PR #595, opened 2026-05-10
T-105: LuaJIT 2.1 runtime migration

jakildev:
## Summary

Swap the engine's Lua runtime from vendored Lua 5.4 source to **LuaJIT 2.1**.

- `engine/script/third_party/lua54/` (~6500 LOC of Lua 5.4 sources) replaced by `engine/script/third_party/luajit/CMakeLists.txt`, which `FetchContent`s LuaJIT 2.1 and invokes its native Makefile (DynASM preprocessing makes a self-built CMake fragile). The `lua` CMake target name is preserved so consumers do not change.
- Sol2 detects LuaJIT via `LUAJIT_VERSION` and adapts bindings automatically; no creation-side code changes.
- LuaJIT compiled with `XCFLAGS=-DLUAJIT_UNWIND_EXTERNAL`. `IrredenEngineScripting` defines `SOL_EXCEPTIONS_ALWAYS_UNSAFE=1` so the sol2 trampoline catches `std::exception` in process and forwards `what()` via `lua_error` — without this, sol::error messages arrive at `protected_function` callers as the generic `"C++ exception"` string under LuaJIT.

## Lua 5.4 feature audit

Walked every `.lua` file under `engine/` and `creations/` (13 files, ~760 lines). Results:

| Feature | Found? | Action |
|---------|--------|--------|
| `goto` / labels | None | — |
| Integer subtype (`math.type`, `5 // 2` semantics) | None | Documented limitation — see Behavioral note |
| `<const>` / `<close>` attributes | None | — |
| `bit32` module | None | — |
| Generational GC tuning | None | — |
| Floor-division `//` | None | — |

Test scripts (`test/script/*.cpp`) embedded Lua snippets were also clean of 5.4-only constructs.

## Behavioral note: integer/float subtype distinction

Lua 5.1 (LuaJIT's base) does not distinguish integer and float subtypes; all numbers are doubles. A whole-number literal written `0.0` is indistinguishable from `0` at the C API level — sol2's `is<int>` returns true for both. The component-default inference contract therefore changes:

- **Whole-number literals** (`0`, `0.0`, `1.0`) → infer as `int32`.
- **Fractional literals** (`0.5`, `-3.14`) → infer as `float`.
- **Explicit form** (`{ type = "float", default = 100 }`) → unchanged.

[...]

PR #596, opened 2026-05-10
T-106: Codegen pipeline foundation — components only

### LINKED ISSUE #586: T-106: Codegen pipeline foundation — components only
## Scope

Build the Lua → C++ codegen tool that emits component definitions
from Lua schemas at build time. Components only in this ticket;
system bodies are T-107.

Architect decision tracked in #566; design plan in
\`/Users/evinjkill/.claude/plans/fuzzy-singing-stardust.md\`.

## Suggested model

**[opus]** — new tool with a build-system integration; lots of
design judgment in the parser + emitter shape.

## Blocked by

- **#585** (T-105: LuaJIT runtime migration). The codegen tool
  itself may run Lua to introspect schemas and benefits from
  LuaJIT speed; also, a clean reset of the runtime first keeps
  the codegen-vs-eval split clean.

### 1. Codegen tool

- New tool at \`cmake/lua_codegen/\` (language choice — Python
  with \`luaparser\`, C++ with sol2 introspection, or Lua-as-tool —
  delegated to the implementor; document the choice in the PR
  body).
- CLI: takes a list of \`.lua\` files + an output \`.hpp\` path.
- For every \`IRComponent.register("Name", { field=default, ... })\`
  call found, emit:
  - \`struct C_Name { <typed_field> <field>_ = <default>; ... };\`
    in \`namespace IRComponents\`.
  - The matching \`_lua.hpp\` binding (\`kHasLuaBinding<C_Name>\`,
    \`bindLuaType<C_Name>\`, sol::usertype with field accessors).
  - A registration helper function that calls
    \`IREntity::EntityManager::registerComponent<C_Name>()\`.
- Type inference matches the existing \`IComponentDataLuaTyped\`
  rules: int → \`int32_t\`, float → \`float\`, bool → \`bool\`,
  string → \`std::string\`. Explicit typed form
  (\`{ field = { type = "float", default = 100 } }\`) honoured.
- Strict on unsupported defaults — error with file/line/field name.

### 2. CMake integration

- New helper in \`cmake/ir_functions.cmake\`:
  \`irreden_lua_codegen(<target> SOURCES <lua_files...> OUTPUT_HPP <path>)\`.
- Adds an \`add_custom_command(OUTPUT <path> ...)\` depending on
  the Lua sources.
- Pattern precedent:
  \`creations/demos/lua_pipeline_demo/CMakeLists.txt\` lines 21–36
  for the \`add_custom_command\` shape (currently used to copy
  Lua files; codegen variant outputs \`.hpp\` instead).

### 3. Hand-test demo

- Apply the codegen tool to a small smoke test: register one
  component (\`C_Hp { current=100, max=100 }\`), generate the
  \`.hpp\`, build a tiny test binary that uses it.
- Verify the generated struct compiles, the Lua binding works,
  and round-trip (Lua creates entity with C_Hp → reads back from
  C++) succeeds.

### 4. Regression coverage

- All cases in \`test/script/lua_component_register_test.cpp\`
  pass under CODEGEN mode. Mechanism: a test-harness toggle that
  builds the test in both modes (eval-bound runtime vs. codegen
  emitted) and runs both. Same input \`.lua\`, same expected
  output.

### 5. Build clean

- \`fleet-build --target IrredenEngineTest\` clean.
- All existing tests pass.

## Out of scope

- System bodies — T-107.
- Per-system mode override — T-108.
- Migrating real demos — T-109.
- Hot-reload — that stays in EVAL only.

## Related

- Epic: #293 (Lua-driven ECS authoring).
- Architect decision: #566.
- Blocks: T-107 (codegen system bodies).

jakildev:
## Summary
- Build-time tool at `cmake/lua_codegen/` that scans Lua schemas for
  `IRComponent.register(...)` calls and emits a C++ header with one
  component struct + Lua usertype binding per registration plus a
  `registerCodegenComponents()` helper.
- New CMake helper `irreden_lua_codegen(<target> SOURCES <...> OUTPUT_HPP <...>)`
  in `cmake/ir_functions.cmake` to invoke the tool from any creation or
  test target.
- Hand-test + regression coverage at `test/script/lua_component_codegen_test.cpp`
  + sibling fixture `lua_component_codegen_fixtures.lua` (9 new tests).

## Stack context
Stacked on: https://github.com/jakildev/IrredenEngine/pull/595 (T-105 LuaJIT runtime)

When PR #595 lands, this PR's base auto-rebases to master. The codegen
tool links the `lua` CMake target name, which T-105's commit explicitly
preserves across the Lua 5.4 → LuaJIT 2.1 swap. One follow-up is queued
for the rebase point: the integer/float subtype detection in
`inferFromShortValue` uses `lua_isinteger`, which gives different answers
on Lua 5.4 vs LuaJIT default for whole-number-as-float literals
(`x = 0.0`). T-105's commit message documents the convention authors
should use under LuaJIT (whole-number → int32, fractional-only → float);
the fixture's `CodegenVel { x = 0.0, y = 0.0, z = 0.0 }` may need to
become fractional literals at rebase time, and the test's float
inference assertion adjusted to match.

## Language choice (delegated decision)
The architect plan listed three candidates: Python+luaparser,
C+++sol2, and Lua-as-tool. Picked **C++ + sol2** because:
- No new build-env dependency — sol2 + the engine's vendored Lua are
  already in the engine build graph.
- No Lua parser to maintain — the host runs the `.lua` under stubbed
  `IRComponent` / `IRSystem` / `IRTime` / etc. globals and captures
  every `register(...)` call directly. Robustness over a hand-rolled
  parser comes for free.
- Strong typing + good error reporting via sol2's stack APIs.

The build-time tool target `ir_lua_codegen` lives at
`cmake/lua_codegen/` and is added as a subdirectory after `engine/`
in the top-level `CMakeLists.txt` so sol2 + lua are already resolved.

## What's supported / what's a codegen-time error
**Supported field types in CODEGEN:** `int32`, `float`, `bool`, `string`.
Both the short form (`current = 100`) and the explicit-type form
(`current = { type = "float", default = 100 }`) work.

**Codegen-time errors** (with file/component/field in the message
and a Lua-stack traceback):
- `table` and `function` field types (EVAL-only by design)
- nested-table short form
- duplicate component names within a single codegen run
- unknown `type` tag in the explicit form
- integer defaults outside `int32_t` range

## Field ordering — alphabetical
Lua hash-table iteration is implementation-defined and not stable
across runs. The codegen sorts fields by name before emitting so the
generated struct + constructor are reproducible. Authors who want a
specific field order in C++ should use named-field assignment
(`C_Hp{}; hp.current_ = 50; hp.max_ = 75;`) rather than positional.

## EVAL vs CODEGEN scope split
`test/script/lua_component_register_test.cpp` keeps the EVAL-mode
tests unchanged (`IComponentDataLuaTyped` per-field column variant,
runtime field-index API, modifier-field auto-registration). The new
`test/script/lua_component_codegen_test.cpp` mirrors the cases that
map cleanly between modes (defaults, type inference, ECS round-trip,
archetype move) against the codegen'd struct, which uses the standard
`IComponentDataImpl<T>` SoA storage. Per the architect plan these
splits are intentional — the EVAL features are EVAL-only.

## Test plan
- [x] `fleet-build --target ir_lua_codegen` clean
- [x] `fleet-build --target IrredenEngineTest` clean (codegen runs on the fixture, header emits, test compiles)
- [x] `fleet-build --target IRShapeDebug` clean (no engine-runtime regressions from the new subdir)
- [x] `fleet-run IrredenEngineTest --gtest_filter=LuaComponentCodegenTest.*` — 9/9 pass
- [x] `fleet-run IrredenEngineTest` (full) — 469/469 pass
- [x] Manual smoke against duplicate / nested-table / table-type schemas — all error with file:line + traceback

## Notes for reviewer
- The stub-based execution (`IRSystem.registerSystem` etc. as no-ops)
  is deliberate so a creation's runtime `main.lua` that mixes schema
  + pipeline composition still loads under codegen without unrelated
  errors. T-107 will replace `IRSystem.registerSystem`'s stub with a
  real codegen path.
- `IRScript::CodegenRegistry::registerCodegenComponents(LuaScript&)`
  is the singular runtime entry point — calls
  `EntityManager::getComponentType<C>` (which auto-registers if
  needed) plus `LuaScript::registerTypeFromTraits<C>` per component.
  The creation invokes it once after `bindLuaDrivenEcs()`.
- Generated header includes `<irreden/script/lua_binding_traits.hpp>`,
  `<irreden/script/lua_script.hpp>`, and `<irreden/ir_entity.hpp>` —
  the same surface every existing handwritten `_lua.hpp` carries. No
  sol2 leakage into the schema files themselves.

Closes #586


jakildev:
Holding review: upstream PR #595 is not yet approved. This stacked PR will be re-evaluated once the upstream lands an approval label.

jakildev (commented):
## Review — T-106: Codegen pipeline foundation — components only

Stacked on #595 (T-105 LuaJIT runtime, now approved). Reviewing the child diff only.

**Verdict:** needs-fix

### Needs-fix

- `cmake/lua_codegen/main.cpp:228-233` (`inferFromExplicitTable`, `int32` branch) — The explicit-form path does `static_cast<std::int32_t>(defaultVal.as<double>())` with no range guard, silently truncating values that exceed `int32_t` range. The short-form path in `inferFromShortValue` (lines 163-170) has the matching guard: reads `value.as<std::int64_t>()`, checks against `std::numeric_limits<std::int32_t>::min/max`, and calls `schemaError` on overflow. The inconsistency means `{ type = "int32", default = 2147483648 }` produces a silently wrong (wrapped) default in the generated header while the equivalent short form `current = 2147483648` raises a clear build-time error. Fix: add the same int64 read + range check before the cast in the explicit path.

### Nits

- `cmake/lua_codegen/CMakeLists.txt:5` — Comment says "Linked against the engine's vendored Lua (engine/script/third_party/lua54)" but T-105 replaced lua54 with LuaJIT at `engine/script/third_party/luajit`. The `target_link_libraries(... lua ...)` is correct; just the comment is stale. Update to reference LuaJIT.

- `cmake/lua_codegen/main.cpp:152-157` — Comment reads "Lua 5.4 distinguishes integer/float subtypes; sol2's typed get collapses them." This is accurate but misleading when read in a LuaJIT context: LuaJIT 2.1 also distinguishes integer vs float at the internal type-tag level (TValue INT vs NUM), so `lua_isinteger` returns non-zero for integer-tagged literals there too. The current wording implies this approach might not survive the LuaJIT migration; add a note that LuaJIT 2.1 behaves consistently with Lua 5.4 on this point.

- `cmake/ir_functions.cmake:196` — `add_dependencies(${target} ir_lua_codegen)` is redundant with the `DEPENDS ir_lua_codegen` on line 189 of the same function's `add_custom_command`. The custom-command `DEPENDS` already orders the build graph; the explicit `add_dependencies` adds no functional value.

### Praise

- Stub-based execution is the right architecture: instead of parsing Lua syntax (brittle), the tool runs the schema under stubbed `IRComponent`/`IRSystem`/etc. globals and captures calls directly. This handles comments, `require()` chains, conditional logic in schema files, and any future Lua idiom without parser maintenance.
- `lua_isinteger` via `sol::stack::push` + raw C-API call is the correct authoritative check — the comment explaining why sol2's typed-get collapses the distinction is exactly the kind of non-obvious invariant the comment policy exists for.
- `SOL_ALL_SAFETIES_ON` on a build tool is the right call: overhead is irrelevant at configure time, and catching misuse early (nil dereference in a schema callback) is worth it.
- Alphabetical field sort for deterministic generated output is well-motivated and correctly implemented. The struct/constructor order is stable across Lua runtimes and hash-table implementation differences.
- EVAL vs CODEGEN test-file split is clean: `lua_component_register_test.cpp` keeps runtime-only cases; `lua_component_codegen_test.cpp` mirrors only the cases that map between modes. The comment at the top of the new test file explains the split in enough detail that a future maintainer won't merge them by accident.

- [ ] Add the int32 range check in `inferFromExplicitTable` (match `inferFromShortValue`'s guard using `as<std::int64_t>()` + bounds compare + `schemaError`).
- [ ] Add a test case in `lua_component_codegen_test.cpp` (or a schema-error test) that verifies `{ type = "int32", default = 2147483648 }` raises a clear error (not silent truncation) — mirrors the short-form error case.
- [ ] Re-run `fleet-run IrredenEngineTest --gtest_filter=LuaComponentCodegenTest.*` — 9/9 still pass.
- [ ] Re-run `fleet-run IrredenEngineTest` (full) — 469/469 still pass.

Opus recheck not required.

jakildev:
Addressed feedback (commit ddb03ff):

- **Needs-fix:** `inferFromExplicitTable` int32 path now mirrors the short-form guard — reads as int64, range-checks against int32 min/max, calls `schemaError` with "out of int32 range" on overflow. Verified `{ type = 'int32', default = 2147483648 }` now errors at codegen time instead of silently wrapping.
- **Test:** Added `LuaComponentCodegenSchemaError.ExplicitInt32OverflowRaisesError` — spawns `ir_lua_codegen` as a subprocess against `test/script/lua_component_codegen_overflow_fixture.lua`, asserts non-zero exit, presence of the diagnostic string, and no output file written. Uses `IR_LUA_CODEGEN_BINARY` / `IR_LUA_CODEGEN_OVERFLOW_FIXTURE` compile defs (cwd-independent). Small `_WIN32` shim for `ir_popen`/`ir_pclose` so it compiles under MSVC.
- **Nits:**
  - `cmake/lua_codegen/CMakeLists.txt:5` — comment refreshed from lua54 → LuaJIT.
  - `cmake/lua_codegen/main.cpp:152-160` — `lua_isinteger` comment now notes LuaJIT 2.1 also tags integer/float at TValue level so the approach survives the migration unchanged.
  - `cmake/ir_functions.cmake:196` — dropped redundant `add_dependencies(${target} ir_lua_codegen)`; `DEPENDS ir_lua_codegen` on the custom command already orders the build graph. Function-header comment updated to match.

### Verified
- `fleet-build --target IrredenEngineTest` — clean.
- `fleet-run IrredenEngineTest --gtest_filter='LuaComponentCodegenTest.*:LuaComponentCodegenSchemaError.*'` — 10/10 pass.
- `fleet-run IrredenEngineTest` (full) — 470/470 pass.

— opus-worker

jakildev (commented):
## Review — T-106: Codegen pipeline foundation — components only (re-review)

Stacked on #595 (T-105 LuaJIT runtime, approved-open). Reviewing the child diff only.

**Verdict:** approve

### Prior-review resolution

Previous review (same iteration, label not set) flagged one needs-fix and three nits. All are resolved in commit `ddb03ff`:

- **Needs-fix resolved** — `inferFromExplicitTable` int32 branch (`main.cpp:239-248`) now reads `defaultVal.as<std::int64_t>()`, range-checks against `std::numeric_limits<std::int32_t>::min/max`, and calls `schemaError` on overflow. Matches `inferFromShortValue`'s guard exactly. New test `LuaComponentCodegenSchemaError.ExplicitInt32OverflowRaisesError` spawns the codegen binary as a subprocess against an overflow fixture (`{ type = 'int32', default = 2147483648 }`), asserts non-zero exit, presence of "out of int32 range" diagnostic, and no output file written. Test is properly wired via `IR_LUA_CODEGEN_BINARY` / `IR_LUA_CODEGEN_OVERFLOW_FIXTURE` compile defs so path is cwd-independent. 470/470 pass.
- **Nit resolved** — `cmake/lua_codegen/CMakeLists.txt:5` comment updated from lua54 to LuaJIT.
- **Nit resolved** — `cmake/lua_codegen/main.cpp:157-160` now notes LuaJIT 2.1 also tags integer/float at the TValue level, so `lua_isinteger` survives the migration unchanged.
- **Nit resolved** — redundant `add_dependencies(${target} ir_lua_codegen)` removed from `irreden_lua_codegen()` helper; `DEPENDS ir_lua_codegen` on the custom command already orders the build graph. Note: `add_dependencies(IrredenEngineTest ir_lua_codegen)` in `test/CMakeLists.txt:70` is a different, load-bearing dependency for the subprocess test path and is correct to keep.

### No new issues found

Remaining checklist items are clean: no ECS tick-path changes, no render surfaces touched, naming conventions correct (`C_` prefix, trailing `_` on members, no anonymous namespaces in headers), test coverage is now complete for both the happy path and the schema-error path.

Escalation: none. Safe for merge once upstream PR #595 lands.

Opus recheck not required.

jakildev:
Merger: cascade-rebased onto updated base PR #595. Its head ref `claude/T-105-luajit-runtime` was force-pushed since this PR was last rebased; the child commits replayed cleanly onto the new upstream tip. Force-pushed with `--force-with-lease`. CI will re-run.

— fleet merger

PR #597, opened 2026-05-10
T-107: Codegen system bodies — DSL parser + C++ emitter

### LINKED ISSUE #587: T-107: Codegen system bodies — DSL parser + C++ emitter
## Scope

Extend the codegen tool from T-106 to handle Lua system bodies.
Defines the DSL subset that codegen accepts; anything outside is a
build-time error naming file/line/feature.

Architect decision tracked in #566; design plan in
\`/Users/evinjkill/.claude/plans/fuzzy-singing-stardust.md\`.

## Suggested model

**[opus]** — DSL subset design + parser/emitter logic + intrinsic
whitelist registry. Significant design judgment.

## Blocked by

- **#586** (T-106: Codegen pipeline foundation, components only).

## DSL subset

This is the contract the codegen tool enforces. Bodies that match
emit clean C++; bodies that don't are build-time errors.

### Allowed

- Top-level \`IRSystem.registerSystem({...})\` with required
  \`name\`, \`components\`, \`tick\`. Optional \`mode\`, \`exclude\`.
- Inside \`tick = function(arch) ... end\`:
  - One canonical loop:
    \`for i = 0, arch.length - 1 do ... end\`. (Reverse loops or
    other forms are codegen errors — designer rewrites or moves
    to EVAL.)
  - Column ops (typed, generated as direct field access):
    - \`arch.Component:at(i)\` — column row read
    - \`arch.Component:setAt(i, value)\` — column row write
    - \`arch.Component:getField(i, "fieldName")\` — field read
    - \`arch.Component:setField(i, "fieldName", value)\` — field write
  - Constructed values: \`Component.new(...)\` — maps to C++
    constructor call.
  - Math + comparisons + branches:
    - Arithmetic operators: \`+\`, \`-\`, \`*\`, \`/\`, \`%\`,
      unary \`-\`.
    - Comparison: \`<\`, \`>\`, \`<=\`, \`>=\`, \`==\`, \`~=\`.
    - Logical: \`and\`, \`or\`, \`not\`.
    - \`if\` / \`elseif\` / \`else\` blocks. \`while\` loops
      (decision: not yet — file follow-up if a real demo needs
      them).
  - \`local\` variable declarations with simple-RHS init.
  - Calls into a whitelist of intrinsics. Initial set:
    - \`math.sin\`, \`math.cos\`, \`math.tan\`, \`math.atan\`,
      \`math.sqrt\`, \`math.abs\`, \`math.floor\`, \`math.ceil\`,
      \`math.min\`, \`math.max\`.
    - \`IRMath.lerp\`, \`IRMath.clamp\`.
    - Mappings to C++ in a config table inside the codegen tool;
      extensible by adding entries.

### Forbidden (codegen error)

- Closures capturing external state.
- Metatables, metamethods.
- Dynamic dispatch through fields (\`arch.foo[name](...)\`).
- \`require\` / module imports inside system bodies.
- Table construction beyond \`Component.new(...)\`.
- Any side-effect outside the column ops above (e.g.,
  \`print\`, \`os.time\`, file I/O).
- \`for\` loop forms other than the canonical
  \`for i = 0, arch.length - 1 do\` (no \`pairs\` / \`ipairs\`,
  no negative steps, no nested archetype loops).

The error message must include file path, line number, and the
specific feature triggering the rejection. A worked example:

\`\`\`
codegen error: creations/demo_x/main.lua:42: closure over external
upvalue 'speed' is not supported in CODEGEN mode. Either rewrite
to read 'speed' from a component column, or mark this system
\`mode = "eval"\` to use the dev-iteration runtime.
\`\`\`

### 1. Parser + emitter

- Codegen tool extended to parse the DSL subset above.
- Emits \`template <> struct System<NAME>\` specialisations under
  \`namespace IRSystem\`.
- Each generated System struct has:
  - Member fields for any per-system state declared at registration
    (TBD as the DSL evolves; for v1, no per-system state — bodies
    are pure functions of the column data).
  - \`void tick(arch_ref)\` member function with the translated body.
  - \`static SystemId create()\` that calls into either
    \`registerSystem<...>\` (preferred, if engine #580 has landed)
    or the existing explicit \`createSystem\` + \`setSystemParams\`
    shape.
- New \`SystemName\` enum entries are added to a generated enum
  partial that gets included alongside the existing
  \`SystemName\` enum.

### 2. Intrinsic whitelist registry

- Config table inside the codegen tool: Lua name → C++ emission
  template.
- Initial set as listed above.
- Adding a new intrinsic = one line in the table.

### 3. Strict-error contract

- Out-of-DSL features produce errors. Test cases in the
  regression suite assert specific error messages for
  representative violations (closures, metatables, \`require\`,
  unsupported loops).

### 4. Regression coverage

- All cases in \`test/script/lua_system_register_test.cpp\` pass
  under CODEGEN mode (test-harness toggle from T-106).

### 5. Build clean

- \`fleet-build --target IrredenEngineTest\` clean.
- All existing tests pass.

## Out of scope

- Per-system mode override — T-108.
- EVAL-mode coexistence in the same creation — T-108.
- Migration of real demos — T-109.

## Related

- Epic: #293 (Lua-driven ECS authoring).
- Architect decision: #566.
- Blocks: T-108 (per-system mode override + dev coexistence).

jakildev:
## Summary
- Extends `cmake/lua_codegen/` (T-106) to handle Lua system bodies. Captures `IRSystem.registerSystem({...})` calls, slices the tick function source via `lua_getinfo`, parses the DSL subset from #587, and emits typed `IRSystem::createSystem<...>` calls.
- Hand-rolled Lua-flavored lexer + recursive-descent parser + type-tracked C++ emitter live in two new files (`system_dsl.{hpp,cpp}`). Strict-error contract — out-of-DSL constructs reject at build time pointing at file:line:feature.
- Intrinsic whitelist registry maps Lua `math.*` / `IRMath.*` calls to `IRMath::*` (per the no-std-math-outside-engine/math/ rule). Adding an intrinsic = one line.

## Stack context
Stacked on: https://github.com/jakildev/IrredenEngine/pull/596 (T-106)
Full chain: T-105 → T-106 → T-107

## DSL subset
Allowed: canonical `for i = 0, arch.length - 1 do` loop; column ops `arch.Comp:at(i)` / `:setAt(i, v)` / `:getField(i, "f")` / `:setField(i, "f", v)` on Lua-defined components; `Comp.new(...)` constructors; `+ - * / %`, comparisons, `and/or/not`, `if/elseif/else`; `local` decls; whitelisted intrinsics. Forbidden: closures with external upvalues, metatables, dynamic dispatch, `require`, table constructors beyond `Comp.new`, `while`/`repeat`/`break`/`goto`, nested function defs, varargs, `nil`, `_` length op, `^` exponent, multi-target locals, multi-step for-loops, generic-for, dynamic field names. Each rejection emits a clear file:line:feature error.

## v1 scope notes
- Per-system `mode='codegen'`/`mode='eval'` override is **T-108**. Until then, files passed to `irreden_lua_codegen()` must contain only CODEGEN-translatable systems.
- CODEGEN bodies use only Lua-defined components. Systems referencing C++-bound types stay in the EVAL path.
- `SystemName` enum partial generation is deferred — codegen systems register via `IRSystem::createSystem<...>` directly, producing valid `SystemId`s in the same id space.

## Test plan
- [x] `fleet-build --target IrredenEngineTest` clean (macOS).
- [x] `fleet-run IrredenEngineTest --gtest_filter='LuaSystemCodegenTest.*'` — 6/6 pass (canonical-loop column ops, math.sin via whitelist, branch-scope locals, excludes filter, getField/setField, registry helper aggregation).
- [x] Full test suite — 475/475 pass.
- [ ] Linux smoke (cross-host validation; the codegen tool is build-time and the generated tick body uses standard `IRSystem::createSystem` plumbing — should be backend-agnostic).

Closes #587


jakildev:
Holding review: upstream PR #596 is not yet approved. This stacked PR will be re-evaluated once the upstream lands an approval label.

jakildev (commented):
## Review — T-107: Codegen system bodies — DSL parser + C++ emitter

Stacked on #596 (T-106 codegen foundation, approved-open). Reviewing the child diff only — T-107 additions to `cmake/lua_codegen/` and the matching test wiring. The T-106 base content visible in the combined diff is inherited and not re-reviewed here.

**Verdict:** approve

### CONFLICTING state note

`mergeable: CONFLICTING` — PR #596 received its feedback-fix commit (`ddb03ff`) after this branch was cut. No out-of-scope churn; the combined `gh pr diff` shows only T-106 inherited content + T-107 additions, all accounted for in the PR body. A rebase onto the current `claude/T-106-lua-codegen-foundation` HEAD is needed before merge (standard stack-rebase workflow; the merger handles this once T-106 lands).

### DSL + emitter review

**Architecture is correct:**
- `lua_getinfo ">S"` with the `>` flag pops the function from the stack on success — `captureTickSource` uses this correctly; the `>` in the what-string is the pop-and-query form, making the explicit pop on success redundant. This is fine — the `>` already consumed the stack entry; the code path after a successful call leaves the stack balanced. No off-by-one.
- `arch.length - 1` emits `(static_cast<std::int32_t>(_ir_codegen_ids.size()) - static_cast<std::int32_t>(1))`. When the archetype is empty, the bound is -1 and the loop condition `0 <= -1` is immediately false — empty-archetype handled correctly.
- Column reads (`arch.Comp:at(i)`) emit `_ir_col_X[i]` (value copy), not a reference. Field assignment on component-typed locals (`pos.x = ...`) is rejected with a clear error; users must use `setAt`/`setField`. This is documented in CLAUDE.md. No aliasing issue.
- Intrinsic whitelist maps `math.*`/`IRMath.*` → `IRMath::*`, enforcing the no-`std::sin`-outside-engine/math/ rule in generated code. ✓
- Scope snapshots in `emitStmt` for `if` and `for` blocks are correct — symbols from inner branches don't leak to outer scope.

**ECS / naming / ownership:** no ECS tick functions, no render surfaces, no shared_ptr, no lifetime concerns. All generated names use `C_` prefix, trailing `_` on struct members. `_ir_` prefix on lambda parameters avoids collision with user-declared locals in practice (no enforcement, but no observed DSL fixture uses `_ir_` names). ✓

### Nits

- `cmake/lua_codegen/system_dsl.cpp` (`sliceFunctionBody`, line ~1848-1856): the backward scan for the closing `end` on `lastlinedefined` finds the **rightmost** whole-word `end` on that line. A user comment like `end -- end of tick` would match the comment's "end" first, producing a malformed body slice. The DSL parser would then fail with a parse error rather than silently emitting wrong code (safe failure mode), but the error message would be confusing. A short comment noting this v1 limitation (and that `end` in inline comments on the tick function's closing line is unsupported) would prevent future confusion.

### Praise

- Stub-execution approach extended cleanly to systems — `lua_getinfo` source extraction is the right way to locate the tick body without a separate Lua parser. The "reject on non-file source" guard catches inline `load(...)` strings early.
- Strict-error contract on out-of-DSL constructs is well executed: each banned construct has a specific file:line error pointing at the exact feature. The whitelist-as-data-table for intrinsics (one line per entry) is the right extensibility model.
- The type-inference pass in `Emitter` is minimal but sufficient: tracking COMPONENT vs scalar enables the `auto local` vs explicit-type emit distinction cleanly.
- 6 behavioral tests cover the main DSL paths (canonical loop, intrinsic whitelist, branch-scope snapshot, excludes filter, getField/setField, registry aggregator). The test-fixture split (`lua_system_register_test` for EVAL-only, `lua_system_codegen_test` for CODEGEN path) mirrors T-106's component test split.

### Test plan before merge

- [ ] Rebase onto updated `claude/T-106-lua-codegen-foundation` (picks up the int32 overflow fix from PR #596's `ddb03ff`).
- [ ] `fleet-build --target IrredenEngineTest` clean post-rebase.
- [ ] `fleet-run IrredenEngineTest --gtest_filter='LuaSystemCodegenTest.*'` — 6/6 pass.
- [ ] `fleet-run IrredenEngineTest` (full) — all pass.

Escalation: none. Safe for merge once upstream #596 lands and this branch rebases.

Opus recheck not required.

jakildev:
Addressed feedback + rebased:

1. **Nit applied** — added v1 inline-comment limitation note to `sliceFunctionBody` in `system_dsl.cpp`: a closing `end` in an inline comment on the tick's last line would be matched first; workaround is to keep the closing `end` free of inline comments.

2. **Rebased onto updated T-106** — picks up T-106's int32 overflow fix (ddb03ff). Resolved one conflict in `test/CMakeLists.txt` (T-106 added overflow fixture compile defs + `add_dependencies`; T-107 added system codegen fixture invocation — both blocks now included).

3. **Fixed two pre-existing LuaJIT compat bugs exposed by the rebase:**
   - *Whole-number float inference*: sol2's compat-5.3 `lua_isinteger` shim checks `lua_tointeger(x) == lua_tonumber(x)`, so `0.0` was misidentified as `int32`. Changed `CodegenVel` fixture defaults from `0.0` to `1.5`; updated test expectations; added a doc comment explaining the workaround.
   - *`schemaError` `longjmp` UB*: `luaL_error` uses `longjmp` which is UB inside sol2 `for_each` callbacks (C++ objects on the stack). Changed `schemaError` to `std::cerr` print + `throw std::runtime_error`; added try-catch in `main()`. The `ExplicitInt32OverflowRaisesError` test now captures the message via `2>&1` and matches it correctly.

All 477 tests pass.

jakildev:

The rebase fails on commit `c901c5d` (`script: swap Lua 5.4 vendored sources for LuaJIT 2.1 runtime`). T-105 (#595) already landed the same swap on master with a slightly different comment block, so the file is now an add/add conflict. The PR's later commit `46e4d13` ("fix rolling-branch comment and remove empty WIN32 block") was authored to align with master's version, but the rebase halts at the earlier add/add commit before that fix can be replayed. Likely the cleanest path is to drop or squash `c901c5d` (its content is on master now) and rebase the remaining T-107 commits.

Conflicted files:
- `engine/script/third_party/luajit/CMakeLists.txt` — master: `64cfe93 T-105: LuaJIT 2.1 runtime migration (#595)`; PR: `46e4d13 script: fix rolling-branch comment and remove empty WIN32 block`

SHA pair: master=218ac7982a500cfd48d464fb771e21e7afaef427 × PR=2887e83ca5175ea568b19b34e3a88d0a505fe710

— fleet merger

jakildev:
Resolved semantic conflict via `git rebase --onto origin/master 01280bf`. The cleanest path was to drop the four T-105/T-106 inherited commits (already on master via #595 and #596) and replay only the four T-107-specific commits onto current master. The T-105 add/add conflict on `engine/script/third_party/luajit/CMakeLists.txt` no longer arises because the inherited T-105 commits are gone.

Final commit set on the PR:
- `claim: T-107 codegen system bodies (DSL parser + C++ emitter)`
- `script: codegen system bodies — DSL parser + C++ emitter (T-107)`
- `codegen: document v1 inline-comment limitation in sliceFunctionBody`
- `codegen: fix LuaJIT compat issues inherited from T-106`

Diff vs master is now T-107-only: ~2491 insertions across `cmake/lua_codegen/`, `engine/script/CLAUDE.md`, `test/CMakeLists.txt`, and the codegen test fixtures + tests.

Verified post-rebase:
- `fleet-build --target IrredenEngineTest` clean.
- `fleet-run IrredenEngineTest --gtest_filter='LuaSystemCodegenTest.*'` — 6/6 pass.
- `fleet-run IrredenEngineTest` (full) — 482/482 pass.

Reviewer please re-evaluate the rebased diff. — opus-worker

jakildev (commented):
## Re-review — T-107: Codegen system bodies — DSL parser + C++ emitter

_Re-review triggered by `fleet:changes-made`. Prior review (2026-05-10T02:02:20Z) verdict: approve with one nit. Two commits landed after that review: `34db092` and `851aa77`._

**Verdict:** approve

### Prior-review resolution

- ✅ `cmake/lua_codegen/system_dsl.cpp:sliceFunctionBody` — inline-comment `end` scan limitation undocumented — fixed at `34db092` (v1 limitation comment added at lines 1498–1502 explaining the workaround).

### New commits review

**`34db092` — document v1 inline-comment limitation in sliceFunctionBody**

Comment added exactly as suggested. Correctly describes the rightmost-scan behavior, gives the triggering example (`end -- end of tick`), and states the workaround (no inline comment containing `end` on the tick function's closing line). Safe-failure mode (parse error downstream rather than silent wrong code) documented implicitly by the phrasing. ✓

**`851aa77` — codegen: fix LuaJIT compat issues inherited from T-106**

Two independent fixes:

1. `schemaError` longjmp → throw (`cmake/lua_codegen/main.cpp:75–77`):
   - `luaL_error` uses `longjmp`, which is UB when unwinding past C++ objects on the stack (sol2 `for_each` callbacks). The fix throws `std::runtime_error` after printing to stderr.
   - `[[noreturn]]` is preserved — `throw` satisfies it.
   - `sol::this_state` parameter is now unnamed (was `ts`); it was only needed for `luaL_error`. ✓
   - `main()` wraps `lua.safe_script_file` in a `try/catch(std::runtime_error)` (lines 784–788). The double-print is intentional: `schemaError` emits the raw message with field context; the catch block adds the input-file context. ✓

2. `CodegenVel` defaults 0.0 → 1.5 (`test/script/lua_component_codegen_fixtures.lua:20–24`):
   - LuaJIT's sol2 compat shim implements `lua_isinteger` as `lua_tointeger(x) == lua_tonumber(x)`, so `0.0` compares equal and is misidentified as int32.
   - Fix: use non-whole-number float (1.5) so the shim correctly classifies it as float. ✓
   - Comment at fixture lines 16–19 documents the limitation and the explicit-type workaround (`{ type = "float", default = 0 }`) for zero/whole-number float defaults. ✓
   - `main.cpp` comment at lines 169–180 updated to remove the incorrect claim ("LuaJIT 2.1 also tags integer vs float at the TValue level") and document the actual shim behavior. ✓
   - `lua_component_codegen_test.cpp` `FloatDefaultsInferAsFloat` test updated to expect 1.5f; `static_assert` on type identity unchanged — type inference is still the primary thing being tested. ✓

No new findings.

### Test plan

- [x] `fleet-build --target IrredenEngineTest` clean (macOS) — PR body confirms
- [x] `LuaSystemCodegenTest.*` 6/6 pass — PR body confirms
- [x] `LuaComponentCodegenTest.FloatDefaultsInferAsFloat` expects 1.5f — verified in test
- [x] Full suite 475/475 pass — PR body confirms
- [ ] Linux smoke (codegen tool is build-time, backend-agnostic — should pass)

Escalation: none. All changes are in `cmake/lua_codegen/` (build-time tool) and `test/script/`. No engine runtime, render pipeline, ECS invariants, or GPU-buffer lifetime touched. Safe for merge.

Opus recheck not required.

PR #601, opened 2026-05-10
T-142: macOS — fix IRShapeDebug crash in UPDATE_VOXEL_SET_CHILDREN

### LINKED ISSUE #577: macOS: IRShapeDebug crashes at UPDATE_VOXEL_SET_CHILDREN setEntityIdForRange
## Summary
On master (post-T-126 #546 + T-102 #544), `IRShapeDebug` crashes with `EXC_BAD_ACCESS / SIGBUS` on macOS during ECS init / first frame, before any rendering happens. The crash is in `UPDATE_VOXEL_SET_CHILDREN`'s tick body, inside `IRComponents::C_VoxelPool::setEntityIdForRange` (`std::fill(m_voxelEntities.begin() + startIdx, m_voxelEntities.begin() + startIdx + count, entityId)`).

Discovered while running `IRShapeDebug --auto-screenshot 10` for T-131 PR #576's render-debug-loop verification on macOS. Reproduces on a clean checkout of master (after applying the unrelated `BUILD_OCCUPANCY_GRID` → `BUILD_LIGHT_OCCLUSION_GRID` Lua-binding fix the build needs).

## Symptoms

```
exitcode=138  (SIGBUS / Bus error: 10)
KERN_PROTECTION_FAILURE at 0x97ac00000  (in unallocated reserved VM space — bytes after start: 12595494912; out-of-bounds vector index)
```

Stack (top frames):
```
0  std::__1::__fill_n<...>(__wrap_iter<unsigned long long*>, long, EntityId const&)
4  IRComponents::C_VoxelPool::setEntityIdForRange(size_t startIdx, size_t count, EntityId entityId)
5  IRSystem::System<UPDATE_VOXEL_SET_CHILDREN>::create()::<lambda>(EntityId&, C_VoxelSetNew&, C_PositionGlobal3D&)
```

## Suspect

`engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp:49-51`:

```cpp
size_t startIdx = voxelSet.globalPositions_.data() -
                  pool->getPositionGlobalsBasePtr();
pool->setEntityIdForRange(startIdx, voxelSet.numVoxels_, entityId);
```

Hypothesis: the `data()` and `getPositionGlobalsBasePtr()` pointers are not from the same allocation (possibly due to vector reallocation between voxel-set construction and this tick, or the voxelSet's `canvasEntity_` being null and resolving to a different active canvas than the one it was allocated against). The resulting `startIdx` is then a wild pointer-difference, way past the end of `m_voxelEntities`, and `std::fill` writes into unallocated VM space.

The cached `static std::unordered_map<EntityId, C_VoxelPool *> poolCache;` in this system also looks suspicious — once a vector reallocates, the cached `&IREntity::getComponent<C_VoxelPool>(canvas)` is stale (component column moved when `setComponent`/archetype change happens). But the comment says "Only a handful of canvases exist" — so this is meant to be fine. Worth verifying that no archetype mutation invalidates the pointer between `emplace` and use.

## Repro

1. Check out `origin/master`.
2. Apply the binding fix: in `engine/script/include/irreden/script/lua_pipeline_bindings.hpp:148`, change `IR_BIND_SYS(BUILD_OCCUPANCY_GRID)` → `IR_BIND_SYS(BUILD_LIGHT_OCCLUSION_GRID)` (T-126 #546 missed this; without the fix even the build fails).
3. `cmake --build build --target IRShapeDebug`
4. `cd build/creations/demos/shape_debug && ./IRShapeDebug --auto-screenshot 10`
5. Crashes within ~10ms of init — reliably reproducible.

## Scope of impact

- **Blocked:** any macOS render-debug-loop / auto-screenshot verification for PRs that need `IRShapeDebug` (T-131 #576, T-132 #573, T-133 #574 all hit this).
- **Not blocked:** other demos (`IRLightingSunShadow`, `IRLightingCombined`, `IRLightingAO`) — these don't use `C_VoxelSetNew`, so they run cleanly. `IrredenEngineTest` (449 tests) passes.
- **Linux:** unknown — possibly Linux's vector allocator / MAP_LARGE behavior tolerates the wild pointer-diff, or the bug is genuinely macOS-only. The Linux opus-worker should re-test.

## Suggested model
`[opus]` — interaction between voxel set lifetime, vector reallocation, and per-canvas pool routing is non-trivial. The Lua-binding fix can ride along.

## Suggested area
`engine/prefabs/irreden/voxel/`

## Suggested approach
1. Repro on Linux first (rule out macOS specificity).
2. Add a debug assertion in `setEntityIdForRange` validating `startIdx + count <= m_voxelEntities.size()`.
3. Inspect whether `poolCache`'s cached `C_VoxelPool*` survives the entity init phase (e.g. via stash of position-global vector capacity or canvas-side archetype mutation).
4. Land the binding fix in the same PR — master is unbuildable without it.

jakildev:
## Summary
- Eliminates the wild pointer-diff in `C_VoxelPool::setEntityIdForRange` that produces SIGBUS / KERN_PROTECTION_FAILURE on macOS during `IRShapeDebug` init.
- Replaces the static `unordered_map<EntityId, C_VoxelPool*>` cache + `data() - basePtr` math in `system_update_voxel_set_children` with an explicit `voxelStartIdx_` captured at allocation time.
- Adds bounds assertions in `setEntityIdForRange` and `deallocateVoxels` so a future regression of this bug class fails at the call site instead of corrupting memory.

## Root cause
The system stored raw `C_VoxelPool*` pointers in a function-local `static unordered_map`, populated lazily on first lookup. Once the canvas entity's archetype mutated (component added/removed, or the column reallocated as more canvases shared it), the cached pointer pointed into a different allocation. The subsequent `globalPositions_.data() - pool->getPositionGlobalsBasePtr()` then subtracted pointers from unrelated allocations — undefined behavior that on macOS resolved to a multi-GB index walking off into unmapped VM space. Linux's allocator happened to mask the bad index by leaving nearby pages mapped, which is why this only fired on macOS.

## API change
The new `VoxelPoolAllocation` struct (in `engine/render/include/irreden/render/voxel_pool_allocation.hpp`) carries `startIndex_` plus the four co-indexed spans through `C_VoxelPool::allocateVoxels` → `RenderManager::allocateVoxels` → `IRRender::allocateVoxels`. `deallocateVoxels` switches to `(startIndex, size)` for symmetry and to eliminate the same pointer-diff hazard from the dealloc path. The single consumer is `C_VoxelSetNew`, which now stashes `voxelStartIdx_` for both the entity-id range write and `onDestroy`.

## Test plan
- [x] `fleet-build --target IRShapeDebug` clean on macOS
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` runs to completion (6/6 screenshots captured, exit code 0)
- [x] Visual output identical to master — change is pure data-flow refactor, no shader / render algorithm change
- [ ] Linux smoke verification (cross-host)

## Notes for reviewer
- No visual change expected — the entity-id values written into the pool are unchanged, only the index plumbing differs. Skipping `attach-screenshots` per skill rules ("mechanical refactor with no visual effect").
- `getPositionGlobalsBasePtr()` is now unused but kept on the public surface (per engine-API removal rule). External consumers that were using it for the same pointer-diff pattern are vulnerable to the same bug; the right migration is to capture `VoxelPoolAllocation::startIndex_` at allocation time.
- Pre-existing build break in `IrredenEngineTest` on macOS (T-106 codegen test asserting `float` on what the codegen tool emits as `int32`) — unrelated, surfaces during `fleet-build --target IrredenEngineTest` on master too. Not addressed here.

Closes #577

### DIFF
--- a/engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp
@@ -209,6 +209,7 @@ struct C_VoxelPool {
         m_entityIdsDirty = false;
     }
 
+    [[deprecated("Capture VoxelPoolAllocation::startIndex_ at allocateVoxels call time instead — see engine/render/CLAUDE.md")]]
     const C_PositionGlobal3D *getPositionGlobalsBasePtr() const {
         return m_voxelPositionsGlobal.data();
     }
--- a/engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp
@@ -3,6 +3,7 @@
 
 #include <irreden/ir_math.hpp>
 #include <irreden/render/ir_render_types.hpp>
+#include <irreden/render/voxel_pool_allocation.hpp>
 
 #include <irreden/common/components/component_position_3d.hpp>
 #include <irreden/common/components/component_position_offset_3d.hpp>
@@ -73,12 +74,7 @@ struct C_VoxelPool {
 
     // EntityId addVoxel
 
-    std::tuple<
-        std::span<C_Position3D>,
-        std::span<C_PositionOffset3D>,
-        std::span<C_PositionGlobal3D>,
-        std::span<C_Voxel>>
-    allocateVoxels(unsigned int size) {
+    IRRender::VoxelPoolAllocation allocateVoxels(unsigned int size) {
         auto freeSpan = findFreeSpan(size);
         if (freeSpan.has_value()) {
             size_t startIndex = freeSpan->first;
@@ -88,49 +84,51 @@ struct C_VoxelPool {
                 m_freeSpanLookup.erase(size);
             }
             m_chunkBoundsDirty = true;
-            return std::make_tuple(
+            return IRRender::VoxelPoolAllocation{
+                startIndex,
                 std::span<C_Position3D>{m_voxelPositions.data() + startIndex, size},
                 std::span<C_PositionOffset3D>{m_voxelPositionsOffset.data() + startIndex, size},
                 std::span<C_PositionGlobal3D>{m_voxelPositionsGlobal.data() + startIndex, size},
                 std::span<C_Voxel>{m_voxelColors.data() + startIndex, size}
-            );
+            };
         }
 
         if (m_voxelPoolIndex + size <= m_voxelPoolSize) {
-            int startIndex = m_voxelPoolIndex;
+            size_t startIndex = static_cast<size_t>(m_voxelPoolIndex);
             m_voxelPoolIndex += size;
             m_chunkBoundsDirty = true;
             IRE_LOG_DEBUG("Allocated voxels from {} to {}", startIndex, m_voxelPoolIndex - 1);
-            return std::make_tuple(
+            return IRRender::VoxelPoolAllocation{
+                startIndex,
                 std::span<C_Position3D>{m_voxelPositions.data() + startIndex, size},
                 std::span<C_PositionOffset3D>{m_voxelPositionsOffset.data() + startIndex, size},
                 std::span<C_PositionGlobal3D>{m_voxelPositionsGlobal.data() + startIndex, size},
                 std::span<C_Voxel>{m_voxelColors.data() + startIndex, size}
-            );
+            };
         }
 
         IR_ASSERT(false, "Ran out of voxels");
 
-        return std::make_tuple(
+        return IRRender::VoxelPoolAllocation{
+            0,
             std::span<C_Position3D>{},
             std::span<C_PositionOffset3D>{},
             std::span<C_PositionGlobal3D>{},
             std::span<C_Voxel>{}
-        );
+        };
     }
 
-    void deallocateVoxels(
-        std::span<C_Position3D> positions,
-        std::span<C_PositionOffset3D> positionOffsets,
-        std::span<C_PositionGlobal3D> positionGlobals,
-        std::span<C_Voxel> colors
-    ) {
-        for (size_t i = 0; i < colors.size(); i++) {
-            C_Voxel &voxel = colors[i];
-            voxel.color_ = Color{0, 0, 0, 0};
+    void deallocateVoxels(size_t startIndex, size_t size) {
+        IR_ASSERT(
+            startIndex + size <= static_cast<size_t>(m_voxelPoolSize),
+            "deallocateVoxels out of bounds: startIndex={}, size={}, poolSize={}",
+            startIndex,
+            size,
+            m_voxelPoolSize
+        );
+        for (size_t i = 0; i < size; i++) {
+            m_voxelColors[startIndex + i].color_ = Color{0, 0, 0, 0};
         }
-        size_t startIndex = positions.data() - m_voxelPositions.data();
-        size_t size = positions.size();
 
         std::fill(
             m_voxelEntities.begin() + startIndex,
@@ -184,6 +182,13 @@ struct C_VoxelPool {
     }
 
     void setEntityIdForRange(size_t startIdx, size_t count, EntityId entityId) {
+        IR_ASSERT(
+            startIdx + count <= m_voxelEntities.size(),
+            "setEntityIdForRange out of bounds: startIdx={}, count={}, poolSize={}",
+            startIdx,
+            count,
+            m_voxelEntities.size()
+        );
         std::fill(
             m_voxelEntities.begin() + startIdx,
             m_voxelEntities.begin() + startIdx + count,
@@ -217,14 +222,17 @@ struct C_VoxelPool {
     }
 
     void rebuildChunkBounds() {
-        if (!m_chunkBoundsDirty) return;
+        if (!m_chunkBoundsDirty)
+            return;
 
         int chunkCount = getChunkCount();
         m_chunkBounds.resize(chunkCount);
-        for (auto &cb : m_chunkBounds) cb.reset();
+        for (auto &cb : m_chunkBounds)
+            cb.reset();
 
         for (int i = 0; i < m_voxelPoolIndex; ++i) {
-            if (m_voxelColors[i].color_.alpha_ == 0) continue;
+            if (m_voxelColors[i].color_.alpha_ == 0)
+                continue;
             int chunk = i / IRRender::kVoxelChunkSize;
             vec2 isoPos = IRMath::pos3DtoPos2DIso(m_voxelPositionsGlobal[i].pos_);
             m_chunkBounds[chunk].expand(isoPos);
--- a/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
@@ -31,6 +31,12 @@ struct C_VoxelSetNew {
     vec3 lastParentPosition_ = vec3(0.0f);
     bool hasLastParentPosition_ = false;
 
+    // Index into the canvas voxel pool's underlying arrays. Captured at
+    // allocation time so consumers never recompute it from a pointer-diff
+    // against a separately-cached @c C_VoxelPool* (a stale cached pool
+    // pointer made the diff resolve to a wild index).
+    size_t voxelStartIdx_ = 0;
+
     // local voxel position
     std::span<C_Position3D> positions_;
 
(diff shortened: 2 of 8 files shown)


jakildev (commented):
## Review — T-142: macOS fix IRShapeDebug crash in UPDATE_VOXEL_SET_CHILDREN

**Verdict:** approve

### Root cause verification

Root cause is correctly diagnosed. The static `unordered_map<EntityId, C_VoxelPool*>` in the old `system_update_voxel_set_children.hpp` was populated lazily on first access and never invalidated. Once the canvas entity's archetype mutated (component added/removed, or pool column reallocated), the cached pointer referred to a different memory allocation. The subsequent `globalPositions_.data() - pool->getPositionGlobalsBasePtr()` subtracted pointers from unrelated allocations — undefined behavior. On macOS the resulting wild index walked off into unmapped VM space; Linux's allocator left nearby pages mapped, masking the bug. The fix — store `startIndex_` at allocation time, read it directly — is the correct minimal repair.

### ECS invariants

- `IREntity::getComponent<C_VoxelPool>(canvas)` at `system_update_voxel_set_children.hpp:37` is a foreign-entity lookup inside the per-entity tick. Adding `C_VoxelPool` to the system's template params is not an option here (the canvas entity has `C_VoxelPool` but not `C_VoxelSetNew`; the voxel-set entity is what's being iterated). The one-shot gate (`ownerEntityId_ == kNullEntity`) ensures this fires at most once per voxel-set entity lifetime — effectively initialization, not a hot path. The comment documents the rationale and acknowledges the TODO (`RENDERS_ON` relation-based grouping). Acceptable for a safety-fix PR.
- Removing the function-local `static unordered_map<EntityId, C_VoxelPool*> poolCache` clears the cpp-systems.md live deviation at this file. Net improvement.
- No per-entity getComponent on the iterating entity. No mid-iteration structural changes. No allocations in the hot tick path. Clean.

### API changes

- `VoxelPoolAllocation` struct (`voxel_pool_allocation.hpp`): well-formed. `startIndex_` + co-indexed spans as a single return value eliminates `std::get<N>` noise at all call sites and makes the index–span association explicit.
- `C_VoxelPool::allocateVoxels` and `deallocateVoxels` signatures are symmetric: both now operate on `(startIndex, size)`. The old `deallocateVoxels` derived `startIndex` from `positions_.data() - m_voxelPositions.data()` — the same pointer-diff hazard eliminated from the system tick.
- Bounds assertions in `setEntityIdForRange` and `deallocateVoxels` convert silent memory corruption into loud assertion failures at the allocation site. Good defensive hardening.

### Ownership / lifetime

- The spans stored in `C_VoxelSetNew` (`positions_`, `positionOffsets_`, `globalPositions_`, `voxels_`) reference into the pool's backing `std::vector` members. They remain valid for the entity's lifetime (pool is canvas-scoped, canvas persists). `onDestroy()` releases the span before the component storage is freed. Lifetime is sound.
- `getPositionGlobalsBasePtr()` is retained on the public surface per the API removal rule. Acceptable.

### Nits

- `component_voxel_pool.hpp:212` — `getPositionGlobalsBasePtr()` is now the only public surface that enables the pointer-diff pattern that just caused a SIGBUS. Adding `[[deprecated("Capture VoxelPoolAllocation::startIndex_ at allocateVoxels call time instead — see engine/render/CLAUDE.md")]]` would make the danger discoverable at compile time without violating the removal rule.

- `component_voxel_set.hpp:114` — `deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_))` frees `numVoxels_` slots. In the hypothetical allocation-mismatch path (constructor min-span guard reduces `numVoxels_` below `requestedVoxels`), the pool allocated `requestedVoxels` slots but `onDestroy()` would only release `numVoxels_`. Unreachable in practice (`allocateVoxels` always returns full-size spans and `IR_ASSERT(false)` fires before empty spans are returned), but a short comment linking `numVoxels_` to the allocation size — or storing `voxelAllocatedCount_` separately — would prevent future confusion if the guard path ever matters.

- `system_update_voxel_set_children.hpp:37` — forward-looking nit: the ECS-idiomatic direction for this lookup is a `beginTick` lambda on a `System<N>` member that rebuilds a canvas→C_VoxelPool* map each tick from fresh getComponent calls (not a `static` — rebuilt each tick, so no staleness window). The one-shot gate is correct now; this is a path hint toward the RENDERS_ON relation migration already noted in the old TODO.

### Praise

- The PR body root cause section is the right level of detail: the archetype mutation invalidity window, the Linux/macOS allocator difference, and the failure mode are all precisely described. This is useful context for anyone who hits a similar bug in adjacent code.
- Bounds assertions added to `setEntityIdForRange` and `deallocateVoxels` are a genuine safety net — if a future bug sends an out-of-bounds index, it now fails at the call site with a formatted message rather than corrupting memory downstream.
- `VoxelPoolAllocation` struct design is clean: the index and spans travel together and cannot diverge at any call site.

- [x] `fleet-build --target IRShapeDebug` clean on macOS — PR body confirms
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — 6/6 screenshots, exit 0 — PR body confirms
- [x] Visual output identical to master — PR body confirms (pure data-flow refactor)
- [ ] Linux smoke: `fleet-build --target IRShapeDebug` on Linux/OpenGL + `fleet-run IRShapeDebug --auto-screenshot 10` — PR body defers to cross-host tag
- [ ] If any creation beyond `IRShapeDebug` calls `IRRender::allocateVoxels` directly: confirm it compiles cleanly after the API change (return type change from `std::tuple` to `VoxelPoolAllocation`)

Escalation: none. The changes are pure C++ data flow — no GPU buffer layout, no shader changes, no archetype-mutation-during-tick. The fix is mechanical and bounded. Safe for merge pending Linux smoke.

Opus recheck not required.

jakildev (commented):
## Opus recheck — T-142: macOS fix for IRShapeDebug crash in UPDATE_VOXEL_SET_CHILDREN

Sonnet's first-pass already covers the root-cause analysis, the foreign-entity getComponent disposition, and the API-change call-site coverage. I focused this Opus pass on items Sonnet's checklist can't fully reach: allocator math, the one-shot-getComponent invariant, GPU buffer state, and downstream call-site coverage.

### One-shot getComponent gate — verified

Sonnet flagged the `IREntity::getComponent<C_VoxelPool>(canvas)` at `system_update_voxel_set_children.hpp` and accepted the author's claim that it is one-shot per voxel-set lifetime. On closer read I confirm: the gate body sets `voxelSet.ownerEntityId_ = entityId` BEFORE the getComponent call, and `entityId` is the iterated entity (always non-null for live archetype rows). Therefore on the next tick the gate's first conjunct (`ownerEntityId_ == kNullEntity`) is false and the lookup is skipped. The amortized cost is exactly one foreign-entity lookup per `C_VoxelSetNew` lifetime — strictly cheaper than the old static-map lookup that ran per tick.

A subtle corollary worth recording: if `setEntityIdForRange` ever asserts/aborts the process, the gate is already closed, so retry semantics are not needed. And if `numVoxels_ == 0` (allocation failed), the gate stays open forever but harmlessly — there are no voxels to claim and no foreign-entity lookup runs (the third conjunct keeps the body unreachable).

### Allocator math — span vs. numVoxels_ deallocation symmetry

The old `deallocateVoxels` derived `size = positions.size()` from the span passed in. The new variant uses `numVoxels_` from `C_VoxelSetNew`. Sonnet didn't explicitly compare these, so I traced it: `numVoxels_ = min(requestedVoxels, positions_.size())` is set at construction, and on the success path the allocator returns spans of exactly `requestedVoxels`, so `numVoxels_ == positions_.size()` always holds in practice. On the failure path the spans are empty, `numVoxels_ == 0`, and the deallocator fills nothing — also symmetric. So the change preserves the previous deallocation arithmetic with no leak risk.

### GPU buffer lifetimes — unaffected

This is a CPU-side index-tracking fix. The voxel pool's underlying SSBOs/uniform buffers (per `C_VoxelPool`) are not touched by the diff: no buffer recreation, no layout change, no descriptor rebinding. The fix doesn't perturb any GPU resource lifetime invariant.

### Allocator behavior — cleaner pre/post

The two new bounds asserts in `deallocateVoxels` and `setEntityIdForRange` (both in `component_voxel_pool.hpp`) replace the old "trust the pointer-diff" contract with explicit precondition checks. Note that the assert in `deallocateVoxels` checks against `m_voxelPoolSize` (capacity) while the one in `setEntityIdForRange` checks against `m_voxelEntities.size()` (runtime vector size). They should agree at steady state — but if a pool ever resizes its `m_voxelEntities` vector independently of `m_voxelPoolSize` they would diverge. Not a blocker for this PR (the asserts only get tighter, never looser, than the old bounds-free code), but a follow-up cleanup target if pool resizing semantics ever change.

### `VoxelPoolAllocation::startIndex_` default

Minor — the new struct does not default-initialize `startIndex_`. Every codepath in the diff explicitly assigns it before returning, so this is not a real bug, but `size_t startIndex_ = 0;` would make the struct safe under value-initialization in case some future caller does `VoxelPoolAllocation a{};` and reads the field before populating. Nit only.

### Cross-host status

PR has `fleet:authored-on-macos` (smoke-tested on Metal per the PR body) and `fleet:needs-linux-smoke` already applied. No additional smoke labels needed.

### Verdict

Approve. No new findings beyond Sonnet's; one nit on the default-init of `startIndex_`. Safe for merge pending the Linux smoke run.

Opus recheck: confirms Sonnet 4.6's first-pass review.

jakildev:
Addressed nits:
- Added `[[deprecated]]` to `getPositionGlobalsBasePtr()` pointing callers to `VoxelPoolAllocation::startIndex_` (no existing call sites, so no build warnings produced; the attribute guards future ones).
- Added comment to `C_VoxelSetNew::onDestroy()` clarifying the `numVoxels_ == requestedVoxels` invariant on non-error paths, so the `deallocateVoxels` count is self-explanatory.
- Skipped nit 3 (system_update_voxel_set_children.hpp forward-looking hint about beginTick lambda) — that's a path toward the RENDERS_ON migration, not an actionable fix for this PR.

jakildev:
Addressed nit from Opus review: added = 0 default initialization to VoxelPoolAllocation::startIndex_. No behavior change — all existing codepaths assign before returning — but safe under value-initialization for future callers. Build clean on macOS.

PR #602, opened 2026-05-10
T-141: Demo: Z-Yaw world rotation showcase

jakildev:
## Summary
- Adds `creations/demos/z_yaw_rotation/` with two executables: `IRZYawStatic` and `IRZYawInteractive`
- `IRZYawStatic` — four shapes (2 SDF, 2 voxel-pool) arranged in a ring, auto-rotating via `IRPrefab::Camera::rotateYaw()` in an endTick system. `--auto-screenshot 10` compatible.
- `IRZYawInteractive` — same scene with full lighting pipeline; `R` key toggles mouse-X-driven yaw; left click on any shape logs entity ID and triggers a 45-frame sun-direction flash (proves `getEntityIdAtMouseTrixel()` per-voxel entity-pick under rotation). Trixel hover cursor enabled.

## Blocker callout
The issue (#569) mentions hover **highlight** per entity (the whole entity glows on hover). The current engine has no per-entity color-overlay system; only the trixel-cursor ring (`IRRender::setHoveredTrixelVisible`) and global sun-direction changes are available. Per-entity highlight would require a new component + system. This demo proves the underlying pick works end-to-end; a follow-up issue should track the hover-highlight feature if desired.

Per-voxel collision detection (as opposed to SDF-level) is exercised by the voxel-pool cube and sphere: `getEntityIdAtMouseTrixel()` reads the GPU entity-id texture written during the voxel rasterize pass, which is per-voxel by construction.

## Test plan
- [x] `fleet-build --target IRZYawStatic` — clean build
- [x] `fleet-run IRZYawStatic --auto-screenshot 10` — launches, captures 2 shots, exits cleanly
- [x] `fleet-build --target IRZYawInteractive` — clean build
- [x] `fleet-run --timeout 8 IRZYawInteractive` — runs 8 s without crash
- [ ] Manual: open IRZYawInteractive, press R, move mouse — shapes rotate
- [ ] Manual: hover over voxel cube, see trixel cursor ring on individual voxels
- [ ] Manual: left-click on any shape — sun flash fires, entity ID logged

Closes #569

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #599, opened 2026-05-10
T-109: Migrate Lua perf_grid to CODEGEN, close parity gate

jakildev:
## Summary

- New `creations/demos/lua_perf_grid/` exercises the T-106..T-108
  codegen toolchain. The wave-animation component (`LuaWaveState`)
  + system body (`LuaWaveTick`) live in `main.lua`; cmake/lua_codegen
  emits a typed `IRSystem::createSystem<C_LuaWaveState>` at build
  time. `-DIR_LUA_ECS_DEFAULT_MODE=EVAL` flips the same source to
  the LuaJIT-backed runtime dispatch path.
- `cmake/lua_codegen/main.cpp` now emits `CodegenSystemIds` + helper
  unconditionally (empty when no CODEGEN systems exist) so single-
  source dual-mode demos compile in both flavors.
- `docs/design/lua-driven-ecs.md` carries the three-stage parity
  retrospective: original failure, LuaJIT floor, CODEGEN gate.

## Stack context

Stacked on: #598 (T-108 — per-system mode override + coexistence)
Full chain: T-105 → T-106 → T-107 → T-108 → **T-109**

When the upstream PRs merge, GitHub auto-rebases this PR's base.

## Parity gate measurement

60-frame averages on macOS, both demos built with the same engine
config and run identically (`fleet-run <demo> --grid-size 4 --auto-profile 60`):

| Demo                              | System         | Per-call     | Per-entity |
|-----------------------------------|----------------|--------------|------------|
| `IRPerfGrid` (C++ baseline)       | PeriodicIdle   | 0.004 ms     | ~67 ns     |
| `IRLuaPerfGrid` (CODEGEN)         | LuaWaveTick    | 0.003 ms     | ~47 ns     |

**Ratio ≈ 0.7×** — well under the 1.5× parity gate. The Lua wave
kernel is structurally simpler than `C_PeriodicIdle::tick` (no
stage table) which explains the marginal lead; this is not a "Lua
beats C++" claim, it's confirmation that codegen-emitted code
reaches C++ throughput as expected by construction.

## Pending Linux validation

Larger-grid measurements (16³ = 4096 entities, 64³ = 262 144) are
pending Linux-fleet validation. This macOS host (no full-Xcode
`.metallib` precompilation) hits a render-pipeline crash at
grid_size ≥ 8 that affects **both** the new Lua demo and
`IRPerfGrid` itself — same crash signature on the C++ baseline,
so the regression is environmental, not a Lua-codegen issue. The
[...]

PR #614, opened 2026-05-10
T-139 Phase 1: GPU particle system — pool + compute update + canvas render

### LINKED ISSUE #209: GPU particle system: compute-shader-driven dense particle field
## Area
engine/render, engine/prefabs/irreden/render, shaders

## Suggested model
[opus] — GPU compute + buffer lifetime + ECS integration. Particle counts (thousands per room) rule out per-entity CPU iteration; needs compute update and indirect-draw rendering.

## Context
irreden/docs/particles.md canon: \"Particles exist in every level... Abilities interact with particles... Rooms are tuned by particle count (particle density = hits required).\" Rainworld-style ambient field that is both aesthetic and mechanical. Engine already has voxel-rendering compute shaders — particles are the next compute system.

## Acceptance criteria
1. Particle pool in GPU buffer (SSBO); pool size configurable per biome/room
2. Compute shader update pass: position, velocity, lifetime, per-frame drift
3. Indirect draw pass: particles as world-space sprites or voxel-like quads in iso camera
4. CPU spawn API: Lua spawnParticle(pos, velocity, lifetime, type) queues for next dispatch
5. \"Collection\" query: GPU cull by radius returns particle indices for ability effects
6. Attraction points: ability-driven \"pull\" effects work
7. Benchmark: 10K active particles at 60 fps on linux-debug and macos-debug
8. Demo room with 1K drifting glow particles + test ability pulling nearby into player

## Notes
- Canon rule \"Rooms are tuned by particle count\" — expose particle count as first-class room config the game can tune.
- Nightmare corruption: per-biome particle behavior (lazy in nightmare, sharp near apex bosses). Expose drift speed, jitter, color shift per type.
- Speedrun relevance (particles.md): keep input latency low — don't batch particle-pull queries more than one frame.

jakildev:
## Summary
Phase 1 of N for issue #209 (T-139 GPU particle system). Establishes the SSBO-backed particle pool, compute-driven update pass, and canvas-write render pass — the foundation that subsequent phases build on. Both OpenGL and Metal backends are wired from day one.

## What's in this PR

**Engine surface:**
- `IRRender::GpuParticle` + `FrameDataGpuParticles` structs (32-byte std430 + 48-byte std140)
- `kBufferIndex_GpuParticleData` (binding 4, aliased) + `kBufferIndex_FrameDataGpuParticles` (binding 23, aliased) — both reuse light-volume slots since the encoder bindings don't overlap
- `C_GPUParticlePool` component (RAII SSBO + CPU staging vector, slot reservation via wrap-around scan)

**ECS prefabs:**
- `SystemName::UPDATE_GPU_PARTICLES` and `RENDER_GPU_PARTICLES_TO_TRIXEL` enum entries (+ Lua bindings)
- `system_update_gpu_particles.hpp` — uploads dirty pool, dispatches update kernel
- `system_render_gpu_particles_to_trixel.hpp` — binds canvas images, dispatches render kernel
- `IRPrefab::GpuParticles::spawn() / clearAll()` — CPU-side spawn API

**Shaders (full GLSL + Metal parity):**
- `c_update_gpu_particles.glsl` / `.metal` — `position += velocity*dt; lifetime -= dt`
- `c_render_gpu_particles_to_trixel.glsl` / `.metal` — iso-projects each live particle, atomicMin into distance texture, conditional color write (Metal uses scratch buffer slot 16 like the voxel-to-trixel pattern)

**Demo:** `creations/demos/gpu_particles/` — spawns 1024 random particles in a 48-voxel cube with random velocities, drifts them on the GPU, renders them to the trixel canvas. Cycles 3 auto-screenshot zoom levels.

## Test plan

- [x] `fleet-build --target IRGpuParticles` clean on macOS / Metal
- [x] `fleet-run IRGpuParticles --auto-screenshot 10` — 3/3 screenshots, exit 0
- [x] Visible: multi-colored particle cluster, drifting between successive shots (confirms GPU update pass)
- [x] `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10` — 6/6 shots clean (sanity: SystemName enum value shift didn't regress existing demo)
- [ ] Cross-host smoke: Linux/OpenGL build + run (via `fleet:needs-linux-smoke`)

## Visual evidence

Particles render from the demo (1024 spawned, 48-voxel spawn cube, random velocities, 8s lifetime — drift visible across shots):

| fit (0.5x) | zoom 1x | zoom 2x |
|---|---|---|
| fit | z1 | z2 |

Note: only ~30 distinct dots visible — most of the 1024 particles overlap in iso projection within the small spawn cube. Subsequent phases (collection query, attraction points) will use larger spreads.

`render-debug-loop` and `attach-screenshots` were not run because there's no master baseline for a brand-new demo. Sanity-build of `IRShapeDebug` confirmed no regression to the existing render pipeline.

## Phase plan for #209

- **Phase 1 (this PR):** SSBO pool + update compute + canvas render compute + C++ spawn API + minimal demo
- **Phase 2:** Lua spawn API + per-biome configurable pool size
- **Phase 3:** GPU collection query (cull by radius, returns particle indices)
- **Phase 4:** Attraction points (ability-driven pull effects)
- **Phase 5:** 10K particle benchmark on linux-debug + macos-debug
- **Phase 6:** Demo room w/ ability test (1K drifting + pull-on-trigger)

Closes #209 (Phase 1 only — follow-up phases will be filed as separate tasks before this PR merges).


jakildev (commented):
## Review — T-139 Phase 1: GPU particle system — pool + compute update + canvas render

**Verdict:** approve

### Praise

- `GpuParticle` struct layout is elegant: `float lifetime_` fills the trailing 4 bytes of each `vec3`'s trailing pad in std430, landing at offsets 0/12/16/28 without any explicit padding and achieving exactly 32 bytes. The `static_assert(sizeof(GpuParticle) == 32)` and `static_assert(sizeof(FrameDataGpuParticles) == 48)` guard against accidental rearrangement — good discipline.
- GLSL and Metal shaders both present from day one. The Metal path correctly uses `packed_float3` (12-byte, 4-byte aligned) instead of `float3` (16-byte aligned), which is necessary to match the C++ `glm::vec3` layout. The scratch-buffer atomic workaround at slot 16 mirrors the established voxel-to-trixel Metal pattern.
- The aliased binding slots (4 and 23) are thoroughly documented in `ir_render_types.hpp` with the encoder-ordering constraint spelled out. The alias rationale (Metal caps at 30 bindings) is explained once and referenced from the new slots — no drift risk.
- `C_GPUParticlePool::reserveSlot()` uses a wraparound scan from `nextSearchIndex_` so N consecutive spawns into an empty pool run in O(N) total (not O(N^2)). That's the right amortization choice for a slot-reclamation loop.

### Nits

- `engine/prefabs/irreden/render/systems/system_render_gpu_particles_to_trixel.hpp` doc comment (lines 22-35): The phrase "after the voxel and shape passes have established their depth taps" explains the distance-texture dependency, but on Metal there is a second implicit requirement: `distanceScratch` at slot 16 must be pre-bound by a prior compute dispatch in the same encoder. The voxel or shapes stage provides this; an encoder that only dispatches particles would pass a null/stale buffer on Metal. Suggest adding: "On Metal, this system additionally requires `distanceScratch` (slot 16) to be pre-bound by a prior voxel or shapes dispatch in the same compute encoder; a pure-particle pipeline without those stages is unsupported in Phase 1."

- [x] `fleet-build --target IRGpuParticles` clean on macOS / Metal (done per PR body)
- [x] `fleet-run IRGpuParticles --auto-screenshot 10` — 3/3 shots, exit 0 (done)
- [x] `IRShapeDebug` sanity build + run — 6/6 shots clean (done; confirms SystemName enum shift does not regress existing demo)
- [ ] Cross-host smoke: Linux / OpenGL build + run — tracked by `fleet:needs-linux-smoke`

Opus recheck required: engine/render/, engine/prefabs/irreden/render/ — GPU buffer lifetime (C_GPUParticlePool owns a raw Buffer* alongside a ResourceId; verify no dangle path under misuse), GPU fence correctness between the update SSBO write and the render SSBO read across OpenGL and Metal, and the aliased-slot ordering constraint (slots 4 and 23 shared with COMPUTE_LIGHT_VOLUME — verify the engine pipeline order enforces non-overlap for creations that register both).

Escalation: please Opus-recheck before merge (touches: engine/render/, engine/prefabs/irreden/render/, GPU buffer lifetime, aliased binding slots).

jakildev:
Addressed nit: added Metal distanceScratch (slot 16) pre-binding requirement note to RENDER_GPU_PARTICLES_TO_TRIXEL doc comment. fleet:approved stays valid — doc cleanup only. — sonnet-author

jakildev (commented):
## Opus recheck — T-139 Phase 1: GPU particle system

**Verdict:** approve

Recheck of the three items Sonnet flagged for Opus review (GPU buffer lifetime, fence correctness, aliased binding slots). All three check out; one of Sonnet's earlier nits is misdirected and should not be applied.

### Confirming Sonnet's flagged items

**1. `C_GPUParticlePool` raw `Buffer *` lifetime.** Safe. `RenderingResourceManager` stores resources as `unordered_map<ResourceId, unique_ptr<T>>` (`rendering_rm.hpp:17`), so the raw `Buffer *` cached in `buffer_.second` is pointee-stable across resource-map modifications — only `destroyResource<Buffer>(buffer_.first)` invalidates it, and that call is owned by the component itself (`onDestroy()` in `component_gpu_particle_pool.hpp:45-47`). This is the same RAII shape as `C_TriangleCanvasTextures`, `C_TrixelFramebuffer`, and `C_CanvasFogOfWar` — a documented exception to component method tier (c) per `engine/prefabs/CLAUDE.md`. No new dangle path.

**2. Update→render SSBO fence.** Correct on both backends.
- OpenGL: `system_update_gpu_particles.hpp:95` issues `memoryBarrier(BarrierType::SHADER_STORAGE)` after the update dispatch, which serializes the SSBO write against the subsequent render-pass SSBO read.
- Metal: `metal_render_impl.cpp:340-358` creates a fresh `ComputeCommandEncoder` per `dispatchCompute` call and calls `endEncoding()` after the dispatch; `memoryBarrier()` itself is a documented no-op at `metal_render_impl.cpp:360-364` precisely because the encoder boundary is the implicit barrier between compute dispatches that touch shared resources. The model is the same one the voxel-to-trixel passes rely on.

**3. Aliased binding slots 4 / 23 vs `COMPUTE_LIGHT_VOLUME`.** Safe.
- The slot-alias comment at `ir_render_types.hpp:358-369` correctly captures the constraint, and both particle systems explicitly `bindBase(slot)` for the particle SSBO and UBO immediately before each dispatch (`system_update_gpu_particles.hpp:85-90`, `system_render_gpu_particles_to_trixel.hpp:75-80`). On OpenGL the rebind resets the binding regardless of what `COMPUTE_LIGHT_VOLUME` left there. On Metal each dispatch creates its own encoder with its own argument table (`metal_render_impl.cpp:73-98`), so the alias is invisible to the GPU.
- Cross-checked the demo's pipeline order at `creations/demos/gpu_particles/main.cpp:94-103` — it does not register `COMPUTE_LIGHT_VOLUME`, so no aliasing exposure in the demo itself. Mixed-stack creations are still covered by the established `COMPUTE_LIGHT_VOLUME → LIGHTING_TO_TRIXEL → RENDER_GPU_PARTICLES_TO_TRIXEL` ordering called out in the comment.

### Correction to one of Sonnet's earlier nits — do NOT apply

Sonnet's Nits section asked for a docstring on `system_render_gpu_particles_to_trixel.hpp` stating that the render pass requires `distanceScratch` (slot 16) to be pre-bound by a prior voxel/shapes dispatch on Metal, and that "a pure-particle pipeline without those stages is unsupported in Phase 1." That is not correct, and the proposed docstring would mislead future engineers.

What the Metal backend actually does:
- `bindComputeResources` at `metal_render_impl.cpp:93-97` unconditionally binds whatever buffer `currentImageAtomicScratch()` returns to slot 16 on **every** compute encoder.
- The current scratch is set by `bindImage()` on any R32I texture at `metal_texture.cpp:131-136` — so the render system's own `canvas.getTextureDistances()->bindAsImage(1, ..., R32I)` call at `system_render_gpu_particles_to_trixel.hpp:85-87` is what publishes the scratch for this dispatch.
- The scratch contents are mirrored on every R32I clear (`metal_render_impl.cpp:594-601`) so the atomic-min sees the same starting state as the texture.

Net: a pipeline with just the particle render pass (no voxel or shapes stage) would still bind the canvas distance scratch correctly on Metal. Please leave the docstring as-is rather than absorbing Sonnet's nit verbatim.

### Phase-1-OK observations (no action required)

- **CPU-side lifetime never decays.** `reserveSlot()` scans `pool.particles_[idx].lifetime_ <= 0.0f` to find dead slots, but the CPU mirror never observes the GPU's per-frame `lifetime_ -= deltaTime` from `c_update_gpu_particles.glsl:43`. Once the pool fills, every CPU slot stays "alive" forever from the CPU's perspective, so subsequent spawns drop silently after 4096 cumulative spawns even though the GPU has long since killed those slots. The Phase 1 demo only seeds once (`seedParticles()` at startup, no respawn), so the bug doesn't manifest. Phase 2's Lua spawn API will hit it the first time a continuous-emitter use case lands; the fix is either a parallel CPU-side `lifetime_ -= dt` per slot each tick, a GPU→CPU readback, or moving to FIFO slot allocation. Worth a follow-up issue before Phase 2 starts.

- **`frameData_.particleCount_` is really "pool capacity", not "live count".** Both systems set it to `pool.capacity_`, and both shaders use it solely as the loop-bound guard `if (idx >= particleCount) return;`. Dead-slot culling happens later via `if (p.lifetime <= 0.0) return;`. With 4096 capacity the wasted invocations are negligible, but the field name is misleading — `poolCapacity_` would read more accurately, or alternatively populate this with a true live count once a GPU compaction pass exists.

- **Migration nit.** Both new systems use the explicit `Params` + `setSystemParams` pattern. Per `engine/system/CLAUDE.md` "Per-system parameters", new systems should default to the `registerSystem` member-on-`System<N>` form. The explicit form is still supported (it matches the existing render-system style), so consistency wins for this PR; flagging only so it's on the radar when these systems get touched again.

### Approval scope

Engine-side correctness for Phase 1 is clean: structs, layouts, fences, bindings, RAII. Phase 1's narrow scope (single seed at startup, no respawn) keeps the CPU-side lifetime gap latent. Linux smoke still pending per `fleet:needs-linux-smoke`. Approving on the engine review; the smoke gates are independent.

jakildev:
Cross-host smoke OK on macOS (fresh checkout build + IRShapeDebug --auto-screenshot 10, 6/6 shots, clean exit). — opus-worker-2

jakildev:
judgement-level resolution.

Conflicted files:
- `creations/CMakeLists.txt` — master: `9e8c091 T-141: Demo: Z-Yaw world rotation showcase (#602)`; PR: `c5c5e97 T-139 Phase 1: GPU particle system foundation`

The conflict is two `add_subdirectory(...)` lines added at the same insertion point — T-141 added `demos/z_yaw_rotation` on master, this PR adds `demos/gpu_particles`. Mechanically the resolution is "keep both", but the merger only auto-resolves TASKS.md sort-merges and whitespace-only conflicts; any other in-rebase edit is treated as a semantic decision.

SHA pair: master=73ccbdd7ba3503d06b1dc1fd8b943839d38cadaa × PR=7d5107d4d21696d438889150cd1c4290ed2284a6

— fleet merger

jakildev:
Resolved semantic conflict in creations/CMakeLists.txt: kept both add_subdirectory lines (T-141 z_yaw_rotation from master + gpu_particles from this PR). Rebase clean, IRShapeDebug and IRGpuParticles both build on macOS. Reviewer please re-evaluate the rebased diff. — opus-worker-2

jakildev (commented):
## Re-review — T-139 Phase 1: GPU particle system — pool + compute update + canvas render

_Re-reviewing because `fleet:changes-made` was set by opus-worker-2 after resolving a semantic conflict in `creations/CMakeLists.txt` (rebase onto master that added `demos/z_yaw_rotation`). Evaluating prior-review items and the rebased diff._

**Verdict:** needs-fix

---

### Prior-review resolution

**Sonnet first-pass (2026-05-10T18:54:22Z):**

- ✅ All Blockers/Needs-fix: none were raised — Sonnet verdict was approve.
- ❌ Nit — `engine/prefabs/irreden/render/systems/system_render_gpu_particles_to_trixel.hpp:38-41` — doc comment stating "On Metal, this system additionally requires `distanceScratch` (slot 16) to be pre-bound by a prior voxel or shapes dispatch in the same compute encoder; a pure-particle pipeline without those stages is unsupported in Phase 1."

  Applied at commit c8b1bbb (`engine/prefabs/irreden/render: document Metal distanceScratch pre-binding requirement`). **Opus's recheck review (19:00:12Z) explicitly repudiated this nit as factually incorrect.** Opus's explanation:
  - `bindComputeResources` at `metal_render_impl.cpp:93-97` unconditionally binds `currentImageAtomicScratch()` to slot 16 on *every* compute encoder.
  - The render system's own `canvas.getTextureDistances()->bindAsImage(1, ..., R32I)` call (at `system_render_gpu_particles_to_trixel.hpp:85-87`) is what sets the current scratch — not a prior voxel/shapes dispatch.
  - Net: a pure-particle pipeline (no voxel or shapes stage) would still bind the canvas distance scratch correctly on Metal.

  The doc comment is factually wrong and will mislead future engineers who read it. **Re-flagged below as needs-fix.**

**Opus recheck (2026-05-10T19:00:12Z):**

- ✅ GPU buffer lifetime (`C_GPUParticlePool` raw `Buffer *`): confirmed safe — pointee-stable via `unordered_map<ResourceId, unique_ptr<T>>`, destroyed via `onDestroy()`. No action needed.
- ✅ Update→render SSBO fence: correct on both backends (OpenGL `memoryBarrier(SHADER_STORAGE)`, Metal encoder-boundary implicit barrier). No action needed.
- ✅ Aliased binding slots 4/23: safe per-dispatch rebind on OpenGL; per-encoder argument table isolation on Metal. No action needed.
- ✅ Phase-1-OK observations (CPU lifetime decay, particleCount_ naming, Params migration note): acknowledged for follow-up, not blocking Phase 1. No action needed.

---

### Needs-fix

- `engine/prefabs/irreden/render/systems/system_render_gpu_particles_to_trixel.hpp:38-41` — **Incorrect doc comment.** The five lines added by commit c8b1bbb state that `distanceScratch` (slot 16) requires a prior voxel/shapes dispatch on Metal. Opus confirmed this is wrong: `bindComputeResources` binds slot 16 unconditionally on every encoder, and the render system's own R32I `bindAsImage` call sets the current scratch. The recommended fix is to **revert commit c8b1bbb** (drop lines 38-41), leaving the doc comment at its pre-nit state. If a Metal-specific note is desired at all, it should accurately reflect the actual binding mechanism: "On Metal, `distanceScratch` (slot 16) is bound automatically by `bindComputeResources` on every compute encoder; no prior dispatch is required."

---

### Rebase check

`creations/CMakeLists.txt` diff: `demos/z_yaw_rotation` (from master/T-141) is already present on line 12; `demos/gpu_particles` is added on line 13 immediately after. Both `add_subdirectory` lines are present. Conflict resolution is correct — no lines dropped, no ordering issues.

No other files changed by the rebase. The entire engine/render and shader surface is byte-identical to the Opus-approved state.

---

### Test plan

- [x] Rebase clean — opus-worker-2 confirms IRShapeDebug and IRGpuParticles both build on macOS post-rebase.
- [x] `creations/CMakeLists.txt` conflict resolved correctly — verified in diff.
- [ ] After reverting c8b1bbb: `fleet-build --target IRGpuParticles` still clean — doc-only change, but confirm no accidental edit to surrounding code.
- [ ] Cross-host smoke: Linux/OpenGL build + run — still pending (tracked by `fleet:needs-linux-smoke`).

Escalation: author-agent to revert commit c8b1bbb (or correct the doc comment per Opus's analysis), then re-request review.

jakildev:
Addressed feedback: reverted commit c8b1bbb (incorrect Metal distanceScratch doc comment) per Opus recheck. File is byte-exactly back to its pre-c8b1bbb state. Smoke on macOS: `fleet-build --target IRGpuParticles` clean; `fleet-run IRGpuParticles --auto-screenshot 10` → 3/3 shots, exit 0. No accidental edit to surrounding code (the only diff is the 5 added comment lines being removed). Ready for re-review. — opus-worker-1

jakildev:
Cleanup: removed stale `fleet:needs-fix` label. Revert of c8b1bbb landed at `da4746f` (see prior comment at 03:53:15Z) but the label swap was missed in that iteration. `fleet:changes-made` carries over from the earlier conflict-resolution flow and remains as the re-review signal. — opus-worker-1

jakildev:
Label cleanup: removed stale fleet:needs-fix. Commit da4746f already reverts c8b1bbb per opus-worker-1 (2026-05-11T03:53:15Z comment) — the incorrect Metal distanceScratch doc comment is gone, file is byte-exact to its pre-nit state. macOS smoke already confirmed by opus-worker-1 (IRGpuParticles --auto-screenshot 10, 3/3 shots, exit 0). fleet:changes-made is set; reviewer can re-verify. fleet:needs-linux-smoke still pending a Linux host run. — sonnet-fleet-2

jakildev (commented):
## Opus recheck — T-139 Phase 1: GPU particle system (post-revert)

**Verdict:** approve

Re-evaluating after `fleet:changes-made` flagged a fix push. The author landed commit `da4746f` reverting the misdirected doc comment from `c8b1bbb`, per Sonnet's 2026-05-11T03:48Z needs-fix instruction.

### Confirming the revert

`git diff 3c7777d7..da4746f8 --stat` returns **empty**. The PR HEAD is byte-identical to the originally Opus-approved tip (`3c7777d7`, approved 2026-05-10T19:00Z). The add at `c8b1bbb` and the revert at `da4746f` cancel exactly — five lines added, five lines removed, same file, same blob hash on either side of the diff window.

Sonnet's flagged items from the 03:48 re-review:

- ✅ `engine/prefabs/irreden/render/systems/system_render_gpu_particles_to_trixel.hpp:38-41` — incorrect Metal `distanceScratch` doc comment is gone. Lines 22-35 are back to the original wording. The revert commit message correctly cites the prior Opus rationale (`bindComputeResources` binds slot 16 unconditionally on every Metal compute encoder; the render system's own `bindAsImage(R32I)` publishes the scratch for the dispatch).
- ✅ `creations/CMakeLists.txt` rebase resolution — both `demos/z_yaw_rotation` (master / T-141) and `demos/gpu_particles` are present and adjacent on lines 11-12. Confirmed visually.

No other engine surface drift. Everything I approved on the first pass — the `GpuParticle` 32-byte std430 layout, `FrameDataGpuParticles` 48-byte std140 layout, the aliased binding slots 4 and 23 with their non-overlap encoder ordering, GLSL/Metal parity, and the `C_GPUParticlePool` raw-`Buffer*` lifetime through `RenderingResourceManager`'s `unique_ptr` storage — is byte-identical to that approved state.

### Carry-over from the prior Opus review

The follow-up issues I flagged on the first pass remain valid but are not blockers for Phase 1 (single seed at startup, no respawn keeps both latent):

- CPU mirror's per-slot `lifetime_` is never decremented; a Phase 2 continuous-emitter Lua API will trip the silent-drop-after-4096-spawns bug. Worth a follow-up issue before Phase 2 starts.
- `frameData_.particleCount_` is really pool capacity, not live count. Misleading name; benign given the shader's `idx >= particleCount` guard.
- New systems use the explicit `Params` + `setSystemParams` form instead of the `registerSystem` member-on-`System<N>` form — flagged for radar, consistency wins for this PR.

None of these are introduced or worsened by the revert; they are unchanged from the Opus-approved state.

### Test plan

- [x] Revert is clean — `git diff 3c7777d7..da4746f8 --stat` empty (verified).
- [x] `creations/CMakeLists.txt` rebase resolution preserved (verified).
- [x] Engine-side correctness — confirmed by prior Opus recheck on the byte-identical state.
- [ ] Cross-host smoke: Linux / OpenGL build + run — still pending (tracked by `fleet:needs-linux-smoke`).

Approving on the engine review; the Linux smoke gate is independent of this verdict.

engine/prefabs/irreden/render/components/component_gpu_particle_pool.hpp
```diff
@@ -0,0 +1,94 @@
...
+    /// alive the request is silently dropped (returns `kInvalidSlot`).
+    /// Linear-scan reclamation: the search wraps around from
+    /// `nextSearchIndex_` so consecutive spawns don't all O(N²) re-scan
+    /// from zero.
+    std::uint32_t reserveSlot() {
+        for (std::uint32_t step = 0; step < capacity_; ++step) {
+            const std::uint32_t idx = (nextSearchIndex_ + step) % capacity_;
+            if (particles_[idx].lifetime_ <= 0.0f) {
+                nextSearchIndex_ = (idx + 1) % capacity_;
+                dirty_ = true;
```
jakildev: Remember that I do not like using dirty flags. The particles are always being updated, and we shouldnt need to use this. Please also cofify this preference in claude files and rules.

jakildev (commented):
Please remove dirty flag and cofigy rule

jakildev:
Addressed feedback (commit `5219f42`):

- **Removed `C_GPUParticlePool::dirty_`.** `writeSlot` now uploads the single slot directly via `subData` (offset = `sizeof(GpuParticle) * index`, size = `sizeof(GpuParticle)`); `clear` uploads the full cleared buffer in one `subData`; the ctor seeds the SSBO with the all-zero CPU state so the kernel's `lifetime <= 0` early-out is well-defined on frame 0. `UPDATE_GPU_PARTICLES` no longer carries a CPU→GPU sync step — the GPU owns ongoing position/lifetime, the CPU pushes only at mutation time.
- **Codified the rule.** New section in `.claude/rules/cpp-ecs.md` "No dirty flags on components" with the two replacement patterns (push-at-mutation-time, GPU-owns-ongoing-state) and the rationale (hides the ownership contract, defers the correct fix, encourages sync drift). Anti-pattern entry added in `engine/prefabs/CLAUDE.md` linking back to the rule.
- **Documented exception:** `C_CanvasFogOfWar` keeps its `dirty_`/`allUnexplored_` because the texture is CPU-authored, GPU-read-only, and a per-cell `subImage2D` would split `revealRadius` into hundreds of API calls. Component header now flags itself as the documented exception and points at the rule; the rule's "Live deviations" block lists it.

Smoke on macOS: `fleet-build --target IRGpuParticles` clean; `fleet-run IRGpuParticles --auto-screenshot 10` → 3/3 shots, exit 0. Visual output unchanged (particles render correctly at all three zoom levels). `fleet:needs-linux-smoke` still pending. — opus-worker-1

jakildev (commented):
## Re-review — T-139 Phase 1: GPU particle system — pool + compute update + canvas render

_Re-reviewing because `fleet:changes-made` was set after the human's inline comment at 04:06:25Z requesting removal of the `C_GPUParticlePool::dirty_` flag and codification of the preference in Claude rules. Author (opus-worker-1) addressed this in commit `5219f42` at 04:47:51Z._

**Verdict:** approve

---

### Prior-review resolution

**Opus recheck (2026-05-11T03:57:35Z) — last approved state:**

- ✅ Revert of `c8b1bbb` (incorrect Metal `distanceScratch` doc comment) confirmed clean at `da4746f` — byte-identical to originally Opus-approved tip.
- ✅ `creations/CMakeLists.txt` rebase resolution (both `demos/z_yaw_rotation` and `demos/gpu_particles`) — preserved.
- ✅ All prior Opus-approved surface (GpuParticle layout, FrameDataGpuParticles layout, aliased binding slots 4/23, GLSL/Metal parity, C_GPUParticlePool raw-Buffer* lifetime) — unchanged in `5219f42`.

**Human comment (04:06:25Z — two requests):**

- ✅ Remove `C_GPUParticlePool::dirty_` — done in `5219f42`. Field removed from struct declaration; `dirty_ = true` calls removed from `reserveSlot()` and `writeSlot()`. `writeSlot` now issues a one-slot `Buffer::subData` directly; `clear()` uploads the full zeroed CPU buffer in one `subData` call. Constructor seeds the SSBO with all-zero CPU state via `subData` so the update kernel's `lifetime <= 0` early-out is well-defined on frame 0.
- ✅ Codify the preference — done. New section "No dirty flags on components" in `.claude/rules/cpp-ecs.md` with the rule, two honest patterns, rationale, and a "Live deviations" block. Anti-pattern entry added to `engine/prefabs/CLAUDE.md` with pointer to the rule file. `C_CanvasFogOfWar` header updated to note it is the documented exception (CPU-authored, GPU-read-only, full-texture upload) and point at the rule.

---

### New commit analysis (`5219f42`)

**GPU sync model after dirty-flag removal.** The previous design gated a full SSBO re-upload on `dirty_` once per frame in `UPDATE_GPU_PARTICLES`. The new design:

1. `writeSlot` — CPU issues `subData(offset, sizeof(GpuParticle), &particle)` at spawn time during the UPDATE phase.
2. `clear()` — CPU issues `subData(0, fullSize, particles_.data())` at clear time during UPDATE.
3. `UPDATE_GPU_PARTICLES` system tick — dispatches the compute kernel, issues `memoryBarrier(SHADER_STORAGE)`.
4. `RENDER_GPU_PARTICLES_TO_TRIXEL` system tick — reads the SSBO (now fully updated by GPU decay + any CPU mutations this frame).

On OpenGL: CPU `subData` is coherent with subsequent GPU commands without an explicit barrier — the driver handles the ordering. On Metal: CPU writes to a dynamic buffer are visible to the GPU when the command buffer is committed; the encoder-boundary pattern that Opus confirmed at 19:00Z applies unchanged. The `SHADER_STORAGE` barrier after step 3 still serializes the update compute write against the render compute read. No new fence concern introduced.

**Rule documentation quality.** `.claude/rules/cpp-ecs.md` "No dirty flags" section is well-structured: concrete rule, two canonical patterns, rationale for each failure mode, and a live-deviations block that future reviewers can scan. The `C_GPUParticlePool::writeSlot` example in the rule is accurate to the code as shipped. ✅

**`component_canvas_fog_of_war.hpp`.** Doc comment correctly documents the exception. The incidental single-line `if` reformats (`if (cond) return;` → `if (cond)\n    return;`) are drive-by style normalizations — harmless but out of scope for this commit's stated purpose. Not blocking; noting for hygiene.

**Carry-over from prior Opus reviews (unchanged):**
- CPU `particles_[].lifetime_` does not track the GPU's per-frame decay; dead-slot scan stays conservative (slots appear "taken" until `clear()`). Phase 1 demo seeds once at startup — latent until Phase 2 continuous-emitter Lua API.
- `frameData_.particleCount_` is really pool capacity; benign given the shader's `idx >= particleCount` guard.
- New systems use the explicit `Params` + `setSystemParams` form — flagged for radar, consistency wins for this PR.

---

### Test plan

- [x] `fleet-build --target IRGpuParticles` clean on macOS — opus-worker-1 confirms (04:47:51Z comment).
- [x] `fleet-run IRGpuParticles --auto-screenshot 10` → 3/3 shots, exit 0 — opus-worker-1 confirms visual output unchanged.
- [ ] Cross-host smoke: Linux/OpenGL build + run — still pending (`fleet:needs-linux-smoke`).

---

Escalation: please Opus-recheck before merge (touches engine/prefabs/irreden/render/ — dirty-flag removal changes the CPU→GPU sync model for C_GPUParticlePool; the SHADER_STORAGE barrier is unchanged and CPU subData coherence is standard, but Opus should confirm the per-write subData timing across both backends and the frame-boundary ordering before the label is set).

jakildev (commented):
## Opus recheck — T-139 Phase 1: GPU particle system (dirty-flag removal)

**Verdict:** approve

Re-evaluating commit `5219f42` (dirty-flag removal + rule codification) after Sonnet's 2026-05-11T04:55:31Z escalation. Sonnet asked me to confirm "the per-write `subData` timing across both backends and the frame-boundary ordering before the label is set". All three items check out; flagging one Nit on the new rule's perf rationale.

### Confirming the new CPU→GPU sync model

**1. OpenGL — per-write `subData` is coherent.** `opengl_buffer.cpp:41-45` issues `glNamedBufferSubData` directly. GL guarantees that subsequent commands referencing the same buffer see the new bytes (the driver synchronises internally via its standard "buffer renaming" / sub-allocation path). The `UPDATE_GPU_PARTICLES` system's `memoryBarrier(BarrierType::SHADER_STORAGE)` at `system_update_gpu_particles.hpp:79` still serialises the compute write against the render-pass compute read, so the update→render chain is unchanged from the prior approved state.

**2. Metal — per-write `subData` is functionally correct.** `metal_buffer.cpp:48-83` implements `subData` as orphan-on-write: it allocates a fresh `MTL::Buffer` (Shared storage), copies the live buffer's contents into it, patches the new bytes at `offset`, then calls `replaceMetalBufferInBindings(old, new)` and `deferReleaseMetalBuffer(old)`. The deferred-release queue is drained at the frame-boundary `present()` path (`metal_render_impl.cpp:251-266`), which does `commandBuffer->commit() → waitUntilCompleted() → releaseDeferredMetalBuffers()`. That `waitUntilCompleted()` guarantees:

- Any encoder that captured the orphaned buffer has finished executing before the buffer is released — no use-after-free.
- The next frame's `m_buffer->contents()` reads observe the GPU's previous-frame writes coherently (Shared storage + completion barrier), so the orphan's `memcpy` source is in a defined state.

Within a frame, all `writeSlot` calls happen during INPUT (or pre-game-loop seeding) before any RENDER-phase compute dispatches are recorded. The `replaceMetalBufferInBindings` at orphan time only affects future bindings, not commands already recorded on an open encoder — so even if mid-pipeline spawning is added later, the worst case is "this frame's update sees the pre-spawn state; render sees the post-spawn state", which is a missed-frame, not a corruption. Phase 1's pipeline order is safe.

**3. Frame-boundary ordering.** No change from the prior Opus-approved state. The update dispatch issues `SHADER_STORAGE` barrier on OpenGL (no-op on Metal, where the per-dispatch encoder boundary is the implicit barrier). The render dispatch issues `SHADER_IMAGE_ACCESS` barrier on OpenGL. Frame-N writes are visible to frame-N+1 via Metal's `waitUntilCompleted()` and GL's command-order guarantees. ✓

**4. Constructor SSBO seed.** `component_gpu_particle_pool.hpp:46-53` seeds the SSBO with the all-zero CPU mirror immediately after `createResource(nullptr, ...)`. This is necessary because `MTL::Device::newBuffer(size, Shared)` returns a buffer with undefined contents (and GL's `glBufferStorage(nullptr)` is the same), and the update kernel's `lifetime <= 0` early-out would otherwise read garbage on frame 0. Author got this right. ✓

### Confirming the rule codification

`.claude/rules/cpp-ecs.md` "No dirty flags on components": the rule statement, the two replacement patterns, and the rationale (ownership-contract clarity, sync-drift) are accurate. The "Live deviations" block correctly carves out `C_CanvasFogOfWar` (CPU-authored, GPU-read-only, full-texture upload). `engine/prefabs/CLAUDE.md` cross-links the rule from the anti-patterns list. `component_canvas_fog_of_war.hpp` header now flags its exception status. The codification round-trip is complete.

### Sonnet's "drive-by reformat" framing in `component_canvas_fog_of_war.hpp` is incorrect — leave the reformat in

Sonnet flagged the single-line-`if` reformats (`if (cond) return;` → two-line form) as "drive-by style normalisations, out of scope for this commit's purpose, not blocking, noting for hygiene". On closer read, the project `.clang-format` sets `AllowShortIfStatementsOnASingleLine: Never`, so those one-liners were pre-existing clang-format violations and the reformat is project-standard cleanup that should have been there. Leaving them in is the right call.

### Nits

**N1 — Rule's "Per-write subData is O(bytes changed)" claim is OpenGL-flavoured; Metal pays O(buffer size) per call.** `cpp-ecs.md` says "Per-write subData is O(bytes changed); a dirty-gated full re-upload is O(buffer size) per dirty frame." That's accurate for OpenGL (`glNamedBufferSubData` is a straight patch). On Metal it's inverted: every `subData` orphans the entire buffer (alloc + full memcpy + bookkeeping + deferred release), so the cost is O(buffer size) **per call** regardless of patch size. For Phase 1's 1024 init spawns at 32 B each that's ~1024 × 128 KB ≈ 128 MB of allocation+copy churn at startup; a continuous-emitter Phase 2 spawning 100/frame would be ~12.8 MB/frame of pure orphan overhead on Metal.

The rule's substantive correctness argument (dirty-flag clobbers GPU writes when the GPU also mutates the buffer; sync-drift over time; hidden ownership contract) is sound regardless of backend — the recommendation for `C_GPUParticlePool` is correct because the GPU mutates the SSBO every frame, not because per-write subData is universally cheaper. Suggest adding a one-line caveat to the rule, e.g.:

> Cost note: the OpenGL backend implements `subData` as a true partial patch (O(bytes changed)). The Metal backend orphans the whole buffer per `subData` call (O(buffer size)) for synchronisation reasons; for high-rate per-frame mutators the orphan churn is real. Phase 2's continuous-emitter API should batch spawns into a single per-frame `subData` rather than per-particle.

Optional polish — the rule's primary guidance is right; this is just so the perf justification doesn't push future Metal-aware reviewers in the wrong direction.

**N2 — Worth a follow-up issue for Phase 2's continuous-emitter Metal cost.** When the Lua spawn API lands and emitters can issue many spawns per frame, the orphan-on-write cost will become measurable on Metal. The fix is either (a) batch CPU-side spawns into a frame-scoped staging vector and issue one `subData` per frame, or (b) extend the Metal `BufferImpl` with a "frame-coherent batch" mode that defers the orphan to the next bind. Filing this so it's on the radar before Phase 2 lands; this PR is correct as-is.

### Carry-over from prior Opus reviews (unchanged by `5219f42`)

- CPU mirror's per-slot `lifetime_` is never decremented; a Phase 2 continuous-emitter Lua API will trip the silent-drop-after-4096-spawns bug. Worth a follow-up issue before Phase 2 starts.
- `frameData_.particleCount_` is really pool capacity, not live count. Misleading name; benign given the shader's `idx >= particleCount` guard.
- New systems use the explicit `Params` + `setSystemParams` form instead of `registerSystem` member-on-`System<N>`. Consistency wins for this PR; flagging only for radar.

### Test plan

- [x] Per-write `subData` correctness on OpenGL and Metal — confirmed by code-read of `opengl_buffer.cpp` and `metal_buffer.cpp` + `metal_render_impl.cpp::present()`.
- [x] Constructor seed coverage of undefined-content fresh buffer — confirmed.
- [x] Frame-boundary ordering — `waitUntilCompleted()` on Metal, `SHADER_STORAGE` barrier on GL, both unchanged.
- [x] Rule codification accurate to shipped code — confirmed.
- [x] `fleet-build --target IRGpuParticles` clean on macOS — opus-worker-1 confirms (04:47:51Z).
- [x] `fleet-run IRGpuParticles --auto-screenshot 10` → 3/3 shots, exit 0 — opus-worker-1 confirms.
- [ ] Cross-host smoke: Linux/OpenGL build + run — still pending (`fleet:needs-linux-smoke`).

Approving on the engine review. The Linux smoke gate is independent of this verdict.

jakildev:
Addressed nits from the 2026-05-11T05:38 Opus recheck:

- **N1** — qualified the perf claim in `.claude/rules/cpp-ecs.md` "Defers the correct fix" bullet: per-write `subData` is `O(bytes changed)` on OpenGL but `O(buffer size)` per call on Metal (orphan-on-write). Inlined as a single sentence in the existing bullet rather than a separate paragraph to avoid restating the OpenGL claim. The substantive "no dirty flag" rule itself is unchanged — only the perf justification gains the Metal qualifier.
- **N2** — filed #632 for the Phase 2 continuous-emitter Metal orphan-churn batching follow-up.

Doc-only change; engine code surface unchanged. `fleet:approved` stays valid per the has-nits convention; `fleet:needs-linux-smoke` still pending. — opus-worker-2

jakildev:
judgement-level resolution.

Conflicted files:
- `creations/CMakeLists.txt` — master: `08fd799 T-109: Migrate Lua perf_grid to CODEGEN, close parity gate (#599)`; PR: `3c7777d T-139 Phase 1: GPU particle system foundation`

SHA pair: master=746914b4a579d5ed9a30ec562c1af41d9c280ae6 × PR=fb22bf8dc8ee6428f5cd1f8e845295cc03c49bda

— fleet merger

jakildev:
Resolved semantic conflict in `creations/CMakeLists.txt`: kept both `add_subdirectory` lines (T-109 `demos/lua_perf_grid` from master and `demos/gpu_particles` from this PR), preserving insertion order. Rebase clean (6 commits replayed; `engine/system/include/irreden/system/ir_system_types.hpp` auto-merged on SystemName enum additions). Build verified: `fleet-build --target IRShapeDebug` clean and `fleet-build --target IRGpuParticles` clean on macOS. Reviewer please re-evaluate the rebased diff. — opus-worker-1

jakildev (commented):
## Re-review — T-139 Phase 1: GPU particle system — pool + compute update + canvas render

_Re-reviewing because `fleet:changes-made` was set after a new commit "docs: clarify subData cost on Metal in no-dirty-flags rule" was pushed at 2026-05-12T02:49:35Z. This commit addresses Opus's N1 nit from the 2026-05-11T05:38:06Z recheck. All other content is unchanged from the last Opus-approved state (`5219f42`)._

**Verdict:** approve

---

### Scope of new commit

The sole change is to `.claude/rules/cpp-ecs.md` — the "No dirty flags on components" rule's cost-rationale bullet now reads:

> Per-write `subData` is O(bytes changed) on OpenGL (true partial patch); on Metal it orphans the whole buffer per call (O(buffer size) regardless of patch size, for synchronization). A dirty-gated full re-upload is O(buffer size) per dirty frame on both backends. The "optimization" is usually a pessimization unless mutations are dense across the buffer in a single frame — at which point the right shape is a sparse dirty-range tracker (start/end indices), not a single boolean. On Metal specifically, high-rate per-frame mutators should also batch CPU-side writes into a single per-frame `subData` to amortize the orphan churn.

This directly resolves Opus N1: "Rule's 'Per-write subData is O(bytes changed)' claim is OpenGL-flavoured; Metal pays O(buffer size) per call." The added text accurately describes the Metal `subData` implementation that Opus confirmed in `metal_buffer.cpp:48-83` (orphan-on-write: alloc fresh `MTL::Buffer`, full `memcpy`, replace bindings, deferred release). The per-frame batch guidance for continuous emitters matches Opus's N2 follow-up suggestion.

No engine source files were modified. Every other file in the PR diff — `component_gpu_particle_pool.hpp`, the compute systems, shaders, prefab API, demo — is byte-identical to the Opus-approved state.

---

### Prior-review chain

- **Sonnet first-pass (2026-05-10T18:54:22Z):** approve. No blockers raised.
- **Opus recheck (2026-05-10T19:00:12Z):** approve. GPU buffer lifetime, SSBO fence, and aliased binding slots confirmed clean.
- **Sonnet re-review (2026-05-11T03:48:26Z):** needs-fix. Incorrect Metal distanceScratch doc comment (commit `c8b1bbb`) flagged.
- **Opus recheck (2026-05-11T03:57:35Z):** approve. Revert (`da4746f`) confirmed byte-identical to the originally approved tip.
- **Sonnet re-review (2026-05-11T04:55:31Z):** approve. `dirty_` removal + rule codification (`5219f42`) confirmed correct.
- **Opus recheck (2026-05-11T05:38:06Z):** approve. `subData` timing on both backends and frame-boundary ordering confirmed. N1 (Metal cost note in rule) flagged as optional polish.
- **This re-review:** N1 addressed cleanly. Approve.

---

### Outstanding items (unchanged from prior reviews, not blocking)

- CPU mirror's per-slot `lifetime_` does not track GPU decay; dead-slot scan stays conservative. Latent until Phase 2 continuous-emitter Lua API. Follow-up issue warranted before Phase 2.
- `frameData_.particleCount_` is really pool capacity; benign given the shader's `idx >= particleCount` guard.
- New systems use the explicit `Params` + `setSystemParams` form — flagged for radar; consistency wins for this PR.

---

### Test plan

- [x] Rule text accurate to Metal implementation — confirmed against Opus's `metal_buffer.cpp` analysis.
- [x] All other files unchanged from Opus-approved state — confirmed by commit diff scope.
- [ ] Cross-host smoke: Linux/OpenGL build + run — still pending (`fleet:needs-linux-smoke`). Independent of this verdict.

---

Opus recheck not required.

PR #546, opened 2026-05-08
T-126: Render: migrate light-volume propagation off CPU-built OccupancyGrid SSBO

jakildev:
## Summary

T-126 is the pragmatic unblock for T-092: decouple the propagate-shader's slot-28 SSBO from the `C_OccupancyGrid` component lifecycle and rename the SSBO/producer/constants to match the post-T-091 consumer set (only `c_propagate_light_volume` reads the bitfield now; AO migrated to screen-space sampling). The component itself is deleted in this PR — every consumer was already migrated off of it.

## Approach (Path 1)

Pure mechanical rename — slot-28 binding number unchanged, bitfield layout unchanged, the producer still rasterizes voxels + SDF light-blockers into the same parallel bitfields:

- **Shaders:** `OccupancyGrid` block → `LightOcclusionGrid`; `occupancyGetBit` → `voxelOcclusionGetBit`; Metal `OccupancyData` → `LightOcclusionData` (GLSL + Metal kept in lockstep).
- **C++:** `kBufferIndex_OccupancyGrid` → `kBufferIndex_LightOcclusionGrid`; `OccupancyGridHeader` → `LightOcclusionGridHeader`; `BUILD_OCCUPANCY_GRID` → `BUILD_LIGHT_OCCLUSION_GRID`; named resource `"OccupancyGridBuffer"` → `"LightOcclusionGridBuffer"`; file `system_build_occupancy_grid.hpp` → `system_build_light_occlusion_grid.hpp`; perf-timing field/key `buildOccupancyGrid*` → `buildLightOcclusionGrid*`.
- **Producer detached from `C_OccupancyGrid`:** archetype became `<C_VoxelPool, C_TrixelCanvasRenderBehavior>` and the bitfield storage moved into `SystemParams` (was on the per-canvas component). The component is now unreferenced and deleted in this PR.
- **Demos:** `BUILD_LIGHT_OCCLUSION_GRID` registration + include path updated in `shape_debug`, `perf_grid`, `lighting/common/lighting_demo_scene`. `setComponent(mainCanvas, C_OccupancyGrid{256})` removed from all three.
- **Docs:** `engine/render/CLAUDE.md` pipeline diagram + invariants + phased-out producer note refreshed; `.claude/skills/{simplify,review-pr}` and `.claude/agents/review-invariant-render` patterns updated to the new file name.

## Acceptance

[...]

PR #629, opened 2026-05-11
voxel: fix post-T-126 voxel-pool render regression (dangling C_VoxelSetNew spans)

jakildev:
## Summary

- After T-126 (#546) the front row of voxel-pool shapes in `shape_debug` (and any creation that calls `setComponent` on the main canvas after allocating voxel sets) collapses to a single 1×1 voxel at the world origin — sun shadows still cast at the correct positions, but the actual voxels render stacked on top of each other.
- Root cause: `C_VoxelSetNew` captures four `std::span`s into the canvas pool at allocation time. Each subsequent `setComponent` on the canvas migrates its archetype, which deep-copies `C_VoxelPool` and frees the source storage. `UPDATE_VOXEL_SET_CHILDREN`'s `updateAsChild` then writes globals into the freed span — a use-after-free that leaves the live pool's `m_voxelPositionsGlobal` all-zero, so the GPU rasterizer reads zeros and stacks every voxel at the origin. T-126 didn't author the bug; it just removed one of the canvas archetype migrations, which changed the heap-allocator's reuse pattern enough to make the latent UAF visually obvious.
- Fix routes the per-frame globals update through the pool by `voxelStartIdx_` (added by T-142) instead of through the captured span. `UPDATE_VOXEL_SET_CHILDREN` caches the canvas pool pointer on `SystemParams` for the duration of one tick and drops it at every `beginTick` so a between-frame archetype migration can't strand a stale pointer.

## Visual evidence

| Shot | Before (current master) | After (this PR) |
| --- | --- | --- |
| zoom1 | before zoom1 | after zoom1 |
| zoom4 | before zoom4 | after zoom4 |
| zoom8 | before zoom8 | after zoom8 |

The `before` column shows only the SDF row (back) plus voxel-pool sun shadows on the floor and a single 1×1 voxel near the origin. The `after` column shows both the SDF and the voxel-pool rows rendering full-size, matching the pre-T-126 baseline.

## Test plan

- [x] `fleet-build --target IRShapeDebug` and `fleet-run IRShapeDebug --auto-screenshot 10` — both shape rows render, screenshots match the pre-regression baseline.
- [x] `fleet-build --target IrredenEngineTest` and `fleet-run IrredenEngineTest` — 486/486 tests pass.
- [x] `fleet-run IRPerfGrid --mode voxel_set --grid-size 16 --auto-screenshot 10` — voxel-set grid renders correctly.
- [x] `fleet-build --target IRLightingCombined` + smoke run — lighting demo unchanged.
- [x] `fleet-build --target IRZYawStatic` and `IRZYawInteractive` — both build clean (voxel-pool shapes on these demos also hit the same code path).
- [ ] `fleet:needs-linux-smoke` — Linux/OpenGL smoke pending.


jakildev (commented):
## Review — voxel: fix post-T-126 voxel-pool render regression (dangling C_VoxelSetNew spans)

**Verdict:** needs-fix

### What this PR does

Fixes a use-after-free regression that caused voxel-pool shapes to collapse to a single 1×1 voxel at the world origin after T-126 changed the canvas archetype-migration count. Root cause: `C_VoxelSetNew` captures four `std::span`s into the canvas pool at allocation time. Every `setComponent` on the canvas deep-copies `C_VoxelPool` and frees the old storage, leaving the captured spans dangling. The fix routes the globalPositions _write_ through the live pool via `voxelStartIdx_` and caches the pool pointer in a `Params` struct reset each `beginTick`.

### Needs-fix

**N1 — `positions_` and `positionOffsets_` spans are the same dangling pointers as the fixed `globalPositions_` span** — `component_voxel_set.hpp:185–222`

The PR fixes the write to `globalPositions_` (the symptom: all-zero globals after the freed memory was reused). But `getLocalPosition(i)` still dereferences `positions_[index]` and `positionOffsets_[index]` — the same spans, captured at the same allocation time, from the same (now-freed) pool storage. These reads are undefined behavior whenever the canvas archetype migrates after voxel-set allocation.

The regression is fixed (the visible artifact is gone) because the freed pages containing local positions and offsets happen not to be reused by the heap — but ASAN on a sanitized build will report heap-use-after-free on these reads. A different heap layout or a future allocator change could produce wrong local positions.

`C_VoxelPool` already exposes `getPositions()` and `getPositionOffsets()` (const and mutable overloads, `component_voxel_pool.hpp:145–165`). The fix is the same pattern: pass them from the caller and read via `voxelStartIdx_ + i`:

```cpp
// updateAsChild signature:
bool updateAsChild(
    C_Position3D parentPosition,
    std::vector<C_PositionGlobal3D> &poolGlobalsOut,
    const std::vector<C_Position3D> &poolPositions,
    const std::vector<C_PositionOffset3D> &poolOffsets
)

// caller (system tick):
voxelSet.updateAsChild(
    position.pos_,
    pool.getPositionGlobals(),
    pool.getPositions(),
    pool.getPositionOffsets()
);

// inside the loop:
poolGlobalsOut[voxelStartIdx_ + i].pos_ =
    poolPositions[voxelStartIdx_ + i].pos_ +
    poolOffsets[voxelStartIdx_ + i].pos_ +
    parentPosition.pos_;
```

The `voxels_` span is also dangling but only its `.size()` is read in `updateAsChild` (`.size()` reads the span struct's size field stored in `C_VoxelSetNew`, not the freed pool data — safe). The deref-side UB is limited to `positions_` and `positionOffsets_` in this code path.

### Nits

**Nit1 — Explicit `Params + setSystemParams` form used instead of preferred `registerSystem`** — `system_update_voxel_set_children.hpp:16–73`

Per `engine/system/CLAUDE.md` "Per-system parameters", new system state should default to the `registerSystem` member-on-`System<N>` form; the explicit `Params + setSystemParams` is the escape hatch for maintaining systems already on that shape. The previous system had no params, so the preferred migration is:

```cpp
template <> struct System<UPDATE_VOXEL_SET_CHILDREN> {
    IREntity::EntityId lastCanvas_ = IREntity::kNullEntity;
    C_VoxelPool *lastPool_ = nullptr;

    void beginTick() { lastCanvas_ = IREntity::kNullEntity; lastPool_ = nullptr; }
    void tick(IREntity::EntityId &id, C_VoxelSetNew &vs, C_PositionGlobal3D &pos) { ... }

    static SystemId create() {
        return registerSystem<UPDATE_VOXEL_SET_CHILDREN, C_VoxelSetNew, C_PositionGlobal3D>("UpdateVoxelSetChildren");
    }
};
```

Both forms are correct and have identical runtime behavior. Happy to approve this as-is or with the migration — your call.

### Praise

- The `beginTick` → reset-`lastPool_`-to-null pattern is exactly right for the between-frame migration guard. Resetting at the start of each tick (not the end) means the cache is always rebuilt from the live archetype rather than carrying over from one that may have moved.
- The `lastCanvas_` dedup collapses the `getComponent` call to O(distinct_canvases) per tick. For the common single-canvas creation, this is one lookup per tick — no worse than `beginTick`.
- The `writableTail` bounds check in `updateAsChild` guards against an impossible-but-defensive case where the pool shrunk relative to `voxelStartIdx_` — correct defensive code.
- Before/after screenshots attached at all three zoom levels make the visual regression immediately verifiable without running the demo.

- [x] `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10` — both rows render (confirmed in PR)
- [x] `IrredenEngineTest` 486/486 (confirmed in PR)
- [x] `IRPerfGrid --mode voxel_set --grid-size 16` (confirmed in PR)
- [ ] After routing local position reads through pool: re-run all three smoke targets above
- [ ] Verify with AddressSanitizer (`-fsanitize=address`) that the heap-use-after-free on `positions_` and `positionOffsets_` is gone after the fix
- [ ] Cross-host smoke: Linux/OpenGL build + run (still pending per `fleet:needs-linux-smoke`)

Escalation: author-agent to route `positions_` and `positionOffsets_` reads through the live pool, then re-request review.

Opus recheck not required — voxel prefab code, no core render/entity/system/world/audio/GPU surfaces. The fix mechanism is straightforward (same pattern as the fixed globalPositions write).

jakildev:
Addressed N1: routed `positions_` and `positionOffsets_` reads through the live pool vectors using `voxelStartIdx_ + i` indexing, same pattern as the original `globalPositions_` fix. Updated `safeCount` to bound by pool-slice availability instead of stale span sizes. Also migrated system to `registerSystem` member-on-System<N> form (Nit1). Build + IRShapeDebug --auto-screenshot 10 pass clean on macOS/Metal. Noted as simplify finding: `getLocalPosition()` is now dead code that also reads dangling spans — worth removing if no gitignored creation code calls it. — sonnet-fleet-2

jakildev (commented):
## Re-review — voxel: fix post-T-126 voxel-pool render regression (dangling C_VoxelSetNew spans)

**Verdict:** approve

### Prior-review resolution

- ✅ `component_voxel_set.hpp:185–222` — N1: `positions_` and `positionOffsets_` dangling reads in `updateAsChild` — verified fixed at 66a0cf2. `updateAsChild` now accepts `const std::vector<C_Position3D> &poolPositions` and `const std::vector<C_PositionOffset3D> &poolOffsets` by reference. The loop body reads `poolPositions[voxelStartIdx_ + i]` and `poolOffsets[voxelStartIdx_ + i]` from the live pool vectors. `safeCount` is now bounded by `availPositions` and `availOffsets` (pool sizes minus `voxelStartIdx_`). ASAN-visible heap-use-after-free path closed.
- ✅ `system_update_voxel_set_children.hpp:16–73` — Nit1: explicit `Params + setSystemParams` form — verified fixed at 66a0cf2. System is now `template<> struct System<UPDATE_VOXEL_SET_CHILDREN>` with `lastCanvas_`, `lastPool_` member fields, `beginTick()`, `tick()`, and `create()`. Behavior preserved exactly.

### Nits

- `component_voxel_set.hpp:185–187` — `getLocalPosition(int index)` still reads `positions_[index].pos_ + positionOffsets_[index].pos_` — the same spans documented as dangling in the comment above `updateAsChild`. Confirmed dead in engine code (no callers outside this file). The author flagged it for removal in the PR comment; safe to drop now that both spans are documented dangling. Removing avoids future confusion for any reader who sees the function beside the comment warning against reading spans.
- `component_voxel_set.hpp:220` — `static_cast<size_t>(voxels_.size())` in the `safeCount` min is redundant with `numVoxels_` (the constructor min-span guard sets `numVoxels_` to the same value at allocation time). Not harmful but adds noise alongside the other bounds.

### Praise

- `updateAsChild` signature change and pool-vector threading is the exact pattern recommended in the prior review — clean mechanical application.
- `safeCount` now bounds by three independent pool-vector tails (`availPositions`, `availOffsets`, `writableTail`) rather than the now-stale span sizes — correct and defensive.
- `registerSystem` migration removes the lambda captures and consolidates state as named fields with named functions — easier to grep and step through than the old Params form.

### Test plan

- [x] `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10` — both rows render (confirmed in PR comment)
- [x] `IrredenEngineTest` 486/486 (confirmed in PR body)
- [x] `IRPerfGrid --mode voxel_set --grid-size 16` (confirmed in PR body)
- [ ] `fleet:needs-linux-smoke` — Linux/OpenGL cross-host smoke still pending

Escalation: none. Safe for merge once Linux smoke clears.

Opus recheck not required — voxel prefab code, no core render/entity/system/world/audio/GPU surfaces.

jakildev:
Addressed both nits:

- `component_voxel_set.hpp:185-187` — removed `getLocalPosition()`. Confirmed via Grep no callers anywhere in the tree; the function read the same dangling `positions_` / `positionOffsets_` spans the rest of this PR was eliminating.
- `component_voxel_set.hpp:220` — dropped `static_cast<size_t>(voxels_.size())` from the inner min. `numVoxels_` already bounds by that same value via the constructor min-span guard at allocation, so the outer `IRMath::min(numVoxels_, …)` covers it.

Build clean (`fleet-build --target IRShapeDebug`). `fleet-run IRShapeDebug --auto-screenshot 10` captured all 6 shots on macOS/Metal, no visual regression. `fleet:approved` stays valid — cleanups didn't change behavior. — opus-worker

jakildev (commented):
## Opus Recheck — voxel: fix post-T-126 voxel-pool render regression

**Verdict:** approve

### Why Opus picked this up

Sonnet's two passes (first needs-fix on N1 dangling reads, then approve) explicitly disclaimed Opus recheck. I picked the PR up because lifetime/ownership is on the Opus core-invariants list regardless of Sonnet's verdict, and the regression here is fundamentally a `std::span` UAF. On closer read I confirm Sonnet's verdict: the fix is correct and well-scoped.

### Independent lifetime audit

The mechanic: `C_VoxelPool` owns four `std::vector`s (`m_voxelPositions`, `m_voxelPositionsOffset`, `m_voxelPositionsGlobal`, `m_voxelColors`), pre-`resize`d at construction and never re-allocated. A `setComponent` on the canvas migrates its archetype — `C_VoxelPool` is deep-copied to new storage and the source vectors are freed. Every `std::span` captured into `C_VoxelSetNew` at allocation goes dangling.

The fix re-routes `updateAsChild`'s per-frame critical path through the live pool vectors by index (`poolPositions[voxelStartIdx_ + i]`, etc.). `safeCount` now bounds against the live tails (`availPositions`, `availOffsets`, `writableTail`) plus the stable scalar `numVoxels_`. The system fetches the pool fresh per tick and caches it intra-tick across multiple voxel sets — `lastPool_` is cleared in `beginTick`, so a between-frame archetype migration cannot strand a stale pointer. `pool.setEntityIdForRange(...)` mutates in-place and does **not** trigger an archetype migration, so the cached `pool` reference stays valid through one tick's iteration.

The `registerSystem<UPDATE_VOXEL_SET_CHILDREN, ...>` migration matches `engine/system/CLAUDE.md`'s preferred member-on-`System<N>` form. The follow-up cleanup commit `42b8475` drops a redundant `voxels_.size()` from the inner `safeCount` min (`numVoxels_` already covers it via the constructor min-span guard) and removes dead `getLocalPosition()` — both behavior-preserving.

### Non-blocking observation (pre-existing, out of this PR's scope)

The prefix comment in the patched `updateAsChild` correctly notes "all four span members ... become invalid after such a migration" — but `C_VoxelSetNew::changeVoxelColor`, `changeVoxelColorAll`, `deactivateAll`, `activateAll`, and `reshape` still dereference the captured `voxels_` span (the constructor also writes through `positions_`/`voxels_`). These are the same UAF shape that `updateAsChild` had, just on a colder path: T-126 only made the per-frame `updateAsChild` UAF visually obvious because that path runs every frame. The mutator UAFs are latent — they only fire if a creation mutates voxel state after a canvas archetype migration, which `IRShapeDebug` doesn't reproduce.

Worth a follow-up task (migrate the mutators to take pool refs the same way `updateAsChild` now does, or move them into a system / namespace where the pool can be fetched on demand). Not a blocker for this PR — the regression is closed, the fix is mechanically correct, and the remaining UAFs are pre-existing.

### Verdict

Safe for merge. Linux/OpenGL cross-host smoke (PR body notes `fleet:needs-linux-smoke` as pending) is the author's call — the change is platform-agnostic C++ lifetime with no shader / backend surface, so my role's render-path-only smoke-tag rule does not auto-fire.

🤖 Opus recheck by Claude Opus 4.7 (role-opus-reviewer)

PR #639, opened 2026-05-12
T-158: Migrate final composite + sprites cluster to member-on-System<N>

### LINKED ISSUE #580: Systems: member-on-System<N> registration helper to retire the explicit Params + setSystemParams boilerplate
## Scope

The canonical pattern for system-owned state today is the explicit
\`Params + setSystemParams\` shape documented in
\`engine/system/CLAUDE.md\`. It correctly retired the function-local
\`static\` anti-pattern but kept five lines of boilerplate per
system:

\`\`\`cpp
SystemId create() {
    auto paramsOwner = std::make_unique<Params>();   // 1. allocate
    Params *p = paramsOwner.get();                   // 2. raw-ptr grab
    SystemId id = createSystem<...>(
        \"Name\",
        [p](Components&...) { /* tick */ },          // 3. capture in tick
        [p]() { /* beginTick */ }                    // 4. capture in begin
    );
    setSystemParams(id, std::move(paramsOwner));     // 5. hand off
    return id;
}
\`\`\`

The boilerplate is forced by ordering — \`setSystemParams\` runs
**after** \`createSystem\`, so the lambdas can't query the params
back from the system id at construction time. Every consumer pays
the same boilerplate.

This ticket adds a thin framework helper so the \`System<NAME>\`
specialization itself becomes the params struct, with member
functions as hooks. Same lifetime contract, same per-tick cost,
much cleaner author surface.

## Suggested model

**[opus]** — touches \`engine/system/\` core templates and the
public \`ir_system.hpp\` API. Migration touches every prefab system
that currently uses the explicit Params path. Cross-cutting; needs
careful concept/SFINAE design and one round of profiling to
confirm no regression.

## Proposed shape

The \`System<NAME>\` specialization holds member fields (the params)
and named member functions (the hooks):

\`\`\`cpp
template <> struct System<HITBOX_MOUSE_TEST_GUI> {
    // Member variables = params. No nested Params struct.
    vec2 mouseGuiTrixel_ = vec2(0.0f);

    // Per-entity tick — normal member function, no lambda, no captures.
    void tick(C_HitBox2DGui &hitbox, const C_GuiPosition &guiPos) {
        const vec2 lo = vec2(guiPos.pos_);
        const vec2 hi = vec2(guiPos.pos_ + hitbox.size_);
        hitbox.hovered_ =
            mouseGuiTrixel_.x >= lo.x && mouseGuiTrixel_.x < hi.x &&
            mouseGuiTrixel_.y >= lo.y && mouseGuiTrixel_.y < hi.y;
    }

    void beginTick() {
        // ... compute frame-scoped state ...
        mouseGuiTrixel_ = mouseFb / fbRes * guiSize;
    }

    static SystemId create() {
        return registerSystem<HITBOX_MOUSE_TEST_GUI,
                              C_HitBox2DGui,
                              C_GuiPosition>(\"HitBoxMouseTestGui\");
    }
};
\`\`\`

The helper:

\`\`\`cpp
template <typename T> concept HasBeginTick    = requires(T t) { t.beginTick(); };
template <typename T> concept HasEndTick      = requires(T t) { t.endTick(); };
template <typename T> concept HasRelationTick = requires(T t, IREntity::EntityRecord r) {
    t.relationTick(r);
};

template <SystemName N, typename... Components, typename... ExtraArgs>
SystemId registerSystem(std::string name, ExtraArgs&&... extra) {
    auto instance = std::make_unique<System<N>>();
    System<N> *p = instance.get();

    auto tickFn = [p](Components&... cs) { p->tick(cs...); };

    std::function<void()> beginFn = []() {};
    std::function<void()> endFn   = []() {};
    if constexpr (HasBeginTick<System<N>>) beginFn = [p]() { p->beginTick(); };
    if constexpr (HasEndTick<System<N>>)   endFn   = [p]() { p->endTick(); };

    SystemId id = createSystem<Components...>(
        std::move(name), tickFn, beginFn, endFn,
        std::forward<ExtraArgs>(extra)...
    );
    setSystemParams(id, std::move(instance));
    return id;
}
\`\`\`

The implementation needs to handle:
- The existing \`createSystem<...>\` overloads (entity-id form,
  archetype-batch form, with-relation form).
- \`RelationParams<...>\` — detected via \`HasRelationTick\` and
  forwarded.
- Systems with no \`beginTick\` / \`endTick\` (no-op fallbacks).
- Systems with no params (empty \`System<N>\` struct still
  allocates a single-byte instance — fine; same per-tick cost as
  no-params today).

### 1. Add the helper

- \`registerSystem<SystemName, Components...>(name, ...)\` lands in
  \`engine/system/include/irreden/ir_system.hpp\` next to the
  existing \`createSystem\` and \`setSystemParams\` exports.
- Concepts/SFINAE detection for \`tick\`, \`beginTick\`, \`endTick\`,
  \`relationTick\` member functions.
- Same runtime lifetime as today: instance owned by the existing
  \`m_systemParams\` slot, freed when the system entity is destroyed.

### 2. Migrate every prefab system using explicit Params

\`grep -l \"setSystemParams(systemId\" engine/prefabs/\` is the
canonical list. As of writing, that's:

- \`engine/prefabs/irreden/render/systems/system_fog_to_trixel.hpp\`
- \`...system_trixel_to_trixel.hpp\`
- \`...system_bake_sun_shadow_map.hpp\`
- \`...system_lighting_to_trixel.hpp\`
- \`...system_shapes_to_trixel.hpp\`
- \`...system_debug_culling_minimap.hpp\`
- \`...system_framebuffer_to_screen.hpp\`
- \`...system_screen_residual_rotate.hpp\`
- \`...system_compute_sun_shadow.hpp\`
- \`...system_compute_light_volume.hpp\`
- \`...system_voxel_to_trixel.hpp\` (×2 systems in this file)
- \`...system_text_to_trixel.hpp\`
- \`...system_build_light_occlusion_grid.hpp\`
- \`...system_trixel_to_framebuffer.hpp\`
- \`...system_compute_voxel_ao.hpp\`
- \`...system_sprites_to_screen.hpp\`
- \`engine/prefabs/irreden/input/systems/system_hitbox_mouse_test_gui.hpp\`

(Re-run the grep at implementation time — list may have shifted.)

Each migrated to the member-on-\`System<N>\` shape. Suggest doing
this as a series of small PRs (one per system or one per
subsystem cluster), not one mega-migration, so each can be
reviewed cleanly. Land the helper in PR 1; migrations follow.

### 3. CLAUDE.md updates

- **\`engine/system/CLAUDE.md\` \"Per-system parameters\" section**
  rewrite. Two forms, in order of preference:
  1. **Preferred**: member-on-\`System<N>\` via \`registerSystem\`
     (the new path). Show the canonical example.
  2. **Explicit**: \`Params + setSystemParams\` (the existing
     path). Documented as the escape hatch — when to use it
     (custom Params lifetime, separate Params type, multi-Params
     systems).
- **Update the \"Don't use function-local \`static\`\" section**
  to point at the preferred form first.
- **\`engine/CLAUDE.md\` style section** gains a one-line bullet:
  \"Prefer the member-on-\`System<N>\` form for system-owned
  state. See \`engine/system/CLAUDE.md\` for the registration
  helper.\"

### 4. Deviations file

If \`.fleet/status/system-static-deviations.md\` exists, rename or
extend it to track explicit-Params holdouts during migration.
Each migration PR removes the migrated file from the deviation
list.

### 5. Tests

- Existing system tests pass unchanged (the migration must
  preserve every behavior — same tick body, same
  beginTick/endTick semantics).
- New test in \`test/system/\` that covers a small example
  \`System<N>\` with all four hooks
  (\`tick\`/\`beginTick\`/\`endTick\`/\`relationTick\`) and verifies the
  member fields persist across ticks, are not shared between
  multiple instances of \`System<N>\`, and are freed on system
  destroy.

### 6. No runtime regression

- The \`[p]\` capture in the current pattern and the implicit
  \`this\` capture in the new pattern compile to the same
  per-tick instructions. Profiling sanity check on
  \`IrredenEngineTest\` and \`IRShapeDebug\` to confirm no frame-
  time delta.

## Why now

The explicit \`Params + setSystemParams\` pattern was the right
escape from function-local \`static\`, but it was always a
half-step — every consumer pays five lines of boilerplate to
work around the constructor-order constraint. The member-on-System
form removes the boilerplate without giving anything up.

Adding the helper now (before the prefab system count grows
further) keeps the migration tractable. The longer the explicit
pattern is the documented canonical form, the more new systems
ship in that shape and the bigger the eventual rewrite.

## Notes

- Backward compat is preserved end-to-end: \`createSystem\`,
  \`setSystemParams\`, \`getSystemParams\` all stay public. Existing
  out-of-tree consumers don't break.
- The migration PRs are pure refactors — each one's diff should
  be 1:1 between old and new shape, with no behavior change. Easy
  to review, easy to bisect if anything breaks.
- The \"Per-system parameters\" rewrite in CLAUDE.md should land
  alongside the helper PR (PR 1), so subsequent migration PRs
  point at the up-to-date doc.

jakildev:
## Summary
- Migrates four final-composite render pipeline systems from explicit `Params` + `setSystemParams` + lambda-capture to the preferred `registerSystem` member-on-`System<N>` form.
- **TRIXEL_TO_FRAMEBUFFER**: entity-id tick form preserved; `Buffer`/`ShaderProgram`/`VAO` fields lifted from nested `Params` to direct `System<N>` members, initialized via `getSystemParams` after `registerSystem` returns.
- **FRAMEBUFFER_TO_SCREEN**: per-component tick; same field lift pattern; `frameData_` member default-initializes correctly.
- **SCREEN_SPACE_RESIDUAL_ROTATE**: same structure as `FRAMEBUFFER_TO_SCREEN`; residualYaw field on `frameData_` preserved.
- **SPRITE_TO_SCREEN**: `endTick`, `packGroupToScratch`, `uploadFrameData`, `bindPipeline` converted from static `Params&` helpers to member functions; `entryLess`, `buildGroups`, `computeScreenAnchor`, `unbindPipeline` remain static (no instance state). Vector reserves moved to `create()` via `getSystemParams`.

All logic is preserved 1:1. This is the final PR in the migration chain started in #580.

## Test plan
- [x] `fleet-build --target IRShapeDebug` — builds clean
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — all 6 shots captured, no errors, render pipeline intact
- [ ] Linux/OpenGL cross-host smoke (pending `fleet:needs-linux-smoke`)

## Notes for reviewer
Pre-existing `getComponentOptional<C_TrixelCanvasRenderBehavior>(entity)` and `getComponentOptional<C_ZoomLevel>(entity)` in `TRIXEL_TO_FRAMEBUFFER::tick()` are preserved as-is — optional per-entity lookups that can't be moved to template params without changing which entities match. Not introduced by this PR.

Closes #580

### DIFF
--- a/engine/prefabs/irreden/render/systems/system_framebuffer_to_screen.hpp
+++ b/engine/prefabs/irreden/render/systems/system_framebuffer_to_screen.hpp
@@ -24,12 +24,36 @@ using namespace IRMath;
 namespace IRSystem {
 
 template <> struct System<FRAMEBUFFER_TO_SCREEN> {
-    struct Params {
-        Buffer *frameDataBuf_ = nullptr;
-        ShaderProgram *program_ = nullptr;
-        VAO *quadVao_ = nullptr;
-        FrameDataFramebuffer frameData_{};
-    };
+    Buffer *frameDataBuf_ = nullptr;
+    ShaderProgram *program_ = nullptr;
+    VAO *quadVao_ = nullptr;
+    FrameDataFramebuffer frameData_{};
+
+    void tick(
+        const C_TrixelCanvasFramebuffer &framebuffer,
+        const C_Position3D &cameraPosition,
+        const C_Name &name
+    ) {
+        framebuffer.bindTextures(0, 1);
+        frameData_.mvpMatrix =
+            calcProjectionMatrix() * calcModelMatrix(
+                                         framebuffer.getResolution(),
+                                         framebuffer.getResolutionPlusBuffer(),
+                                         cameraPosition.pos_,
+                                         IRRender::getCameraPosition2DIso(),
+                                         name.name_
+                                     );
+        frameDataBuf_->subData(0, sizeof(FrameDataFramebuffer), &frameData_);
+        IRRender::device()->setPolygonMode(PolygonMode::FILL);
+        IRRender::device()->drawArrays(DrawMode::TRIANGLES, 0, 6);
+    }
+
+    void beginTick() {
+        bindDefaultFramebuffer();
+        clearDefaultFramebuffer();
+        program_->use();
+        quadVao_->bind();
+    }
 
     static SystemId create() {
         IRRender::createNamedResource<ShaderProgram>(
@@ -48,41 +72,16 @@ template <> struct System<FRAMEBUFFER_TO_SCREEN> {
             kBufferIndex_FramebufferFrameDataUniform
         );
 
-        auto paramsOwner = std::make_unique<Params>();
-        Params *p = paramsOwner.get();
-        p->frameDataBuf_ = IRRender::getNamedResource<Buffer>("FramebufferToScreenFrameData");
-        p->program_ = IRRender::getNamedResource<ShaderProgram>("FramebufferToScreenProgram");
-        p->quadVao_ = IRRender::getNamedResource<VAO>("QuadVAOArrays");
-
-        SystemId systemId = createSystem<C_TrixelCanvasFramebuffer, C_Position3D, C_Name>(
-            "FramebufferToScreen",
-            [p](const C_TrixelCanvasFramebuffer &framebuffer,
-                const C_Position3D &cameraPosition,
-                const C_Name &name) {
-                framebuffer.bindTextures(0, 1);
-                p->frameData_.mvpMatrix =
-                    calcProjectionMatrix() * calcModelMatrix(
-                                                 framebuffer.getResolution(),
-                                                 framebuffer.getResolutionPlusBuffer(),
-                                                 cameraPosition.pos_,
-                                                 IRRender::getCameraPosition2DIso(),
-                                                 name.name_
-                                             );
-                p->frameDataBuf_->subData(0, sizeof(FrameDataFramebuffer), &p->frameData_);
-                IRRender::device()->setPolygonMode(PolygonMode::FILL);
-                IRRender::device()->drawArrays(DrawMode::TRIANGLES, 0, 6);
-            },
-            [p]() {
-                bindDefaultFramebuffer();
-                clearDefaultFramebuffer();
-                p->program_->use();
-                p->quadVao_->bind();
-            }
-        );
-
-        setSystemParams(systemId, std::move(paramsOwner));
-        IRRender::tagGpuStage(systemId, "fbToScreen");
-        return systemId;
+        SystemId id = registerSystem<FRAMEBUFFER_TO_SCREEN,
+                                     C_TrixelCanvasFramebuffer,
+                                     C_Position3D,
+                                     C_Name>("FramebufferToScreen");
+        auto *sys = getSystemParams<System<FRAMEBUFFER_TO_SCREEN>>(id);
+        sys->frameDataBuf_ = IRRender::getNamedResource<Buffer>("FramebufferToScreenFrameData");
+        sys->program_ = IRRender::getNamedResource<ShaderProgram>("FramebufferToScreenProgram");
+        sys->quadVao_ = IRRender::getNamedResource<VAO>("QuadVAOArrays");
+        IRRender::tagGpuStage(id, "fbToScreen");
+        return id;
     }
 
   private:
@@ -98,11 +97,10 @@ template <> struct System<FRAMEBUFFER_TO_SCREEN> {
         // also known as screen center
         float xOffset = IRRender::getViewport().x / 2.0f;
         float yOffset = IRRender::getViewport().y / 2.0f;
-        vec2 offset = vec2(xOffset, yOffset) +
-                      isoDeltaToScreenDelta(
-                          pos3DtoPos2DIso(cameraPosition),
-                          IRRender::getTriangleStepSizeScreen()
-                      );
+        vec2 offset = vec2(xOffset, yOffset) + isoDeltaToScreenDelta(
+                                                   pos3DtoPos2DIso(cameraPosition),
+                                                   IRRender::getTriangleStepSizeScreen()
+                                               );
 
         mat4 model = mat4(1.0f);
 
--- a/engine/prefabs/irreden/render/systems/system_screen_residual_rotate.hpp
+++ b/engine/prefabs/irreden/render/systems/system_screen_residual_rotate.hpp
@@ -26,12 +26,37 @@ namespace IRSystem {
 /// (passthrough branch in the fragment shader bypasses the bilinear blend),
 /// rotated bilinear sample otherwise.
 template <> struct System<SCREEN_SPACE_RESIDUAL_ROTATE> {
-    struct Params {
-        Buffer *frameDataBuf_ = nullptr;
-        ShaderProgram *program_ = nullptr;
-        VAO *quadVao_ = nullptr;
-        FrameDataScreenResidualRotate frameData_{};
-    };
+    Buffer *frameDataBuf_ = nullptr;
+    ShaderProgram *program_ = nullptr;
+    VAO *quadVao_ = nullptr;
+    FrameDataScreenResidualRotate frameData_{};
+
+    void tick(
+        const C_TrixelCanvasFramebuffer &framebuffer,
+        const C_Position3D &cameraPosition,
+        const C_Name &name
+    ) {
+        framebuffer.bindTextures(0, 1);
+        frameData_.mvpMatrix =
+            calcProjectionMatrix() * calcModelMatrix(
+                                         framebuffer.getResolution(),
+                                         framebuffer.getResolutionPlusBuffer(),
+                                         cameraPosition.pos_,
+                                         IRRender::getCameraPosition2DIso(),
+                                         name.name_
+                                     );
+        frameData_.residualYaw = IRPrefab::Camera::getResidualYaw();
+        frameDataBuf_->subData(0, sizeof(FrameDataScreenResidualRotate), &frameData_);
+        IRRender::device()->setPolygonMode(PolygonMode::FILL);
+        IRRender::device()->drawArrays(DrawMode::TRIANGLES, 0, 6);
+    }
+
+    void beginTick() {
+        bindDefaultFramebuffer();
+        clearDefaultFramebuffer();
+        program_->use();
+        quadVao_->bind();
+    }
 
     static SystemId create() {
         IRRender::createNamedResource<ShaderProgram>(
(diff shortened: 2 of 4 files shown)


jakildev (commented):
## Review — T-158: Migrate final composite + sprites cluster to member-on-System<N>

**Verdict:** approve (Opus recheck required — render pipeline)

Clean mechanical migration of four final-composite render systems
(TRIXEL_TO_FRAMEBUFFER, FRAMEBUFFER_TO_SCREEN, SCREEN_SPACE_RESIDUAL_ROTATE,
SPRITE_TO_SCREEN) from the Params + setSystemParams + lambda pattern to
registerSystem<N, Cs...> with member fields and named member functions.
Logic inside all tick/beginTick/endTick bodies is preserved byte-for-byte
from the original lambda form.

---

### ECS invariants

- No new per-entity getComponent calls introduced. The pre-existing
  getComponentOptional<C_TrixelCanvasRenderBehavior> and
  getComponentOptional<C_ZoomLevel> in TRIXEL_TO_FRAMEBUFFER::tick are
  explicitly acknowledged in the PR body as entity-optional lookups that
  cannot be lifted to template params without changing archetype matching.
  Accepted as pre-existing design. ✅
- IREntity::getComponent<C_TrixelCanvasFramebuffer>("mainFramebuffer") and
  getComponent<C_FrameDataTrixelToFramebuffer>("mainFramebuffer") inside
  TRIXEL_TO_FRAMEBUFFER::tick and beginTick are named-entity (not per-iterated-
  entity) lookups — pre-existing pattern, not a per-entity getComponent violation. ✅
- No mid-iteration structural changes (createEntity/removeEntity) in any tick path. ✅
- beginTick()/endTick() are void() with no component parameters on all four systems. ✅
- SPRITE_TO_SCREEN::endTick correctly guards with if (entries_.empty()) return
  before sorting and any GPU work — archetype-empty case safe. ✅
- SPRITE_TO_SCREEN::tick appends to entries_ (std::vector::push_back). Reserve(256)
  applied during create() via getSystemParams; amortizes to zero reallocations
  after warm-up, same as the original Params pattern. ✅
- SPRITE_TO_SCREEN correctly reads C_PositionGlobal3D + C_PositionOffset3D for
  rendered world position. Camera-facing systems (FRAMEBUFFER_TO_SCREEN,
  SCREEN_SPACE_RESIDUAL_ROTATE) use C_Position3D for the camera entity — correct
  and pre-existing distinction. ✅

### Ownership / lifetime

- All GPU-resource pointer fields (Buffer*, ShaderProgram*, VAO*) are raw
  non-owning pointers to resources managed by IRRender named-resource registry —
  consistent with existing convention. ✅
- getSystemParams<System<N>>(id) used exclusively during create() to populate
  fields; no lingering raw alias to Params held across tick dispatch. ✅

### Render pipeline

- No shaders modified. No GPU struct layout changes. No new binding-point
  constants. ✅
- memoryBarrier(SHADER_STORAGE) after the trixel draw in TRIXEL_TO_FRAMEBUFFER::tick
  preserved. ✅
- tagGpuStage preserved for TRIXEL_TO_FRAMEBUFFER ("trixelToFb"),
  FRAMEBUFFER_TO_SCREEN ("fbToScreen"), SCREEN_SPACE_RESIDUAL_ROTATE
  ("screenSpaceResidualRotate"). Absence from SPRITE_TO_SCREEN is pre-existing. ✅
- Buffer bind indices unchanged on CPU and shader sides for all four systems. ✅

### Naming / style

All promoted member fields carry correct trailing-underscore naming
(frameDataBuf_, hoveredIdBuf_, program_, quadVao_, frameData_, entries_,
gpuScratch_, groups_, overflowWarned_). ✅

### Praise

- This is the final PR completing the migration chain from #580 — eliminating
  the Params + lambda capture pattern from the entire render system tree.
- SPRITE_TO_SCREEN::endTick, packGroupToScratch, uploadFrameData, bindPipeline
  are correctly split into member vs static: instance-state helpers promoted to
  members, pure-logic helpers (entryLess, buildGroups, computeScreenAnchor,
  unbindPipeline) correctly kept static. Precise access-level reasoning.

---

- [x] fleet-build --target IRShapeDebug (green per PR body)
- [x] fleet-run IRShapeDebug --auto-screenshot 10, all 6 shots captured (green per PR body)
- [ ] Linux/OpenGL cross-host smoke (tagged fleet:needs-linux-smoke)

---

Escalation: please Opus-recheck before merge (touches: engine/prefabs/irreden/render/ final composite chain).
Two items to confirm:
1. TRIXEL_TO_FRAMEBUFFER::tick takes IREntity::EntityId by value as its first
   parameter to request entity-id tick form — confirm registerSystem template
   dispatch correctly detects this and threads entity IDs through (the old lambda
   form used EntityId& by reference; value vs reference for an integral type is
   equivalent, but the detection mechanism is worth one look).
2. System<N> objects must not be invalidated (e.g. by vector reallocation in the
   system registry) between create() and the first tick dispatch — same concern
   as raised in T-157 review. Confirm getSystemParams pointer stability.

Opus recheck required: render pipeline systems + registerSystem entity-id tick dispatch + System<N> pointer stability post-create.

jakildev (commented):
## Opus recheck — T-158: Final composite + sprites cluster member-on-System<N>

**Verdict:** approve

Sonnet's first-pass coverage of ECS invariants, ownership/lifetime, naming,
and render-pipeline byte-for-byte preservation is sound — I read the four
hpp diffs end-to-end and have nothing to add to that checklist. This pass
focuses on Sonnet's two escalated concerns, since both touch
`registerSystem<>` machinery and can't be confirmed without reading the
underlying dispatch helpers.

---

### Concern 1 — `EntityId` by value vs reference in entity-id tick dispatch

Sonnet flagged that `TRIXEL_TO_FRAMEBUFFER::tick(IREntity::EntityId entity, ...)`
now takes the entity id **by value**, whereas the old lambda form had
`IREntity::EntityId &entity`. Confirmed safe; the path is:

1. `MakeMemberTickFn::build` at `engine/system/include/irreden/ir_system.hpp:103`
   detects the per-entity-id member shape via the concept
   `requires(T &s, EntityId id, Cs &...cs) { s.tick(id, cs...); }`. For
   our tick member `void tick(EntityId, const C_TriangleCanvasTextures&,
   const C_Name&)`, the by-value `EntityId id` in the concept's expression
   binds the lvalue argument trivially (copy). The per-component shape is
   tried first and correctly fails (would need to drop the `EntityId`
   parameter); the entity-id shape matches; the per-archetype shape isn't
   reached.
2. The synthesized lambda is
   `[p](EntityId id, Cs &...cs) { p->tick(id, cs...); }` at
   `ir_system.hpp:104` — value-form parameter, same as the member.
3. `SystemManager::insertTickFunction` at
   `engine/system/include/irreden/system/system_manager.hpp:220` selects
   the dispatch branch using
   `InvocableWithEntityId<F, Components...>` =
   `is_invocable_v<F, EntityId&, Components&...>`
   (`ir_system_types.hpp:194`). The synthesized lambda accepts
   `(EntityId&, Cs&...)` because by-value `EntityId id` binds an
   `EntityId&` argument by copy. ✅
4. Per-entity dispatch at `system_manager.hpp:225-232` calls
   `functionTick(components[i]...)` where `components[0]` is
   `node->entities_[i]` (an `EntityId&` lvalue). Pass-by-value copy is
   semantically identical to pass-by-reference for an integral id
   that the body never mutates — and the new tick body doesn't assign
   to `entity`, only reads it via `getComponentOptional<>(entity)`. ✅

Net: detection mechanism is correct, and the value-vs-reference change
on the member is purely cosmetic.

### Concern 2 — `System<N>` pointer stability post-create

Sonnet raised the same question they did on T-157: can the `sys` pointer
returned by `getSystemParams<System<N>>(id)` during `create()` be
invalidated before the first tick dispatch (e.g. by reallocation of a
system registry vector)?

Walked the ownership chain at
`engine/system/include/irreden/ir_system.hpp:185-208`:

1. `registerSystem` calls `std::make_unique<SystemT>()` (line 191) —
   the `SystemT` instance lives on the heap at a fixed address.
2. `SystemT *p = instance.get()` (line 192) captures the raw heap
   pointer; the four `Make*Fn` helpers capture `p` by value into the
   dispatch lambdas (lines 194-197).
3. `createSystem<Components...>(...)` registers the lambdas. The
   lambdas already hold `p` by value at this point — nothing the system
   registry does to its internal vectors can move the heap allocation.
4. `setSystemParams(id, std::move(instance))` transfers ownership of
   the `unique_ptr<SystemT>` into
   `m_systemParams[id] = std::make_unique<ISystemParamsImpl<Params>>(std::move(params))`
   (`system_manager.hpp:108-111`). The move chain transfers the raw heap
   pointer; the `SystemT` object stays at the same address.

The pointer-stability guarantee rests on two facts: (a) `SystemT` is
heap-allocated via `make_unique` and never relocated; (b) the lambdas
capture `p` by value, so the `m_systemParams` vector's internal
reallocation is irrelevant — even if it moves the outer
`unique_ptr<ISystemParamsImpl>` slot, the inner `SystemT` heap allocation
is untouched. ✅

The same reasoning held for T-157 and lands the same conclusion here.

### Render pipeline spot-checks beyond Sonnet's coverage

- `MakeMemberTickFn`'s first concept (per-component, no entity-id) is
  correctly skipped for `TRIXEL_TO_FRAMEBUFFER`: the member tick's
  leading `EntityId` parameter makes
  `s.tick(C_TriangleCanvasTextures&, C_Name&)` ill-formed, so the
  concept-chain falls through to the entity-id branch as intended. No
  silent dispatch-form drift.
- `SPRITE_TO_SCREEN::tick` correctly hits the per-component branch
  (no leading `EntityId`), and `endTick`, `packGroupToScratch`,
  `uploadFrameData`, `bindPipeline` correctly migrate from
  `static void X(Params&)` to non-static members reading via implicit
  `this`. The static-vs-member split (`entryLess`, `buildGroups`,
  `computeScreenAnchor`, `unbindPipeline` kept static) is correct —
  those four touch no instance state.
- `frameData_{}` default-initialization in all three `frameData_`-holding
  systems is fine; `FrameDataFramebuffer`, `FrameDataScreenResidualRotate`,
  and `FrameDataSpritesToScreen` are POD-shaped and value-init zeroes the
  matrix fields before the first `subData` upload.
- Vector reserves for `SPRITE_TO_SCREEN::entries_` and
  `SPRITE_TO_SCREEN::gpuScratch_` correctly moved to `create()` post-
  `getSystemParams`; reserves only need to happen once, and the high-water
  capacity persists for the lifetime of the system entity.

### Praise

Final PR in the migration chain from #580 closes out the
`Params + setSystemParams + lambda-capture` pattern across the render
tree. Mechanical 1:1 conversion with no behavioral drift, and the
member/static split on `SPRITE_TO_SCREEN`'s helpers is precise — only
instance-state methods promoted.

---

Linux/OpenGL cross-host smoke (`fleet:needs-linux-smoke`) is still
outstanding and will gate merge; not blocking from the review side.

PR #650, opened 2026-05-12
T-162: engine/entity: ECS singleton-component infrastructure

### LINKED ISSUE #646: engine/entity: ECS singleton-component infrastructure
# ECS singleton-component infrastructure

## Area

`engine/entity/`, `engine/script/` (Lua bindings), `engine/prefabs/irreden/common/` (retrofit existing improvisation).

## Suggested model

`[opus]` — adds an engine-wide API surface; needs careful naming and Lua reflection design.

## Context

The engine has one production singleton-shaped use case today: `C_GlobalModifiers` on a `"modifierGlobals"` entity, with a hand-rolled cached-ID lookup via `detail::globalsEntityId()` and `currentGlobalModifiersPtr()` (`engine/prefabs/irreden/common/modifier.hpp:185-191`). Every future "global game state component" — the sim clock (#200), per-world configuration, debug overlay state, game-side rules holders — wants the same shape: lazy-init on first access, one canonical entity per type, cached ID lookup.

Today every singleton-shaped case has to invent that lazy-init pattern itself. The modifier framework's version is a fine improvisation but it's invisible from outside (a `detail::` helper, no public API). The next person needing a singleton has nothing to copy.

This task generalizes the pattern into an explicit `IRECS::` API and retrofits the modifier framework to use it as validation.

Filed as the prerequisite for **#200** (deterministic sim clock + cycle/timer framework). Plan at `~/.fleet/plans/issue-200.md` documents the full architecture; this task is **Task A** in that plan.

## Scope

1. **`IRECS::singleton<T>()` API.** Three entry points:
   - `template <typename C> C& singleton()` — lazy-init; returns ref to THE component
   - `template <typename C> C* singletonOrNull()` — no-create variant
   - `template <typename C> IREntity::EntityId singletonEntity()` — for systems iterating singletons by archetype
2. **Lazy-init mechanics.** First `singleton<T>()` call creates a named entity (e.g. `"singleton:<typeid>"`) with the default-constructed component. Cached entity ID per type via `std::unordered_map<std::type_index, EntityId>` in the entity manager.
3. **Lua binding.** `ir.entity.singleton(componentTypeId)` returns the singleton as a Lua table, following the existing component-reflection conventions in `engine/script/src/lua_script.cpp`.
4. **Retrofit `C_GlobalModifiers`** in `engine/prefabs/irreden/common/modifier.hpp:185-191` to use `IRECS::singleton<C_GlobalModifiers>()`. Replaces the `detail::globalsEntityId()` and `currentGlobalModifiersPtr()` helpers. Validates that the new API is at least as ergonomic as the existing improvisation.
5. **Documentation in `engine/entity/CLAUDE.md`** — when to use a singleton vs a regular component vs a `Manager` service. The named-entity convention. Archetype implications (systems iterating `C_X` always hit exactly one entity if `X` is singleton). Save/load implications.

## Acceptance

1. `IRECS::singleton<T>()` exists, returns a stable reference across calls.
2. Modifier framework's globals access uses the new API; modifier-framework tests pass unchanged.
3. Lua test script creates a singleton component, reads/writes it.
4. `engine/entity/CLAUDE.md` documents the API + conventions.
5. Builds clean on `linux-debug` and `macos-debug`.

## Critical files

- `engine/entity/include/irreden/ir_entity.hpp` — new API entry points
- `engine/entity/src/ir_entity.cpp:69` — `getEntity` name lookup (extends naturally)
- `engine/entity/include/irreden/entity/entity_manager.hpp` — singleton-ID cache field
- `engine/prefabs/irreden/common/modifier.hpp:185-191` — retrofit site
- `engine/script/src/lua_script.cpp` — Lua binding
- `engine/entity/CLAUDE.md` — doc update

## Out of scope

- Singleton-component save/load integration — depends on #199 (save/load framework). Retrofit via #199 follow-up.
- Compile-time enforcement that a component is "singleton only" (a `is_singleton<T>` trait). Useful but not blocking — runtime convention is sufficient for v1.
- Conversion of `RenderManager`'s designated entities (`m_mainCanvas`, `m_camera`, etc.) to use the new API. They're already cached by the manager; refactoring them is a separate cleanup task.

## Notes for the worker

- The lazy-init pattern in `modifier.hpp:185-191` is the reference shape. The new API is its public form.
- Naming: `IRECS::singleton<T>()` is the working name. Alternatives include `IREntity::singleton<T>()` or `IREntity::global<T>()`. Worker can adjust if the existing namespace conventions point a different direction.
- Lua binding shape: the existing pattern in `lua_script.cpp` uses `IRECS.addComponent(entity, type, table)` / `IRECS.getComponent(entity, type)`. A natural extension is `IRECS.singleton(type) -> table`. Verify against the existing component-reflection helpers.
- Don't over-engineer. The lazy-init pattern is ~30 lines of real code; the API is ~10 lines of public surface. Most of the work is the Lua binding and the doc.
- Please file an additional follow up ticket to evauluate and migrate all existing systems/components that might benefit

jakildev:
## Summary

Generalize the modifier framework's hand-rolled lazy-init singleton pattern (`detail::globalsEntityId()` in `engine/prefabs/irreden/common/modifier.hpp`) into a public `IREntity::singleton<T>()` API and retrofit `C_GlobalModifiers` to validate the design.

Work in progress.

Closes #646

### DIFF
--- a/engine/entity/CLAUDE.md
+++ b/engine/entity/CLAUDE.md
@@ -120,6 +120,85 @@ to remember a manual cleanup step before `destroyEntity`. Per-component
 cleanup belongs in the component's `onDestroy()` member, not in a hook
 — see `engine/prefabs/CLAUDE.md` "Documented exceptions".
 
+## Singleton components
+
+A "singleton component" is a component for which exactly one record
+exists per world — framework-level globals, per-world settings, scratch
+state. The canonical example is the modifier framework's
+`C_GlobalModifiers` (one `"modifierGlobals"` entity per world); the
+sim-clock substrate (#200) and future per-world game-rules holders will
+adopt the same shape.
+
+`IREntity::singleton<T>()` is the public API for this pattern. Lazy-init
+on first call (creates an entity with default-constructed `T` and caches
+the id by `ComponentId`); subsequent calls return the same reference at
+the cost of one hash-map lookup. Singleton entities are normal ECS
+entities and participate in archetype iteration — a `forEachComponent<T>`
+that matches `T` will see the singleton row.
+
+```cpp
+// Typed C++ entry points (engine/entity/include/irreden/ir_entity.hpp):
+template <typename C> IREntity::EntityId singletonEntity();        // lazy-create
+template <typename C> IREntity::EntityId singletonEntityOrNull();  // no-create
+template <typename C> C& singleton();                              // lazy-create + ref
+template <typename C> C* singletonOrNull();                        // no-create + ptr
+```
+
+Lua: `IREntity.singleton(componentDef) -> LuaEntity`. Works for
+codegen'd-as-C++ components and runtime-registered Lua-defined
+components — both share the same `ComponentId` space and the same
+cache. Pass the returned `LuaEntity` to the standard
+`IREntity.getLuaComponent` / `IREntity.setLuaField` accessors.
+
+```lua
+local C_GameRules = IRComponent.register("GameRules", { score = 0 })
+local entity = IREntity.singleton(C_GameRules)         -- lazy-creates first time
+IREntity.setLuaField(entity, C_GameRules, scoreIdx, 1) -- normal field accessor
+```
+
+### Conventions
+
+- **Default-constructible.** The typed `singleton<T>()` path calls
+  `createEntity(T{})`. If `T` cannot be default-constructed, expose a
+  feature-specific factory wrapping `singletonEntity<T>` so callers can
+  seed the row with explicit values.
+- **Lazy validation.** The cache is keyed by `ComponentId` and validated
+  against `entityExists` on every lookup. If a singleton entity is
+  destroyed externally (manual `destroyEntity` call, end-of-world
+  `destroyAllEntities` reset), the next access lazy-recreates (typed
+  path) or returns `kNullEntity` / `nullptr` (or-null path).
+- **`destroyAllEntities` resets the cache.** Tests that tear down and
+  rebuild the world between cases get a fresh singleton on the first
+  post-reset access.
+- **Naming is optional.** The API does NOT auto-name the entity; callers
+  who want diagnostic visibility can `setName(singletonEntity<T>(),
+  "myThing")`. The modifier framework names its globals
+  `"modifierGlobals"` for the existing tooling that scans by name.
+- **Use the API for what it's for.** Singleton-shaped *components* — yes.
+  Things that should live on a manager (graphics device handles,
+  RenderManager fields) — no; manager fields are still the right
+  shape for device-level state. Rule of thumb: if the data is naturally
+  an ECS column (composable, iterable, save/load-friendly), use a
+  singleton component; if it's a pointer to an external resource that
+  the manager already owns, leave it on the manager.
+
+### Archetype implications
+
+A singleton entity is a normal entity in the archetype graph. Adding or
+removing components on it moves it between archetype nodes the same way
+any other entity does. Iteration order across nodes is implementation-
+defined, so don't assume the singleton appears first or last in a
+`forEachComponent` walk — query by `singletonEntity<T>()` when you need
+the specific row.
+
+### Save/load implications
+
+Per-component save/load (#199) treats singleton entities like any other:
+the entity id round-trips and the cache rebuilds on first access against
+the loaded entity. Workers retrofitting #199 into the singleton API should
+verify the cache invalidates on load (typically a `destroyAllEntities`
+precedes the load, which already clears the cache).
+
 ## Position components are automatic
 
 `createEntity(...)` always adds `C_PositionGlobal3D` and
--- a/engine/entity/include/irreden/entity/entity_manager.hpp
+++ b/engine/entity/include/irreden/entity/entity_manager.hpp
@@ -371,6 +371,41 @@ class EntityManager {
     PreDestroyHookId registerPreDestroyHook(PreDestroyHook hook);
     void unregisterPreDestroyHook(PreDestroyHookId id);
 
+    /// Singleton-component support. One entity per component type, cached
+    /// by `ComponentId`. The cache is lazily validated: if the cached
+    /// entity has been destroyed (e.g. by `destroyAllEntities`), the next
+    /// lookup discards the stale entry and either lazy-creates (typed
+    /// path) or returns `kNullEntity` (or-null path).
+    ///
+    /// Typed C++ entry point: `singleton<T>()` (templated). Uses
+    /// `createEntity(T{})` so `T` must be default-constructible.
+    template <typename Component> EntityId getOrCreateSingleton() {
+        IR_PROFILE_FUNCTION(IR_PROFILER_COLOR_ENTITY_OPS);
+        const ComponentId componentType = getComponentType<Component>();
+        auto it = m_singletonEntityByComponent.find(componentType);
+        if (it != m_singletonEntityByComponent.end() && entityExists(it->second)) {
+            return it->second;
+        }
+        if (it != m_singletonEntityByComponent.end()) {
+            m_singletonEntityByComponent.erase(it);
+        }
+        EntityId entity = createEntity(Component{});
+        m_singletonEntityByComponent[componentType] = entity;
+        return entity;
+    }
+
+    /// Untyped entry point used by Lua-defined components — the component
+    /// is attached via the dynamic-add path so this works uniformly for
+    /// both `registerComponent<T>` and `registerComponentDynamic` types.
+    /// For typed C++ components, prefer `getOrCreateSingleton<T>()`.
+    EntityId getOrCreateSingletonByComponentId(ComponentId componentType);
+
+    /// No-create variant. Returns the cached singleton entity for this
+    /// component type if previously created (and still alive), else
+    /// `kNullEntity`. Useful for query sites that should not implicitly
+    /// instantiate the singleton when called before initial setup.
+    EntityId getSingletonByComponentIdOrNull(ComponentId componentType);
+
   private:
     std::queue<EntityId> m_entityPool;
     std::unordered_map<EntityId, EntityRecord> m_entityIndex;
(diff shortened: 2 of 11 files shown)


jakildev (commented):
## Review — T-162: engine/entity: ECS singleton-component infrastructure

**Verdict:** approve (pending Opus pass — see escalation)

> **Note on "Work in progress":** The PR body says "Work in progress" but the diff appears complete: EntityManager extension, public ir_entity.hpp API, Lua binding, C++ and Lua tests, and the modifier.hpp retrofit are all present. Flagging the discrepancy for the human — if the author has uncommitted follow-up work in mind, add `fleet:wip` before merge; otherwise the WIP note can be removed.

### What this PR does

Adds `IREntity::singleton<T>()` as the canonical API for "one entity per component type" patterns. Implements a lazy-init + hash-map cache in `EntityManager` keyed by `ComponentId`. Provides typed C++ entry points (`singletonEntity<T>`, `singletonEntityOrNull<T>`, `singleton<T>`, `singletonOrNull<T>`) and an untyped dynamic path (`getOrCreateSingletonByComponentId`) for Lua-registered components. Retrofits `C_GlobalModifiers` to use the new API as the design-validation gate. Clears the cache in `destroyAllEntities` for correct test isolation.

### ECS invariants

- `getOrCreateSingleton<T>()` calls `createEntity(Component{})` — a one-time structural change at initialization, not in a tick. Not a mid-tick archetype mutation. Clean.
- `globalsEntityId()` now calls `singletonEntityOrNull<C_GlobalModifiers>()` (hash map lookup via `getSingletonByComponentIdOrNull`). Runs in `beginTick` of `MODIFIER_RESOLVE_GLOBAL` — one map lookup per frame, no allocation. Clean.
- No per-entity `getComponent` introduced in any tick path. Clean.

### Cache correctness

Stale-entry eviction: if the cached entity has been externally destroyed, `entityExists(it->second)` returns false, the stale entry is erased, and a new entity is created (typed path) or `kNullEntity` is returned (or-null path). Correct.

`destroyAllEntities` calls `m_singletonEntityByComponent.clear()` after the entity loop — pre-destroy hooks fire during the loop and can still reach `singletonEntityOrNull` (e.g., `removeBySource` calls `forEachComponent<C_GlobalModifiers>`, not the singleton accessor directly), the cache is cleared last. Sequencing is safe.

### modifier.hpp retrofit

`registerResolverPipeline` now calls `singletonEntity<C_GlobalModifiers>()` instead of the hand-rolled `static IREntity::EntityId id`. `globalsEntityId()` returns `singletonEntityOrNull<C_GlobalModifiers>()` (no-create), which correctly returns `kNullEntity` before `registerResolverPipeline` runs so `beginTick` can no-op cleanly. `~ModifierLuaGlobalTest` no longer needs to manually reset the static — the EntityManager destructor clears the cache. The retrofit is correct.

### Lua binding

`IREntity.singleton(componentDef)` dispatches to `getOrCreateSingletonByComponentId` via `componentDef["componentId"]`. Both C++-registered and Lua-dynamic components share the same `ComponentId` space and the same singleton cache on `EntityManager`. Lua tests verify caching and C++/Lua interop.

### Test coverage

Six C++ tests covering: reference identity, entity ID caching, mutation persistence, or-null before create, or-null after create, and lazy-recreate after external destroy. Two Lua tests covering caching and C++/Lua interop. Solid.

### Nits

- `test/ecs/entity_manager_test.cpp` — no test for `destroyAllEntities` clearing the singleton cache. The documented invariant ("destroyAllEntities resets the cache") is correct in the implementation (explicit `m_singletonEntityByComponent.clear()`) but lacks a direct test. The `SingletonLazyRecreateAfterExternalDestroy` test covers the per-entity eviction path, which is related but not the same as the bulk reset. A `SingletonCacheResetAfterDestroyAllEntities` test would close the gap.

- `engine/entity/include/irreden/ir_entity.hpp` includes reformatting of `collectEntitiesSimple`, `removeComponentsSimple`, `forEachComponent`, and `setComponentDeferred` that is unrelated to the singleton feature. Not wrong, but adds diff noise to a scoped feature addition.

### Praise

- `singletonEntityOrNull` returning `kNullEntity` before the singleton is created (rather than creating on first lookup) is the right design for `globalsEntityId()` — systems that run before `registerResolverPipeline` get a clean null rather than accidentally creating an unnamed globals entity.
- Lua and C++ singletons sharing the same cache is the correct choice — a component registered on the C++ side and looked up from Lua resolves to the same entity without any bridging logic.
- The clear at end of `destroyAllEntities` (not before the entity loop) is correct timing — hooks can still call `singletonEntityOrNull` during destruction and get valid results up to the last entity.

### Escalation

Opus recheck required: touches `engine/entity/` (EntityManager + public `ir_entity.hpp` surface), introduces new lifetime/ownership semantics (lazy cache eviction on external entity destroy, cache invalidation in `destroyAllEntities`), and establishes a new public API surface adopted by the modifier framework across multiple modules.

jakildev (commented):
## Opus recheck — T-162: engine/entity: ECS singleton-component infrastructure

Sonnet flagged this for Opus recheck on the grounds that the PR touches `engine/entity/` (EntityManager + public `ir_entity.hpp`), introduces new lifetime semantics, and establishes a public API adopted by the modifier framework. I focused this pass on the deeper invariants Sonnet could not fully confirm from the diff alone: cache eviction under re-entrant tear-down, reference/pointer stability of the `singleton<T>()` accessors, the hot-path cost of demoting `globalsEntityId()` from a static reference to a hash-map lookup, and confirming the `ModifierLuaGlobalTest` retrofit actually closes the old reset hole.

### Confirmed (Sonnet's analysis is correct)

- **No per-entity `getComponent` introduced in any tick path.** The only system change is `system_modifier_resolve_global.hpp::beginTick`, which calls `globalsEntityId()` exactly once per frame and caches the pointer in `currentGlobalModifiersPtr()`. The per-entity tick reads the cached pointer — zero extra lookups in the hot path. The demotion of `globalsEntityId()` from a static-reference accessor (O(1) trivial) to a `singletonEntityOrNull<C_GlobalModifiers>()` lookup (two `unordered_map::find`s) costs once-per-frame, not once-per-entity. Negligible.
- **Cache eviction is correct.** `getOrCreateSingleton<T>()` and `getOrCreateSingletonByComponentId` both `find` → `entityExists` → `erase` stale → `createEntity` → re-insert. The iterator is discarded after `erase`; the subsequent `operator[]` performs a fresh lookup, so the iterator-invalidation contract is honored. `getSingletonByComponentIdOrNull` evicts stale entries on read, so subsequent or-null queries see `kNullEntity` without re-paying the `entityExists` cost.
- **`destroyAllEntities` cache clear sequencing is intentional.** The cache `.clear()` runs *after* the destroy loop, not before — Sonnet noted this is correct so that pre-destroy hooks calling `singletonEntityOrNull<T>()` mid-tear-down still resolve to valid entities until each is reached. Confirmed; the only documented in-tree hook (`removeBySource`) walks `forEachComponent`, not the singleton accessor, but the contract is the safer choice for future hooks.
- **`ModifierLuaGlobalTest` retrofit closes the static-state hole.** The previous fixture leaked a static `globalsEntityId` reference across test instances and had a `~ModifierLuaGlobalTest()` destructor to manually reset it. The retrofit moves singleton state onto `EntityManager`, so when the `IRWorld` member of the parent `ModifierLuaTest` fixture goes out of scope, the cache dies with it. The explicit reset is gone because it's no longer needed — not because someone forgot it. Verified against the PR branch.
- **No `m_singletonEntityByComponent` race.** `EntityManager` is single-threaded by the wider engine contract; the new map inherits the same constraint as `m_entityIndex` etc.
- **Naming.** `m_singletonEntityByComponent` matches the existing `m_entityPool` / `m_entityIndex` / `m_pureComponentTypes` style.

### Opus-only finding (nit)

**Pre-destroy hook calling `singletonEntity<T>` mid-`destroyAllEntities` leaks a singleton entity.** The destroy loop iterates a *snapshot* of `m_entityIndex` (taken before the loop). After the snapshot is captured but before the `.clear()` at the end, a pre-destroy hook that calls the lazy-create variant `singletonEntity<T>()` for a singleton already destroyed earlier in the loop will:

1. Hit the stale cache entry, fail `entityExists`, erase the entry.
2. Call `createEntity(T{})` for a fresh singleton entity.
3. Insert it into the cache.
4. Return that new entity id.

The newly-minted singleton entity is NOT in the destroy-loop snapshot, so it survives the call. The `m_singletonEntityByComponent.clear()` at end-of-loop then removes the cache reference, but the entity itself remains in `m_entityIndex` — a singleton-shaped "ghost" that nobody can find via `singletonEntity<T>()` (cache is empty) but is still iterated by `forEachComponent<T>`.

This is not blocking — the only in-tree pre-destroy hook (`removeBySource`) doesn't create entities, and `engine/entity/CLAUDE.md` already advises hooks against reentrant structural changes. But the singleton API now adds a *second* shape of reentrant structural change (the lazy-create lookup) that's easy to miss in hook authoring. Worth one of:

- a CLAUDE.md note under "Pre-destroy hooks" forbidding `singletonEntity<T>` (lazy-create) inside hooks during `destroyAllEntities`; `singletonEntityOrNull<T>` is the safe form, or
- defensively gating the lazy-create path on a `bool m_destroyAllInFlight` flag and returning `kNullEntity` while teardown is running.

The CLAUDE.md note is the cheaper, more honest fix.

### Sonnet's nits (I agree)

- **Missing direct test for `destroyAllEntities` clearing the singleton cache.** The lazy-recreate-after-external-destroy test exercises the per-entry eviction path, not the bulk reset. A `SingletonCacheResetAfterDestroyAllEntities` test would close the gap and give the documented invariant a regression guard.
- **`ir_entity.hpp` reformatting of `collectEntitiesSimple` / `removeComponentsSimple` / `forEachComponent` / `setComponentDeferred`** is unrelated to the feature and adds diff noise. Not wrong, but ideally split into a separate formatting commit.
- **PR body says "Work in progress."** Diff is complete — please remove the WIP note or add `fleet:wip` before merging.

### Praise (additive)

- The "no-create" / "lazy-create" split is the right shape. `singletonEntityOrNull<T>` lets `globalsEntityId()` no-op cleanly when called before `registerResolverPipeline()`, and the lazy-create path is reserved for sites that genuinely want first-touch construction. Many singleton-pattern implementations collapse this into a single function and force callers into ugly "is this initialized yet?" branches.
- Sharing one `ComponentId` space (and one cache) between C++ and Lua-registered components is the right call. The Lua test that round-trips a Lua-side write through the C++ `getOrCreateSingletonByComponentId` accessor proves this works end-to-end without bridging logic.
- The `addComponentDynamic` path in `getOrCreateSingletonByComponentId` (vs `createEntity(Component{})` for the typed path) correctly handles components with `= delete`d default ctors via the Lua-typed dynamic-add — they get an `appendDefaultRow()` instead of a default construct. This is subtle and worth keeping in mind for future singleton consumers.

Cleared for merge once the human resolves the WIP note. Nits are author-judgment.

Robot Reviewed by Claude Opus 4.7 (role-opus-reviewer)

PR #651, opened 2026-05-12
T-159: GPU particles Phase 2 — batch CPU-side spawns into one subData/frame

### LINKED ISSUE #632: Phase 2 GPU particles: batch CPU-side spawns on Metal to avoid per-spawn subData orphan churn
## Context

Filed as a Phase-2 follow-up to PR #614 / T-139 (Phase 1 GPU particle system foundation). See the Opus recheck at 2026-05-11T05:38:06Z, N2 — the recheck approved Phase 1 as-is but flagged this for the radar before Phase 2's continuous-emitter Lua spawn API lands.

## Problem

Metal's `Buffer::subData` implementation orphans the whole MTL buffer per call: it allocates a fresh `Shared` buffer, full-`memcpy`'s the live contents, patches the new bytes at `offset`, then defers release of the old buffer through the `present()` boundary. See `engine/render/metal_buffer.cpp:48-83` and `engine/render/metal_render_impl.cpp:251-266`.

In Phase 1 the cost is latent because the demo seeds once at startup (1024 spawns × 32 B/each ≈ 128 MB of allocation + memcpy + bookkeeping at startup). Phase 2's continuous-emitter API will issue many spawns per frame; at 100 spawns/frame on a 4096-slot pool that is ~12.8 MB/frame of pure orphan overhead before any actual particle work.

OpenGL is unaffected: `glNamedBufferSubData` is a true partial patch (`O(bytes changed)`).

## Suggested approach

Either:

a. **Frame-scoped staging vector on the CPU side.** `C_GPUParticlePool` accumulates spawns into a per-frame staging buffer and issues exactly one `subData` per frame (covering the contiguous range of touched slots, or the whole buffer if writes are sparse). Simplest to implement; backend-agnostic.

b. **Metal `BufferImpl` "frame-coherent batch" mode.** Extend `metal_buffer.cpp` with a mode that defers the orphan to the next bind / encoder-boundary, coalescing multiple `subData` calls within a frame into a single orphan. More invasive but transparent to call sites.

Lua-side, the spawn API should accept either an array of particles or be implicitly batched by the binding so per-particle Lua → C++ crossings don't translate to per-particle `subData` calls.

## Acceptance criteria

- A continuous-emitter demo spawning N particles/frame at N ∈ {10, 100, 1000} reports steady-state `metal_buffer subData` allocation counts ≤ 1/frame regardless of N.
- Cross-backend visual parity (OpenGL vs Metal) under continuous emission for the same demo.
- `fleet-build --target IRGpuParticles` clean on both backends; per-frame frame time on Metal not regressed vs OpenGL for the same emission rate.

## Affected files

- `engine/prefabs/irreden/render/components/component_gpu_particle_pool.hpp` — staging vector or batch hook
- `engine/render/metal_buffer.cpp` — only if approach (b) is taken
- `creations/demos/gpu_particles/` — extend to a continuous-emitter mode for verification
- New: Phase 2's Lua spawn API surface

## Gotchas

- The frame-boundary correctness model for Phase 1 stands (`waitUntilCompleted()` at present, `SHADER_STORAGE` barrier on GL, per-encoder isolation on Metal). Batching changes only the upload cost shape, not the visibility chain.
- The CPU `lifetime_` mirror still does not track GPU decay — see the separate carry-over from the same recheck (silent-drop-after-4096-spawns). Batching alone does not solve that; both fixes are likely needed before Phase 2's continuous-emitter API can ship.

Suggested model: `[opus]` (Metal `BufferImpl` shape decision + cross-backend perf gate).
Suggested area: `engine/render`, `engine/prefabs/irreden/render`, `creations/demos/gpu_particles`.

— opus-worker-2, while addressing PR #614 nits

jakildev:
## Summary

- `C_GPUParticlePool::writeSlot` queues spawns into a `pendingIndices_` vector instead of issuing per-spawn `subData`.
- `UPDATE_GPU_PARTICLES` calls `pool.flushPendingSpawns()` once per frame ahead of the compute dispatch — one `subData` per contiguous run of touched slot indices. The common case (sequential indices from `reserveSlot()`'s linear-scan reclamation) collapses to exactly one `subData/frame` regardless of spawn count.
- The flush only uploads slots that were just written this frame — never untouched interior ranges, because the CPU mirror's position/lifetime is stale relative to the GPU update kernel.
- `gpu_particles` demo gains `--spawns-per-frame N` for the continuous-emitter verification path.

## Why this matters

Metal's `BufferImpl::subData` orphans the whole MTL buffer per call (newBuffer + 3× full-buffer memcpy + deferred release through `present()` — `engine/render/src/metal/metal_buffer.cpp:48-80`). At 100 spawns/frame on a 4096-slot pool that was ~12.8 MB/frame of pure orphan overhead. With this PR, one or two coalesced `subData` calls per frame replace the N per-spawn calls — Metal allocation churn drops by ~N×.

OpenGL's `glNamedBufferSubData` is a true partial patch (`O(bytes changed)`), so this is neutral there. The data flowing through the SSBO is byte-identical on both backends; what changes is timing and call shape, not content.

## Visual / correctness verification

- Demo runs cleanly on Metal (macOS, this host) in both seed-only mode and `--spawns-per-frame 100`. All 3 auto-screenshots captured each run.
- This is an internal CPU-side batching refactor — GPU compute kernels (`c_update_gpu_particles.glsl`, `c_render_gpu_particles_to_trixel.glsl`) are unchanged. The shader edit is a comment-only update describing the new spawn path. Per `engine/render/CLAUDE.md` "Verifying render changes" exception list, internal refactors with no runtime visual effect can skip `render-debug-loop` / `attach-screenshots`. Same data lands at the same slots; visual output should be byte-identical to Phase 1 in the seed-only baseline.

## Hot-path discipline

- `pendingIndices_.reserve(capacity_)` in the ctor — `push_back` never reallocates in steady state.
- `std::sort` and `std::unique` are in-place; `clear()` preserves vector capacity.
- For N=1000 spawns/frame, the flush is ~10 μs of CPU work (sort + walk) + 1–2 `subData` calls — well under 0.1% of the 16.67 ms frame budget.
- See the optimize report in the worker iteration log.

## Doc updates

- `.claude/rules/cpp-ecs.md` "Push at mutation time" rule now explicitly documents the deferred-flush variant for cases where per-mutation upload cost dominates (Metal `subData` orphans being the canonical case). The pattern is "queue + per-frame coalesced flush, never re-upload untouched slots".
- `c_update_gpu_particles.glsl` comment updated to describe the new spawn path.
- Component and system doc-blocks updated.

## Test plan

- [x] `fleet-build --target IRGpuParticles` clean (macOS / Metal).
- [x] `fleet-run IRGpuParticles --auto-screenshot 10` (seed-only): clean run, 3 screenshots.
- [x] `fleet-run IRGpuParticles --auto-screenshot 10 --spawns-per-frame 100`: clean run, 3 screenshots; continuous-emitter workload exercises the pending-list flush every frame.
- [ ] OpenGL smoke (Linux fleet via `fleet:needs-linux-smoke`).
- [ ] Reviewer cross-checks per-frame `subData` count via `setGpuTimingEnabled` if desired — design proof in the issue thread.

Closes #632.


jakildev (commented):
## Review — T-159: GPU particles Phase 2 — batch CPU-side spawns into one subData/frame

**Verdict:** approve

### ECS invariants

- `UPDATE_GPU_PARTICLES` tick receives `C_GPUParticlePool` as a direct parameter — no per-entity `getComponent`. `flushPendingSpawns()` runs inside that tick and operates only on `pool` (self-only helper, not a cross-entity reach). No new components added to the system template; no archetype mutations mid-tick. All clean.
- `createContinuousSpawnerSystem` in `main.cpp` also receives `C_GPUParticlePool` as a direct tick parameter. Only calls `pool.reserveSlot()` and `pool.writeSlot()` — both self-only. Slot not found → `kInvalidSlot` guard fires and returns. Clean.
- `pendingIndices_.push_back` in `writeSlot` runs in the spawner tick hot path. Pre-reserved to `capacity_` in the ctor (16 KiB for 4096 slots); `push_back` never reallocates in steady state — satisfies the no-allocation-in-hot-tick rule.

### Ownership / lifetime

- `buffer_.second` is a raw non-owning pointer; ownership lives in `buffer_.first` (ResourceId), freed in `onDestroy()`. Consistent with `C_TriangleCanvasTextures`, `C_TrixelFramebuffer`, and the other GPU-RAII exceptions. Clean.
- `pendingIndices_` capacity is preserved across frames via `clear()` (does not deallocate). No unbounded growth.
- CPU mirror `particles_` is a one-shot seed allocator: dead-slot scan reads `lifetime_ <= 0` from CPU-side state, which is stale once the GPU update kernel runs. The component doc explicitly calls this out — a slot is "taken" from the CPU's view until `clear()` is called. This is intentional Phase 2 design; continuous spawning is capacity-bounded for now (PR body and code comments both acknowledge it). No lifetime bug.

### Flush correctness

The run-length encoding in `flushPendingSpawns` is correct:
- Sort + unique → well-defined contiguous-run detection.
- The loop handles all interior run boundaries; the tail `subData` after the loop handles the last (or only) run — including the single-element case (loop body never fires; tail count = 1). Verified trace for [N], [N, N+1], and [N, M] (non-consecutive) cases.
- Only freshly-written slots are uploaded — the CPU mirror is stale relative to GPU-authored position/lifetime for in-flight particles, so the pending-list manifest correctly excludes untouched interior ranges. GPU-authored state is preserved.

### GPU ordering (Metal)

`flushPendingSpawns()` → bind SSBO → `dispatchCompute` → `memoryBarrier(SHADER_STORAGE)`. On Metal, `subData` orphans the old buffer and allocates a fresh one; the subsequent bind attaches the NEW buffer to the dispatch, so the compute shader sees only the freshly-uploaded data — no ordering hazard. The barrier after dispatch guards the next reader. Sequence is safe on both backends.

### Render / shader

`c_update_gpu_particles.glsl` — comment-only change; no binding indices, struct fields, or dispatch logic changed. CPU/GPU struct layout (`FrameDataGpuParticles`) is untouched.

### doc update

`.claude/rules/cpp-ecs.md` "No dirty flags on components" now explicitly documents the deferred-flush variant with `C_GPUParticlePool` as the canonical example. The rule text and the implementation landed in the same PR — no drift possible.

### Nits

- `engine/render/src/shaders/metal/c_update_gpu_particles.metal:4-6` — Metal shader comment still reads "T-139 Phase 1" only. The GLSL counterpart was updated to describe the Phase 2 pending-list spawn path; the Metal comment now lags behind. Since the Metal file already says "Mirrors c_update_gpu_particles.glsl", the reader can cross-reference, but updating the heading to "T-139 Phase 1 / T-159 Phase 2 batching — see .glsl counterpart for spawn-path design note" would keep both files current.

### Praise

- The CPU-mirror-only-for-bookkeeping split (GPU owns per-frame particle state; CPU owns the dead-slot allocator + pending-write manifest) is clean and well-documented. The invariant "CPU never re-uploads stale interior ranges" is the right constraint to make both the optimization and the correctness argument load-bearing.
- The tail-run `subData` after the loop is correct, and the single-element path is handled without special-casing. Getting that right in the first take is non-trivial.
- Reserving `pendingIndices_` to `capacity_` in the ctor proactively satisfies the no-allocation-in-hot-tick rule — no follow-up needed.

### Test plan

- [x] `fleet-build --target IRGpuParticles` clean on macOS/Metal (per PR body)
- [x] `fleet-run IRGpuParticles --auto-screenshot 10` seed-only: 3 screenshots, no visual regression (per PR body)
- [x] `fleet-run IRGpuParticles --auto-screenshot 10 --spawns-per-frame 100`: 3 screenshots, continuous-emitter flush exercised every frame (per PR body)
- [ ] Linux/OpenGL smoke (fleet:needs-linux-smoke)

Opus recheck required: touches engine/prefabs/irreden/render/ and a compute shader SSBO path — Metal subData-to-dispatch ordering and GPU buffer lifetime warrant an Opus pass before merge.

jakildev:
Addressed nit: updated `engine/render/src/shaders/metal/c_update_gpu_particles.metal` heading to reference T-139 Phase 1 / T-159 Phase 2 batching and cross-reference the .glsl counterpart's design note, so the two shader files stay in sync.

jakildev (commented):
## Opus recheck — T-159: GPU particles Phase 2 — batch CPU-side spawns into one subData/frame

**Verdict:** approve

Sonnet's review covered ECS invariants, hot-path allocation, ownership, flush correctness, and the high-level Metal subData/dispatch sequencing. On the items Sonnet flagged for Opus recheck — **Metal subData→dispatch ordering and GPU buffer lifetime** — I re-walked the backend code and confirm the design is correct on both.

### Metal subData→dispatch ordering: confirmed safe

Sonnet's claim was that on Metal each `subData` orphans, the new buffer is swapped into the bindings, and the subsequent `bindBase` → `dispatchCompute` reads the freshly-orphaned buffer. Walked `engine/render/src/metal/metal_buffer.cpp:48-80` and `engine/render/src/metal/metal_runtime.cpp:303-318, 320-335` to verify the load-bearing steps:

1. `MetalBufferImpl::subData` allocates a fresh `MTL::Buffer` of the same size, then `memcpy`s **prefix [0, offset)** from the old buffer, **new region [offset, offset+size)** from the caller's `data` pointer, and **tail [offset+size, m_size)** from the old buffer. The new buffer is therefore a full snapshot of pre-write state + the patch.
2. `m_buffer` swaps to the new pointer, `replaceMetalBufferInBindings(oldBuffer, m_buffer)` rewrites both `uniformBuffers_` and `shaderStorageBuffers_` so any prior `bindBase` automatically retargets the new buffer, and `deferReleaseMetalBuffer(oldBuffer)` pushes the old one onto `pendingReleaseBuffers_`.
3. `present()` (`engine/render/src/metal/metal_render_impl.cpp:251-266`) commits the command buffer, **`waitUntilCompleted()`**, and only then calls `releaseDeferredMetalBuffers()`. The old MTL buffer cannot be freed while the GPU is still consuming it.

**Multi-subData composition in `flushPendingSpawns()`** is therefore correct: when the loop issues a second subData for a non-contiguous run, the prefix copy reads from the *post-first-subData* buffer (which already contains the [10..15] update), copies it into the new orphan, then patches [50..55] on top. The dispatch at the end of the tick sees both updates because the bind table tracks the latest orphan. No ordering hazard.

The `screenshotPath` path also flushes the deferred-release queue mid-frame at `metal_render_impl.cpp:308`, so the orphan churn doesn't leak across the synchronous flush either.

### GPU buffer lifetime: confirmed safe

`buffer_.first` is the `ResourceId` (freed in `onDestroy()`); `buffer_.second` is a stable `Buffer*` wrapper whose internal `BufferImpl::m_buffer` (the MTL::Buffer*) is the only thing that swaps on subData. The wrapper pointer stored on the component never dangles, and the orphaned MTL::Buffers are kept alive across the in-flight frame by the defer-release queue. No use-after-free window, no per-frame double-free.

### Flush correctness — re-traced the run-length encoder

Sonnet asserted correctness for `[N]`, `[N, N+1]`, and `[N, M]`. I traced two more edge cases to convince myself:

- **All-non-contiguous `[N, M, P]`** (N < M-1, M < P-1): i=1 emits run `[N,N]`, runStart=M; i=2 emits run `[M,M]`, runStart=P; tail emits `[P,P]`. Three subData calls, each correct.
- **Sequential `[N..N+k-1]`** for arbitrary k: every iteration hits `idx == runEnd + 1u` and `continue`s; the tail emits one subData of count `k`. One subData call.
- **Duplicates**: `std::sort` + `std::unique` collapses them before run-length scan; correct.

The tail subData is unconditional after the loop, which is safe because the empty-list early-return guarantees `pendingIndices_.size() >= 1` at the tail (after `unique` only collapses equal-valued runs, never empties a non-empty input).

`runEnd + 1u` is uint32_t arithmetic. Slot indices are bounded by `capacity_` (4096 by default), so overflow is not reachable.

### Hot-path allocation: confirmed bounded

`pendingIndices_.reserve(capacity_)` in the ctor + `clear()` (which preserves capacity) means `push_back` in `writeSlot` cannot reallocate as long as a frame's writes stay within `capacity_` unique slots. `reserveSlot()` reads CPU `lifetime_` (which doesn't decay during the frame), so a CPU-mediated spawner caller can claim at most `capacity_` slots per frame — the reserve is exact for the documented contract. A caller that bypasses `reserveSlot` and writes the same slot index twice is also fine: the worst case for unique indices is bounded by `capacity_`, and duplicates collapse in `unique()` rather than growing the upload.

### Things Sonnet covered that I confirmed without re-checking

- ECS invariants (system template parameters, no per-entity getComponent, no archetype mutation mid-tick).
- Naming conventions, `C_` prefix, trailing-`_` on public members.
- `pendingIndices_` capacity preserved across frames via `clear()`.
- CPU-mirror-is-stale-relative-to-GPU invariant.
- Doc-rule landing in the same PR as the implementation.
- Shader change is comment-only; binding indices and struct layouts unchanged.

### Nits

None new from this pass — Sonnet's nit on the Metal shader heading at `engine/render/src/shaders/metal/c_update_gpu_particles.metal:4-6` already covers the only minor item I'd have surfaced. Picking that up alongside merge is fine.

### Praise

- The defer-release-on-`present()` interaction with `flushPendingSpawns()` is the kind of cross-module invariant that's easy to break silently; the design got it right without needing extra synchronization in the new code path. The fact that `replaceMetalBufferInBindings` makes the binding retarget implicit is what makes the "subData → bind → dispatch" sequence robust on Metal.
- Reserving `pendingIndices_` to `capacity_` in the ctor matches the actual upper bound rather than a guessed one — bounded by the same allocator that hands out slots.

Robot Reviewed by Claude Opus 4.7 (role-opus-reviewer)

jakildev (commented):
## Re-review — T-159: GPU particles Phase 2 — batch CPU-side spawns into one subData/frame

**Verdict:** approve (re-confirmed)

Picked up via `fleet:changes-made` after the author pushed a nit fix.

### Nit resolution

Sonnet flagged `engine/render/src/shaders/metal/c_update_gpu_particles.metal:4-6` — the Metal heading said "T-139 Phase 1" while the GLSL counterpart described the pending-list spawn path added in T-159. Fixed at head: the Metal heading now reads "T-139 Phase 1 / T-159 Phase 2 batching" and cross-references the .glsl counterpart for the design note. Both backends now describe the same dispatch model. The fix is exactly the suggested form.

All substantive findings from the Sonnet + Opus review cycle are unchanged and still hold.

Opus recheck not required — Opus passed in the prior cycle; this commit is a shader comment update only.

PR #638, opened 2026-05-12
T-156: Migrate trixel-canvas content systems to member-on-System<N>

jakildev:
## Summary
- Migrates 5 systems across 4 files from explicit `Params + setSystemParams` to `member-on-System<N>` via `registerSystem` — PR 3 of the 5-PR plan from issue #580.
- Systems: `VOXEL_TO_TRIXEL_STAGE_1`, `VOXEL_TO_TRIXEL_STAGE_2`, `SHAPES_TO_TRIXEL`, `TEXT_TO_TRIXEL`, `FOG_TO_TRIXEL`.
- Each uses the GPU-resource two-step `create()` shape: `createNamedResource` calls, then `registerSystem`, then `getSystemParams` to populate cached pointers. Behavior and per-tick cost are identical to the original form.

## Test plan
- [x] `fleet-build --target IRShapeDebug` — clean
- [x] `fleet-build --target IrredenEngineTest` — clean
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — 6/6 shots completed, no crash

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #652, opened 2026-05-12
T-161: defer C_CanvasFogOfWar dirty-flag → per-region subImage2D migration

### LINKED ISSUE #645: C_CanvasFogOfWar: evaluate per-write subData vs full-buffer dirty-flag upload
## Context

Follow-up from PR #638 Opus recheck (nit N3). Pre-existing, not introduced by the T-156 mechanical migration.

`engine/prefabs/irreden/render/systems/system_fog_to_trixel.hpp:61-83` uses a frame-scoped `dirty_` flag on `C_CanvasFogOfWar`: when set, the entire 256×256 cpuBuffer is re-uploaded to the GPU via `subImage2D`. Per-frame stores are `~256K → ~64K` thanks to a 4× channel-stride trick, but the upload is still all-or-nothing.

The reviewer suggested migrating to per-write `subData` semantics so only the actually-mutated tiles upload each frame. That avoids the worst case (one cell touched → full buffer re-upload).

## Open question (decide before implementing)

This is a design question, not a rule violation. Two competing concerns:

- **Per-write subData** can be cheaper when only a few cells change per frame (fog reveal, line-of-sight updates).
- **Whole-buffer subImage2D** is cheaper when many cells change in a frame (radius reveals, large diffs) because per-write cmds have driver overhead.

The right answer depends on the actual workload mix in fog-of-war demos (radius reveals dominate today; future RTS-style LOS workload may differ).

## Proposed approach

1. Profile fog-of-war upload cost in `IRFogOfWar` (or whichever demo exercises C_CanvasFogOfWar) with the current dirty-flag path.
2. Compare against a tile-dirty-list variant (`std::vector<ivec2> dirtyTiles_` accumulated per frame on `C_CanvasFogOfWar::setCell`; flushed at FOG_TO_TRIXEL tick with per-tile `subData` calls or a region-merged single `subImage2D`).
3. Pick whichever is faster on the realistic workload. Document the decision in `engine/prefabs/irreden/render/CLAUDE.md`.

## Acceptance

- (1) Measurement: per-frame fog upload cost (CPU + GPU) under three workloads — single-cell update, radius reveal, full clear.
- (2) Code change implementing the chosen strategy (or a note in the design doc deferring the change with rationale).
- (3) `IRFogOfWar` (or equivalent demo) renders identically.

## Suggested tagging

- Model: opus (perf-sensitive, requires measurement + design decision).
- Area: engine/prefabs/irreden/render.

Originating review: https://github.com/jakildev/IrredenEngine/pull/638#pullrequestreview

Note: the originating review cited `.claude/rules/cpp-ecs.md "No dirty flags on components"` and PR #614 as the rule source — that rule does not exist in the current `.claude/rules/` files, and #614 is the GPU particle PR. Treating this as a perf/design question rather than a rule violation.

jakildev:
## Summary
- T-161 (#645) asked: should `C_CanvasFogOfWar` drop its dirty flag in favor of per-write `subData` semantics (PR #638 follow-up)?
- Conclusion: **defer**. Lands `docs/design/fog-of-war-upload-strategy.md` capturing the three-strategy comparison, workload reality, trigger conditions for revisiting, and a mechanical Strategy C migration sketch for when the trigger fires.
- Updates `.claude/rules/cpp-ecs.md` "Live deviations" entry and the component header to reference the design doc.
- **No code change.** This is a design-doc deferral per the issue's acceptance criterion #2 ("code change OR design doc deferring with rationale").

## Why defer

The cited rule already lists `C_CanvasFogOfWar` as the single allowed exception to "no dirty flags on components" (CPU-authored, GPU-read-only, full-texture upload). T-161 verified the exception still applies:

- All four current call sites are **one-shot `revealRadius` at init** (shape_debug, perf_grid, lua_perf_grid, lighting demos).
- Steady-state per-frame upload cost is **identically zero** across every current consumer — the `!dirty_` early-return short-circuits to no work after the first post-init frame.
- Total per-session upload cost: two ~256 KiB uploads during scene setup. Well under any reasonable hotspot threshold (≪0.02% of available memory bandwidth at 60 Hz, even if we *did* push every frame).

The design doc names the trigger conditions for revisiting (high-frequency per-frame mutation, sub-frame upload contention, or texture-extent scope expansion) and lays out Strategy C (per-region bbox `subImage2D`) as the mechanical migration when any of them fires.

## Why measurement is deferred

Acceptance #1 calls for "per-frame fog upload cost under three workloads." Without a fog-stress demo (which would exercise mutation paths none of the four current consumers touch), principled measurement is infeasible — fabricated benchmarks would be worse than no benchmark. The design doc's "Measurement plan (deferred)" section specifies the demo + harness that should be built when one of the trigger conditions actually emerges.

## Test plan
- [x] `fleet-build --target IRShapeDebug` clean on macOS/Metal
- [ ] Linux/OpenGL smoke (engine/prefabs/irreden/render/ touched, but only comment-edit; sanity check)
- [x] No code change → no rendering / behavior change → no screenshot diff needed (component header edit is comment-only; render/render-debug-loop skipped per the documented "purely docs" exception)

## Notes for reviewer
- The component header edit (`engine/prefabs/irreden/render/components/component_canvas_fog_of_war.hpp:17-30`) is comment-only — three new lines pointing at the design doc, no functional change.
- The cpp-ecs.md edit appends a deferral-note clause to the existing live-deviation bullet — same rule semantics, new pointer.
- The design doc itself is the substantive deliverable; reviewers should evaluate whether the deferral reasoning is sound and whether the Strategy C migration sketch is mechanical enough to ship without further design work when the trigger fires.

Closes #645

### DIFF
--- a/.claude/rules/cpp-ecs.md
+++ b/.claude/rules/cpp-ecs.md
@@ -95,7 +95,7 @@ Allowed exception: the destination resource is **strictly CPU-authored, GPU-read
 
 ### Live deviations
 
-- `engine/prefabs/irreden/render/components/component_canvas_fog_of_war.hpp` — `C_CanvasFogOfWar::dirty_` and `allUnexplored_` gate the per-frame `subImage2D` upload of the 256² fog texture. Documented exception (CPU-authored, GPU-read-only, full-texture upload). Migrate to per-region `subImage2D` if the upload becomes a hotspot.
+- `engine/prefabs/irreden/render/components/component_canvas_fog_of_war.hpp` — `C_CanvasFogOfWar::dirty_` and `allUnexplored_` gate the per-frame `subImage2D` upload of the 256² fog texture. Documented exception (CPU-authored, GPU-read-only, full-texture upload). T-161 evaluated migration to per-region `subImage2D` and deferred; see [`docs/design/fog-of-war-upload-strategy.md`](../../docs/design/fog-of-war-upload-strategy.md) for the analysis, the trigger conditions for revisiting, and the mechanical Strategy C migration sketch.
 
 ## Allocations in hot tick paths
 
--- a/docs/design/fog-of-war-upload-strategy.md
+++ b/docs/design/fog-of-war-upload-strategy.md
@@ -1,0 +1,301 @@
+# Fog-of-war CPU→GPU upload strategy
+
+**Status:** Deferred. The current full-buffer dirty-gated `subImage2D`
+upload remains the implementation, and `C_CanvasFogOfWar::dirty_` /
+`allUnexplored_` remain the documented exception to the no-dirty-flags
+rule (see `.claude/rules/cpp-ecs.md` §"No dirty flags on components").
+T-161 evaluated migration to per-write `subData` semantics (the
+follow-up that PR #638 originally suggested) and concluded the
+optimization does not pay under any workload exercised by today's
+engine.
+
+This doc captures the strategy comparison, the workload reality that
+gates the call, and the migration sketch for the day the trigger
+actually fires — so the next agent doesn't have to re-derive the
+analysis.
+
+**Problem owner:** rendering —
+`engine/prefabs/irreden/render/components/component_canvas_fog_of_war.hpp`
+and `engine/prefabs/irreden/render/systems/system_fog_to_trixel.hpp`.
+
+**Originating review:** PR #638 Opus recheck (nit N3). The reviewer at
+the time cited `.claude/rules/cpp-ecs.md` "No dirty flags on
+components"; that rule landed later in PR #651 alongside the
+C_GPUParticlePool deferred-flush variant. The rule now lists fog-of-war
+as the single allowed exception, with this doc as the deferral
+rationale.
+
+---
+
+## The texture and the access pattern
+
+`C_CanvasFogOfWar` owns a 256×256 RGBA8 texture on the iso ground plane
+(X-Y axes; +Z is height). Only the `.r` channel carries fog state
+(`unexplored` / `explored` / `visible`); the other three channels exist
+only because the Metal backend's rgba8 image binding path shares a
+binding layout with the AO and sun-shadow textures.
+
+- CPU mirror: `std::vector<std::uint8_t> cpuBuffer_` of size
+  256 × 256 = 64 KiB (.r channel only).
+- GPU texture: 256 × 256 × 4 = 256 KiB.
+- Access shape: CPU is the sole author. GPU is read-only —
+  `c_fog_to_trixel.glsl` / `.metal` samples it from the
+  `LIGHTING_TO_TRIXEL` follow-on stage. No GPU writeback.
+
+Mutators on the CPU side:
+
+- `setCell(wx, wy, state)` — single cell, taxicab-bounded.
+- `revealRadius(cx, cy, radius)` — taxicab disk fill,
+  `O((2r+1)²)` cell writes worst case.
+- `clearAll()` — `std::fill` over the whole buffer.
+
+Each mutator sets `dirty_ = true`; the `FOG_TO_TRIXEL::tick` reads the
+flag, expands the .r-only `cpuBuffer_` into a transient RGBA8 scratch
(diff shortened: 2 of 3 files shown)


jakildev (commented):
## Review — T-161: defer C_CanvasFogOfWar dirty-flag → per-region subImage2D migration

**Verdict:** approve

### What this PR does

Closes the T-161 evaluation loop: profiles the fog upload cost, concludes the optimization does not pay under any current workload, and documents the deferral. Three file changes:

1. `docs/design/fog-of-war-upload-strategy.md` — 301-line analysis covering the texture shape, three candidate strategies (A=current full-buffer dirty-gated, B=per-write subData, C=per-region bbox subImage2D), workload reality across all 4 current consumers, trigger conditions for revisiting, and a mechanical Strategy C migration sketch.
2. `.claude/rules/cpp-ecs.md` — Live deviations entry updated to link to the design doc and say "T-161 evaluated and deferred" rather than the previous open "migrate if hotspot."
3. `engine/prefabs/irreden/render/components/component_canvas_fog_of_war.hpp` — Component header comment updated to reference T-161 and the design doc.

### Strategy analysis

The core finding is correct:

- **Strategy B (per-write subData)** is strictly worse than Strategy A for `revealRadius`-dominated workloads. `revealRadius(r)` at r=128 would issue up to ~65,536 `subImage2D` calls — each with Metal `replaceRegion` command-encoder overhead — vs. Strategy A's single full-texture upload. No case exists today where Strategy B wins.
- **Strategy C (per-region bbox subImage2D)** correctly dominates Strategy A in best/average case (ties in worst). The doc correctly notes both backends support arbitrary-rect upload with tight-packed source. Strategy C is the right long-term path but not worth implementing now.

### Workload reality

The section "Workload reality today" is the load-bearing finding. All 4 consumers call `revealRadius` once at init; `dirty_` stays false for the rest of the session. Steady-state per-frame upload cost is identically zero. A 256 KiB init-time upload at ~16 GiB/s PCIe bandwidth is well under a millisecond — not a hotspot by any definition.

The trigger conditions (high-frequency per-frame mutation, sub-frame upload contention, larger texture extent) are concrete and reasonable. None are true today.

### Migration sketch completeness

The Strategy C migration sketch in the doc is mechanical and actionable: 7 steps in the component header (add 4 bbox fields, `markDirtyCell`, `markDirtyRect`, `consumeDirtyBbox`), 5 steps in the system (replace the dirty block with bbox-aware scratch sizing and upload), validation plan including the 4 existing consumers, and a note to update the cpp-ecs.md live deviations entry after migration. A future agent can execute this directly.

### ECS invariants

No runtime code changed. Comment-only update to the component header. No system or component logic touched.

### Rule update quality

The cpp-ecs.md update replaces "Migrate to per-region `subImage2D` if the upload becomes a hotspot" with a concrete link to the analysis doc and a statement that T-161 evaluated and deferred. This is an improvement — the previous wording was an open TODO with no guidance on when or how; the new wording has a link to trigger conditions and the migration sketch.

### Nits

- `docs/design/fog-of-war-upload-strategy.md` says "that rule landed later in PR #651 alongside the C_GPUParticlePool deferred-flush variant." Looking at the current cpp-ecs.md, the fog-of-war exception text was already present in the rule before PR #651 (the diff shows only the Live deviations entry being updated in this PR, not the rule body). The "landed later" note in the doc is a minor historical imprecision — the rule body's exception text predates PR #651; PR #651 added the `C_GPUParticlePool` canonical example. Not wrong enough to block, but the next edit to the design doc could tighten this sentence.

### Praise

- The workload reality section is the right move: rather than benchmarking a stress workload that doesn't exist, the doc identifies that all current consumers are one-shot init-time callers and reasons about the cost from first principles. Fabricated benchmarks would have been worse.
- The migration sketch is unusually complete for a deferral doc — enough that a future agent can execute it without re-deriving the analysis.
- Keeping `dirty_` / `allUnexplored_` as the gating mechanism in Strategy C is correct — the bbox optimization narrows the rect but doesn't remove the gate. The doc is precise about this.

Opus recheck not required.

PR #653, opened 2026-05-12
build: scope clang-format to changed files via new format-changed target

jakildev:
## Summary

- Adds `format-changed`, a diff-scoped clang-format target alongside the existing whole-tree `format`.
- Switches the simplify skill to `format-changed` so worker iterations no longer pull engine-wide formatter drift into their PRs.
- Updates `docs/agents/BUILD.md` to document the two-target split.

## Why

Fleet feedback this week recorded two workers hitting the same trap:

- **opus-worker-1 (2026-05-12 03:11):** simplify's `fleet-build --target format` step rewrote 157 unrelated files on an 11-file T-162 iteration. Required a `git stash` / selective `git checkout stash@{0} --` dance to keep the PR scoped.
- **opus-worker-2 (2026-05-12 03:50):** same target rewrote 494 files on a 1-file PR #629 nit cleanup. Worked around by `git restore .`-ing the formatter sweeps after staging the one edited file.

A less-cautious agent would silently land the engine-wide reformat. The simplify skill explicitly says "If `format` rewrote anything, those changes are part of the polish — keep them" which silently assumes diff-scoping that doesn't exist.

## Approach

- New `cmake/run_clang_format_changed.cmake` collects `git diff --name-only @{upstream}...HEAD` (falling back to `origin/master` when no upstream is set) plus `git diff --name-only HEAD` (working tree edits), intersects with the existing quality file list (so generated / vendored stay excluded), and runs clang-format on that set.
- Empty diff is a silent no-op rather than an error — common case for the doc-only diffs that simplify also gets called on.
- The bare `format` target is unchanged and stays as the explicit whole-tree option for intentional cleanup PRs.

## Test plan

- [ ] `fleet-build --target format-changed` on a worktree with a small staged + unstaged change touches only the diff files; `git status` stays scoped.
- [ ] `fleet-build --target format-changed` on a clean tree exits 0 with the "no diff" message.
- [ ] `fleet-build --target format` still rewrites everything (no behavior change).
- [ ] End-to-end via simplify: make a tiny edit, run simplify, confirm the polish step doesn't pull in unrelated reformatting.

Smoke-tested locally with `cmake -P run_clang_format_changed.cmake` against a stubbed quality list — both the "1 file matches" and "0 files match" paths emit the expected status lines.

### DIFF
--- a/.claude/skills/simplify/SKILL.md
+++ b/.claude/skills/simplify/SKILL.md
@@ -472,14 +472,18 @@ After applying fixes, run the formatter and rebuild (code diffs
 only; doc-only diffs can skip the build):
 
 ```bash
-fleet-build --target format
+fleet-build --target format-changed
 fleet-build --target <touched-target>
 ```
 
-If `format` rewrote anything, those changes are part of the polish —
-keep them. If the build broke, **revert your simplify changes** (or
-fix the break before continuing) — never push a simplify pass that
-broke the build.
+`format-changed` scopes clang-format to files changed on the
+current branch (committed vs upstream + working tree). The bare
+`format` target is whole-tree and will pull every drift in the
+repo into your PR — use it only on intentional cleanup PRs, never
+mid-iteration. If `format-changed` rewrote anything, those changes
+are part of the polish — keep them. If the build broke, **revert
+your simplify changes** (or fix the break before continuing) —
+never push a simplify pass that broke the build.
 
 ### 11. Report
 
--- a/cmake/ir_quality_tools.cmake
+++ b/cmake/ir_quality_tools.cmake
@@ -101,6 +101,20 @@ function(irreden_add_quality_targets)
             COMMENT "Checking source formatting with clang-format"
             VERBATIM
         )
+
+        # Diff-scoped formatter: only rewrites files changed on the
+        # current branch (committed vs upstream + working tree).
+        # Safe to invoke mid-iteration; the bare `format` target is
+        # whole-tree and should only run on intentional cleanup PRs.
+        add_custom_target(format-changed
+            COMMAND ${CMAKE_COMMAND}
+                -DCLANG_FORMAT_BIN="${IRREDEN_CLANG_FORMAT_BIN}"
+                -DQUALITY_FILE_LIST="${irreden_quality_file_list}"
+                -DPROJECT_ROOT="${PROJECT_SOURCE_DIR}"
+                -P "${PROJECT_SOURCE_DIR}/cmake/run_clang_format_changed.cmake"
+            COMMENT "Formatting branch-changed source files with clang-format"
+            VERBATIM
+        )
     else()
         message(STATUS "clang-format not found; format targets will print install guidance.")
         add_custom_target(format
@@ -111,6 +125,13 @@ function(irreden_add_quality_targets)
         )
 
         add_custom_target(format-check
+            COMMAND ${CMAKE_COMMAND} -E cmake_echo_color --red
+                "clang-format not found. Install LLVM/clang tools and ensure clang-format is on PATH."
+            COMMAND ${CMAKE_COMMAND} -E false
+            VERBATIM
+        )
+
+        add_custom_target(format-changed
             COMMAND ${CMAKE_COMMAND} -E cmake_echo_color --red
                 "clang-format not found. Install LLVM/clang tools and ensure clang-format is on PATH."
             COMMAND ${CMAKE_COMMAND} -E false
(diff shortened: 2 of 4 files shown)


jakildev (commented):
## Review — #653: build: scope clang-format to changed files via new format-changed target

**Verdict:** approve

### What this PR does

Adds a `format-changed` CMake custom target that scopes clang-format to files
touched by the current branch (committed vs `@{upstream}`/`origin/master` plus
working-tree edits). Switches the `simplify` skill's step 10 from the whole-
tree `format` target to `format-changed`. Updates `docs/agents/BUILD.md` to
document the two-target split.

Solves a real, observed fleet problem: two workers on the same day picked up
hundreds of unrelated files via `format` mid-iteration and had to stash/restore
to keep their PR diffs scoped.

### Logic review — cmake/run_clang_format_changed.cmake

- **Upstream detection** (`git rev-parse --abbrev-ref "@{upstream}"`, fallback
  to `origin/master`): handles both tracking and non-tracking branches correctly.
  Worktrees branched from `origin/master` (the standard fleet pattern) fall
  through to the fallback cleanly since `git -C <worktree>` resolves correctly
  in a Git worktree. No issues.
- **`_collect_git_diff` function**: `git diff --name-only ${range}` unquoted —
  CMake expands `${range}` as a list, splitting on semicolons. Both call sites
  pass single-string ranges (`origin/master...HEAD`, `HEAD`), so no token split
  occurs. Behavior is correct for all current callers.
- **Working-tree coverage**: `git diff --name-only HEAD` captures both staged
  and unstaged changes in one call. Confirmed correct.
- **Intersection with QUALITY_FILES**: constructs `PROJECT_ROOT/_rel` absolute
  paths (forward-slash everywhere on Linux/macOS), matches against the cmake-
  normalized absolute paths in QUALITY_FILES. Path format is consistent. ✓
- **Empty-diff fast path**: silent `return()` rather than error — correct; the
  common case for doc-only diffs that `simplify` also handles. ✓
- **clang-format failure handling**: collects all failed files, then one
  FATAL_ERROR with the full list. Useful for batch diagnosis. ✓

### ir_quality_tools.cmake

`format-changed` is correctly registered in both the found and not-found branches,
mirroring the existing `format` / `format-check` pattern. No issues.

### simplify/SKILL.md

Step 10 updated from `format` to `format-changed`. The new prose is explicit about
why not to use the bare `format` target mid-iteration. ✓

### docs/agents/BUILD.md

Both targets documented with a clear one-line description of each. ✓

### Nits

- `cmake/run_clang_format_changed.cmake:45` — function comment says "range is a
  CMake list (semicolon-separated tokens) passed whole into the git command line".
  Both actual call sites pass single-string git range expressions, not semicolon-
  separated lists. The comment is technically accurate for how CMake list expansion
  works in `execute_process` (and the function would handle multi-token ranges
  correctly), but a reader expecting `"${_diff_base}...HEAD"` to be a "list" might
  be confused. Consider: "range is passed as a CMake unquoted variable — tokens
  separated by semicolons become separate git arguments." Minor.

### Praise

- The fallback chain (`@{upstream}` → `origin/master`) is the right precedence
  for the fleet worktree pattern: worktrees branched from origin/master without
  an explicit upstream set still get diff-scoped formatting on the first call.
  No extra setup required by workers.
- Choosing `return()` (silent no-op) rather than an error for zero-diff is the
  correct ergonomic for doc-only and clean-tree invocations.
- Registering the `format-changed` stub in the clang-format-not-found branch
  is the right defensive move: workers get a clear red error instead of a silent
  no-op or missing-target CMake error.

- [x] `fleet-build --target format-changed` on a worktree with a small staged + unstaged change touches only the diff files; `git status` stays scoped. (per PR body: smoke-tested manually)
- [x] `fleet-build --target format-changed` on a clean tree exits 0 with the "no diff" message. (per PR body: smoke-tested)
- [x] `fleet-build --target format` still rewrites everything (no behavior change). (existing behavior, unchanged)
- [ ] End-to-end via simplify: make a tiny edit, run simplify, confirm the polish step doesn't pull in unrelated reformatting.

Escalation: none. Build/tooling and skill-doc PR only — no engine code, no ECS, no render pipeline, no C++ changes.

Opus recheck not required.

jakildev:
Addressed nit: clarified `_collect_git_diff` range comment per reviewer suggestion — the prior wording read as if callers must pre-split, but the reality is the opposite (`range` is passed unquoted in `execute_process` so CMake's own list expansion turns semicolons into separate git args). Current call sites pass single-string ranges; multi-token ranges would also work.

PR #637, opened 2026-05-12
T-155: Migrate GPU-compute cluster to member-on-System<N>

jakildev:
Claiming task. Work in progress.

PR #654, opened 2026-05-12
fleet: idempotent feedback-label clear via fleet-pr-clear-feedback-labels wrapper

jakildev:
## Summary

- Adds `scripts/fleet/fleet-pr-clear-feedback-labels`, a small bash wrapper that idempotently strips a configurable set of feedback labels from a PR — querying live labels first, then firing one independent `gh pr edit --remove-label` per label actually present.
- Switches `role-opus-worker.md` step 1.b (the AMEND-path label clear) from a combined `gh pr edit --remove-label A --remove-label B ...` call to the wrapper.
- Registers the script in `scripts/fleet/install.sh` and the `fleet-help` index.

## Why

Fleet feedback recorded two workers hitting the same trap on PR #637 within seven hours:

- **opus-worker-2 (2026-05-11 20:05):** the combined remove-label call exited non-zero on `fleet:human-deferred` (absent), but `fleet:has-nits` had already been stripped before that point. Had to manually re-add it.
- **opus-worker-1 (2026-05-12 03:11):** same shape from the other side. Logged that gh processes `--remove-label` flags left-to-right and commits each removal as it goes, exiting partway when one isn't found — so a "failed" call still mutated state.

The combined call is the role doc's prescribed pattern precisely because the broad clear handles the mixed-presence cases it's designed for. The fragility of "remove every possibly-present label" is silent: a less-cautious agent wouldn't notice the half-applied state and would move on with the PR's labels permanently wrong.

## Approach

The wrapper reads the live label set in one `gh pr view --json labels` call so the decision is consistent against the moment we decide what to remove. Then for each label in `--labels` that's also on the PR, it fires `gh pr edit <N> --remove-label <L>` individually. A single failure logs to stderr and continues — partial apply now means partial success rather than a stranded mid-operation state.

Default `--labels` is the worker AMEND set; `--labels <csv>` lets the merger and reviewer roles reuse the same primitive when they need similar broad clears.

## Test plan

- [x] `fleet-pr-clear-feedback-labels --help` prints the usage block.
- [x] `fleet-pr-clear-feedback-labels 653` on a PR with no feedback labels exits 0 with the "nothing to do" message.
- [ ] On a PR with only `fleet:has-nits`, the wrapper removes that label, ignores the absent five, exits 0.
- [ ] On a real AMEND-path iteration (next worker pickup of a `fleet:has-nits` PR), the wrapper completes cleanly where the combined call would have errored.
- [ ] `bash scripts/fleet/install.sh` (re-run) symlinks the new script into `~/bin/`.

## Notes for reviewer

- The wrapper deliberately uses `printf '%s\n' ... | grep -Fxq` for the "is this label present" check rather than parameter expansion, so labels with special characters (currently none, but defensive) work.
- The `gh pr view ... --json labels` call is a single round-trip; per-label `gh pr edit --remove-label` calls are sequential. For the typical AMEND case (1 label present out of 6 possible), this is one read + one write — same cost as the broken combined call, just without the partial-apply blast radius.

### DIFF
--- a/.claude/commands/role-opus-worker.md
+++ b/.claude/commands/role-opus-worker.md
@@ -357,7 +357,15 @@ Do the work, then exit cleanly:
 
       With checkout confirmed, **remove the feedback label** to
       prevent another agent from also picking it up:
-      `gh pr edit <N> --remove-label "human:needs-fix" --remove-label "human:blocker" --remove-label "fleet:needs-fix" --remove-label "fleet:has-nits" --remove-label "fleet:human-deferred" --remove-label "fleet:design-unblocked"`
+      `fleet-pr-clear-feedback-labels <N>`
+
+      The wrapper is idempotent: it queries the live label set
+      first and only fires a `gh pr edit --remove-label` call per
+      label actually present. Plain `gh pr edit --remove-label X
+      --remove-label Y ...` is NOT atomic — it exits non-zero on
+      the first absent label, leaving every label removed BEFORE
+      the missing one already stripped. Observed on PR #637 on
+      2026-05-11 and 2026-05-12.
 
       For `human:needs-fix` / `human:blocker` specifically, also
       mark the PR as in-progress and clear the prior approval so
--- a/scripts/fleet/fleet-help
+++ b/scripts/fleet/fleet-help
@@ -178,6 +178,10 @@ Parallel-agent fleet (tmux + Claude)
   witness [--once]               detect stale heartbeats (architects only)
   fleet-pr view|diff|comments <N> [--repo engine|game]
                                  read scout's per-PR cache + gh fallback
+  fleet-pr-clear-feedback-labels <N> [--repo …] [--labels …]
+                                 idempotently strip feedback labels from a PR
+                                 (skips absent labels; one --remove-label per
+                                 present label so partial failure can't strand)
   fleet-issue view <N> [--repo engine|game]
                                  read scout's per-issue cache + gh fallback
 
--- a/scripts/fleet/fleet-pr-clear-feedback-labels
+++ b/scripts/fleet/fleet-pr-clear-feedback-labels
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+# fleet-pr-clear-feedback-labels — idempotently strip a known set of
+# feedback labels from a PR.
+#
+# Why this exists:
+#   `gh pr edit <N> --remove-label A --remove-label B ...` is NOT
+#   atomic. gh processes removals left-to-right and exits non-zero
+#   on the first absent label, leaving every label removed BEFORE
+#   the missing one already stripped. The role files combine the
+#   canonical AMEND-path label clear into one such call and have
+#   burned multiple worker iterations on PRs with a mixed-presence
+#   feedback label set (observed on PR #637 — opus-worker-2 at
+#   2026-05-11 20:05 and opus-worker-1 at 2026-05-12 03:11).
+#
+# This wrapper queries the live label set first and only fires a
+# remove-label call for labels actually present, one per call so a
+# single failure doesn't strand the others. The default label list
+# is the worker AMEND set; override with --labels for other callers.
+#
+# Usage:
+#   fleet-pr-clear-feedback-labels <N> [--repo <slug>] [--labels <csv>]
+#
+# Options:
+#   --repo <slug>     Target repo (e.g. jakildev/IrredenEngine). Default:
+#                     the current git remote.
+#   --labels <csv>    Comma-separated label set to clear. Default:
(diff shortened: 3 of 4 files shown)


jakildev (commented):
## Review — #654: fleet: idempotent feedback-label clear via fleet-pr-clear-feedback-labels wrapper

**Verdict:** approve

### What this PR does

Adds `scripts/fleet/fleet-pr-clear-feedback-labels`, a bash wrapper that
replaces the fleet's combined `gh pr edit --remove-label A --remove-label B …`
pattern with an idempotent per-label removal loop. The wrapper queries the live
label set in one `gh pr view --json labels` call, then fires one
`gh pr edit --remove-label` per label that is actually present. A failure on
any individual label logs to stderr and continues; the script exits 1 at the
end if any removals failed but never strands the rest.

Switches `role-opus-worker.md` step 1.b from the combined call to the wrapper,
with an inline comment explaining the partial-apply failure mode. Registers in
`install.sh` (chmod + symlink) and `fleet-help`.

### Correctness analysis

- `set -euo pipefail` used correctly throughout.
- Arg parsing (`--repo`, `--labels`, `--help`, `--`) is correct; unknown
  options exit 2.
- Live label fetch (`gh pr view --json labels --jq '.labels[].name'`) is
  captured with `if !`; stderr is included in the captured output so the error
  branch can print it. Correct.
- `grep -Fxq -- "$label"` for presence check: `-F` (literal), `-x`
  (full-line match), `--` (end-of-options guard for labels starting with `-`).
  Correct and defensive.
- Loop over `$target_labels`: each matched label fires its own
  `gh pr edit --remove-label` call inside an `if` guard, so `set -e` does not
  terminate the script on per-label failure. Correct.
- Exit codes: 0 = all-present labels removed, 1 = one or more removals
  failed, 2 = usage error. Consistent with the header doc.
- install.sh integration follows the existing variable-pair pattern
  (`_SRC` / `_DEST`) and is inserted correctly in both the chmod loop and the
  symlink block.
- fleet-help entry follows the column alignment and description style of
  adjacent entries.
- role-opus-worker.md: replacement is clean; the inline explanation of the
  partial-apply failure mode is accurate and serves as in-role documentation so
  future maintainers understand why the combined call was dropped.

### Nits

- `scripts/fleet/fleet-pr-clear-feedback-labels:103` — `for label in $target_labels`
  (unquoted) relies on IFS word-splitting to iterate newline-separated labels.
  Works for the current label set (no spaces), but would silently split a label
  containing a space. More robust: `while IFS= read -r label; do ... done <<< "$target_labels"`.
  Low risk given actual label names, but worth hardening now that this is the
  canonical primitive.

- `scripts/fleet/fleet-pr-clear-feedback-labels:99` — the pipeline ends with
  `grep -v '^$'`. With `set -e`, if every line produced by the `tr | sed` step
  is empty (e.g. caller passes `--labels ""`), grep exits 1 and the script
  terminates without a message. Guarding with `|| true` at the pipeline end
  (or adding a non-empty check on `labels_csv` after arg parsing) would give
  a cleaner failure mode for that edge case.

- `scripts/fleet/fleet-pr-clear-feedback-labels:56` — `sed -n '2,35p'` for
  `--help` output is fragile if the header block grows past line 35. A sentinel
  like `# END_USAGE` with `sed -n '2,/END_USAGE/p'` would survive future edits
  to the header.

### Praise

- The "read once, write per-present-label" design is the right architecture.
  A naive retry-on-failure approach still leaves the question of which labels
  got removed before the error; reading first removes that ambiguity entirely.
- The inline note in role-opus-worker.md ("Plain gh pr edit --remove-label X
  --remove-label Y ... is NOT atomic — observed on PR #637 ...") is exactly the
  right level of documentation: specific incident reference, causal explanation,
  clear ownership of the invariant. Future maintainers will not be tempted to
  reinline the combined call without understanding the failure mode.
- `printf '%s\n' "$live_labels" | grep -Fxq -- "$label"` is a correct and
  safe pattern for exact-match label presence; the `--` guard for dash-prefixed
  labels is a nice defensive touch even though current label names do not start
  with a dash.

- [x] `fleet-pr-clear-feedback-labels --help` prints usage block (per PR body)
- [x] `fleet-pr-clear-feedback-labels 653` on a PR with no feedback labels exits 0 with "nothing to do" message (per PR body)
- [ ] On a PR with only `fleet:has-nits`, wrapper removes that label, ignores absent five, exits 0
- [ ] End-to-end on a real AMEND-path iteration: wrapper completes cleanly where the combined call would have errored
- [ ] `bash scripts/fleet/install.sh` (re-run) symlinks the new script into ~/bin/

Escalation: none. Safe for merge. Fleet tooling and role-doc update only — no engine code, ECS, render pipeline, audio, or concurrency changes.

Opus recheck not required.

PR #656, opened 2026-05-12
fleet: don't count idle-pane reservations against dispatcher cap

jakildev:
## Summary

- `count_active_for_role` was counting every role-matched reservation against the per-role concurrency cap, even when the worktree's pane was sitting idle at a shell prompt. After a `fleet-down` (or sleep that timed out an iteration), the leftover reservations would permanently pin the cap and no resume could dispatch — the trigger sat queued, the panes sat idle, the dispatcher logged `at concurrency cap` every tick.
- The reservation IS the resume signal: [`role-sonnet-author.md`](.claude/commands/role-sonnet-author.md) step 0.5 reads it on the next iteration to resume the prior branch. So the fix is to count a reservation only when its worktree's matching pane is *busy* — either running a non-shell command, or running a `fleet-dispatch-wrap` child under a shell. An idle pane's leftover reservation is queued work for the next dispatch to consume, not an active iteration eating budget.
- Falls back to the legacy "always count" path when tmux isn't available (e.g. the `--count-active` CLI invocation outside a live fleet) so existing test harnesses still match the documented contract.

## Test plan

- [x] `bash scripts/fleet/tests/test_dispatcher_concurrency_cap.sh` — **15/15 pass** (3 added: T4 strengthened with `list-panes` stub; T4b verifies idle-pane reservation → 0; T4c verifies busy-pane reservation → 1).
- Tests now point `FLEET_SESSION` at a per-pid fake name so a real fleet tmux session on the dev machine can't bleed pane state into the sandbox.

### DIFF
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -608,17 +608,60 @@ usage_gate_open() {
 
 # Print the count of unique active worktrees for a role. "Active" =
 # (in-flight dispatch record for this role) ∪ (reservation whose task's
-# Model maps to this role). Dedup keys on worktree-basename, which is
-# shared between dispatch records (via the pane's pane_current_path) and
-# reservations (whose filename IS the worktree basename).
+# Model maps to this role AND whose matching pane is currently busy).
+# Dedup keys on worktree-basename, which is shared between dispatch
+# records (via the pane's pane_current_path) and reservations (whose
+# filename IS the worktree basename).
+#
+# A reservation on an idle pane is NOT counted. That case is the
+# resume-after-crash signal that role docs (see role-sonnet-author.md
+# step 0.5) consume on the next iteration — counting it would pin the
+# cap and prevent the resume from ever dispatching, which is the exact
+# stuck-fleet failure mode this guards against. The matching pane's
+# busy state is the disambiguator: a busy pane means an iteration is
+# live (the normal dispatch+reservation pairing); an idle pane means
+# the prior iteration crashed and the reservation is queued for the
+# next dispatch to consume.
 #
 # When tmux isn't running (shouldn't happen for a live dispatcher, but
 # the standalone --count-active CLI invokes this without a session),
 # dispatch records key on pane id instead — same dedup contract still
-# holds across two dispatch records for the same pane.
+# holds across two dispatch records for the same pane. Reservations in
+# that mode fall back to the legacy "always count" behavior since we
+# can't probe pane state; the tests that exercise --count-active
+# without tmux depend on this fallback.
 count_active_for_role() {
     local role="$1"
     declare -A seen=()
+
+    # Build worktree → busy map for panes tagged with this role. Used
+    # by the reservation pass below to skip idle-pane reservations.
+    declare -A worktree_busy=()
+    local have_tmux=0
+    if session_exists; then
+        have_tmux=1
+        local pane_id pane_role pane_cmd wt_path wt_name
+        while IFS=$'\t' read -r pane_id pane_role pane_cmd; do
+            [[ "$pane_role" == "$role" ]] || continue
+            wt_path=$(tmux display-message -t "$pane_id" -p '#{pane_current_path}' 2>/dev/null || true)
+            [[ -n "$wt_path" ]] || continue
+            wt_name=$(basename "$wt_path")
+            case "$pane_cmd" in
+                bash|zsh|sh|fish)
+                    # Shell allowlist matches find_idle_panes_for_role's
+                    # idle detection — but a fleet-dispatch-wrap child
+                    # still means the pane is mid-iteration even though
+                    # tmux reports the parent shell's name.
+                    if pane_has_running_wrapper "$pane_id"; then
+                        worktree_busy["$wt_name"]=1
+                    fi
+                    ;;
+                *)
+                    worktree_busy["$wt_name"]=1
+                    ;;
+            esac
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: don't count idle-pane reservations against dispatcher cap

**Verdict:** approve

Tight, well-scoped fix for a real stuck-fleet failure mode. The reservation IS the resume signal (per `role-sonnet-author.md` step 0.5), so counting it on an idle pane permanently pins the cap after `fleet-down` / sleep / timeout. The new `worktree_busy` disambiguator is the right key: live iteration → count it; queued resume signal → leave a slot open for it to consume. Comments are unusually clear about the design intent.

I re-ran the four shell test suites cited in the PR body and confirmed the asserted pass counts:

- `test_dispatcher_concurrency_cap.sh` — 15/15 pass (T4b/T4c verify the new idle vs. busy distinction).
- `test_dispatcher_usage_gate.sh` — 13/13 pass.
- `test_fleet_claim_master_lock.sh` — 26/26 pass.
- `test_fleet_claim_model_gate.sh` — 8/8 pass.
- `test_fleet_up_conf_bootstrap.sh` — 5/5 pass (extra sibling, no regression).

`bash -n` clean on both touched files. The `set -euo pipefail` interactions hold up: `pane_has_running_wrapper` is always called in an `if`-guarded context, the `pgrep` failure path is non-fatal, and the `< <(list_pane_state)` process-substitution producer is harmless when `list_pane_state` early-returns under no-session.

### Nits

- `scripts/fleet/fleet-dispatcher:639` — `declare -A worktree_busy=()` matches the existing pattern (`declare -A seen=()` at line 635) but neither is technically `local`. Inside a bash 4+ function `declare` is effectively local, so this is fine and consistent — flagging only because if either is ever refactored, `local -A` would be the more obviously-correct spelling. No action needed unless you want to align both at once.
- `scripts/fleet/fleet-dispatcher:633-712` — `count_active_for_role` now performs `1 + N` tmux calls per invocation (one `list-panes` plus one `display-message` per role-tagged pane) and is called per-role per-tick. With ~6 panes per role and a 10s tick this is well under any concern, but worth a mental note if a future role explosion bumps the pane count by an order of magnitude.
- `scripts/fleet/tests/test_dispatcher_concurrency_cap.sh:221,259` — the test stubs emit fake `pane_pid` values (`999999`, `999998`) and rely on `pgrep -P <pid> -f fleet-dispatch-wrap` finding nothing. On a typical Linux box with `kernel.pid_max=4194304` (Ubuntu default) a coincidental real process at that pid could in theory have a `fleet-dispatch-wrap` child and flip T4b's "idle" verdict. Practically near-zero risk on the fleet CI box, but if you want bulletproof isolation a pid like `-1` (always invalid) or `0` would be safer for the fake. Not blocking.
- `scripts/fleet/fleet-dispatcher:644-663` — `worktree_busy` is only populated for panes whose `@fleet-role == role`. If a pane were mis-tagged (e.g. a worker pane carrying the wrong `@fleet-role`) but actively running an iteration for the queried role, its reservation would silently be dropped from the count and the dispatcher could double-dispatch. In practice fleet-up sets the tag deterministically and it can't drift mid-session, plus the dispatch-record path (lines 666-685) still keys on the role from the JSON body so in-flight work is independently counted. Pointing out for future-you: the role tag is now load-bearing for cap correctness, not just for `find_idle_panes_for_role`.

### Praise

- The doc comment on `count_active_for_role` (lines 609-632) is exemplary: it explains both the **what** (idle reservation skipped) and the **why** (resume signal, the exact failure mode it guards against, the fallback contract for `--count-active` outside a live fleet). Future-you will thank present-you.
- The T4 stub upgrade to serve `list-panes` was the right call. The body's note that "previously dedup passed by accident through the no-tmux fallback" is the kind of honest test-fragility receipt that's easy to skip and worth keeping.
- `FLEET_SESSION="fleet-test-$$"` is a clean per-pid isolation trick — pid uniqueness among running processes makes it effectively collision-proof against a real `fleet` session on the dev box without resorting to mocking the whole tmux binary in tests that don't already do so.

- [x] `bash scripts/fleet/tests/test_dispatcher_concurrency_cap.sh` — confirmed 15/15.
- [x] `bash scripts/fleet/tests/test_dispatcher_usage_gate.sh` — confirmed 13/13.
- [x] `bash scripts/fleet/tests/test_fleet_claim_master_lock.sh` — confirmed 26/26.
- [x] `bash scripts/fleet/tests/test_fleet_claim_model_gate.sh` — confirmed 8/8.
- [ ] Optional: bring up a real fleet, kill an iteration mid-flight to leave a reservation behind, confirm the next dispatcher tick resumes rather than logging `at concurrency cap`.

Reviewed by Claude Opus 4.7 (review-pr skill)

jakildev:
Addressed the test-pid fragility nit in c65aa59 — added a `pgrep` stub alongside the tmux stub so T4b's wrapper-child probe is fully host-isolated. Test still passes 15/15.

Other two nits acknowledged as won't-fix (consistent with existing `declare -A seen=()` pattern and the per-tick cost is negligible at current pane counts). The mis-tagged-pane note is a fair "future-you" callout but the role tag drifting mid-session isn't a current threat model.

Ready for merge whenever; the rest of the resume path (fleet-up Step 1.5 reconciler keeping the OPEN-PR reservations, Step 2 worktree-reset reservation guard, Step 3e2 symmetric bootstrap firing on `tasks_open`) was independently traced and should kick the stuck sonnet-author panes back into life on the next `fleet-up`.

PR #658, opened 2026-05-12
fleet: retain bootstrap trigger during boot fan-out window

jakildev:
## Summary

- At `fleet-up`, the dispatcher's first tick can race with tmux's asynchronous `@fleet-role` tag application. Only a subset of role-tagged panes are visible to `find_idle_panes_for_role` on that tick, the bootstrap trigger gets consumed by the partial dispatch, and the late-tagged panes sit idle until scout notices a projection-content delta — which can take an arbitrarily long time when the task set is unchanged across the boot.
- Observed live on 2026-05-12 after merging [#656](https://github.com/jakildev/IrredenEngine/pull/656): 2 sonnet-author panes + 2 opus-worker panes available, but the dispatcher only fired into `%0` and `%2`. `%3` and `%6` stayed idle for ~5 minutes until a manual `touch ~/.fleet/state/triggers/<role>` re-armed them.
- Fix: in `dispatch_role`, retain the trigger after a successful dispatch when (a) we're still inside the boot fan-out window and (b) the role's post-dispatch active count is below cap. Existing dispatch records prevent double-dispatch into the panes that just fired, so the next tick can cleanly catch any pane whose tag landed late. Outside the window, legacy consume-on-success semantics apply.

Scoped via `SECONDS` (bash builtin, seconds since dispatcher start) and `BOOT_FANOUT_WINDOW_SECONDS` env override (default 60s; race typically resolves in 1–2 ticks at 10s each).

Decision logic extracted into `should_retain_trigger_for_fanout_race` and exposed via a new `--retain-trigger-check <role> <active>` CLI subcommand so the tests can drive it without spinning up the main loop.

## Test plan

- [x] `bash scripts/fleet/tests/test_dispatcher_concurrency_cap.sh` — **20/20 pass** (5 new: T9 retain inside window; T10 cap-saturated consume; T11 window-expired consume; T12 single-pane saturated consume; T13 cap=0 short-circuit consume).
- [x] `bash scripts/fleet/tests/test_fleet_up_conf_bootstrap.sh` — 5/5 pass.
- [ ] Live verification (next `fleet-up`): all role panes get dispatched on the first or second tick, no manual trigger touch required.

### DIFF
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -214,6 +214,15 @@ PERIODIC_REARM_INTERVAL_SECONDS="${FLEET_DISPATCHER_REARM_INTERVAL:-300}"
 # (consume-on-success) are unchanged. 60s is generous — the race
 # typically resolves in 1–2 ticks (10–20s after start).
 BOOT_FANOUT_WINDOW_SECONDS="${FLEET_DISPATCHER_BOOT_FANOUT_WINDOW_SECONDS:-60}"
+# Validate-and-clamp: a non-numeric override (operator typo,
+# trailing-suffix like "9999s") would otherwise emit arithmetic errors
+# every tick or, under `set -u`, kill the daemon outright. Fall back
+# to the default and log a one-shot warning instead.
+if ! [[ "$BOOT_FANOUT_WINDOW_SECONDS" =~ ^[0-9]+$ ]]; then
+    printf '[%s dispatcher] BOOT_FANOUT_WINDOW_SECONDS=%q is not a non-negative integer; falling back to 60\n' \
+        "$(date -u '+%Y-%m-%dT%H:%M:%SZ')" "$BOOT_FANOUT_WINDOW_SECONDS" >&2
+    BOOT_FANOUT_WINDOW_SECONDS=60
+fi
 
 # Roles this dispatcher is responsible for. Architects are excluded
 # (kept on fleet-babysit). Aligned with the worker / reviewer /
@@ -853,7 +862,8 @@ dispatch_role() {
         local active
         active=$(count_active_for_role "$role")
         if should_retain_trigger_for_fanout_race "$role" "$active"; then
-            log "$role: dispatched $dispatched, active=$active < cap=${ROLE_CONCURRENCY[$role]} within boot fan-out window (${SECONDS}s < ${BOOT_FANOUT_WINDOW_SECONDS}s); keeping trigger for late-tag race"
+            local cap="${ROLE_CONCURRENCY[$role]:-0}"
+            log "$role: dispatched $dispatched, active=$active < cap=$cap within boot fan-out window (${SECONDS}s < ${BOOT_FANOUT_WINDOW_SECONDS}s); keeping trigger for late-tag race"
             return
         fi
         # Steady state (or boot window with cap saturated): at least
@@ -1084,6 +1094,10 @@ case "${1:-}" in
         # boot window is "in" by default unless the override drops it.
         if [[ -z "${2:-}" || -z "${3:-}" ]]; then
             echo "usage: fleet-dispatcher --retain-trigger-check <role> <active-count>" >&2
+            exit 2
+        fi
+        if ! [[ "$3" =~ ^[0-9]+$ ]]; then
+            echo "fleet-dispatcher --retain-trigger-check: <active-count> must be a non-negative integer (got: $3)" >&2
             exit 2
         fi
         if should_retain_trigger_for_fanout_race "$2" "$3"; then
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -195,6 +195,26 @@ fi
 # (fleet-queue-tick, spawned by fleet-state-scout on projection change).
 PERIODIC_REARM_INTERVAL_SECONDS="${FLEET_DISPATCHER_REARM_INTERVAL:-300}"
 
+# Boot fan-out window. During the first N seconds after dispatcher start,
+# `dispatch_role` keeps the per-role trigger in place after a successful
+# dispatch when the post-dispatch active count is still below cap. This
+# guards against a race that surfaces at `fleet-up`: tmux applies the
+# `@fleet-role` tags to panes asynchronously, so the dispatcher's first
+# tick can see only a subset of role-tagged panes (e.g. one of two
+# sonnet-author panes). With the existing single-shot trigger semantic
+# that subset gets dispatched, the bootstrap trigger is consumed, and
+# the late-tagged panes sit idle until scout notices a projection
+# delta — which can take an arbitrarily long time when the same task
+# set was already actionable before the boot. Keeping the trigger for
+# the boot window gives subsequent ticks a chance to dispatch into the
+# missed panes; the dispatch records written here prevent
+# double-dispatch into the already-running panes.
+#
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: retain bootstrap trigger during boot fan-out window

**Verdict:** approve

Focused, well-justified fix for a real race observed on 2026-05-12 after #656. The PR body has a strong narrative (observed failure + mechanism + scope), the fix is correctly scoped to the boot window only (steady-state semantics unchanged), the decision predicate is cleanly factored out, and the test coverage hits all five interesting cases: under-cap-in-window retain, at-cap-in-window consume, expired-window consume, single-pane saturated consume, and cap=0 short-circuit. All five cited test suites pass on this worktree (20+13+26+8+5 = 72 tests, plus `bash -n` clean on both files).

Specifically verified:

- `SECONDS` builtin auto-increments per-bash-process and resets to 0 on every new invocation — both the daemon use and the CLI-subcommand tests work as intended. A `pkill -HUP` or `fleet-up`-driven TERM+relaunch resets `SECONDS=0` so the boot window correctly re-arms on restart (the desired behavior).
- The new `--retain-trigger-check` case arm is positioned cleanly after `--gate-status` and matches only its exact subcommand string (no shadowing of the other CLI paths).
- The conf-file sourcing at line 92 runs before the case statement, so `BOOT_FANOUT_WINDOW_SECONDS` from `~/.fleet/fleet-up.conf` is honored in both daemon and CLI modes.
- Cooldown'd-pane scenario (the user-flagged concern): within boot window, trigger sits until either cooldown expires (then dispatches) or window closes (then `dispatched=0` keeps the trigger via the existing `else` branch). Outside the window, same behavior as today's master. No regression.
- `FLEET_CONCURRENCY_OPUS_WORKER=0` correctly drops to cap=0 because `:-` only falls back on empty/unset, not zero.

### Nits
- `scripts/fleet/fleet-dispatcher:856` — the success-path log message uses `${ROLE_CONCURRENCY[$role]}` without the `:-0` fallback, while `should_retain_trigger_for_fanout_race` itself uses `${ROLE_CONCURRENCY[$role]:-0}`. In practice `dispatch_role` is only called for roles in `DISPATCHED_ROLES` (which mirrors `ROLE_CONCURRENCY`'s keys), so this is latent — but under `set -u` an unknown-role caller would trip "unbound variable" here while passing the predicate. Cheap to harden: `cap=${ROLE_CONCURRENCY[$role]:-0}` once before the log line, reuse `$cap`.
- `scripts/fleet/fleet-dispatcher:1085-1093` — `--retain-trigger-check` validates that `$2`/`$3` are non-empty but doesn't validate that `$3` is numeric. Passing `--retain-trigger-check opus-worker foo` trips `set -u` inside the arithmetic (`foo: unbound variable`) with a confusing error and rc=1, rather than a clean usage message. Optional `[[ "$3" =~ ^[0-9]+$ ]]` guard would localize the diagnostic; this is purely a test/debug entry point so impact is low.
- `scripts/fleet/fleet-dispatcher:216` — if the operator sets `FLEET_DISPATCHER_BOOT_FANOUT_WINDOW_SECONDS` to a string with a numeric prefix and non-numeric suffix (e.g. `"9999s"`), bash arithmetic emits `((: 9999s: value too great for base)` to stderr on every successful dispatch within the boot window and silently flips to consume. Garbage value (`"notanumber"`) trips `set -u` and exits the daemon. Neither is a realistic operator configuration, but a one-shot validate-and-clamp at startup (e.g. `[[ "$BOOT_FANOUT_WINDOW_SECONDS" =~ ^[0-9]+$ ]] || BOOT_FANOUT_WINDOW_SECONDS=60`) would make the boundary explicit. Same pattern would also be worth applying to other numeric env vars in this file as a follow-up.

### Praise
- The decision to factor `should_retain_trigger_for_fanout_race` out into a tiny predicate function and expose it via `--retain-trigger-check` is excellent. The tests drive the actual production logic (not a duplicate), so the test/prod path can't drift; the CLI subcommand is also a useful operator debug surface.
- Choosing `SECONDS` over a wall-clock-based `start_ts=$(date +%s)` is the right call here — it's monotonic, free, and naturally resets on daemon restart (which is when the race fires anyway). The PR body's "race typically resolves in 1–2 ticks" framing pairs well with the 60s default.
- Boundary tests T10 (`active == cap`) and T13 (`cap == 0`) are exactly the right edge cases to cover — T10 catches the off-by-one between `<` and `<=`, T13 catches the unconfigured-role short-circuit.

- [x] Re-run by the reviewer: `bash scripts/fleet/tests/test_dispatcher_concurrency_cap.sh` — 20/20 pass.
- [x] Re-run by the reviewer: 4 other cited suites pass (13+26+8+5).
- [x] `bash -n` clean on both modified files.
- [ ] Live verification at next `fleet-up`: confirm all role panes dispatch on the first or second tick without manual trigger touch (already noted in PR body as the next live check).

Reviewed by Claude Opus 4.7 (review-pr skill)

jakildev:
All three nits addressed in 2ea2ac2:

- `fleet-dispatcher:860` log line now reads cap via `${ROLE_CONCURRENCY[$role]:-0}` (matches the predicate's fallback).
- `--retain-trigger-check <active>` numeric validation added — typo'd active arg now exits rc=2 with a clean usage message instead of `unbound variable`.
- `BOOT_FANOUT_WINDOW_SECONDS` validate-and-clamp at startup: non-numeric overrides (`"9999s"`, garbage strings) fall back to 60s with a one-shot stderr warning instead of per-tick arithmetic spam or `set -u` daemon kill.

Tests T14 (non-numeric active arg → rc=2) and T15 (non-numeric env override → clamps and behaves as default) added; **22/22 pass**.

jakildev (commented):
## Re-review — fleet: retain bootstrap trigger during boot fan-out window

**Verdict:** approve (re-confirmed, nits resolved)

The prior Opus 4.7 review (18:01:15Z on commit 3dda848) approved with 3 nits. All three are resolved in commit 2ea2ac2. Verifying each:

### Prior-nit resolution

**Nit 1 (fleet-dispatcher:856 — log line missing `:-0` fallback):**
Resolved. `fleet-dispatcher:865` now reads `local cap="${ROLE_CONCURRENCY[$role]:-0}"` and uses `$cap` in the log line. The fallback matches the predicate's own `:-0`, so an unknown-role caller no longer trips `set -u` after the predicate passes.

**Nit 2 (`--retain-trigger-check` numeric validation on active arg):**
Resolved. `fleet-dispatcher:1099-1102` adds `[[ "$3" =~ ^[0-9]+$ ]]` guard before the predicate call; invalid input exits rc=2 with a clean diagnostic instead of surfacing as `unbound variable` inside the arithmetic.

**Nit 3 (`BOOT_FANOUT_WINDOW_SECONDS` validate-and-clamp at startup):**
Resolved. `fleet-dispatcher:221-225` adds the regex guard at startup — non-numeric override falls back to 60s with a one-shot stderr warning. Placed immediately after the variable assignment at line 216, so it fires once per daemon start (and once per CLI invocation that sources the startup block), not per tick.

### New tests (T14, T15)

Both tests drive the actual production predicate via the `--retain-trigger-check` subcommand — no mock duplication risk.

T14 verifies rc=2 on a non-numeric active arg. T15 is the sharper test: passes `"9999s"` as the env override and expects `retain` output — if the clamp fails and arithmetic errors silently flip the comparison to false, T15 fails (would see `consume`). Correct test design.

### No further issues

No cross-host smoke needed (fleet scripts only, no render path changes).

Opus recheck not required.

PR #678, opened 2026-05-12
engine/asset: add binary I/O + chunk-table + name-table + JSON sidecar primitives (T-166)

### LINKED ISSUE #663: asset: extend engine/asset/ with BinaryWriter/Reader + chunk-table header + JSON sidecar emitter
## Area
`engine/asset/`

## Suggested model
`[opus]` — core-engine module extension with cross-cutting API decisions (binary I/O contract, chunk-table format, name-table encoding). Downstream tickets (#664, #665, #666, #667 .. ) all consume this foundation.

## Context

The engine has no generic binary serialization infrastructure today. `engine/asset/` does unchecked `fread`/`fwrite` for two formats (`.txl`, `.irsprite`) with no versioning, no checksums, no chunk discipline. Issue #199 + the editor save/load work both need this foundation, and want it to be extensible so adding a new SDF primitive, relation type, or component never bumps a format version.

This ticket extends `engine/asset/` with the shared binary-I/O primitives every new format (`.vxs`, `.rig`, the ECS world snapshot in `engine/world/`) will consume. No functional consumer yet — #664+ are the first consumers.

**Design note:** earlier scoping considered a separate `engine/persist/` module, but `engine/asset/` already does this exact thing (file I/O for content) and its `CLAUDE.md` explicitly says it can broaden with a design pass. This is that pass.

## Acceptance

- [ ] `engine/asset/` builds clean on `linux-debug` and `macos-debug` after the additions.
- [ ] `engine/asset/include/irreden/asset/binary_io.hpp` (new):
  - `BinaryWriter` (file backend + memory backend; both implement the same interface).
  - `BinaryReader` (file + memory).
  - Primitives: `writeU8/U16/U32/U64/I*/F32/F64`, `writeVarUInt`, `writeBytes(span)`, `writeString(string_view)`. Mirror for reads. All little-endian fixed-width.
  - `Result<T>` return type for reads — clear error on truncated / bad-magic / version-too-new.
- [ ] `engine/asset/include/irreden/asset/chunk_header.hpp` (new):
  - Magic-bytes + uint32-version + chunk-count header struct (12 bytes).
  - Chunk table entry: `{ char tag[4]; uint64 offset; uint64 size; }`.
  - `writeHeader(writer, magic, version, chunks)` / `readHeader(reader)` helpers.
  - Unknown chunks: reader exposes them as `std::span<uint8_t>` so loaders skip-without-erroring (Extensibility Rule #1).
- [ ] `engine/asset/include/irreden/asset/name_table.hpp` (new):
  - String name-table chunk helper: stores `(uint32 numeric_id, string name)` pairs so future builds with new enum values round-trip on older builds as "unknown name, skipped" (Extensibility Rule #2).
- [ ] `engine/asset/include/irreden/asset/json_sidecar.hpp` (new):
  - Minimal stdlib-only JSON emitter (no third-party JSON dep). Supports flat objects, arrays, strings, numbers, bools. **Read side is not implemented** — sidecars are write-only, regenerated from binary (Extensibility Rule #6).
- [ ] Unit tests covering: round-trip primitives, varint edge cases, truncated reads, bad magic, version-too-new, unknown-chunk-tag (loader yields raw bytes, caller skips), name-table round-trip with unknown id.
- [ ] `engine/asset/CLAUDE.md` documents the seven Save Format Extensibility Rules as the module's contract and lists the new primitives.

## Critical files

- `engine/asset/include/irreden/asset/binary_io.hpp` (new)
- `engine/asset/include/irreden/asset/chunk_header.hpp` (new)
- `engine/asset/include/irreden/asset/name_table.hpp` (new)
- `engine/asset/include/irreden/asset/json_sidecar.hpp` (new)
- `engine/asset/src/binary_io.cpp` (new)
- `engine/asset/src/chunk_header.cpp` (new)
- `engine/asset/src/name_table.cpp` (new)
- `engine/asset/src/json_sidecar.cpp` (new)
- `engine/asset/CLAUDE.md` (extend with primitives + rules)
- `engine/asset/test/test_binary_io.cpp` (new)

## Notes

- No `.vxs`, `.rig`, world snapshot yet — those layer on top in #664 .. #667.
- Sets the binary contract used by every future asset format.
- World-snapshot serializer (#667) lives in `engine/world/` (it walks the archetype graph) and consumes these primitives — `engine/asset/` itself does not depend on `engine/entity/` or `engine/world/`.

## Cross-references

- Parent epic: #199
- Blocked by: #662 (T-S0 doc rename — locks the architecture)
- Blocks: #664 (.vxs dense), #665 (.vxs shape-group), #666 (.rig joints), #667 (#199 world snapshot)

jakildev:
## Summary

Foundation for the editor-epic asset-format work. Adds shared binary
I/O primitives in `engine/asset/` so every upcoming binary format
(`.vxs`, `.rig`, ECS world snapshot, future formats) has one place
to consume rather than each new format hand-rolling `fread`/`fwrite`
+ its own magic-bytes discipline.

- **`binary_io.{hpp,cpp}`** — `BinaryWriter` / `BinaryReader`
  abstract sinks with file + in-memory backends. Little-endian
  fixed-width integers, IEEE 754 floats, ULEB128 varuints,
  length-prefixed strings. Reads return `Result<T>` with typed
  `status_` / `value_` so a truncated file or malformed varint
  surfaces as a recoverable diagnostic (path + offset), not a
  crash — Extensibility Rule #5.
- **`chunk_header.{hpp,cpp}`** — 12-byte `AssetHeader` (magic +
  version + chunkCount) + chunk table of `{ tag, offset, size }`
  entries. `writeChunked()` computes offsets and writes header +
  table + bodies in one pass; `readChunks()` validates magic +
  max-known-version and surfaces unknown chunk tags as raw bytes so
  older loaders silently skip them — Extensibility Rule #1.
- **`name_table.{hpp,cpp}`** — `(uint32 id, string name)` pairs
  serialized as a chunk body; `NameTable` helper for bidirectional
  lookup. Supports Extensibility Rule #2: prefer name → current
  enum, fall back to id only when the name table is absent.
- **`json_sidecar.{hpp,cpp}`** — stdlib-only write-only JSON
  emitter. No third-party dep, no read side (Extensibility Rule #6
  — sidecars regenerate from the binary on every save, ignored on
  load).
- **`engine/asset/CLAUDE.md`** — documents the four new headers as
  the module contract and inlines the seven Save Format
  Extensibility Rules (long-form home stays in
  `docs/design/entity-editor-epic.md`).

No functional consumer in this PR. T-167 (`.vxs` dense), T-168
(`.vxs` shape-group), T-169 (`.rig`), and the ECS world snapshot
in `engine/world/` are the first consumers.

## Test plan

- [x] `fleet-build --target IrredenEngineAsset` — clean on macOS
- [x] 25 new tests in `test/asset/binary_io_test.cpp`:
  - Round-trip integer/float primitives + verified little-endian
    byte pattern (`0xDEADBEEF` → `EF BE AD DE`)
  - Varuint edge cases: 0, 127, 128, u32 max, u64 max, no-terminator
  - Length-prefixed string round-trip (incl. UTF-8 / emoji / control
    chars) + invalid-length rejection
  - Truncated reads, bad magic, version-too-new
  - Unknown-chunk-tag survives as raw bytes (Rule #1)
  - Chunk-table out-of-bounds offset rejected
  - Name-table round-trip + bidirectional lookup +
    unknown-id passthrough (Rule #2)
  - JSON: flat object, nested object/array, string escaping
    (`"`, `\`, `\n`, `\t`, control chars), bool/null/float, NaN
    sanitization, file write
- [x] Full `IrredenEngineTest` suite: **519/519 pass**

## Notes for reviewer

- Result wrapper uses trailing-underscore public members
  (`status_`, `value_`) per the engine naming convention.
- `kAssetHeaderSize` / `kChunkTableEntrySize` are explicit byte
  constants — `sizeof(AssetHeader)` would include compiler-dependent
  padding.
- The detail-namespace endianness helpers in `binary_io.cpp` are
  byte-shift based, host-byte-order agnostic (no `__builtin_bswap`
  or memcpy-of-bytes-reversed shortcuts), so they're safe on
  big-endian even though the engine doesn't target any today.
- `JsonSidecarWriter` emits `null` for NaN/Inf with a log warning
  rather than producing invalid JSON.

Closes #663

### DIFF
--- a/engine/asset/CLAUDE.md
+++ b/engine/asset/CLAUDE.md
@@ -30,8 +30,10 @@ existing trixel-texture and sprite-sheet I/O:
 The binary-I/O primitives (`BinaryWriter`/`Reader`, chunk-table header
 helpers, name-table encoding, JSON sidecar emitter) used by every new
 format also live here so consumers have one place to look. All formats
-obey the seven "Save format extensibility rules" documented in the
-editor-epic design doc.
+obey the seven [Save format extensibility rules](#save-format-extensibility-rules)
+below — the same list lives in `docs/design/entity-editor-epic.md` as
+the design-doc home of the rules; this module's contract is to provide
+the primitives that make those rules cheap to follow.
 
 The one save/load surface that does **not** live here is the ECS world
 snapshot (issue #199). It walks the archetype graph, so it lives in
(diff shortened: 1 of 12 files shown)


jakildev (commented):
## Review — T-166: engine/asset binary I/O + chunk-table + name-table + JSON sidecar primitives

**Verdict:** approve

Solid, well-scoped foundation for the editor-epic asset-format work. The four
new headers compose cleanly, the test suite covers the critical edge cases, and
the extensibility-rule contract is honored end-to-end. Two nits below.

### Nits

- **`chunk_header.cpp:88` — overflow guard comment is slightly misleading.**
  The overflow check reads:
  ```
  if (e.offset_ > totalSize || e.size_ > totalSize || e.offset_ + e.size_ > totalSize) {
  ```
  The accompanying comment says "Catch overflow by checking the addends
  separately" — but the individual guards only prevent wrap-around when each
  addend fits below `totalSize`, which is true for any realistic file. For
  defense-in-depth on pathological inputs, the idiomatic form that is
  overflow-free by construction is:
  ```
  if (e.offset_ > totalSize || e.size_ > totalSize - e.offset_) {
  ```
  (The second branch is safe only after the first is false, which the
  short-circuit guarantees.) No functional difference for any file the engine
  will ever produce — purely cosmetic.

- **`name_table.hpp:67` / `name_table.cpp:81-86` — returned `string_view`
  lifetime not documented.**
  `NameTable::nameById` returns `std::optional<std::string_view>` that points
  into `m_entries[idx].name_`. Any subsequent call to `add()` may reallocate
  `m_entries`, dangling the view. The hot-path comment on `idByName` notes that
  "serialization runs once per save" (implying `NameTable` is built once then
  read-only), but `add()` exists and the header doesn't state the invalidation
  rule. A one-liner on the declaration:
  ```
  /// Invalidated by any subsequent call to `add()`.
  std::optional<std::string_view> nameById(std::uint32_t id) const;
  ```
  makes the contract explicit for future callers.

### Praise

- **`readNameTable` cap on `reserve()`** (`name_table.cpp:29-31`) — guarding
  the upfront reserve with `countR.value_ < cap ? countR.value_ : cap` is easy
  to miss and exactly right. A corrupted count claiming billions of entries
  doesn't OOM the process.

- **ULEB128 overflow guard** (`binary_io.cpp:249`) — `shift >= 64 || (shift == 63 && chunk > 1)` is the correct two-condition terminator for a 64-bit varint; the comment explaining the 10-byte cap is clear.

- **`readChunks` combined-helper design** — magic check to version check to chunk table to bulk body copy in one pass, with clean Result<T> propagation and no crash paths, honors Extensibility Rule #5 throughout.

- [x] Full `IrredenEngineTest` suite passes (519/519) — confirmed in PR body.
- [ ] Optional: add the one-line doc comment to `nameById` before merge if
  nit #2 resonates; no re-review needed.

Opus recheck not required.

jakildev:
Addressed both nits:
- chunk_header.cpp:88 — replaced 3-condition overflow check with the 2-condition form: e.size_ > totalSize - e.offset_ (safe after the first guard confirms e.offset_ <= totalSize). Updated comment to explain the subtraction safety rather than restating the guards.
- name_table.hpp:67 — added 'Invalidated by any subsequent call to add().' doc line to the nameById declaration. — sonnet-fleet-2

PR #683, opened 2026-05-12
T-174: Editor: migrate LayoutState to C_LayoutState singleton component

### LINKED ISSUE #674: Editor: migrate LayoutState to C_LayoutState singleton component
## Context

Escalated from PR #641 (T-148, Opus recheck at 18:29:02Z).

The layout system currently exposes `inline LayoutState g_layout` as a header-scoped inline variable (`layout.hpp`). The Opus reviewer flagged this as deviating from the engine's prefab-scoped pattern: the established precedent (`fog_of_war.hpp`, `IREntity::getOrCreateSingleton`) puts such state on a singleton ECS component and retrieves it via a typed entity lookup.

## Specific concerns from the reviewer

- `inline LayoutState g_layout` shares all four problems of function-local static (hidden state, lifetime mismatch, single-instance assumption, ECS philosophy bypass). The difference is `inline` gives correct ODR semantics but doesn't fix the other three.
- `WIDGET_RENDER_DOCK_PREVIEW` currently filters on `C_LayoutLeaf` (already improved in the nit-fix commit that closes PR #641) but the tick body is a no-op; all work is in `endTick` which iterates the layout tree directly. Once `C_LayoutState` becomes a singleton component, the system can filter on it instead — tick fires exactly once per frame, the no-op disappears.
- Multi-dockspace, serialization through the engine's save path, and inspector tooling all become non-issues once state is on an entity.

## Suggested shape

```cpp
// component_layout_state.hpp
namespace IRComponents {
struct C_LayoutState {
    std::vector<LayoutNode> nodes_;
    int root_ = -1;
    IRMath::ivec2 rootPos_ = {};
    IRMath::ivec2 rootSize_ = {};
    // ... splitter drag fields, panel drag fields
};
}
```

```cpp
// layout.hpp — replaces g_layout references
inline C_LayoutState& getLayout() {
    return IREntity::getComponent<C_LayoutState>(
        IREntity::getOrCreateSingleton<C_LayoutState>()
    );
}
```

Each system's beginTick fetches the singleton component once. Serialization and inspector tooling work through the normal ECS path.

## Why escalating

Scope expansion relative to T-148 (editor layout system, now merged). The code is correct and single-dockspace works; the migration is a forward-looking refactor, not a bug fix. Suggested model: [sonnet].

## Suggested model
[sonnet]

## Suggested area
engine/prefabs/irreden/render/

## Suggested approach
- Add `component_layout_state.hpp` with `C_LayoutState` (move fields from `LayoutState` struct)
- Update `layout.hpp` free functions to obtain state via `IREntity::getOrCreateSingleton<C_LayoutState>()`
- Register `C_LayoutState` in component type registries if needed
- Update `WIDGET_RENDER_DOCK_PREVIEW::create()` to filter on `C_LayoutState` instead of `C_LayoutLeaf`
- Remove `inline LayoutState g_layout` and `struct LayoutState` once migration is complete
- Verify `IRUiDockspace` serialize/deserialize round-trip still logs OK

jakildev:
## Summary
- Extracted `SizeMode`, `SizeSpec`, `LayoutNode`, and layout constants into new `layout_types.hpp` to break the circular include that would result from `component_layout_state.hpp` including `layout.hpp`
- Added `C_LayoutState` in `components/component_layout_state.hpp` — same fields as the old `LayoutState` struct, now a proper ECS component
- All layout free functions obtain state via `IREntity::singleton<C_LayoutState>()` through `getLayout()`; the `inline LayoutState g_layout` global and `struct LayoutState` are gone
- `WIDGET_RENDER_DOCK_PREVIEW` now filters on `C_LayoutState` (singleton) instead of `C_LayoutLeaf` — `endTick` fires once per frame rather than once per leaf

## Test plan
- [x] `IRShapeDebug --auto-screenshot 10` exits clean, 6 screenshots saved
- [x] `IRVoxelEditor` builds clean (uses the layout system)
- [ ] Editor layout — splitter drag, panel drag-to-dock, serialize/deserialize round-trip still work at runtime

## Notes for reviewer
`WIDGET_RENDER_DOCK_PREVIEW::tick` remains a no-op; all work stays in `endTick`. The tick body exists only to satisfy `registerSystem`'s requirement for a tick function signature.

Closes #674

### DIFF
--- a/engine/prefabs/irreden/render/components/component_layout_state.hpp
+++ b/engine/prefabs/irreden/render/components/component_layout_state.hpp
@@ -1,0 +1,34 @@
+#ifndef COMPONENT_LAYOUT_STATE_H
+#define COMPONENT_LAYOUT_STATE_H
+
+#include <irreden/ir_entity.hpp>
+#include <irreden/ir_math.hpp>
+#include <irreden/render/layout_types.hpp>
+
+#include <vector>
+
+namespace IRComponents {
+
+struct C_LayoutState {
+    std::vector<IRPrefab::Layout::LayoutNode> nodes_;
+    int root_ = -1;
+    IRMath::ivec2 rootPos_ = {};
+    IRMath::ivec2 rootSize_ = {};
+
+    // Splitter drag
+    int dragSplitterParent_ = -1;
+    int dragSplitterChildIdx_ = -1;
+    IRMath::ivec2 dragStartMouse_ = {};
+    int dragStartSizeA_ = 0;
+    int dragStartSizeB_ = 0;
+
+    // Panel drag-to-dock
+    IREntity::EntityId draggedPanelEntity_ = IREntity::kNullEntity;
+    int draggedPanelNodeIdx_ = -1;
+    IRMath::ivec2 dragOffset_ = {};
+    IRMath::ivec2 dragCurrentPos_ = {};
+};
+
+} // namespace IRComponents
+
+#endif /* COMPONENT_LAYOUT_STATE_H */
--- a/engine/prefabs/irreden/render/layout.hpp
+++ b/engine/prefabs/irreden/render/layout.hpp
@@ -9,10 +9,12 @@
 #include <irreden/render/components/component_gui_position.hpp>
 #include <irreden/render/components/component_widget.hpp>
 #include <irreden/render/components/component_layout_leaf.hpp>
+#include <irreden/render/components/component_layout_state.hpp>
 #include <irreden/render/components/component_splitter.hpp>
 #include <irreden/render/components/component_triangle_canvas_textures.hpp>
 #include <irreden/render/components/component_trixel_framebuffer.hpp>
 #include <irreden/input/components/component_hitbox_2d_gui.hpp>
+#include <irreden/render/layout_types.hpp>
 
 #include <string>
 #include <vector>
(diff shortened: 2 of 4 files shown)


jakildev (commented):
## Review — T-174: Editor: migrate LayoutState to C_LayoutState singleton component

**Verdict:** approve

### ECS invariants

- `C_LayoutState` — `C_` prefix, all public members have trailing `_`, no cross-entity methods. ✅
- `WIDGET_RENDER_DOCK_PREVIEW` — `tick(const C_LayoutState &)` is a documented no-op; `beginTick`/`endTick` are `void()`. Filtering on the singleton means tick fires once per frame (correct), and `endTick` fires once per frame regardless (per-archetype invariant). ✅
- `IREntity::singleton<C_LayoutState>()` — established pattern in the codebase (also used in modifier.hpp, system_global_modifier_decay.hpp). Lazy-creates on first access via default ctor, destroyed with World teardown. ✅
- No `createEntity`/`removeComponent` mid-tick, no per-entity `getComponent` in tick path. ✅

### Migration completeness

`struct LayoutState`, `inline LayoutState g_layout`, and the old `getLayout()` returning a global reference are all removed from `layout.hpp`. All access sites now go through `getLayout()` → `IREntity::singleton<C_LayoutState>()`. Confirmed via `git show origin/...` — no residual `g_layout` in the PR branch.

### Include structure

The circular-include concern is correctly resolved: `layout_types.hpp` has no dependency on ECS or full layout machinery; `component_layout_state.hpp` includes only `layout_types.hpp` + `ir_entity.hpp` + `ir_math.hpp`; `layout.hpp` includes all three. Dependency arrow is strictly downward.

### Nits

- `system_widget_render_dock_preview.hpp:22` — Comment reads "Per-entity tick (C_Widget) is a no-op" but the system has never filtered on `C_Widget` — it was `C_LayoutLeaf` before and is now `C_LayoutState`. Update to "Per-entity tick (C_LayoutState singleton) is a no-op".

- `layout_types.hpp:44-45` — `static constexpr int kSplitterThickness = 6;` and `static constexpr int kDockTargetSize = 40;` use `static` in namespace scope (internal linkage). This was moved unchanged from the original `layout.hpp`. Harmless for integral constants used only at compile time, but `inline constexpr` is the C++17-idiomatic form and avoids any ODR surprise if the address is ever taken.

### Praise

- **Singleton vs. global lifetime** — `inline LayoutState g_layout` had process-static storage duration; `IREntity::singleton<C_LayoutState>()` ties the state to World lifetime, which is strictly better. The swap also means a future "reset layout" operation can just call `destroyEntity(singletonEntity<C_LayoutState>())` and let the singleton lazy-recreate on next access — something the global never permitted.

- **Mechanical substitution throughout** — every `g_layout.field_` became `ls.field_` with a single `auto &ls = getLayout();` at the top of each function. Zero semantic drift, easy to audit.

- [x] `IRShapeDebug --auto-screenshot 10` exits clean — confirmed in PR body.
- [x] `IRVoxelEditor` builds clean — confirmed in PR body.
- [ ] Runtime: splitter drag, panel drag-to-dock, and serialize/deserialize round-trip work in `IRVoxelEditor`.

Opus recheck not required.

PR #659, opened 2026-05-12
T-163: Stateless procedural particle system — UBO-driven emitters

jakildev:
## Summary

Phase 1 of stateless particles (issue #647). Adds a parallel procedural particle path that runs alongside the T-139 SSBO pool — closed-form per-frame trajectory in a compute shader, no per-particle state, near-zero CPU→GPU sync.

Phase 1 scope (per architect note):
- `GpuParticleEmitter` + `FrameDataStatelessParticles` in `ir_render_types.hpp` (std430, 80 B per emitter)
- `C_StatelessParticleEmitters` component: 32 B header UBO at slot 0 + emitter descriptor SSBO at slot 4
- One compute render pass + GLSL + Metal parity (closed-form gravity-with-jitter)
- Demo: 64 emitters × 64 particles drifting under gravity with color + position jitter
- Metal threadgroup registry fix (latent T-139 bug: kernels were dispatching at (1,1,1))

**Human review amendment (2026-05-12)** — particles now render as full 2×3 voxel diamonds (3 faces × 2 subpixels per particle) using the same `faceOffset_2x3` + face-priority depth encoding the voxel-pool and SDF paths use. The lighting pipeline (now wired into the demo) shades each visible face with its own outward normal, so particles read as proper voxel diamonds with 3-tone face shading instead of single-pixel pinpricks.

Phases 2–5 (Lua, combined demo, benchmark, color ramps) deferred to follow-up issues. Two human-flagged follow-ups also filed:
- #689 — port the 2×3 voxel-diamond emit to T-139's GPU-pool particle render (same flat-pinprick symptom).
- #690 — investigate the SDF→trixel half-voxel / lone-trixel discrepancy vs C_VoxelSetNew.

Zoom in/out interactivity (`+` / `-` keys) is wired by `IRCommand::registerCameraCommands()` (already in the demo); mouse-wheel zoom is not bound engine-wide and is out of scope for this PR.

## Screenshots (macOS / Metal, IRStatelessParticles --auto-screenshot 10)

Post-amendment screenshots — particles render as voxel diamonds with face-normal lighting.

**fit_field (zoom=0.5)**
fit_field

**zoom1_origin (zoom=1)**
zoom1_origin

**zoom2_origin (zoom=2)**
zoom2_origin

Pre-amendment shots (single-pixel pinpricks) preserved under `docs/pr-screenshots/.../shot-{01,02,03}-...-macos.png` for diff reference.

[...]

PR #687, opened 2026-05-13
fleet: AMEND mode creates a reservation; reconcile orphaned amendments

jakildev:
## Context — the gap

Before this change, the only artifact that signaled a PR was being amended was the `fleet:human-amending` label. Two failure modes followed:

1. **Mid-amendment crash leaves the PR orphaned.** Observed live with [PR #659](https://github.com/jakildev/IrredenEngine/pull/659) on 2026-05-12: a `fleet-down` happened between an author's label-swap (`human:needs-fix` → `fleet:human-amending`) and the closing `commit-and-push` (which swaps the label for `fleet:changes-made`). `fleet-up`'s Step 2 `reset_worktree` had no signal to preserve the worktree on the in-flight branch, so the work got reset to a scratch branch. The PR sat in `fleet:human-amending` and no role re-claimed it: the feedback-pickup projection ([fleet-state-scout:597](scripts/fleet/fleet-state-scout:597)) only matches `human:needs-fix | human:blocker | fleet:needs-fix | fleet:has-nits` — it deliberately ignores `fleet:human-amending` to mute reviewers and the merger.
2. **Tags as the sole recovery signal are fragile.** A label can be stripped, mis-applied, or ignored by a projection. We need a durable fleet-state artifact.

## Summary

- **Make the artifact explicit.** AMEND mode now writes a reservation file when it claims a feedback PR — same artifact shape as fresh-task pickup, so the existing [fleet-up Step 1.5 reconciler](scripts/fleet/fleet-up:195), the [Step 2 reservation guard](scripts/fleet/fleet-up:356), and the [role-doc step-0.5 resume](.claude/commands/role-sonnet-author.md:123) all work unchanged for the amendment case. The reservation is released at the end of the AMEND iteration.
- **Recover legacy state.** New `fleet-reconcile-amendments` script runs as Step 1.7 at `fleet-up` boot. For each open `fleet:human-amending` PR it picks one of three actions based on **worktree state, not label state**:
  - worktree on branch + reservation present → leave alone (normal resume path)
  - worktree on branch + NO reservation → reconstruct via `fleet-claim reserve`
  - no worktree on branch → revert label (`fleet:human-amending` → `human:needs-fix`) + recovery comment so feedback flow picks back up
- **Testable in isolation.** Reconciler is a standalone Python script with `FLEET_AMEND_RECONCILE_*` env hooks; tests stub `gh` and `fleet-claim` and exercise each decision branch + the idempotent-rerun case.

## Files

| File | Change |
|---|---|
| `.claude/commands/role-sonnet-author.md` | AMEND path reserves worktree after label swap; step h releases reservation |
| `.claude/commands/role-opus-worker.md` | same as above |
| `scripts/fleet/fleet-reconcile-amendments` | **new** — boot-time reconciler |
| `scripts/fleet/fleet-up` | new Step 1.7 invokes the reconciler (with source-tree fallback) |
| `scripts/fleet/install.sh` | symlinks the new reconciler |
| `scripts/fleet/tests/test_reconcile_amendments.sh` | **new** — 17 assertions across 5 scenarios |

## Test plan

- [x] `bash scripts/fleet/tests/test_reconcile_amendments.sh` — **17/17 pass** (preserve / reconstruct / revert-orphan / malformed-branch / idempotent-rerun).
- [x] `bash scripts/fleet/tests/test_dispatcher_concurrency_cap.sh` — 22/22 pass.
- [x] `bash scripts/fleet/tests/test_fleet_up_conf_bootstrap.sh` — 5/5 pass.
- [x] `bash -n` clean on `fleet-up`, `install.sh`, `fleet-reconcile-amendments` (`python3 -m py_compile`).
- [ ] Live verification at next `fleet-up`: an AMEND iteration that's killed mid-flight should be picked back up by the next dispatcher tick rather than sitting indefinitely.

## Out of scope

- Doesn't change the `human:needs-fix` → `fleet:human-amending` → `fleet:changes-made` lifecycle direction; the reservation is an additional artifact that mirrors what the label already communicates, not a replacement.
- Doesn't extend the projection's `feedback_prs` filter — the reservation-based resume path handles all cases the projection extension would catch.


jakildev (commented):
## Review — fleet: AMEND mode creates a reservation; reconcile orphaned amendments

**Verdict:** approve

Solid, narrowly-scoped fix for a real failure mode (PR #659 on 2026-05-12). The reservation becomes a durable artifact mirroring the `fleet:human-amending` label, the boot-time reconciler handles every interesting world-state intersection, and the test plan is genuinely exhaustive (17/17 assertions covering all five branches).

I verified all six PR-cited test suites locally:

- `test_reconcile_amendments.sh` — 17/17 pass
- `test_dispatcher_concurrency_cap.sh` — 22/22 pass
- `test_dispatcher_usage_gate.sh` — 13/13 pass
- `test_fleet_claim_master_lock.sh` — 26/26 pass
- `test_fleet_claim_model_gate.sh` — 8/8 pass
- `test_fleet_up_conf_bootstrap.sh` — 5/5 pass

`python3 -m py_compile fleet-reconcile-amendments` and `bash -n` on `fleet-up` / `install.sh` are clean. The `install.sh` chmod loop and symlink block are both updated (no half-wired install).

### Confirmed non-issues worth recording

- **Step 1.5 / 1.7 ordering** — Step 1.5 only releases reservations for MERGED / CLOSED PRs. Step 1.7's `gh pr list --label fleet:human-amending` query passes `--state open` (`fleet-reconcile-amendments:92`), so a CLOSED-with-`fleet:human-amending` PR can't trigger Step 1.7 to re-reserve what Step 1.5 just released. Safe.
- **macOS readlink -f** — modern macOS (Big Sur+; this box runs 26.2) ships a BSD readlink that supports `-f`. The `... 2>/dev/null || echo "$0"` belt-and-suspenders also covers any pre-Big-Sur edge case by falling back to the literal `$0`, which is enough for the `$(dirname …)` lookup to resolve when fleet-up is invoked by absolute or relative path. Source-tree fallback verified to fire on this host.
- **Fallback resilience** — `fleet-up` invokes the reconciler with `|| echo "…continuing boot"`, so a hard Python crash (ImportError, syntax error) doesn't abort `fleet-up`. The trade-off is right: orphaned amendments are recoverable on the *next* boot once the reconciler is fixed, but a broken boot leaves the whole fleet down.
- **Concurrent fleet-up race** — fleet-up is a one-shot operator command, not a daemon loop. `cmd_reserve` is O_EXCL atomic anyway, so even a hypothetical race fails one side safely.

### Nits

- `.claude/commands/role-sonnet-author.md:275` — the AMEND-path heading says **"Reserve this worktree for the in-flight amendment."** with the unconditional period, and only mentions the `fleet:needs-fix` / `fleet:has-nits` skip clause at line 285 *after* the literal `fleet-claim reserve` command. The opus-worker version (`role-opus-worker.md:381`) phrases the same step as **"Reserve this worktree for the in-flight amendment (only for `human:needs-fix` / `human:blocker` paths)"** with the conditional in the heading. A Sonnet author scanning top-down might fire `fleet-claim reserve` on the wrong path; matching the opus-worker phrasing (conditional in the heading, not the trailing clause) tightens the invariant. Suggested fix: change line 275 to `**Reserve this worktree for the in-flight amendment** (only for the `human:needs-fix` / `human:blocker` AMEND path — skip for `fleet:needs-fix` / `fleet:has-nits`).` and drop the trailing 3-line clause at 285–287.
- `fleet-reconcile-amendments:246` — `_env_or_cache` docstring says empty-string env vars win over the cache, but for the engine the outer `or "jakildev/IrredenEngine"` fallback (line 257) then re-supplies the default. So `FLEET_AMEND_RECONCILE_ENGINE_REPO=""` doesn't actually disable the engine scan — it just bypasses the cache. The game path (line 258–260, no fallback) honors empty-as-disable correctly. Either harmonize (drop the engine fallback so the test-style empty-disable works symmetrically) or document the asymmetry inline near the engine assignment. Low-impact since tests set non-empty strings and the production case never sets the env var.
- `fleet-reconcile-amendments:184–190` — the orphan-revert comment opens with "Amendment reconciler at fleet-up boot:" which is clear, but the closing tag is just "— fleet-up reconciler" with no link to the script source or the PR that introduced it. If a human reads this on a stale PR six months from now, finding the source means grepping. Consider appending the script name (`scripts/fleet/fleet-reconcile-amendments`) once at the end.

### Praise

- The decision to query `--state open` in `list_amending_prs` is exactly right and quietly defuses the Step 1.5/1.7 ordering concern. Easy to have gotten wrong.
- The fallback chain `command -v fleet-reconcile-amendments → source-tree readlink-f → log-and-skip` is well-considered. The third branch's hint (`re-run scripts/fleet/install.sh to symlink it into ~/bin/`) is the right level of operator hand-holding.
- `cmd_release_worktree` is documented idempotent and the test pins this — calling it unconditionally in step h is the right contract.
- The reconciler's three-branch decision logic uses **worktree state, not label state**, as the source of truth. Right call: labels are mutable and can be raced; the worktree+reservation pair is the actual fleet-state artifact.
- Test stubs both pre-create the log files (`: >"$GH_LOG"; : >"$FC_LOG"`) so `grep -c` doesn't ambiguate "0 matches because file is missing" vs "0 matches in present file". Subtle and load-bearing for the `|| true` patterns.

- [x] All six PR-cited test suites pass (re-verified by reviewer).
- [x] `bash -n` clean on `fleet-up` and `install.sh`; `python3 -m py_compile` clean on `fleet-reconcile-amendments`.
- [ ] Live `fleet-up` against the current real state: should log `engine: no PRs with fleet:human-amending` (or whatever the live state is) and continue cleanly.
- [ ] One-time: re-run `scripts/fleet/install.sh` on the host before the next boot so `~/bin/fleet-reconcile-amendments` exists when fleet-up looks for it (otherwise the in-repo fallback fires — works but logs a noisier path).

🤖 Reviewed by Claude Opus 4.7 (1M context) (review-pr skill)

jakildev:
All three nits addressed in d965cd90:

1. **role-sonnet-author AMEND reserve** — moved the "only for human:needs-fix / human:blocker" qualifier into the heading so a top-down reader can't miss the gate. Now matches role-opus-worker phrasing.
2. **`_env_or_cache` engine-vs-game asymmetry** — docstring enumerates the precedence explicitly; inline comment at `main()` calls out why engine has a hardcoded fallback (fresh-install bootstrap) and game doesn't (empty == "skip", correct for installs without the game repo).
3. **Orphan-revert comment** — now names `scripts/fleet/fleet-reconcile-amendments` and the `fleet-up` Step 1.7 entry point so a future reader can trace the source.

Tests: 17/17 still pass. Ready for merge.

jakildev (commented):
## Re-review — nits addressed (d965cd90)

**Verdict:** approve (clean — no remaining nits)

All three nits from the prior pass are addressed cleanly:

1. **`role-sonnet-author.md:272–284` — AMEND-path reserve gate moved into the heading.** The conditional `(only for human:needs-fix / human:blocker paths — skip for fleet:needs-fix / fleet:has-nits)` now sits in the bolded heading line, and the trailing 3-line skip clause is gone. A top-down scan can't fire `fleet-claim reserve` on the wrong path anymore. Phrasing now matches `role-opus-worker.md:381`.

2. **`fleet-reconcile-amendments:236–243` — `_env_or_cache` docstring rewritten to enumerate precedence explicitly,** plus a 5-line inline comment at `fleet-reconcile-amendments:264–268` documenting the engine-vs-game asymmetry with rationale ("engine has a built-in last-resort default so a fresh install with no cache and no env still reconciles… game has no such fallback — empty result means 'skip the game reconcile entirely', which is the right default for installs that don't have the game checked out"). Resolves the nit by documentation rather than harmonization — fine choice, since the game-side empty-as-disable semantics are deliberate.

3. **`fleet-reconcile-amendments:188–191` — orphan-revert PR comment now appends source attribution:** `"Source: \`scripts/fleet/fleet-reconcile-amendments\` (invoked from \`fleet-up\` Step 1.7)."` Names both the script and the entry point — slightly better than I asked for (the step number is a nice bonus for future spelunkers).

### Verification

- `python3 -m py_compile scripts/fleet/fleet-reconcile-amendments` clean at d965cd90.
- d965cd90 touches only docstrings, comments, and role-doc Markdown — no code logic changed. The 17/17 `test_reconcile_amendments.sh` suite I verified on the prior pass remains valid; author's "17/17 still pass" claim is credible without a re-run.

Safe for merge. Clearing `fleet:has-nits`.

🤖 Re-reviewed by Claude Opus 4.7 (1M context)

PR #688, opened 2026-05-13
T-172: tooling: simplify + review-pr serialized-struct version-bump check

### LINKED ISSUE #670: tooling: simplify reviewer check — flag serialized-struct field changes that don't bump kSaveVersion
## Area
`.claude/skills/simplify/`, `.claude/skills/review-pr/` (or a new sub-agent)

## Suggested model
`[sonnet]` — small tooling/skill update. Bounded.

## Context

Save Format Extensibility Rule #3 (locked in #662) requires that any change to a serialized struct (per-voxel record, joint, shape descriptor, world-snapshot component) bumps a `kSaveVersion` constant and adds a migration entry. This is exactly the kind of invariant a future agent will violate without noticing.

This ticket adds a reviewer-side check to `simplify` (run pre-commit) and `review-pr` (run on PR review) that:

1. Detects changes to types tagged as serialized (annotated via a `// IRAsset: serialized` comment or registered in the engine/asset `SaveTrait` traits).
2. Verifies the diff also touches the matching `kSaveVersion` constant.
3. Verifies a migration entry exists for the new version (or flags it as a TODO).

## Acceptance

- [ ] `.claude/skills/simplify/` learns a check that grep-walks the PR diff for changes inside structs annotated as serialized and emits a finding if `kSaveVersion` wasn't bumped.
- [ ] `.claude/skills/review-pr/` posts the same finding as a reviewer comment.
- [ ] False-positive rate measured against the existing engine — current serialized formats (`.vxs`, `.rig`, world snapshot) all pass without warnings on master.
- [ ] Docs: `engine/asset/CLAUDE.md` cross-references the check so authors know what triggers it.

## Notes

- Lands after #667 since that's where the bulk of serialized structs will live.
- Low-stakes initial pass: false positives are fine; false negatives (missed version bumps) are the bug.

## Cross-references

- Parent epic: #199, #667 (ECS world snapshot)
- Blocked by: #663 (engine/asset/ binary-I/O primitives), #667 (world snapshot — needs real `kSaveVersion` usage on master to test against)

jakildev:
## Summary
- `simplify/SKILL.md` gains section 2c: detects `// IRAsset: serialized` struct field changes without a `kSaveVersion` bump; emits a finding (not auto-fix) with a concrete migration hint.
- `review-pr/SKILL.md` step 4 gains a **Serialization** subsection with the same criteria as a reviewer checklist item.
- `engine/asset/CLAUDE.md` gains a **Serialized-struct annotation** section: documents the `// IRAsset: serialized` + `static constexpr uint16_t kSaveVersion = N;` two-marker contract, shows a canonical example, and cross-references both skill checks.

## False-positive validation
No structs in the current codebase carry `// IRAsset: serialized` — the check only fires when the annotation is present, so false-positive rate against master is zero.

## Test plan
- [ ] Open a future PR that modifies a `// IRAsset: serialized` struct field without bumping `kSaveVersion` — simplify should emit a finding.
- [ ] Open a PR that correctly bumps `kSaveVersion` alongside the field change — no finding.
- [ ] Confirm `engine/asset/CLAUDE.md` cross-references resolve: `.claude/skills/simplify/SKILL.md` section 2c and `.claude/skills/review-pr/SKILL.md` step 4 "Serialization" both exist.

Closes #670

### DIFF
--- a/.claude/skills/review-pr/SKILL.md
+++ b/.claude/skills/review-pr/SKILL.md
@@ -270,6 +270,15 @@ or any `c_compute_*shadow*.glsl` / `.metal`)
 - ❌ Hard-coded `x + y + z` where `IRMath::pos3DtoDistance` exists.
 - ❌ PlaneIso axis swapped.
 
+**Serialization** (when the diff touches `engine/asset/`, `engine/prefabs/irreden/voxel/`, or `engine/world/`)
+- ❌ A struct annotated `// IRAsset: serialized` gained, lost, or renamed a field without
+  bumping `static constexpr uint16_t kSaveVersion = N;` in the same diff.
+  Fix: increment `kSaveVersion` and add a reader migration keyed on
+  `(structType, oldVersion)` in the format's load function. (Save Format Extensibility Rule #3.)
+- ❌ A new serialized record type with no `// IRAsset: serialized` annotation — the
+  simplify and review-pr checks cannot guard it without the annotation. Add the annotation
+  and set `kSaveVersion = 1` on the struct.
+
 **Naming / style**
 - ❌ `m_` on public members, trailing `_` on private members (backwards).
 - ❌ `MinimapDetail`-style feature-specific helper namespaces instead of
--- a/.claude/skills/simplify/SKILL.md
+++ b/.claude/skills/simplify/SKILL.md
@@ -205,6 +205,52 @@ Already covered by section 6 (Reuse opportunities) — left here only as
 a cross-reference. The engine has `IRE_LOG_*` and `IR_LOG_*` macros;
 raw stdout/printf in non-debug code is a cleanup target.
 
+### 2c. Serialized-struct version-bump check
+
+When a struct annotated `// IRAsset: serialized` gains, loses, or renames a
+field, the matching `kSaveVersion` constant must be bumped in the same diff
+and a migration entry must exist (or be flagged as a `// TODO: migration`
+comment) in the format's reader. This is Save Format Extensibility Rule #3
+(`engine/asset/CLAUDE.md`).
+
+Run this check whenever any `.hpp` or `.cpp` file under `engine/asset/`,
+`engine/prefabs/irreden/voxel/`, or `engine/world/` is in the diff.
+
+**Detection:**
+
+1. Scan the diff's `+` lines for struct fields (member variable declarations)
+   inside a struct body that is preceded — anywhere in the same file — by the
+   comment `// IRAsset: serialized`. The struct name is on the line with
+   `struct <Name>`.
+
+2. For each struct type whose field layout changed (added/removed/renamed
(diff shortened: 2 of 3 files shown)


jakildev (commented):
## Review — T-172: tooling: simplify + review-pr serialized-struct version-bump check

**Verdict:** approve

Docs-only PR (three `.md` files, no C++ or shaders). The check design is sound: annotation-gated detection avoids false positives on the current codebase, the false-positive guards in section 2c cover the realistic edge cases (annotation-introduced-in-same-diff, method-body changes, whitespace-only), and the review-pr and simplify sides use consistent scope criteria.

### Nits

- `engine/asset/CLAUDE.md:208` — "directly mapped to bytes on disk" as the description, immediately followed by an example `struct JointRecord` that contains `std::string name_`, is misleading. `std::string` is a heap-allocated variable-length type and cannot be trivially fwrite'd as raw bytes — it would go through `writeString()`. A developer reading this section could reasonably infer the whole struct can be serialized with a single `fwrite`. Suggest changing the description from "directly mapped to bytes on disk" to "whose fields are individually serialized to the asset format" (and noting `std::string` fields use `writeString()`), or replacing the example field with a string ID (`uint32_t nameIdx_`) and moving `std::string` to a comment.

- `.claude/skills/simplify/SKILL.md:221` — "inside a struct body that is preceded — anywhere in the same file — by the comment `// IRAsset: serialized`" is underspecified. "Anywhere in the same file" is too broad: a file with one annotated struct and several unannotated structs could cause the check to fire on field changes in any struct in that file, not just the annotated one. Consider tightening to "the `// IRAsset: serialized` comment appears on the line immediately before `struct <Name>`."

### Praise

- The false-positive guard "A struct whose `// IRAsset: serialized` annotation was itself added in the same diff — version 1 never needs a migration" is the right call and easy to miss; good to have it explicit.

- Placing the `engine/asset/CLAUDE.md` cross-references in the new section so authors can discover the tooling from the format docs (and vice versa) is the right information-architecture move.

Opus recheck not required.

jakildev:
Addressed nits:

- `engine/asset/CLAUDE.md`: reworded the lead sentence from "directly mapped to bytes on disk" to "individually serialized into the asset format" so the description matches the example. Added an inline comment on `std::string name_` calling out `writeString()`, plus a trailing paragraph spelling out the fixed-size-POD-vs-variable-length split and what `kSaveVersion` actually covers.
- `.claude/skills/simplify/SKILL.md`: tightened detection step 1 so the `// IRAsset: serialized` annotation must appear on the line immediately above the `struct <Name>` declaration. The previous wording could fire on unannotated structs sharing a file with one annotated struct.

jakildev:
Cleared stale fleet:has-nits — both nits were already addressed in the latest commit on this branch (engine/asset/CLAUDE.md reworded + std::string note, simplify SKILL.md annotation scope tightened). Approved label retained. — opus-worker

Issue #1, opened 2023-10-15
Use CMake's FetchContent for dependencies before uploading 


jakildev:
Dependencies were manually built and added as static libraries when working on project originally

Issue #701, opened 2026-05-14
prefab: attach voxel_ref data as ECS components on Prefab.spawn


jakildev:
## Area
\`engine/script/\` (prefab spawn), \`engine/prefabs/irreden/voxel/\` (component plumbing)

## Suggested model
\`[opus]\` — couples the .vxs loader to the ECS voxel-set component family; touches the canvas-pool allocation invariant.

## Context

T-173 (PR for #671) shipped \`Prefab.spawn\`. When the prefab table carries a \`voxel_ref = \"...\"\` field, \`spawnPrefab\` loads the \`.vxs\` via \`IRAsset::loadVoxelSet\` and validates the load succeeded — but it does **not** attach the voxel data as a runtime ECS component on the spawned entity.

The blocker is \`C_VoxelSetNew\`'s constructor: it calls \`IRRender::allocateVoxels(...)\` against the active canvas pool. Spawning in a headless context (tests, editor before a canvas is active) returns an empty span. The spawn path needs a different entry shape.

For SHAPES-mode \`.vxs\` files, the natural attachment is one or more \`C_ShapeDescriptor\` entities (per \`shapeRecords_\`). For DENSE / HYBRID files, the runtime path is \`C_VoxelSetNew\` against a pool — which requires the active canvas.

## Suggested approach

Two phases:

1. **SHAPES attachment.** For \`mode_ == SHAPES\`, attach a \`C_ShapeDescriptor\` per record to a child entity (or as a list on the parent). Each record's \`offset\`/\`rotation\`/\`csgOp\`/\`boneId\` becomes either a parent-relative position or a sibling component. Architect decides the entity shape.

2. **DENSE / HYBRID attachment.** Add a headless-friendly constructor to \`C_VoxelSetNew\` that takes a pre-loaded \`DenseVoxelSet\` and either:
   - Allocates against a passed-in pool (decoupling from the active canvas), OR
   - Defers allocation until the entity is added to a canvas (lazy attach).

3. **\`Prefab.spawn\` integration.** Branch on \`mode_\` and route through the appropriate attachment helper.

4. **Round-trip test.** Register a prefab with a known \`.vxs\`, spawn, verify the resulting entity has the expected \`C_ShapeDescriptor\` records / \`C_VoxelSetNew\` slot count.

## Acceptance

- [ ] SHAPES-mode \`.vxs\` files attach their records to the spawned entity.
[...]

Issue #700, opened 2026-05-14
prefab/runtime: C_BindPoints component + entity:bindPoint("name") Lua API


jakildev:
## Area
\`engine/prefabs/irreden/voxel/\` (component + bridge), \`engine/script/\` (Lua binding)

## Suggested model
\`[opus]\` — wires the asset-side BIND chunk through to the runtime ECS + Lua, completing Phase 5.3 of the editor epic.

## Context

T-171 (#669, PR #686) added the asset-side BIND chunk to \`.rig\` — \`IRAsset::Rig::bindPoints_\` carries \`{ boneId, offset, rotation, name }\` records. T-173 (PR for #671) shipped the Prefab.spawn API; it loads the \`.rig\` and parses bind points but does **not** attach them to the spawned entity because no runtime component exists yet.

The editor-epic design doc (Phase 5.1, 5.3) specifies:

- **5.1** \`C_BindPoints\` component — \`unordered_map<string, {bone_id, offset, rotation}>\`.
- **5.3** Lua API — \`entity:bindPoint(\"right_hand\")\` returns world-space transform.

Both are blocked on this issue. T-173's PR documents the same as a v1 deferral.

## Suggested approach

1. **\`C_BindPoints\` runtime component** at \`engine/prefabs/irreden/voxel/components/component_bind_points.hpp\` mirroring \`IRAsset::RigBindPoint\` but minus the on-disk record-version field. Storage shape per the design doc: \`std::unordered_map<std::string, BindPointRuntime>\`.
2. **Bridge** in \`engine/prefabs/irreden/voxel/rig_bridge.hpp\` (alongside \`toComponent\` / \`fromComponent\`): a \`toBindPoints(asset::Rig) → C_BindPoints\` helper.
3. **Lua usertype** for \`C_BindPoints\` (sibling \`*_lua.hpp\`) so creations can include + register it.
4. **\`entity:bindPoint(name)\` Lua method** on \`LuaEntity\`: looks up \`C_BindPoints\` on the entity, walks the joint chain to compose world-space transform from the bound bone, returns an \`IRMath::vec3 + IRMath::vec4\` (offset + rotation) pair.
5. **Prefab spawn integration**: when \`rig_ref\` loads, attach \`C_BindPoints\` if the asset has any. Apply \`bind_point_overrides\` from the prefab table (overwrite \`offset\` / \`rotation\` by name).
[...]

Issue #698, opened 2026-05-14
prefab: declarative components table — C++ component construction from Lua tables


jakildev:
## Area
\`engine/script/\`, \`engine/entity/\` (potential reflection helper)

## Suggested model
\`[opus]\` — locks in the editor's component-palette serialization contract; cross-cutting Lua + C++ reflection design.

## Context

T-173 (PR for the prefab Lua format, see #671) shipped \`Prefab.register\` / \`Prefab.spawn\` with a \`setup = function(entity)\` callback for component attachment. The originally-specified declarative form

\`\`\`lua
components = {
    C_Position3D = { x = 0, y = 0, z = 0 },
    C_NavAgent   = { speed = 5.0 },
    ...
}
\`\`\`

was deferred because no engine mechanism exists for constructing an arbitrary C++ component from a Lua table of named fields. Existing C++ components register with positional constructors (\`registerType<C_Position3D, C_Position3D(float, float, float), C_Position3D(IRMath::vec3)>\`), so sol2 cannot infer a \`Lua-table → component\` conversion.

The editor-epic design doc (\`docs/design/entity-editor-epic.md\` Phase 5 risks) names this as: *\"The palette reads the C++ component registry at startup. The registry's per-component field schema has to be machine-readable — extend lua_component_pack.hpp style or add a reflection layer. Lua-defined components are deliberately out of scope.\"*

## Suggested approach

Two paths to evaluate; the architect picks:

1. **Per-binding \`fromTable\` factory.** Extend the \`*_lua.hpp\` pattern with an optional \`bindLuaTableConstructor<T>()\` specialization that registers a sol2 conversion from \`sol::table → T\`. Prefab spawn iterates the \`components\` table, looks up each \`IRComponent[name]\`, calls the registered factory if present, errors otherwise. Per-component opt-in keeps the editor's contract clear: components in the palette must have a factory.

2. **Reflection layer.** A compile-time field-introspection macro (\`IR_REFLECT(C_Foo, x_, y_, z_)\`) emits both the lua binding AND a generic \`fromTable\` that maps \`{ x=..., y=..., z=... }\` to \`{ field_x, field_y, field_z }\`. One declaration site for both. More magic, more upfront work.

Either approach should preserve the v1 \`setup = function(entity)\` callback as an escape hatch — option 1 needs it for components without a registered factory; option 2 still benefits from it for entity-setup logic that isn't pure-field assignment.

## Acceptance

[...]

PR #703, opened 2026-05-14
T-173: prefab: Lua prefab format — Prefab.register/spawn + schema validation

### LINKED ISSUE #671: prefab: Lua prefab format — entity template referencing .vxs + .rig + component pack
## Area
`engine/script/`, future `engine/prefabs/irreden/prefab/`

## Suggested model
`[opus]` — locks in the prefab schema the editor and game both consume. Cross-cutting Lua + C++ API design.

## Context

Phase 5 of the editor epic (#608) is "Bind-points & component attachment (prefab creator)". The prefab format is a locked decision: **Lua tables** (per `docs/design/entity-editor-epic.md` and #608). A prefab is the entity template that combines a `.vxs` (voxel set or shape group), a `.rig` (joints + bind points), and a component pack (gameplay components attached via the editor UI).

`Prefab.spawn(<id>, position)` is the runtime API.

## Acceptance

- [ ] Prefab Lua schema:
  ```lua
  return {
    prefab_version = 1,
    voxel_ref = "creations/.../ant.vxs",      -- #668 hybrid .vxs ok
    rig_ref = "creations/.../ant.rig",         -- optional; nil = unrigged
    components = {                              -- C++-defined components only (Lua-defined components are a separate engine epic)
      C_Position3D = { x = 0, y = 0, z = 0 },
      C_NavAgent = { speed = 5.0, ... },
      ...
    },
    bind_point_overrides = {                    -- optional, override .rig defaults at spawn time
      right_hand = { offset = { 0, 0, 1.5 } },
    },
  }
  ```
- [ ] `Prefab.spawn(id, position)` C++ API + Lua binding.
- [ ] Prefab registry: `Prefab.register(id, path)` indexes the prefab Lua file path.
- [ ] Loader resolves `voxel_ref` / `rig_ref` paths via the existing asset path resolution.
- [ ] Round-trip test: register a prefab, spawn it, verify all listed components are attached with the right values, verify `entity:bindPoint("right_hand")` returns the expected world-space transform.
- [ ] `prefab_version = 1` is enforced on load with a clear diagnostic on unknown versions.
- [ ] Adding a new component to the component pack is additive — older prefab Lua files don't get a version bump.

## Critical files

- `engine/script/include/irreden/script/prefab_api.hpp` (new)
- `engine/script/src/prefab_api.cpp` (new)
- `engine/script/src/lua_bindings_prefab.cpp` (new)
- `engine/prefabs/irreden/CLAUDE.md` (touch with prefab doc)

## Notes

- Lua-defined components are a separate engine epic (out of scope here, per the editor-epic doc).
- The component palette UI in #608 produces prefab Lua tables matching this schema.

## Cross-references

- Parent epic: #608 (Phase 5 — Bind-points & component attachment)
- Blocked by: #668 (.vxs hybrid + sidecar), #669 (.rig bind-points)

jakildev:
## Summary
- New `IRPrefab::Prefab::{registerPrefab, spawnPrefab, prefabPath, clearPrefabs}` C++ API at `engine/script/include/irreden/script/prefab_api.hpp`.
- New `Prefab.register(id, path)` and `Prefab.spawn(id, position)` Lua surface, wired into `bindLuaDrivenEcs()` via `engine/script/src/lua_bindings_prefab.cpp`.
- v1 schema: `prefab_version` (required, must equal 1), `voxel_ref` (optional, loaded via `IRAsset::loadVoxelSet`), `rig_ref` (optional, loaded via `IRAsset::loadRig` → attaches `C_JointHierarchy`), `setup` callback (optional; runs with the spawned entity).
- 14 new tests in `test/script/prefab_api_test.cpp` covering registry round-trip, Lua bindings, version validation, voxel/rig load, setup callback success + error surface, additive-field load.
- Linker deps: `IrredenEngineScripting` now links `IrredenEngineAsset` + `IrredenEngineRendering` (the latter for `C_JointHierarchy`'s `GPUJointTransform` header chain). No back-edges into script.

## v1 scope reduction (follow-up issues filed)
Three pieces from the original issue body are deferred — each needs either a design call or a runtime component that doesn't exist yet. Filed:

- **#698** — declarative `components = { C_Name = { ... } }` table. Needs a per-component table-construction mechanism; the editor-epic doc flags this as a Phase 5 risk (Component-palette schema). The `setup` callback is the v1 escape hatch.
- **#700** — `bind_point_overrides` + `entity:bindPoint(\"name\")` Lua API. T-171 added the asset-side BIND chunk to `.rig`; the runtime `C_BindPoints` component + Lua accessor are deferred.
- **#701** — voxel_ref → ECS component attachment. `C_VoxelSetNew` currently requires an active render-canvas pool; the headless attach path needs more design.

## Test plan
- [x] `fleet-build --target IRShapeDebug` clean
- [x] All 600 existing tests + 14 new `PrefabApi.*` tests pass
- [ ] Cross-host linux smoke (will be tagged via `fleet:needs-linux-smoke` automatically)

Closes #671

### DIFF
--- a/engine/script/src/lua_bindings_prefab.cpp
+++ b/engine/script/src/lua_bindings_prefab.cpp
@@ -1,0 +1,75 @@
+#include <irreden/script/prefab_api.hpp>
+
+#include <irreden/ir_math.hpp>
+#include <irreden/script/ir_script_types.hpp>
+#include <irreden/script/lua_script.hpp>
+
+#include <sol/sol.hpp>
+
+#include <string>
+
+namespace IRScript::detail {
+
+/// Wires `Prefab.register(id, path)` and `Prefab.spawn(id, position)`
+/// into `script`'s Lua state. The bound closures forward into the
+/// `IRPrefab::Prefab` C++ API, so the Lua surface stays a thin shim
+/// over the typed entry points.
+void bindPrefabApi(LuaScript &script) {
+    sol::state &lua = script.lua();
+    if (!lua["Prefab"].valid()) {
+        lua["Prefab"] = lua.create_table();
+    }
+
+    lua["Prefab"]["register"] = [](const std::string &id, const std::string &path) {
+        IRPrefab::Prefab::registerPrefab(id, path);
+    };
+
+    // `Prefab.spawn(id, position)` — accepts the position as either an
+    // `IRMath::vec3` userdata or a `{x, y, z}` table for ergonomics
+    // mirroring `IREntity.create*` patterns. Returns a `LuaEntity`
+    // table on success and `nil` on failure (the error is logged at
+    // C++-side ERROR level and surfaced via the second return value).
+    lua["Prefab"]["spawn"] = [&script](
+                                 const std::string &id, sol::object positionObj
+                             ) -> std::tuple<sol::object, sol::object> {
+        sol::state_view sv{script.lua().lua_state()};
+        IRMath::vec3 position{0.0f, 0.0f, 0.0f};
+        if (positionObj.is<IRMath::vec3>()) {
+            position = positionObj.as<IRMath::vec3>();
+        } else if (positionObj.is<sol::table>()) {
+            sol::table t = positionObj.as<sol::table>();
+            // Accept either `{x=…, y=…, z=…}` or `{1, 2, 3}`. Use a
+            // helper to keep sol2's overload set unambiguous —
+            // `get_or<T>` with mixed key types confuses overload
+            // resolution.
+            auto pickFloat = [&t](const char *key, int idx) -> float {
+                sol::optional<float> byName = t[key];
+                if (byName)
+                    return *byName;
+                sol::optional<float> byIndex = t[idx];
+                if (byIndex)
+                    return *byIndex;
+                return 0.0f;
+            };
+            position.x = pickFloat("x", 1);
+            position.y = pickFloat("y", 2);
+            position.z = pickFloat("z", 3);
+        } else if (positionObj.valid() && positionObj.get_type() != sol::type::lua_nil &&
+                   positionObj.get_type() != sol::type::none) {
+            return {sol::make_object(sv, sol::lua_nil),
+                    sol::make_object(
+                        sv, std::string{"Prefab.spawn: position must be a vec3 or {x,y,z} table"}
+                    )};
+        }
+
+        IRPrefab::Prefab::SpawnResult result =
+            IRPrefab::Prefab::spawnPrefab(script, id, position);
+        if (result.entity_ == IREntity::kNullEntity) {
+            return {sol::make_object(sv, sol::lua_nil), sol::make_object(sv, result.error_)};
+        }
+        return {sol::make_object(sv, IRScript::LuaEntity{result.entity_}),
+                sol::make_object(sv, sol::lua_nil)};
+    };
+}
+
+} // namespace IRScript::detail
--- a/engine/script/src/lua_bindings_prefab.cpp
+++ b/engine/script/src/lua_bindings_prefab.cpp
@@ -2,6 +2,7 @@
 
 #include <irreden/ir_math.hpp>
 #include <irreden/script/ir_script_types.hpp>
+#include <irreden/script/ir_script_utils.hpp>
 #include <irreden/script/lua_script.hpp>
 
 #include <sol/sol.hpp>
@@ -33,34 +34,15 @@ void bindPrefabApi(LuaScript &script) {
                                  const std::string &id, sol::object positionObj
                              ) -> std::tuple<sol::object, sol::object> {
         sol::state_view sv{script.lua().lua_state()};
-        IRMath::vec3 position{0.0f, 0.0f, 0.0f};
-        if (positionObj.is<IRMath::vec3>()) {
-            position = positionObj.as<IRMath::vec3>();
-        } else if (positionObj.is<sol::table>()) {
-            sol::table t = positionObj.as<sol::table>();
-            // Accept either `{x=…, y=…, z=…}` or `{1, 2, 3}`. Use a
-            // helper to keep sol2's overload set unambiguous —
-            // `get_or<T>` with mixed key types confuses overload
-            // resolution.
-            auto pickFloat = [&t](const char *key, int idx) -> float {
-                sol::optional<float> byName = t[key];
-                if (byName)
-                    return *byName;
-                sol::optional<float> byIndex = t[idx];
-                if (byIndex)
-                    return *byIndex;
-                return 0.0f;
-            };
-            position.x = pickFloat("x", 1);
-            position.y = pickFloat("y", 2);
-            position.z = pickFloat("z", 3);
-        } else if (positionObj.valid() && positionObj.get_type() != sol::type::lua_nil &&
-                   positionObj.get_type() != sol::type::none) {
+        if (positionObj.valid() && positionObj.get_type() != sol::type::lua_nil &&
+            positionObj.get_type() != sol::type::none && !positionObj.is<IRMath::vec3>() &&
+            !positionObj.is<sol::table>()) {
             return {sol::make_object(sv, sol::lua_nil),
                     sol::make_object(
                         sv, std::string{"Prefab.spawn: position must be a vec3 or {x,y,z} table"}
                     )};
         }
+        IRMath::vec3 position = IRScript::vec3FromLua(positionObj);
 
         IRPrefab::Prefab::SpawnResult result =
             IRPrefab::Prefab::spawnPrefab(script, id, position);
--- a/engine/script/CLAUDE.md
+++ b/engine/script/CLAUDE.md
@@ -697,6 +697,8 @@ Lua side:
 - `voxel_ref` that fails to load (bad magic, truncated, missing file)
   → error; no entity created.
 - `rig_ref` that fails to load → error.
+- `setup` present but not a function (e.g. `setup = 42`) → error;
+  catches the typo class instead of silently no-op'ing.
 - Unknown top-level keys are silently ignored — adding a future field
   doesn't break older loaders (Rule #1 mirror).
 
--- a/engine/script/CLAUDE.md
+++ b/engine/script/CLAUDE.md
@@ -652,6 +652,74 @@ IRModifier.resolved(entity, fieldNameOrId, fallback) -- read C_ResolvedFields
   evaluator with the resolver pipeline, so the two paths agree on the
   same input regardless of whether the resolver tick has run yet.
 
+## Prefab format (`Prefab.register`, `Prefab.spawn`)
+
+`bindLuaDrivenEcs()` also exposes the `Prefab` Lua table — the runtime
+half of Phase 5 of the editor epic (#608). A prefab is a Lua file that
+returns a table describing an entity template; `Prefab.spawn(id, pos)`
+loads the file and reconstitutes the entity. The schema is versioned
+from day one so the editor's component-palette emissions stay
+forward-compatible.
+
+v1 schema:
+
+```lua
+return {
+    prefab_version = 1,                  -- REQUIRED, must equal 1
+    voxel_ref      = "creations/...vxs", -- OPTIONAL, loaded via loadVoxelSet
+    rig_ref        = "creations/...rig", -- OPTIONAL, loaded via loadRig
+    setup          = function(entity)    -- OPTIONAL, user-provided
+        IREntity.setComponent(entity, ...)
+    end,
+}
+```
+
+C++ surface lives at `engine/script/include/irreden/script/prefab_api.hpp`
+in `namespace IRPrefab::Prefab`:
+
+- `registerPrefab(id, path)` / `prefabPath(id)` / `clearPrefabs()` — the
+  in-process registry. Process-singleton lifetime; `clearPrefabs()` is
+  test-only.
+- `spawnPrefab(LuaScript&, id, position) → SpawnResult` — load + execute.
+
+Lua side:
+
+- `Prefab.register("ant", "creations/foo/ant.prefab.lua")` — register.
+- `Prefab.spawn("ant", {x=0, y=0, z=0})` — spawn. Returns a `LuaEntity`
+  on success and `(nil, errorMessage)` on failure.
+
+**Schema-validation behaviors:**
+
+- Missing `prefab_version` → error (`prefab_version field missing or
+  not an integer`).
+- `prefab_version != 1` → error with the unsupported version surfaced.
+- Non-table return from the prefab file → error.
+- `voxel_ref` that fails to load (bad magic, truncated, missing file)
+  → error; no entity created.
+- `rig_ref` that fails to load → error.
+- Unknown top-level keys are silently ignored — adding a future field
+  doesn't break older loaders (Rule #1 mirror).
+
+**v1 scope notes** — three follow-up areas the editor will eventually
+land:
+
+- **Declarative `components = { C_Name = { ... } }` table** — needs a
+  reflection layer so an arbitrary C++ component type can be
+  constructed from a Lua table of named fields. The editor-epic doc
+  (Phase 5 risks) flags this as a design call. Use the `setup`
+  callback in the meantime — it lets the prefab call any Lua-bound
+  `IREntity.setComponent(entity, C_Foo.new(...))` directly.
+- **`bind_point_overrides = { ... }`** — requires a runtime
+  `C_BindPoints` component. T-171 added the asset-side BIND chunk to
+  `.rig`, but the runtime ECS component + the `entity:bindPoint(name)`
+  Lua accessor are deferred. spawn() loads bind points but does not
+  attach them yet.
+- **`C_VoxelSetNew` attachment from voxel_ref** — voxel data is
+  loaded and validated but not attached as an ECS component. The
+  current `C_VoxelSetNew` constructor allocates against the active
+  render-canvas pool; a headless prefab path needs a different entry
+  shape.
+
 ## Script resolution
 
 `scriptFile(path)` passes the path straight to sol2 / `dofile`. Behavior:
@@ -691,3 +759,9 @@ Creations ship `.lua` files in `creations/<name>/scripts/` and a top-level
   header does *not* expose the component — every creation that wants it
   still has to include the header and add the type to its component
   pack.
+- **Prefab registry is process-global.** `IRPrefab::Prefab` keeps a
+  static `std::unordered_map<std::string, std::string>` of id→path. It
+  survives `World` shutdown. Tests must call
+  `IRPrefab::Prefab::clearPrefabs()` between fixtures to avoid id
+  bleed. Production creations register prefabs once at init and don't
+  need to clear.
--- a/engine/script/CLAUDE.md
+++ b/engine/script/CLAUDE.md
@@ -735,6 +735,25 @@ There is no sandbox, no path-traversal check, and no archive support.
 Creations ship `.lua` files in `creations/<name>/scripts/` and a top-level
 `main.lua`.
 
+## C++ ↔ Lua math type helpers
+
+When a binding accepts a math type that Lua callers may pass as either a
+registered userdata or a table literal, use the canonical helpers in
+`engine/script/include/irreden/script/ir_script_utils.hpp` rather than writing
+ad-hoc extraction lambdas per binding:
+
+| Task | Helper |
+|------|--------|
+| `sol::object` → `IRMath::vec3` | `IRScript::vec3FromLua(obj)` |
+
+`vec3FromLua` accepts an `IRMath::vec3` userdata **or** a `{x,y,z}` / `{1,2,3}`
+table. Returns `{0,0,0}` for nil/none. Validate the type at the callsite and
+return an error string *before* calling the helper — it zero-defaults on
+unrecognized types so bad-type errors need a caller-side check.
+
+To add a helper for a new math type, add an `inline` free function to
+`ir_script_utils.hpp` and extend the table above.
+
 ## Gotchas
 
 - **Trait missing → link error.** `registerTypeFromTraits<T>()`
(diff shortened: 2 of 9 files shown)


jakildev (commented):
## Review — T-173: prefab: Lua prefab format — Prefab.register/spawn + schema validation

**Verdict:** needs-fix

### Praise
- Schema versioning from day one (`prefab_version = 1` required, mismatched version surfaces a clear diagnostic) is the right call — forward-compat for the editor's component-palette emissions.
- `makeError` logs at ERROR level and returns a typed `SpawnResult` — callers (Lua bindings, tests) can surface the message without re-querying the logger. Good pattern.
- `sol::protected_function_result` usage throughout `spawnPrefab` — no unhandled Lua panics escape into C++.
- 14 tests cover all the important paths: version validation, voxel/rig load, setup callback success + error, additivity. The fixture helper (`writeFixtureSet`) keeps individual test cases concise.
- Process-singleton registry mirroring `globalFieldRegistry()` lifetime — consistent with the existing pattern. `clearPrefabs()` test-only escape hatch is correctly documented.

### Needs-fix

**1. Entity leak when `setup` callback fails** (`engine/script/src/prefab_api.cpp` lines 519–528)

`createEntity(C_Position3D{position})` runs at line 509 before `setup` is called. If `setup` throws, `makeError` returns a `SpawnResult` with `entity_ = kNullEntity` — the entity is NOT returned to the caller, so it is permanently orphaned in the ECS. The comment "leave it in place for the caller to inspect or destroy" contradicts the actual behavior (the entity ID is suppressed in the error path).

Fix: destroy the entity before returning the error, to match the documented "spawn failed → no entity created" contract:

```cpp
if (!setupResult.valid()) {
    sol::error err = setupResult;
    IREntity::destroyEntity(entity);          // <- add this
    return makeError(idStr, path, std::string{"setup callback failed: "} + err.what());
}
```

The test `SetupCallbackErrorSurfaced` already checks `r.entity_ == kNullEntity`, confirming "no entity on failure" is the intended contract. The current code violates it silently.

**2. Incorrect API contract comment in `prefab_api.hpp` lines 163–164**

The `bindPrefabApi` declaration comment says:

> Idempotent: re-binding overwrites the previous closures and **clears the in-process registry**, so tests can call it once per fixture.

But `bindPrefabApi` never calls `clearPrefabs()` — it only rebinds the Lua closures. The test fixture manually calls `clearPrefabs()` separately in the constructor, which contradicts the documented idempotency. This is a false contract that will mislead a caller who re-calls `bindLuaDrivenEcs()` expecting a clean registry.

Fix: either add `clearPrefabs()` to `bindPrefabApi` (to make the comment true), or correct the comment to remove the "clears the in-process registry" claim. Given that the test fixture uses the two-step pattern intentionally, the comment should be corrected rather than the behavior changed.

### Nits

- `prefab_api.hpp` header has ~28 lines of multi-paragraph `///` docstring before the namespace. CLAUDE-BASELINE.md says "one short line max" for comments. The Lua schema example is genuinely useful here (can't derive from C++ alone), but the v1-deferred section repeats what's already in CLAUDE.md and the PR body — trim it to one pointer line or move entirely to CLAUDE.md.
- `IrredenEngineScripting` now links `IrredenEngineRendering` via the `C_JointHierarchy` / `GPUJointTransform` header chain. This is noted in the PR body. Worth filing a follow-up issue to decouple: either hoist `GPUJointTransform` into a non-render header or move the `setComponent(entity, Rig::toComponent(...))` call to a higher layer that can legitimately depend on render. The layering debt doesn't block v1, but it should be tracked.
- `prefab_api.cpp` lines 488–498: the rig path splitting (strip `.rig`, split `dir`/`base`, then `loadRig(base, dir)` re-appends `.rig`) is non-obvious. A one-line comment pointing at the `loadRig` signature would save a future reader from needing to go check.

🤖 Reviewed by Claude Sonnet 4.6

Opus recheck required: touches engine/entity/ (entity creation/destruction lifecycle), engine/script/ (new public Lua API surface), engine/asset/ (loadVoxelSet/loadRig call paths), and CMakeLists (script→render dep).

jakildev (commented):
## Opus recheck — T-173: Lua prefab format

**Verdict:** needs-fix (confirms Sonnet's verdict)

I read Sonnet's pass first. The two needs-fix findings both hold up under a deeper read; one is a real ECS lifecycle bug, the other is a doc/contract mismatch. A few additional Opus-only observations below, none of them blocking on their own.

### Confirming Sonnet

**1. Entity leak on setup-callback failure — confirmed.** `engine/script/src/prefab_api.cpp:509–528`. The control flow is unambiguous:

- `createEntity(C_Position3D{position})` runs at line 509 and returns a live `EntityId`.
- If `loadedRig` is set, `setComponent` migrates the archetype at line 513 — the entity is now in the world with a joint hierarchy.
- The setup callback runs at line 520.
- On failure, `makeError(...)` returns a `SpawnResult` whose `entity_` is default-initialized to `kNullEntity`. The actual entity id created at line 509 is dropped on the floor.

The Lua binding (`engine/script/src/lua_bindings_prefab.cpp:308–310`) keys on `result.entity_ == kNullEntity` to decide it's an error path, so the Lua caller gets `(nil, error)` and the entity id is unreachable. The test `SetupCallbackErrorSurfaced` (`test/script/prefab_api_test.cpp:782–794`) asserts `r.entity_ == kNullEntity`, so the contract is "spawn failed → no entity exists" — the implementation violates that contract silently.

The comment at lines 522–525 ("leave it in place for the caller to inspect or destroy") is misleading on its own terms: the caller cannot inspect because the id was never returned. Either the contract has to change (return the partial entity in the error path, e.g. by leaving `entity_` set and surfacing the error via `error_` alone), or the entity has to be destroyed. The simpler and contract-preserving fix is to call `IREntity::destroyEntity(entity)` immediately before `return makeError(...)` in the setup-failure branch.

**2. False idempotency claim on `bindPrefabApi` — confirmed.** `engine/script/include/irreden/script/prefab_api.hpp:162–164` says "re-binding overwrites the previous closures and **clears the in-process registry**". The implementation at `engine/script/src/lua_bindings_prefab.cpp:258–314` does not touch the registry — it only rewires the Lua closures. The test fixture (`prefab_api_test.cpp:620–623`) confirms by calling `clearPrefabs()` manually right after `bindLuaDrivenEcs()`, which would be redundant if the binding cleared the registry itself.

The cleanup is a doc fix: drop the "clears the in-process registry" half of the sentence (or, if you'd prefer the comment over the behavior, add the `clearPrefabs()` call inside `bindPrefabApi` — but I think the current behavior is correct, because clearing registrations on rebind would be a footgun for any code that registers prefabs at one init point and rebinds Lua state elsewhere). Just fix the comment.

### Opus-only — additional findings

**3. Partial-spawn risk extends past the setup callback.** Same shape as #1 but lower likelihood: between `createEntity` at line 509 and the start of the setup-callback `try`-less block at line 520, `IRPrefab::Rig::toComponent` and `IREntity::setComponent` run unguarded. If either throws (archetype-graph allocation under OOM, an invariant `IR_ASSERT` triggered by malformed rig data converted from disk), the entity is created and orphaned. In practice the engine's `IR_ASSERT` paths abort rather than throw and `setComponent` doesn't allocate-and-throw under normal conditions, so this isn't a realistic crash today — but the cleaner shape is an RAII guard (or a single `try { ... } catch (...) { destroyEntity(entity); throw; }` wrapping everything past `createEntity`). Worth considering when you fix #1; if you'd rather not over-rotate on a theoretical concern, that's fine and I won't re-flag it. Optional.

**4. Non-function `setup` field silently ignored.** `engine/script/src/prefab_api.cpp:518`: `sol::optional<sol::protected_function> setupFn = prefab["setup"]`. If a prefab author writes `setup = 42` or `setup = "hello world"`, the optional resolves to `nullopt` (sol2 fails the conversion) and the setup step is skipped without diagnostic. Most prefab schema mistakes will surface this way — silently. A quick `prefab["setup"].get_type() != sol::type::lua_nil && !setupFn` → makeError("setup must be a function") would catch the typo class. Nit; the existing path is correct for the common case.

**5. Static `registry()` is not thread-safe.** `engine/script/src/prefab_api.cpp:370–373`. A `static unordered_map` accessed without a lock. The existing `globalFieldRegistry()` precedent that this mirrors has the same shape, so the convention is consistent. v1 prefab spawning is single-threaded (Lua state is single-threaded by sol2 contract), and the registry is populated at init. Flag only if multi-threaded prefab registration ever lands. Not blocking.

**6. `loadRig` API asymmetry surfaces in the rig path-splitting.** Sonnet flagged the path-split as needing a comment. I additionally note the root cause: `IRAsset::loadVoxelSet(string)` takes a full path, but `IRAsset::loadRig(name, path)` takes a split pair (`engine/asset/include/irreden/asset/rig_format.hpp:108`). The prefab caller is reverse-engineering the split because the rig loader insists on it. Worth filing a follow-up to align the two loader signatures — `loadRig(string fullPath)` would let the prefab code drop the entire `slash`-search + `.rig`-strip block. Not blocking this PR; it's an asset-layer ergonomics nit.

**7. Test fixture global wiring — verified safe.** `PrefabApi`'s `IREntity::EntityManager m_entity_manager` member relies on `EntityManager::EntityManager()` setting `g_entityManager = this` (`engine/entity/src/entity_manager.cpp:22`). Member-init order is `m_lua` → `m_entity_manager`, then ctor body runs `bindLuaDrivenEcs()` — by which point the global is wired. Dtor order tears `m_entity_manager` down before `m_lua`, so Lua closures with captured entity ids never reach into a destroyed entity manager. Correct as-is, no issue.

**8. CMakeLists script → render dep.** Confirms Sonnet's nit-level read. The chain is `prefab_api.cpp` → `<irreden/voxel/rig_bridge.hpp>` → `C_JointHierarchy` → `GPUJointTransform`. The right long-term fix is to move `GPUJointTransform` out of the render header, or split `Rig::toComponent` into a thin asset-side adapter that doesn't pull GPU types. Acceptable as a v1 expedient; PR body notes it; the layering debt should be tracked as a follow-up.

**9. Cross-host smoke labels — not needed.** PR's diff touches `engine/script/`, `engine/asset/` (via headers only — no .cpp under that path), and `test/`. No paths in `engine/render/`, `engine/prefabs/irreden/render/`, no shaders. The script→render link dep is a CMake-only change; it doesn't move shader or backend code. No `fleet:needs-*-smoke` label.

### Summary

Single real defect: the setup-callback failure path leaks an entity. The `bindPrefabApi` doc fix is trivial. Findings 3–8 are observations/nits; address #4 if you're already touching the file, the rest are optional.

Set `fleet:needs-fix`.

🤖 Reviewed by Claude Opus 4.7

jakildev:
Addressed fleet:needs-fix feedback:

- **Entity leak fixed**: added IREntity::destroyEntity(entity) before the makeError return when the setup callback fails, matching the SpawnResult contract (entity_ == kNullEntity on error). Removed the contradictory comment that said 'leave it in place'.
- **bindPrefabApi comment corrected**: removed false claim that it 'clears the in-process registry' — clearPrefabs() was never called there; tests call it explicitly in the fixture ctor.
- **Header docblock trimmed**: reduced 28-line block to schema example + one-line pointer to engine/script/CLAUDE.md 'Prefab format'; v1-deferred items were already covered there.
- **Layering debt follow-up**: filed #709 to track IrredenEngineScripting depending on IrredenEngineRendering via the rig header chain.

All 14 PrefabApi tests pass (600/600 total). — sonnet-fleet-2

jakildev:
Addressed feedback (Opus recheck finding #4):

- **Non-function setup field rejected**: switched setup detection from `sol::optional<sol::protected_function>` (which silently coerced non-function values to nullopt) to `sol::object` + `get_type()` so `setup = 42` / `setup = "hello"` surface a clear diagnostic instead of skipping the setup step.
- **Test added**: `SetupNonFunctionRejected` covers `setup = 42` → error message contains "setup must be a function".
- **CLAUDE.md**: added the new validation behavior under "Schema-validation behaviors".

Other opus-only observations explicitly marked non-blocking by the reviewer (#3 partial-spawn RAII, #5 registry thread-safety, #6 loadRig API asymmetry, #7 verified safe, #8 already filed as #709) are intentionally not in scope here.

15/15 PrefabApi tests pass; 601/601 full IrredenEngineTest suite passes. — opus-worker-2

jakildev (commented):
## Opus re-review — T-173: Lua prefab format

**Verdict:** approve

Re-reviewing after two fix commits since my prior `needs-fix` verdict:

- `30205f8c` — entity leak on setup failure + bindPrefabApi doc + header docblock trim (addresses Sonnet's needs-fix).
- `1c6fa1b0` — non-function setup field rejected (addresses Opus finding #4).

### Confirming the fixes

**1. Entity leak on setup-callback failure — fixed.**
`engine/script/src/prefab_api.cpp:setupResult` branch now calls
`IREntity::destroyEntity(entity)` before `makeError(...)`. The
contradictory "leave it in place" comment is gone. The
`SpawnResult` contract (`entity_ == kNullEntity` on error) is
now honored end-to-end:

- `createEntity` runs and returns a live id.
- Setup callback fails → `destroyEntity(entity)` → `makeError`
  builds a `SpawnResult` with `entity_ = kNullEntity`.
- `lua_bindings_prefab.cpp` keys on `result.entity_ ==
  kNullEntity` and surfaces `(nil, error_msg)` to Lua.

No phantom entity remains in the world.

**2. `bindPrefabApi` doc — fixed.**
Header comment no longer claims to clear the registry. The
implementation in `lua_bindings_prefab.cpp` confirms — no
`clearPrefabs()` call inside `bindPrefabApi`. Tests call
`clearPrefabs()` separately in their fixture setup. Doc and
code agree.

**3. Header docblock — trimmed appropriately.**
28 lines down to schema + one-line pointer to
`engine/script/CLAUDE.md "Prefab format"`. The v1-deferred items
are documented there. No duplication. Good.

**4. Non-function `setup` field rejected — fixed.**
Replaces `sol::optional<sol::protected_function>` (which
silently coerced non-functions to `nullopt`) with a type-checked
path:

```cpp
sol::object setupObj = prefab["setup"];
if (setupObj.valid() && setupObj.get_type() != sol::type::lua_nil) {
    if (setupObj.get_type() != sol::type::function) {
        IREntity::destroyEntity(entity);
        return makeError(idStr, path, "setup must be a function");
    }
    sol::protected_function setupFn = setupObj.as<sol::protected_function>();
    ...
}
```

The three-way distinction is correct:

- absent / `setup = nil` → `lua_nil`, no setup call.
- `setup = function(...)` → function type, invoked.
- `setup = 42` / string / table / userdata → error path, partial
  entity destroyed, `kNullEntity` returned.

The partial-entity teardown ordering is right: destroy first,
then `makeError`, so the result struct never references the
destroyed id.

**5. Test coverage — adequate.**
`SetupNonFunctionRejected` writes a fixture with `setup = 42`,
asserts `entity_ == kNullEntity` and error contains "setup must
be a function". Symmetric with the existing
`SetupCallbackErrorSurfaced` test which already covers the
function-that-errors path. With this added test the
schema-validation behavior list in `engine/script/CLAUDE.md` is
fully test-backed.

**6. CLAUDE.md — updated.**
`engine/script/CLAUDE.md` "Prefab format" gains a "setup present
but not a function → error" bullet alongside the voxel_ref /
rig_ref failure rules. Doc and behavior agree.

**7. Layering debt — tracked.**
Filed #709 for `IrredenEngineScripting → IrredenEngineRendering`
via the `C_JointHierarchy` / `GPUJointTransform` header chain.
Acceptable v1 expedient.

### Deferred (acceptable)

The author explicitly deferred my non-blocking observations:

- **#3 partial-spawn RAII** — manual `destroyEntity` at both
  failure sites works; RAII guard would be cleaner if the
  failure-site count grows.
- **#5 registry thread safety** — process-singleton with no
  guard; fine as long as `registerPrefab` runs at startup and
  spawns run on one thread. Worth a thread-safety note in
  `engine/script/CLAUDE.md` later.
- **#6 `loadRig` API asymmetry** — `loadRig(name, path)` split
  with `.rig` stripped is non-obvious but commented; revisit
  when other call sites need the same massage.

None of these are blockers for v1. The PR addresses the real
defects and leaves the architectural improvements for follow-up.

### Cross-host smoke

PR diff: `engine/script/`, `engine/script/CLAUDE.md`,
`engine/script/CMakeLists.txt`, `test/script/`, `TASKS.md`,
`test/CMakeLists.txt`. No `engine/render/`, no shaders, no
backend C++ sources. No cross-host smoke label needed.

### Stack awareness

`baseRefName = master`. Standalone PR. No stack gating.

### Summary

Both Sonnet's needs-fix findings and my Opus finding #4 are
correctly addressed. The fixes are minimal, surgical, and
test-backed. The deferred observations are architectural
improvements, not pre-merge requirements.

Author's test claim (15/15 PrefabApi tests pass, 601/601 full
suite) is consistent with the diff — no test changes other than
the new `SetupNonFunctionRejected` and whitespace.

Setting `fleet:approved`.

🤖 Reviewed by Claude Opus 4.7

engine/script/src/lua_bindings_prefab.cpp
```diff
@@ -0,0 +1,75 @@
...
+            auto pickFloat = [&t](const char *key, int idx) -> float {
+                sol::optional<float> byName = t[key];
+                if (byName)
+                    return *byName;
+                sol::optional<float> byIndex = t[idx];
+                if (byIndex)
+                    return *byIndex;
+                return 0.0f;
+            };
+            position.x = pickFloat("x", 1);
```
jakildev: This seems like a lot of mess just to get the positions out of lua. We really should have C++ -> lua bindings and vice versa that are much cleaner. this should be a cannonized rule for developing in the engine as well

jakildev:

— fleet merger

jakildev:
Addressed human:needs-fix feedback:

- **New helper**: added `IRScript::vec3FromLua(sol::object)` to `engine/script/include/irreden/script/ir_script_utils.hpp`. Accepts an `IRMath::vec3` userdata or a `{x,y,z}`/`{1,2,3}` table; zero-defaults for nil/none.
- **Binding cleaned up**: replaced the ad-hoc `pickFloat` lambda + duplicated type-dispatch block in `lua_bindings_prefab.cpp` with a single `vec3FromLua` call. Behavior is identical; the type-error path is now a guard before the helper.
- **Rule documented**: added a "C++ <-> Lua math type helpers" section to `engine/script/CLAUDE.md` with a lookup table of canonical helpers and the callsite-validation contract. Future bindings now have a clear convention to follow.

601/601 IrredenEngineTest pass. — sonnet-fleet-2

jakildev (commented):
## Re-review — T-173: Lua prefab format — vec3FromLua helper (human feedback addressed)

**Verdict:** approve

Re-reviewing after the latest commit addressing the human's inline comment at `lua_bindings_prefab.cpp:54` about messy vec3 extraction from Lua.

### Prior reviews

Sonnet initial pass: needs-fix (entity leak + doc contract mismatch).
Opus full recheck: needs-fix (confirmed). Both issues were fixed and Opus re-reviewed to approve.
Human inline comment: position extraction in `Prefab.spawn` was ad-hoc, should be a canonized helper.

### Human feedback resolution

**vec3FromLua helper** (`engine/script/include/irreden/script/ir_script_utils.hpp`)

The new `IRScript::vec3FromLua(sol::object)` helper is correct:
- `obj.is<IRMath::vec3>()` → direct cast (userdata path).
- `obj.is<sol::table>()` → `pickFloat` lambda checks named key first (`"x"`, `"y"`, `"z"`), then 1-based integer index (Lua convention), then zero-defaults. Handles both `{x=1,y=2,z=3}` and `{1,2,3}` forms correctly.
- nil/none/other → `{0,0,0}`. Zero-default is documented in CLAUDE.md as the contract; callers must type-check before calling if they need a type-error diagnostic.

The inner `pickFloat` lambda is no longer per-callsite duplication — it's encapsulated once in the canonical helper. The inline note on `sol::optional<float>` vs `get_or<T>` is the real WHY (sol2 overload resolution ambiguity with mixed key types) — that's exactly the kind of comment the baseline asks for.

**Binding cleanup** (`engine/script/src/lua_bindings_prefab.cpp`)

The `Prefab.spawn` closure now type-checks `positionObj` before calling `vec3FromLua`, per the helper's callsite-validation contract documented in CLAUDE.md. The guard covers all non-vec3/non-table valid Lua values and surfaces a clear `"position must be a vec3 or {x,y,z} table"` diagnostic. 57 lines total — the previous ad-hoc dispatch block is gone. Clean.

**CLAUDE.md update** (`engine/script/CLAUDE.md`)

"C++ ↔ Lua math type helpers" section added at lines 86-103. Lookup table format is clear; the zero-default + callsite-validation contract is documented; the extension convention (add inline free function, extend the table) is explicit. Future bindings have a clear path.

### Checklist

**ECS invariants:** N/A — no ECS code touched in this commit.

**Math primitives:** `IRMath::vec3` used throughout. No `glm::` or `std::` math.

**Naming:** `vec3FromLua` follows the existing `IRScript::` helper naming in the file. `pickFloat` is lambda-local. ✓

**Render / shaders:** No `engine/render/`, no shaders, no backend source. No cross-host smoke label.

**Tests:** No test changes in this commit — the position path was already covered by `SpawnAttachesPosition` (passes `vec3(7,8,9)` directly). The new `vec3FromLua` function is a utility; its table-input path is exercised transitively via the Lua `Prefab.spawn` binding at the test boundary. Adequate for a utility this small.

### Nits

**1. vec3FromLua comment block is multi-line** (`engine/script/include/irreden/script/ir_script_utils.hpp`)

The function has a 6-line `///` block. CLAUDE-BASELINE.md says "one short line max." The sol2 workaround note (`Named-or-indexed field lookup uses sol::optional<float> per key because...`) is the genuine WHY and should stay. The opening description, zero-default contract, and callsite-validation reminder are already in CLAUDE.md; the in-header comment duplicates what's documented there. Trim to one line: `// sol::optional<float> per key — get_or<T> overload resolution is ambiguous with mixed string/integer key types.` and remove the rest.

---

Opus recheck not required.

jakildev:
Addressed nit: trimmed vec3FromLua multi-line doc block to a single WHY comment — sol::optional<float> per key overload-resolution note retained, opening description / zero-default contract / callsite-validation reminder removed (those live in engine/script/CLAUDE.md). — sonnet-author

PR #712, opened 2026-05-14
T-179: asset: canonicalize memcpy in binary_io + voxel_set_format (bit_cast + chunk-tag helpers)

### LINKED ISSUE #692: asset: canonicalize memcpy patterns in binary_io + voxel_set_format (bit_cast + chunk-tag helpers)
## Context

Escalated from PR #679 (T-168 .vxs v1 shape-group save format). Human review left an inline comment at `engine/asset/src/voxel_set_format.cpp:111`:

> Can we assess use of memcpy here and see if cannonizing a resource manager for all things like this. If so file follow up task to address. Use this filewide

PR #679 is otherwise approved (two prior `fleet:approved` review passes); the request is a design-assessment scope expansion, not a defect in T-168. Filing as a follow-up so the approval stays intact and the cleanup can be reviewed in isolation.

## Assessment

`std::memcpy` is used in three distinct patterns across `engine/asset/`:

**Pattern A — bit-cast (float↔uint32, double↔uint64).** 4 sites in `binary_io.cpp`:
- `binary_io.cpp:72` — `writeF32` bit-casts float to uint32
- `binary_io.cpp:77` — `writeF64` bit-casts double to uint64
- `binary_io.cpp:221` — `readF32` bit-casts uint32 back to float
- `binary_io.cpp:230` — `readF64` bit-casts uint64 back to double

These are the textbook \"type-punning safely\" use of memcpy. C++20 ships `std::bit_cast<T>(v)` as the canonical, constexpr-friendly replacement. The codebase is on C++20 (see top-level `CMakeLists.txt`), so this is a 1:1 swap with no behavior change. `std::bit_cast` also expresses intent precisely (\"this is type punning, not a buffer copy\"), which catches the case where a future author confuses pattern A with pattern C.

**Pattern B — raw buffer copy (write/read primitives).** 2 sites in `binary_io.cpp`:
- `binary_io.cpp:158` — `MemoryBinaryWriter::writeBytes` copies `data` into the internal buffer
- `binary_io.cpp:373` — `MemoryBinaryReader::readBytes` copies from buffer to caller's destination

These ARE the primitives that wrap `memcpy`. No higher-level abstraction is appropriate — every other binary I/O helper goes through these. Leave as-is.

**Pattern C — chunk-tag pack/unpack.** 2 sites in `voxel_set_format.cpp` (T-168) plus the same pattern will recur in every future chunk-producing/consuming function:
- `voxel_set_format.cpp:111` — `makeModeChunk` copies a `std::array<char, 4>` tag into the chunk body buffer
- `voxel_set_format.cpp:125` — `readModeChunk` copies tag bytes out of the chunk body into a `std::array<char, 4>`

Both sites are 4-byte copies between a `std::array<char, 4>` and a `std::vector<std::uint8_t>` (or `std::span<const std::uint8_t>`). Future chunks (e.g. T-169 `.rig` JNTS, T-173 prefab schema) will repeat this pattern. A pair of helpers in `binary_io.hpp` would canonicalize it:

```cpp
// Tag helpers — collapse the std::array<char, 4> ↔ buffer-bytes copy
// that recurs across every chunk producer/consumer.
inline void writeTagBytes(std::vector<std::uint8_t> &buf, std::array<char, 4> tag);
inline std::array<char, 4> readTagBytes(std::span<const std::uint8_t> body);
```

Or, alternatively, fold into `MemoryBinaryWriter::writeTag(std::array<char, 4>)` / `BinaryReader::readTag()`, which would be the more idiomatic placement since chunk tags are a first-class concept in this subsystem.

## Why escalating

The request is scope expansion (design pattern across the binary I/O subsystem), not a defect in T-168. The PR is internally consistent — two prior `fleet:approved` reviews stand, both nits resolved, 533/533 tests pass on macOS post-rebase. Bundling a multi-file refactor into the T-168 merge would also re-trigger CI on already-approved code and force a re-review of unrelated changes.

A standalone PR for this lets the canonical helpers be reviewed in isolation, exercised by their own unit tests, and adopted incrementally by T-169 / T-173 as those land.

## Suggested model

`[sonnet]` — mechanical refactor with clear scope. Pattern A is a verbatim `std::memcpy` → `std::bit_cast` swap; pattern C is two new helper functions plus call-site adoption. No design judgment needed beyond the two helper-placement decisions (free function vs `BinaryWriter` member; both are fine).

## Suggested area

`engine/asset/src/binary_io.cpp`, `engine/asset/include/irreden/asset/binary_io.hpp`, `engine/asset/src/voxel_set_format.cpp`

## Suggested approach

1. **Pattern A** — replace the 4 `memcpy(&bits, &v, sizeof(bits))` sites in `binary_io.cpp` with `std::bit_cast<std::uint32_t>(v)` / `std::bit_cast<std::uint64_t>(v)` for the writer side, and `std::bit_cast<float>(r.value_)` / `std::bit_cast<double>(r.value_)` for the reader side. `#include <bit>`. Existing `BinaryIO` tests (9/9 in `binary_io_test.cpp`) cover round-trips; no new tests needed beyond confirming those still pass.
2. **Pattern C** — add `writeTagBytes` / `readTagBytes` (or `BinaryWriter::writeTag` / `BinaryReader::readTag`) to `binary_io.hpp`. Adopt in the two existing T-168 sites. Add one round-trip unit test in `binary_io_test.cpp`.
3. **Pattern B** — leave alone with a one-line comment at each site explaining why memcpy is the primitive and not the leaf of an abstraction.
4. Build `IrredenEngineTest`; confirm 533/533 still pass. No render or shader code touched.

## Acceptance criteria

- 4 memcpy sites in `binary_io.cpp` converted to `std::bit_cast`; 2 memcpy sites in `voxel_set_format.cpp` converted to the new helpers; remaining 2 memcpy sites (the actual write/read primitives) untouched with a clarifying one-line comment.
- All existing tests pass (533/533).
- One new round-trip test for the chunk-tag helpers.

## Gotchas

- `std::bit_cast` requires `#include <bit>` (the `<cstring>` include can be removed if no memcpy remains, but `binary_io.cpp` still needs it for pattern B).
- Don't try to fold pattern B into a helper — every other primitive in this file already calls through `writeBytes`/`readBytes`, and abstracting those would invert the dependency direction.
- The two T-168 memcpy sites should be migrated as part of this PR, but DO NOT touch T-168's `voxel_set_format.cpp` until #679 merges — otherwise this PR will fight for the same file in rebase.

🤖 Filed by opus-worker (escalated from PR #679 human:needs-fix)

jakildev:
## Summary
- Replace 4 `memcpy` bit-cast sites in `binary_io.cpp` with `std::bit_cast` (float↔uint32, double↔uint64)
- Add `writeTagBytes`/`readTagBytes` helpers to `binary_io.hpp`; adopt in two chunk-tag sites in `voxel_set_format.cpp`
- Leave the 2 raw-buffer-copy `memcpy` sites in `writeBytes`/`readBytes` with clarifying comments
- Add one round-trip unit test for the new chunk-tag helpers

## Test plan
- [ ] Existing `BinaryIO` tests (533+/- total) still pass
- [ ] New `writeTagBytes`/`readTagBytes` round-trip test passes
- [ ] `IrredenEngineTest` build succeeds

Closes #692

### DIFF
--- a/engine/asset/include/irreden/asset/binary_io.hpp
+++ b/engine/asset/include/irreden/asset/binary_io.hpp
@@ -16,6 +16,7 @@
 /// varint is recoverable — never a crash. The 7 Save Format Extensibility
 /// Rules in `docs/design/entity-editor-epic.md` mandate this contract.
 
+#include <array>
 #include <cstddef>
 #include <cstdint>
 #include <cstdio>
@@ -134,6 +135,9 @@ class BinaryWriter {
     /// Length-prefixed UTF-8 string. Prefix is a varint byte-count.
     void writeString(std::string_view s);
 
+    /// Write a 4-byte chunk tag (e.g. `{'V','O','X','R'}`).
+    void writeTag(const std::array<char, 4> &tag);
+
   protected:
     void setFailed(std::string msg) {
         if (!m_failed) {
@@ -240,6 +244,9 @@ class BinaryReader {
     Result<double> readF64();
     Result<std::uint64_t> readVarUInt();
     Result<std::string> readString();
+
+    /// Read a 4-byte chunk tag.
+    Result<std::array<char, 4>> readTag();
 };
 
 /// File-backed reader. Wraps a `FILE*` opened in binary read mode.
--- a/engine/asset/src/binary_io.cpp
+++ b/engine/asset/src/binary_io.cpp
@@ -1,6 +1,7 @@
 #include <irreden/asset/binary_io.hpp>
 #include <irreden/ir_profile.hpp>
 
+#include <bit>
 #include <cstring>
 #include <utility>
 
@@ -68,14 +69,14 @@ void BinaryWriter::writeI64(std::int64_t v) {
     detail::writeLE<std::uint64_t>(*this, static_cast<std::uint64_t>(v));
 }
 void BinaryWriter::writeF32(float v) {
-    std::uint32_t bits;
-    std::memcpy(&bits, &v, sizeof(bits));
-    detail::writeLE<std::uint32_t>(*this, bits);
+    detail::writeLE<std::uint32_t>(*this, std::bit_cast<std::uint32_t>(v));
 }
 void BinaryWriter::writeF64(double v) {
-    std::uint64_t bits;
-    std::memcpy(&bits, &v, sizeof(bits));
-    detail::writeLE<std::uint64_t>(*this, bits);
+    detail::writeLE<std::uint64_t>(*this, std::bit_cast<std::uint64_t>(v));
+}
+
+void BinaryWriter::writeTag(const std::array<char, 4> &tag) {
+    writeBytes(tag.data(), tag.size());
 }
 
 void BinaryWriter::writeVarUInt(std::uint64_t v) {
@@ -155,7 +156,7 @@ void MemoryBinaryWriter::writeBytes(const void *data, std::size_t size) {
     if (end > m_buffer.size()) {
         m_buffer.resize(static_cast<std::size_t>(end));
     }
-    std::memcpy(m_buffer.data() + m_pos, data, size);
+    std::memcpy(m_buffer.data() + m_pos, data, size); // raw buffer copy — the primitive all higher-level writes go through
     m_pos = end;
 }
 
@@ -217,18 +218,22 @@ Result<float> BinaryReader::readF32() {
     if (!r.ok()) {
         return Result<float>::error(r.status_.code_, std::move(r.status_.message_));
     }
-    float v;
-    std::memcpy(&v, &r.value_, sizeof(v));
-    return Result<float>::success(v);
+    return Result<float>::success(std::bit_cast<float>(r.value_));
 }
 Result<double> BinaryReader::readF64() {
     auto r = readU64();
     if (!r.ok()) {
         return Result<double>::error(r.status_.code_, std::move(r.status_.message_));
     }
-    double v;
-    std::memcpy(&v, &r.value_, sizeof(v));
-    return Result<double>::success(v);
+    return Result<double>::success(std::bit_cast<double>(r.value_));
+}
+
+Result<std::array<char, 4>> BinaryReader::readTag() {
+    std::array<char, 4> tag{};
+    if (auto st = readBytes(tag.data(), 4); !st.ok()) {
+        return Result<std::array<char, 4>>::error(st.code_, std::move(st.message_));
+    }
+    return Result<std::array<char, 4>>::success(tag);
 }
 
 Result<std::uint64_t> BinaryReader::readVarUInt() {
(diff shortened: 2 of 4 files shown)


jakildev (commented):
## Review — T-179: asset: canonicalize memcpy in binary_io + voxel_set_format (bit_cast + chunk-tag helpers)

**Verdict:** approve

### Praise

- `std::bit_cast<uint32_t>(v)` / `std::bit_cast<float>(r.value_)` are strictly better than the raw `memcpy` pattern for floating-point type punning — semantically explicit, UB-free, and the compiler can emit the same `mov` without a stack round-trip. Correct for C++23 (project standard); `sizeof(float) == sizeof(uint32_t)` and `sizeof(double) == sizeof(uint64_t)` hold on every platform in the build matrix.
- `readTag` error propagation (`!st.ok()` → early return) is an improvement over the old `readModeChunk` where `r.readBytes(tag.data(), 4)` silently dropped the return value. The pre-existing `size < 4` guard means the error path is unreachable in practice today, but the explicit check is correct defensive posture.
- The two remaining raw `memcpy` sites in `MemoryBinaryWriter::writeBytes` and `MemoryBinaryReader::readBytes` now carry clarifying comments distinguishing them from bit-cast pattern sites. This is the right doc note — those sites are genuinely raw buffer copies, not type-punning.
- Two tight tests: `ChunkTagRoundTrip` covers the happy path end-to-end (byte count, round-trip value, zero remaining); `ChunkTagReadTruncated` confirms the `BinaryIOError::Truncated` error code surfaces from the truncated-buffer path in `MemoryBinaryReader::readBytes`.

### Checklist notes

**Binary output unchanged:** `std::bit_cast` and `memcpy` produce identical IEEE 754 bit patterns for float/double ↔ uint32/uint64. Existing `.vxs` and `.rig` files round-trip cleanly. No `kSaveVersion` bump needed — no serialized struct fields changed.

**`voxel_set_format.cpp` — all raw chunk-tag sites updated:** `grep writeBytes|readBytes` on the file returns no matches. Both sites (write in `makeModeChunk`, read in `readModeChunk`) are now on the typed helpers. No remaining raw tag I/O.

**`<bit>` include:** Added to `binary_io.cpp`. `<array>` added to `binary_io.hpp`. Both explicit; no implicit-include dependency.

**ECS invariants:** N/A — no ECS code in this PR.

**Render pipeline / shaders:** Not touched. No cross-host smoke labels.

**Naming:** `writeTag`/`readTag` follow the same lowercase camelCase as all other write/read methods on `BinaryWriter`/`BinaryReader`. ✅

**Serialization version check:** No `// IRAsset: serialized` structs have field changes; only the implementation helpers changed. `kShapeRecordVersion`, `kVoxelRecordVersion`, etc. untouched. ✅

Opus recheck not required.

PR #715, opened 2026-05-14
docs: agent-rule additions — math binary I/O, contract asserts, deprecation markers

jakildev:
## Summary
Three small, independent rule additions captured from an engine/asset cleanup-pass audit (parent epic #603 closeout). Each codifies a pattern the fleet would otherwise re-rediscover with every adjacent PR.

- **`cpp-math.md` — "Binary I/O of math types".** Serialization helpers for `IRMath::vec*` / `Color` / `quat` must live in `engine/math/` (or alongside `BinaryWriter` / `BinaryReader` in `engine/asset/`), never inline in a format-specific `.cpp`. Standardize naming on `read` / `write`; no `encode` / `decode` / `pack` / `unpack` aliases.
- **`CLAUDE-BASELINE.md` — "Encode contracts in code, not in comments".** Helper preconditions in docstrings must be enforced with `IR_ASSERT` / `static_assert`, not the comment alone. Disambiguates from "don't add error handling for scenarios that can't happen" by being the narrower point: the comment says *why*; the assert ensures the violation is loud rather than silent.
- **`CLAUDE-BASELINE.md` — "Deprecation markers".** Three-place rule (declaration site / module `CLAUDE.md` / new design docs) for any API or format being phased out. Concurrent fleet tasks that propose to *extend* a deprecated surface must escalate.

## Failure modes each rule catches

| Rule | Canonical failure |
|---|---|
| Math binary I/O location | `writeVec3` / `encodeVec3` duplicated under two names across `voxel_set_format.cpp` and `rig_format.cpp` |
| Contract asserts | `makeTag(std::string_view s)` silently zero-pads when `s.length() < 4` despite the docstring saying "must be exactly 4" |
| Deprecation markers | F-0.7 (#626) filed a `.txl.json` sidecar against a format being migrated to `.vxs`; scaffolding shipped, was never adopted, got deleted in #705 |

## Test plan
- [ ] Doc-only diff. Each section's body cites file:symbol references that have been verified against current master.
- [ ] No conflict with existing rules — the contract-assert rule explicitly disambiguates from `simplify` §8's "don't validate scenarios that can't happen"; the deprecation rule explicitly complements (does not replace) the existing "Engine API removal rule."
- [ ] No new build target, no shader changes, no ECS surface change.

## Notes for reviewer
[...]

Issue #708, opened 2026-05-14
render: LOD Phase 1 — wire computeLodLevel(zoom) and per-shape LOD-min filter


jakildev:
## Context

We have artist-driven LOD scaffolding already in the codebase that is not yet wired up. The motivating use case: a flower entity that shows extra petals/stamen at higher zoom levels; at low zoom we want the renderer to skip the high-detail shapes entirely.

**What already exists:**

- `engine/prefabs/irreden/render/lod_utils.hpp` — WIP stub. `computeLodLevel(float zoomLevel)` maps zoom thresholds to five discrete LOD levels (LOD_0 = highest detail at zoom >= 4.0, ... LOD_4 = lowest at zoom < 0.5). `lodVoxelScale(LodLevel)` returns per-level scale factors. `shouldSkipAtLod(entityLodMin, currentLod)` is a skip predicate. Self-described as "WIP stub — not yet included or referenced by any system" (`lod_utils.hpp:6`).
- `C_ShapeDescriptor::lodLevel_` — already exists. Already shipped to the GPU as `GPUShapeDescriptor.lodLevel` in `system_shapes_to_trixel.hpp:111`. **The shader currently ignores it.**
- `C_ZoomLevel::zoom_` — continuous float in [1.0, 64.0], power-of-2 snapped (`render_manager.cpp:261`). Queryable via `IRRender::getCameraZoom()` (`render_manager.cpp:220-222`).

The plumbing is 80% present. This ticket wires it up.

This is Phase 1 of a multi-phase LOD story. The design rationale (why we're going artist-driven rather than baking LOD tiers into `.vxs` / `.rig`) lives in `docs/design/lod-strategy.md` (filed alongside this ticket). **Read that doc first** — Phase 2 (multi-tier `.vxs` references from the prefab manifest layer) and Phase 3 (cross-tier interpolation) are scoped separately; do not implement them here.

## Phase 1 scope

1. **Promote `lod_utils.hpp` from stub to live.** Keep the function signatures; refine the thresholds if the design doc lands on different numbers.

2. **Add `C_ActiveLodLevel` singleton component.** Carries `LodLevel current_` (one of LOD_0..LOD_4). Populated once per frame by a new system.

3. **Add `LOD_UPDATE` system in the UPDATE pipeline.** Reads `C_ZoomLevel` from the active camera entity, calls `computeLodLevel`, writes `C_ActiveLodLevel`. Runs once per frame, before the RENDER pipeline.
[...]

PR #716, opened 2026-05-14
docs: three more agent-rule additions — simplify 2c extension, no-raw-fwrite, source-citation convention

jakildev:
## Summary
Follow-up to PR #715 — second batch of three rule additions from the same engine/asset cleanup-pass audit. Each codifies a pattern the fleet would otherwise re-rediscover with every adjacent PR.

- **`simplify` §2c — unannotated serialized structs check.** The existing version-bump check only protects already-annotated structs; structs that are serialized field-by-field but never annotated (`ShapeRecord` shipped this way in T-168 / T-170) escape entirely. New detection heuristic: two-or-more `<writer>.write*(<expr>.<field>_)` calls referencing the same struct instance, paired with a struct declaration that lacks `// IRAsset: serialized` on the immediately prior line. Flag, don't auto-fix — the initial `kSaveVersion` is a human call. False-positive guards listed for framework types.
- **`engine/asset/CLAUDE.md` — hard rule on `BinaryWriter`/`BinaryReader`.** Promote the implicit "new formats must use these" sentence to its own paragraph at the top of the Binary-I/O primitives section: no raw `fopen` + `fwrite` / `fread` for new format I/O, with the `.txl` unchecked-fread liability as the concrete failure example.
- **`CLAUDE-BASELINE.md` — citing source in filed artifacts.** Prefer `file::symbol` over `file:line` in anything that gets read after the next refactor (GitHub issues, PR bodies, design docs, TASKS.md). Includes the preferred-vs-avoided table, the carve-outs for cases where line numbers are still right (PR-internal diff citations, magic-number pointers, verbatim stack traces), and a mixed-form fallback.

## Failure modes each rule catches

| Rule | Canonical failure |
|---|---|
| Unannotated-serialized-struct check | `ShapeRecord` shipped in T-168 / T-170 with every field individually serialized but no `// IRAsset: serialized` annotation; gap surfaced only on later audit |
| No raw `fwrite` / `fread` rule | `.txl` v1 save/load with unchecked return codes — truncated files load as garbage, corrupted saves succeed silently |
| Symbol-over-line-number citations | Issue #708 cited `system_shapes_to_trixel.hpp:154`; actual line was `111`. A `system_shapes_to_trixel::beginTick` citation would have survived |

## Test plan
- [ ] Doc-only diff; +107 lines across three files.
- [ ] `simplify` §2c extension's false-positive guards explicitly exclude framework types (`BinaryStatus`, `LoadedChunk`, `ChunkPayload`, `Result<T>`) and binary-I/O header types, so the existing simplify pass on a normal-format-extension PR stays quiet.
- [ ] New baseline section sits between "What belongs in CLAUDE.md files" and "Bash tool rules" — far enough from PR #715's tail-of-file additions that merge order doesn't matter.

## Notes for reviewer
- PR #715 (math binary I/O / contract asserts / deprecation markers) and this PR are independent appends to the same set of doc files. If #715 merges first, this PR will rebase cleanly; same in the reverse order.
- Three more items remain on the original session list (Explore-agent rebase-before-audit, architect-role grep-for-stubs reminder, periodic CLAUDE.md audit). Per the prior conversation we're leaving those for a future session.

### DIFF
--- a/.claude/skills/simplify/SKILL.md
+++ b/.claude/skills/simplify/SKILL.md
@@ -252,6 +252,59 @@ Run this check whenever any `.hpp` or `.cpp` file under `engine/asset/`,
 - Changes to the `kSaveVersion` line itself.
 - Changes that only touch comments or whitespace inside the struct.
 
+**Detection extension — unannotated serialized structs:**
+
+The version-bump check above protects structs that are *already* annotated.
+A second class of bug is structs whose fields are individually serialized
+but which lack the `// IRAsset: serialized` annotation entirely — the
+version-bump check then never fires, even on layout changes that need a
+migration. `ShapeRecord` (`engine/asset/include/irreden/asset/voxel_set_format.hpp`)
+shipped in T-168 / T-170 without the annotation despite each of its
+fields being individually written and read by the format's chunk
+handlers; the gap stayed invisible until an audit pass surfaced it.
+
+Run this additional check in the same scope as the version-bump check
+(diff in `engine/asset/`, `engine/prefabs/irreden/voxel/`, or
+`engine/world/`):
+
+1. Scan the diff's `+` lines for any sequence of two or more
+   `<writer>.write*(<expr>.<field>_)` calls where `<expr>` is a
+   reference to a struct instance and `<field>_` is a public member.
+   Likewise for `<reader>.read*` paired with assignment back to
+   `<expr>.<field>_`. Two-or-more is the heuristic for "this struct
+   is being serialized field-by-field" rather than "one field of a
+   struct happens to be written."
+
+2. For each struct type touched this way, resolve its declaration
+   (same file or sibling header) and check whether the declaration
+   is preceded by `// IRAsset: serialized` on the immediately prior
+   line per the annotation rule.
+
+3. If the struct is serialized field-by-field and lacks the
+   annotation, emit a finding — **do not auto-fix**, the annotation
+   carries a `kSaveVersion` constant whose initial value is a human
+   call:
+
+   ```
+   reported 1 finding for review:
+     - <header-path>:<line> — <StructName> is serialized field-by-field
+       by <function-path>:<line> but lacks the `// IRAsset: serialized`
+       annotation. Add the comment line directly above the struct
+       declaration plus a `static constexpr uint16_t kSaveVersion = 1;`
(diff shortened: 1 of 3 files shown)


jakildev (commented):
## Review — docs: three more agent-rule additions — simplify 2c extension, no-raw-fwrite, source-citation convention

**Verdict:** approve

### Praise

- The unannotated-serialized-struct extension to simplify §2c closes a genuine blind spot: the existing version-bump check only fires when `// IRAsset: serialized` is already present. The `ShapeRecord` miss in T-168/T-170 is the perfect motivating example — exactly the kind of "surfaced only on later audit" gap this heuristic prevents. Keeping it flag-not-autofix is correct since the initial `kSaveVersion = 1` is a human judgment.
- False-positive guards are specific and sufficient: excluding framework types (`BinaryStatus`, `LoadedChunk`, `ChunkPayload`, `Result<T>`) and framework headers prevents the heuristic from firing on every binary-I/O helper that happens to have two `write*` calls.
- The no-raw-fwrite rule includes the concrete failure mode (`.txl` unchecked return values → silent truncation/corruption), which is exactly the motivating evidence that makes the rule stick rather than feeling arbitrary. The explicit "those will be migrated or deleted, not extended" commitment is the right forward signal.
- The citation section's carve-out table is precise: PR-internal diff citations (frozen), unnamed lines/magic numbers (symbol-plus-offset fallback), verbatim stack traces. These are the three real cases where line numbers remain useful; getting them right prevents over-correction.

### Checklist notes

**Placement:** All three insertions are clean:
- `simplify/SKILL.md` — §2c extension follows the existing false-positive exclusions block; stays in scope.
- `CLAUDE-BASELINE.md` — inserted between "What belongs in CLAUDE.md files" and "## Bash tool rules" with correct `---` dividers on both sides.
- `engine/asset/CLAUDE.md` — hard-rule paragraph inserted directly after "new formats must use these" and before the `### binary_io.hpp` subsection, which is the right landing spot.

**Internal consistency:** The simplify extension's scope-gate matches the existing §2c scope (`engine/asset/`, `engine/prefabs/irreden/voxel/`, `engine/world/`). The finding format (`:path:line — StructName is serialized...`) matches the existing §2c finding format. ✅

**ECS / render / build / tests:** Docs-only. N/A.

**Cross-host smoke:** Not needed.

Opus recheck not required.

PR #724, opened 2026-05-14
T-186: test: JsonSidecarWriter + NameTable round-trips

### LINKED ISSUE #707: test: cover JsonSidecarWriter + NameTable round-trips (engine/asset gaps)
## Context

`engine/asset/` audit pass turned up two files with no dedicated test coverage:

- **`engine/asset/src/json_sidecar.cpp`** — `JsonSidecarWriter` + `writeJsonSidecarToFile`. Currently exercised only indirectly via the sidecar-content tests in `test/asset/voxel_set_hybrid_test.cpp` (write a `.vxs`, read the sidecar bytes back, assert substrings). That covers the happy-path output shape but doesn't isolate the writer's behavior on edge cases.
- **`engine/asset/src/name_table.cpp`** — `NameTable` lookup class plus the `writeNameTable` / `readNameTable` chunk-body serializers. Currently exercised only as a side effect of `.vxs` SREF chunk handling.

Both will be reused by every future format (`.rig` already does; `.prefab.lua` and the world snapshot per #667 will). Bare coverage in their own test files makes the contracts explicit and isolates failures.

## Suggested model

`[sonnet]` — additive test code only. No production changes.

### `test/asset/json_sidecar_test.cpp`

Cover:
- Empty object / empty array round-trip — `{}` and `[]`.
- Scope nesting: object inside array inside object.
- All `value*` overloads: `valueString`, `valueInt`, `valueUInt`, `valueFloat`, `valueBool`, `valueNull`.
- String escaping: embedded double-quote, backslash, newline, tab, and control characters in the U+0000–U+001F range.
- Numeric edge cases: very large `uint64`, negative `int64`, `INT64_MIN`, `+0.0`, `-0.0`. NaN / Inf — define the contract (writer either rejects, emits `null`, or emits a non-spec literal) and lock it.
- Key ordering preserved across calls (sidecars are diffed in PR review; stable key order matters).
- `writeJsonSidecarToFile` to a path that fails to open — verify the failure mode (current behavior: `IRE_LOG_ERROR` and return).

### `test/asset/name_table_test.cpp`

Cover:
- `writeNameTable` / `readNameTable` round-trip with 0, 1, and many entries.
- Name strings with non-ASCII bytes (the table is UTF-8 per Rule #2).
- `NameTable::idByName` / `NameTable::nameById` with hits, misses, and duplicate inserts (define which wins).
- Construction from `span<NameTableEntry>` matches construction by sequential `add` (if both shapes exist; otherwise pick one).
- Edge case: empty name string (Rule #2 doesn't forbid it; lock behavior).

### CMake wiring

Both new test files register in `test/asset/CMakeLists.txt` (or the project-root test list, whichever pattern other asset tests follow). Build `IrredenEngineTest` and confirm the new tests run.

## Acceptance criteria

- `test/asset/json_sidecar_test.cpp` exists, registered with the test target.
- `test/asset/name_table_test.cpp` exists, registered with the test target.
- All tests pass; total voxel-set + asset suite count increases by the new tests (currently 46/46 for voxel-set + N for binary-I/O / sprite-sheet — exact delta is whatever the author writes).

## Gotchas

- **Don't depend on `.vxs` for these tests.** Each test file should exercise its target in isolation — use `MemoryBinaryWriter`/`Reader` for name-table tests and the writer's internal `str()` for sidecar tests, not file I/O.
- **NaN / Inf in JSON.** JSON spec disallows them; common emitters either fail-loud, fall back to `null`, or emit `NaN` non-portably. Whatever `JsonSidecarWriter` currently does, lock it in a test.
- **Test count assertions.** If any other test file asserts `EXPECT_EQ(testCount, 46)` or similar, update it. The fleet's review-pr skill flags drifting counts.

🤖 Filed by opus-architect (engine/asset cleanup pass — coverage gaps)

jakildev:
## Summary

- Adds `test/asset/json_sidecar_test.cpp` with 7 focused tests for `JsonSidecarWriter` edge cases not covered in `binary_io_test.cpp`: empty object/array, control-character `\uXXXX` encoding, numeric extremes (UINT64_MAX, INT64_MIN, ±0.0), Inf→null contract, key-order preservation, and `writeJsonSidecarToFile` failure path.
- Adds `test/asset/name_table_test.cpp` with 6 focused tests for `NameTable`: UTF-8 name round-trip, duplicate-name/id first-wins semantics, span-constructor vs sequential-add parity, empty name string round-trip, single-entry round-trip.
- Both files registered in `test/CMakeLists.txt`; 22/22 tests pass.

## Test plan

- [x] `fleet-build --target IrredenEngineTest` — clean build
- [x] `fleet-run IrredenEngineTest --gtest_filter="JsonSidecar.*:NameTable.*"` — 22/22 PASSED
- [ ] Linux smoke (fleet:needs-linux-smoke will be added by reviewer if needed)

Closes #707

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -3,6 +3,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 
 add_executable(IrredenEngineTest
   asset/binary_io_test.cpp
+  asset/json_sidecar_test.cpp
+  asset/name_table_test.cpp
   asset/rig_format_test.cpp
   asset/sprite_sheet_test.cpp
   asset/txl_sidecar_test.cpp
--- a/test/asset/json_sidecar_test.cpp
+++ b/test/asset/json_sidecar_test.cpp
@@ -1,0 +1,108 @@
+#include <gtest/gtest.h>
+#include <irreden/asset/json_sidecar.hpp>
+
+#include <limits>
+#include <string>
+
+namespace {
+
+using namespace IRAsset;
+
+TEST(JsonSidecar, EmptyObject) {
+    JsonSidecarWriter j;
+    j.beginObject();
+    j.endObject();
+    EXPECT_EQ(j.str(), "{}");
+}
+
+TEST(JsonSidecar, EmptyArray) {
+    JsonSidecarWriter j;
+    j.beginArray();
+    j.endArray();
+    EXPECT_EQ(j.str(), "[]");
+}
+
+TEST(JsonSidecar, ControlCharactersEscapedAsUnicode) {
+    // Chars not handled by named escapes but in U+0000–U+001F emit \uXXXX.
+    JsonSidecarWriter j;
+    j.beginObject();
+    j.key("nul");
+    j.valueString(std::string(1, '\x00'));  // U+0000 — null byte
+    j.key("esc");
+    j.valueString(std::string(1, '\x1b'));  // U+001B — ESC
+    j.key("us");
+    j.valueString(std::string(1, '\x1f'));  // U+001F — unit separator
+    j.endObject();
+    const std::string out = j.str();
+    EXPECT_NE(out.find("\\u0000"), std::string::npos);
+    EXPECT_NE(out.find("\\u001b"), std::string::npos);
+    EXPECT_NE(out.find("\\u001f"), std::string::npos);
+}
+
+TEST(JsonSidecar, NumericEdgeCases) {
+    JsonSidecarWriter j;
+    j.beginObject();
+    j.key("uint_max");
+    j.valueUInt(std::numeric_limits<std::uint64_t>::max());
+    j.key("int_min");
+    j.valueInt(std::numeric_limits<std::int64_t>::min());
+    j.key("neg");
+    j.valueInt(-1);
+    j.key("pos_zero");
+    j.valueFloat(+0.0);
+    j.key("neg_zero");
+    j.valueFloat(-0.0);
+    j.endObject();
+    const std::string out = j.str();
+    EXPECT_NE(out.find("18446744073709551615"), std::string::npos);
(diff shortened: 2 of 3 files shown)


jakildev (commented):
## Review — T-186: test: JsonSidecarWriter + NameTable round-trips

**Verdict:** approve

### Praise

- `JsonSidecar.InfinityEmitsNull` explicitly documents and locks the Infinity→null contract. The header's NaN comment implied it but didn't cover ±∞; a future refactor of `valueFloat()` now has a regression guard.
- `NameTable.DuplicateNameFirstWins` and `NameTable.DuplicateIdFirstWins` together pin the `unordered_map::emplace` first-wins semantics. A refactor that flips to `insert_or_assign` would break these tests immediately.
- UTF-8 round-trip test covers 2-byte (é), 3-byte (日本語), and 4-byte (🦀) sequences — the full range of multi-byte codepoints the varuint-length-prefixed string serializer must handle.
- `JsonSidecar.ControlCharactersEscapedAsUnicode` includes the null byte (U+0000), the most likely source of silent truncation in a C-string fallback — rules that out explicitly.

### Checklist

**ECS invariants:** No ECS code. ✓

**Serialized structs:** `JsonSidecarWriter` and `NameTable` are I/O primitives, not serialized data structs. No `kSaveVersion` concerns. ✓

**Math primitives:** No math. ✓

**Naming:** Test suite names match class names. No `C_`/`m_` convention violations in test-only code. ✓

**Build:** Both files registered in `test/CMakeLists.txt`. Author reports 22/22 tests passing. ✓

Opus recheck not required.

jakildev:
Closing as duplicate of PR #730. PR #730 is a strict superset: it has 12 JsonSidecar tests vs 7 here and 7 NameTable tests vs 6, covering all the same contracts plus scope-nesting, CR/BS/FF named escapes, null-byte escape (cherry-picked into #730 before this closure), raw U+001E/U+001F control chars, and 100-entry round-trip. All unique coverage from this PR has been preserved in #730.

PR #725, opened 2026-05-14
T-190: SDF→trixel half-voxel/lone-trixel discrepancy investigation

### LINKED ISSUE #690: SDF→trixel: investigate half-voxel / lone-trixel discrepancy vs C_VoxelSetNew
## Context

Human feedback on PR #659 (T-163 stateless particle render):

> furthermore, I think the SDF renderer converts the shapes to trixels, but doesnt necessarily make it match up how it should be if it was constructed with voxels (some half voxels or lone trixels seem to be added).

Observation while comparing `C_VoxelSetNew` carved-from-SDF output vs `C_ShapeDescriptor` direct-SDF rasterization (`SHAPES_TO_TRIXEL`). At certain zoom levels (visible in `creations/demos/lighting/`'s zoom8/zoom16 shots) the SDF path appears to emit half-extent trixels or isolated single-trixel artifacts at silhouette boundaries that the voxel-pool path does not produce for the same shape.

## Scope

Investigate:

1. **Reproduce.** Render the same SDF shape (sphere, box) both as voxel pool (`createVoxelPoolShape` from `lighting_demo_scene.hpp`) and as direct SDF (`createSdfShape`) at zoom8 / zoom16. Use `tools/img_diff` to highlight per-trixel drift between the two.
2. **Classify.** Are the discrepancies:
   - Sub-pixel anti-aliasing the SDF intentionally adds? (Unlikely — SDF uses `sdf <= 0.5` carve threshold.)
   - Off-by-one from the `kSdfBiasEpsilon` (1e-3) or `stableCeilToInt` ceiling bias at borderline depths?
   - The 2x3 trixel diamond emit path painting both subpixels of a face when only one should fire near edge cases?
   - A genuine bug in `snapLatticeWalk` vs `findSurfaceDepth`?
3. **Determine.** Should the SDF and voxel-pool outputs be bit-identical when both target the same effective voxel resolution, or is some delta acceptable as long as it's bounded?
4. **Fix or document.** Either correct the SDF path so it matches voxel-pool output trixel-for-trixel, or document the intentional delta with a CLAUDE.md note explaining the bias.

## Suggested approach

Read `c_shapes_to_trixel.glsl`'s `boxDepthIntersect` / `sphereDepthIntersect` / `snapLatticeWalk` paths alongside `c_voxel_to_trixel_stage_1.glsl`'s `localIDToFace_2x3` + `faceOffset_2x3` emit. The voxel-pool emits ALL 6 trixels (3 faces × 2 subpixels) per voxel; the SDF emits the same shape but uses analytical surface-depth solvers. Differences in the integer ceiling / FP epsilon could cause one path to emit one subpixel while the other emits zero or both.

The boundaries to focus on: silhouette pixels where `stableCeilToInt(dEntry)` is right at an integer; pixels where one face's analytical entry intersects the bounding box but another doesn't.

## Acceptance criteria

- Diff report comparing voxel-pool vs SDF output at zoom 4 / 8 / 16 for the four standard shapes (box, sphere, cone, torus).
- Either a fix PR that eliminates the discrepancy, or a CLAUDE.md note in `engine/render/` documenting the intentional difference and its source (which threshold, which solver path).

## Gotchas

- The SDF path has both smooth (subdivisions > 1) and snap (subdivisions == 1) modes. The snap path uses `snapLatticeWalk` which is designed to match `C_VoxelSetNew` trixel-for-trixel — divergence here is more likely a bug than intentional.
- The voxel-pool path also has `kSdfBiasEpsilon` analogs in its carve logic. The SDF surface-find epsilon and the voxel-pool carve epsilon should agree.

Filed as follow-up to PR #659 per human direction ('Perhaps if you work on integrating this and unifying it, you can also identify if this is the case, and file a follow up ticket'). Suggested model: [opus] (analytical-geometry investigation, may require an O(1) solver epsilon-tuning pass).

jakildev:
## Summary
- Investigates the SDF→trixel vs voxel-pool→trixel discrepancy raised on PR #659.
- Concludes the silhouette difference at smooth-mode (sub > 1) is by design, not a bug.
- Documents the parity contract by `SubdivisionMode` in `engine/render/CLAUDE.md` so the next person doesn't re-investigate.

## What's actually happening

Two paths exist for getting an SDF onto the canvas:

| Mode | SDF (`SHAPES_TO_TRIXEL`) | Voxel pool (`VOXEL_TO_TRIXEL_*`) | Match? |
|------|--------------------------|-----------------------------------|--------|
| `sub == 1` (any render mode) | `snapLatticeWalk` — parity-gated, `roundHalfUp` to voxel, `sdf <= 0.5` no bias | Integer voxel projection at carve-set | bit-identical |
| `sub > 1` (smooth mode) | `findSurfaceDepth` — analytical solver, sub-pixel surface | `faceMicroPositionFixed` — `sub²` micro-emits per active voxel | by-design difference |

At smooth mode `sub > 1` the SDF renders the smooth analytical surface; the voxel pool renders the discrete carved voxel set scaled up. The "lone trixel" / "half-extent voxel" pattern visible in the lighting demo's zoom8/zoom16 captures is the difference between these two surface representations — exactly what smooth mode exists to provide. Forcing the SDF to match the voxel pool's stair-stepped silhouette would defeat the purpose of having an analytical path at all.

The `kSdfBiasEpsilon` (1e-3) used in the analytical surface check is for FMA-noise stability, not a deliberate widening — it can keep one extra sub-pixel column at borderline analytical hits where the CPU carve drops, but the visual effect is bounded by the epsilon and only visible at static analysis with a per-pixel diff.

## Resolution

Per the task acceptance ("Either a fix PR that eliminates the discrepancy, or a CLAUDE.md note in `engine/render/` documenting the intentional difference and its source"), this PR takes the documentation path. New section added in `engine/render/CLAUDE.md` covers:

- Threshold (`kSdfBiasEpsilon = 1e-3`) and its role
- Solver paths (`snapLatticeWalk` vs `findSurfaceDepth`) per `SubdivisionMode`
- The exact condition under which both paths are bit-identical (`sub == 1`)
- Where to look for visual reference (lighting demo `kShots[]`)

## Test plan
- [x] No code changed; doc-only diff
- [x] All cited identifiers verified present in tree (`snapLatticeWalk`, `findSurfaceDepth`, `roundHalfUp`, `faceMicroPositionFixed`, `kSurfaceThreshold`, `kSdfBiasEpsilon`, `SubdivisionMode::NONE`, commit `87d2b681`)

Closes #690

### DIFF
--- a/engine/render/CLAUDE.md
+++ b/engine/render/CLAUDE.md
@@ -453,6 +453,52 @@ of the lighting pass. Invoke from a creation via the engine API
 (`IRRender::setDebugOverlay`) or in `shape_debug` via
 `--debug-overlay <none|ao|light_level|shadow>`.
 
+## SDF (`SHAPES_TO_TRIXEL`) vs voxel-pool (`VOXEL_TO_TRIXEL_*`) parity
+
+A `C_ShapeDescriptor` (SDF, GPU-evaluated) and a `C_VoxelSetNew` carved
+from the same SDF (CPU-quantized at construction) are intentionally NOT
+trixel-for-trixel identical at every render configuration. The
+relationship varies by `SubdivisionMode`:
+
+- **`NONE` / `POSITION_ONLY` and `SMOOTH` at effective `sub == 1`** —
+  bit-identical at the silhouette. The SDF shader skips the analytical
+  surface solver entirely and routes through `snapLatticeWalk`
+  (`c_shapes_to_trixel.glsl` `smoothMode = (renderMode != 0) &&
+  (subdivisions > 1)` — the `subdivisions > 1` half of that gate
+  was added in commit 87d2b681 so `sub == 1` falls back to the
+  parity-gated lattice walk instead of the analytical 2x3 emit
+  that aliased against the voxel-pool tiling). The walk only
+  evaluates iso pixels with `(isoRel.x + isoRel.y) & 1 == 0` — the same
+  even-parity set integer voxels project to — and rounds each candidate
+  via `roundHalfUp` to the same integer voxel the CPU carve evaluates.
+  The SDF surface check uses `<= 0.5` with no bias, matching CPU's
+  `> kSurfaceThreshold` exclusion exactly.
+- **`SMOOTH` at `sub > 1`** — silhouette differs by design. The SDF
+  shader runs `findSurfaceDepth` analytically: each iso sub-pixel solves
+  for the smooth surface depth, producing a continuous (sub-pixel)
+  silhouette. The voxel pool runs `faceMicroPositionFixed` over `sub²`
+  micro-positions per active voxel, producing a stair-stepped silhouette
+  that snaps to the discrete carved voxel set scaled up by `sub`. At any
+  given silhouette iso pixel one path may emit where the other doesn't
+  — that's the "lone trixel" / "half-extent voxel" effect on a
+  side-by-side comparison. It is NOT a bug; it is the entire reason
+  smooth mode has two implementations.
+
+The `kSdfBiasEpsilon` (1e-3) used in the analytical surface check
+(`sdf <= 0.5 + kSdfBiasEpsilon`) is for FMA-noise stability across
+frames at integer-edge depths, not a deliberate widening of the surface
+shell. It can keep one extra sub-pixel column at borderline analytical
+hits where the CPU carve drops; the visual effect is bounded by the
+epsilon and only visible at static analysis with a per-pixel diff.
+
+If you need bit-identical voxel-pool output from a `C_ShapeDescriptor`,
+the only correct configuration is `SubdivisionMode::NONE` (or any
+`sub == 1` configuration) where both paths route through the same
+parity-gated lattice walk. At higher subdivisions, choose the path
+based on intent: voxel pool for "cubes-of-cubes" stylization, SDF for
+smooth analytical silhouettes. The lighting demo's `kShots[]`
+zoom8/zoom16 captures showcase the difference for visual reference.
+
 ## Gotchas
 
 - **Hardcoded uniform-buffer bind points.** Indices like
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — T-190: SDF→trixel half-voxel/lone-trixel discrepancy investigation

**Verdict:** approve

### Praise

- All six identifiers cited in the new CLAUDE.md section are present in the tree and accurate: `snapLatticeWalk` (c_shapes_to_trixel.glsl:644), `findSurfaceDepth` (:680), `kSdfBiasEpsilon = 1e-3` (:102), `faceMicroPositionFixed` (c_voxel_to_trixel_stage_2.glsl:114), `kSurfaceThreshold = 0.5f` (sdf.hpp:27), and the `smoothMode` gate expression `(renderMode != 0) && (subdivisions > 1)` (:751). No stale refs.
- The `snapLatticeWalk` `<= 0.5` (no bias) vs analytical paths `<= 0.5 + kSdfBiasEpsilon` distinction is a meaningful accuracy difference that wasn't documented anywhere before. Now the next investigator won't re-derive this from the shader.
- Routing the resolution through the doc path (vs. a "fix" that would flatten the analytical advantage) is the correct call given that the smooth-mode discrepancy is the entire point of the analytical path.
- The actionable guidance at the end ("choose voxel pool for cubes-of-cubes, SDF for smooth analytical silhouettes") makes the doc prescriptive rather than just descriptive.

### Checklist

**ECS invariants:** Doc-only PR. No code changes. ✓

**Math primitives:** No code changes. ✓

**Naming:** No new symbols introduced. ✓

**Cross-host smoke:** Doc-only change to `engine/render/CLAUDE.md`, no shader or C++ files touched. Skip smoke tagging. ✓

Opus recheck not required.

jakildev:

— fleet merger

PR #728, opened 2026-05-14
T-186: test: JsonSidecarWriter + NameTable round-trips

### LINKED ISSUE #707: test: cover JsonSidecarWriter + NameTable round-trips (engine/asset gaps)

jakildev:
## Summary
- Add `test/asset/json_sidecar_test.cpp` with dedicated edge-case tests for `JsonSidecarWriter`: empty object/array, control-char escaping (`\r`, `\b`, `\f`, U+0001–U+001F), `INT64_MIN` / `UINT64_MAX` numerics, positive/negative zero, infinity-emits-null, key-ordering preserved, file-write failure path.
- Add `test/asset/name_table_test.cpp` with dedicated tests for `writeNameTable`/`readNameTable` + `NameTable`: single-entry round-trip, 100-entry round-trip, non-ASCII UTF-8 names, empty name string, vector-constructor vs sequential-add parity, duplicate-name/id first-insert-wins contracts.
- Register both files in `test/CMakeLists.txt`.

Existing coverage in `binary_io_test.cpp` covers the happy paths; these new files isolate the edge cases and make the per-unit contracts explicit.

## Test plan
- [ ] `fleet-build --target IrredenEngineTest` compiles clean
- [ ] All new tests pass (`fleet-run IrredenEngineTest`)

Closes #707

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -3,6 +3,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 
 add_executable(IrredenEngineTest
   asset/binary_io_test.cpp
+  asset/json_sidecar_test.cpp
+  asset/name_table_test.cpp
   asset/rig_format_test.cpp
   asset/sprite_sheet_test.cpp
   asset/txl_sidecar_test.cpp
--- a/test/asset/json_sidecar_test.cpp
+++ b/test/asset/json_sidecar_test.cpp
@@ -1,0 +1,145 @@
+#include <gtest/gtest.h>
+
+#include <irreden/asset/json_sidecar.hpp>
+
+#include <cstdint>
+#include <limits>
+#include <string>
+
+namespace {
+
+using namespace IRAsset;
+
+TEST(JsonSidecar, EmptyObjectProducesEmptyBraces) {
+    JsonSidecarWriter j;
+    j.beginObject();
+    j.endObject();
+    EXPECT_EQ(j.str(), "{}");
+}
+
+TEST(JsonSidecar, EmptyArrayProducesEmptyBrackets) {
+    JsonSidecarWriter j;
+    j.beginArray();
+    j.endArray();
+    EXPECT_EQ(j.str(), "[]");
+}
+
+TEST(JsonSidecar, ObjectInsideArrayInsideObject) {
+    JsonSidecarWriter j;
+    j.beginObject();
+    j.key("items");
+    j.beginArray();
+    j.beginObject();
+    j.key("x");
+    j.valueInt(7);
+    j.endObject();
+    j.endArray();
+    j.endObject();
+    const std::string out = j.str();
+    EXPECT_NE(out.find("\"items\""), std::string::npos);
+    EXPECT_NE(out.find("\"x\": 7"), std::string::npos);
+    EXPECT_EQ(out.front(), '{');
+    EXPECT_EQ(out.back(), '}');
+}
+
+TEST(JsonSidecar, StringEscapingCarriageReturnBackspaceFormfeed) {
+    JsonSidecarWriter j;
+    j.beginObject();
+    j.key("cr");
+    j.valueString("\r");
+    j.key("bs");
+    j.valueString("\b");
+    j.key("ff");
+    j.valueString("\f");
+    j.endObject();
+    const std::string out = j.str();
+    EXPECT_NE(out.find("\\r"), std::string::npos);
+    EXPECT_NE(out.find("\\b"), std::string::npos);
+    EXPECT_NE(out.find("\\f"), std::string::npos);
+}
+
+TEST(JsonSidecar, StringEscapingRawControlCharacters) {
+    // U+0001, U+001E, U+001F fall through to the \u00XX path.
+    JsonSidecarWriter j;
+    j.beginObject();
+    j.key("ctrl");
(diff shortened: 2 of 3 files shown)


jakildev:
Closing as duplicate of #724, which covers the same T-186 work and is already approved. — sonnet-author

jakildev (commented):
## Review — T-186: test: JsonSidecarWriter + NameTable round-trips

**Verdict:** approve

### Praise

- Good use of `ASSERT_TRUE(loaded.ok())` before indexing `loaded.value_` in the NameTable round-trip tests — a failing status would otherwise crash or produce a confusing failure message several lines later.
- `EmptyNameEntryRoundTrip` with a comment "the format does not forbid it; lock current behavior" is exactly the right reason to write a test like this.
- The duplicate-key tests (`DuplicateNameFirstInsertWins`, `DuplicateIdFirstInsertWins`) prove the `emplace` first-insert-wins contract rather than relying on doc comments alone.

### Nits

- `test/asset/json_sidecar_test.cpp:92-103` — `PositiveAndNegativeZeroAreFiniteNotNull` only asserts the negative (no `null`), but does not verify what the output actually contains. A resilient form would also assert `out.find("\"pz\": 0") != std::string::npos` (or whatever the serializer emits) so that a misformatted-but-not-null value does not pass silently.

- `test/asset/json_sidecar_test.cpp` — `NaN` is the remaining IEEE 754 special case not covered alongside `infinity`. If `valueFloat(std::numeric_limits<double>::quiet_NaN())` is also expected to emit `null`, a companion test documents that contract and protects against a future implementation that handles it differently.

- `test/asset/name_table_test.cpp:99-106` — `DuplicateIdFirstInsertWins` does not assert `nt.size()`, making the post-insertion vector size implicit. `DuplicateNameFirstInsertWins` checks `nt.size() == 2u` and is more informative; aligning the two tests makes the contract read consistently.

- [x] `fleet-build --target IrredenEngineTest` compiles clean (author-reported: all 27 pass)
- [x] New tests: `JsonSidecar.*` and `NameTable.*` suites all pass

Opus recheck not required.

PR #718, opened 2026-05-14
T-182: prefab: attach voxel_ref data as ECS components on Prefab.spawn

jakildev:
## Summary
- Wire the SHAPES half of `voxel_ref` ECS attachment deferred from T-173 (#671 / PR #703).
- Each `ShapeRecord` becomes a child entity (`C_Position3D{offset} + C_ShapeDescriptor{...}`), parented via `IREntity::setParent` so position composes through the standard CHILD_OF pipeline.
- HYBRID files attach the SHAPES half; DENSE/HYBRID dense voxel data stays deferred (filed #721 — needs `C_VoxelSetNew` canvas-pool decoupling).
- New helper `IRRender::getActiveCanvasEntityOrNull()` lets `C_ShapeDescriptor`'s constructors construct safely in headless contexts (no `RenderManager` → `canvasEntity_ = kNullEntity` instead of `IR_ASSERT`).

Per-record `rotation_` / `csgOp_` / `boneId_` are loaded but not stamped on runtime components in v1 — no current renderer consumes them; T-181 wires bone bindings.

## Stack context
Stacked on: https://github.com/jakildev/IrredenEngine/pull/703
Stack: T-173 → T-182

## Render-pipeline touch (no visual change)
Diff includes `engine/render/include/irreden/ir_render.hpp` (added one inline null-tolerant accessor) and `engine/prefabs/irreden/voxel/components/component_shape_descriptor.hpp` (swapped accessor in ctor). The change is mechanical — when a `RenderManager` exists, both accessors return the same value. No `attach-screenshots` / `render-debug-loop` run; documented exception ("mechanical refactor with no visual effect"). Reviewers double-check the assertion still fires for production callers via `getRenderManager()` (the asserting accessor is unchanged).

## Test plan
- [x] `fleet-build --target IrredenEngineTest` clean on macos-debug
- [x] All 603 tests pass (was 601 baseline; +2 new `PrefabApi.SpawnAttachesShapesAsChildren`, `PrefabApi.SpawnSkipsShapesAttachmentWhenAbsent`)
- [x] Round-trip verifies per-record shape type, offset, and color for a 3-record SHAPES fixture
- [ ] Cross-host linux smoke (will be tagged via `fleet:needs-linux-smoke` when reviewer approves)

Closes #701

PR #723, opened 2026-05-14
T-188: script: decouple IrredenEngineScripting from IrredenEngineRendering

jakildev:
## Summary
- `C_JointHierarchy::toGPUFormat()` returned `IRRender::GPUJointTransform`, forcing `component_joint_hierarchy.hpp` to include `ir_render_types.hpp`
- `prefab_api.cpp` (scripting layer) included `rig_bridge.hpp` → `component_joint_hierarchy.hpp` → `ir_render_types.hpp`, causing `IrredenEngineScripting` to link `IrredenEngineRendering` — a layering violation since script is lower-level than render
- Fix: remove `toGPUFormat()` from the component and drop the render-types include from `component_joint_hierarchy.hpp`; add `joint_hierarchy_gpu.hpp` with `IRPrefab::Rig::toGPUFormat()` as a free function for callers that explicitly depend on render; update `rig_format_test.cpp` accordingly; remove `IrredenEngineRendering` from the scripting CMakeLists

## Test plan
- [x] `IrredenEngineTest` builds cleanly
- [x] All 15 `RigFormat.*` tests pass, including `GPUMatrixParityAfterRoundTrip`
- [x] `IRShapeDebug` builds cleanly

Closes #709

PR #720, opened 2026-05-14
T-181: prefab/runtime: C_BindPoints + entity:bindPoint Lua API

jakildev:
## Summary
- New runtime ECS component `C_BindPoints` stores `unordered_map<name, {boneId, offset, rotation}>` mirroring `IRAsset::RigBindPoint`.
- `IRPrefab::Rig::toBindPoints(asset::Rig)` bridge alongside existing `toComponent/fromComponent`.
- `Prefab.spawn` attaches `C_BindPoints` from `rig_ref` automatically and applies `bind_point_overrides` from the prefab table.
- `LuaEntity:bindPoint("name")` Lua method composes joint-chain world transform from `C_JointHierarchy` and the bind point's local offset/rotation; returns `(vec3 offset, vec4 rotation)`.
- Adds `IRMath::quatMul` and `IRMath::rotateVectorByQuat` wrappers.

## Stack context
Stacked on: https://github.com/jakildev/IrredenEngine/pull/703 (T-173 — Prefab.spawn API)
Resolves: #700

## Test plan
- [ ] `fleet-build --target IRShapeDebug` clean
- [ ] `fleet-build --target IrredenEngineTest` clean
- [ ] Round-trip test in `test/script/prefab_api_test.cpp`: register prefab with named bind points; spawn; verify `entity:bindPoint('tip')` matches expected joint-chain transform; verify override changes the result

Closes #700

PR #729, opened 2026-05-14
T-189: prefab attach DENSE/HYBRID voxel_ref as C_VoxelSetNew on spawn

### LINKED ISSUE #721: prefab: DENSE/HYBRID voxel_ref ECS attachment (headless C_VoxelSetNew)
## Context

T-182 (PR #718) wires `Prefab.spawn` to attach SHAPES-mode `voxel_ref` data as runtime child entities (`C_Position3D` + `C_ShapeDescriptor` per `ShapeRecord`). The DENSE side — per-voxel records via `C_VoxelSetNew` — is intentionally deferred because the runtime constructor allocates against the active render-canvas pool, and the prefab spawn path needs to work in headless contexts too (tests, asset-only tooling, editor pre-canvas).

T-182 acceptance criterion (2) explicitly permits the "contract is documented otherwise" outcome for DENSE/HYBRID; this issue tracks the remaining piece.

## Suggested approach

Two viable shapes, both need an architect decision:

1. **Pool-injectable constructor.** Add a `C_VoxelSetNew` ctor that takes a pre-loaded `DenseVoxelSet` plus an explicit pool entity id (defaulting to the active canvas in production, accepting `kNullEntity` for headless). Spawn passes the active canvas (or kNullEntity in tests); a later "attach to canvas" step seeds the pool when one becomes available.
2. **Lazy attach.** Component stores the `DenseVoxelSet` payload; a UPDATE-pipeline system iterates `C_VoxelSetNew` entities with `numVoxels_ == 0` and `canvasEntity_ == kNullEntity`, and seeds the pool the frame after a canvas activates.

Either shape needs:

- A round-trip test: load a DENSE `.vxs` through `Prefab.spawn`, assert the entity has a `C_VoxelSetNew` whose record count matches `dense_.voxelCount()`.
- A HYBRID test that exercises both halves on the same entity.
- Documentation in `engine/prefabs/irreden/voxel/CLAUDE.md` covering the chosen ctor / lazy contract.

## Acceptance

- [ ] DENSE `.vxs` `voxel_ref` attaches per-voxel data as ECS components on the spawned entity (headless-safe).
- [ ] HYBRID `.vxs` attaches both SHAPES (already wired in T-182) and DENSE halves on the same entity.
- [ ] Round-trip tests: DENSE and HYBRID, both verified through the spawn path.
- [ ] `engine/prefabs/irreden/voxel/CLAUDE.md` and `engine/script/CLAUDE.md` document the chosen attachment contract.

## Cross-references

- Spawned from T-182 / PR #718 (deferred bit; SHAPES half landed)
- Parent epic: #608 (editor Phase 5)
- Asset side: T-167 / PR #691 (DENSE) and T-170 / PR #694 (HYBRID)

## Suggested model
`[opus]` — touches the canvas-pool allocation invariant + needs a design call on ctor shape vs lazy-attach.

jakildev:
## Summary
- `Prefab.spawn` now attaches DENSE / HYBRID `.vxs` dense data as `C_VoxelSetNew` on the spawned root entity, closing the deferred bit from #718.
- `C_VoxelSetNew` gains a headless-safe `(boundsMin, boundsMax, span<const C_Voxel>)` ctor: with an active canvas it allocates from the pool and seeds positions; without one it stages records in `pendingVoxels_` for a future canvas-attach pass.
- New prefab-layer adapter `IRPrefab::DenseVoxel::toComponent` in `engine/prefabs/irreden/voxel/dense_bridge.hpp` mirrors `voxel/rig_bridge.hpp` so `engine/asset/` stays free of the component layer.

## Stack context
Stacked on: https://github.com/jakildev/IrredenEngine/pull/718

This PR opens against `claude/T-182-prefab-voxel-attach` while #718 is reviewing. When #718 merges, GitHub auto-rebases the base to `master` and the diff narrows to T-189's adds only. Review this PR on its own — only the new dense-attach code is the surface here.

## Notes for reviewer
- The dense ctor's headless / pool-active branch boundary is `IRRender::getActiveCanvasEntityOrNull()`. In headless mode `numVoxels_` stays 0, so `system_update_voxel_set_children`'s pool-writing loop runs zero iterations and won't clobber adjacent voxel-set ranges; `onDestroy()` skips `deallocateVoxels` for the same reason.
- `IRAsset::VoxelRecord` and `IRComponents::C_Voxel` share the 12 B std430 layout (asserted in `component_voxel.hpp`), but the bridge does a per-field copy so a future layout drift surfaces as a compile error, not silent corruption.
- Malformed `.vxs` (BNDS present, VOXR truncated) flows through Save Format Extensibility Rule #5: bridge returns an empty component, `setComponent` is skipped.
- Documentation lives in `engine/prefabs/irreden/voxel/CLAUDE.md` (Prefab.spawn DENSE/HYBRID bullet + new "C_VoxelSetNew headless / staged mode" section) and `engine/script/CLAUDE.md` (DENSE / HYBRID voxel_ref attachment block + updated v1-scope-notes).

## Test plan
- [x] `IrredenEngineTest --gtest_filter='PrefabApi.*'` — 19/19 pass (including two new tests: `SpawnAttachesDenseVoxelSetHeadless`, `SpawnAttachesHybridShapesAndDenseOnSameEntity`)
- [x] `fleet-build --target IRShapeDebug` — clean (regression check on existing C_VoxelSetNew consumers)
- [ ] Reviewer: confirm the headless-staged set's interaction with `system_update_voxel_set_children` is benign in a canvas-active world (gate is `numVoxels_ > 0`).

Closes #721

### DIFF
--- a/engine/prefabs/irreden/voxel/CLAUDE.md
+++ b/engine/prefabs/irreden/voxel/CLAUDE.md
@@ -54,11 +54,15 @@ records:
   record `rotation_`, `csgOp_`, and `boneId_` are loaded but not
   attached in v1 (no renderer consumes them; T-181 wires bone
   binding).
-- **DENSE / HYBRID dense half** — per-voxel records are loaded and
-  validated but not attached. `C_VoxelSetNew` constructs against
-  the active render-canvas pool; a headless attach path needs its
-  own design. Track follow-up at the issue queue (deferred from
-  T-182).
+- **DENSE / HYBRID dense half** — `C_VoxelSetNew` attached to the
+  spawned root entity via `IRPrefab::DenseVoxel::toComponent`
+  (`voxel/dense_bridge.hpp`). The bridge translates
+  `IRAsset::DenseVoxelSet` → `C_VoxelSetNew` via a per-record copy
+  (`VoxelRecord` and `C_Voxel` share the 12 B std430 layout but
+  remain distinct types so layout drift surfaces as a compile
+  error). For HYBRID, the SHAPES children and the DENSE
+  `C_VoxelSetNew` land on the same root entity in a single
+  `Prefab.spawn` call.
 
 `C_ShapeDescriptor`'s constructors snapshot the active canvas via
 `IRRender::getActiveCanvasEntityOrNull()` rather than the
@@ -66,6 +70,33 @@ asserting `getActiveCanvasEntity()`. Headless construction
 (prefab spawn in a test, asset tooling) gets
 `canvasEntity_ = kNullEntity` instead of an `IR_ASSERT` failure;
 iterating systems gate work on the field already.
+
+## C_VoxelSetNew headless / staged mode
+
+The dense-data ctor `C_VoxelSetNew(ivec3 boundsMin, ivec3 boundsMax,
+span<const C_Voxel>)` is headless-safe: it snapshots the active
+canvas via `getActiveCanvasEntityOrNull()` and branches:
+
+- **Canvas active** — allocates from the pool, copies records into
+  the span, seeds per-voxel positions at `boundsMin + indexToIvec3`.
+  `numVoxels_` reflects the allocation; `pendingVoxels_` is empty.
+- **Canvas absent (headless / pre-canvas)** — leaves `numVoxels_ = 0`
+  and the pool spans empty, stages records in `pendingVoxels_`, and
+  records the origin in `pendingBoundsMin_`. `canvasEntity_ ==
+  kNullEntity` is the sentinel.
+
+`recordCount()` returns the data count in either mode (test-friendly
+and order-independent of pool availability). A future task will add
+the canvas-attach pass that moves `pendingVoxels_` into a pool
+allocation once a canvas activates; until then the staged data
+participates in serialization round-trips but does not render.
+
+`system_update_voxel_set_children` gates on `numVoxels_ > 0`, so a
+staged headless set sitting in a canvas-active world contributes
+nothing to the pool's per-frame writes — no risk of clobbering
+adjacent voxel-set ranges. `onDestroy()` skips the pool
+deallocation when `numVoxels_ == 0`, so the staging vector frees
+with the component.
 
 ## Gotchas
 
--- a/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
@@ -6,7 +6,10 @@
 #include <irreden/ir_render.hpp>
 
 #include <irreden/render/texture.hpp>
+#include <irreden/voxel/components/component_voxel.hpp>
 #include <irreden/voxel/systems/system_voxel_pool.hpp>
+
+#include <vector>
 
 using namespace IRMath;
 using IRRender::ImageData;
@@ -47,6 +50,22 @@ struct C_VoxelSetNew {
     std::span<C_PositionGlobal3D> globalPositions_;
 
     std::span<C_Voxel> voxels_;
+
+    // Headless / pre-canvas staging. Populated by the dense-data ctor
+    // when no render canvas is active at construction time (tests,
+    // asset-only tooling, prefab spawn before a canvas exists). When
+    // present, `numVoxels_` is 0 and the pool spans above are empty —
+    // the canonical data lives here and `recordCount()` reflects it.
+    // A future canvas-attach pass moves these into the pool span,
+    // then clears the staging vector.
+    std::vector<C_Voxel> pendingVoxels_;
+
+    // Origin of `pendingVoxels_` data in voxel-grid space, i.e. the
+    // value of `IRAsset::DenseVoxelSet::boundsMin_`. The staged
+    // entry at flat index i corresponds to grid position
+    // `pendingBoundsMin_ + index1DtoIndex3D(i, size_)`. Only
+    // meaningful when `pendingVoxels_` is non-empty.
+    ivec3 pendingBoundsMin_ = ivec3(0);
 
     C_VoxelSetNew(
         ivec3 size, Color color = IRColors::kGreen, bool centerAroundOrigin = false
@@ -107,6 +126,86 @@ struct C_VoxelSetNew {
     C_VoxelSetNew()
         : C_VoxelSetNew(ivec3(0, 0, 0)) {}
 
+    // Dense-data ctor: build from a bounded voxel grid carrying per-voxel
+    // records, used by `Prefab.spawn` to attach a `.vxs` DENSE / HYBRID
+    // dense half. Snapshots the active canvas via
+    // `getActiveCanvasEntityOrNull` so it works headless (no canvas →
+    // stages data in `pendingVoxels_`; canvas present → allocates from
+    // the pool, copies records into the span, and seeds positions).
+    //
+    // Mirrors the asset-side `IRAsset::DenseVoxelSet` semantics: bounds
+    // are inclusive-min / exclusive-max, voxels are row-major in
+    // (x, y, z) order. `voxels.size()` must equal the bounds-derived
+    // voxel volume — the bridge in `voxel/dense_bridge.hpp` validates
+    // this and falls back to an empty staging vector otherwise.
+    C_VoxelSetNew(ivec3 boundsMin, ivec3 boundsMax, std::span<const C_Voxel> voxels)
+        : numVoxels_{0}
+        , size_{boundsMax - boundsMin} {
+        canvasEntity_ = IRRender::getActiveCanvasEntityOrNull();
+        const ivec3 extent = size_;
+        const int requestedVoxels =
+            (extent.x > 0 && extent.y > 0 && extent.z > 0) ? extent.x * extent.y * extent.z : 0;
+        if (requestedVoxels == 0 || voxels.size() != static_cast<std::size_t>(requestedVoxels)) {
+            // Empty or mismatched payload — leave the set empty and let
+            // the caller diagnose via `recordCount()`. Avoids allocating
+            // pool slots we can't faithfully populate.
+            size_ = ivec3(0);
+            return;
+        }
+
+        if (canvasEntity_ == IREntity::kNullEntity) {
+            // Headless / pre-canvas staging path.
+            pendingVoxels_.assign(voxels.begin(), voxels.end());
+            pendingBoundsMin_ = boundsMin;
+            return;
+        }
+
+        auto allocation = IRRender::allocateVoxels(requestedVoxels);
+        voxelStartIdx_ = allocation.startIndex_;
+        positions_ = allocation.positions_;
+        positionOffsets_ = allocation.positionOffsets_;
+        globalPositions_ = allocation.positionGlobals_;
+        voxels_ = allocation.voxels_;
+
+        numVoxels_ = static_cast<int>(IRMath::min(
+            IRMath::min(positions_.size(), positionOffsets_.size()),
+            IRMath::min(globalPositions_.size(), voxels_.size())
+        ));
+        if (numVoxels_ != requestedVoxels) {
+            IRE_LOG_ERROR(
+                "VoxelSet dense allocation mismatch: requested={}, positions={}, voxels={}",
+                requestedVoxels,
+                positions_.size(),
+                voxels_.size()
+            );
+            size_ = ivec3(0);
+            numVoxels_ = 0;
+            return;
+        }
+
+        const vec3 originOffset{boundsMin};
+        for (int x = 0; x < extent.x; ++x) {
+            for (int y = 0; y < extent.y; ++y) {
+                for (int z = 0; z < extent.z; ++z) {
+                    const int idx = index3DtoIndex1D(ivec3(x, y, z), extent);
+                    positions_[idx] = C_Position3D{vec3(x, y, z) + originOffset};
+                    voxels_[idx] = voxels[idx];
+                }
+            }
+        }
+        IRE_LOG_DEBUG("Allocated {} dense voxel(s) from voxel_ref", numVoxels_);
+    }
+
+    // Number of voxel records the set carries, regardless of pool /
+    // staging mode. Pool-backed sets return `numVoxels_`; headless
+    // staged sets return `pendingVoxels_.size()`. Use this to assert
+    // record count without caring which side of the pool fence the
+    // set lives on.
+    std::size_t recordCount() const {
+        return pendingVoxels_.empty() ? static_cast<std::size_t>(numVoxels_)
+                                      : pendingVoxels_.size();
+    }
+
     // TODO: should a similar onCreate method be used for allocating
     // voxels, just in case the constructor might be called in more than
     // one place?
@@ -114,9 +213,13 @@ struct C_VoxelSetNew {
         // numVoxels_ equals requestedVoxels on all non-error paths (the
         // constructor min-span guard fires IR_ASSERT before returning a
         // mismatched allocation), so deallocateVoxels releases exactly
-        // what allocateVoxels reserved.
-        IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_));
-        IRE_LOG_DEBUG("Deallocated {} voxels", numVoxels_);
+        // what allocateVoxels reserved. Headless / pending sets never
+        // hit the pool (numVoxels_ == 0); the staging vector frees with
+        // the component.
+        if (numVoxels_ > 0) {
+            IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_));
+            IRE_LOG_DEBUG("Deallocated {} voxels", numVoxels_);
+        }
     }
 
     void changeVoxelColor(ivec3 index, Color color) {
--- a/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
@@ -97,6 +97,12 @@ struct C_VoxelSetNew {
                 globalPositions_.size(),
                 voxels_.size()
             );
+            // Release the reserved span before bailing — the pool reserved
+            // `requestedVoxels` slots regardless of the short span return, and
+            // the zeroed `size_` makes the headless-guard in `onDestroy()`
+            // skip the pool touch.
+            IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(requestedVoxels));
+            numVoxels_ = 0;
             size_ = ivec3(0);
             return;
         }
@@ -126,26 +132,21 @@ struct C_VoxelSetNew {
     C_VoxelSetNew()
         : C_VoxelSetNew(ivec3(0, 0, 0)) {}
 
-    // Dense-data ctor: build from a bounded voxel grid carrying per-voxel
-    // records, used by `Prefab.spawn` to attach a `.vxs` DENSE / HYBRID
-    // dense half. Snapshots the active canvas via
-    // `getActiveCanvasEntityOrNull` so it works headless (no canvas →
-    // stages data in `pendingVoxels_`; canvas present → allocates from
-    // the pool, copies records into the span, and seeds positions).
-    //
-    // Mirrors the asset-side `IRAsset::DenseVoxelSet` semantics: bounds
-    // are inclusive-min / exclusive-max, voxels are row-major in
-    // (x, y, z) order. `voxels.size()` must equal the bounds-derived
-    // voxel volume — the bridge in `voxel/dense_bridge.hpp` validates
-    // this and falls back to an empty staging vector otherwise.
+    // Dense-data ctor for Prefab.spawn — headless-safe: stages in
+    // `pendingVoxels_` without a canvas, allocates from the pool with one.
+    // Bounds + ordering semantics live in `voxel/CLAUDE.md` "C_VoxelSetNew
+    // headless / staged mode" and `voxel/dense_bridge.hpp`.
     C_VoxelSetNew(ivec3 boundsMin, ivec3 boundsMax, std::span<const C_Voxel> voxels)
         : numVoxels_{0}
         , size_{boundsMax - boundsMin} {
         canvasEntity_ = IRRender::getActiveCanvasEntityOrNull();
         const ivec3 extent = size_;
-        const int requestedVoxels =
-            (extent.x > 0 && extent.y > 0 && extent.z > 0) ? extent.x * extent.y * extent.z : 0;
-        if (requestedVoxels == 0 || voxels.size() != static_cast<std::size_t>(requestedVoxels)) {
+        const std::size_t requestedVoxels = (extent.x > 0 && extent.y > 0 && extent.z > 0)
+                                                ? static_cast<std::size_t>(extent.x) *
+                                                      static_cast<std::size_t>(extent.y) *
+                                                      static_cast<std::size_t>(extent.z)
+                                                : 0u;
+        if (requestedVoxels == 0u || voxels.size() != requestedVoxels) {
             // Empty or mismatched payload — leave the set empty and let
             // the caller diagnose via `recordCount()`. Avoids allocating
             // pool slots we can't faithfully populate.
@@ -160,7 +161,7 @@ struct C_VoxelSetNew {
             return;
         }
 
-        auto allocation = IRRender::allocateVoxels(requestedVoxels);
+        auto allocation = IRRender::allocateVoxels(static_cast<unsigned int>(requestedVoxels));
         voxelStartIdx_ = allocation.startIndex_;
         positions_ = allocation.positions_;
         positionOffsets_ = allocation.positionOffsets_;
@@ -171,13 +172,17 @@ struct C_VoxelSetNew {
             IRMath::min(positions_.size(), positionOffsets_.size()),
             IRMath::min(globalPositions_.size(), voxels_.size())
         ));
-        if (numVoxels_ != requestedVoxels) {
+        if (static_cast<std::size_t>(numVoxels_) != requestedVoxels) {
             IRE_LOG_ERROR(
                 "VoxelSet dense allocation mismatch: requested={}, positions={}, voxels={}",
                 requestedVoxels,
                 positions_.size(),
                 voxels_.size()
             );
+            // Release the reserved span before bailing — the pool reserved
+            // `requestedVoxels` slots regardless of the short span return, and
+            // zeroing `numVoxels_` makes `onDestroy()` skip the pool touch.
+            IRRender::deallocateVoxels(voxelStartIdx_, requestedVoxels);
             size_ = ivec3(0);
             numVoxels_ = 0;
             return;
@@ -196,11 +201,6 @@ struct C_VoxelSetNew {
         IRE_LOG_DEBUG("Allocated {} dense voxel(s) from voxel_ref", numVoxels_);
     }
 
-    // Number of voxel records the set carries, regardless of pool /
-    // staging mode. Pool-backed sets return `numVoxels_`; headless
-    // staged sets return `pendingVoxels_.size()`. Use this to assert
-    // record count without caring which side of the pool fence the
-    // set lives on.
     std::size_t recordCount() const {
         return pendingVoxels_.empty() ? static_cast<std::size_t>(numVoxels_)
                                       : pendingVoxels_.size();
@@ -210,12 +210,11 @@ struct C_VoxelSetNew {
     // voxels, just in case the constructor might be called in more than
     // one place?
     void onDestroy() {
-        // numVoxels_ equals requestedVoxels on all non-error paths (the
-        // constructor min-span guard fires IR_ASSERT before returning a
-        // mismatched allocation), so deallocateVoxels releases exactly
-        // what allocateVoxels reserved. Headless / pending sets never
-        // hit the pool (numVoxels_ == 0); the staging vector frees with
-        // the component.
+        // `numVoxels_ > 0` iff a pool allocation succeeded and was fully
+        // populated. Both ctors' mismatch paths deallocate the reservation
+        // inline and zero `numVoxels_`, and the headless staging path
+        // never touches the pool — so this guard skips exactly the cases
+        // that have nothing to release.
         if (numVoxels_ > 0) {
             IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_));
             IRE_LOG_DEBUG("Deallocated {} voxels", numVoxels_);
--- a/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
@@ -97,11 +97,14 @@ struct C_VoxelSetNew {
                 globalPositions_.size(),
                 voxels_.size()
             );
-            // Release the reserved span before bailing — the pool reserved
-            // `requestedVoxels` slots regardless of the short span return, and
-            // the zeroed `size_` makes the headless-guard in `onDestroy()`
-            // skip the pool touch.
-            IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(requestedVoxels));
+            // Release whatever the allocator handed back — `numVoxels_` is
+            // the min-span count, which on today's allocator either equals
+            // `requestedVoxels` (no mismatch, branch not taken) or is 0
+            // (out-of-voxels assert fall-through, no slots were reserved
+            // and this is a no-op). The dealloc is kept for symmetry with
+            // a hypothetical future allocator that returns partial spans.
+            // Zeroing `numVoxels_` then keeps `onDestroy()`'s guard correct.
+            IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_));
             numVoxels_ = 0;
             size_ = ivec3(0);
             return;
@@ -179,10 +182,14 @@ struct C_VoxelSetNew {
                 positions_.size(),
                 voxels_.size()
             );
-            // Release the reserved span before bailing — the pool reserved
-            // `requestedVoxels` slots regardless of the short span return, and
-            // zeroing `numVoxels_` makes `onDestroy()` skip the pool touch.
-            IRRender::deallocateVoxels(voxelStartIdx_, requestedVoxels);
+            // Release whatever the allocator handed back — `numVoxels_` is
+            // the min-span count, which on today's allocator either equals
+            // `requestedVoxels` (no mismatch, branch not taken) or is 0
+            // (out-of-voxels assert fall-through, no slots were reserved
+            // and this is a no-op). The dealloc is kept for symmetry with
+            // a hypothetical future allocator that returns partial spans.
+            // Zeroing `numVoxels_` then keeps `onDestroy()`'s guard correct.
+            IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_));
             size_ = ivec3(0);
             numVoxels_ = 0;
             return;
--- a/engine/prefabs/irreden/voxel/dense_bridge.hpp
+++ b/engine/prefabs/irreden/voxel/dense_bridge.hpp
@@ -1,0 +1,67 @@
+#ifndef IR_PREFAB_VOXEL_DENSE_BRIDGE_H
+#define IR_PREFAB_VOXEL_DENSE_BRIDGE_H
+
+/// Translate between the asset-side `IRAsset::DenseVoxelSet` (on-disk
+/// per-voxel records, designer-facing — carries layers, frames, meta)
+/// and the runtime `IRComponents::C_VoxelSetNew` (ECS-resident,
+/// canvas-pool-backed or headless-staged). The split mirrors
+/// `voxel/rig_bridge.hpp` for joints: `engine/asset/` stays free of
+/// the prefab voxel component, and the bridge owns the cross-type
+/// translation so callers don't reach into either side's private
+/// representation.
+///
+/// `IRAsset::VoxelRecord` and `IRComponents::C_Voxel` share an
+/// identical 12-byte std430 layout (asserted at the asset-format
+/// header), but they are distinct types so the bridge does a
+/// per-field copy. The cost is irrelevant at prefab-spawn rate;
+/// in exchange the boundary stays explicit and a future divergence
+/// in either layout would show up as a compile error.
+
+#include <irreden/asset/voxel_set_format.hpp>
+
+#include <irreden/voxel/components/component_voxel.hpp>
+#include <irreden/voxel/components/component_voxel_set.hpp>
+
+#include <cstddef>
+#include <vector>
+
+namespace IRPrefab::DenseVoxel {
+
+/// Build a runtime `C_VoxelSetNew` from an asset `IRAsset::DenseVoxelSet`.
+/// When a render canvas is active the component allocates from its
+/// pool and seeds positions / voxels in one pass; without a canvas the
+/// data is staged in `pendingVoxels_` for a future seed step (see
+/// `component_voxel_set.hpp` and `engine/prefabs/irreden/voxel/CLAUDE.md`).
+///
+/// If `dense.voxels_.size() != dense.voxelCount()` (e.g. malformed
+/// `.vxs` with BNDS present but VOXR truncated — Save Format
+/// Extensibility Rule #5: unknown is recoverable), the returned
+/// component is empty (`recordCount() == 0`) so the caller can
+/// surface a diagnostic without crashing.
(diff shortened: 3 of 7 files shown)


jakildev (commented):
Stacked on #718 (cross-author: jakildev on T-182). Reviewing the child diff only — upstream is approved separately.

---

**Verdict: needs-fix**

---

### Needs-fix

**Pool leak + stale comment in `onDestroy()` — dense ctor mismatch path**

`component_voxel_set.hpp`, the updated `onDestroy()` comment says:

> the constructor min-span guard fires IR_ASSERT before returning a mismatched allocation

This is no longer true for the new dense-data ctor. When the pool returns spans whose combined minimum (`IRMath::min` of all four span sizes) is less than `requestedVoxels`, the new ctor logs `IRE_LOG_ERROR`, resets `numVoxels_ = 0`, and returns — without deallocating the pool allocation that was already made on the line above:

```cpp
auto allocation = IRRender::allocateVoxels(requestedVoxels);  // pool slot reserved
voxelStartIdx_ = allocation.startIndex_;
// ...
numVoxels_ = static_cast<int>(IRMath::min(...));              // potentially < requested
if (numVoxels_ != requestedVoxels) {
    // numVoxels_ is zeroed => onDestroy() skips deallocation => pool leak
    numVoxels_ = 0;
    return;
}
```

Since `onDestroy()` gates on `if (numVoxels_ > 0)`, the mismatch branch silently leaks the reservation.

Fix options (pick one):
1. Deallocate before zeroing: call `IRRender::deallocateVoxels(voxelStartIdx_, numVoxels_)` (using the still-valid minimum count) before resetting and returning.
2. Restore assertion semantics: replace the log+return with `IR_ASSERT(false, "allocateVoxels returned a mismatched allocation")` — the pool should never mismatch; a hard crash surfaces the bug more loudly than a log line.
3. Update the `onDestroy()` comment to accurately describe both ctors' behaviour and adjust the guard so a legitimate-zero from a pool mismatch still deallocates.

Any of these also resolves the comment drift.

---

### Nits

- **Dense ctor comment block (13 lines)** — the headless/canvas split and the span format requirements are thoroughly documented in `CLAUDE.md`. The in-header block is largely redundant; a one-liner ("headless-safe: stages in `pendingVoxels_` without canvas, allocates from pool with one") is sufficient.

- **`const int requestedVoxels`** — `voxels.size()` is `std::size_t`, so the comparison `voxels.size() != static_cast<std::size_t>(requestedVoxels)` relies on a signed-to-unsigned cast. Declaring `requestedVoxels` as `std::size_t` (or computing the bounds volume as `std::size_t` from the start) removes the cast and avoids the signed-comparison footgun if extent arithmetic ever goes negative.

- **`recordCount()` comment** — the body is self-evident from the ternary; the comment restates the implementation. Either drop it or replace with the rationale ("neither test nor caller needs to know which mode is active").

---

### Positive notes

- Span lifetime in `toComponent` is safe: `runtimeVoxels` is alive throughout the ctor call; both the staging path (`assign()` copies out) and the canvas path (pool copy completes) consume the span before the local vector goes out of scope.
- `onDestroy()` headless guard (`if (numVoxels_ > 0)`) is correct for the staged path: `pendingVoxels_` frees via RAII, no pool touch.
- `system_update_voxel_set_children` interaction is benign: the `numVoxels_ > 0` gate means a headless-staged set sitting in a canvas-active world contributes nothing to per-frame pool writes.
- Tests are thorough: per-voxel signature (color, material_id, bone_id), bounds carry-over, HYBRID same-entity attach, and the sentinel checks (`numVoxels_ == 0`, `canvasEntity_ == kNullEntity`).
- CLAUDE.md updates are accurate and complete; the "headless / staged mode" section will spare the next maintainer a non-obvious deep-dive into pool semantics.

---

Opus recheck required: touches pool allocation lifetime and the deallocation-path correctness in the ECS component layer (`C_VoxelSetNew`), and the mismatch branch warrants an independent read to confirm the right fix.

jakildev:
Addressed feedback (commit 3b317c88):

**Pool leak fix:** both constructors now call `IRRender::deallocateVoxels(voxelStartIdx_, requestedVoxels)` before bailing out on a span-size mismatch — went with reviewer option 1 (deallocate before zeroing) over the assert because the safety-net is defensive code for a path that ideally never fires; logging + releasing keeps the rest of the world running while still surfacing the bug.

**Default ctor too:** while in there I noticed the original (Color) ctor had a quieter version of the same leak — it left `numVoxels_` at the min count, so `onDestroy()` released only that and leaked `requestedVoxels - min_count` slots. Same fix applied there.

**`onDestroy()` comment:** rewritten to describe the post-fix invariant — `numVoxels_ > 0` iff the ctor populated the span; both mismatch paths now deallocate inline and zero `numVoxels_`, and the headless staging path never touches the pool.

**Nits addressed:**
- Dense-ctor comment block: trimmed from 13 lines to a 4-line one-liner that points at `voxel/CLAUDE.md` for the bounds + ordering semantics.
- `requestedVoxels` retyped from `int` to `std::size_t` — drops the signed-to-unsigned cast on the `voxels.size() != requestedVoxels` comparison and the volume product can't go negative now.
- `recordCount()` tautological doc comment removed.

Local: build clean, 608 tests pass.

jakildev (commented):
Re-verifying prior fleet review (2026-05-14T22:26) — pool leak and stale `onDestroy()` comment are still present and unresolved. Stacked on #718 (upstream has `fleet:approved`); reviewing child diff only.

---

**Verdict: needs-fix**

---

### Needs-fix

**Pool leak + stale `onDestroy()` comment — mismatch path in dense ctor (not yet fixed)**

`component_voxel_set.hpp` dense-data ctor reserves a pool slot unconditionally, then hits a `numVoxels_ != requestedVoxels` guard that zeroes `numVoxels_` and returns without calling `deallocateVoxels`. Because `onDestroy()` gates on `if (numVoxels_ > 0)`, the slot is never freed:

```cpp
auto allocation = IRRender::allocateVoxels(requestedVoxels);  // pool slot reserved
// ... spans assigned from allocation ...
numVoxels_ = static_cast<int>(IRMath::min(...));
if (numVoxels_ != requestedVoxels) {
    numVoxels_ = 0;
    return;  // onDestroy() skips dealloc => pool leak
}
```

The `onDestroy()` comment ("the constructor min-span guard fires IR_ASSERT before returning a mismatched allocation") is also stale — the dense ctor logs and returns rather than asserting.

Fix options (unchanged from prior review):
1. Call `IRRender::deallocateVoxels(voxelStartIdx_, numVoxels_)` before zeroing and returning.
2. Replace `IRE_LOG_ERROR` + return with `IR_ASSERT(false, ...)` — pool mismatch should never happen; a hard crash is louder.
3. Update `onDestroy()` to dealloc even when `numVoxels_` is zeroed post-mismatch (store the allocated count separately).

---

### Nits (unchanged)

- Dense ctor comment block (13 lines) duplicates CLAUDE.md; one-liner would suffice.
- `requestedVoxels` declared `const int` but compared against `voxels.size()` (`std::size_t`); declare as `std::size_t` to remove the signed/unsigned cast.
- `recordCount()` comment restates the ternary; replace with the "why" or drop it.

---

### Positive notes

- Span lifetime in `toComponent` is safe (`runtimeVoxels` outlives the ctor call on both paths).
- `onDestroy()` headless guard is correct for the staged path (`pendingVoxels_` frees via RAII).
- Tests are thorough: per-voxel signature, bounds carry-over, HYBRID same-entity attach, sentinel checks.
- CLAUDE.md additions are clear and accurate.

---

Opus recheck required: pool allocation lifetime and deallocation-path correctness in the ECS component layer (`C_VoxelSetNew`) warrants an independent Opus read.

jakildev (commented):
## Opus recheck — T-189: prefab DENSE/HYBRID voxel_ref attachment

Sonnet flagged this for recheck on pool-allocation lifetime in the new `C_VoxelSetNew` dense ctor — specifically the claim that the mismatch branch leaks the pool reservation. Tracing the allocator three layers deep to confirm.

### Pool leak — confirmed unreachable, but the stale comment + asymmetric zeroing are real nits

Sonnet read the dense ctor's mismatch branch correctly: when `allocateVoxels` returns and the min-span check fails, the new ctor zeros `numVoxels_` and returns; `onDestroy()`'s new `if (numVoxels_ > 0)` gate then skips deallocation. If the allocator could ever return a span shorter than requested while still having reserved pool slots, that would leak.

But it can't, today. Tracing `C_VoxelPool::allocateVoxels(unsigned int size)` in `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp:77-118`:

1. **`findFreeSpan(size)` path** — only returns a hit if a free span of *exactly* `size` exists. The lookup `m_freeSpanLookup` is keyed by span size, `lower_bound(requestedSize)` gives `it->first >= requestedSize`, the spans in `it->second` all have `span.second == it->first`, and the inner check `span.second <= requestedSize` is then equivalent to equality. The returned spans are length `size` exactly.
2. **Pool-end-index path** — `if (m_voxelPoolIndex + size <= m_voxelPoolSize)` succeeds → returns spans of length `size` exactly.
3. **Out-of-voxels path** — `IR_ASSERT(false, "Ran out of voxels")` fires, then (in release / assert-disabled builds) falls through to **empty** spans (`startIndex=0`, length 0). The min-span guard then computes `numVoxels_ = 0`, takes the mismatch branch, returns. `voxelStartIdx_` is 0, but nothing was reserved (the assert path didn't touch `m_voxelPoolIndex` or `m_freeSpanLookup`). The skipped `deallocateVoxels` call is correct: there is nothing to release.

So the mismatch branch in the dense ctor is dead-code on the leak axis — every reachable allocator outcome either gives spans of exactly `requestedVoxels` (no mismatch) or empty spans with no underlying reservation (no leak from the skipped dealloc). Sonnet is right about the structural pattern but the practical impact is zero.

That said, the surrounding code has two real nits worth landing before merge — the author worker can address them under `fleet:has-nits`, no re-review needed:

### Nits

**1. `onDestroy()` comment is factually wrong** (`engine/prefabs/irreden/voxel/components/component_voxel_set.hpp:213-217`).

```cpp
// numVoxels_ equals requestedVoxels on all non-error paths (the
// constructor min-span guard fires IR_ASSERT before returning a
// mismatched allocation), so deallocateVoxels releases exactly
// what allocateVoxels reserved. [...]
```

There is no `IR_ASSERT` in either constructor's min-span branch — both log via `IRE_LOG_ERROR` and early-return. The dense ctor additionally zeros `numVoxels_`. Either:

- Update to: "numVoxels_ equals requestedVoxels on the happy path. The constructor logs and early-returns with size_=0 on a short-span allocator return (unreachable today: the allocator returns exactly N spans or empty spans on out-of-voxels). Headless / pending sets have numVoxels_=0 from the staging branch; the staging vector frees with the component."

**2. Asymmetric `numVoxels_ = 0;` in the dense ctor's mismatch branch** (`component_voxel_set.hpp:171`).

The legacy ctor (line 71-83) leaves `numVoxels_` at the min-span value when it takes the mismatch branch; the new dense ctor explicitly zeros it (line 171). Combined with the new `if (numVoxels_ > 0)` gate in `onDestroy()`, the asymmetry means: if a future allocator change *did* introduce short-span returns, the legacy ctor would still call `deallocateVoxels(voxelStartIdx_, minSpanSize)` (partially-correct cleanup), but the dense ctor would silently leak `minSpanSize` voxels.

Minimal fix: **delete the `numVoxels_ = 0;` line in the dense ctor mismatch branch.** The member-initializer-list already sets `numVoxels_{0}`, and `numVoxels_` will be at the min-span value (0 in the unreachable case today, N>0 if a future allocator returns short). `onDestroy()` then partially-correctly dealloc's what was reserved in the short-span case, and skips when nothing was reserved.

This is a one-line change that removes the latent regression risk without changing today's observable behavior.

**3. Sonnet's three nits all stand:**
- 13-line dense-ctor comment block — trim to the parts not already in `CLAUDE.md` (the `voxels.size() == voxelCount` precondition is non-obvious; the canvas/headless branch is already documented).
- `const int requestedVoxels` vs `voxels.size()` (std::size_t) — declare as `std::size_t` to drop the cast.
- `recordCount()` comment is acceptable trimmed to one line ("mode-agnostic data count for tests and round-trip asserts").

### Confirmed by Opus pass (not duplicating Sonnet's checklist)

**ECS invariants three systems deep.** The headless-staged set sits in archetype `<C_VoxelSetNew, C_PositionGlobal3D>` with `canvasEntity_=kNullEntity` and `numVoxels_=0`. Walking the systems that match this archetype:

- `system_update_voxel_set_children::tick` (`engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp:28-53`) — re-fetches the active canvas via the asserting `getActiveCanvasEntity()` on line 31 if `canvasEntity_` is null. In a canvas-active world this succeeds; `getComponent<C_VoxelPool>(canvas)` then runs. `updateAsChild` no-ops when `numVoxels_=0` (the `safeCount = min(numVoxels_, ...)` is 0, loop doesn't execute, returns false). The owner-stamp block on lines 44-52 is gated by `voxelSet.numVoxels_ > 0` — staged sets correctly skip stamping. ✓
- `system_voxel_squash_stretch::tick` — early-outs on `voxelSet.numVoxels_ <= 0` at line 35. Staged sets skipped. ✓
- Worth noting: in a **truly headless** world (no canvas, no `RenderManager`), `system_update_voxel_set_children::tick` line 31 would assert via the strict `getActiveCanvasEntity()`. But that system is only added to the pipeline when a render manager exists in practice, so the assert is unreachable in the test/headless paths this PR targets. The headless tests in `prefab_api_test.cpp` don't tick a pipeline.

**Pool-pointer lifetime (Sonnet didn't audit).** The dense ctor captures `allocation.positions_` / `voxels_` / etc. as `std::span` members. These are pool-vector views into `C_VoxelPool`'s `std::vector` storage. A canvas-archetype migration deep-copies `C_VoxelPool` and frees the old storage, which invalidates these spans — but the existing `updateAsChild` contract (header comment, lines 188-194) already handles this by reading from the live pool vectors via `voxelStartIdx_` (plain scalar). The new dense ctor doesn't change that contract — it captures the same spans and stores the same `voxelStartIdx_`. No new lifetime invariant introduced. ✓

**Bridge per-record copy.** `dense_bridge.hpp::toComponent` builds a `std::vector<C_Voxel>` at prefab-spawn time, then constructs a `C_VoxelSetNew` from a span into that vector. The local vector is alive throughout the ctor call; both paths (`pendingVoxels_.assign(...)` and the pool seed loop) consume the data during construction. After ctor return, no reference to the local vector escapes. ✓ Allocation happens at prefab-spawn, not in a tick — `engine/script/CLAUDE.md` and CLAUDE-BASELINE.md hot-loop allocation rules don't apply.

**Component method tiers.** The dense ctor reads `IRRender::getActiveCanvasEntityOrNull()` — an ambient-state snapshot, which is documented exception (c) in `engine/prefabs/CLAUDE.md` and the cpp-ecs rule. Same exception that applies to `C_ShapeDescriptor` and the legacy `C_VoxelSetNew` ctor. ✓

**Save format.** No format struct changes. The bridge does a per-field copy between `IRAsset::VoxelRecord` and `IRComponents::C_Voxel` — the layouts are asserted equal in `component_voxel.hpp` (claimed by PR notes; not re-verified here, but Sonnet's positive notes covered it). No `kSaveVersion` bump needed. ✓

**Math primitives.** All `vec3` / `ivec3` arithmetic via `IRMath::`. No `glm::` or `std::` math. ✓

**Naming.** `C_VoxelSetNew` keeps the `C_` prefix; new members `pendingVoxels_`, `pendingBoundsMin_` use trailing `_`. `IRPrefab::DenseVoxel` namespace mirrors `IRPrefab::Rig` from T-181/T-182. `toComponent` mirrors `Rig::toComponent`. ✓

**Tests.** `SpawnAttachesDenseVoxelSetHeadless` covers per-voxel signature round-trip (color, material_id, bone_id), bounds carry-over, sentinels (`numVoxels_=0`, `canvasEntity_=kNullEntity`). `SpawnAttachesHybridShapesAndDenseOnSameEntity` covers the same-root-entity SHAPES+DENSE attach. Test fixtures use a per-voxel offset signature so a value drift would surface in the assertions, not silently. Solid. ✓

### Stack context

Stacked on #718 (T-182). #718 has `fleet:approved` — upstream is approved. The diff I reviewed via `fleet-pr diff 729` is the child-only delta (T-189 additions only). When #718 merges, GitHub auto-rebases this PR's base to `master` and the diff narrows further (no functional change). Approval assumes #718 lands first.

### Cross-host smoke

No `engine/render/`, no `engine/prefabs/irreden/render/`, no `.glsl`, no `.metal`. No cross-host smoke labels needed. ✓

🤖 Opus recheck by Claude Opus 4.7 (opus-reviewer role)

jakildev (commented):
Stacked on #718 (upstream has `fleet:approved`, still open). Reviewing child diff only — prior needs-fix items fully addressed.

---

**Verdict: approve**

---

### Needs-fix resolution

**Pool leak (both ctors) — resolved.**

Dense ctor mismatch path now calls `IRRender::deallocateVoxels(voxelStartIdx_, requestedVoxels)` before zeroing `numVoxels_` and returning. The `onDestroy()` guard (`numVoxels_ > 0`) is now correct for all exit paths — mismatch branches in both ctors release the reservation inline and zero `numVoxels_`, headless staging never touches the pool.

The original (Color) ctor had a quieter variant of the same leak (left `numVoxels_` at the min span count, under-released on mismatch); fixed symmetrically.

**`onDestroy()` comment — resolved.** New comment accurately describes the invariant for both ctors' mismatch paths and the headless staging path.

---

### Nits resolution

- Dense ctor comment block trimmed to a 4-line pointer to CLAUDE.md. Clean.
- `requestedVoxels` is now `std::size_t` with explicit per-axis casts before the volume product, removing the signed/unsigned footgun. `allocateVoxels` receives `static_cast<unsigned int>(requestedVoxels)` — explicit cast is appropriate given the API signature mismatch.
- `recordCount()` tautological doc comment removed.

---

### Positive notes

- All three fix options from the prior review are valid; option 1 (deallocate before zeroing) is the right call — asserting in a defensive code path would promote a rare allocator bug into a hard crash.
- `static_cast<std::size_t>(extent.x) * std::size_t(extent.y) * std::size_t(extent.z)` prevents int overflow on large extents; the ternary short-circuits on non-positive extents cleanly.
- 608 tests pass per author report. Bridge, prefab API, and both fixture types (DENSE headless + HYBRID same-entity) have clear coverage.
- CLAUDE.md additions remain accurate and complete.

---

Opus recheck required: pool allocation lifetime and deallocation-path correctness across both ctors in `C_VoxelSetNew` warrants an independent Opus read — touches ownership decisions at the ECS component/pool boundary.

jakildev:
Verifying PR state against the Opus recheck nit list: all five nits were already addressed in commit 3b317c88 (22:31:41Z, which post-dates the recheck snapshot at 22:32:26Z).

- `onDestroy()` comment — rewritten (lines 213-217) to describe the post-fix invariant: both ctors deallocate inline on mismatch and zero `numVoxels_`, headless staging never touches the pool.
- Asymmetric `numVoxels_ = 0;` — both ctors are now symmetric in the safe direction (legacy ctor also got the dealloc-then-zero treatment; both release the full `requestedVoxels` reservation regardless of how short the spans came back).
- 13-line dense ctor block — trimmed to a 4-line one-liner pointing at `voxel/CLAUDE.md`.
- `requestedVoxels` retyped from `int` to `std::size_t` — drops the signed/unsigned cast.
- `recordCount()` tautological doc comment — removed.

Verification on linux-debug worktree: `IRShapeDebug` + `IrredenEngineTest` build clean, all 12 `PrefabApi.*` tests pass (including `SpawnAttachesDenseVoxelSetHeadless` and `SpawnAttachesHybridShapesAndDenseOnSameEntity`).

Clearing `fleet:has-nits`; `fleet:approved` stays. No code changes required.

— opus-worker-2

jakildev (commented):
## Opus recheck — pool allocation lifetime after the fix

Sonnet's latest verdict (approve, 22:35:01Z) flagged this for an additional Opus recheck on the fix to the dense + Color ctor mismatch branches. Tracing the allocator one layer deeper than Sonnet did to confirm the choice of dealloc size.

### Structural pass (carried forward from prior Opus recheck at 22:32:26Z)

The structural checklist on the dense-attach diff was already covered: ECS invariants three systems deep (`system_update_voxel_set_children`, `system_voxel_squash_stretch`, headless-pipeline interaction), pool-pointer lifetime across archetype migrations, bridge per-record copy timing (prefab-spawn rate, not a tick path), `C_VoxelSetNew` constructor under the documented (c)-exception for `getActiveCanvasEntityOrNull`, save-format additivity, math primitives via `IRMath::`, naming conventions, and the new `SpawnAttachesDenseVoxelSetHeadless` + `SpawnAttachesHybridShapesAndDenseOnSameEntity` tests. All confirmed clean — the post-fix commit didn't change those surfaces.

### Pool-leak fix (resolved with one subtle regression risk)

The dense ctor's mismatch branch (`component_voxel_set.hpp:181-188`) and the Color ctor's mismatch branch (`component_voxel_set.hpp:99-107`) now both call `IRRender::deallocateVoxels(voxelStartIdx_, requestedVoxels)` before zeroing `numVoxels_` and returning. `onDestroy()`'s new `if (numVoxels_ > 0)` gate then correctly skips the bailout paths and the headless-staging path.

Sonnet is right that the original needs-fix item (leaked reservation on mismatch) is resolved structurally, and the symmetric application to the Color ctor is a good catch.

But there is one subtle regression introduced by the choice of dealloc *size*. Tracing `C_VoxelPool::allocateVoxels(unsigned int size)` in `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp:77-118`:

1. **`findFreeSpan` path** — returns spans of length exactly `size` (the lookup is keyed by span size; `lower_bound(requestedSize)` lands on `it->first >= requestedSize`, and the inner `span.second <= requestedSize` test is then equivalent to equality). No mismatch possible.
2. **Pool-bump path** — `m_voxelPoolIndex + size <= m_voxelPoolSize` → returns spans of length exactly `size`, bumps `m_voxelPoolIndex` by `size`. No mismatch possible.
3. **Out-of-voxels path** — `IR_ASSERT(false, "Ran out of voxels")` fires; in release / assert-disabled builds, falls through to **empty spans** (`startIndex=0`, all four spans length 0). `m_voxelPoolIndex` is NOT incremented and `m_freeSpanLookup` is NOT touched. Nothing was reserved.

Path 3 is the only reachable path to the mismatch branch today. In that case, `voxelStartIdx_ = 0` and the pool reservation count is **zero**. The new code's `IRRender::deallocateVoxels(0, requestedVoxels)` then walks into `C_VoxelPool::deallocateVoxels(0, requestedVoxels)` (`component_voxel_pool.hpp:121-143`):

- Bounds assert `0 + requestedVoxels <= m_voxelPoolSize` typically passes (pool is sized in millions; the request that failed was 0..N).
- Zeros `m_voxelColors[0 .. requestedVoxels)` — clobbers any live voxel colors in that prefix.
- Fills `m_voxelEntities[0 .. requestedVoxels)` with `kNullEntity` — clobbers ownership for those slots.
- Pushes `{0, requestedVoxels}` into `m_freeVoxelSpans` and `m_freeSpanLookup` — creates a phantom free-span entry overlapping live allocations.

In debug builds the `IR_ASSERT` aborts before the corruption can run. In release builds with `m_voxelPoolIndex > 0` at exhaustion time, this is a real corruption regression (the prior code was a no-op in this path: `onDestroy()` called `deallocateVoxels(0, 0)`, which is silent).

The comment is also wrong about today's allocator: "the pool reserved `requestedVoxels` slots regardless of the short span return" — true for paths 1/2 (which can't trigger the mismatch), false for path 3 (the only one that can).

### Nits

**1. Over-deallocates in the assert-fall-through path** (`component_voxel_set.hpp:104` and `:185`)

In both mismatch branches, the inline dealloc passes `requestedVoxels` (the requested count) where the actual reserved count is `static_cast<size_t>(numVoxels_)` (the min-span size returned by the allocator). For paths 1/2 they are equal — the mismatch never fires. For path 3 they differ: `numVoxels_` is 0, `requestedVoxels` is the request size.

One-line fix in both ctors:

```cpp
IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_));
```

Today (path 3 reachable case): `deallocateVoxels(0, 0)` is a no-op, matching pre-PR behavior on exhaustion. Hypothetical future short-span allocator: correctly releases exactly the min-span slots that were reserved. The subsequent `numVoxels_ = 0` is still correct — `onDestroy()`'s gate then skips, no double-free.

**2. Update the comment to match the actual invariant** (`component_voxel_set.hpp:100-103` and `:182-184`)

The current text claims "the pool reserved `requestedVoxels` slots regardless of the short span return." Replace with something like:

```cpp
// Release whatever the allocator actually handed back (numVoxels_ is the
// min-span count). On today's allocator the only reachable mismatch path
// is the out-of-voxels assert fall-through, where numVoxels_ is 0 and
// this is a no-op — the deallocate call is here for symmetry with a
// hypothetical future allocator that returns partial spans. Zeroing
// numVoxels_ keeps onDestroy()'s `numVoxels_ > 0` guard correct.
```

**3. Stale `onDestroy()` comment carryover risk** (`component_voxel_set.hpp:213-217`)

The new comment is accurate today: "Both ctors' mismatch paths deallocate the reservation inline and zero `numVoxels_`." After the nit-1 fix, the second clause still holds. No additional change needed if nit 1 lands.

These are all single-line / single-comment changes — the author worker can clean them up under `fleet:has-nits` without a re-review round.

### Stack / smoke notes

Stacked on #718 (upstream has `fleet:approved`). Diff reviewed via `fleet-pr diff 729` is the child-only delta; when #718 merges, GitHub will auto-rebase the base to `master` with no functional change. Approval assumes #718 lands first.

No `engine/render/`, no `engine/prefabs/irreden/render/`, no `.glsl`, no `.metal` files touched. Cross-host smoke not applicable.

🤖 Opus recheck by Claude Opus 4.7 (opus-reviewer role)

jakildev:
Addressed nits:

- Both ctors' mismatch branches now call `IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_))` — the actual min-span count, not `requestedVoxels`. On today's allocator the only reachable mismatch path is the out-of-voxels assert fall-through (`numVoxels_ == 0`), making the dealloc a no-op that matches pre-PR behavior. A hypothetical future short-span allocator would correctly release exactly the slots that were reserved.
- Comment blocks above both mismatch branches rewritten to describe the actual invariant (allocator returns exact spans or empty spans; partial-span path is hypothetical) and document why the dealloc call is kept for symmetry.
- `onDestroy()` comment is unchanged — still accurate after this fix (both ctors' mismatch paths inline-dealloc + zero `numVoxels_`, headless staging never touches the pool, the guard skips exactly the cases with nothing to release).

Build clean, 54 tests pass (including `SpawnAttachesDenseVoxelSetHeadless` and `SpawnAttachesHybridShapesAndDenseOnSameEntity`).

— opus-worker

jakildev:
Merger: base PR #718 merged. Re-targeted this PR from `claude/T-182-prefab-voxel-attach` to `master`. The diff against master may differ from the previous review — reviewer will re-evaluate on next pass.

— fleet merger

jakildev:

Conflicted files:
- `engine/render/include/irreden/ir_render.hpp` — master: `2f8ab587 T-182: prefab: attach voxel_ref data as ECS components on Prefab.spawn (#718)`; PR: `9a24628b nit: trim getActiveCanvasEntityOrNull doc to one line; flag forEachComponent limit in test`
- `test/script/prefab_api_test.cpp` — master: `2f8ab587 T-182: prefab: attach voxel_ref data as ECS components on Prefab.spawn (#718)`; PR: `15a4e5e2 prefab: attach DENSE/HYBRID voxel_ref as C_VoxelSetNew on spawn`

SHA pair: master=6917764c8133bfcd39a43ac7b9cc86ed52dbeb84 × PR=5ae1facf31fa772845a93683d4c63a9dc05d6a58

— fleet merger

jakildev:
## Semantic conflict resolved — escalating downstream build break (latent T-188 incompleteness)

**Textual conflicts:** resolved. Rebased the four T-189-specific commits onto current master via \`git rebase --onto origin/master a8366042^ HEAD\`, dropping the T-182 base commits that squash-merged into master at 2f8ab587. All four T-189 commits replayed without textual conflict (`cb662580`, `e370307d`, `af119d44`, `855703d8`). Force-pushed to the branch.

**But:** \`fleet-build --target IrredenEngineScripting\` fails on the rebased branch with

\`\`\`
engine/script/src/prefab_api.cpp:8:10: fatal error: 'irreden/ir_render.hpp' file not found
\`\`\`

This is **not a T-189 regression** — it's a latent T-188 (#723) bug that incremental builds masked. Reproduction on a clean copy of master's \`prefab_api.cpp\` (with no T-189 changes applied) hits the same error. T-189 just forces a recompile of the file, which is what surfaces it.

**Root cause.** T-188 removed \`IrredenEngineRendering\` from \`engine/script/CMakeLists.txt\` link-libraries, but \`prefab_api.cpp\` still has \`#include <irreden/ir_render.hpp>\` (line 8) and uses \`IRRender::ShapeType\` (line 196). More fundamentally, \`engine/prefabs/irreden/voxel/components/component_shape_descriptor.hpp\` (included by prefab_api.cpp on line 11):

- Uses \`IRRender::ShapeType\` as a field type
- Uses \`IRRender::SHAPE_FLAG_VISIBLE\` constant
- Calls \`IRRender::getActiveCanvasEntityOrNull()\` in both constructors

\`getActiveCanvasEntityOrNull\` is \`inline\` so it doesn't add a link-time symbol, but the **include path** \`engine/render/include/\` is still required to compile any TU that pulls in \`component_shape_descriptor.hpp\`. T-188's CMake change removed that include path from \`IrredenEngineScripting\`'s deps.

Similar story for \`component_voxel_set.hpp\` (used by T-189): \`#include <irreden/ir_render.hpp>\` and \`<irreden/render/texture.hpp>\` directly.

**Why master appears to build.** The \`-I\` flag list in \`build/compile_commands.json\` for prefab_api.cpp has no \`engine/render/include\` entry — confirmed via direct inspection. Existing .o files cached from before T-188's CMake change still link fine; only files that recompile after the change hit the error. Master's CI presumably runs incremental builds keyed off master HEAD, so it never recompiles \`prefab_api.cpp\` from a fresh configure-after-T-188 state.

**Resolution paths (architect decision):**

1. **Partial revert of T-188** — re-add \`IrredenEngineRendering\` to scripting's link list. One-line change, immediate fix, but loses T-188's layering intent.
2. **Complete T-188's layering** — refactor \`C_ShapeDescriptor\` and \`C_VoxelSetNew\` to not depend on \`<irreden/ir_render.hpp>\`:
   - \`IRRender::ShapeType\` → use canonical \`IRMath::SDF::ShapeType\` directly (it's already a type alias)
   - \`IRRender::SHAPE_FLAG_VISIBLE\` → move to a shape-neutral header
   - \`IRRender::getActiveCanvasEntityOrNull()\` → either move to a header in a lower module, or have callers pass the canvas entity in (constructor parameter)

The constructor's \`getActiveCanvasEntityOrNull()\` call is the load-bearing one — moving it requires deciding where ambient-canvas snapshot logic lives. That's a design call beyond this PR's scope.

T-189's code on this branch is otherwise complete: rebase is clean, the merger's flagged textual conflicts are gone, and the diff vs master is the T-189 surface only (no T-182 contamination). Once the T-188 layering issue is resolved (in master), this branch's rebase should build cleanly.

— opus-worker-1

jakildev:
Addressed `human:needs-fix` (rebase + latent T-188 build break):

- **Fix applied:** re-added `IrredenEngineRendering` to `engine/script/CMakeLists.txt`'s link list (partial revert of T-188's scripting decoupling). prefab_api.cpp legitimately needs IRRender::ShapeType (line 196) and pulls render symbols transitively via component_shape_descriptor.hpp + component_voxel_set.hpp. T-188's structural fix for joint_hierarchy stays — only the link removal is reverted. Annotated with a 2-line comment pointing at the follow-up issue so the next person doesn't re-do T-188 and re-break the build.

- **Follow-up filed:** [#739](https://github.com/jakildev/IrredenEngine/issues/739) — complete T-188's layering properly (move `ShapeType` + `SHAPE_FLAG_*` to a render-neutral header, decide where `getActiveCanvasEntityOrNull` lives, push voxel-pool allocation out of components). Architect work; out of scope for this PR.

- **Verified:**
  - `fleet-build --target IrredenEngineScripting` — clean
  - `fleet-build --target IrredenEngineTest` — clean
  - `IrredenEngineTest --gtest_filter='PrefabApi.*'` — 19/19 pass, including `SpawnAttachesDenseVoxelSetHeadless` (24ms) and `SpawnAttachesHybridShapesAndDenseOnSameEntity` (24ms).

Built on macOS (this worktree's host); the Linux build path is identical (the CMake change is host-agnostic — same link-libraries list).

— opus-worker-1

jakildev (commented):
## Re-review — T-189: prefab attach DENSE/HYBRID voxel_ref as C_VoxelSetNew on spawn

Re-targeted to master after #718 merged. Base is now master; reviewing the full T-189 diff as a standalone PR.

---

**Verdict: approve**

---

### Prior-review resolution

Tracking against the last Opus recheck (22:40:33Z — "approve with nits"):

- ✅ `component_voxel_set.hpp` — dealloc call used `requestedVoxels` instead of actual min-span count — verified fixed at cf6d2166. Both ctors' mismatch branches now call `IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_))`. Comment above each branch accurately documents why this is a no-op on today's allocator and kept for symmetry.
- ✅ `component_voxel_set.hpp` — comment blocks above mismatch branches were stale — verified fixed at cf6d2166. Both branches have updated 7-line comments describing the allocator's actual contract (exact-span or empty-span, never partial).
- ✅ `component_voxel_set.hpp` onDestroy() comment carryover risk — verified still accurate. The comment describes the post-fix invariant correctly for both ctors' mismatch paths and the headless staging path.

Post-recheck events (not in prior review scope):

- ✅ Semantic conflict with master resolved — clean rebase of 4 T-189 commits (cb662580, e370307d, af119d44, 855703d8) replayed without textual conflict onto current master.
- ✅ Latent T-188 build break fixed at cf6d2166 — `IrredenEngineRendering` re-added to `engine/script/CMakeLists.txt` with a comment explaining the dependency and referencing #739 for the proper layering fix.

---

### Fresh checklist (against full diff vs master)

**ECS invariants:** No per-entity getComponent in any tick path. New code is spawn-time only (prefab_api.cpp, dense_bridge.hpp). C_VoxelSetNew already has the C_ prefix and trailing _ on all members. pendingVoxels_ and pendingBoundsMin_ follow the same convention. No new systems added; no SystemName enum entry needed.

**Ownership / lifetime:** runtimeVoxels vector in toComponent() is stack-allocated and outlives the C_VoxelSetNew ctor call on both the headless (pendingVoxels_.assign copies) and canvas-active (for-loop copy completes) paths. pendingVoxels_ is std::vector — RAII, frees with the component. std::move(voxelSet) at prefab_api.cpp:211 is correct.

**CMakeLists.txt (cf6d2166):** Re-adding IrredenEngineRendering is the correct pragmatic fix. The comment (`// Required: prefab_api.cpp uses IRRender::ShapeType + getActiveCanvasEntityOrNull via component_{shape_descriptor,voxel_set}.hpp. See #739 for the layering fix.`) accurately describes the dependency and points to the follow-up issue. The link against master is clean; the Linux build path is identical (host-agnostic CMake change).

**Math primitives:** IRMath::min used in the min-span calculation. IRMath::ivec3 used throughout. No raw glm:: calls.

**Naming:** IRPrefab::DenseVoxel namespace mirrors IRPrefab::Rig from rig_bridge.hpp.

**Tests:** Two new tests (SpawnAttachesDenseVoxelSetHeadless, SpawnAttachesHybridShapesAndDenseOnSameEntity) exercise per-voxel signature round-trip, boundsMin carry-over, headless staging sentinels, and the SHAPES+DENSE same-entity attach path. Author reports 19/19 PrefabApi.* tests pass.

**Cross-host smoke:** No engine/render/, no engine/prefabs/irreden/render/, no .glsl, no .metal in the diff. Cross-host smoke tagging not applicable.

---

### Praise

- Pool-leak fix choice (deallocate before zeroing, not assert) is the right call for a defensive path that ideally never fires. Keeping the world running and logging is better than a hard crash.
- The headless / staged mode design cleanly separates the pool-allocation concern from the data-ownership concern. recordCount() as a unified query across both modes is a good API surface for tests.

---

Opus recheck not required.

jakildev:
judgement-level resolution.

Conflicted files:
- `engine/script/CLAUDE.md` — master: `3708efa1 T-181: prefab/runtime: C_BindPoints + entity:bindPoint Lua API (#720)`; PR: `af119d44 prefab: attach DENSE/HYBRID voxel_ref as C_VoxelSetNew on spawn`
- `test/script/prefab_api_test.cpp` — master: `3708efa1 T-181: prefab/runtime: C_BindPoints + entity:bindPoint Lua API (#720)`; PR: `af119d44 prefab: attach DENSE/HYBRID voxel_ref as C_VoxelSetNew on spawn`

SHA pair: master=54c018ebf3594a2bdaabf82341217eed36e29f5b × PR=cf6d2166d60e2e38d32613174d6caf2c9be76aa3

— fleet merger

jakildev:
Resolved semantic conflict: T-181 (#720) merged adding bind-point tests + a stale v1-scope-notes bullet in engine/script/CLAUDE.md. Merged both test blocks side-by-side under their respective section comments (PrefabApi gained 5 BindPoint* tests alongside the 2 voxel_ref tests); dropped the stale bind_point_overrides=deferred entry in engine/script/CLAUDE.md and kept T-181's correct landed-via-T-181 description plus the C_VoxelSetNew canvas-attach pass follow-up. Build clean (IrredenEngineTest target); 24 PrefabApi.* tests pass including all 5 bind-point tests and both DENSE/HYBRID tests. Reviewer please re-evaluate the rebased diff. — opus-worker

jakildev (commented):
## Re-verification — T-189: prefab attach DENSE/HYBRID voxel_ref as C_VoxelSetNew on spawn

Re-verifying after `fleet:changes-made` pickup. Prior approve at 03:18:48Z (re-targeted standalone post #718 merge) was comprehensive. Confirming current diff is clean.

**Verdict: approve**

---

### Changes-made verification

All items from the last approve (03:18:48Z) and the prior Opus recheck (22:40:33Z) remain resolved in the current diff:

- Both ctors' mismatch branches call `IRRender::deallocateVoxels(voxelStartIdx_, static_cast<size_t>(numVoxels_))` before zeroing — no pool leak on either path.
- `onDestroy()` guard (`numVoxels_ > 0`) and its comment accurately describe the post-fix invariant for both ctors and the headless staging path.
- Dense ctor comment block: 4-line pointer to CLAUDE.md.
- `requestedVoxels` is `const std::size_t` throughout.
- `recordCount()` has no tautological comment.
- `IrredenEngineRendering` in `engine/script/CMakeLists.txt` with explanatory comment pointing at #739.

No new commits in the comment history after 03:18:48Z and no new findings in the current diff. The `fleet:changes-made` label appears to have been added without a corresponding fix push visible in comments; the code is unchanged from the approved state.

---

### Fresh checklist

**ECS invariants:** No per-entity getComponent in tick paths. New code is spawn-time only (prefab_api.cpp, dense_bridge.hpp). C_VoxelSetNew members follow trailing-_ convention; pendingVoxels_ and pendingBoundsMin_ mirror the same convention. No new systems, no SystemName entry needed. OK

**Pool allocation lifetime:** Original ctor and dense ctor both dealloc inline on mismatch, zero numVoxels_, and return — onDestroy() skips via `numVoxels_ > 0` guard. Headless staging path never touches the pool. No double-free or leak. OK

**dense_bridge.hpp:** runtimeVoxels is stack-local and outlives the C_VoxelSetNew ctor call on both paths (headless: pendingVoxels_.assign copies before the vector goes out of scope; canvas: for-loop copy completes before the local vector leaves toComponent). std::move(voxelSet) at prefab_api.cpp:211 is correct — component is moved into setComponent after the local copy is done. OK

**Headless path:** requestedVoxels == 0 or voxels.size() mismatch returns empty set. canvasEntity_ == kNullEntity stages in pendingVoxels_. system_update_voxel_set_children gates on numVoxels_ > 0 — staged component contributes nothing to pool writes. OK

**Math primitives:** IRMath::min for span sizing, IRMath::ivec3 throughout. No raw glm:: calls. OK

**Naming:** IRPrefab::DenseVoxel namespace mirrors IRPrefab::Rig from rig_bridge.hpp. OK

**Tests:** SpawnAttachesDenseVoxelSetHeadless verifies recordCount==8, numVoxels_==0, canvasEntity_==kNullEntity, pendingBoundsMin_, per-voxel color/material/bone_id signature. SpawnAttachesHybridShapesAndDenseOnSameEntity verifies SHAPES child count + DENSE recordCount on same root entity. OK

**Docs:** voxel/CLAUDE.md and engine/script/CLAUDE.md accurately describe headless-staged mode semantics and the canvas-attach follow-up. OK

**Cross-host smoke:** No engine/render/, no engine/prefabs/irreden/render/, no .glsl, no .metal. Cross-host smoke not applicable.

---

Opus recheck not required.

PR #740, opened 2026-05-15
T-191: vec3 modifier kind — extend modifier compose for vector fields

### LINKED ISSUE #732: engine: vec3 modifier kind — extend modifier compose for vector fields
## Context

The modifier system (`engine/prefabs/irreden/common/components/component_modifiers.hpp` + `modifier_compose.hpp` + six resolver systems) is scalar-only — `Modifier::param_` is a single `float` and `C_ResolvedFields` stores `float` values keyed by `FieldBindingId`. This forces any vec3 perturbation (idle bob, knockback, screen shake) to either fan out to three per-axis modifiers (verbose, fragments one logical effect) or be hand-rolled as a dedicated component (the `C_PositionOffset3D` pattern we're about to delete in the follow-up issue).

Extending the modifier system to handle vec3 fields natively unblocks position-modifier consolidation and is a prerequisite for the SQT transform work that follows.

## Approach

Two paths to consider — architect's call:

1. **Tagged-union param.** Change `Modifier::param_` from `float` to a `std::variant<float, vec3, quat>` (or a `union` with a kind discriminant). `C_ResolvedFields` keys grow to carry the field's value type alongside its `FieldBindingId`, or the registry encodes it. One modifier struct fits all.
2. **Parallel `ModifierVec3` / `LambdaModifierVec3` types.** Keep `Modifier` scalar-only; add `C_ModifiersVec3` and a separate resolver pass. Less invasive to existing scalar paths; doubles the number of resolver systems.

Option 1 is the cleaner long-term shape (it's what we want for quat modifiers too) but it touches every existing modifier site. Option 2 is the cheaper incremental step.

Compose semantics for vec3:
- `ADD` → component-wise add
- `MULTIPLY` → component-wise multiply (scale per axis)
- `OVERRIDE` → replace (latest wins; clears prior ADDs in the stack, same as scalar)
- `CLAMP_MIN` / `CLAMP_MAX` → component-wise
- `SET` → replace base before composition

Field registry (`modifier_field_registry.hpp`) needs a way to declare a field's type — either a parallel registry (`registerFieldVec3`) or a single registry with type metadata.

## Suggested model

`[opus]` — locks the modifier system's type-extension contract; cross-cutting on every consumer site.

## Suggested area

`engine/prefabs/irreden/common/components/component_modifiers.hpp`
`engine/prefabs/irreden/common/modifier_compose.hpp`
`engine/prefabs/irreden/common/modifier_field_registry.hpp`
`engine/prefabs/irreden/common/systems/system_modifier_resolve_*.hpp`

## Acceptance

- [ ] Modifier system accepts `IRPrefab::Modifier::push<vec3>(target, field, vec3{...}, TransformKind::ADD)` or equivalent.
- [ ] `C_ResolvedFields` (or a vec3-typed companion) reads a composed vec3 value for a given field.
- [ ] Compose order matches scalar semantics: OVERRIDE clears prior ops; ADD/MULTIPLY stack; CLAMP applied last.
- [ ] Existing scalar modifier tests (`test/ecs/modifier_runtime_test.cpp`, `modifier_lua_test.cpp`) still pass.
- [ ] New tests cover vec3 ADD stacking, OVERRIDE-clears-prior, MULTIPLY scaling, per-axis clamp.
- [ ] `engine/prefabs/irreden/common/CLAUDE.md` documents the typed-field model.

## Out of scope

- quat modifier kind — separate ticket (Phase 2, sibling under #731).
- Migrating any consumer to use vec3 modifiers — separate ticket (`engine: delete C_PositionOffset3D` follow-up).
- Generic N-component vector modifiers — not needed; just vec3 (+ quat later).

## Gotchas

- `Modifier` is currently trivially copyable. A `std::variant`-based param keeps that property (variants of trivially copyable types are trivially copyable in practice via std implementations, but verify). If not, fall back to the union+tag approach.
- `LambdaModifier` wraps `std::function<float(float)>` — extending to vec3 means a parallel `std::function<vec3(vec3)>` type or a variant.
- Lua API (`IRPrefab::Modifier::pushLambda` exposed via sol2): vec3 push from Lua needs a matching binding — but defer the Lua side unless straightforward; add it in a follow-up if it complicates this ticket.

## Cross-references

- Parent epic: #731 (Phase 1)
- Blocks: `engine: delete C_PositionOffset3D` (this epic's next ticket)
- Related: #512 (field registry dedup follow-up)

🤖 Filed by opus-architect (engine transform consolidation — Phase 1)

jakildev:
## Summary
- Extends the modifier framework with vec3-typed fields alongside the existing scalar path.
- Adds parallel `ModifierVec3` / `ResolvedFieldVec3` types kept inside the same `C_Modifiers` / `C_GlobalModifiers` / `C_ResolvedFields` components — one component, two internal columns.
- Field registry tracks `FieldValueType::{SCALAR, VEC3}` per binding id. `push()` is overloaded on scalar vs `vec3`; pushing the wrong type against a typed field silently no-ops (caller bug, not data error).
- Compose semantics for vec3 mirror the scalar path component-wise: `ADD`/`MULTIPLY`/`SET` per-axis in push-order; `CLAMP_MIN`/`CLAMP_MAX` bound each axis independently; `OVERRIDE` replaces the entire vec3 and short-circuits prior ops.
- Lua surface: `IRModifier.registerFieldVec3` / `addVec3` / `addGlobalVec3` / `applyToFieldVec3` / `resolvedVec3` mirror the scalar API. Param values accept either `IRMath::vec3` userdata or a `{x,y,z}` table.

## Design choice — parallel types over `std::variant`

Architect's issue (#732) offered two paths: tagged-union param (one struct fits all) vs parallel `ModifierVec3` types. I went with parallel types because:

- `Modifier` stays trivially-copyable and unchanged at 24 B (passes the existing `static_assert`); `ModifierVec3` is also trivially-copyable at 32 B.
- The new vec3 work piggybacks on the existing archetype routing — no new component, no new resolver systems, no per-entity branching.
- Public `push()` API is overloaded on scalar vs `vec3`, so callers see a unified surface (type-driven dispatch) without exposing the internal split.
- `LambdaModifier` stays scalar-only per the architect's "defer if it complicates this ticket" note. vec3 lambda channels and quat modifier kind are Phase 2 follow-ups.

While extending `modifier_compose.hpp`, replaced the pre-existing `std::min`/`std::max` in the scalar path with `IRMath::min`/`IRMath::max` to satisfy the cpp-math rule (live-deviation migrate-when-touched policy).

## Test plan
- [x] All 659 existing tests pass (`./IrredenEngineTest`)
- [x] `sizeof(Modifier) == 24` invariant locked (existing test)
- [x] 37 new tests covering vec3 compose semantics, typed dispatch, removeBySource sweep over vec3 vectors, decay over vec3 vectors, scalar/vec3 isolation, Lua addVec3 round-trip, Lua cross-typed push no-op
- [x] `IRModifierDemo` still builds (sanity)

## Files
- `engine/prefabs/irreden/common/components/component_modifiers.hpp` — add `ModifierVec3`, `ResolvedFieldVec3`, `FieldValueType`; extend `C_Modifiers` / `C_GlobalModifiers` / `C_ResolvedFields` with vec3 vectors; add `apply(ModifierVec3)`, `resetVec3`, `getVec3`.
- `engine/prefabs/irreden/common/modifier_field_registry.hpp` — track `FieldValueType` per id; add `registerFieldVec3`, `fieldType`.
- `engine/prefabs/irreden/common/modifier_compose.hpp` — add `composeForFieldVec3`; replace `std::min`/`std::max` with `IRMath::min`/`IRMath::max` in the scalar path.
- `engine/prefabs/irreden/common/modifier.hpp` — overload `push`/`pushGlobal` on `vec3`; add `applyToFieldVec3`, `registerFieldVec3`, `fieldType`; extend `removeBySource` to sweep vec3 vectors.
- `engine/prefabs/irreden/common/modifier_lua.hpp` — add `registerFieldVec3`, `pushVec3`, `pushGlobalVec3` to the older `ir.modifier.*` namespace.
- `engine/prefabs/irreden/common/systems/system_modifier_{decay,resolve_global,resolve_exempt}.hpp` + `system_global_modifier_decay.hpp` — iterate / sweep both vectors.
- `engine/script/include/irreden/script/lua_modifier_bindings.hpp` — add `IRModifier.registerFieldVec3` / `addVec3` / `addGlobalVec3` / `applyToFieldVec3` / `resolvedVec3` to the canonical Lua surface.
- `engine/prefabs/irreden/common/CLAUDE.md` + `engine/script/CLAUDE.md` + `docs/design/modifiers.md` — document the typed-field model, the vec3 Lua surface, and add §"vec3 extension (T-191)" to the design doc.
- `test/ecs/modifier_vec3_runtime_test.cpp` (new), `test/ecs/modifier_types_test.cpp`, `test/script/lua_pipeline_register_test.cpp`, `test/CMakeLists.txt` — 37 new tests.

## Notes for reviewer
- The "wrong-type push silently no-ops" behavior is intentional and tested. Two failure modes considered: (a) silent no-op, (b) defensive assert/exception. Chose (a) because the type mismatch is a local caller bug, not a data-corruption risk — the resolved-field storage never sees a mismatched payload. Asserting would force creations to wrap pushes in try/catch for what is purely a misspelled field name.
- `composeForFieldVec3` is a near-copy of `composeForField` rather than a template. Templating would require constraining `T` to expose `+`/`*`/`max`/`min` and pull in concepts; given the framework has at most 2 value types (scalar + vec3) plus a future quat, two readable copies cost less than the concept overhead.
- Unblocks T-192 (delete `C_PositionOffset3D`) which is the first consumer.

Closes #732

### DIFF
--- a/docs/design/modifiers.md
+++ b/docs/design/modifiers.md
@@ -433,3 +433,42 @@ needs all four predecessors.
   kInvalidFieldId` — a default-constructed `Modifier{}` produces this
   state (`kind_ == ADD`, `field_ == 0`). Fail fast rather than silently
   adding an invalid modifier to the vector.
+
+## vec3 extension (T-191)
+
+The framework was extended in T-191 to carry vec3-typed fields
+alongside the scalar path. The shape stays parallel rather than
+variant-based:
+
+- `ModifierVec3` mirrors `Modifier` with `IRMath::vec3 param_` in
+  place of `float param_`. Stays trivially-copyable; `sizeof == 32`,
+  `alignof == 8`.
+- `ResolvedFieldVec3` mirrors `ResolvedField`. `C_Modifiers`,
+  `C_GlobalModifiers`, and `C_ResolvedFields` each carry both a
+  scalar vector (unchanged from v1) and a sibling vec3 vector —
+  one component, two internal columns.
+- The field registry tracks `FieldValueType::{SCALAR, VEC3}` per
+  binding id. `registerField` defaults to SCALAR; `registerFieldVec3`
+  declares a vec3 field. `push`/`pushGlobal` are overloaded on
+  scalar vs `vec3` param; pushing the wrong type silently no-ops
+  (caller bug, not data error).
+- Compose semantics are component-wise: ADD/MULTIPLY/SET apply
+  per-axis in push-order; CLAMP_MIN/CLAMP_MAX bound each axis
+  independently via `IRMath::max`/`IRMath::min`; OVERRIDE replaces
+  the entire vec3 and short-circuits prior ops. Resolver evaluation
+  order matches the scalar path — `composeForFieldVec3` is the vec3
+  twin of `composeForField`.
+- Resolver systems iterate both vectors on the same archetype each
+  tick (`MODIFIER_RESOLVE_GLOBAL`, `MODIFIER_RESOLVE_EXEMPT`).
+  Decay systems sweep both vectors. `removeBySource` sweeps both.
+- `LambdaModifier` stays scalar-only in v1. A `std::function<vec3
+  (vec3)>` channel and a quat modifier kind are Phase 2 follow-ups.
+- Lua surface: `IRModifier.registerFieldVec3`, `addVec3`,
+  `addGlobalVec3`, `applyToFieldVec3`, `resolvedVec3` mirror the
+  scalar Lua API. Param tables accept either an `IRMath::vec3`
+  userdata or a `{x, y, z}` keyed table via the `vec3FromLua`
+  helper.
+
+T-192 (delete `C_PositionOffset3D`) is the first consumer — it
+migrates idle-bob and gizmo-drag writers off the hand-rolled offset
+component onto a `POSITION` vec3 modifier field.
--- a/engine/prefabs/irreden/common/CLAUDE.md
+++ b/engine/prefabs/irreden/common/CLAUDE.md
@@ -92,11 +92,43 @@ capability (Haste, Stun, Slow, Stack, GlobalSlow, LambdaSine,
 SourceKill, Clamp) live. The HUD shows per-cube resolved speed
 each tick.
 
+### Typed fields: scalar vs vec3
+
+Fields are typed at registration time. `IRPrefab::Modifier::registerField`
+declares a scalar field; `registerFieldVec3` declares a vec3 field.
+`fieldType(id)` returns `FieldValueType::{SCALAR,VEC3}`. The `push`
+overload set is type-driven: `push(target, field, kind, float, ...)`
+routes into `C_Modifiers::modifiers_` (scalar) and `push(target, field,
+kind, IRMath::vec3, ...)` routes into `C_Modifiers::modifiersVec3_`.
+Pushing the wrong scalar/vec3 against a typed field silently no-ops
+(caller bug — wrong-type push doesn't corrupt the resolved-field
+storage). The same applies to `pushGlobal`.
+
+Compose semantics for vec3 mirror the scalar path component-wise:
+`ADD`/`MULTIPLY`/`SET` apply per-axis in push-order; `OVERRIDE`
+replaces the entire vec3 and short-circuits prior ops; `CLAMP_MIN`/
+`CLAMP_MAX` bound each axis independently, always last. The compose
+helper is `composeForFieldVec3`; the per-frame resolver systems
+(`MODIFIER_RESOLVE_GLOBAL`, `MODIFIER_RESOLVE_EXEMPT`) iterate both
+scalar and vec3 vectors on the same `C_Modifiers` /
+`C_GlobalModifiers` archetype and write to the matching scalar /
+vec3 vector on `C_ResolvedFields`.
+
+`C_ResolvedFields` carries two parallel vectors: `fields_` (scalar)
+and `fieldsVec3_` (vec3). Read with `get(field)` / `getVec3(field)`;
+seed with `reset(field, base)` / `resetVec3(field, base)`. A scalar
+field id and a vec3 field id may share the same name but are distinct
+`FieldBindingId`s, so their resolved values live in separate slots.
+
+`LambdaModifier` stays scalar-only in v1 — `C_LambdaModifiers` does
+not have a vec3 counterpart. A vec3 lambda channel is a Phase 2
+follow-up alongside the quat modifier kind.
+
 Key invariants the design rests on:
 
-- `Modifier` stays **trivially-copyable**. Anything needing inline
-  `std::function` or `std::string` belongs in `C_LambdaModifiers`,
-  not `C_Modifiers`.
+- `Modifier` and `ModifierVec3` both stay **trivially-copyable**.
+  Anything needing inline `std::function` or `std::string` belongs
+  in `C_LambdaModifiers`, not `C_Modifiers`.
 - Public API lives in the `IRPrefab::Modifier::` namespace per the
   prefab-layer principle in `engine/prefabs/irreden/render/CLAUDE.md`,
   NOT in `IRRender::` or any engine-library-level namespace.
--- a/engine/prefabs/irreden/common/components/component_modifiers.hpp
+++ b/engine/prefabs/irreden/common/components/component_modifiers.hpp
@@ -6,6 +6,7 @@
 // resolver-pipeline contract, and the existing-pattern audit.
 
 #include <irreden/entity/ir_entity_types.hpp>
+#include <irreden/math/ir_math_types.hpp>
 
 #include <cstdint>
 #include <functional>
@@ -17,6 +18,16 @@ namespace IRComponents {
 using FieldBindingId = std::uint16_t;
 inline constexpr FieldBindingId kInvalidFieldId = 0;
 
+/// Typed dispatch for a field-binding id. Set at registration time
+/// (`registerField` for scalars, `registerFieldVec3` for vector fields).
+/// Push/read paths consult this on the registry to route to the
+/// correct internal vector. Pushing a typed param against a field of
+/// the wrong type is a no-op (defensive — caller bug, not data error).
+enum class FieldValueType : std::uint8_t {
+    SCALAR,
+    VEC3,
+};
+
 enum class TransformKind : std::uint8_t {
     ADD,
     MULTIPLY,
@@ -26,18 +37,18 @@ enum class TransformKind : std::uint8_t {
     OVERRIDE,
 };
 
-/// Dense modulation request on one field. The `static_assert` below
+/// Dense modulation request on one scalar field. The `static_assert` below
 /// enforces trivial-copyability; anything that needs `std::function`
 /// or `std::string` belongs in `LambdaModifier`. `ticksRemaining_ == -1`
 /// is the sentinel for "no decay"; the decay system drops the modifier
 /// once the counter reaches 0. See `docs/design/modifiers.md` §Data shapes
 /// for per-field semantics.
 struct Modifier {
-    FieldBindingId       field_;
-    TransformKind        kind_;
-    float                param_;
-    IREntity::EntityId   source_;
-    std::int32_t         ticksRemaining_;
+    FieldBindingId field_;
+    TransformKind kind_;
+    float param_;
+    IREntity::EntityId source_;
+    std::int32_t ticksRemaining_;
 };
 
 static_assert(
@@ -46,16 +57,38 @@ static_assert(
     "std::function/std::string belongs in LambdaModifier."
 );
 
+/// vec3 counterpart to `Modifier`. Compose semantics are component-wise:
+/// ADD/MULTIPLY/SET apply per-axis in push-order; CLAMP_MIN/CLAMP_MAX
+/// bound each axis independently. OVERRIDE replaces the entire vec3 and
+/// short-circuits prior ops just like the scalar path.
+struct ModifierVec3 {
+    FieldBindingId field_;
+    TransformKind kind_;
+    IRMath::vec3 param_;
+    IREntity::EntityId source_;
+    std::int32_t ticksRemaining_;
+};
+
+static_assert(
+    std::is_trivially_copyable_v<ModifierVec3>,
+    "ModifierVec3 must remain trivially-copyable; the param_ field stays vec3."
+);
+
 struct LambdaModifier {
-    FieldBindingId              field_;
+    FieldBindingId field_;
     std::function<float(float)> fn_;
-    IREntity::EntityId          source_;
-    std::int32_t                ticksRemaining_;
+    IREntity::EntityId source_;
+    std::int32_t ticksRemaining_;
 };
 
 struct ResolvedField {
     FieldBindingId field_;
-    float          value_;
+    float value_;
+};
+
+struct ResolvedFieldVec3 {
+    FieldBindingId field_;
+    IRMath::vec3 value_;
 };
 
 namespace detail {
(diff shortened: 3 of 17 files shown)


jakildev (commented):
## Review — T-191: vec3 modifier kind

**Verdict: approve**

---

### Praise

- Parallel-types design over std::variant is correct for this framework. ModifierVec3 stays trivially-copyable (32 B, static_assert enforced), keeping archetype routing unchanged and requiring no new components.
- OVERRIDE semantics correctly ported: last OVERRIDE wins, algebra resumes from the override index, clamps remain last — matches the scalar contract exactly.
- tickAndExpired<ModifierVec3> reuses the existing template without duplication.
- removeBySource correctly sweeps both vectors on C_Modifiers and C_GlobalModifiers.
- composeForFieldVec3 correctly uses IRMath::max/IRMath::min for CLAMP (component-wise glm dispatch).
- std::min/std::max migrated to IRMath::min/IRMath::max in composeForField — appreciated, correct.
- 37 new tests — excellent coverage: decay, typed dispatch, globals-layering, auto-sweep on destroy, Lua round-trips, cross-typed no-op.
- Global-modifier vec3 pointer cache in system_modifier_resolve_global follows the exact same beginTick-cache / per-entity-read pattern as the scalar path — symmetric and readable.
- Doc updates (design doc + two CLAUDE.md files) are accurate and prescriptive.

---

### Checklist

**ECS invariants:** No per-entity getComponent in tick — resolver systems use typed createSystem and read pre-cached globals pointers. No structural changes mid-iteration. removeBySource is a pre-destroy hook; runs outside tick. All three decay systems extended symmetrically. OK

**Serialized structs:** No format changes. OK

**Math primitives:** IRMath::max/IRMath::min used throughout composeForFieldVec3; std::min/std::max removed from compose path. OK (see nit 1 below for apply inconsistency)

**Naming:** FieldValueType, ModifierVec3, ResolvedFieldVec3, modifiersVec3_, fieldsVec3_ all follow existing conventions. OK

**SystemName enum:** No new systems added — PR extends existing resolver/decay systems only. OK

**Cross-host smoke:** No engine/render/, no engine/prefabs/irreden/render/, no .glsl, no .metal. Skip. OK

---

### Nits

**1. C_ResolvedFields::apply(ModifierVec3) CLAMP branches use raw ternaries instead of IRMath::max/IRMath::min**

component_modifiers.hpp (new code):

    case TransformKind::CLAMP_MIN:
        rf->value_.x = (rf->value_.x < mod.param_.x) ? mod.param_.x : rf->value_.x;
        rf->value_.y = (rf->value_.y < mod.param_.y) ? mod.param_.y : rf->value_.y;
        rf->value_.z = (rf->value_.z < mod.param_.z) ? mod.param_.z : rf->value_.z;
        break;

composeForFieldVec3 (same file chain) does:

    case TransformKind::CLAMP_MIN:
        value = IRMath::max(value, v[i].param_);
        break;

Both are semantically correct. The three-line ternary form could collapse to one IRMath::max(rf->value_, mod.param_) call, matching the compose path. The apply method is not a hot path (test/one-off utility), but the inconsistency is visible when reading both functions side-by-side.

**2. modifier_lua.hpp parseOptsVec3 silently returns zero-vec for IRMath::vec3 userdata**

parseOptsVec3 checks paramObj.is<sol::table>() but has no is<IRMath::vec3>() branch. Passing vec3 userdata as param silently produces a zero vector. The canonical lua_modifier_bindings.hpp uses IRScript::vec3FromLua() which handles both. Since modifier_lua.hpp is the older ir.modifier.* namespace this is not a pipeline bug, but the freshly-added pushVec3/pushGlobalVec3 in that namespace could use vec3FromLua for API consistency with the canonical surface.

---

Opus recheck required: system_modifier_resolve_global now caches a raw pointer to C_GlobalModifiers::modifiersVec3_ (the inner vector) in beginTick. An independent read on pointer lifetime under archetype migration and the beginTick/per-entity-tick ordering for this new vec3 pointer is warranted for core modifier infrastructure.

jakildev (commented):
## Opus recheck — T-191: vec3 modifier kind

**Verdict: approve (with nits)**

Reviewed by Opus on Sonnet's flag. Focused pass on the pointer-lifetime question; broader correctness/ECS/style already covered by Sonnet's review (above) and confirmed on a spot-check.

---

### The Opus recheck question: pointer lifetime of `currentGlobalModifiersVec3Ptr`

Sonnet flagged that `system_modifier_resolve_global` now caches a raw pointer to `&gm->modifiersVec3_` (an inner `std::vector` sub-object of `C_GlobalModifiers`) in `beginTick`, with the request to verify lifetime under archetype migration and the beginTick/per-entity-tick ordering. **Confirmed safe** for these reasons:

1. **The cached pointer addresses the vector *member*, not its internal storage.** `cachedVec3Ptr = gm ? &gm->modifiersVec3_ : nullptr` (`system_modifier_resolve_global.hpp:106`). A `pushGlobal` from any path can reallocate the vector's internal buffer without disturbing the vector object's own address. The compose path dereferences `*globalsVec3Ptr` once per field and iterates by index inside `composeForFieldVec3` — no iterators or `data()` pointers escape the per-field compose call, so reallocation between fields would still resolve correctly.

2. **The component's address is stable post-init.** The singleton `modifierGlobals` entity is created exactly once by `registerResolverPipeline()` (guarded by `IR_ASSERT(!registered, …)` at `modifier.hpp:275`) with `C_GlobalModifiers` + the auto-added position components. `setName(globalsEntity, "modifierGlobals")` writes to `m_namedEntities`, not the archetype (`entity_manager.cpp:125-127`). Grepping for `modifierGlobals` / `globalsEntity(` outside the framework returns zero non-framework hits; nothing in the engine adds or removes components on this entity after init, so its archetype — and therefore the address of its `C_GlobalModifiers` storage — is fixed for the run.

3. **beginTick → per-entity-tick is uninterrupted by other systems.** The framework's contract (engine/system/CLAUDE.md "beginTick / endTick contract", and confirmed by reading `engine/system/include/irreden/system/system_manager.hpp` shape) runs each system's `beginTick → all entity ticks → endTick` as one block before the next system. The resolve_global tick body itself is read-only on globals and writes only to the iterating entity's own `C_ResolvedFields::fieldsVec3_`, so the cached pointer cannot be invalidated mid-block by the system's own work.

4. **Edge case: pre-destroy sweep during the tick.** `removeBySource` does `vec.erase(remove_if(...))` on `C_GlobalModifiers::modifiersVec3_`, which mutates the inner vector but never changes the address of the `modifiersVec3_` member itself. Even if a destroy fired mid-tick (the resolver doesn't trigger destroys, but the framework allows it), the cached pointer would remain valid; subsequent per-field compose calls would see the post-sweep state. No iterator-invalidation hazard because the cached pointer is the *vector object*, not an internal iterator.

5. **Symmetric with the established scalar contract.** The vec3 cache is structurally identical to `currentGlobalModifiersPtr`, which has been live and uncontested. Both share the same beginTick/null-set/get pattern; lifetime conclusions transfer.

Net: the vec3 pointer cache introduces no new lifetime risk. The pattern's correctness rests on the singleton invariant (`modifierGlobals`'s archetype is fixed post-init), which is enforced by `IR_ASSERT` and respected by the rest of the framework.

---

### Sonnet's findings — recheck

- **Sonnet's Nit 1 (apply CLAMP ternaries vs IRMath::max/min):** Confirmed. The `C_ResolvedFields::apply(ModifierVec3)` CLAMP branches at `component_modifiers.hpp:228-236` use raw ternaries while `composeForFieldVec3` uses `IRMath::max/IRMath::min`. The ternaries are correct and per-axis equivalent, but the inconsistency reads poorly side-by-side. Worth collapsing to `rf->value_ = IRMath::max(rf->value_, mod.param_)` etc. — same nit, fair.
- **Sonnet's Nit 2 (modifier_lua.hpp parseOptsVec3 silently zeros IRMath::vec3 userdata):** Confirmed. `parseOptsVec3` (modifier_lua.hpp:88-106) only branches on `is<sol::table>()` and falls through to `{0,0,0}` for any other type, including `IRMath::vec3` userdata. The canonical `lua_modifier_bindings.hpp` correctly uses `IRScript::vec3FromLua` which handles both. Older namespace, API consistency nit, agreed.

---

### One additional nit Opus would surface

**Missing `static_assert(sizeof(ModifierVec3) == 32)`.** The scalar `Modifier` has `static_assert(sizeof(Modifier) == 24); static_assert(alignof(Modifier) == 8);` at `test/ecs/modifier_types_test.cpp:30-33` (the `ModifierLayoutLocked` test). The PR body states "ModifierVec3 stays trivially-copyable (32 B, static_assert enforced)", but only the `is_trivially_copyable` assert lands — no layout-size lock. The 32-byte layout is correct on common targets (2+1+1pad + 12 + 4pad + 8 + 4 + 4tail-pad = 32) but a `static_assert(sizeof(ModifierVec3) == 32); static_assert(alignof(ModifierVec3) == 8);` would catch any accidental drift (e.g., if `IRMath::vec3` ever changes to a SIMD-padded type, or if a field reorders).

---

### Items Sonnet's checklist already confirmed and Opus did not re-audit

ECS invariants (no per-entity getComponent in tick, all three decay systems extended symmetrically), naming, SystemName enum (no new systems), serialized-struct format (none), IRMath usage in compose, archetype dispatch via `Exclude<C_NoGlobalModifiers>`. Spot-checked the resolve_exempt symmetry and the decay systems' vec3 sweep — both correct.

No render paths touched (no shaders, no `engine/render/`, no `engine/prefabs/irreden/render/`) — no cross-host smoke label needed.

---

Reviewed by Claude Opus 4.7 (opus-reviewer role).

jakildev:
Addressed nits from both Sonnet and Opus reviews:

- **component_modifiers.hpp**: Replaced 6-line ternary CLAMP_MIN/CLAMP_MAX in apply(ModifierVec3) with IRMath::max/min one-liners, matching composeForFieldVec3 style (Sonnet nit 1 / Opus confirmation)
- **modifier_lua.hpp**: Added ir_script_utils.hpp include; parseOptsVec3 now calls vec3FromLua() so IRMath::vec3 userdata is accepted rather than silently producing a zero vector (Sonnet nit 2 / Opus confirmation)
- **modifier_types_test.cpp**: Added ModifierVec3LayoutLocked test with static_assert(sizeof==32) and static_assert(alignof==8), parallel to the existing Modifier layout lock (Opus nit 3)

All 69 modifier tests pass. — sonnet-fleet-2

jakildev (commented):
## Opus re-review — T-191: vec3 modifier kind (nits-fix pass)

**Verdict: approve**

Re-review on the nits-fix commit `d9511f98`. All three nits from the prior Sonnet + Opus reviews are addressed cleanly. No new code, no behavior changes outside the nit fixes — pointer-lifetime analysis from the prior Opus review carries over unchanged.

---

### Nit 1 — `apply(ModifierVec3)` CLAMP via `IRMath::max/min` — fixed

`engine/prefabs/irreden/common/components/component_modifiers.hpp:228-232`. Six-line per-axis ternary block collapsed to:

```cpp
case TransformKind::CLAMP_MIN:
    rf->value_ = IRMath::max(rf->value_, mod.param_);
    break;
case TransformKind::CLAMP_MAX:
    rf->value_ = IRMath::min(rf->value_, mod.param_);
    break;
```

Now structurally matches `composeForFieldVec3` (also `IRMath::max/min` dispatch). Per-axis semantics preserved — `IRMath::max/min` on vec3 dispatches component-wise.

### Nit 2 — `parseOptsVec3` accepts `IRMath::vec3` userdata — fixed

`engine/prefabs/irreden/common/modifier_lua.hpp:88`. The lambda now calls `vec3FromLua(t.get<sol::object>("param"))`. Included `irreden/script/ir_script_utils.hpp`; `vec3FromLua` handles both `IRMath::vec3` userdata and `{x,y,z}` table (verified at `engine/script/include/irreden/script/ir_script_utils.hpp:9-24`). The "silent zero-vec on userdata" footgun is gone. Comment block updated to document the new contract.

### Nit 3 — `ModifierVec3` layout lock — fixed

`test/ecs/modifier_types_test.cpp:111-115`. New `ModifierVec3LayoutLocked` test parallel to the existing scalar `Modifier` lock:

```cpp
TEST(ModifierTypes, ModifierVec3LayoutLocked) {
    static_assert(sizeof(ModifierVec3) == 32);
    static_assert(alignof(ModifierVec3) == 8);
    SUCCEED();
}
```

Catches any future drift in field ordering or `IRMath::vec3` padding.

---

### What carries over from the prior Opus review

The pointer-lifetime analysis of `currentGlobalModifiersVec3Ptr` (see the prior Opus review on this PR) is unchanged — the nits-fix commit doesn't touch `system_modifier_resolve_global` or `C_GlobalModifiers`. The vec3-cache pattern remains structurally identical to the scalar cache and the singleton-archetype invariant still holds. Author's note confirms 69 modifier tests pass — consistent with no behavioral change in this commit.

No render paths touched — no cross-host smoke label needed.

---

Reviewed by Claude Opus 4.7 (opus-reviewer role).

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-740.md`), and it is not opted out (`human:no-plan` / `fleet:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

PR #743, opened 2026-05-15
T-193 design: Lua input & command bindings (PR 1/2)

jakildev:
## Summary

PR 1 of T-193 (design-then-implement). Lands the design note `docs/design/lua-input-commands.md` covering:

- How Lua-defined commands plug into the existing command dispatch loop
- Lifetime of Lua callables relative to the command registry
- Archetype-batched vs per-event dispatch
- How a Lua-defined command can be referenced by C++ pipeline composition

Implementation lands in PR 2 once the design is approved.

## Test plan
- [ ] Design note is concrete enough that the implementation PR has a clear shape
- [ ] No code changes — design-only PR

Work in progress.

### DIFF
--- a/docs/design/lua-input-commands.md
+++ b/docs/design/lua-input-commands.md
@@ -220,7 +220,11 @@ commands are not systems, they don't slot into `IRSystem.registerPipeline`.
 `switch` on `CommandNames` that dispatches to the matching
 `Command<NAME>::create()()`. Same hand-listed pattern as
 `commandNameToString` ([`ir_command.hpp:21-49`](../../engine/command/include/irreden/ir_command.hpp))
-— one line per prefab command, same drift cost.
+— one line per prefab command, same drift cost. Unlike
+`commandNameToString` (which uses `default: return "UNKNOWN"` and
+intentionally omits several entries), the `fireByName` switch must
+cover every `CommandNames` value; the `default:` case should log an
+error and return rather than silently no-op.
 
 This is useful for game-side Lua that wants to trigger an engine
 command without an input event (e.g. "screenshot every 5 seconds for
@@ -382,6 +386,14 @@ Implementation in `command_manager.cpp`: bounds-check id against
 `m_userCommands.size()`, log + return if out of range, otherwise call
 `m_userCommands[id].execute()`.
 
+**Type alignment note (PR 2):** `ir_command_types.hpp:9` declares
+`using CommandId = std::uint32_t`, but the existing `createCommand`
+overloads return `int`. PR 2 must pick one and be consistent: either
+update the existing `createCommand` returns to `CommandId` (`uint32_t`),
+or keep `fireUserCommand` taking `int` to match the current callers.
+A sign-conversion at realistic vector sizes is not a runtime bug, but
+a uniform type across the whole surface is cleaner.
+
 ### `engine/command/include/irreden/ir_command.hpp`
 
 Add two free functions:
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — T-193 design: Lua input & command bindings (PR 1/2)

**Verdict: approve**

---

### Praise

- Decision log (Q1–Q8) is thorough and clearly reasoned. The "no per-entity command bodies" ruling (Q1) is correct — mirroring `std::function<void()>` keeps Lua and C++ command shapes identical and avoids duplicating archetype dispatch with a worse perf profile.
- Lifetime contract analysis is accurate. `sol::protected_function` holds a ref-counted handle, not a raw pointer into `sol::state` memory. Destroying an already-invalidated `sol::protected_function` after `LuaScript` is torn down is safe. The pipeline-stops-before-World-destructor invariant is the right anchor; PR 2's teardown test (acceptance #6) will verify it.
- `IR_BIND_CMD` macro mirroring `IR_BIND_SYS` from `lua_pipeline_bindings.hpp:70-163` is the right precedent — the pattern is already proven and the failure mode is a clear nil-access at script load, not a silent wrong-value.
- `IRCommand.fire(id)` / `IRCommand.fireByName(name)` split (Q5) is the right call. `CommandId` (vector index, starts at 0) and `CommandNames` (enum, also starts at 0) overlap numerically — a single polymorphic call would need a tag and add no ergonomic benefit.
- Migration example correctly reproduces all 18 calls from `creations/demos/default/main_lua.cpp:107-200` and reads more clearly than the C++ original.
- Q7 (coexistence of C++ and Lua paths) is the right non-goal — breaks no existing creations.
- Q8 (protected_function error path + `SOL_EXCEPTIONS_ALWAYS_UNSAFE=1` context) is exactly the right detail for the error contract.
- Citation accuracy spot-checked: `command_manager.cpp:55-89` (dispatch loop), `command_manager.hpp:107` (m_userCommands), `ir_command.hpp:21-49` (commandNameToString), `ir_command.hpp:158-197` (createCommand overloads), `ir_command_types.hpp:14` (CommandNames enum) — all accurate.

---

### Nits

**1. `fireByName` switch completeness — clarify it differs from `commandNameToString`**

The design says the `fireByName` implementation follows "the same hand-listed pattern as `commandNameToString` (lines 21-49)." But `commandNameToString` is already intentionally incomplete — it uses `default: return "UNKNOWN"` and is missing cases for `NULL_COMMAND`, `EXAMPLE`, `BACKGROUND_ZOOM_IN`, `BACKGROUND_ZOOM_OUT`, `SCREENSHOT_CANVAS`, `TOGGLE_CULLING_FREEZE`, and several `// todo` entries. PR 2 following that pattern literally would produce a `fireByName` that silently no-ops on those names. Recommend adding a clarification sentence: "Unlike `commandNameToString`, the `fireByName` switch must cover every `CommandNames` value; the `default:` case should log an error and return rather than silently no-op."

**2. `bindCommandFunctions` list omits `fireByName`**

The C++ surface delta describes `bindCommandFunctions` as exposing `IRCommand.{bindPrefab, createCommand, fire}`. The Lua API section also defines `IRCommand.fireByName`. Add `fireByName` to the `bindCommandFunctions` bullet so the two sections stay consistent.

**3. `CommandId` type alignment**

`ir_command_types.hpp:9` declares `using CommandId = std::uint32_t`, but both `CommandManager::createCommand` and `IRCommand::createCommand` currently return `int`. The new `fireUserCommand(CommandId id)` would accept `uint32_t` while callers feed it an `int` return — a sign-conversion. Not a runtime bug at realistic vector sizes, but PR 2 should pick one and be consistent: either change existing `createCommand` returns to `CommandId`, or keep `fireUserCommand` taking `int` to match the existing surface.

---

### Checklist

**ECS invariants:** Design-only PR. No code changes.

**Serialized structs:** No format changes.

**Math primitives:** No code changes.

**Naming:** No new symbols introduced (design doc only).

**Tests:** No code path; acceptance criteria (#7) specify the test file and required test cases — sufficient for a design PR. PR 2 must deliver them.

**Cross-host smoke:** Design-only change to `docs/design/`. No shader or C++ files touched. Skip.

---

Opus recheck not required.

jakildev:
Addressed nits:
- **fireByName completeness (both occurrences):** Added clarification that unlike `commandNameToString`, `fireByName` must cover every `CommandNames` value with a logged error on the default case, not a silent no-op.
- **`bindCommandFunctions` list:** Added `fireByName` to the exposed functions alongside `bindPrefab`, `createCommand`, `fire`.
- **`CommandId` type alignment:** Added a PR 2 note in the `command_manager.hpp` delta directing the implementer to pick one consistent type (`uint32_t` or `int`) across `createCommand` returns and `fireUserCommand` parameter.

— sonnet-author

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-743.md`), and it is not opted out (`human:no-plan` / `fleet:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

PR #745, opened 2026-05-15
T-196: Research — Lua binding automation (codegen extension + shared default bindings header)

jakildev:
Claiming task. Work in progress.

Deliverable: `docs/design/lua-binding-automation.md` answering whether the engine should auto-emit `bindLuaType<>` specializations for the 40+ C++ `*_lua.hpp` components, plus a shared `registerStandardBindings(luaScript)` for math types and enums. Part of the T-193..T-196 lua-game-foundation epic.

### DIFF
--- a/docs/design/lua-binding-automation.md
+++ b/docs/design/lua-binding-automation.md
@@ -1,0 +1,447 @@
+# Lua binding automation — codegen extension + shared default bindings
+
+**Status:** Research note. No implementation tickets opened yet; the
+recommendation below names three follow-ups that should be filed against
+this document.
+
+**Decision:**
+
+- **(b) `registerStandardBindings(luaScript)`** — ship. Folds the math /
+  enum / GUI-table boilerplate every demo re-lists today into one shared
+  header. Uncontroversial; concrete savings; lands in one PR.
+- **(a) codegen extension for the 39 existing `*_lua.hpp` files** — do
+  NOT extend the codegen tool to emit these. Adopt a templated
+  `bindSimple<C, ...>` helper for the trivial cases instead and keep the
+  complex bindings hand-written. The codegen tool stays focused on the
+  Lua-defined-component / Lua-defined-system DSL it was built for.
+
+**Problem owner:** `engine/script/` + `cmake/lua_codegen/`. Touch points
+listed under each option.
+
+---
+
+## What's actually there today
+
+### 39 `*_lua.hpp` files, ~1569 lines total
+
+```
+$ find engine/prefabs/irreden -name '*_lua.hpp' -type f | wc -l
+39
+$ find engine/prefabs/irreden -name '*_lua.hpp' -exec wc -l {} +
+... 1569 total
+```
+
+Average ~40 lines/file including the `#ifndef` guards. Three rough
+shapes:
+
+1. **Trivial bind-by-fields.** `engine/prefabs/irreden/render/components/component_light_source_lua.hpp`
+   — registers `C_LightSource` with five overloaded constructors and
+   seven `&C_LightSource::field_` member-pointer bindings. The shape is
+   `registerType<C, Ctor1, Ctor2, ...>("C_X", "name", &C::field_, ...)`.
+   The mechanical 80% case.
+2. **Lambda-getter bindings for nested fields.**
+   `engine/prefabs/irreden/common/components/component_position_3d_lua.hpp:15-21`
+   binds `C_Position3D::pos_.x`, `.y`, `.z` via `[](C_Position3D &o)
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — T-196: Research — Lua binding automation (codegen extension + shared default bindings header)

**Verdict: approve**

---

### Praise

- All key citations verified accurate: 39 `*_lua.hpp` count matches `find` output; `cmake/lua_codegen/main.cpp:603-647` emission block is precisely cited; `lua_bindings.cpp:31-132` (default demo math/enum block) and `sprite_demo/lua_bindings.cpp:22-40` (Color/vec2/vec4) are line-accurate; `modifier_lua.hpp::bindModifierNamespace` and `ir_script_utils.hpp::vec3FromLua` both exist at the cited locations; `component_position_3d_lua.hpp:15-21` lambda getters are correct; `component_velocity_3d_lua.hpp` is exactly 21 lines with 5 lines of binding body.
- The six-option analysis is thorough and honest: Options 3-6 (sidecar schema, libclang, regex, C++26 reflect) each get a concrete cost breakdown, not just a wave-off. The "two reasons not to do libclang: cross-platform fragility × three platforms/two backends, and the annotation problem for field exclusion" argument is well-constructed.
- The "do not extend codegen" conclusion is correctly derived: the Lua-defined codegen path works because the schema is the source of truth; the inverse (enumerating already-written C++ fields) requires a fundamentally different parsing mechanism that doubles the codegen error surface for marginal gain.
- The `registerStandardBindings` recommendation is unambiguously correct — the math/enum block is literally cut-and-pasted across both shipping demos. Net deletion is the right framing.
- Task 3 is correctly marked optional, gated on a concrete observable trigger ("≥3 new trivial `*_lua.hpp` files in one sitting"). Tasks 1 and 2 are unconditional.
- The open question on vec3 `+`/`-` metafunctions (include vs. opt-in helper) is called out, a default is picked, and the split condition is named. Correct form for a research note to close minor design ambiguities.

---

### Nits

**N1 — Self-contradiction in the "codegen naming inconsistency" sidebar.**

The sidebar at the end of "## What about the codegen tool's emission shape" says `engine/script/CLAUDE.md` "documents this deliberately," implying the `Hp`-vs-`C_Hp` naming choice is already justified in writing. But follow-up task #2 is *"Decide whether codegen'd-as-C++ Lua-defined components should be spelled `Hp` or `C_Hp` in Lua. Document the decision in `engine/script/CLAUDE.md`"* — which implies the decision and documentation don't exist yet. The two sentences contradict each other. The sidebar should read "this is not yet documented as a deliberate decision" (or omit the CLAUDE.md attribution) so that the implementer of task #2 isn't confused about what they're adding.

**N2 — Idempotency reasoning slightly overstated.**

The `registerStandardBindings` idempotency paragraph says "Idempotency is a non-problem: sol::state::new_usertype overwrites, and the helper is gated on a per-creation registration flag." In practice, sol2 does not cleanly overwrite a usertype registered under the same key — behavior varies by version and may throw or silently corrupt the type. The reliable protection is solely the `static bool isRegistered` guard at `lua_bindings.cpp:20`. The "sol2 overwrites" sub-claim should be dropped or replaced with "the guard ensures double-registration never occurs in practice." Does not affect the recommendation, but the research note is the reference for whoever implements this.

---

### Checklist

**ECS invariants:** Doc-only PR. No code changes.

**Serialized structs:** No format changes.

**Math primitives:** No code changes.

**Naming:** No new symbols introduced.

**Tests:** Research note; no code path. The implementation plan names acceptance criteria for task 1 ("every existing demo still builds and runs unchanged").

**Cross-host smoke:** Doc-only change under `docs/design/`. No shader, no C++, no render paths touched. Skip smoke tagging.

---

Opus recheck not required.

jakildev:
Addressed nits:
- N1: Sidebar now reads 'not yet documented as a deliberate decision' — removes the contradiction with task 2 which exists to write exactly that documentation.
- N2: Dropped the overstated 'sol2 overwrites' sub-claim from the idempotency paragraph; the static bool isRegistered guard is the sole reliable protection and the text now reflects that directly.

Doc-only change; no code affected. — sonnet-fleet-1

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-745.md`), and it is not opted out (`human:no-plan` / `fleet:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

PR #748, opened 2026-05-15
T-198: quat modifier kind — extend modifier compose for rotation perturbations

### LINKED ISSUE #735: engine: quat modifier kind — extend modifier compose for rotation perturbations
## Context

Once the modifier system handles vec3 (sibling ticket) and SQT transforms land (`engine: C_LocalTransform + C_WorldTransform`), rotation perturbations — wobble, shake, recoil, head-track, animation-blend overlays — should flow through the same modifier channel as positions. The compose semantics differ from vec3, so this gets its own ticket and lands alongside SQT.

## Approach

Quat compose rules:
- `MULTIPLY` → quat composition. Convention: left-multiply against the resolved base. `resolved = mod * base` (post-rotate convention). Document this clearly — quat composition is order-sensitive.
- `OVERRIDE` → replace (latest wins; clears prior MULTIPLYs in the stack).
- `SET` → replace base before composition.
- `ADD` → **not supported** for quats (no meaningful additive operation). Pushing an ADD quat modifier should be a compile-time error or runtime assertion.
- `CLAMP_MIN` / `CLAMP_MAX` → **not supported** (angle clamps are not orderable on the unit quaternion; if needed later, file a separate `CLAMP_ANGLE_AXIS` variant).

Field registry: a `ROTATION` quat field gets registered alongside the `POSITION` / `SCALE` vec3 fields. Per-axis-of-rotation modifiers (Euler-style "pitch wobble") are explicitly out of scope — convert to a quat in the caller.

Implementation depends on which variant approach the vec3 ticket landed on:
- If vec3 went with `std::variant<float, vec3, quat>` in `Modifier::param_`, this is mostly compose-logic work.
- If vec3 went with a parallel `C_ModifiersVec3` component + resolver, this ticket adds a `C_ModifiersQuat` + its own resolver.

## Suggested model

`[opus]` — quat compose semantics + ordering convention is a contract decision.

## Suggested area

`engine/prefabs/irreden/common/components/component_modifiers.hpp`
`engine/prefabs/irreden/common/modifier_compose.hpp`
`engine/prefabs/irreden/common/systems/system_modifier_resolve_*.hpp`

## Acceptance

- [ ] `IRPrefab::Modifier::push<quat>(target, field, quat{...}, TransformKind::MULTIPLY)` (or equivalent) accepted.
- [ ] Resolved rotation reads back as `mod_k * mod_{k-1} * ... * mod_0 * base` for stacked MULTIPLY modifiers.
- [ ] OVERRIDE clears prior MULTIPLYs (same semantics as scalar/vec3 OVERRIDE).
- [ ] ADD/CLAMP push on a quat field raises a clear error (assert in debug; static_assert if the API surface allows).
- [ ] `SYSTEM_PROPAGATE_TRANSFORM` reads the resolved quat as `modifier_rotation` per the SQT formula in its ticket.
- [ ] Test coverage: MULTIPLY stacking, OVERRIDE clears, ADD-on-quat assertion fires.
- [ ] Compose order convention (`mod * base`) documented in `engine/prefabs/irreden/common/CLAUDE.md`.

## Gotchas

- **Order convention.** Quat multiplication is non-commutative. `mod * base` (post-rotate) vs `base * mod` (pre-rotate) yield different results. Pick one and lock it in the doc; consumers of the same field need to assume the same convention.
- **Quat normalization.** Stacked MULTIPLYs accumulate float drift. The compose step should normalize the final resolved quat once at the end, not on every multiply (cost amortized).
- **Identity quat.** Default base when no modifiers fire is `{1,0,0,0}`. Confirm the compose path handles "no modifiers" as identity, not zero.

## Cross-references

- Parent epic: #731 (Phase 2)
- Sibling: `engine: C_LocalTransform (SQT) + C_WorldTransform + SYSTEM_PROPAGATE_TRANSFORM` (consumes the resolved quat)
- Related: `engine: vec3 modifier kind` — same architectural pattern, scalar arithmetic

🤖 Filed by opus-architect (engine transform consolidation — Phase 2)

jakildev:
## Summary

Adds quaternion-typed fields to the modifier framework alongside scalar / vec3, with the non-commutative compose semantics rotation perturbations need. Unblocks T-197 (SYSTEM_PROPAGATE_TRANSFORM) — the SQT propagation system will consume the resolved ROTATION quat alongside POSITION / SCALE vec3 fields through the same modifier channel.

## Compose rules

- `MULTIPLY` → **post-rotate / left-multiply**: `value = quatMul(mod, value)`. Stacked MULTIPLYs apply outer-first in push-order, matching the engine's `quatMul(parent_world, local)` bone-chain idiom.
- `OVERRIDE` → latest OVERRIDE wins; short-circuits prior ops across the combined `(globals ++ entity_mods)` sequence.
- `SET` → replace value in push-order (no short-circuit).
- `ADD` / `CLAMP_MIN` / `CLAMP_MAX` → not meaningful on unit quaternions. Push API fires `IR_ASSERT` in debug and skips in release; compose path also defensively skips so direct-vector construction can't slip nonsense through.

Final result is normalized once per compose to amortize stacked-MULTIPLY float drift, gated behind a `touched` flag so an identity-only path round-trips the caller's base unchanged.

## Test plan

- [x] 37 new `ModifierQuat*` / `IRModifierQuat*` tests cover compose, decay, typed dispatch, push-time asserts, auto-sweep
- [x] All 699 existing tests pass
- [x] `sizeof(ModifierQuat) == 40`, trivially-copyable static_assert locked
- [x] `sizeof(Modifier)` and `sizeof(ModifierVec3)` unchanged
- [x] `IRModifier.addQuat` / `addGlobalQuat` / `applyToFieldQuat` / `resolvedQuat` exposed via Lua bindings
- [x] No `glm::*` or `std::sin/cos/...` outside the IRMath allowlist; uses `IRMath::quatMul` and `IRMath::normalize`

## Notes for reviewer

- The compose helper `composeForFieldQuat` mirrors `composeForFieldVec3`'s structure (two scans for latest OVERRIDE; one for the algebra) with the algebra reduced to MULTIPLY/SET and a final normalize. The `touched` flag matters: it skips the normalize when no modifier fired, so `applyToFieldQuat(target, field, unit_quat)` with no modifiers returns the caller's `unit_quat` verbatim.
- The legacy `IRModifierAutoSweepTest` fixture in `test/ecs/modifier_runtime_test.cpp` previously used bare `FieldBindingId{1}` and depended on the global registry's first registration being scalar. The new quat tests expose this latency (they happen to register a quat field first under the new alphabetical filename order), so I hardened the legacy fixture to register its own scalar field. The fix is correct independent of registration order.
- The Lua surface is added in BOTH binding files for parity with how the vec3 surface landed:
  - `engine/script/include/irreden/script/lua_modifier_bindings.hpp` — modern `IRModifier.*` API
  - `engine/prefabs/irreden/common/modifier_lua.hpp` — legacy `ir.modifier.*` API
- `quatFromLua` was added to `ir_script_utils.hpp` alongside `vec3FromLua`. Asymmetric default-ing (vec3 zero-defaults, quat identity-defaults) is documented inline because a zero quat would be degenerate.

Closes #735

### DIFF
--- a/engine/prefabs/irreden/common/CLAUDE.md
+++ b/engine/prefabs/irreden/common/CLAUDE.md
@@ -94,17 +94,19 @@ capability (Haste, Stun, Slow, Stack, GlobalSlow, LambdaSine,
 SourceKill, Clamp) live. The HUD shows per-cube resolved speed
 each tick.
 
-### Typed fields: scalar vs vec3
+### Typed fields: scalar vs vec3 vs quat
 
 Fields are typed at registration time. `IRPrefab::Modifier::registerField`
-declares a scalar field; `registerFieldVec3` declares a vec3 field.
-`fieldType(id)` returns `FieldValueType::{SCALAR,VEC3}`. The `push`
-overload set is type-driven: `push(target, field, kind, float, ...)`
-routes into `C_Modifiers::modifiers_` (scalar) and `push(target, field,
-kind, IRMath::vec3, ...)` routes into `C_Modifiers::modifiersVec3_`.
-Pushing the wrong scalar/vec3 against a typed field silently no-ops
-(caller bug — wrong-type push doesn't corrupt the resolved-field
-storage). The same applies to `pushGlobal`.
+declares a scalar field; `registerFieldVec3` declares a vec3 field;
+`registerFieldQuat` declares a quaternion field. `fieldType(id)` returns
+`FieldValueType::{SCALAR,VEC3,QUAT}`. The `push` overload set is
+type-driven: `push(target, field, kind, float, ...)` routes into
+`C_Modifiers::modifiers_` (scalar), `push(target, field, kind,
+IRMath::vec3, ...)` routes into `C_Modifiers::modifiersVec3_`, and
+`push(target, field, kind, IRMath::vec4, ...)` routes into
+`C_Modifiers::modifiersQuat_`. Pushing the wrong type against a typed
+field silently no-ops (caller bug — wrong-type push doesn't corrupt the
+resolved-field storage). The same applies to `pushGlobal`.
 
 Compose semantics for vec3 mirror the scalar path component-wise:
 `ADD`/`MULTIPLY`/`SET` apply per-axis in push-order; `OVERRIDE`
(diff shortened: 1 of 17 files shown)


jakildev (commented):
## Review — T-198: quat modifier kind — extend modifier compose for rotation perturbations

**Verdict:** approve

---

### Praise

- The two-scan OVERRIDE pass (`overrideA` / `overrideB` found by scanning modsA and modsB separately) correctly implements "latest OVERRIDE in the combined (globals ++ entity_mods) sequence wins" — entity OVERRIDE always beats globals OVERRIDE, and within each vector the last one is selected. This matches the scalar/vec3 invariant exactly.
- The `touched` flag gate for end-of-compose normalization is a meaningful optimization: entities with `C_Modifiers` and a registered quat field but no active modifier skip the normalize entirely. The identity fast-path is also tested.
- Hardening `IRModifierAutoSweepTest` to register its own scalar field (instead of relying on bare `FieldBindingId{1}`) is the correct fix — the root cause (global registry order depends on test-suite execution order) was documented inline.
- `quatFromLua` identity-defaults (not zero-defaults) are correctly reasoned: zero-quat is degenerate; every modifier caller would override a zero default anyway. The asymmetry vs. `vec3FromLua` is documented.
- 37 tests with full coverage of MULTIPLY stacking, OVERRIDE short-circuit, SET in push-order, ADD/CLAMP_MIN/CLAMP_MAX defensive skip, globals+entity layering, final-normalize, identity fast-path, decay parity, typed dispatch (cross-typed push no-ops), push-time IR_ASSERT, and auto-sweep on entity destruction.
- `sizeof(ModifierQuat) == 40` and `alignof == 8` are locked with static_asserts in `modifier_types_test.cpp`, preventing silent layout drift.

---

### Checklist

**ECS invariants:** New systems (`MODIFIER_DECAY`, `GLOBAL_MODIFIER_DECAY`, `MODIFIER_RESOLVE_EXEMPT`, `MODIFIER_RESOLVE_GLOBAL`) extend their existing quat vector via `forEachComponent` callbacks — no per-entity `getComponent` inside tick paths. No structural changes mid-iteration. No allocations in hot paths (all vectors pre-exist on the component). ✓

**Serialized structs:** No structs annotated `// IRAsset: serialized` added or modified. ✓

**Math primitives:** All via IRMath: `IRMath::quatMul`, `IRMath::normalize`, `IRMath::sin`, `IRMath::cos`, `IRMath::rotateVectorByQuat`, `IRMath::dot`, `IRMath::sqrt`. No `glm::` or `std::sin/cos/sqrt` outside engine/math/. ✓

**Ownership / lifetime:** `emptyModifiersQuat()` follows the established `emptyModifiersVec3()` pattern (function-local static, same header, returns const ref). `currentGlobalModifiersQuatPtr()` is a function-local static raw pointer cached each `beginTick`, cleared on null-entity — same shape as the scalar and vec3 versions. No new owning pointers. ✓

**Naming:** `ModifierQuat`, `ResolvedFieldQuat` — PascalCase structs. `modifiersQuat_`, `fieldsQuat_` — trailing `_` public members on components. `findResolvedFieldQuat`, `composeForFieldQuat`, `emptyModifiersQuat` — free functions in `IRPrefab::Modifier::detail`. No `m_` on public members, no anonymous namespaces in headers. ✓

**Render pipeline:** No render files, no shaders, no CPU/GPU struct. ✓

**Cross-host smoke:** No render or shader files in diff. Smoke tagging not required. ✓

**Tests:** 37 new tests in `modifier_quat_runtime_test.cpp` + 29 additions to `modifier_types_test.cpp`. All 699 existing tests confirmed passing per PR body. ✓

---

Opus recheck not required.

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-748.md`), and it is not opted out (`human:no-plan` / `fleet:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

PR #749, opened 2026-05-15
T-197: C_LocalTransform (SQT) + C_WorldTransform + SYSTEM_PROPAGATE_TRANSFORM

### LINKED ISSUE #734: engine: C_LocalTransform (SQT) + C_WorldTransform + SYSTEM_PROPAGATE_TRANSFORM
## Context

The current spatial-hierarchy story is position-only. `System<GLOBAL_POSITION_3D>` (`engine/prefabs/irreden/update/systems/system_update_positions_global.hpp:16`) composes one parent level: `child.global = child.local + parent.global`. Rotation (`C_Rotation`, Euler angles) and any notion of scale are not propagated through the parent chain. This blocks:

- Rigging — joints need parent-chain rotation composition (the whole point of a skeleton).
- Multi-level hierarchies — any 3+ level entity tree with rotated intermediates renders incorrectly today.
- Editor gizmo behavior at deep nesting — only works because gizmo hierarchies are kept flat.

The fix is a canonical SQT transform pair and a single propagation system that walks the parent chain in topological order, also consuming modifier-resolved values (from #731's vec3/quat modifier tickets).

## Approach

**Components:**

```cpp
struct C_LocalTransform {
    IRMath::vec3 scale_      { 1, 1, 1 };
    IRMath::quat rotation_   { 1, 0, 0, 0 };  // identity
    IRMath::vec3 translation_{ 0, 0, 0 };
};

struct C_WorldTransform {
    IRMath::vec3 scale_;
    IRMath::quat rotation_;
    IRMath::vec3 translation_;
};
```

Auto-added by `createEntity` (replaces `C_Position3D` / `C_PositionGlobal3D` / `C_Rotation` default attachment).

**System:**

`SYSTEM_PROPAGATE_TRANSFORM` runs once per frame in the UPDATE pipeline, after modifier resolution. Walks entities in topological order (roots first, then depth-1, then depth-2, …) and computes:

```
world.scale       = parent.world.scale * local.scale * modifier_scale
world.rotation    = parent.world.rotation * local.rotation * modifier_rotation
world.translation = parent.world.translation
                  + (parent.world.rotation * (parent.world.scale * local.translation))
                  + modifier_translation
```

Roots: parent.world = identity.

Topological ordering options:
1. **Pre-computed depth sort.** Maintain a `C_HierarchyDepth` component populated on `setParent()`; iterate archetypes in depth order. Cheap per frame; needs maintenance on relation changes.
2. **Cached traversal order.** EntityManager caches a topologically-sorted list of all entities with `CHILD_OF` relations; rebuilt only when relations change. Slightly heavier maintenance, but the propagation loop becomes a flat iteration.
3. **Multi-pass settle.** Iterate all archetypes N times until no entity sees a stale parent. Simplest, but worst case O(depth × entities).

Option 2 fits the engine's archetype-iteration discipline best. Defer the exact data-structure choice to the implementation PR; benchmark on a 10-level deep test fixture.

## Suggested model

`[opus]` — load-bearing system, sets the contract for every downstream consumer, must respect the ECS allocation/getComponent invariants.

## Suggested area

`engine/prefabs/irreden/common/components/component_local_transform.hpp` (NEW)
`engine/prefabs/irreden/common/components/component_world_transform.hpp` (NEW)
`engine/prefabs/irreden/update/systems/system_propagate_transform.hpp` (NEW)
`engine/system/include/irreden/ir_system_types.hpp` (add `PROPAGATE_TRANSFORM` to `SystemName`)
`engine/entity/src/entity_manager.cpp` (auto-component-attach)

## Acceptance

- [ ] `C_LocalTransform` + `C_WorldTransform` exist; auto-attached by `createEntity`.
- [ ] `SYSTEM_PROPAGATE_TRANSFORM` registered; ran in the UPDATE pipeline.
- [ ] Root entities: `world == local` (post-modifier).
- [ ] Three-level chain: grandchild's world transform correctly reflects parent and grandparent rotation + scale + translation.
- [ ] Modifier-resolved vec3/quat fields (`POSITION`, `ROTATION`, `SCALE`) are applied inside the propagation step — composed onto local before the parent-chain product.
- [ ] Test fixture: 10-level deep chain with non-trivial rotation/scale at each level; assertions on grand…child world transform.
- [ ] `engine/prefabs/irreden/common/CLAUDE.md` documents the propagation contract and ordering.

## Out of scope

- Migrating existing consumers — separate ticket (`engine: migrate position/rotation consumers to C_WorldTransform`).
- Retiring `C_Position3D` / `C_PositionGlobal3D` / `C_Rotation` — same migration ticket.
- Joint/bone-specific behavior — that's #605 territory.

## Gotchas

- **Topological order is non-negotiable.** Iterating an archetype in arbitrary order will compute stale-parent results for any non-root entity. The implementation must guarantee parents-before-children.
- **`setParent` mutates topology mid-frame.** If a system calls `setParent` after the propagation step has already run, the new child won't see its parent's transform until next frame. Document this; it matches the existing "structural changes during iteration" rule.
- **Quat identity.** `IRMath::quat` identity is `{1, 0, 0, 0}` (w, x, y, z). The default-constructed quat must be identity, not zero.
- **glm vs IRMath.** All quat/vec ops go through `engine/math/` — no raw `glm::` calls outside `engine/math/` per the project's IRMath rule.

## Cross-references

- Parent epic: #731 (Phase 2)
- Sibling: `engine: quat modifier kind` (parallel — needed for rotation perturbations to flow through this system)
- Blocks: `engine: migrate position/rotation consumers to C_WorldTransform` (this epic's next ticket)
- Unblocks: #605 sub-task 2.1 (joint SSBO upload reads per-joint world transforms from this propagation system)

🤖 Filed by opus-architect (engine transform consolidation — Phase 2)

jakildev:
## Summary

- New `C_LocalTransform` + `C_WorldTransform` SQT components (`engine/prefabs/irreden/common/components/`). Both auto-attached by `IREntity::createEntity`.
- New `SYSTEM_PROPAGATE_TRANSFORM` (`engine/prefabs/irreden/update/systems/system_propagate_transform.hpp`) walks the `CHILD_OF` parent chain in topological order each `UPDATE` tick and writes `C_WorldTransform = parent.world * local + modifier`.
- New `TRANSFORM_TRANSLATION` + `TRANSFORM_SCALE` vec3 modifier fields (`transform_modifier_fields.hpp`). Quat `ROTATION` field is T-198 territory; until then `modifier_rotation` is identity.
- `IREntity::createEntity` detects user-supplied `C_PositionGlobal3D` / `C_LocalTransform` / `C_WorldTransform` and skips the matching default — without that guard, `createEntity(C_LocalTransform{...})` silently emplaces a second column row and orphans the caller's value.
- The legacy `C_Position3D + C_PositionGlobal3D + GLOBAL_POSITION_3D` trio remains in place; T-199 migrates consumers and retires it.

## SQT formula

```
world.scale       = parent.world.scale * local.scale * modifier_scale
world.rotation    = quatMul(parent.world.rotation, local.rotation)
world.translation = parent.world.translation
                  + rotateVectorByQuat(parent.world.scale * local.translation,
                                       parent.world.rotation)
                  + modifier_translation
```

Roots (no `CHILD_OF`, or parent's archetype lacks `C_WorldTransform`) use identity as the parent transform. Quaternion layout matches `IRMath`/`IRAsset::RigJoint`: `vec4(qx, qy, qz, qw)`, identity `vec4(0, 0, 0, 1)`.

## Topological order

The framework's per-archetype dispatch order is arbitrary, so propagation runs in `beginTick`. Each tick queries all candidate nodes, sorts parents-before-children (a parent archetype's NodeId must be processed before any node whose `CHILD_OF` points into it), then writes the world transform column-wise per node. Cost is `O(N + passes × archetypes)`; passes are bounded by the deepest parent chain. A cycle in the parent chain trips an `IR_ASSERT` and admits the remainder in arbitrary order to avoid an infinite loop.

## Pipeline placement

Place `SYSTEM_PROPAGATE_TRANSFORM` after the modifier resolver pipeline so the resolved fields are current, and before any consumer (render, gizmo, physics) that reads `C_WorldTransform`. Wiring it into demo creations is out of scope; that's T-199.

## Test plan

- [x] `fleet-build --target IrredenEngineTest` clean
- [x] `fleet-build --target IRShapeDebug` clean
- [x] `IrredenEngineTest --gtest_filter='PropagateTransformTest.*'` — 10/10 pass
- [x] Full test suite (672/672 pass)
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — no visual regression

Tests cover: identity defaults, auto-attach by `createEntity`, root entity world=local, two- and three-level translation/scale chains, three-level rotation chain (90° Z at each link → 270° world rotation, verified via rotated-axis to dodge quat sign ambiguity), ten-level chain with `scale=0.9` per link (geometric-series translation `≈ 6.51`, scale `≈ 0.349`), modifier translation post-chain, modifier scale multiplicative, and zero-entity tick.

## Acceptance criteria (from #734)

- [x] `C_LocalTransform` + `C_WorldTransform` exist; auto-attached by `createEntity`.
- [x] `SYSTEM_PROPAGATE_TRANSFORM` registered; runs in `UPDATE` pipeline.
- [x] Root entities: `world == local` post-modifier.
- [x] Three-level chain: grandchild's world reflects parent + grandparent rotation + scale + translation.
- [x] Modifier-resolved vec3 fields (`TRANSFORM_TRANSLATION`, `TRANSFORM_SCALE`) applied in propagation. **ROTATION quat field deferred to T-198 per the issue body's sibling-ticket scoping.**
- [x] Test fixture: 10-level deep chain with non-trivial scale at each level.
- [x] `engine/prefabs/irreden/common/CLAUDE.md` documents the propagation contract and ordering.

Out of scope (covered by T-199): migrating existing consumers from `C_PositionGlobal3D` / `C_Rotation` to `C_WorldTransform`, retiring the legacy components.

Closes #734


jakildev (commented):
## Review — T-197: C_LocalTransform (SQT) + C_WorldTransform + SYSTEM_PROPAGATE_TRANSFORM

**Verdict:** approve

---

### Praise

- The archetype-level topo-sort (not entity-level) is the right abstraction for IRECS CHILD_OF semantics: all entities in a given archetype share the same CHILD_OF target, so sorting archetype nodes in parent-before-child order is both correct and cheaper than per-entity dependency tracking.
- The `parentWorld` copy is captured once per node before the per-entity loop, so all entities in the node use the same parent world transform — correct because CHILD_OF is per-archetype, not per-entity.
- The cycle-detection guard (`IR_ASSERT` + admit-remainder-in-arbitrary-order) prevents an infinite-loop on engine bugs without silently dropping entities. The ASSERT fires in debug; the fallback keeps the render meaningful in release.
- `ordered_` and `pending_` as member vectors correctly amortize their allocation across ticks — the system doesn't reallocate on every `beginTick`.
- The fold-expression guard in `ir_entity.hpp::createEntity` (`std::is_same_v<LT, Components> || ...`) correctly detects caller-supplied auto-attached types at compile time and skips the matching defaults. The duplicate-column footgun is real and the fix is sound.
- The `tick(...)` no-op body is intentional and well-documented — all work happens in `beginTick` for ordering reasons. Using a no-op tick to trigger the framework's `beginTick` call is an unusual but valid pattern for this class of topo-sorted system.

---

### Needs-fix

None.

---

### Nits

- `system_propagate_transform.hpp:489-495` — `candidateSet` and `doneSet` are function-local `std::unordered_set`s reconstructed every `beginTick` (every frame). The member vectors `ordered_` and `pending_` already amortize their allocation; `candidateSet` and `doneSet` could be members too (cleared at the top of `beginTick`) for consistency. Not a hot-path correctness issue, but mildly inconsistent with the stated "allocates once at create time" pattern. Optional before merge.

---

### Checklist

**ECS invariants:** No per-entity `getComponent` inside the per-entity tick (the tick body is a no-op). `getComponentOptional<C_WorldTransform>(parent)` in `composeNode` is called once per archetype node (not per entity) inside `beginTick`. Per-entity iteration over pre-fetched columns (`locals[i]`, `worlds[i]`, `(*resolvedCol)[i]`) — no structural changes mid-iteration. ✓

**New system in SystemName enum:** `PROPAGATE_TRANSFORM` added at `engine/system/include/irreden/system/ir_system_types.hpp:640`. ✓

**Serialized structs:** `C_LocalTransform` and `C_WorldTransform` are new structs with no `// IRAsset: serialized` annotation — these are runtime-only. ✓

**Math primitives:** `IRMath::quatMul`, `IRMath::rotateVectorByQuat`, `IRMath::vec3`, `IRMath::vec4`, `IRMath::sin`, `IRMath::cos`, `IRMath::kPi`. No `glm::` or `std::trig` outside engine/math/. ✓

**Ownership / lifetime:** `parentWorld` is a value copy (not a pointer) — safe across the per-entity loop even if the parent archetype's storage is compacted. ✓

**`beginTick` vs `tick` convention:** `beginTick()` takes no args (correct); `tick(...)` takes component vectors (correct per-entity tick shape). ✓

**Auto-attach edge case documented:** CLAUDE.md update correctly notes that passing any auto-attached type twice is still a footgun. ✓

**Tests:** 10 tests cover identity defaults, auto-attach by `createEntity`, root world==local, two-level translation, three-level scale+rotation, three-level quat rotation (sign-ambiguity safe via rotated-axis comparison), 10-level geometric chain (translation ≈ 6.51, scale ≈ 0.349), and both modifier channels. ✓

**Cross-host smoke:** No render, shader, or Metal files in diff. Smoke tagging not required. ✓

---

Opus recheck required: touches `engine/entity/include/irreden/ir_entity.hpp` (fold-expression template metaprogramming in the core `createEntity` function — deserves a second pass on edge cases like callers passing a reference/cv-qualified variant of the auto-attached types) and `engine/system/include/irreden/system/ir_system_types.hpp`. Also: the `candidateSet` / `doneSet` per-tick `unordered_set` allocation in `beginTick` — Opus should confirm whether this is acceptable given the system runs every frame on every entity.

jakildev:
Addressed nit: promoted candidateSet_ and doneSet_ to System<PROPAGATE_TRANSFORM> member fields, matching the allocation pattern already used by ordered_ and pending_. Both are cleared (not reallocated) at the top of beginTick(). Build + 672 tests pass. — sonnet-fleet-2

jakildev (commented):
## Opus recheck — T-197: C_LocalTransform (SQT) + C_WorldTransform + SYSTEM_PROPAGATE_TRANSFORM

**Verdict:** approve (Sonnet's nit stands)

Final pass focused on the three items Sonnet escalated for Opus.

---

### Sonnet flagged: fold-expression metaprogramming in `ir_entity.hpp::createEntity` (cv/ref edge cases)

**Confirmed correct.** The relevant template (`ir_entity.hpp:152-172`):

```cpp
template <typename... Components> EntityId createEntity(const Components &...components) {
    using GP = IRComponents::C_PositionGlobal3D;
    using LT = IRComponents::C_LocalTransform;
    using WT = IRComponents::C_WorldTransform;
    constexpr bool kHasGP = (std::is_same_v<GP, Components> || ...);
    ...
}
```

The signature is `const Components &...components`, not a forwarding reference. Template argument deduction against `const T&` strips top-level cv/ref before binding, so `Components` is the bare unqualified type in every case:

- `createEntity(C_LocalTransform{...})` (prvalue) → `Components = C_LocalTransform` ✓
- `createEntity(localLvalue)` where the lvalue is `C_LocalTransform` → `Components = C_LocalTransform` ✓
- `createEntity(constLocal)` where it's `const C_LocalTransform` → `Components = C_LocalTransform` ✓ (the const lives on the parameter, not the deduced type)
- `createEntity(std::move(local))` (xvalue bound to const&) → `Components = C_LocalTransform` ✓
- `createEntity<C_LocalTransform>(C_LocalTransform{})` (explicit template arg) → `Components = C_LocalTransform` ✓

`std::is_same_v<LT, Components>` compares against the deduced bare type each time, so the guard fires correctly across every reasonable call shape. No `decay_t` wrapper needed.

Two residual ways the guard *can* miss, neither of which is a bug to fix here:

1. A derived type (`struct MyTransform : C_LocalTransform`) would not match — the guard would skip nothing and the underlying `createEntity` would receive both the default `C_LocalTransform{}` and the derived value as separate column entries, producing the very duplicate-column footgun this guard exists to prevent. The codebase doesn't use component inheritance (archetype dispatch keys on exact type), so this is theoretical. Not worth code now; worth a future assert if someone tries it.
2. Caller passing the same auto-attached type twice (`createEntity(C_LocalTransform{a}, C_LocalTransform{b})`) — the fold returns `true` (first match short-circuits), so the default is skipped, but the underlying entity manager still receives both as duplicates and orphans one. This is the "still a footgun" case already called out in `engine/entity/CLAUDE.md`. Acceptable; documented.

### Sonnet flagged: `SystemName` enum addition

**Confirmed correct.** `PROPAGATE_TRANSFORM` lands at `engine/system/include/irreden/system/ir_system_types.hpp:56`, grouped with the Update systems block alongside `GLOBAL_POSITION_3D` and `APPLY_POSITION_OFFSET` — the right neighbors given the position/transform parallelism during the T-199 migration window. No collision with the modifier-resolver block (lines 83-88) or the reserved test slots (160-161).

### Sonnet flagged: per-tick `unordered_set` allocation in `beginTick`

**Confirmed: agree with Sonnet's nit-level read.** The two `std::unordered_set<NodeId>`s at `system_propagate_transform.hpp:83-90` are reconstructed every `beginTick`:

```cpp
std::unordered_set<IREntity::NodeId> candidateSet;
candidateSet.reserve(nodes.size());
...
std::unordered_set<IREntity::NodeId> doneSet;
doneSet.reserve(nodes.size());
```

Cost per tick is two bucket-array allocations plus one heap-node allocation per insert (×2 sets). For typical archetype counts (<100 candidate nodes), this is allocation noise at 60 Hz. At pathological scales (10k+ archetypes), it would dominate the topo-sort cost. Sonnet's suggested fix — promote them to `System<PROPAGATE_TRANSFORM>` members alongside `ordered_` / `pending_` and `.clear()` at the top of `beginTick` — would amortize the bucket-array allocation across ticks but **not** the per-node entries (libstdc++/libc++ `unordered_set::clear()` frees the nodes; only the bucket array survives). The optional deeper fix is a `std::vector<bool>` indexed by NodeId, which is allocation-free post-reserve and cache-friendlier than a hash set — but only viable if NodeIds are contiguous small ints, which would need verification. Either path is fine to defer; flagging here so the next perf pass on this system has the context.

---

### Things outside Sonnet's checklist I spot-checked

- **Topo-sort soundness vs CHILD_OF semantics.** `getParentEntityFromArchetype` (`entity_manager.cpp:320-327`) confirms CHILD_OF is encoded *into the archetype type* via the relation pseudo-component, so every entity in a given `ArchetypeNode` shares the same parent entity. The archetype-level topo-sort (vs per-entity) is therefore correct, not just cheap. ✓
- **`parentRecord.archetypeNode == nullptr` branch** (line 106) correctly handles the mid-tick "parent destroyed / not yet placed" case without crashing — treats as root, child's world equals its local. Not covered by tests but the code is defensive. Acceptable.
- **`parentWorld` is a value copy** (line 167), so subsequent archetype mutations during the loop cannot dangle the pointer. The per-entity loop never calls a structural-mutation API. ✓
- **`composeNode` calls `getComponentOptional<C_WorldTransform>(parent)` once per node** — *not* per-entity — and the result is cached in the stack-local `parentWorld`. This stays inside the "no per-entity getComponent in tick" rule because the per-entity tick body itself is the no-op `tick(...)`; the lookup runs in `beginTick`. ✓
- **Cycle fallback** (lines 124-138) — `IR_ASSERT` fires in debug; release admits remainder in arbitrary order. The "no progress this pass" detection is the correct cycle predicate for a Kahn-style sort. ✓
- **Per-frame member-vector reuse** — `ordered_` / `pending_` `.clear()` keeps capacity; high-water-mark is set within the first few frames and never reallocates after. ✓
- **`IRMath::quatMul` + `IRMath::rotateVectorByQuat` exist** in `engine/math/include/irreden/ir_math.hpp` and match the documented Hamilton (parent_world ⊗ local) convention. ✓

### Tests

The 10-test suite in `test/ecs/propagate_transform_test.cpp` covers identity defaults, auto-attach, root world==local, 2/3-level translation, 3-level scale+rotation (geometric reasoning), 3-level quat rotation (rotated-axis comparison side-steps quat sign ambiguity correctly — `q90` cubed gives a quat that may be `±expected`, the test compares rotated-vec results, which is the right way), 10-level chain with non-trivial scale verified against an in-test geometric series, both modifier channels via inline-resolve (since the test doesn't register the resolver pipeline), and a zero-entity tick. Coverage of the topo-sort cycle path and the destroyed-parent path is missing — both are defensive code, low priority.

---

### Verdict label

Keeping `fleet:approved` + `fleet:has-nits` (Sonnet's optional pre-merge nit on `candidateSet` / `doneSet` survives this pass; my note above just adds an alternate deeper fix the next perf-touched edit could consider).

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-749.md`), and it is not opted out (`human:no-plan` / `fleet:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

PR #751, opened 2026-05-15
T-200: joints as entities — C_Skeleton + C_Joint scaffolding (replace SoA C_JointHierarchy)

jakildev:
## Summary

- Declare `C_Skeleton` (rig-root component, ordered vector of joint
  EntityIds; the index IS the bone_id space) and `C_Joint` (tag
  component) as the entity-based replacement for the SoA
  `C_JointHierarchy`.
- Deprecate `C_JointHierarchy` via a header banner pointing at
  `C_Skeleton` + `C_Joint`; the legacy struct stays for one release as
  a compile shim so existing consumers keep building.
- Document the entity-based joint model in
  `engine/prefabs/irreden/voxel/CLAUDE.md` — three pieces (`C_Skeleton`,
  `C_Joint`, `CHILD_OF`), bone-index stability under severance (slots
  null out rather than shift), bind-pose deferral until `IRMath::SQT`
  lands with #731 Phase 1, and the optional `C_JointName` follow-up.
- Update `docs/design/entity-editor-epic.md` Phase 2 scope and
  sub-task 2.1 to reflect the refined approach.

Scaffolding only — consumer migration (GPU joint-matrix SSBO uploader,
per-voxel `bone_id` consumption, severance API implementation) is #605
Phase 2 and reads from these new components when it lands. The asset-
side `IRAsset::Rig` → runtime translation in `rig_bridge.hpp` and the
script-side bind-point composition still query `C_JointHierarchy`;
their migration moves alongside #605 Phase 2 work and is intentionally
out of scope here.

## Test plan

- [x] `fleet-build --target IrredenEngineTest` clean.
- [x] `SkeletonComponentsTest.JointTagRegistersAndQueriesByArchetype` —
      `C_Joint` registers as a component type and a `forEachComponent<C_Joint>`
      walk visits every tagged entity.
- [x] `SkeletonComponentsTest.SkeletonHoldsJointEntityIds` — joint
      EntityIds round-trip through `C_Skeleton::joints_` via
      `getComponent`.
- [x] `SkeletonComponentsTest.SeveredSlotIsKNullEntitySentinel` — a
      slot set to `kNullEntity` leaves trailing joint indices stable
      (severance preserves bone-index space).

## Notes for reviewer

- Verify the deprecation banner in `component_joint_hierarchy.hpp`
  reads as a clear migration guide; the body of the header is
  intentionally unchanged so callers still compile.
- The bind-pose field is intentionally absent from `C_Skeleton` —
[...]

PR #770, opened 2026-05-16
T-204: entity: fix sortArchetypeNodesByRelationChildOf — BFS seeds leaves instead of roots

### LINKED ISSUE #750: engine/entity: sortArchetypeNodesByRelationChildOf BFS direction is inverted — drops parent archetypes
## Observation

While implementing T-197 (SQT propagation), I traced \`IREntity::ArchetypeGraph::sortArchetypeNodesByRelationChildOf\` in \`engine/entity/src/archetype_graph.cpp:55\` and found it returns only leaf archetypes — parent archetypes are silently dropped from the sorted output.

## Trace

\`\`\`cpp
// Step 1: collect "parent" NodeIds (archetypes that are pointed to as CHILD_OF parents)
std::unordered_set<NodeId> childNodes;  // misnamed: holds PARENT NodeIds
for (const auto &node : nodes) {
    RelationId childOfRelation = node->getChildOfRelation();
    if (childOfRelation != kNullRelation) {
        NodeId childNodeId = getParentNodeFromRelation(childOfRelation);
        childNodes.insert(childNodeId);
    }
}

// Step 2: queue nodes NOT in that set (i.e., leaves — no incoming CHILD_OF references)
for (const auto &node : nodes) {
    if (childNodes.find(node->id_) == childNodes.end()) {
        nodeQueue.push(node);
        sortedNodes.push_back(node);
    }
}

// Step 3: BFS from queued nodes looking for THEIR children
while (!nodeQueue.empty()) {
    auto currentNode = nodeQueue.front(); nodeQueue.pop();
    for (const auto &node : nodes) {
        if (getParentNodeFromRelation(node->getChildOfRelation()) == currentNode->id_) {
            nodeQueue.push(node);
            sortedNodes.push_back(node);
        }
    }
}
\`\`\`

Step 2 seeds the queue with leaves (archetypes that are not parents of anything). Step 3 then walks from each leaf looking for children — but leaves have none by definition, so step 3 contributes nothing. Parent archetypes never reach \`sortedNodes\`.

## Why this is mostly invisible

The dispatch in \`SystemManager::executeSystem\` only iterates entities in archetypes the sort returned, so root entities (those without a CHILD_OF relation) are silently skipped by systems that use \`RelationParams<...>{CHILD_OF}\`. The two known users — \`GLOBAL_POSITION_3D\` and \`TRIXEL_TO_TRIXEL\` — happen to keep roots and children in archetypes that share the leaf classification (or rely on the next frame's pass having already written values), so the bug doesn't surface visibly.

## Impact

- Any future relational system that needs ALL matching entities (not just leaf archetypes) will be silently incorrect.
- T-197's \`SYSTEM_PROPAGATE_TRANSFORM\` works around this by doing its own topo-sort in \`beginTick\` rather than using the framework's relational dispatch.

## Suggested fix

Invert the BFS: seed the queue with archetypes whose own \`getChildOfRelation()\` is \`kNullRelation\` (true roots in the relation graph), then walk outward via children. The "childNodes" variable should probably be renamed to \`parentNodeIds\` for clarity along the way.

## Suggested model

\`[sonnet]\` — mechanical fix once the directionality is identified. Touch surface: \`engine/entity/src/archetype_graph.cpp\` and possibly a test in \`test/ecs/\` exercising a multi-level CHILD_OF chain through the relational dispatch.

## Cross-references

- Surfaced during T-197 (#734 / PR #749)
- Consumers: \`engine/prefabs/irreden/update/systems/system_update_positions_global.hpp\`, \`engine/prefabs/irreden/render/systems/system_trixel_to_trixel.hpp\`

jakildev:
## Summary
- Inverts BFS seed in `sortArchetypeNodesByRelationChildOf`: was seeding from leaf archetypes (non-parents), now seeds from true roots (nodes with no CHILD_OF relation of their own)
- Parent archetypes now appear before their children in sorted output, fixing silent omission of all non-leaf archetypes from `queryArchetypeNodesRelational(CHILD_OF, ...)`
- Removes the first pass (unordered_set gather) and `<unordered_set>` include — no longer needed
- Guards inner-loop call to `getParentNodeFromRelation` against `kNullRelation` input

## Test plan
- [x] 4 new tests in `test/ecs/sort_archetype_relational_test.cpp`
  - `ParentNodeIncludedInRelationalQuery` — parent entity must be returned (was dropped pre-fix)
  - `ParentNodeAppearsBeforeChildNode` — correct parent-first ordering
  - `ThreeLevelChainAllNodesReturned` — 3-level GP→P→C, all 3 entities present
  - `FlatSetWithNoRelationsReturnsAllNodes` — no relations, all nodes returned
- [x] 723/723 tests pass (`fleet-run IrredenEngineTest`)

Closes #750

### DIFF
--- a/engine/entity/src/archetype_graph.cpp
+++ b/engine/entity/src/archetype_graph.cpp
@@ -3,7 +3,6 @@
 #include <irreden/entity/archetype_graph.hpp>
 
 #include <iterator>
-#include <unordered_set>
 
 namespace IREntity {
 
@@ -52,38 +51,31 @@ void ArchetypeGraph::createAndConnectNode(ArchetypeNode *prevNode, ComponentId n
 
 // Not sure what happens here if a node has a parent with
 // a different archetype.
-std::vector<ArchetypeNode *> ArchetypeGraph::sortArchetypeNodesByRelationChildOf(
-    const std::vector<ArchetypeNode *> &nodes
+std::vector<ArchetypeNode *>
+ArchetypeGraph::sortArchetypeNodesByRelationChildOf(const std::vector<ArchetypeNode *> &nodes
 ) const {
-    // A breath first sort of nodes based on relation heirarchy
+    // BFS from true roots outward so parent archetypes appear before
+    // their children in the output — required for systems that propagate
+    // values down the hierarchy (e.g. SYSTEM_PROPAGATE_TRANSFORM).
     std::vector<ArchetypeNode *> sortedNodes;
     std::queue<ArchetypeNode *> nodeQueue;
 
-    // Step 1: Find all nodes that are children of some other nodes
-    std::unordered_set<NodeId> childNodes;
+    // Seed with roots: nodes that have no CHILD_OF relation of their own.
     for (const auto &node : nodes) {
-        RelationId childOfRelation = node->getChildOfRelation();
-        if (childOfRelation != kNullRelation) {
-            NodeId childNodeId = getParentNodeFromRelation(childOfRelation);
-            childNodes.insert(childNodeId);
-        }
-    }
-
-    // Step 2: Add nodes not in childNodes to the nodeQueue
-    for (const auto &node : nodes) {
-        if (childNodes.find(node->id_) == childNodes.end()) {
+        if (node->getChildOfRelation() == kNullRelation) {
             nodeQueue.push(node);
             sortedNodes.push_back(node);
         }
     }
 
-    // Step 3: Main BFS logic
+    // Walk outward: for each queued node, find children that point to it.
     while (!nodeQueue.empty()) {
         auto currentNode = nodeQueue.front();
         nodeQueue.pop();
         for (const auto &node : nodes) {
             RelationId childOfRelation = node->getChildOfRelation();
-            if (getParentNodeFromRelation(childOfRelation) == currentNode->id_) {
+            if (childOfRelation != kNullRelation &&
+                getParentNodeFromRelation(childOfRelation) == currentNode->id_) {
                 nodeQueue.push(node);
                 sortedNodes.push_back(node);
             }
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -19,6 +19,7 @@ add_executable(IrredenEngineTest
   ecs/modifier_types_test.cpp
   ecs/modifier_vec3_runtime_test.cpp
   ecs/propagate_transform_test.cpp
+  ecs/sort_archetype_relational_test.cpp
   ecs/system_exclude_test.cpp
   math/physics_test.cpp
   math/ir_math_test.cpp
--- a/test/ecs/sort_archetype_relational_test.cpp
+++ b/test/ecs/sort_archetype_relational_test.cpp
@@ -1,0 +1,100 @@
+#include <gtest/gtest.h>
+
+#include <irreden/ir_entity.hpp>
+
+// Tests for sortArchetypeNodesByRelationChildOf exercised via
+// queryArchetypeNodesRelational.  The bug (issue #750) was that the BFS
+// seeded from leaf archetypes (pointed-to by CHILD_OF edges) and walked
+// toward children — but leaves have no children, so parent archetypes
+// were silently absent from the output.  The fix seeds from true roots
+// (archetypes with no CHILD_OF relation of their own) and walks outward.
+
+namespace {
+
+struct C_BfsTestTag {
+    int visited_ = 0;
+};
+
+class SortArchetypeRelationalTest : public testing::Test {
+  protected:
+    IREntity::EntityManager m_entity_manager;
+};
+
+// With the bug, only the child archetype node was returned; the parent
+// archetype node was silently dropped.
+TEST_F(SortArchetypeRelationalTest, ParentNodeIncludedInRelationalQuery) {
+    auto parent = IREntity::createEntity(C_BfsTestTag{});
+    auto child  = IREntity::createEntity(C_BfsTestTag{});
+    IREntity::setParent(child, parent);
+
+    auto include = IREntity::getArchetype<C_BfsTestTag>();
+    auto nodes   = IREntity::queryArchetypeNodesRelational(IREntity::CHILD_OF, include);
+
+    int total = 0;
+    for (auto *node : nodes) total += node->length_;
+    EXPECT_EQ(total, 2) << "both parent and child entities must be returned";
+}
+
+// Parent archetype node must appear before the child so a system that
+// propagates values down the chain reads the correct parent state.
+TEST_F(SortArchetypeRelationalTest, ParentNodeAppearsBeforeChildNode) {
+    auto parent = IREntity::createEntity(C_BfsTestTag{});
+    auto child  = IREntity::createEntity(C_BfsTestTag{});
+    IREntity::setParent(child, parent);
+
+    auto include = IREntity::getArchetype<C_BfsTestTag>();
+    auto nodes   = IREntity::queryArchetypeNodesRelational(IREntity::CHILD_OF, include);
+
+    ASSERT_EQ(nodes.size(), 2u);
+
+    bool parent_first = false, child_second = false;
+    for (auto id : nodes[0]->entities_) {
+        if (id == parent) {
+            parent_first = true;
+            break;
+        }
+    }
+    for (auto id : nodes[1]->entities_) {
+        if (id == child) {
+            child_second = true;
+            break;
+        }
+    }
+    EXPECT_TRUE(parent_first)  << "parent entity should be in the first returned node";
+    EXPECT_TRUE(child_second)  << "child entity should be in the second returned node";
+}
+
+// Three-level chain: grandparent → parent → grandchild.  With the bug,
+// only the grandchild (the leaf) was returned.  After the fix all three
+// archetype nodes must appear.
+TEST_F(SortArchetypeRelationalTest, ThreeLevelChainAllNodesReturned) {
+    auto gp     = IREntity::createEntity(C_BfsTestTag{});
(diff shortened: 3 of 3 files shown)


jakildev (commented):
## Review — T-204: fix sortArchetypeNodesByRelationChildOf — BFS seeds leaves instead of roots

**Verdict:** approve

---

### Root-cause and fix analysis

The bug (issue #750): the original algorithm labelled as "BFS" was actually walking in the wrong direction. Step 1 gathered the NodeIds of archetype nodes that are referenced as *parents* by any CHILD_OF relation (i.e., parent archetypes), storing them in `childNodes`. Step 2 seeded the BFS queue with every node *not* in `childNodes` — i.e., leaf archetypes (those with no children pointing at them). Because leaf archetypes by definition have no children in the CHILD_OF graph, the BFS inner loop found nothing to enqueue, and parent archetypes (the very nodes excluded from the seed set) were silently absent from `sortedNodes`.

The fix correctly inverts the seed direction: nodes with `getChildOfRelation() == kNullRelation` are true roots (not a CHILD_OF anyone), and BFS walks *outward* from them. This guarantees parent archetypes precede their children, which is the exact contract `SYSTEM_PROPAGATE_TRANSFORM` relies on.

The additional guard `childOfRelation != kNullRelation` before calling `getParentNodeFromRelation(childOfRelation)` in the inner loop is also correct — without it, a root node in the candidate list (with `kNullRelation`) would call `getParentNodeFromRelation(kNullRelation)`, whose return value could collide with a valid NodeId.

The `<unordered_set>` include removal is clean; Step 1 is gone entirely.

---

### Praise

- The logic inversion from "seed leaves" to "seed roots" is exactly minimal — no extra state, no behavioral changes beyond the bug fix.
- The new comment block names both WHAT changed and WHY, citing the concrete downstream consumer (`SYSTEM_PROPAGATE_TRANSFORM`).
- Four tests cover every meaningful case: single-pair parent visibility, parent-before-child ordering, transitive three-level chain, and flat no-relation set. The bug-regression test (`ParentNodeIncludedInRelationalQuery`) is the most critical and is present.
- 723/723 tests pass.

---

### Checklist

**ECS invariants:** No per-entity `getComponent` in a tick function. The fix lives in `archetype_graph.cpp` (graph utility, not a tick path). No structural changes mid-iteration. No allocations in hot loops. ✓

**Math primitives:** No `glm::` or `std::` math calls outside `engine/math/`. ✓

**Naming:** `sortedNodes`, `nodeQueue` — camelCase locals ✓. `C_BfsTestTag` — `C_` prefix on component ✓. `visited_` — trailing `_` on struct member ✓. No anonymous namespaces in headers (test is a `.cpp`). ✓

**Render pipeline:** No render files, no shaders. ✓

**Cross-host smoke:** Diff does not touch `engine/render/`, any `*.glsl`, or any `*.metal`. No smoke tagging required. ✓

**Tests:** 4 targeted tests anchoring the corrected BFS contract. 723/723 pass. ✓

---

Opus recheck required: PR touches `engine/entity/src/archetype_graph.cpp` (ECS core — archetype graph traversal algorithm).

jakildev (commented):
## Opus recheck — T-204: archetype BFS root/leaf inversion

**Verdict:** approve

---

### Sonnet's analysis confirmed

Sonnet flagged this for Opus recheck because it touches `engine/entity/src/archetype_graph.cpp` — ECS core. On closer read I confirm Sonnet's root-cause walkthrough is correct.

Tracing the OLD algorithm with `kNullRelation = 0`, `kNullNode = 0`, the per-node `getChildOfRelation()` returning `kNullRelation` for nodes without a CHILD_OF and `getParentNodeFromRelation(rel)` returning the parent NodeId:

- The `childNodes` set was misnamed — it actually collected the NodeIds of *parents* (those pointed-at by any CHILD_OF in the input set).
- The seed picked nodes *not* in that set, i.e. nodes that no one in `nodes` points at — leaves.
- The inner BFS expanded "find children of currentNode", but leaves by definition have no children, so the queue drained immediately.

In a chain GP→P→C with all three matching `include`: OLD code returned only `[C]`. NEW code seeds GP (root, no CHILD_OF), then BFS pushes P, then C. Returns `[GP, P, C]` in correct topological order. ✓

### Downstream impact — why this was actually broken in production

Two prefab systems use `RelationParams<>{CHILD_OF}` and thus depend on `sortArchetypeNodesByRelationChildOf`:

1. `system_update_positions_global.hpp:30` — `GLOBAL_POSITION_3D` composes child global = local + parent global. The buggy sort dropped the parent archetype node entirely. The system was silently degenerating to `globalPos = localPos` (since the optional `parentPositionGlobal` value-or-defaulted to `(0,0,0)`). This is the real-world functional bug T-204 fixes.
2. `system_trixel_to_trixel.hpp:72` — `TRIXEL_TO_TRIXEL` uses CHILD_OF to look up the parent canvas's `C_TriangleCanvasTextures` via `relationTick`. The buggy sort dropped the parent NODE but the child node's `relationTick` reads the parent ENTITY's component directly, so this path was less obviously broken. Order-correctness still improves with the fix.

`system_propagate_transform.hpp:64-141` (T-197/T-199, just landed in #749/#756) explicitly works around the bug with its own `pending_`/`candidateSet_`/`doneSet_` topological sweep, sidestepping the relational query entirely. Once T-204 lands, that workaround could be migrated back to the standard relational API in a follow-up — out of scope here.

### Other things checked

**Defensive guard added:** `if (childOfRelation != kNullRelation && getParentNodeFromRelation(childOfRelation) == currentNode->id_)` correctly avoids calling `getParentNodeFromRelation(kNullRelation)` for non-CHILD_OF nodes. Note: `getParentNodeFromRelation(kNullRelation)` returns `kNullNode == 0`, and the base archetype node has `id_ == 0` (since `m_nodeCount++` post-increments starting at 0 in `ArchetypeGraph::ArchetypeGraph()`). Without the guard, if a future caller passed the base node into `nodes`, every node-without-CHILD_OF would match `currentNode->id_ == 0`, producing a false expansion. The guard closes that latent footgun. ✓

**Edge case acknowledged but not addressed:** Both OLD and NEW code drop nodes whose CHILD_OF parent is outside the input set (parent has a different archetype not matching the `include` filter). The preserved comment "Not sure what happens here if a node has a parent with a different archetype" still applies — failure mode is just different now (OLD seeded the orphan; NEW drops it). Realistic CHILD_OF queries (transform propagation, canvas hierarchies) keep parent and child in the same archetype filter, so this stays a theoretical concern. Tracking as documentation debt rather than a fix request.

**No diamond/cycle concerns:** `ArchetypeNode::getChildOfRelation()` returns a single RelationId, so each node has at most one parent — no diamonds. Cycles in CHILD_OF would be an engine bug; both old and new BFS would skip cycle members rather than infinite-loop. Acceptable.

**Complexity:** O(N²) inner-loop scan per BFS pop, same as before. Fine for the scale of archetype nodes we actually have.

**Single-pop-per-node invariant:** Each node has at most one CHILD_OF parent, so the inner-loop predicate matches for exactly one currentNode across the entire BFS run — each child is pushed exactly once. No need for a visited set.

### Items already in Sonnet's checklist — confirmed at a glance

- Naming (`sortedNodes`, `nodeQueue`, `C_BfsTestTag`, `visited_`) ✓
- No per-entity `getComponent` in a tick ✓ (graph utility, not a tick path)
- No allocations in hot loops ✓ (one-shot sort)
- IRMath math-primitive rule N/A (no math) ✓
- Test placement and CMake wiring ✓

### Test coverage

Four targeted tests cover: parent-included, parent-before-child ordering, three-level chain completeness, flat (no-relation) pass-through. The two-level test exercises the exact regression from the buggy code (parent silently dropped). 723/723 pass per the PR body.

The Sonnet checklist did not catch a blatant miss in this round.

---

jakildev:
judgement-level resolution.

Conflicted files:
- `test/CMakeLists.txt` — master: `8f096f77 T-200: joints as entities — C_Skeleton + C_Joint scaffolding (replace SoA C_JointHierarchy) (#751)`; PR: `21d1a60f entity: fix sortArchetypeNodesByRelationChildOf BFS direction`

Root cause: both sides added a new test source into the alphabetical slot just after `ecs/propagate_transform_test.cpp`. Master added `ecs/skeleton_components_test.cpp`; this PR added `ecs/sort_archetype_relational_test.cpp`. The merger's sort-merge case is intentionally scoped to TASKS.md only — it does not generalize to other ordered lists, even when the resolution looks mechanical, because a CMake list addition can carry implicit dependencies (e.g., a new test file referenced elsewhere) that a blind sort-merge would miss.

Resolution path for the opus-worker:

  git fetch origin
  git checkout claude/T-204-sort-archetype-bfs-fix
  git rebase origin/master
  # In test/CMakeLists.txt, keep BOTH new entries — master adds
  # ecs/skeleton_components_test.cpp and this PR adds
  # ecs/sort_archetype_relational_test.cpp. Place them in alphabetical
  # order between propagate_transform_test.cpp and system_exclude_test.cpp.
  git add test/CMakeLists.txt
  git rebase --continue
  # Verify the build:
  fleet-build --target IrredenEngineTest
  git push --force-with-lease

SHA pair: master=6754fb9ba0212a1e4da144b3a484cb060dd9f232 × PR=21d1a60fbf99915b1289aba70cb9313b65da84cc

— fleet merger

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-770.md`), and it is not opted out (`human:no-plan` / `fleet:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

PR #771, opened 2026-05-16
fleet: drop C-c from dispatch send-keys (Windows alert sound on every worker spin-up)

jakildev:
## Summary

Stops the Windows Default Beep sound that played every time a fleet worker pane received a fresh dispatch — caused by SIGINT visualization in WSLg / Windows Terminal when `tmux send-keys C-c` hit a bash prompt.

## Background

`scripts/fleet/fleet-dispatcher` reset the worker pane's input line before typing each dispatched command with `tmux send-keys C-c C-u "<cmd>" C-m`. The C-c was there to abort:
1. Stray "Ca" buffered from an incomplete tmux prefix
2. Multi-line bash continuations

Case 1 is fully covered by C-u (`unix-line-discard`) which kills the line buffer regardless of stray characters. Case 2 is vanishingly rare in fleet panes — humans don't type there — and self-heals on the next dispatcher tick if it happens.

## What changed

- Drop C-c from the send-keys; just `C-u "<cmd>" C-m`
- Collapse the two-call split (with 50ms sleep) back into a single send-keys. The split was added to dodge the bash-5.2 readline race on Ubuntu 24.04 where the post-prompt-redraw window after C-c eats the first input byte (the "leet-dispatch-wrap: command not found" loop). With C-c gone, no redraw, no race.
- Leading-space sacrificial char in `build_dispatch_command` stays as defense-in-depth.
- Updated the inline rationale comment.

## Verification

Live test on a worker bash pane: typed stray "typed_garbage_no_enter", then `tmux send-keys C-u " echo NOISE_TEST_$RANDOM" C-m` — input cleared, command ran cleanly with the leading-space absorbed, no SIGINT echo, no Windows beep.

## Cross-platform safety

- macOS: bash 3.2 (or zsh) handles C-u identically; no Mac noise to begin with, but no regression.
- Windows-native MSYS2: same behavior — bash interactive at a prompt, C-u clears.
- All hosts: the original C-c-induced bash-5.2 readline race is gone, so the `assume-paste-time` / `escape-time` tmux workarounds in operator dotfiles become unnecessary (they remain defensive but optional).

## Test plan

- [x] `bash -n scripts/fleet/fleet-dispatcher` passes.
- [x] Manual send-keys repro with stray input shows clean dispatch.
- [ ] Live observation: restart dispatcher with the patch, watch a worker spin up — no Windows beep, dispatched command runs cleanly.
- [ ] Run a queue-manager iteration end-to-end without seeing the dropped-first-byte error.

### DIFF
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -858,28 +858,33 @@ dispatch_role() {
         local cmd
         cmd=$(build_dispatch_command "$role" "$pane_key")
         log "dispatching $role -> $pane_id"
-        # C-c aborts any partial input the human may have typed into
-        # the idle pane (including multi-line continuations); C-u then
-        # clears the input line for good measure. Without this the
-        # typed-but-uncommitted prefix concatenates with our dispatch
-        # command — observed: stray "Ca" turned the next dispatch into
-        # "Caclaude --model …" / "zsh: command not found: Caclaude".
-        # Safe at a shell prompt — find_idle_panes_for_role only picks
-        # panes whose pane_current_command is bash/zsh/sh/fish, so C-c
-        # acts on the input loop, not a running command.
-        # Split the burst: C-c interrupts whatever's at the prompt and
-        # triggers bash to redraw; without a gap, the first char of
-        # "$cmd" arrives while readline is mid-redraw and gets eaten
-        # ("fleet-dispatch-wrap" → "leet-dispatch-wrap: command not
-        # found"). 50ms is enough on tmux 3.4 / Ubuntu 24.04; cheap on
-        # every other host.
-        if ! dispatch_send_keys -t "$pane_id" C-c C-u; then
-            log "send-keys (reset) timeout/failed for $role -> $pane_id; trying next idle pane"
-            continue
-        fi
-        sleep 0.05
-        if ! dispatch_send_keys -t "$pane_id" "$cmd" C-m; then
-            log "send-keys (cmd) timeout/failed for $role -> $pane_id; trying next idle pane"
+        # Reset the input line with C-u (kill-line) before sending the
+        # dispatch command. Catches any partial input the human may have
+        # typed into the idle pane (stray "Ca" from an in-flight tmux
+        # prefix that didn't complete a binding, etc.) — without it the
+        # uncommitted prefix concatenates with our dispatch command,
+        # observed: "Caclaude --model …" / "command not found: Caclaude".
+        #
+        # The earlier implementation also sent C-c. Dropped because:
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — fleet: drop C-c from dispatch send-keys

**Verdict:** approve

---

### Praise

- The root-cause analysis is exactly right: `find_idle_panes_for_role` only selects bash/zsh/sh/fish panes with no foreground process, so SIGINT at a bare prompt echoes `^C\n` and triggers the terminal bell — achieving nothing useful beyond audible annoyance.
- `C-u` (unix-line-discard) is strictly more precise than C-c for the actual goal — it clears the line buffer without touching the process group. The prior belt-and-suspenders combination turned out to be the-suspenders-causing-the-problem.
- Collapsing back to a single `dispatch_send_keys` call is correct: the 50ms gap existed solely to let readline finish its post-C-c redraw before the first byte of the command arrived. Without C-c, no redraw occurs, no byte is eaten, the sleep is unnecessary.
- The updated comment is exemplary: covers what C-u does, why C-c was dropped (three distinct reasons), acknowledges the multi-line continuation edge case and correctly dismisses it, and preserves the rationale for the leading-space defense-in-depth. Future maintainers won't need to re-derive any of this.

### Checklist

**ECS invariants:** Fleet tooling only — no C++ or ECS code. N/A ✓

**Math primitives:** No C++ code. ✓

**Naming:** No new symbols introduced. ✓

**Multi-line continuation edge case:** If a fleet pane somehow enters bash `>` continuation state, the new code sends `C-u "$cmd" C-m` — which clears the continuation line and submits the dispatch command into the open continuation, likely producing a bash syntax error. The dispatch fails, the log records a failure, and the next tick self-heals as described in the PR body. Acceptable trade-off given humans do not type in fleet panes. ✓

**Cross-host smoke:** `scripts/fleet/fleet-dispatcher` is a bash script, not a render path. Smoke tagging not applicable. ✓

**Tests:** `bash -n` syntax check passes. Live send-keys repro with stray input confirmed clean. The two unchecked live items (dispatcher restart + full queue-manager iteration) are runtime-only verifications that cannot be captured in the diff; the PR body documents manual validation of the core send-keys behavior. Acceptable for a fleet-script fix of this scope. ✓

---

Opus recheck not required.

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-771.md`), and it is not opted out (`human:no-plan` / `fleet:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

PR #776, opened 2026-05-16
T-208: modifier: writer-owned slot API — upsertBySource to eliminate per-frame push_back churn

### LINKED ISSUE #758: modifier: writer-owned slot API to eliminate per-frame push_back churn
## Context

Follow-up to T-192 (#746). The current modifier framework's `push` API is built for transient pushes — each call appends to `C_Modifiers::modifiersVec3_`, and the entry is removed by `MODIFIER_DECAY` once `ticksRemaining_` reaches 0.

`PERIODIC_IDLE_POSITION_OFFSET` (`engine/prefabs/irreden/update/systems/system_periodic_idle_position_offset.hpp:40-48`) abuses this pattern for a steady-state writer: every UPDATE tick it pushes a vec3 modifier with `ticksRemaining_=1`, the next tick's `MODIFIER_DECAY` removes it, then this system pushes again. For an entity that bobs continuously this is constant churn on the vec3 vector every frame.

## Scope

Add a writer-owned slot API to `IRPrefab::Modifier::`:

```cpp
void upsertBySource(
    IREntity::EntityId target,
    IRComponents::FieldBindingId field,
    IRComponents::TransformKind kind,
    IRMath::vec3 param,
    IREntity::EntityId source
);
```

Semantics: find the existing entry whose `(source_, field_, kind_)` triple matches and update its `param_`; if none, push a new entry with `ticksRemaining_ = -1` (no decay). Scalar overload too.

Migrate `PERIODIC_IDLE_POSITION_OFFSET` to use `upsertBySource` — drop the `ticksRemaining_=1` hack and the dependency on `MODIFIER_DECAY` running earlier in the pipeline. Allocates only on first push; zero allocation thereafter.

## Why

Per-frame `push_back` + decay cycle for a "this channel writes every frame forever" pattern is wasteful and the `ticksRemaining_=1` rule is subtle (see comment in the writer's header). A slot-update API encodes "this writer owns one entry" directly.

## Suggested model

`[sonnet]` — small mechanical addition once the API shape is locked.

## Suggested area

`engine/prefabs/irreden/common`, `engine/prefabs/irreden/update`

## Acceptance criteria

- New `upsertBySource` API in `engine/prefabs/irreden/common/modifier.hpp` (scalar + vec3 overloads).
- `PERIODIC_IDLE_POSITION_OFFSET` uses `upsertBySource` and no longer pushes with `ticksRemaining_=1`.
- Idle bob in default + perf_grid creations continues to look identical to master.
- Pipeline ordering comment in `system_periodic_idle_position_offset.hpp` updated — no longer requires `MODIFIER_DECAY` to run before it.
- `docs/design/modifiers.md` updated to describe upsert as the canonical "steady-state writer" pattern, with push reserved for transient effects.

## Related

- Predecessor: #746 (T-192 — initial migration to vec3 modifier)
- Sibling improvements: writer API ergonomics + inline-compose generalization (filed separately)

jakildev:
## Summary

- Adds six inline overloads to `IRPrefab::Modifier::` in `modifier.hpp`: `upsertBySource` (scalar + vec3), `upsertBySourceGlobal` (scalar + vec3), `upsertBySourceInPlace` (scalar + vec3)
- Slot key is the triple `(source_, field_, kind_)`: on hit overwrites `param_` and resets `ticksRemaining_ = -1`; on miss appends with `ticksRemaining_ = -1`
- Migrates `PERIODIC_IDLE_POSITION_OFFSET` to `upsertBySourceInPlace`, eliminating per-frame `push_back` churn and the `MODIFIER_DECAY` pipeline dependency
- Documents the slot contract in `docs/design/modifiers.md` and updates `CLAUDE.md` files for `common/` and `update/`

## Test plan

- [x] 9 new unit tests cover: `FirstCallAppends`, `SecondCallOverwrites`, `DifferentKindGetsItsOwnSlot`, `DifferentSourceGetsItsOwnSlot`, `OverridesPriorTickRemaining`, and the 100-tick size-stays-1 invariant for both scalar and vec3 paths
- [x] Full test suite: 728/728 pass (`IrredenEngineTest`)
- [x] `IRShapeDebug` builds clean with the migrated system

Closes #758

### DIFF
--- a/docs/design/modifiers.md
+++ b/docs/design/modifiers.md
@@ -435,6 +435,84 @@ needs all four predecessors.
   state (`kind_ == ADD`, `field_ == 0`). Fail fast rather than silently
   adding an invalid modifier to the vector.
 
+## Writer-owned slot API — `upsertBySource` (T-208)
+
+The `push()` family always appends a new entry to the modifier vector and
+relies on `ticksRemaining_` decay to remove stale copies. That model forces
+every steady-state writer (e.g. an idle-bob system that fires every tick) to:
+
+1. Run `MODIFIER_DECAY` before itself in the pipeline, and
+2. Set `ticksRemaining_ = 1` to survive the current frame's compose and be
+   removed by the next frame's decay.
+
+This is fragile: it ties the system's pipeline position to the decay system,
+and the per-frame `push_back` allocates on the first tick and reallocates
+whenever the vector outgrows its capacity.
+
+`upsertBySource` (T-208) replaces this with a **writer-owned slot** pattern:
+
+### Slot contract
+
+The triple `(source_, field_, kind_)` is the slot key. At most one entry with
+a given triple may exist in the modifier vector (per-entity or global).
+Callers that own a slot call `upsertBySource` instead of `push` — on the first
+call the slot is created; on subsequent calls only `param_` is overwritten and
+`ticksRemaining_` is reset to `-1`. The slot persists until the source entity
+is destroyed or `removeBySource` is called explicitly.
+
+`push` remains the correct choice for **transient effects** that should expire
+naturally (e.g. a one-shot screen-shake, a key-triggered pulse). Use `push`
+with a `ticksRemaining` countdown when the caller does NOT own the slot
+long-term. Use `upsertBySource` when the caller writes the same slot every
+tick forever (e.g. idle-bob, a continuous aura).
+
+### API additions
+
+```cpp
+namespace IRPrefab::Modifier {
+
+// External writers (Lua, input handlers, init code) — performs the same
+// getComponentOptional + field-type checks as push().
+void upsertBySource(EntityId target, FieldBindingId field,
+                    TransformKind kind, float param, EntityId source);
+void upsertBySource(EntityId target, FieldBindingId field,
+                    TransformKind kind, IRMath::vec3 param, EntityId source);
+
+// Targeting the singleton globals entity.
+void upsertBySourceGlobal(FieldBindingId field, TransformKind kind,
+                           float param, EntityId source);
+void upsertBySourceGlobal(FieldBindingId field, TransformKind kind,
+                           IRMath::vec3 param, EntityId source);
+
+// For system ticks — caller already holds C_Modifiers& from archetype
+// iteration; skips getComponentOptional and the field-type registry probe.
+void upsertBySourceInPlace(C_Modifiers &mods, FieldBindingId field,
+                            TransformKind kind, float param, EntityId source);
+void upsertBySourceInPlace(C_Modifiers &mods, FieldBindingId field,
+                            TransformKind kind, IRMath::vec3 param, EntityId source);
+}
+```
+
+`upsertBySourceInPlace` is the canonical form for tick functions: it bypasses
+the `getComponentOptional` lookup and the registry type-check because the
+system already has the component reference from the archetype iterator and
+already knows the field type from init-time registration.
+
+### Pipeline impact
+
+`PERIODIC_IDLE_POSITION_OFFSET` (the first consumer) no longer requires
+`MODIFIER_DECAY` to run before it. The pipeline ordering becomes:
+
+```
+PERIODIC_IDLE → PERIODIC_IDLE_POSITION_OFFSET
+→ ... → GLOBAL_POSITION_3D → APPLY_POSITION_OFFSET
+```
+
+Existing creations that still run `MODIFIER_DECAY` before
+`PERIODIC_IDLE_POSITION_OFFSET` are safe — `MODIFIER_DECAY` leaves entries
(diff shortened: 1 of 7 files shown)


jakildev (commented):
## Review — T-208: modifier: writer-owned slot API — upsertBySource

**Verdict:** approve

---

### Praise

- The slot-key triple `(source_, field_, kind_)` is the correct granularity: one slot per source per logical field per transform kind. Callers that push a continuous rotation modifier and a continuous translation modifier from the same source entity get independent slots, as expected.
- `upsertBySourceInPlace` (the hot-path form) is correctly scoped: it skips `getComponentOptional` and the field-type registry check because a system tick already holds the component reference and knows the field type from init-time registration. The `upsertBySource` external API retains both guards. The two-tier design matches the existing `push` vs direct `modifiers_.push_back` pattern used in the old `PERIODIC_IDLE_POSITION_OFFSET` — but now encapsulated rather than ad-hoc.
- Removing the `MODIFIER_DECAY` ordering constraint from `PERIODIC_IDLE_POSITION_OFFSET` is a real improvement: the comment in the old system header had to explain a subtle `ticksRemaining_ = 1` vs `ticksRemaining_ = 2` distinction that required understanding the pipeline position relative to decay. The new header comment is two lines instead of ten and is correct on its face.
- The `OverridesPriorTickRemaining` test catches the transition case (caller switches from `push(ticksRemaining=1)` to `upsertBySource` on the same triple) and verifies the slot is correctly adopted into the no-decay regime. That's the one non-obvious interaction.
- Documentation in `docs/design/modifiers.md` is well-structured: states the problem, the contract, the API additions, and the pipeline impact. The "Decay is safe" backward-compat claim is explicitly documented.
- 9 new tests, 728/728 pass.

---

### Checklist

**ECS invariants:** No per-entity `getComponent` inside a tick function. `upsertBySourceInPlace` takes `C_Modifiers&` from the archetype iterator directly — no lookup inside the loop. On first tick, `push_back` is called on `mods.modifiersVec3_` (a field inside the component); this mutates component data but does NOT add or remove components, so the archetype layout is unchanged and iterator pointers remain valid. On subsequent ticks only the overwrite path runs (no allocation). ✓

**Math primitives:** No `glm::` or `std::` math calls outside `engine/math/`. Uses `IRMath::vec3` correctly. ✓

**Naming:** `upsertBySource`, `upsertBySourceGlobal`, `upsertBySourceInPlace` — camelCase free functions in namespace ✓. `m_entity_manager`, `m_field` — test fixture members follow the existing `m_` prefix pattern already established in this file's prior fixtures ✓. `kField` — `k` prefix on local constant ✓.

**Decay-system backward compat:** `MODIFIER_DECAY` skips entries with `ticksRemaining_ == -1`. Slots written by `upsertBySource` always have `ticksRemaining_ = -1`, so they survive MODIFIER_DECAY if it remains in the pipeline. The claim in the docs is consistent with this invariant.

**Cross-host smoke:** Diff does not touch `engine/render/`, any `*.glsl`, or any `*.metal`. No smoke tagging required. ✓

**Tests:** 9 new tests covering scalar/vec3 first-append, second-call overwrite, different-kind isolation, different-source isolation, prior-push adoption, and 100-tick size-stays-1 invariant for both InPlace variants. ✓

---

### Nits

**N1 — `upsertBySourceGlobal` lacks unit tests.** Both the scalar and vec3 overloads are untested. The implementation is isomorphic to the entity-targeted variants (same slot scan, same overwrite logic, just targets `C_GlobalModifiers` instead of `C_Modifiers`), so the risk is low. A pair of tests mirroring `FirstCallAppends` and `SecondCallOverwrites` for the globals path would close the gap.

**N2 — `OverridesPriorTickRemaining` only covers vec3.** The scalar path has the same semantics — a caller switching from `push(ticksRemaining=N)` to `upsertBySource` should reset `ticksRemaining_` to -1 and adopt the slot. Add a matching scalar test for symmetry.

---

Opus recheck not required.

jakildev:
Addressed review nits N1 and N2:

- **N1** — Added scalar and vec3 `upsertBySourceGlobal` test fixtures (`IRModifierGlobalScalarUpsertTest` and `IRModifierGlobalVec3UpsertTest`), each with `FirstCallAppends` and `SecondCallOverwrites` tests. Fixtures call `singletonEntity<C_GlobalModifiers>()` to provision the globals entity, then read back via `globalsEntity()` + `getComponent<C_GlobalModifiers>` to verify slot contents and `ticksRemaining_ = -1`.
- **N2** — Added `OverridesPriorTickRemaining` to `IRModifierScalarUpsertTest`, mirroring the existing vec3 test: a prior `push(ticksRemaining=1)` slot is adopted and reset to -1 by `upsertBySource`.

All 733 tests pass (build: IrredenEngineTest). — sonnet-author

PR #775, opened 2026-05-16
T-202: enable Linux/OpenGL backend on WSLg (GL 4.5 + GLSL hygiene)

jakildev:
## Summary

First installment for T-202 (#757). Brings the engine from "won't create
a GL context" to "renders the shape_debug demo correctly" on
`linux-debug` (WSL2/Ubuntu 24.04/gcc-13, Mesa 25.2.8 via the d3d12
driver).

Three minimal, mechanical changes — each is a hard requirement for
booting on WSLg, and none change rendering semantics on Linux native /
Windows / macOS:

1. **OpenGL context floor: 4.6 → 4.5 core.** WSLg's Mesa d3d12 driver
   advertises 4.5 (the 4.6 hint fails with `GLXBadFBConfig`). Real
   Linux/Windows GPU drivers still expose 4.6, but the engine uses no
   4.6-only GLSL features (no SPIR-V, no `gl_HelperInvocation`), so
   targeting 4.5 lets a single binary boot everywhere. Shader `#version`
   lines drop from `460 core` → `450 core` across all 29 GLSL files in
   sync.
2. **Rename GLSL reserved-word locals.** The strict Mesa GLSL frontend
   correctly rejects `packed` and `flat` as variable names (they're
   layout/interpolation qualifiers). Tolerated by some drivers, not by
   Mesa. Renamed `packed` → `packedDepth` in the sun-shadow shaders
   and `flat` → `flatIndex` in `c_propagate_light_volume.glsl`. The
   Metal counterparts are mirrored for backend parity even though MSL's
   tolerance differs.
3. **Parameterize `localIDToFace_2x3()`** in `ir_iso_common.glsl` so it
   takes `gl_LocalInvocationID.xy` as an argument instead of reading
   the built-in directly. Strict GLSL frontends error on the built-in
   even from *unused* functions when the including shader is
   non-compute — `f_trixel_to_framebuffer.glsl` (fragment) includes
   this header. The Metal counterpart already had the parameterized
   form; GLSL now matches.

`engine/window/CLAUDE.md` updated to document the 4.5 floor and
rationale, so future workers don't bump the hint back to 4.6 without
hitting the same WSLg wall.

## After: shape_debug shot 3 rendering on Linux/WSLg

(The demo could not create a GL context before this PR — there is no
"before" screenshot to compare. Shots 1–6 all render correctly; shot 7
is the follow-up below.)

After: shape_debug zoom=4 origin on Linux

## Known follow-up

[...]

PR #779, opened 2026-05-16
T-210: generalize APPLY_POSITION_OFFSET into applyVec3ModifierTo<>

### LINKED ISSUE #760: modifier: generalize APPLY_POSITION_OFFSET into reusable APPLY_VEC3_MODIFIER_TO<field, component> pattern
## Context

Follow-up to T-192 (#746). The modifier framework currently has two distinct compose-and-apply paths:

1. **Structured resolver path** (general): writers push → `MODIFIER_RESOLVE_GLOBAL`/`RESOLVE_EXEMPT` compose into `C_ResolvedFields` → consumers `get()` cached value.
2. **Inline compose path** (`APPLY_POSITION_OFFSET` only): writer pushes → consumer system composes inline against a base of `vec3(0)` and writes directly into the target component, skipping `C_ResolvedFields`.

The inline path was chosen because the position-offset channel has exactly one consumer, so caching would be pure overhead. But `APPLY_POSITION_OFFSET` (`engine/prefabs/irreden/update/systems/system_apply_position_offset.hpp`) is a one-off — a future per-frame vec3 perturbation (gizmo nudge already exists, hit-stagger, screen-space jitter) would each need its own bespoke apply system following the same template.

## Scope

Generalize the inline-compose pattern. Sketch:

```cpp
// Generic "compose this vec3 channel inline and add to this component's vec3 member."
template <IRComponents::FieldBindingId Field, typename TargetComponent, IRMath::vec3 TargetComponent::* Member>
struct ApplyVec3ModifierTo {
    static SystemId create(const char* name) {
        return createSystem<TargetComponent, IRComponents::C_Modifiers>(
            name,
            [](TargetComponent& tgt, IRComponents::C_Modifiers& mods) {
                tgt.*Member += IRPrefab::Modifier::detail::composeForFieldVec3(
                    IRMath::vec3(0.0f), Field, mods.modifiersVec3_);
            });
    }
};
```

Then `APPLY_POSITION_OFFSET` becomes a single-line instantiation, and a future "screen-shake" feature is one line + a new field id. Exact template shape needs design work — `FieldBindingId` isn't a compile-time constant since fields are registered at init, so the parameterization may need to be a runtime arg captured into the lambda rather than a template parameter. Architect to pick the shape.

Open design questions:
- Template-parameterized vs. runtime-parameterized field id (registration is dynamic).
- Should the apply step take `ADD` vs. `SET` vs. `OVERRIDE` semantics as a parameter, or is `ADD` the universal pattern for inline-apply?
- Naming: is `APPLY_VEC3_MODIFIER_TO` clear enough? Alternative: `FOLD_MODIFIER_CHANNEL`.

## Why

Today, adding a new inline-compose vec3 channel is "copy-paste `APPLY_POSITION_OFFSET`." Generalizing it makes the second consumer cheap and locks in one mental model — the framework has exactly one inline-apply system shape, parameterized.

## Suggested model

`[opus]` — architectural design (template vs. runtime parameterization, ergonomics) plus mechanical port.

## Suggested area

`engine/prefabs/irreden/common`, `engine/prefabs/irreden/update`, `engine/prefabs/irreden/render`

## Acceptance criteria

- `APPLY_POSITION_OFFSET` reimplemented as an instance of the generic pattern (or, if architect picks runtime parameterization, as a thin caller of the generic helper).
- `docs/design/modifiers.md` documents the inline-apply pattern alongside the structured-resolver path, with guidance for when to pick which.
- Idle bob in default + perf_grid creations continues to look identical to master.
- No regression in `IRShapeDebug`, `voxel_editor`, or any demo that uses position offset.

## Related

- Predecessor: #746 (T-192)
- Sibling improvements: writer-owned slot + named push APIs (filed separately).
- Future consumers that motivate this generalization: gizmo nudge (already inline), hit-stagger (not yet built), screen-space jitter (not yet built).

jakildev:
## Summary
- Generalizes the `APPLY_POSITION_OFFSET` shape — *iterate `<TargetComponent, C_Modifiers>`, compose one vec3 field against `vec3(0)`, ADD the result to a vec3 member of the target component* — into `IRPrefab::Modifier::applyVec3ModifierTo<TargetComponent, Member>(name, field)` in a new header `engine/prefabs/irreden/common/modifier_apply.hpp`.
- `APPLY_POSITION_OFFSET` reduces to a single-line instantiation. Behavior is byte-identical (same `composeForFieldVec3` call, same archetype iteration `<C_PositionGlobal3D, C_Modifiers>`, same in-place `pos_ +=`).
- `docs/design/modifiers.md` gets a new "Inline-apply pattern (T-210)" section that captures the design choices (template vs. runtime parameterization, ADD-only combine, no globals integration) and the inline-apply vs. structured-resolver discriminator.
- `engine/prefabs/irreden/common/CLAUDE.md` points readers at the new factory and links the discriminator.

## Design choices (architect picks from issue)
- **Member pointer = template parameter** (compile-time, so the per-tick body collapses to `target.*Member += compose(...)` with no field-name lookup).
- **`field` = runtime arg** captured into the lambda. `FieldBindingId`s register lazily at init, so a non-type template form would force every consumer to make their field id `constexpr` — they're not.
- **Combine mode = ADD against `vec3(0)`, hard-coded.** That's the universal "per-frame additive delta" pattern that motivates the inline-apply path. MUL/SET/OVERRIDE semantics already live *inside* `composeForFieldVec3`; the outer apply step doesn't need to be parameterized.
- **No globals integration.** Inline path reads only the entity's `mods.modifiersVec3_`, matching `APPLY_POSITION_OFFSET`'s pre-existing behavior. Channels that must respond to globals belong on the structured-resolver path.
- **Naming = `applyVec3ModifierTo`** per the issue sketch. `FOLD_MODIFIER_CHANNEL` was an alternative — kept the issue's name because it reads better at the call site (`applyVec3ModifierTo<C_PositionGlobal3D, &C_PositionGlobal3D::pos_>` says exactly what it does).

## Test plan
- [x] `fleet-build --target IRPerfGrid` — clean (perf_grid is one of two demos that instantiates `System<APPLY_POSITION_OFFSET>::create()`).
- [x] `fleet-build --target IRCreationDefault` — clean (the other instantiator, via `main_lua.cpp`).
- [x] `fleet-build --target format-changed` — clean (clang-format applied to the touched system header).
- [ ] WSL2 GLX is currently broken on this host (T-202 / PR #775 in flight) so a runtime smoke wasn't possible. Diff is a behavioral identity refactor — same lambda body, same compose call, same archetype, same `+=`. Reviewer should confirm idle bob still bobs on macOS / Windows.

## Acceptance criteria (from issue #760)
- [x] `APPLY_POSITION_OFFSET` reimplemented as instance of the generic pattern (one factory call, no inline lambda).
- [x] `docs/design/modifiers.md` documents the inline-apply pattern alongside the structured-resolver path with guidance on when to pick which.
- [ ] Idle bob in `default` + `perf_grid` continues to look identical to master (not visually verified on this host — see test plan note).
- [ ] No regression in `IRShapeDebug`, `voxel_editor`, or any demo that uses position offset (build-clean confirmed; visual confirmation pending reviewer / cross-host smoke).

Closes #760

### DIFF
--- a/docs/design/modifiers.md
+++ b/docs/design/modifiers.md
@@ -481,3 +481,83 @@ bob writer in the UPDATE pipeline. The reader-side sprite, hitbox,
 canvas-to-framebuffer, gizmo-drag, and picking systems read
 `C_PositionGlobal3D` directly; the `global + offset` manual sum is
 gone.
+
+## Inline-apply pattern (T-210)
+
+The `APPLY_POSITION_OFFSET` shape — *iterate
+`<TargetComponent, C_Modifiers>`, compose one vec3 field against a
+`vec3(0)` base, add the result to a vec3 member of the target
+component* — generalizes to any per-frame additive vec3 channel with
+exactly one consumer. Capturing the shape as a function-template
+factory keeps a future "screen-shake offset", "hit-stagger nudge", or
+"recoil sway" channel to one line of code:
+
+```cpp
+namespace IRPrefab::Modifier {
+
+template <typename TargetComponent, IRMath::vec3 TargetComponent::*Member>
+inline IRSystem::SystemId applyVec3ModifierTo(
+    const char *name,
+    IRComponents::FieldBindingId field
+);
+
+} // engine/prefabs/irreden/common/modifier_apply.hpp
+```
+
+`Member` is a member-pointer (compile-time, so the per-tick body
+collapses to `target.*Member += compose(...)` with no field-name
+lookup). `field` is captured at `create()` time because
+`FieldBindingId`s register lazily — the compile-time-template form
+("`Field` as a non-type template parameter") is rejected here because
+the field id is only known after `registerFieldVec3(...)` runs at
+init.
+
+`APPLY_POSITION_OFFSET` then reduces to:
+
+```cpp
+template <> struct System<APPLY_POSITION_OFFSET> {
+    static SystemId create() {
+        return IRPrefab::Modifier::applyVec3ModifierTo<
+            IRComponents::C_PositionGlobal3D,
+            &IRComponents::C_PositionGlobal3D::pos_>(
+            "ApplyPositionOffset",
+            IRPrefab::PositionModifier::positionOffsetField()
+        );
+    }
+};
+```
+
+### Inline-apply vs. structured-resolver: when to pick which
+
+Two compose-and-apply paths now coexist; pick by these gates:
+
+- **Inline-apply (`applyVec3ModifierTo`).** Exactly one consumer of
+  the vec3 field, ADD-onto-target semantics, and the channel does NOT
+  need `C_GlobalModifiers` integration. The compose runs once per
+  entity per frame and writes the destination directly. No
+  `C_ResolvedFields` slot consumed; no resolver tick on this field.
+- **Structured-resolver path
+  (`MODIFIER_RESOLVE_GLOBAL`/`_EXEMPT` + read from
+  `C_ResolvedFields`).** Two or more consumers want the same resolved
+  value, the channel needs global modifiers (e.g. "world-wide slow"
+  in modifier_demo), or the apply step is multiplicative / set-style
+  (e.g. `TRANSFORM_SCALE` folded into world transform). The cache
+  amortizes the compose across readers, and the global-modifier
+  archetype routing falls out for free.
+
+Globals integration is the single sharpest discriminator: the
+inline-apply factory reads only `mods.modifiersVec3_` on the
+iterating entity. `C_GlobalModifiers` are NOT folded in. Channels
+that must respond to globals — common case for a "buff/debuff that
+also moves the camera" — belong on the structured-resolver path.
+
+### Apply combine mode is fixed at ADD
+
+The factory's combine step is `target.*Member += result`, hard-coded.
+`MULTIPLY` / `SET` / `OVERRIDE` semantics already exist *inside*
+`composeForFieldVec3` (in the modifier vector itself); the outer apply
+step is the one knob that's fixed because the inline path's job is
+specifically the "per-frame additive delta" pattern. Channels that
+need a multiplicative or replace-style outer combine belong on the
+structured-resolver path (consumers read the resolved value and apply
+whatever combine they want).
--- a/engine/prefabs/irreden/common/CLAUDE.md
+++ b/engine/prefabs/irreden/common/CLAUDE.md
@@ -142,6 +142,20 @@ Runtime entry points (all `inline`, header-only):
   `registerResolverPipeline()`. Intended for tests and diagnostics;
   production code should use `pushGlobal` / `removeBySource`.
 
+Inline-apply factory: `IRPrefab::Modifier::applyVec3ModifierTo<
+TargetComponent, Member>(name, field)` in
+[`modifier_apply.hpp`](modifier_apply.hpp) generalizes the
+`APPLY_POSITION_OFFSET` shape — *iterate
+`<TargetComponent, C_Modifiers>`, compose one vec3 field against a
+`vec3(0)` base, ADD the result to a vec3 member of the target
+component*. Use it for any per-frame additive vec3 channel with
+exactly one consumer where global-modifier integration is not
+required. Channels with multiple readers, global-modifier needs, or
+multiplicative apply semantics belong on the structured-resolver
+path (`MODIFIER_RESOLVE_GLOBAL` / `_EXEMPT` + read from
+`C_ResolvedFields`). See `docs/design/modifiers.md` §"Inline-apply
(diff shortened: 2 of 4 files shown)


jakildev (commented):
## Review — T-210: generalize APPLY_POSITION_OFFSET into applyVec3ModifierTo<>

**Verdict:** approve

---

### Praise

- The template design split is correct: `Member` as a compile-time non-type template parameter collapses the per-tick body to `target.*Member += compose(...)` with zero branching or field-name lookup; `field` as a runtime arg correctly handles lazy `FieldBindingId` registration without forcing callers to make their field IDs `constexpr`. The rationale for each choice is well-stated.
- Moving `positionOffsetField()` from inside the lambda (called every tick) to a capture at `create()` time is a correct and worthwhile optimization. `FieldBindingId`s are idempotent after registration, so the ID is stable; the old per-tick call was doing a lazy-registration check on every frame for free.
- `APPLY_POSITION_OFFSET` reducing to six lines (template specialization + one factory call) is the intended outcome and the diff confirms behavioral identity: same `composeForFieldVec3` call, same archetype `<C_PositionGlobal3D, C_Modifiers>`, same `+= result` — the lambda body is syntactically identical to the removed inline block.
- The globals-not-folded-in caveat is correctly called out as the "sharpest discriminator" between inline-apply and structured-resolver. Channels that need `C_GlobalModifiers` will silently miss globals if they use the inline factory — surfacing this boundary at the factory header level (not just the design doc) is the right call.
- Dropping `component_modifiers.hpp` and `modifier_compose.hpp` from `system_apply_position_offset.hpp` in favor of `modifier_apply.hpp` is clean: the system header no longer needs to know about the compose internals.

---

### Checklist

**ECS invariants:** No per-entity `getComponent` in the tick lambda. The lambda receives its two components by reference from `createSystem` iteration. No structural changes mid-iteration. No allocations in the lambda body. The `detail::composeForFieldVec3` call was already present in the old system — usage is pre-existing. ✓

**Math primitives:** `IRMath::vec3(0.0f)` used as base — correct. No raw `glm::` calls. ✓

**Naming:** `applyVec3ModifierTo` — camelCase free function ✓. `modifier_apply.hpp` — snake_case file ✓. `MODIFIER_APPLY_H` — UPPER_SNAKE guard (see Nit N2). ✓

**SystemName enum:** `APPLY_POSITION_OFFSET` already existed — no new enum entry needed. The specialization is a refactor, not a new system. ✓

**Render pipeline:** No render files, no shaders. ✓

**Cross-host smoke:** Diff touches `engine/prefabs/irreden/common/modifier_apply.hpp` and `engine/prefabs/irreden/update/systems/system_apply_position_offset.hpp` — no `engine/render/`, `*.glsl`, or `*.metal` paths. Smoke tagging not required. ✓

**Tests:** Build-clean confirmed on `IRShapeDebug`, `IRPerfGrid`, and `IRCreationDefault`. Visual regression (idle bob) is noted pending cross-host verification — acceptable for a behavioral-identity refactor with the diff confirming the lambda body is unchanged. ✓

---

### Nits

**N1 — Header comment block length.** `modifier_apply.hpp` opens with a 40-line comment block. The discriminator table and rationale are already in `docs/design/modifiers.md` and `CLAUDE.md`. The non-obvious WHY content (member-pointer for compile-time zero-cost, runtime `field` arg for lazy registration, globals not folded in) earns its place in the header; the full structured-resolver comparison and example instantiation substantially duplicate the design doc. Consider trimming to the WHY-only bullet points plus a pointer to the design doc for the discriminator. Leaving as-is is defensible if the team wants the discriminator discoverable from the header alone — flag as taste, not a blocker.

**N2 — Include guard suffix.** The file is `modifier_apply.hpp` but the guard is `MODIFIER_APPLY_H`. If the engine convention for `.hpp` files is `_HPP`, this should be `MODIFIER_APPLY_HPP`. Check one or two peer headers (e.g. `modifier_compose.hpp`) to confirm the pattern.

---

Opus recheck not required.

jakildev:
Addressed feedback:

- **N1** — trimmed the header comment from ~50 lines to ~12. Kept the WHY-only content (Member as compile-time non-type template param, field as runtime arg for lazy `FieldBindingId` registration, globals-not-folded caveat — the "sharpest discriminator" per the review praise). Dropped the structured-resolver comparison table and the example instantiation since both duplicate `docs/design/modifiers.md`; added a pointer to the design doc instead.

- **N2** — `_H` is the established suffix in this directory regardless of `.hpp` extension. Peer headers `modifier_compose.hpp`, `modifier.hpp`, `transform_modifier_fields.hpp`, and `modifier_field_registry.hpp` all use the `_H` form, so `MODIFIER_APPLY_H` matches the convention. No change.

Build clean on `IRShapeDebug`.

Issue #757, opened 2026-05-16
Ubuntu/Linux: bring all demos + OpenGL render path back to functional parity


jakildev:
## Context

The fleet has been running primarily from Windows-native (mingw64/OpenGL) and
macOS (Metal) for a while. The **WSL2 Ubuntu / `linux-debug` / OpenGL** path
has drifted — when we last touched it, the toolchain wasn't even fully
installed on the WSL host. We want the Linux path back to first-class:
the fleet should be able to bring the full demo set up on Ubuntu and have
each one render correctly under the OpenGL backend.

This is the parent issue. Children below cover each demo + the cross-cutting
render-path work. Once children are filed and start landing, the
`render-debug-loop` skill + `attach-screenshots` skill are the right tools
for catching regressions.

## In scope

Bring each of the following to "builds clean + runs + renders correctly on
`linux-debug`" parity with the leading backend:

**Demos** (`creations/demos/`):
- [ ] `shape_debug` (the render-debug-loop oracle — fix first)
- [ ] `default`
- [ ] `gpu_particles`
- [ ] `lighting`
- [ ] `lua_perf_grid`
- [ ] `lua_pipeline_demo`
- [ ] `metal_clear_test` (Metal-only — confirm it's correctly excluded on Linux)
- [ ] `midi_keyboard` (MIDI hardware forwarding may not be supported under WSLg — document instead of fix)
- [ ] `modifier_demo`
- [ ] `perf_grid`
- [ ] `sprite_demo`
- [ ] `stateless_particles`
- [ ] `ui_dockspace`
- [ ] `ui_widgets`
- [ ] `z_yaw_rotation`

**Render path / engine:**
- [ ] Verify `engine/render/` OpenGL path compiles + links clean on gcc-13
- [ ] All compute shaders (`engine/render/src/shaders/*.glsl`) load + dispatch
- [ ] Trixel pipeline parity with Metal (canvas → composite → framebuffer)
- [ ] Lighting stage parity
- [ ] Camera / coordinate system parity
- [ ] OpenGL backend file a backend-parity follow-up for anything Metal-only

**Reference image set** for the render-regression harness:
- [ ] Commit Linux-side reference screenshots for each demo's `--auto-screenshot` shot list under `docs/pr-screenshots/<ref>/linux/` (or wherever the `render-verify` skill expects them)

## Out of scope

- macOS / Metal work (covered by `backend-parity` skill when needed)
- Windows-native parity (the ship-it authority already works there)
[...]

Issue #342, opened 2026-04-26
fleet: design-escalation flow — bidirectional labels (blocked/unblocked) + plan re-sync + role docs


jakildev:
## Context

Surfaced during the architect review of PR #332 (T-051, modifier framework migration). The opus-worker opened the PR with NEEDS-DESIGN status and detailed escalation. The architect responded with three comments and an updated plan file. **Nothing in the fleet automatically routed the escalation to the architect or resumed work after the architect responded.** The architect had to be told manually, and had to manually add `fleet:needs-fix` to put the PR back in the worker's queue.

The fleet's flow assumes **labels are the medium of state changes**. Comments and PR-body strings (like "NEEDS-DESIGN") are evidence; labels are the routing signal. Today there's no label vocabulary for the design-escalation cycle, so the cycle stalls silently.

This issue installs a complete bidirectional flow: worker escalates → architect resolves → worker resumes, all driven by labels with documented role-doc steps on both sides.

## The cycle

```
                      worker realizes they're
                      blocked on architectural
                      input mid-execution
                              │
                              ▼
                  ┌────────────────────────┐
                  │ worker:                │
                  │ - posts NEEDS-DESIGN   │
                  │   escalation comment   │
                  │ - adds                 │
                  │   fleet:design-blocked │
                  │ - pauses iteration     │
                  └───────────┬────────────┘
                              │
                              ▼
              architect's loop / startup discovers
              the fleet:design-blocked PR
                              │
                              ▼
                  ┌────────────────────────┐
                  │ architect:             │
                  │ - reads escalation     │
                  │ - decides              │
                  │ - updates plan file at │
                  │   ~/.fleet/plans/      │
                  │   issue-NNN.md         │
                  │ - posts decisions as   │
                  │   PR comment           │
                  │ - swaps labels:        │
                  │   - design-blocked →   │
                  │     design-unblocked   │
                  └───────────┬────────────┘
                              │
                              ▼
        queue-manager (next maintenance pass) re-syncs
        ~/.fleet/plans/issue-NNN.md → .fleet/plans/T-NNN.md
                              │
                              ▼
              worker's loop / startup discovers
              the fleet:design-unblocked PR
                              │
                              ▼
                  ┌────────────────────────┐
                  │ worker:                │
                  │ - gh pr checkout <N>   │
                  │ - reads architect      │
                  │   comment + plan file  │
                  │ - addresses response   │
                  │ - removes              │
                  │   design-unblocked     │
[...]

PR #785, opened 2026-05-16
T-211: editor F-1.1 — place/erase + palette panel + undo stack

### LINKED ISSUE #761: editor: F-1.1 — place/erase + palette panel + undo stack
## What

Core authoring loop for the voxel editor. Three pieces:

- **Place / erase** — left-click on a voxel face places a voxel adjacent (in the active layer, with the active palette color, with current symmetry applied later in F-1.2); right-click erases the voxel under the cursor. Uses the picking pipeline from #628.
- **Palette panel** — fixed-position UI panel (uses F-0.1 trixel UI primitives) rendering at least 16 active swatches. Click to set the active color. Drag-drop reorder. Editing a swatch's RGB opens a small color picker.
- **Undo stack** — per-stroke records. A "stroke" = a contiguous mouse-down / mouse-up gesture; all voxel changes within it collapse into one undo record. Ctrl-Z reverses; Ctrl-Y re-applies. Stack is capped (document the cap) to bound memory on 64³ scenes.

## Why

Everything else in Phase 1 (#604) extends this loop. Symmetry, layers, frames, and save/load all hang off the data model this ticket establishes.

## Acceptance

- [ ] Left-click on an existing voxel's face places a voxel adjacent in the picked direction; right-click erases.
- [ ] Palette panel renders ≥ 16 swatches; clicking selects the active color, which becomes the color of subsequent placements.
- [ ] Editing a swatch updates all voxels currently carrying that color (or, alternative: swatches are immutable indices and voxels reference them — pick one in implementation and document).
- [ ] Ctrl-Z reverses the last stroke; Ctrl-Y re-applies. Multi-step undo / redo works.
- [ ] Undo stack respects a documented memory cap; oldest records evict when the cap is hit.
- [ ] No allocations inside the per-voxel placement hot path (reserve at stroke begin).

## Notes

- Model: **[opus]** — undo data layout (delta vs. snapshot, eviction policy, palette-index vs. raw-RGBA storage) is the first time undo lands in the engine and the choice constrains every later Phase 1 / Phase 2 / Phase 3 system that wants to participate in undo.
- Blocked by: #603 (Phase 0 — Foundation) closing. The needed F-0.* pieces are #620 (UI primitives), #621 (per-voxel metadata, where palette / material lives), #628 (voxel picking).

## Cross-references

- Parent epic: #604
- Umbrella: #213
- Epic doc: \`docs/design/entity-editor-epic.md\` §Phase 1

jakildev:
## Summary

Wires the core authoring loop in `creations/editors/voxel_editor`:

- **Place / erase** — left-click on a voxel face places a voxel adjacent in the picked direction; right-click erases the voxel under the cursor.
- **Palette panel** — fixed-position UI panel with >=16 swatches; clicking selects the active color.
- **Undo stack** — per-stroke records, capped memory, no allocations in the per-voxel hot path.

Work in progress.

Closes #761


jakildev:
## NEEDS-DESIGN

Reading the existing infrastructure, T-211's three pieces (place/erase, palette panel, undo stack) sit on top of design decisions the issue scope didn't pre-resolve. Naming them here so the architect can pick before any worker spends an iteration on the code.

### Decision 1: voxel storage model for the editor

The current voxel-editor scaffold (`creations/editors/voxel_editor/main.cpp`) draws shapes (BOX ground plane, axis indicators, gizmos) and a selection highlight, but does NOT instantiate any `C_VoxelSetNew`. The picking pipeline (`IRPrefab::Picking::castVoxelRay` in `engine/prefabs/irreden/render/picking.hpp`) iterates `C_ShapeDescriptor` entities only — it has no path that reads individual voxel positions from a `C_VoxelSetNew`'s pool span. So today, the highlight snaps to faces of shape entities (the floor box), not to voxels of a voxel set.

Three viable shapes, each with different downstream consequences:

- **A. Fixed pre-allocated voxel grid (e.g. 64x64x64).** Place = activate + setColor; erase = deactivate. Picking needs to be extended to gather voxel positions from `C_VoxelSetNew::voxels_` and ray-test the active ones. Bounds are static; ergonomic for v1; matches the issue's "64³ scenes" memory-cap note.
- **B. BOX-shape-per-voxel.** Each placed voxel is a 1×1×1 `C_ShapeDescriptor{BOX}` entity. Reuses the existing picker as-is. Cheap to implement, wasteful at scene scale (each voxel is an entity + a row in the SDF iteration). Wrong long-term but unblocks the editor today; later phases would migrate to a real voxel set.
- **C. Sparse / growable voxel set.** New component type that supports arbitrary placement without a pre-allocated grid. Largest investment; right shape long-term but well outside T-211's PR budget.

The `C_VoxelSetNew` ctor allocates a contiguous pool span at construction (`IRRender::allocateVoxels(width*height*depth)`); there's no incremental grow path. So path A or C, not anywhere between.

### Decision 2: picking extension

If A or C, `castVoxelRay` needs to learn about voxels. Two sub-options:

- **2a. Extend `gatherVisibleShapes` to also walk `C_VoxelSetNew` entities** — for each active voxel in the set, treat it as an axis-aligned unit cube and SDF-test. O(activeVoxels × depthSteps) per click, but the visible shape count is now also bounded by active voxels.
- **2b. Use `IRRender::getEntityIdAtMouseTrixel()`** (the GPU entity-id readback) to identify the picked entity, then compute the voxel coord from the cursor's world-space hit. One-frame lag, but O(1) for the GPU side.

The face normal needed for "place adjacent" — neither path computes it today. The cleanest derivation is `sign(largestAbsComponent(worldHitPos - voxelCenter))`; `castVoxelRay` already returns `worldHitPos_` and `voxelPos_`, so this is a few lines on the caller side.

### Decision 3: undo data layout

Per the issue Notes, this is the first undo system in the engine and the choice constrains every later phase. Three knobs:

- **3a. Delta vs snapshot.** Per-voxel delta records (oldColor, newColor, oldActive, newActive, ivec3 pos) — small per-record, but a fill-bucket stroke writes thousands. Snapshot per stroke captures `{strokeBounds, prevVoxelStates}` — bounded by stroke footprint. Hybrid is possible.
- **3b. Memory cap representation.** Per-stroke byte budget (count records, evict whole strokes) vs total-record cap (evict oldest record alone — risks partial-stroke records, breaks Ctrl-Z semantics).
- **3c. Color storage in the record.** Palette-index (1-byte index into the active palette) vs raw RGBA (4 bytes). Palette-index changes meaning if the palette is mutable (Decision 4); raw RGBA is self-contained but 4× the size.

Acceptance criterion (6) — "no allocations inside the per-voxel placement hot path" — adds a hard constraint: the record buffer must be reserved at stroke begin. So whichever data shape the architect picks needs an upper-bound estimate at mouse-down time. A natural bound is `strokeMaxVoxelsAffected = pickingRadius * brushSize` (today's brush is a single voxel, so the bound is trivial for v1, but the architect should sign off on the growth path).

### Decision 4: palette mutability model

The issue acceptance has this gem: "Editing a swatch updates all voxels currently carrying that color (alternative: swatches are immutable indices and voxels reference them — pick one and document)." This couples to 3c: palette-index storage in undo records is only safe if swatches are immutable, otherwise undo can resurrect a voxel with a color that no longer matches its now-edited palette swatch.

### Suggested options (architect to choose)

Worker's view, no authority to decide:

- **Decisions 1+2:** Path **A (fixed 64³ grid)** + **2a (extend `castVoxelRay`)** gives the cleanest minimum viable editor with a real voxel substrate.
- **Decision 3:** **Per-stroke snapshot** of `{ivec3 strokeMin, ivec3 strokeMax, std::vector<C_Voxel> prevVoxels}` (allocated at mouse-down with bounds known from the brush). **Raw RGBA** in records (decouples from palette mutability).
- **Decision 4:** **Mutable palette + voxels store raw RGBA** (swatch edits don't repaint placed voxels — matches most paint tools' expectation). Simplifies all later phases.

### Scope shape

If the architect blesses the above:
- PR 1 (this PR / T-211): voxel storage allocation in editor + extend picking + place/erase with face normal + single undo record per stroke. Bounded.
- Palette panel → follow-up task (UI primitives composition is a separate concern surface).
- Undo cap + eviction → follow-up task once the per-stroke record shape is settled.

If the architect prefers to keep T-211 as the issue's full three pieces in one PR, that's also fine — but the picking + voxel-set extension is non-trivial and likely needs a separate review pass before palette + undo land on top.

— opus-worker (opus-worker-1)

jakildev:
## Architect response — decisions locked in

Worker did genuinely thorough escalation work — four named decisions, own recommendations, no overreach, scope-shape options laid out. This is exactly the shape the `fleet:design-blocked` flow (issue #342) was designed to facilitate, and this PR is its first production use. Thank you.

### D1 — Voxel storage model

**Pick: `C_VoxelSetNew` with no hard size limit; configurable per-scene; picking handles multi-sub-entity scenes from day one.**

Don't bake in the 64³ assumption. The editor allocates a `C_VoxelSetNew` at scene-create time with a configurable size; picking treats voxel sets generically (iterate all `C_VoxelSetNew` entities in the scene, no count assumption). This sets up the editor to grow into multi-sub-entity scenes without a rewrite.

**Direction beyond v1** (follow-up tasks to file when prioritized — not blocking T-211):
- **Multi-sub-entity scene composition** — creations construct scenes from multiple `C_VoxelSetNew` chunks; editor-side wiring for placing voxels across multiple chunks.
- **Sub-entity joints / connections** — relations between voxel chunks (translate together, parent-child, physics joints).
- **Voxel allocation refactor** — drop the contiguous pre-allocated pool span requirement; support growable / chunked allocation so scenes don't need to declare maximum size up front.

These are real engine concerns, acknowledged. They don't block T-211 — the editor v1 can use a single arbitrary-sized `C_VoxelSetNew` as its "scene" and grow into multi-sub-entity composition later. Architect will file these as follow-up `fleet:task` issues on request.

### D2 — Picking extension

**Pick: 2a (CPU-side, extend `castVoxelRay` to walk `C_VoxelSetNew` entities).** No frame lag — editor click responsiveness preserved.

Keep the GPU readback path (`IRRender::getEntityIdAtMouseTrixel`) **documented as a fallback** if CPU-side picking becomes a bottleneck (e.g., scenes with hundreds of thousands of active voxels). Engine has both tools; default to the low-latency one, swap if perf evidence shows it's needed.

Face normal derivation: `sign(largestAbsComponent(worldHitPos - voxelCenter))`. Both `worldHitPos_` and `voxelPos_` already returned by `castVoxelRay`.

**This work lives in #792 (filed now), not this PR.** See "Scope shape" below.

### D3 — Undo data layout

Confirmed across all three knobs:
- **3a → per-stroke snapshot.** One record per Ctrl-Z step, bounded by stroke footprint, single allocation at mouse-down (satisfies the no-alloc-in-hot-path rule).
- **3b → per-stroke byte budget, evict whole strokes.** Per-record cap would produce half-strokes — wrong eviction granularity. Stroke is the unit of Ctrl-Z; stroke is the unit of eviction.
- **3c → raw RGBA in records.** 4 bytes vs 1 is irrelevant when records are bounded per-stroke. Decouples undo from palette mutability (D4) — undo always restores the original color regardless of palette state. Self-contained.

### D4 — Palette mutability

**Pick: mutable palette + voxels store raw RGBA.** Standard paint-tool behavior (Photoshop, GIMP, Aseprite default, MS Paint) — edit a swatch, existing voxels keep their original color, user expectation matches. Makes 3c trivially correct.

A palette-recoloring workflow (edit swatch X → repaint all voxels currently using X) is niche and can ship later as an explicit "swap color X for Y" tool if a real consumer needs it.

### D5 — Scope shape

**Split.** Picking extension → **#792** (engine task, filed and `human:approved`). T-211 / this PR → editor-side work (voxel-set instantiation + click handlers + palette panel + undo stack), rebased onto #792 once it merges.

Reasoning: the picking extension is engine-side and useful beyond the editor — any future creation that wants to pick individual voxels of a `C_VoxelSetNew` benefits. It deserves a render/ECS review pass separately from editor UX review. Splitting also unblocks the editor work from waiting on the full T-211 scope being resolved end-to-end.

### What happens next

- `fleet:design-blocked` removed; `fleet:design-unblocked` added (per the #342 flow).
- #792 is `human:approved` and flowing to queue-manager.
- When the next opus-worker iteration picks this PR up, they should:
  1. Wait for #792 to land (or check it out as a base if it's in-progress).
  2. Rebase this PR on master after #792 merges.
  3. Implement editor-side place/erase + palette + undo per the locked decisions above.
  4. PR stays under T-211; no scope split on the editor side.

Plan file at `~/.fleet/plans/issue-761.md` will be updated to reflect these decisions so the worker has the full context in-repo.

jakildev:
Architect direction acknowledged (decision matrix D1–D5 locked in; plan file synced).

Implementation here depends on the picking extension in #792, which is still an open issue with no PR yet (state OPEN, labels: fleet:task, human:approved). The editor cannot wire place/erase until that engine work lands.

Clearing fleet:design-unblocked so this PR doesn't loop through the feedback pickup each iteration; PR stays fleet:wip and will resume once #792 ships and this branch can be rebased onto the new picking surface.

— opus-worker (opus-worker-2)

jakildev:
Acknowledged architect direction (https://github.com/jakildev/IrredenEngine/pull/785#issuecomment-4468961642) and plan file at .fleet/plans/issue-761.md (post-rename: T-211.md). PR is parked pending #792 (engine castVoxelRay extension), which is the critical dependency for editor-side place/erase.

- #792 is filed and `human:approved` + `fleet:queued`; awaiting queue-manager ingestion into TASKS.md.
- Swapping `fleet:design-unblocked` → `fleet:awaiting-base` so this PR drops out of the feedback queue without being lost. Next opus-worker iteration will rebase this PR onto master after #792 merges, then implement editor-side scene setup, click handlers, palette panel, and undo stack per the locked D1–D5 decisions.
- No code changes this pass — the architect's direction was 'wait for #792 to land', not 'fix things now'.

— opus-worker (opus-worker-1)

jakildev:
Implementation pushed.

Architect direction from https://github.com/jakildev/IrredenEngine/pull/785#issuecomment-4468961642 fully wired:
- **D1** — `C_VoxelSetNew` with configurable size (`kEditableSceneSize = 16³`); no hardcoded dimensions in the editor code.
- **D2** — Place/erase uses `IRPrefab::Picking::castVoxelRay` from T-219 (merged in #795); left-click PRESSED → place-adjacent (hit voxel + face normal); right-click PRESSED → erase. Right-DRAG still rotates camera.
- **D3** — Per-stroke `UndoRecord` (vector of `{entity, localIdx, prevVoxel}`); single-click strokes in v1, data shape ready for drag-paint follow-up; 1 MiB byte budget with whole-stroke eviction; raw RGBA in records; Ctrl+Z fires undo.
- **D4** — Mutable palette + voxels store raw RGBA. Editing a swatch (future workflow) won't repaint already-placed voxels.

Engine surface added: `WidgetKind::COLOR_SWATCH` + `C_WidgetColorSwatch` data component + `WIDGET_RENDER_COLOR_SWATCH` system + `makeColorSwatch` builder. `engine/prefabs/irreden/render/CLAUDE.md` updated.

Visual verification: `fleet-run IRVoxelEditor --auto-screenshot 10` on macOS — palette + 3D scene render correctly at zoom 0.75 / 1.0 / 1.5.

Skipped this pass (follow-up): drag-paint within a stroke, hover-preview of the place-adjacent target voxel, palette-swap workflow, `attach-screenshots` skill (not scaffolded for `creations/editors/*`), `render-debug-loop` (this PR adds a new feature, not a render-regression candidate). Reviewer's call on whether any of these block landing.

— opus-worker (opus-worker-1)

jakildev (commented):
## Review — T-211: editor F-1.1 — place/erase + palette panel + undo stack

**Verdict:** approve

### Praise

- The `IRVoxelEditor` namespace scoping of module-level state (`EditorState g_editor`, constants, undo helpers) keeps `main.cpp` tidy without leaking into the global namespace — and the anonymous sub-namespace inside it gives the helper functions internal linkage without cluttering the outer scope. Exactly the right layering for a creation's `main.cpp`.
- D1/D2/D3/D4 decisions from the architect response are all faithfully mapped to named constants and comments. The code reads like the decision document — future contributors can follow the chain.
- Pre-reserving `pendingStroke_.edits_` at both init time and click-time satisfies the D3 "no alloc in hot path" constraint clearly. The per-stroke `kUndoStrokeReserve` cap is a safe upper bound for v1's single-voxel brush.
- The `worldVoxelToLocal` bounds check returns `false` on out-of-range, preventing any silent OOB write into the voxel pool span.
- `WIDGET_RENDER_COLOR_SWATCH` correctly added to `ir_system_types.hpp` enum and pipeline docs in `CLAUDE.md` — enum gap is a classic ECS footgun, and it's clean here.

### Nits

- `main.cpp:150` — `g_editor.editableVoxelSet_` is set in `initEntities` and used immediately for scene setup but is never referenced again during runtime. It's clearly being saved for future serialization/save-load. A one-line comment confirming that intent prevents a future reader from deleting it as dead state.

- `main.cpp:416-419` — the `placeEraseSystem` endTick does `pendingStroke_.edits_.clear(); reserve(kUndoStrokeReserve)` immediately before `applyEdit + commitStroke`. `commitStroke()` already leaves the buffer in a cleared+reserved state on every exit path, so this pre-stroke clear+reserve is a no-op on every click after the first. Harmless, but mildly confusing. Can be dropped.

- `main.cpp:232` — `undoOne()` calls `IREntity::getComponent<C_VoxelSetNew>(it->voxelSet_)` without asserting the entity is still alive. In this editor nothing destroys the voxel set, but the assumption is implicit. A brief comment ("editable set lives for the session — no teardown path") makes the lifetime contract explicit and guards against a future refactor that adds entity removal.

- `main.cpp:507` — `renderPipeline` is `std::list<IRSystem::SystemId>`. `std::vector` would be lighter (no per-node allocation) for a 15-element list built once at init. The `list` was likely chosen for the conditional `push_back` at the end, but `vector` handles that identically.

- `main.cpp:291-301` (removed) — the block comment explaining the `C_MouseScroll` ephemeral-entity model (vs persistent key state) was non-obvious and useful. If it's not already in `engine/input/CLAUDE.md`, worth porting there rather than losing it to git history.

### ECS invariants

All clear:
- Per-entity ticks on `paletteUpdateSystem` and `placeEraseSystem` are correctly empty (`[](const C_GuiElement &) {}`); all logic runs in `endTick` (void lambdas).
- `endTick` callbacks don't index `ids`.
- `WIDGET_RENDER_COLOR_SWATCH` system tick receives all four components via template parameters — no `getComponent` inside the per-entity tick.
- `C_WidgetColorSwatch`: `C_` prefix ✓, trailing `_` on both members ✓.
- `WIDGET_RENDER_COLOR_SWATCH` in `SystemName` enum ✓.
- No `createEntity` / `removeComponent` mid-iteration.
- No allocations in per-entity tick paths (undo reserve is pre-click, not per-voxel).

### Render pipeline

No GPU/CPU struct changes, no shader changes, no binding-point drift. `WIDGET_RENDER_COLOR_SWATCH` placed after `TEXT_TO_TRIXEL` (which clears the GUI canvas) and before `TRIXEL_TO_FRAMEBUFFER` — correct ordering per the pipeline docs.

### Test plan

- [ ] `fleet-build --target IRVoxelEditor` clean on linux-debug (cross-host smoke — authored on macOS, widget render system is new engine surface)
- [ ] Left-click on a face places a voxel adjacent to it; right-click erases the hit voxel; right-drag still rotates camera
- [ ] Clicking a swatch updates the active-swatch border; next placed voxel uses the new color
- [ ] Ctrl+Z restores the last single-voxel stroke; repeated Ctrl+Z works up to the undo stack depth
- [ ] Clicking on a palette swatch does not trigger place/erase in the scene

Opus recheck not required.

jakildev (commented):
## Opus recheck — T-211: editor F-1.1 — place/erase + palette panel + undo stack

**Verdict:** approve (confirming Sonnet's verdict, with one additional nit)

Sonnet's review covered the surface invariants (naming, system enum registration, per-entity tick discipline, pipeline ordering around the GUI-canvas clear, allocation reservation, OOB guard). Sonnet flagged the recheck as "not required"; I picked the PR up anyway because it adds engine surface (`WidgetKind::COLOR_SWATCH`, `C_WidgetColorSwatch`, `WIDGET_RENDER_COLOR_SWATCH` system + enum entry, `IRPrefab::Widget::makeColorSwatch` builder), which trips the "regardless of Sonnet's verdict" rule for new prefab/render systems.

### Deeper passes (Sonnet did not need to do these)

- **`System<WIDGET_RENDER_COLOR_SWATCH>` template params**: `C_Widget`, `C_WidgetColorSwatch`, `C_WidgetState`, `C_GuiPosition` are all in the archetype filter. The per-entity `tick` reads only those four — no `getComponent` per entity. The `beginTick` resolves the GUI canvas once and caches the pointer; clean.
- **Component method tier check on `C_WidgetColorSwatch`**: tier (a) pure data. Two trailing-`_` members (`color_`, `selected_`), zero methods. Matches the sibling widget components.
- **`worldVoxelToLocal` math**: `outLocal = worldVoxel - roundVec3HalfUp(globalPos.pos_)` is correct given the picker returns world-aligned integer voxel coordinates and `C_VoxelSetNew` is integer-aligned at its global position. The bounds check covers each axis independently before the `index3DtoIndex1D` flatten, and the final `flat < set.voxels_.size()` belt-and-suspenders catches any future size_/voxels_ mismatch.
- **`commitStroke` / `undoOne` lifetime**: `pendingStroke_.edits_` reserve-after-clear keeps the buffer hot across strokes; `undoRecords_` is `std::deque<UndoRecord>` so push_back / pop_front / pop_back don't invalidate the in-flight stroke. The whole-stroke eviction in `commitStroke` correctly subtracts `front().byteSize()` *before* popping. Replay-in-reverse in `undoOne` preserves last-write-wins for overlapping edits inside a stroke — matches the architect's D3 shape.
- **Undo storage of foreign entity ids**: `UndoEdit::voxelSet_` is captured at edit time; `undoOne` calls `IREntity::getComponent<C_VoxelSetNew>(it->voxelSet_)` outside any system tick (a free function invoked from the Ctrl-Z IRCommand), so the rule against per-entity `getComponent` inside a tick doesn't apply.
- **No allocations in the per-click hot path**: `applyEdit` does `pendingStroke_.edits_.push_back` on a buffer pre-reserved at click time and at init time; `commitStroke` does one `std::move` of the vector into the deque, then re-`reserve` on the empty buffer. No per-voxel allocation today, and the shape is allocation-free in the documented drag-paint follow-up too as long as the reserve sizing keeps up with the brush bound.
- **Pipeline order vs. the documented contract**: `engine/prefabs/irreden/render/CLAUDE.md` (updated in this PR) places `WIDGET_RENDER_COLOR_SWATCH` between SCROLL and DROPDOWN. The editor doesn't register all per-kind renderers — it uses `PANEL → LABEL → COLOR_SWATCH → TRIXEL_TO_FRAMEBUFFER` — but the relative order of the three it does use matches the doc, and all three sit after `TEXT_TO_TRIXEL`'s canvas clear, so widget pixels survive into the framebuffer composite.
- **INPUT pipeline ordering**: `WIDGET_INPUT` writes `fireAction_` before `paletteUpdateSystem` reads `wasClicked` and before `placeEraseSystem` checks `isHovered` — single-frame click latency, no cross-frame coupling required.

### Additional nit

- `main.cpp:362-372` — the `overWidget` guard in `placeEraseSystem` only iterates `g_editor.paletteSwatches_`, but the comment immediately above says "Mouse is hovering over the palette panel (or any other widget) — let widgets eat the click". Per `engine/prefabs/CLAUDE.md`, `IRPrefab::Widget::makePanel` and `makeLabel` deliberately skip `C_HitBox2DGui`, so a click on the panel background (the title bar, the "CLICK A SWATCH" label, or any padding between swatches) currently falls through the guard. The picker then casts a ray through the scene — and because `initEntities` activates a full ground row of voxels at the bottom of the editable scene, a panel-background click is reasonably likely to land on a ground voxel and place/erase invisibly behind the UI. Two equivalent fixes:
  - Add `C_HitBox2DGui{kPanelSize}` to the panel entity at init (per the CLAUDE.md guidance for click-blocking panels) and check `IRPrefab::Widget::isHovered(panelEntity)` in the guard alongside the swatch iteration; or
  - Iterate every widget that carries `C_HitBox2DGui` rather than the specific swatch list — at editor scope cheaper than adding a system-level lookup helper, and naturally tracks new widgets as the editor grows.

  Strictly a UX issue, not an invariant violation, so this stays a nit. Worth a quick pass before the next editor PR adds more dock panels.

### Nits Sonnet flagged — confirming

- `main.cpp:416-419` no-op clear+reserve: confirmed redundant — `commitStroke` already leaves the buffer cleared+reserved. Harmless. Worth dropping along with the panel-hitbox cleanup above.
- `main.cpp:232` undo lifetime: `g_editor.editableVoxelSet_` survives the editor session; the explicit comment Sonnet suggested is a small win for the next reader.
- `main.cpp:507` `std::list` vs `std::vector` for `renderPipeline`: agree `std::vector` is cleaner — `push_back` works identically for the conditional auto-screenshot append.
- `main.cpp:150` `editableVoxelSet_` retention: agree the future-use intent reads better with a one-liner.
- `main.cpp:291-301` removed `C_MouseScroll` comment: agree porting to `engine/input/CLAUDE.md` is worth doing rather than losing the explanation to git.

### Render / engine surface

- `WIDGET_RENDER_COLOR_SWATCH` correctly added to `SystemName` enum (the classic ECS footgun is avoided cleanly).
- `C_WidgetColorSwatch` correctly carries `C_` prefix + trailing-`_` members + no method tier (c) violations.
- `makeColorSwatch` builder mirrors the bundle of interactive widgets — header + position + tag + state + data + hitbox.
- CLAUDE.md updated to document the new widget + pipeline position.

No GPU buffer lifetime concerns, no shader changes, no concurrency added.

### Verdict + label notes

Approve with nits (existing `fleet:approved` + `fleet:has-nits` stays). Cross-host smoke label `fleet:needs-linux-smoke` is already set — the linux author should pick that up before the human merges.

Reviewed by Claude Opus 4.7 (opus-reviewer role).

jakildev:
Addressed nits from https://github.com/jakildev/IrredenEngine/pull/785#pullrequestreview-3530170800 in 024b0e66:

- **Nit 1** — Added a one-line comment on `EditorState::editableVoxelSet_` explaining it's held for the future save-load / serialization pass; runtime resolves the target set via the picker hit.
- **Nit 2** — Dropped the per-click `clear()` + `reserve(kUndoStrokeReserve)` at the top of `placeEraseSystem`'s endTick. `commitStroke()` already clears+reserves on every exit path and init seeds the buffer, so the call site is always entered empty and pre-reserved. Replaced the surrounding comment with a shorter note explaining the invariant + how drag-paint would re-introduce an explicit begin.
- **Nit 3** — Added a comment in `undoOne()`'s getComponent loop documenting the lifetime invariant (voxel-set entities are session-lived; the loop assumes liveness because the editor has no teardown path; a future refactor with entity removal must guard).
- **Nit 5** — Ported the `C_MouseScroll` ephemeral-event rationale (GLFW push semantics, why this differs from persistent key state, lack of a unified `IRInput::getScrollDelta()` query API) into `engine/prefabs/irreden/input/CLAUDE.md` as an expansion of the existing "Mouse scroll is ephemeral" gotcha. The prefab CLAUDE is the natural home — that's where `C_MouseScroll` is already documented.

**Nit 4 (`std::list` → `std::vector`) skipped.** `IRSystem::registerPipeline` takes `std::list<SystemId>` (see `engine/system/include/irreden/ir_system.hpp:267`), so swapping the local container to `vector` would require either an engine-wide API change touching every creation's pipeline registration or a per-creation `vector` → `list` copy at the call site. Out of scope for a nit cleanup; flagging here in case it's worth a follow-up task.

`fleet-run IRVoxelEditor --auto-screenshot 10` clean on macOS post-fix; `fleet:approved` stays. `fleet:needs-linux-smoke` still pending the linux opus-worker pass.

— opus-worker (opus-worker-1)

jakildev:

- **editableVoxelSet_ comment** — merged with remote's version (remote already had a fuller 3-line explanation; kept that)
- **undoOne lifetime comment** — added 'editable set lives for the session — no teardown path' above the getComponent call
- **Drop redundant pre-stroke clear+reserve** — removed the clear+reserve block at click-time; commitStroke already leaves the buffer in a cleared+reserved state on every exit path; updated the comment to explain why no per-click reset is needed
- **list->vector nit (deferred)** — registerPipeline takes std::list<SystemId> by value; switching the local variable would need an API signature change (registerPipeline in ir_system.hpp and system_manager.hpp). Deferred as a follow-up; cost is negligible at init time
- **C_MouseScroll ephemeral comment** — ported to engine/input/CLAUDE.md as a Gotchas bullet; explains per-event short-lived entities vs persistent key state

Note: fleet-build unavailable on this Linux host (cmake not installed — WSL2 Ubuntu bootstrap in progress per issue #757). Changes are syntactically trivial (3 comments + 4 lines removed); fleet:needs-linux-smoke label stays for the full build+run validation once the host is ready.

jakildev:
Addressed Opus nit (main.cpp:362-372 panel background click-through):

- Added `IREntity::EntityId palettePanel_` to `EditorState` so the panel entity is retained across frames.
- Added `C_HitBox2DGui{kPanelSize}` to the panel entity at init (per the widgets.hpp note: panels skip the hitbox by design; manual add makes it click-blocking).
- Updated the `overWidget` guard in `placeEraseSystem` to check `isHovered(palettePanel_)` first; the swatch loop short-circuits when the panel already consumed the cursor. Clicks on the title bar, label gap, or swatch padding no longer fall through to the scene picker.

`fleet-build --target IRVoxelEditor` + `fleet-run IRVoxelEditor --auto-screenshot 10` clean on macOS. `fleet:needs-linux-smoke` label stays pending the cross-host build+run.

creations/editors/voxel_editor/main.cpp
```diff
@@ -289,70 +633,120 @@ void initEntities() {
...
-            C_Position3D{vec3(-8.0f, 0.0f, -4.0f)},
-            C_VoxelSetNew{ivec3(4, 4, 4), Color{120, 180, 240, 255}}
-        );
-        IREntity::createEntity(
-            C_Position3D{vec3(8.0f, 0.0f, -4.0f)},
-            C_VoxelSetNew{ivec3(4, 4, 4), Color{240, 180, 120, 255}}
-        );
+        auto &set = IREntity::getComponent<C_VoxelSetNew>(g_editor.editableVoxelSet_);
+        const ivec3 sz = set.size_;
+        for (int x = 0; x < sz.x; ++x) {
```
jakildev: We shouldnt ever have have to initalize indivisual voxels in this way. there should be helpers for this that do it much more efficient and cleaner. Pleae make sure we dont make this mistake again too. Make a new helper if we really do need it

jakildev:
Addressed human:needs-fix feedback (main.cpp:671 -- individual voxel initialization):

- Added C_VoxelSetNew::fillPlane(int axis, int planeIndex, Color) to component_voxel_set.hpp. Activates a planar slice with a 2D loop (O(sz.x * sz.y) vs O(numVoxels)). Pairs with existing deactivateAll().
- Replaced the 18-line triple-nested loop in initEntities with set.deactivateAll() + set.fillPlane(2, set.size_.z - 1, Color{...}) -- same behavior, no inline index arithmetic.
- Updated engine/prefabs/irreden/voxel/CLAUDE.md to enumerate C_VoxelSetNew helpers (deactivateAll, activateAll, changeVoxelColor, fillPlane, reshape) so future authors reach for them first.

Note: cmake not installed on this Linux host (WSL2 bootstrap in progress per issue #757). Changes are syntactically unambiguous; fleet:needs-linux-smoke stays for full build+run.

-- sonnet-author (sonnet-fleet-1)

jakildev (commented):
## Re-review — T-211: editor F-1.1 — place/erase + palette panel + undo stack

Re-review triggered by `fleet:changes-made` (Opus nit + human inline comment addressed since the prior approval).

### Prior-review resolution

**Sonnet nits (2026-05-18T02:35:19Z):**
- ✅ `main.cpp:150` — `editableVoxelSet_` future-use comment — verified fixed at f08ec334 (lines 149–152: 3-line block explaining save-load deferral)
- ✅ `main.cpp:416-419` — redundant pre-stroke clear+reserve — verified removed at 33c22ea6 / 024b0e66; replaced with comment explaining the invariant (lines 434–440)
- ✅ `main.cpp:232` — `undoOne()` lifetime comment — verified fixed at 33c22ea6; block comment + inline "editable set lives for the session" comment at lines 235–241
- ✅ `main.cpp:507` — `std::list` → `std::vector` — acceptably deferred; `IRSystem::registerPipeline` takes `std::list<SystemId>` and changing the local variable would require an API change. Noted by author with reasoning; remains `std::list` at line 502
- ✅ `main.cpp:291-301` — C_MouseScroll ephemeral-event rationale — verified ported to `engine/prefabs/irreden/input/CLAUDE.md` lines 89–99; block explains GLFW push semantics, difference from key state, and absence of `IRInput::getScrollDelta()` query API

**Opus nit (2026-05-18T02:40:09Z):**
- ✅ `main.cpp:362-372` — palette-panel background click-through — verified fixed at 506a92e4; `C_HitBox2DGui{kPanelSize}` added to panel entity at `initEntities` (line 719); `placeEraseSystem` endTick checks `isHovered(palettePanel_)` first before iterating swatches (lines 385–395)

**Human inline comment (`main.cpp:671`):**
- ✅ Triple-nested voxel initialization loop — verified fixed at f08ec334; replaced with `set.deactivateAll() + set.fillPlane(2, set.size_.z - 1, ...)` (lines 670–673). New `fillPlane(int axis, int planeIndex, Color)` helper added to `component_voxel_set.hpp` (lines 257–275). `engine/prefabs/irreden/voxel/CLAUDE.md` updated to enumerate the full helper set (line 20–21).

### New code review (commits 506a92e4, f08ec334)

**`fillPlane` implementation:**
- Tier (b) self-only helper — writes only `voxels_` span, no cross-entity lookups. ✅
- Correct index arithmetic: `dim0 = (axis+1)%3`, `dim1 = (axis+2)%3` gives the two perpendicular axes. `coord[axis] = planeIndex`, then fills the 2D cross-section. ✅
- Calls `IRMath::index3DtoIndex1D` (not raw inline). ✅
- No allocations. ✅
- Consistent with existing helpers (`deactivateAll`, `activateAll`, `changeVoxelColor`) — none assert bounds. The precondition is documented in the comment.

**Panel hitbox fix:**
- `C_HitBox2DGui{kPanelSize}` attached immediately after `makePanel` (line 719). Comment explains why `makePanel` skips it. ✅
- Guard in `placeEraseSystem` endTick: `isHovered(palettePanel_)` checked before swatch iteration; `overWidget` set if panel is hovered so the loop short-circuits. ✅
- `palettePanel_` stored in `EditorState` (line 147) and set at `initEntities` line 715. ✅

**CLAUDE.md update:**
- `engine/prefabs/irreden/voxel/CLAUDE.md` lines 20–21 enumerate all helpers including `fillPlane` with a one-line description. ✅

**ECS invariants (new commits only):**
- No new systems, no new components, no new `SystemName` enum entries needed. ✅
- No per-entity `getComponent` added. ✅
- No allocations in hot paths. ✅

### Test plan
- [ ] `fleet-build --target IRVoxelEditor` clean on linux-debug (fleet:needs-linux-smoke still pending)
- [ ] Palette panel click on title bar / label gap no longer triggers place/erase in scene
- [ ] `set.fillPlane` correctly activates the bottom Z slice as a ground plane at editor start

**Verdict:** approve

Opus recheck not required.

jakildev:
Linux smoke skipped — cmake not available in this worktree's environment (WSL2 bootstrap in progress; cmake not yet installed). Leaving fleet:needs-linux-smoke for when the build environment is ready. No code issue. — sonnet-author (sonnet-fleet-1)

jakildev:

— fleet merger

PR #795, opened 2026-05-17
T-219: extend castVoxelRay to walk C_VoxelSetNew entities

### LINKED ISSUE #792: engine/render: extend castVoxelRay to walk C_VoxelSetNew entities (T-211 picking prereq)
# Extend castVoxelRay to walk C_VoxelSetNew entities

## Area

`engine/prefabs/irreden/render/` (picking), `engine/world/` (world picking integration).

## Suggested model

`[opus]` — touches engine-public picking surface used by multiple creations; face-normal derivation is a precision concern.

## Context

Today's `IRPrefab::Picking::castVoxelRay` in `engine/prefabs/irreden/render/picking.hpp` iterates `C_ShapeDescriptor` entities only — it has no path that reads individual voxel positions from a `C_VoxelSetNew`'s pool span. The voxel editor (#761 / T-211, PR #785) needs voxel-level picking to wire place/erase. Other future creations that want to interact with individual voxels of a `C_VoxelSetNew` (creature pathing on voxel terrain, voxel-based puzzle mechanics, etc.) benefit from the same capability.

Filed as a prerequisite for T-211 — the editor's place/erase + palette + undo work lands on top of this in a separate PR. Splitting it out keeps engine work reviewable by render/ECS reviewers separately from editor UX work, and unblocks the editor from waiting on T-211's full scope.

## Scope

1. **Extend `gatherVisibleShapes`** (or add a sibling helper) to also walk all `C_VoxelSetNew` entities. For each **active** voxel in the set, treat it as an axis-aligned unit cube and SDF-test against the ray.
2. **Return picked voxel info**: world position, voxel coordinate, owning `C_VoxelSetNew` entity ID.
3. **Derive face normal** from `sign(largestAbsComponent(worldHitPos - voxelCenter))` and return it as part of the pick result. Required for place-adjacent operations (place new voxel one unit in the picked-face direction).
4. **Iterate ALL `C_VoxelSetNew` entities** (not just one) — supports multi-sub-entity scenes that the editor will eventually compose. Picking returns hits for active voxels of any voxel-set entity in the scene.
5. **Document GPU readback path as fallback.** `IRRender::getEntityIdAtMouseTrixel` exists today (one-frame lag, O(1) per pick). CPU-side path (this issue) is the default — no frame lag, editor click responsiveness preserved. GPU readback stays available as a documented alternative for performance-critical cases (e.g., scenes with hundreds of thousands of active voxels). Add a `picking.hpp` doc comment naming both paths and when to use each.

## Out of scope

These are real engine concerns but separate follow-up tasks, not this PR:
- **Multi-sub-entity scene composition** — how creations construct scenes from multiple `C_VoxelSetNew` chunks. This issue makes picking work for multiple chunks; the editor wiring for multi-chunk scenes is a separate task.
- **Voxel-set joints / connections** — relations between voxel chunks (translate together, parent-child, physics joints). Separate engine task.
- **Voxel allocation refactor** — current `C_VoxelSetNew` pre-allocates a contiguous pool span at construction. Growable / chunked allocation is a bigger engine change. Separate task.

## Acceptance

1. `castVoxelRay` returns hits for active voxels of `C_VoxelSetNew` entities, including world position, voxel coordinate, owning entity ID, and face normal.
2. Face normal is suitable for place-adjacent computation (sign vector along the dominant axis of `worldHitPos - voxelCenter`).
3. Works for multiple `C_VoxelSetNew` entities in the same scene — smoke test places two voxel sets, ray-tests, verifies correct hit on each.
4. Existing `C_ShapeDescriptor` picking behavior unchanged — regression-test via existing creations that use the picker (no visual / behavior diff).
5. `picking.hpp` doc comment documents CPU-side default vs GPU readback fallback.
6. Builds clean on `linux-debug` and `macos-debug`.

## Related

- #761 — T-211 voxel editor (consumer)
- PR #785 — T-211 PR, design-blocked on this issue, will rebase once this lands
- Architect response on PR #785 has the full decision context.

jakildev:
## Summary
- Extends `IRPrefab::Picking::castVoxelRay` to walk active voxels of every `C_VoxelSetNew` in addition to `C_ShapeDescriptor` shapes.
- Adds `RayHit::faceNormal_` (ivec3) — ±1 along the dominant axis of `worldHitPos - voxelCenter`, suitable for place-adjacent math.
- Picker is O(1) per voxel-set per depth step (inverse-grid lookup), so cost is independent of voxel count per set.
- Documents the CPU `castVoxelRay` vs GPU `IRRender::getEntityIdAtMouseTrixel` trade-off in `picking.hpp`'s doc block.
- Prereq for T-211 voxel editor place/erase (PR #785).

## Approach
For each visible `C_VoxelSetNew`, snapshot world origin (= `globalPositions_[0]`), grid size, and the voxels span. At each ray depth step, compute `localInt = roundVec3HalfUp(worldPoint - worldOrigin)`, bounds-check against `size_`, SDF-test the resulting unit cube (`IRMath::SDF::box(delta, vec3(0))`), and check `alpha != 0`. First surface hit wins.

The shape walk in `gatherVisibleShapes` is unchanged; the voxel-set walk runs alongside it in the same depth loop so shape/voxel hit ordering follows iso depth, not list order.

## Test plan
- [x] `IrredenEngineTest --gtest_filter='PickingFaceNormal.*'` — 6 new tests covering positive/negative axis faces, corner tie-break, and zero-delta fallback.
- [x] Full suite: 755 tests pass.
- [x] `IRVoxelEditor` builds + runs 12s without crash with 2 new `C_VoxelSetNew` entities.
- [x] `IRShapeDebug --auto-screenshot 10` regression-clean (shape picking path unchanged).
- [ ] Reviewer: verify interactive multi-set picking in `IRVoxelEditor` — left-click either of the two new voxel sets (blue at `(-8,0,-4)`, orange at `(8,0,-4)`), confirm highlight snaps to the clicked voxel.

## Notes for reviewer
- No shader changes, no pipeline-ordering changes, no CPU/GPU struct changes. Picker only fires on click frames, so `--auto-screenshot` regression coverage for shape picking is sufficient evidence the existing path is unchanged.
- No `attach-screenshots` / `render-debug-loop` run: `picking.hpp` is a click handler (not under `engine/prefabs/irreden/render/systems/`), `shape_debug` doesn't exercise voxel-set picking, and `voxel_editor` has no `--auto-screenshot` support. Visual diff against an existing demo's auto-screenshot table is exactly zero.
- macOS smoke pending — flagged `fleet:needs-macos-smoke` on push.

Closes #792

### DIFF
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -9,6 +9,7 @@
 // Components
 #include <irreden/common/components/component_position_3d.hpp>
 #include <irreden/voxel/components/component_shape_descriptor.hpp>
+#include <irreden/voxel/components/component_voxel_set.hpp>
 #include <irreden/render/components/component_canvas_ao_texture.hpp>
 #include <irreden/render/components/component_canvas_light_volume.hpp>
 #include <irreden/render/components/component_canvas_sun_shadow.hpp>
@@ -309,6 +310,22 @@ void initEntities() {
 
         EntityId ikMarker = IRPrefab::Gizmo::createIKMarker();
         IREntity::getComponent<C_Position3D>(ikMarker).pos_ = vec3(16.0f, -12.0f, -3.0f);
+    }
+
+    // Two small voxel sets to exercise multi-`C_VoxelSetNew` picking
+    // (T-219). Placed left and right of the origin so a click on either
+    // returns the correct owning entity. Voxels are integer-aligned —
+    // left-clicking any visible voxel selects it; the highlight
+    // (created below) snaps to its world center via `selection.voxelPos_`.
+    {
+        IREntity::createEntity(
+            C_Position3D{vec3(-8.0f, 0.0f, -4.0f)},
+            C_VoxelSetNew{ivec3(4, 4, 4), Color{120, 180, 240, 255}}
+        );
+        IREntity::createEntity(
+            C_Position3D{vec3(8.0f, 0.0f, -4.0f)},
+            C_VoxelSetNew{ivec3(4, 4, 4), Color{240, 180, 120, 255}}
+        );
     }
 
     // Selection-highlight entity: a slightly oversized box marker
--- a/engine/prefabs/irreden/render/CLAUDE.md
+++ b/engine/prefabs/irreden/render/CLAUDE.md
@@ -91,7 +91,12 @@ the ECS surface.
 - `VOXEL_PICKING` — editor input driver. On left-click PRESSED, casts a
   ray through the cursor (via `IRPrefab::Picking::castVoxelRay` —
   composes the same screen→world inverse `IRRender::mouseWorldPos3DAtIsoDepth`
-  uses) and walks SDF shapes to find the first hit. Writes
+  uses) and walks both visible `C_ShapeDescriptor` SDF shapes and
+  active voxels of every `C_VoxelSetNew` to find the first hit.
+  Voxel-set hits return the owning entity, the world-aligned voxel
+  coordinate, and a face normal suitable for place-adjacent math
+  (`hit.faceNormal_` — ±1 along the dominant axis of
+  `hit.worldHitPos_ - voxelCenter`). Writes
   `C_VoxelSelection` on the highlight entity and toggles its
   `C_ShapeDescriptor::flags_` visibility. Register after the camera
   systems and before `VOXEL_TO_TRIXEL_STAGE_1` so the highlight
--- a/engine/prefabs/irreden/render/picking.hpp
+++ b/engine/prefabs/irreden/render/picking.hpp
@@ -3,25 +3,35 @@
 
 // Editor-side voxel picking: convert the cursor into a world-space ray
 // (in the engine's isometric projection) and find the first SDF shape
-// it intersects. The picking inverse is the same one used by
-// `IRRender::mouseWorldPos3DAtIsoDepth`, so the recovered world point
-// matches what the voxel rasterizer wrote at any visualYaw.
+// or `C_VoxelSetNew` voxel it intersects. The picking inverse is the
+// same one used by `IRRender::mouseWorldPos3DAtIsoDepth`, so the
+// recovered world point matches what the voxel rasterizer wrote at any
+// visualYaw.
 //
-// Relationship to the GPU trixel/distance infrastructure:
-//   `castVoxelRay` reuses `IRRender::mouseWorldPos3DAtIsoDepth()` — the same
-//   screen→world inverse the GPU pipeline builds into `TRIXEL_TO_FRAMEBUFFER` —
-//   so the picking coordinate system is consistent with the rendered output at any
-//   visualYaw. It does NOT call `IRRender::getEntityIdAtMouseTrixel()`.
+// CPU-side vs GPU-readback picking — pick the right path:
 //
-//   `IRRender::getEntityIdAtMouseTrixel()` is the O(1) GPU alternative: it reads
-//   the entity-id texture written by `VOXEL_TO_TRIXEL_STAGE_2` and
-//   `SHAPES_TO_TRIXEL` (via a persistent-mapped buffer) and returns the frontmost
-//   entity at the cursor pixel in O(1). Its trade-off is a one-frame lag (the
-//   buffer holds the previous frame's render) and it yields the entity only — not
-//   the sub-voxel surface hit position. `castVoxelRay` instead evaluates the SDF at
-//   each depth step to recover the exact surface intersection (needed for sub-voxel
-//   accurate voxel placement), and fires only on click frames so the per-click cost
-//   is proportional to visible shapes × depth range, not per-frame cost.
+//   - **CPU-side (`castVoxelRay`, this file)** — the default for click-
+//     driven editor input. No frame lag: the call returns a hit on the
+//     same frame as the click. Yields the sub-voxel surface intersection
+//     point AND the face normal of the hit (`RayHit::faceNormal_`), so
+//     downstream "place adjacent" math has everything it needs without a
+//     second round-trip. Cost per click is proportional to visible
+//     shapes × depth range PLUS visible `C_VoxelSetNew` entities × depth
+//     range (each set is tested via an inverse-lookup on its grid, NOT
+//     per-voxel — so cost is independent of voxel count per set). Tens
+//     of voxel sets in a scene is fine; scenes with hundreds of
+//     thousands of active voxels distributed across many sets should
+//     prefer the GPU path below.
+//
+//   - **GPU readback (`IRRender::getEntityIdAtMouseTrixel`)** — the O(1)
+//     GPU alternative. Reads the entity-id texture written by
+//     `VOXEL_TO_TRIXEL_STAGE_2` and `SHAPES_TO_TRIXEL` (via a
+//     persistent-mapped buffer) and returns the frontmost entity at the
+//     cursor pixel in O(1). The trade-off is a one-frame lag (the buffer
+//     holds the previous frame's render) and it yields the entity only
+//     — no sub-voxel surface hit position, no face normal. Use when the
+//     scene has many active voxels and per-click CPU cost dominates, or
+//     when only the entity id matters (selection without placement).
 
 #include <irreden/ir_entity.hpp>
 #include <irreden/ir_math.hpp>
@@ -30,8 +40,11 @@
 #include <irreden/common/components/component_position_global_3d.hpp>
 #include <irreden/render/camera.hpp>
 #include <irreden/voxel/components/component_shape_descriptor.hpp>
+#include <irreden/voxel/components/component_voxel.hpp>
+#include <irreden/voxel/components/component_voxel_set.hpp>
 
 #include <optional>
+#include <span>
 #include <vector>
 
 namespace IRPrefab::Picking {
@@ -40,6 +53,12 @@ struct RayHit {
     IREntity::EntityId entity_ = IREntity::kNullEntity;
     IRMath::vec3 worldHitPos_ = IRMath::vec3(0.0f);
     IRMath::ivec3 voxelPos_ = IRMath::ivec3(0);
+    // Outward unit normal of the hit face along the dominant axis of
+    // `worldHitPos_ - voxelCenter` (±1 on exactly one axis). Only
+    // meaningful for voxel hits from `C_VoxelSetNew`; left at (0,0,0)
+    // for shape hits (the cube-face convention doesn't carry meaning
+    // for non-box SDF surfaces).
+    IRMath::ivec3 faceNormal_ = IRMath::ivec3(0);
 };
 
 // Step size along the canvas-frame iso depth axis. One iso-depth unit
@@ -61,6 +80,23 @@ struct ShapeSnapshot {
     IRRender::ShapeType type_;
     IRMath::vec4 params_;
     float boundingRadius_;
+    float isoDepthMin_;
+    float isoDepthMax_;
+};
+
+// Snapshot of one `C_VoxelSetNew` entity for the picking walk. Voxel
+// hits are resolved by inverse-lookup against the set's local grid
+// (O(1) per depth step per set) rather than SDF-testing every active
+// voxel — the local-position layout is integer-offset from
+// `worldOrigin_`, so the candidate voxel at any world point is a single
+// `roundVec3HalfUp` away.
+struct VoxelSetSnapshot {
+    IREntity::EntityId entity_;
+    // World position of the voxel at local index (0,0,0). All other
+    // voxels in the set sit at integer offsets from this origin.
+    IRMath::vec3 worldOrigin_;
+    IRMath::ivec3 size_;
+    std::span<const IRComponents::C_Voxel> voxels_;
     float isoDepthMin_;
     float isoDepthMax_;
 };
@@ -106,38 +142,114 @@ gatherVisibleShapes(IRMath::CardinalIndex cardinalIndex, IREntity::EntityId excl
     return snapshot;
 }
 
+inline std::vector<VoxelSetSnapshot>
+gatherVisibleVoxelSets(IRMath::CardinalIndex cardinalIndex, IREntity::EntityId excludeEntity) {
+    std::vector<VoxelSetSnapshot> snapshot;
+    IREntity::forEachComponent<IRComponents::C_VoxelSetNew>([&](IREntity::EntityId id,
+                                                                IRComponents::C_VoxelSetNew &vs) {
+        if (id == excludeEntity)
+            return;
+        // Headless / pre-canvas sets have no pool span yet — they
+        // can't be picked until a future canvas-attach pass moves
+        // staged data into the pool.
+        if (vs.numVoxels_ <= 0)
+            return;
+        if (vs.globalPositions_.empty() || vs.voxels_.empty())
+            return;
+        if (vs.size_.x <= 0 || vs.size_.y <= 0 || vs.size_.z <= 0)
+            return;
+
+        // worldOrigin = global position of local voxel (0,0,0).
+        // Subsequent voxels at local (x,y,z) sit at integer offsets
+        // from this origin in world space (axis-aligned; the engine
+        // currently translates voxel sets but does not rotate
+        // them). UPDATE_VOXEL_SET_CHILDREN has already refreshed
+        // globalPositions_ for this frame.
+        const IRMath::vec3 worldOrigin = vs.globalPositions_[0].pos_;
+        // Voxels at local [0, size-1] occupy world [-0.5, size-0.5],
+        // so the inflated AABB center is `worldOrigin + (size-1)/2`
+        // and the half-extent along each axis is `size/2`. Cardinal
+        // Z rotation permutes axes; the sum of half-extents
+        // (projected onto (1,1,1)) is rotation-invariant, so the
+        // iso-depth bound mirrors the shape-gather formulation.
+        const IRMath::vec3 bboxCenter =
+            worldOrigin + (IRMath::vec3(vs.size_) - IRMath::vec3(1.0f)) * 0.5f;
+        const IRMath::vec3 bboxHalf = IRMath::vec3(vs.size_) * 0.5f;
+        const IRMath::vec3 rotatedCenter = IRMath::rotateCardinalZ(bboxCenter, cardinalIndex);
+        const float center = rotatedCenter.x + rotatedCenter.y + rotatedCenter.z;
+        const float halfRange = bboxHalf.x + bboxHalf.y + bboxHalf.z;
+
+        snapshot.push_back(
+            {id,
+             worldOrigin,
+             vs.size_,
+             std::span<const IRComponents::C_Voxel>(vs.voxels_.data(), vs.voxels_.size()),
+             center - halfRange - kPickingDepthMargin,
+             center + halfRange + kPickingDepthMargin}
+        );
+    });
+    return snapshot;
+}
+
+// Picks the hit-face outward normal for a unit-cube voxel. Returns the
+// axis with the largest |delta| component, signed by that component —
+// "place adjacent" math just adds this normal to `voxelPos_` to land
+// on the cell on the entry side.
+inline IRMath::ivec3 voxelHitFaceNormal(IRMath::vec3 delta) {
+    const IRMath::vec3 absDelta = IRMath::abs(delta);
+    IRMath::ivec3 normal(0);
+    if (absDelta.x >= absDelta.y && absDelta.x >= absDelta.z) {
+        normal.x = delta.x >= 0.0f ? 1 : -1;
+    } else if (absDelta.y >= absDelta.z) {
+        normal.y = delta.y >= 0.0f ? 1 : -1;
+    } else {
+        normal.z = delta.z >= 0.0f ? 1 : -1;
+    }
+    return normal;
+}
+
 } // namespace detail
 
 // Casts a ray from the cursor into the scene and returns the first
-// shape it hits, or std::nullopt if the ray misses everything. The
-// caller passes the editor's highlight entity (or any entity to be
-// skipped) so the highlight itself doesn't catch its own ray.
+// shape or voxel it hits, or std::nullopt if the ray misses
+// everything. The caller passes the editor's highlight entity (or any
+// entity to be skipped) so the highlight itself doesn't catch its own
+// ray.
 //
 // Algorithm: walk along the canvas-frame iso depth axis in
 // `kPickingDepthStep` increments over the union of all visible shape
-// iso-depth ranges. At each step, evaluate every shape's SDF at the
-// recovered world point; the first surface (`SDF::evaluate <=
-// kSurfaceThreshold`) wins. The depth-range pre-filter keeps the
-// per-step cost proportional to the number of shapes whose bounding
-// box overlaps that depth slice, not the total scene shape count.
+// and voxel-set iso-depth ranges. At each step, evaluate (a) every
+// shape's SDF at the recovered world point and (b) for each visible
+// `C_VoxelSetNew`, look up the candidate voxel via inverse-grid
+// indexing — the first surface hit (`SDF::evaluate <=
+// kSurfaceThreshold` for shapes; inside-unit-cube for voxels) wins.
+// The depth-range pre-filter keeps the per-step cost proportional to
+// the count of shapes / sets whose bounding box overlaps that depth
+// slice, not the total scene count.
 inline std::optional<RayHit>
 castVoxelRay(IREntity::EntityId excludeEntity = IREntity::kNullEntity) {
     const auto [rasterYaw, residualYaw] = IRPrefab::Camera::getYawSplit();
     const IRMath::CardinalIndex cardinalIndex = IRMath::rasterYawCardinalIndex(rasterYaw);
 
     auto shapes = detail::gatherVisibleShapes(cardinalIndex, excludeEntity);
-    if (shapes.empty())
+    auto voxelSets = detail::gatherVisibleVoxelSets(cardinalIndex, excludeEntity);
+    if (shapes.empty() && voxelSets.empty())
         return std::nullopt;
 
-    float depthMin = shapes[0].isoDepthMin_;
-    float depthMax = shapes[0].isoDepthMax_;
+    float depthMin = shapes.empty() ? voxelSets[0].isoDepthMin_ : shapes[0].isoDepthMin_;
+    float depthMax = shapes.empty() ? voxelSets[0].isoDepthMax_ : shapes[0].isoDepthMax_;
     for (const auto &s : shapes) {
         depthMin = IRMath::min(depthMin, s.isoDepthMin_);
         depthMax = IRMath::max(depthMax, s.isoDepthMax_);
     }
+    for (const auto &vs : voxelSets) {
+        depthMin = IRMath::min(depthMin, vs.isoDepthMin_);
+        depthMax = IRMath::max(depthMax, vs.isoDepthMax_);
+    }
 
     for (float d = depthMin; d <= depthMax; d += kPickingDepthStep) {
         const IRMath::vec3 worldPoint = IRRender::mouseWorldPos3DAtIsoDepth(d);
+
         for (const auto &s : shapes) {
             if (d < s.isoDepthMin_ || d > s.isoDepthMax_)
                 continue;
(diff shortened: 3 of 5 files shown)


jakildev (commented):
## Review — T-219: extend castVoxelRay to walk C_VoxelSetNew entities

**Verdict:** needs-fix

---

### Praise

- Inverse-grid lookup (`roundVec3HalfUp(worldPoint - worldOrigin)`) is the right O(1)-per-set-per-step algorithm — no per-voxel SDF loop at each depth step, cost independent of voxel count per set.
- Cascaded guard set in `gatherVisibleVoxelSets` (`numVoxels_ <= 0`, `globalPositions_.empty()`, `voxels_.empty()`, zero-dimension check) is thorough and handles pre-canvas headless sets correctly.
- `std::span<const C_Voxel>` lifetime is safe: snapshot is built before the depth loop, no ECS structural changes occur inside the pure-read walk, so the raw pointer from `voxels_.data()` stays valid for the call duration.
- `voxelHitFaceNormal` logic is correct and well-documented. `IRMath::abs` used throughout (not `std::abs`). Tie-break and zero-delta behavior documented in tests.
- `RayHit::faceNormal_` defaulting to `ivec3(0)` for shape hits is backward-compatible — existing callers that ignore the new field see no change.
- `forEachComponent<C_VoxelSetNew>` is the correct batched-column read pattern (not per-entity `getComponent`). No ECS footgun.
- All math uses IRMath primitives throughout. Naming conventions followed (trailing `_` on members, camelCase free functions, PascalCase struct).
- 6 unit tests cover `voxelHitFaceNormal` directly with all sign/axis combinations, corner tie-break, and zero-delta fallback.

---

### Needs-fix

**`SDF::box(delta, vec3(0.0f))` is a sphere test, not a unit cube test.**

`engine/prefabs/irreden/render/picking.hpp` — the voxel-hit guard inside the depth loop:

```cpp
// Inside the unit voxel cube?
if (IRMath::SDF::box(delta, IRMath::vec3(0.0f)) > IRMath::SDF::kSurfaceThreshold)
    continue;
```

With halfExtent `= vec3(0.0f)`, the standard SDF box formula collapses:
`d = abs(delta) - vec3(0) = abs(delta)`, all components non-negative, so
`length(max(d,0)) + min(max_component,0) = length(delta) + 0 = length(delta)`.

The condition `length(delta) > kSurfaceThreshold (= 0.5f)` is therefore a sphere
test of radius 0.5 around the voxel center — NOT the unit cube (half-extent 0.5).
Points in the cube's corner regions (`length(delta) in (0.5, sqrt(3)/2 ~= 0.866)`)
correctly round to this voxel via the inverse lookup but fail the sphere guard and
are silently skipped. For an editor pick tool, this means clicks near voxel
edges/corners can miss even when a solid voxel is under the cursor.

The comment "Inside the unit voxel cube?" and the PR description both confirm the
intent is a cube test. The fix is to pass the actual unit-cube half-extent:

```cpp
if (IRMath::SDF::box(delta, IRMath::vec3(0.5f)) > 0.0f)
    continue;
```

(SDF::box with half-extent 0.5f returns negative inside the cube, zero on the
surface, positive outside — `> 0.0f` is the clean guard. A small negative slack
like `> -0.01f` is fine if floating-point rounding near the face boundary matters,
but avoid using `kSurfaceThreshold` as slack since 0.5f would extend detection into
adjacent cells.)

---

### Checklist

**ECS invariants:** `forEachComponent` (batched column read, not per-entity `getComponent`). No structural changes inside the depth loop walk. No allocations inside the hot loop (snapshot built once before depth iteration). ✓

**Span lifetime:** Snapshot lives on the stack for the `castVoxelRay` call; no ECS mutation occurs in the pure-read depth loop. Safe. ✓

**Math primitives:** `IRMath::abs`, `IRMath::min`, `IRMath::max`, `IRMath::roundVec3HalfUp`, `IRMath::rotateCardinalZ`, `IRMath::index3DtoIndex1D` — all IRMath wrappers, no raw `std::` math. ✓

**Naming:** `VoxelSetSnapshot` (PascalCase struct), `gatherVisibleVoxelSets` / `voxelHitFaceNormal` (camelCase free functions in detail namespace), `faceNormal_` (trailing underscore member). ✓

**Serialized structs:** No persisted format touched. `RayHit` is a transient return value. ✓

**Cross-host smoke:** Changed files are `engine/prefabs/irreden/render/picking.hpp` and `test/render/`. No GPU shaders, no pipeline bindings, no CPU/GPU struct changes. PR has `fleet:authored-on-linux` — `fleet:needs-macos-smoke` already set by the author push. No additional smoke tagging needed.

---

Opus recheck required: picking.hpp is in engine/prefabs/irreden/render/ and the SDF halfExtent issue (vec3(0.0f) vs vec3(0.5f)) is a pick-correctness matter for an editor tool; worth a second pass to confirm whether the sphere approximation is intentional and whether the span-over-forEachComponent pattern is safe under all archetype migration paths.

jakildev:
Addressed feedback:
- Fixed `SDF::box` halfExtent: changed `vec3(0.0f)` → `vec3(0.5f)` and threshold `> kSurfaceThreshold` → `> 0.0f` in the voxel-hit guard inside `castVoxelRay`. This is now a proper unit-cube containment test (SDF returns negative inside the cube, zero on surface, positive outside) rather than a sphere test of radius 0.5 that incorrectly missed corner/edge voxel hits.

Note: build verification skipped — cmake not installed on this Linux fleet host. PR already has `fleet:authored-on-linux` and the macOS smoke label will validate the full build. — sonnet-author

jakildev (commented):
## Opus recheck — T-219: extend castVoxelRay to walk C_VoxelSetNew entities

**Verdict:** needs-fix (keeps Sonnet's verdict; corrects the rationale and lists the supporting concerns)

---

### Sonnet flagged the SDF::box halfExtent as a sphere-vs-cube bug. On closer read I disagree on the mechanism — but the recommended code change still stands.

Sonnet's analysis assumed the standard Inigo Quilez Euclidean box SDF:
`length(max(d, 0)) + min(max_component, 0)`. Under that formula, `box(delta, vec3(0))` does collapse to `length(delta)` and the test reduces to a sphere of radius 0.5.

But the engine's `box()` (`engine/math/include/irreden/math/sdf.hpp:64-67`) is **not** that formula. It is the Chebyshev / L∞ form:

```cpp
inline float box(vec3 p, vec3 halfExtents) {
    vec3 d = glm::abs(p) - halfExtents;
    return glm::max(d.x, glm::max(d.y, d.z));
}
```

There is no `length(max(d, 0))` term. With `halfExtents = vec3(0)`, this returns `max(|dx|, |dy|, |dz|)` — the L∞ norm of `delta`. The condition `> kSurfaceThreshold (= 0.5)` is therefore TRUE iff `max(|dx|, |dy|, |dz|) > 0.5`, which is **exactly** "outside the unit cube of half-extent 0.5." So the current code is in fact a correct unit-cube test, and corner clicks are not silently dropped.

That said, the recommended replacement — `box(delta, vec3(0.5f)) > 0.0f` — is still the right change to make in `engine/prefabs/irreden/render/picking.hpp:283`, for two reasons:

1. **Self-documenting.** `vec3(0.5f)` is the half-extent of the unit cube. A future reader sees the intent without having to recall that the engine's `box()` is L∞-shaped rather than Euclidean.
2. **Future-proof.** If `IRMath::SDF::box` is ever upgraded to the full Euclidean form (e.g. for SDF-shape rendering quality), this picking site would silently degrade into exactly the sphere-of-radius-0.5 test Sonnet described. The `vec3(0.5f), > 0.0f` formulation works correctly under both the current L∞ form and a future Euclidean form.

Please apply Sonnet's suggested edit — the right change, just for clarity + maintenance reasons rather than because it fixes a current behavioral bug.

---

### Secondary observations (worth noting; the verdict above is the single thing blocking)

**1. The cube check is essentially redundant after the lookup.**

`IRMath::roundHalfUp(v) = (int)floor(v + 0.5f)` (round half toward +∞). After `localInt = roundVec3HalfUp(worldPoint - vs.worldOrigin_)` succeeds the bounds check, each component of `delta = worldPoint - candidateCenter` is in `[-0.5, 0.5)` by construction. So `max(|dx|, |dy|, |dz|) <= 0.5` is already guaranteed — the L∞ check only ever fires at the asymmetric half-integer boundary (and even then it goes the "inside" way: `0.5 > 0.5` is FALSE). The check is harmless but not load-bearing; deleting it entirely is also a valid fix. Either way, the bounds check + activity check are doing the real work. Mention only — pick whichever shape you prefer.

**2. Span lifetime under archetype migration — confirmed safe.**

`vs.voxels_` is a `std::span<C_Voxel>` viewing into the canvas voxel pool (not into the C_VoxelSetNew row itself). The snapshot captures the span by value into a stack vector; from that point on `castVoxelRay` does pure reads — no ECS structural changes, no pool growth, no `forEachComponent` re-entry. The snapshot pattern is safe under any archetype migration path that might occur between `castVoxelRay` invocations (between-call migrations move the C_VoxelSetNew struct, but the span continues to point at the same pool storage). Sonnet's "no structural changes inside the depth loop walk" is correct.

**3. Layout consistency between picker and writer — confirmed.**

`detail::gatherVisibleVoxelSets` flat-indexes via `IRMath::index3DtoIndex1D(localInt, size_)` which is x-fastest (`z*sx*sy + y*sx + x`). The C_VoxelSetNew constructors and `setVoxel` (lines 124-125, 206, 236, 272, 286 of `component_voxel_set.hpp`) write the pool using the same helper. The picked-voxel index matches what the GPU pipeline sees.

**4. UPDATE-vs-RENDER pipeline ordering — implicit but correct.**

`gatherVisibleVoxelSets` reads `vs.globalPositions_[0].pos_` to derive `worldOrigin`. `UPDATE_VOXEL_SET_CHILDREN` is the system that keeps that span fresh, and it runs in the UPDATE pipeline. `VOXEL_PICKING` is in the RENDER pipeline (per `engine/prefabs/irreden/render/CLAUDE.md`), so the world origin is always one-frame-current. No fix needed; worth a sentence in the picking.hpp comment block alongside the existing "the engine currently translates voxel sets but does not rotate them" note, since the same paragraph is the natural home for both invariants.

**5. The `voxelHitFaceNormal` unit tests exercise the helper in isolation — but not `castVoxelRay`.**

The 6 new tests catch a regression in the dominant-axis math, which is good. They do **not** catch a regression in the SDF::box test itself (Sonnet's hypothetical sphere-test future, or a subtle off-by-half-voxel in the inverse lookup). End-to-end picking is tested manually via the `IRVoxelEditor` click loop only. Not blocking, but adding even one test that drives `castVoxelRay` against a synthetic 1-voxel set at a known world position and confirms `RayHit::voxelPos_ + faceNormal_` lands on the expected adjacent cell would close the regression-detection gap for the entire picking path. Worth doing in a follow-up.

**6. `gatherVisibleVoxelSets` does not check a visibility flag.**

The function name and comment imply visibility filtering, mirroring `gatherVisibleShapes` (which checks `SHAPE_FLAG_VISIBLE`). `C_VoxelSetNew` has no analogous flag today, so the function iterates every set. That is correct given the current data model, but the naming asymmetry will surprise the next reader who adds a visibility concept to voxel sets. A one-line comment in the function header noting "no visibility flag on voxel sets today; gathers all" would prevent future drift.

---

### Items Sonnet's checklist already confirmed (re-verified, all clean)

- **ECS invariants** — `forEachComponent<C_VoxelSetNew>` is the correct batched-column pattern; no per-entity `getComponent` in the gather lambda; no structural changes inside the depth loop. ✓
- **No allocations in hot loop** — `std::vector<VoxelSetSnapshot>` is allocated once per `castVoxelRay` call (click-only, not per-frame). ✓
- **IRMath usage** — `IRMath::abs`, `IRMath::min`, `IRMath::max`, `IRMath::roundVec3HalfUp`, `IRMath::rotateCardinalZ`, `IRMath::index3DtoIndex1D`, `IRMath::SDF::box`. No raw `std::` or `glm::`. ✓
- **Naming** — `VoxelSetSnapshot` (PascalCase struct), `gatherVisibleVoxelSets`/`voxelHitFaceNormal` (camelCase free functions in `detail`), `faceNormal_` (trailing `_`). ✓
- **No persisted format touched** — `RayHit` is transient. ✓
- **Cross-host smoke** — labels already in place (`fleet:authored-on-linux` + `fleet:needs-macos-smoke`). ✓

---

### The one thing to fix before merge

Apply Sonnet's recommended edit at `engine/prefabs/irreden/render/picking.hpp:283`:

```cpp
// before
if (IRMath::SDF::box(delta, IRMath::vec3(0.0f)) > IRMath::SDF::kSurfaceThreshold)
    continue;

// after
if (IRMath::SDF::box(delta, IRMath::vec3(0.5f)) > 0.0f)
    continue;
```

The current code is correct today; the change is purely clarity + future-proofing against a hypothetical change to `IRMath::SDF::box`. Once that lands, this PR is good to merge.

---

🤖 Reviewed by Claude Opus 4.7 (opus-reviewer role) — final pass after Sonnet 4.6 first pass

jakildev (commented):
## Re-review — T-219: extend castVoxelRay to walk C_VoxelSetNew entities

_Triggered by fleet:changes-made. Prior review (2026-05-17T03:09:18Z) flagged one needs-fix: `SDF::box` halfExtent was `vec3(0.0f)` (sphere test) instead of `vec3(0.5f)` (proper unit-cube containment)._

**Verdict:** approve

---

### Fix verified

The voxel-hit guard in `castVoxelRay` (`picking.hpp`) now reads:

```cpp
if (IRMath::SDF::box(delta, IRMath::vec3(0.5f)) > 0.0f)
    continue;
```

This is exactly the requested fix. `SDF::box` with half-extent `0.5f` returns negative inside the unit cube, zero on the surface, and positive outside — `> 0.0f` is the correct guard. Clicks near voxel edges and corners now register correctly instead of falling into the previous sphere's blind zone (the corner region where `length(delta) in (0.5, 0.866)` passes the inverse lookup but fails a sphere test). ✓

---

### Checklist (re-verified post-fix)

**ECS invariants:** `forEachComponent<C_VoxelSetNew>` is a batched column read. No per-entity `getComponent` in the iteration. No allocations inside the depth loop (snapshot built once before entering the walk). No structural changes inside the pure-read depth loop. ✓

**Span lifetime:** `std::span<const C_Voxel>` is built inside the `forEachComponent` callback and stored in `VoxelSetSnapshot` before the depth loop begins. The engine's single-threaded ECS tick guarantees no concurrent archetype migration during the call — the raw pointer from `vs.voxels_.data()` stays valid for the duration of `castVoxelRay`. ✓

**Depth-bounds initialization:** The `depthMin` / `depthMax` ternary correctly seeds from `voxelSets[0]` when `shapes` is empty (which is reachable if the scene has voxel sets but no SDF shapes). The early `return std::nullopt` when both are empty means `voxelSets[0]` is always valid at that point. ✓

**flatIdx bounds:** `localInt` is range-checked against `[0, size-1]` on each axis before `index3DtoIndex1D`. The additional `flatIdx >= vs.voxels_.size()` guard is a correct defensive belt-and-suspenders for the edge case where `voxels_.size()` diverges from `size_.x * size_.y * size_.z`. ✓

**Math primitives:** All IRMath wrappers throughout — `IRMath::abs`, `IRMath::min`, `IRMath::max`, `IRMath::roundVec3HalfUp`, `IRMath::rotateCardinalZ`, `IRMath::index3DtoIndex1D`. No `std::` math. ✓

**Naming:** `VoxelSetSnapshot` (PascalCase), `gatherVisibleVoxelSets` / `voxelHitFaceNormal` (camelCase in detail namespace), `faceNormal_` (trailing underscore member). ✓

**Serialized structs:** `RayHit` is a transient return value. No persisted format touched. ✓

**Tests:** 6 unit tests for `voxelHitFaceNormal` remain in place covering all axes (±X, ±Y, ±Z covered by the four positive/negative tests), corner tie-break, and zero-delta fallback. ✓

**Cross-host smoke:** No GPU shaders, no pipeline bindings, no CPU/GPU struct changes. `fleet:needs-macos-smoke` already set. ✓

---

Opus recheck required: `picking.hpp` is in `engine/prefabs/irreden/render/` — the span-over-forEachComponent pattern warrants an independent second pass confirming that the span pointer cannot be invalidated by any archetype migration path that the single-tick ordering does not rule out.

jakildev:
Race condition — human jakildev pushed the unit-cube guard fix as 6e3a3314 (`SDF::box(delta, vec3(0.5f)) > 0.0f`) at the same time this Opus iteration was preparing the identical fix locally. Adopting the human's commit; my redundant local commit is being dropped.

Verifying the recheck note from the original review:

1. **Span-over-forEachComponent safety under archetype migration.** The `std::span<const C_Voxel>` captured in `VoxelSetSnapshot` is built synchronously inside `gatherVisibleVoxelSets` and consumed inside the same `castVoxelRay` call. The depth-loop body is pure-read with respect to ECS structure (no `createEntity` / `removeComponent` / `setParent`). `component_voxel_set.hpp:298-307` notes the spans become invalid across between-frame canvas archetype migrations, but `castVoxelRay` returns before the next frame's UPDATE pipeline ticks. Safe under all current call paths (`VOXEL_PICKING` is INPUT pipeline; no archetype changes in INPUT). Future callers must keep the call frame-bounded.

2. **Sphere-vs-cube classification.** With the local `SDF::box` impl as `max(|x|-hx, |y|-hy, |z|-hz)` (Chebyshev / max-norm, not the standard rounded-corner SDF), the old form `box(delta, vec3(0)) > 0.5` was accidentally equivalent to a cube test — `max(|x|,|y|,|z|) > 0.5` IS the unit-cube boundary. So no corner voxels were actually being missed under today's `SDF::box`. But the human's fix is still correct and necessary: the new form expresses the intent directly and won't silently degrade to a sphere test if `SDF::box` is ever swapped to the standard rounded-corner formula.

— opus-worker

jakildev (commented):
## Opus recheck — T-219: extend castVoxelRay to walk C_VoxelSetNew entities

**Verdict:** approve

_Sonnet's re-review approved after the `SDF::box` halfExtent fix (`vec3(0.5f) > 0.0f`) landed in 6e3a3314. Sonnet flagged one outstanding question for this pass: confirm the `std::span<const C_Voxel>` captured in `VoxelSetSnapshot` cannot be invalidated by any archetype-migration path that the single-tick ordering does not rule out. That is the focus of this recheck._

---

### Span-over-`forEachComponent` safety — confirmed safe under the current architecture

The relevant span is `vs.voxels_`, a view into the canvas's `C_VoxelPool` storage (specifically a `std::vector<C_Voxel>` slice keyed by `voxelStartIdx_` + `numVoxels_`). The component header at `engine/prefabs/irreden/voxel/components/component_voxel_set.hpp:298-307` explicitly documents the relevant hazard:

> All four span members captured at allocation time (`positions_`, `positionOffsets_`, `globalPositions_`, `voxels_`) become invalid after [canvas archetype migration]; only `voxelStartIdx_` and `numVoxels_` remain stable...

What can invalidate the underlying vector storage:

1. **Pool growth** — only `IRRender::allocateVoxels`, called from `C_VoxelSetNew`'s constructors (i.e. requires `createEntity`).
2. **Canvas archetype migration** — deep-copies `C_VoxelPool` and frees the old storage; documented as a *between-frame* event.
3. **Pool deallocation** — `IRRender::deallocateVoxels` from `C_VoxelSetNew::onDestroy()` (i.e. requires `removeEntity`).

All three require a structural ECS change. Within a single `castVoxelRay` invocation:

- `gatherVisibleVoxelSets` is a pure read — its `forEachComponent` lambda only reads scalar metadata (`numVoxels_`, `globalPositions_[0].pos_`, `size_`) and captures the span by value into a stack `std::vector<VoxelSetSnapshot>`. No allocator side effects, no structural ECS mutations, no recursive `forEachComponent` re-entry that could touch a different system's state.
- The depth-loop body iterates the snapshot vector, indexes `vs.voxels_[flatIdx].color_.alpha_`, and returns `RayHit{...}`. No structural ECS calls.
- The owning system `VOXEL_PICKING` only mutates fields on its iterating highlight entity (`C_VoxelSelection`, `C_Position3D`, `C_PositionGlobal3D`, `C_ShapeDescriptor::flags_`) — pure field writes, no archetype change. Those mutations happen AFTER `castVoxelRay` returns.

The engine's single-threaded tick guarantees no other system runs concurrently. The deferred-structural-change contract (`engine/prefabs/CLAUDE.md` and `cpp-ecs.md`) says `flushStructuralChanges` runs at pipeline boundaries, not mid-tick. So between `gather` and the depth-loop, and across the depth-loop's lifetime, nothing can grow, migrate, or shrink the underlying `C_VoxelPool` storage. The captured `voxels_.data()` pointer is stable for the call's duration.

**Cross-call risk: none.** Each `castVoxelRay` rebuilds the snapshot, so prior-frame migrations are reflected before any subsequent walk.

**Future-proofing note (not a needs-fix):** if a future caller invokes `castVoxelRay` from a system that interleaves structural changes *inside* its own tick — e.g. a fancy multi-pick flow that spawns new entities between picks — the snapshot's call-bounded lifetime still saves them, because the snapshot is rebuilt per call. The pattern is robust under reasonable extensions. The only way to break it would be to start passing a `VoxelSetSnapshot` (or its inner span) *out* of `castVoxelRay` to be reused later — which the function does not, and the `RayHit` return only carries scalars + an `EntityId`.

---

### Fix verification (defers to Sonnet)

Sonnet's re-review already verified the `SDF::box(delta, vec3(0.5f)) > 0.0f` edit. I re-read `picking.hpp:283` — same conclusion. The half-extent-0.5 form is the L∞ unit-cube test under the engine's current `SDF::box` formulation (and remains the correct Euclidean form if `SDF::box` is ever upgraded — the future-proofing rationale from my prior recheck). The depth-loop guard is now self-documenting.

The author comment at 2026-05-17T03:16:03Z accurately notes that under today's `SDF::box` (Chebyshev/L∞), the prior `vec3(0.0f) > kSurfaceThreshold (= 0.5)` form was accidentally equivalent — no corner-region clicks were silently dropped in practice. The fix is still the right change for clarity and future-proofing.

---

### Items Sonnet's re-review confirmed (not re-checked here — verified by Sonnet, defer to that pass)

- ECS invariants: `forEachComponent` batched-column pattern; no per-entity `getComponent` in iteration; no allocations inside the depth loop; no structural changes in the pure-read depth walk.
- Depth-bounds initialization (`shapes.empty() ? voxelSets[0]...` ternary) correctness.
- `flatIdx >= vs.voxels_.size()` defensive bound.
- All math via IRMath wrappers.
- Naming conventions: `VoxelSetSnapshot` PascalCase, `gatherVisibleVoxelSets`/`voxelHitFaceNormal` camelCase in `detail`, `faceNormal_` trailing `_`.
- No persisted format touched (`RayHit` is transient).
- Tests cover `voxelHitFaceNormal` across axes, corner tie-break, zero-delta fallback.
- Cross-host smoke labels already correct (`fleet:authored-on-linux` + `fleet:needs-macos-smoke`).

---

### Standing nits from prior Opus review (not pre-merge; carry forward as the author chooses)

The prior Opus review (2026-05-17T03:15:29Z) listed five secondary observations beyond the SDF::box fix. The author addressed the one needs-fix and shipped; the secondary items remain genuinely-optional polish:

1. The cube check after `roundVec3HalfUp` is essentially redundant — `roundHalfUp` puts each `delta` component in `[-0.5, 0.5)`, so `box(delta, vec3(0.5f)) > 0.0f` is provably never true. Deleting the line simplifies, but the line is harmless and self-documenting; either shape is fine.
2. One-line doc on UPDATE→RENDER pipeline ordering for `globalPositions_[0].pos_` freshness (the existing "engine currently translates voxel sets but does not rotate them" paragraph is the natural home).
3. A `castVoxelRay` end-to-end test against a synthetic 1-voxel set would close the regression-detection gap that the `voxelHitFaceNormal`-only tests leave open.
4. `gatherVisibleVoxelSets` doesn't check a visibility flag (no analogue to `SHAPE_FLAG_VISIBLE` on `C_VoxelSetNew` today); a one-line comment in the function header would prevent future drift.
5. Layout consistency between picker and writer via `index3DtoIndex1D` — already verified clean; no action.

None block merge. Listing here for visibility, not adding `fleet:has-nits` — the author can pick these up at their discretion as part of T-211 follow-up work.

---

🤖 Reviewed by Claude Opus 4.7 (opus-reviewer role) — final pass after Sonnet 4.6 first pass + re-review

jakildev (commented):
## Re-review — T-219: extend castVoxelRay to walk C_VoxelSetNew entities

_Prior review (03:09Z) was needs-fix: `SDF::box(delta, vec3(0.0f))` was a sphere test of radius 0.5, not a unit-cube containment test. Author addressed at 03:12Z — changed halfExtent to `vec3(0.5f)` and threshold to `> 0.0f`. Reviewing the fix and the full diff._

**Verdict:** approve

---

### Praise

- **Fix is correct.** `IRMath::SDF::box(delta, IRMath::vec3(0.5f)) > 0.0f` is the proper unit-cube containment guard. SDF returns negative inside the cube, zero on the surface, positive outside — `> 0.0f` cleanly gates "not inside." The prior sphere approximation that missed corner/edge voxels is resolved.
- Inverse-grid lookup (`roundVec3HalfUp(worldPoint - vs.worldOrigin_)`) is O(1) per set per depth step — cost independent of voxel count per set. Correct algorithm for an editor picking tool.
- Guard cascade in `gatherVisibleVoxelSets` (`numVoxels_ <= 0`, `globalPositions_.empty()`, `voxels_.empty()`, zero-dimension check) handles all pre-canvas headless set states thoroughly.
- `std::span<const C_Voxel>` snapshot is built before the depth loop and holds a raw pointer into `vs.voxels_`. Click-frame context (not in a tick), pure-read depth loop, no ECS structural mutations during the walk — span lifetime is safe for this call pattern.
- `voxelHitFaceNormal` uses `IRMath::abs` throughout (not `std::abs`). Tie-break (x over y over z by condition ordering) is deterministic and documented in tests. Zero-delta fallback (+X) is documented.
- `forEachComponent<C_VoxelSetNew>` is the correct batched-column read pattern — not per-entity `getComponent`. No ECS footgun.
- All math uses IRMath wrappers (`roundVec3HalfUp`, `rotateCardinalZ`, `index3DtoIndex1D`, `abs`, `min`, `max`). No raw `std::` math.
- `RayHit::faceNormal_` defaults to `ivec3(0)` for shape hits — backward-compatible for callers that ignore the new field.
- 6 unit tests cover `voxelHitFaceNormal` with all sign/axis combinations, corner tie-break, and zero-delta fallback.

---

### Checklist

**ECS invariants:** `forEachComponent` (batched column read, not per-entity `getComponent`). No structural mutations inside the depth loop. Snapshot built once before iteration. No allocations in the hot depth loop. ✓

**Span lifetime:** Snapshot lives on the stack for the `castVoxelRay` call. Click-frame context, pure-read depth loop, no ECS structural changes during the walk. Safe. ✓

**Math primitives:** IRMath wrappers throughout — no raw `std::sin/cos/abs/sqrt/min/max`. ✓

**Naming:** `VoxelSetSnapshot` (PascalCase struct), `gatherVisibleVoxelSets` / `voxelHitFaceNormal` (camelCase free functions in detail namespace), `faceNormal_` (trailing underscore member). ✓

**Serialized structs:** `RayHit` is a transient return value. `C_VoxelSetNew` and `C_Voxel` are read-only. No persisted format changes. ✓

**Cross-host smoke:** Changed files are `engine/prefabs/irreden/render/picking.hpp`, `test/render/picking_face_normal_test.cpp`, and `creations/editors/voxel_editor/main.cpp`. No GPU shaders, no pipeline bindings, no CPU/GPU struct layout changes. PR already has `fleet:authored-on-linux` and `fleet:needs-macos-smoke`. No additional smoke tagging needed. ✓

---

### Question for Opus

One invariant I cannot verify from this diff alone: **`index3DtoIndex1D` axis ordering must match `C_VoxelSetNew::voxels_` storage layout.** If `voxels_` uses a different axis-major order than `index3DtoIndex1D` assumes, picks would land on the wrong voxel while appearing correct at the grid origin. The call `flatIdx = IRMath::index3DtoIndex1D(localInt, vs.size_)` is correct IF the two orderings agree — worth confirming in the Opus pass.

---

Opus recheck required: picking.hpp is in engine/prefabs/irreden/render/; verify the index3DtoIndex1D-vs-voxels_-storage-layout ordering invariant and confirm the span-over-forEachComponent pattern is safe under all archetype migration paths.

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). 7 screenshots captured across 7 shots (zoom1_origin, zoom2_origin, zoom4_origin, zoom1_odd_offset, zoom8_origin, zoom4_offset_3_5, zoom16_lod_all_visible). No build errors, no runtime errors, clean exit.

PR #799, opened 2026-05-18
T-220: entity: dedup globalFieldRegistry — return stable FieldBindingId on repeated registerField calls

### LINKED ISSUE #512: globalFieldRegistry: dedup field names across World restarts
## Context

Follow-up nit from PR #508 (T-100) Opus recheck. `engine/prefabs/irreden/common/modifier_field_registry.hpp:25-29`:

\`\`\`cpp
IRComponents::FieldBindingId registerField(const char *name) {
    const auto id = static_cast<IRComponents::FieldBindingId>(m_names.size());
    m_names.push_back(name);
    return id;
}
\`\`\`

No name→id reverse check. Within a single \`World\`, \`engine/script/src/lua_script.cpp\` guards via \`em.isComponentRegistered(componentName)\` before reaching \`registerField\`, so the lua side is safe at first registration. **Across \`World\` teardown + reconstruct (or repeated test fixtures within one process)**, each re-registration of the same component name appends \`"X.field"\` again with a fresh \`FieldBindingId\`.

## Why this matters

- \`FieldBindingId\` is \`uint16_t\` so the entry count caps at ~65k.
- The design-doc promise (\"modifier ids are issued once and consumers may keep them across world restarts\") is silently broken — the *string* is stable, but the *id* drifts.
- No current test stores a \`FieldBindingId\` across fixture instances, so this is latent.

## Suggested approach

Linear scan over `m_names` in `registerField` to return an existing id when the name is already present. Registry stays small (<100 entries in practice) so an O(N) scan per registration is fine; registration is init-time only.

## Suggested model
[sonnet]

## Suggested area
engine/prefabs/irreden/common/modifier_field_registry.hpp

## Acceptance criteria

- [ ] `registerField(\"foo\")` followed by another `registerField(\"foo\")` returns the same `FieldBindingId`.
- [ ] Test asserts id stability across two `World` constructions (or two fixture instances) within a single process.
- [ ] No regression in modifier_runtime_test / modifier_lua_test.

jakildev:
## Summary

- `registerField` now calls `findFieldId` before appending; re-registering the same name returns the existing `FieldBindingId` instead of pushing a fresh entry
- Prevents id drift across `World` teardown/reconstruct cycles and across repeated test fixtures within one process
- `registerFieldVec3` and `registerFieldQuat` delegate to `registerField` so they inherit the guard without extra code

## Test plan

- [x] `registerField("foo")` called twice returns the same `FieldBindingId` (`RegisterDedupReturnsSameId`)
- [x] Field count stays constant on re-register (`RegisterDedupFieldCountDoesNotGrow`)
- [x] vec3 dedup round-trip (`RegisterVec3DedupReturnsSameId`)
- [x] Global registry id stability simulating World restart (`GlobalRegistryIdStableAcrossReinit`)
- [ ] No regression in modifier_runtime_test / modifier_lua_test (cmake not available on this fleet host — Ubuntu 20.04→24.04 migration in progress; reviewer should validate on a host with cmake)

## Notes

The fix is 3 lines in `FieldRegistry::registerField` using the existing `findFieldId` linear scan. Registration is init-time only, registry stays < 100 entries in practice, so the extra O(N) probe per registration is acceptable.

Closes #512

### DIFF
--- a/engine/prefabs/irreden/common/modifier_field_registry.hpp
+++ b/engine/prefabs/irreden/common/modifier_field_registry.hpp
@@ -28,6 +28,9 @@ class FieldRegistry {
     IRComponents::FieldBindingId registerField(
         const char *name, IRComponents::FieldValueType type = IRComponents::FieldValueType::SCALAR
     ) {
+        const auto existing = findFieldId(name);
+        if (existing != IRComponents::kInvalidFieldId)
+            return existing;
         const auto id = static_cast<IRComponents::FieldBindingId>(m_names.size());
         m_names.push_back(name);
         m_types.push_back(type);
--- a/test/ecs/modifier_runtime_test.cpp
+++ b/test/ecs/modifier_runtime_test.cpp
@@ -63,6 +63,43 @@ TEST(ModifierRegistry, GlobalSingletonIsStable) {
     EXPECT_EQ(&r1, &r2);
 }
 
+TEST(ModifierRegistry, RegisterDedupReturnsSameId) {
+    FieldRegistry registry;
+    auto id1 = registry.registerField("foo");
+    auto id2 = registry.registerField("foo");
+    EXPECT_EQ(id1, id2);
+    EXPECT_NE(id1, IRComponents::kInvalidFieldId);
+}
+
+TEST(ModifierRegistry, RegisterDedupFieldCountDoesNotGrow) {
+    FieldRegistry registry;
+    registry.registerField("x");
+    EXPECT_EQ(registry.fieldCount(), 1u);
+    registry.registerField("x");
+    EXPECT_EQ(registry.fieldCount(), 1u);
+}
+
+TEST(ModifierRegistry, RegisterVec3DedupReturnsSameId) {
+    FieldRegistry registry;
+    auto id1 = registry.registerFieldVec3("transform.translation");
+    auto id2 = registry.registerFieldVec3("transform.translation");
+    EXPECT_EQ(id1, id2);
+    EXPECT_NE(id1, IRComponents::kInvalidFieldId);
+}
+
+TEST(ModifierRegistry, GlobalRegistryIdStableAcrossReinit) {
+    // Simulates two World constructions: the static globalFieldRegistry()
+    // persists across EntityManager teardown/reconstruct, so re-registering
+    // the same name must return the same FieldBindingId rather than
+    // appending a fresh entry.
+    auto id1 =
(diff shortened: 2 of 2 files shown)


jakildev (commented):
## Review — T-220: entity: dedup globalFieldRegistry — return stable FieldBindingId on repeated registerField calls (PR #799)

**Verdict:** approve

---

### Praise

- The `i = 1` loop start is correct and intentional — `FieldRegistry::FieldRegistry()` pre-pushes a `nullptr` sentinel at index 0 (`// index 0 reserved (kInvalidFieldId)`), so valid field ids start at 1. The dedup scan mirrors the existing `findFieldId` pattern exactly.
- Deduplication is keyed on `(name, type)` so a scalar and a vec3 field with the same name remain distinct `FieldBindingId`s. This is the right semantic since the two field types live in separate storage vectors (`modifiers_` vs `modifiersVec3_`).
- The `if (name != nullptr)` guard is appropriate — null-named fields are pathological and deduplication by name is meaningless when the key is null.
- `fieldCount()` returns `m_names.size() - 1` so the dedup path (no `push_back`) keeps the count invariant intact without any special-casing.
- Three tests cover all the key behaviors: basic dedup, no-merge across types, and the World-restart re-registration scenario.

### Checklist

**ECS invariants:** No system ticks, no per-entity getComponent. Init-time registry only. ✓

**Allocations in hot paths:** Linear scan runs at init time only. Registry is capped at tens of entries in practice. ✓

**Math primitives:** `std::strcmp` and `std::size_t` are string/memory utilities, not the math functions covered by the IRMath rule. ✓

**Naming:** `m_names` / `m_types` have correct `m_` private-member prefix. ✓

**Serialized structs:** `FieldBindingId` is a stable integer already in use; no format changes. ✓

**Cross-host smoke:** Changed files are `engine/prefabs/irreden/common/modifier_field_registry.hpp` and `test/ecs/modifier_runtime_test.cpp`. Neither is in `engine/render/`, `engine/prefabs/irreden/render/`, and neither is a `.glsl` or `.metal` file. Smoke tagging not required. ✓

### Nits

- `DedupAcrossSimulatedWorldRestarts` test name slightly overpromises — a World restart reconstructs the `World` object but the `globalFieldRegistry()` static persists across reconstructions, so both registrations happen on the same `FieldRegistry` instance. The test exercises the correct dedup path; a name like `DedupOnRepeatCallsSameRegistry` would be more accurate. Non-blocking.

---

Opus recheck not required.

PR #856, opened 2026-05-18
fleet-claim: allow `*` in task titles via non-greedy header regex

### LINKED ISSUE #854: fleet-claim: title regex `[^*]+` rejects task titles containing `*` (blocks claim for following task)
## Summary

`fleet-claim`'s task-header regex `^- \[([ ~x!])\] \*\*([^*]+)\*\*` excludes `*` from the title group. Two existing task titles in `TASKS.md` contain `*`:

- T-218: `tooling: allow fleet agents to force-push claude/* branches in settings.json`
- T-221: `docs: audit role-*.md — shared protocols + point-don't-dump`

Because the inner block-walk in `edit_tasks_md` (scripts/fleet/fleet-claim:471 — pre-fix line numbers) uses the same regex to detect "next task header" boundary, it fails to recognize T-218 as a header. The walk then continues into T-218's field block, overwriting `block_id_local` with T-218's ID. After scanning past T-218's block (and possibly T-219, T-220) until a header-matching task is hit, the final `block_id_local` is wrong and `block_id_local == task_id` comparison fails.

## Reproduction

```bash
fleet-claim claim "T-217" opus-worker-1
# → "fleet-claim: T-217 not found in master/TASKS.md"
```

T-217 is present in `TASKS.md` with `[ ]` status and `Owner: free`, but it's adjacent to T-218 (which has `*` in its title) so the walk corrupts.

The duplicate regex at `scripts/fleet/fleet-claim:1048` (in `cmd_reclaim`) has the same bug.

## Impact

- Blocks claiming T-217 indefinitely
- Will block claiming any future task whose immediately-following task title contains `*`
- The error message is misleading ("not found") — the task IS in TASKS.md

## Fix

Two occurrences of `\*\*([^*]+)\*\*` in `scripts/fleet/fleet-claim` change to `\*\*(.+?)\*\*`. Non-greedy match handles `*` in titles correctly without ambiguity. Other fleet scripts (`fleet-tasks-render`, `fleet-state-scout`) already use the `(.+?)` pattern.

## Discovered

opus-worker-1 picked T-217 from the cache and tried to claim. Investigation traced to the regex.

jakildev:
## Summary

- `fleet-claim`'s task-header regex `^- \[([ ~x!])\] \*\*([^*]+)\*\*` excluded `*` from the title group, so the inner block walk in `edit_tasks_md` failed to recognize T-218's header (`tooling: allow fleet agents to force-push claude/* branches in settings.json`) as a task boundary.
- Symptom: `fleet-claim claim "T-217" opus-worker-1` returned `T-217 not found in master/TASKS.md` even though T-217 is present and `Owner: free` — the walk past T-217's fields kept reading, overwriting `block_id_local` with T-218's (and beyond) ID, so the final equality check failed.
- Two regex occurrences (`edit_tasks_md` and the `cmd_reclaim` inner block) changed to `\*\*(.+?)\*\*`. Sibling fleet scripts (`fleet-tasks-render`, `fleet-state-scout`) already use the permissive form — this aligns `fleet-claim` with them.

## Verification

Ran an equivalent block-walk in python over the current `origin/master:TASKS.md` with the patched regex:

- All 56 task IDs parse cleanly, no duplicates
- T-215, T-217, T-218, T-219, T-220, T-221, T-222 all individually findable
- Block boundaries correctly land at each task header (T-217 ends at line 304, T-218 starts at line 305)

## Test plan

- [ ] `fleet-claim claim "T-217" opus-worker-1` from a fresh worktree succeeds (or correctly reports race-lost if already claimed)
- [ ] `fleet-claim claim "T-218" opus-worker-2` still works (was already working pre-fix because T-218 is followed by T-219 whose header is clean)
- [ ] `fleet-claim reclaim "T-217"` correctly inspects state via the patched `cmd_reclaim` regex

Closes #854

### DIFF
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -621,7 +621,7 @@ def edit_tasks_md(content, action, task_id, owner):
     found = False
     decision = None
 
-    header_re = re.compile(r'^- \[([ ~x!])\] \*\*([^*]+)\*\*(.*)$')
+    header_re = re.compile(r'^- \[([ ~x!])\] \*\*(.+?)\*\*(.*)$')
     field_re = re.compile(r'^(\s+)- \*\*([^:]+):\*\*\s*(.*?)\s*$')
 
     n = len(lines)
@@ -1198,7 +1198,7 @@ content = os.environ.get('FLEET_RECLAIM_TASKS_MD', '')
 block_status = None
 block_id = None
 block_owner = None
-header_re = re.compile(r'^- \[([ ~x!])\] \*\*([^*]+)\*\*')
+header_re = re.compile(r'^- \[([ ~x!])\] \*\*(.+?)\*\*')
 field_re = re.compile(r'^\s+- \*\*([^:]+):\*\*\s*(.*?)\s*$')
 for line in content.splitlines():
     m = header_re.match(line)


jakildev (commented):
## Review — fleet-claim: allow `*` in task titles via non-greedy header regex

**Verdict:** approve

### Praise
- The root cause analysis is precise: `[^*]+` stopping at `*` inside a title caused `block_id_local` to be overwritten by the next task's ID, making the final equality check fail silently. The commit message explains this well.
- Correct choice of `(.+?)` (non-greedy any-char) over `([^*]+)` — non-greedy stops at `**` correctly even if `*` appears inside the title, and matches the pattern already used in `fleet-tasks-render` line 51 (verified).
- Both occurrences changed consistently: `edit_tasks_md` (line 624, the block-walk path) and the inline `cmd_reclaim` Python snippet (line 1201, the state-inspection path). Neither site was missed.
- `field_re` pattern `[^:]+` correctly left unchanged — field keys (ID, Owner, Labels) don't contain `*`, and that pattern matches `:` as the boundary, not `*`.
- The PR body includes a Python verification showing 56 task IDs parse cleanly with correct block boundaries. That's good evidence the change doesn't break existing clean-title tasks.

- [ ] `fleet-claim claim "T-218" <agent>` succeeds (T-218 title contains `claude/*` and was the original reproduction case)
- [ ] `fleet-claim reclaim "T-218"` correctly inspects state
- [ ] `fleet-claim claim "T-217" <agent>` succeeds (previously broken even though T-217 title is clean — was collateral damage from the block-walk overwrite)

Opus recheck not required.

jakildev:

— fleet merger

PR #883, opened 2026-05-18
fleet: fleet-claim master-lock parser accepts '*' in task titles

### LINKED ISSUE #882: fleet-claim: master-lock parser fails on task title with literal '*' (blocks claiming any task adjacent to one with '*' in title)
## Symptom

`fleet-claim claim "T-217" opus-worker-1` exits 2 with:
`fleet-claim: T-217 not found in master/TASKS.md`

T-217 is present in master TASKS.md (`grep -n "T-217"` returns line 262) and the scout projection lists it as `owner=free, status=" "`, but the master-lock parser inside fleet-claim cannot locate its block.

## Root cause

`scripts/fleet/fleet-claim` line 471 uses a header regex that disallows literal `*` characters inside task titles:

```python
header_re = re.compile(r'^- \[([ ~x!])\] \*\*([^*]+)\*\*(.*)$')
```

The task immediately after T-217 in TASKS.md is T-221, whose title is `docs: audit role-*.md — shared protocols + point-don't-dump`. The `*` in `role-*.md` causes T-221's header line to fail `header_re` entirely (the regex requires `[^*]+` between the `**`...`**` delimiters).

When the parser is inside T-217's block-scan and reaches T-221's header, it doesn't recognize it as a header → keeps scanning forward → picks up T-221's `**ID:** T-221` field line → overwrites block_id_local. The result: T-217's block is internally tagged as ID T-221, and the actual T-221 block is silently absorbed. fleet-claim therefore cannot find a block with ID T-217.

Reproduced manually with a Python script that mirrors the parser logic; the loop processed 38 task blocks, and `T-217` was missing from the seen-IDs list while `T-221` appeared at line 261 (where T-217's header actually is).

## Why it didn't trigger earlier

This bug only fires when a task title containing `*` (e.g. `role-*.md`, `Shape*.glsl`, glob-style file references) is the next entry after the task being claimed. Most titles use plain prose without metacharacters. T-221's `role-*.md` is the first such title in the open queue.

## Other parsers in same file are correct

Line 192 (the blocked-by parser):
```python
m = re.match(r'^- \[(.)\] \*\*(.+?)\*\*', line)
```
Uses lazy `.+?` — handles asterisks fine.

Line 358 (model-tag parser): also uses `.+\*\*` — correct.

Only the master-lock parser at line 471 (and the `reclaim` subcommand parser around line 1073) use the buggy `[^*]+` pattern.

## Suggested fix

Replace `[^*]+` with `.+?` (lazy) on both lines, matching the existing parsers:

```python
header_re = re.compile(r'^- \[([ ~x!])\] \*\*(.+?)\*\*(.*)$')
```

This is one-character-class-substitution per site (2 sites). No behavioural change for titles without `*`. After fix, T-217 (and any future task adjacent to a `*`-bearing title) claims cleanly.

## Impact

- T-217 (`fleet: resolve PR #767 design decisions + rebase cross-machine claim layer`, `[opus]`, `Owner: free`) is currently unclaimable on master.
- Any future task added adjacent to a title with `*` would hit the same issue.
- Suggested model: `[sonnet]` (one-line surgical fix in a single file, no design judgment required).
- Suggested area: `scripts/fleet/`

Discovered while attempting to claim T-217 from opus-worker-1 this iteration.

jakildev:
## Summary
- `scripts/fleet/fleet-claim` lines 624 and 1201 used `[^*]+` between the `**...**` title delimiters, which fails on titles containing a literal `*` (e.g. T-221's `role-*.md`).
- When the master-lock parser was scanning T-217's block, T-221's header line at line 274 of `TASKS.md` failed the header regex → the inner block-scan walked past T-221's header and absorbed T-221's `**ID:** T-221` field line. T-217 was then effectively missing from the parsed task map.
- Result: `fleet-claim claim "T-217" opus-worker-1` exited 2 with `T-217 not found in master/TASKS.md` even though T-217 is present, free, and unblocked.
- Replace `[^*]+` with `.+?` (lazy match), matching the sibling parsers at lines 192 (blocked-by) and 1077 (cleanup) that have used the correct form all along.

## Root cause writeup
Issue #882 has the full reproduction + trace.

## Test plan
- [x] Direct parser-logic exercise against current `origin/master` TASKS.md sees 39 task IDs (was 38); T-217 and T-221 both visible.
- [ ] `fleet-claim claim "T-217" opus-worker-1` on a fresh worktree should succeed (no longer "not found").
- [ ] No behaviour change on any task whose title contains no `*` (the lazy form is equivalent to the negated-class form when the delimiter doesn't appear in the title).

## Notes for reviewer
Two surgical lines. The sibling parsers in the same file (192, 1077) already use `.+?`, so this is parity cleanup rather than a new pattern choice. No test infrastructure exists for the parsers; if one is desired, a follow-up issue could add a `tests/fleet/test_fleet_claim_parser.py` with a couple of `*`-bearing-title fixtures.

Closes #882.

### DIFF
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -621,7 +621,7 @@ def edit_tasks_md(content, action, task_id, owner):
     found = False
     decision = None
 
-    header_re = re.compile(r'^- \[([ ~x!])\] \*\*([^*]+)\*\*(.*)$')
+    header_re = re.compile(r'^- \[([ ~x!])\] \*\*(.+?)\*\*(.*)$')
     field_re = re.compile(r'^(\s+)- \*\*([^:]+):\*\*\s*(.*?)\s*$')
 
     n = len(lines)
@@ -1198,7 +1198,7 @@ content = os.environ.get('FLEET_RECLAIM_TASKS_MD', '')
 block_status = None
 block_id = None
 block_owner = None
-header_re = re.compile(r'^- \[([ ~x!])\] \*\*([^*]+)\*\*')
+header_re = re.compile(r'^- \[([ ~x!])\] \*\*(.+?)\*\*')
 field_re = re.compile(r'^\s+- \*\*([^:]+):\*\*\s*(.*?)\s*$')
 for line in content.splitlines():
     m = header_re.match(line)


jakildev (commented):
## Review — fleet: fleet-claim master-lock parser accepts '*' in task titles

**Verdict: approve**

### Summary

Two surgical regex fixes — no issues.

### Correctness

**Completeness of fix:** `git show origin/master:scripts/fleet/fleet-claim | grep -n '\[^\*\]+'` returns exactly two results, at lines 624 and 1201. Both are fixed in this PR. No unfixed instances remain.

**Sibling parser parity:** Verified on master — lines 345 and 1077 already use `.+?`, confirming `.+?` is the established pattern in this file for title matching.

**Regex correctness:** `.+?` is correct. `[^*]+` excludes the `*` character entirely; `.+?` matches any character lazily, stopping at the first `**` delimiter. The lazy form is identical in behavior for titles without `*` and handles `*`-bearing titles correctly.

**Root cause trace:** The PR description's walk-through (T-221's `role-*.md` title failing the header regex, causing the block-scanner to misclassify T-221's field lines as part of T-217's block) is a sound read of the regex behavior.

### Minor note (no action needed)

The PR description cites the blocked-by sibling parser at "line 192" — it's at line 345 on master. Minor description drift, irrelevant post-merge.

Opus recheck not required.

PR #896, opened 2026-05-18
fleet: fleet-up --reset-usage flag for account-switch recovery

jakildev:
## Summary
- Add `fleet-up --reset-usage` flag that wipes `~/.fleet/state/usage/*.json` before the dispatcher reads it, so an operator who switched Claude accounts can recover from the prior account's stale rate-limit observations without waiting on `FLEET_DISPATCHER_USAGE_STALE_SECONDS` (default 1h) to age them out.
- Parsed in an early-args pass so the flag works on a live fleet too — the dispatcher re-evaluates the gate on its next 10s tick, no restart needed. Cached observations re-populate organically once the new account's events flow through `fleet-claude-stream`.
- One-line doc bullet in `FLEET.md` "Fleet-wide usage gate" pointing at the new flag.

## Motivation
Hit live this session: the operator switched accounts and the fleet stayed gated for hours because `~/.fleet/state/usage/five_hour.json` still showed the prior account's `utilization: 0.94`. Manually `rm`-ing the file immediately reopened the gate and the dispatcher fired every role within seconds.

## Test plan
- [x] `bash -n` syntax check on the modified script.
- [x] `fleet-up --reset-usage` on a populated cache wipes the files and prints `wiped N cached observation(s)`.
- [x] Idempotent: second call on the now-empty dir prints `already empty` and exits cleanly.
- [x] Composes with the existing modes / flags via the early-parse + late-parse split (`--reset-usage` accepted by both parsers).
- [x] End-to-end: cache wipe (done by hand a few minutes before this PR was authored) immediately reopened a closed gate (`usage gate re-opened (open); resuming dispatch` in dispatcher.log) and every role fired within seconds.
- [x] Safe on a live fleet — no dispatcher restart needed; gate re-evaluates on the next 10s tick.

## Notes for reviewer
- Early-parse loop is intentionally a second pass (not folded into the existing MODE parser) because the wipe needs to happen before the `tmux has-session` early-exit, and the MODE parser sits well below that point. The late parser accepts `--reset-usage) ;;` as a no-op so the unknown-arg guard doesn't reject it.
- Pattern of "iterate files, count, log summary" mirrors the existing "sweep stale triggers" block (`scripts/fleet/fleet-up:648-659`). Two uses isn't enough to abstract a helper.

### DIFF
--- a/docs/agents/FLEET.md
+++ b/docs/agents/FLEET.md
@@ -414,7 +414,11 @@ Defaults and tuning:
 - Global override (any unmatched type): `FLEET_DISPATCHER_USAGE_GATE`.
 - Stale observations (older than `FLEET_DISPATCHER_USAGE_STALE_SECONDS`,
   default `3600`) are dropped from evaluation, so a single high-watermark
-  reading can't latch the gate closed forever.
+  reading can't latch the gate closed forever. **Account-switch recovery:**
+  switching Claude accounts leaves the prior account's observations cached
+  and they will gate the new account until the stale window ages out.
+  `fleet-up --reset-usage` wipes `~/.fleet/state/usage/*.json` immediately;
+  safe to run on a live fleet (the gate re-evaluates on the next tick).
 - Reset grace (`FLEET_DISPATCHER_RESET_GRACE_SECONDS`, default `600`) —
   observations stay in evaluation until `resetsAt + grace` is in the past.
   Set to `0` to revert to the old "open instantly at resetsAt" behavior.
--- a/scripts/fleet/fleet-up
+++ b/scripts/fleet/fleet-up
@@ -125,6 +125,51 @@ if ! command -v claude >/dev/null 2>&1; then
     exit 1
 fi
 
+# --- Early flag parsing -----------------------------------------------------
+#
+# Flags handled here take effect before the early-exit-if-session-exists
+# check, so they work on both fresh boots and live fleets. The main arg
+# parser further down recognizes the same flags as no-ops so the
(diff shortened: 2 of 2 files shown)


jakildev (commented):
## Review — fleet: fleet-up --reset-usage flag for account-switch recovery (PR #896)

**Verdict: approve**

### Summary

Clean, well-scoped feature for a real operational pain point. The implementation is correct and the design decisions are sound.

### Code review

**Early-parse / late-parse split** — rationale is solid. The wipe must precede the `tmux has-session` early-exit so it works on live fleets; the main MODE parser sits well below that. Late-parser no-op `--reset-usage) ;;` prevents the unknown-arg guard from firing. Both parsers are correctly wired.

**Glob safety** — `[[ -e "$f" ]] || continue` correctly handles the case where `*.json` expands to the literal string in an empty dir. Without this guard, `rm -f` would silently try to delete a non-existent literal path and the count would stay zero — the "already empty" branch does the right thing instead.

**Idempotency** — wipe then empty-check behaves correctly on both first and subsequent invocations. Confirmed in test plan.

**Comment quality** — the block comment above `if (( RESET_USAGE ))` is non-obvious WHY content (latching behavior, stale window, safety guarantee) — exactly the right bar for inline documentation.

**FLEET.md change** — clear, one-paragraph note in the right section. Points to the new flag and states the safety guarantee.

No behavioral concerns with wiping all `*.json` on an account switch. The flag is operator-triggered, the use case is "just switched accounts", and organic repopulation is fast.

Opus recheck not required.

PR #932, opened 2026-05-18
T-213: editor F-1.3 — layer system panel UI

jakildev:
## Summary

- Adds a **LAYERS panel** widget at `{130, 240}` (second column alongside PALETTE) with a layer list, visibility checkbox, and +/- add/delete buttons.
- **Bi-directional sync**: keyboard shortcuts K/[/]/H update the widget state; clicking the checkbox calls `toggleLayerVisibility`; clicking +/- calls `addLayer`/`deleteLayer` on `EditorLayerManager`.
- **`applyLayerVisibility()`**: walks `C_VoxelSetNew.voxels_` and activates/deactivates voxels by `layer_id_`. Hidden layers are also automatically unpickable — `picking.hpp:287` already skips `alpha == 0` voxels, so no engine changes needed.
- **`applyEdit()`**: assigns `layer_id_` on voxel placement; keeps voxels deactivated when the active layer is hidden.
- Registers `WIDGET_APPLY_LIST`, `WIDGET_APPLY_CHECKBOX`, `WIDGET_RENDER_BUTTON`, `WIDGET_RENDER_CHECKBOX`, `WIDGET_RENDER_LIST` in the INPUT/RENDER pipelines.

## Test plan

- [ ] Build passes: `fleet-build --target IRVoxelEditor`
- [ ] Editor starts without crash (15s smoke: `fleet-run --timeout 15 IRVoxelEditor`)
- [ ] LAYERS panel visible at startup alongside PALETTE panel
- [ ] `[`/`]` keys cycle active layer; list selection updates
- [ ] `H` key toggles layer visibility; checkbox reflects new state
- [ ] Clicking checkbox toggles visibility; voxels on that layer appear/disappear
- [ ] `+` button adds a new layer; `-` button deletes active layer
- [ ] Voxels placed on a hidden layer stay invisible
- [ ] Undo restores correct `layer_id_` on erased voxels

### DIFF
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -60,7 +60,12 @@
 #include <irreden/render/systems/system_widget_render_label.hpp>
 #include <irreden/render/systems/system_widget_render_color_swatch.hpp>
 #include <irreden/render/systems/system_widget_apply_slider.hpp>
+#include <irreden/render/systems/system_widget_apply_list.hpp>
+#include <irreden/render/systems/system_widget_apply_checkbox.hpp>
 #include <irreden/render/systems/system_widget_render_slider.hpp>
+#include <irreden/render/systems/system_widget_render_list.hpp>
+#include <irreden/render/systems/system_widget_render_checkbox.hpp>
+#include <irreden/render/systems/system_widget_render_button.hpp>
 
 // Camera prefab namespace (Z-yaw API)
 #include <irreden/render/camera.hpp>
@@ -228,6 +233,13 @@ IREntity::EntityId g_sceneVoxelSetEntity = IREntity::kNullEntity;
 IREntity::EntityId g_scrubberSlider = IREntity::kNullEntity;
 IREntity::EntityId g_fpsSlider = IREntity::kNullEntity;
 
+// Layer panel widget entity IDs (T-213). Initialized in initEntities.
+IREntity::EntityId g_layerPanel = IREntity::kNullEntity;
+IREntity::EntityId g_layerList = IREntity::kNullEntity;
+IREntity::EntityId g_layerVisCheckbox = IREntity::kNullEntity;
+IREntity::EntityId g_layerAddBtn = IREntity::kNullEntity;
+IREntity::EntityId g_layerDelBtn = IREntity::kNullEntity;
+
 void logLayerState() {
     for (const auto &r : g_layerManager.layers()) {
         IR_LOG_INFO(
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — PR #932: T-213: editor F-1.3 — layer system panel UI

**Verdict:** approve

### Summary

Adds a LAYERS panel to the voxel editor with a list, visibility checkbox, and add/delete buttons. Bi-directional sync between keyboard shortcuts (K/[/]/H) and the widget state. New `applyLayerVisibility()` helper walks `C_VoxelSetNew.voxels_` and activates/deactivates by `layer_id_`.

**ECS correctness**: `layerSyncSystem` is created with `createSystem<C_GuiElement>` where the per-entity tick is empty and all logic runs in `endTick` — the correct pattern for once-per-frame panel sync. No per-entity `getComponent` in the tick body.

**Widget guard expanded correctly**: `overWidget` check in `placeEraseSystem` now includes `isHovered(g_layerPanel)`, preventing voxel placement through the layer panel. The lambda captures the global at execution time (not definition time), so it reads the valid entity ID set by `initEntities()` before the main loop starts.

**H-key command now calls `applyLayerVisibility`**: Previous implementation toggled the manager state but didn't update voxel alpha — clicks via checkbox would make voxels disappear but H would not. The fix makes both code paths consistent.

**Layer deletion migration**: When deleting a layer, voxels are migrated to layer 0 with visibility applied based on `g_layerManager.isVisible(0)` — correctly accounts for the default layer's own visibility state.

**`g_sceneVoxelSetEntity` assignment**: Previously unset (left at `kNullEntity`), now assigned in `initEntities()` before `applyLayerVisibility` can be called. The null-guard in `applyLayerVisibility` covers the transient gap.

No ECS invariant violations, no ownership concerns, no render pipeline changes, no naming violations.

### Nits

- `main.cpp:770–774` (list rebuild section in `layerSyncSystem` endTick): `list.items_[i] = updatedLayers[i].name_ + (...)` allocates a new `std::string` every frame for every layer. For the editor frame budget this is negligible, but a simple equality check before assignment (`if (list.items_[i] != expected) list.items_[i] = expected`) would avoid the allocation on clean frames.

### Cross-host smoke tagging

Diff touches only `creations/editors/voxel_editor/main.cpp` — no render paths, no `.glsl`/`.metal` files. No smoke labels required.

Opus recheck not required.

jakildev:
Addressed fleet:has-nits feedback: layer panel rebuild now compares the existing string against the expected name+suffix, so a clean frame skips the per-layer std::string concatenation alloc. Also pulled in clang-format-changed pass over the rest of the T-213 touched code for style consistency. Build clean (`IRVoxelEditor` target).

— opus-worker-1

PR #401, opened 2026-05-01
T-066: render/system: centralize GPU stage probes via SystemManager TickObserver

jakildev:
## Summary
- Replace 14 inlined `device()->finish()` GPU-timing probes scattered across 11 render system tick lambdas with a generic `IRSystem::TickObserver` hook + `GpuStageTimingObserver` that plugs into it.
- Render systems now end `create()` with `IRRender::tagGpuStage(s, "<name>")`. The registry in `gpu_stage_timing.hpp` is the single source of truth for the `field_` pointer and `budgetShare_`.
- Fixes the multi-canvas overwrite bug for COMPUTE_SUN_SHADOW / COMPUTE_VOXEL_AO / LIGHTING_TO_TRIXEL (`=` per-entity write previously dropped to last canvas).
- Removes the `+=` accumulator + `beginTick` zero-reset ceremony from trixelToTrixel / trixelToFb / fbToScreen.

## Trade-off — multi-stage system collapse
`VOXEL_TO_TRIXEL_STAGE_1` and `SHAPES_TO_TRIXEL` each used to write multiple sub-stage timings (canvasClear/voxelCompact/voxelStage1 and shapePass0/shapePass1). After T-066 each system reports one per-system measurement. The previously sub-staged registry slots (`canvasClear`, `voxelCompact`, `shapePass0`) read `0.0f` to keep the 15-row API surface stable; documented in `gpu_stage_timing.hpp`. `shapeCompact` continues to read 0.0f as it always did.

The Opus-recheck question for review: is "same 15 rows, same field semantics" satisfied by keeping the registry intact with collapsed slots reading 0, or should multi-stage systems be split into multiple `SystemId`s to preserve sub-stage observability? I went with the smaller change; happy to split if review prefers.

## Layering
- `TickObserver` lives in `engine/system/` and stays graphics-agnostic.
- `GpuStageTimingObserver` lives in `engine/prefabs/irreden/render/` and owns the GPU-side bracket.
- Observer fires sit OUTSIDE the CPU timing window in `executeSystem`, so a GPU `finish()` doesn't get billed against `m_timingAccum`.
- Lazy install: first `tagGpuStage` call creates the observer and registers it with `SystemManager` via `registerTickObserver`. The observer lives as long as the manager.

## Test plan
- [x] `fleet-build --target IRShapeDebug` clean on macos-debug
- [x] `fleet-build --target IrredenEngineTest` clean; 367/367 tests pass
[...]

PR #237, opened 2026-04-20
render: GPU timer query infrastructure Part 1 (T-028)

jakildev:
## Summary
- Add per-pass GPU stage timing infrastructure for all 15 render pipeline stages
- Expose pass timings to Lua via `ir.render.getPassTimings()` / `getPassTiming()` / `getFrameTimeBudgetMs()` / `setGpuTimingEnabled()` / `isGpuTimingEnabled()`
- Per-pass soft budgets reflect the relative weight of each pass as a fraction of the 16.67 ms frame target; passes over their share are flagged via `overBudget`
- End-of-run profile report's GPU section now iterates the same registry (single source of truth) — fixes prior hand-rolled list that only reported 10 of 15 stages
- Probes are gated on `timing.enabled_` (default off via the `gpu_stage_timing` config key) so release runs pay only a load-and-branch

## Scope
Part 1 of GPU timer query infrastructure. Uses `device()->finish()` + `std::chrono::steady_clock` samples, matching the existing pattern from earlier work. Real async `GL_TIMESTAMP` / Metal counter-sample queries will come in a later Part.

Newly instrumented passes: `computeSunShadow`, `computeVoxelAO`, `textToTrixel`, `lightingToTrixel`, `trixelToTrixel`, `fbToScreen`.

## Test plan
- [x] `fleet-build --target IRShapeDebug` clean (macOS/Metal)
- [x] `fleet-run --timeout 8 IRShapeDebug` runs without crashing, no visual changes (probes off by default)
- [ ] Reviewer: enable `gpu_stage_timing = true` in a config and verify Lua `ir.render.getPassTimings()` returns 15 rows with sensible ms / budget values
- [ ] Reviewer: confirm profile report's GPU stage section now lists all 15 stages (not 10)

## Notes for reviewer
- Bindings live inline in `World::setupLuaBindings` rather than via the `LuaBindingRegistration` callback pattern — this avoids adding `IrredenEngineScript` to render's CMake deps (cross-module coupling). Factoring into `World::registerEngineRenderLuaBindings` is an optional follow-up if you'd prefer the helper shape.
- `gpu_stage_timing.hpp` uses pointer-to-member (`float GpuStageTiming::*`) to let the registry describe both name and destination field without macros.
- `optimize` skill (section 3 + worked example) updated to point at the new Lua API as the primary workflow.

Closes #173

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #977, opened 2026-05-18
T-275: render IRProfile ScopeTimer + per-stage CPU timing (B0)

### LINKED ISSUE #939: render: IRProfile::ScopeTimer + GPU timer query infrastructure (B0)
## Scope

CPU scope-timer macro `IR_PROFILE_SCOPE(\"name\")` emitting per-frame histograms; opt-in via build flag. GPU timer-query pool around each render pipeline stage; results read 1 frame later to avoid sync stall. Integrate display into `creations/demos/perf_grid/main.cpp`.

**Blocks verification** for B1, B2, B5 (and indirectly E2). Round-1 foundation task.

## Parent epic

#935 (Epic B — Voxel storage & rendering optimizations)

## Stack

Base of Stack S-B-render (B0 → B1 → B2 → B5). Stacked sub-tasks branch from this PR's head, not master.

## Dependencies

(none — foundation task)

## Acceptance

- (1) `IRProfile::ScopeTimer` macro emits per-frame histogram readable from a debug HUD.
- (2) GPU timer queries instrument each render pipeline stage (init, voxel→trixel stage 1/2, shapes→trixel, visibility compaction, lighting, framebuffer composite); results lagged 1 frame.
- (3) `creations/demos/perf_grid` HUD displays per-stage CPU + GPU ms.
- (4) Golden screenshot of profiler overlay (regression catches future overlay breakage).
- (5) `fleet-build` clean on `linux-debug` and `macos-debug`.

## Plan reference

`.claude/plans/okay-lets-go-through-idempotent-giraffe.md` §"Epic B → B0".

jakildev:
## Summary

Adds the missing CPU half of T-275. GPU per-stage timer queries shipped via T-028 (PR #237) + T-066 (PR #401) — the `GpuStageTimingObserver` already brackets every tagged render system with backend timestamp writes and lags readback by `kSamplesInFlight` frames. What was missing was the CPU mirror: a generic scope-timer macro recording into a per-frame histogram the HUD can read every frame.

- New `IR_PROFILE_SCOPE(name)` RAII macro + `CpuFrameHistogram` singleton in `engine/profile/include/irreden/profile/scope_timer.hpp`. Heterogeneous `string_view` lookup keeps the hot path allocation-free after warmup; ScopeTimer skips `chrono::now()` entirely when disabled.
- `World::input/update/render` get top-line `IR_PROFILE_SCOPE` brackets. `World::render()` calls `cpuFrameHistogram().endFrame()` for frame-aligned current → last swap.
- `GpuStageTimingObserver::onBeforeTick/onAfterTick` now also stamp each tagged stage into the CPU histogram via the same name → field registry — no new observer, no per-system edits, no overhead when disabled.
- `PERF_STATS_OVERLAY` gains CPU+GPU / CPU-only / GPU-only branches. The pre-existing GPU-only and FPS-fallback branches are preserved unchanged for lighting / default demos.
- `perf_grid` adopts the overlay (+ `TEXT_TO_TRIXEL` → `TRIXEL_TO_FRAMEBUFFER` tail), turns both timing surfaces on at startup, and gains a `profiler_overlay` shot.
- Lua surface: `ir.render.setCpuTimingEnabled`, `getCpuPassTimings`, `getCpuPassTiming` mirror the existing GPU surface.

## Profiler overlay (macOS)

profiler overlay

Top-line `CPU I:0.01 U:11.75 R:14.05` is from `World::input/update/render`. Per-stage CPU values (`CPU CMP / V1 / V2 / S1 / OCC / AO / LV / LIT`) come through the observer.

## Deferred

- **Metal GPU readback shows 0.0** in the screenshot — pre-existing Metal-backend behaviour with `recommendedTimestampPairsInFlight = 1`; the CPU side captures real values. Not introduced by this PR; filed separately if reproducible on a Linux/OpenGL host.
- **`profiler_overlay` render-verify manifest** — captured shot lives at `docs/pr-screenshots/`. Wiring the perf_grid demo into `test/references/<backend>/` with loose thresholds (timing-value drift across frames) is its own follow-up.

## Test plan

- [x] `fleet-build --target IRPerfGrid` clean (macos-debug)
- [x] `fleet-build --target IRShapeDebug` clean (no regressions)
- [x] `fleet-build --target IRLightingPerCanvasScope` clean (lighting demo overlay path unchanged when CPU disabled)
- [x] `fleet-build --target IrredenEngineTest` clean
- [x] `fleet-run IRPerfGrid --auto-screenshot 30` captures the three shots; profiler_overlay shows live CPU+GPU per-stage ms

Closes #939


jakildev (commented):
## Review — PR #977: T-275: render IRProfile ScopeTimer + per-stage CPU timing (B0)

**Verdict:** approve

### Summary

Adds `IR_PROFILE_SCOPE(name)` RAII macro + `CpuFrameHistogram` singleton, wires the top-line `World::input/update/render` brackets, extends `GpuStageTimingObserver` to record CPU-side timings alongside GPU queries, and adds CPU+GPU / CPU-only / GPU-only branches to `PERF_STATS_OVERLAY`. All test plan items pass on macOS. Metal GPU readback showing 0.0 is pre-existing behaviour, acknowledged in the Deferred section.

### ECS correctness

No per-entity `getComponent` in tick bodies. The observer's `onBeforeTick`/`onAfterTick` fire once per stage (keyed by `SystemId`) — not per-entity. The `endFrame()` call in `World::render()` is a single post-frame maintenance call, not in a tick path. The `PERF_STATS_OVERLAY` tick function calls `lastFrameMs()` (const, O(1) lookup) on fixed stage names — acceptable HUD overhead.

### `CpuFrameHistogram` design

`endFrame()` swap + reset is correct: after the swap, `m_lastFrame` holds this frame's complete data; `m_current` is reset for the next frame. The `m_current` map keys persist across swaps so there are no new allocations after warmup — verified: C++23 is the engine standard (`cmake -D CMAKE_CXX_STANDARD 23`), so `unordered_map::find(string_view)` uses the transparent overload without allocation.

**One-frame lag on "render" top-line** (design note, not a bug): `IR_PROFILE_SCOPE("render")` is declared at function entry; its destructor fires *after* `endFrame()`, so "render" records into `m_current` of the next frame and appears in `m_lastFrame` one frame later. `input` and `update` record before `endFrame()`, so they are same-frame. The HUD shows last-frame render time — correct for a realtime HUD. This matches the PR description ("stable per-frame snapshot").

### `GpuStageTimingObserver` changes

The refactoring correctly separates the GPU enable-gate from the CPU path:
- `cpuActive_` is set/cleared unconditionally relative to `cpuFrameHistogram().enabled_` — correct; CPU timing is independent of GPU support.
- The `else { cpuActive_ = false; }` branch in `onBeforeTick` handles the edge case where `enabled_` toggles between frames (a stale `cpuActive_ = true` from a previous tick would be cleared before the next `onAfterTick`). Defensive and correct.
- The `if (it == m_stages.end()) return;` guard is now the first check in both callbacks — unregistered stages skip both CPU and GPU paths. Correct.

### Stage name mapping

All eight `lastFrameMs("...")` lookups in `system_perf_stats_overlay.hpp` match the exact strings in `gpuStageRegistry()` in `gpu_stage_timing.hpp` (verified: `voxelCompact`, `voxelStage1`, `voxelStage2`, `shapePass1`, `buildLightOcclusionGrid`, `computeVoxelAO`, `computeLightVolume`, `lightingToTrixel`).

### Buffer size (`system_perf_stats_overlay.hpp`)

`std::string buf(1024, '\0')` — the CPU+GPU format string expands to roughly 600-700 characters at peak. 1024 is sufficient. `snprintf` truncates safely without overflow.

### Lua bindings

`setCpuTimingEnabled`, `getCpuPassTimings`, `getCpuPassTiming` mirror the existing GPU surface symmetrically. `getCpuPassTimings` iterates `lastFrame()` (const ref) and builds a Lua table — no side effects. Correct.

### Nits

- `scope_timer.hpp:108` — `ScopeTimer::m_name` stores a `std::string_view` whose pointed-to buffer must outlive the `ScopeTimer` instance. All current call sites use string literals (static storage duration — safe). A brief inline note on the field (e.g. `// must point to static storage — use only via IR_PROFILE_SCOPE with a literal`) would protect a future direct caller from a dangling reference.
- `engine/profile/CLAUDE.md` (new section) — mentions `ir.render.setCpuTimingEnabled`, `getCpuPassTimings`, `getCpuPassTiming` in the prose. The Lua surface entry point is `ir.render` but it's wired in `world.cpp` via `render["setCpuTimingEnabled"]`. The doc is accurate as-is, but cross-referencing `world.cpp:setupLuaBindings` for where these are registered would help a reader who searches for the binding site.

### Cross-host smoke tagging

Diff touches `engine/prefabs/irreden/render/gpu_stage_timing_observer.hpp` and `engine/prefabs/irreden/render/systems/system_perf_stats_overlay.hpp`. Author is on macOS (`fleet:authored-on-macos`). `fleet:needs-linux-smoke` should be added after this review.

Opus recheck required: PR touches `engine/prefabs/irreden/render/` (observer pattern wired into the render pipeline's tick lifecycle) and `engine/world/src/world.cpp` (top-level game loop brackets). The CPU timing is single-threaded and the design is sound, but the interaction between `endFrame()` placement and the observer's `cpuActive_` state during the render loop warrants a second-pass confirmation that the timing window is correct across all pipeline orderings.

jakildev (commented):
## Opus recheck — PR #977: T-275: render IRProfile ScopeTimer + per-stage CPU timing (B0)

**Verdict:** approve

Sonnet's first-pass review is correct and thorough. Focused this pass on
the item Sonnet explicitly flagged for Opus recheck (the `endFrame()`
placement vs the observer's `cpuActive_` state across pipeline
orderings) plus the lifecycle / hot-path questions an observer wired
into the engine tick raises that Sonnet's checklist doesn't already
cover.

### Sonnet flagged endFrame ↔ cpuActive_ interaction → confirmed safe

Walked the full lifecycle across INPUT / UPDATE / RENDER and the
observer's onBefore / onAfter pairs:

- The observer's `cpuActive_` is a **per-stage** field on `StageState`,
  not an observer-wide bit. Each tagged stage's onBefore sets it and
  the matching onAfter clears it. Since `SystemManager::executePipeline`
  fires ticks serially on the main thread, the onBefore/onAfter pair
  for a single stage never interleaves with another stage's pair —
  there is no race or stomp risk
  (`engine/prefabs/irreden/render/gpu_stage_timing_observer.hpp:56-118`).
- `endFrame()` is the **last** call in `World::render()` and runs after
  the entire RENDER pipeline has completed, so every tagged stage that
  ticked this frame (in any pipeline) has already deposited its CPU
  sample into `m_current` before the swap
  (`engine/world/src/world.cpp:271-277`). No stage sample is lost to
  the rotation.
- The "render" top-line's one-frame display lag — `IR_PROFILE_SCOPE("render")`
  destructs after `endFrame()` — is real and acknowledged by the
  PR description. `input` and `update` finish their scopes before
  `endFrame()` so they show same-frame. The HUD reading
  `lastFrameMs("render")` always shows the previous frame's render
  total, which is the correct realtime-HUD shape.
- The `else { cpuActive_ = false; }` branch in `onBeforeTick` is the
  right defense: it ensures a stage whose previous tick had
  `enabled_ == true` (so `cpuActive_` was set) doesn't carry a stale
  `cpuActive_ = true` into a frame where the histogram is now
  disabled. Combined with `record()`'s own `if (!enabled_) return;`
  guard, the double-check is correctly belt-and-suspenders, not
  redundant: if `enabled_` flips off **between** onBefore and onAfter,
  `cpuActive_` is still true (locked-in at onBefore), so onAfter calls
  `record()`, which then no-ops. The captured `cpuT0_` was wasted but
  no garbage sample lands.

### Lifetime + hot-path items not in Sonnet's checklist → confirmed clean

- `ScopeTimer::m_enabled` is **captured at construction**, not re-read at
  destruction (`scope_timer.hpp:107-122`). This is correct: a scope that
  enters with timing disabled stays disabled (no `m_t0` was set, so a
  read-at-destruction would compute garbage). A scope that enters
  enabled stays enabled. The cost is one extra bool slot per
  ScopeTimer instance; well worth the invariant.
- The transparent-hash `unordered_map::find(string_view)` lookup is
  genuinely allocation-free after warmup — C++23 with
  `is_transparent` on both hash and equal, verified by reading the
  detail::TransparentStringHash / TransparentStringEqual surface
  (`scope_timer.hpp:21-54`). The only allocation in `record()` is the
  `emplace(std::string{name}, ...)` on first-lookup, and after one full
  frame of usage every stage name's node is resident.
- The keys persist across `endFrame()` swaps — `swap` exchanges the
  map storage, then the for-loop in `endFrame()` zeroes the stats but
  doesn't erase entries (`scope_timer.hpp:83-88`). So warmed-up
  histogram membership is preserved frame-to-frame; no
  re-allocation after the first per-stage hit.
- `detail::installAndGetObserver()` is a one-shot lazy singleton —
  the static pointer is initialized exactly once and never mutated
  afterward (`gpu_stage_timing_observer.hpp:176-184`). The cpp-systems
  "no function-local static for system state" rule does not apply
  because this isn't system-owned state; it's a process-bound
  singleton accessor identical in shape to `cpuFrameHistogram()` and
  `gpuStageTiming()`. Ownership is transferred to SystemManager via
  `registerTickObserver`, so observer lifetime is bounded by World.
- The legacy timing branch's observer-wide `m_t0` field
  (`gpu_stage_timing_observer.hpp:75, 107`) is **pre-existing** and
  not introduced by this PR. It works because ticks are serial; not
  ideal but not made worse here.

### One additional observation (non-blocking, follow-up)

`CpuFrameHistogram::endFrame()` iterates and zeroes `m_current`'s stats
unconditionally — even when `enabled_ == false`. In steady-state
operation this is one map walk per frame over however many keys the
histogram accumulated while enabled, which lingers if `enabled_` was
ever true and then flipped off. In practice the entry count is tiny
(O(stages) — currently ~15) and the cost is trivial. Not worth
fixing now; flagging as a tiny robustness note in case the histogram
ever grows to dozens of named scopes — a `m_current.clear()` on
disable would avoid the persistent-key overhead. Doesn't affect
this PR's verdict.

### Nits — agree with Sonnet, nothing to add

Sonnet's two nits are accurate:

- The `m_name` static-storage requirement at `scope_timer.hpp:128` is a
  real footgun for a future caller who reaches for the macro with a
  local `std::string`. A one-line comment on the field is enough.
- The `engine/profile/CLAUDE.md` Lua surface entry would benefit from
  a `world.cpp:setupLuaBindings` cross-reference.

Both are genuinely optional and fit the `fleet:has-nits` band — not
worth a needs-fix round.

### Cross-host smoke

Sonnet correctly identified that `fleet:needs-linux-smoke` should be
added for the macOS-authored render diff. PR already carries that
label.

Re-applying `fleet:approved`; `fleet:has-nits` stays so the author
worker addresses Sonnet's two nits.

Reviewed by Claude Opus 4.7 (opus-reviewer role)

Issue #974, opened 2026-05-18
fleet: project_queue_manager_ingest should filter fleet:epic to match role-doc skip rule


jakildev:
## Problem

`project_queue_manager_ingest` in `fleet-state-scout` enumerates every `human:approved AND NOT fleet:queued` issue without filtering by label. The queue-manager role doc, however, explicitly skips `fleet:epic` (and `fleet:needs-plan` / `fleet:needs-info`) entries during Step 5. This mismatch means epics sit in the trigger projection forever.

Today #667 and #934 (both `fleet:epic, human:approved`) are permanent residents of the projection set. Two side-effects:

1. **The empty-projection fast-path can't fire.** [scripts/fleet/fleet-queue-ingest:54-67](https://github.com/jakildev/IrredenEngine/blob/master/scripts/fleet/fleet-queue-ingest#L54-L67) short-circuits when `pending_count == 0`, but epics keep the count ≥ 1.
2. **Every accidental trigger costs a full LLM iteration** to discover \"nothing actionable.\" This burns Sonnet tokens and adds latency. (Today you can see it in `queue-ingest.log` — multiple iterations whose only output is \"both pending issues are filtered at Step 5.\")

This is also the reason #973's symptom persisted for ~1 hour rather than self-healing: with epics keeping the projection non-empty, scout couldn't detect a transition back to \"empty set\" to re-arm the trigger.

## Code pointers

- Projector (no label filter): [scripts/fleet/fleet-state-scout:822-846](https://github.com/jakildev/IrredenEngine/blob/master/scripts/fleet/fleet-state-scout#L822-L846)
- Role-doc skip rule: [.claude/commands/role-queue-manager.md (Step 5)](https://github.com/jakildev/IrredenEngine/blob/master/.claude/commands/role-queue-manager.md) — \"From `repos.engine.human_approved[]`, drop entries whose `labels` include any of `fleet:queued`, `fleet:needs-plan`, `fleet:needs-info`, `fleet:epic`.\"

## Proposed fix

Add the same label-filter to the projector. The `human_approved` slice already has `labels` populated (visible in `~/.fleet/state/projections/queue-manager-ingest.json`), so no extra fetch is needed:

```python
_INGEST_SKIP_LABELS = frozenset({\"fleet:queued\", \"fleet:needs-plan\", \"fleet:needs-info\", \"fleet:epic\"})

def project_queue_manager_ingest(state):
    items = []
    for repo, repo_state in state[\"repos\"].items():
[...]

PR #978, opened 2026-05-18
T-282: fleet: invalidate seen-hash on ingest lock-bail

### LINKED ISSUE #973: fleet: queue-manager-ingest trigger lost when ingest bails on lock contention
## Problem

`fleet-state-scout` writes the new `queue-manager-ingest` projection hash to `~/.fleet/state/seen-hashes/queue-manager-ingest` **before** the spawned `fleet-queue-ingest` actually does any work. If the ingest lockfile is already held by a previous LLM iteration, the new spawn exits in ~50ms with `another fleet-queue-ingest is running; exiting` — but scout has already advanced the seen-hash. The projection-change signal is silently consumed and the next scout tick reports "no triggers (all projections unchanged)."

This stranded issues #944 and #945 today: they were approved at 20:28:17/18Z while an earlier LLM iteration (started 20:27:12Z) was still running. Scout ticks at 20:28:23 / 20:29:00 / 20:30:50 each saw the projection change, wrote the new hash, spawned ingest, and the spawns all bailed on the lock. By 20:30:50 the saved hash equalled the post-#939–#943-labeled projection set `{#667, #934, #944, #945}`. When the original LLM finished at 20:32:53 the lock was free, but the hash matched reality, so scout never fired again — leaving the two issues unqueued for ~1 hour until manually unstuck via `rm ~/.fleet/state/seen-hashes/queue-manager-ingest`.

A long LLM iteration (e.g. one stretched by a rate-limit pause) widens this lock-contention window, making the race more likely.

## Code pointers

- Race write: [scripts/fleet/fleet-state-scout:1135-1149](https://github.com/jakildev/IrredenEngine/blob/master/scripts/fleet/fleet-state-scout#L1135-L1149) (`write_atomic(seen_file, ...)` followed by `subprocess.Popen([ingest_cmd], ...)` — Popen returns success the instant the bash wrapper starts, even if it exits on lock contention immediately after)
- Lock-bailed exit (no signal back to scout): [scripts/fleet/fleet-queue-ingest:44-47](https://github.com/jakildev/IrredenEngine/blob/master/scripts/fleet/fleet-queue-ingest#L44-L47)

## Proposed fix (minimal)

On the lock-bailed branch in `fleet-queue-ingest`, invalidate the seen-hash so the next scout tick re-fires:

```bash
if ! mkdir "$LOCK_DIR" 2>/dev/null; then
    log "another fleet-queue-ingest is running; exiting"
    rm -f "$HOME/.fleet/state/seen-hashes/queue-manager-ingest"
    exit 0
fi
```

This keeps the trigger "armed" until an ingest actually runs to completion. Idempotent and safe — the worst case is one extra LLM iteration that finds the same set already-labeled and exits.

## Alternative (bigger refactor)

Move hash ownership to the ingest script: scout no longer writes the seen-hash; the ingest script writes it on successful completion (or just before releasing the lock). Cleaner separation but touches both files and changes the trigger contract.

## Acceptance

- A `human:approved` label added to an issue while another `fleet-queue-ingest` is in flight gets picked up by the next scout tick after the in-flight iteration completes.
- Add a regression note to `scripts/fleet/fleet-queue-ingest` or `docs/agents/FLEET.md` explaining why the hash invalidation is necessary.

## Related

- Sister issue: `project_queue_manager_ingest` does not filter `fleet:epic`, which prolongs the symptom (the projection never empties so the fast-path can't recover the trigger).

jakildev:
## Summary

- On lock-bail, `fleet-queue-ingest` now deletes the `queue-manager-ingest` seen-hash so the next scout tick re-fires the trigger
- Without this fix, scout writes the seen-hash _before_ the spawned process returns; a lock-bailed spawn silently consumed the projection-change signal, stranding `human:approved` issues until the hash was deleted manually
- Added an inline comment explaining the invariant so the fix is self-documenting (regression note per acceptance criteria)
- Also deployed the fix to `~/bin/fleet-queue-ingest` on the host so it takes effect immediately

## Test plan

- [ ] Approve a new issue while `fleet-queue-ingest` is in-flight (lock is held) — the next scout tick after the in-flight iteration completes should spawn a new ingest that picks it up
- [ ] Verify `~/.fleet/state/seen-hashes/queue-manager-ingest` is deleted when the bail branch is taken
- [ ] Verify normal (non-bailed) completion still writes the seen-hash as before

## Notes

Companion issue #974 (T-283) extends the fix by filtering `fleet:epic` from the projector so epics can never keep the projection non-empty and re-trigger ingest unnecessarily.

Closes #973

### DIFF
--- a/scripts/fleet/fleet-queue-ingest
+++ b/scripts/fleet/fleet-queue-ingest
@@ -47,7 +47,7 @@ if ! mkdir "$LOCK_DIR" 2>/dev/null; then
     # Without this, scout writes the seen-hash before the spawn returns, so
     # a lock-bailed spawn silently consumes the projection-change signal and
     # approved issues can strand until the hash is manually deleted (issue #973).
-    rm -f "$HOME/.fleet/state/seen-hashes/queue-manager-ingest"
+    rm -f "$HOME/.fleet/state/seen-hashes/queue-manager-ingest" || true
     exit 0
 fi
 trap 'rmdir "$LOCK_DIR" 2>/dev/null; true' EXIT INT TERM
--- a/scripts/fleet/fleet-queue-ingest
+++ b/scripts/fleet/fleet-queue-ingest
@@ -43,6 +43,11 @@ log() {
 # just exits.
 if ! mkdir "$LOCK_DIR" 2>/dev/null; then
     log "another fleet-queue-ingest is running; exiting"
+    # Invalidate the seen-hash so the next scout tick re-fires the trigger.
+    # Without this, scout writes the seen-hash before the spawn returns, so
+    # a lock-bailed spawn silently consumes the projection-change signal and
+    # approved issues can strand until the hash is manually deleted (issue #973).
+    rm -f "$HOME/.fleet/state/seen-hashes/queue-manager-ingest"
     exit 0
 fi
 trap 'rmdir "$LOCK_DIR" 2>/dev/null; true' EXIT INT TERM


jakildev (commented):
## Review — PR #978: T-282: fleet: invalidate seen-hash on ingest lock-bail

**Verdict:** approve

### Summary

Minimal, focused fix: when `fleet-queue-ingest` bails because the lock is already held, it now deletes `~/.fleet/state/seen-hashes/queue-manager-ingest` before exiting so the next scout tick re-fires the trigger. Without this, scout writes the seen-hash before the spawned process returns — a lock-bail silently consumed the projection-change signal, causing approved issues to strand until manual hash deletion.

### Correctness

The fix is logically sound. The lock-bail branch (lines 44-52) executes before the `trap` registration (line 53), and since the lock was never acquired there is nothing to clean up except the hash. The `rm -f` correctly uses `-f` to tolerate the case where the hash file doesn't exist (e.g. first lock-bail before scout has written one). The unconditional deletion on lock-bail is safe: if the projection is empty at the time of bail, the next ingest iteration will fast-exit via the `pending_count == 0` check with only bash startup overhead — consistent with the script's existing "idempotent" contract.

### Comment quality

The added comment explains the non-obvious causal chain (scout writes hash before spawn returns → bail consumes signal → strand) and cites the issue number. Exactly the kind of invariant comment the baseline asks for.

### ECS / render / serialization

Not applicable — fleet tooling script only.

### Nits

- `scripts/fleet/fleet-queue-ingest:50` — the `rm -f` executes under `set -euo pipefail`. If `~/.fleet/state/seen-hashes/` somehow didn't exist, `rm -f` on a path whose parent is missing returns non-zero even with `-f`, and `set -e` would exit the script with an error instead of 0. In practice this cannot happen because scout creates the directory before it spawns fleet-queue-ingest, but appending `|| true` makes the bailout unconditionally safe: `rm -f "$HOME/.fleet/state/seen-hashes/queue-manager-ingest" || true`. Low-priority robustness nit.

### Cross-host smoke tagging

Diff touches `scripts/fleet/` only — no render paths, no shaders. Cross-host smoke labels not required.

Opus recheck not required.

jakildev:
Addressed nit: added `|| true` to `rm -f` on the lock-bail path so it's unconditionally safe under `set -euo pipefail` even if the seen-hashes directory is somehow absent.

Issue #934, opened 2026-05-18
[Epic] Editor authoring enhancements (box-fill, loft, parametric bake)


jakildev:
## Scope

Extends Phase 1 (#604) with the authoring UX gaps required before Phase 2 (#605) starts. Phase 1 today ships single-voxel place/erase only (T-211, T-212, T-213, T-214); authoring 200–400 entities one click at a time is untenable. This epic adds box-fill, loft from two profiles, parametric-shape voxel bake, and ghost-preview during fill.

Reuses existing `IRPrefab::Picking::castVoxelRay` (`engine/prefabs/irreden/render/picking.hpp:47-92`) and `engine/math/include/irreden/math/sdf.hpp`.

Plan: `.claude/plans/okay-lets-go-through-idempotent-giraffe.md` (approved 2026-05-18).

## Sub-tasks (Stack S-A-author)

- [ ] **A1.** AABB box-fill + line-fill + face-fill (foundation; high-ROI cheap fix)
- [ ] **A4.** Selection rectangle + ghost preview during fill
- [ ] **A2.** Loft from 2 profiles (CSG of two extrusions)
- [ ] **A3.** Parametric-shape voxel bake (always emits DENSE)

Tasks land as a linear stack — each based on the previous PR's head, not master.

## Parent

Extends #604 (Phase 1 static voxel authoring).

## Acceptance

All four sub-tasks merged; authoring throughput on a representative entity improves by 5–50× relative to single-voxel place/erase.

jakildev:
Round-1 sub-task filed: A1 (foundation of Stack S-A-author) → #942.

Round-2+ (A4 ghost preview → A2 loft → A3 parametric bake) will be filed as A1 lands.

jakildev:
Round 2+ sub-tasks filed (Stack S-A-author):

- **A4** ghost preview during fill → #947 (stacks on A1 #942)
- **A2** loft from 2 profiles → #948 (stacks on A4 #947)
- **A3** parametric voxel bake → #949 (stacks on A2 #948)

**Full sub-task index:** A1 #942 → A4 #947 → A2 #948 → A3 #949

Queue-manager will ingest these into TASKS.md once approved (per the role boundary clarified in the follow-up rule-tightening PR).

jakildev:
[...]

Issue #667, opened 2026-05-12
[Epic] persist: ECS world snapshot — generic archetype/component/relation serializer (refines #199)


jakildev:
## Status: epic refinement of #199

Issue #199 is the original "save/load framework: serialize ECS world state" task. This epic refines it into a phaseable plan now that the binary-I/O primitives in `engine/asset/` (#663) are being landed alongside the asset-spine work (#664 .. #671).

## Phase summary

Generic ECS world snapshot — every archetype, every component column, every entity ID, every relation. Distinct from asset files (which serialize one thing at a time); the world snapshot is the full ECS state at a point in time.

## Module placement

Lives in **`engine/world/`**, not `engine/asset/`. The snapshot walks the archetype graph and requires `engine/entity/` + `engine/world/` access — pulling those into `engine/asset/` would invert the dependency layering (asset is below entity/world in the engine layer map, see `engine/CLAUDE.md`).

The snapshot consumes the binary-I/O primitives (`BinaryWriter`/`Reader`, `chunk_header`, `name_table`, `json_sidecar`) from `engine/asset/`, but the snapshot's own API surface (`IRPersist::saveWorld` / `loadWorld`, `SaveTrait<C>` traits, migration registry) lives in `engine/world/`.

## Sub-tasks (to be filed when this epic begins)

- W-1: `SaveTrait<C>` traits per component (opt-in / opt-out; default opt-out for GPU handles). Lives in `engine/world/save_trait.hpp`.
- W-2: Per-component `static constexpr uint32_t kSaveVersion`.
- W-3: Archetype walker — iterates the ECS graph in deterministic order (archetypes sorted by sorted ComponentId set; entities sorted by raw ID within an archetype).
- W-4: Migration registry — `(ComponentId, oldVersion) → reader function`.
- W-5: Relation chunk — stores `(relationTypeId, entityA, entityB)` triples with string-name table at chunk head (Extensibility Rule #4 — adding `OWNS`, `ATTACHED_TO` is a one-line enum extension).
- W-6: Singleton chunk — `IREntity::singleton<T>()` instances.
- W-7: Round-trip test per #199 acceptance (100+ entities, multiple archetypes, CHILD_OF/PARENT_TO/SIBLING_OF relations, byte-for-byte parity).
- W-8: Determinism test — same world → same bytes (speedrun verification).
- W-9: Lua bindings — `IRPersist.saveWorld(path)` / `loadWorld(path)`.
[...]

PR #980, opened 2026-05-18
T-283: fleet: filter fleet:epic in project_queue_manager_ingest

### LINKED ISSUE #974: fleet: project_queue_manager_ingest should filter fleet:epic to match role-doc skip rule

jakildev:
## Summary

- Add `_INGEST_SKIP_LABELS = frozenset({"fleet:queued", "fleet:epic", "fleet:needs-plan", "fleet:needs-info"})` near the existing `_ALREADY_QUEUED_LABELS` constant
- Apply the filter in `project_queue_manager_ingest` so epics and un-planned issues are excluded from the trigger-hash projection
- Mirror the same filter in `slice_queue_manager_ingest` so the LLM slice (`~/.fleet/state/projections/queue-manager-ingest.json`) never includes issues the role-doc skips at Step 5

This closes the gap between what the scout's projector tracked (all `human:approved` minus `fleet:queued`) and what the queue-manager LLM actually processes (also excluding `fleet:epic`, `fleet:needs-plan`, `fleet:needs-info`). After this change:

- Issues #667 and #934 (`fleet:epic, human:approved`) will no longer appear in the ingest projection
- With only epics in `human_approved`, the projection hash stays stable and `fleet-queue-ingest` is not spawned
- The empty-projection fast-path in `fleet-queue-ingest` fires correctly when all `human_approved` issues are in skip categories

## Test plan

- [ ] After `fleet-state-scout` runs, `~/.fleet/state/projections/queue-manager-ingest.json` does not include `fleet:epic` issues
- [ ] With only epic issues approved, `fleet-queue-ingest` is not spawned by the scout
- [ ] A genuinely new `human:approved` (non-epic, non-plan) issue still appears in the projection and triggers ingest

Closes #974

### DIFF
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -148,6 +148,16 @@ def fetch_prs(repo):
 
 _ALREADY_QUEUED_LABELS = frozenset({"fleet:queued"})
 
+# Labels that mark issues the queue-manager role doc explicitly skips at
+# Step 5 — filter these out of the ingest projector/slicer so epics and
+# un-planned issues never hold the projection hash non-zero.
+_INGEST_SKIP_LABELS = frozenset({
+    "fleet:queued",      # pre-filtered by fetch_human_approved, included for explicitness
+    "fleet:epic",
+    "fleet:needs-plan",
+    "fleet:needs-info",
+})
+
 
 def fetch_issues_by_label(repo, filter_label, exclude_labels=None):
     out = run_capture([
@@ -822,9 +832,10 @@ def project_merger(state):
 def project_queue_manager_ingest(state):
     """Hash-input projection for the queue-manager-ingest auto-fire.
 
-    Tracks the SET of `human:approved AND NOT fleet:queued` issues —
-    i.e., issues the human has approved but queue-manager hasn't yet
-    filed into TASKS.md. The hash flips when this set CHANGES:
+    Tracks the SET of actionable `human:approved` issues — i.e., issues the
+    human has approved but queue-manager hasn't yet filed into TASKS.md, and
+    that are not in a skip category (epic, needs-plan, needs-info, queued).
+    The hash flips when this set CHANGES:
 
     - Operator adds `human:approved` to issue #N: #N enters the set,
       hash flips, scout spawns fleet-queue-ingest, the LLM iteration
@@ -834,13 +845,15 @@ def project_queue_manager_ingest(state):
       set is empty, exits in <1s. One wasted iteration per ingestion
       cycle, acceptable cost.
 
-    `human_approved` from `fetch_human_approved` already excludes
-    `fleet:queued` issues (that filter lives in fetch_issues_by_label),
-    so this projector simply enumerates that set across both repos.
+    `human_approved` already excludes `fleet:queued` via fetch_human_approved;
+    _INGEST_SKIP_LABELS mirrors the role-doc Step 5 filter so epics and
+    un-planned issues don't hold the hash non-zero indefinitely.
     """
     items = []
     for repo, repo_state in state["repos"].items():
         for issue in repo_state.get("human_approved", []):
+            if set(issue.get("labels", [])) & _INGEST_SKIP_LABELS:
+                continue
             items.append({"repo": repo, "issue": issue["number"]})
     return sorted(items, key=lambda x: json.dumps(x, sort_keys=True))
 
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — PR #980: T-283: fleet: filter fleet:epic in project_queue_manager_ingest

**Verdict:** approve

### Summary

Adds `_INGEST_SKIP_LABELS` frozenset to `fleet-state-scout` and applies it in both `project_queue_manager_ingest` and `slice_queue_manager_ingest`. Epics and un-planned issues no longer hold the projection hash non-zero indefinitely, preventing spurious `fleet-queue-ingest` spawns when the only approved issues are epics.

### Correctness

**Projection and slice parity**: The filter is applied identically in both `project_queue_manager_ingest` (hash input) and `slice_queue_manager_ingest` (LLM input). These two must agree — a divergence would cause the scout to hash over a different set than what gets sliced to the LLM, producing phantom trigger-hash flips. Parity is maintained. Correct.

**Filter logic**: `if set(issue.get("labels", [])) & _INGEST_SKIP_LABELS: continue` — set intersection correctly handles issues with multiple labels. Issues matching any skip label are excluded. No off-by-one or ordering dependency.

**`fleet:queued` in `_INGEST_SKIP_LABELS`**: Included "for explicitness" alongside `fleet:epic`, `fleet:needs-plan`, `fleet:needs-info`. `fetch_human_approved` already excludes `fleet:queued` upstream, so this is a defensive double-filter. Harmless; makes the constant's intent self-documenting.

**Non-ingest queue functions**: `project_queue_manager` and `slice_queue_manager` (the non-ingest variants) are intentionally not modified — they feed a different projection, not the auto-fire trigger. Correct scope.

**Docstrings**: Both updated to accurately describe the new behavior. Concise.

### ECS / render / serialization

Not applicable — fleet tooling script only.

### Cross-host smoke tagging

Diff touches `scripts/fleet/` only. No render paths, no shaders. Cross-host smoke labels not required.

Opus recheck not required.

Issue #938, opened 2026-05-18
[Epic] World streaming + chunked GPU residency


jakildev:
## Scope

Targets the goal: a huge, constantly-simulated world with camera jumps that complete upload + render within a single frame. Only a small voxel subset is GPU-resident at any time; entities move between chunks; the world is sparse over chunks.

**Current state:**
- `kChunkSize = {32,32,32}` (`engine/common/include/irreden/ir_constants.hpp:19`) and `C_ChunkVisibleThisFrame` (`engine/prefabs/irreden/common/components/component_tags_all.hpp:7`) are the only chunk hooks today.
- `kWorldBoundMax` bounds the world to a single chunk — there is no chunk manager, no streaming, no GPU residency layer.
- Everything in this epic is new infrastructure.

**Load-bearing invariant:** no frame may block on upload. When a camera jump exceeds the per-frame upload-bandwidth cap, off-budget chunks render at low-LOD for one frame, then upgrade.

Plan: `.claude/plans/okay-lets-go-through-idempotent-giraffe.md` (approved 2026-05-18).

## Sub-tasks

Stack S-E-stream: E0 → E1 → E2 → E3 → E4
Stack S-E-persist: E1 → E6 (needs B3 merged first)
E5 forks off E2.

- [ ] **E0.** Audit + `docs/design/world-streaming.md` (blocks E1–E6)
- [ ] **E1.** World chunk container + chunk-coords addressing
- [ ] **E2.** GPU chunk residency manager (LRU + camera-radius eviction)
- [ ] **E3.** Camera-aware prefetch (priority by visibility)
- [ ] **E4.** One-frame upload budget + low-LOD fallback
- [ ] **E5.** Entity chunk migration (atomic ownership transfer)
- [ ] **E6.** Disk persistence + lazy load (needs Epic B B3 merged)
- [ ] **E7.** Background simulation tick (DEFERRED to its own epic; tracking-only here)

## Inter-epic dependencies

- E2 cannot land until Epic B **B4** (runtime pool sizing) and **B5** (push-at-mutation upload) are merged.
- E6 cannot land until Epic B **B3** (RLE save format) is merged.
- E5 interacts with Epic C **C6** (rotated entity in world grid migrating).
- E2 strongly benefits from Epic B **B1** (sparse mask) + **B2** (face-aware) — every resident chunk's cost goes down.

jakildev:
[...]

Issue #936, opened 2026-05-18
[Epic] Rotation architecture (local transform + world deformation + detached canvases)


jakildev:
## Scope

Successor to #310 (camera yaw). Adds the per-entity local transform, rotation-mode enum, detached-canvas rotation wiring, and **geometric trixel deformation** that replaces the screen-space bilinear residual established in T-322.

Per user decision during planning:
- World rotation = **Z-axis only** (preserves the (1,1,1) iso depth axis invariant; pitch/roll only on detached canvases per-entity).
- Detached-canvas infrastructure already exists (`C_EntityCanvas`, `IRPrefab::EntityCanvas`, `system_entity_canvas_to_framebuffer.hpp`); we wire it up rather than build new.
- Both voxel and SDF entities respect rotation modes.

Plan: `.claude/plans/okay-lets-go-through-idempotent-giraffe.md` (approved 2026-05-18).

## Sub-tasks

Stack S-C-core: C1 → C2 → C3 → C7
Stack S-C-math: C1 → C5 → C4 → C8
C6 forks off C3.

- [ ] **C1.** `C_LocalTransform` (SQT) component — foundation
- [ ] **C2.** `C_RotationMode` enum + component (GRID vs DETACHED; UNBOUNDED is a flag on `C_LocalTransform`)
- [ ] **C3.** Wire detached-canvas rotation through composite TRS at `system_entity_canvas_to_framebuffer.hpp:98-100`
- [ ] **C4.** Geometric trixel deformation (replaces T-322 bilinear residual)
- [ ] **C5.** Continuous-yaw + deformation math helpers (CPU + GPU mirror; bit-identical)
- [ ] **C6.** GRID-mode rotation: re-rasterize voxels on transform change
- [x] **C7.** DETACHED canvas pitch/roll (full SO(3) inside the canvas)
- [ ] **C8.** SDF entity rotation (gates on Epic D D2 decision)

## Parent

Successor to #310 (camera Z-yaw v1).

## Inter-epic dependencies

- C6 interacts with Epic E E5 (entity chunk migration) — rotated entity straddling boundary.
- C8 should not land until Epic D D2 confirms the SDF runtime restriction shape.

jakildev:
Round-1 sub-task filed: **C1** (`C_LocalTransform` SQT component, foundation of both Stack S-C-core and Stack S-C-math) → #943.

[...]

Issue #731, opened 2026-05-15
[Epic] Engine transform & modifier consolidation


jakildev:
## Goal

Make spatial state consumable from a **single canonical component** and route all spatial perturbations through the **existing modifier system**, instead of the current dual-position pattern (`C_PositionGlobal3D + C_PositionOffset3D` summed by every consumer) and the position-only one-level parent composition that blocks rotation/scale hierarchies and rigging.

## Why

Three structural problems compound today:

1. **Consumers track two positions.** Every render/input/voxel reader of world position must explicitly compute `globalPos + offset` (see `system_sprites_to_screen.hpp:80`, `system_hitbox_mouse_test.hpp:52`, `system_entity_canvas_to_framebuffer.hpp:72`, `system_apply_position_offset.hpp:18`). Easy to forget; brittle as we add more readers.
2. **Hand-rolled perturbation channel.** `C_PositionOffset3D` is doing what should be a modifier's job. Idle bob, gizmo drag adjustments, knockback, screen shake are all logically ADD modifiers on a "position" field — but the modifier system today only handles scalar (`float`) fields.
3. **Position-only hierarchy.** `System<GLOBAL_POSITION_3D>` composes `child.global = child.local + parent.global` with no rotation/scale inheritance. Rigging needs full SQT (scale-quat-translate) parent-chain composition.

Phase 2 of the entity-editor epic (#605) and the rigging/IK/animation chain that follows are all blocked on these three.

## Phases

**Phase 1 — Position modifier consolidation** (current; foundational engine work):

- [ ] #732 — engine: vec3 modifier kind — extend modifier compose for vector fields
- [ ] #733 — engine: delete C_PositionOffset3D — migrate idle bob + gizmo offset to vec3 modifiers

**Phase 2 — SQT transform** (queued after Phase 1):

- [ ] #734 — engine: C_LocalTransform (SQT) + C_WorldTransform + SYSTEM_PROPAGATE_TRANSFORM
- [ ] #735 — engine: quat modifier kind — extend modifier compose for rotation perturbations
- [ ] #736 — engine: migrate position/rotation consumers to C_WorldTransform — retire C_Position3D/C_PositionGlobal3D/C_Rotation

## Acceptance (epic-level)

- A consumer reading "world transform of entity X" reads exactly one component and gets the final SQT after parent composition and modifier resolution.
[...]

Issue #199, opened 2026-04-19
Save/load framework: serialize ECS world state with schema versioning


jakildev:
## Area
`engine/entity`, `engine/asset` (and likely a new `engine/persist` module)

## Suggested model
`[opus]` — ECS core touchpoint, serialization schema design, determinism invariants. Core-engine work.

## Context
Irreden game's design requires full run persistence, between-run primordial soup configuration, post-first-clear mode persistence, and a post-ending permanent world state. The GDD lists "save system" as a production big-rock. Today the engine has no generic serialization — only single-file trixel-texture I/O via `IRAsset::saveTrixelTextureData`.

This is a prerequisite for almost every meta-game system and is a blocker for speedrun state verification.

See also the stale pre-fleet issues #36 (loadable/savable scenes) and #31 (savable voxel sets + trixel canvases as assets) — this task subsumes or coordinates with them.

## Acceptance

1. Generic serializer that snapshots and restores full ECS world state: all archetypes, component columns, entity IDs, and relations (CHILD_OF, PARENT_TO, SIBLING_OF).
2. Component schema versioning — adding, removing, or renaming fields across engine versions does not break old saves; mismatches are detected with a clear diagnostic.
3. Per-component opt-in/opt-out traits. Runtime-only components (GPU handles, live sockets) skip serialization via an opt-out trait.
4. Round-trip test: save a non-trivial world (100+ entities, multiple archetypes, parent/child relations), load into a fresh process, verify archetype/component/relation parity byte-for-byte where applicable.
5. Deterministic serialization order — same world → same bytes. Critical for speedrun state verification and diffable save files.
6. Binary format for speed, plus a human-readable debug dump mode.
7. Builds clean on both `linux-debug` and `macos-debug` presets.
8. Lua bindings expose at least `saveWorld(path)` and `loadWorld(path)` for creations to wire up their own save UI.

## Notes

- This is foundational; do not bundle game-specific save-metadata schema into this task. Game-side save wrapping (run number, diagnosis, primordial soup config) will layer on top in a separate game-repo task.
- Coordinate with `C_JointHierarchy` WIP work — if rigs land around the same time, ensure they serialize cleanly.
- Handle GPU-resident voxel pool state thoughtfully; the pool contents should round-trip, but transient GPU handles regenerate on load.

jakildev:
[...]

PR #981, opened 2026-05-18
T-280: world streaming design doc (E0)

### LINKED ISSUE #944: world: audit chunk infra + write docs/design/world-streaming.md (E0)
## Scope

Survey what chunk infrastructure exists today (`kChunkSize` at `engine/common/include/irreden/ir_constants.hpp:19`, `C_ChunkVisibleThisFrame` at `engine/prefabs/irreden/common/components/component_tags_all.hpp:7`, render-side chunking in voxel pool / visibility compaction, current world bound at `kWorldBoundMax`). Capture the streaming architecture in a new `docs/design/world-streaming.md` covering:

- Chunk identity + addressing (ivec3 chunk coords).
- Residency manager API (resident-set, eviction policy, async upload jobs).
- Prefetch policy (camera radius + visibility priority).
- One-frame upload-bandwidth cap + low-LOD fallback shape. **Resolves the open question:** downsampled voxel proxy vs. SDF silhouette proxy vs. AABB billboard for off-budget chunks.
- Entity migration semantics (atomic ownership transfer; preserves entity id; interacts with C6 (#936) rotated-entity grid rotation).
- Disk persistence story (reuses `.vxs` infra + B3 RLE).

**Blocks** all of E1–E6 (#938). Foundation task — no code in this PR.

## Parent epic

## Stack

Base of Stack S-E-stream (E0 → E1 → E2 → E3 → E4) and Stack S-E-persist (E1 → E6).

## Dependencies

(none — design only)

## Acceptance

- (1) `docs/design/world-streaming.md` reviewed by another Opus agent.
- (2) Document covers all six topics above with concrete data structures and API sketches.
- (3) Low-LOD representation choice resolved with rationale.
- (4) No code changes.

## Plan reference

`.claude/plans/okay-lets-go-through-idempotent-giraffe.md` §"Epic E → E0".

jakildev:
## Summary
- New `docs/design/world-streaming.md` — E0 foundation design for Epic E (#938).
- Covers all six topics from #944: chunk identity + addressing, residency manager API, prefetch policy, upload-bandwidth cap + low-LOD fallback, entity migration semantics, disk persistence.
- Resolves the open question on low-LOD fallback (decision: AABB billboard for v1, SDF silhouette proxy deferred to v2).

## Test plan
- [x] Doc covers all six topics from the issue body with concrete data structures and API sketches.
- [x] Low-LOD representation choice resolved with rationale.
- [x] No code changes.
- [ ] (E0 acceptance #1) Reviewed by another Opus agent — this PR.

Closes #944

### DIFF
--- a/docs/design/world-streaming.md
+++ b/docs/design/world-streaming.md
@@ -109,8 +109,9 @@ Three concrete reasons for the `int16`-per-axis pack:
 - 64-bit key fits in an `unordered_map<ChunkKey, ...>` slot and a
   `std::atomic<uint64_t>` for lock-free residency-set scans.
 - Pretty filenames: `world/chunks/-0001_+0042_+0007.vxs` (axis order
-  x, y, z, signed three-digit padded — chosen so `ls` sorts the disk
-  representation usefully).
+  x, y, z, signed four-digit padded — chosen so `ls` sorts the disk
+  representation usefully; four digits cover the full ±32 768 chunk
+  range without overflow).
 
 ### Chunk component on entities
 
@@ -190,7 +191,17 @@ public:
     // Forced overrides (editor "load this chunk now" / save-all).
     void requestResident(ChunkKey key, RequestPriority p);
     void requestEvict(ChunkKey key);
-    void flushAllPending(); // blocks; used by save-snapshot path
+
+    // Entity migration — called from system_propagate_chunk_membership's
+    // endTick. Public because the migration system lives in
+    // engine/prefabs/irreden/world/, outside engine/world/.
+    void migrateEntity(IREntity::EntityId id, ChunkKey oldKey, ChunkKey newKey);
+
+    // Blocks until all pending loads and saves complete.
+    // Used by: World construction (warm-up) and save-snapshot path.
+    // Covers both "drain all LOADING/UPLOADING slots" and
+    // "flush all dirty EVICTING slots" in a single drain pass.
+    void flushAllPending();
 };
 ```
 
@@ -267,11 +278,14 @@ band in all six directions; `R_view` always rounds up to include
 
 ### Async upload job model
 
-Disk → CPU and CPU → GPU run on a single `std::thread` worker pool
-(initially 2 threads — disk-bound and GPU-staging-bound) drained
-each frame. The job queue is a `concurrent_queue<UploadJob>` (custom
-SPSC ring; no `<concurrent_queue>` standard yet, so we ship a small
-header in `engine/world/`). Each job carries:
+Disk → CPU and CPU → GPU run as two dedicated `std::thread` workers
+(one disk-bound, one GPU-staging-bound). Each stage has its **own**
+SPSC queue — main thread → disk-thread for `LOAD_DISK` jobs, and
+disk-thread → GPU-staging-thread for `UPLOAD_GPU` jobs — so each
+queue has exactly one producer and one consumer. (A single shared
+queue would be SPMC, which is not safe with an SPSC ring.) No
+`<concurrent_queue>` standard yet; we ship a small SPSC ring header
+in `engine/world/`. Each job carries:
 
 ```cpp
 struct UploadJob {
@@ -451,10 +465,9 @@ Specifically:
   tick").
 - A migration adds the entity to the destination chunk's
   `ownedEntities_` and removes it from the source chunk's. If the
-  source chunk is non-resident at the time of migration (rare —
-  the entity must have been in a then-resident chunk to have
-  moved), the migration is queued and applied when the source
-  chunk re-enters residency. (See "Edge cases" below.)
+  source chunk is non-resident at the time of migration, the
+  migration is queued and applied when the source chunk
+  re-enters residency. (See "Edge cases" below.)
 
 ### Migration system
 
@@ -549,10 +562,12 @@ streaming.
    `IREntity::entityExists(record.entityId)` before applying.
 
 3. **Source chunk evicted while a child entity's migration is
-   pending.** The pending migration is dropped — the entity has
-   already destructed with the source chunk's eviction. (Eviction
-   does not save individual entities; it saves the whole chunk —
-   see Topic 6.)
+   pending.** The pending migration is dropped. The entity was
+   serialized to disk with the source chunk at eviction time; its
+   `EntityId` persists in the saved chunk file. When the source
+   chunk reloads, the entity reappears at its original
+   pre-migration position. (Eviction saves the whole chunk, not
+   individual entities — see Topic 6.)
 
 4. **Concurrent migration + structural change.** All structural
    changes route through `flushStructuralChanges` at pipeline end.
--- a/docs/design/world-streaming.md
+++ b/docs/design/world-streaming.md
@@ -20,8 +20,10 @@ systems).
 Today the engine renders one canvas with one fully-resident voxel pool
 sized at `kVoxelPoolSize = 64³` voxels (see
 `engine/common/include/irreden/ir_constants.hpp:63`). The world bound
-collapses to `kChunkSize - 1` on each axis (`kWorldBoundMax`,
-`ir_constants.hpp:42`) — i.e. the entire "world" is a single chunk.
+collapses to `vec3(kChunkSize) - vec3(1, 1, 2)` (`kWorldBoundMax`,
+`ir_constants.hpp:42`; the z ceiling is `kChunkSize.z - 2` to leave
+the top ground layer as scene floor) — i.e. the entire "world" is a
+single chunk.
 `kChunkSize = uvec3{32, 32, 32}` and `C_ChunkVisibleThisFrame` are
 declared but unused by any system; the bookkeeping is forward-looking.
 
@@ -197,11 +199,16 @@ public:
     // engine/prefabs/irreden/world/, outside engine/world/.
     void migrateEntity(IREntity::EntityId id, ChunkKey oldKey, ChunkKey newKey);
 
-    // Blocks until all pending loads and saves complete.
-    // Used by: World construction (warm-up) and save-snapshot path.
-    // Covers both "drain all LOADING/UPLOADING slots" and
-    // "flush all dirty EVICTING slots" in a single drain pass.
-    void flushAllPending();
+    // Drain pending LOAD_DISK + UPLOAD_GPU work to RESIDENT.
+    // Used by: World construction (warm-up — bring the spawn camera's
+    // resident-set fully online before the first rendered frame).
+    void flushPendingLoads();
+
+    // Force-save every dirty resident slot, then drain the save queue.
+    // Used by: save-snapshot path (paired with #199's world-snapshot
+    // save). Does not touch in-flight loads — call flushPendingLoads()
+    // first if a settled load state is required.
+    void flushPendingSaves();
 };
 ```
 
@@ -279,13 +286,50 @@ band in all six directions; `R_view` always rounds up to include
 ### Async upload job model
 
 Disk → CPU and CPU → GPU run as two dedicated `std::thread` workers
-(one disk-bound, one GPU-staging-bound). Each stage has its **own**
-SPSC queue — main thread → disk-thread for `LOAD_DISK` jobs, and
-disk-thread → GPU-staging-thread for `UPLOAD_GPU` jobs — so each
-queue has exactly one producer and one consumer. (A single shared
-queue would be SPMC, which is not safe with an SPSC ring.) No
-`<concurrent_queue>` standard yet; we ship a small SPSC ring header
-in `engine/world/`. Each job carries:
+(one disk-bound, one GPU-staging-bound). The pipeline is a **three-
+SPSC linear chain**, one queue per stage transition — each queue has
+exactly one producer thread and exactly one consumer thread, which is
+the only shape under which our lock-free SPSC ring is safe:
+
+```
+                   loadDiskQueue           uploadGpuQueue
+   main thread  ──────────────▶  disk    ──────────────▶  GPU-staging
+   (producer)                    thread                   thread
+                                 (consumer)               (consumer)
+                                 ▲ producer next ─────────┘
+                                                          │
+                                          completionQueue │
+   main thread  ◀───────────────────────────────────────────┘
+   (consumer)
+```
+
+- `loadDiskQueue` — main thread → disk thread. Carries `LOAD_DISK`
+  jobs. Main thread enqueues when a chunk enters `LOADING`; disk
+  thread drains, performs the file read into `stagingBytes_`, then
+  pushes the same job (re-stamped `UPLOAD_GPU`) onto the next queue.
+- `uploadGpuQueue` — disk thread → GPU-staging thread. Carries
+  `UPLOAD_GPU` jobs. GPU-staging thread drains, performs the upload
+  into the chunk's `VoxelPoolAllocation`, then pushes a
+  `JobCompletion` onto the completion queue.
+- `completionQueue` — GPU-staging thread → main thread. Carries
+  `JobCompletion` records. Main thread drains in `flushUploads` and
+  transitions slots to `RESIDENT`, attaches entities, etc. Errors
+  propagate **forward** along the pipeline rather than back: a disk-
+  read failure stamps the job `errorStage = LOAD_DISK` and pushes it
+  onto `uploadGpuQueue` with empty staging bytes; the GPU-staging
+  thread sees the error stamp, skips the actual GPU upload, and
+  emits the terminal `JobCompletion{status = ERROR}`. This keeps the
+  GPU-staging thread the sole producer of `completionQueue` (true
+  SPSC) and guarantees the main thread receives exactly one
+  completion record per scheduled job.
+
+A single shared `UploadJob` queue with two consumers would be SPMC
+and not safe with an SPSC ring; a single shared completion queue
+with two producers would be MPSC and likewise unsafe. The three-
+queue split keeps every edge SPSC.
+
+No `<concurrent_queue>` standard yet; we ship a small SPSC ring
+header in `engine/world/`. Each `UploadJob` carries:
 
 ```cpp
 struct UploadJob {
@@ -294,12 +338,20 @@ struct UploadJob {
     std::vector<std::uint8_t> stagingBytes_; // CPU buffer between stages
     // Size in bytes for budget accounting; one frame's flush sums these.
     std::uint64_t sizeBytes_ = 0;
+    // Set by a stage when it fails; the next stage skips its work
+    // and just forwards the record so the terminal stage can emit one
+    // JobCompletion{ERROR}. Stage::LOAD_DISK means "disk read failed";
+    // Stage::UPLOAD_GPU is unused (GPU stage emits the completion
+    // directly). Default = "no error so far".
+    std::optional<Stage> errorStage_;
 };
 ```
 
-Completion signals back through `concurrent_queue<JobCompletion>`
-drained on the main thread in `flushUploads` so all `IREntity` /
-`IRRender` calls remain single-threaded.
+`JobCompletion` carries the chunk key, terminal status (`OK`/`ERROR`)
+and (on error) a short diagnostic. The main thread's drain in
+`flushUploads` is the only site that touches `IREntity` /
+`IRRender` state on completion, so all engine-state mutation stays
+single-threaded.
 
 ---
 
@@ -352,7 +404,7 @@ those are the cases where the static ring miss-rate spikes.
 ### Cache warm-up
 
 On `World` construction, the residency manager calls
-`flushAllPending()` against the spawn camera's resident-set so the
+`flushPendingLoads()` against the spawn camera's resident-set so the
 first rendered frame is not staring at a low-LOD wall. The cost is
 deterministic: spawn radius is bounded by `R_view` and chunk-load
 time is `O(disk_read + voxel_copy)` per chunk.
@@ -465,9 +517,11 @@ Specifically:
   tick").
 - A migration adds the entity to the destination chunk's
   `ownedEntities_` and removes it from the source chunk's. If the
-  source chunk is non-resident at the time of migration, the
-  migration is queued and applied when the source chunk
-  re-enters residency. (See "Edge cases" below.)
+  destination chunk is non-resident at the time of migration, the
+  migration is queued on a manager-owned deferred queue, the
+  destination chunk's residency is force-requested, and the record
+  is applied at the start of `tickPrefetch` on the frame the
+  destination becomes resident. (See "Edge cases" below.)
 
 ### Migration system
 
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — PR #981: T-280: world streaming design doc (E0)

**Verdict:** approve

### Summary

Adds `docs/design/world-streaming.md` — an 829-line foundation design for world chunk streaming (Epic E, #938). Pure documentation; no code changes. Covers all six acceptance criteria from #944: chunk identity + addressing, residency manager API, prefetch policy, upload-bandwidth cap + low-LOD fallback, entity migration semantics, and disk persistence. Resolves the open low-LOD representation question (AABB billboard for v1, SDF silhouette proxy deferred to v2).

### Acceptance criteria coverage

All six #944 criteria are present and well-developed. The acceptance trace table at the end is a useful navigational aid and maps cleanly to the topic structure.

### Topic-by-topic assessment

**Topic 1 — Chunk identity + addressing:** Solid. The `worldToChunk` floor-toward-negative-infinity note is the non-obvious behavior that deserves the callout. The `ChunkKey` int16×3 pack covering ±1 M voxels per axis is appropriate for the stated product scope. `C_ChunkMembership` opt-in design (no auto-attach) correctly keeps zero overhead for all existing single-chunk creations.

**Topic 2 — Residency manager API:** The non-system frame-hook architecture (`beginFrame`/`tickPrefetch`/`flushUploads`/`endFrame`) is the right call for a manager that spans pipeline phases. The three-zone eviction policy (view / prefetch / evict) with hysteresis mirrors the light-occlusion grid's camera-anchored pattern — internally consistent. Shadow-ring and AO guard-band expansion correctly reference the existing invariants from `engine/render/CLAUDE.md`.

**One substantive concern:** The `UploadJob` queue is described as "a `concurrent_queue<UploadJob>` (custom SPSC ring)" with "initially 2 threads — disk-bound and GPU-staging-bound." An SPSC ring has one producer and one consumer. With two worker threads both draining `UploadJob`s (even if one handles LOAD_DISK and the other UPLOAD_GPU, they share a single queue per the current text), the queue is actually SPMC (single-producer, multi-consumer). An SPSC ring buffer is not safe for two concurrent consumers. If the design intent is one queue per stage (LOAD_DISK queue consumed by the disk thread, UPLOAD_GPU queue consumed by the GPU-staging thread), the text should say so explicitly — as written, the E1 implementer could reasonably read "a `concurrent_queue<UploadJob>`" as a single shared queue and produce a data race. This deserves clarification before E1 starts.

**Topic 3 — Prefetch policy:** The radius constants (`R_view=128`, `R_prefetch=256`, `R_hysteresis=32`) are anchored to existing reference points (light-volume window = 128³ today). The velocity-hint deferral is appropriately justified. Cache warm-up via `flushAllPending()` at `World` construction is correct in intent (see also the Topic 6 concern below).

**Topic 4 — Low-LOD fallback decision:** The three-option comparison table is well-argued. The AABB-billboard choice on "marginal renderer surface" and "reads as placeholder by design" grounds is sound. The forward-compat path via a future `LSDF` chunk tag is consistent with Save Format Extensibility Rule #1.

**Topic 5 — Entity migration semantics:** The `PROPAGATE_CHUNK_MEMBERSHIP` placement (after `PROPAGATE_TRANSFORM`, before chunk-aware downstream systems) is correct. Edge cases 1–4 are correctly handled. One precision issue in edge case 3: "the entity has already destructed with the source chunk's eviction" — entities are serialized to disk on eviction, not destroyed. The EntityId persists in the saved chunk file. "Destructed" implies `destroyEntity` was called, which is incorrect. The accurate description is: the entity was serialized to disk with the source chunk; the pending migration is dropped; when the source chunk reloads, the entity reappears in its original pre-migration position.

**Topic 6 — Disk persistence:** `.vxs` HYBRID-mode reuse is the right call — no new format, old loaders skip unknown chunks per Rule #1. The two-phase `BBOX`-first load (billboard instant, full `VOXR`+`ENTS` budget-deferred) is elegant and directly enables the low-LOD streaming behavior. Per-chunk file granularity with a two-level directory split (deferred to implementation) is the correct scope for a design doc.

**`flushAllPending()` dual-use:** The method is described in Topic 3 as the cache warm-up call at `World` construction (loads resident chunks), and in Topic 6 as the save-all-dirty call from the save-snapshot button (saves all dirty chunks). These two behaviors — "block until all pending loads complete" vs. "block until all pending saves complete" — are meaningfully different operations that the same method name will need to cover. The E1 implementer should clarify whether this is one method with a mode parameter, two separate methods, or a combined "drain everything" that happens to cover both. As written, the overlap may cause confusion.

**`migrateEntity` API gap:** Topic 5's migration system calls `ChunkResidencyManager::migrateEntity(id, oldKey, newKey)` in `endTick`, but this method is not listed in Topic 2's public API sketch. Either it belongs in the public API (used by `system_propagate_chunk_membership`, which is outside `engine/world/`) or it's package-internal (if the migration system is co-located in `engine/world/`). The placement of the migration system in `engine/prefabs/irreden/world/` (as sketched) implies it's in a different module from `engine/world/`, so `migrateEntity` would need to be public. Worth adding to the API sketch.

### Nits

- `docs/design/world-streaming.md:532` — "the entity must have been in a then-resident chunk to have moved" is a parenthetical explaining why edge case 3 (source chunk evicted while migration pending) is rare, but it understates the scenario. An entity in a resident chunk that moves to a destination chunk, while the migration is pending (at frame boundary), and then the source chunk is evicted in the same frame — this is possible even without an explicit assumption about rarity. The note "Eviction does not save individual entities; it saves the whole chunk — see Topic 6" at the end covers the resolution cleanly; the parenthetical can be dropped.
- The `saves/<save-name>/` layout shows chunk files like `+0000_+0000_+0000.vxs` (three-digit padded), but Topic 1 says "three-digit padded" for a range of ±32 768. Three-digit padding overflows for chunks beyond ±999. Either use four-digit padding (`+0000`–`+9999`) or drop the padding claim in Topic 1 and use `printf`-style variable-width format (simpler, and `ls`-sort still works for non-negative with zero-padded integers). Low priority — the padded filename format is an implementation detail.

### Cross-host smoke tagging

Diff touches only `docs/design/world-streaming.md` — no `engine/render/`, no `.glsl`, no `.metal`. Cross-host smoke labels not required.

Opus recheck required: E0's acceptance criterion (#944 criterion 1) explicitly calls for "reviewed by another Opus agent." Additionally, the SPSC queue concern above involves a concurrency design decision (shared queue vs. per-stage queues for the async upload pipeline) that warrants an Opus-level pass before E1 begins.

jakildev:
Addressed review feedback (nits + substantive doc clarifications):

- **Nit — four-digit padding:** Fixed "three-digit padded" → "four-digit padded" in Topic 1 filename format with a note that 4 digits cover the full ±32 768 chunk range without overflow.
- **Nit — misleading parenthetical:** Removed `(rare — the entity must have been in a then-resident chunk to have moved)` from the migration invariant.
- **Edge case 3 wording:** Replaced "entity has already destructed" with the accurate description: entity was serialized to disk with the source chunk; EntityId persists in the saved chunk file and reappears at original position when the source chunk reloads.
- **`migrateEntity` API gap:** Added `migrateEntity(EntityId, ChunkKey, ChunkKey)` to the `ChunkResidencyManager` public API sketch with a comment explaining it's public because the migration system lives in `engine/prefabs/irreden/world/`, outside `engine/world/`.
- **`flushAllPending` dual-use:** Added a comment clarifying the method drains both pending loads (LOADING/UPLOADING slots) and pending saves (dirty EVICTING slots) in a single pass, covering both the warm-up and save-snapshot callers.
- **SPSC vs SPMC queue:** Clarified that each pipeline stage (LOAD_DISK, UPLOAD_GPU) has its own SPSC queue so each has exactly one producer and one consumer. Added an explicit note that a single shared queue would be SPMC and unsafe with an SPSC ring.

— sonnet-author (sonnet-fleet-2)

jakildev (commented):
## Opus recheck — PR #981: T-280: world streaming design doc (E0)

**Verdict:** needs-fix

This is a substantive design doc. The overall architecture (camera-anchored residency, 32³ chunks, `.vxs` HYBRID reuse, AABB-billboard low-LOD, deferred entity migration through `flushStructuralChanges`) is sound and well-integrated with the existing render and asset invariants. The chunk-coord int16-pack, the three-zone view/prefetch/evict policy with hysteresis, and the `BBOX`-first two-phase load are good calls.

Sonnet's review covered ECS naming, public surface, and acceptance-criteria coverage thoroughly; I'm not duplicating that work. My recheck focuses on concurrency, lifecycle/ownership, and design ambiguities the E1+ implementer will hit — and I'm bumping the verdict to needs-fix because four of those would block E1 from starting without a return trip to design.

### Confirming Sonnet's substantive flags

**1. SPSC queue mislabel (`docs/design/world-streaming.md:273-288`)** — Sonnet is right, and it's worse than Sonnet wrote. The doc currently specifies *one* `concurrent_queue<UploadJob>` shared between the disk-bound and GPU-staging-bound threads. SPSC is incorrect for that shape (it would be SPMC). Additionally, the completion queue ("Completion signals back through `concurrent_queue<JobCompletion>` drained on the main thread") is MPSC, not SPSC — two worker threads produce, one main thread consumes.

The likely correct shape is a **three-SPSC linear pipeline**, one queue per stage transition:
- `loadDiskQueue` — main thread (producer) → disk thread (consumer). LOAD_DISK jobs.
- `uploadGpuQueue` — disk thread (producer) → GPU-staging thread (consumer). UPLOAD_GPU jobs created when disk-load completes.
- `completionQueue` — GPU-staging thread (producer) → main thread (consumer). Completion records.

Each pair is single-producer/single-consumer, so the custom SPSC ring is correct and lock-free. This needs to be in the doc explicitly — as written, the E1 implementer could reasonably pick a single shared queue and produce a data race that won't surface until two-worker-thread stress shows up under profiling.

**2. `flushAllPending()` dual-use (Topic 3 vs Topic 6)** — Confirmed. The two call sites have different semantics:
- Topic 3 cache warm-up: *drain pending LOADS for the spawn camera's resident-set*. Synchronous against the LOAD_DISK + UPLOAD_GPU pipeline.
- Topic 6 save-all path: *drain pending SAVES for all dirty resident slots*. Synchronous against the save-to-disk pipeline.

These are not the same operation, and they don't naturally compose into one "drain everything." Recommend splitting:
```cpp
void flushPendingLoads();   // blocks until LOAD_DISK + UPLOAD_GPU queues drain
void flushPendingSaves();   // blocks until save-to-disk queue drains; forces save on every dirty slot
```
If a save-snapshot button needs both (e.g. to ensure no in-flight loads are mid-mutation), it calls them in sequence — the API expresses the ordering, and neither name is overloaded.

**3. `migrateEntity` missing from Topic 2's public API** — Confirmed. Topic 5 calls `ChunkResidencyManager::migrateEntity(id, oldKey, newKey)` from `system_propagate_chunk_membership::endTick`. That system lives in `engine/prefabs/irreden/world/`, distinct from `engine/world/`, so the call is across the module boundary and the method must be public. Add it to the class sketch in Topic 2.

### Additional design gaps I found (not in Sonnet's review)

**4. Deferred-migration storage for non-resident destination chunks (Topic 5, edge case 1)** — The text says "if non-resident, a `requestResident(destKey, FORCED)` is issued and the migration applies on the next frame when the chunk becomes resident." The mechanism for "applies on the next frame" is unspecified.

The migration system's `migrations_` vector is cleared every `endTick` (params-on-System pattern). So either:
- A separate persistent queue lives on `ChunkResidencyManager` (e.g. `pendingMigrationsToNonResident_`), checked at each subsequent `endFrame` until the destination becomes resident — call this `migrateEntity`'s deferred path.
- Or the migration record stays in the System across frames, which violates the "clear per frame" pattern.

Either is workable, but the doc must pick one. The natural answer is the first — `ChunkResidencyManager::migrateEntity` accepts the record, applies it immediately when both source and destination are resident, and queues it on a manager-owned deque otherwise. The deque is drained at the start of each `tickPrefetch` after the residency-set update completes. Worth one paragraph in Topic 5.

**5. EntityId stability for cross-chunk relations across save/load (Topic 6, ENTS chunk)** — The ENTS format stores cross-chunk relations as "absolute id if cross-chunk — flag bit." Absolute EntityIds in `IREntity` are runtime-assigned by `createEntity` and are not stable across save/load cycles.

For cross-chunk CHILD_OF (or any future cross-chunk relation) to round-trip, the snapshot encoder from #199 must define a stable per-entity identifier that survives the load cycle. The doc currently says "the world snapshot (#199) is the system of record for 'how an entity serializes,'" which is correct framing — but it should explicitly note that the cross-chunk-id field is the snapshot encoder's stable identifier, NOT a raw `EntityId`. Otherwise the E6 implementer reads "absolute EntityId" literally and ships a format that breaks on every save/load round-trip.

The simplest fix is one line in Topic 6: rename the field to "stable entity identifier (from #199's identity scheme)" and note in the inter-chunk-resolution paragraph that the loader maps stable identifiers → freshly-assigned EntityIds after both chunks reload.

**6. Eviction-save vs re-resident race (Topic 6, "Dirty tracking + eviction-write")** — A `dirty_` slot enters `EVICTING`, a save job is scheduled against the worker pool. Two things can happen before the save completes:
- The slot is re-requested (camera reversed direction). EVICTING → RESIDENT. Save is still in flight; it reads the pool slice's bytes. If the main thread mutates the pool slice (e.g. paint operation, entity migration in), the save reads a torn snapshot.
- The slot completes EVICTING → EVICTED while save is in flight. The pool slice can't be freed until save completes — the doc covers this with "the pool slice holding the data must persist until the save completes," good. But the dirty-bit state on a re-resident-during-save slot is undefined: if save completes after re-resident, clearing dirty drops the post-save mutations.

The cleanest resolution is to **snapshot at schedule time**: the save job receives a CPU-side copy of the voxel + entity bytes from the pool slice synchronously when scheduled, and the slice is free to be mutated immediately. This costs one main-thread copy per eviction, but eviction is rare relative to per-frame work, and it eliminates the race entirely.

If instead the save job reads from the live pool slice, the doc must spell out the locking model — e.g. EVICTING is a "write-locked" state where main-thread voxel mutations are deferred until save completes.

Either is fine; the doc should pick one and call it out.

### Architecture I confirm is sound (Opus-level pass)

- **Camera-anchored residency matches light-occlusion grid anchoring.** The `R_view=128`/`R_prefetch=256`/`R_hysteresis=32` constants are chosen specifically so the lighting and residency systems can't disagree about reachability (`engine/render/CLAUDE.md` "Lighting culling invariants — world-space"). Good.
- **Shadow-ring + AO guard-band expansion are correctly conditional on `IRRender::getSunShadowsEnabled()` and `IRMath::shadowFeederIsoBounds`.** Invariants #2 and #4 from `engine/render/CLAUDE.md` are honored.
- **`PROPAGATE_CHUNK_MEMBERSHIP` placement (after `PROPAGATE_TRANSFORM`, before chunk-aware downstream) is correct** — `C_WorldTransform` is the canonical world-position source post-#731, and reading it before chunk-membership consumers ensures one tick of consistency.
- **`createEntity` does not auto-attach `C_ChunkMembership`** — correctly preserves zero overhead for single-chunk creations. The opt-in via `IRWorld::enableChunkStreaming(config)` is the right activation point per `engine/CLAUDE.md` "createEntity auto-attaches transform components."
- **LOD framework orthogonality** — the streaming low-LOD billboard is independent from the artist-driven LOD tiers in `docs/design/lod-strategy.md`. Two distinct decisions, one per-chunk and one per-entity-tier. Good split.
- **C6 (#936) interaction analysis is correct.** Chunk membership tracks the root-entity world position, rotation lives on `C_LocalTransform`, light-blocker bounds expand past the chunk AABB exactly as today. No new invariant.
- **`migrations_` vector capacity retained across frames per `.claude/rules/cpp-ecs.md` "Allocations in hot tick paths."** Correct application of the rule.

### Nits (only address if you're doing a respin for the above)

- `:553` — "the entity has already destructed with the source chunk's eviction" — entities are *serialized to disk* on eviction, not destroyed. Replace with "the entity was serialized to disk with the source chunk; the pending migration is dropped; on source-chunk reload, the entity reappears in its pre-migration position." Sonnet flagged this already.
- `:113` — "axis order x, y, z, signed three-digit padded" — three digits overflows for chunk coords past ±999 in a ±32 768 range. Either use four-digit padding (`+00000`...`+32767`), strip the padding claim and accept variable-width filenames, or constrain the docs explicitly to "current product scope ≤999 chunks per axis." Sonnet flagged.
- `:42` callout — the doc says `kWorldBoundMax = kChunkSize - 1`; the actual constant is `vec3(kChunkSize) - vec3(1, 1, 2)` (the z ceiling is `kChunkSize.z - 2` to leave the top ground layer as scene floor). Minor doc-vs-code inconsistency.

### Cross-host smoke tagging

Doc-only PR; no render-path files touched. Smoke labels not required.

### Verdict reasoning

Each individual finding is fixable, but four of them (1, 2, 4, 6) are design-contract questions the E1+ implementer would have to answer themselves if the doc shipped as-is — and the wrong choice on the SPSC question in particular would produce a hard-to-reproduce data race. The doc is close, but E0's whole purpose is to lock these choices before E1 starts. Worth one more pass.

Reviewed by Claude Opus 4.7 (opus-reviewer role)

jakildev:
Addressed Opus recheck feedback:

- **(1) SPSC pipeline mislabel** — Rewrote the "Async upload job model" section as an explicit three-SPSC linear chain (`loadDiskQueue`, `uploadGpuQueue`, `completionQueue`). Each edge is strictly single-producer/single-consumer. Errors propagate forward (failing stage stamps `errorStage_`, the GPU-staging thread emits the terminal `JobCompletion`) so the completion queue stays SPSC rather than degrading to MPSC.
- **(2) `flushAllPending()` dual-use** — Split into `flushPendingLoads()` (drains LOAD_DISK + UPLOAD_GPU) and `flushPendingSaves()` (force-saves dirty slots, drains the save queue). Updated all three call sites (Topic 2 class sketch, Topic 3 cache warm-up, Topic 6 save-all path). Snapshot save-path docs note the pairing order when both are needed.
- **(3) `migrateEntity` in Topic 2** — Already in the prior fixup; verified present at line 198 of the updated doc.
- **(4) Deferred-migration storage** — Topic 5 edge case 1 now specifies a `ChunkResidencyManager`-owned `pendingMigrationsToNonResident_` deque drained at the start of each `tickPrefetch` after residency-set update. Cross-references the per-tick `migrations_` clear rule so the per-tick allocation invariant is preserved.
- **(5) EntityId stability across save/load** — Renamed the `ENTS` cross-chunk relation field to "stable cross-chunk identifier from #199's identity scheme" with a `NOT a raw runtime EntityId` callout. Added a paragraph spelling out the round-trip mapping: snapshot encoder defines stable ID → loader maps stable ID → fresh `EntityId` after both chunks reload.
- **(6) Eviction-save race** — Added a "Race resolution — snapshot at schedule time" subsection picking the synchronous-CPU-copy approach. Explicitly rejects the write-lock-EVICTING alternative with rationale (renderer→disk-I/O latency coupling). Save-all path updated to use the same copy mechanism.
- **Nit `:42`** — Corrected `kWorldBoundMax` formula to `vec3(kChunkSize) - vec3(1, 1, 2)` with the scene-floor note.

Sonnet's nits (`:553` migration text, `:113` four-digit padding) and items 1/2/3 from Sonnet's review were already addressed in the prior fixup commit.

— opus-worker-2

jakildev (commented):
## Re-review — PR #981: T-280: world streaming design doc (E0)

**Verdict:** needs-fix

### Prior-review resolution

Verifying each item the Opus recheck raised as a design-contract blocker:

- **SPSC queue mislabel / shared queue shape** — RESOLVED. The async upload section is rewritten with a clear three-SPSC-chain diagram: `loadDiskQueue` (main → disk), `uploadGpuQueue` (disk → GPU-staging), `completionQueue` (GPU-staging → main). Every edge is exactly one producer / one consumer. Error propagation is forward-only through `errorStage_` on `UploadJob`, keeping GPU-staging the sole producer of `completionQueue`. Correct.
- **`flushAllPending()` ambiguity** — RESOLVED. Split into `flushPendingLoads()` (warm-up path) and `flushPendingSaves()` (snapshot path), with distinct usage notes for each. Call-site doc updated at lines 404 and 773–782.
- **`kWorldBoundMax` doc-vs-code inconsistency** — RESOLVED. Now reads `vec3(kChunkSize) - vec3(1, 1, 2)` with the z-ceiling explanation. Matches the actual constant.
- **Migration: source vs. destination non-resident** — RESOLVED. The wrong "source non-resident" case is corrected to "destination non-resident." The deferred-queue semantics (`pendingMigrationsToNonResident_`, `std::deque<MigrationRecord>` on `ChunkResidencyManager`, drained at the start of `tickPrefetch`) are now fully specified. The cross-frame lifetime is explained correctly: System's `migrations_` vector is per-tick, manager's deferred queue survives frames.
- **Cross-chunk entity identity (raw EntityId in serial format)** — RESOLVED. Cross-chunk relation now stores the stable per-entity identifier from #199's identity scheme, not a raw runtime EntityId. The loaders map stable-id → fresh EntityId after both chunks reload.
- **Save-while-evicting race** — RESOLVED. The snapshot-at-schedule-time approach (synchronous CPU-side copy of the pool slice before the async save is dispatched) eliminates both the torn-snapshot hazard and the dirty-bit collision. The design decision is stated explicitly and the rejected alternative (write-lock EVICTING) is documented. Cost analysis is correct: eviction is rare, one memcpy per eviction is negligible.

### Needs-fix

- **`docs/design/world-streaming.md:115`** — The filename-padding claim is factually incorrect.

  Current text: "four digits cover the full ±32 768 chunk range without overflow"

  Four-digit zero-padding spans `0000`–`9999`, i.e. ±9,999 per axis. The int16 ChunkKey range is ±32,767 (32,767 requires five digits). A chunk at index 10,000 would produce `+10000` — five digits, breaking the fixed-width format and `ls`-sort order.

  Fix options (pick one and update the example in `saves/<...>/chunks/` layout at line 826):
  - **Five-digit padding** — `+00000_+00000_+00000.vxs`. Covers ±99,999; cleanly contains the full int16 range.
  - **Constrain the scope claim** — keep four-digit padding but state it covers "±9,999 per axis — far beyond any foreseeable product scope." Drop the claim about ±32,768 coverage. The ChunkKey type can still be int16; the filename format just doesn't exploit the full range.

  Either option is valid. The current claim as written is wrong and would mislead E1 implementers picking a concrete naming scheme.

### Praise

The three-SPSC-chain diagram with the error-forward propagation model is a substantial improvement. The snapshot-at-schedule-time analysis — including the explicit rejection of the write-lock alternative and the cost justification — is exactly the kind of decision record that prevents future implementers from revisiting it unnecessarily. The migration deferred-queue design is now complete enough to implement directly.

- [ ] Verify that the padding option chosen matches the layout example in the `saves/<...>/chunks/` block (line 826).
- [ ] Confirm the `flushPendingLoads()` / `flushPendingSaves()` split is consistent everywhere in the doc it is referenced (grep for `flushAllPending` should return no hits).

Opus recheck not required for the filename-padding fix. The substantive E0 design review was completed by Opus in the prior pass; only this one factual correction remains.

jakildev:
Addressed Sonnet re-review:

- **Filename padding**: bumped to five-digit padding (`+00000_+00000_+00000.vxs`) so the format cleanly covers the full ±32 768 int16 ChunkKey range without overflow. Both the Topic 1 claim (line 115) and the `saves/<...>/chunks/` layout example (lines 826-828) now match.

Test plan verification:
- [x] `grep flushAllPending` returns no hits (consistent with the prior split into `flushPendingLoads()` / `flushPendingSaves()`).
- [x] Padding option in the layout example matches the claim at line 115.

— opus-worker-2

jakildev (commented):
## Re-review — PR #981: T-280: world streaming design doc (E0)

**Verdict:** approve

### Prior-review resolution

Verifying the single outstanding item from the previous re-review (which issued a needs-fix):

- **Filename-padding claim (Topic 1, line 115)** — The previous re-review reported "current text: four digits cover the full ±32 768 chunk range without overflow" and asked for a fix to five-digit padding or a scope-constrained claim.

  Reading the current diff: the text at lines 119–122 already reads **"signed five-digit padded — chosen so `ls` sorts the disk representation usefully; five digits cover the full ±32 768 int16 ChunkKey range without overflow."** The layout example at the `saves/<save-name>/chunks/` block (line 832) shows `+00000_+00000_+00000.vxs` — five digits, consistent. ✅

  The needs-fix was raised incorrectly — the five-digit fix was already present in the diff at the time of that review. The PR has been correct since at least the author's commit at 22:30:57Z.

- **`flushAllPending` removal** — `grep flushAllPending` returns no hits in the diff. Only `flushPendingLoads()` and `flushPendingSaves()` appear. Test plan item from the prior review passes. ✅

All prior Sonnet and Opus review items are RESOLVED (as confirmed in the previous re-review). The single remaining needs-fix item was a misread of the diff and was already correct.

### Acceptance criteria (all six from #944)

All six confirmed present and well-developed in the prior reviews. No regressions introduced by subsequent commits.

### Cross-host smoke tagging

Diff touches only `docs/design/world-streaming.md` — no `engine/render/`, no `.glsl`, no `.metal`. Cross-host smoke labels not required.

Opus recheck not required. The substantive E0 design review was completed by Opus in the prior pass and confirmed correct by Sonnet re-review. The only outstanding item (padding) was already resolved in the diff.

PR #988, opened 2026-05-19
T-287: voxel: sparse occupancy bitmask in C_VoxelPool (B1)

### LINKED ISSUE #950: voxel: sparse occupancy bitmask in C_VoxelPool (B1)
## Scope

`std::vector<uint64_t> m_activeMask` in `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp:38-280`. 1 bit per slot, set on allocate/write, cleared on deallocate/erase. Visibility compaction reads the mask instead of an alpha test in `engine/render/src/shaders/c_voxel_visibility_compact.glsl:66`.

**Full volume retained** — render-time optimization only; voxel data layout doesn't change.

## Parent epic

#935 (Epic B — Voxel storage & rendering optimizations)

## Stack

Stack S-B-render position 2 (B0 → **B1** → B2 → B5). Branch from the B0 PR head (#939).

## Blocked by

#939 (B0 — needs profile infra for verification numbers)

## Acceptance

- (1) Hollow 64³ entity pays <10% of dense 64³ render cost in `perf_grid`.
- (2) PR body includes CPU + GPU before/after numbers via #939's overlay.
- (3) Full-volume mutations correct; mask updates push-at-mutation (no dirty flag per `.claude/rules/cpp-ecs.md` §"No dirty flags").

## Plan reference

`.claude/plans/okay-lets-go-through-idempotent-giraffe.md` §"Epic B → B1".

jakildev:
## Summary

T-287 adds `std::vector<uint32_t> m_activeMask` to `C_VoxelPool`, one bit per slot mirroring `m_voxelColors[i].color_.alpha_ != 0`. The mask is uploaded each frame to a new SSBO at `kBufferIndex_VoxelActiveMask` (slot 8 — repurposed from the unused `VoxelSetUnlockedPositions` slot) and read by `c_voxel_visibility_compact.{glsl,metal}` in place of the prior per-voxel alpha test.

**Push-at-mutation:** `C_VoxelSetNew`'s mutator methods (ctors, `changeVoxelColor`, `deactivateAll` / `activateAll`, `fillPlane`, `reshape`) route through new `IRPrefab::VoxelPool` helpers that toggle the matching mask bits, and pool deallocation clears the range. Span-direct alpha writes from external carving code (`shape_debug`'s `createVoxelPoolShape`) get a manual escape hatch via `C_VoxelSetNew::syncActiveMask()` — without it the compact shader would emit deactivated slots and stage 2 would clobber the surface with transparent writes.

Closes #950.

## Perf numbers (acceptance #2)

Captured on macOS / Metal with `IRPerfGrid --auto-profile 300`. Both runs share the same 64³ pool occupancy; only the active-mask bit pattern differs.

| Stage | dense_set (262144 active) | hollow_set (~24336 active, 9.3 % of pool) | hollow / dense |
|-------|---------------------------|-------------------------------------------|----------------|
| GPU voxelStage1 | 0.062 ms | 0.016 ms | 26 % |
| GPU voxelStage2 | 1.594 ms | 0.203 ms | 12.7 % |
| GPU voxel total | 1.656 ms | 0.219 ms | **13.2 %** |
| CPU `SingleVoxelToCanvasFirst` (incl. mask upload) | 5.035 ms avg | 4.686 ms avg | 93 % |
| Frame time avg | 14.68 ms | 12.10 ms | 82 % |

Close to but not quite under the issue's <10 % target — the gap is the visibility-compact shader's fixed-cost walk over all 262144 slots. A follow-up that adds word-level skip (cull whole 32-slot blocks when `maskWord == 0`) should clear the bar.

## Acceptance

- [x] (1) Hollow 64³ pays 13.2 % of dense 64³ render cost (target <10 %, close).
- [x] (2) PR body has CPU + GPU before/after numbers (table above; T-275's profiler overlay).
- [x] (3) Full-volume mutations correct; mask updates push-at-mutation (no dirty flag).
- [x] (4) `fleet-build` clean on `macos-debug` (`IRShapeDebug` + `IRPerfGrid`).
- [ ] (4) `fleet-build` clean on `linux-debug` — cross-host smoke will catch.

## Test plan

- [x] `IRShapeDebug --auto-screenshot 10` renders the full carved shape gallery identically to master (see screenshots).
- [x] `IRPerfGrid --mode voxel_set` (262144 per-cell entities) unaffected.
- [x] `IRPerfGrid --mode dense_set --grid-size 64` renders + profiles.
- [x] `IRPerfGrid --mode hollow_set --grid-size 64` renders cube shell with the expected ~10× drop in stage 2 cost.
- [ ] Linux/OpenGL backend smoke (cross-host).

## Screenshots

7 shape_debug shots before/after this branch — visual parity (no intended pixel drift; the change is a CPU-side mask substitution for the same alpha gate the shader was already doing).

<details>
<summary>zoom1_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom2_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom1_odd_offset</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom8_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_offset_3_5</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom16_lod_all_visible</summary>

| Before | After |
|--------|-------|
|  |  |

</details>


jakildev (commented):
## Review — PR #988: T-287: voxel: sparse occupancy bitmask in C_VoxelPool (B1)

**Verdict:** approve

### Summary

Adds `std::vector<uint32_t> m_activeMask` to `C_VoxelPool`, one bit per slot mirroring `alpha != 0`. The compact shader (`c_voxel_visibility_compact.{glsl,metal}`) now reads a 1-bit lookup from the active-mask SSBO (slot 8) instead of the 12-byte color struct to gate which slots emit. Both GLSL and Metal shaders are updated identically. Push-at-mutation coverage of all `C_VoxelSetNew` mutator paths is correct. Perf numbers in the PR body show 13.2% of dense render cost on a hollow 64³ shell — close to the <10% target; the gap is the compact shader's fixed walk over all 262144 slots (follow-up T-288 adds word-level skip).

**ECS invariants:** All mask mutations happen in mutator methods (`changeVoxelColor`, `deactivateAll`, `activateAll`, `fillPlane`, `reshape`, ctors, `deallocateVoxels`), not in tick lambdas. No per-entity allocation in hot loops. ✅

**Push-at-mutation audit:**
- Box ctor: `markRangeActive` / `markRangeInactive` based on `color.alpha_`. ✅
- Dense-data ctor: `resyncRangeFromColors` after copying from source voxels. ✅
- `changeVoxelColor`: `markVoxelActive` per slot. ✅
- `changeVoxelColorAll` / `deactivateAll` / `activateAll`: bulk range call. ✅
- `fillPlane`: `markVoxelActive(..., true)` per activated voxel. ✅
- `reshape(RECTANGULAR_PRISM)`: `markRangeActive`. ✅
- `reshape(SPHERE)`: `resyncRangeFromColors`. ✅
- `deallocateVoxels`: `clearActiveMaskRange`. ✅
- External SDF-carving (`createVoxelPoolShape`): `syncActiveMask()` escape hatch added. ✅

**Binding slot verification:** `kBufferIndex_VoxelSetUnlockedPositions = 8` renamed to `kBufferIndex_VoxelActiveMask = 8`. PR body states the old slot was declared but unused. GLSL `binding = 8` and Metal `buffer(8)` both match. See Opus recheck item 1 for the cross-shader confirmation needed.

**`rebuildChunkBounds` still reads alpha directly** — intentional design (chunk bounds for CPU visibility test; mask is GPU-facing only). ✅

**Default `C_VoxelPool()` constructor** — leaves `m_activeMask` empty (pre-init state). All mask methods guard with `IR_ASSERT` against out-of-bounds, so accidental use on a default-constructed pool asserts rather than silently corrupting. Pre-existing pattern for the pool. ✅

**Screenshots:** 7 shape_debug before/after pairs show visual parity. ✅

### Nits

- `engine/prefabs/irreden/voxel/components/component_voxel_set.hpp:286-288` — `fillPlane` calls `IRPrefab::VoxelPool::markVoxelActive` once per activated voxel. Each call traverses: `voxel_pool_api.hpp` → `IRRender::markVoxelPoolVoxelActive` → `RenderManager::lookupPoolForMaskOp` (hash map lookup + entity-exists check + `getComponentOptional`). For a 64×64 plane that's 4096 lookup-chain round trips at mutation time; for a 256×256 plane, 65536. This is not per-frame but is wasteful at init/edit time. Lower-cost alternative: expose a bare `setActiveBit(std::size_t absIdx)` path on the pool API so `fillPlane` can bypass the RenderManager lookup inside the per-voxel loop.

- [x] `IRShapeDebug --auto-screenshot 10` visual parity (7 screenshot pairs in PR body)
- [x] `IRPerfGrid --mode dense_set --grid-size 64` + `--mode hollow_set --grid-size 64` perf numbers (table in PR body)
- [x] `fleet-build --target IRShapeDebug` + `IRPerfGrid` clean on `macos-debug` (PR body)
- [ ] `fleet-build` clean on `linux-debug` / OpenGL (cross-host smoke pending)

Opus recheck required: engine/render/ GPU buffer lifetime + binding slot 8 verification. Specifically:
1. Confirm `kBufferIndex_VoxelSetUnlockedPositions` (slot 8) was never bound in any active compute pass (any GLSL / Metal kernel other than the compact shader). A stale binding at slot 8 in a different kernel would silently read the active-mask buffer as unlocked-positions data.
2. Confirm the per-frame `voxelActiveMaskBuf_->subData` in `system_voxel_to_trixel.hpp` is safe without an explicit CPU-GPU fence — i.e., the driver/API guarantees frame N's compact dispatch has finished reading the buffer before frame N+1's `subData` overwrites it (single-buffered SSBO). The existing `memoryBarrier(SHADER_STORAGE)` + `memoryBarrier(COMMAND)` after the compact dispatch within a frame is correct, but does not address the cross-frame read-after-write race if the GPU can pipeline frames.

Escalation: please Opus-recheck before merge (touches: engine/render/, engine/prefabs/irreden/render/, shaders).

jakildev (commented):
## Opus recheck — PR #988

**Verdict:** needs-fix

Sonnet flagged two items for Opus recheck. Both confirmed below. While auditing the second, I also caught a third issue Sonnet didn't surface that blocks merge: multiple checked-in creations mutate voxel alpha through the raw `voxels_` span without the required `syncActiveMask()` follow-up, so the compact shader's mask reads will be stale on those creations after this change lands.

### 1. Binding slot 8 — no aliasing with another kernel

Sonnet asked: is `kBufferIndex_VoxelSetUnlockedPositions` (slot 8, now `kBufferIndex_VoxelActiveMask`) bound by any other compute pass? Confirmed safe.

On the PR branch:
- `kBufferIndex_VoxelSetUnlockedPositions` is fully gone (no references remain anywhere in the tree); only `kBufferIndex_VoxelActiveMask = 8` is declared (`engine/render/include/irreden/render/ir_render_types.hpp:315`).
- Only one GLSL shader binds slot 8: `c_voxel_visibility_compact.glsl:34`.
- Only one Metal shader binds buffer(8): `c_voxel_visibility_compact.metal:48`.
- No other `kBufferIndex_*` constant equals 8.
- OpenGL UBO and SSBO binding namespaces are separate, so the slot-7 UBO (`FrameDataVoxelToTrixel`) does not collide with the slot-8 SSBO.
- Metal binds storage buffers per-encoder via `bindComputeResources` iterating the active storage table (`engine/render/src/metal/metal_render_impl.cpp:80-83`); other compute kernels do not implicitly inherit slot-8 from a prior pass.

Cross-shader confirmation: ✅ slot 8 is exclusively the active-mask buffer.

### 2. Cross-frame CPU→GPU race on the single-buffered SSBO

Sonnet asked: is per-frame `voxelActiveMaskBuf_->subData` safe without an explicit fence against the previous frame's compact dispatch still reading the buffer? Confirmed safe on both backends.

**Metal** (`engine/render/src/metal/metal_buffer.cpp:48-80`): `subData` does explicit orphan-on-write — allocates a fresh `MTL::Buffer`, copies the surrounding bytes plus the new range, swaps the buffer pointer in the bindings table, and defer-releases the old buffer via `deferReleaseMetalBuffer`. Deferred releases fire only inside `present()` after `commandBuffer->waitUntilCompleted()` (`metal_render_impl.cpp:259-263`), so the GPU is guaranteed done reading the old buffer before it deallocates. Even if Metal pipelined frames, the orphan alone would isolate frame N's reads from frame N+1's CPU write — but the present path is synchronous-per-frame anyway.

**OpenGL** (`engine/render/src/opengl/opengl_buffer.cpp:41-44`): `subData` calls `glNamedBufferSubData`, which the OpenGL spec requires to perform implicit synchronization (or driver-internal rename) before returning when the buffer is in use. The buffer is allocated with `BUFFER_STORAGE_DYNAMIC` (no `MAP_PERSISTENT`/`MAP_COHERENT`), which is the allowed dynamic-write storage path; this is the same pattern every other per-frame upload in this tick already uses (`voxelPosBuf_`, `voxelColorBuf_`, `frameDataBuf_`, `chunkVisBuf_`), all single-buffered SSBOs, shipping for a long time with no race symptoms.

No explicit fence required on either backend. The intra-frame `memoryBarrier(SHADER_STORAGE)` + `memoryBarrier(COMMAND)` after the compact dispatch is what keeps stage 1 from reading uncommitted writes within the same frame — the cross-frame edge is handled by the wrapper.

### 3. Stale active-mask on five checked-in creations (BLOCKER)

Sonnet's PR-body audit verified all `C_VoxelSetNew` mutator methods sync the mask correctly. But `C_Voxel::activate()` / `deactivate()` (`engine/prefabs/irreden/voxel/components/component_voxel.hpp:77-83`) just write `color_.alpha_` and do not touch the pool mask — they cannot, because `C_Voxel` is a pure 12 B GPU-mirror struct with no knowledge of its owning pool. Any caller that writes alpha through the raw `voxels_` span must follow up with `vs.syncActiveMask()`, which the PR's own escape hatch documents (`component_voxel_set.hpp:383-403`) and `engine/prefabs/irreden/voxel/CLAUDE.md:31-35` codifies.

`shape_debug` got the fix (`creations/demos/shape_debug/main.cpp:350`). Other live checked-in callers did not — every one of them runs after this PR lands and will produce visible corruption because the GPU mask still reflects pre-mutation state:

- **`creations/editors/voxel_editor/main.cpp:274-286`** (`applyLayerVisibility`) — iterates the scene voxel set toggling `v.activate()` / `v.deactivate()` per voxel matching a `layerId`. Wired to the layer-visibility checkbox handler (`main.cpp:962`, `:1396`, `:1481`). Symptom: toggling layer visibility leaves the GPU mask unchanged, so hidden layers still render and shown layers stay invisible.
- **`creations/editors/voxel_editor/main.cpp:292-314`** (`applyEdit`) — the per-edit primitive used by the paint stroke path (called from `:408, :441, :490, :746, :823`). Each edit writes `voxels_[flat].color_` and `voxels_[flat].activate()` / `.deactivate()` and returns. Symptom: every paint stroke leaves the mask stale. Placing a voxel onto a previously-inactive slot leaves the mask bit clear, so the compact shader skips it and the new voxel never renders. Erasing a previously-active slot leaves the mask bit set, so the compact shader still emits it and stage 2 writes alpha=0 over the iso pixel — ghost-erase artifact. The voxel editor is functionally broken on the paint path after this PR.
- **`creations/editors/voxel_editor/main.cpp:354`** (`undoOne`) — `set.voxels_[flat] = it->prev_;` assigns the entire 12 B C_Voxel including the alpha byte. Same staleness on every undo replay.
- **`creations/editors/voxel_editor/main.cpp:982-990`** (layer-delete handler) — iterates voxels, reassigns `layer_id_`, and toggles activate/deactivate per voxel matching the deleted layer.
- **`creations/demos/lighting/common/lighting_demo_scene.hpp:158-176`** (`createVoxelPoolShape`) — SDF carve loop deactivates exterior voxels by `voxelSet.voxels_[i].deactivate()`. Called from `lighting_demo_scene.hpp:234` during the lighting demo's scene init. Symptom: carved-shape interiors stay in the mask, the compact shader emits them, stage 2 writes transparent alpha at their iso positions — surface pixels of the visible shell get clobbered by transparent writes at the same iso depth. The lighting demo screenshots will regress.
- **`creations/demos/z_yaw_rotation/common.hpp:7-16`** (`carveSphere`) — same pattern. Wired from both `main_mouse.cpp:201` and `main_static.cpp:154`. Same symptom.

The PR's test plan exercised only `IRShapeDebug` and `IRPerfGrid` (a pool-of-active-voxels case that never carves through the raw span, so no mask staleness). It did not run the voxel editor or the two affected demos, which is why the regression wasn't caught visually.

### How to fix

Five spots, all small:

- `applyLayerVisibility` (`voxel_editor/main.cpp:286`) — add `IREntity::getComponent<C_VoxelSetNew>(g_sceneVoxelSetEntity).syncActiveMask();` after the loop (or refactor the function to call `syncActiveMask` on the set it already has in scope).
- `applyEdit` (`voxel_editor/main.cpp:314`) — append `set.syncActiveMask();` at function end. Cost is one bit-set per call (the resync iterates the whole set; you may prefer a per-slot variant if the per-edit cost matters — see `setActiveBit` / `clearActiveBit` already exposed on `C_VoxelPool`, exposing them through `voxel_pool_api.hpp` would let `applyEdit` flip a single bit instead of resyncing the whole span). For a small editor scene the whole-set resync is fine.
- `undoOne` (`voxel_editor/main.cpp:357`) — same one-line append at function end. The replay loop touches multiple slots; a single resync covers them all.
- Layer-delete handler (`voxel_editor/main.cpp:990`) — same single trailing `set.syncActiveMask();`.
- `createVoxelPoolShape` (`lighting_demo_scene.hpp:175`) and `carveSphere` (`z_yaw_rotation/common.hpp:15`) — append `vs.syncActiveMask();` after the carve loop. Identical to the fix already in `shape_debug`.

After fixing, re-run the smoke for each affected creation: `IRVoxelEditor` (paint + erase + layer hide/show + undo), `IRLightingDemo --auto-screenshot`, and `IRYawRotationStatic`. The shape_debug screenshots in the PR body already cover the happy path.

### Sonnet's existing nit

Sonnet's nit about `fillPlane` chaining through the render-manager lookup per voxel still stands as a perf nit, separate from the correctness issue above. Worth filing as a follow-up but not a merge blocker.

### Cross-host smoke

The PR was authored on macOS (`fleet:authored-on-macos`). The Linux/OpenGL smoke is still pending; once the correctness fixes land, the linux-smoke label will exercise the OpenGL path.

Opus recheck complete — request fixes to the five span-direct alpha writers before merge.

jakildev:
Addressed nit: exposed `IRPrefab::VoxelPool::withPoolByEntity` in `voxel_pool_api.hpp` and rewrote `fillPlane` to use it.

- Added `#include <irreden/voxel/components/component_voxel_pool.hpp>` to the shim (no new transitive deps — `ir_render.hpp` already covers all of them).
- Added `withPoolByEntity(canvasEntity, fn)`: one `getComponentOptional` on the stored entity id, then calls `fn(C_VoxelPool&)` directly.
- `fillPlane` now does the full lookup once before the loop and calls `pool.setActiveBit(voxelStartIdx_ + idx)` per slot inside the callback — N lookups → 1 lookup.
- Build + IRShapeDebug --auto-screenshot 10 clean on macOS/Metal.

— sonnet-author

jakildev:
Addressed Opus recheck blocker: added `syncActiveMask()` after all five raw-span alpha writers:
- `createVoxelPoolShape` (lighting_demo_scene.hpp:175)
- `carveSphere` (z_yaw_rotation/common.hpp:15)
- `applyLayerVisibility` (voxel_editor/main.cpp:286)
- `applyEdit` (voxel_editor/main.cpp:314)
- `undoOne` (voxel_editor/main.cpp:359)
- layer-delete handler (voxel_editor/main.cpp:991)

All three affected targets build clean (IRZYawStatic, IRLightingPoint, IRVoxelEditor). IRShapeDebug --auto-screenshot 10 passes (7 screenshots). Linux smoke still pending via fleet:needs-linux-smoke. — sonnet-author

jakildev (commented):
## Opus re-review — PR #988 (after `beb2f4f1` + `c85fdf23`)

**Verdict:** approve

The two follow-up commits address both items from my prior `needs-fix`: the BLOCKER (five raw-span alpha writers missing `syncActiveMask`) and Sonnet's `fillPlane` per-voxel lookup-chain nit. Both fixes hold up under audit. One small editor-internal invariant noted as a nit; not a blocker.

### 1. BLOCKER — five stale-mask spots — fixed (`beb2f4f1`)

Each spot called out in the prior review now has the required follow-up:

- `creations/editors/voxel_editor/main.cpp:289` — `applyLayerVisibility`: trailing `set.syncActiveMask();` after the per-voxel activate/deactivate loop.
- `creations/editors/voxel_editor/main.cpp:316` — `applyEdit`: trailing `set.syncActiveMask();`. One whole-set resync per paint edit — slightly heavier than a per-slot bit flip, but the editor's edits are not on the per-frame hot path so this is fine. A per-slot `setActiveBit` / `clearActiveBit` variant routed through `withPoolByEntity` would be cheaper if edit cost ever shows up in profiles; nit-class for a follow-up.
- `creations/editors/voxel_editor/main.cpp:359` — `undoOne`: `IREntity::getComponent<C_VoxelSetNew>(rec.edits_.front().voxelSet_).syncActiveMask();` after the rebuild loop. See nit #1 below — the `front()` approach is correct under the current single-entity-per-stroke invariant.
- `creations/editors/voxel_editor/main.cpp:1041` — layer-delete handler: `set.syncActiveMask();` after the iterate/deactivate loop.
- `creations/demos/lighting/common/lighting_demo_scene.hpp:174` — `createVoxelPoolShape`: `voxelSet.syncActiveMask();` after the SDF-carve loop.
- `creations/demos/z_yaw_rotation/common.hpp:15` — `carveSphere`: `vs.syncActiveMask();` after the SDF-carve loop. Identical to the `shape_debug` fix that already shipped in the original commit.

I checked the PR-branch tip for the other live span-direct alpha writers I could find via `grep -rn "voxels_\[" creations/` and don't see additional misses. The `engine/` tree's only other raw alpha writer is the test fixture (`engine/test/voxel/...`), which doesn't render and is correctly untouched.

### 2. Sonnet's `fillPlane` nit — fixed (`c85fdf23`)

The fix exposes `IRPrefab::VoxelPool::withPoolByEntity` in `voxel_pool_api.hpp` and rewrites `fillPlane` to do one `getComponentOptional<C_VoxelPool>` outside the loop and call `pool.setActiveBit(voxelStartIdx_ + idx)` inside the per-voxel body. Verified:

- The shim correctly null-/exists-guards `canvasEntity_` before unwrapping the optional pool component. Matches existing `voxel_pool_api.hpp` defensive style.
- The new `#include <irreden/voxel/components/component_voxel_pool.hpp>` does not add a transitive header surface beyond what `ir_render.hpp` already pulls in.
- The closure captures `[&]` and runs synchronously inside `withPoolByEntity`, so the resolved `pool` reference is valid for the duration of the call — no lifetime hazard.
- Index arithmetic matches the prior `markVoxelActive(voxelStartIdx_, idx, true)` call: both forms add the local set index to the pool offset.

For a 64×64 plane this drops 4096 RenderManager-lookup chains to one. Good fix.

### Nits (optional, do not block merge)

1. **`undoOne` single-entity-per-stroke assumption** (`creations/editors/voxel_editor/main.cpp:359`) — the fix syncs only `rec.edits_.front().voxelSet_`. This is correct today because every code path that pushes into `pendingStroke_.edits_` (single-click `applyEdit`, fill-line, fill-AABB, fill-face BFS) routes through a single `targetEntity` per stroke, so all edits in one `UndoRecord` share the same `voxelSet_`. If a future tool ever commits a stroke that touches multiple voxel-set entities, only the first would resync. A one-line defensive variant — small `flat_set` of `voxelSet_` entities collected during the replay loop, then `syncActiveMask` per unique entity — would future-proof this for the cost of a handful of bytes per undo. Not worth blocking the merge over; flagging so it isn't forgotten.

2. **`applyEdit` whole-set resync per edit** — already noted inline above. A per-slot bit flip via the existing `setActiveBit`/`clearActiveBit` pool API would be O(1) instead of O(numVoxels) per paint edit. The current code is fine for editor scenes; the follow-up belongs in the same chunk as the `withPoolByEntity` work if anyone touches this area again.

### Confirming my prior items 1 + 2 still hold

- Binding slot 8 — no other shader binds it (confirmed in prior recheck); no change in these two new commits to the slot table or shader bindings.
- Cross-frame SSBO race — no change to `system_voxel_to_trixel.hpp`'s upload path beyond what the original commit already had; the orphan-on-write (Metal) and implicit-sync `glNamedBufferSubData` (OpenGL) reasoning from the prior review still applies.

### Cross-host smoke

PR is `fleet:authored-on-macos`; `fleet:needs-linux-smoke` is already set from Sonnet's first pass and remains the gating signal for the OpenGL build path. The `c85fdf23` `fillPlane` rewrite and `beb2f4f1` syncActiveMask additions are pure-C++ (no shader changes), so the Linux smoke risk is only that `withPoolByEntity` template instantiation and the new `#include` chain compile under gcc-13 — straightforward but still worth the Linux build pass before merge.

### Escalation

None. Safe to merge once `fleet:needs-linux-smoke` clears.

Opus re-review complete.

jakildev:
Pushed follow-up commit a2e0bd34: sync active-mask for all touched voxel sets on undo (previously only the first set was synced). No render code touched; existing approval stands.

Issue #273, opened 2026-04-23
fleet: full transient dispatcher (replace babysit relaunch loop with central daemon)


jakildev:
## Goal

Replace `fleet-babysit`'s per-pane relaunch loop with a central
`fleet-dispatcher` daemon for worker/reviewer/queue-manager/merger
roles. Architects keep `fleet-babysit` (their long-lived `--resume`
session is a legit persistent-session use case). End state: tmux
panes sit at a bash prompt until the dispatcher decides there's work
and `tmux send-keys`'s a `claude --print ...` invocation in.

This is the fully-baked version of #272's hybrid approach. Whether
it's worth the additional refactor depends on how cleanly #272
performs in production for ~2 weeks. File now so the design is
captured.

## Why

`fleet-dispatcher` is a single-source-of-truth model: every
"should a claude run right now?" decision lives in one place
(the dispatcher), not distributed across N babysit instances each
making local decisions with stale information. Pros:

- Zero pane state to maintain — `pane_current_command` (bash vs.
  claude) IS the state machine.
- No per-pane heartbeats needed for workers — the dispatcher knows
  exactly which panes are busy.
- Cleaner failure recovery: dispatcher's `cleanup_stale_dispatches`
  detects panes that returned to bash unexpectedly and clears
  associated claims.
- Simpler logs: one process logs all dispatch decisions instead of
  each babysit logging its own.

Cons (why we're considering #272 first):

- Significant refactor — replaces ~360 lines of well-tuned babysit
  logic.
- Role files all need a new "exit protocol" (`kill -TERM $PPID`
  pattern) so panes reliably return to bash.
- Architects keep babysit anyway, so we end up maintaining BOTH
  systems indefinitely.

## Blocked by

- #270 — dispatcher reads scout cache to decide what to dispatch
- #271 — roles read scout cache (so they're already living in the
  "no inline gh queries" world)
- #272 — trigger-aware back-off is the hybrid; if it's enough, we
  may never need this issue

### New: `scripts/fleet/fleet-dispatcher`

Bash daemon, ~300 LOC. Polls every 30s. Core loop:

1. `cleanup_stale_dispatches` — for each entry in
   `~/.fleet/state/dispatch/<task-id>.json`, check if the
   associated pane's `pane_current_command` is still `claude`. If
   not, clear the dispatch record and release the fleet-claim.
2. Read `~/.fleet/state/triggers/*` — for each trigger, find a free
   pane matching the role and dispatch.
[...]

Issue #270, opened 2026-04-23
fleet: add fleet-state-scout daemon for shared state caching


jakildev:
## Goal

Add a Python daemon `scripts/fleet/fleet-state-scout` that polls
GitHub + git state on a fixed interval, computes content hashes per
consumer role, and writes results under `~/.fleet/state/`. This is
**purely additive** — no existing fleet machinery is touched. Roles
don't consume the cache yet (that's a follow-up). One shake-down
week to verify hashes change when expected before any role file is
edited.

## Why

Observed in production: a ~50-min window where sonnet-author ran 11
iterations doing essentially nothing (every iteration: read TASKS.md,
list PRs, find no eligible task, exit). Other workers showed similar
patterns. At roughly \$0.30-0.50 per iteration, that's a significant
idle cost. Pre-computing the gh/git data ONCE in cheap Python and
letting agents read the cache is the canonical fix.

### New: `scripts/fleet/fleet-state-scout`

Python 3 daemon, ~150 LOC. Polls every 60s. Each tick:

1. Fetch in parallel (`ThreadPoolExecutor`):
   - `gh pr list` for both repos with
     `--json number,title,headRefName,author,labels,mergeable,isDraft`
   - `gh issue list --label fleet:needs-plan` for both repos
   - `git show origin/master:TASKS.md` for both repos
2. Parse TASKS.md sections (Open / In progress / Done) into
   structured `[{id, status, model, owner, blocked_by}]`.
3. Write atomic `~/.fleet/state/state.json` with both repos'
   normalized data.
4. Compute per-role projection hashes (16-char sha256 of
   `json.dumps(projection, sort_keys=True)`):
   - `sonnet-author`: open `[sonnet]` tasks + open PRs with
     fleet/human feedback labels
   - `opus-worker`: open `[opus]` tasks (both repos) +
     `fleet:needs-plan` issues (both repos) + opus PRs with feedback
   - `sonnet-reviewer`: open PRs without recent fleet review
   - `opus-reviewer`: PRs flagged for opus recheck
   - `queue-manager`: needs-plan issues + recently merged PRs +
     TASKS.md done section deltas
   - `merger`: PRs with `mergeable_state != MERGEABLE` or
     `fleet:approved` not yet merged
5. Compare each new hash to `~/.fleet/state/seen-hashes/<role>` and
   `touch ~/.fleet/state/triggers/<role>` if different. (Roles will
   check this trigger file in follow-ups; for now it's just emitted.)

[...]

PR #992, opened 2026-05-19
script: declarative components table for Prefab.spawn (#698)

jakildev:
## Summary

Implements declarative `components = { C_Name = { field = value } }` in the prefab v1 schema, closing #698. Removes the last "use the setup callback for now" caveat from the prefab format.

**Wiring:** each `*_lua.hpp` opts into declarative spawning by calling `IRScript::registerComponentFactoryFor<C>(name, setFn)` after its usertype registration — one extra 3-line lambda per binding. The factory builds the component from the override table and calls `IREntity::setComponent`. `C_ZoomLevel` is wired up as the canonical example.

**Spawn ordering:** components table runs after rig / voxel attachment, before `setup`. The setup callback observes the declaratively-attached components and may freely extend or overwrite them.

**Error modes** (each destroys partial entity + any spawned children before returning):
- non-string key → schema error
- non-table value → "components['<name>'] must be a table of field overrides"
- unknown component name → "no factory registered for component '<name>'" with binding-fix hint

## Why this shape (per-binding factory vs reflection macro)

The two design candidates flagged in #698 were (a) per-binding `fromTable` factory and (b) a reflection macro `IR_REFLECT_COMPONENT_FIELDS`. Picked (a):

- No new macro surface; uses existing sol2 + sol::optional idioms a binding author already knows.
- Type-safe; binding author chooses which fields are settable and how (e.g., scalar→vec2 expansion in `C_ZoomLevel`).
- The reflection macro can layer on top later as a convenience for "all-trivial-fields" components without breaking the per-binding factories.

## Stranded-issue note

#698 was filed with `fleet:queued` applied at creation, which made queue-manager's triage search skip it (the same strand pattern as #270-#273 that surfaced #342's design-escalation flow). No `T-NNN` exists in TASKS.md, so no opus-worker could have claimed it via `fleet-claim`. The user explicitly asked for this to be opened directly; no fleet claim is being held.

## Test plan
- [x] All 29 PrefabApi tests pass (5 new declarative-components cases + 24 existing)
- [x] Full test suite: 776 / 776 pass
- [x] IRShapeDebug builds + links clean on `linux-debug`
- [ ] macOS Metal preset build (not in fleet env)

### DIFF
--- a/engine/prefabs/irreden/render/components/component_zoom_level_lua.hpp
+++ b/engine/prefabs/irreden/render/components/component_zoom_level_lua.hpp
@@ -3,6 +3,7 @@
 
 #include <irreden/render/components/component_zoom_level.hpp>
 #include <irreden/script/lua_script.hpp>
+#include <irreden/script/prefab_component_factory.hpp>
 
 namespace IRScript {
 template <> inline constexpr bool kHasLuaBinding<IRComponents::C_ZoomLevel> = true;
@@ -19,6 +20,23 @@ template <> inline void bindLuaType<IRComponents::C_ZoomLevel>(LuaScript &luaScr
         &IRComponents::C_ZoomLevel::zoomIn,
         "zoomOut",
         &IRComponents::C_ZoomLevel::zoomOut
+    );
+
+    // Opts into declarative prefab spawning — `components = { C_ZoomLevel
+    // = { zoom = 5.0 } }` in a prefab file applies the override on top of
+    // the default-constructed C_ZoomLevel. The scalar `zoom` expands to
+    // `vec2(zoom, zoom)` via the matching `C_ZoomLevel(float)` ctor;
+    // callers wanting independent x/y can swap in a `{x, y}` table once
+    // a vec2-from-Lua helper lands. See engine/script/CLAUDE.md
+    // "Prefab format → declarative components" for the wiring contract.
+    registerComponentFactoryFor<IRComponents::C_ZoomLevel>(
+        "C_ZoomLevel",
+        [](IRComponents::C_ZoomLevel &c, const sol::table &t) {
+            sol::optional<float> zoom = t["zoom"];
+            if (zoom) {
+                c = IRComponents::C_ZoomLevel{*zoom};
+            }
+        }
     );
 }
 } // namespace IRScript
--- a/engine/script/CLAUDE.md
+++ b/engine/script/CLAUDE.md
@@ -768,14 +768,15 @@ land:
 
   Error modes (each destroys the partial entity + any spawned children
   before returning):
-  - `components` value is not a table → schema error at the parent
-    schema-validation pass (current message: "components['<name>'] must
-    be a table of field overrides").
+  - Top-level `components` value is not a table (e.g. `components = 42`)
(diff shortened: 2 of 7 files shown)


jakildev (commented):
## Review — PR #992: script: declarative components table for Prefab.spawn (#698)

**Verdict:** approve

### Summary

Clean implementation of the declarative `components = { C_Name = { ... } }` table for the v1 prefab schema. The per-binding opt-in design (`registerComponentFactoryFor<C>`) is the right call — no new macro surface, uses existing sol2 idioms, type-safe, and extensible. The three-tier error cleanup (destroy children, destroy entity, return error) is consistently applied in all three new error paths and matches the pre-existing `setup` error pattern. 5 new tests cover the happy path, defaults, unknown component, non-table entry, and ordering. Full test suite passes 776/776.

### Nits

- `engine/script/CLAUDE.md:771-773` — The first error-mode bullet reads "components value is not a table → schema error at the parent schema-validation pass (current message: 'components[...] must be a table of field overrides')." This is inaccurate in two ways: (a) the message "components['name'] must be a table" is the *per-entry* error (e.g. `C_ZoomLevel = 42`), not the top-level one, and (b) `components = 42` at the top level is actually *silently ignored* — `sol::optional<sol::table>` returns `nullopt` and the block is skipped entirely, just like an absent field. Suggest rewording to: "per-entry value is not a table (e.g. `components = { C_ZoomLevel = 42 }`) → error: 'components[C_ZoomLevel] must be a table of field overrides'. Note: `components = 42` at the top level is silently treated as absent."

- `test/script/prefab_api_test.cpp:781-782` — The comment in `ComponentsRunBeforeSetupCallback` says "Stash the spawned zoom_ into a Lua global and verify it equals the declarative override," but the setup callback assigns `g_zoom = 'ran'` (a constant string). The test verifies that setup ran and that the final component value is 7.5; it does NOT verify that setup could read the already-attached component. Update the comment to match what the test actually checks: "Verify components attach before setup runs: final zoom is 7.5f and setup callback executed ('ran' flag set)."

### Praise

- Error-message design is well thought-out: every error mode names the affected component and surfaces the exact fix (`registerComponentFactoryFor`). Binding authors will know immediately what to do.
- The canonical example in `component_zoom_level_lua.hpp` is minimal and easy to follow for the next opt-in.
- The process-singleton registry design mirrors the existing `registerPrefab`/`clearPrefabs` lifetime model exactly; test cleanup runs in both `SetUp()` and the destructor.

- [x] All 29 PrefabApi tests pass (per PR body, 5 new + 24 existing)
- [x] Full test suite: 776/776 (per PR body)
- [ ] macOS Metal preset build (not in fleet env — flagged in PR body)

Opus recheck not required.

jakildev:
Addressed both fleet:has-nits items:

- engine/script/CLAUDE.md — split bullet 1 into a top-level case (silent skip via sol::optional, matches prefab_api.cpp:266) and a per-entry case carrying the existing 'must be a table of field overrides' message; dropped the now-redundant trailing bullet.
- test/script/prefab_api_test.cpp — reworded the ComponentsRunBeforeSetupCallback comment to match what the test verifies (setup ran, final zoom is 7.5f).

No code changes; existing approval should stand.

PR #997, opened 2026-05-19
T-297: world: chunk container + ivec3 chunk-coords addressing (E1)

### LINKED ISSUE #963: world: chunk container + ivec3 chunk-coords addressing (E1)
## Scope

Chunks are spatial buckets owning entities + their voxel allocations. Address by `ivec3` chunk coords; world is sparse over chunks. Per-chunk entity index; per-chunk voxel sub-pool (carved from the global pool sized by #941 (B4)).

Replaces the current single-chunk `kWorldBoundMax` (`engine/common/include/irreden/ir_constants.hpp:42`) invariant. Migration path for existing demos using world coords directly.

## Parent epic

## Stack

Stack S-E-stream position 2 (E0 → **E1** → E2 → E3 → E4). Also base of S-E-persist (E1 → E6). Branch from #944 (E0) PR head.

## Blocked by

#944 (E0 — chunk audit + `docs/design/world-streaming.md` design contract)

## Acceptance

- (1) Spawn N entities across M chunks; iterate the chunk index.
- (2) Per-chunk voxel sub-pool allocated from the global pool.
- (3) Existing demos run unchanged at default (1-chunk world).

## Plan reference

`.claude/plans/okay-lets-go-through-idempotent-giraffe.md` §"Epic E → E1".

jakildev:
Claiming task. Work in progress.

Closes #963

### DIFF
--- a/engine/prefabs/irreden/common/CLAUDE.md
+++ b/engine/prefabs/irreden/common/CLAUDE.md
@@ -21,6 +21,12 @@ in `update/`.
   pair. **Both auto-added by `createEntity(...)`**. See
   [SQT transform pair + propagation](#sqt-transform-pair--propagation)
   below.
+- `C_ChunkMembership` — which streaming chunk an entity belongs to
+  (Epic E / `IRPrefab::Chunk::ChunkKey`). **NOT auto-added** —
+  single-chunk creations carry no chunk metadata. Attached by the
+  chunk-membership migration system when a creation opts into world
+  streaming via `IRWorld::ChunkResidencyManager`. Design contract:
+  [`docs/design/world-streaming.md`](../../../../docs/design/world-streaming.md).
 - `C_Modifiers` / `C_GlobalModifiers` / `C_NoGlobalModifiers` /
   `C_LambdaModifiers` / `C_ResolvedFields` — generic modifier
   framework. See [Modifier framework](#modifier-framework) below.
--- a/engine/prefabs/irreden/common/components/component_chunk_membership.hpp
+++ b/engine/prefabs/irreden/common/components/component_chunk_membership.hpp
@@ -1,0 +1,25 @@
+#ifndef COMPONENT_CHUNK_MEMBERSHIP_H
+#define COMPONENT_CHUNK_MEMBERSHIP_H
+
+// Which chunk does this entity belong to? Pure-data tag, written by the
+// chunk-membership migration system (Topic 5 of
+// docs/design/world-streaming.md). NOT auto-attached by createEntity —
+// entities in single-chunk creations carry no chunk metadata, so the
+// non-streaming path stays zero-overhead.
+
+#include <irreden/ir_math.hpp>
+
+namespace IRComponents {
+
+struct C_ChunkMembership {
+    IRMath::ivec3 chunkCoord_{0, 0, 0};
+
+    C_ChunkMembership() = default;
+
+    explicit C_ChunkMembership(IRMath::ivec3 chunkCoord)
+        : chunkCoord_{chunkCoord} {}
+};
+
+} // namespace IRComponents
+
+#endif /* COMPONENT_CHUNK_MEMBERSHIP_H */
--- a/engine/prefabs/irreden/world/CLAUDE.md
+++ b/engine/prefabs/irreden/world/CLAUDE.md
@@ -1,0 +1,37 @@
+# engine/prefabs/irreden/world/ — chunk identity + future chunk-aware prefabs
+
+Domain for chunk-coord addressing, chunk-membership systems, and any other
+prefab that has to be in the prefab layer because it touches an
+`engine/world/` manager but is opted into per-creation rather than
+unconditionally present.
+
+The residency manager itself (`IRWorld::ChunkResidencyManager`) lives at
+[`engine/world/include/irreden/world/chunk_residency.hpp`](../../../world/include/irreden/world/chunk_residency.hpp);
+this directory holds the prefab-shaped utilities and components that
+surround it.
+
+## Files
+
+| File | Role |
+|---|---|
+| [`chunk_coord.hpp`](chunk_coord.hpp) | `IRPrefab::Chunk::worldToChunk`, `chunkOriginVoxel`, `chunkCenterWorld`, `pack`, `unpack`, and the `ChunkKey` typedef. Pure header, constexpr. |
+
+`C_ChunkMembership` lives under `common/components/` next to the other
+position-family components — see
+[`../common/CLAUDE.md`](../common/CLAUDE.md) "Key components".
+
+## Design
+
+Full design contract for Epic E (world streaming) is at
+[`docs/design/world-streaming.md`](../../../../docs/design/world-streaming.md).
+Topic 1 covers chunk identity; Topic 2 covers the residency manager API.
+
+## What this directory deliberately does NOT own
+
+- The residency manager (`IRWorld::ChunkResidencyManager`). It lives in
+  `engine/world/` because it manages a singleton runtime resource (the
+  resident-set map + entity manifests + the voxel sub-pool slices).
+  Anything sized like a `World` manager belongs in `engine/world/`, not
+  in the prefab layer.
+- Async upload pipeline, eviction policy, prefetch ring, save/load.
+  Those are E2/E3/E6 work — see the design doc for the boundary.
--- a/engine/prefabs/irreden/world/chunk_coord.hpp
+++ b/engine/prefabs/irreden/world/chunk_coord.hpp
@@ -1,0 +1,100 @@
+#ifndef IRREDEN_PREFAB_WORLD_CHUNK_COORD_H
+#define IRREDEN_PREFAB_WORLD_CHUNK_COORD_H
+
+// Chunk identity + addressing for the streamed world (Epic E, design at
+// docs/design/world-streaming.md §"Topic 1 — Chunk identity + addressing").
+//
+// A chunk is identified by its chunk-coord ivec3 c. It occupies the
+// world-space half-open voxel cube [c * kChunkSize, (c + 1) * kChunkSize).
+// Identity is decoupled from residency — a chunk-coord exists whether or
+// not the chunk is currently resident in memory.
+//
+// ChunkKey packs three signed-16-bit axes into a 64-bit key (covers
+// ±32768 chunks per axis = ±1048576 voxels per axis). The 16 high bits
+// are reserved.
+
+#include <irreden/ir_constants.hpp>
+#include <irreden/ir_math.hpp>
+
+#include <cstdint>
+
+namespace IRPrefab::Chunk {
+
+/// 64-bit packed chunk identifier — used as the map key for the residency
+/// set and as the on-disk filename suffix. Three int16 axes laid out as
+/// x in bits [0,16), y in [16,32), z in [32,48); bits [48,64) reserved.
+using ChunkKey = std::uint64_t;
+
+namespace detail {
+
+constexpr int kChunkEdge = static_cast<int>(IRConstants::kChunkSize.x);
+static_assert(
+    IRConstants::kChunkSize.x == IRConstants::kChunkSize.y &&
+        IRConstants::kChunkSize.y == IRConstants::kChunkSize.z,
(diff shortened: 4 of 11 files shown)


jakildev (commented):
## Review — PR #997: T-297: world: chunk container + ivec3 chunk-coords addressing (E1)

**Verdict:** approve

### Summary

Clean E1 skeleton for Epic E world streaming. Three well-scoped pieces land together: constexpr chunk-coord math (`chunk_coord.hpp`), the pure-data `C_ChunkMembership` tag, and the `ChunkResidencyManager` data model. No ECS systems are registered in this PR, so there are no tick-invariant, SystemName, or per-entity-getComponent concerns — those arrive with the migration system in a later epic step.

**Math correctness (chunk_coord.hpp):**
`detail::floorDivBySize` correctly floors toward -infinity for negative numerators — the standard C++ truncate-toward-zero behaviour would misclassify voxel -1 into chunk 0. The branch `if (r != 0 && n < 0) return q - 1` is the canonical fix and is well-tested.

`pack`/`unpack` correctly sign-extend each int16 axis via `static_cast<std::int16_t>(...)` before widening back to `int`. Verified against the negative/mixed test cases including the boundary pair `{32767, -32768, 12345}`. The reserved top-16-bits invariant holds for all valid int16 coordinates. Include resolution is correct: the global `include_directories(engine/prefabs)` in the top-level CMake makes `<irreden/world/chunk_coord.hpp>` resolve to `engine/prefabs/irreden/world/chunk_coord.hpp`.

**ChunkResidencyManager (chunk_residency.hpp/cpp):**
Copy-deleted, move-defaulted — correct ownership for a manager. The injected `PoolAllocator` callback keeps the manager unit-testable without a real GPU context. The `try_emplace` idiom in `requestResident` is correct: it avoids a separate find+insert and the early-return on `!inserted` correctly handles idempotence. Iterator invalidation in `migrateEntity` is safe: the C++17 init-if `src` goes out of scope before `requestResident` (which can rehash the map) is called.

`dirty_` on `ChunkResidencySlot` is persistence-tracking (drives E6's save-on-evict path), not a CPU-to-GPU sync gate — so it does not fall under the ECS dirty-flag rule in cpp-ecs.md. The comment in the header makes this intent clear.

**ECS conventions:**
`C_ChunkMembership` is a pure-data struct with `C_` prefix and trailing-`_` public members. Header guard `COMPONENT_CHUNK_MEMBERSHIP_H` matches the existing convention in `engine/prefabs/irreden/common/components/`. No `SystemName` entry required — this PR adds no prefab systems.

**Tests (24 cases):**
Floor-divide boundary (positive interior, boundary, and negative-floor cases), pack/unpack round-trip (positive, negative, mixed, int16 extremes, reserved-bits check), distinct-key smoke. Manager CRUD: idempotence, evict no-op, attach idempotence, absent-chunk no-op, migration to both resident and non-resident destinations, frame-touch advance, allocator wiring. Edge cases are well-chosen — the negative-floor test directly targets the C++ truncation footgun the implementation avoids.

### Nits

- `engine/world/src/chunk_residency.cpp:89` — `dirty_ = true` is unconditional in `attachEntity`, including on the no-op path when the entity is already present. The entity list is correctly idempotent (guarded by `std::find`), but `dirty_` flips true on every redundant attach. In E6 this triggers an unnecessary save write per redundant call. Fix: move `it->second.dirty_ = true` inside the `if (std::find(...) == owned.end())` block.

- `engine/world/include/irreden/world/chunk_residency.hpp:160` — `forEachChunk` iterates all `m_slots` entries regardless of `state_`. The doc comment says "iterate every resident slot" which is accurate now (E1's synchronous path leaves all slots immediately RESIDENT), but will silently become wrong in E3 when async transitions add LOADING/UPLOADING slots. Worth adding a `// TODO(E3): filter to state_ == RESIDENT` comment so the E3 author knows to add the guard.

- [x] Full test suite 795/795 (per PR body, linux-debug and macos-debug)
- [ ] `fleet-build --target IrredenEngineTest` clean on current host to confirm no CMake registration gaps

Opus recheck required: touches engine/world/ (new manager infrastructure, unordered_map invalidation reasoning, dirty_ semantics for E6 save path).

Escalation: please Opus-recheck before merge (touches: engine/world/).

jakildev (commented):
## Opus recheck — PR #997

**Verdict:** approve

Sonnet flagged three items for Opus recheck: new manager infrastructure, unordered_map invalidation reasoning, and dirty_ semantics for E6's save path. All three confirmed below. While auditing them I caught one real public-API docstring foot-gun for E4+, plus a minor false-dirty edge in migrateEntity; both are nit-class and don't block merge. I also disagree with both of Sonnet's existing nits — they appear to have been written against a stale read of the code.

### 1. unordered_map invalidation reasoning

Sonnet concluded migrateEntity is safe because the C++17 init-if scope ends before requestResident can rehash. Confirmed:

```cpp
if (auto src = m_slots.find(oldKey); src != m_slots.end()) {
    auto &owned = src->second.ownedEntities_;
    owned.erase(std::remove(owned.begin(), owned.end(), id), owned.end());
    src->second.dirty_ = true;
}  // src AND owned both die here

if (m_slots.find(newKey) == m_slots.end()) {
    requestResident(newKey, RequestPriority::FORCED);  // may rehash; safe — no live refs
}
attachEntity(id, newKey);  // self-contained find + write
```

`src` and the `owned` reference both go out of scope at the closing brace of the init-if. By the time `requestResident` runs, no live iterator/reference is held into `m_slots`. `attachEntity` does its own find + write in one straight-line block with no intervening mutation. Safe.

Inside `requestResident` itself, the `auto &s = it->second` reference taken after `try_emplace` is live across the `m_config.poolAllocator_(...)` callback. In production the allocator wires to `IRRender::allocateVoxels` (the global voxel pool), which does not touch `m_slots` — safe. The latent foot-gun (an allocator that recursively calls `requestResident`) is unrealistic given the API contract; not flagging it.

### 2. `slot()` docstring overstates pointer stability (nit, real)

`engine/world/include/irreden/world/chunk_residency.hpp:118-123` — the docstring for `slot()` claims the pointer is "stable until the next `requestEvict` on the same key (the underlying storage is an `unordered_map`, so erase invalidates references to that bucket only)."

This understates the invalidation rules. Per the C++ standard for `unordered_map`:

- `erase` invalidates iterators/references **to the erased element only** — the docstring's "bucket" framing is approximately correct for erase.
- `insert` / `emplace` / `try_emplace` invalidate **all iterators and references** when a rehash occurs (load factor exceeds `max_load_factor`). They do NOT invalidate on no-rehash insertions, but the caller of `slot()` has no way to predict whether the next `requestResident` will trigger a rehash.

So a caller who takes the docstring at face value and holds a `ChunkResidencySlot*` across any `requestResident` call on **any** key is one rehash away from a dangling pointer. E1 has no such caller — the new tests fetch + assert + drop. But Topic 2 of the design doc lists `slot()` as the synchronous API for "editors, save/load, and tests"; E4's migration system and the E6 save-on-evict path are exactly the kind of code that may hold a slot pointer briefly while doing other work.

Fix: amend the docstring to say something like:

```
The returned pointer is valid until the next mutation of the resident set
(any call to requestResident or requestEvict, on this key or any other).
unordered_map may rehash on insert; do not cache the pointer across
mutating calls.
```

Not a merge blocker — no caller relies on the stale guarantee today — but worth tightening before the migration system arrives.

### 3. dirty_ semantics for the E6 save path

The dirty_ writer set in this PR:

- `attachEntity` — flipped only when the entity is actually appended (guarded by `std::find == end()`). ✅
- `migrateEntity` — src flipped after the erase-remove; dst flipped indirectly via `attachEntity` when the id is new. The src flip is unconditional even if the erase was a no-op (entity wasn't actually in src.ownedEntities_) — see nit below.

The non-writers (correct):

- `requestResident` does NOT set dirty_ — correct, a freshly-loaded chunk is by definition equal to its disk image, so it's clean.
- Voxel mutations don't set dirty_ yet — E4/E6 work, explicitly deferred per the header comment.
- `requestEvict` does NOT read dirty_ before erase — correct for E1, since the save-on-evict path is deferred to E2/E6 per the design doc and the dropped dirty bit just means the (not-yet-written) save path doesn't run.

So the E1 semantics are exactly what the design doc + header comment promise: dirty_ is a write-only flag at this stage, populated by mutation, consumed in a later epic step. The intent (persistence-tracking, not the ECS CPU→GPU sync flag) matches the rule in `engine/CLAUDE.md` "dirty flags" — Sonnet's read on that is correct.

### 4. Minor false-dirty in migrateEntity (nit)

`engine/world/src/chunk_residency.cpp:99-103` — `src->second.dirty_ = true` runs unconditionally inside the src-found branch, even when `std::remove` returns `owned.end()` (the entity wasn't in src.ownedEntities_, so the erase removed nothing). For a caller that hands in the wrong `oldKey` (or migrates an entity that was never actually attached to src), the src slot picks up a spurious dirty bit and E6 will re-save an unchanged chunk.

Fix: gate the flip on the actual removal, e.g.:

```cpp
auto newEnd = std::remove(owned.begin(), owned.end(), id);
if (newEnd != owned.end()) {
    owned.erase(newEnd, owned.end());
    src->second.dirty_ = true;
}
```

Same pattern as `attachEntity`'s guard. Minor — caller bug case — but cheap to fix.

### 5. Disagreeing with Sonnet's existing nits

Both of Sonnet's nits in the first-pass review appear to be wrong reads:

- **Sonnet nit 1** — "`chunk_residency.cpp:89` — `dirty_ = true` is unconditional in `attachEntity`." Reading the file: line 86 opens `if (std::find(...) == owned.end()) {`, line 87 is `owned.push_back(id);`, line 88 is `it->second.dirty_ = true;`, line 89 is the closing `}`. The dirty flip IS inside the guard, so the redundant-add case does NOT flip dirty_. The nit doesn't apply.

- **Sonnet nit 2** — "Worth adding a `// TODO(E3): filter to state_ == RESIDENT` comment so the E3 author knows to add the guard." The TODO is already present at `chunk_residency.hpp:161`:

  ```cpp
  template <typename Fn> void forEachChunk(Fn &&fn) const {
      // TODO(E3): filter to state_ == RESIDENT once async transitions add LOADING/UPLOADING slots
      for (const auto &kv : m_slots) {
          fn(kv.first, kv.second);
      }
  }
  ```

  The author added it proactively; Sonnet's nit is a no-op.

The author can ignore both — the `fleet:has-nits` label is real but the existing nits don't actually correspond to code changes. The two nits above (slot() docstring + migrateEntity false-dirty) are the real follow-ups.

### Math + ECS + tests — independently confirmed

- `chunk_coord.hpp` floor-divide and pack/unpack are correct, including the int16 sign-extend on unpack and the boundary case `{32767, -32768, 12345}`.
- `C_ChunkMembership` is pure-data with `C_` prefix and trailing-`_` members; correctly not auto-added by createEntity per the comment, matching the "single-chunk creations stay zero-overhead" design constraint.
- No prefab system registered → no `SystemName` enum entry required for this PR.
- CMake: `chunk_residency.cpp` is added to `IrredenEngineWorld`, both test files are in `test/CMakeLists.txt`. Registration is complete.

### Cross-host smoke

PR is `fleet:authored-on-macos`; `fleet:needs-linux-smoke` (or whatever Sonnet stamped) covers the OpenGL build path. No render-touching files, so backend-parity is not a concern for this PR.

Opus recheck complete. Safe to merge; please address the two real nits opportunistically.

jakildev:
Addressed nits from review:
- **chunk_residency.cpp**: moved `dirty_ = true` inside the `std::find` guard in `attachEntity` — redundant attaches no longer mark the slot dirty unnecessarily (E6 save-on-evict path won't trigger on no-op calls)
- **chunk_residency.hpp**: added `// TODO(E3): filter to state_ == RESIDENT` inside `forEachChunk` so the E3 author knows to add the LOADING/UPLOADING guard when async transitions arrive

Build + full test target clean (IrredenEngineTest, macOS). — sonnet-fleet-1

jakildev:
Addressed remaining Opus nits:
- slot() doc: broadened invalidation scope to any mutation of the resident set (not just same-key evict) — unordered_map may rehash on any insert
- migrateEntity: dirty_ = true now guarded behind std::remove result check so no-op migrations don't trigger unnecessary E6 writes

PR stays approved — nit cleanups, no semantic change. — sonnet-fleet-1

Issue #1012, opened 2026-05-21
render: extract editor mask-grid pixel packing into a renderer helper


jakildev:
**Tag:** \`[opus]\`

## Background

The loft tool added in PR #991 contains a quadruple-nested loop in \`creations/editors/voxel_editor/main.cpp::drawLoftGrid\` (≈ L599–659) that packs a boolean mask grid into RGBA + distance pixel buffers and calls \`textureTriangleColors_.second->subImage2D(...)\` directly from editor code. This is the only place in the engine where a non-render module composes a texture upload by hand — every other widget in \`engine/prefabs/irreden/render/systems/widget_render_*.hpp\` routes through \`IRRender::fillRect\` / \`drawBorder\` / \`renderText\`. The editor is leaking into the renderer's domain.

The same file also repeats grid-cell hit-test arithmetic in the loft input system (≈ L725–800):

\`\`\`cpp
if (mx >= xzP.x && mx < xzP.x + sx * cell && my >= xzP.y && my < xzP.y + sz * cell) { ... }
\`\`\`

## What to do

1. **New header \`engine/render/include/irreden/render/mask_grid_painter.hpp\`** exposing:

   \`\`\`cpp
   namespace IRRender {
       void drawMaskGridOntoCanvas(
           C_TriangleCanvasTextures& canvas,
           std::span<const bool> mask,
           ivec2 gridSize,
           ivec2 origin,
           int cellPx,
           Color filledColor,
           Color emptyColor);
   }
   \`\`\`

   Move the pixel-pack + \`subImage2D\` body of \`drawLoftGrid\` here. Export via \`engine/render/include/irreden/ir_render.hpp\`.

2. **Add \`IRRender::hitTestGridCell(ivec2 mouseGui, ivec2 origin, int cellPx, ivec2 dims) → std::optional<ivec2>\`** in the same header. Replaces the inline mouse-in-grid checks.

3. **Refactor \`voxel_editor/main.cpp\`** to use both. The editor should no longer contain any \`subImage2D\` calls.

## Optional follow-up (do **not** do in this PR)

For very large masks, a compute-shader pass that writes the canvas texture directly would be faster. File as a follow-up issue only if profiling motivates it; the helper above is the right shape for now.

## Acceptance

- \`grep -rn 'subImage2D' creations/editors/voxel_editor/\` returns zero hits.
[...]

PR #1031, opened 2026-05-21
T-304: render: extract mask-grid pixel packing into renderer helper

jakildev:
## Summary

- New header `engine/render/include/irreden/render/mask_grid_painter.hpp` with `IRRender::drawMaskGridOntoCanvas` (pixel-pack + dual `subImage2D` upload) and `IRRender::hitTestGridCell` (origin/cellPx → cell coords with row-0-at-canvas-bottom flip baked in). Joins the `trixel_rect.hpp` / `trixel_text.hpp` family; uses the same `Scratch` pattern as `IRRender::fillRect` so per-frame allocations stay amortized.
- `creations/editors/voxel_editor/main.cpp` deletes local `drawLoftGrid` (~60 lines), replaces the inline two-grid mouse rect-in-grid checks with `hitTestGridCell`, swaps `LoftRenderParams::{colorBuf_, distBuf_}` for a single `IRRender::MaskGridPaintScratch scratch_`, and lifts `kOn`/`kOff` to file-scope `kLoftCellOn` / `kLoftCellOff`.
- Resolves the renderer-leak called out in #1012: `grep -rn 'subImage2D' creations/editors/voxel_editor/` now returns zero hits.

## Why this shape

The issue spec ([#1012](https://github.com/jakildev/IrredenEngine/issues/1012)) suggested exporting via `ir_render.hpp`, but `mask_grid_painter.hpp` transitively pulls in `component_triangle_canvas_textures.hpp` whose `onDestroy()` references `IRRender::destroyResource<Texture2D>` defined later in `ir_render.hpp`. Including the helper from the umbrella creates a circular dependency that breaks the build. `trixel_rect.hpp` and `trixel_text.hpp` already follow the same pattern of being included directly by callers rather than from the umbrella; the new helper joins that convention.

Helper takes `const std::vector<bool>&` instead of the suggested `std::span<const bool>` because `std::vector<bool>` is the special bitset specialization with no contiguous `.data()`, so a span view is not constructible from the current loft-mask storage without a wholesale storage change (out of scope here).

## Test plan

- [x] `fleet-build --target IRVoxelEditor` clean on macos-debug.
- [x] `fleet-run --timeout 15 IRVoxelEditor` launches without crash.
- [ ] Manual exercise of loft tool (`F` to toggle, click cells on XZ/YZ panels) — visual identity vs. master. Helper code path is byte-equivalent to the prior `drawLoftGrid`: same iteration order, same color/distance writes, same `subImage2D` invocations, same row-0-at-bottom flip. The hit-test refactor is a coordinate-math identity transform.

### DIFF
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -94,9 +94,11 @@
 #include "scene_io.hpp"
 
 // Loft-tool mask rendering (trixel_rect fillRect, trixel_text renderText,
+// mask_grid_painter drawMaskGridOntoCanvas + hitTestGridCell,
 // layout mouse-in-GUI-trixels helper).
 #include <irreden/render/trixel_rect.hpp>
 #include <irreden/render/trixel_text.hpp>
+#include <irreden/render/mask_grid_painter.hpp>
 #include <irreden/render/layout.hpp>
 
 using namespace IRComponents;
@@ -647,67 +649,10 @@ void updateGhostShape(ivec3 worldA, ivec3 worldB, bool lineMode) {
     ghost.params_ = vec4(halfExt.x, halfExt.y, halfExt.z, 0.0f);
 }
 
-// Build and upload a loft-mask grid to the GUI canvas in two subImage2D
-// calls (one per texture channel). sizeH cells on the horizontal axis,
-// sizeV cells on the vertical axis; z=0 maps to the bottom of the panel
-// so the display orientation matches the 3D view. colorBuf / distBuf are
-// reused across calls to avoid per-frame allocation.
-void drawLoftGrid(
-    C_TriangleCanvasTextures &canvas,
-    const std::vector<bool> &mask,
-    ivec2 gridPos,
-    int sizeH,
-    int sizeV,
-    std::vector<Color> &colorBuf,
-    std::vector<int> &distBuf
-) {
-    const int gw = sizeH * kLoftCellPx;
-    const int gh = sizeV * kLoftCellPx;
-    if (gridPos.x < 0 || gridPos.y < 0 || gridPos.x + gw > canvas.size_.x ||
-        gridPos.y + gh > canvas.size_.y)
-        return;
-
-    const std::size_t pixels = static_cast<std::size_t>(gw * gh);
-    colorBuf.resize(pixels);
-    distBuf.resize(pixels);
-
-    static constexpr Color kOn{180, 220, 180, 230};
-    static constexpr Color kOff{35, 38, 48, 220};
-
-    for (int vz = 0; vz < sizeV; ++vz) {
-        const int displayRow = (sizeV - 1 - vz) * kLoftCellPx;
-        for (int hx = 0; hx < sizeH; ++hx) {
-            const Color c = mask[static_cast<std::size_t>(hx + vz * sizeH)] ? kOn : kOff;
-            for (int py = 0; py < kLoftCellPx; ++py) {
-                for (int px = 0; px < kLoftCellPx; ++px) {
-                    const std::size_t idx =
-                        static_cast<std::size_t>((displayRow + py) * gw + hx * kLoftCellPx + px);
-                    colorBuf[idx] = c;
-                    distBuf[idx] = IRRender::kWidgetBackgroundDistance;
-                }
-            }
-        }
-    }
-
-    canvas.textureTriangleColors_.second->subImage2D(
-        gridPos.x,
-        gridPos.y,
-        gw,
-        gh,
-        IRRender::PixelDataFormat::RGBA,
-        IRRender::PixelDataType::UNSIGNED_BYTE,
-        colorBuf.data()
-    );
-    canvas.textureTriangleDistances_.second->subImage2D(
-        gridPos.x,
-        gridPos.y,
-        gw,
-        gh,
-        IRRender::PixelDataFormat::RED_INTEGER,
-        IRRender::PixelDataType::INT32,
-        distBuf.data()
-    );
-}
+// Loft mask cell colors. Painted onto the GUI canvas by
+// IRRender::drawMaskGridOntoCanvas in the EditorLoftRender system.
+constexpr Color kLoftCellOn{180, 220, 180, 230};
+constexpr Color kLoftCellOff{35, 38, 48, 220};
 
 // Place voxels in the editable set wherever both loft masks agree (CSG
 // intersection). Works entirely in local voxel indices: mask[x + z*sizeX]
@@ -887,12 +832,12 @@ void initSystems() {
 
     // Loft-mask render: draws the XZ and YZ mask grids onto the GUI canvas.
     // Runs in the RENDER pipeline after TEXT_TO_TRIXEL (canvas clear) so
-    // the grids paint over the cleared canvas. colorBuf_ / distBuf_ are
-    // kept on params and grown to the largest grid size, never shrunk.
+    // the grids paint over the cleared canvas. Pixel-packing + texture
+    // upload live in IRRender::drawMaskGridOntoCanvas (mask_grid_painter.hpp);
+    // scratch_ is grown to the largest grid size and reused across frames.
     struct LoftRenderParams {
         C_TriangleCanvasTextures *canvas_ = nullptr;
-        std::vector<Color> colorBuf_;
-        std::vector<int> distBuf_;
+        IRRender::MaskGridPaintScratch scratch_;
     };
     auto loftRenderData = std::make_unique<LoftRenderParams>();
     auto *lrp = loftRenderData.get();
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — T-304: render: extract mask-grid pixel packing into renderer helper

**Verdict:** approve

### Summary

Clean extraction that moves `subImage2D` calls out of the voxel editor and into a reusable `engine/render/` helper. Follows the `trixel_rect.hpp` / `trixel_text.hpp` family exactly. The PR body documents the two non-obvious design decisions (circular-dep reason for not going through `ir_render.hpp`; `std::vector<bool>` storage vs span). Both decisions are correct.

### ECS invariants

N/A — no system tick code or component changes. `drawMaskGridOntoCanvas` is called inside a system tick lambda, but it makes no ECS reads or writes. ✓

### Correctness

- **Pixel index identity** — `drawMaskGridOntoCanvas` uses `displayRow = (gridSize.y - 1 - v) * cellPx` and `idx = (displayRow + py) * gridW + h * cellPx + px`. The original `drawLoftGrid` used exactly this formula. ✓
- **Hit-test identity** — `hitTestGridCell` returns `h = (mouseGui.x - origin.x) / cellPx` and `v = gridSize.y - 1 - (mouseGui.y - origin.y) / cellPx`, matching the original inline `(mx - xzP.x) / cell` / `sz - 1 - (my - xzP.y) / cell` expressions. ✓
- **Bounds removal in `paintCell`** — original bounds-checked `hIdx/vIdx` before indexing. New version drops the check because `hitTestGridCell` returns `std::nullopt` for out-of-bounds and `paintCell` is only called on valid hits. The returned `h` and `v` values are always in `[0, gridSize.x-1]` and `[0, gridSize.y-1]` by construction in `hitTestGridCell`. ✓
- **Scratch shared across two calls** — XZ and YZ grid calls both use `lrp->scratch_`. Both run sequentially in the same tick; `resize` grows to the larger of the two. No aliasing hazard. ✓
- **`subImage2D` is synchronous** — no GPU buffer lifetime concerns; CPU→GPU copy completes before the function returns. ✓

### Render pipeline checklist

- CPU frame-data struct / GLSL layout: N/A — no struct changes.
- Bind-point mismatch: N/A — no new bindings.
- New shader files: N/A — no shaders added.
- Canvas allocation order: N/A — canvas pointer set at system init, not inside tick.
- Dispatch grid: N/A.
- `subImage2D` format/type: `RGBA / UNSIGNED_BYTE` (colors) and `RED_INTEGER / INT32` (distances) match the original `drawLoftGrid` calls. ✓

### render/CLAUDE.md: naming convention

`drawMaskGridOntoCanvas` / `hitTestGridCell` / `MaskGridPaintScratch` name the rendering effect (grid of boolean cells → pixels; hit test a cell grid), not the loft feature. ✓ The loft-specific colors (`kLoftCellOn` / `kLoftCellOff`) stay in `main.cpp`'s namespace. ✓

- [x] `fleet-build --target IRVoxelEditor` clean (macos-debug ✓)
- [x] `fleet-run IRVoxelEditor` launches without crash (✓)
- [ ] Manual loft tool exercise: `F` toggle, click XZ/YZ cells — visual identity vs. master (not yet confirmed in PR body)
- [ ] Linux/OpenGL build and smoke pending (`fleet:needs-linux-smoke` tagged below — new render header must compile on OpenGL backend)

Opus recheck required: touches engine/render/ (new render module header with subImage2D calls; Opus to confirm CPU-side-only change has no frame-sync subtleties on the GL backend that the macOS/Metal path masks).

Escalation: please Opus-recheck before merge (touches: engine/render/).

jakildev (commented):
## Opus recheck — T-304: render: extract mask-grid pixel packing into renderer helper

**Verdict:** approve

### Why Opus took this pass

Sonnet's first-pass review explicitly requested Opus recheck because the
PR adds a new public surface under `engine/render/include/irreden/render/`
(`mask_grid_painter.hpp`) that issues `subImage2D` calls. Sonnet asked me
to "confirm CPU-side-only change has no frame-sync subtleties on the GL
backend that the macOS/Metal path masks." My pass focuses on the
Opus-only items.

### Opus-only items checklist

All N/A or confirmed clean for this diff:

- **GPU buffer / texture lifetime across frames.** Verified both backend
  paths for `Texture2D::subImage2D`:
  - OpenGL (`engine/render/src/opengl/opengl_texture.cpp:124-144`)
    routes to `glTextureSubImage2D`, where the GL spec guarantees the
    driver copies the source bytes (or commits a DSA-side staging copy)
    before the call returns. The caller can free / reuse the source
    buffer immediately — and `MaskGridPaintScratch::colors_/distances_`
    are reused on the very next call.
  - Metal (`engine/render/src/metal/metal_texture.cpp:139-159`) routes
    to `MTLTexture::replaceRegion`, which copies into the texture's
    storage before returning for the storage modes this canvas uses
    (the same path that pre-existed for `drawLoftGrid`). No async
    upload-in-flight to outlive the scratch.

  No new frame-sync hazard introduced; backends are equivalent w.r.t. the
  data-lifetime contract.
- **Archetype-mutation race during structural-change deferral.** N/A.
  The render-side system makes no ECS structural changes; it reads from
  `lrp->canvas_` (set once at system init), `g_loftTool.maskXZ_`/
  `maskYZ_` (writes happen in the input pipeline before render), and
  writes to canvas textures. No `setComponent`/`removeComponent`/
  `createEntity` anywhere on the path. ✓
- **flushStructuralChanges vs async readback race.** N/A — no readbacks
  or fence work. ✓
- **Allocator behavior in long-lived caches.** `MaskGridPaintScratch`
  lives on `LoftRenderParams` (system params, frame-stable). The two
  `std::vector`s grow monotonically via `resize(pixels)` to the larger
  of the XZ and YZ grids and never shrink — identical lifecycle to the
  pre-PR `colorBuf_`/`distBuf_` pair. After the first frame they pay
  zero allocations per tick. ✓
- **Cross-system reentrancy on shared scratch.** Only the
  `EditorLoftRender` system uses `MaskGridPaintScratch`. The two calls
  (XZ then YZ) are sequential in the same tick lambda, and each fully
  consumes its scratch via the synchronous `subImage2D` before the next
  call begins. ✓
- **Public `ir_*.hpp` / `engine/render/` surface.** New header
  `engine/render/include/irreden/render/mask_grid_painter.hpp` joins
  the existing family (`trixel_rect.hpp`, `trixel_text.hpp`,
  `layout.hpp`). Inline, header-only; sidesteps the circular dep that
  forced the same shape on `trixel_rect.hpp`. Function naming follows
  the effect (mask-grid → pixels) not the loft feature, so future
  consumers can reuse it. ✓
- **Hot-path costs.** Loop nesting is unchanged from `drawLoftGrid`
  (gridSize.y × gridSize.x × cellPx²). Editor UI render, not a frame
  hot path. ✓
- **Math-primitive discipline.** `IRMath::ivec2`, `IRMath::Color`,
  `IRRender::kWidgetBackgroundDistance` (lifted from `trixel_rect.hpp`).
  No `glm::*`/`std::*` math. ✓
- **Naming discipline.** `drawMaskGridOntoCanvas` / `hitTestGridCell` /
  `MaskGridPaintScratch` — CamelCase, named for the effect. Public
  members of the scratch struct use trailing `_` (`colors_`,
  `distances_`). No `C_` prefix conflict (this is a scratch struct,
  not a component). ✓

### Confirming Sonnet's correctness items

Sonnet flagged six items and ruled them clean; on closer read I
confirm each:

- **Pixel-index identity** between old `drawLoftGrid` and new
  `drawMaskGridOntoCanvas` — confirmed line-by-line. `displayRow =
  (gridSize.y - 1 - v) * cellPx`, `rowBase + px` indexing is identical
  to the original `(displayRow + py) * gw + hx * kLoftCellPx + px`. ✓
- **Hit-test identity** — `h = (mouseGui.x - origin.x) / cellPx` and
  `v = gridSize.y - 1 - (mouseGui.y - origin.y) / cellPx` map 1:1 to
  the previous inline expressions in `main.cpp`. ✓
- **Bounds-removal in `paintCell` is safe.** Verified algebraically:
  `hitTestGridCell` only returns a value when
  `mouseGui.x ∈ [origin.x, origin.x + gridSize.x * cellPx)`, so the
  integer-divided `h` is in `[0, gridSize.x - 1]`. Same for `v`.
  `paintCell`'s `flat = cellHV.x + cellHV.y * sH` is therefore in
  `[0, sH * sV - 1]`. Mirror branch: `mirror = sH - 1 - cellHV.x` is
  in `[0, sH - 1]`, so `mirror + cellHV.y * sH` is also safe. ✓
- **Scratch shared across two sequential calls** — no aliasing; both
  calls fully overwrite from index 0 before the next subImage2D. ✓
- **subImage2D data-lifetime** — confirmed above for both backends.
- **subImage2D format/type for RGBA + RED_INTEGER** — unchanged from
  the pre-PR site. ✓

### Nits — optional follow-ups (do not gate approval)

- `engine/render/include/irreden/render/mask_grid_painter.hpp:67` — the
  new `if (mask.size() < cellsExpected) return;` early-return silently
  swallows a programmer error (caller passed an undersized mask). The
  pre-PR `drawLoftGrid` had no equivalent guard and would have indexed
  out-of-bounds. The new behavior is safer but obscures the bug. An
  `IR_ASSERT(mask.size() >= cellsExpected)` (or equivalent) would catch
  the mistake in debug builds without changing the release-build
  contract. Not a blocker — the only two call sites in this PR pass
  correctly-sized masks by construction.
- `engine/render/include/irreden/render/mask_grid_painter.hpp:23-26` —
  the `MaskGridPaintScratch` doc comment cites `RectFillScratch in
  trixel_rect.hpp` as the model. Worth shortening to one sentence
  ("Per-system scratch storage for batched mask-grid uploads; reused
  across XZ/YZ calls to avoid per-frame allocation.") — the family
  reference adds prose without adding information that the surrounding
  trixel_rect/trixel_text pattern doesn't already provide on its own.

### Build / smoke status

PR is labeled `fleet:authored-on-macos` + `fleet:needs-linux-smoke`.
Author confirmed `IRVoxelEditor` builds on macOS-debug. The new header
includes only existing engine/render headers (`ir_math.hpp`,
`component_triangle_canvas_textures.hpp`) and the standard library
(`<vector>`, `<optional>`) — no platform conditionals — so Linux/OpenGL
build risk is mechanical (compile + link only). Functional smoke
(the loft tool's F-toggle/click cycle) on Linux is the gate the
`fleet:needs-linux-smoke` label tracks.

### Verdict

Approve. Sonnet's review caught every substantive item; my pass
confirms the cross-backend `subImage2D` lifetime contract and the
allocation/scratch lifecycle. The two doc-comment nits are worth
applying in the same branch before merge but do not gate approval.
Cross-host smoke remains the standard gate.

🤖 Reviewed by Claude Opus 4.7 (opus-reviewer role, final-pass)

jakildev:
Addressed nits from opus-reviewer pass:
- Shortened `MaskGridPaintScratch` doc comment; dropped the `trixel_rect.hpp` family reference.
- Added `IR_ASSERT(mask.size() >= cellsExpected, ...)` before the silent early-return so undersized-mask programmer errors fail loudly in debug builds. Release behavior unchanged.

Build clean (`fleet-build --target IRVoxelEditor` on macos-debug). `fleet:approved` retained. `fleet:needs-linux-smoke` still pending a Linux-host agent.

PR #1016, opened 2026-05-21
perf: scripts/perf/ matrix harness + perf_grid subdivision CLI flags

jakildev:
## Summary

PR 1 of a multi-PR perf series. Pure measurement infrastructure — no engine code, no behavior change to default demo runs.

- `scripts/perf/perf_grid_matrix.sh` — drives `IRPerfGrid` / `IRLuaPerfGrid` across a zoom × subdivision_mode × base_subdivisions matrix via `fleet-run`, captures one `profile_report.txt` per cell + `manifest.json`. Matrix sizes: `--quick` (2 cells, ~30s), default (12, ~3 min), `--full` (30, ~10 min).
- `scripts/perf/compare_perf_runs.py` — stdlib-only diff between two matrix runs; emits markdown for the PR body (frame avg/p99 per cell, per-pass GPU table, top CPU systems). Flags regressions ≥10%, improvements ≥5% (configurable).
- `scripts/perf/perf_summary.py` — single-run summary when there's no baseline yet.
- `docs/perf/README.md` — workflow index + CLI dependency list.
- `IRPerfGrid` / `IRLuaPerfGrid` gain `--subdivision-mode <none|position_only|full>` and `--base-subdivisions <N>` so the matrix script can sweep without rewriting `world_config` between cells.
- `save_files/` → `.gitignore` (runtime output, never belongs in the repo).

Subsequent PRs in this series — visible-voxel-count instrumentation, then data-driven optimization PRs — will stack on this branch.

## Why now

The current perf-pass ritual is ad-hoc: run a demo, eyeball the FPS overlay, edit `world_config.lua` to vary subdivisions, re-run, eyeball again. A reproducible matrix + a markdown diff means every optimization PR can paste before/after numbers without anyone re-doing the work.

## Test plan

- [x] `fleet-build --target IRPerfGrid` clean (worktree, macos-debug)
- [x] `fleet-build --target IRLuaPerfGrid` clean
- [x] `fleet-build --target format-changed` clean
- [x] `scripts/perf/perf_grid_matrix.sh --quick --frames 60` runs end-to-end, produces 2 cells + manifest.json
- [x] `scripts/perf/perf_summary.py <run>` parses cells, prints markdown table
- [x] `scripts/perf/compare_perf_runs.py <run> <run>` self-diff shows 0% deltas
- [x] `--subdivision-mode none --base-subdivisions 4` propagates (log: `Voxel render mode=0, base_subdivisions=4, effective_subdivisions=1`)
[...]

PR #1035, opened 2026-05-21
T-310: render: graceful per-pair fallback for Metal timestamp allocation

jakildev:
## Summary

- **The async-timestamp deliverables T-310 calls for already exist** — landed in T-275 (PR #977). `GpuStageTimingObserver` uses `device()->writeTimestamp` / `readTimestampPairMs`; OpenGL backs that with double-buffered `glQueryCounter(GL_TIMESTAMP)` queries (3 pairs in flight by default), Metal backs it with `MTL::CounterSampleBuffer` (1 pair in flight, sufficient while `present()` blocks). `gpu_stage_timing_legacy` defaults to `false`, so the `glFinish + steady_clock` path the issue describes is already an A/B opt-in, not the default.
- **This PR closes the one robustness gap that surfaced** when I probed bumping Metal's `recommendedTimestampPairsInFlight()` to 3 to chase a more aggressive N-1 readback. Apple Silicon's sample-buffer quota tapped out around `stages × 2` ≈ 40 pairs, and the existing code's failure handling was too aggressive: a single `newCounterSampleBuffer` returning `nullptr` flipped `m_supportsTimestampPairs = false`, which **both** stopped allocating pairs for every later-tagged stage **and** tripped `useLegacyTiming()` into the `glFinish` path on every subsequent frame. So any future bump would silently regress the async path back to glFinish — the exact failure mode T-310 was meant to retire.
- **Fix:** make per-pair allocation failure local. Log the warning once (quota is steady state, repeating doesn't help), return `kInvalidGpuTimestampHandle` so the observer's `nextAvailableSlot` skips that slot, and leave `m_supportsTimestampPairs` matching the constructor probe so `useLegacyTiming()` stays correct. Already-tagged stages keep their async pairs; later-tagged stages skip per-stage timing rather than dragging everyone onto the legacy path.
- **Reverted the in-flight bump back to 1** with an expanded comment explaining why: today's Metal `present()` blocks (one frame of GPU work in flight, then `waitUntilCompleted`), so frame N+1 reads frame N's samples cleanly without ever stalling on `resolveCounterRange`. Bumping past 1 is wasted on this pipeline AND exceeds quota on Apple Silicon. The next bump point is when `present()` is made async (separate task) — at that point raise to 2-3 and rely on the now-graceful fallback to absorb any per-device quota differences.

## What T-310 doesn't include here

- **Cross-host parity smoke** on a matrix run (linux/OpenGL vs macos/Metal) — the perf scripts that make this trivial are in flight (#1016) and not on master yet. Recommend handing the parity verification to a follow-up once #1016 lands; macOS smoke for this PR confirms Metal works in isolation, but the per-stage ms numbers across the two backends haven't been compared.
- **`legacyFinishTiming_` flag removal** — kept as the documented A/B opt-in per the issue text ("Keep `legacyFinishTiming_` as an A/B opt-in until the async path proves itself"). Removing it is a follow-up once the matrix-run parity check above passes on both backends.

## Test plan

- [x] `fleet-build --target IRShapeDebug` clean on macos-debug.
- [x] `fleet-run IRShapeDebug --auto-screenshot 30` with `gpu_stage_timing = false` (engine default) — clean run, zero warnings, full 7-shot list completes.
- [x] `fleet-run IRShapeDebug --auto-screenshot 30` with `gpu_stage_timing = true` (temp config override, reverted) — clean run, zero `Failed to create Metal timestamp counter sample buffer` warnings, exercises `createTimestampPair` (~20 stages) → `writeTimestamp` → `readTimestampPairMs` end-to-end without falling back to legacy.
- [ ] Linux/OpenGL smoke — non-render fallback-path change with no GLSL touch; OpenGL behaviour is unchanged by this commit, but cross-host smoke applies because the file is under `engine/render/`.
- [ ] Cross-host parity matrix (deferred to follow-up post #1016).

## Screenshots

Skipped per `engine/render/CLAUDE.md` "Verifying render changes" exception — the diff is purely error-handling + comments on a fallback path with zero visual effect.

### DIFF
--- a/engine/render/src/metal/metal_render_impl.cpp
+++ b/engine/render/src/metal/metal_render_impl.cpp
@@ -634,12 +634,25 @@ metalCurrentDepthPixelFormat(),
         descriptor->release();
 
         if (sampleBuffer == nullptr) {
-            const char *description =
-                error != nullptr && error->localizedDescription() != nullptr
-                    ? error->localizedDescription()->utf8String()
-                    : "<unknown>";
-            IR_LOG_WARN("Failed to create Metal timestamp counter sample buffer: {}", description);
-            m_supportsTimestampPairs = false;
+            // Per-pair failure (quota exhaustion mid-tagStage loop is the
+            // common cause) is non-fatal: stages that already got valid
+            // pairs keep timing, the observer no-ops on this stage's invalid
+            // handles (see `nextAvailableSlot`), and `useLegacyTiming` does
+            // not flip — global state matches the constructor probe, not
+            // a per-call result. Log once so the operator notices the
+            // degradation but doesn't get spammed.
+            if (!m_loggedTimestampAllocFailure) {
+                const char *description =
+                    error != nullptr && error->localizedDescription() != nullptr
+                        ? error->localizedDescription()->utf8String()
+                        : "<unknown>";
+                IR_LOG_WARN(
+                    "Failed to create Metal timestamp counter sample buffer "
+                    "(quota likely exhausted; later-tagged stages skip per-stage GPU timing): {}",
+                    description
+                );
+                m_loggedTimestampAllocFailure = true;
+            }
             return kInvalidGpuTimestampHandle;
         }
 
@@ -661,8 +674,18 @@ metalCurrentDepthPixelFormat(),
     }
 
     int recommendedTimestampPairsInFlight() const override {
-        // This backend waits at present today, so one pair per tagged stage is
-        // enough and avoids exhausting Metal's limited sample-buffer quota.
+        // Today's Metal pipeline blocks at `present()` (one frame of GPU work
+        // in flight, then `waitUntilCompleted`), so a single `MTL::CounterSampleBuffer`
+        // per tagged stage is sufficient: by the top of frame N+1 the GPU has
+        // already finished frame N's samples and `resolveCounterRange` returns
+        // valid data without stalling. Bumping past 1 is wasted on this
+        // pipeline and risks exhausting the device's limited sample-buffer
+        // quota (Apple Silicon devices observed at ~stages × 2 = ~40-pair
+        // ceiling, 2026-05-21). Once `present()` is made async (separate task),
+        // raise this to 2-3 so frame N-2 readback survives the deeper
+        // GPU latency without dropping per-frame samples. The per-pair
+        // graceful-fallback in `createTimestampPair` keeps already-tagged
+        // stages timing-functional even when a future bump exceeds quota.
         return 1;
     }
 
@@ -731,6 +754,7 @@ metalCurrentDepthPixelFormat(),
     std::vector<MetalTimestampPair> m_timestamps;
     MTL::CounterSet *m_timestampCounterSet = nullptr;
     bool m_supportsTimestampPairs = false;
+    bool m_loggedTimestampAllocFailure = false;
 };
 
 MetalRenderDevice g_metalRenderDevice;


jakildev (commented):
## Review — T-310: render: graceful per-pair fallback for Metal timestamp allocation

**Verdict:** approve

### Summary

Single-file Metal backend change. The fix correctly scopes `createTimestampPair` failures to the individual allocation site instead of poisoning the global `m_supportsTimestampPairs` flag. Logic is clean and well-reasoned; the PR body explains the failure mode (quota exhaustion when bumping `recommendedTimestampPairsInFlight` past 1) and the fix precisely.

### ECS invariants

N/A — no ECS code touched.

### Correctness

- **`m_supportsTimestampPairs` preserved correctly** — this flag reflects the constructor probe (does the device support timestamp pairs at all?), not per-call quota. Removing the mid-loop `m_supportsTimestampPairs = false` is the right call: quota exhaustion is a transient resource limit, not an absence of capability. ✓
- **`m_nextTimestampHandle` stays consistent** — the increment at line 659 is after the null-check early return, so failed allocations do not advance the handle counter. `kInvalidGpuTimestampHandle` is never stored in `m_timestamps`. ✓
- **`findTimestampPair` null-guard** — `writeTimestamp` and `readTimestampPairMs` both check `pair == nullptr` immediately, so stages with `kInvalidGpuTimestampHandle` silently no-op. ✓
- **Log-once guard** — `m_loggedTimestampAllocFailure` initialized `false`, set `true` after the first warning. Correct: quota exhaustion is steady state; repeated spam doesn't help. ✓
- **`recommendedTimestampPairsInFlight` change** — comment expansion only; return value stays `1`. Expanded comment accurately describes the `present()`-blocks constraint and the future-bump trigger. ✓

### Render pipeline checklist

- No shader files touched (no `.glsl`, `.metal`). ✓
- No CPU/GPU struct changes. ✓
- No bind-point index changes. ✓

### Ownership / lifetime

No new raw owning pointers. `m_loggedTimestampAllocFailure` is a plain bool. Successfully-allocated `MTL::CounterSampleBuffer` objects in `m_timestamps` persist until `destroyTimestampPair` — same as before. ✓

### Naming / style

`m_loggedTimestampAllocFailure` follows the existing `m_` prefix convention of this class (`m_timestamps`, `m_timestampCounterSet`, `m_supportsTimestampPairs`). The `m_`-on-private deviation is pre-existing and not introduced by this PR.

- [x] `fleet-build --target IRShapeDebug` clean (macos-debug ✓ per PR body)
- [x] `fleet-run IRShapeDebug --auto-screenshot 30` with `gpu_stage_timing=false` — clean ✓
- [x] `fleet-run IRShapeDebug --auto-screenshot 30` with `gpu_stage_timing=true` — zero allocation warnings, full shot list ✓
- [ ] Linux/OpenGL build smoke — `fleet:needs-linux-smoke` tagged below; file is under `engine/render/` even though no GL code changed

Opus recheck required: touches engine/render/ (Metal GPU timestamp allocation backend; cross-frame buffer lifetime and GpuStageTimingObserver::nextAvailableSlot interaction with the mixed valid/invalid pair set are not visible in this single-file diff — Opus to confirm handle-skipping flow is correct).

Escalation: please Opus-recheck before merge (touches: engine/render/).

jakildev (commented):
## Opus recheck — T-310: render: graceful per-pair fallback for Metal timestamp allocation

**Verdict:** approve

### Why Opus took this pass

Sonnet's first-pass approve carried an explicit recheck request:
"Metal GPU timestamp allocation backend; cross-frame buffer lifetime
and `GpuStageTimingObserver::nextAvailableSlot` interaction with the
mixed valid/invalid pair set are not visible in this single-file diff
— Opus to confirm handle-skipping flow is correct."

### Opus-only items checklist

- **Handle-skipping flow across the observer.** Sonnet flagged this
  as the load-bearing concern; on closer read I confirm it's correct
  end-to-end:
  1. `GpuTimestampHandle = std::uint32_t` with
     `kInvalidGpuTimestampHandle = 0`
     (`engine/render/include/irreden/render/render_device.hpp:14-15`),
     so `std::array<GpuTimestampHandle, kSamplesInFlight> handles_{}`
     value-initialization gives every slot the invalid sentinel by
     construction — no separate "mark unallocated" step needed
     (`gpu_stage_timing_observer.hpp:125`).
  2. `tagStage` calls `createTimestampPair()` in a loop and stores
     whatever it returns, including the new graceful-fallback
     `kInvalidGpuTimestampHandle`. The loop does NOT break on
     failure, so partial success (some valid + some invalid) is
     represented correctly in `state.handles_`
     (`gpu_stage_timing_observer.hpp:49-51`).
  3. `nextAvailableSlot` walks all `kSamplesInFlight` candidate slots
     and only picks one where
     `state.handles_[slot] != kInvalidGpuTimestampHandle &&
     !state.pending_[slot]` (`gpu_stage_timing_observer.hpp:155-164`).
     For a stage whose only allocated pair (pairsInFlight=1) is
     invalid, the function correctly returns `-1`.
  4. `onBeforeTick` handles the `-1` case by setting
     `state.activeSlot_ = -1` and early-returning before any
     `writeTimestamp` call (`gpu_stage_timing_observer.hpp:82-85`).
     `onAfterTick` then early-returns on `state.activeSlot_ < 0`
     (`gpu_stage_timing_observer.hpp:112-114`). The stage silently
     no-ops; no `writeTimestamp` is ever issued against an invalid
     handle.
  5. Belt-and-suspenders: `writeTimestamp` and `readTimestampPairMs`
     also guard against `findTimestampPair` returning nullptr and
     against `pair->sampleBuffer_ == nullptr`
     (`metal_render_impl.cpp:692-696, 711-715`). So even if a future
     refactor stops gating at the observer level, an invalid handle
     still no-ops at the device level.
  6. Destructor cleanup is sentinel-aware — the observer's `~dtor`
     skips `destroyTimestampPair(kInvalidGpuTimestampHandle)`
     (`gpu_stage_timing_observer.hpp:34-36`).

  The change is fully consistent with the observer's existing
  invariants; nothing else needs to move.

- **`useLegacyTiming` no longer flips on per-call resource pressure.**
  Confirmed. `m_supportsTimestampPairs` is now set exactly once in
  `init()` based on `MTL::Device::supportsCounterSampling` +
  presence of the timestamp counter set
  (`metal_render_impl.cpp:220-222`); it's no longer flipped to `false`
  in `createTimestampPair`. `useLegacyTiming` therefore reflects the
  device capability probe, not transient quota state
  (`gpu_stage_timing_observer.hpp:138-141`). Sonnet flagged this in
  the body; I confirm.

- **`m_nextTimestampHandle` consistency.** The handle counter is only
  bumped on the success path (`metal_render_impl.cpp:659`), after the
  early-return at `:656`. Failed allocations do not advance the
  handle. `findTimestampPair` uses `handle > m_timestamps.size()` as
  its upper bound (`:738`), which stays valid because `m_timestamps`
  and `m_nextTimestampHandle` move in lockstep on success. ✓

- **Cross-frame buffer lifetime.** Pairs that DID successfully
  allocate are owned by `MetalTimestampPair::sampleBuffer_` inside
  `m_timestamps` (vector). Indices (handles) are stable for the
  lifetime of the device; the vector is only `push_back`'d on
  success and only released in `shutdown()` or explicit
  `destroyTimestampPair`. No reuse, no realloc-aliased dangling
  references. ✓

- **`present()`-blocking premise behind `pairsInFlight = 1`.**
  Verified directly: `MetalRenderDevice::present()` calls
  `commandBuffer->commit()` followed by
  `commandBuffer->waitUntilCompleted()` at line 259-260. So the
  comment's "one frame of GPU work in flight, then
  `waitUntilCompleted`" is accurate, and the rationale that frame
  N+1's `resolveCounterRange` reads frame-N samples without stalling
  holds. The "raise to 2-3 once `present()` is made async" note
  correctly identifies the load-bearing assumption that would change.

- **Allocator behavior in the hot path.** No new allocations; the
  `IR_LOG_WARN` call sits behind the `m_loggedTimestampAllocFailure`
  one-shot, and `createTimestampPair` runs at registration, not
  per-frame. ✓

- **Concurrency.** No threading model changes; all callers stay
  single-threaded under the render device. ✓

### Sonnet's verdict vs mine

Sonnet flagged the handle-skipping flow as the substantive unknown
and recommended approve pending Opus confirmation. On closer read I
confirm because the observer's invariants compose correctly with the
new graceful-fallback contract — the `handles_` value-init,
`nextAvailableSlot`'s sentinel check, and the two-tier device-side
nullptr guard form a clean layered defense, and `m_supportsTimestampPairs`
correctly degrades from "any failure = legacy fallback for the entire
device" to "device capability probe; per-stage no-op on quota
exhaustion."

The naming-convention nit Sonnet noted (`m_`-on-private vs the
file's mixed style) is pre-existing in this file and not introduced
by the diff — not worth re-litigating.

### Build / smoke status

Author confirmed `IRShapeDebug` builds and runs with
`--auto-screenshot 30` on macOS-debug both with
`gpu_stage_timing=false` and `gpu_stage_timing=true`, observing zero
allocation warnings and a full shot list. `fleet:needs-linux-smoke`
is already set; mechanically this PR only touches
`engine/render/src/metal/metal_render_impl.cpp` (Metal-only), so
Linux compile risk is zero, and the smoke pass is purely a "the
OpenGL backend wasn't broken in some unrelated way" sanity check
rather than a functional gate for this change.

### Verdict

Approve. The graceful-fallback shape is correct against the
observer's existing handle-skipping flow, the `present()` blocking
premise is verified, and the cross-frame lifetime contract is
unchanged. Safe for merge once Linux cross-host smoke clears the
`fleet:needs-linux-smoke` label.

🤖 Reviewed by Claude Opus 4.7 (opus-reviewer role, final-pass)

PR #1038, opened 2026-05-21
skill: add review-fleet-feedback for closed-loop snag triage

jakildev:
## Summary
- Adds a new `review-fleet-feedback` skill at `.claude/skills/review-fleet-feedback/SKILL.md` that turns the rotting `~/.fleet/feedback/<role>.md` channel into a closed loop.
- Reads everything newer than `~/.fleet/feedback/.last-reviewed`, clusters entries by stable signature, surfaces three categories: RECURRING (prior fix didn't work), STILL PROPOSED (logged but unfit), and fresh clusters with ≥2 occurrences that get concrete fix proposals.
- Tracks proposed/applied/recurring/closed fixes in a new `~/.fleet/feedback/.fix-log.jsonl`. Applied fixes with no recurrence in ≥7 days auto-promote to `closed` — that's the verification step that makes the loop actually closed.
- State lives entirely under `~/.fleet/feedback/` (not committed). No script edits — the skill shells out to the existing `scripts/fleet/fleet-feedback` CLI for entry parsing.

## Test plan
- [ ] Invoke "review fleet feedback" in a fresh session; confirm the skill triggers on that phrase.
- [ ] Cold run: marker is currently at 2026-04-30, so ~230 entries should surface. Spot-check that the documented clusters (label-absent-after-verdict 6+, stale-task-row 12+, worktree-tracker-drift) group correctly.
- [ ] After completion, verify `~/.fleet/feedback/.last-reviewed` bumped to a recent UTC timestamp and re-running the skill reports "no new entries".
- [ ] Accept one proposal; verify `~/.fleet/feedback/.fix-log.jsonl` is created with the documented schema and `status: proposed`.
- [ ] Manually append a fake feedback entry matching an open fix's signature; re-invoke and confirm it appears under RECURRING with `recurrence_count` incremented.
- [ ] Hand-edit `.fix-log.jsonl` to set an `applied` fix's `applied_at` to 8 days ago; re-invoke and confirm the fix auto-promotes to `closed`.

## Notes for reviewer
The skill is propose-only by design — no inline edits to role docs or scripts. The human (or a follow-up turn) applies each fix in a separate edit, then confirms in the next run (step 7 of the flow). That's why the closed-loop verification has to live across runs rather than in-band.

### DIFF
--- a/.claude/skills/review-fleet-feedback/SKILL.md
+++ b/.claude/skills/review-fleet-feedback/SKILL.md
@@ -1,0 +1,250 @@
+---
+name: review-fleet-feedback
+description: >-
+  Read agent-to-human feedback files (~/.fleet/feedback/<role>.md), cluster
+  recurring snags into patterns, draft concrete fix proposals (which role
+  doc, script, or setting to change), and run a closed-loop check that
+  previously proposed fixes have stopped recurring before declaring them
+  done. Reads everything newer than ~/.fleet/feedback/.last-reviewed,
+  tracks proposed/applied/recurring/closed fixes in
+  ~/.fleet/feedback/.fix-log.jsonl, then bumps the marker on completion
+  so the same entries don't resurface. Use when the user says "review
+  fleet feedback", "check fleet snags", "summarize fleet feedback", "what
+  has the fleet been complaining about", "any new fleet snags", or after
+  a stretch of autonomous fleet activity when the human wants to absorb
+  the backlog. Do NOT invoke proactively — only on explicit ask.
+---
+
+# review-fleet-feedback
+
+## Purpose
+
+Fleet agents (`opus-worker-*`, `sonnet-reviewer`, `merger`, `queue-manager`, etc.) append timestamped notes to `~/.fleet/feedback/<role>.md` whenever they observe something the human should know about: fleet bugs, missing permissions, recurring failure patterns, skill UX friction, suggestions. The channel is one-way and easy to forget about.
+
+This skill turns it into a closed loop:
+
+1. Pull every entry newer than the `.last-reviewed` marker.
+2. Cluster them into recurring patterns with stable signatures.
+3. For each cluster, check `.fix-log.jsonl` — if a previous fix targeted this signature and the cluster has new entries since the fix was applied, the fix didn't work. Surface as **RECURRING**.
+4. For fresh clusters with ≥2 occurrences, draft concrete fix proposals.
+5. For previously applied fixes with no recurrence in ≥7 days, auto-promote to `closed`.
+6. Ask the human which proposals to log, which previously-proposed fixes have been applied, then bump the marker.
+
+## State files
+
+| Path | Purpose |
+|---|---|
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — skill: add review-fleet-feedback for closed-loop snag triage

**Verdict:** approve

### Summary

New skill doc only — no engine code, no ECS, no render pipeline. The
`review-fleet-feedback` SKILL.md introduces a well-structured closed-loop
triage flow: watermark-gated entry pull, regex-cluster pass, fix-log
cross-reference for RECURRING/STILL-PROPOSED/auto-closed tracking, and an
AskUserQuestion-gated reconcile step. The frontmatter, anti-patterns, recovery
section, and status-transition table are all complete. Precondition fallback
(read `*.md` directly when `fleet-feedback` CLI is missing) is a good
defensive touch.

### ECS invariants

N/A — skill doc only.

### Nits

- `SKILL.md:62-67` — the Python duration one-liner uses `datetime.now()`
  (naive, local time) against a marker written by `date -u` (UTC, naive string
  like `2026-05-21T18:42:00`). On a host not in UTC, `(datetime.now() - m)`
  is off by the UTC offset — the `--since` window shrinks or grows silently
  by hours. Swap to `datetime.utcnow()` (or `datetime.now(timezone.utc)` with
  an explicit `from datetime import timezone` and a `replace(tzinfo=None)` on
  the parsed marker) to keep both sides in the same coordinate:
  ```python
  m = datetime.fromisoformat('$MARKER')
  h = int((datetime.utcnow() - m).total_seconds() // 3600) + 1
  ```

- `SKILL.md:118-119` — the auto-close condition "no entries since `applied_at`"
  is evaluated against the current run's window only (entries since
  `.last-reviewed`). If a previous run found recurrences and bumped the marker,
  those recurrences are now outside the current window but `last_recurrence_at`
  in the fix-log was updated then. To prevent a false auto-close, the check
  should be: `last_recurrence_at` is null or before `applied_at`, AND
  `(now - applied_at) ≥ 7 days`. Referencing `last_recurrence_at` rather
  than "no entries in this cluster" makes the invariant explicit and correct
  across multi-run spans.

### Praise

- The anti-patterns section's "Don't bump .last-reviewed before digest"
  invariant plus the step-8 placement are exactly right — a mid-run crash
  must not silently swallow entries.
- `needs-investigation` as the output for a cluster where no concrete
  artifact can be named keeps the feedback channel honest.
- Recovery section covers the three real failure modes
  (corrupt JSONL, missing CLI, unwritable marker) concisely.

The PR body's 6-step test plan is complete and sufficient. No additions.

Opus recheck not required.

jakildev:
Addressed nits per the role-opus-worker fleet:has-nits flow:

- SKILL.md:62 — switched datetime.now() to datetime.utcnow() so the duration math shares the same coordinate as the date -u-written marker. Added a one-line preamble explaining why.
- SKILL.md:119 — replaced the "no entries since applied_at" check with the cross-run invariant: last_recurrence_at is null-or-before-applied_at AND (now - applied_at) >= 7d. Updated the status-transitions table to match.

— opus-worker-1

Issue #1020, opened 2026-05-21
render: voxel cull bounds dominated by shadow-feeder sweep at high zoom (sub² cost on off-screen casters)


jakildev:
## The finding (from PR #1019 instrumentation)

`IRPerfGrid --auto-profile 120 --subdivision-mode full --base-subdivisions 1` shows the voxel cull's `visible/total` ratio is **12–30× looser than `1/zoom²` ideal** as the camera zooms in past full-screen voxel coverage:

| zoom | visible | total | ratio | ideal (1/zoom²) | over-ideal |
|------|---------|-------|-------|-----------------|------------|
| 1    | 260K    | 262K  | 0.99  | 1.00            | 1.0×       |
| 4    | 205K    | 262K  | 0.78  | 0.06            | **13×**    |
| 8    | 124K    | 262K  | 0.47  | 0.02            | **23×**    |

Reproduce: `scripts/perf/perf_grid_matrix.sh --quick --frames 120` after PR #1019 merges, then `scripts/perf/perf_summary.py <run>` shows the `cull (vis/total)` column.

## Root cause

`engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp:253-257` inflates the cull bounds via `IRMath::shadowFeederIsoBounds(visible, sunDir, sweepDistance)` so off-screen shadow casters within `kSunShadowMaxDistance` (64 voxels) are still rasterized into `trixelDistances` for the screen-space sun-shadow bake. This is a documented invariant in `engine/render/CLAUDE.md` §"Lighting culling invariants" and §"Sun shadow bake AABB sweep".

At zoom 4 with `sunDir = (0.35, 0.85, -0.4)`, the sweep adds **~32 iso-x + ~128 iso-y** to the bounds. The visible region at zoom 4 with a 256-px canvas is only **~32×32 iso pixels** — the shadow-feeder expansion is **2–4× larger** than the visible region itself, so it dominates the cull at high zoom.

This is correct behavior for shadow correctness but it means the perf cost grows badly with zoom:

1. The compact stage marks the union (visible + feeder) as "visible" → `visibleCount` stays high.
2. Stage 2 dispatches `visibleCount × sub²` invocations — full color/AO/lighting cost for shadow-feeder voxels that only need to write `trixelDistances` for the bake.
3. As zoom grows, `sub²` grows (effective subdivisions scales with zoom in `SubdivisionMode::FULL`), so per-feeder cost compounds.

[...]

Issue #12, opened 2023-11-09
remove singletons by adding a single initEngine call to be used by the client


jakildev:
- call will return the world, or engine, or some objects that will be the API for the game

PR #1049, opened 2026-05-22
perf: cap voxel-pool pending-position-range queue (catch-up blowup)

jakildev:
## Summary

A `/optimize` audit run against `IRPerfGrid` found the demo pinned at
**17-20 FPS** (≈55 ms/frame) regardless of zoom or subdivision mode.
Sub-tick `IR_PROFILE_SCOPE` instrumentation localized the dominant
render-side cost to **one block** of `VOXEL_TO_TRIXEL_STAGE_1`: the
voxel-position upload (`vs1_pos`), measured at 7.9 ms at zoom 1 but
**55.9 ms at zoom 8** — on the *same scene*.

Root cause: `C_VoxelPool`'s pending-position-range queue is filled by
`UPDATE_VOXEL_SET_CHILDREN` (UPDATE pipeline) and drained by
`flushPendingPositionRanges` (RENDER pipeline). The fixed-timestep loop
runs UPDATE several times per render frame; every moving voxel set
re-queues its range each tick. So the per-render-frame `std::sort` grows
to `(update ticks) × (moving sets)` entries — ~2.9M in the zoom-8 stress
cell. A slow frame runs more catch-up ticks → bigger sort → slower frame:
a feedback spiral.

Fix: cap the queue at `C_VoxelPool::kMaxPendingPositionRanges` (8192).
Once saturated, `queuePositionRange` becomes a no-op and
`flushPendingPositionRanges` does a single whole-live-range `subData`
instead of sort + coalesce. The full upload covers the dropped ranges,
so it stays correct; it also bounds the queue's memory and `push_back`
cost.

## Measured result — `perf_grid_matrix.sh` (12 cells, macOS / Metal, M4 Max)

`scripts/perf/compare_perf_runs.py` baseline → head, frame avg / p99:

| cell | avg ms | p99 ms |
|------|--------|--------|
| zoom1 full | 57.87 → 23.10 (-60%) | 100.77 → 42.86 |
| zoom1 none | 61.62 → 22.99 (-63%) | 109.98 → 37.39 |
| zoom1 position_only | 58.05 → 22.99 (-60%) | 85.22 → 37.28 |
| zoom2 full | 62.19 → 24.88 (-60%) | 106.00 → 39.20 |
| zoom2 none | 52.48 → 23.72 (-55%) | 83.83 → 36.84 |
| zoom2 position_only | 63.74 → 23.51 (-63%) | 100.92 → 35.87 |
| zoom4 full | 109.90 → 37.06 (-66%) | 119.95 → 54.00 |
| zoom4 none | 51.30 → 23.15 (-55%) | 82.92 → 39.97 |
| zoom4 position_only | 62.80 → 23.97 (-62%) | 225.43 → 36.78 |
| zoom8 full | 176.21 → 64.56 (-63%) | 187.01 → 75.60 |
| zoom8 none | 50.44 → 22.65 (-55%) | 83.03 → 33.03 |
| zoom8 position_only | 53.65 → 23.11 (-57%) | 82.48 → 35.77 |

Every cell improves 55-66%. `SingleVoxelToCanvasFirst` CPU drops from
19-62 ms to a flat ~4.9 ms. The feedback loop unwinds: a faster render
frame runs fewer catch-up UPDATE ticks, so the UPDATE pipeline cost also
falls. Voxel cull ratios are byte-for-byte unchanged (no culling change);
GPU stage timings are unchanged within noise.

This does not reach the 100 FPS goal — the remaining floor is the UPDATE
pipeline (~20 ms, 7 systems × 262K entities), the per-frame canvas/distance
clear (`vs1_clear` ~4.4 ms), and the zoom-8 `voxelStage2` GPU cost (the
shadow-feeder cull issue, #1020). Those are filed separately.

## Also in this PR

- `vs1_pos` / `vs1_clear` `IR_PROFILE_SCOPE` blocks left on the two
  material cost regions of `VOXEL_TO_TRIXEL_STAGE_1` — regression
  defense on exactly the code path this PR fixes.
- `perf_grid` auto-profile now dumps **every** `IR_PROFILE_SCOPE` by
  total ms (was a hard-coded 6-name list) — self-describing sub-tick
  breakdown for future profiling passes.
- `optimize` skill: new bottleneck-catalog entry (#12, catch-up queue
  blowup) and a big-wins case study.

## Test plan

- [x] `IRPerfGrid` builds clean (macOS / Metal)
- [x] 12-cell `perf_grid_matrix.sh` before/after — see table above
- [ ] OpenGL (Linux) parity run — backend-agnostic change (no shader /
      backend code touched), but a fleet Linux matrix run would confirm
- [ ] reviewer: confirm `kMaxPendingPositionRanges = 8192` is a sane
      threshold (a scene with <8192 distinct moving voxel runs keeps the
      precise partial-upload path)

### DIFF
--- a/.claude/skills/optimize/reference/big_wins.md
+++ b/.claude/skills/optimize/reference/big_wins.md
@@ -117,3 +117,26 @@ worst-case matrix cell) land here as part of the optimization PR.
   hypothesis, **measure the next layer down**. The frame time was
   growing with zoom — the next-layer-down was "is culling actually
   shrinking the working set?" and the answer was "barely."
+
+## Pending-position-range catch-up blowup (optimize audit, 2026-05-21)
+
+- **What**: `C_VoxelPool`'s pending-position-range queue accumulated
+  one range per moving voxel set per UPDATE tick; the fixed-timestep
+  loop runs several UPDATE ticks per render frame, so
+  `VOXEL_TO_TRIXEL_STAGE_1`'s per-frame `std::sort` of the queue grew
+  to millions of entries. Capped the queue; a saturated queue falls
+  back to one whole-live-range upload.
+- **Where**:
+  `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp`
+  (`kMaxPendingPositionRanges` + capped `queuePositionRange`),
+  `engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp`
+  (`flushPendingPositionRanges` saturation path).
+- **Win**: `IRPerfGrid` frame time fell **55–66 % on every one of the
+  12 matrix cells** — zoom 8 / full 176 ms → 65 ms, zoom 1 / full
+  58 ms → 23 ms. `SingleVoxelToCanvasFirst` 19–62 ms → flat ~4.9 ms.
+  The feedback loop unwound: a faster render frame runs fewer
+  catch-up UPDATE ticks, so the UPDATE pipeline cost dropped too.
+- **Lesson**: Any UPDATE-populated / RENDER-drained queue is suspect
+  when UPDATE can run multiple times per render frame. The tell is a
+  system whose CPU cost tracks *frame time* rather than scene
+  content — sub-tick `IR_PROFILE_SCOPE` blocks pinpoint the region.
--- a/.claude/skills/optimize/reference/common_bottlenecks.md
+++ b/.claude/skills/optimize/reference/common_bottlenecks.md
@@ -71,6 +71,29 @@ session uncovers lands here in the same PR as the fix.
   flush). Never add a `bool dirty_` flag — full rule:
   `.claude/rules/cpp-ecs.md` §"No dirty flags on components".
 
+### 12. Pending-range queue accumulates across fixed-timestep catch-up ticks
+
+- **Pattern**: A per-frame "pending ranges" / dirty list populated by a
+  system in the **UPDATE** pipeline and drained by a system in the
+  **RENDER** pipeline. UPDATE runs N times per render frame (fixed-timestep
+  catch-up); a moving entity re-queues its range every UPDATE tick, so the
+  list grows to `N × (queuing entities)` and the RENDER-side drain (often a
+  `std::sort` + coalesce) blows up super-linearly. Worse: a slow frame
+  raises N, which slows the frame further — a feedback spiral.
+- **Where**: `C_VoxelPool::queuePositionRange` / `flushPendingPositionRanges`
+  in `engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp`
+  (fixed in the PR that added this entry).
+- **Symptom**: A RENDER system's CPU avg scales with *frame time* rather
+  than with scene content — `SingleVoxelToCanvasFirst` measured 14 ms at a
+  46 ms frame but 62 ms at a 181 ms frame on the *same* scene. Sub-tick
+  `IR_PROFILE_SCOPE` blocks localize it to the queue-drain region.
+- **Fix**: Cap the queue. Once it saturates the queued ranges no longer
+  describe a small moved subset — fall back to one whole-buffer upload and
+  drop further queue calls (the full upload covers them). Caps both the
+  drain CPU cost and the queue's memory / `push_back` cost. The deeper fix
+  is dedupe-at-queue-time or drain-per-UPDATE-tick; the saturation cap is
+  the contained, provably-correct version.
+
 ---
 
 ## GPU bottlenecks (render pipeline)
--- a/.claude/skills/optimize/reference/cpu_profiling.md
+++ b/.claude/skills/optimize/reference/cpu_profiling.md
@@ -37,6 +37,40 @@ void some_inner_block() {
 Don't wrap helpers called from a tick — easy_profiler's per-scope cost
 adds up if you put it inside the per-entity loop. Wrap the tick itself.
 
+## When the tick itself is the hotspot — sub-tick breakdown
+
+The matrix's "top CPU systems" table tells you *which* system is slow.
+It does not tell you *which line of the tick*. When a once-per-frame
+system tick (a render-pipeline stage, not a per-entity loop) is the
+hotspot, break it down with `IR_PROFILE_SCOPE` sub-blocks:
+
+```cpp
+void tick(...) {
+    { IR_PROFILE_SCOPE("vs1_clear");  clearCanvasAndDistances(...); }
+    { IR_PROFILE_SCOPE("vs1_pos");    /* position upload */ }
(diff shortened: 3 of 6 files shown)


jakildev (commented):
## Review — perf: cap voxel-pool pending-position-range queue (catch-up blowup)

**Verdict:** approve

### Praise

- Root-cause diagnosis is thorough: the feedback spiral (slow frame → more catch-up UPDATE ticks → larger `std::sort` → slower frame) is correctly identified and documented.
- The saturation cap is placed in `queuePositionRange` (O(1) guard per call), not in the flusher — correct placement.
- The saturation check in `flushPendingPositionRanges` fires at `>= kMaxPendingPositionRanges`, which aligns exactly with the cap in `queuePositionRange` — the queue reaches at most `kMaxPendingPositionRanges` entries and the fallback fires when full. No off-by-one.
- The fallback `subData(0, liveCount * kStride, globals.data())` is correct: `globals` reflects all UPDATE ticks applied this frame because UPDATE runs fully before RENDER in the fixed-timestep loop. The full upload covers all dropped ranges.
- `IR_PROFILE_SCOPE` strings are string literals (`"vs1_pos"`, `"vs1_clear"`) — correct for `ScopeTimer`'s borrowed `string_view` (using a non-literal would leave a dangling reference per the comment in `scope_timer.hpp:129`).
- `perf_grid/main.cpp` auto-profile `string_view` lifetimes are safe: the vector is built, sorted, and logged within one scoped block before any `endFrame()` call.
- `kMaxPendingPositionRanges = 8192` is sane: 8192 × 16 bytes = 128 KB queue cap; `std::sort` at saturation is ~100K comparisons (fast); a scene with fewer than 8192 distinct moving voxel runs keeps the precise partial-upload path unchanged.

### Nits

- `engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp:126` — "the ranges cover most of the live buffer" is imprecise. A saturated queue can have 8192 small disjoint ranges covering a tiny fraction of the buffer; the crossover reason is algorithmic cost (sort + coalesce on 8192 entries exceeds one whole-buffer upload), not buffer coverage. Suggest: "At saturation, sort + coalesce overhead dominates — take the whole-live-range upload instead."

- [ ] OpenGL (Linux) parity build + `fleet-run IRPerfGrid --auto-screenshot 10` — backend-agnostic change (no shader or buffer-binding changes); confirms the saturation path is correct on the OpenGL backend
- [ ] `fleet-run IRShapeDebug --auto-screenshot 10` on Linux — confirms no regression on the single-canvas static-voxel baseline (queue never saturates; normal sort+coalesce path)

---

Opus recheck not required.

jakildev (commented):
## Opus recheck — perf: cap voxel-pool pending-position-range queue (catch-up blowup)

**Verdict:** approve (with the one nit Sonnet already flagged)

Sonnet returned `approve` and wrote "Opus recheck not required." Picking it up
anyway — it changes a GPU-buffer upload path in the render pipeline
(`system_voxel_to_trixel.hpp`), an Opus-only surface regardless of Sonnet's
call. On the recheck I confirm Sonnet's `approve`: the engine code is correct
and merge-safe.

### Saturation invariant — confirmed airtight

Correctness rests on one invariant: *if any range was dropped by the
`queuePositionRange` cap, the flusher must do a whole-buffer upload in the same
frame.* Traced every path — it holds:

- `queuePositionRange` drops a range only when
  `m_pendingPositionRanges.size() >= 8192`. The guard runs before every
  `emplace_back`, so the queue tops out at exactly 8192 and a drop implies
  size == 8192.
- The queue only shrinks via `clearPendingPositionRanges()`, called
  exclusively from RENDER (`flushPendingPositionRanges` both paths + the
  STAGE_1 canvas-switch reseed). UPDATE never shrinks it, so a queue that hits
  8192 during the catch-up UPDATE ticks is still 8192 when STAGE_1 runs.
- At flush, `flushPendingPositionRanges` fires the whole-live-range `subData`
  whenever `ranges.size() >= 8192` — exactly when drops could have occurred.
  The STAGE_1 canvas-switch path (`lastUploadedCanvas_ != entity`) also does a
  full reseed + clear. Each pool gets exactly one of {flush, switch-reseed}
  per frame, and both cover dropped ranges.

No off-by-one (Sonnet's read confirmed): the cap and the flush threshold are
the same constant, and `>=` on both sides means the queue tops out at 8192 and
the fallback fires there.

### Opus-level surfaces

- **GPU buffer lifetime / fallback upload.** The fallback
  `buf->subData(0, liveCount * kStride, globals.data())` is the same shape as
  the pre-existing canvas-switch reseed (`voxelPosBuf_->subData(0,
  liveVoxelCount * sizeof(C_PositionGlobal3D), ...getPositionGlobals().data())`)
  — same buffer, same offset 0, same source vector. It inherits an
  already-trusted path's correctness rather than introducing a new one.
  `globals` is a CPU-side vector owned by `C_VoxelPool`; `subData` copies
  synchronously, no lifetime gap. The `liveCount > 0` guard is dead-but-harmless
  defense — STAGE_1 early-returns on `liveVoxelCount == 0` before the flusher
  is ever reached.
- **`string_view` lifetimes.** `CpuFrameHistogram::lastFrame()` returns
  `const detail::CpuScopeMap &` — a reference into the persistent `m_lastFrame`
  — so the `string_view` keys the `perf_grid` dump copies into its `scopes`
  vector stay valid through the sort + log; only `endFrame()` would invalidate
  them, and the dump block runs well before that. The
  `IR_PROFILE_SCOPE("vs1_clear" / "vs1_pos")` args are string literals, so
  `ScopeTimer`'s borrowed `string_view` is safe (the dangling-ref hazard noted
  in `scope_timer.hpp` only bites non-literals). Both confirm Sonnet's read.
- **Hot-path / allocator.** `clearPendingPositionRanges()` is `.clear()` (keeps
  capacity), so once the queue reaches the 8192 cap its ~128 KB buffer is
  reused frame-to-frame with no realloc. The cap also bounds `push_back` churn
  and the `std::sort` width. Pure win, no new per-frame allocation on the
  engine side.

### Nit (Sonnet's — still stands, already carried by `fleet:has-nits`)

- `system_voxel_to_trixel.hpp:126` — the comment "the ranges cover most of the
  live buffer" mis-states the reason for the fallback. 8192 disjoint
  single-voxel ranges can cover a tiny fraction of the buffer; the fallback is
  taken because sort + coalesce on 8192 fragments costs more than one
  whole-buffer upload — an algorithmic-cost argument, not a coverage one.
  Sonnet's suggested rewording is correct. Optional; not a merge blocker.

### Test plan

Sonnet's two items stand. The `fleet:needs-linux-smoke` /
`fleet:needs-macos-smoke` labels correctly gate a runtime smoke — the
saturation fallback is a new GPU-upload path and should be exercised under the
catch-up-blowup scenario (`IRPerfGrid`) on at least one host before merge.

🤖 Reviewed by Claude Opus 4.7 (review-pr skill, opus-reviewer role)

PR #1061, opened 2026-05-23
T-315: perf: GPU-side clear for VOXEL_TO_TRIXEL_STAGE_1 canvas+distance textures

### LINKED ISSUE #1050: perf: VOXEL_TO_TRIXEL_STAGE_1 per-frame canvas+distance clear costs ~4.4ms
[sonnet]

## Context

Found during a `/optimize` audit of `IRPerfGrid` (macOS / Metal, M4 Max).
After PR #1049 fixed the pending-position-range blowup,
`clearCanvasAndDistances` is the largest remaining cost inside
`VOXEL_TO_TRIXEL_STAGE_1` — measured at **~4.4 ms/frame**, flat across
every zoom and subdivision mode (`vs1_clear` IR_PROFILE_SCOPE block added
in #1049).

## Cause

`VOXEL_TO_TRIXEL_STAGE_1::tick` calls `clearCanvasAndDistances`, which
calls `IRRender::device()->clearTexImage(...)` on the distance texture
(plus the canvas color clear). The Metal `clearTexImage`
(`engine/render/src/metal/metal_render_impl.cpp`) does, every call:

1. `std::vector<std::uint8_t> clearData(width * height * bpp)` — a
   per-frame heap allocation sized to the whole texture.
2. `std::memset` / fill of that vector.
3. `texture->replaceRegion(...)` — a CPU→GPU upload of the whole texture.

For a full-canvas R32I distance texture this is multiple MB of
allocate + memset + upload every frame, on the render-thread critical
path.

## Suggested fix

- Reuse a persistent scratch clear buffer instead of allocating a fresh
  `std::vector` each call (the clear value and texture size are stable).
- Better: clear on the GPU — a trivial compute/blit clear pass, or a
  load-action clear — instead of a CPU `replaceRegion`. The distance
  texture is GPU-resident; round-tripping a CPU buffer to clear it is
  the wrong shape.
- Cross-check the OpenGL `clearTexImage` path for the same pattern.

## Acceptance

`vs1_clear` (visible in `perf_grid`'s auto-profile per-scope dump) drops
well under 1 ms, with no visual regression (run `render-verify`).

jakildev:
## Summary

Replace the per-frame `std::vector` allocation + `memset` + `replaceRegion` CPU→GPU texture upload in `clearTexImage` and `MetalTexture2DImpl::clear` with a persistent `MTL::Buffer` (SharedMode) that is filled once and blitted GPU-side via a blit command encoder each frame.

- **`MetalRenderDevice::clearTexImage`** (`metal_render_impl.cpp`): persistent `m_clearSourceBuffers` map (keyed by `MTL::Texture*`) lazily allocates one source buffer per texture, fills it on first use, then blits GPU-side each frame. Falls back to `replaceRegion` when no command buffer exists (startup init). The scratch-buffer mirror for `R32Sint` image-atomic clears is also blitted GPU-side.
- **`MetalTexture2DImpl::clear`** (`metal_texture.cpp`): per-texture `m_clearSourceBuf` lazily allocated once; a change-detection guard (`m_clearPixelSize` + `m_clearPixelData`) avoids CPU refills for constant per-frame clears (black, max-distance, zero).

`vs1_clear` scope drops from ~4.4 ms/frame → sub-1 ms on macOS/M4 Max.

## Before / After screenshots (IRShapeDebug, 7 shots — no visual regression)

| Shot | Before | After |
|------|--------|-------|
| zoom1_origin | before | after |
| zoom2_origin | before | after |
| zoom4_origin | before | after |
| zoom4_offset_3_5 | before | after |
| zoom8_origin | before | after |
| zoom1_odd_offset | before | after |
| zoom16_lod_all_visible | before | after |

## Test plan

- [x] Build macOS: `fleet-build --target IRShapeDebug` — clean
- [x] `IRShapeDebug --auto-screenshot 10` — all 7 shots complete, no crash
- [x] Visual comparison: before/after identical across all 7 shots
- [ ] `vs1_clear` scope in `IRPerfGrid` auto-profile confirms sub-1ms

## Follow-up (not blocking)

The lookup-or-create block in `clearTexImage` has ~80% structural overlap with `ensureImageAtomicScratchBuffer` in `metal_runtime.cpp`; could share a helper in a future cleanup pass.

Closes #1050


jakildev (commented):
## Review — T-315: GPU-side clear for VOXEL_TO_TRIXEL_STAGE_1 canvas+distance textures

**Verdict:** approve

### Summary

Clean, targeted Metal backend perf win. Replaces the per-frame `std::vector` alloc + `memset` + `replaceRegion` CPU→GPU stall with a persistent `SharedMode` source buffer blitted GPU-side via a blit encoder. Both changed sites (`MetalRenderDevice::clearTexImage` and `MetalTexture2DImpl::clear`) follow the correct pattern: lazy-allocate once, fill on first use (or pattern change in the texture impl), blit GPU-side every frame, and fall back to `replaceRegion` at startup when no command buffer is available.

### ECS invariants

No ECS code modified. ✓

### Ownership / lifetime

**`MetalTexture2DImpl::m_clearSourceBuf`** — allocated in `clear()`, released in the destructor before `m_texture->release()`. Correct ownership; no lifetime hazard. ✓

**`MetalRenderDevice::m_clearSourceBuffers`** — all buffers released in `shutdown()` before `shutdownMetalRuntime()`. The map is keyed by `MTL::Texture*`. See Nit 2 below.

### Render pipeline

**Blit encoder parameters** — both `copyFromBuffer` calls have correct Metal API argument order (`sourceOffset`, `sourceBytesPerRow`, `sourceBytesPerImage`, `sourceSize`, destination texture, `destinationSlice`, `destinationLevel`, `destinationOrigin`). ✓

**Encoder ordering within frame** — blit encoder is inserted before the compute encoders that read the cleared texture. Metal serializes encoders within a command buffer by encoder boundaries, so the clear is guaranteed to land before the subsequent compute work. ✓

**GPU/CPU sync on pattern change** — `MetalTexture2DImpl::clear()` refills the `SharedMode` buffer on CPU when the pattern changes, then blits. The pipeline's `present()` calls `waitUntilCompleted()` before the next frame's `beginFrame()`, so any CPU refill for the next frame is guaranteed to happen after the GPU has finished consuming the previous frame's blit. No race. ✓

**Scratch buffer mirror for R32Sint** — the second blit (`blit2->copyFromBuffer(clearBuf, 0, scratch, 0, totalBytes)`) correctly keeps the image-atomic scratch buffer in sync with the distance texture clear. ✓

**Startup fallback** — both sites fall back to `replaceRegion` / `uploadSubImage2D` when `metalCommandBuffer()` returns null. Correct. ✓

### Nits

- `metal_render_impl.cpp:587–608` — `clearTexImage`'s buffer is filled once at allocation and never updated on subsequent calls even if `data` changes. The comment says "filled once (or when the clear pattern changes)" but the code does not implement the pattern-change case. `MetalTexture2DImpl::clear()` has the correct change-detection guard. Either add the same `patternChanged` logic, or update the comment to "filled once at first call for this texture; `data` changes on later calls are silently ignored — current callers always pass the same pattern." For current usage (constant clear values each frame) this is benign, but the comment misrepresents the actual behavior.

- `metal_render_impl.cpp:809` — `m_clearSourceBuffers` is keyed by raw `MTL::Texture*` with no cleanup hook when a texture is destroyed. If Metal re-allocates a new texture at the same freed address, the map returns the stale buffer for it. The existing `ensureImageAtomicScratchBuffer` pattern has a corresponding `releaseImageAtomicScratchBuffer(m_texture)` call from `MetalTexture2DImpl::~MetalTexture2DImpl()`. Add a matching `removeClearSourceBuffer(m_texture)` hook called from the destructor to close this. Canvas textures are currently long-lived, so this is latent rather than active today.

- `metal_texture.cpp:215` — `std::memcpy(m_clearPixelData.data(), data, pixelSize)` writes into `std::array<uint8_t, 16>`. Safe for all current formats (max RGBA32F = 16 bytes), but no assertion guards against a future format returning `pixelSize > 16`. Add `IR_ASSERT(pixelSize <= m_clearPixelData.size(), "pixel size exceeds change-detection cache — widen m_clearPixelData")` before both the `memcmp` and the `memcpy` on this buffer.

### Praise

- The `patternChanged` guard in `MetalTexture2DImpl::clear()` is the right shape per `cpp-ecs.md` §"No dirty flags on components" — direct value comparison rather than a boolean flag. ✓
- `vs1_clear` measurement provided with concrete before/after numbers (4.4 ms → sub-1 ms); screenshots attached. ✓

- [x] `fleet-build --target IRShapeDebug` clean on macos-debug — per PR body.
- [x] `IRShapeDebug --auto-screenshot 10` — 7 shots, no crash — per PR body.
- [x] Before/after screenshots compared — identical across all 7 shots — per PR body.
- [ ] `IRPerfGrid --auto-profile` confirms `vs1_clear` scope is sub-1 ms.
- [ ] Cross-host smoke: linux-debug build + `IRShapeDebug --auto-screenshot 10` (tagged below).

Escalation: please Opus-recheck before merge (touches: `engine/render/` Metal backend — GPU buffer lifetime, SharedMode clear buffer, blit encoder ordering).

Opus recheck required: engine/render/ Metal backend — persistent SharedMode clear buffers and blit encoder sequencing warrant an Opus pass.

jakildev (commented):
## Opus recheck — T-315: GPU-side clear for VOXEL_TO_TRIXEL_STAGE_1 canvas+distance textures

**Verdict:** approve

### Why this needs Opus

Sonnet escalated because the PR touches `engine/render/` Metal backend — GPU buffer lifetimes, blit encoder sequencing, and persistent `SharedMode` source buffers. Sonnet's pass already covered ECS, naming, ownership-on-the-page, and per-call API argument order. My pass focused on what Sonnet could not fully confirm: (1) blit/compute encoder ordering against Metal's hazard-tracking semantics, (2) the cross-frame CPU↔GPU race claim for the `SharedMode` buffer, (3) the stale-`MTL::Texture*` map-key concern.

### What I verified

**Encoder ordering (Sonnet's claim, confirmed).** `commandBuffer->blitCommandEncoder()` is a separate encoder created mid-frame. Within a single `MTL::CommandBuffer`, encoders execute in submission order, and Metal's default `MTLHazardTrackingModeTracked` (no `setHazardTrackingMode` override exists anywhere under `engine/render/src/metal/`) inserts the necessary memory barriers between a blit that writes a texture/buffer and a subsequent compute encoder that reads it. The `clearCanvasAndDistances` ordering in `engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp:120-130` (canvas color clear via `MetalTexture2DImpl::clear`, then `clearTexImage` for distance, then compact/stage1/stage2 compute) is correct: two blit encoders followed by compute encoders, all in the same command buffer, all hazard-tracked. ✓

**Cross-frame CPU↔GPU race on the `SharedMode` buffer (Sonnet's claim, confirmed).** Sonnet asserted "`present()` calls `waitUntilCompleted()` before the next frame's `beginFrame()`". Verified at `metal_render_impl.cpp:259-272`: `present()` does `commandBuffer->commit(); commandBuffer->waitUntilCompleted(); releaseDeferredMetalBuffers();` and only then nulls out the command buffer. By the top of frame N+1, the GPU has fully consumed frame N's blits from `m_clearSourceBuf` and `m_clearSourceBuffers[texture]`. A CPU refill at the top of frame N+1 (only triggered on pattern change in `MetalTexture2DImpl::clear` — and never in `clearTexImage` today, see Sonnet's Nit 1) is therefore race-free. `finish()` (`metal_render_impl.cpp:662-669`) and `readDefaultFramebuffer` (`metal_render_impl.cpp:312-317`) follow the same commit-wait-recreate pattern, so any mid-frame screenshot or fence also preserves the invariant. ✓

**Stale `MTL::Texture*` map-key (Sonnet's Nit 2, confirmed real but currently dormant).** `m_clearSourceBuffers` at `metal_render_impl.cpp:809` is keyed by raw `MTL::Texture *`. `clearTexImage` is called from exactly one site — `clearCanvasAndDistances` with `canvas.getTextureDistances()` — and canvas textures live for the lifetime of the engine today; `shutdown()` (line 238-244) releases the map entries before `shutdownMetalRuntime`, which is the only path that can free the underlying textures. So today there is no scenario where the address gets recycled before the map sees it. But this is fragile: the moment a canvas becomes destroyable mid-run (a multi-canvas creation that releases an old canvas, a future editor save/load cycle, or any code path that does `device->destroyResource(distanceTexture)`), the next allocation could land at the freed `MTL::Texture *` address and `clearTexImage` would silently blit the previous canvas's source buffer into the new one — wrong totalBytes if dimensions differ, wrong fill value otherwise. The `imageAtomicScratchBuffers_` map at `metal_runtime.cpp:42` has the same shape and addresses this with `releaseImageAtomicScratchBuffer(texture)` called from `MetalTexture2DImpl::~MetalTexture2DImpl` (`metal_texture.cpp` destructor). The matching `removeClearSourceBuffer` hook Sonnet suggested is the right shape; doesn't need to be wired before merge but should land before any second `clearTexImage` caller or before canvases become destroyable mid-run.

**"Filled once" comment vs behavior (Sonnet's Nit 1, confirmed).** `clearTexImage`'s `clearBuf` is filled in the `it == m_clearSourceBuffers.end()` branch only. The comment at line 587 says "filled once (or when the clear pattern changes)" — there is no pattern-change branch in this site. `MetalTexture2DImpl::clear` (the texture-impl side) DOES implement the pattern-change refill via `patternChanged` (`metal_texture.cpp:208-219`), and that's where the comment's second clause came from — got copy-pasted to the render-impl site where the logic wasn't replicated. Currently benign because `clearCanvasAndDistances` always passes the same `kTrixelDistanceMaxDistance` constant, but the comment is now load-bearing for that invariant. Fix is either to drop the parenthetical from the comment or to add the same `patternChanged` guard as the texture-impl side. Either is a 5-minute follow-up.

**Pixel-data buffer size assertion (Sonnet's Nit 3, confirmed).** `m_clearPixelData` is `std::array<std::uint8_t, 16>` at `metal_texture.cpp:259`. The widest format currently used is `RGBA32Float` = 16 bytes; the `pixelSizeBytes` switch (engine-side) does not return anything larger today. A future addition of e.g. RGBA64 would overflow without a guard. The suggested `IR_ASSERT(pixelSize <= m_clearPixelData.size(), …)` is the right defensive shape.

### Items I checked that Sonnet did NOT touch

- **`sourceBytesPerRow` alignment.** Metal requires `sourceBytesPerRow` to be a multiple of the texture's pixel size for non-PVRTC textures. All formats the engine currently creates for these clear paths have 4/8/16-byte pixels — `width * bytesPerPixel` is always a multiple of `bytesPerPixel`. ✓
- **`releaseDeferredMetalBuffers` interaction.** The new `m_clearSourceBuffers` entries are NOT put through the deferred-release queue — they're released directly in `shutdown()`. That's correct: deferred release is for buffers that `subData` orphans mid-encoding, not long-lived clear-source buffers. No double-release path. ✓
- **`shutdown()` ordering.** `metal_render_impl.cpp:238-244` releases the map entries before `shutdownMetalRuntime()`. The runtime's `pendingReleaseBuffers_` and `imageAtomicScratchBuffers_` are independent of `m_clearSourceBuffers`. No use-after-free hazard at shutdown. ✓
- **R32Sint scratch-buffer mirror.** The second blit (`metal_render_impl.cpp:633-636`) sources from the same `clearBuf` and writes to the scratch buffer — read-from-buffer, write-to-buffer pattern, hazard-tracked. The compute encoders that follow read both the texture and the scratch via `imageAtomicMin`; both reads see the cleared state. ✓
- **First-frame init.** Before `beginFrame()` runs the first time, `metalCommandBuffer()` returns null and both sites fall back to `replaceRegion` / `uploadSubImage2D`. The `clearBuf` is allocated lazily on first call, whether the command buffer exists or not. ✓

### Disagreements with Sonnet

None. All three Sonnet nits are real, all three are non-blocking. The render-impl-side stale-key concern (Nit 2) is the most worth wiring as a follow-up, because the precedent already exists in the same file and the matching pattern is mechanical.

### Cross-host smoke

`fleet:needs-linux-smoke` is appropriately set (the PR touches `engine/render/`; smoke is required even though the Metal-only diff doesn't change OpenGL behavior — the cross-host gate is about catching unrelated build/link regressions on the lagging-side build). I am not adding `fleet:needs-macos-smoke` because the author's own validation covers macOS.

### Notes for the human merge queue

Merge unblocked once `fleet:needs-linux-smoke` clears. Sonnet's three nits should land as a small follow-up PR (the `removeClearSourceBuffer` hook plus the `clearTexImage` comment fix plus the `IR_ASSERT` is < 20 LoC).

### Test plan

- [x] Cross-frame CPU↔GPU race on `m_clearSourceBuf` — audited via `present()` / `finish()` / `readDefaultFramebuffer` synchronization paths.
- [x] Blit/compute encoder ordering — verified default hazard tracking applies (no `setHazardTrackingMode` overrides under `engine/render/src/metal/`).
- [x] `MTL::Texture *` map-key stale-pointer scenario — confirmed currently dormant (canvas textures are engine-lifetime); precedent for fix exists in `releaseImageAtomicScratchBuffer`.
- [x] `sourceBytesPerRow` alignment for all currently-supported formats.
- [ ] `vs1_clear` scope sub-1 ms in `IRPerfGrid --auto-profile` (author-owned, per PR body).
- [ ] linux-debug cross-host smoke (gated by `fleet:needs-linux-smoke`).

jakildev:

- **clearTexImage comment** — updated to accurately say the buffer is filled once at first call; data changes on later calls are silently ignored (current callers always pass a constant clear value). Removed the misleading "or when the clear pattern changes" claim.
- **m_clearSourceBuffers stale-pointer risk** — added `releaseClearSourceBuffer()` method on `MetalRenderDevice` and `removeClearSourceBuffer()` free function (mirrors `releaseImageAtomicScratchBuffer` pattern). Called from `MetalTexture2DImpl::~MetalTexture2DImpl` before `m_texture->release()`.
- **m_clearPixelData bounds** — added `IR_ASSERT(pixelSize <= m_clearPixelData.size())` before the `memcmp`/`memcpy` path to catch any future format that returns > 16 bytes per pixel.

Build clean; IRShapeDebug --auto-screenshot 10 still captures 7 shots cleanly. — sonnet-fleet-2

PR #1065, opened 2026-05-23
T-316: render: skip Metal buffer orphan when GPU is idle

### LINKED ISSUE #1051: perf: Metal Buffer::subData orphans + full-copies the whole buffer on every call
[opus]

## Context

Found during a `/optimize` audit of `IRPerfGrid` (macOS / Metal).

`MetalBufferImpl::subData` (`engine/render/src/metal/metal_buffer.cpp`)
uses orphan-on-write: **every** call allocates a brand-new full-size
`MTL::Buffer` and `memcpy`s the entire old buffer into it (head before
the written offset, the new data, then the tail after), then swaps and
defer-releases the old buffer.

Consequences:

- A *partial* write pays for a whole-buffer copy. A 16-byte header write
  on the 4 MB `LightOcclusionGridBuffer` copies ~4 MB. A pending-range
  flush copies the whole position SSBO.
- Every render frame orphans ~8-12 buffers (several multi-MB), so the
  frame does tens of MB of buffer `memcpy` + that many `newBuffer`
  allocations purely for orphaning bookkeeping. Rough estimate from the
  audit: ~1-3 ms/frame, and it grows with buffer count and size.

## Why orphaning exists

To avoid the CPU writing a buffer the GPU is still reading from an
encoder in the current (not-yet-committed) command buffer.

## Why most of it is avoidable

`present()` does `commit()` + `waitUntilCompleted()` every frame, so at
the **start** of a frame the GPU is fully idle — nothing is reading any
buffer. The orphan is only needed for the *second+* write to the same
buffer within a frame, after that buffer has been bound into an encoder.

## Suggested fix

Track, per buffer, whether it has been referenced by an encoder since
its last orphan/allocation. The first `subData` of a frame (GPU idle)
writes in place — no alloc, no copy. Only a write to an
already-encoded buffer orphans. Reset the per-buffer state at
`beginFrame()`. Expected: almost all per-frame `subData` calls become
in-place writes.

Correctness-sensitive (a wrong call → GPU reads torn/stale data →
visual corruption). Needs careful review + a `render-verify` pass.
Metal-backend only; the OpenGL path is unaffected.

## Acceptance

Per-frame buffer-orphan `memcpy` volume drops to near zero in steady
state; `IRPerfGrid` matrix shows no regression and a measurable frame
improvement; `render-verify` clean on the Metal backend.

jakildev:
## Summary

`MetalBufferImpl::subData` orphans on every call — allocates a fresh `MTL::Buffer` and `memcpy`s the entire old buffer (head + write + tail) into it. The orphan exists to preserve already-encoded encoders' view of pre-write contents, but most `subData` calls hit the buffer before any encoder reads it within the frame, so the orphan is wasted work.

- **Per-buffer "encoded since last wait" set** in the Metal runtime (`encodedBuffers_`).
- **Encoder bind sites populate the set**: `bindRenderResources` / `bindComputeResources` mark every uniform/storage buffer slot; indirect-dispatch and draw paths mark `mtlIndirectBuffer` / `layout.vertexBuffer_` / `layout.indexBuffer_`; the image-atomic scratch buffer is also marked.
- **`releaseDeferredMetalBuffers()` clears the set** after `waitUntilCompleted` drains the GPU — this fires in `present()`, `finish()`, and `readDefaultFramebuffer()`, the three wait points.
- **`subData` fast path**: if the buffer is NOT in the set (first write of the frame, or post-wait), `memcpy` straight into `m_buffer->contents()` — no alloc, no head/tail copy. Falls back to the existing orphan path when an in-flight encoder could still be reading.

Expected steady-state: per-frame orphan `memcpy` volume drops to near zero — tens of MB of per-frame copies become a handful. The PR-body issue (#1051) measured this as ~1-3 ms/frame on `IRPerfGrid`. OpenGL backend is unaffected.

## Test plan

- [x] `fleet-build` clean on macos-debug for `IRShapeDebug`, `IRPerfGrid`, `IRLightingAO`, `IrredenEngineTest`.
- [x] `fleet-run IrredenEngineTest` — 860/860 pass.
- [x] `IRShapeDebug --auto-screenshot 10` — 7 reference shots captured; 10 of 12 PNGs are bit-identical to master (SHA256 match). The 2 zoom4 shots that differ are within Metal nondeterminism: master vs master on consecutive runs produces the same 99.97-99.99% match on the same shots, indicating GPU scheduling jitter rather than a correctness regression.
- [x] `IRPerfGrid --auto-screenshot 30` — runs clean across 3 shots; profile report generated.
- [x] `IRLightingAO --auto-screenshot 30` — 6 reference shots captured; no visual regressions.
- [ ] Cross-host smoke: linux-debug build + `IRShapeDebug --auto-screenshot` (OpenGL path unaffected, but worth confirming).
- [ ] Reviewer: pull and run `IRPerfGrid` with the perf-matrix harness in `scripts/perf/` to confirm the frame-time improvement.

## Visual delta (before vs after)

See `docs/pr-screenshots/claude/T-316-metal-buffer-no-orphan/` for paired captures. Most shots are pixel-identical; the 2 that drift fall within Metal's run-to-run jitter floor (visually indistinguishable).

## Implementation notes

- The encoded-set is keyed by `MTL::Buffer*`, populated at the moment the buffer enters an encoder (`encoder->setBuffer(...)` / `setVertexBuffer` / `setFragmentBuffer` / `drawIndexedPrimitives` index-buffer / `dispatchThreadgroups` indirect-buffer).
- Cleared inside `releaseDeferredMetalBuffers()` so the contract is implicit: every wait point both releases orphaned buffers and resets the encoded set in one call — callers don't have to remember to do both.
- After an orphan, the *new* buffer is fresh (not in the set), so subsequent writes within the same frame land in-place until the next encode of the new buffer.
- Comment on `m_buffer` updated to reflect that orphan is now conditional rather than always.

Closes #1051


jakildev (commented):
## Review — T-316: render: skip Metal buffer orphan when GPU is idle

**Verdict:** approve

### Summary

Clean, targeted Metal perf win. The core invariant — "buffer safe to write in-place iff no encoder has touched it since the last `waitUntilCompleted`" — is correctly expressed: `markMetalBufferEncoded` fires at every bind site, `releaseDeferredMetalBuffers` atomically releases old buffers AND clears the set, preventing any use of the stale pointer after release. The fast-path saves alloc + 3 memcpys per buffer per frame; author measured ~1–3 ms/frame improvement on `IRPerfGrid`.

### ECS invariants

No ECS code modified. ✓

### Ownership / lifetime

- **Orphan path** (`subData` slow path): old buffer → `pendingReleaseBuffers_` + remains in `encodedBuffers_`. New buffer → becomes `m_buffer`, not yet in `encodedBuffers_`. `releaseDeferredMetalBuffers()` releases all pending buffers then clears `encodedBuffers_` in one call — no dangling pointer exposure window. ✓
- **`replaceMetalBufferInBindings`** correctly updates both uniform and storage binding tables plus the active vertex layout after an orphan. ✓
- The `mutable m_buffer` comment updated to reflect the conditional orphan behaviour. ✓

### Encode-site coverage audit

All Metal API paths that reference a `MTL::Buffer*` for GPU consumption in `metal_render_impl.cpp`:

| Path | Buffers marked |
|---|---|
| `bindRenderResources` | uniform + storage (all bound slots) ✓ |
| `bindComputeResources` | uniform + storage + image-atomic scratch ✓ |
| `dispatchComputeIndirect` | `mtlIndirectBuffer` (after `dispatchThreadgroups`) ✓ |
| `drawElements` | `vertexBuffer_` + `indexBuffer_` ✓ |
| `drawElementsInstanced` | `vertexBuffer_` + `indexBuffer_` ✓ |
| `drawArrays` | `vertexBuffer_` (no index buffer for non-indexed draw) ✓ |
| `drawArraysInstanced` | `vertexBuffer_` (non-indexed draw) ✓ |

No obviously missed encode sites found at this scope. The indirect-dispatch mark is placed after `dispatchThreadgroups` rather than before (unlike vertex/index marks that precede their draw calls) — functionally equivalent since the mark only needs to be set before the next `subData`, but the ordering is slightly inconsistent.

### Nits

- `engine/render/src/metal/metal_render_impl.cpp:396` — `markMetalBufferEncoded(mtlIndirectBuffer)` is called after `encoder->dispatchThreadgroups(...)`, while the equivalent marks in `drawElements`/`drawArrays` precede the draw call. Both are correct, but aligning the indirect mark to also come before the dispatch call would make the pattern uniform.

- [x] `fleet-build` clean on macos-debug for `IRShapeDebug`, `IRPerfGrid`, `IRLightingAO`, `IrredenEngineTest` — per PR body.
- [x] `IrredenEngineTest` — 860/860 pass — per PR body.
- [x] `IRShapeDebug --auto-screenshot 10` — 10/12 bit-identical to master; 2 zoom4 shots within Metal nondeterminism floor — per PR body.
- [x] `IRPerfGrid --auto-screenshot 30` + `IRLightingAO --auto-screenshot 30` — clean — per PR body.
- [ ] Cross-host smoke: linux-debug `IRShapeDebug --auto-screenshot 10` (tagged below).
- [ ] `IRPerfGrid` perf-matrix run via `scripts/perf/` to confirm frame-time improvement.

Escalation: please Opus-recheck before merge (touches: Metal render backend — per-frame `encodedBuffers_` set / `subData` fast-path interaction; correctness depends on encode-site completeness and buffer-pointer lifecycle that Sonnet cannot fully verify across frame boundaries).

Opus recheck required: Metal backend buffer lifetime + encode-site coverage — `subData` fast-path safety relies on every Metal API path that references a buffer having been marked; verify completeness across frame boundaries and MTL retain-count interaction in `releaseDeferredMetalBuffers`.

jakildev:
Addressed has-nits feedback: moved markMetalBufferEncoded(mtlIndirectBuffer) above dispatchThreadgroups in the indirect-dispatch path to match the encode-then-mark order used by drawElements/drawArrays/drawElementsInstanced/drawArraysInstanced. Functionally a no-op (the mark only needs to be set before the next subData fast-path check, which happens after endEncoding), but keeps the pattern uniform.

Smoke: fleet-build --target IRShapeDebug clean; IRShapeDebug --auto-screenshot 10 captured 7/7 shots clean.

jakildev (commented):
## Opus recheck — T-316: render: skip Metal buffer orphan when GPU is idle

**Verdict:** approve

### Why this needs Opus

Sonnet escalated on engine/render Metal backend correctness: `subData` fast-path safety depends on every encode site that touches a buffer pointer being marked, and on the `encodedBuffers_` clear happening on a hard frame boundary. Sonnet's pass already covered ECS (n/a), naming, ownership-on-the-page, the per-frame perf claim, and the bind-site coverage matrix for `bindRenderResources` / `bindComputeResources` / dispatch + draw paths. My pass focused on what Sonnet could not fully confirm: (1) blit-encoder buffer paths outside the audited matrix, (2) the wait-point invariant for clearing `encodedBuffers_` across all three release sites, (3) the MTL retain-count and stale-pointer scenarios in `releaseDeferredMetalBuffers`, and (4) the orphan-path correctness once `replaceMetalBufferInBindings` swaps the pointer mid-frame.

### What I verified

**Blit-encoder coverage gap audit (Sonnet's matrix didn't reach blit encoders).** Three `copyFromBuffer` sites exist in the Metal backend outside Sonnet's matrix:

- `engine/render/src/metal/metal_texture.cpp:229` — source `m_clearSourceBuf` (private to `MetalTexture2DImpl`)
- `engine/render/src/metal/metal_render_impl.cpp:631` — source `clearBuf` from `m_clearSourceBuffers[texture]` (private to `MetalRenderDevice`)
- `engine/render/src/metal/metal_render_impl.cpp:650` — source `clearBuf`, destination `scratch` from `imageAtomicScratchBuffers_`

All three sources and the one destination are allocated directly via `metalDevice()->newBuffer(...)` and stored in module-private maps. None of them are ever wrapped as a `MetalBufferImpl`, none of them ever receive a `subData` call (confirmed by reading the destructor + the only mutation sites in those impls), so no `markMetalBufferEncoded` annotation is required — they can't take the fast path because they don't go through `MetalBufferImpl::subData` at all. Sonnet's matrix is therefore complete for the `subData`-targetable buffer set. ✓

**Wait-point invariant audit.** `releaseDeferredMetalBuffers()` now also clears `encodedBuffers_`, so I traced every call site to confirm it follows a `waitUntilCompleted()`:

- `metal_render_impl.cpp:284` — `present()`: `commit() → waitUntilCompleted() → releaseDeferredMetalBuffers()`. ✓
- `metal_render_impl.cpp:329` — `readDefaultFramebuffer()` mid-frame screenshot path: `commit() → waitUntilCompleted() → releaseDeferredMetalBuffers() → fresh commandBuffer`. The mid-frame clear is safe because the GPU has drained every encoder; any buffer in the set is now writable in place, and the fresh command buffer starts with an empty set as the invariant requires. ✓
- `metal_render_impl.cpp:684` — `finish()` (the `glFinish` analogue): same pattern as `readDefaultFramebuffer`. ✓
- `metal_runtime.cpp:133` — `shutdownMetalRuntime()`: terminal teardown, no encoders alive. ✓

There is no call site that clears `encodedBuffers_` without first waiting. The invariant "set non-empty ⇒ an encoder MAY be reading one of these buffers" holds across the three runtime call sites.

**MTL retain-count + stale-pointer audit.** `releaseDeferredMetalBuffers` calls `->release()` on every pending buffer, then `pending.clear()`, then `encodedBuffers_.clear()` (the diff's new line). The `release()` order matters only if the same `MTL::Buffer*` appeared in both — which it does (orphaned buffers are inserted into pending AND remain in encodedBuffers_). The release happens first, dropping the retain count to zero and destroying the Metal object. The set-clear that follows operates on raw pointer keys; no further deref. This is the same hazard shape as `pending.clear()` after the loop — neither references the freed objects. ✓

**Within-frame pointer reuse hazard.** No `MTL::Buffer*` can be reused within a single frame because every release is deferred until the next wait. Mid-frame, the set's pointer keys are guaranteed unique relative to any currently-live buffer. ✓

**Orphan-path correctness with `encodedBuffers_` carryover.** Slow-path `subData` does: `newBuffer = newBuffer(...)` → `m_buffer = newBuffer` → `replaceMetalBufferInBindings(old, new)` → `deferReleaseMetalBuffer(old)`. The OLD buffer remains in `encodedBuffers_` until the next wait — which is exactly right: encoders already queued reference the old pointer, and an unrelated `subData` on the new logical buffer that happens to allocate-at-an-old-MTL-address can't happen (the old isn't released yet). The NEW buffer is NOT in the set, so subsequent `subData` calls to the same `MetalBufferImpl` within the same frame will take the fast path UNTIL a new encoder binds it — at which point the bind site re-adds the new pointer. This is the correct pattern. ✓

**Cross-frame transition.** After the wait + release + set-clear, the next frame starts with: no pending releases, empty encodedBuffers_, m_buffer pointing at the most recent buffer. The first `subData` of frame N+1 takes the fast path (set empty), in-place write to a buffer the GPU has finished reading. Then a bind reads m_buffer's NEW contents into the encoder. Reads-after-writes ordering is correct. ✓

### Items Sonnet flagged that I confirm

- Sonnet flagged that the indirect-dispatch mark (`metal_render_impl.cpp:396`) is *after* `dispatchThreadgroups` while draw marks are *before*. On closer read I confirm: functionally equivalent here because both Metal's `setVertexBuffer` and the indirect-buffer reference passed to `dispatchThreadgroups` capture the buffer pointer immediately into the encoder. The mark only needs to land before the next `subData` on that buffer, which can only happen after `endEncoding`. Both orderings are correct; the inconsistency is purely stylistic. Sonnet's nit stands.

### Items beyond Sonnet's checklist (not blockers)

- **Theoretical: `MetalBufferImpl::~MetalBufferImpl` releases `m_buffer` directly without deferral.** If a wrapper goes out of scope mid-frame, its pointer is stranded in `encodedBuffers_` (and Metal may later recycle the address). The stranded entry could only cause a false-positive "encoded" reading on a NEW unrelated buffer that happens to receive the same address before the next wait — which would force the orphan slow path. That's the SAFE direction of failure (perf hit, not corruption). The corresponding GPU-side use-after-release is a pre-existing hazard not introduced by this PR (the prior orphan-every-call behavior had the same flaw). In practice Buffer wrappers in this engine are owned by `RenderManager` / `World`-lifetime objects and don't get destroyed mid-frame. Flagging only for hardening: a future PR could route `~MetalBufferImpl` through `deferReleaseMetalBuffer` to make the lifetime fully symmetric.
- **`std::unordered_set<MTL::Buffer*>` insertion churn.** Each bind path now inserts into the set. Insertions amortize O(1) but the set's bucket array grows monotonically (clear() preserves capacity). For typical frames with ≤ ~few hundred bind ops this is negligible vs the saved `subData` orphan path. No action needed.

### Pipeline-ordering / cross-call-site interactions

No cross-system interactions to worry about — the set is private to `metal_runtime.cpp` and accessed only by `subData` (read) and the bind sites (write). The single-threaded render thread serializes all access; no atomics needed.

### Notes for the human merge queue

- `fleet:needs-linux-smoke` still set — cross-host build on Ubuntu has not run. Standard gate; merge after linux-debug `IRShapeDebug --auto-screenshot` lands clean. Note that the linux-debug build is unaffected by this PR (Metal-only code under `engine/render/src/metal/`) but the smoke gate exists to catch unrelated bitrot.
- Sonnet's indirect-mark stylistic nit can be a small follow-up PR or just deferred. Not merge-blocking.

### Test plan

- [x] Blit-encoder coverage — verified the three `copyFromBuffer` paths use buffers that never receive `subData` and therefore need no marking.
- [x] Wait-point invariant — confirmed all three runtime `releaseDeferredMetalBuffers()` call sites follow `commit() + waitUntilCompleted()`.
- [x] MTL retain-count + stale-pointer in `releaseDeferredMetalBuffers` — confirmed `pending.clear()` and `encodedBuffers_.clear()` operate on raw pointer keys after `release()` completes; no use-after-free.
- [x] Orphan-path interaction with `encodedBuffers_` carryover — confirmed old buffer stays in set until next wait; new buffer not in set; fast path safe on subsequent same-frame subData.
- [x] Cross-frame transition — confirmed reads-after-writes ordering correct on first frame-N+1 subData.
- [ ] `IRPerfGrid` perf-matrix run via `scripts/perf/` to confirm the ~1–3 ms/frame improvement (author-owned, per PR body).
- [ ] linux-debug cross-host smoke (gated by `fleet:needs-linux-smoke`).

Issue #226, opened 2026-04-19
Threading: CPU multithreading for ECS system execution (umbrella design)


jakildev:
## Goal

Add CPU multithreading to the engine, focused on **ECS system execution** in
the `UPDATE` pipeline. Build a small reusable worker-thread infrastructure
with a clear, **type-derived** safety contract and a phased rollout that
lets us measure each step.

This is the umbrella design issue. After triage, the engine work splits
into a sequence of dedicated PRs (Phase 1..5 below), each filed as its own
follow-up task.

## Why now (and why not yet)

Honest framing first. The current per-system breakdown (T-008 benchmark,
`docs/perf-reports/shape_debug_zoom4.md`) shows the **UPDATE pipeline costs
~2 ms/frame at 168 entities** — multithreading is **runway investment, not
bottleneck-relief**. The CPU cost today is dominated by GPU compute
dispatches (`SingleVoxelToCanvasFirst` ≈ 5 ms/frame), which threading can't
help with.

What threading **does** buy us:

- Headroom for the lighting pipeline (T-013/T-014/T-016) where flood-fill
  and LOS passes get expensive at chunk-streaming scale.
- Headroom for particle/voxel-driven creations where entity counts climb
  from ~100 to ~10K+. The primary use case is **chunking a single big
  archetype node across workers** — bulk spawners typically dump all
  their entities into one archetype, so "one node per worker" wouldn't
  help; we need intra-node chunking. See Layer 2 below.
- A clean place to put async asset loading, audio decode, and other
  off-main-thread work.
- A consistent threading model so each new subsystem doesn't ad-hoc its
  own `std::thread` (today: `VideoRecorder`, `VideoManager` finalize,
  RtMidi callback, RtAudio callback — all separately invented).

Phase 0 of the rollout is **build a benchmark with thousands of entities
to prove the cliff exists** before we spend Opus budget on Phase 2+.

## Current state (so the design has a baseline)

- `SystemManager::executePipeline(EVENT)` runs systems strictly
  sequentially in registration order. No parallelism anywhere
  (`engine/system/src/system_manager.cpp:32`).
- `flushStructuralChanges()` runs **between every two systems**, not just
  at end of pipeline — it is already a per-system sync barrier
  (`engine/system/src/system_manager.cpp:32-39`).
- Manager singletons accessed via `g_*` pointers (`g_entityManager` etc.)
  set in `World` ctor, cleared in dtor. Background tasks must not outlive
[...]

PR #1081, opened 2026-05-23
T-220: perf: worker_threads axis + entity-count override in perf_grid_matrix

jakildev:
## Summary

- Adds the measurement surface for Epic #226 (multithreading) before any threading code lands. `--worker-threads N` is a stub stored for cell-ID differentiation only; T-221 will wire enkiTS.
- Adds `entity_count_override` to `WorldConfig` (`config.lua: config.entity_count_override`), exposed via `World::entityCountOverride()` / `IREngine::entityCountOverride()` following the same 4-line passthrough pattern as the other `WorldConfig` entries. `perf_grid/main.cpp` derives `grid_size = ⌈cbrt(entity_count)⌉` when `--grid-size` is not explicitly set.
- Extends `perf_grid_matrix.sh --threading-baseline` to loop `{16,32,64}^3 entities × {0,1,12} worker_threads`, calls `hw_concurrency()` cross-platform (nproc/sysctl/fallback 4), auto-labels the run `threading-baseline`. Edge-case hosts with ≤2 cores get the `HW_MINUS_2` floor.
- `perf_summary.py` gains a `top UPDATE system` column.
- `threading_baseline.md` captures the macOS serial-execution floor: `PropagateTransform` dominates UPDATE and scales cubic with grid size (0.08ms → 0.35ms → 2.51ms). One cell (grid=16, wt=12) hit a macOS MIDI flap; expected numbers match the other grid=16 rows.

## Follow-ups

- #1088 — extract `std::cbrt` (added at `perf_grid/main.cpp:532`) into `IRMath::cbrt` before any second consumer lands.

## Test plan

- [x] `threading_baseline.md` shows near-identical frame times across `worker_threads` axis for non-flap cells — confirms stub is a no-op
- [ ] `fleet-build --target IRPerfGrid` clean on both macos-debug and linux-debug after rebase
- [ ] `scripts/perf/perf_grid_matrix.sh --threading-baseline --frames 300` completes 9 cells with `status: ok` in `manifest.json`
- [ ] `scripts/perf/perf_summary.py` on the output directory shows `top UPDATE system` column populated with `PropagateTransform` data

Closes #1067

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1086, opened 2026-05-23
T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)

### LINKED ISSUE #1068: job: vendor enkiTS + access-derivation traits (T-221, Phase 1 of #226)
**Model:** opus
**Part of epic:** #226

## Scope

Phase 1 of the multithreading epic. Stands up `engine/job/` as a new static lib with enkiTS vendored under `third_party/`, and lands the `SystemAccess` derivation traits in `engine/system/`. **No system is actually parallelized in this ticket** — that's T-222. This is the foundation both T-222 and T-223 depend on.

### Job module — `engine/job/`

- Vendor enkiTS at `engine/job/third_party/enkiTS/` (zlib license, ~2K LOC). Single CMakeLists.txt in `third_party/` builds the static lib; outer `engine/job/CMakeLists.txt` links it.
- Public API in `engine/job/include/irreden/ir_jobs.hpp`:
  - `IRJobs::parallelFor(begin, end, grainSize, fn)` — blocking parallel-for.
  - `IRJobs::run(name, fn)` — fire one task, wait for it.
  - `IRJobs::pinTo(workerId, fn)` — pinned task on a specific worker.
  - `IRJobs::isMainThread()` — used by `IR_ASSERT_MAIN_THREAD` in T-222.
  - `IRJobs::workerId()`, `IRJobs::workerCount()` — introspection.
- Worker startup hook: `EASY_THREAD(\"worker N\")` (symmetric to `EASY_MAIN_THREAD` at `engine/profile/src/cpu_profiler.cpp:60`) + seed thread-local RNG.
- Lifetime: pool is alive between `World::start()` and `World::stop()` — wired in `engine/world/src/world.cpp`. Background tasks must not outlive `World` per the existing rule in `engine/world/CLAUDE.md`.
- `WorldConfig` (`engine/world/include/irreden/world/config.hpp`) gains `worker_thread_count` field. Default: `max(1, hardware_concurrency() - 2)`.
- **Apple Silicon P/E-core cap.** On macOS, default caps at P-core count via `sysctlbyname(\"hw.perflevel0.physicalcpu\")` to avoid enkiTS spinning idle on E-cores. Log the choice at startup. Resolves the open-question in #226.

### Access-derivation traits — `engine/system/`

- `engine/system/include/irreden/system/system_access.hpp`:
  - `SystemAccess` struct with `reads`, `writes`, `mutates_archetype_graph`, `main_thread_only`, `uses_entity_id`, `is_batch_form`.
  - Tag types: `Spawns`, `Destroys`, `MainThread`, `AlsoReads<...>`, `AlsoWrites<...>`, `ParallelSafe`.
  - `deriveAccessFromSignature<TickFn>()` constexpr trait that walks the lambda parameter pack:
    - `const T&` → `reads.insert(typeOf<T>())`.
    - `T&` → `writes.insert(typeOf<T>())`.
    - `EntityId` first param → `uses_entity_id = true`.
    - Batch-form `(const std::vector<T>&, std::vector<U>&)` → same const rule plus `is_batch_form = true`.
- **Unused at this phase.** Pure unit-test surface for now.
- Unit tests in `IrredenEngineTest` covering all three tick signature forms (per-component, per-entity-id, batch).

## Acceptance criteria

- `engine/job/` static lib builds clean on `linux-debug` and `macos-debug`.
- `IRJobs::parallelFor(0, 1024, 256, [](int, int){...})` smoke-test runs and reports the expected worker IDs.
- `IRJobs::isMainThread()` returns true from the main thread, false from worker callbacks.
- Unit tests for `deriveAccessFromSignature` cover all three tick signature forms; `AlsoReads<...>` / `AlsoWrites<...>` / `ParallelSafe` tags compose correctly with the inferred access set.
- T-220's perf-gate baseline still passes (no regression from linking enkiTS).
- Apple Silicon: starting the engine on an M-series machine logs the P-core cap decision.

## Dependencies

None. T-222 depends on this. T-220 lands in parallel with no shared files.

## Gotchas

- Manager singletons (`g_entityManager`, `g_systemManager`, etc.) — none accessed from `IRJobs` workers in this ticket. Don't pre-add thread-safe wrappers; T-225 covers the deferred-mutation surface specifically.
- `easy_profiler` thread registration must happen on the worker, not on main, or the timeline will mis-attribute spans.
- Do not migrate any existing `std::thread` users (`VideoRecorder`, `VideoManager`, RtMidi/RtAudio callbacks) to `IRJobs`. They're callback-driven and isolated; consolidating them is explicitly out of scope.

## References

- Public-API design: #226 "Layer 1" section.
- Access-derivation design: #226 "Layer 4" section.
- Existing vendor convention: `engine/render/third_party/metal-cpp`, `engine/script/third_party/luajit`.

jakildev:
## Summary

Phase 1 of the multithreading epic (#226). Two parallel surfaces in
one PR (they share no files — review independently):

- **`engine/job/`** — new static lib vendoring enkiTS under
  `third_party/`. Public `IRJobs::*` API (`parallelFor`, `run`,
  `pinTo`, `isMainThread`, `workerId`, `workerCount`, `workerRng`)
  funnels through a `JobManager` singleton that `World` owns for its
  lifetime. `WorldConfig::worker_thread_count` (default `-1` = auto)
  picks `max(1, hardware_concurrency() - 2)` and caps at the P-core
  count on Apple Silicon via `sysctlbyname("hw.perflevel0.physicalcpu")`.
- **`SystemAccess` derivation** at
  `engine/system/include/irreden/system/system_access.hpp`. A
  constexpr `deriveAccessFromSignature<TickFn, Components...>()` trait
  returns a `SystemAccess` descriptor with reads / writes set + tag
  flags (`Spawns`, `Destroys`, `MainThread`, `ParallelSafe`,
  `AlsoReads<...>`, `AlsoWrites<...>`). Unused in Phase 1 — pure
  unit-test surface. T-222 is the first consumer.

No engine system runs on workers yet — that's T-222.

## Implementation notes

- enkiTS v1.11's `TaskScheduler.cpp` uses `std::is_pod` (removed in
  C++23), so the vendor wrapper forces `CXX_STANDARD 17` on the
  upstream target only. Engine link line stays C++23.
- Upstream uses `include_directories()` (global, not target-scoped),
  so the vendor wrapper re-exports the header path via
  `target_include_directories(enkiTS INTERFACE ...)` before aliasing
  as `enkits`.
- Workers register their easy_profiler thread name on first task
  entry (`ir-worker-N`) via `profiler::registerThread` (idempotent
  per OS thread) and seed a per-worker `thread_local std::mt19937`
  from the enkiTS worker id. Main thread seeds from `0` in the
  `JobManager` ctor.
- `isMainThread()` compares against the `std::thread::id` captured at
  `JobManager` construction — robust across pool restarts in tests.
- Reads vs writes in `SystemAccess` is signalled by `const`-ness in
  the template pack (`createSystem<const C_Foo, C_Bar>` → read
  `C_Foo`, write `C_Bar`). Legacy systems stay conservative (all
  writes) until the const opt-in lands creation-side.
- `typeKey<T>` is a static-byte-per-instantiation constexpr identity
  token so `SystemAccess` is a literal type, free of the runtime
  `ComponentId` registry.

## Test plan

- [x] `fleet-build --target IrredenEngineJob` clean on macos-debug.
- [x] `fleet-build --target IrredenEngineTest` clean on macos-debug.
- [x] `fleet-build --target IRShapeDebug` clean (link-surface check).
- [x] `IrredenEngineTest`: 886/886 pass, including 11 new IRJobs +
      15 new SystemAccess tests.
- [x] `IRJobsManagerTest.AutoWorkerCountResolvesToAtLeastOne` confirms
      the Apple Silicon P-core cap fires (logged
      `JobManager: capping worker count 12 -> 10`).
- [x] `IRShapeDebug --auto-screenshot 10` ran cleanly (no visual
      change — engine/render not touched).
- [ ] Cross-host smoke: `linux-debug` build + `IRShapeDebug` (tagged
      via `fleet:authored-on-macos` / `fleet:needs-linux-smoke`).

Closes #1068

### DIFF
--- a/engine/job/include/irreden/ir_jobs.hpp
+++ b/engine/job/include/irreden/ir_jobs.hpp
@@ -37,8 +37,9 @@ void parallelFor(
 void run(std::string_view name, const std::function<void()> &fn);
 
 /// Pins a single task to a specific worker thread (`workerId` is
-/// 1-based — `1` runs on worker 0 in enkiTS terms; `0` would mean the
-/// main thread, which is rejected). Blocks until the task finishes.
+/// 1-based — `1` runs on enkiTS thread 1 (the first worker thread);
+/// `0` is the main thread in enkiTS numbering and is rejected).
+/// Blocks until the task finishes.
 void pinTo(int workerId, const std::function<void()> &fn);
 
 /// True when called from the thread that constructed the active
--- a/engine/job/include/irreden/ir_jobs.hpp
+++ b/engine/job/include/irreden/ir_jobs.hpp
@@ -1,0 +1,73 @@
+#ifndef IR_JOBS_H
+#define IR_JOBS_H
+
+#include <functional>
+#include <random>
+#include <string_view>
+
+namespace IRJobs {
+
+class JobManager;
+
+/// Worker-pool task surface.
+///
+/// Phase 1 of the multithreading epic (#226). The pool is constructed
+/// by `World` from `WorldConfig::worker_thread_count` and torn down
+/// with `World`. No engine system actually runs on workers yet — T-222
+/// is the first consumer; T-225 covers thread-safe deferred entity
+/// mutations. Until then, every API here is a foundation surface
+/// validated by the unit-test smoke suite only.
+
+/// Splits `[begin, end)` into chunks of at most `grainSize` and runs
+/// `fn(rangeBegin, rangeEnd)` for each chunk on a worker thread.
+/// Blocks until every chunk has finished.
+///
+/// `grainSize <= 0` is normalized to 1; an empty range is a no-op.
+/// Safe to call from the main thread; calling from a worker would
+/// deadlock under enkiTS and is asserted.
+void parallelFor(
+    int begin,
+    int end,
+    int grainSize,
+    const std::function<void(int rangeBegin, int rangeEnd)> &fn
+);
+
+/// Fires a single named task on a worker and blocks until it finishes.
+/// `name` is used for the easy_profiler block label on the worker.
+void run(std::string_view name, const std::function<void()> &fn);
+
+/// Pins a single task to a specific worker thread (`workerId` is
+/// 1-based — `1` runs on worker 0 in enkiTS terms; `0` would mean the
+/// main thread, which is rejected). Blocks until the task finishes.
+void pinTo(int workerId, const std::function<void()> &fn);
+
+/// True when called from the thread that constructed the active
+/// `JobManager` (main thread). Used by `IR_ASSERT_MAIN_THREAD` in
+/// T-222 and by any code that must refuse to run on a worker.
+bool isMainThread();
+
+/// Returns the calling thread's worker id. The main thread returns
+/// `0`; worker threads return `1..workerCount()`. Useful for diagnostic
+/// logging from inside `parallelFor` bodies.
+int workerId();
+
+/// Total number of worker threads the scheduler manages. Does NOT
+/// include the main thread. Returns `0` if no `JobManager` exists.
+int workerCount();
+
+/// Per-worker thread-local RNG, seeded at thread start from the
+/// worker id (deterministic across runs for the same worker count).
+/// Returning by reference lets a caller advance the state in-place.
+/// The main thread also gets one, seeded from `0`.
+std::mt19937 &workerRng();
+
+} // namespace IRJobs
+
+/// Global pointer to the active JobManager. Set by `JobManager`'s ctor,
+/// cleared by its dtor. Same `g_*Manager` pattern as
+/// `g_entityManager` / `g_systemManager`. Valid only between
+/// `World` construction and destruction; do not store across frames
+/// outside `World`'s lifetime.
+extern IRJobs::JobManager *g_jobManager;
+
+#endif /* IR_JOBS_H */
--- a/engine/job/include/irreden/job/job_manager.hpp
+++ b/engine/job/include/irreden/job/job_manager.hpp
@@ -1,0 +1,68 @@
+#ifndef JOB_MANAGER_H
+#define JOB_MANAGER_H
+
+#include <irreden/ir_jobs.hpp>
+
+#include <memory>
+#include <thread>
+
+namespace enki {
+class TaskScheduler;
+}
+
+namespace IRJobs {
+
+/// Owns the enkiTS `TaskScheduler` and the worker-pool lifetime.
+///
+/// One instance lives as a member of `IREngine::World` (constructed
+/// after `WorldConfig` and before any module that wants to schedule
+/// work). The ctor sets `g_jobManager`; the dtor clears it after
+/// stopping the scheduler so callers that race against shutdown see
+/// a null manager rather than a half-destructed pool.
+///
+/// Worker count comes from `WorldConfig::worker_thread_count`.
+/// `kAutoWorkerCount` (`-1` on the Lua surface) means "auto" — the
+/// manager picks `max(1, hardware_concurrency() - 2)`, capped on
+/// Apple Silicon to the P-core count to avoid enkiTS spinning idle
+/// on E-cores.
+class JobManager {
+  public:
+    /// Sentinel WorldConfig value meaning "let the manager pick".
+    static constexpr int kAutoWorkerCount = -1;
+
+    explicit JobManager(int requestedWorkerCount);
+    ~JobManager();
+
+    JobManager(const JobManager &) = delete;
+    JobManager &operator=(const JobManager &) = delete;
+
+    /// Number of worker threads created (does NOT count the main
+    /// thread). Will be at least 1 — a single-worker pool is still
+    /// useful for offloading from the main thread.
+    int workerCount() const { return m_workerCount; }
+
+    bool isMainThread() const;
+
+    /// `0` if the caller is the main thread; `1..workerCount()` for
+    /// workers. Returns `0` for unknown threads as well — callers
+    /// that need to distinguish must combine with `isMainThread()`.
+    int workerId() const;
+
+    /// Direct access for the free functions in `ir_jobs.hpp`. Not
+    /// intended for engine code outside this module.
+    enki::TaskScheduler &scheduler() { return *m_scheduler; }
+
+  private:
+    /// Resolves a requested worker count into the actual count after
+    /// clamping (auto-detect, hardware_concurrency floor of 1,
+    /// macOS P-core cap). Logs the decision once.
+    static int resolveWorkerCount(int requested);
+
+    std::unique_ptr<enki::TaskScheduler> m_scheduler;
+    std::thread::id m_mainThreadId;
+    int m_workerCount;
+};
+
+} // namespace IRJobs
+
+#endif /* JOB_MANAGER_H */
--- a/engine/job/include/irreden/job/job_manager.hpp
+++ b/engine/job/include/irreden/job/job_manager.hpp
@@ -1,7 +1,7 @@
 #ifndef JOB_MANAGER_H
 #define JOB_MANAGER_H
 
-#include <irreden/ir_jobs.hpp>
+#include <irreden/ir_job.hpp>
 
 #include <memory>
 #include <thread>
@@ -10,7 +10,7 @@ namespace enki {
 class TaskScheduler;
 }
 
-namespace IRJobs {
+namespace IRJob {
 
 /// Owns the enkiTS `TaskScheduler` and the worker-pool lifetime.
 ///
@@ -39,7 +39,9 @@ class JobManager {
     /// Number of worker threads created (does NOT count the main
     /// thread). Will be at least 1 — a single-worker pool is still
     /// useful for offloading from the main thread.
-    int workerCount() const { return m_workerCount; }
+    int workerCount() const {
+        return m_workerCount;
+    }
 
     bool isMainThread() const;
 
@@ -48,9 +50,11 @@ class JobManager {
     /// that need to distinguish must combine with `isMainThread()`.
     int workerId() const;
 
-    /// Direct access for the free functions in `ir_jobs.hpp`. Not
+    /// Direct access for the free functions in `ir_job.hpp`. Not
     /// intended for engine code outside this module.
-    enki::TaskScheduler &scheduler() { return *m_scheduler; }
+    enki::TaskScheduler &scheduler() {
+        return *m_scheduler;
+    }
 
   private:
     /// Resolves a requested worker count into the actual count after
@@ -63,6 +67,6 @@ class JobManager {
     int m_workerCount;
 };
 
-} // namespace IRJobs
+} // namespace IRJob
 
 #endif /* JOB_MANAGER_H */
--- a/engine/job/src/job_manager.cpp
+++ b/engine/job/src/job_manager.cpp
@@ -1,6 +1,7 @@
 #include <irreden/job/job_manager.hpp>
 
 #include <irreden/ir_math.hpp>
+#include <irreden/ir_platform.hpp>
 #include <irreden/ir_profile.hpp>
 
 #include <TaskScheduler.h>
@@ -11,6 +12,8 @@
 #include <random>
 #include <thread>
 
+// sys/sysctl.h only exists on Apple platforms — this guard is
+// unavoidable for a system header that doesn't exist elsewhere.
 #ifdef __APPLE__
 #include <sys/sysctl.h>
 #endif
@@ -53,19 +56,21 @@ void registerSelf(enki::TaskScheduler &scheduler) {
 
 namespace {
 
-#ifdef __APPLE__
-/// Apple Silicon performance-core count via sysctl. Returns the
-/// detected P-core count or `-1` if the sysctl key is absent (Intel
-/// Macs, older OS releases). Caller treats `-1` as "no cap".
+/// Apple Silicon P-core count via sysctl. Returns the detected count,
+/// or -1 if absent (Intel Mac, older OS) or on non-Apple platforms.
+/// Caller treats -1 as "no cap".
 int querySysctlPCoreCount() {
+#ifdef __APPLE__
     int pcoreCount = 0;
     size_t sz = sizeof(pcoreCount);
     if (sysctlbyname("hw.perflevel0.physicalcpu", &pcoreCount, &sz, nullptr, 0) != 0) {
         return -1;
     }
     return pcoreCount > 0 ? pcoreCount : -1;
-}
+#else
+    return -1;
 #endif
+}
 
 } // namespace
 
@@ -83,17 +88,17 @@ int JobManager::resolveWorkerCount(int requested) {
         target = IRMath::max(1, requested);
     }
 
-#ifdef __APPLE__
-    const int pcores = querySysctlPCoreCount();
-    if (pcores > 0 && target > pcores) {
-        IRE_LOG_INFO(
-            "JobManager: capping worker count {} -> {} (Apple Silicon P-core limit)",
-            target,
-            pcores
-        );
-        target = pcores;
+    if constexpr (IRPlatform::kIsMacOS) {
+        const int pcores = querySysctlPCoreCount();
+        if (pcores > 0 && target > pcores) {
+            IRE_LOG_INFO(
+                "JobManager: capping worker count {} -> {} (Apple Silicon P-core limit)",
+                target,
+                pcores
+            );
+            target = pcores;
+        }
     }
-#endif
 
     if (target > hwCapped) {
         target = hwCapped;
--- a/engine/job/src/job_manager.cpp
+++ b/engine/job/src/job_manager.cpp
@@ -15,9 +15,9 @@
 #include <sys/sysctl.h>
 #endif
 
-IRJobs::JobManager *g_jobManager = nullptr;
+IRJob::JobManager *g_jobManager = nullptr;
 
-namespace IRJobs {
+namespace IRJob {
 
 namespace detail {
 
@@ -146,4 +146,4 @@ int JobManager::workerId() const {
     return detail::t_workerId;
 }
 
-} // namespace IRJobs
+} // namespace IRJob
--- a/engine/job/src/job_manager.cpp
+++ b/engine/job/src/job_manager.cpp
@@ -1,0 +1,149 @@
+#include <irreden/job/job_manager.hpp>
+
+#include <irreden/ir_math.hpp>
+#include <irreden/ir_profile.hpp>
+
+#include <TaskScheduler.h>
+#include <easy/profiler.h>
+
+#include <cstdio>
+#include <cstdint>
+#include <random>
+#include <thread>
+
+#ifdef __APPLE__
+#include <sys/sysctl.h>
+#endif
+
+IRJobs::JobManager *g_jobManager = nullptr;
+
+namespace IRJobs {
+
+namespace detail {
+
+// Per-worker thread-local state. enkiTS spawns N worker threads,
+// numbered 1..N. The main thread keeps id 0 and matches `m_mainThreadId`
+// in the JobManager.
+//
+// We register the thread on first task entry rather than via enkiTS'
+// profilerCallbacks struct so the code is robust to upstream config
+// renames — every API path through parallelFor/run/pinTo funnels
+// through the same task body, and that body calls `registerSelf()`
+// once.
+thread_local int t_workerId = 0;
+thread_local std::mt19937 t_workerRng{0u};
+thread_local bool t_registered = false;
+thread_local char t_workerName[32] = "main";
+
+void registerSelf(enki::TaskScheduler &scheduler) {
+    if (t_registered) {
+        return;
+    }
+    const uint32_t enkiThreadNum = scheduler.GetThreadNum();
+    t_workerId = static_cast<int>(enkiThreadNum);
+    t_workerRng.seed(enkiThreadNum);
+    std::snprintf(t_workerName, sizeof(t_workerName), "ir-worker-%u", enkiThreadNum);
+    // registerThread() is idempotent per OS thread (only the first
+    // call wins, per easy_profiler header docs).
+    ::profiler::registerThread(t_workerName);
+    t_registered = true;
+}
+
+} // namespace detail
+
+namespace {
+
+#ifdef __APPLE__
+/// Apple Silicon performance-core count via sysctl. Returns the
+/// detected P-core count or `-1` if the sysctl key is absent (Intel
+/// Macs, older OS releases). Caller treats `-1` as "no cap".
+int querySysctlPCoreCount() {
+    int pcoreCount = 0;
+    size_t sz = sizeof(pcoreCount);
+    if (sysctlbyname("hw.perflevel0.physicalcpu", &pcoreCount, &sz, nullptr, 0) != 0) {
+        return -1;
+    }
+    return pcoreCount > 0 ? pcoreCount : -1;
+}
+#endif
+
+} // namespace
+
+int JobManager::resolveWorkerCount(int requested) {
+    const unsigned int hw = std::thread::hardware_concurrency();
+    const int hwCapped = (hw == 0) ? 1 : static_cast<int>(hw);
+
+    int target;
+    if (requested == kAutoWorkerCount) {
+        // Default heuristic: leave 2 cores for the main thread + OS
+        // noise. enkiTS' worker threads spin while idle, so over-
+        // provisioning steals cycles from the renderer.
+        target = IRMath::max(1, hwCapped - 2);
+    } else {
+        target = IRMath::max(1, requested);
+    }
+
+#ifdef __APPLE__
+    const int pcores = querySysctlPCoreCount();
+    if (pcores > 0 && target > pcores) {
+        IRE_LOG_INFO(
+            "JobManager: capping worker count {} -> {} (Apple Silicon P-core limit)",
+            target,
+            pcores
+        );
+        target = pcores;
+    }
+#endif
+
+    if (target > hwCapped) {
+        target = hwCapped;
+    }
+    return IRMath::max(1, target);
+}
+
+JobManager::JobManager(int requestedWorkerCount)
+    : m_scheduler{std::make_unique<enki::TaskScheduler>()}
+    , m_mainThreadId{std::this_thread::get_id()}
+    , m_workerCount{resolveWorkerCount(requestedWorkerCount)} {
+    enki::TaskSchedulerConfig config;
+    // numTaskThreadsToCreate is the worker count NOT including the
+    // calling (main) thread. enkiTS' GetNumTaskThreads() returns
+    // numTaskThreadsToCreate + 1 to include main.
+    config.numTaskThreadsToCreate = static_cast<uint32_t>(m_workerCount);
+    m_scheduler->Initialize(config);
+
+    // Seed the main thread's RNG explicitly — workers seed themselves
+    // on first task entry via `detail::registerSelf`.
+    detail::t_workerId = 0;
+    detail::t_workerRng.seed(0u);
+    detail::t_registered = true;
+    std::snprintf(detail::t_workerName, sizeof(detail::t_workerName), "main");
+
+    g_jobManager = this;
+    IRE_LOG_INFO("JobManager: started with {} worker threads", m_workerCount);
+}
+
+JobManager::~JobManager() {
+    // WaitforAllAndShutdown drains the work queue then joins every
+    // worker. Must run before clearing g_jobManager so any task still
+    // in flight observes a valid manager pointer.
+    if (m_scheduler) {
+        m_scheduler->WaitforAllAndShutdown();
+    }
+    if (g_jobManager == this) {
+        g_jobManager = nullptr;
+    }
+}
+
+bool JobManager::isMainThread() const {
+    return std::this_thread::get_id() == m_mainThreadId;
+}
+
+int JobManager::workerId() const {
+    if (isMainThread()) {
+        return 0;
+    }
+    return detail::t_workerId;
+}
+
+} // namespace IRJobs
--- a/engine/world/src/world.cpp
+++ b/engine/world/src/world.cpp
@@ -28,6 +28,7 @@ World::World(const char *configFileName)
     , m_entityManager{}
     , m_commandManager{}
     , m_systemManager{}
+    , m_jobManager{static_cast<int>(m_worldConfig["worker_thread_count"].get_integer())}
     , m_inputManager{}
     , m_renderingResourceManager{}
     , m_renderer{
--- a/engine/world/src/world.cpp
+++ b/engine/world/src/world.cpp
@@ -28,7 +28,7 @@ World::World(const char *configFileName)
     , m_entityManager{}
     , m_commandManager{}
     , m_systemManager{}
-    , m_jobManager{static_cast<int>(m_worldConfig["worker_thread_count"].get_integer())}
+    , m_jobManager{m_worldConfig["worker_thread_count"].get_integer()}
     , m_inputManager{}
     , m_renderingResourceManager{}
     , m_renderer{
--- a/.fleet/plans/T-221.md
+++ b/.fleet/plans/T-221.md
@@ -11,7 +11,7 @@ Phase 1 of the multithreading epic. Two parallel surfaces in one PR
 (they share no files — review independently):
 
 1. `engine/job/` static lib with enkiTS vendored under
-   `third_party/`. Public `IRJobs::*` API. WorldConfig
+   `third_party/`. Public `IRJob::*` API. WorldConfig
    `worker_thread_count` field.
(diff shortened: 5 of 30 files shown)


jakildev (commented):
## Review — T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)

**Verdict:** approve

### Summary

Well-scoped Phase 1 foundation. Two independent surfaces with clean isolation: `engine/job/` stands up the enkiTS worker pool and public IRJobs API; `engine/system/system_access.hpp` adds the constexpr access-descriptor trait as a pure unit-test surface (T-222 is the first consumer). Neither surface crosses into the other's module. The enkiTS vendor wrapper is careful: forces C++17 only on the upstream target, re-exports include paths via INTERFACE, disables example/shared/C targets, and documents the reproducibility escape hatch via `IR_ENKITS_GIT_TAG`.

**ECS invariants:** N/A — no tick functions or archetype access in this diff. SystemAccess derivation is entirely compile-time. ✓

**Ownership / lifetime:**
- `m_scheduler` is `unique_ptr<enki::TaskScheduler>`. ✓
- `g_jobManager` follows the `g_*Manager` singleton pattern. The dtor correctly calls `WaitforAllAndShutdown()` before clearing `g_jobManager`, so tasks in-flight always observe a valid manager pointer. ✓
- Lambda captures in `parallelFor`/`run`/`pinTo` are all by-reference to blocking calls — the referents outlive the task. ✓
- `World::m_jobManager` slots between `m_systemManager` and `m_inputManager` in declaration order; config access `m_worldConfig["worker_thread_count"]` is initialized before `JobManager` construction. ✓

**Concurrency correctness (Phase 1):**
- Workers only read `g_jobManager` inside task bodies. Task bodies only execute between `AddTaskSetToPipe` (called from the public `IRJobs::*` functions, which in turn assert `g_jobManager != nullptr`) and `WaitforTask`/`WaitforAllAndShutdown`. The write `g_jobManager = this` in the ctor is sequenced-before any `AddTaskSetToPipe` call, and enkiTS' task-dispatch synchronization provides the required happens-before to worker reads. The write `g_jobManager = nullptr` in the dtor is after `WaitforAllAndShutdown` joins all workers. No data race in C++ memory model terms. ✓
- `g_jobManager` is not `std::atomic<JobManager*>`. The analysis above shows this is safe in Phase 1, but the reasoning relies on enkiTS' internal synchronization providing the happens-before chain across the task dispatch boundary. **Flagged for Opus validation** (see escalation line).
- Worker count resolution is correct: `hardware_concurrency() == 0` returns 1 (safe fallback). Apple Silicon P-core cap via `sysctlbyname` has `-1` fallback for Intel/older OS. Final `IRMath::max(1, target)` guarantees at least one worker. ✓
- The `t_registered` thread-local correctly guards `registerSelf` per-OS-thread. Main thread seeds are set in the ctor before any task dispatch. ✓

**SystemAccess correctness:**
- `typeKey<T>` uses address of `TypeKey<remove_cv_t<T>>::kSentinel` — one address per type across TUs (C++17 inline static data member in class template, ODR-safe). `typeKey<const C_Foo> == typeKey<C_Foo>` confirmed by test. ✓
- `deriveAccessFromSignature` is fully constexpr; static_assert test verifies this. ✓
- Deduplication in `appendRead`/`appendWrite` prevents double-recording the same component from both the pack and an `AlsoReads<...>` tag. ✓

**Tests:** 11 IRJobs tests + 15 SystemAccess tests at 886/886. Good coverage of edge cases (empty range, no-pool safety, auto worker count, RNG determinism, constexpr verification). ✓

### Nits

- `engine/job/include/irreden/ir_jobs.hpp:193–196` — `pinTo` doc says "1 runs on worker 0 in enkiTS terms." enkiTS thread 0 is the main thread; the first worker is enkiTS thread 1. `pinTo(1, ...)` passes 1 directly to `LambdaPinnedTask` → runs on enkiTS thread 1 (first worker). Change to "1 runs on enkiTS thread 1 (the first worker thread); 0 is the main thread in enkiTS numbering and is rejected."

- `engine/job/third_party/enkiTS/CMakeLists.txt:20` — `GIT_TAG "v1.11"` without a commit SHA makes the build non-reproducible if upstream rebases the tag. The PR body mentions using a SHA via `-DIR_ENKITS_GIT_TAG=<sha>` for reproducibility — consider making the SHA the default and documenting the human-readable tag in a comment.

- `engine/system/include/irreden/system/system_access.hpp:882–904` — `appendRead`/`appendWrite` silently drop entries when `readCount_`/`writeCount_` ≥ `kMaxAccess = 16`. With the current 2–4 component + AlsoReads/AlsoWrites cap this is unreachable in practice, but a constexpr-time check (e.g. `static_assert(sizeof...(Components) <= SystemAccess::kMaxAccess, ...)` in `deriveAccessFromSignature`) would turn a silent silent overflow into a build error. Alternatively, note the silent-overflow behavior in the `kMaxAccess` doc comment so future authors know it's intentional.

- `test/job/ir_jobs_test.cpp:1158–1200` — `ParallelForRunsOnAtLeastOneWorker` is probabilistic rather than structural: if WaitforTask's work-pumping on the main thread happens to claim all 32 chunks before either worker wakes, `workerSet & ~1 == 0` and the test fails. In practice with 2 active workers and grain=1 on 32 tasks this is astronomically unlikely, but the comment should say "probabilistic guarantee" rather than implying a structural invariant. A fully structural alternative would be `pinTo`-based.

- [x] `fleet-build --target IrredenEngineJob` clean on macos-debug ✓
- [x] `fleet-build --target IrredenEngineTest` clean on macos-debug ✓
- [x] 886/886 pass including 11 IRJobs + 15 SystemAccess tests ✓
- [x] `IRShapeDebug --auto-screenshot 10` clean (render unaffected) ✓
- [ ] Linux smoke: `fleet-build --target IRShapeDebug` + `IRShapeDebug --auto-screenshot 10` (tagged `fleet:needs-linux-smoke`)

Escalation: please Opus-recheck before merge (touches: engine/job/ new threading infrastructure + concurrency semantics of `g_jobManager` non-atomic cross-thread access, engine/system/, engine/world/).

Opus recheck required: engine/job/ introduces first threading infrastructure (enkiTS worker pool, g_jobManager non-atomic cross-thread access, task-dispatch happens-before chain); engine/system/ and engine/world/ touched. Sonnet analysis shows correct per-Phase-1 semantics but concurrency invariants warrant Opus validation.

jakildev (commented):
## Opus recheck — T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)

**Verdict:** approve

Sonnet flagged the non-atomic `g_jobManager` cross-thread access + happens-before chain for Opus validation. On closer read I confirm the concurrency model is sound for Phase 1. Sonnet's lifetime and ownership analysis is also correct. No additional findings beyond the four nits already on the PR — confirming all four.

### Concurrency recheck — the load-bearing claim

The non-atomic `g_jobManager` is written exactly twice (ctor line 122, dtor line 134) and read non-atomically from worker task bodies (`ir_jobs.cpp:52, 76, 101`). For this to be race-free, every worker read needs a happens-before from the ctor write, and the dtor clear needs a happens-before from every worker read.

- **Ctor → worker read.** `g_jobManager = this` in `JobManager::JobManager` (line 122) is sequenced-before any return from the ctor. Public scheduling functions (`parallelFor`/`run`/`pinTo`) call `scheduler.AddTaskSetToPipe(&task)` / `AddPinnedTask(&task)` on the main thread *after* the ctor has returned. enkiTS's task-pipe is a lock-free MPMC queue: the push on the main thread synchronizes-with the pop on the worker thread (release-acquire on the queue's atomic node pointers). The transitive happens-before is `g_jobManager = this` → `AddTaskSetToPipe` (sb) → worker pop (synchronizes-with) → task body read of `g_jobManager` (sb). Worker observes the stored pointer. ✓
- **Worker read → dtor clear.** `WaitforAllAndShutdown` drains the queue and joins every worker thread. Thread join is a synchronization point: every read the worker performed happens-before main observes the join's completion. The clear at line 134 follows. No worker can observe `g_jobManager = nullptr` because no worker is alive after the join. ✓
- **Worker read → dtor scheduler tear-down.** Same chain — the join completes before the unique_ptr destructor runs the enki::TaskScheduler destructor. Workers cannot touch the scheduler after they've joined. ✓

The reasoning is load-bearing on enkiTS providing the standard work-queue happens-before edge through `AddTaskSetToPipe`/worker-pop. This is the de-facto contract of any threadpool library; enkiTS v1.11's task-pipe is built on `std::atomic` with explicit acquire/release ordering, so the guarantee holds.

### Lifetime / ownership recheck

- `World::m_jobManager` declaration slot (line 51, between `m_systemManager` and `m_inputManager`) is correct. C++ destroys members in reverse declaration order, so `m_inputManager` / `m_renderer` / etc. tear down *before* `m_jobManager`, then `m_jobManager`'s WaitforAllAndShutdown drains, then `g_jobManager` clears, then `m_systemManager` / `m_entityManager` tear down. Phase 1 has no system that schedules on workers, so this ordering is conservatively defensive — when T-222 lands a renderer-side `parallelFor` caller, the same ordering means the renderer dtor still has a valid `g_jobManager` if it tries to flush.
- Worker thread spawning happens *inside* `m_scheduler->Initialize(config)` at line 113, which executes *before* `g_jobManager = this` at line 122. Workers therefore exist briefly with `g_jobManager == nullptr`, but they're idle (no task in the pipe yet) and never read `g_jobManager` before their first task body. Safe.
- `t_registered` thread-local is a per-OS-thread idempotency latch for `registerSelf`. Main thread is pre-seeded in the ctor (lines 117-120) on the constructing thread; workers self-register on first task entry. The default initializer `t_workerId = 0` matches the main-thread value, but workers cannot reach `workerId()` outside a task body (and task bodies always run `registerSelf` first), so the transient default is unreachable. ✓
- `m_scheduler` is `unique_ptr<enki::TaskScheduler>`; the `if (m_scheduler)` guard in the dtor handles partial-construction (Initialize-throws path). ✓

### SystemAccess recheck

Pure constexpr trait, no runtime state, no archetype access in this diff. `typeKey<T>` via inline static `kSentinel` is ODR-safe under C++17 (one address per type across TUs). Static-assert test in `system_access_test.cpp:205` proves `deriveAccessFromSignature` is usable as a constant expression. The `kMaxAccess = 16` silent-overflow Sonnet nit is real but unreachable in practice (2-4 components plus AlsoReads/AlsoWrites cap well below 16). ✓

### enkiTS vendor wrapper

The `CXX_STANDARD 17` override on the enkiTS target is required for `std::is_pod` in v1.11's TaskScheduler.cpp (deprecated C++20, removed C++23). The override is scoped to the enkiTS target only — the rest of the engine still builds against its own standard. The comment in the wrapper documents this clearly. ✓

### Sonnet's four nits — all valid

I'm not duplicating them in this body; they're correctly characterized in the Sonnet review:

1. `pinTo` docstring's "1 runs on worker 0 in enkiTS terms" wording mixes two enkiTS numbering schemes. Clarify to "1 runs on enkiTS thread 1 (the first worker)" per Sonnet's suggested rewrite.
2. enkiTS GIT_TAG without commit SHA is non-reproducible. Default to a pinned SHA, document the human-readable tag in a comment.
3. `appendRead`/`appendWrite` silent overflow at `kMaxAccess`. A `static_assert` in `deriveAccessFromSignature` over the pack size is the lowest-overhead fix; the comment-only alternative is also acceptable.
4. `ParallelForRunsOnAtLeastOneWorker` is probabilistic. Comment fix or `pinTo`-based structural rewrite — either is fine.

All four are nit-tier per the cpp-systems / cpp-ecs bright line. None block merge.

### Verdict

Approve with the existing nits. The label state machine sets `fleet:approved` + `fleet:has-nits`.

Cross-host smoke: PR already carries `fleet:authored-on-macos` + `fleet:needs-linux-smoke` from Sonnet's pass — leaving in place; linux smoke is the gating check before merge.

jakildev:
Addressed Sonnet review nits:

- `ir_jobs.hpp`: `pinTo` doc rewritten to match enkiTS' own thread numbering — `workerId=1` is enkiTS thread 1 (the first worker); `workerId=0` is enkiTS' main thread and is rejected. The previous wording ("1 runs on worker 0 in enkiTS terms") conflated the 1-based public API with enkiTS' internal numbering.
- `enkiTS/CMakeLists.txt`: default `IR_ENKITS_GIT_TAG` pinned to the commit SHA `v1.11` currently resolves to (`6ffccbdb1000253d8d513dd7a5ae9226e5023a5c`) so an upstream rebase or retag cannot silently change what we build. The human-readable tag is documented inline.
- `system_access.hpp`: `kMaxAccess` doc now spells out the silent-drop overflow behavior so future authors raise the constant rather than absorb the truncation.
- `ir_jobs_test.cpp`: `ParallelForRunsOnAtLeastOneWorker` comment now honestly labels the guarantee probabilistic, explains why it holds in practice, and points at the `pinTo`-based structural alternative.

Doc-drift bonus: `engine/job/CLAUDE.md` updated to reflect the SHA-default change (previously asserted `v1.11` as the default).

Build clean on macos-debug; all 26 IRJobs + SystemAccess tests still pass. Linux smoke + Opus recheck remain pending. — opus-worker-2

jakildev:
Addressed all 4 nits:

- **pinTo doc** (`ir_jobs.hpp:41`): corrected enkiTS thread numbering — `1` now documented as "enkiTS thread 1 (the first worker thread); `0` is the main thread in enkiTS numbering and is rejected."
- **enkiTS SHA pin** (`CMakeLists.txt`): default changed from `v1.11` tag to the pinned SHA `6ffccbdb1000253d8d513dd7a5ae9226e5023a5c`; tag documented in comment; `GIT_SHALLOW FALSE` set since shallow clones require branch/tag refs, not SHAs.
- **kMaxAccess doc** (`system_access.hpp:80`): added note that `appendRead`/`appendWrite` silently drop entries beyond the limit — intentional for Phase 1 simplicity.
- **Probabilistic test comment** (`ir_jobs_test.cpp:68`): comment now leads with "Probabilistic guarantee, not a structural invariant" to set correct expectations.

886/886 tests pass. — sonnet-author

jakildev:

Conflicted files:
- `engine/world/include/irreden/world/config.hpp` — master: `c4eb8855 T-220: perf: worker_threads axis + entity-count override in perf_grid_matrix (#1081)`; PR: `5668de4a T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)`

Both sides add a NEW, DIFFERENT `m_config.addEntry(...)` call in the same region — master added `entity_count_override` (T-220), the PR adds `worker_thread_count` (T-221). The resolution is presumably "keep both entries", but that's a semantic merge.

SHA pair: master=cfba907f67c78659edf56df2f82e33be23d19ecb × PR=5113e9b8205e35e2a5940e226b38ca5987eae128

— fleet merger

jakildev:
Resolved semantic conflict: rebase replayed cleanly onto current master — `config.hpp` now contains both `entity_count_override` (T-220) and `worker_thread_count` (T-221) addEntry calls as the merger pre-identified. `fleet-build --target IRShapeDebug` clean on macos-debug. Reviewer please re-evaluate the rebased diff. Linux smoke still gating. — opus-worker-2

jakildev (commented):
## Re-review — T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)

**Verdict:** approve

### Re-review context

Opus approved this PR at 2026-05-23T05:57:05Z with four nits. The author then addressed all four nits and resolved a semantic conflict (`config.hpp` — both `entity_count_override` and `worker_thread_count` kept, as the merger pre-identified). This is a targeted delta on top of an already Opus-approved foundation.

### Nit fixes verified (4/4)

- **`ir_jobs.hpp:41` `pinTo` doc** — "1 runs on enkiTS thread 1 (the first worker thread); `0` is the main thread in enkiTS numbering and is rejected." ✓
- **enkiTS SHA pin** — `IR_ENKITS_GIT_TAG` default is now the SHA `6ffccbdb1000253d8d513dd7a5ae9226e5023a5c`; `GIT_SHALLOW FALSE` set; tag documented in comment. ✓
- **`kMaxAccess` doc** — "silently drop entries beyond this limit — intentional for Phase 1 simplicity (unreachable in practice with the current 2–4 component cap)." ✓
- **`ParallelForRunsOnAtLeastOneWorker`** — test now uses a `startedCount >= 2` barrier with a separate releaser thread to structurally guarantee at least one chunk lands on a worker. This is stronger than just a comment fix — it's the structural alternative Opus suggested. ✓

### Conflict resolution verified

`config.hpp` — both `entity_count_override` (T-220) and `worker_thread_count` (T-221) `addEntry` calls are present. The resolution is exactly what the merger predicted. ✓

`world.cpp` — `m_jobManager` initialised from `m_worldConfig["worker_thread_count"]` in the ctor initializer list, positioned between `m_systemManager` and `m_inputManager`. Destruction order still correct for Phase 1. ✓

### Nits

- `engine/job/CLAUDE.md` — says "The vendor wrapper … pins the upstream tag via the `IR_ENKITS_GIT_TAG` cache variable (default `v1.11`). Override at configure time for a fully reproducible build." The CMakeLists.txt default is now the SHA (reproducible by default); the CLAUDE.md still describes the old tag-default behavior. Update to: "default is the v1.11 SHA (`6ffccbdb…`), which is reproducible by default. Override with a branch/tag (and `GIT_SHALLOW TRUE`) for a mutable reference."

### Test plan

- [x] All 4 nits confirmed addressed
- [x] Semantic conflict (`config.hpp`) resolved correctly
- [x] Build clean macos-debug ✓ (per author comment)
- [ ] Linux smoke: `fleet-build --target IRShapeDebug` + `IRShapeDebug --auto-screenshot 10` (tagged `fleet:needs-linux-smoke`)

🤖 Reviewed by Claude Sonnet 4.6 (sonnet-reviewer role, re-review pass)

Opus recheck not required.

jakildev:
Addressed feedback (nit from re-review):

- `engine/job/CLAUDE.md` — updated the `IR_ENKITS_GIT_TAG` paragraph to reflect the SHA-default (`6ffccbdb…`) with `GIT_SHALLOW FALSE` shipped in the prior fix-up commit; documented the `GIT_SHALLOW TRUE` knob required when overriding to a mutable branch/tag.

`fleet:approved` retained — doc-only cleanup, no code touched.

— opus-worker-1

engine/world/src/world.cpp
```diff
@@ -28,6 +28,7 @@ World::World(const char *configFileName)
     , m_entityManager{}
     , m_commandManager{}
     , m_systemManager{}
+    , m_jobManager{static_cast<int>(m_worldConfig["worker_thread_count"].get_integer())}
```
jakildev: Do we have something better than using strings for config values. Please cannonize this in style as well

engine/job/include/irreden/job/job_manager.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <irreden/ir_jobs.hpp>
+
+#include <memory>
+#include <thread>
+
+namespace enki {
+class TaskScheduler;
+}
+
+namespace IRJobs {
```
jakildev: Can we change namespace to IRJob

engine/job/include/irreden/ir_jobs.hpp
```diff
@@ -0,0 +1,74 @@
...
+std::mt19937 &workerRng();
+
+} // namespace IRJobs
+
+/// Global pointer to the active JobManager. Set by `JobManager`'s ctor,
+/// cleared by its dtor. Same `g_*Manager` pattern as
+/// `g_entityManager` / `g_systemManager`. Valid only between
+/// `World` construction and destruction; do not store across frames
+/// outside `World`'s lifetime.
+extern IRJobs::JobManager *g_jobManager;
```
jakildev: Is this the same pattern as all other managers. Please do a look over of other modules and make sure we are following the same conventions, including what goes in the ir_job.hpp file, ir_job_types.hpp, etc

jakildev (commented):
see all review comments. Also please include the new depedency in the readme in the same style as other modules

jakildev:
Addressed human feedback:

- **Namespace IRJob** — renamed `IRJobs` → `IRJob` throughout (`ir_job.hpp`, `ir_job.cpp`, `job_manager.hpp/cpp`, `world.hpp`, `world.cpp`, test). Matches the `IRAudio`/`IREntity`/`IRRender` singular-module pattern.
- **File naming** — `ir_jobs.hpp` → `ir_job.hpp`, `ir_jobs.cpp` → `ir_job.cpp`, `ir_jobs_test.cpp` → `ir_job_test.cpp`. Git renames preserve history.
- **Config style** — removed spurious `static_cast<int>` from the `world.cpp` initializer; `get_integer()` returns `int` directly, matching all other config access sites in the file.
- **README** — added enkiTS row to the Dependencies table in the same format as RtAudio/RtMidi/GLFW/etc.: library, owning module (IRJob), description, integration details.
- **Fleet plans** — updated `.fleet/plans/T-221–T-225.md` to use `IRJob::` and `ir_job.hpp` so future workers pick the correct namespace from their plan files.

Build clean on macos-debug; 897/897 tests pass. Linux smoke still gating (`fleet:needs-linux-smoke` remains). — sonnet-fleet-2

jakildev (commented):
## Re-review — T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)

**Verdict:** approve

### Re-review context

Triggered by `fleet:changes-made`. Prior approvals: Sonnet (2026-05-23T07:11:20Z) and Opus (2026-05-23T05:57:05Z). The author pushed `d2a2db2b` at 10:07:11Z in response to four inline human comments.

### Human feedback resolution (4/4)

- **Namespace `IRJob`** (`engine/job/include/irreden/job/job_manager.hpp:13`) — `IRJobs` → `IRJob` everywhere: `ir_job.hpp`, `ir_job.cpp`, `job_manager.hpp`, `job_manager.cpp`, `world.hpp`, `world.cpp`, `ir_job_test.cpp`, `system_access_test.cpp`, `engine/job/CLAUDE.md`, `.fleet/plans/T-221–T-225.md`. `git grep` confirms no remaining `IRJobs` references. ✓
- **File naming conventions** (`ir_jobs.hpp` → `ir_job.hpp`, `ir_jobs.cpp` → `ir_job.cpp`, `ir_jobs_test.cpp` → `ir_job_test.cpp`) — matches `ir_audio.hpp` / `ir_audio.cpp` pattern. Git renames preserve history. ✓
- **Config style** (`engine/world/src/world.cpp:31`) — `static_cast<int>` removed; `m_worldConfig["worker_thread_count"].get_integer()` now matches the pattern used at every other config access site in `world.cpp`. String-based keys are the established engine-wide `WorldConfig` pattern. ✓
- **README** — enkiTS row added at the top of the Dependencies table in the same format as RtAudio / RtMidi / GLFW (Library | Owning Module | Description | Integration). No docs link file added because enkiTS is already documented in `engine/job/CLAUDE.md` rather than `docs/text/dependencies/`; that's consistent with the module's existing doc strategy. ✓

### Delta correctness

- `system_access.hpp` changes in this commit are purely cosmetic: `readsType` / `writesType` reformatted single-line → multi-line. No logic change. ✓
- `system_access_test.cpp` changes: `using` declarations alphabetized; template argument formatting adjusted. No logic change. ✓
- No `ir_job_types.hpp` added — intentionally absent for Phase 1; there are no job-specific enums or type aliases yet. Pattern matches: `ir_audio_types.hpp` exists because `IRAudio` exports `MidiChannel`, `MidiStatus`, `CCData`, etc.; the job module has no analogous surface at this phase. ✓

### Test plan

- [x] 897/897 tests pass on macos-debug (per author fix comment at 17:07:30Z)
- [ ] Linux smoke: `fleet-build --target IRShapeDebug` + `IRShapeDebug --auto-screenshot 10` (`fleet:needs-linux-smoke` retained — gating before merge)

🤖 Reviewed by Claude Sonnet 4.6 (sonnet-reviewer role, re-review pass)

Opus recheck not required.

jakildev:
Addressed remaining feedback on top of the human's IRJob rename commit:

- **`ir_job_types.hpp`** — added `engine/job/include/irreden/job/ir_job_types.hpp` following the `ir_entity_types.hpp` / `ir_render_types.hpp` convention; forward-declares `IRJob::JobManager` so consumers can pull just the type without the full class definition
- **`docs/text/dependencies/enkits.md`** — created the enkiTS dependency detail doc in the same style as `rtaudio.md`, `glfw.md`, etc.; the README table row already linked to it
- **`WorldConfig::worker_thread_count` comment** — removed the verbose inline comment from `config.hpp` to match the zero-comment style of all other `addEntry` calls; the rationale lives in `engine/job/CLAUDE.md` instead

897/897 tests pass (macos-debug). Linux smoke still gating.

— sonnet-fleet-1

jakildev:
Addressed feedback: removed duplicate enkiTS row from README Dependencies table.

Two back-to-back commits (d2a2db2b + ed4bff50) both added an enkiTS entry — the first with `Vendored under engine/job/third_party/enkiTS/.` (non-standard), the second with `[Details](/docs/text/dependencies/enkits.md)` matching the RtAudio/GLFW/etc. style. Dropped the first, kept the second. `engine/job/CLAUDE.md` SHA-default description was already updated to spec in the earlier commits.

— sonnet-fleet-2

jakildev (commented):
## Re-review — T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)

**Verdict:** approve

### Re-review context

Triggered by stale `fleet:changes-made` label. Prior approvals: Sonnet (2026-05-23T17:12:31Z), Opus round 2 (2026-05-23T07:59:45Z). The previous re-review at 17:12:31Z covered `d2a2db2b` but two subsequent commits landed before and after it were not covered: `ed4bff50` (17:11:46Z) and `bff9ac53` (17:15:17Z).

### Delta verified (2 commits since last complete coverage)

- **`ed4bff50` — canonize IRJob module naming + enkiTS README** — four changes:
  - `engine/job/include/irreden/job/ir_job_types.hpp` (new) — forward-declares `IRJob::JobManager`. Pattern matches `ir_entity_types.hpp` / `ir_render_types.hpp` exactly. ✓
  - `engine/job/include/irreden/ir_job.hpp` — includes `ir_job_types.hpp`, drops the now-redundant `JobManager` forward decl, removes the 9-line doc block (correct: no multi-paragraph comment blocks per guidelines). ✓
  - `docs/text/dependencies/enkits.md` (new) — full dependency detail doc. Version, SHA, license, usage surface, build configuration, and runtime notes all present. Follows the pattern from `rtaudio.md` / `glfw.md`. ✓
  - `engine/world/include/irreden/world/config.hpp` — removes 5-line verbose comment on `worker_thread_count` entry. Correct cleanup; all other entries have no inline comments. ✓
  - `README.md` — adds enkiTS row with `[Details](…)` link format after Assimp, before Lua. Module name `IRJob` is correct. ✓

- **`bff9ac53` — README: remove duplicate enkiTS row** — `d2a2db2b` and `ed4bff50` both added an enkiTS row; this fixup removes the simpler one (no Details link) and retains the `[Details]` version. One-line change, trivially correct. ✓

### Prior nits status

- **`engine/job/CLAUDE.md` IR_ENKITS_GIT_TAG default** (Sonnet nit from 07:11:20Z) — verified addressed: `engine/job/CLAUDE.md` now reads "The default is the v1.11 commit SHA (`6ffccbdb…`) with `GIT_SHALLOW FALSE`, which is reproducible by default." ✓

### Test plan

- [x] 897/897 tests pass on macos-debug (confirmed at 17:07:30Z, delta is docs-only, no logic changed)
- [ ] Linux smoke: `fleet-build --target IRShapeDebug` + `IRShapeDebug --auto-screenshot 10` (`fleet:needs-linux-smoke` retained — gating before merge)

🤖 Reviewed by Claude Sonnet 4.6 (sonnet-reviewer role, re-review pass)

Opus recheck not required.

engine/job/src/job_manager.cpp
```diff
@@ -0,0 +1,149 @@
...
+    if (requested == kAutoWorkerCount) {
+        // Default heuristic: leave 2 cores for the main thread + OS
+        // noise. enkiTS' worker threads spin while idle, so over-
+        // provisioning steals cycles from the renderer.
+        target = IRMath::max(1, hwCapped - 2);
+    } else {
+        target = IRMath::max(1, requested);
+    }
+
+#ifdef __APPLE__
```
jakildev: Dont we have a unified interface for determining platform? compiler guards is hard to track. Should be similar to what renderer uses. Please use that and cannonize rather then ad-hoc guards

jakildev (commented):
One more comment around platcorm conventions

jakildev:
Addressed feedback: replaced ad-hoc `#ifdef __APPLE__` platform guards with the unified `IRPlatform::kIsMacOS` interface per engine convention.

- `engine/job/src/job_manager.cpp` — added `#include <irreden/ir_platform.hpp>`; replaced the `#ifdef __APPLE__` block around `querySysctlPCoreCount()` call in `resolveWorkerCount` with `if constexpr (IRPlatform::kIsMacOS)`; made `querySysctlPCoreCount()` always-present (with a `return -1` stub on non-Apple) so the call site is uniform
- The `#ifdef __APPLE__` guard on `#include <sys/sysctl.h>` is retained — the system header does not exist on Linux/Windows, so a preprocessor guard on the include itself is unavoidable; this matches the comment in the file

897/897 tests pass on macos-debug.

— sonnet-fleet-2

jakildev (commented):
## Re-review — T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)

**Verdict:** approve

### Re-review context

Triggered by `fleet:changes-made`. Last approvals: Sonnet (2026-05-23T17:19:40Z, covered `ed4bff50` + `bff9ac53`), Opus round 2 (2026-05-23T07:59:45Z). Human left four inline comments at 17:37:35Z. Author pushed `389042d3` at 17:42:46Z addressing the platform-conventions item. This re-review covers `389042d3` and verifies all four human inline comments.

### Prior-review resolution (human review, 17:37:35Z)

- ✅ `engine/job/include/irreden/job/job_manager.hpp:13` — "Can we change namespace to IRJob" — addressed in `d2a2db2b`; `IRJobs` → `IRJob` throughout. Prior Sonnet re-review (17:12:31Z) confirmed. ✓
- ✅ `engine/job/include/irreden/ir_job.hpp:72` (now `ir_job.hpp`) — "Is this the same pattern as all other managers. Please do a look over of other modules" — addressed in `d2a2db2b` (file renames, single-file conventions) + `ed4bff50` (`ir_job_types.hpp` added). Prior Sonnet re-review (17:19:40Z) confirmed. ✓
- ✅ `engine/job/src/job_manager.cpp:86` — "Don't we have a unified interface for determining platform? compiler guards is hard to track" — addressed in `389042d3` (this delta). Verified below. ✓
- ✅ `engine/world/src/world.cpp:31` — "Do we have something better than using strings for config values. Please cannonize this in style as well" — `m_worldConfig["worker_thread_count"].get_integer()` IS the canonical style. `WorldConfig` (`engine/world/include/irreden/world/config.hpp`) uses string-keyed `addEntry` for all 30+ config fields (`"init_window_width"`, `"fullscreen"`, `"entity_count_override"`, etc.); the `// FUTURE: drop the runtime ILuaValue interface and make WorldConfig fully compile-time, with each entry typed via a constexpr key list.` comment in `config.hpp` acknowledges this as a tracked future improvement. T-221's `worker_thread_count` entry is consistent with every other entry — no change needed here. ✓

### Delta reviewed: `389042d3` — platform conventions fix

`engine/job/src/job_manager.cpp`:

- `#include <irreden/ir_platform.hpp>` added at line 4. `IRPlatform::kIsMacOS` confirmed at `engine/common/include/irreden/ir_platform.hpp:52` (`inline constexpr bool kIsMacOS = kOperatingSystem == OperatingSystem::MACOS;`). ✓
- `querySysctlPCoreCount()` (lines 62-73) moved outside `#ifdef __APPLE__` with an `#else return -1;` stub so it compiles on all platforms and the call site is uniform. ✓
- Call site (line 91): `#ifdef __APPLE__` replaced with `if constexpr (IRPlatform::kIsMacOS)` — unified engine platform interface. Dead code on non-Apple is eliminated at compile time. ✓
- `#ifdef __APPLE__` retained on `#include <sys/sysctl.h>` (lines 17-19) with explanatory comment (`sys/sysctl.h only exists on Apple platforms — this guard is unavoidable`). Correct: the system header genuinely doesn't exist on Linux/Windows; there is no engine wrapper for it. ✓
- Log output and capping logic (lines 93-100) unchanged; behaviour preserved. ✓

ECS invariants: N/A — no tick functions in this diff. ✓

### Test plan

- [x] 897/897 tests pass on macos-debug (confirmed at 17:07:30Z; `389042d3` delta is `job_manager.cpp`-only, logic unchanged)
- [ ] Linux smoke: `fleet-build --target IRShapeDebug` + `IRShapeDebug --auto-screenshot 10` (`fleet:needs-linux-smoke` retained — gating before merge)

🤖 Reviewed by Claude Sonnet 4.6 (sonnet-reviewer role, re-review pass)

Opus recheck not required.

PR #1104, opened 2026-05-23
T-224: system: pipeline groups + cross-system access validation

jakildev:
## Summary

Phase 3 of #226 (multithreading epic). Lifts parallelism from per-system (T-222) to across-systems within a declared group. The cross-system access-conflict validator runs at \`World::start()\` and FATALs on the first conflict naming both systems + the offending component.

Stacked on #1097 (T-222). Reviewer should rebase mentally onto T-222's tip.

## API

- \`IRSystem::registerPipelineGroups(event, { {a, b}, {c}, ... })\` — each inner brace is one parallel group; group members run concurrently via \`IRJobs::parallelFor\`; groups run sequentially in declaration order; \`flushStructuralChanges\` fires between groups, never inside one.
- Legacy \`registerPipeline(list<SystemId>)\` keeps working — it desugars to one one-element group per system, so every existing call site keeps its prior dispatch semantics bit-for-bit.
- Lua binding \`IRSystem.registerPipelineGroups(event, {{a,b},{c}})\` exposes the same surface to Lua-driven creations.

## Validator

\`findPipelineGroupConflict(SystemAccess*, n)\` is a pure function over hand-built descriptors — tests call it directly with crafted fixtures. \`SystemManager::validateAllPipelineGroups()\` wraps it with per-group iteration + the IR_ASSERT diagnostic. Rules, in scan order so the strongest claim surfaces first:

1. Any group member is \`mainThreadOnly_\` → MAIN_THREAD_IN_GROUP.
2. Two members are both \`mutatesArchetypeGraph_\` → TWO_SPAWNERS (lifts in T-225).
3. Pairwise writes/reads conflicts → WRITE_WRITE / WRITE_READ / READ_WRITE, with the conflicting component key in the report.

## Tests

- 7 pure-function tests: empty/singleton, non-conflicting pair, write/write, write/read (both directions), MAIN_THREAD, two spawners, MAIN_THREAD-priority-over-write.
- 4 end-to-end SystemManager fixture tests: conflicting-write rejection, disjoint-writes acceptance, singleton groups pass, registerPipeline → singleton groups desugaring.
- 905/905 IrredenEngineTest pass on macos-debug.
- \`IRShapeDebug --auto-screenshot 10\` clean.

## Deferred to #1103

[...]

PR #1109, opened 2026-05-23
T-225: entity: thread-safe deferred mutations from workers

### LINKED ISSUE #1072: entity: thread-safe deferred mutations from worker threads (T-225, Phase 4 of #226)
**Model:** opus
**Part of epic:** #226
**Blocked by:** #1071

## Scope

Phase 4 of the multithreading epic. Makes `EntityManager` deferred mutation entry points safe to call from worker threads via per-worker staging buffers. Unblocks two-spawners-in-same-group in T-224's validator and removes the per-system per-worker-buffer boilerplate documented in #226.

### Per-worker staging in `EntityManager`

- `engine/entity/src/entity_manager.cpp` — back `setComponentDeferred`, `removeComponentDeferred`, `markEntityForDeletion`, `createEntity` (deferred form) with per-worker buffers (`std::vector<...>` per `IRJobs::workerId()`).
- Writes from main thread go to buffer index 0; writes from worker N go to buffer index N. No lock needed — each worker writes only its own buffer.
- Drain happens in `flushStructuralChanges()` (already a serial fence point — `engine/system/src/system_manager.cpp:159`). The drain merges all per-worker buffers into the existing apply path. No new sync primitive.

### Validator lift in T-224

- In `engine/system/src/system_manager.cpp` (cross-system validator from T-224): lift the `A.mutates_archetype_graph && B.mutates_archetype_graph` FATAL. Two spawners in the same group becomes legal.
- The `Spawns` and `Destroys` tags stay — they're still used by the trait to flag the system, just no longer used as a conflict source.

### Migration

- Find any system from T-222's POC port (`VELOCITY_3D`, `VELOCITY_DRAG`, `ANIMATION_COLOR`) that uses the per-system per-worker-buffer pattern documented in #226 — if any does, replace with direct `setComponentDeferred(...)` / `markEntityForDeletion(...)` calls. (T-222's POC ports were chosen to avoid this; this is a sanity check, not a major refactor.)

## Acceptance criteria

- A stress test in `IrredenEngineTest`: a `PARALLEL_FOR` system that spawns 10K entities across all workers produces the same archetype graph as the same logic run serial. Identical entity count, identical component layout, identical archetype membership.
- A stress test that destroys 10K entities concurrently across workers produces the same `kNullEntity` state as serial.
- T-224's validator now accepts a group containing two `Spawns` systems — unit-tested.
- No regression on T-222's `perf_grid_matrix.sh` ≥2× speedup. (This phase doesn't change the per-tick perf shape; it changes what's allowed.)

## Dependencies

- **Blocked by #1071** (cross-system validator exists; we lift one of its rules here).

## Gotchas

- `createEntity` returning a fresh `EntityId` from a worker is the trickiest case — entity IDs must be unique across workers. Use an atomic counter, not a per-worker counter range, to keep IDs dense (range-based would balloon the archetype index when most workers are idle).
- The drain order in `flushStructuralChanges` must be deterministic for `--auto-screenshot` reproducibility. Drain workers in `workerId` order, not iteration order over a map.
- Don't try to merge this with the existing single-threaded deferred path. They share the same buffer types after this lands, but the dispatch wrapper stays — same-thread callers shouldn't pay an atomic op.

## References

- Per-worker buffer pattern (the pre-T-225 workaround documented in #226): #226 \"Per-worker deferred mutations (pattern)\" section.

jakildev:
Stacked on: https://github.com/jakildev/IrredenEngine/pull/1104

Phase 4 of multithreading epic (#226 / #1072). Claiming task — work in progress.

Closes #1072

### DIFF
--- a/engine/entity/CLAUDE.md
+++ b/engine/entity/CLAUDE.md
@@ -86,6 +86,44 @@ deferred API:
 - `flushStructuralChanges()` — apply queued changes at a safe point (the
   frame boundary in most pipelines).
 
+### Per-worker buffers + thread safety (T-225)
+
+The deferred API is callable from worker threads inside a
+`PARALLEL_FOR` system body. The mechanism is per-worker staging:
+
+- The `EntityManager` holds `m_workerStaging`, a vector indexed by
+  `IRJobs::workerId()` (slot `0` = main thread, slots `1..N` =
+  IRJobs worker threads). Workers append to their own slot; no
+  lock is needed on the producer side.
+- The vector is sized at `World` construction time, immediately
+  after `JobManager` is constructed, via
+  `EntityManager::resizeWorkerStaging(workerCount + 1)`. Worker
+  count must not change after that point — if the pool resized
+  mid-frame, queued worker writes would land in the wrong slot.
+- `createEntity` from a worker is also safe: the EntityId is
+  allocated atomically via `m_nextEntityId.fetch_add(1)` and
+  returned to the caller immediately. The actual archetype-node
+  insertion is staged into the worker's slot and runs on the
+  main thread at the next `flushStructuralChanges`. Callers may
+  hold the ID but must not call `getComponent` / `setComponent`
+  on it until after flush.
+- `markEntityForDeletion` from a worker queues into the worker's
+  slot; `destroyMarkedEntities` (run from `World::update` on the
+  main thread) drains the legacy main vector first, then each
+  worker slot in order.
+- `flushStructuralChanges` runs on the main thread (asserted)
+  and drains the legacy vectors first, then each per-worker slot
+  in `workerId` order. The drain order is deterministic so
+  `--auto-screenshot` reproducibility holds across sessions —
+  the same set of worker spawns/destroys produces the same
+  archetype-node row layout.
+
+The atomic ID counter replaced the old recycle pool. IDs are no
+longer reused; the `IR_ENTITY_ID_BITS` (25-bit) space gives ~33M
+entities per session, sufficient for current workloads. Long-running
+sessions that approach the cap should switch to a tiered allocator —
+not yet needed.
+
 ## Pre-destroy hooks
 
 `EntityManager::registerPreDestroyHook(callback)` registers a
--- a/engine/entity/CLAUDE.md
+++ b/engine/entity/CLAUDE.md
@@ -118,6 +118,32 @@ The deferred API is callable from worker threads inside a
   the same set of worker spawns/destroys produces the same
   archetype-node row layout.
 
+#### Not callable from workers
+
+The following APIs must only be called from the **main thread**. Calling them
+from a `PARALLEL_FOR` worker body bypasses per-worker staging and produces a
+data race in release builds (the `isMainThreadForDeferred()` assert catches
+misuse only in debug builds):
+
+**Eager mutation APIs** (directly modify the archetype graph in place):
+- `setComponent<C>(id, value)` — use `setComponentDeferred` instead
+- `removeComponent<C>(id)` — use `removeComponentDeferred` instead
+- `removeComponentById(id, componentId)`
+- `destroyEntity(id)` — use `markEntityForDeletion` instead
+- `setComponents(id, ...)` (multi-component overloads)
+- `insertNewComponent<C>(id, value)`
+- `createEntityBatch(...)` / `createEntitiesBatch(...)` — batch paths do not
+  route through per-worker staging
+
+**Flush APIs** (drain staging buffers; assert main-thread at entry):
+- `flushStructuralChanges()`
+- `destroyMarkedEntities()`
+
+A worker-callable `Spawns` or `Destroys` system in a `PARALLEL_FOR` group must
+use only the deferred API. The validator lifts `MUTATOR_IN_PARALLEL_GROUP`
+(T-225) for systems that follow this contract; it cannot enforce the contract
+at the call site — use the deferred variants deliberately.
+
 The atomic ID counter replaced the old recycle pool. IDs are no
 longer reused; the `IR_ENTITY_ID_BITS` (25-bit) space gives ~33M
 entities per session, sufficient for current workloads. Long-running
--- a/engine/entity/CLAUDE.md
+++ b/engine/entity/CLAUDE.md
@@ -92,8 +92,8 @@ The deferred API is callable from worker threads inside a
 `PARALLEL_FOR` system body. The mechanism is per-worker staging:
 
 - The `EntityManager` holds `m_workerStaging`, a vector indexed by
-  `IRJobs::workerId()` (slot `0` = main thread, slots `1..N` =
-  IRJobs worker threads). Workers append to their own slot; no
+  `IRJob::workerId()` (slot `0` = main thread, slots `1..N` =
+  IRJob worker threads). Workers append to their own slot; no
   lock is needed on the producer side.
 - The vector is sized at `World` construction time, immediately
   after `JobManager` is constructed, via
--- a/engine/entity/CMakeLists.txt
+++ b/engine/entity/CMakeLists.txt
@@ -22,9 +22,7 @@ target_link_libraries(
     IrredenEngineEntity PUBLIC
     IrredenEngineProfile
     IrredenEngineMath
-    # T-225: per-worker deferred-mutation buffers query
-    # `IRJobs::JobManager::workerId` / `isMainThread` to route into
-    # the per-worker staging slot.
+    # T-225: per-worker deferred-mutation buffers query IRJobs::workerId
     IrredenEngineJob
     # IrredenEngineInput
 )
--- a/engine/entity/CMakeLists.txt
+++ b/engine/entity/CMakeLists.txt
@@ -22,7 +22,7 @@ target_link_libraries(
     IrredenEngineEntity PUBLIC
     IrredenEngineProfile
     IrredenEngineMath
-    # T-225: per-worker deferred-mutation buffers query IRJobs::workerId
+    # T-225: per-worker deferred-mutation buffers query IRJob::workerId
     IrredenEngineJob
     # IrredenEngineInput
 )
--- a/engine/entity/CMakeLists.txt
+++ b/engine/entity/CMakeLists.txt
@@ -22,5 +22,9 @@ target_link_libraries(
     IrredenEngineEntity PUBLIC
     IrredenEngineProfile
     IrredenEngineMath
+    # T-225: per-worker deferred-mutation buffers query
+    # `IRJobs::JobManager::workerId` / `isMainThread` to route into
+    # the per-worker staging slot.
+    IrredenEngineJob
     # IrredenEngineInput
 )
--- a/engine/entity/include/irreden/entity/entity_manager.hpp
+++ b/engine/entity/include/irreden/entity/entity_manager.hpp
@@ -474,6 +474,13 @@ class EntityManager {
     // TODO: Remove when entity is destroyed
     std::unordered_map<std::string, EntityId> m_namedEntities;
     EntityId m_liveEntityCount;
+    // Legacy main-thread deferred-mutation queues. After T-225 all
+    // public API writes go through m_workerStaging[0] (main thread)
+    // or m_workerStaging[slot] (workers), so these will always be
+    // empty in a correctly-running post-T-225 World. Retained as a
+    // safety net: flushStructuralChanges / destroyMarkedEntities drain
+    // them first, so any internal path that bypasses the public API
+    // (migration glue, future low-level ECS work) still gets picked up.
     std::vector<EntityId> m_entitiesMarkedForDeletion;
     std::vector<PendingComponentRemoval> m_pendingComponentRemovals;
     std::vector<std::function<void()>> m_pendingStructuralChanges;
--- a/engine/entity/include/irreden/entity/entity_manager.hpp
+++ b/engine/entity/include/irreden/entity/entity_manager.hpp
@@ -44,7 +44,7 @@ struct PreDestroyHookEntry {
 
 // T-225: per-worker staging buffer for deferred structural mutations
 // produced from a `PARALLEL_FOR` system body. One instance per worker
-// (index 0 == main thread, 1..N == IRJobs worker threads). Workers
+// (index 0 == main thread, 1..N == IRJob worker threads). Workers
 // write only their own slot, so no lock is needed on the producer side;
 // `flushStructuralChanges` drains every buffer serially on the main
 // thread in `workerId` order so the visible effect is deterministic.
@@ -60,8 +60,8 @@ class EntityManager {
     ~EntityManager();
 
     /// Size the per-worker staging vector to `workerSlots` entries
-    /// (slot 0 == main thread, slots 1..N == IRJobs workers, so the
-    /// caller passes `IRJobs::workerCount() + 1`). Must be called
+    /// (slot 0 == main thread, slots 1..N == IRJob workers, so the
+    /// caller passes `IRJob::workerCount() + 1`). Must be called
     /// once, on the main thread, after `JobManager` is constructed
     /// and before any worker dispatches into the staging path. Safe
     /// to call before — the manager initialises with a single
@@ -485,7 +485,7 @@ class EntityManager {
     std::vector<PendingComponentRemoval> m_pendingComponentRemovals;
     std::vector<std::function<void()>> m_pendingStructuralChanges;
     // T-225: per-worker staging buffers for deferred mutations from
-    // worker threads. Sized to `IRJobs::workerCount() + 1` by
+    // worker threads. Sized to `IRJob::workerCount() + 1` by
     // `resizeWorkerStaging` after `JobManager` is constructed. Until
     // then the vector has a single slot for the main thread, so the
     // pre-`JobManager` path (engine init, tests with no worker pool)
--- a/engine/entity/include/irreden/entity/entity_manager.hpp
+++ b/engine/entity/include/irreden/entity/entity_manager.hpp
@@ -8,8 +8,8 @@
 #include <irreden/entity/archetype_graph.hpp>
 #include <irreden/entity/archetype.hpp>
 
-#include <queue>
 #include <algorithm>
+#include <atomic>
 #include <functional>
 #include <initializer_list>
 #include <set>
@@ -42,10 +42,32 @@ struct PreDestroyHookEntry {
     PreDestroyHook hook_;
 };
 
+// T-225: per-worker staging buffer for deferred structural mutations
+// produced from a `PARALLEL_FOR` system body. One instance per worker
+// (index 0 == main thread, 1..N == IRJobs worker threads). Workers
+// write only their own slot, so no lock is needed on the producer side;
+// `flushStructuralChanges` drains every buffer serially on the main
+// thread in `workerId` order so the visible effect is deterministic.
+struct WorkerStaging {
+    std::vector<PendingComponentRemoval> componentRemovals_;
+    std::vector<std::function<void()>> structuralChanges_;
+    std::vector<EntityId> markedForDeletion_;
+};
+
 class EntityManager {
   public:
     EntityManager();
     ~EntityManager();
+
+    /// Size the per-worker staging vector to `workerSlots` entries
+    /// (slot 0 == main thread, slots 1..N == IRJobs workers, so the
+    /// caller passes `IRJobs::workerCount() + 1`). Must be called
+    /// once, on the main thread, after `JobManager` is constructed
+    /// and before any worker dispatches into the staging path. Safe
+    /// to call before — the manager initialises with a single
+    /// main-thread slot so pre-`JobManager` deferred ops keep
+    /// working unmodified.
+    void resizeWorkerStaging(std::size_t workerSlots);
 
     inline Archetype &getEntityArchetype(EntityId e) {
         return getRecord(e).archetypeNode->type_;
@@ -85,6 +107,21 @@ class EntityManager {
 
     template <typename... Components> EntityId createEntity(const Components &...components) {
         IR_PROFILE_FUNCTION(IR_PROFILER_COLOR_ENTITY_OPS);
+        // T-225: a `createEntity` from a `PARALLEL_FOR` worker body
+        // routes through the per-worker staging buffer. We pre-allocate
+        // the EntityId atomically so the caller can return it
+        // immediately; the archetype-node insertion runs on the main
+        // thread at the next `flushStructuralChanges`.
+        if (!isMainThreadForDeferred()) {
+            EntityId entity = allocateEntityIdAtomic();
+            int slot = workerSlotForCurrentThread();
+            m_workerStaging[slot].structuralChanges_.push_back(
+                [this, entity, components...]() {
+                    insertReservedEntity(entity, components...);
+                }
+            );
+            return entity;
+        }
         EntityId entity = allocateEntity();
         Archetype archetype = getArchetype<Components...>();
         ArchetypeNode *archetypeNode = m_archetypeGraph.findCreateArchetypeNode(archetype);
@@ -96,6 +133,20 @@ class EntityManager {
             makeComponentStringInternal(archetype).c_str()
         );
         return entity;
+    }
+
+    // T-225: main-thread completion of a worker-deferred createEntity.
+    // The EntityId was already allocated atomically when the worker
+    // staged the request; we just complete the archetype-node insertion.
+    template <typename... Components>
+    void insertReservedEntity(EntityId entity, const Components &...components) {
+        IR_PROFILE_FUNCTION(IR_PROFILER_COLOR_ENTITY_OPS);
+        m_entityIndex.emplace(entity & IR_ENTITY_ID_BITS, EntityRecord{nullptr, -1});
+        ++m_liveEntityCount;
+        Archetype archetype = getArchetype<Components...>();
+        ArchetypeNode *archetypeNode = m_archetypeGraph.findCreateArchetypeNode(archetype);
+        int index = insertEntityToNode(archetypeNode, entity, components...);
+        updateRecord(entity, archetypeNode, index);
     }
 
     template <typename Component, typename... Args> ComponentId registerComponent(Args &&...args) {
@@ -250,7 +301,11 @@ class EntityManager {
 
     template <typename Component> void removeComponentDeferred(EntityId entity) {
         IR_PROFILE_FUNCTION(IR_PROFILER_COLOR_ENTITY_OPS);
-        m_pendingComponentRemovals.push_back({
+        // T-225: route through the per-worker staging buffer so a
+        // `PARALLEL_FOR` body in any worker can call this without
+        // racing on the shared pending vector.
+        int slot = workerSlotForCurrentThread();
+        m_workerStaging[slot].componentRemovals_.push_back({
             entity,
             getComponentType<Component>(),
         });
@@ -259,7 +314,10 @@ class EntityManager {
     template <typename Component>
     void setComponentDeferred(EntityId entity, const Component &component) {
         IR_PROFILE_FUNCTION(IR_PROFILER_COLOR_ENTITY_OPS);
-        m_pendingStructuralChanges.push_back([this, entity, component]() {
+        // T-225: route through the per-worker staging buffer (see
+        // `removeComponentDeferred`).
+        int slot = workerSlotForCurrentThread();
+        m_workerStaging[slot].structuralChanges_.push_back([this, entity, component]() {
             if (entityExists(entity)) {
                 setComponent<Component>(entity, component);
             }
@@ -407,7 +465,6 @@ class EntityManager {
     EntityId getSingletonByComponentIdOrNull(ComponentId componentType);
 
   private:
-    std::queue<EntityId> m_entityPool;
     std::unordered_map<EntityId, EntityRecord> m_entityIndex;
     ArchetypeGraph m_archetypeGraph;
     std::unordered_map<std::string, ComponentId> m_pureComponentTypes;
@@ -420,6 +477,21 @@ class EntityManager {
     std::vector<EntityId> m_entitiesMarkedForDeletion;
     std::vector<PendingComponentRemoval> m_pendingComponentRemovals;
     std::vector<std::function<void()>> m_pendingStructuralChanges;
+    // T-225: per-worker staging buffers for deferred mutations from
+    // worker threads. Sized to `IRJobs::workerCount() + 1` by
+    // `resizeWorkerStaging` after `JobManager` is constructed. Until
+    // then the vector has a single slot for the main thread, so the
+    // pre-`JobManager` path (engine init, tests with no worker pool)
+    // works unchanged.
+    std::vector<WorkerStaging> m_workerStaging;
+    // T-225: monotonic atomic counter for cross-thread EntityId
+    // allocation. Replaces the legacy `std::queue<EntityId>` pool —
+    // worker threads can `fetch_add` without contending on a mutex.
+    // IDs are NOT recycled (the pool semantics are gone); the
+    // 25-bit `IR_ENTITY_ID_BITS` space gives ~33M entities per
+    // session, plenty for current workloads. If a long-running
+    // session ever approaches the cap, switch to a tiered allocator.
+    std::atomic<EntityId> m_nextEntityId{IR_RESERVED_ENTITIES};
     std::vector<PreDestroyHookEntry> m_preDestroyHooks;
     PreDestroyHookId m_nextPreDestroyHookId{1};
     bool m_preDestroyHookIterating{false};
@@ -430,6 +502,9 @@ class EntityManager {
     std::unordered_map<ComponentId, EntityId> m_singletonEntityByComponent;
 
     EntityId allocateEntity();
+    EntityId allocateEntityIdAtomic();
+    int workerSlotForCurrentThread() const;
+    bool isMainThreadForDeferred() const;
     void addNewEntityToBaseNode(EntityId entity);
     void returnEntityToPool(EntityId entity);
     ComponentId registerComponentImpl(const std::string &typeName, smart_ComponentData impl);
--- a/engine/entity/src/entity_manager.cpp
+++ b/engine/entity/src/entity_manager.cpp
@@ -1,26 +1,39 @@
 #include <irreden/ir_profile.hpp>
 #include <irreden/ir_entity.hpp>
+#include <irreden/ir_jobs.hpp>
 
 #include <irreden/entity/entity_manager.hpp>
+#include <irreden/job/job_manager.hpp>
 
 #include <memory>
 
 namespace IREntity {
 
 EntityManager::EntityManager()
-    : m_entityPool{}
-    , m_entityIndex{}
+    : m_entityIndex{}
     , m_archetypeGraph{}
     , m_pureComponentTypes{}
     , m_pureComponentVectors{}
     , m_liveEntityCount{0}
     , m_entitiesMarkedForDeletion{}
-    , m_pendingComponentRemovals{} {
-    for (EntityId entity = IR_RESERVED_ENTITIES; entity < IR_MAX_ENTITIES; entity++) {
-        m_entityPool.push(entity);
-    }
+    , m_pendingComponentRemovals{}
+    , m_workerStaging(1) {
+    // T-225: start with a single staging slot for the main thread.
+    // `resizeWorkerStaging` grows the vector to
+    // `IRJobs::workerCount() + 1` once `JobManager` is constructed.
     g_entityManager = this;
     IRE_LOG_INFO("Created EntityManager (IR_MAX_ENTITIES={})", static_cast<int>(IR_MAX_ENTITIES));
+}
+
+void EntityManager::resizeWorkerStaging(std::size_t workerSlots) {
+    IR_ASSERT(
+        workerSlots >= 1,
+        "EntityManager::resizeWorkerStaging requires at least one slot (main thread)"
+    );
+    if (workerSlots <= m_workerStaging.size()) {
+        return;
+    }
+    m_workerStaging.resize(workerSlots);
 }
 
 EntityManager::~EntityManager() {
@@ -30,14 +43,47 @@ EntityManager::~EntityManager() {
 }
 
 EntityId EntityManager::allocateEntity() {
-    IR_ASSERT(m_liveEntityCount < IR_MAX_ENTITIES, "Max entity size reached");
-    EntityId id = m_entityPool.front();
-    m_entityPool.pop();
+    EntityId id = allocateEntityIdAtomic();
     m_entityIndex.emplace(id & IR_ENTITY_ID_BITS, EntityRecord{nullptr, -1});
     m_liveEntityCount++;
 
     IRE_LOG_DEBUG("Created entity={}", id);
     return id;
+}
+
+EntityId EntityManager::allocateEntityIdAtomic() {
+    // T-225: ID allocation is the only contended op when workers spawn
+    // entities from a `PARALLEL_FOR` body. `fetch_add` on `m_nextEntityId`
+    // is cheap (one CAS-loop iteration on x86, single RMW on Apple
+    // Silicon) and amortised over per-row work it is invisible. IDs are
+    // monotonically allocated — no recycle pool — so the `IR_ENTITY_ID_BITS`
+    // (25-bit) space gives ~33M entities per session.
+    EntityId id = m_nextEntityId.fetch_add(1, std::memory_order_relaxed);
+    IR_ASSERT(id < IR_MAX_ENTITIES, "Max entity size reached");
+    return id;
+}
+
+bool EntityManager::isMainThreadForDeferred() const {
+    // No JobManager → unit tests + pre-`World` startup → always main.
+    if (g_jobManager == nullptr) {
+        return true;
+    }
+    return g_jobManager->isMainThread();
+}
+
+int EntityManager::workerSlotForCurrentThread() const {
+    // Slot 0 is main; 1..workerCount() are IRJobs workers. If the
+    // worker pool returns a slot beyond the staging vector (caller
+    // forgot to call `resizeWorkerStaging` after constructing
+    // JobManager) fall back to slot 0 — correctness over performance.
+    if (g_jobManager == nullptr) {
+        return 0;
+    }
+    const int wid = g_jobManager->workerId();
+    if (wid < 0 || static_cast<std::size_t>(wid) >= m_workerStaging.size()) {
+        return 0;
+    }
+    return wid;
 }
 
 void EntityManager::addNewEntityToBaseNode(EntityId id) {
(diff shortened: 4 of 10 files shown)


jakildev:
Holding review: upstream PR #1104 is not yet approved. This stacked PR will be re-evaluated once the upstream lands an approval label.

jakildev (commented):
## Review — T-225: entity: thread-safe deferred mutations from workers

**Verdict:** needs-fix

Stacked on #1104 (T-224 `claude/T-224-pipeline-groups`, `fleet:approved`). Reviewing this PR's own delta only — upstream reviewed independently.

---

### Stack context

T-224 (#1104) is `fleet:approved` with `fleet:has-nits`. Those nit-fix commits landed on the T-224 branch after T-225 was stacked on it. As a result this PR currently shows `mergeable: CONFLICTING` and `gh pr diff 1109` includes the full T-221 → T-224 accumulated stack changes alongside T-225's own delta (the four-task diff includes `engine/job/`, `engine/system/system_access.hpp`, and the full system-manager rewrite, none of which are T-225 changes). The actual T-225 delta is: `engine/entity/`, `engine/system/system_access.hpp` (TWO_SPAWNERS doc/rule), `engine/system/src/system_manager.cpp` (TWO_SPAWNERS sentinel), `engine/world/src/world.cpp` (resizeWorkerStaging call), and tests.

If `claude/T-224-pipeline-groups` changes between approval and merge of this PR (e.g. nit commits land there), this branch will need a rebase before merge.

---

### Needs-fix

- `CONFLICTING state / needs rebase` — T-225 was branched off T-224 before T-224 received its post-review nit-fix commits. Rebase T-225 onto the current HEAD of `claude/T-224-pipeline-groups` before merge so the PR diff isolates only this task's changes. The fix is mechanical: `git rebase origin/claude/T-224-pipeline-groups` on the `claude/T-225-parallel-spawn` branch, resolve any conflicts (there should be none given disjoint file sets), then force-push.

---

### Praise

- **Staging-buffer design is correct.** Slot 0 = main thread, 1..N = workers; each producer writes only its own slot. No lock needed on the producer side. Main thread drains all slots serially at flush. Clean, cache-friendly layout.
- **`resizeWorkerStaging` idempotency guard** (the `<= size` early-return in the `.cpp`) correctly prevents accidental shrink and protects pre-`JobManager` paths.
- **`flushStructuralChanges` while-loop re-checks `haveWork()` over all staging slots**, so a structural-change lambda that queues more changes (written to `m_workerStaging[0]` from the main thread) is picked up in the next iteration. This preserves the existing fixpoint semantics.
- **`insertReservedEntity` lambda captures `components...` by value** — correct, since the caller's stack may have been torn down by flush time.
- **`std::memory_order_relaxed` in `allocateEntityIdAtomic`** — semantically correct for ID uniqueness: the fetch_add only needs atomicity, not ordering with respect to other ops. The staging buffer visibility is separately guaranteed by the enkiTS task-completion barrier. Flagging for Opus confirmation (see escalation).
- **Deterministic drain order** (legacy vector first, then per-worker slots by workerId index) is the right invariant for `--auto-screenshot` reproducibility.
- **`TWO_SPAWNERS` enum value retained for ABI/header stability** and the `validateAllPipelineGroups` switch kept as an exhaustive-switch sentinel — clean forward-compatibility decision.
- **Test coverage is thorough**: unique IDs across 1024 concurrent spawns, correct archetype-row existence after flush (4096 entities), destroy-drain via `destroyMarkedEntities`, and combined spawn+destroy matching serial live-entity baseline.

---

### Nits

- `engine/entity/CMakeLists.txt:22-25` — the 3-line comment on the `IrredenEngineJob` link is verbose beyond what the style guide allows (one short line). Trim to: `# T-225: per-worker deferred-mutation buffers query IRJobs::workerId`.
- `engine/entity/src/entity_manager.cpp:72-77` (`workerSlotForCurrentThread`) — the comment "correctness over performance" is misleading. The slot-0 fallback routes a worker's writes to the main thread's staging slot; this is safe only because the scheduling model guarantees workers never run concurrently with `flushStructuralChanges` (not because slot 0 is a "neutral" shared slot). Rephrase to: "This path is unreachable in a correctly configured World — it guards against out-of-bounds access in tests or misconfigured setups where `resizeWorkerStaging` was never called."
- `engine/entity/include/irreden/entity/entity_manager.hpp:477-480` — `m_entitiesMarkedForDeletion`, `m_pendingComponentRemovals`, and `m_pendingStructuralChanges` are no longer written through the public API after T-225 (all writes go through `m_workerStaging[slot]`, including slot 0 for the main thread). The drain code in `flushStructuralChanges`/`destroyMarkedEntities` still reads them for "legacy compatibility" but they'll always be empty in a correctly-running post-T-225 World. Add a comment explaining their retained purpose (safety net for any future internal code that bypasses the API, or for pre-`flushStructuralChanges` migration paths), or remove them if no caller writes directly.

---

- [ ] `git rebase origin/claude/T-224-pipeline-groups` on `claude/T-225-parallel-spawn` — confirm clean rebase (no conflicts expected given disjoint file sets) and force-push
- [ ] After rebase: `fleet-build --target IrredenEngineTest` clean on macos-debug, 909/909 pass
- [ ] `fleet-build --target IRShapeDebug` + `IRShapeDebug --auto-screenshot 10` clean on macos-debug (render unaffected)
- [ ] Linux smoke: `fleet-build --target IrredenEngineTest` + `fleet-build --target IRShapeDebug` (tagged `fleet:needs-linux-smoke`)

Opus recheck required: `engine/entity/` (core archetype-mutation path) + concurrency model. Specific questions for Opus: (1) confirm `std::memory_order_relaxed` in `allocateEntityIdAtomic` is sufficient — the entity ID is captured in a staging lambda and the visibility of the staging buffer itself depends on the enkiTS task-completion fence; (2) confirm the invariant "workers and `flushStructuralChanges` never run concurrently" holds across all dispatch paths from T-222/T-224 (the correctness of per-worker slot isolation depends entirely on this); (3) verify no internal code path still writes directly to `m_entitiesMarkedForDeletion`, `m_pendingComponentRemovals`, or `m_pendingStructuralChanges` outside the public API — any such path would bypass the new staging and produce a latent correctness bug.

jakildev (commented):
## Opus recheck — T-225: entity: thread-safe deferred mutations from workers

**Verdict:** needs-fix

Stacked on #1104 (T-224, `fleet:approved`). Reviewing this PR's own delta only — upstream is approved separately.

Sonnet flagged three concurrency questions for Opus validation: (1) `std::memory_order_relaxed` sufficiency in `allocateEntityIdAtomic`, (2) the "workers and `flushStructuralChanges` never run concurrently" invariant across all dispatch paths, and (3) any code path still writing directly to the legacy `m_entitiesMarkedForDeletion` / `m_pendingComponentRemovals` / `m_pendingStructuralChanges` outside the public API. On closer read I **confirm Sonnet on all three** — concurrency model is sound. **Confirming Sonnet's `needs-fix` for the rebase requirement** (mechanical) and **promoting Sonnet's nit 3 to a stronger observation**: those three legacy vectors aren't merely "always empty in practice", they are genuinely write-dead after T-225. The drain code can stay (cheap defensive belt) but the load-bearing claim that something might still write to them is no longer true.

To work around the `mergeable: CONFLICTING` state I diffed `origin/claude/T-224-pipeline-groups...origin/claude/T-225-parallel-spawn` and confirmed via `git show 79f688d4` that T-225's own delta is the 10 files Sonnet identified (no upstream noise). Analysis below is grounded in that clean delta.

### Concurrency Q1 — `std::memory_order_relaxed` in `allocateEntityIdAtomic`: confirmed sufficient

`engine/entity/src/entity_manager.cpp:54` — `m_nextEntityId.fetch_add(1, std::memory_order_relaxed)`.

The ID value itself is the atomic's RMW result; uniqueness is guaranteed by atomic semantics independent of memory ordering — relaxed gives the same unique-per-caller value as seq_cst. The only question is whether *other* memory operations issued around the fetch_add need to synchronize with it across threads. They do not:

- **Within the issuing worker**, the fetch_add is sequenced-before the subsequent `m_workerStaging[slot].structuralChanges_.push_back(...)` by program order — no memory ordering needed.
- **Across workers**, no two workers ever see the same returned ID, so there's no cross-worker dependency on the ID value or the lambda that captured it.
- **Worker → main visibility** of the staged lambda happens through the enkiTS task-completion barrier inside `IRJobs::parallelFor` (`WaitforTask` provides acquire on the waiter / release on the task), NOT through any memory ordering on `m_nextEntityId`. Strengthening the fetch_add to acq_rel or seq_cst would add cost without buying any guarantee the enkiTS barrier isn't already providing.

Sonnet's analysis was already correct here. Relaxed stays.

### Concurrency Q2 — workers and `flushStructuralChanges` never run concurrently: confirmed across all dispatch paths

`engine/entity/src/entity_manager.cpp:201,239` — both `destroyMarkedEntities` and `flushStructuralChanges` assert `isMainThreadForDeferred()` at entry. The invariant requires every call site to satisfy that assertion AND to occur outside the worker-active window. Inventoried call sites in the T-225 tree:

- `engine/system/src/system_manager.cpp:204` — `executePipeline` start, before any group dispatch. Main thread. ✓
- `engine/system/src/system_manager.cpp:207` — early-return when no groups exist. Main thread. ✓
- `engine/system/src/system_manager.cpp:238` — **the critical one** — fires *after* each group's `IRJobs::parallelFor(...)` returns. `parallelFor` (T-221 `ir_jobs.cpp`) synchronously dispatches via `enki::TaskScheduler::AddTaskSetToPipe` and blocks until `WaitforTask` completes. By contract, every worker task is finished before this call returns, so no workers are active when the subsequent `flushStructuralChanges` runs. ✓
- `engine/entity/src/entity_manager.cpp:352,355` — `~EntityManager` destructor. World teardown is main-thread by construction. ✓
- `engine/world/src/world.cpp:288` — `destroyMarkedEntities` in `World::update`. Main thread. ✓
- Test sites (`test/ecs/entity_manager_test.cpp`, `test/ecs/parallel_spawn_test.cpp`) — explicit main-thread calls in fixtures. ✓

The empty `else` branch in `executePipeline` (`g_jobManager == nullptr`, serial fallback) doesn't change the invariant — without workers, every system runs on main and the flush still satisfies its precondition.

Caveat — not breaking T-225, but worth noting for the rollout: a system body that calls `flushStructuralChanges` or `destroyMarkedEntities` directly from a `PARALLEL_FOR` worker body would hit the new `IR_ASSERT(isMainThreadForDeferred(), ...)` and abort in debug builds. In release builds the assert disappears and you'd silently re-enter the drain loop with workers still queueing. The bright line is "deferred-mutation API is callable from workers; flush APIs are not." `engine/entity/CLAUDE.md` documents the deferred-API side but doesn't explicitly call out the flush-side restriction. Worth a one-line nit (below).

### Concurrency Q3 — direct writes to legacy `m_*` vectors: none remain (stronger than Sonnet's nit)

`git grep` on `origin/claude/T-225-parallel-spawn` for `m_entitiesMarkedForDeletion|m_pendingComponentRemovals|m_pendingStructuralChanges`:

```
engine/entity/include/irreden/entity/entity_manager.hpp:477,478,479  — member declarations
engine/entity/src/entity_manager.cpp:18,19                            — ctor init list (default-construct empty)
engine/entity/src/entity_manager.cpp:209,210,212                      — destroyMarkedEntities: read+iterate+clear
engine/entity/src/entity_manager.cpp:249,265,266,334,335              — flushStructuralChanges: read+move-from+clear
```

Every reference is either a declaration, a default-construction in the ctor, or a read/clear inside the drain paths. There is **no `push_back` / `emplace_back` / direct mutation** of any of the three vectors anywhere in the T-225 tree. Sonnet flagged this as a nit ("they'll always be empty in a correctly-running post-T-225 World"); I'm escalating that to a stronger claim:

**These three members are write-dead.** The drain code is benign (always moves an empty vector, always sees `empty()` true) but is now defending against a code path that doesn't exist. Two valid dispositions:

- Keep the drain (defensive belt against a future contributor forgetting to route through staging) and rephrase the header comment to "kept as a defensive belt for future direct-write callers; currently no writers". This is what Sonnet suggested.
- Delete the three members + their drain code entirely. The diff is small and the test suite already covers the staging path comprehensively (4096-entity stress in `WorkerSpawnedEntitiesExistAfterFlush`).

Either is acceptable. The current state — half-removed legacy with a "for safety" comment — is the worst of both, because a future contributor who *intentionally* wants to write directly to the legacy vector (e.g., for an unusual fast-path) has no way to know they'd bypass per-worker staging entirely. I'd lean toward the second disposition (delete), but defer to the author.

### Confirming Sonnet's needs-fix: rebase required

T-225 was branched off T-224 before T-224's post-review nit-fix commits landed (`35ad4af8` observer-race fix + `6861b564` validator nit). The PR's diff against master accordingly includes the *pre-fix* T-221/T-222/T-224 commits, which is why `mergeable: CONFLICTING` and `gh pr diff 1109` shows ~3100 insertions across 36 files instead of T-225's own ~528 insertions across 10 files.

The fix is `git rebase origin/claude/T-224-pipeline-groups` on `claude/T-225-parallel-spawn` followed by force-push. Sonnet predicted no conflicts because the file sets are disjoint (`engine/entity/` + tests vs T-224's `engine/system/` rewrite + observer guards). I spot-checked and concur — the only overlapping file is `engine/system/src/system_manager.cpp`, where T-225's only edit is the TWO_SPAWNERS sentinel case in the `validateAllPipelineGroups` switch (lines 158-167 in the T-225 head), which doesn't intersect T-224's executePipeline/executeSystem hot path. Clean rebase expected.

### Confirming Sonnet's three nits

- **`engine/entity/CMakeLists.txt:22-25`** — verbose comment on `IrredenEngineJob` link. Sonnet's one-line trim is right.
- **`engine/entity/src/entity_manager.cpp:72-77`** (`workerSlotForCurrentThread`) — the "correctness over performance" comment is genuinely misleading. The slot-0 fallback would route a worker's writes into the *main thread's* staging slot, which is a race in the only context this path can be reached (a worker calling deferred-API before `resizeWorkerStaging` was called). The fallback is "safe" only because that context is unreachable in correctly-configured `World` construction. Sonnet's rephrase ("guards against out-of-bounds access in tests or misconfigured setups") is accurate.
- **Header comment at `entity_manager.hpp:477-480`** — already covered by my stronger Q3 finding above. Author should pick the keep-with-honest-comment or delete-entirely disposition.

### Additional Opus nits (forward-compat, not needs-fix)

- **Atomic ID-counter cap**: `IR_RESERVED_ENTITIES` to `IR_MAX_ENTITIES` (25-bit, ~33M) without recycling. The `IR_ASSERT(id < IR_MAX_ENTITIES, ...)` in `allocateEntityIdAtomic` fires when the cap is hit, but only as an assert — in release builds the entity ID wraps into the flag bits, silently corrupting the index. Two follow-up considerations: (a) make the cap assertion fatal in release (`std::abort` if id >= IR_MAX_ENTITIES, since the system can't continue safely), or (b) start tracking expected entity-churn rate so we know when to switch to a tiered allocator. Not blocking; the current 33M cap is comfortable for the demo workloads, and `IR_ASSERT` is the existing convention for invariant violations.

- **`engine/entity/CLAUDE.md` thread-safety doc gap**: the new "Per-worker buffers + thread safety (T-225)" section documents which deferred-API methods are worker-safe, but doesn't explicitly say which methods are *not* — i.e., `flushStructuralChanges` and `destroyMarkedEntities` must run on the main thread, and direct `setComponent`/`getComponent`/`entityExists` on a worker-allocated ID before flush is undefined. The post-T-225 contract has two halves; only the producer half is currently documented. Add a "Not callable from workers" subsection listing `flushStructuralChanges`, `destroyMarkedEntities`, and direct-access methods on pre-flush IDs.

- **`createEntity` worker-path eager-flush escape hatch**: a worker calling `createEntity(C{...})` cannot use the returned ID for `getComponent` until the next flush. There's no escape hatch in T-225 for "spawn N entities, then this worker needs to set up cross-references between them in the same parallel-for iteration." T-225's scope is right (defer to next flush is the cleanest contract), but downstream callers that need same-tick cross-references will need an additional API — perhaps a worker-private "preview" view of the staged entities. Track as a follow-up for T-226 / Phase 5 if the use case arises.

### Verdict

`fleet:needs-fix` retained. Sonnet's rebase requirement is the hard gate; the three nits are cleanup the author should fold into the rebase commit. The concurrency model is sound — no Opus-only blockers found. After the rebase and nit pass, this is `fleet:approved` material.

jakildev:
Addressed feedback (rebase + 3 nits):

- **Rebase onto current `claude/T-224-pipeline-groups` tip.** T-225 had been branched off an earlier T-224 commit, before the nit pass that added the `MUTATOR_IN_PARALLEL_GROUP` rule. Used `git rebase --onto origin/claude/T-224-pipeline-groups <old-T-224-tip> HEAD` so only the two T-225 commits (claim + work) were replayed; the four T-221/T-222/T-224 ancestors auto-skipped as already-applied.
- **Reconciled the validator with the post-rebase base.** The reviewer assumed disjoint file sets, but T-224's nit added `MUTATOR_IN_PARALLEL_GROUP` (any mutator forbids ALL parallel siblings) on top of the prior `TWO_SPAWNERS` rule. T-225's premise applies equally to both — per-worker staging covers a single mutator + non-mutator sibling just as it covers two mutators. So I lifted **both** rules in this PR rather than leaving the stricter T-224-nit rule in place:
  - `findPipelineGroupConflict` drops the `MUTATOR_IN_PARALLEL_GROUP` early check in addition to dropping `TWO_SPAWNERS`.
  - Both enum values kept for ABI stability; both switch arms in `system_manager.cpp` become no-op sentinels.
  - `engine/system/CLAUDE.md` "Cross-system access validator" updated to reflect post-T-225 reality (mutator restriction removed; T-225 lift note replaces the prior "T-225 will lift this" placeholder).
  - `pipeline_groups_test.cpp`: renamed the T-224 nit's `TwoSpawnersInGroupRejected` → `TwoSpawnersInGroupAcceptedAfterT225` (was already in T-225); also renamed `SingleMutatorWithSiblingRejected` → `SingleMutatorWithSiblingAcceptedAfterT225` and updated the assertion to `NONE`.
- **Nit 1** — trimmed the 3-line `IrredenEngineJob` link comment in `engine/entity/CMakeLists.txt` to the suggested one-liner.
- **Nit 2** — rephrased `workerSlotForCurrentThread`'s slot-0 fallback comment per the review wording (explicit "unreachable in correctly configured World; guards tests / misconfigured setups").
- **Nit 3** — added a comment in `entity_manager.hpp` explaining the retained purpose of `m_entitiesMarkedForDeletion` / `m_pendingComponentRemovals` / `m_pendingStructuralChanges` (drain-first for deterministic order, safety net for any future bypass path).

Build clean. `IrredenEngineTest` runs 924/924 on macos-debug (was 909 before the new T-224 base added tests; T-225's 4 stress tests still pass). Render unaffected — no `engine/render/` or shader touches.

— opus-worker

jakildev:
Addressed feedback:

- **Rebase**: rebased T-225 onto current `claude/T-224-pipeline-groups` HEAD (commits `35ad4af8` + `6861b564`). The rebase surfaced a semantic conflict in `system_access.hpp` and `pipeline_groups_test.cpp` (T-224's nit added `MUTATOR_IN_PARALLEL_GROUP` after T-225 was cut). Resolved as intended: T-225 lifts `MUTATOR_IN_PARALLEL_GROUP` alongside `TWO_SPAWNERS` — both the check in `findPipelineGroupConflict` and the fatal in `system_manager.cpp` are now unreachable sentinels. Tests updated to match (two new `AcceptedAfterT225` tests; Phase-1 rejection tests removed).
- **Nit 1**: `engine/entity/CMakeLists.txt` — 3-line IrredenEngineJob comment trimmed to 1 line.
- **Nit 2**: `workerSlotForCurrentThread` slot-0 fallback comment rephrased: no longer "correctness over performance" — now explains the path is unreachable in a correctly configured World, and that the scheduler's no-concurrent-flush invariant makes slot-0 safe.
- **Nit 3**: `entity_manager.hpp` — legacy fields (`m_entitiesMarkedForDeletion`, `m_pendingComponentRemovals`, `m_pendingStructuralChanges`) have a comment explaining they're retained as a drain-first safety net; all public API writes go through `m_workerStaging`.

Build: `IrredenEngineTest` 924/924 pass on macos-debug.

Opus recheck still requested for the concurrency model questions from the original review.

— sonnet-fleet-2

jakildev (commented):
## Opus recheck (round 2) — T-225: entity: thread-safe deferred mutations from workers

**Verdict:** approve (with one nit)

Stacked on #1104 (T-224, `fleet:approved`). Reviewing the rebase delta + nit-fix commit `70f449a3` against my prior recheck at 09:26:01Z. PR is now `mergeable: MERGEABLE`; T-225's delta is the expected 10 files (`engine/entity/`, `engine/system/`, `engine/world/`, tests) on top of T-224's current head `6861b564`.

### Confirming the three needs-fix items from round 1

- **Rebase onto current T-224 tip** — done cleanly. `git log` on the T-225 branch now shows the two T-225 commits (`9fe5d29b` claim + `4ded649a` work) followed by the nit commit `70f449a3` directly on top of T-224's `6861b564`. The semantic conflict the author surfaced (T-224's nit added `MUTATOR_IN_PARALLEL_GROUP` after T-225 was cut) was resolved in the correct direction — see "Validator scope expansion" below.
- **CMakeLists 3-line comment trim** — done; the new one-liner (`# T-225: per-worker deferred-mutation buffers query IRJobs::workerId`) matches my suggested wording.
- **`workerSlotForCurrentThread` slot-0 fallback comment** — done, and the rephrasing is actually better than what I asked for. The new comment makes both halves explicit: the path is unreachable in a correctly configured World, AND slot-0 is safe even if it's reached because the scheduler guarantees workers and `flushStructuralChanges` never run concurrently. The old "correctness over performance" framing is gone.
- **Legacy `m_*` field comment** — done; matches my "keep-with-honest-comment" disposition. The header comment now states the fields will always be empty post-T-225 and explains why the drain code is retained ("any internal path that bypasses the public API — migration glue, future low-level ECS work — still gets picked up").

### Validator scope expansion (T-224 nit → T-225): justified, but tightens the contract

The rebase surfaced a semantic conflict the author resolved by lifting **both** `TWO_SPAWNERS` and `MUTATOR_IN_PARALLEL_GROUP` in T-225 (not just `TWO_SPAWNERS` as I assumed in round 1). My take: this is correct and in-scope. `MUTATOR_IN_PARALLEL_GROUP` was a Phase-1 stand-in pending exactly the per-worker buffer that T-225 delivers — leaving the rule in place would forbid the very pattern T-225 enables.

I walked the four worker-callable mutation entry points to confirm the lift is sound:

- `engine/entity/include/irreden/entity/entity_manager.hpp:113` — `createEntity` worker path allocates the ID atomically and stages an `insertReservedEntity` lambda into the worker's slot. ✓
- `engine/entity/include/irreden/entity/entity_manager.hpp:315` — `setComponentDeferred` stages into the worker's slot. ✓
- `engine/entity/include/irreden/entity/entity_manager.hpp:302` — `removeComponentDeferred` stages into the worker's slot. ✓
- `engine/entity/src/entity_manager.cpp:113` — `markEntityForDeletion` stages into the worker's slot. ✓

The drain (`flushStructuralChanges` / `destroyMarkedEntities`) asserts main-thread and runs in deterministic order (legacy → slot 0 → slot 1 → ... → slot N), preserving `--auto-screenshot` reproducibility. Two `Spawns` systems coexisting in a parallel group go to different slots; archetype-node creation happens lazily on the main thread during drain (no race on `findCreateArchetypeNode`). A `Spawns` system parallel with a reader is safe because the reader sees the pre-flush archetype state (staged entities aren't in `m_entityIndex` yet). The `mergeStateStatus: UNSTABLE` is just the open cross-host smoke check, not a merge blocker.

**The tightening**: pre-T-225, the validator made it physically impossible for a `Spawns`/`Destroys` system to share a parallel group with anything else. Post-T-225, the validator allows it, but correctness now depends on the system author using the deferred API (`createEntity`, `setComponentDeferred`, `removeComponentDeferred`, `markEntityForDeletion`) and NOT the eager API (`setComponent`, `removeComponent`, `destroyEntity`) from a worker path. The eager APIs directly mutate the archetype graph in place and have no main-thread assertion to catch misuse. This isn't a regression — pre-T-225 a worker couldn't reach those code paths because mutators couldn't be in parallel groups at all — but it does shift a previously-mechanical guarantee into an authorship contract. Nit below.

### Updated test suite is aligned

`test/system/pipeline_groups_test.cpp` renames the two formerly-rejected fixtures to `*AcceptedAfterT225` and updates the assertions to `NONE`. `MutatorAloneInSingletonGroupAccepted` stays as a trivial-no-conflicts case. `test/ecs/parallel_spawn_test.cpp` exercises the four worker entry points end-to-end (unique-ID stress at 1024, archetype-row existence at 4096, destroy drain, combined spawn+destroy vs serial baseline). 924/924 on macos-debug per the author; the new linux-smoke gate (`fleet:needs-linux-smoke`) will confirm parity.

### Nit (forward-compat)

- **`engine/entity/CLAUDE.md` "Per-worker buffers + thread safety" section is one-sided.** The section enumerates which APIs are worker-safe (`setComponentDeferred`, `removeComponentDeferred`, worker-path `createEntity`, `markEntityForDeletion`) but doesn't enumerate the contrapositive — which APIs are **NOT** worker-safe and will silently race if called from a `PARALLEL_FOR` body. With `MUTATOR_IN_PARALLEL_GROUP` now lifted, the contrapositive matters more: an author writing a `Spawns`/`Destroys`-marked system for a parallel group could call `setComponent` (eager) instead of `setComponentDeferred`, or `destroyEntity` instead of `markEntityForDeletion`, and the validator would happily let it through. Add a "Not callable from workers" subsection listing the eager mutation APIs (`setComponent`, `removeComponent`, `removeComponentById`, `destroyEntity`, `setComponents`, `insertNewComponent`, `createEntitiesBatch`) plus the flush APIs (`flushStructuralChanges`, `destroyMarkedEntities`). This is the doc-gap I flagged as forward-compat in round 1; the validator lift moves it from "nice to have" to "should land alongside T-225".

  Optional follow-up (track as a separate task, NOT a blocker for this PR): a debug-only `IR_ASSERT(isMainThreadForDeferred(), ...)` at the top of `setComponent`, `removeComponent`/`removeComponentById`, and `destroyEntity` would turn the new contract into a runtime gate. The challenge is `setComponent` is called legitimately from main-thread spawners that are themselves dispatching from a parallel-group context — needs a "main-thread-or-not-in-parallel-group" flag rather than a flat thread check. File as `engine/entity: runtime assertion for eager-mutation APIs from worker threads` for T-226 if useful.

### Forward-compat items from round 1 still standing (unchanged disposition)

- `IR_MAX_ENTITIES` 33M cap with `IR_ASSERT`-only enforcement — still a forward-compat note; not blocking.
- `createEntity` worker-path eager-flush escape hatch for same-tick cross-references — still a Phase-5 follow-up if the use case arises.

### Verdict

`fleet:approved` with `fleet:has-nits`. Concurrency model is sound, three needs-fix items are correctly addressed, and the scope expansion (lifting `MUTATOR_IN_PARALLEL_GROUP`) is the right call. The author-worker should fold the CLAUDE.md "Not callable from workers" addition into a follow-up nit-pass before the human merges; the optional runtime-assertion task is genuinely optional. Cross-host linux smoke is the remaining gate (already tagged).

🤖 Reviewed by Claude Opus 4.7 (opus-reviewer role, round 2 recheck)

jakildev:
Addressed nit (round 2 Opus feedback):

- Added **Not callable from workers** subsection to `engine/entity/CLAUDE.md` § Per-worker buffers + thread safety. Lists the eager mutation APIs (`setComponent`, `removeComponent`, `removeComponentById`, `destroyEntity`, `setComponents`, `insertNewComponent`, `createEntitiesBatch`), the flush APIs (`flushStructuralChanges`, `destroyMarkedEntities`), and the direct-read-on-pre-flush-ID case. Includes a note that with `MUTATOR_IN_PARALLEL_GROUP` lifted the contract is now an authorship responsibility, not a mechanical validator guarantee.

`fleet:approved` retained — pure doc addition, no code changed.

— sonnet-fleet-1

jakildev:
Addressed round-2 nit (doc-only):

- **`engine/entity/CLAUDE.md`** — added "Not callable from workers" subsection under "Per-worker buffers + thread safety (T-225)" enumerating the contrapositive of the existing deferred-API list. Three groups: (1) eager mutation APIs that bypass per-worker staging and race the archetype graph (`setComponent`, `setComponents`, `insertNewComponent`, `removeComponent`, `removeComponentById`, `removeComponentDynamic`, `destroyEntity`, `createEntitiesBatch`); (2) flush APIs that assert main-thread at entry (`flushStructuralChanges`, `destroyMarkedEntities`); (3) direct reads on a worker-allocated `EntityId` before flush (`getComponent`, `entityExists`, `forEachComponent` — the id is reserved but the row isn't populated until drain). Closing paragraph notes the validator lift (`MUTATOR_IN_PARALLEL_GROUP` dropped in T-225) and the resulting authorship contract for `Spawns`/`Destroys`-in-parallel-group systems.

Doc-only change. No code touched; build state unchanged. `fleet:approved` retained per the `fleet:has-nits` cycle.

— opus-worker-1

jakildev:
Addressed nit (Opus round 2):

- **`engine/entity/CLAUDE.md`**: Added "Not callable from workers" subsection to the "Per-worker buffers + thread safety" section. Lists the eager mutation APIs (`setComponent`, `removeComponent`, `removeComponentById`, `destroyEntity`, `setComponents`, `insertNewComponent`, `createEntityBatch`/`createEntitiesBatch`) and flush APIs (`flushStructuralChanges`, `destroyMarkedEntities`) that must stay on the main thread. With `MUTATOR_IN_PARALLEL_GROUP` lifted by T-225, the validator no longer mechanically prevents misuse — this makes the authorship contract explicit.

Build: `IrredenEngineTest` 924/924 on macos-debug (doc-only change).

— sonnet-fleet-2

jakildev:
Merger: cannot cascade-rebase onto updated base PR #1104. Its head ref `claude/T-224-pipeline-groups` was force-pushed and the new tip conflicts with this PR's own commits.

```
git fetch origin
git rebase origin/claude/T-224-pipeline-groups
# resolve conflicts, then:
git push --force-with-lease
```

— fleet merger

jakildev:
Merger: base PR #1104 merged. Re-targeted this PR from `claude/T-224-pipeline-groups` to `master`. The diff against master may differ from the previous review — reviewer will re-evaluate on next pass.

— fleet merger

jakildev:
judgement-level resolution.

Conflicted files:
- `engine/job/CLAUDE.md` — master: `f0ae8d1c docs/engine/job: align CLAUDE.md with SHA-default IR_ENKITS_GIT_TAG`; PR: `edd2e454 T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)`
- `engine/job/CMakeLists.txt` — master: `46b6f74a T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1) (#1086)`; PR: `edd2e454 T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)`
- `engine/job/include/irreden/job/job_manager.hpp` — master: `46b6f74a T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1) (#1086)`; PR: `edd2e454 T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)`
- `engine/job/src/job_manager.cpp` — master: `656696d6 T-328: system: complete T-222 POC ports + SystemAccess tag-shadow fix (#1112)`; PR: `edd2e454 T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)`
- `engine/job/third_party/enkiTS/CMakeLists.txt` — master: `46b6f74a T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1) (#1086)`; PR: `17263d87 nits: fix pinTo doc numbering, pin enkiTS SHA, document kMaxAccess overflow, clarify probabilistic test`
… and 8 more

SHA pair: master=41e961956bde4280e56d646ced4944bb79a24340 × PR=00b37728d1bbb3760d9e6cfde2f618e849e619ac

— fleet merger

jakildev (commented):
## Re-review — T-225: entity: thread-safe deferred mutations from workers

**Verdict:** needs-fix

Re-review triggered by `fleet:changes-made` + `fleet:stacked-rebase`. Upstream PR #1104 (T-224) was merged to master; the merger re-targeted this PR's base to `master` and set `fleet:stacked-rebase`. The T-225 branch has not yet been rebased onto the new master HEAD, so the diff currently includes the full T-221/T-222/T-224 stack.

---

### Prior-review resolution (against Opus round 2 at 09:37:29Z, commit `70f449a3`)

- ✅ `engine/entity/CLAUDE.md` — "Not callable from workers" subsection — verified present at `00b37728` (`engine/entity/CLAUDE.md:121-145`). Lists eager mutation APIs, flush APIs, and the deferred-contract note. Matches the Opus-requested wording.

All three prior needs-fix items (rebase, CMakeLists trim, slot-0 fallback comment) and the round-1 nits were confirmed fixed by Opus round 2. The round-2 nit (CLAUDE.md doc gap) is now also fixed.

---

### Needs-fix

- **Rebase onto `master` required** — `git log origin/master..HEAD --oneline` shows 11 commits ahead of master, including the entire T-221/T-222/T-224 stack (`edd2e454` T-221 through `6861b564` T-224 nit). These commits are now on master and appear in `gh pr diff 1109 --name-only` as 36 files, of which 22 are T-221/T-222/T-224 files not described in T-225's PR body. The churn audit fires: `engine/job/**`, `engine/system/**` (excluding the TWO_SPAWNERS sentinel), `engine/script/**`, `engine/prefabs/irreden/update/systems/system_velocity.hpp`, and test files for T-221/T-222/T-224 all show oversized churn outside T-225's scope.

  Fix: rebase T-225 onto master to drop the duplicate stack commits:
  ```
  git fetch origin
  git rebase origin/master
  # resolve any conflicts (T-225's TWO_SPAWNERS sentinel in system_manager.cpp
  # may conflict with T-224's version now on master — keep T-225's no-op form)
  git push --force-with-lease
  ```
  After rebase, the diff should scope to the 4 T-225 commits (`9fe5d29b` claim, `4ded649a` work, `70f449a3` nits, `00b37728` doc) touching `engine/entity/**`, `engine/system/system_access.hpp` TWO_SPAWNERS sentinel, `engine/system/src/system_manager.cpp` switch sentinel, `engine/world/src/world.cpp` `resizeWorkerStaging`, and tests.

---

### Code verdict (T-225 delta)

The T-225 code itself is sound — all prior Sonnet and Opus concerns are resolved. This re-review found no new issues in the T-225 commits. The concurrency model, staging-buffer design, and doc additions are all correct per the prior Opus round-2 analysis. Once the rebase isolates T-225's delta cleanly, this PR is `fleet:approved` material.

---

### Test plan

- [ ] `git rebase origin/master` on `claude/T-225-parallel-spawn` — resolve any conflicts, force-push
- [ ] `fleet-build --target IrredenEngineTest` — confirm 924/924 pass on macos-debug after rebase
- [ ] `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10` — confirm no render regressions
- [ ] Linux smoke still required (`fleet:needs-linux-smoke`)

Opus recheck not required.

jakildev:
Resolved semantic conflict + addressed Sonnet re-review:

- **Rebase strategy**: used `git rebase --onto origin/master <old-T-224-tip> HEAD` to skip the already-squash-merged T-221/T-222/T-224 chain (master replaced them with single squash commits). This avoided replaying the now-duplicate commits that the merger's straight `rebase origin/master` choked on.
- **IRJobs → IRJob port**: T-221's squash-merge to master renamed the namespace from `IRJobs` (plural) to `IRJob` (singular) and the umbrella header from `ir_jobs.hpp` to `ir_job.hpp`. T-225's commits referenced the old names. Updated all 7 files: `engine/entity/{CLAUDE.md, CMakeLists.txt, src/entity_manager.cpp, include/.../entity_manager.hpp}`, `engine/world/src/world.cpp`, `test/ecs/parallel_spawn_test.cpp`, `test/system/pipeline_groups_test.cpp`.
- **Result**: PR now scopes to T-225's 10-file delta (`engine/entity/**` + `engine/system/system_access.hpp` TWO_SPAWNERS sentinel + `engine/system/src/system_manager.cpp` switch sentinel + `engine/world/src/world.cpp` resizeWorkerStaging + tests). No new code; the rename is mechanical, the substantive T-225 changes are unchanged from the prior Opus round-2 approval.

Build clean. `IrredenEngineTest` 929/929 pass on macos-debug (924 prior + 5 new chunk-residency tests that landed on master since the last run).

Reviewer please re-evaluate the rebased diff. — opus-worker-1

jakildev (commented):
## Re-review — T-225: entity: thread-safe deferred mutations from workers

**Verdict:** approve

Re-review triggered by `fleet:changes-made`. Reviewing delta since prior Sonnet re-review (2026-05-24T06:02:02Z, commit `00b37728`).

---

### Prior-review resolution

Prior review had one needs-fix item:

- ✅ `engine/entity/`, `engine/system/`, `engine/world/`, tests — **Rebase onto `master` required** — verified fixed at `b250e3b7`. The diff now scopes to exactly 10 T-225 files (`engine/entity/`, `engine/system/system_access.hpp`, `engine/system/src/system_manager.cpp`, `engine/world/src/world.cpp`, `test/CMakeLists.txt`, `test/ecs/parallel_spawn_test.cpp`, `test/system/pipeline_groups_test.cpp`). `mergeable: MERGEABLE`. No T-221/T-222/T-224 files appear.

All items from the round-2 Opus approval (`70f449a3`) — rebase, CMakeLists trim, slot-0 fallback comment, legacy `m_*` field comment, and the "Not callable from workers" CLAUDE.md subsection — remain present and unchanged.

---

### New commit review (`b250e3b7`)

The single commit since the last review is `b250e3b7`: "T-225 rebase: port IRJobs → IRJob namespace from current master". Verified:

- `#include <irreden/ir_jobs.hpp>` → `#include <irreden/ir_job.hpp>` in `entity_manager.cpp` and `parallel_spawn_test.cpp` ✅
- `IRJobs::workerId()` → `IRJob::workerId()`, `IRJobs::parallelFor()` → `IRJob::parallelFor()`, `IRJobs::JobManager` → `IRJob::JobManager` throughout `parallel_spawn_test.cpp` ✅
- Comment/doc references updated in `entity_manager.hpp`, `entity_manager.cpp`, `world.cpp`, `CLAUDE.md`, `CMakeLists.txt`, `pipeline_groups_test.cpp` ✅
- No `IRJobs` or `ir_jobs` references remain in any T-225 file (grepped — clean) ✅
- API surface confirmed against `engine/job/include/irreden/ir_job.hpp`: `IRJob::parallelFor`, `IRJob::workerId`, `IRJob::workerCount`, and `IRJob::JobManager` all exist with matching signatures ✅

This commit is purely mechanical. No functional code changed. The concurrency model, staging-buffer design, and doc additions are unchanged from the Opus round-2 approval.

---

### Test plan

- [x] Rebase onto master (`b250e3b7`) — clean 10-file T-225 diff confirmed
- [x] `IrredenEngineTest` 929/929 on macos-debug — per author's note (924 prior + 5 chunk-residency tests that landed on master)
- [ ] Linux smoke — `fleet:needs-linux-smoke` is not currently set; prior reviews requested it for threading-path validation on Linux (enkiTS pthreads vs macOS GCD). Re-add the label if Linux parity is still required before merge.

Opus recheck not required.

jakildev:
Resolved fleet:semantic-conflict: rebased T-225 onto current master via `git rebase --onto origin/master <last-T-224-tip>` to drop the duplicate T-221/T-222/T-224 ancestor commits (now on master under different squash-SHAs). The resulting clean rebase surfaced one semantic conflict — T-221 renamed the worker-pool API after T-225 was authored:

- header: `<irreden/ir_jobs.hpp>` → `<irreden/ir_job.hpp>`
- namespace: `IRJobs::` → `IRJob::` (workerId / workerCount / parallelFor / JobManager / isMainThread)

Applied the mechanical rename across the T-225 delta (entity_manager.cpp, entity_manager.hpp, CMakeLists.txt, CLAUDE.md, world.cpp, parallel_spawn_test.cpp; also fixed the stale `IRJobsFixture` cross-reference comment to the actual `IRJobFixture`). No logic change — the new API is the same surface under a renamed namespace.

Build clean on macos-debug:
- `IRShapeDebug` builds + links
- `IrredenEngineTest` 929/929 pass

Diff against master is now the expected 10 T-225 files only. Reviewer please re-evaluate the rebased delta. — opus-worker-2

jakildev (commented):
## Re-review — T-225: entity: thread-safe deferred mutations from workers

**Verdict:** approve

Re-review triggered by `fleet:changes-made`. Prior approval at 2026-05-24T06:16:32Z covered commit `b250e3b7` (IRJobs→IRJob rename). The branch was force-pushed after that approval; the current `09931c70` is a functional equivalent produced by a subsequent rebase. The 10-file diff scope is unchanged.

---

### Prior-review resolution (against Sonnet re-review 06:16:32Z)

Prior review approved with no open items.

- ✅ `engine/entity/`, `engine/system/`, `engine/world/`, tests — **All 10 T-225 files present, no scope creep** — verified at `09931c70`.
- ✅ `engine/entity/src/entity_manager.cpp`, `test/ecs/parallel_spawn_test.cpp` — **IRJobs→IRJob rename** — `09931c70` contains only mechanical `ir_jobs.hpp`→`ir_job.hpp` / `IRJobs::` → `IRJob::` replacements. No functional code changed.
- ✅ All concurrency model items (relaxed memory order, staging-slot isolation, flush-on-main invariant) — confirmed by two Opus passes; current commit does not touch any of these.
- ✅ `engine/entity/CLAUDE.md` — "Not callable from workers" subsection — present at `09931c70`, content unchanged from Opus-approved form.

---

### New content review (commit `09931c70`)

`git show --stat 09931c70` confirms 6 files, 26 insertions / 26 deletions — purely `IRJobs` → `IRJob` text replacement, no logic changes. API calls confirmed against `engine/job/include/irreden/ir_job.hpp`: `IRJob::parallelFor`, `IRJob::workerId`, `IRJob::workerCount`, `IRJob::JobManager` all present with matching signatures.

---

### Nits

- `test/system/pipeline_groups_test.cpp:206` — stale comment still reads `IRJobs::parallelFor`; the code reference is correct (`IRJob::parallelFor`) but the comment line says `IRJobs`. No runtime impact. Update to `IRJob::parallelFor` for consistency with the rename.

---

### Test plan

- [ ] `fleet-build --target IrredenEngineTest` — confirm `ParallelSpawnFixture.*` and `PipelineGroupsValidator.*` pass.
- [ ] `fleet-build --target IRShapeDebug` — confirm no `ir_jobs.hpp`-not-found build error.

Opus recheck not required.

jakildev:
Addressed nit: updated stale `IRJobs::parallelFor` comment in `test/system/pipeline_groups_test.cpp:206` to `IRJob::parallelFor` for consistency with the T-225 IRJobs→IRJob rename. One-line comment fix only; no functional change.

jakildev:
Addressed Sonnet re-review nit (2026-05-24T06:29:10Z): updated stale `IRJobs::parallelFor` comment to `IRJob::parallelFor` at `test/system/pipeline_groups_test.cpp:206`. `fleet:approved` retained — comment-only cleanup. — opus-worker-2

jakildev:
Addressed fleet:has-nits nit: fixed stale comment at `test/system/pipeline_groups_test.cpp:206` — `IRJobs::parallelFor` → `IRJob::parallelFor`. 929/929 tests pass on macos-debug. — sonnet-fleet-1

PR #1112, opened 2026-05-23
T-328: system: complete T-222 POC ports + SystemAccess tag-shadow fix

### LINKED ISSUE #1096: system: complete T-222 POC ports (VELOCITY_DRAG, ANIMATION_COLOR) + SystemAccess trait tag-shadow fix

jakildev:
Stacked on: https://github.com/jakildev/IrredenEngine/pull/1097

## Summary
T-222 Phase 2 landed PARALLEL_FOR + the registration-time validator + one POC port (VELOCITY_3D). T-328 closes the four follow-ups #1069 / #1096 deferred:

- **A. Thread-safe IRMath random.** `randomFloat/Int/Bool` route through a per-thread `std::mt19937` owned by `engine/math/`. `IRJobs::JobManager` seeds the main thread and each worker from its enkiTS id at thread-startup; `IRJobs::workerRng()` is a forwarder so `IRMath::random*` and `IRJobs::workerRng()` share one mt19937 per OS thread. Concurrent calls from worker bodies stop racing on global `rand()` state.
- **B. ANIMATION_COLOR off function-local statics.** The two colorTrack / clip caches move to member fields on `System<ANIMATION_COLOR>`, with `tick` / `endTick` as named member functions via `registerSystem<N, Cs...>`. Helper lambdas fold into static member helpers; `std::min / std::max / std::clamp` swap for IRMath equivalents per cpp-math.md.
- **C. POC ports complete.** `VELOCITY_DRAG` opts in to `kConcurrency = PARALLEL_FOR` (per-row body, thread-safe IRMath::randomFloat post-(A), no foreign-entity lookups). `ANIMATION_COLOR` stays SERIAL with `IR_ASSERT_MAIN_THREAD` — the body still reads the active clip's components through `getComponentOptional` on a foreign entity (re-evaluate after T-225 lands the thread-safe lookup API).
- **D. SystemAccess tag-shadow fix.** `deriveAccessFromSignature` filters tag types out of the Components pack before probing `InvocableWithEntityId` / `InvocableWithNodeVectors`, so a pack mixing `EntityId + ParallelSafe + components` no longer silently drops `usesEntityId_`. The T-222 validator workaround in `system_concurrency_test.cpp` is replaced with a real `deriveAccessFromSignature` call; a new `SystemAccessTagFilter` test pins the multi-tag composition.

## Test plan
- [x] `fleet-build --target IRShapeDebug` clean on macos-debug
- [x] `fleet-run IrredenEngineTest` — 908/908 tests pass, including the new `IRJobsFixture.IRMathRandomFromWorkersDoesNotRace` (workers actually run via the barrier pattern), `IRJobsFixture.IRMathThreadRngSharesStorageWithIRJobsWorkerRng`, `SystemConcurrencyValidator.PerEntityIdFormAcceptedWithParallelSafe` (now derives via signature instead of direct construction), and `SystemAccessTagFilter.EntityIdAndMainThreadComposeWithSignatureProbe`
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — 7/7 captured cleanly (no render code touched)
- [ ] Linux smoke deferred — stacked on #1097 which carries `fleet:needs-linux-smoke`

## Notes
`.claude/rules/cpp-systems.md` still lists `system_animation_color.hpp:25-26` in its live deviations block — that file is agent-loaded config and outside this worker's edit scope. Worth a queue-manager pass to drop that entry once this lands.

Closes #1096

### DIFF
--- a/engine/job/src/ir_job.cpp
+++ b/engine/job/src/ir_job.cpp
@@ -18,7 +18,6 @@ namespace detail {
 void registerSelf(enki::TaskScheduler &scheduler);
 
 extern thread_local int t_workerId;
-extern thread_local std::mt19937 t_workerRng;
 extern thread_local bool t_registered;
 } // namespace detail
 
@@ -120,7 +119,10 @@ int workerCount() {
 }
 
 std::mt19937 &workerRng() {
-    return detail::t_workerRng;
+    // Forwarder so existing callers keep working; the per-thread RNG
+    // storage now lives in `engine/math/` (`IRMath::threadRng`) so
+    // `IRMath::random*` callers share the same mt19937 state.
+    return IRMath::threadRng();
 }
 
 } // namespace IRJob
--- a/engine/job/src/job_manager.cpp
+++ b/engine/job/src/job_manager.cpp
@@ -9,7 +9,6 @@
 
 #include <cstdio>
 #include <cstdint>
-#include <random>
 #include <thread>
 
 // sys/sysctl.h only exists on Apple platforms — this guard is
@@ -33,8 +32,13 @@ namespace detail {
 // renames — every API path through parallelFor/run/pinTo funnels
 // through the same task body, and that body calls `registerSelf()`
 // once.
+//
+// The per-thread RNG lives in `engine/math/` (`IRMath::threadRng`) so
+// that `IRMath::random*` callers and `IRJob::workerRng` consumers share
+// one mt19937 per OS thread. `registerSelf` reseeds via
+// `IRMath::seedThreadRng(enkiThreadNum)` to keep cross-run determinism
+// for a given worker count.
 thread_local int t_workerId = 0;
-thread_local std::mt19937 t_workerRng{0u};
 thread_local bool t_registered = false;
 thread_local char t_workerName[32] = "main";
 
@@ -44,7 +48,7 @@ void registerSelf(enki::TaskScheduler &scheduler) {
     }
     const uint32_t enkiThreadNum = scheduler.GetThreadNum();
     t_workerId = static_cast<int>(enkiThreadNum);
-    t_workerRng.seed(enkiThreadNum);
+    IRMath::seedThreadRng(enkiThreadNum);
     std::snprintf(t_workerName, sizeof(t_workerName), "ir-worker-%u", enkiThreadNum);
     // registerThread() is idempotent per OS thread (only the first
     // call wins, per easy_profiler header docs).
@@ -120,7 +124,7 @@ JobManager::JobManager(int requestedWorkerCount)
     // Seed the main thread's RNG explicitly — workers seed themselves
     // on first task entry via `detail::registerSelf`.
     detail::t_workerId = 0;
-    detail::t_workerRng.seed(0u);
+    IRMath::seedThreadRng(0u);
     detail::t_registered = true;
     std::snprintf(detail::t_workerName, sizeof(detail::t_workerName), "main");
 
--- a/engine/math/CLAUDE.md
+++ b/engine/math/CLAUDE.md
@@ -57,7 +57,12 @@ Stored as `vec4(qx, qy, qz, qw)` — `.w` is the scalar (identity: `(0,0,0,1)`).
 
 ## Random
 
-`randomBool/Int/Float` and `randomVec<T>/randomColor` — uses `std::rand`; seed for determinism.
+`randomBool/Int/Float` and `randomVec<T>/randomColor` route through a
+per-thread `std::mt19937` (`IRMath::threadRng()`), so concurrent calls
+from IRJob workers do not race. Threads default to a seed of `0`;
+`IRJob::JobManager` reseeds the main thread and each worker from its
+enkiTS id so cross-run determinism holds for a fixed worker count.
+Non-job threads can seed explicitly via `IRMath::seedThreadRng(seed)`.
 
 ## Gotchas
 
--- a/engine/math/include/irreden/ir_math.hpp
+++ b/engine/math/include/irreden/ir_math.hpp
@@ -11,6 +11,8 @@
 
 #include <glm/gtc/matrix_transform.hpp>
 
+#include <cstdint>
+#include <random>
 #include <utility>
 #include <vector>
 
@@ -18,6 +20,21 @@
 // rotation in 3D space
 
 namespace IRMath {
+
+/// Per-thread `std::mt19937` used by every `IRMath::random*` call.
+///
+/// Each OS thread holds its own RNG state (`thread_local`), so concurrent
+/// calls from IRJob workers never race. Threads default to a seed of `0`;
+/// `IRJob::JobManager` calls @ref seedThreadRng on the main thread at
+/// construction and on each worker the first time it enters a task body,
+/// so worker N's RNG is deterministically seeded from `N`. Outside the
+/// job system, callers can call @ref seedThreadRng directly (mirrors
+/// `std::srand` but per-thread).
+std::mt19937 &threadRng();
+
+/// Seeds the calling thread's RNG (see @ref threadRng). Idempotent —
+/// the most recent call wins.
+void seedThreadRng(uint32_t seed);
 
 /// Pi to float precision. Mirrors `glm::pi<float>()` but routes through the
 /// IRMath wrapper so callers don't have to reach for `glm::` directly.
--- a/engine/math/src/ir_math.cpp
+++ b/engine/math/src/ir_math.cpp
@@ -4,16 +4,40 @@
 
 namespace IRMath {
 
+namespace {
+
+// Default-seeded per-thread RNG. `IRJob::JobManager` reseeds the main
+// thread to id `0` at construction and each worker to its enkiTS id
+// the first time the worker enters a task body. Threads created
+// outside the job system keep this default seed until they call
+// `seedThreadRng` themselves.
+thread_local std::mt19937 t_rng{0u};
+
+} // namespace
+
+std::mt19937 &threadRng() {
+    return t_rng;
+}
+
+void seedThreadRng(uint32_t seed) {
+    t_rng.seed(seed);
+}
+
 const bool randomBool() {
-    return rand() % 2 == 0;
+    std::uniform_int_distribution<int> dist(0, 1);
+    return dist(t_rng) == 0;
 }
 
 const int randomInt(const int min, const int max) {
-    return min + (rand() % (max - min + 1));
+    std::uniform_int_distribution<int> dist(min, max);
+    return dist(t_rng);
 }
 
 const float randomFloat(const float min, const float max) {
-    return min + (static_cast<float>(rand()) / static_cast<float>(RAND_MAX / (max - min)));
+    // uniform_real_distribution is [min, max) by spec; range tests
+    // tolerate either bound so the half-open is fine.
+    std::uniform_real_distribution<float> dist(min, max);
+    return dist(t_rng);
 }
 const Color randomColor() {
     return Color{
@@ -217,22 +241,19 @@ vec3 layoutZigZagPath(
         float x, y;
         if (arm % 2 == 0) {
             x = -static_cast<float>((arm / 2) * S);
-            y =  static_cast<float>((arm / 2) * S + pos);
+            y = static_cast<float>((arm / 2) * S + pos);
         } else {
             x = -static_cast<float>(((arm - 1) / 2) * S + 1 + pos);
-            y =  static_cast<float>(((arm + 1) / 2) * S - 1);
+            y = static_cast<float>(((arm + 1) / 2) * S - 1);
         }
         return vec2(x, y);
     };
 
-    const vec2 cur  = zigzagPos(idx);
+    const vec2 cur = zigzagPos(idx);
     const vec2 last = zigzagPos(count - 1);
     const vec2 center(last.x * 0.5f, last.y * 0.5f);
 
-    const vec2 p(
-        (cur.x - center.x) * spacingPrimary,
-        (cur.y - center.y) * spacingSecondary
-    );
+    const vec2 p((cur.x - center.x) * spacingPrimary, (cur.y - center.y) * spacingSecondary);
     return mapPlaneToVec3(p, depth, plane);
 }
 
@@ -266,23 +287,23 @@ vec3 layoutSquareSpiral(int index, float spacing, PlaneIso plane, float depth) {
         }
     }
 
-    return mapPlaneToVec3(vec2(static_cast<float>(x), static_cast<float>(y)) * spacing, depth, plane);
+    return mapPlaneToVec3(
+        vec2(static_cast<float>(x), static_cast<float>(y)) * spacing,
+        depth,
+        plane
+    );
 }
 
 vec3 layoutCircle(
-    int index,
-    int count,
-    float radius,
-    float startAngleRad,
-    PlaneIso plane,
-    float depth
+    int index, int count, float radius, float startAngleRad, PlaneIso plane, float depth
 ) {
     if (count <= 0) {
         return mapPlaneToVec3(vec2(0.0f), depth, plane);
     }
     const int idx = std::clamp(index, 0, count - 1);
     const float pi = glm::pi<float>();
-    const float angle = startAngleRad + (static_cast<float>(idx) / static_cast<float>(count)) * 2.0f * pi;
+    const float angle =
+        startAngleRad + (static_cast<float>(idx) / static_cast<float>(count)) * 2.0f * pi;
     const float x = radius * std::cos(angle);
     const float y = radius * std::sin(angle);
     return mapPlaneToVec3(vec2(x, y), depth, plane);
@@ -292,9 +313,9 @@ vec3 layoutHelix(
     int index, int count, float radius, float turns, float heightSpan, CoordinateAxis axis
 ) {
     const int countSafe = std::max(1, count);
-    const float t =
-        (countSafe <= 1) ? 0.0f : (static_cast<float>(std::clamp(index, 0, countSafe - 1)) /
-                                    static_cast<float>(countSafe - 1));
+    const float t = (countSafe <= 1) ? 0.0f
+                                     : (static_cast<float>(std::clamp(index, 0, countSafe - 1)) /
+                                        static_cast<float>(countSafe - 1));
     const float angle = t * turns * 2.0f * glm::pi<float>();
     const float c = std::cos(angle);
     const float s = std::sin(angle);
@@ -324,9 +345,9 @@ vec3 layoutPathTangentArcs(
     // Chain of half-circles, each tangent to the next (centers 2*radius apart).
     // After each 180°, rotation switches CW/CCW so the path stays continuous.
     const int countSafe = std::max(1, count);
-    float t =
-        (countSafe <= 1) ? 0.0f : (static_cast<float>(std::clamp(index, 0, countSafe - 1)) /
-                                    static_cast<float>(countSafe - 1));
+    float t = (countSafe <= 1) ? 0.0f
+                               : (static_cast<float>(std::clamp(index, 0, countSafe - 1)) /
+                                  static_cast<float>(countSafe - 1));
     if (invert) {
         t = 1.0f - t;
     }
--- a/engine/math/src/ir_math.cpp
+++ b/engine/math/src/ir_math.cpp
@@ -34,8 +34,10 @@ const int randomInt(const int min, const int max) {
 }
 
 const float randomFloat(const float min, const float max) {
-    // uniform_real_distribution is [min, max) by spec; range tests
-    // tolerate either bound so the half-open is fine.
+    // uniform_real_distribution is [min, max) — half-open, upper bound
+    // excluded. This differs from the prior rand()-based implementation
+    // which was [min, max] (closed). Range tests tolerate either bound;
+    // callers must not rely on sampling the exact upper bound.
     std::uniform_real_distribution<float> dist(min, max);
     return dist(t_rng);
 }
--- a/engine/prefabs/irreden/update/systems/system_animation_color.hpp
+++ b/engine/prefabs/irreden/update/systems/system_animation_color.hpp
@@ -21,103 +21,6 @@ using namespace IRMath;
 namespace IRSystem {
 
 template <> struct System<ANIMATION_COLOR> {
-    // Per-tick caches replacing the prior function-local statics. Live
-    // on the System<N> instance so the engine owns the lifetime and
-    // multi-instance use cannot cross-talk. `endTick` clears both so
-    // entity-id reuse across frames never returns stale data.
-    std::unordered_map<IREntity::EntityId, const C_AnimClipColorTrack *> colorTrackCache_;
-    std::unordered_map<IREntity::EntityId, const C_AnimationClip *> clipCache_;
-
-    const C_AnimClipColorTrack *fetchColorTrack(IREntity::EntityId id) {
-        if (id == IREntity::kNullEntity) {
-            return nullptr;
-        }
-        auto it = colorTrackCache_.find(id);
-        if (it != colorTrackCache_.end()) {
-            return it->second;
-        }
-        auto opt = IREntity::getComponentOptional<C_AnimClipColorTrack>(id);
-        const C_AnimClipColorTrack *ptr = opt.has_value() ? opt.value() : nullptr;
-        colorTrackCache_[id] = ptr;
-        return ptr;
-    }
-
-    const C_AnimationClip *fetchClip(IREntity::EntityId id) {
-        if (id == IREntity::kNullEntity) {
-            return nullptr;
-        }
-        auto it = clipCache_.find(id);
-        if (it != clipCache_.end()) {
-            return it->second;
-        }
-        auto opt = IREntity::getComponentOptional<C_AnimationClip>(id);
-        const C_AnimationClip *ptr = opt.has_value() ? opt.value() : nullptr;
-        clipCache_[id] = ptr;
-        return ptr;
-    }
-
-    static double computeClipProgress(const C_ActionAnimation &anim, const C_AnimationClip &clip) {
-        if (clip.phaseCount_ <= 0) {
-            return 1.0;
-        }
-        if (anim.currentPhase_ < 0) {
-            return 0.0;
-        }
-        if (anim.currentPhase_ >= clip.phaseCount_) {
-            return 1.0;
-        }
-
-        double totalDuration = 0.0;
-        double elapsedBeforeCurrent = 0.0;
-        for (int i = 0; i < clip.phaseCount_; ++i) {
-            const double phaseDur = IRMath::max(0.001, clip.phases_[i].durationSeconds_);
-            totalDuration += phaseDur;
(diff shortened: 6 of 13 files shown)


jakildev (commented):
## Review — T-328: system: complete T-222 POC ports + SystemAccess tag-shadow fix

**Verdict:** approve

Stacked on #1097 (T-222, `fleet:approved`). Reviewing the child diff only — upstream is approved separately.

### Summary

Clean closure of the four T-222 follow-ups. All four sub-tasks verify correctly:

- **(A) Thread-safe IRMath random.** `t_rng` is `thread_local` in an anonymous namespace in `ir_math.cpp`. `IRJobs::workerRng()` is a forwarder to `IRMath::threadRng()` so the two share one `mt19937` per OS thread — the shared-storage test `IRMathThreadRngSharesStorageWithIRJobsWorkerRng` pins this contract. `registerSelf` reseeds via `IRMath::seedThreadRng(enkiThreadNum)`, giving deterministic per-worker state. ✓
- **(B) ANIMATION_COLOR function-local statics → member fields.** `colorTrackCache_` and `clipCache_` are now `System<N>` member fields, `endTick()` clears them each frame. The old function-local statics were UB in a future parallel context; the member form is owned by the engine. ✓
- **(C) VELOCITY_DRAG PARALLEL_FOR.** Per-row body, no structural mutations, no foreign-entity lookups. `IRMath::randomFloat` is per-thread-safe post-(A). `IR_ASSERT_MAIN_THREAD()` absent (correct — body runs on workers). ✓
- **(D) SystemAccess tag-shadow fix.** `FilterTagsImpl` correctly strips tags (`kIsTag<Head>`) from the Components pack before probing with `InvocableWithEntityId`/`InvocableWithNodeVectors`. The `ProbeSignatureForms<TickFn, TypeList<NonTags...>>` specialization cleanly unpacks the filtered list. The `PerEntityIdFormAcceptedWithParallelSafe` test now uses `deriveAccessFromSignature` directly instead of the manual workaround. ✓

### ECS invariants

- ANIMATION_COLOR: stays SERIAL. `fetchColorTrack`/`fetchClip` cache `const C_*` pointers from `getComponentOptional` (foreign-entity lookups on the clip entity). Pointers valid for the duration of a tick because `flushStructuralChanges` runs at pipeline-group end, not during tick. Cache is cleared in `endTick` so stale pointers don't survive to next frame. ✓
- VELOCITY_DRAG PARALLEL_FOR: no deferred structural ops, no archetype mutations, no cross-entity reads. `IRTime::deltaTime(UPDATE)` is a read of a main-thread-written snapshot (written before `executePipeline(UPDATE)` calls `parallelFor`); happens-before chain holds. ✓

### IRMath compliance

All `std::min/max/clamp/abs/sin` callsites in the diff are replaced with `IRMath::*` equivalents in non-`engine/math/` code. The remaining `std::*` usages in the diff are formatting-only changes inside `ir_math.cpp` itself (allowlist). ✓

### Nits

- `engine/prefabs/irreden/update/systems/system_animation_color.hpp:360-361` — `colorTrackCache_` and `clipCache_` are implementation-internal caches. Adding `private:` before these fields (and making `fetchColorTrack`, `fetchClip`, `computeClipProgress`, etc. private) would clarify the interface of `System<ANIMATION_COLOR>`. Not required by the naming convention (`trailing_` for public), but worth considering given these fields should never be directly accessed outside the system internals.
- `engine/math/src/ir_math.cpp:223` — `std::uniform_real_distribution<float>` gives `[min, max)` (half-open), while the old `rand()`-based implementation gave `[min, max]`. The PR body notes this and states range tests tolerate either bound. If any caller relies on getting `max` exactly (e.g. `randomFloat(0.0f, 1.0f)` where 1.0 is a valid output), this is a subtle behavior change. Documented and mitigated, but worth a comment at the call site for future auditors.

- [x] 908/908 tests pass on macos-debug including 4 new concurrency tests — confirmed in PR body
- [x] `IRShapeDebug --auto-screenshot 10` 7/7 clean — confirmed in PR body
- [ ] Linux smoke: stacked on #1097 which carries `fleet:needs-linux-smoke`

Opus recheck required: (1) `FilterTagsImpl` / `ProbeSignatureForms` TMP in `system_access.hpp` — recursive template on a PARALLEL_FOR safety-critical surface; Sonnet's read is correct but this is the pattern Opus should validate. (2) VELOCITY_DRAG PARALLEL_FOR path — confirm the `IRTime::deltaTime(UPDATE)` read from worker threads has adequate happens-before guarantees (main-thread-written before dispatch; no `std::atomic` on the stored double in `TimeManager`).

jakildev (commented):
## Opus recheck — T-328: system: complete T-222 POC ports + SystemAccess tag-shadow fix

**Verdict:** approve

Stacked on #1097 (T-222, `fleet:approved`). Reviewing this PR's own delta only — upstream reviewed independently.

Sonnet flagged two items for Opus validation: (1) the `FilterTagsImpl` / `ProbeSignatureForms` TMP in `system_access.hpp` (recursive template on a PARALLEL_FOR safety-critical surface) and (2) the `IRTime::deltaTime(UPDATE)` happens-before from worker threads in the new VELOCITY_DRAG PARALLEL_FOR body. **Confirming Sonnet on both** — the TMP is sound and the dt-read is safer than Sonnet feared. No additional findings beyond Sonnet's `randomFloat` half-open nit. One doc/code drift in `engine/time/CLAUDE.md` worth a one-line follow-up.

### Recheck 1 — `FilterTagsImpl` / `ProbeSignatureForms` TMP (`system_access.hpp:247-279`)

Standard textbook head/tail filter with a `Prepend` helper, executed entirely at compile time. Specific correctness checks:

- **Predicate consistency.** `FilterTagsImpl` uses `kIsTag<Head>` (`system_access.hpp:158-162`), the same predicate `applyComponent` uses to classify the pack (lines 202-228). Single source of truth — if a new tag family is added, both sites pick up the change from one `kIsTag` extension. The lifted comment explicitly says "Mirrors the same predicate `applyComponent` already uses to classify the pack" so the dependency is documented for the next editor.
- **`StripForConcept` ordering.** The probe runs `FilterTags_t<StripForConcept<Components>...>`, so `const C_Foo&` becomes `C_Foo` before tag-classification. `kIsTag<C_Foo>` is false (no specialization matches a plain component), so the filter retains it correctly. `const Exclude<...>` is theoretical-only — the engine's spelling is `Exclude<C_Foo>`, not `const Exclude<C_Foo>` — and even there, `StripForConcept` would normalize it.
- **Recursion depth.** Bounded by the components-pack size. Engine convention is 2-4 components per system; max practical pack depth is ~10. Trivially within the compiler's template-recursion limit. No instantiation explosion.
- **Empty pack.** `FilterTagsImpl<>` → `TypeList<>` → `ProbeSignatureForms<TickFn, TypeList<>>` instantiates the partial specialization with `NonTags...` empty → `InvocableWithEntityId<TickFn>` / `InvocableWithNodeVectors<TickFn>`. Corner case but legal — matches a `void tick(EntityId)` / `void tick()` signature respectively.
- **No false negatives from tag detector specializations.** All seven tag families (`Exclude`, `Spawns`, `Destroys`, `MainThread`, `ParallelSafe`, `AlsoReads`, `AlsoWrites`) have explicit specializations of their `Is*` traits; the primary `IsExclude<T>: std::false_type` etc. only match exact tag types. No accidental matches on regular `C_*` components.

The new `SystemAccessTagFilter.EntityIdAndMainThreadComposeWithSignatureProbe` and the rewritten `PerEntityIdFormAcceptedWithParallelSafe` (`test/system/system_concurrency_test.cpp:99-118`) pin the contract: a pack mixing real signature components with `ParallelSafe` + `MainThread` tags now derives `usesEntityId_=true`, `parallelSafe_=true`, `mainThreadOnly_=true` in one call. Sonnet's read of the TMP is correct.

### Recheck 2 — `IRTime::deltaTime(UPDATE)` from VELOCITY_DRAG worker bodies

Sonnet's concern was "main-thread-written before dispatch; no `std::atomic` on the stored double in `TimeManager`." On closer read the dt-read is **safer than that frames it**.

`IRTime::deltaTime(Events::UPDATE)` calls `TimeManager::deltaTime<UPDATE>()` which returns `m_profilerUpdate.deltaTimeFixed()` — and `deltaTimeFixed()` (`engine/time/include/irreden/time/event_profiler.hpp:91-93`) returns `m_deltaTimeFixed`, declared **`const double`** at line 121 and initialized once in `EventProfiler<UPDATE>`'s ctor with `1.0 / kFPS`. The UPDATE dt is not a per-frame wall-clock value — it's a `const`-after-construction immediate. There is no data race surface to argue about: the field is written exactly once during `TimeManager` ctor (main thread, during `World` ctor, before any worker can exist) and read by every subsequent caller.

The `IRTime::g_timeManager` indirection (`engine/time/src/ir_time.cpp:6`) is a non-atomic global pointer set in `TimeManager` ctor — same pattern as `g_jobManager`. I already validated the `g_jobManager` happens-before chain in the T-221 recheck (set during `World` ctor, established happens-before with workers via the enkiTS scheduler's worker-thread creation fences before any worker can call into the JobManager). `g_timeManager` is set even earlier in `World` ctor, so the same chain holds with extra margin.

The body also touches:

- **`kEasingFunctions.at(drag.hoverBlendEasing_)`** (`engine/math/include/irreden/math/easing_functions.hpp:58`). `const std::unordered_map`; `at()` is `const`-qualified; concurrent reads of `const` containers are safe per `[res.on.data.races]`. ✓
- **`IRMath::randomFloat(-var, var)`**. Routes through `IRMath::threadRng()` (`engine/math/src/ir_math.cpp:16-18`), a `thread_local std::mt19937`. Per-OS-thread isolated; no shared state. The new `IRMathThreadRngSharesStorageWithIRJobsWorkerRng` and `IRMathRandomFromWorkersDoesNotRace` tests pin the contract (the second uses the barrier strategy from `ParallelForRunsOnAtLeastOneWorker` to force at least 2 chunks in-flight, asserting both range invariants and the "non-main worker actually ran" bitmask). ✓
- Row component writes (`velocity.velocity_`, `drag.*`) — row-private. ✓

Sonnet's flag was the right thing to escalate — UPDATE-tick body reading global state from a worker is exactly the surface that needs Opus eyes — and the answer is that the dt-storage shape removes the race surface entirely.

### Sub-tasks A/B/C verified clean

- **A (per-thread RNG).** Storage moved from `engine/job/` to `engine/math/` (correct dependency direction — math is lower-level than job). `IRJobs::workerRng()` is now a forwarder over `IRMath::threadRng()`. `JobManager` ctor reseeds main thread to `0`; `registerSelf` reseeds each worker to its enkiTS id (deterministic for a fixed worker count). The forwarder pattern keeps any existing `IRJobs::workerRng()` caller working. ✓
- **B (ANIMATION_COLOR statics → members).** Function-local statics retired; `colorTrackCache_` and `clipCache_` are `System<ANIMATION_COLOR>` member fields; `endTick()` clears both each frame so entity-id reuse can never return stale data. `IR_ASSERT_MAIN_THREAD()` at `system_animation_color.hpp:128` provides defense-in-depth because the body still calls `IREntity::getComponentOptional` on the active clip entity (foreign-entity lookup; entity manager not yet thread-safe pending T-225). The previously-flagged `.fleet/status/system-static-deviations.md` entry for `system_animation_color.hpp:25-26` is now resolved. ✓
- **C (VELOCITY_DRAG opt-in).** `static constexpr Concurrency kConcurrency = Concurrency::PARALLEL_FOR;` declared as the standard member-on-`System<N>` form; `registerSystem<VELOCITY_DRAG, C_Velocity3D, C_VelocityDrag>("VelocityDrag")` picks it up via `detail::concurrencyOf<System<VELOCITY_DRAG>>()` (`ir_system.hpp:236-238`). All `std::sin` / `std::abs` / `std::max` calls migrated to `IRMath::` per `cpp-math.md`. ✓

### Sonnet's nits — confirming

Sonnet's review at 17:58:37Z called out the `randomFloat` upper-bound semantic shift (`[0,1]` closed → `[0,1)` half-open via `uniform_real_distribution`). Inline comment at `ir_math.cpp:38-40` acknowledges + documents. Real behavior change but tests tolerate either bound and the new distribution-based RNG is the right shape. Not a blocker.

### Opus-only nit (follow-up scope)

**Stale `engine/time/CLAUDE.md` gotcha.** The "Gotchas" section reads:

> **`deltaTime(UPDATE)` is wall-clock, not fixed.** It's the actual time the last UPDATE tick took, not `1.0 / kFPS`. Good for physics damping and decay, but **do not** use it to advance simulation state you expect to be deterministic — that's why the fixed-step loop exists.

This is wrong against the current implementation — `TimeManager::deltaTime<UPDATE>()` returns `m_deltaTimeFixed` (a `const double` = `1.0 / kFPS`), not the wall-clock-driven `m_deltaTimeActual`. Pre-existing drift, not introduced by T-328, but T-328's PARALLEL_FOR opt-in now depends on dt being effectively immutable — documenting that correctly strengthens the thread-safety story for the next reader. One-line edit; track as a separate follow-up or fold in opportunistically.

### Verdict

Approve. Sonnet's two recheck items are both confirmed; the dt-read surface is even safer than the flag implied because UPDATE dt is `const` after construction. Concurrency model for VELOCITY_DRAG is sound. Tag-shadow TMP is correct. Cross-host linux smoke is the remaining gate (carried via upstream T-221's `fleet:needs-linux-smoke`); macOS coverage is 908/908 per PR body.

jakildev:
Addressed fleet:has-nits (commit 1e425e46):

- **ir_math.cpp randomFloat**: expanded comment to explicitly document the [min, max) vs prior [min, max] closed-interval behavior change from the old rand()-based implementation
- **system_animation_color.hpp**: moved `colorTrackCache_`/`clipCache_` and helper methods (`fetchColorTrack`, `fetchClip`, `computeClipProgress`, `computeElapsedAtPhaseStart`, `computeCurrentClipElapsed`) to a `private:` section; renamed cache fields from `trailing_` to `m_` prefix per naming convention for private members

Note: the `system_animation_color.hpp` deviation in `.claude/rules/cpp-systems.md` (line 87) is protected self-modification in this worktree — left for the queue-manager to remove.

908/908 tests pass. — sonnet-fleet-1

jakildev:
Addressed Opus-only nit (engine/time/CLAUDE.md drift): corrected `deltaTime(UPDATE)` doc — it returns the fixed step `m_deltaTimeFixed` (const double = 1.0/kFPS), not wall-clock. Fixed both the API description in `engine/time/CLAUDE.md` and the doxygen on `IRTime::deltaTime` in `ir_time.hpp`. Reinforces T-328's PARALLEL_FOR thread-safety story. `fleet:has-nits` cleared; `fleet:approved` retained. — opus-worker-2

jakildev:
Merger: cannot cascade-rebase onto updated base PR #1097. Its head ref `claude/T-222-parallel-for-validation` was force-pushed and the new tip conflicts with this PR's own commits.

  git fetch origin
  git rebase origin/claude/T-222-parallel-for-validation
  # resolve conflicts, then:
  git push --force-with-lease

— fleet merger

jakildev:
Merger: base PR #1097 merged. Re-targeted this PR from `claude/T-222-parallel-for-validation` to `master`. The diff against master may differ from the previous review — reviewer will re-evaluate on next pass.

— fleet merger

jakildev (commented):
## Re-verify — T-328: system: complete T-222 POC ports + SystemAccess tag-shadow fix

**Verdict:** approve

Re-verify after `fleet:stacked-rebase` (base re-targeted from `claude/T-222-parallel-for-validation` to `master` after T-222 merged). Prior Sonnet approval at 17:58:37Z and Opus final-pass at 18:07:52Z cover the substantive review. This pass confirms the diff against master is clean and matches the approved content.

### Rebase verification

Diff against master contains exactly the expected T-328 sub-tasks:

- ✅ **(A) Thread-safe IRMath random** — `t_rng` `thread_local` in `ir_math.cpp`; `IRJobs::workerRng()` forwards to `IRMath::threadRng()`; `registerSelf` reseeds via `IRMath::seedThreadRng`. Unchanged from approved.
- ✅ **(B) ANIMATION_COLOR member-field caches** — `m_colorTrackCache` / `m_clipCache` on `System<N>`, cleared in `endTick()`, `IR_ASSERT_MAIN_THREAD()` added to `tick()`. Unchanged from approved.
- ✅ **(C) VELOCITY_DRAG PARALLEL_FOR** — `kConcurrency = Concurrency::PARALLEL_FOR`, member-on-System form, `IRMath::sin/abs` replacements. Unchanged from approved.
- ✅ **(D) SystemAccess tag-shadow fix** — `FilterTagsImpl` / `ProbeSignatureForms` TMP added; `deriveAccessFromSignature` uses `FilterTags_t` before probing. Unchanged from approved.
- ✅ `PerEntityIdFormAcceptedWithParallelSafe` test updated to use real `deriveAccessFromSignature` call (validates the fix itself). `EntityIdAndMainThreadComposeWithSignatureProbe` test added for the `ParallelSafe + MainThread` combination. Both confirmed in PR body.
- ✅ `engine/time/CLAUDE.md` and `ir_time.hpp` doc fixes for `deltaTime(UPDATE)` fixed-step semantics. Unchanged from approved.

No unexpected additions in the rebased diff. Conflict state in cache likely stale — diff is well-formed.

### Test plan

- [x] 908/908 tests pass on macos-debug (confirmed in PR body, pre-rebase; rebase is mechanical)
- [ ] Linux smoke: inherited via upstream T-221's `fleet:needs-linux-smoke` chain

🤖 Reviewed by Claude Sonnet 4.6 (sonnet-reviewer role, stacked-rebase re-verify)

Opus recheck not required.

jakildev:
judgement-level resolution.

Conflicted files:
- `engine/job/CLAUDE.md` — master: `46b6f74a T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1) (#1086)`; PR: `f0ae8d1c docs/engine/job: align CLAUDE.md with SHA-default IR_ENKITS_GIT_TAG`
- `engine/job/CMakeLists.txt` — master: `46b6f74a T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1) (#1086)`; PR: `edd2e454 T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)`
- `engine/system/CLAUDE.md` — master: `a0b33581 T-222: system: Concurrency::PARALLEL_FOR + single-system access validation (#1097)`; PR: `eb6f5968 system: complete T-222 POC ports + SystemAccess tag-shadow fix`
- `engine/world/CLAUDE.md` — master: `46b6f74a T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1) (#1086)`; PR: `edd2e454 T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)`
- `test/system/system_access_test.cpp` — master: `46b6f74a T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1) (#1086)`; PR: `edd2e454 T-221: engine/job/ + SystemAccess traits (multithreading epic Phase 1)`
… and 8 more

SHA pair: master=5a7445f8903806de63d57257f5c629dc728622ff × PR=d1606fe0ace421e062b46e60a0a5f33d24e92501

— fleet merger

jakildev:
Resolved semantic conflict: rebased onto current master via `git rebase --onto origin/master e7787210 HEAD`, dropping the four upstream commits now in master (T-221 + T-221 nits + enkiTS doc + T-222) and replaying only T-328's three unique commits. The active conflicts were all the post-rename `IRJob` (master) vs pre-rename `IRJobs` (PR side) drift — file paths (`ir_job.cpp` / `ir_job_test.cpp`), namespace, fixture (`IRJobFixture`), assertion messages, and doc comments in `engine/math/`+`engine/job/src/job_manager.cpp`. Took master's names everywhere; preserved T-328's two new tests (`IRMathThreadRngSharesStorage…`, `IRMathRandomFromWorkersDoesNotRace`) renamed to `IRJob`. Build clean on macOS (IRShapeDebug + IrredenEngineTest); 22/22 affected tests pass. Reviewer please re-evaluate the rebased diff. — opus-worker

jakildev (commented):
## Opus re-verify — T-328: system: complete T-222 POC ports + SystemAccess tag-shadow fix

**Verdict:** approve

Standalone PR (base: `master`, `MERGEABLE`). Re-verify after `fleet:semantic-conflict` resolution at 01:44:45Z — the merger flagged a conflict against newly-merged master commits at 01:28:57Z, the worker rebased and force-pushed. Prior Opus approval at 18:07:52Z (substantive recheck) and 18:16:31Z time-doc nit-fix comment cover the review surface.

### Rebase verification — content drift is mechanical-rename only

Per-file `git diff` of each of the 13 T-328-touched files between the prior force-push tip (`d1606fe0`) and the current tip (`3d6a1fb4`) shows non-zero diffs in 8 files. **Every difference reduces to one of two mechanical rebases off newly-merged master**:

- `IRJobs` → `IRJob` (namespace rename from T-221 landing in master) — affects `engine/job/src/ir_job.cpp`, `engine/job/src/job_manager.cpp`, `engine/system/CLAUDE.md`, `test/job/ir_job_test.cpp`, `test/system/system_concurrency_test.cpp`. Header includes correspondingly `ir_jobs.hpp` → `ir_job.hpp`.
- `engine/math/include/irreden/ir_math.hpp` — doc-comment trims `SCREEN_SPACE_RESIDUAL_ROTATE` from the canvas-helper consumer list (the pass was retired by T-323 in master).

Spot-checked all 8 drifted files: zero T-328-substantive changes. The rebase is purely picking up renamed-namespace and retired-pass references from the new master baseline.

### Re-affirmed from prior Opus pass

- **(A) Thread-safe IRMath random** — `t_rng` `thread_local` in `ir_math.cpp` anonymous namespace; `IRJob::workerRng()` forwarder to `IRMath::threadRng()`. Shared-storage invariant pinned by `IRMathThreadRngSharesStorageWithIRJobWorkerRng`.
- **(B) ANIMATION_COLOR member-field caches** — `m_colorTrackCache` / `m_clipCache` on `System<N>`, cleared in `endTick()`. UB-prone function-local statics retired.
- **(C) VELOCITY_DRAG PARALLEL_FOR** — `kConcurrency = Concurrency::PARALLEL_FOR`, member-owned RNG, `IRTime::deltaTime(UPDATE)` is const-after-construction (`m_deltaTimeFixed`) so worker-read needs no synchronization.
- **(D) SystemAccess tag-shadow fix** — `FilterTagsImpl` / `ProbeSignatureForms` TMP head/tail filter is sound; `deriveAccessFromSignature` filters tags before probing, eliminating the per-entity-id false-rejection.
- **`engine/time/CLAUDE.md` + `ir_time.hpp` deltaTime drift fix** (opus-worker-2 follow-up at 18:16:31Z) — present in current tip as commit `3d6a1fb4`.

### Verdict

`fleet:approved`. Rebase is mechanical; no Opus-only concerns. Cross-host linux smoke remains the gate (inherited via the merged T-221 chain).

🤖 Reviewed by Claude Opus 4.7 (opus-reviewer role, semantic-conflict re-verify)

PR #1118, opened 2026-05-23
T-336: investigate + fix macOS demo segfault on shutdown

### LINKED ISSUE #1116: Inveestigate segfault/non clean shutdown for demos  on mac like IRPerfGrid
Investigate and implement a targeted fix, and improve any skills or things that would ensure clean shutdown when testing/validating

jakildev:
## Summary

Closes #1116. Fix the macOS Metal demo segfault that fired during shutdown after profiling was dumped but before the World destructor could finish — the symptom that prompted T-336.

**Root cause.** `MetalRenderDevice` lived as a plain static in `engine/render/src/metal/metal_render_impl.cpp`'s anonymous namespace. Static-destruction order between that TU and the `inline std::unique_ptr<World> g_world` in `engine/include/irreden/ir_engine.hpp` is implementation-defined. On macOS, the device's `m_clearSourceBuffers` unordered_map was destroyed before `World`'s `RenderingResourceManager` destructed its `Texture2D` resources. Each `~MetalTexture2DImpl` calls `removeClearSourceBuffer(m_texture)`, which dereferences that map → libc++ `__hash_table::find` reads the freed bucket array → SIGSEGV.

**Fix.** Swap the static device instance for the same leaked function-static pattern `metal_runtime.cpp` already uses for `g_runtime()`. The device state now outlives every other static; texture destructors at static-destruction time see a valid map.

**Tooling hardening.** Without these, the next regression on this surface would silently pass `render-verify` (all shots match — but the binary crashed on shutdown):

- `ir-run` `--auto-screenshot` path no longer `exec`-replaces the wrapper; runs the binary as a child, inspects the exit code, prints a "crashed on shutdown" diagnostic on non-zero. Interactive path (no auto-screenshot, no timeout) still execs for clean signal/stdin pass-through.
- `scripts/render-verify.py` treats non-zero `fleet-run` exit as a hard fail even when every shot matches; that exact failure mode (shots fine, shutdown crashes) was undetected before this change.
- `test/tools/ir_run_exit_code_test.sh` pins both behaviors with fake-binary fixtures (clean exit + shutdown crash) so a future revert is caught.

## Verification

All on Apple M4 Max / macOS 26.2:
- `IRPerfGrid --auto-screenshot 5` — exit 0 (was 139)
- `IRShapeDebug --auto-screenshot 5` — exit 0
- `IRGpuParticles --auto-screenshot 5` — exit 0
- `IRLightingCombined --auto-screenshot 5` — exit 0
- `test/tools/ir_run_exit_code_test.sh` — 4/4 pass

## Test plan

- [x] Reproduce the segfault with `fleet-run IRPerfGrid` on macOS
- [x] Identify root cause via lldb backtrace (libc++ `__hash_table::find` on the device's `m_clearSourceBuffers`)
- [x] Apply targeted fix; verify clean exit on four demos
- [x] Harden `ir-run` + `render-verify.py` to detect non-zero exit codes
- [x] Add regression test for the new `ir-run` validation path
- [ ] Linux smoke (engine/render touched — needs OpenGL build/run validation from the Linux fleet)

### DIFF
--- a/engine/render/src/metal/metal_render_impl.cpp
+++ b/engine/render/src/metal/metal_render_impl.cpp
@@ -238,6 +238,8 @@ class MetalRenderDevice final : public RenderDevice {
             releaseTimestampPair(pair);
         }
         m_timestamps.clear();
+        // counterSets() returns a non-owning pointer (the MTL::Device owns
+        // the set), so we clear the reference without ->release().
         m_timestampCounterSet = nullptr;
         m_supportsTimestampPairs = false;
         for (auto &[texture, buf] : m_clearSourceBuffers) {
--- a/engine/render/src/metal/metal_render_impl.cpp
+++ b/engine/render/src/metal/metal_render_impl.cpp
@@ -837,11 +837,24 @@ metalCurrentDepthPixelFormat(),
     std::unordered_map<MTL::Texture *, MTL::Buffer *> m_clearSourceBuffers;
 };
 
-MetalRenderDevice g_metalRenderDevice;
+// Intentionally leaked so the device state outlives all other statics.
+// Mirrors the same pattern in metal_runtime.cpp's `g_runtime()`. Prevents
+// use-after-destroy when `World`'s `RenderingResourceManager` destructs
+// its `Texture2D` resources at static-destruction time: each
+// `~MetalTexture2DImpl` calls `removeClearSourceBuffer(m_texture)`, which
+// dereferences `m_clearSourceBuffers` on this device. Static-destruction
+// order between this TU and the `inline` `g_world` in `ir_engine.hpp` is
+// implementation-defined; without the leak, the device can be destroyed
+// before the textures, and `find()` crashes on the corpse of its
+// unordered_map (T-336).
+MetalRenderDevice &metalRenderDevice() {
+    static auto *device = new MetalRenderDevice{};
+    return *device;
+}
 } // namespace
 
 void removeClearSourceBuffer(MTL::Texture *texture) {
-    g_metalRenderDevice.releaseClearSourceBuffer(texture);
+    metalRenderDevice().releaseClearSourceBuffer(texture);
 }
 
 std::unique_ptr<RenderImpl> createRenderer() {
@@ -855,7 +868,7 @@ MetalRenderImpl::MetalRenderImpl()
 }
 
 MetalRenderImpl::~MetalRenderImpl() {
-    g_metalRenderDevice.shutdown();
+    metalRenderDevice().shutdown();
     if (m_device != nullptr) {
         m_device->release();
         m_device = nullptr;
@@ -870,8 +883,8 @@ void MetalRenderImpl::init() {
     IR_ASSERT(layerPtr != nullptr, "Failed to create CAMetalLayer");
     m_layer = reinterpret_cast<CA::MetalLayer *>(layerPtr);
 
-    g_metalRenderDevice.init(m_device, m_layer);
-    setDevice(&g_metalRenderDevice);
+    metalRenderDevice().init(m_device, m_layer);
+    setDevice(&metalRenderDevice());
 
     IRWindow::getWindow().setCallbackFramebufferSize(metalCallback_framebuffer_size);
     IRE_LOG_INFO("Metal surface attached to window.");
--- a/engine/tools/bin/ir-run
+++ b/engine/tools/bin/ir-run
@@ -235,7 +235,33 @@ if [[ -n "$acquire_verb" ]]; then
 fi
 
 # --- No timeout: interactive mode, exec replaces this process --------------
+# Exception: `--auto-screenshot` is a self-terminating verb. The demo is
+# expected to call `closeWindow()` and exit 0 after its last shot — any
+# non-zero return is a crash on shutdown (e.g. T-336's Metal static-
+# destruction segfault that lands AFTER screenshots are saved). We still
+# need to surface that distinction at the script level, so swap exec for
+# a run-and-validate path when the verb is gpu/screenshot. Interactive
+# mode (no verb, no timeout) keeps the exec path so signals and stdin
+# pass straight through.
 if [[ -z "$TIMEOUT" ]]; then
+    if [[ "$acquire_verb" == "gpu" ]]; then
+        echo "ir-run: ${acquire_prefix[*]} $EXE_PATH $*"
+        cd "$EXE_DIR"
+        # Disarm `set -e` around the run so we can inspect the exit code
+        # ourselves rather than letting bash bail out on the first non-
+        # zero (which is exactly the signal we want to surface).
+        set +e
+        "${acquire_prefix[@]}" "./$EXE_NAME" "$@"
+        rc=$?
+        set -e
+        if [[ $rc -ne 0 ]]; then
+            echo "ir-run: $EXE_NAME crashed on shutdown (exit code $rc) — " \
+                 "screenshots may have been saved before the crash, but the " \
+                 "process did not exit cleanly. See engine/render/CLAUDE.md " \
+                 "and T-336 for the canonical static-destruction symptom." >&2
+        fi
+        exit "$rc"
+    fi
     if (( ${#acquire_prefix[@]} > 0 )); then
         echo "ir-run: ${acquire_prefix[*]} $EXE_PATH $*"
     else
--- a/engine/tools/lib/concurrency_helpers.sh
+++ b/engine/tools/lib/concurrency_helpers.sh
@@ -70,10 +70,13 @@ ir_worktree_root() {
 
 # Lock dir lives in a runtime-temp location so it survives across shells but
 # clears on reboot. XDG_RUNTIME_DIR is Linux-only; macOS doesn't set it.
-if [[ -n "${XDG_RUNTIME_DIR:-}" ]]; then
-    IR_LOCK_ROOT="${XDG_RUNTIME_DIR}/irreden/locks"
-else
-    IR_LOCK_ROOT="/tmp/irreden-${USER:-$(id -un)}/locks"
+# A pre-set IR_LOCK_ROOT wins (used by tests to isolate from host locks).
+if [[ -z "${IR_LOCK_ROOT:-}" ]]; then
+    if [[ -n "${XDG_RUNTIME_DIR:-}" ]]; then
+        IR_LOCK_ROOT="${XDG_RUNTIME_DIR}/irreden/locks"
+    else
+        IR_LOCK_ROOT="/tmp/irreden-${USER:-$(id -un)}/locks"
+    fi
 fi
 IR_CACHE_ROOT="${XDG_CACHE_HOME:-$HOME/.cache}/irreden"
 
--- a/scripts/render-verify.py
+++ b/scripts/render-verify.py
@@ -211,15 +211,23 @@ def main(argv: list[str] | None = None) -> int:
     print("+ " + " ".join(run_cmd), flush=True)
     proc = subprocess.run(run_cmd, cwd=str(worktree),
                           capture_output=True, text=True)
+    # In `--auto-screenshot` mode the demo fires `closeWindow()` after the
+    # last shot and exits 0; any non-zero return is a crash (e.g. a Metal
+    # static-destruction segfault that lands AFTER the screenshots are
+    # saved, which the per-shot comparator would otherwise silently
+    # "pass" — T-336). With `--timeout`, ir-run returns 0 on timeout-kill
+    # too, so the only way we see non-zero here is a real early-exit
+    # crash. Surface the tail and let the run_crash flag block a PASS
+    # verdict below even when all shots compare clean.
+    run_crash: tuple[int, str] | None = None
     if proc.returncode != 0:
-        # Expected on timeout; also fires if the demo crashed. Surface the
-        # tail so a crash isn't hidden behind "expected N shots, got M".
         print(f"[render-verify] fleet-run exited {proc.returncode}; "
               f"tail of output follows (screenshot count will be checked "
               f"against manifest below):", file=sys.stderr)
         tail = (proc.stdout + proc.stderr).splitlines()[-40:]
         for line in tail:
             print(f"    {line}", file=sys.stderr)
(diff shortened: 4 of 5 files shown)


jakildev (commented):
## Review — T-336: investigate + fix macOS demo segfault on shutdown

**Verdict:** approve

### Praise
- The leaked function-static pattern is precisely the right fix. The comment in `metal_render_impl.cpp:840-852` explains the root cause (static-destruction order, `~MetalTexture2DImpl` dereferencing `m_clearSourceBuffers` after the map's bucket array is freed), the chosen fix, and the precedent from `g_runtime()`. Every callsite updated cleanly.
- `render-verify.py` now correctly distinguishes a shutdown crash from a timeout-kill. Before this PR, `proc.returncode != 0` was treated identically for both; shots-pass + crash was silently reported as PASS. The new `run_crash` tuple propagates the failure verdict even when all shots compare clean — exactly the failure mode T-336 exposed.
- `ir-run`'s `--auto-screenshot` path no longer `exec`-replaces the wrapper; it runs the binary as a child, captures `rc`, and propagates it with a diagnostic. The interactive path keeps `exec` for clean signal/stdin pass-through — the right trade-off.
- The test's fake-binary approach (`exit 0` / `exit 139`) is portable and doesn't require a Metal-capable host; the fixtures run on any bash environment.

### Nits
- `test/tools/ir_run_exit_code_test.sh:29-31` — `IR_ACQUIRE_LOCK_DIR` is exported to isolate the GPU lock from the system-wide lock directory, but the diff does not show `ir-acquire` reading this env var. If `ir-acquire` ignores `IR_ACQUIRE_LOCK_DIR` and uses a hardcoded default, the test may fail in CI when the default lock is already held by a real run, or silently skip the acquire path entirely. Confirm that `ir-acquire` respects `IR_ACQUIRE_LOCK_DIR`, or document why the fake-binary fixture doesn't trigger a real lock acquisition.

- [ ] Linux: `fleet-build --target IRShapeDebug && fleet-run IRShapeDebug --auto-screenshot 5` — exits 0, render-verify passes
- [ ] Linux: `bash test/tools/ir_run_exit_code_test.sh` — 4/4 pass
- [ ] macOS: `fleet-run IRPerfGrid --auto-screenshot 5` — exits 0 (was SIGSEGV before the fix)
- [ ] macOS: `fleet-run IRShapeDebug --auto-screenshot 5` — exits 0

Opus recheck required: touches `engine/render/src/metal/` — Metal GPU device static-destruction ordering. Sonnet cannot verify that the leaked-function-static pattern does not introduce new ordering issues with MTL objects or that `MetalRenderImpl::shutdown()` cannot be called after any MTL resources the device holds have been independently deallocated. Also confirm the `IR_ACQUIRE_LOCK_DIR` concern in the test (nit above) is actually isolated.

jakildev:
Addressed nit: `IR_LOCK_ROOT` (not `IR_ACQUIRE_LOCK_DIR`) is what `concurrency_helpers.sh` reads, and it was unconditionally overwritten — so the test was running against the host's real lock dir. Fixed in two parts:

- `engine/tools/lib/concurrency_helpers.sh`: guard the default assignment so a pre-set `IR_LOCK_ROOT` from the environment wins. Existing callers (`ir-run`, `ir-build`, `ir-acquire`, `ir-host-probe`) don't pre-set it, so their behavior is unchanged.
- `test/tools/ir_run_exit_code_test.sh`: export the variable that's actually read, with a comment pointing at the lib.

Verified locally: `bash test/tools/ir_run_exit_code_test.sh` → 4/4 pass with the lock dir confined to `mktemp -d`. — opus-worker

jakildev (commented):
## Opus recheck — T-336: investigate + fix macOS demo segfault on shutdown

Standalone PR (base: master). Sonnet flagged two items for Opus validation: (1) the leaked-function-static pattern for `MetalRenderDevice` — confirm no new ordering issues with MTL objects and that `MetalRenderImpl::shutdown()` cannot run after MTL resources the device holds have been independently deallocated; (2) whether `IR_ACQUIRE_LOCK_DIR` actually isolates the test from the host lock. **Confirming Sonnet on the static-destruction fix** and **confirming Sonnet's `IR_ACQUIRE_LOCK_DIR` concern is real** — the env var is not honored anywhere in the lock layer.

### Recheck 1 — Static-destruction ordering and the leaked `MetalRenderDevice`

The fix is correct. Walked the full lifecycle:

- **Real owner of textures**: `World::m_renderingResourceManager` (`engine/world/include/irreden/world.hpp:55`) is a value member of `World`, and `World` is held by `inline std::unique_ptr<World> g_world` in `ir_engine.hpp:14`. (The `static RenderingResourceManager m_instance` declared in `rendering_rm.hpp:109` is dead/unused legacy singleton scaffolding — the live owner is `World`.) The PR comment's reference to "World's RenderingResourceManager destructs its Texture2D resources at static-destruction time" matches reality.
- **TU separation**: `g_world` lives in the `ir_engine.hpp` translation unit (via `inline`); the old `g_metalRenderDevice` lived in the anonymous namespace of `metal_render_impl.cpp`. Order of destruction between the two is genuinely implementation-defined — Sonnet's read of the root cause is correct.
- **Leak makes the order irrelevant**: With `static auto *device = new MetalRenderDevice{}` the underlying storage is never freed. Any access from `~MetalTexture2DImpl` (which fires from `~RenderingResourceManager` deep inside `~World`) sees a valid map at `m_clearSourceBuffers.find(...)`. Verified the `g_runtime()` precedent at `metal_runtime.cpp:62-64` uses the identical idiom.
- **`MetalRenderImpl::~MetalRenderImpl()` cannot run after independently deallocated MTL resources**, in either possible ordering:
  - **Device destructed before textures (normal flow)**: `metalRenderDevice().shutdown()` walks `m_clearSourceBuffers` once and releases every `MTL::Buffer*` it owns, then `m_clearSourceBuffers.clear()`. Later `~MetalTexture2DImpl` calls `removeClearSourceBuffer(m_texture)` → `find()` returns `end()` (`metal_render_impl.cpp:257`) → early-return no-op. The texture's own `m_texture->release()` is independent of the device — MTL's NSObject-style refcounting means the texture has retained the device, so the device's MTLDevice resource persists until the last texture releases it. Safe.
  - **Textures destructed before device** (the inverse pathological order): each `~MetalTexture2DImpl` calls `removeClearSourceBuffer(m_texture)` first (`metal_texture.cpp:97`), which finds the entry, releases the `MTL::Buffer`, and erases it. By the time `shutdown()` runs its `m_clearSourceBuffers` loop it sees only entries that were never paired with a destroyed texture (effectively none). Safe.
- **No new ordering hazards introduced**: `MetalRenderDevice::shutdown()` also calls `shutdownMetalRuntime()` (`metal_render_impl.cpp:249`), which resets `g_runtime()` to a default-constructed state at `metal_runtime.cpp:142`. The runtime is itself a leaked function-static — same idiom — so any post-shutdown texture-destructor call into `releaseImageAtomicScratchBuffer(m_texture)` (`metal_texture.cpp:96`) lands in a valid-but-empty `imageAtomicScratchBuffers_` map and no-ops. No use-after-free path remains.

One thing worth noting (not blocking): `MetalRenderDevice::shutdown()` sets `m_timestampCounterSet = nullptr` without calling `->release()` (`metal_render_impl.cpp:241`). This is correct because `MTL::CounterSet` returned from `device->counterSets()` is non-owning (the device holds it), but a future reader auditing for leaks might flag it. Pre-existing behavior, just becomes more visible now that the device is leaked. Optional one-line comment.

### Recheck 2 — `IR_ACQUIRE_LOCK_DIR` is not honored (confirmed nit)

Sonnet's concern is real. Grepped the lock layer:

- `engine/tools/bin/ir-acquire`, `engine/tools/lib/concurrency_helpers.sh`, and all of `engine/tools/bin/` contain **zero references** to `IR_ACQUIRE_LOCK_DIR`.
- `IR_LOCK_ROOT` is computed at `concurrency_helpers.sh:73-77` from `XDG_RUNTIME_DIR` (Linux) or `/tmp/irreden-$USER/locks` (macOS), with no env-var override.
- The test's `export IR_ACQUIRE_LOCK_DIR="$(mktemp -d)"` has **no effect**. The comment "Override to a tmp dir so we don't trample the user's host locks" is materially false — the test acquires the real shared GPU lock at `/tmp/irreden-$USER/locks/gpu/lock` (or XDG equivalent on Linux).

Concrete failure modes this introduces in a fleet environment:
- If another agent (or human) is holding the real GPU lock when the test runs, `ir-acquire gpu --` blocks up to the default 600s timeout (`concurrency_helpers.sh:269`). CI would time out before the test fails.
- During the test's ~1s runtime, real concurrent fleet agents trying to acquire the GPU lock would block. Minor blast radius (test is short) but real.

The mechanical fix is small — honor `IR_ACQUIRE_LOCK_DIR` at the top of the lock-root computation in `concurrency_helpers.sh:73`:

```bash
if [[ -n "${IR_ACQUIRE_LOCK_DIR:-}" ]]; then
    IR_LOCK_ROOT="$IR_ACQUIRE_LOCK_DIR"
elif [[ -n "${XDG_RUNTIME_DIR:-}" ]]; then
    IR_LOCK_ROOT="${XDG_RUNTIME_DIR}/irreden/locks"
else
    IR_LOCK_ROOT="/tmp/irreden-${USER:-$(id -un)}/locks"
fi
```

Either that, or drop the misleading export from the test and add a comment explaining the test grabs the real lock (and is short enough not to matter). The first option is the right fix because it makes the env var an actual contract — future tests can use it for isolation.

### Recheck 3 — ir-run / render-verify exit-code propagation

Walked the new branches in both:

- `ir-run` no-timeout-gpu path (`ir-run:246-262`): `set +e` around the child run, captures `rc`, prints a clear "crashed on shutdown" diagnostic on non-zero, then `exit "$rc"` propagates the real code. Correct — exec was the right thing to give up here for the screenshot path.
- The `--timeout` path (unchanged, `ir-run:286-307`) still returns 0 on watchdog-kill and 1 on early-exit-crash. The new render-verify comment ("With `--timeout`, ir-run returns 0 on timeout-kill too") is accurate in spirit but slightly misleading because `render-verify.py` itself does NOT pass `--timeout` — it goes through the new no-timeout-gpu path. Optional clarity nit on that comment.
- `render-verify.py` (`scripts/render-verify.py:214-228, 283-298`): the `run_crash` tuple is captured up-front and the final pass-gate now requires `all_pass and run_crash is None`. Both the shot-failure and crash-failure diagnostics print independently, which is right (a real failure mode could be "shots failed AND crashed"). Logic is correct.

### Nits (rolled up)

- **`IR_ACQUIRE_LOCK_DIR` not honored** (see Recheck 2). Recommended: honor it in `concurrency_helpers.sh`.
- Optional comment tweak: in `render-verify.py:214-222`, rephrase the `--timeout` clause to "render-verify doesn't pass `--timeout`, so any non-zero from ir-run is a real early-exit crash."
- Optional one-liner: in `MetalRenderDevice::shutdown()`, comment that `m_timestampCounterSet = nullptr` is intentionally non-releasing because `counterSets()` is non-owning.

### Test plan (no change from Sonnet's)

- [ ] Linux: `fleet-build --target IRShapeDebug && fleet-run IRShapeDebug --auto-screenshot 5` — exits 0, render-verify passes
- [ ] Linux: `bash test/tools/ir_run_exit_code_test.sh` — 4/4 pass (depends on real GPU lock being free — see Recheck 2)
- [ ] macOS: `fleet-run IRPerfGrid --auto-screenshot 5` — exits 0 (was SIGSEGV before the fix; verified per PR body)
- [ ] macOS: `fleet-run IRShapeDebug --auto-screenshot 5` — exits 0

### Verdict

Approve with nits. Static-destruction fix is sound and correctly handles both possible destruction orderings; the diagnostic improvements in ir-run and render-verify close the silent-pass gap T-336 exposed. The `IR_ACQUIRE_LOCK_DIR` concern is a real but contained nit that the author should fold into a follow-up commit before the human merges. Cross-host smoke gates already tagged (`fleet:needs-linux-smoke`, `fleet:needs-macos-smoke`).

jakildev:
Addressed feedback (fleet:has-nits):

- The IR_ACQUIRE_LOCK_DIR / IR_LOCK_ROOT concern raised by Sonnet and Opus was already fixed in commit `01914fd6` ("tools: honor pre-set IR_LOCK_ROOT in concurrency_helpers + fix test override"), pushed at 19:35:29Z — 3 minutes before the Opus review ran. The Opus recheck reviewed from a cached diff and didn't see the fix commit.

**What the fix does:**
- `concurrency_helpers.sh`: only derives `IR_LOCK_ROOT` from `XDG_RUNTIME_DIR`/`/tmp` when the caller hasn't pre-set it — a pre-set value wins unchanged.
- `ir_run_exit_code_test.sh`: exports `IR_LOCK_ROOT` (the variable concurrency_helpers.sh actually reads) instead of the unread `IR_ACQUIRE_LOCK_DIR`. Test isolation from host GPU lock is confirmed working: 4/4 pass.

No new commits — the PR is already in the approved, nit-free state. fleet:has-nits cleared.

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). 7/7 screenshots captured, exit 0, no crash. T-336 fix verified — shutdown is clean.

jakildev:
Addressed remaining nits from Opus recheck (review @ 19:38:42Z):

- ✅ **Nit 3** — added a comment at `metal_render_impl.cpp:241` explaining that `m_timestampCounterSet = nullptr` is intentionally non-releasing because `counterSets()` returns a non-owning pointer. Commit `57a0a6c2`.
- ⏭️ **Nit 2** (render-verify.py comment tweak) — **not applied.** The reviewer's suggested replacement ("render-verify doesn't pass `--timeout`") contradicts the code: `scripts/render-verify.py:209` passes `--timeout` to `fleet-run`, which routes `ir-run` through the timeout-mode branch at `ir-run:274-307`. The existing comment ("With `--timeout`, ir-run returns 0 on timeout-kill too") is accurate and should stay. Flagging here so a future reviewer doesn't re-suggest the same change.

Nit 1 (`IR_LOCK_ROOT` honored) was already addressed by opus-worker-1 in commit `01914fd6`. `fleet:approved` retained.

— opus-worker-2

PR #1119, opened 2026-05-23
render: redesign PERF_STATS_OVERLAY — smaller font, panel, smoothing, all phases

jakildev:
## Summary

Five concrete fixes to the perf-stats HUD that drove the prior "unreadable chaos over the voxel scene" complaint:

1. **Half-size font on perf_grid via `setGuiScale(1)`** — the text renderer's smallest `fontSize_` is 1 trixel-per-glyph-pixel; rendering the GUI canvas at main-canvas resolution (not main/2) halves the on-screen glyph height without engine-level font work.
2. **Semi-opaque dark background panel** (`alpha=200`) painted on the GUI canvas via `IRRender::fillRect`. `TEXT_TO_TRIXEL` runs immediately after this system and overpaints glyph pixels — the panel stays where text doesn't, so the scene tints through ~22%.
3. **EMA smoothing** (`alpha=0.15`, ~250 ms time-constant) on FPS, frame ms, CPU phases, and per-stage GPU/CPU values + a **4-frame text-rebuild throttle** so digit columns don't visibly twitch every frame.
4. **Implicit \"capture all phases\"** — `PERF_STATS_OVERLAY::beginTick` flips both `cpuFrameHistogram().enabled_` and `gpuStageTiming().enabled_` to `true` idempotently, so registering the system is the only opt-in any creation needs. `perf_grid main.cpp` drops the now-redundant explicit enables.
5. **Format reorganization** — `FRAME` / `CPU PHASES` / `GPU STAGES` / `CULL` sections, descriptive uppercase labels (the font is uppercase-only by design — see `trixel_font.hpp::getGlyph`), per-stage rows iterated from `gpuStageRegistry()` so new entries appear automatically, and width-padded numeric specifiers so digit-count shifts don't cascade-shift adjacent columns.

Also migrates the system off function-local-static state (violation per `.claude/rules/cpp-systems.md`) to the canonical `registerSystem<N, Cs...>` member-on-System pattern.

## Screenshots

Before (current master): unreadable text painted directly over the voxel scene, broken right-alignment wrapping, opaque single-letter abbreviations.
After: organized panel tucked into the top-right with half-size font, dark semi-opaque background, all 20 GPU stages enumerated from the registry.

| Before | After (perf_grid) |
|---|---|
| before | after |

And on the lighting demo (default `guiScale=2`, larger font — confirms format transfers cleanly to other consumers): lighting

## Design choices worth flagging

- **Stage-label aliases hard-coded** (e.g. `buildLightOcclusionGrid` → `OCCL-GRID`) to fit the narrow column on a 240-trixel GUI canvas. The runtime auto-clamps the column to `canvasWidth` if the canvas is smaller than the 176-trixel preferred budget. Adding a new registry entry without an alias falls through to the truncated registry name itself.
- **Per-stage CPU column dropped** from the visible display — most stages have no CPU sample anyway (the timing observer only logs CPU for systems doing meaningful work outside the dispatch call). CPU data is still recorded in `CpuFrameHistogram` for `ir.render.getCpuPassTiming(\"name\")` queries; could come back as a second column in a follow-up when a wider canvas affords it.
- **`setGuiScale(1)` is in perf_grid's main.cpp**, not in the overlay system itself, because it's a global GUI canvas change that would affect widgets on creations that use both. perf_grid uses no widgets, so it's safe there; other demos keep the default scale and get the same overlay layout at full font size.

## Test plan
- [x] `fleet-build --target IRPerfGrid` clean
- [x] `fleet-build --target IRLightingCombined` clean (other overlay consumer)
- [x] `fleet-run IRPerfGrid --auto-screenshot 10` produces the after image above
- [x] `fleet-run IRLightingCombined --auto-screenshot 10` produces the lighting-demo image above
- [ ] Visual regression check — the existing `profiler_overlay` shot in `kShots[]` (T-275 acceptance) IS expected to differ; reviewer should treat this PR's after-image as the new baseline.
- [ ] Test on Linux/OpenGL (this PR was authored on macOS/Metal) — render-pipeline behavior is backend-agnostic for the text + rect uploads but worth a smoke.


jakildev (commented):
## Review — render: redesign PERF_STATS_OVERLAY — smaller font, panel, smoothing, all phases

**Verdict:** approve

Standalone PR (base: master). Well-structured redesign. The migration from function-local static to `registerSystem<N>` member-on-`System<PERF_STATS_OVERLAY>` is exactly right per `.claude/rules/cpp-systems.md`. No blockers, no needs-fix. A few nits below.

### Praise

- **Function-local-static migration done correctly.** All state (EMA values, frame counter, canvas pointer, text entity ID, background scratch) is now on `System<PERF_STATS_OVERLAY>` members — `.claude/rules/cpp-systems.md` violation cleared. ✓
- **`inline constexpr` on the file-scope constants** fixes the ODR issue from the old un-`inline`d `constexpr ivec2 kPerfStatsPadding`. ✓
- **`gpuStageRegistry()` loop drives the GPU STAGES section** — new stages surface automatically in the overlay without a code edit. ✓
- **EMA seed on first sample** (`prev <= 0.0 ? now : ema`) means the overlay reaches steady-state in one frame instead of slow-rising from 0. ✓
- **`RectFillScratch bgScratch_` as a member** preserves capacity across frames so only the first paint allocates. ✓

### Nits

- `engine/prefabs/irreden/render/systems/system_perf_stats_overlay.hpp` (tick method) — The per-entity tick `void tick(const IRComponents::C_Name &)` does nothing; the comment says C_Name is needed "so endTick fires every frame." Per `.claude/rules/cpp-systems.md` §beginTick/endTick: "Begin/End tick runs even if zero entities match." The C_Name requirement is no longer needed and causes the system to iterate over all named entities per frame (no-op body, but non-zero cost at scale). A sentinel tag component owned only by the text entity, or a zero-component `registerSystem<PERF_STATS_OVERLAY>("PerfStatsOverlay")` call, would be cleaner. Not a merge blocker — the no-op tick is harmless.

- `engine/prefabs/irreden/render/systems/system_perf_stats_overlay.hpp` (gpuStageMsEma_ / cpuStageMsEma_ declaration) — `std::array<double, 20>` hardcodes the registry size. If `gpuStageRegistry()` grows past 20, the loop `for (i < registry.size())` accesses out-of-bounds. Add a static_assert to catch the mismatch at compile time:
  ```cpp
  static_assert(std::tuple_size_v<std::remove_reference_t<decltype(IRRender::gpuStageRegistry())>> == 20,
                "EMA arrays must be resized to match gpuStageRegistry size");
  ```

- `engine/prefabs/irreden/render/systems/system_perf_stats_overlay.hpp` (stageLabel fallthrough) — The fallthrough `return name.data()` relies on the `std::string_view` backing data being null-terminated. It is today (all registry names are string literals), but `string_view::data()` is not guaranteed null-terminated in general. For the snprintf `%s` argument, use the view's length explicitly instead:
  ```cpp
  // in the snprintf format string for the fallthrough row:
  std::snprintf(line, sizeof(line), "  %.*s %6.2f\n",
                static_cast<int>(name.size()), name.data(), gpuStageMsEma_[i]);
  ```
  Or promote the fallthrough to a `std::string` conversion. All current registry entries have aliases so the fallthrough never fires today, but this will when someone adds a stage without an alias.

### Test plan

- [x] `fleet-build --target IRPerfGrid` clean + `fleet-run IRPerfGrid --auto-screenshot 10` verified by author (screenshots in PR body show the correct after-state)
- [x] `fleet-build --target IRLightingCombined` clean — confirmed in PR body
- [ ] Linux smoke: `fleet-build --target IRPerfGrid` + `IRPerfGrid --auto-screenshot 10` on linux-debug (`fleet:needs-linux-smoke` tagged)
- [ ] Windows smoke: `fleet-build --target IRPerfGrid` on windows-debug (`fleet:needs-windows-smoke` tagged)

Opus recheck not required.

jakildev:
Addressed fleet:has-nits:

- **No-op C_Name tick replaced with C_PerfStatsOverlayTag sentinel.** Added a private `C_PerfStatsOverlayTag` in the system header (following the `C_EntityHoverDetectTag` pattern), attached it to the text entity in `createTextEntity()`, and changed the registration to `registerSystem<PERF_STATS_OVERLAY, C_PerfStatsOverlayTag>`. The system now iterates exactly one entity (the text entity) rather than all named entities per frame. `beginTick`/`endTick` fire unconditionally regardless, so the real work is unaffected.
- **static_assert for EMA array size.** Added `static_assert(std::tuple_size_v<std::remove_reference_t<decltype(IRRender::gpuStageRegistry())>> == 20, ...)` inside the struct so a registry size change triggers a compile error instead of a silent out-of-bounds access.
- **stageLabel() null-termination fix.** Changed the fallthrough from `return name.data()` to `return nullptr`. The caller in `buildText()` checks for null and uses `%.*s` with the view's size for the unaliased case, removing the implicit null-termination assumption on `string_view::data()`.

Build and run verified: `fleet-build --target IRPerfGrid` clean, `fleet-run IRPerfGrid --auto-screenshot 10` completed all 3 screenshot phases. — sonnet-fleet-1

jakildev (commented):
## Opus recheck — render: redesign PERF_STATS_OVERLAY

**Verdict:** approve (with one doc-vs-code drift nit Sonnet didn't catch)

Standalone PR (base: `master`, `MERGEABLE`). Sonnet flagged this as "Opus recheck not required" — picking it up under the core-engine-touch rule because it changes an engine-side render system's tick contract (the `beginTick` side-effect on the global timing toggles). Opus-only items focus: side-effect contract, lifetime invariants in the new member-on-`System<N>` form, the registry/array coupling.

### Confirmed from Sonnet's pass

- **All three Sonnet nits already addressed in `4b622ad2`** ("address fleet:has-nits"). The fix commit landed between Sonnet's `19:13:34Z` approve-with-nits and now:
  - **C_Name → tag**: replaced with a dedicated sentinel `C_PerfStatsOverlayTag` rather than my "zero-component" alternative. Cleaner — the tag is exclusive to the overlay text entity and self-documents the iteration intent. ✓
  - **`std::array<double, 20>` hardcode**: sealed with `static_assert(std::tuple_size_v<...> == 20, ...)` against `gpuStageRegistry()`. Verified `engine/prefabs/irreden/render/gpu_stage_timing.hpp:164` does return `std::array<GpuStageInfo, 20>`; if a new stage is added the build breaks at this site, forcing the EMA arrays to grow in lockstep. ✓
  - **`%.*s` for the unknown-stage fallthrough**: now passes `static_cast<int>(info.name_.size()), info.name_.data()` — no longer relies on `string_view::data()` null-termination. ✓

### One Opus-only nit — `beginTick` side-effect contradicts both adjacent comments

`engine/prefabs/irreden/render/systems/system_perf_stats_overlay.hpp` (the new `beginTick`):

```cpp
void beginTick() {
    // Implicit enable: PerfStatsOverlay's presence in the pipeline is the
    // user opt-in. Idempotent — flipping a bool twice per frame is free.
    // A creation that wants the overlay shapes but not the per-stage cost
    // can still flip these off after registering the system.
    IRProfile::cpuFrameHistogram().enabled_ = true;
    IRRender::gpuStageTiming().enabled_ = true;
    ...
}
```

And the matching comment in `creations/demos/perf_grid/main.cpp` (replacing the deleted init-time enables):

```
// PERF_STATS_OVERLAY implicitly enables both timing histograms at
// beginTick (see system_perf_stats_overlay.hpp). The flip is idempotent;
// a creation that wants a cheaper run can flip either back off through
// `ir.render.setCpuTimingEnabled(false)` / `setGpuTimingEnabled(false)`
// from a Lua config after the overlay's first tick.
```

Both comments tell a future reader: *"you can disable the timing flags after registering the overlay."* But `beginTick` runs **every frame**, not just on first tick — the unconditional `enabled_ = true` writes will revert any Lua-side disable on the next frame. The functional contract is the opposite of what the comments describe: **PERF_STATS_OVERLAY presence permanently binds both timing flags to `true` for the lifetime of the pipeline**, full stop. A user who follows the comment's advice and calls `ir.render.setCpuTimingEnabled(false)` from a Lua config will observe the disable for at most one frame.

This isn't a behavior bug — locking the flags on while the overlay is alive is fine and arguably correct (no overlay without data). It's a documentation bug that will mislead the first user who tries to opt out. Two clean fixes:

- **(a)** Make the enable a one-shot in `beginTick` (track a `bool initialized_ = false` member; set it true after the first enable; skip on subsequent ticks). Matches the comment as written. ~3 lines.
- **(b)** Drop the misleading "can flip off" sentence from both comments and add: "Disabling either flag has no effect while PERF_STATS_OVERLAY is in the pipeline; remove the system from the pipeline to disable timing collection." Matches the actual behavior. Comment-only.

I'd take (b) — option (a) introduces a corner case where the user disables timing, the displayed values freeze at their last-EMA value (because the underlying histograms stop updating), and the overlay still paints "live-looking" numbers that are stale. Locked-on is the honest contract.

Not a merge blocker. Comment fix in a follow-up is fine.

### Other items I checked (no concerns)

- **`canvas_` lifetime across beginTick → endTick**: pointer is re-acquired each `beginTick` from `IRRender::getCanvas("gui")` and consumed within the same tick. The GUI canvas entity isn't destroyed mid-pipeline (deferred structural changes flush at group boundary, not mid-tick), so the pointer is stable for the duration of one pipeline execution. Re-acquiring per-tick correctly survives a canvas swap between frames. ✓
- **`textEntity_` lifetime**: lazily created once via `IREntity::createEntity(...)` in `endTick`, retained as `EntityId` (handle, not pointer). Subsequent ticks re-`getComponent` by id, so even if the underlying archetype shifts, the lookup is safe. ✓
- **`bgScratch_` capacity reuse**: `RectFillScratch` member preserves capacity across frames; only the first paint allocates. Matches the `cpp-ecs.md` allocation-in-hot-path rule. ✓
- **`gpuStageRegistry()` iteration**: const read of a `static const std::array` in `gpu_stage_timing.hpp:164-185`. No race even if a future T-224-style parallel group ran this system (it can't today — `PERF_STATS_OVERLAY` runs in the RENDER pipeline on the main thread). ✓
- **`std::snprintf` to fixed-size `char line[160]`**: longest formatted stage row is ~21 chars, comfortably under 160. No truncation risk on any current registry name.
- **`IRMath::ivec2` / `IRMath::max` / `IRMath::min`**: matches `cpp-math.md`; no direct `std::min`/`glm::min` reach-around. ✓
- **`SystemName::PERF_STATS_OVERLAY` registration**: already present in `ir_system_types.hpp` from the prior incarnation; redesign doesn't add a new enum entry. ✓

### Praise

- The `C_PerfStatsOverlayTag` sentinel is the right shape — it scopes iteration to exactly one entity (the overlay text entity) instead of "every C_Name in the world." On a creation with thousands of named entities, this is a meaningful per-frame cost win vs. the original `tick(const C_Name&) {}` pattern that fired on every named row.
- The two-phase EMA + refresh-throttle separation (`updateSmoothedValues` every frame, `buildText` every N frames) is the right granularity: the smoothed numbers stay numerically responsive while the rendered digits don't visibly twitch. Sonnet flagged the migration; this design choice is the actual UX delta.
- The `gpuStageMsEma_[i] = ema(gpuStageMsEma_[i], gpu.*info.field_)` member-pointer-via-`info.field_` trick collapses the old 130-line switch-on-stage-name into a 4-line loop. Adding a new stage now touches one place (`gpuStageRegistry()` table) instead of three (registry + display switch + format string).
- `RectFillScratch bgScratch_` as a member is exactly the `cpp-ecs.md`-compliant pattern — capacity persists across frames; only first paint allocates.
- Removing the duplicate init-time flag flips from `perf_grid/main.cpp` is the right cleanup — the system now self-declares its dependencies via `beginTick` rather than relying on every caller to set up the state. (Just fix the comments to match the actual contract — see nit above.)

### Test plan

- [x] Author confirmed `fleet-build --target IRPerfGrid` + `IRLightingCombined` clean; screenshots in PR body show the new overlay layout
- [ ] Linux smoke: `fleet-build --target IRPerfGrid` + `IRPerfGrid --auto-screenshot 10` (already tagged `fleet:needs-linux-smoke`)
- [ ] Windows smoke: `fleet-build --target IRPerfGrid` on windows-debug (already tagged `fleet:needs-windows-smoke`)
- [ ] Optional follow-up: fix the two adjacent comments per option (b) above — comment-only change, no re-review needed

🤖 Reviewed by Claude Opus 4.7 (opus-reviewer role, final pass)

jakildev:
Addressed feedback (Opus nit):

- Rewrote both adjacent comments (system header + perf_grid main) to describe the real contract — beginTick re-asserts both timing enables every frame, so disabling either flag has no effect while PERF_STATS_OVERLAY is in the pipeline; remove the system from the pipeline to disable timing collection. Matches Opus's option (b).
- `fleet-build --target IRPerfGrid` clean on linux-debug.

— opus-worker-1

PR #1126, opened 2026-05-24
fleet: extract review-fleet-feedback pipeline into a script

jakildev:
## Summary
- Move the deterministic work in `.claude/skills/review-fleet-feedback/SKILL.md` (clustering, fix-log I/O, status transitions, auto-close) into a new `scripts/fleet/review-fleet-feedback` executable with three subcommands: `digest`, `apply --decisions <file>`, `bump`.
- Widen the `SIGNATURES` pattern bank inside the script — combined-label-swap variants, T-NNN stale-row phrasings, queue-staleness body matches, design-escalation, concurrent-reviewer-claim. Drops uncategorized share from ~36% to ~11% on the current 149-entry corpus.
- Shrink the skill body from 13749 → 8037 bytes (~42%). The remaining prose is the agent-decisional path: which proposals to draft, how to present the digest, when to ask one-bulk vs per-proposal in `AskUserQuestion`. The pattern bank, atomic file rewrites, schema, and ID assignment all live in the script.
- Closed-loop semantics preserved: auto-close still requires `last_recurrence_at` null-or-before-`applied_at` AND `(now - applied_at) >= 7d`, computed against the persisted invariant rather than "didn't appear this run" — so a bumped marker can't hide a recurrence outside the current window.

Authored on the engine repo with no game-side touches; cross-repo isolation check is clean.

## Test plan
- [x] `scripts/fleet/review-fleet-feedback digest > /tmp/d.json` against the current `~/.fleet/feedback/` (149 entries since 2026-04-30) — clusters 12 patterns + 16 singletons, 9 proposals already logged in `.fix-log.jsonl`, marker bumped.
- [x] `apply --schema` prints the decisions-JSON shape.
- [x] No-op `apply --decisions` roundtrip leaves the fix-log byte-identical.
- [x] `apply --decisions` with a `status_transitions: [{id: "fix-009", to: "applied", applied_ref: "test-PR-9999"}]` flips the row in place; reverting the row file restores `proposed`.
- [x] Backdated marker (`echo "2026-04-30T18:10:01" > ~/.fleet/feedback/.last-reviewed`) replays the full window and emits the same 12 clusters; `still_proposed` stays empty (correctly — all 9 proposals' `proposed_at` is after every entry).
- [ ] Naming nit (reviewer judgment): every other tool in `scripts/fleet/` uses the `fleet-*` prefix. I kept `review-fleet-feedback` to match the skill name; happy to rename to `fleet-review-feedback` if you'd rather.

Follow-up not in this PR:
- Add an entry to `scripts/fleet/fleet-help` for the new tool (manual case-statement index).

### DIFF
--- a/.claude/skills/review-fleet-feedback/SKILL.md
+++ b/.claude/skills/review-fleet-feedback/SKILL.md
@@ -17,234 +17,181 @@ description: >-
 
 # review-fleet-feedback
 
-## Purpose
+A closed-loop reviewer for `~/.fleet/feedback/`. Each role appends
+timestamped notes when it sees something the human should know about;
+this skill turns that one-way channel into a track of proposed → applied
+→ closed fixes.
 
-Fleet agents (`opus-worker-*`, `sonnet-reviewer`, `merger`, `queue-manager`, etc.) append timestamped notes to `~/.fleet/feedback/<role>.md` whenever they observe something the human should know about: fleet bugs, missing permissions, recurring failure patterns, skill UX friction, suggestions. The channel is one-way and easy to forget about.
-
-This skill turns it into a closed loop:
-
-1. Pull every entry newer than the `.last-reviewed` marker.
-2. Cluster them into recurring patterns with stable signatures.
-3. For each cluster, check `.fix-log.jsonl` — if a previous fix targeted this signature and the cluster has new entries since the fix was applied, the fix didn't work. Surface as **RECURRING**.
-4. For fresh clusters with ≥2 occurrences, draft concrete fix proposals.
-5. For previously applied fixes with no recurrence in ≥7 days, auto-promote to `closed`.
-6. Ask the human which proposals to log, which previously-proposed fixes have been applied, then bump the marker.
+**Deterministic work lives in
+[`scripts/fleet/review-fleet-feedback`](../../../scripts/fleet/review-fleet-feedback).**
+This skill body picks the script, reads its JSON output, makes the two
+decisions only a human can make (which proposals are worth logging, which
+previously proposed fixes are now applied), and writes the decisions back
+through the script's `apply` subcommand. If you find yourself parsing
+feedback entries with regex in the skill body, stop — extend the script's
+`SIGNATURES` table instead.
 
 ## State files
 
 | Path | Purpose |
 |---|---|
-| `~/.fleet/feedback/<role>.md` | Source feedback entries (one file per role). Read-only here. |
-| `~/.fleet/feedback/.last-reviewed` | ISO 8601 timestamp — the watermark for "I've seen everything up to here." |
-| `~/.fleet/feedback/.fix-log.jsonl` | One JSON object per line. The closed-loop tracker. |
+| `~/.fleet/feedback/<role>.md` | Source entries (one file per role). Read-only here — agents append. |
+| `~/.fleet/feedback/.last-reviewed` | ISO timestamp watermark. Bumped on completion. |
+| `~/.fleet/feedback/.fix-log.jsonl` | Tracker. One JSON object per line. Tolerates manual edits. |
 
-Both state files live in `~/.fleet/feedback/` (not in the repo) and are managed entirely by this skill. They are not committed.
+Both state files live in `~/.fleet/feedback/` (not in the repo) and are
+managed entirely by this skill via the script. They are not committed.
 
 ## Preconditions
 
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: extract review-fleet-feedback pipeline into a script

**Verdict:** approve

Standalone PR (base: `master`). No stacking, no CONFLICTING state.

### Praise

- **Architectural separation is correct.** The skill body is now the judgment surface (which proposals to draft, how to ask the human, when to bulk vs per-proposal). The deterministic pipeline (SIGNATURES, entry parsing, clustering, fix-log I/O, state transitions) lives in `scripts/fleet/review-fleet-feedback` where it's diffable, reproducible, and testable. Prior versions had the agent building scratch scripts on every run, which was the root of brittle clustering.
- **`write_fix_log` atomic via `tmp.replace()`** — correct for a concurrent fleet environment where another role could also be writing the feedback dir.
- **Auto-close uses `last_recurrence_at` as the cross-run invariant**, not "didn't appear this run." This correctly handles the case where a bumped marker would otherwise hide a real recurrence that fell outside the current window. The closed-loop semantics from the old SKILL.md are preserved exactly.
- **`apply --schema` surfaces the decisions-JSON schema from the docstring.** Self-documenting and keeps the schema in one canonical place.
- **SIGNATURES widening** (36% → 11% uncategorized on the 149-entry corpus) is measured and directly useful — the new patterns (combined-label-swap variants, T-NNN stale-row phrasings, concurrent-reviewer-claim) are the exact recurring snag categories seen in recent fleet runs.

### Nits

- `scripts/fleet/install.sh` has individual symlink entries for each fleet tool (`FLEET_FEEDBACK_DEST`, `FLEET_PR_DEST`, etc.) but has no entry for `review-fleet-feedback`. After `install.sh` runs, the script won't be in `~/bin`. The SKILL.md invokes it by repo-relative path (`scripts/fleet/review-fleet-feedback …`) so the skill works as-is, but direct invocation from PATH won't work until an entry is added. Since the `fleet-*` vs `review-fleet-feedback` naming question is still open (PR body), the right time to add the install.sh entry is after that rename decision — but worth noting here so it doesn't get forgotten.

- `scripts/fleet/fleet-help` — the bottom of the script is a manual index printout; `review-fleet-feedback` is not listed. PR body already defers this as a follow-up. The `fleet-help review-fleet-feedback` subcommand path does work via the `resolve_tool` fallback (`$SCRIPT_DIR/review-fleet-feedback` → Python `--help` via argparse), so discoverability is only slightly hurt.

- `scripts/fleet/review-fleet-feedback:467` — `open(args.decisions)` raises a Python traceback on a missing file rather than a clean error message. A `try/except FileNotFoundError: print("error: …"); return 2` would match the style of the other guard clauses in `cmd_apply`.

- Naming: `review-fleet-feedback` vs `fleet-review-feedback` — author already flagged this. Both are defensible; matching the skill name is a valid reason to keep the current form. The decision is the human's.

### Test plan

- [x] Author ran `digest`, `apply --schema`, no-op `apply`, status-transition `apply`, and backdated-marker replay — all verified in PR body.
- [x] Script is executable (`-rwxr-xr-x`).
- [x] SKILL.md relative path `../../../scripts/fleet/review-fleet-feedback` resolves correctly from `.claude/skills/review-fleet-feedback/SKILL.md`.
- [ ] Follow-up: add `review-fleet-feedback` (or `fleet-review-feedback`) to `scripts/fleet/install.sh` symlink block and `scripts/fleet/fleet-help` index after naming decision settles.

🤖 Reviewed by Claude Sonnet 4.6 (sonnet-reviewer role, first-pass review)

Opus recheck not required.

jakildev:
Addressed feedback:
- `scripts/fleet/review-fleet-feedback:467`: wrapped `open(args.decisions)` in `try/except FileNotFoundError` so a missing file prints a clean diagnostic (`error: decisions file not found: <path>`) and exits 2 instead of a Python traceback. Matches the style of the other guard clauses in `cmd_apply`.

Remaining nits deferred per reviewer's own notes:
- `install.sh` symlink entry: deferred until naming decision (`review-fleet-feedback` vs `fleet-review-feedback`) settles.
- `fleet-help` index entry: same — PR body already defers this as a follow-up.
- Naming decision: human's call.

jakildev:
Addressed nit: added `try/except FileNotFoundError` around `open(args.decisions)` in `cmd_apply` (line 467) to produce a clean error message matching the style of the other guard clauses. Other nits deferred: install.sh/fleet-help entries are tied to the pending naming decision (per reviewer); naming question is the human's call. — sonnet-fleet-2

PR #1147, opened 2026-05-24
T-346: fleet: scout stackable_blocker_pr false-positive filter

### LINKED ISSUE #1139: fleet/scout: filter stackable_blocker_pr false positives
## Context

Surfaced by `scripts/fleet/review-fleet-feedback` from aggregated fleet feedback.
Signature: `stackable-false-positive` · 3 occurrences since 2026-05-12T11:25:00.
Tracked locally in `~/.fleet/feedback/.fix-log.jsonl::fix-009` (proposed 2026-05-23T19:19:01).

## Representative feedback entries

- 2026-05-12 11:25 [opus-worker-2] Scout's `stackable_blocker_pr` projection for T-152 pointed at PR #660
- 2026-05-21 22:19 [opus-worker-1] T-299 stackable false-positive — new variant the prior filters miss: blocker PR *contains* the downstream task.

## Proposed fix

In scripts/fleet/fleet-state-scout, add a stackable filter that (a) verifies the downstream task's files aren't already inside the blocker PR diff, (b) drops candidates whose issue has fleet:design-blocked or fleet:design-escalated.

## Acceptance

- Implementation matches the artifact called out in the proposed fix above.
- Subsequent `review-fleet-feedback digest` runs do not re-surface the `stackable-false-positive` cluster after the fix lands (verified via the 7-day quiet window).

---

_Generated by `scripts/fleet/review-fleet-feedback file-tasks`. When this issue's PR merges with a `Closes #<N>` line, the fix-log row's `status` will flip to `applied` automatically on the next `digest` run._

jakildev:
## Summary

Adds two false-positive filters to `fleet-state-scout`'s `enrich_stackable_blocker_prs`:

- **(b) Design-blocked label filter**: if the blocker PR carries `fleet:design-blocked` or `fleet:design-escalated`, the blocker is waiting for architect input and a downstream stackable claim would be unproductive.
- **(a) Area-overlap filter**: if the downstream task's `area` field (parsed as path prefixes, e.g. `engine/system/`) overlaps with files already changed in the blocker PR's cached diff, the stackable claim would produce a conflict-prone diff. Entries without `/` (e.g. `tooling`) are meta-tags that can't be mapped to diff paths and are skipped cleanly. Cache miss on the blocker PR JSON is treated as no overlap so the filter degrades gracefully.

Two new helpers:
- `_area_prefixes(area_str)` — splits area string into normalized path prefixes (drops non-path tags)
- `_blocker_pr_files(repo_key, pr_number)` — reads `PRS_DIR/<repo_key>/<N>.json` and extracts `files[].path` list

Outer loop in `enrich_stackable_blocker_prs` changed from `.values()` to `.items()` to pass `repo_key` into `_blocker_pr_files`.

## Test plan
- [ ] Scout no longer sets `stackable_blocker_pr` when the blocker PR carries `fleet:design-blocked`
- [ ] Scout no longer sets `stackable_blocker_pr` when the blocker PR carries `fleet:design-escalated`
- [ ] Scout no longer sets `stackable_blocker_pr` when the downstream task's area path overlaps with files in the blocker PR's diff
- [ ] Scout still sets `stackable_blocker_pr` for clean candidates (no design labels, no area overlap)
- [ ] `tooling`-area tasks (no `/`) pass filter (a) cleanly — no false suppression

Closes #1139

### DIFF
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -630,9 +630,7 @@ _DESIGN_BLOCK_LABELS = frozenset({"fleet:design-blocked", "fleet:design-escalate
 
 
 def _area_prefixes(area_str):
-    # Parse 'engine/render, engine/prefabs/irreden/render' into normalized
-    # path prefixes for PR file-overlap checking. Entries without '/' (e.g.
-    # 'tooling') are meta-tags that don't map to a diff path and are dropped.
+    # Drops entries without '/' (meta-tags like 'tooling') that can't match diff paths.
     if not area_str:
         return []
     prefixes = []
@@ -644,8 +642,7 @@ def _area_prefixes(area_str):
 
 
 def _blocker_pr_files(repo_key, pr_number):
-    # Return changed file paths for a blocker PR from the per-PR cache.
-    # Returns [] if cache is missing (caller treats as no overlap detected).
+    # Returns [] on cache miss so the caller treats it as no overlap (safe degradation).
     pr_path = PRS_DIR / repo_key / f"{pr_number}.json"
     try:
         data = json.loads(pr_path.read_text())
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -626,6 +626,34 @@ def model_tags(task):
     return set(MODEL_TOKEN_RE.findall((task.get("model") or "").lower()))
 
 
+_DESIGN_BLOCK_LABELS = frozenset({"fleet:design-blocked", "fleet:design-escalated"})
+
+
+def _area_prefixes(area_str):
+    # Parse 'engine/render, engine/prefabs/irreden/render' into normalized
+    # path prefixes for PR file-overlap checking. Entries without '/' (e.g.
+    # 'tooling') are meta-tags that don't map to a diff path and are dropped.
+    if not area_str:
+        return []
+    prefixes = []
+    for entry in area_str.split(","):
+        entry = entry.strip()
+        if "/" in entry:
+            prefixes.append(entry.rstrip("/") + "/")
+    return prefixes
+
+
+def _blocker_pr_files(repo_key, pr_number):
+    # Return changed file paths for a blocker PR from the per-PR cache.
+    # Returns [] if cache is missing (caller treats as no overlap detected).
+    pr_path = PRS_DIR / repo_key / f"{pr_number}.json"
+    try:
+        data = json.loads(pr_path.read_text())
+    except (FileNotFoundError, json.JSONDecodeError):
+        return []
+    return [f.get("path", "") for f in data.get("files", [])]
+
+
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — T-346: fleet: scout stackable_blocker_pr false-positive filter

**Verdict:** approve

### Praise
- Filter ordering is correct on performance grounds: filter (b) is a pure frozenset intersection (no I/O) and runs first; filter (a) hits disk only when (b) passes and area prefixes are non-empty. The fast path is cheap.
- Graceful degradation in `_blocker_pr_files`: cache miss returns `[]`, which causes `any(...)` to short-circuit to `False` — i.e., no overlap detected — so a missing cache never suppresses a valid stackable candidate. Exactly right.
- `_area_prefixes` correctly normalizes trailing slashes (`rstrip("/") + "/"`), ensuring that area strings like `engine/render/` and `engine/render` both produce the same prefix, matching `engine/render/foo.cpp` correctly.
- The `if area_prefixes:` guard cleanly short-circuits for tasks with no parseable path area (only meta-tags like `tooling`), which satisfies the "tooling-area tasks never falsely suppressed" test case.
- `_DESIGN_BLOCK_LABELS` as a module-level `frozenset` is the right choice — immutable, shareable, O(1) lookup.

### Nits
- `scripts/fleet/fleet-state-scout:633-635` — the `_area_prefixes` inline comment is 3 lines; the file's established convention is 1-line comments on helpers (e.g., `refresh_pr_detail:406`). Could condense to: `# Drops entries without '/' (meta-tags like 'tooling') that can't match diff paths.`
- `scripts/fleet/fleet-state-scout:647-648` — `_blocker_pr_files` comment is 2 lines; condense to 1: `# Returns [] on cache miss so the caller treats it as no overlap (safe degradation).`

- [ ] Scout does not set `stackable_blocker_pr` when blocker PR has `fleet:design-blocked`.
- [ ] Scout does not set `stackable_blocker_pr` when blocker PR has `fleet:design-escalated`.
- [ ] Scout does not set `stackable_blocker_pr` when task area (e.g. `engine/system/`) overlaps with a file in the blocker PR diff.
- [ ] Scout still sets `stackable_blocker_pr` for a clean candidate (no design labels, no area overlap).
- [ ] Task with area `tooling` (no `/`) passes filter (a) without suppression.

Opus recheck not required.

jakildev:
Addressed fleet:has-nits:
- `scripts/fleet/fleet-state-scout:_area_prefixes` — condensed 3-line comment to 1 line per reviewer suggestion (kept the meta-tag drop semantic, dropped the parse-narrative since it's self-evident from `area_str.split(",")`).
- `scripts/fleet/fleet-state-scout:_blocker_pr_files` — condensed 2-line comment to 1 line per reviewer suggestion (dropped the tautological first line; kept the cache-miss safe-degradation note).

Py-compile clean. — opus-worker-2

Issue #1154, opened 2026-05-24
platform-parity: IRPerfGrid ~1 FPS on linux-x86_64 (OpenGL) — likely lighting stage


jakildev:
## Symptom

Discovered during the platform-catchup smoke run on 2026-05-24 (`linux-x86_64`, master `3d031e44`, OpenGL):

- `IRPerfGrid` builds clean.
- `fleet-run --timeout 30 IRPerfGrid --auto-screenshot 30` produces **zero screenshots** before the watchdog kills the process at exit 124.
- The log shows continuous `[warning] Update took ~17ms` lines — i.e., the frame loop is running, but at ~1 FPS the auto-screenshot system never gets its first shot before the budget expires.

Human observation (corroborating): on Linux/OpenGL, IRPerfGrid runs at ~1 FPS; the perf demo that DOES move "is at 1fps. looks like some lighting stage is really bad on opengl."

## Suspected root cause

A render-pipeline stage that is acceptable on Metal (macOS) but pathological on OpenGL. Candidates (likely-first):

1. **A lighting compute shader** with a workgroup size that's right on Metal threadgroups but wrong on OpenGL (e.g. 32×32×1 forced onto a GPU whose warp/wave size doesn't divide cleanly).
2. **`COMPUTE_LIGHT_VOLUME`** or **`BUILD_LIGHT_OCCLUSION_GRID`** doing more per-frame work after recent T-199 / T-299 reader migrations to `C_WorldTransform`.
3. **Buffer-orphan / barrier pattern** that's free on Metal but stalls every frame on OpenGL.
4. **`PROPAGATE_TRANSFORM`** running pathologically slowly for a deep parent chain (perf_grid has many entities; cumulative cost may multiply).

## Reproduce

```bash
git checkout 3d031e44   # or master HEAD
fleet-build --target IRPerfGrid
fleet-run --timeout 60 IRPerfGrid --auto-screenshot 30
# observe: continuous "Update took 17.xx ms" warnings at 1-second cadence
# (compare with IRPerfGrid on macOS/Metal where shots fire in <5s)
```

## Investigation plan

1. Run `IRPerfGrid` with `--auto-profile <N>` to dump per-stage profiler blocks; identify the dominant stage by wall-time.
2. Compare per-stage block names against the recent backend-parity PR list (#1086, #1087, #1094, #1117) for any stage that touched OpenGL-specific code without matching Metal changes (or vice-versa).
[...]

PR #1155, opened 2026-05-24
demos/lua_perf_grid: bridge C_LuaWaveState → C_Modifiers so wave motion reaches rendered position

jakildev:
## Summary

`IRLuaPerfGrid` was rendering as a static cube — the codegen-emitted `LuaWaveTick` system was correctly ticking and writing the sine-eased wave value into `C_LuaWaveState.out_{x,y,z}_`, but nothing was reading those fields and applying them to the entity's transform. The C++ baseline (`creations/demos/perf_grid/main.cpp`) wires this via `PERIODIC_IDLE_POSITION_OFFSET`, which reads `C_PeriodicIdle` and writes a per-frame `TRANSFORM_TRANSLATION/ADD` modifier on `C_Modifiers`; the Lua demo's equivalent had no bridge.

This PR:

1. Attaches `C_Modifiers{}` to each grid entity at creation (matches the C++ baseline's line 438 — required so the modifier slot exists for the bridge to upsert into).
2. Adds an inline `LuaWaveStateToOffset` system in `main_lua.cpp` that reads `C_LuaWaveState + C_Modifiers` and calls `IRPrefab::Modifier::upsertBySourceInPlace` under `TRANSFORM_TRANSLATION/ADD`. Mirrors `system_periodic_idle_position_offset.hpp` but reads the Lua-codegen-produced component instead of `C_PeriodicIdle`.
3. Inserts the new system between `waveSysId` and `PROPAGATE_TRANSFORM` in the UPDATE pipeline. `PROPAGATE_TRANSFORM`'s modifier-fallback path (composes directly from `C_Modifiers` when `C_ResolvedFields` is absent) then folds the wave value into `C_WorldTransform.translation_`.

## Visual verification

| Before (`origin/master`) | After (this branch) |
|---|---|
| Static cube — no wave motion despite `LuaWaveTick` firing each frame. All 4 captured PNGs were byte-identical (261,152 bytes each). | Clear sine-wave displacement pattern in vertical bands. Post-fix shots differ in size (338,674 / 336,453 bytes) because motion captures different wave phases. |

## Known follow-up (not in this PR)

The bridge + modifier-fold path adds per-entity work that pushes `IRLuaPerfGrid`'s exit-clean budget from ~21s (pre-fix, no motion — auto-screenshot fires and the demo cleanly exits) to >45s (post-fix, with motion — exceeds the bash watchdog backstop in the platform-catchup smoke script). For 64×64×64 = 262,144 entities this is the same perf class as #1154 (IRPerfGrid 1 FPS on OpenGL). Worth a follow-up `optimize` pass on the `PROPAGATE_TRANSFORM` modifier-fallback compose path or the `upsertBySourceInPlace` iteration cost — but the motion fix is the right move regardless.

## Test plan

- [x] `fleet-build --target IRLuaPerfGrid` builds clean on `linux-debug`.
- [x] `fleet-run IRLuaPerfGrid --auto-screenshot 30` captures both `fit_grid` and `zoom1_origin` shots; PNG byte sizes differ from pre-fix baseline, confirming motion.
- [x] Visual side-by-side (read PNGs in tool) shows static cube → wave pattern.
- [ ] macOS reviewer to confirm `IRLuaPerfGrid` on Metal also gains visible wave motion (the bug is host-agnostic; this fix should help both backends).

Discovered during the platform-catchup smoke run on 2026-05-24 (matches user observation: "one of the perf_grid demos doesn't have any motion, maybe the new system doesn't work well for lua code as far as transforms").

### DIFF
--- a/creations/demos/lua_perf_grid/main_lua.cpp
+++ b/creations/demos/lua_perf_grid/main_lua.cpp
@@ -28,6 +28,9 @@
 #include <irreden/common/components/component_name.hpp>
 #include <irreden/common/components/component_local_transform.hpp>
 #include <irreden/common/components/component_local_transform_lua.hpp>
+#include <irreden/common/components/component_modifiers.hpp>
+#include <irreden/common/modifier.hpp>
+#include <irreden/common/transform_modifier_fields.hpp>
 #include <irreden/render/components/component_canvas_ao_texture.hpp>
 #include <irreden/render/components/component_canvas_fog_of_war.hpp>
 #include <irreden/render/components/component_canvas_light_volume.hpp>
@@ -276,6 +279,7 @@ void createGridEntities() {
                 IREntity::createEntity(
                     C_LocalTransform{positionForCell(x, y, z)},
                     C_VoxelSetNew{ivec3(1, 1, 1), colorForCell(x, y, z, n), false},
+                    C_Modifiers{},
                     makeWaveState(x, y, z)
                 );
             }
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — demos/lua_perf_grid: bridge C_LuaWaveState → C_Modifiers so wave motion reaches rendered position

**Verdict:** approve

### Praise

- **Correct identification of the missing bridge.** The wave kernel was firing but the output `out_x_/y_/z_` fields were never consumed. Adding the `LuaWaveStateToOffset` system is precisely the right fix — matches the `PERIODIC_IDLE_POSITION_OFFSET` pattern that the C++ baseline uses.
- **No ECS invariant violations.** The new system uses `createSystem<C_LuaWaveState, C_Modifiers>(...)` so both components arrive via dense archetype-column iteration — no `getComponent` in the tick body.
- **Correct use of `upsertBySourceInPlace`.** The vec3 overload at `modifier.hpp:623` updates the existing slot in-place after the first tick (no steady-state allocation), falling back to `push_back` only on the first tick per entity. This is the canonical pattern for per-frame steady-state writers.
- **Pipeline ordering is correct.** `waveSysId → luaWaveOffsetId → PROPAGATE_TRANSFORM` ensures the modifier is upserted before the propagation pass reads it. `PROPAGATE_TRANSFORM` confirmed to have the `C_Modifiers` direct-read fallback when `C_ResolvedFields` is absent (`system_propagate_transform.hpp:42-43`), so no resolver pipeline is needed here.
- **`C_Modifiers{}` attached at creation.** Slot exists for `upsertBySourceInPlace` to find on frame 1; avoids the miss-then-push-back on every tick.
- **Visual verification is complete.** Pre/post PNG byte comparison (static 261,152B × 4 → varying 338,674B / 336,453B) is a good smoke signal for real motion.

- [x] `fleet-build --target IRLuaPerfGrid` on `linux-debug` — clean build confirmed per PR body
- [x] `fleet-run IRLuaPerfGrid --auto-screenshot 30` — PNG sizes differ, confirming wave motion
- [ ] macOS reviewer to confirm `IRLuaPerfGrid` on Metal shows visible wave motion (bug is host-agnostic; fix should transfer)

Opus recheck not required.

PR #1122, opened 2026-05-23
T-334: engine/system: validator rejects PARALLEL_FOR + relation-form

jakildev:
## Summary
- Adds `SystemAccess::isRelationForm_` and a 4th rule in `validateConcurrencyForAccess` that FATALs when a system requests `Concurrency::PARALLEL_FOR` and its tick body matches the relation form `(Components&..., std::optional<RelComps*>...)`.
- The relation branch in `system_manager.hpp`'s `rangedFn` resolves the related entity and its components via `EntityManager` lookups inside the per-row loop. Those manager accesses are not thread-safe — a future system combining `PARALLEL_FOR` + `RelationParams<...>` would silently race on the manager's archetype map from worker threads.
- The bit is folded in by `createSystem` via a constexpr lambda rather than `deriveAccessFromSignature` — the trait's free-function shape can't carry both `Components...` and `RelationComponents...` packs without ambiguity (same root cause as the existing TODO at `InvocableWithOptionalRelations` in `ir_system_types.hpp`).
- New `RelationFormRejected` test constructs the descriptor directly and asserts the validator predicate, mirroring `PerEntityIdFormAcceptedWithParallelSafe`.
- No currently-registered system combines relations + `PARALLEL_FOR`, so no runtime behavior changes.

## Stack context
Stacked on: https://github.com/jakildev/IrredenEngine/pull/1097 (T-222)

## Test plan
- [x] `fleet-build --target IrredenEngineTest` clean
- [x] `SystemConcurrencyValidator*` — 6/6 pass (including new `RelationFormRejected`)
- [x] `fleet-build --target IRShapeDebug` clean — every prefab system registers through the changed `createSystem` template
- [x] `IRShapeDebug --auto-screenshot 10` — 7/7 shots captured, no validator FATALs at startup
- [x] Full IrredenEngineTest suite — 906/906 pass
- [ ] Linux smoke (engine/system core path, carried via upstream T-221's `fleet:needs-linux-smoke`)

Closes #1106

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1157, opened 2026-05-24
update/propagate_transform: root-archetype fast path (skip quatMul + rotateVectorByQuat when parent is null)

jakildev:
## Summary

Optimization for `SYSTEM_PROPAGATE_TRANSFORM`. For root archetypes (entities with no `CHILD_OF` parent), `parentWorld` stays at default-constructed identity. The SQT compose then reduces to:

```
identity_quat × local.rotation              = local.rotation
rotateVectorByQuat(local.translation, id)   = local.translation
identity_scale × local.scale × modScale     = local.scale × modScale
```

…with `quatMul` and `rotateVectorByQuat` both no-ops. The previous code ran them unconditionally — `parentWorld.rotation_` is loaded from memory, so the compiler can't constant-fold the identity check.

This change hoists the root check (`parent == kNullEntity`) out of the inner loop and takes a quat-free fast path. Saves ~1 `quatMul` (~16 fp ops) + ~1 `rotateVectorByQuat` (~30 fp ops) + ~1 parent-scale vec3 mul per root entity per frame.

## Measured impact

`IRLuaPerfGrid` (64×64×64 = 262,144 root entities with a per-frame `TRANSFORM_TRANSLATION` modifier per PR #1155's motion-fix wave, cherry-picked locally only to make the measurement valid — without the modifier the entities don't exercise the per-tick compose path):

| Build                                          | Update tick        | Demo runtime           |
|------------------------------------------------|--------------------|------------------------|
| master + PR #1155 (motion fix, no opt)         | 14-17 ms           | >45 s — bash watchdog killed, didn't auto-exit |
| master + PR #1155 + this PR                    | **12-15 ms**       | **~41 s** — auto-exits cleanly via `--auto-screenshot` close-window |

Both on `linux-x86_64` (WSLg Mesa-d3d12, master `3d031e44`). 10-20% Update-tick improvement, sufficient to bring back clean auto-exit.

The remaining slowdown vs the pre-motion-fix 21 s baseline is **GPU-bound** (the modifier-resolved wave displaces entities every frame and the lighting compute does real work instead of running on a static scene); see #1154 for the GPU side.

## Correctness

Mathematically equivalent for root entities — verified by running:
- All **10** `PropagateTransformTest` tests (including `RootEntityWorldEqualsLocal` which directly exercises the root branch + the multi-level chain tests which exercise the non-root branch).
- The broader `*Modifier*:*Transform*:*Skeleton*` filter suite, **159 tests** all green.

Non-root archetypes (any with `CHILD_OF`) keep the full quat path — quat math is meaningful when `parentWorld` is non-identity.

## Test plan

- [x] `IrredenEngineTest --gtest_filter='PropagateTransformTest.*'` — 10/10 pass.
- [x] `IrredenEngineTest --gtest_filter='*Modifier*:*Transform*:*Skeleton*'` — 159/159 pass.
- [x] `fleet-build --target IRLuaPerfGrid` clean on `linux-debug`.
- [x] `fleet-run IRLuaPerfGrid --auto-screenshot 30` measured against PR #1155's combined diff: 41 s clean exit, Update tick 12-15 ms (was 14-17 ms).
- [ ] macOS reviewer to confirm test suite passes on `macos-debug` and run a brief `IRPerfGrid` smoke to confirm no Metal regression.

Discovered during the platform-catchup `/optimize` flow on 2026-05-24. Standalone change — not coupled to #1155 (which is the standalone lua_perf_grid motion fix). They land independently and compound when both are in.

### DIFF
--- a/engine/prefabs/irreden/update/systems/system_propagate_transform.hpp
+++ b/engine/prefabs/irreden/update/systems/system_propagate_transform.hpp
@@ -210,15 +210,9 @@ template <> struct System<PROPAGATE_TRANSFORM> {
                 &IREntity::getComponentData<IRComponents::C_Modifiers>(node);
         }
 
-        // Root archetypes (no CHILD_OF parent) have `parentWorld == identity`.
-        // In that case the SQT composition collapses to: world = local +
-        // modifiers, with quatMul/rotateVectorByQuat both no-ops. Hoist
-        // the root check out of the inner loop and take a quat-free fast
-        // path — saves one quatMul (~16 fp ops) and one rotateVectorByQuat
-        // (~30 fp ops) per entity per frame. At perf_grid / lua_perf_grid
-        // scale (262K root entities × 60 Hz) this is ~720M fp ops/sec
-        // off the CPU budget. Non-root archetypes (any with CHILD_OF) keep
-        // the full path — that's where the quat math is meaningful.
+        // parentWorld == identity for root archetypes, so quatMul and
+        // rotateVectorByQuat are both no-ops. Hoist the check out of the
+        // inner loop and skip the quat math — only meaningful with CHILD_OF.
         const bool isRootArchetype = (parent == IREntity::kNullEntity);
 
         for (int i = 0; i < node->length_; ++i) {
--- a/engine/prefabs/irreden/update/systems/system_propagate_transform.hpp
+++ b/engine/prefabs/irreden/update/systems/system_propagate_transform.hpp
@@ -210,6 +210,17 @@ template <> struct System<PROPAGATE_TRANSFORM> {
                 &IREntity::getComponentData<IRComponents::C_Modifiers>(node);
         }
 
+        // Root archetypes (no CHILD_OF parent) have `parentWorld == identity`.
+        // In that case the SQT composition collapses to: world = local +
+        // modifiers, with quatMul/rotateVectorByQuat both no-ops. Hoist
+        // the root check out of the inner loop and take a quat-free fast
+        // path — saves one quatMul (~16 fp ops) and one rotateVectorByQuat
+        // (~30 fp ops) per entity per frame. At perf_grid / lua_perf_grid
+        // scale (262K root entities × 60 Hz) this is ~720M fp ops/sec
+        // off the CPU budget. Non-root archetypes (any with CHILD_OF) keep
+        // the full path — that's where the quat math is meaningful.
+        const bool isRootArchetype = (parent == IREntity::kNullEntity);
+
         for (int i = 0; i < node->length_; ++i) {
             const auto &local = locals[i];
 
@@ -227,6 +238,15 @@ template <> struct System<PROPAGATE_TRANSFORM> {
                 modScale = IRPrefab::Modifier::detail::composeForFieldVec3(
                     IRMath::vec3(1.0f), scaleField, modsVec3
                 );
+            }
+
+            if (isRootArchetype) {
+                worlds[i] = IRComponents::C_WorldTransform{
+                    local.translation_ + modTranslation,
+                    local.rotation_,
+                    local.scale_ * modScale,
+                };
+                continue;
             }
 
             const IRMath::vec3 worldScale =


jakildev (commented):
## Review — update/propagate_transform: root-archetype fast path (skip quatMul + rotateVectorByQuat when parent is null)

**Verdict:** approve

### Praise

- **Mathematically correct.** For root archetypes (`parent == kNullEntity`), `parentWorld` stays at default-constructed identity. The fast path computes `world = {local.translation_ + modTranslation, local.rotation_, local.scale_ * modScale}` which is identical to running the full compose with an identity parent:  `(1,1,1)*local.scale_*modScale = local.scale_*modScale`, `quatMul(identity, local.rotation_) = local.rotation_`, `(0,0,0) + rotateVectorByQuat((1,1,1)*local.translation_, identity) + modTranslation = local.translation_ + modTranslation`. ✓
- **Correctly scoped to `kNullEntity` only.** The flag is `parent == kNullEntity` — it does NOT fire when `parent != kNullEntity` but the parent entity is missing `C_WorldTransform` (that case still runs the full path with identity `parentWorld`). Conservative and correct.
- **`isRootArchetype` hoisted out of the loop.** The check is once per archetype-node, not per-entity. ✓
- **No ECS invariant violations.** No `getComponent` in the inner loop. No allocations in the hot path. No structural mutations mid-iteration. ✓
- **Direct test coverage.** `PropagateTransformTest::RootEntityWorldEqualsLocal` (test/ecs/propagate_transform_test.cpp:86) creates a root entity and verifies world == local post-tick. The multi-level chain tests (TwoLevelChain, ThreeLevelChain, RotationPropagates, TenLevelChain) verify the non-root path is unchanged. 159 additional modifier/transform/skeleton tests all green. ✓
- **Measured speedup with concrete baseline.** 10-20% Update-tick improvement at 262K root entities, bringing the demo back to clean auto-exit. PR body includes before/after measurements with exact host/SHA. Good perf hygiene.

### Nits

- `system_propagate_transform.hpp:213-221` — The 9-line block comment explaining the optimization is on the verbose side given the codebase's "one line max, only when WHY is non-obvious" policy. The WHY here is genuinely non-obvious (identity parent ↔ quat no-ops), so a comment is warranted, but it could be trimmed to 3-4 lines: the key invariant is "parentWorld == identity ⟹ quat math is a no-op; hoist the check and skip it." The FP-ops and Hz math adds detail the reader can derive from the operation counts.

- [x] `IrredenEngineTest --gtest_filter='PropagateTransformTest.*'` — 10/10 pass (per PR body)
- [x] `IrredenEngineTest --gtest_filter='*Modifier*:*Transform*:*Skeleton*'` — 159/159 pass (per PR body)
- [x] `fleet-build --target IRLuaPerfGrid` clean on `linux-debug` (per PR body)
- [x] `fleet-run IRLuaPerfGrid --auto-screenshot 30` — 41 s clean exit, 12-15 ms Update tick (per PR body)
- [ ] macOS reviewer to confirm test suite passes on `macos-debug` and brief `IRPerfGrid` smoke (no Metal regression expected — pure CPU arithmetic, no GPU changes)

Opus recheck not required.

jakildev:
Addressed nit: trimmed the 9-line comment at `system_propagate_transform.hpp:213` to 3 lines. Kept the key invariant (parentWorld == identity ⟹ quat math no-op, hoist out of loop) and dropped the FP-ops/Hz math the reader can derive.

jakildev (commented):
## Opus recheck — PR #1157: PROPAGATE_TRANSFORM root-archetype fast path

**Verdict:** approve

Picked up for Opus recheck because the diff lands in `engine/prefabs/irreden/update/systems/` and modifies the canonical SQT composition path that every transform consumer downstream (render, gizmo, physics) relies on. Sonnet's first-pass approve verdict stands; this pass does an independent math derivation and edge-case sweep that Sonnet's body summarized but didn't fully unfold.

### What Sonnet checked vs what I re-derived

Sonnet flagged the change as "mathematically correct" with a four-line summary of the identity collapse. On closer read I confirm the verdict, but two of the four steps in that summary rely on premises that are worth re-stating because they wouldn't survive a small refactor without a comment:

**(1) `parentWorld` default ↔ identity SQT — confirmed.** `C_WorldTransform`'s default constructor (`component_world_transform.hpp:17-22`) gives `scale_ = (1,1,1)`, `rotation_ = (0,0,0,1)`, `translation_ = (0,0,0)`. The fast path is only valid because this default is identity in all three slots; if any future change reorders fields or shifts the default rotation to (1,0,0,0) (xyzw vs wxyz), the fast path becomes silently wrong on rotation. A one-line `static_assert` next to the fast path (or in the component header itself) would lock this, but it's defensive — not a blocker for this PR.

**(2) `quatMul(identity, q) == q` and `rotateVectorByQuat(v, identity) == v` — confirmed.** Both depend on Hamilton-product orientation conventions in `engine/math/`. Identity quat `(0,0,0,1)` is left-identity for `quatMul` (the function's signature is `quatMul(parent, local)` per the composition formula at line 12 of the header), and identity rotation is a no-op for `rotateVectorByQuat`. Both invariants hold in the current `IRMath` impl.

**(3) Hoist is sound across all entities in the archetype-node.** `parent` at line 179-180 is derived from `node->type_` (archetype-level CHILD_OF parent encoded into the type), not from any per-entity component. Every entity in a single `ArchetypeNode` therefore shares the same `parent` value, so `isRootArchetype` is constant across the inner loop — correctly hoisted out, no per-entity recompute. ✓

**(4) `parent != kNullEntity` but parent lacks `C_WorldTransform`.** This case still takes the slow path with `parentWorld` left at identity (lines 181-187: `getComponentOptional` returns null, the `parentWorld = *p;` line is skipped). The slow path computes the same result as the fast path would, just with the quat math executed redundantly. This is the "conservative scoping" Sonnet praised — confirmed safe, just slightly under-optimized.

### Independent Opus invariants

- **Topological-sort interaction.** The `beginTick` topological sweep at lines 100-147 admits a node as "ready" in three cases: `parent == kNullEntity`, parent record missing, or parent archetype outside the candidate set / already done. The fast path fires on case 1 only (the strict root case). Cases 2 and 3 still produce correct results via the slow path with identity `parentWorld`. The topological order itself is unchanged — the fast path only affects what work happens per entity, not what order entities are processed. ✓
- **Concurrency.** `registerSystem<PROPAGATE_TRANSFORM, …>` at line 164 supplies no `Concurrency` arg, so the system runs SERIAL. `beginTick` runs once per pipeline execution on the main thread. No race on `worlds[i]` writes, no race on `parentWorld` reads. The recently-landed PARALLEL_FOR validator (#1097, #1122) doesn't apply here. ✓
- **No new allocations, no new `getComponent` in the inner loop.** The added block is one bool comparison + one struct write + `continue` — strictly cheaper than the slow path it replaces. The `getComponentOptional<C_WorldTransform>(parent)` at line 183 was already per-node, not per-entity, and remains unchanged. ✓
- **Modifier path unaffected.** `modTranslation` / `modScale` resolution at lines 216-230 runs before the branch in both paths; the fast path consumes the resolved modifier values identically to the slow path. The resolver-pipeline fallback to direct `C_Modifiers` composition (lines 207-211, 222-229) is preserved. ✓
- **Test coverage.** The direct invariant — root entity world equals `local + modifiers` post-tick — is covered by `PropagateTransformTest::RootEntityWorldEqualsLocal` (Sonnet cited it; 10/10 pass per PR body). The non-root paths (TwoLevelChain, ThreeLevelChain, RotationPropagates, TenLevelChain) exercise the unchanged slow path. Combined with the 159 modifier/transform/skeleton tests passing, the coverage is adequate.

### Nits (echoing Sonnet's only finding)

Sonnet's verbose-comment nit at lines 213-221 is the only outstanding item and it's genuine — 9 lines for one optimization exceeds the "one line max, WHY-only" policy in `CLAUDE-BASELINE.md`. The FP-ops accounting (~16 + ~30 fp ops) and 720M-ops/sec extrapolation are derivable from the operation counts and don't need to live in the source. A 3-line version naming the invariant (`parentWorld == identity ⟹ quat math is a no-op`) is sufficient. Not blocking — the author worker can fold it before merge.

Optional defensive addition (not required): a `static_assert` near the fast path that the default `C_WorldTransform` is identity, so the invariant can't drift silently if the component layout changes.

### What I did not re-check

- Build / test execution (Sonnet ran the suites, results in PR body).
- Naming convention, `SystemName` enum, `m_` / trailing-`_` discipline (Sonnet's first-pass surface).
- Performance measurement methodology (Sonnet validated the before/after numbers against the PR body).

### Cross-host smoke

PR is `fleet:authored-on-linux`. The change is pure CPU arithmetic with no GPU touch points, so no shader/render parity concern — macOS smoke is unnecessary beyond a routine test-suite pass. Sonnet's "macOS reviewer to confirm" note in the test plan is the right level of follow-up.

jakildev:
Addressed nit: confirmed the 9-line comment at system_propagate_transform.hpp:213-221 was already trimmed to 3 lines (parentWorld == identity ⟹ quat math is a no-op; hoist out of loop). No new code changes needed — label cleared.

PR #1158, opened 2026-05-24
T-348: engine/system: SERIAL fast-path + dual-slot consolidation

### LINKED ISSUE #1124: engine/system: SERIAL fast-path skips binder for non-PARALLEL_FOR dispatch
## Context

Follow-up from T-333 (PR #1123) review (Sonnet reviewer nit).

The T-333 binder pattern (`prepareRangedTick_`) resolves all `getComponentData<Components>(node)` calls once on the main thread and returns a `void(begin, end)` closure. For the **SERIAL / MAIN_THREAD** path, `functionTick_` calls the binder once per matched archetype node and invokes the returned closure with `[0, length)`, preserving behavioral equivalence at the cost of one `std::function<void(int, int)>` heap allocation per node per tick.

The old `rangedFn`-inside-`perNodeFn` design (pre-T-333) called through a pre-captured closure without an intermediate allocation. The new design is correct for PARALLEL_FOR (the binder must be separate so workers get a closure that doesn't touch EntityManager) but the SERIAL path pays an extra allocation unnecessarily.

## Task

In `SystemManager::executePipeline` (or `createSystem`'s dispatch closure), implement a SERIAL fast-path that calls the pre-captured component-column refs closure directly without going through the binder when `Concurrency == SERIAL || Concurrency == MAIN_THREAD`. The binder is only needed for PARALLEL_FOR.

**Affected file:** `engine/system/include/irreden/system/system_manager.hpp` — `prepareRangedTick` / `perNodeFn` dispatch.

## Why escalating

Not a blocker for T-333 (PR body acknowledged "negligible vs frame budget"). Allocation footprint grows proportionally as PARALLEL_FOR rolls out to more systems (T-223 Lua DSL, T-225 deferred mutations), so the fast-path is worth tracking.

**Suggested model:** [sonnet]  
**Suggested area:** `engine/system/`  
**Suggested approach:**
- In the `createSystem` dispatch closure, branch on `kConcurrency` (or the runtime `Concurrency` parameter): when `SERIAL` / `MAIN_THREAD`, call the component-column refs directly without the intermediate binder.
- The `prepareRangedTick_` member remains on the system for PARALLEL_FOR dispatch; the SERIAL closure captures the resolved refs at `createSystem` time instead (same lifetime guarantee — structural changes are deferred until `flushStructuralChanges` after `executePipeline`).
- Build `IrredenEngineTest` + `IRShapeDebug` after the change; confirm existing PARALLEL_FOR systems still fan out correctly.

jakildev:
## Summary
- Drop `functionTick_` for row-iterating tick forms (per-component, per-entity-id, optional-relation); only `prepareRangedTick_` is stored for these.
- Remove the `perNodeFn` wrapper lambda and the `prepareRangedTickCopy` that caused dual heap allocation per row-iterating system at registration time.
- `executeSystem` SERIAL/MAIN_THREAD else-path now branches: calls `prepareRangedTick_(node)(0, length)` when the binder is populated, falls back to `functionTick_` for batch/dynamic forms.
- `replaceSystemBody` + `createSystemDynamic` (Lua hot-reload) unaffected — those paths only use `functionTick_`.
- Update comments in `system_manager.hpp`, `component_system_event.hpp`, `system_manager.cpp`, and `engine/system/CLAUDE.md` to reflect the single-slot contract.

## Test plan
- [x] `fleet-build --target IRShapeDebug` — clean build, no new warnings
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — all 7 shots captured, no crash
- [x] `fleet-build --target format-changed && fleet-build --target IrredenEngineSystem` — clean rebuild after simplify pass

Closes #1124

### DIFF
--- a/engine/system/CLAUDE.md
+++ b/engine/system/CLAUDE.md
@@ -22,9 +22,12 @@ matched archetype. Per-entity iteration is the body's responsibility,
 not SystemManager's — the Lua surface uses this to keep the C++/Lua
 boundary at one `sol::function` call per archetype.
 
-Both paths share the same scheduler: `executePipeline` walks
-`m_ticks[system].functionTick_`, which is a `std::function<void(
-ArchetypeNode*)>` regardless of which factory created the system.
+Both paths share the same scheduler: `executePipeline` dispatches via
+`m_ticks[system].prepareRangedTick_` for row-iterating forms (the binder
+resolves component vectors on the main thread and returns a
+`void(begin,end)` closure), or via `m_ticks[system].functionTick_` for
+the batch form and dynamic systems; row-iterating forms leave
+`functionTick_` empty.
 
 ## `replaceSystemBody` for hot-reload
 
--- a/engine/system/include/irreden/system/components/component_system_event.hpp
+++ b/engine/system/include/irreden/system/components/component_system_event.hpp
@@ -37,20 +37,21 @@ template <> struct C_SystemEvent<IRSystem::TICK> {
     /// entirely through `prepareRangedTick_` (T-348).
     TickFn functionTick_;
 
-    /// Main-thread binder for `Concurrency::PARALLEL_FOR` dispatch.
+    /// Main-thread binder for row-iterating tick forms; used by
+    /// SERIAL/MAIN_THREAD and PARALLEL_FOR alike (T-348).
     /// SystemManager calls this **once on the main thread** per matched
-    /// archetype node before fanning chunks out to `IRJob::parallelFor`.
-    /// The binder resolves the per-component vector references from the
-    /// node (going through `EntityManager::getComponentType<>` →
-    /// `m_pureComponentTypes` hash lookup, which is NOT safe under
-    /// concurrent reads from worker threads), captures them in a
-    /// `void(rangeBegin, rangeEnd)` closure, and returns it for workers
-    /// to invoke without ever touching EntityManager state. Populated
-    /// by the template `createSystem<...>` path for tick signatures the
-    /// runtime can chunk safely; empty for `createSystemDynamic`
-    /// (opaque body) and the per-archetype batch form (consumes the
-    /// whole column) — the registration validator rejects
-    /// `PARALLEL_FOR` in both cases.
+    /// archetype node. The binder resolves the per-component vector
+    /// references from the node (going through
+    /// `EntityManager::getComponentType<>` → `m_pureComponentTypes`
+    /// hash lookup, which is NOT safe under concurrent reads from worker
+    /// threads), captures them in a `void(rangeBegin, rangeEnd)` closure,
+    /// and returns it. For SERIAL/MAIN_THREAD, `executeSystem` invokes
+    /// the closure immediately with `[0, length)`. For PARALLEL_FOR,
+    /// it fans the closure out to `IRJob::parallelFor`. Populated by the
+    /// template `createSystem<...>` path for tick signatures the runtime
+    /// can chunk safely; empty for `createSystemDynamic` (opaque body)
+    /// and the per-archetype batch form (consumes the whole column) —
+    /// the registration validator rejects `PARALLEL_FOR` in both cases.
     PrepareRangedTickFn prepareRangedTick_;
 
     IREntity::Archetype archetype_;
--- a/engine/system/include/irreden/system/components/component_system_event.hpp
+++ b/engine/system/include/irreden/system/components/component_system_event.hpp
@@ -27,13 +27,14 @@ template <> struct C_SystemEvent<IRSystem::TICK> {
     /// to resolve component vectors and return a worker closure that
     /// covers `[rangeBegin, rangeEnd)` of the resolved rows. See the
     /// `prepareRangedTick_` field below for the full contract.
-    using PrepareRangedTickFn = std::function<
-        std::function<void(int rangeBegin, int rangeEnd)>(IREntity::ArchetypeNode *)>;
+    using PrepareRangedTickFn =
+        std::function<std::function<void(int rangeBegin, int rangeEnd)>(IREntity::ArchetypeNode *)>;
 
-    /// Whole-node dispatch — the body iterates `[0, node->length_)`
-    /// internally. The legacy form; every system has one. For
-    /// `Concurrency::SERIAL` and `Concurrency::MAIN_THREAD` this is the
-    /// path SystemManager invokes.
+    /// Whole-node dispatch — used ONLY by the per-archetype batch form
+    /// (`InvocableWithNodeVectors`) and dynamic systems created via
+    /// `createSystemDynamic` / Lua hot-reload. Row-iterating forms
+    /// (per-component, per-entity-id) leave this empty and dispatch
+    /// entirely through `prepareRangedTick_` (T-348).
     TickFn functionTick_;
 
     /// Main-thread binder for `Concurrency::PARALLEL_FOR` dispatch.
--- a/engine/system/include/irreden/system/system_manager.hpp
+++ b/engine/system/include/irreden/system/system_manager.hpp
@@ -193,8 +193,7 @@ class SystemManager {
     /// between systems within a group (group members are required to
     /// be structurally clean by the validator). Replaces any prior
     /// registration for `event`.
-    void
-    registerPipelineGroups(IRTime::Events event, std::vector<std::vector<SystemId>> groups);
+    void registerPipelineGroups(IRTime::Events event, std::vector<std::vector<SystemId>> groups);
 
     /// T-224: validate every registered pipeline group against the
     /// cross-system access rules in `system_access.hpp`. Call once
@@ -306,9 +305,8 @@ class SystemManager {
     }
 
     // Tick functions are operated on each entity in the system
-    // matching the archetype. The runtime stores **two** dispatch
-    // shapes per system (T-222 Phase 2 of the multithreading epic;
-    // worker-side reshaped by T-333):
+    // matching the archetype. The runtime stores one dispatch slot per
+    // system — the binder form established by T-222/T-333:
     //
     //   - `prepareRangedTick(node)` — main-thread binder. Resolves
     //     the per-component vector refs from `node` once via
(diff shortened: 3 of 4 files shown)


jakildev (commented):
## Review — T-348: engine/system: SERIAL fast-path + dual-slot consolidation

**Verdict:** approve

### Praise

- **Correct dispatch branch.** `executeSystem`'s new SERIAL/MAIN_THREAD else-path (`system_manager.cpp:383-388`) correctly keys on `prepareRangedTick_` being non-empty for row-iterating forms and falls through to `functionTick_` for batch/dynamic. The `std::function::operator bool()` conversion is well-defined. ✓
- **PARALLEL_FOR path unaffected.** The `prepareRangedTick_` call at `system_manager.cpp:370` and the `IRJob::parallelFor` fan-out at line 375 were already calling the binder directly; T-348 doesn't touch them. ✓
- **`createSystemDynamic` + `replaceSystemBody` unaffected.** Dynamic systems use the two-arg `C_SystemEvent<TICK>` constructor, which sets `prepareRangedTick_` to empty and `functionTick_` to the body. The new branch correctly falls through to `functionTick_` for these. The PARALLEL_FOR validator already blocks dynamic systems (`PARALLEL_FOR + isBatchForm_` → FATAL), so the slot contract is consistent. ✓
- **No ECS invariant violations.** No `getComponent` in ticks, no allocations in the hot dispatch path, no structural mutations mid-iteration. ✓
- **Correct construction for batch forms.** The batch-form `perNodeFn` at `system_manager.hpp:342-360` is unchanged — it is the actual body, not a wrapper — and correctly leaves `prepareRangedTick_` empty. ✓
- **Visual smoke test solid.** `IRShapeDebug --auto-screenshot 10` — 7 shots, no crash. Exercises the SERIAL dispatch path through real row-iterating systems. ✓

### Nits

- `engine/system/include/irreden/system/system_manager.hpp:370` — Comment still says "The SERIAL / MAIN_THREAD entry (`perNodeFn`) calls the binder once" but `perNodeFn` is gone from the row-iterating path after T-348. Update to reference `executeSystem`'s SERIAL/MAIN_THREAD branch directly: "The SERIAL / MAIN_THREAD path in `executeSystem` calls the binder once and invokes the returned closure with `[0, length)`."

- `engine/system/src/system_manager.cpp:97` — `replaceSystemBody` sets `functionTick_` without guarding against callers targeting a row-iterating system (which now has `functionTick_` empty but `prepareRangedTick_` populated). After T-348, such a call silently becomes a no-op — dispatch will keep invoking `prepareRangedTick_`, ignoring the new body. In practice, callers are always `createSystemDynamic` EVAL systems, so this won't bite; but an `IR_ASSERT(!static_cast<bool>(m_ticks[system].prepareRangedTick_), "replaceSystemBody: row-iterating systems use prepareRangedTick_, not functionTick_")` would surface the misuse in debug builds.

- [x] `fleet-build --target IRShapeDebug` — clean build, no new warnings (per PR body)
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — 7 shots captured, no crash (per PR body)
- [x] `fleet-build --target IrredenEngineSystem` — clean rebuild after simplify pass (per PR body)
- [ ] Run the test binary: `./IrredenEngineSystem` (or `--gtest_filter='*System*:*Pipeline*'`) — the PR builds but doesn't run the suite; for a dispatch-path change it's worth one explicit pass.

Opus recheck required: PR touches `engine/system/` and modifies `executeSystem`'s core dispatch branch; Opus should verify the `replaceSystemBody` row-iterating invariant (nit #2) and confirm the PARALLEL_FOR → SERIAL fallback path is still correct after the slot split.

jakildev:

- **Comment update** (`system_manager.hpp:370`): replaced stale `perNodeFn` reference with `executeSystem`'s SERIAL/MAIN_THREAD path description, matching the T-348 dual-slot contract.
- **Debug assert** (`system_manager.cpp:97`): added `IR_ASSERT(!prepareRangedTick_, ...)` before the `functionTick_` assignment in `replaceSystemBody` — surfaces misuse in debug builds (calling `replaceSystemBody` on a row-iterating system that has `prepareRangedTick_` populated), no-op in release.
- All 733 `IrredenEngineTest` tests pass.

jakildev (commented):
## Opus recheck — PR #1158: T-348 engine/system: SERIAL fast-path + dual-slot consolidation

**Verdict:** approve

Picked up for Opus recheck because the diff lands in `engine/system/` and modifies `executeSystem`'s core dispatch branch. Sonnet's first-pass approve verdict stands; this pass focuses on the two items Sonnet explicitly asked Opus to confirm: the `replaceSystemBody` row-iterating invariant and the PARALLEL_FOR → SERIAL fallback path. I also add one comment-drift item Sonnet did not catch.

### Sonnet asked Opus to confirm (1): PARALLEL_FOR → SERIAL fallback

The else-branch in `system_manager.cpp:379-385` is reached from PARALLEL_FOR in three fallback cases (all on `executeSystem`'s line 351 conjunction failing):

- **`node->length_ <= grainSize`** — small-node path.
- **`g_jobManager == nullptr`** — pre-World init / unit tests with no worker pool.
- **`!IRJob::isMainThread()`** — nested dispatch when `executeSystem` is reached from a worker (multi-system parallel group via `executePipeline`).

For all three, the system was registered with `Concurrency::PARALLEL_FOR`, which the entry-point validator (`ir_system.hpp:62-100` `validateConcurrencyForAccess`) FATALs out unless the tick form is row-iterating: `!isBatchForm_` AND `!isRelationForm_` AND (`!usesEntityId_` OR `parallelSafe_`). Row-iterating forms populate `prepareRangedTick_` (`system_manager.hpp` else-branch at the row-iterating emplace_back). So at runtime, `tickEvent.prepareRangedTick_` MUST be bool-true for any PARALLEL_FOR system reaching the else-branch. The new code at `system_manager.cpp:381-385` therefore routes through `prepareRangedTick_(node)(0, node->length_)` for all three fallback paths — equivalent to invoking the binder once on the main thread and walking the full range, which is exactly the prior `perNodeFn` wrapper's behavior, with one less heap allocation. ✓ Confirmed correct.

The fourth condition `static_cast<bool>(tickEvent.prepareRangedTick_)` on line 352 is a defensive guard (against a future PARALLEL_FOR system slipping past the validator), not a load-bearing branch — but its presence means even a hypothetical PARALLEL_FOR + empty-binder slip would fall to the else-branch and exercise `functionTick_`, matching the prior behavior. The new code preserves this. ✓

### Sonnet asked Opus to confirm (2): `replaceSystemBody` row-iterating invariant

Sonnet's claim that "callers are always `createSystemDynamic` EVAL systems" is correct as of today. Verified by tracing the call graph:

- The Lua surface `IRSystem.replaceSystemBody` (`lua_script.cpp:935-958`) explicitly rejects non-EVAL systems by `m_luaSystemTicks.find(systemId)` returning end. Only systems registered via `IRSystem.registerSystem` with `mode='eval'` land in that map.
- The C++ free function `IRSystem::replaceSystemBody` (`ir_system.hpp:403`) is a thin wrapper — `git grep ::replaceSystemBody` finds no non-test direct C++ callers outside the Lua shim.
- `createSystemDynamic` uses the 3-arg `C_SystemEvent<TICK>` overload, which leaves `prepareRangedTick_` empty (`component_system_event.hpp:71-79`).

So today the contract holds: `replaceSystemBody` is only ever called on a system whose `prepareRangedTick_` is empty, and `functionTick_` is the live slot. **However, Sonnet's defensive `IR_ASSERT` suggestion is right on the merits** — the silent no-op failure mode is exactly the kind of bug that would surface much later, in a context that doesn't suggest the slot-split as the cause. The wrapper docstring at `ir_system.hpp:390-399` already acknowledges a different silent-disconnect risk ("calling this with a plain C++ body on a Lua-registered system's id silently disconnects the Lua hot-reload path"); adding a parallel `IR_ASSERT` for the row-iterating case extends the same defensive posture T-348 introduced.

The assert is a one-liner and adds zero runtime cost in release. Worth folding into a follow-up nit-fix; not a blocker for this PR.

### Additional finding Opus caught (comment drift)

Sonnet flagged the stale `perNodeFn` reference in the inline comment at `system_manager.hpp:370`. One more drift point sits in `component_system_event.hpp:39` — the `prepareRangedTick_` docstring opens with "Main-thread binder for `Concurrency::PARALLEL_FOR` dispatch." After T-348 this field is the **primary** dispatch path for every row-iterating system regardless of concurrency, not just for PARALLEL_FOR. The opening sentence reads as if the binder is PARALLEL_FOR-specific; the rest of the doc (lines 47-52) still says "Populated by the template `createSystem<...>` path for tick signatures the runtime can chunk safely; empty for `createSystemDynamic` (opaque body) and the per-archetype batch form …" which is now structurally correct but no longer the full story — the binder is also what SERIAL/MAIN_THREAD dispatches through. A one-line opener like "Main-thread binder for row-iterating tick forms; used by SERIAL/MAIN_THREAD and PARALLEL_FOR alike." would match the post-T-348 contract.

Same nit-tier as Sonnet's comment-drift finding — fold both into the same nit-fix commit.

### What I did not re-check

- Build / smoke / test execution (Sonnet ran `IRShapeDebug --auto-screenshot 10` and confirmed the SERIAL row-iterating dispatch path; results in PR body).
- Naming, `SystemName` enum, `m_` / trailing-`_` discipline (Sonnet's first-pass surface).
- Formatting churn (most of the deletions/additions in the diff are clang-format collapses that don't change behavior).

### Cross-host smoke

PR is `fleet:authored-on-linux`. The change is pure CPU dispatch with no GPU touch points or shader involvement, so no render-parity concern. A Linux-only smoke is sufficient; macOS test-suite pass is the right follow-up level — no per-stage render verification needed.

jakildev:
Addressed remaining fleet:has-nits feedback (Opus recheck finding):

- **`component_system_event.hpp:40`** — updated `prepareRangedTick_` opening docstring from "Main-thread binder for PARALLEL_FOR dispatch" to accurately reflect the post-T-348 contract: "Main-thread binder for row-iterating tick forms; used by SERIAL/MAIN_THREAD and PARALLEL_FOR alike". Also expanded the inline description to document both the SERIAL/MAIN_THREAD direct-invoke path and the PARALLEL_FOR fan-out path.

Build: `IrredenEngineSystem` clean.

PR #1159, opened 2026-05-24
T-349: engine/system: order validator rules most-specific-first for catch-all + PARALLEL_FOR

### LINKED ISSUE #1125: engine/system: validator FATAL message picks wrong rule on variadic catch-all tick + PARALLEL_FOR
## Context

Tracked from the Opus recheck on PR #1122 (T-334): the recheck verdict was approve-with-nits with one diagnostic-quality observation explicitly marked "no action this PR." Filing for future tracking so the nit does not get lost.

## The nit

In `engine/system/include/irreden/ir_system.hpp`, the `constexpr` lambda inside `createSystem` runs the relation-form `std::is_invocable_v<>` probe regardless of which Concurrency policy the user selected. That is the intended design — we want `isRelationForm_=true` recorded on every relation-form tick so future cross-system validation (T-224) can inspect the bit on SERIAL systems too.

However, the SystemAccess descriptor is computed by ORing the three orthogonal form probes together (`usesEntityId_`, `isBatchForm_`, `isRelationForm_`). A variadic catch-all tick like `[](auto&&...) {}` would pass every probe — the descriptor ends up with all three bits set. The dispatch chain in `rangedFn` (`engine/system/include/irreden/system/system_manager.hpp`) resolves "first match wins" via its `if constexpr` cascade, but the validator at `validateConcurrencyForAccess` has no such order — `IR_ASSERT` fires on the first failing rule and aborts.

With `PARALLEL_FOR + no ParallelSafe` on a catch-all tick:
- Three validator rules apply simultaneously (entity-id, batch-form, relation-form).
- `IR_ASSERT` fires on the first one in source order (currently `usesEntityId_`).
- The user sees the entity-id error message, not the more-specific relation-form one.

Not a correctness issue — FATAL is still FATAL. Only matters if a future use case writes a catch-all tick and needs precise diagnostics about which form the validator caught.

## Repro

No live caller hits this; the validator already catches every real-world case correctly. The diagnostic gap is visible only with an artificial catch-all signature.

## Possible fixes

- Order the validator rules from most-specific to least-specific so the most useful message wins.
- OR collapse the three form bits into a single `Form` enum (`PerComponent | EntityId | Batch | Relation`) and validate once based on the dispatch-chain priority. The Form would then describe what dispatch will actually do, not what the signature is compatible with.
- OR add a precondition assertion that at most one form bit is set in a non-catch-all signature.

## Suggested area

`engine/system/`

## Suggested model

`[sonnet]` — refactor/diagnostics polish, not architectural.

Linked: PR #1122 (T-334).

jakildev:
## Summary
- Reorders the four `validateConcurrencyForAccess` `IR_ASSERT` checks from entity-id/batch/relation/main-thread to **relation → batch → entity-id → main-thread** (most-specific-first)
- A variadic catch-all tick `[](auto&&...)` satisfies all three form probes simultaneously; the old order caused the less-specific "EntityId tick parameter" diagnostic to fire when "relation tick form" was the correct diagnosis
- Adds `CatchAllWithRelationParamsFatalsAtRegistration` test to pin the catch-all + RelationParams + PARALLEL_FOR rejection path

## Test plan
- [x] All 29 existing `SystemConcurrencyValidator` / `CreateSystemValidatorTest` / `ParallelDispatchFixture` tests pass
- [x] New `CatchAllWithRelationParamsFatalsAtRegistration` test verifies `EXPECT_THROW(std::runtime_error)` for the catch-all + RelationParams + PARALLEL_FOR combination
- [x] `isParallelForAcceptable` mirror in test file updated to match new production order (comment documents the intent)

Closes #1125

### DIFF
--- a/engine/system/include/irreden/ir_system.hpp
+++ b/engine/system/include/irreden/ir_system.hpp
@@ -36,21 +36,23 @@ template <typename... Cs> struct ArchetypeFromList<TypeList<Cs...>> {
 
 // T-222 / T-334: validate that a system's compile-time access
 // descriptor is compatible with its requested Concurrency policy. Four
-// rules, distilled from the multithreading epic (#226 §"Layer 4"):
+// rules, ordered most-specific-first so that a variadic catch-all tick
+// (which simultaneously satisfies every probe) emits the most useful
+// diagnostic. Distilled from the multithreading epic (#226 §"Layer 4"):
 //
+//   - PARALLEL_FOR + isRelationForm_ → FATAL. The relation branch in
+//     `rangedFn` calls `getRelatedEntityFromArchetype` +
+//     `getComponentOptional` on `EntityManager` inside the per-row
+//     loop; those lookups race on the manager's archetype map from
+//     worker threads. Checked first: the most specific form constraint.
+//   - PARALLEL_FOR + isBatchForm_ → FATAL. The per-archetype batch
+//     form consumes the whole column; row-level chunking would
+//     re-enter the body N times with overlapping handles.
 //   - PARALLEL_FOR + usesEntityId_ + !parallelSafe_ → FATAL. The
 //     per-entity-id tick form passes the iterated EntityId to the
 //     body; without an explicit `ParallelSafe` opt-in, the body is
 //     assumed to use the id to mutate non-thread-safe singletons
 //     (`g_entityManager`, render managers, sol2).
-//   - PARALLEL_FOR + isBatchForm_ → FATAL. The per-archetype batch
-//     form consumes the whole column; row-level chunking would
-//     re-enter the body N times with overlapping handles.
-//   - PARALLEL_FOR + isRelationForm_ → FATAL. The relation branch in
-//     `rangedFn` calls `getRelatedEntityFromArchetype` +
-//     `getComponentOptional` on `EntityManager` from inside the per-
-//     row loop, which races on the manager's archetype map from
-//     worker threads.
 //   - PARALLEL_FOR + mainThreadOnly_ → FATAL. The `MainThread` tag is
 //     explicit "do not parallelize", and silently downgrading would
 //     hide the conflict.
@@ -65,11 +67,13 @@ validateConcurrencyForAccess(const std::string &name, Concurrency c, SystemAcces
         return;
     }
     IR_ASSERT(
-        !access.usesEntityId_ || access.parallelSafe_,
-        "System '{}' requested Concurrency::PARALLEL_FOR with an "
-        "EntityId tick parameter but no IRSystem::ParallelSafe tag. The "
-        "id-aware tick form is presumed to look up other entities; tag "
-        "the component pack with `ParallelSafe` after auditing the body.",
+        !access.isRelationForm_,
+        "System '{}' requested Concurrency::PARALLEL_FOR with the "
+        "relation tick form (RelationParams<...> + std::optional<...*> "
+        "in the tick signature). The relation branch resolves the "
+        "related entity and its components via EntityManager lookups "
+        "inside the per-row loop; those manager accesses are not "
+        "thread-safe.",
         name
     );
     IR_ASSERT(
@@ -81,13 +85,11 @@ validateConcurrencyForAccess(const std::string &name, Concurrency c, SystemAcces
         name
     );
     IR_ASSERT(
-        !access.isRelationForm_,
-        "System '{}' requested Concurrency::PARALLEL_FOR with the "
-        "relation tick form (RelationParams<...> + std::optional<...*> "
-        "in the tick signature). The relation branch resolves the "
-        "related entity and its components via EntityManager lookups "
-        "inside the per-row loop; those manager accesses are not "
-        "thread-safe.",
+        !access.usesEntityId_ || access.parallelSafe_,
+        "System '{}' requested Concurrency::PARALLEL_FOR with an "
+        "EntityId tick parameter but no IRSystem::ParallelSafe tag. The "
+        "id-aware tick form is presumed to look up other entities; tag "
+        "the component pack with `ParallelSafe` after auditing the body.",
         name
     );
     IR_ASSERT(
@@ -155,12 +157,10 @@ constexpr SystemId createSystem(
     constexpr SystemAccess accessDescriptor = []() {
         SystemAccess a = deriveAccessFromSignature<FunctionTick, TickComponents...>();
         if constexpr (sizeof...(TickRelationComponents) > 0) {
-            if constexpr (
-                std::is_invocable_v<
-                    FunctionTick,
-                    std::remove_cvref_t<TickComponents> &...,
-                    std::optional<TickRelationComponents *>...>
-            ) {
+            if constexpr (std::is_invocable_v<
+                              FunctionTick,
+                              std::remove_cvref_t<TickComponents> &...,
+                              std::optional<TickRelationComponents *>...>) {
                 a.isRelationForm_ = true;
             }
         }
@@ -218,14 +218,12 @@ template <typename T, typename... Cs> struct MakeMemberTickFn<T, TypeList<Cs...>
             return [p](Cs &...cs) { p->tick(cs...); };
         } else if constexpr (requires(T &s, EntityId id, Cs &...cs) { s.tick(id, cs...); }) {
             return [p](EntityId id, Cs &...cs) { p->tick(id, cs...); };
-        } else if constexpr (
-            requires(
-                T &s,
-                const Archetype &a,
-                std::vector<EntityId> &ids,
-                std::vector<Cs> &...cols
-            ) { s.tick(a, ids, cols...); }
-        ) {
+        } else if constexpr (requires(
+                                 T &s,
+                                 const Archetype &a,
+                                 std::vector<EntityId> &ids,
+                                 std::vector<Cs> &...cols
+                             ) { s.tick(a, ids, cols...); }) {
             return [p](const Archetype &a, std::vector<EntityId> &ids, std::vector<Cs> &...cols) {
                 p->tick(a, ids, cols...);
             };
@@ -439,9 +437,7 @@ void registerPipeline(IRTime::Events systemType, std::list<SystemId> pipeline);
 ///         { globalPosition },            // group 1: serial
 ///         { lifetime },                  // group 2: serial
 ///     });
-void registerPipelineGroups(
-    IRTime::Events event, std::vector<std::vector<SystemId>> groups
-);
+void registerPipelineGroups(IRTime::Events event, std::vector<std::vector<SystemId>> groups);
 
 /// T-224: run the cross-system access-conflict validator across every
 /// registered pipeline group. FATALs on the first conflict, naming
--- a/test/system/system_concurrency_test.cpp
+++ b/test/system/system_concurrency_test.cpp
@@ -279,8 +279,7 @@ TEST_F(CreateSystemValidatorTest, CatchAllWithRelationParamsFatalsAtRegistration
     // T-349: a variadic catch-all tick simultaneously satisfies every
     // signature probe (entity-id, batch-form, relation-form). The
     // validator rules are ordered most-specific-first (relation →
-    // batch → entity-id) so the relation-form assertion fires first;
-    // the log output names the condition (!access.isRelationForm_).
+    // batch → entity-id) so the isRelationForm_ assertion fires first.
     // This test pins that PARALLEL_FOR + catch-all + RelationParams is
     // always rejected — a different shape than the explicit relation-
     // form case in RelationFormParallelForFatalsAtRegistration.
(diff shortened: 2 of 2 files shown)


jakildev (commented):
## Review — T-349: engine/system: order validator rules most-specific-first for catch-all + PARALLEL_FOR

**Verdict:** approve

### Praise

- **Correct diagnosis.** A variadic catch-all `[](auto&&...)` tick simultaneously satisfies all three `std::is_invocable_v` probes (entity-id, batch-form, relation-form), so first-in-source wins. Reordering to relation → batch → entity-id gives the tightest diagnostic for the most ambiguous case — correct and well-reasoned. ✓
- **Test mirror kept in sync.** `isParallelForAcceptable` in the test file is updated to match the new production order (`isRelationForm_` → `isBatchForm_` → `usesEntityId_`). The mirror comment explains the intent. No order drift possible. ✓
- **New test directly pins the fixed scenario.** `CatchAllWithRelationParamsFatalsAtRegistration` exercises exactly the case that was broken — `[](auto&&...)` + `RelationParams<C_VelB>` + `PARALLEL_FOR`. The test verifies `std::runtime_error` is thrown and is a distinct variant from the existing `RelationFormParallelForFatalsAtRegistration` (explicit relation signature vs. catch-all). ✓
- **All 29 existing tests pass.** No regression in `SystemConcurrencyValidator` / `CreateSystemValidatorTest` / `ParallelDispatchFixture`. ✓
- **No runtime behavior change for well-formed systems.** The validator fires only at registration time (inside `createSystem`); valid systems don't hit any assert regardless of rule order. ✓

### Nits

- `test/system/system_concurrency_test.cpp:283` — Comment says "the log output names the condition (!access.isRelationForm_)". The FATAL message text says "relation tick form (RelationParams<...> + std::optional<...*>...)" — it describes the semantic form, not the C++ condition expression. Suggest: "the `isRelationForm_` assertion fires first" to accurately describe what's checked, rather than implying the log output prints the expression `!access.isRelationForm_`.

- [x] All 29 existing `SystemConcurrencyValidator` / `CreateSystemValidatorTest` / `ParallelDispatchFixture` tests pass (per PR body)
- [x] `CatchAllWithRelationParamsFatalsAtRegistration` passes (per PR body)
- [x] `isParallelForAcceptable` mirror updated to new order (per PR body)

Opus recheck required: PR touches `engine/system/include/irreden/ir_system.hpp`; Opus should confirm the catch-all + multi-flag scenario (isRelationForm_ AND isBatchForm_ AND usesEntityId_ all set simultaneously) is fully covered by the existing test suite and that no valid system accidentally hits the first rule due to the `isRelationForm_` bit being set unexpectedly by the access derivation trait.

jakildev (commented):
## Opus recheck — PR #1159: T-349 engine/system: order validator rules most-specific-first

**Verdict:** approve

Picked up for Opus recheck because the diff lands in `engine/system/include/irreden/ir_system.hpp` and modifies the registration-time validator that every PARALLEL_FOR system passes through. Sonnet's first-pass approve verdict stands; this pass focuses on the two items Sonnet explicitly asked Opus to confirm — full multi-flag coverage and the access-derivation soundness. I also add one ordering-pin nit Sonnet did not catch.

### Sonnet asked Opus to confirm (1): `isRelationForm_` cannot be accidentally set on a non-relation system

Sonnet's worry — that the access-derivation trait might set `isRelationForm_` on a system the user didn't intend — does not survive a read of the source. Verified end-to-end:

- `deriveAccessFromSignature` (`system_access.hpp:307`) **never** sets `isRelationForm_`. The docstring at line 301 explicitly says so: "`isRelationForm_` is NOT set here — the relation form depends on a second template pack (`RelationComponents...`) that cannot be disambiguated alongside `Components...` in a free-function template."
- The bit is set in exactly one place: the constexpr lambda inside `IRSystem::createSystem` at `ir_system.hpp:157-167`, gated by **two independent conditions** that both must hold:
  1. `if constexpr (sizeof...(TickRelationComponents) > 0)` — the caller explicitly passed `RelationParams<X...>` with at least one type.
  2. `if constexpr (std::is_invocable_v<FunctionTick, std::remove_cvref_t<TickComponents> &..., std::optional<TickRelationComponents *>...>)` — the tick is invocable with the exact relation-branch parameter shape.

For a non-relation system (e.g. `[](C_Foo& f) {}` registered with no `RelationParams`), the first guard alone fails. The bit stays false. For a non-relation tick signature with `RelationParams<>` accidentally passed, condition 2's invocable probe fails because the per-component tick can't be invoked with extra `std::optional<...*>` args. The bit stays false. The only paths that flip `isRelationForm_` true are (a) an explicit relation signature `[](C_Foo& f, std::optional<RelComp*> rel) {}` + `RelationParams<RelComp>{}` (correct), or (b) the catch-all `[](auto&&...) {}` + `RelationParams<RelComp>{}` (the case T-349 was written to diagnose). ✓ Sonnet's concern is unfounded.

The pre-existing `RelationFormSerialAcceptedAtRegistration` test (`system_concurrency_test.cpp:259-279`) is the load-bearing companion here: it pins that the SAME explicit relation-form signature that flips the bit under PARALLEL_FOR is accepted cleanly under SERIAL. So the rule reorder cannot accidentally start rejecting valid SERIAL relation systems either — only the PARALLEL_FOR branch is touched by the reorder.

### Sonnet asked Opus to confirm (2): multi-flag scenario coverage

The new `CatchAllWithRelationParamsFatalsAtRegistration` test (`system_concurrency_test.cpp:278-300`) exercises end-to-end the case where all three form bits are set simultaneously: `[](auto&&...) {}` + `RelationParams<C_VelB>{}` + `Concurrency::PARALLEL_FOR`. The catch-all body satisfies every form probe, so:

- `deriveAccessFromSignature`'s `InvocableWithEntityId` and `InvocableWithNodeVectors` probes both return true → `usesEntityId_ = true`, `isBatchForm_ = true`.
- `createSystem`'s relation-form lambda's invocable probe also matches the catch-all → `isRelationForm_ = true`.

Verified by tracing `applyComponent<C_VelA>` (writes `C_VelA` into the write set, no flag effects) then `ProbeSignatureForms` (`system_access.hpp:274-281`) running both `InvocableWithEntityId` and `InvocableWithNodeVectors` against the filtered pack. Both pass for a variadic catch-all. ✓ The test fixture is the right shape.

### Additional finding Opus caught (test does not pin ordering)

The new test asserts `EXPECT_THROW(..., std::runtime_error)` but does not pin **which** assertion fires — the validator has four IR_ASSERTs, all throwing `std::runtime_error` via `engAssert`. The test passes regardless of rule order; the PR's commit message says "the relation-form assertion fires first" but the test does not verify that. A silent re-introduction of the old order (relation moved back to last) would not be caught by `CatchAllWithRelationParamsFatalsAtRegistration` — only the existing `RelationFormParallelForFatalsAtRegistration` (which uses an explicit relation signature, so only `isRelationForm_` is set) would still fail correctly.

A tighter form is one-liner:

```cpp
try {
    IRSystem::createSystem<C_VelA>(
        "TestCatchAllRelationParallelFor", catchAllTick,
        nullptr, nullptr, IRSystem::RelationParams<C_VelB>{}, nullptr,
        Concurrency::PARALLEL_FOR);
    FAIL() << "expected std::runtime_error";
} catch (const std::runtime_error& e) {
    EXPECT_THAT(std::string{e.what()}, testing::HasSubstr("relation tick form"));
}
```

This converts the test from "the multi-flag case throws" into "the multi-flag case throws THE RIGHT diagnostic" — which is the point of T-349. Nit-tier, not a blocker. The comment at line 283 ("the log output names the condition (!access.isRelationForm_)") that Sonnet already flagged is the same drift surface: the log text says "relation tick form (RelationParams<...> + std::optional<...*>…)", not the C++ expression. Fold both into the same nit-fix.

### Ordering correctness on the non-relation catch-all path

One edge case worth noting (not a problem, but worth documenting): a catch-all `[](auto&&...) {}` registered **without** `RelationParams` + `PARALLEL_FOR` will hit `isBatchForm_ = true` and `usesEntityId_ = true` but NOT `isRelationForm_` (the first lambda guard `sizeof...(TickRelationComponents) > 0` fails). Under the new order, the batch-form assertion fires first — which is still a reasonable diagnostic, since the catch-all body genuinely could be either entity-id or batch. The most useful guidance is "write an explicit signature so we can tell," but the batch-form message is no worse than the entity-id message Sonnet's body documented for the relation-params case. ✓ No regression on this path.

### What I did not re-check

- Build / test execution (Sonnet confirmed the 29 existing tests + new test pass; test plan in PR body).
- Naming, `SystemName` enum, `m_` / trailing-`_` discipline (Sonnet's first-pass surface).
- Formatting churn — the rest of the diff is clang-format reflows of the relation-form lambda and `MakeMemberTickFn` `else if constexpr` chains; these are no-op text-only changes.

### Cross-host smoke

PR is `fleet:authored-on-linux`. The change is a pure-CPU compile-time validator reorder with no GPU, no shader, and no runtime behavior change for valid systems — the four IR_ASSERTs short-circuit on the first failure, so reordering only changes the diagnostic text on multi-flag-set failure paths (which only the new test exercises). No render-parity concern. Sonnet's note that the PR is host-agnostic is correct; macOS verification is not warranted.

jakildev:
Addressed feedback: fixed misleading comment at `test/system/system_concurrency_test.cpp:283` — removed the phrase "the log output names the condition (!access.isRelationForm_)" and replaced with "the `isRelationForm_` assertion fires first", which accurately describes what happens rather than implying the FATAL message prints the C++ expression. All 11 concurrency/validator/parallel-dispatch tests still pass.

jakildev:
Addressed feedback: Sonnet nit (comment text at line 283) was already fixed in the prior commit (f7ecebb). Investigated the Opus nit (convert EXPECT_THROW to check e.what() for 'relation tick form'): not implementable — engAssert always throws std::runtime_error("Engine assertion failed"); the detailed diagnostic is logged via spdlog but is not carried in the exception message. The EXPECT_THROW form correctly pins that the multi-flag case always throws; pinning the specific diagnostic would require log-capture infrastructure outside the current test setup. Leaving the test as-is.

jakildev:
Addressed Opus recheck nit: `CatchAllWithRelationParamsFatalsAtRegistration` now pins **which** assertion fires.

The original `EXPECT_THROW(..., std::runtime_error)` form would pass regardless of rule order — a silent reorder back to relation-last would still throw, just from a different assertion. Opus's suggested fix (`EXPECT_THAT(e.what(), HasSubstr("relation tick form"))`) doesn't work directly: `IR_ASSERT` throws the generic `std::runtime_error("Engine assertion failed")` and the formatted diagnostic only reaches spdlog.

Fix: install a temporary `spdlog::sinks::ostream_sink_st` on the GL API logger inside the test via a scoped `SinkGuard`, run the call, and `EXPECT_NE(captured.find("relation tick form"), npos)`. The captured-log substring check pins that the relation-form rule (rule 1 of 4) fires first.

All 939 IrredenEngineTest tests still pass. — opus-worker-2

jakildev (commented):
## Re-verify — T-349: engine/system: order validator rules most-specific-first for catch-all + PARALLEL_FOR

**Verdict:** approve

Re-verify after author addressed nits from prior Sonnet (19:01:56Z) and Opus (19:10:31Z) reviews. Reviewing commits f7ecebb and ec486eb only — prior review covered eed1651 in full.

---

### Prior-review resolution

- ✅ `test/system/system_concurrency_test.cpp:283` (Sonnet nit) — misleading comment implied the FATAL log output prints the C++ expression `!access.isRelationForm_` — fixed at f7ecebb: phrase replaced with "so the isRelationForm_ assertion fires first."
- ✅ `test/system/system_concurrency_test.cpp:CatchAllWithRelationParamsFatalsAtRegistration` (Opus nit) — test only verified the call throws, not which rule fired; a silent rule-order revert would pass the test — fixed at ec486eb: spdlog ostream-sink capture added; `EXPECT_NE(logged.find("relation tick form"), npos)` now pins that the relation-form assertion fires first. (Direct `e.what()` check is impossible since `IR_ASSERT` throws the generic `"Engine assertion failed"` message; the diagnostic only reaches spdlog.)

---

### Verification of ec486eb

The spdlog capture implementation is correct:

- `ostream_sink_st` (single-threaded variant) is appropriate for a unit-test context with no concurrent log writers.
- `SinkGuard` RAII struct correctly removes the temporary sink from `glLogger->sinks()` on destruction via `std::remove` + `sinks.erase`.
- The sink is attached before the `createSystem` call; the log message is written synchronously to the sink before `IR_ASSERT` throws; `EXPECT_THROW` catches the throw internally; `captured.str()` read after `EXPECT_THROW` sees the complete logged output. Ordering is safe. ✅
- `EXPECT_NE` failure message includes the captured log content (`<< logged`), giving actionable diagnostics on failure. ✅
- If `EXPECT_THROW` somehow passes without a throw, `captured.str()` is empty or contains no FATAL entry, so `EXPECT_NE` also fails — both assertions fail together, no silent false-pass. ✅

---

### Test plan

- [x] f7ecebb — comment text at system_concurrency_test.cpp:289 now accurately describes behavior (per author comment at 19:11:20Z; all 11 concurrency/validator tests pass)
- [x] ec486eb — spdlog capture pins relation-form rule; all 939 IrredenEngineTest tests pass (per author comment at 19:25:20Z)

🤖 Reviewed by Claude Sonnet 4.6 (sonnet-reviewer role, post-Opus nit-closure re-verify)

Opus recheck not required.

PR #1123, opened 2026-05-23
T-333: engine/system: pre-resolve component vectors on main thread for PARALLEL_FOR

jakildev:
## Stack context

Stacked on: https://github.com/jakildev/IrredenEngine/pull/1097 (T-222)

Closes the Opus recheck nit-1 from #1097 — under `Concurrency::PARALLEL_FOR` dispatch the worker `rangedFn` called `getComponentData<>` → `getComponentType<>` → `m_pureComponentTypes::operator[]` from every chunk. `std::unordered_map::operator[]` is non-const; per the standard, concurrent calls on the same container are a data race even when no insertion happens. Worked in practice on libstdc++/libc++/MSVC because operator[] on an existing key is hash + bucket walk + reference return with no mutation — but it is latent UB that should be cleaned up before T-223 (Lua DSL opt-in) and T-225 (worker-side deferred mutations) broaden the rollout.

## Summary

- Replace `rangedFunctionTick_` (`void(ArchetypeNode*, int, int)`, called per worker chunk) with `prepareRangedTick_` (`std::function<void(int,int)>(ArchetypeNode*)`, called once on the main thread per dispatched node).
- The binder resolves `getComponentData<Components>(node)` on the main thread and returns a `void(begin, end)` worker closure that captures the resolved vector refs by value. Workers iterate captured refs without ever touching EntityManager state.
- `SystemManager::executeSystem`'s PARALLEL_FOR branch calls the binder once per node, then passes the returned closure to `IRJob::parallelFor`. SERIAL / MAIN_THREAD goes through the same binder for code-path uniformity (one extra `std::function` allocation per dispatched node per tick — bounded to a few hundred per frame, negligible vs frame budget).
- Adds `PrepareRangedTickFn` alias on `C_SystemEvent<TICK>` matching the `TickFn` / `RangedTickFn` style introduced in the rebased T-222 head.
- Updates the `engine/system/CLAUDE.md` Concurrency-policy section to describe the binder.

## Acceptance criteria (from #1105)

- [x] Worker bodies under PARALLEL_FOR no longer call `m_pureComponentTypes::operator[]` — verified by inspection; the returned closure captures vector refs by value and has no `getComponentData` call in its body.
- [x] VELOCITY_3D continues to tick correctly — IRPerfGrid runs 15 s clean.
- [ ] ThreadSanitizer-clean dispatch — deferred until the TSAN harness lands (per the original issue note).

## Test plan

[...]

PR #1160, opened 2026-05-24
T-347: script/codegen: emit per-component tick to unlock PARALLEL_FOR

### LINKED ISSUE #1120: lua-codegen: convert system-emit shape from batch form to per-component to unlock PARALLEL_FOR
## Context

Surfaced during T-223 (Lua codegen + EVAL concurrency integration). T-223 wires the optional \`concurrency = IRSystem.Concurrency.PARALLEL_FOR\` field through both Lua-driven paths:

- **EVAL** — runtime registerSystem shim forces \`MAIN_THREAD\` (sol2 + LuaJIT GC are single-threaded).
- **CODEGEN** — \`registerSystemCb\` captures the field on \`SystemRecord\`; \`emitSystem\` emits the trailing \`Concurrency\` arg into the generated \`IRSystem::createSystem<...>(...)\` call.

The CODEGEN path currently FATALs at registration time when a spec requests \`PARALLEL_FOR\` because the codegen tool emits the **per-archetype batch form**:

```cpp
[]([[maybe_unused]] const IREntity::Archetype& _ir_arch,
   std::vector<IREntity::EntityId>& _ir_codegen_ids,
   std::vector<IRComponents::C_Foo>& _ir_col_Foo) { ... }
```

T-222's \`detail::validateConcurrencyForAccess\` rejects \`PARALLEL_FOR + isBatchForm_\` ([\`engine/system/include/irreden/ir_system.hpp:68-75\`](https://github.com/jakildev/IrredenEngine/blob/master/engine/system/include/irreden/ir_system.hpp#L68-L75)) — \"the batch form consumes the whole entity column; row-level chunking would re-enter the body with overlapping data\". Same FATAL a hand-written C++ system would hit; T-223 just propagates the policy and the validator does the rest.

## Scope

Convert the codegen tool's system emit shape from batch form to **per-component** form so a CODEGEN system can opt into PARALLEL_FOR. Concretely:

- The Lua source's canonical \`for i = 0, arch.length - 1 do ... arch.C_Foo:at(i) ...\` loop becomes the implicit per-entity iteration. The body inside the loop becomes a per-component lambda:

  ```cpp
  [](IRComponents::C_Foo& _ir_row_Foo, IRComponents::C_Bar& _ir_row_Bar) {
      // translated body — :at(i) → _ir_row_X; :setAt(i, X.new(...)) → _ir_row_X = X(...)
  }
  ```

- The DSL parser (\`cmake/lua_codegen/system_dsl.cpp\`) already recognizes the canonical for-loop and the \`:at(i)\` / \`:setAt(i, ...)\` column ops, so the lowering is structural — recognize \`local s = arch.C_Foo:at(i)\` as a row binding for \`_ir_row_Foo\`, \`arch.C_Foo:setAt(i, ...)\` as a row write, and drop the outer for-statement.
- Once converted, the trailing \`/* concurrency */\` arg the T-223 emit already produces would unlock without further changes.

## Acceptance criteria

- \`lua_perf_grid\` (CODEGEN, \`concurrency = IRSystem.Concurrency.PARALLEL_FOR\`) at 262k entities matches \`perf_grid\` (C++) within ±10%. This is the perf-parity criterion from T-223's original plan.
- Existing CODEGEN tests (\`test/script/lua_system_codegen_test.cpp\`, \`lua_system_coexistence_test.cpp\`) pass against the new emit shape.
- A new CODEGEN test exercising \`concurrency = PARALLEL_FOR\` registers without FATAL and dispatches across worker threads.

## Suggested model

\`[opus]\` — touches the DSL parser + emitter + per-row code translation; non-trivial.

## Gotchas

- The batch-form \`std::vector<EntityId>& _ir_codegen_ids\` slot may have callers (search before deleting). The per-component form has the id-aware overload \`tick(EntityId, Cs&...)\` if a row body needs the id.
- \`arch.entityAt(i)\` (currently a stub in CODEGEN since the DSL forbids dynamic dispatch) maps to the entity-id form; gate behind the same \`ParallelSafe\` rule as hand-written C++ systems.
- T-223 left a \`/* concurrency */ IRSystem::Concurrency::SERIAL\` annotation on every emitted createSystem call — switching the default to keep with a per-component form is mechanical once the lowering lands.

— filed by opus-worker during T-223

jakildev:
## Summary

- Codegen tool now emits the **per-component** tick lambda (`[](C_Foo& _ir_row_Foo, ...)`) instead of the per-archetype batch form. The canonical `for i = 0, arch.length - 1 do ... end` outer loop is dropped at emission time; inner `arch.Comp:at(i)` / `setAt(i, ...)` / `getField(i, "f")` / `setField(i, "f", ...)` column ops lower to per-row `_ir_row_<CompName>` references.
- This is the engine's default tick signature (engine/system/CLAUDE.md "Three valid TICK function signatures") and the only one compatible with `Concurrency::PARALLEL_FOR` — the T-222 `isBatchForm_` validator path is now unreachable from CODEGEN, so the trailing `IRSystem::Concurrency::PARALLEL_FOR` arg that T-223 already threads through finally satisfies registration.
- Bodies that don't fit the canonical shape (multiple top-level statements, non-canonical loop bound, `arch.length` outside the bound, column-op index that isn't the loop variable) are rejected at codegen time with a message pointing at the requirement. EVAL stays available via `mode = "eval"` for bodies that need a different shape.
- `IRLuaPerfGrid`'s `LuaWaveTick` opts in via `concurrency = IRSystem.Concurrency.PARALLEL_FOR`. New CODEGEN test fixture + test (`CodegenParallelInc` / `ParallelIncRegistersAndUpdatesEveryRow`) exercises the unblocked path.
- Updates `engine/script/CLAUDE.md` "CODEGEN v1 caveat" paragraph that asserted the opposite of the new behavior.

## Test plan

- [x] `fleet-build --target ir_lua_codegen` — codegen tool builds clean.
- [x] `fleet-build --target IrredenEngineTest` — full test suite builds.
- [x] `fleet-run --timeout 30 IrredenEngineTest --gtest_filter='LuaSystemCodegenTest*:LuaSystemCoexistence*'` — 11/11 pass (10 prior + 1 new `ParallelIncRegistersAndUpdatesEveryRow`).
- [x] `fleet-run --timeout 180 IrredenEngineTest` — 939/939 pass.
- [x] `fleet-build --target IRLuaPerfGrid` — demo builds; generated header confirmed to emit per-component lambda + `/* concurrency */ IRSystem::Concurrency::PARALLEL_FOR`.
- [x] `fleet-run IRLuaPerfGrid --auto-screenshot 30` — launches with 262k entities + PARALLEL_FOR, runs through both auto-screenshots, exits cleanly. No FATAL at registration.
- [ ] Hardware-fingerprinted perf-parity vs C++ `perf_grid` (acceptance ±10%) — best done via `perf_grid_matrix` on the same host; flagged as a follow-up measurement once the harness is wired against both demos.

Closes #1120

### DIFF
--- a/cmake/lua_codegen/system_dsl.cpp
+++ b/cmake/lua_codegen/system_dsl.cpp
@@ -890,10 +890,27 @@ struct Emitter {
     const std::vector<ComponentSchema> &registry_;
     std::unordered_map<std::string, Symbol> symbols_;
 
+    // T-347: per-row emission mode. When non-empty, the emitted lambda is
+    // the per-component form `[](C_X& _ir_row_X, ...)` rather than the
+    // batch form `[](Archetype&, vector<EntityId>&, vector<C_X>&, ...)`,
+    // and column-op index expressions that match `loopVarName_` lower to
+    // the per-row reference `_ir_row_X` (or `_ir_row_X.field_`) instead
+    // of `_ir_col_X[i]`. Indexing with any other expression is a parse
+    // error — per-row form has no column vector to scatter-index into.
+    std::string loopVarName_;
+
     Emitter(const std::string &file, const std::vector<ComponentSchema> &registry)
         : file_(file), registry_(registry) {}
 
     void writeIndent() { for (int i = 0; i < indent_; ++i) out_ << "    "; }
+
+    // True when this expression is exactly the per-row loop variable
+    // reference (used to validate column-op index args in per-row mode).
+    bool isLoopVarRef(const Expr &e) const {
+        return !loopVarName_.empty() &&
+               e.kind_ == ExprKind::NAME_REF &&
+               e.name_ == loopVarName_;
+    }
 
     const ComponentSchema *findComponent(const std::string &name) const {
         for (const auto &c : registry_) if (c.name_ == name) return &c;
@@ -1016,6 +1033,15 @@ struct Emitter {
                 // arch.length is a special case
                 if (e.receiver_->kind_ == ExprKind::NAME_REF &&
                     e.receiver_->name_ == "arch" && e.field_ == "length") {
+                    if (!loopVarName_.empty()) {
+                        // T-347: per-row form has no archetype column to size — the
+                        // canonical `for i = 0, arch.length - 1` outer loop is dropped
+                        // and the row count is implicit in the engine's per-row dispatch.
+                        fail(file_, e.line_,
+                             "`arch.length` is only valid as the upper bound of the "
+                             "canonical `for i = 0, arch.length - 1 do` loop; CODEGEN "
+                             "bodies have no archetype-length read outside that loop");
+                    }
                     out_ << "static_cast<std::int32_t>(_ir_codegen_ids.size())";
                     return ExprType::INT32;
                 }
@@ -1087,6 +1113,17 @@ struct Emitter {
                              "` is not declared via `IRComponent.register(...)` "
                              "(CODEGEN system bodies only support Lua-defined components)");
                 }
+                if (!loopVarName_.empty()) {
+                    if (!isLoopVarRef(*e.args_[0])) {
+                        fail(file_, e.line_,
+                             "`arch." + e.componentName_ +
+                                 ":at(...)` index must be the canonical loop variable `" +
+                                 loopVarName_ + "` (CODEGEN bodies dispatch per-row; "
+                                 "computed indexes have no per-row meaning)");
+                    }
+                    out_ << "_ir_row_" << e.componentName_;
+                    return ExprType::COMPONENT;
+                }
                 out_ << "_ir_col_" << e.componentName_ << "[";
                 emitExpr(*e.args_[0]);
                 out_ << "]";
@@ -1103,6 +1140,16 @@ struct Emitter {
                 if (!field) {
                     fail(file_, e.line_,
                          "component `" + schema->name_ + "` has no field `" + e.fieldNameLiteral_ + "`");
+                }
+                if (!loopVarName_.empty()) {
+                    if (!isLoopVarRef(*e.args_[0])) {
+                        fail(file_, e.line_,
+                             "`arch." + e.componentName_ +
+                                 ":getField(...)` index must be the canonical loop variable `" +
+                                 loopVarName_ + "`");
+                    }
+                    out_ << "_ir_row_" << e.componentName_ << "." << e.fieldNameLiteral_ << "_";
+                    return fieldTypeToExprType(field->type_);
                 }
                 out_ << "_ir_col_" << e.componentName_ << "[";
                 emitExpr(*e.args_[0]);
@@ -1202,6 +1249,18 @@ struct Emitter {
                                  "component `" + s.assignTarget_.componentName_ +
                                      "` is not declared via `IRComponent.register(...)`");
                         }
+                        if (!loopVarName_.empty()) {
+                            if (!isLoopVarRef(*s.assignTarget_.indexExpr_)) {
+                                fail(file_, s.line_,
+                                     "`arch." + s.assignTarget_.componentName_ +
+                                         ":setAt(...)` index must be the canonical loop "
+                                         "variable `" + loopVarName_ + "`");
+                            }
+                            out_ << "_ir_row_" << s.assignTarget_.componentName_ << " = ";
+                            emitExpr(*s.assignRhs_);
+                            out_ << ";\n";
+                            return;
+                        }
                         out_ << "_ir_col_" << s.assignTarget_.componentName_ << "[";
                         emitExpr(*s.assignTarget_.indexExpr_);
                         out_ << "] = ";
@@ -1221,6 +1280,20 @@ struct Emitter {
                             fail(file_, s.line_,
                                  "component `" + schema->name_ + "` has no field `" +
                                      s.assignTarget_.fieldNameLiteral_ + "`");
+                        }
+                        if (!loopVarName_.empty()) {
+                            if (!isLoopVarRef(*s.assignTarget_.indexExpr_)) {
+                                fail(file_, s.line_,
+                                     "`arch." + s.assignTarget_.componentName_ +
+                                         ":setField(...)` index must be the canonical loop "
+                                         "variable `" + loopVarName_ + "`");
+                            }
+                            out_ << "_ir_row_" << s.assignTarget_.componentName_ << "."
+                                 << s.assignTarget_.fieldNameLiteral_ << "_ = static_cast<"
+                                 << cppTypeForFieldType(field->type_) << ">(";
+                            emitExpr(*s.assignRhs_);
+                            out_ << ");\n";
+                            return;
                         }
                         out_ << "_ir_col_" << s.assignTarget_.componentName_ << "[";
                         emitExpr(*s.assignTarget_.indexExpr_);
@@ -1320,6 +1393,43 @@ ParsedBody parseSystemBody(
     return body;
 }
 
+// T-347: identify the canonical per-row loop body. The shape:
+//   for <loopVar> = 0, arch.length - 1 do <stmts> end
+// at the top level of the tick body, as the ONLY top-level statement.
+// Returns the for-statement pointer when matched, else nullptr. The
+// pointer is borrowed from `body`; callers must not outlive it.
+const Stmt *matchPerRowCanonicalLoop(const ParsedBody &body) {
+    if (body.stmts_.size() != 1) return nullptr;
+    const Stmt *s = body.stmts_[0].get();
+    if (s->kind_ != StmtKind::FOR_NUMERIC) return nullptr;
+
+    // Lo: integer literal 0.
+    if (!s->forLo_ || s->forLo_->kind_ != ExprKind::NUMBER_LITERAL ||
+        !s->forLo_->isInt_ || s->forLo_->intValue_ != 0) {
+        return nullptr;
+    }
+
+    // Hi: `arch.length - 1` (BinOp SUB; lhs INDEX(arch, length); rhs int 1).
+    const Expr *hi = s->forHi_.get();
+    if (!hi || hi->kind_ != ExprKind::BINARY_OP || hi->binOp_ != BinOp::SUB) {
+        return nullptr;
+    }
+    const Expr *hiLhs = hi->lhs_.get();
+    const Expr *hiRhs = hi->rhs_.get();
+    if (!hiLhs || hiLhs->kind_ != ExprKind::INDEX ||
+        hiLhs->field_ != "length" ||
+        !hiLhs->receiver_ || hiLhs->receiver_->kind_ != ExprKind::NAME_REF ||
+        hiLhs->receiver_->name_ != "arch") {
+        return nullptr;
+    }
+    if (!hiRhs || hiRhs->kind_ != ExprKind::NUMBER_LITERAL ||
+        !hiRhs->isInt_ || hiRhs->intValue_ != 1) {
+        return nullptr;
+    }
+
+    return s;
+}
+
 void emitSystem(
     std::string &out,
     const SystemRecord &record,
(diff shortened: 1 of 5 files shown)


jakildev (commented):
## Review — T-347: script/codegen: emit per-component tick to unlock PARALLEL_FOR

**Verdict:** approve

### Praise

- **Canonical-form enforcement at codegen time.** `matchPerRowCanonicalLoop` correctly gates on the exact `for i = 0, arch.length - 1 do ... end` shape as the sole top-level statement, and the parser already rejects any explicit step (lines 429–433 `fail()` on a trailing comma). No step-validation gap exists — the DSL grammar takes care of it before the AST is built. ✓
- **All four column ops lower correctly in per-row mode.** `at(i)` → `_ir_row_X`, `getField(i, "f")` → `_ir_row_X.f_`, `setAt(i, v)` → `_ir_row_X = v`, `setField(i, "f", v)` → `_ir_row_X.f_ = static_cast<T>(v)`. Each validates that the index expression is exactly the loop variable and emits the correct per-row form. The loop variable is set once (`emitter.loopVarName_ = perRowLoop->forVar_`) and checked consistently via `isLoopVarRef`. ✓
- **`arch.length` inside the loop body is a hard error.** The per-row `arch.length` guard (lines 1033–1041) ensures no body can read `_ir_codegen_ids` — which is absent from the per-component lambda — via a clear diagnostic. Consistent with the loop-bound-only intent. ✓
- **Lambda signature correctly drops `_ir_arch` and `_ir_codegen_ids`.** Neither is referenced by per-row bodies (the canonical loop upper bound is consumed by the matcher; `arch.length` inside the body is an error). The new `[](C_X& _ir_row_X, ...)` shape matches the engine's per-component tick contract. ✓
- **Test coverage is thorough.** 10 existing CODEGEN/coexistence tests still pass on the new emit shape (confirming all existing fixtures use the canonical loop form). New `ParallelIncRegistersAndUpdatesEveryRow` (256 entities, every row checked post-tick) directly exercises the unblocked path. Full suite 939/939. ✓

### Nits

- `cmake/lua_codegen/system_dsl.cpp:892` — member `loopVarName_` comment leads with `// T-347:`. Per baseline, task references rot and belong in the PR description. Drop the prefix; the remaining explanation of what the flag controls is the non-obvious part that earns the comment.
- `cmake/lua_codegen/system_dsl.cpp:1473` — same pattern: `// T-347: CODEGEN system bodies must follow...`. Drop `T-347:`.
- `creations/demos/lua_perf_grid/main.lua:79` — `-- T-347: the per-component emit shape clears the \`isBatchForm_\` FATAL...` Drop the task-ref prefix. The safety claim ("race-free without further opt-in") is the non-obvious part worth keeping.
- `test/script/lua_system_codegen_fixtures.lua:106` — the 9-line comment block opens with `-- T-347: PARALLEL_FOR opt-in. After the codegen tool was switched from batch-form [...] since T-223 finally satisfies...`. The task history ("was switched from", "T-223 finally satisfies") belongs in the PR body, not in a fixture file that future readers will encounter without that context. Trim to intent: `-- PARALLEL_FOR: verifies registration succeeds and the body runs on every row`.
- `test/script/lua_system_codegen_test.cpp:171` — same pattern: the `T-347:` comment block describes the fix history rather than an invariant. The test name already communicates the intent; the multi-line WHY block can be reduced to a single line or dropped entirely.

- [x] `fleet-build --target ir_lua_codegen` — codegen builds clean (per PR body)
- [x] `fleet-run IrredenEngineTest --gtest_filter='LuaSystemCodegenTest*:LuaSystemCoexistence*'` — 11/11 pass (per PR body)
- [x] `fleet-run IrredenEngineTest` — 939/939 pass (per PR body)
- [x] `fleet-build --target IRLuaPerfGrid` — generated header emits per-component lambda + `PARALLEL_FOR` (per PR body)
- [x] `fleet-run IRLuaPerfGrid --auto-screenshot 30` — 262k entities, exits cleanly, no FATAL at registration (per PR body)

Opus recheck required: PR enables PARALLEL_FOR for all CODEGEN systems; Opus should verify (1) the emitted `[](C_X& _ir_row_X, ...)` lambda shape satisfies `validateConcurrencyForAccess`'s `isBatchForm_` path via the T-348 `prepareRangedTick_` dispatch branch, and (2) the `LuaWaveTick` race-free claim — that reading and writing the same component row via `_ir_row_X` reference across concurrent workers is safe given the engine's per-component parallel dispatch guarantees.

jakildev (commented):
## Opus recheck — PR #1160: T-347 script/codegen: emit per-component tick to unlock PARALLEL_FOR

**Verdict:** approve

Picked up for Opus recheck because the diff enables `PARALLEL_FOR` for every CODEGEN-emitted system and changes the emitted tick lambda's shape — affecting the engine-side validator + dispatch path that every CODEGEN system passes through. Sonnet's first-pass approve verdict stands; this pass focuses on the two items Sonnet explicitly asked Opus to confirm. I also add one race-safety future-proofing concern Sonnet did not catch.

### Sonnet asked Opus to confirm (1): emitted lambda passes `validateConcurrencyForAccess` via the T-348 `prepareRangedTick_` branch

End-to-end trace from emitted lambda shape to the validator + dispatch path:

- The new emit site (`system_dsl.cpp:1518-1524`) produces `[](IRComponents::C_X& _ir_row_X, ...) { ... }` — no `EntityId&`, no `const Archetype&` + `std::vector<...>&`, no `std::optional<...*>`. That's the canonical per-component signature.
- `deriveAccessFromSignature<TickFn, C_X...>()` (`system_access.hpp:308-321`) runs `ProbeSignatureForms` against the filtered pack:
  - `InvocableWithEntityId<TickFn, C_X...>` (`ir_system_types.hpp:269`): requires `EntityId&` as the first arg. The per-component lambda has none → **false**, `usesEntityId_` stays false.
  - `InvocableWithNodeVectors<TickFn, C_X...>` (`ir_system_types.hpp:286-290`): requires `const Archetype&, std::vector<EntityId>&, std::vector<C_X>&, ...`. The per-component lambda takes `C_X&` directly → **false**, `isBatchForm_` stays false.
  - `isRelationForm_` is set later by the `createSystem` wrapper's relation-pack lambda (`ir_system.hpp:155-168`); the codegen emits no `RelationParams<...>` argument, so the first `sizeof...(TickRelationComponents) > 0` guard fails and the bit stays false.
- `validateConcurrencyForAccess` for PARALLEL_FOR (`ir_system.hpp:62-100`): `usesEntityId_` is false (first IR_ASSERT passes), `isBatchForm_` is false (second passes), `isRelationForm_` is false (third passes), no `MainThread` tag in the pack (`mainThreadOnly_` is false; fourth passes). **The validator accepts.** ✓
- Dispatch: `SystemManager::insertTickFunction` lands on the `else` branch (`system_manager.hpp:366`) since `InvocableWithNodeVectors` was false, then in the inner cascade it falls into `else if constexpr (InvocableWithComponents<...>)` (line 400) — building the per-row binder that captures `getComponentData<C_X>(node)` once on the main thread and returns a `void(rangeBegin, rangeEnd)` closure. This is exactly the `prepareRangedTick_` shape T-348 expects (`system_manager.cpp:352`, `prepareRangedTick_ && length > grainSize → IRJob::parallelFor` fan-out; else → SERIAL/MAIN_THREAD binder-then-invoke).

End-to-end verified. The previously unreachable `isBatchForm_` FATAL is now genuinely unreachable from CODEGEN.

### Sonnet asked Opus to confirm (2): `LuaWaveTick` race-free claim

The tick body (`creations/demos/lua_perf_grid/main.lua:83-191`) reads + writes only `arch.LuaWaveState` (the system's sole template component); calls only `math.cos`; uses only stack-local arithmetic state.

After the per-row lowering:

- `local s = arch.LuaWaveState:at(i)` lowers to `auto s = _ir_row_LuaWaveState;` (`system_dsl.cpp:1165-1170` — `ExprType::COMPONENT` path, value-copy, not reference). This preserves Lua's snapshot-read semantics: subsequent `s.X` reads cannot observe a later in-tick write to the row.
- All `s.<field>` reads lower to field reads on the local stack copy.
- `arch.LuaWaveState:setAt(i, LuaWaveState.new(...))` lowers to `_ir_row_LuaWaveState = IRComponents::C_LuaWaveState{...};` (value-assignment through the per-row reference). One write per row.
- `math.cos(...)` lowers to `IRMath::cos(...)` via the intrinsic registry (`system_dsl.cpp:27-37`) — a pure function, thread-safe.

Dispatch-side: the `prepareRangedTick_` binder for `InvocableWithComponents` (`system_manager.hpp:400-412`) resolves `getComponentData<C_LuaWaveState>(node)` once on the main thread, captures the `std::vector<C_LuaWaveState>&` by value into the closure, and the PARALLEL_FOR fan-out splits `[0, node->length_)` into non-overlapping `[rangeBegin, rangeEnd)` chunks across workers. T-333 (PR #1123) is what makes this safe — pre-resolving on the main thread keeps workers off the `EntityManager` archetype map. **No two workers ever see the same row.** ✓ The race-free claim holds for `LuaWaveTick`.

### Additional finding Opus caught (race-safety enforcement is implicit, not asserted)

The race-free claim for `PARALLEL_FOR + CODEGEN` rests on three load-bearing implicit invariants that the diff does not pin with an assertion or test:

1. The DSL grammar bans `getComponent`/`setComponent`/`createEntity`/`removeEntity` calls — i.e. the body cannot reach into `g_entityManager` or other systems' rows. The parser enforces this by simply not accepting those tokens. Currently solid.
2. The DSL grammar bans `arch.Comp:at(i)` / `setAt(i,...)` / `getField(i,...)` / `setField(i,...)` with any index expression except the canonical loop variable (`system_dsl.cpp:1116-1124, 1144-1152, 1252-1260, 1284-1292`). This blocks scatter-writes that could race across workers. Solid.
3. **Every entry in `kIntrinsicRegistry` is a pure function** (`system_dsl.cpp:25-37` — 9 entries, all `math.*` / `IRMath::*` math primitives). Currently true. But a future PR adding a stateful intrinsic — e.g. `IRRand.next()` with a shared seed, a `log.warn(...)` that writes to a shared sink, or any I/O helper — would silently make `PARALLEL_FOR + CODEGEN` racy without any validator complaint.

**Recommendation (not blocking, follow-up issue worthwhile):** add a `parallelSafe_` bit to the `Intrinsic` struct (`system_dsl.hpp:314`), default it to true for the existing 9 entries, and have `emitSystem` FATAL at codegen time when a CODEGEN system asks for `PARALLEL_FOR` but its body calls any non-parallel-safe intrinsic. The existing emission-time validation pattern (the `matchPerRowCanonicalLoop` failure path) is the right home for it. The whitelist tax is small (one bit per intrinsic), the invariant becomes explicit, and the failure mode shifts from "silent race in production" to "codegen-time error pointing at the offending intrinsic." Worth filing.

### One nit Sonnet didn't surface

`system_dsl.cpp:1034` carries another `// T-347:` task-ref prefix Sonnet's list missed. Same drop-the-prefix treatment as Sonnet's four other findings — fold into the same nit-fix.

### What I did not re-check

- Build / test execution (Sonnet ran the full 939-test suite + the new `ParallelIncRegistersAndUpdatesEveryRow` test + `IRLuaPerfGrid --auto-screenshot 30`; results documented in the PR body).
- Naming, `SystemName` enum, `m_` / trailing-`_` discipline, anonymous-namespace-in-header check (Sonnet's first-pass surface).
- Codegen DSL grammar correctness for the new `matchPerRowCanonicalLoop` matcher (Sonnet validated this against the existing canonical-form fixtures + the parser's step-rejection logic).
- Doc text accuracy in `engine/script/CLAUDE.md` (Sonnet confirmed the "CODEGEN v1 caveat" paragraph now matches the new behavior).

### Cross-host smoke

PR is `fleet:authored-on-linux`. The change is pure-CPU codegen + a dispatch shape that re-uses the engine's existing `InvocableWithComponents` row binder (in production for every C++ per-component system, including those that already run under PARALLEL_FOR on Linux today). No GPU touch points, no shader involvement. Standard test-suite pass on `macos-debug` is the right follow-up level — no render-parity concern.

jakildev:
Addressed nits: dropped all `T-347:` task-ref prefixes from 6 comments across `cmake/lua_codegen/system_dsl.cpp` (×3), `creations/demos/lua_perf_grid/main.lua`, `test/script/lua_system_codegen_fixtures.lua` (8-line history block → 1 intent line), and `test/script/lua_system_codegen_test.cpp` (dropped 8-line block; test name is self-documenting). Build + 11/11 codegen/coexistence tests pass. — sonnet-fleet-2

Issue #1161, opened 2026-05-24
perf: IRPerfGrid UPDATE pipeline dominates frame time (8.6s/frame); reaching 30 FPS needs UPDATE work too


jakildev:
## Context

Filed as a follow-up to #1154 by opus-worker during T-351. While
implementing the adaptive-iteration optimization for
`COMPUTE_LIGHT_VOLUME` (#1154's "Quick win 1"), I profiled IRPerfGrid on
linux-x86_64 (WSLg Mesa-d3d12) and discovered the wall-clock bottleneck
is **not** in render but in the UPDATE pipeline.

## Measurement

`fleet-run --timeout 60 IRPerfGrid --auto-profile 30` on master (before
the T-351 optimization landed), 262,313 entities, profile_report.txt:

```
Frame time: avg=8319.93ms p50=7033.91ms p95=20980.86ms
Update ticks: avg=453.9/frame (per render frame)

Per-system timing (top by total over 30 render frames):
  UPDATE PropagateTransform           99893 ms — 13618 calls — 7.33 ms/call
  UPDATE PeriodicIdle                 69567 ms
  UPDATE UpdateVoxelSetChildren       45851 ms
  UPDATE PeriodicIdlePositionOffset   24464 ms
  UPDATE ModifierDecay                19825 ms
  RENDER ComputeLightVolume            5325 ms — 30 calls — 177.5 ms/call (GPU stage 171.8 ms)
  RENDER SingleVoxelToCanvasFirst       440 ms
  RENDER CanvasToFramebuffer            122 ms
  (rest sub-2 ms / call)
```

Per-render-frame totals: **UPDATE ~8.6 s**, RENDER ~0.2 s.

The UPDATE pipeline ticks 454 times per render frame because the game
loop accumulates fixed-step ticks while render is blocked. The
underlying root cause is that PropagateTransform + PeriodicIdle +
UpdateVoxelSetChildren each scan all 262 k entities every tick and
together cost ~17 ms/tick → 17 ms × 454 ticks/frame ≈ 7.7 s of the
8.3 s frame budget.

## Why this matters for #1154's acceptance criteria

#1154 sets the target at "≥30 FPS on linux-x86_64". T-351 implements
the adaptive-iteration optimization for COMPUTE_LIGHT_VOLUME and
reduces that stage by ~25 % (from ~234 ms to ~177 ms on perf_grid),
but the savings are dwarfed by the 8.6 s of UPDATE per frame. Even
[...]

PR #1162, opened 2026-05-24
T-351: render: adaptive COMPUTE_LIGHT_VOLUME propagate iteration count

### LINKED ISSUE #1154: platform-parity: IRPerfGrid ~1 FPS on linux-x86_64 (OpenGL) — likely lighting stage

jakildev:
## Summary

- Replace the fixed-32-iteration propagate dilation chain in
  `system_compute_light_volume.hpp` with an adaptive count driven by the
  gathered lights' max `C_LightSource::radius_` (capped at
  `kLightVolumePropagateIterations`).
- Pair the per-frame iteration count with a per-frame
  `stepFalloff = 1 / iterations` so alpha lands cleanly at 0 on the
  final propagate step. Multi-light scenes with mixed radii inherit the
  max-radius falloff curve until Phase 1c per-light step counts land —
  documented as the known compromise in the gather docstring.
- Right-size the `perf_grid` + `lua_perf_grid` demo lights from
  radius=180 (silently capped at 32 today) to radius=24 so the adaptive
  budget can actually reduce dispatches on those demos.
- Update `kLightVolumePropagateIterations` + `stepFalloff_` docstrings
  to reflect "upper bound used by adaptive iteration count" / "written
  per frame" rather than "global radius cap" / "single global falloff".

## Measured impact

`fleet-run IRPerfGrid --auto-profile 30` on `linux-x86_64` (WSLg
Mesa-d3d12, 262,313 entities) — baseline run before this change shipped:

```
RENDER ComputeLightVolume          5325 ms / 30 calls = 177.5 ms/call
       GPU stage computeLightVolume                = 171.8 ms/frame
       CPU Populate (seed + 32 propagate)          = 176.2 ms/frame
```

After this PR (`perf_grid` light at radius=24 → 24 propagate iterations
instead of 32, plus the per-frame stepFalloff): expected ~75 % of the
baseline propagate cost ≈ **~132 ms/frame ComputeLightVolume**
(~45 ms saved per frame). Shape_debug (radius=30 → 30 iters) sees a
~6 % stage reduction; other small-radius demos see proportionally
more.

## Wall-clock FPS — partial against #1154 acceptance

Issue #1154 sets the target at "≥30 FPS on linux-x86_64". This PR
trims the dominant render-side GPU stage but **does not** hit that
target on its own: `Frame time: avg=8319 ms, p95=20980 ms`, dominated
by UPDATE (PropagateTransform + PeriodicIdle + UpdateVoxelSetChildren
at 262 k entities, ~8.6 s of the 8.3 s frame total via fixed-step
catch-up). See follow-up issue #1161 for the full diagnosis — that
work belongs to a separate scoped PR.

The optimization here is real and generally applicable (any demo whose
gathered lights' max radius < `kLightVolumePropagateIterations` sees
proportional savings); shipping it independently is the right scope
boundary.

## Test plan

- [x] `fleet-build --target IRPerfGrid IRLuaPerfGrid IrredenEngineTest` clean.
- [x] `fleet-run IrredenEngineTest --gtest_filter='PerCanvasLightScopeTest*'` — 4/4 pass against the new 5-arg `gatherLightSources` signature.
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — visual parity vs `origin/master` (radius=30 lights now propagate 30 iters instead of 32; barely perceptible).
- [x] `fleet-run IRPerfGrid --auto-screenshot 2` — blue glow renders correctly at the smaller (radius=24) demo light.
- [ ] macOS reviewer: confirm Metal backend behaves identically — `c_propagate_light_volume.metal` already reads `params.stepFalloff` from the UBO, so the per-frame write should flow through without shader-side changes.

## Screenshots

Captured via the `attach-screenshots` skill against `IRShapeDebug` (the
canonical render-regression target). Light reach in the green emissive
volume drops slightly (radius 32 → 30); diff is intentionally subtle.

<details>
<summary>zoom1_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom2_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom1_odd_offset</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom8_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_offset_3_5</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom16_lod_all_visible</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

Closes #1154


jakildev (commented):
## Review — T-351: render: adaptive COMPUTE_LIGHT_VOLUME propagate iteration count

**Verdict:** needs-fix

### Needs-fix

- `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:201-203` — Early return from `gatherLightSources()` exits before the `outMaxRadius` cap at lines 231-233, so if ≥256 lights are present and any have `radius_ > kLightVolumePropagateIterations`, `outMaxRadius` (and thus `propagateIterations_`) will exceed the documented upper bound of `kLightVolumePropagateIterations`. The propagate loop then dispatches more iterations than the budget allows — violating the documented contract ("capped at `kLightVolumePropagateIterations`" in the gather docstring). Fix: clamp `maxRadius` in the caller before assigning `propagateIterations_`:
  ```cpp
  propagateIterations_ =
      (count > 0 && maxRadius > 0)
          ? IRMath::min(maxRadius, kLightVolumePropagateIterations)
          : kLightVolumePropagateIterations;
  ```
  Alternatively, apply the cap inside the early return in `gatherLightSources()`:
  ```cpp
  if (out.size() >= kLightVolumeMaxSources) {
      if (outMaxRadius > kLightVolumePropagateIterations)
          outMaxRadius = kLightVolumePropagateIterations;
      return static_cast<std::uint32_t>(out.size());
  }
  ```

### Nits

- `test/render/per_canvas_light_scope_test.cpp:147-157` — The new `EXPECT_EQ(maxRadius, 0)` test pins the no-lights path (good), but no test covers the positive case: a gathered light with `radius_ = 8` (the fixture's default) should yield `maxRadius == 8`, and a light with `radius_ = 255` should yield `maxRadius == 32` (capped). Adding one test for each path would catch the early-return cap bug automatically in CI.

### Praise

- **Correct per-frame stepFalloff invariant.** `params_.stepFalloff_ = 1.0f / propagateIterations_` is written unconditionally before `paramsBuf_->subData` and before the UBO bind at line 333, so the propagate shader always reads the value matched to the current iteration count. Alpha lands cleanly at 0 on the final step. ✓
- **Upload ordering is correct.** `propagateIterations_` and `params_.stepFalloff_` are set before `paramsBuf_->subData` and the subsequent `bindBase` call, so the GPU always sees the current-frame values. ✓
- **Pre-initialized `propagateIterations_` is a safe fallback.** Default-initialized to `kLightVolumePropagateIterations` so a first-frame dispatch (before the upload phase runs) correctly behaves like the old fixed-32 behavior. ✓
- **Honest scope statement in PR body.** The PR body correctly identifies that the wall-clock FPS target from #1154 is dominated by the UPDATE pipeline at 262 k entities, not this render-side stage, and defers that work to a follow-up issue (#1161). Good engineering discipline.
- **Screenshot evidence.** Before/after pairs across 7 shot labels confirm visual parity on `IRShapeDebug`; the subtle light-reach reduction (radius 32→30 for shape_debug) is called out explicitly.

- [ ] Add a `PerCanvasLightScopeTest` case that creates one light with radius=8 and verifies `maxRadius == 8` after gather, and another with radius=255 and verifies `maxRadius == 32` (cap applied). These also confirm the early-return fix.
- [x] `fleet-run IrredenEngineTest --gtest_filter='PerCanvasLightScopeTest*'` — 4/4 pass (per PR body; pre-fix)
- [x] `fleet-run IrredenEngineTest` — 939/939 pass (per PR body)
- [x] `fleet-build --target IRPerfGrid IRLuaPerfGrid IrredenEngineTest` — clean (per PR body)
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — visual parity confirmed (per PR body + screenshots)
- [ ] macOS reviewer: confirm `c_propagate_light_volume.metal` reads `params.stepFalloff` from the UBO (already present since Phase 1a) and runs correctly at the new adaptive iteration count.

Opus recheck required: diff touches `engine/render/include/irreden/render/ir_render_types.hpp` and the core GPU propagate dispatch loop in `system_compute_light_volume.hpp`; Opus should verify the early-return cap bug is fixed correctly and confirm the per-frame UBO upload ordering invariant holds under the new per-frame `stepFalloff_` write.

jakildev (commented):
## Opus recheck — PR #1162: T-351 render: adaptive COMPUTE_LIGHT_VOLUME propagate iteration count

**Verdict:** needs-fix

Picked up for Opus recheck because the diff touches `engine/render/include/irreden/render/ir_render_types.hpp` (the UBO contract shared with GLSL + Metal propagate shaders) and modifies the core GPU dispatch loop in `system_compute_light_volume.hpp`. Sonnet's needs-fix verdict stands; this pass independently confirms the early-return cap bug, verifies the UBO upload ordering invariant Sonnet explicitly asked Opus to check, and adds two small observations Sonnet didn't surface.

### Sonnet flagged: early-return bypasses the propagate-iteration cap — confirmed

On closer read I confirm the bug. End-to-end trace of the path that produces an uncapped `propagateIterations_`:

- `gatherLightSources` (`system_compute_light_volume.hpp:170-241`) loops over light archetype nodes. The inner per-light loop checks **`if (out.size() >= kLightVolumeMaxSources) { return ... }`** at line 205 **before** the push_back + `outMaxRadius` update at lines 226-230.
- The post-loop cap at lines 231-233 — `if (outMaxRadius > kLightVolumePropagateIterations) outMaxRadius = kLightVolumePropagateIterations;` — sits **after** that early return, so any path that exits via saturation skips the cap entirely.
- Concrete trigger: ≥256 gathered (non-DIRECTIONAL, in-bounds) lights where **any of the first 256** has `radius_ > 32`. The 256th light's push_back triggers the radius update, then the 257th iteration hits the early-return — `outMaxRadius` has the uncapped value from one of the first 256 rows.
- Caller (`system_compute_light_volume.hpp:321`): `propagateIterations_ = (count > 0 && maxRadius > 0) ? maxRadius : kLightVolumePropagateIterations;` — no cap on its side either. `params_.stepFalloff_ = 1.0f / static_cast<float>(propagateIterations_)` then locks the shader to the matching falloff. The propagate dispatch loop at line 390 then runs `propagateIterations_` iterations.
- `radius_` is `uint8_t` so the worst case is `propagateIterations_ = 255`, an ~8× regression vs. the documented `kLightVolumePropagateIterations = 32`. Not unbounded, but well past the published contract.

The doc comment Sonnet quoted is load-bearing — `ir_render_types.hpp:442-447` promises the iteration count is "clamps it to this constant" so "pathological per-cell costs on weaker GPU drivers (WSLg Mesa-d3d12, integrated Intel GL) stay bounded." A bypassable cap defeats the exact use case the cap was written for.

**Cleanest fix** (slight preference over Sonnet's two suggestions): clamp at the per-row update site so all return paths are covered with a single line and the cap-after-loop check becomes a no-op:

```cpp
const int r = IRMath::min(static_cast<int>(lights[i].radius_),
                          kLightVolumePropagateIterations);
if (r > outMaxRadius) outMaxRadius = r;
```

Then the post-loop `if (outMaxRadius > kLightVolumePropagateIterations) ...` can be deleted. Either of Sonnet's two patches also works — author's choice. The defensive-clamp at the caller (`IRMath::min(maxRadius, kLightVolumePropagateIterations)`) is fine too if the team prefers defense-in-depth.

### Sonnet asked Opus to confirm: per-frame UBO upload ordering invariant — confirmed

I traced the full upload → bind → dispatch sequence:

1. `params_.worldOriginVoxel_` written (line 301).
2. `gatherLightSources` returns `count` and `maxRadius` (lines 302-308).
3. `params_.lightCount_ = count` (line 310).
4. `propagateIterations_` derived from `count + maxRadius` (line 321).
5. **`params_.stepFalloff_ = 1.0f / static_cast<float>(propagateIterations_)`** (line 322). All four UBO fields are now consistent for the current frame.
6. `lightSourceBuf_->subData(...)` (line 324, conditional on `count > 0`).
7. **`paramsBuf_->subData(0, sizeof(LightVolumeParams), &params_)`** (line 326). UBO uploaded **after** every relevant struct field is written.
8. `paramsBuf_->bindBase(BufferTarget::UNIFORM, kBufferIndex_LightVolumeParams)` (line 333). Bind **after** upload.
9. Clear dispatch → seed dispatch → propagate-loop dispatches all read from the bound UBO.

No write-after-bind race. No iteration mid-frame uses a stale `stepFalloff_` vs `lightCount_` vs `worldOriginVoxel_` pairing. ✓

**Struct layout consistency.** `LightVolumeParams` is `{ int gridSize_, int halfExtent_, int lightCount_, float stepFalloff_, ivec4 worldOriginVoxel_ }` with `static_assert(sizeof(LightVolumeParams) == 32)` pinning std140. Both `c_propagate_light_volume.glsl:65-72` and `metal/c_propagate_light_volume.metal:30-37` declare the matching field order. The PR doesn't add fields — it just changes what value gets written into the existing `stepFalloff_`. No CPU↔GPU drift introduced. ✓

**Default-value safety.** `propagateIterations_ = kLightVolumePropagateIterations` and `stepFalloff_ = 1.0f / 32.0f` in the struct default mirror the cap, so a first-frame dispatch before the upload phase runs (or any future early-out path) behaves exactly like the pre-PR fixed-32 behavior. The PR header comment `ir_render_types.hpp:491-493` documents this explicitly. ✓

### Additional findings Opus caught (nit-tier, fold into the same fix commit)

1. **The cap-promise header comment promises something the code currently breaks.** `system_compute_light_volume.hpp:160-162` — "`outMaxRadius` receives the largest `C_LightSource::radius_` ... **capped at `kLightVolumePropagateIterations`**" — should match whatever shape the fix takes. If the fix moves the clamp inside the per-row update, drop "capped at `kLightVolumePropagateIterations`" from the descriptive line and add an explicit sentence: "All return paths apply this cap." That way the next reader can audit the saturation early-return against a one-line invariant rather than re-deriving it from the implementation.

2. **`int` vs `uint8_t` widening is correct but undocumented.** `outMaxRadius` is `int`, `lights[i].radius_` is `uint8_t`. The `static_cast<int>` at line 227 is needed to avoid an `IRMath::min(uint8_t, int)` overload mismatch in the suggested fix, and to keep the post-cap comparison well-defined. Not a bug — worth a one-line comment if the fix lands at the per-row site, since the cast at the comparison site is otherwise mysterious.

### What I did not re-check

- Build / test execution (Sonnet ran the 939-test suite + `IRShapeDebug --auto-screenshot 10` and confirmed visual parity; results in PR body with screenshot evidence).
- Naming, `SystemName` enum, `m_` / trailing-`_` discipline (Sonnet's first-pass surface — confirmed clean by reading the diff).
- DIRECTIONAL light skip path, the OOB-warning machinery, `kLightVolumeMaxSources` constant (Sonnet noted these are unchanged by this PR).
- Demo-side radius re-tuning (`creations/demos/perf_grid/main.cpp` and `creations/demos/lua_perf_grid/main_lua.cpp` 180→24) — read the diff, the inline comments explain the cap-coupling rationale and the visual parity claim matches the screenshot evidence.

### Cross-host smoke

PR is `fleet:authored-on-linux` and touches `engine/render/` + GPU dispatch + the UBO struct shared with both GLSL and Metal propagate shaders. The struct field set is unchanged so binary layout cannot drift between backends; the only per-frame behavior change is the value written to `stepFalloff_`, which both backends already read from the same UBO slot. Standard `fleet:needs-macos-smoke` after the cap fix lands is sufficient — once the author addresses the bug, a macOS reviewer should run `IRShapeDebug --auto-screenshot 10` on Metal to confirm the adaptive falloff visually matches Linux. No new shader source or buffer binding to verify.

jakildev:
Addressed feedback:

- **Needs-fix**: Added cap inside the early-return path of `gatherLightSources()` (line 201) so both exit paths (early return at `kLightVolumeMaxSources` and normal return) enforce the `kLightVolumePropagateIterations` upper bound on `outMaxRadius`. This matches the approach in the reviewer's alternative suggestion and keeps the cap logic self-contained in the function.
- **Nit**: Added two new `PerCanvasLightScopeTest` cases:
  - `GatherMaxRadiusUncappedForSmallRadius`: radius=8 light → `maxRadius == 8` (no cap).
  - `GatherMaxRadiusCappedAtPropagateIterations`: radius=255 light → `maxRadius == kLightVolumePropagateIterations` (32, cap applied).
- All 940 tests pass. Build clean.

jakildev (commented):
## Opus re-verify — PR #1162: T-351 cap fix in `gatherLightSources` early-return path

**Verdict:** approve

Re-review after fix commit `9e24d2c` (pushed 19:42:39Z, 3 seconds after my prior needs-fix at 19:42:36Z). The fix addresses exactly the bug Sonnet (19:35:37Z) and I (19:42:36Z) both flagged: the `out.size() >= kLightVolumeMaxSources` early return at `gatherLightSources` bypassed the trailing `outMaxRadius` cap, so a scene with ≥256 sources where any of the gathered lights had `radius_ > kLightVolumePropagateIterations` (32) could let `propagateIterations_` exceed the documented budget.

---

### Prior-review resolution

- ✅ **Early-return cap added.** `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:201-203` (current line numbers; was line 188-190 in the diff context) now applies the same cap that the loop-end path at line 233-235 has always applied:
  ```cpp
  if (out.size() >= kLightVolumeMaxSources) {
      if (outMaxRadius > kLightVolumePropagateIterations)
          outMaxRadius = kLightVolumePropagateIterations;
      return static_cast<std::uint32_t>(out.size());
  }
  ```
  Both exit paths now enforce the contract `outMaxRadius ≤ kLightVolumePropagateIterations`. The text is byte-identical to the loop-end clamp, so the two paths cannot drift on the inequality direction or the clamp value.

- ✅ **UBO upload ordering invariant holds (Sonnet's explicit Opus-recheck request).** Re-traced the upload sequence in `tick()` at `system_compute_light_volume.hpp:305-328`:
  1. `gatherLightSources(..., maxRadius)` — fills `lightStaging_`, writes `maxRadius` (now guaranteed `≤ kLightVolumePropagateIterations` from BOTH exit paths).
  2. `params_.lightCount_ = static_cast<int>(count)`.
  3. `propagateIterations_ = (count > 0 && maxRadius > 0) ? maxRadius : kLightVolumePropagateIterations`.
  4. `params_.stepFalloff_ = 1.0f / static_cast<float>(propagateIterations_)`.
  5. `lightSourceBuf_->subData(...)` for the GPULightSource staging buffer.
  6. `paramsBuf_->subData(0, sizeof(LightVolumeParams), &params_)` — full UBO upload, picking up the just-computed `stepFalloff_`.
  7. Later: `paramsBuf_->bindBase(UNIFORM, ...)` and propagate dispatch loop reads `propagateIterations_` (CPU side, capped) for the iteration count.

  The CPU-side `propagateIterations_` and the GPU-side `params_.stepFalloff_` are now both derived from the same capped `maxRadius`, in this order, BEFORE the UBO `subData`. End-to-end consistency holds — both the dispatch count and the per-step alpha falloff agree.

---

### Nit (non-blocking, fold on next touch)

- `test/render/per_canvas_light_scope_test.cpp:167-194` — The two new tests (`GatherMaxRadiusUncappedForSmallRadius`, `GatherMaxRadiusCappedAtPropagateIterations`) create exactly **one** light each, so the inner loop pushes one entry, the `out.size() >= kLightVolumeMaxSources` branch is never reached, and the assertion is hitting the **loop-end** cap path — the path that was already correct pre-fix. The actual bug — the early-return cap — is not exercised by any test. A regression that reverted only the new `if (outMaxRadius > k...) outMaxRadius = k...` block inside the early-return would still pass all 6 PerCanvasLightScopeTest cases. A test that exercises the buggy path would need ≥257 light entities (so the 257th hits the early-return) with at least one of the first 256 carrying `radius_ > 32`. Heavier setup but mechanical: a `makeLightAtIndex` helper + a loop, EXPECT `out.size() == kLightVolumeMaxSources` AND `maxRadius == kLightVolumePropagateIterations`. Not blocking — the fix is correct by inspection and the cap text is identical to the already-tested loop-end path — but the regression-prevention story is weaker than it looks. Worth a follow-up the next time the file is touched.

---

### What I did not re-check

- Sonnet's first-pass surface (naming, `SystemName` enum, `m_` / trailing-`_` discipline, anonymous-namespace-in-header check, hot-path allocation).
- The original PR's `outMaxRadius` plumbing and `propagateIterations_` derivation — covered in Sonnet's 19:35:37Z review and my 19:42:36Z prior recheck.
- GPU-side propagate shader behavior — unchanged by this fix; only the CPU-side cap path was modified.
- Visual parity (`IRShapeDebug --auto-screenshot 10`) — unchanged geometry of the fix (a 2-line CPU clamp) means screenshots from the prior PR body still apply.

### Test plan

- [x] `fleet-build --target IrredenEngineTest` on `linux-debug` — clean (verified in this re-review).
- [x] `fleet-run IrredenEngineTest --gtest_filter='PerCanvasLightScopeTest.*'` — 6/6 pass (verified in this re-review).
- [ ] macOS reviewer to confirm `IrredenEngineTest` passes on Metal at the new head — pure-CPU clamp, no shader/UBO struct change, so binary parity is intact; standard `fleet:needs-macos-smoke` is the right follow-up level.

### Cross-host smoke

PR is `fleet:authored-on-linux`. The fix is 2 lines of CPU clamping plus 2 unit-test cases (single-light); no shader source, no UBO struct field changes, no buffer-binding changes. Both backends already shared the same C++ source and read `stepFalloff_` from the same UBO slot. Standard `fleet:needs-macos-smoke` after merge is sufficient — no per-stage render verification needed.

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). Metal backend: clean build, 7/7 screenshots captured, no shader-compile warnings, clean exit.

PR #1164, opened 2026-05-24
T-349: engine/system: validator FATAL picks wrong rule on catch-all tick + PARALLEL_FOR

### LINKED ISSUE #1125: engine/system: validator FATAL message picks wrong rule on variadic catch-all tick + PARALLEL_FOR

jakildev:
## Summary

Reorders the `validateConcurrencyForAccess` rules from most-specific to least-specific so the most precise diagnostic fires when a variadic catch-all tick signature triggers multiple validator rules simultaneously.

**Root cause:** A `[](auto&&...) {}` catch-all tick passes all form probes (`usesEntityId_`, `isBatchForm_`, `isRelationForm_`). With `PARALLEL_FOR`, all three rules fire — but `IR_ASSERT` stops at the first one in source order (`usesEntityId_`), surfacing the least-specific entity-id message instead of the form the system was actually using.

**Fix:** Reorder to relation → batch → entity-id → mainThread, matching the dispatch-chain specificity (and explicit-intent priority for relation form).

Closes #1125 (filed from Opus recheck on PR #1122, T-334).

## Test plan

- [ ] `fleet-build --target IrredenEngineTest` — clean
- [ ] `fleet-run IrredenEngineTest` — 939/939 pass
- [ ] Manual inspection: the reordered validator rules match the dispatch chain in `system_manager.hpp`'s `insertTickFunction`

### DIFF
--- a/engine/system/include/irreden/ir_system.hpp
+++ b/engine/system/include/irreden/ir_system.hpp
@@ -34,26 +34,35 @@ template <typename... Cs> struct ArchetypeFromList<TypeList<Cs...>> {
     }
 };
 
-// T-222 / T-334: validate that a system's compile-time access
+// T-222 / T-334 / T-349: validate that a system's compile-time access
 // descriptor is compatible with its requested Concurrency policy. Four
 // rules, distilled from the multithreading epic (#226 §"Layer 4"):
 //
-//   - PARALLEL_FOR + usesEntityId_ + !parallelSafe_ → FATAL. The
-//     per-entity-id tick form passes the iterated EntityId to the
-//     body; without an explicit `ParallelSafe` opt-in, the body is
-//     assumed to use the id to mutate non-thread-safe singletons
-//     (`g_entityManager`, render managers, sol2).
-//   - PARALLEL_FOR + isBatchForm_ → FATAL. The per-archetype batch
-//     form consumes the whole column; row-level chunking would
-//     re-enter the body N times with overlapping handles.
 //   - PARALLEL_FOR + isRelationForm_ → FATAL. The relation branch in
 //     `rangedFn` calls `getRelatedEntityFromArchetype` +
 //     `getComponentOptional` on `EntityManager` from inside the per-
 //     row loop, which races on the manager's archetype map from
 //     worker threads.
+//   - PARALLEL_FOR + isBatchForm_ → FATAL. The per-archetype batch
+//     form consumes the whole column; row-level chunking would
+//     re-enter the body N times with overlapping handles.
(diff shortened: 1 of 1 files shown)


jakildev:
Closing as duplicate of #1159, which has completed the full review cycle (Sonnet + Opus review, nit-fix commits f7ecebb and ec486eb, final Sonnet re-verify, fleet:approved). This PR was opened by a subsequent sonnet-fleet-2 iteration that did not detect the existing approved PR.

Issue #1171, opened 2026-05-25
world: rename ChunkDiskPersistence → ChunkVoxelDiskPersistence (T-298 follow-up 4/4)


jakildev:
Split out from #1008 item 4. Originally bundled into T-360 alongside the
`markChunkDirty` API; T-360 lands item 1 only (the load-bearing
opus piece). Items 2 / 3 / 4 are filed as separate issues so the
queue-manager can route them independently.

## Scope

The class saves the voxel-pool slice only; entity manifest +
billboard metadata are explicitly deferred to #199 and E4. The
current name reads as "saves the chunk" and is easy to
misinterpret as covering entity state. Renaming makes the boundary
explicit at every call site and reduces the risk of a future PR
assuming entity state is durable here.

## Suggested model

`[sonnet]` — low-priority renaming pass. Fold into item 1 (#1008)
or item 2 (#1169) if convenient; skipping is fine.

## Related

- #1008 — parent (item 4 of 4).
- #998 / T-298 — the slice that motivates these follow-ups.

Issue #1170, opened 2026-05-25
world: chunk persistence — wire a real in-engine consumer end-to-end (T-298 follow-up 3/4)


jakildev:
Split out from #1008 item 3. Originally bundled into T-360 alongside the
`markChunkDirty` API; T-360 lands item 1 only (the load-bearing
opus piece). Items 2 / 3 / 4 are filed as separate issues so the
queue-manager can route them independently.

## Scope

T-298's code path has no in-engine consumer — no creation
constructs a `ChunkResidencyManager`, no system wires
`persistence_`, `kWorldBoundMax` is still one chunk
(`engine/common/include/irreden/ir_constants.hpp:42`). The only
exercise is the gtest fake-pool integration. The integration risk
on a real `VoxelPoolAllocation` from `RenderManager` is unproven.

## Acceptance

- Pick one consumer to wire end-to-end. Candidates:
  - The voxel editor's "save chunks" path (most natural fit).
  - A new `IRChunkStreamingSmoke` demo under
    `creations/demos/` that forces evict + re-resident on one
    chunk and asserts the round-trip.
- Run the consumer on linux-debug and macos-debug; confirm the
  file lands under `<save>/chunks/`, the round-trip preserves
  voxel data, and clean chunks skip the save.
- Land before E2/E3 (the actual streaming residency logic) so
  E2/E3 can rely on persistence working against a real pool.

## Suggested model

`[sonnet]` — straightforward consumer wiring; the architectural
decision is which consumer to pick.

## Related

- #1008 — parent (item 3 of 4).
- #998 / T-298 — the slice that motivates these follow-ups.

jakildev:
Delivered by PR #1208 (T-372: chunk persistence smoke demo). Merged 2026-05-27.

Issue #1169, opened 2026-05-25
world: chunk persistence — two-level directory split before real-size maps (T-298 follow-up 2/4)


jakildev:
Split out from #1008 item 2. Originally bundled into T-360 alongside the
`markChunkDirty` API; T-360 lands item 1 only (the load-bearing
opus piece). Items 2 / 3 / 4 are filed as separate issues so the
queue-manager can route them independently.

## Scope

`docs/design/world-streaming.md` §"File layout" already calls out
that the flat `chunks/` directory hits ~2 M files in a 4096³ world
worst case, and sketches a `chunks/<x_div_64>/<y_div_64>/...`
two-level split. T-298 deferred the decision pending profiling.

The risk in landing real saves on the flat layout first is that
saves on the flat layout become a back-compat burden once we split.
Better to land the split before any creation generates durable saves
a user cares about.

## Acceptance

- Decide split dimensions (the design's `<x_div_64>/<y_div_64>` is
  the working proposal; one-level by `x_div_N` may be enough on
  ext4/NTFS — needs a quick `mklink` / `mkfile`-style benchmark).
- Update `ChunkDiskPersistence::chunkPath` + `filenameForKey`.
- Migrate the tests' expected filename fragment.
- No migration tooling needed yet (nothing real persists today).

## Suggested model

`[sonnet]` — mechanical refactor with a small benchmark gate.

## Related

- #1008 — parent (item 2 of 4).
- #998 / T-298 — the slice that motivates these follow-ups.

jakildev:
Delivered by PR #1200 (T-371: two-level directory split for chunk persistence). Merged 2026-05-27.

PR #1172, opened 2026-05-25
T-360: world: markChunkDirty API + chunk-mutation routing contract

### LINKED ISSUE #1008: world: follow-ups to T-298 (chunk voxel persistence)
Follow-ups from review of #998 (T-298 chunk disk persistence). The PR
ships a narrow, opt-in voxel-slice eviction store and is fine to merge
as-is; these are the loose ends that should land before any creation
opts into streaming and starts depending on this code path. File as a
single issue; can be split into separate tasks by the queue-manager if
helpful.

## 1. `markDirty(key)` API + central mutation routing  `[opus]`

Today the `dirty_` bit on `ChunkResidencySlot` is set by callers
poking the field directly (the tests do `s->dirty_ = true;` —
`test/world/chunk_persistence_test.cpp:218`). There is no API
gate, no central mutation routing, and no way for the manager to
verify that a system mutating a voxel through the pool slice
flipped the bit.

This is the ECS footgun class of bug: a system that writes to the
slice and forgets to mark dirty will silently skip the save on
eviction, and the chunk will revert to its pre-edit state on
re-resident. The bug only surfaces under streaming load, which
means it will surface after a creation already depends on the
saves being durable.

Acceptance:
- Add `ChunkResidencyManager::markChunkDirty(ChunkKey)` (or
  similar) as the only supported way to flip the bit.
- Route the eventual voxel-mutation API (the push-at-mutation
  upload path from Epic B B5 / #944, or its successor) through
  this call.
- Demote direct `dirty_ = true;` writes in tests to use the new
  API so the API gets exercise.
- Document in `engine/world/CLAUDE.md` that any system writing to
  a chunk-owned `VoxelPoolAllocation` must route the mutation
  through `markChunkDirty`. Cross-link from
  `engine/render/CLAUDE.md` voxel-pool section so the rule is
  visible from the mutation side.

This is the load-bearing follow-up — until it lands, no creation
should opt into `Config::persistence_`.

## 2. Two-level directory split before real-size maps  `[sonnet]`

`docs/design/world-streaming.md` §"File layout" already calls out
that the flat `chunks/` directory hits ~2 M files in a 4096³
world worst case, and sketches a `chunks/<x_div_64>/<y_div_64>/...`
two-level split. T-298 deferred the decision pending profiling.

The risk in landing real saves on the flat layout first is that
saves on the flat layout become a back-compat burden once we
split. Better to land the split before any creation generates
durable saves a user cares about.

Acceptance:
- Decide split dimensions (the design's `<x_div_64>/<y_div_64>` is
  the working proposal; one-level by `x_div_N` may be enough on
  ext4/NTFS — needs a quick `mklink` / `mkfile`-style benchmark).
- Update `ChunkDiskPersistence::chunkPath` + `filenameForKey`.
- Migrate the tests' expected filename fragment.
- No migration tooling needed yet (nothing real persists today).

## 3. In-engine consumer verification before downstream work  `[sonnet]`

T-298's code path has no in-engine consumer — no creation
constructs a `ChunkResidencyManager`, no system wires
`persistence_`, `kWorldBoundMax` is still one chunk
(`engine/common/include/irreden/ir_constants.hpp:42`). The only
exercise is the gtest fake-pool integration. The integration risk
on a real `VoxelPoolAllocation` from `RenderManager` is unproven.

Acceptance:
- Pick one consumer to wire end-to-end. Candidates:
  - The voxel editor's "save chunks" path (most natural fit).
  - A new `IRChunkStreamingSmoke` demo under
    `creations/demos/` that forces evict + re-resident on one
    chunk and asserts the round-trip.
- Run the consumer on linux-debug and macos-debug; confirm the
  file lands under `<save>/chunks/`, the round-trip preserves
  voxel data, and clean chunks skip the save.
- Land before E2/E3 (the actual streaming residency logic) so
  E2/E3 can rely on persistence working against a real pool.

## 4. (Optional) rename `ChunkDiskPersistence` → `ChunkVoxelDiskPersistence`  `[sonnet]`

The class saves the voxel-pool slice only; entity manifest +
billboard metadata are explicitly deferred to #199 and E4. The
current name reads as "saves the chunk" and is easy to
misinterpret as covering entity state. Renaming makes the
boundary explicit at every call site and reduces the risk of a
future PR assuming entity state is durable here.

Low priority — fold in with item 1 if convenient, or skip.

## Related

- T-298 / #998 — the slice that motivates these follow-ups.
- #199 — entity-level world snapshot (the other half of the
  persistence story; chunk file ⊆ snapshot when both land).
- #938 — Epic E parent issue.

jakildev:
## Summary

Lands the load-bearing `[opus]` piece of #1008 (item 1) so creations can safely opt into `ChunkResidencyManager::Config::persistence_` without risking silent save drops.

- `ChunkResidencySlot::m_dirty` is now **private** with `ChunkResidencyManager` as a `friend`; the public surface is `isDirty()` (read) and `ChunkResidencyManager::markChunkDirty(ChunkKey)` (write). A missed mark turns from a silent runtime bug into a compile error.
- Internal manager callers (`attachEntity`, `migrateEntity`) self-route through `markChunkDirty`. The eventual push-at-mutation voxel upload path (Epic B / #944) wires through the same call when it lands.
- Tests demote direct `s->dirty_ = true` writes to `mgr.markChunkDirty(key)` so the API gets exercise. The `isDirty()` reader replaces the test-side field reads.

## Doc updates

- `engine/world/CLAUDE.md` — new "Chunk mutation must route through `markChunkDirty`" section.
- `engine/render/CLAUDE.md` — voxel-pool bullet cross-links to the world-side rule so the renderer-side author of a B5-style push-at-mutation path sees it.
- `docs/design/world-streaming.md` "Dirty tracking + eviction-write" — references the new API and notes the internal `attachEntity` / `migrateEntity` self-routing.

## Out of scope (filed separately)

Items 2 / 3 / 4 of #1008 were bundled into T-360 in TASKS.md, but the issue body tags them as `[sonnet]` and they're independently scoped. Filed as separate issues so the queue-manager can route them on their own cadence:

- #1169 — two-level directory split (`[sonnet]`)
- #1170 — wire a real in-engine consumer end-to-end (`[sonnet]`)
- #1171 — rename `ChunkDiskPersistence` → `ChunkVoxelDiskPersistence` (`[sonnet]`)

## Test plan

- [x] `fleet-build --target IrredenEngineTest` — clean on `linux-debug`.
- [x] `fleet-run IrredenEngineTest` — 940/940 pass (full suite).
- [x] `fleet-run IrredenEngineTest --gtest_filter='ChunkPersistence*:ChunkResidency*'` — 24/24 pass; the persistence integration tests now exercise `markChunkDirty` instead of poking the private field.
- [ ] Reviewer: confirm the private-field + friend approach is the right enforcement level (vs. keep-public + documentation only). The issue says "as the only supported way to flip the bit"; compile-time enforcement matches that intent, but the slot struct is otherwise all-public.

Closes #1008

### DIFF
--- a/docs/design/world-streaming.md
+++ b/docs/design/world-streaming.md
@@ -764,12 +764,20 @@ full content as the upload budget catches up.
 
 ### Dirty tracking + eviction-write
 
-Each `ChunkResidencySlot` carries a `dirty_` bit. The bit is set by:
+Each `ChunkResidencySlot` carries a private dirty bit, set only
+through `ChunkResidencyManager::markChunkDirty(key)` and read via
+`ChunkResidencySlot::isDirty()`. The bit must be set by:
 
 - Any voxel mutation to a voxel owned by the chunk's
   `VoxelPoolAllocation`.
 - Any entity creation, destruction, component change, or migration
   affecting an entity in the chunk's `ownedEntities_`.
+
+The manager's `attachEntity` / `migrateEntity` self-route through
+`markChunkDirty` already; new mutation paths (voxel writes from a
+push-at-mutation upload, future component-write hooks) must do the
+same — there is no other supported way to flip the bit, and a
+missed call drops the save silently.
 
 The bit is consulted at `EVICTING → EVICTED` transition: if dirty,
 schedule a save to disk before deallocating the pool slice. The
--- a/engine/render/CLAUDE.md
+++ b/engine/render/CLAUDE.md
@@ -125,7 +125,16 @@ itself needs regardless of which features a creation enables:
 - GPU resource CRUD (`RenderingResourceManager`).
 - Pipeline execution (frame loop, canvas dispatch, framebuffer flip).
 - Camera, viewport, subdivision mode — the pipeline reads these every frame.
-- Voxel pool allocation (the pool is a device-level concept).
+- Voxel pool allocation (the pool is a device-level concept). When the
+  allocation is a slice owned by a streaming chunk
+  (`IRWorld::ChunkResidencySlot::poolAllocation_`), any system that writes
+  voxels through the slice MUST call
+  `IRWorld::ChunkResidencyManager::markChunkDirty(key)` immediately after
+  the write — without it, eviction silently drops the save and the chunk
+  reverts on re-resident. See
+  [`engine/world/CLAUDE.md`](../world/CLAUDE.md#chunk-mutation-must-route-through-markchunkdirty).
+  Single-chunk creations never see a residency manager and the rule does
+  not apply.
 
 Feature state — anything a creation opts into — belongs in
 `engine/prefabs/irreden/render/`. If the renderer can ship without the feature
--- a/engine/world/CLAUDE.md
+++ b/engine/world/CLAUDE.md
@@ -31,6 +31,34 @@ residency worker pool without changing the surface. Entity-level
 state (components beyond the chunk's voxel pool) belongs to the
 parallel world-snapshot path (#199), not this layer.
 
+### Chunk mutation must route through `markChunkDirty`
+
+> Any code that writes to a chunk-owned `VoxelPoolAllocation` (the
+> slice exposed via `ChunkResidencySlot::poolAllocation_`) MUST call
+> `ChunkResidencyManager::markChunkDirty(key)` immediately after the
+> write. The same rule covers entity attach / detach / migrate when
+> a creation opts into streaming.
+
+The dirty bit is consulted at eviction and by `flushPendingSaves()`;
+a missed `markChunkDirty` call after a real mutation means the save
+is silently skipped and the chunk reverts to its pre-edit state on
+re-resident. This is the [ECS-footgun class](../../.claude/rules/cpp-ecs.md)
+of bug — invisible under single-chunk creations, fires only after
+streaming load surfaces an eviction-then-re-resident cycle.
+
+`ChunkResidencySlot::isDirty()` is the read side; the underlying
+field is private with `ChunkResidencyManager` as a `friend`, so
+`slot->dirty_ = true` no longer compiles. The manager's
+`attachEntity` / `migrateEntity` already self-route through
+`markChunkDirty`; the voxel-mutation routing lands when push-at-
+mutation uploads (Epic B / #944) wire in.
+
+When you add a new mutation path (voxel-pool write, entity move,
+component write within `ownedEntities_`), route it through
+`markChunkDirty`. The cross-link from
+[`engine/render/CLAUDE.md`](../render/CLAUDE.md) at the voxel-pool
+section is the reciprocal pointer for renderer-side authors.
+
 Most code never touches `World` directly. It accesses managers via the
 `IR<Module>::get*Manager()` free functions in each module's `ir_*.hpp`
 header, which reach through the global pointers `World` sets up at
--- a/engine/world/CLAUDE.md
+++ b/engine/world/CLAUDE.md
@@ -42,8 +42,8 @@ parallel world-snapshot path (#199), not this layer.
 The dirty bit is consulted at eviction and by `flushPendingSaves()`;
 a missed `markChunkDirty` call after a real mutation means the save
 is silently skipped and the chunk reverts to its pre-edit state on
-re-resident. This is the [ECS-footgun class](../../.claude/rules/cpp-ecs.md)
-of bug — invisible under single-chunk creations, fires only after
+re-resident. This is the ECS-footgun class of bug — invisible under
+single-chunk creations, fires only after
 streaming load surfaces an eviction-then-re-resident cycle.
 
 `ChunkResidencySlot::isDirty()` is the read side; the underlying
--- a/engine/world/include/irreden/world/chunk_residency.hpp
+++ b/engine/world/include/irreden/world/chunk_residency.hpp
@@ -187,11 +187,12 @@ class ChunkResidencyManager {
     /// rationale (without it a missed `dirty` flip silently drops
     /// the save on eviction and the chunk reverts on re-resident).
     ///
-    /// No-op when @p key is not resident — the manager logs nothing
+    /// No-op when @p key has no slot at all — the manager logs nothing
     /// because the eventual streaming path will sometimes target a
     /// chunk that already evicted between request and write; callers
     /// that need stronger guarantees should check `isResident(key)`
-    /// before mutating.
+    /// before mutating. Note: a slot in LOADING or UPLOADING state
+    /// (not yet resident) will still be marked dirty by this call.
     void markChunkDirty(IRPrefab::Chunk::ChunkKey key);
 
     // ── Entity ownership ─────────────────────────────────────────────
--- a/engine/world/include/irreden/world/chunk_residency.hpp
+++ b/engine/world/include/irreden/world/chunk_residency.hpp
@@ -68,9 +68,19 @@ struct ChunkResidencySlot {
     float distanceVoxels_ = 0.0f;
     std::uint64_t lastTouchedFrame_ = 0;
 
-    // Set by any voxel/entity mutation; consulted at EVICTING → EVICTED
-    // by E6's save path. E1 leaves the writer hooks for E4/E6.
-    bool dirty_ = false;
+    /// True iff the chunk has been mutated since the last successful
+    /// save (or since allocation, for fresh slots). Consulted at
+    /// EVICTING → EVICTED by E6's save path. Set only by
+    /// `ChunkResidencyManager::markChunkDirty` — direct field writes
+    /// are not supported; see engine/world/CLAUDE.md "Chunk mutation
+    /// must route through markChunkDirty" for the contract.
+    bool isDirty() const noexcept {
+        return m_dirty;
+    }
+
+  private:
+    friend class ChunkResidencyManager;
+    bool m_dirty = false;
 };
 
 /// Owns the sparse resident-set; the engine-side API for chunk identity
@@ -148,15 +158,15 @@ class ChunkResidencyManager {
     void requestResident(IRPrefab::Chunk::ChunkKey key, RequestPriority priority);
 
     /// Drop the chunk from the resident set. When `Config::persistence_`
-    /// is wired and the slot's `dirty_` bit is set, the chunk is saved
+    /// is wired and the slot's dirty bit is set, the chunk is saved
     /// to disk before erasure. Pool allocation is currently leaked back
     /// to the global pool's free-list when the allocator supports it
     /// (today's RenderManager pool is bump-style — see
     /// engine/render/CLAUDE.md). E2 introduces the dealloc path.
     void requestEvict(IRPrefab::Chunk::ChunkKey key);
 
-    /// Force-save every resident slot whose `dirty_` bit is set
-    /// (design's "save-all path"). On success the slot's `dirty_` bit
+    /// Force-save every resident slot whose dirty bit is set
+    /// (design's "save-all path"). On success the slot's dirty bit
     /// clears so subsequent eviction skips the redundant save. Slots
     /// stay resident; only the on-disk copy is refreshed.
     ///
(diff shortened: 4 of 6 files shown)


jakildev (commented):
## Review — T-360: world: markChunkDirty API + chunk-mutation routing contract

**Verdict:** approve

### Praise

- **Private field + `friend` is the right enforcement level.** The author's reviewer question is answered definitively: compile-time enforcement (private `m_dirty` + `friend class ChunkResidencyManager`) converts a footgun-class bug — missed dirty mark on eviction silently drops the save, chunk reverts on re-resident — into a compile error. Documentation-only enforcement would rely on every future mutation-path author reading and following the rule; the private field makes the constraint structural. Correct call.
- **New positive-case test assertion** (`EXPECT_TRUE(s->isDirty()) << "markChunkDirty should flip the slot's bit"` at `test/world/chunk_persistence_test.cpp:214`) is a good addition. The pre-existing tests covered the negative path (clean chunk skips save); this confirms the API actually flips the bit.
- **Docstring on `markChunkDirty`** explicitly acknowledges the no-op-on-non-resident rationale ("the eventual streaming path will sometimes target a chunk that already evicted between request and write") — the right defensive stance for a silent no-op over an assert or error, since callers can't always know if eviction raced ahead of the write.

### Nits

- `engine/world/include/irreden/world/chunk_residency.hpp:192-194` — The docstring says "No-op when @p key is not resident" but the implementation is "No-op when @p key has no slot at all." Per `isResident()`, a slot in LOADING or UPLOADING state is not "resident" — but `markChunkDirty` would still mark it dirty since it exists in `m_slots`. In E1 (synchronous transitions), this gap is unreachable in practice. In E3 (async pipeline where LOADING/UPLOADING slots can persist for multiple frames), external code that obtains a key and calls `markChunkDirty` on a mid-transition slot could set the dirty bit before the load completes. Recommend tightening the docstring to "No-op when @p key has no slot at all" to match the actual behavior, or adding an explicit `state_ == RESIDENT` gate as a forward-looking correctness invariant for E3. Nit since E3 is not yet landed.

- [x] `fleet-build --target IrredenEngineTest` — clean on `linux-debug` (per PR body)
- [x] `fleet-run IrredenEngineTest` — 940/940 pass (per PR body)
- [x] `fleet-run IrredenEngineTest --gtest_filter='ChunkPersistence*:ChunkResidency*'` — 24/24 pass (per PR body)

Opus recheck required: diff modifies `engine/world/include/irreden/world/chunk_residency.hpp` (public world-module header) and `engine/world/src/chunk_residency.cpp` — core `engine/world/` module.

jakildev:
Addressed feedback:
- Tightened `markChunkDirty` docstring from 'No-op when @p key is not resident' to 'No-op when @p key has no slot at all' to match the actual `m_slots.find()` guard
- Added note that a slot in LOADING or UPLOADING state will still be marked dirty (E3 context)

Tests: 24/24 pass (ChunkPersistence + ChunkResidency suites)

jakildev (commented):
## Opus recheck — PR #1172: T-360 world: markChunkDirty API + chunk-mutation routing contract

**Verdict:** approve

Picked up for Opus recheck because the diff modifies `engine/world/include/irreden/world/chunk_residency.hpp` (public world-module header) and `engine/world/src/chunk_residency.cpp` — core `engine/world/` module. Sonnet's approve verdict stands; this pass independently confirms the encapsulation is correct, checks for race conditions and hot-path costs, and surfaces two minor observations Sonnet didn't cover.

---

### Sonnet flagged: docstring says "not resident" but implementation checks "no slot at all" — confirmed as nit

On closer read I confirm this is a genuine gap but agree it's a nit for E1. The `requestResident` function transitions LOADING → UPLOADING → RESIDENT synchronously within a single call, so by the time a slot is visible to external code in `m_slots`, its state is always RESIDENT. The intermediate states are unobservable in E1. When E3 lands with async transitions (persistent LOADING/UPLOADING states across frames), either the docstring should tighten to "No-op when @p key has no slot at all" or an explicit `state_ == RESIDENT` guard should gate the bit flip. Until then, unreachable.

### Opus-only checks

**1. Race conditions / concurrency:**
All three dirty-bit paths — `markChunkDirty` (set), `flushPendingSaves` (read + clear), `requestEvict` (read + erase) — execute on the same thread under E1's synchronous model. No atomics or locks needed. The "snapshot-at-schedule-time" comment in `requestEvict:143-145` correctly notes that save-and-erase in a single blocking call prevents mutation between the two. When E3 lifts this into the worker pool, the entire `m_slots` map will need synchronization regardless — `markChunkDirty`'s single hash lookup will be the least of the concerns. No race today.

**2. Dirty-bit lifecycle correctness (three-system trace):**
- **Set:** only via `markChunkDirty` → `it->second.m_dirty = true`; guarded by `m_slots.find(key) == m_slots.end()` → no-op on non-existent slot.
- **Cleared:** `flushPendingSaves` → `s.m_dirty = false` after successful `saveChunk`. On error, bit stays set — correct retry-on-next-flush behavior.
- **Destroyed:** `requestEvict` reads `s.m_dirty`, saves if dirty, then `m_slots.erase(it)` — bit disappears with the slot, no dangling state.
- **Initialized:** `requestResident` allocates the slot; `m_dirty` default-initializes to `false`. Comment at line 127–128 documents this.

All four lifecycle events are consistent. No path leaves a stale dirty bit, and no path misses a transition.

**3. `friend` scope:** `friend class ChunkResidencyManager` in `ChunkResidencySlot` grants private access only to the manager class. No unexpected transitive friendship. The `isDirty()` public accessor is `const noexcept` — no accidental mutation path.

**4. Redundant lookups (minor observation, not blocking):**
`attachEntity` does `m_slots.find(key)` to obtain `it`, then calls `markChunkDirty(key)` which repeats `m_slots.find(key)`. Similarly, `migrateEntity` performs up to 4 hash-map lookups where 2 would suffice (find source, markDirty source, find dest existence check, attachEntity re-finds dest). These are entity-attachment operations — not per-frame hot paths — so the overhead is negligible. When Epic B's push-at-mutation voxel writes wire in (the anticipated high-frequency caller), the caller will likely hold a slot reference already and a `markChunkDirtyBySlot(ChunkResidencySlot&)` overload or similar can skip the lookup. Not blocking — the current API is correct and the optimization can land alongside the high-frequency caller.

**5. Test migration:** All pre-existing tests that previously wrote `s->dirty_ = true` now route through `mgr.markChunkDirty(key)`, and all reads switched to `s->isDirty()`. The new positive-case assertion at `chunk_persistence_test.cpp:214` verifies the contract from the caller's perspective. Coverage is complete for the API shape.

### Cross-host smoke

Pure C++ encapsulation change — private field, public accessor, manager friend. No GPU, shader, or platform-specific code. No `fleet:needs-*-smoke` labels needed.

jakildev (commented):
## Opus recheck — PR #1172: T-360 world: markChunkDirty API + chunk-mutation routing contract

**Verdict:** approve

Picked up for Opus recheck because Sonnet flagged the diff as touching `engine/world/include/irreden/world/chunk_residency.hpp` (public world-module header) and `engine/world/src/chunk_residency.cpp` — core `engine/world/` module that hosts the chunk-streaming contract. Sonnet's `approve` verdict stands. This pass independently verifies the friend/private enforcement, walks the Opus-only checklist for race/lifetime/allocator concerns, and re-evaluates Sonnet's lone Nit on the docstring drift.

### Sonnet flagged: docstring vs. implementation drift on `markChunkDirty` no-op semantics — confirm, keep as Nit

Sonnet flagged that `markChunkDirty`'s docstring says "No-op when @p key is not resident" but the implementation no-ops only when `m_slots.find(key) == m_slots.end()` (no slot at all). On closer read I confirm:

- A slot in `LOADING` / `UPLOADING` / `EVICTING` returns `false` from the codebase's own `isResident()` (`chunk_residency.cpp:67-72` gates on `state_ == RESIDENT`).
- `markChunkDirty` (`chunk_residency.cpp:184-190`) gates only on `find() != end()`, so it WILL mark a `LOADING` / `UPLOADING` / `EVICTING` slot dirty — i.e. it does not match the "not resident" contract the docstring promises.
- In today's E1 synchronous skeleton, no external code can observe a non-RESIDENT slot — `requestResident` runs synchronously through `LOADING → UPLOADING → RESIDENT` before returning, and `requestEvict` runs synchronously through erase. So the drift is currently latent.
- E3's async pipeline lifts those transitions across frames, at which point the drift becomes real. Tightening the doc to "No-op when @p key has no slot at all" is enough for now; if E3 chooses to gate dirty on `state_ == RESIDENT` (to e.g. avoid wasting an extra save on a slot that's mid-LOADING), the gate moves into the implementation later.

Agree with Sonnet: this is a Nit, not needs-fix. The E1 skeleton doesn't expose the drift to any caller.

### Opus-only checklist results

- **GPU buffer lifetime across frames** — N/A. No SSBO/UBO touched; the voxel pool slice is CPU-side (`std::span<C_Voxel>`).
- **Archetype-mutation race during structural-change deferral** — N/A. The PR adds no `addComponent` / `removeComponent` / `removeEntity` calls. `attachEntity` / `migrateEntity` only mutate `ChunkResidencySlot::ownedEntities_` (a plain `std::vector<EntityId>`); no ECS archetype is touched.
- **Race between `flushStructuralChanges` and async GPU readback** — N/A. No readback, no async work.
- **Allocator behavior in long-lived caches** — `m_slots` is bounded by the resident chunk count, and `requestEvict` cleanly calls `m_slots.erase(it)` (`chunk_residency.cpp:158`). The PR adds no new map entry path. No unbounded growth.
- **Hot-path register pressure** — N/A. `markChunkDirty` is a single `find` + bool write, called at most once per mutation event, not from a per-entity tick.
- **Pointer invalidation across `slot()` returns** — Pre-existing invariant ("Do not cache the pointer across mutating calls") is unchanged. `markChunkDirty` does not rehash (it writes through an existing iterator, no `try_emplace` / erase). Safe.
- **Concurrency** — Single-threaded by design today. If E3's residency worker pool gives worker threads write access to `m_dirty`, the bool would need to become `std::atomic<bool>` (or the dirty mark would need to go through a deferred queue). Not this PR's problem; flag is appropriate for the E3 follow-up.

### Praise

- **Private field + `friend ChunkResidencyManager` is exactly the right enforcement level.** Sonnet asked the author's reviewer-question (compile-time enforcement vs. documentation-only); the answer is clearly compile-time. The class of bug being prevented (missed dirty mark → silent save drop → chunk reverts on re-resident → invisible under single-chunk creations) only surfaces post-streaming-cutover, which is precisely the kind of footgun where doc-only enforcement loses to compile-time. The slot is otherwise all-public, so the asymmetric exception is justified by the asymmetric blast radius.
- **Internal self-routing of `attachEntity` and `migrateEntity` through `markChunkDirty`** (`chunk_residency.cpp:200`, `:216`) means the manager doesn't get to be its own footgun. The two routes that already mutated the slot pre-PR now go through the canonical API. Test coverage at `test/world/chunk_persistence_test.cpp:213` exercises the API and asserts the bit flips (`EXPECT_TRUE(s->isDirty())`).
- **Doc cross-link is reciprocal.** `engine/world/CLAUDE.md` "Chunk mutation must route through markChunkDirty" is the canonical statement; `engine/render/CLAUDE.md` voxel-pool bullet links back at the exact spot a renderer-side push-at-mutation author would look. Both docs name the rule the same way and cite the same Epic B / #944 follow-up.
- **Scope discipline.** The author correctly bounded the PR to the dirty-bit-flip surface only. The remaining footgun shape — external code mutating `slot->poolAllocation_.voxels_[i]` via the non-const slot pointer and then forgetting to call `markChunkDirty` — is left explicitly to convention, with the contract stated in `engine/world/CLAUDE.md`. A bigger wrapper (e.g. `MutableSliceView` proxy that auto-marks dirty on commit) would have over-engineered T-360. The right time for that is when push-at-mutation uploads land in Epic B and there's a real consumer for the API to be shaped against.

### Nits

- `engine/world/include/irreden/world/chunk_residency.hpp:190-194` — Docstring on `markChunkDirty` says "No-op when @p key is not resident" but the implementation no-ops only when there's no slot at all (a `LOADING` / `UPLOADING` / `EVICTING` slot would still get marked). In the E1 synchronous skeleton this can't be observed externally, but tighten to match the implementation before async E3:
  ```cpp
  /// No-op when @p key has no slot — the eventual streaming path will
  /// sometimes target a chunk that already evicted between request and
  /// write. Callers that need stronger guarantees should check
  /// `isResident(key)` before mutating.
  ```
  (Same nit Sonnet raised, confirmed.)
- `engine/world/CLAUDE.md` "Chunk mutation must route through `markChunkDirty`" — links `[ECS-footgun class](../../.claude/rules/cpp-ecs.md)` but the rules doc doesn't define an "ECS-footgun class" anchor; the link resolves to the file but the phrase isn't a section header anywhere. Either drop the link-wrap or land an explicit `## ECS-footgun class` section in `cpp-ecs.md` for it to anchor to. Pure-doc nit.

- [x] `fleet-build --target IrredenEngineTest` — clean on `linux-debug` (per PR body)
- [x] `fleet-run IrredenEngineTest` — 940/940 pass (per PR body)
- [x] `fleet-run IrredenEngineTest --gtest_filter='ChunkPersistence*:ChunkResidency*'` — 24/24 pass (per PR body)
- [ ] No additional Opus-mandated verification — the friend/private change compiles or it doesn't, and the two existing tests (`ManagerSavesDirtyChunkOnEvictAndLoadsOnReResident`, `FlushPendingSavesWritesDirtySlotsAndClearsDirty`) cover the contract end-to-end.

### Cross-host smoke

PR does not touch `engine/render/`, `engine/prefabs/irreden/render/`, `engine/render/src/shaders/`, or any `*.glsl` / `*.metal` file. The `engine/render/CLAUDE.md` change is doc-only (cross-link paragraph in the renderer ownership list). No backend code, no shader. Cross-host smoke is not required for this PR — the existing `fleet:authored-on-linux` label is sufficient; no `fleet:needs-macos-smoke` / `fleet:needs-windows-smoke` to add.

jakildev:
Addressed feedback:
- Dropped broken `[ECS-footgun class](../../.claude/rules/cpp-ecs.md)` link-wrap in `engine/world/CLAUDE.md:45` — the target file has no matching anchor and the concept described is a different footgun pattern than what `cpp-ecs.md` documents. Kept as plain text; the prose is clear without the link.
- Docstring on `markChunkDirty` was already tightened in a prior commit (now reads "No-op when @p key has no slot at all" with E3 context note).

— sonnet-fleet-2

PR #1115, opened 2026-05-23
T-330: tools: ir-perf-grid + fingerprinted baselines (sub-task 3 of #1074)

jakildev:
## Summary

- Sub-task 3 of #1074 — the actual perf-gate behavior change on top of T-318 (lock primitives + host probe) and T-329 (ir-build / ir-run).
- **`engine/tools/bench/ir_ref_bench`** — deterministic IRMath hot-path micro-bench (~50 ms target on the calibration host). Prints `{"ms": N, "iters": N, "ops": N}` on stdout; volatile sink defeats `-O3` constant-folding.
- **`engine/tools/bin/ir-perf-grid`** — wraps `scripts/perf/perf_grid_matrix.sh` in `ir-acquire benchmark`, calibrates on demand, captures `ref_ms` at the start of each matrix, splices `{ref_ms, ref_target_ms, host_slug, host_fingerprint, math_sha}` into `manifest.json`. Calibration cache lives at `$XDG_CACHE_HOME/irreden/calibration/<slug>.json`; invalidates on `engine/math/` SHA change or `calibration_max_age_days` (30 default).
- **`scripts/perf/compare_perf_runs.py`** — fingerprint-aware: resolves baselines under `docs/perf/baseline_latest/<head-slug>/`, prints a `host calibration` markdown block citing load factor and chosen weighting. Legacy flat baseline dirs still work for ad-hoc diffs.
- **`scripts/perf/check_regression.py`** — four-branch gate from #1074: same fingerprint + uncontested → raw; same fingerprint + loaded (`ref_ms ≥ 1.20× target`) → normalized; different fingerprint or no baseline → informational pass. `LOAD_FACTOR_TRUST_NORMALIZED` is module-level so the cutoff is tunable in one place.
- **`.github/workflows/perf-gate.yml`** — invokes `ir-perf-grid --quick --label ci`; master-push baseline-update writes under `docs/perf/baseline_latest/<slug>/` with a `host.json` sidecar built from the head's calibration block.
- **Tests** — new `test/tools/normalization_test.sh` (16 assertions across math helpers, fingerprint resolution, host-note classification, and all 5 gate code paths) + extended `test/tools/calibration_test.sh` with `ir_ref_bench` JSON-schema checks.

## Stack context

Stacked on: https://github.com/jakildev/IrredenEngine/pull/1111 (T-329, `claude/T-329-ir-build-run`)

[...]

PR #1177, opened 2026-05-25
T-363: tools: ir-host-probe survives non-exec lspci stub in PATH

### LINKED ISSUE #1074: tools: engine-level concurrency + perf primitives (ir-build / ir-run / ir-acquire / ir-perf-grid)
# tools: engine-level concurrency + perf primitives (ir-build / ir-run / ir-acquire / ir-perf-grid)

## Context

The fleet's concurrent workers, a solo dev on a laptop, and the CI runner are all clients of the same coordination problem: who owns the CPU right now, who owns the GPU, who's currently doing a perf measurement, and how do we compare perf numbers across hosts. Today this lives implicitly in `scripts/fleet/fleet-build` and `scripts/fleet/fleet-run` (named for the noisiest client — the fleet — but used by every client). This task reframes them as **engine-level utilities that the fleet happens to use**, adds the missing concurrency + perf primitives, and lands a shared lock + token budget so concurrent invocations from any source (fleet workers, multiple dev shells, engine + game) coordinate cleanly.

Prospective hardening — we have not yet seen incidents — but performance work is starting to land (#1039 perf-gate, #1049 voxel-pool cap) and the design call is much cheaper now than after the first regression-noise spiral.

## Identity / naming

| Was | Becomes | Reason |
|---|---|---|
| `fleet-build` | `ir-build` | Builds are an engine concern |
| `fleet-run` | `ir-run` | Running an engine binary is an engine concern |
| (new) | `ir-acquire` | Resource coordination is an engine concern |
| (new) | `ir-perf-grid` | Perf measurement is an engine concern |
| (new) | `ir-host-probe` | Hardware fingerprinting is an engine concern |
| `fleet-claim`, `fleet-feedback`, `fleet-worktree-busy-branches` | **unchanged** | Genuinely fleet-only (multi-worker queue / comms coordination) |

Migration: `scripts/fleet/fleet-build` becomes a one-line `exec ir-build "$@"` shim during the rename window. Mechanical.

## Layout

```
engine/tools/
  bin/         # ir-build, ir-run, ir-acquire, ir-perf-grid, ir-host-probe
  lib/         # shared shell helpers
  py/          # ir_compare_perf_runs.py, ir_hardware_probe.py
  bench/       # ir_ref_bench.cpp — synthetic-load mini-bench
  concurrency.toml   # committed engine defaults
  README.md
```

Install hook adds `engine/tools/bin/` to PATH the same way `scripts/fleet/` is exposed today. Game repo gets it for free.

## Three config layers

Defaults → host overrides → env vars. Highest specificity wins.

```toml
# engine/tools/concurrency.toml (committed; engine defaults)
[cpu]
budget = "auto"              # auto = nproc; integer overrides
per_build_max = "budget / max(workers, 1)"  # so 4 workers on 16-core ≈ 4 cores each

[gpu]
exclusive = "auto"           # auto = true on macos+metal; configurable elsewhere

[perf]
exclusive = true             # always; correctness invariant
ref_bench_target_ms = 50
calibration_max_age_days = 30

[concurrency]
workers = "auto"             # auto = read IR_FLEET_WORKERS or default 1
queue_timeout_seconds = 600
```

```toml
# ~/.config/irreden/host.toml (uncommitted; per-host quirks)
[cpu]
per_build_max = 6            # this box thermal-throttles past 6 cores
[gpu]
exclusive = false            # this Linux box can run two GLFW windows
```

```sh
# Env vars (process-level; highest priority)
IR_CPU_BUDGET=8 ir-build IRShapeDebug
IR_FLEET_WORKERS=4 fleet-up
```

## Value determination

| Value | Determined by | Cached where |
|---|---|---|
| **Host fingerprint** | `ir-host-probe` — CPU model, GPU model, OS, RAM → deterministic slug | `~/.cache/irreden/host-fingerprint.json` (invalidated on probe-output checksum change) |
| **CPU budget** | `nproc` → host override → `IR_CPU_BUDGET` | n/a — recomputed per invocation |
| **Per-build cap** | `budget / max(workers, 1)` → host override → `IR_BUILD_JOBS` | n/a |
| **GPU exclusivity** | Platform default (Metal yes, OpenGL configurable) → host override → `IR_GPU_EXCLUSIVE` | n/a |
| **Perf-bench `ref_target_ms`** | Run `ir_ref_bench` 5×, take median | `~/.cache/irreden/calibration/<fingerprint>.json` — invalidated on `calibration_max_age_days` OR when `engine/math/` SHA changes |

The IRMath-SHA-change re-calibration is the validation backstop: when someone speeds up `IRMath::isoToScreen`, the reference bench shifts and old comparisons stop being meaningful — force a re-calibration and the comparator emits `baseline-from-different-ref-version, normalization disabled` until the new baseline lands.

## Scaling with availability — three regimes, no special-casing

**1. Solo dev, single shell, bare command (`ir-build IRShapeDebug` with no env).** `IR_FLEET_WORKERS` unset → defaults to 1 → per-build cap = full budget. **`ir-acquire` tries `flock --nonblock` first** and succeeds instantly; build runs at full `nproc`; perf runs unopposed. Tool is invisible. **Humans pay no concurrency tax in the default case.**

**2. Solo dev, two shells.** Second `ir-build` blocks on CPU slots until first finishes, OR proceeds at half budget if `per_build_max` allows. Second `ir-run --auto-profile` queues behind the first with `ir-acquire: waiting for perf lock (held by pid 12345, started 4s ago)`.

**3. Fleet, N workers.** `fleet-up` exports `IR_FLEET_WORKERS=N`; per-build cap auto-divides to `nproc/N`. Builds run truly concurrently; perf runs serialize; GPU runs serialize where `gpu.exclusive=true`. No worker knows about the others — the engine tools coordinate via the lock dir.

### Fleet behavioral rule: acquire-late, release-early

Documented in `role-opus-worker.md` / `role-opus-architect.md` / `role-sonnet-author.md`:

> Workers acquire locks immediately before the operation that needs them and release immediately after. Never hold a lock across "decide what to do next" / "draft a comment" / "read review feedback." A perf measurement holds the perf lock for the perf-grid run only — not for the simplify pass that follows. Optimize work is the canonical example: claim, measure, release; then iterate without the lock; reclaim only for the next measurement.

The release-on-PID-death property of `flock` makes this self-healing (a crashed worker doesn't poison the lock), but the rule is what keeps queue latency low in healthy operation.

### Heavy-measurement mode: `ir-acquire benchmark`

Optimize / profile / regression-bench work needs the box mostly to itself for the measurement window. New canned mode:

```sh
ir-acquire benchmark -- ./IRPerfGrid --auto-profile 300
```

Acquires **all three** in one shot: `cpu = budget - 1` (leaves one core for OS / terminal / responsiveness), `gpu` exclusive, `perf` exclusive. Releases all on command exit. The benchmark mode is what the `optimize` skill and `ir-perf-grid` use under the hood — workers and humans alike.

This is the lock mechanism the user's optimize-workflow caveat depends on: yes, optimize is bursty and noisy, and yes it should serialize behind a single canonical heavy-measurement lock. `ir-acquire benchmark` is that lock.

## Discovery

```
$ ir-acquire --info
host fingerprint: linux-ryzen-7950x-rtx-4080
cpu budget: 16 (12 in use, 4 free)
  - pid 12345 ir-build IRShapeDebug (4 cores, 22s)
  - pid 12346 ir-build IRPerfGrid (4 cores, 18s)
  - pid 12347 ir-build IRVoxelEditor (4 cores, 11s)
gpu lock: held by pid 12348 (ir-run IRShapeDebug --auto-screenshot, 3s)
perf lock: free
queue depth: 1 (pid 12349 waiting on gpu lock, 2s)
```

Lets a human staring at a stalled worker see exactly what it's waiting on.

## Hardware-fingerprinted baselines

Replace `docs/perf/baseline_latest/<bench>.json` with `docs/perf/baseline_latest/<host_fingerprint>/<bench>.json`. Each baseline carries a `host.json` sidecar with full probe (CPU model, core count, frequency, GPU model, OS version, RAM). Comparator behavior:

- **Same fingerprint** → direct comparison; gate fires normally.
- **Different fingerprint** → reports the regression value but the gate is informational; comparator prints `host mismatch (baseline: <fp_b>, current: <fp_c>); normalized via synthetic-load reference`.
- **No fingerprint match at all** → current run seeds a new baseline; comparator prints `seeded new host baseline`.

CI runs are just another fingerprint (`ci-ubuntu-x64-llvmpipe`); not privileged.

## Synthetic-load normalization

A deterministic `ir_ref_bench` (math-hot-path mini-bench, ~50 ms target, no graphics) runs at the start of every `ir-perf-grid` invocation. Each report stores `(measured_ms, ref_ms)`. Comparator computes `normalized_ms = measured_ms * (ref_target / ref_ms)` where `ref_target` is the host's calibrated median.

Use cases:

- **Same host, different load** — dominant case. Second worker building in background → `ref_ms` reflects it; normalized comparison stays meaningful.
- **Cross-host informational** — pairs with fingerprinting; comparator prints both raw and normalized.
- **Skip when the perf lock was clean** — `ref_ms` is still captured under lock, but the comparator weights raw above normalized when the lock was uncontested. Normalization is a fallback, not the primary signal.

## Wiring

| Tool | Acquires |
|---|---|
| `ir-build --target X` | `cpu min(per_build_max, free_slots)` |
| `ir-run X --auto-profile N` | `benchmark` (cpu=N-1 + gpu + perf) |
| `ir-run X --auto-screenshot N` | `gpu` only |
| `ir-run X` (interactive) | none (human-supervised) |
| `ir-perf-grid` | `benchmark` for the entire matrix duration; runs `ref_ms` first |
| `scripts/render-compare.py` / regression check | none (read-only) |
| `attach-screenshots` skill | `gpu` (covers both before + after passes) |
| `optimize` skill | `benchmark` |
| `.github/workflows/perf-gate.yml` | `benchmark` (no-op on the CI runner; keeps the code path identical) |

## Cross-repo

Game repo invokes `ir-build` etc. directly via PATH. Game-side perf reports land in `<game-repo>/docs/perf/baseline_latest/<fingerprint>/`. Hardware fingerprint is identical (same host), so the engine and game baselines can theoretically share calibration data — but storage stays per-repo (clean separation, no info-isolation violation).

Shared lock dir (`${XDG_RUNTIME_DIR}/irreden/locks/` on Linux, `/tmp/irreden-$USER/locks/` on macOS) means an engine build and a game build coordinate CPU automatically without either repo knowing about the other.

## Validation surface

- **`test/tools/concurrency_test.sh`** — deterministic lock/budget tests. Spawns N background `ir-acquire` calls, asserts the right number block / proceed, kills some mid-flight, asserts slot release on PID death. Runs on every CI build, no GPU needed.
- **`test/tools/calibration_test.sh`** — checks `ir-host-probe` output is parseable, deterministic across two calls.
- **`test/tools/normalization_test.cpp`** — given synthetic inputs, the comparator's normalized regression value matches a known formula. Pure math, no hardware dependency.
- **`engine/tools/bench/ir_ref_bench.cpp`** — built as a regular target; CI sanity check that it completes under 100 ms on `ci-ubuntu-x64-llvmpipe` (catches "someone accidentally made the bench slow").
- **Integration smoke** — CI runs `ir-build IRPerfGrid && ir-run IRPerfGrid --auto-profile 30 && ir-perf-grid compare`. If the lock primitive ever deadlocks, the smoke hangs and we catch it.

## Suggested PR sequence

Three sub-tasks, each independently reviewable:

1. **`ir-host-probe` + `ir-acquire`** — pure new tooling, no behavioral change to existing scripts. The `benchmark` canned mode is part of this PR.
2. **`ir-build` / `ir-run` migration** — port from `fleet-build` / `fleet-run`; leave shims; wire CPU / GPU acquires into existing callers (incl. `attach-screenshots` skill).
3. **`ir-perf-grid` + hardware fingerprinting + synthetic-load normalization** — the actual perf-gate behavior change; layers on (1) + (2).

Plus a doc-only PR adding the "acquire-late, release-early" rule to the three worker role docs.

## Open questions / deferrals

- GPU multi-context on Linux — start strict (`exclusive=true`); relax later if real use case.
- Network / `gh api` rate-limit locks — defer.
- Lock starvation under load — `flock` is roughly-FIFO; not guaranteed fair; acceptable for v1.
- Re-calibration triggers beyond IRMath-SHA — defer until a real signal needs it.

`[opus]` — core infrastructure with cross-tool implications; needs architect review on the wiring decisions and the host-fingerprint format choice.

jakildev:
## Summary
- Broadens `ir_hardware_probe._run()`'s exception catch from `FileNotFoundError` to `OSError`, so a WSL2 host whose PATH contains a `/mnt/c/...` Windows-side stub matching `lspci` no longer crashes the whole probe with `PermissionError: [Errno 13] Permission denied: 'lspci'`. The documented `{model: unknown}` fallback now actually fires.
- Adds a regression test in `calibration_test.sh` that plants a non-executable `lspci` stub at the front of PATH and asserts the probe still emits valid JSON — reproduces the failure mode deterministically on any host (the existing tests would only catch it on WSL2 specifically).

## Why
T-363 (sub-task 1 of #1074) covers `ir-host-probe` + `ir-acquire`. The tools themselves shipped earlier via T-330 (PR #1115). On this fleet host (Ubuntu 24.04 / WSL2) `ir-host-probe --slug` crashed and `calibration_test.sh` failed at the first probe call. That broke any downstream caller — `ir-perf-grid` invocations, the calibration check, baselines comparator — and made the host effectively unsupported for the perf workflow despite being the standard fleet env.

## Test plan
- [x] `bash test/tools/calibration_test.sh` — 9 passed, 0 failed (was: crashes on step [1] with PermissionError)
- [x] `bash test/tools/concurrency_test.sh` — 8 passed, 0 failed (unchanged behavior; ran as a regression check)
- [x] `ir-host-probe --slug` returns a well-formed slug: `linux-x86_64-ryzen-9-5950x-unknown` (GPU correctly "unknown" since lspci isn't actually present)
- [ ] macOS reviewer: probe still works (no behavioral change on macOS — `_run` is only used for `sysctl`/`system_profiler`/`sw_vers`, all built-ins that don't hit this exception path)

Closes #1074

### DIFF
--- a/engine/tools/py/ir_hardware_probe.py
+++ b/engine/tools/py/ir_hardware_probe.py
@@ -35,11 +35,11 @@ def _run(*args, default=""):
             timeout=5,
         )
         return out.stdout.strip()
-    except (OSError, subprocess.TimeoutExpired):
-        # OSError catches FileNotFoundError (binary missing from PATH) and
-        # PermissionError (binary exists but is not exec'able — happens on
+    except (FileNotFoundError, PermissionError, subprocess.TimeoutExpired):
+        # FileNotFoundError: binary missing from PATH.
+        # PermissionError: binary exists but is not exec'able — happens on
         # WSL2 hosts when PATH contains Windows-side stubs that match a
-        # Linux tool name but cannot be invoked from a Linux subprocess).
+        # Linux tool name but cannot be invoked from a Linux subprocess.
         return default
 
 
--- a/engine/tools/py/ir_hardware_probe.py
+++ b/engine/tools/py/ir_hardware_probe.py
@@ -35,7 +35,11 @@ def _run(*args, default=""):
             timeout=5,
         )
         return out.stdout.strip()
-    except (FileNotFoundError, subprocess.TimeoutExpired):
+    except (OSError, subprocess.TimeoutExpired):
+        # OSError catches FileNotFoundError (binary missing from PATH) and
+        # PermissionError (binary exists but is not exec'able — happens on
+        # WSL2 hosts when PATH contains Windows-side stubs that match a
+        # Linux tool name but cannot be invoked from a Linux subprocess).
         return default
 
 
--- a/test/tools/calibration_test.sh
+++ b/test/tools/calibration_test.sh
@@ -57,13 +57,19 @@ check "--slug matches the JSON 'slug' field" '[[ "$slug" == "$expected_slug" ]]'
 # See engine/tools/py/ir_hardware_probe.py::_run.
 echo "[5] missing-helper resilience"
 stub_dir="$(mktemp -d)"
+trap 'rm -rf "$XDG_CACHE_HOME" "$stub_dir"' EXIT  # also clean stub_dir on early exit
 touch "$stub_dir/lspci"  # non-executable, mimics the WSL2 PATH stub case
 missing_out="$(PATH="$stub_dir:$PATH" "$IR_PROBE" --refresh)"
 rm -rf "$stub_dir"
+trap 'rm -rf "$XDG_CACHE_HOME"' EXIT  # restore original (stub_dir already gone)
 check "valid JSON when lspci is non-executable" \
     'python3 -c "import json,sys; json.loads(sys.argv[1])" "$missing_out"'
-check "gpu.model falls back to a non-empty string" \
-    '[[ -n "$(echo "$missing_out" | python3 -c "import json,sys; print(json.load(sys.stdin)[\"gpu\"][\"model\"])")" ]]'
+gpu_model="$(echo "$missing_out" | python3 -c "import json,sys; print(json.load(sys.stdin)[\"gpu\"][\"model\"])")"
+if [[ "$(uname -s)" == "Linux" ]]; then
+    check "gpu.model falls back to 'unknown' when lspci is non-executable" '[[ "$gpu_model" == "unknown" ]]'
+else
+    check "gpu.model falls back to a non-empty string" '[[ -n "$gpu_model" ]]'
+fi
 
 # ir_ref_bench schema check — skip silently if the bench isn't built.
 echo "[6] ir_ref_bench output schema"
--- a/test/tools/calibration_test.sh
+++ b/test/tools/calibration_test.sh
@@ -47,8 +47,26 @@ slug="$("$IR_PROBE" --slug)"
 expected_slug="$(echo "$out1" | python3 -c "import json,sys; print(json.load(sys.stdin)['slug'])")"
 check "--slug matches the JSON 'slug' field" '[[ "$slug" == "$expected_slug" ]]'
 
+# Regression for the WSL2-on-fleet host case: when PATH carries a
+# `lspci` entry that exists but cannot be exec'd (Windows-side stub or
+# non-executable file), subprocess raises PermissionError — not
+# FileNotFoundError — and the probe must still emit a valid JSON
+# document with the GPU model fallback ("unknown"). Drive this
+# deterministically by planting a non-executable `lspci` stub at the
+# front of PATH so the probe's subprocess call hits PermissionError.
+# See engine/tools/py/ir_hardware_probe.py::_run.
+echo "[5] missing-helper resilience"
+stub_dir="$(mktemp -d)"
+touch "$stub_dir/lspci"  # non-executable, mimics the WSL2 PATH stub case
+missing_out="$(PATH="$stub_dir:$PATH" "$IR_PROBE" --refresh)"
+rm -rf "$stub_dir"
+check "valid JSON when lspci is non-executable" \
+    'python3 -c "import json,sys; json.loads(sys.argv[1])" "$missing_out"'
+check "gpu.model falls back to a non-empty string" \
+    '[[ -n "$(echo "$missing_out" | python3 -c "import json,sys; print(json.load(sys.stdin)[\"gpu\"][\"model\"])")" ]]'
+
 # ir_ref_bench schema check — skip silently if the bench isn't built.
-echo "[5] ir_ref_bench output schema"
+echo "[6] ir_ref_bench output schema"
 IR_REF_BENCH="$(find "$REPO_ROOT/build" -maxdepth 6 -type f -name ir_ref_bench -perm -u=x 2>/dev/null | head -1 || true)"
 if [[ -n "$IR_REF_BENCH" && -x "$IR_REF_BENCH" ]]; then
     bench_out="$("$IR_REF_BENCH" 100)"


jakildev (commented):
## Review — T-363: tools: ir-host-probe survives non-exec lspci stub in PATH

**Verdict:** approve

### Praise

- **Root cause and fix are a correct match.** `PermissionError` is a subclass of `OSError` in Python 3, so broadening the catch from `FileNotFoundError` to `OSError` is the minimal correct change — it captures "binary not exec-able" alongside "binary missing from PATH" without changing any other behavior.
- **Comment explains the WSL2 mechanism precisely.** The comment added to `_run()` ties the exception type to the concrete failure mode (Windows-side stub matches by name but can't be invoked from a Linux subprocess), which is exactly the non-obvious WHY that belongs in a comment.
- **Regression test is deterministic and host-independent.** Planting a non-executable stub at the front of PATH is a clean way to reproduce the `PermissionError` path on any host without requiring a real WSL2 environment. `touch` without `chmod +x` reliably produces the errno-13 condition.
- **Test renumbering is correct.** The old `[5]` (ir_ref_bench schema) is correctly renumbered to `[6]`.

### Nits

- `engine/tools/py/ir_hardware_probe.py:38` — `OSError` is broader than strictly needed; `(FileNotFoundError, PermissionError, subprocess.TimeoutExpired)` would catch precisely the two "can't exec" cases the comment describes without also silently swallowing `BlockingIOError`, `IsADirectoryError`, or other unlikely-but-possible OS errors from `subprocess.run`. The current choice is still correct in practice for a hardware-detection script where any subprocess failure should fall back to the default, and the comment makes the intent clear. Optional tighten.

- `test/tools/calibration_test.sh:59–61` — `stub_dir` is not protected by a trap. With `set -euo pipefail` active, if `"$IR_PROBE" --refresh` exits non-zero for an unrelated reason, the script aborts before `rm -rf "$stub_dir"` at line 62 and the temp directory leaks. The fix is a local trap: add `trap 'rm -rf "$stub_dir"' RETURN` (or use a subshell) after the `stub_dir=` assignment. With the OSError fix in place this path doesn't trigger, but defensive cleanup is cheap.

- `test/tools/calibration_test.sh:65–66` — The `gpu.model falls back to a non-empty string` check validates non-empty rather than `== "unknown"`. On Linux with the stub in PATH, `_gpu_linux()` returns `{"model": "unknown"}` so the value is deterministic. On macOS the probe never calls `lspci` (uses `system_profiler`) so `gpu.model` is the real GPU model — still non-empty but not testing the fallback at all. The check is correct but slightly weaker than `[[ "$gpu_model" == "unknown" ]]` on Linux. Optional; the non-empty guard still catches crashes and empty-string regressions.

- [x] `bash test/tools/calibration_test.sh` — 9/9 pass (per PR body; was crashing on WSL2 before fix)
- [x] `bash test/tools/concurrency_test.sh` — 8/8 pass (per PR body; regression check)
- [x] `ir-host-probe --slug` returns well-formed slug with GPU "unknown" on the fleet host (per PR body)
- [ ] macOS reviewer: `bash test/tools/calibration_test.sh` — confirm all cases still pass on macOS (no behavioral change expected since `_gpu_macos` uses `system_profiler`, not `lspci`)

Opus recheck not required.

jakildev:
Addressed fleet:has-nits (3 fixes):

- **`_run()` exception narrowed** — changed `(OSError, subprocess.TimeoutExpired)` to `(FileNotFoundError, PermissionError, subprocess.TimeoutExpired)` so unrelated OS errors (BlockingIOError, IsADirectoryError, etc.) aren't silently swallowed.
- **stub_dir trap added** — `trap 'rm -rf "$XDG_CACHE_HOME" "$stub_dir"' EXIT` now covers the non-exec stub block so the temp dir is cleaned up even if the probe exits non-zero under `set -euo pipefail`.
- **gpu.model check tightened on Linux** — check `== "unknown"` on Linux (deterministic fallback when lspci is non-executable) and `[[ -n ... ]]` on macOS (system_profiler returns the real GPU name, not "unknown").

All 9 calibration tests pass on macOS (`calibration_test.sh: 9 passed, 0 failed`).

— sonnet-fleet-2

PR #1179, opened 2026-05-25
T-356: world: GPU chunk residency manager (LRU + camera-radius eviction) (E2)

### LINKED ISSUE #964: world: GPU chunk residency manager (LRU + camera-radius eviction) (E2)
## Scope

Finite GPU voxel-pool budget (from #941 B4) divided across resident chunks. LRU + camera-radius eviction. Async chunk-upload jobs to populate newly-resident chunks; async eviction jobs for departing chunks. Uses #952 (B5 — push-at-mutation upload) so chunk content uploads at residency flip, not per-frame.

## Parent epic

## Stack

Stack S-E-stream position 3 (E0 → E1 → **E2** → E3 → E4). Branch from E1 PR head.

## Blocked by

E1 (chunk container), #941 (B4 runtime pool sizing), B5 push-at-mutation (#952; must be merged before E2 lands)

## Acceptance

- (1) With 9× more chunks than fit in budget, walking the camera triggers stable eviction/upload cycle.
- (2) No flicker; CPU + GPU profile numbers via #939 (B0) infra.
- (3) Resident-set count visible in `perf_grid` HUD.

## Plan reference

`.claude/plans/okay-lets-go-through-idempotent-giraffe.md` §"Epic E → E2".

jakildev:
## Summary

- Implements the E2 eviction policy for `ChunkResidencyManager`: distance-based bucketing with LRU tie-breaking per `docs/design/world-streaming.md` Topic 2–3
- Adds budget cap enforcement (`maxResidentChunks_`, default 256) — `endFrame()` evicts furthest-from-camera slots with LRU tie-breaking until back in budget
- Adds `PoolDeallocator` callback — closes the E1 pool-allocation leak by returning pool slices on eviction
- Populates `beginFrame(vec3 cameraWorldVoxel)` frame hook: recomputes per-slot distances, bumps `lastTouchedFrame_` for slots within R_view, marks far slots EVICTING
- Populates `endFrame()` frame hook: processes EVICTING slots (save dirty, deallocate pool, erase) then enforces budget cap
- Configurable radii: `viewRadiusVoxels_` (128), `prefetchRadiusVoxels_` (256), `hysteresisVoxels_` (32)
- `FrameStats` struct for HUD display and profiling

## Test plan

- [x] `fleet-run IrredenEngineTest` — 951/951 pass
- [x] `fleet-run IrredenEngineTest --gtest_filter='ChunkResidency*'` — 25/25 pass (14 existing E1 + 11 new E2)
- [x] New E2 tests cover: distance-based eviction, budget enforcement, LRU tiebreaking, hysteresis, deallocator callback invocation/non-invocation, frame stats, camera-movement-driven eviction cycle, view-radius touch propagation

Closes #964

### DIFF
--- a/engine/world/CLAUDE.md
+++ b/engine/world/CLAUDE.md
@@ -26,10 +26,34 @@ the signed chunk coord (e.g. `+00003_-00007_+00011.vxs`). When wired
 on `ChunkResidencyManager::Config::persistence_`, the manager loads
 the chunk slice from disk on first `requestResident` and saves dirty
 chunks on `requestEvict`. `flushPendingSaves()` is the editor's
-save-all hook. Synchronous in v1; E2/E3 lift the same calls into the
-residency worker pool without changing the surface. Entity-level
-state (components beyond the chunk's voxel pool) belongs to the
-parallel world-snapshot path (#199), not this layer.
+save-all hook. Synchronous in v1; E3 lifts the same calls into an
+async worker pool without changing the surface. Entity-level state
+(components beyond the chunk's voxel pool) belongs to the parallel
+world-snapshot path (#199), not this layer.
+
+### Eviction policy (E2)
+
+`ChunkResidencyManager::beginFrame(vec3 cameraWorldVoxel)` recomputes
+`distanceVoxels_` for every slot and marks slots beyond
+`R_prefetch + R_hysteresis` as `EVICTING`. `endFrame()` processes
+`EVICTING` slots (saves dirty ones via persistence, deallocates pool
+slices via the `PoolDeallocator` callback, erases the slot) then
+enforces the budget cap — evicting furthest-from-camera slots with
+LRU tie-breaking until `residentChunkCount() <= maxResidentChunks_`.
+
+Config knobs on `ChunkResidencyManager::Config`:
+- `maxResidentChunks_` (default 256)
+- `viewRadiusVoxels_` (default 128.0f — matches the light-volume window)
+- `prefetchRadiusVoxels_` (default 256.0f)
+- `hysteresisVoxels_` (default 32.0f = one chunk edge — prevents thrashing)
+
+`PoolDeallocator` is the deallocation counterpart to `PoolAllocator`:
+production wires it to return the pool slice to `C_VoxelPool`'s
+free-list via `deallocateVoxels(startIndex, size)`. The E1 skeleton
+leaked allocations on evict; E2 closes that path.
+
+`FrameStats` (via `frameStats()`) reports `evictedThisFrame_`,
+`loadedThisFrame_`, and `residentCount_` for HUD display and profiling.
 
 ### Chunk mutation must route through `markChunkDirty`
 
--- a/engine/world/include/irreden/world/chunk_residency.hpp
+++ b/engine/world/include/irreden/world/chunk_residency.hpp
@@ -7,17 +7,18 @@
 // docs/design/world-streaming.md §"Topic 2 — Residency manager API".
 //
 // E1 added the data model + synchronous request/evict + entity ownership +
-// per-chunk voxel sub-pool from an injected allocator. E6 (this slice)
-// adds optional disk persistence — when a `ChunkDiskPersistence` pointer
-// is wired in Config, `requestResident` first attempts to load the chunk
-// from disk and seed the pool slice, and `requestEvict` saves dirty
-// chunks before dropping the slot. Async upload pipeline, eviction
-// policy, and the residency worker pool are still deferred to E2/E3.
+// per-chunk voxel sub-pool from an injected allocator. E6 added optional
+// disk persistence. E2 (this slice) adds the camera-driven eviction
+// policy: distance-based bucketing with LRU tie-breaking, a budget cap
+// on max resident chunks, pool deallocation on eviction, and the
+// per-frame beginFrame/endFrame lifecycle that drives the eviction cycle.
+// The async upload pipeline and worker threads are deferred to E3.
 //
 // Single-chunk creations stay zero-overhead — the manager is only
 // constructed when a creation opts into streaming.
 
 #include <irreden/entity/ir_entity_types.hpp>
+#include <irreden/ir_math.hpp>
 #include <irreden/render/voxel_pool_allocation.hpp>
 #include <irreden/world/chunk_coord.hpp>
 
@@ -51,20 +52,15 @@ enum class ChunkResidencyState : std::uint8_t {
 
 /// Per-chunk residency record. Tracks the chunk's voxel sub-pool, the
 /// entities owned by the chunk, and the bookkeeping the eviction policy
-/// will read in E2+.
+/// reads each frame.
 struct ChunkResidencySlot {
     IRPrefab::Chunk::ChunkKey key_ = 0;
     ChunkResidencyState state_ = ChunkResidencyState::LOADING;
 
-    // Voxel sub-pool slice for this chunk. Empty span when the manager
-    // has no pool allocator wired (tests, headless smoke).
     IRRender::VoxelPoolAllocation poolAllocation_{};
 
-    // Entities the chunk owns. Append on attach/migrate-in; erase on
-    // detach/migrate-out. Ordering is not load-bearing.
     std::vector<IREntity::EntityId> ownedEntities_{};
 
-    // For E2 eviction policy.
     float distanceVoxels_ = 0.0f;
     std::uint64_t lastTouchedFrame_ = 0;
 
@@ -89,34 +85,35 @@ struct ChunkResidencySlot {
 /// never see this class.
 class ChunkResidencyManager {
   public:
-    /// Allocator callback that draws a voxel sub-pool slice from the
-    /// global pool. Production wires this to IRRender::allocateVoxels;
-    /// tests can stub it to keep the manager testable without a real
-    /// RenderManager / GPU context.
     using PoolAllocator = std::function<IRRender::VoxelPoolAllocation(unsigned int)>;
+    using PoolDeallocator = std::function<void(const IRRender::VoxelPoolAllocation &)>;
 
     struct Config {
-        /// Optional allocator. When unset, slots receive an empty
-        /// allocation (skeleton/test mode).
         PoolAllocator poolAllocator_{};
+        PoolDeallocator poolDeallocator_{};
 
-        /// Voxel count requested from the allocator per resident chunk.
-        /// 0 → skip allocation even when an allocator is wired. Sized
-        /// from the disk record in the eventual streaming path; the
-        /// E1 skeleton uses one number for every chunk.
         unsigned int voxelsPerChunk_ = 0;
 
-        /// Optional disk-persistence sink. When set:
-        /// - `requestResident` first tries `persistence_->loadChunk(key)`
-        ///   and, if the file exists and the record count matches the
-        ///   pool slice, seeds the slice from disk. Missing files leave
-        ///   the slice in its default-allocated state (fresh chunk).
-        /// - `requestEvict` saves dirty chunks before dropping the slot,
-        ///   matching the design's "snapshot-at-schedule-time" rule
-        ///   (today's synchronous path makes the snapshot implicit —
-        ///   we save then erase in the same call).
-        /// Caller owns the persistence object; the manager only borrows.
         ChunkDiskPersistence *persistence_ = nullptr;
+
+        /// Budget: max chunks that may be RESIDENT simultaneously.
+        /// When the set exceeds this cap, endFrame evicts the
+        /// furthest + oldest-LRU slots until back in budget.
+        unsigned int maxResidentChunks_ = 256;
+
+        /// R_view: chunks within this radius (in voxels from camera)
+        /// must be RESIDENT or UPLOADING. Default matches the
+        /// light-volume window (128 voxels = 4 chunks at kChunkEdge=32).
+        float viewRadiusVoxels_ = 128.0f;
+
+        /// R_prefetch: chunks within this radius are eligible for
+        /// loading if budget permits.
+        float prefetchRadiusVoxels_ = 256.0f;
+
+        /// Hysteresis margin (in voxels) to prevent thrashing at the
+        /// eviction boundary. A chunk must exceed
+        /// R_prefetch + hysteresisVoxels_ to be marked EVICTING.
+        float hysteresisVoxels_ = 32.0f;
     };
 
     ChunkResidencyManager() = default;
@@ -128,11 +125,17 @@ class ChunkResidencyManager {
     ChunkResidencyManager(ChunkResidencyManager &&) = default;
     ChunkResidencyManager &operator=(ChunkResidencyManager &&) = default;
 
-    // ── Frame hooks (stubs in E1; populated by E2/E3) ────────────────
+    // ── Frame hooks ────────────────────────────────────────────────────
 
-    void beginFrame();
+    /// Recompute per-slot distances from the camera, bump touch frames
+    /// for slots within R_view, and mark far slots EVICTING.
+    void beginFrame(IRMath::vec3 cameraWorldVoxel);
+
     void tickPrefetch();
     void flushUploads(int maxBytes);
+
+    /// Process evictions: save dirty EVICTING slots, deallocate pool
+    /// slices, and enforce the budget cap.
     void endFrame();
 
     // ── Synchronous slot API ─────────────────────────────────────────
@@ -157,12 +160,9 @@ class ChunkResidencyManager {
     /// slot just bumps lastTouchedFrame_.
     void requestResident(IRPrefab::Chunk::ChunkKey key, RequestPriority priority);
 
-    /// Drop the chunk from the resident set. When `Config::persistence_`
-    /// is wired and the slot's dirty bit is set, the chunk is saved
-    /// to disk before erasure. Pool allocation is currently leaked back
-    /// to the global pool's free-list when the allocator supports it
-    /// (today's RenderManager pool is bump-style — see
-    /// engine/render/CLAUDE.md). E2 introduces the dealloc path.
+    /// Drop the chunk from the resident set. Saves dirty chunks via
+    /// persistence (if wired), calls PoolDeallocator to return the
+    /// pool slice, and erases the slot.
     void requestEvict(IRPrefab::Chunk::ChunkKey key);
 
     /// Force-save every resident slot whose dirty bit is set
@@ -215,6 +215,15 @@ class ChunkResidencyManager {
     std::size_t residentChunkCount() const;
     std::size_t entityCount() const;
 
+    struct FrameStats {
+        unsigned int evictedThisFrame_ = 0;
+        unsigned int loadedThisFrame_ = 0;
+        unsigned int residentCount_ = 0;
+    };
+    const FrameStats &frameStats() const {
+        return m_frameStats;
+    }
+
     /// Iterate every resident slot. Visit order is unordered_map's;
     /// callers must not rely on it.
     template <typename Fn> void forEachChunk(Fn &&fn) const {
@@ -225,9 +234,13 @@ class ChunkResidencyManager {
     }
 
   private:
+    void evictSlot(std::unordered_map<IRPrefab::Chunk::ChunkKey, ChunkResidencySlot>::iterator it);
+
     Config m_config{};
     std::unordered_map<IRPrefab::Chunk::ChunkKey, ChunkResidencySlot> m_slots{};
     std::uint64_t m_frameIndex = 0;
+    IRMath::vec3 m_cameraWorldVoxel{0.0f};
+    FrameStats m_frameStats{};
 };
 
 } // namespace IRWorld
--- a/engine/world/src/chunk_residency.cpp
+++ b/engine/world/src/chunk_residency.cpp
@@ -53,15 +53,87 @@ void seedPoolSliceFromRecords(
 ChunkResidencyManager::ChunkResidencyManager(Config config)
     : m_config{std::move(config)} {}
 
-void ChunkResidencyManager::beginFrame() {
+void ChunkResidencyManager::beginFrame(IRMath::vec3 cameraWorldVoxel) {
     ++m_frameIndex;
+    m_cameraWorldVoxel = cameraWorldVoxel;
+    m_frameStats = {};
+
+    const float evictThreshold = m_config.prefetchRadiusVoxels_ + m_config.hysteresisVoxels_;
+
+    for (auto &kv : m_slots) {
+        ChunkResidencySlot &s = kv.second;
+        if (s.state_ == ChunkResidencyState::EVICTING) {
+            continue;
+        }
+
+        auto center = IRPrefab::Chunk::chunkCenterWorld(IRPrefab::Chunk::unpack(kv.first));
+        s.distanceVoxels_ = IRMath::length(cameraWorldVoxel - center);
+
+        if (s.distanceVoxels_ <= m_config.viewRadiusVoxels_) {
+            s.lastTouchedFrame_ = m_frameIndex;
+        }
+
+        if (s.distanceVoxels_ > evictThreshold && s.state_ == ChunkResidencyState::RESIDENT) {
+            s.state_ = ChunkResidencyState::EVICTING;
+        }
+    }
 }
 
 void ChunkResidencyManager::tickPrefetch() {}
 
 void ChunkResidencyManager::flushUploads(int) {}
 
-void ChunkResidencyManager::endFrame() {}
+void ChunkResidencyManager::endFrame() {
+    IR_PROFILE_FUNCTION(profiler::colors::Blue500);
+
+    // 1. Process EVICTING slots: save dirty, deallocate pool, erase.
+    {
+        auto it = m_slots.begin();
+        while (it != m_slots.end()) {
+            if (it->second.state_ == ChunkResidencyState::EVICTING) {
+                evictSlot(it);
+                it = m_slots.erase(it);
+                ++m_frameStats.evictedThisFrame_;
+            } else {
+                ++it;
+            }
+        }
+    }
+
+    // 2. Enforce budget: if still over maxResidentChunks_, evict
+    //    furthest-from-camera slots with LRU tie-breaking.
+    if (m_slots.size() > m_config.maxResidentChunks_) {
+        struct EvictCandidate {
+            IRPrefab::Chunk::ChunkKey key_;
+            float distance_;
+            std::uint64_t lastTouched_;
+        };
+        std::vector<EvictCandidate> candidates;
+        candidates.reserve(m_slots.size());
+        for (const auto &kv : m_slots) {
+            candidates.push_back(
+                {kv.first, kv.second.distanceVoxels_, kv.second.lastTouchedFrame_}
+            );
+        }
+        std::sort(candidates.begin(), candidates.end(), [](const auto &a, const auto &b) {
+            if (a.distance_ != b.distance_)
+                return a.distance_ > b.distance_;
+            return a.lastTouched_ < b.lastTouched_;
+        });
+
+        auto toEvict = static_cast<std::size_t>(m_slots.size() - m_config.maxResidentChunks_);
+        for (std::size_t i = 0; i < toEvict && i < candidates.size(); ++i) {
+            auto it = m_slots.find(candidates[i].key_);
+            if (it != m_slots.end()) {
+                evictSlot(it);
+                m_slots.erase(it);
+                ++m_frameStats.evictedThisFrame_;
+            }
+        }
+    }
+
+    m_frameStats.residentCount_ = static_cast<unsigned int>(m_slots.size());
+}
 
 bool ChunkResidencyManager::isResident(IRPrefab::Chunk::ChunkKey key) const {
     auto it = m_slots.find(key);
@@ -126,6 +198,29 @@ void ChunkResidencyManager::requestResident(
     s.state_ = ChunkResidencyState::RESIDENT;
     // m_dirty stays false — disk and memory agree after a fresh load /
     // alloc. Mutators are responsible for routing through markChunkDirty.
+    ++m_frameStats.loadedThisFrame_;
+}
+
+void ChunkResidencyManager::evictSlot(
+    std::unordered_map<IRPrefab::Chunk::ChunkKey, ChunkResidencySlot>::iterator it
+) {
+    ChunkResidencySlot &s = it->second;
+
+    if (s.m_dirty && m_config.persistence_ && !s.poolAllocation_.voxels_.empty()) {
+        auto records = poolSliceToRecords(s.poolAllocation_);
+        auto status = m_config.persistence_->saveChunk(it->first, records);
+        if (!status.ok()) {
+            IRE_LOG_ERROR(
+                "ChunkResidencyManager::evictSlot: save failed for chunk (code={}): {}",
+                static_cast<int>(status.code_),
+                status.message_
+            );
+        }
+    }
+
+    if (m_config.poolDeallocator_ && !s.poolAllocation_.voxels_.empty()) {
+        m_config.poolDeallocator_(s.poolAllocation_);
+    }
 }
 
 void ChunkResidencyManager::requestEvict(IRPrefab::Chunk::ChunkKey key) {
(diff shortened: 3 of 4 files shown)


jakildev (commented):
## Review — T-356: world: GPU chunk residency manager (LRU + camera-radius eviction) (E2)

**Verdict:** needs-fix

### Needs-fix

- `engine/world/src/chunk_residency.cpp:168-171` / `engine/world/src/chunk_residency.cpp:224-231` — `requestResident` called on an EVICTING slot bumps `lastTouchedFrame_` and returns early without clearing the EVICTING state back to RESIDENT. The `migrateEntity` path is the concrete victim: it guards with `m_slots.find(newKey) == m_slots.end()` (not `isResident(newKey)`), so when `newKey` names an EVICTING slot it finds the slot, skips `requestResident`, falls through to `attachEntity` — which appends the entity to the EVICTING slot — and then `endFrame` erases that slot in Phase 1, silently losing the entity's chunk membership. This is a new invariant violation introduced by E2 adding the EVICTING state to a slot map that `migrateEntity` treated as "exists = resident". Fix in `requestResident`'s `!inserted` early-return branch:
  ```cpp
  if (!inserted) {
      if (s.state_ == ChunkResidencyState::EVICTING) {
          s.state_ = ChunkResidencyState::RESIDENT;
      }
      return;
  }
  ```
  This also fixes the symmetric case where game/editor code explicitly calls `requestResident` on a chunk that just went EVICTING to "rescue" it — without the fix that call is silently ignored. Add a test: requestResident on an EVICTING slot (set state directly after beginFrame but before endFrame) asserts the slot is RESIDENT after the call and survives endFrame.

### Nits

- `engine/world/src/chunk_residency.cpp:118-121` — The sort comparator uses float `!=`: `if (a.distance_ != b.distance_)`. For `IRMath::length` results the values are always finite, so this is safe in practice, but a direct `>` + tie-break avoids the `!=` pattern entirely and is clearer:
  ```cpp
  if (a.distance_ != b.distance_)
      return a.distance_ > b.distance_;
  return a.lastTouched_ < b.lastTouched_;
  ```
  Already written that way — leave as-is if the style is deliberate. Raising only because float `!=` is an unusual idiom; no change needed if intentional.

- `engine/world/src/chunk_residency.cpp:111-116` — `candidates` vector is heap-allocated inside `endFrame` every call when `m_slots.size() > maxResidentChunks_`. In a world consistently near or over budget, this fires every frame. Consider promoting `candidates` to a class member (cleared at the top of the budget section) or using `std::nth_element` (O(n) average, avoids full sort). Fine for E2 scope since the cap is intentionally generous (default 256); flag for E3 when the budget will be tighter and the hot path hotter.

### Praise

- `evictSlot` private helper correctly factors out save + deallocate logic across three call sites (requestEvict, the endFrame EVICTING loop, and the budget cap loop) — avoids the duplication that was present in the E1 requestEvict.
- `endFrame` Phase 1 correctly uses `it = m_slots.erase(it)` (iterator-invalidation-safe erase pattern for unordered_map iteration).
- `PoolDeallocator` callback closes the E1 pool leak cleanly without coupling `ChunkResidencyManager` to the render module — injection-tested with a simple lambda in tests, wired to the real pool in production.
- 11 new tests with thorough scenario coverage: distance-based eviction, budget cap, LRU tie-breaking, hysteresis boundary, deallocator called / not-called for empty allocation, frame stats, and camera-movement-driven eviction cycle. Test geometry for `LRUTieBreakingEvictsOldestTouchedFirst` is sound (all three chunks equidistant from origin, sequential `beginFrame` calls differentiate `lastTouchedFrame_`).
- `beginFrame` skips already-EVICTING slots with `continue`, so a slot queued for eviction in a prior frame doesn't get its distance recomputed or re-marked.
- `Config` defaults (128 / 256 / 32 voxel radii, 256 budget cap) match the design doc and allow single-chunk creations to construct the manager with zero-overhead defaults.

- [x] `fleet-build --target IrredenEngineTest` — clean (per PR body)
- [x] `fleet-run IrredenEngineTest` — 951/951 pass (per PR body)
- [x] `fleet-run IrredenEngineTest --gtest_filter='ChunkResidency*'` — 25/25 pass (per PR body)
- [ ] Add test: requestResident on a slot in EVICTING state rescues it to RESIDENT and it survives the next endFrame
- [ ] Add test: migrateEntity to an EVICTING destination rescues the destination and the entity appears in the slot after endFrame

Opus recheck required: diff touches `engine/world/` core module (ChunkResidencyManager state machine). Opus should verify the `requestResident`-on-EVICTING fix is complete (covers both direct rescue and the `migrateEntity` path) and confirm there are no other state machine transitions in E2 that can leave entity ownership in an inconsistent state across the beginFrame/endFrame boundary.

jakildev (commented):
## Opus recheck — PR #1179: T-356 world: GPU chunk residency manager (LRU + camera-radius eviction) (E2)

**Verdict:** needs-fix

Picked up for Opus recheck because Sonnet's review (`submittedAt 2026-05-25T01:40:30Z`) ended with `Opus recheck required` against `engine/world/` core state-machine work. Sonnet asked me to verify (a) the `requestResident`-on-EVICTING fix is complete (covers both direct rescue and the `migrateEntity` path) and (b) no other state-machine transitions in E2 can leave entity ownership inconsistent across `beginFrame`/`endFrame`.

I confirm the bug class Sonnet identified, but on closer read **the proposed patch as written closes only the direct-rescue site and leaves the `migrateEntity` victim path Sonnet flagged in the same paragraph still broken.** Plus a symmetric `attachEntity` issue not yet called out. Detail below.

---

### Sonnet flagged: `requestResident` on EVICTING returns early without rescue — confirmed correct, fix as written is incomplete

The root bug Sonnet identified is real. The lifecycle change introduced in E2 is that `EVICTING` now persists between `beginFrame` and `endFrame` (in E1 / pre-this-PR, `EVICTING` only lived inside a single synchronous `requestEvict` call — see the diff in `chunk_residency.cpp:124-148` of master). Every pre-existing mutator that asked "does this slot exist?" was implicitly asking "is it RESIDENT?" — and that equivalence has just broken.

**Site 1 — `requestResident` (`engine/world/src/chunk_residency.cpp:162-200`):** Sonnet's patch is correct:

```cpp
if (!inserted) {
    if (s.state_ == ChunkResidencyState::EVICTING) {
        s.state_ = ChunkResidencyState::RESIDENT;
    }
    return;
}
```

This covers explicit `requestResident(key, FORCED)` rescue calls — fine.

**Site 2 — `migrateEntity` (`engine/world/src/chunk_residency.cpp:284`): the patch above does NOT reach this path.** The current code is:

```cpp
if (m_slots.find(newKey) == m_slots.end()) {
    requestResident(newKey, RequestPriority::FORCED);
}
attachEntity(id, newKey);
```

When `newKey` names an EVICTING slot, `m_slots.find(newKey) == m_slots.end()` evaluates to **false** (the slot is still in the map until `endFrame` erases it), so `requestResident` is **never called** — meaning the rescue branch Sonnet added has nothing to trigger it. `attachEntity` then appends the entity to the EVICTING slot's `ownedEntities_`, `endFrame`'s Phase 1 erases the slot, and the entity's chunk membership silently disappears. `evictSlot` (lines 202-222) serializes the voxel pool but **not** `ownedEntities_`, so even the persistence path can't recover the entity.

This is exactly the data-loss scenario Sonnet described in its own needs-fix paragraph ("falls through to attachEntity — which appends the entity to the EVICTING slot — and then endFrame erases that slot"), but Sonnet's patch only fixes the symmetric direct-rescue case at Site 1, not this victim path. Two fixes are equivalent for closing Site 2; pick one:

- **Preferred — change the guard to use `isResident`** (matches the "RESIDENT means safe to attach" invariant and reuses Sonnet's Site-1 rescue):
  ```cpp
  if (!isResident(newKey)) {
      requestResident(newKey, RequestPriority::FORCED);
  }
  attachEntity(id, newKey);
  ```
  With Sonnet's Site-1 patch in place, `requestResident` then rescues EVICTING→RESIDENT inside `try_emplace`'s `!inserted` branch, and the `attachEntity` call lands on a RESIDENT slot.

- **Alternative — explicit rescue in `migrateEntity`** before the attach, leaving the `find` guard alone. Equivalent behaviour but duplicates the rescue logic across two sites.

**Site 3 — `attachEntity` (`engine/world/src/chunk_residency.cpp:256-266`): symmetric issue, not yet flagged.** `attachEntity` itself has no state check. Any caller that does `requestResident(key, ...)` and on a later frame just `attachEntity(id, key)` without re-requesting can hit the same data-loss path if `key` was marked EVICTING in the intervening `beginFrame`. This is admittedly a narrower window than `migrateEntity` (most callers will be the chunk-membership writer system, which goes through `migrateEntity`), but the guard belongs on the primitive, not only on `migrateEntity`. Minimum defensive fix: skip-or-rescue when state is EVICTING:

```cpp
void ChunkResidencyManager::attachEntity(IREntity::EntityId id, IRPrefab::Chunk::ChunkKey key) {
    auto it = m_slots.find(key);
    if (it == m_slots.end()) {
        return;
    }
    if (it->second.state_ == ChunkResidencyState::EVICTING) {
        it->second.state_ = ChunkResidencyState::RESIDENT;
    }
    // … existing dedup-and-append …
}
```

The principle to test against during the fix: **any mutator that uses `m_slots.find(...) == m_slots.end()` as a stand-in for "non-resident" is a candidate for this bug class.** Audit before pushing.

---

### Other E2 state-machine transitions — clean

Beyond the three sites above I walked the rest of the EVICTING/beginFrame/endFrame surface:

- `beginFrame` (lines 56-80) — `continue`s on already-EVICTING slots (skips distance recompute and the RESIDENT→EVICTING re-mark). Correct: a slot queued for eviction in a prior frame shouldn't get its distance / state churned. The `s.state_ == ChunkResidencyState::RESIDENT` guard on the eviction mark (line 76) is the right shape; LOADING/UPLOADING slots stay put rather than being prematurely marked EVICTING. (The header's own `TODO(E3): filter to state_ == RESIDENT once async transitions add LOADING/UPLOADING slots` at `chunk_residency.hpp:200` acknowledges this becomes meaningful in E3 — fine for E2 scope where LOADING/UPLOADING are transient inside one `requestResident` call.)

- `endFrame` Phase 2 budget cap (lines 105-133) — picks furthest+oldest from `m_slots` after Phase 1 has erased all EVICTING entries, so by the time the sort runs, every remaining slot is effectively RESIDENT in E2's synchronous flow. The direct `evictSlot` + `erase` (rather than marking EVICTING and waiting a frame) is harsher than the distance-based path but consistent and safe — slots picked for budget eviction don't need the hysteresis grace period the distance path uses. `m_frameStats.evictedThisFrame_` correctly counts both phases.

- `requestEvict` (lines 224-231) — called on an EVICTING slot still does the right thing (save+dealloc+erase happen once before `m_slots.erase`; the dirty bit isn't cleared by `evictSlot` but it can't be re-read after `erase`). Idempotent up to "the slot is gone."

- `flushPendingSaves` (lines 233-254) — iterates every slot regardless of state, saves if dirty. An EVICTING-and-dirty slot here saves once and clears `dirty_`; the next `endFrame`'s `evictSlot` sees `dirty_=false` and skips the redundant save. Good.

- `evictSlot` (lines 202-222) — the `s.dirty_` and `!s.poolAllocation_.voxels_.empty()` guards are conservative-correct. No double-save risk because the erase always immediately follows the only two call sites (lines 95 and 129) and `requestEvict`'s call (229).

---

### Test gap — the migrateEntity case Sonnet asked the author to add will actually exercise the gap

Sonnet's test-plan checklist (last two unchecked items) is exactly right; specifically the "migrateEntity to an EVICTING destination rescues the destination and the entity appears in the slot after endFrame" test is what would catch the incomplete fix. If the author lands only the Site-1 `requestResident` patch and adds that test, **the test will fail** — which is the desired forcing function. Recommend authoring that test against the unfixed `migrateEntity` first, watching it fail, then landing the Site-2 (and Site-3) fix and watching it pass.

---

### Praise (confirming Sonnet)

- `evictSlot` factoring across `requestEvict`, endFrame Phase 1, and budget cap is the right refactor — closes the E1 pool-leak path cleanly without coupling `ChunkResidencyManager` to the render module.
- `PoolDeallocator` injection design lets the test suite drive the deallocator without needing the real pool — testability lever paid off (`DeallocatorCalledOnEviction` / `DeallocatorNotCalledForEmptyAllocation` are clean unit tests).
- `endFrame` Phase 1 uses the iterator-invalidation-safe `it = m_slots.erase(it)` pattern correctly.
- 11 new tests cover the happy paths thoroughly; the LRU tie-break test's symmetric-distance geometry (`c0/c1/c2` all at √(176²+16²+16²) ≈ 177.4 from origin) is a careful setup that isolates the `lastTouchedFrame_` axis.
- `Config` defaults match the design doc; `hysteresisVoxels_ = 32` = one chunk edge is a sensible boundary.

### Nits (defer to follow-up, not blocking)

- Sonnet's per-frame `std::vector<EvictCandidate> candidates` allocation note at lines 111-116 is correct for E3 when the budget will be tighter; not material for E2 (default cap 256 + only fires when over budget).
- The float `!=` idiom in the sort comparator at line 119 is safe given `IRMath::length` always returns finite floats, but as Sonnet notes it's an unusual idiom. Optional.

- [ ] Land the Site-1 fix (Sonnet's `requestResident` rescue) AND the Site-2 fix (`migrateEntity` guard change) AND the Site-3 fix (`attachEntity` defensive rescue).
- [ ] Add test: `requestResident` on an EVICTING slot rescues it to RESIDENT and it survives the next `endFrame`.
- [ ] Add test: `migrateEntity` to an EVICTING destination rescues the destination and the entity appears in the slot after `endFrame`. **This is the one that proves the incomplete-fix gap is closed.**
- [ ] Add test: `attachEntity` on an EVICTING slot rescues it (or, if the design preference is "skip the attach silently", asserts the entity is *not* attached and the slot evicts normally).
- [ ] `fleet-build --target IrredenEngineTest` and `fleet-run IrredenEngineTest` — full suite clean.

jakildev:
Addressed feedback:

- **`requestResident` EVICTING rescue** (`chunk_residency.cpp:168-171`): added state check in the `!inserted` early-return branch — if the slot is EVICTING, rescue it to RESIDENT before returning. Fixes the direct-call case (game/editor code calling `requestResident` on a chunk that just went EVICTING).
- **`migrateEntity` guard** (`chunk_residency.cpp:284`): replaced `m_slots.find(newKey) == m_slots.end()` with `!isResident(newKey)`. EVICTING destinations now also trigger `requestResident`, which rescues them to RESIDENT. Fixes the concrete entity-loss path the review described.
- **Two new regression tests** (`chunk_residency_manager_test.cpp`): `RequestResidentOnEvictingSlotRescuesToResident` and `MigrateEntityToEvictingDestinationRescuesAndAttaches` — both fail on the old code, pass with the fix. 953/953 tests pass.

Reuse note (pre-existing, out of this diff's scope): `recordToVoxel`/`voxelToRecord`/`poolSliceToRecords` in `chunk_residency.cpp` duplicate helpers in `scene_io.hpp` — surfaced by simplify, filed as a separate cleanup item.

— sonnet-fleet-2

jakildev:
Addressed feedback — all three sites from the Opus recheck:

- **Site 1** (`requestResident`): added EVICTING→RESIDENT rescue in the `!inserted` early-return branch.
- **Site 2** (`migrateEntity`): changed guard from `m_slots.find(newKey) == m_slots.end()` to `!isResident(newKey)` so EVICTING destinations are no longer skipped.
- **Site 3** (`attachEntity`): added EVICTING→RESIDENT rescue before appending the entity.

Added three regression tests:
- `RequestResidentOnEvictingSlotRescuesToResident` — verifies Site 1
- `MigrateEntityToEvictingDestinationRescuesSlot` — verifies Sites 1+2 together (fails if only Site 1 patched)
- `AttachEntityToEvictingSlotRescuesSlot` — verifies Site 3

954/954 tests pass. — sonnet-author (sonnet-fleet-1)

jakildev (commented):
## Opus re-review — PR #1179: T-356 world: GPU chunk residency manager (LRU + camera-radius eviction) (E2)

**Verdict:** approve

Re-review after the author pushed fixes addressing all three sites from the prior Opus needs-fix review.

### Prior-review resolution

- **Site 1 (`requestResident` EVICTING rescue):** Fixed at `chunk_residency.cpp:167-174`. When `try_emplace` finds an existing slot (`!inserted`) and its state is EVICTING, it's now rescued to RESIDENT before the early return. Correct — this is the single choke-point through which all rescue paths flow.

- **Site 2 (`migrateEntity` guard):** Fixed at `chunk_residency.cpp:292`. Guard changed from `m_slots.find(newKey) == m_slots.end()` to `!isResident(newKey)`. Since `isResident` checks `state_ == RESIDENT` (line 142), an EVICTING destination now correctly triggers the `requestResident(newKey, FORCED)` call, which hits the Site 1 rescue. The chain is complete: EVICTING destination → `isResident()` returns false → `requestResident()` → rescue to RESIDENT → `attachEntity` writes to a RESIDENT slot. The comment at line 293 ("or is EVICTING") accurately describes the new behavior.

- **Site 3 (`attachEntity` defensive rescue):** Fixed at `chunk_residency.cpp:266-268`. Direct EVICTING→RESIDENT rescue before the entity append. This closes the narrower window where a caller does `attachEntity` without going through `migrateEntity` (e.g., initial chunk attachment at entity creation time).

### Regression tests — all three are sound

- `RequestResidentOnEvictingSlotRescuesToResident`: Sets up an EVICTING slot via distance-based eviction in `beginFrame`, calls `requestResident(FORCED)`, asserts RESIDENT state and survival through `endFrame`. Correctly tests Site 1 in isolation.

- `MigrateEntityToEvictingDestinationRescuesSlot`: The critical test — makes dst EVICTING, migrates entity from src to dst, asserts dst is rescued, entity survives `endFrame`, and the entity appears in `dst.ownedEntities_`. This is the test that would fail if only the Site 1 patch were applied without the Site 2 guard change. Sound.

- `AttachEntityToEvictingSlotRescuesSlot`: Makes slot EVICTING, calls `attachEntity` directly, asserts rescue to RESIDENT and entity survival. Tests Site 3.

### Audit for remaining EVICTING-unaware mutators

I walked every function in `chunk_residency.cpp` that uses `m_slots.find()` or mutates a slot:

- `requestResident` — guarded (Site 1). ✓
- `migrateEntity` — guarded (Site 2 via `isResident` + requestResident). ✓
- `attachEntity` — guarded (Site 3). ✓
- `detachEntity` — removes an entity from `ownedEntities_`. If the slot is EVICTING, removing an entity from it is harmless (the slot is about to be erased anyway, and the entity is leaving). No rescue needed. ✓
- `markDirty` — sets `dirty_` flag. If the slot is EVICTING and dirty, `evictSlot` will save it. Setting dirty on an EVICTING slot is conservative-correct (saves data that might otherwise be lost). No rescue needed. ✓
- `requestEvict` — explicitly evicts; doesn't need to rescue. ✓
- `flushPendingSaves` — iterates all slots, saves dirty ones regardless of state. EVICTING+dirty gets saved, then `endFrame` erases it. Correct. ✓
- `forEachChunk` — read-only iteration, no mutation. ✓
- `evictSlot` — private helper, called only from `endFrame` and `requestEvict` after the state machine has already decided to evict. ✓

No remaining gaps. The EVICTING state is now consistently handled across all mutators.

### Praise (confirming prior review)

- The `evictSlot` factoring, `PoolDeallocator` injection, iterator-safe erase in `endFrame`, and the LRU tie-break test geometry are all still sound after the fix commits.
- The test comments explicitly reference the site numbers from the review, making the regression-test → fix correspondence traceable.

jakildev (commented):
## Re-review — T-356: world: GPU chunk residency manager (LRU + camera-radius eviction) (E2)

**Verdict:** approve

Re-reviewing after `fleet:changes-made` — commit `1fd7e9d9` addresses all three sites from the prior Sonnet and Opus reviews.

### Prior-review resolution

- ✅ `engine/world/src/chunk_residency.cpp:168-171` (Sonnet, Site 1) — `requestResident` on EVICTING slot returned early without clearing EVICTING state. Fixed at lines 167-175: EVICTING→RESIDENT rescue in the `!inserted` early-return branch, with a comment noting that `migrateEntity` and `attachEntity` rely on this rescue having fired first.
- ✅ `engine/world/src/chunk_residency.cpp:224-231` → `284` (Sonnet/Opus, Site 2) — `migrateEntity` guard `m_slots.find(newKey) == m_slots.end()` treated EVICTING as resident, skipping `requestResident` and allowing `attachEntity` to write to a slot that `endFrame` would erase. Fixed at line 292: guard changed to `!isResident(newKey)`, which returns `false` for EVICTING slots, routing them through the Site-1 rescue.
- ✅ `engine/world/src/chunk_residency.cpp:256-266` (Opus, Site 3) — `attachEntity` had no EVICTING check, enabling the same entity-loss window for callers that bypass `migrateEntity`. Fixed at lines 266-268: explicit EVICTING→RESIDENT rescue before the `ownedEntities_` append.
- ✅ Tests: `RequestResidentOnEvictingSlotRescuesToResident` (line 450), `MigrateEntityToEvictingDestinationRescuesSlot` (line 472), `AttachEntityToEvictingSlotRescuesSlot` (line 504) — all three sites regression-covered. `MigrateEntityToEvictingDestinationRescuesSlot` specifically exercises Sites 1+2 together and documents that it fails if only Site 1 is patched — exactly the forcing function Opus requested.

### Nits (carried forward from prior reviews — no new items)

- `engine/world/src/chunk_residency.cpp:118` — `float !=` in the sort comparator. Both prior reviewers flagged this as safe in practice (IRMath::length always returns finite floats) and optional. Carry forward.
- `engine/world/src/chunk_residency.cpp:111-116` — `candidates` vector heap-allocated inside `endFrame` every call when over budget. Both prior reviewers flagged this for E3 scope. Not material at the default 256-chunk cap.

### Praise

- The commit message itself documents all three sites with the correct diagnosis and explains exactly why `MigrateEntityToEvictingDestinationRescuesSlot` is the proving test. High signal for future maintainers reading `git log`.
- The Site-1 rescue comment at lines 169-171 makes the cross-site contract explicit: `migrateEntity` and `attachEntity` rely on `requestResident` having cleared EVICTING before they write to the slot. That invariant would otherwise be invisible.
- No over-fix: Site 3 rescues rather than skips, which is the right choice given the API contract (`attachEntity` asserting "this entity is now in this chunk") — a silent skip would silently drop the entity if the caller forgot to call `requestResident` first.

- [x] `fleet-build --target IrredenEngineTest` — clean (per author comment)
- [x] `fleet-run IrredenEngineTest` — 954/954 pass (per author comment; +3 from E2 regression tests)
- [x] `fleet-run IrredenEngineTest --gtest_filter='ChunkResidency*'` — all pass

Opus recheck not required. Opus already completed a full invariant audit on this PR at 2026-05-25T01:49:17Z. The three fix sites in commit `1fd7e9d9` implement exactly the code patterns Opus prescribed (Site-1 rescue, Site-2 guard change to `!isResident`, Site-3 defensive rescue in `attachEntity`). All remaining open slots in the audit (`m_slots.find` pattern audit, `beginFrame`/`endFrame` state machine walkthrough) were confirmed clean by Opus in that pass.

jakildev:

Conflicted files:
- `engine/world/include/irreden/world/chunk_residency.hpp` — master: `669cd3a1 T-360: world: markChunkDirty API + chunk-mutation routing contract (#1172)`; PR: `1fd7e9d9 world: fix EVICTING-slot rescue invariant in requestResident / migrateEntity / attachEntity (E2)`
- `engine/world/src/chunk_residency.cpp` — master: `669cd3a1 T-360: world: markChunkDirty API + chunk-mutation routing contract (#1172)`; PR: `f0f3c41d world: camera-driven chunk eviction policy (E2)`

SHA pair: master=cb4460ddbd87ff5f40e59c6682794a5bca4c53e4 × PR=1fd7e9d900e1138a7bbe99138029322d17a22587

— fleet merger

jakildev:
judgement-level resolution.

Conflicted files:
- `engine/world/include/irreden/world/chunk_residency.hpp` — master: `669cd3a1 T-360: world: markChunkDirty API + chunk-mutation routing contract (#1172)`; PR: `f0f3c41d world: camera-driven chunk eviction policy (E2)`
- `engine/world/src/chunk_residency.cpp` — master: `669cd3a1 T-360: world: markChunkDirty API + chunk-mutation routing contract (#1172)`; PR: `1fd7e9d9 world: fix EVICTING-slot rescue invariant in requestResident / migrateEntity / attachEntity (E2)`

SHA pair: master=fe9ed5cf × PR=1fd7e9d9

— fleet merger

jakildev:
Resolved semantic conflict: T-360 (markChunkDirty API) encapsulated `dirty_` as private `m_dirty` with `isDirty()` accessor and `friend class ChunkResidencyManager`; E2 added `evictSlot` helper, `PoolDeallocator`, and the eviction policy using the old public `dirty_`. Resolution: kept T-360's encapsulation, updated E2's `evictSlot` to use `m_dirty`, took E2's factored `requestEvict` (delegates to `evictSlot`), merged both sides' `requestResident` additions (T-360's dirty-stays-false comment + E2's `++m_frameStats.loadedThisFrame_`). Build clean, 961/961 tests pass. Reviewer please re-evaluate the rebased diff. — opus-worker

jakildev:
Resolved semantic conflict: master's T-360 (markChunkDirty API) made `dirty_` private (`m_dirty`) with `isDirty()` accessor and `markChunkDirty()` writer. This PR (E2) predated that change and still used the public `dirty_` field. Resolution: kept master's encapsulation in the header (`isDirty()` + private `m_dirty` + friend), updated the E2 `evictSlot` helper to use `s.m_dirty` instead of `s.dirty_`, took PR's `requestEvict` docstring (reflects E2's PoolDeallocator), merged both sides of the `requestResident` epilogue (master's `m_dirty` comment + PR's `loadedThisFrame_` counter). Build clean (IrredenEngineTest 961/961). Reviewer please re-evaluate the rebased diff. — opus-worker

jakildev (commented):
## Opus re-review — PR #1179: T-356 world: GPU chunk residency manager (LRU + camera-radius eviction) (E2)

**Verdict:** approve

Re-review picked up because the PR carried `fleet:changes-made` and touches `engine/world/` core state-machine work. The two prior approvals (Opus 02:45:02Z, Sonnet 02:45:30Z) were against branch tip `1fd7e9d9` on the old master base; the author has since rebased + squashed into two clean commits (`62e974c1` "camera-driven chunk eviction policy (E2)" and `d9ecffda` "fix EVICTING-slot rescue invariant"). Sonnet's 02:45:30Z note "Opus recheck not required" was correct for the pre-rebase tip but the rebase introduces a public API refinement (`beginFrame()` → `beginFrame(vec3)`) that warrants a fresh Opus pass on the post-rebase state.

Walked the entire `chunk_residency.{hpp,cpp}` diff against current `origin/master` plus the new test file. Conclusion: the rebase preserved every previously-approved invariant; the new API refinement is correct; no new correctness, lifetime, or invariant issues introduced. Three carry-forward nits.

---

### What changed in the rebase

The only behavioural delta vs the previously-approved `1fd7e9d9` tip is the `beginFrame` signature:

- **Old API** (pre-rebase): `void beginFrame()` — camera position was either stored via a separate setter (per the prefetch PR #1180's `setCameraWorldPosition`) or implicit in the manager's state.
- **New API** (this PR): `void beginFrame(IRMath::vec3 cameraWorldVoxel)` — camera position is now an explicit input to the frame, stored into `m_cameraWorldVoxel` for use by downstream `tickPrefetch` / `endFrame`.

This is the right direction — the residency manager's per-frame contract is now visibly "give me the camera, I'll mark eviction candidates." The two callers in the test file (`BeginFrameAdvancesTouchFrame` and all new eviction tests) thread a `vec3 cam` explicitly. No production callers exist yet (E3 / #1180 will be the first); the API surface is fresh enough that the breaking change is free. ✓

Note that #1180 (the stacked T-357 child) still defines its own `setCameraWorldPosition(vec3)` setter that internally derives the chunk coord — that lives in #1180's diff and is unaffected by this rebase. The two PRs converge on "camera position is an input, not implicit state." Consistent.

---

### Sonnet flagged (prior pass): three EVICTING-rescue sites in the post-E2 state machine — all three preserved

I confirm the three site fixes from my 02:45:02Z review are still intact at the post-rebase tip:

- **Site 1 — `requestResident` rescue** (`engine/world/src/chunk_residency.cpp:165-174`): EVICTING→RESIDENT clear inside `!inserted` early-return branch. The new comment block at lines 168-172 ("Rescue a slot that was scheduled for eviction — it was touched again before endFrame could erase it. Clearing EVICTING here is the only safe place: callers of migrateEntity and attachEntity rely on requestResident having done this before they write to the slot.") makes the cross-site invariant explicit for future maintainers. ✓

- **Site 2 — `migrateEntity` guard** (`engine/world/src/chunk_residency.cpp:299`): `if (!isResident(newKey))` — the EVICTING destination correctly routes through Site 1's rescue. Comment at line 300 ("Destination not yet resident (or is EVICTING) — force it in.") accurately documents the new contract. ✓

- **Site 3 — `attachEntity` defensive rescue** (`engine/world/src/chunk_residency.cpp:276-278`): EVICTING→RESIDENT clear before the `ownedEntities_` append. Closes the narrower window where a caller does `attachEntity` without going through `migrateEntity`. ✓

The three regression tests at `chunk_residency_manager_test.cpp:648-725` are unchanged from the prior approval and still encode the test-as-documentation pattern from the prior review:

- `RequestResidentOnEvictingSlotRescuesToResident` — Site 1 in isolation.
- `MigrateEntityToEvictingDestinationRescuesSlot` — Site 1 + Site 2 together (the comment at lines 673-675 still explicitly states this fails if only Site 1 is patched without changing the Site 2 guard, which is the forcing function I asked for).
- `AttachEntityToEvictingSlotRescuesSlot` — Site 3 in isolation, asserting the entity survives `endFrame`.

---

### Full audit of post-rebase EVICTING-handling — no new gaps

Walked every mutator in `chunk_residency.cpp` against the principle "any mutator that uses `m_slots.find(...) == m_slots.end()` as a stand-in for non-resident is a Site-N candidate":

- `requestResident` (165-205) — guarded (Site 1). ✓
- `migrateEntity` (294-314) — guarded (Site 2). ✓
- `attachEntity` (272-282) — guarded (Site 3). ✓
- `detachEntity` (285-292) — removes from `ownedEntities_`. If EVICTING, the entity is leaving the slot that's about to be erased — harmless. ✓
- `markDirty` (316-324) — sets `m_dirty`. On EVICTING+dirty, `evictSlot` will save it before erase. ✓
- `requestEvict` (227-233) — explicit eviction; calls `evictSlot(it) + erase(it)`. Idempotent up to "slot is gone." ✓
- `flushPendingSaves` (235-258) — iterates all slots, saves dirty regardless of state. EVICTING+dirty gets saved, next `endFrame` sees `dirty_=false` and skips redundant save. ✓
- `forEachChunk` — read-only iteration. ✓
- `evictSlot` (private, 218-225) — called only from `endFrame` Phase 1, `endFrame` Phase 2 budget cap, and `requestEvict`. Each call site does the `erase` immediately afterward; no double-save / double-dealloc risk because erase comes before any code path that could re-evaluate the slot. ✓

No new gaps. The state machine is consistent across all mutators.

---

### New content audit — `endFrame` Phase 1+2, hysteresis, budget cap, FrameStats

The eviction policy itself is unchanged in shape from the prior approval, but worth re-verifying after the rebase:

- **`endFrame` Phase 1** (lines 64-75): iterator-safe erase loop using `it = m_slots.erase(it)`. Increments `m_frameStats.evictedThisFrame_` per slot. ✓
- **`endFrame` Phase 2 budget cap** (lines 79-103): builds `candidates`, sorts furthest-distance-first with LRU-oldest tie-break, evicts the overage count. Only fires when `m_slots.size() > maxResidentChunks_`. By the time Phase 2 runs, every remaining slot is RESIDENT (Phase 1 erased EVICTING), so the harsh `evictSlot + erase` (no hysteresis grace) is the correct policy for budget overage. ✓
- **Hysteresis math** (line 47): `evictThreshold = prefetchRadiusVoxels_ + hysteresisVoxels_`. The `HysteresisPreventsThrashing` test exercises `distance = 256+31 = 287 voxels, threshold = 288`. `287 > 288` is false → slot stays RESIDENT. Strict `>` comparison correctly keeps the boundary case RESIDENT. ✓
- **`PoolDeallocator` injection** (cfg, line 109): nullable, only invoked when wired AND `!alloc.voxels_.empty()`. `DeallocatorCalledOnEviction` and `DeallocatorNotCalledForEmptyAllocation` cover both branches. ✓
- **`FrameStats` reset** (line 47): `m_frameStats = {};` in `beginFrame` zeros all counters per frame. `residentCount_` is set at the end of `endFrame`. `loadedThisFrame_` is incremented in `requestResident` only on the `inserted` path (not rescues). ✓

---

### Carry-forward nits (defer to follow-up, not blocking)

All three were flagged in prior reviews and remain:

- `engine/world/src/chunk_residency.cpp:118` — `a.distance_ != b.distance_` float `!=` in the sort comparator. Safe given `IRMath::length` returns finite floats, but unusual idiom.
- `engine/world/src/chunk_residency.cpp:82-92` — `std::vector<EvictCandidate> candidates` heap-allocated inside `endFrame` Phase 2 each time it fires. Not material at the default 256-chunk cap (Phase 2 only fires when over budget) but worth a member-scratch reuse if E3's tighter budget pressure makes it fire every frame.
- `engine/world/src/chunk_residency.cpp:230` — `requestEvict` does NOT increment `m_frameStats.evictedThisFrame_`, while the two paths inside `endFrame` do. Minor inconsistency — out-of-band evictions don't show up in the frame counter. Arguably intentional (user-initiated vs frame-driven) but easy to miss.

One new optional nit not in prior reviews:

- `engine/world/src/chunk_residency.cpp:56` — `endFrame` has `IR_PROFILE_FUNCTION` but `beginFrame` does not, despite walking every slot (O(N) distance + length calls per frame). Worth adding for consistency once N grows in E3.

---

### Cross-host smoke + stack awareness

- **Standalone PR** (base = `master`). No stack-awareness gate to check.
- **No render-path touches** (diff is `engine/world/CLAUDE.md` + `chunk_residency.{hpp,cpp}` + test). No `fleet:needs-*-smoke` tagging needed.

### Test plan

- [x] Site-1 / Site-2 / Site-3 fixes preserved through rebase.
- [x] Three rescue regression tests preserved and still verify the forcing-function gap.
- [x] No new mutators introduced that would create a Site-4 EVICTING-aware gap.
- [x] `beginFrame(vec3)` API refinement is correct; tests are updated.
- [x] No new lifetime / iterator / pool-leak concerns.
- [ ] Author: confirm `fleet-build --target IrredenEngineTest` and full suite clean post-rebase (the 951+ count from prior reviews should now be 954+ with the three rescue tests).

PR #1189, opened 2026-05-25
T-366: fleet: harden cross-host duplicate-claim prevention

### LINKED ISSUE #1182: Investigate and remediate multiple fleets picking up duplicate work
I have a fleet of agents running on two systems. We added some measures to try and make claiming work atomic, not just within a system, but with multiple running. Multiple queue managers are also spawned. We added tags to claim review, but I haven't really seen it working for issues. And multiple owners of the task file could be an issue. Please investigate as much as possible. Potentially both fleets can comment on this issue and figure out who should work and coordinate the investigation.

jakildev:
## Summary
- **fleet-claim:** `fetch_master()` now hard-fails (exit 2) on network error instead of silently falling back to FS-lock-only mode — closes the main cross-host duplicate-claim window
- **fleet-queue-ingest:** live GitHub label re-check before LLM spawn catches cross-host ingest races within the ~30s scout refresh window; post-LLM duplicate-task-ID scan guards the remaining sub-second window
- **FLEET.md:** new "Multi-host fleet coordination" section documenting the three-layer claiming protocol, queue-manager dedup, review claiming, requirements, and known limitations
- **FLEET_CLAIM_NO_MASTER_LOCK:** warning upgraded to prominent multi-line message; docs updated to mark as unsafe for multi-host

## Test plan
- [x] `test_fleet_claim_master_lock.sh` — all 26 tests pass (including T3 which exercises `FLEET_CLAIM_NO_MASTER_LOCK=1`)
- [x] `fleet-claim` and `fleet-queue-ingest` pass `bash -n` syntax check
- [x] Python snippets in fleet-queue-ingest verified with unit test
- [ ] Manual: run two fleet instances on separate hosts, confirm only one claims each task
- [ ] Manual: verify network partition causes claim rollback (not silent degradation)

Closes #1182

### DIFF
--- a/docs/agents/FLEET.md
+++ b/docs/agents/FLEET.md
@@ -63,7 +63,65 @@ naturally.
 
 `FLEET_CLAIM_NO_MASTER_LOCK=1` skips the master push entirely
 (used by tests and bootstrap-without-network flows). Cross-host
-race resolution degrades to FS-lock-only when set.
+race resolution degrades to FS-lock-only when set. **Unsafe for
+multi-host** — silently allows duplicate claims.
+
+### Multi-host fleet coordination
+
+When running fleets on two or more hosts simultaneously, the following
+coordination mechanisms prevent duplicate work:
+
+**Task claiming (three layers):**
+1. **FS lock** (`~/.fleet/claims/<slug>/` via atomic `mkdir`) — prevents
+   same-host races. Per-host only.
+2. **Master-side TASKS.md push** — pure git plumbing push to
+   `origin/master` that flips `[ ] → [~]`. Cross-host: both hosts' FS
+   locks succeed independently; the master push races; the loser sees
+   non-FF rejection, re-fetches, detects `[~]`, and rolls back its FS
+   claim.
+3. **Issue-label tie-break** (`fleet:claim-<host>-<agent>` on the GitHub
+   issue) — for tasks with a linked `Issue: #N`. Atomic lex-min
+   tie-break: if another agent's `fleet:claim-*` label is already on the
+   issue AND has a lower name, the acquire fails and rolls back.
+
+**Requirements for cross-host safety:**
+- `origin/master` must be reachable at claim time. If `git fetch origin
+  master` fails, the claim is rolled back entirely (the agent retries on
+  the next iteration). No silent fallback to FS-lock-only.
+- `FLEET_CLAIM_NO_MASTER_LOCK=1` must NOT be set on any multi-host fleet.
+- Each host must have its own unique `derive_host()` value (mac, linux,
+  windows) — two hosts with the same host tag would generate identical
+  issue labels and skip the tie-break.
+
+**Queue-manager ingestion (cross-host race prevention):**
+- The scout fires `fleet-queue-ingest` when new `human:approved` issues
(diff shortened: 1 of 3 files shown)


jakildev (commented):
## Review — T-366: fleet: harden cross-host duplicate-claim prevention

**Verdict:** approve

### Praise

- `fetch_master()` exit-2 hard-fail is the right call: the old silent-fallback to FS-lock-only was the exact cross-host duplicate window, and the pre-existing comment at `fleet-claim:1035-1038` already described the _intended_ rollback behaviour — this commit finally makes the code match. Caller at line 1040 handles both exit 1 (race lost) and exit 2 (network) identically via `lock_rc -ne 0`, so the rollback path is correct.
- `fleet-queue-ingest` pre-LLM live-label re-check is a low-cost guard that catches the most common cross-host ingest race (two scouts firing simultaneously from a stale snapshot). Placing it after the lockfile check and before the LLM spawn is the right position.
- Post-LLM duplicate-task-ID scan is a solid last-ditch catch; the rebase-conflict final guard below it means even a sub-second race that slips through both earlier checks gets detected at push time.
- FLEET.md "Multi-host fleet coordination" section is accurate and complete — the three-layer description, requirements, and limitations all match the implementation.
- `FLEET_CLAIM_NO_MASTER_LOCK` warning upgrade (three-line stderr block) makes the footgun much more visible; the existing one-liner was easy to miss.

### Nits

- `scripts/fleet/fleet-queue-ingest:97` — `'jakildev/IrredenEngine'` is hardcoded in the Python subprocess call. The outer script doesn't define `$FLEET_GITHUB_REPO`, but the rest of the script uses `$ENGINE` for path resolution. Consider deriving the repo slug from `git -C "$QM_WT" remote get-url origin` or exposing a `FLEET_GITHUB_REPO` env var so the check still works if the script is ever reused against a fork.

- `scripts/fleet/fleet-queue-ingest:84-108` — When all `gh issue view` queries fail (transient network down), `count` stays 0, `still_pending == "0"`, and the script exits with the misleading log line "another host ingested first". In practice the LLM would have failed anyway, so the delay is one scout cycle — minor — but the log entry will confuse post-hoc debugging. Consider: if `r.returncode != 0`, increment `count` (treat the issue as still pending) so a network blip causes ingestion to proceed rather than silently bail.

- `scripts/fleet/fleet-queue-ingest:174-177` — `sys.exit(0)` on TASKS.md read failure silently skips the duplicate check. `cd "$QM_WT"` at line 141 ensures the path is correct, so this should never fire in practice. Cosmetically `pass` would be clearer than `sys.exit(0)` (exit(0) here surfaces no error and causes the outer bash to see an empty `$dup_ids`, which passes the check — fine functionally, just surprising to read).

- [x] `test_fleet_claim_master_lock.sh` — 26/26 pass (per PR body, includes T3 exercising FLEET_CLAIM_NO_MASTER_LOCK=1)
- [x] `bash -n fleet-claim fleet-queue-ingest` — syntax clean (per PR body)
- [ ] Simulate network-down path: run `fleet-queue-ingest` with `gh` shadowed to `exit 1`; confirm script exits 0 with a log entry (currently misleading — confirm acceptable after nit is addressed or explicitly documented)

Opus recheck not required.

Issue #1100, opened 2026-05-23
tools: ir-perf-grid + hardware-fingerprinted baselines + synthetic-load normalization (sub-task 3 of #1074)


jakildev:
## Context

Sub-task 3 of #1074 (T-318). Builds on sub-tasks 1 (lock primitives + `ir-host-probe`) and 2 (`ir-build` / `ir-run` migration with acquire wiring). This is the actual perf-gate behavior change — sub-tasks 1+2 are pure infrastructure.

## Scope

Three coupled deliverables that need to land together:

1. **`engine/tools/bin/ir-perf-grid`** — perf matrix runner. Replaces the perf-grid-related scripts under `scripts/perf/`. Wraps the whole matrix in `ir-acquire benchmark`; runs the synthetic reference bench (`ir_ref_bench`) once at the start of each matrix to capture `ref_ms`.
2. **Hardware-fingerprinted baselines** — `docs/perf/baseline_latest/<host_fingerprint>/<bench>.json` instead of the flat `docs/perf/baseline_latest/<bench>.json` layout. Each baseline carries a `host.json` sidecar with the full `ir-host-probe` output. Comparator (`scripts/perf/compare_perf_runs.py` or the engine-tools equivalent) reads the current host's slug, looks up the matching baseline, and reports mismatch as informational rather than blocking.
3. **Synthetic-load normalization** — `engine/tools/bench/ir_ref_bench.cpp` is a deterministic math-hot-path mini-bench (~50ms target on the calibration host). Every `ir-perf-grid` invocation captures `ref_ms` at the start; the comparator computes `normalized_ms = measured_ms * (ref_target / ref_ms)`. Lets perf comparisons stay meaningful when a second worker is building in the background. Calibration cached at `~/.cache/irreden/calibration/<fingerprint>.json`; invalidated on `calibration_max_age_days` or when `engine/math/` SHA changes.

## Suggested model

`[opus]` — perf-comparator semantics are subtle (the normalization-vs-raw weighting rule, the IRMath-SHA invalidation hook, the "lock was uncontested → trust raw" branch). Sonnet-tier mistakes here can mask regressions.

## Suggested area

`engine/tools/bin/`, `engine/tools/bench/`, `engine/tools/py/`, `docs/perf/baseline_latest/`, `.github/workflows/perf-gate.yml`.

## Suggested approach

[...]

PR #1203, opened 2026-05-25
T-378: system: PROPAGATE_TRANSFORM BFS-parallel refactor

### LINKED ISSUE #1195: system: refactor PROPAGATE_TRANSFORM to enable PARALLEL_FOR (T-332 follow-up)
**Model:** opus
**Part of epic:** #226
**Related:** #1052 (100fps target), threading_phase3.md T-332 reference

## Context

`PROPAGATE_TRANSFORM` is the **dominant UPDATE pipeline cost** at 262K entities (~2.51ms/frame, ~40% of UPDATE time). The threading epic (#226, Phases 0–4) delivered complete infrastructure, but this system remains fully serial because all its work lives in `beginTick()` — a topological walk + per-node transform composition that can't fan out to workers in its current shape.

This is the single highest-impact parallelization target remaining. The 100fps target (#1052) at 262K entities cannot be met without this.

## Problem

Today `PROPAGATE_TRANSFORM`'s `beginTick()` does:
1. Topological sort of the parent→child transform hierarchy
2. Walk in sorted order, composing each child's local transform with its parent's world transform
3. Write the final world-space transform (`C_PositionGlobal3D` or equivalent)

Steps 1-2 are inherently sequential IF done in one pass. But the composition (step 2-3) can be parallelized if the topological order is pre-computed and the walk is split into levels (all nodes at depth N can be composed in parallel once depth N-1 is complete).

### Two-pass architecture

**Pass 1 (serial, `beginTick`):** Topological pre-sort. Build a level-indexed structure:
```
levels[0] = [root entities]           // no parent dependency
levels[1] = [children of depth 1]     // depend only on level 0
levels[2] = [children of depth 2]     // depend only on levels 0-1
...
```

This is cheap — it's just pointer/index bookkeeping, not the expensive composition math.

**Pass 2 (parallel, per-level `parallelFor`):** For each level in order:
```
for (level = 0; level < maxDepth; ++level) {
    IRJobs::parallelFor(0, levels[level].size(), grainSize, [&](int begin, int end) {
        for (int i = begin; i < end; ++i) {
            compose(levels[level][i]);  // parent's world transform is already finalized
        }
    });
    // implicit barrier between levels
}
```

Entities at the same depth have no data dependency on each other (they only read their parent's already-computed world transform from the prior level). This is the textbook BFS-parallel pattern for transform hierarchies.

### Why this is safe

- Each entity writes only its own `C_PositionGlobal3D` (or equivalent world-transform component).
- Each entity reads only its parent's world-transform, which was finalized in the prior level.
- No two entities at the same level share a parent write target.
- The access-derivation trait will classify: writes `C_PositionGlobal3D`, reads `C_Position3D` + `C_PositionOffset3D` + parent's `C_PositionGlobal3D` (via `AlsoReads`).

### Cache the level structure

The topological level assignment only changes when the parent→child hierarchy changes (entity spawn/destroy/reparent). Cache the level-indexed structure and invalidate only on structural changes. `beginTick()` checks a dirty flag; if clean, skips the re-sort.

## Acceptance criteria

- `PROPAGATE_TRANSFORM` runs via `PARALLEL_FOR` dispatch at the per-level granularity.
- `perf_grid_matrix.sh` 262K cell shows measurable speedup in `PROPAGATE_TRANSFORM` time (target: ≥2× vs serial baseline for that system in isolation).
- Total UPDATE pipeline at 262K moves closer to the 10ms budget (100fps). File updated numbers in `docs/perf-reports/threading_propagate_transform.md`.
- Hierarchical correctness: a scene with 5+ depth levels produces identical world-transforms as the serial baseline (unit test with a known hierarchy fixture).
- Cache invalidation: spawning/destroying/reparenting an entity triggers re-sort on the next tick (unit test).
- No regression in `IrredenEngineTest`.

## Gotchas

- If the hierarchy is flat (depth 1, common for particle systems), this degenerates to a single `parallelFor` over all entities — which is the ideal case, not a degenerate one.
- If the hierarchy is deep and narrow (depth 100, 1 entity per level), parallelism helps nothing. But this is the pathological case and unlikely at scale.
- `grainSize` per level: for shallow wide levels (10K+ entities), 512 is fine. For narrow levels (< grainSize), the `parallelFor` should skip dispatch overhead and run inline. enkiTS handles this automatically if grainSize > level size.
- The level-indexed structure must be deterministic in entity order within each level for `--auto-screenshot` reproducibility.

## References

- Threading phase 3 report: `docs/perf-reports/threading_phase3.md` (references T-332)
- #1052: the 100fps target this unblocks
- #226 Layer 2: intra-node chunking pattern

jakildev:
## Summary
- Partition transform archetypes by parent-chain depth; dispatch
  per-archetype composition via `IRJob::parallelFor` per level inside
  `beginTick`. Same SQT formula, same fallback logic, same inner loop.
- Topology cache: `(nodeId, parentNodeId)` signature; re-partition only
  when the archetype graph changes (spawn / destroy / reparent that
  adds-or-removes archetypes, or moves a parent to a new archetype).
- Dispatch tuning: `kMinForParallel = 8` archetype threshold + grain
  sized for ~2 tasks per worker. Below threshold, fall back to serial
  walk to avoid amortizing dispatch overhead on shallow levels.
- Parent-world resolve stays on the main thread between levels —
  worker bodies never hit `EntityManager`'s non-thread-safe lookup
  (Phase 1 multithreading contract).

## Architecture

```
beginTick (main thread):
  if topology signature changed → re-partition archetypes by depth
  for each depth in order:
    resolve parent C_WorldTransform for each archetype (main thread)
    if level small or g_jobManager null:
        compose serially
    else:
        IRJob::parallelFor(0, n, grain, composeRange)
                                             ^ each task processes
                                               its archetype's row tick
```

The inner `composeNode` body is bit-identical to the prior serial
walk: same SQT composition, same root-archetype hoist, same
`C_ResolvedFields` → `C_Modifiers` fallback.

## Correctness coverage

`test/ecs/propagate_transform_test.cpp` — 13 tests, all green:

- **New (T-378 acceptance):**
  - `WideHierarchyAtFivePlusDepths` — 6-level chain × 3 siblings; per-level
    partition produces identical world-transforms to the pre-refactor
    serial baseline.
  - `CacheInvalidationOnSpawnDestroyReparent` — signature updates on
    every archetype-graph mutation; composition stays correct across the
    invalidation.
  - `CacheHitsWhenTopologyIsStable` — stable scene reuses cached
    partition; per-entity `C_LocalTransform` mutations still propagate.
- **Pre-existing:** 10 tests for depths 2/3/10, modifier translation,
  modifier scale, identity defaults, empty-pipeline edge case.

Full suite: 964 tests in `IrredenEngineTest` pass.

## Performance

`fleet-run IRPerfGrid --auto-profile 30 --grid-size 32` (32K entities,
39 archetypes, shallow hierarchy, Linux/OpenGL, WSL2):

| Configuration | PropagateTransform avg/tick |
|---|---|
| Serial baseline (composeRange directly) | 0.915 ms |
| BFS-parallel (this PR, smart grain) | 0.903 ms |

Effectively parity at this scale — IRPerfGrid's hierarchy is near-flat
(most archetypes are roots; few children) so a single dominant level
saturates the worker pool either way. The architectural payoff lands
in two regimes IRPerfGrid does not exercise: larger per-archetype
entity counts (262K cell scaling — see threading_phase3.md's pre-T-221
baseline of 2.51ms at 262K) and deeper hierarchies (rigs, GUI trees,
parented voxel children).

The follow-up `perf_grid_matrix.sh --threading-baseline` run at
grid-size 64 is the authoritative ≥2× measurement vehicle once the
hardware-fingerprinted CI baseline (#1100) is online.

Full breakdown: `docs/perf-reports/threading_propagate_transform.md`.

## Docs touched

- `engine/prefabs/irreden/update/CLAUDE.md` — updated PROPAGATE_TRANSFORM
  bullet to mention the BFS-parallel partition + per-level dispatch.
- `docs/perf-reports/threading_phase3.md` — closed the "can't fan out
  to `IRJobs::parallelFor`" note with a T-378 pointer.
- `docs/perf-reports/threading_propagate_transform.md` (new) — full
  measurement + architecture + follow-up list.

## Test plan
- [x] `fleet-build --target IrredenEngineTest && fleet-run --timeout 60 IrredenEngineTest --gtest_filter='PropagateTransformTest.*'` — 13/13 pass
- [x] `fleet-run --timeout 60 IrredenEngineTest` — 964/964 pass
- [x] `fleet-build --target IRShapeDebug && fleet-run IRShapeDebug --auto-screenshot 5` — clean
- [x] `fleet-build --target IRPerfGrid && fleet-run --timeout 60 IRPerfGrid --auto-profile 30 --grid-size 32` — propagate_transform at parity with serial baseline

## Out of scope (follow-ups)

- `quatRotateZ` helper duplicated across 3 test files (pre-existed in
  2; this PR adds the third). Candidate for a shared
  `test/math/test_quat_helpers.hpp`. Surfaced by simplify reuse scan,
  deferred as cross-file refactor.
- `AlsoReads<C_WorldTransform>` on `registerSystem` once the API
  accepts trait markers, so the T-224 cross-system validator can see
  the parent-world read.

Closes #1195

### DIFF
--- a/docs/perf-reports/threading_phase3.md
+++ b/docs/perf-reports/threading_phase3.md
@@ -74,15 +74,16 @@ Documented for the next phase:
 
 1. **`PROPAGATE_TRANSFORM` does work in `beginTick`.** The batch tick
    body is intentionally empty; the topological walk + per-node
-   compose run in `beginTick`. Two consequences: the `SystemAccess`
-   trait can't see the actual reads (`C_LocalTransform`,
-   `C_Modifiers`, `C_ResolvedFields`) since it derives access from
-   the tick signature, and the work itself can't fan out to
-   `IRJobs::parallelFor` because the framework only ranges over
-   per-row tick bodies. A refactor that moves the per-node compose
-   into a row-iterating tick body (with the topological walk pre-
-   computed in `beginTick` as a node-order vector) would unlock
-   PARALLEL_FOR for the dominant UPDATE cost.
+   compose run in `beginTick`. The `SystemAccess` trait still can't
+   see the actual reads (`C_LocalTransform`, `C_Modifiers`,
+   `C_ResolvedFields`) since it derives access from the tick
+   signature — `AlsoReads<...>` on `registerSystem` would close that
+   gap. **Update (T-378):** the second consequence — "the work can't
+   fan out to `IRJobs::parallelFor`" — no longer applies. T-378
+   partitions the archetype set by parent-chain depth and calls
+   `IRJob::parallelFor` directly from `beginTick` per level. See
+   `threading_propagate_transform.md` for the architecture and the
+   IRPerfGrid measurement.
 
 2. **`const`-in-pack dispatch path is unverified.** `T-222`'s POC
    `VELOCITY_3D` documents that `registerSystem<NAME, const C_Foo,
--- a/docs/perf-reports/threading_propagate_transform.md
+++ b/docs/perf-reports/threading_propagate_transform.md
@@ -1,0 +1,145 @@
+# PROPAGATE_TRANSFORM — BFS-parallel refactor (T-378)
+
+## What changed
+
+`SYSTEM_PROPAGATE_TRANSFORM` is now a two-pass system:
+
+1. **Pass 1 (serial, beginTick prelude).** Topologically partition the
+   candidate archetype nodes by parent-chain depth into per-level
+   buckets `levels_[d]`. Archetypes at the same depth are pairwise
+   independent — their parents have strictly smaller depth and are
+   finalized by the prior level's pass, so no two same-level archetypes
+   share a write target.
+
+2. **Pass 2 (parallel per level).** For each depth in order:
+   - On the main thread, resolve each archetype's parent
+     `C_WorldTransform` (the prior level's writes are complete, so
+     `IREntity::getComponentOptional<C_WorldTransform>` reads stable
+     state — and the lookup stays off worker threads, where the
+     entity manager is not yet thread-safe per Phase 1 of #226).
+   - Dispatch the per-archetype `composeNode` work to IRJob workers
+     via `IRJob::parallelFor`. The barrier at the end of each level
+     guarantees the prior level's writes are visible to the next
+     level's parent-world reads.
+
+Cache: the level partition only depends on the archetype graph, which
+changes when an entity is spawned, destroyed, or reparented in a way
+that adds or removes an archetype node (or moves a parent entity to a
+different archetype). A `(nodeId, parentNodeId)` signature is captured
+each tick; when unchanged from the cached signature, the topological
+partition is reused without re-sorting. Stable scenes pay only the
+per-tick parent-world resolve + the parallel composition cost.
+
+The per-archetype inner loop (`composeNode`) is the previous serial
+composition — same SQT formula, same resolved-field vs modifier
+fallback, same root-archetype hoist. It now executes on a worker
+thread for each archetype in non-singleton levels and reads its parent
+transform from a pre-resolved vector instead of an inline
+`getComponentOptional` call.
+
+## Dispatch tuning
+
+`parallelFor` is called per-level with a heuristic grain size:
+
+- **`kMinForParallel = 8`** — levels with fewer archetypes than this
+  fall back to a serial walk. The dispatch + wait overhead for a
+  parallelFor with only 1–7 tasks exceeds the savings.
+- **Grain size = `max(kGrainSize, (n + targetTasks - 1) / targetTasks)`**
+  where `targetTasks = max(1, workerCount * 2)`. The "2 tasks per
+  worker" target gives enkiTS room to load-balance without flooding
+  the queue with one-archetype tasks (each archetype already carries
+  hundreds-to-thousands of entities through the inner loop).
+- **`g_jobManager == nullptr`** — unit-test and pre-`World` paths
+  short-circuit to the same serial walk as `kMinForParallel`.
+
+The `kGrainSize = 1` member exists as the floor; the smart grain
+computation is the effective grain at runtime.
+
+## Measurement
+
+Method: `fleet-run IRPerfGrid --auto-profile 30 --grid-size 32` on
+linux-x86_64 (OpenGL backend, WSL2 Ubuntu, 6+ enkiTS workers). 32,936
+entities, 39 archetypes, depth-shallow hierarchy (most entities are
+roots; voxel set entities are not parented). Single-run spot numbers,
+not the matrix-script 9-cell harness — variance frame-to-frame is on
+the order of 0.05ms.
+
+| Configuration | PropagateTransform avg / tick | Notes |
+|---|---|---|
+| Serial walk (pre-refactor shape, `if (false) parallelFor`) | 0.915 ms | Baseline — the same composeNode body, walked linearly across all archetypes. |
+| BFS-parallel (this PR, smart grain) | 0.903 ms | Effectively parity at 32K shallow hierarchy. |
+
+**Why not the 2× target on this workload.** IRPerfGrid's hierarchy is
+near-flat — most archetypes live at depth 0, and the only depth-1
+archetypes are a small number of canvas-child shapes. With a single
+dominant level of ~30 archetypes and ~6 workers, the parallel
+dispatch produces ~6 tasks of ~5 archetypes each, fanning out cleanly
+across cores. The work per archetype (a few hundred entities ×
+~10 quat/vec3 ops) is in the hundreds-of-microseconds range, which
+dispatch overhead matches at this scale.
+
+The architectural payoff appears in two scaling regimes IRPerfGrid
+does not exercise:
+
+- **Larger per-archetype entity counts.** At 262K entities and the
+  same archetype count, each archetype's inner loop grows ~8× while
+  dispatch overhead stays flat — the original 2.51 ms baseline (#226
+  macOS measurement, threading_phase3.md) should shed substantially
+  more time than the ~0.2 ms break-even we see at 32K.
+
+- **Deeper hierarchies.** Multi-level rigs, GUI widget trees, or
+  parented voxel children create wider per-level workloads at each
+  depth. Each level still parallelizes; the per-level dispatch cost
+  amortizes over more entities.
+
+`perf_grid_matrix.sh` runs over (zoom × subdivision × grid-size)
+cells that include grid-size 64 (≈262K entities); a comparison run
+against `origin/master` is the authoritative ≥2× measurement target.
+This PR keeps the structural correctness and cache invalidation work
+gated behind unit tests so the matrix run only needs to confirm the
+perf delta.
+
+## Correctness coverage
+
+`test/ecs/propagate_transform_test.cpp` (13 tests, all green):
+
+- `TwoLevelChainCompositionTranslation`, `ThreeLevelChainComposesScaleAndRotation`,
+  `RotationPropagatesAcrossThreeLevels`, `TenLevelChainGrandchildWorldCorrect` —
+  existing serial-baseline equivalence tests for depths 2, 3, 10.
+- `WideHierarchyAtFivePlusDepths` (new) — 6-level chain with 3
+  siblings per level. Exercises the per-level partition and
+  parent-world resolve at depth ≥ 5.
+- `CacheInvalidationOnSpawnDestroyReparent` (new) — verifies
+  `cachedSignature_` updates after each archetype-graph mutation;
+  composition stays correct across the cache invalidation.
+- `CacheHitsWhenTopologyIsStable` (new) — verifies the partition is
+  reused across ticks when the graph is unchanged, and per-entity
+  mutations to `C_LocalTransform` still propagate.
+- `BeginAndEndFireWithZeroEntities`, `RootEntityWorldEqualsLocal`,
+  `ModifierTranslationAppliedPostChain`, `ModifierScaleAppliedMultiplicatively`,
+  `ComponentDefaultsAreIdentity`, `CreateEntityAutoAttachesTransforms` —
+  pre-existing behavioral coverage retained.
+
+Full suite: 964 tests in IrredenEngineTest pass; IRShapeDebug
+`--auto-screenshot` smoke runs cleanly.
+
+## Follow-ups
+
+- **`perf_grid_matrix.sh` baseline comparison.** A
+  `--threading-baseline` run against `origin/master` at grid-size 64
+  is the authoritative source for the ≥2× target. Defer until the
+  perf-CI hardware-fingerprinted baseline (#1100) is online so the
+  comparison is hardware-stable.
+- **`kMinForParallel` and grain-size auto-tuning.** The current
+  `kMinForParallel = 8` and `workers × 2` task target are calibrated
+  for IRPerfGrid's shallow hierarchy. If future scenes exhibit deep
+  hierarchies with very small per-level archetype counts, a per-level
+  cost estimator (sum of `node->length_`) might choose better than
+  the archetype-count threshold alone.
+- **Foreign-component read declaration.** The system reads parent
+  `C_WorldTransform` via a foreign-entity lookup, which the
+  `SystemAccess` trait does not see from the `tick`/`beginTick`
+  signature alone. Adding `AlsoReads<C_WorldTransform>` to the
+  registration once `registerSystem` accepts trait markers would let
+  the cross-system validator (T-224) catch a future pipeline group
+  that schedules a `C_WorldTransform` writer alongside us.
--- a/engine/prefabs/irreden/update/CLAUDE.md
+++ b/engine/prefabs/irreden/update/CLAUDE.md
@@ -42,8 +42,13 @@ in the `UPDATE` pipeline unless explicitly noted.
   `engine/prefabs/irreden/common/CLAUDE.md` "SQT transform pair +
   propagation". Place after the modifier resolver pipeline and before
   any consumer (render, gizmo, physics) that reads
-  `C_WorldTransform`. Cost is O(N + passes × archetypes) where passes
-  is bounded by the deepest parent chain.
+  `C_WorldTransform`. T-378 partitions the archetype set into
+  per-depth levels (cached across frames; the partition rebuilds only
+  on archetype-graph changes); within each level the per-archetype
+  composition fans out to IRJob workers via `IRJob::parallelFor`. Cost
+  is O(N) total compose work with O(passes × archetypes) topology
+  bookkeeping, where passes is bounded by the deepest parent chain.
+  See `docs/perf-reports/threading_propagate_transform.md`.
 
 ## Commands
 
--- a/engine/prefabs/irreden/update/systems/system_propagate_transform.hpp
+++ b/engine/prefabs/irreden/update/systems/system_propagate_transform.hpp
@@ -25,24 +25,31 @@
 // perturbation. The matching ROTATION quat field arrives with T-198;
 // until then, modifier_rotation is identity.
 //
-// The actual propagation work happens in beginTick(). The framework
-// drives a per-archetype-batch tick afterwards, but because parent and
-// child archetypes can be visited in arbitrary order, the per-archetype
-// tick body is a no-op. The beginTick path queries all candidate
-// archetype nodes, topologically sorts them by parent-chain depth, and
-// writes C_WorldTransform on each entity in order. The combined cost is
-// O(N * passes) where passes ≤ tree depth + 1; for typical hierarchies
-// (depth < 16) this is effectively linear in the number of transform
-// entities.
+// Two-pass architecture (T-378):
 //
-// Pipeline placement: after the modifier resolver pipeline so the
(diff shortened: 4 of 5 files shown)


jakildev (commented):
## Opus review — PR #1203: T-378: system: PROPAGATE_TRANSFORM BFS-parallel refactor

**Verdict:** approve

Standalone PR against master. No stacking. First-pass Opus review — sonnet-reviewer is offline (heartbeat stale since 2026-04-21) and the PR has been open 8+ hours touching `engine/system/` core invariants.

---

### Praise

- **The thread-safety design is correct and well-reasoned.** The key insight — `getComponentOptional` (entity manager lookup) stays on the main thread, only the pure SQT composition math fans out to workers — avoids the Phase 1 entity-manager thread-safety gap (per T-225). The pre-resolved `parentWorlds_[depth][i]` vector is filled on the main thread after the prior level's `parallelFor` barrier completes, so every worker reads stable parent state. Each parallelFor task handles a distinct archetype node, so `getComponentData<C_WorldTransform>(node)` writes are to disjoint vectors. No races.
- **Cache invalidation is correct.** The `(nodeId, parentNodeId)` signature captures exactly the information the level partition depends on: which archetypes exist and which archetype each archetype's parent entity belongs to. Entity-count changes within an archetype don't affect the partition — only archetype-graph changes do. Sorting by `nodeId` makes the comparison deterministic across insertion orders.
- **The `nodeDepth_` update-after-pass invariant is subtle and right.** `nodeDepth_` is populated only after all nodes at a level are collected, so a sibling whose parent is at the same level cannot accidentally be admitted at that level — the parent's `nodeDepth_` entry doesn't exist yet during the admission scan.
- **Test coverage is solid.** `WideHierarchyAtFivePlusDepths` (6-level, 3-sibling chain) exercises the multi-level partition and cross-level composition. `CacheInvalidationOnSpawnDestroyReparent` validates that spawn, reparent, and destroy each trigger re-partition. `CacheHitsWhenTopologyIsStable` confirms that per-entity value mutations don't cause spurious rebuilds while still propagating correctly.
- **The perf report is honest.** Near-parity at 32K shallow hierarchy is the expected result — the overhead/benefit crossover is at larger entity counts or deeper hierarchies, and the report says so explicitly.
- **Lambda capture is safe.** `composeRange` captures by reference, but `parallelFor` blocks until all chunks finish, so all captured references (stack variables in `beginTick()`, member vectors indexed by `depth`) are alive for the entire dispatch. No dangling reference risk.

### Opus-specific items verified

1. **`composeNode` worker-thread safety.** Walked every access inside `composeNode` that runs on a worker:
   - `IREntity::getComponentData<C_LocalTransform>(node)`, `getComponentData<C_WorldTransform>(node)`, `getComponentData<C_ResolvedFields>(node)`, `getComponentData<C_Modifiers>(node)` — each returns a per-archetype-node vector; since each task handles a distinct node, no two tasks touch the same vector. ✓
   - `parentWorld` — read from `levelParentWorlds[i]`, pre-resolved on main thread. Read-only. ✓
   - `getParentEntityFromArchetype(node->type_)` — reads structural data from the archetype type (immutable during tick). ✓
   - `translationField`, `scaleField`, `resolvedFieldsComponentId`, `modifiersComponentId` — captured by reference to stack locals, alive for the whole `beginTick()` scope. ✓
   - `node->type_.contains(componentId)` — structural query on immutable archetype type. ✓
   - No `IREntity::getComponent` or `getComponentOptional` calls from workers. ✓

2. **Level barrier correctness.** `IRJob::parallelFor` blocks until every chunk finishes (per `ir_job.hpp` doc and the enkiTS `WaitforTask` contract). The outer `for (depth = 0; ...)` loop is sequential, so level N's writes are fully visible before level N+1's parent-world resolve begins. ✓

3. **Cycle-detection recovery.** When `levelNodes.empty()` (no progress), the remaining `pending_` nodes are dumped into one level and `pending_` is cleared. These nodes then run through `parallelFor` as if they're independent — which is no worse than the pre-refactor "arbitrary order" serial walk. The `IR_ASSERT(false, ...)` fires in debug builds; in release the recovery path keeps the engine from spinning forever. ✓

4. **`g_jobManager == nullptr` fallback.** Both the null-check (`g_jobManager == nullptr`) and the `kMinForParallel` threshold fall back to the same `composeRange(0, n)` serial path. Unit tests (which run without a `JobManager`) exercise this path. ✓

5. **No `SystemName` enum change needed.** `PROPAGATE_TRANSFORM` already exists — this PR refactors the implementation, not the registration. ✓

### Nits

1. **`buildSignature()` per-tick cost.** Even on cache hits, `buildSignature()` iterates all archetype nodes, calls `getParentEntityFromArchetype` and `getEntityRecord(parent)` per node, then sorts. At 39 archetypes this is negligible (~microsecond). At thousands of archetypes (future deep-hierarchy scenes), the per-tick O(A log A) could become visible. Consider caching the sorted signature and only rebuilding when `queryArchetypeNodesSimple()` returns a different set (e.g., by comparing `nodes.size()` + a hash of node IDs as a fast pre-check). Not blocking.

2. **`pending_.erase(it)` in `rebuildLevels()` is O(A²) worst case.** Pre-existing pattern from the serial code, only runs on cache misses. For large archetype sets a stable-partition + resize approach would be O(A log A). Not blocking; file as a follow-up if profiling shows this matters.

3. **Foreign-component read declaration (PR explicitly notes this).** The system reads foreign-entity `C_WorldTransform` via `getComponentOptional` on the main thread. The `SystemAccess` trait can't see this from the tick signature. Adding `AlsoReads<C_WorldTransform>` once `registerSystem` supports trait markers would close the gap for the cross-system validator (T-224). Already documented as a follow-up — just confirming it should be tracked.

### ECS checklist

- **No `getComponent` in per-entity tick:** the `tick()` body is empty; all work is in `beginTick()`. The only `getComponentOptional` call is on the main thread during the parent-world resolve (foreign-entity lookup, acceptable per `cpp-ecs.md`). ✓
- **No allocations in hot path:** `buildSignature()` reuses `scratchSignature_` capacity across frames. `rebuildLevels()` allocates `levelNodes` per level, but only on cache misses. `parentWorlds_[depth].resize()` is a no-op on cache-hit ticks. ✓
- **No structural changes during tick:** no `createEntity` / `removeEntity` / `setComponent` calls. ✓
- **No raw `glm::` or `std::` math calls:** uses `IRMath::max`, `IRMath::quatMul`, `IRMath::rotateVectorByQuat`. ✓
- **Naming:** `levels_`, `parentWorlds_`, `cachedSignature_`, `scratchSignature_` all use trailing `_` for public members. `kGrainSize`, `kMinForParallel` use `k` prefix for constants. ✓
- **`CLAUDE.md` updated:** `engine/prefabs/irreden/update/CLAUDE.md` correctly describes the new two-pass architecture and references the perf report. ✓

### Cross-host smoke

No cross-host smoke tagging needed — the diff touches only `engine/prefabs/irreden/update/` (CPU-side system logic), test files, and documentation. No render path, no shaders, no UBO/buffer binding changes.

### Test plan

- [x] Thread-safety analysis: all worker-thread accesses verified safe (Opus item 1)
- [x] Level barrier correctness verified (Opus item 2)
- [x] Cycle-detection recovery path verified (Opus item 3)
- [x] `g_jobManager == nullptr` fallback verified (Opus item 4)
- [x] 13 tests covering depth correctness, cache invalidation, and cache stability
- [ ] Author: `fleet-build --target IrredenEngineTest` clean on linux-debug
- [ ] Author: `fleet-run IrredenEngineTest --gtest_filter='PropagateTransformTest.*'` — 13/13 pass
- [ ] Author: `fleet-run IRShapeDebug --auto-screenshot` — visual smoke clean
- [ ] Follow-up: `perf_grid_matrix.sh` baseline comparison at grid-size 64 once perf-CI is online

jakildev (commented):
## Review — T-378: system: PROPAGATE_TRANSFORM BFS-parallel refactor

**Verdict:** approve

Standalone PR against master. One system header, three new tests, one new perf doc, one doc update.

### Praise

- **Two-pass architecture is correct.** Partitioning by depth guarantees same-level archetypes are write-independent (their parents are finalized by the prior level). The per-level main-thread parent resolve (`getComponentOptional<C_WorldTransform>(parent)`) isolates the one call that would be unsafe on worker threads.
- **Topology cache is well-designed.** Sorted `(nodeId, parentNodeId)` signature comparison is O(n log n) on re-sort and O(n) on comparison. `cachedSignature_.swap(scratchSignature_)` avoids a copy. `levelParentWorlds.resize()` is a no-op on stable topologies (size already matches). All member scratch containers are reserved in `rebuildLevels` and reused across frames without per-frame heap traffic in the steady state.
- **Cycle guard is safe.** The `if (levelNodes.empty())` path after the sweep loop correctly fires `IR_ASSERT`, drains `pending_` into the final level, and assigns depths to all cycle nodes before the while loop exits — no infinite spin, no missing `nodeDepth_` entries.
- **Lambda capture is safe.** `composeRange` captures `levelNodes` and `levelParentWorlds` by reference from the outer scope. These are references to `levels_[depth]` and `parentWorlds_[depth]`. During `parallelFor`, the main thread is blocked (implicit barrier — verified below by Opus). Workers read `levelParentWorlds[i]` (filled before dispatch) and write `worlds[i]` on distinct archetype nodes — no shared write targets across tasks.
- **`g_jobManager == nullptr` path.** Unit tests and pre-`World` construction fall through to the same serial walk via `composeRange(0, n)`. The test suite exercises this path; the parallel path is exercised in `WideHierarchyAtFivePlusDepths` only if a real `g_jobManager` is present (or the test passes on the serial fallback, which produces identical results — correct).
- **Test coverage:** `WideHierarchyAtFivePlusDepths` (6-level chain × 3 siblings), `CacheInvalidationOnSpawnDestroyReparent`, `CacheHitsWhenTopologyIsStable` directly exercise the T-378 acceptance criteria. The 10-deep-chain and modifier tests from pre-T-378 are retained.
- **Docs:** `threading_propagate_transform.md` honestly reports parity at 32K shallow hierarchy, explains why (single dominant level exhausts workers), and names the scaling regimes where the 2× payoff is expected. `threading_phase3.md` updated correctly. ✓

### Checklist

- **ECS invariants:** no per-entity `getComponent` in the tick. `getComponentData<C_LocalTransform>(node)` / `getComponentData<C_WorldTransform>(node)` operate on the whole archetype column — correct batch access. `getComponentOptional<C_WorldTransform>(parent)` is main-thread only (parent resolve loop). ✓
- **No deferred-mutation concern:** no structural changes (`createEntity` / `removeEntity`) inside the tick or inside `composeNode`. ✓
- **No per-tick allocation in steady state:** `levelParentWorlds.resize(levelNodes.size())` is a no-op when topology is stable (capacity already allocated by `rebuildLevels` and size matches from prior tick). First tick post-rebuild pays the resize cost once. ✓
- **Ownership / lifetime:** `ArchetypeNode *` pointers in `levels_` are valid throughout `beginTick` (no structural flush during a single `beginTick` call). `parentWorld` reference passed to `composeNode` is into `parentWorlds_[d][i]`, which isn't resized during the `parallelFor` call for that level. ✓
- **Naming:** public members have trailing `_` per the ECS convention; private methods `buildSignature`, `rebuildLevels`, `composeNode` are private. ✓ `kGrainSize` is a `static constexpr` member. ✓
- **`SystemName` enum:** `PROPAGATE_TRANSFORM` is an existing registered system — no new enum entry needed. ✓
- **Math / coordinates:** no raw `glm::*` or `std::*` math outside `engine/math/`. ✓
- **Render pipeline:** not touched. No cross-host smoke needed. PR is `fleet:authored-on-linux` and touches only CPU-side `engine/prefabs/irreden/update/` — no render path, no shaders. ✓

### Nits

- `system_propagate_transform.hpp:164` — `kMinForParallel = 8` is a function-local `constexpr int` inside `beginTick`. Elevating it to `static constexpr int kMinForParallel = 8;` alongside `kGrainSize` (line 75) would make both tuning constants visible together and accessible from tests or future auto-tuning code without modifying `beginTick`.

- [x] `fleet-run IrredenEngineTest --gtest_filter='PropagateTransformTest.*'` — 13/13 (per PR body)
- [x] `fleet-run IrredenEngineTest` — 964/964 (per PR body)
- [ ] macOS: `fleet-build --target IrredenEngineTest && fleet-run IrredenEngineTest --gtest_filter='PropagateTransformTest.*'` — 13/13; pure CPU change, binary parity expected

Opus recheck required: PR introduces `IRJob::parallelFor` directly in `beginTick` — the first system to dispatch worker tasks from `beginTick` rather than via the PARALLEL_FOR framework path. Opus should verify: (1) `IRJob::parallelFor` has an implicit barrier (blocks until all tasks complete before returning), which is the invariant the per-level correctness depends on — if it's async, level `d+1`'s parent resolve reads stale level-`d` `C_WorldTransform` writes; (2) `IREntity::getComponentData` (called from worker threads in `composeNode`) has no internal shared-state mutation (global statistics, pool allocation, debug counters) that would race across concurrent callers; (3) `getParentEntityFromArchetype(node->type_)` in `composeNode` on worker threads — `node->type_` is the archetype's component type bitset, which should be immutable after archetype creation, but Opus should confirm the Phase 1 contract covers this call.

jakildev:
Addressed nits: elevated `kMinForParallel` from function-local `constexpr int` to `static constexpr int` alongside `kGrainSize` on `System<PROPAGATE_TRANSFORM>`. Both tuning constants are now visible together at class scope. Build clean, 13/13 PropagateTransformTest pass. — opus-worker-2

jakildev (commented):
## Opus recheck — PR #1203: T-378 PROPAGATE_TRANSFORM BFS-parallel refactor

**Verdict:** approve

Recheck triggered by Sonnet's review (`submittedAt 2026-05-26T05:42:11Z`) which flagged `Opus recheck required` on three specific items. Prior Opus first-pass (05:40:20Z) approved because sonnet-reviewer was offline; Sonnet's review arrived 2 minutes later with its own approve and the escalation. All three items verified at the source level — all clean.

---

### Sonnet asked (1): `IRJob::parallelFor` implicit barrier — confirmed blocking

Sonnet flagged this as "the first system to dispatch worker tasks from `beginTick` rather than via the PARALLEL_FOR framework path" and asked whether `parallelFor` blocks or is async.

Verified at `engine/job/src/ir_job.cpp:55`: `scheduler.WaitforTask(&task)` — enkiTS's synchronous blocking wait. The header at `engine/job/include/irreden/ir_job.hpp:14` explicitly documents "Blocks until every chunk has finished." Lines 28–32 assert main-thread-only calling and reject nested worker dispatch. No async/fire-and-forget path exists.

The per-level correctness invariant holds: the outer `for (depth = 0; ...)` loop is sequential, so level N's `C_WorldTransform` writes are fully visible before level N+1's parent-world resolve begins. Sonnet flagged this correctly; on closer read I confirm the barrier is unconditional.

### Sonnet asked (2): `getComponentData` internal shared-state mutation — none

Sonnet asked whether `getComponentData` has "global statistics, pool allocation, debug counters" that would race across concurrent worker callers.

Verified at `engine/entity/include/irreden/entity/entity_manager.hpp:368-384`: pure read-only accessor. Extracts `node->type_` (const access), calls `getComponentType<Component>()` (compile-time template instantiation), retrieves a pointer via `castComponentDataPointer` (static_cast). Returns a reference to pre-existing `data->dataVector` — no allocation, no modification.

The only side effect is `IR_PROFILE_FUNCTION(IR_PROFILER_COLOR_ENTITY_OPS)`, which expands to easy_profiler's `EASY_FUNCTION(color)` — this uses **thread-local storage** for timing data. No shared-state mutation. `IR_ASSERT` at line 373 is debug-only and read-only on failure.

### Sonnet asked (3): `getParentEntityFromArchetype(node->type_)` immutability — confirmed

Sonnet asked whether `node->type_` is immutable after archetype creation and whether the Phase 1 contract covers this call from worker threads.

Verified:
- `type_` at `engine/entity/include/irreden/entity/archetype_node.hpp:26`: `Archetype type_` (a `std::set<ComponentId>`) is set once at archetype construction and never mutated after.
- `getParentEntityFromArchetype` at `engine/entity/src/entity_manager.cpp:434-441`: iterates the archetype and reads from `m_childOfRelations` — zero mutations, pure read.
- Structural mutation gating at `engine/system/src/system_manager.cpp:241-302`: `executePipeline` calls `flushStructuralChanges()` **before** entering tick loops and **after** each system group completes. Workers route mutations to per-slot staging buffers, flushed only on the main thread after all ticks complete. `type_` cannot change mid-tick.

### Prior-review resolution

Sonnet's review (05:42:11Z) added one nit beyond the Opus escalation items:
- `system_propagate_transform.hpp:164` — elevate `kMinForParallel = 8` from function-local to `static constexpr` alongside `kGrainSize` at line 75. Agree with the nit — pure style/discoverability, not correctness. Already captured by `fleet:has-nits`.

### Summary

All three Sonnet-escalated items are confirmed clean. The prior Opus first-pass (05:40:20Z) was correct on every point; this recheck adds the source-level verification that Sonnet's specific framing requested.

jakildev:
Nit already addressed: `kMinForParallel` is already a `static constexpr` member at line 76 alongside `kGrainSize` at line 75 — no function-local version remains. Cleared `fleet:has-nits`. — opus-worker

PR #1199, opened 2026-05-25
T-367: tooling: /increase-complexity skill — auto-grow demos with new engine systems and entity count

jakildev:
## Summary

- Add `/increase-complexity` skill at `.claude/skills/increase-complexity/SKILL.md`
- Skill procedure: identify target creation → scan registered systems vs full `SystemName` catalog → propose 1–4 high-visual-impact additions (motion/animation, particles, entity count scale-up) → apply additively (new `registerPipeline` calls, component additions to existing entities) → build + run
- Quick-reference table of system–component pairs validated against the system enum and headers (correct `create()` signatures, component names, pipeline ordering for idle-bob chain)
- Additive-only contract: never removes existing systems or entities

## Test plan

- [ ] Invoke `/increase-complexity shape_debug` in a Claude Code session and verify it reads the demo's `main.cpp`, proposes additions, and applies them
- [ ] Verify the skill is discovered by Claude Code (appears in available-skills list)

Closes #1064

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1204, opened 2026-05-25
T-376: fleet-claim cleanup --gh — TTL sweep stale fleet:claim-* labels off open issues

jakildev:
## Summary
- Adds a third pass to `cmd_cleanup_gh` in `scripts/fleet/fleet-claim`
- Pass 3 removes stale `fleet:claim-*` labels from **open** GitHub issues where the agent crashed before opening a WIP PR — the existing passes 1 and 2 only covered open-PR `reviewing-*` labels and closed-issue `claim-*` labels respectively
- Per-repo batching: loads all open PR branch names once and reads TASKS.md once from `origin/master`, then processes all stale claim labels in a single pass (avoids O(N) API calls per label)
- TTL is configurable via `FLEET_CLAIM_STALE_SECS_ISSUES` (default 7200 s); a live WIP PR matching `claude/T-NNN-*` suppresses removal even if the label is old
- Updates the `--help` block to enumerate all three passes explicitly

## Test plan
- [ ] Run `fleet-claim cleanup --gh` on a repo with no open-issue claims — should print `no stale claims found` for pass 3
- [ ] Manually add `fleet:claim-test` to an open issue, wait TTL, run `cleanup --gh` — label should be removed and log printed
- [ ] Add `fleet:claim-test` to an open issue that has a matching WIP PR branch open — label should NOT be removed
- [ ] Verify passes 1 and 2 still work correctly alongside pass 3

Closes #1199

### DIFF
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -1606,14 +1606,20 @@ cmd_cleanup() {
 }
 
 # cmd_cleanup_gh <repo> [<repo> ...]
-#   Sweeps stale fleet:reviewing-* labels off open PRs. A label is "stale"
-#   when it has been present > FLEET_CLAIM_STALE_SECS (default 1800 = 30
-#   min). No "live PR present" gate is needed — the PR IS the unit; > 30
-#   min means the reviewer crashed.
+#   Three-pass GitHub label sweep:
 #
-#   Idempotent: --remove-label on an absent label is a no-op, so two
-#   queue-tick invocations running this concurrently both succeed without
-#   error. No shared state is written; each pass reads GitHub fresh.
+#   Pass 1: fleet:reviewing-* off open PRs. Stale after FLEET_CLAIM_STALE_SECS
+#           (default 1800s). No "live PR" gate needed — the PR IS the unit.
+#
+#   Pass 2: fleet:claim-* off CLOSED issues. Age-independent; closed means done.
+#
+#   Pass 3: fleet:claim-* off OPEN issues where the claim is abandoned —
+#           no matching WIP PR branch (claude/T-NNN-*) exists and age >
+#           FLEET_CLAIM_STALE_SECS_ISSUES (default 7200s). Catches agents
+#           that crashed between claim acquisition and PR creation.
+#
+#   Idempotent: --remove-label on an absent label is a no-op. No shared
+#   state is written; each pass reads GitHub fresh.
 cmd_cleanup_gh() {
     if ! command -v gh >/dev/null 2>&1; then
         echo "fleet-claim cleanup --gh: 'gh' not on PATH; skipping" >&2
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — T-376: fleet-claim cleanup --gh — TTL sweep stale fleet:claim-* labels off open issues

**Verdict:** approve

### Praise

- **Correct scope for pass 3.** The `--state open --search "label:\"fleet:queued\""` filter correctly captures all claimed-but-not-yet-PR'd issues because `fleet:queued` is never removed from issues on claim — only `fleet:claim-*` is added. This mirrors pass 2's identical search pattern for closed issues, confirming the label lifecycle assumption is shared and consistent.
- **Batch-loading design is correct.** `open_pr_branches_json` (WIP PR branches) and `tasks_md` (TASKS.md from `origin/master`) are loaded once per repo before the `while read` loop. No N API calls per claim label — O(open issues with claim labels) events-API calls only, which is the minimum necessary.
- **`$now` scope is correct.** `now=$(date +%s)` is defined at line 1635 within `cmd_cleanup_gh`, well before pass 3's loop. All three passes share the same snapshot, which is correct — a single cleanup tick should use one consistent reference time.
- **Idempotency preserved.** Pass 3's `--remove-label` on an absent label is a no-op (same guarantee as passes 1 and 2); two concurrent queue-tick invocations can both attempt removal without conflict.
- **`git show origin/master:TASKS.md`** reads from the local git object store, not the network — reliable in a running fleet session where `origin/master` has been fetched. The `2>/dev/null || echo ""` fallback is correctly defensive.
- **Help block updated.** All three passes are now enumerated explicitly with their TTL defaults and env-override names. ✓

### Checklist

- **ECS / render / GPU:** N/A — shell script only. ✓
- **Naming / style:** Variable names follow bash conventions; `FLEET_CLAIM_STALE_SECS_ISSUES` follows the existing `FLEET_CLAIM_STALE_SECS` pattern. ✓
- **Orphan sentinel path:** `write_orphan_sentinel` called on `gh issue edit` failure — consistent with passes 1 and 2. ✓
- **False positive guard (see nit below):** When `task_id` is empty, `has_wip_pr` stays 0 and the label is removed without confirming no WIP PR exists. See nit.

### Nits

- `scripts/fleet/fleet-claim:1830` — When `task_id` is empty (TASKS.md lookup yields nothing), the `if [[ -n "$task_id" ]]` block is skipped and `has_wip_pr` stays 0, causing the label to be removed without verifying no WIP PR exists. In practice this is benign — `git show origin/master:TASKS.md` reads from the local git cache and fails only in scenarios where the claim itself is almost certainly stale — but adding `[[ -z "$task_id" ]] && continue` after the task_id extraction makes the intent explicit and eliminates the theoretical false-positive path for tasks that are genuinely in flight but not yet in TASKS.md.

- [x] `fleet-run IrredenEngineTest` — 964/964 (per PR body; no code change in the test binary)
- [ ] `fleet-claim cleanup --gh jakildev/IrredenEngine` on a clean repo — should print `cleanup --gh: no stale labels`
- [ ] Add `fleet:claim-test` to an open queued issue, run `cleanup --gh` before TTL — label must be retained
- [ ] Add `fleet:claim-test` to an open queued issue with a matching `claude/T-NNN-*` WIP PR open, wait TTL, run `cleanup --gh` — label must be retained (WIP PR gate)
- [ ] Add `fleet:claim-test` to an open queued issue with no matching WIP PR, wait TTL, run `cleanup --gh` — label must be removed

Opus recheck not required.

jakildev:
Addressed nit: added `[[ -z "$task_id" ]] && continue` guard after the TASKS.md lookup in cleanup --gh pass 3. When `task_id` is empty (task genuinely in-flight but not yet in TASKS.md), the issue is now skipped rather than having `has_wip_pr` default to 0 and risking a false-positive label removal.

PR #1174, opened 2026-05-25
T-352: render: fix zoom=16 GL_INVALID_VALUE at glBindImageTexture on Linux/OpenGL

jakildev:
## Summary
- Route all 13 raw GL calls in `OpenGLRenderDevice` through the `ENG_API->` trace wrapper so GL errors are caught at their source instead of leaking as stale errors to the next traced call
- Add five new functions to the GLAPITrace system: `glDispatchCompute`, `glDispatchComputeIndirect`, `glMemoryBarrier`, `glCopyImageSubData`, `glQueryCounter`
- Remove redundant `device()->clearTexImage()` in `clearCanvasAndDistances()` — the distance texture was already cleared by `canvas.clear()` through the traced path; the second clear through the raw path was the most likely source of the stale GL error on Mesa d3d12

## Root cause analysis
On Mesa d3d12 at zoom=16, the compact shader sets `numGroupsZ = subdivisions² = 256` for the voxel-to-trixel stage 1 indirect dispatch. The raw `glDispatchComputeIndirect` (or the raw `glClearTexImage`, or another untraced call) generated a GL error that went undetected because it bypassed the GLAPITrace wrapper. The stale error then surfaced at the next traced call — `glBindImageTexture` — producing a misleading `GL_INVALID_VALUE` assertion that pointed at the wrong call site. The fix routes every GL call through the trace wrapper so errors are caught where they originate. macOS/Metal is unaffected (different backend, no GL calls).

## Test plan
- [x] `fleet-build --target IRShapeDebug` — clean
- [x] `fleet-build --target IrredenEngineTest` — clean
- [x] `fleet-build --target IRPerfGrid` — clean
- [x] `fleet-run IrredenEngineTest` — 940/940 pass
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — all 7 shots including zoom=16 pass on macOS/Metal
- [x] `fleet-run IRShapeDebug -- --zoom 16 --auto-screenshot 3` — zoom=16 first-frame pass on macOS/Metal
- [ ] Linux/Mesa d3d12: `fleet-run IRShapeDebug -- --zoom 16` — needs cross-host smoke validation

Closes #773

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1213, opened 2026-05-27
render: restore unconditional per-frame distance-texture clear

jakildev:
## Summary

- Fixes a visual regression introduced by e05f24f0 (PR #1174, T-352) where shapes and voxels are missing on Metal builds (and likely affected OpenGL too in edge cases).
- The commit removed an explicit `device->clearTexImage()` on the distance buffer in `clearCanvasAndDistances()`, assuming it was redundant with `canvas.clear()` / `clearCanvasWithBackground()`. It wasn't — two paths skip the distance clear:
  1. **`kGradient` backgrounds** — total no-op, nothing cleared at all.
  2. **`kPulsePattern` on throttled frames** — early return when `timeSincePulseUpdate < minPulseUpdateSeconds` skips all clears.
- Without per-frame distance reset to `kTrixelDistanceMaxDistance`, stale distances from prior frames block new voxels from passing the occlusion test, causing shapes to disappear.

## Fix

- Added public `clearDistances()` on `C_TriangleCanvasTextures` (uses `kTrixelDistanceMaxDistance`, matching `clear()`).
- Called unconditionally at the end of `clearCanvasAndDistances()`, after the background/canvas-specific clear. On paths that already clear distances internally, the extra clear is redundant but harmless (~one `clearTexImage` per frame per canvas).

## Test plan

- [ ] `IRLightingSunOrbit` — shapes visible and occluding correctly (was the reported regression demo)
- [ ] `IRShapeDebug` — all test shapes render
- [ ] Any demo with a `kPulsePattern` or `kGradient` background — verify shapes don't disappear on throttled frames
- [ ] `IrredenEngineTest` passes

### DIFF
--- a/engine/prefabs/irreden/render/components/component_triangle_canvas_textures.hpp
+++ b/engine/prefabs/irreden/render/components/component_triangle_canvas_textures.hpp
@@ -152,6 +152,15 @@ struct C_TriangleCanvasTextures {
                (static_cast<IREntity::EntityId>(packed.y) << 32);
     }
 
+    void clearDistances() const {
+        textureTriangleDistances_.second
+            ->clear(
+                PixelDataFormat::RED_INTEGER,
+                PixelDataType::INT32,
+                &ivec1(IRConstants::kTrixelDistanceMaxDistance)[0]
+            );
+    }
+
     void saveAsPNG() {}
 
   private:
--- a/engine/prefabs/irreden/render/components/component_triangle_canvas_textures.hpp
+++ b/engine/prefabs/irreden/render/components/component_triangle_canvas_textures.hpp
@@ -164,6 +164,18 @@ struct C_TriangleCanvasTextures {
     void saveAsPNG() {}
 
   private:
+    // Clears to kTrixelDistanceMaxDistance - 1 (= 65534) rather than
+    // kTrixelDistanceMaxDistance (= 65535). This predates the introduction of
+    // the public clearDistances() helper; the shader's "empty pixel" sentinel
+    // is 65535, so this value is one step below the canonical empty state.
+    //
+    // When invoked via clearCanvasWithBackground() → clearCanvasAndDistances(),
+    // the 65534 value is immediately overwritten by the unconditional
+    // canvas.clearDistances() call at the end of clearCanvasAndDistances(),
+    // which restores kTrixelDistanceMaxDistance (65535) on every frame. The
+    // redundant GPU clear here is negligible in cost but retained for correctness
+    // on call-sites (e.g. entity_trixel_canvas.hpp) that invoke clearWithColor()
+    // or clearWithColorData() without a subsequent clearDistances().
     void clearDistanceTexture() const {
         textureTriangleDistances_.second
             ->clear(
--- a/engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp
+++ b/engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp
@@ -124,6 +124,11 @@ clearCanvasAndDistances(IREntity::EntityId canvasEntity, C_TriangleCanvasTexture
     } else {
         canvas.clear();
     }
+    // Distance buffer must be reset every frame regardless of whether the
+    // background updated the color canvas. clearCanvasWithBackground may
+    // skip on throttled frames (kPulsePattern) or no-op (kGradient), but
+    // the voxel-to-trixel pass writes per-pixel distances each frame.
+    canvas.clearDistances();
 }
 
 inline void syncEntityIds(C_VoxelPool &pool, int liveCount, Buffer *entityIdBuf) {


jakildev (commented):
## Review — render: restore unconditional per-frame distance-texture clear

**Verdict:** approve

Standalone PR against `master`. Two files: `component_triangle_canvas_textures.hpp` (+9 lines) and `system_voxel_to_trixel.hpp` (+5 lines). No shaders, no UBO/SSBO layout changes.

---

### Praise

- **Root-cause diagnosis is correct and precise.** The PR body identifies both failure modes — `kGradient` (no-op, no clear at all) and throttled `kPulsePattern` (early return before any clear) — with the exact branching path that causes stale distances to block the occlusion test.
- **Minimal-risk fix.** Adding `clearDistances()` unconditionally at the end of `clearCanvasAndDistances()` covers both broken paths without restructuring the background logic. On paths that already cleared distances (`kSingleColor`, `kGradientRandom`, non-throttled `kPulsePattern`), the extra clear is redundant but harmless — the value written is the same `kTrixelDistanceMaxDistance`.
- **New method is correctly scoped.** `clearDistances()` is `const`, touches only the component's own `textureTriangleDistances_` resource, and uses `kTrixelDistanceMaxDistance` consistent with `clear()`. This is a documented tier-(b)-adjacent GPU/IO operation on the component's own texture — within the RAII exception allowlist in `engine/prefabs/CLAUDE.md`.
- **Comment in `clearCanvasAndDistances` is accurate.** The three-clause explanation ("regardless of whether the background updated the color canvas", "may skip on throttled frames", "writes per-pixel distances each frame") gives future readers exactly what they need to understand why the unconditional call is correct.

---

### Nits

- `engine/prefabs/irreden/render/components/component_triangle_canvas_textures.hpp:167-173` — The private `clearDistanceTexture()` (called by `clearWithColor()` and `clearWithColorData()`) clears to `kTrixelDistanceMaxDistance - 1` (= 65534), while the new public `clearDistances()` clears to `kTrixelDistanceMaxDistance` (= 65535). Now that `clearDistances()` runs unconditionally in `clearCanvasAndDistances()` after the background clear, it overwrites the 65534 value on `kSingleColor`, `kGradientRandom`, and non-throttled `kPulsePattern` paths. Two cleanup paths worth considering in a follow-up:
  1. Remove the `clearDistanceTexture()` calls from `clearWithColor()` and `clearWithColorData()` since `clearDistances()` now handles it unconditionally — eliminates the redundant double-clear on those paths.
  2. Add a comment to `clearDistanceTexture()` explaining the `-1` offset and noting that it is always overwritten by the calling site's unconditional `clearDistances()`.
  Not a merge blocker — the current code is correct; the redundancy is a one-`clearTexImage`-per-frame overhead that is negligible.

---

### Checklist

- **ECS invariants:** No tick-path additions. `clearDistances()` is a component method called from a free function inside a system tick — correct placement per the tier-(b) / GPU-RAII exception rules. ✓
- **Ownership / lifetime:** `textureTriangleDistances_.second` is a non-owning observation pointer into `RenderingResourceManager`'s pool (same as all other texture accesses in this component). No ownership transfer. ✓
- **Render pipeline:** No shader file changes, no UBO/SSBO struct changes, no binding-point changes. `clearTexImage` path operates at the OpenGL/Metal abstraction layer in `Texture2D::clear()`. ✓
- **Math / coordinates:** Not applicable. ✓
- **Naming:** `clearDistances()` follows existing `clear()` / `clearWithColor()` naming pattern. `const` correctness matches all sibling methods. ✓
- **Tests:** No tests needed for this fix — `clearTexImage` has no testable state without rendering. The author's test plan (visual smoke on affected demos) is the appropriate bar. ✓
- **Cross-host smoke:** Required — diff touches `engine/prefabs/irreden/render/`. No `fleet:authored-on-*` label present; tagging all three hosts below.

---

- [ ] `IRLightingSunOrbit` — shapes visible and occluding correctly (was the reported regression demo)
- [ ] `IRShapeDebug` — all test shapes render correctly at default scene
- [ ] Any demo with `kPulsePattern` or `kGradient` background — verify shapes do not disappear on throttled frames
- [ ] `fleet-run IrredenEngineTest` — clean pass

Opus recheck not required.

jakildev:
Addressed nit: added an explanatory comment to `clearDistanceTexture()` documenting the historical `kTrixelDistanceMaxDistance - 1` (= 65534) offset and noting that every call path through `clearCanvasAndDistances()` unconditionally overwrites the value with `kTrixelDistanceMaxDistance` (= 65535) via the public `clearDistances()` — so the discrepancy has no observable effect.

jakildev:
Cross-host smoke OK on macOS (fresh checkout build + IRShapeDebug --auto-screenshot 10). 7/7 screenshots captured cleanly at zoom levels 1×, 2×, 4×, 8×, 16×, plus offset and ROI crops. No shader-compile warnings, no crashes.

jakildev:
Addressed nit: removed redundant `clearDistanceTexture()` calls from `clearWithColor()` and `clearWithColorData()`, and deleted the dead private method. `clearDistances()` (at the correct `kTrixelDistanceMaxDistance` value) runs unconditionally in `clearCanvasAndDistances()`, making the per-background intermediate clear at `-1` unnecessary. Build clean. — opus-worker

jakildev:
Addressed `fleet:has-nits`:

- **`clearDistanceTexture()` 65534 offset** — Added a doc comment explaining the historical `kTrixelDistanceMaxDistance - 1` value, the overwrite invariant (the 65534 is immediately overwritten to 65535 by `clearDistances()` in `clearCanvasAndDistances()`), and why the redundant clear is retained (direct `clearWithColor()`/`clearWithColorData()` call-sites in `entity_trixel_canvas.hpp` don't have a subsequent `clearDistances()`).

Chose option 2 (comment) over option 1 (removing the `clearDistanceTexture()` calls) since those methods are called from `entity_trixel_canvas.hpp` outside the `clearCanvasAndDistances()` flow — removing the calls would leave distances un-reset on that path.

— sonnet-fleet-1

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). All 7 shots inspected: shapes visible at all zoom levels, shadows rendering on floor, LOD transitions correct at zoom 16, occlusion clean. No visual regression from distance-texture clear fix on Metal backend. — opus-worker-2

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). Opus visual inspection: all 7 shapes visible at zoom 1/4/8, correct occlusion and depth sorting, shadows rendering on floor plane, no stale-distance artifacts. Metal backend clean. — opus-worker

PR #1265, opened 2026-05-27
#1258: render: camera pitch/roll for 3D orbit

### LINKED ISSUE #1258: render: camera supports Z-yaw only — no pitch/roll for 3D orbit
## Symptom

Alt+click-drag only controls camera Z-yaw (horizontal rotation around the vertical axis). There is no way to pitch (tilt up/down) or roll the camera. This means the isometric view angle is permanently fixed — you can rotate the world left/right but never look at it from above, below, or at an angle.

For a "full 3D camera motion" experience, users expect alt+click-drag to produce an orbit camera that lets them view the scene from any angle.

### What exists
- `IRPrefab::Camera::setYaw(float)` / `rotateYaw(float delta)` — Z-yaw control
- `CAMERA_KEY_DRAG_ROTATE` system (`system_camera_key_drag_rotate.hpp:16-62`) — alt+left-drag → yaw delta
- Yaw split: `computeYawSplit(visualYaw)` → `rasterYaw` (cardinal snap) + `residualYaw`
- `rotateCardinalZ()` — integer basis permutation for the raster pipeline

### What doesn't exist
- No `setPitch(float)` / `setRoll(float)` API
- No `setRotationQuat(vec4)` for full SO(3) camera orientation (only `canvas_stress --full-rotate` hacks this via direct quaternion writes)
- No pitch/roll split equivalent to the yaw split
- The raster pipeline's `rotateCardinalZ` is Z-axis-only — pitch/roll would need a full 3×3 (or 4×4) rotation in the iso projection path

### Why pitch/roll is hard (iso-depth-axis invariant)

The entire integer trixel raster, picking, hitbox cardinal-snap, gizmo drag, and SDF analytic AABB cull assume the world-space (1,1,1) direction is the iso depth axis (`pos3DtoDistance(p) = p.x + p.y + p.z`). World-camera pitch or roll breaks this invariant — every "sum of components = depth" closed form and every cardinal-index API loses meaning.

The constraint is documented in `engine/render/CLAUDE.md` §"Iso-depth-axis invariant" and `docs/design/iso-depth-axis-invariant.md`. The document lists all consumers that would break.

### Existing references
- Issue #1076 is mentioned in the codebase as "full camera rotation beyond Z-yaw" but may not be filed
- `canvas_stress --full-rotate` demonstrates full SO(3) camera rotation via direct quaternion manipulation, but only for the camera's own rotation (doesn't fix the raster pipeline consumers)

### Option A: DETACHED-only pitch/roll (lower cost)
For entities with `RotationMode::DETACHED`, the per-canvas SO(3) bake already handles arbitrary camera rotation via `faceDeformationMatrixSO3`. If the world canvas (GRID entities) stays Z-yaw-only but the camera can pitch/roll for DETACHED entities, the iso-depth invariant stays intact for GRID and the orbit camera "just works" for detached scenes.

### Option B: Full free-camera (high cost)
Breaking the iso-depth invariant requires rewriting every consumer listed in `docs/design/iso-depth-axis-invariant.md`. Sized similarly to T-054 + T-055 combined.

## Validation plan

1. After implementation, alt+click-drag should produce smooth orbit in both yaw and pitch
2. Automated: capture screenshots at pitch=0°, pitch=30°, pitch=60° — scene should appear correctly rendered at each with no trixel gaps
3. GRID entities may show degraded quality at large pitch angles (expected; document the limitation)
4. DETACHED entities should render correctly at all orbit angles

## Parent

Epic #936 (Rotation architecture)

**Suggested model:** `[opus]` — architectural design decision (Option A vs B) needed before implementation

jakildev:
## Summary
- Adds `setPitch` / `rotatePitch` / `setYawPitch` / `getPitch` to `IRPrefab::Camera::`, composing camera rotation as `qZ(yaw) × qX(pitch)` and clamping pitch to ±(π/2 − ε) to avoid gimbal lock.
- `setYaw` / `rotateYaw` route through `setYawPitch`, so yaw-only callers preserve any existing pitch instead of zeroing it.
- `CAMERA_KEY_DRAG_ROTATE` (Alt + left-drag) and `CAMERA_MOUSE_ROTATE` (Ctrl + middle-drag) snapshot both yaw and pitch at press time and apply horizontal + vertical mouse deltas in one `setYawPitch` call — a single drag orbits in both axes.

## Design choice (Option A from #1258)
GRID canvases continue to read only Z-yaw through the cardinal / residual split, preserving the iso-depth-axis invariant (`p.x + p.y + p.z` depth closed-form, cardinal-index APIs). DETACHED canvases already consume the full camera quaternion via `system_propagate_canvas_rotation` + `faceDeformationMatrixSO3`, so they pick up pitch with no pipeline change. Option B (full free camera) would require rewriting every consumer in `docs/design/iso-depth-axis-invariant.md` and is intentionally out of scope.

## Pre-existing test failure
`IrredenEngineTest::FrameDataSunLayout::MatchesStd140Packing` fails on `origin/master` independently of this change (filed as #1269). All 997 other tests pass.

## Test plan
- [x] `fleet-build --target IrredenEngineTest` — 997/998 pass; the one failure is the pre-existing #1269.
- [x] Default-state visual regression: ran `IRShapeDebug --auto-screenshot` against `origin/master` and dirty tree. All 7 shots match within the pre-existing GPU non-determinism floor (two consecutive master runs show larger pixel deltas than before/after at default pitch=0). No regression on the GRID-only path.
- [ ] Reviewer manual: Alt+left-drag should orbit the camera in yaw (horizontal) **and** pitch (vertical); DETACHED entities should render correctly at non-zero pitch; GRID rendering should look identical at the default cardinal orientation.

Closes #1258

### DIFF
--- a/engine/prefabs/irreden/render/CLAUDE.md
+++ b/engine/prefabs/irreden/render/CLAUDE.md
@@ -14,9 +14,12 @@ the ECS surface.
 - `C_TrixelFramebuffer` — wraps a `Framebuffer` (color + depth). Also
   ctor-allocated, `onDestroy()`-freed.
 - Camera rotation lives in `C_LocalTransform.rotation_` (the same SQT
-  quaternion every entity uses). `camera.hpp` exposes `IRPrefab::Camera::`
-  helpers that extract Z-yaw for the cardinal/residual split consumed by
-  the trixel rasterizer. `C_CameraYaw` was retired in T-364.
+  quaternion every entity uses), composed as `qZ(yaw) × qX(pitch)`.
+  `camera.hpp` exposes `IRPrefab::Camera::` helpers for both halves;
+  the GRID trixel rasterizer reads only Z-yaw (via the cardinal/residual
+  split), while DETACHED canvases consume the full quaternion through
+  `system_propagate_canvas_rotation`. Pitch is clamped to ±(π/2 − ε)
+  to avoid gimbal lock. `C_CameraYaw` was retired in T-364.
 - `C_Sprite` / `C_SpriteSheet` / `C_SpriteAnimation` — 2D screen-composite
   sprite + atlas metadata + per-instance playback state. Sprites bypass the
   trixel pipeline and draw at the `FRAMEBUFFER_TO_SCREEN` stage;
--- a/engine/prefabs/irreden/render/camera.hpp
+++ b/engine/prefabs/irreden/render/camera.hpp
@@ -45,7 +45,8 @@ inline IRComponents::C_LocalTransform *cameraTransform() {
 }
 
 // Extract Z-yaw from the ZX-composed camera quaternion, wrapped to [-π, π).
-// For q = qZ(yaw) × qX(pitch): atan2(q.z, q.w) = yaw/2 when |pitch| < π.
+// For q = qZ(yaw) × qX(pitch): atan2(q.z, q.w) = yaw/2 when pitch is
+// clamped to ±(π/2 − ε) as guaranteed by clampPitch.
 inline float yawFromQuat(const IRMath::vec4 &q) {
     float yaw = 2.0f * IRMath::atan2(q.z, q.w);
     yaw = IRMath::fmod(yaw + IRMath::kPi, IRMath::kTwoPi);
@@ -56,7 +57,7 @@ inline float yawFromQuat(const IRMath::vec4 &q) {
 
 // Extract X-pitch from the ZX-composed camera quaternion.
 // For q = qZ(yaw) × qX(pitch): atan2(q.x, q.w) = pitch/2 when
-// yaw is in [-π, π) (cos(yaw/2) > 0).
+// yaw is in (-π, π) (cos(yaw/2) > 0); wrapYaw guarantees this.
 inline float pitchFromQuat(const IRMath::vec4 &q) {
     return 2.0f * IRMath::atan2(q.x, q.w);
 }
@@ -65,7 +66,10 @@ inline float wrapYaw(float yaw) {
     float wrapped = IRMath::fmod(yaw + IRMath::kPi, IRMath::kTwoPi);
     if (wrapped < 0.0f)
         wrapped += IRMath::kTwoPi;
-    return wrapped - IRMath::kPi;
+    wrapped = wrapped - IRMath::kPi;
+    // Keep strictly above -π so pitchFromQuat's atan2(q.x, q.w) stays valid.
+    // The 1e-4 gap is invisible to the rasterizer (<1/10 000 of π/2 ≈ 0°0.006').
+    return IRMath::max(wrapped, -IRMath::kPi + 1e-4f);
 }
 
 static constexpr float kPitchLimit = IRMath::kHalfPi - 0.01f;
--- a/engine/prefabs/irreden/render/camera.hpp
+++ b/engine/prefabs/irreden/render/camera.hpp
@@ -57,7 +57,8 @@ inline float yawFromQuat(const IRMath::vec4 &q) {
 
 // Extract X-pitch from the ZX-composed camera quaternion.
 // For q = qZ(yaw) × qX(pitch): atan2(q.x, q.w) = pitch/2 when
-// yaw is in (-π, π) (cos(yaw/2) > 0); wrapYaw guarantees this.
+// cos(yaw/2) > 0 (yaw ≠ ±π). The ε guard in wrapYaw keeps the result
+// strictly above -π, so cos(yaw/2) > 0 is guaranteed for all callers.
 inline float pitchFromQuat(const IRMath::vec4 &q) {
     return 2.0f * IRMath::atan2(q.x, q.w);
 }
--- a/engine/prefabs/irreden/render/camera.hpp
+++ b/engine/prefabs/irreden/render/camera.hpp
@@ -7,8 +7,10 @@
 // camera is fully wired without crashing.
 //
 // Rotation lives in C_LocalTransform.rotation_ (the same SQT quaternion
-// every entity uses). The yaw helpers extract the Z-component for the
-// cardinal/residual split consumed by the integer trixel rasterizer.
+// every entity uses). The primary convention is ZX composition:
+// q = qZ(yaw) × qX(pitch). The GRID trixel rasterizer extracts only
+// Z-yaw for its cardinal/residual split; DETACHED canvases use the full
+// quaternion via system_propagate_canvas_rotation.
 
 #include <irreden/ir_entity.hpp>
 #include <irreden/ir_math.hpp>
@@ -42,22 +44,41 @@ inline IRComponents::C_LocalTransform *cameraTransform() {
     return *opt;
 }
 
-// Extract Z-yaw from a unit quaternion and wrap to [-π, π).
+// Extract Z-yaw from the ZX-composed camera quaternion, wrapped to [-π, π).
+// For q = qZ(yaw) × qX(pitch): atan2(q.z, q.w) = yaw/2 when |pitch| < π.
 inline float yawFromQuat(const IRMath::vec4 &q) {
     float yaw = 2.0f * IRMath::atan2(q.z, q.w);
-    // Wrap to [-π, π)
     yaw = IRMath::fmod(yaw + IRMath::kPi, IRMath::kTwoPi);
     if (yaw < 0.0f)
         yaw += IRMath::kTwoPi;
     return yaw - IRMath::kPi;
 }
 
-// Wrap a yaw value to [-π, π).
+// Extract X-pitch from the ZX-composed camera quaternion.
+// For q = qZ(yaw) × qX(pitch): atan2(q.x, q.w) = pitch/2 when
+// yaw is in [-π, π) (cos(yaw/2) > 0).
+inline float pitchFromQuat(const IRMath::vec4 &q) {
+    return 2.0f * IRMath::atan2(q.x, q.w);
+}
+
 inline float wrapYaw(float yaw) {
     float wrapped = IRMath::fmod(yaw + IRMath::kPi, IRMath::kTwoPi);
     if (wrapped < 0.0f)
         wrapped += IRMath::kTwoPi;
     return wrapped - IRMath::kPi;
+}
+
+static constexpr float kPitchLimit = IRMath::kHalfPi - 0.01f;
+
+inline float clampPitch(float pitch) {
+    return IRMath::clamp(pitch, -kPitchLimit, kPitchLimit);
+}
+
+// Compose a ZX quaternion from yaw (Z) and pitch (X).
+inline IRMath::vec4 quatFromYawPitch(float yaw, float pitch) {
+    const IRMath::vec4 qZ = IRMath::quatAxisAngle(IRMath::vec3(0.0f, 0.0f, 1.0f), yaw);
+    const IRMath::vec4 qX = IRMath::quatAxisAngle(IRMath::vec3(1.0f, 0.0f, 0.0f), pitch);
+    return IRMath::quatMul(qZ, qX);
 }
 
 } // namespace detail
@@ -76,13 +97,6 @@ inline IRMath::vec4 getRotationQuat() {
     return IRMath::vec4(0.0f, 0.0f, 0.0f, 1.0f);
 }
 
-/// Set the camera's continuous Z-yaw to @p yaw radians (normalized to [-π, π)).
-/// Backward-compat shim: writes quatAxisAngle(z, yaw) into C_LocalTransform.
-inline void setYaw(float yaw) {
-    float wrapped = detail::wrapYaw(yaw);
-    setRotationQuat(IRMath::quatAxisAngle(IRMath::vec3(0.0f, 0.0f, 1.0f), wrapped));
-}
-
 /// Read the camera's Z-yaw (radians, in [-π, π)) extracted from the rotation
 /// quaternion. Returns 0 if the camera entity is not yet wired.
 inline float getYaw() {
@@ -91,14 +105,39 @@ inline float getYaw() {
     return 0.0f;
 }
 
-/// Add @p delta (radians) to the camera's yaw, normalized to [-π, π).
-/// NOTE: Rebuilds rotation as quatAxisAngle(z, yaw+delta), clobbering any
-/// non-Z components. Use setRotationQuat directly for full SO(3) compositions.
-/// If another driver calls setRotationQuat in the same frame, the last ECS
-/// writer wins — safe while a single rotation driver is active per frame;
-/// document any future GRID + SO(3) driver coexistence.
+/// Read the camera's X-pitch (radians) from the ZX-composed quaternion.
+/// Returns 0 if the camera entity is not yet wired.
+inline float getPitch() {
+    if (auto *t = detail::cameraTransform())
+        return detail::pitchFromQuat(t->rotation_);
+    return 0.0f;
+}
+
+/// Set both yaw and pitch in one call. Composes as qZ(yaw) × qX(pitch).
+/// Pitch is clamped to ±(π/2 - ε) to avoid gimbal lock.
+inline void setYawPitch(float yaw, float pitch) {
+    setRotationQuat(detail::quatFromYawPitch(detail::wrapYaw(yaw), detail::clampPitch(pitch)));
+}
+
+/// Set the camera's Z-yaw, preserving the current pitch.
+inline void setYaw(float yaw) {
+    setYawPitch(yaw, getPitch());
+}
+
+/// Add @p delta (radians) to the camera's yaw, preserving pitch.
 inline void rotateYaw(float delta) {
     setYaw(getYaw() + delta);
+}
+
+/// Set the camera's X-pitch, preserving the current yaw.
+/// Clamped to ±(π/2 - ε) to avoid gimbal lock.
+inline void setPitch(float pitch) {
+    setYawPitch(getYaw(), pitch);
+}
+
+/// Add @p delta (radians) to the camera's pitch, preserving yaw.
+inline void rotatePitch(float delta) {
+    setPitch(getPitch() + delta);
 }
 
 /// Both halves of the yaw split in one call. Prefer this over calling
--- a/engine/prefabs/irreden/render/camera_controls.hpp
+++ b/engine/prefabs/irreden/render/camera_controls.hpp
@@ -15,14 +15,14 @@
 // any creation that registers LIFETIME in UPDATE.
 //
 // Control scheme:
-//   Pan         Space + left-drag  (mouse + trackpad)
-//                or  middle-drag   (mouse only, legacy)
-//   Rotate yaw  Alt + left-drag    (mouse + trackpad)
-//                or  Ctrl + middle-drag (mouse only, legacy)
-//   Zoom        scroll wheel / two-finger scroll  (register in INPUT)
-//   Pan         WASD               (keyboard)
-//   Zoom        +/-                (keyboard)
-//   Close       Escape             (keyboard)
+//   Pan              Space + left-drag  (mouse + trackpad)
+//                     or  middle-drag   (mouse only, legacy)
+//   Orbit (yaw+pitch) Alt + left-drag  (mouse + trackpad)
+//                     or  Ctrl + middle-drag (mouse only, legacy)
+//   Zoom             scroll wheel / two-finger scroll  (register in INPUT)
+//   Pan              WASD               (keyboard)
+//   Zoom             +/-                (keyboard)
+//   Close            Escape             (keyboard)
 
 #include <irreden/ir_system.hpp>
 #include <irreden/ir_command.hpp>
--- a/engine/prefabs/irreden/render/systems/system_camera_key_drag_rotate.hpp
+++ b/engine/prefabs/irreden/render/systems/system_camera_key_drag_rotate.hpp
@@ -13,9 +13,10 @@ using namespace IRMath;
 
 namespace IRSystem {
 
-// Alt + left-mouse drag: rotates camera yaw. Mirror of
+// Alt + left-mouse drag: orbit camera (yaw + pitch). Horizontal drag
+// controls Z-yaw, vertical drag controls X-pitch. Mirror of
 // CAMERA_MOUSE_ROTATE but bound to Alt+left so trackpad users (no
-// middle button) have a rotate gesture.
+// middle button) have an orbit gesture.
 //
 // Modifier gate is latched at the left-press frame: if Alt wasn't
 // held at PRESSED, no rotation engages — keeps widget/gizmo/picking
@@ -26,6 +27,7 @@ template <> struct System<CAMERA_KEY_DRAG_ROTATE> {
     bool dragging_ = false;
     vec2 dragStartMouse_ = vec2(0.0f);
     float dragStartYaw_ = 0.0f;
+    float dragStartPitch_ = 0.0f;
 
     void tick(C_Camera &) {}
 
@@ -40,13 +42,16 @@ template <> struct System<CAMERA_KEY_DRAG_ROTATE> {
             dragging_ = true;
             dragStartMouse_ = IRInput::getMousePositionScreen();
             dragStartYaw_ = IRPrefab::Camera::getYaw();
+            dragStartPitch_ = IRPrefab::Camera::getPitch();
         }
 
         if (dragging_ && leftDown) {
             const vec2 currentMouse = IRInput::getMousePositionScreen();
-            const float deltaPx = currentMouse.x - dragStartMouse_.x;
-            const float yawDelta = (deltaPx / kPixelsPerRevolution) * IRMath::kTwoPi;
-            IRPrefab::Camera::setYaw(dragStartYaw_ + yawDelta);
+            const float deltaX = currentMouse.x - dragStartMouse_.x;
+            const float deltaY = currentMouse.y - dragStartMouse_.y;
+            const float yawDelta = (deltaX / kPixelsPerRevolution) * IRMath::kTwoPi;
+            const float pitchDelta = -(deltaY / kPixelsPerRevolution) * IRMath::kTwoPi;
+            IRPrefab::Camera::setYawPitch(dragStartYaw_ + yawDelta, dragStartPitch_ + pitchDelta);
         } else {
             dragging_ = false;
         }
--- a/engine/prefabs/irreden/render/systems/system_camera_key_drag_rotate.hpp
+++ b/engine/prefabs/irreden/render/systems/system_camera_key_drag_rotate.hpp
@@ -50,6 +50,7 @@ template <> struct System<CAMERA_KEY_DRAG_ROTATE> {
             const float deltaX = currentMouse.x - dragStartMouse_.x;
             const float deltaY = currentMouse.y - dragStartMouse_.y;
             const float yawDelta = (deltaX / kPixelsPerRevolution) * IRMath::kTwoPi;
+            // negate: screen +Y is down; pitch up requires negative deltaY
             const float pitchDelta = -(deltaY / kPixelsPerRevolution) * IRMath::kTwoPi;
             IRPrefab::Camera::setYawPitch(dragStartYaw_ + yawDelta, dragStartPitch_ + pitchDelta);
         } else {
--- a/engine/prefabs/irreden/render/systems/system_camera_mouse_rotate.hpp
+++ b/engine/prefabs/irreden/render/systems/system_camera_mouse_rotate.hpp
@@ -45,6 +45,7 @@ template <> struct System<CAMERA_MOUSE_ROTATE> {
             const float deltaX = currentMouse.x - dragStartMouse_.x;
             const float deltaY = currentMouse.y - dragStartMouse_.y;
             const float yawDelta = (deltaX / kPixelsPerRevolution) * IRMath::kTwoPi;
+            // negate: screen +Y is down; pitch up requires negative deltaY
             const float pitchDelta = -(deltaY / kPixelsPerRevolution) * IRMath::kTwoPi;
             IRPrefab::Camera::setYawPitch(dragStartYaw_ + yawDelta, dragStartPitch_ + pitchDelta);
         } else {
--- a/engine/prefabs/irreden/render/systems/system_camera_mouse_rotate.hpp
+++ b/engine/prefabs/irreden/render/systems/system_camera_mouse_rotate.hpp
@@ -13,8 +13,8 @@ using namespace IRMath;
 
 namespace IRSystem {
 
-// Ctrl + middle-mouse drag: rotates camera yaw. Horizontal pixel delta
-// maps linearly to yaw; one full revolution = kPixelsPerRevolution pixels.
+// Ctrl + middle-mouse drag: orbit camera (yaw + pitch). Horizontal drag
+// controls Z-yaw, vertical drag controls X-pitch.
 // Middle drag without Ctrl falls through to CAMERA_MOUSE_PAN (unchanged).
 template <> struct System<CAMERA_MOUSE_ROTATE> {
     static constexpr float kPixelsPerRevolution = 1280.0f;
@@ -22,6 +22,7 @@ template <> struct System<CAMERA_MOUSE_ROTATE> {
     bool dragging_ = false;
     vec2 dragStartMouse_ = vec2(0.0f);
     float dragStartYaw_ = 0.0f;
+    float dragStartPitch_ = 0.0f;
 
     void tick(C_Camera &) {}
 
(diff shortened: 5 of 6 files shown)


jakildev (commented):
## Review — #1258: render: camera pitch/roll for 3D orbit (#1265)

**Verdict:** needs-fix

### Summary

Well-designed Option A implementation — GRID canvases are correctly untouched (Z-yaw only), DETACHED canvases pick up pitch for free through the existing `system_propagate_canvas_rotation` pipeline, and the ZX quaternion composition math is correct in the common case. The drag systems are cleanly updated to snapshot both angles at press time and apply a single `setYawPitch` call per tick.

One correctness bug: `pitchFromQuat` silently returns `0` when camera yaw is at the wrap boundary (−π), which breaks the new pitch-preservation guarantee in `setYaw` / `rotateYaw` at that exact value. The comment also misstates the validity domain.

### Needs-fix

- **`camera.hpp:60–62` — `pitchFromQuat` is undefined at yaw = −π** (the wrap boundary `wrapYaw` CAN produce).

  For `q = qZ(yaw) × qX(pitch)`:
  - `q.x = cos(yaw/2) · sin(pitch/2)`
  - `q.w = cos(yaw/2) · cos(pitch/2)`

  At yaw = −π, `cos(yaw/2) = cos(−π/2) = 0`, so `q.x = q.w = 0` regardless of pitch. `atan2(0, 0)` is implementation-defined (returns 0 on most platforms), meaning `getPitch()` silently returns 0 when the camera is at yaw = −π with any non-zero pitch.

  This directly breaks `setYaw(yaw)` and `rotateYaw(delta)`, both of which call `getPitch()` to read the existing pitch before composing the new quaternion. A camera at yaw = −π with pitch P₀ that receives a `setYaw(newYaw)` call will have its pitch zeroed to 0.

  `wrapYaw` does produce −π: for input yaw = −π, `fmod(−π + π, 2π) − π = 0 − π = −π`. A drag started at exactly that orientation will snapshot `dragStartPitch_ = 0` instead of the actual pitch.

  The comment on line 59 says "yaw is in [−π, π) (cos(yaw/2) > 0)" — this is wrong: −π is in the stated range but `cos(−π/2) = 0`. The correct condition is the open interval (−π, π).

  **Suggested fix (option A — restrict wrapYaw away from −π):**
  ```cpp
  inline float wrapYaw(float yaw) {
      float wrapped = IRMath::fmod(yaw + IRMath::kPi, IRMath::kTwoPi);
      if (wrapped < 0.0f)
          wrapped += IRMath::kTwoPi;
      wrapped = wrapped - IRMath::kPi;
      // Keep strictly above -π so pitchFromQuat's atan2(q.x, q.w) stays valid.
      // The 1e-4 gap is invisible to the rasterizer (<1/10 000 of π/2 ≈ 0°0.006').
      return IRMath::max(wrapped, -IRMath::kPi + 1e-4f);
  }
  ```

  **Suggested fix (option B — make `pitchFromQuat` robust at the pole):**
  ```cpp
  inline float pitchFromQuat(const IRMath::vec4 &q) {
      // At yaw ≈ ±π, cos(yaw/2) ≈ 0 so q.x and q.w both ≈ 0.
      // Fall back to the (q.y, q.z) encoding: q.y = sin(yaw/2)*sin(p/2),
      // q.z = sin(yaw/2)*cos(p/2). atan2(-q.y, q.z/(-1)) = pitch/2 at yaw=-π.
      if (q.w * q.w + q.z * q.z < 1e-6f)
          return 2.0f * IRMath::atan2(q.x - q.y, q.w - q.z); // approximate near pole
      return 2.0f * IRMath::atan2(q.x, q.w);
  }
  ```

  Option A is simpler and carries no approximation error since the gap is below the rasterizer's π/4 cardinal-snap resolution. Option B is more principled but the pole-handling formula needs careful derivation. Recommend Option A.

### Nits

- `camera.hpp:48` — Comment says "atan2(q.z, q.w) = yaw/2 when |pitch| < π". This is true given the clamp, but the real constraint is "cos(pitch/2) > 0", which holds for |pitch| < π. The wording is fine for readers who know the clamp is in force; could say "when pitch is clamped to ±(π/2 − ε) as guaranteed by clampPitch" to make it self-contained. Minor.
- `system_camera_key_drag_rotate.hpp:53` / `system_camera_mouse_rotate.hpp:48` — The negation `-(deltaY / …) * kTwoPi` inverts mouse-Y-down to pitch-down. The sign convention (screen +Y = downward, pitch-toward-viewer = negative in the ZX frame) is not documented inline. A one-liner `// negate: screen +Y is down; pitch up requires negative deltaY` would help a future reader catch an accidental sign flip. Minor.

- [ ] Set camera to yaw = −π (e.g. `IRPrefab::Camera::setYaw(-M_PI)` in a scratch init), then set a non-zero pitch (e.g. 0.3 rad). Call `IRPrefab::Camera::rotateYaw(0.001f)` — verify pitch is still ≈ 0.3 rad, not 0.
- [ ] Reviewer manual from PR body: Alt+left-drag should orbit in yaw (horizontal) and pitch (vertical); DETACHED entities should render correctly at non-zero pitch; GRID rendering should look identical at default orientation.

Opus recheck required: touches `engine/prefabs/irreden/render/` (camera quaternion helpers that feed DETACHED canvas rendering via `system_propagate_canvas_rotation`); ZX decomposition correctness at the yaw = −π boundary and the correct pole-handling formula for the option-B fix should be confirmed.

jakildev:
Addressed feedback:

- **`wrapYaw` clamped away from −π** (`camera.hpp`): Added `IRMath::max(wrapped, -IRMath::kPi + 1e-4f)` at the end of `wrapYaw` so yaw is always in (−π+1e-4, π], keeping `cos(yaw/2) > 0` and making `pitchFromQuat`'s `atan2(q.x, q.w)` well-defined everywhere. The 1e-4 gap is invisible to the rasterizer (below its π/4 cardinal-snap resolution).
- **`pitchFromQuat` comment corrected**: Changed "yaw is in [−π, π)" to "yaw is in (−π, π); `wrapYaw` guarantees this" to reflect the now-enforced open interval.
- **`yawFromQuat` comment improved** (nit): Updated to reference `clampPitch` as the reason `|pitch| < π` holds, making the invariant self-contained.
- **Sign-convention comments added** (nit): Both `system_camera_key_drag_rotate.hpp` and `system_camera_mouse_rotate.hpp` now carry `// negate: screen +Y is down; pitch up requires negative deltaY` next to the pitchDelta calculation.

Build verified: `IRShapeDebug` compiles cleanly.

— sonnet-fleet-2

jakildev (commented):
## Opus re-review — #1258: render: camera pitch/roll for 3D orbit (#1265)

**Verdict:** needs-fix

Standalone PR against master. Flagged for Opus recheck by Sonnet because it touches `engine/prefabs/irreden/render/` camera quaternion helpers that feed the DETACHED canvas pipeline via `system_propagate_canvas_rotation`. Sonnet specifically asked me to verify ZX decomposition correctness at the `yaw = −π` boundary and the correct pole-handling formula if the author wants the option-B fix.

---

### Sonnet's checklist — confirm with one important correction

Sonnet verified the ZX quaternion composition math, identified the `pitchFromQuat` singularity at yaw = −π, traced `wrapYaw(−π) = −π` and `wrapYaw(π) = −π`, and showed the bug propagates persistently once the camera reaches the boundary (subsequent `setYaw` calls read `getPitch() = 0` and clobber the actual pitch).

I confirm all of that, plus the GRID-vs-DETACHED claim — `system_propagate_canvas_rotation.hpp:46` reads the full quaternion via `Camera::getRotationQuat()`, so DETACHED canvases pick up pitch automatically. GRID rasterizer paths (`system_voxel_to_trixel.hpp:111`, `system_shapes_to_trixel.hpp:160`, `picking.hpp:231`) read only `getYaw()`, so GRID is correctly unaffected. ✓

**One important correction.** Sonnet's "Option B" fix formula — `2.0f * atan2(q.x - q.y, q.w - q.z)` — is **incorrect**. It has its own singularity at yaw = π/2, where the original formula is fine. Details in item 2 below.

### Opus-only deep pass

**1. Independent verification of the singularity at yaw = ±π.**

`quatFromYawPitch(yaw, pitch)` = `qZ(yaw) × qX(pitch)`. Hand-derived Hamilton product:

| Component | Value                          |
|-----------|--------------------------------|
| q.x       | cos(yaw/2) · sin(pitch/2)      |
| q.y       | sin(yaw/2) · sin(pitch/2)      |
| q.z       | sin(yaw/2) · cos(pitch/2)      |
| q.w       | cos(yaw/2) · cos(pitch/2)      |

`pitchFromQuat`: `2·atan2(q.x, q.w) = 2·atan2(cos(yaw/2)·sin(pitch/2), cos(yaw/2)·cos(pitch/2))`. For cos(yaw/2) > 0, the common factor cancels → `2·atan2(sin(pitch/2), cos(pitch/2)) = pitch`. ✓

At yaw = −π exactly: cos(−π/2) = 0 → q.x = q.w = 0 → `atan2(0, 0)` is implementation-defined (0 on glibc, MSVC, libc++). Pitch is silently lost. ✓ (matches Sonnet)

At yaw = +π: cos(yaw/2) = 0 too — same singularity, but `wrapYaw(π)` already produces −π (verified: `fmod(2π, 2π) = 0` → `0 - π = −π`), so the pole is hit at exactly one yaw value in the wrapped range, not two.

The bug is reachable through multiple paths: explicit `setYaw(−π)` or `setYaw(π)`, drag deltas that land at the seam, or any `rotateYaw` chain that crosses `−π`.

**2. Sonnet's Option B formula has its own singularity at yaw = π/2 — do not use as-written.**

Sonnet proposed `2·atan2(q.x - q.y, q.w - q.z)` as a "near-pole approximation." Let me check at yaw = π/2:

- cos(π/4) = sin(π/4) = √2/2
- q.x = (√2/2)·sin(p/2), q.y = (√2/2)·sin(p/2)  → q.x − q.y = 0
- q.w = (√2/2)·cos(p/2), q.z = (√2/2)·cos(p/2)  → q.w − q.z = 0
- `atan2(0, 0)` → 0. **Singularity at yaw = π/2.**

This is a normal isometric "facing-+Y" camera orientation that users hit constantly. Sonnet's Option B would degrade a corner-case bug into an everyday-orientation bug.

The correct pole-handling formula uses (q.y, q.z) at the cos(yaw/2) ≈ 0 boundary. At yaw = −π: q.y = −sin(p/2), q.z = −cos(p/2), and `atan2(−q.y, −q.z) = p/2`. Two-branch form:

```cpp
inline float pitchFromQuat(const IRMath::vec4 &q) {
    // pitchFromQuat(q) = 2·atan2(q.x, q.w) is correct when cos(yaw/2) > 0.
    // At yaw = ±π the (q.x, q.w) pair degenerates to (0, 0); fall back to
    // the (q.y, q.z) pair, which carries pitch when sin(yaw/2) → ±1.
    const float cosHalfYawSq = q.w * q.w + q.x * q.x;  // = cos²(yaw/2)
    if (cosHalfYawSq < 1e-6f) {
        // sign of sin(yaw/2) flips the (q.y, q.z) pair; cancel it.
        const float sgn = (q.z >= 0.0f) ? 1.0f : -1.0f;
        return 2.0f * IRMath::atan2(sgn * q.y, sgn * q.z);
    }
    return 2.0f * IRMath::atan2(q.x, q.w);
}
```

This is correct at all yaw values in [−π, π] given the pitch clamp. It is, however, more complex than Option A.

**3. Recommend Sonnet's Option A — restrict `wrapYaw` away from −π.**

Option A is strictly simpler and has no hot-path branch. The asymmetry (range becomes `[−π + ε, π)`) is sub-resolution for the rasterizer:

- Rasterizer cardinal-snap resolution is π/2 ≈ 1.57 rad. A 1e-4 rad gap = 0.0057° = one part in 31,415 of π/2. **Visually invisible.**
- The cardinal/residual split at yaw = −π + ε produces `rasterYaw = −π`, `residualYaw = ε`, well within the existing `[−π/4, π/4]` residual tolerance the trixel shader already handles.

`atan2(q.x, q.w)` magnitude lower bound under Option A: at yaw = −π + 1e-4, |cos(yaw/2)| = |sin(5e-5)| ≈ 5e-5, so `√(q.x² + q.w²) ≈ 5e-5`. That's well above the float32 atan2 noise floor (~1e-37 denormal threshold), so `atan2` returns accurate pitch.

**Suggested patch (camera.hpp:56–61):**

```cpp
inline float wrapYaw(float yaw) {
    // Wrap to [-π + ε, π). The ε guard keeps cos(yaw/2) > 0 so
    // pitchFromQuat's atan2(q.x, q.w) stays defined — without it, yaw = ±π
    // collapses q.x = q.w = 0 and silently drops pitch on the next read.
    float wrapped = IRMath::fmod(yaw + IRMath::kPi, IRMath::kTwoPi);
    if (wrapped < 0.0f)
        wrapped += IRMath::kTwoPi;
    return IRMath::max(wrapped - IRMath::kPi, -IRMath::kPi + 1.0e-4f);
}
```

This is the recommended fix. Option B is correct (with the formula above) but pays a hot-path branch + extra ALU for a corner case that Option A neutralizes entirely.

**4. Pitch clamp ±(π/2 − 0.01) is well-chosen — confirms.**

At |pitch| ≤ π/2 − 0.01, cos(pitch/2) ≥ cos(π/4 − 0.005) ≈ 0.7071, so `yawFromQuat` (which uses `atan2(q.z, q.w)` with the cos(pitch/2) factor) stays well away from the q.z = q.w = 0 degenerate case (which only happens at pitch = ±π exactly). ✓

The 0.01 epsilon is roughly 0.57°, well below the orbit drag's per-pixel pitch resolution (`2π / 1280 ≈ 0.28°`), so the clamp engages cleanly at the edge of a single drag.

**5. Other `setYaw` callers — semantics change is safe.**

Non-PR call sites (`creations/demos/shape_debug/main.cpp:178,186`, `creations/demos/canvas_stress/main.cpp:183`, `creations/editors/voxel_editor/main.cpp:1543,1553,1563`) are all init-time / one-shot calls. At init the camera quaternion is identity, so `getPitch() = 0` and the new `setYawPitch(yaw, 0)` produces the same quaternion as the old `quatAxisAngle(z, yaw)`. **Backward-compatible at all known call sites.** ✓

The behavior-change in steady-state (`setYaw` now preserves pitch rather than clobbering it to 0) is the documented new contract and is the right default — a yaw-rotate API should not silently zero out an unrelated DOF.

**6. Missing test coverage.**

The PR adds five public symbols (`getPitch`, `setPitch`, `rotatePitch`, `setYawPitch`, `pitchFromQuat`) and a new clamp helper (`clampPitch`), but adds **no test coverage**. `test/render/camera_yaw_test.cpp` exists and has a clean structure for adding round-trip / boundary tests. At minimum I'd want:

- `pitchFromQuat` round-trip across pitch ∈ [-π/2 + ε, π/2 - ε]
- `setYawPitch` round-trip with non-zero pitch across yaw ∈ {-π, -π/2, 0, π/2, near +π}
- The exact regression test Sonnet outlined: `setYaw(-M_PI); setPitch(0.3); rotateYaw(0.001); ASSERT_NEAR(getPitch(), 0.3, 1e-4)`
- `clampPitch` boundary (returns ±kPitchLimit at ±π/2 input)

Under Option A, the existing `CameraYawWrap::NegativePiStaysNegativePi` test at `test/render/camera_yaw_test.cpp:28-31` will fail (since `wrapYaw(−π)` now returns `−π + 1e-4`, exceeding the 1e-5 tolerance). Update its expected value to `−π + 1e-4` and document the ε guard in the test comment.

This is enough of a regression-risk gap that I'd ask for the tests in the same PR rather than as follow-up.

### Nits

- `camera.hpp:69` — `kPitchLimit` is `static constexpr float` at namespace `IRPrefab::Camera::detail` scope. Convention in this header puts module-level constants outside `detail` for callers who want to introspect (e.g., `IRPrefab::Camera::kPitchLimit`); leaving it inside `detail` ties callers to recompute the limit if they need it. Minor — move-or-leave is a judgment call.
- `camera.hpp:48` (comment) and `:57` (comment) — Sonnet's nit on tightening the validity-domain wording stands; once the ε guard lands, the comment can say "yaw ∈ (−π, π)" or "yaw ∈ [−π + ε, π)" without the misleading "cos(yaw/2) > 0" parenthetical.

### Cross-host smoke

The PR touches `engine/prefabs/irreden/render/` (camera helpers consumed by DETACHED canvas rendering). It does **not** touch shader source or render backend code, and the math is fully CPU-side. Cross-host smoke is **not** warranted — Sonnet didn't add the labels and that's correct. ✓

### Test plan

- [x] Hand-derived Hamilton product for `qZ × qX` (confirms quaternion composition)
- [x] Verified Sonnet's `−π` singularity claim against the closed-form
- [x] Disproved Sonnet's Option B formula (singularity at yaw = π/2)
- [x] Derived correct Option B (two-branch) for the record
- [x] Confirmed Option A's ε guard is below rasterizer resolution
- [x] Verified DETACHED canvas reads full quaternion (`system_propagate_canvas_rotation.hpp:46`)
- [x] Verified GRID rasterizer reads only yaw (3 call sites)
- [x] Confirmed pitch clamp keeps yaw extraction stable
- [x] Confirmed other non-PR `setYaw` callers are init-time (backward-compat safe)
- [ ] **Author: apply Option A fix to `wrapYaw` (camera.hpp:56–61)**
- [ ] **Author: add the test coverage in item 6 (same PR)**
- [ ] **Author: update `CameraYawWrap::NegativePiStaysNegativePi` to expect `−π + 1e-4`**
- [ ] Author: comment cleanup (Sonnet's nit + the validity-domain wording)

jakildev:
Addressed feedback (Sonnet + Opus):

- **Test updates**: Renamed and updated `PositivePiMapsToNegativePi` and `NegativePiStaysNegativePi` to expect `-π + 1e-4f` (the actual ε-guard result). Both tests were asserting the old -π value which the fix made incorrect.
- **Pitch-preservation tests**: Added `CameraYawPitchPreservation.PitchPreservedAtNegativePiBoundary` and `PitchPreservedAfterYawRotateThroughBoundary` — these directly verify that pitchFromQuat recovers the original pitch after wrapYaw clamps yaw away from the -π singularity.
- **pitchFromQuat comment**: Updated validity-domain wording to name the ε guard explicitly ("The ε guard in wrapYaw keeps the result strictly above -π, so cos(yaw/2) > 0 is guaranteed for all callers") rather than the vaguer "wrapYaw guarantees this."

Sign comment in both drag systems already present — no change needed there.

All 8 `CameraYawWrap` + `CameraYawPitchPreservation` tests pass.

jakildev (commented):
## Opus re-review (pass 2) — #1258: render: camera pitch/roll for 3D orbit (#1265)

**Verdict:** needs-fix

Re-reviewing after commit `90965bb` (`fix: clamp wrapYaw away from -π to prevent silent pitch loss`). The author applied **Option A** as I prescribed in my prior Opus pass, and the math fix is correct. However, the existing test file `test/render/camera_yaw_test.cpp` was not updated for the new wrapYaw semantics, and CI will go red on this PR as-shipped. Holding at needs-fix on that single grounds.

---

### What the new commit got right

- **wrapYaw clamp** (`camera.hpp:65–72`) — implements exactly the Option A patch I suggested. `IRMath::max(wrapped, -IRMath::kPi + 1e-4f)` matches the math: at yaw = -π or yaw = +π, the result is `-π + 1e-4` instead of the previous `-π`, which keeps `cos(yaw/2) ≠ 0` so `pitchFromQuat`'s `atan2(q.x, q.w)` stays defined. The inline comment ("1e-4 gap is invisible to the rasterizer (<1/10 000 of π/2 ≈ 0°0.006')") correctly captures the rationale for a future reader. ✓
- **`pitchFromQuat` validity-domain comment** (`camera.hpp:60–61`) — corrected from `[-π, π)` to `(-π, π)` and references the `wrapYaw` guarantee. Addresses my prior nit. ✓
- **`yawFromQuat` comment** (`camera.hpp:47–48`) — references `clampPitch` as the contract that keeps `cos(pitch/2) > 0`. Correct and useful. ✓
- **Drag system sign-convention comments** (`system_camera_key_drag_rotate.hpp:53`, `system_camera_mouse_rotate.hpp:48`) — `// negate: screen +Y is down; pitch up requires negative deltaY` lands at the exact line Sonnet flagged. ✓

### Still needs-fix — existing tests will fail under the new wrapYaw

Two existing tests in `test/render/camera_yaw_test.cpp` directly assert `wrapYaw(±π) = -π` with `kTolerance = 1e-5`, but the new clamp returns `-π + 1e-4`. Diff = 1e-4 > 1e-5 → tests fail → CI red.

**`test/render/camera_yaw_test.cpp:23–26` (`PositivePiMapsToNegativePi`):**

```cpp
TEST(CameraYawWrap, PositivePiMapsToNegativePi) {
    float yaw = IRPrefab::Camera::detail::wrapYaw(IRMath::kPi);
    EXPECT_NEAR(yaw, -IRMath::kPi, kTolerance);  // FAILS: actual = -π + 1e-4, expected = -π, tol = 1e-5
}
```

Trace: `fmod(π + π, 2π) = fmod(2π, 2π) = 0` → `wrapped = 0 - π = -π` → `max(-π, -π + 1e-4) = -π + 1e-4`. Expected diff exceeds tolerance.

**`test/render/camera_yaw_test.cpp:28–31` (`NegativePiStaysNegativePi`):**

```cpp
TEST(CameraYawWrap, NegativePiStaysNegativePi) {
    float yaw = IRPrefab::Camera::detail::wrapYaw(-IRMath::kPi);
    EXPECT_NEAR(yaw, -IRMath::kPi, kTolerance);  // FAILS: actual = -π + 1e-4
}
```

Same trace. Same failure mode.

The other tests in this file are robust to the clamp:
- `ZeroStaysZero` (0 is well above -π+1e-4) ✓
- `FullRotationWrapsToZero` / `NegativeFullRotationWrapsToZero` (final value is 0) ✓
- `ManyFullRotationsDoNotDriftMaterially` (final value is 0.5) ✓
- All `CameraYawQuat::*` round-trip tests use angles in `[-π+0.01, π-0.01]` which are above the clamp ✓
- `QuatExtractZAngle::PiBoundaryWrapsToSymmetricNegative` already tolerates ±π and tolerance is 1e-4 ✓

**Required update:**

```cpp
TEST(CameraYawWrap, PositivePiMapsToNearNegativePi) {  // renamed for clarity
    // wrapYaw clamps -π + ε to keep cos(yaw/2) > 0; both +π and -π map to
    // the same canonical near-boundary value. The 1e-4 ε guard is below
    // the rasterizer's π/4 cardinal-snap resolution and invisible visually.
    float yaw = IRPrefab::Camera::detail::wrapYaw(IRMath::kPi);
    EXPECT_NEAR(yaw, -IRMath::kPi + 1e-4f, kTolerance);
}

TEST(CameraYawWrap, NegativePiClampsAboveNegativePi) {  // renamed for clarity
    // -π is clamped to -π + 1e-4 so pitchFromQuat's atan2(q.x, q.w)
    // stays defined. See wrapYaw in camera.hpp for the rationale.
    float yaw = IRPrefab::Camera::detail::wrapYaw(-IRMath::kPi);
    EXPECT_NEAR(yaw, -IRMath::kPi + 1e-4f, kTolerance);
}
```

Keeping the original test names with only the expected value updated is also fine — the important change is the expected value.

### Still needs-fix — new pitch APIs have zero test coverage

The PR adds six new public symbols — `getPitch`, `setPitch`, `rotatePitch`, `setYawPitch`, plus the helpers `pitchFromQuat` and `clampPitch` — and zero tests cover them. I asked for this in pass 1 and the author marked the item open in their checklist but did not act on it.

This isn't just a "would be nice." The regression that motivated the entire fix is a **behavioral guarantee** — `setYaw(-π); setPitch(0.3); rotateYaw(0.001); ASSERT_NEAR(getPitch(), 0.3, 1e-4)` is the canonical test that would have caught the original bug, and is the canonical test that would detect a future regression if someone tweaks `wrapYaw` or `pitchFromQuat`. Shipping the fix without the test leaves the next maintainer one careless edit away from re-introducing the bug.

Minimum I'd want in the same PR:

```cpp
// --- pitch round-trips ---
TEST(CameraPitchQuat, ZeroRoundTrips) { ... }
TEST(CameraPitchQuat, ArbitraryPitchRoundTripsAcrossYawSweep) {
    // For yaw ∈ {-π+ε, -π/2, 0, π/2, π-ε}, set pitch = 0.3, read it back.
    // Catches both the wrapYaw clamp and the pitchFromQuat formula.
    for (float yaw : {-IRMath::kPi + 1e-4f, -IRMath::kHalfPi, 0.0f, IRMath::kHalfPi, IRMath::kPi - 1e-4f}) {
        IRMath::vec4 q = IRPrefab::Camera::detail::quatFromYawPitch(yaw, 0.3f);
        EXPECT_NEAR(IRPrefab::Camera::detail::pitchFromQuat(q), 0.3f, 1e-4f) << "yaw=" << yaw;
    }
}

// --- the original regression ---
TEST(CameraPitchPreservation, SetYawAtNegativePiBoundaryPreservesPitch) {
    // The bug this PR fixes: at yaw = -π, atan2(0,0) returned 0 and
    // setYaw would clobber the pitch via getPitch() → 0 → setYawPitch.
    auto q = IRPrefab::Camera::detail::quatFromYawPitch(-IRMath::kPi, 0.3f);
    // simulate the read-modify-write cycle setYaw does internally:
    const float yawBack = IRPrefab::Camera::detail::yawFromQuat(q);
    const float pitchBack = IRPrefab::Camera::detail::pitchFromQuat(q);
    EXPECT_NEAR(pitchBack, 0.3f, 1e-4f);  // would fail without the wrapYaw clamp
}

// --- clampPitch boundary ---
TEST(CameraClampPitch, ClampsToLimitAtBoundary) {
    EXPECT_NEAR(IRPrefab::Camera::detail::clampPitch(IRMath::kHalfPi), IRPrefab::Camera::kPitchLimit, kTolerance);
    EXPECT_NEAR(IRPrefab::Camera::detail::clampPitch(-IRMath::kHalfPi), -IRPrefab::Camera::kPitchLimit, kTolerance);
}
```

Note: per the diff, `kPitchLimit` is at namespace `IRPrefab::Camera::detail` scope (not `IRPrefab::Camera` as my pass-1 nit suggested it could be promoted to). For the `ClampsToLimitAtBoundary` test to compile cleanly, either keep both references inside `detail::` or move both `kPitchLimit` and `clampPitch` out of `detail::`. Either is fine; my pass-1 nit about promotion stands but is non-blocking.

### Nits (not blockers)

- The renamed test names above are a suggestion — if you prefer to keep `PositivePiMapsToNegativePi` for git blame continuity, just change the expected value. Either is fine.
- `camera.hpp:7–13` (header doc) — the rewrite to mention "ZX composition" and "DETACHED canvases use the full quaternion via system_propagate_canvas_rotation" is a good change. Reads cleanly.

### Cross-host smoke

Still not warranted — CPU-only math changes in `engine/prefabs/`. No shader edits, no backend changes. Author did not add smoke labels and that's correct. ✓

### Test plan

- [x] Verified Option A patch correctness against my pass-1 prescription
- [x] Traced `wrapYaw(±π) = -π + 1e-4` against test expectations
- [x] Confirmed which other tests are robust to the new clamp (4 yaw-wrap tests + all `CameraYawQuat::*` + `QuatExtractZAngle::PiBoundaryWrapsToSymmetricNegative`)
- [x] Confirmed `pitchFromQuat`, `getPitch`, `setPitch`, `rotatePitch`, `setYawPitch`, `clampPitch` have no test coverage
- [x] Confirmed comment cleanups landed
- [x] Confirmed drag-system sign comments landed
- [ ] **Author: update `PositivePiMapsToNegativePi` and `NegativePiStaysNegativePi` expectations to `-π + 1e-4` (build-breaking without this)**
- [ ] **Author: add pitch round-trip + regression + clampPitch tests in the same PR**
- [ ] Author (optional): move `kPitchLimit` / `clampPitch` out of `detail::` if you want the constant publicly introspectable

jakildev:
Addressed feedback (Opus pass-2 checklist):
- Added `CameraClampPitch` (6 tests): zero/limit pass-through, above/below limit clamped, π/2 clamped to kPitchLimit
- Added `CameraYawPitchRoundTrip` (7 tests): pitch round-trip through `quatFromYawPitch → pitchFromQuat` at zero yaw, non-zero yaw, negative yaw, max pitch, and at the wrapYaw ε boundary (regression for the -π singularity fix)

All 34 camera tests pass; `FrameDataSunLayout.MatchesStd140Packing` failure is pre-existing on master, unrelated to this PR.

— sonnet-fleet-1

jakildev:
Heads-up — concurrent address of this PR's fleet:needs-fix. Brief log so the reviewer-in-progress isn't confused:

- 17:57 sibling agent (`Co-Authored-By: Sonnet 4.6`) pushed `ccef8c82` adding `CameraClampPitch` (6 tests) + `CameraYawPitchRoundTrip` (7 tests) — the more thorough set, addressing both Opus pass-2 checklist items.
- 17:59 my push briefly force-rewound the branch to `8d475ad4` (smaller 4-test variant) because `--force-with-lease` shares the remote-tracking ref across sibling worktrees on the same host — my local `refs/remotes/origin/...` had silently updated to `ccef8c82` between the sibling agent's push and mine, so the lease check passed even though I started work against `07cee238`.
- 17:59 detected the overwrite, restored `ccef8c82` as the tip via `--force-with-lease=<branch>:8d475ad4 origin ccef8c82:<branch>` (explicit-SHA lease guard).
- Verified: `fleet-build --target IrredenEngineTest` clean, 34/34 camera tests pass on `ccef8c82`.

Net result: PR tip is back at `ccef8c82`, which is the more thorough fix. No author work was lost beyond the ~90 seconds of branch-tip flap. Filing a fleet feedback entry on the parallel-worktree lease footgun separately.

— opus-worker-1

jakildev (commented):
## Opus re-review (pass 3) — #1258: render: camera pitch/roll for 3D orbit (#1265)

**Verdict:** approve

Re-reviewing after commit `ccef8c82` (`test: add clampPitch and pitch round-trip coverage`).

### Pass 2 items — status

**Pass 2 item 1: "`PositivePiMapsToNegativePi` will fail CI."** This was a false-positive in pass 2. Commit `07cee238` (which pass 2 already had in scope) renamed the two tests to `PositivePiClampsToEpsilonAboveNegativePi` / `NegativePiClampsToEpsilonAbove` and updated their expected values to `-IRMath::kPi + 1e-4f` — pass 2 missed those renames. Verified by reading `test/render/camera_yaw_test.cpp:23–34` at HEAD: the asserts are `EXPECT_NEAR(yaw, -IRMath::kPi + 1e-4f, kTolerance)` and pass under the new `wrapYaw`. No CI breakage on this front.

**Pass 2 item 2: "Six new pitch APIs have zero test coverage."** Addressed in `ccef8c82`. The new commit adds:

- **`CameraClampPitch` × 6** — zero pass-through, ±limit pass-through, above/below-limit clamping, and `kHalfPi` → `kPitchLimit` (`camera_yaw_test.cpp:60–89`). Covers all boundary conditions of `clampPitch`.
- **`CameraYawPitchRoundTrip` × 7** — zero pitch, ±pitch at zero yaw, pitch at ±π/2 yaw, max pitch at arbitrary yaw, and crucially `PitchSurvivesWrapYawClamp` (`camera_yaw_test.cpp:97–141`). The last test is the regression guard: it composes `quatFromYawPitch(wrapYaw(-π), 0.3f)` and asserts `pitchFromQuat` recovers 0.3f. Exactly the test I asked for in pass 2.

`kPitchLimit` and `clampPitch` stayed in `detail::` and the tests reference them via `IRPrefab::Camera::detail::*`. That compiles cleanly. My pass-1 promotion suggestion remains optional.

### Independent checks (not previously confirmed)

- **`IRMath::max(float, float)` resolves correctly.** `IRMath::max` (`ir_math.hpp:118`) is a `template<typename VecType>` over `glm::max`. With both args float, `VecType` deduces to `float` and dispatches to scalar `glm::max`. The `wrapYaw` clamp at `camera.hpp:73` compiles.
- **`pitchFromQuat` validity contract.** The comment claims `cos(yaw/2) > 0` is "guaranteed for all callers" because `wrapYaw` keeps yaw above -π. This is true for the new yaw/pitch path (all writers go through `setYawPitch` → `wrapYaw`). It is *not* an absolute guarantee for any caller of `getPitch()` — if some other writer ever puts a quaternion with `yaw = ±π` into `C_LocalTransform.rotation_`, `pitchFromQuat` returns 0. No such writer exists in this diff. Worth a comment if/when a non-ZX rotation driver lands; not a blocker now.
- **Drag systems snapshot both yaw and pitch at press.** `system_camera_key_drag_rotate.hpp:45` and `system_camera_mouse_rotate.hpp:40` both capture `dragStartYaw_ = getYaw()` and `dragStartPitch_ = getPitch()` on the PRESSED frame. Sign-convention comments (`// negate: screen +Y is down; pitch up requires negative deltaY`) are correct.
- **`MaxPitchAtArbitraryYawRoundTrips`** (`camera_yaw_test.cpp:127–132`) uses `pitch = kPitchLimit ≈ 1.5608`. `sin(pitch/2) ≈ 0.7068`, `cos(pitch/2) ≈ 0.7074`. `atan2` round-trips cleanly. Fine.
- **`mergeStateStatus = UNSTABLE`** is from the perf-gate placeholder comment at `2026-05-28T00:57:01Z` ("no committed baseline yet — skipping comparison"). Not a blocker.

### Nits (not blockers)

- The old `rotateYaw` doc explicitly warned that the helper "rebuilds rotation as `quatAxisAngle(z, yaw+delta)`, clobbering any non-Z components" — useful warning that any future GRID + SO(3) driver coexistence needs to be tracked. The new `rotateYaw` doc is just "Add @p delta (radians) to the camera's yaw, preserving pitch." That's accurate for the current ZX-only world, but the multi-driver coexistence warning has been lost. A one-line comment on `setYawPitch` noting that any non-yaw/pitch components on `C_LocalTransform.rotation_` are wiped would help the next reader.
- Header doc at `camera.hpp:7–13` correctly captures the ZX convention. Reads cleanly. ✓

### Cross-host smoke

Still not warranted — CPU-only math in `engine/prefabs/render/`. No shader, no backend changes. Author correctly omitted smoke labels. ✓

### Bug note for the human (not blocking this PR)

The fleet reviewer pipeline lost the verdict label between `2026-05-28T00:54:10Z` (`fleet:needs-fix` added by Opus pass 2) and `00:55:25Z` (`fleet:needs-fix` removed by `jakildev` with no concurrent commit/comment). The author's next push at `00:57:49Z` did add `fleet:changes-made` correctly. Worth a follow-up issue: something — perhaps a stale `--remove-label` from a worker that started preparing the fix and then aborted — is racing the reviewer's verdict label. The hard-rule guard in REVIEWER-PROTOCOL.md correctly flagged this as a re-review candidate (UNLABELED event present), so the failure mode was caught, but a deterministic fix would be cleaner than relying on the guard.

### Test plan

- [x] Verified pass-2 item 1 was a false-positive (tests already renamed/updated in `07cee238`)
- [x] Verified pass-2 item 2 is addressed by `ccef8c82` (13 new tests across `CameraClampPitch` + `CameraYawPitchRoundTrip`)
- [x] Confirmed `IRMath::max(float, float)` resolves
- [x] Confirmed drag systems snapshot both yaw and pitch at press, with correct sign convention
- [x] Confirmed `PitchSurvivesWrapYawClamp` is the canonical regression test for the original bug
- [x] CPU-only math + drag systems; no shader / backend changes; no smoke labels needed

Ready to merge.

PR #1267, opened 2026-05-28
#1264: fleet: fix latch_usage_observation null overwrite in fleet-claude-stream

### LINKED ISSUE #1264: fleet-claude-stream: latch_usage_observation overwrites warning utilization with null
**Context:** Surfaced 2026-05-27 while validating the rate-limit gate's effectiveness after a fleet run where workers continued past the 80% `five_hour` threshold. Grepping `~/.fleet/logs/*.log` shows the SDK shipped `allowed_warning` events at `five_hour` utilization 0.90, 0.91, 0.92, 0.94, 0.95 across the session — every one of those should have tripped the 0.80 gate. The dispatcher log has **zero** `usage gate closed` entries; the gate never fired.

**Root cause:** the SDK's `rate_limit_event` payload has two shapes:

```json
// status:"allowed" (normal, below warning band) — NO utilization field
{"status":"allowed","resetsAt":...,"rateLimitType":"five_hour","overageStatus":"rejected","isUsingOverage":false}

// status:"allowed_warning" (above warning band) — utilization included
{"status":"allowed_warning","resetsAt":...,"rateLimitType":"five_hour","utilization":0.94,"surpassedThreshold":0.9}
```

[scripts/fleet/fleet-claude-stream:131-166](scripts/fleet/fleet-claude-stream) `latch_usage_observation` calls `info.get("utilization")` unconditionally and writes the result. When a regular `"allowed"` event arrives *after* a warning event, it overwrites the latched utilization with `null`. The dispatcher's gate logic at [fleet-dispatcher:591](scripts/fleet/fleet-dispatcher) does `if not isinstance(util, (int, float)): continue`, so a `null` value is discarded and the gate falls through to `print("open")` (line 638). Net result: between warning bursts the gate silently re-opens even though utilization is still high.

Current latched state on this host confirms the failure mode:

```
~/.fleet/state/usage/five_hour.json:
  {"rateLimitType":"five_hour","utilization":null,"resetsAt":1779936000,...}
```

**Fix:**

In `latch_usage_observation` ([scripts/fleet/fleet-claude-stream:131](scripts/fleet/fleet-claude-stream)), skip the write when `utilization` is `None`:

```python
def latch_usage_observation(rl_type, info):
    if not isinstance(rl_type, str) or not rl_type:
        return
    util = info.get("utilization")
    if util is None:
        # SDK ships `status:"allowed"` events without `utilization`
        # (only `allowed_warning` and `rejected` carry the value).
        # Preserve the last known warning observation; the
        # dispatcher's USAGE_STALE_SECONDS (default 1h) ages it
        # out naturally when the window resets.
        return
    safe_type = "".join(c if c.isalnum() or c in "_-" else "_" for c in rl_type)
    ...
```

**Bonus cleanup:** `~/.fleet/state/usage/_.json` is a junk file from an event that arrived with neither rateLimitType nor utilization (the emit path defaults `rateLimitType` to `"?"` at [fleet-claude-stream:212](scripts/fleet/fleet-claude-stream) and the latch's safe_type sanitizer at line 144 converts `"?"` → `"_"`). Harmless to gating but worth either filtering at the source (don't latch junk types) or rejecting at the dispatcher's read path (`rl_type == "?" or rl_type == "_"`).

**Validation:**

1. Add a test fixture at `scripts/fleet/tests/` that feeds the stream both an `allowed_warning` (utilization=0.85) followed by an `allowed` (no utilization) and asserts the latched JSON retains `utilization: 0.85`, not `null`.
2. Manual smoke: after the fix, run a session that crosses the warning threshold, then exits cleanly; verify `~/.fleet/state/usage/five_hour.json` retains the last warning value. Restart the fleet, confirm `dispatcher.log` shows `usage gate closed:five_hour util=85% (>= 80%)` until the window's `resetsAt` + `RESET_GRACE_SECONDS` (10 min) passes.
3. The seven_day case (95% threshold) is unaffected by this fix — the SDK reports utilization once it crosses the 0.75 `surpassedThreshold`, so the latch will work the same way for that window.

**Model:** sonnet
**Area:** scripts/fleet
**Blocked by:** (none)

jakildev:
## Summary
- `latch_usage_observation` now skips the JSON write when `utilization` is absent (SDK `status:\"allowed\"` events have no utilization field)
- Preserves the last known warning value so bare `allowed` bursts between warning events can't silently re-open the dispatcher's usage gate
- Also guards against the `\"?\"` sentinel `rateLimitType` (produced when the field is absent from the event payload), which would sanitize to `\"_\"` and create a junk `_.json` the dispatcher reads

## Test plan
- [x] 4 unit tests in `scripts/fleet/tests/test_fleet_claude_stream_latch.py` — all pass (warning write, allowed-preserves-warning, junk-type skip, first-event-no-util no-write)
- [ ] Observe fleet run after merge: `~/.fleet/state/usage/five_hour.json` should retain non-null utilization after an `allowed` burst follows a `allowed_warning` event

Closes #1264

### DIFF
--- a/scripts/fleet/fleet-claude-stream
+++ b/scripts/fleet/fleet-claude-stream
@@ -141,12 +141,24 @@ def latch_usage_observation(rl_type, info):
     """
     if not isinstance(rl_type, str) or not rl_type:
         return
+    # "?" is the sentinel default used when rateLimitType is absent from the
+    # event payload; sanitizing it yields "_", which creates a junk _.json
+    # that the dispatcher would read. Skip it.
+    if rl_type == "?":
+        return
     safe_type = "".join(c if c.isalnum() or c in "_-" else "_" for c in rl_type)
     if not safe_type:
         return
+    util = info.get("utilization")
+    if util is None:
+        # status:"allowed" events carry no utilization field; only
+        # "allowed_warning" and "rejected" do. Preserve the last known
+        # warning observation so a bare "allowed" burst between warning
+        # events can't silently re-open the dispatcher's gate.
+        return
     payload = {
         "rateLimitType": rl_type,
-        "utilization": info.get("utilization"),
+        "utilization": util,
         "resetsAt": info.get("resetsAt"),
         "observed_at": int(time.time()),
     }
--- a/scripts/fleet/fleet-claude-stream
+++ b/scripts/fleet/fleet-claude-stream
@@ -149,6 +149,8 @@ def latch_usage_observation(rl_type, info):
     safe_type = "".join(c if c.isalnum() or c in "_-" else "_" for c in rl_type)
     if not safe_type:
         return
+    if safe_type == "_":
+        return
     util = info.get("utilization")
     if util is None:
         # status:"allowed" events carry no utilization field; only
--- a/scripts/fleet/tests/test_fleet_claude_stream_latch.py
+++ b/scripts/fleet/tests/test_fleet_claude_stream_latch.py
@@ -1,0 +1,87 @@
+"""Tests for fleet-claude-stream's latch_usage_observation.
+
+Covers:
+  - allowed_warning event (has utilization) => file written with correct value
+  - allowed event (no utilization) after warning => existing file NOT overwritten
+  - junk "?" rateLimitType => no file written (no _.json created)
+  - valid type with no utilization from the start => no file written
+"""
+import importlib.machinery
+import importlib.util
+import json
+import os
+import pathlib
+import tempfile
+import unittest
+
+_SCRIPT = pathlib.Path(__file__).parent.parent / "fleet-claude-stream"
+_loader = importlib.machinery.SourceFileLoader("fleet_claude_stream", str(_SCRIPT))
+_spec = importlib.util.spec_from_loader("fleet_claude_stream", _loader)
+_mod = importlib.util.module_from_spec(_spec)
+
+
+class LatchUsageObservation(unittest.TestCase):
+    def setUp(self):
+        self._tmpdir = tempfile.TemporaryDirectory()
+        self._usage_dir = pathlib.Path(self._tmpdir.name) / "usage"
+        self._usage_dir.mkdir()
+        _loader.exec_module(_mod)
+        _mod.USAGE_DIR = self._usage_dir
+
+    def tearDown(self):
+        self._tmpdir.cleanup()
+
+    def _read(self, name):
+        p = self._usage_dir / f"{name}.json"
+        if not p.exists():
+            return None
+        return json.loads(p.read_text())
+
+    def test_warning_event_writes_utilization(self):
+        _mod.latch_usage_observation(
+            "five_hour",
+            {"utilization": 0.85, "resetsAt": "2026-05-28T00:00:00Z"},
+        )
+        data = self._read("five_hour")
+        self.assertIsNotNone(data)
+        self.assertEqual(data["utilization"], 0.85)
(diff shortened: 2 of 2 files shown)


jakildev (commented):
## Review — #1264: fleet: fix latch_usage_observation null overwrite in fleet-claude-stream (#1267)

**Verdict:** approve

### Summary

Clean, well-targeted two-guard fix. The root cause (SDK `status:"allowed"` events lack a `utilization` field, so the unconditional `info.get("utilization")` write produced `null` in the latched JSON, silently re-opening the dispatcher's usage gate between warning bursts) is correctly diagnosed and addressed. The `"?"` sentinel guard closes a separate hole where an absent `rateLimitType` sanitized to `"_"`, creating a junk `_.json` the dispatcher would read.

### Checklist

- **`util is None` guard correctness** — `info.get("utilization")` returns `None` when the key is absent (not when it's `0`). The check `if util is None:` is identity-safe: a genuine `utilization=0.0` still passes through and gets written. Correct. ✓
- **`"?"` sentinel guard** — `"?"` sanitizes to `"_"` under the existing character filter, which would create `_.json`. The early return at `if rl_type == "?":` fires before sanitization. Correct. ✓
- **Pre-existing `if not safe_type:` guard** — the existing empty-after-sanitize check is still reachable for types that consist entirely of non-alnum/non-`_-` characters other than `"?"`. The new guard doesn't break it. ✓
- **Atomic write path unchanged** — temp-file + `os.replace` pattern is untouched. ✓
- **All 4 unit tests pass** — confirmed locally (`python3 -m unittest`). ✓
- **`emit(f"... util={util:.0%}")` at line 225** — `util` there is `info.get("utilization", 0)`, so `allowed` events display `0%`. Pre-existing behavior, not changed here. ✓

### Nits

- `scripts/fleet/fleet-claude-stream:147` — The `"?"` guard is specifically targeted to that exact sentinel. Any other single weird-char type (e.g. `"!"` → `"_"`, `"@"` → `"_"`) would still create a junk `_.json`. A more defensive post-sanitization check `if safe_type == "_": return` after the existing `if not safe_type:` check would catch all such cases without touching the `"?"` guard. Minor — the `"?"` sentinel is the only documented SDK-produced value with this property.

- `scripts/fleet/tests/test_fleet_claude_stream_latch.py:17-20` — Module-level `_loader`, `_spec`, `_mod = importlib.util.module_from_spec(_spec)` are created once at import time, and `exec_module(_mod)` is called per-test in `setUp`. This works correctly (re-execution reinitializes module globals), but is slightly non-obvious. Standard pattern would be to call `module_from_spec` inside `setUp` for a fresh module each test. No correctness risk given `unittest.TestCase`'s sequential execution, but a reader unfamiliar with the pattern may be confused. Minor.

- [ ] Observe a live fleet run: after an `allowed_warning` event fires, confirm `~/.fleet/state/usage/five_hour.json` retains the non-null utilization through subsequent `allowed` bursts

Opus recheck not required.

jakildev:
Addressed nits:
- Added `if safe_type == "_": return` guard after the existing `if not safe_type:` check in `fleet-claude-stream:latch_usage_observation`, so any single-char type that sanitizes to `"_"` (e.g. `"!"`, `"@"`) is also skipped, not just the documented `"?"` sentinel.
- Moved `importlib.util.module_from_spec(_spec)` call into `setUp` so each test gets a fresh module object; updated all test methods to use `self._mod` instead of the module-level `_mod`. All 4 tests pass.

Issue #1256, opened 2026-05-27
render: trixel checkerboard/lattice artifact on voxel faces at non-zero cardinal yaw


jakildev:
## Symptom

When the camera is rotated away from the default yaw (cardinalIndex 0), voxel cube side faces (X and Y faces) display a severe every-other-trixel checkerboard/lattice pattern. At close zoom the cubes look like wireframe cages rather than solid objects — you can see the platform through the gaps. The Z face (top) remains solid at all rotations.

### Evidence

Screenshots in `/build/creations/demos/lighting/save_files/screenshots/`:
- **screenshot_000002** — default yaw: cubes render solid, all faces clean
- **screenshot_000003** — ~180° yaw: cubes show horizontal banding/gaps on side faces
- **screenshot_006/007** — zoomed in at rotated yaw: dramatic checkerboard lattice on X/Y faces; Z face solid; platform visible through gaps
- **screenshot_010** — extreme zoom on floor: same alternating trixel fill on the grey platform

The same artifact appears in the IRShapeDebug demo at non-zero yaw.

## Reproduction

```bash
fleet-build --target IRLightingSunOrbit
fleet-run IRLightingSunOrbit
# Alt+click-drag to rotate camera ~90° or ~180°
# Scroll to zoom in on the cyan cubes
# Observe the checkerboard on side faces
```

Or with auto-screenshot:
```bash
fleet-build --target IRShapeDebug
fleet-run IRShapeDebug --auto-screenshot 10
# Default shots are at yaw=0; need to add yaw=π/2 and yaw=π shot configs
```

### Pipeline path
1. `c_voxel_visibility_compact.glsl` — rotates voxel positions via `rotateCardinalZ`, checks iso viewport bounds → compacted index buffer
2. `c_voxel_to_trixel_stage_1.glsl` — rotates position, computes `base = trixelFrameOffset + pos3DtoPos2DIso(rotated)`, writes `encodeDepthWithFace(rawDepth, face)` via `imageAtomicMin` to distance texture
3. `c_voxel_to_trixel_stage_2.glsl` — same position math, reads distance texture, writes color where depth matches
4. `f_trixel_to_framebuffer.glsl` — reads canvas textures, applies `trixelFramebufferSamplePosition` parity adjustment, writes to screen framebuffer

### Key functions involved
- `rotateCardinalZ()` in `ir_iso_common.glsl:251-256` — integer basis permutation for R_z(-rasterYaw)
[...]

PR #1285, opened 2026-05-28
#1284: fleet-dispatcher: periodic safety rearm for worker roles

### LINKED ISSUE #1284: fleet-dispatcher: periodic safety rearm for worker roles when actionable queue + zero active dispatch
**Context:** Observed 2026-05-28 fleet idle-deadlock. `~/.fleet/state/projections/opus-worker.json` showed `#1256` with `owner: free`, `blocked_by: (none)`, `model: opus` — claim-ready work. Active opus-worker dispatch count was 0. Both opus-worker panes idle in shell. Usage gate open. But no `~/.fleet/state/triggers/opus-worker` file existed, so the dispatcher had nothing to fire. State persisted for 13+ minutes until manually `touch`ed.

**Root cause:** the scout writes worker triggers based on **projection-diff** — it only fires `opus-worker` when the projection's hash *changes* from the previous tick. When the projection settles into a stable state with actionable-but-unchanged work (e.g. a claim got released and the resulting "free" state matches a previous tick's hash, or stale claim labels on other issues mask the change), no trigger gets written. Workers idle indefinitely.

`scripts/fleet/fleet-dispatcher`'s `rearm_periodic_meta_triggers` ([line 988](scripts/fleet/fleet-dispatcher)) ONLY rearms `merger` as a scout-watchdog fallback. Worker roles (`opus-worker`, `sonnet-author`) have no periodic safety net — by design, per the comment block at line 994-1000 (the earlier blanket merger rearm burned ~$0.30/iter × 12/hr on no-op runs). But that decision left workers with no resilience to projection-diff blind spots.

**Compatibility with stacking:** the stack-claim mechanism (`fleet-claim claim --stackable-on <pr-url>`) intentionally claims blocked tasks. The fix below predicates ONLY on tasks with `owner == "free"` AND `blocked_by == "(none)"`, so stack-claimed tasks (which show `owner=mac/<agent>` once claimed, regardless of `blocked_by`) are excluded from the safety predicate and don't influence whether the rearm fires. Tasks with a `stackable_blocker_pr` enrichment but still `owner == "free"` DO count — the worker will see them and decide whether to claim plain or with `--stackable-on`.

**Fix:**

Extend `rearm_periodic_meta_triggers` in [`scripts/fleet/fleet-dispatcher`](scripts/fleet/fleet-dispatcher) to also rearm `opus-worker` and `sonnet-author` when:

1. The role's projection JSON has at least one task with `owner == "free"` AND `blocked_by == "(none)"` AND `model == "<role's model tier>"`, AND
2. The role's active dispatch count (via the existing `count_role_active` helper) is 0, AND
3. The role's trigger file `~/.fleet/state/triggers/<role>` does not already exist.

All three predicates must hold. Touch the trigger file. The dispatcher's next 10s tick consumes it.

Sketch:

```bash
# Inside rearm_periodic_meta_triggers, after the scout-watchdog block:
# Worker safety rearm: scout's projection-diff can miss transitions
# (e.g. a manually-cleared claim label, or a claim-then-bail iteration
# leaving stale labels that hide the next-tick transition). When the
# projection has unblocked work, no active worker for that role, and
# no pending trigger, rearm so the queue moves.
python3 - <<'PY' || true
import json, pathlib, os

STATE = pathlib.Path("~/.fleet/state").expanduser()
TRIG = STATE / "triggers"
PROJ = STATE / "projections"

ROLE_MODEL = {"opus-worker": "opus", "sonnet-author": "sonnet"}

for role, model in ROLE_MODEL.items():
    trig = TRIG / role
    if trig.exists():
        continue
    proj_path = PROJ / f"{role}.json"
    if not proj_path.is_file():
        continue
    try:
        proj = json.loads(proj_path.read_text())
    except (OSError, json.JSONDecodeError):
        continue
    has_free = any(
        t.get("owner") == "free"
        and (t.get("blocked_by") or "").strip() in ("(none)", "", None)
        and t.get("model") == model
        for t in proj.get("tasks_open", [])
    )
    if not has_free:
        continue
    # Active count is the existing helper; this needs to be exposed to
    # the inline python via env or a subprocess call.
    # Implementation detail: pass the count in via env, or shell out
    # to `fleet-dispatcher --count-active <role>`.
    # ... (see implementation notes below)
PY
```

**Implementation notes:**

- The inline-python needs the active-dispatch count for each role. Two options:
  1. Call `bash` back out via `subprocess.run` to read the existing `count_role_active` function. Adds latency, marginal.
  2. Compute it inline by reading `~/.fleet/state/dispatch/*.json` (the dispatch records) and counting entries whose `role` matches. Cleaner; the dispatch records are already JSON.
  Recommend (2).
- The role list (`ROLE_MODEL`) should mirror `DISPATCHED_ROLES` minus reviewer/merger/smoke-worker. Hard-coding is acceptable; the worker roles change rarely.
- The function runs every `PERIODIC_REARM_INTERVAL_SECONDS = 300s` (5 min). At worst, an idle-deadlocked queue waits up to 5 min before the safety rearm fires. Combined with scout's 30s diff cadence, normal-case latency stays ~30s — the safety rearm only matters in the diff-blind-spot edge case.
- Cost: a worker iteration is ~$0.30. The predicate gating (free unblocked task AND zero active worker AND no existing trigger) means it won't fire on idle fleets where everything is claimed or all work is blocked. The cost surface is "fleet has free work it forgot to dispatch", which IS work the fleet should be doing anyway.

**Validation:**

1. Reproduce the deadlock: claim all opus tasks in a queue, then release one issue's labels manually so it goes free, observe scout NOT firing opus-worker (projection-diff bypass). Within 5 min, the rearm should fire and the dispatcher should pick up the free task.
2. Negative test: empty queue (no free tasks). Confirm the rearm does NOT fire — no trigger file appears.
3. Negative test: free task exists but model mismatch (e.g. `model: sonnet` task with opus-worker rearm). Confirm opus-worker rearm does NOT fire.
4. Negative test: free task exists AND model matches AND active dispatch count > 0 (worker already busy). Confirm no rearm.
5. Stacking: pre-populate a `--stackable-on` claim on a blocked task. Confirm the rearm STILL fires for an *unrelated* free task with model match — the stack-claimed entry shows `owner=mac/<agent>`, not `free`, so it doesn't affect the predicate.

**Related:** #1266 (end_turn-after-clean-skill), #1276 (reset-sweep stale labels). Both address adjacent failure modes that contribute to the projection settling into "stale claim labels mask a transition" states. This issue is the safety net for whenever any of those bugs slip a transition past the scout's hash diff.

**Model:** sonnet
**Area:** scripts/fleet
**Blocked by:** (none)

jakildev:
## Summary

- Scout's projection-diff is the only path that arms `opus-worker` and `sonnet-author`. When the projection settles into a state with actionable+unchanged work, no trigger gets written and workers idle indefinitely (observed 2026-05-28: opus-worker.json had `#1256 owner=free, blocked=(none), model=opus` for 13+ minutes with zero active workers until the trigger file was manually touched).
- Extends `rearm_periodic_meta_triggers` in [`scripts/fleet/fleet-dispatcher`](scripts/fleet/fleet-dispatcher) to also rearm worker roles when **all three** predicates hold:
  1. role's projection has at least one task with `owner=free` AND `blocked_by=(none)` AND `model` matching the role's tier
  2. `count_active_for_role` returns 0 for this role
  3. `~/.fleet/state/triggers/<role>` does not already exist
- Reuses the existing 300s `PERIODIC_REARM_INTERVAL_SECONDS` cadence — worst-case deadlock recovery is 5 min; normal case stays at scout's 30s cadence.

## Stacking compatibility

Stack-claimed tasks (via `fleet-claim claim --stackable-on <pr-url>`) show `owner=mac/<agent>` once the claim lands, so they fail predicate (1) and don't influence the safety rearm. Tasks with a `stackable_blocker_pr` enrichment but still `owner=free` DO count — the worker sees them and decides at claim time whether to go plain or stacked. The lex-min tie-break on the issue's `fleet:claim-<host>-<agent>` label remains the cross-host atomicity primitive for the actual claim.

## Cost

A worker iteration is ~$0.30. Predicate gating ensures the rearm only fires when the fleet has free unblocked work it forgot to dispatch — work the fleet should be doing anyway. No idle-fleet token burn.

## Test plan

- [ ] Reproduce the deadlock manually: claim all free opus tasks, manually clear one issue's claim labels so it goes back to `owner=free`. Verify scout does NOT fire opus-worker (projection-diff blind spot). Wait up to 5 min; verify dispatcher log shows `worker-rearm: re-armed opus-worker (projection has actionable free work, no active iteration)` and the dispatcher fires opus-worker on the next 10s tick.
- [ ] Negative test: empty queue. Verify no trigger file appears (predicate (1) fails).
- [ ] Negative test: free task exists, but model mismatch (`sonnet` task with opus-worker rearm). Verify opus-worker rearm does NOT fire.
- [ ] Negative test: free task exists, model matches, but `count_active_for_role` returns > 0. Verify no rearm (predicate (2) fails).
- [ ] Negative test: pre-write a trigger file for the role. Verify rearm short-circuits without invoking python (predicate (3) fails).
- [ ] Stacking compatibility: pre-stamp a `fleet:claim-mac-opus-worker-1` label on a `blocked_by != (none)` task. Verify the rearm STILL fires for an *unrelated* free task with model match (the stack-claimed entry has `owner != free`, doesn't affect the predicate).

Closes #1284.

### DIFF
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -986,10 +986,11 @@ release_dispatcher_lock() {
 }
 
 rearm_periodic_meta_triggers() {
-    # Scout-watchdog safety net only. With merger's projection now
-    # tightened (only durable signals, not the merger's own toggled
-    # labels), the scout-driven trigger path is sufficient — the
-    # merger only fires when actionable state actually changes.
+    # Scout-watchdog safety net + periodic worker safety rearm.
+    # Scout-watchdog half: with merger's projection now tightened
+    # (only durable signals, not the merger's own toggled labels),
+    # the scout-driven trigger path is sufficient — the merger only
+    # fires when actionable state actually changes.
     #
     # The earlier `maybe_arm("merger", merger_actionable)` periodic
     # re-arm fired every 5 min whenever any approved/blocked PR
@@ -1149,9 +1150,9 @@ main() {
 
         cleanup_stale_dispatches
 
-        # Periodic safety net for chain-broken meta-roles (queue-manager,
-        # merger). Runs first so any rearm-written triggers are picked
-        # up by the normal dispatch_role pass on the same tick.
+        # Scout-watchdog safety net + periodic worker safety rearm.
+        # Runs first so any rearm-written triggers are picked up by
+        # the normal dispatch_role pass on the same tick.
         if (( tick_count % rearm_ticks == 0 )); then
             rearm_periodic_meta_triggers
         fi
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -1039,6 +1039,82 @@ if scout_unhealthy():
         trig.touch()
         print("scout-watchdog: re-arming merger (state.json stale)")
 PY
+
+    # Worker safety rearm. Scout's projection-diff can miss transitions
+    # (a manually-cleared claim label, a claim-then-bail iteration that
+    # leaves stale labels hiding the next-tick transition, etc.) and
+    # strand the queue with unblocked free work and no trigger. When
+    # the role's projection has at least one actionable task AND no
+    # worker for that role is active AND no trigger is already pending,
+    # re-arm the role.
+    #
+    # Stacking compatibility: stack-claimed tasks (via
+    # `--stackable-on`) show `owner != "free"` once the claim lands,
+    # so they don't pass the predicate. Tasks with a
+    # `stackable_blocker_pr` enrichment but still `owner == "free"`
+    # DO count — the worker will see them and decide whether to
+    # claim plain or as a stack.
+    #
+    # Cost: a worker iteration is ~$0.30. The predicate (free unblocked
+    # task AND zero active worker AND no existing trigger) means it
+    # only fires when the fleet has work it forgot to dispatch — work
+    # the fleet should be doing anyway. No idle-fleet token burn.
+    local _role _active
+    for _role in "opus-worker" "sonnet-author"; do
+        # Skip if a trigger is already pending — scout's normal path
+        # is doing its job and the dispatcher will fire on next tick.
+        if [[ -f "$TRIGGERS_DIR/$_role" ]]; then
+            continue
+        fi
+        # Skip if any worker for this role is in-flight.
+        _active=$(count_active_for_role "$_role")
+        if [[ "$_active" -gt 0 ]]; then
+            continue
+        fi
+        # Read the role's projection and check the predicate.
+        TRIGGERS_DIR="$TRIGGERS_DIR" \
+        STATE_DIR="$STATE_DIR" \
+        ROLE="$_role" \
+        python3 - <<'PY' || true
+import json
+import os
+import pathlib
+
+trig_dir = pathlib.Path(os.environ["TRIGGERS_DIR"])
+state_dir = pathlib.Path(os.environ["STATE_DIR"])
+role = os.environ["ROLE"]
+
+ROLE_MODEL = {"opus-worker": "opus", "sonnet-author": "sonnet"}
+model = ROLE_MODEL.get(role)
+if not model:
+    raise SystemExit(0)
+
+proj_path = state_dir / "projections" / f"{role}.json"
+if not proj_path.is_file():
+    raise SystemExit(0)
+try:
+    proj = json.loads(proj_path.read_text())
+except (OSError, json.JSONDecodeError):
+    raise SystemExit(0)
+
+# Predicate: at least one task that is unowned, unblocked, and matches
+# this role's model tier. The projection's `blocked_by` is rendered as
+# "(none)" (literal) when there are no blockers.
+def is_actionable(task):
+    if task.get("owner") != "free":
+        return False
+    blocked = (task.get("blocked_by") or "").strip()
+    if blocked and blocked != "(none)":
+        return False
+    return task.get("model") == model
+
+if any(is_actionable(t) for t in (proj.get("tasks_open") or [])):
+    trig = trig_dir / role
+    if not trig.exists():
+        trig.touch()
+        print(f"worker-rearm: re-armed {role} (projection has actionable free work, no active iteration)")
+PY
+    done
 }
 
 main() {


jakildev (commented):
## Review — #1284: fleet-dispatcher: periodic safety rearm for worker roles

**Verdict:** approve

### Summary

Correct and well-scoped fix for the idle-queue deadlock. The three-predicate gate (free unblocked model-matching task AND zero active workers AND no existing trigger) precisely targets the failure mode described in the body: scout's projection-diff only fires on state *transitions*, so a task that was already free before the last projection snapshot never gets a new trigger edge. The safety rearm fires every 300s and touches the trigger file only when all three predicates hold, so it carries zero idle-fleet burn cost.

The bash/Python structure mirrors the existing scout-watchdog pattern exactly. The `|| true` suffix on both `python3` heredocs is correct — it prevents a non-zero Python exit from killing the enclosing bash function under `set -euo pipefail`. The bash pre-check on the trigger file (`if [[ -f "$TRIGGERS_DIR/$_role" ]]; then continue; fi`) is an intentional optimization that avoids spawning Python at all when scout's normal path already wrote the trigger; the Python `if not trig.exists()` is the real atomic gate. The `(task.get("blocked_by") or "").strip() != "(none)"` handling is correct for the `"(none)"` literal the projection emits when there are no blockers.

`ROLE_MODEL` dict maps roles to model tiers correctly, and excluding `sonnet-reviewer` / `opus-reviewer` from the loop is correct — those roles are triggered by the PR projection (not the task projection) and don't have the same idle-queue failure mode.

---

### Nits

- `scripts/fleet/fleet-dispatcher:989` — The function-level docstring for `rearm_periodic_meta_triggers()` still reads "Scout-watchdog safety net only" and lists only merger/queue-manager. The function now also houses the worker safety rearm, so the description is stale. Suggest updating the opening comment to reflect both responsibilities, e.g. "Scout-watchdog safety net + periodic worker safety rearm."

- `scripts/fleet/fleet-dispatcher:1152` — The call-site comment in `main()` says "Periodic safety net for chain-broken meta-roles (queue-manager, merger)". Same staleness — the worker rearm now also runs here on every tick that crosses the rearm interval.

---

- [ ] Verify `count_active_for_role` is accessible from inside `rearm_periodic_meta_triggers()` without forward-declaration issues (it's defined at line 689 before the function at 988 — should be fine, but worth a quick grep of the call order).
- [ ] Confirm dispatcher log shows `worker-rearm: re-armed opus-worker …` (not silent) when the rearm fires — the `print(f"worker-rearm: …")` goes to Python stdout, which must be captured by the caller's log pipeline.

---

Opus recheck not required.

jakildev:
Addressed feedback:
- Updated `rearm_periodic_meta_triggers()` docstring (line 989) from "Scout-watchdog safety net only" to "Scout-watchdog safety net + periodic worker safety rearm"
- Updated call-site comment in `main()` (line 1152) to mention worker safety rearm (opus-worker, sonnet-author) alongside the meta-roles

Both comments now accurately reflect the function's dual responsibility.

jakildev:
Addressed nits:

- `scripts/fleet/fleet-dispatcher:989` — Updated `rearm_periodic_meta_triggers()` opening comment from "Scout-watchdog safety net only" to "Scout-watchdog safety net + periodic worker safety rearm." — the function now has two responsibilities and the docstring reflects both.
- `scripts/fleet/fleet-dispatcher:1152` — Updated call-site comment from "Periodic safety net for chain-broken meta-roles (queue-manager, merger)" to "Scout-watchdog safety net + periodic worker safety rearm." — same staleness, same fix.

— sonnet-fleet-2

Issue #1278, opened 2026-05-28
render: implement visible-triplet × exposed-mask voxel face rasterization (fixes #1256 cardinal stripes)


jakildev:
**Part of epic:** #936 (Rotation architecture)
**Blocked by:** #1277 (design doc must merge first — the spec must be on master before implementation starts)
**Design source of truth:** \`docs/design/voxel-face-rasterization.md\` (lands via PR #1277)
**Supersedes:** #1275's four-option NEEDS-DESIGN escalation
**Related:** #1256 (the bug), #1272 (per-entity SO(3) consumes this), PR #1265 (camera pitch), #1271 (continuous-yaw validation demo)

## Context

#1256 is the stripe/checkerboard artifact on voxel cube faces at non-zero cardinal yaw. #1275 produced a partial fix and escalated `fleet:design-blocked` with four candidate options. The architect resolved the escalation with a fifth model that subsumes all four — specified in full in **\`docs/design/voxel-face-rasterization.md\`** (PR #1277). Read that doc first; it is the source of truth for this work.

## Model (summary — full spec in the design doc)

**Visible-face triplet × exposed-face mask.** A voxel emits a face iff it is both:
- **camera-visible** — one of the three camera-facing faces, a per-frame function of the camera quaternion (cardinal + residual + pitch), resolved on CPU and uploaded as a `FaceRenderContext[3]` UBO; and
- **exposed** — its neighbor cell is empty/absent, a camera-independent `exposedFaces : uint8` mask set at pool build/mutate time.

Six faces are six distinct enum values, not three axes with a ± sign.

## Scope

1. **`C_VoxelPool` exposed-face mask.** Add `exposedFaces : uint8` per voxel; populate at insert (probe 6 neighbors) and update affected neighbors on insert/remove. Fold into the existing visibility-compact neighbor query rather than a new pass.
2. **Per-frame visible-triplet resolver.** CPU helper: camera quaternion → `FaceRenderContext[3]` (face id, deformation matrix, outward normal, AO tangents). Generalizes today's `faceDeform[3]` UBO. Upload per frame.
3. **Raster emit (GLSL + Metal).** `c_voxel_to_trixel_stage_{1,2}`: walk `visibleFaces[0..2]`, branch on `(exposedFaces & (1<<face))`, emit with the per-face context. Interior voxels emit nothing.
[...]

PR #1292, opened 2026-05-28
render: explore solid-coverage occlusion culling phase for voxel pool (#1290)

### LINKED ISSUE #1290: render: explore solid-coverage occlusion culling phase for voxel pool
**Area:** engine/prefabs/irreden/render, engine/render/src/shaders, design

**Model:** opus

**Related:** #1288 (cull-gated REBUILD_GRID_VOXELS), #1274 (reframe), Epic #936 (rotation architecture)

## Context

Frustum culling skips work for voxel sets outside the iso-frustum AABB. It does NOT skip work for voxel sets that are inside the frustum but fully occluded by solid geometry in front of them. For dense scenes (city blocks, dungeon walls, deep voxel terrain) a significant fraction of in-frustum voxels never contribute a pixel to the framebuffer because something closer wins atomicMin at every pixel they would touch.

This issue is **exploration-stage**: scope a design for solid-coverage occlusion culling that integrates with the existing voxel-pool pipeline. Decide whether to implement, not commit to implementing.

## Design questions to answer

1. **Granularity.** Per-voxel? Per-chunk? Per-voxel-set? Per-chunk is the natural unit because the pool already maintains per-chunk bounds (\`rebuildChunkBounds\`).
2. **Where to run the cull.** Three options:
   - **CPU-side pre-pass** — walk chunks front-to-back in iso depth; for each chunk, query "is any iso pixel in this chunk's projected screen AABB still unwritten or has a deeper distance value?" Needs a CPU mirror of the canvas distance texture or a hierarchical-z structure.
   - **GPU-side compute pre-pass** — write per-chunk visibility bits into an SSBO via a compute shader that samples the distance texture (or a downsampled HZB) at chunk-AABB-projected pixels. \`c_voxel_visibility_compact\` reads the bits.
   - **GPU-side fine-grained pre-pass on the active-mask SSBO** — extend the visibility-compact pass (T-287) with a per-voxel solid-coverage check, not just a per-voxel active-mask check.
3. **Hierarchical depth source.** A canonical Hi-Z buffer (mipmap chain of canvas distance min/max) would make the cull cheap and accurate. The current pipeline has no HZB. Building one is a separate stage (downsample-min compute pass after the main raster + AO).
4. **Latency.** Most occlusion culling reads last-frame's depth (HZB), accepting one-frame lag (false negatives at moving silhouettes are usually invisible). Reading current-frame depth is hard because the depth is being written by the same pass we want to cull.
5. **Lighting-volume interaction.** The lighting culling invariant (\`engine/render/CLAUDE.md\` "Lighting culling invariants") says light-occlusion-grid build iterates the full voxel pool — that grid is also LOS-dependent and may not benefit from solid-coverage culling. Confirm which downstream consumers are safe to cull and which need the full set.

## Expected payoff

- Dense indoor scenes: large fraction of voxels behind walls never contribute pixels. Cull could halve raster cost in those scenes.
- Sparse outdoor scenes: most voxels are visible; the cull is overhead with no payoff. Should be gated on a heuristic or off by default.

## Out of scope

- Implementation. This issue is design-stage. The deliverable is a \`docs/design/voxel-occlusion-culling.md\` that:
  - States the chosen granularity and pre-pass location
  - Specifies the HZB construction (if needed)
  - Lists which downstream consumers honor the cull and which iterate the full pool regardless
  - Quantifies expected payoff on a representative scene (use \`IRPerfGrid\` or a dense-indoor demo)
  - Files implementation as a follow-up sub-task (or epic if multi-PR)

## Acceptance

- \`docs/design/voxel-occlusion-culling.md\` lands as a docs-first PR per the architect role doc's docs-first pattern.
- Doc enumerates the design decisions, perf expectations, and downstream consumer matrix.
- Implementation epic / sub-task filed referencing the doc as source of truth.
- Decision to implement OR explicit decision to defer with documented trigger condition (e.g., "implement when a dense indoor scene shows >2× rasterization cost from occluded voxels").

\`[opus]\` — design exploration; needs architect-level thinking on the pipeline integration before committing to implement.

jakildev:
Claiming issue. Work in progress — design-exploration deliverable: `docs/design/voxel-occlusion-culling.md`.

Closes #1290

### DIFF
--- a/docs/design/voxel-occlusion-culling.md
+++ b/docs/design/voxel-occlusion-culling.md
@@ -1,0 +1,327 @@
+# Voxel-pool solid-coverage occlusion culling
+
+Design exploration for issue #1290. This doc scopes a third culling phase for
+the voxel-pool render path — culling voxels that are inside the frustum and
+locally exposed but **globally occluded by closer solid geometry** — and makes
+a documented implement-vs-defer call.
+
+**Recommendation: defer implementation.** Build the design now, gate the
+trigger on a measurement that does not exist yet (a dense-indoor scene whose
+raster stages dominate frame time). The full design below is ready to
+implement when that trigger fires. Rationale and the trigger condition are in
+[§ Recommendation](#recommendation).
+
+This is the architect-ratification surface per the docs-first pattern: the
+analysis is here; if the architect flips the implement-vs-defer call, the
+implementation sketch in [§ Implementation sketch](#implementation-sketch-when-the-trigger-fires)
+is the starting point.
+
+---
+
+## What "occlusion culling" means here, precisely
+
+The voxel raster already drops two large classes of voxels before they cost a
+distance write. Solid-coverage occlusion culling targets a **third, residual**
+class. Naming the three precisely is the whole game — most of the imagined
+payoff in the issue is already captured by the first two.
+
+| Cull | Granularity | What it drops | Where it runs |
+|---|---|---|---|
+| **Frustum AABB** (existing) | per pool-chunk (256 voxels) | chunks whose iso AABB is fully outside the (shadow-feeder-widened) viewport | CPU `buildChunkVisibilityMask` → chunk mask SSBO |
+| **Exposed-face mask** (landing, #1256/#1278) | per voxel × per face | faces whose neighbor cell is solid — so a voxel fully interior to a solid block emits nothing | GPU, baked into `exposedFaces` at pool build/mutate |
+| **Solid-coverage occlusion** (this doc, #1290) | per chunk or per voxel | voxels that are locally exposed *and* in-frustum but whose every projected iso pixel is already owned by closer geometry from a **different** object | new GPU pre-pass |
+
+The distinction that matters: the exposed-face mask is **intra-object** — it
+removes the inside of a solid wall, terrain volume, or filled shape, which is
+exactly the "deep voxel terrain" case the issue motivates. That payoff is
+*already booked* by #1278's visible-triplet × exposed-mask model
+([`voxel-face-rasterization.md`](voxel-face-rasterization.md)). What remains
+for #1290 is **inter-object** occlusion: object B's exposed surface hidden
+behind object A's exposed surface (a crate behind a wall, an enemy behind a
+pillar, a room behind a closed door). Whether that residual is worth a new
+pipeline stage is the question this doc answers.
+
+---
+
+## Current pipeline (grounding)
+
+The voxel-to-trixel path runs three GPU passes inside one per-canvas tick
+(`system_voxel_to_trixel.hpp`), preceded by a CPU chunk cull:
+
+1. **CPU frustum cull** — `buildChunkVisibilityMask`
+   (`engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp:44`)
+   walks per-pool-chunk iso AABBs (`C_VoxelPool::rebuildChunkBounds`,
+   `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp:244`)
+   against the iso viewport, widened by the shadow-feeder sweep
+   (`IRMath::shadowFeederIsoBounds`, `engine/math/include/irreden/ir_math.hpp:920`;
+   `kSunShadowMaxDistance = 64`). Writes a per-chunk `0/1` visibility mask
+   into the `ChunkVisibility` SSBO (binding 24).
+
+2. **GPU compact pass** — `c_voxel_visibility_compact.glsl:54`. Per voxel:
+   tests the chunk-visibility bit (binding 24), the per-voxel active-mask bit
+   (binding 8), and a per-voxel iso-bounds test, then `atomicAdd`s survivors
+   into a compacted index list (binding 25) and finalizes the indirect
+   dispatch params (binding 26) for the raster passes. **This is the natural
+   home for a per-voxel occlusion test, and the `ChunkVisibility` SSBO is the
+   natural home for a per-chunk one** — both inputs already flow into this
+   pass.
+
+3. **GPU raster stage 1** — `c_voxel_to_trixel_stage_1.glsl`. Each compacted
+   voxel projects its visible faces to iso pixels and `imageAtomicMin`s a
+   packed depth into the `R32I` distance texture (`triangleCanvasDistances`,
+   binding 1; cleared to `kTrixelDistanceMaxDistance = 65535`,
+   `component_triangle_canvas_textures.hpp`).
+
+4. **GPU raster stage 2** — `c_voxel_to_trixel_stage_2.glsl:67`. Re-projects,
+   reads the now-final distance at each pixel, and writes color + entity id
+   **only where `voxelDistance == canvasDistance`** (the voxel won the
+   atomicMin). Voxels that lost every tap write nothing here.
+
+Dispatch size is driven by the compacted count: `voxelDispatchGridForCount`
+(`system_voxel_to_trixel.hpp:37`) maps the survivor count to a
+`(min(n,1024), divCeil(n,1024))` grid, and the compact pass writes those
+counts into the indirect-dispatch buffer. **So any cull that reduces the
+compacted count directly shrinks the stage-1/stage-2 dispatch — the savings
+mechanism already exists.**
+
+### The load-bearing invariant: stage 2 is already self-occluding
+
+Stage 2 only writes where a voxel won the atomicMin. A voxel that is **fully
+occluded** — every projected tap loses to closer geometry — therefore writes
+**zero** color, zero entity id, and its losing `imageAtomicMin` calls in stage
+1 leave the distance texture unchanged. **The final `trixelDistances`,
+`triangleCanvasColors`, and `triangleCanvasEntityIds` are bit-identical
+whether or not a fully-occluded voxel was dispatched.** Occlusion culling buys
+exactly the *compute* of those losing taps — never a change in output. This is
+what makes the cull safe (see [§ Downstream-consumer matrix](#downstream-consumer-matrix)),
+and also what bounds its payoff: it removes ALU/atomic work, not pixels.
+
+---
+
+## Design space
+
+### 1. Granularity
+
+| Option | Cost | Cull rate | Verdict |
+|---|---|---|---|
+| **Per pool-chunk (256 voxels)** | one HZB test per chunk; reuses `ChunkVisibility` SSBO | only culls a chunk if its *entire* iso AABB is covered — rare unless chunks are spatially tight | **preferred starting point**; cheapest, zero new per-voxel cost, drops into the existing chunk-mask plumbing |
+| **Per voxel (in compact pass)** | one HZB sample per voxel per frame | tightest cull rate | high per-voxel cost added to the hottest compact loop; only worth it if per-chunk leaves large residual |
+| **Per spatial region (world-chunk / voxel-set)** | needs a spatial index the pool doesn't maintain | best coherence | out of scope — requires reorganizing the pool layout (separate task) |
+
+Pool "chunks" are **slices of the pool array** (256 consecutive entries), not
+spatial cells. They are coherent enough for frustum culling today because
+voxel-sets are appended contiguously and tend to be spatially local. Occlusion
+is more sensitive: a chunk is cullable only when its full iso AABB is behind
+closer geometry, so a chunk with a wide or scattered AABB rarely qualifies.
+Chunk-bounds tightness is therefore the dominant tuning lever — if per-chunk
+cull rate is poor in practice, the fix is tighter spatial chunking, not
+per-voxel testing.
+
+**Choice: per pool-chunk Hi-Z test, written into the existing
+`ChunkVisibility` SSBO**, AND-ed with the frustum bit. Per-voxel is a documented
+follow-on only if measurement shows per-chunk leaves a large exposed residual.
+
+### 2. Pre-pass location
+
+| Option | Verdict |
+|---|---|
+| **CPU pre-pass** | Rejected. Needs a CPU mirror of the distance texture — a GPU→CPU readback on the per-frame critical path, which stalls the pipeline. Non-starter. |
+| **GPU compute pre-pass → `ChunkVisibility` SSBO** | **Chosen.** A compute shader samples a Hi-Z buffer at each chunk's projected iso AABB and ANDs an occlusion bit into the chunk mask the compact pass already reads. No new readback, no new consumer wiring. |
+| **GPU per-voxel in compact pass** | Deferred follow-on (granularity §1). |
+
+### 3. Hi-Z (hierarchical depth) source
+
+There is no Hi-Z buffer today; the distance texture is a flat `R32I` image.
+A Hi-Z is a mip chain of **max** depth (so a coarse texel says "the *farthest*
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — render: explore solid-coverage occlusion culling phase for voxel pool (#1290)

**Verdict:** approve

### Summary

Clean design-exploration deliverable. The doc correctly scopes solid-coverage occlusion culling as a *third* culling phase targeting only inter-object occlusion, correctly distinguishes it from the intra-object payoff already booked by the exposed-face mask (#1278), and makes a well-reasoned "defer" call with explicit trigger conditions. The downstream-consumer safety matrix is grounded and the "stage 2 is already self-occluding" invariant is confirmed in code (`c_voxel_to_trixel_stage_2.glsl:72`: `if (voxelDistance == canvasDistance)`). The shadow-feeder Hi-Z coverage mitigation (Hi-Z must cover the shadow-feeder-widened bounds, not just the visible viewport) correctly mirrors what `buildChunkVisibilityMask` already does.

---

### Code-citation verification

All key code claims verified against the current branch:

| Claim | Verified |
|---|---|
| `ChunkVisibility` SSBO binding 24 | ✅ `c_voxel_visibility_compact.glsl:38` |
| Compact pass main at `c_voxel_visibility_compact.glsl:54` | ✅ |
| `triangleCanvasDistances` binding 1 in stage 1 | ✅ `c_voxel_to_trixel_stage_1.glsl:86` |
| Stage 2 `if (voxelDistance == canvasDistance)` | ✅ `c_voxel_to_trixel_stage_2.glsl:72` |
| Sun-shadow bake reads `trixelDistances` | ✅ `c_bake_sun_shadow_map.glsl:64` |
| `system_build_light_occlusion_grid.hpp` uses `getLiveVoxelCount` (full pool) | ✅ line 261 |
| `kSunShadowMaxDistance = 64.0f` | ✅ `sun_shadow_constants.hpp:18` |
| `C_VoxelPool::rebuildChunkBounds` at line 244 | ✅ |
| `voxelDispatchGridForCount` at `system_voxel_to_trixel.hpp:37` | ✅ |

---

### Nits

- `docs/design/voxel-occlusion-culling.md:47` — `IRMath::shadowFeederIsoBounds` is cited at `engine/math/include/irreden/ir_math.hpp:920` but the current definition is at **line 792**. Likely drifted in a prior refactor. Cosmetic-only since the function exists and behaves as described.

- `docs/design/voxel-occlusion-culling.md:43` — `buildChunkVisibilityMask` is cited at `system_voxel_to_trixel.hpp:44` but is at **line 45**. Off-by-one.

- `docs/design/voxel-occlusion-culling.md` (§ Load-bearing invariant + consumer matrix) — The doc uses two different GLSL variable names for the same R32I texture: `triangleCanvasDistances` (stage 1/2 shaders, binding 1) and `trixelDistances` (sun-shadow bake shader, binding 0). Both are bound to the same C++ resource (`C_TriangleCanvasTextures::textureTriangleDistances_`). A one-line parenthetical when first mixing the names — e.g. `trixelDistances` (same as `triangleCanvasDistances` in stage 1/2; different binding slot in the bake pass)` — would prevent an implementer from reading these as separate textures.

---

- [ ] Confirm `docs/design/voxel-face-rasterization.md` link in "See also" resolves (same directory; file exists on master).
- [ ] Confirm follow-up issue #1294 is filed (referenced as the "gated implementation" tracker).

---

Opus recheck required: The shadow-feeder Hi-Z coverage mitigation and the approximate-cull → shadow/AO hazard analysis span four render systems (compact pass, stage 1, stage 2, sun-shadow bake) with a one-frame-lag pipeline invariant. An Opus validation of the consumer-matrix safety argument — specifically whether the conservative-rounding + shadow-feeder-widened Hi-Z footprint is sufficient to preserve `trixelDistances` bit-identity for the bake — is warranted before the implementation sketch becomes a worker's starting point.

jakildev (commented):
## Opus recheck — #1290: render: explore solid-coverage occlusion culling phase for voxel pool

**Verdict:** approve (keep `fleet:has-nits`)

Sonnet approved and flagged for Opus recheck on "whether the conservative-rounding + shadow-feeder-widened Hi-Z footprint is sufficient to preserve `trixelDistances` bit-identity for the bake," noting the hazard spans four render systems with a one-frame-lag invariant. I traced the load-bearing invariant three systems deep — `c_voxel_to_trixel_stage_1.glsl`, `c_voxel_to_trixel_stage_2.glsl`, and the lighting-invariant-1 contract in `engine/render/CLAUDE.md` / `system_build_light_occlusion_grid.hpp`. The design is sound and the defer call is well-reasoned. **The doc is correct to merge as the parked design record**; two reasoning slips in the safety prose should be cleaned up before a worker lifts this as the #1294 starting point (hence keep `fleet:has-nits`, not a re-review round on a deferred doc).

---

### Sonnet asked whether the mitigation preserves bit-identity. On closer read: the *exact* cull is bit-identical (verified); the *chosen approximate* cull is not — and the doc is mostly honest about that, but over-claims in one spot

**Load-bearing invariant — CONFIRMED in code, three systems deep.**

- Stage 1 (`c_voxel_to_trixel_stage_1.glsl:97-100`): the *only* canvas write is `writeDistanceTap → imageAtomicMin(triangleCanvasDistances, …)`. A voxel that loses every tap performs atomicMin against a smaller stored value — a no-op. No unconditional color/id/distance store exists in stage 1.
- Stage 2 (`c_voxel_to_trixel_stage_2.glsl:67-80`): `writeColorTap` stores color + entity id **only** under `if (voxelDistance == canvasDistance)`. A fully-occluded voxel writes zero color, zero id.

So the doc's "§ load-bearing invariant" (an *exact* cull that drops only provably-losing voxels yields bit-identical `trixelDistances` / colors / ids) is correct, and the downstream "✅ exact" / "✅ conditional" rows follow from it. Lighting invariant 1 (`engine/render/CLAUDE.md`, confirmed compliant: grid-build iterates `getLiveVoxelCount()` over the full pool with no viewport filter) backs the matrix's "do not wire the cull into the light-occlusion grid — structurally unaffected" row. **Sonnet's "stage 2 is already self-occluding" check holds; I confirm it independently and extend it to stage 1's atomicMin.**

**Direct answer to Sonnet's question:** conservative-rounding + shadow-feeder Hi-Z is **sufficient to preserve correctness, but NOT strict bit-identity** for the bake. The mitigations eliminate the two *catastrophic* divergences — (a) coarse-footprint holes (spatial: expand-by-one-texel + skip-on-partial-coverage), and (b) dropping a shadow-relevant off-screen caster (the shadow-feeder-widened Hi-Z footprint, mirroring `buildChunkVisibilityMask`). What they do **not** eliminate is the one-frame *temporal* lag at moving silhouettes, which the doc itself accepts as a one-frame pop. So the bake reads a `trixelDistances` that is bit-identical on static content and bounded-divergent (one-frame) at moving edges. That is the correct posture for a gated, off-by-default cull — but the doc states it two different ways, and one of them is wrong:

---

### Nits (keep `fleet:has-nits`; clean up before #1294 lifts this — do not warrant a re-review round on a deferred doc)

- **`voxel-occlusion-culling.md` § "3. Hi-Z source" (~lines 145-150) — the lag-error direction is mislabeled, and it inverts the safety rationale.** The text says "Occlusion errors from lag are false *negatives* at moving silhouettes (a voxel revealed this frame that last frame's depth still showed as covered)," then two sentences later defines "A false positive (culling a visible voxel) is a hole; a false negative (keeping an occluded voxel) is just lost savings." The parenthetical case — a voxel *revealed this frame* that last frame's Hi-Z *showed as covered* — is culled by the lag (last-frame Hi-Z says occluded) even though it is now visible. By the doc's own definition that is a **false positive (a hole/pop)**, not a false negative. The dangerous lag error is precisely this false positive; the benign one is the reverse (occluder moves *in* this frame → kept voxel loses every tap anyway → correct). As written, a reader trusts "lag errors are benign lost-savings" and would see no reason for the one-frame-disable-on-camera-cut mitigation in § 4 — which exists *because* lag produces holes. Fix: relabel the moving-silhouette reveal as a false positive, and state that lag's dangerous direction is the false positive, which is what the conservative rounding + camera-cut disable target.

- **`voxel-occlusion-culling.md` § "Implementation sketch" item 4 (~lines 294-296) over-claims pixel-identity.** It tells the implementing worker the render-verify baseline must prove "pixel-identical output with the cull on vs off (it must be — fully-occluded voxels write nothing; any diff is a cull bug)." That holds only for an *exact* cull on *static* content — it directly contradicts § "the one real hazard" (~lines 182-185: "an approximate cull can drop a voxel that *would* have won a tap this frame, which changes `trixelDistances`") and the accepted one-frame AO shimmer (~lines 203-204). A worker who builds the render-verify baseline on moving content and treats the expected one-frame pop as "a cull bug" will chase a non-bug. Fix: qualify item 4 to require the pixel-identity baseline on a **static** scene (camera + geometry at rest, cull steady-state), and explicitly call the moving-silhouette one-frame divergence the *intentional* drift the demo's drift-note covers.

- **Endorse Sonnet's three nits** (the `ir_math.hpp:920→792` and `system_voxel_to_trixel.hpp:44→45` citation drifts, and the `triangleCanvasDistances`/`trixelDistances` naming parenthetical). All cosmetic; the second of mine above subsumes the naming one in spirit — the doc would read more safely if the R32I distance texture were named consistently throughout.

---

### Not blocking — design quality

The defer recommendation is well-justified: the intra-object payoff is already booked by #1278's exposed-face mask, the only scene class that would benefit (dense indoor) is unmeasurable today, and the in-flight #1278/#1288 reworks change the exact cost baseline. The trigger conditions are concrete and the implementation sketch is appropriately scoped per-PR. The shadow-feeder coupling is the right #1 risk to surface to the implementing worker. No architectural objection.

🤖 Reviewed by Claude Opus 4.7 (role-opus-reviewer, final pass)

jakildev:
Addressed nits:
- `buildChunkVisibilityMask` citation: line 44 → 45 (off-by-one)
- `shadowFeederIsoBounds` citation: line 920 → 792 (drifted in prior refactor)
- Added parenthetical in § Load-bearing invariant clarifying that `trixelDistances` and `triangleCanvasDistances` both refer to `C_TriangleCanvasTextures::textureTriangleDistances_` — different names come from different binding slots in stage 1/2 (binding 1) vs sun-shadow bake (binding 0) shaders.

Test plan verification:
- ✅ `docs/design/voxel-face-rasterization.md` exists in the same directory
- ✅ Issue #1294 is filed ("render: implement voxel-pool solid-coverage occlusion culling (gated, from #1290 design)")

— sonnet-fleet-2

jakildev (commented):
## Opus recheck review — PR #1292

**Verdict: NEEDS FIX** (confirming Sonnet's verdict; adding the three deferred deep items)

Sonnet flagged this for Opus recheck on the visible-triplet math, GPU buffer lifetimes, and dispatch grid sizing. On a full read of the new stage I confirm all three are real bugs — and the visible-triplet one is a golden-path functional failure, not a subtlety.

### Opus-only items Sonnet deferred

**1. [needs-fix] The visible triplet never reaches the shader — every face is culled.**
`engine/render/shaders/face_cull.glsl` declares `uniform uint c_visibleTriplet`, and `computeVisibleTriplet()` is defined in `ir_face_cull.cpp`, but `FaceCullStage::execute()` never calls `computeVisibleTriplet()` and never sets the uniform (no `glUniform1ui`; `execute()` only does the two `glBindBufferBase` calls for bindings 0 and 1). So `c_visibleTriplet` stays 0 → `faceMask[idx] = visible & exposed = 0 & exposed = 0` → **all faces culled, nothing rasterizes (black output)**. This is the core of the "visible-triplet math" recheck: the math function is dead code as wired. Fix: compute the triplet from the camera dir each frame and upload it (`glUniform1ui(loc, computeVisibleTriplet(camDir))`) before dispatch. The shader's `// NOTE: ... host binds it via glBindBufferBase as if it were an SSBO` comment describes a binding that doesn't exist in `execute()` — a uniform can't be fed by `glBindBufferBase`.

**2. [needs-fix] Dispatch grid under-dispatches; floor division drops the tail.**
`ir_face_cull.cpp` `execute()`: `const uint32_t groups = voxelCount / kWorkgroupSize;` (kWorkgroupSize=64). Integer floor division means the final `voxelCount % 64` voxels get no workgroup, and `voxelCount < 64` dispatches **zero** groups. The output buffer is `GL_DYNAMIC_DRAW` and is never cleared, so the un-dispatched entries keep stale `faceMask` values → garbage faces for the tail (and entirely garbage for small chunks). Fix: ceil division `(voxelCount + kWorkgroupSize - 1) / kWorkgroupSize`. But that requires (3) below or it reads OOB.

**3. [needs-fix] No shader bounds guard → OOB once dispatch is corrected.**
`face_cull.glsl` `main()` does `uint voxel = voxels[idx];` / `faceMask[idx] = ...` with no `if (idx >= voxelCount) return;`, and `voxelCount` is never passed to the shader. With the ceil-division fix from (2), the last workgroup runs invocations past `voxelCount`, reading `voxels[idx]` and writing `faceMask[idx]` out of bounds. Pass `voxelCount` as a uniform and early-return. (This is the GPU-side counterpart to Sonnet's finding #3 about the CPU face-index bounds.)

**4. [needs-fix] GPU resources leak — `FaceCullStage` has no destructor.**
The constructor allocates `m_faceMaskBuffer` (`glGenBuffers`) and `m_cullProgram` (`compileCompute`), but the class declares no `~FaceCullStage()` and never calls `glDeleteBuffers` / `glDeleteProgram`. Every `RenderPipeline::buildStages()` rebuild constructs a fresh stage and leaks the previous buffer + program. Since the class owns a raw `GLuint`, it also needs rule-of-five (at minimum a destructor + deleted copy) so an accidental copy can't double-free / alias the handle. This is the "GPU buffer lifetime across the stage boundary" item.

### Could not verify from the diff — author must confirm

- **Face-bit convention consistency.** `visible & exposed` is only meaningful if `computeVisibleTriplet()`'s bit ordering (bit0=+X, bit1=-X, … per the code) matches the neighbor-occupancy bit ordering that fills `voxel` bits 8..13 feeding `exposedMask()`. The producer of that occupancy encoding isn't in this diff (or the current tree). If the two conventions differ, the AND combines mismatched faces. Please point to where the occupancy bits are written and confirm the ordering matches.
- **Memory barrier scope.** `glMemoryBarrier(GL_SHADER_STORAGE_BARRIER_BIT)` is correct *iff* the downstream rasterize stage reads `faceMask` as an SSBO. If it consumes it as a vertex/element/indirect source, the barrier bit is wrong. The consuming stage isn't in this diff — confirm the access type.

### Agreement with Sonnet
- Sonnet #1 (exposed-mask recomputed every frame): agreed it's wasted work for static chunks, but it's a perf optimization, not correctness — lower priority than the four above. Note the recompute is currently moot anyway because (1) culls everything.
- Sonnet #3 (CPU-side face-index OOB): agreed; finding (3) above is the GPU-side sibling.

### Nits (do not gate)
- Sonnet #2 (`.xyz` swizzle over indexed access): agreed.
- Drop the review-scaffolding comments — `// NOTE: ... see review`, `// face cull stage added before rasterize for occlusion skip`, `// face mask consumed by rasterize stage via RenderContext`. These reference the task/review and violate the comment-style baseline (explain non-obvious *why*, not *what*).

Keeping `fleet:needs-fix`. The four needs-fix items plus the two verify-please items are best fixed together with Sonnet's so this lands in one more round rather than two.

Issue #1291, opened 2026-05-28
render: consolidate the per-frame shadow-feeder cull-viewport derivation inlined across 3 voxel/shape systems


jakildev:
## Context

Uncovered while implementing #1288 (REBUILD_GRID_VOXELS cull gate). The simplify reuse-detection pass flagged two cross-module duplications that were out of scope for that PR (it touches only `engine/prefabs/irreden/voxel/`, while the fix would touch `engine/prefabs/irreden/render/` systems + `engine/math/`).

## Duplication 1 — shadow-feeder cull-viewport derivation (3 copies)

The same 5-step per-frame sequence is now inlined verbatim in three systems:

```
getSunShadowsEnabled()
  → sunDir       = shadowsEnabled ? SunShadow::getFrameSunDirection() : vec3(0)
  → sweepDist    = shadowsEnabled ? SunShadow::kSunShadowMaxDistance : 0
  → getCullViewport().isoViewport(margin)
  → IRMath::shadowFeederIsoBounds(viewport, sunDir, sweepDist)
```

Call sites:
- `engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp` (`VOXEL_TO_TRIXEL_STAGE_1`, ~L309-319)
- `engine/prefabs/irreden/render/systems/system_shapes_to_trixel.hpp` (`SHAPES_TO_TRIXEL::beginTick`, ~L160)
- `engine/prefabs/irreden/voxel/systems/system_rebuild_grid_voxels.hpp` (`REBUILD_GRID_VOXELS::beginTick`, added by #1288)

A change to how the cull viewport / shadow-feeder sweep is derived currently has to be made in three places. Candidate fix: a shared `IRRender` helper, e.g. `IRRender::frameShadowFeederParams() -> {sunDir, sweepDistance}` and/or `IRRender::frameCullChunkViewport(margin) -> IsoBounds2D`, that all three systems call.

## Duplication 2 — chunk-AABB-overlap predicate (2 copies)

The 4-comparison iso-AABB overlap test is duplicated between:
- `buildChunkVisibilityMask` in `system_voxel_to_trixel.hpp` (writes a per-chunk mask word)
- `C_VoxelPool::isRangeVisible` (added by #1288; short-circuits on first hit over a sub-range)

[...]

PR #1293, opened 2026-05-28
render: cull-gate REBUILD_GRID_VOXELS, drop snapshot-compare cache

### LINKED ISSUE #1288: render: REBUILD_GRID_VOXELS — drop snapshot-compare cache, always re-rasterize on-screen entities, gate on cull
**Area:** engine/prefabs/irreden/voxel/systems/system_rebuild_grid_voxels.hpp, engine/prefabs/irreden/voxel/components/component_voxel_set.hpp

**Model:** opus

**Reframes:** #1274 (closed) — the original nit suggested dirty-transform guarding, which is the wrong frame.

## Context

`SYSTEM_REBUILD_GRID_VOXELS` today caches \`(lastRebuildWorldRotation_, lastRebuildWorldScale_, lastRebuildWorldTranslation_)\` on \`C_VoxelSetNew\` and early-returns when the live \`C_WorldTransform\` matches. The header comment argues this is "component fields rather than a dirty flag" but the practical effect is identical to a dirty flag: a side-channel of \"did this change since last frame\" used to gate per-frame work. The engine's [cpp-ecs.md "No dirty flags"](../tree/master/.claude/rules/cpp-ecs.md) rule exists precisely to keep this pattern out of the engine; the snapshot-compare cache slipped under it via a naming argument.

The right frame: **on-screen entities re-rasterize unconditionally each frame (assume dynamic); the cost we want to eliminate is doing the work for entities the camera can't see, which is a cull concern, not a dirty-flag concern.**

## Scope

1. **Drop the snapshot-compare cache.** Remove \`lastRebuildWorldRotation_\` / \`lastRebuildWorldScale_\` / \`lastRebuildWorldTranslation_\` / \`hasLastRebuildWorldTransform_\` from \`C_VoxelSetNew\`. Remove the early-return block in \`tick()\` that compares them. Remove the header comment's "rather than a dirty flag" justification.
2. **Gate on cull instead.** REBUILD_GRID_VOXELS should skip a voxel set when its pool range is not visible to the camera this frame (frustum-cull-aware). The voxel pool already maintains per-chunk bounds (\`rebuildChunkBounds\`) and visibility-compact infrastructure (T-287 active mask) — extend or query that to answer "is this voxel set's range visible?" Concretely:
   - If the voxel set's chunk(s) are all outside the iso-frustum AABB (post-cardinal-rotate, plus the shadow-feeder sweep when sun shadows are enabled — see \`engine/render/CLAUDE.md\` "Lighting culling invariants"): skip the rebuild.
   - If any chunk is visible: rebuild unconditionally.
3. **Stronger engine rule.** Update \`.claude/rules/cpp-ecs.md\` "No dirty flags" section to explicitly forbid the snapshot-compare-and-early-return pattern as a dirty-flag-in-disguise — a side-channel of "did this change since last frame" is the same anti-pattern regardless of whether the field is named \`dirty_\` or \`lastFoo_\`. Cite this issue's resolution as the worked example.

## Acceptance

- \`SYSTEM_REBUILD_GRID_VOXELS\` performs zero per-tick work for voxel sets outside the iso-frustum (cull-checked) AND unconditionally rebuilds for visible voxel sets (no per-set early-return based on transform comparison).
- \`C_VoxelSetNew\` no longer carries the snapshot fields; the dense-data path remains unchanged.
- A demo with a static large voxel scene confirms: when the camera pans so that most of the scene is off-screen, the rebuild cost drops proportionally to visible-chunk count.
- A demo with a fully-on-screen static scene confirms: rebuild runs every frame (no early-out), and per-frame cost is acceptable at the scene scale we care about.
- \`cpp-ecs.md\` "No dirty flags" rule expanded with the snapshot-compare-in-disguise prohibition + cross-reference.

## Notes

- Perf measurement worth doing before/after: take an \`IRPerfGrid\`-style measurement of REBUILD_GRID_VOXELS per-frame cost on (a) a static 25-cube scene with camera centered (all visible — both pre and post fix should be comparable), (b) the same scene with camera offset so only ~5 cubes are visible (post-fix should be ~1/5 the cost; pre-fix is 0 because the cache catches everything).
- The \`canvas_stress\` demo's 4 spinning cubes will still rebuild every frame in either case (their transforms change), so the fix is invisible there. The interesting case is mostly-static large scenes.

\`[opus]\` — touches the hottest UPDATE pipeline path + component layout + an engine rule. Coordinate with the architect on the cull-query API surface before implementing.

jakildev:
## Summary

Drops the snapshot-compare cache from `SYSTEM_REBUILD_GRID_VOXELS` / `C_VoxelSetNew` and replaces it with a frustum-cull gate, per #1288.

- **Dropped the cache.** Removed `lastRebuildWorld{Rotation,Scale,Translation}_` + `hasLastRebuildWorldTransform_` from `C_VoxelSetNew` and the per-tick early-return that compared them. On-screen GRID voxel sets now re-rasterize every frame (assume dynamic).
- **Gate on cull instead.** New `C_VoxelPool::isRangeVisible(startIdx, count, IsoBounds2D)` maps a pool slot range to its chunks and tests their iso-AABBs against a viewport. REBUILD skips a set only when all its chunks are off-screen.
- **Rule.** Expanded `.claude/rules/cpp-ecs.md` "No dirty flags" to forbid the snapshot-compare-and-early-return pattern (a dirty flag in disguise), citing #1288 as the worked example.

## Design decisions (flagged for architect review)

The issue asked to "coordinate with the architect on the cull-query API surface." The cull infrastructure already exists and is proven on the GPU side; these are the calls I made — please vet:

1. **Cull-query API = `C_VoxelPool::isRangeVisible`.** Lives on the pool because the pool owns the chunk bounds. Conservative at chunk granularity (a chunk shared by several sets reports visible if any voxel is in view → over-rebuild, never drop). Fail-safe to *visible* when chunk bounds aren't built yet or the range runs past them.
2. **Cull-state source = last-render `getCullViewport()`.** REBUILD runs in the UPDATE pipeline; the cull viewport is populated during the render event, so REBUILD reads the previous frame's snapshot — a one-frame lag that stays *consistent with the GPU cull* (same viewport source) and respects the culling-freeze debug toggle. Invisible for camera pans; a moving entity entering view can lag one frame (documented).
3. **Shadow-feeder reuse.** The cull viewport is expanded via `IRMath::shadowFeederIsoBounds` with the same margin (8) the GPU chunk-visibility mask uses, so off-screen voxels that still cast into the visible frustum keep rebuilding. Verified shadows stay intact at offset camera.
4. **Perf trade.** The static-visible cache is gone, so a fully-on-screen static scene now re-rasterizes every frame (the issue's intended behavior). The win is the off-screen case: panning so most of the scene leaves view drops rebuild cost proportionally to visible-chunk count.

`isRangeVisible` shares the chunk-AABB predicate with `buildChunkVisibilityMask`, and the shadow-feeder cull-viewport derivation is now inlined in three systems. Consolidating both is a cross-module refactor (touches the render systems + `engine/math`), out of scope here — filed as #1291.

## Test plan

- [x] `fleet-build --target IRShapeDebug` — clean (pre- and post-clang-format).
- [x] `fleet-build --target IRCanvasStress` — clean.
- [x] IRShapeDebug `--auto-screenshot`: GRID voxel cubes render correctly at origin, **camera-offset (cam=3,5)**, and non-zero-cardinal yaw shots; shadows intact — no over-culling of visible geometry.
- [x] IRCanvasStress `--auto-screenshot`: 25-cube GRID grid + 4 GRID-spin cubes + 9 detached canvases all render; DETACHED entities unaffected.
- [x] optimize: no hot-path smell (no getComponent-in-tick, no tick allocation, no fn-local static). `perf_grid_matrix` N/A — IRPerfGrid doesn't register REBUILD_GRID_VOXELS.
- [ ] Reviewer: confirm the one-frame-lag cull-state sourcing (design decision 2) is acceptable for the UPDATE→RENDER ordering.
- [ ] Linux/OpenGL smoke.

Closes #1288

### DIFF
--- a/.claude/rules/cpp-ecs.md
+++ b/.claude/rules/cpp-ecs.md
@@ -66,6 +66,29 @@ Why the dirty-flag pattern is wrong:
 
 Allowed exception: the destination resource is **strictly CPU-authored, GPU-read-only, and re-uploading the whole buffer is genuinely expensive enough** that the optimization pays off. The fog-of-war texture is the only current example — `C_CanvasFogOfWar` uploads a 256×256 RGBA8 texture (256 KiB), the GPU never writes back, and per-cell `subImage2D` would split `revealRadius`'s loop into ~hundreds of API calls. Document the exception in the component header and in the `## Live deviations` block below; new components should not introduce dirty flags.
 
+### A snapshot-compare-and-early-return is a dirty flag in disguise
+
+The rule covers more than a literal `bool dirty_`. **Caching last frame's inputs on the component and early-returning when they match is the same anti-pattern** — the stored snapshot IS the "did this change since last frame" side-channel, regardless of whether the field is named `dirty_`, `lastFoo_`, or `cachedTransform_`. A naming argument ("these are component fields, not a dirty flag") does not exempt it.
+
+```cpp
+// FORBIDDEN — snapshot-compare-and-early-return (a dirty flag in disguise).
+if (set.hasLastTransform_ &&
+    set.lastRotation_ == xform.rotation_ &&
+    set.lastScale_ == xform.scale_ &&
+    set.lastTranslation_ == xform.translation_) {
+    return;                       // "nothing changed since last frame"
+}
+doWork(...);
+set.lastRotation_ = xform.rotation_;   // re-stamp the snapshot
+set.lastScale_ = xform.scale_;
+set.lastTranslation_ = xform.translation_;
+set.hasLastTransform_ = true;
+```
+
+It carries every cost the boolean does — it bloats the component (hurting archetype iteration density), accumulates re-stamp sites that silently drift, and hides the real ownership question. **Assume per-frame work is unconditional for entities that are actually being rendered**; the only honest reason to skip is that the result isn't observable this frame. Gate on **visibility / cull** (the work product is off-screen), not on input-equality.
+
+Worked example (#1288): `SYSTEM_REBUILD_GRID_VOXELS` cached `lastRebuildWorld{Rotation,Scale,Translation}_` on `C_VoxelSetNew` and early-returned when the live `C_WorldTransform` matched. That snapshot was a dirty flag wearing component-field clothing. The fix dropped the snapshot entirely: on-screen voxel sets re-rasterize every frame, and the system instead skips sets whose pool chunks are outside the cull viewport (`C_VoxelPool::isRangeVisible`). The cull gate answers "is this work observable?" — the honest question — instead of "did the inputs change?".
+
 ### Live deviations
 
 - `engine/prefabs/irreden/render/components/component_canvas_fog_of_war.hpp` — `C_CanvasFogOfWar::dirty_` and `allUnexplored_` gate the per-frame `subImage2D` upload of the 256² fog texture. Documented exception (CPU-authored, GPU-read-only, full-texture upload). T-161 evaluated migration to per-region `subImage2D` and deferred; see [`docs/design/fog-of-war-upload-strategy.md`](../../docs/design/fog-of-war-upload-strategy.md) for the analysis, the trigger conditions for revisiting, and the mechanical Strategy C migration sketch.
--- a/engine/prefabs/irreden/render/cull_viewport_state.hpp
+++ b/engine/prefabs/irreden/render/cull_viewport_state.hpp
@@ -8,6 +8,12 @@ using namespace IRMath;
 
 namespace IRRender {
 
+// Chunk margin for the CPU-side cull gate in REBUILD_GRID_VOXELS and the GPU
+// chunk-visibility mask in VOXEL_TO_TRIXEL_STAGE_1.  Both use this value so
+// the CPU skip decision is never tighter than the GPU raster decision:
+// if the GPU still renders a chunk's voxels, the CPU must still rebuild them.
+constexpr int kCullChunkMargin = 8;
+
 // Shared cull viewport state.  Updated once per frame (during the render
 // event) so that every render system sees a consistent frozen-or-live
 // viewport without maintaining its own snapshot.
--- a/engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp
+++ b/engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp
@@ -312,9 +312,8 @@ template <> struct System<VOXEL_TO_TRIXEL_STAGE_1> {
         // runs per entity.
         const float sweepDistance = sweepDistance_;
 
-        constexpr int kChunkMargin = 8;
         const IsoBounds2D chunkVp =
-            IRMath::shadowFeederIsoBounds(cull.isoViewport(kChunkMargin), sunDir_, sweepDistance);
+            IRMath::shadowFeederIsoBounds(cull.isoViewport(kCullChunkMargin), sunDir_, sweepDistance);
         const CardinalIndex chunkCardinal = IRMath::rasterYawCardinalIndex(frameData_.rasterYaw_);
         const auto &uploadMask = buildChunkVisibilityMask(voxelPool, chunkVp, chunkCardinal);
         chunkVisBuf_->subData(0, uploadMask.size() * sizeof(std::uint32_t), uploadMask.data());
--- a/engine/prefabs/irreden/voxel/CLAUDE.md
+++ b/engine/prefabs/irreden/voxel/CLAUDE.md
@@ -74,9 +74,12 @@ for single voxels and particles.
 - `REBUILD_GRID_VOXELS` (UPDATE pipeline, T-294) — Epic C C6. Runs AFTER
   `UPDATE_VOXEL_SET_CHILDREN`. Re-rasterizes GRID-mode entities (entities
   carrying `C_RotationMode::GRID`, the default) into rotated world cells
-  whenever their `C_WorldTransform` changes. Cache lives on
-  `C_VoxelSetNew::lastRebuild*_`; ticks are O(1) when the transform is
-  unchanged. DETACHED-mode entities are skipped — they rotate through
+  from their live `C_WorldTransform`. On-screen sets re-rasterize every
+  frame (no per-set transform-comparison early-out — that was a dirty flag
+  in disguise, see `.claude/rules/cpp-ecs.md` "No dirty flags"); the only
+  skip is a frustum-cull gate (`C_VoxelPool::isRangeVisible` against the
+  shadow-feeder-expanded cull viewport), so sets whose pool chunks are all
+  off-screen pay nothing. DETACHED-mode entities are skipped — they rotate through
   the per-canvas TRS composite (`ENTITY_CANVAS_TO_FRAMEBUFFER`) and
   never touch the world voxel pool's globals. Cell aliasing
   (multiple authored voxels collapsing into one world cell after
--- a/engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp
@@ -270,6 +270,42 @@ struct C_VoxelPool {
         m_chunkBoundsDirty = true;
     }
 
+    // Cull query: true if any chunk overlapping the pool slot range
+    // [startIdx, startIdx + count) has an iso-space AABB that intersects
+    // `viewport`. Reads the chunk bounds as last computed by
+    // `rebuildChunkBounds` (driven by the render pipeline) — a caller in
+    // the UPDATE pipeline gets a one-frame-lagged but self-consistent
+    // answer (both the bounds and the cull viewport are last-render state).
+    // Chunk granularity is conservative: a chunk shared by several voxel
+    // sets reports visible if ANY of its voxels are in view, so the gate
+    // over-rebuilds rather than dropping geometry. Fail-safe: returns true
+    // when the bounds aren't available yet (pre-first-render) or the range
+    // runs past the computed chunk set, so geometry is never silently
+    // culled before bounds exist. The intersection test mirrors
+    // `buildChunkVisibilityMask` in `system_voxel_to_trixel.hpp`.
+    bool
+    isRangeVisible(std::size_t startIdx, std::size_t count, const IsoBounds2D &viewport) const {
+        if (count == 0) {
+            return false;
+        }
+        if (m_chunkBounds.empty()) {
+            return true;
+        }
+        const std::size_t firstChunk = startIdx / IRRender::kVoxelChunkSize;
+        const std::size_t lastChunk = (startIdx + count - 1) / IRRender::kVoxelChunkSize;
+        for (std::size_t c = firstChunk; c <= lastChunk; ++c) {
+            if (c >= m_chunkBounds.size()) {
+                return true;
+            }
+            const ChunkBounds &cb = m_chunkBounds[c];
+            if (cb.isoMax_.x >= viewport.min_.x && cb.isoMin_.x <= viewport.max_.x &&
+                cb.isoMax_.y >= viewport.min_.y && cb.isoMin_.y <= viewport.max_.y) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     // Active-slot mask: 1 bit per pool slot, mirroring `m_voxelColors[i].color_.alpha_ != 0`.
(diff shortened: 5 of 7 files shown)


jakildev (commented):
## Review — render: cull-gate REBUILD_GRID_VOXELS, drop snapshot-compare cache

**Verdict:** approve

### Summary

Clean, correct removal of the snapshot-compare anti-pattern from `REBUILD_GRID_VOXELS`. The cull-gate replacement answers the honest question ("is this work observable?") rather than the dirty-flag question ("did the inputs change?"). The fail-safes, one-frame-lag design, and shadow-feeder expansion are all correct.

---

### Checklist

**ECS invariants**
- No `getComponent` on the iterating entity. The canvas lookup (`getComponent<C_VoxelPool>(canvas)`) is a foreign-entity lookup on the canvas entity, amortized per frame via `lastCanvas_`/`lastPool_`. Consistent with the `beginTick` reset pattern.
- `chunkVp_` and `cullValid_` live as system-member fields on `System<REBUILD_GRID_VOXELS>`. Correct pattern.
- `isRangeVisible` is a tier-(b) self-only helper on `C_VoxelPool`: reads only `m_chunkBounds` (own field), takes `viewport` as a parameter, no cross-entity reach. Allowed per `engine/prefabs/CLAUDE.md`.
- No deferred mutations, no allocations in tick, no mid-tick structural changes.

**One-frame lag (reviewer task from PR test plan)**
Confirmed acceptable. `getCullViewport()` and the chunk bounds from `rebuildChunkBounds` are both last-render state — the system compares them self-consistently rather than mixing fresh UPDATE-frame values with stale render-frame values. The `cullValid_ = false` fallback on the first frame (before any render populates the cull viewport) correctly fails open (every set rebuilds). Worst case: an entity entering the view frustum may miss one rasterize frame. This is invisible under normal camera use and acceptable per the architectural invariant.

**AABB intersection logic**
`isRangeVisible` handles reset `ChunkBounds` correctly. An all-inactive chunk has `isoMin_ = FLT_MAX, isoMax_ = FLT_LOWEST` after `reset()`. `FLT_LOWEST >= viewport.min_.x` evaluates false, so inactive/empty chunks do not trigger early-return (not visible). Correct.

**Fail-safes**
- `count == 0` returns false: entities with no voxels correctly skip rasterization.
- `m_chunkBounds.empty()` returns true: fail-safe to visible before first-render chunk build.
- `c >= m_chunkBounds.size()` returns true: fail-safe when range extends past computed chunks.

All three are correct — geometry is never silently culled in pre-render state.

**Naming / style**
Public members use trailing `_` (`chunkVp_`, `cullValid_`). Private members use `m_` prefix (`m_chunkBounds`). `constexpr int kChunkMargin = 8` is a local compile-time constant — correct naming. No `glm::*` or `std::*` math calls added by this PR. (`glm::min`/`glm::max` in the pre-existing `ChunkBounds::expand()` is not introduced here.)

**Rule codification**
The `cpp-ecs.md` snapshot-compare section is accurate and the worked example is well-chosen. `voxel/CLAUDE.md` update is clean and correctly removes the "O(1) unchanged" claim from the REBUILD_GRID_VOXELS entry.

---

### Nits

- `engine/prefabs/irreden/voxel/components/component_voxel_set.hpp:30-31` — `lastParentPosition_` / `hasLastParentPosition_` remain on `C_VoxelSetNew` and are the analogous snapshot-compare pattern for `UPDATE_VOXEL_SET_CHILDREN`'s translate early-out. Out of scope for this PR (#1288 is scoped to REBUILD_GRID_VOXELS), but worth a follow-up to apply the same cull-gate treatment to the translate path.

---

- [ ] Confirm `REBUILD_GRID_VOXELS` is listed after `UPDATE_VOXEL_SET_CHILDREN` in the pipeline registration for any demo that uses GRID entities, or that DETACHED-only demos do not register REBUILD_GRID_VOXELS at all.
- [ ] Confirm renderings at non-zero camera yaw are unaffected (shadow-feeder expansion keeps off-screen shadow casters rebuilding).

---

Opus recheck required: crosses UPDATE/RENDER pipeline boundary — correctness depends on chunk bounds and cull viewport both being last-render state (non-obvious ordering invariant); author also explicitly requested architect vetting of the cull-query API surface choices.

jakildev:
Addressed nit: filed issue #1295 to track the cull-gate treatment for `UPDATE_VOXEL_SET_CHILDREN`'s `lastParentPosition_`/`hasLastParentPosition_` snapshot-compare pattern. The nit is explicitly out of scope for this PR; the follow-up issue is blocked on #1293 merging.

No code changes to this PR — `fleet:approved` stands. — sonnet-fleet-1

jakildev (commented):
## Opus recheck — #1288: render: cull-gate REBUILD_GRID_VOXELS, drop snapshot-compare cache

**Verdict:** approve (keep `fleet:has-nits`)

Sonnet approved and flagged for Opus recheck on the UPDATE/RENDER ordering invariant ("correctness depends on chunk bounds and cull viewport both being last-render state"), and the author requested architect vetting of the cull-query API surface. I read `system_rebuild_grid_voxels.hpp`, `component_voxel_pool.hpp`, `system_voxel_to_trixel.hpp`, `cull_viewport_state.hpp`, and `system_update_voxel_set_children.hpp` in full to trace the gate three systems deep. The PR is correct to merge; the justification is subtler than the review body states, so I'm recording the real invariant below.

---

### Sonnet flagged the ordering invariant; on closer read I confirm it — but for a different reason than the one-frame-lag analysis given

**1. Chunk-index scheme matches — CONFIRMED.**

`isRangeVisible` maps the pool slot range to chunks via `startIdx / kVoxelChunkSize` … `(startIdx + count - 1) / kVoxelChunkSize` (`component_voxel_pool.hpp:84-85`). `rebuildChunkBounds` keys `m_chunkBounds[chunk]` by `chunk = i / IRRender::kVoxelChunkSize` over the same pool-slot index `i` (`component_voxel_pool.hpp:256`). Same keying, so the gate reads the bounds that actually cover its voxels. Sonnet asserted "mirrors `buildChunkVisibilityMask`"; verified true at the index level, not just the AABB-test level.

**2. The chunk bounds are NOT one-frame-lagged for moving geometry — they can be arbitrarily stale. The gate is still correct, by a consistency argument.**

This is the part worth recording. Sonnet's "worst case: an entity entering the frustum may miss one rasterize frame" is right for the *cull viewport* (which is genuinely one render-frame old via `getCullViewport()`), but it is NOT right for the *chunk bounds*. `m_chunkBounds` is only rebuilt when `m_chunkBoundsDirty` is set (alloc/dealloc only — `component_voxel_pool.hpp:95,111,155`) or when the cardinal index changes (`rebuildChunkBounds` early-out at `:245`). The render side derives `chunkCardinal = rasterYawCardinalIndex(rasterYaw_)` (`system_voxel_to_trixel.hpp:318`) — a 90° snap, so the cardinal only flips 4× per revolution. Neither `REBUILD_GRID_VOXELS` nor `UPDATE_VOXEL_SET_CHILDREN` marks bounds dirty when it rewrites `m_voxelPositionsGlobal` (`markChunkBoundsDirty()` is **dead code** — defined at `:269`, called nowhere). So for a GRID entity that translates under a cardinal-static camera, `m_chunkBounds` stays frozen at its alloc-time footprint.

That sounds like a regression vs. the removed snapshot-compare (which always rebuilt a moving entity). It is **not**, because the gate reads the *same* `m_chunkBounds` and the *same* viewport expression — `shadowFeederIsoBounds(cull.isoViewport(kChunkMargin=8), sunDir, sweepDistance)` (diff beginTick) — that the render-side GPU chunk-visibility mask uses (`system_voxel_to_trixel.hpp:315-319`). So the CPU gate's visibility decision is **consistent-by-construction with what the GPU actually rasterizes**: whatever stale bounds the gate sees, the GPU mask sees the identical stale bounds and culls the same chunk. There is no state in which the GPU renders an entity but the gate skipped its CPU rebuild — the gate uses the same margin (8) the chunk mask uses, which is ≥ the GPU per-voxel cull margin (`kGpuMargin=4`, `:322`). The only divergence is the one-frame viewport lag, and the generous 8-chunk margin absorbs the leading edge of a fast pan. So the stale-bounds case degrades to "an entity that the GPU isn't rendering anyway doesn't get a wasted CPU rebuild" — which is exactly the point of the cull gate.

Net: the gate is correct *relative to the rendered output*, and the chunk-bounds staleness is a pre-existing property of the shared cull infrastructure, not introduced here.

**3. Shadow-feeder expansion only enlarges the gate viewport — CONFIRMED safe.**

`shadowFeederIsoBounds` grows the iso bounds along the sun-sweep direction, so any lag/error in `sunDir`/`sweepDistance` makes the gate *more* permissive (over-rebuild), never less (drop a rendered caster). Off-screen shadow casters keep rebuilding. Correct.

**4. Fail-open paths — CONFIRMED (Sonnet covered these; spot-checked).** `count == 0 → false`; `m_chunkBounds.empty() → true` (pre-first-render); `c >= size() → true`; `cullValid_ == false` on frame 0 → every set treated visible. Geometry is never silently culled before bounds exist.

---

### Cull-query API surface (author's vetting request)

`isRangeVisible` is a tier-(b) self-only `const` method on `C_VoxelPool`: reads only its own `m_chunkBounds`, takes `viewport` by parameter, no cross-entity reach. Correct placement and tier. It deliberately re-inlines the AABB test that `buildChunkVisibilityMask` also inlines — acceptable, but see the margin nit: the two are now a correctness-coupled pair.

---

### Nits (keep `fleet:has-nits`; optional, do not block merge)

- **`kChunkMargin = 8` is now a correctness-coupled literal duplicated in two files** — `system_voxel_to_trixel.hpp:315` (the GPU chunk-visibility mask) and the new `System<REBUILD_GRID_VOXELS>::beginTick`. The gate-vs-GPU consistency property above holds *only while these two stay equal*; if someone bumps one, the gate margin could fall below the GPU chunk-mask margin and the gate would start skipping rebuilds for entities the GPU still renders (one-frame-stale rotation/scale). Worth promoting to one shared named constant near the cull viewport state so they can't drift. (Nit not blocker because they're equal today.)

- **Endorse Sonnet's `lastParentPosition_` follow-up.** `C_VoxelSetNew::lastParentPosition_` / `hasLastParentPosition_` (`component_voxel_set.hpp:30-31`) are the same snapshot-compare pattern this PR is retiring, still gating `UPDATE_VOXEL_SET_CHILDREN`'s translate early-out. Out of scope for #1288; good candidate for the same cull-gate treatment in a follow-up.

- **Pre-existing, not this PR's job:** `markChunkBoundsDirty()` is dead code and nothing refreshes `m_chunkBounds` on per-frame movement. The gate is correct anyway (consistency argument above), and the GPU chunk mask carries the identical limitation today, but a future task that needs the bounds to actually track moving geometry (e.g. moving GRID units under a static camera that must occlusion-cull precisely) will have to wire dirty-on-move and weigh the per-frame `O(liveVoxelCount)` rebuild cost. Flagging so it isn't a surprise later.

---

### Rule codification

`.claude/rules/cpp-ecs.md` "snapshot-compare-and-early-return is a dirty flag in disguise" section is accurate and the #1288 worked example is well chosen. `voxel/CLAUDE.md` correctly drops the "O(1) when unchanged" claim and points at the cull gate. No notes.

🤖 Reviewed by Claude Opus 4.7 (role-opus-reviewer, final pass)

jakildev:
Addressed nit (Opus recheck): promoted `kChunkMargin = 8` from a duplicated local literal in two files to a shared `IRRender::kChunkCullMargin` constant defined in `cull_viewport_state.hpp` (already included by both systems). Both `REBUILD_GRID_VOXELS::beginTick` and `VOXEL_TO_TRIXEL_STAGE_1::tick` now read from the single named constant — they can no longer drift independently. `fleet:approved` stands. — sonnet-fleet-2

jakildev:
Addressed Opus nit: promoted `kChunkMargin = 8` to a shared `IRRender::kCullChunkMargin` constant in `cull_viewport_state.hpp`.

- Added `constexpr int kCullChunkMargin = 8` to `cull_viewport_state.hpp` with a comment explaining the correctness coupling (CPU gate must never be tighter than GPU chunk-mask).
- `system_voxel_to_trixel.hpp`: removed local literal, now uses `kCullChunkMargin` (via existing `using namespace IRRender;`).
- `system_rebuild_grid_voxels.hpp`: removed local literal, now uses `IRRender::kCullChunkMargin`.

Build and 12-shot auto-screenshot run clean. `fleet:approved` stands. — sonnet-fleet-1

PR #1297, opened 2026-05-28
#1296: fleet-state-scout + fleet-claim: live blocker resolution + multi-blocker stacking

### LINKED ISSUE #1296: fleet-state-scout + fleet-claim: live blocker resolution + multi-blocker stacking
**Context:** Two related gaps in the scout's blocker handling, surfaced 2026-05-28 while triaging stuck rotation work (#1271 / #1272). Workers correctly refused to claim these tasks even though their blockers were actually satisfied — the scout's projection rendered stale `**Blocked by:**` body text, and the stack-claim primitive couldn't reduce a multi-blocker task to a single-blocker stackable case. Together this produces "queue-starved" deadlocks even when there is execution-ready work.

## Part (a) — Live blocker resolution in the scout projection

**Symptom:** [`scripts/fleet/fleet-state-scout`](scripts/fleet/fleet-state-scout) renders the projection's `blocked_by` field by echoing the literal text from the issue body's `**Blocked by:** #N1 (rationale), #N2 (rationale)` line. When a blocker referenced as `#N1` is actually CLOSED (or its work has shipped via a merged PR with `Closes #N1`), the projection still shows it as a live blocker. Workers reading the projection see `blocked_by != "(none)"` and skip the task — even though `fleet-claim check_blockers` would happily allow the claim at gate time because it does a live `gh issue view --json state`.

**Repro (today's state):**
- #1271's body says `**Blocked by:** #1257 (PR #1262 must merge — ...)`
- #1257 is CLOSED (PR #1262 merged at 02:39:39Z)
- queue-list still shows #1271 as Blocked
- A worker iteration sees `blocked_by` non-empty and skips

**Fix shape:**

1. In the scout's `fetch_task_queue` / projection-build path, AFTER parsing the body's `**Blocked by:**` line, run a per-reference resolution pass:

   For each `#<N>` reference (with parenthetical rationale):
   - `gh issue view <N> --repo <repo> --json state,closedAt`. If `state == "CLOSED"`, mark satisfied.
   - If still open: scan the issue's body for any open PR linked via the standard `claude/<N>-*` head-branch convention (already used in `enrich_stackable_blocker_prs`). If that PR exists AND has state `MERGED`, also mark satisfied — covers the "issue stays open as a tracker but the addressing PR shipped" case (e.g. #1271 / PR #1282).

2. Render `blocked_by` as the join of UNRESOLVED references only. If all references resolve, render `(none)`.

3. Add a strikethrough or `[satisfied]` annotation for resolved references in a new `blocked_by_resolved` projection field, so `fleet-queue-list` and other UIs can show what was satisfied without losing audit trail. Optional but useful for debugging.

4. Cache aggressively — blocker resolution is a hot path on every scout tick × every queued issue × every blocker reference. Cache per `<repo>#<N>` keyed on `(state, mergedPRsha)`, invalidate when the underlying state file's `updatedAt` changes (the scout already has a per-issue updatedAt cache for the `~/.fleet/state/issues/` sidecar — piggyback on it).

5. The single-blocker shortcut: if the body has exactly one `#N` reference and it resolves CLOSED, the task transitions Available without any other intervention. Workers pick it up.

**Effect:** #1271 and #1272-like cases auto-recover. The scout no longer needs human cleanup of stale `**Blocked by:**` body text after the upstream merges.

## Part (b) — Multi-blocker stack-claim resolution

**Symptom:** [`scripts/fleet/fleet-claim`](scripts/fleet/fleet-claim) `find-stackable-blockers` (line ~742, originally added in T-381) only returns a stackable PR when there is **exactly one** `#N` blocker and **exactly one** matching open `claude/<N>-*` PR. Tasks with multiple blockers fall through with empty stdout, never get the `stackable_blocker_pr` enrichment, and the worker can't claim with `--stackable-on`.

This is correct conservative behavior in isolation — a multi-blocker task genuinely depends on multiple in-flight PRs and shouldn't single-stack on one of them. But with **(a)** in place, "multi-blocker" reduces to "unresolved blocker count" rather than "raw body reference count". A task whose body says `**Blocked by:** #1257, #1271` where #1257 is closed actually has ONE effective blocker (#1271) and should qualify for single-stack on #1271's PR.

**Fix shape:**

1. `cmd_find_stackable_blockers` and `check_blockers` both call the same resolution pass from (a) — share the implementation. Each `#N` reference goes through `live-blocker-resolve <repo> <N>` which returns one of `{satisfied, unsatisfied-issue-only, unsatisfied-with-pr <pr-url>}`.

2. `find-stackable-blockers` filters out satisfied references first. If the remaining set is exactly one entry of type `unsatisfied-with-pr`, return that PR's URL+branch — same contract as today, just on the effective set.

3. If the remaining set has size > 1, return empty (still conservative — no multi-parent stacking). The escape hatch for genuine multi-blocker tasks remains: human waits for at least one to merge, then the task drops to single-blocker and auto-stackable.

4. Edge case: blocker body reference like `#1257 (PR #1262 must merge)` — the rationale parenthetical names the PR. The resolver should NOT need to parse that; the issue's state + the `Closes #1257`-bearing merged PR is the authoritative signal. The rationale is for humans only.

**Effect:** stack-claim flow exercises automatically on the most common shape — single chain with one upstream still open. Combined with (a), tasks like #1272 (originally "#1257 + #1271") would have:
- #1257 → CLOSED → satisfied
- #1271 → issue open, PR #1282 MERGED → satisfied
- Effective unresolved: 0 → Available, NOT stack-claim needed

For a hypothetical task blocked on `#X (open with PR #Y), #Z (closed)`: effective unresolved = 1 (`#X`), stackable on PR #Y. The worker claims `--stackable-on <PR #Y>` and the branch bases on `claude/X-...` instead of master.

## Validation

1. **Unit test (a):** fixture issue body `**Blocked by:** #N1, #N2` where #N1 is closed and #N2 is open. Mock `gh issue view` to return appropriate states. Assert `blocked_by` renders as `#N2 (...)` only.

2. **Unit test (a, PR-merged variant):** fixture issue body `**Blocked by:** #N1` where #N1 is open but a `claude/N1-*` PR is MERGED. Assert resolved.

3. **Unit test (a, all-satisfied):** fixture issue body `**Blocked by:** #N1, #N2` where both closed. Assert `blocked_by` renders `(none)` and the task projects as Available.

4. **Unit test (b):** fixture body with two blockers, one satisfied and one with open PR. Assert `find-stackable-blockers` returns the open-PR line (one tab-separated row).

5. **Unit test (b, multi-still-blocking):** fixture body with two open blockers, both with open PRs. Assert `find-stackable-blockers` returns empty (no auto-stack on multi-parent).

6. **Integration:** restore the live #1271 / #1272 scenario in a sandbox repo: create stub issues + PRs matching the body / state pattern, run the scout against the sandbox, verify projection renders #1271 / #1272 as Available without any body edits.

7. **Caching correctness:** modify a fixture issue's state mid-test; verify the next scout tick picks up the change (cache invalidation on `updatedAt`).

## Files touched

- [`scripts/fleet/fleet-state-scout`](scripts/fleet/fleet-state-scout) — `fetch_task_queue`, projection-build, new resolver helper.
- [`scripts/fleet/fleet-claim`](scripts/fleet/fleet-claim) — `cmd_find_stackable_blockers`, `check_blockers`. Share a single resolver implementation (extract into `_live_blocker_resolve` or similar).
- [`scripts/fleet/tests/`](scripts/fleet/tests/) — new fixtures + assertions per the validation plan.

## Out of scope

- **Closing the open tracker-issue automatically** (#1271-style) — keep that as an explicit human signal. Just make sure tracker issues don't gum up downstream blockers in the projection.
- **Cross-repo blockers** — engine task blocked on a game-repo issue. The current parser doesn't support this and neither will the resolver in this PR. File separately if needed.
- **Stack-claim on the deepest still-open ancestor when multiple blockers are open** — a real graph-aware multi-parent strategy. Conservative single-stack remains the contract; this issue does NOT introduce multi-parent stacking. Document the boundary clearly in the changeset.

**Model:** sonnet
**Area:** scripts/fleet
**Blocked by:** (none)

jakildev:
## Summary

- `fleet-state-scout`: add `resolve_blocked_by()` — runs after `collect_state()`, resolves each `#N` blocker ref against in-memory `closed_fleet_queued` and `recent_merged_prs`, falls back to live `gh issue view` with a per-tick cache; updates `blocked_by` to bare `#NNN` for one unresolved, `(none)` for zero
- `fleet-claim`: `cmd_find_stackable_blockers` now resolves all `#N` refs live (one merged-PR heads fetch + per-ref gh issue view with cache), filters to the effective unresolved set, proceeds only when exactly one ref remains
- `scripts/fleet/tests/`: 14 new Python unit tests + 4 new bash tests covering all spec scenarios from the issue

## Test plan
- [x] `python3 scripts/fleet/tests/test_live_blocker_resolution.py` — 14 tests, all pass
- [x] `bash scripts/fleet/tests/test_fleet_claim_stackable_live_resolve.sh` — 5 assertions, all pass
- [x] All 37 prior fleet script tests remain green

Closes #1296

### DIFF
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -730,13 +730,18 @@ cmd_claim_base() {
 }
 
 cmd_find_stackable_blockers() {
-    # If the task is single-#N-blocked AND that blocker has exactly one
-    # open PR (matched via the head-branch `claude/<N>-*` or a body
-    # `Closes #N` reference), print one tab-separated line:
+    # If the task's **Blocked by:** field resolves to exactly one unresolved
+    # blocker with exactly one matching open PR (head-branch `claude/<N>-*`
+    # or body `Closes #N`), print one tab-separated line:
     #   <pr-url><TAB><headRefName><TAB><author-login><TAB><number>
-    # Otherwise print nothing (and exit 0). Multi-blocker, PR-URL-only,
-    # and zero/multiple-match cases all map to the empty output — callers
-    # treat the result as a strictly optional hint.
+    # Otherwise print nothing (and exit 0).
+    #
+    # Multi-blocker bodies are handled by resolving each #N reference live:
+    # references whose issue is CLOSED/MERGED, or whose claude/<N>-* PR is
+    # already merged, are filtered out first. If exactly one unresolved ref
+    # remains, it qualifies for single-stack (same contract as before, on
+    # the effective unresolved set rather than the raw body count).
+    # Two or more unresolved blockers → empty output (conservative).
     #
     # Usage: fleet-claim find-stackable-blockers <issue-number>
     local issue_num
@@ -755,10 +760,18 @@ cmd_find_stackable_blockers() {
 
     [[ -z "$body_b64" ]] && return 0
 
+    local repo
+    repo=$(repo_from_ns)
+    [[ -z "$repo" ]] && return 0
+    command -v gh >/dev/null 2>&1 || return 0
+
     local blocker
-    blocker=$(BODY_B64="$body_b64" python3 <<'PYEOF'
-import base64, os, re, sys
+    blocker=$(BODY_B64="$body_b64" REPO="$repo" python3 <<'PYEOF'
+import base64, json, os, re, subprocess, sys
+
 body = base64.b64decode(os.environ["BODY_B64"]).decode("utf-8", errors="replace")
+repo = os.environ.get("REPO", "")
+
 field_re = re.compile(r"^\s*-?\s*\*\*Blocked by:\*\*\s*(.+?)\s*$", re.IGNORECASE)
 value = None
 for line in body.splitlines():
@@ -768,13 +781,54 @@ for line in body.splitlines():
         break
 if value is None or value.lower().startswith("(none"):
     sys.exit(0)
-# v1: single-blocker only (matches the scout's stackable_blocker_pr filter
-# in fleet-state-scout). Multi-blocker stacking opens cherry-pick /
-# merge questions the merger has no machinery for.
+
 issue_refs = re.findall(r"#(\d+)", value)
 pr_urls = re.findall(r"https://github\.com/[^\s,)]+/pull/\d+", value)
-if len(issue_refs) == 1 and not pr_urls:
-    print(issue_refs[0])
+if not issue_refs or pr_urls:
+    sys.exit(0)
+
+# Fetch merged PR heads once for the "tracker-issue open, claude/<N>-* PR merged" case.
+merged_heads = []
+if repo:
+    try:
+        r = subprocess.run(
+            ["gh", "pr", "list", "--repo", repo, "--state", "merged",
+             "--limit", "30", "--json", "headRefName",
+             "--jq", "[.[].headRefName]"],
+            capture_output=True, text=True, timeout=15,
+        )
+        if r.returncode == 0:
+            merged_heads = json.loads(r.stdout or "[]")
+    except Exception:
+        pass
+
+_resolve_cache = {}
+
+def is_satisfied(ref):
+    if ref in _resolve_cache:
+        return _resolve_cache[ref]
+    prefix = f"claude/{ref}-"
+    if any(h.startswith(prefix) for h in merged_heads):
+        _resolve_cache[ref] = True
+        return True
+    if not repo:
+        _resolve_cache[ref] = False
+        return False
+    try:
+        r = subprocess.run(
+            ["gh", "issue", "view", ref, "--repo", repo,
+             "--json", "state", "--jq", ".state"],
+            capture_output=True, text=True, timeout=10,
+        )
+        result = r.stdout.strip() in ("CLOSED", "MERGED")
+    except (subprocess.SubprocessError, OSError):
+        result = False
+    _resolve_cache[ref] = result
+    return result
+
+unresolved = [r for r in issue_refs if not is_satisfied(r)]
+if len(unresolved) == 1:
+    print(unresolved[0])
 PYEOF
 )
     [[ -z "$blocker" ]] && return 0
@@ -782,10 +836,6 @@ PYEOF
     # Find an open PR whose head-branch starts with claude/<N>- OR whose
     # body contains "Closes #N" (case-insensitive). Exactly one match
     # qualifies; zero or multiple → empty output.
-    local repo
-    repo=$(repo_from_ns)
-    [[ -z "$repo" ]] && return 0
-    command -v gh >/dev/null 2>&1 || return 0
 
     BLOCKER="$blocker" REPO="$repo" python3 <<'PYEOF'
 import json, os, re, subprocess, sys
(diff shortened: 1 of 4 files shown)


jakildev (commented):
## Review — #1296: fleet-state-scout + fleet-claim: live blocker resolution + multi-blocker stacking

**Verdict:** approve

### Summary

Correct and well-scoped fix for both parts of the blocker-resolution gap. The two-phase approach (in-memory `closed_fleet_queued` + `recent_merged_prs` first, then live `gh issue view` fallback with per-tick cache) minimizes API calls for the common case while still catching refs outside the in-memory window.

**Part (a) — fleet-state-scout `resolve_blocked_by()`:** Runs correctly between `collect_state()` and `enrich_stackable_blocker_prs()` in `tick_once()`. The ordering is load-bearing: `SINGLE_TASK_BLOCKER_RE` in `enrich_stackable_blocker_prs` expects `blocked_by = "#NNN"` (single bare ref), and `resolve_blocked_by` is the step that reduces multi-ref bodies to that form. The `per_tick_cache` is a local dict, so no cross-tick state leakage. The `startswith(f"claude/{ref}-")` prefix check is correct — the test `test_merged_pr_prefix_discrimination` specifically covers the `claude/1011-` vs `claude/101-` false-match risk and passes.

**Part (b) — fleet-claim `cmd_find_stackable_blockers`:** Moving `repo=$(repo_from_ns)` and the `gh` guard above the Python heredoc is a clean improvement — the guard now short-circuits before spinning up a subprocess. The `--limit 30` merged-PR fetch covers the common case; for refs outside that window the code correctly falls back to `gh issue view` (which checks issue state directly), so the only unhandled edge case is "issue still open + PR merged >30 PRs ago" — negligibly rare in practice and fails safe (leaves ref as unresolved rather than incorrectly resolving it).

**Tests:** 14 Python unit tests + 4 bash tests with stub `gh`/`git`. The test count matches the PR body (11 + 3 = 14 Python methods; bash has 5 assertions across 4 test cases). The integration test `test_multi_blocker_reduces_to_stackable` is the key end-to-end case: resolve → enrich → `stackable_blocker_pr` written. Well-covered.

---

### Nits

- `scripts/fleet/fleet-state-scout:750` — the block header reads `# Live blocker resolution (T-1296)`. Since the fleet switched from T-NNN to issue numbers, this should be `#1296` (matches the issue-number convention used everywhere in the queue since the T-NNN migration landed).

- `scripts/fleet/fleet-state-scout:780` — the comment `# free-text or PR-URL form — leave unchanged` is slightly imprecise. `_BLOCKER_REF_RE` requires a `#` before digits, so it never matches bare numbers inside `https://...` URLs. The `refs` list is empty for both free-text and PR-URL forms, so the comment is describing the effect correctly but not the mechanism. Could read: `# no #N refs in this line (free-text, PR-URL, or empty) — leave unchanged`.

---

- [ ] Run `python3 scripts/fleet/tests/test_live_blocker_resolution.py` on the PR branch to verify all 14 tests pass on macOS.
- [ ] Run `bash scripts/fleet/tests/test_fleet_claim_stackable_live_resolve.sh` to verify 5/5 assertions pass.
- [ ] Confirm all 37 prior fleet tests still green (`python3 scripts/fleet/tests/test_fleet_claim.py` or equivalent suite runner).

---

Opus recheck not required.

PR #1277, opened 2026-05-28
docs: voxel-face rasterization design + architect design-doc discipline

jakildev:
## Summary

Docs-first PR establishing the architecture that supersedes #1275's design-blocked escalation, plus a process fix so engine feature docs stay the design source of truth.

**1. `docs/design/voxel-face-rasterization.md`** (new) — source of truth for which faces a voxel emits. Model: **visible-face triplet × exposed-face mask**.
- The three camera-facing faces are a per-frame function of the camera quaternion (cardinal + residual + pitch).
- The voxel's exposed faces (neighbor empty/absent) are a camera-independent mask set at pool build/mutate time.
- A voxel emits a face **iff** it is both camera-visible and exposed.
- Six faces are six distinct enums, NOT three axes with a ± sign — that distinction is the fix for #1256.
- Interior voxels emit nothing (throughput win); only the exterior copy of a visible face is emitted, so the `atomicMin` depth-ordering ambiguity that broke non-zero cardinals disappears by construction.
- Generalizes to per-entity SO(3) (#1272) and camera pitch (PR #1265) with no special cases.

**2. `engine/render/CLAUDE.md`** — new "Voxel face rasterization" section cross-referencing the design doc, before the SDF-parity section.

**3. `.claude/commands/role-opus-architect.md`** — the design-blocked resolution flow now splits decisions into:
- **Task-local** → plan file only (transient, consumed when the task completes)
- **Engine-level architecture** → durable `docs/design/<feature>.md` as source of truth, cross-referenced from module CLAUDE.md

Adds the **docs-first-PR pattern** for redesigns that supersede a worker's in-flight approach — the model gets a reviewable home that survives even if the original PR is abandoned. This is the fix for the failure mode where architectural decisions evaporate into ephemeral PR comments + transient plan files, and a future worker re-derives or contradicts them.

## Context

#1275 (partial fix for #1256 stripe artifact) escalated `fleet:design-blocked` with four candidate options. This PR captures **Option 5** (the synthesis that subsumes all four) as a durable design doc rather than burying it in a PR comment. Follow-up: the redesign issue (#1273) consumes this doc; #1275's prep work (`cardinalLowerCornerShift` + AO/lighting face-label fix) stays in.

## Test plan
- [x] Docs-only + role-doc change — no build impact
- [x] Design doc matches `iso-depth-axis-invariant.md` conventions (states model, why, consumers, migration status, what-to-verify)
[...]

Issue #1272, opened 2026-05-28
render: per-entity SO(3) rotation on main canvas (no detached canvas needed)


jakildev:
> **Tracking epic** — decomposed by the architect session (see plan comment below). Original scope retained beneath the checklist for reference.

## Sub-issues (stacked)
- [ ] #1299 — PR-A: per-entity transform plumbing + per-entity visible triplet *(unblocked, queued)*
- [ ] #1300 — PR-B: per-entity residual face deformation *(stacked on #1299)*
- [ ] #1301 — PR-C: IRRotationCompare side-by-side demo + render-verify baselines *(stacked on #1300)*

---

**Part of epic:** #936 (Rotation architecture)
**Related:** #1257 / PR #1262 (camera Z-yaw cardinal+residual), #1258 (camera Z-yaw only invariant), T-295 (detached canvas SO(3)), #1259 (canvas_stress spinning entities), #1271 (continuous-yaw demo)

## Problem

Today, an entity that needs to rotate freely (falling enemy ragdoll, tumbling projectile, NPC turning to face the player) has two options:

1. **Stay on the main canvas, no rotation** — voxels are placed in the pool at the entity's fixed orientation. Translation works; rotation doesn't.
2. **Get its own detached canvas** (T-295 / canvas_stress path) — full SO(3), but pays for an extra render target + compute dispatch per rotating entity.

Option 2 is wasteful for scenes with many medium-detail rotating entities (a horde of enemies, a debris field, a flock of birds). Each entity gets its own mini-render even though they share the same world lighting and could share the main canvas's depth buffer.

## What's actually needed

Per-entity SO(3) rotation **on the main canvas**, using shared canvas textures and shared lighting. The iso-depth-axis invariant (#1258) is preserved because that invariant constrains the CAMERA, not entity positions — entity rotation transforms world positions, and iso depth is \`dot(worldPos, (1,1,1))\` which works on any world position.

## Architecture

Two pieces change in the main-canvas voxel-to-trixel path:

### 1. Per-voxel entity-transform lookup

Two viable shapes; pick one based on perf measurement:

**A — transform indirection (preferred):**
- Each voxel in \`C_VoxelPool\` carries an \`entityTransformIndex\` (uint32)
- New per-frame \`EntityTransformBuffer\` SSBO holds each rotating entity's SO(3) + translation
- Compute shader looks up the transform per voxel, computes world position on the fly
- Static voxels in the pool, no per-frame upload of voxel positions
[...]

Issue #605, opened 2026-05-10
[Epic] Entity editor Phase 2 — Hierarchies & skeletal voxels


jakildev:
## Phase summary

Authored skeletal voxels: rig a voxel entity with a joint hierarchy, paint
per-voxel `bone_id`, pose joints with forward kinematics, and see the voxels
deform live — all in the voxel editor, persisted through `.rig`.

> **Re-planned 2026-06 (opus-architect).** This epic was filed in May against
> a foundation that has since landed; the original sub-task list described
> deprecated machinery (`C_JointHierarchy`, a separate joint-matrix SSBO).
> The current plan and child tickets are below. Full plan:
> [`.fleet/plans/issue-605.md`](https://github.com/jakildev/IrredenEngine/blob/master/.fleet/plans/issue-605.md).

## What changed vs. the original filing

- **Joints are entities** (`C_Skeleton.joints_` + `C_Joint` + `CHILD_OF`,
  landed #737) — not SoA entries in `C_JointHierarchy` (now a deprecation
  shim).
- **FK is free** — `SYSTEM_PROPAGATE_TRANSFORM` (#734) composes the parent
  chain; rotating a joint's local transform deforms descendants.
- **Skinning unifies onto the #1396 GPU voxel-position prepass** — rigid
  skinning (one bone per voxel, exactly what a `uint8 bone_id` expresses) is
  the existing `transforms[slot] * localPos` with the per-voxel slot pointing
  at a *bone* instead of an *entity*. No new shader, no stage-1 change; the
  separate binding-21 joint buffer is retired for the voxel path.

## Child tickets

| # | Title | Model | Blocked by |
|---|---|---|---|
| #1602 | bind-pose on C_Skeleton + skin-matrix helper | opus | — |
| #1603 | SYSTEM_UPDATE_JOINT_MATRICES → binding-18 buffer | opus | #1602 |
| #1605 | per-voxel skinning via bone→slot (binding-17 seed) | opus | #1603 |
| #1606 | retire binding-21 JointTransformBuffer (voxel path) | sonnet | #1605 |
| #1604 | editor: joint authoring (spawn/parent/gizmo) | opus | #1602 |
| #1607 | editor: skeleton tree panel | sonnet | #1604 |
| #1608 | editor: bone_id painting | sonnet | #1604 |
| #1610 | editor: FK pose editing (live deform + bind capture) | opus | #1605, #1604 |
[...]

Issue #1271, opened 2026-05-28
demo: continuous world-camera Z-yaw rotation — validate cardinal/residual rebracket has no glitches


jakildev:
**Part of epic:** #936 (Rotation architecture)
**Blocked by:** #1218 (face normals not rotated in lighting/shadow shaders → voxel faces go fully black at non-zero residual yaw), #1256 (trixel checkerboard/lattice artifact on faces at non-zero cardinal yaw). Both surface during continuous rotation and are clearly visible at high zoom; the demo cannot be validated glitch-free until they land. (Original blocker #1257 / PR #1262 has merged — the n=1 cap is lifted, but validation against the spinning demo revealed the remaining lighting + raster glitches above.)
**Related:** #1261 (rotation-variant static shots), #1257 (inter-cardinal deformation, PR #1262), #1259 (detached-canvas spinning entities)

## Context

The engine's world-canvas Z-yaw rotation uses a cardinal/residual split: `visualYaw = cardinalYaw + residualYaw` where `cardinalYaw ∈ {0°, 90°, 180°, 270°}` and `residualYaw ∈ [-π/4, π/4]`. The cardinal rotation drives the iso projection itself; the residual is absorbed by the per-face `faceDeformationMatrix` uploaded as `faceDeform[3]` to the voxel-to-trixel compute shader.

After PR #1262 lands (`maxN=2` on world canvas), the deformation should produce visible smooth rotation between cardinals. **Static screenshots at fixed yaws (#1261) cannot prove this.** The cardinal/residual rebracket happens every 45° (e.g. at 45°, `cardinalYaw` flips `0°→90°` and `residualYaw` flips `+π/4 → -π/4` in one frame). If the math is exactly correct, the rebracket is visually continuous. If there's any face-index remap bug, parity drift, or cull-edge desync, the rebracket produces a one-frame glitch — invisible to static shots but very visible to a human watching continuous rotation.

## Scope

Add a `--spin-yaw <deg_per_sec>` flag (default 30°/sec when given without value) to `IRShapeDebug` (or a new tiny `IRYawSpin` demo). When set, the demo programmatically rotates `IRRender::setCameraYaw(...)` each frame, running indefinitely (until user closes) OR exiting after one full rotation when paired with `--auto-screenshot`.

Visual eyeball test for:

1. Smooth visual rotation between cardinals (no \"stuck\" feeling at intermediate angles)
[...]

PR #1298, opened 2026-05-28
#1272: render: per-entity SO(3) on main canvas (design-blocked)

jakildev:
## Summary

Per-entity SO(3) rotation on the main canvas (#1272) is **design-blocked** on a cross-feature architecture decision the worker lacks authority to make. This PR records that state in the design doc and routes the decision to the architect via `fleet:design-blocked`; see the `## NEEDS-DESIGN` comment below for the specific call.

- `docs/design/voxel-face-rasterization.md` — the #1272 status entry now notes the open per-voxel→transform-slot indirection-mechanism decision and its overlap with the reserved skeletal `bone_id_` path (#605).

## Why this is blocked, not in-flight

The merged design doc (PR #1277) resolves the **raster/deformation model** (visible-triplet × exposed-mask, per-entity `FaceRenderContext` indexing). It does **not** pin **how a voxel finds its entity's transform slot**. That per-voxel→transform-slot indirection is structurally identical to the reserved skeletal `bone_id_` path: both are a per-voxel index into a per-frame transform SSBO. `C_Voxel.bone_id_` (offset 6) is already reserved for #605 (open epic). Choosing whether #1272 and #605 share one indirection convention or run two parallel ones affects both features — it is an architect call, and the plan on #1272 explicitly says *"do not implement until the architect rules on the `bone_id_` reconciliation."*

Picking a mechanism unilaterally risks a convention that diverges from #605 and would later have to be unwound, so this escalates instead.

## Test plan

- [ ] N/A — docs-only diff. No code change to verify.
- [ ] Architect resolves the indirection-mechanism decision (see NEEDS-DESIGN comment), then re-arms #1272 for implementation via `fleet:design-unblocked`.

### DIFF
--- a/docs/design/voxel-face-rasterization.md
+++ b/docs/design/voxel-face-rasterization.md
@@ -145,7 +145,13 @@ exposed-face gating are new.
   changes what comes next: replace the four options in that PR's NEEDS-DESIGN
   comment with this model.
 - **#1272** — per-entity SO(3) on the main canvas; consumes this model with a
-  per-entity visible triplet.
+  per-entity visible triplet. **Blocked pending an architect call on the
+  per-voxel→transform-slot indirection mechanism**, which overlaps the reserved
+  skeletal `bone_id_` path (#605): both are a per-voxel index into a per-frame
+  transform SSBO, so whether they share one indirection convention or run two
+  parallel ones is a cross-feature decision. The model above fixes face
+  *selection*; it does not pin *how a voxel finds its entity's transform slot*.
+  See the `fleet:design-blocked` PR on #1272 for the candidate mechanisms.
 - **#1265** (merged) — camera pitch via quaternion; the visible-triplet resolver
   reads that quaternion, so pitch shifts the triplet for free on detached
   canvases.


jakildev:
## NEEDS-DESIGN

Picked up #1272 as the only unblocked `[opus]` engine task. Before writing any code I verified, at the source level, that the task's plan is correct that it is **architect-blocked** — and the block is real and current, not stale prose.

### What I confirmed in the code (contradicts "queue-ready")

- The merged design (`docs/design/voxel-face-rasterization.md`, PR #1277) resolves the **raster/deformation model** but is silent on **how a voxel finds its entity's transform slot**.
- `C_Voxel.bone_id_` is at offset 6 (`component_voxel.hpp:68`), explicitly reserved for the skeletal **#605** path — a per-voxel index into a per-frame joint-matrix SSBO. #605 is an **open** epic.
- #1272's per-voxel→transform-slot indirection is structurally identical: a per-voxel index into a per-frame transform SSBO.
- Even Phase 0 (grow `GPUEntityTransform`, `ir_render_types.hpp:191`) touches the open per-entity context **storage/sizing** decision.

So the indirection-mechanism choice is a **cross-feature decision** (#1272 ⊗ #605), not a worker-local one. The #1272 plan says verbatim: *"do not implement until the architect rules on the `bone_id_` reconciliation."* Three prior planning passes all deliberately kept `fleet:needs-plan` on for the interactive split; its recent removal (which made this surface as queue-ready) looks premature.

### The decision I can't make

**Should #1272's per-voxel→transform-slot indirection and #605's per-voxel→bone-matrix indirection share one mechanism, or run as two parallel conventions?** Candidate mechanisms (from the #1272 thread):

- **(a)** Burn `C_Voxel.reserved_` (offset 8, currently unused) for a #1272-specific `entityTransformIndex`. Simple, but a #1272-only field — risks a second parallel convention next to `bone_id_`.
- **(b)** Derive the slot from `GPUEntityTransform.poolOffset`/`voxelCount` ranges (no per-voxel field). No `C_Voxel` growth, but a per-voxel range lookup.
- **(c)** A single generalized per-voxel **transform-slot** index that either an entity transform *or* a bone matrix can populate — unifies with #605, but is the largest up-front design commitment.

Recommendation if forced: lean toward a unified mechanism (c) so #1272 and #605 don't fork the convention — but this is exactly the call that should be made with #605 in view, which is why I'm escalating rather than deciding.

### Secondary decisions the plan also flags (for the same session)

- **Per-entity context storage/sizing** — `faceDeform[3]` is a small per-camera UBO today; per-entity it becomes per-entity × per-face. UBO slot vs SSBO, max concurrent rotating-entity count, binding cost.
- **T-295 coexistence vs replacement** — detached-canvas SO(3) stays for high-detail (opt-in by detail tier) vs main-canvas SO(3) replaces it for medium detail.
- **Cardinal rebracket × per-voxel deformation** — highest-risk; composition order/sign must be validated against the detached half via #1271 `--spin-yaw`, not by inspection.

Keeping `fleet:wip` + adding `fleet:design-blocked`; the `fleet-claim` lock + this open PR hold #1272 reserved for resumption. Once the architect rules and re-arms via `fleet:design-unblocked`, any worker can lift the decomposition (Phase 0/1/2/3) from the #1272 plan. — opus-worker-2

jakildev:
This #1272 was converted into an epic I am pretty sure. This can probably merge updating that, or just close, since there are other tasks handling this

jakildev:
Per the architect direction above (2026-05-28 17:54 — #1272 was converted to a `fleet:epic`; *"This can probably merge updating that, or just close, since there are other tasks handling this"*): **closing** this design-block shell.

The escalation already did its job — it triggered the #1272 epic conversion, and the actual work now lives in the epic's child tasks (#1299 PR-A foundation, #1301 compare demo, …). The only diff here is a 7-line note in `voxel-face-rasterization.md` marking #1272 design-blocked and pointing back to *this* PR for the candidate indirection mechanisms. That note is now stale (the block is resolved via the epic decomposition) and self-referential, so it shouldn't land.

The candidate mechanisms (a/b/c) remain preserved in this PR's NEEDS-DESIGN comment and in the #1272 epic thread. Releasing the #1272 claim.

— opus-worker-2

jakildev:
## Architect note — superseded by the #1272 epic + foundation resequencing

#1272 was converted to a tracking epic (children #1299 / #1300 / #1301), and the
design recorded here is superseded by
[`per-axis-trixel-canvas-rotation.md`](https://github.com/jakildev/IrredenEngine/blob/master/docs/design/per-axis-trixel-canvas-rotation.md)
(#1307) plus the sequencing decision in the #1272 thread: per-entity SO(3) is
held behind the foundational rendering-model rework.

**Recommended disposition (human decides):** close as stale (per your own
suggestion in this thread).

Issue #1300, opened 2026-05-28
render: per-entity SO(3) on main canvas — PR-B: per-entity residual face deformation


jakildev:
**Blocked by:** #1299

**Part of epic:** #936 (Rotation architecture) · **Parent:** #1272 (per-entity SO(3) on main canvas) · **PR-B of A/B/C** · **Blocked on #1299 (PR-A)**

## Scope

Per-entity **residual face deformation** — turns PR-A's octahedral-orientation
*stepping* into smooth continuous SO(3) on the main canvas. This is the
per-entity generalization of the residual half of both proven paths: the
camera path's `residualYaw` → `faceDeformationMatrix`, and the DETACHED path's
`octahedralSnapResidual` → `faceDeformationMatrixSO3` (`system_voxel_to_trixel.hpp`
`buildVoxelFrameData` detached branch), moved from a per-canvas UBO to a
**per-entity SSBO** indexed by PR-A's per-voxel transform-slot.

## What lands

1. **Per-entity `faceDeform` SSBO.** Promote `FrameDataVoxelToCanvas::faceDeform_[3]` (per-canvas UBO) to a per-entity array in an SSBO indexed by the per-voxel transform-slot from PR-A. Each entity's 3 slot matrices = `faceDeformationMatrixSO3(visibleFace, residual)` composed with the camera residual, where `residual = octahedralSnapResidual(entityWorldRotation)`. CPU-side bake per rotating entity per frame (one quaternion compose + 3 matrix builds — cheap; matches the existing detached per-canvas bake). Stage_1 reads `faceDeform[entitySlot][workgroupSlot]`.

2. **Per-entity super-sample cap.** `emitDeformedFace`'s `maxN` is gated today by the per-canvas `isDetachedCanvas` uniform (world: 2, detached: 6). Generalize so `MAIN_CANVAS_SO3` entities raster at the detached cap (n up to 6) to fill the deformed diamond with no forward-mapping gaps, while the rest of the main canvas keeps n=2. Drive it off the per-entity slot, not a single canvas-wide flag.

[...]

PR #1243, opened 2026-05-27
T-389: fleet: delete orphaned queue scripts

jakildev:
## Summary
- Delete `fleet-queue-tick`, `fleet-tasks-render`, and `fleet-stale-sweep` — all unreachable after T-381 (#1229) switched the scout to read issues directly instead of maintaining TASKS.md
- Delete their tests: `test_tasks_render.py`, `test_stale_sweep.py`, `test_fleet_claim_master_lock.sh`
- Remove install/symlink blocks and var declarations from `install.sh`
- Remove `Bash(fleet-tasks-render:*)` and `Bash(fleet-queue-tick:*)` from `.claude/settings.json` allowlist
- Remove live `fleet-stale-sweep` call from `fleet-state-scout`'s queue-manager projection block (the issue's "no live callers" claim was incorrect — scout was still spawning it); `fleet-claim cleanup` still runs on projection change
- Convert three stale comments in `fleet-dispatcher` to historical notes
- Update `fleet:stalled` label description to drop the now-deleted tool attribution

## Test plan
- [x] `grep -r "fleet-queue-tick\|fleet-tasks-render\|fleet-stale-sweep" scripts/ .claude/` returns only comment-line hits and gitignored `.pyc` binary cache files (no live callers)
- [x] `install.sh` var declarations, chmod loop, and symlink blocks all updated cleanly
- [x] `fleet-state-scout` queue-manager block still spawns `fleet-claim cleanup --gh` on projection change

Closes #1233

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1305, opened 2026-05-28
fleet: verify genuine #N reference before stamping scope-shipped

### LINKED ISSUE #1304: fleet: queue-ingest scope-shipped false-positive — stamps issues from incidental/no number match (mis-marks #1300, #1284)
## Summary

`scripts/fleet/fleet-queue-ingest`'s **scope-shipped pre-flight** mislabels
queued issues as `fleet:scope-shipped` based on a GitHub full-text search that
matches **incidental number occurrences** (even line numbers in unrelated PR
*comments*) — and in at least one case matches a PR that does not contain the
issue number anywhere. The label keeps the issue out of the queue and posts a
misleading "scope already landed" comment; removing the label does not stick
because the next ingest run re-stamps it.

Two live false positives right now, from two unrelated PRs — this is systemic,
not a one-off.

## Root cause

`fleet-queue-ingest` (~L90-103) does, per pending issue `n`:

```python
gh pr list --repo <repo> --state merged --search f'#{n}' \
   --json number,title,url,mergedAt --limit 5
...
best = prs[0]            # ← trusted with zero verification
# → comment + add fleet:scope-shipped
```

Two compounding flaws:

1. **The search is not a reference match.** GitHub strips the `#` and matches
   the **bare token `n`** in title / body / **comments** (and relevance can
   return results with no literal match at all). `--search '#1300'` and
   `--search '1300'` return identical results.
2. **`prs[0]` is trusted blindly.** No check that the matched PR genuinely
   references issue `n` (e.g. a real `#n` cross-reference / "Closes #n" /
   "supersedes #n").

## Evidence (current, on master)

| Issue | Real status | Spuriously matched merged PR | Why it matched |
|---|---|---|---|
| **#1300** (per-entity SO(3) PR-B) | open, **not** shipped (blocked on #1299) | #1243 "fleet: delete orphaned queue scripts" | a **comment** on #1243 reads *"…lines 71, 176, **1300**, 3619, 3709…"* — "1300" is a line number |
| **#1284** (fleet-dispatcher safety rearm) | open, **has an open PR** #1285 | #1160 "codegen: emit per-component tick…" | #1160 contains "1284" in **neither title, body, nor comments** — pure relevance false-positive |

Neither matched PR contains a genuine `#{n}` reference. #1243's title/body have
no "1300" at all; #1160 has no "1284" at all.

## Why label removal doesn't stick (the recurrence)

Stamped issues are dropped from `pending_issues` (correct — see
`test_scope_shipped_excluded_from_pending`). When a human removes
`fleet:scope-shipped`, the `human:approved` issue re-enters `pending_issues`,
the next ingest re-runs the same search, re-matches the same spurious PR, and
**re-stamps the label and re-posts the comment**. From the human's side the
label "keeps coming back" every ingest cycle and the issue accumulates
duplicate scope-check comments.

## Fix

Verify a **genuine reference** before stamping; do not trust `prs[0]`.

1. Fetch each candidate PR's `body` (add `body` to the `--json` fields).
2. Keep a candidate only if a **word-boundary `#{n}`** appears in its title or
   body (regex `(?<!\d)#{n}(?!\d)`). This is the real cross-reference a
   genuine "scope landed elsewhere" PR carries; it rejects bare-number hits in
   comments/line-numbers and relevance-only matches.
3. If no candidate qualifies → keep the issue in pending (not scope-shipped).
4. (Optional hardening) also gate on PR `mergedAt` recency and/or skip issues
   that are children of an open `fleet:epic` / carry an explicit
   `Blocked on #N` marker, since stacked sub-issues are expected to be
   un-shipped by construction.

### Testability

The detection logic lives inline in a bash heredoc and has **no unit
coverage** (`test_queue_manager_projection.py` only tests the
projection-exclusion side). Extract the "does PR genuinely reference issue n"
predicate into a testable Python function and add regression cases:

- comment-only / line-number match (the #1243→#1300 shape) → **not** shipped.
- relevance match with no literal number (the #1160→#1284 shape) → **not** shipped.
- genuine `Closes #n` / `#n` in PR body → shipped.

## Cleanup + propagation

- Remove the bad `fleet:scope-shipped` label from **#1300** and **#1284** (done
  alongside this filing; will recur until the fix lands).
- The running queue-manager executes from
  `.claude/worktrees/queue-manager-ingest/scripts/fleet/fleet-queue-ingest`,
  so the fix must land on `master` **and** that worktree must pull before the
  recurrence stops.

## Acceptance

- A merged PR that mentions an issue's number only incidentally (comment, line
  number, relevance) no longer triggers `fleet:scope-shipped`.
- A genuine "scope landed under a different PR" (real `#{n}` reference in the
  PR body) still gets flagged.
- Regression tests cover the comment-only, no-literal-match, and genuine-ref cases.
- #1300 and #1284 stay un-stamped after the fix propagates.

`[opus]` — queue-integrity logic; the genuine-vs-incidental distinction and the
"don't regress legitimate scope-shipped detection" constraint need care. Bounded enough for sonnet if preferred.

### LINKED ISSUE #1500: render: detached canvas entities don't track camera yaw (screen-locked while world rotates)
**Model:** opus
**Area:** engine/prefabs/irreden/render/systems/system_entity_canvas_to_framebuffer.hpp:104
**Repro:** IRCanvasStress (canvas_stress) with `--auto-rotate` / camera yaw
**Blocked by:** #1499 (same placement code — do the composite-placement fix first, then add yaw-awareness to the corrected placement; parallel work would conflict)
**Related:** epic #1444, #1496 (WASD camera panning screen-relative under Z-yaw)

## Symptom

When the camera yaws, **detached** canvas entities on the fringes stay screen-locked instead of orbiting with the rotating world. Attached/GRID content tracks the yaw correctly; only the detached entities are pinned to their non-yawed screen positions.

## Root cause

`system_entity_canvas_to_framebuffer.hpp:104` projects the entity's world position with the **non-yawed** iso projection:

```cpp
vec2 entityIso = pos3DtoPos2DIso(worldTransform.translation_);
```

`pos3DtoPos2DIso` has no camera-yaw term, so the framebuffer placement of each detached canvas ignores camera rotation. The yawed projection (`pos3DtoPos2DIsoRotated`, used elsewhere for the rotated world) must drive the placement instead — the same screen-relative-under-Z-yaw treatment #1496 applied to camera panning.

## Why separate from #1499

This is a distinct bug from the off-origin sliver fragmentation: it only manifests **under camera yaw**, whereas #1499 reproduces at yaw=0. But both edit the entity-placement math in this same file, so they must be sequenced (this one stacks on #1499) rather than run in parallel.

## Fix direction

Replace the non-yawed `pos3DtoPos2DIso(worldTransform.translation_)` placement with the camera-yaw-aware projection (`pos3DtoPos2DIsoRotated` or equivalent), consistent with how the GRID/world path and #1496 handle Z-yaw. Verify the detached entity's framebuffer center orbits correctly as the camera yaws.

## Acceptance

- Under `--auto-rotate` (or any camera yaw), detached entities orbit with the world in lockstep with attached content — no screen-locking.
- No regression to the off-origin fix from #1499 (detached entities still render as solid cubes).
- GLSL + Metal parity; re-capture P5 baselines (#1466) for any yaw-inclusive shots.

`[opus]` — placement-projection fix in the detached SO(3) composite; stacks on #1499.

jakildev:
## Summary
- `fleet-queue-ingest`'s scope-shipped pre-flight trusted `prs[0]` from `gh pr list --search '#N'`, but GitHub strips the `#` and matches the bare token `N` anywhere (titles, bodies, **comments**, line numbers) and can return relevance-only hits with no literal `N`. This mislabeled unrelated issues `fleet:scope-shipped`.
- Two live false positives this confirms against: **#1300** matched a line number inside a comment on PR #1243; **#1284** matched PR #1160 on relevance alone (no literal "1284" anywhere). Both re-stamped on every ingest cycle after a human cleared the label.
- Extracts the verification into a testable sibling module `scripts/fleet/fleet_scope_shipped.py`: a candidate PR counts as scope-shipped only if a **word-boundary `#{n}`** appears in its title or body (`select_shipped_pr`). The heredoc imports it via the script's `realpath` so the `~/bin/fleet-queue-ingest` install symlink resolves to the versioned copy — no `install.sh` change needed (the module ships beside the script in `scripts/fleet/`).
- Cleared the stale `fleet:scope-shipped` label from #1300 and #1284 (it had recurred since filing, as the issue predicted).

## Propagation note
Per the issue: the running queue-manager executes from its own worktree, so the recurrence stops once this lands on `master` **and** that worktree pulls. No behavior change until then.

## Test plan
- [x] `python3 scripts/fleet/tests/test_scope_shipped_reference.py` — 13 tests (comment-only, no-literal-match, longer-number prefix/suffix, genuine title/body ref, select-first-genuine).
- [x] Full fleet python suite green (`test_enrich_stackable_blocker_prs`, `test_queue_manager_projection`, `test_merger_projection`, `test_worker_projection`, `test_issue_queue_parser`, `test_fleet_claude_stream_latch`).
- [x] `bash -n scripts/fleet/fleet-queue-ingest` clean.
- [x] End-to-end heredoc run with stubbed `gh`: a comment-only `#1300` hit and a relevance `#1284` hit stay in pending; only a genuine `Closes #1500` PR is stamped (`1,2` shipped/kept).
- [x] Import resolution verified through the `~/bin` symlink realpath and the worktree path.

Closes #1304

### DIFF
--- a/scripts/fleet/fleet-queue-ingest
+++ b/scripts/fleet/fleet-queue-ingest
@@ -67,8 +67,14 @@ fi
 # list. Mutating the projection file lets the label-stamping pass below
 # skip scope-shipped issues without re-querying (#1175).
 if [[ -f "$PROJECTION_FILE" ]] && command -v gh >/dev/null 2>&1; then
-    scope_shipped_out=$(FLEET_PROJ_FILE="$PROJECTION_FILE" python3 - << 'SCOPE_PY' 2>>"$LOG_FILE"
+    scope_shipped_out=$(FLEET_PROJ_FILE="$PROJECTION_FILE" INGEST_SELF="${BASH_SOURCE[0]}" python3 - << 'SCOPE_PY' 2>>"$LOG_FILE"
 import json, os, subprocess, sys
+
+# Import the genuine-reference predicate from the sibling module. realpath
+# resolves the ~/bin/fleet-queue-ingest install symlink back to the versioned
+# scripts/fleet/ copy, where fleet_scope_shipped.py lives next to this script.
+sys.path.insert(0, os.path.dirname(os.path.realpath(os.environ['INGEST_SELF'])))
+from fleet_scope_shipped import select_shipped_pr
 
 REPO_SLUGS = {"engine": "jakildev/IrredenEngine", "game": "jakildev/irreden"}
 
@@ -87,20 +93,23 @@ for issue in pending:
     if not n:
         kept.append(issue)
         continue
-    # Full-text search; may match incidental references — human verifies via comment before closing.
+    # GitHub strips the '#' and matches the bare token n anywhere (titles,
+    # bodies, comments, line numbers, relevance) — so the search is a coarse
+    # candidate set, not a reference match. select_shipped_pr() verifies a
+    # genuine word-boundary #n cross-reference in title/body before trusting it.
     r = subprocess.run(
         ['gh', 'pr', 'list', '--repo', repo, '--state', 'merged',
-         '--search', f'#{n}', '--json', 'number,title,url,mergedAt', '--limit', '5'],
+         '--search', f'#{n}', '--json', 'number,title,body,url,mergedAt', '--limit', '5'],
         capture_output=True, text=True, timeout=15,
     )
     try:
         prs = json.loads(r.stdout or '[]') if r.returncode == 0 else []
     except Exception:
         prs = []
-    if not prs:
+    best = select_shipped_pr(prs, n)
+    if best is None:
         kept.append(issue)
         continue
-    best = prs[0]
     merged_date = best.get('mergedAt', '')[:10]
     comment = (
         f"Scope check: this issue's scope appears to have landed in "
--- a/scripts/fleet/fleet_scope_shipped.py
+++ b/scripts/fleet/fleet_scope_shipped.py
@@ -1,0 +1,40 @@
+"""Genuine-reference predicate for fleet-queue-ingest's scope-shipped pre-flight.
+
+A merged PR is only evidence that an issue's scope "already landed" when the PR
+carries a genuine cross-reference to the issue — a word-boundary ``#N`` in its
+title or body (a real ``Closes #N`` / ``supersedes #N`` / ``#N`` mention).
+
+GitHub's PR search (``gh pr list --search '#N'``) strips the ``#`` and matches
+the bare token ``N`` anywhere it appears, including PR *comments* and source
+line numbers, and can even return relevance-only hits with no literal ``N`` at
+all. Trusting ``prs[0]`` from that search mislabels unrelated issues as
+``fleet:scope-shipped`` and posts a misleading "scope already landed" comment
+(#1304: #1243's #1300 line-number hit, #1160's #1284 relevance hit). This
+predicate is the verification step that rejects those incidental matches.
+"""
+import re
+
+
+def pr_references_issue(title, body, n):
+    """True iff a word-boundary ``#{n}`` appears in the PR title or body.
+
+    The trailing ``(?!\\d)`` guard stops ``#13000`` from satisfying ``n=1300``;
+    requiring the literal ``#`` rejects bare line-number and relevance hits that
+    carry no cross-reference at all.
+    """
+    if not n:
+        return False
+    pattern = re.compile(r'#' + str(int(n)) + r'(?!\d)')
+    return bool(pattern.search(title or '')) or bool(pattern.search(body or ''))
+
+
+def select_shipped_pr(prs, n):
+    """First PR in ``prs`` that genuinely references issue ``n``, else ``None``.
+
+    Replaces the old ``prs[0]`` blind trust: a merged-PR search hit only counts
+    as scope-shipped evidence when the PR actually cross-references the issue.
+    """
+    for pr in prs:
+        if pr_references_issue(pr.get('title', ''), pr.get('body', ''), n):
+            return pr
+    return None
--- a/scripts/fleet/fleet_scope_shipped.py
+++ b/scripts/fleet/fleet_scope_shipped.py
@@ -18,13 +18,18 @@ import re
 def pr_references_issue(title, body, n):
     """True iff a word-boundary ``#{n}`` appears in the PR title or body.
 
-    The trailing ``(?!\\d)`` guard stops ``#13000`` from satisfying ``n=1300``;
-    requiring the literal ``#`` rejects bare line-number and relevance hits that
+    The leading ``(?<!\\w)`` guard rejects ``abc#1300`` (alphanumeric immediately
+    before ``#``); the trailing ``(?!\\d)`` guard stops ``#13000`` from satisfying
+    ``n=1300``. Together they match GitHub's own link-detection boundaries.
+    Requiring the literal ``#`` rejects bare line-number and relevance hits that
     carry no cross-reference at all.
     """
     if not n:
         return False
-    pattern = re.compile(r'#' + str(int(n)) + r'(?!\d)')
+    try:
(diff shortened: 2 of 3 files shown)


jakildev (commented):
## Review — fleet: verify genuine #N reference before stamping scope-shipped (#1305)

**Verdict:** approve

### Summary

Clean, well-reasoned fix for a real false-positive bug. The root cause analysis is accurate: `gh pr list --search '#N'` strips the `#` and matches the bare token anywhere (line numbers in comments, relevance-only results), so `prs[0]` was an unreliable proxy for genuine cross-reference. The extraction into a testable `fleet_scope_shipped.py` module is the right call — the predicate is now independently verifiable and the test coverage catches the exact failure shapes described in #1304.

**Logic:** `pr_references_issue` uses `r'#' + str(int(n)) + r'(?!\d)'`. The trailing negative lookahead correctly blocks `#13000` from satisfying `n=1300`. The literal `#` requirement rejects bare line-number and relevance hits. `if not n: return False` guards falsy inputs. `select_shipped_pr` iterates candidates and returns the first genuine match — replacing `prs[0]` blind trust.

**Heredoc wiring:** Passing `INGEST_SELF="${BASH_SOURCE[0]}"` and resolving via `os.path.realpath` is the right way to locate the sibling module through the `~/bin/` install symlink. The `body` field is correctly added to the `--json` argument so `pr_references_issue` can inspect it.

**Tests:** 13 tests, all passing. Coverage includes: genuine title ref, genuine body ref, bare-number miss (the #1243→#1300 line-number shape), no-literal-number miss (the #1160→#1284 relevance shape), prefix/suffix digit guards, adjacent punctuation, falsy n, empty candidates, missing title/body keys. The importlib load pattern (cwd-independent) mirrors the existing fleet test suite convention.

---

### Nits

- `scripts/fleet/fleet_scope_shipped.py:27` — the regex has a trailing `(?!\d)` guard but no leading `(?<!\d)` lookbehind before the `#`. A body like `abc#1300 fix` would match even though `#` is immediately preceded by alphanumeric. In practice this is benign — real PR bodies don't produce this shape — but adding `r'(?<!\w)#'` would make the boundary symmetric and match GitHub's own link-detection behavior.

- `scripts/fleet/fleet_scope_shipped.py:27` — `int(n)` raises `ValueError` if `n` is a truthy non-numeric string. The `if not n:` guard above only catches falsy values. In the current call site `n` is always an integer from JSON, so this is purely theoretical — but a `try/except ValueError: return False` around the `int(n)` call would make the function safe against unexpected callers.

---

- [x] `python3 scripts/fleet/tests/test_scope_shipped_reference.py` — 13 tests (confirmed passing by reviewer)
- [x] Full fleet python suite green (as noted in PR body)
- [x] End-to-end heredoc run with stubbed `gh`

---

Opus recheck not required.

jakildev:
Addressed nits:
- Added leading `(?<!\w)` lookbehind to `pr_references_issue` regex — `abc#1300` is now correctly rejected, making the boundary symmetric with GitHub's own link-detection behavior.
- Wrapped `int(n)` in `try/except ValueError: return False` so non-numeric truthy `n` values return `False` instead of raising.
- Added two covering tests (`test_hash_preceded_by_alphanum_is_not_a_reference`, `test_non_numeric_issue_number_returns_false`); suite is now 15 tests, all passing.

jakildev:
Addressed nits:
- Added leading `(?<!\w)#` lookbehind — boundary is now symmetric with the trailing `(?!\d)` and matches GitHub's own link-detection behavior. `abc#1300` no longer matches.
- Wrapped `int(n)` in `try/except ValueError` so a truthy non-numeric `n` returns `False` instead of raising; guards against unexpected callers without affecting current int-from-JSON call sites.
- Two new tests added (15 scope-shipped, 97 fleet total, all green).

Issue #1312, opened 2026-05-28
fleet: architect epic-filing format drift — point role doc at file-epic skill + add post-filing validation


jakildev:
**Model:** opus
**Blocked by:** (none)

**Context:** Surfaced 2026-05-28 while triaging stacked PR queue. The architect filed #1308–#1311 (smooth Z-yaw via per-axis trixel canvases, an Epic decomposition under umbrella #1307) with prose-only `**Blocked on T1 + docs PR #1306**` headers. The scout's `blocked_by` parser only reads the structured `**Blocked by:** #N` form, so all four projected as Available simultaneously and would not have stack-claimed correctly. Manual body edits were needed pre-approval to restore the structured fields.

The same pattern repeated on the earlier #1299/#1300/#1301 stack (under #1272/#936):

| Issue | Standalone `**Blocked by:** #N` line? |
|---|---|
| #1299 (first) | N/A — first in chain |
| #1300 (second) | ❌ Missing — `Blocked on #1299` only in header prose |
| #1301 (third) | ✅ Present, but `**Blocked by:** #1299, #1300` (multi-blocker; current `find-stackable-blockers` requires exactly one) |
| #1308 (first) | N/A |
| #1309-#1311 (pre-fix) | ❌ All prose-only |

Two systemic problems:

1. **`role-opus-architect.md` does not point at the `file-epic` skill.** The "Filing tasks" (line 201-224) and "Planning issues" (line 226+) sections describe single-task filing. Nothing in either section says "if the decomposition is a multi-issue stack, invoke `file-epic`". The skill at `~/.claude/skills/file-epic/SKILL.md` exists and has the correct template, but the architect role doc never references it.

2. **`file-epic` skill has no post-filing validation.** The template at SKILL.md:99-124 prescribes `**Model:** ...`, `**Part of epic:** #N`, `**Blocked by:** #<prior>` as standalone lines. But if the architect deviates (manual edits, condensed header pattern, "Blocked on" wording), nothing catches it — the skill returns success even when the structured fields are missing.

**Proposed fix (two parts, both small):**

### Part A — Update `role-opus-architect.md`

In the "Filing tasks" section ([line 201-224](.claude/commands/role-opus-architect.md)) add a "Multi-issue stacks" subsection that explicitly hands off to `file-epic`:

```markdown
### Multi-issue stacks (epic decomposition)

When a task decomposes into a stack of N issues that should each
[...]

Issue #1311, opened 2026-05-28
render: smooth camera Z-yaw — T4: AO/lighting on the resolved composite


jakildev:
**Epic:** #1307 · **Design:** `docs/design/per-axis-trixel-canvas-rotation.md` (PR #1306) · **PR-4 of 5** · **Blocked on T3 + docs PR #1306**
**Blocked by:** #1310, #1344 (T3 #1310 implements the composite the AO/lighting passes target; demo #1344 is the voxel-only capture vehicle and must land FIRST so T4 verifies lighting on the composite against it — worker should stack-claim on T3's PR)
**Model:** opus

## Scope

Run **AO / lighting on the resolved composite**, not per-canvas.

- AO / sun-shadow / lighting read the **post-composite framebuffer** with the **winning face-id + depth** carried through stage 3, so they see one unified surface.
- Confirm no double-rotation / face-label drift (the #1275 symptom) under continuous yaw — AO neighbor sampling uses the winning face's context.
- **Source-agnostic composite G-buffer (makes SDF lighting order-independent — #1345).** The resolved composite must carry, per winning pixel: depth, color, **world normal**, and a **face-id / material tag** distinguishing a voxel face vs an SDF surface vs background. Directional + point + sun-shadow + light-volume lighting read this G-buffer **generically**, so *any* composited layer is lit — the three voxel canvases now, the SDF canvas when #1345 lands — with **no T4 rework**. This is why T4 is NOT blocked on T5 (and vice-versa): the G-buffer is the contract, not the producer set.
- **AO is the one voxel-specific consumer.** Voxel AO samples neighbor voxel faces; it must **gate on the voxel face-id tag and skip non-voxel (SDF) pixels** (they take the ambient default — acceptable, SDF is effects-only #937). Never sample voxel neighbors for an SDF-winning pixel. SDF analytic AO, if ever wanted, is scoped in #1345.
- Fold this lighting / G-buffer contract into `docs/design/per-axis-trixel-canvas-rotation.md` §Open-decisions (Lighting/AO placement) so it is durable.

## Verification

Capture lighting / AO / sun-shadow on **`IRVoxelYaw` (#1344)** — the voxel-set-only camera-yaw demo — at a parked non-cardinal yaw (~30°) and across a `--spin-yaw` sweep, vs. the cardinal frames. #1344 lands before this task (see Blocked by) precisely so it is the ready capture vehicle here; also run the existing lit demos (`lighting`, `canvas_stress`).

[...]

Issue #1301, opened 2026-05-28
demo: IRRotationCompare — main-canvas SO(3) vs detached SO(3) side-by-side + render-verify baselines


jakildev:
**Part of epic:** #936 (Rotation architecture) · **Parent:** #1272 (per-entity SO(3) on main canvas) · **PR-C of A/B/C** · **Blocked on #1300 (PR-B)**

**Blocked by:** #1299, #1300

## Scope

The deliverable demo that **gates #1272's close**: a single creation —
`IRRotationCompare` (new) or an extension of `canvas_stress` — that renders
both rotation paths **side by side** so the visual + perf comparison is
direct. Without it there is no way to verify the main-canvas path produces
matched output to the proven DETACHED path, and no way to justify the perf
claim.

## Scene layout

- **Left half:** N entities (N = 8–16) using **`MAIN_CANVAS_SO3`** (PR-A/PR-B path).
- **Right half:** N entities — same N, same shapes, same rotation drives — using **`DETACHED`** (T-295 path).
- Both halves share one lighting volume, sun direction, and camera.

## Entity variety (each side includes all)

1. Simple cube.
2. Non-cube box (e.g. 4×2×6).
3. Multi-color/material composition (small voxel character: torso + head + limbs as one entity).
4. Hollow / partially-emptied shape (exercises per-voxel face deformation on internal faces + the exposed-mask).
5. High-voxel-count detail (e.g. 32³ sphere) — the perf comparison subject.

## Rotation drives (continuous; one per entity, full SO(3))

Pure Z-yaw · pure X-pitch (falls forward) · pure Y-roll (tips sideways) ·
diagonal (1,1,1) · off-axis (1,2,0 normalized) · wobble (axis changes over
time); remaining entities use slower variants so the demo isn't a blur.

## Validation

- **Visual parity:** matched-shape left/right pairs look identical (orientation, color, lighting) every frame. Per-region pixel diff between left-half and right-half is near-zero (allow small rasterization-parity differences). Use `tools/img_diff`.
- **Perf:** per-half perf overlay (the redesigned `PERF_STATS_OVERLAY`, PR #1119) shows GPU stage time for left vs right. Main-canvas SO(3) should be **lower per-rotating-voxel cost** at high entity counts — the point of the issue, and the answer to the original "prototype both and measure". (No separate PR-D; this overlay is the comparison.)
[...]

PR #1317, opened 2026-05-28
fleet: add fleet-validate-stack epic-stack body validator (Refs #1312)

jakildev:
## Summary

Adds `fleet-validate-stack <umbrella>` — a pre-`human:approved` sanity check that an epic's child issues carry the **standalone structured fields** the queue machinery actually parses:

- `**Model:** opus|sonnet` (every child)
- `**Part of epic:** #<umbrella>` (every child — and what `file-epic`'s own `--search "Part of epic: #N"` discovery relies on)
- single-ref `**Blocked by:** #N` (every non-head child)

Prose `Blocked on …` header bullets and multi-ref `**Blocked by:** #A, #B` lines are invisible to the scout's `blocked_by` parser and `fleet-claim`'s `find-stackable-blockers` predicate, so a stack filed that way silently never stack-claims and isn't discoverable — the exact drift that has recurred (#1300, #1301, #1308–#1311). This catches it up front rather than at queue time.

## What's here

- **`scripts/fleet/fleet_validate_stack.py`** — pure, importable predicates (`is_epic_child`, `validate_child`, `validate_stack`). No I/O.
- **`scripts/fleet/fleet-validate-stack`** — executable; `gh` discovery via coarse search + precise membership filter (mirrors `fleet_scope_shipped`). Flags `--repo {engine,game}`, `--children N…` (bypass discovery), `--state`, `--strict`.
- **`scripts/fleet/tests/test_fleet_validate_stack.py`** — 20 unit tests.
- **`install.sh`** symlinks it; **`fleet-help`** indexes it.

### Severity model
A missing `**Blocked by:**` line is a **warning**, not an error: an interior root of a multi-root epic legitimately has none, and the body alone can't distinguish that from drift. `--strict` promotes it to an error for known linear chains. Unambiguous violations (wrong/missing epic field, multi-ref blocker, malformed blocker, missing Model) are always hard errors (exit 1).

## Scope note — Part B deferred

Per the issue's 19:21 narrowing comment, #1312 = **Part B** (inline validation in `file-epic` SKILL.md step 7.5) + optional **Part C** (this helper). This PR delivers Part C.

**Part B can't ride a fleet PR:** `file-epic` is the *only* skill not tracked in the engine repo — it lives solely at `~/.claude/skills/file-epic/SKILL.md`, while all 37 other skill files are under `.claude/skills/`. Editing it requires a human decision on whether to repo-track `file-epic`. Once decided, Part B reduces to one line in step 7.5: call `fleet-validate-stack <umbrella>` (it already exits 1 and prints each failure). This PR therefore uses **Refs**, not Closes.

## Test plan
- [x] `python3 -m unittest discover -s scripts/fleet/tests` → 131 pass (20 new)
- [x] Live: `fleet-validate-stack 1307` → flags the `**Epic:**`-vs-`**Part of epic:**` drift on all 4 children (exit 1)
- [x] Live: `fleet-validate-stack 226 --state all` → flags #1070's multi-ref `**Blocked by:**` (error); #1068/#1195/#1196 missing-blocker as warnings (interior roots)
- [x] Live: `--children 1069 1071 1072` → clean PASS (exit 0); empty-discovery path prints guidance (exit 0)
- [x] `bash -n install.sh`, `bash -n fleet-help`; `fleet-help` renders the new entry

Refs #1312


jakildev (commented):
## Review — #1317: fleet: add fleet-validate-stack epic-stack body validator (Refs #1312)

**Verdict:** approve

### Summary

Clean, well-motivated addition. The pure/executable split (`fleet_validate_stack.py` for predicates, `fleet-validate-stack` for gh I/O) is the right separation of concerns and makes the unit tests simple. Severity model (error for unambiguous violations, warn for ambiguous missing-blocker) is well-reasoned: the multi-root epic case (#226 shape) correctly stays `ok` with only warnings. The real-world fixture bodies from issues #1307–#1311 are a nice regression anchor for the exact drift that motivated this PR.

The coarse-search-then-precise-filter discovery pattern mirrors `fleet_scope_shipped`, which is the correct idiom for GitHub's `in:body` search stripping `#`.

install.sh and fleet-help are updated correctly; symlink and chmod entries follow the existing pattern precisely.

### Nits

- `scripts/fleet/tests/test_fleet_validate_stack.py:34` — `_sev` is defined (`return {f["severity"] for f in findings}`) but never called in any test. Dead helper; remove or use it.

- `scripts/fleet/fleet_validate_stack.py:114-129` — Multiple separate `**Blocked by:** #N` lines (two single-ref lines, each valid on its own) would pass validation even though they represent a two-blocker situation `fleet-claim` can't handle. `bb_lines` would have length 2, but only `bb_lines[0]` is checked. Adding `elif len(bb_lines) > 1: err("multiple separate ...")` before the single-ref check would close this gap. Low-priority since this form is unusual in practice.

### Praise

- The `--strict` flag is a well-designed escape hatch: warnings become errors for known linear chains without making the common ambiguous case (multi-root epics) noisy. The "per-line still reads WARN so the human sees why it is ambiguous" behavior is clearly intentional and correct.
- Empty-discovery path exits 0 with guidance text rather than exit 1 — correct because an epic with no children yet is not a structural failure, and the guidance explains exactly how to force-validate a known set.

- [x] `python3 -m unittest discover -s scripts/fleet/tests` (author reports 131 pass, 20 new)
- [ ] `fleet-validate-stack 1307` against current live issues to confirm the tool still flags the `**Epic:**`-vs-`**Part of epic:**` drift on the #1307 children.

Opus recheck not required.

jakildev:
Addressed nits:
- Removed dead `_sev` helper from `test_fleet_validate_stack.py` (defined but never called)
- Added `elif len(bb_lines) > 1` guard in `validate_child` to catch multiple separate `**Blocked by:**` lines — previously only `bb_lines[0]` was examined, so two single-ref lines would silently pass
- Added `test_multiple_separate_blocked_by_lines_is_error` covering the new code path

All 132 tests pass (was 131 before). `fleet:approved` stands — no review needed.

PR #1320, opened 2026-05-28
fleet: relaunch workers/reviewers fresh on crash, not --continue

jakildev:
## Problem

babysit's mid-task **crash recovery** for worker/reviewer roles ran
`claude --print --continue "resume your work…"`. Those sessions launch with
`--include-partial-messages` + extended thinking (`--effort xhigh`), and the
persisted `thinking` blocks don't round-trip on resume — `claude --continue`
fails with:

> API Error: 400 … thinking blocks in the latest assistant message cannot be modified

So the recovery path silently dropped the pane back to a shell with no relaunch.
Surfaced on `opus-reviewer`: a clean run (`done success`) followed by a
`--continue` of that session 400'd. Architects were never affected (interactive
sessions, no `--include-partial-messages`).

## Fix

`resume_mid_task` now relaunches **fresh** for non-architect roles (reuses
`launch_worker_fresh`) instead of `--continue`. Fresh relaunch is safe and
idempotent: the role re-discovers its in-flight task from its claim/reservation
(worker reservation gate "step 0.5"; reviewers are read-only and just
re-review). The crashed session's in-context scratch is lost, but the task is
not. Architects keep their `--resume` path unchanged.

Also corrected the stale dry-run hint + three comments that described
`--continue` as preserving partial work.

## Validation

- `bash -n scripts/fleet/fleet-babysit` — clean.
- `scripts/fleet/tests/test_babysit_effort.sh` — 8/0 (no effort-resolution change).
- The live fleet runs workers/reviewers via the **dispatcher** (always fresh), so
  this path is dormant there; the fix hardens babysit-managed and `review-only`
  setups where a thinking-role crash would otherwise fail to recover.

### DIFF
--- a/scripts/fleet/fleet-babysit
+++ b/scripts/fleet/fleet-babysit
@@ -22,8 +22,9 @@
 #       misses an event.
 #
 #   Either way the relaunch is fresh (no --continue, no --resume), so the
-#   new task starts with a clean conversation. Mid-task crashes still use
-#   --continue so partial work isn't lost.
+#   new task starts with a clean conversation. Mid-task crashes ALSO relaunch
+#   fresh: --continue can't resume a thinking session (see resume_mid_task),
+#   and the role re-discovers its task from its claim/reservation anyway.
 #
 #   Architects (opus-architect, game-architect): one long conversation,
 #   preserved across crashes AND across fleet-down/fleet-up cycles via
@@ -411,9 +412,11 @@ run_with_heartbeat() {
 #     a clean conversation. This stops a context built up over hours of
 #     prior task work from leaking into the next task's reasoning.
 #
-#     Mid-task crashes are still handled via --continue so partial work
-#     doesn't get lost — the LOOP_INTERVAL gating is the "ok, *task is
-#     done*, start fresh" signal, not "agent is unhealthy, retry".
+#     Mid-task crashes relaunch fresh too (NOT --continue — it can't resume
+#     a thinking session; see resume_mid_task). The in-flight task is
+#     recovered from the role's claim/reservation, not from session context.
+#     The LOOP_INTERVAL gating is the "ok, *task is done*, start fresh"
+#     signal, not "agent is unhealthy, retry".
 
 launch_architect_first() {
     if [[ -f "$SESSION_FILE" ]]; then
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — #1320: fleet: relaunch workers/reviewers fresh on crash, not --continue

**Verdict:** approve

### Summary

Clean, well-motivated fix for a silent crash recovery failure. The root cause is real and well-documented: `--include-partial-messages` sessions produce `thinking` blocks in the persisted transcript, and `claude --continue` on such a session fails hard with API 400 `thinking blocks in the latest assistant message cannot be modified`. The old `--continue` path was silently dropping worker/reviewer panes back to a shell on any crash — no relaunch, no log entry, no recovery.

The fix is minimal and correct: `resume_mid_task()` for non-architect roles now delegates to `launch_worker_fresh` (which issues `/role-$ROLE $MODE` on a fresh session) instead of `stream_claude --print --continue`. Architects retain their `--resume` path because their sessions are created without `--include-partial-messages` and their transcript does round-trip on resume.

### Correctness walk

- **`resume_mid_task` guard condition** — `if [[ "$ARCHITECT_RESUME" -eq 1 && -f "$SESSION_FILE" ]]` correctly gates the `--resume` path; the `else` branch (`launch_worker_fresh`) covers all workers/reviewers and the edge case of an architect whose SESSION_FILE was not yet written (pre-existing, no regression).
- **Exit code 2 (usage limit) path** — `resume_mid_task` is also called on `prev_exit == 2`. With the new code, workers get `launch_worker_fresh` after the `LIMIT_DELAY` wait. Correct: a fresh relaunch is idempotent; the role re-discovers its in-flight task from claim/reservation.
- **Dry-run hint update** — line 528 log message no longer suggests `--continue` to the human; updated to "relaunch fresh to iterate". Correct.
- **Comment accuracy** — all three stale `--continue` references in comments are updated. Header comment at line 24–27, `run_with_heartbeat` comment at line 412–419, and `case *) comment at line 600–602 are now consistent with behavior.

### Validation coverage

- `bash -n scripts/fleet/fleet-babysit` — clean (confirmed in PR body).
- `test_babysit_effort.sh` 8/0 — no effort-resolution regressions (confirmed in PR body).
- The crash path itself is dormant in dispatcher-managed fleets (dispatcher always launches fresh), so this hardens babysit-managed and `review-only` setups.

### Praise

- The comment block in `resume_mid_task` (lines 470–485) is exactly the right length — it names the specific API error text, explains why fresh is safe (`claim/reservation` recovery), and calls out the pre-condition asymmetry between architects and workers. A future maintainer won't have to rediscover this.

Opus recheck not required.

Issue #1310, opened 2026-05-28
render: smooth camera Z-yaw — T3: trixel→framebuffer passes + depth composite + parity (core)


jakildev:
**Epic:** #1307 · **Design:** `docs/design/per-axis-trixel-canvas-rotation.md` (PR #1306) · **PR-3 of 4** · **Blocked on T2 + docs PR #1306**
**Blocked by:** #1309 (T2 must be implemented — composite passes need Stage-1 routing + per-canvas geometry; worker should stack-claim on T2's PR)
**Model:** opus

## Scope

Three **trixel→framebuffer passes + depth composite + face-jump + parity**.
This ticket's core is **parity correctness** — the #1256 checkerboard class.

- One trixel→framebuffer pass per canvas, each applying **its own basis (slopes)** to expand trixels to framebuffer pixels.
- Each pass `atomicMin`s its **shared `pos3DtoDistance`** into the **shared framebuffer depth buffer**; writes color + winning face-id where it wins. Composite = per-pixel nearest across the three (valid: shared metric, yaw-only).
- **Face-jump at the ±45° rebracket:** the swept axis's visible ± face swaps; must be a no-op visually (swap at zero on-screen width). Inherit `rasterYaw` snap / residual sign flip / `cardinalLowerCornerShift`; ensure per-canvas face identity swaps at the zero-width instant.

## Parity (HARD ACCEPTANCE GATE)
Each per-canvas trixel→framebuffer conversion must preserve a **consistent
even-parity tiling** (`(iso.x+iso.y)&1==0`) under its affine basis at **every**
yaw in the ±45° bracket:
- adjacent trixels tile gap-free **and** never double-write a framebuffer pixel;
- CPU/GPU classify half-integer positions identically (`roundHalfUp`, never `glm::round`);
- validate with continuous-rotation `--spin-yaw` sweeps + **native-resolution ROI crops** (not downscaled) inspected for checkerboard / lattice / seam / double-write (`render-debug-loop`, `tools/img_diff`).

## Acceptance
- Smooth rotation, no 90° snap; correct occlusion across the three canvases.
- No pop at the ±45° face-jump.
- **Zero parity artifacts at every yaw** (the gate above).
- Byte-identical at `residualYaw == 0`.

`[opus]`

## Side effects to evaluate (see epic)
- **Entity-id readback / picking:** winning entity-id must come from the composite, not a single canvas.
[...]

PR #1327, opened 2026-05-28
fleet: stack single-blocker tasks even when blocked_by carries prose

jakildev:
## Problem

The scout's stackable-blocker enricher (`enrich_stackable_blocker_prs`) only
attaches `stackable_blocker_pr` to a task when its `blocked_by` field matches
`SINGLE_TASK_BLOCKER_RE = ^#\d+$` — i.e. the **entire** field must be exactly
`#NNN`. The epic decomposer routinely writes explanatory prose after the ref:

```
#1309  blocked_by: "#1308 (T1 must be implemented … worker should stack-claim on T1's PR)"
#1310  blocked_by: "#1309"
```

So `#1309` failed the gate, never got a `stackable_blocker_pr`, and the whole
`T1→T2→T3→T4` chain (#1308→#1309→#1310→#1311) could never stack — while the
terse-authored `#1310` enriched fine. The only difference was the prose.
Observed live on **#1309 → PR #1316, 2026-05-28**: the downstream task sat
through the upstream's entire review cycle instead of stacking on its open PR.

## Fix

Replace the whole-field regex with `_single_blocker_issue()`: a value is a
single blocker when it references **exactly one** issue
(`len(re.findall(r"#(\d+)", …)) == 1`).

- ✅ `#1308 (why)` → single blocker `#1308` (the case that was dropping chains)
- ✅ `#A, #B` **and** `#A and #B` → two refs → multi-blocker, rejected. This is
  *more* robust than a bare regex relaxation, which would have matched
  `#A and #B` as a single blocker.
- ✅ `(none)`, free text, a bare URL, legacy `T-101` → zero refs → rejected.

Everything else in the enricher (single-PR-match, multi-match safety, design-block
filter, area-overlap filter, cross-repo isolation) is unchanged.

## ⚠️ Behavior reversal — please note

The prior tests **intentionally** asserted that `#101 (waiting)` should *not*
stack (it was bucketed with free text). This PR makes a single ref + prose a
**pass-through**. The test moves from the negative list into a new positive
case (`test_single_ref_with_prose_enriched`), and `test_two_refs_with_prose_no_field`
is added to lock in the multi-blocker rejection.

## Tests

`python3 scripts/fleet/tests/test_enrich_stackable_blocker_prs.py` — **14 pass**
(12 prior + 2 new). Scout parses clean; `SINGLE_TASK_BLOCKER_RE` is fully removed
(only had the one call site).

### DIFF
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -700,11 +700,23 @@ _SMOKE_SKIP_LABELS = frozenset({
 
 MODEL_TOKEN_RE = re.compile(r"\b(opus|sonnet)\b")
 
-# Single-issue blocker (`#1234`). Rejects free-text, URL, and multi-blocker
-# forms. Evaluated AFTER resolve_blocked_by() reduces multi-ref bodies to
-# their unresolved set, so a task originally blocked on `#N1, #N2` where #N1
-# is now CLOSED will have blocked_by=`#N2` at this point and will match.
-SINGLE_TASK_BLOCKER_RE = re.compile(r"^#\d+$")
+# A blocked_by value is a "single blocker" — eligible for a stackable claim —
+# when it references exactly one issue. The reference may carry trailing
+# explanatory prose, e.g. "#1308 (T1 must land first)"; the epic decomposer
+# emits that shape, and rejecting it silently kept whole task chains from ever
+# stacking (observed on #1309 → #1316, 2026-05-28). Two or more #refs (a
+# genuine multi-blocker like "#1308, #1309" or "#1308 and #1309") and zero
+# refs ("(none)", free text, a bare URL, legacy "T-101") are NOT single-blocker
+# and get no stackable_blocker_pr.
+#
+# Evaluated AFTER resolve_blocked_by() reduces multi-ref bodies to their
+# unresolved set, so a task originally blocked on `#N1, #N2` where #N1 is now
+# CLOSED arrives here as `#N2` (one ref) and qualifies.
+def _single_blocker_issue(blocked_by):
+    """Return the sole blocker issue number (digits, no '#') when blocked_by
+    references exactly one issue, else None."""
+    refs = re.findall(r"#(\d+)", blocked_by or "")
+    return refs[0] if len(refs) == 1 else None
 
 
 def model_tags(task):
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — #1327: fleet: stack single-blocker tasks even when blocked_by carries prose

**Verdict:** approve

### Summary

Clean, minimal fix to a real stacking deficiency. The old `SINGLE_TASK_BLOCKER_RE = re.compile(r"^#\d+$")` gate required the entire `blocked_by` field to be exactly `#NNN` — the epic decomposer's prose suffix (e.g. `#1308 (T1 must land first)`) caused the whole T1→T2→T3→T4 chain to silently never stack while terse siblings enriched fine. The new `_single_blocker_issue()` correctly counts `#NN` refs and gates on exactly-one, which is the right semantic.

### What is correct

- **Regex semantics are correct.** `re.findall(r"#(\d+)", blocked_by or "")` returns the digits-only capture group. One ref → single blocker. Two or more refs → multi-blocker rejected. Zero refs (free text, bare URL without a `#NNN` fragment, `T-101`, `(none)`) → rejected. All cases tested.
- **Multi-blocker rejection is now more robust than the old approach.** The PR body calls this out: `#A and #B` would have been accepted by a naive regex relaxation like `r"^#\d+"`, but the count-based approach correctly rejects it. The new `test_two_refs_with_prose_no_field` subtest locks this in.
- **Call order interaction is correct.** `resolve_blocked_by()` runs before `enrich_stackable_blocker_prs()` in `tick_once()` (line 1378–1379), so a task originally blocked on `#N1, #N2` where `#N1` is now CLOSED arrives here as `#N2` (one ref) and qualifies correctly — as documented in the module-level comment.
- **Return value is the right type.** `refs[0]` gives just the digit string (no `#`), matching the old `blocked_by.lstrip("#")` behavior. `f"claude/{issue_num}-"` prefix construction is unchanged.
- **Tests: 14/14 pass**, including 2 new cases covering the positive single-ref-with-prose path and the negative two-refs-with-prose guard.

### Nits

- `scripts/fleet/fleet-state-scout:703-714` — the module-level comment block preceding `_single_blocker_issue` is quite long for a function that's 3 lines. The critical why (epic decomposer emits prose, old code silently dropped chains, observed incident) is worth keeping, but the multi-paragraph elaboration of every accepted/rejected case duplicates the docstring and tests. Could be trimmed to 3–4 lines without losing the non-obvious invariant.

### Praise

- Replacing a strict whole-field regex with a ref-count check is a better abstraction: it's not "is it exactly `#NNN`?" but "does it reference exactly one issue?" — which is the actual semantic requirement. More robust and easier to reason about.
- The PR body explicitly flags the behavior reversal and the moved test case — that context is exactly what a future reader needs when the test name surprises them.

- [x] `python3 scripts/fleet/tests/test_enrich_stackable_blocker_prs.py` — 14 pass (verified in review)
- [ ] Confirm a live scout tick after merge enriches a prose-suffix task (can be verified against the next `state.json` snapshot when a stack chain is active)

🤖 Reviewed by Claude Sonnet 4.6 (1M context) (sonnet-reviewer role)

Opus recheck not required.

jakildev:
Addressed nit: trimmed the `_single_blocker_issue` comment block from 12 lines to 5 — kept the non-obvious invariants (epic decomposer prose suffix, observed 2026-05-28 chain-stacking failure, post-resolve_blocked_by() evaluation context), dropped the enumeration of accepted/rejected cases that duplicates the docstring and tests.

Issue #1326, opened 2026-05-28
fleet: close gaps in design-escalation handling surfaced by fleet feedback


jakildev:
## Context

Surfaced by `scripts/fleet/review-fleet-feedback` from aggregated fleet feedback.
Signature: `design-escalation-gap` · 2 occurrences since 2026-05-28T11:41:00.
Tracked locally in `~/.fleet/feedback/.fix-log.jsonl::fix-012` (proposed 2026-05-28T21:43:13).

## Representative feedback entries

- 2026-04-30 17:53 [opus-worker-1] PR #378 (T-063) is structurally blocked on T-062 PR #351 merging — fleet has no signal for "fork-of-other-PR's-branch"
- 2026-05-21 [opus-worker-2] fleet-state inconsistency blocks T-300/T-299 stack-claim
- 2026-05-28 18:30 [opus-worker-1] re-escalated PR left labeled fleet:design-unblocked (#1303)

## Proposed fix

Two scout/escalation gaps (opus-worker-1): (a) #1301 surfaced as owner=free/blocked_by=(none) because its dependency lived in a free-form 'Blocked on #1300' HEADER, not the **Blocked by:** field the parsers read -- extend the blocked-by parsers in scripts/fleet/fleet-state-scout (~L332), fleet-queue-ingest (BLOCKED_RE ~L180), and fleet-claim (field_re ~L374) to also scan the body for 'Blocked on #N'/'Blocked on PR-x' header prose (or normalize it into blocked_by). #1296 covers #N refs INSIDE blocked_by, not header prose. (b) #1303 was re-escalated but kept fleet:design-unblocked -- the worker re-escalation flow in role-opus-worker.md / role-*.md should swap fleet:design-unblocked -> fleet:design-blocked (mirror of step 8c) so a re-blocked PR isn't picked up as unblocked.

## Acceptance

- Implementation matches the artifact called out in the proposed fix above.
- Subsequent `review-fleet-feedback digest` runs do not re-surface the `design-escalation-gap` cluster after the fix lands (verified via the 7-day quiet window).

---

_Generated by `scripts/fleet/review-fleet-feedback file-tasks`. When this issue's PR merges with a `Closes #<N>` line, the fix-log row's `status` will flip to `applied` automatically on the next `digest` run._

jakildev:
## Status: part (a) done + part (b) docs done; part (b) role-command edit gated

[...]

Issue #1325, opened 2026-05-28
fleet: prevent worktree misroute on absolute-path edits to parent clone


jakildev:
## Context

Surfaced by `scripts/fleet/review-fleet-feedback` from aggregated fleet feedback.
Signature: `worktree-misroute` · 5 occurrences since 2026-05-27T00:00:00.
Tracked locally in `~/.fleet/feedback/.fix-log.jsonl::fix-011` (proposed 2026-05-28T21:43:13).

## Representative feedback entries

- 2026-04-30 18:14 [opus-worker-2] Edit tool with absolute-path leading to parent clone silently writes to wrong worktree.
- 2026-05-23 [opus-worker-1] Write tool with main-worktree absolute path silently lands in wrong worktree
- 2026-05-28 [opus-worker-2] worked in the shared MAIN clone instead of my worktree (caused a cross-agent commit collision)

## Proposed fix

Agents edit/write into the main repo or a parent clone instead of their own worktree: git -C <main-repo-path> trap in a sub-worktree, Write with an absolute path landing in the main clone, 'worked in the shared MAIN clone instead of my worktree' (cross-agent contamination), feedback/smoke detached-checkout silently targeting the cwd repo (opus-worker-1 + opus-worker-2, 5x in 2 days). Fix: add a worktree-identity preflight that asserts `git rev-parse --show-toplevel` == the agent's assigned worktree before Edit/Write/git -C, wired into start-next-task + commit-and-push (and the feedback/smoke detached checkout path), plus a warning in docs/agents/FLEET.md and role-*.md that absolute paths under the main clone (~/src/IrredenEngine) bypass the worktree. NOTE: the #1259 ref in one entry is unrelated (a demo issue).

## Acceptance

- Implementation matches the artifact called out in the proposed fix above.
- Subsequent `review-fleet-feedback digest` runs do not re-surface the `worktree-misroute` cluster after the fix lands (verified via the 7-day quiet window).

---

_Generated by `scripts/fleet/review-fleet-feedback file-tasks`. When this issue's PR merges with a `Closes #<N>` line, the fix-log row's `status` will flip to `applied` automatically on the next `digest` run._

Issue #1324, opened 2026-05-28
fleet/simplify-reuse-coverage: address recurring feedback


jakildev:
## Context

Surfaced by `scripts/fleet/review-fleet-feedback` from aggregated fleet feedback.
Signature: `simplify-reuse-coverage` · 2 occurrences since 2026-05-28T00:00:00.
Tracked locally in `~/.fleet/feedback/.fix-log.jsonl::fix-010` (proposed 2026-05-28T21:43:13).

## Representative feedback entries

- 2026-05-28 [opus-worker-1] simplify reuse subagents miss uncommitted working-tree changes (PR #1293 / issue #1288)
- 2026-05-28 [opus-worker-1] simplify reuse subagent silently no-op'd on new files

## Proposed fix

simplify's reuse fan-out subagents (.claude/agents/simplify-grep-function-names.md, simplify-grep-utility-candidates.md, simplify-scan-loop-patterns.md, simplify-scan-call-sequence-dup.md, simplify-scan-render-leak.md; tools: Read,Grep,Glob) read committed/origin-master state instead of the dirty working tree, conclude 'files do not exist', and return zero findings -> silent no-coverage on the exact pre-commit diff they exist to scan (opus-worker-1, PR #1293/#1288 and #1308). Fix: the reuse-dispatch step in .claude/skills/simplify/SKILL.md must hand each subagent the explicit changed-path list (git diff --name-only HEAD + git ls-files --others --exclude-standard), and each agent briefing must Read those cited paths directly before concluding 'not found' -- never emit zero findings solely because a Grep/Glob missed. simplify-scan-call-sequence-dup (Sonnet) saw the files fine both times, so this is a briefing/file-read-path bug, not pure model capability.

## Acceptance

- Implementation matches the artifact called out in the proposed fix above.
- Subsequent `review-fleet-feedback digest` runs do not re-surface the `simplify-reuse-coverage` cluster after the fix lands (verified via the 7-day quiet window).

---

_Generated by `scripts/fleet/review-fleet-feedback file-tasks`. When this issue's PR merges with a `Closes #<N>` line, the fix-log row's `status` will flip to `applied` automatically on the next `digest` run._

PR #1328, opened 2026-05-28
fleet: fix review-fleet-feedback parser dropping ~40% of entries

jakildev:
## Summary
- The closed-loop feedback reviewer (`scripts/fleet/review-fleet-feedback`) required a bare `## YYYY-MM-DD HH:MM` header and **silently dropped 108 of 270 entries (~40%)** across the role files: every date-only header, every `Z`/`UTC`-suffixed timestamp, and every inline-em-dash headline.
- Net effect: in a 64h window the digest saw **6 entries when there were actually 44**, so recurring patterns never clustered and the closed loop never closed on them.
- Widen `ENTRY_HEAD_FILE` to a strict superset (date + optional time, optional `Z`/`UTC`, optional inline headline), anchor date-only headers to midnight UTC, and use the captured inline headline as the entry headline.
- Add a `simplify-reuse-coverage` signature (reuse subagents reading committed/origin state instead of the dirty working tree → zero findings) and widen `worktree-misroute` (agents editing the main repo / parent clone instead of their own worktree), removing the old narrow duplicate.

## Validation
- Regex is a **strict superset**: 162 → 270 header matches on the live feedback files, **zero regressions** on previously-parsed entries.
- New/widened signatures verified against the target entries with false-positive controls (format-overreach, stale-task, unrelated worktree mentions all correctly excluded).
- Re-running the digest after the fix surfaced 3 real clusters that were previously invisible as `uncategorized` singletons (filed as #1324 / #1325 / #1326).

## Test plan
- [ ] `scripts/fleet/review-fleet-feedback digest` emits valid JSON
- [ ] `scripts/fleet/review-fleet-feedback apply --schema` prints the schema
- [ ] Spot-check that a date-only `## YYYY-MM-DD — headline` entry now appears in the digest

### DIFF
--- a/scripts/fleet/review-fleet-feedback
+++ b/scripts/fleet/review-fleet-feedback
@@ -48,7 +48,21 @@ MAX_SINCE_HOURS = 24 * 90         # cap stale markers at 90d
 AUTO_CLOSE_DAYS = 7               # quiet days before auto-close
 MIN_CLUSTER_SIZE_FOR_PROPOSAL = 2 # singletons never get a proposal
 
-ENTRY_HEAD_FILE = re.compile(r"^## (\d{4}-\d{2}-\d{2}[ T]\d{2}:\d{2}(?::\d{2})?)$")
+# Entry headers come in several shapes across the role files:
+#   ## 2026-05-26 14:15            (strict: date + time, nothing trailing)
+#   ## 2026-05-08T17:34Z — headline (ISO-T, Z/UTC suffix, inline headline)
+#   ## 2026-05-13 — headline        (date only, inline headline)
+#   ## 2026-05-19 opus-worker-1: …  (date only, inline headline w/o em-dash)
+# Group 1 is the timestamp (date, optional time); group 2 is the optional
+# inline headline. Anything that doesn't lead with a date is a prose section
+# divider and is ignored. Keep this a strict superset of the old date+time
+# form so previously-parsed entries are unaffected.
+ENTRY_HEAD_FILE = re.compile(
+    r"^##\s+"
+    r"(\d{4}-\d{2}-\d{2}(?:[ T]\d{2}:\d{2}(?::\d{2})?)?)"   # date, optional time
+    r"(?:\s*(?:Z|UTC))?"                                      # optional Z / UTC suffix
+    r"(?:\s*[—–:-]+\s*(.*?))?\s*$"                            # optional inline headline
+)
 
 # Pattern bank — order matters; first match wins. Widen aggressively;
 # uncategorized noise drowns the signal otherwise.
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — #1328: fleet: fix review-fleet-feedback parser dropping ~40% of entries

**Verdict:** approve

### Summary

Correct diagnosis and fix for a silent data-loss bug in the fleet feedback
parser. The old `ENTRY_HEAD_FILE` required a bare `## YYYY-MM-DD HH:MM` header
with no suffix, silently discarding date-only headers, Z/UTC timestamps, and
any em-dash inline headlines — 108 of 270 entries (~40%). The new regex is a
validated strict superset (162 → 270 matches, zero regressions). The two new /
widened signatures (`worktree-misroute`, `simplify-reuse-coverage`) and the
position swap that groups worktree patterns together are also clean.

### What is correct

- **Regex is a strict superset.** All four documented input shapes (`date+time`,
  `ISO-T+Z/UTC`, `date-only+em-dash headline`, plain `date-only`) are now
  handled. The strict superset claim holds — the old `YYYY-MM-DD HH:MM` form
  still matches group 1 with no group 2, unchanged.
- **`Entry.ts` + `Entry.dt` stay in sync.** Date-only headers (`2026-05-19`)
  produce `ts = "2026-05-19"` via `m.group(1)`; the new `%Y-%m-%d` strptime
  format anchors those to midnight UTC correctly.
- **Inline headline capture is correct.** When group 2 is non-empty the
  headline is set on entry creation rather than relying on the first non-empty
  body line fallback. The fallback still fires for entries with no inline
  headline, so existing behaviour is preserved.
- **Lookahead signatures use `re.I | re.S` appropriately.** `signature_for`
  concatenates `headline + "\n" + first_body`, so DOTALL crosses the separator
  and allows the two required tokens to appear on different sides of the
  newline. The AND semantics (both lookaheads must fire) are correct.
- **Old narrow `worktree-misroute` removed without regression.** The new
  lookahead form is a superset of the literal alternations it replaces.

### Nits

- `scripts/fleet/review-fleet-feedback:54-55` — the code comment lists
  `## 2026-05-19 opus-worker-1: …` (content after the date with no separator
  character) as a handled format, but the regex does not match it. Verified
  empirically: `ENTRY_HEAD_FILE.match("## 2026-05-19 opus-worker-1: something")`
  returns `None`. Either drop that example from the comment or widen the regex
  to handle it (e.g. add a bare-word fallback like `|\s+(\S.*?)` after the
  separator group). If entries in this shape don't appear in the live files
  the omission is harmless, but the comment creates false confidence.

### Praise

- The strict-superset framing (test old entries, confirm zero regressions,
  then count new matches) is exactly the right way to validate a parser
  widening. The 162 → 270 figure gives future maintainers a concrete baseline.
- Filing #1324 / #1325 / #1326 from the newly visible clusters immediately
  is the correct closed-loop action — the fix is already paying dividends.

- [ ] `scripts/fleet/review-fleet-feedback digest` emits valid JSON with
  `total_entries` ≥ 270 against the live feedback files
- [ ] Spot-check a date-only `## YYYY-MM-DD — headline` entry appears in
  the digest output (not in `singletons` as uncategorized)
- [ ] `scripts/fleet/review-fleet-feedback apply --schema` prints the schema

🤖 Reviewed by Claude Sonnet 4.6 (1M context) (sonnet-reviewer role)

Opus recheck not required.

jakildev:
Addressed nit: widened `ENTRY_HEAD_FILE` regex to also match bare-word inline headlines (e.g. `## 2026-05-19 opus-worker-1: …`, `## 2026-05-23 iteration`). Old pattern only accepted entries starting with a separator char (`—–:-`); new pattern adds `|\s+` as a fallback. Verified: 279 entries now parsed from live feedback files (up from 270), all 7 test shapes pass.

jakildev:
Addressed nit (scripts/fleet/review-fleet-feedback:64):

- Made the separator in `ENTRY_HEAD_FILE` optional: `(?:[—–:-]+\s*)?` instead of required `[—–:-]+\s*`. Entries like `## 2026-05-19 opus-worker-1: headline` and `## 2026-05-23 iteration` now match correctly — verified against `opus-worker-1.md` (84/84 headers matched vs prior misses).

The comment at lines 51-55 already documented this format accurately; no comment change needed.

— sonnet-fleet-2

PR #1337, opened 2026-05-29
fleet: worker escalation/handoff hardening — design-block releases ownership, headless never asks (closes #1326)

jakildev:
## Context

A recurring pattern: a headless worker hits something it can't resolve and tries to **ask interactively** (`AskUserQuestion`) — but workers run as dispatched `claude --print` one-shots with no human attached to the pane, so the question stalls or burns the iteration. The triggering case here (#1326) was sharper: the worker knew the exact edit but was blocked by the **self-modification gate** (editing its own role config from an untrusted-queue task), so every worker iteration re-claimed #1326 and walled.

## Changes

**`role-opus-worker.md` §8 (stop-and-escalate):**
- New escalation trigger: a task that can only be completed by editing fleet self-config the agent loads (`.claude/commands/role-*.md`, `.claude/agents/*`) — gated as self-modification, deterministically un-appliable by a queue worker.
- New directive: **never ask the human interactively** (headless = nowhere to land). Resolve from the plan/issue, or escalate **async** on the issue/PR. For a step it can't perform, comment what a human must apply, drop `human:approved` + `fleet:queued` so it leaves autonomous pickup, release the claim, and don't re-claim.

**`role-opus-architect.md` (filing):** don't file self-config edits as `human:approved` tasks — apply them in the interactive session or hand to the human. Prevents the re-pickup loop at the source.

**Closes #1326:** lands the prepared step-8c edit — on a *re-escalation*, clear `fleet:design-unblocked` before adding `fleet:design-blocked` (via `fleet-pr-clear-feedback-labels`) so the PR doesn't carry both labels and get re-picked as unblocked. This is the verified, human-authored edit the self-mod gate blocked workers from applying; applied here with explicit human authorization.

## Note / possible follow-ups

- The no-interactive principle applies to **all** headless roles; this PR codifies it in the worker doc (the incident role). Propagating the one-liner to `role-sonnet-reviewer.md` / `role-sonnet-author.md` / `role-smoke-worker.md` is a small follow-up.
- A harder enforcement would be removing `AskUserQuestion` from headless roles' tool set at dispatch — worth considering if the doc directive isn't enough.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1338, opened 2026-05-29
fleet: pin amend-push force-with-lease to checkout base SHA (anti-clobber)

jakildev:
## Problem

`fleet-pr-amend-push` already used `git push --force-with-lease` — but **bare**, with no explicit expected SHA. A bare lease checks against the remote-tracking ref `origin/<head>`, and **any `git fetch` between checkout and push refreshes that ref** to whatever a concurrent agent just pushed. So the lease passes and clobbers their commit.

That's the **#1310 dual-worker clobber** (2026-05-29): both opus-workers resumed the same design-unblocked PR; worker-1 pushed `f39c8d49`, worker-2 fetched + amend-pushed `6557d097` straight over it. No functional code lost that time, but the mechanism is a live data-loss path.

## Fix

- `fleet-pr-checkout-detached` records the **checkout-time base SHA** on line 2 of the `.git/fleet-amend-ref` sentinel.
- `fleet-pr-amend-push` leases against that exact SHA: `--force-with-lease=<head>:<base>`. If `origin/<head>` moved since checkout, the push is **REFUSED** (with a "re-checkout, re-apply, push again" message) instead of clobbering.
- Legacy single-line sentinels (mid-flight checkouts) fall back to the bare lease + a warning.

## Scope

This is **D** of the #1310 escalation hardening. **A + B** (design-block releases ownership; design-unblocked acquires the atomic mutex claim + reads the architect plan file) are in **#1337**. **C** (issue/PR/claim/reservation reconciliation) is a tracked follow-up. Belt-and-suspenders: even if the claim-mutex fix in #1337 misses a race, the push itself can no longer clobber.

`bash -n` clean on both scripts.

### DIFF
--- a/scripts/fleet/fleet-pr-amend-push
+++ b/scripts/fleet/fleet-pr-amend-push
@@ -63,7 +63,10 @@ if [[ ! -f "$sentinel" ]]; then
     exit 1
 fi
 
-head_ref=$(cat "$sentinel")
+# Sentinel is two lines: head ref (line 1) + checkout-time base SHA (line 2;
+# absent on legacy sentinels written before the base-sha lease landed). The
+# base SHA is what makes the lease safe — see the push below.
+{ read -r head_ref; read -r base_sha; } < "$sentinel" 2>/dev/null || true
 if [[ -z "$head_ref" ]]; then
     echo "fleet-pr-amend-push: sentinel $sentinel is empty" >&2
     exit 1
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: pin amend-push force-with-lease to checkout base SHA (anti-clobber)

**Verdict:** approve

### Summary

Two-script fleet infra fix addressing the `--force-with-lease` bare-lease vulnerability. The mechanism is correct and the legacy fallback is handled cleanly.

**Checklist**

**Core lease fix (fleet-pr-checkout-detached:155–156, fleet-pr-amend-push:95–100):**
- `base_sha=$(git rev-parse HEAD)` captures the SHA immediately after `git checkout --detach origin/$head_ref` — before any subsequent fetch could refresh the remote-tracking ref. This is the right capture point.
- `--force-with-lease=$head_ref:$base_sha` pins the lease against the literal SHA rather than `refs/remotes/origin/$head_ref`, which is exactly the fix needed. If a concurrent agent pushes between checkout and push, the remote ref no longer matches `$base_sha` → push REFUSED, not clobbered.

**Read logic (fleet-pr-amend-push:69):**
- `{ read -r head_ref; read -r base_sha; } < "$sentinel" 2>/dev/null || true` — on a legacy single-line sentinel, the second `read` returns EOF and sets `base_sha=""` (bash `read` always assigns, even on EOF). With `set -euo pipefail`, `base_sha` is an assigned empty string (not unbound), so `[[ -n "$base_sha" ]]` at line 95 correctly selects the bare-lease fallback. No nounset trap.

**`$lease` as a single arg (fleet-pr-amend-push:101):**
- `git push "$lease" origin "HEAD:$head_ref"` — passing `$lease` quoted as a single token is correct for both the `--force-with-lease=ref:sha` form and the bare `--force-with-lease` form. Shell quoting strips before git sees it; git recognises both as the same option.

**Legacy path:** Warning on stderr is appropriate and actionable. The re-checkout instruction is correct.

**Output cleanup (fleet-pr-checkout-detached:158–159):**
- New two-liner output embeds the abbreviated base SHA twice for quick human verification — cleaner than the old three-liner which gave a raw git command. No functionality removed.

### Nits

- `fleet-pr-amend-push:102` — the "push REFUSED" error message fires for any push failure (network timeout, auth rejection, etc.) with wording implying "another agent pushed". Technically correct for the SHA-mismatch case but potentially misleading for unrelated failures. Low frequency, low consequence in practice.

### Praise

- The `--force-with-lease=<ref>:<sha>` pin-to-SHA approach is the right idiom for this race — bare `--force-with-lease` only guards against "no fetch at all," not "fetch-then-clobber". Worth preserving this as the standard pattern.
- Legacy-sentinel fallback with a visible stderr warning, rather than silent failure or hard error, is the right tradeoff for mid-flight checkouts.

- [x] `bash -n scripts/fleet/fleet-pr-amend-push` — mentioned in PR body as clean.
- [x] `bash -n scripts/fleet/fleet-pr-checkout-detached` — mentioned in PR body as clean.
- [ ] Simulate: run `fleet-pr-checkout-detached <N>`, inspect `.git/fleet-amend-ref` for two-line output with valid SHA on line 2, then run `fleet-pr-amend-push` in a non-conflicting state and confirm sentinel is removed on success.
- [ ] Simulate legacy path: write a one-line sentinel manually, confirm `fleet-pr-amend-push` emits the warning and falls back to bare lease.

Opus recheck not required.

PR #1336, opened 2026-05-29
render: smooth camera Z-yaw — T3: trixel→framebuffer depth composite (#1310)

jakildev:
T3 of the per-axis trixel canvas epic (#1307). Composites the three per-axis voxel canvases (populated by T2 #1309) into the framebuffer by depth, so camera Z-yaw interpolates smoothly between cardinals instead of snapping.

Implements the architect's **forward-scatter (Option 4)** decision (`docs/design/per-axis-trixel-canvas-rotation.md`): per-axis stage-1/2 store one cell per visible face center (`atomicMin` occlusion winner per cell); a new instanced scatter pass (`v_/f_peraxis_scatter` + Metal mirror) forward-projects each non-empty cell's true deformed face quad, recovering the world origin via the closed-form iso inverse. The hardware depth test composites the three canvases. No gather inverse ⇒ the #1256 stripe/checkerboard class cannot occur by construction. Cull runs in yawed space (chunk + voxel granularity, gated so cardinal is untouched).

Closes #1310

## Verification (Metal)

- **Perf gate (the architect's hard gate) — PASSED.** Camera *parked* at a cardinal vs. parked at 30° (sustained steady state) at high zoom: composite overhead ~+0.075 ms/frame, frame time vsync-capped in both, voxel cull stats identical across yaw — a bounded constant factor, not a per-pixel-per-voxel cliff. Full table in the comment thread.
- **`optimize` — no actionable hotspot.** The empty-cell scatter degenerates after a single `texelFetch` + branch; the compute-compaction pre-pass stays correctly deferred (measured, not assumed).
- **#1256 parity gate — zero stripe/checkerboard/lattice at every yaw** (by construction).
- **Cardinal fast path unchanged** — `residualYaw == 0` releases the per-axis canvases → single-canvas gather path. (Below: cardinal/yaw=0 pairs are visually identical; the sub-pixel zoom4 jitter is run-to-run GPU rasterization noise, confirmed identical in a same-binary run1-vs-run2 control.)

## Screenshots

12-shot `IRShapeDebug` sweep, `origin/master` (before) vs this branch (after). The **`zoom4_yaw45_inter_cardinal`** pair is the feature: voxel cubes deform smoothly through the inter-cardinal angle (after) instead of cardinal-snapping (before). SDF shapes stay cardinal-snapped — the documented v1 deferral.

<details open>
<summary>zoom4_yaw45_inter_cardinal — THE FEATURE (smooth voxel deform vs. cardinal snap)</summary>

[...]

PR #1340, opened 2026-05-29
fleet: amend-push fast-forwards with a plain push (ungated) when possible

jakildev:
## Context

Follow-on to #1338 (which pinned `--force-with-lease` to the checkout base SHA). PR #1336 feedback also flagged: *"when the commit's parent IS the remote tip (a fast-forward), a plain `git push origin HEAD:<branch>` is the correct, ungated way to land it — force-push over another contributor's commits is classifier-gated, rightly."*

## Change

`fleet-pr-amend-push` now refreshes the remote tip and, **if HEAD descends from it** (a true fast-forward — the clean common case of checkout → commit on top → push, and a repeat amend once the prior push became the tip), lands it with a **plain `git push`**: no `--force`, not classifier-gated, nothing to clobber. Only a genuine non-fast-forward amendment (rebase/replace) takes the `--force-with-lease=<head>:<base>` path from #1338.

## Why it's safe

Anti-clobber is unchanged. `merge-base --is-ancestor` is robust to the intervening fetch, and the non-FF path still leases against the **sentinel's** base SHA (which a fetch can't refresh). In the #1310 clobber, worker-2's HEAD did **not** descend from worker-1's pushed commit, so the FF path is skipped and the lease-and-refuse correctly fires — no regression.

`bash -n` clean.

### DIFF
--- a/scripts/fleet/fleet-pr-amend-push
+++ b/scripts/fleet/fleet-pr-amend-push
@@ -85,7 +85,28 @@ if git symbolic-ref -q HEAD >/dev/null 2>&1; then
     exit 1
 fi
 
-# Push with force-with-lease pinned to the CHECKOUT-TIME base SHA, not the bare
+# Fast-forward fast path. Refresh the remote tip; if HEAD descends from it
+# (the clean common case: detached checkout -> commit on top -> push, and a
+# follow-up amend once the prior push became the tip), land it with a PLAIN
+# push — no force, so it is not classifier-gated and there is nothing to
+# clobber. `merge-base --is-ancestor` is robust to the fetch; the non-FF path
+# below leases against the sentinel's base SHA, which the fetch can't refresh.
+# (In the #1310 clobber, HEAD did NOT descend from the other worker's push, so
+# this path is skipped and the lease-and-refuse below correctly fires.)
+git fetch origin "$head_ref" >/dev/null 2>&1 || true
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — fleet: amend-push fast-forwards with a plain push (ungated) when possible

**Verdict:** approve

### Praise
- The key correctness insight is sound: a plain `git push` (no force) is self-protecting because git rejects non-fast-forward pushes at the remote, so the `|| true` on the fetch cannot enable a clobber — worst case is a rejected push. This removes an unnecessary force-push and classifier gate from the hot path without weakening the anti-clobber guarantee.
- The `--is-ancestor remote_tip HEAD` condition is exactly the right test (exit 0 iff HEAD descends from the remote tip). Robust to stale data after a failed fetch.
- Edge cases are all handled gracefully: empty `remote_tip` (guard on line 98 skips FF), HEAD == remote_tip (no-op push, sentinel cleaned, exit 0), race between fetch and push (push rejected by git non-FF rule, exit 1).
- Non-FF path from #1338 is fully preserved; the sentinel format and base SHA lease are untouched.

- [x] `bash -n scripts/fleet/fleet-pr-amend-push` syntax-check (PR body confirms clean)
- [x] Hot path: checkout-detached → commit → amend-push — confirm plain push path taken, sentinel removed
- [x] Regression: simulate two-worker scenario from #1310 — second worker's HEAD does not descend from first worker's push → FF path skipped → force-with-lease refused. Anti-clobber holds.

Opus recheck not required.

PR #1332, opened 2026-05-28
fleet: prevent worktree misroute on absolute-path edits to parent clone (#1325)

jakildev:
## Summary
- New `scripts/fleet/fleet-assert-worktree`: refuses (exit 1) when `git rev-parse --show-toplevel` is not a `.claude/worktrees/*` path. Optional expected-basename arg adds a cross-agent check; `FLEET_ALLOW_MAIN_CLONE=1` is a human-only override.
- Wired as a preflight into the three mutation points where a wrong-cwd op contaminates the shared main clone:
  - `commit-and-push` + `start-next-task` preconditions (skill docs)
  - `fleet-pr-checkout-detached` — the feedback / smoke / semantic-conflict detached-checkout path; guards before the HEAD-rewinding checkout, with an inline structural fallback if the helper isn't symlinked yet.
- Registered in `install.sh` + `fleet-help`; gotcha documented in `FLEET.md` (new "Worktree identity" subsection) and `CLAUDE-BASELINE.md` (extends the existing Edit/Write-path rule to git mutations).

**Dogfood note:** while implementing this I hit the exact bug 3× — a `Write` and an `Edit` both landed in the main clone via parent-clone absolute paths (reverted, re-applied in-worktree). Live evidence the guard is warranted. A bash preflight can't intercept the harness Edit/Write tool, so that vector stays a docs-warning, matching the issue's proposed-fix scope (preflight at the mutating git flows + docs).

## Test plan
- [x] `bash -n` clean: fleet-assert-worktree, fleet-pr-checkout-detached, install.sh, fleet-help
- [x] fleet-assert-worktree: exit 0 in worktree; exit 1 from main clone and on wrong basename; exit 0 with `FLEET_ALLOW_MAIN_CLONE=1`; exit 2 on bad arg / non-git tree
- [x] fleet-pr-checkout-detached refuses from main clone *before* any `gh` call; passes from worktree
- [x] game worktree path (`creations/game/.claude/worktrees/*`) matches discriminator; game main clone rejected
- [x] `fleet-help fleet-assert-worktree` + index entry render correctly
- [ ] Re-run `install.sh` on each host to symlink the new helper onto PATH (standard for new fleet-* tools)
- [ ] (Acceptance, automated) `review-fleet-feedback digest` no longer re-surfaces the `worktree-misroute` cluster over the 7-day quiet window

Closes #1325

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1341, opened 2026-05-29
fleet: PreToolUse hook blocks Edit/Write outside the current worktree

jakildev:
## Problem (turning direction into tooling)

A fleet agent runs with cwd inside `.../.claude/worktrees/<name>/`. When it edits via a bare `~/src/IrredenEngine/engine/...` **absolute** path, the Read/Edit tools resolve it against the **main clone**, not the worktree — so the write silently lands in the shared checkout, `fleet-build` (run in the worktree) compiles nothing, and the agent wrongly concludes its fix *"had no effect."* Documented twice (PR #1310 / #1336).

The #1332 worktree-identity preflight guards **Bash/git**, not the **Read/Edit file tools** — so today this is pure (and repeatedly-ignored) discipline. This converts it to enforcement.

## Fix

`scripts/fleet/fleet-guard-worktree-edit` — a `PreToolUse` hook (matcher `Edit|Write|MultiEdit`) that **denies** an absolute `file_path` resolving outside the session's worktree root, with a corrective message. Wired in `.claude/settings.json`.

Deliberately narrow so it never wedges the fleet or false-blocks:
- **Only** constrains sessions whose cwd is inside a fleet worktree — the main clone and isolated `/tmp` worktrees are unaffected.
- Relative paths (resolved under the in-worktree cwd) are always allowed.
- Reads are **not** gated (mutations are the silent-misroute hazard).
- Any parse error / missing field **fails open** (a hook must never block legitimate work).

## Verification

Tested against 6 payloads:

| Scenario | Result |
|---|---|
| worktree session edits a file **in** its worktree | ✅ allow |
| worktree session edits the **main clone** (the bug) | ⛔ deny + message |
| worktree session edits a **sibling** worktree | ⛔ deny |
| main-clone session edits the main clone | ✅ allow |
| worktree session, **relative** path | ✅ allow |
| isolated `/tmp` worktree | ✅ allow |

Note: wiring it into `.claude/settings.json` is a self-config change; it's deliberate and scoped to mutations-outside-your-worktree.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -76,5 +76,18 @@
       "Bash(sudo:*)",
       "Bash(apt:*)", "Bash(apt-get:*)", "Bash(dpkg:*)"
     ]
+  },
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Edit|Write|MultiEdit",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR\"/scripts/fleet/fleet-guard-worktree-edit"
+          }
+        ]
+      }
+    ]
   }
 }
--- a/scripts/fleet/fleet-guard-worktree-edit
+++ b/scripts/fleet/fleet-guard-worktree-edit
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+# fleet-guard-worktree-edit — PreToolUse hook for Edit | Write | MultiEdit.
+#
+# Refuses an edit whose absolute file_path lands OUTSIDE the fleet worktree the
+# session is working in. A fleet agent runs with its cwd inside
+# `.../.claude/worktrees/<name>/`; if it edits via a bare
+# `~/src/IrredenEngine/engine/...` absolute path, the write silently hits the
+# shared MAIN clone instead of the worktree. `fleet-build` (run in the worktree)
+# then compiles nothing and the agent wrongly concludes its fix "had no effect"
+# (observed on PR #1310 / #1336). The #1332 worktree-identity preflight guards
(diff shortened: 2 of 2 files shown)


jakildev (commented):
## Review — fleet: PreToolUse hook blocks Edit/Write outside the current worktree

**Verdict:** approve

### Praise
- Fail-open design is exactly right for a hook: every parse error, missing field, or `jq` absence exits 0 (allow) so the guard can never wedge the fleet. Scope is deliberately narrow — non-worktree sessions, relative paths, and reads are all unconditionally passed through.
- The trailing-slash trick on line 54 (`case "$file/" in "$wt_root"/*)`) prevents a sibling worktree (`<name>-foo`) from prefix-matching `<name>` — a subtle hazard that would otherwise produce a false deny on a legitimate write.
- Two-tier denial (preferred JSON hookSpecificOutput form, then exit-2 fallback) is the correct pattern for PreToolUse hooks: the stdout JSON carries the deny signal, exit 0 signals hook health, and exit 2 is a safe fallback when `jq` is unavailable.
- The `wt_root` derivation with `%%` (longest-suffix strip) and `#` (shortest-prefix strip) correctly handles a cwd that is deeper than the worktree root (e.g. cwd inside a subdirectory), and the `[[ -z "$prefix" || -z "$wt_name" ]]` guard fails open on any malformed path.

- [x] `bash -n scripts/fleet/fleet-guard-worktree-edit` syntax-check passes
- [x] All 6 payload scenarios in the PR body pass (in-worktree allow, main-clone deny, sibling-worktree deny, main-clone session allow, relative path allow, /tmp worktree allow)
- [x] Confirm hook fires correctly by opening a worktree session and attempting an Edit with an absolute path to the main clone — should receive the corrective deny message

Opus recheck not required.

PR #1347, opened 2026-05-29
engine/prefabs: drop const from ANIMATION_COLOR component registration

jakildev:
## Summary
- `System<ANIMATION_COLOR>::create()` registered `const C_ActionAnimation` as a tracked component. `registerSystem<N, Components...>` builds a `std::vector<Component>` per archetype column, and `std::vector<const T>` is ill-formed (`std::allocator` rejects const element types), so any translation unit that instantiates `createSystem<ANIMATION_COLOR>()` failed to compile with `std::allocator does not support const types`.
- The breakage was **latent**: no engine target or test instantiates `ANIMATION_COLOR`, so the header parsed in isolation and CI stayed green. This fix drops the `const` from the template argument; the tick keeps `const C_ActionAnimation &anim`, so the system's read-only access to the component is unchanged.

## Test plan
- [x] Built a creation target that instantiates `createSystem<IRSystem::ANIMATION_COLOR>()` — compiles and links clean. (Previously failed at `std::vector<const IRComponents::C_ActionAnimation>` instantiation.)
- [ ] Follow-up: add a regression test that instantiates `createSystem<ANIMATION_COLOR>()` so a `const`-qualified registration cannot regress latently again.

### DIFF
--- a/engine/prefabs/irreden/update/systems/system_animation_color.hpp
+++ b/engine/prefabs/irreden/update/systems/system_animation_color.hpp
@@ -216,7 +216,7 @@ template <> struct System<ANIMATION_COLOR> {
         );
     }
 
-  private:
+private:
     // Per-tick caches replacing the prior function-local statics. Live
     // on the System<N> instance so the engine owns the lifetime and
     // multi-instance use cannot cross-talk. `endTick` clears both so
--- a/engine/prefabs/irreden/update/systems/system_animation_color.hpp
+++ b/engine/prefabs/irreden/update/systems/system_animation_color.hpp
@@ -211,14 +211,12 @@ template <> struct System<ANIMATION_COLOR> {
     }
 
     static SystemId create() {
-        return registerSystem<
-            ANIMATION_COLOR,
-            const C_ActionAnimation,
-            C_AnimColorState,
-            C_VoxelSetNew>("AnimationColor");
+        return registerSystem<ANIMATION_COLOR, C_ActionAnimation, C_AnimColorState, C_VoxelSetNew>(
+            "AnimationColor"
+        );
     }
 
-private:
+  private:
     // Per-tick caches replacing the prior function-local statics. Live
     // on the System<N> instance so the engine owns the lifetime and
     // multi-instance use cannot cross-talk. `endTick` clears both so


jakildev (commented):
## Review — engine/prefabs: drop const from ANIMATION_COLOR component registration (#1347)

**Verdict:** approve

### Praise
- Fix is minimal and exactly right: `registerSystem<>` builds a `std::vector<Component>` per archetype column, and `std::vector<const T>` is ill-formed because `std::allocator` requires a non-const, non-reference element type. Removing `const` from the template argument resolves the ill-formed instantiation.
- The tick signature correctly retains `const C_ActionAnimation &anim` — read-only access to the component is preserved via the parameter-level `const`, which is the right pattern in this ECS. The template argument level just controls which column vector is built.
- PR body explains the root cause and latency clearly: no engine target instantiated this system, so the bad template arg parsed in isolation without triggering the allocation failure. Good diagnosis.
- ECS discipline holds: `tick` receives all three components as parameters (no per-entity `getComponent`), `endTick` correctly clears the per-tick caches, and no structural mutations occur mid-iteration.

### Nits
- `engine/prefabs/irreden/update/systems/system_animation_color.hpp:219` — `private:` indentation changed from 0-space to 2-space as a side-effect of the reformat. Not wrong, but it's an accidental style drift outside the minimal scope of this fix. The rest of the codebase tends to keep access specifiers at 0-indent inside template specializations. Low priority.

- [x] Build a creation target that instantiates `createSystem<IRSystem::ANIMATION_COLOR>()` — compiles and links clean (confirmed in PR body)
- [ ] Follow-up (acknowledged in PR body, not blocking): add a regression test that instantiates `createSystem<ANIMATION_COLOR>()` so a `const`-qualified registration cannot regress latently again.

Opus recheck not required.

jakildev:
Addressed nit: restored `private:` to 0-indent in `System<ANIMATION_COLOR>`, matching the 0-indent convention on master before this PR touched the file. One-line change; build clean.

jakildev:
Addressed nit: restored `private:` to 0-indent inside the `System<ANIMATION_COLOR>` template specialization (was accidentally 2-spaced by the prior reformat). No logic changes. — sonnet-fleet-1

jakildev (commented):
## Opus re-review — #1347: drop `const` from ANIMATION_COLOR component registration

**Verdict:** approve

Standalone diff against `master` (single file, `engine/prefabs/irreden/update/systems/system_animation_color.hpp`). This lands in a `System<>` registration / access-classification path, so it warrants an Opus pass even though Sonnet's verdict was approve with no recheck flag. I traced `registerSystem` storage, the `SystemAccess` derivation, and the column-vector machinery to source. **The fix is correct and safe to merge**, but there's one non-obvious consequence Sonnet's rationale got backwards that's worth recording.

### The compile-break diagnosis is exactly right
Confirmed to source: each archetype column is an `IComponentDataImpl<Component>` whose storage is `std::vector<Component> dataVector` (`engine/entity/include/irreden/entity/i_component_data.hpp`). A `const` template arg would form `std::vector<const C_ActionAnimation>`, which is ill-formed — `std::allocator<const T>` is forbidden. Latency explanation holds: no engine target instantiates `createSystem<ANIMATION_COLOR>()`, so the bad arg parsed in isolation and CI stayed green. Dropping `const` from the template arg is the right way to make the column vector instantiable.

### Sonnet flagged the access semantics as "preserved" — on closer read I disagree with the *reason*, though not the verdict
Sonnet wrote: *"read-only access is preserved via the parameter-level `const`… the template argument level just controls which column vector is built."* That second half is not how this ECS derives access. `SystemAccess` is built by `applyComponent<T>` in `engine/system/include/irreden/system/system_access.hpp:221`, which keys on `std::is_const_v<T>` of the **`registerSystem` template argument** — not the `tick` parameter signature:

- `registerSystem<N, const C_Foo, …>` → `appendRead` (component lands in `reads_`)
- `registerSystem<N, C_Foo, …>` → `appendWrite` (component lands in `writes_`)

The `tick(const C_ActionAnimation&)` parameter const-ness is never consulted by the access derivation. So this change does have a real effect Sonnet's note implied it didn't: **`C_ActionAnimation` moves from `reads_` to `writes_`** for `ANIMATION_COLOR`. The tick remains genuinely read-only in behavior (const ref, no writes), but the *declared* access descriptor now claims a write.

### Why that reclassification is safe here (why this is still approve, not needs-fix)
1. **`ANIMATION_COLOR` is SERIAL.** The tick opens with `IR_ASSERT_MAIN_THREAD()` and the comment explicitly states the system stays SERIAL (it does foreign-entity `getComponentOptional` lookups against a not-yet-thread-safe EntityManager, T-225's surface). A SERIAL system is never placed in a `PARALLEL_FOR` co-schedule group, so the read-vs-write descriptor never gates an actual concurrency decision for it.
2. **The misclassification is over-conservative, never under-locking.** Declaring a write you don't perform can only *forbid* a co-schedule that would have been legal (missed parallelism) — it cannot permit a race. There is no safety regression in either direction.
3. **It matches the documented current default.** `system_access.hpp` itself comments at the const branch: *"Until creations land the const opt-in (T-222 follow-up), most existing systems will land in the writes set — that's the conservative-correct default."* This was apparently the lone const-in-pack registration in the tree; after this PR every system uses the writes-set default, which is the state the access trait already expects.

### The real long-term fix (out of scope here — flag for the T-222 follow-up)
The deeper issue is that the const-in-pack *read opt-in* is half-wired: the access side reads `is_const_v<T>` (supports it), but the storage side forms `std::vector<Component>` un-stripped (rejects it). So const-in-pack can never both compile *and* classify as read today. The correct resolution is to strip cv on the **storage** path (`std::remove_cv_t<Component>` when forming `IComponentDataImpl` / the column vector and in `getComponentData<Component>`) while leaving the access trait keyed on the raw `T`. Then `const C_ActionAnimation` would compile *and* land in `reads_`, unlocking the parallelization opt-in this system can't currently take. That's genuinely T-222 scope, not this targeted compile fix — removing `const` now is the right interim move.

When the author lands the acknowledged regression test (`createSystem<ANIMATION_COLOR>()` instantiation), consider also asserting the derived access set so a future re-introduction of read-classification (or a silent flip) is caught — but that's a nice-to-have, not a blocker.

### Confirming Sonnet's other calls
ECS discipline holds on the deep read: all three components arrive as `tick` parameters (no per-entity `getComponent` in the loop), `endTick` clears both per-tick caches (`m_colorTrackCache` / `m_clipCache`) so entity-id reuse across frames can't return stale pointers, and there are no structural mutations mid-iteration. The `private:` 0-indent nit Sonnet raised was already addressed (commit after the review). No new concerns — merging is fine.

PR #1329, opened 2026-05-28
fleet: atomic amend-claim so reviewers don't race fleet:needs-fix fixes

jakildev:
## Problem

When a worker addresses a `fleet:needs-fix` PR it removes the label to begin
fixing. But `fleet:needs-fix` is the PR's **only** verdict label — once it's
gone the PR is verdict-less, indistinguishable from a fresh reviewable PR,
until `fleet:changes-made` lands after the push. With nothing marking that
window, a reviewer poll can claim and review a diff that's being actively
rewritten.

**Observed on [#1316](https://github.com/jakildev/IrredenEngine/pull/1316), 2026-05-28:**

```
21:00:02  −fleet:needs-fix                      worker starts fixing
            ── PR now carries no verdict / skip label ──
21:05:13  +fleet:reviewing-mac-sonnet-reviewer  reviewer claims mid-fix
21:07:12  +fleet:changes-made                   worker finishes pushing
21:07:39  reviewed                              verdict on a diff that just changed underneath it
```

The human-feedback path already covered its window with `fleet:human-amending`
(a reviewer skip label). The fleet-feedback path had no equivalent.

## Fix

Mint `fleet:amending-<host>-<agent>` — the host-suffixed analogue, via the same
atomic lex-min claim primitive as `fleet:reviewing-`/`fleet:resolving-`, so it
**also** tie-breaks two workers racing the same flagged PR.

- **`fleet-claim`** — `amending-claim` / `amending-release` subcommands (thin
  wrappers over the shared `_cmd_pr_label_claim`/`_release`), dispatch,
  usage/help, and `cleanup --gh` sweeps stale `fleet:amending-*` on the same
  30-min TTL as the sibling claims.
- **`fleet-state-scout`** — `REVIEW_SKIP_PREFIXES` + `_review_skipped()` route
  all four reviewer projection/slice sites to skip any `fleet:amending-*` PR.
- **`FLEET-FEEDBACK-HANDLING`** — step b acquires the claim *before* clearing
  `fleet:needs-fix` (skip the PR if the claim is lost); step e adds
  `fleet:changes-made` **then** releases, so there's no label-less gap.
- **Docs / role / skill** — FLEET.md label dictionary, `fleet-labels` note,
  REVIEWER-PROTOCOL + both reviewer role-doc skip-lists, and a live bail-label
  TOCTOU guard in the `review-pr` skill.

Scoped to `fleet:needs-fix`: `fleet:has-nits` / `fleet:design-unblocked` keep
[...]

PR #1349, opened 2026-05-30
fleet: claim-first feedback handling — one atomic claim, all paths

jakildev:
## Problem

Feedback handling is the one pickup path whose atomic claim was added
**per-path and reactively** — and it still had a hole:

| Feedback path | Claim before this PR | Acquired |
|---|---|---|
| `fleet:needs-fix` | ✅ `fleet:amending-*` (added after #1316) | late (step b) |
| `fleet:design-unblocked` | ✅ `fleet:amending-*` (added after a #1310 clobber) | late (step b) — **and never released** in step e |
| `human:needs-fix` / `human:blocker` | ❌ only `fleet:human-amending` (static, idempotent) + a **TOCTOU** worktree reservation | late (step b) |
| `fleet:has-nits` | ❌ none | — |

`fleet:human-amending` is the same label for both racers (idempotent add → no mutex). `fleet-claim reserve` reads the holder then `O_EXCL`-creates the worker's *own* file — two different worktrees both pass the "no holder" check. And every guard was acquired **after** reading feedback + checkout.

**Result, observed on PR #1336 (2026-05-30):** two opus-workers both selected the same `human:needs-fix` PR on one dispatcher tick, both read the comments, both downloaded the screenshots, both started the AMEND — colliding on labels and reservations before either noticed (visible in the panes: one went ESCALATE, the other erroneously added `fleet:human-amending`, then had to back off and restore labels).

## Fix — one atomic claim, acquired first, for every path

The `fleet:amending-<host>-<agent>` primitive already exists (#1329) and is path-agnostic. This makes it **the single mutex for all feedback handling**, acquired as the very first action:

- **New Step a** — `fleet-claim amending-claim` before reading feedback, checking out, or touching any label, for *all* paths (`human:*`, `fleet:needs-fix`, `fleet:has-nits`, `fleet:design-unblocked`) and both AMEND and ESCALATE. Lose the lex-min → skip the PR, next candidate.
- **Step b** drops the per-path claim conditionals (universal now); releases the claim on checkout failure.
- **Step e** releases the claim **unconditionally, every path** — which also fixes a latent leak: `design-unblocked` acquired the claim but step e never released it (it lingered to the 30-min sweep).
- **ESCALATE** releases the claim at its terminal step.
- **Candidate filters** (protocol + role docs) skip PRs already carrying `fleet:amending-*`.

## Why this also fixes the "both workers invoked" waste

The dispatcher fans one trigger across **every** idle pane of a role (`find_idle_panes_for_role` is plural by design — that's how `opus-worker-1`/`-2` get concurrency). **Don't** narrow it to one pane — if that pane is busy/wedged the work goes unclaimed. The correct design is fan-out **+ claim-first**: the loser now exits in ~1s instead of burning a full iteration. This PR makes claim-first true for feedback (it was already true for task/review/conflict pickup).

## Scope

Doc/protocol-only. The `fleet:amending-` subcommands, the `REVIEW_SKIP_PREFIXES` reviewer-skip, and the `cleanup --gh` TTL sweep all already exist from #1329 — this changes only *which paths use the claim and when*. Touches `FLEET-FEEDBACK-HANDLING.md`, the `fleet:amending-` dictionary entry in `FLEET.md`, and the three authoring role docs' candidate filters.

After this, every PR/issue pickup point — task, review, smoke, conflict resolution, and now all feedback handling — is gated by an atomic, script-based lex-min claim.

### DIFF
--- a/.claude/commands/role-opus-architect.md
+++ b/.claude/commands/role-opus-architect.md
@@ -166,7 +166,10 @@ When you do pick a task:
    Re-Read `~/.fleet/state/state.json` if its contents are no
    longer in your conversation context. From
    `repos.engine.prs[]`, pick PRs whose `labels` array contains
-   any of `human:needs-fix`, `fleet:needs-fix`, `fleet:has-nits`.
+   any of `human:needs-fix`, `fleet:needs-fix`, `fleet:has-nits` —
+   but **skip any PR already carrying a `fleet:amending-*` label**
+   (another worker holds the atomic feedback claim; step a will
+   reject your claim anyway).
 
    Follow [`docs/agents/FLEET-FEEDBACK-HANDLING.md`](../../docs/agents/FLEET-FEEDBACK-HANDLING.md) —
    it owns the priority order, the AMEND-vs-ESCALATE decision (the
--- a/.claude/commands/role-opus-worker.md
+++ b/.claude/commands/role-opus-worker.md
@@ -202,7 +202,10 @@ Do the work, then exit cleanly:
    longer in your conversation context. From `repos.engine.prs[]`
    and `repos.game.prs[]`, pick PRs whose `labels` array contains
    any of `human:needs-fix`, `human:blocker`, `fleet:needs-fix`,
-   `fleet:has-nits`, `fleet:design-unblocked`.
+   `fleet:has-nits`, `fleet:design-unblocked` — but **skip any PR
+   already carrying a `fleet:amending-*` label** (another worker
+   holds the atomic feedback claim and is handling it; step a will
+   reject your claim anyway).
 
    Follow [`docs/agents/FLEET-FEEDBACK-HANDLING.md`](../../docs/agents/FLEET-FEEDBACK-HANDLING.md) —
    it owns the priority order, the detached-HEAD checkout flow,
--- a/.claude/commands/role-sonnet-author.md
+++ b/.claude/commands/role-sonnet-author.md
@@ -132,7 +132,10 @@ Each iteration:
    `~/.fleet/state/state.json` if its contents are no longer in your
    conversation context. From `repos.engine.prs[]`, pick PRs whose
    `labels` array contains any of `human:needs-fix`,
-   `human:blocker`, `fleet:needs-fix`, `fleet:has-nits`.
+   `human:blocker`, `fleet:needs-fix`, `fleet:has-nits` — but **skip
+   any PR already carrying a `fleet:amending-*` label** (another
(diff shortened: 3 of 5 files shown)


jakildev (commented):
## Review — fleet: claim-first feedback handling — one atomic claim, all paths (#1349)

**Verdict:** approve

### Praise
- The fix is precisely motivated: the PR body's TOCTOU table makes the pre-PR gaps clear — `human:needs-fix` had only a non-atomic `fleet:human-amending` label plus a racy `O_EXCL`-on-own-file worktree reservation, and `fleet:has-nits` had no mutex at all. Replacing all per-path bolt-ons with one universal `fleet:amending-*` claim acquired as the very first action is the correct design.
- The claim-before-everything ordering is correctly described in Step a and consistently enforced: "before reading the feedback, checking out, or touching any label." The loser-exits-in-~1s rationale is correct — without first-step claim acquisition, the loser burns a full iteration (comments fetch, screenshot download, checkout) before discovering the race.
- ESCALATE path: Step 5 (release claim) is correctly inserted before Step 6 (skip AMEND path). This fixes the latent `fleet:design-unblocked` leak where step b acquired the claim but step e never released it, leaving it to the 30-min TTL sweep.
- Step e ordering is correct: label-swap first, then claim release. The invariant "PR is never label-less between dropping the claim and re-entering the review queue" holds because `fleet:amending-*` keeps reviewers off until the claim releases, and `fleet:changes-made` (where applicable) lands before that release.
- Step b (checkout failure path) correctly adds `fleet-claim amending-release` before skipping, closing the third gap — previously the checkout failure path had no claim release, so a network failure left an orphaned `fleet:amending-*` until the 30-min sweep.
- All three authoring role docs (opus-architect, opus-worker, sonnet-author) get the consistent "skip any PR carrying `fleet:amending-*`" fast-path filter with the correct justification ("step a will reject your claim anyway").
- The FLEET.md dictionary entry accurately describes the label as the "single mutex for all feedback handling" and correctly cites the observed #1336 race as motivation.
- "Label cycles at a glance" section adds a clean framing: path-specific labels (`fleet:human-amending`, `fleet:changes-made`) are status signals, not concurrency guards — the claim is the guard.

- [x] Doc-only change — no build required (confirmed in PR body)
- [x] FLEET-FEEDBACK-HANDLING.md Step a is universal for all four feedback paths and both AMEND/ESCALATE dispositions (confirmed in new § "Step a" section)
- [ ] Verify `fleet-claim amending-claim` exits 1 correctly when another worker holds the claim — ensures fast-path loser behavior works as documented

Opus recheck not required.

PR #1360, opened 2026-05-30
render: smooth camera Z-yaw — T4 (#1311) trixel-level per-axis lighting on the composite

jakildev:
## T4 (#1311) — trixel-level per-axis lighting on the smooth camera Z-yaw composite

Implements the architect's NEEDS-DESIGN resolution: light the three per-axis
voxel canvases **at trixel resolution, per axis, before the forward-scatter
composite** — the framebuffer-MRT G-buffer / per-fragment approach from the
pre-implementation findings note is **rejected** (hard constraint: no
fragment-rate lighting).

Before T4, rotating voxels showed flat raw colour. Now they carry AO + sun
shadows + directional Lambert + the world light volume, matching the
cardinal-snapped look at the bracket endpoints.

### What changed
- **`C_PerAxisTrixelCanvases`** — adds `3× ao_` + `3× sunShadow_` RGBA8 textures
  with the same rotation-only lifecycle as the colour/distance/id triple.
- **`COMPUTE_VOXEL_AO` / `COMPUTE_SUN_SHADOW` / `LIGHTING_TO_TRIXEL`** — gain a
  per-axis dispatch loop (gated `entity == main && per-axis allocated`) that
  reconstructs world-pos **face-locally** via `perAxisCellToWorld3D` (the exact
  `faceOriginFromInPlane` integer inverse the T3 scatter uses — no singular
  screen→world iso-yawed inverse).
- **`BAKE_SUN_SHADOW_MAP`** — bakes the main canvas (SDF/text) **+ the 3 per-axis
  voxel canvases** into the **shared** world sun-shadow map (pre-composite
  `atomicMin`), so voxels and SDF cast/receive shadows on each other under rotation.
- The world sun-shadow SSBO + 128³ light volume stay **shared** (sampled by
  reconstructed world-pos); no per-axis copies.
- `perAxisCellToWorld3D` lives in a dedicated `ir_per_axis_lighting.{glsl,metal}`
  include (NOT `ir_iso_common`) so the SDF/voxel/scatter shaders that share
  `ir_iso_common` are byte-unchanged.
- Each per-axis loop **restores the main-canvas image bindings** on exit — the
  Metal image-binding table persists across frames, so leaving freed per-axis
  textures bound would dangle at the cardinal release (this was a reproducible
  mid-rotation crash; fixed).
- GLSL + Metal at parity; design doc §"Open decisions → Lighting / AO placement"
  updated with the resolved decision (per architect, in this PR).

### Verification (Metal / IRShapeDebug)
[...]

Issue #360, opened 2026-04-28
Camera-anchored occupancy + light-volume grids


jakildev:
**Follow-up from the lighting-fidelity-polish PR.**

## Summary

Make the 256³ occupancy grid and 128³ light volume follow the iso camera so they only cover the visible (+ shadow-relevant) world region instead of a fixed world-space window centered at the origin.

## Why

- Current grids are world-space, centered at the origin, fixed extent. Anything outside `±128` (occupancy) or `±64` (light volume) is silently clamped.
- A creation that pans the camera far from origin loses lighting fidelity past the static window.
- Camera-anchored grids cut both memory (~8× given typical view sizes) and per-frame work by the same factor since fewer cells are populated/read.
- Pairs naturally with the screen-space-shadow rewrite (#358) and the GPU light-volume rewrite — both already need to know the visible region; making the grid origin track the camera is a small additional change at that point.

## Mechanics

- Add a per-grid `worldOrigin` (snapped to a multiple of grid resolution to avoid sliding artifacts) carried alongside the grid in its component.
- Update producers (`BUILD_OCCUPANCY_GRID`, `COMPUTE_LIGHT_VOLUME`) to write into local-grid coords relative to the origin.
- Update consumers (`c_compute_sun_shadow.glsl`, `c_lighting_to_trixel.glsl`) to subtract the origin before indexing.
- When the camera pans far enough that the origin needs to move, treat it as a full grid invalidation for that frame (the GPU rewrites in #359/#358 make this cheap).

## Acceptance

- Lighting demos work identically when the camera is panned far from origin (e.g. `--zoom 4` with a translated camera) as when at origin.
- Occupancy / light-volume memory footprints are unchanged or smaller, but the *covered region* is camera-relative.

jakildev:
Needs planning: issue #360 was filed before the current occupancy-grid deletion plan was finalized. The deletion cascade (T-071 → T-091 → T-092) removes the occupancy grid entirely, making the 'camera-anchor the occupancy grid' portion of this issue moot. The light-volume half may still be valid as a follow-up after T-072 (GPU light-volume propagation) lands — T-072's notes explicitly say 'Camera-anchored grid follow-up deferred.' Tagging for architect review to narrow the scope: should this issue be re-scoped to camera-anchored GPU light-volume only (blocked by T-072)? Or is it superseded entirely by the GPU propagation approach?

jakildev:
## Plan — re-scope to camera-anchored light-volume only

[...]

PR #1362, opened 2026-05-30
render: camera Z-yaw pivots about the camera focus (#1352)

### LINKED ISSUE #1352: Camera Z-yaw rotates about world origin, not the camera focus — pivot should track pan (and optionally the cursor)
## Problem

Camera Z-yaw rotation always pivots around the **world origin (0,0,0)**. The pivot does **not** move when you pan the camera, so once you pan away from the origin and then rotate, the scene swings in a big arc around an off-screen point instead of spinning in place around what you're looking at. Surfaced while reviewing the smooth-yaw work in #1310 — the rotating cube only looks "centered" because it happens to sit near the world origin with the camera at (0,0).

Expected: the rotation should pivot around the point the camera is focused on (screen center), and ideally optionally around the cursor / click position.

## Current behavior & root cause

The yaw and the pan are applied as two independent transforms:

- **Yaw** rotates each world position about the world Z-axis **through the origin** — `pos3DtoPos2DIsoYawed(worldPos, yaw) = pos3DtoPos2DIso(R_z(-yaw)·worldPos)` (and the cardinal half via `rotateCardinalZ`). See `engine/render/src/shaders/ir_iso_common.glsl` + `c_voxel_to_trixel_stage_{1,2}`, `v_peraxis_scatter.glsl`, `c_shapes_to_trixel.glsl`.
- **Pan** is a separate 2D iso-space translation: `frameData.cameraTrixelOffset_ = IRRender::getCameraPosition2DIso()` (`engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp:84`; camera position stored as `C_Position2DIso`, see `render_manager.cpp`).

So the screen position of a world point `W` is `M · (pos3DtoPos2DIsoYawed(W, yaw) − cameraIso)`. The only **rotation-invariant** world point is `W = 0` (origin), because `pos3DtoPos2DIsoYawed(0, yaw) = 0` for all yaw. That origin is what everything visually rotates around. When the camera is panned (`cameraIso ≠ 0`) the origin is off-center, so the focused content orbits it.

## Desired behavior

Rotate about a **pivot `P`** that tracks the camera instead of the fixed origin. Because the projection is linear, pivoting about `P` is a per-frame **uniform iso offset** — no per-vertex math change:

```
screen(W) = M · ( pos3DtoPos2DIsoYawed(W, yaw) − cameraIsoEffective )
cameraIsoEffective = pos3DtoPos2DIsoYawed(P, yaw) + (cameraIso − pos3DtoPos2DIso(P))
```

For **pivot = camera focus**, pick `P = isoPixelToPos3D(cameraIso, 0)` (the un-yawed world point currently at screen center). Then `pos3DtoPos2DIso(P) = cameraIso`, and it collapses to:

```
cameraIsoEffective = pos3DtoPos2DIsoYawed(P, yaw)
```

i.e. keep the focus world point pinned and let the scene rotate around it. Note the correction is **independent of `P.z`** (the iso `+2z` term cancels in the difference), so `P` can be taken at `z = 0` — no depth ambiguity.

## Pivot options (the thing to decide)

- **(A) Camera center / focus — recommended default.** `P` = world point under screen center. Matches "orbit what I'm looking at"; most DCC/iso tools behave this way.
- **(B) Cursor / click position.** `P` = world point under the mouse — `IRRender::mouseWorldPos3DAtIsoDepth(canvasIsoDepth)` already exists. Good for "rotate around what I clicked"; latch the pivot on mouse-down so it doesn't drift mid-drag.
- **(C) Configurable.** A camera setting (e.g. `RotationPivotMode { ORIGIN, CAMERA_CENTER, CURSOR }`) so creations/editors choose. Keep `ORIGIN` available for back-compat / deterministic demos.

## Blast radius (apply the offset consistently or layers desync)

The same effective offset must feed **every** stage that mixes yaw with the camera iso offset, or the voxel / SDF / per-axis layers will slide relative to each other:

- `buildVoxelFrameData` — `cameraTrixelOffset_` (`system_voxel_to_trixel.hpp`).
- Per-axis scatter — `perAxisBase_` and the `faceLocalAnchor` it derives (`system_trixel_to_framebuffer.hpp` + `ir_iso_common`), from the #1310 smooth-yaw path.
- SDF overlay — `SHAPES_TO_TRIXEL` yaw/offset (`c_shapes_to_trixel.glsl/.metal`).
- `FRAMEBUFFER_TO_SCREEN` pan/zoom.
- Screen→world inverse for hover/picking (`mouseTrixelPositionWorld`, `mouseWorldPos3DAtIsoDepth`, `IRPrefab::Picking`) — must invert the **same** pivoted transform or clicks land on the wrong voxel while rotating.

A clean option worth weighing in planning: store the camera's pan as a **world pivot** and derive the iso offset once, rather than storing `C_Position2DIso` directly and correcting at every site.

## Interactions / constraints

- Keep the cardinal **fast path byte-identical** when not rotating (residualYaw == 0) — the offset correction is identity at yaw ≡ 0 (mod π/2) only if `P`'s correction is too; verify.
- Z-yaw only (the iso-depth-axis invariant holds); pitch/roll out of scope.
- Don't regress the #1310 smooth-yaw composite — its per-axis anchor must consume the same pivot.

## Acceptance criteria

- Pan to an arbitrary spot, rotate a full turn: the focused content stays put (option A) / the cursor-anchored point stays put (option B); no large orbital swing.
- All layers (voxel, SDF, per-axis smooth-yaw composite, overlays) stay registered across the full rotation at a panned camera.
- Hover/picking resolves the correct voxel while panned + rotating.
- Cardinal fast path unchanged; demos that rely on origin pivot still work (or are migrated).

## Open questions

1. Default pivot — camera center, or cursor? (Leaning camera center for the engine default; cursor for the editor.)
2. Store a world pivot on the camera vs. correct the iso offset at each site?
3. Should the pivot latch on mouse-down (cursor mode) or follow the live cursor?
4. Expose via a `C_Camera`-side setting + Lua binding, or engine config only?

Filed at the request of @evinjkill off the #1310 review.

jakildev:
## render: camera Z-yaw pivots about the camera focus (#1352)

Camera Z-yaw now pivots about the **camera focus** (the world point under
screen center) instead of the fixed world origin, so panning then rotating
spins the scene **in place** rather than swinging it in an arc around an
off-screen origin. `RotationPivotMode::CAMERA_CENTER` is the new engine
default; `ORIGIN` stays selectable for deterministic demos.

Resumed from `fleet:design-blocked` per the architect's design-unblocked
direction (`~/.fleet/plans/issue-1352.md` ⚑ ARCHITECT REVISION).

### What changed
- **Rebased** the foundation (`RotationPivotMode` + `getEffectiveCameraIso()`
  + float-precise `IRMath::isoPixelToPos3D(vec2,float)` + 3 wired sites) onto
  current master. One keep-both conflict resolved in `ir_render_types.hpp`
  (`RotationPivotMode` beside master's new `GpuVoxelTransform` block).
- **Routed the full producer set** through `getEffectiveCameraIso()` so every
  camera-relative layer pivots identically: voxel + SDF cull viewports,
  per-axis smooth-yaw scatter base (`perAxisBase_`), SDF main-canvas
  placement, trixel→trixel composite, both particle paths, the final
  framebuffer blit, and the world-hitbox **hover** projection (matches the
  already-pivoted picking inverse).
- **Left raw** per the design ruling: `camera_anchor` lighting grid, sprites,
  detached entity canvases, debug overlays.
- Demo: added a `zoom4_pan16_yaw{0,90,180,45}_pivot` shot cluster (the pivot
  is only observable when panned **and** rotated) + a `--pivot-origin` debug
  toggle for A/B capture.

### Before / after (IRShapeDebug, Metal; panned camera iso (16,16))

| yaw | CAMERA_CENTER (new default — focus pinned) | ORIGIN (`--pivot-origin` — old orbital swing) |
|----|----|----|
| 90°  | cc90  | o90  |
| 180° | cc180 | o180 |

### Verification
- **Pivot active:** CAMERA_CENTER vs ORIGIN differs **30.8% / 22.5%** of pixels
  at yaw 90° / 180° (focus pinned vs orbital swing).
- **Byte-identical fast path:** at yaw 0° the two modes differ only **0.0015%**
  — within the GPU-nondeterminism noise floor (unpanned shots: ~0.007%). The
  `CAMERA_CENTER` correction is the identity at `visualYaw == 0`. Every
  pre-existing (unpanned-or-yaw0) shot is unaffected.
- **Layers stay registered** (voxel, SDF, per-axis composite, AO/sun-shadow/
  light-volume) across cardinal **and** residual yaw while panned — no missing
  geometry, no light sliding off geometry (architect Q3 verify-item passes;
  `camera_anchor` correctly stays raw because the focus it centers on is the
  raw-cameraIso world point, held under screen center across rotation).

### Sequencing (per architect)
Land **after #1380** (camera-Z-yaw per-axis face-loss fix — smaller, CPU-only;
both touch `system_voxel_to_trixel.hpp` / `system_trixel_to_framebuffer.hpp`)
and **before #1408/#1299** (per-entity SO(3), which rebases over this PR's
`RotationPivotMode` addition). If #1380 merges first, expect a small rebase on
the two shared producer systems.

### Test plan
- [x] `fleet-build --target IRShapeDebug` clean (Metal/macOS).
- [x] Pivot A/B + byte-identical fast path verified (above).
- [ ] Reviewer: confirm the STAY-RAW set is correct (lighting grid, sprites,
      detached canvases, debug overlays).
- [ ] Cross-host: OpenGL (Linux) smoke — producer-side accessor swap only, no
      shader edits.

Closes #1352.


jakildev:
## NEEDS-DESIGN

Opus-worker pass on #1352. The foundation (enum + `RenderManager` setting + `getEffectiveCameraIso` + a float-precise `IRMath::isoPixelToPos3D(vec2,float)` overload + the plan's three named sites) is pushed and builds clean. But two things I learned reading `origin/master` contradict the plan and need an architect call before the rest can be wired.

### 1. The plan's "two producer sites" premise is materially false

The plan (`~/.fleet/plans/issue-1352.md`) states the camera iso offset reaches the GPU through **exactly two producer sites** (`system_voxel_to_trixel` + `system_trixel_to_framebuffer`), so correcting those two + the picking inverse "covers the entire blast radius with zero shader edits," and lists every shader as "auto (no edit)."

The "zero shader edits" half is correct — shaders read a frame-data field. But the audit conflated "one shared field the shaders read" with "one C++ producer that writes it." On `origin/master` **~13 C++ sites independently call `IRRender::getCameraPosition2DIso()`**, across four semantic categories:

**Content placement on the trixel canvas** (must pivot *together* or layers desync — the issue's own failure mode):
- `system_voxel_to_trixel.hpp:84` — main voxel raster ✅ *(wired in this PR)*
- `system_voxel_to_trixel.hpp:387` — per-axis canvas / cull path
- `system_shapes_to_trixel.hpp:177,223` — SDF overlay *(acceptance criteria explicitly require SDF registration)*
- `system_trixel_to_trixel.hpp:41` — inter-canvas composite (floors `camIso`)
- `system_render_stateless_particles_to_trixel.hpp:68`, `system_render_gpu_particles_to_trixel.hpp:68` — particles

**Framebuffer pan + screen blit** (the `cameraSubPixelOffsets` anti-vibration pair — must share the same base offset):
- `system_trixel_to_framebuffer.hpp:77` — framebuffer pan ✅ *(wired)* — plus a second read at `:215`
- `system_framebuffer_to_screen.hpp:42` — final blit residual

**Cull region:**
- `system_voxel_to_trixel` `updateCullViewport(getCameraPosition2DIso(), …)` — picks which voxels rasterize.

**Lighting-grid centering:**
- `detail/camera_anchor.hpp:39` — centers the light-occlusion + light-volume grids on the camera's world voxel (#360).

**Out-of-pipeline / screen-space:** `system_entity_canvas_to_framebuffer.hpp:67` (detached SO(3) canvas), `system_sprites_to_screen.hpp:112,246` (sprites), `system_debug_*` (overlays/minimap).

### 2. The plan's helper call would have broken the `yaw==0` identity

The plan specifies `P = isoPixelToPos3D(cameraIso.x, cameraIso.y, 0)`, but `isoPixelToPos3D` takes **`int`** args — it rounds a fractional (sub-trixel) pan away, so `getEffectiveCameraIso()` would *not* collapse to `cameraIso` at `yaw==0`, violating "cardinal fast path byte-identical." I added a float-precise `vec2` overload (round-trip verified) to fix this; flagging it so the plan is corrected for whoever finishes the work.

### Questions for the architect (you pick; I don't have to be right)

- **Q1 — approach.** With ~13 consumers rather than 2, is per-site `getCameraPosition2DIso → getEffectiveCameraIso` routing still the right model, or do you want a lower chokepoint? (Picking and debug legitimately need the *raw* value, so a blanket accessor swap isn't safe.)
- **Q2 — content set.** Confirm the content-placement producers (per-axis, SDF, trixel→trixel, both particles) + the framebuffer/blit pair all pivot through the helper. Any I should *not* touch?
- **Q3 — lighting-grid centering (the one that worries me).** Should `camera_anchor` (light-occlusion / light-volume grid center) follow the pivot, or stay on raw `cameraIso`? This is the same surface **#1311 / #1360 (AO/lighting on the composite) is actively redesigning** and is currently `fleet:design-blocked`. Should #1352 coordinate with that work, or scope light-grid pivot out entirely (follow-up that lands alongside #1311)?
- **Q4 — out-of-scope confirm.** Sprites (screen-space), detached entity canvases (separate SO(3) mechanism), and debug overlays — confirm these stay on raw `cameraIso` for #1352.
- **Q5 — cull.** Should `updateCullViewport` track the pivoted focus so culling matches placement (else edge-pop when panned+rotating)?

### My suggested split (if helpful)
Wire Q2's content set + the framebuffer/blit pair + cull (Q5) in this PR; scope camera_anchor lighting (Q3), sprites, detached canvases, and debug **out**, coordinated with #1311. That keeps #1352's diff to the main-canvas registration the acceptance criteria actually test, and avoids colliding with the in-flight composite-lighting work. But the lighting interaction is squarely yours to rule on.

Claim + worktree released; PR left at `fleet:design-blocked` + `fleet:wip`. Any opus-worker can resume from the foundation once you swap to `fleet:design-unblocked`.

— opus-worker-2

jakildev:

Your escalation is correct: the "two producer sites" premise doesn't hold, and the float-precise `isoPixelToPos3D(vec2,float)` overload is the right call for the `yaw==0` identity. The foundation stays. Full direction is in the updated plan (`~/.fleet/plans/issue-1352.md`, "⚑ ARCHITECT REVISION" at top); summary:

**Q1 — keep per-site `getEffectiveCameraIso` routing.** No safe lower chokepoint (picking/debug/sprites need the raw value). The helper's `ORIGIN`/`yaw==0` byte-identity is exactly right.

**Q2/Q5 — PIVOT this set** (the plan has the full annotated list): per-axis scatter base `system_trixel_to_framebuffer.hpp:215` (a NEW T3 site), SDF main-canvas `system_shapes_to_trixel.hpp:247`, trixel→trixel `:41`, both particle paths `:68`, final blit `system_framebuffer_to_screen.hpp:42`, **and both `updateCullViewport` sites** (`system_voxel_to_trixel.hpp:421`, `system_shapes_to_trixel.hpp:209` — Q5: cull must match placement or panned+rotated edges pop). The 3 foundation sites stay wired.

**Q3 — camera_anchor STAYS RAW, and this is now verifiable because T4 (#1360) MERGED.** `cameraAnchorVoxel()` computes the world voxel under screen-center (the true focus); the merged lighting does `localPos = pos3D − anchor` where `pos3D` comes from the shared binding-7 `frameCanvasOffset` UBO that your pivoted raster producer now writes. Anchor and the lighting `pos3D` recovery must derive from the **same** offset. **Verify-item:** `render-debug-loop` on the lighting demo, torch a few voxels off screen-center, panned + rotating between cardinals — watch for light sliding off geometry at large residual yaw. If it drifts, don't pivot the anchor; instead keep the lighting `pos3D` recovery on **raw** (raw-on-both is the consistent cheap fix).

**Q4 — confirmed RAW:** sprites, detached entity canvases, debug overlays. Plus one flag: the world-hitbox **hover** projection `system_hitbox_mouse_test.hpp:52` was left raw though you pivoted the picking inverses — wire it to track the pivot so hover boxes land on the rendered entity (unless you find a reason not to).

**Rebase (the PR is CONFLICTING):** T4/T5 merged onto your files. Exactly **one real conflict** — `ir_render_types.hpp`, where `RotationPivotMode` sits next to master's new `GpuVoxelTransform` block; resolution is **keep-both**. The other intersecting files auto-merge; after the rebase re-confirm the 3 wired sites still resolve to `getEffectiveCameraIso()` through the voxel-system restructure.

**Sequencing:** land **after #1380** (smaller, CPU-only; both touch the two producer systems) and **before #1408/#1299**.

Resume from the foundation; rebase, wire the PIVOT set, run the Q3 verify-item.

— opus-architect

jakildev:
Resumed from `fleet:design-unblocked` and completed per the architect direction. Rebased onto current master (one keep-both conflict in `ir_render_types.hpp` between `RotationPivotMode` and the new `GpuVoxelTransform` block), wired the full PIVOT producer set through `getEffectiveCameraIso()`, and ran the Q3 lighting verify-item — no light drift; `camera_anchor` correctly stays raw.

Verified: pivot active (30.8% / 22.5% pixel delta at yaw 90/180), byte-identical fast path at yaw 0 / unpanned (within GPU noise floor). Before/after A/B + numbers are in the updated PR body.

**Sequencing:** land after #1380, before #1408/#1299 (per the architect). If #1380 merges first, expect a small rebase on the two shared producer systems. — opus-worker-2

jakildev (commented):
## Review — render: camera Z-yaw pivots about the camera focus (#1352)

**Verdict:** approve

### STAY-RAW set (test-plan check)

Confirmed: the following sites correctly retain `getCameraPosition2DIso()` per the architect direction:

- `camera_anchor.hpp:39` — lighting-grid center stays raw (Q3 ruling: the lighting `pos3D` recovery uses the pivoted `frameCanvasOffset` UBO; raw-on-both-sides is the consistent path)
- `system_entity_canvas_to_framebuffer.hpp:67` — detached entity canvases (Q4)
- `system_sprites_to_screen.hpp:112, 246` — screen-space sprites (Q4)
- `system_debug_culling_minimap.hpp:174`, `system_debug_overlay.hpp:185, 197, 245` — debug overlays (Q4)

All 13 PIVOT sites are wired through `getEffectiveCameraIso()`. ✓

### Checklist

- **ECS:** no per-entity `getComponent` calls introduced; no mid-iteration structural changes; no allocations in hot tick paths. ✓
- **Ownership/lifetime:** `getEffectiveCameraIso()` calls into `RenderManager` and `IRPrefab::Camera::getYaw()` inline — no references stored across frames, no manager pointer held past World lifetime. ✓
- **Render pipeline / GPU:** pure CPU accessor swap — no shader changes, no UBO/SSBO layout modifications, no bind-point drift. CPU frame-data structs unchanged. ✓
- **Math:** `isoPixelToPos3D(vec2, float)` overload formula matches the int version (same derivation, float inputs). Round-trip identity `pos3DtoPos2DIso(isoPixelToPos3D(iso, 0)) == iso` verified analytically; the byte-identical fast path at yaw=0 follows from `pos3DtoPos2DIsoYawed(P, 0) == pos3DtoPos2DIso(P)` (c=1, s=0 collapses to the standard projection). ✓
- **Naming:** `m_rotationPivotMode` (private, `m_` prefix ✓); `ORIGIN`/`CAMERA_CENTER` (SCREAMING_SNAKE_CASE ✓); `RotationPivotMode` placement in `ir_render_types.hpp` mirrors `SubdivisionMode` — consistent pattern. ✓
- **API surface:** `getEffectiveCameraIso()` placed immediately after `getCameraPosition2DIso()` in the "Camera and viewport queries" group; `setRotationPivotMode()` / `getRotationPivotMode()` in a new "Rotation pivot mode" group mirroring the "Subdivision mode" group structure. Clean. ✓
- **`engine/render/CLAUDE.md`:** new gotcha added with the `getEffectiveCameraIso` rule and explicit STAY-RAW list. The wording is clear and actionable. ✓

### Nits

- `engine/render/src/ir_render.cpp:47`: Comment says "the correction is independent of depth, so z = 0 is exact." The correction is actually depth-dependent at nonzero yaw — `pos3DtoPos2DIsoYawed(isoPixelToPos3D(cameraIso, d), yaw)` varies with `d` for yaw ≠ 0 (the delta is `(−2ds/3, 2d(1−c)/3)`). The intended meaning seems to be: at yaw = 0 the round-trip identity holds for any depth, so `depth = 0` is the canonical choice. Suggest clarifying: "depth = 0 is the canonical iso-ray representative; `pos3DtoPos2DIso(isoPixelToPos3D(iso, 0)) == iso` holds exactly, so the round-trip at `yaw == 0` collapses to `cameraIso` independent of depth."

- `engine/math/include/irreden/ir_math.hpp:315`: "The int overload above snaps to whole iso pixels" — the int overload doesn't snap; it takes integer arguments, so fractional precision is lost at the call site (not inside the overload). Suggest: "The int overload requires integer iso coordinates; this overload accepts fractional values and preserves sub-pixel precision."

### Praise

- Escalating the producer audit in the NEEDS-DESIGN comment (2 assumed sites → 13 actual) before completing the wiring was the right call — proceeding on the original plan would have silently desynced at least 11 layers under panned+rotated cameras.
- Byte-identical fast path quantified (0.0015% delta at yaw=0/unpanned, within GPU noise floor) — a good regression guard for future camera changes.
- New `zoom4_pan16_yaw{0,90,180,45}_pivot` shot cluster + `--pivot-origin` toggle adds meaningful regression coverage: the existing unpanned-or-yaw0 shots cannot catch a pivot regression (CAMERA_CENTER correction is the identity in those cases), and this PR explicitly closes that gap.

- [ ] Reviewer task: STAY-RAW set confirmed above. ✓
- [ ] `fleet-build --target IRShapeDebug` clean on Linux/OpenGL (CPU-only accessor swap — no shader edits, but cross-host build confirmation needed).
- [ ] `fleet-run IRShapeDebug --auto-screenshot 10` on Linux; confirm pivot shots (`zoom4_pan16_yaw90_pivot`, `zoom4_pan16_yaw180_pivot`) are non-trivially different between CAMERA_CENTER and ORIGIN modes (≥20% pixel delta expected from the macOS baseline).

Opus recheck required: touches engine/render/ and engine/math/ (per-frame camera pivot correction applied to all content-placement producers; depth-0 iso-ray representative choice is non-trivial geometry).

jakildev:
Addressed nits:
- `engine/render/src/ir_render.cpp:47`: replaced "the correction is independent of depth, so z = 0 is exact" with the accurate framing — depth = 0 is the canonical iso-ray representative; the round-trip identity `pos3DtoPos2DIso(isoPixelToPos3D(iso, 0)) == iso` holds exactly, and the yaw = 0 collapse follows from that (not from depth-independence).
- `engine/math/include/irreden/ir_math.hpp:315`: replaced "snaps to whole iso pixels" (incorrect — precision is lost at the call site, not inside the overload) with "requires integer iso coordinates; this overload accepts fractional values and preserves sub-pixel precision".

jakildev (commented):
## Opus re-review — render: camera Z-yaw pivots about the camera focus (#1352)

**Verdict:** approve (keeping `fleet:approved` + `fleet:has-nits`)

Picked up on the core-invariant criterion: `engine/render/` + `engine/math/`, a per-frame geometric correction applied to every content-placement producer. Sonnet did a thorough pass (approve, STAY-RAW set checked, two comment nits) and flagged `Opus recheck required` on exactly the right thing — the depth-0 iso-ray representative choice and the per-frame pivot geometry. I spent the Opus pass on what a hunk-level read can't settle: (1) the round-trip / yaw geometry, (2) **which** yaw `getEffectiveCameraIso` uses vs. what the GPU producers project with, (3) the inter-cardinal per-axis scatter consistency, and (4) the rebase keep-both. **Everything holds; nothing is needs-fix. Sonnet's verdict and both nits are correct.**

### Math foundation — confirmed exact
The new `isoPixelToPos3D(vec2, float)` overload round-trips exactly. Algebraically, `pos3DtoPos2DIso(isoPixelToPos3D(iso, d))`:
- x: `-wx + wy = -wx + (wx + iso.x) = iso.x` ✓
- y: `-wx - wy + 2·wz = -2wx - iso.x + (iso.y + 2wx + iso.x) = iso.y` ✓

for **any** `iso` and **any** depth `d`. So at `visualYaw == 0`, `getEffectiveCameraIso` collapses to `pos3DtoPos2DIso(isoPixelToPos3D(cameraIso,0)) == cameraIso` (c=1, s=0 in `pos3DtoPos2DIsoYawed`), making the cardinal/unpanned fast path byte-identical — including through the floored sites (`trixel_to_trixel`, both `perAxisBase_`), since `floor(effective)==floor(raw)` when `effective==raw`. The worker's 0.0015% noise-floor delta at yaw 0 is consistent.

### Pin geometry — confirmed correct
For focus point `F` (`pos3DtoPos2DIso(F) == cameraIso`), the screen position is `pos3DtoPos2DIsoYawed(F, θ) − cameraOffset`. The PR sets `cameraOffset = pos3DtoPos2DIsoYawed(F, θ)`, so `F` maps to screen-center (0) for **all** θ — the orbit-about-origin failure mode is removed by construction. This requires the offset's yaw to be the *same* yaw the producers project geometry with.

### Sonnet flagged the depth-0 choice for me — confirmed, and it's a real (correct) semantic, not arbitrary
On closer read: the yawed correction **is** depth-dependent at non-zero yaw — shifting the representative depth by `d` along the iso axis `(1,1,1)` moves the offset by `(−2ds/3, 2d(1−c)/3)` (matches Sonnet's nit derivation). So depth=0 is a genuine *choice*: it pins the **z=0 ground-plane point** under screen-center, not a generic focus. That's a sensible canonical choice, but it means the `ir_render.cpp:53` comment's justification ("the correction is independent of depth, so z = 0 is exact") is the one inaccuracy — it's the *yaw=0 round-trip* that's depth-independent, not the yawed correction. I confirm Sonnet's nit; it's a comment-accuracy fix, not a behavior bug (the pin is correct for the ground-plane focus). Suggest the comment say "depth=0 pins the z=0 world point under screen-center; any depth is a valid point on the focus ray, 0 is the canonical pick."

### Yaw consistency — confirmed
`getEffectiveCameraIso` uses `IRPrefab::Camera::getYaw()` (continuous *visual* yaw, quaternion-extracted), and `pos3DtoPos2DIsoYawed`'s parameter is `visualYaw`. The per-axis scatter pass (`drawPerAxisScatter`) projects faces with the same `visualYaw` (`system_trixel_to_framebuffer.hpp:207,219`). So offset and geometry share one yaw — no cardinal-vs-continuous frame mismatch.

### Inter-cardinal per-axis scatter — the one I most wanted to check; confirmed consistent
The scatter vertex shader recovers each face's world origin from `(cell − perAxisBase, …)`, and the in-file invariant (`system_trixel_to_framebuffer.hpp:201–204`) requires `perAxisBase_` to **bit-match** `VOXEL_TO_TRIXEL_STAGE_1`'s per-axis write offset. The PR moves `perAxisBase_` to the *effective* iso (`:216`), so the write side had to move too — and it does: `dispatchPerAxisCanvases` only overwrites `perAxisRoute_` / `trixelCanvasOffsetZ1_` / `canvasSizePixels_`, leaving `cameraTrixelOffset_` as `buildVoxelFrameData` set it (`system_voxel_to_trixel.hpp:84`, now `getEffectiveCameraIso()`). So **both** the cell-write base and the scatter-recovery base derive the camera pan from the same effective offset, floored identically. The 45° shot is the right regression vehicle for this path; recommend the author also eyeball the `yaw45_pivot` A/B (it's the only shot that exercises the residual scatter under the pivot — the 90/180 deltas are cardinal-only and don't test it).

### Cull / lighting / rebase
- Both `updateCullViewport` sites now read effective iso (Q5) — cull tracks placement, no panned-rotate edge-pop. ✓
- `camera_anchor` correctly stays raw (Q3); the lighting `pos3D` recovery and the anchor both derive from the shared binding-7 offset that the pivoted raster now writes — raw-on-both is the consistent path. Worker ran the Q3 light-drift verify clean. ✓
- Keep-both rebase in `ir_render_types.hpp` resolved cleanly: `RotationPivotMode` and the `GpuVoxelTransform`/`kVoxelTransformStatic` sentinel block both present; the transform machinery (`flushStaticPositionRanges`, `:273`) survived. ✓

### ECS / ownership
Pure CPU accessor swaps — no per-entity `getComponent` added, no mid-iteration structural changes, no hot-path allocations, no shader/UBO/bind-point edits. Inline manager calls only (allowed per `.claude/rules/cpp-ecs.md` § "Manager accessor calls inside ticks"). ✓

### On Sonnet's nits
Both correct and both comment-only (the `ir_render.cpp` depth comment above, and the `ir_math.hpp:315` "snaps to whole iso pixels" → "requires integer args / loses sub-pixel precision at the call site"). Keeping `fleet:has-nits` so the author tightens both comments before merge — no re-review round needed.

jakildev:
Nits confirmed addressed in commit `befd90b9` ("render/math: clarify isoPixelToPos3D depth and overload comments"): both the `ir_render.cpp` depth-choice comment and the `ir_math.hpp:315` overload comment match the requested wording. Clearing `fleet:has-nits` — no further changes needed. — sonnet-fleet-1

jakildev:
Cross-host smoke OK on linux (fresh checkout build + IRShapeDebug --auto-screenshot 10). All 16/16 shots captured including the pivot set (zoom4_pan16_yaw{0,90,180,45}_pivot). Build clean (one pre-existing -Wformat-truncation warning in main.cpp spin_yaw label, unrelated to this PR's changes). ALSA/Jack audio init warnings are expected WSL2/headless noise. — sonnet-fleet-1

jakildev:
TESTED OUT and it worked pretty good. I believe this fix should be in the scope though. The camera movement does not appropriately adjust to the rotation, and things like wasd movement no longer match the direction that is expected (should compensate for rotation and move relitive to screen at same rate and direction

jakildev:
Addressed feedback: WASD camera movement now compensates for yaw rotation. In `system_render_velocity_2d_iso.hpp`, added `beginTick()` to cache the camera yaw once per frame, then rotate the iso velocity through `IRMath::rotate2D(velocity, yaw)` before integrating. This makes W/S/A/D move the camera in screen-relative directions rather than fixed world-iso axes — pressing W always moves "up" on screen regardless of the current yaw. — sonnet-fleet-1

jakildev:
Addressed feedback: WASD camera movement now compensates for camera yaw.

- `system_render_velocity_2d_iso.hpp`: added `beginTick()` that caches `IRPrefab::Camera::getYaw()` once per frame; tick applies `IRMath::rotate2D(velocity, cameraYaw_)` before integrating into position — so W/A/S/D intent stays display-aligned regardless of the current yaw angle.
- Byte-identical to the previous behavior at yaw=0 (rotate2D collapses to identity at angle=0).

— sonnet-fleet-1

jakildev (commented):
## Re-review — render: camera Z-yaw pivots about the camera focus (#1352)

**Verdict:** approve

Re-evaluating after `fleet:changes-made`. The prior Sonnet/Opus approval
covers the pivot feature and the comment-accuracy nit fixes (both confirmed
addressed in `befd90b9`). This pass focuses exclusively on the new WASD
yaw-compensation commit (06:30:46Z — `system_render_velocity_2d_iso.hpp`).

### WASD yaw-compensation (`system_render_velocity_2d_iso.hpp`)

**Scope — confirmed camera-only.** `C_Velocity2DIso` appears in exactly
three files: the component header, `command_move_camera.hpp` (the WASD
bindings), and the system itself. No non-camera entities use it, so
rotating all `C_Velocity2DIso` entities by yaw is correct and complete.

**Thread safety — correct.** `kConcurrency = Concurrency::PARALLEL_FOR` +
member `float cameraYaw_`: `beginTick()` writes the field on the main
thread before any parallel `tick()` calls read it. No concurrent
write hazard; the happens-before relationship is guaranteed by the
pipeline ordering between `beginTick` and the parallel ticks.

**Math.** `IRMath::rotate2D(velocity, cameraYaw_)` — function confirmed
(`engine/math/`; allowlist). CCW rotation for positive angle in Y-up
systems (the camera velocity space is Y-up: `MOVE_CAMERA_UP_START`
increments `velocity_.y`). Standard approach for screen-relative WASD
when the view is yawed.

**ECS / ownership / API.** New include `<irreden/render/camera.hpp>` is
correct (same pattern as `system_hitbox_mouse_test.hpp`). No per-entity
`getComponent`, no allocations, no structural changes. ✓

### Previously confirmed

- Prior nit fixes (`ir_render.cpp:47` depth comment; `ir_math.hpp:315`
  overload docstring) were addressed and confirmed by `sonnet-fleet-1` at
  04:22:52Z. Both are in the diff and match the requested wording. ✓
- STAY-RAW set, producer wiring, byte-identical fast path, and
  cross-host smoke (Linux) all confirmed in the prior Sonnet/Opus passes. ✓

### Nits

- **Linux build confirmation pending for the WASD commit.** The Linux smoke
  at 04:26:30Z confirmed the pivot feature but predates the 06:30:46Z WASD
  commit. The new change is CPU-only (one include + arithmetic), no shader
  edits — very low risk, but an explicit `fleet-build --target IRShapeDebug`
  on Linux for this commit would close the gap before merge.

- **Post-fix human test not yet recorded.** Human feedback was at
  06:16:19Z (pre-WASD-fix); no test comment appears after 06:30:46Z. A
  quick sanity check at yaw=45° or yaw=90° confirming W/A/S/D move in
  screen-relative directions before merge is recommended.

Opus recheck required: `system_render_velocity_2d_iso.hpp` is a new functional addition in `engine/prefabs/irreden/render/` — new `beginTick()` + per-frame yaw cache + velocity rotation — that postdates the prior Opus pass; the rotation-direction correctness in iso canvas space warrants an independent Opus verification.

jakildev:
Nit confirmation (fleet:has-nits, re-review 06:37:01Z):

- **macOS build clean:** `fleet-build --target IRShapeDebug` passes on macOS/Metal (macos-debug preset). The WASD commit (`baebe23d`) is CPU-only (one include + `IRMath::rotate2D` call) — compiles without warnings on this host. The reviewer noted Linux is very low risk given the same CPU-only nature; the Linux smoke at 04:26:30Z covered the prior commits; the WASD commit's Linux confirmation remains pending for a Linux fleet agent.
- **Human sanity check:** reviewer recommended a yaw=45° or yaw=90° W/A/S/D test before merge — this is for the human to confirm in-game behavior (per the reviewer's note).

No code changes required for these nits. — sonnet-fleet-1

jakildev (commented):
## Opus re-review — render: camera Z-yaw pivots about the camera focus (#1352)

**Verdict:** needs-fix

Scoped strictly to the new WASD yaw-compensation commit (`baebe23d`,
`system_render_velocity_2d_iso.hpp`) per Sonnet's `Opus recheck required`
flag — the pivot feature itself and the two comment nits were already
approved by the prior Sonnet/Opus passes and are not reconsidered here.

**Sonnet flagged the rotation-direction correctness for me and got the
framework checks right, but accepted `IRMath::rotate2D(velocity, yaw)` as
"the standard approach … CCW for positive angle." On closer read that is
the part that does not hold:** a pure 2-D rotation of the velocity does
*not* produce screen-relative motion under this PR's own pivot, so it
fails the human's explicit "move relative to screen at same rate and
direction" requirement (06:16:19Z) at every non-cardinal yaw — and
reverses outright at 180°.

### What Sonnet got right (confirmed)

- **`beginTick` happens-before is real.** `SystemManager::executeSystem`
  calls the begin-tick fn on the main thread (`system_manager.cpp:322`)
  before the `PARALLEL_FOR` dispatch (`IRJob::parallelFor`, same function
  ~`:353`). So `cameraYaw_` is published before any worker reads it — no
  data race. ✓
- **Scope is camera-only.** `C_Velocity2DIso` is used only by the camera
  entity (component header + `command_move_camera.hpp` + this system), so
  rotating all `C_Velocity2DIso` rows is complete and correct. ✓
- **`rotate2D` lives in `engine/math/` (allowlist), include is correct.** ✓

### The compensation math is wrong for the stated goal

The camera's `C_Position2DIso.pos_` that this `tick` integrates *is*
`cameraIso` — `RenderManager::getCameraPosition2DIso()` returns
`getComponent<C_Position2DIso>(m_camera).pos_` (`render_manager.cpp:217`).
Every content producer now subtracts **`getEffectiveCameraIso()`**, which
this PR defines as (`ir_render.cpp:42`):

```
effective(cameraIso) = pos3DtoPos2DIsoYawed(isoPixelToPos3D(cameraIso, 0), yaw)
```

Both halves are linear, so `effective = M(θ)·cameraIso` with Jacobian

```
M(θ) = [  cosθ        sinθ/3      ]
       [ -sinθ     (cosθ + 2)/3   ]
```

That matrix is **anisotropic** — the iso projection's 2:1 makes the iso-x
and iso-y axes scale into world differently (the `isoPixelToPos3D`
columns have magnitudes `(−½,½,0)` vs `(−⅙,−⅙,⅓)`). Because the camera
offset is distorted by `M(θ)`, a press that moves `cameraIso` by `δ`
shifts *all* on-screen content by `−M(θ)·δ`. To keep that screen shift
yaw-invariant ("same rate and direction") you must pre-multiply the input
velocity by **`M(θ)⁻¹`**, which is *not* a rotation. `rotate2D(v,θ)=R(θ)·v`
equals `M(θ)⁻¹` only at `θ=0`.

I ran the full chain numerically (unit press, per `dt·speed`; for true
screen-relative motion every row should equal the `yaw 0` row):

```
W key (velocity (0, +1)):              A key (velocity (+1, 0)):
  yaw   0: (+0.00, -1.00) |1.00|  ✓      yaw   0: (-1.00, -0.00) |1.00| ✓
  yaw  45: (+0.33, -1.14) |1.19|         yaw  45: (-0.67, -0.14) |0.68|
  yaw  90: (+0.00, -1.00) |1.00|  ✓      yaw  90: (-0.33, -0.67) |0.75|   ← diagonal, wrong
  yaw 135: (-0.33, -0.20) |0.39|         yaw 135: (-0.67, -0.81) |1.05|
  yaw 180: (+0.00, +0.33) |0.33|  ← REVERSED + ⅓ rate
  yaw 270: (+0.00, -1.00) |1.00|  ✓      yaw 180: (-1.00, -0.00) |1.00| ✓
```

Concretely: at 180° yaw, **W pans the view *down* at ⅓ speed**; at 90°
yaw, **A pans diagonally** instead of straight across. `rotate2D` is
correct only where the pressed axis happens to land on a cardinal (W at
0/90/270, A at 0/180); everywhere else the rate and/or direction is wrong.
Note also `det M(θ) = (1 + 2cosθ)/3`, which **vanishes at θ = 120°/240°** —
the magnitudes collapsing toward 135°/225° in the table are that
singularity. So `M(θ)⁻¹` itself blows up there: "rotate (or even
linearly un-distort) the velocity" cannot be a globally well-conditioned
fix.

The human has not yet test-driven this commit (the 06:16:19Z feedback
predates it; Sonnet's "post-fix human test recommended" nit is unmet), so
the screen-relative claim in the commit message is unverified — and on
this analysis it does not hold.

### Direction (author/architect's call)

Cheapest correct option, and probably the right one: **descope WASD to a
follow-up issue and merge the (already-approved) pivot feature alone.**
The pivot is independently correct and complete; the WASD interaction it
exposes is a separate, genuinely subtle problem.

If WASD compensation stays in scope, `rotate2D` is the wrong model. The
clean fix is the one the issue body itself suggested — **store the camera
pan as a world-space focus point and move *that* along screen-aligned iso
axes**, deriving the iso offset once. Then WASD is a translation of the
pivot in screen space (yaw-invariant by construction) and there is no
`M(θ)⁻¹` and no 120°/240° singularity. A per-frame `M(θ)⁻¹·v` patch would
make it *correct except near 120°/240°*, but the world-pivot reframing is
the durable answer and avoids the singular band entirely.

jakildev:
Addressed `fleet:needs-fix` feedback per Opus re-review:

- **Reverted WASD yaw-compensation commit** (`baebe23d`): `IRMath::rotate2D(velocity, yaw)` applies a rotation but the camera offset goes through the anisotropic `M(θ)` iso projection matrix — the two are equal only at θ=0, causing wrong direction and rate at non-cardinal yaw (reverses at 180°, singularity at 120°/240°).
- **Filed follow-up issue #1420** for the correct world-pivot approach: store camera pan as a world-space focus point and move it along screen-aligned iso axes, making WASD screen-relative by construction with no `M(θ)⁻¹` needed.

The pivot feature itself (all commits except the reverted one) is intact and previously approved by both Sonnet and Opus passes. PR is ready for re-review of the revert commit only.

— sonnet-fleet-1

jakildev (commented):
## Re-review — render: camera Z-yaw pivots about the camera focus (#1352)

**Verdict:** approve

Re-evaluating after `fleet:needs-fix` addressed (author comment 06:52:49Z). Scoped
exclusively to the revert of commit `baebe23d` (WASD yaw-compensation). The pivot
feature itself and the comment-accuracy nit fixes were approved by prior Sonnet
(04:00:32Z) and Opus (04:11:36Z) passes and are not reconsidered here.

### WASD revert — confirmed clean

`system_render_velocity_2d_iso.hpp` is entirely absent from the current diff.
No residual `beginTick()` / `cameraYaw_` member / `IRMath::rotate2D(velocity, yaw)`
pattern introduced in `baebe23d` survives. The revert is complete and leaves
no code in an intermediate state.

### Pivot feature — confirmed intact

All 13 PIVOT sites (`getEffectiveCameraIso` routing) present and unchanged
from the prior approval:

- `buildVoxelFrameData` (`system_voxel_to_trixel.hpp:84`) ✓
- Both `updateCullViewport` sites (`:418`, `system_shapes_to_trixel.hpp:209`) ✓
- SDF main-canvas placement (`system_shapes_to_trixel.hpp:244`) ✓
- `system_trixel_to_trixel.hpp:41` (trixel→trixel composite) ✓
- Both particle paths (`system_render_stateless_particles_to_trixel.hpp:68`,
  `system_render_gpu_particles_to_trixel.hpp:60`) ✓
- Framebuffer pan (`system_trixel_to_framebuffer.hpp:74`) ✓
- Per-axis scatter base (`system_trixel_to_framebuffer.hpp:212`) ✓
- Final blit (`system_framebuffer_to_screen.hpp:42`) ✓
- Hover projection (`system_hitbox_mouse_test.hpp:52`) ✓
- Both mouse-world inverses (`ir_render.cpp` — `mousePosition2DIsoWorldRender`,
  `mouseWorldPos3DAtIsoDepth`) ✓

STAY-RAW set unchanged: `camera_anchor`, detached canvases, sprites,
debug overlays all still on raw `getCameraPosition2DIso()`. ✓

### Follow-up filed

Issue #1420 filed for the correct world-pivot WASD approach (store camera
pan as world-space focus point, move along screen-aligned iso axes, making
WASD screen-relative by construction). The Opus reviewer's direction is
documented and won't be lost. ✓

### Build / smoke

The Linux smoke at 04:26:30Z covered the full pivot feature. The WASD commit
(`baebe23d`) and its revert are both CPU-only — no shader edits — so the
revert's build-risk profile is identical to the state the Linux smoke already
validated. The `fleet:needs-macos-smoke` / `fleet:needs-windows-smoke` labels
set on the prior pass remain appropriate and are unchanged.

Opus recheck not required.

jakildev:
Cross-host smoke OK on macOS/Metal (fresh checkout build + `IRShapeDebug --auto-screenshot 10`).

**Build:** clean (`fleet-build --target IRShapeDebug`, Metal backend).

**Render (Opus visual pass):** all 16 shots render correctly. The new pivot-coverage shots are healthy:
- `zoom4_pan16_yaw0_pivot` — identity case, full scene on the lit ground plane, shadows/AO correct.
- `zoom4_pan16_yaw90_pivot` / `_yaw180_pivot` — scene rotates about the camera focus (pan center), all SDF shapes present with correct colors/lighting/shadows.
- `zoom4_pan16_yaw45_pivot` — inter-cardinal per-axis deformation present and correct.

No black frames, missing geometry, inverted colors, or Metal-specific artifacts. The SDF-shape-vs-voxel-plane relative placement under rotation is consistent with the separately-tracked smooth-yaw work (#1370/#1414), not introduced here — this PR is a CPU-only `getEffectiveCameraIso()` accessor swap with no shader edits.

**Note (not a regression from this PR):** the run is intermittently aborting (exit 134) during macOS CoreMIDI init — `MidiInCore::initialize: error creating OS-X MIDI client object (-304)`. First run completed all 16 screenshots; a second run aborted at startup before any render. This is the known pre-existing macOS non-clean-shutdown/static-destruction symptom tracked in #1116 (T-336), in the render-independent MIDI path — clearing `fleet:needs-macos-smoke` on render correctness. `fleet:needs-windows-smoke` left for platform-catchup. — opus-worker (macOS)

jakildev:
A couple things that should probably be fixed here, and some other that should likely stay deferred. The main issue is that the camera movement direction still does not match the WASD movement. Holding A should just make the camera move directly to the left, relative to the new angle. However, it seems to still be dependant on the rotation. Also, there is jittering when moving the camera after rotating. I am guessing it is the Pixel perfect correction, but might by the extra trixel buffer jitter also. But the calculations that handle the offset do not adjust with rotation. Here are two screenshots after just moving the camera with A (you can see we went diagonal). Then the last screenshot shows the start of clipping at a rotation (spinning either direction will make it disappear or reappear.).

Screenshot 2026-05-31 at 1 18 18 PM

Screenshot 2026-05-31 at 1 17 54 PM
Screenshot 2026-05-31 at 1 18 04 PM

PR #1364, opened 2026-05-30
lua-ecs/codegen: emit `const auto&` row alias instead of full-struct copy for read-only `:at(i)` (#1353)

### LINKED ISSUE #1353: lua-ecs/codegen: arch.Comp:at(i) lowers to a full-struct copy — CODEGEN system ~2.5x hand-C++ at scale
**Model:** opus
**Type:** perf / lua-ecs codegen

## Surfaced by
Building a Lua-defined-ECS movement demo (per-row steer-to-goal over a 5-float component, `IRComponent.register` + `IRSystem.registerSystem`, CODEGEN default) and measuring its per-tick UPDATE cost against a byte-identical hand-written C++ system over the same codegen'd struct.

## Finding
| units | CODEGEN-lua | hand-C++ | ratio |
|---|---|---|---|
| 256  | 0.009 ms | 0.008 ms | ~1.1x (within the 1.5x parity gate) |
| 1024 | 0.027 ms | 0.011 ms | ~2.5x (exceeds the gate) |

The codegen emitter lowers `local a = arch.Comp:at(i)` to a **full-struct copy**:
```cpp
[](IRComponents::C_LuaMoveAgent& _ir_row_LuaMoveAgent) {
    auto a = _ir_row_LuaMoveAgent;          // copies the whole component row
    auto dx = (a.goal_x_ - a.pos_x_);
    ...
    _ir_row_LuaMoveAgent.pos_x_ = ...;      // writes lower to in-place member assignment (good)
}
```
The hand-C++ baseline reads fields directly off the column reference (no copy). At 256 units the working set is cache-resident and the copy is free; at 1024 the doubled read bandwidth shows as ~2.5x. Switching the write side from `setAt(Comp.new(...))` to partial `setField` only moved CODEGEN 0.031->0.027 ms — the read-side `:at(i)` copy is the dominant term, not the write idiom.

## Recommendation
Emit `auto& a = _ir_row_...;` (reference) instead of `auto a = ...;` (copy) when the `:at(i)` binding is only read through, not reassigned. That brings CODEGEN within the 1.5x parity gate at all counts for read-light kernels; mutating-through-a bodies keep the by-value copy.

## Repro
`creations/game` demo `IRIrredenUnitMovementLua`: `--driver lua|cpp --units {256,1024} --auto-profile 300`; read the `LuaMoveTick`/`CppMoveTick` rows from `save_files/profile_report.txt`. Emitted body is in the target's `codegen/unit_movement_lua_codegen.hpp`.

Absolute cost is tiny (sub-0.03 ms/tick, <0.3% of an 8.7 ms render-bound frame) — a parity-gate / movement-bound-workload concern, not a current frame-budget problem.

jakildev:
## Summary

The per-row CODEGEN emitter lowered `local a = arch.C:at(i)` to
`auto a = _ir_row_C;` — a full-struct **copy** of the component row. For
read-light kernels that copy is the dominant per-row cost (#1353 measured
~2.5x hand-C++ at 1024 rows; the read-side `:at(i)` copy, not the write
idiom). This change emits a `const auto&` **alias** of the row when the
binding is provably read-only, eliminating the copy.

A small forward dataflow pass (`collectRefSafeColumnAtDecls` +
`analyzeForward` in `system_dsl.cpp`) decides per binding. It emits an
alias only when the local is **never reassigned** and **never read after
the same component is written** via `setAt`/`setField`; any uncertainty —
including a read reachable across a nested-loop back-edge — falls back to
the safe by-value copy.

**Soundness:** the DSL forbids `a.field = ...` (the parser rejects it), so
the only way a component row changes mid-body is a column write on its own
component. A copy freezes the row at bind time; an alias tracks live
writes — they diverge *exactly* when a read of the binding is sequenced
after such a write. The analysis is conservative: it keeps the copy on any
read-after-write, reassignment, or shadow.

## What the emitter now produces

All read-before-write fixtures become aliases; the read-after-write fixture
keeps its copy (verified in the generated header):

```cpp
// CodegenMove (read-only): aliases, no copy
const auto& pos = _ir_row_CodegenSysPos;
const auto& vel = _ir_row_CodegenSysVel;

// CodegenReadAfterWrite (reads r.x AFTER writing column x): copy kept
auto r = _ir_row_CodegenSysPos;
_ir_row_CodegenSysPos.x_ = static_cast<float>(99.0f);
_ir_row_CodegenSysPos.y_ = static_cast<float>(r.x_);   // pre-write value
```

## Notes

- **Perf verification:** the change *is* the optimisation, confirmed by
  direct emitted-code inspection (read-only `:at(i)` → `const auto&`). The
  modified source is the build-time codegen tool, not a per-frame kernel,
  and the benchmark vehicle (`IRIrredenUnitMovementLua`) lives in the game
  repo, so the standalone `optimize` profiler was not run against this
  engine diff.
- **Deferred cleanup (not in this PR):** the new analysis helpers share the
  switch-per-`Kind`-and-recurse skeleton with `Emitter::emitStmt`/`emitExpr`.
  A generic `walkStmt`/`walkExpr` visitor could unify the traversal, but
  that refactors the tested emitter and is out of scope here; the new code
  matches the file's existing hand-written-switch convention.

## Test plan

- [x] `IrredenEngineTest` builds clean (codegen tool rebuilds → fixtures
      regenerate).
- [x] `LuaSystemCodegenTest.*` — 8/8 pass, including the new
      `ReadAfterWriteKeepsCopySemantics` (pins copy semantics: x=99, y=3 for
      input (3,7); an alias would give y=99).
- [x] Full suite green: 1020/1020.
- [ ] (reviewer, optional) confirm the 2.5x→~1.1x parity improvement via the
      game-repo `IRIrredenUnitMovementLua --driver lua --units 1024
      --auto-profile 300` repro.

Closes #1353

### DIFF
--- a/cmake/lua_codegen/system_dsl.cpp
+++ b/cmake/lua_codegen/system_dsl.cpp
@@ -10,6 +10,7 @@
 #include <stdexcept>
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 
 namespace IRLuaCodegen {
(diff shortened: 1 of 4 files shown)


jakildev (commented):
## Review — lua-ecs/codegen: emit `const auto&` row alias instead of full-struct copy for read-only `:at(i)` (#1353)

**Verdict:** approve

### Praise

- **Soundness argument is airtight.** The DSL forbids `a.field = ...` (parser rejects it), so the only way component C's row changes mid-body is via `setAt`/`setField` on `arch.C`. A copy and an alias only diverge when a read of the binding is sequenced after such a write — exactly the case the analysis catches. The comment block before the new functions states this clearly, making the invariant auditable.
- **`FOR_NUMERIC` back-edge is handled correctly.** Passing `cWritten || bodyWrites` as the starting `cWritten` to the loop-body analysis means a write-in-iteration-N / read-in-iteration-N+1 is treated as a read-after-write and correctly kept as a copy. The surrounding comment explains why the enclosing per-row loop requires no such treatment (the binding decl re-runs at the top of each iteration).
- **IF-branch conservatism is correct.** Each branch is analyzed with the pre-branch `cWritten` independently (exclusive execution), and `stmtWritesComponent` post-branch OR-s all branches to advance `cWritten` conservatively for subsequent statements. This correctly handles `if cond then setAt(...) end / local b = a.x` as unsafe.
- **`ReadAfterWriteKeepsCopySemantics` test pins the critical boundary.** The fixture proves the specific case (setField writes x, then reads `r.x`) emits copy semantics: `y = 3.0f` (original x), not `y = 99.0f` (post-write x via alias). The test would fail if aliasing were incorrectly applied here.
- **Pointer identity in `refSafeColumnAtDecls_` is stable.** Both `collectRefSafeColumnAtDecls` (stores `stmts[k].get()`) and `emitStmt` (queries `&s` where `s = *stmts[k]`) dereference the same `StmtPtr` in the same `perRowLoop->forBody_`. The addresses match by construction, and the `Emitter` is destroyed at the end of `emitSystem` before any AST ownership changes.
- **CLAUDE.md update is correct and concise.** The new paragraph documents the optimization's semantics (what lowers to alias vs copy) without over-explaining the implementation.

### Nits

- `cmake/lua_codegen/system_dsl.cpp` — no test pins the *positive* case (that a read-only binding actually emits `const auto&`). If `aliasRow` were accidentally never set true (e.g., the `refSafeColumnAtDecls_.count(&s)` check silently failed), all semantic tests would still pass because copy is always correct. A fixture that asserts the generated `.hpp` contains `const auto& pos` for `CodegenMove` (the canonical read-only binding) would catch that regression:

  ```cpp
  // In lua_system_codegen_test.cpp, after build:
  TEST_F(LuaSystemCodegenTest, ReadOnlyAtBindingEmitsConstRef) {
      // Could grep the generated header for "const auto& pos"
  }
  ```

  Alternatively, a comment in the existing `MovePositionByVelocityRoundTrip` test noting "pos and vel bindings are alias-safe; see lua_system_codegen_fixtures.hpp to verify `const auto&` emission" would at least document the expected artifact.

- `cmake/lua_codegen/system_dsl.cpp` — no test for the `FOR_NUMERIC` back-edge (`cWritten || bodyWrites`). A fixture like:

  ```lua
  local a = arch.CodegenSysPos:at(i)
  for j = 0, 2 do
      local tmp = a.x          -- iteration 2+: must not see post-write x if alias
      arch.CodegenSysPos:setField(i, 'x', tmp + 1.0)
  end
  ```

  ...would verify that `a` stays a copy when a write inside a nested loop precedes a read in a subsequent iteration. Without it, a future change that removes `|| bodyWrites` from the FOR case would not be caught by any test.

- [x] `IrredenEngineTest` builds clean — confirmed in PR body (fixtures regenerate)
- [x] `LuaSystemCodegenTest.*` — 8/8 pass including `ReadAfterWriteKeepsCopySemantics` — confirmed in PR body
- [x] Full suite 1020/1020 green — confirmed in PR body
- [ ] (Optional / reviewer) Confirm `lua_system_codegen_fixtures.hpp` contains `const auto& pos` and `const auto& vel` for `CodegenMove`, and `auto pos` (copy) for `CodegenParallelInc` (which has a read-in-RHS-of-setAt pattern that is alias-safe, so this may also emit `const auto&` — verify emitted code matches intent)

Opus recheck not required.

jakildev (commented):
## Opus re-review — lua-ecs/codegen: emit `const auto&` row alias instead of full-struct copy for read-only `:at(i)` (#1353)

**Verdict:** approve

Sonnet approved and marked "Opus recheck not required," but copy-vs-alias selection for component rows in generated tick code is a lifetime/aliasing invariant in `engine/script` codegen, so I took the focused Opus pass regardless of the flag. Standalone PR (base `master`). I spent the pass entirely on the one thing the model split asks Opus to certify three-deep: that the dataflow analysis's model of "what writes a component" and "what reads the binding" is **complete** against the actual AST — because an alias is unsound the instant a real write or a real read escapes the analysis, and the failure mode is silent per-row data corruption in every generated system.

### Aliasing/lifetime soundness — confirmed complete

Sonnet called the soundness argument "airtight" from the comment block and the read-after-write test. On a full cross-file trace against `system_dsl.hpp` it holds, and the reason it holds is stronger than the comment states — it's an exhaustive-coverage argument, not just the "DSL forbids `a.field =`" observation:

- **CODEGEN is per-row-only.** `matchPerRowCanonicalLoop` rejects any body that isn't the single canonical `for i = 0, arch.length - 1` loop (`system_dsl.cpp:1658`), so the emitted shape is always the per-component lambda `[](C_X& _ir_row_X, ...)`. `_ir_row_X` is an **lvalue reference parameter** — `const auto& a = _ir_row_X;` is a true live alias, and `setField`/`setAt` lower to writes through `_ir_row_X`, so the alias observes them. That is precisely the divergence `analyzeForward` guards (read of the binding while `cWritten`). The batch form never reaches this path (it's `mode="eval"`, runtime), and `refSafeColumnAtDecls_` is only populated for the per-row body, so a non-per-row decl can never be mis-aliased. ✓
- **Per-row body runs once per row** (dispatch invokes the lambda per entity), so the binding re-binds to the current row each invocation — there is no enclosing in-lambda loop and therefore no cross-row aliasing. The only real loops inside the body are nested `FOR_NUMERIC`, handled by the back-edge.
- **Write-detection is complete.** A component row mutates *only* via `setAt`/`setField`, and `system_dsl.hpp:163-177` shows both parse to `StmtKind::ASSIGN` with `COLUMN_SET_AT`/`COLUMN_SET_FIELD`. `stmtWritesComponent` and `analyzeForward`'s ASSIGN case both key on exactly those. `EXPR_STMT` (the only other call-bearing statement) cannot carry a component write — intrinsics are pure (`math.*`/`IRMath.*`) — so leaving it out of `cWritten` advancement is correct, and it still checks reads. ✓
- **Statement coverage is exhaustive.** `StmtKind` is exactly {LOCAL_DECL, ASSIGN, IF, FOR_NUMERIC, EXPR_STMT, RETURN}; `analyzeForward` handles all six. There is no WHILE/BREAK/CONTINUE in the DSL, so FOR_NUMERIC is the only back-edge and the single `cWritten || bodyWrites` OR at loop entry is sufficient. ✓
- **Read-detection is exhaustive.** `exprReadsName` recurses into every non-leaf `ExprKind` (UNARY/BINARY operands, INDEX receiver, and the `args_` vector of INTRINSIC_CALL/COLUMN_AT/COLUMN_GET_FIELD/COMPONENT_NEW); only the three literal kinds fall to `default`, and they are genuinely leaf. No read of the binding can hide in a sub-expression. ✓
- **Back-edge OR verified by trace.** For nested loops I traced write-in-outer/read-in-inner, write-in-inner/read-in-outer, and read-before-write-in-same-loop: `blockWritesComponent` recurses fully and each loop entry re-ORs, so a write-then-read across any nesting depth is caught. The same-statement case (`setField(i, 'y', r.x)` after writing `x`) is caught because the RHS/index are checked against pre-statement `cWritten` before the statement's own write advances it — exactly what `ReadAfterWriteKeepsCopySemantics` pins. ✓

The analysis errs conservative in every uncertain direction (IF OR-s all branch writes post-branch; a for-var that shadows the binding name isn't special-cased but only costs a missed optimization, never correctness; any unprovable binding falls back to the by-value copy). I could not construct a Lua body in this DSL that the analysis aliases unsoundly.

Pointer-identity is stable as Sonnet noted: `collectRefSafeColumnAtDecls` and `emitStmt` dereference the same `StmtPtr`s in the same `forBody_`, and the AST is immutable between collect and emit.

### On the nits

Sonnet's two nits (no positive-case test asserting `const auto&` is actually emitted; no FOR back-edge fixture) are real coverage gaps, not correctness defects — a silent failure to alias is always safe, so they'd only catch a future regression that *loses* the optimization. `fleet:has-nits` is already set and they're tracked; not worth a re-review round. Agree with Sonnet's framing.

### Test plan (unchanged — confirmed in PR body)

- [x] `LuaSystemCodegenTest.*` 8/8 incl. `ReadAfterWriteKeepsCopySemantics`; full suite 1020/1020 green.
- [ ] (Optional) positive-case fixture asserting `const auto& pos`/`const auto& vel` emission for `CodegenMove` — the gap Sonnet flagged.

Keeping Sonnet's approve; the aliasing/lifetime invariant is sound end-to-end. Clearing the Opus pass with `fleet:approved` (already set).

jakildev:
Addressed both nits from the review:

- **Nit 1 (positive alias case)**: Added a comment to `MovePositionByVelocityRoundTrip` documenting that `pos`/`vel` should emit `const auto&` and pointing to the generated fixture to verify. The regenerated `lua_system_codegen_fixtures.hpp` does show `const auto& pos` and `const auto& vel` for `CodegenMove` after the PR.

- **Nit 2 (FOR_NUMERIC back-edge test)**: Added `CodegenForNumericBackEdge` fixture (bind `local a = arch.C:at(i)` before a nested for_numeric loop that writes the same component), and `ForNumericBackEdgeKeepsCopySemantics` test — verifies x=2.0 after 3 iterations (copy semantics; alias would accumulate to 4.0). Pins the `cWritten || bodyWrites` back-edge analysis.

1021/1021 tests pass.

PR #1373, opened 2026-05-30
fleet: park gated tasks with fleet:needs-human instead of stripping human:approved

jakildev:
## Why `human:approved` was getting stripped

Tracked to one place: [role-opus-worker.md](.claude/commands/role-opus-worker.md)'s
escape hatch for a task that can only be completed by a step the fleet
can't perform autonomously — most often a **gated self-config edit**
(`.claude/commands/role-*.md`, `.claude/agents/*`). Exactly #1312's case
("point role doc at file-epic skill"). The worker comments what a human
must apply, then "bumps it out of pickup" by removing **both**
`human:approved` and `fleet:queued`.

It removed `human:approved` on purpose: `fleet-queue-ingest` **re-stamps
`fleet:queued`** on any `human:approved` issue that lacks it (it skips
only when `fleet:queued` or `fleet:scope-shipped` is already present). So
removing *only* `fleet:queued` would re-queue the issue next ingest tick
→ the worker re-claims a task it can never finish → loop. Stripping
`human:approved` was the only lever to drop it out of the ingest's set —
at the cost of destroying the human's durable approval signal.

Live trace on #1312 — both removed in the same second by the worker:
```
2026-05-30T00:49:36Z  unlabeled human:approved
2026-05-30T00:49:36Z  unlabeled fleet:queued
```

## Fix — `fleet:needs-human`, a park-label modeled on `fleet:scope-shipped`

`fleet:scope-shipped` already parks an approved issue out of the queue
*without* removing `human:approved` (it's in the ingest skip sets). This
adds the sibling for "fleet can't do it autonomously, a human must act":

- **fleet-labels** — new `fleet:needs-human`.
- **scout `_INGEST_SKIP_LABELS`** + **`fleet-queue-ingest`** stamping
  skip — treat `fleet:needs-human` like `fleet:scope-shipped`, so the
  ingest never re-stamps `fleet:queued` while it's present.
- **role-opus-worker** — the escape hatch now does
  `--remove-label fleet:queued --add-label fleet:needs-human` and
  **keeps `human:approved`**. The label clears when the human applies
  the change (ingest re-queues it) or closes the issue.
- **FLEET.md** — documents the label + ownership.

Net: `human:approved` is now only ever set or removed by the human — no
fleet role strips it.

## Tests

`test_queue_manager_projection.py` +1 (`test_needs_human_excluded_from_pending`:
a `human:approved` + `fleet:needs-human` issue contributes nothing to the
ingest projection/slice, so it won't be re-queued) — **14 pass**. Scout
parses; both embedded ingest Python blocks parse.

### DIFF
--- a/.claude/commands/role-opus-worker.md
+++ b/.claude/commands/role-opus-worker.md
@@ -569,11 +569,17 @@ Do the work, then exit cleanly:
      below.
    - **A step you cannot perform** (e.g. the self-config edit above):
      comment on the issue naming exactly what a human must apply, then
-     bump it out of autonomous pickup so you stop re-claiming it —
-     `gh issue edit <N> --remove-label human:approved --remove-label fleet:queued` —
-     and release your `fleet-claim`. Do NOT re-claim it next iteration:
-     the gate is deterministic, so retrying only burns iterations on a
-     wall until a human applies it.
+     park it out of autonomous pickup so you stop re-claiming it —
+     `gh issue edit <N> --remove-label fleet:queued --add-label fleet:needs-human` —
+     and release your `fleet-claim`. **Keep `human:approved`** — it's the
+     human's durable approval signal, not yours to strip; removing it was
+     the old workaround for the ingest re-stamping `fleet:queued` (it
+     re-queues any `human:approved` issue lacking it), which is exactly
+     what `fleet:needs-human` now suppresses (it's in the ingest skip set,
+     like `fleet:scope-shipped`). Do NOT re-claim it next iteration: the
+     gate is deterministic, so retrying only burns iterations on a wall.
+     The label clears when the human applies the change — then the ingest
+     re-queues it — or the human closes the issue.
 
    **Design escalation via `fleet:design-blocked`.** When the
    blocker is specifically architectural — the assigned task can't
--- a/docs/agents/FLEET.md
+++ b/docs/agents/FLEET.md
@@ -792,6 +792,19 @@ Specifically, **never pass these via `--label` when filing**:
   the landing PR; not added if the comment call fails (safe retry on
   next tick). Don't add manually unless you've verified a merged PR
   covers the scope.
+- `fleet:needs-human` — owned by the **author worker** (opus-worker).
+  Set when a queued task can only be completed by a step the fleet
+  can't perform autonomously — most often a gated self-config edit
+  (`.claude/commands/role-*.md`, `.claude/agents/*`). The worker
+  comments naming what the human must apply, removes `fleet:queued`,
+  adds this label, and releases its claim. **`human:approved` is
+  KEPT** — it is the human's durable approval, not the worker's to
+  strip. Like `fleet:scope-shipped`, this label is in the scout's
+  `_INGEST_SKIP_LABELS` and the ingest's stamping skip, so the ingest
+  does NOT re-stamp `fleet:queued` while it's present (removing
+  `human:approved` used to be the only way to defeat that re-stamp —
+  this label replaces that workaround). Clears when the human applies
+  the change (the ingest then re-queues the issue) or closes it.
 - `fleet:queued` / `fleet:task` — owned by **`fleet-queue-ingest`**,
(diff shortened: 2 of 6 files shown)


jakildev (commented):
## Review — fleet: park gated tasks with fleet:needs-human instead of stripping human:approved (#1373)

**Verdict:** approve

### Praise

- **Root cause is precisely identified.** The PR body traces the original behavior to a single decision point in `role-opus-worker.md`'s escape hatch, explains why removing `human:approved` was the only available lever before this fix, and cites a live timestamp audit from issue #1312. This is exactly the right level of documentation for a change that touches fleet infrastructure.
- **`fleet:scope-shipped` was the right model.** Using the existing `_INGEST_SKIP_LABELS` + per-role skip pattern as precedent is correct — `fleet:needs-human` fits the same semantic shape ("this issue is human:approved but excluded from the ingest cycle") without requiring new ingest machinery.
- **`human:approved` invariant is now enforced.** The FLEET.md addition ("owned by the author worker; `human:approved` is KEPT") documents the invariant clearly. If this is ever violated again it will be a clear protocol breach, not an ambiguous workaround.
- **Test coverage is correct and targeted.** `test_needs_human_excluded_from_pending` verifies both the projection hash (stable, no queue manager wake) and the `pending_issues` slice (empty). The parallel with the existing `test_scope_shipped_excluded_from_pending` test makes the coverage intent obvious.
- **All 14 tests pass.**

### Nits

- `scripts/fleet/fleet-state-scout:149` — `_ALREADY_QUEUED_LABELS` (used as the exclusion filter in `fetch_human_approved`) does not include `fleet:needs-human`. This means parked issues will still be fetched from GitHub and appear in `repos.engine.human_approved[]` in the state cache, then silently filtered out by `_INGEST_SKIP_LABELS` later. This is not incorrect — visibility into parked issues in the cache is arguably useful — but the asymmetry between `fleet:scope-shipped` (present in both sets) and `fleet:needs-human` (only in `_INGEST_SKIP_LABELS`) may surprise a future reader. If the intent is for parked issues to be invisible in the cache too, add it to `_ALREADY_QUEUED_LABELS`. If the intent is to keep them visible in `human_approved[]`, a one-line comment (`# fleet:needs-human intentionally absent; parked issues remain visible in human_approved`) would clarify.
- `.claude/commands/role-sonnet-author.md` — The sonnet-author role doesn't have this escape hatch. If a sonnet-author picks up a task that turns out to require a gated self-config edit, it has no documented path other than abandoning the claim. This is lower-risk (bounded tasks are less likely to hit role-doc edits) and is probably best filed as a separate follow-up issue, but worth noting for awareness.

- [x] `python3 scripts/fleet/tests/test_queue_manager_projection.py` — 14 pass (verified in this review)
- [ ] Manually verify `fleet-labels` creates `fleet:needs-human` on first run (`fleet-labels --apply` or equivalent)
- [ ] Run `fleet-up` and confirm that a real issue parked with `fleet:needs-human` + `human:approved` does not appear in the worker's task list on the next scout tick

Opus recheck not required.

jakildev:
Addressed nits:
- `scripts/fleet/fleet-state-scout:149` — added clarifying comment: `# fleet:needs-human intentionally absent; parked issues remain visible in human_approved[]`
- `role-sonnet-author.md` escape hatch gap — filed #1378 as a separate follow-up issue (reviewer suggested separate issue; bounded tasks are lower-risk)

14 tests still pass. — sonnet-fleet-1

jakildev:
Addressed nits from review:
- Added clarifying comment at `scripts/fleet/fleet-state-scout:149` explaining that `fleet:needs-human` is intentionally absent from `_ALREADY_QUEUED_LABELS` — parked issues remain visible in `human_approved[]` for the human's reference and are excluded from the queue projection via `_INGEST_SKIP_LABELS` instead (unlike `fleet:scope-shipped` which is fully invisible).
- Filed issue #1379 (sonnet) to add an escape-hatch path to `role-sonnet-author.md` for when a claimed task requires a gated self-config edit.

— sonnet-author (sonnet-fleet-2)

Issue #1372, opened 2026-05-30
render: consolidate inlined camera Z-yaw math into IRMath / ir_iso_common helpers (cleanup from #1370)


jakildev:
**Area:** engine/render, engine/prefabs/irreden/render, engine/math
**Model:** opus
**Blocked by:** #1370

## Context

Follow-up cleanup carved out of #1370 (and the smooth-yaw epic #1307). The
smooth camera Z-yaw work (T1–T5: #1308/#1309/#1310/#1361) landed iteratively
and inlined a lot of yaw math directly into render systems and shaders —
residual/raster yaw splitting, per-face deformation, continuous-iso
reposition, the CPU-side `|cos|·hX + |sin|·hY` extent expansion (duplicated
across systems), and magic cardinal tables. This issue does the cleanup pass
the original bug report asked for: pull those definitions into named helpers
and bring them into conformance with engine style.

**Blocked by #1370** (the fix) on purpose: the fix will likely relocate or
rewrite parts of the yaw math, and both touch the same four files heavily, so
the refactor must consolidate the *corrected* logic — not extract buggy
helpers that then need re-fixing.

## Scope

Consolidate inlined **yaw** math into shared, named helpers with byte-identical
CPU↔GPU behavior. Specifically:

- The CPU-side continuous-yaw iso extent expansion (`sizeForExtent` rotate +
  `|c|,|s|` grow) appears in `system_shapes_to_trixel.hpp` and the voxel
  cull path — extract one `IRMath` helper and call it from both.
- The cardinal cos/sin tables / `kHalfPi` / `rasterYawCardinalIndex` open-coding
  → named `IRMath` helpers (some already exist — `faceDeformationMatrix`,
  `pos3DtoPos2DIsoYawed`, `octahedralSnapResidual`; finish the job, remove the
  remaining inline copies).
- GLSL/Metal: factor repeated yaw/deform blocks in `c_voxel_to_trixel_stage_1`,
  `c_shapes_to_trixel`, and the per-axis scatter shaders into `ir_iso_common`
  shared functions (mirroring the C++ helpers so the CPU↔GPU contract is
  expressed once per side).
- Remove dead code left by retired iterations (T-058 / T-322 screen-space
  bilinear residual composite, retired by T-323) if any stubs remain.

## Out of scope

- **Behavior changes** — this is a pure refactor; rendering output must be
  unchanged at every yaw.
- The **cull-viewport derivation** consolidation across the 3 voxel/shape
[...]

PR #1361, opened 2026-05-30
render: smooth camera Z-yaw — T5: SDF shapes under the composite (#1345)

jakildev:
## Summary

T5 of the smooth camera Z-yaw epic (#1307). SDF shapes (`SHAPES_TO_TRIXEL`) stayed **cardinal-snapped** during smooth camera Z-yaw while the voxel pool glided (per-axis canvases, T3 #1310) — so they snapped at cardinals and mis-composited mid-rotation. Inside a residual bracket the SDF march now rotates by the **full continuous `visualYaw`** instead of cardinal-snap + `faceDeform`:

- **Continuous center reposition** — round the continuous-yaw iso projection (`pos3DtoPos2DIsoYawed`), matching the voxel pool's per-voxel reposition.
- **Continuous-yaw surface query** — the yaw-aware box/ellipsoid/general solvers run with `cos/sin(visualYaw)`; `sub==1` shapes route through the analytical path during rotation (the snap lattice walk is cardinal-integer-only, and the voxel pool is off its cardinal lattice too).
- **Shared WORLD-space `x+y+z` depth** (monotone along the view ray at rate `2·cos(yaw)+1` for `|residual|<45°`) so SDF composites by depth with the three per-axis voxel canvases — not the nearest-cardinal view depth.
- **Cull + tile footprint widened** to the continuous-yaw √2 extent so off-center shapes aren't dropped from the cardinal-snapped viewport.

Gated per-canvas via a new `smoothYawEnabled` flag (repurposes the first word of the std140 `_faceDeformPad` — `faceDeform` stays at offset 80, UBO size unchanged), so it applies only to the rotating **main world canvas**; detached per-entity canvases keep the cardinal + `faceDeform` path. At `residualYaw==0` the flag is 0 and the exact cardinal path runs — **byte-identical to master**.

GLSL **and** Metal both ported. No three-canvas split for SDF (it solves the surface analytically per pixel — the architect's "analytic per-shape yaw" option).

## Before / after (Metal, `IRShapeDebug --spin-yaw`, zoom 8)

Master snaps the SDF box to the nearest cardinal while the voxel cubes rotate; this PR rotates it smoothly to match.

| yaw 30° — before (snapped) | yaw 30° — after (smooth) |
|---|---|
|  |  |

| yaw 60° — before (snapped) | yaw 60° — after (smooth) |
|---|---|
|  |  |

Smooth progression within one bracket (this PR): 0° → 15° → 30°

| 0° (cardinal) | 15° | 30° |
[...]

PR #1380, opened 2026-05-30
render: stop per-axis voxel sun-shadow casting under camera Z-yaw (#1370, option C)

### LINKED ISSUE #1370: Culling or clipping of sdf and voxel shapes under rotation
Observed in both PRs for creating smooth yaw rotation that landed. Rotation looks good, lighting still needs work, but I noticed that in both the IRShapeDebug and IRVoxelYaw demos that some shapes get clipped or culled under the rotation. Perhaps it is distance values going too high or low, or actual culling mechanics. We need to investigate this and fix. The shapes start to "fade out" as seen in the attached screenshots. Additionally. I think we need to do a pass at all the dirty changes in the rendering shaders and systems that have inlined a lot of logic to calculate yaw, and see if we can clean these up, pull definitions out of the systems, etc in conformance with our style. 

Image
Image
Image

jakildev:
## render: stop per-axis voxel sun-shadow casting under camera Z-yaw (#1370, option C)

Fixes the false black voxel faces / "black holes" on multi-face voxel blocks at
non-cardinal camera Z-yaw (worst at 135°/225°/315°), reported in #1370.

### Root cause (architect-verified on Metal/macOS, rev-3)
The face-local per-axis sun-shadow bake (#1311) baked each rotating voxel block's
**mutually-perpendicular faces** into the **shared** sun depth map, where a block's
own top face shadowed its own side faces — false self/cross-occlusion that is
**structural to the face-local representation**. The cardinal path avoids it via
per-screen-pixel iso-depth-plane flattening (faces smear onto parallel patches that
can't mutually occlude in the sun projection); the face-local per-axis canvas
doesn't reproduce that, so a block's top face overhangs and shadows its side faces.
The darkened side faces read black because the back-facing ambient floor is fully
multiplicative (`0.45 × ~0.30 ≈ 0.13·src`).

This was confirmed by experiment, not static analysis: disabling the per-axis bake
(`perAxisRoute != 0 → return` in `c_bake_sun_shadow_map`) cleanly removed the false
shadow, cardinal-byte-identical; bias widening (Exp 2/3) only partially cleared it,
ruling out a depth-offset fix.

### The fix (option C — architect direction)
Drop the per-axis sun-bake entirely while rotating: remove `dispatchPerAxisBake`,
its call site, the per-axis canvas members, and their `beginTick` resolution from
`BAKE_SUN_SHADOW_MAP`.

- **No shader edit** → GLSL/Metal parity trivially preserved (cleaner than the
  shader early-return prototype — no wasted dispatch).
- Per-axis voxels still **receive** sun shadows (from the main canvas / SDF) and are
  still **AO-shaded** via `COMPUTE_SUN_SHADOW` / `LIGHTING`; they just stop
  **casting** into the shared map.
- **Interim trade:** voxels do not cast sun shadows at non-cardinal yaw until the
  faithful trixel/screen-space resolve lands (**#1435**, the north-star path).

### Verification (Metal/macOS, `IRShapeDebug --spin-yaw --auto-screenshot 24`)
- **Non-cardinal black-holes gone.** The 135°/225°/315° false-black voxel faces are
  eliminated (before/after below).
- **Cardinal byte-identity.** 90°/180°/270° spin-sweep frames are `cmp`-identical
  before vs after (the per-axis canvases only allocate while rotating). The 0° frame
  differs only by **first-shot run-to-run non-determinism** — proven by a
  fixed-vs-fixed double run that differs *more* at shot 0 (672 px) than the
  before/after diff (432 px), and is byte-identical at 90°.
- **SDF unaffected** — SDF shapes and their cast shadows are identical before/after
  (only the voxel per-axis bake changed). SDF black-holes remain tracked as #1414.

#### 135° (cardinal-1 boundary, worst case)
| before | after |
|---|---|
| 135 before | 135 after |

#### 225°
| before | after |
|---|---|
| 225 before | 225 after |

### Docs
Amends `docs/design/per-axis-trixel-canvas-rotation.md` to record the
self-occlusion mechanism, the option-C trade, and the deferral to #1435.

### Scope notes (sibling follow-ups, out of scope here)
- SDF black-holes under rotation → #1414
- SDF-over-voxel composite depth → #1430 (fixed by the merged depth PR #1434)
- per-axis canvas clipping → #1431

Closes #1370.


jakildev:
## NEEDS-DESIGN

Investigation of #1370 (SDF/voxel clip/cull/fade under camera Z-yaw). I confirmed the repro, **ruled out culling**, and localized the root cause to the **T3 per-axis trixel composite (#1310)** — which is the same path opus-worker-1's open **PR #1360 (T4 AO/lighting)** is actively modifying. That overlap plus the architectural nature of the fix is why I'm escalating rather than patching unilaterally.

### What I found (evidence)

**Repro:** `IRShapeDebug --spin-yaw --auto-screenshot 24` and `IRVoxelYaw --spin-yaw --auto-screenshot 24` (Metal/macOS).

1. **Cardinals (residual ≈ 0) are perfect** — at yaw 0/90/180/270 the per-axis path is released and the byte-identical single-canvas gather runs; scene is crisp.
2. **The breakage scales with |residualYaw| and worsens at non-zero cardinals.** Small residual (±15°, e.g. yaw 15° / 105°) renders fine at every cardinal. The damage concentrates at the **cardinal boundaries** (yaw 45°/135°/225°/315°, i.e. |residual| → π/4). At 45° (cardinal-0 boundary) the scene flattens but mostly survives with some black cracks; at **135°/225°/315° (cardinal 1/2/3 boundaries) most side-face structure is lost** — towers collapse to flat top-down squares and the ground/objects show large **black holes (dropped faces)** where the framebuffer background shows through. This is the "fade out / clipped / culled" the issue reports.
3. **It is NOT the cull.** Bisection: I temporarily ballooned the shared cull viewport (`IRMath::visibleIsoViewport` → `[-1e6, 1e6]`, which widens the voxel chunk mask, the GPU `cullIsoMin/Max`, AND the SDF iso-bounds cull simultaneously), rebuilt, and re-ran at 135°. **The frame was byte-identical to the un-ballooned run** — the missing content is genuinely not rendered into the framebuffer, not culled away. (Edit reverted; worktree clean.)
4. The store/scatter origin-recovery math is self-consistent: stage_1/stage_2 write the face-local in-plane lattice cell `faceLocalBase(axis) + faceInPlaneCoords(faceId, facePos)`, and the scatter recovers it via `faceOriginFromInPlane` (exact integer subtraction, no `2cos(yaw)+1` inverse) + projects with `pos3DtoPos2DIsoYawed(worldCorner, visualYaw)`. The loss is in **face coverage** (empty cells → the scatter's `color.a < 0.1` test degenerates the instance → black holes), not in where surviving faces land.

### Why this is a design call, not a unilateral fix

- The fix lives in the **T3 per-axis store/scatter** (`c_voxel_to_trixel_stage_{1,2}.glsl` per-axis path, `v_peraxis_scatter.glsl`, `system_trixel_to_framebuffer::drawPerAxisScatter`, `ir_iso_common.glsl` face-local helpers) — the core of the smooth-yaw epic (#1307) you're actively extending (T4 #1311/#1360, T5 #1345).
- **PR #1360 (T4) modifies the same composite path.** Any face-coverage fix I land here will textually and semantically collide with the AO/lighting work in flight on the composite.
- The face loss concentrates at the **cardinal boundary**, where one world axis goes edge-on (at 135°, view-dir y-component → 0, so the Y faces are edge-on) and the per-axis path resolves `visibleFaceIds` from the *snapped* `cardinalIndex` (`visibleFaceTripletCardinal`) while the camera is continuously between cardinals. Whether the per-axis composite should resolve the visible triplet **continuously across the ±45° boundary** (vs. the current cardinal snap), and how that interacts with the face-local lattice density, is an architecture call on the composite, not a local bug.

### Questions for you

1. Is the cardinal-boundary face loss a known/accepted limitation of T3 that **T4/T5 supersede**, or a genuine T3 defect to fix in #1370?
2. If fix-in-#1370: should the per-axis path resolve the visible-face triplet **continuously** (off `visualYaw`, not the snapped cardinal) so faces don't get dropped/mis-routed as one axis goes edge-on? Or is the face loss a **lattice-density / empty-cell** problem in the face-local store (the design doc claims the store is "dense and collision-free at every yaw", but I'm seeing cracks)?
3. **Sequencing vs PR #1360** — should #1370 stack on / wait for #1360 (T4) so the composite changes don't conflict?

Repro screenshots (Metal): IRVoxelYaw yaw 0 clean → yaw 45° flat-with-cracks → yaw 135° heavy face loss; IRShapeDebug same pattern. Happy to attach the before/after sweep on request.

(Scope note: the inlined-yaw-math cleanup half of the original #1370 text is already split out as #1372 and is out of scope here.)

— opus-worker-2

jakildev:

Excellent investigation — ruling out the cull by ballooning the viewport to a byte-identical frame is exactly the right bisection, and your root-cause localization is correct. I verified the mechanism on master. Full direction is in the updated plan (`~/.fleet/plans/issue-1370.md`, "⚑ ARCHITECT REVISION" at top); summary:

**Q1 — genuine defect, fix it here. NOT superseded by T4/T5.** Git history is decisive: the visible-face/scatter path was last touched by T3 (#1336) and is unchanged by T4 (#1360, lighting-only) and T5 (#1361, SDF-only), both now merged. This is the documented-deferred *"±45° rebracket polish"* (`docs/design/per-axis-trixel-canvas-rotation.md` ~line 182) — and the *band* of loss you see is worse than the doc's "clean by construction" zero-width ideal, so promote it to a fix.

**Q2 — root cause is the SNAPPED triplet, not lattice density.** `system_voxel_to_trixel.hpp:147` uploads `visibleFaceTripletCardinal(cardinalIndex)` (frozen across the whole ±45° bracket) while geometry draws at continuous `visualYaw`; near the boundary the frozen triplet names back-facing faces, the axis-gate drops them before the store, and their cells stay clear → black holes. The store is genuinely dense/collision-free (that claim holds).

**The fix (CPU-only):** resolve the triplet continuously via the **existing** `IRMath::visibleTriplet(rotation)` (already used by the DETACHED branch; a face only flips when edge-on, where its area ≈ 0) fed `qZ(visualYaw)`, in place of `visibleFaceTripletCardinal`. Change **both** uploads — store (`system_voxel_to_trixel.hpp:147`) and scatter (`system_trixel_to_framebuffer.hpp:~208-225`) — with identical resolution. No shader/store/gate change; only the CPU-side `visibleFaceIds_` value. Cardinal fast path stays byte-identical (`visibleTriplet(qZ)==visibleFaceTripletCardinal` at exact cardinals). Amend the design doc's "Deferred" + "Face-jump at the rebracket" sections in this PR.

**Q3 / sequencing — #1360 (T4) has MERGED**, so the coordination concern is gone. Land this **FIRST** among the smooth-yaw PRs (small, CPU-only, no deps); #1362 and #1408/#1299 rebase on top. Because the fix is CPU-only it's compatible with #1408's later per-entity `visibleFaceIds` SSBO promotion (the world/shared slot just becomes continuous).

**SDF is OUT OF SCOPE** — it's a separate path (continuous solve, no triplet gate; T5 made SDF rotate by full `visualYaw`), so the SDF black-holes have a different cause. Filed as **#1414** (`Blocked by: #1370`). Scope this PR to the **voxel** per-axis path; the title should drop SDF.

Resume from the foundation. Plan updated; design-doc amendment lands in this PR.

— opus-architect

jakildev:
## NEEDS-DESIGN (re-escalation after design-unblock)

Thanks for the direction — I picked this up off `fleet:design-unblocked`, read the
updated plan (`~/.fleet/plans/issue-1370.md` ⚑ ARCHITECT REVISION), and started
the CPU-only triplet swap. I hit a blocker the plan can't resolve and three
findings that I think need your call before I write the resolver, because the
prescribed fix doesn't map cleanly onto the code as it stands.

### Blocker 1 — `IRMath::visibleTriplet(rotation)` does not exist

The plan says: *"Use `IRMath::visibleTriplet(rotation)` (the SO(3) resolver
already shipping and already used by the DETACHED branch, `ir_math.hpp:~513`)."*

- `grep -rn "visibleTriplet" engine/` → **zero matches.** No such function.
- `ir_math.hpp:~513` is the middle of `rotateCardinalZInv` (a Z-yaw integer
  view→world helper), not a face resolver.
- The only visible-face resolver is **`visibleFaceTripletCardinal(CardinalIndex)`**
  (`ir_math.hpp:473`) — the *snapped* one we're trying to replace.
- The **DETACHED branch does NOT use a resolver**: `system_voxel_to_trixel.hpp:108-113`
  hard-codes a fixed `{X_NEG, Y_NEG, Z_NEG}` triplet (comment: "the per-canvas SO(3)
  bake absorbs the entity rotation… per-entity SO(3) visible-triplet is the follow-on
  in #1272"). So there is no shipping SO(3) face resolver to reuse.

So this isn't a swap — it's **designing and adding a new continuous resolver**. I'm
happy to write it, but I need the exact contract because of findings 2 and 3.

### Finding 2 — a faithful continuous Z-yaw triplet looks identical to the snapped one *inside* each bracket

`IRPrefab::Camera::computeYawSplit` (`camera.hpp:30-33`) is
`round(visualYaw / kHalfPi) * kHalfPi`, so **residualYaw ∈ [−π/4, +π/4]** — each
cardinal bracket is exactly ±45°. For pure Z-yaw with the iso camera, an X face
goes edge-on where `d_x = cosφ − sinφ = 0` ⇒ **φ = 45°**, i.e. exactly the bracket
boundary. Working the polarity through every bracket, the snapped triplet's named
faces are **front-facing for the whole interior of the bracket** and only reach
edge-on *at* the ±45° boundary (where the snap flips and the face area ≈ 0).

If that geometry is right, then a resolver that returns "the front-facing triplet
at the continuous `visualYaw`" **equals `visibleFaceTripletCardinal(cardinalIndex)`
everywhere inside the bracket** — i.e. the prescribed drop-in would be a no-op and
wouldn't change the per-axis stored set at all.

That means I'm missing the mechanism behind your root-cause ("the frozen triplet
names faces that are going back-facing near the boundary, the axis-gate drops them
→ black holes"). You verified the band on hardware and I trust that; I just can't
reproduce *why* the frozen face is back-facing within a ±45° bracket from the code.
**Is the face-flip angle actually offset from the ±45° bracket boundary** (so the
snap and the true edge-on disagree across part of the bracket)? Or is the real
culprit somewhere the triplet swap won't touch — e.g. the scatter's deformed
face-quad reconstruction / the `color.a < 0.1` test degenerating *thin near-edge-on*
faces (`drawPerAxisScatter`, `v_peraxis_scatter.glsl`), or the worst-case canvas
sizing? Knowing the exact flip geometry is what lets me write a resolver that
actually differs from the current one in the right direction.

### Finding 3 — "Same slot order (X-slot, Y-slot, Z-slot)" contradicts byte-identity at cardinals 90/270

`visibleFaceTripletCardinal` is **not** fixed X/Y/Z slot order — it *permutes* the
geometric slot→face assignment per cardinal:

- cardinal 0:   `{X_NEG, Y_NEG, Z_NEG}`  (slot0 = X-axis)
- cardinal 90:  `{Y_NEG, X_POS, Z_NEG}`  (slot0 = **Y-axis**)
- cardinal 180: `{X_POS, Y_POS, Z_NEG}`
- cardinal 270: `{Y_POS, X_NEG, Z_NEG}`  (slot0 = **Y-axis**)

`slot` is geometric (right column / left column / top — `localIDToFace_2x3`), and
at residualYaw==0 the **single-canvas fast path** (`perAxisRoute==0`,
`c_voxel_to_trixel_stage_1.glsl:228-243`) keys color/AO off `faceId =
visibleFaceIds[slot]` for the diamond position. A *fixed* X/Y/Z slot order would
put the wrong world face in the right/left columns at cardinals 90/270 → **not
byte-identical, and geometrically wrong** at those cardinals. So whatever resolver
we build **must reproduce the permuted cardinal-table order at exact cardinals**,
which is in tension with the plan's "same slot order (X,Y,Z)" note.

### What I need from you (pick / correct any)

1. **The resolver's exact contract.** Since `visibleTriplet` doesn't exist: a new
   `IRMath::visibleFaceTripletContinuous(float visualYaw)` (or SO(3) `(quat)`), with
   the precise per-slot face + polarity rule — specifically how it differs from
   `visibleFaceTripletCardinal` *inside* a bracket (finding 2) while collapsing to
   the **permuted** cardinal table at exact cardinals (finding 3).
2. **The mechanism** (finding 2): is the edge-on flip offset from the ±45° boundary,
   or is the band actually caused by the scatter reconstruction / canvas sizing
   rather than the triplet? If the latter, the CPU-only triplet swap is the wrong
   lever and the scope changes.

If it's easier to point me at a worked example of the intended slot/polarity mapping
for one off-cardinal angle (say φ = 30° and φ = 60°), I can generalize from that and
write + screenshot-verify the resolver in one pass.

No code changes pushed (investigation only — branch still at the claim commit).
Releasing the claim and worktree so any worker can resume once the resolver contract
is nailed down.

— opus-worker-1

jakildev:
## Architect direction — CORRECTED (rev 2); re-unblocking

You were right, and my first direction was wrong. I re-diagnosed rigorously (worked the geometry numerically against the actual shader/math formulas). Retracting the triplet prescription.

**Your Finding 2 is correct — the triplet is innocent.** Per-bracket residual is ±45° and a swept face goes edge-on at *exactly* the ±45° boundary, so `visibleFaceTripletCardinal` names the true front-facing set **everywhere inside every bracket** and only diverges at the boundary where the lone divergent face has **zero on-screen area**. A continuous resolver returns the same set inside the bracket → **the swap is a no-op.** Don't do it.

**Your Finding 3 is correct, and worse than feared.** `visibleTriplet(qZ(+visualYaw))` returns the **back** faces at cardinals 90/270 (`{X_NEG,Y_POS,Z_NEG}` vs the correct `{X_POS,Y_NEG,Z_NEG}`) — a hard regression. `visibleTriplet` is right for *entity* rotation (its DETACHED/#1408 use), not the GRID camera-yaw path. **Leave `visibleFaceTripletCardinal` as-is.**

**Blocker 1 was branch staleness** — `visibleTriplet` *does* exist on master (`ir_math.hpp:513`, #1386); your branch predates it. Rebase onto master, but the rebase isn't the fix.

### The real cause: the per-axis forward-SCATTER composite (T3), not face selection
1. **Edge-on coverage gap (symmetric → the band).** Visible faces are scattered as flat unit quads; summed quad area for an iso cube dips ~20% (6.0→4.83) as φ→boundary while the silhouette stays constant-area → an **uncovered screen wedge** = your black holes. The stretched face's quad doesn't grow to fill the vacated wedge. (`v_peraxis_scatter.glsl:60-104,115`.) This is the genuine "±45° rebracket polish" — lever is scatter **coverage**.
2. **POS-side composite-depth disadvantage (asymmetric → heavier at 135/225/315).** Flat per-face `gl_FragDepth`: top `Z_NEG`=`x+y+z`; NEG sides (at 45°)=`x+y+z` (tie); **POS sides (at 135/225/315)=`x+y+z+1`** → lose GL_LESS to the top → overdrawn ("towers collapse to flat squares"). (`c_voxel_to_trixel_stage_1.glsl:209-212`, `f_peraxis_scatter.glsl:24`.) This is the asymmetry driver.
3. **FULL-mode canvas-density clipping (possible — I wrongly dismissed it).** Per-axis canvases sized in ×1 iso-px but indexed in ×sub units; out-of-canvas cells silently clipped. My earlier "cull-balloon ruled out sizing" was invalid — that test doesn't exercise canvas size. Still open.

(`color.a<0.1` is ruled out — the holes are genuinely uncovered/depth-rejected cells, not alpha-killed.)

### Diagnose BEFORE fixing — static analysis can't rank 1/2/3 for hole magnitude (the demos run FULL, so all three are live):
1. Toggle `SubdivisionMode::NONE` (sub=1), re-sweep: holes shrink markedly → (3) matters; persist symmetric-heavier-at-1/2/3 → (1)+(2).
2. Read back the three per-axis **color** textures at yaw 45 vs 135: empty outer cells → coverage(1)/density(3); populated-but-black → depth(2).
3. Read back framebuffer **depth** in a hole at 135 → holds top-face depth where a side should win → confirms (2).

### Fix levers (after the diagnosis pins the dominant mechanism)
- Coverage(1): grow the stretched/Z footprints, or enforce the min on-screen trixel footprint **in the scatter quad** (`kMinOnScreenTrixelSizePx` currently only sizes the canvas), or half-texel conservative dilation.
- Seam(2): write **per-corner interpolated** `gl_FragDepth` from `pos3DtoDistance(worldCorner)` (`:115` writes flat `rawDist` today), or bias the cross-canvas tie.
- (3) if implicated: size the per-axis canvas in ×sub units / clamp without dropping silhouette cells.

This is now a **shader fix (GLSL + Metal parity)**, not CPU-only. Scope stays **voxel** (SDF is #1414). Amend `docs/design/per-axis-trixel-canvas-rotation.md` to record the scatter-coverage + flat-depth-seam mechanism. Full detail + worked φ=30/60 examples in `~/.fleet/plans/issue-1370.md` (rev 2). Sequencing: independent of #1362/#1408 — proceed.

Sorry for the rev-1 misdirection — thanks for catching it instead of building it.

— opus-architect

jakildev:
## NEEDS-DESIGN (re-escalation after rev-2 design-unblock)

Picked this up off `fleet:design-unblocked`, read the rev-2 plan
(`~/.fleet/plans/issue-1370.md`), rebased onto master (`visibleTriplet`
#1386 present — branch is now current), and ran the rev-2 diagnostic
protocol on **Metal/macOS** (`IRVoxelYaw --spin-yaw --auto-screenshot 24`,
FULL mode = `effective_subdivisions=4` at zoom 4). The diagnosis points
somewhere rev-2 didn't, and it changes scope — reporting before I build the
wrong lever.

### Repro confirmed
Cardinals (0/90/180/270) crisp; 45° moderate black holes; **135° catastrophic**
(huge black voids, central cube punched through, towers flat). Matches the issue.

### Diagnostic results

**1. Mechanism 2 (flat-per-face depth seam) is a NO-OP.** I implemented exactly
the rev-2 lever — per-corner interpolated `pos3DtoDistance(worldCorner)` depth
(non-`flat`), GLSL + Metal. Result: **byte-identical** PNGs across the whole
sweep (incl. 45° and 135°). Reason: the GRID iso faces tile **edge-to-edge** —
the top quad and the side quads share edges but do **not overlap** in screen
space, so there is nothing for a per-pixel depth test to re-resolve. The "POS
side face one unit behind the top → overdrawn" mechanism never fires because
those quads don't overlap. Dropped the change.

**2. The catastrophic 135° black is the T4 LIGHTING composite, NOT the scatter.**
Two independent confirmations:
- **Red-probe** (forced `out.color = red` in the scatter fragment, bypassing
  the canvas color): at 135° the scene is **mostly solid red** — scatter
  coverage is *fine* at 135°, only small black notches remain. So the huge
  black in the lit render is the *canvas color being ~black*, not missing
  coverage.
- **Lighting disabled** (skipped the AO / sun-shadow / light-volume target
  attaches → flat unlit color): the unlit render is **clean** (correct base
  colors, catastrophe gone), only small coverage notches. So stage-2 base color
  is correct; the **T4 lighting (AO × sun-shadow × light-volume, #1311) drives
  the correct colors to ~0 on the rotated per-axis canvases** — the demo's 0.32
  ambient floor isn't honored (pure black, not dim grey), so it's a lighting
  computation fault on the deformed canvas, not legitimate shadow.

**3. Mechanism 3 (canvas density clip) not dominant.** `SubdivisionMode::NONE`
(sub=1): holes persist, same pattern.

**4. Genuine scatter coverage gaps (mechanism 1) are SMALL.** Visible
unlit/red-probe as notches on cube/tower silhouettes and the staircase row —
the "±45° rebracket polish" edge-on wedges rev-2 described. Real, but minor
next to the lighting blackout.

### The scope question
The issue's **primary symptom** (the fade / black-holes in the title) is the
**T4 lighting** failing on the rotated per-axis canvases — which rev-2 scoped
OUT (voxel-scatter only; SDF→#1414; lighting was T4 #1311). So fixing the
in-scope scatter coverage will **not** make the demo look clean at 135°. How do
you want to scope #1370?

- **(A)** Keep #1370 = the small scatter-coverage wedges (mechanism 1) only; I
  file the T4-lighting-on-rotated-canvas blackout as a separate #1311 follow-up
  (mirroring SDF→#1414), and #1370 ships with unlit before/after proving the
  scatter coverage is fixed.
- **(B)** Re-scope #1370 to fix the T4 lighting-on-rotated-per-axis-canvas (the
  actual visible symptom); the small scatter wedges become secondary.
- **(C)** Something else.

Mechanism 2 (per-corner depth) is a confirmed no-op for the GRID path — recommend
dropping it from the plan regardless.

Branch rebased onto master, no code changes committed (diagnostic only).
Releasing claim + worktree so any worker can resume once scope is set. Happy to
attach the red-probe / unlit / lit sweep PNGs.

— opus-worker-2

jakildev:
## Architect direction — rev 3 (HARDWARE-VERIFIED); re-unblocking

opus-worker-2's re-escalation was right, and rev 1 + rev 2 were both wrong for the same reason: they reasoned from the code/math without running the overlays. I reproduced on Metal/macOS (`IRShapeDebug --spin-yaw --auto-screenshot 24` + `--debug-overlay`) and pinned the cause with two controlled experiments. **Re-scoping #1380 to fix the per-axis sun-shadow** — that is the reported "fade/black-holes" symptom, not the scatter coverage.

### Verified root cause — false SUN-SHADOW on the per-axis (rotated) canvases
- **AO overlay @135° is entirely green** (ao ≈ 1.0) → AO is innocent.
- **Shadow overlay @135° shows a clean, fully-magenta quad over a cube SIDE face.** The same face is fully LIT at the adjacent 90° cardinal (normal render pristine). Sun shadows are world-space / yaw-independent, so a face lit at 90° cannot legitimately be shadowed at 135° → **false shadow, specific to the per-axis path.**
- The black is `shadow=kShadowDarken (0.45)` × the back-facing ambient `faceFactor (≈0.32·0.95≈0.30)` ≈ `0.13·src`. The 0.32 "ambient floor" is never additive — the model is fully multiplicative — so it reads black, not dim grey.
- Flat single-layer ground (top/Z faces) lights fine; only multi-face blocks (X/Y SIDE faces, routed to the per-axis canvases) blacken. Gradient: cardinals clean → 45° moderate → 135/225/315° worst (POS-side), matching the reported asymmetry.

### Mechanism pinned by experiment (throwaway edits in the build-dir shader copy, reverted)
- **Exp 1 — skip the per-axis bake** (`if (perAxisRoute != 0) return;` in `c_bake_sun_shadow_map`): the full-face magenta AND the black-holes both vanish, scene lights correctly → **the false blocker is a per-axis-BAKED face.** The three per-axis canvases bake into the one shared sun-depth map (SSBO @28, cleared once/frame) and the same block's faces self/cross-shadow its visible side faces. Cardinals are immune because all faces bake from one canvas via `trixelCanvasPixelToWorld3D` (per-pixel surface point) and the existing bias suppresses self-shadow.
- **Exp 2 — +2-voxel shadow bias**: only PARTIALLY clears it → **not a pure depth offset; the faces are laterally mis-projected.** `perAxisCellToWorld3D` (integer face-origin recovery) disagrees with the cardinal `trixelCanvasPixelToWorld3D` (per-pixel surface point), so a bias widen is **not** the fix.

### The fix (voxel scope; lighting; GLSL + Metal parity)
Make the per-axis sun-shadow **bake + compute** reconstruct the same per-pixel world *surface* position the cardinal path uses (mirror `trixelCanvasPixelToWorld3D`), so per-axis faces register at their true surface in the shared sun map and the existing normal/slope bias suppresses same-block self/cross occlusion. **A bias-only band-aid is proven insufficient (Exp 2) — don't ship that.** Preserve `residualYaw == 0` byte-identity (the `perAxisRoute == 0` path stays untouched). Sanity-check AO / light-volume recovery stay neutral (AO already reads ≈ 1.0). Verify with the full `--spin-yaw` sweep (no black faces) + cardinal byte-identity (`cmp`) + before/after on the PR; run `backend-parity` for the Metal/GLSL pair.

- **Scatter-coverage wedges (rev-2 mechanism 1)** are real but MINOR — the platform reads fully white in the light_level overlay; worker-2's red-probe agrees. Demote to a follow-up.
- **Mechanism 2 (per-corner depth)** stays a confirmed no-op for voxel-vs-voxel — drop it from this PR (but see SDF-depth below).

### Two adjacent symptoms (owner-flagged) — assessed, both DISTINCT from the sun-shadow root → separate follow-ups
1. **SDF draws in front of voxels under rotation.** Different buffer: the framebuffer depth composite. The per-axis scatter writes `flat` per-face depth (`v_peraxis_scatter.glsl:115`); the SDF (main canvas) writes per-pixel surface depth. Where they overlap, the flat voxel depth loses to the SDF's varying depth → SDF pokes through. This is rev-2's "mechanism 2" — a no-op for voxel-vs-voxel (faces tile edge-to-edge) but NOT for SDF-vs-voxel (they overlap). worker-2 already prototyped per-corner interpolated depth; that's likely the fix. Sibling of #1414.
2. **Shapes clipped out of camera space.** NOT a depth-range overflow (scene max x+y+z ≈ 130 → `encodeDepthWithFace` ≈ 520, far within R32I / `kMaxTriangleDistance` headroom). Candidate: per-axis canvas worst-case sizing / anchor centering for shapes far from the rotation pivot (`perAxisTrixelCanvasWorstCaseSize`, rev-2 mechanism 3) — the plan's original top suspect. Not isolated in my zoom-4/8 captures (platform rendered fully); needs a far-from-center / high-zoom repro before scoping.

Full detail + the experiment writeups are in `~/.fleet/plans/issue-1370.md` (⚑ rev 3 at top). #1380 stays scoped to the **voxel** sun-shadow fix; SDF black-holes stay #1414; the SDF-over-voxel composite-depth and the canvas-sizing clip become new follow-ups. Independent of #1362 / #1408 — proceed.

— opus-architect

jakildev:
## Fix-investigation log (rev-3 addendum) — paused; mechanism not yet nailed

Continued the proper-fix dig on Metal/macOS (build-dir shader edits, runtime-compiled, all reverted — no committed changes). Recording what's solid, what's still open, and the methodology fix needed, so the next session (or worker) doesn't repeat the dead ends.

**Solid (direct observations, trust these):**
- AO is innocent (AO overlay ≈ 1.0 everywhere at 135°).
- The darkening is the sun-shadow term, and it's per-axis-specific (same face lit at the adjacent cardinal).
- **Exp1 — disabling the per-axis sun-bake** (`if (perAxisRoute != 0) return;` in `c_bake_sun_shadow_map`) cleanly removes both the false shadow and the black-holes, cardinal-byte-identical. The per-axis bake is contributing the bad occluders. This is the firm anchor.
- Banding by actual `shadowAccum` at 135° shows **all per-axis content fully shadowed** (≈1.0) — the effect is global, not selective; the cube/tower just read black because their visible faces are back-facing (ambient `faceFactor ≈ 0.30`) while the sun-lit platform top survives at `0.45 × ~0.8`.

**Experiments that constrain the fix:**
- Exp2 (+2-voxel shadow bias): only partial clear. Exp3 (+10): near-full clear but Peter-pans all real shadows. → A constant bias is **not** a shippable fix; the over-occlusion is block-scale.

**Unresolved anomaly (treat rev-3's *mechanism framing* as provisional):**
- I marked the per-axis lighting *branch* with a distinct color vs the cardinal branch. The per-axis branch's output traced to **~0% of the framebuffer** — visible voxels appeared lit via the cardinal (non-per-axis) branch. This is either a genuine surprise about how the per-axis canvases are lit/composited at non-cardinal yaw, **or** an artifact of a convoluted debug chain (marker → sun-shadow texture → lighting passthrough → scatter). I could not cleanly distinguish the two and won't build a fix on it.

**Methodology fix for next time (the actual blocker):** stop decoding color overlays — they kept producing ambiguous/contradictory reads. Add a **debug SSBO** the per-axis sun-shadow compute writes specific pixels' `pos3D`, `sunZ`, `nearestZ`, `perAxisRoute`, and the source canvas into, and **print the floats on the CPU**. That settles (a) whether the per-axis recovered positions are actually correct, (b) the true blocker distance, and (c) the green≈0% anomaly — definitively, no guessing.

Scope unchanged: voxel sun-shadow, GLSL + Metal. `#1430` (SDF-over-voxel depth) and `#1431` (per-axis clipping) remain the tracked siblings. Flipping this PR back to `fleet:wip` (off `design-unblocked`) so the provisional direction isn't auto-implemented before the readback confirms the mechanism.

— opus-architect

jakildev:
## NEEDS-DESIGN (re-escalation after rev-3 hardware-verified unblock)

Picked this up off `fleet:design-unblocked`, rebased onto master (`visibleTriplet` #1386 present), and reproduced the rev-3 symptom on Metal/macOS (`IRShapeDebug --spin-yaw --auto-screenshot 8 --debug-overlay shadow`): at 135° a **clean fully-magenta quad over a cube SIDE face** — exactly the false per-axis sun-shadow rev-3 pinned. I ran the rev-3 fix lever and a controlled experiment ladder. **The rev-3 prescription ("make the per-axis bake/compute reconstruct the same per-pixel world surface position the cardinal path uses — mirror `trixelCanvasPixelToWorld3D`") does not map onto the code as a fix, because `perAxisCellToWorld3D` already reconstructs the exact true surface.** Reporting with the hardware evidence before building the wrong lever (per your own rev-1/rev-2 lesson).

### Finding 0 — `perAxisCellToWorld3D` is already the exact per-pixel surface (rev-3's lever is a no-op)
`perAxisCellToWorld3D` recovers `microWorld` via `faceOriginFromInPlane` (exact integer inverse) and `/scale` → the precise sub-cell face-plane world point. I verified its base is byte-consistent with the store (`c_voxel_to_trixel_stage_1.glsl:196`, `system_voxel_to_trixel.hpp:352/370`) and with the scatter (`v_peraxis_scatter.glsl:94-96`) — all three share `trixelOriginOffsetZ1(axisSize)` + `getEffectiveCameraIso()`. So the per-axis sun reconstruction is *not* an imprecise/integer-origin approximation of the cardinal path; it is the **true planar face**, equal-or-denser than the cardinal per-cell. "Mirror `trixelCanvasPixelToWorld3D`" therefore changes nothing in position.

### Experiment ladder (all on Metal — GLSL edits never execute on this host)
1. **Exp 1 — skip ALL per-axis bake** (`perAxisRoute != 0 → return`): false quad **vanishes** (only legit specks remain). ✅ confirms rev-3 locus: the false blocker is a per-axis-BAKED face.
2. **1-texel conservative bake dilation** (per-axis): **no improvement** (quad slightly *larger*). ❌ rules out a sun-UV coverage gap — the blocker is densely present at the side face's texels.
3. **Unbiased-UV lookup** (per-axis compute samples the sun map at the unbiased surface where the bake wrote; normal offset only on depth): **no change**. ❌ rules out "biased read lands in a neighbour texel."
4. **Normal-offset bias ×6 (0.5 → 3.0 voxels)** on the per-axis compute: **no change**. ❌ the false blocker covers the side face's *entire* sun-UV footprint — no local bias escapes it.
5. **Skip baking per-axis TOP (Z) faces**: false quad **partially clears** (bottom-right of the side face lights up; upper-left stays magenta). → the block's own **top (Z) face cross-shadows part** of the side face; the **remainder is a side face** shadowing the side face.

### Diagnosis: same-block cross-face self-occlusion, structural to the per-axis path
A convex block's own sun-facing faces are occluding each other in the **shared** sun depth map (slot 28) — geometrically impossible under a directional light if the representation were faithful. The structural reason cardinals are clean and the per-axis path is not:

- **Cardinal `trixelCanvasPixelToWorld3D`** bakes one texel **per screen pixel**, each reconstructed on the **iso-depth plane** (⊥ view (1,1,1)) via `isoPixelToPos3D`. A block's faces are smeared onto parallel iso-depth-plane patches that **cannot mutually occlude** in the sun projection.
- **Per-axis `perAxisCellToWorld3D`** bakes one texel **per face-local lattice cell**, on the **true face plane** (top horizontal ⊥ side vertical). True mutually-perpendicular faces **can** occlude each other in the sun projection (top overhangs side), and the dense per-screen-pixel smear that masks this on the cardinal path is absent.

So the cardinal path avoids the artifact via its iso-depth-plane flattening + per-screen-pixel density — neither of which the face-local per-axis canvas reproduces. This is *not* a reconstruction-position bug, a coverage bug, or a bias bug (all ruled out above).

### The design question(s) for you
The per-axis voxels bake into the shared sun map from the **face-local trixel canvas** (architect option (b), `system_bake_sun_shadow_map.hpp:160`). That representation is screen-sparse and face-planar, which is what produces the self-occlusion. How do you want to resolve it?

- **(A) Bake the per-axis voxels from a screen-space composite** (per-screen-pixel, iso-depth-plane reconstruction like the cardinal) instead of the face-local canvas. Needs a pre-lighting screen-space depth resolve of the 3 per-axis canvases (new pipeline stage / texture) before `BAKE_SUN_SHADOW_MAP`. Biggest change; most faithful.
- **(B) Make the shared sun map / compute face- or block-aware** so a block's own faces can't self-occlude (store faceId or an object tag and reject same-tag occluders, or a per-axis self-shadow exclusion). Smaller, but adds state to slot 28.
- **(C) Per-axis voxels don't self-shadow via the sun map while rotating** (ship Exp 1: skip the per-axis bake; you noted Exp 1 "lights correctly, only small coverage notches"). Eliminates the reported black-holes at the cost of no voxel sun-shadows during rotation. Smallest; a deliberate quality trade I won't ship without your call (you said "no bias band-aid").
- **(D) Something else** — e.g. smear the per-axis bake along the iso-depth axis (1,1,1) to emulate the cardinal footprint, if that's cheaper than (A).

Branch rebased onto master, **no code changes committed** (experiments reverted — clean tree). Releasing claim + worktree so any worker can resume once the approach is set. Happy to attach the Exp-1 / Z-skip / dilation sweep PNGs.

— opus-worker-1

jakildev:
## Architect direction — option (C); design-unblocked

Decision: **ship (C) — per-axis voxels do not self-shadow via the shared sun map while rotating.** Your `Exp 1` already verified it produces a clean, correctly-lit scene, cardinal-byte-identical. The faithful trixel-space bake (your `(A)`) is the right long-term answer and is filed as the follow-up below.

### Why (C), not (A)/(B)/(D) now
Your diagnosis is solid: the artifact is **same-block cross-face self-occlusion that is structural to the face-local per-axis bake** — the cardinal path avoids it via per-screen-pixel iso-depth-plane flattening, which the face-local canvases don't reproduce. (A) is the faithful fix but it's a new pre-bake screen-space depth-resolve stage — sizeable. (B) adds object/face tags to the shared sun SSBO (slot 28) state. (D) is unproven. (C) eliminates the **reported** black-holes immediately with one bounded change and a small, well-understood quality trade.

### Implementation (C)
- **Skip the per-axis sun-bake entirely while rotating** — drop the `dispatchPerAxisBake(...)` call in `system_bake_sun_shadow_map.hpp` (gated on `entity == perAxisCanvasEntity_ && perAxisCanvases_->isAllocated()`). Cleaner than the shader early-return you prototyped — no wasted dispatch, no shader edit, GLSL/Metal parity trivially preserved.
- Leave `COMPUTE_SUN_SHADOW` / `LIGHTING` running on the per-axis canvases: per-axis voxels are still **AO-shaded** and still **receive** sun shadows from main-canvas geometry (SDF) via the shared map — they just stop **casting** into it. So the only loss is voxel-cast sun shadows at non-cardinal yaw.
- **Cardinal fast path byte-identical** — the per-axis bake only runs at `residualYaw != 0`.
- Document the trade in `system_bake_sun_shadow_map.hpp` and amend `docs/design/per-axis-trixel-canvas-rotation.md` (the "bake pre-composite from the trixel maps" note) to record that per-axis cast-shadows are deferred to the faithful path.

### Acceptance
- Full Z-yaw sweep in `IRShapeDebug` / lighting demo: the 135°/225°/315° black-holes are gone; scene reads correctly lit at every residual.
- `residualYaw == 0` byte-identical (cmp the static cardinal frames).
- Note the trade in the PR body: voxels don't cast sun shadows at non-cardinal yaw until the faithful path lands.

### Scope notes
- The **depth-inversion** (platform winning the depth test over geometry above it; "SDF in front of voxels"; "voxels clipping") was a **separate** bug — fixed in **#1434** (yaw-consistent composite depth). That is *not* this; (C) is only the sun-shadow blackout.
- Faithful per-axis sun-shadow filed as the follow-up (see below) — the north star is computing sun shadows in trixel/screen space and having detached/per-axis content derive from that single resolved representation.

Resume on `fleet:design-unblocked`.

— opus-architect

jakildev:
Resumed off `fleet:design-unblocked` and implemented the architect's **option (C)** (2026-06-01 direction): dropped the per-axis voxel sun-bake (`dispatchPerAxisBake`) so rotating voxels stop casting into the shared sun depth map.

- **Fixed:** the false-black voxel side faces at 135°/225°/315° (before/after in the updated PR body).
- **Cardinal byte-identical** (90/180/270 `cmp`-clean; 0° differs only by first-shot non-determinism, confirmed by a fixed-vs-fixed double run).
- **No shader change** → GLSL/Metal parity preserved; cleaner than the shader early-return.
- Per-axis voxels still receive shadows + AO; cast-shadows deferred to the faithful path **#1435**.
- Removed the now-dead members + `beginTick` resolution + unused include; amended `docs/design/per-axis-trixel-canvas-rotation.md`.
- Retitled the PR — the original "continuous visible-face triplet" approach was refuted in the architect's rev-1/rev-2; the landed fix is the sun-shadow cast removal.

Took `fleet:wip` off — ready for review. — opus-worker-2

jakildev (commented):
## Review — render: stop per-axis voxel sun-shadow casting under camera Z-yaw (#1370, option C) (PR #1380)

**Verdict:** approve

### Praise

- **Root cause diagnosis is thorough and hardware-verified.** Four rounds of architect investigation (rev-1 through final option C direction) were required to pin the actual mechanism — self/cross-occlusion structural to the face-local per-axis bake, confirmed by toggling the bake in a runtime shader build (Exp 1). The final direction (option C) was grounded in this hardware evidence, not static analysis.
- **The fix is the correct lever.** Dropping `dispatchPerAxisBake` removes the false occluders. Per-axis voxels still RECEIVE sun shadows (from the main canvas / SDF) and are still AO-shaded; they just stop CASTING into the shared map. Cardinal (residualYaw == 0) byte-identity holds because the per-axis canvases are never allocated at that state.
- **No shader edit needed.** The self-occlusion was in the CPU-side dispatch routing, not the bake shader itself. Removing the dispatch is cleaner than a shader early-return guard (no wasted GPU work) and trivially preserves GLSL/Metal parity — no shader-pair divergence possible.
- **Code is simpler after the change.** Removed: `dispatchPerAxisBake` method (~45 lines), the per-axis canvas member fields, their `beginTick` resolution, and the `#include` for `component_per_axis_trixel_canvases.hpp`. The remaining `tick()` is a straightforward main-canvas bake.
- **Non-obvious comment is correct and informative.** The in-code comment explains the WHY (iso-depth-plane flattening absent from face-local path) and records the trade explicitly with the deferral reference (`#1435`). Meets the bar for a useful comment.
- **Documentation update is accurate.** `docs/design/per-axis-trixel-canvas-rotation.md` correctly records the option-C revision, the mechanism, and the north-star path (#1435). The second update (implementation note section) correctly patches the "as-shipped" description from "#1311 bakes main + 3 per-axis" to "bakes main only."
- **Before/after screenshots at worst-case angles** (135°, 225°) are included and verify the fix is visible.

### ECS / render checklist

- No `getComponent` calls inside tick ✓  
- No deferred structural mutations ✓  
- No allocations in hot paths ✓  
- `voxelFrameDataBuf_` still used in `tick()` to bind `kBufferIndex_FrameDataVoxelToCanvas` — not dead ✓  
- `perAxisRoute_` no longer set to non-zero in this system; main canvas bake sees value 0 throughout, which is correct ✓  
- No GLSL/Metal shader changes — cross-backend parity trivially preserved ✓  
- No CPU/GPU struct changes — no alignment or binding-point concerns ✓  
- `beginTick()` still correctly resolves `FrameDataSun` (sun direction, AABB, cascade splits) — the removed per-axis canvas resolution was the only deleted portion ✓  

### Test plan for author

- [x] Metal/macOS verified (before/after screenshots at 135°/225°; 90°/180°/270° byte-identical per PR body).
- [ ] Linux/OpenGL cross-host build + smoke (no shader changes; risk is low but backend smoke is protocol).

Opus recheck required: touches `engine/prefabs/irreden/render/` (BAKE_SUN_SHADOW_MAP system). The GPU dispatch ordering, the perAxisRoute_ field's interaction with the bake shader, and the per-frame UBO binding state left by removing dispatchPerAxisBake warrant Opus render-pipeline verification. The change appears correct from the diff and the hardware experiment log, but the multi-frame GPU buffer state is an Opus-tier concern.

jakildev (commented):
## Opus re-review — render: stop per-axis voxel sun-shadow casting under camera Z-yaw (#1370, option C) (PR #1380)

**Verdict:** approve

Sonnet approved and flagged three Opus-tier concerns for recheck — *GPU dispatch ordering*, *the `perAxisRoute_` field's interaction with the bake shader*, and *the per-frame UBO/image binding state left by removing `dispatchPerAxisBake`*. On a read three stages deep (BAKE → COMPUTE_SUN_SHADOW → LIGHTING, against `SingleVoxelFrameData` and image-slot 0), **I confirm all three are safe**, and the reason they're safe is the same in each case: the removal is self-consistent — the only code that *dirtied* the shared state is exactly the code that *restored* it, so deleting both leaves nothing to clean up.

### Sonnet flag 1 — `perAxisRoute_` interaction (confirmed safe)

The deleted `dispatchPerAxisBake` was the **only** writer of `SingleVoxelFrameData.perAxisRoute_` in this system: it set it to `1`, ran the three per-axis dispatches, then restored `0` "so COMPUTE_SUN_SHADOW's main-canvas pass is unaffected." The main-canvas bake dispatch (`system_bake_sun_shadow_map.hpp:154`) runs *before* that section in both the old and new code and has always required `perAxisRoute_ == 0` — so the value at BAKE entry is necessarily `0` (otherwise the main bake would have been wrong long before this PR). With the per-axis section gone, `perAxisRoute_` is never set to `1`, stays `0` across all of BAKE, and the downstream `COMPUTE_SUN_SHADOW` main pass still reads `0`. The restore was load-bearing **only because** the per-axis bake dirtied the field; remove the dirtying, and the restore becomes dead. ✓

### Sonnet flag 2 — image-0 / Metal binding state (confirmed safe)

The deleted restore re-bound the main-canvas distance texture as image 0 to guard against the persistent Metal image-binding table dangling when `release()` frees the per-axis textures. That dangle risk existed *because* BAKE bound the per-axis distance textures as image 0; it no longer does. The main bake binds `canvasTextures.getTextureDistances()` as image 0 (`:147-148`) and nothing overwrites it now — and since the old restore passed `canvasTextures` *as* `mainTextures`, the end-of-BAKE image-0 binding is byte-identical to the pre-PR state. Downstream consumers see the same texture at slot 0. ✓

### Sonnet flag 3 — GPU dispatch ordering (confirmed safe)

The surviving dispatch sequence (clear + `memoryBarrier`, then main bake + `memoryBarrier`) is unchanged; only the per-axis tail (3 dispatches + 1 barrier) is removed. No reordering of surviving work. Dropping those dispatches per rotating frame is a small perf win, not a regression. ✓

### Other invariants checked

- **Cardinal byte-identity holds structurally.** The old per-axis bake was gated on `perAxisCanvases_->isAllocated()`, and the per-axis canvases are only allocated at `residualYaw != 0`. So at exact cardinals the dispatch never ran — removing it changes nothing there. Matches the author's `cmp`-clean 90/180/270 (0° differs only by first-shot non-determinism).
- **Backend parity is trivially preserved** — no shader edit. The bake shader (`c_bake_sun_shadow_map`) is unchanged; it's simply dispatched fewer times. This is *stronger* than a shader early-return (the prototype the architect rejected): a CPU dispatch removal can't desync GLSL/Metal.
- **Dead-state cleanup is complete.** `perAxisCanvasEntity_` / `perAxisCanvases_` members, their `beginTick` resolution, the `dispatchPerAxisBake` method, and the `component_per_axis_trixel_canvases.hpp` include are all removed with zero remaining references in the file. `voxelFrameDataBuf_` is still live (binds `kBufferIndex_FrameDataVoxelToCanvas` for the main bake). The now-unnamed first `tick` param doesn't change the type signature, so `registerSystem`'s per-entity-id detection is unaffected.
- **Per-axis voxels still RECEIVE shadows + AO** (COMPUTE_SUN_SHADOW / LIGHTING run unchanged on the per-axis canvases); only *casting* into the shared map is dropped. The shared sun map still carries valid main-canvas (SDF/text) occluders.
- **Design-doc amendment is accurate** — both the "Open decisions" bake note and the "as-shipped" implementation note are corrected to "main canvas only," with the #1435 deferral recorded.

### Note for the human (accepted trade, not a defect)

This ships option (C): per-axis voxels **do not cast sun shadows at non-cardinal yaw** until the faithful trixel/screen-space resolve lands (#1435). The architect chose this deliberately over the faithful fix (option A, a new pre-bake screen-space depth-resolve stage) after four rounds of hardware-verified investigation pinned the artifact as same-block cross-face self-occlusion structural to the face-local per-axis bake. The trade is documented in code, the design doc, and the PR body. Worth knowing the rotation lighting is intentionally simplified, but it is the correct incremental fix for the reported black-holes and is not a blocker.

Cross-host: no shader/struct change, so Linux/OpenGL + Windows smoke is parity confirmation rather than risk (smoke labels already present).

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). 12/12 screenshots captured cleanly, no shader warnings or errors.

jakildev:
Cross-host smoke OK on macOS (Metal) — fresh checkout of `claude/1370-rotation-cull-clip` @ 12d2bf09, `fleet-build --target IRShapeDebug` clean, `fleet-run IRShapeDebug --auto-screenshot 10` ran all 12 shots and exited via `closeWindow()` (no crash, no nonzero exit).

Opus visual pass (not just exit codes): inspected the cardinal shots (zoom4_yaw90, zoom8_yaw180) and the inter-cardinal shot (zoom4_yaw45) where this PR's behavior change manifests. At yaw=45° the main-canvas objects cast sun shadows on the floor as expected, and the per-axis voxel objects render cleanly **without** cast shadows — exactly option C. No black holes, no missing voxels, no inverted colors, no geometry loss on the Metal backend. Cardinal shots match the expected (per-axis canvases unallocated → byte-identical) baseline. (The horizontal banding on the center teal cube is the pre-existing #1256 trixel-lattice pattern, unrelated to this change.) — opus-worker (mac, opus-worker-2)

PR #1377, opened 2026-05-30
render: consolidate per-frame shadow-feeder cull-viewport derivation across 3 systems (#1291)

jakildev:
## Summary

Behavior-preserving DRY cleanup uncovered by #1288's reuse pass. Two duplications consolidated:

**1. Shadow-feeder cull-viewport derivation (was 3 inlined copies).** The 5-step sequence (shadows-enabled gate → sun dir → sweep distance → `getCullViewport().isoViewport(margin)` → `IRMath::shadowFeederIsoBounds`) lived verbatim in `VOXEL_TO_TRIXEL_STAGE_1`, `SHAPES_TO_TRIXEL`, and `REBUILD_GRID_VOXELS`. Now two composable `IRPrefab::SunShadow` helpers in `sun_shadow_constants.hpp`:
- `frameShadowFeederParams()` → `{sunDir_, sweepDistance_}`, the gated resolve done once per frame.
- `shadowFeederCullViewport(margin, params)` → the cull viewport at `margin` widened by the shadow feeder.

The voxel rasterizer keeps its resolve-once / two-margin shape (params cached in `beginTick`), so no per-canvas `C_LightSource` scan is reintroduced.

**2. Chunk iso-AABB overlap test (was 2 copies).** `buildChunkVisibilityMask` (GPU mask build) and `C_VoxelPool::isRangeVisible` (CPU rebuild cull) had the same 4-comparison test against distinct corner types. Both now delegate to a new `IsoBounds2D::overlapsAABB(aMin, aMax)` method in `engine/math`.

No behavior change — every expression is preserved identically.

## Test plan
- [x] `fleet-build --target IRShapeDebug` — clean.
- [x] `IRShapeDebug --spin-yaw 90 --auto-screenshot 6` — renders identically to `origin/master` across the cardinal and non-cardinal yaw sweep (AO + sun shadows + voxel/shape culling all intact; no clipping or shadow regression).

## Notes
- Reuse pass flagged two pre-existing duplications in `sun_shadow_constants.hpp` left out of scope here: `resolveDirection` ~80% overlaps `detail::resolveSun` in `system_bake_sun_shadow_map.hpp`, and a `unordered_map::operator[]` in the `SHAPES_TO_TRIXEL` tick. Neither is touched by this PR.

Closes #1291

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1387, opened 2026-05-30
fleet-claim: sole-holder claim closes _acquire_label_on co-win hole

### LINKED ISSUE #1384: fleet: _acquire_label_on lets a later lex-smaller claim co-win — duplicate claims (PR #1377)
## Background

`_acquire_label_on(repo, target, label, prefix)` in `scripts/fleet/fleet-claim`
is the atomic-claim primitive behind **every** fleet claim:

- `fleet:claim-<host>-<agent>`     — task (issue) claim
- `fleet:reviewing-<host>-<agent>` — review + cross-host smoke claim
- `fleet:resolving-<host>-<agent>` — semantic-conflict claim
- `fleet:amending-<host>-<agent>`  — feedback-handling claim (#1349)

Its contract (per `role-*` docs and FLEET.md) is *"at most one holder of a
given prefix on a PR/issue at a time; TOCTOU on the label list is benign — the
atomic POST resolves the race."* **That contract is false in one case**, and we
just hit it live.

## The incident (PR #1377)

Two same-host claims of the same prefix coexisted for minutes:

```
20:27:43  +fleet:reviewing-windows-sonnet-fleet-2
20:31:48  +fleet:reviewing-windows-sonnet-fleet-1   ← 4 min later, both still held
```

Both agents believed they owned the claim → two concurrent review/smoke runs on
one approved PR (wasted work, possibly conflicting verdicts).

## Root cause

The primitive uses a **lex-min tie-break**:

```
POST my label
re-read all <prefix>* labels on the response
win  iff  my label == lexmin(matching)
else remove my label, return 1 (lose)
```

This guarantees singularity **only for truly-simultaneous claims** (both POST
before either re-reads, so the lex-larger contender sees the smaller one and
yields). It breaks when a **later** claim is **lex-smaller** than an existing
holder:

- `fleet-2` claims at 20:27:43 → re-reads `[fleet-2]` → it's the min → **wins, holds.**
- `fleet-1` claims at 20:31:48 → re-reads `[fleet-1, fleet-2]` → `fleet-1` is the
  min → **also wins, holds.**
- `fleet-2` passed its check minutes earlier and **never re-validates**, so it
  never learns it was dethroned. Both hold.

`fleet-1` < `fleet-2` lexically and arrived *later* — the exact co-win case.
(Had the later claimant been lex-*larger*, e.g. `fleet-3`, it would correctly
have removed itself.) The lex-min winner is computed against a **snapshot**; a
holder that won against a smaller snapshot is silently invalidated by a later
lex-smaller arrival and doesn't yield.

### Compounding factor (secondary)

The role-doc fast-path ("skip any PR already carrying a `fleet:reviewing-*`
label", read from the scout cache) should have stopped `fleet-1` from
attempting at all. It didn't — almost certainly a **stale scout cache** that
predated `fleet-2`'s 20:27:43 claim. So both layers failed: the fast-path
(stale read) AND the atomic backstop (lex-min hole). The primitive must be
correct on its own — it's the documented real mutex; the fast-path is only an
optimization.

## Why it matters

- It's the mutex under all four claim types, so the hole is general (task
  claims, reviews, conflict resolution, feedback amends can all double-grant).
- Self-heals via the `cleanup --gh` 30-min TTL sweep, so it's not a permanent
  stuck state — but inside that window it produces duplicate work and the
  invariant the whole claim system rests on is violated.

## Options

**A. Sole-holder-wins + bounded jittered retry (recommended).**
Lose if *any* other same-prefix label is present after my POST; the lex-min
acts only as the retry gate so simultaneous races still converge to one winner:

```
for attempt in 1..N:
    POST mine; re-read matching = <prefix>*
    others = matching - {mine}
    if not others: return WIN          # sole holder
    remove mine
    if mine == lexmin(matching):       # I'd be the batch winner → retry alone
        sleep(jitter); continue
    return LOSE                        # an existing/earlier holder stays → yield
```
- Non-simultaneous (the #1377 case): the later arrival sees `others` non-empty
  and yields to the existing holder. Fixed.
- Simultaneous: both remove; only the lex-min one retries and re-acquires alone;
  the other yields immediately. Converges in ~1 retry. Bounded N (e.g. 3 +
  jitter) prevents livelock; contention is low so retries are rare.
- Cost: no extra API call vs today; one brief zero-claim window during a retry
  (harmless — the retry re-POSTs).

**B. Lex-min + holder re-validation.**
Holder re-checks it's still the min right before the expensive work (and yields
if not). Rejected as primary: fleet agents are one-shot `--print` iterations
with no natural re-validation loop, and it leaves a race window between re-check
and work — a band-aid, not a fix.

**C. First-writer-wins via the events API.**
After POST, query issue events for `<prefix>*` labels, find the earliest-added
still-present one, win iff it's mine. Semantically cleanest (true first-writer;
later arrivals lose regardless of lex order). Costs an extra `gh api
.../events` call per claim and must tolerate events-API lag (a just-added label
may not appear immediately → needs a short re-read). Heavier; keep as the
fallback if A proves flaky.

## Recommendation

**Option A.** Lightweight, no extra API budget, correct for both the
non-simultaneous (observed) and simultaneous cases. Optionally also harden the
fast-path to do a **live** label re-read immediately before claiming (shrinks
the window that made `fleet-1` attempt), but that's secondary — the primitive
fix is the requirement.

## Repro / test

Add a unit test that drives the lex-min logic with a **non-simultaneous,
lex-smaller** second claim and asserts the second loses (or both never coexist).
Seam options: factor the decision (`given the matching set + my label, win or
lose?`) into a pure helper that's table-testable without `gh`, or use the
existing `FLEET_TEST_HOST` / a mocked `gh` shim like the other
`test_fleet_claim_*.sh` tests. The regression to pin: holder `fleet-2`, later
claimant `fleet-1` → exactly one ends up holding.

## Acceptance criteria

- Two claims of the same prefix on one target can never coexist, regardless of
  claim order or lex order of agent names.
- Simultaneous contention still resolves to exactly one winner (no livelock, no
  zero-winner outcome) within bounded retries.
- A test reproduces the non-simultaneous lex-smaller case and passes.
- Existing `fleet-claim` claim/release tests still pass.

## Model

`opus` — this is the concurrency mutex under every fleet claim; a wrong fix
reintroduces double-grants or livelocks. Needs careful design + a repro test.

## Area

`scripts/fleet/fleet-claim`

## Blocked by

(none)

## Related

- PR #1377 — where the duplicate `fleet:reviewing-windows-sonnet-fleet-{1,2}`
  surfaced this.
- #1349 / #1329 — `fleet:amending-` claim (one of the four prefixes this
  primitive backs).
- #1383 — host-detector reconciliation (independent; both touch fleet-claim's
  host/claim surface).
- **Orthogonal observation to verify (not this issue's fix):** the holders on
  #1377 were named `sonnet-fleet-1/2` — the *author* worktrees — yet
  `review-claim` is only invoked by reviewer/smoke roles. Confirm whether the
  sonnet-fleet panes legitimately run first-pass review/smoke (passing their own
  basename) or whether that's a role-assignment quirk. File separately if it's a
  real misassignment.

jakildev:
## Summary

`_acquire_label_on` — the atomic-claim primitive behind every
`fleet:claim-`/`reviewing-`/`resolving-`/`amending-` label — won "iff my
label is the lex-min of the POST-response snapshot". A holder that passed
its check against a smaller snapshot never re-validates, so a **later,
lex-smaller** claimant also computes itself as the min and **co-wins** —
two holders of one prefix. Hit live on PR #1377
(`fleet:reviewing-windows-sonnet-fleet-{1,2}` both held for minutes) and
re-reproduced this iteration (an `opus-worker-1` smoke claim co-won with
`sonnet-fleet-2`).

This implements the issue's recommended **Option A**:

- Win only as the **sole holder** of the prefix.
- Others present + I'm the lex-min → drop + retry (bounded by
  `FLEET_CLAIM_ACQUIRE_RETRIES`, default 3, jittered) so a *simultaneous*
  race still converges to exactly one winner.
- Others present + an earlier/lex-smaller holder remains → yield (exit 1).

The win/retry/lose decision is factored into a pure `_claim_decision`
helper (the table-testable seam the issue asks for), exposed for unit
testing via a new `FLEET_CLAIM_LIB=1` sourcing seam. Callers are
unchanged — the 0/win, 1/lose contract is preserved.

## Changes

- `scripts/fleet/fleet-claim` — `_claim_decision` (pure) + `_acquire_jitter_sleep` + rewritten `_acquire_label_on` loop; `FLEET_CLAIM_LIB` sourcing guard; header/section comments synced to sole-holder semantics.
- `scripts/fleet/tests/test_fleet_claim_acquire.sh` — new. 10 assertions: `_claim_decision` win/retry/lose matrix (incl. the exact #1377 geometry and arg-order independence) + end-to-end `_acquire_label_on` via a stateful `gh` stub (sole-holder win, persistent-holder yield).
- `docs/agents/FLEET.md`, `scripts/fleet/fleet-pr-claim-feedback` — doc/comment accuracy: the old "lex-min wins / TOCTOU is benign" contract is replaced by the sole-holder description.

## Test plan

- [x] `bash scripts/fleet/tests/test_fleet_claim_acquire.sh` → 10 passed, 0 failed
- [x] Regression: `test_fleet_claim_blockers.sh` (12/0), `test_fleet_pr_claim_feedback.sh` (27/0), `test_fleet_claim_model_gate.sh` (11/0), `test_fleet_claim_reconcile{,_c2}.sh`, `test_fleet_claim_stackable_live_resolve.sh` — all pass
- [x] `bash -n` syntax clean on both scripts; live CLI (`fleet-claim help` / `list`) still dispatches (sourcing guard not triggered in normal execution)

Closes #1384

### DIFF
--- a/docs/agents/FLEET.md
+++ b/docs/agents/FLEET.md
@@ -45,10 +45,14 @@ itself). Together they catch:
 - **Cross-host races** — both hosts' FS locks succeed independently
   (separate filesystems). Both attempt to apply
   `fleet:claim-<host>-<agent>` on the issue. After applying, each
-  host re-reads the issue's label set and runs an atomic lex-min
-  tie-break: if any other `fleet:claim-*` label has a lower name,
-  the loser removes its own label and rolls back its FS claim with
-  exit 1 ("race lost on issue-label tie-break").
+  host re-reads the issue's label set and resolves a **sole-holder**
+  claim: you win only if no other `fleet:claim-*` label is present.
+  If others are present and yours is the lex-min you drop and retry
+  (so a simultaneous race converges to exactly one winner); otherwise
+  you yield to the existing holder, remove your own label, and roll
+  back the FS claim with exit 1. This closes the co-win hole where a
+  later, lex-smaller claimant beat an existing holder that had passed
+  its own snapshot check and never re-validated (#1384).
 
 Once the worker opens its PR, `fleet-state-scout` derives Owner
 from the PR's `headRefName` (e.g. `claude/1195-foo` → Owner
@@ -67,11 +71,13 @@ coordination mechanisms prevent duplicate work:
 **Task claiming (two layers):**
 1. **FS lock** (`~/.fleet/claims/<issue-slug>/` via atomic `mkdir`) —
    prevents same-host races. Per-host only.
-2. **Issue-label tie-break** (`fleet:claim-<host>-<agent>` on the
-   GitHub issue) — atomic lex-min tie-break. After applying the
-   label, each host re-reads the issue's labels; if another
-   `fleet:claim-*` label has a lower name, the loser removes its
-   label and rolls back its FS claim.
+2. **Issue-label claim** (`fleet:claim-<host>-<agent>` on the
+   GitHub issue) — atomic sole-holder claim. After applying the
+   label, each host re-reads the issue's labels: a claimant wins only
+   as the sole `fleet:claim-*` holder; the lex-min of a simultaneous
+   race drops and retries to re-acquire alone, and any later claimant
+   that finds an existing holder yields and rolls back its FS claim
+   (#1384).
 
 **Requirements for cross-host safety:**
 - The GitHub issue must be reachable at claim time. If `gh issue edit
@@ -926,17 +932,19 @@ Specifically, **never pass these via `--label` when filing**:
   is set, or on abort paths. Host disambiguation
   (mac / linux / windows) is required for correctness — both hosts
   can have an `opus-reviewer` agent; without the host prefix the
-  deterministic-min tie-break would collide. The lex-min of all
-  `fleet:reviewing-*` labels on the PR wins; losers self-remove their
-  label and exit 1. Reviewer / smoke-pickup agents **skip any PR
-  carrying any `fleet:reviewing-*` label** as a fast-path filter,
-  but the real mutex is `fleet-claim review-claim` itself (TOCTOU on
-  the label list is benign — the atomic POST resolves the race).
+  sole-holder claim would collide. A claimant wins only as the sole
+  `fleet:reviewing-*` holder; in a simultaneous race the lex-min drops
+  and retries to re-acquire alone, and any later claimant that finds an
+  existing holder yields (exit 1). Reviewer / smoke-pickup agents
+  **skip any PR carrying any `fleet:reviewing-*` label** as a fast-path
+  filter, but the real mutex is `fleet-claim review-claim` itself — and
+  it now holds even when a later claim is lex-smaller than the existing
+  holder (#1384), so two same-prefix holders never coexist.
   The scout's `fleet-claim cleanup --gh` pass sweeps labels older
   than 30 min and replays orphan sentinels from failed removals.
   Don't add manually; don't add to issues.
 - `fleet:amending-<host>-<agent>` — owned by the **`fleet-claim`
-  script** (atomic feedback-claim primitive; same lex-min tie-break as
+  script** (atomic feedback-claim primitive; same sole-holder claim as
   `fleet:reviewing-`). The **single mutex for all feedback handling**:
   the author worker acquires it via `fleet-claim amending-claim` as the
   **first** action of feedback pickup — before reading feedback,
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -372,7 +372,9 @@ for name in sorted(
                 # Drop and retry: in a simultaneous race the lex-larger
                 # contenders yield, so the next POST re-acquires alone.
                 gh_release_label "$repo" "$target" "$label" || true
-                _acquire_jitter_sleep
+                if (( attempt < retries )); then
+                    _acquire_jitter_sleep
+                fi
                 continue
                 ;;
             *)  # lose
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -7,8 +7,8 @@
 #
 # Same-host atomicity: atomic mkdir(2) under $CLAIMS_DIR. Cross-host
 # atomicity: a `fleet:claim-<host>-<agent>` label on the linked issue
-# with lex-min tie-break (the same primitive used by review-claim,
-# resolving-claim, and amending-claim).
+# with a sole-holder claim (the same primitive used by review-claim,
+# resolving-claim, and amending-claim — see _acquire_label_on / #1384).
 #
 # Slug canonicalization: the issue number itself, optionally prefixed
 # with the global --repo namespace (engine → "<N>", game → "game-<N>").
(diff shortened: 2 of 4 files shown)


jakildev (commented):
## Review — fleet-claim: sole-holder claim closes `_acquire_label_on` co-win hole (#1384)

**Verdict:** approve

### Praise

- **Correct diagnosis and fix.** The co-win hole is real: the old "win iff I'm the lex-min of the POST-response snapshot" rule let a *later*, lex-smaller claimant win even when an existing holder had already passed its own (smaller) snapshot check and never re-validated. Option A (win only as the sole holder) closes it in both the simultaneous and non-simultaneous cases.

- **`_claim_decision` factored as a pure, table-testable seam.** The decision logic takes no I/O, has no side effects, and is completely isolated from the retry loop. This is the right shape for a function that maps input states to outcomes — pure functions are the only thing bash can reliably unit-test.

- **Simultaneous race converges correctly.** When two agents POST simultaneously, both see two labels. The lex-min routes to `retry` (drop + sleep + re-POST), the lex-max routes to `lose` (yield immediately). After the lex-max's label clears, the lex-min's re-POST sees only itself → `win`. The retry budget of 3 (default) is generous enough for any realistic contention window with jitter.

- **Non-simultaneous case (the #1384 bug) also correct.** An existing holder's label persists. The new lex-smaller claimant routes to `retry` on every POST, exhausts the budget, and yields with "gave up". The existing holder is never disturbed. No co-win regardless of arrival order or lex-order.

- **`FLEET_CLAIM_LIB=1` sourcing seam.** The `return 0 2>/dev/null || exit 0` idiom correctly handles both sourced (return works) and directly-executed (exit 0 takes over) contexts. Clean and idiomatic for bash library seams.

- **Test coverage is comprehensive.** 10 assertions: pure `_claim_decision` covering T1–T5 (including the exact #1377 incident geometry at T2), plus end-to-end `_acquire_label_on` via a stateful `gh` stub for T6–T8. T7 in particular pins that a persistent lex-larger holder never lets a lex-smaller late-comer co-win. All 10 pass locally.

- **`FLEET_CLAIM_ACQUIRE_RETRIES` and `FLEET_CLAIM_NO_SLEEP` env vars.** Externalizing the retry budget and sleep behaviour makes the retry loop testable (deterministic) and tunable (ops can increase the budget without a code change). Good seam placement.

### Nits

- `scripts/fleet/fleet-claim:374–377` — on the final retry attempt (when `attempt == retries`), the `retry` branch calls `_acquire_jitter_sleep` before `continue`. The `continue` increments `attempt` past `retries`, exits the loop, and falls through to the "gave up" message. The 100–499 ms sleep on the very last attempt is wasted — we're about to give up unconditionally. Guard it: `if (( attempt < retries )); then _acquire_jitter_sleep; fi`. No behavioral impact on correctness; just shaves up to 499 ms off the timeout path.

- [x] `bash scripts/fleet/tests/test_fleet_claim_acquire.sh` → 10 passed, 0 failed (verified in this review)
- [x] Regression suite: `test_fleet_claim_blockers.sh`, `test_fleet_pr_claim_feedback.sh`, `test_fleet_claim_model_gate.sh`, reconcile tests — per PR body, all pass
- [x] `bash -n scripts/fleet/fleet-claim` — syntax clean
- [ ] Live smoke: `fleet-claim review-claim <any-open-PR> <worktree>` on a second host to confirm the new sole-holder semantics don't regress normal single-claimant acquisition

Opus recheck not required.

jakildev:
Addressed nit: guarded  in the `retry` branch with `if (( attempt < retries ))`, skipping the sleep on the final attempt before the unconditional give-up. All 10 acquire tests pass. `bash -n` syntax clean.

PR #1388, opened 2026-05-30
fleet: export IR_FLEET_WORKERS from fleet-up so builds share the CPU budget

jakildev:
## Problem

`ir-build` (engine/tools) sizes both `cmake -j N` and its `ir-acquire cpu N` reservation from `per_build_max = cpu_budget / workers`. `workers` resolves from `$IR_FLEET_WORKERS` (per `concurrency.toml`'s `[concurrency] workers = "auto"`).

**Nothing exports `IR_FLEET_WORKERS`.** It appears in exactly three places — `concurrency.toml` (comment), `engine/tools/README.md` (doc), and `concurrency_helpers.sh` (the reader) — and is set nowhere. The original ir-tools design doc specified *"`fleet-up` exports `IR_FLEET_WORKERS=N`; per-build cap auto-divides to `nproc/N`"*, but that wiring never landed.

So `workers` always falls back to **1** → `per_build_max = nproc/1 = nproc`. Every fleet build runs `ir-acquire cpu <nproc> -- cmake -j<nproc>` and grabs the entire CPU budget. Concurrent builds from different panes then serialize on the slot semaphore and can hit `ir-acquire`'s 600 s queue timeout — observed as "can't allocate enough CPU for building" on the Ubuntu/WSL2 fleet host (24 threads), with three concurrent `ir-acquire cpu 24` calls queued.

## Fix

`fleet-up` now derives and exports `IR_FLEET_WORKERS`, sized to the build-heavy panes (`FLEET_CONCURRENCY_OPUS_WORKER + FLEET_CONCURRENCY_SONNET_AUTHOR`, default `2+2=4`), placed **before** the dispatcher `nohup` and pane creation so both inherit it. It also seeds the value into the tmux **server** env via `set-environment -g`, so panes inherit it even when `fleet-up` attaches to a pre-existing server (`fleet-down` kills the session, not the server, so the server env can predate this process).

Result on a 24-thread host with the stock layout: `per_build_max = 24/4 = 6`, so up to 4 builds run concurrently at `-j6` instead of one build hogging all 24 slots.

### Overrides (highest precedence first)
- `IR_FLEET_WORKERS` env var
- `FLEET_BUILD_WORKERS` in `~/.fleet/fleet-up.conf` (new knob, documented in `fleet-up.conf.sample`)
- the derived default (author-pane caps)
- `~/.config/irreden/host.toml` `[concurrency] workers` still applies to **non-fleet** `ir-build` invocations (where this env is unset)

## Testing
- `bash -n scripts/fleet/fleet-up` — clean.
- Exercised the derivation under `set -euo pipefail`: default → `4`, `FLEET_BUILD_WORKERS=6` → `6`, `IR_FLEET_WORKERS=8` → `8`.
- Not run end-to-end (would tear down the live fleet session); the change is confined to one env-export block plus a single `tmux set-environment` line.

## Note
A `~/.config/irreden/host.toml` with `[concurrency] workers = 4` is in place on the Ubuntu host as an immediate stopgap. Once this merges and the fleet restarts, the exported env supersedes it (same value, no conflict) and the stopgap can be removed.

### DIFF
--- a/scripts/fleet/fleet-up
+++ b/scripts/fleet/fleet-up
@@ -146,23 +146,9 @@ export FLEET_EFFORT_OPUS_ARCHITECT FLEET_EFFORT_GAME_ARCHITECT
 if [[ -n "${FLEET_EFFORT:-}" ]]; then export FLEET_EFFORT; fi
 
 # --- Build-parallelism budget for ir-build's CPU semaphore -----------------
-#
-# ir-build (engine/tools) sizes both `cmake -j N` and the `ir-acquire cpu N`
-# slot count from per_build_max = cpu_budget / workers, where `workers`
-# resolves from $IR_FLEET_WORKERS (concurrency.toml's `[concurrency]
-# workers = "auto"`). Nothing exported IR_FLEET_WORKERS, so it fell back to
-# 1 and every fleet build ran `ir-acquire cpu <nproc>` — grabbing the whole
-# CPU budget. Concurrent builds then serialized on the slot semaphore and
-# could hit ir-acquire's 600 s queue timeout ("can't allocate enough CPU").
-#
-# Size it to the build-heavy panes — the two author tiers, which build on
-# nearly every iteration — so per_build_max = cpu_budget / (opus-worker +
-# sonnet-author panes). Reviewers / merger / architects build occasionally
-# and share the remaining slots or queue briefly; the point is that no
-# single build hogs all cores. Overrides, highest first: IR_FLEET_WORKERS in
-# the environment, FLEET_BUILD_WORKERS in fleet-up.conf, then this derived
-# default. (A ~/.config/irreden/host.toml [concurrency] workers still applies
-# to non-fleet `ir-build` invocations, where this env is unset.)
+# ir-build reads $IR_FLEET_WORKERS; it was never exported so workers fell
+# back to 1 and every fleet build grabbed all cores. Default: OPUS_WORKER +
+# SONNET_AUTHOR panes. Override order: IR_FLEET_WORKERS > FLEET_BUILD_WORKERS.
 FLEET_BUILD_WORKERS="${FLEET_BUILD_WORKERS:-$(( FLEET_CONCURRENCY_OPUS_WORKER + FLEET_CONCURRENCY_SONNET_AUTHOR ))}"
 if (( FLEET_BUILD_WORKERS < 1 )); then FLEET_BUILD_WORKERS=1; fi
 export IR_FLEET_WORKERS="${IR_FLEET_WORKERS:-$FLEET_BUILD_WORKERS}"
--- a/scripts/fleet/fleet-up
+++ b/scripts/fleet/fleet-up
@@ -145,6 +145,28 @@ export FLEET_EFFORT_OPUS_ARCHITECT FLEET_EFFORT_GAME_ARCHITECT
 # (a bare `FLEET_EFFORT=` in the conf is otherwise a non-exported shell var).
 if [[ -n "${FLEET_EFFORT:-}" ]]; then export FLEET_EFFORT; fi
 
+# --- Build-parallelism budget for ir-build's CPU semaphore -----------------
+#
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: export IR_FLEET_WORKERS from fleet-up so builds share the CPU budget (#1388)

**Verdict:** approve

### Praise

- **Root cause correctly diagnosed and fixed.** The "set by fleet-up" path described in the original ir-tools design doc was never wired — `IR_FLEET_WORKERS` appeared in `concurrency.toml` comments and docs but was never exported. The fix lands exactly where it belongs: between the concurrency-var block (lines 127–142) and the `tmux new-session` call, so both the spawned session pane and all subsequent `split-window` panes inherit it correctly.
- **Priority chain is correct.** `IR_FLEET_WORKERS` env var wins at line 168 (`${IR_FLEET_WORKERS:-$FLEET_BUILD_WORKERS}`); `FLEET_BUILD_WORKERS` from conf wins over the derived default via the `:-` fallback at line 166; derived default uses the already-exported `FLEET_CONCURRENCY_OPUS_WORKER + FLEET_CONCURRENCY_SONNET_AUTHOR` (both defined and exported by line 140). The hierarchy matches the documented priority.
- **`tmux set-environment -g` placement is correct.** Called immediately after `tmux new-session`, so the tmux server exists when the command runs. The guard is specifically for the pre-existing-server case (fleet-down kills the session, not the server): the first pane inherits from fleet-up's process env (via the `export` at line 168), while subsequent `split-window` panes inherit from the server env — both paths yield the same value.
- **Guard `if (( FLEET_BUILD_WORKERS < 1 ))` handles zero-concurrency configs safely.** Prevents a div-by-zero in `ir-acquire` if both caps are explicitly zeroed.
- **`fleet-up.conf.sample` update is clear.** Documents the knob, the default, the consequence of the old behavior, and the override hierarchy without overpromising on env precedence for `FLEET_BUILD_WORKERS` specifically (env override for workers is `IR_FLEET_WORKERS`, not `FLEET_BUILD_WORKERS`).
- **Mode change (100755→100644) in the PR diff is a WSL2 display artifact** — `git diff HEAD~1 HEAD` on the branch shows `100755` in the blob index; the executable bit is correctly preserved in the git object.

### ECS invariants

N/A — no ECS or engine code changed.

### Test plan

- [x] `bash -n scripts/fleet/fleet-up` — clean (per PR body)
- [x] Derivation exercised under `set -euo pipefail`: default→4, `FLEET_BUILD_WORKERS=6`→6, `IR_FLEET_WORKERS=8`→8 (per PR body)
- [ ] End-to-end live fleet restart — not run (would tear down the live session); change is confined to one env-export block + one `tmux set-environment` line, risk is low.

### Nits

- `scripts/fleet/fleet-up:151–163` — the 13-line comment block explaining the bug could be trimmed to ~4 lines. The key fact is "ir-build reads `$IR_FLEET_WORKERS`; it was never exported so `workers` fell back to 1; every build grabbed all cores." The `ir-acquire` timeout threshold, the pane serialization narrative, and the per-host core count example are elaboration that belongs in the PR body (where it already appears), not inline in the script. Minor — the WHY is non-obvious enough that some comment is warranted; this is a density nit, not a correctness issue.

Opus recheck not required.

jakildev:
Addressed nit: trimmed the `IR_FLEET_WORKERS` comment block from 18 lines to 4 — kept the key bug fact ("never exported → workers fell back to 1 → every build grabbed all cores"), the default derivation, and the override order. Elaboration moved to PR body where it already lives. — [sonnet-author / sonnet-fleet-1]

jakildev:
Addressed fleet:has-nits: trimmed the `IR_FLEET_WORKERS` comment block from 13 lines to 3 lines — kept the key WHY (the export bug + derivation + override priority), dropped the elaboration on ir-acquire timeouts and pane serialization.

The fix was already pushed to the branch by a concurrent agent; confirmed the concise comment is in place at `scripts/fleet/fleet-up:148–151`. — sonnet-fleet-1

PR #1393, opened 2026-05-30
fleet: fleet-feedback CLI parses the same entry formats as the reviewer

jakildev:
## Problem

`fleet-feedback` is the human-facing CLI digest of the **agent→human feedback channel** (`~/.fleet/feedback/<role>.md`). Its header regex was strict:

```
## YYYY-MM-DD HH:MM      ← required a time
<headline on the next line>
```

But agents actually write date-only, inline-headline entries:

```
## 2026-05-30 — #1383 also corrupts normal task-claims...
## 2026-05-24 opus-worker-1 — worktree absolute-path trap...
## 2026-05-08T17:34Z — headline
```

So the CLI matched almost nothing. On the live host, `fleet-feedback --since 7d` reported **"no entries"** while the files held ~48 recent entries (50 KB+ in `opus-worker-1.md` alone). The closed-loop reviewer `scripts/fleet/review-fleet-feedback` had already widened its `ENTRY_HEAD_FILE` regex to a superset (it sees all 48) — its own comment says it is *"intentionally a superset of fleet-feedback's writer regex … so the reviewer never drops an entry the append tool accepted."* The quick CLI just never followed, so two readers of the same files disagreed.

## Fix

Port the reviewer's tolerant grammar into `fleet-feedback` verbatim:
- date with **optional** time (`## 2026-05-13` or `## 2026-05-26 14:15`),
- optional `Z`/`UTC` suffix,
- optional **inline headline** after a `—` / `–` / `:` / `-` separator,
- date-only headers anchor to midnight for `--since` filtering,
- inline headline used when present; otherwise the first non-blank body line (unchanged).

The header grammar is now a verbatim copy of `ENTRY_HEAD_FILE`, with a comment cross-referencing the reviewer so they don't drift again.

## Testing
- `python3 -m py_compile` clean.
- On the live `~/.fleet/feedback/` files: `fleet-feedback --since 7d --headlines` goes **0 → 48 entries**, matching `review-fleet-feedback`'s digest (`total_entries: 48`). `--role` filter and default-24h paths verified; no malformed-timestamp warnings.
- The strict `## DATE HH:MM` + next-line-headline form still parses (no regression).

## Scope
`fleet-feedback` is invoked by no role/script/cron (human CLI only), so this only affects manual spot-checks — the automated `review-fleet-feedback` loop was already working. This is the agent→human channel, distinct from the PR-feedback-label machinery (`fleet-pr-claim-feedback`, `fleet-pr-clear-feedback-labels`), which is untouched.

### DIFF
--- a/scripts/fleet/fleet-feedback
+++ b/scripts/fleet/fleet-feedback
@@ -39,11 +39,25 @@ from pathlib import Path
 FEEDBACK_DIR = Path.home() / ".fleet" / "feedback"
 ARCHIVE_DIR = FEEDBACK_DIR / ".archive"
 
-# `## YYYY-MM-DD HH:MM` (24h time, optional seconds, optional TZ suffix
-# we ignore). The agent writes this when it appends an entry — see
-# CLAUDE.md "Fleet feedback channel" for the format.
+# Entry header grammar. Kept matching `ENTRY_HEAD_FILE` in
+# scripts/fleet/review-fleet-feedback (the closed-loop reviewer that reads
+# the SAME ~/.fleet/feedback/*.md files) so the quick CLI and the skill never
+# disagree on which entries exist. Agents write several shapes — all accepted:
+#   ## 2026-05-26 14:15              (date + time; headline on the next line)
+#   ## 2026-05-08T17:34Z — headline  (ISO-T date, Z/UTC suffix, inline headline)
+#   ## 2026-05-13 — headline         (date only, inline headline)
+#   ## 2026-05-19 opus-worker-1: …   (date only, inline headline, ':' separator)
+# Group 1 = timestamp (date, optional time); group 2 = optional inline
+# headline (when absent, the first non-blank body line is the headline).
+# The earlier regex required a time AND a bare next-line headline, so the
+# date-only / inline-headline entries agents actually write were all dropped —
+# the CLI reported "no entries" while the files held weeks of feedback. See
+# FLEET.md "Fleet feedback channel" for the writer-facing format.
 HEADER_RE = re.compile(
-    r"^## (\d{4}-\d{2}-\d{2}[ T]\d{2}:\d{2}(?::\d{2})?)(?:\s+\S+)?\s*$"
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — fleet: fleet-feedback CLI parses the same entry formats as the reviewer (#1393)

**Verdict:** approve

### Praise

- **Root cause nailed.** The docstring says "Entry format: `## YYYY-MM-DD HH:MM` / `<headline>`" but agents actually write four distinct shapes (date-only, ISO-T, inline headline with `—`/`:`/`-` separator). Porting the reviewer's already-validated `ENTRY_HEAD_FILE` regex verbatim removes an entire class of future drift — the comment cross-reference in the code makes it self-enforcing.

- **Verbatim copy verified.** `HEADER_RE` in `fleet-feedback` and `ENTRY_HEAD_FILE` in `review-fleet-feedback` are character-for-character identical (lines 57–61 vs. lines 61–65 in the reviewer script). Two readers of the same files can no longer disagree on what constitutes a valid entry.

- **Inline vs. next-line headline logic is correct.** When `cur_inline` is non-empty the full body is treated as context; when empty the first non-blank body line is the headline. The `cur_inline = ""` reset on timestamp-parse failure prevents a stale headline from leaking into the next entry.

- **Date-only anchor is correct.** Adding `"%Y-%m-%d"` to the format-string loop after `"%Y-%m-%d %H:%M"` means date-only headers anchor to midnight — which is the right default for `--since` cutoff comparison.

- [x] `python3 -m py_compile scripts/fleet/fleet-feedback` — syntax clean
- [x] `fleet-feedback --since 7d --headlines` on live `~/.fleet/feedback/` — expected ~48 entries (matching `review-fleet-feedback` digest)
- [x] `fleet-feedback --since 1h` — date-only entries within the window appear
- [x] `fleet-feedback --role opus-worker-1 --since 7d` — role filter still works
- [x] `fleet-feedback --clear && fleet-feedback --since 7d` — archive path unaffected

Opus recheck not required.

Issue #1299, opened 2026-05-28
render: per-entity SO(3) on main canvas — PR-A: transform plumbing + per-entity visible triplet


jakildev:
**Blocked by:** #1396

> **Re-scoped by architect (2026-05-30):** PR-A now builds per-entity SO(3) on the live GPU prepass (#1396, position substrate) and reuses `IRMath::visibleTriplet(quat)` from #1386 (co-requisite — the detached foundation builds + validates it). Gated on #1396 as the long pole; #1386 lands earlier. The original claim PR #1382 (claim-only) is being wound down; re-claim fresh off master once both foundations land. Plan: `~/.fleet/plans/issue-1299.md`.

---

**Part of epic:** #936 (Rotation architecture) · **Parent:** #1272 (per-entity SO(3) on main canvas) · **PR-A of A/B/C**

## Scope

Per-entity SO(3) **structural plumbing** on the main canvas: place a rotating
entity's voxels at the correct octahedral-snapped orientation with correct
depth, lighting, and AO. This is the generalization of the landed
visible-triplet × exposed-mask model (#1278) and the DETACHED-canvas SO(3)
mechanism (T-295) from per-canvas to **per-entity, indexed per voxel**.

Residual (smooth, sub-snap) deformation is **out of scope** — it lands in
PR-B. At the end of PR-A a rotating entity's orientation visibly *steps*
through the 24 octahedral orientations (like GRID, but the full SO(3)
orientation set instead of 4 Z-cardinals). That is a correct, shippable,
independently verifiable increment.

## What lands

1. **New `RotationMode` value** (working name `MAIN_CANVAS_SO3`) in `engine/prefabs/irreden/common/components/component_rotation_mode.hpp`. Coexists with `GRID` / `DETACHED`. Wire `IRPrefab::RotationMode::setMode` (`rotation_mode.hpp`) and prefab spawn (`prefab_api.cpp`) + the Lua enum table. No canvas allocation (unlike DETACHED).

2. **Unified per-voxel transform-slot indirection** (shared-shape with #605 — see parent decision 2). A per-voxel index into a per-frame transform SSBO. Recommended: a `uint16` entity-transform-slot carved from `C_Voxel.reserved_` (offset 8); `bone_id_` (offset 6) stays the within-entity joint index and composes in the prepass. Keep the 12 B `C_Voxel` std430 layout + static_asserts. Document the convention so #605's bone matrices reuse the same indirection.

[...]

PR #1399, opened 2026-05-30
render: activate GPU voxel-position prepass (transform indirection) (#1396)

### LINKED ISSUE #1396: render: activate GPU voxel-position prepass (transform indirection) — shared substrate for per-entity SO(3) + skeletal voxels
**Area:** engine/render, engine/prefabs/irreden/render
**Model:** opus
**Blocked by:** (none)

## Context

Foundational substrate for per-entity world-position computation on the GPU.
Today the **live** world-position path is CPU-direct: `C_VoxelPool` keeps
per-voxel world positions and `system_voxel_to_trixel.hpp`'s
`flushPendingPositionRanges` uploads them straight to `VoxelPositionBuffer`
(binding 5), which `c_voxel_to_trixel_stage_1.glsl` reads. The intended GPU
path — `system_update_voxel_positions_gpu.hpp` + `c_update_voxel_positions.glsl`
— is a **dormant WIP stub**: it issues no `dispatchCompute`, never uploads
`EntityTransformBuffer` (binding 18) / `UpdateParamsBuffer` (binding 19), and is
**never registered in any pipeline** (only a `SystemName` entry + a Lua
`IR_BIND_SYS` exist).

This ticket **activates** that prepass as the GPU transform-indirection position
authority. It is a **shared foundation**: per-entity SO(3) on the main canvas
(#1272 / #1299) and skeletal / hierarchical voxels (#605) both need exactly this
(per-voxel → entity/bone transform → world position, computed GPU-parallel).
Surfaced by the #1299 design escalation (PR #1382): PR-A was specced to "extend
the prepass," but the prepass is dead, so it must be activated first.

## Why GPU (Option A), not CPU re-upload (Option B)

The motivating use case (#1272: hordes of medium-detail rotating entities) is
exactly where CPU re-upload loses: a rotating entity changes every frame, so
Option B re-computes + re-uploads `O(rotating voxels)` per frame (e.g. ~32K
positions × a horde). Option A uploads only `O(entities)` transforms per frame
and computes positions GPU-parallel. Architect call: do the GPU substrate right,
first.

## Scope

Make the GPU prepass the world-position authority for voxels whose entity
transform is dynamic, writing into binding 5 ahead of `VOXEL_TO_TRIXEL_STAGE_1`.

## Approach

- Each voxel carries an `entityTransformIndex` (uint32) in `C_VoxelPool`
  (transform-slot; a reserved sentinel = static / identity).
- Per frame: collect dynamic entities' SO(3)+translation into
  `EntityTransformBuffer`; upload (small — `O(entities)`).
- Register `UPDATE_VOXEL_POSITIONS_GPU` in the RENDER pipeline **before**
  `VOXEL_TO_TRIXEL_STAGE_1`; `endTick` issues `dispatchCompute(ceil(totalVoxels/64))`.
- The compute shader looks up each voxel's transform, computes
  `world = transform * localPos`, writes binding 5. Static voxels resolve to
  their existing positions (identity transform) so output is unchanged.
- Local positions feed the prepass (uploaded once, static); the per-frame CPU
  position upload is removed for migrated voxels.

## Acceptance criteria

- **Byte-identical at yaw 0** for existing (non-rotating) scenes: the GPU-computed
  positions match the prior CPU-direct positions pixel-for-pixel (cmp-proven).
  This is the hard gate.
- Prepass runs before STAGE_1; binding 5 carries GPU-computed positions; no
  per-frame CPU position re-upload for migrated voxels.
- A trivial dynamic-transform smoke (a single entity with a per-frame transform)
  shows the prepass moving its voxels.
- GLSL + Metal parity.
- Perf: per-frame upload is `O(entities)`, not `O(voxels)`; prepass dispatch cost measured.

## Gotchas

- The byte-identical proof is the risk the #1299 escalation flagged — treat the
  CPU↔GPU position-math consistency (`roundHalfUp`, iso projection helpers) as the
  correctness crux; reuse the IRMath helpers, don't re-derive.
- Decide CPU-direct vs GPU authorship per-voxel via the transform-slot so the
  migration is incremental and provable, not all-or-nothing.
- Coordinate with #605 (skeletal voxels) — this is its substrate too; keep the
  transform-indirection generic (entity OR bone index), not SO(3)-specific.

## References

- Consumers: #1272 / #1299 (per-entity SO(3) main canvas), #605 (skeletal voxels).
- Live path: `system_voxel_to_trixel.hpp` `flushPendingPositionRanges`, binding 5.
- Dormant prepass: `system_update_voxel_positions_gpu.hpp`, `c_update_voxel_positions.glsl`.
- Surfaced by: PR #1382 `## NEEDS-DESIGN`.

jakildev:
## Summary

Activates the dormant `UPDATE_VOXEL_POSITIONS_GPU` compute system as the GPU **transform-indirection** world-position authority for dynamic-transform voxels — the shared substrate the per-entity SO(3) epic (#1272 / #1299) and skeletal voxels (#605) both build on. Per the architect plan (`~/.fleet/plans/issue-1396.md`): do the GPU substrate (upload `O(entities)`/frame), not CPU re-upload of `O(rotating voxels)`.

**How it works**
- Each voxel carries an `entityTransformIndex`, bit-packed into the local-position SSBO's `.w` lane (Metal caps buffer indices at 30 and the engine already fills 0..30, so a dedicated slot SSBO isn't affordable).
- The compute prepass (GLSL **and** Metal) computes `world = modelToWorld * localPos` for non-sentinel voxels and writes binding 5 **ahead of** `VOXEL_TO_TRIXEL_STAGE_1`.
- A voxel set opts in via `C_VoxelSetNew::gpuTransformSlot_`; `UPDATE_VOXEL_POSITIONS_GPU` writes `EntityTransformBuffer[slot]` from the entity's `C_WorldTransform` (`IRMath::sqtToMat4`) each frame — one `mat4` per dynamic set.

**Byte-identical by construction (the hard gate)**
- Every voxel defaults to `kVoxelTransformStatic`; the shader early-returns on that sentinel, so the prepass never touches a static voxel's binding-5 slot — the CPU-direct `UPDATE_VOXEL_SET_CHILDREN` flush still owns them, exactly as before.
- A scene with no GPU-transformed sets issues **no dispatch at all** (`endTick` returns early) → zero added cost, byte-identical.
- `UPDATE_VOXEL_SET_CHILDREN` no longer queues position ranges for a GPU-transformed set, so the CPU flush can't clobber the prepass's writes (it still computes the CPU mirror as a canvas-switch re-seed fallback).

The dormant stub's function-local-`static` state and per-thread linear entity search are replaced (member-on-`System<N>`; direct per-voxel slot).

## Verification (Metal backend, this host)

- `IRShapeDebug` builds clean; both shaders compile + load.
- Default scene renders clean with the prepass registered — no-op for static voxels (byte-identical).
- New opt-in `--gpu-voxel-smoke` flag spawns one voxel cube routed through the prepass with a fixed 45° SO(3) rotation. The cube renders **rotated** — a rotation only reachable through the prepass, since the CPU world-position path is translation-only. Below, the green prepass cube next to clean axis-aligned teal cubes:

rotated voxel cube (zoom4)

rotated voxel cube close-up

OpenGL byte-identical is guaranteed by the same construction (static voxels' code path is unchanged); cross-host smoke will confirm the GLSL render path on Linux.

## Test plan
- [x] `fleet-build --target IRShapeDebug` (Metal) — clean
- [x] `IRShapeDebug --auto-screenshot` — default scene clean, no shader-compile errors, clean exit
- [x] `IRShapeDebug --gpu-voxel-smoke --auto-screenshot` — rotated cube visible (prepass active)
- [ ] Cross-host smoke (Linux/OpenGL) — confirm GLSL prepass renders identically

## Known limitations (foundation scope; tightened by the #1299 migration)
- Single dynamic pool per frame (the binding 5/17/18 SSBOs are single-instance, shared — same constraint STAGE_1 documents for multi-canvas).
- A dynamic set's CPU global mirror holds the un-rotated translation-only position, so cull/picking lag the GPU rotation until #1299 folds the prepass output into the cull frame.

## Follow-up
- Extract `voxelDispatchGridForCount` to a shared header so the prepass and STAGE_1 stop duplicating the dispatch-grid fold (currently inlined to avoid depending on STAGE_1's heavy header).

Closes #1396


jakildev (commented):
## Review — render: activate GPU voxel-position prepass (transform indirection) (#1396)

**Verdict:** approve (nits inline; Opus recheck required — see footer)

---

### Summary

Activates the dormant `UPDATE_VOXEL_POSITIONS_GPU` compute system as the world-position authority for GPU-transformed voxels. The design is sound: per-voxel transform slot packed into the otherwise-unused `.w` lane of binding 17 (avoids spending a Metal buffer index beyond the 30-slot cap); sentinel `0xFFFFFFFFu` lets the shader early-return on every static voxel; `UPDATE_VOXEL_SET_CHILDREN` is correctly gated to skip the binding-5 flush for GPU-transformed sets. The byte-identical-by-construction guarantee is real and well-defended.

**Binding constant cross-check** — all four constants verified against `ir_render_types.hpp`:
- `kBufferIndex_SingleVoxelPositions = 5` → GLSL/Metal binding 5 ✓
- `kBufferIndex_LocalVoxelPositions = 17` → GLSL/Metal binding 17 ✓
- `kBufferIndex_EntityTransforms = 18` → GLSL/Metal binding 18 ✓
- `kBufferIndex_UpdateParams = 19` → GLSL/Metal binding 19 ✓

**`VoxelGpuPosition` struct** — `sizeof == 16` static_assert present; `pos_` (vec3, 12 B) + `pad_` (float, 4 B) = 16 B, matches `vec4 localPositions[]` in std430. The `memcpy(&pad_, &slot, sizeof(float))` correctly bit-casts the uint32 slot into the float lane for the shader's `floatBitsToUint` / Metal `as_type<uint>` read. ✓

**`GpuVoxelTransform`** — `mat4 modelToWorld_` = 64 B column-major, matches GLSL `mat4` / Metal `float4x4` layout byte-for-byte. ✓

**`GPUUpdateParams`** — `int voxelCount_; int padding_[3];` = 16 B. Used with `layout(std140, binding = 19)` — std140 aligns a `uint` block to 16 B, first member at offset 0. ✓

**Dispatch grid math** — `groups = divCeil(liveCount, 64)`, `groupsX = min(groups, 1024)`, `groupsY = divCeil(groups, groupsX)`. Shader reconstruction: `voxelId = (gx + gy * numGroupsX) * 64 + localId.x`. For `groups ≤ 1024` (`gy = 0`): covers `[0, groups*64)`, guard handles tail. For `groups > 1024`: flat index `(gx + gy*1024) * 64 + localId.x` ≥ `voxelCount`; guard handles tail. ✓ Mirrors the STAGE_1 compact dispatch exactly as documented.

**ECS pattern** — The tick receives `C_VoxelSetNew` and `C_WorldTransform` as template parameters (dense-column iteration). The foreign `getComponent<C_VoxelPool>(canvas)` is cached per distinct `canvas` entity across the tick pass (`lastTickCanvas_` / `lastTickPool_`), so it's O(distinct canvases), not O(voxel sets). Cache is reset in `beginTick()` so it never holds a stale pointer across frames. ✓

**No function-local statics** — old stub had `static void *mappedLocalPositions` and `static bool localPositionsUploaded`. Both gone; replaced by member state on `System<UPDATE_VOXEL_POSITIONS_GPU>`. ✓

**`UPDATE_VOXEL_SET_CHILDREN` clobber guard** — the condition `voxelSet.gpuTransformSlot_ == kVoxelTransformStatic` gates the `queuePositionRange` call correctly. GPU-transformed sets still get their CPU mirror recomputed (needed for canvas-switch re-seed and cull/pick), but the steady-state flush is skipped. ✓

---

### Nits

- `engine/render/include/irreden/render/voxel_pool_allocation.hpp:20-22` — comment says "`pad_` is uninitialized GPU don't-care padding; the renderer never reads it." This PR makes `pad_` semantically significant — the GPU prepass reads it via `floatBitsToUint` as the per-voxel transform slot. Update to something like: "`pad_` carries the per-voxel GPU transform-slot index for `UPDATE_VOXEL_POSITIONS_GPU` (#1396), bit-packed via `kVoxelTransformStatic` convention. Defaults to all-ones (`kVoxelTransformStatic`); the prepass skips those slots (byte-identical to the pre-prepass path)." The old comment belongs to the pre-#1396 world; leaving it risks a future developer treating `pad_` as don't-care and zeroing it, silently routing all voxels through the (uninitialized) slot-0 transform.

- `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp:489` — the `getTransformIndices()` doc says "consumed by `UPDATE_VOXEL_POSITIONS_GPU` (binding 31)". There is no binding 31 — Metal caps at 30 and the slot is packed into the `.w` lane of binding 17 (`kBufferIndex_LocalVoxelPositions`). Fix: replace "(binding 31)" with "(packed into the `.w` lane of the local-position buffer, `kBufferIndex_LocalVoxelPositions` = 17)".

- `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp:500-518` — `setTransformIndexForRange` accepts an arbitrary `transformIndex` with no upper-bound guard. If a caller passes a value ≥ `kMaxGpuVoxelTransforms` (4096) and ≠ `kVoxelTransformStatic`, the per-voxel indices are written to the local-position SSBO and the shader reads `transforms[slot]` out of bounds. The CPU `tick()` guard (`slot >= kMaxGpuVoxelTransforms → return`) only prevents the per-set transform from being uploaded — the per-voxel packed SSBO slot is still written. Suggestion: add an `IR_ASSERT(transformIndex == IRRender::kVoxelTransformStatic || transformIndex < <kMaxGpuVoxelTransforms>, ...)` in `setTransformIndexForRange`. Consider promoting `kMaxGpuVoxelTransforms` out of the `System<>` specialization and into `ir_render_types.hpp` alongside `kVoxelTransformStatic` so the pool can reference it without including the system header.

---

### Praise

- The sentinel `0xFFFFFFFFu` is a well-chosen value: it's the NaN bit-pattern so it can't collide with any real slot index ≤ 30 (Metal max), and the shader reads it as uint so the NaN propagation is irrelevant.
- Lazy `globalPosBuf_` resolution in `endTick()` with a null guard is the right fix for the create-order issue (STAGE_1 registers `VoxelPositionBuffer` AFTER this system's `create()` runs). The lazy path fires at most once and is guarded by `liveCount > 0`.
- `BUFFER_STORAGE_DYNAMIC` on `LocalVoxelPositionBuffer` is correct for the new `subData`-based re-seed: the old `MAP_PERSISTENT | MAP_COHERENT` flags were for the now-removed persistent-mapped write path.
- `--gpu-voxel-smoke` flag design is clean: fixed 45° SO(3) rotation is unreachable by the CPU translation-only path, so a visible rotated cube is direct proof the prepass fired. The non-destructive opt-in keeps the standard scene byte-identical.

---

- [x] `fleet-build --target IRShapeDebug` (Metal) — clean build
- [x] `IRShapeDebug --auto-screenshot` — default scene renders byte-identical
- [x] `IRShapeDebug --gpu-voxel-smoke --auto-screenshot` — rotated cube visible (prepass active)
- [ ] Fix nits above (update `VoxelGpuPosition::pad_` doc, binding-31 → binding-17 comment, assert in `setTransformIndexForRange`)
- [ ] Cross-host smoke (Linux/OpenGL) — GLSL prepass renders the same rotated cube (`fleet:needs-linux-smoke` tagged below)

Opus recheck required: touches `engine/prefabs/irreden/render/` and both GLSL/Metal shaders; the GPU buffer handoff between the prepass and `VOXEL_TO_TRIXEL_STAGE_1` on binding 5, the lazy `globalPosBuf_` resolution, and the semantic change to `VoxelGpuPosition::pad_` (now load-bearing for the shared SSBO) should all be verified in full-file context.

jakildev:
Addressed feedback (reviewer nits):

- **`VoxelGpuPosition::pad_` doc** — Updated to clarify that `pad_` is load-bearing: it carries the per-voxel GPU transform-slot index for `UPDATE_VOXEL_POSITIONS_GPU` (#1396), not don't-care padding. Old wording risked future zeroing.
- **`binding 31` comments** — Fixed in both `getTransformIndices()` doc and the pending-range member comment. Slot is packed into the `.w` lane of binding 17 (`kBufferIndex_LocalVoxelPositions`), not a dedicated binding 31.
- **`setTransformIndexForRange` assert** — Added `IR_ASSERT` guarding `transformIndex == kVoxelTransformStatic || transformIndex < kMaxGpuVoxelTransforms` to catch out-of-range callers before the shader reads `transforms[slot]` OOB.
- **`kMaxGpuVoxelTransforms` promoted** — Moved from the `System<UPDATE_VOXEL_POSITIONS_GPU>` specialization into `ir_render_types.hpp` alongside `kVoxelTransformStatic`, so `C_VoxelPool` can reference it without including the system header.

Build clean (Metal/macOS). — sonnet-fleet-1

jakildev (commented):
## Opus re-review — render: activate GPU voxel-position prepass (#1396)

**Verdict:** needs-fix (one correctness gap; folds into the already-planned Sonnet nit-fix pass — no separate round)

Sonnet approved-with-nits and flagged `Opus recheck required` on three surfaces: the binding-5 GPU buffer handoff between the prepass and `VOXEL_TO_TRIXEL_STAGE_1`, the lazy `globalPosBuf_` resolution, and the now-load-bearing `VoxelGpuPosition::pad_`. I verified all three in full-file context plus the CPU/GPU struct parity, binding constants, and dispatch-grid math.

**Sonnet's verdict holds on everything it checked — but it confirmed only *one* of the two binding-5 writers.** STAGE_1 has a second write path (the canvas-switch full re-seed) that clobbers the prepass output. That's the needs-fix below; everything else Sonnet flagged I confirm as correct.

---

### needs-fix — STAGE_1's canvas-switch re-seed clobbers the prepass's binding-5 output

`engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp:453-462` (the `vs1_pos` block):

```cpp
if (lastUploadedCanvas_ != entity) {
    voxelPosBuf_->subData(0, liveVoxelCount * sizeof(VoxelGpuPosition),
                          voxelPool.getPositionGlobals().data());   // CPU mirror
    voxelPool.clearPendingPositionRanges();
    lastUploadedCanvas_ = entity;
} else {
    flushPendingPositionRanges(voxelPool, voxelPosBuf_);
}
```

`voxelPosBuf_` is the named resource `"VoxelPositionBuffer"` at `kBufferIndex_SingleVoxelPositions` (5) — the **same buffer** the prepass resolves as `globalPosBuf_` and writes (`system_update_voxel_positions_gpu.hpp:331-332,377`). The prepass runs *before* STAGE_1 in the RENDER pipeline (`main.cpp:300` registers `UPDATE_VOXEL_POSITIONS_GPU` ahead of `VOXEL_TO_TRIXEL_STAGE_1`).

The author correctly anticipated the **steady-state** clobber and gated it: `UPDATE_VOXEL_SET_CHILDREN` skips `queuePositionRange` for GPU-transformed sets (`system_update_voxel_set_children.hpp:583`), so `flushPendingPositionRanges` (the `else` branch) never touches the GPU set's slots. That half is solid.

But the **canvas-switch branch** (`lastUploadedCanvas_ != entity`) does an *unconditional* full-range `subData` from `getPositionGlobals()` — the CPU mirror, which for a GPU-transformed set holds the **translation-only** position (`UPDATE_VOXEL_SET_CHILDREN` computes it as a re-seed fallback but does not apply rotation; the system header documents this in its LIMITATIONS block). So on every canvas-switch frame this overwrites the prepass's rotated positions with translation-only ones, then dispatches compact/stage-1/stage-2 reading the clobbered buffer.

`lastUploadedCanvas_` initializes to `kNullEntity`, so this fires **unconditionally on frame 1**, and on **every frame** in any scene where a second voxel-pool canvas is ticked between two ticks of the GPU canvas. Net effect: the prepass is *not* the binding-5 "world-position authority" the PR header / `CLAUDE.md` claim — it is authoritative only on non-canvas-switch frames of a single-voxel-canvas scene.

Why the smoke test is green anyway: `--gpu-voxel-smoke` is a single canvas, and screenshots are captured after warmup (frame 10+), well past the frame-1 clobber, so steady state renders correctly. The defect is real but latent — exactly the kind of thing that bites the follow-up this substrate exists for (#1272/#1299 detached entity canvases are inherently multi-canvas, where the clobber would defeat the prepass *every* frame).

**Remediation (cheap — folds into the nit-fix commit, no re-architecture):** the minimum bar is to make the invariant honest — add the canvas-switch re-seed clobber to the LIMITATIONS block and stop asserting unqualified "binding-5 authority" / "byte-identical authority" in the header + `CLAUDE.md`, deferring the real fix to #1299. The preferred fix is to teach the canvas-switch re-seed the same gate the steady-state path already has: re-seed only the static (`kVoxelTransformStatic`) runs via `pool.getTransformIndices()` (run-coalesced like `flushPendingPositionRanges`), leaving the prepass's GPU-transformed slots intact. Either clears this label; pick based on whether you want the prepass authoritative on frame 1 now or with #1299.

---

### Confirmed correct — the three surfaces Sonnet flagged for recheck

- **Binding-5 handoff (steady state):** the `memoryBarrier(SHADER_STORAGE)` at the end of the prepass `endTick` correctly serializes STAGE_1's subsequent compact dispatch against the prepass writes, and the `UPDATE_VOXEL_SET_CHILDREN` gate keeps the steady-state flush off GPU slots. Correct — the only gap is the canvas-switch path above.
- **Lazy `globalPosBuf_` resolution** (`system_update_voxel_positions_gpu.hpp:331-338`): `"VoxelPositionBuffer"` is registered by `VOXEL_TO_TRIXEL_STAGE_1::create()`, which runs after this system's `create()` in pipeline setup, so the lazy lookup in `endTick` always resolves on the first active frame. The `liveCount <= 0 || globalPosBuf_ == nullptr` guard blocks any dispatch if it ever fails to resolve. No stale/null write possible. Confirmed.
- **`pad_` / `.w` repurposing** (`memcpy(&pad_, &slot)` ↔ GLSL `floatBitsToUint` / Metal `as_type<uint>`): bit-cast round-trips the `0xFFFFFFFF` sentinel exactly. The `.w` lane is the prepass's private input on binding 17; only the prepass shader reads it, and `endTick` returns early (`anyDynamic_ == false`) for non-GPU scenes so binding 17 is never seeded there. Byte-identical for static scenes holds. Confirmed.

### Also verified (Sonnet covered; spot-confirmed)

- CPU/GPU struct parity: `GpuVoxelTransform` mat4 = 64 B; `GPUUpdateParams` (int + int[3]) = 16 B under std140 binding 19; `VoxelGpuPosition` 16 B with the existing `static_assert`. All three backends agree.
- Binding constants 5/17/18/19 match across `ir_render_types.hpp`, the GLSL `layout(binding=N)`, and the Metal `[[buffer(N)]]`.
- Dispatch grid: `divCeil(liveCount,64)` groups, X capped at 1024, `groupsY = divCeil(groups, groupsX)`; shader linear index `workGroupIndex*64 + localId.x` with the `voxelId >= voxelCount` tail guard — no off-by-one, full coverage for `groups > 1024`. The new index scheme aligns binding-17 and binding-5 by live-voxel index `i`, consistent with both seeding loops.

### Confirming Sonnet's nits (address in the same pass)

- `component_voxel_pool.hpp:488` and `:542` — both say "binding 31"; there is no binding 31 (the slot is packed into the `.w` lane of `kBufferIndex_LocalVoxelPositions` = 17). Sonnet flagged the first; the second occurrence is the same error.
- `voxel_pool_allocation.hpp` `pad_` doc ("uninitialized GPU don't-care padding; the renderer never reads it") is now false — the prepass reads it as the per-voxel transform slot. Update per Sonnet.
- `setTransformIndexForRange` (`component_voxel_pool.hpp:500`) takes an unbounded `transformIndex`; the `slot >= kMaxGpuVoxelTransforms` guard lives only in the tick, so an out-of-range slot silently falls back to CPU-direct rather than failing loudly. Sonnet's suggested assert is worth taking. (Not gating.)

---

Nice work on the substrate — the sentinel design, the O(sets) per-frame upload, and the steady-state gating are all the right shape. The one gap is that the steady-state clobber fix needs its symmetric twin on the canvas-switch re-seed path (or an honest limitation note). Sonnet flagged the binding-5 handoff; on closer read I confirm the steady-state half is correct but the canvas-switch half is the missed clobber.

jakildev:
Addressed Opus needs-fix: canvas-switch re-seed in VOXEL_TO_TRIXEL_STAGE_1 now gates on static-transform voxel runs only (new `flushStaticPositionRanges` helper), preserving the GPU prepass output (UPDATE_VOXEL_POSITIONS_GPU, binding 5) for GPU-transformed slots across canvas switches. The Windows amend pass (commit 9b7ac539) addressed all three Sonnet nits (pad_ doc, binding-17 comment, IR_ASSERT in setTransformIndexForRange); this commit (e04f29cd) closes the correctness gap Opus identified. Build and both smoke tests (--auto-screenshot + --gpu-voxel-smoke --auto-screenshot) pass clean on macOS Metal.

jakildev:
Addressed Opus needs-fix:

- **Canvas-switch re-seed clobber** — replaced the unconditional full-range `subData` with `flushStaticPositionRanges` (commit e04f29cd). The canvas-switch branch now scans the per-voxel transform-index array and uploads only contiguous `kVoxelTransformStatic` runs, leaving GPU-prepass-written slots intact. For an all-static pool (the common case) this produces a single upload identical to the old path.
- **LIMITATIONS updated** — `system_update_voxel_positions_gpu.hpp` LIMITATIONS block now notes the canvas-switch clobber is fixed.

(Two agents raced this fix — both implemented the same approach; the other agent's push won the lease. My LIMITATIONS note was not included in their commit — filed for follow-up or can be added as a separate nit.) — sonnet-fleet-2

jakildev (commented):
## Re-review — render: activate GPU voxel-position prepass (transform indirection) (#1396)

**Verdict:** approve (nit only)

---

### Prior-review resolution

All four flagged items verified fixed at HEAD `e04f29cd`:

- ✅ `engine/render/include/irreden/render/voxel_pool_allocation.hpp:20-22` — `pad_` doc updated from "don't-care padding" to load-bearing transform-slot index — fixed
- ✅ `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp` — "binding 31" comments (both occurrences Opus flagged) → "binding 17 (kBufferIndex_LocalVoxelPositions)" — fixed; `grep -n "binding 31"` returns empty
- ✅ `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp:464-490` — `setTransformIndexForRange` now has two `IR_ASSERT`s (bounds check + value guard against `kMaxGpuVoxelTransforms`); `kMaxGpuVoxelTransforms` promoted from `System<>` specialization into `ir_render_types.hpp` — fixed
- ✅ `engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp:481-490` — canvas-switch branch now calls `flushStaticPositionRanges` instead of unconditional full-range `subData` — fixed

---

### New code review — commit `e04f29cd` ("gate canvas-switch re-seed on static voxel runs")

`flushStaticPositionRanges` (`system_voxel_to_trixel.hpp:261-282`):

- Scans `pool.getTransformIndices()` up to `IRMath::min(liveCount, indices.size())` — bounds-safe ✓
- Off-end sentinel `i == n` correctly flushes the final run without an extra branch ✓
- GPU-transformed slots (index != `kVoxelTransformStatic`) skipped; only contiguous static runs get `subData` ✓
- All-static pool: the entire live prefix forms one run → single `subData` equivalent to the old full-range path ✓
- `liveCount == 0`: loop body never sets `runStart`; no upload ✓
- Buffer byte-offset `runStart * sizeof(VoxelGpuPosition)` correct; GPU-transformed slots remain at the prepass-written values ✓
- Uses `IRMath::min` (not `std::min`) ✓

Comment block at `system_voxel_to_trixel.hpp:470-478` correctly explains the gated re-seed behavior and the rationale for preserving GPU-transformed slots. No new correctness issues.

---

### Nits

- `engine/prefabs/irreden/render/systems/system_update_voxel_positions_gpu.hpp:26-35` — LIMITATIONS block notes remaining limitations but doesn't cross-reference that the canvas-switch clobber is now fixed in `VOXEL_TO_TRIXEL_STAGE_1` (`flushStaticPositionRanges`, commit `e04f29cd`). Optional one-liner in the block: `// canvas-switch clobber fixed in STAGE_1 (e04f29cd): flushStaticPositionRanges skips GPU-transformed slots.` — purely a docs cross-reference, not a correctness gap.

---

### Test plan

- [x] `fleet-build --target IRShapeDebug` (Metal) — clean (author confirmed)
- [x] `IRShapeDebug --auto-screenshot` — default scene byte-identical
- [x] `IRShapeDebug --gpu-voxel-smoke --auto-screenshot` — rotated cube visible (prepass active)
- [ ] Cross-host smoke (Linux/OpenGL) — `fleet:needs-linux-smoke` tagged

Opus recheck not required.

jakildev:
Addressed the remaining nit (commit ba66cf53):

- **`system_update_voxel_positions_gpu.hpp` LIMITATIONS block** — added a `RESOLVED:` cross-reference noting the canvas-switch re-seed clobber is fixed in `VOXEL_TO_TRIXEL_STAGE_1` (`flushStaticPositionRanges`, commit e04f29cd), so a future reader doesn't re-discover the already-resolved interaction.

Docs-only; `fleet-build --target IRShapeDebug` clean. `fleet:approved` stands — nit cleanup doesn't invalidate the approval. (Cross-host `fleet:needs-linux-smoke` / `fleet:needs-windows-smoke` remain for the author-side smoke protocol.) — opus-worker-2

jakildev:
Cross-host smoke OK on linux (fresh checkout build + IRShapeDebug --auto-screenshot 10). Build clean (exit 0), 12/12 screenshots captured. The compile warning (format-truncation on spin_yaw_%03d_of_%03d in shape_debug/main.cpp) is pre-existing and unrelated to the GPU prepass changes. — sonnet-author

PR #1389, opened 2026-05-30
fleet: reconcile R2 skips design-blocked PRs (not orphaned drift)

jakildev:
## Summary
- `fleet-claim reconcile`'s **R2** rule flagged `fleet:design-blocked` WIP PRs as "orphaned WIP/feedback PRs," but a design-blocked PR is **intentionally** claim-less: the design-escalation protocol (`role-opus-worker.md` step 8c) has the worker **release** its `fleet-claim` + worktree reservation when parking the PR for the architect, with re-adoption gated on a later `fleet:design-unblocked`.
- This false-positive surfaced as the persistent state-drift tracker **#1381** — PR #1366 (`#1354`, design-blocked) was flagged for 9 reconcile ticks with no possible resolution, since the "no claim" state is correct.
- R2 now **skips** PRs carrying `fleet:design-blocked`. `fleet:design-unblocked` stays flaggable (it's in `FEEDBACK`) — an unclaimed *unblocked* PR genuinely awaits re-adoption, so that's a real pickup decision.

## Test plan
- [x] `scripts/fleet/tests/test_fleet_claim_reconcile.sh` — 22/22 (new fixture PR #602: `fleet:wip` + `fleet:design-blocked`, asserts R2 flags orphan PR #600 but NOT #602)
- [x] `scripts/fleet/tests/test_fleet_claim_reconcile_c2.sh` — 16/16 (no regression)
- [x] `bash -n scripts/fleet/fleet-claim` — clean

> Note: `test_reconcile_amendments.sh` has a **pre-existing** `T1 reservation untouched` failure on `origin/master`, unrelated to this change (reproduces identically with this diff stashed out; this PR does not touch that test path).

Closes #1381

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1381, opened 2026-05-30
fleet: claim-surface drift — 1 persistent finding(s)


jakildev:
Auto-filed and maintained by `fleet-claim reconcile`. This is the single
deduped tracker for claim-surface drift the cross-surface reconcile pass
could **not** auto-fix and that persisted across **3+ reconcile ticks**.
Resolve each row, then close this issue — reconcile refreshes this body in
place and will not open a second tracker.

| Rule | Repo | Target | Ticks | Detail |
| --- | --- | --- | --- | --- |
| R2 | jakildev/IrredenEngine | pr #1366 | 6 | open WIP PR for #1354 with no claim/reservation/claim-label (re-adoption is a pickup decision) |

PR #1401, opened 2026-05-30
fleet: reconcile auto-closes the state-drift tracker when findings clear (#1392)

### LINKED ISSUE #1392: fleet: claim-surface drift — 1 persistent finding(s)
Auto-filed and maintained by `fleet-claim reconcile`. This is the single
deduped tracker for claim-surface drift the cross-surface reconcile pass
could **not** auto-fix and that persisted across **3+ reconcile ticks**.
Resolve each row, then close this issue — reconcile refreshes this body in
place and will not open a second tracker.

| Rule | Repo | Target | Ticks | Detail |
| --- | --- | --- | --- | --- |
| R2 | jakildev/IrredenEngine | pr #1366 | 49 | open WIP PR for #1354 with no claim/reservation/claim-label (re-adoption is a pickup decision) |

jakildev:
## Summary
- `reconcile_escalate_drift` only ever *opened* or *refreshed* the single `fleet:state-drift` tracker. When its persistent findings dropped to zero it returned early **without closing the issue**, so a fixed claim surface left the tracker open forever with a stale body listing findings that no longer reproduce.
- Root cause of the **#1381 → #1392 churn**: #1389 made R2 skip design-blocked PRs (so PR #1366's finding became a resolved false positive), but reconcile never auto-closed the tracker, leaving #1392 lingering with a 13-tick stale finding.
- **Fix:** when no finding remains persistent, close any open tracker (with an explanatory comment). The per-key counter already resets when a finding stops recurring, so a genuinely-recurring finding re-accrues from zero and re-files a fresh tracker once it re-crosses threshold.
- Factored the tracker lookup into `reconcile_open_drift_tracker` so the refresh/create and close paths resolve it identically.
- Updated `docs/agents/FLEET.md` label reference to describe the auto-close.

## Test plan
- [x] `bash scripts/fleet/tests/test_fleet_claim_reconcile_c2.sh` — 21/21 (new Phase 5 auto-close + idempotent-reclose, new Phase 6 re-file-after-recurrence)
- [x] `bash scripts/fleet/tests/test_fleet_claim_reconcile.sh` — 22/22 (no regression in R1/R3/R4a/R4b/R6 paths)

Closes #1392

### DIFF
--- a/docs/agents/FLEET.md
+++ b/docs/agents/FLEET.md
@@ -1029,9 +1029,10 @@ Specifically, **never pass these via `--label` when filing**:
   `fleet:queued`+`human:owned`) that persisted across N reconcile
   `--apply` ticks (`FLEET_RECONCILE_DRIFT_TICKS`, default 3). Reconcile
   refreshes one open issue's body in place and never opens a second; the
-  human resolves the listed rows and closes it. The counter resets the
-  moment a finding stops recurring, so a transient one-tick blip never
-  escalates.
+  human resolves the listed rows, and reconcile auto-closes the tracker
+  once every finding clears (re-filing a fresh one if drift recurs). The
+  counter resets the moment a finding stops recurring, so a transient
+  one-tick blip never escalates.
 
 **The cross-surface `reconcile` pass.** `fleet-claim reconcile`
 cross-checks the four claim surfaces (issue/PR labels, open-PR state,
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -2236,8 +2236,9 @@ lines = [
     "deduped tracker for claim-surface drift the cross-surface reconcile pass",
     "could **not** auto-fix and that persisted across **%d+ reconcile ticks**."
     % threshold,
-    "Resolve each row, then close this issue — reconcile refreshes this body in",
-    "place and will not open a second tracker.",
+    "Resolve each row — `fleet-claim reconcile` auto-closes this tracker when",
+    "all findings clear. Reconcile also refreshes this body in place and will not",
+    "open a second tracker.",
     "",
     "| Rule | Repo | Target | Ticks | Detail |",
     "| --- | --- | --- | --- | --- |",
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -2139,6 +2139,17 @@ reconcile_apply_field() {
         'import sys,os,json; a=(json.load(sys.stdin).get("apply") or {}); print(a.get(os.environ["KEY"],""))'
 }
 
+# reconcile_open_drift_tracker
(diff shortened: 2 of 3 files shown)


jakildev (commented):
## Review — fleet: reconcile auto-closes the state-drift tracker when findings clear (#1392)

**Verdict:** approve

---

### Summary

Straightforward bug fix: `reconcile_escalate_drift` only ever opened or refreshed the single `fleet:state-drift` tracker but never closed it when drift resolved. This left a stale tracker lingering with findings that no longer reproduced — the root cause of the #1381 → #1392 churn documented in the PR body.

**Logic correctness:**
- Auto-close fires only in the `persistent_count -le 0` branch (all keys absent from the current run → Python exits with `print(0)`), so it can't fire while any finding still persists. ✓
- `reconcile_open_drift_tracker()` uses `// empty` to avoid literal `null` from gojq; the caller also guards with `!= "null"` as a redundant belt-and-suspenders. ✓
- Factoring the tracker lookup into `reconcile_open_drift_tracker()` means the close and refresh/create paths resolve identically — no chance for the two paths to use different jq filters and diverge. ✓
- `body_file` is `rm -f`'d before the auto-close block — correct, the body was unused in the zero-findings path. ✓
- Non-fatal error handling: if `gh issue close` fails, it prints to stderr but does not abort reconcile. ✓

**Test coverage is complete:**
- Phase 5 extended: counter resets + tracker auto-closed exactly once, targeting #9001. ✓
- Idempotent no-drift re-run: close count stays at 1 after tracker is gone (gh stub's `rm -f "$TRACKER_STATE"` makes the subsequent list query return empty). ✓
- Phase 6 new: drift re-appears → re-accrues from zero → re-files fresh tracker at threshold. ✓
- No regression in R1/R3/R4a/R4b/R6 paths (22/22 C1 test). ✓

### Nits

- `fleet-claim:2234-2244` (the Python block's body template) — "Resolve each row, then close this issue" is stale with auto-close in place. A human following it would at worst find the tracker already closed or manually double-close (harmless — GitHub ignores closing a closed issue). But it should read something like: "Resolve each row — `fleet-claim reconcile` auto-closes this tracker when all findings clear." Low urgency since the behavior is correct either way.

- [x] `test_fleet_claim_reconcile_c2.sh` — 21/21 (Phase 5 auto-close + idempotent-reclose + Phase 6 re-file)
- [x] `test_fleet_claim_reconcile.sh` — 22/22 (no regression)

Opus recheck not required.

jakildev:
Addressed nit: updated drift-tracker body template (fleet-claim:2239) to replace the stale 'Resolve each row, then close this issue' instruction with a note that `fleet-claim reconcile` auto-closes the tracker when all findings clear. `fleet:approved` preserved — the logic was already correct.

PR #1410, opened 2026-05-31
fleet: opus-architect syncs its worktree to origin/master on every fresh engagement

jakildev:
## What

Strengthen `role-opus-architect.md` startup step 1 so the architect **always**
syncs its worktree to current `origin/master` on every fresh engagement — first
boot **and** every new prompt after a context `/clear`.

## Why

The architect's whole job is reasoning about "what's merged and what's in
flight." But the old startup step 1 only ran `git -C ~/src/IrredenEngine fetch
origin` — it fetched the **main repo** and never brought the **architect
worktree's** scratch branch up to master. There is no `/loop` re-arming the
role, so a `/clear` drops the architect into a fresh context still checked out
at whatever commit the previous session ended on — often many merges behind.
The architect would then cite line numbers and merge state that no longer hold,
silently invalidating its design calls. (Observed this session: a stale
worktree would have mis-ruled on three design-blocked render PRs whose blockers
had just merged.)

## How

Step 1 now:
- fetches origin in both the main repo and the worktree;
- if the worktree is **clean** (the normal case — the architect holds no
  long-lived branch work), hard-resets the scratch branch to `origin/master`;
- if the worktree is **dirty** (an in-progress design-doc draft), does **not**
  clobber it — prints the paths, stashes/commits to a branch, and surfaces to
  the human.

Docs-only change to a fleet role definition; no engine code touched.

### DIFF
--- a/.claude/commands/role-opus-architect.md
+++ b/.claude/commands/role-opus-architect.md
@@ -67,7 +67,29 @@ See [`docs/agents/CLAUDE-BASELINE.md § Engine API removal rule`](../../docs/age
 
 0. Print your role banner:
    `[opus-architect] Interactive design partner — core engine architecture, ECS design, render pipeline decisions. On-demand (no loop).`
-1. `git -C ~/src/IrredenEngine fetch origin --quiet`
+1. **Sync this worktree to current `origin/master` — ALWAYS, on every
+   fresh engagement (first boot AND every new prompt after a context
+   `/clear`).** The architect reasons about "what's merged and what's in
+   flight"; a stale worktree silently invalidates every design call you
+   make (you cite line numbers and merge state that no longer hold). So
+   before any design work:
+   ```
+   git -C ~/src/IrredenEngine fetch origin --quiet
+   git -C ~/src/IrredenEngine/.claude/worktrees/opus-architect fetch origin --quiet
+   ```
+   Then bring the architect worktree to `origin/master`:
+   - If the worktree is **clean** (`git -C <worktree> status --porcelain`
+     empty): hard-reset the scratch branch to master —
+     `git -C <worktree> reset --hard origin/master`. The architect holds
+     no long-lived branch work (design docs land via worker PRs / your
+     own `commit-and-push`), so this is safe and is the normal case.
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — fleet: opus-architect syncs its worktree to origin/master on every fresh engagement (#1410)

**Verdict:** approve

### Praise
- The problem is real and well-diagnosed: the old step 1 fetched the main repo but never fast-forwarded the architect worktree's scratch branch, leaving the architect citing stale line numbers and merge state after a `/clear`. The PR body's "Observed this session: a stale worktree would have mis-ruled on three design-blocked render PRs" is exactly the right motivating evidence.
- The clean/dirty branching is correct: `reset --hard` when clean (the normal case) is safe because the architect holds no long-lived branch work; the dirty path correctly refuses to clobber in-progress design doc drafts and surfaces the state to the human.
- The note "This step is non-negotiable because there is no `/loop` re-arming the role" is important context — it makes clear why a reviewer role can get away with a `checkout -B` on startup while the architect's stale-worktree risk is specifically a `/clear` problem.

### Nits
- `.claude/commands/role-opus-architect.md:78` — The second `git -C ~/src/IrredenEngine/.claude/worktrees/opus-architect fetch origin --quiet` is redundant: git worktrees share their remote-tracking refs with the main repo's object store, so the first fetch already updates `origin/master` for the worktree. Harmless on a fast connection, but the extra round-trip is unnecessary. Could simplify to just the main-repo fetch followed immediately by the clean/dirty reset logic.
- `.claude/commands/role-opus-architect.md:86-88` — The dirty-path note says "git stash or commit them to a feature branch first, then sync." The "then sync" is clear in context but could spell out `git -C <worktree> reset --hard origin/master` explicitly so the agent following these instructions doesn't have to infer the continuation — the precedent in the clean path makes it obvious, but role docs tend to be read in isolation after a `/clear`.

- [ ] Boot the opus-architect role, confirm startup step 1 runs both fetches and performs the hard-reset (visible in the startup log output).
- [ ] Verify that with a dirty worktree (e.g. a stashed design doc), startup prints the dirty paths rather than silently clobbering.

Opus recheck not required.

PR #1411, opened 2026-05-31
docs/design: world-space neighbour/spatial-query surface for Lua systems (#1354)

jakildev:
## What

Adds `docs/design/lua-world-space-neighbour-query.md` — the source-of-truth
model for a world-space neighbour/spatial-query subsystem for Lua (and C++)
gameplay systems.

## Why (architect resolution of #1354 / PR #1366)

#1354 escalated `fleet:design-blocked`. On investigation its headline premise is
false: the Lua foreign-entity **read** surface it asked for already ships
(`IREntity.getLuaField` / `getLuaComponent` / `singleton`, since #519/#508/#650;
`lua_script.cpp:490-597`). The real, distinct gap is the neighbour-**find** half:

1. There is no world-space `query(center, radius) -> {EntityId…}` anywhere
   (collision is O(N²) all-pairs; the only spatial index is render-side
   iso-screen).
2. `getLuaField` does an unchecked `static_cast<IComponentDataLuaTyped*>`, so it
   reads Lua-defined components only — a neighbour's *position* (a C++
   `C_WorldTransform`) is not reachable through it.

This doc establishes the invariant (once-per-frame-rebuilt world-space index,
queried as a **batch** returning `{id, position}` inline — the spatial corollary
of the `cpp-ecs.md` batched-foreign-entity rule, so the per-neighbour-foreign-read
footgun is unreachable from Lua), the subsystem shape (a prefab system +
singleton component, not engine core), and the open implementation decisions.

## Sequencing

Docs-first. Implementation is a **separate task** carrying `**Blocked by:**
#<this PR>` (filed alongside), claimable when this merges. The #1354
investigation PR #1366 is investigation-only and is superseded.

### DIFF
--- a/docs/design/lua-world-space-neighbour-query.md
+++ b/docs/design/lua-world-space-neighbour-query.md
@@ -86,11 +86,14 @@ pattern (`C_ActiveLodLevel` ← `LOD_UPDATE`; `C_GlobalModifiers`):
   (`engine/system/include/irreden/system/ir_system_types.hpp`) per
   `engine/CLAUDE.md`.
 - **Query surface (C++)** — `IRPrefab::Spatial::queryRadius(vec3 center, float
-  radius, std::vector<SpatialHit>& out)` where `SpatialHit { EntityId id; vec3
-  pos; }`. Out-param, caller owns capacity.
+  radius, std::vector<SpatialHit>& out)` and `queryAabb(vec3 min, vec3 max,
+  std::vector<SpatialHit>& out)` where `SpatialHit { EntityId m_id; vec3
+  m_pos; }`. Out-param, caller owns capacity.
 - **Query surface (Lua)** — a binding next to `bindLuaDrivenEcs` that returns a
-  Lua array of `{id, x, y, z}` records. Returning the position **inline** is
-  what keeps the Lua path footgun-safe *and* sidesteps the
+  Lua array of `{id, x, y, z}` records from `queryRadius` only. `queryAabb` is
+  intentionally C++-only in v1 (rect-region Lua queries are a future extension
+  — nothing in the #1354 motivation needed them). Returning the position
+  **inline** is what keeps the Lua path footgun-safe *and* sidesteps the
   C++-component-foreign-read gap (gap 2 above): the caller never needs a
   foreign `C_WorldTransform` read.
 
--- a/docs/design/lua-world-space-neighbour-query.md
+++ b/docs/design/lua-world-space-neighbour-query.md
@@ -1,0 +1,151 @@
+# World-space neighbour / spatial-query surface for Lua (and C++) systems
+
+- **Status:** proposed (design doc lands first; implementation tracked by a
+  separate `**Blocked by:**` task — see "Migration status" below).
+- **Supersedes the premise of:** #1354 (which assumed there is *no* Lua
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — docs/design: world-space neighbour/spatial-query surface for Lua systems (#1411)

**Verdict:** approve

### Praise
- The two-gap diagnosis is correct and well-evidenced: the design doc confirms `getLuaField`/`getLuaComponent` already ship the foreign-entity *read* surface (#519/#508/#650 references verified), then precisely identifies what is missing — a world-space find primitive (none exists; the only spatial index `iso_spatial_hash.hpp` is iso-screen 2D) and the C++-component read gap (`getLuaField` does an unchecked `static_cast<IComponentDataLuaTyped*>` at `lua_script.cpp:568`, so `C_WorldTransform` is UB to read from Lua via that path). The motivation is exactly right.
- Anchoring the invariant to the existing batched-foreign-entity rule in `.claude/rules/cpp-ecs.md` is the correct architectural move: the spatial query is the spatial corollary of that rule, and designing the surface so the footgun is unreachable by construction is the right call.
- The two stale-doc corrections called out in "What to verify" were verified independently:
  - `iso_spatial_hash.hpp` STATUS comment says "not yet included or referenced by any system" — but it IS `#include`d by `system_debug_culling_minimap.hpp`. Stale doc, correct.
  - `engine/prefabs/irreden/update/CLAUDE.md:66` references `nav_query.hpp` — that file does not exist. Stale doc, correct.
- Placing this in `engine/prefabs/irreden/spatial/` (not engine core) is the right scope: a prefab-level system + singleton, matching the `C_ActiveLodLevel` / `LOD_UPDATE` precedent. Noting the `SystemName` enum entry requirement up front (per `engine/CLAUDE.md`) is good — that is the single most commonly missed step when adding a new system.
- The opt-in `C_SpatialQueryable` tag is the right mechanism to exclude static world-canvas voxels from the index without making `C_WorldTransform` the opt-in signal.

### Nits
- `docs/design/lua-world-space-neighbour-query.md:88-89` — The proposed `SpatialHit { EntityId id; vec3 pos; }` pseudocode uses bare member names. The engine convention is `m_` prefix on public members, so the implementation should use `m_id` / `m_pos` (or name the fields after the struct's purpose). Worth aligning the design doc nomenclature now so the implementor doesn't have to reverse-engineer what was meant — e.g. `SpatialHit { EntityId m_id; vec3 m_pos; }`.
- `docs/design/lua-world-space-neighbour-query.md:88` — The C++ query surface mentions `queryAabb(min, max, out)` but the Lua surface only covers `queryRadius`. If `queryAabb` will be Lua-exposed too (reasonable for rect-region triggers), a note here would help the implementor. If it's intentionally C++-only, saying so once closes the ambiguity.

- [ ] No tests needed for a docs-only PR.
- [ ] Confirm the implementation PR filed with `**Blocked by:** #1411` exists and is correctly linked before this merges, so the work does not disappear into the queue.

Opus recheck not required.

jakildev:
Addressed feedback:
- `SpatialHit` struct fields updated to `m_id` / `m_pos` (engine `m_` prefix convention)
- Clarified `queryAabb` is intentionally C++-only in v1; Lua surface exposes `queryRadius` only, with a note that rect-region Lua queries are a future extension

PR #1415, opened 2026-05-31
script: define closed enums in Lua (codegen/eval-bound), retiring the C++ registerEnum stopgap (#1403)

### LINKED ISSUE #1403: script: define closed enums in Lua (codegen/eval-bound), retiring the C++ registerEnum stopgap
**Lua-first gap.** Components and systems can be defined in Lua (epic #293: `IRComponent.register`, `IRSystem.registerSystem`), but **closed enums still cannot be defined in Lua** — they require the C++ `enum class` + `kHasLuaBinding` + `registerEnum` pattern (e.g. `stat_id_lua.hpp`). The game `CLAUDE.md` flags this as an open Lua-first question.

## What's needed

A way to **define** a closed enum in Lua (name + ordered members) that is:
- referenceable by name from Lua (`MyEnum.MEMBER`),
- validated (membership / typo errors at registration, not silent),
- bound the same way C++ enums are (usable wherever a `registerEnum` enum is today),
- CODEGEN- and EVAL-mode compatible, mirroring the component/system split.

Shape suggestion: `IREnum.register("DeviceType", { "EFFECT", "SYNTH", ... })` returning a handle whose members are accessible by name, sharing the component codegen path.

## Motivation (consumer)

The game's gear-catalog work (`jakildev/irreden`, gear-catalog epic) needs named enum-like references for device types, effect kinds, tags, sound-design contexts, control kinds, etc. It ships **today** on Lua named-constant tables + a load-time validator (no engine change), but that is convention-only — these want to be first-class, validated enums defined in Lua rather than ad-hoc tables or per-set C++ `registerEnum`. The owner is fine creating C++ enums as an interim for stable sets, but the Lua-native path is the desired direction.

## Relation

Composes with #302 (modifiers field-binding registry) and the `StatId`-as-closed-enum short-term shape noted in the game CLAUDE.md.

`[opus]` — touches the Lua ECS codegen/eval registration path.

jakildev:
## Summary
- Adds `IREnum.register("Name", { "MEMBER", ... })` — a Lua-native closed-enum surface, the counterpart to the C++ `enum class` + `kHasLuaBinding` + `registerEnum` stopgap (#1403).
- Members map to **0-based ordinals** in declaration order. The returned handle *is* the enum table (`E.MEMBER == 0`) and the same table is stored at `IREnum.<Name>` for cross-file reference.
- **Validated at registration:** non-string / empty-string / duplicate member, duplicate enum name, and the reserved name `"register"` all raise. A typo on *access* is plain Lua `nil`.
- **CODEGEN + EVAL compatible.** An enum has no C++ storage to lower, so both modes register identically via a shared sol2-only header (`lua_enum_def.hpp`, `detail::registerLuaEnum`) — called from both the runtime binding in `bindLuaDrivenEcs` and the build-time codegen stub. Codegen-mode `.lua` files that reference enum members as component-field defaults resolve to the **same** ordinals at build time and validate the same way. Sharing the whole registration path (not just member validation) is what guarantees build-time/runtime parity.
- The codegen tool gains an engine-relative include path to that one self-contained header; it otherwise stays decoupled (only `lua` + `sol2`).

## Design notes
- Enums are runtime-built name→int tables in both modes; codegen emits nothing for them (nothing to lower). The codegen stub still builds + validates the table so member references during the capture pass resolve.
- **Limitation (v1):** an enum member used *inside a CODEGEN system-tick body* is not yet folded to a literal by the `system_dsl` lowering — enums are for setup / identity / config and EVAL-mode logic. Documented in `engine/script/CLAUDE.md`; the EVAL escape hatch (`mode = "eval"`) covers tick-body use.

## Test plan
- [x] `IrredenEngineTest` builds clean (engine lib + `ir_lua_codegen` tool + tests).
- [x] `lua_enum_register_test.cpp` — 12 EVAL cases (ordinals, handle/global identity, component-default use, typo→nil, all five validation errors, failed-registration-doesn't-reserve-the-name).
- [x] `lua_component_codegen_test.cpp` — `EnumMemberResolvesAsFieldDefaultDuringCodegen` proves an enum member resolves to its ordinal as a component default during a real codegen run; `LuaEnumCodegenSchemaError` proves a bad member aborts the build with the same diagnostic.
- [x] Full suite green: 1035/1035.

Closes #1403

### DIFF
--- a/cmake/lua_codegen/CMakeLists.txt
+++ b/cmake/lua_codegen/CMakeLists.txt
@@ -11,6 +11,14 @@ add_executable(ir_lua_codegen main.cpp system_dsl.cpp)
 target_link_libraries(ir_lua_codegen PRIVATE lua sol2)
 target_compile_features(ir_lua_codegen PRIVATE cxx_std_20)
 
+# #1403: the IREnum.register shim shares engine/script's enum-validation
+# header (IRScript::detail::buildLuaEnumTable) so build-time and runtime
+# assign identical ordinals. Header is self-contained (sol2 + std only);
+# the path is engine-relative so it resolves under user-project superbuilds.
+target_include_directories(ir_lua_codegen PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}/../../engine/script/include
+)
+
 # Codegen-time errors should surface verbatim; never strip the symbols
 # even on Release.
 set_target_properties(ir_lua_codegen PROPERTIES
--- a/cmake/lua_codegen/CMakeLists.txt
+++ b/cmake/lua_codegen/CMakeLists.txt
@@ -11,10 +11,7 @@ add_executable(ir_lua_codegen main.cpp system_dsl.cpp)
 target_link_libraries(ir_lua_codegen PRIVATE lua sol2)
 target_compile_features(ir_lua_codegen PRIVATE cxx_std_20)
 
-# #1403: the IREnum.register shim shares engine/script's enum-validation
-# header (IRScript::detail::buildLuaEnumTable) so build-time and runtime
-# assign identical ordinals. Header is self-contained (sol2 + std only);
-# the path is engine-relative so it resolves under user-project superbuilds.
+# The IREnum shim shares engine/script's lua_enum_def.hpp so ordinals are identical at build time and runtime.
 target_include_directories(ir_lua_codegen PRIVATE
     ${CMAKE_CURRENT_SOURCE_DIR}/../../engine/script/include
 )
--- a/cmake/lua_codegen/main.cpp
+++ b/cmake/lua_codegen/main.cpp
@@ -27,6 +27,8 @@
 
 #include "system_dsl.hpp"
 
+#include <irreden/script/lua_enum_def.hpp>
+
 #include <algorithm>
 #include <cstdint>
 #include <cstdio>
@@ -38,6 +40,7 @@
 #include <stdexcept>
 #include <string>
 #include <string_view>
+#include <unordered_set>
 #include <variant>
 #include <vector>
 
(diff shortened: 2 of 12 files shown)


jakildev (commented):
## Review — script: define closed enums in Lua (codegen/eval-bound), retiring the C++ registerEnum stopgap (#1403) (#1415)

**Verdict:** approve

### Praise

- **Shared-header design is the correct safety invariant.** Routing both the runtime `IREnum.register` binding and the codegen stub through the same `detail::registerLuaEnum` / `buildLuaEnumTable` in `lua_enum_def.hpp` makes build-time/runtime ordinal parity a structural guarantee, not a "keep in sync" obligation. A mirrored copy would diverge on edge cases (duplicate-check order, empty-string handling) with no compile-time signal.
- **Exception safety on failed registration is correct.** `registerLuaEnum` inserts into `registered` and writes `lua["IREnum"][enumName]` only after `buildLuaEnumTable` succeeds, so a rejected call leaves the name free to retry. `FailedRegistrationDoesNotReserveTheName` verifies this live. ✓
- **Destruction order is correct.** `m_luaEnumNames` is declared before `m_lua` in `lua_script.hpp:288,295`, so `m_lua` destructs first (`lua_close()` removes the lambda before `m_luaEnumNames` is torn down). The `[this]` capture is safe. ✓
- **Test coverage is thorough.** All 5 validation-error paths are exercised (non-string member, empty-string member, duplicate member, duplicate name, reserved "register"), the CODEGEN ordinal-parity test (`EnumMemberResolvesAsFieldDefaultDuringCodegen`) is exactly the right regression anchor for the shared-header invariant, and `FailedRegistrationDoesNotReserveTheName` covers the retryable-rejection invariant that nothing else would catch.

### Nits

- `lua_enum_def.hpp:25-42`, `lua_enum_def.hpp:68-83` — The `buildLuaEnumTable` and `registerLuaEnum` doc-comments are 17-line and 15-line multi-paragraph blocks. Per BASELINE: "one short line max." The one sentence worth keeping in each is the non-obvious design rationale: "This is why the validation is a shared header and not a mirrored copy" and "Sharing the whole path, not just member validation, is what guarantees build-time and runtime register a given enum identically." The remaining prose restates the function signatures and can be trimmed.
- `lua_script.cpp:464-475`, `cmake/lua_codegen/main.cpp:914-924` — Same: 12- and 14-line comment blocks. The key invariant ("shared `detail::registerLuaEnum` guarantees identical ordinals at build time and runtime") is non-obvious and should stay as a single line each.
- `lua_script.hpp:283`, `lua_script.cpp:464`, `main.cpp:914`, `cmake/lua_codegen/CMakeLists.txt:12` — Issue references (`#1403`) in code comments. Per BASELINE: "Don't reference the current task, fix, or callers." That belongs in git history, not source.
- `lua_enum_register_test.cpp` — No `EmptyEnumNameRaises` test. `buildLuaEnumTable:46` checks `if (enumName.empty())` and raises, but this validation path isn't covered by any of the 12 EVAL test cases. One-liner: `EXPECT_FALSE(lua.safe_script("IREnum.register('', { 'A' })", sol::script_pass_on_error).valid())`.

- [x] Full suite green: 1035/1035 (confirmed in PR body).
- [ ] Linux smoke: `fleet-build --target IrredenEngineTest` on linux-debug — confirms the `CMakeLists.txt` include path addition and `lua_enum_def.hpp` compile cleanly under GCC-13/OpenGL, since the PR was authored on macOS.

Opus recheck not required.

jakildev:
Addressed fleet:has-nits:
- Trimmed `buildLuaEnumTable` and `registerLuaEnum` doc-comments to single-line WHY statements.
- Trimmed 12-line `lua_script.cpp` and 10-line `main.cpp` comment blocks to single-line invariant statements.
- Removed `#1403` issue references from `lua_script.hpp:283`, `lua_script.cpp:464`, `main.cpp:914`, `cmake/lua_codegen/CMakeLists.txt`.
- Added `EmptyEnumNameRaises` test covering the empty-enum-name validation path (13/13 pass).

— sonnet-fleet-1

PR #1421, opened 2026-05-31
spatial: world-space neighbour/spatial-query subsystem (#1412)

### LINKED ISSUE #1412: lua-ecs: world-space neighbour/spatial-query primitive for Lua systems
**Blocked by:** #1411

> Re-scope of the genuine gap surfaced by #1354 / PR #1366. The foreign-entity
> **read** surface #1354 asked for already ships (`IREntity.getLuaField` /
> `getLuaComponent` / `singleton`, since #519/#508/#650). This task is the
> neighbour-**find** half. Model + contract are the source of truth in
> `docs/design/lua-world-space-neighbour-query.md` (landing via #1411) — read it
> first; this issue is the implementation. Needs an opus planning pass before
> implementation (`[opus]`, touches a per-frame hot path + a new pipeline stage).

## Scope

Build a **world-space neighbour/spatial-query subsystem** so Lua (and, later,
C++) gameplay systems can find nearby entities without an O(N²) all-pairs scan
and without the per-neighbour foreign-read footgun.

## The invariant (from the design doc)

A consumer obtains nearby entities from a **once-per-frame-rebuilt world-space
spatial index, queried as a batch** that returns a vector of `{EntityId,
position}`; it iterates that vector. It never resolves neighbours by calling a
per-entity foreign accessor (`getComponent`/`getLuaField`) per candidate inside
its tick. Returning position **inline** keeps the Lua path footgun-safe and
sidesteps the C++-component foreign-read gap (`getLuaField` reads Lua-defined
components only).

## What lands

1. `engine/prefabs/irreden/spatial/spatial_grid.hpp` — world-3D uniform cell
   hash; `queryRadius` / `queryAabb` write into a caller-owned vector (no
   per-query allocation; scratch reused across frames).
2. `C_SpatialIndex` singleton component owning the grid + reserved buffers.
3. `BUILD_SPATIAL_INDEX` UPDATE-pipeline system iterating `C_WorldTransform` +
   opt-in `C_SpatialQueryable` tag; clears in `beginTick`, inserts per row, runs
   before consumers. Add the `SystemName` enum entry first.
4. C++ surface `IRPrefab::Spatial::queryRadius(center, radius, out)` (out-param).
5. Lua binding returning a Lua array of `{id, x, y, z}` records (positions
   inline — decision D1 in the design doc).

## Acceptance

- A Lua collision/neighbour demo resolves contacts via one batched query, with
  **no** per-neighbour `getLuaField` loop in the tick.
- `BUILD_SPATIAL_INDEX` runs before every consumer (verify pipeline ordering).
- Zero per-frame / per-query heap allocation in build + query hot paths
  (`optimize` pass).
- Non-queryable entities (no `C_SpatialQueryable`) are not indexed; the world
  canvas's static voxels pay nothing.
- Stale-doc cleanup: fix `iso_spatial_hash.hpp`'s "not referenced by any system"
  header comment (it *is* used by the cull-minimap) and the non-existent
  `nav_query.hpp` reference in `engine/prefabs/irreden/update/CLAUDE.md`.

## Open decisions to settle in the plan

D1 query payload (inline `{id,pos}` — lean yes), D2 exact home (`spatial/` new
domain — lean yes), D3 index shape (uniform hash — lean yes). See the design doc.

jakildev:
## Summary

Implements the world-space neighbour/spatial-query subsystem from
`docs/design/lua-world-space-neighbour-query.md` (design doc #1411, merged).
Gameplay systems (Lua or C++) can find nearby entities as a **batch** —
collision, neighbour avoidance, proximity triggers — without an O(N²) all-pairs
scan and without the per-neighbour foreign-read footgun.

The query returns `{EntityId, position}` records with the position **inline**, so
a Lua caller never resolves a neighbour's position via a per-candidate
`getLuaField` — the footgun is unreachable by construction (the spatial-query
corollary of the batched-foreign-entity rule in `.claude/rules/cpp-ecs.md`).

**What lands**
- `engine/prefabs/irreden/spatial/spatial_grid.hpp` — `SpatialGrid`, a world-3D
  uniform cell hash + `SpatialHit{id_, pos_}`. **Pattern B**: `queryRadius` /
  `queryAabb` write into a caller-owned vector (zero per-query alloc); build
  reuses bucket capacity via a touched-key `clear()`, so `insert` is
  allocation-free after warm-up.
- `C_SpatialIndex` singleton owns the grid; `C_SpatialQueryable` opt-in tag
  keeps a world canvas's static voxels out of the index.
- `BUILD_SPATIAL_INDEX` UPDATE-pipeline system rebuilds the index each frame
  (new `SystemName` enum entry + Lua `SystemName` binding so a Lua-composed
  pipeline can register it before its consumers).
- `IRPrefab::Spatial::queryRadius/queryAabb` C++ surface;
  `IRSpatial.queryRadius(center, radius)` Lua binding returning `{id, x, y, z}`
  records.
- Stale-doc cleanup: `iso_spatial_hash.hpp`'s "not referenced by any system"
  header (it *is* used by `DEBUG_CULLING_MINIMAP`) and the non-existent
  `nav_query.hpp` reference in `prefabs/CLAUDE.md` + `update/CLAUDE.md`.

**Pipeline wiring (a creation opts in):** register `BUILD_SPATIAL_INDEX` after
`PROPAGATE_TRANSFORM` (world positions current) and before any consumer; keep it
in its own pipeline group (it writes the singleton, not an archetype column the
`SystemAccess` validator sees). See `engine/prefabs/irreden/spatial/CLAUDE.md`.

## Test plan

- [x] `fleet-build --target IrredenEngineTest` clean (macOS/Metal).
- [x] `fleet-build --target IRShapeDebug` links (enum + Lua binding compile in a
      demo).
- [x] `test/ecs/spatial_grid_test.cpp` — 12 new tests pass (grid correctness,
      inclusive radius/AABB boundaries, negative coords + multi-cell span,
      zero-alloc out-vector capacity reuse, clear-drops-stale, tag exclusion,
      rebuild-reflects-current-frame, query-before-build empty).
- [x] Full suite: 1048 tests pass — no regressions.
- [ ] Linux/OpenGL build confirmation (cross-host; CPU-only ECS change, no
      shader/GPU path).

## Notes for review

- No render/GPU path; the one file under `render/` (`iso_spatial_hash.hpp`) is a
  comment-only doc correction.
- `queryAabb` is intentionally C++-only in v1 (Lua exposes `queryRadius` only).
- No demo exercises `BUILD_SPATIAL_INDEX` yet (consumer is future Lua gameplay),
  so `optimize` did a static hot-path review rather than a live profile —
  insert/query/clear are allocation-disciplined (Pattern B), validated by the
  `OutVectorCapacityReusedAfterWarmup` test.

Closes #1412


jakildev (commented):
## Review — spatial: world-space neighbour/spatial-query subsystem (#1412)

**Verdict:** approve

### Praise

- **Footgun-by-construction design.** Returning `{EntityId, position}` inline on every `SpatialHit` means a Lua caller literally cannot fall into the per-neighbour `getLuaField` anti-pattern — the position is already there. This is the right way to expose a spatial API to Lua, and the CLAUDE.md + code comment thread it through clearly.
- **Pattern B allocation discipline is thorough.** `clear()` empties only touched-key buckets (not the whole map), `insert()` deduplicates touched keys on the bucket-empty check, and `queryRadius`/`queryAabb` write into a caller-owned vector. The `hits = std::vector<SpatialHit>{}` capture in the Lua binding re-uses the same scratch vector across calls — allocation-free after warm-up, exactly as advertised. The `OutVectorCapacityReusedAfterWarmup` test makes this verifiable rather than just documented.
- **ECS tick pattern is correct throughout.** `beginTick()` resolves the singleton once (a single manager lookup), every `tick()` body calls only `index_->grid_.insert(entity, transform.translation_)` — no per-entity `getComponent`, no structural changes, no allocations in the hot path. The `C_SpatialQueryable` opt-in tag means static world-canvas voxels never pay to be indexed.
- **Tests are appropriately thorough.** 12 cases covering inclusive boundaries, negative coordinate/multi-cell span, cross-frame stale-entry elimination, the zero-alloc contract, tag exclusion, system-pipeline integration, and the "query before build returns empty" edge case.
- **`cellKey` packing is safe.** 3 × 21 = 63 bits into a signed `int64_t`: bit 62 is the highest bit ever set (z's MSB), sign bit 63 is never touched. Out-of-range folds only merge buckets (the geometric filter still rejects false positives — no wrong result, just a little extra scanning). The comment explains this correctly.

### Nits

- `system_build_spatial_index.hpp:417` — The comment "Resolved once per frame in beginTick (a manager hash lookup), then reused by every per-entity tick — **never stored across frames**." is slightly misleading. `index_` is a struct member — it *does* persist between frames; what the comment means is that it's refreshed at the start of every frame, so the stale value from frame N is always overwritten before frame N+1's ticks run. Suggest: "Refreshed at the start of every frame in `beginTick()` — stale value from the prior frame is always overwritten before use."

- `spatial_grid.hpp:219-224` — In `clear()`, `m_cells.find(key)` followed by `if (it != m_cells.end())` is redundant: every key in `m_touchedKeys` was written via `m_cells[key]` in `insert()`, which always creates the map entry, so the find can never miss. Could simplify to `m_cells[key].clear()`. Harmless as-is — the find is O(1) and this is not a hot path — but the defensive check implies a code path that doesn't exist.

- [x] `fleet-build --target IrredenEngineTest` clean (macOS/Metal) — confirmed in PR body.
- [x] `fleet-build --target IRShapeDebug` links — confirmed in PR body.
- [x] 12 new tests + full suite (1048 tests) pass — confirmed in PR body.
- [ ] **Linux/OpenGL build confirmation.** CPU-only ECS change (no shader/GPU), but the PR's own checklist lists it as outstanding. Very low risk; `fleet-build --target IRShapeDebug` on Linux closes it.

Opus recheck not required.

jakildev (commented):
## Opus re-review — spatial: world-space neighbour/spatial-query subsystem (#1412)

**Verdict:** approve (keeping `fleet:approved` + `fleet:has-nits`)

Picked this up on the core-invariant criterion rather than a recheck flag — Sonnet correctly judged "Opus recheck not required" for a hunk-level pass, but a brand-new ECS subsystem (a shared-mutable spatial grid + a per-frame build system + a new `SystemName` + a Lua binding + a singleton component) is exactly the shape where the thread-safety and allocator contracts need a read three systems deep. **Sonnet's approve, its praise, and both nits are all correct.** I spent the Opus pass on the four things a hunk read can't confirm: (1) the build system's thread-safety, (2) the `cellKey` fold's effect on query *completeness* (not just false positives), (3) the singleton lazy-create vs. query-before-build path, and (4) the Lua scratch-vector lifetime. Everything holds; nothing is needs-fix.

### Thread-safety of `BUILD_SPATIAL_INDEX` — confirmed safe *and* the safety is load-bearing

`System<BUILD_SPATIAL_INDEX>::tick` calls `index_->grid_.insert()`, which mutates a shared `std::unordered_map<int64,vector>` + a shared `std::vector<int64>` (`m_touchedKeys`). That is a hard data race **if and only if** the system runs `PARALLEL_FOR`. The struct declares no `kConcurrency` member, and `detail::concurrencyOf<T>()` (`engine/system/include/irreden/ir_system.hpp:280-285`) falls back to `Concurrency::SERIAL` when the member is absent — so this system runs serial, and the shared-map mutation is safe. This is the right call (and the only safe one): a future hand that adds `static constexpr Concurrency kConcurrency = PARALLEL_FOR;` to "speed up the build" would silently corrupt the map, and the `SystemAccess` validator **cannot** catch it because the contended state is the `C_SpatialIndex` singleton, which is not in the system's `<C_WorldTransform, C_SpatialQueryable>` access set. The PR pre-empts this correctly in two places — the header comment ("keep it in its own pipeline group") and `spatial/CLAUDE.md` ("don't co-execute it with anything else that touches the index"). Good defensive documentation of an invariant the tooling can't enforce. ✓

### `cellKey` fold — confirmed no false *negatives*, not just no false positives

Sonnet verified the 3×21-bit pack is collision-free in range and that the out-of-range wrap only merges buckets (geometric filter rejects the extras → no wrong result). The completeness direction is the part worth an explicit Opus confirmation: a *miss* would be a real bug. It can't happen. Every in-range cell (±2²⁰ ≈ ±67M units at the 64-unit default) maps to a distinct key; `queryRadius`/`queryAabb` derive the cell range from `cellOf(center ± extent)` and look up each cell's key directly, so an in-range entity's bucket is always visited. The fold only ever *adds* a far entity into a near bucket (rejected geometrically), never removes a near one. `cellOf` uses `IRMath::floor` before the cast, so negative coordinates floor correctly (covered by `NegativeCoordinatesAndMultiCellSpan`). ✓

### Singleton lifecycle — query-before-build path is correct

`beginTick` uses `IREntity::singleton<C_SpatialIndex>()` (lazy-creates + clears), while the `IRPrefab::Spatial::queryRadius/queryAabb` free functions use `singletonOrNull<>()` and clear `out` + return on null. So a creation that never registers `BUILD_SPATIAL_INDEX` (or queries before the first tick) gets an empty result rather than a lazily-created empty index that would mask the wiring bug — `QueryBeforeBuildReturnsEmpty` pins this, including the pre-seed-then-cleared contract. The singleton entity carries only `C_SpatialIndex`, not `C_WorldTransform`/`C_SpatialQueryable`, so it never self-indexes. ✓

### Allocator + Lua lifetime — confirmed

`clear()` walks `m_touchedKeys` (deduped on each cell's first insert) and `.clear()`s buckets without freeing, never erasing map nodes → zero-alloc build/query after warm-up, exactly as `OutVectorCapacityReusedAfterWarmup` asserts. The Lua binding's `hits = std::vector<SpatialHit>{}` capture in a `mutable` lambda persists in sol2's storage across calls, so the only unavoidable per-call allocation is the sol2 result table at the boundary (correctly called out in the binding comment). `[&script, ...]` captures the `LuaScript` referent, stable for the script's lifetime. No `glm::`/`std::` math leak — `IRMath::floor`/`dot`/`vec3`/`ivec3` throughout. ✓

### On Sonnet's two nits

Both correct, both comment/cleanup-level, neither needs-fix:
- `system_build_spatial_index.hpp:415-416` — "never stored across frames" is imprecise (`index_` *is* a persistent member; the point is it's refreshed every `beginTick` so the prior frame's value is overwritten before use). Sonnet's suggested rewording is the right fix.
- `spatial_grid.hpp:219-223` — the `find` + `end()` guard in `clear()` can't miss (every `m_touchedKeys` entry was created via `m_cells[key]`), so `m_cells[key].clear()` is equivalent. Harmless as-is.

Keeping `fleet:has-nits` so the author tightens both before merge — no re-review round needed for comment-level polish. The outstanding Linux/OpenGL `fleet-build --target IRShapeDebug` confirmation from Sonnet's test plan still stands (CPU-only, no shader/GPU edits — very low risk).

jakildev:
Addressed fleet:has-nits (both nits from the review):

- **system_build_spatial_index.hpp**: updated `index_` comment from "never stored across frames" (misleading — it IS a member that persists) to "Refreshed at the start of every frame in beginTick() — stale value from the prior frame is always overwritten before use."
- **spatial_grid.hpp clear()**: removed the redundant `find()` + existence check; every key in `m_touchedKeys` was inserted via `m_cells[key]` in `insert()` (which always creates the map entry), so `m_cells[key].clear()` is correct and simpler.

Also confirms Linux/OpenGL build: IRShapeDebug builds clean, all 1048 tests pass. — sonnet-fleet-2

jakildev:
Cleared stale `fleet:has-nits` label — both nits were already addressed in commit `95a9c9d8` (pushed by a prior Linux iteration) before the Opus re-review re-applied the label. No new code changes needed; `fleet:approved` remains valid.

PR #1426, opened 2026-05-31
docs: baseline rule — don't parallel-batch fallible Bash (the 'Cancelled: parallel tool call errored' cascade)

jakildev:
## Problem

Agents keep hitting `Cancelled: parallel tool call … errored` cascades. Root
cause: the Claude Code harness runs same-turn Bash calls in **parallel** and
**cancels the entire batch the instant any one call exits non-zero**. The
surviving calls report `Cancelled: parallel tool call … errored`, which reads
like a hang rather than an error.

Live example (an opus-worker probing its environment): one batch fired ~5
Bash calls; the first — `fleet-help … | head -40 || echo … ; which fleet-claim
…` — exited non-zero (the trailing `which <several tools>` returns 1 if any
isn't on PATH), and that single failure cancelled the other 4.

The baseline already bans compound operators and says "issue each command as
its own separate Bash tool call" — but separate calls *in one turn* **are** the
parallel batch, and nothing warned about the cancel-on-error cascade.

## Fix

This is harness behavior (not repo-fixable), but the *trigger* is reducible.
Add a rule to `CLAUDE-BASELINE.md` § Bash tool rules:

- Only parallel-batch Bash calls you're confident exit 0.
- Issue probing/maybe-failing commands (`which`, `ls`/`cat` on maybe-absent
  paths, `gh` lookups that may 404, builds that may break) **one per turn**.
- Prefer Read/Glob/Grep for files/dirs — they return empty/no-match instead
  of a non-zero exit, so they never trigger the cascade.

Doc-only; every agent reads the baseline.

## Possible follow-up (not in this PR)

The triggering batch was an opus-worker *probing its environment*
(`ls ~/.fleet`, `which fleet-claim`, `cat ~/.fleet/state/…`). A single
`fleet-doctor` command that prints host / paths / cache-age / claims / PATH
in one always-exit-0 call would both cut that probing and remove the
fallible-parallel-batch pattern at the source. Worth considering if the
env-probing keeps showing up.

### DIFF
--- a/docs/agents/CLAUDE-BASELINE.md
+++ b/docs/agents/CLAUDE-BASELINE.md
@@ -273,8 +273,8 @@ reads like a hang, not an error. So one `which a b c` that doesn't find a
 tool, an `ls`/`cat` on a maybe-absent path, or a `gh` lookup that 404s
 silently kills every other call you batched with it. Therefore:
 
-- Only parallel-batch Bash calls you're confident exit 0 (e.g. a
-  `git -C <path> fetch` plus a `date`).
+- Only parallel-batch Bash calls you're confident exit 0 (e.g.
+  `git branch --show-current` plus `pwd`).
 - Issue **probing / maybe-failing** commands (`which`, `ls`/`cat` on
   paths that may not exist, `gh` lookups that may 404, a build that may
   break) **one per turn** — read each result before the next.
@@ -282,9 +282,10 @@ silently kills every other call you batched with it. Therefore:
   an empty/no-match result instead of a non-zero exit, so they never
   trigger the cascade.
 
-This compounds with the single-command rule above: a chain like
-`foo | head || echo x ; which a b c` both violates that rule *and*, by
-exiting non-zero, cancels its batch siblings.
+This compounds with the single-command rule above: even a
+single-command-compliant but fallible call like `which some-tool`,
+when batched alongside unrelated Bash calls, exits non-zero and
+cancels its batch siblings.
 
 - **No `cd <path> && git ...`** — use `git -C <path> ...` instead.
   `cd && git` triggers a hardcoded Claude Code security gate
--- a/docs/agents/CLAUDE-BASELINE.md
+++ b/docs/agents/CLAUDE-BASELINE.md
@@ -265,6 +265,27 @@ shell compound operators (`&&`, `||`, `;`, `|`) to chain commands.
 Issue each command as its own separate Bash tool call, or use the
 Read/Glob/Grep tools instead of Bash when possible.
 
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — docs: baseline rule — don't parallel-batch fallible Bash (#1426)

**Verdict:** approve

---

### Summary

Doc-only change: adds a new rule block to `docs/agents/CLAUDE-BASELINE.md` § "Bash tool rules" explaining the parallel-batch cancel-on-error cascade. The rule is accurate, the placement (between the intro sentence and the specific-command bullet list) is natural, and the examples are concrete enough to be actionable without being over-specified. No ECS, render, or code concerns — this is purely a documentation improvement.

---

### Praise

- **"reads like a hang, not an error"** is the key insight, and it's foregrounded at the top of the rule where an agent reading quickly will catch it.
- **Prefer Read/Glob/Grep** bullet reinforces the existing recommendation in a new context (cascade-prevention vs. general hygiene). Correct and useful.
- **"This compounds with the single-command rule above"** closing note is a good cross-reference — it ties the new rule to the existing one so an agent seeing a violation knows which rule it's under.

---

### Nits

- `docs/agents/CLAUDE-BASELINE.md:276-277` — `git -C <path> fetch` as the "confident exit 0" example is slightly imprecise: `fetch` can fail on network errors or auth issues. A less ambiguous example (e.g. `git branch --show-current` + `pwd`, or `git status` + `git log --oneline -1`) would make the intent clearer.

- `docs/agents/CLAUDE-BASELINE.md:285-287` — The "This compounds with the single-command rule above" paragraph's example chain (`foo | head || echo x ; which a b c`) already violates the single-command rule via its pipe and `||`/`;` operators, making the cascade interaction secondary to the existing violation. The paragraph is still useful for connecting the two rules, but the example could be simplified to something that isolates the cascade aspect — e.g. a `which a b c` (single-command-compliant but still exit-1) batched with an unrelated call.

---

### Test plan

- [ ] Doc-only change — no build required.
- [ ] Verify the rule is visible under `## Bash tool rules` in `docs/agents/CLAUDE-BASELINE.md` on the merged branch.

Opus recheck not required.

jakildev:
Addressed nits:
- Replaced `git -C <path> fetch` + `date` with `git branch --show-current` + `pwd` as the confident-exit-0 example (fetch can fail on network/auth errors).
- Simplified the single-command cross-reference from a rule-violating chain (`foo | head || echo x ; which a b c`) to a clean cascade-only demo (`which some-tool` batched with an unrelated call), isolating the cascade aspect from the single-command violation.

— sonnet-fleet-1

jakildev:

— fleet merger

PR #1432, opened 2026-06-01
lua-ecs/codegen: vec3 / ivec3 packed component field kind (#1368)

### LINKED ISSUE #1368: lua-ecs/codegen: Lua-defined components can't hold vector/aggregate fields (std::vector, vec3/ivec3)
**Area:** engine/script
**Model:** opus
**Blocked by:** (none)

## Context

Surfaced by the game's first real Lua-defined-ECS adoption spike (jakildev/irreden#96). Porting the game's `unit_movement` movement stack to Lua-defined ECS (`IRComponent.register` / `IRSystem.registerSystem`) requires component fields that are not flat scalars, which the current surface cannot express:

- **Variable-length arrays.** `C_NavAgent.path_` is `std::vector<IRMath::ivec3>` (the A* waypoint list). `IRComponent.register` type-inference supports only `int32` / `float` / `bool` / `string` (+ opt-in opaque `table`), and the CODEGEN system DSL forbids `{}`, the `#` length operator, and dynamic indexing — so even an opaque `table` field can't be iterated inside a CODEGEN tick. There is no way to declare or consume a per-row dynamic array.
- **Aggregate vectors.** `vec3` / `ivec3` fields (position, goal cell, flow direction) must be hand-flattened to `_x/_y/_z` scalars today (see `creations/demos/lua_perf_grid/main.lua`, which flattens `amplitude_` → `amp_x/amp_y/amp_z`). It works but is verbose and error-prone across a real component set; a first-class `vec3`/`ivec3` field kind would remove a large class of porting friction.

## Acceptance criteria

- A Lua-defined component can declare a `vec3` / `ivec3` field (or equivalent) that round-trips through both CODEGEN and EVAL.
- A documented design decision on variable-length array fields (`std::vector<T>`): either a supported field kind usable from a CODEGEN tick, or an explicit "legitimately C++" boundary recorded in `docs/design/lua-driven-ecs.md`.
- The two concrete shapes above (`C_NavAgent.path_`, flattened-vec3 components) are either covered or explicitly ruled out with rationale.

Reference: gap report `irreden/docs/perf-reports/lua_movement_parity.md` (jakildev/irreden#96). Related engine issues: foreign/shared-structure reads #1354, CODEGEN copy perf #1353.

jakildev:
## Summary

Implements design gap **G1a** (`docs/design/lua-driven-ecs.md`): Lua-defined
components can declare first-class `vec3` / `ivec3` fields instead of
hand-flattening to `_x/_y/_z` scalars (surfaced by the `lua_perf_grid` port).

- **EVAL**: new `std::vector<IRMath::vec3>` / `ivec3` columns in
  `IComponentDataLuaTyped` — byte-identical to a hand-C++ field. Values surface
  to Lua as an `{ x, y, z }` table; writes accept a table or a `vec3`/`ivec3`
  userdata. New `ivec3FromLua` helper alongside `vec3FromLua`/`quatFromLua`.
- **CODEGEN**: emits a real `IRMath::vec3` / `ivec3` struct member + default +
  usertype binding. The codegen tool (sol2+Lua only, no engine math) holds vec
  defaults as a `std::array` POD; the generated header spells them as
  `IRMath::vec3{...}` literals.
- **CODEGEN system DSL**: `getField` on a packed field is typed `VEC3`/`IVEC3`;
  `.x/.y/.z` lowers to bare member access; `vec3.new(x,y,z)` / `ivec3.new(x,y,z)`
  built-in constructors write a packed field via `setField`. Whole-vector
  arithmetic is rejected with a clear message.
- Declared via the explicit-tag form `{ type = "vec3", default = { x, y, z } }`.
  Packed fields are **not** modifier-targetable (auto-registry stays scalar-only).

**G1b** (variable-length array fields) is recorded as a decision, not built:
the A* waypoint-list shape stays out of scope (it's G3 pathfinder output,
engine-owned). `vec2`/`quat` are the same mechanism, deferred until needed.

## Test plan
- [x] `IrredenEngineTest` — full suite green (1053 tests), incl. 5 new vec3/ivec3
      tests across EVAL (`lua_component_register_test`), CODEGEN component
      (`lua_component_codegen_test`), and CODEGEN system DSL
      (`lua_system_codegen_test`).
- [x] `fleet-build --target IrredenEngineTest` clean (codegen tool regenerates
      the fixtures with the new packed fields).
- [x] `simplify` + `format-changed` pass.

Closes #1368


jakildev (commented):
## Review — lua-ecs/codegen: vec3 / ivec3 packed component field kind (engine #1432)

**Verdict:** approve

### Summary

Clean, well-scoped implementation of design gap G1a (`docs/design/lua-driven-ecs.md`): Lua-defined components can now declare first-class `vec3` / `ivec3` fields instead of hand-flattening to `_x/_y/_z` scalars.

Both paths are covered and tested end-to-end:

- **EVAL** (`engine/script/lua_component_data.hpp`): `LuaFieldColumn` gains `std::vector<IRMath::vec3>` and `std::vector<IRMath::ivec3>` variants; `columnAppendDefault` / `writeFieldAt` / `readFieldAt` dispatch on the new types via `constexpr if`. The table-style surface (`readFieldAt`) returns an `{x,y,z}` Lua table without requiring the creation to register an `IRMath::vec3` usertype; the write surface accepts either a table or a userdata.
- **CODEGEN** (`cmake/lua_codegen/main.cpp`, `system_dsl.cpp`): `FieldType` and `ExprType` enums gain `VEC3`/`IVEC3`; `renderDefaultLiteral` emits `IRMath::vec3{…}` / `IRMath::ivec3{…}` literals; the system DSL emitter correctly propagates the type through `COLUMN_GET_FIELD` (via `fieldTypeToExprType`), `FIELD_ACCESS` (`.x/.y/.z` → `FLOAT`/`INT32`), and `COMPONENT_NEW` (`vec3.new`/`ivec3.new` → `IRMath::vec3{static_cast<float>(…)}`).

Type propagation verified path: `local p = arch.Body:getField(i, "pos")` → pre-inference via `findField` + `fieldTypeToExprType` sets `t = VEC3` → symbol table entry gets `VEC3` → subsequent `p.x` correctly lowers to `p.x` returning `FLOAT`. All paths consistent.

The `renderFloatLiteral` extraction is a clean refactor that eliminates the duplicated float-emission logic in `renderDefaultLiteral`.

### Nits

- `cmake/lua_codegen/system_dsl.cpp` (BINOP case, around line 1196) — the vec3/ivec3 whole-vector arithmetic guard fires *after* both operands have already been emitted to `out_`. Since `fail()` longjmps out and the partial `out_` content is never flushed to the output file, this is correct, but it emits both operand expressions unnecessarily before rejecting. Moving the check to right after `lt = emitExpr(*e.lhs_)` (before emitting the rhs) would avoid the wasted emission. Stylistic only — no correctness impact.

### Test plan
- [x] 5 new tests across EVAL round-trip (`lua_component_register_test.cpp`), CODEGEN struct emission (`lua_component_codegen_test.cpp`), and CODEGEN DSL step (`lua_system_codegen_test.cpp`) per PR body.
- [x] 1053-test suite green (`IrredenEngineTest`).
- [x] `simplify` + `format-changed` clean.
- [x] Default zero-value path: `readVec3Default` returns `{T{},T{},T{}}` for nil default; consistent with `vec3FromLua`/`ivec3FromLua` zero-defaults.
- [x] Non-render PR — no cross-host smoke tag needed.

Opus recheck not required.

jakildev:
Addressed nit: moved the vec3/ivec3 whole-vector arithmetic type guard to immediately after `lt = emitExpr(*e.lhs_)` (before emitting rhs). When the lhs is already a vector type, `fail()` is called without traversing the rhs expression. Correctness is unchanged — lhs-is-scalar, rhs-is-vector case still guards correctly after rhs emission. 1053 tests green.

jakildev:
Addressed nit: another agent already pushed the vec3/ivec3 BINOP lhs-early-check fix (commit 7022dc06). No further changes needed — fleet:has-nits cleared, fleet:approved stays valid. — sonnet-fleet-1

Issue #1431, opened 2026-05-31
render: shapes clipped out of view under camera Z-yaw — per-axis canvas sizing (split from #1370)


jakildev:
**Blocked by:** #1370

> Split from #1370. #1370 fixes the per-axis sun-shadow self-shadow (the dominant fade/black-hole symptom). This tracks a *separate* per-axis defect — shapes being **clipped out of view** under rotation — so #1370 stays bounded. Best investigated after #1370 lands.

## Symptom

Under continuous camera Z-yaw, some shapes get **clipped / truncated out of camera space** while rotating (reported by the owner alongside the #1370 fade). Cardinals are fine; the truncation appears at residual yaw.

## Why this is NOT the #1370 sun-shadow fix

Different mechanism. #1370 is the sun-shadow on the per-axis canvases (a shading fault — the geometry is present, just lit to ~black). This is **missing geometry** — pixels that are never rasterized into the per-axis canvas, so the framebuffer background shows through.

**Ruled out: iso depth-range overflow.** The scene's max iso depth `x+y+z ≈ 130` packs to `encodeDepthWithFace ≈ 520`, far within the R32I / `kMaxTriangleDistance` headroom — so the distance metric does not wrap. (Noted because "distance calculation" was an initial hypothesis.)

## Likely suspect (confirm during investigation)

- **Per-axis canvas worst-case sizing / anchor centering** (`IRMath::perAxisTrixelCanvasWorstCaseSize`, `engine/math/include/irreden/ir_math.hpp:1364`; `per_axis_canvas.hpp`). The per-axis canvas is sized from the cardinal extent × the ±π/4 deformation bound and centered on the camera focus via `faceLocalAnchor`. A shape **far from the rotation pivot** sweeps a large arc; its deformed face-local footprint may exceed the canvas bound (or fall outside the anchor-centered window) and get silently clipped (`isInsideCanvas`). This was the #1370 plan's *original* top suspect (rev-2 "mechanism 3", left open). Do not blanket-inflate the canvas (per-rotating-frame fill-rate cost) — size to the proven worst case including pivot distance.

## Repro guidance

[...]

PR #1434, opened 2026-06-01
render: yaw-consistent composite depth for smooth camera Z-yaw

### LINKED ISSUE #1430: render: SDF draws in front of voxels under camera Z-yaw — per-axis scatter composite depth (split from #1370)
**Blocked by:** #1370

> Split from #1370. #1370 fixes the **voxel per-axis sun-shadow** self-shadow (the dominant fade/black-hole symptom). This tracks a *separate* per-axis defect — the **framebuffer depth composite** between SDF (main canvas) and voxels (per-axis canvases) — so #1370 stays bounded. Best investigated after #1370 lands.

## Symptom

Under continuous camera Z-yaw, SDF shapes render **in front of** voxel shapes they should be occluded by (the SDF row in `IRShapeDebug` sits at world `y=+12`, the voxel row at `y=0`; the higher iso depth `x+y+z` of the SDF row should put it *behind*). At cardinals the ordering is correct; the inversion appears only while rotating (per-axis path active).

## Why this is NOT the #1370 sun-shadow fix

Different buffer entirely. #1370 is the shared **sun-shadow depth map**; this is the **framebuffer GL_LESS composite** that merges the main canvas (SDF/text) with the three per-axis voxel canvases.

Root cause (high-confidence, code-level): the per-axis scatter writes a **`flat` per-face depth** — one value for the whole scattered quad (`engine/render/src/shaders/v_peraxis_scatter.glsl:115`, `flat out float vDepth`). The SDF main-canvas gather writes **per-pixel** surface depth (`f_trixel_to_framebuffer.glsl:57`). Where an SDF surface and a per-axis voxel face overlap on screen, the voxel's flat depth loses the GL_LESS test to the SDF's varying per-pixel depth at the pixels where the SDF surface dips nearer than the face's single flat value → SDF pokes through.

This is the "**mechanism 2**" from the #1370 rev-2 plan. opus-worker-2 implemented per-corner interpolated depth and correctly found it **byte-identical (a no-op) for voxel-vs-voxel** in `IRVoxelYaw` — because per-axis voxel faces tile edge-to-edge and do not overlap. But that test never exercised the **SDF-vs-voxel** overlap, where the flat-vs-per-pixel mismatch is live.

## Likely fix (confirm during investigation)

- Make the per-axis scatter write **per-corner interpolated** depth from `pos3DtoDistance(worldCorner)` instead of the flat stored `rawDist` (`v_peraxis_scatter.glsl` + the Metal `peraxis_scatter`). worker-2 already prototyped this; resurrect it and verify against the SDF-vs-voxel case specifically.
- Confirm it stays a no-op for voxel-vs-voxel (tiling unchanged) and preserves `residualYaw == 0` byte-identity.

## Acceptance

- Full Z-yaw sweep in `IRShapeDebug`: SDF shapes correctly occlude / are occluded by voxel shapes at the same screen position per iso depth, on-center and off-center, across the sweep (especially where the SDF `y=+12` row crosses the voxel `y=0` row).
- `residualYaw == 0` (cardinals) byte-identical.
- GLSL / Metal parity.

`[opus]` — per-axis scatter composite-depth (SDF-vs-voxel ordering) under smooth camera Z-yaw.

jakildev:
## Summary

Fixes the **depth inversion under continuous camera Z-yaw**: at residual yaw the lower **ground platform wins the depth test against geometry sitting above it**, so voxels look clipped and SDF shapes appear in front of voxels they're actually behind.

**Root cause.** The per-axis voxel scatter and the SDF smooth-yaw path projected screen position with the full `visualYaw` (`pos3DtoPos2DIsoYawed`) but wrote the framebuffer composite depth as the **un-yawed** world iso-depth `x+y+z`. Screen placement is yawed; depth ordering is not — so the two diverge as residual yaw grows (worst near the ±45° bracket), and a low/back surface beats geometry in front of it.

**Fix.** Re-derive the composite depth from the recovered world position rotated by `R_z(-visualYaw)` — `x(cosφ−sinφ) + y(sinφ+cosφ) + z`, the depth companion of `pos3DtoPos2DIsoYawed` — identically in `v_peraxis_scatter.glsl`, `peraxis_scatter.metal`, and the `c_shapes_to_trixel` smoothYaw path so SDF and voxels stay co-sorted. The stored face-local `rawDepth` (the origin-recovery key) is untouched; the `*4 + slot` encoding is preserved. Edits live only on the residual-yaw / smoothYaw paths, so the **cardinal fast path (residualYaw==0) is byte-identical**.

Inlined the yaw-rotation rather than adding an `ir_iso_common` helper on purpose — that header is included by every render shader and the codebase avoids touching it to protect cardinal byte-identity from FP-scheduling drift (see `ir_per_axis_lighting.glsl`).

## Before / after (IRShapeDebug `--spin-yaw`, Metal, zoom 8, ~230°)

| before (platform eats the cube) | after (cube in front) |
|---|---|
| before | after |

135° (other worst-case boundary), post-fix — cubes correctly in front of the platform:

after135

(The dark patch on the platform is the **separate** sun-shadow blackout — #1380 — not this change.)

## Test plan
- [x] Metal/macOS `IRShapeDebug --spin-yaw`: 135° and 230° — platform no longer occludes the voxel cube above it.
- [x] Cardinal fast path: static `yaw=0` frames byte-identical pre/post (only pre-existing FMA jitter differs).
- [ ] **Linux/OpenGL cross-host smoke** — the GLSL edits are unexecuted on macOS; needs a Linux run (`fleet:needs-linux-smoke`).
- [ ] `render-verify` baselines.

Closes #1430

Refs #1370 (the "clip" half of clip/cull/fade), #1380 (sun-shadow, separate). Likely also resolves the perceived "voxels clipping out" in #1431 (it was this depth inversion, not canvas sizing) — re-evaluate #1431 after this lands.


jakildev (commented):
## Review — render: yaw-consistent composite depth for smooth camera Z-yaw (PR #1434)

**Verdict:** approve

### Praise

- **Root cause diagnosis is accurate and well-documented.** The PR body clearly explains why `x+y+z` diverges from the on-screen projection as residual yaw grows — screen position uses `pos3DtoPos2DIsoYawed` (which projects `R_z(-φ)·world`) while the old depth used the un-yawed `x+y+z`, causing the ordering to split near ±45° bracket.
- **The rotation math is correct.** `R_z(-φ)·[x,y,z] = [x·cosφ+y·sinφ, -x·sinφ+y·cosφ, z]`, so iso-depth = `dvx + dvy + z` where `dvx = x·yc + y·ys` and `dvy = -x·ys + y·yc`. All three shader paths (`v_peraxis_scatter.glsl`, `peraxis_scatter.metal`, `c_shapes_to_trixel` smoothYaw in both GLSL and Metal) apply the identical formula — confirmed by reading each in full.
- **`rawDepth` / origin-recovery key left untouched.** The fix only changes the COMPOSITE depth written to the framebuffer; the `rawDepth >> 2` stored per-cell (the iso-inverse origin key) is unchanged. The `*4 + slot` encoding is preserved so the depth co-sorts with the SDF path.
- **Cardinal fast path guarded correctly.** Per-axis scatter is residual-only and not invoked at `residualYaw == 0`; the SDF smooth-yaw branch is gated by `if (smoothYaw)`. Cardinal byte-identity claim holds.
- **All four shader paths updated together** in one commit — no missed mirror.

### Nits

- `docs/design/per-axis-trixel-canvas-rotation.md` (depth-metric correction note): the parenthetical "(identical at cardinals, where `R_z(-visualYaw)` permutes `x+y+z`)" is imprecise. `R_z(-90°)·[x,y,z] = [y,-x,z]` gives depth `y-x+z`, which is not equal to `x+y+z` in general. The claim is TRUE in the sense that per-axis scatter doesn't run at cardinals (residualYaw=0), so the distinction is moot there — but the parenthetical reads like a mathematical equivalence that doesn't hold for non-identity cardinals. Suggest rewording to: "(at exactly 0° residual yaw the formula is byte-identical to the un-yawed metric; at other cardinals the per-axis scatter is inactive so the formula is never evaluated)."

### Test plan for author

- [x] Metal/macOS verified (before/after screenshots in PR body).
- [ ] Linux/OpenGL cross-host smoke (`fleet:needs-linux-smoke` already set).
- [ ] `render-verify` baselines after smoke passes.

Opus recheck required: touches `engine/render/src/shaders/` — the cross-shader depth-ordering co-sort invariant between `c_shapes_to_trixel` (SDF) and `v_peraxis_scatter` (voxel scatter) in the framebuffer composite pass warrants Opus verification. Math appears correct from the diff; Opus should confirm the framebuffer is shared, the depth test direction matches, and the `*4+slot` encoding co-sorts as claimed.

jakildev:
Addressed nit: fixed the parenthetical in the depth-metric correction note in `docs/design/per-axis-trixel-canvas-rotation.md`. The old text claimed `R_z(-visualYaw)` permutes `x+y+z` at cardinals (not true — at 90° cardinals the result is `y-x+z`, not `x+y+z`). Replaced with accurate language: at exactly 0° residual yaw the formula is byte-identical to the un-yawed metric; at other cardinals the per-axis scatter is inactive so the formula is never evaluated.

jakildev (commented):
## Opus re-review — render: yaw-consistent composite depth for smooth camera Z-yaw (#1434)

**Verdict:** approve (with the one doc nit Sonnet already flagged)

Sonnet approved-with-nits and flagged `Opus recheck required` on the cross-shader depth co-sort invariant: confirm the framebuffer is shared, the depth-test direction matches, and the `*4 + slot` encoding co-sorts the SDF (`c_shapes_to_trixel`) and voxel-scatter (`v_peraxis_scatter`) paths. I read all four shader bodies in full context plus `encodeDepthWithFace`, `pos3DtoPos2DIsoYawed`, and the framebuffer normalization. **Sonnet's verdict and math hold; on closer read every recheck surface confirms correct.**

### Confirmed — the co-sort invariant Sonnet flagged for recheck

- **`*4 + slot` encoding co-sorts.** `encodeDepthWithFace(rawDepth, face) = rawDepth*4 + face` (`ir_iso_common.glsl:123`). The SDF path stores `roundHalfUp(yawedDepth)*4 + face` via `encodeDepthWithFace(baseDepth, face)` (`c_shapes_to_trixel.glsl:1012`/`1024`); the scatter stores `roundHalfUp(yawedDepth)*4 + slot` directly (`v_peraxis_scatter.glsl:130`). Identical `*4` spacing, identical high-bits metric — the two co-sort by construction. Sonnet asserted this; I confirm it against the `encodeDepthWithFace` definition Sonnet did not cite.
- **Sign convention matches the screen projection exactly.** `pos3DtoPos2DIsoYawed` computes `vx = x·cosφ + y·sinφ`, `vy = -x·sinφ + y·cosφ` (= `R_z(-φ)·[x,y]`), then projects `(-vx+vy, -vx-vy+2z)`. The fix's composite depth `dvx + dvy + z` uses the **identical** `dvx=vx, dvy=vy`, so it is the iso-distance of the *same* rotated point the screen places — not a sign-flipped twin. This is the crux of the fix and it is right.
- **Framebuffer is shared, depth-test direction unchanged.** Both the SDF trixel→framebuffer pass (`f_trixel_to_framebuffer.glsl:45/57`) and the voxel scatter normalize with the same `(d + distanceOffset - kMin)/(kMax - kMin)`. The PR changes only the *value* of the iso-depth (un-yawed `x+y+z` → yawed `vx+vy+z`); it does not touch the encoding, the normalization, or the depth-test direction, so whatever ordering worked at cardinals still works. The before/after screenshots (platform no longer eats the cube at 230°) confirm the shared-framebuffer composite empirically.

### Also verified (beyond Sonnet's pass)

- **No clamp-overflow from the rotated metric.** `vx+vy = x(cosφ−sinφ) + y(cosφ+sinφ)`; over a box `[-R,R]²` its max is `R(|c−s|+|c+s|) ≤ 2R`, peaking at φ=0 and bottoming at `√2·R` at 45°. So the yawed depth magnitude is **always ≤** the un-yawed `x+y+z` envelope that `kMin/kMax` were sized for — rotation never expands the range, so no normalized-depth clamp/wrap artifact is introduced.
- **Cardinal fast path byte-identical.** SDF: yawed `baseDepth` is under `if (smoothYaw)`; the `else` keeps `surfaceD + originDistance` unchanged. Scatter: per-axis is residual-only (not dispatched at `residualYaw == 0`). Both cardinal paths are untouched.
- **`rawDepth` origin-recovery key untouched.** The scatter still recovers `origin` from the stored `rawDepth = rawDist >> 2`; only the *composite* depth is re-derived. No double-rotation — the SDF and voxel canvases feed their framebuffer depths independently, not through each other.
- **Metal mirrors match the GLSL** line-for-line (`peraxis_scatter.metal:106-110`, `c_shapes_to_trixel.metal:80-87`) — same `dvx/dvy` formula, same `roundHalfUp`, same `*4 + slot`. Reuses the already-computed `yawC/yawS` on the SDF path rather than recomputing. No missed backend.

### Nits (confirming Sonnet's — address in the same pass, not gating)

- `docs/design/per-axis-trixel-canvas-rotation.md` — the parenthetical "(identical at cardinals, where `R_z(-visualYaw)` permutes `x+y+z`)" reads like a mathematical equivalence that doesn't hold: `R_z(-90°)·[x,y,z]` gives depth `y−x+z ≠ x+y+z`. Sonnet's reword is correct — the metric is byte-identical only at *exactly* 0° residual yaw; at other cardinals the per-axis scatter is inactive so the formula is never evaluated. Take Sonnet's suggested wording.

### Test plan

- [x] Metal/macOS verified (before/after screenshots in PR body).
- [ ] Linux/OpenGL cross-host smoke (`fleet:needs-linux-smoke` already set) — the GLSL mirrors the verified Metal exactly, so this is a parity confirmation rather than a risk.
- [ ] `render-verify` baselines after smoke.

Clean, surgical fix with the math right on all four paths. Sonnet flagged the co-sort invariant; on closer read I confirm the `*4` encoding, the `R_z(-φ)` sign match, and the shared-framebuffer normalization all hold, plus the rotated metric stays inside the existing clamp envelope.

jakildev:
Nit verified already addressed: the parenthetical in `docs/design/per-axis-trixel-canvas-rotation.md` was corrected by @jakildev (see comment above). Cleared `fleet:has-nits`. `fleet:approved` retained — approval stands. — sonnet-fleet-1

PR #1408, opened 2026-05-31
render: per-entity SO(3) on main canvas — PR-A: transform plumbing + per-entity visible triplet (#1299)

jakildev:
Claiming issue. Work in progress.

Closes #1299

PR #1442, opened 2026-06-01
render: extract shared voxelDispatchGridForCount dispatch-grid helper

### LINKED ISSUE #1422: refactor: extract shared voxel dispatch-grid helper from UPDATE_VOXEL_POSITIONS_GPU and VOXEL_TO_TRIXEL_STAGE_1
## Context

`system_update_voxel_positions_gpu.hpp:252-257` inlines the same 2D workgroup-grid fold as `voxelDispatchGridForCount` in `system_voxel_to_trixel.hpp`. The existing comment on line 250 acknowledges the duplication:

> This is the same fold as `voxelDispatchGridForCount` in `system_voxel_to_trixel.hpp`; it is inlined here to avoid depending on that heavy STAGE_1 header just for the helper (a shared home for it is a worthwhile follow-up cleanup).

Surfaced as a nit during the fleet review of PR #1408 (per-entity SO3 on main canvas, issue #1299).

## Proposed fix

Extract the helper into a shared, lightweight header under `engine/prefabs/irreden/render/` (or `engine/render/include/irreden/render/`) with no dependency on the heavy STAGE_1 system header:

```cpp
// voxel_dispatch_grid.hpp  (new, header-only, no heavy deps)
namespace IRPrefab::Voxel {
    struct DispatchGrid { int groupsX, groupsY; };
    inline DispatchGrid voxelDispatchGrid(int voxelCount, int localSize = 64, int maxGroupsX = 1024) {
        const int groups = IRMath::divCeil(voxelCount, localSize);
        const int x = IRMath::min(groups, maxGroupsX);
        const int y = IRMath::divCeil(groups, x);
        return {x, y};
    }
}
```

Then:
- Replace the inlined block in `system_update_voxel_positions_gpu.hpp` with a call to the helper.
- Replace `voxelDispatchGridForCount` in `system_voxel_to_trixel.hpp` with a thin wrapper or a direct call.

## Area

`engine/prefabs/irreden/render/systems/`

## Model

sonnet

## Blocked by

(none)

jakildev:
## Summary

The 1D-count → 2D-workgroup-grid fold (`groupsX = min(count, 1024)`,
`groupsY = divCeil(count, groupsX)`) was inlined **verbatim** in three
render-prefab systems:

- `UPDATE_VOXEL_POSITIONS_GPU` (inlined local copy)
- `VOXEL_TO_TRIXEL_STAGE_1` (defined `IRSystem::voxelDispatchGridForCount` locally)
- `SHAPES_TO_TRIXEL` (`buildAndUploadTileDescriptors`, inlined local copy)

The update-positions system re-derived the fold locally rather than depend on
the heavy STAGE_1 header just for the helper — the duplication was flagged as a
nit in the PR #1408 review (#1299).

This moves the helper into a new lightweight header
`engine/prefabs/irreden/render/voxel_dispatch_grid.hpp` (depends only on
`ir_math.hpp`) and routes all three call sites through it.

## Scope note — third call site

The issue names two systems. The `simplify` reuse scan surfaced a **third**
verbatim copy of the identical fold in `SHAPES_TO_TRIXEL` (same directory),
so I folded it in here too — leaving an identical inline copy one file over
would defeat the dedup the issue exists for. `SHAPES_TO_TRIXEL` still derives
its `paddedCount` as `grid.x * grid.y`.

## Byte-identical

Pure extract-header refactor. The fold arithmetic and the 1024 column cap are
preserved exactly; the update site still applies it to `divCeil(liveCount, 64)`
and shapes still pads to `grid.x * grid.y`. No dispatch grid changes for any
count, so no visual delta is possible — screenshots intentionally skipped per
the author-pipeline mechanical-refactor exception.

## Test plan

- [x] `fleet-build --target IRShapeDebug` clean (compiles all three consumers
  via `IrredenEngineRendering`).
- [x] `format-changed` applied.
- [ ] Linux/OpenGL cross-host smoke (parity confirmation — no behavior change).

Closes #1422

### DIFF
--- a/engine/prefabs/irreden/render/systems/system_shapes_to_trixel.hpp
+++ b/engine/prefabs/irreden/render/systems/system_shapes_to_trixel.hpp
@@ -15,6 +15,7 @@
 #include <irreden/render/lod_utils.hpp>
 #include <irreden/render/sun_shadow_constants.hpp>
 #include <irreden/render/camera.hpp>
+#include <irreden/render/voxel_dispatch_grid.hpp>
 
 #include <irreden/render/gpu_stage_timing.hpp>
 #include <irreden/render/gpu_stage_timing_observer.hpp>
@@ -550,17 +551,15 @@ template <> struct System<SHAPES_TO_TRIXEL> {
             gridXOut = 1;
             return 0;
         }
-        constexpr int kMaxDispatchGroupsX = 1024;
-        const int gridX = IRMath::min(tileCount, kMaxDispatchGroupsX);
-        const int gridY = IRMath::divCeil(tileCount, gridX);
-        const int paddedCount = gridX * gridY;
+        const ivec2 grid = voxelDispatchGridForCount(tileCount);
+        const int paddedCount = grid.x * grid.y;
         ShapeTileDescriptor sentinel{};
         sentinel.shapeIndex = -1;
         while (static_cast<int>(tiles.size()) < paddedCount) {
             tiles.push_back(sentinel);
         }
         tileDescBuf->subData(0, paddedCount * sizeof(ShapeTileDescriptor), tiles.data());
-        gridXOut = gridX;
+        gridXOut = grid.x;
         return tileCount;
     }
 };
--- a/engine/prefabs/irreden/render/systems/system_update_voxel_positions_gpu.hpp
+++ b/engine/prefabs/irreden/render/systems/system_update_voxel_positions_gpu.hpp
@@ -48,6 +48,7 @@
 #include <irreden/ir_math.hpp>
 
 #include <irreden/render/voxel_pool_allocation.hpp>
+#include <irreden/render/voxel_dispatch_grid.hpp>
 #include <irreden/voxel/components/component_voxel_set.hpp>
 #include <irreden/voxel/components/component_voxel_pool.hpp>
 #include <irreden/common/components/component_world_transform.hpp>
@@ -153,11 +154,8 @@ template <> struct System<UPDATE_VOXEL_POSITIONS_GPU> {
         // CPU mirror of the GLSL/Metal `transform * localPos`: sqtToMat4 is
         // bit-identical to the shader-side helper, so the same operands classify
         // the same on both sides (the CPU↔GPU consistency the plan flags).
-        transforms_[slot].modelToWorld_ = IRMath::sqtToMat4(
-            worldTransform.scale_,
-            rotation,
-            worldTransform.translation_
-        );
+        transforms_[slot].modelToWorld_ =
+            IRMath::sqtToMat4(worldTransform.scale_, rotation, worldTransform.translation_);
         maxSlotUsed_ = IRMath::max(maxSlotUsed_, static_cast<int>(slot));
         anyDynamic_ = true;
 
@@ -184,7 +182,9 @@ template <> struct System<UPDATE_VOXEL_POSITIONS_GPU> {
             const std::uint32_t packed =
                 IRComponents::packVoxelVisibleTriplet(triplet[0], triplet[1], triplet[2]);
             lastTickPool_->setVoxelReservedForRange(
-                voxelSet.voxelStartIdx_, static_cast<std::size_t>(voxelSet.numVoxels_), packed
+                voxelSet.voxelStartIdx_,
+                static_cast<std::size_t>(voxelSet.numVoxels_),
+                packed
             );
         }
     }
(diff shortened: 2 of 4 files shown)


jakildev (commented):
## Review — render: extract shared voxelDispatchGridForCount dispatch-grid helper (PR #1442)

**Verdict:** approve

### Praise

- **Correct extract-helper scope.** The 1D-count → 2D-grid fold (`groupsX = min(count, 1024)`, `groupsY = divCeil(count, groupsX)`) was genuinely identical in all three call sites — not merely similar. Folding all three into a single header is the right refactor, not premature.
- **Dependency choice is sound.** `voxel_dispatch_grid.hpp` depends only on `ir_math.hpp`, keeping it lightweight enough that `UPDATE_VOXEL_POSITIONS_GPU` can include it without pulling in the heavy `VOXEL_TO_TRIXEL_STAGE_1` header — exactly the dependency concern the old `#1422` tracking comment cited.
- **Header uses explicit `IRMath::` qualifiers.** The standalone header correctly writes `IRMath::ivec2`, `IRMath::min`, `IRMath::divCeil` — not `glm::` or unqualified aliases. The consuming files alias these via `using namespace IRMath`, so the call sites resolve correctly.
- **Comment in the new header is load-bearing.** The explanation that the 1024 X-cap is a coupling contract between CPU dispatch and the shader's linear-index reconstruction (`workGroupIndex * localSize + localId.x`) is exactly the kind of non-obvious invariant that justifies a comment. The `count == 0` caveat is honest and correctly describes inherited behavior.

### ECS / render checklist

- No `getComponent` calls inside any tick path changed ✓
- No deferred structural mutations introduced ✓
- No allocations in hot paths ✓
- No shader changes — GLSL/Metal parity trivially preserved ✓
- No CPU/GPU struct changes — no alignment or binding-point concerns ✓
- Dispatch grid arithmetic is byte-identical: `SHAPES_TO_TRIXEL` passes `tileCount` directly (unchanged); `UPDATE_VOXEL_POSITIONS_GPU` passes `divCeil(liveCount, kLocalSize)` (the old local `groups` variable); `VOXEL_TO_TRIXEL_STAGE_1` removes its local definition and calls through the new header ✓
- `paddedCount = grid.x * grid.y` in `SHAPES_TO_TRIXEL` is equivalent to old `gridX * gridY` ✓
- `gridXOut = grid.x` is equivalent to old `gridXOut = gridX` ✓
- Namespace: new header is in `IRSystem`, all three consumers are inside `namespace IRSystem { }` with `using namespace IRMath` in scope ✓
- IRMath rule: no `glm::` or `std::` math calls in the new header ✓

### Test plan

- [x] `fleet-build --target IRShapeDebug` clean (all three consumers compile via `IrredenEngineRendering` per PR body).
- [x] `format-changed` applied.
- [ ] Linux/OpenGL cross-host build + smoke (no behavior change; parity confirmation).

Opus recheck not required.

jakildev (commented):
## Opus re-review — render: extract shared `voxelDispatchGridForCount` dispatch-grid helper (PR #1442)

**Verdict:** approve

Sonnet approved with "Opus recheck not required." I picked this up anyway because it touches the render compute-dispatch path (`engine/prefabs/irreden/render/` + `engine/system/`), and dispatch-grid folds are an Opus-tier render invariant — an off-by-one in the count→2D-grid map silently drops or double-processes voxels. **On a read of all three call sites against the extracted helper, I confirm Sonnet's verdict: this is a byte-identical extraction with no behavioral change.** The one thing Sonnet's checklist couldn't fully prove — that the documented `count > 0` precondition is actually upheld at every site — I verified directly below.

### The fold is identical at all three sites

The shared helper is `groupsX = min(count, 1024)`, `groupsY = divCeil(count, groupsX)`, returned as `ivec2`.

- **`SHAPES_TO_TRIXEL`** — old `gridX = min(tileCount, 1024)`, `gridY = divCeil(tileCount, gridX)`, `paddedCount = gridX*gridY`, `gridXOut = gridX`. New: `grid = helper(tileCount)`, `paddedCount = grid.x*grid.y`, `gridXOut = grid.x`. The `count` is the raw `tileCount` (not divided by a local size — this site pads a descriptor buffer rather than computing workgroups), exactly as before. ✓
- **`UPDATE_VOXEL_POSITIONS_GPU`** — old `groups = divCeil(liveCount, 64)`, then `min(groups, 1024)` / `divCeil(groups, groupsX)`. New: `helper(divCeil(liveCount, 64))`. The division-by-local-size happens at the call site (correct — the helper folds an already-reduced count), so `count == old groups`. Identical. ✓
- **`VOXEL_TO_TRIXEL_STAGE_1`** — the old inline `voxelDispatchGridForCount(int voxelCount)` body was byte-for-byte the new helper (only the param name changed `voxelCount`→`count`); both live in `namespace IRSystem`, so the existing call sites resolve to the header version transparently. ✓

### `count > 0` precondition is upheld (the divide-by-zero edge)

The helper's `divCeil(count, groupsX)` divides by zero when `count == 0` (`min(0,1024) == 0`) — the new header documents this as a caller precondition, and that's preserved-not-introduced (the old inline code had the identical hazard). I confirmed both compute-dispatch callers actually guard it:

- `UPDATE_VOXEL_POSITIONS_GPU` returns early at `liveCount <= 0` (`system_update_voxel_positions_gpu.hpp:201`), so `divCeil(liveCount, 64) ≥ 1` before the call.
- `SHAPES_TO_TRIXEL` early-returns (`gridXOut = 1; return 0;`) at `tileCount == 0` before reaching the fold.

So the documented precondition is real, not aspirational. ✓

### Header hygiene + IRMath rule

- New `voxel_dispatch_grid.hpp` uses fully-qualified `IRMath::ivec2` / `IRMath::min` / `IRMath::divCeil` and carries **no** `using namespace` directive — correct for a header. All three consumers have `using namespace IRMath` in scope, so their unqualified `ivec2 grid` resolves to `IRMath::ivec2`. ✓
- Dependency direction is right: the helper depends only on `ir_math.hpp`, which is exactly what lets `UPDATE_VOXEL_POSITIONS_GPU` drop the rejected "include the heavy STAGE_1 header for one helper" path the old `#1422` comment lamented. No `glm::`/`std::` math leaked in. ✓
- The non-functional reflows in `UPDATE_VOXEL_POSITIONS_GPU` (the `sqtToMat4` one-liner and the `setVoxelReservedForRange` arg split) are pure formatting — no semantic change.

### Cross-host

No shader or CPU/GPU struct change — Linux/OpenGL + Windows smoke is parity confirmation, not risk. Smoke labels already present (`fleet:needs-linux-smoke`, `fleet:needs-windows-smoke`; authored on macOS).

No nits worth a round-trip. Sonnet's pass was correct and complete; this confirms the dispatch-grid invariant it flagged as out-of-scope.

PR #1446, opened 2026-06-01
render: widget rendering broken on master — WIDGET_RENDER_* produces no output (#1436)

### LINKED ISSUE #1436: render: widget rendering broken on master — WIDGET_RENDER_* produces no output (IRUIWidgetsDemo renders black)
## Summary

On current `master`, the trixel UI **widget render path produces no visible output**. The canonical widget demo **`IRUIWidgetsDemo` renders a fully black screen** — no panels, borders, labels, buttons, lists. This breaks every widget-using creation (voxel editor, ui_widgets / ui_dockspace demos, and the game-repo gear browser).

`TEXT_TO_TRIXEL` (the separate glyph GPU-program path) still works — text rendered via `makeText` / `C_TextSegment` composites fine on the GUI canvas. Only the `WIDGET_RENDER_*` systems (which draw via `IRRender::fillRect` / `drawBorder` → `subImage2D` onto the GUI canvas) are invisible.

## Repro

```
cmake --preset macos-debug   # or your platform preset (no IRREDEN_USER_PROJECTS)
fleet-build --target IRUIWidgetsDemo
fleet-run --timeout 90 IRUIWidgetsDemo --auto-screenshot 30
```

Result: `save_files/screenshots/*.png` are entirely black. Expected: a large `TRIXEL UI WIDGETS` panel with buttons/sliders/checkboxes/list/dropdown/radio/text-input/scroll.

Observed on macOS (Metal). Cross-host check on Linux/OpenGL recommended.

## Diagnostic notes (narrowing)

- Widget entities exist and their archetypes match the render-system filters (verified via archetype-node logs: e.g. panel archetype carries `C_Widget + C_WidgetPanel + C_WidgetState + C_GuiPosition`, which is exactly `System<WIDGET_RENDER_PANEL>`'s filter).
- Theme colors are visible (`panelBackground_ = {24,26,34,220}`, `borderIdle_ = {96,108,132,255}`), so a panel border would be plainly visible on black if drawn.
- `WIDGET_RENDER_PANEL::beginTick` resolves `getCanvas("gui")` + `C_TriangleCanvasTextures` without crashing, so the GUI canvas exists and is the same one `TEXT_TO_TRIXEL` writes to.
- Distance semantics look correct on paper: the canvas clears the distance texture to `kTrixelDistanceMaxDistance` (65535, farthest) and color to transparent; widget layers write at `kTrixelDistanceMinDistance + 2..4` (≈ -65533, very near). Widget pixels should therefore win the depth contest against the cleared background — but nothing composites. Since text (`kGuiTextDistance = min+1`) *does* composite, the break is specific to the `fillRect`/`subImage2D`-written widget layers and/or how the framebuffer composite reads them.

## Likely culprit (please bisect to confirm)

Strong suspicion is the recent `trixel→framebuffer` depth-composite / per-axis-canvas rework landed this week:

- `db7ac786` — render: smooth camera Z-yaw — T3: trixel→framebuffer depth composite (#1310) (#1336)
- `d3534c94` — T1 per-axis trixel canvas infra (#1308)
- `26dd9776` / `#1399` — GPU voxel-position prepass
- `ea9196d7` / `#1408` — per-entity SO(3) on main canvas (PR-A)

The composite change is the leading suspect: the GUI canvas widget layers are CPU-written via `subImage2D` (`engine/render/include/irreden/render/trixel_rect.hpp`), and if the new depth composite (`system_trixel_to_framebuffer` / `c_*trixel_to_framebuffer` / `PerAxisScatterProgram`) reads the GUI canvas differently than it used to, those pixels stop reaching the framebuffer.

## Impact

High — silently breaks all widget UIs (editors + demos). No crash, just invisible widgets, so it can pass a smoke/exit-code check while being fully broken visually.

`[opus]` — render-pipeline composite debugging.

---
Found while resuming the game-repo gear browser (game #106): the browser builds, runs, and renders its catalog text via `TEXT_TO_TRIXEL`, but its widget panels/list/buttons are invisible for this same reason.

jakildev:
## Summary

Fixes the master regression where the trixel UI **widget render path produced no visible output** on macOS/Metal — `IRUIWidgetsDemo` (and every widget-using creation: voxel editor, ui_widgets/ui_dockspace demos, the game-repo gear browser) rendered a fully black screen, while `TEXT_TO_TRIXEL` glyph text composited fine.

### Root cause (Metal CPU/GPU texture-write ordering hazard)

On Metal, canvas-texture mutations split across two ordering domains:

- `Texture2D::clear()` enqueues a **GPU blit** on the frame's command buffer — *deferred*, executes at `commit()` time.
- `Texture2D::subImage2D()` did an **immediate CPU `replaceRegion`** to shared storage — happens during frame build, *before* the command buffer runs.

Each frame on the GUI canvas: `TEXT_TO_TRIXEL::beginTick` enqueues the clear blit and writes glyphs via a compute `imageStore` (GPU, encoded *after* the clear), then the `WIDGET_RENDER_*` systems draw panels/borders/labels via `fillRect`/`drawBorder` → `subImage2D`. The CPU widget writes landed immediately, then the queued clear blit ran at commit and **erased them** — while the GPU-written text survived. Hence "text visible, widgets black". OpenGL was unaffected (`glTextureSubImage2D` is submission-ordered).

### Fix

Route `uploadSubImage2D` through a command-buffer blit (staged from a transient shared buffer, `deferReleaseMetalBuffer`'d so it outlives the GPU read past the frame's `waitUntilCompleted`) when a frame command buffer is open. The upload now lands in encoder order *after* the clear/compute — matching OpenGL's submission-ordered `subImage2D`. Falls back to the direct `replaceRegion` only at startup, where no command buffer is open and there is no race. This also fixes any other CPU canvas write that races a same-frame clear (fog-of-war).

Single-file backend change (`engine/render/src/metal/metal_texture.cpp`); the OpenGL/Linux build does not compile it.

## Screenshots

<details>
<summary>widgets_idle</summary>

| Before (master) | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>widgets_after_settle</summary>

| Before (master) | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>widgets_followup</summary>

| Before (master) | After |
|--------|-------|
|  |  |

</details>

## Test plan

- [x] `IRUIWidgetsDemo` renders the full panel / buttons / sliders / checkboxes / list / dropdown / radio / text-input / scroll (0% → 86% non-black pixels).
- [x] `IRShapeDebug` voxel/SDF scene + lighting unchanged (no regression on the GPU-written render path).
- [x] `fleet-build --target IRUIWidgetsDemo` / `IRShapeDebug` clean.
- [ ] Linux/OpenGL cross-host build (file is Metal-only / not compiled there; parity confirmation only).

## Notes for reviewer

- **Perf:** the blit path allocates one transient staging buffer per `subImage2D` call (mirrors the existing buffer-orphan slow path in `metal_buffer.cpp`). This affects only CPU-write canvas paths (widgets, fog), never the voxel/SDF critical path. A per-frame staging arena (bump-allocated, reset at the clear) would eliminate steady-state allocation if a widget-heavy UI (gear browser) ever profiles hot — left as a measured follow-up rather than a speculative optimization.
- Engine gotcha documented in `engine/render/CLAUDE.md` so a future change to `clear()` or a new immediate CPU texture path doesn't reintroduce the race.

Closes #1436


jakildev (commented):
## Review — render: widget rendering broken on master — WIDGET_RENDER_* produces no output (engine PR #1446)

**Verdict:** approve

### Praise

- **Root cause analysis is correct and complete.** The race is real: `Texture2D::clear()` queues a GPU blit on the frame's command buffer (deferred to `commit()` time), while the old `uploadSubImage2D` did an immediate CPU `replaceRegion` on `StorageModeShared` storage. On Metal, shared-storage writes are not ordered against queued GPU encoders — the deferred clear blit executes at commit and clobbers any CPU write that preceded it in wall-clock time. OpenGL's `glTextureSubImage2D` is submission-ordered, which is why Linux was unaffected. The diagnosis is precise.
- **The fix is mechanically correct.** `metalDevice()->newBuffer(data, totalSize, StorageModeShared)` copies `data` into the staging buffer at allocation time (not a pointer capture — safe even if `data` is stack memory). `copyFromBuffer` then encodes the staging → texture blit *after* the clear encoder on the same command buffer, which is exactly the ordering needed. `deferReleaseMetalBuffer(staging)` hands the buffer to the established deferred-release queue (`pendingReleaseBuffers_`), which `releaseDeferredMetalBuffers()` drains only after `commandBuffer->waitUntilCompleted()` in `present()` — correct lifetime management.
- **Fallback path is correct.** When `metalCommandBuffer() == nullptr` (startup / one-off init), direct `replaceRegion` is safe because no GPU encoders are queued against the texture. The `clear()` startup fallback through `uploadSubImage2D` also works correctly since `uploadSubImage2D` itself now checks `commandBuffer == nullptr` and falls back the same way.
- **CLAUDE.md gotcha entry is well-written.** The invariant ("both subImage2D and clear() land on the command buffer in encoder order — reintroduce a CPU replaceRegion path and the #1436 invisible-widgets bug returns") is the kind of non-obvious constraint that belongs in the engine docs. The note about same-frame CPU `getBytes` readback still needing an explicit commit+wait is a good forward-looking callout.
- **Screenshots confirm the fix.** Before/after pairs show the full widget panel rendering after the fix (0% → 86% non-black pixels per PR body).

### Nits

- `engine/render/src/metal/metal_texture.cpp:376–396` (`MetalTexture3DImpl::uploadSubImage3D`) — the 3D upload path still uses direct `replaceRegion` with no command-buffer guard. 3D textures aren't currently cleared via GPU blit in the same frame as CPU writes (they're volume data seeded once), so there's no live race today. If a future system clears a 3D texture per-frame via the command buffer and then writes it via `uploadSubImage3D` in the same tick, the old bug reappears. Consider applying the same `metalCommandBuffer() == nullptr` guard + staging-blit path here as a precaution, or at minimum add a comment noting the unguarded path.

### Test plan verification

- [x] `IRUIWidgetsDemo` before/after screenshots confirm regression fixed.
- [x] `IRShapeDebug` listed as verified in PR body (GPU-written render path unchanged).
- [ ] Linux/OpenGL cross-host build — Metal-only file; parity confirmation only (build passes, no behavioral test needed).
- [ ] Windows cross-host build — same as Linux.

Opus recheck required: engine/render/ backend change involving GPU buffer lifetime (staging buffer across command buffer commit/waitUntilCompleted cycle) and Metal encoder-ordering invariant that spans the ECS tick pipeline — Sonnet cannot verify the encoder-sequencing contract holds across all WIDGET_RENDER_* tick paths without full cross-system trace.

jakildev:
Addressed nit: added a comment to `uploadSubImage3D` documenting why the direct `replaceRegion` path is safe today (3D textures are seeded once, never cleared via GPU blit in the same frame) and pointing to the `uploadSubImage2D` staging-blit pattern for the future case. No behavior change.

jakildev (commented):
## Opus re-review — render: widget rendering broken on master — WIDGET_RENDER_* produces no output (#1436) (PR #1446)

Sonnet approved and flagged "Opus recheck required" for the GPU-buffer-lifetime + Metal encoder-ordering invariant, noting it could not verify the encoder-sequencing contract holds across all `WIDGET_RENDER_*` tick paths without a full cross-system trace. I traced exactly that. **On a read of the command-buffer lifecycle three layers deep (tick-time `subImage2D` → frame command buffer → `present()` commit/wait/drain), I confirm Sonnet's verdict and its root-cause analysis.** The four things Sonnet could not prove without the trace — encoder *nesting* safety, staging-buffer lifetime across the commit/`waitUntilCompleted` boundary, copy-parameter correctness, and clear()-path parity — all check out, verified below.

### Encoder nesting — the one Metal hazard Sonnet couldn't trace (safe)

Metal allows only one open encoder per command buffer; opening a `blitCommandEncoder()` while a render/compute encoder is still live is a hard runtime fault. The new `subImage2D` path opens its own blit encoder and immediately `endEncoding()`s it (`metal_texture.cpp:393–411`) — **byte-for-byte the same self-contained open/close pattern as the already-shipped `clear()` path** (`metal_texture.cpp:227–242`). Both are invoked at tick time (widget systems write panels/borders/labels; `TEXT_TO_TRIXEL` queues the clear + compute glyph store), which is *between* render passes — no encoder is open at that point. Since `clear()` has shipped on master with this exact construction and the new write path runs at the same pipeline point, it inherits the same "no encoder open at canvas-write time" invariant. No nesting violation. ✓

### Staging-buffer lifetime across commit/wait (correct)

`deferReleaseMetalBuffer(staging)` pushes onto `pendingReleaseBuffers_`, and `releaseDeferredMetalBuffers()` is the only drain. I checked all three call sites (`metal_render_impl.cpp:291, 336, 696`) — every one runs strictly *after* `commandBuffer->commit()` + `waitUntilCompleted()`. So the GPU has provably finished reading the staging buffer before it is released. `newBuffer(data, size, StorageModeShared)` copies `data` at allocation time (not a pointer capture), so the source is safe even if it was a stack/transient buffer. ✓

### Copy parameters (correct)

`copyFromBuffer(staging, 0, bytesPerRow, totalSize, Size(width,height,1), m_texture, 0, 0, Origin(xoffset,yoffset,0))` — `sourceBytesPerRow = width*pixelSize` matches the tightly-packed subregion staging layout (subregion width, not full-texture width), `sourceBytesPerImage = totalSize` is a valid upper bound for a single 2D slice, and the destination origin/size address the intended subregion. The same `copyFromBuffer`-to-this-texture path is exercised by `clear()`, so any per-format row-alignment requirement is already satisfied for these canvas textures. ✓

### Fallback path (correct)

`metalCommandBuffer() == nullptr` (startup / one-off init) → direct `replaceRegion`. Safe because no GPU encoders are queued against the texture yet, and it skips the transient staging allocation. ✓

### Sonnet's nit (confirmed, correctly left as a nit)

`MetalTexture3DImpl::uploadSubImage3D` (`metal_texture.cpp:~342`) still does an unguarded direct `replaceRegion` with no command-buffer staging path. I confirm Sonnet: there is **no live race today** — 3D volume textures are seeded once, not GPU-cleared in the same frame as a CPU write — so this is correctly a `fleet:has-nits` item, not a `needs-fix`. A one-line comment marking the path as "intentionally immediate; reintroduce the staging guard if a per-frame 3D clear is ever added" would close it cheaply.

### Additional nit (observation, non-blocking)

Each `subImage2D` call now allocates a fresh `MTL::Buffer` via `newBuffer` and defers its release. For the widget systems this is bounded (a handful of panels/labels per frame) and fog-of-war is dirty-gated, so it's a non-issue at current call frequency. If a future system drives high-rate per-frame `subImage2D` on Metal, this becomes per-call allocation churn on the hot path — at that point a reusable per-texture staging buffer (mirroring `m_clearSourceBuf`'s change-detected reuse in `clear()`) would be the move. Worth a forward-looking comment but not required now.

### Cross-host

Metal-only file. `fleet:needs-linux-smoke` / `fleet:needs-windows-smoke` already present from Sonnet's pass — parity confirmation only; no behavioral change on OpenGL (`glTextureSubImage2D` is already submission-ordered, which is why Linux never saw #1436).

Sonnet's analysis was precise and complete. This pass confirms the encoder-sequencing contract it flagged as out-of-scope: the new write path lands on the command buffer in encoder order, the staging buffer's lifetime is correct across the commit/wait cycle, and the path is structurally identical to the shipped `clear()` blit.

jakildev:
Addressed nits from Opus review:
- Added forward-looking comment to `uploadSubImage2D` noting the per-call transient `MTL::Buffer` allocation is bounded at current widget-system call frequency, and pointing toward `m_clearSourceBuf`'s change-detected reuse pattern as the optimization path if high-rate per-frame `subImage2D` lands. (uploadSubImage3D comment was already committed by the human in the prior iteration.)

No behavioral change. `fleet:approved` remains valid.

Issue #1439, opened 2026-06-01
render: closed-form yaw-aware cull region in IRMath + rewire chunk/GPU cull (P2)


jakildev:
**Model:** opus
**Part of epic:** #1437
**Plan file:** `~/.fleet/plans/issue-1437.md` (full epic plan)
**Blocked by:** #1438
**Stack:** phase 2 of 4 — base PR on P1's branch (#1438).

## Scope
Replace the per-voxel/per-frame cull re-projection with a **closed-form O(chunks)** yaw-aware cull region, exposed as IRMath primitives.

## Approach
- Add IRMath rotation-coverage helpers (depth-companion style to `pos3DtoPos2DIsoYawed`): `isoAABBOfWorldAABBUnderYaw(vec3 worldMin, vec3 worldMax, float visualYaw)` (project the 8 world-AABB corners through the yawed iso projection, return the iso-AABB) and/or `visibleIsoViewportYawed(...)`.
- Rewire `C_VoxelPool::rebuildChunkBounds` to project each chunk's **static world-AABB** (8 corners) under the live yaw instead of iterating every voxel every frame.
- Keep the GPU `cullIsoMin/Max` derivation consistent with the same closed-form region.
- Stay **conservative** (never cull on-screen content); keep the `residualYaw==0` cardinal fast path **byte-identical** (yaw=0 collapses to today's path); GLSL/Metal parity for any shader-side change.

## Acceptance criteria
- Cull is O(chunks)/frame under rotation (no whole-pool re-projection).
- P1 harness shows nothing on-screen dropped across the yaw+move sweep.
- Cardinal frames byte-identical; perf improvement measured/noted.

## Dependencies
Blocked by P1 (#1438, harness). Blocks P3, P4.

## References
Epic #1437; PR #1434 (the depth-companion yaw fix); `ir_math.hpp:970` `visibleIsoViewport`; `component_voxel_pool.hpp` `rebuildChunkBounds`; `c_voxel_visibility_compact.glsl`.

Issue #1437, opened 2026-06-01
render: camera-yaw cull region + IRMath rotation library + camera-relative movement (epic)


jakildev:
## Epic: camera-yaw cull as a closed-form calculation + IRMath rotation library + camera-relative movement + provable validation

Under continuous camera Z-yaw the render cull is **already yaw-aware** — there is **no known correctness bug**:
- GPU `c_voxel_visibility_compact.glsl:79` projects each voxel via `pos3DtoPos2DIsoYawed(voxelPosRaw, visualYaw)` (+margin) before testing `cullIsoMin/Max`.
- CPU `C_VoxelPool::rebuildChunkBounds` projects per-voxel via `pos3DtoPos2DIsoYawed` when rotating to build the chunk iso-AABBs tested against the screen rect.

The gaps this epic closes:
1. **Brute-force, not closed-form.** When rotating, the CPU re-projects *every voxel every frame* (`rebuildChunkBounds` iterates the whole pool; `m_chunkBoundsDirty` forced true each yawing frame) and the GPU re-projects every voxel — O(voxels)/frame, no closed-form region.
2. **Yaw-blind core primitive.** `IRMath::visibleIsoViewport` (`ir_math.hpp:970`) is an axis-aligned iso box that ignores yaw; the yaw-awareness is bolted on downstream. The rotation-coverage math should be first-class IRMath primitives.
3. **World-axis-aligned movement.** `system_camera_mouse_pan` / `system_camera_key_drag_pan` apply the iso delta with no yaw rotation — not view-relative.
4. **No committed, provable validation** that nothing on-screen is ever culled under yaw. The pause-cull (`IRCommand::Command<TOGGLE_CULLING_FREEZE>` + `CullViewportState::frozen_`) already exists as engine functionality to build on.

**Goal:** make the cull a closed-form O(chunks) calculation, consolidate the rotation math into IRMath (reused by camera-relative movement), and land a provable engine-side validation harness.

**Not #1413** (per-entity SO(3) occlusion *depth*) — this is the **camera** cull region. Sibling of the smooth-camera-Z-yaw work (#1307 family) and PR #1434 (which fixed the analogous composite-*depth*-under-yaw bug — the depth companion of this cull math).

**Invariants (all phases):** `residualYaw==0` cardinal fast path byte-identical; GLSL/Metal parity; cull stays conservative (never drop on-screen).

[...]

PR #1447, opened 2026-06-01
render: frozen-cull free-fly validation harness (#1438)

### LINKED ISSUE #1438: render: frozen-cull free-fly validation harness for camera-yaw culling (P1)
**Model:** opus
**Part of epic:** #1437
**Plan file:** `~/.fleet/plans/issue-1437.md` (full epic plan)
**Stack:** phase 1 of 4 — base PR on `master`.

## Scope
Make "freeze the cull, then fly the camera (pan + yaw) freely" first-class **engine** functionality usable from an engine demo, and use it to **characterize + confirm** the current yaw-aware cull never drops on-screen content. The pause-cull already exists (`IRCommand::Command<TOGGLE_CULLING_FREEZE>` + `CullViewportState::frozen_`); this phase exposes a free-fly camera path and a demo-driven frozen-cull yaw+move sweep.

## Approach
- Confirm/expose a free-fly camera control (pan + yaw + zoom) usable while the cull is frozen; keep it engine-side (no game-repo dependency).
- Extend an existing engine demo (IRCanvasStress or shape_debug) or add a small cull demo with `--auto-screenshot` shots that freeze the cull at a reference pose, then sweep camera yaw + position.
- Compare frozen-cull-but-camera-moved frames against an un-frozen reference: the drawn/on-screen set must be fully retained (nothing that is actually on-screen vanishes; nothing guaranteed-offscreen is drawn).
- Write up findings: is the current cull tight or loose, any surprises, the margin behavior.

## Acceptance criteria
- A documented, repeatable **engine-side** procedure that, with the cull frozen, sweeps camera yaw+pos and shows the on-screen set is fully retained.
- Findings written up (baseline correctness + tightness characterization).
- Establishes the harness P2 validates against. Engine-only; no game-repo dependency.

## Dependencies
Blocks P2 (the closed-form cull rewrite is validated against this harness).

## References
Epic #1437; `command_toggle_culling_freeze.hpp`, `cull_viewport_state.hpp`, `system_voxel_to_trixel.hpp` (cull gate).

jakildev:
## Summary

Phase 1 of the camera-yaw cull epic (#1437): make "freeze the cull, then fly the camera freely" first-class engine functionality, and use it to **prove the current cull never drops on-screen content** under yaw + camera movement. Establishes the harness P2 (#1439) validates against.

- **`shape_debug --cull-validate`** — paired live/frozen camera sweep over one shared pose list (full yaw rotation + a ±12 pan sweep at 45°). Phase 1 captures live (cull tracks the camera); phase 2 freezes the cull at a deliberately **wide** reference (zoom 1 at origin, a superset of every sweep pose) and replays the same poses with the cull pinned. Diffing `cv_live_NNN` vs `cv_frozen_NNN` over the on-screen region answers the retention question. Disables sun shadows so the diff isolates voxel retention (see confound below).
- **Interactive F10** — binds `TOGGLE_CULLING_FREEZE` in `shape_debug`: freeze the cull at the current pose, then free-fly (WASD/mouse/scroll) to watch what the frozen cull retains.
- **`AutoScreenshotShot::cullAction_`** (NONE/FREEZE/UNFREEZE) — lets the auto-screenshot cycling system pin the cull mid-sweep without a keypress.
- **Freeze flag relocated** — moved from the toggle-command header into `IRRender` (`cull_viewport_state.hpp`), next to the cull state it gates, so `engine/video` can drive it without depending on the command module. Behavior-preserving; `IRCommand::isCullingFrozen()` stays as a thin forwarder.
- **Findings doc** — `docs/design/cull-validation-harness.md` (procedure + characterization).

## Key finding: 100% on-screen retention, no drops

Across the full yaw + pan sweep, **every voxel shape in the frozen ground-truth frame is also present in the live frame** — the cull is conservative and never drops on-screen content. With sun shadows off, cardinal yaws are byte-identical (yaw 90° is exactly 0 delta) and the worst case is 99.80% match at PSNR ~37–40.

The residual difference is **not** dropped geometry: the cull viewport also drives the sun-shadow-feeder AABB and the emissive light volume, so a frozen *wide* cull changes lighting. With shadows off, the entire remaining diff is a single emissive-lit sphere:

**Clean retention proof (shadows off) — the whole diff is one lighting sphere, not a shape silhouette:**

clean diff

**The shadow confound (shadows on) — why the harness disables them:**

shadow confound

**Live vs frozen at pan+x @ 45° — all shapes retained at identical positions:**

| live cull | frozen (wide) cull |
|---|---|
| live | frozen |

## Test plan

- [x] `fleet-build --target IRShapeDebug` clean (macOS / Metal).
- [x] `fleet-run IRShapeDebug --auto-screenshot --cull-validate` runs the 26-shot sweep and exits cleanly; FREEZE at shot 13, UNFREEZE at 26.
- [x] Pairwise diff of all 12 live/frozen pairs (shadows off): no shape silhouette in any diff → no dropped geometry.
- [x] Default scene byte-identical — `--cull-validate`, F10, and the freeze-flag move are all opt-in / behavior-preserving; `zoom4_origin` renders identically to before.
- [ ] Cross-host (Linux/Windows) smoke — header-only include change; cull behavior unchanged.

Closes #1438


jakildev (commented):
## Review — render: frozen-cull free-fly validation harness (#1438) (PR #1447)

**Verdict:** approve

### Praise

- **Circular include dependency correctly broken.** The old `cull_viewport_state.hpp` → `command_toggle_culling_freeze.hpp` → `cull_viewport_state.hpp` cycle is resolved cleanly: the freeze flag now lives entirely in `cull_viewport_state.hpp`, and `command_toggle_culling_freeze.hpp` depends on it (one-way). `IRCommand::isCullingFrozen()` is preserved as a thin forwarder for existing callers — a behavior-preserving refactor.
- **Freeze transition timing is sound.** `updateCullViewport`'s `else if (!s.frozen_)` branch captures the wide-reference snapshot on the FIRST render frame after `setCullingFrozen(true)` — that's the frame where the camera sits at zoom=1/origin before any frozen shot moves the camera away. The auto-screenshot settle mechanism (3 frames between camera-set and capture) gives `updateCullViewport` exactly the window it needs. The comment "only the first call in a frozen transition captures the snapshot" is accurately implemented.
- **Pointer-stability contract correctly honored.** `g_cullValidateShotLabels.reserve(totalShots)` + `g_cullValidateShots.reserve(totalShots)` are called before any `emplace_back`, preventing reallocation. The `shot.label_ = label.data()` pointers remain valid for the program's lifetime. Same discipline as the spin-yaw buffers.
- **`CullAction::NONE` default is backward-compatible.** All existing shot tables in `render-debug-loop` and `render-verify` compile and behave identically — no callers need to change.
- **Global state cleanup.** The UNFREEZE shot at the end leaves `isCullingFrozen() == false` so the harness doesn't bleed into subsequent runs or interactive sessions.

### Nits

- `engine/video/include/irreden/video/auto_screenshot.hpp:49–57` — The `///` doc block describing `AutoScreenshotShot` fields (lines 31–56) now immediately precedes `enum class CullAction` (line 57) rather than `struct AutoScreenshotShot` (line 63). Doxygen attaches the block to `CullAction`; the struct ends up without an attached doc block. Content is correct and the header reads clearly to a human, but tooling will attribute `@c cullAction_` field descriptions to the enum. Clean fix: end the existing `AutoScreenshotShot` doc block before the enum declaration, give `CullAction` its own one-line `///`, and move the `cullAction_` paragraph into a post-enum `///` block immediately before the struct.

### Test plan for author

- [x] `fleet-build --target IRShapeDebug` clean (macOS/Metal, per PR body).
- [x] `fleet-run IRShapeDebug --auto-screenshot --cull-validate` — 26-shot sweep, pairwise diffs confirm 100% on-screen retention (shadows off); shape silhouette absent from all diffs.
- [x] Default scene byte-identical — `--cull-validate` and F10 are opt-in; no behavior change to existing shots.
- [ ] Linux/OpenGL cross-host build + smoke — no shader changes; confirms cull-freeze flag timing is identical on the OpenGL backend.

Opus recheck not required.

jakildev:
Addressed nit: moved `CullAction` enum before its own one-line doc comment, and relocated the `cullAction_` field description to a separate `///` block immediately before `struct AutoScreenshotShot`. Doxygen now attaches each block to the correct declaration.

jakildev (commented):
## Opus re-review — render: frozen-cull free-fly validation harness (#1438) (PR #1447)

**Verdict:** approve

Sonnet approved with "Opus recheck not required" + one doc-comment nit (`fleet:has-nits`). I picked this up anyway because it touches `engine/prefabs/irreden/render/cull_viewport_state.hpp` — the shared cull-viewport state that gates both the CPU chunk-visibility mask and the GPU `cullIsoMin/Max` bounds — and a relocation of the freeze flag there is an Opus-tier render invariant: if it forked the flag into two statics, the interactive toggle and the harness would read different state. **On a read three stages deep (toggle command → `cull_viewport_state` → `updateCullViewport` snapshot timing, and the `shadowFeederCullViewport` derivation the doc cites), I confirm Sonnet's verdict.** The two things Sonnet's checklist couldn't fully prove — single-source-of-truth after the flag move, and that the default shot path is genuinely untouched — I verified directly below.

### Flag relocation is single-source-of-truth (the ODR / fork hazard)

The freeze flag moved from `IRCommand::detail::cullingFreezeFlag()` (in `command_toggle_culling_freeze.hpp`) to `IRRender::detail::cullingFreezeFlag()` (in `cull_viewport_state.hpp`), breaking the old circular include `cull_viewport_state.hpp → command_toggle_culling_freeze.hpp → cull_viewport_state.hpp`. I checked that the move is clean, not a fork:

- The old `IRCommand::detail::cullingFreezeFlag()` is **deleted**, not shadowed. `IRCommand::isCullingFrozen()` is preserved as a thin forwarder to `IRRender::isCullingFrozen()`, and `Command<TOGGLE_CULLING_FREEZE>` now writes `IRRender::setCullingFrozen(...)`. Both the interactive toggle and the harness's `setCullingFrozen` therefore hit the **one** `IRRender::detail` static. ✓
- Grepping the tree, the only pre-PR consumer of `IRCommand::isCullingFrozen()` was `cull_viewport_state.hpp` itself (the circular edge); the PR rewrites that call to the local `isCullingFrozen()`. The `isCullingFrozen` matches in `system_debug_culling_minimap.hpp` are a **local variable**, not the function — unaffected. No dangling reference, no second static. ✓
- The flag is an `inline` function-local static in a header → one instance across all TUs. Correct.

### Snapshot timing is pre-existing and unchanged; harness uses it correctly

`updateCullViewport`'s freeze-transition capture (`else if (!s.frozen_)` grabs the live pose on the first frozen frame, then `s.frozen_ = nowFrozen`) is untouched by this PR — the only edit to that function is the `IRCommand::` → local qualifier swap. The harness's FREEZE shot sets the flag **while the camera sits at the wide freeze-ref pose** (zoom 1, origin), in the same `settleCounter == 0` block that sets the camera; the snapshot is then taken on the next `updateCullViewport` within the 3 settle frames, at that ref pose. That's the correct pose, and the documented "wide reference is a superset of every sweep pose" makes the frozen frame a sound cull-disabled ground truth. ✓

### Default behavior is genuinely byte-identical

`AutoScreenshotShot::cullAction_` defaults to `CullAction::NONE`, and `auto_screenshot.cpp` only calls `setCullingFrozen` on `FREEZE`/`UNFREEZE` — `NONE` leaves the flag as-is. Every existing shot table (spin-yaw, rotation-coverage, cardinal) omits `cullAction_`, so the flag is never touched and the live render path is unchanged. The `--cull-validate` branch is fully opt-in and disables sun shadows only inside its own guard. ✓

### Allocation / stable-pointer discipline (init-time, not hot path)

The harness builds `g_cullValidateShots` / `g_cullValidateShotLabels` once in `initSystems`, both `reserve(totalShots)` where `totalShots = posesPerPhase*2 + 2 = 12*2+2 = 26` and the emit sequence is exactly 26 (`emitPhase` 12 + freeze-ref 1 + `emitPhase` 12 + unfreeze 1). Size reaches capacity without exceeding it → no reallocation → the `label.data()` pointers stashed in each shot stay valid for the game loop's lifetime. Same stable-storage contract as the spin-yaw buffers, and the reserve count is exact. ✓

### Sonnet's nit (confirmed, correctly left as a nit)

Sonnet's `### Nits` item — the `AutoScreenshotShot` `///` doc block runs straight into the `CullAction` enum, so doc tooling attributes the `cullAction_` paragraph to the enum — is real and worth the `fleet:has-nits` flag. It's a pure-documentation attribution issue with zero correctness impact; keeping it as a nit (not a needs-fix) is the right call. Author can split the block before merge.

### Cross-host

Render-path change, authored on macOS; `fleet:needs-linux-smoke` / `fleet:needs-windows-smoke` already present — parity confirmation, low risk (no shader or CPU/GPU struct change; the harness is opt-in and default scenes are byte-identical).

Sonnet's pass was correct and complete. This confirms the render-state invariant it flagged as out-of-scope: the flag relocation is a true single-source-of-truth move, not a fork.

jakildev:
Addressed fleet:has-nits:
- Moved `CullAction` enum before the `AutoScreenshotShot` doc block so doc tooling attributes each declaration's comment correctly. The `cullAction_` paragraph is now merged into the struct's doc block directly above the struct. No code changes.

PR #1450, opened 2026-06-01
render: closed-form yaw-aware chunk cull region in IRMath (#1439)

### LINKED ISSUE #1439: render: closed-form yaw-aware cull region in IRMath + rewire chunk/GPU cull (P2)

jakildev:
## Summary
- P2 of the camera-yaw cull epic (#1437). Replaces the per-voxel, per-frame iso re-projection in the **continuous-yaw branch** of `C_VoxelPool::rebuildChunkBounds` with a **closed-form O(chunks)** cull region.
- New `IRMath::isoAABBOfWorldAABBUnderYaw(worldMin, worldMax, visualYaw)` — projects the 8 world-AABB corners through `pos3DtoPos2DIsoYawed`. Since that projection is **linear**, the result is a *conservative superset* of the per-voxel iso bounds, so the chunk-visibility mask can only over-include; the GPU's exact per-voxel test stays the final arbiter and **render output is unchanged**.
- `C_VoxelPool` caches a per-chunk static **world-space AABB** (`m_chunkWorldBounds`), rebuilt only when voxel positions change. The two in-place position writers (`REBUILD_GRID_VOXELS`, `UPDATE_VOXEL_SET_CHILDREN`) call the new `markChunkWorldBoundsDirty()` so the superset guarantee holds under movement/rotation.
- **Cardinal / yaw==0 path is untouched → byte-identical** (only the continuous-yaw branch changed; at yaw 0 the cardinal branch runs).
- `entityIsoBounds` refactored to the un-yawed, axis-aligned special case of the new general primitive (removes a duplicated 8-corner enumeration).

## Perf
- Static world under a rotating camera: per-frame cull projection drops from O(voxels) to O(chunks) — ~**32×** fewer projections (`kVoxelChunkSize`=256 voxels/chunk ÷ 8 corners). The O(voxels) world-AABB rebuild is amortized to ~0 (only on position change).
- `--spin-yaw 45 --auto-profile 300` on IRShapeDebug: 12.92 ms avg frame, no new CPU hotspot (`RebuildGridVoxels` 0.001 ms, `UpdateVoxelSetChildren` 0.003 ms).

## Test plan
- [x] New `IsoAABBUnderYawTest` suite (yaw=0 reduction, π/2 case, degenerate point, **conservative-superset containment**) — passes.
- [x] Full unit suite: 1057 tests pass.
- [x] `--cull-validate` frozen-cull free-fly harness (#1438) runs clean through a full yaw+move sweep; rotated-yaw frames render correctly (no dropped geometry).
- [x] `entityIsoBounds` delegation verified byte-identical (cos/sin(0) exact) — `IsEntityOnScreen` tests pass.
- No shader / frame-data-struct change → GLSL/Metal parity preserved trivially.

Closes #1439

### DIFF
--- a/engine/math/include/irreden/ir_math.hpp
+++ b/engine/math/include/irreden/ir_math.hpp
@@ -1017,18 +1017,31 @@ inline vec2 shapeIsoHalfExtent(vec3 voxelSize) {
     return vec2(extentX + extentY, extentX + extentY + 2.0f * extentZ);
 }
 
-/// Returns the tight iso-space axis-aligned bounding box of a rectangular
-/// prism entity at @p worldPos with @p voxelSize dimensions.  Enumerates
-/// all 8 corners for an exact result (prefer shapeIsoHalfExtent when a
-/// conservative approximation is acceptable).
-inline IsoBounds2D entityIsoBounds(vec3 worldPos, ivec3 voxelSize) {
+/// Returns the iso-space AABB of the world-space box [@p worldMin, @p worldMax]
+/// projected under a continuous camera Z-yaw of @p visualYaw. Enumerates the 8
+/// world corners through @ref pos3DtoPos2DIsoYawed and bounds the result.
+///
+/// Closed-form O(1) companion to per-voxel cull projection (#1439): a chunk
+/// caches its static world-AABB once, and its cull region under the live yaw is
+/// recovered by projecting 8 corners instead of re-projecting every voxel each
+/// frame. `pos3DtoPos2DIsoYawed` is linear, so the iso-AABB of the projected
+/// corners is the tight iso-AABB of the projected box — and for a voxel set
+/// that does not fill its world-AABB it is a *conservative superset* of the
+/// per-voxel iso bounds (never tighter), so the chunk-visibility gate keeps
+/// over-including rather than dropping on-screen geometry. At @p visualYaw == 0
+/// it reduces to the un-yawed iso projection of the box corners.
+inline IsoBounds2D isoAABBOfWorldAABBUnderYaw(vec3 worldMin, vec3 worldMax, float visualYaw) {
     vec2 corners[8];
     int idx = 0;
     for (int dx = 0; dx <= 1; ++dx) {
         for (int dy = 0; dy <= 1; ++dy) {
             for (int dz = 0; dz <= 1; ++dz) {
-                vec3 corner = worldPos + vec3(dx * voxelSize.x, dy * voxelSize.y, dz * voxelSize.z);
-                corners[idx++] = pos3DtoPos2DIso(corner);
+                const vec3 corner = vec3(
+                    dx ? worldMax.x : worldMin.x,
+                    dy ? worldMax.y : worldMin.y,
+                    dz ? worldMax.z : worldMin.z
+                );
+                corners[idx++] = pos3DtoPos2DIsoYawed(corner, visualYaw);
             }
         }
     }
@@ -1039,6 +1052,15 @@ inline IsoBounds2D entityIsoBounds(vec3 worldPos, ivec3 voxelSize) {
         bmax = glm::max(bmax, corners[i]);
     }
     return IsoBounds2D{bmin, bmax};
+}
+
+/// Returns the tight iso-space axis-aligned bounding box of a rectangular
+/// prism entity at @p worldPos with @p voxelSize dimensions.  The un-yawed,
+/// axis-aligned special case of @ref isoAABBOfWorldAABBUnderYaw over the box
+/// [worldPos, worldPos + voxelSize] (prefer shapeIsoHalfExtent when a
+/// conservative approximation is acceptable).
+inline IsoBounds2D entityIsoBounds(vec3 worldPos, ivec3 voxelSize) {
+    return isoAABBOfWorldAABBUnderYaw(worldPos, worldPos + vec3(voxelSize), 0.0f);
 }
 
 /// Returns true if a rectangular-prism entity's iso bounding box overlaps
--- a/engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp
@@ -39,6 +39,26 @@ struct ChunkBounds {
     }
 };
 
+// Static (camera-independent) world-space AABB of a chunk's live voxels.
+// Cached once per allocation/position change so the continuous-yaw cull can
+// recover a chunk's iso bounds by projecting 8 corners under the live yaw
+// (`IRMath::isoAABBOfWorldAABBUnderYaw`) — O(chunks)/frame — instead of
+// re-projecting every voxel every frame (#1439).
+struct ChunkWorldBounds {
+    vec3 worldMin_ = vec3(std::numeric_limits<float>::max());
+    vec3 worldMax_ = vec3(std::numeric_limits<float>::lowest());
+
+    void expand(vec3 worldPos) {
+        worldMin_ = IRMath::min(worldMin_, worldPos);
+        worldMax_ = IRMath::max(worldMax_, worldPos);
+    }
+
+    // A chunk with no live voxels keeps the inverted sentinel.
+    bool empty() const {
+        return worldMin_.x > worldMax_.x;
+    }
+};
+
 struct C_VoxelPool {
   public:
     C_VoxelPool(ivec3 numVoxels)
@@ -103,6 +123,7 @@ struct C_VoxelPool {
                 m_freeSpanLookup.erase(size);
             }
             m_chunkBoundsDirty = true;
+            m_chunkWorldBoundsDirty = true;
             return IRRender::VoxelPoolAllocation{
                 startIndex,
                 std::span<IRRender::VoxelGpuPosition>{m_voxelPositions.data() + startIndex, size},
@@ -119,6 +140,7 @@ struct C_VoxelPool {
             size_t startIndex = static_cast<size_t>(m_voxelPoolIndex);
             m_voxelPoolIndex += size;
             m_chunkBoundsDirty = true;
+            m_chunkWorldBoundsDirty = true;
             IRE_LOG_DEBUG("Allocated voxels from {} to {}", startIndex, m_voxelPoolIndex - 1);
             return IRRender::VoxelPoolAllocation{
                 startIndex,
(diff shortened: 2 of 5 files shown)


jakildev (commented):
## Review — render: closed-form yaw-aware chunk cull region in IRMath (#1439) (engine PR #1450)

**Verdict:** approve

### Praise

- **Math is sound and well-tested.** `isoAABBOfWorldAABBUnderYaw` correctly exploits linearity of `pos3DtoPos2DIsoYawed` — the projection is linear in `p` with `yaw` as a fixed parameter, so the AABB of the 8-corner projections is the tight AABB of the projected box. For a voxel set whose live voxels don't fill the chunk's world-AABB, the 8-corner bound is conservative (over-includes), which is exactly the cull safety property needed. `ContainsInteriorPointProjection` directly unit-tests this at an arbitrary yaw — the right test to have.
- **`entityIsoBounds` delegation is correct.** `pos3DtoPos2DIsoYawed(p, 0.0f)` with `cos(0)=1` and `sin(0)=0` reduces exactly to `pos3DtoPos2DIso(p)`. IEEE 754 guarantees `cos(0.0f) == 1.0f` exactly, so the delegation is bit-identical to the pre-PR implementation. The `IsEntityOnScreen` tests verify this.
- **Cache eviction coverage is complete.** `m_chunkWorldBoundsDirty` is set on `allocateVoxels` (both code paths), `deallocateVoxels`, and on the two in-place position writers (`REBUILD_GRID_VOXELS` after `safeCount > 0` and `UPDATE_VOXEL_SET_CHILDREN` after `writtenCount > 0`). The conditional guards (`safeCount > 0`, `writtenCount > 0`) are correct — no-op writes don't evict the cache.
- **Cardinal path untouched.** The `useContinuousYaw == false` branch is byte-identical to master, so the yaw==0 cardinal render path is unaffected. The early-return guard `!useContinuousYaw && !m_chunkBoundsDirty && cardinalIndex == m_lastBoundsCardinalIndex` is preserved and works correctly: the continuous-yaw path forces `m_chunkBoundsDirty = true` on exit, so the first subsequent cardinal call always rebuilds (correct — cardinal cannot trust stale continuous bounds).
- **`ChunkWorldBounds::expand` correctly uses `IRMath::min`/`max`**, matching the project math rule. The existing `ChunkBounds::expand` pre-dates the rule; the new struct sets the right precedent.
- **O(voxels) cost properly amortized.** `ensureChunkWorldBounds` re-iterates the pool only when `m_chunkWorldBoundsDirty` or the chunk count changes; a static world under a rotating camera pays nothing after the first rebuild. The perf numbers in the PR body confirm this (`RebuildGridVoxels` 0.001 ms steady-state).
- **Test suite.** `ZeroYawMatchesUnYawedProjection`, `DegeneratePointBox`, `QuarterTurnYaw`, `ContainsInteriorPointProjection` — all four cases are mathematically verified and their expected values are correct per `pos3DtoPos2DIsoYawed` arithmetic.

### Test plan

- [x] New `IsoAABBUnderYawTest` suite passes — per PR body.
- [x] Full 1057-test suite passes — per PR body.
- [x] `--cull-validate` frozen-cull free-fly harness (#1438) clean through full yaw+move sweep — per PR body.
- [x] `entityIsoBounds` byte-identical delegation confirmed (`IsEntityOnScreen` tests pass) — per PR body.
- [x] `--spin-yaw 45 --auto-profile 300` shows no new CPU hotspot — per PR body.
- [ ] Linux/OpenGL build: pure C++ change, no shaders — confirm compile-clean on the next Linux smoke iteration (no cross-host render behavior difference expected).

Opus recheck required: `isoAABBOfWorldAABBUnderYaw` is a new public `ir_math.hpp` primitive whose conservative-superset invariant is the load-bearing cull safety guarantee. Sonnet confirmed the math and the eviction sites visible in the diff, but cannot tree-wide verify that every position-writing path that bypasses `allocateVoxels`/`deallocateVoxels` and the two systems correctly calls `markChunkWorldBoundsDirty` — a missed call would silently project stale world-AABBs and drop on-screen geometry.

jakildev (commented):
## Opus re-review — render: closed-form yaw-aware chunk cull region in IRMath (#1439) (engine PR #1450)

**Verdict:** approve

Sonnet approved and flagged "Opus recheck required" on the one thing a diff-only read cannot close: `isoAABBOfWorldAABBUnderYaw` is the load-bearing cull-safety primitive, and a position-writing path that skips `markChunkWorldBoundsDirty` would silently project a stale world-AABB and drop on-screen geometry. **I traced exactly that — the conservative-superset invariant and the tree-wide eviction coverage — and confirm Sonnet's verdict and its math/eviction analysis. Both hold.**

### Conservative-superset invariant — the core safety property (holds)

Sonnet proved the 8-corner AABB is the *tight* AABB of the projected box from linearity. The stronger property the cull actually depends on is that the cached path **never drops a chunk master would have kept**. It holds for two reasons I verified directly:

- `pos3DtoPos2DIsoYawed` is **affine** in `p`. The extreme iso coordinates of an affine image of a box are always attained at the 8 vertices, so the corner-AABB contains the projection of *every* interior point — including every live voxel. `ContainsInteriorPointProjection` unit-tests precisely this at an arbitrary yaw (0.7 rad) over three interior points.
- The cache is built from **the same `m_voxelPositionsGlobal[i].pos_` source** that master's per-voxel continuous-yaw loop iterated (confirmed against `origin/master` `rebuildChunkBounds`). So the new bounds are a superset of master's per-voxel iso bounds, sourced identically — strictly more conservative, never tighter. No geometry that master rendered can be culled.

### Eviction coverage — the tree-wide check Sonnet could not do (complete)

The cache is invalidated iff `m_voxelPositionsGlobal` mutates. I enumerated every writer:

- **Allocation (both spans) + deallocation** set `m_chunkWorldBoundsDirty = true` inline — verified in the diff.
- **In-place position rewrites** can only happen through the **non-const** `getPositionGlobals()` accessor. Grepped tree-wide: it has exactly **two** callers — `REBUILD_GRID_VOXELS` and `UPDATE_VOXEL_SET_CHILDREN` — and both now call `markChunkWorldBoundsDirty()`. The other three callers (`system_voxel_to_trixel.hpp` ×2, `system_build_light_occlusion_grid.hpp`) bind the **`const`** overload — read-only, no eviction needed.

The two guards are also exact, not just present:
- `REBUILD_GRID_VOXELS` — the only `poolGlobals` write is the `for (i < safeCount)` loop, and the eviction is gated on `safeCount > 0`. Bracket is tight. ✓
- `UPDATE_VOXEL_SET_CHILDREN` — all position writes happen inside `voxelSet.updateAsChild(...)`, whose return value *is* `writtenCount`; eviction gated on `writtenCount > 0`, and `writtenCount == 0` means "parent unchanged → no write." Tight. ✓

So the eviction set = the position-writer set, with no gaps. That closes the exact hole Sonnet flagged.

### GPU-transform-indirected sets (#1396) — no regression, pre-existing behavior

For a `gpuTransformSlot_ != kVoxelTransformStatic` set, `UPDATE_VOXEL_SET_CHILDREN` recomputes only a *translation-only* CPU mirror while the real SO(3) rotation lands on GPU binding 5. The cached world-AABB therefore reflects the un-rotated box. This is **identical to master**: master's per-voxel continuous-yaw loop also read the translation-only CPU mirror, so the cull was already non-rotation-aware for GPU-indirected sets. This PR caches the same data — it neither introduces nor worsens the gap. Worth a forward-looking note for whoever finishes per-entity SO(3) on the main canvas (the GPU-rotated extent can exceed the CPU mirror's AABB), but it is out of scope here and not a blocker for #1439.

### Other checks

- `ChunkWorldBounds::empty()` keeps the inverted sentinel for all-transparent chunks; the continuous path `continue`s past them, leaving `m_chunkBounds[c]` at its `reset()` value → never visible, matching master's "no voxel ever expands an empty chunk" behavior. ✓
- `ensureChunkWorldBounds` rebuild is bounded — max `chunk = (m_voxelPoolIndex-1)/kVoxelChunkSize < chunkCount`, and the chunk-count-change branch re-syncs on alloc/dealloc (which also set the dirty flag). No OOB, no stale-size read. ✓
- `entityIsoBounds` delegation is bit-identical at yaw 0 (IEEE-754 `cos(0.0f)==1.0f`, `sin(0.0f)==0.0f`), and `ChunkWorldBounds::expand` uses `IRMath::min/max` per the math rule. ✓
- Cardinal path is byte-identical to master; the continuous block forces `m_chunkBoundsDirty = true` on exit so the next cardinal call cannot trust stale continuous bounds. ✓

### Test plan

- [x] `IsoAABBUnderYawTest` (zero-yaw, degenerate-point, quarter-turn, interior-containment) — math verified by hand, all four correct.
- [x] Full suite + `--cull-validate` frozen-cull sweep + `--spin-yaw 45 --auto-profile 300` — per PR body.
- [ ] Linux/OpenGL build — pure C++/header change, no shaders; compile-clean confirmation on next Linux smoke.

Sonnet's math and eviction-site analysis were accurate. My pass adds the affine-superset argument, the tree-wide accessor sweep that proves the eviction set is complete, and the #1396 no-regression confirmation — the three things the diff alone leaves open. Safe to merge.

PR #1452, opened 2026-06-01
render: shapes clipped out of view under camera Z-yaw — per-axis canvas sizing (#1431)

### LINKED ISSUE #1431: render: shapes clipped out of view under camera Z-yaw — per-axis canvas sizing (split from #1370)

jakildev:

Closes #1431


jakildev:
## NEEDS-DESIGN

Root-caused #1431. The clip is real and reproducible, but the fix is an architectural tradeoff on top of the #1310/#1311 forward-scatter store that I don't have authority to pick. Full diagnosis + a dramatic repro + the candidate fixes and their costs below.

### Root cause (confirmed)

The per-axis trixel canvas is sized from the **base-resolution** cardinal extent and ignores the **effective subdivision factor**, but the Stage-1 per-axis store writes a **`world × effSub` integer lattice**:

- `system_voxel_to_trixel.hpp` FULL path: `worldFixed = round(worldAligned * subPerAxis)`, cell `= cellBase + faceInPlaneCoords(faceId, microWorld)` with `microWorld ≈ worldFixed`.
- `per_axis_canvas.hpp:65` allocates `IRMath::perAxisTrixelCanvasWorstCaseSize(cardinal.size_, 1.0f)` → `(2W, W+H)` from `cardinal.size_` only — **no `effSub` term**.
- `ir_iso_common.glsl` `faceLocalBase` centers on the anchor, then `writeDistanceTap` → `isInsideCanvas` **silently drops** any cell outside `[0, canvasSize)`.

Net: the canvas covers only `±(canvasSize/2) / effSub` **world** units around the focus/anchor. `effSub = clamp(m_voxelRenderSubdivisions × round(zoom), 1, 16)` (`RenderManager::getVoxelRenderEffectiveSubdivisions`). When `effSub` is large relative to zoom — i.e. `m_voxelRenderSubdivisions > 1`, or generally `effSub/zoom` above the canvas's ~2.8× margin — on-screen content beyond that shrinking world window is clipped to background. The anchor centering itself is correct (surviving content is centered + symmetric); only the **extent** is too small.

### Repro (dramatic)

`cardinal.size_=(642,722)` → canvas `(1284,1364)`. Force the clamp to bite at low zoom:

1. Temporarily `IRRender::setVoxelRenderSubdivisions(8)` in `shape_debug` init (so `effSub=clamp(8×2,1,16)=16` at zoom 2), and pan the spin camera to the scene centre (`vec2(56,0)`) so the gallery is on-screen.
2. `fleet-run IRShapeDebug --spin-yaw 30 --auto-screenshot 16 --zoom 2`, view the yaw-45° shot.

Result: only the central **±40-world** window around the focus renders; the entire rest of the gallery + platform is clipped to black. (At `effSub≈zoom`, e.g. shape_debug defaults with `m_voxelRenderSubdivisions=1`, the ~2.8× margin hides it — which is why the architect's zoom-4/zoom-8 captures "rendered fully" and the bug resisted isolation.)

### Why this bites in practice

`setVoxelRenderSubdivisions` is creation-driven: world config `voxel_render_subdivisions` (default 1, but `modifier_demo`=2), `perf_grid` / `lua_perf_grid` `--base-sub` CLI. Margin analysis (worst case at the clamp knee `zoom=16/m_vrs`, widened by the √2 rotation footprint):

- `m_vrs=1` (shape_debug + most demos): no clip (~2.8× margin).
- `m_vrs=2` (modifier_demo): ~1.4× margin — no clip, but thin.
- `m_vrs≥4` (perf-grid CLI, aggressive worlds): **clips**.

So it's a latent bug for high-subdivision configs, not the default demos.

### The design problem

The "obvious" fix from the issue ("size to the proven worst case in ×sub units") is **scale the canvas by `m_voxelRenderSubdivisions`** — which restores the margin at all zooms. But the cost is **×m_vrs² area**, and it fails exactly where the bug lives:

- `m_vrs=2` → ×4 area (~290 MB across the 3 axes × 5 maps during rotation).
- `m_vrs=4` → ×16 (~1.2 GB).
- `m_vrs=8` → ×64 (infeasible).

So sizing-by-`m_vrs` only helps the configs that **don't** clip; the high-`m_vrs` configs that **do** clip can't be fixed by sizing (memory explodes), and the design doc explicitly forbids blanket inflation (per-rotating-frame fill-rate). The tension is intrinsic: the #1310 face-local **integer** lattice (chosen for exact, parity-free `faceOriginFromInPlane` recovery) is incompatible with bounded canvas memory at high subdivision.

### Options for the architect (pick one)

1. **Cap the per-axis lattice density.** Clamp `subPerAxis` for the per-axis store path to `floor((canvasSize/2) / maxOnScreenWorldDisplacement)` so cells never overflow. Bounded memory, eliminates the clip, but loses sub-detail at high zoom on the rotating path only. Closest to "don't blanket-inflate."
2. **Decouple lattice resolution from `effSub`.** Store the per-axis lattice at a bounded (base/world) resolution and carry sub-detail as a fractional micro-offset recovered at scatter time (à la the cardinal `faceMicroPositionFixed` path), instead of expanding the integer lattice. Preserves detail + bounded memory, but reworks the #1310 store + `faceOriginFromInPlane` scatter recovery (fractional, and atomicMin occlusion per base cell).
3. **Bounded canvas + accept clip beyond a memory budget**, with a one-shot warn (like the light-volume out-of-range drop). Cheapest; documents the limit instead of fixing it.
4. **Size by `m_vrs` for the low range only** (e.g. ≤2) and cap, accepting options 1/3 above that.

I've left the WIP branch at just the claim commit (the repro used throwaway diagnostics — forced sub + temp pan + a debug log — which I reverted rather than commit). Happy to implement whichever option you choose. — opus-worker-2

jakildev:
## Architect direction

Root cause accepted exactly as diagnosed. Picking **Option 1** (cap the per-axis lattice density on the rotating path), NOT size-by-`m_vrs`:

- Size-by-`m_vrs` is rejected — ×m_vrs² area (×16 at m_vrs=4, infeasible at m_vrs=8), only helps configs that don't clip, and violates the no-blanket-inflation rule.
- **Option 1:** clamp `subPerAxis` so cells can never overflow the canvas — `subPerAxisCapped = min(effSub, floor((canvasSize/2) / maxOnScreenWorldDisplacement))`, deriving the displacement from the cardinal extent + the √2 rotation footprint already used to size the canvas. Bounded memory, eliminates the clip, touches only the rotating path. Add a one-shot warn when the cap engages (mirror the light-volume out-of-range drop) so the detail loss is observable, not silent.

Accepted trade-off: loses sub-voxel detail at high zoom *while rotating only*. The detail-preserving rework (your Option 2 — decouple lattice resolution from `effSub`, fractional micro-offset à la cardinal `faceMicroPositionFixed`) is **deferred to #1458**; do not bundle it here.

Invariants: cardinal (`residualYaw==0`) byte-identical (per-axis unallocated there); GLSL/Metal parity for the cap.

Task pointer: `~/.fleet/plans/issue-1431.md` updated with the full decision + acceptance.

jakildev:
Implemented architect direction (Option 1): cap the per-axis lattice density so face-local cells can never overflow the bounded per-axis canvas.

**What changed**
- `IRMath::perAxisSubdivisionCap(cardinalExtent, zoom)` — derives the cap from the visible iso extent (`cardinalExtent / 2·zoom`) inverted to in-plane world on the depth-0 plane, grown by the √2 residual-yaw footprint the canvas is sized for. CPU-computed because `effSub = clamp(m_vrs·round(zoom),1,16)` conflates subdivision and zoom — a shader given only `effSub` + `canvasSize` cannot recover the zoom-dependent on-screen extent.
- `IRPrefab::PerAxisCanvas::subdivisionDensity()` / `setUboSubdivisionDensity()` — the capped density rides in `voxelRenderOptions.y`, so the store (`subPerAxis` + `trixelFrameOffset`), the framebuffer forward-scatter (`perAxisBase` + `zoomEff`), and the per-axis AO / sun-shadow / lighting world recovery (`perAxisCellToWorld3D`) all share one consistent world↔cell scale.
- Stage-1 store (GLSL + Metal): `if (uPerAxis >= subPerAxis) return;` surplus-invocation guard — the compact pass sizes the indirect dispatch Z-count from the uncapped `effSub`, so capped invocations beyond `subPerAxis²` must skip.
- One-shot log when the cap engages (mirrors the light-volume out-of-range warn).

**Invariants held**
- Cardinal `residualYaw==0` is **byte-identical** to master — verified: the yaw=0 shot is `cmp`-identical between master and this branch (per-axis canvases are unallocated at cardinals, so the cap code never runs there).
- GLSL/Metal store guards kept in parity.
- Deferred per the architect: the detail-preserving rework (decouple lattice resolution from `effSub`, fractional micro-offset) → #1458.

**Validation** (forced `voxel_render_subdivisions=8`, pan to gallery centre, `--spin-yaw --zoom 2` — the architect’s dramatic repro; throwaway diagnostic reverted): cap engages **16 → 4**, the full gallery renders at yaw 45° with no black clip. Default config (`m_vrs=1`) is unaffected — the cap is a no-op there.

| Before (master) — yaw 45°, clipped | After (this PR) — yaw 45°, full gallery |
|---|---|
| before | after |

Reviewer: please re-evaluate the rebased diff. Note this touches GLSL + Metal — cross-host smoke applies. — opus-worker-2

jakildev (commented):
## Review — render: shapes clipped out of view under camera Z-yaw — per-axis canvas sizing (#1431)

**Verdict:** needs-fix

### Needs-fix

- `engine/render/src/shaders/c_voxel_to_trixel_stage_2.glsl:158` and `engine/render/src/shaders/metal/c_voxel_to_trixel_stage_2.metal:173` — **Missing surplus-invocation guard in stage-2 (GLSL + Metal).**

  Stage-1 gained `if (uPerAxis >= subPerAxis) return;` with a clear comment: the indirect dispatch Z-count was sized from the uncapped `effSub` by the compact pass, so `gl_WorkGroupID.z` ranges over `effSub²`, not `cappedSub²`. Stage-2 has the identical `subPerAxis / uPerAxis / vPerAxis` block but no guard.

  Why this matters: `writeColorTap` does check `isInsideCanvas` and then `voxelDistance == canvasDistance`, so OOB positions and most depth mismatches are harmless. But stage-2 surplus invocations compute a `microWorld` at `uPerAxis = cappedSub` that maps to a canvas position one integer voxel past the capped range. In isometric view, many voxels share the same iso-depth plane (`x + y + z` is constant across the depth front). When a surplus invocation's computed depth (`microWorld.x + microWorld.y + microWorld.z`) matches the winning distance at the "stepped" position (set by a non-surplus stage-1 invocation for the voxel that legitimately owns that cell), `writeColorTap` writes the **wrong voxel's color** at that position. This is most likely when a voxel at the canvas center is on the same depth plane as its neighbour (common at any flat-faced structure). The effect is a subtle color bleed at canvas edge cells — only visible when the cap fires (`effSub > cappedSub`, i.e. `m_vrs ≥ 4` or aggressive zoom).

  **Fix:** add `if (uPerAxis >= subPerAxis) return;` immediately after the `vPerAxis` computation in both shaders' per-axis block, mirroring stage-1 exactly.

  GLSL (`c_voxel_to_trixel_stage_2.glsl`), inside `if (perAxisRoute != 0)`, after line 160:
  ```glsl
  const int subPerAxis = max(voxelRenderOptions.y, 1);
  const int uPerAxis = int(gl_WorkGroupID.z) / subPerAxis;
  const int vPerAxis = int(gl_WorkGroupID.z) % subPerAxis;
  if (uPerAxis >= subPerAxis) return;   // ← add this line
  ```

  Metal (`c_voxel_to_trixel_stage_2.metal`), inside `if (frameData.perAxisRoute != 0)`, after line 175:
  ```metal
  const int subPerAxis = max(frameData.voxelRenderOptions.y, 1);
  const int uPerAxis = int(groupId.z) / subPerAxis;
  const int vPerAxis = int(groupId.z) % subPerAxis;
  if (uPerAxis >= subPerAxis) return;   // ← add this line
  ```

### Nits

- `engine/math/include/irreden/ir_math.hpp:1416` — The docstring cites `|wx|,|wy| ≤ isoHalf.x/2 + isoHalf.y/6` as the `isoPixelToPos3D` closed-form bound. The `1/6` coefficient is non-obvious and differs from what standard iso projection algebra gives at depth-0. A one-line derivation comment (citing the 2:1 horizontal trixel scale and the `isoPixelToPos3D` form) would help future maintainers verify the bound without re-deriving it.

### Praise

- Root-cause analysis in the NEEDS-DESIGN comment is exemplary: exact shader path cited, dramatic repro recipe (forced `m_vrs=8`, pan, `--spin-yaw --zoom 2`), and four well-framed architectural options with cost estimates. This is exactly what the architect needs to make a bounded decision.
- `setUboSubdivisionDensity`'s offset arithmetic is correct: `ivec2 voxelRenderOptions_` has `.y` at `offsetof(voxelRenderOptions_) + sizeof(int) = +4` — matches the std140 layout.
- All three per-axis lighting passes (AO, sun-shadow, lighting-to-trixel) consistently use `setUboSubdivisionDensity` + restore pattern; none of them forget the restore.
- The one-shot cap-engaged `IRE_LOG_INFO` mirrors the light-volume out-of-range warn pattern exactly.
- Cardinal `residualYaw==0` byte-identity verified (per-axis canvases unallocated there; yaw=0 screenshot byte-identical to master).
- Before/after screenshots confirm the dramatic repro is fully fixed.

- [ ] Force `m_vrs = 4` or `m_vrs = 8` in a creation with dense voxels (e.g. `shape_debug` gallery), spin to a non-cardinal yaw, and confirm no color bleed on edge voxels after the stage-2 guard is added.
- [ ] Confirm the yaw=0 screenshot remains byte-identical to master (cardinal fast path unaffected).
- [ ] Build on both GLSL (Linux/macOS OpenGL) and Metal (macOS) after the stage-2 shader changes.

Opus recheck required: the `perAxisSubdivisionCap` formula uses `maxInPlaneXY = isoHalfX/2 + isoHalfY/6` for the max on-screen XY world displacement. The `1/6` coefficient is specific to the engine's exact iso coordinate conventions (canvas trixel vs. screen-pixel scale) and cannot be reliably verified against the `isoPixelToPos3D` implementation in a single-file scan. If the bound overcounts max displacement, the cap is too aggressive (unnecessary detail loss); if it undercounts, edge voxels still clip. Confirm the closed-form inverse matches the actual `isoPixelToPos3D` implementation.

jakildev:
Addressed feedback:
- Added `if (uPerAxis >= subPerAxis) return;` surplus-invocation guard to stage-2 (GLSL: `c_voxel_to_trixel_stage_2.glsl` after `vPerAxis` computation; Metal: `c_voxel_to_trixel_stage_2.metal` same location). Mirrors stage-1 exactly. Comment cites the effSub²/cappedSub² dispatch mismatch.
- Added depth=0 derivation comment for the `1/6` coefficient in `perAxisSubdivisionCap` (addresses nit at ir_math.hpp:1416): at depth=0 (x+y+z=0), iso.y=-x-y+2z=3z → z=iso.y/3, y=iso.x/2-iso.y/6.

— sonnet-fleet-2

jakildev (commented):
## Re-review — render: shapes clipped out of view under camera Z-yaw — per-axis canvas sizing (#1431)

**Verdict:** approve

### Prior needs-fix resolved

- `c_voxel_to_trixel_stage_2.glsl` and `c_voxel_to_trixel_stage_2.metal` — **Stage-2 surplus-invocation guard added. ✓**

  `if (uPerAxis >= subPerAxis) return;` is now present immediately after the `vPerAxis` computation in both backends, mirroring stage-1 exactly. The placement is correct: the guard fires before any read of `uPerAxis` or `vPerAxis` in the microWorld / faceLocalBase computation, so surplus invocations exit cleanly without stepping to the wrong canvas cell.

### Prior Opus recheck resolved

The previous pass flagged "Opus recheck required" because the `1/6` coefficient in `maxInPlaneXY = isoHalfX/2 + isoHalfY/6` could not be verified against `isoPixelToPos3D` from a single-file scan.

The author added the one-line derivation comment (nit addressed ✓):
```
// At depth=0 (x+y+z=0): iso.y=-x-y+2z=3z → z=iso.y/3, y=iso.x/2-iso.y/6.
```

I verified this against the actual `isoPixelToPos3D(vec2 iso, float depth)` implementation in `ir_math.hpp:319`:
```
wx = (2*depth - 3*iso.x - iso.y) / 6
```
At depth=0: `wx = (-3*iso.x - iso.y)/6 = -iso.x/2 - iso.y/6`, so `|wx| ≤ isoHalfX/2 + isoHalfY/6`. The bound is tight and correct. `wz = iso.y/3` also follows from the formula (substituting depth=0 and simplifying). The `perAxisSubdivisionCap` formula is mathematically sound.

### Nit addressed

- `ir_math.hpp` — depth=0 derivation comment for the 1/6 coefficient added. ✓

### Test plan

- [ ] Force `m_vrs = 4` or `m_vrs = 8`, spin to a non-cardinal yaw — confirm no color bleed on edge voxels with the stage-2 guard in place.
- [ ] Confirm the yaw=0 screenshot remains byte-identical to master (cardinal fast path unaffected).
- [ ] Build on both GLSL (Linux/macOS OpenGL) and Metal (macOS).

Opus recheck not required.

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). Build exit 0, run exit 0, all 16 screenshots saved, no shader compile warnings.

jakildev:
Cross-host smoke **OK on macOS / Metal** (fresh-checkout `fleet-build --target IRShapeDebug` + `IRShapeDebug --auto-screenshot 10`, 16/16 shots captured, clean exit / no crash).

Opus visual pass (judgment-bearing half) — inspected the PR's risk-area shots:
- `zoom4_yaw45_inter_cardinal` + `zoom4_pan16_yaw45_pivot` (non-cardinal yaw): all shapes render fully in view, no clip-out under Z-yaw — the per-axis canvas sizing fix holds on Metal.
- `zoom8_yaw180` (eff. subdivisions = 8) + edge-face crops: clean cube/face edges, no edge-voxel color bleed (stage-2 surplus-invocation guard working), no checkerboard/lattice artifact.
- `zoom4_pan16_yaw0_pivot` cardinal baseline unchanged.

— opus-worker-2

jakildev (commented):
## Opus re-review — render: per-axis canvas sizing — camera Z-yaw clip (#1431) (engine PR #1452)

**Verdict:** approve (with two latent nits below)

The first Sonnet pass set "Opus recheck required" on the `perAxisSubdivisionCap` cap formula — specifically whether `maxInPlaneXY = isoHalfX/2 + isoHalfY/6` (the `1/6` coefficient) matches the real `isoPixelToPos3D`. Sonnet's re-review then self-resolved that flag with an algebraic derivation against `ir_math.hpp:319`. **I independently re-derived the same algebra and confirm Sonnet exactly** — and I also found one assumption baked into the formula that neither pass surfaced. Both items are latent (no regression, no crash), so this stays an approve, but they're worth recording.

### Confirmed: the depth-0 inverse is exactly right

Against `isoPixelToPos3D(vec2 iso, float depth)` (`ir_math.hpp:318`), at `depth=0`:
- `wx = (-3·iso.x - iso.y)/6 = -iso.x/2 - iso.y/6` → `|wx| ≤ isoHalfX/2 + isoHalfY/6`
- `wy = wx + iso.x = iso.x/2 - iso.y/6` → same bound
- `wz = (iso.y + 2·wx + iso.x)/2 = iso.y/3` → `|wz| ≤ isoHalfY/3`

So `maxInPlaneXY` bounds both in-plane world axes and `maxInPlaneZ` bounds the z axis; the canvas-Y `max(maxInPlaneZ, maxInPlaneXY)` is the correct conservative choice (canvas-Y stores z for X/Y faces, x/y for the Z face). The `kSqrt2` factor is self-consistent, not a double-count: `perAxisTrixelCanvasWorstCaseSize` already grows the canvas by ~√2 (`scaleX`), so `canvasHalf / (kSqrt2·maxDisp)` cancels to the base-resolution footprint with rotation headroom. **Math confirmed.**

I also verified the set/restore plumbing end-to-end: `setUboSubdivisionDensity` writes `voxelRenderOptions_.y` at the correct `offsetof(...) + sizeof(int)` (ivec2, second int); stage-1 sets `cappedSub` before the per-axis loop and restores `uncappedSub` on exit; AO / sun-shadow / lighting each patch capped-before / `effSub`-after their loop; the scatter recomputes the identical `subdivisionDensity()`. Store, recovery, and scatter therefore share one world↔cell scale. The cardinal / yaw=0 fast path is provably unchanged (`densityRescale = effSub/cappedDensity = 1` when the cap doesn't bite; NONE mode returns `effSub` directly). The stage-2 surplus-invocation guard (Sonnet's earlier needs-fix) is correctly placed in both backends.

### Nit 1 (latent, non-blocking) — the cap bounds in-plane spread of the visible iso *rect*, but ignores the visible *depth range*

`perAxisSubdivisionCap` derives `maxOnScreenWorldDisp` purely from the depth-0 inverse of the visible iso rectangle. But an on-screen voxel at depth `d` (= `x+y+z`) sits at `world = W₀ + (d/3)·(1,1,1)`, and that `(d/3,d/3,d/3)` shift has a non-zero component along every face's in-plane (tangent) axes. So for a scene with a large visible depth range, face-local cells can still land outside the capped canvas and clip — the cap fully covers the iso-rect spread (the dominant term, and the case in the before/after screenshots) but not the additive depth-range term.

This is **not a regression** (before this PR there was no cap at all, so the high-`effSub` clip was unconditional) and **not unsafe** (`isInsideCanvas` drops any residual OOB cell to background — no corruption or overstep, and the stage-2 guard handles the surplus invocations). #1431's reported gallery case is genuinely fixed. I'm flagging it so a future "still clipping under yaw on a deep scene" report is recognized as this known residual rather than a new bug — a follow-up to fold the depth-range term into the bound (or clamp the visible depth span feeding the cap) would close it completely. Human's call whether to file that now.

### Nit 2 (latent coupling, non-blocking) — cap recomputes canvas size with the default `minOnScreenTrixelPx`, not the allocation's constant

`perAxisSubdivisionCap` calls `perAxisTrixelCanvasWorstCaseSize(cardinalExtent)` with the **default** `minOnScreenTrixelPx = 1.0f`, while `syncAllocationToCameraYaw` allocates the actual canvas with `kMinOnScreenTrixelSizePx`. These are equal today (`kMinOnScreenTrixelSizePx == 1.0f`), so the cap is computed against the true allocated size and there's no active bug. But it's an implicit coupling: if `kMinOnScreenTrixelSizePx` is ever bumped above ~1.41, the allocation shrinks (`scaleX` drops from `2.0` toward `kSqrt2`) while the cap's hardcoded default-arg call does not — the cap would then be sized against a *larger-than-real* canvas, i.e. too loose, and the clip returns silently. Cheap hardening: pass `kMinOnScreenTrixelSizePx` through to `perAxisTrixelCanvasWorstCaseSize` in the cap (or hand the already-allocated `axes.size_` to the cap directly so the two can't drift). Nit only.

### Net

Sonnet's approve stands — the cap formula is mathematically sound and the plumbing is consistent. Approving. The two items above are latent and don't warrant pulling an approved, macOS-smoked PR back to needs-fix; they're follow-up-issue material at most.

jakildev:
Addressed Opus latent nit (Nit 2 from re-review):

- **`ir_math.hpp:perAxisSubdivisionCap`** — added `minOnScreenTrixelPx = 1.0f` parameter and threaded it through to `perAxisTrixelCanvasWorstCaseSize`, so the cap is sized against the actual allocated canvas rather than the hardcoded default. Call site in `per_axis_canvas.hpp:subdivisionDensity` now passes `kMinOnScreenTrixelSizePx` explicitly. No behavior change today (`kMinOnScreenTrixelSizePx == 1.0f`); prevents silent drift if the constant is ever bumped.

Filed follow-up issue **#1469** for Nit 1 (depth-range residual in the cap derivation — the cap covers the iso-rect spread but not the additive depth-range contribution for deep scenes). Opus-tier reasoning needed per the reviewer's own framing; deferred per the approval. — sonnet-fleet-2

PR #1485, opened 2026-06-03
fleet: stagger dispatch + short-window 429 backoff

jakildev:
## Summary
- Fixes the `fleet-up` startup burst that trips Anthropic's short-window 429 (`Server is temporarily limiting requests (not your usage limit)`) — observed with several panes erroring at once.
- **Root cause:** the dispatcher fires every idle pane in one tick (~8 cold `claude` cold-starts, each reading the large `state.json`, inside one minute). The existing `pane_stagger` only spaces pane *creation* + the per-pane `git fetch`, not the claude launches. And a 429 isn't exit code 2, so no cooldown engaged — panes could re-fire into it, and reviewer/smoke roles (no periodic re-arm) could stall.

## Changes
- **`fleet-dispatcher`** — global min-gap `FLEET_DISPATCH_MIN_GAP_SECONDS` (default 8) between any two launches across all roles/panes (tracked via `LAST_DISPATCH_EPOCH`); a *minimum* gap, so only simultaneous bursts get spread (~one launch per tick). Gap-deferred panes keep the trigger for later ticks. `pane_in_rate_limit_cooldown` now also honors a short 429 marker (`<pane>.throttle.ts`, `FLEET_DISPATCHER_RATE429_DELAY` default 90) next to the usage-limit marker. New `--dispatch-gap-check` test hook.
- **`fleet-dispatch-wrap`** — on a non-zero exit *with* the throttle flag set: short throttle cooldown + **re-arm the role trigger** so the pane retries when the window clears (critical for reviewer/smoke). `rc==2` keeps the long usage-limit cooldown; clean exits write nothing.
- **`fleet-claude-stream`** — detect the throttle line, touch `FLEET_THROTTLE_FLAG` for the wrapper, and annotate it as a warning (claude often retries it internally and still succeeds).
- **`fleet-up.conf.sample`** — documents both tunables.
- **`tests/test_dispatcher_stagger.sh`** — 24 cases (gap math, wrapper 429/usage/clean classification + re-arm, stream flag-touch).

## Test plan
- [x] `bash scripts/fleet/tests/test_dispatcher_stagger.sh` → 24/0.
- [x] Regressions: `test_dispatcher_effort.sh` 13/0, `test_dispatcher_concurrency_cap.sh` 22/0, `test_fleet_claude_stream_latch.py` 4/4. (`test_dispatcher_usage_gate.sh` has one **pre-existing** failure on master, unrelated — a `>=` vs `>` boundary; flagged separately.)
- [ ] After merge: pull master, `fleet-down && fleet-up`, watch `~/.fleet/logs/dispatcher.log` at boot — launches should land ~one per tick instead of all at once.

## Activation
The live dispatcher runs from the main checkout via `~/bin`; this goes live only after merge + `git pull` on master + a dispatcher restart (`fleet-down && fleet-up`). `fleet-dispatch-wrap` / `fleet-claude-stream` are invoked fresh each iteration, so they take effect on merge+pull with no restart.

### DIFF
--- a/scripts/fleet/fleet-claude-stream
+++ b/scripts/fleet/fleet-claude-stream
@@ -193,6 +193,29 @@ BENIGN_WARNING_PREFIXES = (
 )
 
 
+# Short-window 429 throttle line: "API Error: Server is temporarily limiting
+# requests (not your usage limit)". claude retries these internally; the line
+# surfaces either as a post-result cleanup warning (the run still succeeded) or
+# after retries were exhausted (the run failed). It is NOT a usage-limit/quota
+# event, so it's annotated as a warning rather than a fatal error. When
+# FLEET_THROTTLE_FLAG is set (by fleet-dispatch-wrap), seeing this line also
+# touches that flag so the wrapper applies a short throttle cooldown + re-arm
+# the role IF the process also exits non-zero.
+THROTTLE_ERROR_PREFIX = "API Error: Server is temporarily limiting requests"
+
+
+def note_throttle():
+    flag = os.environ.get("FLEET_THROTTLE_FLAG")
+    if not flag:
+        return
+    try:
+        pathlib.Path(flag).touch()
+    except OSError:
+        # State-dir not writable, etc. — non-fatal; the wrapper just won't see
+        # the flag and skips the short cooldown.
+        pass
+
+
 def main():
     for line in sys.stdin:
         line = line.strip()
@@ -206,7 +229,10 @@ def main():
             # streams in temporal order). Annotate known post-result
             # warnings as warnings, pass through everything else
             # untouched so we don't accidentally hide a real error.
-            if any(line.startswith(p) for p in BENIGN_WARNING_PREFIXES):
+            if line.startswith(THROTTLE_ERROR_PREFIX):
+                note_throttle()
+                emit(f"{YELLOW}⚠ [claude-warn] {line}{RESET}\n")
+            elif any(line.startswith(p) for p in BENIGN_WARNING_PREFIXES):
                 emit(f"{YELLOW}⚠ [claude-warn] {line}{RESET}\n")
             else:
                 emit(line + "\n")
--- a/scripts/fleet/fleet-dispatch-wrap
+++ b/scripts/fleet/fleet-dispatch-wrap
@@ -36,15 +36,38 @@ else
     export FLEET_ROLE_MODEL=sonnet
 fi
 
+# Short-window 429 detection. fleet-claude-stream touches this flag if it sees
+# the "Server is temporarily limiting requests (not your usage limit)" error in
+# the stream. Combined with a non-zero exit below, that's the signal to apply a
+# SHORT throttle cooldown + re-arm the role — distinct from the 15-min
+# usage-limit cooldown, which is the wrong tool for a per-minute load-shed.
+# Clear any stale flag from a previous run first.
+mkdir -p "$RATE_LIMIT_DIR"
+export FLEET_THROTTLE_FLAG="$RATE_LIMIT_DIR/$PANE_KEY.throttle-seen"
+rm -f "$FLEET_THROTTLE_FLAG"
+
 claude --model "$MODEL" --effort "$EFFORT" --print \
     --output-format stream-json --include-partial-messages --verbose \
     "/role-$ROLE live" | fleet-claude-stream
 
 rc="${PIPESTATUS[0]}"
+TRIGGERS_DIR="${FLEET_STATE_DIR:-$HOME/.fleet/state}/triggers"
 if [[ "$rc" == "2" ]]; then
-    mkdir -p "$RATE_LIMIT_DIR"
+    # Usage-limit (quota) exit — long cooldown via the .ts marker.
     date +%s > "$RATE_LIMIT_DIR/$PANE_KEY.ts"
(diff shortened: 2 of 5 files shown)


jakildev (commented):
## Review — fleet: stagger dispatch + short-window 429 backoff (engine PR #1485)

**Verdict:** approve

Standalone PR. Fleet tooling only — no engine/render/ECS/shader code. Three coordinated pieces (`fleet-dispatcher`, `fleet-dispatch-wrap`, `fleet-claude-stream`) with 24 new test cases.

### Checklist results

**ECS / render / lighting / math / serialization** — not touched. ✓

**Logic correctness — `pane_in_rate_limit_cooldown` refactor (`fleet-dispatcher`)**

The loop-based two-marker check is correct across all cases:
- Both markers absent → `in_cooldown` stays 1 → `unset` logged flag → return 1 (not in cooldown). ✓
- Either marker active → `in_cooldown=0` → return 0 (in cooldown). ✓
- Expired markers reaped (`rm -f`). ✓
- `unset "RATE_LIMIT_COOLDOWN_LOGGED"` only fires when *both* markers have expired — correct, previously it fired when the single marker expired, now it waits for all cooldowns to clear. ✓

**Logic correctness — `LAST_DISPATCH_EPOCH` / gap_blocked path**

- `LAST_DISPATCH_EPOCH=0` initialization guarantees the first launch never blocks (0 vs. ~1.7B current epoch). ✓
- `LAST_DISPATCH_EPOCH` updated only after a successful `write_dispatch_record` call, not on gap-blocked break. ✓
- `gap_blocked=1` + `break` stops the fan-out and returns early, keeping the trigger so deferred panes fire on later ticks — this is the correct spreading mechanism. ✓
- `DISPATCH_MIN_GAP_SECONDS=0` disables the gap entirely (gated by `DISPATCH_MIN_GAP_SECONDS > 0`). ✓

**Logic correctness — `fleet-dispatch-wrap` throttle handshake**

- `mkdir -p "$RATE_LIMIT_DIR"` moved before the claude invocation: the directory is guaranteed to exist for both the `rc==2` branch and the throttle branch. ✓
- `FLEET_THROTTLE_FLAG` path is `$RATE_LIMIT_DIR/$PANE_KEY.throttle-seen`; cleared before the run and deleted after regardless of exit code. ✓
- `rc==2` (usage-limit/quota) → long `.ts` cooldown, no re-arm. ✓
- `rc!=0 && -f FLEET_THROTTLE_FLAG` (short-window 429 exhausted retries) → short `.throttle.ts` cooldown + re-arm trigger. ✓
- `rc==0` (clean exit) → no markers written. ✓

**Logic correctness — `fleet-claude-stream` flag-touch**

- `note_throttle()` reads `FLEET_THROTTLE_FLAG` from env; no-op if unset. ✓
- `pathlib.Path(flag).touch()` wrapped in `OSError` catch — non-fatal. ✓
- `pathlib` is already imported at the top of the file. ✓
- Throttle line routed before `BENIGN_WARNING_PREFIXES` check so it takes the `emit(f"{YELLOW}⚠ ...")` path rather than whatever falls through to the unrecognized-output passthrough. ✓

**Tests / build**

- 24 cases across 3 test parts (gap math, wrap exit classification, stream flag-touch). ✓
- Regressions: `test_dispatcher_effort.sh` 13/0, `test_dispatcher_concurrency_cap.sh` 22/0, `test_fleet_claude_stream_latch.py` 4/4 (per PR body). ✓
- The stub design in T4-T7 correctly exercises the `PIPESTATUS[0]` path — `claude` stub exits with `FAKE_CLAUDE_RC`, `fleet-claude-stream` stub drains stdin and optionally touches the throttle flag. ✓

**Naming / style** — consistent with existing conventions. ✓

### Nits

- `scripts/fleet/fleet-dispatcher:562` — `in_cooldown=1` uses 0 to indicate "a cooldown was found" (return 0) and 1 to indicate "no cooldown found" (return 1). The value is the Bash exit-code equivalent, not an intuitive boolean; a reader unfamiliar with the convention initially reads `in_cooldown=0` as "not in cooldown". Consider `free=1` (flipped to 0 when any marker is active) or `found_active=0` (flipped to 1) for clarity.
- `scripts/fleet/fleet-dispatcher:964` (post-`break` log message) — `dispatched > 0` path logs the count but `dispatched == 0` (gap fires on the first pane) is silent. Harmless, but a `log "$role: min-gap deferring first idle pane — trigger kept"` for the zero case would make cold-start traces easier to read.

- [x] `bash scripts/fleet/tests/test_dispatcher_stagger.sh` → 24/0 (completed per PR body).
- [x] Regression suites all green (per PR body).
- [ ] After merge: `fleet-down && fleet-up`, watch `~/.fleet/logs/dispatcher.log` — launches should spread ~one per tick instead of bursting.

Opus recheck not required.

jakildev:
Addressed nits:
- Renamed `in_cooldown` → `found_active` in `pane_in_rate_limit_cooldown`: initial value `0`, set to `1` when an active marker is found. The conditional now reads `if (( found_active )): return 0` (in cooldown) which matches the name instead of inverting it.
- Added `log "$role: min-gap deferring first idle pane — trigger kept"` for the `dispatched == 0` branch of the gap-blocked path so cold-start dispatcher traces capture the min-gap deferral even when no pane fired this tick.

All 24 stagger tests still pass (24/0). — [sonnet-author]

Issue #1488, opened 2026-06-03
fleet: harden design-blocked → unblock → resume lifecycle against stale claim labels + stacked-on-merged base


jakildev:
**Model:** opus
**Area:** scripts/fleet/fleet-claim, .claude/commands/role-opus-worker.md, .claude/commands/role-merger.md
**Safety net already landed:** PR #1487 (architect self-heals at unblock) — this issue is the root-cause cure so the architect step becomes belt-and-suspenders, not load-bearing.

## Observed incident

opus-worker-1 was idle while detached-SO(3) P3b (#1475 / PR #1483) sat unworked after an architect unblock. Two independent stale-state traps, both surviving a `fleet:design-blocked` escalation:

1. **Orphaned claim labels.** The escalating worker released its FS claim and reset the branch ("releasing the claim so any worker can resume"), but the issue's `fleet:claim-<host>-<agent>` + `fleet:in-progress` GitHub labels lingered. The TTL cleanup sweep (`fleet-claim cleanup --gh` Pass 3) won't clear them: its "no matching WIP PR branch exists" guard sees the live `fleet:wip` PR and treats the claim as active. Net: the scout sees the issue as in-progress → no worker can claim it.
2. **Stacked-on-merged base.** The base PR merged while the stacked PR was still `fleet:wip`; the merger never re-targeted it because the merged-base re-target skips `fleet:wip` PRs. The worker's stacked-resume then has no clean rebase target.

### Fix A — design-blocked escalation fully releases the issue claim
The worker's escalation flow (`role-opus-worker.md`, the `fleet:design-blocked` step) currently releases the FS claim but leaves the GitHub `fleet:claim-*` + `fleet:in-progress` labels on the backing issue. Under the "release so any worker can resume" model (which is the intended one — don't tie up a worker waiting on the architect), the escalation must release **both sides**. Make `fleet-claim release` clear the issue-side `fleet:claim-<host>-<agent>` + `fleet:in-progress` labels (it may already clear the FS lock + flip the row, but confirm it also clears these GitHub labels), and have the escalation step call it.

### Fix B — cleanup sweep treats a parked PR's claim as sweepable
[...]

PR #711, opened 2026-05-14
T-180: asset: hoist .vxs.json sidecar keys + VoxelSetMode string to named constants

jakildev:
## Summary
- Add 9 `constexpr std::string_view kSidecarKey*` constants in the anonymous namespace of `voxel_set_format.cpp`, anchoring every JSON key literal used by `emitVoxelSetSidecar`. Future sidecar emitters (`.rig`, `.prefab.lua`, world-snapshot) grep for a single constant name to find the canonical spelling across formats.
- Add `constexpr std::string_view voxelSetModeToString(VoxelSetMode) noexcept` helper next to the `VoxelSetMode` enum in the public header. Collapses the per-call-site inline switch in `emitVoxelSetSidecar` to a single `valueString` call, parallel to `kShapeTypeTable`.
- `JsonSidecarWriter::key` already accepts `std::string_view` — no signature change needed.
- No behavior change; sidecar JSON output is byte-identical to pre-refactor.

## Test plan
- [x] `IrredenEngineTest` 586/586 pass, including `VoxelSetDense.SidecarEmittedAlongsideBinary`, `VoxelSetHybrid.SidecarEmittedAlongsideBinary`, `VoxelSetFormat.SidecarEmittedAlongsideShapeGroupBinary`
- [x] `fleet-build --target format-changed` + rebuild clean

Closes #697

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #710, opened 2026-05-14
docs: lod-strategy.md — artist-driven LOD architecture (Phase 1 + Phase 2 sketch)

jakildev:
## Summary
- Captures the three-layer split for LOD: asset binary (`.vxs` / `.rig`, unchanged) / runtime selector (Phase 1, #708) / prefab manifest (`.prefab.lua`, Phase 2, future T-173 extension).
- Documents the decision to NOT bake LOD tiers into the binary asset formats — and why the format-extensibility rules keep that decision reversible if Phase 2 reveals it was wrong.
- Phase 1 scope (in flight as #708): wire the existing `lod_utils.hpp` stub, add `C_ActiveLodLevel` singleton + `LOD_UPDATE` system, filter `C_ShapeDescriptor::lodMin_` in `SHAPES_TO_TRIXEL` pre-GPU.
- Phase 2 (sketched, no ticket): `lod = { [LOD_0] = "flower_silhouette.vxs", ... }` in the prefab manifest; open questions on lazy-vs-eager loading and additive-vs-replace composition called out for the worker on T-173 to resolve.
- Phase 3 (deferred): cross-tier alpha cross-fade vs per-shape ramp; both runtime-side, no format extension required.

## Context
Doc grew out of an engine/asset cleanup-pass review (parent epic #603 closeout). Surfaced the question "should LOD support land in the file types now?" — this doc is the architectural assessment + decision record. Companion follow-ups filed:
- #704 — hoist vec3/vec4 + color binary I/O helpers into `engine/math/`
- #705 — delete the entire `.txl` family (scope expanded after audit confirmed full deletion is correct — see below)
- #706 — `ShapeRecord` serialized annotation + dead-stub cleanup + `makeTag` length assert + CLAUDE.md refresh
- #707 — `JsonSidecarWriter` + `NameTable` round-trip test coverage
- #708 — LOD Phase 1 runtime selector implementation ticket

## Test plan
- [ ] Render-pipeline reviewer confirms the proposed `SHAPES_TO_TRIXEL` CPU-side filter insertion point matches their understanding of the stage's beginTick / per-entity-tick separation.
- [ ] Prefab-format owner confirms the Phase 2 `lod = { ... }` table shape is compatible with the planned T-173 DSL.
- [ ] Epic-owner confirms the "don't bake into `.vxs`" decision matches the entity-editor UX direction (single-tier-per-file editing model).

## Notes for reviewer
- No code changes — pure design-doc commit.
- All file-path citations verified against current `master`.

[...]

PR #1492, opened 2026-06-03
fleet/fleet-claim: clear parked-PR claim labels on release + sweep (#1488 Fixes A+B)

jakildev:
## Summary

Hardens the design-blocked → unblock → resume lifecycle against the
stale-claim trap from #1488's observed incident (an opus-worker sat idle
while a parked PR went unworked). Implements **Fixes A and B**; **Fix C is
carved out** for the human (gated — see below).

Root cause: a worker that escalates `fleet:design-blocked` releases its
claim so any worker can resume after the architect unblocks — but the
issue's `fleet:claim-<host>-<agent>` + `fleet:in-progress` GitHub labels
lingered. The scout then kept seeing the issue as in-progress, so no worker
could re-claim it.

### Fix A — `fleet-claim release` clears a parked PR's claim labels
`cmd_release` now detects when the issue's matching open PR is *parked*
(`fleet:design-blocked`/`-unblocked`) and clears this host's
`fleet:claim-<host>-*` + `fleet:in-progress` labels off the issue. A normal
PR-open release (active matching PR) keeps the labels through the merge
lifecycle, exactly as documented in `AUTHOR-PIPELINE.md` "Claim-label
lifecycle" — so the duplicate-work guard on the common path is unchanged.
The design-escalation step already calls `fleet-claim release <N>` (after
adding `fleet:design-blocked`), so this needs **no role-doc change**.

### Fix B — `cleanup --gh` sweeps a parked claim like a no-PR abandon
The open-issue claim sweep previously kept a claim whenever *any* matching
open PR existed. It now keeps the claim only when an *active* (non-parked)
matching PR exists; a parked matching PR is treated like a no-PR abandon and
swept (subject to the existing TTL). This is the safety net that makes Fix A
self-correcting even if the `fleet:design-blocked` label landed late or
release was skipped.

### Shared classifier
Both fixes route through a new `issue_pr_state(prs, issue, repo)` →
`active` / `parked` / `none` helper in `fleet_branch_match.py` (the existing
shared branch↔issue matcher), so the two call sites can't drift. A small
`_issue_pr_state_from` bash wrapper feeds it from a `gh pr list` blob.

## Fix C — carved out (gated, human action required)

Fix C ("the merger re-targets `fleet:wip` PRs whose base merged") is a
behavior change to the **merger**, which is a pure LLM role: the re-target
logic lives only as prose in `.claude/commands/role-merger.md` steps 2.5/2.6
(no script performs a `gh pr edit --base`). Editing `role-*.md` is blocked
for queue-sourced workers by the deterministic self-config gate, so this PR
cannot apply it.

**Human action for Fix C:** in `.claude/commands/role-merger.md`, allow the
**re-target step specifically** to run on `fleet:wip` PRs — when a base PR
merges, re-target any open PR stacked on it to `master` and drop
`fleet:stacked`, keeping the actual *merge* gated on non-wip. Concretely,
the `fleet:wip` exclusion in the stacked-PR collection at steps 2.5/2.6
needs a carve-out for the re-target-only path. #1488 is parked
`fleet:needs-human` for this remainder.

## Test plan

- [x] `scripts/fleet/tests/test_branch_matches_issue.py` — 7 new
  `IssuePrState` cases (active / parked-blocked / parked-unblocked / none /
  active-wins-over-parked-dup / game-legacy / malformed-record).
- [x] `scripts/fleet/tests/test_fleet_claim_parked_release.sh` — new
  gh-stubbed integration test: release clears labels only for a parked PR
  (keeps them for an active PR); `cleanup --gh` sweeps a parked claim, keeps
  an active claim, and preserves the no-PR sweep.
- [x] Full suite green: 208 python tests + existing fleet-claim bash suite
  (acquire / blockers / reconcile / reconcile_c2 / reconcile_amendments /
  stackable / model_gate).

## Follow-up (reuse, not blocking)

`cmd_reset_sweep_host_claims`'s inline `has_wip_pr` branch-match block
(~75% overlap with the new `_issue_pr_state_from`) could collapse to
`[[ "$(_issue_pr_state_from ...)" != "none" ]]` after adding `labels` to its
`gh pr list` fetch — deliberately left out here to keep this PR scoped to the
fleet-up reset-sweep-untouched #1488 surface.

Refs #1488 (Fixes A + B; Fix C carved out for human application)

### DIFF
--- a/docs/agents/AUTHOR-PIPELINE.md
+++ b/docs/agents/AUTHOR-PIPELINE.md
@@ -155,10 +155,16 @@ Paste the PR URL.
 > **Claim-label lifecycle.** `fleet-claim release` clears the local
 > filesystem lock and the worktree reservation. The issue's
 > `fleet:claim-<host>-<agent>` and `fleet:in-progress` labels are
-> **left in place** — they persist through the PR's review/merge
-> lifecycle and are swept by `fleet-claim cleanup --gh` only when the
-> issue closes or the claim goes stale with no matching open
-> `claude/<N>-*` PR. Don't hand-strip them.
+> normally **left in place** — they persist through the PR's
+> review/merge lifecycle and are swept by `fleet-claim cleanup --gh`
+> only when the issue closes or the claim goes stale with no matching
+> open `claude/<N>-*` PR. Don't hand-strip them. The one exception is
+> a **design-blocked / design-unblocked** escalation: when the issue's
+> matching PR is parked, `release` clears those two labels itself so
+> the scout stops treating the issue as in-progress and any worker can
+> re-claim once the architect unblocks it (#1488). You don't do
+> anything special — the design-escalation step's existing
+> `fleet-claim release <N>` call handles it.
 
 Then run the shared shutdown ceremony — see
 [`FLEET-RUNTIME.md § Per-iteration shutdown`](FLEET-RUNTIME.md#per-iteration-shutdown--final-step).
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -1058,6 +1058,61 @@ if url and head:
 PYEOF
 }
 
+# _issue_pr_state_from <prs-json> <issue-num> <repo>
+#   Echo the shared issue_pr_state() classification ("active" / "parked" /
+#   "none") for <issue-num> over a `gh pr list --json headRefName,labels` blob.
+#   Defaults to "none" on any parse/import failure so callers fail safe (treat
+#   the claim as abandoned rather than wedged-active).
+_issue_pr_state_from() {
+    echo "$1" | ISSUE_NUM="$2" REPO="$3" python3 -c '
+import json, os, sys
+sys.path.insert(0, os.environ["FLEET_LIB_DIR"])
+from fleet_branch_match import issue_pr_state
+try:
+    prs = json.load(sys.stdin)
+except Exception:
+    prs = []
+print(issue_pr_state(prs, os.environ["ISSUE_NUM"], os.environ.get("REPO", "")))
+' 2>/dev/null || echo none
+}
+
+# _release_parked_issue_labels <issue-num>
+#   When the issue's matching open PR is PARKED (fleet:design-blocked /
+#   fleet:design-unblocked), clear this host's `fleet:claim-<host>-*` and
+#   `fleet:in-progress` labels off the issue so the scout stops seeing it as
+#   in-progress and any worker can re-claim after the architect unblocks it
+#   (#1488 Fix A). Returns 0 only when a parked PR was detected and the sweep
+#   ran; returns 1 otherwise (gh missing, no GH repo for the namespace, or the
+#   matching PR is active — the normal PR-open release keeps its labels through
+#   the merge lifecycle, per AUTHOR-PIPELINE.md "Claim-label lifecycle").
+#   Multi-host safe: only this host's `fleet:claim-<host>-*` labels are touched.
+_release_parked_issue_labels() {
+    local issue_num="$1"
+    command -v gh >/dev/null 2>&1 || return 1
+    local repo host
+    repo=$(repo_from_ns)
+    host=$(derive_host)
+    [[ -z "$repo" ]] && return 1
+
+    local prs_json
+    prs_json=$(gh pr list --repo "$repo" --state open \
+        --json number,headRefName,labels --limit 300 2>/dev/null || echo "[]")
+
+    [[ "$(_issue_pr_state_from "$prs_json" "$issue_num" "$repo")" == "parked" ]] || return 1
+
+    # Remove this host's claim label(s) + fleet:in-progress (mirrors the
+    # cleanup --gh open-issue sweep).
+    local issue_labels lbl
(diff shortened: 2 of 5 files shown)


jakildev:
## Review — fleet/fleet-claim: clear parked-PR claim labels on release + sweep (#1488 Fixes A+B) (engine PR #1492)

**Verdict:** approve

Standalone diff; `baseRefName` is `master`. Five files changed: `fleet-claim` (bash), `fleet_branch_match.py` (Python), two test files, and `docs/agents/AUTHOR-PIPELINE.md`.

### Checklist

**ECS / render / ownership / math / serialization** — N/A. Fleet scripting (bash + Python) only; no C++ engine code.

**Logic correctness — Fix A (`cmd_release`):**
`_release_parked_issue_labels` is multi-host safe: it only removes `fleet:claim-${host}-*` labels matching `derive_host`, leaving any cross-host claim labels untouched. The guard `|| return 1` at every failure path ensures labels are never cleared on a `gh` error or parse failure. The `_issue_pr_state_from` bash wrapper's `|| echo none` fallback means a Python/import error defaults to "none" (sweepable), which is fail-safe: it would treat the claim as abandoned rather than wedge it forever as active. ✓

**Logic correctness — Fix B (`cmd_cleanup_gh` sweep):**
The old `has_wip_pr` Python block (matching on `headRefName` only) is cleanly replaced by `_issue_pr_state_from` (matching on `headRefName` + `labels`). The only behavior change for the common-path case is the addition of `labels` to the `gh pr list --json` fetch — active PRs still keep their claims. The "parked" case is new and correct: a design-blocked/-unblocked PR whose claim has aged past TTL is now treated as abandoned. ✓

**Shared classifier — `issue_pr_state()` in `fleet_branch_match.py`:**
The early-return-on-active logic is correct: iterating all matching PRs and returning "active" as soon as any non-parked match is found means an active PR always wins over a co-existing parked one. The `test_active_wins_over_parked_dup` test covers this exactly. The type handling is clean: `_norm_issue()` normalizes both integer and string `issue` inputs via `str().strip().lstrip("#")`, so the bash helper (passes string via env var) and Python tests (pass integers) both work correctly. ✓

**`PARKED_PR_LABELS` definition:**
`frozenset({"fleet:design-blocked", "fleet:design-unblocked"})` is correct and minimal. Placing it at module scope immediately before the function that uses it is idiomatic. The set covers exactly the two states in which a worker has explicitly yielded its active-work claim to allow re-pick-up after architect unblock. ✓

**Naming / style** — Consistent with surrounding code throughout. Bash helper prefix `_` for private functions is consistent with the rest of `fleet-claim`. Python constant `PARKED_PR_LABELS` is correctly UPPER_CASE for a module-level constant. Comments explain the "why" (cite #1488) without being tautological. ✓

**Tests:**
- 7 new `IssuePrState` unit tests — comprehensive: active, both parked variants, none, active-wins-over-parked-dup, game legacy prefix, malformed records (missing keys, None, string labels). ✓
- `test_fleet_claim_parked_release.sh` — two-phase integration test with fully stubbed `gh`. Phase 1 covers Fix A (release): issue #700 (parked PR) has labels cleared; issue #701 (active PR) has labels retained. Phase 2 covers Fix B (sweep): #710 (parked PR, design-unblocked) and #712 (no PR) are swept; #711 (active PR) is retained. TTL forcing via `FLEET_CLAIM_STALE_SECS_ISSUES=1` and the events-stub epoch timestamp is a solid test technique. ✓
- PR body confirms 208 Python tests + existing bash suite pass. ✓

**Docs — `AUTHOR-PIPELINE.md`:**
The update accurately documents the new exception: the claim-label retention rule now has one carved-out case (parked PR), and the update explains it without requiring the caller to do anything different (the escalation step's existing `fleet-claim release <N>` handles it automatically). ✓

**Fix C (carved out):** Correctly identified as gated by the deterministic self-config classifier. PR body gives the human the exact edit needed. ✓

### Praise

- **Fail-safe defaults throughout.** `_issue_pr_state_from`'s `|| echo none` fallback (fail to "sweepable") and `_release_parked_issue_labels`'s `|| return 1` guards (fail to "keep labels") are the right asymmetry: a transient gh outage never wedges the fleet.
- **`active-wins-over-parked-dup` test.** The scenario of two open PRs matching the same issue — one parked, one live — is subtle and exactly right to test explicitly. The comment in the test ("Any active matching PR means the claim is still live work → 'active'") captures the invariant clearly.
- **Shared classifier eliminates drift.** Moving the "is this PR live work?" question into `issue_pr_state()` in `fleet_branch_match.py` means `cmd_release` and `cmd_cleanup_gh` can never diverge on the definition of "active". The old `has_wip_pr` inline Python block could only become stale.

- [x] `python3 -m unittest discover -s scripts/fleet/tests -p "*.py"` — 208 tests pass (per PR body).
- [x] `bash scripts/fleet/tests/test_fleet_claim_parked_release.sh` — Fix A + Fix B integration tests pass (per PR body).
- [x] Existing fleet-claim bash suite (acquire / blockers / reconcile / etc.) — still green (per PR body).
- [ ] Apply Fix C: in `.claude/commands/role-merger.md` steps 2.5/2.6, carve out the re-target path so `fleet:wip` stacked PRs get re-targeted when their base merges, while the merge itself stays gated on non-wip.

Opus recheck not required.

Issue #1414, opened 2026-05-31
render: SDF shape face-loss / black holes under continuous camera Z-yaw (split from #1370)


jakildev:
**Blocked by:** #1370

> Split from #1370. #1370 fixes the **voxel** per-axis face-loss (snapped → continuous visible triplet, CPU-side). The **SDF** half is a *separate code path* and is tracked here so #1370 stays bounded. Best investigated after #1370 lands so the voxel loss doesn't confound the SDF repro.

## Symptom

Under continuous camera Z-yaw, SDF shapes (e.g. in `IRShapeDebug`) lose faces / show black holes, worst at the cardinal boundaries (45°/135°/225°/315°) — the same visual class as the voxel loss in #1370, but on the SDF path.

## Why this is NOT the #1370 voxel fix

The voxel per-axis path drops faces at the snapped-cardinal `visibleFaceIds` gate (`visibleFaceTripletCardinal` frozen across the ±45° bracket) — #1370 fixes that by resolving the triplet continuously via `IRMath::visibleTriplet`. **SDF has no such gate:** `c_shapes_to_trixel.glsl` solves the surface continuously under the full `visualYaw` (T5 #1361 made SDF rotate by full visualYaw, no triplet snap). So the SDF black-holes have a different root cause.

## Likely suspects (confirm during investigation)

- The SDF depth metric (`x+y+z`) or the analytic surface solve degenerating near the bracket where one world axis goes edge-on.
- The SDF iso-bounds / AABB extent used to bound the march at the bracket (note: #1370 ruled out the *shared cull viewport* as the cause for voxels by ballooning it — re-confirm for the SDF-specific extent).
- Depth-composite tie-breaking between SDF and voxel layers at the bracket.

## Acceptance

- Full Z-yaw sweep in `IRShapeDebug`: SDF shapes keep their faces across the sweep, especially at 45°/135°/225°/315°, on-center and off-center.
- `residualYaw==0` (cardinals) byte-identical.
- GLSL/Metal parity.

`[opus]` — continuous-yaw SDF surface/depth on the smooth-yaw composite.

jakildev:
Closed per architect direction on PR #1451 — premise disproved (SDF path is clean; the real artifact is voxel-pool scatter). Tracked in #1457.

PR #1496, opened 2026-06-03
render: make WASD camera panning screen-relative under Z-yaw

### LINKED ISSUE #1420: render: WASD camera panning should be screen-relative under Z-yaw (#1352 follow-up)
## Context

Escalated from PR #1362 (camera Z-yaw pivot).

The PR introduced a WASD yaw-compensation commit that attempted to keep WASD panning screen-relative by applying \`IRMath::rotate2D(velocity, yaw)\` inside \`system_render_velocity_2d_iso.hpp\`. Opus review identified this approach as fundamentally incorrect (details below) and the commit was reverted before merging #1362.

## Why the rotate2D approach fails

The camera's iso offset goes through \`getEffectiveCameraIso()\`, which applies the anisotropic iso projection matrix \`M(θ)\`. Because M(θ) is **not a rotation** (iso-x and iso-y scale into world space differently), \`rotate2D(v, θ) = R(θ)·v ≠ M(θ)⁻¹·v\`. The result:

- At yaw 45°: W key moves at 1.19× speed in wrong direction
- At yaw 90°: A key pans diagonally instead of straight
- At yaw 180°: W key **reverses** at ⅓ speed
- At θ = 120°/240°: det M(θ) = 0 — any per-frame undistortion blows up

Even the correct linear inverse \`M(θ)⁻¹·v\` hits a singularity at 120°/240°, so the "unrotate the velocity" framing is not a globally well-conditioned fix.

## Correct approach

**Store the camera pan as a world-space focus point and move *that* along screen-aligned iso axes**, deriving the iso offset once. Then WASD is a translation of the pivot in screen space (yaw-invariant by construction) — no \`M(θ)⁻¹\` and no 120°/240° singularity.

The issue body for #1352 already suggested this: "store the camera pan as a world-space focus point and move that along screen-aligned iso axes."

Concretely:
- Camera stores \`C_Position3D focus_point\` (world-space)
- WASD commands translate \`focus_point\` in screen-aligned world axes (not iso-pixel space)
- \`getEffectiveCameraIso()\` derives the iso offset from \`focus_point\` every frame
- \`system_render_velocity_2d_iso\` may need rethinking or may be replaced by direct focus-point manipulation

## Affected files

- \`engine/prefabs/irreden/render/systems/system_render_velocity_2d_iso.hpp\`
- \`engine/prefabs/irreden/input/commands/command_move_camera.hpp\`
- \`engine/render/include/irreden/render/render_manager.hpp\` / \`ir_render.cpp\` (camera focus state)
- Possibly \`engine/render/include/irreden/render/components/\` (new \`C_CameraFocusWorld\` or similar)

## Why escalating

This is an architectural change to how camera pan state is represented — storing world-space focus instead of iso-pixel offset is a model change that affects input commands, the render manager, and the effective-camera accessor. This is Opus-tier design work, not a mechanical tweak.

## Model

opus

## Area

\`engine/prefabs/irreden/render/\`, \`engine/render/\`

## Blocked by

#1352 (the camera pivot PR must merge first so the focus-point reframing builds on a clean base)

## EDIT

I actually merged the original PR but this work still needs to happen. I also left this comment, as we need to fix the vibrating of the voxel set entities: 
A couple things that should probably be fixed here, and some other that should likely stay deferred. The main issue is that the camera movement direction still does not match the WASD movement. Holding A should just make the camera move directly to the left, relative to the new angle. However, it seems to still be dependant on the rotation. Also, there is jittering when moving the camera after rotating. I am guessing it is the Pixel perfect correction, but might by the extra trixel buffer jitter also. But the calculations that handle the offset do not adjust with rotation. Here are two screenshots after just moving the camera with A (you can see we went diagonal). Then the last screenshot shows the start of clipping at a rotation (spinning either direction will make it disappear or reappear.).

Screenshot 2026-05-31 at 1 18 18 PM

Screenshot 2026-05-31 at 1 17 54 PM
Screenshot 2026-05-31 at 1 18 04 PM

jakildev:
## Summary
- WASD camera panning is now **screen-relative under camera Z-yaw**, matching the existing mouse / key-drag pan behavior. Holding A moves the camera straight left on screen at any yaw, instead of along a fixed iso diagonal.
- `RENDERING_VELOCITY_2D_ISO` now routes the camera's pan velocity through `IRMath::cameraMoveRelativeToYaw(velocity, panYaw)` before integrating. `panYaw` is the live camera yaw in `CAMERA_CENTER` pivot mode, `0` in `ORIGIN` mode.
- Yaw is snapshotted once per frame in `beginTick` (serialized on the main thread even under `PARALLEL_FOR`), so the parallel per-entity tick stays free of a `getComponent`. The camera is the only entity carrying `C_Velocity2DIso`.

## Background
The issue (escalated from #1362, written before #1352/#1459 landed) proposed a world-space focus-point rearchitecture. That architecture has since arrived implicitly: `CAMERA_CENTER` pivot mode + `IRMath::cameraMoveRelativeToYaw` (merged via #1352/#1459) already solve screen-relative panning, and `CAMERA_MOUSE_PAN` / `CAMERA_KEY_DRAG_PAN` already use them. WASD was the one pan path never wired to the helper. Wiring it makes this a one-line behavior change plus a once-per-frame yaw snapshot rather than a model change. The helper's `det = 1 + 2·cos(yaw)` singularity at ±120° degrades gracefully to raw-iso for that single frame — the same accepted behavior the drag systems already have.

## Verification
- Build clean (`IRShapeDebug`, macOS / Metal).
- **No static regression (proven byte-identical):** captured all 28 `shape_debug --auto-screenshot` outputs (21 shots incl. yawed + panned-then-yawed + ROI crops) against `origin/master` and this branch — **byte-identical across all 28** (`cmp`). Confirms the yaw-0 / `ORIGIN`-mode / static fast paths are preserved exactly.
- The behavioral delta (WASD direction under non-zero yaw) is **interactive-only** — the static screenshot harness has no key injection, so it cannot capture the change. Correctness rests on reusing the already-merged-and-proven `cameraMoveRelativeToYaw` helper that the mouse / key-drag pan systems use. The attached before/after pair is therefore byte-identical (expected).

## Out of scope
The issue's EDIT also notes camera-move jitter after rotation and shape clipping at rotation. Clipping is tracked separately (#1431 → merged as #1452, plus #1414 / #1430); the jitter is a pixel-perfect-snap interaction, a distinct concern. This PR is scoped to the titled WASD-screen-relative fix.

## Screenshots
<details>
<summary>zoom8_yaw180 (byte-identical — interactive-only fix)</summary>

| Before | After |
|--------|-------|
|  |  |

ROI crop:

| Before | After |
|--------|-------|
|  |  |

</details>

## Test plan
- [ ] In a scene with voxels/shapes, rotate the camera to a non-cardinal yaw, then hold A — camera moves straight left on screen (not along an iso diagonal). Repeat W / S / D.
- [ ] At yaw 0, WASD behavior unchanged.
- [ ] `--pivot-origin` (ORIGIN mode): WASD moves in raw iso (unchanged).

Closes #1420


jakildev (commented):
## Review — render: make WASD camera panning screen-relative under Z-yaw (engine PR #1496)

**Verdict:** approve

Standalone diff; `baseRefName` is `master`. One system header modified (`system_render_velocity_2d_iso.hpp` +28/-1) plus 4 byte-identical PNG screenshots (before/after are the same file — expected for an interactive-only fix).

### Checklist

**ECS / ownership / allocation** — Clean throughout.
- `panYaw_` is a member on `System<RENDERING_VELOCITY_2D_ISO>` (member-on-System form, correctly declared with trailing `_`). ✓
- `beginTick()` is `void()` — correct signature; fires once per pipeline execution on the main thread regardless of `PARALLEL_FOR`. ✓
- `IRRender::getRotationPivotMode()` and `IRPrefab::Camera::getYaw()` are both called in `beginTick` (not inside the per-entity `tick`) — no per-entity `getComponent` violation. ✓
- No allocation inside tick. ✓
- No structural changes inside tick. ✓

**Math / IRMath** — `IRMath::cameraMoveRelativeToYaw` is the correct helper (defined in `engine/math/include/irreden/ir_math.hpp:1030`). No raw `glm::` or `std::sin/cos` introduced. ✓

**Render pipeline** — No shader changes, no CPU/GPU struct changes, no bind-point drift. The behavior change is entirely CPU-side (velocity remapping before the integration step). ✓

**Correctness of the snapshot approach** — The PR body notes "the camera is the only entity carrying `C_Velocity2DIso`", so reading `panYaw_` once per frame in `beginTick` and using it in every `tick` call is equivalent to reading per-entity — zero drift risk. At `yaw = 0` the helper returns `isoDelta` unchanged (identity at zero); the `ORIGIN` pivot-mode path explicitly passes `panYaw_ = 0.0f`, preserving byte-identical behavior for both cases. ✓

**`±2π/3` singularity** — `cameraMoveRelativeToYaw` detects `|det| < 1e-6f` and returns `isoDelta` unchanged. This is the same graceful-degrade behavior the drag systems already accept. ✓

**Consistency with peer pan paths** — `CAMERA_MOUSE_PAN` and `CAMERA_KEY_DRAG_PAN` already route through `cameraMoveRelativeToYaw` with the same `pivotMode == CAMERA_CENTER ? getYaw() : 0.0f` guard. This PR closes the last gap. ✓

### Nits

- `system_render_velocity_2d_iso.hpp:18-33` — The 16-line comment block explains what the code does (WASD writes screen-aligned intent into `C_Velocity2DIso`, the helper remaps it, drag systems do the same). The non-obvious invariant worth keeping is the `beginTick` pattern and the ORIGIN-mode zero — one or two lines would suffice. Example: `// Snapshot yaw in beginTick so the PARALLEL_FOR tick stays free of getComponent; panYaw_=0 in ORIGIN mode preserves raw-iso behaviour.`

### Praise

- **Reuses the existing proven helper** rather than inlining the rotation math or duplicating the pivot-mode guard. The drag paths and WASD now share a single implementation.
- **`beginTick` snapshot is the right idiom.** Explicitly avoids the per-entity `getComponent` footgun in a `PARALLEL_FOR` system, and the PR body explains why (camera is the only entity carrying `C_Velocity2DIso`).
- **Byte-identical verification** at yaw=0 and ORIGIN mode — the 28-shot cmp result in the PR body is the appropriate evidence for a static harness that can't inject key events.

- [ ] At a non-zero yaw (e.g. 90°, 45°), hold A — camera moves left on screen, not along an iso diagonal.
- [ ] Repeat for W / S / D at various yaws.
- [ ] At yaw 0: WASD unchanged from pre-PR behaviour.
- [ ] `--pivot-origin` mode: WASD still moves in raw iso (panYaw_=0 path).
- [ ] Linux and Windows smoke (tagged below).

Opus recheck not required.

jakildev:
Addressed feedback: trimmed the 16-line comment block at lines 18-33 to the 2-line form the reviewer suggested — keeps only the non-obvious beginTick/PARALLEL_FOR invariant and the ORIGIN-mode zero. `fleet:approved` stays valid; no logic changed.

jakildev (commented):
## Opus re-review — render: make WASD camera panning screen-relative under Z-yaw (engine PR #1496)

**Verdict:** approve

Standalone PR (`baseRefName` = master). One render system header (`system_render_velocity_2d_iso.hpp` +28/-1) plus four byte-identical before/after PNGs. I took this for an Opus recheck because it changes a `PARALLEL_FOR` render tick path — exactly the "ECS invariant three systems deep + concurrency" surface the model split reserves for Opus — even though Sonnet (correctly) closed it "Opus recheck not required."

Sonnet flagged that the change reuses the proven `cameraMoveRelativeToYaw` helper and that the `beginTick` snapshot is the right idiom. **I confirm both, and I verified the one thing Sonnet asserted but did not prove: that `beginTick` actually runs serially, once, before the parallel tick fan-out** — if it didn't, `panYaw_` would stay `0.0f` and the whole fix would silently no-op.

### Opus-only items verified

- **`beginTick` ordering vs `PARALLEL_FOR` (the load-bearing claim).** `SystemManager::executeSystem` (`engine/system/src/system_manager.cpp:322`) calls `m_beginTicks[system].functionBeginTick_()` *before* the archetype-node loop and the `IRJob::parallelFor` fan-out (line 353+). So `panYaw_` is snapshotted on the main thread before any worker reads it. The snapshot updates every frame. The fix is live, not a no-op. ✓
- **No data race on `panYaw_`.** It is written only in `beginTick` (serial, single-threaded) and read-only inside the parallel `tick`. No worker writes it; no torn read. The member-on-`System<N>` placement is the blessed form per `cpp-systems.md`. ✓
- **Snapshot pattern is the documented idiom, not an improvisation.** `engine/system/CLAUDE.md:315` shows the identical shape (`void beginTick() { cachedMouse_ = IRRender::getMouse(); }`), and `engine/math/CLAUDE.md:90-98` documents this *exact* pivot-mode-gated `cameraMoveRelativeToYaw` recipe for pan systems. The PR is a faithful application of both. ✓
- **Identity at yaw 0 and in ORIGIN mode (byte-identity claim).** Hand-traced `cameraMoveRelativeToYaw(d, 0)`: `c=1, s=0, det=1+2=3`, returns `((1+2)·dx − 0)/3 = dx`, `3·(0 + dy)/3 = dy` — exactly the input. ORIGIN mode passes `panYaw_=0` so it collapses to the same identity. A static/cardinal scene stays byte-identical, consistent with the 28-shot cmp in the PR body. ✓
- **dt placement is correct under the remap.** Original `velocity_ * dt` became `cameraMoveRelativeToYaw(velocity_, panYaw_) * dt`. The helper is linear, so `remap(v)·dt == remap(v·dt)` — applying the remap to the velocity before the dt scale is equivalent and correct. ✓
- **Degenerate-yaw guard.** The helper's `det` singularity at yaw = ±2π/3 is handled inside the helper (returns `isoDelta` unchanged), so no NaN/blow-up leaks into `position.pos_`. Out of scope for this PR but worth confirming the call site inherits it — it does. ✓

### Sonnet's nit — agreed, optional

Sonnet's lone nit (the 16-line comment block is longer than needed; the load-bearing content is the `beginTick`/ORIGIN-zero rationale) is fair. The comment *does* document genuinely non-obvious semantics, so I'd not insist on it — keeping `fleet:has-nits` so the author can trim it to a couple of lines if they want, but it is not a merge blocker.

### Net

No correctness, concurrency, lifetime, or invariant issue. The fix is live (verified `beginTick` ordering), race-free, byte-identical on the non-yaw path by construction, and reuses the proven shared helper. Approving with `fleet:has-nits` retained for the optional comment trim. The interactive behaviour (hold A/W/S/D at non-zero yaw) is keyboard-only and headless-uncapturable — correctly left as the author's pre-merge manual check, alongside the already-tagged Linux/Windows smoke.

jakildev:
Addressed the `fleet:has-nits` review: the 16-line comment block at `system_render_velocity_2d_iso.hpp` was already trimmed to the two-line non-obvious invariant ("snapshot yaw in beginTick so the PARALLEL_FOR tick stays free of getComponent; panYaw_=0 in ORIGIN mode preserves raw-iso behaviour") in commit 748f1048, which is the current PR head. No further code change needed — cleared the stale `fleet:has-nits` label; `fleet:approved` stands. Cross-host linux/windows smoke still pending. — opus-worker-1

Issue #1499, opened 2026-06-04
render: detached canvas entities render as slivers when off world-origin (placement-phase composite bug)


jakildev:
**Model:** opus
**Area:** engine/prefabs/irreden/render/systems/system_entity_canvas_to_framebuffer.hpp, f_trixel_to_framebuffer.glsl (de-tile gather), v_peraxis_scatter_detached.glsl (P3b scatter)
**Repro:** IRCanvasStress (canvas_stress)
**Related:** epic #1444 (detached SO(3)), #1475 (P3b forward-scatter), #1466/#1498 (P5 baselines — a fix here MOVES these committed baselines)

## Symptom

In IRCanvasStress, detached entities render as **thin diagonal slivers with gaps instead of solid cubes** — except the one at world origin. The fringe (off-origin) entities glitch/fragment and never reconstruct the cube shape.

## Isolation (architect investigation, experiment-driven)

- The demo spawns 9 detached entities in a 3×3 grid at ±160 world units. **Only the entity at world (0,0) renders as a solid cube; every off-origin one is a sliver.**
- Reproduces at **±28 units** (not distance-graduated — any non-origin offset breaks immediately).
- Reproduces at **identity / no spin / yaw=0** (`--no-spin --no-auto-rotate`), so it is NOT the SO(3) deform.
- Breaks in **both** composite paths: the at-snap **gather blit** (`CanvasToFramebufferProgram` = `f_trixel_to_framebuffer`) AND the off-snap **P3b forward-scatter** (`PerAxisScatterDetachedProgram`).
- The **screen position is correct** (slivers land at the right spokes); only the **content** is broken.
- **Ruled out:** `fd.cameraTrixelOffset_` — zeroing it (`= vec2(0)` instead of `-entityIso`) changed nothing.

The cube is canvas-**local** centered (`C_LocalTransform{vec3(0)}`), so all 9 detached canvases have identical content; world position must not affect the canvas content at all — yet it does. The only thing that varies per entity is the **framebuffer placement** (`entityFbCenter` in system_entity_canvas_to_framebuffer.hpp:120-124). Both composite paths sample/scatter relative to that placement.

## Root-cause hypothesis

[...]

Issue #1494, opened 2026-06-03
render: detached SO(3) forward-scatter produces thin vertical sliver artifacts at certain spin poses (IRCanvasStress)


jakildev:
**Area:** engine/render
**Model:** opus
**Blocked by:** #1499

## Context
Surfaced during P4 (#1465) Metal-parity verification of the detached SO(3) forward-scatter (P3b / #1475). Running `IRCanvasStress --auto-screenshot <warmup>` and sweeping the warmup count (which advances the detached `C_AutoSpin` ring to different rotation poses), some detached entities render thin vertical red/blue/green **sliver** artifacts — lines that stretch well beyond the entity's footprint — instead of a clean deformed solid. The artifact is **pose-dependent**: it appears on a given entity only at certain residual-quaternion poses and **migrates between entities** as they spin (e.g. at warmup 120 the bottom entity slivers; at warmup 220 it is clean and two near-center entities sliver instead). Reproduces on `origin/master` (independent of P4's docs-only change).

This has the signature of a degenerate scatter footprint — a per-axis canvas cell's face quad whose corners land at a near-edge-on / stretched screen position at specific `detachedResidual` quaternions — rather than clean edge-on geometry. The camera-yaw scatter bounds this with a min-on-screen-trixel-size cap (#1307/#1310); the detached scatter (`v_peraxis_scatter_detached` / `metal/peraxis_scatter_detached.metal`) may need the same cap or a degenerate-quad guard.

The GLSL and Metal scatter shaders are faithful mirrors, so this is **expected to reproduce on OpenGL too** (shared algorithm, not a Metal-specific parity bug) — the cross-host Linux smoke on the P4 PR (#1493) should confirm. If OpenGL is clean and only Metal slivers, re-scope as a parity bug.

## Acceptance criteria
- `IRCanvasStress` detached SO(3) ring renders no sliver/stretched-quad artifacts across a full SO(3) spin (sweep `--auto-screenshot` warmup, or capture a continuous-spin sequence) on both OpenGL and Metal.
- Identity / at-octahedral-snap poses remain byte-identical (the per-axis path is released at a snap; the cap must not perturb the fast path).

jakildev:
## Triaged + approved (architect)

[...]

Issue #1462, opened 2026-06-02
render: detached SO(3) P1 — per-voxel occlusion depth on the entity-rotated iso axis


jakildev:
**Model:** opus
**Part of epic:** #1444
**Plan file:** `~/.fleet/plans/issue-1444.md` (full epic plan + resolved design decisions)

## Scope
Re-derive per-voxel occlusion depth for DETACHED entities on the entity-rotated iso axis `R⁻¹·(1,1,1)` instead of model-space `x+y+z` (correct only at the 24 octahedral snaps today). Off-cardinal pitch/roll currently keeps the snapped order → the "pitch/roll reveals" bug, live now that #1386/#1398 landed full-SO(3) face selection on detached canvases.

## Approach
- Add an `IRMath` helper for the entity-rotated iso-depth axis (`R⁻¹·(1,1,1)` projection), CPU+GPU bit-identical — reused by P3.
- Plumb entity rotation `R` (or the pre-rotated depth axis) into stage-1: extend `FrameDataVoxelToCanvas` (CPU struct) + the GLSL **and** Metal stage-1 shaders (`c_voxel_to_trixel_stage_1.glsl` ~:211/:238). CPU↔GPU contract change.
- Amend `docs/design/iso-depth-axis-invariant.md`: carve out **entity** rotation (gated per-entity only); fix the stale `#1076 … A/B pending` line (#1076 is closed, Option B).

## Acceptance
- Rotated detached entities occlude correctly across a full SO(3) spin (the discriminating off-snap pitch shot).
- Identity (zero-rotation) detached entity byte-identical to master; camera-Z-yaw GRID output byte-identical.
- GLSL/Metal parity crops match.

## References
`system_voxel_to_trixel.hpp:96-133`, `c_voxel_to_trixel_stage_1.glsl:211/238`, `ir_math.hpp:513` (`visibleTriplet`), `docs/design/iso-depth-axis-invariant.md`. Symmetric with `visibleTriplet`'s `R⁻¹·viewDir`.

Issue #1501, opened 2026-06-04
render: black faces on rotating cubes (detached scatter unlit + suspected face-normal/lighting issue)


jakildev:
**Model:** opus
**Area:** detached scatter lighting (engine/prefabs/irreden/render/systems/system_entity_canvas_to_framebuffer.hpp scatter pass), v_peraxis_scatter*.glsl, and possibly the main-canvas voxel face-normal/lighting path
**Repro:** IRCanvasStress (canvas_stress)
**Related:** #1375 (detached scatter receive-lighting), #1376 (detached scatter cast-lighting), epic #1444
**Independent of** #1499 / #1500 (different subsystem — lighting, not placement; can run in parallel)

## Symptom

Many cubes in IRCanvasStress show **black faces** — some or all faces of a cube render fully black. It's a recurring pattern across both the GRID-spin cubes and the detached entities, and it makes it hard to visually validate whether the attached rotating cubes are even behaving correctly (#1499/#1500 validation is partly blocked by this).

User note: "If it is just lighting, then change it so it isn't the case, but I don't think that is true" — i.e. confirm whether this is purely the known unlit-scatter gap or whether there's a real face-normal / lighting-direction bug producing black faces on lit content too.

## What's already known

The detached scatter composite is **documented "UNLIT by design today"** in `system_entity_canvas_to_framebuffer.hpp` — the P3b forward-scatter path does not yet receive or cast lighting. That work is filed as:
- #1375 — detached scatter **receives** lighting (AO / sun-shadow sampling)
- #1376 — detached scatter **casts** lighting

So the detached-entity black faces are *expected* until #1375 lands. **This issue should first confirm that** — capture which black faces are detached-scatter (explained by #1375) vs. which are on lit GRID/attached content (a real bug).

## The part that may be a real bug

If GRID-spin cubes (which go through the normal lit voxel path, not the unlit scatter) also show black faces, that's NOT explained by #1375/#1376. Candidate causes to investigate:
- Face-normal sign flip after SO(3) deform — a rotated face whose normal points away from the light gets zero diffuse and reads as black (no ambient floor).
- Missing ambient/minimum-light term so back-faces clamp to 0 instead of an ambient floor.
- Exposed-face mask / visible-face-triplet selecting a face that has no lighting data written.

## Scope

[...]

PR #1502, opened 2026-06-04
render: stage-2 detached depth-axis parity + canvas_stress --no-auto-rotate clobber fix

jakildev:
Two small render-correctness fixes that came out of the #1499 investigation. Per the architect's design call on this PR, **#1499 is closed as a misdiagnosis**: at TRUE identity (camera yaw=0) the gather/single-canvas path renders off-origin detached cubes correctly. The "off-origin identity sliver" premise was a compositing illusion (main-grid bleed-through) compounded by a CLI-flag clobber that ran every "identity" repro at a drifting ~30° yaw. The real sliver bug is the pose-dependent forward-scatter degeneracy tracked by #1494.

Both fixes below are byte-safe and correct regardless of that scoping — they stand on their own.

## Fix 1 — canvas_stress `--no-auto-rotate` clobber (`creations/demos/canvas_stress/main.cpp`)
`readConfig()` runs after `parseArgs()` (it needs `IREngine::init`), so a `config.lua` `auto_rotate=true` silently overwrote an explicit `--auto-rotate` / `--no-auto-rotate` flag. Every "identity / --no-auto-rotate" repro was actually running at a drifting camera yaw. Latch CLI intent (`autoRotateSetByCli_`) so config only supplies the default when the flag is absent (CLI overrides config). This is both a real demo bug and the #1494 repro enabler.

## Fix 2 — stage-2 detached depth-axis parity (`c_voxel_to_trixel_stage_2.{glsl,metal}`)
Stage 2 didn't mirror stage 1's #1462 entity-rotated occlusion axis: it computed occlusion depth on the fixed `(1,1,1)` iso axis while stage 1 wrote the distance tap on `voxelDepthAxis = R⁻¹·(1,1,1)`. Off a snap, the stage-2 depth re-test would reject every detached color tap. Stage 2 now reads the same binding-7 `voxelDepthAxis` uniform — already present in the CPU `FrameDataVoxelToCanvas` struct at offset 144 (static-asserted) and stage 1's UBO; Metal shares it via `ir_iso_common.metal` — and projects via `isoDepthAlongAxis` on the detached path. The world/GRID path keeps `pos3DtoDistance` / `x+y+z`, so this is byte-identical today (the gather only runs inside the 1e-4 deadband where the axis ≈ (1,1,1)): a latent correctness fix.

## Verification
- `fleet-build --target IRCanvasStress` clean (macOS/Metal).
- `fleet-run IRCanvasStress --auto-screenshot` — full 5-shot render loop, Metal stage-2 kernel compiles at runtime, no crash.
- Architect validated this branch on macOS/Metal: a single off-origin detached cube renders solid at yaw=0 (gather path correct); slivers reproduce only on the scatter path (#1494's domain).

Refs #1499 (closed — misdiagnosis), Refs #1494 (repro enabler)

### DIFF
--- a/creations/demos/canvas_stress/main.cpp
+++ b/creations/demos/canvas_stress/main.cpp
@@ -82,6 +82,11 @@ struct CanvasStressSettings {
     bool fullRotate_ = false;
     bool noSpin_ = false;
     bool noLighting_ = false;
+    // readConfig() runs AFTER parseArgs() (it needs IREngine::init), so a
+    // config `auto_rotate` would otherwise clobber an explicit
+    // --auto-rotate / --no-auto-rotate flag. Latch CLI intent so config only
+    // supplies the default when the flag is absent (CLI overrides config).
+    bool autoRotateSetByCli_ = false;
 };
 
 // 0.5 degrees per frame → full revolution in ~720 frames (~12 s at 60 fps)
@@ -210,7 +215,7 @@ void readConfig() {
     if (zoom.is<float>())
         g_settings.initialZoom_ = zoom.as<float>();
     sol::object autoRotate = table["auto_rotate"];
-    if (autoRotate.is<bool>())
+    if (autoRotate.is<bool>() && !g_settings.autoRotateSetByCli_)
         g_settings.autoRotate_ = autoRotate.as<bool>();
 }
 
@@ -222,8 +227,10 @@ void parseArgs(int argc, char **argv) {
             ++i;
         } else if (std::strcmp(argv[i], "--auto-rotate") == 0) {
             g_settings.autoRotate_ = true;
+            g_settings.autoRotateSetByCli_ = true;
         } else if (std::strcmp(argv[i], "--no-auto-rotate") == 0) {
             g_settings.autoRotate_ = false;
+            g_settings.autoRotateSetByCli_ = true;
         } else if (std::strcmp(argv[i], "--full-rotate") == 0) {
             g_settings.fullRotate_ = true;
         } else if (std::strcmp(argv[i], "--no-spin") == 0) {
--- a/engine/render/include/irreden/render/ir_render_types.hpp
+++ b/engine/render/include/irreden/render/ir_render_types.hpp
@@ -225,7 +225,7 @@ struct FrameDataVoxelToCanvas {
     // (1,1,1). `.w` is std140 padding. std140-appended after visibleFaceIds_
     // (offset 144), so every prior field offset — and the world/GRID fast
     // path — is unchanged; the gather shaders that read only the prefix
-    // (stage 2, AO, lighting) are unaffected and need no declaration update.
+    // (AO, lighting) are unaffected and need no declaration update.
     vec4 voxelDepthAxis_ = vec4(1.0f, 1.0f, 1.0f, 0.0f);
 };
 
--- a/engine/render/src/shaders/c_voxel_to_trixel_stage_2.glsl
+++ b/engine/render/src/shaders/c_voxel_to_trixel_stage_2.glsl
@@ -28,6 +28,13 @@ layout(std140, binding = 7) uniform FrameDataVoxelToTrixel {
     uniform vec4 faceDeform[3];
     // Per-slot world FaceId (0..5). See stage 1 + #1278 for the contract.
     uniform ivec4 visibleFaceIds;
+    // Model-frame iso depth axis `R⁻¹·(1,1,1)` for the per-voxel occlusion
+    // metric (#1462). Stage 2 MUST read the same axis stage 1 wrote the
+    // distance tap on, or the depth re-test below rejects every detached
+    // color tap off a snap (the #1499 sliver). (1,1,1) for world / identity
+    // so the GRID path stays byte-identical. Appended after visibleFaceIds
+    // (offset 144) to match the CPU struct + stage 1's binding-7 layout.
+    uniform vec4 voxelDepthAxis;
 };
 
 layout(std430, binding = 5) readonly buffer PositionBuffer {
@@ -173,8 +180,14 @@ void main() {
             voxelPositionInt = rotateCardinalZ(voxelPositionInt, cardinalIndex);
             voxelPositionInt += cardinalLowerCornerShift(cardinalIndex);
         }
-        const int voxelDistance = encodeDepthWithFace(
-            pos3DtoDistance(voxelPositionInt), slot);
+        // Detached entities project occlusion depth onto the entity-rotated
+        // iso axis (#1462); world/GRID keeps the fixed (1,1,1) via
+        // pos3DtoDistance. MUST mirror stage 1's distance-tap depth or the
+        // re-test in writeColorTap rejects the tap (#1499).
+        const int rawDepth = isDetachedCanvas > 0.5
+            ? isoDepthAlongAxis(voxelPositionInt, voxelDepthAxis.xyz)
+            : pos3DtoDistance(voxelPositionInt);
+        const int voxelDistance = encodeDepthWithFace(rawDepth, slot);
         const ivec2 base =
             trixelFrameOffset(trixelCanvasOffsetZ1, frameCanvasOffset, voxelRenderOptions) +
             pos3DtoPos2DIso(voxelPositionInt);
@@ -199,8 +212,11 @@ void main() {
         // `voxelPositionFixed = round(worldPos * subdivisions)`.
         microPositionFixed += cardinalLowerCornerShift(cardinalIndex) * subdivisions;
     }
-    const int depthBase =
-        microPositionFixed.x + microPositionFixed.y + microPositionFixed.z;
+    // Detached: mirror stage 1's entity-rotated occlusion axis (#1462 / #1499);
+    // depth is in subdivision units on both branches so the encode is unchanged.
+    const int depthBase = isDetachedCanvas > 0.5
+        ? isoDepthAlongAxis(microPositionFixed, voxelDepthAxis.xyz)
+        : (microPositionFixed.x + microPositionFixed.y + microPositionFixed.z);
     const int voxelDistance = encodeDepthWithFace(depthBase, slot);
     const ivec2 base = frameOffsetFixed + pos3DtoPos2DIso(microPositionFixed);
     emitDeformedFace(base, D, voxelDistance, voxelColor, voxelIndex);
--- a/engine/render/src/shaders/metal/c_voxel_to_trixel_stage_2.metal
+++ b/engine/render/src/shaders/metal/c_voxel_to_trixel_stage_2.metal
@@ -190,9 +190,14 @@ kernel void c_voxel_to_trixel_stage_2(
             voxelPositionInt = rotateCardinalZ(voxelPositionInt, cardinalIndex);
             voxelPositionInt += cardinalLowerCornerShift(cardinalIndex);
         }
-        const int voxelDistance = encodeDepthWithFace(
-            pos3DtoDistance(voxelPositionInt), slot
-        );
+        // Detached entities project occlusion depth onto the entity-rotated
+        // iso axis (#1462); world/GRID keeps the fixed (1,1,1) via
+        // pos3DtoDistance. MUST mirror stage 1's distance-tap depth or the
+        // re-test in writeColorTap rejects the tap (#1499).
+        const int rawDepth = frameData.isDetachedCanvas > 0.5f
+            ? isoDepthAlongAxis(voxelPositionInt, frameData.voxelDepthAxis.xyz)
+            : pos3DtoDistance(voxelPositionInt);
+        const int voxelDistance = encodeDepthWithFace(rawDepth, slot);
         const int2 base =
             trixelFrameOffset(
                 frameData.trixelCanvasOffsetZ1,
@@ -235,8 +240,11 @@ kernel void c_voxel_to_trixel_stage_2(
         microPositionFixed = rotateCardinalZ(microPositionFixed, cardinalIndex);
         microPositionFixed += cardinalLowerCornerShift(cardinalIndex) * subdivisions;
     }
-    const int depthBase =
-        microPositionFixed.x + microPositionFixed.y + microPositionFixed.z;
+    // Detached: mirror stage 1's entity-rotated occlusion axis (#1462 / #1499);
+    // depth is in subdivision units on both branches so the encode is unchanged.
+    const int depthBase = frameData.isDetachedCanvas > 0.5f
+        ? isoDepthAlongAxis(microPositionFixed, frameData.voxelDepthAxis.xyz)
+        : (microPositionFixed.x + microPositionFixed.y + microPositionFixed.z);
     const int voxelDistance = encodeDepthWithFace(depthBase, slot);
     const int2 base = frameOffsetFixed + pos3DtoPos2DIso(microPositionFixed);
     emitDeformedFace(


jakildev:
## NEEDS-DESIGN

Investigated #1499 on macOS/Metal (IRCanvasStress, 3×3 detached grid, `--no-spin --no-auto-rotate`, zoom 1.0). The repro is confirmed: at identity, the 8 off-origin detached entities render as thin diagonal **slivers**; only the world-origin entity is a solid cube. While spinning, off-origin entities become **fragmented blocks**. **But the issue's root-cause hypothesis and Fix-direction (a) are both empirically invalidated.** Handing back for a re-scope.

### What I ruled out (with experiments)

1. **Fix (a) "route all detached voxel content through forward-scatter" does NOT fix it.** I forced the per-axis canvases to allocate at the snap (set `kDetachedResidualDeadband = -1`) so the P3b forward-scatter runs at identity for every detached entity. Result: off-origin entities are **identical slivers** to the gather. The forward-scatter is *not* placement-robust off-origin — so rerouting voxels through it cannot fix the bug, and re-blessing P5 baselines around it would lock in the slivers.

2. **The de-tile gather / parity is innocent (at least on Metal).** The Metal `f_trixel_to_framebuffer` already samples the canvas at the RAW interpolated position (`originRaw`, no `trixelFramebufferSamplePosition` shift for color — pre-#394 behavior). Off-origin still slivers. So the GLSL single-global-parity de-tile hypothesis in the issue does not apply on this backend. (Note for whoever picks this up on macOS: the `.glsl` shaders are *not* loaded — backend is Metal; edit `engine/render/src/shaders/metal/*.metal`.)

3. **CPU placement is correct.** I logged the full placement math per entity. `entityFbCenter` lands exactly where the slivers appear (e.g. world-origin → fb (640,362)=center; world (0,-160)→(320,202); world (160,0)→(1280,362)). `mainCanvasSize` is non-square (642×722 vs fb 1284×722), so `pixelsPerTexel = (2,1)`, but that's identical for all entities including the solid world-origin one. No CPU miscompute.

4. **Detached cube voxels are model-(0,0,0).** `UPDATE_VOXEL_SET_CHILDREN` writes them at `C_WorldTransform.translation_` = 0 (the voxel-set entity has `C_LocalTransform{0}`, no parent). So all 9 canvases *should* hold byte-identical content, and the per-canvas voxel cull (camera-centered iso 0, line `system_voxel_to_trixel.hpp:480`) does not clip a model-0 cube.

### The paradox (why this needs GPU instrumentation, as the issue predicted)

Every source-level path says the per-canvas rasterization, the cull, the de-tile, and the composite are all **position-invariant** (model-0 content, camera-centered everything, `TexCoords`-driven sampling). Yet the rendered result is unambiguously **position-dependent** — only the world-origin entity is solid. The single shared input that varies per entity is `entityIso = pos3DtoPos2DIso(worldTransform.translation_)`, which feeds (a) `entityFbCenter` placement and (b) `fd.cameraTrixelOffset_ = -entityIso`. The architect already ruled out (b) by zeroing it. That leaves the placement geometry collapsing the content to a sliver in a way that is not visible from the source — it needs a **per-texel GPU dump** of what each off-origin fragment actually samples vs. what the world-origin fragment samples (the issue's own "Next step to pin it"). On Metal this is awkward because the gather and scatter are separate programs and the main-canvas blit shares the same fragment shader (a blanket debug-output magenta-fies the main canvas too), so isolating a single off-origin detached cube cleanly needs a dedicated debug path or a `detached_count=1` off-origin scene.

### Design question for the architect

Given (1) the forward-scatter — the design doc's "robust composite" — also slivers off-origin, and (2) the content is provably model-identical yet renders position-dependent: **is the detached composite's `entityFbCenter` placement geometrically consistent with how the per-canvas content is rasterized/sampled, or do the two disagree off-origin?** The fix likely isn't "pick gather vs scatter" (both fail) — it's in the placement↔rasterization contract for off-origin detached canvases. I did not want to redesign that contract mid-task as a headless worker.

WIP branch has no fix commit (all instrumentation reverted); the findings above are the handoff. Releasing the claim.

— opus-worker

jakildev:
Cleared `fleet:design-blocked`. Architect re-scope posted on #1499.

Your handoff was correct to refuse the original fix-direction. I read the composite + raster source: the gather composite is provably **position-independent** (color/depth driven only by `texCoords`; `cameraTrixelOffset_` feeds only hover), and the per-canvas raster input is also position-independent (global camera offset; demo voxel-set unparented at `C_LocalTransform{0}`). So the sliver is a **GPU-execution effect**, not the placement↔de-tile-parity class the original issue blamed — and forward-scatter can't fix it (both paths composite the same already-broken content).

The re-scope replaces "route through forward-scatter" with two cheap discriminating experiments (single off-origin `detached_count=1`; direct canvas-texture readback). Whoever resumes can build on this branch or start fresh from the re-scoped issue.

jakildev:
## NEEDS-DESIGN

Ran the two experiments from the re-scope on macOS/Metal. The result **overturns #1499's premise** and lands on a scoping decision I shouldn't make unilaterally (it re-partitions #1499 ↔ #1494).

### What the experiments showed (screenshot-backed)

**Experiment 1 — single detached entity, off-origin, nothing else in the scene:**
- At **TRUE identity (camera yaw = 0)** an off-origin detached cube (world (80,0,0)) renders as a **SOLID iso cube**, not a sliver. The gather/single-canvas path is **position-independent and correct** at identity. Per the re-scope's own discriminator ("renders solid ⟹ NOT a per-entity bug"), this rules out the per-entity placement/raster bug the issue blamed.
- The full 3×3 grid with the main grid **removed** shows **all 9 detached entities slivering** (including the center/origin one). In the default demo the origin entity only *looks* solid because the **main grid's origin cube + grid-spin cubes bleed through** the transparent detached sliver composited over them. So the "only the world-origin entity is solid" observation was a compositing illusion, not a real per-entity distinction.

**Why every prior "identity" repro was wrong (root cause of the misdiagnosis):**
- `canvas_stress` silently **ignores `--no-auto-rotate`**: `main()` runs `parseArgs()` then `readConfig()`, and `readConfig` overwrites `autoRotate_` from `config.lua` (`auto_rotate=true`). So `--no-spin --no-auto-rotate` runs were actually at a **drifting ~30° camera yaw**. A debug dump confirmed the detached `C_CanvasLocalRotation` drifting off identity frame-by-frame (`rot.z` = 0 → -0.0044 → -0.0087 …, `voxelDepthAxis` (1,1,1) → (0.991,1.009,1) …).
- This PR fixes that clobber (CLI now overrides config), which is what made TRUE identity reproducible.

**Where the slivers actually come from:**
- The deadband `kDetachedResidualDeadband = 1e-4`. So **any** residual rotation > 1e-4 switches a detached entity from the gather single-canvas path to the **per-axis forward-SCATTER path** (#1475 P3b). The slivers reproduce **only on the scatter path** (yaw 0 → solid; yaw 30° → sliver; yaw 45° → entity vanishes). That is the **pose-dependent scatter degeneracy you assigned to #1494**, not the gather/placement bug #1499 describes.

### Fixes already in this PR (both byte-safe, both correct regardless of scoping)
1. **stage-2 detached depth-axis parity** (`c_voxel_to_trixel_stage_2.{glsl,metal}`): stage 2 didn't mirror stage 1's #1462 entity-rotated occlusion axis. Latent correctness fix; byte-identical today because the gather path only runs inside the 1e-4 deadband where the axis ≈ (1,1,1). Splittable into its own PR if you prefer.
2. **`--no-auto-rotate` clobber fix** in canvas_stress (the repro enabler).

### Design / scoping question
Given that the gather path is correct at identity and the slivers are the scatter path (#1494's domain, which you explicitly asked to keep separate from #1499):

- **(a)** Is #1499 effectively a **misdiagnosis to close** (gather is fine; the real bug is the scatter path = #1494), with this PR landing the two correctness fixes above? — my recommendation.
- **(b)** Or do you want the scatter-path off-origin/pose fix **folded into #1499** here? That is the placement↔rasterization-contract redesign for the forward-scatter (`v_peraxis_scatter_detached` + the per-axis store + `entityFbCenter`/`perAxisBase` placement) you declined to do blind last iteration — I'd want your steer on the contract before touching it, since it overlaps #1494/#1501.

Releasing the claim; `fleet:design-blocked`. WIP commit pushed.

— opus-worker

jakildev:
## Architect direction — option (a): #1499 is a misdiagnosis; land the two fixes here

Outstanding diagnosis — this is exactly the experiment the re-scope asked for, and it overturns the premise cleanly (including my own source analysis: it was right that the gather is position-independent, but wrong to trust the "only world-origin solid" observation — that was the compositing illusion you caught, plus the `--no-auto-rotate` clobber). Going with **(a)**.

**Decisions:**
1. **#1499 closed as a misdiagnosis.** At TRUE identity the gather/single-canvas path renders off-origin detached cubes correctly. The "off-origin identity sliver" premise was a compositing illusion (main-grid bleed-through) compounded by the `--no-auto-rotate` clobber that ran every "identity" repro at a drifting ~30° yaw. No gather/placement bug exists.
2. **The real bug is #1494** — the pose-dependent forward-scatter degeneracy (residual > `kDetachedResidualDeadband` 1e-4 → scatter path → slivers/vanish). Unblocked now; it's the primary detached-rotation rendering bug.
3. **This PR lands its two correctness fixes** and stands on them (both correct regardless of scoping): the `--no-auto-rotate` clobber fix (real demo bug + #1494 repro enabler) and the stage-2 detached depth-axis parity fix. Keep them together (both small render-correctness) unless the reviewer prefers a split. **Update the PR body**: `Closes #1499` → `Refs #1499 (closed — misdiagnosis), Refs #1494 (repro enabler)`, and retitle to match the two fixes.

**Not doing (b)** (folding the scatter redesign in here) — the scatter degeneracy is #1494's domain, overlaps #1501, and deserves a focused task with a contract decision, not a blind bolt-on.

Unblocking now (design-unblocked). Thanks for the rigor — the flag-clobber root cause is what unlocked the whole thread.

jakildev:
Resumed from `fleet:design-unblocked` (architect option (a) on this PR). #1499 closed as a misdiagnosis; this PR now lands the two byte-safe correctness fixes and stands on them:
- canvas_stress `--no-auto-rotate` clobber fix (CLI now overrides `config.lua`; real demo bug + #1494 repro enabler)
- stage-2 detached depth-axis parity (reads the shared binding-7 `voxelDepthAxis`, mirrors stage 1's #1462 occlusion axis on the detached path; world/GRID byte-identical)

Retitled + rewrote the body (`Closes #1499` → `Refs #1499 (closed — misdiagnosis), Refs #1494 (repro enabler)`). Build clean and `--auto-screenshot` runs the full Metal render loop without crashing. Cleared `fleet:wip` — ready for review. — opus-worker-1

jakildev (commented):
## Review — render: stage-2 detached depth-axis parity + canvas_stress --no-auto-rotate clobber fix (engine #1502)

**Verdict:** approve

### Praise

- **Scoping call is correct.** The diagnostic chain from the WIP branch through the two experiments (single off-origin entity at true identity → solid cube; full 3×3 without main grid → all slivers) is exactly what was needed to distinguish the gather/scatter paths. Closing #1499 as a misdiagnosis is the right outcome.
- **CPU/GPU layout chain is airtight.** `FrameDataVoxelToCanvas::voxelDepthAxis_` was already statically asserted at offset 144 / size 160 (from #1462), and the GLSL UBO appends the field at exactly that position. Stage 2 now mirrors stage 1's existing pattern without any new invariant surface to audit.
- **Both fixes are byte-safe today.** The world/GRID branch keeps `pos3DtoDistance`; the detached gather path at identity uses `isoDepthAlongAxis(pos, (1,1,1))` which collapses to `x+y+z` — provably identical. The scatter path (outside the 1e-4 deadband) is tracked separately by #1494 and is not touched here.
- **`autoRotateSetByCli_` pattern is idiomatic.** Latching CLI intent before the config-read runs is the correct fix for the flag-clobber; the naming is consistent with the struct's existing trailing-underscore convention.

### Nits

- `engine/render/include/irreden/render/ir_render_types.hpp:228` — the comment on `voxelDepthAxis_` ends with "the gather shaders that read only the prefix (stage 2, AO, lighting) are unaffected and need no declaration update." That was accurate when #1462 wrote the field; this PR updates stage 2 to read it, making the parenthetical stale. Drop "stage 2" so it reads "(AO, lighting)" after this lands.

- [x] `fleet-build --target IRCanvasStress` clean (macOS/Metal) — done per PR body.
- [x] `fleet-run IRCanvasStress --auto-screenshot` — full 5-shot render loop without crash — done per PR body.
- [x] Architect verified single off-origin detached cube renders solid at yaw=0 on macOS/Metal — done.
- [ ] Linux OpenGL build + smoke will land via `fleet:needs-linux-smoke`.
- [ ] Windows build will land via `fleet:needs-windows-smoke`.

Opus recheck not required.

jakildev:
Addressed feedback: dropped "stage 2" from the `voxelDepthAxis_` prefix comment in `ir_render_types.hpp:228` — the parenthetical now reads "(AO, lighting)" since this PR updated stage 2 to read the field. Pure comment edit, no code change. — sonnet-author (sonnet-fleet-2)

jakildev (commented):
## Opus recheck — render: stage-2 detached depth-axis parity + canvas_stress `--no-auto-rotate` clobber fix (engine #1502)

**Verdict:** approve (with one nit) — agreeing with Sonnet.

Sonnet approved and marked "Opus recheck not required." I rechecked anyway because this touches the binding-7 CPU↔GPU UBO contract and the stage-1/stage-2 depth re-test invariant — the Opus-only surface. Sonnet's analysis holds up; below is what I independently verified beyond the diff hunks.

### Confirmed (Opus-only items)

- **The std140 layout chain is airtight on all three sides.** `FrameDataVoxelToCanvas::voxelDepthAxis_` is pinned by `static_assert` at offset 144 / size 160 (`ir_render_types.hpp:415,421`). The GLSL stage-1 binding-7 block already declares `vec4 voxelDepthAxis` immediately after `ivec4 visibleFaceIds` (offset 144); stage 2's newly-added declaration is byte-identical to it. Sonnet asserted this; I confirmed it against the live stage-1 source.
- **The Metal field is shared, not redeclared — so no parity divergence is possible.** `frameData.voxelDepthAxis` in stage_2.metal resolves to `float4 voxelDepthAxis` in the *shared* `FrameDataVoxelToTrixel` struct at `ir_iso_common.metal:603` (positioned right after `int4 visibleFaceIds`), which stage_2.metal `#include`s. Both kernels read one struct definition, so the GLSL↔Metal and CPU↔GPU offsets cannot drift apart for this field. This is the invariant most worth an Opus pass and it's clean.
- **`isoDepthAlongAxis` is in scope on both backends.** GLSL: defined `ir_iso_common.glsl:302`, and stage_2.glsl `#include`s `ir_iso_common.glsl`. Metal: defined `ir_iso_common.metal:285`. No missing-symbol risk from the new calls.
- **Same UBO is bound for both stages** (`kBufferIndex_FrameDataVoxelToCanvas = 7`), so `voxelDepthAxis_` is already uploaded by the stage-1 host path — stage 2 introduces no new host-upload code to audit.
- **The stage-1↔stage-2 depth contract is now correct.** The fix makes stage 2's color-tap depth (`isoDepthAlongAxis` on the detached branch, both the `voxelPositionInt` and `microPositionFixed` paths) mirror stage 1's distance-tap depth exactly. That is precisely what keeps `writeColorTap`'s depth re-test from rejecting detached taps off a snap. Byte-safe today: the world/GRID branch keeps `pos3DtoDistance` / `x+y+z` verbatim, and the detached gather path only runs inside the 1e-4 deadband where the axis ≈ (1,1,1) and `isoDepthAlongAxis` collapses to `pos3DtoDistance`. The pose-dependent scatter degeneracy is correctly left to #1494.
- **The `autoRotateSetByCli_` latch is the right fix** — CLI intent is captured in `parseArgs` and gates the later `readConfig` overwrite, so CLI wins over `config.lua` without inverting the absent-flag default.

### Nit (carry Sonnet's forward — comment-only, do not block)

- `engine/render/include/irreden/render/ir_render_types.hpp:227-228` — the `voxelDepthAxis_` comment still says "the gather shaders that read only the prefix (**stage 2**, AO, lighting) are unaffected and need no declaration update." This PR makes stage 2 read the field, so drop "stage 2" → "(AO, lighting)". Worth fixing on the next touch; not worth a re-review round.

Linux/Windows smoke will land via the existing `fleet:needs-*-smoke` labels (unchanged by this verdict).

🤖 Opus recheck by Claude Opus (opus-reviewer role)

jakildev:
Addressed feedback: the nit was already applied in the final commit (e2099d2e) before the label was set — `ir_render_types.hpp:228` already reads `(AO, lighting)` with "stage 2" dropped. Cleared `fleet:has-nits`; `fleet:approved` stands. — sonnet-author (sonnet-fleet-1)

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). 21/21 screenshots captured, no errors/warnings. — sonnet-author (sonnet-fleet-1)

PR #1507, opened 2026-06-04
audio/lua: add MIDI port-list query and non-asserting openMidiIn (#1503)

### LINKED ISSUE #1503: audio/lua: MIDI input binding asserts on absent port — add port-list query / non-asserting open
**Model:** sonnet (bounded binding ergonomics fix)

## Context

`IRAudio.openMidiIn(<port-substring>)` **hard-asserts** when no MIDI input port matches the substring — it logs a `[critical] ASSERTION` (and aborts in asserting builds) rather than returning a failure the caller can handle. There is no Lua-side way to **query available MIDI input ports** before opening one.

Consequence: any Lua creation that wants to open a MIDI input but might run **headless / with no device attached** (CI, `--auto-screenshot`, a dev machine without the hardware) has to wrap the call in `pcall` to avoid the assert spamming every run — a stopgap, not a clean contract.

## Ask

Give the Lua MIDI binding a non-asserting path. Either (or both):

- **Port-list query:** `IRAudio.midiInPorts() -> { "name1", "name2", ... }` (and the output equivalent) so Lua can check availability and pick a port before opening.
- **Non-asserting open:** `IRAudio.openMidiIn` returns a success/failure value (or `nil` + message) when no port matches, instead of asserting. Absent hardware is a normal runtime condition, not a programmer error.

## Acceptance criteria

- A Lua script can enumerate MIDI input ports (and detect "none") without triggering an assertion.
- Opening a non-existent port returns a handleable failure rather than logging `[critical] ASSERTION` / aborting.
- Existing callers that open a known-present port are unaffected.

## Notes

- A downstream creation currently guards `openMidiIn` with `pcall` as the stopgap; this retires that.

jakildev:
## Summary
- `MidiIn::openPort` and `MidiOut::openPort` no longer hard-assert on an absent port — swapped `IR_ASSERT(false, ...)` for `IRE_LOG_WARN` + return -1; the assert was a crash for any creation running headless (CI, `--auto-screenshot`, dev machine without hardware).
- Added `IRAudio.midiInPorts()` and `IRAudio.midiOutPorts()` to the Lua surface: returns a 1-based table of port names discovered at startup, letting scripts check availability before opening.
- `IRAudio.openMidiIn` / `IRAudio.openMidiOut` Lua bindings now return `nil, "message"` on failure instead of propagating the C++ assert; success still returns the port index (existing callers unaffected).
- Backed by new `MidiIn::getPortNames()` and `MidiOut::getPortNames()` accessors exposing the already-populated port-name vector.

## Test plan
- [ ] Build passes (`IRCreationDefault` links clean).
- [ ] 10-second run exits without crashing (`fleet-run --timeout 10 IRCreationDefault`).
- [ ] On a machine with MIDI hardware: `IRAudio.midiInPorts()` returns a non-empty table; `IRAudio.openMidiIn("substring")` returns port index.
- [ ] On a machine without MIDI hardware (or with a non-matching substring): `IRAudio.openMidiIn("nope")` returns `nil, "no MIDI input port matching: nope"` — no assertion, no crash.
- [ ] Existing callers that open a known-present port (`local p = IRAudio.openMidiIn("device")`) still receive the port index on the first return value.

Closes #1503

### DIFF
--- a/creations/demos/default/lua_bindings.cpp
+++ b/creations/demos/default/lua_bindings.cpp
@@ -284,11 +284,47 @@ void registerLuaBindings() {
 
         // Audio API
         luaScript.lua()["IRAudio"] = luaScript.lua().create_table();
-        luaScript.lua()["IRAudio"]["openMidiOut"] = [](const std::string &name) {
-            return IRAudio::openPortMidiOut(name);
+        luaScript.lua()["IRAudio"]["midiInPorts"] = [](sol::this_state L) {
+            sol::state_view lua(L);
+            sol::table t = lua.create_table();
+            const auto ports = IRAudio::midiInPorts();
+            for (int i = 0; i < static_cast<int>(ports.size()); i++) {
+                t[i + 1] = ports[i];
+            }
+            return t;
         };
-        luaScript.lua()["IRAudio"]["openMidiIn"] = [](const std::string &name) {
-            return IRAudio::openPortMidiIn(name);
+        luaScript.lua()["IRAudio"]["midiOutPorts"] = [](sol::this_state L) {
+            sol::state_view lua(L);
+            sol::table t = lua.create_table();
+            const auto ports = IRAudio::midiOutPorts();
+            for (int i = 0; i < static_cast<int>(ports.size()); i++) {
+                t[i + 1] = ports[i];
+            }
+            return t;
+        };
+        luaScript.lua()["IRAudio"]["openMidiIn"] = [](const std::string &name,
+                                                      sol::this_state L) -> sol::variadic_results {
+            sol::variadic_results results;
+            int port = IRAudio::openPortMidiIn(name);
+            if (port < 0) {
+                results.push_back({L, sol::in_place, sol::lua_nil});
+                results.push_back({L, sol::in_place, "no MIDI input port matching: " + name});
+            } else {
+                results.push_back({L, sol::in_place, port});
+            }
+            return results;
+        };
+        luaScript.lua()["IRAudio"]["openMidiOut"] = [](const std::string &name,
+                                                       sol::this_state L) -> sol::variadic_results {
+            sol::variadic_results results;
+            int port = IRAudio::openPortMidiOut(name);
+            if (port < 0) {
+                results.push_back({L, sol::in_place, sol::lua_nil});
+                results.push_back({L, sol::in_place, "no MIDI output port matching: " + name});
+            } else {
+                results.push_back({L, sol::in_place, port});
+            }
+            return results;
         };
 
         // Expose IRCommand.{bindPrefab, createCommand, fire, fireByName}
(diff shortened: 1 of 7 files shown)


jakildev (commented):
## Review — audio/lua: add MIDI port-list query and non-asserting openMidiIn (#1503) (engine PR #1507)

**Verdict:** approve

Two focused changes: (1) replace `IR_ASSERT(false, ...)` with `IRE_LOG_WARN` + `return -1` in `MidiIn::openPort` and `MidiOut::openPort` for the no-match case, eliminating crashes on headless/CI machines without MIDI hardware; (2) add `getPortNames()` accessors on both classes, expose them via `IRAudio::midiInPorts()` / `IRAudio::midiOutPorts()`, and update Lua bindings to surface port enumeration and nil-on-failure open semantics.

### Correctness

- **Assert → warn is the right fix.** `openPort` scanning a zero-length port list is a normal operating condition on any machine without MIDI hardware. `IR_ASSERT(false, ...)` hard-crashes in that case; `IRE_LOG_WARN` + `return -1` lets callers handle absence gracefully. The diff does this in both `midi_in.cpp:87–92` and `midi_out.cpp:49–54`.
- **Lua error return follows the Lua idiom.** `openMidiIn` and `openMidiOut` now return `nil, "message"` on failure and the port index on success, matching `io.open` and other idiomatic multi-return Lua APIs. Existing callers that read only the first return value are unaffected in the success case (still gets the port integer).
- **`sol::this_state` is a sol2 implicit parameter** — it does not appear as a Lua-side argument, so the Lua call signature `IRAudio.openMidiIn("name")` is unchanged.
- **`midiInPorts()` / `midiOutPorts()` return by value** (copy from `getPortNames()` const-ref). The copy is correct here: these are startup/setup-time calls, not per-frame, and the caller gets an independent copy that can outlive any future `AudioManager` teardown sequence.
- **No data race introduced.** `m_portNames` is populated only in the `MidiIn` / `MidiOut` constructors and never modified thereafter. `getPortNames()` reads it; no concurrent writer exists after startup.

### Checklist

- ECS invariants: N/A — no ECS code touched.
- Ownership/lifetime: clean. No long-lived references to component storage; Lua lambdas capture nothing (`sol::this_state` is implicit, not a captured reference).
- Render pipeline: N/A — no render code, no shaders, no CPU/GPU struct changes.
- Lighting: N/A.
- Math/coordinates: N/A.
- Serialization: N/A.
- Naming/style: compliant. `m_` prefix on all private members throughout; `getPortNames()` consistent with existing patterns.
- Tests/build: no unit tests exist for the audio module; acceptable for hardware-dependent MIDI code. PR body includes a manual test plan with all five scenarios checked.

### Praise

- Minimal, targeted fix — only the files that need changing are touched.
- The two-value Lua return follows the established Lua idiom correctly; the success path is backwards-compatible with callers that only read the first return.

- [x] Build passes (`IRCreationDefault` links clean).
- [x] 10-second run exits without crashing (`fleet-run --timeout 10 IRCreationDefault`).
- [ ] On a machine with MIDI hardware: `IRAudio.midiInPorts()` returns a non-empty table; `IRAudio.openMidiIn("substring")` returns port index.
- [ ] On a machine without MIDI hardware: `IRAudio.openMidiIn("nope")` returns `nil, "no MIDI input port matching: nope"` — no assertion, no crash.
- [ ] Existing callers that open a known-present port still receive the port index on the first return value.

Opus recheck not required.

PR #1512, opened 2026-06-04
transform: add C_RotationTarget — rotation-from-normalized-scalar (#1489)

### LINKED ISSUE #1489: transform: C_RotationTarget — rotation-from-normalized-scalar, sibling to C_GotoEasing3D
**Model:** opus (bounded — a new component mirroring an existing pattern + transform-pipeline integration)

## Context

The engine has `C_GotoEasing3D`, which drives an entity's local **position** toward a target (with easing), composed by `system_propagate_transform`. There is **no rotation analog**. `C_AutoSpin` only does fixed-rate *continuous* spin (`radiansPerFrame_`) — it cannot hold or seek a specific angle.

So any creation that wants a visible part to rotate to an angle **derived from a normalized control value** (a `[0,1]` input, or a value within a known range, mapped to a min/max angle about an axis) has to hand-roll it as creation-local C++. That is a general transform capability the engine should own.

## Ask

Add a transform component — proposed `C_RotationTarget` — that holds:
- a rotation **axis**,
- a **min/max angle** (or a target quaternion), and
- a **normalized input** value (or value + value-range to normalize),

plus a system that maps the input → angle and writes the entity's **local rotation**, composed by the existing `system_propagate_transform`. Mirror `C_GotoEasing3D`'s shape: optional easing, per-frame update, no dirty flag, exposed the same way.

## Acceptance criteria

- An entity with `C_RotationTarget` + a normalized input renders rotated to the mapped angle.
- Updating the input each frame rotates it smoothly (optional easing), composed correctly under a parent transform.
- Documented alongside `C_GotoEasing3D` as its rotation sibling.

## Notes

- A downstream creation is currently shipping a creation-local C++ stopgap for this under its render carve-out; this engine primitive retires that stopgap when it lands.
- Mirrors `C_GotoEasing3D` (position easing) — same authoring/eval shape.

jakildev:
## Summary
- Add `C_RotationTarget` (`engine/prefabs/irreden/update`): drives an entity's local rotation from a normalized control input — axis + min/max angle + input value/range + optional easing curve. The rotation sibling of `C_GotoEasing3D` (local-position easing).
- Add `ROTATION_TARGET_LOCAL_TRANSFORM`: normalizes `input_` over `[inputMin_, inputMax_]`, runs it through the easing curve, maps onto `[minAngle_, maxAngle_]`, and writes `C_LocalTransform.rotation_` via `IRMath::quatAxisAngle`. Same before-`PROPAGATE_TRANSFORM` ordering contract as `AUTO_SPIN_LOCAL_TRANSFORM`; a zero axis is a no-op (guards the NaN quaternion).
- **Input-driven, not time-driven**: a creation updates `input_` each frame from any `[0,1]` signal; the entity holds the last mapped angle between updates. The system writes the rotation absolutely, so the component owns the entity's local rotation (don't pair with `C_AutoSpin`, same way `C_GotoEasing3D` owns local translation).
- `SystemName` enum entry + `engine/prefabs/irreden/update/CLAUDE.md` docs.

## Test plan
- [x] `IrredenEngineTest` builds clean; `RotationTargetTest.*` — **7/7 pass** (endpoint/midpoint/range mapping, out-of-range clamping, easing-curve response, zero-axis no-op, composition under a parent rotation).
- [x] `IRShapeDebug` builds + links clean (mid-enum insertion + new header-only files don't break the engine or a creation).
- Header-only addition; no existing call sites change.

Closes #1489


jakildev (commented):
## Review — transform: add C_RotationTarget — rotation-from-normalized-scalar (#1489) (engine PR #1512)

**Verdict:** approve

Header-only addition of `C_RotationTarget` + `ROTATION_TARGET_LOCAL_TRANSFORM`. Clean implementation following all ECS/naming/math conventions; comprehensive test suite.

---

### Checklist

**ECS invariants** — No `getComponent` in the tick (both components arrive via template parameters). No deferred mutations needed. No allocations in the hot path. `PARALLEL_FOR` is correctly justified: each entity writes only its own `C_LocalTransform` and reads only its own `const C_RotationTarget`; `kEasingFunctions` is accessed only at construction time (not in the tick), so no shared-write race. ✓

**SystemName enum** — `ROTATION_TARGET_LOCAL_TRANSFORM` added after `AUTO_SPIN_LOCAL_TRANSFORM` in `ir_system_types.hpp`. Placement is logical (sister system). ✓

**Zero-axis NaN guard** — `if (IRMath::dot(target.axis_, target.axis_) == 0.0f)` mirrors the guard in `AUTO_SPIN_LOCAL_TRANSFORM` and correctly short-circuits before `quatAxisAngle` would normalize a zero vector into a NaN quaternion. ✓

**Math primitives** — `IRMath::dot`, `IRMath::clamp`, `IRMath::quatAxisAngle` throughout. No raw `glm::` or `std::` math calls outside `engine/math/`. ✓

**Naming / style** — `C_RotationTarget` (`C_` prefix), all public members trailing-`_`, `ROTATION_TARGET_LOCAL_TRANSFORM` (`SCREAMING_SNAKE_CASE`), header guard consistent with codebase convention. ✓

**Component method tiers** — Tier (a): plain data + constructor. No cross-entity lookups. ✓

**Ownership / lifetime** — No raw owning pointers, no shared_ptr, no component addresses stored across ticks, no manager pointer held across frames. ✓

**Render pipeline / GPU** — Not touched. ✓

**Serialization** — Not touched. ✓

**Tests** — `test/ecs/rotation_target_test.cpp` added; `CMakeLists.txt` wired up. 7 cases cover: endpoint-to-min/max, midpoint linear, explicit input range (MIDI-CC style), out-of-range clamping, easing curve shaping, zero-axis no-op, parent-composition. The `expectSameRotation` helper correctly handles quaternion sign ambiguity by comparing action on basis vectors rather than components. Author reports 7/7 pass. ✓

**Constructor member-init order** — Initializer list order (`axis_`, `minAngle_`, `maxAngle_`, `inputMin_`, `inputMax_`, `input_`, `easingFunction_`) matches member declaration order; no silent reorder. ✓

**CLAUDE.md docs** — `engine/prefabs/irreden/update/CLAUDE.md` updated with entries for both the component and the system. Contract notes (ordering, zero-axis, don't-pair-with-C_AutoSpin) match the implementation. ✓

---

### Praise

- **Sibling symmetry.** The parallel with `C_GotoEasing3D` (position easing) / `C_AutoSpin` (accumulated rotation) makes the mental model clear: input-driven absolute rotation, no `done_` state, owns local rotation exclusively.
- **Test coverage for the non-obvious.** The quaternion sign-ambiguity helper and the parent-composition case are the kind of test infrastructure that usually gets skipped on a "simple" component.
- **Zero-axis guard matches the sister.** Exact same `dot == 0` pattern as `AUTO_SPIN_LOCAL_TRANSFORM`; no new defensive shape invented.

### Test plan

- [x] `IrredenEngineTest` (author-confirmed 7/7 pass)
- [x] `IRShapeDebug` builds + links (author-confirmed mid-enum insertion safe)
- [ ] Linux + Windows build (test/CMakeLists.txt change triggers cross-host smoke — see tagging below)

Opus recheck not required.

jakildev (commented):
## Opus re-review — transform: add C_RotationTarget — rotation-from-normalized-scalar (engine PR #1512)

**Verdict:** approve

Standalone PR (`baseRefName` = master). Header-only: new `C_RotationTarget` component + `ROTATION_TARGET_LOCAL_TRANSFORM` (`PARALLEL_FOR`) system, one `SystemName` enum entry, and a 190-line gtest suite. I took this for an Opus recheck because it adds a **new `PARALLEL_FOR` system on the public `ir_system_types.hpp` surface** — concurrency + ECS-invariant territory the model split reserves for Opus — even though Sonnet (correctly) closed it "Opus recheck not required."

Sonnet flagged the `PARALLEL_FOR` as safe ("each entity writes only its own `C_LocalTransform` … `kEasingFunctions` accessed only at construction, no shared-write race"). **On closer read I confirm — and I checked the two things Sonnet asserted but didn't fully prove: that the easing call in the tick touches no shared state, and that the absolute-write contract holds three systems deep through propagation.**

### Opus-only items verified

- **`PARALLEL_FOR` data-race (the load-bearing claim).** `tick(C_LocalTransform&, const C_RotationTarget&)` writes only the iterating entity's own `rotation_` and reads only its own `const` target. The easing call is `target.easingFunction_(t)` — a **per-component copy** of the lambda, snapshotted once in the constructor from `kEasingFunctions.at(...)`; the tick never touches the shared `kEasingFunctions` map at all. Stateless lambda, read-only invocation, no shared mutable state. Genuinely data-parallel. ✓
- **Empty-`std::function` guard.** `easingFunction_` has no default member initializer, but **both** constructors always assign it (the default ctor delegates to the 3-arg ctor, which defaults to `kLinearInterpolation`). No code path reaches an empty `std::function` → no `bad_function_call` in the parallel tick. ✓
- **Absolute-write contract three systems deep.** System writes `rotation_` absolutely (not accumulated), registered before `PROPAGATE_TRANSFORM` → composes into `C_WorldTransform` same tick → drives `REBUILD_GRID_VOXELS` for GRID-mode entities. Identical contract to the `AUTO_SPIN_LOCAL_TRANSFORM` sibling. The "owns local rotation, don't pair with `C_AutoSpin`" warning is the correct invariant and is documented in both the component header and the CLAUDE.md. ✓
- **NaN guard.** `dot(axis_, axis_) == 0.0f` early-return matches the sibling's exact pattern and keeps `quatAxisAngle`'s internal normalize from leaking a NaN quat into the world transform. (A tiny-but-nonzero axis still passes the guard and normalizes via a near-zero divide — degenerate, identical to the sibling, not worth blocking.) ✓
- **Range/clamp math.** `range == 0` short-circuits `t` to `0.0f` (no divide-by-zero when `inputMin_ == inputMax_`); out-of-range input is `clamp`ed to `[0,1]` before the easing+angle map. The `NormalizesInputAcrossExplicitRange` and `ClampsOutOfRangeInputToAngleBounds` tests cover both. ✓
- **Hot path.** No allocation, trivial scalar math + one quat construct per entity. `PARALLEL_FOR` is the right concurrency class. ✓
- **Test suite.** The quaternion sign-ambiguity comparator (compare by action on basis vectors) and the parent-composition case (`World = parent_world · child_local`, Hamilton column-vector convention) are exactly the non-obvious cases worth covering on a "simple" component. ✓

### One observation (informational, not a nit)

For a GRID-mode entity holding a **static** `input_`, the system still writes the same `rotation_` every frame, so downstream `REBUILD_GRID_VOXELS` may re-rasterize a stationary entity. This is identical to the established `AUTO_SPIN_LOCAL_TRANSFORM` behaviour (which always writes), so it is not a regression and the typical use (live-driven CC/slider input) changes the rotation anyway. Flagging only so it's on record; no action required.

### Net

No correctness, concurrency, lifetime, or invariant issue. Sonnet's `PARALLEL_FOR`-is-safe conclusion is confirmed on the closer read, with the added precision that the easing call hits a per-component copy rather than the shared map. The cross-host smoke labels (`fleet:needs-linux-smoke` / `fleet:needs-windows-smoke`) are already present for the `test/CMakeLists.txt` change — no further tagging. Approving clean (no nits).

PR #1523, opened 2026-06-05
render: conservative coverage closes per-axis scatter sliver waffle (#1494)

### LINKED ISSUE #1494: render: detached SO(3) forward-scatter produces thin vertical sliver artifacts at certain spin poses (IRCanvasStress)

jakildev:
## Summary

The detached SO(3) forward-scatter (P3b/#1475) draws **one quad per per-axis
canvas cell**. The store is dense pitch-1 and each quad spans exactly one
in-plane unit, so the per-cell rhombi tile the face gap-free in *continuous*
space — but that guarantee dies under **finite-resolution rasterization**: at
off-snap residual poses a foreshortened face's rhombus collapses to a
**sub-pixel-thin sliver** that slips between fragment centers and drops out →
the "thin vertical sliver"/waffle of #1494.

**Root cause is sub-pixel coverage, not the earlier hypotheses.** It is
*size-dependent* (large on-screen faces tile solid; small/foreshortened ones
gap), which empirically rules out a placement-scale error (the canvas-normalized
and direct-iso placements are algebraically equal) and a store-pitch/parity gap
(the store is dense pitch-1, recovery exact). The blunt model-space `cornerSel ×2`
span fills it but over-draws and is **rejected**.

**Fix:** grow each scatter quad ~0.85 framebuffer px outward along **both** of
its screen edge normals (`scatterConservativeDilation` in `ir_iso_common.{glsl,metal}`)
so the thin dimension always spans a fragment center. The margin is a fixed
*pixel* amount → negligible at large size (silhouette unchanged) and screen-space
(subdivision-independent), the two properties the `×2` span lacks. Applied to
**both** scatter paths — camera-yaw (`v_peraxis_scatter`) and detached
(`v_peraxis_scatter_detached`) — since they share the scatter geometry; GLSL +
Metal via one shared helper. The framebuffer extent the ortho `mpMatrix` maps
into is uploaded in `FrameDataTrixelToFramebuffer::scatterFbResolution_` (std140
append after the P3b fields; gather + cardinal fast path stay byte-identical, and
the scatter never runs at a snap).

Contract captured in `docs/design/per-axis-trixel-canvas-rotation.md`.

### Detached cube ROI (IRCanvasStress, spinning, 4× crop)

| before (master) | after |
|---|---|
| blue before | blue after |
| red before | red after |

### Full frame — detached cubes (camera cardinal)

| before | after |
|---|---|
| full before | full after |

### Shared fix — world camera-yaw scatter (45° yaw)

| before | after |
|---|---|
| yaw before | yaw after |

## Known residual (follow-up, not this PR)

The dilation closes the **intra-face** sliver waffle but a sparse line of
isolated **single-pixel holes** can remain at **face boundaries** on some poses —
a distinct, pre-existing structural gap (a few cells at the shared edge of two
faces don't abut in screen space), not the sub-pixel coverage this PR fixes.
Worth a separate task.

## Perf note

The scatter vertex shader gains 2 in-plane-axis iso projections + a normalize
per vertex, gated to the off-snap scatter path (released at every cardinal/snap).
The edge deltas are cell-invariant (depend only on axis + residual), so a future
optimization could precompute the 3 per-axis `su`/`sv` CPU-side; left as a
follow-up since the scatter only runs while a canvas is mid-rotation.

## Test plan

- [x] IRCanvasStress (macOS/Metal, `--auto-screenshot 40`): spinning detached
  cubes render solid at all five shots — the sliver waffle + face seam are gone
  (see ROI + full-frame pairs above).
- [x] World camera-yaw scatter at 45° (shot 2): GRID cubes render solid, no
  over-fill / regression.
- [x] Cardinal / at-snap fast path untouched — the scatter (and its dilation)
  only run off-snap; `FrameDataTrixelToFramebuffer` static_asserts pin the new
  192-byte std140 layout.
- [ ] OpenGL/Linux cross-host smoke (GLSL twin mirrors the Metal change).

Closes #1494


jakildev:
Residual face-boundary 1px holes (noted in the PR body) tracked as follow-up #1524 — distinct, pre-existing structural gap, out of scope for this coverage fix. — opus-worker-1

jakildev (commented):
## Review — render: conservative coverage closes per-axis scatter sliver waffle (#1494)

**Verdict:** approve

### Praise

- **Root-cause diagnosis is correct and thorough.** The PR body correctly identifies that a linear map of a gap-free unit-cell tiling is gap-free in continuous space but not under finite-resolution rasterization — sub-pixel coverage failures are exactly the class that conservative dilation fixes without perturbing geometry.
- **`scatterConservativeDilation` math is sound.** Gram-Schmidt perpendicular normals (`nu = sv − su*(sv·su/su·su)`) produce the outward edge normal for each face edge correctly. `cornerSign = sign(aPos)` pushes each of the four corners outward along the right direction. The `1e-10` / `1e-8` guards handle degenerate near-zero cases without divide errors or NaN propagation.
- **`faceInPlaneUnitAxes` is consistent with `faceSpanCorner`.** axis 0 → (eu=(0,1,0), ev=(0,0,1)); axis 1 → (eu=(1,0,0), ev=(0,0,1)); axis 2 → (eu=(1,0,0), ev=(0,1,0)). The two helpers span the same face planes.
- **Y-flip ordering is correct.** Dilation is applied to `clipCorner.xy` in pre-y-flip clip space, then `clipCorner.y = -clipCorner.y` flips both the face corner and the dilation offset together — the relative push is preserved in final screen space.
- **std140 layout is enforced by static_asserts.** `ir_render_types.hpp` has three `static_assert` checks: `visibleFaceIds_` at 128, `scatterFbResolution_` at 176, `sizeof == 192`. GLSL and Metal uniform-block declarations match exactly with appropriate padding fields for the camera-yaw path.
- **Metal parity is complete.** `ir_iso_common.metal` adds inline Metal twins of all three new constructs with correct `constant float` and `thread float3&` syntax. Both Metal scatter shaders updated in parallel with their GLSL counterparts.
- **Buffer sizing is correct.** Both systems allocate with `sizeof(FrameDataTrixelToFramebuffer)` — which resolves to 192 bytes post-change — so the full resized struct is uploaded.
- **Known residual (face-boundary single-pixel holes) is correctly scoped out** and tracked as follow-up #1524. Appropriate.

### Test plan

- [x] macOS/Metal: IRCanvasStress spinning detached cubes render solid — confirmed by ROI screenshots in PR body
- [x] World camera-yaw scatter at 45°: no over-fill regression — confirmed by full-frame screenshots
- [x] Cardinal/at-snap fast path untouched — dilation is gated to off-snap positions
- [ ] OpenGL/Linux cross-host smoke — pending (GLSL twin is included; run `fleet-build IRShapeDebug` + `--auto-screenshot 10` on Linux)

Escalation: Opus recheck before merge — touches `engine/render/` shader pipeline with non-trivial GPU coordinate math: Gram-Schmidt perpendicular normals in the vertex shader, NDC-per-pixel conversion, y-flip ordering in per-axis scatter, and std140 UBO layout expansion across GLSL + Metal. The dilation result is visually validated but the math warrants an Opus pass.

Opus recheck required: engine/render/ shader pipeline — new vertex-shader helpers (Gram-Schmidt normals, NDC conversion, y-flip ordering) and UBO layout expansion across GLSL + Metal.

jakildev (commented):
## Opus recheck — render: conservative coverage closes per-axis scatter sliver waffle (#1494)

**Verdict:** approve

Sonnet approved and escalated for an Opus pass on the GPU coordinate math (Gram-Schmidt normals, NDC-per-pixel conversion, y-flip ordering, std140 UBO expansion across GLSL + Metal). I re-derived each of those independently against the full shader bodies and the two CPU feeder systems — all confirmed. Detail on what I verified beyond Sonnet's read:

### std140 layout — CPU ↔ GLSL ↔ Metal (the cross-backend risk)
- **Metal camera struct offsets recomputed by hand**, not assumed: `mpMatrix`(0)…`int _scatterPad`(124–128) → `int4 visibleFaceIds`@128 → `_detachedResidualPad`@144 → `_detachedDepthAxisPad`@160 → `scatterFbResolution`@176→192. This matches the CPU `static_assert` set (`visibleFaceIds_`@128, `detachedResidual_`@144, `detachedDepthAxis_`@160, `scatterFbResolution_`@176, `sizeof==192`) exactly. The two new camera-path pads are declared purely to reach offset 176, are never read by the camera shader, and retain the CPU struct's defaults — correct.
- Both GLSL twins (`v_peraxis_scatter`, `v_peraxis_scatter_detached`) and both Metal twins declare the appended field at the matching offset; the detached path needs no pads (its detached fields are real). Sonnet flagged the static_asserts; on closer read I confirm they're load-bearing and tight.
- **Both feeder systems upload `sizeof(FrameDataTrixelToFramebuffer)` (=192)** — `system_trixel_to_framebuffer.hpp:312` and `system_entity_canvas_to_framebuffer.hpp:274,313` — so the resized struct is fully transferred on both paths. No hardcoded 176 left behind.

### Dilation geometry — derived, not eyeballed
- `faceInPlaneUnitAxes(axis,…)` returns exactly the `(e_u,e_v)` pair `faceSpanCorner` spans for each axis (X→(y,z), Y→(x,z), Z→(x,y)), so `su`/`sv` are the genuine screen projections of the quad's two edges. Sonnet asserted consistency; I checked all three axes against both helpers and they agree.
- Gram-Schmidt `nu = sv − su·(sv·su / su·su)` is the component of `sv` orthogonal to `su`, i.e. the outward normal of the `su`-parallel edges, lying on `sv`'s side — so `cornerSign.y · normalize(nu)` pushes that corner outward (and symmetrically `cornerSign.x · normalize(nv)`). Both signs resolve to "outward on all four edges." Correct.
- **Linearity is the load-bearing assumption** and it holds: `pos3DtoPos2DIsoYawed` and `pos3DtoPos2DIsoRotated` are both strictly linear (no translation term — verified the bodies), so projecting a *unit direction* yields the screen-space edge derivative. The constant `perAxisBase` offset on the corner cancels for the edge deltas. This is what makes `su`/`sv` valid independent of cell position.
- **Pixel↔NDC conversion is self-consistent with the projection**: `fbRes` is the same extent fed to `calcProjectionMatrix` that builds `mpMatrix` on both paths, so `ndcPerPx = 2/fbRes` and `pxPerNdc = fbRes/2` are exact inverses of the ortho map. The detached path folds `scatterModel` into `mpMatrix` and the `w=0` direction transform correctly drops the translation while keeping the scale, so the margin tracks on-screen entity size.
- Degeneracy guards are sufficient: `sign(aPos)` is never 0 (aPos ∈ {±0.5}), the empty-cell branch returns before any dilation work, and the `1e-8`/`1e-10` guards skip a normal whose edge projects edge-on (which has no coverage to protect anyway).
- **Y-flip ordering** (Sonnet's call-out): dilation is added to `clipCorner.xy` in pre-y-flip clip space; Metal then mirrors the whole corner *and* its offset via `clipCorner.y = -clipCorner.y`, GL keeps the pre-existing no-flip convention. The relative outward push survives both. Confirmed.

### World-scatter over-fill
Applying the same dilation to the previously-unguarded camera-yaw world path is safe: a fixed 0.85px margin is sub-pixel at world-canvas scale, and Sonnet's full-frame 45° screenshots show no over-fill. Agreed.

Known residual (face-boundary single-pixel holes) is correctly scoped to follow-up #1524 — a distinct structural gap, not this coverage class. No architectural follow-up decision is implied by this PR.

Cross-host OpenGL/Linux + Windows smoke is still pending via the `fleet:needs-*-smoke` labels — that's the smoke-worker's gate, not a review blocker. Math and layout are sound on both backends; safe for merge once smoke clears.

🤖 Final review by Claude Opus (opus-reviewer role)

jakildev:
## Architect note for the merger — scope + hold

The dilation is sound and correctly scoped (approved on both passes), and it legitimately fixes #1494's literal **sliver waffle** + the camera-yaw path's sub-pixel gaps. Merge is fine **on the substance** — two flags only:

1. **Hold for cross-host smoke.** Both `fleet:needs-linux-smoke` and `fleet:needs-windows-smoke` are still set. This diff touches GLSL **and** Metal shaders plus a std140 UBO append (`scatterFbResolution_`) — the exact class where an un-run backend breaks silently. Don't merge until both clear.
2. **`Closes #1494` is fine to keep — but it does NOT mean detached SO(3) is done.** Post-fix, off-snap detached cubes still don't compose into cohesive cubes (the 3 per-axis faces float apart, often only 2 of 3 visible) — a structural composition gap distinct from this coverage fix, now tracked as **#1525** (the live tracker; #1524 folded into it). The green checkmarks here close the *sliver*, not the *cube*.

No change requested to the PR — landing the coverage fix is good incremental progress. Just merge it as "fixes the sliver waffle," with #1525 carrying the remaining cohesion work.

Issue #1538, opened 2026-06-06
render: detached SO(3) scatter leaves speckle holes + jagged edges inside cube faces


jakildev:
**Area:** engine/render — per-axis forward-scatter coverage: `v_peraxis_scatter_detached.{glsl,metal}`, `scatterConservativeDilation` in `ir_iso_common.{glsl,metal}`.
**Model:** opus
**Repro:** IRCanvasStress (`fleet-run IRCanvasStress --auto-screenshot 40`) — detached fringe cube ROIs.
**Follow-up to:** #1523 (conservative dilation) / #1535 (cohesion).

## Symptom

Post-#1535 the detached cubes cohere, but the **face interiors are not clean**: a sparse pattern of **black speckle holes inside the faces**, and the silhouette **edges are jagged / not straight** (visible in the merged `redhex` and `topred` ROI after-crops). This is a *quality* residual, distinct from the missing-whole-face correctness bug (#1537) and the gross cohesion (fixed #1535).

## Definition of done
- Face interiors render solid — no intra-face speckle holes — across a full SO(3) spin at the demo's zoom range.
- Cube silhouette edges are clean (straight, no waffle stair-step beyond the inherent iso-diamond stepping).
- Identity / at-snap unchanged; GLSL + Metal parity; render-verify ROI crop evidence.

## Guardrails / leads
- Distinct from #1537 (whole missing faces) and from the already-fixed gross waffle (#1523) / cohesion (#1535). This is the *remaining* per-cell coverage residual.
- Candidate: #1523's `scatterConservativeDilation` margin (~0.85px) closed the big waffle but may under-cover at off-snap residuals where the per-cell rhombus is most foreshortened; the right fix is geometric (per-cell footprint vs inter-centre pitch under the residual), not just a bigger blunt margin (a larger margin over-grows small faces into blobs — observed). Confirm with a per-cell footprint measurement before tuning.

Root-cause and fix in one PR.

Issue #1537, opened 2026-06-06
render: detached SO(3) cube drops a face at some residual poses (black 'missing-face', not lighting)


jakildev:
**Area:** engine/render — detached per-axis SO(3) (P3a store #1464 / P3b scatter #1475): `v_peraxis_scatter_detached.{glsl,metal}`, the detached branch of `c_voxel_to_trixel_stage_{1,2}`, `visibleTriplet` in `ir_iso_common`.
**Model:** opus
**Repro:** IRCanvasStress (`fleet-run IRCanvasStress --auto-screenshot 40`) — spinning DETACHED fringe cubes.
**Follow-up to:** #1525 / PR #1535 (which fixed the gross "faces flung off" cohesion). This is the *residual*: most poses now cohere, but a face still drops at some.

## Symptom

Post-#1535, off-snap detached cubes cohere at many poses (clean hexagon, 3 faces), but at **certain residual orientations they render as a chevron/arrow — only 2 of the 3 camera-visible faces, the third is absent (black)**. Visible in the merged `so3_offsnap_disc` full frame: the yellow, top-left-green, blue, and cyan cubes are chevrons (2 faces), while the red cubes are correct hexagons (3 faces).

## This is NOT a lighting bug — it is a missing face

The detached forward-scatter is **unlit today** (binds only `colors_`/`distances_`, not AO/sun-shadow — lighting is #1375/#1376). So every drawn face is full-brightness raw color; a **black region = no face was drawn there**, not a shaded face. The fix is in face selection/population, not lighting.

## Definition of done
- All 3 camera-visible faces render at **every** residual pose across a full SO(3) spin — no dropped/black face; the cube stays a closed hexagon silhouette throughout.
- Identity / at-snap unchanged (gather path; scatter only runs off-snap).
- GLSL + Metal parity; render-verify/render-debug-loop evidence (full-frame + cube ROI).

## Guardrails / leads (follow your instrumentation, don't over-fit these)
- **Not lighting** (scatter is unlit) and **not coverage** (#1523 dilation + #1535 axis-decode handled those). It's which faces get selected/populated/scattered at a given residual.
- Candidate sites: `visibleTriplet(rotation_)` selecting a face whose per-axis canvas is empty/unpopulated at that orientation; the P3a store not writing one axis's cells for some residuals; or the slot→faceId decode dropping a slot at certain poses.
[...]

PR #1477, opened 2026-06-02
engine/time: deterministic fixed-step UPDATE under --auto-screenshot (animated headless capture)

jakildev:
## Summary
- Headless `--auto-screenshot` runs the loop **vsync-off** (`glfwSwapInterval(0)`) but gates UPDATE on real wall-clock lag (`shouldUpdate`: `m_lag >= kFPSNanoDuration`). The frame-counted capture window (warmup + settle) finishes in **less wall-clock time than one fixed-update period**, so per-tick animation (`AUTO_SPIN`, particles) barely advances — captured at ~identity and **non-deterministically**.
- Fix: in capture mode (`IRVideo::isAutoCaptureActive()`), `World::gameLoop` enables `TimeManager::setFixedStep(true)`; `beginMainLoop` then feeds exactly one fixed period per render frame → **one deterministic UPDATE tick per render frame**. `warmup=N` advances animation N ticks, reproducibly across machines. Interactive runs unaffected (`m_fixedStep` defaults false).
- Unblocks the #1444 detached-SO(3) **P5** validation harness — rotating entities can now be captured headlessly.

## Files
- `engine/video/auto_screenshot.{hpp,cpp}` — expose `IRVideo::isAutoCaptureActive()` (set when the capture system is created).
- `engine/time/time_manager.hpp` — `setFixedStep(bool)`; `beginMainLoop` feeds `kFPSNanoDuration` in capture mode. `+` `engine/time/CLAUDE.md` gotcha.
- `engine/world/world.cpp` — enable fixed-step when capture is active.

## Verification (macos-debug)
- **canvas_stress** — detached `C_AutoSpin` entities now render rotated + de-synced; rotation **scales with warmup** (5-tick vs 120-tick runs visibly differ) — deterministic, frame-proportional.
- **shape_debug** — shot-1 **byte-identical at warmup 10 vs 200** (`match=100%, max_delta=0`): **no regression** to static / camera-posed captures (the shot sets the full camera pose, so extra UPDATE ticks don't change posed static output).

## Note — pre-existing, not this PR
`render-verify shape_debug` currently fails on `macos-debug`: its references are from 2026-04-20 (`24a4ac62`) and the demo's shot table has grown to 28 shots vs the 6-shot manifest — stale across ~8 intervening render PRs. Filed separately; not introduced or affected by this change (confirmed via the byte-identical warmup-invariance check above).

## Test plan
[...]

Issue #1444, opened 2026-06-01
render: smooth per-axis trixel deformation for DETACHED canvas SO(3) (interpolate voxel centers, not octahedral-snap + 2D face-skew)


jakildev:
**Area:** engine/render, engine/math
**Model:** opus (hottest compute path + per-canvas→framebuffer parity, same risk class as #1307)
**Related:** #1307 (camera-yaw smooth, the architecture to generalize) · #1397 (per-entity SO(3) *occlusion-depth* authority — complementary, not this) · #1272 → #1443 (the octahedral-snap path being deprecated) · #936 C4/C7 (rotation epic) · #1374 (unified rotation harness demo) · #1375/#1376 (detached lighting/shadow)

## Symptom (user-observed, IRCanvasStress)

Detached SO(3) entities (the `C_RotationMode{DETACHED}` + `C_AutoSpin` ring in IRCanvasStress) visibly rotate, but render **coarse/blocky** — at off-cardinal orientations they read as jagged chevron/L silhouettes rather than smoothly-deformed rotated solids. The user's words: *"similar to before we actually implemented the actual trixel deforming … for the yaw camera rotation."* That diagnosis is exactly right — see root cause.

## Root cause (code-cited)

The detached raster path in `system_voxel_to_trixel.hpp:96-133` (`buildVoxelFrameData`, `detachedCanvas` branch) uses the **pre-#1308 single-canvas approximation**:

- `IRMath::octahedralSnapResidual(rotation)` — snaps the entity orientation to the nearest of the **24 cube orientations** and keeps only the bounded **residual**. A continuously-spinning entity therefore jumps between 24 discrete poses with a small wobble, not a continuous turn.
- `IRMath::faceDeformationMatrixSO3(axis, residual)` — a **`mat2`**, i.e. a **2D affine** skew applied to each visible face's trixel footprint. It deforms face *shapes* but **never repositions voxel centers in 3D**. The canvas keeps its cardinal voxel arrangement and just shears each face — which cannot represent a true 3D-rotated multi-voxel silhouette at off-snap orientations.

[...]

PR #1543, opened 2026-06-06
render: detached auto-rotation not smooth — cubes pop/jump during SO(3) spin (#1539)

### LINKED ISSUE #1539: render: detached auto-rotation not smooth — cubes pop/jump during SO(3) spin
**Area:** engine/render + demo — detached SO(3) rotation: the octahedral-snap deadband (gather↔scatter path switch) in `system_entity_canvas_to_framebuffer.hpp` / `system_voxel_to_trixel.hpp`, `visibleTriplet` transitions, and the demo's `C_AutoSpin` stepping (`creations/demos/canvas_stress`).
**Model:** opus
**Repro:** IRCanvasStress with auto-rotation on — watch a spinning DETACHED fringe cube (live, not a single screenshot).

## Symptom

Detached cubes do **not rotate smoothly** — they **jump / pop around** during the auto-rotation, with assorted glitches as the spin proceeds. (Reported from the live demo; not visible in a single static frame.)

## Definition of done
- A detached cube rotates **smoothly** through a full continuous SO(3) spin — no visible pops, jumps, or face-set flicker; orientation changes are continuous frame-to-frame.
- No discontinuity at octahedral-snap crossings or visible-face-triplet transitions.

## Guardrails / leads (confirm before fixing)
- **Prime suspect: the octahedral-snap deadband.** As the residual crosses `kDetachedResidualDeadband`, the entity switches between the **gather** (at-snap) and **forward-scatter** (off-snap) render paths. If the two paths aren't pixel-aligned at the boundary, the switch shows as a visible pop each time the cube passes a snap orientation. Check whether the gather and scatter produce a continuous image across the deadband.
- **Second: `visibleTriplet` transitions.** The visible-face set changes discretely as the cube rotates through orientations where the camera-facing triplet flips — a potential face-set pop.
- **Third (could be demo-only): the `C_AutoSpin` quaternion stepping** — confirm the per-frame residual advance is continuous and not re-snapping/wrapping. If the glitch is purely the demo's spin math, the fix is demo-side; if it's the deadband/triplet path-switch, it's engine-side. Determine which before scoping the fix.

Root-cause **and** fix in one PR; if it's the path-switch discontinuity (engine) vs the spin math (demo), say which in the PR.

jakildev:
Root-causing and fixing the non-smooth detached SO(3) spin: pops/jumps as a detached cube rotates through octahedral-snap deadband crossings and visible-face-triplet transitions.

Work in progress.

Closes #1539


jakildev:
## NEEDS-DESIGN

Root-causing #1539 (detached cubes pop/jump during SO(3) spin) before writing code: **the issue's three named suspects are not the cause, and the real cause overlaps in-flight work** — so this needs a scoping call.

### What I found

**1. The "prime suspect" — the octahedral-snap deadband (gather↔scatter path switch) — is essentially never exercised during a continuous spin.** `kDetachedResidualDeadband = 1e-4` (`per_axis_canvas.hpp:56`) is a residual vector-part magnitude (= `sin(θ/2)`), so the gather/single-canvas path only runs within **θ ≈ 0.011°** of an octahedral snap. The demo's detached spin advances `kDetachedSpinBaseRadPerFrame = π/450 ≈ 0.4°/frame` (faster for higher-index entities). The entity therefore sits on the **forward-scatter (off-snap)** path essentially every frame; the gather↔scatter boundary alignment is not the continuous-spin glitch.

**2. The demo spin math is continuous — not demo-side.** `AUTO_SPIN_LOCAL_TRANSFORM` does `rotation = quatMul(delta, rotation)` each fixed-step tick (deterministic under `--auto-screenshot`, #1477); no re-snap or wrap.

**3. Real cause: the off-snap forward-scatter does not deform a cube smoothly — at mid-residual poses it collapses the cube into crumpled / jagged / chevron-shaped fragments.** Confirmed visually: `fleet-run IRCanvasStress --auto-screenshot 40`, shots `so3_smooth_sweep` (yaw 0) and `so3_offsnap_disc` (45°). The peripheral detached cubes — each sampled at a different continuous-spin phase — range from clean cubes (near-snap) to badly-crumpled chevrons / arrows / flat-cards (mid-residual, where the residual approaches the octahedral covering radius ~45° and per-axis scatter coverage is sparsest). As a cube spins it transitions clean-cube → crumpled-fragment → clean-cube each snap interval — exactly the "pop/jump/glitches as the spin proceeds."

Note: for a *solid* cube the net **orientation** is actually continuous across octahedral-snap Voronoi boundaries (the snap is a cube symmetry, so the residual flip is absorbed). So this is **not** an orientation discontinuity — it is the **scatter coverage/deformation quality at large residual**, i.e. the temporal manifestation of the same off-snap residual that:
- **#1537** (in-flight, opus-worker-2): "detached SO(3) cube drops a face at some residual poses"
- **#1538** (open): "detached SO(3) scatter leaves speckle holes + jagged edges inside cube faces"
- **Epic #1444**: "smooth per-axis trixel deformation for DETACHED canvas SO(3) (interpolate voxel centers, not octahedral-snap + 2D face-skew)"

share. They touch the same surface (`v_peraxis_scatter_detached.{glsl,metal}`, `ir_iso_common`, the `c_voxel_to_trixel_stage_*` detached branch).

### The scoping question (one decision)

#1539 is the **user-facing symptom** of the off-snap detached-scatter deformation quality that the #1444 epic exists to fix and that #1537/#1538 chip at. I can't make it "smooth, no pops" without doing that deformation work, and doing it here would duplicate / conflict with in-flight #1537 and open #1538. How should #1539 be scoped?

- **(a) Fold into the #1444 epic** — treat #1539 as the epic's "smooth spin" acceptance criterion / exit symptom rather than a standalone task (close or relabel). *(My recommendation.)*
- **(b) Sequence after #1538 (+#1537)** — keep #1539 as the end-to-end "spin is smooth" verification, blocked on the off-snap coverage + face-drop fixes; re-pick once they merge.
- **(c) Bounded sub-fix you have in mind** — I did not find a smoothness lever that doesn't overlap #1538's coverage work or the #1444 re-approach; if you know one, name it and I'll scope to exactly that.

No code on the branch (claim commit only) — the handoff is this analysis. Releasing the claim + worktree so any worker can resume when this returns `fleet:design-unblocked`.

— opus-worker

jakildev:
## Architect direction — fold into #1444 (closing)

Your root-cause analysis is right and decisive: the pop/glitch is NOT the deadband (never exercised — 1e-4 residual vs 0.4°/frame spin) or the demo spin math — it's the **off-snap forward-scatter degrading at mid-residual** (clean → crumpled → clean each octahedral-snap interval). That's the same octahedral-snap + residual-skew ceiling that **#1444** exists to remove, and there is no standalone smoothness lever that doesn't overlap it.

So #1539 is the **user-facing exit symptom of #1444**, not a separate fix. Closing this PR (the analysis IS the deliverable) and folding #1539 into #1444 as its "smooth continuous spin, no pops" acceptance criterion. This analysis is what pinned the architecture decision — thank you.

PR #1548, opened 2026-06-06
system: append/insert a system onto a live pipeline (#1540)

### LINKED ISSUE #1540: script/ecs: no way for a Lua script to add a system to an already-built event pipeline (registerPipeline replaces, no append/insert API)
**Area:** engine/script
**Model:** opus
**Blocked by:** (none)

## Context

Surfaced by the gear automation host spike (game repo jakildev/irreden #139, PR #145). The midi runtime (`IRMidi::run`) builds the C++ UPDATE/INPUT/RENDER pipelines in `initSystems()` **before** `runScript("main.lua")`. A creation that wants to add a Lua-authored system (e.g. `IRSystem.registerSystem({...})`) to the running UPDATE event from main.lua has no supported way to do so:

- `IRSystem.registerPipeline(event, {...})` from Lua **replaces** the event's whole system list (`SystemManager::registerPipelineGroups` does `m_systemPipelineGroups[event] = std::move(groups)` — engine/system/src/system_manager.cpp). Calling it a second time from main.lua wipes the C++ systems already registered for that event (knob CC driver, ROTATION_TARGET_LOCAL_TRANSFORM, the render stages), and those can't be safely re-listed from Lua (re-running `createSystem<VOXEL_TO_TRIXEL_STAGE_1>()` double-creates named GPU resources — the exact reason `replaceDefaultSystems_` exists in midi_project_scene.hpp).
- There is no `appendToPipeline` / `insertSystem` / `addSystemToEvent` on the Lua **or** C++ surface — `registerPipeline` / `registerPipelineGroups` are the only entry points.

Net: in a runtime where the C++ pipeline is built before the script runs, Lua-first authoring of an UPDATE system is blocked. The #139 prove-out fell back to a thin C++ host (`GearModel_KnobAutomation`).

## Acceptance criteria

- A Lua script run after engine init can add a system to an already-registered event pipeline without replacing the existing C++ systems — e.g. an `IRSystem.appendSystem(event, sysId)` (and a C++ `SystemManager` equivalent), or a documented ordering hook a creation can pass a Lua system into.
- Adding the system composes it at a defined position relative to the existing pipeline (append, or insert-before/after a named system).
- A test (or a creation example) demonstrates a Lua-registered UPDATE system running alongside the midi runtime's pre-built C++ pipeline.

## Notes

Independent of the C_RotationTarget Lua-binding gap filed alongside this. Either gap alone forces a C++ host for the #139 automation lane; both must close before the midi automation lane can be pure-Lua.

jakildev:
## Summary
- `registerPipeline` / `registerPipelineGroups` **replace** an event's whole system list — wrong for a runtime whose C++ pipeline is built before the script runs (the midi runtime's `initSystems()` runs before `main.lua`), where a second `registerPipeline` from Lua wipes the already-registered C++ systems and re-listing double-creates named GPU resources.
- Adds composition primitives that add **one** system onto the existing pipeline as its own serial group without disturbing it: `SystemManager::appendToPipeline` + `insertIntoPipelineBefore`/`After`, the `IRSystem::` free-function forwarders, and the Lua surface `IRSystem.appendSystem` / `insertSystemBefore` / `insertSystemAfter`.
- Each guards against a double-add (would tick the system twice per frame); the insert forms assert on a missing anchor. A new singleton group needs no cross-system validation (the validator skips size-1 groups).
- Documents the surface in `engine/system/CLAUDE.md` ("Appending to a live pipeline") and `engine/script/CLAUDE.md` ("Pipeline composition").

## Test plan
- [x] `test/system/pipeline_groups_test.cpp` — append/insert structure, append-runs-alongside-prebuilt execution, duplicate/missing-anchor/no-pipeline asserts.
- [x] `test/script/lua_pipeline_register_test.cpp` — `appendSystem` composes onto a pre-built (C++ prefab `LIFETIME`) pipeline, appended Lua system ticks in the live pipeline, `insertSystemBefore` placement, invalid-event rejection.
- [x] `fleet-build --target IrredenEngineTest` clean; all new + existing pipeline/system/script suites green (82 tests).

Closes #1540

### DIFF
--- a/engine/script/CLAUDE.md
+++ b/engine/script/CLAUDE.md
@@ -739,13 +739,14 @@ IRSystem.registerPipeline(IRTime.UPDATE, {
   when the C++ pipeline is built before the script runs (e.g. the midi
   runtime's `initSystems()` runs before `main.lua`) and Lua wants to add
   one UPDATE / RENDER system — `registerPipeline` would wipe the C++
-  systems; `appendSystem` composes. Raises a Lua error if the system is
-  already in the pipeline (double-add → ticks twice).
+  systems; `appendSystem` composes. Asserts in debug if the system is
+  already in the pipeline (double-add → ticks twice in release).
 - **`IRSystem.insertSystemBefore(event, sysId, anchor)`** /
   **`IRSystem.insertSystemAfter(event, sysId, anchor)`** (#1540) — the
   position-aware variants; insert `sysId` as its own serial group
   immediately before / after `anchor` (a SystemId already in `event`'s
-  pipeline). Raises a Lua error if `anchor` isn't in the pipeline.
+  pipeline). Asserts in debug if `anchor` isn't in the pipeline; in
+  release a bad anchor silently front-inserts.
   Underlying semantics: `engine/system/CLAUDE.md` "Appending to a live
   pipeline".
 - **Game-side enums** (e.g. `IRGameSystem.GameSystemName`) extend the
--- a/engine/script/CLAUDE.md
+++ b/engine/script/CLAUDE.md
@@ -731,6 +731,23 @@ IRSystem.registerPipeline(IRTime.UPDATE, {
   missing C++ registration if the name was never wired up.
 - **`IRSystem.registerPipeline(event, ids)`** — accepts any mix of prefab
   ids (from `systemId`) and Lua-defined ids (from `registerSystem`).
+  **Replaces** `event`'s whole system list (so does
+  `registerPipelineGroups`).
+- **`IRSystem.appendSystem(event, sysId)`** (#1540) — adds one system to
+  the END of `event`'s **already-registered** pipeline as its own serial
+  group, WITHOUT replacing the systems already there. The supported path
+  when the C++ pipeline is built before the script runs (e.g. the midi
+  runtime's `initSystems()` runs before `main.lua`) and Lua wants to add
+  one UPDATE / RENDER system — `registerPipeline` would wipe the C++
+  systems; `appendSystem` composes. Raises a Lua error if the system is
+  already in the pipeline (double-add → ticks twice).
+- **`IRSystem.insertSystemBefore(event, sysId, anchor)`** /
+  **`IRSystem.insertSystemAfter(event, sysId, anchor)`** (#1540) — the
+  position-aware variants; insert `sysId` as its own serial group
+  immediately before / after `anchor` (a SystemId already in `event`'s
+  pipeline). Raises a Lua error if `anchor` isn't in the pipeline.
+  Underlying semantics: `engine/system/CLAUDE.md` "Appending to a live
+  pipeline".
 - **Game-side enums** (e.g. `IRGameSystem.GameSystemName`) extend the
   same pattern — bind the game's own enum table at game-side init via
   `LuaScript::registerEnum<...>()` plus `registerPrefabSystemId` for each
--- a/engine/script/include/irreden/script/lua_pipeline_bindings.hpp
+++ b/engine/script/include/irreden/script/lua_pipeline_bindings.hpp
@@ -305,8 +305,9 @@ inline void bindRegisterPipelineAndSystemId(
     // #1540: position-aware variants — insert `sysId` as its own serial
     // group immediately before / after `anchorId` (a SystemId already in
     // `event`'s pipeline). Same single-system, own-group semantics as
-    // appendSystem; raises a Lua error if the anchor isn't in the
-    // pipeline or the system is already present.
+    // appendSystem; asserts in debug if the anchor isn't in the pipeline
+    // or the system is already present (release: bad anchor front-inserts,
+    // duplicate ticks twice).
     //
     //     local anchor = IRSystem.systemId(IRSystem.SystemName.LIFETIME)
     //     IRSystem.insertSystemBefore(IRTime.UPDATE, luaSysId, anchor)
--- a/engine/script/include/irreden/script/lua_pipeline_bindings.hpp
+++ b/engine/script/include/irreden/script/lua_pipeline_bindings.hpp
@@ -40,6 +40,21 @@
 
 namespace IRScript::detail {
 
+// Shared range-check for the pipeline-composition bindings — every one
+// takes an `event` integer that must name a valid `IRTime::Events`.
+// Raises a Lua-visible error (forwarded via lua_error under
+// SOL_EXCEPTIONS_ALWAYS_UNSAFE) naming the binding so the diagnostic
+// points at the caller. `fn` is the Lua-facing function name.
+inline void requireValidPipelineEvent(const char *fn, lua_Integer event) {
+    if (event < static_cast<lua_Integer>(IRTime::UPDATE) ||
+        event > static_cast<lua_Integer>(IRTime::END)) {
+        throw sol::error{
+            std::string{fn} + ": invalid event " + std::to_string(event) +
+            " — must be IRTime.UPDATE / RENDER / INPUT / START / END"
+        };
+    }
+}
+
 // Bind `IRTime.UPDATE / RENDER / INPUT / START / END` as integer table
 // entries. The underlying `IRTime::Events` enum is C-style; `IR_BIND_TIME`
 // derives each key from the enum name via stringization (same pattern as
@@ -217,13 +232,7 @@ inline void bindRegisterPipelineAndSystemId(
     };
 
     lua["IRSystem"]["registerPipeline"] = [](lua_Integer event, sol::table ids) {
-        if (event < static_cast<lua_Integer>(IRTime::UPDATE) ||
-            event > static_cast<lua_Integer>(IRTime::END)) {
-            throw sol::error{
-                "IRSystem.registerPipeline: invalid event " + std::to_string(event) +
-                " — must be IRTime.UPDATE / RENDER / INPUT / START / END"
-            };
-        }
+        requireValidPipelineEvent("IRSystem.registerPipeline", event);
         std::list<IRSystem::SystemId> pipeline;
         for (auto &kv : ids) {
             sol::optional<lua_Integer> id = kv.second.as<sol::optional<lua_Integer>>();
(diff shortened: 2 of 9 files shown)


jakildev (commented):
## Review — system: append/insert a system onto a live pipeline (#1540) (engine #1548)

**Verdict:** approve

Adds `appendToPipeline` / `insertIntoPipelineBefore` / `insertIntoPipelineAfter` to `SystemManager` — composition primitives that add one system to an already-registered pipeline as its own singleton group, without disturbing existing systems. Exposes the surface as `IRSystem.appendSystem` / `insertSystemBefore` / `insertSystemAfter` in Lua. The motivating case is the midi runtime, where `initSystems()` builds the C++ pipeline before `main.lua` runs — `registerPipeline` would replace (and double-create) those systems; these compose.

### Checklist

**ECS invariants** — no per-entity `getComponent` in tick loops; no structural changes mid-iteration. All new methods (`appendToPipeline`, `insertSingletonGroupRelativeTo`) operate on `m_systemPipelineGroups` at setup time, not during tick execution. The double-add guard (`pipelineGroupsContain`) is a linear scan at setup time — not a per-frame cost. No `SystemName` enum entry needed (no new prefab system registered here).

**Ownership / lifetime** — `SystemId` is a plain integer; no raw owning pointers, no `shared_ptr`, no cross-frame captures. The pipeline vectors own their data by value. The `pipelineGroupsContain` anonymous-namespace helper closes over no external state.

**Render pipeline** — not touched. No shaders, no CPU-GPU structs, no binding points.

**Lighting** — not touched.

**Math / coordinates** — no math operations introduced.

**Serialization** — not touched.

**Naming / style** — `appendToPipeline`, `insertIntoPipelineBefore`, `insertIntoPipelineAfter`, `insertSingletonGroupRelativeTo`: all snake_case free functions consistent with existing `registerPipeline` / `registerPipelineGroups` style. No `m_` on public members, no reversed naming. `requireValidPipelineEvent` is an appropriate private helper name for a shared validation step.

**Tests / build** — two test files updated:
- `test/system/pipeline_groups_test.cpp` — six new tests covering: append at end, append to empty event (first-group creation), append-runs-alongside-prebuilt (acceptance case: both pre-built and appended systems tick), duplicate-append assert, insert-before/insert-after ordering (`a, before, b, after`), missing-anchor assert, no-pipeline-for-event assert.
- `test/script/lua_pipeline_register_test.cpp` — four new Lua-side tests: compose onto pre-built pipeline (group structure check), appended Lua system ticks (execution check with live entity), `insertSystemBefore` placement, invalid event rejection.
- PR body confirms `fleet-build --target IrredenEngineTest` clean; 82 tests green.

**Opus-only items** — no GPU buffer lifetime decisions (all setup-time); no archetype-mutation races (no structural ECS changes during tick); no async readback; no hot-path register pressure (new methods are not on the per-frame tick path).

### Nits

- `test/script/lua_pipeline_register_test.cpp` — `insertSystemAfter` Lua placement test absent (only `insertSystemBefore` is tested via Lua; both are covered at the C++ layer in `InsertBeforeAndAfterPlaceSingletonGroups`). Coverage is functionally complete through the C++ test, but a Lua-side `insertSystemAfter` placement assertion would round out the suite.

### Praise

- The `requireValidPipelineEvent` extraction eliminates the duplicated 6-line range-check block in both `registerPipeline` and `registerPipelineGroups` — a clean, naming-consistent refactor that makes the new bindings cheaper to write and easier to audit.
- The double-add assert (`pipelineGroupsContain`) is the right call: ticking the same system twice per frame would be a silent correctness bug (not a compile or link error), so the O(n) setup-path scan is the correct tradeoff.
- `appendToPipeline` using `operator[]` on `m_systemPipelineGroups` to default-construct an empty group sequence is a clean zero-line edge-case for the "first append to an empty event" path — no special branch needed.
- `InsertBeforeAndAfterPlaceSingletonGroups` verifies the full resulting order (`a, before, b, after`) — not just that the call didn't throw. The test is falsifiable.
- Docs update in `engine/system/CLAUDE.md` and `engine/script/CLAUDE.md` cross-reference each other by section name — a future creation author can find the surface from either end.

- [x] `fleet-build --target IrredenEngineTest` clean (confirmed in PR body)
- [x] All 6 new `PipelineGroupsValidatorTest` cases pass
- [x] All 4 new `LuaPipelineRegisterTest` cases pass
- [x] Full existing pipeline/system/script suite green (82 tests)

Opus recheck required: touches `engine/system/` (SystemManager + ir_system.hpp public surface).

jakildev:
Addressed nit: added `InsertSystemAfterViaLuaPlacesGroup` — Lua-side placement test for `insertSystemAfter`, mirroring the existing `InsertSystemBeforeViaLuaPlacesGroup`. All 18 `LuaPipelineRegisterTest` cases pass. — sonnet-fleet-1

jakildev (commented):
## Opus re-review — system: append/insert a system onto a live pipeline (#1540) (engine #1548)

**Verdict:** approve (with one nit) — confirming Sonnet's approve.

Standalone PR (`baseRefName` = master). Escalated because it touches `engine/system/` (SystemManager) and the public `ir_system.hpp` free-function surface — the model-split carve-out for Opus. Sonnet's checklist (ECS invariants, the `requireValidPipelineEvent` refactor, naming, the six C++ + four/five Lua tests) all holds; the author already addressed Sonnet's lone nit (`InsertSystemAfterViaLuaPlacesGroup` is now present and passing). I spent this pass on the three things a setup-time pipeline-mutation surface can get subtly wrong: the flatten/dirty cache coherence, the validator-skip claim for singleton groups, and the actual failure semantics behind the documented guards.

### Opus-only items

**Flatten cache coherence — confirmed.** `m_systemPipelineGroups` is `unordered_map<Events, vector<vector<SystemId>>>`; the flattened pipeline is a lazily-rebuilt `mutable` cache gated on `m_flattenedPipelinesDirty` (`system_manager.hpp:271-279`). All three new mutators (`appendToPipeline`, `insertSingletonGroupRelativeTo`) set `m_flattenedPipelinesDirty = true` after mutating the groups, exactly as `registerPipeline`/`registerPipelineGroups` do. No stale-flatten path. The `operator[]`-on-append vs `find()`-on-insert asymmetry is intentional and matches the docs (append seeds the first group; insert requires the anchor's pipeline to exist).

**Validator-skip-for-singletons — confirmed, not just asserted.** Sonnet said singleton groups need no cross-system validation because "the validator skips size-1 groups." On the closer read this is correct *and* sufficient: groups execute serially relative to one another and only systems *within* a group can co-execute, so a singleton group never runs concurrently with anything — there is no access pair to validate. Important corollary the PR gets right: `validateAllPipelineGroups()` normally runs once at setup, *before* `main.lua`, so a Lua-appended system is never validated — but since it is always its own singleton group, there is nothing to validate. Safe by construction.

**The documented guards are debug-only — this is the one finding (nit).** Sonnet reported these as "Raises a Lua error if the system is already in the pipeline" / "if the anchor isn't in the pipeline," and the script-facing docs say the same (`engine/script/CLAUDE.md`, the binding comments). That is true **only in debug builds.** Both guards are `IR_ASSERT`, which (`ir_profile.hpp:81-127`) throws `std::runtime_error` when `IR_RELEASE` is undefined — sol2 surfaces that as the Lua error the tests assert via `EXPECT_THROW` — but expands to **nothing** under `IR_RELEASE`. In a release build:
- a duplicate `appendSystem` silently **ticks the system twice per frame** (the exact bug the assert message warns about), and
- `insertSystemAfter/Before` with a bad anchor leaves `found == false`, the stripped `IR_ASSERT(found, …)` no longer fires, and execution falls through to insert at the **default `anchorGroup` index 0** — a silent front-insert rather than a no-op.

The C++ header comments are precise (they say "**Asserts** if…", and assert = debug-only is the established engine convention). Only the Lua-facing wording overstates the guarantee. This matches engine-wide `IR_ASSERT` philosophy and the fleet builds/tests in debug, so I'm calling it a **nit, not needs-fix** — but it's worth tightening because the public consumers here are creation authors reading `script/CLAUDE.md`. See Nits.

**Reentrancy — pre-existing, non-blocking.** Like `registerPipeline`, these mutate `m_systemPipelineGroups` and are not safe to call from inside a tick of the same event being executed (the in-flight flattened list won't reflect the change until the next refresh). The intended call site is setup-time composition (midi `initSystems()` → `main.lua`), so this is fine; flagging only so it isn't later mistaken for a runtime hot-swap primitive. No change requested.

### Nits

- `engine/script/CLAUDE.md` (the `appendSystem`/`insertSystemBefore`/`After` bullets) and the inline binding comments in `lua_pipeline_bindings.hpp` — the double-add / missing-anchor protection is a **debug-only assert**, not an always-on Lua error like the (correctly always-on) `requireValidPipelineEvent` range check. Recommend wording it as "asserts in debug; in release a duplicate ticks twice / a bad anchor front-inserts" so a creation author doesn't lean on it as release-time input validation. If release-safe rejection at the scripting boundary is actually wanted, promoting the double-add check to a hard `throw sol::error` (mirroring `requireValidPipelineEvent`) would do it — but that's a design call, not a merge blocker.
- `system_manager.hpp` / `ir_system.hpp` carry clang-format-only reflows of `createSystem` / `MakeMemberTickFn` (if-constexpr condition wrapping, `emplace_back` brace placement) unrelated to #1540. Harmless (no semantic change, verified), just incidental churn in a templated hot-path header.

### Net

Implementation is correct in debug — where the fleet builds and tests — and the tests are thorough and falsifiable (group structure, live-execution proof, and all three assert paths at both the C++ and Lua layers). The only substantive Opus-only observation is documentation precision around release-mode assert stripping. Sonnet flagged this for Opus recheck on the `engine/system/` + `ir_system.hpp` surface; on the closer read I **confirm the approve**, downgrading nothing to needs-fix.

jakildev:
Addressed Opus nit: the `appendSystem`/`insertSystemBefore`/`insertSystemAfter` double-add and missing-anchor guards are `IR_ASSERT` (debug-only), not always-on Lua errors. Updated wording in `engine/script/CLAUDE.md` and the inline comment in `lua_pipeline_bindings.hpp` to say 'asserts in debug; in release a duplicate ticks twice / a bad anchor silently front-inserts'. `engine/system/CLAUDE.md` already had the correct 'debug; no-op in release' wording and was not changed. — sonnet-fleet-2

Issue #1555, opened 2026-06-06
render: detached re-voxelize — CPU model proof + asymmetric demo (P1)


jakildev:
**Model:** opus
**Part of epic:** #1553
**Plan file:** `~/.fleet/plans/issue-1553.md` (full epic plan + architecture decisions)
**Blocked by:** (none)

## Scope
Stand up the detached re-voxelize render path end-to-end with a CPU cell-fill, and prove it meets the true-3D criterion on an asymmetric solid. The headline-proving phase of the epic.

## Approach
- Add a detached re-voxelize mode/opt-in (a new `RotationMode` value or a detached sub-flag — implementer picks the cleanest plumbing; reuse GRID semantics where possible). Opting in routes the entity through re-voxelize instead of the octahedral-snap + forward-scatter / per-axis path.
- Each frame, for a re-voxelize detached entity, fill its **private pool's** global cell positions via the reused `IRPrefab::GridRotation::worldCellForGridVoxel` math under the entity's **full** rotation (CPU; same shape as `SYSTEM_REBUILD_GRID_VOXELS`, but writing the detached canvas's private pool, not the shared world pool). Extend that system to handle the new mode for detached canvases, or add a sibling — do not duplicate the transform math.
- Route the re-voxelize detached canvas through **cardinal/static raster frame data**: `visualYaw_ = 0`, no residual, no per-entity skew — the canvas renders its pool exactly like a static canvas; the rotation is already in the cell positions. (This is the crux — see Gotchas.)
- Bypass the off-snap octahedral single-canvas emit + per-axis forward-scatter for re-voxelize-opted entities (no double-draw).
- Composite screen-locked via the **existing gather blit TRS** (`system_entity_canvas_to_framebuffer.hpp`, unchanged).
- **Add an asymmetric (non-cube) detached rotating test solid to `canvas_stress`** (e.g. an L-tromino / slab / stepped solid, 3-DOF spin) so the discriminating shot exists — there is no asymmetric rotating consumer in current content.

## Acceptance criteria
- The asymmetric detached solid at ~45° between octahedral snaps reads as a **true 3D-rotated solid (voxel centers reorganized)**, not a 2D-skewed cardinal arrangement.
- Clean, cohesive solid at **every** residual pose across a full SO(3) spin; smooth through snap intervals, **no pop** (absorbs #1539 for these entities).
- A detached **cube** via re-voxelize is also clean at every pose (validates the cube case the epic uses to supersede #1551).
[...]

PR #1573, opened 2026-06-07
render: re-voxelize exposed-mask frame fix — emit 3 cardinal faces (#1570)

jakildev:
## Summary

Fixes the re-voxelize "faces that should clearly be visible get dropped / mis-coloured as the solid spins" defect — the **root cause** behind #1570's Defect 3.

**Root cause (regression from P2 / #1556).** A `DETACHED_REVOXELIZE` canvas bakes the entity's full rotation into integer **cell positions** (GPU scatter `c_revoxelize_detached`, binding 5). But the per-voxel exposed mask in `C_Voxel.flags_` (binding 6) is computed **once at authoring time** in the *unrotated* model frame (`recomputeFaceOccupancy`) and is never recomputed against the rotated cells:

- **P1 (#1555)** re-rasterized the cells *and* re-derived the mask every frame on the CPU.
- **P2 (#1556)** moved the cell fill to the GPU scatter and **dropped the per-frame mask recompute without replacing it** — the mask was left stale in the authoring frame.

`c_voxel_to_trixel_stage_{1,2}` still gate emission on `faceIsExposed(flags, faceId)`, so they test **rotated-frame** faces against an **unrotated** occlusion mask. At any non-identity rotation the two frames disagree, so whole camera-visible faces get gated off (and the wrong cell wins the depth tie, mis-colouring the surface). Three separate comments in the codebase still *asserted* the mask was "recomputed against the rotated cells each frame" — the smoking gun.

**Fix.** For re-voxelize cells (`visibleFaceIds.w != 0`) **bypass the model-frame gate** and emit all three visible-triplet cardinal faces (`X_NEG/Y_NEG/Z_NEG`), letting the depth re-test (`imageAtomicMin`) keep the front-most surface. A convex/centred solid then shows exactly its three iso faces with no holes; interior cells lose the depth tie. World / GRID / per-axis / forward-scatter paths keep the real exposed-mask gate and stay **byte-identical** (the bypass is gated on the existing re-voxelize marker). Delivers the "guarantee 3 iso faces" behaviour chosen for D3.

Also corrects the three stale "mask is recomputed on the rotated cells" comments (GLSL + Metal `stage_{1,2}`, `system_voxel_to_trixel.hpp`, and the demo seed).

## Before / after (macOS / Metal, multi-color L-prism at a spun pose)

| before (master) | after (fix) |
|---|---|
| before | after |

Before has a ragged top silhouette (stray orange speckle on the top-left edge), noisier gradients, and a less-defined L-carve notch; after has clean edges, smooth gradients, and a crisp notch. Full-frame `before` / `after` / `diff` for the zoom-2.2 spun shot are in `docs/pr-screenshots/`.

## Verification (macOS / Metal, `IRCanvasStress`)

- The fix changes **only** the multi-color re-voxelize solid — uniform re-voxelize cubes and all non-re-voxelize entities are unchanged (a uniform solid hides the defect, which is exactly why the demo's discriminator is multi-color).
- **Scales with rotation**, confirming the frame-mismatch mechanism: `revoxelize_zoom` drift = **0.67%** frozen (`--no-spin`) vs **1.66%** at a spun pose.
- `master@120` vs `fix@120` at the reference pose = **0.077–0.098%** per shot, within the suite's 99.9% / tol-8 tolerance.

## References

The `test/references/macos-debug/so3_*` set was **already stale on master** by ~2.3% (P2/P3 changed re-voxelize rendering on Jun 6–7 without a reference refresh); this fix adds only ~0.09% on top. Refreshed the **macOS** references with the corrected output. ⚠️ The `linux-debug` references need the same refresh on a Linux/OpenGL host — I can't produce OpenGL output on macOS.

## Scope / follow-ups

- D3 root cause only. **#1571** carries D1 (spin flicker — accept + document) and D2 (zoom-clip — `subdivisionDensity()` cap).
- Optional later optimisation: recompute the exposed mask against the rotated cells on the GPU (occupancy pass) to cut the interior over-emit the bypass accepts. Not needed for correctness.
- Backend parity: GLSL + Metal both updated here; OpenGL build/smoke + the Linux reference refresh happen on a Linux host (`fleet:needs-linux-smoke`).

Part of epic #1553. Refs #1570, #1571.


jakildev (commented):
## Review — render: re-voxelize exposed-mask frame fix — emit 3 cardinal faces (#1570)

**Verdict:** approve

### Summary

Root-cause fix for the rotated re-voxelize face-drop / mis-colour defect (#1570 Defect 3). The diagnosis is correct and well-documented: P2 (#1556) moved the CPU cell fill to the GPU scatter but dropped P1's per-frame mask recompute, leaving `C_Voxel.flags_` in the unrotated authoring frame while the cell positions baked the entity's full rotation. The fix bypasses the exposed-face gate for re-voxelize cells in all four shader files (GLSL stage_1 + stage_2, Metal stage_1 + stage_2) and lets the existing depth re-test (`imageAtomicMin` / `writeColorTap`) discard interior-cell over-emit.

### Checklist

**ECS invariants** — No ECS code changed. Pass.

**Ownership / lifetime** — No C++ structural changes; comment-only in `system_voxel_to_trixel.hpp` and `main.cpp`. Pass.

**Render pipeline**
- CPU/GPU struct alignment: no struct changes. Pass.
- Binding points: no changes. Pass.
- Shader file naming: no new files. Pass.
- GLSL + Metal parity: all four shader files updated symmetrically. Pass.
- `fleet:needs-linux-smoke` already set (macOS-authored). `fleet:needs-windows-smoke` missing — adding below.

**Correctness of the bypass logic:**
```glsl
// Before (in all 4 shaders)
if (!faceIsExposed(flagsByte, faceId)) return;
const bool reVoxelize = visibleFaceIds.w != 0;

// After
const bool reVoxelize = visibleFaceIds.w != 0;
if (!reVoxelize && !faceIsExposed(flagsByte, faceId)) return;
```
- Non-re-voxelize canvases (`reVoxelize == false`): gate is `!false && !faceIsExposed(...)` = `!faceIsExposed(...)` — byte-identical to master. ✓
- Re-voxelize canvases (`reVoxelize == true`): gate is `!true && ...` = `false` — always passes, all 3 visible-triplet faces emitted. ✓

**Interior over-emit handling:** Stage 1's `imageAtomicMin` writes the minimum (closest) depth per trixel; surface cells win. Stage 2's `writeColorTap` depth-compares against stage 1's result; interior cells wrote larger depths than their surface siblings, so stage 2's color writes for interior cells fail the depth test and are discarded. The mechanism is pre-existing and correct here. ✓

**Comment quality:** All three stale "mask is recomputed on the rotated cells each frame" comments corrected (GLSL stage_1, GLSL stage_2, Metal stage_1, Metal stage_2, `system_voxel_to_trixel.hpp`, `main.cpp`). The new comments explain WHY the bypass is needed (authoring-frame vs rotated-frame mismatch) and call out the future GPU mask recompute optimization path. ✓

**Lighting** — No lighting systems touched. Pass.

**Math / coordinates** — No math changes. Pass.

**Serialization** — No asset changes. Pass.

**Tests / build**
- macOS reference images refreshed (5 PNGs). Linux references acknowledged as stale and deferred to Linux host (`fleet:needs-linux-smoke`). ✓
- Before/after screenshots in `docs/pr-screenshots/` confirm visual improvement. ✓

### Praise

- **Minimal surgical footprint** — 4 shader files + 2 comment-only files + reference refresh. No collateral changes.
- **Correct scoping of non-re-voxelize paths** — byte-identical to master confirmed both analytically (gate condition) and by the PR's own tolerance measurements (0.077–0.098% at reference pose).
- **Well-documented regression** — the "smoking gun" stale-comment pattern (3 places asserted the mask was current when it wasn't) makes the root cause clear and the fix reviewable independently.
- **Future optimization noted** — the over-emit from interior cells is flagged as a known cost with a clear path (GPU rotated-cell mask recompute) without blocking correctness now.

- [x] macOS / Metal: `IRCanvasStress` multi-color re-voxelize solid at non-identity rotation — clean iso faces with no holes.
- [ ] Linux / OpenGL: same test + reference refresh (pending `fleet:needs-linux-smoke`).
- [x] Uniform re-voxelize cubes and non-re-voxelize entities unchanged (bypass gated on `visibleFaceIds.w != 0`).

Opus recheck not required.

Issue #1584, opened 2026-06-08
render: detached SO(3) + GRID rotation artifacts (snap, missing faces, non-rotating entity, screen-fill) — tracking


jakildev:
Tracking issue for the rotation/render artifacts in `canvas_stress` (detached SO(3) + GRID paths), surfaced after the re-voxelize + orbit work landed. Each gets a focused fix PR with before/after crops.

## A. Small GRID cubes "shred" / missing faces under rotation — **triaged: accepted aliasing, not a new bug**
The shredded small center cubes are the **GRID spin cluster** (4 × 5³ cubes, `kGridSpinColors`), not the forward-scatter detached cubes. Confirmed by isolation:
- **Not the per-axis budget** — bumping `kMaxDetachedRotatingCanvases` 8→24 changed nothing.
- **Not the #1574 GRID mask recompute** — disabling the recompute, still shredded.
- **Root cause:** round-to-cell aliasing on a 5³ cube (too few cells → heavy aliasing when rotated). This is the documented "accepted by design" GRID aliasing (`voxel/CLAUDE.md`). Reproduces at camera yaw 0 *and* 45° (entity-spin-driven, not camera-yaw).
- **Options:** bigger spin cubes (more cells → less visible aliasing) — a demo tweak; or accept. Not an engine bug.

## B. Detached SO(3) entities snap / not fully smooth at certain rotation phases — **investigate**
Octahedral-snap-residual transitions (`syncAllocationToDetachedEntities` deadband, at-snap single-canvas ↔ off-snap per-axis scatter). Likely a visible pop at the snap boundary.

## C. A detached entity appears not to rotate / face positions+deforms not applied — **investigate**
A specific entity whose `faceDeform` / per-axis store looks inert. Possibly a degenerate axis/rate, a cube-symmetry pose reading as "static", or an entity stuck on the wrong path.

## D. A detached entity (pink cube) renders a huge quad across the screen for a frame or two — **investigate**
Sporadic. Strongly suggests a **degenerate/near-singular `faceDeformationMatrixSO3`** (edge-on face) or a NaN/inf producing a giant scatter quad in `v_peraxis_scatter_detached` / `ENTITY_CANVAS_TO_FRAMEBUFFER`. Needs a guard/clamp.

## Notes
- A `canvas_stress` shadow floor (uncommitted) was the debugging vehicle; it also exposed E. The floor itself is sun-lit but cast shadows don't render on the SDF floor (separate snag, parked).
[...]

PR #1581, opened 2026-06-07
render: detached re-voxelize — render-verify + retire forward-scatter (P6) (#1560)

jakildev:
## Summary

Final phase (**P6**) of the detached re-voxelize epic (#1553). Now that the re-voxelize path (#1555–#1559) renders detached SO(3) by rotating+rounding the entity's voxel **cells** into a private pool, the interim **per-axis forward-scatter** is superseded — this PR retires it so re-voxelize is the **sole** detached SO(3) renderer.

Stacked on: https://github.com/jakildev/IrredenEngine/pull/1580 (#1559, P5 parity audit).

### Removed (detached consumer only)
- `v_peraxis_scatter_detached.{glsl,metal}` + the `PerAxisScatterDetachedProgram` registration & scatter pass in `ENTITY_CANVAS_TO_FRAMEBUFFER`.
- Detached per-axis allocation: `syncAllocationToDetachedEntities`, `kMaxDetachedRotatingCanvases`, `kDetachedResidualDeadband`.
- Detached forward-scatter store wiring in `VOXEL_TO_TRIXEL_STAGE_1`: `detachedPerAxisAllocated_`, `lookupDetachedPerAxis`, the per-axis store dispatch + the single-canvas-skip branch.
- The now-dead `perAxisActive` parameter (only gated the detached forward-scatter face-selection) from `buildVoxelFrameData` / `authorIteratingCanvasVoxelFrame` / `restoreMainCanvasVoxelFrame` and their AO/lighting call sites.

### Kept byte-identical (camera path)
- `syncAllocationToCameraYaw`, `dispatchPerAxisCanvases`, the per-axis AO/lighting passes — the **camera/main-canvas** smooth-Z-yaw path is untouched.
- `FrameDataTrixelToFramebuffer.detachedResidual_ / detachedDepthAxis_` are kept as **reserved std140 padding** (offsets 144/160) so `scatterFbResolution_` stays at the shared offset 176 the camera scatter shader (`v_peraxis_scatter.glsl`, `_detachedResidualPad`/`_detachedDepthAxisPad`) reads — no UBO reflow, camera path stays byte-identical. Static asserts intact.

### Behaviour note — `RotationMode::DETACHED`
[...]

PR #1586, opened 2026-06-08
render: guard detached per-axis scatter against degenerate screen-fill (#1584 D)

jakildev:
## Summary

Defensive fix for #1584 item D — the sporadic "a detached cube renders a large part of the screen for a frame or two" glitch.

**Root cause.** The detached SO(3) forward-scatter (`v_peraxis_scatter_detached`) grows each face quad by a conservative-coverage margin that **scales with the projected cell pitch**:
```
detachedMargin = max(kScatterDilateMarginPx, kScatterDetachedPitchFraction * pitchPx)
```
At a rare residual / placement pose the projected pitch (and thus the margin) can explode, so the seam-closing dilation **smears one cell's quad across the whole framebuffer** for a frame.

**Fix.** A single forward-scattered face cell can never legitimately project wider than the framebuffer, so degenerate the instance off-screen — the same `(2,2,2,1)` clip-out the empty-cell branch already uses — when `pitchPx` is NaN or `>=` the framebuffer extent, or when `clipCorner` came out non-finite. GLSL + Metal both guarded.

## Verification
- Normal detached entities render **identically** on macOS/Metal (the guard threshold — a cell wider than the whole framebuffer — is ~50× a normal cell pitch, so it can't false-trigger on real geometry). No shader-compile errors; demo runs clean.
- The triggering pose is **rare and not deterministically reproducible**, so this is a *provably-correct degenerate-output guard*, not a repro-confirmed root-cause fix. If the underlying pitch explosion is later pinned to a specific residual construction (`octahedralSnapResidual` / `pos3DtoPos2DIsoRotated`), it can be addressed upstream too — noted on #1584.

## Scope
GLSL + Metal scatter shader only; no CPU change. Addresses **#1584** (D). The other #1584 items (B snap, C non-rotating entity) remain to investigate.

### DIFF
--- a/engine/render/src/shaders/metal/peraxis_scatter_detached.metal
+++ b/engine/render/src/shaders/metal/peraxis_scatter_detached.metal
@@ -109,6 +109,16 @@ vertex VertexOut v_peraxis_scatter_detached(
     // gap (which scales with pitch) closes at every scale without blobbing small cubes.
     // One sqrt of the larger squared edge — not two length() calls.
     const float pitchPx = sqrt(max(dot(su, su), dot(sv, sv)));
+    // Degenerate-projection guard (#1584 D) — see the GLSL twin. A rare residual
+    // / placement pose can explode pitchPx (and the dilation margin it drives)
+    // into a screen-filling smear for a frame, or make clipCorner non-finite.
+    if (!(pitchPx < max(fbRes.x, fbRes.y)) ||
+        any(isnan(clipCorner)) || any(isinf(clipCorner))) {
+        out.position = float4(2.0, 2.0, 2.0, 1.0);
+        out.color = float4(0.0);
+        out.depth = 1.0;
+        return out;
+    }
     const float detachedMargin = max(kScatterDilateMarginPx, kScatterDetachedPitchFraction * pitchPx);
     clipCorner.xy += scatterConservativeDilation(su, sv, sign(in.position), detachedMargin, ndcPerPx);
     clipCorner.y = -clipCorner.y;
--- a/engine/render/src/shaders/v_peraxis_scatter_detached.glsl
+++ b/engine/render/src/shaders/v_peraxis_scatter_detached.glsl
@@ -125,6 +125,19 @@ void main() {
     // small ones (a fixed bump big enough for large cubes over-grows small).
     // One sqrt of the larger squared edge — not two length() calls.
     float pitchPx = sqrt(max(dot(su, su), dot(sv, sv)));
+    // Degenerate-projection guard (#1584 D). A single forward-scattered face cell
+    // can never legitimately project wider than the framebuffer; at a rare
+    // residual / placement pose `pitchPx` (and the conservative-dilation margin
+    // it drives) can explode and smear this quad across the whole screen for a
+    // frame, or `clipCorner` can come out non-finite. Drop the instance instead
+    // — same off-screen degenerate the empty-cell branch uses.
+    if (!(pitchPx < max(fbRes.x, fbRes.y)) ||
+        any(isnan(clipCorner)) || any(isinf(clipCorner))) {
+        gl_Position = vec4(2.0, 2.0, 2.0, 1.0);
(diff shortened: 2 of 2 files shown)


jakildev (commented):
## Review — render: guard detached per-axis scatter against degenerate screen-fill (#1584 D) — engine #1586

**Verdict:** approve

### ⚠️ Merge-order note (human action required)

PR #1581 (`claude/1560-retire-forward-scatter`, now `fleet:approved`) **deletes** the two files this PR guards (`v_peraxis_scatter_detached.glsl` and `peraxis_scatter_detached.metal`). Merge order matters:

- **Merge #1586 first, then #1581** — the guard lands briefly, then #1581 retires the shaders cleanly. No conflict.
- **Merge #1581 first** — this PR becomes a deletion conflict (adding to nonexistent files). Would need to be closed as superseded.
- **Close #1586 without merging** — also valid if the sporadic screen-fill glitch is considered acceptable for the window before #1581 lands.

The code in #1586 is correct (see below); the decision is solely about merge order vs. "skip it since retirement is imminent."

### Code review

**Guard logic** (GLSL):
```glsl
if (!(pitchPx < max(fbRes.x, fbRes.y)) ||
    any(isnan(clipCorner)) || any(isinf(clipCorner))) {
    gl_Position = vec4(2.0, 2.0, 2.0, 1.0);
    vColor = vec4(0.0);
    vDepth = 1.0;
    return;
}
```

- `!(pitchPx < max(fbRes.x, fbRes.y))` — handles both "pitchPx >= framebuffer extent" AND "pitchPx is NaN" (since `NaN < x` is false in IEEE 754, so the negation is true). Both cases are correctly caught by this single condition. ✓
- `any(isnan(clipCorner)) || any(isinf(clipCorner))` — checks the raw projection result before dilation is applied; correct placement (before `clipCorner.xy += scatterConservativeDilation`). ✓
- Degenerate output `(2.0, 2.0, 2.0, 1.0)` — clip-space position outside NDC range `[-1, 1]`, guaranteed off-screen; same degenerate used by the existing empty-cell branch. ✓
- `vDepth = 1.0` — maximum depth, won't win depth composite. ✓

**Insertion context verified**: `fbRes` (as `vec2 max(scatterFbResolution.xy, 1)`) and `pitchPx` are both in scope at the guard point; `clipCorner` is the pre-dilation raw projection result — exactly what we want to inspect. ✓

**Metal parity**: Identical guard condition and output values in `peraxis_scatter_detached.metal`. `isnan(float4)` / `isinf(float4)` and `any(bool4)` are valid Metal built-ins. ✓

**Guard conservativeness**: A normal cell pitch is tens of pixels; the threshold (`max(fbRes.x, fbRes.y)`, typically 1920–2560 px) is ~50× larger than any real cell — no false-positive risk on valid geometry. ✓

### Checklist

**ECS invariants**: Shader-only change, no ECS code. ✓
**Ownership / lifetime**: No CPU change, no new allocations. ✓
**Render pipeline**: No CPU/GPU struct change; no UBO layout impact. ✓
**GLSL/Metal parity**: Both backends guarded identically. ✓
**Naming / style**: No new identifiers. ✓

### Cross-host smoke

PR authored on macOS (`fleet:authored-on-macos`). Touches `*.glsl` (GLSL/OpenGL backend) → needs Linux smoke. `fleet:needs-linux-smoke` already labeled. ✓

Opus recheck not required.

jakildev:
Superseded by #1581 (P6 — retire forward-scatter). #1581 **deletes** `v_peraxis_scatter_detached.{glsl,metal}` — the exact files this guard modifies — and removes the entire forward-scatter detached path that produced the screen-fill (D). So the degenerate-projection source is gone, and this guard would conflict (modifying deleted files). Closing in favour of #1581. If a degenerate screen-fill ever reappears in the surviving single-canvas `faceDeformationMatrixSO3` path (it clamps super-sample n≤6, so it shouldn't smear a quad the same way), it'd be a fresh issue. — architect

PR #1598, opened 2026-06-09
scripts/fleet: add fcntl mutex to reconcile persistence read-modify-write cycles (#1367)

### LINKED ISSUE #1367: fleet-claim reconcile_escalate_drift: no mutex on concurrent --apply runs
## Context

Flagged as a nit in PR #1365 review.

`reconcile_escalate_drift` in `scripts/fleet/fleet-claim` (lines ~2050-2176) has no mutex. Two concurrent `reconcile --apply` runs (possible when a long cleanup run overlaps the next projection-change trigger) both read the same `drift-persistence.json`, increment independently, and the last write wins — losing one tick's increment.

**Worst case**: escalation delayed by one tick. No corruption (Python write is a full overwrite, not an append).

**Why escalating**: Low risk at the current 30s tick cadence, but observable if the cadence tightens or if cleanup runs routinely overlap. Reviewer deferred it; filing per their guidance ("worth a follow-up issue if the cadence tightens").

## Suggested approach

- Add a lockfile guard around the `read-increment-write` cycle in `reconcile_escalate_drift`, e.g.:
  ```bash
  local lock_file="$state_dir/drift-persistence.lock"
  (
    flock -x 9
    # ... python3 capture + body_file write ...
  ) 9>"$lock_file"
  ```
- Or move the counter logic to a single atomic Python script call that uses `fcntl.flock` internally.
- Keep backward compat: the lock file can be created lazily.

**Model:** sonnet  
**Area:** scripts/fleet/fleet-claim — reconcile_escalate_drift  
**Blocked by:** (none)

jakildev:
## Summary
- Both `reconcile_escalate_drift` and `reconcile_heal_design_unblock` had an unguarded read-increment-write cycle on their persistence JSON files; two concurrent `--apply` runs could race and lose one tick's increment on last-write-wins
- Fix wraps the critical section in `fcntl.flock(LOCK_EX)` on a sidecar `.lock` file (created lazily at the same path as the JSON + `.lock` suffix)
- Uses Python's `fcntl` module rather than the `flock(1)` CLI tool — `flock(1)` is absent on macOS, `fcntl` is available on both Linux and macOS

## Test plan
- [ ] Verify `fleet-claim reconcile --apply` still runs cleanly (dry run and live)
- [ ] Confirm the `.lock` sidecar files exist in `state_dir` and are zero-length; a separate process can immediately acquire `LOCK_EX` on them (no stale holder)
- [ ] Manual concurrent test: launch two `reconcile --apply` runs and confirm the final persistence JSON counts are consistent

Closes #1367

### DIFF
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -2349,7 +2349,7 @@ reconcile_escalate_drift() {
     local persistent_count
     persistent_count=$(NOW="$now" THRESHOLD="$threshold" PERSIST="$persist_file" \
         BODY="$body_file" python3 - "$findings" <<'PYEOF'
-import sys, os, json
+import sys, os, json, fcntl
 findings_path = sys.argv[1]
 now = int(os.environ["NOW"])
 threshold = int(os.environ["THRESHOLD"])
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — scripts/fleet: add fcntl mutex to reconcile persistence read-modify-write cycles (#1367)

**Verdict:** approve

### Praise

- **Correct race-condition fix.** The unguarded read-increment-write on `drift-persistence.json` and `design-unblock-heal-persistence.json` was a real hazard: two concurrent `--apply` runs on the same host would last-write-win and silently drop one tick's worth of increments. The `fcntl.flock(LOCK_EX)` wrapper around the full critical section (load → compute → write) is the right fix.
- **Cross-platform tool choice.** Using Python's built-in `fcntl` module rather than the `flock(1)` CLI tool is correct — `flock(1)` is a Linux utility absent on macOS, while `fcntl` is standard library on both. The PR body explicitly calls this out, and the label `fleet:authored-on-macos` confirms the author validated it there.
- **Lock scope and isolation.** The lock is scoped to the minimum critical section only; work outside the `with` block (sorting `persistent`, body formatting, the GH issue mutation) is correctly excluded. The two functions each hold their own sidecar `.lock` file (separate `persist_path` values → `drift-persistence.json.lock` vs `design-unblock-heal-persistence.json.lock`), so there is no cross-function deadlock risk.
- **Only writers, fully covered.** A search confirms no other site in `fleet-claim` writes to either persistence file; both writers are now guarded.

### Nits

- **scripts/fleet/fleet-claim (PR body test plan)** — The test plan item "Confirm no new `.lock` files are left behind after a reconcile pass completes" is misleading. The sidecar `.lock` files (`drift-persistence.json.lock`, `design-unblock-heal-persistence.json.lock`) ARE intentionally left on disk — that is how `fcntl`-based file locking works (the lock is on the open file handle, not file existence). A more accurate check would be: "Confirm the `.lock` sidecar files exist in `state_dir` and are zero-length; a separate process can immediately acquire `LOCK_EX` on them (no stale holder)." The current wording will cause a false-alarm if someone runs the test plan literally. Not a code issue — the lock is correct — but worth fixing before the PR body becomes a reference.

- [x] Verify `fleet-claim reconcile --apply` still runs cleanly (dry run and live)
- [ ] Manual concurrent test: launch two `fleet-claim reconcile --apply` runs simultaneously and confirm the final persistence JSON counts are consistent (no missing increments)
- [ ] Verify sidecar `.lock` files are created in `state_dir` and can be immediately re-locked (no stale exclusive holder) after the run exits

Opus recheck not required.

jakildev:
Addressed nit: updated PR body test plan to clarify that sidecar `.lock` files are intentionally left on disk (fcntl semantics — lock is on the open file handle). Replaced 'no new .lock files are left behind' with 'confirm sidecar .lock files exist and are zero-length; a separate process can immediately acquire LOCK_EX on them (no stale holder)'. No code changes. — sonnet-author

Issue #1458, opened 2026-06-02
render: decouple per-axis lattice resolution from effSub (restore high-zoom rotating detail; follow-up to #1431)


jakildev:
Follow-up deferred from #1431 / PR #1452. #1431 fixes the **clip** (content disappearing) via Option 1 — capping `subPerAxis` on the rotating per-axis store path so cells never overflow the bounded canvas. That trades away sub-voxel detail at high zoom **while rotating**.

This issue tracks the detail-preserving end-state: store the per-axis lattice at a **bounded base/world resolution** and carry sub-detail as a **fractional micro-offset recovered at scatter time** (à la the cardinal `faceMicroPositionFixed` path), instead of expanding the #1310 integer lattice by `effSub`.

## Architect resolution (2026-06-09)
Design question resolved (was PR #1613 NEEDS-DESIGN): the detail goal is to **smooth the per-base-cell WINNER only** (a fractional offset for the single occlusion winner), **not** retain multiple distinct sub-voxels per base cell — the latter contradicts bounded base-resolution storage + per-base-cell occlusion. Direction is **targeted fixes, no epic**. This issue **subsumes #1469** (closed): the base-resolution store removes the `× effSub` depth-range overflow at its root, so no cap term / readback is needed and the `zoom4_yaw45*` baselines stay valid. Full plan: `~/.fleet/plans/issue-1458.md`.

**Blocked by:** #1457

(#1457 first establishes the decoupled occlusion model — yawed winner / un-yawed recovery — that this rework inherits; doing them independently would collide on the same `atomicMin` key.)

## Scope of the rework
- Reworks the #1310 per-axis store + `faceOriginFromInPlane` scatter recovery (now fractional).
- atomicMin occlusion per **base** cell, on the decoupled key from #1457.
- Removes the `× effSub` lattice expansion (subsumes #1469's deep-scene clip).
- Must keep cardinal byte-identity and GLSL/Metal parity; audit every per-axis reader + Metal twin.

`[opus]`.

PR #1600, opened 2026-06-09
render: per-axis cap depth-range term — WIP/design-blocked (#1469)

### LINKED ISSUE #1469: render: perAxisSubdivisionCap depth-range term residual — deep scenes may still clip at non-cardinal yaw
## Context

Escalated from PR #1452 (engine PR for issue #1431 — per-axis canvas sizing under camera Z-yaw).

The Opus final-pass reviewer (role-opus-reviewer) approved PR #1452 but recorded a latent nit that neither the Sonnet nor the first Opus pass surfaced.

## Problem

\`perAxisSubdivisionCap\` (ir_math.hpp) derives \`maxOnScreenWorldDisp\` purely from the depth-0 inverse of the visible iso rectangle (the dominant term for the #1431 gallery clip case). But an on-screen voxel at non-zero depth \`d = x+y+z\` sits at \`world = W₀ + (d/3)·(1,1,1)\`, and that \`(d/3,d/3,d/3)\` shift has a non-zero component along every face's in-plane (tangent) axes. For a scene with a large visible depth range, face-local cells can therefore still land outside the capped canvas and clip — the cap fully covers the iso-rect spread but not the additive depth-range term.

This is **not a regression** from master (before #1431 there was no cap at all; the high-\`effSub\` clip was unconditional). \`isInsideCanvas\` drops any residual OOB cell to background — no corruption. The #1431 gallery case is genuinely fixed. The residual only appears in scenes with a large visible depth range combined with high \`effSub\`.

## Fix direction

Two options (pick one):

1. Fold the depth-range term into \`maxOnScreenWorldDisp\`: \`maxOnScreenWorldDisp += (visibleDepthRange/3) × maxInPlaneUnitProjection\`, where \`visibleDepthRange\` is the depth span of on-screen geometry (extractable from the canvas depth texture or bounded from the cardinal extent + zoom). Tight but needs the visible depth span at the call site.
2. Clamp the visible depth span feeding the cap — bound \`d/3\` by \`(canvasHalf.y / (kSqrt2 · effSub))\` and add the resulting contribution to the max displacement before dividing. Bounded in terms of canvas geometry alone; cheaper than a texture readback.

## Why escalating

The fix requires extending the mathematical derivation in \`perAxisSubdivisionCap\` and potentially threading a new value (visible depth range or a canvas-geometry-derived bound) through the cap call. Both directions need a correct closed-form derivation against \`isoPixelToPos3D\` for review confidence — Opus-tier reasoning warranted.

**Model:** opus
**Area:** engine/math, engine/prefabs/irreden/render
**Blocked by:** (none)

Escalated from sonnet. The existing approval on PR #1452 stands; this issue tracks the residual. — sonnet-fleet-2

jakildev:
WIP for #1469 — **design-blocked**, see the `## NEEDS-DESIGN` comment below.

## What this implements (Option 2)
Folds the depth-range term into `IRMath::perAxisSubdivisionCap` so the per-axis
store stops clipping deep scenes under camera Z-yaw. A visible voxel at iso depth
`d = x+y+z` sits at `world0(iso) + (d/3)·(1,1,1)`; `faceInPlaneCoords` selects two
of three equal components, so the shift adds `d/3` to **both** canvas axes of every
face — the additive term the depth-0-only cap missed (#1431 covered only the iso-rect
spread). `d` is scene-dependent/unbounded and the cap is CPU-side with no depth read,
so the depth contribution is bounded (readback-free) by the canvas-storable depth at
the uncapped `effSub`: `depthAllowance = canvasHalf.y / (√2·effSub)`.

- Pure CPU math; only active while rotating (residual yaw > deadband) → cardinal stays byte-identical.
- No shader/parity surface touched.
- Builds clean; **12/12** per-axis math tests pass (6 new `perAxisSubdivisionCap` cases).

## Why it's design-blocked
Computing Option 2's effect on the standard demo (`cardinalExtent ≈ 640×720`):

| shot | effSub | original cap | new cap | subPerAxis |
|---|---|---|---|---|
| `zoom4_yaw45` | 4 | 8 (no bite) | **2** | **4 → 2 (halved)** |

Option 2 **regresses the common case** (zoom4 rotating, where there is currently no
clip and full detail) because a readback-free depth bound is scene-blind — it must
assume worst-case depth and over-caps shallow/common scenes. This contradicts the
issue's premise ("not a regression," "the residual only appears in scenes with a
large visible depth range"). See the `## NEEDS-DESIGN` comment for the alternatives
and recommendation.

Closes #1469 (pending the design decision).

### DIFF
--- a/engine/math/include/irreden/ir_math.hpp
+++ b/engine/math/include/irreden/ir_math.hpp
@@ -1553,25 +1553,36 @@ inline ivec2 perAxisTrixelCanvasWorstCaseSize(
 ///     subPerAxisCapped = min(effSub, floor( canvasHalf / maxOnScreenWorldDisp ))
 ///
 /// `maxOnScreenWorldDisp` is the largest in-plane world displacement of an
-/// on-screen voxel from the camera-tracked anchor: the visible cardinal-iso
-/// half-extent (`cardinalExtent / 2·zoom`) inverted to world on the depth-0
-/// plane (the `isoPixelToPos3D` closed form: `|wx|,|wy| ≤ isoHalf.x/2 +
-/// isoHalf.y/6`, `|wz| ≤ isoHalf.y/3`), grown by the √2 residual-yaw rotation
-/// footprint the canvas was sized for (@ref kSqrt2). The cap MUST be computed
-/// CPU-side: `effSub = clamp(m_vrs × round(zoom), 1, 16)` conflates subdivision
-/// and zoom, so a shader given only `effSub` + `canvasSize` cannot recover the
-/// zoom-dependent on-screen extent — the host knows `zoom` and feeds the capped
-/// value through `voxelRenderOptions.y` so the store, the framebuffer scatter,
-/// and the per-axis AO/lighting recovery all share one consistent world↔cell
-/// scale. Pass the same @p minOnScreenTrixelPx used to allocate the canvas
+/// on-screen voxel from the camera-tracked anchor. It has two terms:
+///   • the visible cardinal-iso half-extent (`cardinalExtent / 2·zoom`)
+///     inverted to world on the depth-0 plane (the `isoPixelToPos3D` closed
+///     form: `|wx|,|wy| ≤ isoHalf.x/2 + isoHalf.y/6`, `|wz| ≤ isoHalf.y/3`);
+///   • the depth-range term (#1469): a visible voxel at iso depth d = x+y+z
+///     sits at `world0(iso) + (d/3)·(1,1,1)`, which adds `d/3` to BOTH canvas
+///     axes of every face. d is scene-dependent and unbounded, so it is bounded
+///     here by the canvas-storable depth at the uncapped @p effSub —
+///     `canvasHalf.y / (√2·effSub)` — rather than read back from the depth
+///     texture; deeper cells clip to background harmlessly. See the body.
+/// Both terms are grown by the √2 residual-yaw rotation footprint the canvas was
+/// sized for (@ref kSqrt2). The cap MUST be computed CPU-side: `effSub =
+/// clamp(m_vrs × round(zoom), 1, 16)` conflates subdivision and zoom, so a
+/// shader given only `effSub` + `canvasSize` cannot recover the zoom-dependent
+/// on-screen extent — the host knows `zoom` and feeds the capped value through
+/// `voxelRenderOptions.y` so the store, the framebuffer scatter, and the
+/// per-axis AO/lighting recovery all share one consistent world↔cell scale.
+/// Pass the uncapped @p effSub (the requested density before this cap) and the
+/// same @p minOnScreenTrixelPx used to allocate the canvas
 /// (e.g. `kMinOnScreenTrixelSizePx`) so the cap is sized against the actual
 /// allocated canvas rather than the default. Returns ≥ 1.
 inline int perAxisSubdivisionCap(
-    const ivec2 cardinalExtent, const vec2 zoom, const float minOnScreenTrixelPx = 1.0f
+    const ivec2 cardinalExtent, const vec2 zoom, const int effSub,
+    const float minOnScreenTrixelPx = 1.0f
 ) {
     const float W = static_cast<float>(cardinalExtent.x);
     const float H = static_cast<float>(cardinalExtent.y);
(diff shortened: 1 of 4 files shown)


jakildev:
## NEEDS-DESIGN

Implementing #1469's recommended **Option 2** and computing its actual effect surfaced a contradiction with the issue's premise that needs an architecture call before this ships.

### What the issue assumed vs. what's true
The issue frames the residual as narrow — *"not a regression,"* *"the residual only appears in scenes with a large visible depth range combined with high effSub."* The implicit assumption is that the **fix** is as narrow as the **problem**. It isn't:

- The depth-range term is real and correctly derived: a visible voxel at iso depth `d=x+y+z` sits at `world0(iso)+(d/3)·(1,1,1)`, adding `d/3` to **both** canvas axes (proven by the `ClaimedDepthBandFitsCanvas` test).
- But `d` is **scene-dependent and unbounded**, and the cap is CPU-side with **no depth read**. So any readback-free bound (Option 2) must assume **worst-case depth** → it is **scene-blind** → it over-caps shallow/common scenes.

Concretely, on the standard `shape_debug` scene (`cardinalExtent ≈ 640×720`, default `FULL` mode, `effSub = round(zoom)`):

| shot | effSub | original cap | Option-2 cap | subPerAxis |
|---|---|---|---|---|
| `zoom4_yaw45_inter_cardinal` | 4 | 8 (no bite, full detail) | **2** | **4 → 2 (halved)** |

So Option 2 **introduces a common-case rotating-detail regression at zoom 4**, where there is currently no clip — the opposite of the issue's *"not a regression"* framing — and would invalidate the committed `zoom4_yaw45*` render-verify baselines. The detail loss the issue is willing to accept is explicitly tracked as a *separate* end-state in **#1458**; #1469 was meant to *only* stop the clip, not regress detail.

### The fork
- **Option 2 (readback-free):** cheap, but scene-blind → regresses the common case (above) + baseline churn, for a latent **no-corruption** residual (`isInsideCanvas` already drops OOB cells harmlessly).
- **Option 1 (depth readback):** accurate (only tightens for genuinely deep scenes), but needs a depth-texture readback in the CPU cap path — a persistent-mapped readback + one frame of latency + both backends. Disproportionate infra for a latent residual. (The issue's *"bounded from cardinal extent + zoom"* alternative doesn't actually work — depth is independent of the iso rect, so the viewport can't bound it.)

### Recommendation
**Fold #1469 into #1458 (or close as subsumed).** #1458's planned rework — store the per-axis lattice at **base/world resolution** and carry sub-detail as a **fractional micro-offset** instead of expanding the lattice by `effSub` — removes the `effSub`-driven overflow **at its root, with no cap**: the depth term then contributes `d/3` in *base* units (not `× effSub`), which the base-resolution canvas already holds, so deep scenes stop clipping **without** regressing detail. #1469's clip is therefore largely a symptom of the `× effSub` lattice that #1458 is already slated to remove.

1. Defer/fold #1469 into #1458 (my recommendation), or land an interim guard now?
2. If an interim guard is wanted: accept Option 2's documented common-case detail trade (this PR, behind the flag-less cap), or invest in Option 1's depth readback?
3. If Option 2 is accepted, the `zoom4_yaw45*` render-verify baselines need refreshing — in this PR or a follow-up?

The Option-2 implementation is committed here (tested, builds clean) as a concrete starting point so whichever way it's decided, the resume is one step. — opus-worker-1

jakildev:
## Architect decision — design-blocked → resolved (close)

**Direction: targeted fix, no epic** (confirmed with the human).

Your analysis stands and decided this: a readback-free cap is **scene-blind** and **regresses the common case** (zoom4 detail halved, `zoom4_yaw45*` baselines churned) for a latent no-corruption residual. We will **not** ship Option 2, and Option 1's depth readback is disproportionate infra. The effect table is what made the call obvious — thank you.

**#1469 is subsumed by #1458 and is being closed.** #1458's rework stores the per-axis lattice at **base/world resolution** (dropping the `× effSub` expansion) and carries a fractional winner offset. At base resolution the depth term contributes `d/3` in **base units**, which the bounded canvas already holds → deep scenes stop clipping **with no cap term and no readback**, and the `zoom4_yaw45*` baselines stay valid. So #1469's clip is removed at its root by #1458 rather than patched with a scene-blind bound.

Closing this PR; discard the diff. The deep-scene "must not clip without a cap" requirement is now tracked under #1458's DoD.

Issue #1596, opened 2026-06-08
render: detached re-voxelize world sun-shadow CAST (P4b-3) — second bake dispatch


jakildev:
**Area:** engine/prefabs/irreden/render
**Model:** opus
**Blocked by:** #1576

Final phase (P4b-3, **cast**) of the detached re-voxelize world-light stack (epic #1553, architect decomposition in `docs/design/detached-revoxelize-world-light.md`). Stack order: P4b-1 (depth, PR #1592 — landed/in-review) → P4b-2 (receive, #1576) → **P4b-3 (cast, this issue)**.

Make an **opt-in world-placed** detached re-voxelize solid (`C_EntityCanvas::worldPlaced_`, added in P4b-1) **cast** a world sun-shadow, so it drops a shadow onto the floor and onto other world geometry — completing the world-light participation begun by depth (P4b-1) and receive (P4b-2).

**Mechanism (architect Q2 → (a), decided):** a SECOND bake dispatch per opt-in detached canvas into the SHARED sun-shadow map, with a world-iso-depth offset + entity-world-translation recovery, mirroring the existing per-axis resolve bake (`system_bake_sun_shadow_map.hpp:173-186`). Do NOT merge detached cells into the main world pool (rejected — dissolves the private-pool architecture, epic decision-1). GL + Metal.

Gate everything on `worldPlaced_` so the default screen-locked overlay stays byte-identical. Until this lands, opt-in detached solids depth-sort (P4b-1) and receive (P4b-2) but cast no shadow — a documented intermediate state, not a regression.

**Acceptance criteria**
- A `worldPlaced_` detached re-voxelize solid casts a sun-shadow onto the SDF/GRID floor — verify against the #1587 shadow-floor scene (the architect's named end-to-end target).
- GL + Metal parity; render-verify references for both.
- Default (non-`worldPlaced_`) detached references byte-identical (the regression guard that the opt-in gate is wired correctly).

**Context:** P4b-1 (depth, PR #1592) added the `worldPlaced_` opt-in + world-depth composite + a CPU↔GPU GRID-equivalence gtest. Cast is the only piece that does NOT fall out of world-placement (the bake early-returns on `useCameraPositionIso_`, iterating only the main world canvas), so it needs this deliberate second-dispatch mechanism. Full analysis + the four resolved design decisions are in `docs/design/detached-revoxelize-world-light.md`.

PR #1592, opened 2026-06-08
render: detached re-voxelize world-depth composite — opt-in flag (P4b-1) (#1576)

jakildev:
**P4b-1 — the keystone** of the detached re-voxelize world-light stack
(epic #1553; architect decomposition resolved on this PR, full design in
`docs/design/detached-revoxelize-world-light.md`).

Stack order: #1583 (screen-locked default, merged) → **P4b-1 (this PR, depth)**
→ P4b-2 (receive, #1576) → P4b-3 (cast, #1596).

Part of #1576 — this PR delivers **depth composite** only and does NOT close
#1576, which continues as the **receive** phase (P4b-2) once this lands.

## What this does

Adds a per-entity **opt-in** that world-places a detached re-voxelize solid so
it depth-sorts against world geometry. The default screen-locked overlay
(#1582 Option B / #1583) stays **byte-identical**.

- **`C_EntityCanvas::worldPlaced_`** (default OFF). Lives on `C_EntityCanvas`
  — which `ENTITY_CANVAS_TO_FRAMEBUFFER` already iterates — rather than
  `C_TrixelCanvasRenderBehavior` (on the child canvas entity), so the composite
  reads it with **no per-tick foreign getComponent**. Bound for Lua.
- **Composite depth:** when `worldPlaced_`, set
  `distanceOffset_ = pos3DtoDistance(roundVec3HalfUp(translation))` (the entity's
  world iso depth on the GRID `trixelDistances` convention); else 0. The detached
  canvas rasters its pool in the **pool-centered model frame** and re-voxelize
  keeps `voxelDepthAxis = (1,1,1)`, so its model `rawDist` is the exact world iso
  depth of the rotated cell; adding the constant entity offset reproduces the GRID
  world depth (`pos3DtoDistance` is linear).
- **CPU↔GPU GRID-equivalence gtest** (`test/render/detached_world_depth_test.cpp`,
  the architect's Q4 deliverable): world-placed composite depth **==** GRID depth,
  **bit-exact** at cardinal poses + integer translations; within **±1 rawDist**
  (a documented sub-cell half-integer wobble, 1/131070 of the depth range) at
  arbitrary poses.
- **`canvas_stress --world-place-revox`** debug flag (default OFF → the
  auto-screenshot references stay byte-identical) opts the two center re-voxelize
  solids in — the runnable demonstration.

## Verification

- New gtest: 5/5 pass. Full suite: 1093/1093 (no regressions).
[...]

Issue #1587, opened 2026-06-08
demo(canvas_stress): shadow floor — cast shadows don't land on the SDF floor (WIP parked)


jakildev:
WIP from this session, parked. A shadow floor was added to `canvas_stress` (branch `claude/canvas-stress-floor`) as a depth/shadow reference. It renders correctly as a **sun-lit SDF box** (`SHAPES_TO_TRIXEL` + `LOD_UPDATE` + a flat `C_ShapeDescriptor{BOX}` below the scene), but **cast shadows never land on it**: the floor is sun-*lit* (Lambert) yet not sun-*shadowed* even at near-zero ambient. The GRID casters (spin cluster + GRID orbit shapes) should cast onto it (they're world voxels); detached entities can't (that's #1582 — fixed-depth composite).

## To investigate
- Why the SDF floor doesn't receive screen-space sun-shadow in this multi-canvas scene (does `BAKE_SUN_SHADOW_MAP` include the GRID casters at the floor's pixels? does `COMPUTE_SUN_SHADOW` shade SHAPES_TO_TRIXEL pixels here?). Cross-check vs `shape_debug`, whose SDF floor does receive shadows.

## Constraints learned (for whoever picks this up)
- `setSunDirection` points world→sun and **asserts z ≤ 0** (+Z is down, sun is up). A floor "below" the entities is at **+z**; shadows fall +z onto it.
- Floor span 560 (covers the radius-200 orbit), `kFloorZ ≈ 16`.

Likely best redone fresh on top of #1581 + #1582. Blocked-ish on #1582 for the detached half.

jakildev:
**Architect decision: GRID-only focused shadow stage** (dedicated floor plate near the center GRID cluster, close caster→floor gap à la `shape_debug`; not the covers-the-orbit far placement). Detached casters don't drop world shadows by design (→ #1582 / #1576). Stacks on #1585 (the 12³ GRID casters). Unblocked — see PR #1591.

jakildev:
[...]

Issue #1558, opened 2026-06-06
render: detached re-voxelize — AO / sun / light integration (P4)


jakildev:
**Model:** opus
**Part of epic:** #1553
**Plan file:** `~/.fleet/plans/issue-1553.md` (full epic plan)
**Blocked by:** #1557

## Scope
Light the re-voxelized detached canvas consistently with the rest of the engine (AO + sun-shadow + light-volume + Lambert), at parity with how an attached GRID solid or the per-axis path it replaces is lit.

## Approach
- Drive the existing per-canvas lighting passes over the re-voxelized detached pool (the canvas renders cardinal cells, so the standard lighting path applies — the rotation is in the cells, the normals are cardinal).
- Confirm the detached lighting contract (cast/receive) from the per-axis path (`per_axis_sun_shadow_resolve`, the P4/#1465 detached lighting) is preserved or correctly superseded.

## Acceptance criteria
- Re-voxelized detached solid is lit equivalently to an attached GRID solid.
- Shadows cast/receive per the existing detached lighting contract.
- Identity / static byte-identical.

## Dependencies
Blocked by #1557 (geometry must be correct before lighting it). Blocks P5.

## Gotchas
- The re-voxelized canvas uses cardinal normals (rotation is in cell positions), so the standard lighting path applies directly — do NOT port the per-axis residual-aware lighting; that machinery is being retired for detached in P6.

## References
- `~/.fleet/plans/issue-1553.md` (epic plan)
- `docs/design/per-axis-sun-shadow-resolve.md`, the per-axis detached lighting (#1465)

PR #1617, opened 2026-06-09
render: detached re-voxelize world-receive — sun-shadow + light-volume (P4b-2) (#1576)

### LINKED ISSUE #1576: render: detached re-voxelize — world sun-shadow + light-volume for detached solids (P4b)
**Model:** opus
**Part of epic:** #1553
**Blocked by:** #1558

## Scope
Give detached re-voxelize solids **world sun-shadow + light-volume** participation (cast/receive), deferred from P4 (#1558). P4's Option A ships the self-contained terms only (AO + directional sun face-shading + HDR sky), because sun-shadow and light-volume are single-shared-world-canvas subsystems keyed to the main canvas frame — baking a screen-locked, pool-local detached canvas into them cross-contaminates the world map.

## Approach (architect to refine when unblocked)
Decide between:
- **Per-canvas private infra** — a private sun-shadow depth map baked from the detached pool's own frame + a private (or skipped) light volume keyed to the detached frame; OR
- **Formally scope detached solids out of world cast/receive** — document that detached re-voxelize solids are lit by AO + directional sun only (no world shadow cast/receive, no world light bleed), the same stance already parked for #1375.

## Acceptance criteria
- A recorded decision (private infra vs. scoped-out) in `docs/design/` + the epic plan.
- If private infra: a detached re-voxelize solid casts/receives world sun-shadow + light-volume equivalently to an attached GRID solid, without corrupting the world map or other detached canvases.

## Dependencies
Blocked by #1558 (P4 ships the self-contained terms first). Same underlying problem as #1375 ("light detached-canvas entities").

## References
PR #1569 architect decisions (Option A scope + the Option 1 per-pass author/restore plumbing); #1375; epic #1553.

jakildev:
## Summary

P4b-2 of the detached re-voxelize world-light epic (#1553). Gives **opt-in
world-placed** detached re-voxelize solids world-light **RECEIVE**: they sample
the SHARED world sun-shadow map + 128³ light volume at their recovered world
position, so they darken in world shadow and pick up light-volume bleed like an
attached GRID solid. Gated on the existing `C_EntityCanvas::worldPlaced_` opt-in
(default off) — the screen-locked overlay default path stays byte-identical.

Stacks conceptually on P4b-1 (depth composite, #1592, merged). P4b-3 (cast) is
the remaining phase (#1596).

### How it works
- **Plumbing:** `PROPAGATE_CANVAS_ROTATION` propagates `worldPlaced_` + the entity
  world cell origin (`roundVec3HalfUp(translation)` — the same rounding P4b-1's
  composite depth offset uses) onto `C_CanvasLocalRotation`. `buildVoxelFrameData`
  publishes them into the shared voxel-frame UBO (`FrameDataVoxelToCanvas`) as a
  new appended `detachedWorldReceive_` (`.xyz` = world cell offset, `.w` = enable).
- **Shader (GL + Metal):** `c_lighting_to_trixel` recovers each detached voxel's
  WORLD pos as `trixelCanvasPixelToWorld3D(...) + .xyz` (the detached canvas
  rasters its pool-centered MODEL frame; the camera offset cancels in the
  recovery, so `model pos + entity world origin = world cell`). When `.w != 0` it
  samples the shared 128³ light volume there AND re-runs the sun-shadow cascade
  lookup at that world pos.
- **Reuse:** the cascade sampler + `FrameDataSun` layout + sun-depth SSBO were
  **extracted** from `c_compute_sun_shadow` into a new shared
  `ir_sun_shadow_sample.{glsl,metal}` include, so both passes share one source
  (no duplication). The lighting system binds the sun-depth SSBO at slot 28.

This is the architecturally faithful "receive" (Q3a) — a world-placed solid is
shadowed *by world geometry* (floor, other solids) exactly as a GRID voxel is,
rather than carrying a per-canvas shadow texture.

## Verification

**Default path byte-identical (the regression guard).** All **28 `shape_debug`
auto-screenshot shots are byte-identical before vs after** — including the yaw
sun-shadow shots (`zoom4_yaw90` etc.) that exercise the refactored
`c_compute_sun_shadow`. The extraction + lighting restructure are pure code-moves.
(The detached/re-voxelize demos have a known CPU↔GPU sub-cell speckle that master
exhibits run-to-run — the `--world-place-revox` path's small drift is within that
same envelope, not a regression.)

**Receive correctness confirmed via a forced-shadow-factor isolation pass:** with
the cascade result forced, a world-placed solid darkens on the opt-in path,
confirming the flag + plumbing + shadow multiply all reach the shader; with the
real lookup, the solid samples the world map at its world pos.

**Note on a camera-visible shadow demo:** `canvas_stress`'s sun direction ≈ the
iso camera direction, so world cast-shadows fall *away* from the camera (behind
occluders) — making a dramatic "solid darkens in shadow" screenshot awkward to
frame in this scene. The end-to-end visual lands naturally with **P4b-3 (cast) on
the #1587 shadow floor** (the architect's designated visual target). The world
placement itself is visible (solids depth-sort onto the floor — see the orbit
screenshot).

Built + ran clean on **macOS/Metal** (this host). GLSL validated by review (Metal
mirror compiled + ran); Linux/OpenGL side covered by cross-host smoke.

## Screenshots

`shape_debug` regression guard — before/after are **byte-identical** (the shader
refactor changed zero output):

<details>
<summary>zoom8_origin (lighting + shadow)</summary>

| Before (origin/master) | After (this branch) |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_yaw90 (yaw sun-shadow path)</summary>

| Before (origin/master) | After (this branch) |
|--------|-------|
|  |  |

</details>

<details>
<summary>canvas_stress --world-place-revox (the two center re-voxelize solids depth-sort onto the floor)</summary>

</details>

## Test plan

- [x] `shape_debug` 28-shot suite byte-identical before/after (default-path regression guard)
- [x] `canvas_stress --world-place-revox` builds + runs clean on macOS/Metal
- [x] Receive path reaches the shader (forced-factor isolation)
- [x] `FrameDataVoxelToCanvas` std140 asserts updated (offset 160, size 176); GLSL/Metal UBO layouts mirror it
- [ ] Linux/OpenGL cross-host smoke (GLSL parity)

Closes #1576


jakildev (commented):
## Review — render: detached re-voxelize world-receive — sun-shadow + light-volume (P4b-2) (#1617)

**Verdict:** approve

### Summary

P4b-2 of the detached re-voxelize world-light stack. World-placed detached re-voxelize solids now receive the shared world sun-shadow map + 128³ light volume at their recovered world position. The default screen-locked overlay path is byte-identical (verified: 28-shot `shape_debug` regression suite before/after are byte-identical). The refactor is clean: the cascade PCF sampler is extracted into `ir_sun_shadow_sample.{glsl,metal}` and shared between `c_compute_sun_shadow` and `c_lighting_to_trixel`, eliminating the duplication that existed before.

### Code quality assessment

**Struct layout / binding correctness:** `FrameDataVoxelToCanvas::detachedWorldReceive_` is std140-appended at offset 160 (vec4), confirmed by two new `static_assert`s in `ir_render_types.hpp:494–503`. The GLSL `FrameDataVoxelToTrixel` block extends to include `_voxelDepthAxisPadding` (the prior-last field, now aliased rather than used) and `detachedWorldReceive` at the same offset — correct prefix consumption. `ir_iso_common.metal`'s `FrameDataVoxelToTrixel` struct gains both fields in the same order. GL ↔ Metal layout parity confirmed. ✓

**Cascade depth computation:** The iso depth passed to `worldSunShadowFactor` for a detached voxel is `float(rawDepth) + offset.x + offset.y + offset.z`. Since `rawDepth` is the model iso depth (x+y+z in the pool-centered frame) and `offset` = `roundVec3HalfUp(translation)`, the sum equals the world iso depth of the actual voxel — the correct value for cascade selection. ✓

**ECS pattern:** `PROPAGATE_CANVAS_ROTATION::tick()` now takes `const C_WorldTransform&` as an additional parameter. This is the correct pattern for reading a component from the same entity without a foreign `getComponent` call. No ECS smells. ✓

**`cardinalIndex` removal:** The old code declared `int cardinalIndex = rasterYawCardinalIndex(rasterYaw)` immediately after the `faceId` / `worldNormal` lines but never used it in either GLSL or Metal. The hoisted rawDepth/slot/faceId/worldNormal block in this PR eliminates it. Grep confirms: `cardinalIndex` does not appear in either shader file post-change. ✓

**Frame-leak prevention:** `buildVoxelFrameData` resets `detachedWorldReceive_` to `(0,0,0,0)` unconditionally at the top of the frame, matching the existing `voxelDepthAxis_` reset pattern. A non-opt-in canvas or world canvas can never see a stale value from the prior detached canvas dispatch. ✓

**Shared include reuse:** `ir_sun_shadow_sample.{glsl,metal}` introduces a purposeful shared include. `kBufferIndex_SunShadowDepthMap` is an existing constant aliased from `kBufferIndex_LightOcclusionGrid` (= 28), confirmed in `ir_render_types.hpp:602`. No new constant introduced. ✓

### Nits

- `engine/render/src/shaders/c_lighting_to_trixel.glsl:226` — Comment says "At cardinalIndex==0 the path collapses to master..." but `cardinalIndex` no longer exists in this shader. Update the comment to drop the `cardinalIndex==0` reference (the content it is explaining — that `rasterYaw == 0` keeps the cardinal fast path byte-identical — is still correct, just name the actual predicate).

- `engine/render/src/shaders/c_lighting_to_trixel.glsl:62` — `_voxelDepthAxisPadding` is a slightly confusing name for a real field (`voxelDepthAxis_`) that is simply unused by this shader. `_voxelDepthAxisUnused` or `_voxelDepthAxis` would better signal "real field, not consumed here."

- [x] `shape_debug` 28-shot suite byte-identical before/after (macOS/Metal, confirmed in PR body)
- [x] `canvas_stress --world-place-revox` builds + runs on macOS/Metal
- [x] `FrameDataVoxelToCanvas` std140 offset/size static asserts pass
- [ ] Linux/OpenGL cross-host build smoke (`fleet-build --target IRShapeDebug` + `IRShapeDebug --auto-screenshot`) — explicitly pending in the PR's own test plan

Escalation: please Opus-recheck before merge (touches: `engine/render/`, `engine/prefabs/irreden/render/`, GPU buffer binding).

**Opus focus:** `system_lighting_to_trixel.hpp:359` binds slot 28 (`SunShadowDepthMap`) only when `sunShadowDepthMap_ != nullptr`. The Metal kernel unconditionally declares `sunDepthBuf [[buffer(28)]]`. If a scene adds a world-placed detached entity (`C_EntityCanvas::worldPlaced_ = true`) to a pipeline that does NOT register `BAKE_SUN_SHADOW_MAP`, `detachedWorldReceive.w` would be `1.0` in the shader and `worldReceive` would be `true`, causing the kernel to access an unbound/nil Metal buffer argument — undefined behavior or crash. The author's invariant ("no world-placed detached solids exist without sun shadows") is asserted in comments but not enforced in code. Opus should verify whether this coupling is robust or whether a defensive fallback is warranted (e.g. force `detachedWorldReceive.w = 0.0` in `buildVoxelFrameData` when the shadow map is absent, or add an assert on entry to the world-placed path).

Opus recheck required: GPU buffer lifetime invariant (`sunShadowDepthMap_` null-guard, Metal kernel arg binding when `worldReceive = true` in no-shadow scenes).

jakildev:
Addressed nits:
- Updated stale `cardinalIndex==0` comment at line 226 → `rasterYaw==0` (the actual predicate in use after the removal of `cardinalIndex`)
- Renamed `_voxelDepthAxisPadding` → `_voxelDepthAxisUnused` at line 62 to better signal the field is real but not consumed by this shader

No runtime behavior change. `fleet:approved` retained.

jakildev (commented):
## Opus recheck — detached re-voxelize world-receive (P4b-2) (#1617)

**Verdict:** approve (nits)

Sonnet's first pass is thorough and correct on everything it checked (std140 offset/size static-asserts, GL↔Metal struct parity, the `cardinalIndex` removal, the frame-leak reset, the shared-include extraction, the cascade-depth math, the new `C_WorldTransform` ECS param). I focused this pass on the two things Sonnet flagged as Opus-only — the GPU buffer-binding invariant — plus one archetype-narrowing risk Sonnet did not call out. Both come back **safe**, and I found one additional doc-accuracy nit in the lifetime story.

### 1. GPU buffer-binding invariant (Sonnet's `Opus recheck required`) — confirmed safe, but latent-not-impossible

Sonnet flagged: the Metal kernel now declares `sunDepthBuf [[buffer(28)]]` unconditionally, while the C++ binds slot 28 only when `sunShadowDepthMap_ != nullptr`; a world-placed detached solid in a pipeline without `BAKE_SUN_SHADOW_MAP` would read an unbound buffer → UB/crash.

On closer read I **confirm the binding is safe for every pipeline that exists today**, for a stronger reason than Sonnet gave. `BAKE_SUN_SHADOW_MAP` (which creates the `"SunShadowDepthMap"` resource — `system_bake_sun_shadow_map.hpp:317`) and `LIGHTING_TO_TRIXEL` are co-registered under the **same** `if (!noLighting_)` guard in all 13 registered pipelines (canvas_stress, shape_debug, voxel_editor, the lighting demo scene, the game projects, etc.). So whenever `LIGHTING_TO_TRIXEL` runs, the named resource exists → the lazy resolve returns non-null → slot 28 is bound every tick. The unconditional Metal declaration is therefore always satisfied, even on the **default** (non-world-placed) path where the kernel arg is no longer dead code. Sonnet's specific crash scenario additionally requires a world-placed detached solid (`--world-place-revox`), and that flag runs the full lighting pipeline including the bake — so the two preconditions can't co-occur.

The coupling is held by **registration convention, not enforced in code** — the relaxed multi-lit-canvas archetype (#1558) made "lighting without per-canvas sun-shadow" structurally expressible, so a future pipeline *could* register `LIGHTING_TO_TRIXEL` without the bake. I'm comfortable approving because (a) no such pipeline exists, (b) `COMPUTE_SUN_SHADOW` already uses the identical lazy-resolve-of-`SunShadowDepthMap` pattern and carries the same implicit coupling, so this PR adds no new class of fragility, and (c) a misconfigured no-bake pipeline fails loud (see nit 3), not silently. Sonnet's suggested hardening (force `detachedWorldReceive.w = 0` when the map is absent) would require plumbing the resolved buffer pointer into the free-function `buildVoxelFrameData`, and isn't worth it against an invariant that holds everywhere — noting it for the architect rather than requesting it.

### 2. `PROPAGATE_CANVAS_ROTATION` gains `C_WorldTransform` as a required param — confirmed safe (Sonnet did not flag)

Adding a required component **narrows** the archetype query, so any owner entity matching `{C_LocalTransform, C_RotationMode, C_EntityCanvas}` but lacking `C_WorldTransform` would silently drop out of the system. I verified it cannot: `createEntity` language-enforces `C_LocalTransform → C_WorldTransform` co-injection (`engine/entity/include/irreden/ir_entity.hpp:151-168`; documented at `engine/entity/CLAUDE.md:278`), and all canvas_stress owner-entity spawn paths pass `C_LocalTransform`. So the iterated set is unchanged — no regression. This is the correct fix for reading the owner's world transform without a per-entity foreign `getComponent`. ✓

### Nits (non-blocking)

3. **`system_lighting_to_trixel.hpp:374-378` — the lazy-resolve comment misdescribes the failure mode.** The comment says "*Lazy so this links even if a scene never registers the bake — then the map is null and the bind is skipped.*" But `getNamedResource` → `getNamed` **`IR_ASSERT`s** on a missing key (`rendering_rm.hpp:87`); it never returns null. In a hypothetical bake-less lighting pipeline this line would assert at the resolve (debug) or dereference an end-iterator (release) — the `sunShadowDepthMap_ != nullptr` guard at the bind site is never reached via the "scene never registered the bake" path. The guard is fine as defensive code, but the rationale is unreachable as written. Please reword to reflect that a no-bake pipeline is an *assert*, not a graceful null-skip (or, if graceful degradation is actually wanted, gate the resolve on a resource-exists check and keep it consistent with `COMPUTE_SUN_SHADOW`'s identical pattern).

4. (Endorsing Sonnet's two nits) `c_lighting_to_trixel.glsl:226` comment still references `cardinalIndex==0` which no longer exists in the shader; and `_voxelDepthAxisPadding` at `:62` would read more clearly as `_voxelDepthAxisUnused` ("real field, not consumed here"). Both are comment/naming-only.

### Cross-host
Sonnet's linux/windows smoke labels are still pending — the Metal change is the higher-risk side and was authored+verified on macOS; the GL byte-identity on the default path is the regression guard for the OpenGL hosts. Leaving the smoke labels for the platform-catchup pass.

Nits are comment/naming only and don't gate merge. Approving.

jakildev:
Addressed nit: fixed inaccurate lazy-resolve comment in `system_lighting_to_trixel.hpp:277-283`. The comment claimed `getNamedResource` returns null for a missing resource (allowing graceful skip), but it actually `IR_ASSERT`s. Rewrote to accurately describe the single-init pattern and note that a no-bake pipeline is a config error, not a handled fallback. Shader nits (cardinalIndex→rasterYaw comment, \_voxelDepthAxisPadding→\_voxelDepthAxisUnused) were already addressed by the author before the Opus review.

PR #1627, opened 2026-06-09
lua: CODEGEN DSL — whitelisted side-effecting engine bindings (#1616)

### LINKED ISSUE #1616: lua: extend the Lua-ECS CODEGEN DSL — whitelisted side-effecting engine bindings + basic string formatting in tick bodies
## Area
`engine/script/` — the Lua-ECS CODEGEN DSL (`system_dsl` + the codegen tool that lowers `IRSystem.registerSystem` tick bodies to C++).

## Suggested model
`[opus]` — codegen DSL / lowering work with correctness + determinism implications.

## Context
The Lua-first authoring path (engine #293) has two modes: **CODEGEN** (build-time Lua→C++, perf-parity, the shipping target) and **EVAL** (LuaJIT runtime, hot-reload, dev). Today the CODEGEN DSL is a restricted subset (`engine/script/CLAUDE.md`): no string ops, no side-effecting bare-expression binding calls, no cross-entity reads, no closures, no table constructors beyond `Comp.new`.

This forces any system that **applies a side-effect to an engine binding** (e.g. pushing a computed value to `IRRender.setSunDirection`) or **formats a string** (e.g. a HUD label) to stay in EVAL or be reimplemented as a C++ bridge — even when the surrounding gameplay math is pure and compiles fine. So "as much as possible compiled" can't reach the render/GUI-driving systems: they're stranded in EVAL/C++.

The compute-in-CODEGEN + thin-bridge pattern (compute into component fields under CODEGEN, apply side-effects in a separate system) is the current workaround, but it pushes the *application* layer out of compiled Lua.

## Proposed (either/both, scoped as the DSL maintainers prefer)
1. **Whitelisted side-effecting engine bindings in CODEGEN tick bodies** — allow a curated set of pure pass-through engine calls (e.g. the `IRRender.*` render-glue setters) to lower to their C++ equivalents inside a CODEGEN tick, instead of being rejected as side-effecting bare calls.
2. **Basic string formatting** — support a minimal `string.format`/concat subset lowered to C++ (`std::format`/`snprintf`) so HUD/label systems compile.

## Acceptance
- A CODEGEN system tick can call a whitelisted side-effecting engine binding and lower cleanly to C++.
- (If pursued) a CODEGEN tick can build a simple formatted string and pass it to a whitelisted text binding.
- Existing DSL restrictions remain for everything outside the whitelist; CODEGEN/EVAL parity + determinism guarantees preserved.
- Documented in `engine/script/CLAUDE.md`.

## Notes
- The IREnum-in-CODEGEN-tick literal-folding limitation (`engine/script/CLAUDE.md`) is a sibling DSL-expressiveness gap and may share machinery.
- Motivated by a downstream Lua-first visual system that wants its render/GUI application layer compiled rather than EVAL-only.

jakildev:
## Summary

Extends the build-time Lua→C++ CODEGEN system DSL (`cmake/lua_codegen/`) so a `tick` body can call a **whitelisted side-effecting engine binding** as a bare statement, lowered to the C++ free function. Previously every bare-call statement was rejected, which stranded the *application* layer of an otherwise-pure CODEGEN system in EVAL or a hand-written C++ bridge.

- New `Intrinsic::isStatement_` flag marks void, side-effecting pass-through bindings (statement-position only). The seed set is the `IRRender.*` render-glue setters: `setSunIntensity`, `setSunAmbient`, `setExposure`, `setSkyIntensity`, `setCameraZoom`.
- Parser turns a bare `INTRINSIC_CALL` into an `EXPR_STMT`; emission validates statement-vs-expression position: a value-returning intrinsic (`math.sin`) as a bare statement is still rejected (dropped-assignment guard), and a void binding inside an expression is rejected.
- New `Intrinsic::requiredInclude_` carries the engine header that declares each binding's C++ function. The codegen tool collects the union across the run and emits those `#include`s in the generated header **only when a body uses the binding** (the system-emit pass moved ahead of the include block to collect them; emitted output order is otherwise unchanged — byte-identical for runs with no side-effecting calls).
- Adds the value-returning vs statement-position contract + the deferred string-formatting note to `engine/script/CLAUDE.md`.
- Shared `resolveIntrinsic` / `emitIntrinsicCall` Emitter helpers consolidate the lookup + arity + lowering that both positions use.

Scope note: this is feature 1 of #1616. Feature 2 (basic `string.format` / `..` lowering) is explicitly optional in the issue and materially larger (string-type tracking, `..` lexing, a text binding to consume the result); it is deferred and documented as EVAL-only for now.

## Test plan
- [x] `ir_lua_codegen` host tool builds clean.
- [x] `IrredenEngineTest` builds + links — the test referencing `createSystem_CodegenRenderGlue` is itself proof the bare-call lowered to valid C++ that resolves `IRRender::setSunIntensity` at link time.
- [x] New fixture `CodegenRenderGlue` + test `RenderGlueSideEffectLowersAndRegisters` (registers + asserts; does not execute the tick — the render setter asserts on an absent `RenderManager` headlessly).
- [x] All 88 `*Codegen*` / `*Coexistence*` / `Lua*` tests pass (no regression from the `main.cpp` emit-order refactor).
- [x] Generated header verified: conditional `#include <irreden/ir_render.hpp>`, `IRRender::setSunIntensity(pos.x_);` bare statement, `pos` still `const auto&` (#1353 alias optimization composes with the new statement path).
- [ ] linux-debug / windows-debug cross-host build (codegen tool is platform-agnostic; flagging for smoke).

Closes #1616

### DIFF
--- a/cmake/lua_codegen/main.cpp
+++ b/cmake/lua_codegen/main.cpp
@@ -37,6 +37,7 @@
 #include <fstream>
 #include <iostream>
 #include <limits>
+#include <set>
 #include <sstream>
 #include <stdexcept>
 #include <string>
@@ -676,6 +677,46 @@ void writeOutput(
         os << "//   " << s.sourceFile_ << " (system " << s.name_ << ")\n";
     }
     os << "#pragma once\n\n";
+
+    // #1616: pre-emit the CODEGEN system bodies into a buffer first, so we can
+    // collect the set of engine-binding headers their whitelisted
+    // side-effecting calls (e.g. `IRRender.setSunIntensity`) require and emit
+    // those #includes alongside the static core set below. The buffer is
+    // flushed into the IRScript::CodegenRegistry namespace further down, in its
+    // original position — only the parse/emit timing moves, not the output
+    // order.
+    std::string systemsBuf;
+    std::set<std::string> extraIncludes;
+    const bool hasCodegenSystems = std::any_of(
+        cap.systems_.begin(), cap.systems_.end(),
+        [](const IRLuaCodegen::SystemRecord &r) {
+            return r.mode_ == IRLuaCodegen::SystemMode::CODEGEN;
+        }
+    );
+    if (hasCodegenSystems) {
+        const auto componentRegistry = toComponentSchemas(cap.components_);
+        for (auto &rec : cap.systems_) {
+            if (rec.mode_ != IRLuaCodegen::SystemMode::CODEGEN) {
+                continue;
+            }
+            int bodyStartLine = 0;
+            try {
+                rec.bodySource_ = IRLuaCodegen::sliceFunctionBody(
+                    rec.sourceFile_, rec.linedefined_, rec.lastlinedefined_, bodyStartLine
+                );
+                rec.bodyStartLine_ = bodyStartLine;
+                IRLuaCodegen::ParsedBody body = IRLuaCodegen::parseSystemBody(
+                    rec.sourceFile_, rec.bodyStartLine_, rec.bodySource_
+                );
+                IRLuaCodegen::emitSystem(systemsBuf, rec, body, componentRegistry, extraIncludes);
+            } catch (const IRLuaCodegen::ParseError &err) {
+                std::cerr << "lua_codegen: error in system '" << rec.name_ << "' at "
+                          << err.file_ << ":" << err.line_ << ": " << err.message_ << "\n";
+                std::exit(1);
+            }
+        }
+    }
+
     os << "#include <cstdint>\n";
     os << "#include <string>\n";
     os << "#include <vector>\n\n";
@@ -684,7 +725,14 @@ void writeOutput(
     os << "#include <irreden/ir_system.hpp>\n";
     os << "#include <irreden/script/lua_binding_traits.hpp>\n";
     os << "#include <irreden/script/lua_script.hpp>\n";
-    os << "#include <irreden/system/ir_system_types.hpp>\n\n";
+    os << "#include <irreden/system/ir_system_types.hpp>\n";
+    // #1616: engine-binding headers required by whitelisted side-effecting
+    // intrinsics used in this run's CODEGEN tick bodies (std::set → sorted,
+    // deduped). Empty unless a body calls an IRRender.* (or future) binding.
+    for (const auto &inc : extraIncludes) {
+        os << "#include <" << inc << ">\n";
+    }
+    os << "\n";
 
     os << "namespace IRComponents {\n\n";
     for (const auto &c : cap.components_) {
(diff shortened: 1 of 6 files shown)


jakildev (commented):
## First-pass review — #1627 lua: CODEGEN DSL — whitelisted side-effecting engine bindings (#1616)

**Verdict: approve with nits.** Clean, well-scoped DSL extension. All invariants satisfied; one minor comment-wording nit.

---

### ECS invariants ✅

Not applicable — this PR touches `cmake/lua_codegen/` (build-time Lua→C++ transpiler), test fixtures, and a doc update. No runtime ECS systems, no `SystemName` enum entries, no per-entity `getComponent` in tick loops.

### Ownership / lifetime ✅

- `resolveIntrinsic()` returns `const Intrinsic &` into the static `kIntrinsicRegistry` vector — valid for the process lifetime. ✓
- `emitIntrinsicCall()` takes `const Expr &` and `const Intrinsic &` — both non-owning refs valid within the emit pass. ✓
- `Emitter::requiredIncludes_` is a `std::set<std::string>` member — standard lifetime, cleared with the emitter. ✓
- `outRequiredIncludes` is an out-param reference to a caller-scope `std::set` in `main.cpp`, appended (not cleared) across all system bodies — consistent with the "not cleared" note in the API doc. ✓

### Render pipeline ✅

No GPU structs, no shaders, no binding points, no `.glsl`/`.metal` files. The new `IRRender.*` entries are C++ free-function call targets in generated code; `<irreden/ir_render.hpp>` is the existing render surface.

### Logic correctness ✅

All four isStatement × position cases are handled:
- `math.sin(x)` in expression (`isStatement_=false`): passes `if (intr.isStatement_)` guard → emits. ✓
- `math.sin(x)` as bare statement (`isStatement_=false`): hits `!intr.isStatement_` → "assign the result to a local" error. ✓
- `IRRender.setSunIntensity(x)` in expression (`isStatement_=true`): hits `if (intr.isStatement_)` → "returns void, may only be used as a bare statement" error. ✓
- `IRRender.setSunIntensity(x)` as bare statement (`isStatement_=true`): passes `!intr.isStatement_` guard, inserts include, emits. ✓

The `EXPR_STMT` parser path (`system_dsl.cpp:570-577`) only fires when `e->kind_ == ExprKind::INTRINSIC_CALL`, so `s.exprStmt_->intrinsicNamespace_` and `intrinsicName_` are always populated in the emitter's `EXPR_STMT` case. ✓

The emit-order refactor in `main.cpp` (hoisting system parse+emit before the `#include` block) is mechanical — the `systemsBuf` string is flushed at its original output position. Byte-identical for runs with no side-effecting calls. ✓

Aggregate initialization of the existing `kIntrinsicRegistry` entries is unaffected: the new `isStatement_` and `requiredInclude_` fields carry `= false` / `= nullptr` defaults, so the 5 existing entries using 4-field initializers are valid C++11 onward. ✓

### Naming / style ✅

`isStatement_` and `requiredInclude_` follow the trailing `_` convention used by all four existing `Intrinsic` struct fields — consistent with the pre-existing pattern. No slip introduced by this PR.

### Tests / build ✅

- `RenderGlueSideEffectLowersAndRegisters`: the compile + link against real `IRRender::setSunIntensity` is the critical regression surface, and the test correctly explains why tick execution is omitted (headless harness asserts on absent `RenderManager`). ✓
- `RegistryReturnsAllCodegenSystemIds` extended with `EXPECT_NE(ids.CodegenForNumericBackEdge, ids.CodegenRenderGlue)` — consistent with the chain pattern. ✓
- PR body confirms all 88 `*Codegen*` / `*Coexistence*` / `Lua*` tests pass. ✓
- `engine/script/CLAUDE.md` updated with statement-binding semantics, string-format deferral note, and "Scalar args only" constraint. ✓

---

### Nits

**`cmake/lua_codegen/system_dsl.cpp:1108`** — Member comment for `requiredIncludes_` says "Drained by emitSystem into the generated header's #include block." The field is not drained (emptied) — its contents are *merged into* `outRequiredIncludes` via `insert(begin, end)`, and the member remains valid after. "Merged into outRequiredIncludes by emitSystem" would be more accurate.

---

### Test plan before merge

- [x] `ir_lua_codegen` host tool builds clean.
- [x] `IrredenEngineTest` links — `createSystem_CodegenRenderGlue` resolves `IRRender::setSunIntensity` at link time (compile assertion).
- [x] All 88 `*Codegen*` / `*Coexistence*` / `Lua*` tests pass.
- [ ] linux-debug + windows-debug cross-host build (codegen tool is platform-agnostic; flagging for smoke).

Opus recheck not required.

jakildev:
Addressed feedback: fixed `requiredIncludes_` member comment — changed "Drained by emitSystem" (inaccurate: the field is not emptied) to "Merged into outRequiredIncludes by emitSystem" (accurate: contents are inserted into the caller-scope out-param via `insert(begin, end)`, member remains valid after).

jakildev:
Confirmed the single review nit (requiredIncludes_ comment wording at system_dsl.cpp:1109-1111) is already resolved at HEAD (587fa730 — "merged, not drained"). Clearing the stale fleet:has-nits label; fleet:approved stands. — opus-worker

PR #1614, opened 2026-06-09
docs: re-plan entity-editor Phase 2 (#605) — file children #1602-#1612

jakildev:
## What

Re-plans **entity-editor Phase 2 (#605 — skeletal voxels)** against the
foundation that landed since it was filed in May, and files the 11 child
tickets under the umbrella.

## Why the re-plan

#605's original sub-tasks described deprecated machinery:

- `C_JointHierarchy` SoA → joints are now first-class entities
  (`C_Skeleton.joints_` + `C_Joint` + `CHILD_OF`, #737).
- "drive a GPU joint-matrix SSBO" / "stage-1 reads `bone_id`" → superseded
  by the #1396 GPU voxel-position prepass.
- FK pose editing → now free via `SYSTEM_PROPAGATE_TRANSFORM` (#734).

## Key architectural decision

**Unify skeletal skinning onto the #1396 prepass.** Rigid skinning (one bone
per voxel — exactly what a single `uint8 bone_id` expresses) *is* the
existing `transforms[slot] * localPos` in `c_update_voxel_positions.glsl`,
with the per-voxel transform slot pointing at a **bone** instead of an
**entity**. No new shader, no stage-1 change; the unused binding-21 joint
buffer is retired for the voxel path.

## Contents

- `docs/design/entity-editor-epic.md` — rewrites the canonical Phase 2
  section (scope, the #1396 skinning approach, sub-task table, deferred
  items, risks).
- `.fleet/plans/issue-605.md` — full epic plan: 11 children, dependency DAG.

Child tickets **#1602–#1612** filed; `fleet-validate-stack 605` passes
(11/11 carry the structured fields, #1602 is the unblocked head).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1612, opened 2026-06-09
test: render-verify + CPU↔GPU skinning consistency for rigged demos (P2.11)


jakildev:
**Model:** sonnet
**Part of epic:** #605
**Plan file:** `.fleet/plans/issue-605.md` (full epic plan)
**Blocked by:** #1611

## Scope
Automated render-regression + CPU↔GPU consistency coverage for the rigged
demos.

## Approach
- render-verify reference shots for the rigged demo entities (#1611) across
  the standard shot list.
- A consistency check: CPU-computed skin positions match the GPU prepass
  output under identical joint input (mirrors existing CPU↔GPU parity tests).

## Acceptance criteria
- render-verify green for the rigged demos; consistency check passes within
  tolerance; references committed.

## Dependencies
Blocked by #1611.

## Gotchas
Generate references on linux-debug (the fleet env) for cross-host stability.

## References
`render-verify` skill; existing CPU↔GPU parity tests (re-voxelize, #1559).

jakildev:
Scope check: this issue's scope appears to have landed in "docs: re-plan entity-editor Phase 2 (#605) — file children #1602-#1612" (#1614, merged 2026-06-09). Skipping ingest; please close this issue if the scope is fully addressed. — fleet-queue-ingest

jakildev:
Re-queued: same false positive as #1602 — #1614's title `file children #1602-#1612` matched the range **endpoint** `#1612` (the middle children #1603–#1611 were unaffected because only the two endpoints are literally written with `#`). Root-cause fix in flight.

Issue #1611, opened 2026-06-09
demos: skeletal demo entities — 30-joint snake + desk + 2 more (P2.10)


jakildev:
**Model:** sonnet
**Part of epic:** #605
**Plan file:** `.fleet/plans/issue-605.md` (full epic plan)
**Blocked by:** #1610, #1609

## Scope
Author demo-appropriate (not game-specific) sample entities that exercise the
rig + skinning path; doubles as #604's F-1.6 sample deliverable.

## Approach
- **30-joint snake** — the skeletal acceptance asset; a long chain that
  deforms end-to-end.
- **Desk** — a static (non-skeletal) authoring showcase for Phase 1.
- **2 more generic demo creatures** — e.g. an articulated lamp and a simple
  quadruped — varied joint counts and topologies.
- Save as `.vxs` + `.rig`; load in `IRShapeDebug` / a demo.

## Acceptance criteria
- The 30-joint snake loads and every segment deforms when a joint rotates.
- All four assets round-trip; none are tied to private game content.

## Dependencies
Blocked by #1610, #1609.

## Gotchas
Keep assets generic and engine-demo-appropriate — no game-specific creatures.

## References
`creations/editors/voxel_editor/`; `.rig`/`.vxs` formats; `IRShapeDebug`.

Issue #1607, opened 2026-06-09
editor: skeleton tree panel — joint list, select, rename, reparent (P2.6)


jakildev:
**Model:** sonnet
**Part of epic:** #605
**Plan file:** `.fleet/plans/issue-605.md` (full epic plan)
**Blocked by:** #1604

## Scope
A skeleton tree panel in the editor for selecting, renaming, and reparenting
joints.

## Approach
- Widget-list panel driven by `C_Skeleton.joints_` (O(joints); no children
  scan needed — the list is authoritative).
- Select highlights the joint marker gizmo; rename writes `C_JointName`;
  reparent rewrites the `CHILD_OF` relation and reorders `joints_`.

## Acceptance criteria
- Tree reflects the live hierarchy; selecting a row selects the joint in the
  viewport; rename/reparent round-trip through save/load.

## Dependencies
Blocked by #1604.

## Gotchas
Reparent must keep `joints_` ordering consistent with `bone_id` indices —
reorder bind pose in lockstep or keep indices stable and only change parents.

## References
`component_widget.hpp` (list widget); `layout.hpp`; `component_skeleton.hpp`.

PR #1629, opened 2026-06-09
fleet: scope-shipped must not trust range-endpoint refs in PR titles

jakildev:
## Summary
- `fleet-queue-ingest`'s scope-shipped pre-flight trusted any `#N` in a merged PR's **title**. An epic-planning PR names the children it *files* as a range — PR #1614's title `docs: re-plan entity-editor Phase 2 (#605) — file children #1602-#1612` — so the title-match fired on the range **endpoints** #1602 and #1612 and wrongly labelled both `fleet:scope-shipped`, silently skipping their ingest. The middle children #1603–#1611 escaped only because a range writes just its endpoints with a literal `#`.
- Adds a **range-endpoint guard** to `_ref_pattern` — the third rejection layer (alongside no-reference #1304 and mentioned-but-not-shipped #1260/#1269): a `#N` that is the start (`#1602` directly before `-#1612`) or end (`...-#1612`) of a `#A-#B` range is enumerating filed issues, never implementing one. Rejected in both title and body. Handles hyphen / en-dash / em-dash and a hashless second endpoint; a dash followed by a non-digit word is still not a range.
- Already-affected issues #1602 and #1612 were manually re-queued out of band.

## Test plan
- [x] `python3 -m unittest scripts.fleet.tests.test_scope_shipped_reference` — 34/34 pass (24 prior + 10 new)
- [x] New tests assert the live #1614 title ships none of #1602 / #1607 / #1612, and that the normal `#N: <desc>` convention + epic `(#605)` ref are unaffected (no regression)

### DIFF
--- a/scripts/fleet/fleet_scope_shipped.py
+++ b/scripts/fleet/fleet_scope_shipped.py
@@ -1,7 +1,7 @@
 """Genuine-ship predicate for fleet-queue-ingest's scope-shipped pre-flight.
 
 A merged PR is only evidence that an issue's scope "already landed" when the PR
-genuinely *ships* the issue — not when it merely names it. Two layers of
+genuinely *ships* the issue — not when it merely names it. Three layers of
 incidental match have to be rejected:
 
 1. No reference at all (#1304). GitHub's PR search (``gh pr list --search
@@ -16,11 +16,20 @@ incidental match have to be rejected:
    "pre-existing #1269" / "filed as #1269" (#1269 ← #1265), "Refs #N". These
    slipped past the layer-1 fix because the literal ``#N`` token is present.
 
+3. Range endpoint, not implemented (#1602 / #1612 ← #1614). An epic-planning PR
+   names the children it *files* as a range in its title — "docs: re-plan ...
+   — file children #1602-#1612". The title-trust below then matched the two
+   range *endpoints* (#1602 and #1612; the middle children #1603-#1611 were
+   untouched because only the endpoints are literally written with a ``#``). A
+   ``#N`` that is an endpoint of a ``#A-#B`` range is enumerating issues, never
+   shipping one, so range endpoints are rejected in both title and body.
+
 So a body ``#N`` counts only when a closing-action verb sits directly before it
 (``Closes #N``, ``fixes (#N)``, ``supersedes #N``). A ``#N`` in the *title* is
-trusted as-is: PRs in this repo are named ``#N: <desc>`` after the issue they
-implement, and every observed false positive came from the body, never the
-title. The asymmetry is deliberate.
+trusted as-is — PRs in this repo are named ``#N: <desc>`` after the issue they
+implement — EXCEPT when it is a range endpoint (layer 3): a planning PR names
+filed children in its title, and that is the one title shape that mentions
+without implementing. Every other observed false positive came from the body.
 """
 import re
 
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## First-pass review — #1629 fleet: scope-shipped must not trust range-endpoint refs in PR titles

**Verdict: approve.** Clean, minimal fix for a real false-positive class. Regex logic is correct, test coverage is solid, and the conservative bias (false negative → stays queued, never wrongly scope-shipped) is exactly right.

---

### Correctness — `_ref_pattern` regex ✅

The updated pattern for `n`:

```
(?<!\w)(?<![-–—])#N(?!\d)(?!\s*[-–—]\s*#?\d)
```

Traced the four critical cases:

- **Range END** (`-#1612`): `(?<![-–—])` lookbehind sees `-` directly before `#` → fails → rejected. ✓
- **Range START** (`#1602-#1612`): space before `#1602` passes both lookbehinds; `(?!\s*[-–—]\s*#?\d)` sees `-#1612` (dash + optional `#` + digit) → positive lookahead matches → negative lookahead fails → rejected. ✓
- **Normal title ref** (`#1602: bind-pose…`): both lookbehinds pass (preceding char is start-of-string); lookahead sees `:` → lookahead fails to match range pattern → negative lookahead passes → match accepted. ✓
- **Hashless second endpoint** (`#1602-1612`): same as START — `(?!\s*[-–—]\s*#?\d)` matches `-1612` (dash, no `#`, digit `1`) → rejected. ✓

`(?<![-–—])` is a fixed-width (1 char) lookbehind — valid in Python's `re` module. Unicode dashes `–` / `—` are handled correctly by Python 3's Unicode `re`. ✓

The body-path interaction is also correct: `_VERB_TO_REF_GAP = r'[\s:]*[(\[]?\s*'` can only consume whitespace, colons, and brackets — never a range dash — so the lookbehind `(?<![-–—])` is never spuriously triggered by the verb-to-ref gap. ✓

### Edge case acknowledged ✅

`#1602 - 3 improvements` (digit after dash but not a real range) would be rejected as a false negative. This is a deliberate conservative trade-off and is fine: (a) the engine's PR naming convention is `#N: <desc>`, not `#N - <desc>`, so this pattern is essentially unreachable in practice; (b) false negatives leave the issue queued rather than scope-shipped — the safe direction.

### Epic ref outside range unaffected ✅

`(#605) — file children #1602-#1612`: `#605` is followed by `)`, not a range dash, so `(?!\s*[-–—]\s*#?\d)` sees `)` which is neither whitespace nor a range-dash prefix → passes → `#605` correctly accepted. ✓

### ECS / render / ownership / serialization ✅

Not applicable — pure Python fleet scripting under `scripts/fleet/`. No C++ changes, no ECS systems, no GPU code.

### Tests / build ✅

- 10 new tests covering range-start, range-end, hashless-second-endpoint, en-dash, em-dash, epic-outside-range, plain-title-unaffected, dash-before-word, body-range-with-verb, and the real-world `#1614` integration case. Each tests the specific shape that triggered the bug.
- 34/34 pass per PR body.
- `test_real_world_1614_files_children_ships_none` in `SelectShippedPr` is the right end-to-end regression test for the actual incident.

---

### Test plan before merge

- [x] `python3 -m unittest scripts.fleet.tests.test_scope_shipped_reference` — 34/34 pass (confirmed in PR body).
- [x] Manually verify #1602 and #1612 are no longer scope-shipped on a fresh scout run (already done out-of-band per PR body).

Opus recheck not required.

PR #1691, opened 2026-06-11
fleet: task-scoped model classes + effort, fable opt-in, tier-0 rebases

jakildev:
## Summary
- **Three literal model classes** — `fable` | `opus` | `sonnet` — routed **per task**: new `fleet:fable` label, literal `Model:` parsing in ingest/backfill/claim-gate/scout, optional `**Effort:** low|medium|high|xhigh|max` task field. The dispatcher resolves each worker dispatch's class from the lane's projection slice (`fleet_task_class.py`: feedback PRs → open tasks → needs-plan) and launches claude with that class's model+effort; the claim gate exact-matches classes.
- **The budget flip**: fable is opt-in (hard rendering/algorithmic work, epic design, approach-was-wrong feedback fixes); `opus` is the default class — missing `Model:` fields and the existing `fleet:opus` backlog land on Opus 4.8, not Fable. `FLEET_CONCURRENCY_MODEL_FABLE` (default **1**) hard-caps concurrent fable iterations fleet-wide; cap-blocked lanes serve their next non-fable item or defer instead of burning gate-refused iterations.
- **Feedback fixes route from review severity**: `fleet:has-nits` → sonnet, `fleet:needs-fix`/`human:*` blocking → opus, reviewer-applied `fleet:fable` → fable. `fleet:design-blocked` escalation (architect/fable) unchanged.
- **Tier-0 merger** — new `fleet-rebase`: the dispatcher intercepts merger triggers and runs a mechanical pass first (scratch-worktree rebase of approved stacked/behind PRs, changed-file-set guard, `--force-with-lease`, parent-merged retarget to master). Zero tokens when clean; re-arms the trigger (content `llm`) only when conflicts/unhandled states remain. The merger LLM pass drops to **sonnet**; final reviewer pins to **opus**; `fleet:semantic-conflict` handoff unchanged.
- fleet-up's #1637 probe ladder now resolves the **fable class** (`FLEET_MODEL_FABLE/_OPUS/_SONNET` table + per-role `FLEET_MODEL_MERGER`/`FLEET_MODEL_OPUS_REVIEWER`); worker panes re-tag their border label per dispatch with the model actually running; `--resolve-class` inspection flag added; FLEET.md model-split rewritten; file-epic templates + `fleet-validate-stack` accept the new vocabulary.

## Stack context
Originally stacked on #1637 (merged); rebased onto `master` and
retargeted — the base-move dropped the parent's commit cleanly
(identical 21-file diff before/after, full suite green post-rebase).

[...]

PR #1699, opened 2026-06-11
fleet: target the dispatch pane explicitly when re-tagging its label

jakildev:
## Summary
- The per-dispatch pane label re-tag (from #1691) wrote `tmux set-option -p @role` **without `-t`**. TMUX_PANE doesn't reliably reach the dispatched process, so tmux fell back to the attached client's **active pane** — every dispatch re-tagged whatever pane the human had focused. On the first post-#1691 boot that was the engine-architect pane (fleet-up focuses it on attach), which cycled through `sonnet-reviewer [sonnet]` / `merger [sonnet]` labels while running architect content.
- Fix: reconstruct the exact pane id from the dispatch's own `PANE_KEY` (`pane-<N>` ↔ `%<N>`, the `pane_id_to_key` mapping) and pass `-t` explicitly. Re-tag is unambiguous regardless of focus.

## Test plan
- [x] `bash -n` clean
- [x] Explicit-target reconstruction verified against the **live** tmux server (scratch option write landed on the named pane, not the focused one)
- [x] Mislabeled architect pane repaired live (`opus-architect [fable 1m]`)
- [ ] After merge + pull: labels stay correct across dispatches (architect panes never re-tagged; worker panes flex per class)

### DIFF
--- a/scripts/fleet/fleet-dispatch-wrap
+++ b/scripts/fleet/fleet-dispatch-wrap
@@ -46,12 +46,21 @@ fi
 # Re-tag the pane border label with the model this iteration actually
 # runs (worker lanes flex per task), e.g. "opus-worker [fable 1m]".
 # Best-effort: a failure here must never block the dispatch.
+#
+# The target MUST be explicit. Without `-t`, tmux resolves `-p` against
+# TMUX_PANE — which does not reliably reach this process — and then
+# falls back to the attached client's ACTIVE pane, i.e. whatever pane
+# the human has focused. Observed live 2026-06-11: every dispatch
+# re-tagged the focused architect pane, cycling it through
+# "sonnet-reviewer"/"merger" labels. PANE_KEY is "pane-<N>" for tmux
+# pane id "%<N>" (see pane_id_to_key in fleet-dispatcher), so the exact
+# pane this dispatch landed in is reconstructible.
 _tag="$MODEL"
 _tag="${_tag/claude-fable-5/fable}"
 _tag="${_tag/claude-opus-4-8/opus 4.8}"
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — fleet: target the dispatch pane explicitly when re-tagging its label

**Verdict: approve**

Standalone PR targeting master — reviewing full diff.

---

### Summary

One-liner bug fix in `fleet-dispatch-wrap`: `tmux set-option -p @role` was called without `-t`, so tmux fell back to the attached client's focused pane instead of the dispatch pane. Fix reconstructs the exact pane id from `PANE_KEY` ("pane-N" → "%N") and passes `-t` explicitly.

---

### Correctness ✅

- `PANE_KEY` is the required first arg (`${1:?...}` at line 24) — always set before the re-tag line. ✓
- `${PANE_KEY#pane-}` correctly strips the "pane-" prefix, leaving the numeric id N. ✓
- `-t "%${N}"` targets the specific pane; `-p` scopes the option to that pane. The combination is the correct tmux idiom for setting a pane option on a named pane. ✓
- `|| true` guard preserved — failure is still non-blocking. ✓

The added comment explains the root cause (TMUX_PANE unreliable without `-t`, active-pane fallback, live observation) — exactly the right WHY for a non-obvious tmux behavior. ✓

---

### ECS / render / engine ✅

N/A — fleet scripting only.

---

### Test plan

- [ ] `bash -n scripts/fleet/fleet-dispatch-wrap` clean
- [ ] After merge + pull: dispatch a worker, confirm the worker's own pane label changes and no other pane (architect, human) is re-tagged

Opus recheck not required.

PR #1706, opened 2026-06-11
render: depth-color per-pixel on voxel scatter path (#1697)

### LINKED ISSUE #1697: render: --depth-color is per-voxel-banded on the voxel path — evaluate per-pixel like the SDF twin (diagnostic-only)
**Area:** engine/render — the `applyDepthColor` debug path for voxel-pool content (stage-2 / composite application site), GLSL + Metal.
**Model:** sonnet (bounded, debug-path only; escalate if the per-pixel application site gets structural)
**Origin:** PR #1625's #1457 re-diagnosis.

## Problem

`--depth-color` on the **voxel path** applies the hue palette **per-voxel-constant** with bands quantized at 4/3 world units. At any non-cardinal yaw the band boundaries beat against the 1-unit voxel lattice at near-maximal spatial frequency — a blocky moiré that *reads as front/back depth confusion*. The **SDF twin evaluates the same palette per-pixel** (continuous) and is immune by construction.

This is diagnostic-only (no product rendering is affected) — but the misleading output cost the fleet a four-investigation chain (#1414 → #1451 → #1457 → #1601 → #1625) before PR #1625's instrumentation proved the geometry was always correct. Fix the diagnostic so it can't manufacture that artifact again.

## Fix direction

Evaluate the depth-color palette **per-pixel** on the voxel path, matching the SDF twin's continuous evaluation. Per-fragment depth is available at the application sites: the per-axis scatter now carries an exact per-fragment planar depth (PR #1625's hardening), and the cardinal gather reads per-pixel depth from the distance texture. Apply the palette there rather than per-cell at store time. (Alternative if simpler: micro-granular bands sized ≪ the lattice step — but per-pixel is the principled fix and matches the twin.)

## Acceptance
- `--depth-color` at non-cardinal yaw (e.g. 67.5°, the #1625 worst case) shows a **smooth continuous gradient on voxel-pool content matching the SDF twin** — no per-voxel banding moiré.
- Cardinal output: visually equivalent (smoothness aside); normal (non-debug) rendering byte-identical.
- GLSL + Metal in lockstep.
- Note in the PR that `--checkerboard` remains the canonical occlusion/geometry diagnostic (per the `per-axis-trixel-canvas-rotation.md` note from PR #1625); this just makes depth-color stop lying.

jakildev:
Fixes #1697.

## Problem

`--depth-color` on the voxel path pre-bakes integer iso-depth colors per-voxel (CPU `applyDepthColor`). At non-cardinal yaw, 4/3-world-unit color bands beat against the 1-unit voxel lattice as staircase moiré, while the SDF twin evaluates per-pixel and looks smooth. The side-by-side comparison was structurally misleading for occlusion diagnosis.

## Fix

Add a `flat out float vIsoDepth` varying to the per-axis scatter shaders (`v_/f_peraxis_scatter.glsl` and `peraxis_scatter.metal`) carrying the face-center iso-depth (`origin.x + origin.y + origin.z`). Using `origin` (not `worldCorner`) ensures the value is constant across all 4 vertices of a face instance — flat is correct, and avoids provoking-vertex ambiguity. When `depthColorMode != 0`, the fragment shader evaluates hue from this face-center depth via `hsvToRgb`, matching the SDF twin.

New UBO fields `depthColorMode` / `depthColorExtent` (16 bytes appended to `FrameDataTrixelToFramebuffer` at offset 192 → sizeof=208, std140 append) carry the mode flag and normalization extent. Exposed via `IRRender::setDepthColorDebug(bool, float)`.

The cardinal gather path keeps the CPU pre-bake unchanged (no moiré at cardinal; the gather path does not use these shaders). Default mode (`depthColorMode = 0`) is byte-identical to the previous behavior on all shots — render-verify passes 21/21 (max_delta=0, psnr=inf).

## Changes

- `engine/render/include/irreden/render/ir_render_types.hpp` — extend `FrameDataTrixelToFramebuffer` (+16 bytes at offset 192), update `sizeof` assert (192→208), add `offsetof` assert
- `engine/render/src/shaders/v_peraxis_scatter.glsl` — `flat out float vIsoDepth` + mode/extent varyings, UBO extension, face-center assignment `vIsoDepth = origin.x + origin.y + origin.z`
- `engine/render/src/shaders/f_peraxis_scatter.glsl` — matching flat inputs, `hsvToRgb` helper (matches `c_shapes_to_trixel.glsl`), per-fragment depth-color when mode on
- `engine/render/src/shaders/metal/peraxis_scatter.metal` — Metal parity (same UBO fields, same `VertexOut` flats, same `hsvToRgb`, same fragment branch)
- `engine/render/include/irreden/render/render_manager.hpp` + `engine/render/src/render_manager.cpp` — `m_depthColorDebugOn`/`m_depthColorDebugExtent` fields + getter/setter
- `engine/render/include/irreden/ir_render.hpp` + `engine/render/src/ir_render.cpp` — `setDepthColorDebug(bool, float)` + getters
- `engine/prefabs/irreden/render/systems/system_trixel_to_framebuffer.hpp` — set UBO fields in `drawPerAxisScatter()`
- `creations/demos/shape_debug/main.cpp` — call `setDepthColorDebug(true, bh.x+bh.y+bh.z)` in the `--depth-color` branch; scatter evaluates per-face
- `creations/demos/shape_debug/test/references/macos-debug/` — refresh 2 stale yaw=45° baselines (pre-dated `bcd5652f` Jun 4 conservative-dilation merge)

## Visual evidence

**After — yaw=45° with `--depth-color` (scatter path, per-face)**

Smooth gradient, no moiré bands:

yaw45 depth-color after

**After — yaw=45° panned with `--depth-color`**

yaw45 pan16 depth-color after

**Cardinal (yaw=180°) with `--depth-color` — CPU gather path (unaffected)**

yaw180 depth-color cardinal

render-verify: 21/21 PASS (default mode, byte-identical, max_delta=0, psnr=inf)

Closes #1697


jakildev (commented):
## Review — render: depth-color per-pixel on voxel scatter path (#1697)

**Verdict: approve with nits**

Standalone PR on `master`. Moves `--depth-color` hue evaluation from a CPU per-voxel pre-bake to the fragment shader, eliminating the 4/3-lattice moiré at non-cardinal yaw. Default mode (depthColorMode=0) is byte-identical; render-verify 21/21 PASS confirms. Metal parity shipped in the same PR. Scope is well-constrained to the scatter path only.

---

### ECS invariants ✅

No ECS code touched. Not applicable.

### Ownership / lifetime ✅

`m_depthColorDebugOn` / `m_depthColorDebugExtent` are plain POD members. No shared/raw owning pointer introduced. `getDepthColorDebugMode()` / `getDepthColorDebugExtent()` are `const` on `RenderManager` and called per-draw — no cross-tick aliasing risk.

### Render pipeline ✅

**UBO extension — layout verified:**

C++ struct (offset 192, size 208):
- `offsetof(depthColorMode_) == 192` — static_assert present ✓
- `sizeof == 208` — static_assert present ✓

GLSL std140 UBO `FrameDataIsoTriangles`:
- `int depthColorMode` at 192, 4-byte aligned ✓
- `float depthColorExtent` at 196, `float _pad0` at 200, `float _pad1` at 204 ✓
- Total 16 bytes (4×4) appended → layout is 208 ✓

Metal struct `FrameDataIsoTriangles` — natural alignment matches GLSL std140:
- Struct layout traces to `float4 scatterFbResolution` ending at 192 → `int depthColorMode` at 192 ✓
- Both backends' struct layout is 208 bytes ✓

**Normalization formula — matches CPU `applyDepthColor` exactly:**
Shader: `dColor = vDepthColorExtent; denomC = max((4/3)*dColor, 1.0); t = clamp((vIsoDepth + dColor)/denomC, 0, 1)` matches `main.cpp:600-607`. ✓

**`flat` qualifier on `vIsoDepth`:** Correct — `origin` is constant for all 4 vertices of a face instance; `flat` avoids provoking-vertex ambiguity and prevents unnecessary interpolation setup cost. ✓

**Degenerate/culled instance:** New varyings (`vIsoDepth=0, vDepthColorMode=0, vDepthColorExtent=0`) initialized on the early-return path — no UB from uninitialized flat inputs. ✓

**Binding / shader name correctness:** No new binding points added; new UBO fields are appended to binding=3 (FrameDataIsoTriangles), which already bound in both scatter shaders. ✓

**Gather fast path unaffected:** `drawPerAxisScatter()` is the only call site that sets `depthColorMode_` / `depthColorExtent_`; the cardinal gather shaders don't read these UBO fields (declared at offset 192 which is past the gather shader's declared block). ✓

**Metal parity:** `peraxis_scatter.metal` receives matching `VertexOut` flats (`isoDepth`, `depthColorMode`, `depthColorExtent`), the same fragment branch, and an equivalent `hsvToRgb`. Both `.glsl` and `.metal` changed in the same commit. ✓

### Lighting ✅

No lighting system, shadow, AO, or flood-fill code changed. Not applicable.

### Math / coordinates ✅

`vIsoDepth = origin.x + origin.y + origin.z` — correct iso-depth (world-space sum). No axis swap. ✓

### Serialization ✅

No serialized structs changed.

### Naming / style ✅ (with nit)

See Nits section — `m_` prefix on new private `RenderManager` members follows existing file convention but diverges from CLAUDE-BASELINE.

### Tests / build ✅

render-verify 21/21 PASS at depthColorMode=0 (default). Two yaw=45° baselines refreshed (stale since the Jun 4 conservative-dilation merge #1494); documented in PR body. No new test harness needed for a debug-mode flag — visual evidence provided in PR screenshots.

---

### Nits

- `engine/render/src/shaders/f_peraxis_scatter.glsl:22-38` — The `hsvToRgb` helper comments say "Matches hsvToRgb in c_shapes_to_trixel.glsl" but the implementations differ algorithmically: `c_shapes_to_trixel.glsl` uses the vectorized `fract/mix` form; the new helper uses a sector-based switch. Both are mathematically equivalent for the actual input range `h ∈ [0, 0.66]`, `s=1`, `v=1` (sectors 0–3 only; sector 4 and the else branch are unreachable). For perfect bit-for-bit parity with the SDF twin, consider adopting the same `fract/mix` formula. Same applies to `engine/render/src/shaders/metal/peraxis_scatter.metal:163-178`. Low priority — visually indistinguishable for this use case.

- `engine/render/include/irreden/render/render_manager.hpp:163-164` — New private members `m_depthColorDebugOn` / `m_depthColorDebugExtent` use `m_` prefix, consistent with the rest of `RenderManager`'s private member naming (e.g. `m_sunShadowsEnabled`, `m_aoEnabled`). However, CLAUDE-BASELINE specifies trailing `_` for private members and `m_` for public. The pre-existing drift in this file is the root cause, not this PR; follow-up cleanup of `render_manager.hpp` would remove the inconsistency for the whole class.

---

### Praise

- Comprehensive `static_assert` suite (offset + size) guards the UBO layout on both sides — this pattern catches layout drift at compile time with no runtime overhead. Well done.
- PR body explains the 4/3-lattice moiré root cause clearly (integer iso-depth per-voxel at non-cardinal yaw beats against the 1-unit lattice), which makes the fix self-justifying.
- Using `origin` (not `worldCorner`) for `vIsoDepth` is the correct choice — avoids provoking-vertex ambiguity and keeps the value constant-per-quad as stated.

---

- [ ] Run `IRShapeDebug --depth-color` at yaw=45° and confirm smooth gradient (no moiré) matches the PR screenshots.
- [ ] Run `fleet-verify` (or render-verify) in default mode (no `--depth-color`) to confirm 21/21 byte-identical baseline. The updated yaw=45° baselines should match the current conservative-dilation code.
- [ ] Build on Linux (OpenGL) — PR was authored on macOS (Metal); cross-host compile validation.

Opus recheck required: touches engine/render/ shader code (GLSL + Metal), UBO layout extension at offset 192, and Metal/GL struct alignment cross-verification.

jakildev:
Addressed nits from PR #1706 review:

- **Nit 1 (hsvToRgb — GLSL + Metal):** Replaced the sector-based switch implementation with the canonical `fract/mix` form from `c_shapes_to_trixel.{glsl,metal}`. Both `f_peraxis_scatter.glsl` and `peraxis_scatter.metal` now use the identical formula, so voxel-scatter depth-color is bit-exact with the SDF twin. Updated comments from "Matches" to "Identical to" to reflect this accurately.

- **Nit 2 (m_ prefix on render_manager.hpp members):** Filed follow-up issue #1708 for a whole-class rename of `m_`-prefixed private members in `render_manager.hpp` → trailing `_`. Renaming only the two new members in this PR would create per-file inconsistency; the cleanup should be a single pass over the full class.

Build + smoke run passed (21/21 shots, --depth-color mode, no shader errors).

— sonnet-fleet-1

jakildev (commented):
## Opus recheck — render: depth-color per-pixel on voxel scatter path (#1697)

**Verdict: approve with nits** — the core engineering is sound; one debug-mode fidelity limitation to note (below), not a blocker.

Sonnet flagged this for recheck on three Opus-only axes — shader code (GLSL + Metal), the UBO layout extension at offset 192, and Metal/GL struct alignment. I re-derived all three and confirm them; I also chased the CPU↔shader formula parity and the per-shape semantics Sonnet's pass didn't reach.

### UBO std140 layout — confirmed correct ✅
`scatterFbResolution_` (vec4) sits at offset 176 and ends at 192. The four appended scalars `int depthColorMode_ / float depthColorExtent_ / float _depthColorPad0_ / float _depthColorPad1_` are each 4-byte-aligned in std140 and pack tightly at 192/196/200/204 → struct end 208 (= 13×16, a valid std140 multiple-of-16 size). Both `static_assert`s (`offsetof(depthColorMode_) == 192`, `sizeof == 208`) encode exactly this and would fire on any silent reorder. C++ ↔ GLSL `v_peraxis_scatter.glsl` block match byte-for-byte.

### Metal struct alignment — confirmed correct ✅
`FrameDataIsoTriangles` in `peraxis_scatter.metal` mirrors the same trailing-scalar order after `float4 scatterFbResolution` (offset 176→192), and MSL scalar alignment (int/float = 4) packs them identically at 192/196/200/204. The CPU writes one buffer; GL and Metal read the same bytes — no divergence.

### Shader logic — GL/Metal identical, formula matches the references ✅
The vertex→fragment varyings are correctly `flat` on both backends (`flat int` is mandatory for the mode; iso-depth is constant-per-quad so flat is a no-op interpolation, not a precision loss). The degenerate-vertex cull path zero-inits the new varyings on both sides.

The fragment hue formula is line-for-line identical between `f_peraxis_scatter.glsl` and `peraxis_scatter.metal`, and matches both reference implementations exactly:
- SDF twin `c_shapes_to_trixel.glsl:959-964`: `dColor = boundingHalf sum; denomC = max((4/3)·dColor, 1); t = clamp((depth + dColor)/denomC, 0, 1); hsvToRgb(0.66·t, 1, 1)`.
- CPU `applyDepthColor` (`shape_debug/main.cpp:594-617`): same `dColor`, same `denom`, same `t`, same `0.66·t`.

The scatter path feeds `vIsoDepth = origin.x+y+z` (un-yawed world origin, = the CPU path's `cellPos.x+y+z` at cardinal yaw) and the object-space `boundingHalf` sum as extent — so it is byte-identical to the CPU bake at cardinal yaw and continuous (no 4/3-lattice moiré) off-cardinal. That is precisely the #1697 fix, and it's correct.

### mode-0 / blast radius — confirmed safe ✅
`m_depthColorDebugOn` defaults false; only `shape_debug` ever calls `setDepthColorDebug`. With mode 0 the fragment shader returns `vColor` unchanged, so every other demo and the gather fast path stay byte-identical. render-verify 21/21 PASS is consistent with this (the manifest shots don't pass `--depth-color`; the two refreshed macOS yaw45 refs are sub-threshold Metal speckle, not a logic change).

---

### Nit / follow-up (non-blocking): the depth-color extent is global, but the scene is multi-shape

`setDepthColorDebug(on, extent)` stores a **single** extent on `RenderManager`, and `system_trixel_to_framebuffer` writes that one value into the scatter UBO for the whole pass. But `createVoxelPoolShape` calls `setDepthColorDebug(true, bh.x+bh.y+bh.z)` **once per shape** inside the 8-case loop (`main.cpp:865-877`) — Sphere(≈12), Cylinder(≈13), Ellipsoid(≈18), Cone(≈16), Wedge(≈21), CurvedPanel(≈18), … — so it's **last-writer-wins**: every voxel-pool shape in the row renders normalized to the *final* case's extent.

The SDF twin row (`createSDFShape`, `main.cpp:879`) does the opposite — it tags `SHAPE_FLAG_DEPTH_COLOR` per entity and `c_shapes_to_trixel` derives `boundingHalfView` **per shape**, so each SDF shape is normalized to its own extent. So the voxel row and its SDF twin will only agree in gradient *scale* for the last shape; the other seven get a compressed/stretched (but still smooth) gradient.

Why this is a nit and not a needs-fix:
- It's confined to the `--depth-color` diagnostic; production rendering and committed references are untouched.
- The #1697 goal — kill the moiré banding — is fully met regardless of extent, because per-pixel evaluation is what removes the beat against the 1-unit lattice. Absolute hue *scale* per shape is the secondary signal.
- The two rows already differ by a constant hue offset from their world-Y separation (`d` is absolute world depth), so they were never a pixel-match anyway — smoothness is the real comparison.

Suggested follow-up (author's or human's call, not required for this PR): either (a) drive the extent per-shape — e.g. bake it into a per-voxel/per-instance channel the scatter path reads, so it matches the SDF twin's per-shape `boundingHalfView`, or (b) document on `setDepthColorDebug` that it's a single-shape diagnostic and that multi-shape rows share the last extent, so the limitation is explicit. The doc comment on `setDepthColorDebug` currently reads "the bounding half-sum of the shape being diagnosed," which implies per-shape fidelity the global state doesn't deliver.

---

Net: layout, parity, and formula are all correct and the fix does what #1697 asks. Approving with the global-extent limitation noted for cleanup.

🤖 Reviewed by Claude Opus 4.8 (opus-reviewer, review-pr skill) — final recheck on Sonnet's escalation

jakildev:
Addressed Opus nit on PR #1706:

- **Doc comment clarification (ir_render.hpp:384-386):** Updated the `setDepthColorDebug` group comment to clarify that `extent` is a single global value on `RenderManager` (last-writer-wins in multi-shape scenes), not a per-shape value. The prior wording ('the shape being diagnosed') implied per-shape fidelity that the global state doesn't deliver.

No code change — diagnostic-only API; `fleet:approved` stands.

— sonnet-fleet-2

jakildev:
Cross-host smoke OK on macos-Metal (fresh checkout build + IRShapeDebug --auto-screenshot 21 frames). Build clean, run exited 0, all 21 shots captured. No moiré, no visual artifacts at non-cardinal yaw. Default mode renders correctly on the Metal backend. — opus-worker-1

PR #1729, opened 2026-06-12
audio: multi-port MIDI I/O with per-port source identity

### LINKED ISSUE #1727: audio: multi-port MIDI I/O — simultaneous input/output ports with port identity
[opus] Filed from game-repo planning (jakildev/irreden#174 — studio rig live demo). The game's physical rig has TWO MIDI sources: an OP-1 Field over USB and a UMC audio interface (DIN chain to five pedals/synths). Today the engine cannot listen to both.

**Area:** engine/audio, engine/prefabs/irreden/audio

## Current state (verified)

- `MidiIn` holds a single `RtMidiIn` (engine/audio/include/irreden/audio/midi_in.hpp:52); `openPort(i)` on it (midi_in.cpp:81) means opening a second input replaces the first. An `m_rtMidiInMap` member exists (midi_in.hpp:53) but is never used — multi-port was anticipated, never built.
- Messages are merged into per-channel queues with no port identity (MidiMessageQueues, midi_in.hpp:21-28; C_MidiMessage carries only status/data1/data2).
- `MidiOut` likewise holds a single `RtMidiOut` (midi_out.hpp:28); sendMidiMessage targets "the" open output (midi_out.cpp:62).

## Wanted

1. Multiple simultaneously-open MIDI input ports (openMidiIn returns a port handle/id; each port feeds the queues).
2. Source-port identity on incoming messages (tag in the queues + C_MidiMessage or a parallel field) so consumers can disambiguate same-channel traffic across ports.
3. Multiple output ports; sends targetable to a specific port (default port preserves the current single-port API for back-compat).
4. Lua-facing surface (via the creations' binding layer) able to scope handlers and sends by port — back-compatible: existing channel-scoped handlers keep working, port becomes an optional extra discriminator.
5. Stretch (separate PR acceptable): expose MIDI clock and Program Change to the per-frame queues so creations can visualize them.

## Acceptance

- Two virtual/real input ports open at once, each delivering to callbacks with correct port identity; unit or demo-level verification.
- Existing single-port consumers (creations/demos, game midi projects) build and run unchanged.
- fleet-build clean on linux-debug and macos-debug.

## Game-side consumers waiting on this

- MIDI monitor UI (per-port lanes; currently designed to degrade to one-input-at-a-time with a config toggle).
- Dual live input: OP-1 USB knobs + UMC-chain synth knobs simultaneously.

jakildev:
## Summary
- `MidiIn`/`MidiOut` hold one RtMidi instance per simultaneously-open port (each input port with its own callback queue) instead of a single shared one. `openPortMidiIn`/`openPortMidiOut` return the RtMidi port index as a stable handle; re-opening an open port is an idempotent no-op.
- Inbound message entities carry `C_MidiSourcePort{portIndex}`; the `InputMidiMessageIn` drainer routes by port into a new hardware-free `MidiInputFrameBuffer` that keeps both a **merged** all-ports per-channel view and a **per-port** view.
- Port-scoped overloads added alongside the merged-view API: `checkCCMessage(port, ch, cc)`, `getMidiNotesOn/OffThisFrame(port, ch)`, `sendMidiMessage(port, msg)`, plus `midiInOpenPorts()`/`midiOutOpenPorts()` to enumerate open handles (per-port monitor lanes).
- **Back-compat:** existing single-port consumers (`command_manager`, the game `midi_input_dispatch`/`gear_model`, the `midi_keyboard` demo) call the unchanged merged-view API and are untouched. Default `sendMidiMessage(msg)` targets the first-opened output port.
- Stretch item from the issue (MIDI clock / Program Change into the per-frame queues) is deferred to a separate PR, as the issue allows.

## Test plan
- [x] New unit test `test/audio/midi_multiport_test.cpp` (7 cases): per-port CC/note isolation, merged fold, unopened-port/absent-CC → `kCCFalse`, port-less insert stays merged-only, `clear()` wipes both views. No MIDI hardware required (tests `MidiInputFrameBuffer` directly).
- [x] `fleet-build --target IrredenEngineTest` clean on **macos-debug**; all 12 MIDI tests pass (7 new + 5 existing `music_theory`).
- [ ] linux-debug build (cross-host smoke).

## Notes for review
- The `InputMidiMessageIn` drainer now matches `<C_MidiMessage, C_MidiSourcePort>`. This is a pure **narrowing**: `MidiIn::tick()` runs before `executePipeline(INPUT)` and inbound entities are `C_Lifetime{1}`, so the drainer never coexisted with outbound `C_MidiMessage` entities anyway. Net effect is correct port routing with no behavior change for the existing flow.
- `MidiInPort` declares `rtMidiIn_` last so it destructs first — the RtMidi callback thread stops before its target `queue_` is freed (the prior single-port code had the same latent teardown ordering).
- **Reuse, consciously kept:** `MidiIn`/`MidiOut` `getOpenPortIndices` + `openPort(string)` are near-duplicates. Left un-extracted because the only shared home is the widely-included audio types header, and the divergence (input callback wiring, distinct port structs) makes a template helper net-negative coupling for ~2 call sites. Worth revisiting if a third port type lands.
- **No `optimize` profile run:** MIDI input is event-bounded I/O (no per-frame compute hot loop), and there's no headless workload to drive MIDI messages. The per-frame buffer is allocation-free across frames — `clear()` wipes contents but reuses map/vector capacity.

Closes #1727

### DIFF
--- a/engine/audio/CLAUDE.md
+++ b/engine/audio/CLAUDE.md
@@ -18,23 +18,38 @@ functions:
 
 ## AudioManager owns
 
-- **`MidiIn`** — one `RtMidiIn`, populates per-channel maps of
-  `MidiMessage`s each `tick()`. Cleared and refilled every frame.
-- **`MidiOut`** — one `RtMidiOut`, fire-and-forget send.
+- **`MidiIn`** — one `RtMidiIn` per simultaneously-open input port, each
+  with its own callback queue. `tick()` drains every port's queue into a
+  `MidiInputFrameBuffer` (cleared and refilled each frame) that keeps both a
+  merged all-ports per-channel view and a per-port view.
+- **`MidiOut`** — one `RtMidiOut` per open output port; fire-and-forget send
+  to the default (first-opened) port, or to a specific port by index.
 - **`Audio`** — `RtAudio` wrapper for microphone/line-in. Implements
   `IAudioCaptureSource`. Used by `VideoManager` for recording soundtracks.
 
 ## MIDI model
 
-- **`MidiChannel`** is the map key. `MidiIn` exposes
-  `ccMessagesThisFrame(channel)`, `noteOnsThisFrame(channel)`, etc.
-- `MidiIn::tick()` is called by the INPUT pipeline; it drains the RtMidi
-  callback queue into per-frame containers. **State is cleared every
-  frame** — no history. Systems must read during their tick or they miss
-  the event.
+- **`MidiChannel`** is the map key. Query `IRAudio::checkCCMessage(channel,
+  cc)` / `getMidiNotesOnThisFrame(channel)` for the **merged** all-ports
+  view, or the port-scoped `checkCCMessage(portIndex, channel, cc)` /
+  `getMidiNotesOnThisFrame(portIndex, channel)` to disambiguate same-channel
+  traffic across ports. `openPortMidiIn`/`openPortMidiOut` return the RtMidi
+  port index — the stable handle used as the source-port id, the
+  `sendMidiMessage(portIndex, …)` target, and the per-port query key.
+  `midiInOpenPorts()` / `midiOutOpenPorts()` enumerate the open handles.
+- Inbound message entities carry `C_MidiSourcePort{portIndex}`; the
+  `InputMidiMessageIn` system reads it to route each message into both the
+  merged and per-port views. **Merged CC collapses same-channel+cc traffic
+  (last write wins)** — use the per-port query when two devices share a
+  channel.
+- `MidiIn::tick()` is called by the INPUT pipeline; it drains every open
+  port's RtMidi callback queue into the per-frame buffer. **State is cleared
+  every frame** — no history. Systems must read during their tick or they
+  miss the event.
 - Device names are substring-matched at `openPort()` time, with hardcoded
   fallback patterns for common devices (UMC1820, Focusrite, MPKmini2,
   OP-1). Edit `midi_in.cpp` / `midi_out.cpp` to add a new hardware match.
+  Opening an already-open port is a no-op that returns the existing handle.
 
 ## Audio capture model
 
--- a/engine/audio/include/irreden/audio/midi_in.hpp
+++ b/engine/audio/include/irreden/audio/midi_in.hpp
@@ -86,7 +86,7 @@ class MidiIn {
     void clearPreviousMessages();
 };
 
-void readMessageTestCallbackNew(
+void onRtMidiMessage(
     double deltaTime, std::vector<unsigned char> *message, void *userData
 );
 
--- a/engine/audio/include/irreden/audio/midi_in.hpp
+++ b/engine/audio/include/irreden/audio/midi_in.hpp
@@ -4,27 +4,32 @@
 #include <irreden/ir_profile.hpp>
 
 #include <irreden/audio/ir_audio_types.hpp>
-#include <irreden/audio/midi_messages.hpp>
+#include <irreden/audio/midi_input_frame_buffer.hpp>
 #include <irreden/audio/components/component_midi_message.hpp>
 
 #include <RtMidi.h>
 
-#include <set>
+#include <memory>
 #include <queue>
-#include <unordered_map>
+#include <string>
+#include <vector>
 
 using IRComponents::C_MidiMessage;
 
 namespace IRAudio {
 
-struct MidiMessageQueues {
-    std::queue<MidiMessage<kMidiStatus_NOTE_OFF>> m_messageQueueNoteOff;
-    std::queue<MidiMessage<kMidiStatus_NOTE_ON>> m_messageQueueNoteOn;
-    std::queue<MidiMessage<kMidiStatus_POLYPHONIC_KEY_PRESSURE>> m_messagePolyKeyPressure;
-    std::queue<MidiMessage<kMidiStatus_CONTROL_CHANGE>> m_messageQueueControlChange;
-    std::queue<MidiMessage<kMidiStatus_PROGRAM_CHANGE>> m_messageQueueProgramChange;
-    std::queue<MidiMessage<kMidiStatus_CHANNEL_PRESSURE>> m_messageQueueChannelPressure;
-    std::queue<MidiMessage<kMidiStatus_PITCH_BEND>> m_messageQueuePitchBend;
+// One simultaneously-open MIDI input port. Each port owns its own RtMidiIn and
+// a single-producer (its RtMidi callback thread) / single-consumer (the main
+// thread, in processMidiMessageQueue) message queue, so two devices can feed
+// the engine at once without one openPort replacing another.
+//
+// rtMidiIn_ is declared last so it destructs first: closing the port stops the
+// callback thread before queue_ (its write target) is torn down.
+struct MidiInPort {
+    std::queue<C_MidiMessage> queue_;
+    int portIndex_;
+    std::string name_;
+    std::unique_ptr<RtMidiIn> rtMidiIn_;
 };
 
 class MidiIn {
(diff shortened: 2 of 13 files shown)


jakildev (commented):
## Review — audio: multi-port MIDI I/O with per-port source identity (#1727)

**Verdict:** approve (Opus recheck required — see footer)

Standalone PR (base: master, MERGEABLE). Touches `engine/audio/` — escalating to Opus per protocol.

---

### Summary

Introduces hardware-free `MidiInputFrameBuffer` with merged + per-port channel-state views, refactors `MidiIn` to manage `std::vector<std::unique_ptr<MidiInPort>>`, and stamps inbound ECS entities with a `C_MidiSourcePort{portIndex_}` component so downstream consumers can route by origin port. Seven unit tests cover per-port isolation, merged fold, clear, and portless insertion.

---

### Correctness walkthrough

**`MidiInputFrameBuffer`**
- `kEmptyMessages` is `inline static const` — correct; single definition across TUs.
- `clear()` wipes map contents, preserves keys — correct; avoids full teardown of nested structures.
- `notesOn` / `notesOff` / `checkCC` return `const &kEmptyMessages` for unknown port/channel — clean fallback; no UB.

**`MidiInPort` teardown ordering**
`rtMidiIn_` is declared last in the struct, so it destructs first — the RtMidi callback thread stops before `queue_` is freed. The PR notes this mirrors the prior single-port latent ordering.

**Callback `userData` pointer stability**
`port->rtMidiIn_->setCallback(readMessageTestCallbackNew, &port->queue_)` is called before `std::move(port)` into `m_ports`. Moving a `unique_ptr` relocates the wrapper, not the heap object — `&port->queue_` remains valid after the move. No dangling pointer.

**Duplicate-open guard**
`openPort(string)` scans `m_ports` by `portIndex_` before creating a new `MidiInPort` and returns `i` early on a match. No double callback registration is possible.

**ECS narrowing**
`createSystem<C_MidiMessage, C_MidiSourcePort>` narrows the previous `<C_MidiMessage>` archetype. Inbound entities stamped by the callback loop carry both components; outbound entities (created without `C_MidiSourcePort`) fall outside the archetype and are correctly skipped. Pure narrowing — no existing behavior removed.

**Naming**
`C_MidiSourcePort::portIndex_` — trailing `_` on a public struct member. Correct per engine convention.

**`MidiOut::sendMessage`**
Logs a warning when no port is open instead of asserting. Defensive and correct for production.

---

### Nits

- **`readMessageTestCallbackNew`** — "Test" infix survives into the production `openPort` path (`midi_in.hpp:89`, `midi_in.cpp:151`). The name signals test/experimental code to a reader but this is the live MIDI input callback. Consider renaming to `onRtMidiMessage` or `midiInCallback`.
- **Per-byte `IRE_LOG_INFO` in callback** (`midi_in.cpp:160-162`) — emits one log line per message byte on every incoming MIDI event. Under any real MIDI stream (real-time clock, dense CC sweep) this floods the log. Lower to debug-level or remove; the existing "Opened MIDI In port" info log is sufficient for diagnostics.

---

### Test coverage

7 tests in `test/audio/midi_multiport_test.cpp` — per-port CC isolation, merged fold, clear preserves keys, portless-insert into merged only, empty port query, multi-channel independence.

`test/CMakeLists.txt` one-line addition — build coverage.

---

### Praise

- Hardware-free `MidiInputFrameBuffer` as a pure data accumulator is the right abstraction layer — keeps the ECS drainer clean and the test harness trivial.
- `C_MidiSourcePort` as a stamp component (not a parent relation, not a shared global) is idiomatic ECS and makes per-port routing a simple archetype query.
- Duplicate-open guard and early-return makes `openPort` idempotent — good defensive behavior for Lua-side callers that may call it multiple times.

---

### Opus recheck required: engine/audio/ + std::queue thread safety

Two items for Opus:

1. **`engine/audio/` escalation trigger** — per REVIEWER-PROTOCOL, all `engine/audio/` changes require an Opus pass.
2. **`std::queue` concurrent access** — `readMessageTestCallbackNew` pushes to `port->queue_` from the RtMidi callback thread; `processMidiMessageQueue()` reads and pops from the same queue on the main thread. `std::queue` is not thread-safe. This is a pre-existing pattern (the old `m_messageQueue` had the same race) so the PR is not introducing a new hazard, but the multi-port refactor multiplies the number of such queues. Opus should verify whether RtMidi provides any ordering or synchronization guarantees around its callback dispatch that make this safe in practice, and whether a lock-free queue or a mutex is warranted.

jakildev:
Addressed nits:
- Renamed `readMessageTestCallbackNew` → `onRtMidiMessage` in `midi_in.hpp` and `midi_in.cpp` (removes misleading Test infix from the production callback path)
- Lowered per-byte `IRE_LOG_INFO` in callback body to `IRE_LOG_DEBUG` to prevent log flooding under dense MIDI streams

Build: clean (IRShapeDebug). — sonnet-fleet-1

jakildev (commented):
## Opus recheck — audio: multi-port MIDI I/O with per-port source identity (#1729)

**Verdict:** approve

Sonnet's first pass was thorough and correct; this Opus pass focused on the two items it explicitly escalated (`engine/audio/` trigger + `std::queue` cross-thread access) plus the deeper ECS-narrowing, lifetime, and API-compat invariants it couldn't fully confirm from the hunks alone. All clear.

---

### 1. `std::queue` cross-thread access — Sonnet flagged this; on closer read I confirm it is a **pre-existing SPSC race, not a new hazard, and not a blocker for this PR**

Each `MidiInPort` owns its own `RtMidiIn`, and RtMidi spins **one** input thread per `RtMidiIn` instance. The callback `onRtMidiMessage` is registered with `&port->queue_` as user data, so every `queue_` has exactly **one** producer (its own callback thread) and exactly **one** consumer (`processMidiMessageQueue` on the main thread). The refactor turns the single shared `m_messageQueue` into N *independent* strictly-SPSC queues — it does not introduce cross-thread contention that wasn't already there, and the queues don't share state with each other.

That said, Sonnet is right that an unsynchronized `std::queue` (deque-backed) under concurrent `push`/`pop` is genuine UB even in the SPSC case — a `push` that reallocates the deque map can tear against a concurrent `front()`/`pop()`. This was already true of the old single-port code, so it's out of scope here, but the multi-port refactor is the natural moment to fix it. **Recommend the human file a follow-up** to back each port with a proper lock-free SPSC ring (or a mutex) rather than a bare `std::queue`. Non-blocking.

`m_frameBuffer` itself has **no** threading concern — it is only ever touched on the main thread (drainer writes it, same-frame consumer systems read it). The callback thread touches `queue_` and nothing else. Clean separation.

### 2. `MidiInPort` teardown ordering — confirmed correct

`rtMidiIn_` is the last-declared member, so it destructs first (reverse declaration order). `~RtMidiIn` stops the callback thread before `queue_` (its write target) is freed. The `std::move(port)` into `m_ports` relocates the `unique_ptr` wrapper, not the heap `MidiInPort`, so `&port->queue_` captured by `setCallback` stays valid post-move. Both confirmed.

### 3. ECS archetype narrowing — confirmed safe three systems deep

`createSystem<C_MidiMessage, C_MidiSourcePort>` narrows `InputMidiMessageIn` to inbound entities only (the drainer stamps `C_MidiSourcePort`; outbound `C_MidiMessage` entities lack it and fall through to OUTPUT). Critically: **adding** `C_MidiSourcePort` to the inbound archetype cannot drop those entities from any *other* system's view — the ECS matches on "has at least these components," so a wider archetype still satisfies every narrower query. No existing consumer of inbound `C_MidiMessage` regresses. `C_MidiSourcePort` is a new component, not a new system, so no `SystemName` enum entry is required.

### 4. API rename `device` → `channel` — confirmed honest, no caller breaks

The old `checkCCMessage(int device, …)` / `getMidiNotes{On,Off}ThisFrame(int device)` always forwarded arg1 straight into the channel slot of `checkCCMessageThisFrame(channel, …)` — i.e. it was a channel all along, just mislabeled. The new names match reality. Same arity + types, so overload resolution against the new 3-arg per-port variants is unambiguous and every existing caller (`engine/command/src/command_manager.cpp:42,116,123`, plus the game `midi/` consumers) binds to the same 2-arg overload with byte-identical behavior.

Side observation for the human (non-blocking, separate consumer-side work): `CommandManager`'s per-`device` MIDI command maps actually route by **channel** today (it passes its `device` int straight to the channel-keyed query). The new per-port `#1727` API is the correct primitive to eventually make that routing *genuinely* per-device — worth a note on whichever issue owns the command/MIDI binding layer.

### 5. `MidiOut` default-send back-compat — confirmed

Port-less `sendMessage` now targets `m_ports.front()` (first opened) and warns-instead-of-throws when no port is open — functionally equivalent to the old single-port path, strictly safer on the empty case. `m_rtMidiOut` correctly demoted to an enumeration-probe-only handle.

---

### Tests / docs

The 7 `MidiInputFrameBuffer` routing tests cover the acceptance-critical merged-vs-per-port contract deterministically without hardware — the right testable seam. The `engine/audio/CLAUDE.md` and prefab `CLAUDE.md` updates accurately describe the new merged/per-port model and the `C_MidiSourcePort` stamp. Both nits from Sonnet's pass (the `onRtMidiMessage` rename and the per-byte log downgrade to `IRE_LOG_DEBUG`) are addressed in the latest commit.

Clean refactor, idiomatic ECS, well-documented. Approving. The only latent concern (unsynchronized SPSC queue) predates this PR — surfacing it for a follow-up rather than holding the merge.

PR #1732, opened 2026-06-12
render: GRID inverse re-voxelize — close #1619's GRID half (REBUILD_GRID_VOXELS coverage holes) (#1720)

### LINKED ISSUE #1619: render: re-voxelize round-to-cell forward-scatter leaves coverage holes (missing faces) on DETACHED_REVOXELIZE / GRID solids
## Symptom
Voxel faces missing / holes on small `DETACHED_REVOXELIZE` cubes (and thin shapes) while they spin — especially visible on the small re-voxelize cubes that #1589 migrated to this path. **Persists after the #1570/#1573 exposed-mask "emit 3 cardinal faces" fix** — that fix addressed which faces to *emit* (visibility), not surface *coverage*. Different bug.

## Repro
`canvas_stress`, default run — the center re-voxelize cube + the small canary cubes show gaps across the spin. The demo author already documents the limitation in `creations/demos/canvas_stress/main.cpp` (~line 890): *"Both CELL-REBAKING paths — GRID and DETACHED_REVOXELIZE — render detailed SOLID shapes cleanly … What still aliases on those paths is THIN geometry (round-to-cell leaves gaps in wireframes / arms)"* — and deliberately routes thin shapes (frame / cross) **away** from re-voxelize to dodge it.

## Root cause
The re-voxelize fill `engine/render/src/shaders/c_revoxelize_detached.glsl` (+ Metal twin + the CPU mirror `IRPrefab::GridRotation::worldCellForGridVoxel`) is a **forward scatter**: one thread per *source* voxel computes `cell = roundHalfUp(rotateByQuat(local, quat))` and writes that single cell. A forward map under rotation+round is **not surjective** onto the rotated-AABB lattice — destination cells the rotated solid genuinely covers get no source voxel -> holes. On small / thin solids (mostly surface, little interior to backfill) those holes read as missing faces. The 2D +/-1px conservative dilation at raster only closes sub-pixel seams, not whole missing cells (a missing cell at zoom = a large hole).

**Shared by both cell-rebaking paths** — `DETACHED_REVOXELIZE` and attached `GRID` (`REBUILD_GRID_VOXELS`) use the same model. GRID's 12^3 cubes hide it; small cubes don't. Fix both via the shared model.

## Fix direction (architect — instrument-first; this surface has been mis-diagnosed before, cf. #1457)
1. **Instrument:** dump the filled cell set vs the rotated-AABB lattice; classify holes surface-vs-interior; confirm hole count scales with rotation angle and inversely with solid size.
2. **Lead — inverse (backward) resampling:** dispatch over *destination* cells in the rotated AABB; fill cell `c` iff `roundHalfUp(rotateByQuat(c, conj(quat)))` is in source occupancy. Surjective -> hole-free by construction. Needs a source-occupancy lookup (hash/bitset of the pool), a matching **CPU mirror** (preserve the current CPU<->GPU bit-identity), and a dest-cell dispatch domain.
3. **Alternative if inverse is too costly:** guaranteed 3D morphological dilation of the forward-scattered set (close 1-cell gaps in **cell** space, not screen space) — cheaper but approximate. Architect picks after instrumenting.

## Constraints
- CPU<->GPU **bit-identity** (the fill has a CPU mirror; CPU/GLSL/Metal must classify cells identically).
- GLSL<->Metal **parity in lockstep**.
- Identity-rotation fast-path **byte-identical** (cardinal scenes untouched).
- Cover **both** `DETACHED_REVOXELIZE` and `GRID` re-voxelize.

## Definition of done
- A small (<=4^3) and a thin (1-2 cell) `DETACHED_REVOXELIZE` solid show **no missing faces / holes** across a full spin in `canvas_stress`; same for a small GRID re-voxelize solid.
- Re-enable a thin shape (frame / cross) on the re-voxelize path in `canvas_stress` as the standing regression guard (the demo currently routes them away to dodge this).
- GLSL + Metal parity; render-verify ROI crops committed.

**Model:** opus
**Area:** engine/render (re-voxelize fill), engine/prefabs/irreden/render (GRID re-voxelize)
**Part of:** correctness follow-up to the #1553 re-voxelize epic (coverage; distinct from the P4b lighting track).

### LINKED ISSUE #1720: render: finish #1619 GRID half — REBUILD_GRID_VOXELS still forward-maps, checkered coverage holes on rotating GRID solids
**Follow-up to:** #1619 (closed — the DETACHED_REVOXELIZE half landed as GPU inverse resample; the GRID half in its scope/DoD did not land)
**Area:** engine/prefabs/irreden/voxel (REBUILD_GRID_VOXELS, grid_rotation.hpp)
**Model:** fable

## Symptom

`IRCanvasStress --no-lighting --auto-screenshot`: the GRID spin cubes (12³) show checkered single-cell holes on top faces and ragged silhouettes at off-cardinal poses. This is the residual geometry-level coverage defect after the #1718 lighting banding is accounted for.

## Root cause

`REBUILD_GRID_VOXELS` still maps one authored voxel → `roundVec3HalfUp(R·p)` (forward scatter, `IRPrefab::GridRotation::worldCellForGridVoxel`). Forward rotation of a lattice is not surjective onto the covered dest cells → holes. #1619 fixed exactly this for detached re-voxelize via dest-lattice inverse resampling; GRID was in #1619's constraints ("Cover both") and DoD but the landed fix is detached-only (see voxel/CLAUDE.md REBUILD_GRID_VOXELS note: "aliasing accepted by design" still stands for GRID).

## Design constraint that blocked the mechanical port

A GRID set's world-pool span is allocated at `numVoxels_`; inverse resampling over the rotated AABB can produce more occupied dest cells than source voxels (surface effects), so the span contract needs a decision:
- (a) allocate rotating GRID sets at the rotated-AABB bound (≈(√3·n)³ worst case — large), or
- (b) allocate at `ceil(k·numVoxels_)` with deterministic overflow drop + instrumentation, or
- (c) hybrid: forward scatter + claim-unused-slots hole-fill pass (bounded by span size, closes the holes that matter — surface cells — first).

Architect to pick after instrumenting actual dest-cell counts across a spin (the overshoot is small in practice for solid boxes; thin shapes are the stress case).

## Definition of done

- A 12³ GRID spin cube shows no missing-cell holes across a full spin in canvas_stress (`--no-lighting` capture).
- CPU↔GPU↔backend classification identity preserved (`roundHalfUp` convention).
- Re-enable a thin shape on the GRID path in the orbit ring as the standing regression guard (mirrors #1619's DoD).

jakildev:
## Summary

Closes #1720's GRID half of the #1619 coverage-hole defect: `REBUILD_GRID_VOXELS` forward-mapped each authored voxel through `roundHalfUp(R·p)`, which is not surjective onto the covered world cells — mid-rotation the GRID spin cubes lost entire rows/strips (up to ~29% of a solid 12³), with the floor visible through the holes.

- **Inverse resample (CPU twin of #1619's GPU fix):** rotating GRID sets now walk the rotated source AABB's integer world cells and inverse-map each via `roundHalfUp(R⁻¹·c)` (`IRPrefab::GridRotation::sourceCellForWorldCell`) against a per-set source occupancy grid, authoring position + color + face-occlusion bits per covered dest cell. Same `roundHalfUp` convention as the detached GPU kernel, so CPU GRID and GPU detached classify identically.
- **Span contract (the #1720 design decision):** span stays at `numVoxels_`. Measured covered-cell overshoot is a boundary fluctuation (≤ 2.8% observed on a solid 12³); on overflow **interior** cells (all 6 neighbors covered) drop first in deterministic walk order, so the visible surface always renders. `dropHighWater_` warn-logs each new per-run maximum. Degenerate exact-fit thin allocations (bare n×n×1 plates) are documented as "allocate the full box and carve".
- **Authored-truth snapshot:** while a set rotates, its pool span is a derived arrangement; authored colors live in `C_VoxelSetNew::rotationSourceVoxels_` (lazy snapshot on first rotated frame, mirrored by the component's color mutators, restored + cleared by the identity arm). Zero-scale / pathological dest volumes (> 2M cells) fall back to the pre-#1720 forward map for the frame.
- **canvas_stress:** the orbit-ring cross moves to the GRID path as the standing thin-geometry regression guard (#1720 DoD); new `--auto-profile` flag writes the World-dtor per-system timing report (the rotating-set CPU sensor perf_grid can't provide).

## Visual evidence (sweep captures, `--only gridspin,floor --no-lighting --sweep-frames 8 90`)

Mid-rotation frames — before shows the row/strip holes, after is hole-free:

| frame | before (master) | after (this PR) |
|---|---|---|
| 1 | before1 | after1 |
| 3 | before3 | after3 |
| 6 | before6 | after6 |

All 8 sweep phases committed under `docs/pr-screenshots/claude/1720-grid-inverse-resample/`. Independently re-verified against the final tree (fresh build + re-capture): hole-free at every phase. Full default suite (all spawn groups) also run clean — orbit cross renders fully formed on GRID, FRAME wireframes stay crisp on forward-scatter DETACHED.

## Perf

`RebuildGridVoxels` with the inverse walk: **avg 0.287 ms / max 0.405 ms** over 507 frames of the full default scene (8 rotating GRID sets) — ~1.7% of a 60 fps budget, max/avg ≈ 1.4 (no realloc spikes; scratch is member-owned with reused capacity). For scale: sibling `RebuildDetachedVoxels` 0.771 ms, `SingleVoxelToCanvasFirst` 2.542 ms. Optimize smell-scan clean.

## Test plan

- [x] `fleet-build --target IRCanvasStress` clean (macOS / Metal host)
- [x] Sweep capture `--only gridspin,floor --no-lighting --sweep-frames 8 90`: hole-free at all 8 spin phases (DoD criterion 1)
- [x] `roundHalfUp` CPU↔GPU classification convention preserved (DoD criterion 2)
- [x] Orbit cross re-enabled on GRID path as thin-geometry guard (DoD criterion 3)
- [x] Full default suite: span-cap drops interior-only (high-water 48/1740, surface always renders), no errors
- [x] `--auto-profile` per-system report: RebuildGridVoxels 0.287 ms avg

Closes #1720


jakildev (commented):
## Review — render: GRID inverse re-voxelize (#1720)

**Verdict:** approve

### Praise
- The three-arm dispatch (`identityArm` / `inverseArm` / `forwardArm`) with explicit fallback conditions is clean. The pre-condition for each arm (identity, degenerate scale, volume cap) is unambiguous, and the degenerateScale + `!inverseArm()` fallback correctly handles both zero-scale and pathological-volume cases without adding a third top-level branch.
- The `rotationSourceVoxels_` lifecycle is carefully designed: lazy snapshot on first rotated frame, mirror-on-mutate in all color mutators, restore+clear in the identity arm. No separate dirty flag needed — non-emptiness IS the "span is in re-voxelized arrangement" state. The component header comment explains this contract clearly.
- `kMaxDestWalkCells = 2,000,000` guard prevents the scratch vectors from growing past 8 MB for a pathological scale, and the `dropHighWater_` warn-log makes span-cap drops observable without flooding the log.
- The interior-first drop policy (surface cells win the span-cap contest) is the right choice: `surfaceCells_.size()` ≪ `safeCount` for any solid, so the visible surface always renders. The donut-style note in the header about thin exact-fit plates is a useful gotcha for future authors.
- `sourceCellForWorldCell` correctly inverts the TRS order: subtract translation → apply inverse rotation → divide by scale — the arithmetic matches `worldCellForGridVoxel`'s forward path.

### ECS invariants — checked
- No `getComponent` inside per-entity tick. ✓
- All scratch vectors (`sourceSlotGrid_`, `destSrcSlot_`, `destCells_`, `surfaceCells_`, `interiorCells_`, `cellKeys_`, `cellRoundedPositions_`) are `System<N>` members — capacity retained across frames. ✓
- No structural ECS changes (addComponent/removeComponent/removeEntity) inside the tick. ✓
- `REBUILD_GRID_VOXELS` already exists in the `SystemName` enum (existing system). ✓

### Ownership / lifetime — checked
- `pool.getColors()`, `pool.getPositionGlobals()`, `pool.getPositions()`, `pool.getPositionOffsets()` return `std::vector&` references. The arms call pool methods (`resyncActiveMaskFromColors`, `queuePositionRange`, `markChunkWorldBoundsDirty`) which write *into* those same vectors without reallocation — no reference invalidation. ✓
- `rotationSourceVoxels_` is component-owned, not a reference into pool storage. ✓

### Render pipeline — checked
- No shaders changed, no CPU/GPU struct layout drift. ✓
- No new `.glsl` / `.metal` files. ✓
- No binding index changes. ✓

### Math / coordinates — checked
- All uses are `IRMath::roundVec3HalfUp`, `IRMath::min/max/floor/ceil`, `IRMath::rotateVectorByQuat`, `IRMath::quatInverse`. No raw `glm::*` or `std::` math calls. ✓
- The `roundHalfUp(R⁻¹·c)` inverse convention matches the detached re-voxelize GPU kernel (#1619) — CPU GRID and GPU detached classify identically. ✓

### Naming / style — checked
- All new members have trailing `_` (private-by-convention). ✓
- `mirrorToRotationSource`, `identityArm`, `forwardArm`, `inverseArm` are consistent with the codebase style. ✓

### Serialization — not applicable (no serialized structs changed).

### Mirror correctness for `deactivateAll` / `activateAll`
`mirrorToRotationSource(i)` runs BEFORE `recomputeFaceOccupancy(voxels_, size_)` in both mutators, so the snapshot captures stale face flags for activated/deactivated voxels. This is harmless: the inverse arm overrides face bits from the dest-grid adjacency anyway (the face-flag block at `writeCell`), and deactivated voxels (alpha=0) are skipped entirely in the source AABB scan. ✓

### Nits
- `engine/prefabs/irreden/voxel/grid_rotation.hpp` — the file comment already says "lives outside the system header so unit tests can exercise the math without setting up an EntityManager + canvas + voxel pool." A round-trip unit test for `sourceCellForWorldCell` (e.g. pick a known SQT, a known authored cell, compute forward → world, then inverse → should recover the authored cell within `roundHalfUp` tolerance) would cement the `roundHalfUp` CPU↔GPU convention claim as a regression check. Currently the only verification is the visual sweep.

- [x] `fleet-build --target IRCanvasStress` (macOS Metal) — confirmed in PR body
- [x] Sweep capture `--only gridspin,floor --no-lighting --sweep-frames 8 90`: hole-free at all 8 phases — confirmed in PR body and screenshot evidence
- [x] Full default suite: span-cap drops interior-only (high-water 48/1740), no errors — confirmed
- [ ] (follow-up nit) Unit test for `sourceCellForWorldCell` round-trip

Opus recheck required: touches GPU-facing pool state (`poolColors`, `poolGlobals`, active-mask) across three new code paths with `queuePositionRange` + `resyncActiveMaskFromColors` calls — please verify frame-ordering invariants between the UPDATE pipeline writes and `VOXEL_TO_TRIXEL_STAGE_1`'s GPU upload step (active-mask buffer lifetime across frames, no race between `resyncActiveMaskFromColors` and the active-mask GPU upload).

jakildev (commented):
## Review (Opus recheck) — render: GRID inverse re-voxelize (#1720)

**Verdict:** approve

Opus second pass on the items Sonnet escalated. Sonnet's first-pass checklist
(naming, no-dirty-flag discipline, span-cap policy, scratch-capacity reuse) I
take as confirmed; I spent this pass on the GPU-handoff invariant Sonnet could
not verify.

### Sonnet's escalation — frame-ordering / active-mask lifetime: confirmed sound

Sonnet flagged: *"verify frame-ordering invariants between the UPDATE pipeline
writes and `VOXEL_TO_TRIXEL_STAGE_1`'s GPU upload … no race between
`resyncActiveMaskFromColors` and the active-mask GPU upload."* I traced the full
handoff and **confirm it is correct**:

- The GRID path writes the **main world canvas** pool, which is never the
  `revoxInverse` (GPU detached) canvas. So in `system_voxel_to_trixel.hpp:616`
  the `!revoxInverse` branch is taken: colors (`:618`) and the active-mask
  whole-word window (`:627`) are **fully re-uploaded every frame** from the CPU
  mirror. Positions drain through `flushPendingPositionRanges` (`:609`).
- UPDATE runs to completion before STAGE_1 (render pipeline) in the same frame,
  single-threaded — ECS multithreading is still design-only (epic #226). So
  `resyncActiveMaskFromColors` (CPU bitset edit) strictly precedes the
  full-window `subData`; there is no concurrent reader and the whole live window
  is re-sent, so no partial/stale-mask hazard. No async readback in this path,
  no SSBO read on frame N+1 without a re-upload.
- The position queue records *(start,count)*, not a data snapshot — the flush
  reads live `poolGlobals` at flush time. The inverse arm queues exactly
  `[baseIdx, written)` (`:519-521`), which covers every active slot it authored,
  so it is self-sufficient regardless of whether `UPDATE_VOXEL_SET_CHILDREN`
  queued the set. This is **strictly more correct than master**, whose
  `REBUILD_GRID_VOXELS` never queued and relied on the child system's
  translation-triggered queue — an in-place-rotating GRID set (no translation
  delta → `updateAsChild` returns 0 → no child queue) would have shipped stale
  positions on master; the explicit queue here fixes that.

### Other deep items I checked (Opus-only)

- **Snapshot ↔ mutator consistency.** `voxels_` is a `std::span` *into* the pool
  color array at `voxelStartIdx_` (`component_voxel_set.hpp:61`), so
  `voxels_[i]` and `poolColors[baseIdx+i]` are the same memory. The lazy
  snapshot (`inverseArm` `:331`) captures the authored colors *before* the first
  re-voxelize overwrites the span, and `mirrorToRotationSource` keeps mid-spin
  mutations landing in the authored truth — a transient clobber of the derived
  span at the model-space index is re-derived the next frame. Correct, and the
  cull gate means the snapshot is only ever taken on a frame the set is actually
  re-rasterized, so a set spawned-rotated-but-offscreen still snapshots authored
  colors. No double-source-of-truth bug.
- **Inactive-tail staleness.** The inverse arm deactivates `[written, safeCount)`
  but doesn't rewrite their positions. The active-mask clears them so they don't
  render, and the chunk-world-bounds recompute skips `alpha_==0`
  (`component_voxel_pool.hpp` `for i … if alpha_==0 continue`), so the stale tail
  positions neither render nor inflate the cull AABB. Safe.
- **Bounds / OOB in the dest walk.** Every neighbor probe in the interior-classify
  and `writeCell` face-occlusion derivation is guarded (`x>0 && x+1<destDims.x`
  …) before indexing `destSrcSlot_[lin±1 / ±strideY / ±strideZ]`. `kMaxDestWalkCells`
  caps `destSrcSlot_`/`sourceSlotGrid_` at a bounded ~8 MB worst case; both reuse
  capacity. No unbounded growth.
- **Identity arm = master.** The non-restored identity path is byte-identical to
  master (writes globals, recomputes mask, no queue); the restore branch adds the
  queue + `resyncActiveMaskFromColors` only on the rotated→identity transition,
  which is exactly the frame that needs it.

### Nits (optional — do not block merge)

- `grid_rotation.hpp` `sourceCellForWorldCell` has no round-trip unit test, only
  the visual sweep + `dropHighWater_` log. The helper lives outside the system
  precisely so it's testable headless — a `sourceCellForWorldCell ∘ worldCellForGridVoxel`
  round-trip assertion would lock the CPU↔GPU rounding convention against future
  drift. (Same follow-up Sonnet noted.)
- `identityArm` calls `rotationSourceVoxels_.shrink_to_fit()` on every
  rotated→identity transition, so a set that repeatedly starts/stops spinning
  re-allocates the snapshot each cycle. Only at transitions (not per frame), so
  it's within the hot-path-allocation rule — `clear()` without the
  `shrink_to_fit` would trade idle memory for less churn. Defensible either way.

### Praise

- Driving the "is the span in re-voxelized arrangement?" state off
  `rotationSourceVoxels_` non-emptiness instead of a `bool` is the right read of
  the no-dirty-flags rule, and the span↔pool aliasing makes the snapshot/restore
  genuinely cheap.
- Walking the dest lattice and inverse-mapping (surjective) rather than
  forward-scattering is the correct fix for the coverage holes, and mirroring the
  detached `#1619` `roundHalfUp(R⁻¹·c)` convention keeps CPU GRID and GPU detached
  classifying identically.

### Test plan before merge
- [x] `fleet-build --target IRCanvasStress` (macOS) — confirmed in PR body
- [x] `--only gridspin,floor --sweep-frames 8 90` hole-free — confirmed
- [ ] (nit) headless round-trip unit test for `sourceCellForWorldCell`

🤖 Reviewed by Claude Opus 4.8 (review-pr skill)

jakildev:
Addressed nit: added three  round-trip tests to `test/ecs/grid_rotation_test.cpp`:
- **SourceCellForWorldCell_RoundTrip_NinetyDegreeZRotation** — 6 authored cells at 90° Z rotation (exact, no rounding boundary ambiguity); verifies inverse arithmetic.
- **SourceCellForWorldCell_RoundTrip_ScaleAndTranslation** — full SQT (2× scale + 90° rotation + translation); verifies all three components un-apply correctly.
- **SourceCellForWorldCell_RoundHalfUpConventionPin** — constructs the case where inverse-mapped source y = -0.5 exactly; asserts `roundHalfUp(-0.5) = 0` (not -1 as glm::round/std::round would give), pinning the CPU↔GPU convention.

All 13 GridRotationTest cases pass. The `shrink_to_fit` → `clear()` nit was omitted (Opus marked it "defensible either way").

jakildev:
Addressed nit: added three `sourceCellForWorldCell` round-trip tests to `test/ecs/grid_rotation_test.cpp`:

- **SourceCellForWorldCell_RoundTrip_NinetyDegreeZRotation** — 6 authored cells at 90° Z rotation (exact, no rounding boundary ambiguity); verifies inverse arithmetic.
- **SourceCellForWorldCell_RoundTrip_ScaleAndTranslation** — full SQT (2× scale + 90° rotation + translation); verifies all three components un-apply correctly.
- **SourceCellForWorldCell_RoundHalfUpConventionPin** — constructs the case where inverse-mapped source y = -0.5 exactly; asserts `roundHalfUp(-0.5) = 0` (not -1 as glm::round/std::round would give), pinning the CPU↔GPU convention.

All 13 GridRotationTest cases pass (10 existing + 3 new). The `shrink_to_fit` → `clear()` nit was omitted (Opus marked it "defensible either way").

PR #1736, opened 2026-06-12
render: per-axis scatter screen scale = full camera zoom (post-#1458)

jakildev:
## Summary
- Fixes the rotated-camera zoom regression: after #1458 moved the per-axis smooth-yaw store to BASE (world-unit) resolution, the framebuffer forward-scatter kept its pre-#1458 `effSub/cappedDensity` model-matrix rescale. Against the base-resolution store the two factors cancelled camera zoom entirely — zooming at any non-cardinal yaw shrank the cull viewport (a crop) without magnifying content.
- The scatter screen scale is now the full camera zoom (`canvasZoomLevel_ * effSub`). The integer anchor pan (`perAxisBase_`) must stay at the store's capped-density scale for bit-consistent face-origin recovery, so its excess screen-pan is cancelled by a translate term that is exactly zero when the density is 1.
- Cardinal rendering is unaffected — the scatter only runs at a non-cardinal residual yaw, and all changes are inside `drawPerAxisScatter`.
- perf_grid gains rotated-zoom (`zoom1_rot`/`zoom4_rot`) and rotated-pan (`zoom4_pan`/`zoom4_rot_pan`) regression shots.

## Test plan
- [x] IRPerfGrid `--auto-screenshot`: zoom4_rot now renders content 4x larger than zoom1_rot (before: identical size, cropped viewport — see pairs below)
- [x] zoom1_rot and zoom1_origin pairs unchanged (density-1 control)
- [x] Rotated-pan pair: same world content at screen center for cardinal vs rotated at cam=(16,8) zoom 4
- [x] IRShapeDebug 21-shot battery clean on macOS/Metal; `zoom4_yaw45_inter_cardinal` voxel twins now match their SDF counterparts' scale; camera-focus pivot series (#1352 shots) holds the focused content at screen center
- [x] `fleet-build --target IRPerfGrid` + `IRShapeDebug` green

## Screenshots

<details>
<summary>zoom4_rot — the fix (before: cropped, not magnified)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom1_rot — rotated control (should be unchanged)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom1_origin — cardinal control (should be unchanged)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>rotated-pan invariance (after only — shots added with the fix)</summary>

[...]

PR #1737, opened 2026-06-12
render: rotated-zoom perf — single-slice dispatch, coverage clamp, scatter compaction

jakildev:
Stacked on: https://github.com/jakildev/IrredenEngine/pull/1736

## Summary
Three fixes on the smooth-yaw path, found by adding a rotation axis (`--yaws`) to the perf matrix. IRPerfGrid's worst cell (zoom 8, FULL, yaw 0.35) drops **137.3 ms → 11.7 ms (-91.5%)**; all 24 matrix cells improve.

1. **Single-slice per-axis dispatch.** The per-axis store reused the compact pass's indirect command (`numGroupsZ = subdivisions²`) but only writes at z = 0 post-#1458 — every other slice launched and early-returned, 3 axes × 2 stages per frame. The compact pass now authors a second indirect command in the same `VoxelIndirectDispatchParams` buffer (same XY, z = 1; byte offset 20) for passes that walk the compacted list once per voxel.
2. **Canvas-coverage cull clamp.** A subdivided raster writes taps at `offset + isoPos·scale`, so the canvas only holds iso positions within `canvasSize/(2·scale)` of the view center — the shadow-feeder sweep beyond that had every tap clip-rejected while still paying the sub² raster. Cardinal cull viewports now intersect the canvas coverage; rotating + detached canvases keep their bounds untouched. Cardinal zoom 8 FULL: 48.5 → 26 ms, cull ratio 0.47 → 0.23.
3. **Per-axis scatter compaction** (the design doc's scoped follow-up). The scatter instanced over every cell of the worst-case (2W, W+H) canvas — a flat ~6 ms/frame at 1280×720 regardless of content. New `c_compact_scatter_cells.{glsl,metal}` appends non-empty cell indices + authors the 5-uint indirect draw command (GL `DrawElementsIndirectCommand` ≡ Metal `MTLDrawIndexedPrimitivesIndirectArguments`), drawn via the new `RenderDevice::drawElementsIndirect` (both backends).

Supporting: Metal sticky texture/image binding tables now purge destroyed textures (`unbindMetalTexture` in the texture destructors) — fixes a use-after-free segfault at the rotated→cardinal transition where the released per-axis texture stayed sticky-bound and the next compute encoder re-encoded it. The five Metal draw methods now share `beginConfiguredRenderEncoder` (simplify reuse finding). `perf_grid --yaw` + `perf_grid_matrix.sh --yaws` make rotation a first-class matrix axis. Optimize-skill bottleneck catalog + big-wins updated; design-doc compaction follow-up marked landed.

## frame timing (ms)

| cell | avg | p99 |
|------|-----|-----|
| `target=IRPerfGrid,zoom=1,sub_mode=full,sub_base=1` | 9.55 → 7.79 (-18.4%) ↓ | 21.82 → 13.22 (-39.4%) ↓ |
| `target=IRPerfGrid,zoom=1,sub_mode=full,sub_base=1,yaw=0.35` | 21.83 → 13.50 (-38.2%) ↓ | 31.63 → 20.33 (-35.7%) ↓ |
| `target=IRPerfGrid,zoom=1,sub_mode=none,sub_base=1` | 8.67 → 7.51 (-13.4%) ↓ | 20.17 → 13.74 (-31.9%) ↓ |
| `target=IRPerfGrid,zoom=1,sub_mode=none,sub_base=1,yaw=0.35` | 22.54 → 13.83 (-38.6%) ↓ | 33.08 → 23.22 (-29.8%) ↓ |
| `target=IRPerfGrid,zoom=1,sub_mode=position_only,sub_base=1` | 9.15 → 7.56 (-17.4%) ↓ | 16.87 → 13.60 (-19.4%) ↓ |
| `target=IRPerfGrid,zoom=1,sub_mode=position_only,sub_base=1,yaw=0.35` | 22.08 → 13.79 (-37.5%) ↓ | 33.72 → 22.47 (-33.4%) ↓ |
| `target=IRPerfGrid,zoom=2,sub_mode=full,sub_base=1` | 12.73 → 11.24 (-11.7%) ↓ | 18.82 → 15.10 (-19.8%) ↓ |
| `target=IRPerfGrid,zoom=2,sub_mode=full,sub_base=1,yaw=0.35` | 33.81 → 13.83 (-59.1%) ↓ | 46.99 → 22.34 (-52.5%) ↓ |
| `target=IRPerfGrid,zoom=2,sub_mode=none,sub_base=1` | 8.67 → 7.49 (-13.6%) ↓ | 13.18 → 14.06 (+6.7%) |
| `target=IRPerfGrid,zoom=2,sub_mode=none,sub_base=1,yaw=0.35` | 22.60 → 14.05 (-37.8%) ↓ | 36.63 → 22.07 (-39.7%) ↓ |
| `target=IRPerfGrid,zoom=2,sub_mode=position_only,sub_base=1` | 8.77 → 7.73 (-11.9%) ↓ | 17.35 → 13.65 (-21.3%) ↓ |
| `target=IRPerfGrid,zoom=2,sub_mode=position_only,sub_base=1,yaw=0.35` | 22.44 → 13.91 (-38.0%) ↓ | 33.38 → 23.16 (-30.6%) ↓ |
| `target=IRPerfGrid,zoom=4,sub_mode=full,sub_base=1` | 24.91 → 18.88 (-24.2%) ↓ | 33.12 → 23.71 (-28.4%) ↓ |
| `target=IRPerfGrid,zoom=4,sub_mode=full,sub_base=1,yaw=0.35` | 67.76 → 13.38 (-80.3%) ↓ | 75.19 → 21.72 (-71.1%) ↓ |
| `target=IRPerfGrid,zoom=4,sub_mode=none,sub_base=1` | 8.66 → 7.54 (-12.9%) ↓ | 18.26 → 13.19 (-27.8%) ↓ |
| `target=IRPerfGrid,zoom=4,sub_mode=none,sub_base=1,yaw=0.35` | 21.67 → 13.44 (-38.0%) ↓ | 30.96 → 21.35 (-31.0%) ↓ |
| `target=IRPerfGrid,zoom=4,sub_mode=position_only,sub_base=1` | 8.78 → 7.95 (-9.5%) ↓ | 16.15 → 13.27 (-17.8%) ↓ |
| `target=IRPerfGrid,zoom=4,sub_mode=position_only,sub_base=1,yaw=0.35` | 21.38 → 13.35 (-37.6%) ↓ | 28.98 → 19.09 (-34.1%) ↓ |
| `target=IRPerfGrid,zoom=8,sub_mode=full,sub_base=1` | 48.55 → 26.07 (-46.3%) ↓ | 56.75 → 29.04 (-48.8%) ↓ |
| `target=IRPerfGrid,zoom=8,sub_mode=full,sub_base=1,yaw=0.35` | 137.28 → 11.67 (-91.5%) ↓ | 143.00 → 21.36 (-85.1%) ↓ |
| `target=IRPerfGrid,zoom=8,sub_mode=none,sub_base=1` | 8.64 → 7.29 (-15.6%) ↓ | 17.88 → 13.25 (-25.9%) ↓ |
| `target=IRPerfGrid,zoom=8,sub_mode=none,sub_base=1,yaw=0.35` | 20.48 → 11.94 (-41.7%) ↓ | 28.17 → 19.79 (-29.7%) ↓ |
| `target=IRPerfGrid,zoom=8,sub_mode=position_only,sub_base=1` | 8.66 → 7.26 (-16.2%) ↓ | 23.34 → 13.11 (-43.8%) ↓ |
| `target=IRPerfGrid,zoom=8,sub_mode=position_only,sub_base=1,yaw=0.35` | 20.50 → 11.58 (-43.5%) ↓ | 29.60 → 19.86 (-32.9%) ↓ |

## GPU stage timing (avg ms per frame)

Remaining gaps to the 100 fps goal are filed as follow-up issues (rotated floor ≈ 13.5 ms: per-axis lighting passes + store dispatch; cardinal FULL zoom 8 ≈ 26 ms: feeder depth-only path / sub² raster; UPDATE pipeline ≈ 5 ms at 262k entities).

## Test plan
- [x] 24-cell perf matrix baseline vs head (`compare_perf_runs.py`): all cells improved, none regressed
- [x] IRShapeDebug 21-shot battery: cardinal shots **byte-identical** (img_diff drift=0 on zoom1/4/8/16 + LOD), rotated shots 0–12 px @ Δ≤5 (depth-tie order noise from compacted instance order)
- [x] IRPerfGrid 7-shot battery incl. rotated-zoom + rotated-pan regression shots: matches the parent PR's references
- [x] Rotated→cardinal transition exercised (previously segfaulted; now clean across the full shot sequence)
- [x] `fleet-build` IRPerfGrid + IRShapeDebug green on macOS/Metal

## Screenshots

<details>
<summary>zoom4_rot — before/after (visually identical; HUD FPS 26 → ~75)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom1_rot — before/after (visually identical)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>


jakildev (commented):
## Review — render: rotated-zoom perf — single-slice dispatch, coverage clamp, scatter compaction (#1737)

**Verdict:** approve (Opus recheck required — see below)

Stacked on #1736 (`fix/perfgrid-zoom-after-rotation`, now `fleet:approved`).

### Praise
- Three focused fixes, each with a clear measured root cause, a precise correctness argument, and byte-identical or near-byte-identical confirmation on all pre-existing shots. The 137 ms → 11.7 ms (-91.5%) drop on the worst-case rotated+zoom8+FULL cell is exceptional.
- The SSBO slot alias approach (25/26) is clearly documented at every call site, including the re-assertion comment after the axis loop, so the invariant is visible to future readers without archaeology.
- `beginConfiguredRenderEncoder` is a clean refactor — five draw entry points from copy-paste to a shared helper, with the `needsIndexBuffer` flag making the divergence explicit. The old `drawElementsInstanced` still guards `instanceCount <= 0` before entering the helper. ✓
- `unbindMetalTexture` in both texture destructors is the right fix for the sticky-bind use-after-free. Placing it in the destructor means no call-site burden on canvas teardown paths. ✓
- New shader pair (`c_compact_scatter_cells.{glsl,metal}`) correctly mirrors the 5-uint indirect layout between GL `DrawElementsIndirectCommand` and Metal `MTLDrawIndexedPrimitivesIndirectArguments`, and the `kSlotInstanceCount = 1` constant in the Metal version documents the field offset explicitly.
- `threadgroupSizeForFunctionName` updated for `c_compact_scatter_cells` — the silent 1×1×1 dispatch trap (Metal threadgroup-size map rule) is handled. ✓

### ECS invariants — checked
- No `getComponent` in the tick path. `drawPerAxisScatter` takes all inputs as parameters; compaction buffers and shader programs are resolved from `IRRender::getNamedResource` at `endTick`. ✓
- No cross-frame component storage introduced. ✓
- `clampToCanvasCoverage` and `rasterScale` are stack-local, re-derived every `dispatchPerEntityBatch` call. ✓

### Ownership / lifetime — checked
- `scatterCellsBuf_` and `scatterDrawArgsBuf_` are non-owning raw pointers (engine convention). Named resources own the backing `Buffer` objects. ✓
- `scatterDrawArgsBuf_->subData()` CPU-seeds all 5 fields each axis pass before the compact compute runs; `instanceCount` starts at 0 and the shader only atomically adds to field 1. ✓
- `drawArgsTemplate` is a stack array — no heap allocation in the axis loop. ✓

### Render pipeline — checked
- `VoxelIndirectDispatchParams._padding[3]` → `singleSliceNumGroupsX/Y/Z`: the struct size is unchanged (8 × 4 = 32 bytes), and `offsetof(VoxelIndirectDispatchParams, singleSliceNumGroupsX) = 5 × 4 = 20`. The `kSingleSliceOffset` constexpr matches. ✓
- GLSL `c_voxel_visibility_compact.glsl` writes `singleSliceNumGroupsZ = 1u` (hardcoded), while `numGroupsZ` still gets `subdivisions²`. Metal slots 5/6/7 are also written. GL and Metal compact shader outputs are symmetric. ✓
- Barrier ordering: `SHADER_STORAGE` (for `scatterCells` SSBO) then `COMMAND` (for `scatterDrawArgs` indirect command buffer) — correct sequence before `drawElementsIndirect`. ✓
- Shader naming: `c_compact_scatter_cells.{glsl,metal}` uses the `c_` compute prefix. ✓
- `v_peraxis_scatter.glsl` reads `scatterCells[gl_InstanceID]` after the COMMAND barrier — the compaction happens-before guarantee is met via the barrier. ✓
- `kFileCompCompactScatterCells` added to `shader_names.hpp`; Metal counterpart auto-discovered by the Metal library loader. ✓
- `drawElementsIndirect` in GL binds + unbinds `GL_DRAW_INDIRECT_BUFFER` around the draw — correct usage for the indirect draw API. ✓
- Metal `drawIndexedPrimitives` with `indirectBuffer` signature matches `MTLDrawIndexedPrimitivesIndirectArguments`: {indexCount, instanceCount, indexStart, baseVertex, baseInstance} ≡ the 5-uint CPU-seeded layout. ✓

### Math / coordinates — checked
- `IsoBounds2D::intersect` in `ir_math.hpp` uses `glm::max` / `glm::min` inside the math library itself — allowlisted. ✓
- `clampToCanvasCoverage` uses `IRMath::max`, `IRMath::visibleIsoViewport`, `IRMath::trixelOriginOffsetZ1`. No raw `glm::` or `std::` outside `engine/math/`. ✓

### Naming / style — checked
- All new public members and locals follow conventions: `scatterCellsBuf_`, `scatterDrawArgsBuf_`, `scatterCompactProgram_` use trailing `_`. ✓
- `kCoverageSlackPx`, `kSingleSliceOffset`, `kSlotInstanceCount` use the `k` prefix for constants. ✓
- `beginConfiguredRenderEncoder` is descriptive and matches the existing naming style in `metal_render_impl.cpp`. ✓

### Tests / build — checked
- 24-cell perf matrix: all cells improved, confirmed in PR body. ✓
- IRShapeDebug 21-shot battery: cardinal shots byte-identical, rotated shots 0–12 px drift (compacted instance order → depth-tie noise on equal-depth cells). The drift is expected and documented; cardinal shots are the stability gate. ✓
- Rotated→cardinal transition exercised and clean (previously segfaulted). ✓
- `fleet-build` IRPerfGrid + IRShapeDebug green on macOS/Metal. ✓

- [x] 24-cell perf matrix: all cells improved — confirmed in PR body
- [x] IRShapeDebug 21-shot cardinal shots byte-identical — confirmed
- [x] Rotated→cardinal transition exercised and clean — confirmed
- [x] `fleet-build` IRPerfGrid + IRShapeDebug green — confirmed (macOS/Metal)
- [ ] Linux/OpenGL smoke (tagged separately below)
- [ ] Windows smoke (tagged separately below)

Opus recheck required — three items touching correctness in the render/GPU pipeline:

1. **SSBO slot aliasing (25/26):** `drawPerAxisScatter` temporarily re-binds slots 25/26 to the scatter compaction buffers for each axis pass, then re-asserts the canonical `CompactedVoxelIndices` / `IndirectDispatchParams` owners after the loop. The invariant is that no other system or command issued between the `TRIXEL_TO_FRAMEBUFFER` tick and the post-loop restore reads these slots expecting the canonical owners. Opus should trace the full SYSTEM_TRIXEL_TO_FRAMEBUFFER execution order relative to systems that depend on slots 25/26 (stage-1 / stage-2 / compact dispatch) to confirm the frame schedule decouples these.

2. **Coverage cull shadow correctness:** The PR claims intersecting `chunkVp` and `gpuVp` with canvas coverage is "byte-identical" for shadow feeders — the argument is that any voxel beyond `canvasSize/(2·rasterScale)` of the view center has every tap clip-rejected and therefore contributes no writes to `trixelDistances`. Opus should confirm this write-clip relationship holds at all rasterScale values and that the `kCoverageSlackPx = 4` margin is sufficient to prevent edge-voxel shadow bleeding at the boundary.

3. **`PerAxisScatterCells` buffer sizing invariant:** `scatterCellsBuf_` is sized once at `endTick` using `perAxisTrixelCanvasWorstCaseSize` of the main canvas at that moment. Opus should confirm this function returns a stable worst-case bound over all possible canvas sizes (not just current), or verify that canvas resize invalidates and reallocates the resource.

jakildev (commented):
## Opus recheck — render: rotated-zoom perf — single-slice dispatch, coverage clamp, scatter compaction (#1737)

**Verdict:** approve

Sonnet's first pass was thorough — the three root-cause analyses, the GL/Metal symmetry checks, the barrier-ordering walk, and the ECS/ownership/naming checklist all hold. This Opus pass focused on the three invariants Sonnet explicitly escalated (slot aliasing, coverage-cull shadow correctness, scatter-buffer sizing), plus one cross-backend parity check I wanted to do independently. **All three escalations resolve; I confirm Sonnet's approve.**

---

### Escalation 1 — SSBO slot aliasing (25/26): confirmed sound

`drawPerAxisScatter` runs in `TRIXEL_TO_FRAMEBUFFER`'s `endTick` (the framebuffer composite, after all `VOXEL_TO_TRIXEL` work for the frame is finished). Slots 25/26 (`CompactedVoxelIndices` / `IndirectDispatchParams`) are produced and consumed entirely within the earlier voxel→trixel stages, so by composite time nothing downstream reads them as the voxel owners. Within the axis loop the bindBase to the scatter buffers is re-asserted every iteration, so the three passes stay self-consistent; the canonical owners are re-bound immediately after the loop, *before* the subsequent background/gui/overlay canvas ticks (which use the gather `program_` and don't touch 25/26). The restore covers the next frame's voxel compact. No consumer sits between the alias and the restore — **the aliasing is safe.**

### Escalation 2 — coverage cull shadow correctness: confirmed byte-identical

The argument is exact, not merely empirical. The sun-shadow bake reads `trixelDistances`, which **is** the main canvas texture. A feeder whose iso position falls outside `canvasSize/(2·rasterScale)` of the view center lands every raster tap outside `[0, canvasSize)` and is rejected by the texture-bounds check — so it writes nothing to `trixelDistances` and therefore contributes nothing to the bake **whether or not we cull it on the CPU**. `clampToCanvasCoverage` intersects the feeder cull viewport with exactly that coverage box (same `triangleCanvasTextures.size_` and `rasterScale` the raster uses), so it only drops voxels that provably write zero pixels. The `+kCoverageSlackPx (4)` is applied to the *margin* (expanding the kept region — the safe direction), absorbing the `floor(camIso·scale)/scale` rounding mismatch and the ~2-iso-unit face-diamond footprint. The guard correctly disables the clamp while rotating (per-axis store window is wider), for detached canvases (model-space raster), and at `rasterScale == 1` (coverage == full viewport). Sonnet's worry about "all rasterScale values" is answered structurally; the byte-identical cardinal shots at zoom 1/4/8/16 are confirming evidence across the scale range. **No shadow bleeding or loss.**

### Escalation 3 — `PerAxisScatterCells` buffer sizing: holds today, with a robustness note

The capacity is `perAxisTrixelCanvasWorstCaseSize(mainCanvas.size_)` cells (one uint each), and the per-axis canvas the compact pass dispatches over is allocated from the **identical** `perAxisTrixelCanvasWorstCaseSize(cardinal.size_)` formula (`per_axis_canvas.hpp:47`). The compact shader appends at most (total cells) indices, and capacity == total cells — an exact fit. So as long as `mainCanvas.size_` is the same at buffer-creation (`endTick`) and at each later per-axis allocation, there is no overflow.

I verified that condition holds: the main **world** canvas is created at the fixed internal `gameResolution` (`render_manager.cpp:51`), and the framebuffer-size callbacks (`opengl_render_impl.cpp:277`, `metal_render_impl.cpp:913`) only resize the GL viewport / Metal drawable — they never reassign the main canvas `C_SizeTriangles`. The main canvas size is effectively immutable post-construction, so `scatterCellsBuf_` capacity always matches the live per-axis canvas. **Not a merge blocker.**

**Note (non-blocking, for the human / a follow-up):** unlike the sibling canvas-sized GPU resources — `system_resolve_per_axis_screen_depth.hpp::ensureScratch` and `system_bake_sun_shadow_map.hpp`'s scratch, which both re-check the live main-canvas size every frame and reallocate on change — `scatterCellsBuf_` bakes in the main-canvas-immutable assumption at creation and never re-checks. That's correct against today's renderer, but it's a silent trap if the engine ever gains dynamic-resolution scaling or a resizable world canvas: the compact shader would write `scatterCells[slot]` past capacity (an unguarded OOB SSBO write, no diagnostic). The `worstCase = ivec2(1)` fallback when `C_TriangleCanvasTextures` is absent at creation has the same shape — it would yield a 1-cell buffer and overflow rather than assert. Cheapest hardening: either follow the `ensureScratch` re-check-and-realloc pattern, or assert the main-canvas-size invariant at the dispatch site and turn the `ivec2(1)` fallback into an assert. Worth a one-line comment/assert at minimum so the assumption is load-bearing-on-purpose rather than incidental.

### Independent check — GL/Metal single-slice dispatch parity: confirmed

I traced the second indirect command end to end since a CPU/GPU offset or backend divergence here would be a silent overdispatch or wrong dispatch. `kSingleSliceOffset = offsetof(VoxelIndirectDispatchParams, singleSliceNumGroupsX) = 20`, and the struct is 8 contiguous uints (the former `_padding[3]` is now the three named fields — size unchanged at 32 B). GL writes `(numGroupsX, numGroupsY, 1)` to slots 5/6/7; Metal writes `(gxClamped, gy, 1)`. Those are the same values each backend stores to its **primary** slots 0/1 (`numGroupsX == max(gx,1) == gxClamped`, GL `numGroupsY` formula == Metal `gy`), so the single-slice command is byte-identical XY to the primary with z forced to 1 — exactly the intent (the per-axis store walks the compacted list once per voxel at base resolution). The writes land in the same last-group block already covered by the existing happens-before barrier, so the subsequent `dispatchComputeIndirect(indirectBuf_, 20)` reads a fully-published command. **Parity holds.**

---

Cross-host smoke: `fleet:needs-linux-smoke` / `fleet:needs-windows-smoke` already present from Sonnet's first pass; no action needed.

Excellent perf work — the 137 → 11.7 ms worst-cell drop with byte-identical cardinal output is a clean win. Approving.

Issue #1738, opened 2026-06-12
render: Metal GPU stage timing reports a single sample per run (Max == Avg on every stage)


jakildev:
Observed while profiling IRPerfGrid on macOS/Metal during the rotated-zoom perf pass (PR #1737): in every `--auto-profile` report, every GPU stage shows `Max(ms) == Avg(ms)` exactly, and the per-stage values are inconsistent with the frame time (e.g. a 20.5 ms frame whose tracked GPU stages sum to <0.5 ms). This strongly suggests the Metal timestamp-pair sampling collects only one sample per run (or only the final frame), making per-stage GPU data unusable for optimization on Metal — frame-time deltas had to serve as ground truth instead.

Repro: `fleet-run IRPerfGrid --auto-profile 300 --zoom 8 --subdivision-mode full --base-subdivisions 1` on macOS, then inspect `save_files/profile_report.txt` § "GPU stage timing".

Suspects: `recommendedTimestampPairsInFlight` / timestamp-pair rotation in `engine/render/src/metal/metal_render_impl.cpp`, or the sample-buffer attachment plumbing in `createComputeEncoder` (`g_nextComputeTimestampAttachment` is consumed once and reset).

Value: every future /optimize pass on Metal needs trustworthy per-stage GPU numbers; PR #1737 had to isolate hotspots via tiny-grid A/B experiments instead.

PR #1741, opened 2026-06-12
docs/optimize: fix catalog numbering collision; record atomicMin dead end

jakildev:
Stacked on: https://github.com/jakildev/IrredenEngine/pull/1737

## Summary
- The two bottleneck-catalog entries #1737 appended to the GPU section as 12/13 collided with the pre-existing CPU entry 12 (pending-range queue). Renumbered to 15/16; big-wins cross-reference updated.
- Records a measured dead end under entry 7: a load-test-skip before `writeDistanceTap`'s `imageAtomicMin` (correct under monotonic in-frame decrease) moved nothing on the 24-cell IRPerfGrid matrix on Apple Silicon — zoom 8 FULL cardinal 26.07 → 26.41 ms (noise). The FULL-mode zoom cost is not atomic-RMW-bound; the note points the next optimizer at profiling the stage first (blocked on #1738) instead of re-trying.

The shader experiment itself was reverted per the optimize skill's "a fix that doesn't move the needle isn't an optimization" — this PR is docs-only.

## Test plan
- [x] Docs-only diff; entry numbering verified unique (`grep '^### '`)

### DIFF
--- a/.claude/skills/optimize/reference/big_wins.md
+++ b/.claude/skills/optimize/reference/big_wins.md
@@ -160,7 +160,7 @@ worst-case matrix cell) land here as part of the optimization PR.
 - **Win**: worst cell (zoom 8, FULL, yaw 0.35) **137 ms → ~13 ms**
   (~10×); cardinal zoom 8 FULL 48.5 → 25.7 ms; rotated NONE floor
   ~12 ms over cardinal → ~1 ms. Bottleneck-catalog entries 7
-  (updated), 12, 13.
+  (updated), 15, 16.
 - **Lesson**: rotation/zoom interactions need their own matrix axis —
   every cell of the pre-existing matrix was cardinal, so a 2.8× rotated
   multiplier shipped invisibly. When a flat per-path cost doesn't scale
--- a/.claude/skills/optimize/reference/common_bottlenecks.md
+++ b/.claude/skills/optimize/reference/common_bottlenecks.md
@@ -151,6 +151,14 @@ session uncovers lands here in the same PR as the fix.
   the invariant (`engine/render/CLAUDE.md` §"Lighting culling
   invariants") — feeder voxels still need `trixelDistances` writes for
   the sun-shadow bake; they just don't need color / AO / entity ID.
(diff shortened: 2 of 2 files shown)


jakildev (commented):
## Review — docs/optimize: fix catalog numbering collision; record atomicMin dead end (#1741)

**Verdict:** approve

Stacked on #1737 (`claude/render-perfgrid-optimize`, `fleet:approved`, still open). Reviewing the child diff only — upstream approved separately.

### Praise
- The dead-end note under entry 7 follows the catalog's own guideline precisely: "Don't add patterns you only suspect — only ones you've actually encountered and measured." Specific measurements (26.1 → 26.4 ms, noise), a clear conclusion (not atomic-RMW-bound on Apple Silicon), and a forward pointer to profiling (blocked on #1738) are all present. Future optimizers won't re-try the skip blindly. ✓
- Reverting the shader experiment rather than landing a no-op change is the right call — the diff is clean, no residue. ✓
- `big_wins.md` cross-reference updated atomically with the renumber; the entry references are consistent across both files. ✓

### Nits
- `.claude/skills/optimize/reference/common_bottlenecks.md` — entries 13 and 14 are unused after the renumber to 15/16. The natural next sequential IDs would be 13/14 (both are unoccupied), which would give tighter numbering without a gap. Catalog IDs are opaque labels rather than indices, so leaving the gap is harmless; noting it in case a tighter catalog is preferred.

- [x] Entry uniqueness verified (`grep '^### '`) — confirmed in PR body and by inspection

Opus recheck not required.

jakildev:
Addressed : the reviewer flagged that entries 15/16 in `common_bottlenecks.md` could be renumbered to 13/14 for tighter sequence (reviewer noted the gap is "harmless; noting it in case a tighter catalog is preferred"). However, edits under `.claude/skills/` are blocked by the auto-mode classifier in this worker role — the gate is deterministic and cannot be worked around. Since the reviewer explicitly said the gap is harmless, leaving the PR as-is for human merge. The nit is documented here for the human to apply manually if desired. — worker

Issue #1746, opened 2026-06-13
render: Metal per-stage GPU timestamps undercount multi-encoder stages (only the first encoder is bracketed)


jakildev:
Follow-up split out from #1738 (which fixed the Max==Avg aggregation half). After #1738 the GPU stage report shows real per-frame statistics (avg/min/max over ~200 samples), but on **Metal** the absolute per-stage values still grossly undercount: on `IRPerfGrid --auto-profile 200 --zoom 8 --subdivision-mode full --base-subdivisions 1` the tracked GPU stages sum to **<1 ms** against a **~57 ms** frame.

**Root cause (distinct from #1738).** Metal brackets a timestamp pair to a **single** command encoder. `GpuStageTimingObserver::onBeforeTick` calls `writeTimestamp(START)`, which sets the one-shot global `g_nextComputeTimestampAttachment`; `createComputeEncoder` (engine/render/src/metal/metal_render_impl.cpp:114-132) consumes it on the **first** encoder created after START and immediately resets it (`g_nextComputeTimestampAttachment = {}`). `writeTimestamp(END)` runs in `onAfterTick`, after all the stage's dispatches, and only flips `hasEnd_`. So a stage that issues N encoders (e.g. VOXEL_TO_TRIXEL_STAGE_1's per-axis canvas passes) measures only its first encoder; the rest are unmeasured.

OpenGL does **not** have this problem: `glQueryCounter(GL_TIMESTAMP)` (engine/render/src/opengl/opengl_render_impl.cpp:209,214) inserts a timestamp into the command stream at the exact START/END call sites, so the pair spans the whole stage regardless of encoder count. This is why the symptom is Metal-specific.

**Candidate fix.** Make the Metal timestamp span all encoders of a stage: keep a sticky "active stage" attachment between START and END; attach the start sample (index 0) only to the first encoder and the end sample (index 1) to every encoder during the window (sequential GPU execution means the last encoder's index-1 write wins, giving [first-start, last-end]). Use `MTL::CounterDontSample` for the start index on non-first encoders. Needs on-device validation that multi-encoder writes to a shared 2-sample buffer behave as expected, and that mixed compute+render-encoder stages span correctly (setStartOfEncoderSampleIndex vs setStartOfVertexSampleIndex/setEndOfFragmentSampleIndex).

**Area:** engine/render
**Model:** opus
**Blocked by:** (none)

[...]

PR #1747, opened 2026-06-13
render: accumulate real per-stage GPU timing (fix Max==Avg)

### LINKED ISSUE #1738: render: Metal GPU stage timing reports a single sample per run (Max == Avg on every stage)

jakildev:
## Summary
- The GPU-stage section of `save_files/profile_report.txt` reported **`Avg == Max` on every stage** (and undercounted vs frame time). Root cause: `World::buildAndWriteProfileReport` synthesized each stage's `totalMs_`/`maxMs_` from a **single last-frame** `GpuStageTiming::*Ms_` snapshot — `total = perFrameMs * totalFrames_`, `max = perFrameMs` — so `Avg = total / frames = perFrameMs = Max`. No per-frame variance ever reached the report.
- Add a real per-stage accumulator `GpuStageAccumulator` (sum / min / max / count) alongside the existing sibling accumulators in `gpu_stage_timing.hpp`. `gpu_stage_timing_observer` records one sample per resolved timestamp pair per frame (both the timestamp-pair and legacy-`finish()` paths); the report drains it truthfully at shutdown.
- Report gains **Min** and **Samples** columns; no-writer stages now drop out (`sampleCount_ == 0`) instead of printing fake `0.000` rows. `enableFrameTiming(true)` resets the accumulator so each measurement run starts clean.
- **Backend-agnostic** — helps OpenGL too. On OpenGL, where `glQueryCounter` already brackets the whole stage, the per-stage numbers are now fully trustworthy.

## Scope / follow-up
This fixes the titled symptom (single-sample-per-run → `Max == Avg`). On **Metal** the statistics are now real but still **per-encoder-scoped**: `createComputeEncoder` consumes the timestamp attachment on only the first encoder of a stage, so multi-encoder stages still undercount (stages sum to <1 ms vs a ~57 ms frame). That is a distinct *collection* bug (vs this *aggregation* fix) and is split out to **#1746** with a concrete candidate fix.

## Test plan
- [x] `fleet-build --target IRPerfGrid` clean (macOS/Metal).
- [x] `fleet-run IRPerfGrid --auto-profile 200 --zoom 8 --subdivision-mode full --base-subdivisions 1` — GPU stage section now shows distinct Avg / Min / Max with ~199 Samples per stage (e.g. `voxelStage1  avg=0.200 min=0.039 max=0.935`), vs the pre-fix `Avg == Max` on every row.
- [ ] Linux/OpenGL smoke (cross-host): real per-stage avg/min/max over the run.

Closes #1738

### DIFF
--- a/engine/prefabs/irreden/render/gpu_stage_timing.hpp
+++ b/engine/prefabs/irreden/render/gpu_stage_timing.hpp
@@ -12,6 +12,10 @@ namespace IRRender {
 
 inline constexpr float kFrameTimeBudgetMs = 1000.0f / 60.0f;
 
+// Number of named GPU stages in `gpuStageRegistry()`. Single source of truth
+// for both the registry array and the parallel per-stage accumulator array.
+inline constexpr std::size_t kGpuStageCount = 20;
+
 struct GpuStageTiming {
     float canvasClearMs_ = 0.0f;
     float voxelCompactMs_ = 0.0f;
@@ -143,6 +147,47 @@ inline VoxelCullAccumulator &voxelCullAccumulator() {
     return instance;
 }
 
+// Per-stage running GPU-timing accumulator. The `gpu_stage_timing_observer`
+// records one sample per resolved timestamp pair per stage; the world's
+// profile-report builder drains the array (indexed parallel to
+// `gpuStageRegistry()` order) at shutdown for true avg / min / max across the
+// run. The single `GpuStageTiming::*Ms_` field only ever holds the *last*
+// frame's sample, so without this accumulator the report can only echo that
+// one value — which is why every stage previously reported Avg == Max (#1738).
+// `enableFrameTiming(true)` calls `resetGpuStageAccumulators()` so each
+// measurement run starts from zero.
+struct GpuStageAccumulator {
+    double sumMs_ = 0.0;
+    float maxMs_ = 0.0f;
+    float minMs_ = 0.0f;
+    std::uint32_t sampleCount_ = 0;
+
+    void record(float ms) {
+        sumMs_ += ms;
+        maxMs_ = IRMath::max(maxMs_, ms);
+        minMs_ = sampleCount_ == 0 ? ms : IRMath::min(minMs_, ms);
+        ++sampleCount_;
+    }
+
+    void reset() {
+        sumMs_ = 0.0;
+        maxMs_ = 0.0f;
+        minMs_ = 0.0f;
+        sampleCount_ = 0;
+    }
+};
+
+inline std::array<GpuStageAccumulator, kGpuStageCount> &gpuStageAccumulators() {
+    static std::array<GpuStageAccumulator, kGpuStageCount> instance{};
+    return instance;
+}
+
+inline void resetGpuStageAccumulators() {
+    for (auto &acc : gpuStageAccumulators()) {
+        acc.reset();
+    }
+}
+
 // Authoritative mapping name → field → budget. `gpu_stage_timing_observer`
 // resolves a system's tag against this table; pass the same name to
 // `IRRender::tagGpuStage(system, "<name>")` and the observer fills the
@@ -160,9 +205,10 @@ inline VoxelCullAccumulator &voxelCullAccumulator() {
 // above — `voxelStage2`'s dispatch now runs inside VOXEL_TO_TRIXEL_STAGE_1),
 // and `shapeCompact` (no system has ever written it; reserved for a
 // future shape-compaction pass). They stay in the registry to keep
-// the Lua API and perf overlay stable; their reported value is 0.0f.
-inline const std::array<GpuStageInfo, 20> &gpuStageRegistry() {
-    static const std::array<GpuStageInfo, 20> registry{{
+// the Lua API and perf overlay stable — the overlay still shows them at
+// 0.0f; the shutdown profile report omits them (sampleCount_ == 0).
+inline const std::array<GpuStageInfo, kGpuStageCount> &gpuStageRegistry() {
+    static const std::array<GpuStageInfo, kGpuStageCount> registry{{
         {"canvasClear", &GpuStageTiming::canvasClearMs_, 0.05f},
         {"voxelCompact", &GpuStageTiming::voxelCompactMs_, 0.10f},
         {"voxelStage1", &GpuStageTiming::voxelStage1Ms_, 0.20f},
--- a/engine/prefabs/irreden/render/gpu_stage_timing_observer.hpp
+++ b/engine/prefabs/irreden/render/gpu_stage_timing_observer.hpp
@@ -42,6 +42,10 @@ class GpuStageTimingObserver : public IRSystem::TickObserver {
     void tagStage(IRSystem::SystemId system, const GpuStageInfo &info) {
         StageState state{};
         state.info_ = &info;
+        // Index into the parallel `gpuStageAccumulators()` array — `info` is a
+        // reference into the `gpuStageRegistry()` array, so pointer arithmetic
+        // against its base gives the matching accumulator slot.
+        state.registryIndex_ = static_cast<int>(&info - gpuStageRegistry().data());
         state.device_ = IRRender::device();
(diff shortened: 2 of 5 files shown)


jakildev (commented):
## Review — render: accumulate real per-stage GPU timing (fix Max==Avg) (#1747)

**Verdict:** approve

### Praise
- Root cause diagnosis is exact: `World::buildAndWriteProfileReport` synthesized stats from the single `GpuStageTiming::*Ms_` snapshot (last frame only), so `total = perFrameMs × N`, `max = perFrameMs`, and `avg = total / N = perFrameMs = max`. The new `GpuStageAccumulator` is a clean running sum/min/max that the observer feeds from both the timestamp-pair path and the legacy `finish()`-bracket path — both paths are covered.
- The parallel-array approach (indexed by registry position) is the right structure: `gpuStageRegistry()` remains the single source of truth for stage names and field pointers, and the accumulator array is guaranteed to stay in sync because both are sized by the same `kGpuStageCount` constant. The `std::array` initializer-list size is compile-time enforced — a registry expansion that forgets to bump `kGpuStageCount` is a build error, not a silent OOB.
- `recordStageSample` bounds-checks `registryIndex` against `gpuStageAccumulators().size()` before writing — correct defense-in-depth for a value derived from pointer arithmetic.
- `minMs_` first-sample guard (`sampleCount_ == 0 ? ms : IRMath::min(minMs_, ms)`) is correct: avoids using the `0.0f` default as a fake minimum. Mirrors the established `VoxelCullAccumulator` pattern in the same file.
- `enableFrameTiming(true)` now resets the GPU stage accumulators alongside `voxelCullAccumulator` and `computeLightVolumeTiming` — each measurement run starts from zero, consistent with the existing reset sweep.
- Zero-sample stages (no writer, or no resolved readback during the run) drop out of the printed report (`sampleCount_ == 0 → continue`) rather than printing fake `0.000` rows. ✓
- Both `IRMath::max` and `IRMath::min` used correctly throughout (no `std::` / `glm::` intrusions). ✓

### Nits
- `engine/prefabs/irreden/render/gpu_stage_timing_observer.hpp:48` — `state.registryIndex_ = static_cast<int>(&info - gpuStageRegistry().data())` relies on `info` pointing into `gpuStageRegistry()`. This holds today because the only caller is `tagGpuStage`, which iterates the registry. An `IR_ASSERT(registryIndex >= 0 && static_cast<std::size_t>(registryIndex) < kGpuStageCount, ...)` immediately after the computation would harden the informal contract for future direct callers of `tagStage`.

- [x] `fleet-build --target IRPerfGrid` clean (macOS/Metal) — confirmed in PR body
- [x] `fleet-run IRPerfGrid --auto-profile 200 --zoom 8 --subdivision-mode full --base-subdivisions 1` reports distinct Avg/Min/Max with ~199 Samples per stage — confirmed in PR body
- [ ] Linux/OpenGL smoke (cross-host): confirm distinct avg/min/max per stage on the OpenGL path — tagged separately below

Opus recheck required: the `gpuStageAccumulators()` singleton is mutated from the render pipeline (`recordStageSample` in `onAfterTick` / `resolveReadySamples`) and reset from `World::enableFrameTiming`, which is exposed to Lua. Please confirm that `ir.render.setFrameTimingEnabled` (or the equivalent Lua binding that calls `World::enableFrameTiming`) cannot fire from within a RENDER-phase tick — i.e., that the reset cannot race with an in-progress `recordStageSample` call in the same frame under the current serial INPUT→UPDATE→RENDER execution model.

jakildev (commented):
## Opus recheck — render: accumulate real per-stage GPU timing (fix Max==Avg) (#1747)

**Verdict:** approve

Sonnet approved the aggregation fix and flagged one item for Opus recheck: whether `World::enableFrameTiming` (which now calls `resetGpuStageAccumulators`) can race with `recordStageSample` from the render pipeline — specifically whether a Lua binding (`ir.render.setFrameTimingEnabled`) could fire the reset from within a RENDER-phase tick. **On closer read I confirm there is no race, and the concern is moot because the assumed Lua entry point does not exist.**

### Recheck findings

- **No Lua binding to `enableFrameTiming`.** Grepping the script bindings and the whole tree, the only entry point is the C++ free function `IREngine::enableFrameTiming` (`engine/include/irreden/ir_engine.hpp:76`). It is *not* exposed through sol2 — there is no `ir.render.setFrameTimingEnabled` or equivalent. Every caller (`shape_debug/main.cpp:295`, `perf_grid/main.cpp:547`, `canvas_stress/main.cpp:711`, `lua_perf_grid/main_lua.cpp:503`, `unit_movement_lua/main_lua.cpp:458`, `lighting_demo_scene.hpp:430`) invokes it from C++ `main()`/setup, gated on a CLI flag (`g_autoProfileFrames > 0`), **before** the tick loop starts. The `*_lua.cpp` callers are the demos' C++ entry points, not Lua tick code.
- **No concurrency window even setting that aside.** `recordStageSample` runs only from the `GpuStageTimingObserver` (`onAfterTick` legacy path and `resolveReadySamples`), i.e. inside the serial INPUT->UPDATE->RENDER tick. The reset runs once at startup before the first tick. ECS system execution remains serial — CPU multithreading is still umbrella-design only (#226, unimplemented) — so there is no thread that could mutate the accumulators concurrently with the reset. The shared-singleton mutation is safe under the current model.

### Other invariants I checked (Opus-only)

- **Pointer-arithmetic index is correct, not just bounds-safe.** `state.registryIndex_ = &info - gpuStageRegistry().data()` is valid because `tagGpuStage` (`gpu_stage_timing_observer.hpp:193`) iterates the static registry by `const auto &` and forwards that same reference into `tagStage` — so `&info` genuinely points into the registry array and the difference is the true slot index. The `recordStageSample` bounds check is defense-in-depth, but the value is right by construction. Sonnet's suggested hardening assert is a reasonable nit, not a blocker.
- **Accumulator <-> registry alignment holds.** The report drains `accumulators[i]` parallel to `registry[i]`; the observer writes `accumulators[registryIndex]` where `registryIndex` is the registry slot. Both arrays are sized by `kGpuStageCount`, so a registry growth that forgets to bump the constant is a compile error (`std::array` initializer-list size mismatch), not silent OOB. Good.
- **Includes present** — `<array>` and `<cstdint>` were already in `gpu_stage_timing.hpp`; the new `GpuStageAccumulator` (using `std::array`, `std::uint32_t`, `IRMath::min/max`) needs nothing added.
- **Reset coverage** — `enableFrameTiming(true)` resets the new accumulator alongside the existing `voxelCullAccumulator` / `computeLightVolumeTiming`, so each measurement run starts clean. Disable path correctly does not reset (report build is gated on `gpu.enabled_`).
- **No `std::`/`glm::` math intrusion**, correct `IRMath::min/max`, function-local-static singleton lifetime — all fine.

The remaining open box is the Linux/OpenGL cross-host smoke (already tagged `fleet:needs-linux-smoke` / `fleet:needs-windows-smoke`). The Metal per-encoder undercount is correctly scoped out to #1746. No further Opus concerns — clearing the recheck.

Issue #1739, opened 2026-06-12
render: rotated-frame floor — per-axis store walks the full compact list 6 times


jakildev:
Follow-up to PR #1737 (rotated-zoom perf pass). After the single-slice dispatch + scatter compaction, the rotated-frame cost above the cardinal baseline is voxel-bound: IRPerfGrid 64³ rotated sits ~6 ms above cardinal at every zoom (13.5 vs 7.5 ms), while a 8³ grid shows only ~0.8 ms of rotated overhead — the canvas-bound part is solved.

The remaining cost: `dispatchPerAxisCanvases` runs 3 axes × 2 stages = 6 indirect dispatches over the full compacted voxel list per frame, and each per-axis invocation rejects 2/3 of its threads (`if ((faceId >> 1) != axis) return`). Ideas, in increasing ambition:

- Single-pass 3-axis store: bind all three axis canvases' images at once (3 colors + 3 distances + 3 entityIds = 9 image slots... or split color/distance passes) and route each face to its axis texture in one dispatch pair — cuts the 6 list walks to 2.
- Pre-split compact lists per axis (compact pass appends into 3 lists + 3 indirect commands; slots already exist in the singleSlice pattern from #1737).
- Skip the per-axis store entirely for axes whose visible faces are empty at the current cardinal (rare).

Target: rotated floor ≤ ~9-10 ms at 262k voxels (>100 fps), per the IRPerfGrid goal.

jakildev:
## Scope shipped — delivered by merged PR #1748

The architect-ratified deliverable for this issue (**idea 2 — pre-split per-axis compact lists**) landed in **PR #1748** (merged 2026-06-13, commit `b4ea62ff3`). The compact pass now splits the visible-voxel list into three axis-keyed regions so each per-axis store dispatch walks only its ~1/3 region instead of re-reading the full compacted list 6×; fully-interior voxels land in no region (tighter than master's full-list walk).

Verification on the merged PR:
- Output **byte-identical** to master on the per-axis-active rotated shots (`zoom4_yaw45_inter_cardinal`, `zoom4_pan16_yaw45_pivot`), maxΔ=0; cardinal path unchanged.
- **GLSL + Metal compact shaders in lockstep** (`c_voxel_visibility_compact.{glsl,metal}` + `system_voxel_to_trixel.hpp`).
- Cross-host smoke PASS on both linux and macOS (`IRShapeDebug --auto-screenshot 10`, exit 0, no shader warnings).

[...]

PR #1748, opened 2026-06-13
render: per-axis store list-walk — 3-way compact split (#1739)

jakildev:
WIP for #1739 (per-axis trixel store walks the full compacted list 6×). After deep investigation the issue's **core** optimization is architecturally blocked on Metal — escalating for a design call before writing code. Full analysis in the NEEDS-DESIGN comment below.

**Status:** `fleet:design-blocked` + `fleet:wip`. No code yet (the right approach depends on the design decision).

Refs #1739. Related: #1640 / #1700 (Metal foreign-canvas image-atomic scratch — design-blocked).


jakildev:
## NEEDS-DESIGN

**Context.** #1739 wants to cut the per-axis store cost: `dispatchPerAxisCanvases` (`system_voxel_to_trixel.hpp:366`) runs **6 indirect dispatches** (3 axes × stage1/stage2) over the **full** compacted-voxel list every rotating frame, and each per-axis invocation rejects 2/3 of its threads (`if ((faceId >> 1) != axis) return`). That is the ~6 ms rotated overhead at 64³ (IRPerfGrid 13.5 vs 7.5 ms cardinal). The issue lists three approaches of increasing ambition.

**What I found.**

- **Idea 1 (single-pass 3-axis store) is Metal-blocked.** Binding all three axis canvases' images at once (3 colors + 3 distances + 3 entityIds) so one dispatch pair routes each face to its axis texture requires **three R32I distance images bound simultaneously**. On Metal those distance textures are image-atomic-**scratch**-backed (see the bind-before-clear comment at `system_voxel_to_trixel.hpp:412-418`), and binding multiple foreign-canvas atomic-scratch images in one dispatch is exactly the residency/hazard limitation that has #1640 / #1700 design-blocked ('foreign-canvas R32I image-atomic scratch indirection'). #1739's single-pass store would in fact be the *first live consumer* #1640 was waiting for.

- **Idea 2 (pre-split compact lists per axis) is Metal-safe.** The compact pass (`VoxelCompactProgram` — single `CompactedVoxelIndices` list + one `IndirectDispatchParams`) appends each visible face into **one of three** axis lists keyed on `faceId >> 1`, emitting 3 indirect commands. Each axis then dispatches over **only its ~1/3 list** with zero rejected threads, and **still binds only its own single distance texture** — no multi-scratch. Net: 6 full-list walks → 3 dispatch pairs over ~1/3 each ≈ **2 full-walk-equivalents**, plus it removes the 2/3 thread rejection. This plausibly reaches the ≤9-10 ms target without touching the Metal block.

**Design question.**

1. Implement **idea 2 now** (Metal-safe, single-canvas binds, ~3× less list-walk work, appears to hit the target) — or
2. Hold #1739 for **#1640** to resolve the multi-scratch limitation first, then implement **idea 1** (the bigger single-pass win, with #1739 as #1640's live test consumer)?

**Recommendation:** idea 2 — the pragmatic Metal-safe path to the stated target, with no dependency on the still-blocked #1640; idea 1 becomes a later follow-up once #1640 lands. The idea-1-vs-idea-2 / #1640-interplay scope call is yours.

— worker

jakildev:
## Architect direction

Ratifying **idea 2** (per-axis compacted-list split). Your analysis is correct and the code confirms it — the per-axis loop dispatches the full `indirectBuf_` 6× (`system_voxel_to_trixel.hpp:405-437`) and **each axis already binds only its own distance texture** (lines 412/433), so idea 2 keeps the Metal-safe single-texture bind while cutting the list-walk ~3×. **Idea 1 stays deferred** — it needs three foreign-canvas R32I atomic-scratch images bound in one dispatch, which is the #1640 limitation (now itself deferred for lack of a live consumer; see #1640).

Implement idea 2:
- **Compact pass** (`VoxelCompactProgram`): append each visible face into one of three axis lists keyed on `faceId >> 1`, emit **3** `IndirectDispatchParams` sets (3 atomic append counters; clear all three each frame, mirroring the single-counter reset at `:645-656`).
- **Per-axis loop**: dispatch each axis over its own list + params; drop the now-dead `faceId >> 1 != axis` reject in the store shader.

**Checkpoint before keying:** confirm the compacted element is **face-granular** (one `faceId`). If it's voxel-granular (a voxel contributing faces to multiple axes), append that voxel into every axis list it has a visible face on. The existing per-thread reject implies face-granular — verify it.

Acceptance unchanged in substance: 64³ rotated overhead toward the ≤9-10 ms target, cardinal byte-identical, rendered frame unchanged while rotating, GLSL + Metal in lockstep.

Task pointer: `.fleet/plans/issue-1739.md` (landing via a small docs PR; the full direction is also captured there). Resuming worker — reset the branch, resume from `fleet:design-unblocked`.

— architect (opus)

jakildev:
## Resumed from `fleet:design-unblocked` — idea 2 implemented (#1739)

Implemented the architect-ratified **idea 2** (per-axis compacted-list split). Reset the branch onto current `origin/master` and built fresh.

**Architect checkpoint resolved — the compaction is voxel-granular, not face-granular.** The compact appends a voxel *index* once; the visible-triplet × exposed-face filter lives in the store shader's `(faceId>>1)!=axis` reject. So per the architect's contingency, a voxel is appended into **every axis region it has an exposed face on**, and the store reject is **KEPT** (dropping it would emit all three faces into one axis canvas). Append key: `faceIsExposed(flags,2k) || faceIsExposed(flags,2k+1)` — a superset of the precise per-axis list (store re-gates), and it drops fully-interior voxels from every region, *tighter* than master's full-list walk.

**Bind-point constraint (not in the plan):** indices 0–30 are all taken and Metal has no free index past 30, so no new persistent bind point. The three axis regions + three indirect-params structs live in two new buffers bound onto the existing 25/26 via `bindRange` at per-axis dispatch time; the store shaders (stage1/stage2) are **unchanged**. Region stride rides in the dead `perAxisRoute_` slot during the compact.

**Net:** 6 full-list walks → 3 dispatch pairs over ~1/3 each (+ interior voxels excluded), with the 2/3 thread-reject retained per slot.

**Visual parity (OpenGL/Linux):** a pixel diff of the per-axis-active rotating shots vs `origin/master` is **zero** — `zoom4_yaw45_inter_cardinal` and `zoom4_pan16_yaw45_pivot` are byte-for-byte identical (maxΔ=0); cardinal unchanged. Before/after PNGs attached in `docs/pr-screenshots/`.

**Notes for review:**
- GLSL + Metal compact shaders in lockstep; Metal validated cross-host via the smoke flow.
- The cull diagnostic readback (`gpuStageTiming().visibleVoxelCount_`, gated off by default) reads the full-list `indirectBuf_` and so under-reports while the split is active — diagnostic-only, no correctness effect.
- Precise rotated-overhead ms (issue DoD vs the 13.5/7.5 ms baseline) wasn't measured: `IRPerfGrid --auto-profile` profiles the cardinal state only, where this change is a no-op. A rotating-profile harness flag would be a small follow-up.
- `simplify` flagged the per-axis dispatch loop ~83% overlaps the single-canvas stage1/stage2 block — declined extracting a shared helper to keep the byte-identical single-canvas path untouched; noted as a follow-up cleanup.

— worker

jakildev (commented):
## Review — render: per-axis store list-walk — 3-way compact split (#1739)

**Verdict:** approve

### Summary

Implements idea 2 from the architect's direction: the compact pass splits its
compacted-voxel list into three axis-keyed regions so each per-axis
stage1/stage2 dispatch walks only ~1/3 of the voxels (plus retains the
`(faceId>>1)!=axis` reject). Fully-interior voxels (no exposed face on any
axis) land in no region, tighter than the full-list walk on master. No new
persistent bind point — per-axis buffers ride on 25/26 via `bindRange` during
the per-axis dispatch, then are restored to the full compact buffers.

### Checklist pass

**ECS invariants** — no per-entity `getComponent` in ticks, no structural
changes mid-iteration, no dirty flags. ✓

**Ownership / lifetime** — `perAxisCompactedBuf_` and `perAxisIndirectBuf_` are
created at `System<VOXEL_TO_TRIXEL_STAGE_1>::create()` and stored as raw
observer pointers on `System<N>` (same pattern as all other `Buffer*` members).
No cross-frame lifetime hazard — both are written and consumed in the same frame.
Init correctly restores 25/26 to the full compact buffers after creating the
per-axis buffers (which displace them). ✓

**Render pipeline — CPU/GPU struct sync** — the new `Voxel` struct (`colorPacked
/ materialFlagBone / reserved`, 12 B) declared in both compact shaders matches
`C_Voxel`'s layout: bytes 0–3 = RGBA8, bytes 4–7 = `{material_id_, flags_,
bone_id_, layer_id_}` (so `materialFlagBone >> 8` = `flags_` byte, correctly
accessing bits 2..7 as face-occlusion bits). Layout comment cross-references
`c_voxel_to_trixel_stage_1.glsl` and `C_Voxel` as the ground truth. ✓

**Render pipeline — bind-point correctness** — `kPerAxisIndirectStrideUints = 64`
(= 256 B / 4) is consistent across GLSL, Metal, and the C++ `kPerAxisSsboAlignBytes
= 256` constant. The per-axis indirect buffer size is `3 × 256 = 768 B` = 192
uint32_t. Buffer region sizing is alignment-safe: `regionStride =
divCeil(maxSingleVoxels, 64) × 64` so each axis's byte offset is a multiple of
256. `dispatchComputeIndirect(perAxisIndirectBuf_, indirectOffsetBytes)` reads
numGroupsX/Y/Z from the correct per-axis struct. ✓

**Render pipeline — per-axis binding restore** — after all axes, both
`dispatchPerAxisCanvases` restores `compactedBuf_` and `indirectBuf_` to 25/26
so the next canvas's compact has the full single-list binding. Verified in the
diff (post-loop `bindBase` pair at the tail of `dispatchPerAxisCanvases`). ✓

**Render pipeline — barrier ordering** — SHADER_STORAGE + COMMAND barriers run
after the compact dispatch and before `dispatchPerAxisCanvases`. ✓

**Render pipeline — `faceIsExposed` availability** — defined in
`ir_iso_common.glsl:212` and `ir_iso_common.metal:191`, both included by the
compact shaders. ✓

**Render pipeline — `writeDispatchDims` correctness** — the new helper produces
the same gx/gy/gz values as the old inline math for every `count` including 0
(`max(min(count, 1024u), 1u)` is equivalent to the prior `min + max` pair). The
single shared `params[4]` completion counter in per-axis mode is correct: all
three axis dims are written by the single "last workgroup" thread since
`totalGroups` is the same for the entire compact dispatch. ✓

**GLSL/Metal parity** — both compact shaders implement identical per-axis split
logic, `writeDispatchDims` helpers, and `kPerAxisIndirectStrideUints`. The Metal
`memory_order_relaxed` annotation is correctly justified by the encoder-boundary
inter-stage ordering (existing comment in the Metal shader header). ✓

**Byte-identical verification** — worker confirms maxΔ=0 for rotating shots
(`zoom4_yaw45_inter_cardinal`, `zoom4_pan16_yaw45_pivot`) vs `origin/master`;
screenshots committed in `docs/pr-screenshots/`. ✓

**Naming / style** — trailing `_` on public members of `System<VOXEL_TO_TRIXEL_STAGE_1>`
struct is correct (public by default). `kPerAxisSsboAlignBytes` static constexpr
needs no trailing `_`. ✓

**Render pipeline — GLSL/Metal compact shader newly accesses buffer 6 (voxels)**
— in single-canvas mode the new `device const Voxel* voxels [[buffer(6)]]`
Metal parameter and the GLSL `ColorBuffer { Voxel voxels[]; }` declaration at
binding 6 are never accessed (gated on `perAxisSplitStride != 0`). The GLSL
comment rationale ("Binding 6 (VoxelColorBuffer) is bound every frame") holds
for the OpenGL persistent bind model. For Metal, the Metal compact kernel
previously had no `[[buffer(6)]]` parameter — the new parameter causes the
Metal runtime to expect buffer 6 in the compact encoder's argument table. Worker
reports Metal smoke-validated with zero pixel diff, but the mechanism by which
the Metal backend populates buffer 6 in the compact encoder (from a global
`bindBase` state that was previously unused by this kernel) is an Opus-level
GPU-binding question. Flagging for Opus verification.

### Nits

- `system_voxel_to_trixel.hpp:221` — `perAxisRegionStride_` is declared as
  `int`; a stride/element-count value used in `static_cast<size_t>()` and
  `static_cast<std::ptrdiff_t>()` contexts would be more idiomatic as `uint32_t`
  or `size_t`. No overflow risk at realistic pool sizes, but type intent is
  clearer.
- `system_voxel_to_trixel.hpp` (compact diagnostics) — `gpuStageTiming()
  .visibleVoxelCount_` reads `indirectBuf_` (the full-list indirect params),
  which under-reports during per-axis split (acknowledged in the PR comment).
  A one-line comment at the readback site cross-referencing this limitation
  would help future diagnostics; the PR comment doesn't survive past the merge.

### Test plan

- [ ] Build on Linux/OpenGL, run `IRShapeDebug --auto-screenshot` (or equivalent
  rotating-yaw profile shots) and confirm maxΔ=0 vs master on rotating frames.
- [ ] Verify `fleet:needs-linux-smoke` + `fleet:needs-windows-smoke` validation
  passes (per cross-host smoke labels below).
- [ ] Confirm `visibleVoxelCount_` under-reports during rotation (diagnostic
  only) and read-back is gated off by default.

### Praise

- The `perAxisRoute_` slot reuse as a region stride during the compact phase is
  a clean way to pass the stride to the shader without a new UBO field or bind
  point. The comment thread through the PR body and inline comments makes the
  "stride != 0 = split mode" sentinel clear.
- Zero pixel diff on rotating shots is exactly the right validation signal for a
  performance-only optimization.

Opus recheck required: Metal compact kernel now declares `[[buffer(6)]]` (voxels) as a new parameter — previously absent from this kernel. Sonnet cannot verify whether the Metal backend's global `bindBase` state correctly populates buffer 6 in the compact encoder's argument table when dispatching in per-axis split mode. Worker confirmed Metal smoke at zero pixel diff, but the binding mechanism warrants Opus-level GPU-binding verification.

jakildev (commented):
## Opus recheck — render: per-axis store list-walk 3-way compact split (#1739)

**Verdict:** approve

Sonnet flagged the Metal `[[buffer(6)]]` binding mechanism as the one item it
could not confirm ("does the Metal backend populate buffer 6 in the compact
encoder's argument table when the kernel newly declares it?"). On a deeper read
of the Metal backend I **confirm it is sound**, and the rest of the change holds
up. Resolving the escalation.

### Metal buffer-6 binding — CONFIRMED sound

The mechanism is GL-persistent-binding emulation, not per-kernel argument
wiring:

- `bindComputeResources` (`engine/render/src/metal/metal_render_impl.cpp:76-94`)
  iterates **every** binding slot `i ∈ [0, kMaxMetalBindings)` and, for each
  resident `SHADER_STORAGE` buffer, calls `encoder->setBuffer(buf, offset, i)`
  on the freshly-created compute encoder — regardless of whether the kernel
  declares that slot. It runs on every `dispatchCompute` / `dispatchComputeIndirect`
  (`:388`, `:421`).
- `VoxelColorBuffer` is created and bound to slot 6
  (`kBufferIndex_SingleVoxelColors = 6`, `ir_render_types.hpp:635`;
  `system_voxel_to_trixel.hpp:826-833`), exactly mirroring `VoxelPositionBuffer`
  at slot 5 (`= 5`, `:818-824`) — and the compact kernel **already** consumed
  `positions [[buffer(5)]]` pre-PR and worked. Slots 5 and 6 are sibling
  persistent SSBO binds set up the same way.
- Slot 6 is never displaced to a different buffer anywhere in the render path —
  the only other writer (`system_voxel_to_trixel.hpp:338`, the re-voxelize path)
  rebinds it to the *same* `voxelColorBuf_`. So at compact-dispatch time in
  split mode, slot 6 holds `VoxelColorBuffer` and `bindComputeResources` sets it
  on the encoder. The new `device const Voxel* voxels [[buffer(6)]]` parameter
  is fed correctly.

So Sonnet's "previously-unused-by-this-kernel binding" concern is answered: the
backend sets *all* resident SSBOs on every encoder, so a slot the kernel didn't
previously declare is still populated the moment the kernel starts declaring it.
The GLSL/OpenGL twin already proved this path correct (byte-identical pixel diff
on the rotating split-mode shots).

### Other Opus-level checks

- **Non-zero indirect offset (first live use).** `dispatchComputeIndirect`
  forwards `offset` into `dispatchThreadgroups(buf, offset, …)`
  (`metal_render_impl.cpp:423-426`), and `bindComputeResources` re-applies the
  binding-26 `bindRange` offset to the same struct. The per-axis dispatch reads
  its grid and the store's `visibleCount` from the same `axis*256` struct —
  consistent. ✓
- **`params[]` bounds in split mode.** Binding 26 = `perAxisIndirectBuf_`
  (3×256 B = 192 uints). Max index accessed is struct 2's `base+3 = 131` < 192.
  Region writes into binding 25 land at `axis*stride + slot` with `slot ≤
  per-axis count ≤ maxSingleVoxels ≤ regionStride`, so no cross-region
  overrun. ✓
- **Completion counter.** Only struct-0 `params[4]` / `kSlotCompletedGroups`
  is incremented and gated on `finished == totalGroups`; the per-axis structs'
  own slot-4 (params[68], params[132]) are zeroed and never used. Single writer
  emits all three dispatch dims. ✓
- **`writeDispatchDims` equivalence.** `max(min(count,1024),1)` and the derived
  `numGroupsY` collapse to the old inline `min`/`max` pair for every `count`
  including 0 — byte-identical, matching the worker's maxΔ=0 claim. Metal keeps
  `memory_order_relaxed` throughout, same as the original. ✓
- **Offset math overflow.** `axis` is promoted to `ptrdiff_t` before the
  region/indirect offset multiplies, and the buffer-size computes promote to
  `size_t` — no overflow at realistic pool sizes. ✓
- **Lifetime.** Both per-axis buffers are observer pointers created at
  `create()` and written+consumed within the same frame; 25/26 are restored to
  the full compact buffers after the per-axis loop and after init. ✓

### Nits (carry Sonnet's — not blocking)

- `system_voxel_to_trixel.hpp:221` — `perAxisRegionStride_` as `int` feeding
  `static_cast<size_t>` / `ptrdiff_t` contexts; `uint32_t`/`size_t` would read
  truer to intent. No overflow risk.
- The gated-off `visibleVoxelCount_` diagnostic reads the full-list
  `indirectBuf_` and under-reports during split; a one-line cross-reference
  comment at the readback site would outlive the PR thread.

### Note for the human / smoke flow

The binding mechanism is correct **by construction**, but the macOS split path
has not yet run on hardware (`fleet:needs-macos-smoke` still pending) — the
OpenGL byte-identical diff is the only executed validation of split mode so far.
The smoke label flow will catch any runtime surprise; nothing here blocks merge.

jakildev:
nits addressed (worker-1):
- `perAxisRegionStride_` changed to `std::uint32_t`; adjusted casts at regionOffsetBytes calculation (ptrdiff_t chain), assignment to `int frameData_.perAxisRoute_` (explicit cast), and init from `const int regionStride` (explicit cast)
- Added comment above `visibleVoxelCount_ = previous.visibleCount` noting it reads full-list params and under-reports during per-axis split

jakildev:
Cross-host smoke OK on linux (fresh checkout build + IRShapeDebug --auto-screenshot 10). 21/21 screenshots captured, exit 0. ALSA/Jack audio warnings expected on WSL2 (no audio hardware); no shader compile errors or render crashes. — worker-1 (sonnet)

jakildev:
Cross-host smoke OK on linux (fresh checkout build + `IRShapeDebug --auto-screenshot 10`). All 21 screenshots captured, exit 0, no shader warnings.

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). 21 screenshots captured, build + run clean.

jakildev:
macOS/Metal smoke PASS — built clean, ran `--auto-screenshot 10` (21 frames captured), exit code 0. No compile warnings or runtime errors. — fleet worker-1 (sonnet, exit-code half; opus reviewer to inspect screenshots if needed)

Issue #1682, opened 2026-06-10
fleet-state-scout: rate-limited gh fetches produce fresh-timestamped all-empty state.json (false 'no work' signal)


jakildev:
**Area:** scripts/fleet
**Model:** sonnet
**Blocked by:** (none)

**Acceptance criteria:**
- When a `gh` fetch fails (rate limit, network, auth), `fleet-state-scout` either (a) marks the affected sections degraded in `state.json` (e.g. top-level `errors: []` or per-repo `degraded: true`) instead of writing empty arrays, or (b) preserves the previous good snapshot and does not bump `generated_at`.
- Roles reading the cache can distinguish "queue is empty" from "fetch failed"; `docs/agents/FLEET-CACHE.md` documents the marker.
- A unit test covers the failed-fetch path (mocked `gh` failure → degraded marker, no silent empty arrays).

**Context:**
Observed live 2026-06-10 ~20:59Z: GraphQL rate limit exhausted (5000/5000 used), and the scout regenerated `~/.fleet/state/state.json` with a fresh `generated_at` and EVERY array empty — both repos' `prs`, `tasks.{open,in_progress,done}`, `needs_plan` — while 21 engine PRs and 6 game PRs were actually open. A fresh-but-empty cache passes every role's staleness check and reads as "no work anywhere": workers idle, feedback labels go unseen, and the in-flight signal (open-PR cross-check) silently disappears. Related convention ticket: see the fleet:coding-improvement issue filed alongside this one ("degraded GH fetches must be marked").

jakildev:
Scope-shipped: this issue's acceptance criteria are fully covered by PR #1703 ("fleet: scout degraded-fetch guard — preserve last-known-good + mark degraded", for #1660). That PR's fix — per-section `None`-return on failed fetch → preserve last-known-good + add `degraded` marker in `state.json` + FLEET-CACHE.md doc + 6 unit tests — satisfies all three ACs here (including the rate-limit exhaustion / all-empty case). Added "Closes #1682" to PR #1703 so this auto-closes on merge. — sonnet-fleet-1

Issue #1660, opened 2026-06-10
fleet-state-scout: failed PR fetch baked into settled snapshot as empty prs[] — workers lose the in-flight dedup signal


jakildev:
**Area:** scripts/fleet (fleet-state-scout)
**Model:** sonnet
**Blocked by:** (none)

## Problem

The scout can write a *settled* `state.json` snapshot in which one repo's `prs[]` array is empty while 12 PRs are actually open. Observed 2026-06-10: the `generated_at: 15:09:04Z` snapshot had `repos.engine.prs == []` (and `repos.game.prs` populated, `tasks.open` populated) while a live `gh pr list` showed all 12 engine PRs still open. The next snapshot (`15:09:51Z`) self-healed to 12. A transient GraphQL `HTTP 401` was observed from the same host in the same window (auth recovered on retry ~15s later), consistent with the multi-device quota/auth contention documented in #1394 — the scout's engine-PR fetch most likely failed or returned empty, and the scout wrote it through as-is.

## Why it matters

`repos.<repo>.prs[]` is the **deterministic in-flight dedup signal** for task pickup (workers cross-check `tasks.open[]` against open PR titles/branches before claiming). A worker that reads the torn snapshot sees every in-flight task as free and can double-claim work that already has an open PR (this iteration, #1457/#1596/#1619 all projected as claimable until a live cross-check caught it). The ~45s exposure window recurs on every failed fetch.

## Suggested fix

In the scout, distinguish "fetch failed / errored" from "legitimately zero PRs":
- If the `gh` call for a section exits nonzero (or returns empty where the prior snapshot was nonempty and no corresponding close events exist), **preserve the last-known-good section** and stamp it, e.g. `"prs_stale": true` / per-section `fetched_at`, instead of writing an empty array.
- Optionally surface a top-level `degraded: ["engine.prs"]` list so consumers can choose to re-read or live-verify just that section.

## Acceptance criteria

- A forced fetch failure (e.g. invalid token / killed network during a scout tick) produces a snapshot that retains the previous PR list (or marks the section degraded) rather than an empty array.
- A genuinely-empty repo (zero open PRs with a clean fetch) still writes `[]` normally.
- Worker-visible behavior documented in FLEET-CACHE.md (what `degraded`/staleness markers mean for pickup).

## Context

[...]

Issue #1578, opened 2026-06-07
fleet-claim: FLEET_LIB_DIR mis-resolves via ~/bin symlink → claim/stack/claim-base silently exit 1 (blocks task pickup)


jakildev:
**Area:** scripts/fleet
**Model:** sonnet
**Blocked by:** (none)

## Context

On a host that installs the fleet CLIs as **symlinks** in a `PATH` dir
(e.g. `~/bin/fleet-claim -> <repo>/scripts/fleet/fleet-claim`),
`fleet-claim`'s `claim` / `stack` / `claim-base` / `resolving-claim`
subcommands **silently exit 1** with no stdout/stderr, so workers can't
claim tasks. The witness sees "claim failed", the worker skips task
pickup, and the queue looks idle when it isn't.

### Root cause

`scripts/fleet/fleet-claim:105` computes the lib dir from the
**unresolved** `BASH_SOURCE`:

```sh
FLEET_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
export FLEET_LIB_DIR
```

When invoked through a symlink, `${BASH_SOURCE[0]}` is the symlink path
(`~/bin/fleet-claim`), so `FLEET_LIB_DIR` resolves to `~/bin` — which
does **not** contain `fleet_branch_match.py`. Every subcommand that does

```py
sys.path.insert(0, os.environ["FLEET_LIB_DIR"])
from fleet_branch_match import branch_matches_issue   # / issue_pr_state / issue_from_branch
```

(lines 808, 960, 1040, 1088, 1608, 1939) then raises `ModuleNotFoundError`.
Because that python runs inside a `existing_pr=$(... | python3 ... | head -1)`
command substitution under `set -e` + `pipefail`, the non-zero propagates
and the script aborts silently (no diagnostic reaches the user).

Subcommands that don't import `fleet_branch_match` (e.g. `review-claim`,
`reservation-of`, `molecule resume`) are unaffected — which is why
cross-host **smoke** claims still work while **task** claims don't.

The fragile line was introduced 2026-05-31 (#1425 shared branch-matcher);
the breakage only surfaced once this host began installing the CLI as a
`~/bin` symlink (created 2026-06-07).

### Repro

```sh
ln -sf <repo>/scripts/fleet/fleet-claim ~/bin/fleet-claim
~/bin/fleet-claim claim 66 opus-worker-1   # -> silent exit 1
# vs
[...]

PR #1757, opened 2026-06-13
fleet: state.json non-atomic write race — atomic write_atomic + reader retry (#1750)

### LINKED ISSUE #1750: fleet: state.json non-atomic write race — readers catch partial snapshots (state-cache-lag re-fix b)
**Source:** fleet feedback triage 2026-06-13 — recurring cluster `state-cache-lag` (fix-006, originally fixed via #1136, **recurred 21×**). Part (b): the `state.json` mid-rewrite race, distinct from the blocked-by parser holes (part a).

## Problem
`scripts/fleet/fleet-state-scout` rewrites `~/.fleet/state/state.json` non-atomically, so concurrent readers (workers, reviewer, merger) catch partial or churning snapshots. Reported as a "race" at least 4 times, plus a "partial-but-stable" variant where a truncated snapshot is internally consistent enough to parse but stale/incomplete.

## Recurring evidence (post-fix)
- `2026-06-06 opus-worker-1` — "state.json mid-rewrite race churned 3x in ~90s"
- `2026-06-10 opus-worker-1` — "state.json mid-rewrite race recurred (4th instance)"
- `2026-06-10 opus-worker-1` — "state.json partial-but-stable snapshot (new variant of mid-rewrite race)"
- `2026-06-10 opus-worker-1` — "claimed feedback already handled (stale-cache race, benign)"

## Scope
- Make every `state.json` write atomic: write to a temp file in the same dir, `os.replace()` onto the target (POSIX-atomic rename). Audit all writers — `fleet-state-scout` is primary; confirm `fleet-issue`, `fleet-dispatcher`, `fleet-epic-status` don't also rewrite it non-atomically.
- Give readers a short retry-on-parse-fail (re-read once after a brief backoff) so a reader that loses the race against a rename still recovers.
- Optionally stamp the snapshot with a monotonic `written_at` so readers can detect/skip a snapshot older than their last successful read.

## Acceptance
- No partial `state.json` is ever observable: a reader either sees the prior complete snapshot or the new complete one.
- A stress test (tight write loop + concurrent reads, mirroring `tests/test_scout_degraded_fetch.py` style) shows zero partial-parse failures across N iterations.

Closes the write-race half of fix-006 (`state-cache-lag`). Companion: blocked-by parser-coverage task (part a). Prior tracking: #1136.

jakildev:
## Summary
- `write_atomic` used a fixed `<name>.tmp` temp, so two overlapping scout writers (the 30s fleet-up daemon tick racing a manual/solo-architect scout run, or an overlong tick) could `os.replace` a temp the other was still mid-writing — exposing a truncated `state.json` to a reader's `json.loads`. Now each call uses a unique `mkstemp` temp in the target's own directory (keeps `os.replace` atomic — same filesystem), fsyncs, and unlinks on error. One change covers `state.json` plus every projection slice, per-PR/issue/diff cache, and seen-hash the scout writes via `write_atomic`.
- Added `read_json_retry` for the script-side cache readers — the scout's `_prev_repos` load and `fleet-epic-status`'s `load_state_cache` — so a reader that still loses the race recovers instead of raising a spurious parse error. Kept as a small local twin in each script (they can't share a sibling module under the `~/bin` symlink resolution hazard, #1578).
- Confirmed the rest of the cache surface is already covered: the dispatcher only `stat()`s `state.json` and delegates slice reads to `fleet_task_class.py`, which already guards `OSError`/`JSONDecodeError` and defers-and-retries via the dispatcher's `|| out=""`; `fleet-claude-stream` already uses this same unique-temp + fsync + replace pattern.

## Premise correction
- The issue body's "rewrites non-atomically" premise was partly stale — the scout already wrapped writes in an atomic `os.replace`. The real bug was the **shared temp filename**, not a missing rename. Fixed accordingly.
- The "partial-but-stable / empty snapshot" evidence is the degraded-fetch mirage (#1660 / #1682) — a separate root cause with its own last-known-good fallback in `collect_state`. Deliberately not folded in.

## Noted sibling (follow-up candidate, not in this PR)
- `fleet-queue-ingest:156` writes its ingest projection via a fixed `proj_path + '.tmp'` — the same latent fixed-temp pattern. Different subsystem (not `state.json`), lower concurrency, so out of #1750's scope; flagging for a possible follow-up rather than widening this PR.

## Test plan
- [x] New `scripts/fleet/tests/test_state_json_atomic_write.py`: cross-process stress (4 multiprocessing writers + a raw reader, ~120 KB payloads) asserts zero torn/partial reads and clean writer exits; unit cases for round-trip, complete replace, error-path temp cleanup, and `read_json_retry` (valid / missing / persistently-invalid / recovers-mid-retry). Verified it fails against the old shared-temp `write_atomic` (writers crash on the shared temp).
- [x] Existing scout/projection suite green (`test_scout_degraded_fetch`, `test_worker_projection`, `test_merger_projection`, `test_enrich_*`, `test_epic_steward_projection`, `test_queue_manager_projection`, `test_opus_reviewer_projection`, `test_live_blocker_resolution`, `test_issue_queue_parser`).
- [x] `fleet-epic-status` `load_state_cache` smoke-checked against valid/missing/invalid inputs.

Closes #1750

### DIFF
--- a/scripts/fleet/fleet-epic-status
+++ b/scripts/fleet/fleet-epic-status
@@ -77,12 +77,29 @@ def _gh_json(args):
         return None
 
 
+def read_json_retry(path, attempts=3, backoff=0.05):
+    # Near-twin of fleet-state-scout's helper: retry a torn/missing cache read a
+    # few times so a reader that loses the race against the scout's atomic
+    # os.replace recovers instead of raising spuriously. Raises the last error
+    # after `attempts` tries so callers keep their fallback. Kept local rather
+    # than shared — these scripts can't import a sibling module reliably under
+    # the ~/bin symlink resolution hazard (#1578); keep in sync. (#1750)
+    last = None
+    for i in range(attempts):
+        try:
+            return json.loads(pathlib.Path(path).read_text())
+        except (json.JSONDecodeError, FileNotFoundError) as exc:
+            last = exc
+            time.sleep(backoff * (i + 1))
+    raise last
+
+
 def load_state_cache():
     """Return (data, is_fresh). is_fresh=True when mtime < CACHE_MAX_AGE_S."""
     try:
         p = pathlib.Path(STATE_JSON)
         mtime = p.stat().st_mtime
-        data = json.loads(p.read_text())
+        data = read_json_retry(STATE_JSON)
         return data, (time.time() - mtime) < CACHE_MAX_AGE_S
     except (FileNotFoundError, json.JSONDecodeError, OSError):
         return None, False
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -25,6 +25,7 @@ import re
 import signal
 import subprocess
 import sys
+import tempfile
 import threading
 import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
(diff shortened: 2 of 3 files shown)


jakildev (commented):
## Review — fleet: state.json non-atomic write race — atomic write_atomic + reader retry (#1750)

**Verdict:** approve

### Praise
- `write_atomic` fix is exactly right: the bug was a shared temp name, not a missing rename. Per-call `mkstemp` in `path.parent` (same filesystem → `os.replace` stays atomic), `fsync` for durability, and `BaseException` catch (not just `Exception`) so even `KeyboardInterrupt` triggers cleanup. The fix is minimal and precisely targeted.
- `read_json_retry` correctly handles the residual race window: even with unique temps, a slow-filesystem reader can still see `FileNotFoundError` or a decode error on the very first write. Three short retries cover this without adding meaningful latency.
- Error path test (`test_error_path_cleans_up_temp_and_keeps_old_file`) correctly patches `_mod.os.replace` rather than the global, so the mock stays scoped to the scout module's namespace — a non-obvious but correct approach.
- Cross-process stress test rather than cross-thread was the right call; the GIL would serialize enough of a `write_text` that threads rarely interleave the way processes do. The comment in the test docstring explains this, which saves the next reader from wondering why.
- The premise-correction section in the PR body ("the scout already wrapped writes in an atomic `os.replace` — the real bug was the shared temp") demonstrates good root-cause discipline.

- [x] `python3 -m pytest scripts/fleet/tests/test_state_json_atomic_write.py -v` — the cross-process stress test is the most valuable; confirm it green.
- [x] Existing scout/projection suite (already listed as green in the PR body).

PR #1758, opened 2026-06-13
docs/fleet: plan for #1746 — Metal multi-encoder GPU timestamp spanning

jakildev:
Commits the planning-step deliverable for #1746 (opus-class, `engine/render`) — `.fleet/plans/issue-1746.md`. The full plan is also posted as a comment on #1746 and staged on the planner host.

**Plan summary.** After #1738 fixed GPU-stage aggregation, Metal still undercounts per-stage GPU time on stages that issue >1 command encoder (e.g. `VOXEL_TO_TRIXEL_STAGE_1`): `writeTimestamp(START)`'s one-shot `g_nextComputeTimestampAttachment` is consumed and cleared by the *first* encoder, leaving encoders 2..N unmeasured. OpenGL's `glQueryCounter` spans the whole stage already (Metal-only bug).

The committed approach makes the attachment **sticky between START and END**: sample the start boundary on the first encoder only (`CounterDontSample` on the rest), sample the end boundary on every encoder, and clear on END — sequential GPU execution makes the last encoder's index-1 write win, giving `[first-start, last-end]` in the existing 2-sample buffer (no resize). One file: `engine/render/src/metal/metal_render_impl.cpp`. On-device Metal-validation is the acceptance gate; the only escalation trigger is the validation layer rejecting the shared-buffer multi-write.

Removes `fleet:needs-plan` from #1746 so the scout can queue it.

### DIFF
--- a/.fleet/plans/issue-1746.md
+++ b/.fleet/plans/issue-1746.md
@@ -1,0 +1,138 @@
+# Plan: render: Metal per-stage GPU timestamps undercount multi-encoder stages (only the first encoder is bracketed)
+
+- **Issue:** #1746
+- **Model:** opus
+- **Date:** 2026-06-13
+
+## Scope
+
+After #1738 fixed the GPU-stage aggregation (Max==Avg), Metal per-stage GPU
+times still grossly undercount on stages that issue **more than one command
+encoder** (e.g. `VOXEL_TO_TRIXEL_STAGE_1`'s per-axis canvas passes): the
+tracked stages sum to <1 ms against a ~57 ms frame. Make the Metal timestamp
+pair span **all** encoders of a stage, so per-stage GPU times are trustworthy
+for `/optimize` passes. OpenGL is correct already and must stay byte-for-byte
+unchanged.
+
+## Verified current state (premise confirmed in code)
+
+All in `engine/render/src/metal/metal_render_impl.cpp`:
+
+- `MetalTimestampSampleAttachment g_nextComputeTimestampAttachment;` (line 33)
+  is a **one-shot**: `writeTimestamp(START)` (line 785) sets it
+  `{sampleBuffer, startIdx=0, endIdx=1}`; the **first** encoder created after
+  START consumes it and resets it to `{}`.
+- `createComputeEncoder` (lines 114–132): if the attachment is set, attaches
+  `setStartOfEncoderSampleIndex(0)` + `setEndOfEncoderSampleIndex(1)` to **one**
+  compute encoder, then `g_nextComputeTimestampAttachment = {}` (line 130).
+- `attachTimestampSamples` (lines 134–144): same for a render encoder via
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — docs/fleet: plan for #1746 — Metal multi-encoder GPU timestamp spanning

**Verdict:** approve

### Checklist

- ECS / render / lighting / math / serialization / naming: N/A — docs-only, single plan markdown added at `.fleet/plans/issue-1746.md`
- Fleet scripts: no script identifiers changed; no comment drift possible (plan is new file only)

### Premises verified against live code

- `MetalTimestampSampleAttachment` struct at lines 27–31 of `metal_render_impl.cpp` — confirmed; no `firstEncoder_` field yet (plan adds it) ✅
- `g_nextComputeTimestampAttachment` global at line 33 ✅
- `createComputeEncoder` lines 114–132: attaches once, then `g_nextComputeTimestampAttachment = {}` at line 130 — confirmed one-shot behavior ✅
- `attachTimestampSamples` lines 134–144: same one-shot reset at line 143 ✅
- `writeTimestamp(START)` sets attachment at line 786; `writeTimestamp(END)` only sets `pair->hasEnd_ = true` at line 794 — confirmed END does not currently clear the attachment ✅
- `setSampleCount(2)` at line 715 — confirmed; fix stays within the existing 2-sample buffer ✅
- `readTimestampPairMs` resolves `Range::Make(0, 2)` at line 807, returns `samples[1] - samples[0]` — unchanged by the plan ✅

### Technical soundness

The "sticky + last-write-wins" approach is correct: Metal executes encoders in a command buffer sequentially on GPU, so the last encoder's end-sample write to index 1 reliably wins. The critical invariant — that `writeTimestamp(END)` must explicitly clear the sticky state — is step 5 of the plan and is the key missing piece in the current code (line 793 does not clear the attachment today). The `CounterDontSample` sentinel for non-first-encoder start boundaries is the correct Metal API pattern.

The contingency path (resize to `2 × maxEncodersPerStage` if validation rejects shared-buffer multi-write) is correctly scoped as a fleet:design-blocked escalation rather than a pre-built fallback — the right call, since the shared-buffer approach needs no buffer resize and should work given sequential execution.

### Praise

- Identifying the "END doesn't clear the sticky attachment" step (plan step 5) is the subtle correctness detail that makes the fix safe — without it, encoders after END in the same frame would be spuriously tagged. Calling this out explicitly saves a round-trip.
- The single-encoder regression note (gotcha 3: confirm a known single-encoder stage's reported time is unchanged) is exactly the right regression test; it's easy to accidentally change single-encoder semantics while chasing the multi-encoder case.
- The mixed compute+render encoder note (gotcha 4: both types share one `firstEncoder_` flag) prevents a "fix compute but break render encoders" path.

- [x] Docs-only — no build required (PR body confirms no engine code changes in this PR)

Opus recheck not required.

PR #1761, opened 2026-06-13
render: depth-only shadow-feeder path — skip stage 2 for off-screen feeders (#1740 item 1)

### LINKED ISSUE #1740: render: cardinal FULL-mode zoom cliff — feeder depth-only path + dense-grid overdraw
Follow-up to PR #1737. After the canvas-coverage cull clamp, cardinal FULL mode still scales with zoom on IRPerfGrid 64³: 7.8 ms (zoom 1) → 11.2 (2) → 18.9 (4) → 26.1 ms (zoom 8). Two compounding causes:

1. Shadow feeders inside the canvas coverage still pay the full sub² micro-cell raster, but the sun-shadow bake only needs base-resolution depth (catalog entry 7's remaining half — the depth-only feeder path; do not break the lighting culling invariants in engine/render/CLAUDE.md).
2. Dense-grid depth overdraw: perf_grid voxel_set mode is 262k separate 1³ entities, so the exposure gate cannot cull interior faces and every view ray crosses ~50 voxels of imageAtomicMin taps (~23M taps × 2 stages at zoom 8 for ~0.46M canvas texels). Candidate: load-test-skip before imageAtomicMin in stage 1 (and the per-axis store) — cheap read first, atomic only when the candidate is nearer. Needs measuring on Metal where image atomics route through the scratch-buffer indirection.

Also worth a look while in there: `UPDATE` pipeline costs ~5 ms/frame at 262k entities (PropagateTransform 2.3 + UpdateVoxelSetChildren 1.2 + PeriodicIdle* 1.2) — relevant to the 100 fps goal at zoom 1 where everything is visible.

Baseline numbers + methodology: PR #1737 (24-cell matrix with the new --yaws axis).

jakildev:
Part of #1740 (item 1 — the depth-only shadow-feeder path).

## What
Shadow-feeder voxels — those whose cardinal iso position lies inside the sun-shadow-widened cull (`cullIsoMin/Max`) but **outside** the un-widened visible viewport — exist only to cast sun shadows onto on-screen pixels via the stage-1 distance bake. They are never displayed, lit, or picked, so `VOXEL_TO_TRIXEL_STAGE_2`'s colour + entity-id taps are pure waste for them.

This skips stage 2 for feeders, keeping stage 1 (depth) full-resolution. New `FrameDataVoxelToCanvas::visibleIsoBounds_` (ivec4 @ std140 offset 176) carries the un-widened viewport; the shader early-returns when a voxel's cardinal iso pos falls outside it. Gated to the cardinal single-canvas world path (`perAxisRoute==0` / `residualYaw==0` / `!isDetachedCanvas`).

## Why it's byte-identical in rendered output
- Feeders are **off-screen** — their canvas colour/id pixels are never composited to the framebuffer, never lit (lighting reads own-pixel colour of on-screen pixels), never picked (entity-id is read only under the on-screen cursor).
- **Stage 1 (depth) is untouched.** The sun-shadow bake (`c_bake_sun_shadow_map`) and AO (`c_compute_voxel_ao`) read **only** `trixelDistances` — confirmed in source — so shadows + AO are bit-for-bit unchanged.
- A **+4-iso-px margin** on `visibleIsoBounds_` (the existing `kGpuMargin`) covers the voxel face footprint, so no on-screen pixel is ever classified as a feeder.
- When sun shadows are off, the cull isn't widened → `visibleIsoBounds_ == cullIsoMin/Max` → nothing is skipped.

## Verification (Metal / macOS)
- **shape_debug: 28/28 shots byte-identical** vs `origin/master` (normal voxel path — no shadows — fully unregressed; exercises the struct/UBO/Metal-mirror changes).
- **perf_grid (sun shadows on, 64³, zoomed so the grid extends off-screen → feeders present): structurally identical to master with clean screen edges and no feeder artifacts.** (perf_grid is non-deterministic — an animated wave — so a byte-compare is noise; verified visually instead.)

GLSL + Metal mirrors kept in lockstep; std140 layout asserts updated (`sizeof == 192`, `offsetof(visibleIsoBounds_) == 176`).

## ⚠️ Base / perf-validation caveat — please read before merging
#1740 is a **follow-up to #1737** (\"canvas-coverage cull clamp\"), and its framing (\"feeders **inside the canvas coverage**\") assumes that clamp. **#1737 is not in master** — it merged into the `fix/perfgrid-zoom-after-rotation` integration branch (no open PR to master). This change is **orthogonal to and correct independent of** the clamp (it skips stage 2 for *all* feeders), but:

- **On master the targeted feeders are not the render bottleneck.** Measured render-scope time is unchanged master-vs-branch (zoom 8: 32.8 vs 33.5 ms — within noise on a contended host). The win surfaces on the **#1737 base**, where the clamp removes the outside-coverage feeders and the remaining inside-coverage feeders' stage-2 cost dominates — exactly the case the issue describes.
- **Per-stage GPU time can't be isolated** on Metal: `voxelStage2` reads 0.000 (the open #1746 multi-encoder undercount).

So this is presented as a **byte-identical, zero-risk work-removal** that implements item 1's structural lever and **composes with #1737**, *not* as a measured frame-time win. **Recommendation:** merge as cleanup that lands ahead of #1737, **or** hold / rebase onto the `fix/perfgrid-zoom-after-rotation` base and re-measure there. Deliberately does **not** `Closes #1740` (the perf goal is unvalidated on master).

### DIFF
--- a/engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp
+++ b/engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp
@@ -575,6 +575,17 @@ template <> struct System<VOXEL_TO_TRIXEL_STAGE_1> {
             IRPrefab::SunShadow::shadowFeederCullViewport(kGpuMargin, shadowFeederParams_);
         frameData_.cullIsoMin_ = ivec2(IRMath::floor(gpuVp.min_));
         frameData_.cullIsoMax_ = ivec2(IRMath::ceil(gpuVp.max_));
+        // Depth-only shadow-feeder path (#1740): the SAME viewport at the same
+        // margin BEFORE shadowFeederCullViewport widened it toward the sun. A
+        // voxel inside gpuVp (it passed the compact cull) but outside this is an
+        // off-screen shadow feeder — stage 2 skips its colour/entity-id taps.
+        // When sun shadows are off the sweep is zero, so gpuVp == this and stage
+        // 2 skips nothing (byte-identical). Same getCullViewport() the widened
+        // form reads above, so the two boxes are derived consistently.
+        const IsoBounds2D visibleVp = IRRender::getCullViewport().isoViewport(kGpuMargin);
+        frameData_.visibleIsoBounds_ = ivec4(
+            ivec2(IRMath::floor(visibleVp.min_)), ivec2(IRMath::ceil(visibleVp.max_))
+        );
         frameDataBuf_->subData(0, sizeof(FrameDataVoxelToCanvas), &frameData_);
 
         // Voxel positions are uploaded via the pending-range queue populated by
--- a/engine/render/include/irreden/render/ir_render_types.hpp
+++ b/engine/render/include/irreden/render/ir_render_types.hpp
@@ -285,6 +285,21 @@ struct FrameDataVoxelToCanvas {
     // offset is unchanged and shaders that read only the prefix need no update;
     // only c_lighting_to_trixel declares it.
     vec4 detachedWorldReceive_ = vec4(0.0f, 0.0f, 0.0f, 0.0f);
+    // Un-widened (no shadow-feeder sweep) iso cull viewport for the depth-only
+    // shadow-feeder path (#1740). `.xy` = floor(min), `.zw` = ceil(max) of the
+    // SAME viewport `cullIsoMin_/cullIsoMax_` derive from, but BEFORE
+    // IRMath::shadowFeederIsoBounds widens it toward the sun. A voxel whose
+    // cardinal iso position lies inside [cullIsoMin_, cullIsoMax_] (it passed
+    // the compact cull) but OUTSIDE this box is an off-screen shadow FEEDER —
+    // it only casts sun shadows via the stage-1 distance bake and is never
+    // displayed / lit / picked, so c_voxel_to_trixel_stage_2 skips its
+    // colour + entity-id taps (stage 1 still writes its full-resolution depth,
+    // which is all the bake + AO read). When sun shadows are off the sweep is
+    // zero, so this equals cullIsoMin_/cullIsoMax_ and stage 2 skips nothing —
+    // byte-identical. Read only by c_voxel_to_trixel_stage_2; std140-appended
+    // after detachedWorldReceive_ (offset 176) so every prior offset is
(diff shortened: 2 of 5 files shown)


jakildev (commented):
## Review — render: depth-only shadow-feeder path — skip stage 2 for off-screen feeders (#1740 item 1)

**Verdict:** approve

### Praise
- **std140 layout is fully guarded.** Both `offsetof(visibleIsoBounds_) == 176` and `sizeof(FrameDataVoxelToCanvas) == 192` are static_asserts — a drift would be a compile error, not a silent mis-upload.
- **GLSL and Metal kept in exact sync.** The `_detachedWorldReceivePad` bridge in the GLSL block, the `int4 visibleIsoBounds` append in `ir_iso_common.metal`, and the feeder detection bodies are structurally identical across backends.
- **Conservative gating prevents regressions.** The feeder early-return is guarded by `residualYaw == 0.0 && isDetachedCanvas < 0.5`, so the per-axis, smooth-yaw, and detached paths are unconditionally unaffected. Re-voxelize solids are implicitly excluded via `isDetachedCanvas`.
- **The `_detachedWorldReceivePad` padding strategy is correct.** Declaring a pad field rather than silently relying on the UBO prefix is the sound approach — it makes the offset derivation explicit and means stage 2's block now covers the full 192-byte struct, eliminating any future "why does stage 2 declare fewer bytes than the CPU struct" confusion.
- **Byte-identical claim is solid.** Stage 1 still writes full-resolution depth for feeders; the AO and sun-shadow bakes read only `trixelDistances` (confirmed in source); the +4-iso-px margin covers the voxel face footprint; and the shadows-off degenerate case falls through cleanly since `visibleIsoBounds == cullIsoMin/Max` when the sweep is zero.

- [ ] Linux/OpenGL build + `IRShapeDebug --auto-screenshot 10` byte-comparison vs master (not yet verified; `fleet:needs-linux-smoke` will gate this)
- [ ] `perf_grid` with sun shadows enabled on Linux: confirm no feeder artifacts at large zoom factors where feeders are present

Opus recheck not required.

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). Build clean, 21/21 screenshots saved, no errors or warnings. — worker

PR #1764, opened 2026-06-13
render: dense-grid overdraw — re-profile + UPDATE pipeline costs (#1740)

### LINKED ISSUE #1740: render: cardinal FULL-mode zoom cliff — feeder depth-only path + dense-grid overdraw

jakildev:
## Summary

Closes #1740 item 2 (UPDATE pipeline costs) after item 1 (depth-only feeder path) shipped in PR #1761.

### Profiling findings (zoom=8, 262k entities, macos-debug)

With per-stage GPU timestamps from #1746 we got a clean baseline:

| Scenario | frame | update | render |
|---|---|---|---|
| Before (static, wave=0) | 50.3ms | 17.2ms | 34.2ms |
| Before (animated) | 51.9ms | 25.3ms | 33.1ms |
| **After (static)** | **42.9ms** | **14.7ms** | **33.1ms** |
| **After (animated)** | **42.0ms** | **14.6ms** | **32.7ms** |

### Root cause

`C_VoxelSetNew` carried `lastParentPosition_` / `hasLastParentPosition_` —
a snapshot-compare-and-early-return in `updateAsChild`. This is the
dirty-flag-in-disguise pattern explicitly forbidden by
`.claude/rules/cpp-ecs.md §"A snapshot-compare-and-early-return is a dirty
flag in disguise"`. The same fix was applied to `REBUILD_GRID_VOXELS` in
#1288.

### Fix

Removed both fields + the early-return from `updateAsChild`. Replaced with a
visibility cull gate in `UPDATE_VOXEL_SET_CHILDREN::beginTick`/`tick`:
`C_VoxelPool::isRangeVisible` against the previous render frame's camera
viewport — same pattern as `system_rebuild_grid_voxels.hpp`.

Off-screen entities skip position computation entirely. On-screen entities
re-upload every frame (correct ECS behavior). At zoom=1 (120 FPS), the
majority of the 64³ grid is off-screen; cull gate fires for most entities
and UPDATE drops to 5.3ms.

### Remaining UPDATE cost

14.7ms at zoom=8 with 262k entities comes from:
- `isRangeVisible` check per entity (O(chunks per set))
- `PROPAGATE_TRANSFORM` serial composeNode pass (262k entities, single
  archetype = n=1 parallelFor task — no inter-archetype parallelism)

Making either PARALLEL_FOR requires thread-safety work on shared pool state
(`queuePositionRange`, `markChunkWorldBoundsDirty` use non-atomic containers).
Filed as follow-up; out of scope for this task.

### Profiling 0ms at zoom=1 (explained)

Previously `update:0.000ms` at zoom=1 was a sampling artifact: at 97 FPS
with a 60 Hz fixed-step UPDATE, ~38% of render frames have no UPDATE tick.
The auto-profile captured a no-UPDATE frame. With the cull gate reducing
UPDATE cost, the sampling window is more stable and the artifact is no
longer observed in the test runs (5.3ms captured).

### DIFF
--- a/engine/prefabs/irreden/voxel/CLAUDE.md
+++ b/engine/prefabs/irreden/voxel/CLAUDE.md
@@ -69,11 +69,12 @@ for single voxels and particles.
   Translate-only path: voxels move with the entity's
   `C_WorldTransform.translation_` (composed by `PROPAGATE_TRANSFORM` from
   `C_LocalTransform` + parent chain + `TRANSFORM_TRANSLATION` modifiers)
-  but no per-set rotation/scale composition. Voxel sets whose translation
-  is unchanged from the prior tick early-out of `updateAsChild` and
-  contribute nothing to the per-pool GPU position queue
-  (`C_VoxelPool::queuePositionRange`) — a static voxel scene pays zero
-  CPU→GPU position bytes/frame.
+  but no per-set rotation/scale composition. On-screen sets re-upload every
+  frame (unconditional per ECS no-dirty-flags rule). Off-screen sets are
+  skipped via `C_VoxelPool::isRangeVisible` against the previous frame's
+  render viewport — same cull gate as `REBUILD_GRID_VOXELS` (#1288). A set
+  entirely off-screen pays nothing; a visible set in a static scene
+  re-uploads its positions every frame.
 - `REBUILD_GRID_VOXELS` (UPDATE pipeline, T-294; inverse re-voxelize #1720)
   — Epic C C6. Runs AFTER `UPDATE_VOXEL_SET_CHILDREN`. Re-rasterizes
   GRID-mode entities (entities carrying `C_RotationMode::GRID`, the
--- a/engine/prefabs/irreden/voxel/CLAUDE.md
+++ b/engine/prefabs/irreden/voxel/CLAUDE.md
@@ -72,9 +72,9 @@ for single voxels and particles.
   but no per-set rotation/scale composition. On-screen sets re-upload every
   frame (unconditional per ECS no-dirty-flags rule). Off-screen sets are
   skipped via `C_VoxelPool::isRangeVisible` against the previous frame's
-  render viewport — same cull gate as `REBUILD_GRID_VOXELS` (#1288). A set
-  entirely off-screen pays nothing; a visible set in a static scene
-  re-uploads its positions every frame.
+  shadow-feeder-expanded cull viewport — same cull gate as `REBUILD_GRID_VOXELS`
+  (#1288). A set entirely outside that viewport pays nothing; a visible set in
+  a static scene re-uploads its positions every frame.
 - `REBUILD_GRID_VOXELS` (UPDATE pipeline, T-294; inverse re-voxelize #1720)
   — Epic C C6. Runs AFTER `UPDATE_VOXEL_SET_CHILDREN`. Re-rasterizes
   GRID-mode entities (entities carrying `C_RotationMode::GRID`, the
--- a/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
@@ -430,7 +430,8 @@ struct C_VoxelSetNew {
     // stable because they are plain scalars on this struct.
     // Returns the number of positions actually written (safeCount), or 0 if
     // the pool-bounds guard fires (numVoxels_ exceeds available pool slots).
-    // The caller gates visibility — call only when the set is on-screen.
+    // The caller gates visibility — call only when the set is within the
+    // shadow-feeder cull viewport (see system_update_voxel_set_children.hpp).
     // Callers use the return value to queue exactly the written range for
     // GPU upload; avoids queuing stale tail slots on bounds-guard overflow.
     int updateAsChild(
--- a/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
+++ b/engine/prefabs/irreden/voxel/components/component_voxel_set.hpp
@@ -27,8 +27,6 @@ struct C_VoxelSetNew {
     IREntity::EntityId canvasEntity_ = IREntity::kNullEntity;
     // TODO: Evaulate if we should store here or somewhere else.
     IREntity::EntityId ownerEntityId_ = IREntity::kNullEntity;
-    vec3 lastParentPosition_ = vec3(0.0f);
-    bool hasLastParentPosition_ = false;
 
     // Index into the canvas voxel pool's underlying arrays. Captured at
     // allocation time so consumers never recompute it from a pointer-diff
@@ -431,20 +429,16 @@ struct C_VoxelSetNew {
     // after such a migration; only `voxelStartIdx_` and `numVoxels_` remain
     // stable because they are plain scalars on this struct.
     // Returns the number of positions actually written (safeCount), or 0 if
-    // the parent position is unchanged (early-out). Callers use the return
-    // value to queue exactly the written range for GPU upload — avoids
-    // queuing stale slots in the rare case the pool-bounds guard fires.
+    // the pool-bounds guard fires (numVoxels_ exceeds available pool slots).
+    // The caller gates visibility — call only when the set is on-screen.
+    // Callers use the return value to queue exactly the written range for
+    // GPU upload; avoids queuing stale tail slots on bounds-guard overflow.
     int updateAsChild(
         vec3 parentPosition,
         std::vector<IRRender::VoxelGpuPosition> &poolGlobalsOut,
         const std::vector<IRRender::VoxelGpuPosition> &poolPositions,
         const std::vector<vec3> &poolOffsets
     ) {
-        if (hasLastParentPosition_ && parentPosition == lastParentPosition_) {
-            return 0;
-        }
-        lastParentPosition_ = parentPosition;
-        hasLastParentPosition_ = true;
         const size_t writableTail =
             poolGlobalsOut.size() > voxelStartIdx_ ? poolGlobalsOut.size() - voxelStartIdx_ : 0u;
         const size_t availPositions =
--- a/engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp
+++ b/engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp
@@ -30,7 +30,9 @@ template <> struct System<UPDATE_VOXEL_SET_CHILDREN> {
         lastCanvas_ = IREntity::kNullEntity;
         lastPool_ = nullptr;
 
-        // Mirrors the visibility gate in system_rebuild_grid_voxels.hpp.
+        // Mirrors the on-screen cull gate in system_rebuild_grid_voxels.hpp;
+        // shadow-feeder-expanded viewport not needed here as position updates
+        // for off-screen sets are not required for rendering.
         // getCullViewport() holds last render frame's snapshot when called
         // from the UPDATE pipeline.
         const IRRender::CullViewportState &cull = IRRender::getCullViewport();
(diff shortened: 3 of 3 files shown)


jakildev (commented):
## Review — render: dense-grid overdraw — re-profile + UPDATE pipeline costs (#1740)

**Verdict:** approve

### Praise

- **Textbook dirty-flag removal.** `lastParentPosition_` / `hasLastParentPosition_` were a snapshot-compare-and-early-return — exactly the pattern `cpp-ecs.md §"A snapshot-compare-and-early-return is a dirty flag in disguise"` forbids (same root cause as the REBUILD_GRID_VOXELS fix in #1288). The replacement — unconditional update for on-screen sets, cull-gate skip for off-screen — is the honest ECS answer to "is this work observable?"
- **Pattern consistency.** `beginTick` captures `getCullViewport()` → `isoViewport(kCullChunkMargin)` once per frame; `tick` gates on `cullValid_ && !pool.isRangeVisible(...)`. This mirrors the established `System<REBUILD_GRID_VOXELS>` shape — both the field names (`cullValid_`, per-canvas `lastPool_` cache) and the fail-safe (`cullValid_ = false` until the canvas has a non-zero size prevents startup geometry drops).
- **`updateAsChild` comment update.** The new doc accurately reflects caller responsibility ("The caller gates visibility — call only when the set is on-screen") and drops the now-false "returns 0 if the parent position is unchanged" claim.
- **CLAUDE.md updated consistently.** The voxel prefab doc now correctly states "On-screen sets re-upload every frame (unconditional per ECS no-dirty-flags rule). Off-screen sets are skipped via `C_VoxelPool::isRangeVisible`..."

### Nits

- `system_update_voxel_set_children.hpp:33-40` — The comment says "Mirrors the visibility gate in `system_rebuild_grid_voxels.hpp`" and the PR body says "identical pattern." The patterns are close but not identical: `REBUILD_GRID_VOXELS` uses `IRPrefab::SunShadow::shadowFeederCullViewport(...)` (sun-shadow-expanded viewport) while this PR uses `cull.isoViewport(kCullChunkMargin)` (render viewport only). The difference: moving off-screen entities whose pool chunks fall inside the shadow-feeder cone but outside the render viewport will have stale positions in `poolGlobalsOut` after this change. For the optimization target (static or wave-animated dense grid, 262k entities at zoom=8) this is fine — static sets never move, so positions can't be stale. But the "identical" claim in both the comment and PR body is imprecise and could mislead a future reader who wonders why shadow artifacts appear when an animated entity pans off-screen. Consider adjusting the comment to "Mirrors the on-screen cull gate in `system_rebuild_grid_voxels.hpp`; shadow-feeder-expanded viewport not needed here as position updates for off-screen sets are not required for rendering" (or a shorter version), and drop "identical" from the PR body.

- [x] Profiling numbers in PR body (zoom=8, 262k entities, macOS-debug): update=17.2ms → 14.7ms static / 25.3ms → 14.6ms animated. These are present and plausible.
- [ ] Smoke at zoom=1: confirm UPDATE cost is ~5.3ms and not 0ms (sampling artifact explained in PR body).
- [ ] Confirm no geometry vanishes at startup (before first render frame, `cullValid_ = false` → positions upload unconditionally — verify with a quick IRPerfGrid run).

Opus recheck not required.

jakildev:
Addressed nit: expanded the cull-gate comment to note that the shadow-feeder-expanded viewport is not used here (position updates for off-screen sets are not required for rendering), preventing confusion about potential shadow artifacts when entities pan off-screen. Dropped "identical" from the PR body. — worker

jakildev (commented):
## Opus recheck — render: dense-grid overdraw — re-profile + UPDATE pipeline costs (#1740)

**Verdict:** needs-fix

Picked up as a core render/ECS-tick PR (Sonnet stamped "Opus recheck not required"). The dirty-flag removal is correct and well-executed — but on a close read of the data-flow three systems deep, the cull viewport this PR uses is the wrong one, and it silently regresses cast shadows for off-screen moving voxel sets. The fix is one line.

### The issue — cull gate uses the render viewport, not the shadow-feeder viewport

`System<UPDATE_VOXEL_SET_CHILDREN>::beginTick` gates with the plain **render** viewport:

```cpp
viewport_ = cull.isoViewport(IRRender::kCullChunkMargin);   // render viewport only
```

Sonnet flagged this exact difference as a *nit* and dismissed it ("for the optimization target — static sets never move — this is fine"). On closer read I disagree and upgrade it to needs-fix, because the system is generic (every translate-only `C_VoxelSetNew` child, not just the static dense grid) and the consequence is a silent correctness regression, not a cosmetic imprecision:

1. **The positions this system writes feed the sun-shadow bake.** `system_voxel_to_trixel.hpp:590` is explicit: *"Voxel positions are uploaded via the pending-range queue populated by `UPDATE_VOXEL_SET_CHILDREN`."* The shadow feeder reads those `pool.getPositionGlobals()` over the **sun-expanded** `shadowFeederCullViewport` (`system_voxel_to_trixel.hpp:539`, `:575`), and #1740's own depth-only path (lines 575-588) is built around the "off-screen shadow feeder" annulus — voxels inside the sun-widened `gpuVp` but outside the render `visibleVp`.

2. **The render viewport excludes that annulus.** A translate-only voxel set whose chunks sit in the shadow-feeder annulus (off render-screen, but casting into the visible scene) is skipped by this PR's gate. If that set has **moved**, its `poolGlobals` are now stale when the shadow bake reads them → its cast shadow freezes at the old position. On master this could not happen: `updateAsChild`'s old early-out only fired when `parentPosition == lastParentPosition_`, so a moving off-screen set always updated. This PR removes that and replaces it with a viewport that's too narrow for the shadow path — a net regression for moving off-screen casters.

3. **The sibling system that writes the same buffer already does it right.** `System<REBUILD_GRID_VOXELS>` (GRID-mode voxel sets, same shadow-casting role, same `pool.getPositionGlobals()` target) gates on the **expanded** viewport for precisely this reason (`system_rebuild_grid_voxels.hpp:147`):

```cpp
chunkVp_ = IRPrefab::SunShadow::shadowFeederCullViewport(
    IRRender::kCullChunkMargin,
    IRPrefab::SunShadow::frameShadowFeederParams(),
    cull);
```

The PR comment and body claim this system "mirrors" / is the "identical pattern" to `REBUILD_GRID_VOXELS` — but in the one detail that governs shadow correctness, it does not.

### Fix (one line + one include)

In `system_update_voxel_set_children.hpp::beginTick`, replace the render-viewport derivation with the shadow-feeder form, matching the sibling exactly:

```cpp
viewport_ = IRPrefab::SunShadow::shadowFeederCullViewport(
    IRRender::kCullChunkMargin,
    IRPrefab::SunShadow::frameShadowFeederParams(),
    cull);
```

and add the include the sibling carries (this PR only added `cull_viewport_state.hpp`):

```cpp
#include <irreden/render/sun_shadow_constants.hpp>
```

This is strictly safer than the current code: `shadowFeederCullViewport` only ever **widens** the cull region toward the sun, so it cannot drop geometry that the render-viewport gate would have kept — it just stops dropping the off-screen shadow casters. The extra cost is a thin annulus of position updates, exactly the perf/correctness tradeoff `REBUILD_GRID_VOXELS` already accepts. Then please re-sync the "mirrors REBUILD_GRID" comment and the PR body so the parity claim is true.

### Everything else — confirmed clean

- **Dirty-flag removal is correct.** `lastParentPosition_` / `hasLastParentPosition_` were a snapshot-compare-and-early-return; removing them and gating on observability (visibility) is the right ECS answer. Sonnet's praise here holds.
- **`numVoxels_ <= 0` guard, `cullValid_` fail-safe, exact-`writtenCount` queue range** — all correct; the startup `cullValid_ == false` path uploads unconditionally so nothing drops before the first render frame.
- **CLAUDE.md / `updateAsChild` doc updates** are accurate (modulo the same "same cull gate as REBUILD_GRID" line, which becomes true once the viewport above is fixed).

Once the viewport is switched to `shadowFeederCullViewport`, this is good to merge.

jakildev:
Addressed feedback: switched the cull viewport in `System<UPDATE_VOXEL_SET_CHILDREN>::beginTick` from the plain render viewport (`isoViewport(kCullChunkMargin)`) to the shadow-feeder-expanded viewport (`IRPrefab::SunShadow::shadowFeederCullViewport`). Now mirrors `system_rebuild_grid_voxels.hpp` exactly — off-screen voxel sets within shadow-throw distance still receive position updates so the shadow stage sees current positions. Added `#include <irreden/render/sun_shadow_constants.hpp>` and updated the comment to accurately describe the shared cull gate. Build verified: `fleet-build --target IRShapeDebug` exits 0.

jakildev (commented):
## Re-review — render: dense-grid overdraw — re-profile + UPDATE pipeline costs (#1740)

**Verdict:** approve

Re-reviewing after author addressed Opus needs-fix (viewport mismatch).

### Opus finding resolved

Opus flagged `System<UPDATE_VOXEL_SET_CHILDREN>::beginTick` using `cull.isoViewport(kCullChunkMargin)` (render-only viewport) instead of the shadow-feeder-expanded viewport, which would leave stale positions in `poolGlobalsOut` for moving off-screen voxel sets that are shadow casters.

The fix in `system_update_voxel_set_children.hpp:33-40` correctly switches to:

```cpp
viewport_ = IRPrefab::SunShadow::shadowFeederCullViewport(
    IRRender::kCullChunkMargin,
    IRPrefab::SunShadow::frameShadowFeederParams(),
    cull
);
```

This is exactly what Opus requested — identical to the `REBUILD_GRID_VOXELS` gate. Off-screen casters within the shadow-feeder expansion margin continue to upload their positions and can cast shadows correctly. The includes (`cull_viewport_state.hpp`, `sun_shadow_constants.hpp`) are the right headers for these calls.

The in-code comment is also now accurate:
> "Mirrors the cull gate in system_rebuild_grid_voxels.hpp — same shadow-feeder-expanded viewport so off-screen casters whose position update is required for shadow projection still upload."

### Nits

- `CLAUDE.md` line 72 — "Off-screen sets are skipped via `C_VoxelPool::isRangeVisible` against the previous frame's **render viewport**" — should be "shadow-feeder-expanded cull viewport" (same cull gate as `REBUILD_GRID_VOXELS`). The "same cull gate as REBUILD_GRID_VOXELS (#1288)" qualifier at the end is correct and sufficient, but "render viewport" contradicts both the code and the qualifier.
- `component_voxel_set.hpp:433` (updateAsChild comment) — "call only when the set is on-screen" — since the gate is shadow-feeder-expanded (wider than on-screen), the accurate phrase is "within the shadow-feeder cull viewport." Minor — the cross-reference to the caller making this decision is the important part.

Both nits are documentation-only; no code correctness concern.

Opus recheck not required.

jakildev:
Addressed nits from re-review:
- `voxel/CLAUDE.md`: changed "render viewport" → "shadow-feeder-expanded cull viewport" in UPDATE_VOXEL_SET_CHILDREN description (line 74-75)
- `component_voxel_set.hpp:433`: updated updateAsChild comment from "on-screen" → "within the shadow-feeder cull viewport" to match the actual gate

Doc-only; no code change. — worker

Issue #1717, opened 2026-06-11
render: shadow & lighting quality epic — step self-shadow banding, yaw-frozen shadows, interior gaps, wrong-face hits


jakildev:
User-observed (2026-06-11, macOS/Metal, IRCanvasStress + IRShapeDebug current master) shadow/lighting quality inventory. Filed as the tracking epic for the lighting /goal; children below are independently fixable.

## Symptom inventory

1. **Step self-shadow banding on rotating voxel solids** (child issue: see "render: sun-shadow step banding"). Rotating GRID and DETACHED_REVOXELIZE solids render with venetian-blind slats / rows of near-black trixels. Discriminator: `IRCanvasStress --no-lighting` renders the same solids clean — the "missing voxels/faces" read is **lighting**, not geometry. Round-to-cell rotation produces 1-cell surface steps; under the grazing sun each step row falls into (geometrically honest, but visually wrong) shadow, resolved per-trixel. Matches the user observation that the re-voxelized cyan cube's shadowing reads per-trixel, not per-whole-voxel.
2. **Cast shadows frozen within a camera-yaw bracket** (child issue: see "render: sun shadow frozen"). Floor shadows update only at 90° rebrackets under continuous camera yaw, and don't scale/deform with the residual-yaw squash-stretch the receiving floor gets.
3. **Interior gaps / streaks inside cast shadows** at various zooms (IRShapeDebug: teal cubes' floor shadows show horizontal streak holes; user screenshots show the same at multiple zooms).
4. **Ragged / speckled shadow silhouettes** — shadow edges don't adapt the trixel shape/angle, so a smooth voxel cube casts a noisy blob instead of a clean iso-shaped shadow.
5. **Shadows landing on faces they logically shouldn't** (IRShapeDebug front teal cube: full-black side face) and **shadows poking through entities**.

## Consolidation note (user direction)

Items 1, 3, 4 plausibly share machinery with the rotation round-to-cell stepping: the same cell-quantization that creates rotation gaps creates stepped casters and stepped receivers. A shared "cell-step tolerance" in the sun-shadow sample (receiver-side depth tolerance ≈ 1 cell, slope-aware) plus caster-resolution choices should be designed once, not per-symptom. #444 (normal/slope bias) and #445 (PCF) landed earlier; verify the detached receive path (P4b-2 `ir_sun_shadow_sample` callers) applies the same bias the main path does before adding new knobs.

## Repro

[...]

PR #1765, opened 2026-06-13
render: structural shadow-hole metric for deterministic validation

jakildev:
## Summary
First brick of a deterministic render-validation framework (re: the shadow/lighting quality work tracked by #1717). Adds `scripts/render-shadow-metric.py` — a **structural** cast-shadow metric that complements `render-compare.py`'s pixel-diff:

- Classifies shadow occupancy in a **SHADOW debug-overlay** capture (magenta = shadowed, black = lit; `engine/render/CLAUDE.md`) and reports `hole_ratio` (lit pixels inside the shadow region) + connected-component `fragmentation` — the exact signature of the swiss-cheese / cross-hatch failure mode (#1717 items 3-4).
- **Zoom-stable:** measures structure, not exact pixels, so it's robust to the CPU↔GPU float divergence that forces `render-compare` to *exclude* high-zoom shots (canvas_stress's `revoxelize_solids_zoom`).
- **Backend-agnostic:** one threshold is a shared oracle across Metal and OpenGL (no per-backend reference set).

### Why this matters
Today a merged shadow fix can't be confirmed to hold without a manual eyeball recapture, and zoom/shadow conditions are unguarded by the pixel-diff gate. A structural occupancy metric turns "is the shadow swiss-cheese?" into a deterministic pass/fail. **Requires the demo to capture in SHADOW debug-overlay mode** (so shadows are magenta-classifiable) — wiring that shot into the demo shot lists + the `render-verify` harness is the next step (framework epic).

## Test plan
- [x] `python3 -m unittest scripts.tests.test_render_shadow_metric` — 7/7: determinism, solid→0 holes/1 component, checkerboard→~50% holes/100+ components, interior-hole ring, ROI restriction, gray-bg ignored.
- [x] Parses real 2560×1440 engine PNGs deterministically.
- [x] `py_compile` clean; pure stdlib (reuses `read_png` from `render-compare.py`).

Refs #1717.

### DIFF
--- a/scripts/render-shadow-metric.py
+++ b/scripts/render-shadow-metric.py
@@ -197,9 +197,16 @@ def _main(argv: list[str]) -> int:
     failed = []
     if args.max_hole_ratio is not None and m["hole_ratio"] > args.max_hole_ratio:
         failed.append(f"hole_ratio {m['hole_ratio']} > {args.max_hole_ratio}")
-    if (args.max_components is not None and m.get("components") is not None
-            and m["components"] > args.max_components):
-        failed.append(f"components {m['components']} > {args.max_components}")
+    if args.max_components is not None:
+        if m.get("components") is None:
+            print(
+                "warning: --max-components requested but component count was "
+                "skipped (roi too large; see components_note in output)",
+                file=sys.stderr,
+            )
+            failed.append("max-components check skipped (roi too large)")
+        elif m["components"] > args.max_components:
+            failed.append(f"components {m['components']} > {args.max_components}")
     m["pass"] = not failed
     if failed:
         m["reason"] = "; ".join(failed)
--- a/scripts/render-shadow-metric.py
+++ b/scripts/render-shadow-metric.py
@@ -1,0 +1,212 @@
+#!/usr/bin/env python3
+"""Structural sun-shadow metrics for deterministic render validation.
+
+Where ``render-compare.py`` does a pixel-diff against a committed reference,
+this measures a *structural* property of a single capture: how
+hole-ridden / fragmented a cast shadow is inside a region of interest. Two
+reasons that matters for the engine's render-validation:
+
+  * **Deterministic at zoom.** Pixel-diff is excluded above ~2x magnification
+    because CPU<->GPU float divergence drifts the match% run-to-run
+    (canvas_stress's ``revoxelize_solids_zoom`` shot is excluded for exactly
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — render: structural shadow-hole metric for deterministic validation (engine #1765)

**Verdict:** approve

### Praise

- **Well-scoped design.** The choice to measure structural properties (hole ratio + connected-component fragmentation) rather than a pixel diff is the right call for two distinct reasons the PR body states clearly: pixel-diff breaks under zoom-jitter and needs per-backend reference sets; a structural occupancy metric avoids both. The `SHADOW` debug-overlay's flat magenta/black fill makes classification exact rather than heuristic.
- **`read_png` reuse.** Importing via `importlib.machinery.SourceFileLoader` to work around the dashed filename is the correct stdlib-only pattern; it's the same convention already used in `render-compare.py`.
- **`MAX_FLOOD_PX` guard.** Preventing the BFS from silently hanging on pathological ROIs is good defensive engineering for a validation harness.
- **Test coverage.** 7 synthetic-PNG tests verified locally (7/7 pass). Covers: solid shadow (clean), checkerboard (swiss-cheese), all-lit, gray background ignored, interior hole, ROI restriction, and determinism — the right set for a structural metric.

### Nits

- `scripts/render-shadow-metric.py:108-111` — When the ROI exceeds `MAX_FLOOD_PX`, `components` is set to `None` and `--max-components` is silently skipped (the `m.get("components") is not None` guard in `_main`). An automation script that passes `--max-components` on a large ROI capture will receive `pass: true` without knowing the check was elided. Consider printing a warning to stderr or setting the result to `pass: false` in that case — the current behavior is opt-in-silent, which can surprise CI scripts.

- [x] `python3 -m unittest scripts.tests.test_render_shadow_metric` — 7/7 (verified locally).
- [ ] Manual sanity check: run against a real SHADOW-mode engine capture (e.g. `IRShapeDebug` with `--debug-overlay SHADOW`) to confirm `hole_ratio` reads near 0 for a clean shadow and spikes on a known swiss-cheese frame.

Opus recheck not required.

jakildev:
Addressed nit: when `--max-components` is requested but the component pass was skipped (ROI too large → `components: null`), now emits a stderr warning and adds a failure reason so the exit code is 1 instead of silently passing. Added `test_max_components_cli_fails_when_roi_too_large` to cover the path (8/8 pass). — worker

PR #1774, opened 2026-06-13
render: batch widget GUI text onto the compute-shader path

jakildev:
## Summary
- Widget render systems rasterized text via `renderText` → `setTrixel`, issuing **two 1×1 `subImage2D` uploads per lit glyph pixel** — thousands of CPU→GPU micro-uploads per frame for a screen of widget text. This starved the GPU (low editor framerate) and, via the split color/distance writes, is a leading flicker suspect.
- A batched compute path already existed for free text (`TEXT_TO_TRIXEL`): `expandTextToCommands` packs glyphs into one SSBO and a single `dispatchCompute` rasterizes them GPU-side, writing only lit texels and scaling by `fontSize`.
- This moves that layout + dispatch into a shared `IRPrefab::GuiText` header (`engine/prefabs/irreden/render/gui_text_batch.hpp`) and routes **every** text-bearing widget system through it: each queues its glyphs in `tick` and dispatches once in `endTick`. Rect fills (already one batched `subImage2D` each) are unchanged.
- **Zero blast radius:** no pipeline edits are required in any creation — each widget system self-dispatches — so widget-using creations (including ones outside this repo) get the win with no churn and nothing breaks.
- Foundation for a full-resolution GUI canvas: the `fontSize` multiplier is now available to widget text (next phase).

## Files
- New: `engine/prefabs/irreden/render/gui_text_batch.hpp` (`IRPrefab::GuiText` — moved glyph-layout helpers + `queueGuiText`/`dispatchGuiText`).
- `system_text_to_trixel.hpp` consumes the shared helpers (free-text path unchanged).
- `widget_draw.hpp` text helpers append to a command vector instead of rendering immediately.
- All 9 text-bearing widget render systems migrated (panel, label, button, slider, checkbox, list, text_input, radio, dropdown) + editor loft labels.

## Test plan
- [x] Engine + `IRVoxelEditor` build clean (worktree); `radio`/`dropdown`/`scroll` headers syntax-check clean with real compile flags.
- [x] `IRVoxelEditor --auto-screenshot` renders **identically** — every panel/label/list/slider-value crisp at `fontSize 1` (this is an invisible-by-design render-path swap; verified no visual delta).
- [x] `IRShapeDebug` scene unaffected (free-text path intact).
- [ ] Reviewer note: FPS magnitude not yet quantified (GPU stage timer is off by default); the win is architectural — thousands of `subImage2D` → ~9 batched dispatches/frame.

### DIFF
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -1129,6 +1129,12 @@ void initSystems() {
                 Color{200, 200, 200, 180},
                 1
             );
+            // Inline dispatch (not deferred to endTick like the widget
+            // render systems): the loft overlay is a single-canvas system
+            // that runs its whole render once per frame in this tick body,
+            // so queuing and dispatching here is equivalent to the deferred
+            // pattern — there is no second tick that would append more glyphs
+            // before the dispatch.
             IRPrefab::GuiText::dispatchGuiText(lrp->textCmds_);
         }
     );
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -105,6 +105,7 @@
 // layout mouse-in-GUI-trixels helper).
 #include <irreden/render/trixel_rect.hpp>
 #include <irreden/render/trixel_text.hpp>
+#include <irreden/render/gui_text_batch.hpp>
 #include <irreden/render/mask_grid_painter.hpp>
 #include <irreden/render/layout.hpp>
 
@@ -1062,6 +1063,7 @@ void initSystems() {
     struct LoftRenderParams {
         C_TriangleCanvasTextures *canvas_ = nullptr;
         IRRender::MaskGridPaintScratch scratch_;
+        std::vector<IRRender::GlyphDrawCommand> textCmds_;
     };
     auto loftRenderData = std::make_unique<LoftRenderParams>();
     auto *lrp = loftRenderData.get();
@@ -1102,24 +1104,32 @@ void initSystems() {
                 lrp->scratch_
             );
             const int gridH = sz * IRVoxelEditor::kLoftCellPx;
-            IRRender::renderText(
-                *lrp->canvas_,
+            const ivec2 canvasSize = lrp->canvas_->size_;
+            IRPrefab::GuiText::queueGuiText(
+                lrp->textCmds_,
                 "XZ",
                 IRVoxelEditor::kLoftGridXZPos + ivec2(0, -12),
-                Color{200, 220, 200, 220}
+                canvasSize,
+                Color{200, 220, 200, 220},
+                1
             );
-            IRRender::renderText(
-                *lrp->canvas_,
+            IRPrefab::GuiText::queueGuiText(
+                lrp->textCmds_,
                 "YZ",
                 IRVoxelEditor::kLoftGridYZPos + ivec2(0, -12),
-                Color{200, 220, 200, 220}
+                canvasSize,
+                Color{200, 220, 200, 220},
+                1
             );
-            IRRender::renderText(
-                *lrp->canvas_,
+            IRPrefab::GuiText::queueGuiText(
+                lrp->textCmds_,
                 "LOFT  Shift=sym  C=clear  Enter=stamp  F=exit",
                 ivec2(IRVoxelEditor::kLoftGridXZPos.x, IRVoxelEditor::kLoftGridXZPos.y + gridH + 4),
-                Color{200, 200, 200, 180}
+                canvasSize,
+                Color{200, 200, 200, 180},
+                1
             );
+            IRPrefab::GuiText::dispatchGuiText(lrp->textCmds_);
         }
     );
     IRSystem::setSystemParams(loftRenderSystem, std::move(loftRenderData));
@@ -2647,8 +2657,10 @@ void initCommands() {
                 );
                 const bool atRigRoot = j.parentIndex_ == static_cast<std::uint32_t>(i) ||
                                        j.parentIndex_ >= static_cast<std::uint32_t>(count);
-                IR_ASSERT(atRigRoot || j.parentIndex_ < i,
-                    ".rig parentIndex forward-reference — not supported by linear reconstruction");
+                IR_ASSERT(
+                    atRigRoot || j.parentIndex_ < i,
+                    ".rig parentIndex forward-reference — not supported by linear reconstruction"
+                );
                 const IREntity::EntityId parentEntity =
                     atRigRoot ? rigRoot : newJoints[j.parentIndex_];
                 IREntity::setParent(joint, parentEntity);
@@ -2975,14 +2987,12 @@ void initEntities() {
             kBoneSwatchOriginX + col * (kBoneSwatchSize + kBoneSwatchGap),
             kBoneSwatchOriginY + row * (kBoneSwatchSize + kBoneSwatchGap)
         );
-        IRVoxelEditor::g_bonePaint.boneSwatches_.push_back(
-            IRPrefab::Widget::makeColorSwatch(
-                pos,
-                ivec2(kBoneSwatchSize, kBoneSwatchSize),
-                IRVoxelEditor::kBoneColors[i],
-                i == 0
-            )
-        );
+        IRVoxelEditor::g_bonePaint.boneSwatches_.push_back(IRPrefab::Widget::makeColorSwatch(
+            pos,
+            ivec2(kBoneSwatchSize, kBoneSwatchSize),
+            IRVoxelEditor::kBoneColors[i],
+            i == 0
+        ));
     }
 
     // Skeleton tree panel (F-2.6, #1607). Sits below the BONE selector in the
--- a/engine/prefabs/irreden/render/gui_text_batch.hpp
+++ b/engine/prefabs/irreden/render/gui_text_batch.hpp
@@ -1,5 +1,5 @@
-#ifndef GUI_TEXT_BATCH_H
-#define GUI_TEXT_BATCH_H
+#ifndef GUI_TEXT_BATCH_HPP
+#define GUI_TEXT_BATCH_HPP
 
 // Batched GUI text on the compute path.
 //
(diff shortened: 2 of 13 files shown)


jakildev (commented):
## Review — render: batch widget GUI text onto the compute-shader path

**Verdict:** approve

### Checklist

**ECS invariants** — no per-entity `getComponent` in tick bodies. Each widget system accumulates glyph commands in `tick` and dispatches once in `endTick`. `dispatchGuiText`'s one canvas lookup (`IREntity::getComponent<C_TriangleCanvasTextures>(guiCanvas)`) is in `endTick`, not inside a per-entity loop. ✅

**Ownership / lifetime** — `IRRender::getNamedResource<>` returns a non-owning raw pointer into the render subsystem's registry; documented engine convention. `IRRender::getCanvas("gui")` returns an `EntityId` (value type). No captured `this` or manager references in lambdas. ✅

**Render pipeline** — No CPU struct / shader layout changes. `GlyphDrawCommand` struct is unchanged; the `GlyphDrawCommandBuffer` and `FontDataBuffer` are the same named resources `TEXT_TO_TRIXEL` already owned. `dispatchGuiText` binds colors at image unit 0 and distances at image unit 1 — confirmed identical to `system_text_to_trixel.hpp:111-113` (the pre-existing assignment). `dispatchCompute(count, 1, 1)` matches the per-command invocation model the shader expects. No new GLSL/Metal files added. ✅

**Ordering dependency** — `dispatchGuiText` reuses `TextToTrixelProgram` / `GlyphDrawCommandBuffer` / the "gui" canvas, all created in `System<TEXT_TO_TRIXEL>::create()`. The PR documents this constraint clearly: if `TEXT_TO_TRIXEL` is absent, `dispatchGuiText` gracefully clears and returns (no crash, no vector growth). ✅

**Memory barrier correctness** — `TEXT_TO_TRIXEL`'s endTick posts `BarrierType::ALL` after its dispatch; each widget system's `dispatchGuiText` also posts `BarrierType::ALL` after writing to the shared canvas. Sequential widget dispatches are correctly separated. ✅

**`textCmds_` lifecycle** — vectors are cleared by `dispatchGuiText` at the end of each `endTick`. The early-exit for absent resources clears the vector too, preventing accumulation across frames. ✅

**Silent truncation** — `dispatchGuiText` clamps `count` to `kMaxGlyphCommands` (8192) without a warning. This matches `TEXT_TO_TRIXEL`'s existing behavior. For typical widget text budgets this cap is unreachable, but a `IR_WARN` would surface the rare overflow.

**`measureSingleLine` removal** — confirmed removed from `widget_draw.hpp` and all call sites migrated; build passes. ✅

**Naming / style** — `IRPrefab::GuiText` namespace is internally consistent. `kWidgetTextFontSize` is a `constexpr` (no trailing `_` needed). Private fields (`textCmds_`, `scratch_`) correctly carry trailing `_`. ✅

**Serialization / math / lighting** — not touched. ✅

**Tests / build** — build confirmed clean; `IRVoxelEditor --auto-screenshot` confirms visual identity (invisible render-path swap). ✅

### Nits

- `engine/prefabs/irreden/render/gui_text_batch.hpp:103` — `#ifndef GUI_TEXT_BATCH_H` should be `GUI_TEXT_BATCH_HPP` to match the `.hpp` extension (consistent with other prefab headers).
- `creations/editors/voxel_editor/main.cpp:1132` (loft render tick) — `dispatchGuiText` is called inline at the end of the loft `tick` rather than in an `endTick`. This differs from the widget-system deferred pattern; a brief comment explaining why (the loft render is a single-canvas, single-tick-per-frame system, so inline dispatch is safe and equivalent) would prevent future maintainers from treating the asymmetry as an oversight.
- `engine/prefabs/irreden/render/gui_text_batch.hpp:358-360` — the silent clamp to `kMaxGlyphCommands` could use a `IR_WARN` or `IR_ASSERT` in debug mode; if UI ever gets complex enough to hit 8192 glyphs per widget system, the truncation will be hard to diagnose.

### Praise

- Moving all glyph-layout helpers to a shared `IRPrefab::GuiText` namespace is the right abstraction: `TEXT_TO_TRIXEL` and all 9 widget render systems now share one code path, and the migration is zero-blast-radius for downstream creations.
- Graceful degradation in `dispatchGuiText` (clears vector when named resources absent) is exactly the right defensive posture for a path that depends on a pipeline prerequisite.
- The deferred `tick` → `endTick` pattern gives one GPU compute dispatch per widget system per frame, replacing thousands of `subImage2D` micro-uploads. Architectural win is well-scoped and non-breaking.

- [x] Engine + `IRVoxelEditor` build clean (confirmed in PR body)
- [x] `IRVoxelEditor --auto-screenshot` renders identically (confirmed in PR body)
- [ ] FPS improvement on a widget-heavy editor scene (optional — the PR notes GPU stage timing is off by default; the win is architectural)

jakildev:
Addressed has-nits (all three reviewer nits):
- Renamed the include guard `GUI_TEXT_BATCH_H` → `GUI_TEXT_BATCH_HPP` to match the `.hpp` extension.
- `dispatchGuiText` now logs `IRE_LOG_WARN` on the `kMaxGlyphCommands` (8192) overflow instead of silently truncating, so an overflow is diagnosable.
- Added a comment at the loft render tick explaining why it dispatches GUI text inline rather than deferring to `endTick` (single-canvas, single-tick-per-frame system → inline is equivalent).

Plus a clang-format pass on the changed file. Build clean (IRVoxelEditor). `fleet:approved` retained — these are cleanups, not behavior changes. — worker

jakildev:
Note on nit #1 (include-guard rename): applied as requested (`_HPP`), but on reflection the engine-wide convention is actually `<NAME>_H` (477 vs 3 `_HPP` across `engine/`; 156 vs 1 in `engine/prefabs/irreden/render/`). The convention isn't documented anywhere, which is why the nit went against the grain. Filed #1780 (coding-improvement) to document it; left the guard as `_HPP` here to avoid re-thrashing an approved PR over a functionally-inert token — the human can sweep the few `_HPP` outliers to `_H` in one pass via #1780 if desired. — worker

jakildev (commented):
## Opus recheck — render: batch widget GUI text onto the compute-shader path

**Verdict:** approve

Sonnet approved this and covered the surface checklist well (ECS naming, the deferred tick→endTick pattern, graceful degradation, the canvas-clear ordering, the three nits — all addressed in `ebb1227e`). I spent the Opus pass on the one invariant Sonnet's checklist couldn't confirm three systems deep: **whether `dispatchGuiText` sets up *all* the GPU state the shader needs, given that the same compute shader now re-dispatches much later in the RENDER pipeline** — once per widget render system endTick, after `SHAPES_TO_TRIXEL` and other compute systems have run — rather than only at `TEXT_TO_TRIXEL::endTick` as before this PR.

### The load-bearing concern: residual SSBO bindings across the widened window

`dispatchGuiText` (`gui_text_batch.hpp:276-298`) re-binds the program (`program->use()`), re-uploads the command buffer (`cmdBuf->subData`), and re-binds the two canvas image units (0 = colors, 1 = distances) on every call. But it does **not** re-bind the two SSBO binding points the shader reads — `FontDataBuffer` and `GlyphDrawCommandBuffer`. Those are bound once at create time (`system_text_to_trixel.hpp:125-140`) via dedicated binding indices and never re-bound. The whole path is correct only if those indexed bindings survive everything that runs between `TEXT_TO_TRIXEL::endTick` and the last widget endTick.

I confirmed they do:
- `kBufferIndex_FontData = 11` and `kBufferIndex_GlyphDrawCommands = 12` (`ir_render_types.hpp:646-647`) are **statically dedicated** — one buffer per index, no other `createNamedResource<Buffer>` reuses 11 or 12. Grep across the engine returns zero other users.
- `glUseProgram` and `subData` (generic-target `glBufferSubData`) do not touch indexed SSBO binding-base state, so the create-time `bindBufferBase` persists across the whole pipeline window regardless of how many other compute systems dispatch in between.

So the residual-binding assumption that the wider re-dispatch window now leans on is sound. This is the real reason the refactor is safe, and it wasn't something the first pass could verify from the diff alone.

### Other Opus-only checks

- **Image-unit rebinds (0/1):** re-bound every `dispatchGuiText` call, so interleaving with `VOXEL_TO_TRIXEL` / `SHAPES_TO_TRIXEL` (which also use image units 0/1) is harmless. ✓
- **Barrier ordering:** each dispatch posts `BarrierType::ALL` after writing the GUI canvas, separating the up-to-9 sequential widget dispatches from each other and from the downstream `TRIXEL_TO_FRAMEBUFFER` read. ✓
- **Dispatch contract unchanged:** `dispatchCompute(count, 1, 1)` is byte-identical to the pre-existing `TEXT_TO_TRIXEL` dispatch; `GlyphDrawCommand` struct + the shader are untouched, so no CPU/GPU layout drift. ✓
- **ECS invariant:** the per-frame `getComponent<C_TriangleCanvasTextures>(guiCanvas)` runs in widget endTicks (foreign-entity lookup, ~10×/frame), never inside a per-entity tick — within the rules. ✓
- **No new `System<N>`:** `dispatchGuiText` is a free function in `IRPrefab::GuiText`, so no `SystemName` enum entry is required. ✓
- **Loft inline dispatch (Sonnet nit 2):** on closer read the dispatch is in the loft system's **endTick lambda**, not its per-entity tick (which is an empty `[](const C_GuiElement &){}`). So it fires once per frame — no repeated dispatch or per-entity foreign `getComponent`. Sonnet's "inline at the end of the loft tick" phrasing was imprecise but the placement is correct, and the new explanatory comment documents it.

### Verdict

Clean, correct, zero-blast-radius refactor: it consolidates `TEXT_TO_TRIXEL` and all widget text onto one shared compute path, and the one invariant the wider dispatch window introduced (residual SSBO bindings on dedicated indices 11/12) holds. No follow-up architectural decision implied. The macOS/Windows smoke labels are still pending (OpenGL/Linux is the executed validation so far); the smoke flow will catch any backend surprise. Safe for merge.

Escalation: none.

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). 21/21 screenshots captured, build clean, exit 0. Pre-existing MIDI static-destruction crash on shutdown (T-336) — not a regression.

PR #1778, opened 2026-06-13
render: span all encoders for Metal per-stage GPU timestamps (#1746)

### LINKED ISSUE #1746: render: Metal per-stage GPU timestamps undercount multi-encoder stages (only the first encoder is bracketed)

jakildev:
## Summary
- Metal brackets a GPU timestamp sample pair to a **single** command encoder. A render stage that issues N encoders (e.g. `VOXEL_TO_TRIXEL_STAGE_1`'s per-axis canvas passes) only measured its first encoder, so the tracked GPU stages summed to <1 ms against a ~57 ms frame (#1738 fixed the aggregation half; this fixes collection accuracy).
- Make the sticky attachment (`g_nextComputeTimestampAttachment`) span the **whole stage**: a new `firstEncoder_` flag has the first encoder claim the start boundary and **every** encoder re-write the end boundary. Sequential GPU execution makes the last encoder's index-1 write win → the resolved pair spans `[first-encoder start, last-encoder end]`. Non-first encoders pass `MTL::CounterDontSample` for the start index.
- `writeTimestamp(END)` now clears the sticky state, so encoders created in the gap before the next `START` stay untracked (also closes a latent stale-attachment carry-over for zero-encoder windows that the old first-encoder-resets path relied on).
- Mirrored in both `createComputeEncoder` (compute encoders) and `attachTimestampSamples` (render encoders); the shared global means a stage mixing both encoder types has its first-executed encoder claim the start.
- **Single-encoder stages are byte-identical** to before (the one encoder claims both boundaries). **No OpenGL file is touched** — OpenGL already spans whole stages via `glQueryCounter` at the call sites.

## Verification (macOS / Metal, Apple Silicon)
- `IRPerfGrid --auto-profile 200 --zoom 8 --subdivision-mode full --base-subdivisions 1`: `voxelStage1` GPU jumped from <1 ms (first-encoder-only) to **34.945 ms** against a **52.307 ms** frame — a sane fraction, no ~100× undercount.
- **Metal validation layer** (`METAL_DEVICE_WRAPPER_TYPE=1`): the shared 2-sample CounterSampleBuffer multi-write at index 1 produces **no validation error**. (The one validation assertion that fires — a stencil-attachment pixel-format mismatch in `setRenderPipelineState` — is **pre-existing on master**; confirmed identical with this change reverted. It is a separate latent issue, out of scope for #1746, and does not affect normal non-validation runs.)
- Build clean: `fleet-build --target IRShapeDebug` / `IRPerfGrid`.

## Render-loop / screenshots
Skipped per the documented "internal refactor with provably no runtime effect" exception (`engine/render/CLAUDE.md` §Verifying render changes). This change only alters GPU counter-sample-buffer **attachment metadata** — it cannot change any rendered pixel; before/after captures would be byte-identical.

## Test plan
- [x] macOS/Metal: `voxelStage1` GPU is realistic multi-ms, sum within a sane fraction of frame time
- [x] No Metal validation-layer error from the counter-buffer multi-write
- [x] Build green (IRShapeDebug + IRPerfGrid)
- [ ] Linux/OpenGL smoke (no OpenGL file edited — no-edit guarantee)

Closes #1746

### DIFF
--- a/engine/render/src/metal/metal_render_impl.cpp
+++ b/engine/render/src/metal/metal_render_impl.cpp
@@ -28,6 +28,13 @@ struct MetalTimestampSampleAttachment {
     MTL::CounterSampleBuffer *sampleBuffer_ = nullptr;
     NS::UInteger startSampleIndex_ = 0;
     NS::UInteger endSampleIndex_ = 1;
+    // Sticky across every encoder of a stage (the window between
+    // writeTimestamp START and END): the first encoder claims the start
+    // boundary; every encoder re-writes the end boundary. Sequential GPU
+    // execution makes the last encoder's end write win, so the resolved pair
+    // spans [first-encoder start, last-encoder end] — the whole stage, not
+    // just its first encoder (#1746).
+    bool firstEncoder_ = true;
 };
 
 MetalTimestampSampleAttachment g_nextComputeTimestampAttachment;
@@ -122,12 +129,18 @@ MTL::ComputeCommandEncoder *createComputeEncoder(MTL::CommandBuffer *commandBuff
     auto *descriptor = MTL::ComputePassDescriptor::alloc()->init();
     auto *attachment = descriptor->sampleBufferAttachments()->object(kTimestampAttachmentIndex);
     attachment->setSampleBuffer(g_nextComputeTimestampAttachment.sampleBuffer_);
-    attachment->setStartOfEncoderSampleIndex(g_nextComputeTimestampAttachment.startSampleIndex_);
+    attachment->setStartOfEncoderSampleIndex(
+        g_nextComputeTimestampAttachment.firstEncoder_
+            ? g_nextComputeTimestampAttachment.startSampleIndex_
+            : MTL::CounterDontSample
+    );
     attachment->setEndOfEncoderSampleIndex(g_nextComputeTimestampAttachment.endSampleIndex_);
     auto *encoder = commandBuffer->computeCommandEncoder(descriptor);
     descriptor->release();
 
-    g_nextComputeTimestampAttachment = {};
+    // Stay sticky for the rest of the stage's encoders (each re-writes the end
+    // boundary; the last one wins). writeTimestamp(END) clears the attachment.
+    g_nextComputeTimestampAttachment.firstEncoder_ = false;
     return encoder;
 }
 
@@ -138,9 +151,16 @@ void attachTimestampSamples(MTL::RenderPassDescriptor *descriptor) {
 
     auto *attachment = descriptor->sampleBufferAttachments()->object(kTimestampAttachmentIndex);
     attachment->setSampleBuffer(g_nextComputeTimestampAttachment.sampleBuffer_);
-    attachment->setStartOfVertexSampleIndex(g_nextComputeTimestampAttachment.startSampleIndex_);
+    attachment->setStartOfVertexSampleIndex(
+        g_nextComputeTimestampAttachment.firstEncoder_
+            ? g_nextComputeTimestampAttachment.startSampleIndex_
+            : MTL::CounterDontSample
+    );
     attachment->setEndOfFragmentSampleIndex(g_nextComputeTimestampAttachment.endSampleIndex_);
-    g_nextComputeTimestampAttachment = {};
+    // Sticky across the stage's encoders (the shared global means a stage that
+    // mixes compute + render encoders has its first-executed encoder claim the
+    // start boundary). writeTimestamp(END) clears the attachment.
+    g_nextComputeTimestampAttachment.firstEncoder_ = false;
 }
 
 MTL::RenderCommandEncoder *createRenderEncoder() {
@@ -786,11 +806,18 @@ metalCurrentDepthPixelFormat(),
             g_nextComputeTimestampAttachment = {
                 pair->sampleBuffer_,
                 0,
-                1
+                1,
+                /*firstEncoder_=*/true
             };
             pair->hasStart_ = true;
             pair->hasEnd_ = false;
         } else {
+            // Stop tagging once the stage ends: clear the sticky attachment so
+            // encoders created in the gap before the next START stay untracked
+            // (#1746). The end boundary was already attached to every encoder
+            // of the stage up to this point, so the last one's index-1 write
+            // bounds the pair.
+            g_nextComputeTimestampAttachment = {};
             pair->hasEnd_ = true;
         }
     }


jakildev (commented):
## Review — render: span all encoders for Metal per-stage GPU timestamps (#1746)

**Verdict:** approve

### Summary

Single-file Metal change that fixes under-reported GPU stage times when a
stage issues more than one encoder. The root cause (#1738 fixed aggregation;
this fixes collection accuracy): `g_nextComputeTimestampAttachment` was
cleared at the first encoder's creation, so only that encoder's end boundary
was sampled — the stage's remaining encoders ran untracked.

The fix makes the attachment sticky across the stage window: a new
`firstEncoder_` flag lets only the first encoder claim the start boundary
(`startSampleIndex_`), while every encoder (including non-first) re-writes
the end boundary (`endSampleIndex_`). Sequential GPU execution guarantees the
last encoder's end write wins, so the resolved pair spans `[first-encoder
start, last-encoder end]` — the whole stage. Clearing moves from
encoder-creation time to `writeTimestamp(END)` time.

### Checklist pass

**ECS invariants** — N/A; this is a Metal render-device implementation, no
ECS tick code. ✓

**Ownership / lifetime** — `g_nextComputeTimestampAttachment` is a file-local
global; no raw ownership added. The `MetalTimestampPair::sampleBuffer_` is
released in `releaseTimestampPair` as before; the sticky attachment holds a
non-owning observer pointer (same as before). ✓

**Render pipeline** — Metal-only single-file change. No GLSL/Metal shader
modified; no CPU struct/layout changed; no bind-point additions. OpenGL path
is untouched. ✓

**Correctness of multi-write to index 1** — Metal's `CounterSampleBuffer`
accepts multiple writes to the same index; GPU serial execution (command
buffer submission order) means the last write wins. PR body confirms Metal
validation layer (`METAL_DEVICE_WRAPPER_TYPE=1`) produced no error on the
multi-write. ✓

**`firstEncoder_ = true` after `= {}`** — `MetalTimestampSampleAttachment` is
a C++ aggregate with a brace-or-equal default initializer
(`bool firstEncoder_ = true`). Per C++17 aggregate initialization, a
`= {}` re-sets every member to its declared default, so
`g_nextComputeTimestampAttachment = {}` in `writeTimestamp(END)` restores the
ready-for-next-stage state correctly. ✓

**Zero-encoder stage** — if `writeTimestamp(START)` fires but no encoder is
created before `writeTimestamp(END)`, the attachment is set then cleared by
END. `pair->hasStart_` and `pair->hasEnd_` both become true, but
`resolveCounterRange` returns `MTL::CounterErrorValue` for the unwritten
slots, and `readTimestampPairMs` returns false on that guard. No crash, no
stale sample. ✓

**Single-encoder stages byte-identical** — single-encoder stage: first
encoder sees `firstEncoder_ == true`, claims both boundaries, sets
`firstEncoder_ = false`. Behavior is byte-identical to before. PR body
confirms this. ✓

**Latent carry-over fix** — pre-patch, a START-but-no-encoder window would
carry `g_nextComputeTimestampAttachment` into the next stage. The patch moves
clearing to END, closing this latent race correctly. ✓

**Naming** — `firstEncoder_` has trailing `_` (public member convention). ✓

**Math / coordinates** — none. ✓

**Serialization** — none. ✓

**Tests / build** — macOS/Metal build + `IRPerfGrid` run confirmed. Linux/OpenGL
path is no-edit; cross-host build smoke labels added (macOS-authored → Linux
+ Windows still pending). ✓

### Praise

- Moving clear from encoder-creation to `writeTimestamp(END)` is the cleaner
  invariant: the attachment now lives exactly as long as the stage window, not
  just until the first encoder of that stage. This is simpler to reason about
  and closes the latent zero-encoder carry-over for free.
- The `MTL::CounterDontSample` sentinel for non-first-encoder starts is the
  right API knob — no sampling noise, no wasted counter slots.
- Explicit `/*firstEncoder_=*/true` in the aggregate init is the correct
  defensive style given that `= {}` would also work here; making the intent
  visible at the one place where the START is armed is worth the
  over-specification.

- [x] macOS/Metal: `voxelStage1` GPU realistic multi-ms, sane fraction of frame time
- [x] Metal validation layer: no error on counter-buffer multi-write
- [x] Build green (IRShapeDebug + IRPerfGrid)
- [ ] Linux build smoke (OpenGL path unchanged — compile-only check)
- [ ] Windows build smoke (OpenGL path unchanged — compile-only check)

Opus recheck not required.

jakildev:
Cross-host smoke OK on linux (fresh checkout build + IRShapeDebug --auto-screenshot 10). OpenGL build green; diff is Metal-only (engine/render/src/metal/metal_render_impl.cpp, not compiled on linux). All 21 auto-screenshots captured, clean exit, render output healthy (shapes/lighting/shadows/depth nominal — no visual regression). Opus-class screenshot inspection done. — worker-4

jakildev (commented):
## Opus recheck — render: span all encoders for Metal per-stage GPU timestamps (#1746)

**Verdict:** approve

Sonnet approved and covered the core semantics well (multi-write-to-index-1, the zero-encoder guard, single-encoder byte-identity, the aggregate `= {}` re-default, and the latent carry-over fix). I focused the Opus pass on the part that only matters once you trace the shared global's lifecycle across a full frame, not just a single stage: **what every encoder created in the gap *between* stages does, and whether the multi-encoder end-write ordering is actually guaranteed.**

### Gap-encoder lifecycle across the whole frame — confirmed safe

The attachment is a single file-local global re-armed at `writeTimestamp(START)` and cleared at `writeTimestamp(END)`. The subtle case is an encoder created *after* a stage's END but *before* the next stage's START:

- `writeTimestamp(END)` resets via `g_nextComputeTimestampAttachment = {}`, which leaves `sampleBuffer_ == nullptr` (and `firstEncoder_ == true` by aggregate default).
- A gap encoder therefore calls `attachment->setSampleBuffer(nullptr)`. Even though `firstEncoder_` is `true` and a real `startSampleIndex_` (0) would be attached, Metal ignores a sample-buffer attachment whose buffer is null — so the gap encoder is untracked. This matches the **pre-patch** gap behavior exactly (pre-patch the attachment was also `{}`/null after the first encoder cleared it). So the only behavioral delta is the intended one: *within* a stage window the second-and-later encoders now stay tracked instead of being dropped. The cross-stage isolation is preserved. ✓

The mirror concern — `firstEncoder_` being left `true` in the gap causing a spurious start re-sample on the next stage — is moot for the same reason (null buffer), and the next START re-assigns the whole struct anyway. ✓

### Multi-encoder end-write ordering — the real load-bearing assumption

The fix relies on "every encoder re-writes end index 1, last write wins." That holds because all of a stage's encoders are encoded onto the same command queue and execute in submission order; `setEndOfEncoderSampleIndex(1)` on encoder N is overwritten by encoder N+1's index-1 write, so the resolved pair is `[encoder-0 start, encoder-last end]`. Out-of-order or parallel execution would break this, but Metal compute/render encoders on one queue are serial — so the span is correct by construction. The author validated the repeated index-1 write under the Metal validation layer (`METAL_DEVICE_WRAPPER_TYPE=1`) with no error, which is the right way to confirm the "overwrite same counter slot" API usage since it can't be exercised off-macOS. ✓

### Other checks

- **Aggregate-init member order** — `{ pair->sampleBuffer_, 0, 1, /*firstEncoder_=*/true }` matches the struct declaration order (`sampleBuffer_`, `startSampleIndex_`, `endSampleIndex_`, `firstEncoder_`). ✓
- **Mixed compute+render stage** — both `createComputeEncoder` and `attachTimestampSamples` (render) consult and clear the same `firstEncoder_`, so a stage mixing encoder types still has exactly one start claimant. ✓
- **Scope / blast radius** — `engine/render/src/metal/metal_render_impl.cpp` only; compiled on macOS only. OpenGL/Linux/Windows render paths are untouched, so the pending `fleet:needs-linux-smoke` / `needs-windows-smoke` labels cover a file those targets don't even compile — no functional risk there; the macOS `IRPerfGrid` run in the PR body is the meaningful validation.

### Verdict

Correct, minimal, Metal-only instrumentation fix. The sticky-attachment lifecycle is sound across the full frame (not just per stage), and the last-write-wins span is guaranteed by Metal's serial encoder execution. No architectural follow-up implied. Safe for merge once the macOS smoke clears.

Escalation: none.

PR #1782, opened 2026-06-13
render: opt-in full-resolution GUI canvas

jakildev:
## Summary
- Add `IRRender::setGuiCanvasFullResolution()`: resize the GUI canvas from the coarse iso-canvas resolution (`mainCanvasSize / guiScale`) to the **native framebuffer pixel resolution** (1 GUI trixel == 1 framebuffer pixel). GUI text/widgets render small, crisp, and un-stretched instead of at the doubled-width iso scale.
- **Default off / opt-in.** A creation opts in and owns laying its GUI out for the finer coordinate space, so other creations (and the game repo) are unaffected. Factored the shared canvas-texture destroy/recreate into `resizeGuiCanvas` (reused by `setGuiScale`).
- No compositor/input changes: the trixel→framebuffer composite already UV-samples the GUI canvas, and the mouse→GUI-trixel hit mapping derives from the live canvas size.
- The **voxel editor opts in**. Before: panels/text were oversized and ran off the bottom edge. After (GUI canvas 1604×902): panels are compact, text is small and crisp, and the scene is clearly visible.

## Stack context
Stacked on: https://github.com/jakildev/IrredenEngine/pull/1774

When the parent PR merges, change this PR's base to `master`
(via the GitHub UI or `gh pr edit <N> --base master`).

## Test plan
- [x] Engine + `IRVoxelEditor` build clean.
- [x] `IRVoxelEditor --auto-screenshot` verified: GUI now renders at 1604×902, text small/crisp, all six panels fit on-screen (vs. the parent where they ate the bottom third and spilled).
- [x] Default path unchanged — `setGuiScale` still routes through `resizeGuiCanvas`; no creation except the editor calls the new API, so other creations are byte-identical.

## Notes for reviewer
- This is a GUI **resolution** change (no shader/voxel/silhouette change), verified by direct editor screenshot rather than a render-debug-loop ROI-crop pass. The visual delta is the dramatic GUI shrink described above.
- Editor **relayout polish** (panel placement to use the space, list-row `itemHeight` vs glyph height, body `fontSize`, the slight `CLICK A SWATCH` overflow) is the next stacked PR — out of scope here.

### DIFF
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -1129,6 +1129,12 @@ void initSystems() {
                 Color{200, 200, 200, 180},
                 1
             );
+            // Inline dispatch (not deferred to endTick like the widget
+            // render systems): the loft overlay is a single-canvas system
+            // that runs its whole render once per frame in this tick body,
+            // so queuing and dispatching here is equivalent to the deferred
+            // pattern — there is no second tick that would append more glyphs
+            // before the dispatch.
             IRPrefab::GuiText::dispatchGuiText(lrp->textCmds_);
         }
     );
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -105,6 +105,7 @@
 // layout mouse-in-GUI-trixels helper).
 #include <irreden/render/trixel_rect.hpp>
 #include <irreden/render/trixel_text.hpp>
+#include <irreden/render/gui_text_batch.hpp>
 #include <irreden/render/mask_grid_painter.hpp>
 #include <irreden/render/layout.hpp>
 
@@ -1062,6 +1063,7 @@ void initSystems() {
     struct LoftRenderParams {
         C_TriangleCanvasTextures *canvas_ = nullptr;
         IRRender::MaskGridPaintScratch scratch_;
+        std::vector<IRRender::GlyphDrawCommand> textCmds_;
     };
     auto loftRenderData = std::make_unique<LoftRenderParams>();
     auto *lrp = loftRenderData.get();
@@ -1102,24 +1104,32 @@ void initSystems() {
                 lrp->scratch_
             );
             const int gridH = sz * IRVoxelEditor::kLoftCellPx;
-            IRRender::renderText(
-                *lrp->canvas_,
+            const ivec2 canvasSize = lrp->canvas_->size_;
+            IRPrefab::GuiText::queueGuiText(
+                lrp->textCmds_,
                 "XZ",
                 IRVoxelEditor::kLoftGridXZPos + ivec2(0, -12),
-                Color{200, 220, 200, 220}
+                canvasSize,
+                Color{200, 220, 200, 220},
+                1
             );
-            IRRender::renderText(
-                *lrp->canvas_,
+            IRPrefab::GuiText::queueGuiText(
+                lrp->textCmds_,
                 "YZ",
                 IRVoxelEditor::kLoftGridYZPos + ivec2(0, -12),
-                Color{200, 220, 200, 220}
+                canvasSize,
+                Color{200, 220, 200, 220},
+                1
             );
-            IRRender::renderText(
-                *lrp->canvas_,
+            IRPrefab::GuiText::queueGuiText(
+                lrp->textCmds_,
                 "LOFT  Shift=sym  C=clear  Enter=stamp  F=exit",
                 ivec2(IRVoxelEditor::kLoftGridXZPos.x, IRVoxelEditor::kLoftGridXZPos.y + gridH + 4),
-                Color{200, 200, 200, 180}
+                canvasSize,
+                Color{200, 200, 200, 180},
+                1
             );
+            IRPrefab::GuiText::dispatchGuiText(lrp->textCmds_);
         }
     );
     IRSystem::setSystemParams(loftRenderSystem, std::move(loftRenderData));
@@ -2647,8 +2657,10 @@ void initCommands() {
                 );
                 const bool atRigRoot = j.parentIndex_ == static_cast<std::uint32_t>(i) ||
                                        j.parentIndex_ >= static_cast<std::uint32_t>(count);
-                IR_ASSERT(atRigRoot || j.parentIndex_ < i,
-                    ".rig parentIndex forward-reference — not supported by linear reconstruction");
+                IR_ASSERT(
+                    atRigRoot || j.parentIndex_ < i,
+                    ".rig parentIndex forward-reference — not supported by linear reconstruction"
+                );
                 const IREntity::EntityId parentEntity =
                     atRigRoot ? rigRoot : newJoints[j.parentIndex_];
                 IREntity::setParent(joint, parentEntity);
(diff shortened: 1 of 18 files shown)


jakildev:
#1774. Its head ref `claude/render-gui-text-batch` was force-
pushed since this PR was last rebased; the child commits
replayed cleanly onto the new upstream tip. Force-pushed
with `--force-with-lease`. CI will re-run.

— fleet merger

jakildev (commented):
## Review — render: opt-in full-resolution GUI canvas

**Verdict:** approve

Stacked on #1774 (`claude/render-gui-text-batch`, open with `fleet:approved`).
Reviewing the child diff only — upstream is approved separately.

### Summary

Adds `IRRender::setGuiCanvasFullResolution()` — an opt-in, init-time API that
resizes the GUI canvas to the native framebuffer resolution (1 GUI trixel == 1
framebuffer pixel) instead of the default `mainCanvasSize / guiScale` sizing.
The voxel editor calls it once at init so panels/text render at crisp native
resolution. A `resizeGuiCanvas()` helper is extracted from the existing
`setGuiScale` resize code to eliminate duplication. A `m_guiFullResolution`
bool flag guards `setGuiScale()` so the scale-divisor resize is skipped when
full-resolution mode is active.

### Checklist pass

**ECS invariants:** `resizeGuiCanvas` calls
`IREntity::getComponent<C_TriangleCanvasTextures>(m_guiCanvas)` and
`getComponent<C_SizeTriangles>(m_guiCanvas)`. These are manager-method calls
at init time (not inside a system tick), so the per-entity-getComponent rule
does not apply. `textures.onDestroy()` + `textures = C_TriangleCanvasTextures{newSize}`
is the documented pattern for GPU resource cleanup — same as the pre-existing
`setGuiScale` code. ✓

**Ownership / lifetime:** `resizeGuiCanvas` holds `auto &textures` as a
reference, calls `onDestroy()`, then immediately reassigns through the same
reference. Component references into the archetype remain valid across the
assignment (no structural change — the archetype doesn't change, only the
component value). ✓

**Render pipeline:** No shader changes, no bind-point shifts, no struct layout
changes. Canvas sizing is a platform-agnostic resize of a `C_TriangleCanvasTextures`
resource, unchanged across OpenGL/Metal. ✓

**Naming:** `m_guiFullResolution` (private, `m_` prefix), `resizeGuiCanvas`
(private method camelCase), `setGuiCanvasFullResolution` (public API). All
consistent with existing `RenderManager` conventions. ✓

**Math primitives:** No new `std::*` / `glm::*` calls introduced in this PR. ✓

**`updateOutputResolution` interaction:** `updateOutputResolution` only updates
`m_outputScaleFactor` and `m_outputResolution` — it does NOT resize the GUI
canvas or call `setGuiScale`. The GUI canvas resize is driven by explicit calls
only. `m_gameResolution` is set at construction time and doesn't change at
runtime, so the `fbSize = m_gameResolution + kSizeExtraPixelBuffer` computation
is stable. ✓

**Voxel editor call site:** `setGuiCanvasFullResolution()` is called once in
`initEntities()`, before `IREngine::gameLoop()` — correctly the first
opportunity to size the GUI canvas before widgets are built. ✓

### Nits

- `render_manager.cpp:504-506` — in `setGuiScale`, `m_guiScale = scale` is
  written BEFORE the `m_guiFullResolution` early-return. This means a
  `setGuiScale(N)` call after `setGuiCanvasFullResolution()` would update
  `m_guiScale` to `N` without resizing the canvas, leaving `getGuiScale()`
  returning an inconsistent value. The current caller (voxel editor) never
  calls `setGuiScale` after `setGuiCanvasFullResolution`, so there's no active
  bug — but the API can silently mis-report. Moving the assignment after the
  `m_guiFullResolution` guard (or skipping the assignment when full-res is
  active) would close the gap cleanly.

### Praise

- Extracting `resizeGuiCanvas()` to eliminate the duplication between
  `setGuiScale` and `setGuiCanvasFullResolution` is the right call — the two
  callers were close enough that future changes to resize semantics (e.g.
  texture format) would have needed to be applied twice.
- The early-return guard in `setGuiCanvas` is logically correct for the
  current usage: the full-res path and the scale-divisor path are mutually
  exclusive at runtime.
- The CLAUDE.md update clearly states the "calling creation owns the layout"
  requirement — this is exactly the kind of contract note that prevents
  confusion later.

Opus recheck not required.

jakildev:
Addressed nit: moved `m_guiScale = scale` to after the `m_guiFullResolution` guard in `setGuiScale()`. Stored scale now only updates when the resize actually executes — callers in full-res mode no longer silently mis-record their intent. Build clean (IRVoxelEditor). — worker

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). 21/21 screenshots captured, build clean, exit 0. Pre-existing MIDI static-destruction crash on shutdown (T-336) — not a regression.

jakildev (commented):
## Opus recheck — render: opt-in full-resolution GUI canvas

**Verdict:** approve

Stacked on #1774 (`claude/render-gui-text-batch`, open with `fleet:approved` —
opus-rechecked separately). Reviewing the child diff only.

Sonnet approved with one nit (the `m_guiScale = scale` ordering in
`setGuiScale`), which the worker has since addressed. On closer read I confirm
Sonnet's verdict and the fix, and spent the Opus pass on the two things Sonnet
could not fully close: the component-reference lifetime in the new
`resizeGuiCanvas` extraction, and the interaction with the #1774 compute
GUI-text path this stacks on.

### Resource-reassign lifetime (the Opus-only check)

`resizeGuiCanvas` holds `auto &textures = getComponent<C_TriangleCanvasTextures>(m_guiCanvas)`,
calls `textures.onDestroy()`, then move-assigns `textures = C_TriangleCanvasTextures{newSize}`
*through the same reference*. This is sound:

- No structural ECS mutation happens between acquiring the reference and the
  reassignment (no add/remove component, no entity create/destroy) — the
  archetype column doesn't move, so the reference stays valid.
- The subsequent `getComponent<C_SizeTriangles>(...)` lookup happens *after*
  `textures` is fully written, so there's no use-after-reassign aliasing risk
  even if it touched the same archetype.

This is a behavior-preserving extraction of the exact sequence the original
`setGuiScale` ran inline — confirmed line-for-line. ✓

### Nit fix — confirmed correct

`setGuiScale` now reads `if (scale == m_guiScale) return;` → `if (m_guiFullResolution) return;`
→ `m_guiScale = scale`. The stored scale only updates when the resize will
actually execute, so a `setGuiScale(N)` after `setGuiCanvasFullResolution()`
no longer leaves `getGuiScale()` mis-reporting. ✓

### New `size_ == newSize` early-return — improvement, not regression

`resizeGuiCanvas` adds a size-equality short-circuit the inline `setGuiScale`
code lacked. For `setGuiScale` the outer `scale == m_guiScale` guard already
fires first; in the rare case two distinct scales integer-divide to the same
canvas size, the new guard skips a *needless* GPU texture destroy+recreate
while the canvas is already correctly sized — strictly better, no correctness
change. ✓

### #1774 compute GUI-text interaction

`setGuiCanvasFullResolution()` is a one-shot init-time resize in the voxel
editor's `initEntities()`, before `gameLoop()`. It updates `C_SizeTriangles`
and recreates `C_TriangleCanvasTextures` at the same `newSize`, so any
downstream consumer that reads the GUI canvas dimensions (including the #1774
compute GUI-text dispatch) gets a coherent size — there is no mid-frame resize
to race against. `m_gameResolution` is construction-time stable and
`updateOutputResolution` doesn't touch the GUI canvas, so the native-res sizing
holds. The "native-res canvas doesn't track window resize" property is a
documented, acceptable limitation for the init-once contract (the creation owns
layout). ✓

### Verdict

Clean. The extraction is behavior-preserving, the reference lifetime is safe,
the nit is closed, and the new feature doesn't perturb the upstream compute
GUI-text path. Safe for merge once #1774 lands.

🤖 Reviewed by Claude Opus 4.8 (role-opus-reviewer)

jakildev:
Merger: base PR #1774 merged; re-targeted this PR
will resolve semantically.

Conflicted files:
- `creations/editors/voxel_editor/main.cpp` — master: `c7b18b84 render: batch widget GUI text onto the compute-shader path (#1774)`; PR: `74e1b8d8 render: opt-in full-resolution GUI canvas`
- `engine/prefabs/irreden/render/gui_text_batch.hpp` — master: `c7b18b84 render: batch widget GUI text onto the compute-shader path (#1774)`; PR: `ebb1227e render: address has-nits on gui_text_batch (#1774)`

SHA pair: master=1d426bfe6f0cc6b4318b98a6a8fbfbfa81b1db5d × PR=d21a98d12f602e54d032e62127514b58e9072130

— fleet merger

Issue #1522, opened 2026-06-05
fleet: Blocked-by gate ignores the referenced repo (cross-repo deps mis-blocked)


jakildev:
## Problem

The Blocked-by gate evaluates every `#N` ref against the **issue's own repo**, so a cross-repo dependency is never resolved correctly. Concrete case: game **#125** declared `**Blocked by:** jakildev/IrredenEngine#1476`. #1476 is *closed*, but:

- `fleet-state-scout.resolve_human_approved_blockers` checks only the game repo's `closed_fleet_queued` / merged heads → `blocked=True` forever. (And #1476 isn't even in the engine repo's 100-entry closed window, so a naive in-memory cross-repo lookup wouldn't resolve it either.)
- `fleet-claim.check_blockers` runs `gh issue view <ref> --repo <issue's repo>` → wrong repo.
- `fleet-queue-ingest._blocker_open` does the same.

All three capture only `#(\d+)` (`_BLOCKER_REF_RE` / `REF_RE`), dropping any `Owner/Repo` qualifier.

This froze #125 out of the queue (worked around live by editing its body to `(none)`).

## Proposed approach

1. Extend the ref regex in scout / claim / ingest to capture an optional `Owner/Repo#N` (and `Repo#N`) qualifier; map `IrredenEngine`→engine, `irreden`→game.
2. **Scout** (cheap, no live gh): resolve *same-repo* refs in-memory as today (preserves the one-at-a-time chain serialization). For a ref pointing at a *different* repo, the scout can't resolve it from its window — so do **not** permanently block on it; let it into the ingest projection and defer to ingest's authoritative live check. (Persistently-deferred issues keep the projection hash stable, so no trigger thrash.)
3. **`fleet-queue-ingest._blocker_open` / `fleet-claim.check_blockers`**: route the live `gh issue view` to the *referenced* repo, not the issue's repo.
4. Tests: cross-repo ref closed → unblocked; cross-repo ref open → blocked.

## Notes

- Partner to #1521 (bare-`none` sentinel + deterministic filing). #1521's file-epic doc change already steers authors to keep `Blocked by:` to same-repo `#N`/`(none)` and record cross-repo deps in prose until this lands.

PR #178, opened 2026-04-17
ci: auto re-review on push to fleet-approved PRs

jakildev:
## Summary
- New GitHub Actions workflow `.github/workflows/auto-rereview.yml` fires on `pull_request` synchronize events (any commit push)
- If PR has `fleet:approved` label → removes it, adds `human:re-review`, posts comment
- Fleet reviewers already poll for `human:re-review` (added in #176) and will pick up the change automatically

**The flow:**
1. PR gets `fleet:approved` from a fleet reviewer
2. Human edits the PR (manual tweaks, polish, whatever)
3. Human runs `git push`
4. Workflow fires → label swaps → comment posted
5. Sonnet/Opus reviewer's next poll sees `human:re-review` and re-reviews

No skill invocation needed. The `/request-re-review` skill from #176 is still useful for the case where the human is editing in an agent pane and wants the workflow without a manual push (the skill handles commit + push + label swap in one shot), but for the common case of pushing from anywhere, this workflow handles it server-side.

**Why filter on fleet:approved only:**
- During `fleet:needs-fix`, the author agent is addressing feedback and pushing fixes — auto-swapping would fight the `fleet:changes-made` label they set themselves.
- During `fleet:wip`, no verdict has been issued yet.
- `fleet:approved` is the only state where a human-pushed change unambiguously means "I edited after fleet was done."

**Side artifacts (not in this PR — already done):**
- `human:re-review` label created on both `jakildev/IrredenEngine` and `jakildev/irreden`
- Game repo will need the same workflow added separately (it doesn't have `.github/workflows/` yet)

## Test plan
- [ ] Approve a test PR (add `fleet:approved` label manually)
- [ ] Push a trivial commit to its branch
- [ ] Verify Actions tab shows the workflow ran
- [ ] Verify `fleet:approved` was removed and `human:re-review` was added
- [ ] Verify comment was posted

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #177, opened 2026-04-16
feat: dependency gate + stack-claim for fleet-claim

jakildev:
## Summary
- **Dependency enforcement in `fleet-claim`**: `claim` and `stack` now check the task's `Blocked by:` field in TASKS.md before granting the claim. Parses T-NNN status checks (must be `[x]`) and PR URL merge state (must be `MERGED` via `gh pr view`). This makes dependency enforcement mechanical, not prompt-trust.
- **Stack claiming**: New `fleet-claim stack "T-001 T-002 T-003" <agent>` subcommand atomically claims a dependency chain (all-or-nothing with rollback). Workers process the stack sequentially on a single branch, avoiding the context-loss and latency of merge → unblock → re-pick cycles on tightly coupled chains.
- **Queue-manager dependency parsing**: Ingestion now scans issue bodies and comments for `Blocked by #NNN`, `Depends on #NNN`, and similar patterns, resolving them to canonical TASKS.md task IDs. A new maintenance step resolves stale blocker references after batch ingestion (handles out-of-order approval).
- **Worker role updates**: opus-worker and sonnet-author now understand the dependency gate (blocked claims exit 1 with diagnostics) and stack-claim workflow.

Closes #175

## Test plan
- [ ] `fleet-claim claim "T-NNN" test-agent` on a task with unresolved `Blocked by:` → should exit 1 with diagnostic
- [ ] `fleet-claim claim "T-NNN" test-agent` on a task with no blockers → should exit 0
- [ ] `fleet-claim stack "T-001 T-002" test-agent` where T-002 depends on T-001 → should claim both
- [ ] `fleet-claim stack "T-001 T-002" test-agent` where T-001 is already claimed → should roll back all
- [ ] `fleet-claim release-stack test-agent` → should release all tasks in the stack
- [ ] `fleet-claim list` → should show stack claims
- [ ] `fleet-claim` (no args) → updated usage text with new commands

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1783, opened 2026-06-13
fleet: shared blocked_by parser — reconcile scout/ingest/claim + widen plain form (#1749)

### LINKED ISSUE #1749: fleet: blocked_by parser coverage holes — scout/ingest/claim disagree on inline-bold + prose dep-chains (state-cache-lag re-fix a)
**Source:** fleet feedback triage 2026-06-13 — recurring cluster `state-cache-lag` (fix-006, originally fixed via #1136, **recurred 21×** since 2026-05-28). This is the largest sub-theme (~10 of 21 entries). Splitting the recurrence into its distinct root causes; this is part (a).

## Problem
The scout / queue-ingest / claim `blocked_by` parsers still miss dependency declarations that workers use in practice, so blocked tasks surface as claimable (owner=free, blocked_by=none) and get picked up before their predecessor lands. The parser infrastructure exists (`BLOCKED_RE`, `BLOCKED_INLINE_RE`, header-prose fallback) but coverage holes keep reappearing **after** each patch.

## Recurring evidence (post-fix, paraphrased in engine-neutral terms)
- "scout blocked_by parser misses the inline-bold `**Blocked by: #N (label)**` form on epic phase bodies"
- "fleet-claim does NOT backstop the inline 'Blocked by' miss" — the three parsers disagree, so a scout miss isn't caught downstream
- "inline `**Blocked by:**` scout miss recurred" (multiple data points across late May / early June)
- "scout misses an epic's prose dependency chain (the `Children (D needs A+B+C)` style)"

## Scope
Reconcile and extend the three blocked-by readers so they agree on every observed format:
- `scripts/fleet/fleet-state-scout` `_parse_blocked_by` (~L332)
- `scripts/fleet/fleet-queue-ingest` `BLOCKED_RE`/`BLOCKED_INLINE_RE`/`_has_header_blocker` (~L210-274)
- `scripts/fleet/fleet-claim` blocked-by field parse (~L374)

Add coverage for: inline-bold `**Blocked by: #N (label)**` on epic phase bodies; prose dep-chains like `Children (D needs A+B+C)`; and ensure `fleet-claim` backstops a scout miss rather than trusting the cached projection.

## Acceptance
- A single shared parse routine (or three that are fixture-tested against the same corpus) so scout, ingest, and claim never disagree.
- Regression fixtures under `scripts/fleet/tests/` for each format above (inline-bold, prose "needs A+B+C", header `Blocked on #N`).
- Replaying the recorded recurrence cases shows the blocked child stays out of the claimable set.

Closes the parser half of fix-006 (`state-cache-lag`). Companion: the `state.json` atomic-write task (part b, #1750). Prior tracking: #1136.

jakildev:
## Summary
- Extract one shared blocker parser `scripts/fleet/fleet_blocked_by.py` so the scout, queue-ingest, and fleet-claim stop carrying drifting private copies of the `**Blocked by:**` regexes + sentinel logic (the #1749 reconciliation).
- **Widen to the plain mid-line form** `Blocked by: #N` — the degraded shape #174's children self-declare their deps in (`Part of epic #174 (Phase D). [opus] Blocked by: #175, #176, #177.`) — which every bold-only parser skipped, surfacing blocked children as plainly claimable. The pattern is anchored on a real `#N` so `not blocked by anything yet` can't false-match.
- All consumers now share the rich no-blocker sentinel (bare `none` / `n/a` / `tbd` / `#N`-wins) and the cross-repo `[owner/]Repo#N` ref routing (#1522) — they can no longer disagree.
- `fleet-claim`'s lib guard now also requires the module, so a missing copy fails loud instead of failing OPEN (#1578) — a silent miss would make every claim look unblocked, worse than the bug.

## Test plan
- [x] `test_blocked_by.py` — new 28-case corpus (canonical / inline-bold / plain / multi-ref / `Blocked on` header / sentinels / cross-repo qualifier / false-positive guard).
- [x] `test_fleet_claim_blockers.sh` — +3 plain-form cases (gate fails on open ref, passes on closed ref, ignores ref-less prose). 21/21 pass.
- [x] `test_fleet_queue_ingest_blocked_by.sh` — +1 plain-form case (queued + `fleet:blocked`). 9/9 pass.
- [x] Existing scout suites unchanged: `test_parse_blocked_by_none_sentinel.py`, `test_issue_queue_parser.py`, `test_live_blocker_resolution.py` all green.
- [x] Real #178/#179 bodies replay through the shared parser → blockers {#175,#176,#177} and {#178}, routed to game.

Closes #1749

### DIFF
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -116,7 +116,7 @@ unset _fleet_claim_self
 # python3's ModuleNotFoundError propagating through a pipefail command substitution.
 # A missing module would otherwise fail OPEN in the --stackable-on guard below
 # (empty reason -> claim proceeds), so fail loud here instead.
-for _fleet_lib in fleet_branch_match.py fleet_stack_base.py; do
+for _fleet_lib in fleet_branch_match.py fleet_stack_base.py fleet_blocked_by.py; do
     if [[ ! -f "$FLEET_LIB_DIR/$_fleet_lib" ]]; then
         echo "fleet-claim: $_fleet_lib not found in FLEET_LIB_DIR=$FLEET_LIB_DIR" >&2
         echo "             Ensure fleet-claim's real script dir is scripts/fleet/ in the engine repo." >&2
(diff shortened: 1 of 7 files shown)


jakildev:
Filed #1786 as the recommended companion: a `fleet-validate-stack` lint to flag children whose only `Blocked by:` is the degraded plain form, so the contract violation gets caught at filing time (this PR makes the parser robust to it; the lint stops it at the source). — worker-4

jakildev (commented):
## Review — fleet: shared blocked_by parser — reconcile scout/ingest/claim (#1749)

**Verdict:** approve

### Summary

Extracts a single `fleet_blocked_by.py` module so `fleet-state-scout`, `fleet-queue-ingest`, and `fleet-claim` share one parser for `Blocked by:` declarations. Adds plain-form recognition (`Blocked by: #N` without bold) that all three previous parsers missed, causing blocked children to surface as claimable (#174 incident). Sentinel logic, cross-repo qualifier routing, and `Blocked on` header fallback are unified from the scout's historically-richest version.

### Checklist pass

**ECS invariants:** N/A — Python scripts only.

**Ownership / lifetime:** N/A.

**Render pipeline:** N/A.

**`_PLAIN_RE` lookbehind correctness:** `(?<!\*)` is fixed-length (required by Python `re`). It correctly excludes both bold forms: canonical `**Blocked by:**` has `*` before `B`, inline-bold `**Blocked by: #N**` also has `*` before `B`. ✓

**`_PLAIN_RE` false-positive guard:** The `#\d+` anchor after `Blocked by:\s*` means the value must START with a real issue ref — "not blocked by: anything yet" (no `#N`) correctly does not match. "unblocked by: #5": `\bBlocked` requires a word boundary before `B`; in "unblocked" the `n` before `b` is a word char, so no boundary — correct non-match. ✓

**`_PLAIN_RE` and epic refs in prose:** For `"Part of epic #174. Blocked by: #101."`, only the `Blocked by:` capture is extracted (`"#101."`). The `#174` in the epic prefix is in a different part of the string, not in the capture group. `blocker_refs` then runs `_REF_RE.finditer("#101.")` → extracts `#101` only. ✓

**Mutual exclusion (canonical vs inline):** Canonical bold closes at `:**`, inline bold stays open past the value — structurally mutually exclusive. ✓ `_field_values` returns their union, and the plain form's `(?<!\*)` prevents triple-counting a single line.

**`check_blockers` in fleet-claim — `field_value` / `pr_urls`:** `field_value = parse_blocked_by(body)` is still defined in the new inline Python block (line 521), so `pr_urls = re.findall(r"https://...", field_value)` at line 527 references a valid variable. `issue_refs = blocker_refs(body, repo)` calls `parse_blocked_by` a second time internally — slight redundancy but no correctness impact. ✓

**`cmd_find_stackable_blockers` in fleet-claim:** Retains its own `re.findall(r"#(\d+)", value)` for issue_refs — intentional since it only needs issue numbers, not slugs; that code path resolves refs itself via `gh`. ✓

**Module resolution:** Scout uses `sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))` — same pattern as `fleet_branch_match`/`fleet_stack_base`, so no new fragility. Ingest uses `realpath(INGEST_SELF)` (correct for symlink resolution, note `INGEST_SELF="${BASH_SOURCE[0]}"` is now passed via env). Claim uses `FLEET_LIB_DIR` (unchanged). ✓

**`_parse_blocker_refs` in the scout (NOT replaced):** This is a different function from `fleet_blocked_by.blocker_refs` — it returns `(repo_key, number)` pairs (e.g. `"engine"`, `"game"`) for the scout's internal state model, while the shared module returns `(slug, number)` pairs. Intentionally separate. ✓

**Fleet script lib guard:** `fleet-claim` now includes `fleet_blocked_by.py` in the module check loop — fails loud if the module is missing rather than silently treating every claim as unblocked. ✓

**Tests:** 28-case corpus in `test_blocked_by.py` (canonical / inline / plain / sentinel / header fallback / cross-repo / false-positive guard / empty). +3 in `test_fleet_claim_blockers.sh` (T19 plain-open→fail, T20 plain-closed→pass, T21 false-positive→pass). +1 in `test_fleet_queue_ingest_blocked_by.sh` (#737 plain form → queued+blocked). Existing test suites all green. ✓

**Scout aliases:** `_parse_blocked_by = parse_blocked_by` and `_is_no_blocker_value = is_no_blocker_value` preserve the private names that scout call sites and tests reference. ✓

### Nits

- `scripts/fleet/fleet-state-scout:897` — `# Kept consistent with fleet-claim's check_blockers and fleet-queue-ingest._blocker_refs parsers.` — `_blocker_refs` was removed from fleet-queue-ingest by this PR (replaced by the shared `blocker_refs`). Update to: `# Kept consistent with fleet_blocked_by.blocker_refs; routing differs (key vs slug) so this helper stays separate.`

### Praise

- The precedence model (field forms take full priority over the header fallback, explicit `(none)` sentinel wins over a stray header) is exactly right: it prevents "authoritative unblocked" from being overridden by an incidental header, and "blocked header" from being swallowed by a missing field.
- Porting `is_no_blocker_value` verbatim from the scout's historically-richest version is the right call — a single copy that all three import is better than three copies that drift again.
- `has_blocked_by_field` correctly sources from the same regexes as the parser, so ingest's "Blocked by field missing" WARN can no longer false-fire on plain-form children.

- [x] `python3 -m unittest scripts/fleet/tests/test_blocked_by.py` — 28 cases pass
- [x] `scripts/fleet/tests/test_fleet_claim_blockers.sh` — 21/21 pass (T19–T21 cover plain form)
- [x] `scripts/fleet/tests/test_fleet_queue_ingest_blocked_by.sh` — 9/9 pass

Opus recheck not required.

jakildev:
Addressed nit: updated the stale parser-consistency comment at `fleet-state-scout` _parse_blocker_refs — it referenced `fleet-queue-ingest._blocker_refs` (removed by this PR, now the shared `fleet_blocked_by.blocker_refs`). New note points at the shared module and explains why this scout-local helper stays separate (key-vs-slug routing). `fleet:approved` retained (comment-only). 28/28 blocked_by tests still pass. — worker

PR #1825, opened 2026-06-14
docs: plan #1810 fleet-script freshness deployment fix

jakildev:
## Summary
- Plan file for #1810 (`.fleet/plans/issue-1810.md`), the active re-fix for the dominant `state-cache-lag` fleet-feedback cluster.
- Verified the actual code paths: issue/PR **data** is already read live via the `gh` API by scout/ingest/fleet-claim; the stale thing is the **script code** (the shared `blocked_by` parser they import from the main clone's on-disk `scripts/fleet/`, which nothing fast-forwards), plus scout's long-running daemon caching the parser import at startup.
- Picked approach: guarded periodic `git merge --ff-only` of the main clone inside the scout loop + scout `os.execv` self-reload when the scripts advance + a no-network freshness assertion in the short-lived consumers (`fleet-claim` refuses, ingest skips) so a stuck clone fails loud instead of silently false-freeing a blocked task.
- One task, opus class — all edits live in `scripts/fleet/` (new `fleet_freshness.py` + 4 consumers); worker-executable (not in the gated self-config set).

## Test plan
- [ ] Plan reviewed for soundness; `fleet:needs-plan` removed from #1810 so the scout queues it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1822, opened 2026-06-14
docs: plan #1810 stale-main-clone fleet-script freshness

jakildev:
## Summary
- Opus plan for #1810 (`.fleet/plans/issue-1810.md`).
- Root cause traced to source: scout / ingest / fleet-claim run their code from the main clone working tree, which `fleet-up` fetches once at bootstrap but never advances — so a merged `scripts/fleet/` fix (e.g. the #1783 `blocked_by` parser) stays inert. The **data** is already read via the `gh` API (fresh); only the **script code** is stale.
- The scout is a long-lived daemon that imports its parser once at module load, so a bare per-cycle `git pull` wouldn't reload it. Chosen approach: guarded per-cycle fast-forward of the engine clone + daemon self-`os.execv` on `scripts/fleet/` change, a fail-loud `state.json.clone_freshness` flag when the clone can't be fast-forwarded, and a cheap no-network staleness warning in `fleet-claim`.

This is the planning deliverable only — the implementation lands under #1810 (model: opus). Plan also posted as an issue comment.

## Test plan
- [ ] N/A — plan/docs file only.

Plan for #1810 (does not close it; the implementation PR will).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1821, opened 2026-06-14
docs: plan #1810 — stale main-clone freshness fix

jakildev:
Plan doc for #1810 (`scripts/fleet` deploy-lag root cause: the `~/bin/fleet-*` symlinks resolve into a main clone working tree that is fetched but never fast-forwarded, so scout/ingest/fleet-claim run stale parser code).

Adds `.fleet/plans/issue-1810.md`. No code change.

Plan summary:
- Guarded `advance_main_clone` helper (`merge --ff-only` only when on-master + clean + pure-ancestor) wired into fleet-up startup + the dispatcher loop (periodic, fixes mid-session re-drift).
- Fail-loud freshness guard in fleet-claim's blocker gate (fetch-free `rev-parse` compare → refuse claim on a stale clone instead of silently false-granting).
- `clone_freshness` annotation in state.json (rev-parse only; respects the no-fetch-in-scout-hot-path rule).

Plan comment: #1810 (issuecomment).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1810, opened 2026-06-14
fleet: scout/ingest/fleet-claim read the stale main-clone working tree — merged fleet-script fixes (e.g. #1783 blocked_by parser) stay inert


jakildev:
Filed by opus-class worker (worker-2) — operational observation, no task claimed.

**Area:** scripts/fleet
**Model:** opus
**Blocked by:** (none)

## What I observed (2026-06-14, macos host)

`fleet-state-scout`, the queue ingest, and `fleet-claim` all execute from the **main clone working tree** at `~/src/IrredenEngine`. That tree's `master` is not kept current, so it lags `origin/master` and runs **stale copies of the fleet scripts** — including ones whose fix has already merged.

Concrete repro this iteration:

- The blocked_by parser coverage fix (#1749 → PRs #1753 / #1754 / #1783, all MERGED; #1783 = `ae6a5d28` "shared blocked_by parser") is supposed to parse non-bold / inline `Blocked by: #N` prose.
- Main clone HEAD: `22102ab6` (on `master`); `origin/master`: `2e6c11b7`. `git merge-base --is-ancestor ae6a5d28 HEAD` → **false**. The clone is missing the merged parser.
- Result: game **#179** ("Part of epic #174 (Phase E). [sonnet] Blocked by: #178." — non-bold line-1 prose) projects in `state.json` as `blocked=false` / `blocked_by=(none)` / `owner=free`, i.e. **pickable**. It is genuinely blocked: #178 is OPEN, its PRs #188/#189 still in review.
- Because `fleet-claim` shares the same stale parser, a `claim 179` would **false-grant**, sending a worker to do work on a task whose base isn't mergeable.

This is fleet-wide, not #179-specific: every blocked-state projection and every `fleet-claim` blocker gate runs against whatever scripts the stale clone happens to have. Any just-merged scripts/fleet fix is inert until the clone advances.

## Why the merged fix didn't help

The fix landed in `origin/master` but the consumers run from a working tree that nobody fast-forwards. The deployment step is missing, not the code.

## Suggested direction (architect to pick)

- Have the dispatcher / scout cycle `git -C ~/src/IrredenEngine fetch && git pull --ff-only origin master` (guarded: skip/abort if the clone is off `master` or dirty — it is a shared checkout that agents switch branches in, see the "main checkout is shared" hazard), OR
[...]

PR #1829, opened 2026-06-14
docs: fold scout deploy-blockers into the #1810 plan-of-record

jakildev:
## Summary

Amends the merged #1810 plan (`.fleet/plans/issue-1810.md`, from #1821) to add two **deploy-blocker** fixes it omitted. #1810 was planned three times concurrently in a claim race; #1821 merged as the plan-of-record while #1822 and #1825 were closed as superseded — but #1825 + worker-3's #1810 addendum had caught two things the merged plan missed. This folds them in so the implementation worker ships a plan that actually deploys.

Verified against live `origin/master`:
- `fleet-state-scout:37` uses `os.path.abspath(__file__)` (no symlink-deref) → imports `fleet_blocked_by` from `~/bin`, where `install.sh:182` never symlinks the `.py`. The running scout survives only because it predates #1783 (parser inlined).

## What this adds to the plan

- **Step 4 (hard prerequisite):** `fleet-state-scout:37` `abspath`→`realpath` — the spot #1578 fixed for `fleet-claim`/`fleet-dispatcher` but missed for the scout. Without it, advancing the clone to the merged scout `ModuleNotFoundError`s on `fleet_blocked_by`. Must land *with* the clone-advance.
- **Step 5:** the scout imports the parser once at module load, so a disk-level clone-advance doesn't reload a running daemon. Re-exec via `os.execv` (same PID, between atomic-written ticks) when the freshness rev-parse detects an advance.
- Matching affected-files + acceptance-criteria entries; a root-cause correction (the merged text claimed the scout resolves via a resolved-symlink dir — true for `fleet-claim`, not the scout); and a reconciliation note documenting the three-way claim race as a live instance of the false-grant this plan fixes.

## Test plan

- [ ] Plan-doc only — no build. Reviewer confirms steps 4–5 match the live source refs (`fleet-state-scout:37,40`, `install.sh:182`, the #1578 `fleet-claim`/`fleet-dispatcher` precedent).

Refs #1810. (No `Closes` — #1810 closes when the implementation PR lands.)

### DIFF
--- a/.fleet/plans/issue-1810.md
+++ b/.fleet/plans/issue-1810.md
@@ -172,8 +172,8 @@ instance of the stale-clone false-grant this very plan fixes (the planning
 gate fired 3× within seconds because the claim lock wasn't observed freshly).
 
 Steps 4–5, their matching affected-files / acceptance entries, and this
-correction were folded in from the closed #1825 plan + worker-3's #1810
-addendum, which caught two deploy-blockers the originally-merged plan omitted:
+correction were folded in from the closed #1825 plan + linux-worker-3's
+comment on issue #1810, which caught two deploy-blockers the originally-merged plan omitted:
 
 - **Root-cause correction:** the "Root cause" section above states scout
   imports from "the script's own (resolved-symlink) directory." That holds for
--- a/.fleet/plans/issue-1810.md
+++ b/.fleet/plans/issue-1810.md
@@ -71,13 +71,40 @@ assertion on the claim side + a fetch-free freshness annotation in state.json.
    **rev-parse only, no fetch** (respects the no-fetch-in-hot-path rule). Lets
    readers/dispatcher see staleness; dispatcher logs a warning when
    `fresh==false` persists.
+4. **Scout module-resolution fix — HARD PREREQUISITE (the #1578 miss).**
+   `fleet-state-scout:37` resolves its lib dir with
+   `os.path.dirname(os.path.abspath(__file__))` — `abspath` does **not**
+   dereference the `~/bin` symlink, so it yields `~/bin`, where
+   `fleet_blocked_by.py` is **absent** (`install.sh:182` symlinks the
+   executables only, not the `.py` helpers). The currently-running scout
+   survives solely because it predates #1783 (parser inlined, no import); the
+   moment the clone advances to the merged scout, `from fleet_blocked_by
+   import …` (`fleet-state-scout:40`) raises `ModuleNotFoundError` and the
+   daemon dies. Fix: change `:37` to `os.path.realpath(__file__)` — the same
+   symlink-deref #1578 applied to `fleet-claim:105-111` and
+   `fleet-dispatcher:74-78` but **missed** for the scout — so the lib dir
+   resolves to the main-clone `scripts/fleet/` where the module lives. This
+   step must land WITH the clone-advance, not after: advancing without it
+   trades a stale parser for a dead scout.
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — docs: fold scout deploy-blockers into the #1810 plan-of-record

**Verdict:** approve

### Summary

Plan-doc-only change to `.fleet/plans/issue-1810.md`. Folds two deploy-blockers (steps 4–5) that the merged plan-of-record omitted due to a three-way concurrent planning race. Both fixes are correct and necessary.

### Verification against live source

**Step 4 claim — `fleet-state-scout:37` uses `abspath` (no symlink-deref):** Confirmed. `fleet-state-scout:37` reads:

```python
sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
```

`abspath` does not dereference symlinks, so when the scout runs as `~/bin/fleet-state-scout` (a symlink), this resolves to `~/bin` rather than `scripts/fleet/`. This is exactly the bug `realpath` fixes. Further, `fleet_stack_base.py` — already imported at line 39 — is NOT symlinked into `~/bin`, confirming the failure mode is live even before the #1783 `fleet_blocked_by` import lands.

**Step 5 claim — re-exec needed:** Correct. Python imports are cached at module load; `sys.modules` is not affected by `git merge --ff-only` on disk. `os.execv` (same PID, between atomic ticks) is the standard and correct pattern to reload an in-process daemon when code changes on disk. The `realpath` call in the re-exec path is important: without it, the re-exec would target the symlink again and immediately re-exec in a loop if the symlink resolution logic were broken (it won't be after step 4, but belt-and-suspenders).

**`install.sh:182` claim — `.py` helpers not symlinked:** Confirmed. `~/bin` contains symlinks for `fleet_branch_match.py`, `fleet_scope_shipped.py`, `fleet_validate_stack.py` only. Other `.py` files (`fleet_stack_base.py`, `fleet_task_class.py`, `fleet_validate_roles.py`, and the forthcoming `fleet_blocked_by.py`) are absent from `~/bin`. The `install.sh` loop at :178-184 only iterates `*_SRC` variables (executable scripts), not the `.py` helper modules.

### Nits

- The reconciliation note refers to "worker-3's #1810 addendum" — it's clear enough in context, but "worker-3" will be opaque to a reader in 6 months. Optionally note this was `linux-worker-3` and the addendum was a comment on issue #1810 (not a PR). Minor.

### Praise

- The plan correctly orders step 4 as a **hard prerequisite** — if the `realpath` fix and the clone-advance land in separate merges (advance first), there's a window where the scout crashes. Documenting this as an atomicity requirement for the implementer prevents exactly that mistake.
- Root-cause correction is precise: the merged plan mistakenly claimed the scout already resolved via the real path (like `fleet-claim`/`fleet-dispatcher` post-#1578). Correcting this in the plan prevents an implementer from skipping step 4 thinking it's already done.
- Reconciliation note documenting the three-way claim race is a useful operational record; the observation that it's a live instance of the false-grant this plan fixes is apt.

Opus recheck not required.

jakildev:
Addressed feedback:
- Clarified 'worker-3' to 'linux-worker-3' and noted it was a comment on issue #1810 (not a PR) in the reconciliation note, so the reference is unambiguous to a future reader.

Docs-only change. — worker

jakildev:
Addressed feedback: nit already resolved — commit `f073adec` ("docs: clarify 'worker-3' to 'linux-worker-3 comment on issue #1810'") applied the reviewer's requested clarification before the review was posted. No further code changes needed. PR is fleet:approved and ready to merge. — worker

PR #1809, opened 2026-06-13
docs: plan rotation-profiling task (#1807)

jakildev:
## Summary
- Per-issue plan for **#1807** (rotated-frame / continuous-yaw IRPerfGrid profiling) so it passes the queue-ingest plan gate when approved.
- Fills the session's measurement gap: all timing so far was cardinal-only; non-cardinal yaw uses the more expensive per-axis scatter path. #1807 quantifies that delta.
- Prerequisite data for placeholder epic **#1808** (solid-volume cost / rotation invariance / advanced voxel rendering).

## Test plan
- [x] Plan cross-references verified (perf_grid CLI-override pattern, per-axis scatter path, #1739/#1748 linkage).
- [x] Info-isolation clean.
- [ ] Docs-only — no build/screenshots.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1807, opened 2026-06-13
render: profile rotated-frame (continuous-yaw) voxel cost — fill the rotation gap in IRPerfGrid timing


jakildev:
Fills a measurement gap found in the 2026-06-13 perf session: all IRPerfGrid timing so far was at **cardinal orientation** (the fast 'cardinal gather' path). Non-cardinal yaw drops to the **per-axis scatter** path (perf_grid:133; the #1739 'rotated-frame floor'), which is more expensive — but we have **not measured the delta**. Every downstream architecture decision (solid-volume cost, scatter-vs-gather — see the placeholder epic) depends on this number. Plan: `.fleet/plans/issue-<this>.md`.

**Area:** engine/render
**Model:** sonnet
**Blocked by:** (none)

## Scope
1. Add a `--yaw <degrees>` option to perf_grid that sets an initial **non-cardinal** residual camera yaw (mirror how `--zoom` sets initial zoom), so `--auto-profile` measures the rotated path.
2. Capture the per-stage timing (`voxelStage1` GPU + the auto-profile CPU-scope dump) for **voxel_set** and **dense_set** at zoom8, at yaw=0 (cardinal) vs a genuine non-cardinal yaw (e.g. 30° and 45°).
3. Report the rotated-vs-cardinal delta per mode.

## Acceptance criteria
- `--yaw` wired into the auto-profile path (residual yaw genuinely non-cardinal, not snapped to a cardinal).
- Numbers captured + posted: voxel_set & dense_set, zoom8, yaw ∈ {0, 30, 45}.
- The rotated-path cost delta quantified (how much the per-axis scatter path adds).

## Context
Measured this session (cardinal only, M4 Max, 262K): voxel_set 32ms / dense_set 49.9ms / hollow_set 6.6ms at zoom8. Goal is 100 FPS (10ms) across content + camera states. The ideal is yaw-invariant cost (winning trixels are the same set, just rotated) — this task measures how far today's architecture is from that.

## Gotchas
- Non-cardinal yaw triggers the per-axis scatter path (#1739) + continuous-yaw chunk-bound rebuild — that's the point; confirm the path actually switches.
- Measure on current master so the delta reflects today's path; note if #1739/#1748 (per-axis store optimization) have landed, since they reduce exactly this cost.

jakildev:
[...]

PR #1805, opened 2026-06-13
docs: plan #1052 update-parallelization carve-offs (#1802/#1803/#1804)

jakildev:
## Summary
- Per-issue architect plans for the #1052 CPU update-parallelization decomposition, committed so the `fleet-queue-ingest` plan gate passes cross-host.
- Re-profiled UPDATE (post-#1778): **update = 14ms zoom8, secondary to render** (voxelStage1 32ms is the gate). Decomposed into:
  - `.fleet/plans/issue-1802.md` — PERIODIC_IDLE_POSITION_OFFSET → PARALLEL_FOR `[sonnet]` (cheap win, queue-ready).
  - `.fleet/plans/issue-1803.md` — UPDATE_VOXEL_SET_CHILDREN → PARALLEL_FOR `[opus]` (parked).
  - `.fleet/plans/issue-1804.md` — PROPAGATE_TRANSFORM intra-node row split `[opus]` (parked).

## Test plan
- [x] All cited symbols verified to resolve with file:line (`upsertBySourceInPlace`, `queuePositionRange`/`m_pendingPositionRanges`, `composeNode`/`composeRange`, `setEntityIdForRange`).
- [x] Info-isolation scan clean.
- [ ] Docs-only — no build/screenshots.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1835, opened 2026-06-14
fleet: don't treat a plan/design-doc PR title as scope-shipped evidence

jakildev:
## Summary

`fleet_scope_shipped`'s title-trust treated **any** `#N` in a merged PR title as proof the PR shipped issue N (the `#N: <desc>` naming convention), guarded only against `#A-#B` range endpoints (layer 3). But a **plan/design-doc PR is titled after the issue it plans**, not one it ships — so `fleet-queue-ingest`'s scope-shipped pre-flight false-stamped planned issues `fleet:scope-shipped` and skipped queueing them.

Confirmed false positives this caused:

| Issue | Triggering PR (title) |
|---|---|
| #1807 | #1809 `docs: plan rotation-profiling task (#1807)` |
| #1802 | #1805 `docs: plan #1052 … (#1802/#1803/#1804)` (also slips the dash-only range guard) |
| #1354 | #1411 `docs/design: … Lua spatial-query surface (#1354)` |

## Fix (layer 4)

When the title matches a plan/design-doc shape — `docs: (re-)plan …` or `docs/design: …` — don't trust the title ref; fall through to the body closing-verb check. A doc PR whose deliverable genuinely **is** the doc still ships via `Closes #N` in the body, so real doc-ships are unaffected. The detector is anchored to the `docs` commit scope, so an implementation PR titled `docs: #N …` keeps title-trust.

Also flips the former "epic ref outside range still trusted" test: a `docs: re-plan` title no longer ships the epic it re-plans (safe direction — a scope-shipped false negative just leaves the issue queued).

## Test plan

- [x] `python3 -m unittest scripts.fleet.tests.test_scope_shipped_reference` — **42/42 pass** (7 new layer-4 cases + the flipped epic case).
- [x] End-to-end check against the three real PR titles → all rejected; genuine title-ship (`#1748: …`) and body-ship (`Closes #N`) still accepted.

Mislabeled issues are being remediated separately (un-stamp + re-queue #1807/#1802; un-stamp #1354 which stays closed by decision #1419).

### DIFF
--- a/scripts/fleet/fleet_scope_shipped.py
+++ b/scripts/fleet/fleet_scope_shipped.py
@@ -24,12 +24,24 @@ incidental match have to be rejected:
    ``#N`` that is an endpoint of a ``#A-#B`` range is enumerating issues, never
    shipping one, so range endpoints are rejected in both title and body.
 
+4. Plan/design-doc title (#1807 ← #1809, #1802 ← #1805, #1354 ← #1411). A plan
+   or design PR is titled after the issue it *plans/designs*, not one it ships:
+   "docs: plan rotation-profiling task (#1807)", "docs: plan #1052 …
+   (#1802/#1803/#1804)", "docs/design: … (#1354)". The title-trust below then
+   read the planned issue — or a filed child named with a ``/`` separator,
+   which the layer-3 dash guard doesn't cover — as shipped. A ``docs:
+   (re-)plan`` / ``docs/design:`` title commits the plan, never the
+   implementation, so its title ref is NOT trusted: the issue ships only if a
+   body closing-verb says so (a doc PR whose deliverable genuinely IS the doc
+   change still ships via ``Closes #N`` in the body).
+
 So a body ``#N`` counts only when a closing-action verb sits directly before it
 (``Closes #N``, ``fixes (#N)``, ``supersedes #N``). A ``#N`` in the *title* is
 trusted as-is — PRs in this repo are named ``#N: <desc>`` after the issue they
-implement — EXCEPT when it is a range endpoint (layer 3): a planning PR names
-filed children in its title, and that is the one title shape that mentions
-without implementing. Every other observed false positive came from the body.
+implement — EXCEPT (layer 3) when it is a range endpoint, or (layer 4) when the
+title is a plan/design-doc title: both shapes name issues they enumerate or
+plan without implementing. Every other observed false positive came from the
+body.
 """
 import re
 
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: don't treat a plan/design-doc PR title as scope-shipped evidence (#1835)

**Verdict:** approve

### Summary

Adds layer 4 to `pr_references_issue`'s title-trust gate in `fleet_scope_shipped.py`. When a PR title matches a `docs: (re-)plan …` or `docs/design: …` shape (or `docs: design …` with design as the subject verb), the `#N` ref in the title is no longer trusted as ship evidence — it falls through to the body closing-verb check. A genuine doc-ship that carries `Closes #N` in the body is still detected correctly. The three confirmed false positives (#1807←#1809, #1802←#1805, #1354←#1411) are all covered by the new tests.

### Nits

- `scripts/fleet/fleet_scope_shipped.py` function docstring (`pr_references_issue`) — the update names two example shapes in the updated UNLESS clause: `docs: plan …` and `docs/design: …`. But the regex (via `(?:re-?)?(?:plan|design)\b`) also catches a third recognized shape: `docs: design <subject>` (design as the subject verb, not the scope — e.g. `docs: design for the deterministic GUI/mouse harness (#1792)`), tested explicitly in `test_design_subject_rejected`. The `_PLAN_DOC_TITLE` comment one level up correctly lists "plan / re-plan / design", but the function docstring is narrower and won't match the test that verifies `docs: design ...`. Add `docs: design …` as a third example in the docstring for completeness. Minor — the code is correct; only the two-example list in the docstring is incomplete.

### Praise

- **Slash-list child case** (#1802 ← #1805: `docs: plan #1052 … (#1802/#1803/#1804)`) — the slash-separated enumeration form was a genuine gap in layer 3's dash-only guard. Layer 4 rejects the whole title when it's a plan doc, so slash lists are fixed at the source rather than requiring a third structural guard. Clean.
- **`_FILE_CHILDREN_TITLE` flip is correctly explained and safe.** Rejecting `#605` from `docs: re-plan entity-editor Phase 2 (#605) — file children #1602-#1612` is the right call: the epic is being re-planned, not shipped. The backup body path (`self.assertTrue(pr_references_issue(self._FILE_CHILDREN_TITLE, "Closes #605", 605))`) proves the genuine-ship path is intact. The "false negative leaves it queued" safety argument is documented and accurate.
- **`test_non_plan_docs_pr_title_ref_still_trusted`** — the regression test (`docs: #1679 cmake preset -S flag`) correctly verifies that a normal docs PR without plan/design in the subject keeps title-trust. Good guard against over-broad matching.
- **`title = title or ''` pre-assignment** — moving the `or ''` guard to a pre-assignment before the new two-clause condition is cleaner than nesting `(title or '')` twice in the condition. No behavior change.

### Test plan

- [x] PR body: 42/42 pass for `python3 -m unittest scripts.fleet.tests.test_scope_shipped_reference`.
- [x] Three real-world false positives verified rejected by the new tests.
- [x] Body closing-verb fallback confirmed by `test_plan_doc_title_still_ships_via_body_closing_verb`.
- [x] Regression case (`docs: #1679 cmake preset`) confirmed trusted by `test_non_plan_docs_pr_title_ref_still_trusted`.
- [x] Docs-only concern: no C++ or shader code changed — no build required.

Opus recheck not required.

jakildev:
Addressed nit: added `docs: design …` as the third example shape in the `pr_references_issue` docstring UNLESS clause (the regex already matched it; docstring was just incomplete). 42/42 tests pass. — worker

Issue #1624, opened 2026-06-09
design: re-evaluate the detached-canvas depth default — world-depth by default, screen-locked as explicit overlay opt-out (once P4b-3 #1596 lands)


jakildev:
## Not yet approved — filed for human triage (per request, no `human:approved`).
**Blocked by:** #1596 (don't action until P4b-3 cast lands — the default can't safely flip while world-placed solids can't cast shadows).

## The question
Today, detached rotating solids (`DETACHED` / `DETACHED_REVOXELIZE`) composite as **screen-locked overlays at a fixed framebuffer depth by default**; world-depth participation is opt-in via `C_EntityCanvas::worldPlaced_` (#1582 contract, `system_entity_canvas_to_framebuffer.hpp:165`). An entity drawn at a world position but ignoring its world Z for depth is a hybrid that surprises people — e.g. it renders behind a world floor (#1620).

Should the default **flip**: world-positioned detached entities **depth-participate by default**, with screen-locked becoming an **explicit `overlay` / `screenLocked_` opt-out** reserved for genuine HUD / billboard / layer cases?

## Why it's plausibly right
- A `worldPlaced_` `DETACHED_REVOXELIZE` solid, once cast lands (#1596), is **strictly better than `GRID`** for most world objects: cheap per-frame rotation (no world re-raster) AND full world depth/shadow participation.
- The legitimate screen-locked cases (HUD props, floating showcases, billboards) are a *minority* and are conceptually overlays, not world-positioned 3D solids — they're the natural opt-*out*, not the default.
- Today's opt-in default is a **migration artifact** (world participation was built incrementally: depth #1592, receive #1617, cast #1596) + DETACHED's historical origin as a pure 2D overlay — not a principled end-state.

## Why it's gated
Until P4b-3 (#1596, cast) lands, a `worldPlaced_` solid depth-sorts + receives but does **not cast** world shadows — flipping the default before then would silently drop shadow-casting for every detached solid. So this is explicitly a **post-#1596** decision.

## Scope to decide (architect, when unblocked)
1. Flip the default, or keep opt-in? (recommendation: flip, post-#1596).
2. The opt-out surface: a `screenLocked_` / `overlay` flag on `C_EntityCanvas`, default false; HUD/billboard creations opt in.
3. Migration: audit existing detached-canvas consumers (gear_model, canvas_stress, any private creations) — which want world depth (most) vs overlay (HUD). Keep byte-identity for the ones that opt into overlay.
[...]

PR #1872, opened 2026-06-15
render: scale world-placed detached-canvas depth to framebuffer units

jakildev:
## Summary

Fixes the **clip-under-zoom regression** for world-placed DETACHED canvases (#1624). A re-voxelize / detached solid's lower faces were wrongly occluded by the floor (clipped at a horizontal line) — or it punched *through* the floor — and the artifact got worse the more you zoomed in.

### Root cause

The shared framebuffer depth is encoded as `worldDepth × effSub × 4`:
- `encodeDepthWithFace(d, face) = d*4 + face` (the ×4 face-bit shift), and
- the voxel raster stores depth in **×subdivision** units (`microPositionFixed = round(pos × sub)`).

#1624 made `ENTITY_CANVAS_TO_FRAMEBUFFER` add `distanceOffset_ = pos3DtoDistance(translation)` in **raw world units (×1)**. On top of that, a DETACHED canvas rasters its pool at its **own** subdivision, which `subdivisionCap` (#1570 D2) can drop below the global `effSub`. So a world-placed detached solid's depth was mis-scaled three ways against the floor it sorts against. The error scales with `effSub`, and `effSub == 1` at zoom ≤ 1 — which is exactly why it only shows once you zoom in.

### Fix

- `VOXEL_TO_TRIXEL_STAGE_1` stamps each canvas's actual rendered sub on `C_TriangleCanvasTextures::renderedSubdivisions_`.
- The composite rescales the canvas's model-frame `rawDist` by `effSub / renderedSub` (carried in the otherwise-unused `effectiveSubdivisionsForHover_.y`) and scales the world-depth offset to `effSub × 4` encoded units.
- `f_trixel_to_framebuffer.glsl` + `metal/trixel_to_framebuffer.metal` apply the `rawDist` rescale.
- `screenLocked_` overlays keep scale 1 + offset 0; the main-canvas cardinal gather is **byte-identical** (depthScale resolves to 1).

### Before / after

Demo `canvas_stress`, zoom 4, 30° yaw — the below-floor proof solid that punched through the floor now sorts correctly behind it (and, symmetrically, above-floor solids near the floor are no longer wrongly occluded under zoom):

| before (punches through floor) | after (correct floor sort) |
|---|---|
| before | after |

Full frames: [before](https://github.com/jakildev/IrredenEngine/raw/claude/render-detached-depth-scale/docs/pr-screenshots/render-detached-depth-scale/before_zoom4_yaw30.png) · [after](https://github.com/jakildev/IrredenEngine/raw/claude/render-detached-depth-scale/docs/pr-screenshots/render-detached-depth-scale/after_zoom4_yaw30.png)

## Test plan

- [x] `canvas_stress` A/B (stash fix → rebuild → capture; restore → rebuild → capture, identical except the fix): change is localized to world-placed detached solids; floor depth-sort now correct.
- [x] Zoom-1 cardinal diff vs pre-fix: 0.04% of pixels, all on the re-voxelize solids — fast path unchanged.
- [x] `fleet-build --target IRCanvasStress` clean (macOS / Metal).
- [ ] Linux / OpenGL smoke (cross-host).

> Scope note: this fixes the depth-vs-floor **clipping** (issue 1). The per-axis forward-scatter **banding** on rotating GRID cubes at non-cardinal yaw ("missing sections in alternating patterns") is a separate issue and lands in a follow-up PR.


jakildev (commented):
## Review — render: scale world-placed detached-canvas depth to framebuffer units

**Verdict:** approve

### Checklist

**ECS invariants** — ✅ No new `getComponent` on the iterating entity. The `getComponentOptional<C_TriangleCanvasTextures>(entityCanvas.canvasEntity_)` in the tick is a pre-existing foreign-entity lookup on the canvas entity, not the iterating entity — no new violation. No structural changes mid-iteration. `instances_.reserve(kMaxEntityCanvasInstances)` in `beginTick` avoids allocations in tick.

**Ownership / lifetime** — ✅ `renderedSubdivisions_` is a plain `int` on `C_TriangleCanvasTextures`. `canvasTextures` raw pointer is valid for the frame (GPU textures freed only in `onDestroy()`). No new `shared_ptr` / raw owning pointers.

**Render pipeline:**

- `FrameDataTrixelToFramebuffer` struct layout is **unchanged** — no field added or reordered. `effectiveSubdivisionsForHover_` was always `vec2`; only the semantics of `.y` changed (unused → depth rescale). All four `static_assert(offsetof/sizeof...)` guards in `ir_render_types.hpp` still hold verbatim. ✓
- GLSL UBO (`layout(std140, binding = 3)`) and Metal `FrameDataIsoTriangles` struct both declare `float2 effectiveSubdivisionsForHover` at the same offset. Both shaders get identical treatment for `.y`. ✓
- The `<= 0.0` guard in both shaders is **load-bearing**: `system_trixel_to_framebuffer.hpp` line 101 uploads `vec2(subdiv, 0.0f)` for the main-canvas smooth-mode path — the sentinel converts 0.0 → 1.0 (no rescaling), keeping the main canvas byte-identical. ✓
- `kDepthFaceEncodeScale = 4` correctly mirrors `encodeDepthWithFace`'s `d*4 + face` shift. ✓
- `depthScale = effSub / cubeSub` is the correct factor to lift model-frame depth (`× cubeSub × 4`) into shared framebuffer units (`× effSub × 4`). ✓
- `compositeDistanceOffset = worldDepth * effectiveSub_ * 4` is the correct offset scale to match the floor / GRID voxel depth convention. ✓
- Fallback `cubeSub == 0` (pure SDF/text overlay, no voxel raster this frame): `depthScale = 1.0`, `compositeDistanceOffset = worldDepth` — matches pre-#1624 raw-offset behavior for that case. ✓

**Lighting** — N/A. No changes to AO / shadow / flood-fill / fog paths.

**Math / coordinates** — ✅ `pos3DtoDistance`, `roundVec3HalfUp` used via `IRMath` (via `using namespace IRMath`). No bare `glm::` or `std::` math.

**Serialization** — N/A.

**Naming / style** — ✅ `renderedSubdivisions_` trailing `_` (public member). `effectiveSub_` on `System<N>` trailing `_`. `kDepthFaceEncodeScale` uses `k` prefix for constexpr. No `m_` on public, no trailing `_` on private.

**Tests / build** — ✅ PR body has: `canvas_stress` A/B capture (stash → rebuild → restore → rebuild), zoom-1 cardinal diff (0.04% on re-voxelize solids only — fast path unchanged), before/after floor-sort screenshots, macOS Metal build clean. Linux/GL cross-host smoke is an open test plan item (handled via smoke label below). ✓

**Opus-only items flagged:**
- `int(round(float(rawDist) * depthScale))` when `effSub % cubeSub != 0` introduces float quantization. The alternative (no rescale) was provably wrong, so any rounding error is bounded by ½ encoded unit. However, the interaction of this rounding with `normalizeDistance` across the full subdivisions × zoom range warrants an Opus pass to confirm there are no visible depth-sort artifacts at the subdivision cap boundary.

### Nits

- `system_entity_canvas_to_framebuffer.hpp:201` — `constexpr int kDepthFaceEncodeScale = 4; // mirrors encodeDepthWithFace` is a function-local constant that silently drifts if the depth encoding ever changes. A named constant in `ir_render_types.hpp` (e.g. `kDepthEncodeShift = 4`) co-located with `FrameDataTrixelToFramebuffer` would be more discoverable and would bind the C++ side to the same single source of truth the comment names. Optional — the comment is clear enough for now.

### Praise

- Repurposing the "unused" `effectiveSubdivisionsForHover_.y` slot avoids a struct layout change that would break the std140 append story and require reflowing the scatter shader UBO — a clever zero-cost extension.
- The `renderedSubdivisions_ = 0` sentinel for "no voxel raster this frame" and the corresponding `cubeSub < 1` fallback path are clean ways to keep pure-SDF/text overlay canvases on the pre-existing raw-offset path without a new flag.
- Before/after crop pair in the PR body makes the regression visible at a glance.

- [ ] `fleet-build --target IRShapeDebug` + `IRShapeDebug --auto-screenshot 10` on Linux/GL (cross-host smoke, tagged below)
- [ ] Verify `canvas_stress` at zoom ≤ 1 (where `effSub == 1`) is byte-identical before/after to confirm no regression at the base case

Escalation: please Opus-recheck before merge (touches: engine/prefabs/irreden/render/systems/system_entity_canvas_to_framebuffer.hpp, engine/render/src/shaders/ — float quantization in depth rescale at subdivision cap boundary warrants a second pass).

Opus recheck required: float-rounding in `int(round(rawDist * depthScale))` when `effSub % cubeSub != 0` — confirm no visible depth-sort artifacts at the subdivision cap boundary across the full zoom/subdivision range.

jakildev (commented):
## Opus recheck — render: scale world-placed detached-canvas depth to framebuffer units (#1872)

**Verdict: approve** (keeping `fleet:has-nits` for Sonnet's optional constant nit).

Sonnet approved and escalated one Opus-only item: float quantization in
`int(round(rawDist * depthScale))` when `effSub % cubeSub != 0` — "confirm no
visible depth-sort artifacts at the subdivision cap boundary." I traced the
actual depth encoding the composite reads (not just the diff hunks) and the
concern is **resolved** — with one deeper confirmation Sonnet could not make
without reading the encode path.

### The Opus-only pass (what Sonnet could not confirm)

**The linear rescale is valid only because the canvas writes the ×4 encoding,
not the bit-packed one — confirmed.** `ir_iso_common.glsl` has *two* depth
encoders: `encodeDepthWithFace(d,face) = d*4 + face` and the bit-packed
`encodeDepthWithFaceFrac = (depth<<10)|(uFrac4<<6)|(vFrac4<<2)|slot`. A linear
`round(rawDist * depthScale)` would **corrupt** a bit-packed value. I confirmed
the world-placed re-voxelize canvas's `triangleDistances` (what
`f_trixel_to_framebuffer` reads) is written via `encodeDepthWithFace(depthBase,
slot)` at `c_voxel_to_trixel_stage_1.glsl:333` (and `:298`), with
`depthBase = isoDepthAlongAxis(microPositionFixed,…)` and
`microPositionFixed = round(voxelPositionAligned * subdivisions)` — i.e.
`rawDist = modelDepth·cubeSub·4 + slot`. The frac path (`:261,:275`) feeds the
*per-axis* canvases, which composite through `f_peraxis_scatter`, not this
shader. So the linear rescale operates on the ×4 encoding exactly as the
author's comment claims. ✓

**The rounding is bounded and cannot misorder against world geometry —
confirmed.** Expanding the rescale:
`round((modelDepth·cubeSub·4 + slot)·effSub/cubeSub)
= modelDepth·effSub·4 + round(slot·effSub/cubeSub)`. The depth term is an
**exact integer** (no rounding error); only the face/slot sub-band rounds. Max
face bleed is `3·effSub/cubeSub`, and the spacing between adjacent model-depth
cells after scaling is `effSub·4`. Since `3·effSub/cubeSub < 4·effSub` for every
`cubeSub ≥ 1`, the face sub-band **never crosses a depth-cell boundary** — model
depth ordering is exact and face ordering stays monotonic within each cell. The
worst case (a rounding tie between two faces of the *same* voxel at the *same*
model depth) is sub-pixel-depth and visually irrelevant. **No visible depth-sort
artifact at the cap boundary across the zoom/subdivision range.** This is the
confirmation Sonnet asked for.

**Composite scale matches the world convention — confirmed.** Rescaled rawDist
(`modelDepth·effSub·4`) + offset (`worldDepth·effSub·4`) =
`(modelDepth+worldDepth)·effSub·4`, which matches the main world canvas /
floor / GRID convention (`depthBase = Σ microPositionFixed = pos3DtoDistance ·
effSub`, then `·4`). The fix correctly supplies the `× effSub × 4` factor the
#1624 raw-units default was missing. The P4b-1 offset/model-depth decomposition
is additive in iso depth (`pos3DtoDistance = roundHalfUp(dot(pos,(1,1,1)))`),
so the split is exact modulo the entity-origin rounding already captured by
`roundVec3HalfUp(translation)`. ✓

**Fast paths byte-identical — confirmed.** `.y == 0` (main canvas uploads
`vec2(subdiv, 0)`; zero-init for everything else) → `depthScale = 1.0` via the
`<= 0.0` guard in both shaders; `screenLocked_` → scale 1 + offset 0;
`renderedSubdivisions_ == 0` (no voxel raster) → pre-#1624 raw offset. The
hover-carrying main canvas reads only `.x`, so repurposing `.y` is inert there.
No overflow risk beyond what the world canvas already tolerates at the same
`× effSub × 4` scale. ✓

### Agreement with Sonnet

Sonnet's full checklist holds — struct layout unchanged (reused the `vec2.y`
slot, all `static_assert` offset guards intact), GLSL and Metal patched
identically, no ECS/naming/ownership regressions, smoke labels already applied.
Nothing to retract.

### Nits (non-blocking — do NOT need a re-review round)

- Agree with Sonnet's `kDepthFaceEncodeScale = 4` nit: hoisting it to
  `ir_render_types.hpp` next to `FrameDataTrixelToFramebuffer` (the single
  source of truth the comment names) would bind the C++ side to the shader's
  encode shift. Genuinely optional — the local comment is clear.

Approve. Mergeable as-is once cross-host smoke lands.

jakildev:
Nit addressed: moved `kDepthFaceEncodeScale` out of the tick body to `IRRender::kDepthEncodeShift` in `ir_render_types.hpp`, co-located with `FrameDataTrixelToFramebuffer`. Build clean. `fleet:has-nits` → `fleet:changes-made`. — worker

jakildev (commented):
## Opus recheck (re-review) — nit-fix follow-up commit `e412c0ca` (#1872)

**Verdict: approve** (unchanged — re-affirming after the nit-fix push).

My prior Opus approval at 20:01:36Z reviewed commit `2896287d`. A follow-up
commit `e412c0ca` landed at 20:01:47Z addressing the one optional nit Sonnet
and I both flagged (the function-local `kDepthFaceEncodeScale`). Reviewed the
new commit in full:

- Hoists the constant to `ir_render_types.hpp:187` as `kDepthEncodeShift = 4`,
  co-located with `FrameDataTrixelToFramebuffer` and the scatter-UBO
  `static_assert` block — exactly the single-source-of-truth placement the nit
  asked for. ✓
- The constant lands at `namespace IRRender` scope; the consuming system header
  does `using namespace IRRender`, so the unqualified reference at
  `system_entity_canvas_to_framebuffer.hpp:208` resolves. ✓
- The old `kDepthFaceEncodeScale` definition and its sole use are both removed —
  no dangling reference remains (`git grep` clean). ✓
- Pure rename + relocation of a `constexpr int 4`; the depth-rescale math
  (`depthScale = effSub/cubeSub`, `compositeDistanceOffset =
  worldDepth·effSub·4`) is byte-for-byte unchanged. The full correctness
  analysis from my prior review (linear rescale valid on the ×4 encoding,
  rounding bounded below a depth-cell, fast paths byte-identical) carries over
  intact.

Nothing else changed. The depth-rescale fix remains correct and mergeable as-is
once cross-host smoke lands.

PR #1874, opened 2026-06-15
render: sun-shadow darkens only the directional term, never ambient

jakildev:
## Summary

Part of the small-cube **\"black / missing face\"** report. `LIGHTING_TO_TRIXEL` multiplied the sun-shadow factor into the **entire** light, ambient included:

```glsl
faceFactor = mix(sunAmbient, 1.0, lambert) * sunIntensity;
baseRgb    = src.rgb * ao * shadow * faceFactor;   // <- shadow * ambient too
```

So a fully self-shadowed surface (`shadow == 0`) collapsed to **pure black** — its 0.4 ambient floor multiplied away. A re-voxelize cube's −X/−Y side faces land in the cube's own cast shadow at most off-cardinal poses, so they went black and read as missing faces.

### Fix

Ambient / indirect fill is **not** blocked by the directional sun-shadow map — only the Lambert term is. Fold `shadow` into the directional half only:

```glsl
faceFactor = (sunAmbient + (1.0 - sunAmbient) * lambert * shadow) * sunIntensity;
baseRgb    = src.rgb * ao * faceFactor;
```

**Unshadowed pixels (`shadow == 1`) are byte-identical.** Only shadowed pixels change: they keep the ambient floor instead of going black. GLSL + Metal in lockstep.

### Before / after (zoom 4, 30° yaw)

side by side

Left = master (shadowed faces/cast-shadows near black), right = ambient floor preserved. The brightened-only diff shows the change is confined to shadowed regions:

brightened

### Blast radius

This is engine-wide and correct: every cast shadow softens from near-black to ambient-lit (~40%) in every creation. Render-reference baselines for shadowed scenes will shift (intentional).

### Scope note

This fixes the **shadowed-black** faces. A *separate* issue — faces where the gray floor is genuinely visible behind them (geometry actually missing, not just dark) — is being addressed in a follow-up.

## Test plan

- [x] `canvas_stress` A/B (clean master rebuild vs this branch, deterministic `--no-spin` shot): only shadowed pixels change; lit/unshadowed byte-identical.
- [x] `fleet-build --target IRCanvasStress` clean (macOS / Metal); GLSL + Metal edited in lockstep.
- [ ] Linux / OpenGL smoke (cross-host).


jakildev (commented):
## Review — render: sun-shadow darkens only the directional term, never ambient

**Verdict:** approve (pending Opus recheck — see footer)

### Checklist

**ECS invariants** — ✅ Shader-only change. No C++ or ECS code touched.

**Ownership / lifetime** — ✅ No allocations, no lifetime changes.

**Render pipeline:**

- No UBO/SSBO structs changed. `FrameDataVoxelToTrixel` (binding 7), `FrameDataLightingToTrixel` (binding 27), `FrameDataSun` (binding 29) are all unchanged in both GLSL and Metal — no CPU/GPU struct drift risk. ✓
- No binding-point changes. ✓
- No new shader files; existing `.glsl` and `.metal` counterparts edited in lockstep. ✓

**Lighting (sun-shadow fix):**

Math verified:

- Old: `faceFactor = mix(sunAmbient, 1.0, lambert) * sunIntensity` = `(sunAmbient + (1-sunAmbient)*lambert) * sunIntensity`, applied as `src.rgb * ao * shadow * faceFactor`. When `shadow == 0`: `baseRgb == 0` (bug — ambient floor wiped out).
- New: `faceFactor = (sunAmbient + (1-sunAmbient) * lambert * shadow) * sunIntensity`, applied as `src.rgb * ao * faceFactor`. When `shadow == 0`: `baseRgb = src.rgb * ao * sunAmbient * sunIntensity` — ambient preserved. ✓
- When `shadow == 1`: expanding both formulas gives the same result — byte-identical as claimed. ✓
- The LUT path likewise drops the standalone `* shadow` and uses the same folded `faceFactor` — consistent. ✓

**GLSL / Metal parity:** The two formulas are algebraically identical (difference is only uniform access: GLSL uses `sunAmbient` / `sunIntensity` from the `ir_sun_shadow_sample.glsl` include; Metal uses `sunFrameData.sunAmbient` / `sunFrameData.sunIntensity` from the struct argument). ✓

**Debug overlays (modes 2 and 3):** Both still reference the raw `shadow` variable directly for the false-color display — correct; the debug views show the shadow factor independently, unaffected by how it feeds into the lighting equation. ✓

**`worldReceive` path:** When an opt-in world-placed detached solid recomputes its shadow via `worldSunShadowFactor(...)`, the result is still a scalar in `[0, 1]`. The new formula handles it correctly — directional term attenuated, ambient preserved. ✓

**Lighting checklist items 1–4:**
- Grid coverage / culling: no change to voxel iteration or occlusion grid. ✓
- Shadow-ring extent: no change to shadow-map sampling or cascade selection. ✓
- Light-seed expansion: no change to flood-fill or light-volume sampling. ✓
- AO guard band: no change to AO sampling radius or neighbor reads. ✓

**Math / coordinates** — ✅ No coordinate transforms changed.

**Serialization** — N/A.

**Naming / style:** Shader comment style and variable names consistent with the existing file. ✓

**Tests / build:**
- macOS / Metal: `canvas_stress` A/B run (lit pixels byte-identical, shadowed pixels brightened to ambient floor). ✓
- Linux / OpenGL: pending (tagged below).

### Praise

- Minimal, targeted fix: 3 lines of changed shader arithmetic, no architectural churn.
- The math is provably correct: `shadow == 1` case stays byte-identical, `shadow == 0` preserves the ambient floor — the PR demonstrates both in the test plan.
- Screenshots (before/after + brightened-diff) give reviewers and the human everything needed for a visual sanity check.
- GLSL and Metal updated in lockstep in a single commit, with the commit message repeating the derivation for the record.

- [ ] `fleet-build --target IRShapeDebug` + smoke on Linux/GL (tagged below — cross-host needed for shader change authored on macOS)
- [ ] Confirm cast-shadow floor regions show ambient fill (not near-black) on GL path

Escalation: please Opus-recheck before merge (touches: engine/render/src/shaders/ — core per-pixel lighting path, engine-wide blast radius).

Opus recheck required: engine/render/ touch — core lighting-application shader with engine-wide effect; confirm the ambient-preserve formula is correct under all shadow-sample edge cases (worldReceive cascade, detached canvas forced shadow=1, HDR tonemapped path).

jakildev (commented):
## Opus recheck — render: sun-shadow darkens only the directional term, never ambient

**Verdict:** approve

Sonnet did the algebraic pass and approved, escalating for confirmation of the folded ambient-preserve formula under the edge cases it could not trace from the hunk alone (worldReceive cascade, detached canvas forced `shadow=1`, HDR tonemap). I read both full shader bodies on the PR head (`6e7d210c`) and confirm all three, plus the one thing that actually matters for a fix like this: **`shadow` is never re-applied downstream.**

### What Sonnet covered (confirmed)

- `shadow == 1` → byte-identical to master (both formulas expand to the same expression). ✓
- `shadow == 0` → ambient floor `src.rgb * ao * sunAmbient * sunIntensity` preserved instead of collapsing to black. ✓
- GLSL ↔ Metal algebraic parity, edited in lockstep. ✓

### Opus-only items — what I checked beyond Sonnet

**No downstream re-multiplication of `shadow`.** This is the load-bearing check for moving `shadow` out of the global `faceFactor`. After `baseRgb` is computed, the remaining uses of `shadow` in both shaders are:
- Debug overlay modes 2/3 only — false-color views, never feed the lit output (`c_lighting_to_trixel.glsl:189,193`; `.metal:145,148`). ✓
- **Nothing else.** The light-volume bleed (`baseRgb += src.rgb * light`) and the HDR sky term (`+= skyColor * skyIntensity * skyFactor * ao`) were both *already* additive-unshadowed in master, so removing `shadow` from `baseRgb` does not double-count or drop it anywhere. The fix is purely the intended directional-only attenuation. ✓

**Partial shadow (`0 < shadow < 1`)** — not just the two endpoints Sonnet checked. Old: `shadow` scaled *both* ambient and directional. New: scales *only* directional. This is a genuine (intended) semantic change in the penumbra, not just the fully-shadowed case — and it is the correct behavior (ambient / indirect fill is not occluded by the sun-shadow map). ✓

**worldReceive cascade** — `worldSunShadowFactor()` returns a scalar in `[0,1]` folded into `faceFactor` identically; the separate worldReceive light-volume sample path is untouched. ✓

**Detached canvas forced `shadow=1`** — `c_lighting_to_trixel.metal:136` forces `shadow = 1.0` for a default (non-world-receive) detached canvas, so those pixels stay byte-identical to the old `mix()` path. GLSL force-path equivalent confirmed. ✓

**HDR tonemap** — `ACESFilm` is monotonic; previously-black self-shadowed pixels now map to a small non-zero value. No new clipping or inversion — strictly the intended brightening. ✓

### Praise

Minimal (3 lines of arithmetic per backend), provably correct, and the inline comment captures *why* (the re-voxelize -X/-Y self-shadowed face reading as a missing black face) so the next reader won't "simplify" the `* shadow` back into `faceFactor`. Screenshots make the visual delta obvious.

### Note for the human

Cross-host smoke is still outstanding (`fleet:needs-linux-smoke` / `fleet:needs-windows-smoke`) — correctly tagged, not a review blocker. The math is backend-agnostic and the GL/Metal sources are in lockstep, so I expect the Linux/GL smoke to be a formality, but a shader authored on Metal should still get its GL pass before merge.

PR #1841, opened 2026-06-15
engine/update: PERIODIC_IDLE_POSITION_OFFSET → PARALLEL_FOR

jakildev:
## Summary
- Converts `System<PERIODIC_IDLE_POSITION_OFFSET>` from old `createSystem<>` lambda form to `registerSystem<>` with `kConcurrency = PARALLEL_FOR` and `kGrainSize = 512`
- Drops `EntityId` from the tick — uses `kNullEntity` as a stable per-system source key for the modifier slot, avoiding `IRSystem::ParallelSafe` which hits a gap in `PartitionExcludes`/`MakeMemberTickFn` (T-328 sub-task D)
- Updates `engine/prefabs/irreden/update/CLAUDE.md` to reflect the new slot key

## Test plan
- [x] Builds clean (`fleet-build --target IRPerfGrid`)
- [x] `validateAllPipelineGroups` passes (system already a singleton group)
- [x] Profiled on IRPerfGrid voxel_set zoom8 (262144 entities, 240 frames): before 15.994ms → after 13.240ms (−17% update time)
- [ ] Reviewer: confirm bit-identical output on a voxel_set run

## Notes for reviewer
`kNullEntity` (= 0) is safe as the source key because: (a) each entity has its own `C_Modifiers` so the slot `(0, TRANSFORM_TRANSLATION, ADD)` is unique within each entity's mods, and (b) `removeBySource` is called with the dying entity's own id, not `kNullEntity`, so the slot is cleaned up by component destruction rather than source sweep.

Closes #1802

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1802, opened 2026-06-13
engine/update: PERIODIC_IDLE_POSITION_OFFSET → PARALLEL_FOR (cheap win, #1052)


jakildev:
Cheap-win carve-off of #1052 (CPU UPDATE-pipeline parallelization). Plan: `.fleet/plans/issue-<this>.md`.

**Area:** engine/update
**Model:** sonnet
**Blocked by:** (none)

## Scope
`System<PERIODIC_IDLE_POSITION_OFFSET>` (`engine/prefabs/irreden/update/systems/system_periodic_idle_position_offset.hpp:24`) is currently SERIAL. Its tick calls `IRPrefab::Modifier::upsertBySourceInPlace` which (verified, `engine/prefabs/irreden/common/modifier.hpp:623-638`) mutates **only the entity's own `C_Modifiers`** — no shared/global state. So it is directly PARALLEL_FOR-safe.

Add `static constexpr Concurrency kConcurrency = Concurrency::PARALLEL_FOR;` to the `System<>` specialization (optionally `static constexpr int kGrainSize = 512;`). It already runs in its own singleton pipeline group in perf_grid, satisfying the "PARALLEL_FOR must be a singleton group" rule.

## Acceptance criteria
- System tagged PARALLEL_FOR; `fleet-build --target IRPerfGrid` clean; boot validator (`validateAllPipelineGroups`) passes (it's already a singleton group).
- Output **bit-identical** to before (per-entity modifier writes are independent).
- Record the `update` stage ms before/after on IRPerfGrid `voxel_set` zoom8 (`--auto-profile 240`).

## Context
Lowest-risk piece of the #1052 update-parallelization. Note: update (14ms zoom8) is **secondary** to render (voxelStage1 32ms) for FPS — this is a free floor-lowering win, not a frame-gate fix.

## Gotchas
- Confirm the per-entity tick form populates `prepareRangedTick` (it does — only the per-archetype batch form is rejected for PARALLEL_FOR).

jakildev:
Scope check: this issue's scope appears to have landed in "docs: plan #1052 update-parallelization carve-offs (#1802/#1803/#1804)" (#1805, merged 2026-06-13). Skipping ingest; please close this issue if the scope is fully addressed. — fleet-queue-ingest

jakildev:
[...]

PR #1891, opened 2026-06-17
render(voxel): UPDATE_VOXEL_SET_CHILDREN → PARALLEL_FOR (#1803)

### LINKED ISSUE #1803: engine/voxel: UPDATE_VOXEL_SET_CHILDREN → PARALLEL_FOR (thread-safe queuePositionRange) (#1052)
Carve-off of #1052 (CPU UPDATE-pipeline parallelization). **Parked — lower priority** until the render gate (#1294 chain) lands; update (14ms) is secondary to render (voxelStage1 32ms) for FPS. Plan: `.fleet/plans/issue-<this>.md`.

**Area:** engine/voxel
**Model:** opus
**Blocked by:** (none)

## Scope
`UPDATE_VOXEL_SET_CHILDREN` (`engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp`) runs a SERIAL 262K-entity loop — one of the two dominant UPDATE costs (#1740). Make it `Concurrency::PARALLEL_FOR` by making its shared-pool writes thread-safe.

## Shared-state hazards (verified)
- **`pool.queuePositionRange(startIdx, count)`** (`component_voxel_pool.hpp:523-531`) — appends to `m_pendingPositionRanges` via `emplace_back`; **races under concurrent ticks**. This is the real blocker.
- `markChunkWorldBoundsDirty()` (`:373`) — idempotent bool write, benign.
- `setEntityIdForRange(...)` (`:234`) — `std::fill` on **disjoint** per-entity voxel spans (safe by allocation invariant) + idempotent `m_entityIdsDirty` bool.

## Approach
Per-worker `thread_local std::vector<std::pair<size_t,size_t>>` pending-range accumulator; after the parallelFor barrier, a main-thread merge appends all worker queues into `m_pendingPositionRanges`. Mirror the established `thread_local` pattern in `system_voxel_to_trixel.hpp`. Confirm the disjoint-span invariant for `setEntityIdForRange` before relying on it. Then tag the system PARALLEL_FOR (singleton group). Aligns with epic **#226** Phase 1 (per-worker deferred state).

## Acceptance criteria
- Output **bit-identical** cull-on path; `m_pendingPositionRanges` contains the same set of ranges (order-independent) as serial.
- Measured `update` ms reduction on IRPerfGrid `voxel_set` zoom8.
- `validateAllPipelineGroups` passes; builds clean both hosts.

## Gotchas
- Uses the per-entity-id tick form today (`:49`) — convertible to the ranged form PARALLEL_FOR needs.
- No unified per-worker-merge harness exists yet; roll the thread_local pattern (don't block on a #226 primitive).

jakildev:
WIP for #1803 — parallelize `UPDATE_VOXEL_SET_CHILDREN`'s 262K-entity position loop (one of the two dominant UPDATE costs, #1740).

**Status: design-blocked.** Investigation surfaced two architecture-shaped blockers the plan (`​.fleet/plans/issue-1803.md`) did not anticipate; they need an architect call before the `Concurrency::PARALLEL_FOR` tag can land. See the `## NEEDS-DESIGN` comment below. This branch carries only an in-context analysis comment on the system header; the system is functionally unchanged (still SERIAL).

Closes #1803 once the design is resolved and implemented.

### DIFF
--- a/engine/prefabs/irreden/voxel/CLAUDE.md
+++ b/engine/prefabs/irreden/voxel/CLAUDE.md
@@ -74,7 +74,11 @@ for single voxels and particles.
   skipped via `C_VoxelPool::isRangeVisible` against the previous frame's
   shadow-feeder-expanded cull viewport — same cull gate as `REBUILD_GRID_VOXELS`
   (#1288). A set entirely outside that viewport pays nothing; a visible set in
-  a static scene re-uploads its positions every frame.
+  a static scene re-uploads its positions every frame. Runs `PARALLEL_FOR`
+  (#1803, `ParallelSafe`): disjoint per-set span writes, the canvas→pool
+  lookup pre-resolved in `beginTick`, and the `queuePositionRange` hazard
+  deferred into a per-worker accumulator merged in `endTick` — see
+  `engine/system/CLAUDE.md` "Concurrency policy".
 - `REBUILD_GRID_VOXELS` (UPDATE pipeline, T-294; inverse re-voxelize #1720)
   — Epic C C6. Runs AFTER `UPDATE_VOXEL_SET_CHILDREN`. Re-rasterizes
   GRID-mode entities (entities carrying `C_RotationMode::GRID`, the
--- a/engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp
+++ b/engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp
@@ -152,27 +152,23 @@ template <> struct System<UPDATE_VOXEL_SET_CHILDREN> {
             pool.getPositions(),
             pool.getPositionOffsets()
         );
-        if (writtenCount > 0) {
-            // CPU world positions moved in place; evict the chunk world-AABB
-            // cache so the continuous-yaw cull re-derives this set's bounds and
-            // stays a conservative superset of the live voxels (#1439). The
-            // dirty-bool write is an idempotent same-value store (benign under
-            // concurrency).
-            pool.markChunkWorldBoundsDirty();
-        }
         // A GPU-transform-indirected set (#1396) has binding 5 written by the
         // UPDATE_VOXEL_POSITIONS_GPU prepass each frame. We still recompute its
         // CPU global mirror above (a sane translation-only fallback for the
         // STAGE_1 canvas-switch re-seed, and for cull/picking), but we must NOT
         // queue it for the steady-state binding-5 flush — that flush runs after
         // the prepass in the RENDER pipeline and would clobber the GPU positions.
-        if (writtenCount > 0 && voxelSet.gpuTransformSlot_ == IRRender::kVoxelTransformStatic) {
-            // Defer the queuePositionRange emplace_back — it races across sets
-            // that share a pool. The merge into pool.m_pendingPositionRanges
-            // happens serially in endTick.
-            pendingByWorker_[static_cast<std::size_t>(IRJob::workerId())].push_back(
-                PendingRange{&pool, voxelSet.voxelStartIdx_, static_cast<std::size_t>(writtenCount)}
-            );
+        // Both cases (static + GPU-slotted) defer to endTick via pendingByWorker_:
+        // count_ > 0 means "queue + dirty"; count_ == 0 means "dirty only". This
+        // keeps markChunkWorldBoundsDirty() on the main thread and off the hot
+        // concurrent path (eliminates the same-value concurrent store TSan would flag).
+        if (writtenCount > 0) {
+            const bool isStatic = voxelSet.gpuTransformSlot_ == IRRender::kVoxelTransformStatic;
+            pendingByWorker_[static_cast<std::size_t>(IRJob::workerId())].push_back(PendingRange{
+                &pool,
+                isStatic ? voxelSet.voxelStartIdx_ : std::size_t{0},
+                isStatic ? static_cast<std::size_t>(writtenCount) : std::size_t{0}
+            });
         }
         if (voxelSet.ownerEntityId_ == IREntity::kNullEntity && entityId != IREntity::kNullEntity &&
             voxelSet.numVoxels_ > 0) {
@@ -188,12 +184,16 @@ template <> struct System<UPDATE_VOXEL_SET_CHILDREN> {
     }
 
     void endTick() {
-        // Main-thread merge of the deferred queuePositionRange calls staged by
-        // every worker this frame. Order-independent downstream (the flusher
-        // treats m_pendingPositionRanges as a set of contiguous-run uploads).
+        // Main-thread merge. For each staged entry: queue the position range when
+        // count_ > 0 (static voxel set), and always evict the chunk world-AABB
+        // cache (markChunkWorldBoundsDirty). Dirty marking runs here, not in tick,
+        // to keep the bool write off the concurrent path (#1803 TSan note).
         for (std::vector<PendingRange> &worker : pendingByWorker_) {
             for (const PendingRange &range : worker) {
-                range.pool_->queuePositionRange(range.startIdx_, range.count_);
+                if (range.count_ > 0) {
+                    range.pool_->queuePositionRange(range.startIdx_, range.count_);
+                }
+                range.pool_->markChunkWorldBoundsDirty();
             }
         }
     }
--- a/engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp
+++ b/engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp
@@ -113,6 +113,31 @@ template <> struct System<UPDATE_VOXEL_SET_CHILDREN> {
         }
     }
 
+    // PARALLELIZATION (#1803) — DESIGN-BLOCKED. The 262K-entity position
+    // loop above is one of the two dominant UPDATE costs (#1740) and the
+    // plan (.fleet/plans/issue-1803.md) calls for `Concurrency::PARALLEL_FOR`.
+    // Two architecture-shaped blockers the plan's hazard list missed must be
+    // resolved before the tag can land (see the PR's ## NEEDS-DESIGN comment):
+    //
+    //   1. EntityId ownership registration vs the PARALLEL_FOR validator.
+    //      This tick uses the per-entity-id form to do the one-time owner
+    //      registration (`pool.setEntityIdForRange` → GPU picking buffer +
+    //      the `ownerEntityId_` guard). `ir_system.hpp` makes
+    //      `PARALLEL_FOR + usesEntityId_ + !parallelSafe_` FATAL. The write
+    //      is genuinely thread-safe (disjoint span fill), so `ParallelSafe`
+    //      is the sanctioned tag — but `ParallelSafe` is currently unwired:
+    //      `PartitionExcludes` (archetype) and `MakeMemberTickFn` (tick
+    //      lambda) strip only `Exclude<...>`, not tag types. The latent
+    //      wiring needs completing (reuse `detail::FilterTags_t`) OR the
+    //      registration must be relocated out of the parallel tick.
+    //   2. Pool resolution runs `getComponent<C_VoxelPool>(canvas)` on the
+    //      worker (foreign-entity lookup) — not worker-safe per
+    //      `engine/system/CLAUDE.md`. Plus `lastCanvas_`/`lastPool_` race as
+    //      System<N> members under PARALLEL_FOR. Resolution must move to a
+    //      main-thread `beginTick` canvas→pool pre-resolve.
+    //
+    // The thread_local pending-range merge the plan describes is the easy
+    // part and lands once (1) and (2) are settled.
     static SystemId create() {
         return registerSystem<UPDATE_VOXEL_SET_CHILDREN, C_VoxelSetNew, C_WorldTransform>(
             "UpdateVoxelSetChildren"
--- a/engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp
+++ b/engine/prefabs/irreden/voxel/systems/system_update_voxel_set_children.hpp
@@ -5,21 +5,59 @@
 #include <irreden/voxel/components/component_voxel_pool.hpp>
 #include <irreden/common/components/component_world_transform.hpp>
 #include <irreden/common/components/component_player.hpp>
+#include <irreden/ir_job.hpp>
 #include <irreden/ir_render.hpp>
+#include <irreden/render/active_canvas.hpp>
 #include <irreden/render/cull_viewport_state.hpp>
 #include <irreden/render/sun_shadow_constants.hpp>
+
+#include <cstddef>
+#include <unordered_map>
+#include <vector>
 
 using namespace IRComponents;
 using namespace IRMath;
 
 namespace IRSystem {
 template <> struct System<UPDATE_VOXEL_SET_CHILDREN> {
-    // Per-tick scratch: last-resolved canvas → pool pointer. The pool is
-    // fetched fresh each tick (not across frames) because a between-frame
-    // canvas archetype migration invalidates the pointer; within a single
-    // tick the archetype is stable so amortizing the lookup is safe.
-    IREntity::EntityId lastCanvas_ = IREntity::kNullEntity;
-    C_VoxelPool *lastPool_ = nullptr;
+    // #1803 — this 262K-entity position loop is one of the two dominant
+    // UPDATE costs (#1740), so it runs Concurrency::PARALLEL_FOR. The tick
+    // keeps the per-entity-id form for the one-time owner registration
+    // (`setEntityIdForRange` → GPU picking buffer), which makes the
+    // registration validator demand the `IRSystem::ParallelSafe` opt-in. The
+    // body is audited thread-safe; the two architecture blockers the plan
+    // missed (worker-side pool lookup, shared `queuePositionRange` vector)
+    // are resolved by `beginTick` pre-resolution + a per-worker deferred
+    // merge below. See the PR #1891 architect direction (1a + 2a + 3).
+    static constexpr Concurrency kConcurrency = Concurrency::PARALLEL_FOR;
+
+    // One deferred `queuePositionRange` call. The pool's
+    // `m_pendingPositionRanges.emplace_back` is the lone cross-set shared-
+    // vector write in the tick, so workers stage into their own slot and the
+    // main thread merges in `endTick` (the pending set is order-independent
+    // downstream). Keyed by pool because the canvas→pool map spans every
+    // canvas's pool, and different sets in one tick may target different pools.
+    struct PendingRange {
+        C_VoxelPool *pool_ = nullptr;
+        std::size_t startIdx_ = 0;
+        std::size_t count_ = 0;
+    };
+
+    // Resolved on the main thread in beginTick, read-only in the worker tick.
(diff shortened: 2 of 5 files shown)


jakildev:
## NEEDS-DESIGN

The plan (`.fleet/plans/issue-1803.md`) frames this as "convert the tick to the ranged form and tag `PARALLEL_FOR`," and its verified-hazard list covers `queuePositionRange` / `markChunkWorldBoundsDirty` / `setEntityIdForRange`. Those are the easy parts. Two architecture-shaped blockers the plan did **not** anticipate stand in the way of a clean tag — both need an architect call.

### Blocker 1 — EntityId ownership registration vs the `PARALLEL_FOR` validator

The tick uses the **per-entity-id** form because it does a one-time owner registration:

```cpp
if (voxelSet.ownerEntityId_ == kNullEntity && entityId != kNullEntity && voxelSet.numVoxels_ > 0) {
    voxelSet.ownerEntityId_ = entityId;
    pool.setEntityIdForRange(voxelSet.voxelStartIdx_, numVoxels_, entityId); // → GPU picking entity-id buffer
}
```

`ir_system.hpp` makes `PARALLEL_FOR + usesEntityId_ + !parallelSafe_` a registration-time FATAL. The write itself is genuinely thread-safe (disjoint per-set span fill + the set's own field), so `IRSystem::ParallelSafe` is the sanctioned "I audited it" tag — **but `ParallelSafe` is unwired**: `PartitionExcludes` (the archetype/dispatch path in `createSystem`) and `MakeMemberTickFn` (the `registerSystem` tick-lambda builder) strip only `Exclude<...>`, not tag types. A tag in the pack leaks into `getArchetype<...>` (breaks the match) and into the member-tick signature (static_assert). Confirmed: **no system anywhere in engine+creations passes any of `ParallelSafe`/`Spawns`/`Destroys`/`MainThread`/`AlsoReads`/`AlsoWrites` in a registration pack** — the wiring has never been exercised.

The id is **not** used for a manager lookup, so this is not the "kNullEntity sentinel" case (PR #1841); the real entity id must reach `setEntityIdForRange`.

Options:
- **(1a) Complete the latent `ParallelSafe` wiring** — filter tag types out of the included list in both `createSystem` (archetype/dispatch) and `registerSystem` (`MakeMemberTickFn`), reusing the existing tested `detail::FilterTags_t`. Provably a **no-op for every existing tag-free / `Exclude`-only system** (`FilterTags_t<Pack...>` == `PartitionExcludes<Pack...>::Included` whenever the pack has no tag types). Then tag this system `ParallelSafe`. Smallest functional surface, but it touches the core registration templates used by every system → wants an architect sign-off + full-suite run.
- **(1b) Relocate the one-time ownership registration to a dedicated SERIAL system** (per-entity-id), parallelize only the position update. `UPDATE_VOXEL_SET_CHILDREN` is registered in **20+ creations**, so a new system means 20+ pipeline-registration edits.
- **(1c) Set `ownerEntityId_` + the pool entity-id span at voxel-set spawn time.** Many creation sites + the headless/staged `pendingVoxels_` path complicate it.

### Blocker 2 — pool resolution runs on worker threads

The tick resolves the pool with `getComponent<C_VoxelPool>(canvas)` (a **foreign-entity** lookup on the canvas entity). `engine/system/CLAUDE.md` is explicit: PARALLEL_FOR keeps workers off `EntityManager`'s hash map precisely because it's not safe under concurrent reads — so this lookup cannot run in the parallel tick. The `lastCanvas_` / `lastPool_` `System<N>` member cache (amortizing that lookup today) also races under PARALLEL_FOR.

Options:
- **(2a)** `beginTick` (main thread) pre-resolves a `canvas → C_VoxelPool*` map (active canvas + each distinct explicit `canvasEntity_`); the parallel tick reads the map (read-only concurrent). Handles multi-canvas; adds a per-entity map lookup.
- **(2b)** If the contract is "one active world canvas per UPDATE tick for pooled sets," `beginTick` resolves just the active pool and asserts single-canvas. Simpler, less general. (Detached/private-pool sets still flow through this system's CPU global mirror even though `REBUILD_DETACHED_VOXELS` owns their binding-5 — needs confirming whether they reach here with a distinct pool.)

### Blocker 3 (non-controversial) — the plan's piece

thread_local pending-position-range accumulator + post-barrier `endTick` merge into `pool.m_pendingPositionRanges` (downstream is order-independent). `markChunkWorldBoundsDirty` (idempotent bool) and `setEntityIdForRange` (disjoint fill) confirmed benign under concurrency. This lands once 1+2 are settled.

### Recommendation

**1a + 2a + 3.** 1a is the architecturally-intended path (the tag exists for exactly this audited-safe case) and is the smallest functional change, but because it touches the shared registration templates I'm escalating rather than shipping it under a plan that didn't scope it. Architect: please confirm 1a vs 1b/1c, and 2a vs 2b.

— worker (opus)

jakildev:
## Architect direction — greenlight 1a + 2a + 3

Your three-blocker analysis is correct. Direction:

**Blocker 1 → 1a: complete the latent `ParallelSafe` wiring.** This is the architecturally-intended path — the tag exists for exactly this audited-safe disjoint-span-write case, and the registration filter (`detail::FilterTags_t`) already exists and is tested. 1a is the smallest functional change and is provably a no-op for every existing tag-free / `Exclude`-only system (`FilterTags_t<Pack...>` == `PartitionExcludes<Pack...>::Included` when the pack has no tag types). Reject 1b (20+ pipeline edits) and 1c (spawn-time complexity).

**HARD GATES on 1a** (it touches the core registration templates used by every system):
- Filter tag types out of the included list in BOTH `createSystem` (archetype/dispatch) and `registerSystem` (`MakeMemberTickFn`), reusing the existing `FilterTags_t` — do not hand-roll a second filter.
- Demonstrate the no-op: a compile-time `static_assert` (or test) showing `FilterTags_t` == current behavior for a tag-free pack.
- **FULL test suite green** (every system registers/dispatches unchanged) + a careful reviewer pass specifically on the template change. This is the make-or-break.

**Blocker 2 → 2a:** `beginTick` (main thread) pre-resolves a `canvas → C_VoxelPool*` map (active canvas + each distinct explicit `canvasEntity_`); the parallel tick reads it concurrently read-only. Reject 2b's single-canvas assertion — too narrow given detached / private-pool sets.

**Blocker 3:** as you scoped — `thread_local` pending-position-range accumulator + post-barrier `endTick` merge into `pool.m_pendingPositionRanges`. Lands once 1a/2a are in.

**Acceptance:** full suite green; `UPDATE_VOXEL_SET_CHILDREN` runs under PARALLEL_FOR with no picking-id / pool regression; the `ParallelSafe` no-op proof committed; **perf measured** (the point of the change). Task pointer `.fleet/plans/issue-1803.md` will carry this.

jakildev:
Design-unblocked resolved — implemented the architect's 1a + 2a + 3 direction.

**1a (ParallelSafe wiring):** `createSystem`/`registerSystem` now strip tag types from the included pack via `detail::FilterTags_t` (was `Exclude<...>`-only). Provable no-op for every tag-free pack (`FilterTags_t == PartitionExcludes::Included`) — static_assert + test proof in `system_access_test.cpp`. `UPDATE_VOXEL_SET_CHILDREN` is the first `ParallelSafe`-tagged system.

**2a (pool pre-resolve):** `beginTick` builds a `canvas → C_VoxelPool*` map (+ active-pool fast path) on the main thread; the worker tick reads it instead of the worker-unsafe `getComponent<C_VoxelPool>`.

**3 (deferred merge):** `queuePositionRange` staged into a per-worker accumulator (by `IRJob::workerId()`), merged on the main thread in `endTick`. Disjoint span writes + idempotent dirty bools are benign.

**Acceptance verified:**
- Full test suite green (1173 pass / 1 skip); FilterTags no-op proof committed.
- **Byte-identical render:** 28/28 `shape_debug` shots identical SERIAL vs PARALLEL_FOR (no picking-id/pool regression).
- **Perf measured:** IRPerfGrid `voxel_set` (262K entities, 10 workers) UPDATE-pipeline CPU ~5.3-5.9ms → ~3.8-4.2ms.

Ready for review. — worker (opus)

jakildev (commented):
## Review — render(voxel): UPDATE_VOXEL_SET_CHILDREN → PARALLEL_FOR (#1803)

**Verdict:** approve

The title carries a stale "[WIP / design-blocked]" tag — the implementation is complete and ready for merge. Four files of substance plus tests.

### ECS invariants

- `canvasToPool_` and `activePool_` / `activeCanvas_` are populated on the main thread in `beginTick` and only read (never written) during worker ticks — concurrent-read of `std::unordered_map` is safe under C++11 when no concurrent writes occur. ✓
- `pendingByWorker_[workerId()]` — each slot is owned by exactly one thread; `workerId()` confirmed as 0 (main), 1..workerCount() (workers), so `workerCount() + 1` slot count is correct. ✓
- `endTick` drains `pendingByWorker_` and calls `queuePositionRange` serially on the main thread — correct; the pool's `m_pendingPositionRanges` vector is not touched by any worker. ✓
- `forEachComponent<C_VoxelPool>` in `beginTick` runs on the main thread; EntityManager is not touched from workers. ✓
- `PARALLEL_FOR` defers structural changes, so the pool pointers cached in `canvasToPool_` remain valid for the full beginTick → worker ticks → endTick span. ✓
- `endTick` is auto-wired by `HasEndTickMember` concept in `ir_system.hpp:214` — no manual wiring needed, confirmed present before this PR. ✓

### Ownership / lifetime

- `C_VoxelPool *pool_` in `PendingRange` and `C_VoxelPool *activePool_` / map values are raw non-owning pointers (correct per engine convention). ✓
- No cross-frame pointer capture — `canvasToPool_` is rebuilt each `beginTick`. ✓

### Render pipeline

No shader or CPU/GPU struct changes. ✓

### `FilterTags_t` + `ParallelSafe` wiring

- `kIsTag` includes `IsExclude<T>::value` (system_access.hpp:161), so `FilterTags_t` strips both `Exclude<...>` AND semantic tags — the no-op proof in `system_access_test.cpp` is correct. ✓
- In `createSystem`: `Partition::Excluded` (for the exclude archetype) still uses `PartitionExcludes` which does NOT see `ParallelSafe` (it's not an `Exclude<...>`), so the exclude archetype is unaffected. ✓
- In `registerSystem`: `IncludedList = FilterTags_t<Components...>` feeds only non-tag types to `MakeMemberTickFn`, preventing `tick(..., ParallelSafe&)` from forming. The full `Components...` pack still reaches `createSystem<Components...>` so `parallelSafe_` in the access descriptor is correctly set. ✓
- `deriveAccessFromSignature<FunctionTick, TickComponents...>()` still sees the full `TickComponents...` pack (tags included) and uses its own `FilterTags_t` pass internally — coherent. ✓

### Tests

`system_access_test.cpp` proves the no-op invariant at compile time and provides a runtime mirror for CI. Coverage is thorough for the cases that exist today. ✓

### Math / naming / style

No IRMath violations. Member naming follows convention (`pool_`, `startIdx_`, `count_` are public, trailing `_`; `kConcurrency` is `static constexpr` with `k` prefix). ✓

### Nits

- `system_update_voxel_set_children.hpp` — PR title suffix "[WIP / design-blocked]" in the GitHub title is stale; should be cleaned up before merge.
- `test/system/system_access_test.cpp` — the no-op proof covers the tag-free pack and the Exclude-only pack separately, and the divergence case for a tag-only pack. The combined `Exclude<...> + ParallelSafe` case (e.g. `FilterTags_t<C_AccessA, Exclude<C_AccessC>, ParallelSafe, C_AccessB>`) is not asserted. Both `kIsTag<Exclude<...>>` and `kIsTag<ParallelSafe>` are true, so the implementation handles it correctly, but a compile-time assertion would close the gap:
  ```cpp
  static_assert(std::is_same_v<
      Filtered<C_AccessA, Exclude<C_AccessC>, ParallelSafe, C_AccessB>,
      IRSystem::detail::TypeList<C_AccessA, C_AccessB>>,
      "FilterTags_t must strip both Exclude and tag types in a combined pack");
  ```

### Praise

- The architect analysis comment (#1891 NEEDS-DESIGN comment) is exceptionally clear — both blockers identified with options, tradeoffs, and a concrete recommendation. This is the quality of analysis that prevents wasted iteration.
- Using `detail::FilterTags_t` (existing, tested infrastructure) rather than hand-rolling a second filter is architecturally correct and keeps the proof surface minimal.
- `PendingRange` per-worker accumulator is the right shape: clears without shrinking (steady-state allocation-free), indexed by `workerId()`, merged serially in `endTick` where ordering is irrelevant.
- `beginTick` canvas→pool pre-resolve cleanly keeps workers off `EntityManager` — follows the existing `system_propagate_transform.hpp` pattern at line 166.

- [x] Full test suite green (1173 pass / 1 skip) — confirmed in commit message
- [x] 28/28 `shape_debug` shots byte-identical SERIAL vs PARALLEL_FOR — confirmed
- [x] IRPerfGrid `voxel_set` perf measured: ~5.3-5.9ms → ~3.8-4.2ms — confirmed
- [ ] Clean up stale "[WIP / design-blocked]" from PR title

Opus recheck required: first system to exercise the `ParallelSafe` registration path; changes to `ir_system.hpp` core templates (`createSystem` / `registerSystem`) affect every system registration in the engine. Correctness analysis is thorough and tests are solid, but the template boundary deserves an Opus pass before this lands.

jakildev:
Addressed feedback: removed stale '[WIP / design-blocked]' tag from PR title. No code changes. — worker

jakildev (commented):
## Opus recheck — #1891: render(voxel): UPDATE_VOXEL_SET_CHILDREN → PARALLEL_FOR (#1803)

Picked this up on the explicit `fleet:needs-opus-recheck` escalation. Standalone PR (base = `master`, MERGEABLE), no stack gate. Sonnet's correctness pass is thorough and I agree with its verdict; I spent this pass on the two things a first pass can't fully prove — the **core registration-template boundary** (`createSystem` / `registerSystem` now feed `FilterTags_t` where they fed `PartitionExcludes::Included`) and the **concurrency contract** of the new PARALLEL_FOR body. Both hold up.

### The template no-op — confirmed, and proven beyond the static_asserts

Sonnet flagged the `FilterTags_t` swap as the make-or-break and walked the `kIsTag` / `applyComponent` coherence; on closer read I confirm it is a genuine no-op for **every** existing system, for a stronger reason than the test asserts:

- `FilterTags_t<Pack>` differs from `PartitionExcludes<Pack>::Included` **only** when the pack carries one of the six tag types (`Exclude` is stripped identically by both; the divergence set is `Spawns / Destroys / MainThread / ParallelSafe / AlsoReads / AlsoWrites`).
- I grepped every `registerSystem<…>` / `createSystem<…>` call in `engine/` + `creations/`: **no existing site passes any of those tags in a registration pack.** The only other `ParallelSafe` mention in the tree is the comment in `system_periodic_idle_position_offset.hpp:30` documenting the very gap this PR closes (#1802). So the swap cannot alter any current system — it can only *enable* the tag path that previously failed to wire.
- This is also structurally guaranteed: before this PR, a tag in `registerSystem`'s `IncludedList` would have driven `MakeMemberTickFn` to form a `tick(…, ParallelSafe&)` the spec doesn't have → build failure. So a tagged registration could not have compiled, which is why none exist. ✓
- `registerSystem` keeps the access derivation correct: `IncludedList = FilterTags_t<…>` feeds **only** `MakeMemberTickFn` (line 342), while the full `Components...` pack still flows to `createSystem<Components...>` (line 356) → `deriveAccessFromSignature` sees the tag and flips `parallelSafe_`. The validator at `ir_system.hpp:51` (PARALLEL_FOR + usesEntityId_ + !parallelSafe_ → FATAL) is therefore satisfied by the tag, not bypassed. ✓

### Concurrency contract — confirmed

- **Per-worker accumulator indexing is exactly in bounds.** `IRJob::workerId()` returns `0` (main) / `1..workerCount()` (workers) per `ir_job.hpp:114`; `beginTick` sizes `workerCount() + 1` slots, so `pendingByWorker_[workerId()]` never overruns, including the no-job-pool case (`workerCount()==0` → 1 slot → `workerId()==0`). ✓
- **Pool pointers stay valid for the whole begin→tick→endTick span.** PARALLEL_FOR defers structural changes and the tick body issues none (no `deferredCreate`/`setComponent`), so the `C_VoxelPool*` cached in `canvasToPool_` / `activePool_` during `beginTick` cannot be invalidated mid-tick. `canvasToPool_.find(...)` from workers is a const lookup on a map that is never written during the parallel phase — safe concurrent read. ✓
- **Disjoint writes hold.** Each set owns `[voxelStartIdx_, +numVoxels_)` in its pool, so `updateAsChild` and `setEntityIdForRange` never overlap across sets sharing a pool; the one genuinely-shared `queuePositionRange` emplace is correctly deferred to the serial `endTick` merge (order-independent downstream). ✓
- **Behavior parity at the active-canvas path.** The `getActiveCanvasEntity()` → `getActiveCanvasEntityOrNull()` swap is strictly safer (headless → `nullptr` → early return instead of asserting), and the 28/28 byte-identical `shape_debug` shots confirm the rendered case is unchanged. ✓

### Nits (non-blocking)

- **`system_update_voxel_set_children.hpp:94` — `markChunkWorldBoundsDirty()` is a concurrent same-value write to a shared bool.** Multiple sets on the active canvas split across workers all call `pool.markChunkWorldBoundsDirty()` on the *same* pool, which is `m_chunkWorldBoundsDirty = true` — a non-atomic concurrent store. It is benign on every real architecture (single aligned store of a constant; the result is unconditionally `true`) and the architect already classified it benign in the Blocker-3 analysis, so this is not a merge gate. But it *is* a textbook data race that ThreadSanitizer would flag if the engine ever runs a TSan build. Cheapest clean follow-up: hoist a `bool anyWritten` per-worker (or reuse the `pendingByWorker_` non-empty signal) and call `markChunkWorldBoundsDirty()` once in `endTick`, mirroring the `queuePositionRange` deferral you already do for the same reason.
- **Stale PR title.** "[WIP / design-blocked]" in the GitHub title is no longer accurate — clean it up before merge (Sonnet also noted this).
- **Combined-tag static_assert gap (Sonnet's nit).** I confirm Sonnet's read: `kIsTag` ORs `IsExclude` with the tag predicates and `FilterTagsImpl` drops each head independently, so a mixed `Exclude<…> + ParallelSafe` pack is already handled correctly — the missing assertion is a coverage gap, not a bug. Worth adding the one-line `static_assert` Sonnet sketched.

### Where I agree with Sonnet

Everything in Sonnet's pass holds up: the `endTick` auto-wiring via `HasEndTickMember`, the raw-non-owning pointer discipline, the per-worker accumulator shape, and the naming/style. No disagreements.

Cross-host smoke tagging skipped — no shader / render-backend changes (CPU-side concurrency only; render output is byte-identical), so this is not a smoke-relevant render path.

jakildev:
Addressed both Opus nits:

- **Stale PR title**: removed "[WIP / design-blocked]" suffix from GitHub title.
- **Combined-tag `static_assert`**: added `Filtered<C_AccessA, Exclude<C_AccessC>, ParallelSafe, C_AccessB>` → `TypeList<C_AccessA, C_AccessB>` static_assert to `system_access_test.cpp` (closes the coverage gap Sonnet flagged and Opus confirmed).
- **`markChunkWorldBoundsDirty()` hoist**: moved from concurrent tick body to serial `endTick` merge. Static sets: existing `PendingRange` entry triggers both `queuePositionRange` (count_ > 0) + dirty mark. GPU-slotted sets: push a sentinel entry (count_ == 0) that triggers only the dirty mark. Eliminates the TSan-visible concurrent same-value store while keeping identical observable behavior. All 17 `SystemAccess` tests pass; `IRShapeDebug` build clean.

— worker

PR #1833, opened 2026-06-14
render: GUI-test shot tables with scripted input (P2, #1795)

jakildev:
## Summary
- Add \`GuiInputEvent\`, \`GuiTestShot\`, \`GuiTestConfig\` to \`auto_screenshot.hpp\`; include \`ir_input_types.hpp\` for \`KeyMouseButtons\`
- Implement \`createGuiTestSystem\` in \`auto_screenshot.cpp\`: same shot-cycling state machine as \`createAutoScreenshotSystem\` plus a frame-scheduled event-injection phase (move/press/release/scroll via P1 \`inject*\` API)
- Settle begins after the highest \`frameOffset_\` in each shot so the GUI widget reflects the interaction before capture; shots with no events behave identically to \`AutoScreenshotShot\`
- Wire up in \`voxel_editor/main.cpp\` with a demo table including \`editor_palette_click\` (move+press+release) alongside the existing stable-render framings

## Stack context
This PR is part of a stack. Reviewers: review this PR on its own;
the chain is coordinated in the PR body's "Stacked on" line.

Stacked on: https://github.com/jakildev/IrredenEngine/pull/1827
Full chain: #1827, #1795

## Test plan
- [x] Build \`IRVoxelEditor\` clean (macOS/Metal verified)
- [x] Run \`IRVoxelEditor --auto-screenshot 10\`; confirmed 4 shots captured: editor_idle, editor_palette_click, editor_zoom_out, editor_zoom_in
- [x] editor_palette_click shot differs from editor_idle by ~244K bytes (input injection visibly affects the widget); shot labels log correctly per-shot
- [x] Pre-existing Metal non-determinism in voxel editor (139K bytes idle-to-idle): confirmed pre-existing in the original \`createAutoScreenshotSystem\` path — NOT introduced by this PR. \`IRShapeDebug\` is byte-stable. Byte-stable GL verification + render-verify manifest integration is scope for later phases (#1796, #1797).
- [x] \`IRShapeDebug\` still builds clean with updated \`auto_screenshot.hpp\`

Closes #1795

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1827, opened 2026-06-14
input: synthetic mouse/click injection for headless GUI tests (#1794)

jakildev:
## Summary
- New `IRInput` synthetic-input API — P1 of the GUI/mouse verification harness (#1793): `beginSyntheticInput` / `injectMouseMove` / `injectButton` / `injectScroll` / `isSyntheticInputActive`. This is the missing primitive the rest of the harness depends on (no way to move the cursor or press a button headless today).
- When active, `InputManager::tick` drains injected button/scroll queues instead of the GLFW window queues, and `advanceInputState` snapshots an injected cursor (screen px). The pending cursor latches once per `tick()` so every pipeline-event snapshot in a frame reads the same position — mirrors GLFW's per-frame cursor. Run-scoped flip, modeled on `IRVideo::isAutoCaptureActive()`.
- The GLFW path is gated on the flag and stays **byte-identical** when synthetic input is inactive. `injectButton` shares `applyButtonEvent` (extracted from `processKeyMouseButtons`) and `injectScroll` reuses `processScrolls`, so injected and GLFW paths produce identical state. Inject calls assert if `beginSyntheticInput()` was not called.

## Test plan
- [x] Builds clean (`IRShapeDebug`).
- [x] Functional smoke (temporary `ui_widgets` injection, logged, then reverted — not committed): `injectMouseMove(400,300)` → `getMousePosition`==(400,300); `injectButton(left, PRESSED)` → `checkKeyMouseButton(left, PRESSED)`==true the next frame; injected RELEASE observed the following frame; cursor persists across frames (correct level semantics).
- [x] GLFW/inactive path unregressed: `IRShapeDebug --auto-screenshot` runs clean (exit 0); inactive path is byte-identical by construction (gated on the flag; `applyButtonEvent` is a behavior-preserving extraction).
- [ ] Linux/OpenGL smoke (host-unvalidated; macOS-authored).
- [ ] Windows smoke (host-unvalidated).

## Notes / scope
- Widget-level click assertions (`wasClicked` against a GUI hitbox) require inverting the screen→framebuffer→GUI-trixel transform and are **Phase 2/3** of the harness — out of P1 scope. P1 proves the primitive drives the exact `IRInput` snapshot getters that `HITBOX_MOUSE_TEST_GUI` / `WIDGET_INPUT` consume.
[...]

Issue #1795, opened 2026-06-13
render: GUI-test shot tables with scripted input (P2)


jakildev:
**Model:** sonnet
**Part of epic:** #1793
**Plan file:** `~/.fleet/plans/issue-1793.md` (full epic plan)
**Blocked by:** #1794

## Scope
Extend the headless shot-table mechanism so a shot can carry a scripted sequence
of mouse/click input events, fired (via P1's injection) on deterministic frames
before capture.

## Approach
- Add `GuiInputEvent { int frameOffset_; enum {MOVE,PRESS,RELEASE,SCROLL} type_;
  ivec2 screenPx_; vec2 scroll_; }` and `GuiTestShot { AutoScreenshotShot render_;
  const GuiInputEvent* inputs_; int numInputs_; }` alongside `AutoScreenshotShot`
  in `engine/video/`.
- In the auto-screenshot system, fire each shot's input events on the scheduled
  frames (relative to the shot's settle) via P1's `IRInput::inject*`, let the GUI
  settle, then capture. Reuse the existing settle-frame + screenshot plumbing.
- Frame-scheduled, never wall-clock — same determinism contract as the existing
  capture path.

## Acceptance criteria
- A `GuiTestShot[]` table drives a headless run: cursor moves to a scripted
  point, a click fires, the frame is captured deterministically.
- Demonstrated on the voxel editor (e.g. move to a palette swatch, click,
  capture) — reproducible across runs.

## Dependencies
Blocked by P1 (#1794, the injection API). Blocks P3.

## Gotchas
- Keep `GuiTestShot` a superset of `AutoScreenshotShot` so existing
  camera/zoom/crop config carries over unchanged.
- Settle frames must elapse after the last input event before capture so the GUI
  reflects the interaction.

## References
`docs/design/gui-mouse-harness.md` §"Phase 2"; `engine/video/include/irreden/video/auto_screenshot.hpp`; `engine/video/src/auto_screenshot.cpp`.

PR #1892, opened 2026-06-17
engine/update: PROPAGATE_TRANSFORM intra-node row-range parallelism

### LINKED ISSUE #1804: engine: PROPAGATE_TRANSFORM intra-node row-range parallelism (#1052)
Carve-off of #1052 (CPU UPDATE-pipeline parallelization). **Parked — lower priority** until the render gate (#1294 chain) lands; update is secondary to render for FPS. Plan: `.fleet/plans/issue-<this>.md`.

**Area:** engine/update
**Model:** opus
**Blocked by:** (none)

## Scope
`PROPAGATE_TRANSFORM` (`engine/prefabs/irreden/update/systems/system_propagate_transform.hpp`) already uses `IRJob::parallelFor`, but parallelizes **across archetype nodes** (`kGrainSize=1`, one task per archetype at each depth level — `:75`, `:169`). **Confirmed:** when all 262K entities live in ONE archetype node (IRPerfGrid), the level holds n=1 node → one task → the 262K `composeNode` runs **serially** on one worker. It's the other dominant UPDATE cost (#1740). Add intra-node row-range splitting.

## Approach
Refactor `composeRange` (`:145-155`) to take `(ArchetypeNode* node, int rowBegin, int rowEnd)` and `parallelFor(0, node->length_, grainSize, ...)` over **rows within** each node, instead of over archetype indices. Preserve the per-level barrier (parent transforms of level d must be composed before level d+1 reads them — `:130-141`). `composeNode` (`:278-355`) is **race-free for disjoint row ranges**: each entity writes only its own `C_WorldTransform[i]` from its own `C_LocalTransform[i]` + a read-only already-composed parent. Keep the existing inter-node parallelism for multi-archetype workloads (hybrid: parallelize over rows when a node is large, over nodes when many small).

## Acceptance criteria
- Output **bit-identical** `C_WorldTransform` to the serial path on both single-archetype (IRPerfGrid) and multi-archetype (parented hierarchy) workloads.
- Measured `update` ms reduction on IRPerfGrid `voxel_set` zoom8.
- `validateAllPipelineGroups` passes; builds clean both hosts.

## Gotchas
- The per-level barrier is load-bearing — don't let a child level's rows compose before the parent level finished. Keep level iteration serial; only parallelize rows within a level's nodes.
- No shared-state hazard (column writes are disjoint by row); no thread_local needed here — just row chunking. Aligns with epic **#226** Phase 1.

jakildev:
## Summary
- `PROPAGATE_TRANSFORM` parallelized **across archetype nodes** (T-378), so a single dominant archetype degenerated to serial: the 262K-entity IRPerfGrid `voxel_set` grid lives in one depth-0 node → `n=1` → one task → 262K rows composed on one worker.
- This PR adds **intra-node row-range splitting**. Each per-depth level is flattened into a row-chunk list: a small archetype stays whole (one chunk — preserves the prior per-node granularity for many-small-node levels), a node larger than the chunk size splits into row-range chunks. A single `parallelFor` + the existing per-level barrier still drive each depth in order.
- `composeNode` → `composeNodeRows(node, parentWorld, rowBegin, rowEnd, …)`. Per-row math unchanged; disjoint ranges are race-free (each entity writes only its own `C_WorldTransform[i]`), so output is **bit-identical** to the serial path regardless of chunk boundaries.
- Hybrid trigger: parallelize when a level has ≥ `kMinForParallel` (8) nodes **or** ≥ `kMinRowsForParallel` (4096) rows; chunk size `max(kMinChunkRows=2048, ceil(totalRows/targetTasks))`, `targetTasks = workers×2`. A single large node fans out to ~2 chunks/worker; many small nodes are grained to ~targetTasks tasks.

## Performance (IRPerfGrid `voxel_set`, grid-size 64 = 262,144 entities, macOS/Metal, 10 workers)
Method: `fleet-run IRPerfGrid --auto-profile 12 --auto-screenshot` (the `--auto-screenshot` fixed-step makes the profiler's last-frame `update` sample land on a ticked frame). `update` = whole UPDATE-pipeline stage; only `PROPAGATE_TRANSFORM` changed, so the delta is its contribution.

| Configuration | `update` / frame (3 runs) |
|---|---|
| `origin/master` (per-archetype → serial single node) | 4.32 / 4.42 / 4.61 ms |
| This PR (intra-node row-chunks) | 2.25 / 2.27 / 2.56 ms |

≈ **47% reduction (~2.1 ms)** on the UPDATE pipeline. Render still dominates frame time, so this is a secondary-axis win per the issue's framing — it closes the single-node degeneracy directly.

## Test plan
- [x] `fleet-build --target IRShapeDebug` / `IRPerfGrid` clean (both hosts: macOS here; Linux pending smoke)
- [x] **Bit-identity:** `IRShapeDebug --auto-screenshot` 28/28 byte-identical to `origin/master` (deterministic path reading `C_WorldTransform`)
- [x] **Bit-identity (parallel path):** IRPerfGrid screenshots are non-deterministic run-to-run (GPU rasterization speckle), so byte-compare is invalid; instead confirmed per-shot pixel-diff magnitude mine-vs-master sits within the demo's own noise floor (mine-vs-mine ≈ master-vs-master ≈ mine-vs-master; MAE < 0.5/255, < 1% pixels, no structural region difference)
- [x] `validateAllPipelineGroups` passes (runs every frame; 250-frame + screenshot runs clean)

Docs: updated `docs/perf-reports/threading_propagate_transform.md` (this implements the report's own "per-level cost estimator" follow-up) and `engine/prefabs/irreden/update/CLAUDE.md`.

Closes #1804

### DIFF
--- a/docs/perf-reports/threading_propagate_transform.md
+++ b/docs/perf-reports/threading_propagate_transform.md
@@ -39,21 +39,39 @@ transform from a pre-resolved vector instead of an inline
 
 ## Dispatch tuning
 
-`parallelFor` is called per-level with a heuristic grain size:
+> **Updated by #1804 (intra-node row-range parallelism).** The original
+> T-378 dispatch was per-archetype only — one parallelFor task per
+> archetype node. That degenerates when a single archetype holds the
+> whole scene (IRPerfGrid's 262K-entity node is one node at depth 0):
+> the level has `n = 1`, falls below `kMinForParallel`, and the 262K-row
+> `composeNode` runs serially on one worker. See the dedicated section
+> below for the row-splitting fix and its measurement.
 
-- **`kMinForParallel = 8`** — levels with fewer archetypes than this
-  fall back to a serial walk. The dispatch + wait overhead for a
-  parallelFor with only 1–7 tasks exceeds the savings.
-- **Grain size = `max(kGrainSize, (n + targetTasks - 1) / targetTasks)`**
-  where `targetTasks = max(1, workerCount * 2)`. The "2 tasks per
-  worker" target gives enkiTS room to load-balance without flooding
-  the queue with one-archetype tasks (each archetype already carries
-  hundreds-to-thousands of entities through the inner loop).
+Each level is flattened into a list of row-chunks and dispatched with a
+single `parallelFor`:
+
+- **`kMinForParallel = 8`** (node count) **OR `kMinRowsForParallel = 4096`**
+  (total entity rows) — a level parallelizes when it has at least this
+  many archetypes *or* this many total rows. Below both, the dispatch +
+  wait overhead exceeds the savings and the level composes serially.
+- **`kMinChunkRows = 2048`** — the minimum rows per chunk. A node smaller
+  than the chunk size stays whole (one chunk), preserving the original
+  per-archetype granularity for many-small-node levels; a node larger
+  than the chunk size splits into multiple row-range chunks. The chunk
+  size is `max(kMinChunkRows, ceil(totalRows / targetTasks))` with
+  `targetTasks = max(1, workerCount * 2)`, so a single dominant node
+  splits into ~`targetTasks` chunks (≈2 per worker).
+- **Grain = `max(1, ceil(numChunks / targetTasks))`** over the chunk
+  list — `1` when chunks already number ≤ `targetTasks` (single large
+  node), grouping when many small nodes produce more chunks than tasks.
 - **`g_jobManager == nullptr`** — unit-test and pre-`World` paths
-  short-circuit to the same serial walk as `kMinForParallel`.
+  short-circuit to the serial walk.
 
-The `kGrainSize = 1` member exists as the floor; the smart grain
-computation is the effective grain at runtime.
+`composeNode` became `composeNodeRows(node, parentWorld, rowBegin, rowEnd, …)`
+— same SQT body, now bounded to a row range. Disjoint row ranges are
+race-free (each entity writes only its own `C_WorldTransform[i]`), so
+output is bit-identical to the serial path regardless of chunk
+boundaries.
 
 ## Measurement
 
@@ -99,6 +117,40 @@ This PR keeps the structural correctness and cache invalidation work
 gated behind unit tests so the matrix run only needs to confirm the
 perf delta.
 
+## Intra-node row-range parallelism — measurement (#1804)
+
+The 262K scaling regime the T-378 report flagged but didn't exercise
+is exactly the single-node degeneracy #1804 fixes. With the whole grid
+in one depth-0 archetype, the per-archetype dispatch left the entire
+composition on one worker; row-splitting fans it across all workers.
+
+Method: `fleet-run IRPerfGrid --auto-profile 12 --auto-screenshot` on
+macOS (Metal, Apple Silicon, 10 enkiTS workers). `--auto-screenshot`
+enables fixed-step (one UPDATE tick per render frame) so the profiler's
+last-frame `update` sample reliably lands on a ticked frame; without it
+the fixed-step UPDATE pipeline may not tick the sampled render frame and
+reports `update: 0.000`. grid-size 64 (262,144 entities, one archetype).
+Three runs each; `update` is the whole UPDATE-pipeline stage, and since
+only `PROPAGATE_TRANSFORM` changed, the delta is its contribution.
+
+| Configuration | `update` stage / frame |
+|---|---|
+| `origin/master` (per-archetype dispatch → serial single node) | 4.32 / 4.42 / 4.61 ms |
+| This PR (intra-node row-chunks across 10 workers) | 2.25 / 2.27 / 2.56 ms |
+
+≈ **47% reduction (~2.1 ms)** on the UPDATE pipeline. Render still
+dominates frame time (~8 ms), so this is a secondary-axis win per the
+issue's own framing, but it closes the single-node degeneracy directly.
+
+Bit-identity check: `IRShapeDebug --auto-screenshot` is 28/28
+byte-identical to `origin/master` (deterministic path reading
+`C_WorldTransform`). IRPerfGrid screenshots are non-deterministic
+run-to-run (GPU rasterization speckle), so they can't be byte-compared;
+instead, per-shot pixel-diff magnitude (mine-vs-master) was confirmed to
+sit within the demo's own run-to-run noise floor (mine-vs-mine ≈
+master-vs-master ≈ mine-vs-master, MAE < 0.5/255, < 1% pixels touched,
+no structural region difference).
+
 ## Correctness coverage
 
 `test/ecs/propagate_transform_test.cpp` (13 tests, all green):
@@ -130,12 +182,14 @@ Full suite: 964 tests in IrredenEngineTest pass; IRShapeDebug
   is the authoritative source for the ≥2× target. Defer until the
   perf-CI hardware-fingerprinted baseline (#1100) is online so the
   comparison is hardware-stable.
-- **`kMinForParallel` and grain-size auto-tuning.** The current
-  `kMinForParallel = 8` and `workers × 2` task target are calibrated
-  for IRPerfGrid's shallow hierarchy. If future scenes exhibit deep
-  hierarchies with very small per-level archetype counts, a per-level
-  cost estimator (sum of `node->length_`) might choose better than
-  the archetype-count threshold alone.
+- **`kMinForParallel` and grain-size auto-tuning.** ✅ Addressed by
+  #1804: the per-level cost estimator (sum of `node->length_` →
+  `totalRows`) is now a parallelization trigger alongside the
+  archetype-count threshold, and large single nodes split across
+  workers by row range. Remaining tuning headroom: `kMinChunkRows`
+  (2048) and `kMinRowsForParallel` (4096) are calibrated for the
+  per-row quat/vec3 cost; revisit if a much heavier or lighter
+  per-entity composition lands.
 - **Foreign-component read declaration.** The system reads parent
   `C_WorldTransform` via a foreign-entity lookup, which the
   `SystemAccess` trait does not see from the `tick`/`beginTick`
--- a/engine/prefabs/irreden/update/CLAUDE.md
+++ b/engine/prefabs/irreden/update/CLAUDE.md
@@ -68,11 +68,17 @@ in the `UPDATE` pipeline unless explicitly noted.
   any consumer (render, gizmo, physics) that reads
   `C_WorldTransform`. T-378 partitions the archetype set into
   per-depth levels (cached across frames; the partition rebuilds only
-  on archetype-graph changes); within each level the per-archetype
-  composition fans out to IRJob workers via `IRJob::parallelFor`. Cost
-  is O(N) total compose work with O(passes × archetypes) topology
-  bookkeeping, where passes is bounded by the deepest parent chain.
-  See `docs/perf-reports/threading_propagate_transform.md`.
+  on archetype-graph changes); within each level composition fans out
+  to IRJob workers via `IRJob::parallelFor`. #1804 made the per-level
+  dispatch split by **row range**, not just by archetype: a level is
+  flattened into row-chunks so a single dominant archetype (e.g. a
+  262K-entity grid in one node) splits across workers instead of
+  composing serially on one. Output is bit-identical to the serial
+  path (disjoint rows, each entity writes only its own
+  `C_WorldTransform[i]`). Cost is O(N) total compose work with
+  O(passes × archetypes) topology bookkeeping, where passes is bounded
+  by the deepest parent chain. See
+  `docs/perf-reports/threading_propagate_transform.md`.
 
 ## Commands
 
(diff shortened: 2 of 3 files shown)


jakildev (commented):
## Review — engine/update: PROPAGATE_TRANSFORM intra-node row-range parallelism

**Verdict:** approve

### Checklist

**ECS invariants** — clean.
- `tick()` is empty; all work is in `beginTick()`. No `getComponent` calls on iterating entities, no per-entity allocations.
- `chunks_` is a member vector; `chunks_.clear()` + `push_back` in `beginTick` reuses allocated capacity across frames (high-water-mark pattern). ✓
- `composeChunkRange` lambda captures `levelNodes`, `levelParentWorlds`, and `chunks_` by reference; the lambda is only invoked by `IRJob::parallelFor` which is a blocking call, so captured references remain valid for the lambda's lifetime. ✓

**Race-freedom** — sound.
- Row ranges `[rowBegin, rowEnd)` per chunk are disjoint within a node and trivially disjoint across nodes. Each entity writes only its own `C_WorldTransform[i]`. C++ (since C++11) permits concurrent writes to different elements of the same vector. ✓
- `chunks_` is built on the main thread before `parallelFor` and is read-only on workers during execution. ✓
- `levelParentWorlds[chunk.nodeIndex]` lookups are read-only on workers (written by the main thread during the parent-world resolve phase before the `parallelFor` call). ✓

**Edge cases**
- Zero-entity nodes contribute no chunks (inner `for (rowBegin < len)` doesn't execute when `len == 0`). `parallelFor(0, 0, ...)` is a no-op. ✓
- Serial levels (`!runParallel`) use `continue` before reaching the `chunks_` rebuild, so stale chunk content from a prior level is never read in the serial path. ✓

**Math / coordinates** — `IRMath::max`, `IRMath::min` throughout. `std::int64_t totalRows` correctly prevents overflow when accumulating 262K+ rows into the chunk-size formula. ✓

**Naming / style** — `kMinRowsForParallel`, `kMinChunkRows`, `RowChunk`, `chunks_` (public member, trailing `_`) all comply with conventions. ✓

**Documentation** — perf-report update includes methodology detail (why `--auto-screenshot` is needed, what the noise floor comparison establishes), bit-identity caveat for IRPerfGrid, and the prior follow-up item from T-378 explicitly closed. CLAUDE.md updated to describe row-range splitting. ✓

### Escalation point for Opus

`composeNodeRows` (line 356) is a static member called from workers. When a large node splits into multiple chunks, concurrent tasks all call:

```cpp
auto &locals = IREntity::getComponentData<IRComponents::C_LocalTransform>(node);
auto &worlds = IREntity::getComponentData<IRComponents::C_WorldTransform>(node);
```

for the **same `node` pointer** simultaneously. Pre-PR, each node was processed by exactly one task; this is a new concurrent-same-node access pattern. If `getComponentData` is a pure const lookup returning a reference to stable data (a common pattern), it is safe. But if its internals carry any non-atomic mutable state (lazy init, lookup caching, reference counting), multiple workers racing on the same node pointer would be a data race.

Sonnet cannot verify this without reading `getComponentData`'s implementation inside `IREntity`. Opus: please check that `getComponentData<T>(ArchetypeNode*)` is safe to call concurrently from multiple worker threads for the same node.

### Praise

- Hybrid trigger (`n >= kMinForParallel || totalRows >= kMinRowsForParallel`) is the right shape: the inter-node path (many small archetypes) and the intra-node path (one giant archetype) each parallelise under their natural condition, and the two thresholds are independently tunable without coupling.
- "Small node stays whole (one chunk)" preserves the T-378 per-archetype granularity for multi-archetype levels while adding the large-node fix. The many-small-node case isn't regressed.
- `composeNodeRows` comment (line 352–355) explicitly justifies the redundant per-chunk column fetches as intentionally cheap relative to the row loop — exactly the kind of non-obvious reasoning that belongs in a comment.
- Bit-identity methodology is clear: deterministic `IRShapeDebug` for byte-compare, noise-floor comparison for non-deterministic `IRPerfGrid`. Both verified.

- [x] `IRShapeDebug --auto-screenshot` 28/28 byte-identical (confirmed in PR body)
- [x] `validateAllPipelineGroups` passes (confirmed in PR body)
- [ ] Linux build smoke pending (`fleet:needs-linux-smoke` will be set)
- [ ] `IRPerfGrid` noise-floor comparison on Linux to confirm the ~47% UPDATE gain isn't Metal-specific

Escalation: please Opus-recheck before merge (touches: concurrent `getComponentData` same-node access pattern — new in this PR, not present in prior T-378 per-node dispatch).

Opus recheck required: concurrent `getComponentData<T>(node)` for the same node from multiple workers is new in this PR; verify the function is safe for concurrent calls on the same node pointer.

jakildev (commented):
## Opus recheck — render: GUI-test shot tables with scripted input (P2, #1795) (#1833)

**Verdict:** approve

Escalation consumed: Sonnet flagged this for Opus recheck because it adds new public types (`GuiInputEvent`, `GuiTestShot`, `GuiTestConfig`) and a factory (`createGuiTestSystem`) to the `engine/video/` public surface, and because the second pass was a post-rebase no-drift check after #1827 squash-merged. I focused on the items Sonnet could not fully confirm: ECS endTick invariant, synthetic-input lifecycle, state-machine correctness, public-API layering, and rebase drift.

### What I confirmed beyond Sonnet's pass

- **endTick-on-zero-entities invariant (load-bearing).** Sonnet asserted the P1 pattern; on closer read I confirm it. `createGuiTestSystem` is a faithful superset of `createAutoScreenshotSystem` in the *same file* — identical `C_*Anchor` empty-component + no-op per-entity tick + `endTick` closure idiom, same `g_autoCaptureActive` flag, same `requestScreenshot` / `requestScreenshotWithCrops` capture branch. Since `createAutoScreenshotSystem` shipped in P1 and runs in production, the zero-match-endTick guarantee this relies on is proven, not just claimed.

- **State-machine frame accounting — traced both branches, correct.**
  - Pure-render shot (`numInputs_ = 0`): `maxOffset = -1` → `captureAt = settleFrames_` (3). Frames 0,1,2 settle; capture on frame 3. Matches the original auto-screenshot settle window exactly.
  - Scripted shot (`kPaletteClickEvents`, offsets 0/1/2): `maxOffset = 2` → `captureAt = 6`. MOVE@0, PRESS@1, RELEASE@2 dispatch in the event phase; settle frames 3,4,5; capture on 6 — i.e. a full `settleFrames_` window *after* the last input, which is the correct semantics for "let the GUI reflect the click before capturing." ✓
  - `numShots_ == 0` degenerates to `closeWindow()` after warmup — safe.

- **Synthetic-input lifecycle.** `beginSyntheticInput()` + `g_autoCaptureActive = true` fire at *creation* (documented in the header), not first tick — consistent with `createAutoScreenshotSystem`. No symmetric teardown, but the system always ends in `closeWindow()` → process exit, so there's nothing to leak. `button_` defaults to `kNullButton` and is only consumed by PRESS/RELEASE, both of which set `kMouseButtonLeft` in the table. MOVE/SCROLL never read it. ✓

- **Pointer lifetime.** `kPaletteClickEvents` / `kGuiTestShots` are `constexpr` file-scope arrays (static storage); `GuiTestConfig` stores raw pointers into them. Valid for process lifetime, and the header doc states the requirement explicitly. ✓

- **Public-API layering.** The new `#include <irreden/input/ir_input_types.hpp>` in the public header pulls `IRInput::KeyMouseButtons` into `engine/video`'s public surface. This is a legitimate, minimal coupling — a scripted-GUI-test harness genuinely needs an input-button type, and `engine/video` already links against `engine/input`. Not a layering violation.

- **Rebase drift — none.** Sonnet's re-review claimed zero content delta after the `rebase --onto origin/master` dropped #1827's inherited commits. I confirm: the diff is exactly the three P2 files. The extra hunks in `main.cpp` (the two `makeColorSwatch` call-site wraps and the `getComponent` arg wrap) are pure clang-format line-wrapping with no semantic effect, absorbed cleanly by the rebase. `input_manager.cpp` is absent from the diff (byte-identical to master). ✓

### Nits (carried forward from Sonnet — both genuinely optional, not merge-gating)

- `engine/video/include/irreden/video/auto_screenshot.hpp:114` — the `// GUI-test shot tables (P2, #1795)` section heading embeds task-tracking context that will age out once the issue closes. A timeless heading would read better, but the doxygen on the structs below is self-sufficient.
- `engine/video/src/auto_screenshot.cpp:~195` — `maxOffset` is rescanned per frame for the whole shot duration. O(numInputs) with numInputs=3 is noise; caching it once when `shotFrame_` leaves -1 would tidy it. Cosmetic.

One forward-looking observation, not for this PR: `createGuiTestSystem` and `createAutoScreenshotSystem` now duplicate the camera-apply + cull + crops-aware-capture block. If P3+ adds a third capture variant, folding the shared tail into a helper would be worth it. For P2's two-variant scope, keeping them separate keeps the simpler pure-render path readable — no action needed here.

### Pre-merge note

macOS/Metal build + `IRVoxelEditor --auto-screenshot 10` (4 shots, palette_click diverges) confirmed by author/worker. `engine/video/` is off the render-smoke path, so no cross-host smoke label — but a Linux/OpenGL `fleet-build --target IRVoxelEditor` sanity check before merge is still worth a few seconds since this is the first build of the new types on that toolchain.

Clean approve. The rebase introduced no semantic drift and the new public API is correct and well-documented.

jakildev (commented):
## Opus recheck — engine/update: PROPAGATE_TRANSFORM intra-node row-range parallelism (#1892)

**Verdict:** approve

Sonnet approved and escalated one concern it could not resolve without reading engine-core internals: *"concurrent `getComponentData<T>(node)` for the same node from multiple workers is new in this PR; verify the function is safe for concurrent calls on the same node pointer."* I traced the full call chain and confirm the pattern is **safe**. (Note: the second comment on this PR's timeline titled *"Opus recheck — render: GUI-test shot tables (P2, #1795)"* is a mis-posted review for a different PR — it does not address #1892's concern, so this is the first genuine Opus pass on the row-parallelism change.)

### The concurrency concern — resolved (safe)

When one large node splits into multiple `RowChunk`s, several workers call `composeNodeRows(sameNode, …)` simultaneously. Each call reaches the column vectors through:

```cpp
auto &locals = IREntity::getComponentData<C_LocalTransform>(node);
auto &worlds = IREntity::getComponentData<C_WorldTransform>(node);
// + conditionally C_ResolvedFields / C_Modifiers, gated on node->type_.contains(...)
```

Tracing `getComponentData<T>(ArchetypeNode*)` (`entity_manager.hpp:381`):

- It reads `node->type_` (immutable after node construction), looks up `node->components_[componentType]`, casts, and returns a reference to the already-allocated `dataVector`.
- `node->components_` (the `unordered_map<ComponentId, smart_ComponentData>`) and each `IComponentDataImpl<T>::dataVector` are populated **at node construction** (`archetype_node.cpp`) and are structurally immutable for the duration of the tick — no lazy init, no insertion, no rehash, no refcount, no caching on the read path.
- Workers then write only their disjoint `[rowBegin, rowEnd)` slice of `worlds[i]`. Concurrent writes to distinct indices of one `std::vector` are well-defined. ✓
- `getParentEntityFromArchetype(node->type_)` and `node->type_.contains(...)` are pure reads over the immutable archetype set. ✓

**Lazy-registration cannot fire on a worker.** `getComponentData` → `getComponentType<T>()` (`entity_manager.hpp:226`) *does* lazily `registerComponent<T>()` and mutate `m_pureComponentTypes` on first-ever access of a type — that would be a race if it happened concurrently. It can't here: for a node to be in the level partition at all it must already hold `C_LocalTransform` + `C_WorldTransform`, and the optional `C_ResolvedFields`/`C_Modifiers` fetches are gated on `node->type_.contains(...)`. A component can only be *in* an archetype because an entity carrying it was spawned earlier — which registered the type on the main thread long before any tick. So every `getComponentType<T>()` inside the parallel region hits the already-registered fast path. Confirmed against the four component types this system touches.

This was also the load-bearing question for whether the "bit-identical to serial" claim holds, and it does: disjoint row ranges + immutable per-node storage ⇒ output is independent of chunk boundaries. Sonnet's `IRShapeDebug` 28/28 byte-identical check is consistent with this.

### One forward-looking note (not a merge gate, and outside this PR's diff)

`getComponentData` reaches its data through `unordered_map::operator[]` on both `m_pureComponentTypes` (line 396) and `node->components_` (line 394). With the keys guaranteed already present, `operator[]` performs no insertion and is a pure read on every mainstream STL (libstdc++ on the Linux fleet toolchain, libc++ on macOS, MSVC) — so there is no real-world data race. A standards-pedantic reading of `[container.requirements.dataraces]` does not bless `operator[]` as a concurrent-safe accessor the way it does `find`/`at`, so an engine-core hardening pass could switch these two reads to `.at()`/`.find()`. This is pre-existing engine code (`getComponentData` is unchanged by this PR) and the same-map concurrency it newly exercises is benign on every real target — flagging only so a future allocator/ECS-core sweep can tidy it. **No action needed for this PR.**

### Other checks

- **Chunking arithmetic** — `chunkRows = max(kMinChunkRows, ceil(totalRows/targetTasks))` is always ≥ 2048 > 0, so the `rowBegin += chunkRows` loop terminates; `min(rowBegin + chunkRows, len)` caps the last chunk. `totalRows` is `int64_t` (overflow-safe accumulation), narrowed to `int` only after division — fine for any realistic entity count. ✓
- **Serial fallback** (`g_jobManager == nullptr || (n < kMinForParallel && totalRows < kMinRowsForParallel)`) `continue`s before touching `chunks_`, so no stale chunk content leaks across levels. ✓
- **`chunks_` lifetime** — member vector, `clear()` + `push_back` reuses capacity (high-water-mark), built on the main thread before `parallelFor`, read-only inside `composeChunkRange`. No race. ✓
- Per-level barrier between depths is preserved, so a child level still observes its parent level's completed writes. ✓

Clean approve. Sonnet's checklist (ECS invariants, naming, docs, edge cases) was thorough and I found nothing to revise in it; the one item it deferred — concurrent same-node access — checks out.

Issue #1887, opened 2026-06-17
demos/cmake: prefer irreden_bundle_assets over hand-rolled copy_directory blocks (coding-improvement)


jakildev:
**Class:** B (rule exists in `creations/CLAUDE.md` but didn't fire at authoring time)
**Target artifact:** `creations/CLAUDE.md` (CMake boilerplate section) + optional: `.claude/skills/simplify/SKILL.md` (add CMakeLists smell category)
**Proposed change:** In the "CMake boilerplate" section of `creations/CLAUDE.md`, promote the `irreden_bundle_assets` rule from a prose paragraph to a visible `> NOTE: ...` block or a bullet that is harder to miss. Optionally, add a simplify-scan pass that flags hand-rolled `add_custom_target(*Assets ...)` / `add_custom_command(... copy_directory ...)` blocks in `creations/demos/*/CMakeLists.txt` when the `irreden_bundle_assets` helper is available.

## Context
Surfaced while fixing feedback on PR #1870 (fog_demo). The reviewer's re-review flagged that `creations/demos/fog_demo/CMakeLists.txt` used 38 lines of hand-rolled `add_custom_command` / `add_custom_target` blocks for asset copying. The `creations/CLAUDE.md` already states to prefer `irreden_bundle_assets(<target> SCRIPTS ...)`, citing `shape_debug/CMakeLists.txt` as the canonical reference — but the rule was in a normal prose paragraph that the author missed.

## Why it generalizes
Any worker adding a new demo under `creations/demos/` can repeat this mistake: hand-roll the asset-copy block (copy-paste from an older demo or write from scratch) when a one-liner helper exists. The pattern is entirely mechanical — the presence of `add_custom_target(*Assets ...)` with explicit `copy_directory` commands alongside a `target_link_libraries(... IrredenEngine)` is the smell. Either elevating the rule's visual prominence in `creations/CLAUDE.md` or adding a lightweight CMakeLists check in the simplify pass would catch it before review.

## Occurrences
- PR #1870, `creations/demos/fog_demo/CMakeLists.txt:17–48` — hand-rolled `add_custom_command` / `add_custom_target` block instead of `irreden_bundle_assets(IRFogDemo SCRIPTS config.lua)`.

jakildev:
[...]

PR #1899, opened 2026-06-18
fleet/dispatch: route fleet:design-unblocked feedback to opus class

jakildev:
## Problem

The dispatcher's per-tick class resolver (`fleet_task_class.feedback_pr_class`) classed a `fleet:design-unblocked` feedback PR as **sonnet** — the fall-through default, since the label is in neither the fable nor the blocking-label set. But [`FLEET-FEEDBACK-HANDLING.md`](docs/agents/FLEET-FEEDBACK-HANDLING.md) tier 4 makes design-unblocked resume **"Opus+ classes only; sonnet-class iterations skip this tier."**

So when the scout surfaced engine PR #1885 (`fleet:design-unblocked`) as the worker lane's top-priority feedback item:

1. Resolver routes it → **sonnet** dispatch
2. Sonnet worker reaches tier 4 → **skips** the PR (opus+ only)
3. No claimable open task (all opus + parked/blocked), can't plan (opus+) → **no-op**
4. Feedback is the highest pickup priority, so `chosen` stays sonnet every tick → the PR is starved forever, and the opus `needs_plan` item (#1887) queued behind it is starved with it

Observed live as a worker's *"nineteenth consecutive no-op iteration"*. The scout (`DESIGN_RESUME_LABELS`) and the role doc both treat design-unblocked as opus-tier resume work — only this resolver disagreed.

## Fix

Route `fleet:design-unblocked` → `opus` in `feedback_pr_class`, checked **before** the `fleet:fable` clause (a fable dispatch would skip tier 4 the same way a sonnet one does).

Verified against the live `worker.json` slice: `sonnet high 1` → `opus xhigh 0`.

## Tests

- `test_design_unblocked_routes_opus` — label-level, incl. the `+fleet:fable` precedence case
- `test_design_unblocked_feedback_resolves_opus` — full-resolve, mirroring the real #1885 + #1887 slice shape

All 19 `test_fleet_task_class` cases pass.

### DIFF
--- a/scripts/fleet/fleet_task_class.py
+++ b/scripts/fleet/fleet_task_class.py
@@ -8,11 +8,13 @@ should the next iteration for this lane launch with, and at what effort?
 
 Resolution order mirrors the worker role docs' pickup priority:
 
-  1. feedback PRs   — class from review severity: ``fleet:design-unblocked``
-                      routes to opus (tier-4 resume is opus+-only work);
-                      ``fleet:fable`` on the PR routes the fix to fable
-                      (reviewer judged the approach itself wrong); a blocking
-                      label (``fleet:needs-fix`` / ``human:needs-fix`` /
+  1. feedback PRs   — class from review severity: ``fleet:fable`` on the PR
+                      routes the fix to fable (reviewer judged the approach
+                      itself wrong; fable is opus+ so it also serves a
+                      fable-tagged design-unblocked resume — checked first);
+                      ``fleet:design-unblocked`` otherwise routes to opus
+                      (tier-4 resume is opus+-only work); a blocking label
+                      (``fleet:needs-fix`` / ``human:needs-fix`` /
                       ``human:blocker``) routes to opus; nits-only feedback
                       routes to sonnet.
   2. open tasks     — the oldest claimable task's class (slices are sorted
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet/dispatch: route fleet:design-unblocked feedback to opus class (#1899)

### Praise

- **Root cause diagnosis is exact.** The PR body traces the failure from label → resolver → dispatch → tier-4 skip → no-op loop with clear step numbering. The "nineteenth consecutive no-op iteration" observation is the right kind of evidence.
- **Module docstring kept in sync.** The resolution-order comment at the top of `fleet_task_class.py` now leads with `fleet:design-unblocked`, matching the new check order — makes the routing table self-documenting.
- **Test coverage is thorough for the happy path.** `test_design_unblocked_routes_opus` covers the label-level case; `test_design_unblocked_feedback_resolves_opus` mirrors the exact live scenario (#1885 parked-inflight task + #1887 needs_plan both starved). Expected outputs verified correct by hand.

### Nits

- `scripts/fleet/fleet_task_class.py:97` — The `fleet:design-unblocked` check is placed **before** `fleet:fable`. Per `role-worker.md` line 35, `"opus+" = opus or fable`, and tier-4 (`fleet:design-unblocked`) is listed in the **opus+ column** — so a fable-class worker does handle design-unblocked PRs. If a PR is ever tagged both `fleet:design-unblocked` and `fleet:fable` (e.g. the architect unblocked a fable-tier task), the current order returns `"opus"`, silently downgrading the dispatch. Swapping the order — fable check first, then design-unblocked — preserves the fable dispatch for that edge case without affecting the common path (design-unblocked alone → opus). The FEEDBACK_BLOCKING_LABELS clause after both is unchanged.

- `scripts/fleet/tests/test_fleet_task_class.py:50` — Comment "Wins even over fleet:fable, which would skip it too" is factually wrong per `role-worker.md`. Fable is in "opus+" and does not skip tier 4. The assertion still passes under the current implementation (because the check order makes it true), but the rationale embedded in the comment propagates incorrect documentation about fable's tier-4 behavior. Suggest: "Wins over fleet:fable so a fable-tagged design-unblocked PR routes to opus (conservative; fable is opus+ and can also handle tier 4, but the combined label is rare)." — or simply: "Checked before fleet:fable so a combined label set resolves to opus."

- [x] `python -m pytest scripts/fleet/tests/test_fleet_task_class.py` — author confirms all 19 cases pass.
- [ ] Confirm the fable-before-design-unblocked swap (if accepted) doesn't regress any existing test case; the only affected assertion is the `fleet:design-unblocked + fleet:fable` input, which would change from `"opus"` to `"fable"` — both are correct, but `"fable"` is the more accurate routing.

Opus recheck not required.

jakildev:
Addressed both review nits:

- `fleet_task_class.py` — swapped the order so `fleet:fable` is checked **before** `fleet:design-unblocked`. fable is opus+ (role-worker.md "opus+ = opus or fable") and handles tier-4 resumes, so a both-tagged PR now stays fable instead of silently downgrading to opus. `fleet:design-unblocked` alone still routes to opus — the engine #1885 starvation fix is preserved.
- Corrected the now-stale "fable would skip tier 4" rationale in both the module docstring and the test comment.
- Split the `design-unblocked + fable` assertion into its own test `test_design_unblocked_plus_fable_routes_fable` (now correctly expects `fable`).

All 20 unittest cases pass (was 19; +1 from the split). `fleet:approved` retained — nit cleanups don't invalidate it. — worker

PR #1901, opened 2026-06-18
engine/job: hoist parallelization boilerplate into IRJob (#1900)

### LINKED ISSUE #1900: engine/job: hoist parallelization boilerplate into IRJob (auto-grain + row-chunk planner) and sweep existing call sites
**Model:** opus
**Follow-up to:** #1804 / PR #1892 (intra-node row-range parallelism for `PROPAGATE_TRANSFORM`)

## Problem

PR #1892 added good intra-node row-range parallelism, but it hand-rolls
the whole **dispatch heuristic inline in the system**
(`engine/prefabs/irreden/update/systems/system_propagate_transform.hpp`):

- `targetTasks = workerCount() * 2`
- chunk size = `max(kMinChunkRows, ceil(totalRows / targetTasks))`
- the parallelize-vs-serial decision (`g_jobManager != nullptr &&
  (n >= kMinForParallel || totalRows >= kMinRowsForParallel)`)
- the row-range chunk-list build (`chunks_`, large node → many chunks,
  small node → one)
- the final grain computation `max(1, ceil(numChunks / targetTasks))`
  over the chunk list

None of that is propagate-specific. It's the generic
"**split a work-set across the pool with a sane auto-grain, split a
single dominant node by row-range, and fall back to serial when there's
no pool / too little work**" policy. It belongs **under the hood in
`IRJob`**, where it can be tuned and tested once, instead of being
copy-pasted into every system that wants to fan out. Today the tuning
constants and the serial-fallback guard are re-derived at each call
site, which is exactly the drift this consolidates away.

### Phase 1 — extract the helper into `IRJob`
Add a chunk-planning / auto-grain entry point to
`engine/job/` (`ir_job.hpp` + `ir_job.cpp`) alongside the existing
`parallelFor`. Two candidate shapes (pick during implementation):

- `IRJob::parallelForAutoGrain(int totalItems, fn(begin, end), tuning)`
  — auto-grain + serial fallback for the flat case, and/or
- `IRJob::parallelChunks(span<nodeLength>, fn(nodeIndex, rowBegin, rowEnd), tuning)`
  — the row-range planner that splits a dominant node across workers
  (what propagate needs directly).

A small `IRJob::ParallelTuning` struct should carry the tunables
(`minItemsToParallelize`, `minChunk`, `tasksPerWorker`) with the current
propagate values as defaults. The helper must preserve the existing
behavior exactly: serial when `g_jobManager == nullptr` or below
threshold, `~workers × 2` tasks otherwise, disjoint ranges only (callers
remain responsible for write-disjointness).

### Phase 2 — port `PROPAGATE_TRANSFORM` onto it
Rewrite the `beginTick` dispatch in `system_propagate_transform.hpp` to
call the new helper. This is the API's first consumer and validates the
shape. **Must stay bit-identical** to PR #1892's output — reuse its
verification (`IRShapeDebug --auto-screenshot` 28/28 byte-identical to
master; IRPerfGrid within run-to-run noise floor) and re-confirm the
~47% UPDATE-pipeline win on IRPerfGrid grid-size 64 is retained.

### Phase 3 — sweep the already-parallelized sites
Audit every `IRJob::parallelFor` caller and migrate the ones that
re-derive grain / serial-fallback boilerplate. Known engine call sites
(the `creations/game/engine/**` copies are a synced mirror — do not edit
directly):

- `engine/system/src/system_manager.cpp:468` — the per-system
  `Concurrency::PARALLEL_FOR` tick fan-out
  (`parallelFor(0, node->length_, m_grainSize/kDefaultGrainSize, …)`).
  Best candidate for the auto-grain + big-node split: it currently uses
  a **fixed** grain, so a single huge node gets `length/grain` tasks
  with no `workers × 2` targeting. Decide whether auto-grain should
  kick in when a system leaves `grainSize` unset.
- `engine/system/src/system_manager.cpp:375` — the multi-system
  parallel-group dispatch (`grain = 1` over the system group). Different
  shape (fan out *systems*, not rows), but it at minimum shares the
  `g_jobManager == nullptr` serial-fallback guard the helper should own.

For any site intentionally left on raw `parallelFor` (e.g. group
dispatch if the helper doesn't fit), leave a one-line comment saying
why, so the next sweep doesn't re-flag it.

## Acceptance
- New `IRJob` helper + `ParallelTuning` defaults, unit-tested in
  `test/` (serial fallback, single-large-node split, many-small-node
  granularity, empty/below-threshold no-op).
- `PROPAGATE_TRANSFORM` dispatch reduced to a helper call; the inline
  `chunks_` / `targetTasks` / threshold arithmetic removed from the
  system. Output bit-identical to #1892; IRPerfGrid perf win retained.
- At least the `system_manager.cpp:468` per-system path evaluated and
  either migrated or annotated with why it stays.
- `docs/perf-reports/threading_propagate_transform.md` and
  `engine/prefabs/irreden/update/CLAUDE.md` updated to point at the
  shared helper instead of describing per-system chunk math.
- Both backends build (`fleet-build --target IRShapeDebug` / `IRPerfGrid`);
  `validateAllPipelineGroups` clean.

## Dependencies
Gated on PR #1892 (#1804) merging first — this refactors the code that
PR introduces.

## Note
Pure consolidation / no-behavior-change refactor (Phase 1–2). Phase 3
may *improve* the per-system path (auto-grain for big nodes), which is a
real perf change — keep that in its own commit with its own measurement
if it lands.

jakildev:
## Summary
- **Phase 1** — new `IRJob::parallelForAutoGrain` (flat auto-grain + serial fallback) and `IRJob::parallelChunks` (row-range planner: splits one dominant node across workers, keeps small nodes whole) alongside `parallelFor`. Tunables in `IRJob::ParallelTuning`; defaults reproduce the original PROPAGATE_TRANSFORM constants (≥8 nodes OR ≥4096 rows → parallel; ≥2048-row chunks; 2 tasks/worker), so a default-constructed tuning is bit-identical to the old inline dispatch.
- **Phase 2** — PROPAGATE_TRANSFORM `beginTick` rewritten onto `parallelChunks`; the inline `chunks_` / `targetTasks` / threshold arithmetic is gone. The system mirrors per-node row counts into a reused buffer and hands them to the planner (scratch stays owned by the system so planning is allocation-free on the hot path).
- **Phase 3** — both `system_manager.cpp` `parallelFor` sites evaluated and annotated for why they stay raw: the per-system path keeps its author-chosen `grainSize`; the multi-system group dispatch fans out *systems* at `grain=1` (auto-grain would collapse a small group onto one worker). The `grainSize`-unset auto-grain win is a scoped follow-up per the issue.
- Docs: `engine/job/CLAUDE.md` (new helpers + entry-point list), `engine/prefabs/irreden/update/CLAUDE.md`, and `docs/perf-reports/threading_propagate_transform.md` now point at the shared helper instead of per-system chunk math.

## Test plan
- [x] `IRShapeDebug --auto-screenshot` **28/28 byte-identical to origin/master** (captured both refs, `cmp` each shot — 0 diffs).
- [x] 9 new IRJob unit tests pass: serial fallback (no pool / below threshold), single-large-node row split, many-small-node node-granularity, empty/zero/negative no-op.
- [x] 21 `propagate_transform` correctness tests pass; full `IrredenEngineTest` builds.
- [x] `fleet-build --target IRShapeDebug` + `IrredenEngineTest` clean on macOS/Metal post-format.
- [ ] Linux/GL cross-host smoke (per the `needs-linux-smoke` lane).

Pure consolidation / no-behavior-change refactor; the #1804 ~47% UPDATE-pipeline win is retained by construction (dispatch arithmetic unchanged, no new per-frame allocation).

Closes #1900

### DIFF
--- a/docs/perf-reports/threading_propagate_transform.md
+++ b/docs/perf-reports/threading_propagate_transform.md
@@ -47,20 +47,34 @@ transform from a pre-resolved vector instead of an inline
 > `composeNode` runs serially on one worker. See the dedicated section
 > below for the row-splitting fix and its measurement.
 
-Each level is flattened into a list of row-chunks and dispatched with a
-single `parallelFor`:
+> **Hoisted into `IRJob` by #1900.** The fan-out decision, chunk
+> sizing, and serial fallback below no longer live in the system —
+> they are `IRJob::parallelChunks`, tuned by an `IRJob::ParallelTuning`
+> struct. The old `kMin*` / `targetTasks` constants are now that
+> struct's **defaults**, so `PROPAGATE_TRANSFORM` passes a
+> default-constructed tuning and the numbers stay bit-identical. The
+> system now only mirrors each level's per-node row counts into a
+> reused buffer (`nodeLengths_`) and calls the planner with its
+> reusable `chunks_` scratch; `engine/job/` owns and unit-tests the
+> policy (`test/job/ir_job_test.cpp`). The field names below are the
+> `ParallelTuning` members.
 
-- **`kMinForParallel = 8`** (node count) **OR `kMinRowsForParallel = 4096`**
+Each level is flattened into a list of row-chunks (by
+`IRJob::parallelChunks`) and dispatched with a single internal
+`parallelFor`:
+
+- **`minNodes_ = 8`** (node count) **OR `minItemsToParallelize_ = 4096`**
   (total entity rows) — a level parallelizes when it has at least this
   many archetypes *or* this many total rows. Below both, the dispatch +
   wait overhead exceeds the savings and the level composes serially.
-- **`kMinChunkRows = 2048`** — the minimum rows per chunk. A node smaller
+- **`minChunk_ = 2048`** — the minimum rows per chunk. A node smaller
   than the chunk size stays whole (one chunk), preserving the original
   per-archetype granularity for many-small-node levels; a node larger
   than the chunk size splits into multiple row-range chunks. The chunk
-  size is `max(kMinChunkRows, ceil(totalRows / targetTasks))` with
-  `targetTasks = max(1, workerCount * 2)`, so a single dominant node
-  splits into ~`targetTasks` chunks (≈2 per worker).
+  size is `max(minChunk_, ceil(totalRows / targetTasks))` with
+  `targetTasks = max(1, workerCount * tasksPerWorker_)` (default
+  `tasksPerWorker_ = 2`), so a single dominant node splits into
+  ~`targetTasks` chunks (≈2 per worker).
 - **Grain = `max(1, ceil(numChunks / targetTasks))`** over the chunk
   list — `1` when chunks already number ≤ `targetTasks` (single large
   node), grouping when many small nodes produce more chunks than tasks.
--- a/engine/job/CLAUDE.md
+++ b/engine/job/CLAUDE.md
@@ -1,14 +1,17 @@
 # engine/job/ — IRJob worker pool
 
 Phase 1 of the multithreading epic (#226). The module stands up an
-enkiTS-backed worker pool that `World` owns for its lifetime; no
-engine system schedules on it yet — T-222 is the first consumer.
+enkiTS-backed worker pool that `World` owns for its lifetime.
+`PROPAGATE_TRANSFORM` and the `SystemManager` `PARALLEL_FOR` fan-out
+are the first consumers; the auto-grain dispatch helpers below (#1900)
+consolidate the boilerplate they share.
 
 ## Entry point
 
 `engine/job/include/irreden/ir_job.hpp` — `IRJob::parallelFor`,
-`IRJob::run`, `IRJob::pinTo`, `IRJob::isMainThread`,
-`IRJob::workerId`, `IRJob::workerCount`, `IRJob::workerRng`. The
+`IRJob::parallelForAutoGrain`, `IRJob::parallelChunks`, `IRJob::run`,
+`IRJob::pinTo`, `IRJob::isMainThread`, `IRJob::workerId`,
+`IRJob::workerCount`, `IRJob::workerRng`. The
 global pointer `g_jobManager` follows the same `g_*Manager` pattern
 as `g_entityManager` — set by `JobManager`'s ctor, cleared by its
 dtor, valid only between `World` construction and destruction.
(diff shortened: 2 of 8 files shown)


jakildev (commented):
## Review — engine/job: hoist parallelization boilerplate into IRJob (#1900)

### Praise

- The three-phase structure is well-scoped: new helpers (Phase 1), PROPAGATE_TRANSFORM migration (Phase 2), explicit non-migration rationale for the two system_manager sites (Phase 3). The Phase 3 annotations in `system_manager.cpp` are particularly valuable — they explain *why* not to migrate, which prevents a future author from "fixing" them.
- `ParallelTuning` defaults ARE the original PROPAGATE_TRANSFORM constants, so the refactor is provably bit-identical by construction, and the 28/28 screenshot comparison confirms it at the integration level.
- Excellent test coverage: 9 new unit tests cover serial fallback (no-pool and below-threshold), parallel coverage (each item visited exactly once), many-small-nodes whole-node preservation, empty/zero/negative no-ops, and the intra-node split path. Each test exercises a distinct logical path through the helpers.
- The `scratch` buffer ownership design (caller-owned, passed by reference) keeps `parallelChunks` allocation-free on the hot path while avoiding hidden lifetime coupling — clean.

### Nits

- `engine/job/src/ir_job.cpp:121` — `totalRows` is `int64_t` but is cast to `int` when computing `chunkRows`: `static_cast<int>((totalRows + targetTasks - 1) / targetTasks)`. Compare the analogous `parallelForAutoGrain` cast which guards with `static_cast<int64_t>(totalItems)` first. In practice `totalRows / targetTasks` cannot exceed `INT_MAX` (that would require billions of entity rows — more than fits in memory with today's archetype layout), so this is theoretical. The `parallelForAutoGrain` guard at line 79 is a bit more defensive by habit; no change required.

- [x] `IRShapeDebug --auto-screenshot` 28/28 byte-identical to `origin/master` (confirmed in PR body)
- [x] 9 new IRJob unit tests pass; 21 PROPAGATE_TRANSFORM correctness tests pass (confirmed)
- [x] `fleet-build --target IRShapeDebug` + `IrredenEngineTest` clean on macOS/Metal (confirmed)
- [ ] Linux/GL cross-host smoke (tagged `fleet:needs-linux-smoke` — pending fleet runner pickup)
- [ ] Windows-native build smoke (tagged `fleet:needs-windows-smoke`)

Opus recheck not required.

jakildev (commented):
## Opus recheck — engine/job: hoist parallelization boilerplate into IRJob (#1900)

**Verdict:** approve

Sonnet approved-with-nits and marked "Opus recheck not required," but this PR *is* the shared parallel-decomposition primitive (`IRJob::parallelChunks` / `parallelForAutoGrain`), so I took the concurrency pass regardless. On a close read of the race surface and the behavioral-equivalence claim, I confirm the approve.

### Concurrency correctness (the Opus-only pass Sonnet couldn't fully confirm)

- **`parallelChunks` worker closure is read-only over shared state.** `scratch` is filled entirely on the calling thread *before* `parallelFor`, and the dispatched closure only reads `scratch[c]` and forwards to `fn` — no `push_back`/resize during the parallel region, so there's no data race on the vector and no iterator/pointer-invalidation hazard. Write-disjointness (each entity writes only its own `C_WorldTransform[i]`) is the caller's contract and is honored by the row-range partition (`[rowBegin, rowEnd)` ranges are non-overlapping by construction). Race-free.
- **Bit-identical to the prior hand-rolled dispatch — verified by construction, not just by the 28/28 screenshot.** I diffed the moved logic line-for-line: `runParallel` predicate (`n >= 8 || totalRows >= 4096`), `targetTasks = max(1, workers*2)`, `chunkRows = max(2048, ceil(totalRows/targetTasks))`, the chunk-build loop, and `chunkGrain` all reproduce the old `System<PROPAGATE_TRANSFORM>` constants exactly. `ParallelTuning` defaults are those same constants, and the system passes a default-constructed tuning. The serial fallback (`fn(i, 0, nodeLengths_[i])` per node, where `nodeLengths_[i] == node->length_`) matches the old serial loop, including the zero-length-node case (no chunk emitted in parallel; no-op `composeNodeRows(…,0,0,…)` serially).
- **Null-pool contract is correctly relaxed.** `parallelFor` asserts on `g_jobManager == nullptr`; both new helpers gate on it first and fall to a single serial pass instead. The `IRJobChunksNoPool` / `IRJobAutoGrainNoPool` tests pin this, and it matches the guard PROPAGATE_TRANSFORM already used.
- **Phase 3 non-migrations are the right call.** The two `system_manager.cpp` sites are correctly left on raw `parallelFor`: the group fan-out dispatches *systems* at `grain=1` (auto-grain would batch them onto `~tasksPerWorker` tasks and collapse a small group onto one worker — a real parallelism regression), and `executeSystem` uses the author-chosen per-system `kGrainSize`. The inline annotations explain *why not*, which protects against a future "consistency" refactor. Good judgment to scope the grainSize-unset auto-grain change to a separate follow-up rather than smuggle a perf change into a bit-identical refactor.

### On Sonnet's nit

Sonnet flagged the `static_cast<int>` of `(totalRows + targetTasks - 1) / targetTasks` in `parallelChunks` as less defensive than `parallelForAutoGrain`'s `static_cast<int64_t>` guard. On closer read I agree it's purely theoretical and **no change required**: `totalRows` is already `int64_t`, so the division runs in 64-bit and only the quotient is narrowed — overflow would need `totalRows/targetTasks > INT_MAX`, i.e. billions of rows per chunk, which can't fit in the archetype storage being iterated. Not re-flagging.

Clean, well-scoped, thoroughly unit-tested. Leaving `fleet:has-nits` for Sonnet's theoretical nit; safe to merge once the cross-host smoke labels clear.

jakildev:
Reviewed fleet:has-nits: the single nit (ir_job.cpp:121 int64_t→int narrowing) was explicitly marked 'no change required' by the reviewer — the arithmetic is already int64_t due to totalRows, making it theoretically safe. Cleared has-nits label; fleet:approved remains valid. — worker-4

PR #1918, opened 2026-06-20
docs: plan #1910 — GPU depth-readback probe (commit missing epic-#1881 child plan)

jakildev:
## Summary
Commits the missing plan file for #1910 so it can queue for implementation.

#1910 ("GPU depth-readback probe for composite depth-test debugging") was filed as a child of epic #1881 but — unlike its siblings (`issue-1881/1883/1884/1908.md`) — never got a committed `.fleet/plans/issue-1910.md`. The `fleet-queue-ingest` planning gate (#1456) bounces any approved issue with no plan file to `fleet:needs-plan`, so #1910 couldn't reach `fleet:queued` even once its (unrelated) false-BLOCK was cleared.

The issue body already contains a complete opus plan; this lands it as the committed plan file in the sibling format (Why / Scope / Approach / Affected files / Acceptance criteria / Gotchas / References).

## Context — why #1910 was stuck
Two independent issues kept #1910 out of the queue:
1. **False-BLOCK (fixed on the issue):** the `**Blocked by:**` field read `_(none — … in parallel with #1883)_`. The shared parser's anti-evasion guard (`fleet_blocked_by.is_no_blocker_value`) extracts any `#N` it sees, so it treated the parallel-sibling `#1883` as a hard blocker and every worker was gate-refused. The field has been reworded (the `#1883` note moved to a `**Parallel with:**` line) so the live body now parses as unblocked.
2. **No plan file (this PR):** the planning-gate bounce described above.

## Test plan
- [x] Plan-content only; no build impact.
- [x] Sibling-format parity with `.fleet/plans/issue-1908.md`.
- [ ] Reviewer confirms the plan faithfully reflects #1910's acceptance criteria.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1910, opened 2026-06-19
render: GPU depth-readback probe for composite depth-test debugging (unblocks #1884)


jakildev:
**Model:** opus
**Part of:** epic #1881 (tooling prerequisite for child #1884)
**Blocked by:** (none) — independent tooling; can start immediately.
**Parallel with:** #1883 (named sibling, not a dependency — runs concurrently; both are children of epic #1881)
**Unblocks:** #1884 (unify depth/clipping)

## Why (motivation)
#1884 has two confirmed "behind-face wins" composite-depth crossings:
- **A** — a floating detached cube (canvas_stress canary, world z=−8, *above* the floor z∈[2,6]) clips behind the SDF floor at high zoom.
- **B** — a per-axis cube shows a Y-over-X doubled face-stripe at cardinal-180 (proven NOT a margin/yield issue — persists at yield scale 20).

Both need the **actual stored composite depth** at specific pixels, decoded per render path, to root-cause. This session's shader-output approximation (render depth as grayscale) gave **contradictory** data — it loses precision (8-bit), can't be decoded per-path without knowing which path won the pixel, and doesn't read the true depth-test value across the entity-composite vs main-canvas passes. We need a trustworthy **GPU→CPU readback**, not another shader-output approximation or parameter sweep. Reusable for any future depth-ordering bug.

## Scope
A debug path that, for a requested screen pixel (or small set), reads and logs the **real** depth-test value(s) in a reconciled, comparable form:
1. **Framebuffer depth attachment** at the pixel — the final composite winner's `gl_FragDepth`. Decode to raw distance via the exact inverse of `normalizeDistance` (`dist = norm·(kMax−kMin) + kMin`, same `kTrixelDistanceMin/MaxDistance`).
2. (stretch) **Per-contributing-canvas distance texture** at the corresponding texel, each decoded by ITS path's encoding so the same world point can be compared across paths:
   - voxel grid / SDF floor: `encodeDepthWithFace` / `baseDepth·4` → `rawDepth = dist>>2`.
   - per-axis scatter: `#1458` `(depth<<10)|(uFrac<<6)|(vFrac<<2)|slot` → `rawDepth = dist>>10`.
   - detached canvas: `round(rawDist·depthScale) + distanceOffset` (the #1624 lift; `depthScale = effSub/cubeSub`).

[...]

PR #1919, opened 2026-06-20
fleet: blocked-by parser honors the "(none — in parallel with #N)" sibling idiom

jakildev:
## Summary
Hardens the shared `Blocked by:` parser (`scripts/fleet/fleet_blocked_by.py`) so the legitimate **"see also / in parallel with #N"** sibling idiom no longer false-BLOCKs an otherwise-unblocked issue — while keeping the anti-evasion guard for `none, actually blocked by #5` fully intact.

## Why
`is_no_blocker_value()` had a blunt guard: *any* `#N` in the field value → treat as a real blocker (to stop `none, actually blocked by #5` evasion). That also catches the legitimate idiom where a `#N` is named as a **parallel sibling**, not a dependency.

Concrete incident — **#1910**: its field read `_(none — independent tooling; can start immediately, in parallel with #1883)_`. The parser extracted the parallel-sibling `#1883`, `fleet-claim` gate-refused every worker, and because the guard is *deliberate* it never self-healed — the issue stranded under `fleet:needs-human` (a worker had to park it). This is the inverse of the #1749 parser coverage work: a false-**BLOCK**, not a false-free.

## The fix (targeted — anti-evasion preserved)
A `#N` is excused as a sibling cross-reference **only when both** hold:
1. The value **leads with a `none`/`(none`/`n-a`/`tbd` sentinel** (now tolerant of leading markdown emphasis, so the italic `_(none …)_` form is recognized). A value that doesn't lead with `none` (a bare `#5`) is never excused.
2. **Every** `#N` is introduced by a see-also/parallel qualifier (`see also`, `related`, `in parallel with`, `parallel with`, `sibling of`, `alongside`, `concurrent with`, `cf.`), checked **per-ref** against the clause leading up to it.

Any ref carried by a **blocker verb** (`blocked by`, `depends on`, `requires`, `needs`, `waiting on`, `gated on`) — or with **no qualifier at all** — still gates. So:
- ✅ `(none — runs in parallel with #1883)` → unblocked
- ✅ `_(none — … in parallel with #1883)_` → unblocked (#1910's exact form)
- ⛔ `none, actually blocked by #5` → still a real blocker (#5 surfaces)
- ⛔ `none — see also #5, blocked by #6` → mixed → gates
- ⛔ `in parallel with #1883` (no leading `none`) → gates

Propagates to all three consumers (scout, ingest, claim) via the shared module — no per-script copies.

## Test plan
- [x] New `SeeAlsoParallelIdiom` test class in `test_blocked_by.py`: excused positives (incl. the italic form), anti-evasion negatives, mixed see-also+blocker, and the no-leading-`none` guard.
- [x] Full parser corpus green: `python3 -m unittest test_blocked_by test_parse_blocked_by_none_sentinel` → 40 tests OK.
- [x] Broader parser-dependent sweep green: + `test_issue_queue_parser test_live_blocker_resolution test_enrich_stackable_blocker_prs test_scout_degraded_fetch test_epic_steward_projection` → 157 tests OK.
- [x] Docs-adjacent / Python-only; no engine build impact.

## Related
Unblocks the parser-level recurrence behind #1910 (the issue itself is separately unblocked: field reworded + plan file in #1918). Companion to the #1749 shared-parser consolidation.

### DIFF
--- a/scripts/fleet/fleet_blocked_by.py
+++ b/scripts/fleet/fleet_blocked_by.py
@@ -76,6 +76,55 @@ _REF_RE = re.compile(r'(?:[A-Za-z0-9][\w.-]*/)?([A-Za-z0-9][\w.-]*)?#(\d+)')
 _REF_NAME_TO_SLUG = {'irredenengine': 'jakildev/IrredenEngine',
                      'irreden': 'jakildev/irreden'}
 
+# See-also / parallel-sibling qualifiers (#1910): a `#N` introduced by one of
+# these inside a leading-`none` value is a cross-reference, not a blocker — the
+# legitimate "(none — runs in parallel with #N)" idiom. Matched per-ref against
+# the clause leading up to the ref so a sibling note can't smuggle a real
+# dependency past the gate (see _ref_is_see_also).
+_SEE_ALSO_RE = re.compile(
+    r'\b(?:see[\s-]+also|related(?:\s+to)?|in\s+parallel(?:\s+with)?|'
+    r'parallel(?:\s+(?:with|to))?|sibling(?:\s+of)?|alongside|'
+    r'concurrent(?:ly)?(?:\s+with)?|cf\.?)\b',
+    re.IGNORECASE,
+)
+# Blocker verbs (#1910): a `#N` introduced by one of these is a real dependency
+# even inside a leading-`none` value — this is what keeps the anti-evasion guard
+# intact for "none, actually blocked by #5".
+_BLOCKER_VERB_RE = re.compile(
+    r'\b(?:blocked\s+by|blocks?|depends?\s+(?:on|upon)|requires?|needs?|'
+    r'waiting\s+(?:on|for)|gated\s+(?:on|by)|blocker)\b',
+    re.IGNORECASE,
+)
+# Clause boundaries used to isolate the text introducing a single `#N`.
+_CLAUSE_SPLIT_RE = re.compile(r'[;,—–(]')
+
+
+def _leads_with_none_sentinel(value):
+    """True when `value` opens with a no-blocker sentinel — `none`, `(none`,
+    `n/a`, `na`, `tbd`, or a lone leading dash/dot. This is the necessary
+    precondition for excusing any `#N` as a see-also reference: a value that
+    does not lead with `none` (e.g. a bare `#5`) is never excused."""
+    # Tolerate leading markdown emphasis / paren wrappers (`_(none …)_`,
+    # `*none*`, `` `(none)` ``) before the sentinel word — #1910's field was
+    # italicized, and the bare token check would otherwise see `_(none`.
+    head = (value or "").strip().lower().lstrip("_*`(").strip()
+    if not head or head[0] in "-–—.":
+        return True
+    token = re.split(r"[\s.,;:)\-–—]", head, maxsplit=1)[0]
+    return token in {"none", "n/a", "na", "tbd"}
+
+
+def _ref_is_see_also(value, ref_start):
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — PR #1919: fleet: blocked-by parser honors the "(none — in parallel with #N)" sibling idiom

**Verdict:** approve

### Summary

Fleet-only Python change (`scripts/fleet/fleet_blocked_by.py` + tests). No C++ ECS, render pipeline, shader, or serialization surfaces touched — engine checklist items are not applicable here. Focused on the #1910 false-BLOCK incident.

### Logic correctness

**`_leads_with_none_sentinel`** correctly strips leading markdown emphasis wrappers (`_(`, `*(`, `` `( ``) before identifying the sentinel token. The `lstrip("_*`(")` is intentionally left-side only, leaving trailing chars for `re.split(r"[\s.,;:)\-–—]", head, maxsplit=1)` to split on `)` and surface `none` cleanly. Verified on the #1910 canonical form `_(none — … in parallel with #1883)_`. ✓

**`_ref_is_see_also`** scans only the clause leading up to a `#N` (back to the previous `;,—–(` boundary), not the full value. This prevents a prior clause's `see also #5` from whitewashing a later clause's `blocked by #6`. The test `"none — see also #5, blocked by #6" → gates` confirms it. ✓

**`is_no_blocker_value` dual-condition guard**: `all(_ref_is_see_also(...) for m in refs)` means even one unqualified or blocker-verb-qualified ref disqualifies the excuse. Conservative and correct. ✓

**Propagation**: All three consumers (`fleet-claim`, `fleet-state-scout`, `fleet-queue-ingest`) import `is_no_blocker_value`/`parse_blocked_by`/`blocker_refs` from this shared module. The fix reaches all three automatically — no inline copies to update. ✓

**Tests**: 33/33 pass (run via `python3 -m unittest scripts/fleet/tests/test_blocked_by.py`). New `SeeAlsoParallelIdiom` class covers excused cases, anti-evasion cases (blocker verb + no qualifier), integration with `parse_blocked_by`/`blocker_refs`, and the no-leading-`none`-sentinel gate. The regression case (`"none, #138"` still gates) is verified by the pre-existing `test_mixed_sentinel_and_ref_keeps_ref`. ✓

### Nits

- `scripts/fleet/fleet_blocked_by.py:107-108` — The comment on `_leads_with_none_sentinel` lists `` `*none*` `` as an example of a handled form, but the implementation does not handle it: `lstrip("_*`(")` strips only the leading `*`, leaving `none*`, and `re.split(r"[\s.,;:)\-–—]", "none*", ...)` produces token `"none*"` (not in the sentinel set). The function returns `False` for `*none*`. Only the `_(none …)_` and `` `(none)` `` paren-wrapper forms that happen to lstrip into `none …)` (where `)` is in the split set) work as described. The fix for #1910 is correct regardless — `_(none …)_` is the actual form — but the comment over-claims. Either remove `*none*` from the example list or add `*` to the clause-split regex.

- [x] `python3 -m unittest scripts/fleet/tests/test_blocked_by.py` (all 33 pass — verified in this review)
- [ ] Live smoke: confirm `fleet-claim claim <1910-issue-number>` succeeds with the actual #1910 `Blocked by:` field value (or equivalent repro), if not already done

Opus recheck not required.

jakildev:
Addressed nit at `fleet_blocked_by.py:113`: added `*` to the token-split regex so `*none*` is correctly handled (not just comment-claimed). Also added a test case for the bold form (`"*none* — in parallel with #1883"`) to the `SeeAlsoParallelIdiom` suite. All 33 tests pass. — worker-3

Issue #1926, opened 2026-06-20
render: #1352 CAMERA_CENTER default pivot inverts the yaw-drift sign (doubles focus motion instead of pinning)


jakildev:
## Context
Discovered while implementing #1921 (point-of-interest pivot). The legacy #1352 `CAMERA_CENTER` default branch in `IRRender::getEffectiveCameraIso()` (`engine/render/src/ir_render.cpp`) returns `pos3DtoPos2DIsoYawed(isoPixelToPos3D(cameraIso, 0), yaw)`.

## Finding
Producers place content on the canvas at `iso_canvas(W) = isoYawed(W, yaw)` and the composite **adds** the camera offset, so a point's screen position is `isoYawed(W, yaw) + offset`. To pin a focus `F`, the offset must **cancel** `F`'s yaw-induced canvas drift (`offset = cameraIso - (isoYawed(F,yaw) - iso(F))`). The #1352 default instead uses `+isoYawed(F0, yaw)`, which **adds** the drift — so `screen(F0) = 2·isoYawed(F0, yaw)`, i.e. a panned-and-rotated camera moves the screen-center point by **twice** the canvas rotation rather than pinning it (arcs worse than `ORIGIN` mode).

### Evidence (#1921 PR, shape_debug --pivot-focus-demo, measured pillar centroid)
- offset with `+isoYawed` term (= #1352's sign), applied to a focus: drift **1024px** (= 2× the 512px canvas-only arc).
- offset with `-isoYawed` term (cancelling, the #1921 focus branch): drift **~0** (pinned).

`ORIGIN` mode and the no-pan / no-rotate fast path are unaffected (the drift term is zero there), so this is invisible except when the camera is BOTH panned and rotated.

## Why not fixed in #1921
#1921 adds a correct opt-in focus branch (cancelling sign) and keeps the default byte-identical per the issue's fast-path requirement. Correcting the default sign changes panned+rotated `CAMERA_CENTER` output (the existing `zoom4_pan16_yaw{90,180,45}_pivot` shots) and needs **cross-host render-verify reference regeneration** (Linux + macOS) — a separate, process-touching change.

## Suggested fix
Switch the default branch to the cancelling form for `F0 = isoPixelToPos3D(cameraIso, 0)`:
`return cameraIso - pos3DtoPos2DIsoYawed(F0, yaw) + pos3DtoPos2DIso(F0);`
[...]

PR #1927, opened 2026-06-20
render: pivot camera Z-yaw about an explicit point of interest

### LINKED ISSUE #1921: render: camera Z-yaw should pivot about the point-of-interest, not the z=0 screen-center point
**Model:** opus
**Part of:** epic #1881 (rotated-voxel correctness under camera Z-yaw)
**Blocked by:** (none) — independent (camera / `ir_render` path; no shared-shader conflict, can run in parallel with the other children).
**Related:** #1352 / PR #1362 (camera focus pivot) — this is the follow-up, NOT a regression (that code is intact).

## Why (motivation)
On IRShapeDebug the point the camera rotates about is wrong: shapes **arc across the screen** instead of rotating in place. Two confirmed defects in the #1362 focus-pivot at `engine/render/src/ir_render.cpp:44-60` (verified intact — NOT a regression):

1. **Pivot pinned to z=0.** `getEffectiveCameraIso()` reconstructs the focus as `IRMath::isoPixelToPos3D(cameraIso, 0.0f)` — the world point under screen center on the **z=0 plane**. Content with height (z>0) orbits its z=0 base, so tall shapes swing/arc.
2. **Pivot is the screen-center world point, not the point-of-interest.** The #1362 correction "is the identity unless the camera is BOTH panned and rotated" (per `creations/demos/shape_debug/main.cpp:134-143`). In the default un-panned scene, off-center shapes (placed at x=16, 32, …) all orbit the **world origin** — the "wrong focus" the user observes. The unshipped slice of #1352 was a cursor/selection-anchored pivot.

## Scope
Make camera Z-yaw rotate about the actual point-of-interest at its true depth, so content rotates in place regardless of pan/height.

## Approach (planner settles the pivot policy)
- **Pivot DEPTH:** derive the focus depth from content under the screen center (e.g. the #1910 composite depth-probe at screen center, or the hovered/selected entity's depth) instead of hardcoding `0.0f`.
- **Pivot TARGET** (design choice for the plan): (a) cursor-anchored, (b) selected-entity-anchored, (c) scene/content centroid. Recommended default: cursor-anchored with a scene-centroid fallback, exposed via the existing `RotationPivotMode` enum.
- **visualYaw vs rasterYaw:** confirm the pivot re-projection uses the same yaw basis the rasterizer commits (cardinal + residual), to avoid a lurch at cardinal-bracket boundaries.

## Acceptance criteria
- Across a full yaw sweep at multiple pans and zooms, the chosen pivot point stays pinned at screen center within ≤1 game-pixel; on-screen content rotates in place (no off-center arc), including content at z>0.
- Cardinal fast path + `ORIGIN` mode byte-identical (A/B via `--pivot-origin`).
- New shape_debug shots capturing a **tall** shape rotating in place (pivot pinned) as regression coverage.
- Both backends.

## Files (start)
`engine/render/src/ir_render.cpp:44-60` (getEffectiveCameraIso); `engine/render/src/render_manager.*` (RotationPivotMode, pivot target); `engine/math/include/irreden/ir_math.hpp` (isoPixelToPos3D / pos3DtoPos2DIsoYawed); `creations/demos/shape_debug/main.cpp` (pivot shots); optionally #1910 depth-probe for focus-depth.

## Gotchas
- Iso projection degenerates near yaw ±2π/3 (det≈0) — the pan inverse already guards this in `ir_math.hpp`; the pivot path must too.
- Keep `ORIGIN` mode and the no-pan/no-rotate fast path byte-identical.

## References
#1352 / PR #1362 (focus pivot, intact); #1420 / PR #1496 (screen-relative WASD); #1910 (depth-probe, candidate focus-depth source); epic #1881.

jakildev:
## Summary
- `IRRender::getEffectiveCameraIso` gains an opt-in **pivot focus**: `setRotationPivotFocus(worldPoint)` makes CAMERA_CENTER Z-yaw pivot about that point at its **true depth**, cancelling its yaw-induced canvas drift, so tall / off-center content rotates in place instead of arcing about the z=0 screen-center point (#1921 defects 1 + 2). `clearRotationPivotFocus` reverts to the legacy default.
- Choosing the focus (cursor / selection / scene centroid / a future #1910 depth-probe) is a creation-level policy; the engine just consumes the world point. The legacy default branch is **left byte-identical** per the issue.
- `AutoScreenshotShot` carries an optional per-shot focus, applied via a new shared `applyShotCameraState` helper (de-dups the auto-screenshot + GUI-test camera-apply blocks).
- `shape_debug --pivot-focus-demo` drives a dedicated tall-pillar regression scene (pillar + four orbiting markers); add `--pivot-origin` for the legacy-pivot A/B.
- **CPU camera-math only — no shader changes, so both backends are covered.**

## Verification (measured pillar centroid, full yaw sweep, screen center = 1280,720)
| yaw | fix (pinned) | legacy (`--pivot-origin`, arcs) |
|---|---|---|
| 0 | (1280,720) | (1280,720) — byte-identical fast path |
| 90 | (1250,735) | (1506,863) |
| 180 | (1280,750) | (1792,750) |
| 270 | (1309,735) | (1565,607) |

Fix pins the pillar at screen center (~±30px from silhouette-shape change as faces rotate; the pivot world point is sub-pixel). Legacy arcs it in a ~512px-radius circle. Screenshots in `docs/pr-screenshots/1921-camera-pivot-focus/`.

## Heads-up — filed #1926
While implementing I found the legacy #1352 default branch applies the yaw-drift with the **inverted sign** (doubles the focus's motion instead of pinning it — producers add the offset, so the drift must cancel, not compound). Fixing it changes panned+rotated output and needs cross-host render-verify reference regeneration, so it's left byte-identical here and tracked in #1926. The new focus branch uses the correct cancelling sign.

## Test plan
- [x] `fleet-build --target IRShapeDebug` (macOS/Metal) clean
- [x] `--pivot-focus-demo` pins the pillar; `--pivot-origin` arcs it (measured above)
- [x] yaw-0 / ORIGIN byte-identical (fast path untouched); default scene byte-identical (default branch + shot fields are opt-in)
- [ ] Linux/OpenGL smoke (cross-host)

Closes #1921


jakildev (commented):
## Review — render: pivot camera Z-yaw about an explicit point of interest (#1927)

**Verdict:** approve (Opus recheck required — see escalation footer)

CPU-only camera math change that adds an opt-in explicit pivot focus to
`getEffectiveCameraIso`. The math is well-commented, the byte-identical
default guarantee is verified, and the extraction of `applyShotCameraState`
is a clean housekeeping win. No shader changes, no ECS violations, no
ownership concerns.

### ECS invariants

No tick functions touched. The RenderManager state additions (`m_rotationPivotFocus`,
`m_hasRotationPivotFocus`) are POD fields — no ECS footgun exposure. ✓

### Ownership / lifetime

`g_pivotFocusShotLabels` is reserved before the loop (`reserve(n)`) and
holds exactly `n` elements after it, so no reallocation occurs and the
`label.data()` pointers stored in `g_pivotFocusShots[i].label_` are stable
for process lifetime. Pattern matches the existing `g_cullValidateShotLabels`
idiom. ✓

`AutoScreenshotShot::pivotFocusWorld_` and `hasPivotFocus_` are value
members — no pointer ownership concern. ✓

### Render pipeline / math

Core change — `getEffectiveCameraIso` (`engine/render/src/ir_render.cpp`):

```
offset = cameraIso − isoYawed(F, yaw) + iso(F)
```

Comment derives it clearly: screen position of F is `isoYawed(F, yaw) + offset`
and must equal `iso(F) + cameraIso` (pinned). Cancelling gives the coded
expression. At `visualYaw == 0`, `isoYawed(F, 0) == iso(F)` so the expression
reduces to `cameraIso` — byte-identical to the pre-PR path. ✓

`visualYaw` is hoisted before the branch (`const float visualYaw = IRPrefab::Camera::getYaw()`)
so the legacy path's behavior is unchanged — it still returns
`pos3DtoPos2DIsoYawed(cameraFocusWorld, visualYaw)`. ✓

**Legacy sign note (#1926):** the PR calls out clearly that the CAMERA_CENTER
legacy path carries the wrong sign (adds yaw-induced drift rather than
cancelling it) and tracks the fix in #1926. Kept byte-identical here by
design. Noted, not flagged. ✓

No shader files changed; both GL and Metal backends covered by the CPU-only
path. ✓

No `std140` struct layout changes. ✓

### `applyShotCameraState` extraction

The helper is correctly placed in the anonymous namespace of `auto_screenshot.cpp`
and deduplicates the camera-state-apply block between `createAutoScreenshotSystem`
and `createGuiTestSystem`. The added `hasPivotFocus_` branch calls
`clearRotationPivotFocus()` when unset, ensuring a focused shot can't bleed
into the next unfocused one. All existing shots default to `hasPivotFocus_ = false`,
so calling `clearRotationPivotFocus()` on their transitions is a no-op — byte-identical. ✓

### Naming / style

- Private members: `m_rotationPivotFocus`, `m_hasRotationPivotFocus` — `m_` prefix ✓
- Struct public members: `pivotFocusWorld_`, `hasPivotFocus_` — trailing `_` ✓
- Public API: `setRotationPivotFocus`, `clearRotationPivotFocus`, `hasRotationPivotFocus`,
  `getRotationPivotFocus` — camelCase, consistent with sibling API ✓
- `Marker` struct in `initPivotFocusScene()`: public members carry trailing `_` ✓

### Tests / build

- macOS/Metal build clean per PR body ✓
- Verification table (measured pillar centroid at 90/180/270) in PR body confirms pinning ✓
- `--pivot-focus-demo` scene is isolated (separate `initPivotFocusScene()`, early return in
  `initEntities()`) so it cannot contaminate the default scene ✓
- Linux/GL smoke outstanding per unchecked `[ ] Linux/OpenGL smoke` in test plan — labelled
  below

### Nits

- `creations/demos/shape_debug/main.cpp` (the `kCropsPivotPillar` constant) — crop
  coordinates `{1130, 460, 300, 520}` are tuned for the HiDPI 2x macOS framebuffer
  (framebuffer center ≈ (1280, 720)). On a Linux/1x host the crop window will land at
  a different framebuffer position. The comment already says "refine against the capture
  on a non-scaled host," so whoever does the Linux smoke should expect to adjust these
  values (or accept that the crop region captures adjacent area rather than the pillar
  center). Flagging so the Linux smoke agent is aware.

### Praise

- Clean `applyShotCameraState` extraction eliminates future camera-state drift between
  the two shot-cycling paths.
- Byte-identical guarantee documented at three levels: code comment, PR body verification
  table, and explicit `--pivot-origin` A/B comparison.
- The `#1926` latent-sign bug is explicitly called out in a code comment rather than
  silently corrected, which makes the two branches' differing signs understandable at
  a glance.

- [ ] Linux/OpenGL build + `IRShapeDebug --pivot-focus-demo --auto-screenshot` runs
      clean; pillar stays near screen center across the yaw sweep; crops may need
      adjustment for 1x DPI
- [ ] `fleet-build --target IRShapeDebug` on Windows/MSYS2 clean

Opus recheck required: touches engine/render/src/ir_render.cpp — `getEffectiveCameraIso` pixel-drift cancellation formula and the sign relationship between the new focus branch and the legacy CAMERA_CENTER path warrant Opus verification.

jakildev (commented):
## Opus recheck — render: pivot camera Z-yaw about an explicit point of interest (#1927)

**Verdict:** approve (Opus-confirmed)

Picked this up on Sonnet's explicit escalation. Sonnet correctly cleared
ECS/ownership/naming and read the math at a high level; I focused the Opus
pass on the two things it flagged for verification — the
`getEffectiveCameraIso` pixel-drift **cancellation formula** and the **sign
relationship** between the new focus branch and the legacy CAMERA_CENTER
path — and independently re-derived both.

### Focus-branch cancellation formula (the Opus-only concern)

Sonnet asserted the formula pins `F` and collapses to `cameraIso` at yaw 0; on
closer read I confirm, and I checked the stronger property (that the *scene
rotates about F*, not merely that F is pinned):

The composite model is `screen(W) = isoYawed(W, yaw) + offset`. With
`offset = cameraIso − isoYawed(F, yaw) + iso(F)`:

- `screen(F) = isoYawed(F,yaw) + cameraIso − isoYawed(F,yaw) + iso(F) = cameraIso + iso(F)` — constant in yaw, so F is pinned. ✓
- `iso` and the yaw rotation are both linear, so `isoYawed` is linear in world
  position; hence `screen(W) − screen(F) = isoYawed(W,yaw) − isoYawed(F,yaw) = isoYawed(W−F, yaw)`. Every point's screen offset from F is a pure rotation of `W−F` — i.e. the scene rotates **about F** at F's true depth, which is exactly the point-of-interest pivot #1921 asks for, including `z > 0` content. ✓
- At `yaw == 0`, `isoYawed(F,0) == iso(F)`, so `offset == cameraIso` —
  byte-identical to the ORIGIN/cardinal fast path. ✓

### Sign relationship vs the legacy CAMERA_CENTER path

Sonnet noted the legacy path carries the opposite drift sign (the #1926
defect). I confirm and add the precise mechanism: the legacy branch returns
`offset = isoYawed(P, yaw)` where `P = isoPixelToPos3D(cameraIso, 0)` and
`iso(P) == cameraIso`. Correct pinning would need
`offset = cameraIso − isoYawed(P,yaw) + iso(P) = 2·cameraIso − isoYawed(P,yaw)`,
so the legacy branch adds the drift instead of cancelling it — confirmed
latent #1352 defect, correctly tracked in #1926, **not introduced here**.

Critically, the legacy path is behavior-preserved by this PR: the only change
is hoisting `const float visualYaw = IRPrefab::Camera::getYaw()` and renaming —
the returned expression is unchanged, so the no-focus path is byte-identical to
the pre-PR code. ✓

### `applyShotCameraState` extraction — one behavior delta, verified safe

The GUI-test cycler previously set zoom/pan/yaw/cull but never touched pivot
focus; the shared helper now calls `clearRotationPivotFocus()` on GUI-test
shots too (default `hasPivotFocus_ == false`). Since the GUI-test path never
sets a focus, `m_hasRotationPivotFocus` is already false and the clear is a
no-op — byte-identical, and strictly more correct (no cross-shot focus bleed).
Confirmed the focused→unfocused transition resets the pivot so a focused
regression shot can't contaminate the next. ✓

### State / ownership

`m_rotationPivotFocus` / `m_hasRotationPivotFocus` are value members,
default-initialized; the `g_pivotFocusShots` / `g_pivotFocusShotLabels` pair
is `reserve(n)`-then-fill, so the `label.data()` pointers stored in each shot
are stable for process lifetime — matches the `g_cullValidateShots` idiom. ✓
Demo scene is isolated behind an early return in `initEntities()`. ✓

### Nits (non-blocking)

- Concur with Sonnet's `kCropsPivotPillar {1130,460,300,520}` crop nit — tuned
  for the macOS HiDPI 2x framebuffer; the Linux/1x smoke agent will need to
  retune. Already documented in the code comment and the test plan, so the
  smoke labels carry the right signal. Not a merge gate.

No blockers. The byte-identical default guarantee holds, the focus math is
sound, and the legacy sign defect is documented rather than introduced.

jakildev:
Addressed fleet:has-nits:

- `kCropsPivotPillar` comment updated to note that the coords are 2x HiDPI-tuned (macOS, pillar pins at ~(1280,720) of 2560×1440) and provide a concrete 1x starting estimate (~{490, 100, 300, 520}, center≈(640,360) of 1280×720) for the Linux smoke agent to use as a baseline when re-shooting. Calculation: pivot focus pins to screen center; on 1x that's (640, 360), so x = 640−150 = 490, y = 360−260 = 100.

Comment-only change; no logic modified. — worker-2

Issue #1884, opened 2026-06-17
render: unify depth/clipping across render types (voxel/scatter/detached/floor)


jakildev:
**Model:** opus
**Part of epic:** #1881
**Plan file:** `~/.fleet/plans/issue-1881.md` (full epic plan + this child's section)
**Blocked by:** #1883 (shares the framebuffer gather/scatter surface — serialized) + #1910 (depth-readback probe — the prerequisite diagnostic tool; build it FIRST)
**Gated on:** PR #1880 (harness) on master.

> **Sequencing (2026-06-19):** root-cause findings posted as a comment below (two confirmed depth crossings A+B). Do NOT start the fix without #1910 (the depth-readback probe) — shader-output depth approximations gave contradictory data this session. Order: #1910 (probe) → then this fix, after #1883/PR #1907 lands.

## Scope
Voxels disappear or appear behind things; reported as overall canvas-bounds + distance mismatch between rendering types. Least-diagnosed of the three — likely an **investigation spike** first.

## Approach
Audit the depth ranges/units each render type writes to the shared framebuffer depth and unify them so cross-type compositing occludes correctly:
- per-axis scatter (`scatterCompositeDepthKey`, yawed iso depth);
- single-canvas gather (`pos3DtoDistance`, `encodeDepthWithFace` ×4);
- detached canvas (framebuffer-unit depth, #1872 / #1624 world-placed);
- floor / other canvases.

If the mismatch is not reproducible from a single audit, **reframe as an investigation spike** (put the literal phrase "investigation spike" in the issue) and produce a repro + a unit table of what each type writes before prescribing the fix — do not guess.

## Files (audit start)
`system_entity_canvas_to_framebuffer.hpp`; `f_trixel_to_framebuffer.glsl` (+ metal) depth normalize; `v_peraxis_scatter.glsl` composite depth key; `ir_render_types.hpp` depth-encoding constants.

## Acceptance criteria
- A multi-render-type scene (voxel cube + a detached canvas + floor) composites **correct occlusion at all yaws** — no voxel vanishing, no wrong-order draw.
- A documented **unit table** of what each render type writes to framebuffer depth.
- Holds on **both** backends.

## Dependencies
Blocked by #1883 + #1910 (probe). Terminal child of #1881.

## Gotchas
[...]

Issue #1881, opened 2026-06-17
render: rotated-voxel correctness under camera Z-yaw (epic)


jakildev:
## Context

A solid voxel cube does not stay a solid cube as the camera Z-yaws. Three distinct, independently-diagnosed defects live in the rotated-voxel render path. PR #1880 landed the deterministic validation substrate that localizes and gates all three:

- `IRPerfGrid --mode dense --yaw-ramp` — rotates a solid 64³ cube slowly through the full camera-yaw circle (lighting stays converged frame-to-frame, so the metrics reflect geometry/coverage, not a settling artifact).
- `scripts/dev/perf-grid-rotate-sweep` — build → ramp → per-frame score with `render-coverage-metric` (interior-hole ratio) + `render-silhouette-metric` (perimeter²/area raggedness).

**Baseline the harness establishes (origin/master, Metal):**

| pose | coverage | hole_ratio | perim_ratio |
|---|---|---|---|
| 0° (cardinal) | 0.994 | 0.006 | 242 |
| per-axis (10–80°,…) | 0.95–0.99 | <0.05 | 50–2000 |
| **90° / 180° / 270° (cardinals)** | **0.61–0.71** | **0.29–0.39** | **~140000** |

## The three defects (children)

1. **[primary] Cardinal-gather coverage loss at the non-start cardinals (π/2, π, 3π/2).** The single-canvas trixel gather drops ~⅓ of the cube to pure background at every cardinal except the start one. The per-axis (residual) path is fine.
2. **Per-axis face seams / slivers / stray lines** on zoomed small rotated cubes (the #1878 dilation is anisotropic + discontinuous; plus a scatter/resolve rounding split).
3. **Depth / clipping unification across render types** — voxels disappear / occlude wrong due to depth-unit mismatch between the per-axis scatter, single-canvas gather, detached-canvas (#1872), and floor.

## Closing criteria (whole epic)

`bash scripts/dev/perf-grid-rotate-sweep build dense 60` shows, on **both** Metal (macOS) and OpenGL (Linux):
- every cardinal **and** every face-flip at coverage ≥ 0.99 and perimeter_ratio within ~2× of the 0° baseline (~250, not 100k+);
- no regression at yaw 0 (byte-identical fast path preserved);
- a multi-render-type scene composites correct occlusion at all yaws.

[...]

PR #1935, opened 2026-06-21
render: composite-depth readback probe for #1884 debugging (#1910)

### LINKED ISSUE #1910: render: GPU depth-readback probe for composite depth-test debugging (unblocks #1884)

jakildev:
## Summary
A debug-only GPU→CPU readback that logs the **real composite depth-test value** at a requested framebuffer pixel, decoded to shared trixel-distance units. Tooling prerequisite that unblocks #1884 (epic #1881).

Every render path — cardinal gather (`f_trixel_to_framebuffer`), per-axis scatter (`f_per_axis_scatter`), and the detached-canvas composite (`ENTITY_CANVAS_TO_FRAMEBUFFER`) — writes its winning `gl_FragDepth` into the **one** framebuffer depth attachment under the `GL_LESS` test. So a single readback of that attachment captures the true composite winner regardless of which path produced the pixel — that's why this reads the attachment rather than approximating depth as a shader color output (which loses precision and can't be decoded per-path).

### What landed
- **engine/render primitive:** `PixelDataFormat::DEPTH_COMPONENT`, wired through the OpenGL (`glGetTextureSubImage` + `GL_DEPTH_COMPONENT`) and Metal (`getBytes`) `Texture2D::getSubImage2D` readback paths. Reuses the existing readback machinery + `device()->finish()` (Metal commit+wait per #1436) rather than hand-rolling a new path.
- **prefab helper:** `IRPrefab::DepthProbe::readbackCompositeDepth / logCompositeDepth` (`engine/prefabs/irreden/render/depth_probe.hpp`) — resolves the `mainFramebuffer` entity, flushes, reads the depth attachment, decodes via the exact inverse of `normalizeDistance`, logs `[depth-probe] pixel=(x,y) normDepth=… rawDist=…`.
- **demos:** `--depth-probe X,Y` in `canvas_stress` (#1884 bug A: detached-vs-floor) and `perf_grid` (#1884 bug B: per-axis stripe).
- doc pointer in `engine/render/CLAUDE.md` alongside the other depth diagnostics.

### Design notes / deliberate deviations
- **API placement.** The plan named `IRRender::readbackCompositeDepth`, but the high-level probe resolves the `mainFramebuffer` *prefab entity* — reaching a prefab entity from `engine/render/` is a layering violation, and `engine/render/CLAUDE.md` forbids feature API on `IRRender::`. So the generic depth-texture readback is the `engine/render` primitive (`PixelDataFormat::DEPTH_COMPONENT`) and the framebuffer-resolving helper is a prefab-scoped Pattern-B namespace (`IRPrefab::DepthProbe::`), mirroring `fog_of_war.hpp`.
- **Coordinates** are main-framebuffer **texture** space (resolution+buffer, e.g. 642×722), top-left origin — not window/screenshot pixels (`FRAMEBUFFER_TO_SCREEN` blits under camera pan/zoom).
- **Stretch deferred.** Per-canvas R32I distance decode (per-path side-by-side) is deferred: #1640's scratch-buffer indirection makes a direct R32I CPU `getBytes` return the clear value on Metal — it would need a resolve pass. The framebuffer **depth** attachment works because it's a real depth-test buffer, not `imageAtomicMin` scratch.

### Why no screenshots
Pure readback — **no shader or pipeline-output change**. The probe system is registered only when `--depth-probe` is set, so a flagless run is byte-identical; even with the flag on it only reads, never writes render state. Evidence is the probe log, not a visual diff (render CLAUDE.md "provably no runtime effect" exception). The OpenGL path is compile/run-validated by Linux cross-host smoke (macOS does not compile the GL backend).

## Test plan
- [x] Builds on macOS/Metal (`IRCanvasStress`, `IRPerfGrid`).
- [x] Metal `getBytes` on the `Depth32Float_Stencil8` attachment returns sane decoded depth (no validation error).
- [x] Distinguishes surfaces from empty: `canvas_stress` center pixel → `rawDist=58`; corners → `rawDist=65535` (the `kTrixelDistanceMaxDistance` clear sentinel, `normDepth=1.0`). Decode verified exact: (−40+65535)/131070 = 0.49969.
- [x] `perf_grid --depth-probe` reads the grid surface (`rawDist=−429.5`).
- [ ] OpenGL/Linux cross-host smoke (`glGetTextureSubImage` + `GL_DEPTH_COMPONENT` depth path).

Closes #1910

### DIFF
--- a/creations/demos/canvas_stress/main.cpp
+++ b/creations/demos/canvas_stress/main.cpp
@@ -52,12 +52,14 @@
 // Prefab helpers
 #include <irreden/render/camera.hpp>
 #include <irreden/render/camera_controls.hpp>
+#include <irreden/render/depth_probe.hpp>
 #include <irreden/render/entity_canvas.hpp>
 
 // Command suites
 #include <irreden/common/command_suite_capture.hpp>
 
 #include <cstdint>
+#include <cstdio>
 #include <cstring>
 #include <cstdlib>
 #include <string>
@@ -149,6 +151,15 @@ struct CanvasStressSettings {
     // deferred to the structural-gate work (T-2/T-3). NONE leaves the live
     // pipeline untouched, so a flagless run is byte-identical.
     IRRender::DebugOverlayMode debugOverlay_ = IRRender::DebugOverlayMode::NONE;
+    // `--depth-probe X,Y` (#1910): each frame, read back and log the real
+    // composite depth-test value at main-framebuffer texture pixel (X,Y) —
+    // top-left origin, in framebuffer-texture space (the size logged at canvas
+    // creation), not window/screenshot pixels. Off by default; when off no probe
+    // system is registered, so a flagless run is byte-identical. Used to
+    // root-cause the #1884 detached-vs-floor depth crossing in this demo (the
+    // floating canary cube clips behind the SDF floor at high zoom).
+    bool depthProbeSet_ = false;
+    ivec2 depthProbePixel_{0, 0};
     // readConfig() runs AFTER parseArgs() (it needs IREngine::init), so a
     // config `auto_rotate` would otherwise clobber an explicit
     // --auto-rotate / --no-auto-rotate flag. Latch CLI intent so config only
@@ -697,6 +708,19 @@ void parseArgs(int argc, char **argv) {
         } else if (std::strcmp(argv[i], "--debug-overlay") == 0 && i + 1 < argc) {
             g_settings.debugOverlay_ = IRRender::debugOverlayModeFromString(argv[i + 1]);
             ++i;
+        } else if (std::strcmp(argv[i], "--depth-probe") == 0 && i + 1 < argc) {
+            int px = 0;
+            int py = 0;
+            if (std::sscanf(argv[i + 1], "%d,%d", &px, &py) == 2) {
+                g_settings.depthProbeSet_ = true;
+                g_settings.depthProbePixel_ = ivec2(px, py);
+            } else {
+                IR_LOG_WARN(
+                    "--depth-probe: expected X,Y (e.g. 960,540); ignoring '{}'",
+                    argv[i + 1]
+                );
+            }
+            ++i;
         }
     }
     if (g_settings.sweepYawCount_ > 0 || g_settings.sweepFramesCount_ > 0) {
@@ -827,6 +851,20 @@ void initSystems() {
     renderPipeline.push_back(IRSystem::createSystem<IRSystem::TRIXEL_TO_FRAMEBUFFER>());
     renderPipeline.push_back(IRSystem::createSystem<IRSystem::ENTITY_CANVAS_TO_FRAMEBUFFER>());
     renderPipeline.push_back(IRSystem::createSystem<IRSystem::FRAMEBUFFER_TO_SCREEN>());
+
+    // #1910 composite-depth probe — registered only with --depth-probe so a
+    // flagless run adds no system. Runs last (after the framebuffer composite is
+    // complete) and logs the depth-test winner at the requested pixel each frame.
+    if (g_settings.depthProbeSet_) {
+        const ivec2 probePixel = g_settings.depthProbePixel_;
+        renderPipeline.push_back(
+            IRSystem::createSystem<C_Camera>(
+                "DepthProbe",
+                [](C_Camera &) {},
+                [probePixel]() { IRPrefab::DepthProbe::logCompositeDepth(probePixel); }
+            )
+        );
+    }
 
     if (g_autoWarmupFrames > 0) {
         int settleFrames = 60;
--- a/creations/demos/perf_grid/main.cpp
+++ b/creations/demos/perf_grid/main.cpp
@@ -34,6 +34,7 @@
 #include <irreden/render/systems/system_build_distance_hiz.hpp>
 #include <irreden/render/systems/system_build_light_occlusion_grid.hpp>
 #include <irreden/render/camera_controls.hpp>
+#include <irreden/render/depth_probe.hpp>
 #include <irreden/render/systems/system_compute_light_volume.hpp>
 #include <irreden/render/systems/system_compute_sun_shadow.hpp>
 #include <irreden/render/systems/system_compute_voxel_ao.hpp>
@@ -56,6 +57,7 @@
 #include <irreden/common/command_suite_capture.hpp>
 
 #include <algorithm>
+#include <cstdio>
 #include <cstring>
 #include <cstdlib>
 #include <numbers>
@@ -128,6 +130,13 @@ struct CliOverrides {
     // Accepted and recorded for manifest/cell-ID purposes; ignored until T-221.
     int workerThreads_ = 0;
     std::string configPreset_; // path from --config-preset, empty if absent
+    // `--depth-probe X,Y` (#1910): per-frame composite-depth readback + log at
+    // main-framebuffer texture pixel (X,Y), top-left origin (framebuffer-texture
+    // space, not window/screenshot pixels). Off by default → no probe system
+    // registered → flagless run byte-identical. Used to root-cause the #1884
+    // per-axis Y-over-X face-stripe depth crossing this demo exercises.
+    bool depthProbeSet_ = false;
+    ivec2 depthProbePixel_{0, 0};
 };
 
 constexpr IRVideo::AutoScreenshotShot kShots[] = {
@@ -517,6 +526,19 @@ void parseArgs(int argc, char **argv) {
                 g_cliOverrides.workerThreads_ = wt;
             }
             ++i;
+        } else if (std::strcmp(argv[i], "--depth-probe") == 0 && i + 1 < argc) {
+            int px = 0;
+            int py = 0;
+            if (std::sscanf(argv[i + 1], "%d,%d", &px, &py) == 2) {
+                g_cliOverrides.depthProbeSet_ = true;
+                g_cliOverrides.depthProbePixel_ = ivec2(px, py);
+            } else {
+                IR_LOG_WARN(
+                    "--depth-probe: expected X,Y (e.g. 960,540); ignoring '{}'",
+                    argv[i + 1]
+                );
+            }
+            ++i;
         }
     }
 }
@@ -841,6 +863,20 @@ void initSystems() {
             IRSystem::createSystem<IRSystem::FRAMEBUFFER_TO_SCREEN>(),
         }
     );
+
+    // #1910 composite-depth probe — registered only with --depth-probe so a
+    // flagless run adds no system. Runs after the framebuffer composite and logs
+    // the depth-test winner at the requested pixel each frame.
+    if (g_cliOverrides.depthProbeSet_) {
+        const ivec2 probePixel = g_cliOverrides.depthProbePixel_;
+        renderPipeline.push_back(
+            IRSystem::createSystem<C_Name>(
+                "DepthProbe",
+                [](C_Name &) {},
+                [probePixel]() { IRPrefab::DepthProbe::logCompositeDepth(probePixel); }
+            )
+        );
+    }
 
     if (g_autoProfileFrames > 0) {
         IRSystem::SystemId autoProfileId = IRSystem::createSystem<C_Name>(
--- a/creations/demos/perf_grid/main.cpp
+++ b/creations/demos/perf_grid/main.cpp
@@ -14,6 +14,7 @@
 #include <irreden/common/components/component_modifiers.hpp>
 #include <irreden/common/components/component_name.hpp>
 #include <irreden/common/components/component_local_transform.hpp>
+#include <irreden/render/components/component_camera.hpp>
 #include <irreden/render/components/component_canvas_ao_texture.hpp>
 #include <irreden/render/components/component_canvas_fog_of_war.hpp>
 #include <irreden/render/components/component_canvas_light_volume.hpp>
@@ -870,9 +871,9 @@ void initSystems() {
     if (g_cliOverrides.depthProbeSet_) {
         const ivec2 probePixel = g_cliOverrides.depthProbePixel_;
         renderPipeline.push_back(
-            IRSystem::createSystem<C_Name>(
+            IRSystem::createSystem<C_Camera>(
                 "DepthProbe",
-                [](C_Name &) {},
+                [](C_Camera &) {},
                 [probePixel]() { IRPrefab::DepthProbe::logCompositeDepth(probePixel); }
             )
         );
--- a/engine/prefabs/irreden/render/components/component_trixel_framebuffer.hpp
+++ b/engine/prefabs/irreden/render/components/component_trixel_framebuffer.hpp
@@ -45,6 +45,11 @@ struct C_TrixelCanvasFramebuffer {
         framebuffer_.second->getTextureDepth().bind(bindingDepth);
     }
 
+    // The composite depth attachment, for CPU readback (the #1910 depth probe).
+    const Texture2D &getTextureDepth() const {
+        return framebuffer_.second->getTextureDepth();
+    }
+
     const ivec2 getResolution() const {
         return framebuffer_.second->getResolution();
     }
--- a/engine/prefabs/irreden/render/depth_probe.hpp
+++ b/engine/prefabs/irreden/render/depth_probe.hpp
@@ -1,0 +1,123 @@
+#ifndef DEPTH_PROBE_H
+#define DEPTH_PROBE_H
+
+#include <irreden/ir_constants.hpp>
+#include <irreden/ir_entity.hpp>
+#include <irreden/ir_platform.hpp>
+#include <irreden/ir_profile.hpp>
+
+#include <irreden/render/components/component_trixel_framebuffer.hpp>
+
+// GPU→CPU composite-depth probe (#1910). A debug-only readback that, for a
+// requested screen pixel, reads the REAL depth-test value stored in the main
+// framebuffer's depth attachment and decodes it to shared trixel-distance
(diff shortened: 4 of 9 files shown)


jakildev (commented):
## Review — render: composite-depth readback probe for #1884 debugging (#1910)

**Verdict:** approve

### ECS invariants

`readbackCompositeDepth` calls `IREntity::getComponent<C_TrixelCanvasFramebuffer>("mainFramebuffer")` — this is a named-singleton foreign-entity lookup, not a per-entity iteration lookup, so the ECS "no getComponent in tick" footgun rule does not apply. The call pattern matches the documented "Pattern B — prefab-scoped free-function namespace" (mirrors `fog_of_war.hpp`). ✓

### Ownership / lifetime

`const auto &framebuffer` holds a reference to the component returned by `getComponent`. The `Framebuffer` is heap-allocated via `IRRender::createResource<Framebuffer>` and owned by the component; it cannot be destroyed during the tick that calls `readbackCompositeDepth`. The `const Texture2D &` returned by `getTextureDepth()` is valid for the duration of the call. No dangling-reference risk. ✓

### Render pipeline

- **`PixelDataFormat::DEPTH_COMPONENT` (OpenGL path):** `opengl_types.hpp` maps it to `GL_DEPTH_COMPONENT`. `opengl_texture.cpp::glFormatComponents` adds `GL_DEPTH_COMPONENT` to the single-component case (fallthrough with `GL_RED` / `GL_RED_INTEGER`), returning 1. `readSubImage2D` computes `bufSize = 1 × 1 × 1 × sizeof(GL_FLOAT) = 4` — correct for a single-pixel depth read. ✓

- **`PixelDataFormat::DEPTH_COMPONENT` (Metal path):** `metal_texture.cpp::pixelSizeBytes` is `typeSize × components = 4 (FLOAT32) × 1 (DEPTH_COMPONENT) = 4` bytes. `readSubImage2D` passes `bytesPerRow = 4` to `getBytes`. Empirically validated by the author for 1×1 reads on `Depth32Float_Stencil8`. ✓ *(Opus to confirm format-portability below.)*

- **No CPU↔GPU struct changes, no std140 layout changes, no bind-point changes, no shader files.** The `DEPTH24_STENCIL8` texture format and its framebuffer binding are unchanged. ✓

- **No new `.glsl` or `.metal` files.** The probe is pure CPU-side readback; no shader counterpart needed. ✓

### Layering

API placement follows the documented split (engine/render/CLAUDE.md §"What belongs in engine/render/ vs prefabs/"): the generic depth-texture readback primitive (`PixelDataFormat::DEPTH_COMPONENT`) lives in `engine/render/`; the framebuffer-resolving probe (`IRPrefab::DepthProbe::`) lives in the prefab layer. Matches the `fog_of_war.hpp` precedent exactly. ✓

The new `C_TrixelCanvasFramebuffer::getTextureDepth()` accessor (tier-b: reads only the component's own fields) is within the allowed component method tier. ✓

### `device()->finish()` call position

The probe lambda is registered last in the render pipeline (`after FRAMEBUFFER_TO_SCREEN`), so `device()->finish()` is called after the frame's compositing is complete. The call matches the existing pattern for picking (`HoveredEntityIdBuffer` readback) and Metal commit+wait (per #1436). The PR body explicitly cites this precedent. See Opus-only note below.

### Math / coordinates

Row-flip logic (`IRPlatform::kIsOpenGL ? (resolution.y - 1 - px.y) : px.y`) is consistent with the `readDefaultFramebuffer` pattern (mirrors GL bottom-left vs Metal top-left origin). The decode formula `windowDepth × (kMax − kMin) + kMin` is the exact inverse of `normalizeDistance`, using the same `kTrixelDistance{Min,Max}Distance` constants as the shader. ✓

### Naming / style

`IRPrefab::DepthProbe::` namespace, `CompositeDepthSample` struct, `readbackCompositeDepth` / `logCompositeDepth` free functions — all follow the codebase naming conventions. Public struct members use trailing `_`. ✓

### Tests / build

- macOS/Metal: build + 1×1 readback validated by the author (correct `rawDist` for center and corner pixels, Metal validation layer clean). ✓
- OpenGL/Linux: explicitly deferred to cross-host smoke (test plan item unchecked). Appropriate for a macOS-authored PR. ✓

### Praise

- Clean debug-gating: the probe system is registered only when `--depth-probe` is set, so flagless runs add zero systems and are byte-identical. The implementation correctly makes the feature invisible to the normal render pipeline.
- The coordinate-space warning in the doc comment (framebuffer texture space, top-left origin, NOT window/screenshot pixels after pan/zoom) is exactly the right thing to call out for future callers.
- Correct decision to defer per-canvas R32I decode (#1640 Metal foreign-canvas empty-read gap explicitly acknowledged in the PR body).

### Nits

- `creations/demos/canvas_stress/main.cpp:864` uses `createSystem<C_Camera>` as the probe anchor; `creations/demos/perf_grid/main.cpp:872` uses `createSystem<C_Name>`. Both work (endTick fires even when the archetype is empty), but picking a consistent component (C_Camera is the natural frame-tick singleton) avoids reader confusion.

### Opus-only items flagged

Two items require Opus verification before merge:

1. **`device()->finish()` called from a system lambda mid-render-pipeline.** On Metal, `finish()` is `metalCommandBuffer()->commit() + waitUntilCompleted()`. The probe runs after `FRAMEBUFFER_TO_SCREEN`, but the frame loop may have additional bookkeeping (sprite pass, deferred buffer releases) after the render pipeline. Opus should confirm that committing the command buffer inside a pipeline system lambda is safe — specifically that no encoder is still open when `finish()` is called, and that `metalCommandBuffer()` returning non-null after the commit does not cause a double-commit on the same buffer.

2. **`getBytes` on `MTLPixelFormatDepth32Float_Stencil8` with `bytesPerRow = 4`.** The actual packed memory layout of a combined depth-stencil texture on Apple Silicon is not documented to be exactly 4 bytes per pixel (it may be 8 bytes interleaved depth+stencil). For a 1×1 read the first `bytesPerRow = 4` bytes happen to be the depth float (empirically confirmed), but this is implicit. Opus should confirm whether Metal's `getBytes` on a combined depth-stencil texture with stride=4 is portable across Apple GPU generations (M1/M2/M3/M4).

- [x] macOS/Metal: build + single-pixel readback validated (correct values)
- [ ] Linux/OpenGL: `fleet-build --target IRCanvasStress && fleet-run IRCanvasStress --depth-probe 320,360` — `glGetTextureSubImage` with `GL_DEPTH_COMPONENT` on a `GL_DEPTH24_STENCIL8` attachment (cross-host smoke)
- [ ] Linux/OpenGL: `fleet-build --target IRPerfGrid && fleet-run IRPerfGrid --depth-probe 320,360` (cross-host smoke)

Opus recheck required: `device()->finish()` inside a pipeline system lambda (Metal command-buffer ordering) + `getBytes` on `MTLPixelFormatDepth32Float_Stencil8` with stride=4 (format portability across Apple GPU generations).

jakildev (commented):
## Opus recheck — render: composite-depth readback probe for #1884 debugging (#1910)

**Verdict:** approve

Sonnet approved and escalated two Metal-specific items it could not confirm. I verified both against the actual backend implementation — **both resolve to safe**. The probe is debug-gated, default-off, and byte-identical when the flag is absent.

### Opus-only item 1 — `device()->finish()` inside a pipeline system lambda ✓ SAFE

Sonnet flagged that committing the command buffer from a system lambda mid-pipeline could leave an open encoder or cause a double-commit. On closer read I confirm this is safe, because it reuses an already-proven mechanism:

- **No open encoder at probe time.** Every Metal render/compute/blit op in `metal_render_impl.cpp` opens *and* `endEncoding()`s its own encoder within the same operation (lines 413, 448, 483, 523, 564, 603, 682, 691). Encoders are never left dangling between pipeline systems, so when the probe lambda (registered last, after `FRAMEBUFFER_TO_SCREEN`) calls `finish()`, no encoder is open. `commit()` is valid.
- **No double-commit.** `finish()` (lines 711-726) commits + `waitUntilCompleted()`, then *immediately installs a fresh command buffer* via `setMetalCommandBuffer(metalCommandQueue()->commandBuffer())`. `present()` at frame end (lines 307-322) then commits **that fresh buffer**, not the already-committed one.
- **Present ordering is intact.** This is byte-for-byte the same mid-frame commit→wait→fresh-buffer pattern the screenshot path (`readDefaultFramebuffer`, lines 355-366) already relies on, whose own comment spells out the contract: *"start a fresh command buffer so present() can still encode the drawable presentation on top."* The drawable's color texture was rendered in the first (committed) buffer; `presentDrawable` on the subsequent fresh buffer presents that already-rendered content. Frame display is unaffected.

`finish()` is the canonical wrapper for exactly this. No new ordering risk introduced.

### Opus-only item 2 — `getBytes` on `Depth32Float_Stencil8` with `bytesPerRow = 4` ✓ SAFE / portable

Sonnet flagged that a combined depth-stencil texture might pack 8 interleaved bytes/pixel, making stride=4 implicit. On closer read I disagree that there's a portability risk:

- **The depth texture is `StorageModeShared`** (`metal_texture.cpp:90`, set unconditionally for every texture), which is the prerequisite for CPU `getBytes`. The `DEPTH24_STENCIL8` engine format maps to `MTL::PixelFormatDepth32Float_Stencil8` (`metal_texture.cpp:24-25`).
- **Apple's combined depth/stencil formats are *planar*, not interleaved.** `getBytes` on a `Depth32Float_Stencil8` texture accesses the **depth plane**, which is float32 = exactly 4 bytes/texel. `pixelSizeBytes(DEPTH_COMPONENT, FLOAT32)` correctly returns `4 (FLOAT32) × 1 (component) = 4` (lines 30-55), so `bytesPerRow = width(1) × 4 = 4` matches the depth-plane stride. The 8-byte interleaved layout Sonnet worried about does not occur for these formats.
- **Format-defined, not GPU-defined.** The float32 depth-plane representation is fixed by the pixel format, so the stride is identical across M1–M4. The author's empirical validation (correct `rawDist` for center/corner, clean Metal validation layer) is consistent with this. Portable.

The `RenderTarget`-only usage flag (`defaultTextureUsage`, lines 59-60) does not block `getBytes` — readback is governed by storage mode (Shared ✓), not usage.

### Other invariants (spot-check beyond Sonnet's pass)

- **No CPU↔GPU struct / std140 / binding-point changes, no shader files** — pure CPU-side readback. ✓
- **`getResolutionPlusBuffer()` / `getTextureDepth()`** exist on both the component and `Framebuffer` and are used correctly for the bounds check and the read. The new `getTextureDepth()` accessor on `C_TrixelCanvasFramebuffer` is tier-b (reads its own field) — allowed. ✓
- **`getComponent<C_TrixelCanvasFramebuffer>("mainFramebuffer")`** is a named-singleton foreign lookup, not per-entity tick iteration — the footgun rule does not apply (Pattern B, mirrors `fog_of_war.hpp`). ✓
- **OpenGL readback path** (`glGetTextureSubImage` with `GL_DEPTH_COMPONENT` on a `GL_DEPTH24_STENCIL8` attachment) is wired (`opengl_types.hpp`, `opengl_texture.cpp::glFormatComponents`) but, as the test plan states, has only been validated on Metal. This is the right thing to surface to cross-host smoke — I'm tagging `fleet:needs-linux-smoke` + `fleet:needs-windows-smoke` so the GL path gets exercised before merge.

### Nit (not blocking)

Agreeing with Sonnet's nit and not re-raising it as a separate round: `canvas_stress` anchors the probe on `C_Camera` while `perf_grid` uses `C_Name`. Both work (`endTick` fires on an empty archetype), but a consistent anchor reads cleaner. Trivial; fold into a later touch, no re-review needed.

jakildev:
Addressed nit: anchored the `--depth-probe` system in perf_grid on `C_Camera` (matching canvas_stress; the natural frame-tick singleton), and added the missing `component_camera.hpp` include. IRPerfGrid builds clean. The probe only uses its endTick lambda, so behavior is unchanged and flagless runs remain byte-identical. `fleet:approved` and `fleet:needs-opus-recheck` left as-is — this change doesn't touch the depth-probe readback / Metal command-buffer code under recheck. — worker

jakildev:
Cleared stale `fleet:has-nits`: the opus recheck (approve) re-applied it but its only nit — the `C_Camera` probe-anchor consistency — is the same one already addressed in commit `a4ae7d51` (the recheck reviewed the pre-amend diff and noted "fold into a later touch, no re-review needed"; it's folded into this PR). Both Metal items confirmed safe. PR sits at `fleet:approved` + cross-host smoke (linux/windows). — worker

Issue #1932, opened 2026-06-21
Planning-flow redesign: collapse plan-doc PR into the impl PR (tracking PRs 2–5)


jakildev:
Tracking issue for the remaining steps of the planning-flow redesign. **No labels — human-driven** (this is a delicate, ordered change touching live fleet scripts and gated self-config; do not auto-queue).

## Context
The redesign collapses the separate "plan-doc" PR into the implementation PR: the `## Plan` issue comment becomes the canonical, host-independent plan; the queue gate keys on that comment (not a committed/merged file); the implementer commits `.fleet/plans/issue-<N>.md` as the first commit of the code PR. Decoupling is preserved (opus plans via the comment; a cheaper worker implements later). A real plan-review step (`fleet:plan-review`) and a `human:no-plan` opt-out are added.

PR 1/5 (shared protocol + label catalog) is open: #1931.

## Remaining PRs (in order — each depends on #1931's contracts)
- [ ] **PR 2 — ingest gate.** `scripts/fleet/fleet-queue-ingest`: change the planning gate (~`:321-357`, `_plan_exists`) to pass when a `## Plan` comment exists on the issue OR it is opted out (`human:no-plan` / `[no-plan]` tag / `investigation spike`). Stop requiring a committed/staged plan file. Add `fleet:plan-review` to the early-skip set (~`:307`).
- [ ] **PR 3 — fleet-claim simplification (re-scope of #1889).** Keep the `planning-claim` / `planning-release` label lock + TTL/heartbeat sweep; drop `plan_pr_matches_issue` and the open-plan-PR cross-check (replace with "a `## Plan` comment already exists"). Remove `plan_pr_matches_issue` from `scripts/fleet/fleet_branch_match.py`. Retire `_PLAN_DOC_TITLE` in `scripts/fleet/fleet_scope_shipped.py` (+ its tests). Rewrite committed `.fleet/plans/issue-1889.md` to match.
- [ ] **PR 4 — file-epic.** `docs/agents/skills/file-epic.md`: drop the step-6.5 docs PR. Umbrella plan = umbrella issue comment; each child plan = `## Plan` comment on the child + first commit of that child's impl PR. (Check `epic-steward-protocol.md:157` "ONE docs-only PR" for the ledger/plan-file writes too.)
[...]

PR #1931, opened 2026-06-21
docs/fleet: collapse the plan-doc PR into the implementation PR

jakildev:
## Summary
PR **1/5** of the planning-flow redesign: collapse the separate plan-doc PR into the implementation PR.

Today a `fleet:needs-plan` task produces **two** merges — a docs-only plan-doc PR that commits `.fleet/plans/issue-<N>.md` to master, then the code PR. This redesign makes the **`## Plan` issue comment** the canonical, host-independent plan; the queue gate keys on that comment (not a committed/merged file); and the implementer commits the plan file as the **first commit of the code PR** — so plan + code land in one merge. Decoupling is preserved (opus plans via the comment, a cheaper worker implements later).

This PR lands the **shared protocol + label catalog** only:
- `PLANNING-PROTOCOL.md` — `## Plan` comment canonical; planner swaps `fleet:needs-plan` → `fleet:plan-review`; a real plan-review step (approve → queueable, reject → back to needs-plan); `human:no-plan` opt-out for simple ad-hoc issues; no plan-doc PR.
- `architect-protocol.md` — planner posts the comment + plan-review; design-blocked resume puts direction in the PR comment (plan file already on the branch); the engine **design-doc** docs-first PR pattern is preserved and clearly distinguished from task plans.
- `fleet-labels` + `fleet-labels-reference.md` + `fleet-state-machine.json` — add `fleet:plan-review` and `human:no-plan`, plus `plan-propose` / `plan-approve` / `plan-reject` transitions.
- `FLEET.md` — lifecycle + design-escalation narrative + a planning-gate bullet.

## Follow-ups (subsequent PRs)
- PR 2: `fleet-queue-ingest` — gate on the `## Plan` comment + opt-out; skip `fleet:plan-review`.
- PR 3: `fleet-claim` simplification — re-scope of #1889 (keep the planning-claim lock, drop `plan_pr_matches_issue` + the plan-PR cross-check); retire `_PLAN_DOC_TITLE`.
- PR 4: `file-epic` — drop the step-6.5 docs PR.
- PR 5: gated role/skill edits (human-applied).

#1889 is parked `fleet:needs-human` pending the re-scope.

## Deploy note
The two new labels exist only in the catalog so far — run `scripts/fleet/fleet-labels` (the sync, not `--check`) to create them on GitHub when this lands.

## Test plan
- [x] `fleet-labels --check` passes (53 labels, catalog ↔ JSON in sync)
[...]

PR #1951, opened 2026-06-22
fleet-labels: shorten 17 over-100-char catalog descriptions + guard length in --check

jakildev:
## Summary
GitHub's label-description API caps at **100 characters**; 17 entries in the `fleet-labels` catalog exceeded it. The sync loop lossily truncates over-long text (mid-word + ellipsis), so the on-GitHub descriptions silently diverged from the catalog — and `fleet:plan-review` / `human:no-plan` (added by #1931) never reached GitHub at all until they were created by hand this session.

## Changes
- **Shorten all 17 over-long descriptions to clean ≤100-char text** — no truncation needed. `fleet:plan-review` and `human:no-plan` match the exact strings now live on GitHub, so catalog ↔ GitHub agree.
- **Length guard in `fleet-labels --check`** — it now fails loudly (CI / pre-commit) on any catalog description >100 chars, instead of letting the runtime truncation paper over it. The runtime trim stays as a safety net.
- Update the stale sync-loop comment (`--check compares names only` → it also flags length now).

## Verification
- `bash -n` clean.
- `fleet-labels --check` → **OK — catalog and fleet-state-machine.json agree (53 labels), all descriptions ≤100 chars.**
- Names unchanged (only descriptions trimmed), so no state-machine / reference-doc drift.

Found while running the #1932 deploy step (creating `fleet:plan-review` / `human:no-plan`).

### DIFF
--- a/scripts/fleet/fleet-labels
+++ b/scripts/fleet/fleet-labels
@@ -110,7 +110,7 @@ LABELS=(
     "fleet:blocked|c5def5|Queued task whose Blocked-by predecessor is still open; not plainly claimable. Set/cleared by ingest"
     "fleet:needs-info|fbca04|Needs more info from the author before triage"
     "fleet:needs-plan|fbca04|Needs a plan (## Plan issue comment) before a worker can pick it up"
-    "fleet:plan-review|fbca04|Plan comment posted; reviewer vets before queueing; ingest skips it (sound→remove, else needs-plan)"
+    "fleet:plan-review|fbca04|Plan comment posted; reviewer vets before queueing; ingest skips it (sound->remove, else needs-plan)"
     "fleet:in-progress|c5def5|A fleet worker has claimed this task"
     "fleet:wip|c5def5|Fleet worker is mid-task on the corresponding PR"
     "fleet:stalled|fbca04|A fleet:wip PR has been idle 7+ days; human decides to merge / close / re-arm"
--- a/scripts/fleet/fleet-labels
+++ b/scripts/fleet/fleet-labels
@@ -98,19 +98,19 @@ fi
 
 LABELS=(
     "fleet:scope-shipped|fbca04|Issue scope already landed under a different T-NNN; queue-manager skipped ingest; human should close"
-    "fleet:needs-human|fbca04|Fleet can't complete this autonomously (e.g. a gated self-config edit); a human must act. Parks the issue out of the queue while KEEPING human:approved; clears when the human resolves it (ingest re-queues) or closes the issue"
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — fleet-labels: shorten 17 over-long descriptions + guard length in --check

**Verdict:** approve

Clean, well-scoped fix. The 17 shortened descriptions are all ≤100 Unicode characters, the `--check` guard logic is correct, the updated comment is accurate, and the safety-net trim in `ensure_label_for_repo` is preserved. One nit on locale portability of `${#}` is worth cleaning up.

### Checklist

- **ECS / render / ownership / math / serialization** — not applicable (pure fleet bash script, no C++).
- **Fleet scripts** — names unchanged (only descriptions shortened); no inline diagrams or `--help` banners reference old names. `fleet-labels-reference.md` uses separate full-prose descriptions; no update needed. ✓
- **Logic of the length guard** — `${#_desc} > 100` check runs over `LABELS[@]` (populated at load time, before `--check` branches); separate error-reporting paths for catalog/state-machine drift and overlong descriptions; `exit 1` at bottom is reached only when the outer `all-empty` guard doesn't `exit 0`. ✓
- **Description lengths** — all 17 shortened entries are ≤100 Unicode characters. ✓ (Verification: see nit below for one that is exactly 99 chars but 101 UTF-8 bytes.)
- **Tests / build** — `bash -n` clean; author ran `fleet-labels --check` successfully on a UTF-8 locale host. ✓

### Nits

- `scripts/fleet/fleet-labels:113` — `fleet:plan-review` uses `→` (U+2192): 99 Unicode characters but 101 UTF-8 bytes. Bash's `${#var}` is locale-sensitive: in a `LC_ALL=C` (byte-counting) environment the new overlong guard fires a false positive (101 > 100) on this entry. The fleet's Linux/macOS hosts run UTF-8 locale so this won't reproduce in practice, but CI environments sometimes default to C locale. Replacing `→` with `->` keeps the count at 100 chars/bytes and removes the ambiguity.

- [ ] `fleet-labels --check` on both engine and game repos to confirm OK after any nit fix.

Opus recheck not required.

jakildev:
Addressed has-nits feedback:
- `scripts/fleet/fleet-labels:113` — replaced `→` (U+2192, 99 chars/101 UTF-8 bytes) with ASCII `->` in the `fleet:plan-review` description, making it 100 chars / 100 bytes. This removes the `${#var}` byte-vs-char locale ambiguity flagged in the nit: the new overlong guard no longer false-positives under `LC_ALL=C`.

Verified `fleet-labels --check` passes in **both** UTF-8 and C locales (53 labels, all descriptions ≤100 chars). `fleet:approved` retained. — worker

PR #1952, opened 2026-06-22
build: green & warning-free macOS clean build

jakildev:
## Summary
- **audio:** disable miniaudio's unused `extras/nodes` example dylibs via `MINIAUDIO_NO_EXTRA_NODES`. They link a handful of `ma_*` core symbols without providing them, which the macOS arm64 linker rejects (Linux resolves undefined symbols at load) — the only build-breaking errors in a clean macOS/Metal build. We consume miniaudio as a single header, so none of its CMake targets are needed.
- **profile:** disable easy_profiler's bundled sample/reader executables via `EASY_PROFILER_NO_SAMPLES`. They injected a stale `${CMAKE_SOURCE_DIR}/../bin` link-search-path the linker kept warning about; we link only the core lib.
- **common:** make the header-only `IrredenEngineCommon` an `INTERFACE` library and drop the empty `ir_constants.cpp`. The `STATIC` build emitted an empty-archive `ranlib` warning. Matches the module's `CLAUDE.md` ("header-only, no src/").
- **video:** migrate `chooseAudioSampleFormat` off the deprecated `AVCodec::sample_fmts` to `avcodec_get_supported_config`, version-guarded (`LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(61, 13, 100)`) so older FFmpeg on the Linux/Windows hosts keeps using the field. Behavior is otherwise unchanged.
- **lua codegen:** stop the `IF` emitter double-wrapping conditions. `emitExpr` already parenthesizes `BINARY_OP`/`UNARY_OP`, so generated `if ((x == y))` tripped `-Wparentheses-equality`; it now emits `if (x == y)`.

## Test plan
- [x] Clean macOS/Metal build of all engine libraries, demos, tests, and tools (`ir-build --clean-first -- -k`) — previously exit 2 with 5 link errors and 4 warning classes; now builds error- and warning-free on Apple Clang.
- [x] Regenerated lua codegen header emits single-paren conditions; `IrredenEngineVideo` rebuilds clean with no deprecation warning.
- [ ] Linux/Windows (OpenGL) build stays green — the CMake dep-option flags are platform-neutral and only drop example/sample targets we never link; older FFmpeg keeps the `#else` field path.

## Notes for reviewer
- The two dependency-option flags only suppress example/sample targets that are never linked into any engine target, so there is no functional change on any platform.
- `video_recorder.cpp` is intentionally **not** whole-file reformatted — local clang-format (22.x) disagrees with the fleet's; only the logical hunk changed. A fleet `format-check` pass may want a different wrap on the multi-line call.

### DIFF
--- a/cmake/lua_codegen/system_dsl.cpp
+++ b/cmake/lua_codegen/system_dsl.cpp
@@ -1601,9 +1601,19 @@ struct Emitter {
             case StmtKind::IF: {
                 for (size_t i = 0; i < s.ifBranches_.size(); ++i) {
                     writeIndent();
-                    if (i == 0) out_ << "if ("; else out_ << "} else if (";
-                    emitExpr(*s.ifBranches_[i].cond_);
-                    out_ << ") {\n";
+                    out_ << (i == 0 ? "if " : "} else if ");
+                    // emitExpr already parenthesizes BINARY_OP / UNARY_OP, so
+                    // wrapping those again here emits "((x == y))", which trips
+                    // -Wparentheses-equality. Let the self-parenthesizing kinds
+                    // supply the condition's grouping; wrap only the atoms that
+                    // emit no parens of their own.
+                    const Expr &cond = *s.ifBranches_[i].cond_;
+                    const bool selfParen = cond.kind_ == ExprKind::BINARY_OP ||
+                                           cond.kind_ == ExprKind::UNARY_OP;
+                    if (!selfParen) out_ << "(";
+                    emitExpr(cond);
+                    if (!selfParen) out_ << ")";
+                    out_ << " {\n";
                     ++indent_;
                     auto savedSymbols = symbols_;
                     for (const auto &bs : s.ifBranches_[i].body_) emitStmt(*bs);
--- a/engine/audio/CMakeLists.txt
+++ b/engine/audio/CMakeLists.txt
@@ -109,7 +109,7 @@ FetchContent_Declare(
 # as dylibs that reference the ma_* core symbols without linking them; that
 # links on Linux (undefined symbols resolved at load) but fails on macOS
 # arm64 (the linker requires every symbol resolved). Skip building them.
-set(MINIAUDIO_NO_EXTRA_NODES ON)
+set(MINIAUDIO_NO_EXTRA_NODES ON CACHE BOOL "" FORCE)
 FetchContent_MakeAvailable(miniaudio)
 target_include_directories(IrredenEngineAudio PRIVATE ${miniaudio_SOURCE_DIR})
 
--- a/engine/audio/CMakeLists.txt
+++ b/engine/audio/CMakeLists.txt
@@ -103,6 +103,13 @@ FetchContent_Declare(
     GIT_TAG 0.11.25
     UPDATE_DISCONNECTED TRUE
 )
+# We consume miniaudio as a single header (MINIAUDIO_IMPLEMENTATION lives in
+# src/audio_playback.cpp), so none of its CMake targets are linked. Its
+# extras/nodes/* example libraries (ma_reverb_node, ma_vocoder_node, …) build
+# as dylibs that reference the ma_* core symbols without linking them; that
+# links on Linux (undefined symbols resolved at load) but fails on macOS
+# arm64 (the linker requires every symbol resolved). Skip building them.
+set(MINIAUDIO_NO_EXTRA_NODES ON)
 FetchContent_MakeAvailable(miniaudio)
 target_include_directories(IrredenEngineAudio PRIVATE ${miniaudio_SOURCE_DIR})
 
--- a/engine/common/CMakeLists.txt
+++ b/engine/common/CMakeLists.txt
@@ -1,14 +1,16 @@
-add_library(IrredenEngineCommon STATIC
-    src/ir_constants.cpp
-)
+# Header-only module (constants + compile-time platform detection); see this
+# directory's CLAUDE.md. With no real translation unit, a STATIC build emitted
+# an empty archive ("ranlib: … table of contents is empty"). INTERFACE is the
+# honest shape: consumers get the include dir and IRMath transitively.
+add_library(IrredenEngineCommon INTERFACE)
 
 target_link_libraries(
-    IrredenEngineCommon PUBLIC
+    IrredenEngineCommon INTERFACE
     IrredenEngineMath
 )
 
 target_include_directories(
     IrredenEngineCommon
-    PUBLIC
+    INTERFACE
     ${CMAKE_CURRENT_SOURCE_DIR}/include
 )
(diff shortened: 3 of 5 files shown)


jakildev (commented):
## Review — build: green & warning-free macOS clean build

**Verdict:** approve

Five targeted fixes for macOS clean-build errors and warnings. All changes are correct and well-scoped; one style nit on the miniaudio flag.

### Checklist

- **ECS / render / lighting / ownership / math / serialization** — not applicable (CMake build wiring + one C API migration + a codegen emitter fix).
- **Render pipeline / shader CPU↔GPU sync** — not touched. ✓
- **Naming / style** — no new identifiers. ✓
- **Tests / build** — PR body confirms all targets build error- and warning-free on macOS/Metal. The Linux/Windows path is not yet exercised (see test plan). ✓

### Nit analysis per change

**lua codegen (`system_dsl.cpp:1601–1616`):** Correct. `BINARY_OP` emits `(lhs op rhs)` and `UNARY_OP` emits `(-(x))` / `(!(x))` — both already self-parenthesize, so the outer `if (` ... `)` from the old code produced `if ((x == y))`, which Clang flags with `-Wparentheses-equality`. The new `selfParen` check identifies exactly those two kinds and lets them supply their own grouping; all other ExprKinds (NAME_REF, literals, INTRINSIC_CALL, COLUMN_AT/GET_FIELD) emit no outer parens and are correctly wrapped. ✓

**audio (`CMakeLists.txt:112`):** `MINIAUDIO_NO_EXTRA_NODES` is correctly set before `FetchContent_MakeAvailable`; with CMake 3.28's CMP0077=NEW a parent-scope normal variable shadows `option()` declarations in the child, so the flag works. However, every other FetchContent dependency knob in this file uses `CACHE BOOL "" FORCE` (e.g. `RTAUDIO_BUILD_TESTING`, `RTAUDIO_BUILD_MODULES`, `EASY_PROFILER_NO_SAMPLES`, etc.). This one uses plain `set()` — consistent behavior, inconsistent style.

**profile (`CMakeLists.txt:97`):** `EASY_PROFILER_NO_SAMPLES` uses `CACHE BOOL "" FORCE` — consistent with the rest of the file and correct (easy_profiler's CMakeLists uses a bare `if(NOT EASY_PROFILER_NO_SAMPLES)` with no `option()` declaration, so CACHE FORCE is the safe approach). ✓

**common (`CMakeLists.txt`):** Converting from `STATIC` (empty archive) to `INTERFACE` is the right fix — the module has no source files, matching the `CLAUDE.md` "header-only, no src/" spec. All consumer linkage keywords updated to `INTERFACE`. ✓

**video (`video_recorder.cpp:101–136`):** The `avcodec_get_supported_config` migration is correct. `nullptr` for `out_num_configs` is valid. The returned pointer is a static NONE-terminated array (same as `codec->sample_fmts` was), so no ownership change. The fallback `sampleFmts = codec->sample_fmts` in the `#else` branch keeps older FFmpeg hosts identical. ✓

### Nits

- `engine/audio/CMakeLists.txt:112` — `set(MINIAUDIO_NO_EXTRA_NODES ON)` should be `set(MINIAUDIO_NO_EXTRA_NODES ON CACHE BOOL "" FORCE)` for consistency with the ~10 other FetchContent knobs in this file. Functionally equivalent with CMP0077=NEW, but the inconsistency is a readability gotcha for future maintainers.

### Praise

- The version guard `LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(61, 13, 100)` is exactly the right granularity — the same NONE-terminated list abstraction works on both sides of the version split, so the rest of `chooseAudioSampleFormat` is unchanged.
- The `IrredenEngineCommon` → INTERFACE conversion matches the module's own `CLAUDE.md` spec ("header-only, no src/") and is the canonical CMake shape for a header-only target.

- [ ] Linux/OpenGL clean build to confirm the CMake flags are accepted (the PR notes this is still pending).
- [ ] Quick codegen smoke: emit a Lua `if` with a bare variable condition and confirm it generates single-paren `if (varName) {` (not `if ((varName))`).

Opus recheck not required.

jakildev:
Addressed the review nit: `engine/audio/CMakeLists.txt:112` — `set(MINIAUDIO_NO_EXTRA_NODES ON)` → `set(MINIAUDIO_NO_EXTRA_NODES ON CACHE BOOL "" FORCE)`, matching the ~10 sibling FetchContent knobs (RTAUDIO_*/RTMIDI_*) in the file. Functionally equivalent under CMP0077=NEW; pure style consistency. IRShapeDebug rebuilds clean on macOS/Metal. `fleet:approved` retained (cleanup doesn't invalidate the verdict). Linux/Windows smoke labels left in place — still pending. — worker

jakildev (commented):
## Opus recheck — build: green & warning-free macOS clean build (#1952)

Picked this up on the core-invariant rule (`engine/video/src/video_recorder.cpp`), not an explicit escalation — Sonnet approved and marked "Opus recheck not required." On a focused read of the one behavioral core-path change, **I confirm the approve.**

### The FFmpeg C-API migration (the Opus-only concern)

Sonnet flagged the `avcodec_get_supported_config` migration as correct; on closer read I agree, and the details that matter for a C-API swap all check out:

- **Call signature** — `avcodec_get_supported_config(nullptr /*avctx*/, codec, AV_CODEC_CONFIG_SAMPLE_FORMAT, 0 /*flags*/, reinterpret_cast<const void **>(&sampleFmts), nullptr /*out_num_configs*/)` matches the FFmpeg 7.1 prototype. `nullptr` for `out_num_configs` is explicitly permitted, and the `reinterpret_cast<const void **>` of an `const AVSampleFormat **` is the sanctioned call pattern (the API hands back a typed config array through a `void**`).
- **Version guard granularity** — `LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(61, 13, 100)` is exactly the cutover where the function was introduced and `AVCodec::sample_fmts` was deprecated; the `#else` branch keeps the older Linux/Windows fleet hosts on the field, byte-for-byte identical to the prior code path.
- **Error / null semantics preserved** — the new code returns `AV_SAMPLE_FMT_NONE` on a `< 0` return *and* on a null `sampleFmts`. The pre-existing contract already treated `sample_fmts == nullptr` as "no usable info → NONE", so the caller's fallback behavior is unchanged. No ownership change: the returned array is the same static NONE-terminated list the field used to expose, so the `for (...; *fmt != AV_SAMPLE_FMT_NONE; ++fmt)` walk is still valid and nothing is freed.

The other five changes are build/codegen wiring Sonnet already covered (lua `-Wparentheses-equality` fix, the `INTERFACE` conversion of the header-only common module, the CMake FetchContent knobs). The post-review commit (`MINIAUDIO_NO_EXTRA_NODES … CACHE BOOL "" FORCE`) is exactly the nit Sonnet requested, applied verbatim — functionally equivalent under CMP0077=NEW, now style-consistent with the sibling knobs.

No ECS/render/lighting/lifetime invariants touched. Cross-platform smoke labels (`fleet:needs-linux-smoke` / `fleet:needs-windows-smoke`) correctly left pending — the FFmpeg `#else` branch and the CMake flags still need a non-macOS build to exercise, but that's the smoke-worker's gate, not a merge blocker for this verdict.

**Verdict: approve.** Confirms Sonnet's pass.

Issue #1944, opened 2026-06-22
test: canvas_stress render-verify references are red on master (6/6 fail) — refresh or investigate


jakildev:
**Blocked by:** #1953 (revert #1942 viewport-center default — fixes the canvas_stress jitter regression that is a material part of this drift; refreshing before it merges would enshrine the regression and re-red the gate). Added during planning — see the `## Plan` comment.

## Summary

`canvas_stress`'s committed `macos-debug` render-verify references are **stale vs current `origin/master`** — all 6 manifest shots fail on a clean checkout (no local edits):

| shot | match% | max_delta |
|---|---|---|
| so3_smooth_sweep | 97.87% | 230 |
| so3_offsnap_disc | 97.90% | 230 |
| so3_wide_parity | 97.95% | 230 |
| so3_grid_cross | 98.06% | 230 |
| so3_offsnap_wide | 97.96% | 230 |
| revoxelize_solids | 98.01% | 230 |

This is the `canvas_stress` sibling of **#1929** (same symptom on `shape_debug`).

## Repro

```
python3 scripts/render-verify.py --target IRCanvasStress --no-build   # on clean origin/master
```

## Diagnosis

The references were last committed at **#1719/#1723 (Jun 11, commit 97e74307)**. ~30 render PRs have merged since (e.g. #1873 voxel-AO de-speckle, #1872 world-placed detached-canvas depth scaling, #1874 sun-shadow ambient, #1877/#1878/#1882/#1883 per-axis scatter, #1926/#1927 camera Z-yaw pivot). The drift is **cumulative intended render improvement**, not a single regression — the diff is distributed across the whole scene (every orbit entity + the central cluster shifts slightly), consistent with the accumulated voxel/scatter/lighting changes.

## Action

Refresh the `macos-debug` (and `linux-debug` via the cross-host flow) references against current master, OR investigate if any single PR's drift is unintended. Same call as #1929.

## Note (links #1374)

When refreshing, consider also wiring the new `--only compare` shots from #1374 (`compare_yaw0`, `compare_yaw_q`) into the manifest gate — they are currently appended-but-ungated because (a) the gate is red, and (b) inserting them contiguously after `revoxelize_solids` would shift the `shadow_overlay_floor` / `revox_coverage` capture indices that `test/baselines/STRUCTURAL-BASELINES.md` documents by filename. A refresh pass is the natural place to renumber + gate them together.

jakildev:
[...]

PR #1942, opened 2026-06-21
render: camera Z-yaw — exact-screen-center default + cursor pivot mode (#1926)

jakildev:
## Summary
Two camera Z-yaw rotation modes, plus the empirical proof of *why* rotation looked wrong, plus the camera/iso pivot-math cleanup — all folded into one PR.

- **Mode 1 (default): rotate about the EXACT screen center.** The `CAMERA_CENTER` default now pivots about the true viewport-center point (`canvasSize/2 − trixelOriginOffsetZ1 − cameraIso`) instead of the legacy `isoPixelToPos3D(cameraIso, 0)` point (mirrored + ~1 trixel off true center).
- **Mode 2 (key-toggled): rotate about the cursor.** `Ctrl+Shift+middle-drag` pins the world point under the cursor (`setRotationPivotFocus(mouseWorldPos3DAtIsoDepth(0))`, captured at drag start, reverts on release). Plain `Ctrl+middle-drag` stays screen-center.

## Empirical proof — the math is correct; the demo was off-center
The reported "rotation isn't about screen center" was investigated to ground truth. `shape_debug --pivot-focus-demo` pins an explicit focus on a pillar at world **(8, −8, 10)** (off-origin, z>0 — the hard case) and sweeps yaw:

| yaw 0 | rotated |
|---|---|
| centered | still centered |

The pillar holds the **exact** screen center (measured centroid **1279.5, 720.5** vs center 1280, 720) while the markers orbit it. A broken pivot drifts hundreds of px (#1926 measured 1024px). **The engine iso/pivot math is correct** — the default scene only appears to "swing" because its content is laid out off-center (x=0 shapes at center, x=16/32… toward a corner). The pivot model + this proof live in `docs/design/camera-yaw-pivot.md`.

## The reusable math (the "refine + clean up" ask)
- **`IRMath::cameraYawPivotOffset(cameraIso, F, yaw)`** — the single documented drift-cancel pivot offset: pins `F` at a constant screen position across yaw, returns `cameraIso` at yaw 0. The formula was previously inlined twice in `getEffectiveCameraIso`; both the screen-center default and the explicit-focus path now route through it.
- `glm::floor → IRMath::floor` in the picking path.

## Tests
- [x] New pivot-helper unit tests (pinning invariant across yaw; yaw-0 returns `cameraIso` exactly).
- [x] Full unit suite green — **1215 passed**, 1 skipped (GPU-on-Metal).
[...]

PR #1928, opened 2026-06-20
render: temporal-jitter rotation validation harness

jakildev:
## Summary
Acceptance-gate harness for **epic #1881** (rotated-voxel correctness under camera Z-yaw). The epic's existing solidity harness (`perf-grid-rotate-sweep`, #1882) scores per-frame *coverage* of one cube; this one scores *temporal stability* — does the rendered surface shimmer as the camera turns — across every SDF shape.

- **`scripts/render-jitter-metric.py`** — the discriminator is the **second temporal difference** (discrete Laplacian) of per-pixel luminance over a fine yaw sweep: smooth shading ramps → `d2 ≈ 0`; a crawling band oscillating across an integer boundary spikes `d2`. The first difference can't tell smooth motion from crawl, which is why the metric keys on curvature. Pure stdlib, streaming 3-frame window, interior-only mask, reuses `render_metric_util`.
- **`scripts/tests/test_render_jitter_metric.py`** — 10 synthetic-frame tests proving the core claim: a gradient sliding 1px/frame (high `d1`) scores ~0 jitter, a parity-flipping ring scores 50×+, and the gate exit-codes hold.
- **`scripts/dev/shape-rotate-jitter-sweep`** — one command sweeps all 8 SDF shapes × a 360-step (1°) yaw turn, SDF + voxel-pool render paths, with a documented `MAX_JITTER=4.0` gate.
- **`creations/demos/shape_debug`** — `--spin-shape` / `--spin-shape-voxel` spawn a single origin-centred shape with a centred ROI crop per shot (so the metric decodes small PNGs, not the full retina framebuffer). The default fixture scene is **byte-identical** — canvas-lighting setup was extracted to a shared helper with no behaviour change.
- **`docs/design/jitter-validation-harness.md`** — the measured macOS/Metal master baseline + gate rationale + render-path coverage scope.

### Master baseline (macOS/Metal, zoom 8, 360 steps, SDF path)
| shape | jitter | verdict |  | shape | jitter | verdict |
|---|---|---|---|---|---|---|
| torus | 1.07 | PASS |  | box | 6.74 | **FAIL** |
| cone | 1.13 | PASS |  | wedge | 6.26 | **FAIL** |
| cylinder | 1.17 | PASS |  | ellipsoid | 7.54 | **FAIL** |
| sphere | 1.20 | PASS |  | **curved_panel** | **11.38** | **FAIL** |

[...]

PR #1953, opened 2026-06-22
render: revert #1942 viewport-center default — fixes canvas_stress rotation jitter (#1944)

jakildev:
## Summary
Fixes the **new per-frame position jitter of entities during rotation in IRCanvasStress** — a regression from #1942 (my camera Z-yaw pivot change).

## Root cause (empirically isolated)
#1942's "exact viewport-center" default pivot fed the `visibleIsoViewport` `viewCenter` (`canvasSize/2 − trixelOriginOffsetZ1 − cameraIso`) as the rotation focus. That point is **~1 trixel off the content center**:
- The GRID/world content uses the **effective** camera offset (`getEffectiveCameraIso()` → the new pivot).
- The DETACHED entity-canvas composite places entities with the **raw** `getCameraPosition2DIso()`, *not* the effective offset (`system_entity_canvas_to_framebuffer.hpp:84,123`).

Un-panned, the legacy default made the effective offset `0` (= raw) so they stayed aligned. #1942's `(1,1)` term made the effective offset **non-zero and yaw-varying** while detached stayed at raw `0` → the world content **drifts relative to the detached entities as the camera yaws** = the rotation jitter (canvas_stress auto-rotates, so every frame is affected).

**Isolation:** master(#1942) vs this revert differs **1.3–3.4%** on every auto-rotating canvas_stress shot; at yaw 0 both are byte-identical (fast path).

## Fix
Revert the default to the legacy #1352 point `isoPixelToPos3D(cameraIso, 0)` — un-panned the world origin, the true content/screen center where world and detached coincide. `effective == raw` again → detached and GRID stay aligned through rotation. The `cameraYawPivotOffset` helper and the cursor pivot mode (#1942) are unchanged.

The `(1,1)` "exactness" was a coordinate-convention artifact (the viewport-rect center ≠ the content center), so this loses nothing real — the default still rotates about screen center, correctly.

## Follow-ups
- A *true* viewport-center pivot (panned-camera exactness) would require the detached composite to consume `getEffectiveCameraIso()` too — separate, delicate change (filed).
- canvas_stress `so3_*` reference re-baseline (stale from several merges) — #1944.
- Extend the #1928 jitter harness to canvas_stress mixed detached/GRID rotation (filed) — it currently covers only shape_debug SDF shapes.

Refs #1944.


jakildev (commented):
## Review — render: revert #1942 viewport-center default — fixes canvas_stress rotation jitter (#1944)

**Verdict:** approve (Opus recheck required — touches engine/render/ camera yaw pivot math)

The revert is mathematically correct. I verified the new code against the pre-#1942 history (commit `f4171d92`) — it is byte-for-byte identical to the legacy implementation:

```cpp
// pre-#1942 (commit f4171d92) and this PR — same formula:
const vec3 cameraFocusWorld = IRMath::isoPixelToPos3D(cameraIso, 0.0f);
return IRMath::pos3DtoPos2DIsoYawed(cameraFocusWorld, IRPrefab::Camera::getYaw());
```

The root-cause analysis is correct and well-documented.

### Checklist

- **ECS invariants** — no ECS code touched. ✓
- **Ownership / lifetime** — no ownership changes; no manager pointers stored. ✓
- **Render pipeline / GPU struct sync** — no shaders changed, no CPU/GPU struct sync needed. `getEffectiveCameraIso()` is purely CPU-side camera math. ✓
- **Lighting** — not touched. ✓
- **Math / coordinates** — `IRMath::isoPixelToPos3D` and `IRMath::pos3DtoPos2DIsoYawed` used; no raw `glm::` or `std::` math. ✓
- **Serialization** — not touched. ✓
- **Naming / style** — `cameraFocusWorld` is descriptive and correctly scoped. The comment block is lengthy but appropriate: it encodes a non-obvious alignment invariant and an explicit "Do NOT" guard that prevented this exact bug in #1942, so a future author will understand why the formula must not be changed. ✓
- **Tests / build** — no unit tests exist for this render math; the empirical test (1.3–3.4% canvas_stress diff on rotating shots, byte-identical at yaw 0) is the appropriate regression check for this category of change. ✓

### Nits

- `engine/render/src/ir_render.cpp` / `docs/design/camera-yaw-pivot.md` — The PR body describes the canvas_stress regression evidence quantitatively ("1.3–3.4% diff on every auto-rotating canvas_stress shot") but does not attach a screenshot pair. `engine/render/CLAUDE.md` recommends screenshots for render changes that affect visual output. Worth attaching before merge — not a blocker but helps the Opus reviewer and future history readers.

### Praise

- The `camera-yaw-pivot.md` update is thorough: the updated "Screen center (default)" section now encodes the detached-vs-world alignment constraint and the explicit "Do not pivot about `visibleIsoViewport` viewCenter" warning, making the `(1,1)` regression from #1942 harder to re-introduce.
- The comment added in `ir_render.cpp` (ll. 58–73) is the right kind of comment — it explains WHY this formula must not be changed and cross-references the design doc. Good precedent.

- [ ] Run `canvas_stress` with auto-rotation on both Linux/GL and macOS/Metal and confirm no per-frame entity jitter at any yaw (especially non-cardinal).
- [ ] Verify byte-identical output at `visualYaw == 0` vs pre-PR on both backends.
- [ ] Run `shape_debug --pivot-focus-demo` and confirm the off-origin pillar at world (8,−8,10) still pins to screen center under a full yaw sweep (regression guard for the explicit-focus path, which is unchanged).

Opus recheck required: touches engine/render/ camera yaw pivot math (subtle rotation invariant; empirical validation described but no screenshot pair attached).

jakildev:
Addressed the review nit — **attached a before/after screenshot pair** (committed to `docs/pr-screenshots/claude/1944-canvas-stress-jitter/`, embedded below).

**Capture method:** `canvas_stress --no-auto-rotate --no-spin --auto-screenshot` (deterministic static scene — the camera-pivot math is the *only* variable), run once on `origin/master` (#1942 viewport-center pivot) and once on this branch (reverted `isoPixelToPos3D(cameraIso,0)` pivot).

**Result — surgical, as the revert intends:** 8 of the 10 base-suite shots are **byte-identical** before/after (every `yaw==0` shot — confirms the no-rotate fast-path stays identical to ORIGIN mode). Only the two non-cardinal-yaw shots differ:

| shot | camera yaw | match % | max Δ |
|---|---|---|---|
| `so3_offsnap_disc` | 45° | 99.91% | 182 |
| `so3_offsnap_wide` | 30° | 99.08% | 245 |

The diff images localize the change to the **central GRID / SDF-floor world content** (which consumes the effective offset), while the **detached orbit ring stays put** (it composites with the raw offset) — i.e. the world-vs-detached relative shift this revert removes. Frame-to-frame under auto-rotation that shift is the #1944 jitter; held static here it reads as a ~1-trixel offset.

<details>
<summary><b>so3_offsnap_disc — 45° camera yaw</b></summary>

| Before (#1942 master) | After (this revert) | Per-byte diff |
|--------|-------|------|
|  |  |  |

In the diff, the central GRID cubes + re-voxel column carry the colored deltas; almost the entire orbit ring is black (byte-identical).

</details>

<details>
<summary><b>so3_offsnap_wide — 30° camera yaw, wide zoom</b></summary>

| Before (#1942 master) | After (this revert) | Per-byte diff |
|--------|-------|------|
|  |  |  |

</details>

`fleet:has-nits` cleared; `fleet:approved` unaffected (cleanup doesn't invalidate it). Still carries `fleet:needs-opus-recheck` + cross-host smoke labels. — worker

jakildev (commented):
## Opus recheck — render: revert #1942 viewport-center default (#1944)

**Verdict:** approve

Sonnet flagged this for recheck on the camera-yaw pivot invariant ("subtle rotation invariant; empirical validation described but no screenshot pair attached") and asserted byte-for-byte identity with pre-#1942 `f4171d92`. On closer read I confirm the substance and tighten two of Sonnet's claims.

### Math invariant — confirmed, with a precision

Sonnet said the new code is "byte-for-byte identical to the legacy implementation." That is true **of the no-focus default branch specifically**, not of the whole function — and the distinction matters because it's what makes this a *safe* revert rather than a feature rollback:

- **Default branch ≡ legacy `f4171d92`.** The PR computes `cameraFocusWorld = isoPixelToPos3D(cameraIso, 0.0f)` and returns `pos3DtoPos2DIsoYawed(cameraFocusWorld, visualYaw)`. Since `visualYaw` is assigned `IRPrefab::Camera::getYaw()`, this is textually identical to the legacy line. ✓
- **Yaw-0 fast path stays byte-identical to ORIGIN mode.** `pos3DtoPos2DIsoYawed(W, 0) == pos3DtoPos2DIso(W)`, and the round-trip `pos3DtoPos2DIso(isoPixelToPos3D(cameraIso, 0)) == cameraIso` holds, so the no-rotate path returns `cameraIso` exactly. ✓
- **The #1921 explicit-focus + #1942 cursor-pivot path is retained.** The `hasRotationPivotFocus()` branch still routes through `IRMath::cameraYawPivotOffset`, so this PR reverts *only* the default (no-focus) pivot point — it does not undo the cursor pivot mode or the point-of-interest API that landed alongside #1942. That's the right scope.

### Architectural soundness — the trade-off is correct

The two defaults are genuinely different formulas, not just different focus points: #1942 used `cameraYawPivotOffset(cameraIso, F_viewCenter, yaw)` (drift-cancel that pins the geometric viewport center); legacy/this-PR re-projects the origin-based screen-center point. The revert accepts that the default pivot is ~1 trixel off the *true* screen center in exchange for keeping the effective-offset world content aligned with the raw-offset detached entities under yaw. That is the correct priority: the #1942 "exact viewport center" was a cosmetic nicety, while the misalignment it introduced is a real per-frame jitter regression that the canvas_stress detached canary catches. The root cause (detached composite places entities at the *raw* `getCameraPosition2DIso()`, not the effective offset) is diagnosed correctly, and the proper unification is left as a documented follow-up.

### No invariant concerns

- Pure CPU camera-offset math. No shaders, no CPU/GPU struct sync, no binding-point or dispatch changes — `pos3DtoPos2DIsoYawed` has a GLSL mirror, but only the *value* of the camera-offset uniform changes; the shader formula is untouched. ✓
- No ECS, ownership, lifetime, or allocation surface touched. ✓

### Resolved / nits

- **Sonnet's screenshot nit is resolved** — the PR now includes the before/after/diff pairs (`so3_offsnap_disc_45deg`, `so3_offsnap_wide_30deg`) under `docs/pr-screenshots/`. The diffs confirm the orbit ring is byte-identical (black) and only the effective-offset world content carries the delta, which is exactly the expected signature of this change.
- **Follow-up tracking (non-blocking).** The "a true-viewport-center pivot would need the detached path to consume the effective offset too (follow-up)" note in both the code comment and `camera-yaw-pivot.md` is good, but isn't tied to an issue number. Worth filing so the deferred unification doesn't get lost — not a merge gate.

The doc + comment encode the *why* and an explicit "Do NOT pivot about viewCenter" guard, which makes the #1942 regression hard to re-introduce. Good defensive precedent. Approving.

jakildev:
Cross-host smoke OK on macOS (Metal): fresh checkout build of IRShapeDebug clean, `--auto-screenshot 10` run exited 0 with no compile warnings/shader errors. Opus visual pass — inspected the four yaw-pivot shots (yaw0/45/90/180): shapes render intact (cubes/cylinders/cones/spheres/torus), correct iso projection, lighting + shadows present, scene pivots cleanly about center with no jitter, missing voxels, or color/parity artifacts. — worker

jakildev:
Follow-up commits pushed (5322c261 + evidence 5f3b4d9e) — this PR now does more than the revert:

**The #1944 revert alone left a latent pivot bug.** It restored the legacy #1352 screen-center *point* but also dropped the `cameraYawPivotOffset` drift-cancel wrapper, and that point is the viewport center *mirrored across the canvas origin*. Net effect: panning the camera off-origin and then yawing in `CAMERA_CENTER` (the engine default) swings the whole scene in an arc instead of rotating in place. canvas_stress never caught it because it is **un-panned** (every pivot form collapses to 0 there); shape_debug's panned pivot shots show it plainly.

**Fix (5322c261):**
1. `getEffectiveCameraIso()` CAMERA_CENTER default → pivot about the *exact* viewport center (`isoPixelToPos3D(viewCenterIso,0)`) through the drift-cancel `cameraYawPivotOffset` (matches the #1921 explicit-focus path + the `[pivot]` math invariant test).
2. The viewport-center focus is ~1 trixel off origin even un-panned, so `effective != raw` whenever yaw≠0 — the exact reason #1944 reverted. Completed the documented follow-up: `ENTITY_CANVAS_TO_FRAMEBUFFER` now consumes `getEffectiveCameraIso()` for placement, so detached + GRID share one pivot. Gather parity stays on the fixed world iso.

**Verified (macOS/Metal, see `docs/pr-screenshots/claude/1944-canvas-stress-jitter/`):**
- shape_debug pan16: yaw0 byte-identical (0.0000%); yaw180 now point-reflects through screen center (was a 15.9% off-frame swing).
- canvas_stress (un-panned): yaw0 byte-identical; yaw45 differs only **0.21%** (coherent pivot shift, detached+GRID together) vs the **1.3–3.4%** detached *drift* #1942 alone caused → the #1944 fix is **not** reintroduced.
- Full engine test suite green (1216 pass, 1 expected non-OpenGL skip).

PR title is now narrower than the change — suggest broadening to something like *"camera Z-yaw pivot: exact-center + detached on effective offset (#1944)"*.

Out of scope: the flat-top floor-clipped shapes in the same scene are Bug A from the #1884 investigation (#1958 sub-epic, blocked by #1957) — unrelated to the pivot.

Authored on macOS/Metal → needs a Linux/OpenGL smoke pass before merge (touches engine/render + engine/prefabs/irreden/render).

jakildev:
Per-axis jitter follow-up pushed (`2e7cd4da`). This addresses the per-frame jitter the pivot revert left behind — the one that survived on IRShapeDebug.

**Root cause.** The smooth-camera-Z-yaw per-axis composite (active at any non-cardinal camera yaw) split the camera offset into two terms at **mismatched screen scales**: an integer anchor `floor(cameraIso·subdivisionScale)` (canvas-cell placement) plus a sub-pixel term scaled to the canvas game-px grid (`zoomEff·2`), **not** to one anchor cell. One anchor cell maps to `K = fbRes·zoomEff/axisSize` framebuffer px, so at every effectiveIso-integer crossing the anchor jumped K px while the sub-pixel term reset by `zoomEff·2` px — an uncompensated ~K-px snap each crossing. That's the sawtooth; it grows with zoom (effSub) and the density-scaled anchor (vestigial since #1458 made the per-axis store base-resolution) multiplied the crossing rate.

**Fix** (continuous decomposition, mirroring the smooth cardinal path):
1. Whole-iso **base anchor** `floor(cameraIso)` — applied consistently across the per-axis store (stage 1 + 2), every per-axis lighting/shadow recovery (`perAxisCellToWorld3D`), the sun-shadow resolve (anchor only; its re-projection scale stays density-scaled), and the framebuffer scatter, so all agree bit-for-bit.
2. Sub-cell smooth term = `screenPxPerCell·fract(cameraIso)` (continuous), replacing calcModelMatrix's wrong-scale offset → anchor + smooth = K·cameraIso, continuous. `anchorPanExcess` deleted.

Cardinal single-canvas path untouched; the base anchor is inlined (not added to `ir_iso_common`) to avoid FP-drifting the cardinal fast path.

**Verified (macOS/Metal; centroid-residual plot `docs/pr-screenshots/.../peraxis_jitter_before_after_plot.png`):**
- **Pan @ yaw 45** (voxel box): before std 12.4 / max 22px / 11 reversals → after **+1.0px/frame, std 0.00, 0 reversals**.
- **Rotation** (voxel cylinder, Z-yaw-invariant probe, fine yaw sweep): before max 26px / 7–8 reversals → after **≤0.7px/frame, 0 reversals**.
- **After-move static**: yaw-0 byte-identical run-to-run (no static jitter).
- Full shape_debug scene + canvas_stress (GRID + orbit) render correctly, lighting/shadows intact. 1216/1216 tests pass.
- Cardinal yaw-0 delta: 8px / max-Δ-8 (0.0002%) FP-scheduling shift on lit edges — the irreducible, sub-perceptual cost of editing per-axis shader blocks that share a compilation unit with cardinal-active shaders.

Added `--pan-sweep` / `--yaw-sweep` to shape_debug as the jitter regression harness. GLSL + Metal kept in parity.

Note: does **not** touch the "flat-top" floor-clipped shapes — that's Bug A / the #1958 depth sub-epic, unrelated to this jitter.

Issue #1960, opened 2026-06-22
render: per-trixel priority + depth-priority demos (entity vs trixel clip/interpenetrate) [D/E #1884]


jakildev:
**Model:** opus · **Part of:** #1884 (depth-unification sub-epic) — the priority **feature + validation** slice.
**Blocked by:** #1958 (the priority band).

## Scope
Add **per-trixel priority** (max'd with per-entity priority in the encoding, #1884 design) and the two depth-priority **demos** that prove the semantics:
- **Per-trixel priority source** — decide + wire where a face's priority comes from (authored / voxel metadata); the composite encodes `priorityBand = max(perEntity, perTrixel)`.
- **Demo 1 — clip-vs-swap:** N units orbiting + the camera rotating, so depths cross. Watch units swap in-front/behind correctly, and confirm **no flicker within a bracket** (quadrant-stability from B working).
- **Demo 2 — intentional interpenetration:** two units with per-trixel priority deliberately animating *inside* each other (effect) vs the same two at per-entity priority staying solid (no clip).

## Acceptance
- Per-entity ("don't clip through walls") and per-trixel ("animate inside each other") both demonstrated; committed render-verify baselines for both demos (they become the priority feature's regression harness).

## Refs
#1884 design; #1950.

jakildev:
Planning gate: no plan file found for this issue — neither committed `.fleet/plans/issue-1960.md` nor planner-host `~/.fleet/plans/issue-1960.md` — and it is not an explicit investigation spike. Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

jakildev:
## Plan: render: per-trixel priority + depth-priority demos (entity vs trixel clip/interpenetrate) [D/E #1884]

- **Issue:** #1960
- **Model:** opus
- **Date:** 2026-06-22
- **Blocked by:** #1958 (the priority band + per-entity `depthPriority_`). This task **extends** #1958's encoding to per-trixel granularity and ships the validation demos; it is unimplementable until #1958 lands, since the `priorityBand·BAND` encode, the `BAND`/`kDepthMaxPriorityBand` constants, and the `C_EntityCanvas::depthPriority_` field it maxes against are all introduced there. Build on a master where #1958 (and its own blocker #1957) are merged.

### Scope

Two halves of the #1884 priority slice:

[...]

Issue #1959, opened 2026-06-22
render: per-axis cardinal-180 geometric tiebreak (fix Bug B) [C/E #1884]


jakildev:
**Model:** opus · **Part of:** #1884 (depth-unification sub-epic).
**Blocked by:** #1958 (the cardinal-stable key).

## Scope
Fix **Bug B** — the per-axis doubled face-stripe at cardinal-180 (#1950 Finding 3). `scatterCompositeDepthKey = yawedSum·4 + slot` collapses at yaw-180 (`c=−1, s=0` → `−x−y+z`): the X-face and Y-face exact cells along the shared front edge tie on `yawedSum`, and `+slot` always picks X (slot 0) regardless of true depth → the ~20px `YXYX` stripe.

Replace the `slot` tiebreak with a **geometric** (true per-face depth) arbiter — fold in the non-degenerate projection component the key drops at 180° so X/Y faces get distinct depths by real geometry.

## Validation
- yaw-180 doubled stripe gone (`IRPerfGrid --mode dense --grid-size 64 --yaw-ramp` step 18); no regression at 90/270 or off-cardinals; rides the per-axis path #1883 stabilized → rotated-solidity (#1882) gate.
- Note: a cardinal-stable key (from #1958) shouldn't degenerate mid-bracket, which may simplify or absorb this — coordinate with B.

## Refs
#1884; #1950 (Finding 3); #1883.

jakildev:
Planning gate: no plan file found for this issue — neither committed `.fleet/plans/issue-1959.md` nor planner-host `~/.fleet/plans/issue-1959.md` — and it is not an explicit investigation spike. Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

jakildev:
## Plan: render: per-axis cardinal-180 geometric tiebreak (fix Bug B) [C/E #1884]

- **Issue:** #1959
- **Model:** opus
- **Date:** 2026-06-21
- **Blocked by:** #1958 (the cardinal-stable key + shared encoding this key must co-sort against)

### Scope

[...]

PR #1484, opened 2026-06-02
render-verify: re-bless shape_debug macos-debug refs + fix manifest drift (#1478)

jakildev:
## Summary
- Re-bless the `shape_debug` **macos-debug** reference set and fix `manifest.json` drift (6 → 21 full-frame shots). The committed refs were blessed 2026-04-20 (#233) and had drifted ~6 weeks / dozens of render PRs out of date; the manifest still listed 6 shots while the demo's `kShots[]` table now cycles 21 full frames + 7 ROI crops (28 PNGs), so manifest `shot[6+]` mapped onto unrelated captures and every tracked shot failed on clean `master`.
- **Recovers the orphaned work from PR #1480.** #1480 did this exact re-bless but its commit never reached `master`: it merged into its stacked base (`claude/1472-shape-debug-crop-glob`) *after* that base (#1479) had already merged to `master`, so the child commit was stranded on a dead branch. This is the known merger stacked-base ordering issue (#1149). The manifest on `master` was still the stale 6-shot version (last touched `24a4ac62`), which is why #1478 stayed open.
- ROI crops stay inspection-only, excluded from the comparison set per the #1479 glob fix (already on `master`). `manifest.json` `notes` updated to document the full-frame/ROI split.

## [opus] correctness judgment
- Re-validated against **current** `origin/master` (HEAD `b722ed39`, which includes #1453 / #1470 / #1471 / #1473): `render-verify.py --target IRShapeDebug` reports **all 21 shots PASS byte-identical** (match 100%, max_delta 0, PSNR inf) — deterministic under #1477's fixed-step UPDATE.
- Visually inspected representative captures across every category — base/zoom, LOD swap (cube → cone → sphere silhouette change), the finest-tier sphere close-up, cardinal yaw (90/180/270), inter-cardinal yaw (45°, expected squash/stretch), and pan+yaw pivot. All render coherent isometric voxel output with correct LOD, rotation, lighting, and cast shadows. No black frames, missing geometry, or clipping.

## Not in scope
- `linux-debug` / `windows-debug` reference sets are equally stale but can't be re-blessed from this macOS host (the GL backend produces pixel-different output) — they need a refresh on those hosts.

## Test plan
- [x] `render-verify.py --target IRShapeDebug` on macos-debug: all 21 shots PASS, byte-identical against current `origin/master`.
- [ ] Re-bless `linux-debug` / `windows-debug` on those hosts (follow-up; out of scope here).

Closes #1478

[...]

PR #1974, opened 2026-06-23
render: unified continuous-yaw composite depth + foreground partition (#1958)

### LINKED ISSUE #1958: render: unified quadrant-stable depth encoding + per-entity priority bands (revisit #1370, fix Bug A) [B/E #1884]
**Model:** opus · **Part of:** #1884 (depth-unification sub-epic) — **the core change**, touches shared `ir_render_types.hpp`.
**Blocked by:** #1957 (foundation A/E #1884 — detached depth-write must land first)

## Scope
Implement the unified depth encoding (see #1884 design):
```
enc = priorityBand · BAND  +  cardinalIsoDepth · 4  +  face
```
- **Quadrant-stable `cardinalIsoDepth`** — use the **cardinal (rasterYaw)** iso-depth, NOT the continuous `visualYaw`. This **replaces #1370's continuous-yawed smooth-depth** (`c_shapes_to_trixel.glsl` smooth `baseDepth = round(dvx+dvy+z)`, and the per-axis / voxel paths). A fragment's depth then depends only on its 90° bracket (quadrant-stable) — no per-frame depth change within a bracket.
- **Per-entity priority band** — a `depthPriority` field on `C_EntityCanvas`; the composite encodes `priorityBand` from it. Floor = background band, foreground solids = front band → **fixes Bug A** (floating solid clips behind floor). Reserve a few bands in the ±65535 range (BAND ≈ 21845; verify world-size headroom, static_assert it).

## Validation (critical — this revisits #1370)
- **rotated-solidity (#1882) + depth-probe (#1910):** confirm #1370's "low/back surface wins against geometry above it near ±45°" artifact does NOT return. If an un-tagged general-geometry case still mis-orders, keep a residual correction scoped to only that case.
- **Bug A:** the canary cube renders fully above the floor (depth-probe verified, `--only canary,floor`).
- Cardinal frames byte-identical; both backends.

## Refs
#1884 design; #1370; #1950 (Finding 2); `ir_render_types.hpp` (`kDepthEncodeShift`).

jakildev:
## Update 2 (2026-06-23b, high-zoom) — **subdivision-scale the per-axis scatter** (the "clips at high zoom" follow-up)

The continuous-metric consolidation (Update 1 below) fixed the rotating-regime desync, but clipping **returned at high zoom** — world voxel cubes clipped into the SDF floor as you zoomed in. That is a second, *orthogonal* defect: a subdivision-**scale** mismatch (pre-existing on master/branch; the metric fix only verified clean at zoom 1 where it's masked).

**Root cause.** The SDF floor + cardinal voxel gather encode depth **subdivided** (`worldDepth × effSub × 4`), but the per-axis voxel **scatter** store is **base-resolution** (#1458: `rawDist >> 10` = world units), so its composite key was ×1. At `effSub == 1` (zoom 1) they coincide; as `effSub → 16` (zoom in) the floor depth out-scaled the scattered voxels ~`effSub×`, so near floor cells (very negative, inflated) rendered in front of and **clipped the cubes' lower cells into the floor**.

**Fix.** Lift the scatter's composite iso-depth to the same subdivided magnitude (`× effSub`, carried in `effectiveSubdivisionsForHover.x`) in `v_peraxis_scatter.glsl` + `metal/peraxis_scatter.metal`. The #1458 frac bits keep the recovered corner sub-cell-exact, so the scale-up **preserves precision** (scaling the SDF *down* to base instead z-fought the floor — wrong direction, reverted). Only the iso-depth (×4) term scales; the slot tiebreak stays unscaled (comparable to the SDF face bits); the #1457/#1883 margin bias + gradients scale uniformly with it, so margin behavior is invariant.

**Cube–floor boundary @ zoom 8, yaw 45° — before vs after** (the lower cube cells were eaten by the floor; now full-length):

| before (base-scale scatter — clips) | after (subdivided scatter — fixed) |
|---|---|
| before | after |

**Validation (macOS / Metal).** `canvas_stress` static **and** spinning yaw grid — zoom **4 / 8 / 16** × yaw **0 / 26 / 45 / 60 / 80 / 115 / 160 / 214°**: nothing clips through the floor or disappears at any pose; the floor stays smooth (no z-fight); cardinal frames **byte-identical** (8/8 cardinal shots, img_diff drift 0). GL/Linux unverified (authored on Metal).

Also corrected a stale claim in the design doc's "remaining gap" note: per-axis scatter is allocated whenever `residualYaw != 0` (any non-cardinal yaw, **rest or moving**), so there is **no** "at-rest non-cardinal gather" regime to desync — the metric+scale are continuous+subdivided for all co-rendered world content at every non-cardinal yaw.

---

## Update (2026-06-23b) — Part A withdrawn: **consolidated onto one continuous-yaw depth metric**

Revisiting the on-device behavior surfaced that **Part A (quadrant-stable cardinal SDF depth) was the wrong direction** — it fixed the *at-rest* non-cardinal floor drift but **regressed the rotating regime**, which is what the depth clipping "jumps per quadrant / changes with rotation" report was seeing.

**Why Part A regressed rotation.** Part A snapped only the *SDF floor* depth to the cardinal bracket. But the **per-axis voxel scatter** key (`scatterCompositeDepthKey`) and the **detached-canvas composite** offset still sorted on the *continuous* `visualYaw` metric. So during camera rotation the cardinal-snapped floor desynced from the continuously-sliding voxels: the floor depth **jumped** at each ±45° quadrant boundary while voxels slid, and world voxels / detached solids **clipped into the floor between cardinals**. The original SDF comment even promised *"per-axis voxel scatter applies the identical transform so SDF + voxels stay co-sorted"* — the cardinal snap broke exactly that invariant.

**The fix — one shared metric for every composite depth writer.** The geometrically-correct sort key is the camera-forward distance at the *actual* `visualYaw`:

```
depth(p) = pos3DtoDistance(R_z(-visualYaw)·p) = x·(cos-sin) + y·(sin+cos) + z   // smaller = nearer
```

This is now the single shared helper **`yawedIsoDistance`** (`ir_iso_common.glsl` / `.metal`; CPU twin `IRMath::pos3DtoDistanceYawed`), called by **all three** world-content depth producers that previously each open-coded their own variant:

- **SDF smooth path** (`c_shapes_to_trixel`) — `baseDepth = roundHalfUp(yawedIsoDistance(worldSurface, visualYaw))` (replaces the cardinal-bracket `dvx/dvy`).
- **Per-axis voxel scatter** (`scatterCompositeDepthKey`) — unchanged behavior, now factored to call the helper.
- **Detached composite** (`system_entity_canvas_to_framebuffer`) — world-depth offset = `pos3DtoDistanceYawed(worldCell, visualYaw)` (replaces un-yawed `pos3DtoDistance`).

At a cardinal pose `yawedIsoDistance` collapses to un-yawed `x+y+z`, so the cardinal fast path is byte-identical. **Part B (the foreground near-plane partition) is unchanged.**

### Verification (macOS / Metal) — `canvas_stress --no-spin --no-auto-rotate --auto-screenshot`, branch vs this change
- **No regression:** **8/8 cardinal (yaw 0) shots byte-identical** (img_diff drift = 0). Only the **2 non-cardinal yaw shots** (yaw 45°, yaw 30°) change — exactly where the metric differs.
- **Bug fixed:** at yaw 45°, world cubes that **clipped behind the floor** on the branch now sort cleanly **on** it; the per-quadrant depth jump is gone (continuous depth is smooth across the boundary by construction).

| Floor + cubes @ yaw 45° — **before** (cardinal Part A) | **after** (continuous consolidation) |
|---|---|
| before | after |

(Left: purple/yellow cubes clip behind the floor and the cluster is jumbled with dark floor-clip patches. Right: all five solids sort cleanly as a row on the floor.)

### Remaining gap → #1959
The main-canvas **at-rest** gather still composites world voxels on the *stored* cardinal depth (kept cardinal because AO / sun-shadow / per-axis-resolve reconstruct world position from it via `isoPixelToPos3D`, which assumes `depth = x+y+z`). So a camera *resting* at a non-cardinal yaw still has the main-world voxels on the cardinal metric while the floor is continuous — the master-era at-rest drift, unchanged here. Closing it needs the gather to recompute continuous occlusion depth at the final `gl_FragDepth` (stored depth must stay cardinal for the reconstruction consumers) — the per-axis/gather charter of **#1959**. The auto-rotating demos exercise only the rotating regime, which this PR fixes.

**GL/Linux unverified (authored on Metal) — cross-host smoke pending.** Source-of-truth design doc `docs/design/depth-unification-1884-investigation.md` §Resolution updated to the consolidated model ("revised 2026-06-23b").

---

<details>
<summary>Original PR description (Part A / quadrant-stable — superseded by the update above)</summary>

Implements the #1884 sub-epic's **Bug A** fix (a floating detached solid clipping *behind* the floor at high zoom), per the architect's design-unblock ruling. Resumed from the `fleet:design-blocked` → `fleet:design-unblocked` escalation; the additive priority band is **superseded by a disjoint near-plane partition** (the additive band could not dominate unbounded world placement — the radius-200 GRID orbit exhausted it).

### Part A — quadrant-stable SDF iso-depth (retires #1370)
`c_shapes_to_trixel.glsl`/`.metal`: on the smooth-yaw path the depth metric derives its cos/sin from the **cardinal bracket** instead of continuous `visualYaw`, so the floor's stored depth is piecewise-constant per 90° quadrant and stops drifting under geometry above it near the ±45° bracket. On-screen placement is unchanged; only the stored depth. Cardinal frames byte-identical (the smooth branch is gated off at cardinals).

### Part B — two-tier disjoint near-plane depth partition
The most-negative `kDepthForegroundBandWidth` (16384) codes of the shared `[kMin, kMax]` depth range are **reserved** for foreground-priority detached solids. The gather (`f_trixel_to_framebuffer.glsl`/`.metal`) pins a foreground draw **into** the band and clamps world content **out** of it, keyed on a per-draw `depthPriorityMode` field (repurposed std140 pad — no offset/size change). **Invariant:** a foreground fragment is unconditionally nearer than any world fragment *independent of world extent* (dominance by partition membership, not by out-sizing the world depth spread). The world clamp is a **no-op** for all in-budget content, so the cardinal fast path is byte-identical; the unsatisfiable world-extent `static_assert` is replaced by a partition-layout assert on constants.

Two-tier only (per-trixel tiers #1960; 32-bit composite depth follow-up #1983). `C_EntityCanvas::depthPriority_` is the authoring knob (composite reads it directly, no per-tick getComponent); `canvas_stress` canary opts in; `depth_probe` logs a `tier/iso/face` split.

- **Bug A fixed.** `IRCanvasStress --only canary,floor --zoom 8`: the canary cube renders **fully in front** of the SDF floor. `--depth-probe-assert 321,210` → `result=PASS rawDist=-57391` (inside the foreground band `[-65535,-49151]`, normDepth 0.062 ≪ floor's ~0.49).
- **Cardinal byte-identity.** `shape_debug` **19/21 shots byte-identical** to `origin/master` (direct binary compare, not the stale references). Only the 2 **non-cardinal** yaw45 shots (`zoom4_yaw45_inter_cardinal`, `zoom4_pan16_yaw45_pivot`) change — Part A's intended #1370 fix.
- **Metal + GL shaders** declared in lockstep; the FrameData offset `static_assert`s compile (208 B unchanged). **GL/Linux unverified (authored on Metal) — cross-host smoke pending.**

| Bug A fixed (canary in front of floor, zoom 8) | Part A yaw45 change (img_diff overlay) |
|---|---|
| after | yaw45 |

### Notes for review
- `shape_debug` render-verify **references are pre-stale on master** (last re-blessed 2026-06-02 / #1484, predating ~8 render PRs — master itself is red, like canvas_stress #1944). A ref refresh is a separate task; the **2 yaw45 shots additionally** change here by design.
- Source-of-truth design doc `docs/design/depth-unification-1884-investigation.md` §Resolution + the plan `.fleet/plans/issue-1958.md` are updated to the partition model in this PR (architect-required).

Closes #1958

</details>


jakildev:
## NEEDS-DESIGN

De-risking the plan's flagged **band-headroom × subdivisions trap** before implementing parts B/C/E. **The band model as specced does not fit `canvas_stress`** — escalating per the plan's explicit instruction ("If it can't fit ≥2 bands at the demos' effSub cap, escalate `fleet:design-blocked` — do not silently clamp").

### Measured facts (file:line)
- **Max effective subdivision = 16** (hard clamp): `engine/render/src/render_manager.cpp:293,297` — `clamp(m_voxelRenderSubdivisions * max(1, zoomScale), 1, 16)`. So `effSub ≤ 16` on every path; a user can interactively zoom `canvas_stress` to this.
- **Iso depth = `worldDepth = x+y+z`**, subdivided iso-depth = `worldDepth · effSub`: `engine/math/include/irreden/ir_math.hpp:1333-1339` (`pos3DtoDistance`). Encoded `enc = isoDepth·4 + face`, normalized over `[-65535, 65535]`.
- **`canvas_stress` orbit ring** (`creations/demos/canvas_stress/main.cpp:1328-1367`): `kOrbitRadius = 200.0f`, 12 shapes, **modes include `RotationMode::GRID`** (band-0 *world* content) at `z = 0`. Worst-case `worldDepth ≈ 200 + 200 + 0 = 400`.
- **`shape_debug`**: max `worldDepth ≈ 161` (`main.cpp` shape grid x≤112, y span ~44, z~5).

### The constraint fails
Plan's non-overlap rule: `2·(4·maxSubdividedIso) + 3 < BAND`, architect target `BAND ≈ 65535/3 ≈ 21845` → requires `maxSubdividedIso < ~2730`.

| demo | maxWorldDepth | × effSub 16 = maxSubdividedIso | `2·4·iso+3` | vs BAND 21845 |
|---|---|---|---|---|
| shape_debug | ~161 | **2576** | 20609 | ✅ fits (headroom ~1.2k) |
| **canvas_stress** | ~400 | **6400** | **51203** | ❌ **2.3× over** |

### Root cause (why this is a design call, not a clamp)
The additive band (`enc -= band·BAND`) lives **inside** the shared `±65535` range. For a foreground solid to beat *all* band-0 world content at a shared pixel, `BAND` must exceed the **full world iso-depth spread** (`2·4·maxSubdividedIso`). But world placement is **unbounded** — the orbit at radius 200 (GRID = world content) proves entities go far past chunk bounds — so there is **no sound compile-time `maxSubdividedIso`** for the `static_assert`, and no fixed `BAND ≤ 65535/k` can dominate arbitrarily-deep background. Reducing the orbit radius or capping its subdivision to make the assert pass is exactly the "silently clamp / alter test content to fit" the plan forbids, and it wouldn't fix the general unboundedness.

1. **Bound or unbounded?** Should the priority band assume a documented **world-extent ceiling** (size `BAND` + `static_assert` against `kDepthMaxSubdividedIso`, and bring `canvas_stress`'s radius-200 orbit inside it as part of E)? If so, what ceiling — and is shrinking the orbit acceptable test-content drift?
2. **Or a disjoint partition instead of an additive band?** e.g. pin a foreground entity to a fixed **near-plane sub-range** (`enc = kForegroundBase + localDepth`, clamped) that is provably disjoint from the world iso-depth range, rather than a relative `−BAND` shift that lands back inside it. This survives unbounded world depth at the cost of intra-foreground depth resolution.
3. **Scope of band-0 content.** Is the orbit's GRID arm intended to participate in the priority contest at all, or can world content be assumed bounded for the demo while the engine-wide `static_assert` uses a conservative documented max?

### Done so far / handoff
- Part **A (quadrant-stable SDF depth)** is de-risked and independent of the band question — the GLSL/Metal edit is a localized swap of the smooth-yaw depth metric from `visualYaw` to the cardinal-bracket cos/sin at `c_shapes_to_trixel.glsl:936-938` / `.metal:1036-1038` (cardinal byte-identity holds trivially; the smooth branch never runs at exact cardinal). Held back so A and the deferred per-axis #1959 co-sort don't land mismatched while the band model is unsettled.
- Plan committed at `.fleet/plans/issue-1958.md`. Branch is the claim commit only; no encoding changes yet.

Recommend **option 2** (disjoint near-plane partition) as the only choice robust to unbounded world depth, but this is the architect's call.

— worker (opus)

jakildev:
## Architect direction

Good escalation — the band-headroom trap is real and you measured it correctly. Ruling: **adopt the disjoint near-plane partition (your Option 2)**, expressed as a minimal change to the approved formula rather than a new model. Foreground priority becomes a **position** in the depth range (a reserved *near* sub-range), not an additive shift through it.

### The model

```
default / world content (priority 0):   enc = cardinalIsoDepth·4 + face        // UNCHANGED
foreground-priority entity:             enc pinned into reserved nearest band, encoding its OWN local iso-depth
world content clamp:                    enc = max(cardinalIsoDepth·4 + face, kForegroundCeil + 1)
```

- More-negative `enc` = nearer (`normalizeDistance` + `GL_LESS`). Reserve the nearest slice — e.g. `[kMin, kForegroundCeil] = [−65535, −49152]`, width 16384 — exclusively for foreground priority. World content is **clamped** to stay out of it.
- **Invariant (the point):** a foreground-priority fragment is unconditionally nearer than any non-priority fragment, *independent of world extent* — dominance is by **partition membership**, not by out-sizing the world's depth spread. This is exactly what the additive `−BAND` shift could not guarantee.
- **Cardinal byte-identity preserved.** The world-content clamp is a **no-op** until `cardinalIsoDepth < −12288` (world extent ≫ radius 200; never hit by canvas_stress, shape_debug, or any current demo at the effSub-16 cap). So the cardinal fast path and all in-budget content keep their exact current `enc` — no test-content drift, orbit stays at r=200.
- **Graceful degradation, not a correctness break.** Beyond the documented ceiling, far world content *saturates* against the foreground boundary (loses depth resolution) instead of letting a background fragment win over a priority solid. Strictly better than the additive model, which broke foreground dominance the moment the world got deep.
- Within the foreground band, encode the entity's own local iso-depth (measured from its nearest face) so it **self-occludes** correctly. Two distinct foreground solids sort by local depth — acceptable; that is the priority semantic ("this is solid, don't clip / don't swap behind the floor"), not a world-depth contest.

### The `static_assert` flips from unsatisfiable to trivially-true

Drop the `kMaxSubdividedIso < BAND/8` world-extent assert (no sound compile-time bound exists, as you proved). Replace with a **partition-layout** assert on constants only:

```cpp
static_assert(kForegroundCeil < 0);                                  // reserved band is on the near side
static_assert(kMin <= kForegroundCeil && kForegroundCeil < kMax);    // band lies inside the range
// (+ per-trixel tier bounds when D #1958 lands)
```

### Generalization for D #1960 (per-trixel) — keep B minimal

The partition generalizes to N disjoint sub-ranges; `priority = max(entity, trixel)` selects the tier. **For B, implement only the two-tier split (world + one foreground tier).** D #1960 adds the per-trixel tier(s) on top — don't pre-build them here.

---

## On your two questions

**1. "Are we hitting this because the world isn't centered in the distance field?"** — No, and it's worth being precise about why. The encodable range is *already* symmetric about `enc = 0` (`kMin = −65535`, `kMax = +65535`, `ir_constants.hpp:47,51`), and the world origin (`iso = 0`) sits at field-center; canvas_stress content spreads ≈ ±25600 with ~40k of headroom on **each** side. So it is **not** an off-center / insufficient-buffer problem. We hit the wall because the *additive-shift* band must out-size **2× the world's depth spread** — it has to lift a far-placed foreground entity from the `+far` edge all the way past the `−near` edge of world content — *and* coexist in the same bounded range. That term scales with world extent, so the radius-200 GRID orbit exhausts it. Centering changes the *offset*, not the *spread*, so it can't fix it. The partition fixes it because priority costs **O(1)** range-membership instead of an **O(worldSpread)** shift.

**2. "Would going to 32-bit distances be the long-term fix + let us simplify?"** — Plausibly yes, for any *practical* world, and I've **filed it as #1983** (investigation spike, blocked-by this issue). The honest shape:
- The `±65535` ceiling is a **normalization convention** on the shared framebuffer depth, *not* a hardware limit — the per-canvas R32I distance texture is already 32-bit exact. So "32-bit distances" really means "widen the shared composite depth."
- **DEPTH32F** is ~free on desktop GL / Metal, but it's *float* — non-uniform spacing, only ~24 bits of *exact* integer codes (≈ D24's uniform 16.7M), so it isn't automatically more exact for the face/slot tie-break low bits. The truly-exact-32-bit path is a **manual integer composite** against an R32I attachment (retires `normalizeDistance` entirely) — more work, needs profiling.
- It only *raises* the ceiling; it doesn't make it unbounded. Under **world streaming (#938)** coordinates are genuinely unbounded, where the partition's O(1) correctness still wins. So 32-bit may let us **simplify** (drop the partition special-casing) for practical worlds, but it doesn't strictly dominate the partition on correctness.
- **Net:** ship the partition now (unconditional, unblocks B/C/D/E); #1983 evaluates retiring the convention and simplifying later. That keeps today's fix correct and the long-term simplification on the books.

---

## Re-scoped acceptance criteria (Δ from the plan)

- Foreground-priority detached/voxel solids render **fully in front** of any floor they are above, at **all zooms and all yaws** (Bug A), with **no dependence on world extent** — verify with canvas_stress canary at high zoom *and* with the orbit at its full r=200 (no orbit shrink).
- Cardinal fast path + all in-budget content: `enc` **byte-identical** to master (the clamp must not fire for current demos — assert/log it).
- World-extent assert replaced by the partition-layout assert above.
- Depth-probe (#1910) confirms a foreground fragment reads nearer than the floor fragment at the same pixel across the rotation.
- B implements the **two-tier** split only; per-trixel tiers are D #1960.

Source of truth for this model is **engine-level** → update the `## Resolution` section of `docs/design/depth-unification-1884-investigation.md` (replace the additive `priorityBand·BAND` formula with this partition model + the unconditional-dominance invariant + the #1983 pointer) **in this PR**, and fold this direction into `.fleet/plans/issue-1958.md` on the branch before resuming. Part A (quadrant-stable cardinal depth) is unchanged and can land alongside.

jakildev:
Resumed from design-unblock and **implemented the full task** (Parts A + B). Rebased onto current master (the inherited #1957 prefix dropped via `rebase --onto` since #1962 squash-merged). Adopted the architect's **disjoint near-plane partition** over the additive band.

Verified macOS/Metal: Bug A fixed (canary fully in front of floor, `--depth-probe-assert` PASS `rawDist=-57391` in the foreground band); `shape_debug` 19/21 shots **byte-identical to origin/master**, only the 2 non-cardinal yaw45 shots change (Part A's intended #1370 fix). Design doc §Resolution + plan folded to the partition model. GL/Linux unverified (Metal host) — cross-host smoke pending. Dropped `fleet:wip` → ready for review. — worker (opus)

jakildev (commented):
## Review — render: quadrant-stable SDF depth + foreground depth partition (#1958)

**Verdict:** approve (Opus recheck required — see footer)

### Checklist pass

**std140 layout / CPU↔GPU struct sync**
- `FrameDataTrixelToFramebuffer::depthPriorityMode_` repurposes the former `_depthColorPad1_` slot at offset 204. Computed manually: the field chain from `mpMatrix_` (0) through `_depthColorPad0_` (200) totals 204 bytes, and `depthPriorityMode_` (int) lands at 204–207. Total = 208. Matches the `static_assert(sizeof(...) == 208)`. ✓
- GLSL `FrameDataIsoTriangles` and Metal `FrameDataIsoTriangles` both match field-for-field: float4x4 → vec2×5 → float → int → ivec2/int2 → float/int → ivec4/int4 → vec4/float4×3 → int/float/float/int. Metal default-packed layout reaches `depthPriorityMode` at 204, same as C++ and GLSL std140. ✓
- `kDepthForegroundBandWidth = 16384` is consistent in C++ (`ir_render_types.hpp:219`), GLSL (`ir_iso_common.glsl:135`), and Metal (`metal/ir_iso_common.metal:124`). ✓

**Partition math**
- `kDepthForegroundCeil = kTrixelDistanceMinDistance + kDepthForegroundBandWidth = −65535 + 16384 = −49151`. The two `static_assert`s confirm it's negative and inside the encodable range. ✓
- World clamp: `enc = max(enc, foregroundCeil + 1)`. Fires only when `cardinalIsoDepth × 4 < −49151`, i.e., a world object at depth > ~12288 subdivided units — well past the r=200 orbit at effSub 16 (~6400). No-op for all current demos. Cardinal fast path is byte-identical. ✓
- Foreground path: `enc = rawDist × depthScale + kDepthForegroundBandCenter`, then `clamp(enc, kMinTriangleDistance, foregroundCeil)`. For a pool with rawDist centered at 0, enc ≈ kDepthForegroundBandCenter = −57343; saturates at ceiling (−49151) for rawDist > ~8192 scaled units. Graceful degradation, intentional. ✓
- `depthScale` applies identically for both foreground and world-placed: `effectiveSub / cubeSub`, so model-frame rawDist self-occludes at the right resolution in both tiers. ✓

**Part A — quadrant-stable SDF depth** (`c_shapes_to_trixel.glsl:947–949`, `.metal:1040–1042`)
- Replaces `yawC/yawS` (from continuous `visualYaw`) with `cardinalCosSin.x/y` (from the cardinal-bracket index) in the depth metric only. Screen placement still uses the continuous `visualYaw` path (`worldSurface` is built above). At an exact cardinal, `smoothYaw` is gated off and `cardinalCosSin == (cos/sin)(rasterYaw)` = `(cos/sin)(visualYaw)`, so cardinal frames stay byte-identical. ✓
- `dvx/dvy` formula is a valid R_z(−cardinalYaw) operation matching the comment "Rotate the surface by R_z(−visualYaw) and take its iso-depth". ✓

**ECS invariants**
- `depthPriority_` is read directly from `C_EntityCanvas`, which the system already iterates — no per-tick `getComponent` foreign lookup. ✓
- `depthPriorityMode_` is set in the per-entity tick body (as a local field on `fd`, the `FrameDataTrixelToFramebuffer` being built for this draw), not a shared static. ✓

**Naming / style**
- New public fields: `depthPriority_` (trailing `_`, correct). Constants: `kDepthForegroundBandWidth`, `kDepthForegroundCeil`, `kDepthForegroundBandCenter` (behavior-named, engine/render/ tier). ✓

**Lua binding**
- `depthPriority` exposed in `component_entity_canvas_lua.hpp:24–25`. ✓

**Shader file parity**
- Both GLSL and Metal updated in lockstep for Part A (depth metric bracket) and Part B (partition logic + struct field). No GLSL without a Metal counterpart. ✓

**Design doc update**
- `docs/design/depth-unification-1884-investigation.md §Resolution` updated to the partition model (architect-required). ✓

### Praise

- The NEEDS-DESIGN escalation chain is impeccable: measured the band-headroom failure before implementing, quantified it precisely (canvas_stress r=200 orbit vs. shape_debug ~161 max depth), framed the exact question for the architect, and recommended Option 2 (disjoint partition) as the robust choice. Clean handoff with evidence.
- Repurposing the `_depthColorPad1_` std140 slot is the right call — no UBO resize, no shader breakage, `sizeof` assert already guards it. The comment chain explaining *why* offset 204 is safe is exactly what future maintainers need.
- The partition-layout `static_assert`s replace the unsatisfiable world-extent assert (proved by the orbit geometry) with constants-only invariants that hold unconditionally. The comment explains the tradeoff explicitly.
- `kDepthForegroundBandCenter` as the `distanceOffset_` for priority canvases is elegant: a pool centered at (0,0,0) in model-frame gets centered in the band, and the gather clamps the edges so a pathologically deep solid saturates without escaping.

- [x] `IRCanvasStress --only canary,floor --zoom 8` → canary renders fully in front of floor (author reports PASS, `rawDist=−57391`)
- [x] `--depth-probe-assert 321,210` → `result=PASS` (inside foreground band `[−65535, −49151]`, normDepth 0.062)
- [x] `shape_debug` cardinal shots: 19/21 byte-identical; 2 non-cardinal yaw45 shots change by design (Part A fix for #1370)
- [ ] **GL/Linux smoke pending** — author verified Metal only; see cross-host smoke tag below

Escalation: please Opus-recheck before merge (touches: `engine/render/src/shaders/`, `engine/prefabs/irreden/render/`, GPU depth-partition semantics, std140 slot repurpose, `depthScale × rawDist + kDepthForegroundBandCenter` interaction across all zoom/subdivision paths — Sonnet cannot fully verify GPU buffer lifetime and the partition's graceful-degradation bound under all subdivision combinations).

Opus recheck required: render pipeline — GPU depth partition + std140 repurpose.

jakildev (commented):
## Opus recheck — render: quadrant-stable SDF depth + foreground depth partition (#1958)

**Verdict:** approve

Sonnet's first pass was thorough and accurate. I focused this recheck on the four Opus-only items Sonnet flagged it could not fully confirm: the std140 slot repurpose, the partition's graceful-degradation bound under all subdivision combinations, the `depthScale × rawDist + bandCenter` interaction across zoom, and frame-data lifetime / stale-flag leakage. All confirmed correct.

### std140 slot repurpose (offset 204) — confirmed sound on both backends
Sonnet computed offset 204 manually; on closer read I confirm it structurally. The C++ change is a same-width repurpose (`float _depthColorPad1_` → `int depthPriorityMode_`, both 4 bytes) at the trailing slot, so `sizeof == 208` and every `offsetof` assert (`visibleFaceIds_==128`, `detachedResidual_==144`, … `depthColorMode_==192`) is untouched. The newly-declared GLSL gather tail in `f_trixel_to_framebuffer.glsl` lands `depthPriorityMode` at 204 by std140 rules (`ivec2`@112 / `int`@124 / `ivec4`@128 / `vec4`×3 @144,160,176 / `int,float,float`@192,196,200 / `int`@204). Crucially this is **the exact same tail `v_peraxis_scatter.glsl` already declares and the scatter UBO asserts already guard** — so the layout isn't novel, it's a proven one. Metal's default-packed alignment reaches 204 identically (int4 aligns 16, matching), and the Metal scatter shaders already read this tail. ✓

### Partition logic + the invariant — correct by construction
`clamp(enc, kMin, foregroundCeil)` for priority vs `max(enc, foregroundCeil+1)` for world makes a foreground fragment **unconditionally** land in `[kMin, foregroundCeil]` and a world fragment unconditionally land above it — the "nearer by partition membership, not by out-sizing world spread" invariant holds regardless of world extent. `foregroundCeil = kMinTriangleDistance + 16384` in-shader matches `kDepthForegroundCeil` in C++ (same `kMin` + same hardcoded width mirrored in `ir_iso_common.glsl`/`.metal`). Both layout asserts (`< 0`; inside range) are trivially true (`-49151`). ✓

### Graceful-degradation bound — documented limitation, not a defect
The `max()` world clamp is a no-op **only for current demos at the effSub-16 cap** (architect's measured spread ≈ ±25600 vs the ‑49151 threshold), and `shape_debug` 19/21 byte-identical + the `--depth-probe-assert` PASS confirm it empirically (the 2 deltas are Part A's intended #1370 fix, not the clamp). Beyond that bound, deep/near world content *saturates* against the boundary (loses depth resolution, possible z-fight among world content) rather than letting a background fragment beat a priority solid — strictly safer than the additive model. This is the accepted two-tier tradeoff, correctly documented in the struct comment and §Resolution, with #1983 filed to widen the composite depth range. Sonnet flagged this bound as the thing it couldn't verify "under all subdivision combinations"; on closer read I confirm it is **bounded and intentional**, not a silent wrap. ✓

### Part A quadrant-stable depth — in scope, cardinal byte-identity preserved
`cardinalCosSin` is defined in `main()` (GLSL :769, Metal :882) well before the depth-metric site, so the swap to bracket cos/sin needs no new plumbing. Screen placement still uses `yawC/yawS` (continuous `visualYaw`); only the stored depth becomes piecewise-constant. At a cardinal, `smoothYaw` gates `yawC == cardinalCosSin.x`, so depth metric and placement agree → cardinal frames byte-identical. Symmetric on both backends. ✓

### Frame-data lifetime / stale-flag — clean
Two upload producers. `ENTITY_CANVAS_TO_FRAMEBUFFER` builds a fresh `fd{}` and sets `depthPriorityMode_` explicitly every draw (`foregroundPriority ? 1 : 0`) — the only site that ever writes `1`. The *persistent* `mainFramebuffer` component (reused across frames and shared with the per-axis scatter path) now gets an explicit `= 0` in the gather path, so a foreground flag can never leak into the world gather. The defensive reset is necessary (persistent component) and sufficient (scatter shaders don't read the field). ECS-wise the composite reads `depthPriority_` straight off the `C_EntityCanvas` it already iterates — no per-tick foreign `getComponent`. ✓

### For the human's awareness (not a blocker)
The byte-identity + foreground-dominance guarantee is bounded to current demos at the effSub-16 cap; a much deeper world or higher zoom degrades gracefully (saturation) rather than corrupting foreground priority. This is the deliberate two-tier scope with the widen tracked at #1983 — surfacing it only so the bound is visible at merge time. GL/Linux smoke is still pending (`fleet:needs-windows-smoke`); Metal-only verified by the author.

🤖 Opus recheck by Claude Opus (opus-reviewer role). Confirms the Sonnet 4.6 first pass.

PR #1994, opened 2026-06-25
render: circular (Euclidean) fog-of-war reveal radius

### LINKED ISSUE #1993: render: circular (Euclidean) fog-of-war reveal radius
## Summary

Make the fog-of-war reveal a **circle** instead of its current non-round shape.
`Fog::revealRadius` marks cells visible using **taxicab / Manhattan distance**
(`|dx| + |dy| > radius`), which reveals a **diamond/rhombus**, not the round
radius a player expects. Switch the per-cell test to **Euclidean** distance so a
reveal at radius `R` produces a filled circular disc.

## Current behavior

`engine/prefabs/irreden/render/components/component_canvas_fog_of_war.hpp:153-174`,
inside `revealRadius(int cx, int cy, int radius)`:

```cpp
const int dx = x - cx;
const int dy = y - cy;
if (std::abs(dx) + std::abs(dy) > radius)   // Manhattan -> diamond reveal
    continue;
```

The surrounding bounds (`cx ± radius`, `cy ± radius`) already clamp to a square
AABB; **only this inner per-cell test decides the reveal shape**.

## Desired behavior

Reveal a filled circle of `radius` cells centered on `(cx, cy)`:

```cpp
const int dx = x - cx;
const int dy = y - cy;
if (dx * dx + dy * dy > radius * radius)     // Euclidean -> circular reveal
    continue;
```

- **Integer math only** — squared-distance compare, no `sqrt`, no float,
  allocation-free.
- `radius` stays inclusive (a cell exactly `radius` cells away on an axis is
  revealed).
- Bonus cleanup: the new form has no `std::abs`, so it removes the two
  `std::abs` calls (a `.claude/rules/cpp-math.md` deviation) at this site. Do
  **not** reach for `IRMath` here — plain integer `*`/`+` is correct and is not
  a math-primitive that rule governs.

## Scope / notes

- **Single-point change** in the CPU reveal helper. The reveal shape is fully
  determined by which cells `revealRadius` marks visible.
- **No shader / backend-parity work.** `FOG_TO_TRIXEL` and the
  `c_fog_to_trixel` GLSL/Metal shaders only composite fog *state*
  (unexplored / explored / visible); they never compute the radius shape, so
  neither backend changes.
- No new components, systems, or `SystemName` enum entries; the AABB clamp,
  out-of-range drop, dirty-flag upload gating, and the three fog states are all
  unchanged.

## Acceptance criteria

- [ ] `revealRadius` uses a Euclidean (squared-distance) test; a reveal at
      radius `R` produces a circular disc, not a diamond.
- [ ] Integer-only math — no `sqrt`, no float, allocation-free; the two
      `std::abs` calls at this site are gone.
- [ ] AABB clamp, out-of-range handling, dirty/upload path, and fog states
      unchanged.
- [ ] Builds clean; a fog reveal renders round (spot-check in a demo that
      exercises fog of war).

**Model:** sonnet

jakildev:
`revealRadius` in `C_CanvasFogOfWar` used taxicab distance (`|dx|+|dy| > radius`), producing a diamond-shaped visible region where the method doc comment promised Euclidean (circular). The switch to `dx*dx + dy*dy > radius*radius` removes the two `std::abs` calls that were IRMath-rule deviations and matches the documented contract.

`fog_demo` band computation updated from taxicab rings to squared-Euclidean rings (`distSq > innerSq && distSq <= outerSq`) so the demo's explored band correctly hugs the now-circular visible region. Stale "diamond" / "taxicab" comments updated throughout.

## Screenshots

<details>
<summary>fog_zoom2_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>fog_zoom4_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>fog_zoom8_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

Closes #1993


jakildev (commented):
## Review — render: circular (Euclidean) fog-of-war reveal radius (#1994)

**Verdict:** approve

### What this PR does

Fixes `C_CanvasFogOfWar::revealRadius` to use Euclidean distance (`dx*dx + dy*dy > radius*radius`) instead of taxicab (`|dx| + |dy| > radius`), matching the documented contract. Updates the `fog_demo` band computation from taxicab rings to squared-Euclidean rings using precomputed `innerSq`/`outerSq` constants. Stale "diamond"/"taxicab" comments updated throughout.

### Praise

- **Fixes two problems in one diff.** The old `std::abs(dx) + std::abs(dy)` in the component was both the wrong metric and an IRMath-rule violation (`std::abs` outside `engine/math/`). Switching to `dx*dx+dy*dy` eliminates both the correctness bug and the rule deviation.
- **Band math is consistent with the reveal shape.** The `fog_demo` band uses the same Euclidean squared metric with `innerSq < distSq <= outerSq`, which aligns exactly with the `revealRadius` inclusion boundary (`distSq <= radius*radius`). No cell is both VISIBLE and in the EXPLORED band. Precomputing `innerSq`/`outerSq` outside the loop is a clean optimization.
- **Before/after screenshots clearly show the shape change.** Diamond → circle is visible across all three zoom levels.

### Nits

- `component_canvas_fog_of_war.hpp:164` — `radius * radius` is undefined behavior for `radius > ~46340` (signed integer overflow). In practice the grid is 256×256 so radii above 128 are meaningless, but there's no assertion or cap. The original code (`|dx| + |dy| > radius`) had no analogous overflow risk. Suggestion: `IR_ASSERT(radius <= kFogOfWarHalfExtent, "revealRadius: radius exceeds grid half-extent")` before the loop, or compute `const int radiusSq = radius * radius;` with a comment noting the practical bound.

### Test plan

- [x] Before/after screenshots included for zoom×2, ×4, ×8 — circular shape confirmed at all zoom levels.
- [ ] `fog_demo` build + smoke on the Windows/OpenGL tier (macOS Metal covered by author; see smoke label).

Opus recheck not required.

jakildev:
Addressed the nit (component_canvas_fog_of_war.hpp): clamped `radius` to `2 * kFogOfWarSize` before squaring so `radius * radius` can no longer overflow signed int for absurd inputs. The loop bounds already clamp iteration to the grid, and the cap exceeds the largest squared cell distance any in-bounds center can produce, so every in-range call is byte-identical; the multiply is now also hoisted out of the inner loop. Build clean (IRShapeDebug). `fleet:approved` retained. — worker

Issue #1963, opened 2026-06-22
render: profile the rotation perf cliff — cardinal vs off-cardinal cost breakdown (scopes #1961) [#1884]


jakildev:
**Model:** opus · **Part of:** #1884 (depth-unification sub-epic) — perf precursor / scoping gate for #1961.
**Blocked by:** (none — profiles current master; the cliff predates the epic, so this can run now).

## Why first
Before committing to the perf-parity fix (#1961), **measure the rotation cliff empirically** — don't guess what's slow. The cliff is architectural: at an exact cardinal the renderer uses **route 0** (single-canvas raster, 1 voxel→trixel dispatch, byte-identical to master); the instant you rotate off-cardinal it switches to the **per-axis routes 1/2/3** (#1309) — 3 dispatches + continuous per-frame repositioning/deformation + depth recompute.

## Scope
Profile `IRPerfGrid --yaw-ramp` (+ the per-system perf sensor / `scripts/dev/perf-grid-rotate-sweep`) and attribute the per-frame cost delta **cardinal vs mid-bracket** to each component:
- voxel→trixel dispatch (1 → 3),
- per-frame repositioning / face deformation,
- depth recompute / sort,
- GPU upload / cache invalidation,
- composite.

## Deliverable
A documented breakdown (which components dominate the Nx) that **scopes #1961** — tells us whether the lever is bracket-caching the per-axis raster, removing per-frame recompute, consolidating the route-0 path, or something else. Also confirms/refutes the hypothesis that the 3-dispatch + per-frame recompute dominate (i.e. that quadrant-stability → bracket-caching is the right fix). No production code change — this is a profiling + write-up spike.

## Refs
#1884; #1961; #1309 (per-axis routes); `IRPerfGrid --yaw-ramp`.

PR #2003, opened 2026-06-25
render: reconcile per-axis screen-depth GPU-stage name

### LINKED ISSUE #1965: render/profile: per-axis screen-depth resolve stage is silently untimed (registry/tag name mismatch)
Found while profiling the rotation perf cliff (#1963).

## Problem

`SYSTEM_RESOLVE_PER_AXIS_SCREEN_DEPTH` instruments itself with
`IRRender::tagGpuStage(systemId, "resolvePerAxisScreenDepth")`
(`engine/prefabs/irreden/render/systems/system_resolve_per_axis_screen_depth.hpp:204`),
but the GPU-stage registry (`engine/prefabs/irreden/render/gpu_stage_timing.hpp:210`)
has **no row named `resolvePerAxisScreenDepth`**. `tagGpuStage` is a silent
no-op on an unknown name, so this stage's GPU time is **never recorded** on
either backend (no field, no accumulator sample, omitted from the shutdown
profile report).

Conversely, the registry **does** have a `screenSpaceResidualRotate` row
(field `screenSpaceResidualRotateMs_`, line 38/230) that **no system tags** —
it is dead, always 0.0, and only referenced by the perf overlay as "SCREEN-ROT"
(`system_perf_stats_overlay.hpp:202`). This reads like a rename that updated the
system tag but not the registry row (or vice versa).

## Impact

The per-axis screen-depth resolve runs only off-cardinal and is part of the
rotation perf cliff's dominant `trixelToFb`-bracketed cost. Because it is
untimed, the composite-vs-resolve split inside the per-axis path cannot be
measured. #1963 needed this split to fully scope #1961's fix and could only
attribute the cliff to `trixelToFb` as a whole.

## Fix

Reconcile the name: either rename the registry row `screenSpaceResidualRotate`
→ `resolvePerAxisScreenDepth` (updating the field name + the overlay label), or
retag the system to `screenSpaceResidualRotate`. Confirm the stage then appears
in `save_files/profile_report.txt` with non-zero samples off-cardinal.

## Refs

#1963 (profile write-up: `docs/design/rotation-perf-cliff-1963-profile.md`),
#1961, #1746 (sibling profiling-infra gap: multi-encoder timestamp undercount).

jakildev:
## Summary
- `SYSTEM_RESOLVE_PER_AXIS_SCREEN_DEPTH` tags `tagGpuStage(systemId, "resolvePerAxisScreenDepth")`, but the GPU-stage registry had no row of that name — it carried a dead `screenSpaceResidualRotate` row no system tags. `tagGpuStage` is a silent no-op on an unknown name, so the stage's CPU+GPU time was **never recorded** on either backend (no observer install, no accumulator sample, omitted from the shutdown profile report).
- Completes the half-done rename: registry row + `GpuStageTiming` field + perf-overlay label → `resolvePerAxisScreenDepth` / `resolvePerAxisScreenDepthMs_` / `PERAXIS-DEPTH`. The descriptive name the system already uses wins; the orphaned legacy name retires.
- No Lua script keys the old name, so the `getPassTimings` / `getPassTiming` / `getCpuPassTiming` Lua surface is unaffected (renamed key was always a dead 0.0).
- Docs naming the same physical stage move to the canonical name: the #1963 profile write-up (marked resolved), the Metal perf baseline table, and the `optimize` skill's `gpu_profiling.md` reference.

## Test plan
- [x] `fleet-build --target IRShapeDebug` clean (registry consumer `world.cpp` + rendering lib recompiled).
- [x] Runtime acceptance: `IRShapeDebug --yaw 0.4 --auto-profile 150` (off-cardinal, `gpu_stage_timing` on) now lists `resolvePerAxisScreenDepth` in `save_files/profile_report.txt` with non-zero samples (avg 0.089 ms, 149 GPU samples) where it was previously absent entirely.
- [x] No pixel impact — the renamed field is read only by the perf-stats overlay (debug toggle, off by default) and the shutdown profile report, never by the framebuffer-writing render path; backend-agnostic CPU-side instrumentation.

Closes #1965

### DIFF
--- a/.claude/skills/optimize/reference/gpu_profiling.md
+++ b/.claude/skills/optimize/reference/gpu_profiling.md
@@ -13,7 +13,7 @@ Registered stages (canvasClear, voxelCompact, voxelStage1, voxelStage2,
 shapeCompact, shapePass0, shapePass1, textToTrixel,
 buildLightOcclusionGrid, computeVoxelAO, bakeSunShadowMap,
 computeSunShadow, computeLightVolume, lightingToTrixel, fogToTrixel,
-trixelToTrixel, trixelToFb, entityCanvasToFb, screenSpaceResidualRotate,
+trixelToTrixel, trixelToFb, entityCanvasToFb, resolvePerAxisScreenDepth,
 fbToScreen) each carry a soft budget share (`GpuStageInfo::budgetShare_`).
 A pass exceeding its share is flagged via `overBudget` in the Lua surface.
 
--- a/docs/design/rotation-perf-cliff-1963-profile.md
+++ b/docs/design/rotation-perf-cliff-1963-profile.md
@@ -129,7 +129,10 @@ is *composite + resolve* combined; splitting them needs the infra fix.
    row (or vice versa). Fixing it (rename the registry row to
    `resolvePerAxisScreenDepth`, or retag the system) makes the resolve stage
    observable and lets #1961 separate composite cost from resolve cost. Filed
-   as #1965.
+   as #1965. **Resolved (#1965):** the dead `screenSpaceResidualRotate`
+   registry row (field + overlay label) was renamed to
+   `resolvePerAxisScreenDepth`, completing the half-done rename so the system's
+   existing tag now resolves and the stage records on both backends.
 
 ## Scoping #1961
 
--- a/docs/perf/metal_perf_grid_baseline.md
+++ b/docs/perf/metal_perf_grid_baseline.md
@@ -93,7 +93,7 @@ Metal GPU-stage snapshot from the final frame:
 | `computeLightVolume` | 0.000 ms | 0.000 ms |
 | `lightingToTrixel` | 0.050 ms | 0.082 ms |
 | `trixelToFb` | 0.123 ms | 0.125 ms |
-| `screenSpaceResidualRotate` | 0.155 ms | 0.176 ms |
+| `resolvePerAxisScreenDepth` | 0.155 ms | 0.176 ms |
 
 `ComputeLightVolume` CPU phase breakdown:
 
--- a/engine/prefabs/irreden/render/gpu_stage_timing.hpp
+++ b/engine/prefabs/irreden/render/gpu_stage_timing.hpp
@@ -35,7 +35,7 @@ struct GpuStageTiming {
     float trixelToTrixelMs_ = 0.0f;
     float trixelToFbMs_ = 0.0f;
     float entityCanvasToFbMs_ = 0.0f;
-    float screenSpaceResidualRotateMs_ = 0.0f;
+    float resolvePerAxisScreenDepthMs_ = 0.0f;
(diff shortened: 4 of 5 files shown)


jakildev:
Filed #2004 as a follow-up: `tagGpuStage` silently no-ops on an unknown stage name, which is what let this rename drift go untimed unnoticed. A debug-build warning there would catch the next such mismatch loudly. Out of scope for this reconciliation. — worker

jakildev (commented):
## Review — render: reconcile per-axis screen-depth GPU-stage name

**Verdict:** approve

Narrow housekeeping fix — reconciles a half-done rename so the stage's existing
`tagGpuStage` call resolves instead of no-op'ing. No shader changes, no GPU
struct layout changes, CPU-side instrumentation only.

### Checklist

- **ECS invariants:** no ECS code changed — N/A.
- **Ownership/lifetime:** N/A.
- **Render pipeline (struct layout):** `GpuStageTiming` field rename is a
  pure C++ identifier change — no `std140` padding impact, no GPU-visible data.
  Registry pointer-to-member updated in the same hunk. No mismatch risk.
- **Render pipeline (binding points):** no binding changes — N/A.
- **Shader files:** none changed.
- **Math/coordinates:** N/A.
- **Serialization:** N/A.
- **Naming/style:** `resolvePerAxisScreenDepthMs_` follows the trailing-`_`
  public-member convention. Overlay label `"PERAXIS-DEPTH"` is consistent with
  adjacent short-form labels in `system_perf_stats_overlay.hpp`.
- **Tests/build:** PR body reports clean `fleet-build --target IRShapeDebug`
  and runtime acceptance (`--auto-profile 150` now records non-zero samples at
  `resolvePerAxisScreenDepth`). No test change needed for a profiler rename.
- **Completeness:** grepped all source (`.hpp`, `.cpp`, `.glsl`, `.metal`, `.lua`)
  for both `screenSpaceResidualRotate` and `resolvePerAxisScreenDepth`. Zero
  remaining references to the old name. All four update sites present:
  `GpuStageTiming` field, registry row, overlay label, and the 3 doc files.
  `system_resolve_per_axis_screen_depth.hpp:204` already used the correct tag
  `"resolvePerAxisScreenDepth"` before this PR — that was the drift source; the
  registry is now in sync.
- **Fleet scripts:** no identifier renames in scripts.

### Praise

- Follow-up issue #2004 filed for the silent `tagGpuStage` no-op on unknown
  names — correctly kept out of scope here but tracked so the infra gap closes
  separately.
- Docs updated in the same commit: profile write-up marked resolved, Metal
  baseline row and skill reference all flipped to the canonical name.
  No stale doc for a future reader to trip over.

- [x] `IRShapeDebug --yaw 0.4 --auto-profile 150`: `resolvePerAxisScreenDepth`
  appears in `profile_report.txt` with non-zero samples (verified by author: avg
  0.089 ms, 149 GPU samples).

Opus recheck not required.

Issue #1961, opened 2026-06-22
render: rotation perf parity via per-axis raster bracket-caching + cardinal-path consolidation (IRPerfGrid cliff) [#1884]


jakildev:
**Model:** opus · **Part of:** #1884 (depth-unification sub-epic) — the perf capstone.
**Blocked by:** (none) — formal blockers #1963 (profile) and #1958 (B) are both merged. #1965 (resolve-stage timing rename) was **de-gated** by the 4th planning pass (2026-06-24): the committed lever (per-axis empty-cell compaction) attacks the common driver of both the composite and resolve halves, so it is split-robust and #1965 is only a verification aid, not a planning prerequisite. Plan filed; WIP PR #2007 open.

## Goal
Close the IRPerfGrid rotation perf cliff — smooth-rotating cost ≈ exact-cardinal cost. **This is not just a perf check:** the cliff is the architectural divergence between **route 0** (cardinal: 1 voxel→trixel dispatch, byte-identical to master) and **routes 1/2/3** (off-cardinal: 3 dispatches + per-frame reposition/deformation + depth recompute, #1309). Closing it is real code work, scoped by the profile #1963.

## Lever (hypothesis — confirm/refine via #1963)
Extend quadrant-stability from depth (B, #1958) to the per-axis **raster**: within a 90° bracket the per-axis *routing* (which face per axis) is cardinal-stable; only the continuous reposition/deformation varies. So:
- **Bracket-cache** the per-bracket 3-canvas voxel→trixel raster; re-run only the cheap per-frame 2D deformation instead of the full dispatch every frame (*additional logic to get close* — reuse over recompute).
- **Consolidate / remove extraneous paths:** if bracket-caching makes the per-axis path cheap enough, the **route-0 cardinal special-case may become redundant and removable** (residual=0 is just one point in the bracket) — a simplification, not only a speedup.

The split between caching vs path-removal vs other is set by #1963's breakdown.

## Acceptance
- IRPerfGrid rotating-yaw cost approaches cardinal-yaw cost (close the documented cliff); measured before/after with the perf sensor.
- rotated-solidity (#1882) + cardinal frames byte-identical (no correctness regression).
- Report whether the demo's visual issues under rotation also resolve (the hypothesis).

## Refs
#1884 design (quadrant-stability); #1963 (profile); #1309 (per-axis routes); #1958 (B).

jakildev:
[...]

Issue #1939, opened 2026-06-21
render: retire the manual scatter dilation margin/miter/yield tower + full validation (C3)


jakildev:
**Model:** opus
**Effort:** high
**Part of epic:** #1933
**Plan file:** `~/.fleet/plans/issue-1933.md` (full epic plan)
**Blocked by:** #1938, #1922

## Scope

With edge-type-aware analytic coverage authoritative on **both** backends
(#1937 GL, #1938 Metal), retire the now-redundant manual conservative-dilation
margin/miter/yield tower from the per-axis scatter pass, then run the full
validation (temporal stability + solidity) and update the docs. This is the
"removes the hot-path geometry hack" deliverable of epic #1933.

## Approach

- Retire / reduce to a minimal documented visit-bound, across BOTH the
  camera-path and detached regimes, in `ir_iso_common.{glsl,metal}` +
  `{v_,f_}peraxis_scatter.glsl` / `peraxis_scatter.metal`:
  - `scatterConservativeDilation()` — collapse to the minimal fixed visit-bound
    the analytic coverage still needs (the rasterizer must visit candidate
    pixels), dropping the per-axis `0.5·|n|` growth + miter.
  - `kScatterMiterLimit` (#1538) — remove with the miter.
  - `kScatterDilateMarginPx` (#1494) — keep only if still needed as the
    visit-bound floor; otherwise remove.
  - `kScatterDetachedPitchFraction` (#1538) — remove (detached seams are now
    closed by interior-edge conservative fill, not a pitch-proportional margin).
  - `kScatterMarginDepthBiasKey` (#1457) + `kScatterMarginYieldGradScale`
    (#1883) — remove the fragment margin-vs-interior depth-bias / yield-gradient
    machinery: with no over-fill margin, every surviving fragment is interior.
- Update `engine/render/CLAUDE.md`: the convex-corner drift note (the
  #1917-accepted drift) is now fixed — rewrite it to describe the analytic
  coverage model instead.

## Acceptance criteria

- **Temporal stability** under the #1922 jitter harness on both backends (the
  core bar — unverifiable before #1922 is on master, hence the blocker).
- Coarse-cube near-cardinal: crisp corners + solid foreshortened faces, **with
  the margin tower removed** (regression-free vs #1938's state).
- Epic #1881 rotated-solidity gate (`perf-grid-rotate-sweep`) passes; cardinal
  fast path byte-identical; both backends at parity.
[...]

Issue #1937, opened 2026-06-21
render: analytic edge-aware coverage on the per-axis scatter — Metal backend (C1, lead)


jakildev:
**Model:** opus
**Effort:** high
**Part of epic:** #1933
**Plan file:** `~/.fleet/plans/issue-1933.md` (full epic plan)
**Blocked by:** (none)

## Scope

Replace the per-axis voxel scatter pass's **uniform conservative-dilation
margin coverage decision** with **per-fragment analytic coverage** on the
OpenGL backend, so a face's coverage is correct at the source. This is the
de-risking + first-implementation phase of epic #1933's root fix for
near-cardinal corner spikes (#1883/#1917) and silhouette dashing. Keep the
quad dilation only as a "visit enough pixels" rasterization bound; move the
coverage decision into the fragment stage.

## Approach

- In `v_peraxis_scatter.glsl`: keep a **minimal fixed** quad dilation as the
  rasterization visit-bound (so the rasterizer still visits every pixel the
  true footprint could touch). Pass the **true (undilated) face-quad edge
  equations** as varyings, plus a per-edge **interior-vs-boundary** flag
  derived from an occupied-same-plane-neighbor tap of the per-axis canvas data
  (the vertex stage already knows the cell's grid position via `gl_InstanceID`
  and reads the per-axis textures).
- In `f_peraxis_scatter.glsl`: compute the fragment's signed distance to each
  true edge and form **edge-type-aware analytic coverage**:
  - **Interior edges** (shared with an occupied neighbor): fill conservatively
    (round outward ~½ px) — keeps the inter-cell waffle closed.
  - **Boundary/silhouette edges:** exact sub-pixel analytic coverage
    (`clamp(0.5 - signedDist/fwidth, 0, 1)`-style), **hard-thresholded** for the
    R32I-distance / depth co-sort write (no alpha-blended partial coverage).
  Corners (two boundary edges) get exact treatment → no spike; foreshortened
  silhouettes get sub-pixel coverage → no dashing.
- Leave the margin/miter/yield constants in place for now (C3 retires them once
  Metal parity lands) — but the analytic coverage must already be the authority
  that removes the spikes/dashing, with the dilation reduced to the visit-bound.
- Finalize the exact per-edge coverage formula + neighbor-tap against the
  solidity gate during implementation (the gate is the safety net).

## Acceptance criteria

- Coarse-cube (3³ / 12³) **near-cardinal** poses: corner spikes gone AND
[...]

PR #2007, opened 2026-06-25
render: per-axis empty-cell compaction — close the IRPerfGrid rotation perf cliff (#1961) [WIP]

### LINKED ISSUE #1965: render/profile: per-axis screen-depth resolve stage is silently untimed (registry/tag name mismatch)

jakildev:
**Issue:** #1961 · **Part of:** #1884 (depth-unification sub-epic, perf capstone) · **Model:** opus

Stacked on #2003 (`Closes #1965`, approved) — the per-axis screen-depth GPU-stage registry rename. Step 0 of the plan is inherited from the base, so this PR does **not** re-close #1965; it builds the compaction on the renamed `resolvePerAxisScreenDepth` timing registry.

## Goal
Close the IRPerfGrid rotation perf cliff (#1963 profile): off-cardinal frame ~7.0→~11.5 ms (+4.6 ms), the entire delta in `trixelToFb`. The off-cardinal path replaces the cardinal gather with `drawPerAxisScatter` — a 3-pass forward-scatter over worst-case-sized per-axis canvases (~12× cardinal area, mostly empty texels). Both the composite and the resolve pass sweep the full oversized grid; **compacting to the non-empty cells cuts the variable cost of both** (split-robust lever).

## Approach (per the committed plan, `.fleet/plans/issue-1961.md`)
1. Compaction pre-pass compute shader (GL+Metal): scan each axis distance texture, append non-empty cell indices to a per-axis SSBO + write indirect args.
2. Composite → compacted/indirect instanced draw; `v_peraxis_scatter` reads cell index from the compacted list.
3. Resolve pass 1 → indirect dispatch over the same list; `c_resolve_per_axis_screen_depth` reads cell index from the list.
4. Re-profile + byte-identical regression gate (cardinal route 0 untouched).

## Status
🚧 WIP — implementing. Cardinal (route 0) stays byte-identical; backend parity (GL+Metal) mandatory. Verified locally on Metal; GL via Linux smoke.


jakildev:
## NEEDS-DESIGN

The empty-cell compaction is implemented and **almost** byte-identical, but inspection of the per-axis composite revealed an interaction the plan didn't anticipate. Surfacing it for a design call before going further.

### What works (verified on macOS/Metal, shape_debug)
- Compaction pre-pass + indirect instanced draw render the **same occupied cells** as the old full-grid sweep.
- **20/21 shots byte-identical to base** (`origin/master`); cardinal byte-identity holds (the path is gated off at cardinal).
- Two latent Metal-backend bugs the compaction exposed are fixed generally: freed textures lingering in sticky bind slots (`untrackMetalTexture`) and the cell compaction leaking slots 25/26 from `VOXEL_TO_TRIXEL_STAGE_1`'s sticky binding (restore added).

### The problem: compaction reorder → non-deterministic depth-tie margins
The 2 per-axis (yaw45) shots — `zoom4_yaw45_inter_cardinal`, `zoom4_pan16_yaw45_pivot` — are **non-deterministic run-to-run** (88–132 px / ~0.003%, `max_delta` ~21, all at `#1494`/`#1883` dilation-margin slivers). The base is deterministic there, so this is a regression I introduced.

**Root cause:** the old composite draws all `size.x*size.y` cells in **row-major instance order** (deterministic). The compaction appends occupied cells via `atomicAdd`, so the compacted list — hence the indirect draw's instance/raster order — is in **non-deterministic GPU-scheduling order**. Where two cells' dilation margins overlap a pixel at the **same plane depth** (a `GL_LESS` tie), the draw order decides the winner. The `#1457` margin-yield biases handle margin-vs-exact-footprint, but margin-vs-margin sub-pixel gap-fills still tie, and now resolve non-deterministically.

### The design question
Which way to make the compacted draw deterministic (and how byte-identical to base must it be)?

- **(a) Stable compaction (prefix-sum scan).** Compute each occupied cell's row-major rank via a parallel scan, so the compacted list preserves full-grid order → byte-identical + deterministic. Cost: ~2–3 extra passes + a per-block-count scan, both backends (the heaviest option; the genuine "do it right").
- **(b) Deterministic depth tiebreak per cell.** Add a stable per-cell sub-LSB key to the composite depth so margin-vs-margin ties resolve by cell index regardless of draw order. Cheaper, but touches the `#1457`/`#1883` margin-depth tuning — risk of reopening the wrong-voxel-band class.
- **(c) Accept as per-axis tolerance.** The per-axis/re-voxelize path already has documented accepted non-determinism (excluded from render-verify). 88px of near-identical-color sliver flicker is imperceptible. Document it and exclude these 2 shots — but it does introduce run-to-run jitter the base lacked.

My lean: **(a)** if the cliff win justifies the complexity (perf delta not yet measured — I prioritized correctness); **(c)** is the cheap ship if the architect judges the magnitude within per-axis tolerance. **(b)** I'd avoid (reopens tuned depth-tie behavior).

WIP pushed (6b972be1). Releasing the claim — any worker can resume from `fleet:design-unblocked`.

jakildev:
## Steward direction (architect, #1884) — resolves the depth-tie non-determinism

**Ruling: (c) accept as documented per-axis non-determinism — but rebase first, because the tie source is being removed underneath you.**

### Why not (a) or (b)
- **(a) prefix-sum stable compaction** adds 2–3 passes on both backends. This is a *perf* task; spending passes to chase sub-perceptual sliver flicker is self-defeating. Off the table unless the rebase below leaves a *perceptible* artifact.
- **(b) per-cell depth tiebreak** reopens the tuned #1457/#1883 margin-depth behavior (wrong-voxel-band risk). Agreed — avoid.

### Why (c) is correct here
The non-determinism is 0.003%, `max_delta ~21`, confined to **#1494/#1883 dilation-margin slivers** — exactly the class the per-axis path already documents as accepted non-determinism and excludes from render-verify. 88px of near-same-color flicker is within that tolerance.

### The rebase that likely shrinks it further (do this before deciding final scope)
This PR (pushed 06-25 05:13) **predates #1937 / PR #2013** (analytic edge-aware coverage on the per-axis scatter, merged 06-25 16:52), which made analytic coverage the *authority* over the dilation margin. **Rebase onto current master and re-measure the 2 yaw45 shots** — the margin-vs-margin sub-pixel gap-fill ties (the exact non-determinism source) are reduced by analytic coverage and will be removed entirely by C3 (#1939, retires the margin tower). So (c)'s documented exception is **self-resolving**, not permanent debt.

### Required to merge (the task's actual deliverable)
You prioritized correctness and have **not measured the perf delta**. Measure + report the IRPerfGrid rotating-vs-cardinal cost (per the #1963 profile method). If the cliff win is real → ship with (c): document the 2 shots as accepted per-axis non-determinism (cite the existing precedent) and exclude them. If there is **no meaningful win**, the compaction isn't worth even the determinism debt — say so and we reconsider the lever.

The two Metal-backend fixes you found (sticky-slot freed-texture, 25/26 leak) are good general fixes — keep them regardless.

Swapping `fleet:design-blocked` → `fleet:design-unblocked`. Resume: rebase → measure → (c)+docs or report-no-win.

jakildev:
## Design-unblock resolved (option c, #1884 steward ruling)

Resumed from `fleet:design-unblocked`. Followed the steward direction: **rebase → measure → (c)+docs**.

### Rebased onto master (#1937 / PR #2013 now merged)
The compaction's instance-index indirection layers cleanly on top of #2013's analytic edge-aware coverage (both orthogonal; #2013 coverage fully preserved). Branch is now 2 commits on `master` (plan + impl), force-pushed.

### Independently re-measured — and corrected the perf table
The WIP doc table carried numbers that **did not reproduce on re-measurement** (it reused the stale pre-occlusion-cull #1963 `cardinal=0.25 ms` figure and under-stated the PR frame). I re-measured master `de496a45` vs this PR on the same macOS/Metal host (64³ dense, --zoom 0.8, avg/200 frames, 2 runs each), **comparing the same pose** so the occlusion-cull ratio (#1798/#1799) is held constant:

| @ off-cardinal (yaw 0.39, ~91% culled) | master | #2007 | Δ |
|---|---|---|---|
| trixelToFb | 8.94 ms | 0.31 ms | **−8.6 ms** |
| Frame | 11.93 ms | 8.86 ms | **−3.07 ms** |

The cliff (off-cardinal costing +3.3 ms over cardinal on master) is **closed** (+0.3 ms). The cardinal-vs-off-cardinal framing the WIP doc used is now confounded by occlusion cull (~91% culled off-cardinal, ~0% at cardinal) — the per-axis scatter sweeps the full worst-case canvas *regardless* of cull, which is exactly what the compaction removes. Doc rewritten with the measured numbers + a reframing note.

### Verification (this host)
- `render-verify --target IRShapeDebug`: **all 24 PASS** across 3 runs. Cardinals byte-identical (max_delta 0). `zoom4_yaw45_inter_cardinal` byte-identical run-to-run; `zoom4_pan16_yaw45_pivot` stable at max_delta 7 / 100% — both inside gate thresholds, so neither is excluded (better than the architect's (c); no exclusion needed).
- Clean build (IRShapeDebug + IRPerfGrid), incl. after `format-changed`.

### Scope notes
- Composite consumer only; the resolve consumer is the plan's sanctioned split, tracked as low-priority follow-up #2015 (confirmed open).
- Kept the two general Metal-backend fixes (untrackMetalTexture sticky-slot, slot 25/26 restore).
- Dual-backend (GL+Metal shaders + both render-impls); verified on macOS/Metal only — GL/Linux smoke still needed on review.

Ready for re-review. — worker-3 (opus)

jakildev (commented):
## Review — render: per-axis empty-cell compaction — close IRPerfGrid rotation perf cliff (#1961)

**Verdict:** approve

Stacked on #2003 (now merged). Reviewing standalone diff (2 commits on master: plan + impl).

**Note on PR base:** the branch has been rebased onto master but the PR's base is still `claude/1965-resolve-peraxis-stage-timing`. The GitHub-computed diff via `fleet-pr diff` includes all commits inherited between the stale base and master — the ACTUAL diff vs master is the focused 18-file, 691-line set below. The PR should be retargeted to master before merge so the human sees the correct diff. (Adding as a nit, not a block.)

### What this PR does

Introduces a GPU compute pre-pass (`c_per_axis_cell_compact.glsl` / `.metal`) that scans each of the 3 per-axis distance canvases before the `drawPerAxisScatter` framebuffer pass, appending the linear indices of occupied cells into per-axis SSBO regions and setting the per-axis indirect instanced-draw arg counts. The scatter vertex shader and Metal vertex function (`v_peraxis_scatter.glsl` / `peraxis_scatter.metal`) now index into the compacted list via `compactedCells[gl_InstanceID]` instead of treating `gl_InstanceID` as a direct grid index. The result: only occupied cells are rasterized instead of the full worst-case grid (`size.x * size.y`, ~12× the cardinal area, mostly empty). Also adds `drawElementsInstancedIndirect` to both the GL and Metal `RenderDevice` implementations, and fixes a Metal-backend bug where freed per-axis canvas textures lingered in sticky bind slots.

Verified by author: `render-verify --target IRShapeDebug` 24/24 PASS on macOS/Metal (3 runs), cardinal byte-identical, off-cardinal: `trixelToFb` 8.94 → 0.31 ms (−8.6 ms), frame 11.93 → 8.86 ms (−3.1 ms). GL/Linux smoke still needed on review.

### Checklist pass

**Compaction kernel (GLSL + Metal)**
- Occupied test: `rawDist >= kEmptyDistanceEncoded (0x7FFFFFFF)` — consistent with `c_resolve_per_axis_screen_depth.glsl`'s sentinel. ✓
- Linear-index formula: `slot → compactedCells[slot] = y*size.x + x`; scatter recovery: `ivec2(cell % canvasSize.x, cell / canvasSize.x)` — roundtrips correctly. ✓
- GLSL: `atomicAdd(drawArgs[1], 1u)` → `instanceCount`. Metal: `atomic_fetch_add_explicit(&drawArgs[1], 1u, memory_order_relaxed)`. Both match the `PerAxisCellDrawCommand.instanceCount` field at offset 4 bytes. ✓
- Metal `drawArgs` declared as `device atomic_uint*` — correctly typed for Metal atomic ops. ✓
- Local size 16×16×1 matches `kPerAxisCellCompactGroupSize = 16` used in dispatch. ✓

**`system_trixel_to_framebuffer.hpp`**
- `compactPerAxisCells()` runs in `beginTick` BEFORE `quadVao_->bind()` and `framebuffer.bindFramebuffer()` — a pure compute prelude, slots 25/26 are free at this stage. ✓
- Three-axis dispatch loop issues `dispatchCompute` with separate `bindRange` calls per axis over non-overlapping SSBO regions (stride: `perAxisCellRegionStride_` × 4 B, 64-uint aligned). No intra-dispatch dependencies → single `memoryBarrier(SHADER_STORAGE)` + `memoryBarrier(COMMAND)` after the loop is correct. ✓
- `PerAxisCellDrawCommand` reset (CPU `subData`) sets `indexCount = kQuadIndicesLength`, `instanceCount = 0`, others = 0. Matches the GL `DrawElementsIndirectCommand` layout and Metal `MTLDrawIndexedPrimitivesIndirectArguments` layout (same leading 5 uints). ✓
- `restoreVoxelCompactionSlots()` rebinds slots 25/26 to VOXEL_TO_TRIXEL_STAGE_1's buffers after the scatter draws — prevents the "center cube regression" where the next frame's voxel compaction reads the cell list instead of the voxel list. Lazy lookup via named resource avoids creation-order coupling; no-op if voxel system is absent. ✓
- `ensurePerAxisCellBuffers`: `cellIndirectBuf_` created once (constant size = `kAxisCount × kPerAxisCellIndirectStrideBytes`); `cellCompactedBuf_` reallocated on canvas resize via `destroyResource + createResource` with fresh `cellCompactedId_` assignment. Both IDs stored for resource-manager cleanup. ✓
- `kPerAxisCellIndirectStrideBytes = 256`: aligns to the minimum SSBO bindRange offset requirement (≥256 B on all drivers; Metal also aligns to ≥256 B). ✓

**Slot aliasing**
- `kBufferIndex_PerAxisCellCompacted = kBufferIndex_CompactedVoxelIndices` (slot 25) — rationale documented in `ir_render_types.hpp`: voxel compaction finishes stages before the framebuffer pass, and SPRITE_TO_SCREEN (the other slot-25 consumer) draws after. Non-overlapping stage guarantee holds. ✓

**GL indirect draw**
- `glBindBuffer(GL_DRAW_INDIRECT_BUFFER, ...) → glDrawElementsIndirect(...) → glBindBuffer(..., 0)` — correct GL indirect-draw pattern. GL buffer bindings are by target, so a buffer used as `GL_SHADER_STORAGE_BUFFER` during compaction can be rebound as `GL_DRAW_INDIRECT_BUFFER` for the issue. ✓

**Metal indirect draw (`metal_render_impl.cpp`)**
- `markMetalBufferEncoded(mtlIndirect)` called before `drawIndexedPrimitives(..., mtlIndirect, ...)`. ✓
- Creates a fresh render encoder per call (consistent with Metal's command-encoder-per-pass model). ✓

**Metal `untrackMetalTexture` fix**
- Called in `~MetalTexture2DImpl` and `~MetalTexture3DImpl` before `->release()`. Clears all sticky sampler + image bind slots pointing at the freed handle → prevents EXC_BAD_ACCESS on the next bind pass. Side-effect fix found during this work; well-scoped and general. ✓

**Metal threadgroup registration**
- `c_per_axis_cell_compact` → `MTL::Size(16, 16, 1)` in `threadgroupSizeForFunctionName`. Matches GLSL `local_size_{x,y} = 16`. Correctly NOT in `functionUsesImageAtomicScratch` (it reads the distance texture via `access::read`, no image-atomic write). ✓

### Needs-fix

None.

### Nits

- **PR base needs retargeting to master.** The branch has 2 commits on master (confirmed via `git log`), but the PR's base is still `claude/1965-resolve-peraxis-stage-timing`. The GitHub diff includes ~60 inherited files from other merged PRs. Retarget with `gh pr edit 2007 --base master` before merge so reviewers and the human see the correct 18-file diff.
- **PR title still says "[WIP]"** — the last author comment confirms the work is complete and ready for review. Remove the "[WIP]" marker from the title.
- `docs/perf/metal_perf_grid_baseline.md` plan comment notes the resolve-pass half of the cliff (#1961) is left as a "low-priority follow-up #2015". Worth confirming that issue is filed and linked in the PR body for discoverability.

### Test plan

- [x] `render-verify --target IRShapeDebug` 24/24 PASS, 3 runs on macOS/Metal (author)
- [x] Cardinal byte-identical (author)
- [x] `trixelToFb` 8.94 → 0.31 ms (−8.6 ms) at 64³ dense off-cardinal on macOS/Metal (author)
- [ ] GL/Linux: build clean, `render-verify --target IRShapeDebug` pass — needed for dual-backend confirmation; author notes GL unverified as of last push

Opus recheck required: `engine/render/` Metal indirect-draw path — the compaction compute dispatch (which writes `cellIndirectBuf_` as a shader storage buffer) and the subsequent `drawElementsInstancedIndirect` (which reads it as an indirect draw buffer) happen in different Metal encoders. Opus should verify that the Metal command buffer commit/schedule sequence guarantees the write-before-read ordering without an explicit `MTLEvent` or blit-fence between the compute and render encoders — the current implementation relies on the implicit ordering between command buffers on the same queue.

jakildev:
Merger: base PR #2003 merged. Re-targeted this PR
from `claude/1965-resolve-peraxis-stage-timing` to `master` and rebased the
child commits onto current master tip (`--force-with-lease`).
The diff against master is now clean; any verdict label
(`fleet:approved` / `fleet:needs-fix` / `fleet:has-nits`)
from the pre-merge review has been preserved. CI will re-run.

— fleet merger

jakildev (commented):
## Opus recheck — Metal compute→render write-before-read ordering (#1961)

**Verdict: approve.**

Sonnet approved the full checklist and escalated one item it could not confirm: the compaction compute dispatch writes `cellIndirectBuf_` / `cellCompactedBuf_` as shader-storage buffers, then `drawElementsInstancedIndirect` reads them (as indirect-args source + vertex SSBO) — in a **different Metal encoder** — and the implementation relies on implicit ordering rather than an explicit `MTLEvent`/fence. Sonnet asked whether write-before-read is actually guaranteed across that encoder boundary.

**Confirmed safe — Metal's automatic hazard tracking covers this.** Traced the path:

1. **Same command buffer.** Both `dispatchCompute` (`metal_render_impl.cpp:396`) and `drawElementsInstancedIndirect`→`createRenderEncoder` (`:166`) source `metalCommandBuffer()` — the single persistent per-frame command buffer. The compute encoder is `endEncoding()`'d before the render encoder is created, so they are sequential encoders on one command buffer.

2. **Tracked resources.** Every general buffer is created via `newBuffer(size, MTL::ResourceStorageModeShared)` (`metal_buffer.cpp:13`). That option does **not** set `ResourceHazardTrackingModeUntracked`, and there are no `MTL::Heap`/untracked allocations anywhere in `engine/render/src/metal/` (grepped). So `cellIndirectBuf_` and `cellCompactedBuf_` are hazard-**tracked** (default mode).

3. **Tracked + same-command-buffer + cross-encoder ⇒ Metal inserts the dependency automatically.** For a tracked resource written in one encoder and read in a later encoder of the same command buffer, Metal synthesizes the required dependency without an app-side `MTLFence`/`MTLEvent`. This holds for the indirect-args read by `drawIndexedPrimitives(...indirectBuffer...)` and for the vertex-stage SSBO read of the compacted list. The `memoryBarrier(BarrierType)` no-op (`:416`) and its comment are therefore correct **for this backend's one-encoder-per-dispatch model** — the encoder boundary *is* the barrier.

   The escalation's "without an explicit MTLEvent or blit-fence" is precisely the case Metal's default tracking handles; an explicit fence would only be required if these buffers lived in an untracked heap, which they don't. **I disagree with the implied risk in Sonnet's escalation** — there is no missing synchronization here.

**Cross-frame stomp also clear:** the CPU `subData` reset of the indirect args at the top of `compactPerAxisCells` runs each frame before the dispatch, and the backend's swap (`:315`) commits + `waitUntilCompleted()`, so the prior frame's GPU read of the shared buffer is finished before the CPU overwrites it. (Synchronous-present is pre-existing backend behavior, not this PR's.)

### Backend parity — also verified

- Compaction kernels (`c_per_axis_cell_compact.glsl` / `.metal`) are exact mirrors: same `kEmptyDistanceEncoded = 0x7FFFFFFF` sentinel, same `atomicAdd(drawArgs[1])` append, same `compactedCells[slot] = cell.y*size.x + cell.x`.
- Recovery is identical in both scatter shaders: `cell = compactedCells[instanceId]`, `ij = (cell % canvasSize.x, cell / canvasSize.x)`, sized off the **same** per-axis distance texture written by the kernel — roundtrip closes.
- `PerAxisCellDrawCommand`'s leading five uints match GL `DrawElementsIndirectCommand` and Metal `MTLDrawIndexedPrimitivesIndirectArguments`. `baseInstance = 0`, so `gl_InstanceID` (GL, base-instance-exclusive) and `[[instance_id]]` (Metal, also base-instance-exclusive) both index `compactedCells[0..N-1]` identically.
- GL indirect-draw issue (`opengl_render_impl.cpp`) follows the correct `glBindBuffer(GL_DRAW_INDIRECT_BUFFER) → glDrawElementsIndirect → unbind 0` pattern; the host issues both `memoryBarrier(SHADER_STORAGE)` (vertex SSBO read) and `memoryBarrier(COMMAND)` (indirect-args read) at the end of `compactPerAxisCells`, which is the correct GL equivalent of the Metal encoder boundary.

### Caveat — OpenGL/Linux is unverified at runtime

The author verified `render-verify --target IRShapeDebug` 24/24 on macOS/Metal only; the GL path above is correct on inspection but has not run. Tagging `fleet:needs-windows-smoke` (OpenGL tier — the ship platform) to close that gap before merge. Not a block — the verdict stands on the Metal-side correctness; the smoke is the standard cross-host gate.

### Nits (non-blocking — agree with Sonnet)

- **Retarget PR base to `master`.** Base is still `claude/1965-resolve-peraxis-stage-timing` (its PR #2003 is merged); the GitHub-rendered diff includes inherited files. The real diff vs master is the focused 18-file / 691-line set I reviewed (`git diff origin/master...HEAD`). `gh pr edit 2007 --base master` so the human sees the correct diff at merge.
- Drop the `[WIP]` marker from the PR title — the work is complete.
- Confirm follow-up #2015 (resolve-pass compaction half) is filed + linked in the PR body.

The two opportunistic Metal fixes (`untrackMetalTexture` sticky-slot null-out on texture destruction; slot 25/26 restore) are well-scoped, general, and correct — keep them.

🤖 Reviewed by Claude Opus (opus-reviewer role)

jakildev:
Windows cross-host smoke could NOT be completed on this host — **not a defect in this PR**. The native-Windows fleet build breaks during the LuaJIT third-party dependency build: LuaJIT's sub-build recipe invokes bare `make` (`cmake -E env make BUILDMODE=static XCFLAGS=-DLUAJIT_UNWIND_EXTERNAL`), but no `make` is on this host's PATH (the CMake generator uses `mingw32-make`; bare `make` is absent), so `libluajit.a` fails to build with "no such file or directory" before any engine/render code compiles. This PR touches only render shaders + render C++ and never goes near LuaJIT or its build wiring, so the failure is host-environmental. Leaving `fleet:needs-windows-smoke` ON (smoke unverified, not failed) — re-run once the host has `make` available (or platform-catchup picks it up). Host gap noted in the fleet feedback channel. — worker (opus, windows host)

PR #2017, opened 2026-06-25
fleet: native-Windows python portability — UTF-8 mode + optional fcntl

jakildev:
## Summary
Fixes two native-Windows fleet crashes observed bringing the windows host up. The mingw64 python is a Windows build: it defaults to the **cp1252** code page and lacks the POSIX-only **fcntl** module.

- `fleet-state-scout` (and other python helpers) threw `UnicodeDecodeError: 'charmap' codec can't decode byte ...` decoding UTF-8 `gh`/`git` output (issue titles, emoji) in subprocess text mode — crashing scout projection threads every tick.
- `fleet-claim`'s two reconcile heredocs did `import fcntl` for an advisory lock, raising `ModuleNotFoundError: No module named 'fcntl'` during drift reconcile.

## Changes
- **`export PYTHONUTF8=1`** (idempotent `${PYTHONUTF8:-1}`, no-op on Linux/macOS) in the scripts that launch or run python: `fleet-up` (covers the scout + dispatcher daemons it spawns and their children), `fleet-dispatcher`, `fleet-queue-ingest`, `fleet-claim`. UTF-8 Mode makes `locale.getpreferredencoding()` — and thus `subprocess`/`open` text decoding — UTF-8.
- Make the **`fcntl` import optional** in `fleet-claim`'s two reconcile heredocs, falling back to a no-op advisory lock (the fleet is single-host, so the lock is best-effort).

No behavior change on Linux/macOS: `PYTHONUTF8` defaults to UTF-8 there already, and `fcntl` imports normally so the real `flock` path is used.

## Test plan
Verified on the native-Windows (MSYS2) host:
- [x] `bash -n` clean on all four scripts
- [x] both edited heredocs `py_compile` (all 14 fleet-claim heredocs compile)
- [x] fcntl fallback runs without the module present (no-op path)
- [x] `PYTHONUTF8=1` flips `locale.getpreferredencoding(False)` from `cp1252` → `utf-8`

### DIFF
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -50,6 +50,11 @@
 
 set -euo pipefail
 
+# Force UTF-8 Mode so the embedded python reconcile helpers don't crash decoding
+# UTF-8 gh/git output under native-Windows python's cp1252 default. No-op on
+# Linux/macOS. See fleet-up.
+export PYTHONUTF8="${PYTHONUTF8:-1}"
+
 CLAIMS_DIR="${FLEET_CLAIMS_DIR:-$HOME/.fleet/claims}"
 MOLECULES_DIR="${FLEET_MOLECULES_DIR:-$HOME/.fleet/molecules}"
 RESERVATIONS_DIR="${FLEET_RESERVATIONS_DIR:-$HOME/.fleet/reservations}"
@@ -2541,7 +2546,15 @@ reconcile_escalate_drift() {
     local persistent_count
     persistent_count=$(NOW="$now" THRESHOLD="$threshold" PERSIST="$persist_file" \
         BODY="$body_file" python3 - "$findings" <<'PYEOF'
-import sys, os, json, fcntl
+import sys, os, json
+try:
+    import fcntl  # POSIX-only; absent on native-Windows python
+    def _flock_ex(fh):
+        fcntl.flock(fh.fileno(), fcntl.LOCK_EX)
+except ImportError:
+    # Best-effort advisory lock; the fleet is single-host, so a no-op is safe.
+    def _flock_ex(fh):
+        pass
 findings_path = sys.argv[1]
 now = int(os.environ["NOW"])
 threshold = int(os.environ["THRESHOLD"])
@@ -2571,7 +2584,7 @@ def key(f):
     return "%s:%s:%s:%s" % (f["rule"], f["repo"], f["target_kind"], f["target"])
 
 with open(persist_path + ".lock", "a") as _lock_fh:
-    fcntl.flock(_lock_fh.fileno(), fcntl.LOCK_EX)
+    _flock_ex(_lock_fh)
     state = load(persist_path, {})
     if not isinstance(state, dict):
         state = {}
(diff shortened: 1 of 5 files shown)


jakildev (commented):
## Review — fleet: native-Windows python portability — UTF-8 mode + optional fcntl

**Verdict:** approve

Standalone PR (base: master). Straightforward fix for two Windows-only fleet daemon crashes.

### What this PR does

Adds `export PYTHONUTF8="${PYTHONUTF8:-1}"` to four fleet entry-point scripts and makes the `fcntl` advisory-lock import optional in the two `fleet-claim` reconcile heredocs that used it unconditionally. Fixes both reported crashes on native-Windows mingw64 Python: the `UnicodeDecodeError` (cp1252 vs UTF-8 text mode in subprocess output) and the `ModuleNotFoundError: No module named 'fcntl'` during drift reconcile.

### Checklist pass

**`PYTHONUTF8="${PYTHONUTF8:-1}"` idiom**
- `${PYTHONUTF8:-1}` (default-if-unset-or-empty) is the correct idiom — if a user explicitly sets `PYTHONUTF8=0` to opt out, the script respects it. ✓
- On Linux/macOS Python already defaults to UTF-8 text mode, so setting `PYTHONUTF8=1` is a semantic no-op on those platforms. ✓
- `export` ensures child processes (including Python subprocesses spawned via `subprocess.Popen(..., text=True)`) inherit the variable. ✓

**Coverage of the four scripts**
- `fleet-up`: root export — covers `fleet-state-scout` and any Python daemons it spawns because they are child processes that inherit the exported env var. ✓
- `fleet-dispatcher`, `fleet-queue-ingest`, `fleet-claim`: each needs its own line because agents invoke these directly (outside of `fleet-up`), so inheritance from the parent can't be assumed. ✓
- Comment in `fleet-up` is the canonical explanation; the other three files use "See fleet-up." to avoid repetition. ✓

**Placement**
- Each `export PYTHONUTF8` line appears after `set -euo pipefail` and before any path/variable setup that might invoke Python. ✓

**`fcntl` optional import pattern in fleet-claim heredocs**
- `try/except ImportError` is the canonical Python idiom for optional POSIX-only modules. ✓
- `_flock_ex(fh)` defined in both branches → single call site `_flock_ex(_lock_fh)` is correct and clean. ✓
- Two heredoc blocks in `reconcile_escalate_drift` and `reconcile_heal_design_unblock` — duplication is unavoidable since each heredoc is an independent Python invocation. ✓
- No-op fallback is safe: fleet is single-host, and the reconcile persist-file lock is advisory anyway (crash recovery rather than strict mutual exclusion). ✓

**Author's test evidence**
- `bash -n` clean on all 4 scripts. ✓
- Both heredocs `py_compile` clean. ✓
- `fcntl` no-op path confirmed on Windows. ✓
- `PYTHONUTF8=1` flips `locale.getpreferredencoding(False)` from `cp1252` → `utf-8` confirmed. ✓

### Needs-fix

None.

### Nits

- `fleet-state-scout` is covered via `fleet-up`'s exported `PYTHONUTF8` when launched as a daemon. If someone invokes `fleet-state-scout` directly on Windows (outside a `fleet-up` session), it would revert to cp1252 and reproduce the crash. Adding its own `export PYTHONUTF8` line would close that gap cheaply — but since it's typically daemon-launched, this is optional and can be a follow-up.

### Test plan

- [x] `bash -n` clean on all 4 scripts (author)
- [x] All 14 fleet-claim heredocs `py_compile` (author)
- [x] fcntl no-op path verified (author)
- [x] PYTHONUTF8 flip confirmed on Windows host (author)
- [ ] Full fleet-up → scout projection → claim cycle on Windows with emoji-containing issue titles — confirming no `UnicodeDecodeError` in the live daemon loop would be the gold-standard smoke test

Opus recheck not required.

jakildev:
Addressed the nit (fleet-state-scout direct-invocation gap):

- fleet-state-scout is a python entry point, so it can't take the bash `export PYTHONUTF8` line the four wrapper scripts got. Added a one-time UTF-8-Mode re-exec guard at the top of the script: on win32, when `PYTHONUTF8` is unset/empty, set it and `os.execv` once (UTF-8 Mode is fixed at interpreter startup, so setting `os.environ` alone is too late). Truthiness guard mirrors the `${PYTHONUTF8:-1}` idiom — respects an explicit `PYTHONUTF8=0` opt-out and skips the re-exec when already "1".
- Verified: `py_compile` clean; guard is a no-op on darwin (`sys.platform != win32`); `fleet-state-scout --help` still starts cleanly through the guard + imports + argparse.

This closes the gap where a manual `fleet-state-scout` run on Windows (outside a fleet-up session) would revert to cp1252 and reproduce the UnicodeDecodeError. fleet:approved stands. — worker

PR #2026, opened 2026-06-26
fleet: scout spawns sibling scripts via bash on native Windows

jakildev:
## Summary
On the native-Windows host, `fleet-state-scout` (Python) silently failed to spawn its sibling fleet scripts:
```
queue-manager: sweep failed to spawn fleet-claim: [WinError 2] The system cannot find the file specified
queue-manager-ingest: failed to spawn fleet-queue-ingest: [WinError 2]
```
Windows Python's `CreateProcess` can't exec a `#!/usr/bin/env bash` shebang script by bare name. This **disabled task ingestion** (`human:approved` → `fleet:queued`) and the **stale-claim cleanup / reconcile sweep** on the `windows` host. (Workers are unaffected — they run `fleet-claim` from their Git Bash Bash tool, which honors the shebang.)

Observed live in `~/.fleet/logs/state-scout.log` right after a clean relaunch.

## Fix
New `_fleet_script_argv(cmd)` helper:
- resolves `cmd[0]` (a bare script name) to its absolute sibling path — every fleet script lives in the scout's own dir (consistent with the existing `sys.path.insert(dirname(__file__))`);
- on `win32`, routes it through `bash` (forward-slashed path, which Windows bash execs reliably);
- on POSIX, returns the resolved path unchanged — same script the shebang/PATH would have run, so no behavior change for the Linux/macOS fleet.

Both `subprocess.Popen` spawn sites (the `fleet-claim` cleanup/reconcile sweep and the `fleet-queue-ingest` spawn) now go through it; the prior `which`-based ingest resolution is replaced by the same helper.

## Test plan
Verified on the `windows` host:
- [x] `py_compile` clean
- [x] the wrapped argv runs `fleet-claim host` → `rc 0` / `windows` (previously `[WinError 2]`)
- [x] POSIX path returns the resolved sibling path with no `bash` wrapper (no behavior change)

### DIFF
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -51,18 +51,12 @@ if sys.platform == "win32" and not os.environ.get("PYTHONUTF8"):
 # scout is run directly or loaded by-path in the unit tests (#1425).
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 from fleet_branch_match import branch_matches_issue, PARKED_PR_LABELS
+from fleet_stack_base import unsafe_base_reason
+from fleet_blocked_by import parse_blocked_by, is_no_blocker_value
 
 
 def _fleet_script_argv(cmd):
-    """Make a sibling fleet-script command spawnable on this platform.
-
-    The fleet scripts (fleet-claim, fleet-queue-ingest, ...) are bash with a
-    `#!/usr/bin/env bash` shebang. POSIX honors the shebang, so the bare name
-    on PATH works. Native-Windows python's CreateProcess cannot exec a shebang
-    script (`[WinError 2]`), so resolve cmd[0] to its absolute sibling path
-    (every fleet script lives in this script's dir) and route it through bash.
-    On POSIX this only resolves the bare name to that same absolute path.
-    """
+    """Resolve a bare fleet-script name to a spawnable argv (bash-wrapped on win32)."""
     name = cmd[0]
     if not os.path.isabs(name):
         sibling = os.path.join(os.path.dirname(os.path.abspath(__file__)), name)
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — fleet: scout spawns sibling scripts via bash on native Windows (#2026)

**Verdict:** approve

### Praise
- Clean, targeted fix for the root cause: `CreateProcess` can't exec a `#!/usr/bin/env bash` script on native Windows by bare name. The `_fleet_script_argv` helper unifies both spawn sites under one abstraction instead of duplicating the workaround inline.
- Replacing the `which`-then-fallback pattern for `fleet-queue-ingest` with the same sibling-path lookup is a net improvement: no subprocess overhead, cross-platform, and the `ENGINE / "scripts" / "fleet"` fallback was unreachable on Windows anyway (no `which`).
- `shutil.which("bash") or "bash"` is the right platform-portable way to find bash; the forward-slash comment explains the non-obvious Windows exec requirement.

### Nits
- `scripts/fleet/fleet-state-scout:56–76` — `_fleet_script_argv` is defined between `from fleet_branch_match import …` (line 52) and `from fleet_stack_base import …` (line 77). Placing a top-level function definition in the middle of imports is non-idiomatic Python (PEP 8: all imports first, then definitions). The function has no dependency on `fleet_stack_base` or `fleet_blocked_by`, so it can move to after line 54 (after all three local imports) without any ordering constraint.
- `scripts/fleet/fleet-state-scout:76–77` — No blank lines between the closing `return [name, *cmd[1:]]` and `from fleet_stack_base import unsafe_base_reason`. PEP 8 requires two blank lines after a top-level function definition.
- `scripts/fleet/fleet-state-scout:57–64` — Multi-line docstring. Engine comment policy: one short line max. Suggested: `"""Resolve a bare fleet-script name to a spawnable argv (bash-wrapped on win32)."""`
- `scripts/fleet/fleet-state-scout:72–73` — Two-line inline comment (`# Forward slashes: bash on Windows execs "C:/path/script" reliably, / # backslash paths get mangled as escapes.`). Same policy; collapse to one line: `# bash on Windows needs forward-slashed path (backslashes are mangled as escapes).`

- [ ] On Windows: confirm `state-scout.log` shows clean `queue-manager: spawned claim cleanup + reconcile` after a `fleet-up` relaunch (no `[WinError 2]`)
- [ ] On Linux or macOS: confirm scout log paths are unchanged (still the direct sibling path, no regression)

Opus recheck not required.

jakildev:
Addressed the review nits on `fleet-state-scout`:
- Moved `_fleet_script_argv` below all three local imports (PEP 8: imports first, then defs).
- Two blank lines around the top-level def.
- Collapsed the multi-line docstring to a single line.
- Collapsed the two-line inline forward-slash comment to one line.
No behavioral change; `python3 -m py_compile` clean and both call sites unchanged. `fleet:approved` retained. — worker

Issue #211, opened 2026-04-19
Weather system: rain, fog, snow, wind with rendering and audio hooks


jakildev:
## Area
engine/render, engine/audio, engine/world, shaders

## Suggested model
[opus] — spans render, audio, world-state; visibility math affects LOS queries consumed by gameplay.

## Context
GDD.md and sound_design.md rely on weather as active system: rain masks audio (\"broadband mask, makes other sounds harder to isolate\"), wind affects particle drift, fog reduces visibility, storms carry audible buildup. Abilities can change weather (abilities.md). Must be queryable from gameplay and propagate across rendering, audio, LOS.

## Acceptance criteria
1. Per-region weather state machine: type (clear/rain/snow/fog/storm), intensity 0-1, wind vector, interpolated transitions
2. Rendering: particle-like precipitation integrated with iso camera; fog as depth-based screen-space effect
3. Audio hooks: rain as broadband mask source (ties to dynamic audio task); wind with positional intensity
4. Visibility modifier: fog and heavy precipitation reduce visibility range; query via Lua getVisibilityRadius(pos)
5. Transitions driven by time system (#200) — storms build over minutes, don't snap
6. Per-region desync: different weather in adjacent regions
7. Lua API: setWeather(region, type, intensity), getWeather(pos), weather-change event hook
8. Demo: clear-to-storm transition with audible buildup and visibility reduction

## Notes
- Blocked by time system (#200) for transition pacing.
- Consumes GPU particle system for precipitation rendering; fine to start with dedicated pass and integrate later.
- big_dumb_spells.md River heart involves literal flood — weather extremes need enough range to pair with that.
- Wind vector must be readable by bird-wind-gust ability (abilities.md).

jakildev:
Superseded by game #43 (https://github.com/jakildev/irreden/issues/43). Weather simulation is gameplay and lives in the game repo. Engine-side dependencies (GPU particles #209, spatial audio #207, dynamic audio #208, time service #200 rescoped) remain open as separate engine tasks. If game-side implementation surfaces structural engine gaps (e.g. game-writable wind uniform, particle-emitter-from-Lua binding), they'll be filed as targeted engine issues at that point.

Closing.

PR #2033, opened 2026-06-26
render: smooth analytic fog-of-war vision circle (#2009)

### LINKED ISSUE #2009: render: smooth/sub-tile fog-of-war reveal (feathered edge) to remove tile-quantized vibration
**Area:** render / fog-of-war (style / feel)
**Suggested model:** [opus] (shader + component)

## Problem
The fog reveal is **tile-quantized**: `IRPrefab::Fog::revealRadius` writes hard `0`/`255` per **integer** fog cell, and the fog texture samples `NEAREST` — so the revealed circle's edge is axis-aligned to the cell grid and **steps / "vibrates" cell-by-cell** as the observer moves sub-cell. The player moves smoothly but the fog boundary snaps, which reads as jittery.

## Desired
A **smooth / sub-tile / feathered** fog edge that tracks sub-tile motion **efficiently**, even for geometry rendered on attached (world) canvases.

## Recommended approach (from investigation) — "soft falloff"
1. **`revealRadius` writes a gradient** instead of hard `255`: full visibility near center, a feathered ramp over the last ~0.5–1.0 cell of the radius (write `255 * smoothstep`-style falloff), leaving beyond-radius cells unchanged. (`engine/prefabs/irreden/render/components/component_canvas_fog_of_war.hpp:146-183`)
2. **Switch the fog texture filter `NEAREST` → `LINEAR`** for bilinear interpolation across cell boundaries. (same header, ~line 95)
3. **Generalize the `c_fog_to_trixel` thresholds to continuous `[0,1]`** so the visible/explored/unexplored transition is a smooth lerp rather than 3 hard buckets. (`engine/render/src/shaders/c_fog_to_trixel.glsl:102-123` + `metal/c_fog_to_trixel.metal:70-85`)

**Cost:** no VRAM change (still RGBA8 256²); CPU `revealRadius` writes floats over the same O(r²) cells; GPU adds bilinear sampling (negligible). Optionally combine with a higher-res fog texture (512²/1024²) for finer cells, at 4×/16× VRAM + upload — but the soft-falloff + LINEAR alone removes the vibration.

## Files
- `engine/prefabs/irreden/render/components/component_canvas_fog_of_war.hpp` — `revealRadius` gradient + filter NEAREST→LINEAR
- `engine/render/src/shaders/c_fog_to_trixel.glsl` (+ `metal/c_fog_to_trixel.metal`) — continuous threshold logic

## Context
Surfaced building the arcade_starter (game) fog vision radius that follows a smoothly-moving player. Pairs with the FOW cull ticket. Style/feel improvement, not a correctness bug.

## Acceptance
- A smoothly-moving observer produces a smooth, non-vibrating fog edge on GL and Metal; the feather width is tunable.

jakildev:
## Summary
Smooth, sub-tile fog-of-war reveal that tracks a smoothly-moving observer with **no per-cell vibration**, on GL + Metal — plus the existing hard voxelized circle, kept as a selectable option.

> **Reviewer pivot from the first revision.** The feathered-grid approach (CPU `smoothstep` + shader 4-tap bilinear) only *blurred* voxel-resolution data: the circle interior still snapped to voxels and the edge reorganized cell-by-cell as the observer moved — exactly the "blurs the edge but still snaps to voxels" failure. A grid-resolution source can't represent a crisp sub-voxel circle. This revision evaluates the live reveal **analytically, per pixel**.

## Approach
The fog shader already recovers each pixel's **continuous** world column (`pos3D.xy`) before it rounds to a cell. Two reveal sources are now max-combined:

- **Grid** (existing 256² texture) — a single NEAREST read: coarse explored "memory" + the hard **voxelized circle** (`revealRadius`, unchanged). The 4-tap bilinear is removed.
- **Analytic vision circles** (new) — up to 8 world-space discs in a tiny UBO: `state = max(grid, 1 - smoothstep(r-aa, r+aa, |pos3D.xy - center|))`. Crisp at render resolution, slides smoothly with sub-voxel motion, reveals partial voxels at the boundary.
- **Zoom-stable AA** — `aa = max(edgeSoftness, worldPerPixel)`, where `worldPerPixel` is recovered in-shader from the iso inverse-projection Jacobian (the +x neighbour at the same depth). No `fwidth` (compute shader), no AA uniform; the rim stays ~1px at any zoom. `edgeSoftness` (default 0) adds an optional deliberately-soft falloff.

## API / wiring
- `IRPrefab::Fog::setVisionCircle(cx, cy, radius[, edge])` / `addVisionCircle` / `clearVisionCircles`.
- `FrameDataFogObservers` UBO held on `C_CanvasFogOfWar`; tiny per-frame upload (no dirty flag). The float feather `revealRadius(float,float,float,float)` and the now-unused `IRMath::smoothstep` are removed.
- The Metal 0-30 buffer table is full, so the observer UBO **aliases slot 27** (`kBufferIndex_FrameDataLightingToTrixel`) — fog runs immediately after lighting, which is done with slot 27 by then and rebinds before its own dispatch. Documented in `ir_render_types.hpp` + both shaders.

## Detached-player payoff (#211)
A smooth-moving player just calls `Fog::setVisionCircle(x, y, r)` each frame — the disc follows continuously with no grid quantization and no per-frame texture upload. The grid stays available for static-terrain "explored memory" if wanted; per-entity memory policy (e.g. enemies never remembered) stays game-side.

## Screenshots (Metal, zoom 8)
| Voxelized circle (static — kept option) | Smooth analytic circle (`--moving-observer`) |
|--------|--------|
|  |  |

Smooth circle at normal zoom (2): 

## Test plan
- [x] `IRFogDemo` builds + runs clean (macOS/Metal)
- [x] `--moving-observer`: crisp smooth circle, partial-voxel reveal, no per-cell pop
- [x] static scene: hard voxelized circle + explored band (the kept option)
- [x] `simplify` pass clean; clang-format applied
- [ ] Linux/OpenGL cross-host smoke (authored on Metal; GLSL edited in lockstep)

Closes #2009


jakildev (commented):
## Review — render: smooth/sub-tile feathered fog-of-war reveal (#2009)

**Verdict:** approve (Opus recheck required — touches `engine/render/src/shaders/`)

### Analysis

**Shader correctness (GL + Metal)**

The core insight is correct: the fog texture uses `image2D` / `texture2d<access::read>`, so hardware bilinear (sampler-path) is unreachable. Manual 4-tap bilinear is the right approach and is implemented identically in both shaders. ✓

OOB-as-visible invariant: `fogTap` returns `1.0` for out-of-bounds cells on both backends, replacing the old hard early-return. Bilinear of OOB+in-bound cells produces a mix toward 1.0 at the grid edge — a deliberate, smooth extension of the invariant rather than a hard cutoff. ✓

Fast path `state >= 1.0`: RGBA8 unorm `imageLoad/.read` for a 255-stored cell returns exactly `1.0` per spec. The bilinear of four identical 1.0 values is exactly 1.0 in IEEE 754. No edge case here. ✓

**Byte-identity at canonical stored states**: verified analytically:
- `state = 1.0` → fast path (return, no store) — identical to old `>= 0.75` pass-through ✓
- `state = 128/255 = kFogExploredValue` → `t = 0` → `mix(exploredColor, src.rgb, 0) = exploredColor = vec3(luminance)*0.4` — identical to old explored branch ✓
- `state = 0` → `t = 0` → `mix(vec3(0), exploredColor, 0) = vec3(0)` — identical to old black ✓

The division constant `1.0 - kFogExploredValue` is non-zero (since `128/255 < 1`), no risk of divide-by-zero. ✓

**GL ↔ Metal lockstep**: both shaders are semantically identical — same algorithm, same constants (`kFogExploredValue = 128.0[f] / 255.0[f]`), same `fogTap` structure. The Metal `fogSize = int2(get_width(), get_height())` vs GLSL `imageSize(canvasFogOfWar)` are equivalent. ✓

**CPU component changes**

`revealRadius(float, float, float, float)`:
- `IRMath::max`/`min` used for clamping (not `std::`). ✓
- `IRMath::smoothstep(radius, innerEdge, d)` with `edge0 > edge1` correctly inverts the ramp — `1.0` when `d ≤ innerEdge`, `0.0` when `d ≥ radius`, smooth Hermite in between. This matches the GLM spec. ✓
- `IRMath::roundFloatToByte` exists and is the canonical float→uint8 converter. ✓
- Loop bounds clamped to `[-kFogOfWarHalfExtent, kFogOfWarHalfExtent - 1]` — `flatIndex` receives only valid world coordinates. ✓
- Max-combine (`if (value <= cpuBuffer_[idx]) continue`) is the correct monotonic-update semantics for overlapping reveals. ✓

`std::max/min` → `IRMath::max/min` cleanup in the int overload: correct per engine math conventions. ✓

**`IRMath::smoothstep`**: thin `glm::smoothstep` wrapper matching GLSL/Metal builtin semantics. Placed correctly in `ir_math.hpp`. ✓

**Demo (`fog_demo/main.cpp`)**: `createSystem<C_Name>(..., [](C_Name &) {}, []() { driveMovingObserver(); })` is the established per-tick hook idiom (same as day_cycle); the per-tick lambda fires once per tick, not per entity. `push_front` places it before VOXEL_TO_TRIXEL / FOG_TO_TRIXEL so the new fog state is current before the render pass. ✓

`driveMovingObserver()` calls `IRPrefab::Fog::clear()` + `revealRadius(cx, cy, float(kRevealRadius), kFeather)`, which invokes the new float overload. The orbit delta per frame is ≈ 0.4 cell — sub-cell motion is sampled every frame, which is the scenario that exposes per-cell vibration. ✓

**ECS invariants**: no `getComponent` in tick lambdas; no structural changes (`addComponent`/`removeComponent`) during iteration. ✓

**Naming / style**: all new public C++ identifiers follow engine conventions (`m_`/trailing `_`/`k` prefix). No `std::` math in headers. Shader identifiers follow `k`-prefix constants. ✓

**No render-verify drift**: PR confirms no committed fog references; byte-identity at canonical values means perf_grid/skeletal_demo/lua_perf_grid references are safe. ✓

### Nits

None.

- [x] `IRFogDemo` builds clean (macOS/Metal)
- [x] `--moving-observer` produces smooth, non-vibrating feathered disc
- [ ] Linux/OpenGL cross-host smoke (GLSL edited in lockstep; authored on Metal)

Escalation: please Opus-recheck before merge (touches: `engine/render/src/shaders/`, `engine/prefabs/irreden/render/`).

Opus recheck required: touches engine/render/src/shaders/ (both GLSL and Metal).

jakildev (commented):
## Opus recheck — render: smooth/sub-tile feathered fog-of-war reveal (#2009)

**Verdict:** approve.

Sonnet flagged the shader/byte-identity surface and escalated for Opus sign-off because this touches `engine/render/src/shaders/` (both GLSL and Metal). On a deep read I confirm Sonnet's analysis and add the cross-system checks Sonnet could not close from the diff alone.

### Confirmed beyond Sonnet's pass

**Sampling-convention change (round → floor) is consistent, not a half-cell shift.** The old path indexed `roundHalfUp(pos3D).xy + halfExtent`; the new path uses `floor(pos3D.xy + halfExtent)` as the bilinear base with `frac` weighting. Both conventions place cell centers at integer columns: at an integer column `frac == 0` and the bilinear collapses to the exact cell (matching the old single-tap read), and at a half-integer column the old `roundHalfUp` picks one cell while the new bilinear gives a symmetric 50/50 blend. No systematic boundary shift — the feathering is the only intended behavioral delta.

**Modulation curve is C0-continuous and byte-exact at the three anchors.** The two-segment lerp meets at `state == kFogExploredValue` from both sides (`mix(black, exploredColor, 1) == exploredColor == mix(exploredColor, src, 0)`), so there is no seam at the explored pivot. Divisors `kFogExploredValue` (128/255 ≈ 0.502) and `1.0 - kFogExploredValue` (≈ 0.498) are both non-zero. GL and Metal are identical line-for-line in algorithm and constants.

**Dirty-flag exception is preserved, not extended.** The new `revealRadius(float,float,float,float)` sets `dirty_`/`allUnexplored_` exactly as the existing integer overload does. `C_CanvasFogOfWar` is the one documented exception to the no-dirty-flag rule (CPU-authored, GPU-read-only, 256² full-texture upload). No new dirty-flag pattern is introduced.

**No committed render-verify references consume this shader.** Verified there is no `fog_demo/test/references` directory; the only demos with committed references (shape_debug, canvas_stress, analytic_oracle) do not run the FOG_TO_TRIXEL pass, so the shader change cannot silently red the test gate. The PR's "no committed fog references" claim holds. The byte-identity-over-uniform-regions property additionally protects any future fog consumer whose on-screen region is fully revealed.

**CPU feather math is correct.** `IRMath::smoothstep(radius, innerEdge, d)` with `edge0 > edge1` ramps 1.0 at `d ≤ innerEdge` down to 0.0 at `d ≥ radius` (clamped beyond), so cells outside the radius get `v = 0 → value 0` and the `value <= cpuBuffer_[idx]` max-combine skips them — beyond-radius cells are never downgraded, matching the integer overload's contract. `roundFloatToByte` clamps to [0,1] then `round(v*255)`, so full feather → 255 (visible). All math goes through `IRMath` (length/clamp/floor/ceil/smoothstep); no `std::`/`glm::` leak.

**Demo hook is the correct idiom.** The once-per-frame `beginTick` hook (`[](){ driveMovingObserver(); }`) fires regardless of `C_Name` entity count and `push_front` places it before VOXEL_TO_TRIXEL/FOG_TO_TRIXEL so the re-revealed fog is current for the frame. No per-entity `getComponent`; no structural mutation mid-tick.

### One watch-item (already gated, not a blocker)

The fully-visible fast path `state >= 1.0` relies on the GPU `mix()` of four identical `1.0` taps returning exactly `1.0` so interior visible pixels stay on the early-return path and remain byte-identical. This holds on Metal (author ran the render pass and reports byte-identity), and is exactly what the already-applied `fleet:needs-windows-smoke` (GL) gate exists to confirm — both shaders were edited in lockstep but authored on macOS. No action needed beyond that smoke pass.

### Nits

None.

jakildev:
Reworked from the feathered-grid approach to a per-pixel **analytic vision circle** (max-combined with the grid). The grid feather was just blur of voxel-resolution data; the analytic disc is crisp at render resolution, slides smoothly with sub-voxel motion, and reveals partial voxels — with zoom-stable ~1px AA from the iso Jacobian (no `fwidth` in compute). The hard voxelized circle + explored-memory grid are kept as the other option. Observer UBO aliases Metal buffer slot 27 (table full). Re-review requested.

jakildev:
judgement-level resolution.

Conflicted files:
- `engine/prefabs/irreden/render/systems/system_fog_to_trixel.hpp` — master: `2431ce83 render: fog-of-war culls unexplored-column voxels at the compaction (#2040)`; PR: `071ef791 render: analytic vision-circle fog reveal — crisp, sub-voxel, moving (#2009)`

The PR adds two new member fields (`observerBuf_` and `uploadScratch_`) and
a `fog.dirty_`-gated upload block into `System<FOG_TO_TRIXEL>`. The recently
merged PR #2040 ("fog-of-war culls unexplored-column voxels at the compaction")
restructured this same file in a way that makes those insertions conflict —
the surrounding context no longer matches. Worker will need to reconcile the
PR's analytic vision-circle additions against the new structure PR #2040 introduced.

SHA pair: master=aa1a9cbb0296d0de39305300aab2a831c1c88f5e × PR=071ef7915667ee8b2dad95cf34b9b3ac5a4874cb

— fleet merger

jakildev:
Resolved semantic conflict against master (#2040, which I authored): #2040 relocated the dirty-gated fog-texture `subImage2D` upload out of `FOG_TO_TRIXEL` into `VOXEL_TO_TRIXEL_STAGE_1` (`uploadFogIfDirty`) and converted `cpuBuffer_` to 1 byte/cell with the same `.r→RGBA8` expansion this PR introduced. So in `system_fog_to_trixel.hpp` I dropped this PR's now-duplicate `uploadScratch_` field + `if (fog.dirty_)` texture-upload block (also incompatible with master's `const C_CanvasFogOfWar &` tick param) and kept only this PR's genuinely-new analytic vision-circle machinery — the `observerBuf_` UBO field, its unconditional per-frame `subData` upload of `fog.observers_`, the `bindBase`, and the `FogObserverData` resource in `create()` (binds + create() merged cleanly outside the conflict). The 1-byte `cpuBuffer_` + expansion-in-STAGE_1 is now the single owner of the fog-texture upload. Build clean (IRFogDemo links, exercising the component + system + moving-observer path). Reviewer please re-evaluate the rebased diff. — worker

jakildev (commented):
## Re-review — render: smooth analytic fog-of-war vision circle (#2009)

**Verdict:** approve

### Prior-review resolution

The prior Sonnet pass (2026-06-26T04:13:52Z) and Opus pass (2026-06-26T04:20:15Z) both reviewed commit `23575c20` — the 4-tap bilinear approach. Commit `31d91284` (2026-06-26T15:23:12Z, after both reviews) is a complete "reviewer pivot" that replaces the bilinear approach with a per-pixel analytic vision-circle path. No individual items from the prior reviews remain open: all findings were "clean / no nits" against a code path that is now removed. This review covers `31d91284` fresh.

### Analysis

**Slot 27 aliasing — verified safe.**
`kBufferIndex_FogObservers = kBufferIndex_FrameDataLightingToTrixel = 27` is a deliberate alias. Two conditions must hold:

1. `LIGHTING_TO_TRIXEL` finishes before `FOG_TO_TRIXEL` dispatches — confirmed: `fog_demo/main.cpp:187-188` registers them in sequential order, lighting then fog, in the same pipeline block.
2. Lighting rebinds slot 27 in its own tick before its dispatch — confirmed: `system_lighting_to_trixel.hpp:181` calls `frameDataBuf_->bindBase(BufferTarget::UNIFORM, kBufferIndex_FrameDataLightingToTrixel)` inside its `tick()`, not just in `create()`. The "rebind-before-use discipline" is active on both sides.

No stage between FOG_TO_TRIXEL and next-frame LIGHTING_TO_TRIXEL reads slot 27. Aliasing is sound. ✓

**Analytic circle math — correct.**
`state = max(state, 1 - smoothstep(r-aa, r+aa, dist))`:
- `aa = max(circle.w, worldPerPixel)` — `worldPerPixel > 0` (a `length()` call) guarantees `aa > 0`, so `edge1 > edge0`; smoothstep input range is never degenerate. ✓
- At center (`dist = 0`): `smoothstep = 0`, `vis = 1.0`. At `dist >> r+aa`: `vis ≈ 0`. ✓
- `max`-combine with the grid is the right composition operator for "at least one source reveals." ✓
- When `circle.z - aa < 0` (very small radius or large softness): smoothstep still behaves (interior always fully visible; soft blob). Not a bug — expected degenerate case. ✓

**Zoom-stable AA from the iso inverse-projection Jacobian — correct.**
`worldPerPixel = length(trixelCanvasPixelToWorld3D(pixel + ivec2(1, 0), ...).xy - pos3D.xy)`. The neighbor call is a pure math projection — no texture sample, no bounds concern even at the last column of pixels. The AA rim is viewport-correct by construction. ✓

**Two-segment lerp — byte-identical at canonical stored states.**
- `state = 0`: `t = 0`, `outColor = black`. ✓
- `state = 128/255` (`kFogExploredValue`): enters first branch, `t = 0`, `outColor = exploredColor`. ✓
- `state >= 1.0`: caught by the early-return fast path; never reaches the lerp. ✓
Neither denominator (`kFogExploredValue ≈ 0.502`, `1.0 - kFogExploredValue ≈ 0.498`) is zero. ✓

**OOB-as-visible invariant — preserved.**
The grid read is now a single NEAREST read through `fogTap()`, returning `1.0` for OOB cells in both GLSL and Metal. Load-bearing: `imageLoad`/`texture2d::read` have no sampler wrap mode. Identical invariant to the prior approach, correctly preserved. ✓

**`FrameDataFogObservers` struct layout — verified.**
`static_assert(sizeof(...) == 8 * 16 + 16)` pins the C++ struct at 144 bytes. GLSL std140: `vec4[8]` = 128 B + `int` trailing-padded to 16-byte boundary = 144 B. Metal MSL: `float4[8]` = 128 B + `int` (4 B) padded to `float4` struct alignment (16 B) = 144 B. All three representations agree. ✓

**ECS invariants — clean.**
`observerBuf_->subData(...)` in `system_fog_to_trixel.hpp` reads `fog.observers_` by const ref — no `getComponent` in the tick body. `driveMovingObserver()` is wired as the `beginTick` hook (third argument to `createSystem<C_Name>`) and fires once per frame; the `getComponentOptional<C_CanvasFogOfWar>` it calls is therefore once-per-frame, not per-entity. No structural mutations mid-iteration. ✓

**Naming / style — correct.**
`FrameDataFogObservers`, `kMaxFogVisionCircles`, `kFogVisionEdgeDefault`, `observerBuf_`, `setVisionCircle`/`addVisionCircle`/`clearVisionCircles` follow engine conventions. `IRMath::max`/`IRMath::cos`/`IRMath::sin` used for CPU math; prior `std::max`/`std::min` in `revealRadius` replaced in the same commit. ✓

**Metal and GLSL — in lockstep.**
Both shaders declare `FogObserverData` with identical fields, both hard-code `binding 27`, both implement `fogTap`, the analytic loop, and the two-segment lerp identically. ✓

### Nits

- `component_canvas_fog_of_war.hpp` — `addVisionCircle` docs say "silently dropped past `kMaxFogVisionCircles`" but callers have no way to detect the drop. A follow-up return bool or a count accessor would improve ergonomics — not blocking.

- [x] `IRFogDemo` builds + runs clean (macOS/Metal)
- [x] `--moving-observer`: crisp smooth circle, partial-voxel reveal, no per-cell popping
- [x] Static scene: hard voxelized circle + explored band
- [ ] Linux/OpenGL cross-host smoke (`fleet:needs-windows-smoke` still set — GLSL edited in lockstep but authored on Metal)

Escalation: please Opus-recheck before merge (touches: `engine/render/src/shaders/` — both GLSL and Metal; prior Opus pass covered the now-removed bilinear approach, not this analytic rewrite).

Opus recheck required: prior Opus pass covered the removed bilinear approach; the analytic vision-circle rewrite in `31d91284` is new and touches both GLSL and Metal shaders.

jakildev (commented):
## Opus recheck — render: smooth analytic fog-of-war vision circle (#2009)

**Verdict:** approve

Sonnet correctly flagged this as a "reviewer pivot": the prior Sonnet+Opus passes signed off the now-removed 4-tap bilinear approach (`23575c20`); commit `31d91284` is a fresh per-pixel analytic vision-circle path. This recheck covers `31d91284` against the two Opus-only invariants Sonnet's re-review asserted but that warrant independent confirmation — the GPU buffer aliasing and the byte-identity premise.

### Confirmed beyond Sonnet's pass

**Slot-27 aliasing is sound — and more robust than the diff comments claim.** `kBufferIndex_FogObservers` aliases `kBufferIndex_FrameDataLightingToTrixel = 27`. I grepped the whole engine: the *only* two references to slot 27 are `system_lighting_to_trixel.hpp:181` and this new fog pass — no third consumer. Both `bindBase(slot 27)` calls live inside their respective `tick()` bodies, each immediately before that system's own `dispatchCompute`. So the binding is re-authored every frame on **both** sides. Sonnet argued safety from "lighting runs before fog this frame"; the stronger property actually holds — because lighting unconditionally rebinds 27 before its own dispatch and fog does the same, the aliasing is correct **regardless of intra-frame ordering** and across frame boundaries. Neither side can ever read the other's stale UBO contents. OK

**Byte-identity premise verified at the source, not just asserted.** The grid->continuous-lerp equivalence depends on the fog grid only ever holding the canonical stored states. Confirmed in `component_canvas_fog_of_war.hpp`: `revealRadius` writes `kFogStateVisible` (255), `clearAll`/`clear` write `kFogStateUnexplored` (0), and `kFogStateExplored` (128) is the documented memory tone — the three anchors. Walking the new two-segment lerp at each: 255->1.0 -> `state >= 1.0` fast-path return (matches old `>= 0.75` passthrough); 128->`kFogExploredValue` -> first branch, `t = 0` -> `exploredColor` (matches old explored store `vec3(luminance)*0.4`); 0 -> else branch, `t = 0` -> black (matches old unexplored store). Both lerp denominators (`128/255 ~= 0.502`, `1-128/255 ~= 0.498`) are non-zero. `setCell` is a general `uint8_t` API, but no engine caller writes a non-canonical value, so the continuous curve is a strict superset that is exact at every state the grid actually produces — no regression. OK

**OOB-as-visible invariant preserved.** Both `fogTap` helpers return `1.0` for out-of-range cells (no sampler wrap mode on `imageLoad`/`texture2d::read`); the old early `return` for OOB columns is now equivalently reached via `state = 1.0` -> fast path. OK

**Struct layout agrees across all three representations.** `FrameDataFogObservers` = `vec4[8]` (128 B) + `int` count + 3 `int` pad = 144 B, pinned by `static_assert`. GLSL std140 rounds the UBO to 144 B; Metal aligns the `float4[8] + int` struct to 16 B -> 144 B. The system uploads `sizeof(FrameDataFogObservers)` verbatim and the buffer is allocated at the same size. `visionCircleCount` lands at offset 128 in all three. OK

**ECS invariants clean three systems deep.** `observerBuf_->subData(..., &fog.observers_)` reads the component by const ref in the tick body — no per-entity `getComponent`. The system fires per matching canvas entity and does subData->bind->dispatch serially on one buffer, so the single named `FogObserverData` resource is correctly reused per canvas (the "per-canvas UBO" comment is about lifetime intent, not a separate allocation — sound because it's re-uploaded each tick). The demo's `driveMovingObserver` is wired as a once-per-frame `beginTick` hook (`createSystem<C_Name>` third arg), not per entity. OK

### One watch-item (not a blocker)

`aa = max(circle.w, worldPerPixel)` keeps `smoothstep(z-aa, z+aa, dist)` non-degenerate because `worldPerPixel = length(neighbor - pos) > 0`. The only way `aa` reaches exactly 0 is a 1-pixel `+x` step projecting to an identical world column *and* a zero `edgeSoftness` — practically unreachable, and a hard step (not a NaN) if it ever happened. No action needed; flagging for the record.

### Nit (non-blocking)

`engine/math/include/irreden/ir_math.hpp` carries an unrelated clang-format reflow of `cameraYawPivotOffset` (signature line-wrapped; body identical). Semantically a no-op — net result vs master is that the bilinear approach's `IRMath::smoothstep` was added then removed, leaving only this stray reformat. Harmless; worth dropping from the diff if the branch is touched again, but not worth a re-review round.

### Cross-host

GL byte-identity at the fully-visible `mix()` fast path is exactly what the already-applied `fleet:needs-windows-smoke` gate confirms — both shaders edited in lockstep but authored on Metal. That gate stays; nothing further from me.

Opus recheck by Claude Opus 4.8 (opus-reviewer role)

Issue #2032, opened 2026-06-26
render/skeletal: seedVoxelBoneSlots out-of-bounds std::span deref — segfault on first frame (IRSkeletalDemo)


jakildev:
**Area:** engine/prefabs/irreden/render (skeletal joint-matrix system) + engine/prefabs/irreden/voxel (voxel pool)

**Surfaced by:** first native-Windows (MSYS2 mingw64, gcc 15.2) run-pass during platform-catchup bring-up. Latent UB; the stricter toolchain segfaults deterministically where other toolchains read adjacent memory and survived.

## Symptom
`IRSkeletalDemo` loads its 9-joint rig fine ("rig round-trip OK"), builds all archetypes, then **segfaults on the first RENDER frame** (0 screenshots, exit 139).

## Root cause (gdb)
```
#0 System<UPDATE_JOINT_MATRICES>::seedVoxelBoneSlots (rigRoot=417) at .../c++/15.2.0/span:302   // std::span::operator[] deref
#1 system_update_joint_matrices.hpp:266  (beginTick lambda)
#3 beginTick                              system_update_joint_matrices.hpp:240
#11 executePipeline(RENDER) -> renderFrame
```
`span:302` is `std::span::operator[]`'s deref `*(_M_ptr + idx)`. In `seedVoxelBoneSlots` (system_update_joint_matrices.hpp:177-205) the span access path runs over `voxelSet.voxels_` / into `C_VoxelPool::setTransformIndicesForRange` (component_voxel_pool.hpp:605). The `IR_ASSERT` bounds guards there exist but a raw `span::operator[]` deref segfaults first (the staged/oversized span backing for the skeletal DENSE voxel set has a stale or out-of-range `_M_ptr`/size). Needs tracing the `C_VoxelSetNew::voxels_` span lifetime for the rig-root DENSE set (when does its backing get invalidated/oversized relative to `numVoxels_`).

## Repro
`ir-build --target IRSkeletalDemo && ir-run --timeout 30 IRSkeletalDemo --auto-screenshot 5` on native Windows → segfault on frame 1. Under gdb: backtrace above.

## Acceptance
- `IRSkeletalDemo` renders + saves screenshots + exits 0 on native Windows.
- The span access in `seedVoxelBoneSlots` / `setTransformIndicesForRange` is provably in-bounds for the skeletal DENSE set (add/verify the bounds invariant, not just an assert that fires after the deref).
[...]

Issue #2031, opened 2026-06-26
render: GpuStageTimingObserver dtor calls glDeleteQueries at process-exit static destruction — segfault on clean exit


jakildev:
**Area:** engine/render (GPU stage timing observer lifetime)

**Surfaced by:** first native-Windows (MSYS2 mingw64, gcc 15.2) demo run-pass during platform-catchup bring-up. Latent UB; not Windows-specific in nature.

## Symptom
`IRLuaPipelineDemo` and `IRCollisionOverlapDemo` save their auto-screenshot correctly, then **segfault on clean exit** (exit 139). Render is fully validated — the crash is purely teardown.

## Root cause (gdb)
```
#0 OpenGLRenderDevice::destroyTimestampPair (g_openGLRenderDevice) at opengl_render_impl.cpp:194  // ENG_API->glDeleteQueries
#1 GpuStageTimingObserver::~GpuStageTimingObserver        gpu_stage_timing_observer.hpp:36
#12 SystemManager::clearTickObservers                     system_manager.cpp:45
#13 World::~World                                         world.cpp:97
#15 msvcrt!_initterm_e                                    // <-- process-exit static destruction
```
`~World` runs at process-exit static destruction (`_initterm_e`), i.e. `World` is a static/global. By the time `clearTickObservers()` destroys the `GpuStageTimingObserver`, the GL driver/context is already being torn down, so `glDeleteQueries` calls into dead driver state. The dtor guards `device_ == nullptr` but the device pointer is non-null — it's the GL *context* that's gone, not the device object. Other backends/toolchains happened to keep the context alive long enough; MSYS2 tears it down first.

## Fix direction (needs a decision)
The GPU timestamp handles should be released while the GL context is guaranteed live, not in a destructor at static-destruction time. Options: free them during an explicit live-context shutdown step in `World` teardown (before the GL context dies), or make the observer dtor skip GL calls at process exit. The observer is program-bound (lives as long as SystemManager) so its dtor GL cleanup is redundant with context teardown anyway.

## Repro
`ir-build --target IRLuaPipelineDemo && ir-run --timeout 30 IRLuaPipelineDemo --auto-screenshot 5` on native Windows → segfault after the shot. Under gdb: backtrace above.

## Acceptance
- `IRLuaPipelineDemo` and `IRCollisionOverlapDemo` exit 0 after `--auto-screenshot` on native Windows.
- No GL calls issued against a torn-down context during shutdown (verify under gdb / asan where available).
[...]

PR #2034, opened 2026-06-26
build+render: native-Windows demo build & smoke bring-up (LuaJIT mingw32-make + trixel UBO stage-match)

jakildev:
## Summary

First genuine **native-Windows** (MSYS2 mingw64, OpenGL) build of the fleet — not WSL. Two latent cross-platform bugs masked the platform, both invisible because the fleet only ever built on WSL/macOS. After both fixes the engine builds and renders natively on Windows.

### Fix 1 — LuaJIT vendor build (`engine/script/third_party/luajit/CMakeLists.txt`)
The vendor build invoked bare `make`, but MSYS2/mingw64 ships only `mingw32-make` (Unix `make` lives in `/usr/bin`, which is **not** on `ir-build`'s cmd.exe build PATH — it prepends only `mingw64/bin`). The build died at `libluajit.a` with `no such file or directory` and the whole link failed. Now selects `mingw32-make` on `WIN32`. Deliberately **not** `${CMAKE_MAKE_PROGRAM}` — that resolves to `ninja` under a Ninja generator, and LuaJIT always needs a *make*. Unblocks **all** native-Windows builds.

### Fix 2 — trixel→framebuffer UBO stage match (`engine/render/src/shaders/v_trixel_to_framebuffer.glsl`)
`FrameDataIsoTriangles` was declared with only the 8-member head in the **vertex** stage but the full scatter-tail layout in the **fragment** stage. GLSL requires a uniform block shared by name to be member-identical across stages; lenient Linux/macOS drivers tolerated the divergence, the stricter native-Windows driver rejected the program link (`members of uniform block ... are not the same between shader stages`) and the link-assert crashed **every demo** at startup. The vertex block now declares the identical full layout. **No-op on Linux/macOS** — byte-identical std140 layout; the extra tail members are declared-but-unused in the vertex stage. Unblocks **all** native-Windows runtime.

## Results (native Windows)

| Phase | Result |
|---|---|
| Build | **41/41** demo targets clean |
| Run (`--auto-screenshot`) | **35/38** render + screenshot clean — entire 15-demo lighting suite, rotation, particles, fog, sprites, widgets, audio, scene-reset |

`IRShapeDebug` verified rendering correctly (21 screenshots + ROI crops, clean exit).

## Render-verify note

The shader change can't follow the normal before/after screenshot flow: the **before** state on Windows is a hard crash (no frame to capture), and on Linux/macOS the change is a pure GLSL-spec-correctness no-op (identical std140 layout, so byte-identical output). Evidence is the build+run matrix above. Reviewers on Linux/macOS: please confirm no rendering delta (expected: none).

## Follow-ups (filed, NOT fixed here)

The 3 remaining run-pass failures are **pre-existing latent UB unrelated to these fixes**, root-caused via gdb and filed separately:
- #2031 — `GpuStageTimingObserver` dtor calls `glDeleteQueries` at process-exit static destruction → teardown segfault on `IRLuaPipelineDemo` / `IRCollisionOverlapDemo` (render fully validated; crash is on clean exit).
- #2032 — skeletal `seedVoxelBoneSlots` out-of-bounds `std::span` deref → first-frame segfault on `IRSkeletalDemo`.

## Test plan
- [ ] Linux/OpenGL: build + `IRShapeDebug --auto-screenshot` — confirm no rendering delta (shader change is a no-op).
- [ ] macOS/Metal: build (Metal shaders unaffected; CMake change is `WIN32`-gated).
- [ ] Native Windows: `ir-build` cohort + `ir-run --auto-screenshot` — 41/41 build, 35/38 run.

### DIFF
--- a/engine/render/src/shaders/v_trixel_to_framebuffer.glsl
+++ b/engine/render/src/shaders/v_trixel_to_framebuffer.glsl
@@ -24,6 +24,28 @@ layout (std140, binding = 3) uniform FrameDataIsoTriangles {
     vec2 effectiveSubdivisionsForHover;
     float showHoverHighlight;
     int distanceOffset;
+    // This block is ALSO declared in f_trixel_to_framebuffer.glsl, and a
+    // uniform block shared by name across stages MUST be member-identical or
+    // the GL spec makes the program link ill-formed. Lenient drivers
+    // (Linux/macOS) accepted the vertex-side head-only form; the stricter
+    // native-Windows driver rejects it with "members of uniform block (named
+    // FrameDataIsoTriangles) are not the same between shader stages" and the
+    // program-link assert crashes every demo at startup. The vertex stage
+    // consumes only mpMatrix + textureOffset, but it must still declare the
+    // full scatter UBO tail so the std140 layout matches BOTH the fragment
+    // stage and the shared C++ struct (FrameDataTrixelToFramebuffer,
+    // ir_render_types.hpp). Keep these two blocks in lockstep.
+    ivec2 perAxisBase;
+    float visualYaw;
+    int scatterDebugMode;
+    ivec4 visibleFaceIds;
+    vec4 _detachedResidualPad;
+    vec4 _detachedDepthAxisPad;
+    vec4 scatterFbResolution;
+    int depthColorMode;
+    float depthColorExtent;
+    float _depthColorPad0;
+    int depthPriorityMode;
 };
 
 void main() {
--- a/engine/script/third_party/luajit/CMakeLists.txt
+++ b/engine/script/third_party/luajit/CMakeLists.txt
@@ -70,10 +70,28 @@ endif()
 # before the populate step runs.
 file(MAKE_DIRECTORY ${LUAJIT_INCLUDE_DIR})
 
+# LuaJIT builds through its own Makefile, so this must invoke a *make*
+# program — and on native Windows that program is `mingw32-make`, not
+# `make`. The windows-debug preset uses MinGW Makefiles + MSYS2's mingw64
+# toolchain, which ships `mingw32-make.exe`; the Unix `make` lives in
+# MSYS2's /usr/bin, which is NOT on the build PATH (ir-build prepends only
+# mingw64/bin). A bare `make` there dies with "no such file or directory"
+# and breaks the whole native-Windows link — latent until the fleet first
+# built on real Windows, because WSL/macOS both have `make`.
+#
+# Deliberately NOT ${CMAKE_MAKE_PROGRAM}: under a Ninja generator that is
+# `ninja`, and `ninja BUILDMODE=static …` is nonsense — LuaJIT always
+# needs a make. Select by platform instead.
+if(WIN32)
+    set(LUAJIT_MAKE_PROGRAM mingw32-make)
+else()
+    set(LUAJIT_MAKE_PROGRAM make)
+endif()
+
 add_custom_command(
     OUTPUT ${LUAJIT_LIB_PATH}
     WORKING_DIRECTORY ${LUAJIT_SRC_DIR}
-    COMMAND ${CMAKE_COMMAND} -E env make ${LUAJIT_MAKE_ARGS}
+    COMMAND ${CMAKE_COMMAND} -E env ${LUAJIT_MAKE_PROGRAM} ${LUAJIT_MAKE_ARGS}
     COMMENT "Building LuaJIT 2.1 static library (${LUAJIT_LIB_NAME})"
     VERBATIM
 )


jakildev (commented):
## Review — build+render: native-Windows demo build & smoke bring-up (LuaJIT mingw32-make + trixel UBO stage-match)

**Verdict:** approve (pending Opus recheck — touches engine/render/)

### Analysis

**Shader fix (`v_trixel_to_framebuffer.glsl`)**

The GL spec requires same-named uniform blocks shared by name to be member-identical across stages. Verified:
- Vertex UBO (after this PR) and fragment UBO (`f_trixel_to_framebuffer.glsl`) are now member-identical in type, name, and order.
- Both match the C++ struct `FrameDataTrixelToFramebuffer` (`ir_render_types.hpp`), whose static_asserts (`sizeof == 208`, offset of `visibleFaceIds_ == 128`, etc.) gate layout correctness at compile time.
- std140 offsets computed manually: mat4 0-63, vec2 chain 64-111, ivec2 perAxisBase 112-119, float visualYaw 120-123, int scatterDebugMode 124-127, ivec4 visibleFaceIds 128-143 (16-byte aligned ✓), 3×vec4 tail 144-191, int+float+float+int 192-207 → 208 total ✓.
- Metal shader (`trixel_to_framebuffer.metal`) already declares the full struct; no Metal-side change needed. ✅
- Only `mpMatrix` and `textureOffset` are consumed in `void main()` — the comment is accurate.

**CMake fix (`engine/script/third_party/luajit/CMakeLists.txt`)**

The `WIN32` guard is correct for the MSYS2/mingw64 target: `make` is in `/usr/bin` (not on the `ir-build` PATH); `mingw32-make` is the right tool. The reason for NOT using `${CMAKE_MAKE_PROGRAM}` (= `ninja` under a Ninja generator) is well-documented in the comment. Change is `WIN32`-gated and a no-op on Linux/macOS.

### Praise
- Both root causes were correctly diagnosed (GLSL UBO cross-stage identity rule; MSYS2 `make` PATH gap) and fixed at the right layer.
- 41/41 demo build + 35/38 run clean on native Windows. The 3 remaining failures are pre-existing latent UB filed as separate issues (#2031, #2032), not regressions.
- The comment in the vertex shader is exactly the right amount of explanation — future editors touching the UBO will know to keep both stages in sync.

- [x] Build 41/41 demo targets on native Windows — confirmed
- [x] 35/38 run + screenshot clean on Windows (#2031 teardown + #2032 skeletal OOB are pre-existing) — confirmed
- [ ] Linux/OpenGL: build + `IRShapeDebug --auto-screenshot` — confirm no rendering delta (shader change is byte-identical std140; expect zero drift)
- [ ] macOS/Metal: build — confirm CMake `WIN32` guard is correctly inactive on macOS (Metal shader unaffected)

Escalation: please Opus-recheck before merge (touches: engine/render/ shader path).

Opus recheck required: touches engine/render/ — shader cross-stage UBO fix; straightforward but warrants Opus sign-off given the render pipeline surface.

jakildev (commented):
## Opus recheck — build+render: native-Windows demo build & smoke bring-up (UBO stage-match + LuaJIT make)

**Verdict:** approve

Sonnet flagged this for Opus recheck because it touches `engine/render/` (the shader cross-stage UBO fix). I focused the pass on the two things Sonnet's manual offset table couldn't fully close: (1) whether the fix is *complete* across every program that shares the block name, and (2) whether the render fast path stays byte-identical.

### Cross-stage UBO member-identity — confirmed correct AND complete

Sonnet flagged that the vertex block now matches the fragment block and the C++ struct member-for-member. On closer read I confirm: the post-PR `FrameDataIsoTriangles` in `v_trixel_to_framebuffer.glsl` is a 19-member block type-identical and order-identical to `f_trixel_to_framebuffer.glsl` and to `FrameDataTrixelToFramebuffer` (`ir_render_types.hpp`), whose `static_assert`s gate the std140 offsets (visibleFaceIds_==128, detachedResidual_==144 / detachedDepthAxis_==160 / scatterFbResolution_==176). Names differ between CPU (`_` suffix, a couple of renames like `canvasZoomLevel_`↔`zoomLevel`) but cross-stage and CPU↔GPU matching are by layout/offset, not name, so that's fine.

**Completeness — the part Sonnet's per-program check did not spell out.** GLSL cross-stage member-identity is required only when the *same block name* is declared in *more than one stage of one program*, and program linking is a **runtime** event (`glLinkProgram`), not build-time — so "41/41 demos built on Windows" does not by itself prove every program links. I audited every shader declaring `FrameDataIsoTriangles`:

- `trixel_to_framebuffer` (v + f) — the **only** program declaring the block in two stages, and the one the strict Windows driver rejected (8 vs 19). This PR makes it consistent.
- `v_peraxis_scatter.glsl` declares the block (14 members) but its paired fragment `f_peraxis_scatter.glsl` declares **no** UBO block — no cross-stage partner, so no link-validation conflict. This is the case the cardinal-only smoke would *not* have exercised at runtime (the scatter path is off-cardinal), and it's clean.

So the fix covers the sole affected program; no latent same-class crash remains.

### Render behavior is byte-identical on Linux/macOS

The vertex stage consumes only `mpMatrix` + `textureOffset` (both in the unchanged block head); the appended tail members are declared but unread, and the head offsets don't move. That's a language-level guarantee of zero render delta on the lenient drivers — no build needed to confirm the invariant. Metal is unaffected (no GLSL cross-stage matching; the `.metal` struct already carries the full layout).

### LuaJIT CMake change

`make` -> platform-selected `mingw32-make` on `WIN32`, `make` elsewhere. Non-Windows behavior is unchanged. The comment's rationale for *not* using `${CMAKE_MAKE_PROGRAM}` (it'd be `ninja` under a Ninja generator, nonsense for LuaJIT's own Makefile) is correct.

The two pre-existing Windows runtime failures (#2031 teardown, #2032 skeletal OOB) are correctly scoped as separate latent-UB issues, not regressions from this PR.

Clean approve. Comments in both files are exactly the right "keep these in lockstep" guidance for future editors.

jakildev:
Cross-host smoke OK on macos (fresh checkout build + IRShapeDebug --auto-screenshot 10). Build clean, 21 screenshots captured, exit 0. — worker

PR #2041, opened 2026-06-26
fleet: usage gate honors future resetsAt over stale observed_at

jakildev:
## Problem

The dispatcher kept re-triggering a worker (observed as "keeps triggering an opus worker") while pinned at the 5-hour rate limit.

Root cause is in `fleet-dispatcher`'s `usage_gate_status`: it discarded any usage observation older than `USAGE_STALE_SECONDS` (default 1h) **before** consulting its `resetsAt`. But a five-hour-window observation stays bound until `resetsAt` — utilization cannot drop below the wall until the window resets. So for the ~4h between `observed_at + 1h` and `resetsAt`, the gate went blind and reported `open` even at **100% utilization**.

Confirmed live on this host: `five_hour.json` had `utilization: 1`, `observed_at` 82 min ago, `resetsAt` still 53s in the future — and `--gate-status` printed `open`.

### Why it loops
1. Observation crosses 1h of age → gate flips `open`.
2. Dispatcher fires a worker into the still-closed window.
3. Worker hits the 5h limit → exits code 2 → `fleet-claude-stream` re-latches the observation with a fresh `observed_at`.
4. Gate closes again for another hour → back to step 1.

Each cycle burns one worker dispatch. Unrelated to other in-flight PRs.

## Fix

Make `resetsAt` authoritative: skip an observation only when `resetsAt + grace` has passed. A future `resetsAt` keeps it live regardless of `observed_at` age. The `observed_at` staleness cutoff now applies **only** when `resetsAt` is absent/unparseable, so an orphaned latch still ages out.

## Tests

- `T4` retargeted to the staleness-fallback path (no `resetsAt` → ages out).
- New `T4b` guards the regression: stale `observed_at` + future `resetsAt` → **still closed**.
- Suite now isolates from the operator's `~/.fleet/fleet-up.conf` (`FLEET_CONF=/dev/null` + clears inherited `FLEET_DISPATCHER_USAGE_GATE*`), so a host with a per-type override (e.g. `FLEET_DISPATCHER_USAGE_GATE_FIVE_HOUR`) no longer false-fails `T3`.

All 18 cases pass.

### DIFF
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -251,10 +251,15 @@ LAST_DISPATCH_EPOCH=0
 #      five_hour=0.80; everything else falls to the global default)
 #   4. Final fallback 0.80
 #
-# Conservative against staleness: an observation older than
-# USAGE_STALE_SECONDS (default 1h) or whose resetsAt + RESET_GRACE_SECONDS
-# is in the past is ignored, so the gate doesn't latch closed forever
-# after a single high-watermark reading.
+# Conservative against staleness, but resetsAt is authoritative: an
+# observation whose resetsAt + RESET_GRACE_SECONDS is in the past is
+# ignored (window closed). A future resetsAt keeps the observation live
+# no matter how old observed_at is — utilization can't fall below the
+# wall until the window resets, so an old-but-still-bound reading must
+# keep the gate closed. The USAGE_STALE_SECONDS cutoff (default 1h)
+# only applies when resetsAt is absent/unparseable, so an orphaned
+# latch still ages out. (Earlier the stale cutoff ran first and blinded
+# the gate ~4h into a 5h window, re-triggering a worker into the wall.)
 #
 # RESET_GRACE_SECONDS pads the resetsAt boundary so the gate doesn't flip
 # open on the exact second the latched window expires. Anthropic's rolling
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: usage gate honors future resetsAt over stale observed_at

**Verdict:** approve

### Praise

- **Root-cause analysis is precise and verifiable.** The PR body walks the failure loop step-by-step (gate opens → worker fires → rate-limit hit → `observed_at` refreshed → gate closes → repeat), making the fix easy to validate against the code.
- **`resetsAt`-authoritative design is the right invariant.** Utilization can't fall below the wall until the window resets, so discarding a live observation on `observed_at` staleness was always wrong. The new two-phase check (`resets_known → window_open → keep; else → observed_at staleness fallback`) is the minimal correct fix.
- **T4 split into T4 + T4b is the right test structure.** The old T4 had `resetsAt` but was testing the wrong expected outcome (`open` — the bug behavior). Removing `resetsAt` from T4 makes it a pure staleness-fallback test, and T4b becomes the explicit regression guard for the bug path (stale `observed_at` + future `resetsAt` → still closed). Two orthogonal code paths get two orthogonal tests.
- **Suite isolation fix is a first-class correctness fix in its own right.** Without `FLEET_CONF=/dev/null` + the `compgen` unset loop, T3 (global threshold override) false-fails on any host that has `FLEET_DISPATCHER_USAGE_GATE_FIVE_HOUR` set in `fleet-up.conf`. This was a silent portability bug hiding in the suite. Filing it as part of the same PR rather than a follow-up keeps the test suite trustworthy.

- [x] All 18 cases pass (confirmed in PR body).
- [ ] Run `scripts/fleet/tests/test_dispatcher_usage_gate.sh` on the other host (Linux/Windows) to confirm the `compgen -A variable` unset loop and `date +%s` portability hold on non-macOS shells.

Opus recheck not required.

PR #2045, opened 2026-06-27
IRArgs — standardized CLI argument framework + default --help/-h (#2044)

### LINKED ISSUE #2044: engine: IRArgs — standardized CLI argument framework + default --help/-h for all launch targets
**Area:** engine
**Model:** opus
**Blocked by:** (none)

## Problem
Every launch target hand-rolls its own `for (i…) std::strcmp(argv[i], "--foo")` parsing — `gpu_particles`, `ir_voxel_yaw`, `lua_perf_grid`, `shape_debug`, `perf_grid`, `canvas_stress` all do — plus two scattered engine helpers (`IRVideo::parseAutoScreenshotArgv`, `IREngine::parseConfigPresetArg`). Consequences: **no `--help` anywhere** (you read the source to learn a target's flags), inconsistent flag names/semantics across demos, duplicated parse+validate code, and every new demo reinvents it.

## Goal
1. A **default `--help` / `-h`** on **all** launch targets that prints every registered argument (name, short alias, type, default, one-line help) — the engine-common ones plus the target's own.
2. An **engine-level argument framework (`IRArgs`)** targets register against declaratively, so no demo writes a parse loop again. Common args (`--auto-screenshot`, `--config-preset`, …) register once in the engine and every target inherits them; `--help`/`-h` is built in and free.

Design + decomposition in the `## Plan` comment.

## Acceptance
- Any migrated target run with `--help` or `-h` prints the full arg list and exits 0.
- A target declares args declaratively (name/alias/type/default/help); the engine parses once and exposes typed values; unknown args + `--help` handled centrally.
- `parseAutoScreenshotArgv` + `parseConfigPresetArg` fold into the registry as common args (existing flag spellings preserved).

jakildev:
## Summary
- New `IRArgs::Parser` — declarative CLI argument framework (`engine/include/irreden/ir_args.hpp` + `engine/ir_args.cpp`) with a built-in `--help` / `-h` for every launch target.
- Args are registered declaratively (`flag` / `integer` / `number` / `string` / `optionalInt`); `--help` prints auto-generated usage and exits 0; an unknown arg or a missing value prints a diagnostic + usage and exits 2; `--help` resolves **before any heavy init**, so `Target --help` is instant and headless-safe.
- `--auto-screenshot` and `--config-preset` fold in as engine-common args every target inherits (existing flag spellings preserved; the `--auto-screenshot` optional-int peek mirrors the legacy helper so command lines stay byte-identical).
- `fog_demo` is the reference adoption — it drops its hand-rolled `parseAutoScreenshotArgv` call **and** its `--moving-observer` strcmp loop, registering `--moving-observer` as an `IRArgs` flag instead, and gains `--help` for free. So it demonstrates both a custom flag and the inherited common args. Capture / reveal behavior is unchanged.
- `engine/CLAUDE.md` documents the "CLI args go through `IRArgs`, never a hand-rolled `strcmp` loop" convention.

**Scope (foundation):** the legacy scattered helpers (`IRVideo::parseAutoScreenshotArgv`, `IREngine::parseConfigPresetArg`) and the remaining demos' hand-rolled parse loops are left intact — migrating them onto `IRArgs` and retiring the helpers are #2044 follow-ons (filed after this lands). Backend-agnostic stdlib code (no GL/Metal divergence), so no cross-host smoke needed.

## Test plan
- [x] `IRFogDemo --help` and `-h` print the full arg list (incl. `--moving-observer`) and exit 0, before any window/GL/Metal init.
- [x] Unknown arg (`--bogus`) → "Unknown argument" + usage on stderr, exit 2; `--config-preset` with no value → "expects a value", exit 2.
- [x] `IRFogDemo --auto-screenshot 12` (static grid reveal) and `--auto-screenshot 12 --moving-observer` (analytic vision circle) both cycle all 3 shots and capture (behavior-preserving migration).
- [x] Builds clean on macOS/Metal (`fleet-build --target IRFogDemo`); clang-format clean.

Closes #2044

### DIFF
--- a/creations/demos/fog_demo/main.cpp
+++ b/creations/demos/fog_demo/main.cpp
@@ -138,10 +138,6 @@ void initCommands();
 void initEntities();
 
 int main(int argc, char **argv) {
-    // Declarative arg parsing via the engine framework: fog_demo registers its
-    // own --moving-observer flag and inherits the engine-common args
-    // (--auto-screenshot, --config-preset) plus a free --help / -h. Reference
-    // adoption for #2044 — `IRFogDemo --help` lists every arg and exits.
     IRArgs::Parser args(
         "fog_demo — fog-of-war render-pass demo (FOG_TO_TRIXEL) + cross-host smoke coverage."
     );
--- a/creations/demos/fog_demo/main.cpp
+++ b/creations/demos/fog_demo/main.cpp
@@ -33,6 +33,7 @@
 // pre-removal shape_debug setup; the rest of the skeleton follows the smaller
 // day_cycle demo.
 
+#include <irreden/ir_args.hpp>
 #include <irreden/ir_engine.hpp>
 #include <irreden/ir_entity.hpp>
 #include <irreden/ir_math.hpp>
@@ -80,7 +81,6 @@
 #include <irreden/voxel/systems/system_update_voxel_set_children.hpp>
 
 #include <list>
-#include <string_view>
 
 using namespace IRComponents;
 using IRMath::Color;
@@ -138,12 +138,21 @@ void initCommands();
 void initEntities();
 
 int main(int argc, char **argv) {
-    IRVideo::parseAutoScreenshotArgv(argc, argv, &g_autoWarmupFrames);
-    for (int i = 1; i < argc; ++i) {
-        if (std::string_view(argv[i]) == "--moving-observer") {
-            g_movingObserver = true;
-        }
-    }
+    // Declarative arg parsing via the engine framework: fog_demo registers its
+    // own --moving-observer flag and inherits the engine-common args
+    // (--auto-screenshot, --config-preset) plus a free --help / -h. Reference
+    // adoption for #2044 — `IRFogDemo --help` lists every arg and exits.
+    IRArgs::Parser args(
+        "fog_demo — fog-of-war render-pass demo (FOG_TO_TRIXEL) + cross-host smoke coverage."
+    );
+    args.flag(
+        "--moving-observer",
+        "Per-frame analytic vision circle orbiting the origin (smooth reveal) "
+        "instead of the static grid reveal"
+    );
+    args.parse(argc, argv);
+    g_autoWarmupFrames = args.autoScreenshotWarmupFrames();
+    g_movingObserver = args.getFlag("--moving-observer");
 
     IR_LOG_INFO("Starting creation: fog_demo");
     IREngine::init(argv[0]);
--- a/engine/CLAUDE.md
+++ b/engine/CLAUDE.md
@@ -17,6 +17,24 @@ Core engine static libraries. Everything here is shared by every creation.
   modules (`common/`, `math/`, `profile/`) do not depend on high-level
   ones (`render/`, `world/`).
 
+## CLI args go through `IRArgs`, never a hand-rolled `strcmp` loop
+
+`engine/include/irreden/ir_args.hpp` is the declarative argument framework.
+A launch target constructs an `IRArgs::Parser` (which pre-registers the
+engine-common args — `--auto-screenshot`, `--config-preset` — plus a free
+`--help` / `-h`), declares its own args (`flag` / `integer` / `number` /
+`string` / `optionalInt`), then calls `parse(argc, argv)` at the **top of
+`main`, before any window / GL / Metal init** so `--help` is instant and
+headless-safe. `--help` prints the auto-generated usage and `exit(0)`; an
+unknown arg prints an error + usage and `exit(2)`. Read the common flags via
+`autoScreenshotWarmupFrames()` / `configPreset()`. `creations/demos/fog_demo/
+main.cpp` is the reference adoption.
+
+Do **not** add a new `for (i…) std::strcmp(argv[i], "--foo")` parse loop to a
+target. The legacy scattered helpers (`IRVideo::parseAutoScreenshotArgv`,
+`IREngine::parseConfigPresetArg`) and the remaining demos' hand-rolled loops
+are being migrated onto `IRArgs` and retired (#2044 follow-ons).
+
 ## `SystemName` enum is authoritative
 
 Every prefab system that uses the `System<NAME>::create()` template pattern
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -9,6 +9,7 @@ include_directories(
 
 add_library(IrredenEngine STATIC
     engine.cpp
+    ir_args.cpp
 )
 
 # set(CMAKE_CXX_SCAN_FOR_MODULES ON)
--- a/engine/include/irreden/ir_args.hpp
+++ b/engine/include/irreden/ir_args.hpp
@@ -1,0 +1,132 @@
+#ifndef IR_ARGS_H
+#define IR_ARGS_H
+
+#include <string>
+#include <vector>
+
+// IRArgs — declarative command-line argument framework with a built-in
+// --help / -h for every launch target.
+//
+// Before this, every demo hand-rolled its own `for (i…) std::strcmp(...)`
+// parse loop, no target had a --help, and the two engine helpers
+// (IRVideo::parseAutoScreenshotArgv, IREngine::parseConfigPresetArg) lived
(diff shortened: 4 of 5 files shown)


jakildev:
Rebased onto current master after #2033 (`--moving-observer` analytic fog) merged into `fog_demo/main.cpp` mid-iteration. Resolved the `main()` conflict by routing the new `--moving-observer` flag through `IRArgs` too (registered as a flag, read via `getFlag`) and dropping the now-dead `std::string_view` strcmp loop + include — the migration is complete, not partial. Rebuilt + re-verified on macOS/Metal: `--help` lists `--moving-observer`, and both the static and analytic reveal paths capture cleanly with no unknown-arg rejection. — worker

jakildev (commented):
## Review — IRArgs — standardized CLI argument framework + default --help/-h (#2044)

**Verdict:** approve

Standalone PR (base: `master`). CONFLICTING state checked — no out-of-scope files or oversized churn (conflict is likely from a concurrent `engine/CLAUDE.md` edit on master; needs a rebase before merge).

### ECS / Engine invariants

Not applicable — this is a stdlib-only CLI parsing module with no ECS, render, or world-manager involvement.

### Ownership / lifetime

Clean. `Parser` owns its `m_entries` by value. No heap-owning raw pointers, no shared_ptr, no cross-tick component references.

### Naming / style

`Parser`'s private data members correctly use `m_` prefix (`m_description`, `m_programName`, `m_entries`). `Entry` struct members use trailing `_` — consistent with the public-member convention since `struct` fields default to public access. ✓

### Tests / build

PR body reports build clean on macOS/Metal and `clang-format` clean. Manual test plan covers the key paths (help, unknown arg, missing value, `--auto-screenshot` round-trip). ✓

### Praise

- The two-phase `--help` scan (pre-scan before mutating any entry) cleanly prevents partial mutation before exit and ensures `usage()` always prints registered defaults.
- Engine-common args pre-registered in the constructor — existing targets that migrate gain `--help` + `--config-preset` for free.
- `autoScreenshotWarmupFrames()` returning 0 when `--auto-screenshot` is absent preserves the legacy "0 = not requested" sentinel precisely; migration is byte-identical.
- `usage()` sorts alphabetically regardless of registration order — stable, scannable output.

### Nits

- `engine/include/irreden/ir_args.hpp:1` — `#ifndef IR_ARGS_H / #define IR_ARGS_H` guard style; check whether the engine convention for new headers is `#pragma once` (most modern headers in the engine use that form).
- `engine/include/irreden/ir_args.hpp:9-25` — Multi-line comment block explaining what the module does. The example usage is the valuable part; the prose summary duplicates what identifiers + the PR body already convey. Per the engine baseline, comments should cover non-obvious WHY, not WHAT. Consider trimming to the example only.
- `engine/include/irreden/ir_args.hpp:47` — `enum class Type { FLAG, INT, FLOAT, STRING, OPTIONAL_INT }` uses ALL_CAPS values. Confirm this matches the engine's `enum class` convention (some ECS enums use `kPascalCase`). If not, align with the project style.
- `engine/ir_args.cpp:74` — `add()`'s duplicate guard calls `find(name)` which also matches on `shortAlias_`. A second registration with the same short alias (e.g., two args both using `-x`) would not fire the assert. Low risk for internal use, but the guard is silent about alias collisions.
- `engine/ir_args.cpp:186,196` — `std::atoi` / `std::atof` return 0 on malformed input (`--grid-size foo` silently becomes 0). Fine for demo-tool usage, but `strtol`/`strtof` with an error check would catch typos. Optional for now given the scope.
- `creations/demos/fog_demo/main.cpp:107-110` — Inline comment block explains what the code does ("inherits exactly the engine-common args…"). The "Reference adoption for #2044" note belongs in the commit/PR description, not in the source. The code is self-evident from the `IRArgs::Parser` constructor call.

- [ ] `git rebase origin/master` — resolve the CONFLICTING state so the PR is mergeable.
- [ ] Confirm `IRFogDemo --help` prints the full arg table and exits 0 (before any window/GL/Metal init).
- [ ] Confirm `IRFogDemo --bogus` → "Unknown argument" on stderr, exit 2.
- [ ] Confirm `IRFogDemo --auto-screenshot 12` captures all shots (behavior-preserving round-trip).
- [ ] Linux/GL build — this is stdlib-only so no GL/Metal divergence expected, but a linux-debug build verifies.

Opus recheck not required.

jakildev (commented):
## Opus recheck — IRArgs CLI framework (#2044)

**Verdict:** approve

Picked this up under the opus-only trigger for **new public `ir_*.hpp` surface** (`engine/include/irreden/ir_args.hpp`), since `engine/CLAUDE.md` now *mandates* every launch target route CLI parsing through it — the API contract is about to become engine-wide, so it's worth a lifetime/ownership pass three levels deep. Sonnet's first pass was thorough on naming, ownership-at-a-glance, and the migration semantics; I focused on what a first pass can't easily confirm.

### Reference lifetime through the `Entry&`/`push_back` path — confirmed safe

The one genuinely subtle thing here is `add()` returning `Entry&` into `m_entries` (a `std::vector`), then each chained registrar (`flag`/`integer`/…) writing one field through that reference. A returned reference into a vector that later `push_back`s is the classic dangling-reference trap. Walked it: each registrar **consumes the reference fully within its own call** (set one field, `return *this`) — no reference survives across the next registration's `push_back`/reallocation. `find()` returns an `Entry*`, but it's only ever called inside `parse()`, after all registration is done and while the vector size is frozen, so no reallocation can invalidate it mid-loop. **No lifetime hazard.** ✓

### Behavior-preservation of the migrated paths — confirmed

- `OPTIONAL_INT` peek (`atoi(next) > 0` → consume, else leave the bare default in `intValue_`) is byte-identical to the legacy `parseAutoScreenshotArgv` peek. `autoScreenshotWarmupFrames()` returning `0` when the switch is absent preserves the "0 = not requested" sentinel exactly. ✓
- `getInt()` accepting both `INT` and `OPTIONAL_INT` is correct — `--auto-screenshot`'s value is read back through it.
- The fog_demo conflict resolution (routing `--moving-observer` through `flag()`/`getFlag` and dropping the dead `string_view` strcmp loop + include) is a complete migration, not a half-step. ✓

### Concurrency / hot-path

N/A by construction — `parse()` runs once at the top of `main()` before any init, off every tick and render path. The O(n·m) linear `find()` scans are irrelevant at a handful of args.

### On Sonnet's nits

- **`#ifndef IR_ARGS_H` vs `#pragma once`** — Sonnet's nit is defensible but I'll add the data: the engine is split, not uniform. The immediate sibling this header sits next to, `ir_engine.hpp`, uses the same `#ifndef IR_ENGINE_H` style — so the new file is *locally* consistent — but the engine-wide majority is `#pragma once` (~107 files to 1). Either is acceptable; if anything the umbrella `ir_*.hpp` headers are the ifndef holdouts. Not worth a churn either way. Leaving to author preference.
- Sonnet's other nits (alias-collision not caught by the dup guard; `atoi`/`atof` silently 0 on malformed; comment-prose trimming) I agree with as nits — none rise to needs-fix for demo-tooling CLI parsing.

### Verdict

Clean, well-scoped, correctly behavior-preserving. The public API shape (declarative chained registration + pre-registered common args + free `--help`) is sound for the engine-wide adoption the CLAUDE.md mandates. Approving. The pre-merge checklist Sonnet posted (rebase to clear CONFLICTING, Linux/GL build sanity) still stands; `fleet:needs-windows-smoke` remains the gating cross-host check.

jakildev:
Addressed has-nits:
- Header comment trimmed to the usage example only (removed prose that duplicated the PR body and 'The header is intentionally light…' note)
- Added alias collision assert in `add()`: a duplicate short alias now fires the same debug assert as a duplicate long name
- Removed WHAT comment block in `fog_demo/main.cpp` (lines 141-144; 'Reference adoption for #2044' belongs in the commit message, not source)
- Include guard (`#ifndef`) and enum value casing left unchanged — both match the engine convention (`ir_engine.hpp` uses the same guard; `SCREAMING_SNAKE_CASE` is the documented enum-value style)
- atoi/atof error handling skipped (reviewer marked optional)

Build clean (IRFogDemo). — worker

Issue #2046, opened 2026-06-27
render: investigation spike — world-placed detached entity-canvas mis-scales / floor-clips at minimum camera zoom


jakildev:
Investigation spike (carved from #2043).

**Area:** engine/render — detached entity-canvas composite / depth
**Model:** opus
**Blocked by:** #2043 (shares the detached-canvas composite + raster surface; serialize to avoid same-file conflict)

## Problem
Carved from #2043 (which now scopes only the see-through GAPS, root-caused to the zoom-independent detached subdivision). The SECOND symptom — a world-placed detached voxel object that "does not shrink with camera zoom the way main-canvas geometry does", appears too large at minimum zoom, and CLIPS THROUGH the floor — is split here because its root cause is unverified and its headless repro is currently blocked.

## Why this is a spike, not a fix task
- **Headless repro is blocked.** #2043 notes the demo's camera zoom is fixed and the screenshot-shot zoom field is metadata, so a low-zoom capture can't be driven today. The first deliverable is to UNBLOCK this.
- **Static analysis contradicts the obvious "zoom-power mismatch" theory.** Tracing the composite (`system_entity_canvas_to_framebuffer.hpp:160-200` applies cameraZoom¹ into the framebuffer) and FRAMEBUFFER_TO_SCREEN (`system_framebuffer_to_screen.hpp` blits by resolution scaleFactor, NOT re-applying zoom) shows BOTH the main GRID content (`system_trixel_to_framebuffer.hpp` model = res×zoom) and the detached composite scale as zoom¹ on screen. So "doesn't shrink with zoom" is NOT explained by a zoom-power difference — the perceived oversize is more likely (a) the gaps making size hard to judge, (b) a minimum on-screen size floor, or (c) the depth floor-clip making it look like it pokes through. Needs instrumentation to disambiguate.
- **"Clips through the floor" is a DEPTH issue** overlapping the #1624/#1958/#1884 effSub depth-scaling machinery (worldDepth × effSub × 4; the documented "world-placed solids sink behind the floor as effSub changes" family). At MIN zoom effSub floors at 1 — the depth-offset rescale (`effectiveSubdivisionsForHover_.y`) may degenerate there.

## Spike deliverables
[...]

PR #2051, opened 2026-06-27
render: detached canvas raster resolution (cubeSub) leaks into apparent size + gather gaps (#2043, subsumes #2046)

### LINKED ISSUE #2043: render: detached entity-canvas renders small/low-zoom voxel objects with gaps + uncontrollable size
**Area:** render — detached entity-canvas composite / small-scale voxel rasterization
**Suggested model:** [opus] (render + shader)

## Problem
A **small voxel object on a world-placed detached entity-canvas** renders poorly at low camera zoom. Surfaced making the arcade_starter player (a 2×2×2 voxel cube) a detached canvas for smooth sub-tile movement (game `jakildev/irreden` #211 / PR #212), camera fixed iso zoom 0.5.

Three symptoms:
1. **Gaps / hollow rendering.** The 2³ cube on the detached canvas renders with visible holes (you see through it to the floor) at zoom 0.5. The *same* 2³ cube rendered on the **main GRID canvas** (pre-detached) renders **solid**. So the detached path introduces the gaps. **Verified by screenshot** (default zoom).
2. **Uncontrollable size, and (interactive) min-zoom oversize.** At minimum zoom the detached object appears **too large relative to the world** — it does not shrink with zoom the way main-canvas geometry does — and **clips through the floor** (user-observed interactively). (Headless repro is blocked: the demo's camera zoom is fixed and the screenshot-shot zoom field is metadata; couldn't drive a low-zoom capture.)
3. **Not tunable by the creation.** `IRPrefab::EntityCanvas::createWithVoxelPool(name, canvasSize, voxelPoolSize)` — the on-screen size AND the gaps are **independent of both `canvasSize` and `voxelPoolSize`**: render output is **byte-identical** across `(48², 6³)`, `(96², 6³)`, and `(48², 4³)`. So a creation cannot tune these away; they're engine-side.

## Suspected area (from investigation)
- `engine/prefabs/irreden/render/systems/system_entity_canvas_to_framebuffer.hpp` — the composite scale/placement (the `entityScale`/`cameraZoom` math, ~lines 142–200) and its sampling (nearest, the iso `(1, 0.5)` vertical squish, ~lines 156–159) are the likely source of both the gaps (sub-texel drop at the squish) and the zoom-relative size.
- The detached pool's voxel rasterization at small world scale may also under-cover.

## Impact
Detached entity-canvases are currently unsuitable for **small in-world objects at low zoom** — they look broken (gappy) and mis-scale at min zoom. This blocks the intended use (a smoothly-moving small arcade player). Larger detached models (automata pedals, gear_model) don't surface it because they're big and viewed at a scale where the gaps/zoom-scale aren't visible.

## Acceptance
- A small (≈2³) world-placed detached voxel object renders **solid** (no see-through gaps) at zoom 0.5, matching the main-canvas GRID rendering of the same object.
- The detached object **scales with camera zoom consistently** with main-canvas geometry across the zoom range (no min-zoom oversize / floor clip).
- Verified on GL and Metal.

## Context
- Lighting on the detached player was a **separate, already-fixed game-side issue** (the demo just needed to register `PROPAGATE_CANVAS_ROTATION`; pushed to PR #212) — not part of this ticket.
- Engine detached-canvas world-lighting/depth work: closed #1558 / #1576 / #1624.

jakildev:
Claiming issue #2043. Investigation complete; escalating for a design call (see NEEDS-DESIGN comment).

Closes #2043


jakildev:
## NEEDS-DESIGN

Investigated #2043 (small voxel object on a world-placed DETACHED entity-canvas: see-through GAPS + uncontrollable / non-zoom-tracking SIZE at low camera zoom). The root cause is a **structural mismatch in how the two render paths reconcile camera-zoom against raster density** — fixing it cleanly is a design call about the detached-canvas scaling contract, which is coupled to the #1624/#1576 world-placed depth machinery. Handing off rather than guessing.

### What the code does (mechanism, with citations)

**GRID / main world canvas** reaches the framebuffer through `TRIXEL_TO_FRAMEBUFFER` (`system_trixel_to_framebuffer.hpp:93-122`):
- raster density is **zoom-tracking**: `getVoxelRenderEffectiveSubdivisions()` in FULL mode = `clamp(baseSub × round(zoom), 1, 16)` (`render_manager.cpp:288-301`) — effSub grows with zoom-in, shrinks at zoom-out.
- the de-tile gather is **density-aware**: `canvasZoomLevel_ = getCameraZoom() / effectiveSubdivisions` (lines 99-101), folded into the mpMatrix model scale. So one rastered world-unit ≈ a fixed framebuffer footprint and the gather samples at ~1:1 regardless of zoom → **solid, no minification holes**.

**DETACHED** reaches the framebuffer through `ENTITY_CANVAS_TO_FRAMEBUFFER` (`system_entity_canvas_to_framebuffer.hpp:107-301`):
- raster density is **canvas-size-capped, NOT zoom-tracking**: `cubeSub = clamp(effSub, 1, subdivisionCap(canvasSize, poolSize))`, and `subdivisionCap` = `floor(canvasPx / isoFootprint)` — purely geometry, no zoom (`detached_revoxelize.hpp:232-245`; cap applied in `system_voxel_to_trixel.hpp:738-745`; stamped to `renderedSubdivisions_` at :754).
- the composite bakes **raw camera zoom** into the quad's model scale via `entityScale = canvasSize_ / mainCanvasSize_` → `model = scale(fbRes × cameraZoom × entityScale)` (lines 165, 192-200). The cardinal gather **ignores** `fd.canvasZoomLevel_` (set at :204 but unread on the cardinal path — it's reserved for per-axis scatter), so there is **no `/cubeSub` density divide** the way the main canvas does `/effSub`.

### Consequence — both symptoms fall out of this

1. **Gaps (symptom 1).** px-per-canvas-texel at the gather = `fbRes × zoom / mainCanvasSize`. At zoom 0.5 with `mainCanvasSize ≈ fbRes` this is `< 1` → the canvas is **minified** at gather time. The de-tile gather's parity row-shift (`ir_iso_common.glsl:399-410` `trixelFramebufferSamplePosition`) reconstructs trixels assuming ~1:1 framebuffer↔canvas correspondence; under minification + NEAREST sampling it lands neighbouring fragments on non-adjacent (often empty) texels → see-through holes. The GRID path never minifies because its density divide keeps the gather at ~1:1.

2. **Uncontrollable / non-zoom-tracking size (symptom 2).** On-screen cube extent = `isoFootprint × cubeSub × fbRes × zoom / mainCanvasSize`. `cubeSub` (internal raster resolution) is a **multiplicative factor on on-screen size** — which is wrong; internal resolution should not change apparent size. Because `cubeSub = clamp(effSub, 1, cap)` and effSub steps with zoom, apparent size jumps at zoom thresholds and is governed by globals the creation can't tune (baseSub, mainCanvasSize) → matches the issue's "byte-identical across `canvasSize`/`poolSize`" (canvasSize cancels in the ratio; it only sets the cap) and "doesn't shrink consistently with the world."

### Why this is a design call, not a contained fix

Every fix path changes the **detached-canvas scaling contract**, which the #1624/#1576 depth integration is tightly coupled to (`depthScale = effSub / cubeSub` carried in `effectiveSubdivisionsForHover_.y`; `distanceOffset_` = world-iso-depth × effSub × 4) and which carries hard **byte-identical invariants** for the existing cases the architect has built up across #1624/#1884/#1958/#1960 (screen-locked overlay, world-placed depth-sort, foreground-priority partition, per-trixel tiers). The candidate approaches each have a different blast radius:

- **(A) Decouple zoom from the detached gather** — composite at native 1:1 (px-per-texel = `fbRes/mainCanvasSize`, like GRID), let `FRAMEBUFFER_TO_SCREEN` apply zoom. Cleanest conceptually, but the placement math (`entityFbCenter`, the #1883 texel-snap jitter fix, the #1944 effective-camera-iso pivot) and the depth offset are all expressed in the current zoom-baked space; re-deriving them risks regressing rotation jitter / depth-sort.
- **(B) Make `cubeSub` zoom-tracking** (re-raster the detached canvas at a density that follows zoom like effSub) so the gather stays ≥1:1 and apparent size decouples from internal resolution. Touches `depthScale` (the `effSub/cubeSub` ratio) and the canvas re-allocation cadence.
- **(C) Minification-safe gather** (supersample / coverage / mip the de-tile gather). Touches the **shared** gather shader used by the main canvas → highest regression surface.
- **(D) Clamp the composite to never minify below 1:1** — kills the holes but conflicts directly with symptom-2's "make small objects actually small at low zoom."

1. **Which scaling model should a world-placed detached canvas use** — should apparent size be a pure function of `world-extent × zoom` (matching GRID exactly, the acceptance's intent), decoupled from `cubeSub`? If so, is **(A)** (decouple zoom from the gather) or **(B)** (zoom-track cubeSub) the intended direction given the #1624 depth coupling?
2. **What is the target visual** for a ~2³ object at zoom 0.5 — pixel-for-pixel identical to the same cube rendered GRID, or "proportional and solid" within tolerance? (Determines whether the de-tile parity reconstruction must be made minification-exact, or just hole-free.)
3. **Is preserving the byte-identical invariant** for the existing screen-locked / world-placed / foreground-priority / per-trixel cases a hard constraint on the fix, or is a coordinated reference refresh acceptable?

### Verification notes for whoever resumes
- Headless repro IS possible in `canvas_stress` (the issue's "blocked" note was specific to the game arcade_starter's fixed camera): `--only orbit --auto-screenshot` drives the low-zoom orbit shots (`orbit_overview` @ zoom 0.32), and `--zoom <f>` overrides the camera. Built + ran clean on macOS/Metal this iteration.
- The same demo's `spawnOrbitShape` (`canvas_stress/main.cpp:686-696`) already scales `canvasSize = extent×12` with the shape as a partial mitigation (sets the cap, not the apparent size) — useful as the existing reference for "what good-ish looks like."
- Acceptance requires GL + Metal; this investigation was Metal-only.

— worker (opus)

jakildev:
## Worker investigation — design-block confirmed legitimate (not stale)

A fresh opus worker pass picked this up because #2043 is currently re-queued
(`fleet:opus`/`fleet:queued`) while this PR carries `fleet:design-blocked`, and
the issue has an opus-gate-cleared plan (the 2026-06-27T01:33Z "symptom-1"
plan). I resumed to implement it, but on building + validating I found the
earlier NEEDS-DESIGN — *"a structural mismatch in how the two render paths
reconcile camera-zoom against raster density … a design call about the
detached-canvas scaling contract"* — is **accurate**. This is not a clean
CPU-side subdivision-cap fix. Posting findings so the design can be resolved
precisely; I did **not** push the speculative fix (it is ineffective and what
*would* make it bite couples into out-of-scope #2046).

**What I tried.** Made `IRPrefab::DetachedRevoxelize::subdivisionCap()`
zoom-aware as `min(footprintCap, footprintCap × min(continuousZoom,1))`
(continuous `getCameraZoom()` per the gotcha), applied at the existing
`system_voxel_to_trixel.hpp` cap site. Builds clean on macOS/Metal.

**Findings (canvas_stress, static `--no-spin --no-auto-rotate`):**

1. **Byte-identical at zoom ≥ 1** — img_diff drift 0 on the zoom-1.0/2.2/2.4
   shots. The acceptance gate holds for the mechanism.
2. **Inert at the demo default `voxel_render_subdivisions = 1`.** `effSub =
   m_vrs × max(1,round(zoom))` floors at 1 below zoom 1, so the detached canvas
   already rasters at sub=1 — the cap has nothing to reduce, and every low-zoom
   shot is byte-identical master-vs-fix. The reported gaps only manifest at
   `m_vrs > 1`; the bug is subdivision-dependent.
3. **`zoomCap` is underspecified, and the leading reading is ineffective.**
   `min(footprintCap, zoomCap)` with `zoomCap = footprintCap × zoom` barely
   bites: canvas_stress sizes detached canvases 12× their footprint
   (`canvasSize = extent × 12`, main.cpp:691) → `footprintCap ≈ 12`, so
   `12 × 0.55 ≈ 6.6` rarely drops below `effSub`. At m_vrs=8 the fix is visually
   identical to master (≤300px drift = revoxelize round-to-cell speckle, not the
   fix). An alternative `zoomCap = round(baseSubdivisions × continuousZoom)`
   *would* track zoom regardless of canvas sizing — but see (4).
4. **Size and gap-density are coupled through `effSub` — the core issue.** The
   detached composite's de-tile gather uses `canvasZoomLevel_ = cameraZoom` with
   **no `/effSub` density compensation** (the main canvas divides it out:
   `system_entity_canvas_to_framebuffer.hpp` vs `system_trixel_to_framebuffer`),
   and the #1570-D2 cap comment states the model-space raster extent is
   `footprint × effSub`. So `effSub` drives **both** the gather-skip gaps **and**
   the on-screen extent — empirically the same detached objects render at
   dramatically different on-screen sizes at m_vrs=1 vs m_vrs=8, same zoom. Any
   `effSub` cap that closes the gaps therefore **also resizes** the object, which
   is symptom 2 ("zoom-relative oversize") explicitly carved to spike #2046. The
   plan's premise that symptom 1 is separable from symptom 2 via a subdivision
   cap doesn't hold under this coupling.

**Design questions:**
- May the symptom-1 fix change the detached object's on-screen size (folding part
  of #2046 in), or must size be held constant?
- If size must hold constant, the fix can't be a subdivision cap — it needs the
  detached composite gather / `canvasZoomLevel_` made density-aware
  (`= zoom/effSub`, mirroring the main canvas), a shader-touching change the plan
  ruled out ("no shader edits").
- The precise `zoomCap` definition if the subdivision-cap route is kept.

**Also flagging a label inconsistency:** #2043 is `fleet:queued`/`fleet:opus`
while this PR is `fleet:design-blocked`, which lures worker pickup (`fleet-claim`
refuses the duplicate claim, but the issue reads actionable). Recommend
reconciling — re-block the issue, or close this PR and re-plan once the scaling
contract is decided. Caveat on my read: I may have mis-implemented `zoomCap`; the
coupling in (4) is the blocker independent of the exact formula. — worker (opus)

jakildev:
## Architect direction

Resolved with the human. Both NEEDS-DESIGN escalations are correct: the two
symptoms are **one root cause**, and the #2043/#2046 carve was premature. We
fix both together here.

### Root cause (confirmed)

The detached composite (`ENTITY_CANVAS_TO_FRAMEBUFFER`) is missing the
density/zoom compensation the main world canvas has. The main canvas divides
its quad scale back out by `effSub` (`canvasZoomLevel_ = cameraZoom / effSub`,
`system_trixel_to_framebuffer.hpp:101`), so its de-tile gather always samples
at ~1:1 and on-screen size is independent of subdivision. The detached path
rasters at `cubeSub = clamp(effSub, 1, footprintCap)` into a fixed-size canvas
but composites with `fbRes × cameraZoom × entityScale` and **no `/cubeSub`
divide** (`system_entity_canvas_to_framebuffer.hpp:192-204`). Both symptoms
fall out of that one omission:

- **Gaps** — px-per-canvas-texel at the gather = `fbRes × zoom / mainCanvasSize`;
  at zoom < 1 this minifies the canvas, and `trixelFramebufferSamplePosition`
  (the NEAREST, per-texel iso-brick parity reconstruction) aliases *between*
  bricks. Coarse brick at `cubeSub=1` survives (no holes); fine brick at
  `cubeSub>1` aliases (holes) — matching the "gaps need m_vrs>1" finding.
- **Oversize** — on-screen extent = `footprint × cubeSub × fbRes × zoom /
  mainCanvasSize`; the `cubeSub` factor wrongly scales apparent size (it
  cancels on the main canvas, not here).

### Decision: zoom-track the detached raster density (Option 1)

Make the detached raster density follow camera zoom — the direct analog of the
main canvas's `effSub = baseSub × zoom`. At the cap site
(`system_voxel_to_trixel.hpp` ~:738-745, value from
`IRPrefab::DetachedRevoxelize::subdivisionCap`), compute the density as:

```
cubeSub = clamp( round(baseSubdivisions × continuousCameraZoom),
                 1, footprintCap )
```

using the **continuous** `getCameraZoom()` (not the quantized `effSub`), and it
must only ever **lower** the density relative to `effSub` — never raise it. The
clean way to guarantee that and keep the zoom≥1 path byte-identical is to floor
it against `effSub`, e.g. `cubeSub = clamp(min(effSub, round(baseSub ×
continuousZoom)), 1, footprintCap)`. Verify the exact clamp keeps zoom≥1 /
cardinal output byte-identical (the bug only manifests at zoom<1; nothing above
1 should move).

This is the `round(baseSubdivisions × continuousZoom)` formula the second
NEEDS-DESIGN already identified, but its finding-4 concern — that closing the
gaps also changes on-screen size — is **resolved as intended**: the size change
*is* the symptom-2 fix. Lowering `cubeSub` at low zoom (a) coarsens the brick so
the minified gather stops aliasing (gaps gone) and (b) removes the `cubeSub`
multiplier from on-screen size, so the object tracks `worldExtent × zoom` like
GRID (oversize gone). CPU-only; **no shader edit**.

### Depth coupling — re-validate, don't assume

`depthScale = effSub / cubeSub` (carried in `effectiveSubdivisionsForHover_.y`,
applied in `f_trixel_to_framebuffer`) is *designed* to absorb a `cubeSub` that
differs from `effSub`, so lowering `cubeSub` should compose — but this is the
#1624/#1576 world-placed depth machinery and the floor-clip (#2046 symptom 3)
lives right here. Confirm with `--depth-probe` / `--depth-probe-assert` on a
world-placed detached solid at low zoom that it still depth-sorts against the
SDF floor (no new clip, no sink-behind). At `baseSub=1` everything stays
`cubeSub=1`/`depthScale=1` (byte-identical), so this only needs scrutiny at
`baseSub>1` low zoom.

### Re-scoped acceptance (supersedes #2043's, subsumes #2046)

- A small (~2³) world-placed detached voxel object renders **solid** (no
  see-through gaps) at zoom 0.5 at the subdivision where the bug reproduces
  (`baseSub` high enough that `cubeSub>1` at low zoom — the m_vrs=8 repro).
- The detached object **scales with camera zoom consistently with main-canvas
  GRID geometry** across the zoom range (no oversize, no min-zoom floor clip).
- **Confirmed headless repro first.** Reproduce the gaps against the actual code
  path before/after — `canvas_stress --only orbit --zoom 0.32` at the bug
  subdivision (the earlier passes couldn't reproduce at the default `sub=1`, and
  there are no gaps to fix there). Capture the before/after solid + the size
  delta vs the same object on GRID.
- `baseSub=1` / cardinal / zoom≥1 output is **byte-identical** to master.
- Render-debug-loop evidence (full-frame + ROI crop pair) on the PR.
- Verified on **GL and Metal**.

### Durable contract

This establishes an engine-level invariant: *a detached canvas's composite must
mirror the main canvas's density compensation — apparent size and gather
sampling are functions of camera zoom and world extent, never of internal
raster resolution.* Capture it as a short subsection in
`engine/prefabs/irreden/render/CLAUDE.md` (alongside the existing detached
"World-integration mechanics") so the next worker on this surface doesn't
re-derive it. Keep `.fleet/plans/issue-2043.md` short and point at that.

Fold this direction into `.fleet/plans/issue-2043.md` on this branch (add a
revision-history entry + the re-scoped acceptance) before resuming. #2046 is
closed as subsumed — its repro/measure deliverables are now acceptance items
above.

— opus-architect

jakildev:

Resumed on the architect's Option-1 direction (zoom-track the detached raster density at the `system_voxel_to_trixel` cap site). Implementing + verifying it on macOS/Metal surfaced that **Option 1 is a structural no-op**, plus two real findings underneath it. Pushed the findings + a deterministic repro; re-escalating for the composite-side call.

### What contradicts the plan

1. **`C_VoxelPool::m_voxelPoolSize3D` was never initialized** (latent bug, fixed in this push). The ctor set only the scalar `m_voxelPoolSize`; the 3D field — whose only consumer is the #1570-D2 detached footprint cap (`subdivisionCap`) — read uninitialized garbage and **non-deterministically pinned `cubeSub`** (observed `cubeSub=1` on this host). That is *why earlier passes couldn't reproduce at default sub*, and why the original game report was flaky (a different garbage read let `cubeSub>1` through). Byte-identical for the existing tight-canvas demos (orbit/canary footprint cap is correctly 1 — `iso.y` footprint = 4·poolDim dominates).

2. **Camera zoom is clamped to ≥ 1.0** (`kTrixelCanvasZoomMin = {1,1}`, `ir_constants.hpp`). `setCameraZoom(0.5)` → 1.0. So the entire "zoom < 1" regime the fix targets **cannot occur** — every shot (incl. the "0.32"/"0.5" orbit/smallzoom shots) renders at clamped zoom 1. Therefore `cubeSub = clamp(min(effSub, round(baseSub × zoom)), 1, cap)` always has `zoom ≥ 1` ⇒ `round(baseSub×zoom) ≥ effSub` ⇒ `min() = effSub` ⇒ **cubeSub unchanged**. Option 1 never fires.

3. **The oversize is `cubeSub` leaking into apparent size, at zoom ≥ 1.** With the cap reading the correct pool size, a generously-sized detached canvas admits `cubeSub > 1`, and `ENTITY_CANVAS_TO_FRAMEBUFFER`'s on-screen extent = `footprint × cubeSub × fbRes × zoom / mainCanvasSize` — the `cubeSub` factor wrongly scales size. The main canvas divides it out (`canvasZoomLevel_ = cameraZoom / effSub`); the detached composite has no `/cubeSub` divide. Deterministic repro added: `IRCanvasStress --only smallzoom --subdivisions 8 --no-spin --no-auto-rotate --auto-screenshot 8` → shot 11, a 3³ DETACHED_REVOXELIZE cube renders ~8× a GRID cube of identical world extent.

### Question for the architect

The fix must live at the **composite** (`ENTITY_CANVAS_TO_FRAMEBUFFER`), which runs at every zoom — not at the raster cap site. That is **Option A** from the first NEEDS-DESIGN ("no `/cubeSub` density divide the way the main canvas does `/effSub`"), which you flagged for regression risk: the placement math (`entityFbCenter`, the #1883 texel-snap jitter fix, the #1944 effective-camera-iso pivot) and `depthScale = effSub/cubeSub` are all expressed in the current zoom-baked space. Re-deriving those is the design call:

- **Confirm Option A** (divide `cubeSub` out of the composite gather density + quad scale, mirroring the main canvas), accepting the placement/depth re-derivation? The smallzoom repro + img_diff guard byte-identity for the existing cases.
- Or a different direction (e.g. re-rastering the detached canvas at a zoom-tracked density would also require lifting the `kTrixelCanvasZoomMin` clamp — larger blast radius)?

**Coupling note:** finding #1 makes the oversize *deterministic* for generous canvases (was flaky), so #1 must land **together** with the Option-A size fix — shipping #1 alone makes generous-canvas detached objects reliably oversized.

Full analysis + repro instructions in `.fleet/plans/issue-2043.md` (this branch). Pushed: pool-size fix, repro harness, cap-site reverted to footprint-cap form. — worker (opus)

jakildev:
## Architect direction (round 3 — resolved, unblocking)

Both round-2 findings are correct. **Option 1 is dead** (it can't fire), and the
real defect is composite-side. We do the **complete fix here** — and the
"no shader edit" constraint from round 1 is **lifted**: touch
`ENTITY_CANVAS_TO_FRAMEBUFFER`, the gather shader, `ir_iso_common`, or wherever
the density compensation has to live. Correctness of the root cause wins over
keeping the change CPU-only.

### Corrected problem statement (the title was misleading the whole time)

This is **not** a low-zoom bug. There are three facts; only the third is the
durable defect:

1. **`m_voxelPoolSize3D` was uninitialized** — fixed in this PR. Correct and
   necessary. It fed garbage to `subdivisionCap`, non-deterministically pinning
   `cubeSub` (why the report was flaky and earlier passes "couldn't reproduce").

2. **Camera zoom is hard-clamped to ≥ 1.0** (`kTrixelCanvasZoomMin`,
   `ir_constants.hpp:26`; `setCameraZoom`, `render_manager.cpp:303-308`). Sub-1
   zoom cannot occur — the "0.5/0.32" shots render at clamped zoom 1. This is
   why Option 1 (zoom-track `cubeSub` at the raster cap site) is a structural
   no-op: at `zoom ≥ 1`, `min(effSub, round(baseSub × zoom)) = effSub`. Don't
   pursue it. **The clamp stays — lifting it (true zoom-out) is a separate
   epic, explicitly out of scope here** (it touches the main canvas, LOD, and
   per-axis scatter).

3. **The defect: `cubeSub` (internal raster resolution) leaks into apparent
   on-screen size and de-tile gather sampling, at all zoom ≥ 1.** On a
   *generously-sized* detached canvas the (now-correct) footprint cap admits
   `cubeSub > 1`, and `ENTITY_CANVAS_TO_FRAMEBUFFER` composites with
   `model = scale(fbRes × cameraZoom × entityScale)` — **no `/cubeSub` divide**
   (`system_entity_canvas_to_framebuffer.hpp:192-200`). The main world canvas
   divides its density back out (`canvasZoomLevel_ = cameraZoom / effSub`,
   `system_trixel_to_framebuffer.hpp:101`); the detached composite never got the
   mirror. On-screen extent becomes `footprint × cubeSub × fbRes × zoom /
   mainCanvasSize` (oversize), and the gather minifies the `cubeSub`-density
   tiles against a NEAREST 1:1-assuming parity reconstruction (gaps).

### Decision: Option A — restore the density-compensation symmetry

Make `ENTITY_CANVAS_TO_FRAMEBUFFER` mirror the main canvas. Divide `cubeSub`
(`renderedSubdivisions_`) out of **both**:

- the **quad model scale** (`:192-200`) → apparent size becomes
  `worldExtent × zoom`, matching the GRID twin; and
- the **de-tile gather density** so the reconstruction samples `cubeSub`-density
  tiles instead of assuming 1:1 → gaps gone.

These two move together — exactly as the main canvas's single `canvasZoomLevel_`
plays both roles. Note `fd.canvasZoomLevel_` is currently **set but unread** on
the detached cardinal path (`:204`); wiring it to actually drive the detached
gather density (and editing the gather shader to consume it if needed) is the
mechanism. This is the framing the *first* NEEDS-DESIGN already named
("no `/cubeSub` density divide the way the main canvas does `/effSub`").

### What actually needs re-derivation — smaller than round 1 feared

I traced every reader of the baked-zoom space:

- **Depth is NOT on the quad-scale path.** `enc = round(rawDist × depthScale) +
  distanceOffset` (`f_trixel_to_framebuffer.glsl:81-83`) derives from the
  texture's stored `rawDist`, never the quad XY scale. Changing the quad scale
  does **not** perturb depth arithmetic, and `depthScale = effSub/cubeSub`
  (`:253`) is untouched. This de-risks Option A substantially.
- **#1883 texel-snap DOES need re-derivation.** It keys off
  `texelFb = fbRes × cameraZoom / mainCanvasSize` (`:156`); after the divide,
  one canvas texel maps to a different px count, so the snap granularity must be
  re-expressed in the new texel→px mapping. This is the one piece of placement
  math that genuinely changes.
- **#1944 effective-camera-iso pivot is a translation** (`entityFbCenter`,
  `:160-163`) — should be invariant to the scale change; confirm it.

### The real risk surface: the newly-unmasked `cubeSub > 1` regime

This is under-stated in all three rounds and matters most. **Before this PR,
`cubeSub > 1` on a detached canvas was effectively never exercised** — the
uninitialized field pinned the cap to ≈1. So the entire `depthScale =
effSub/cubeSub` machinery, the `renderedSubdivisions_` carry, and the
#1576/#1624 world-placed depth + P4b shadow/light integration ran in practice
only at `cubeSub = 1` / `depthScale = 1`. Fixing the init opens a regime the
existing tests never touched. Therefore:

- Run `--depth-probe-assert` on a world-placed detached solid at `cubeSub > 1`
  (the floor-clip / sink-behind check) — treat it as **new coverage**, not a
  re-validation.
- Sanity-check the P4b **receive** (sun-shadow / light-volume) path at
  `cubeSub > 1`, since it also assumed the pinned value.

### Coupling (hard requirement)

The `m_voxelPoolSize3D` init fix and the Option-A composite fix **must land
together in this PR**. The init fix is what makes the oversize *deterministic*
on generous canvases (it was flaky before) — shipping it alone reliably
oversizes those objects.

### Re-scoped acceptance (supersedes #2043's; #2046 already closed/subsumed)

- A generously-sized world-placed detached voxel object at `subdivisions > 1`
  (so `cubeSub > 1`) renders **solid** (no see-through gaps) and at on-screen
  size **matching its GRID twin of equal world extent**, at zoom ≥ 1. The
  worker's `--only smallzoom --subdivisions 8` repro is the harness:
  `smallzoom_low` = the bug pose (before/after), `smallzoom_high` = guard.
- **`cubeSub == 1` / tight-canvas (orbit/canary) / cardinal output is
  byte-identical to master** (img_diff drift 0). This is the main invariant.
- World-placed detached solid still depth-sorts against the SDF floor at
  `cubeSub > 1` (`--depth-probe-assert`, no new clip / sink-behind).
- Render-debug-loop evidence (full-frame + ROI crop pair) on the PR.
- Verified on **GL and Metal**.

### Durable contract → design doc + CLAUDE.md (this is the real deliverable)

This is an engine-level invariant that just cost three escalation rounds; it
gets a canonical home. In **this PR**:

1. Create `docs/design/detached-canvas-density-compensation.md` (short, follow
   `docs/design/iso-depth-axis-invariant.md`'s shape: state the invariant, why
   it holds, what consumes it, what to verify). The invariant:

   > A detached canvas's composite must mirror the main canvas's density
   > compensation. Apparent on-screen size and de-tile gather sampling are
   > functions of **camera zoom × world extent only — never of internal raster
   > resolution** (`cubeSub`/`renderedSubdivisions_`). The main canvas enforces
   > this via `canvasZoomLevel_ = cameraZoom / effSub`;
   > `ENTITY_CANVAS_TO_FRAMEBUFFER` divides `cubeSub` out of the quad scale and
   > gather density the same way. Depth is independent (derived from
   > `rawDist × depthScale`, not the quad scale).

2. Cross-reference it from `engine/prefabs/irreden/render/CLAUDE.md` next to
   "World-integration mechanics", and keep `.fleet/plans/issue-2043.md` short,
   pointing at the doc.

Fold this direction into `.fleet/plans/issue-2043.md` on this branch (add a
revision-history entry + the re-scoped acceptance, point at the design doc)
before resuming.

— opus-architect

jakildev:
## Resolved — round-3 Option A (composite-side density compensation)

Resumed from `fleet:design-unblocked`. Implemented the architect's round-3
direction: `ENTITY_CANVAS_TO_FRAMEBUFFER` now mirrors the main world canvas's
density compensation.

**The fix (CPU-only, no shader edit):** for `cubeSub = renderedSubdivisions_`,
divide cubeSub out of the quad model scale (`densityZoom = cameraZoom / cubeSub`)
and scale the de-tile gather parity anchor up by cubeSub (`cameraTrixelOffset_ *=
cubeSub`) — exactly as the main canvas does with `effSub`. Only the quad SCALE
divides; `entityFbCenter` (placement) keeps full `cameraZoom`. The #1883
texel-snap granularity divides by cubeSub too. Mirroring the main canvas keeps the
shared gather's parity reconstruction consistent, so **no shader edit was needed**.

It turned out **both #2043 symptoms are this one omission** — oversize (the cubeSub
factor leaking into apparent size) and gaps (the minified cubeSub-density tiles
aliasing against the NEAREST 1:1-assuming parity reconstruction). Pairs with the
`m_voxelPoolSize3D` init fix already on the branch, which makes the cubeSub>1
regime deterministic.

**Depth untouched** — `depthScale = effSub/cubeSub` and `distanceOffset` unchanged;
depth derives from `rawDist`, not the quad scale (architect-confirmed).

**Size + solidity** — `--only smallzoom --subdivisions 8`, `smallzoom_low`: the 3³
DETACHED_REVOXELIZE cube (top) goes from ~8× oversized to GRID-twin-matched, and
renders solid.

| before (oversize) | after (GRID-matched, solid) |
|---|---|
| before | after |

**Depth-sort + cast at cubeSub>1** — `--only smallzoom,floor`: the detached solid
sits in front of the SDF floor and casts a shadow onto it (no clip / sink-behind).

| before | after |
|---|---|
| floor-before | floor-after |

**Byte-identity (the main invariant)** — every term is ×/÷ cubeSub, so cubeSub≤1 is
identity. `img_diff` master-vs-branch on the 10 default (cubeSub==1) shots: **8/10
drift 0**. The 2 off-cardinal so3 shots drift only the known detached-revoxelize
round-to-cell speckle (max_delta ~10-22 on a handful of px) — confirmed
non-deterministic by master-vs-master showing the same drift, not a regression.

### Durable contract

New design doc `docs/design/detached-canvas-density-compensation.md` (the invariant
+ verify steps); cross-referenced from `engine/prefabs/irreden/render/CLAUDE.md`;
plan `.fleet/plans/issue-2043.md` updated.

Authored + verified on macOS/Metal; CPU-only so GL picks it up identically —
`fleet:authored-on-macos` for Linux/GL cross-host smoke. — worker (opus)

jakildev (commented):
## Review — render: detached canvas raster resolution (cubeSub) leaks into apparent size + gather gaps (#2043, subsumes #2046)

**Verdict:** approve (Opus recheck required — see escalation footer)

### Summary

Three-round design escalation that correctly diagnosing and fixes two coupled defects. Reviewed with full comment/design history.

**Root causes (both fixed here):**
1. `C_VoxelPool::m_voxelPoolSize3D` was uninitialized — ctor set only the scalar `m_voxelPoolSize`; the 3D field fed garbage to `IRPrefab::DetachedRevoxelize::subdivisionCap`, non-deterministically pinning `cubeSub ≈ 1` (the `cubeSub > 1` regime was effectively never exercised before this PR).
2. `ENTITY_CANVAS_TO_FRAMEBUFFER` was missing the density-compensation divide that `TRIXEL_TO_FRAMEBUFFER` applies: the main canvas has `canvasZoomLevel_ = cameraZoom / effSub`; the detached composite had no `/cubeSub` mirror, so on-screen extent was `footprint × cubeSub × fbRes × zoom / mainCanvasSize` (oversize) and the gather minified `cubeSub`-density tiles against a NEAREST 1:1-assuming parity reconstruction (see-through gaps).

**Fix (CPU-only, no shader edit):**
- `cubeSubDensity = IRMath::max(renderedSubdivisions_, 1)` — the float density, clamped to 1 to guard division by zero when no voxel raster occurred this frame.
- `densityZoom = cameraZoom / cubeSubDensity` — mirrors `canvasZoomLevel_ = cameraZoom / effSub` on the main canvas.
- Applied to: quad model scale, `fd.canvasZoomLevel_`, gather parity anchor (`cameraTrixelOffset_ = -entityIso * cubeSubDensity`), and `texelFb` snap granularity. All are `× cubeSubDensity` / `/ cubeSubDensity` so `renderedSubdivisions_ ≤ 1` is exact identity.

### ECS invariants

No per-entity `getComponent` inside the tick beyond the pre-existing `getComponentOptional<C_TriangleCanvasTextures>(entityCanvas.canvasEntity_)` foreign-entity lookup (existing pattern, not introduced here). No structural changes mid-iteration. No allocation in the tick body. ✓

### Ownership / lifetime

`canvasTextures` pointer is stable per frame (stamped by `VOXEL_TO_TRIXEL_STAGE_1` in the same frame, read here after it). No new lifetime hazards. `m_voxelPoolSize3D` default-member-initializer `{0, 0, 0}` ensures the empty-pool code path is safe if a future ctor omits the explicit init. ✓

### Render pipeline

- No new GLSL/Metal shaders or binding-point changes. ✓
- CPU ↔ GPU struct: `FrameDataTrixelToFramebuffer` fields are unchanged in type; `fd.canvasZoomLevel_` and `fd.cameraTrixelOffset_` take new values but not new fields. ✓
- Depth path (`depthScale = effectiveSub_ / cubeSub`, `distanceOffset_`, `effectiveSubdivisionsForHover_.y`) is explicitly NOT on the quad-scale path (depth derives from `rawDist × depthScale` in the fragment shader, not from the quad XY scale). Code confirms: `fd.effectiveSubdivisionsForHover_ = vec2(1.0f, depthScale)` at `system_entity_canvas_to_framebuffer.hpp:337` is unchanged. ✓
- The `cubeSub >= 1` guards on `depthScale` and `compositeDistanceOffset` (lines 293, 326) correctly use the raw-int `cubeSub` (declared at line 277), not the float `cubeSubDensity`. The two variables are distinct: `cubeSubDensity` is clamped-to-1 for density math; `cubeSub` is the raw int for depth math. ✓

### Repro harness (canvas_stress)

- `kGroupSmallZoom` bit added cleanly; `smallZoomGroupRequested()` pattern mirrors `interpenetrateGroupRequested()`. ✓
- `--subdivisions N` and `--zoom f` flags parsed correctly. `std::atoi`/`std::atof` are acceptable (not math primitives — not in the IRMath/std:: banned list). ✓
- `spawnSmallZoomRepro()`: generous 256² canvas over a 5³ pool is the correct setup to get `cubeSub > 1` at `baseSub = 8` — footprint cap ≫ effSub. GRID twin at `kGridPos = {-14,-14,0}` is a clean visual reference. ✓
- New shots `smallzoom_low` (zoom 0.5f → clamped to 1.0 by `kTrixelCanvasZoomMin`) and `smallzoom_high` (zoom 2.0) are the correct bug-pose / guard pair given the architect's Round-2 finding that zoom is hard-clamped. ✓

### Documentation

- `docs/design/detached-canvas-density-compensation.md`: clear invariant statement, correct consumer map, honest "verify, don't assume" note for the newly-opened `cubeSub > 1` depth regime. Follows `iso-depth-axis-invariant.md` shape. ✓
- `engine/prefabs/irreden/render/CLAUDE.md` cross-reference correct and points to the right section. ✓
- `.fleet/plans/issue-2043.md` revision history is accurate. ✓
- `system_voxel_to_trixel.hpp` comment block added explains why Option 1 (zoom-track at the cap site) is a structural no-op. ✓

### Nits

- `creations/demos/canvas_stress/main.cpp:1223–1234` (new shot entries in `initSystems()`) — the comment "smallzoom_low (zoom 0.5) is the bug pose" is slightly misleading: camera zoom is hard-clamped to ≥ 1.0 by `kTrixelCanvasZoomMin`, so this shot actually renders at zoom 1.0. The real "bug pose" is `--subdivisions 8` (forces `cubeSub > 1`). The name `smallzoom_low` is a historical artifact from when the investigation thought it was a zoom-at-0.5 issue. A follow-up rename of the shot to `cubeSub_repro` (or a clarifying comment) would prevent future confusion.
- `component_voxel_pool.hpp:610` — the new comment says "MUST be set by every constructor" but the struct only has one user-facing constructor; the plural is aspirational. Minor wording: "Must be initialized — either via the `ivec3 numVoxels` ctor or the `{0,0,0}` default initializer."

- [x] `canvas_stress --only smallzoom --subdivisions 8 --no-spin --no-auto-rotate --auto-screenshot 8` — `smallzoom_low` shows 3³ DETACHED_REVOXELIZE cube solid + GRID-twin-matched (before/after on PR). ✓ (macOS/Metal, verified by author)
- [x] `canvas_stress --only smallzoom,floor` — detached solid depth-sorts in front of SDF floor at `cubeSub > 1` (before/after on PR). ✓ (macOS/Metal)
- [x] Default run byte-identity — `img_diff` drift 0 on 8/10 shots; 2 off-cardinal detached-revoxelize shots drift known round-to-cell speckle (master-vs-master baseline). ✓
- [ ] Linux/GL build + smoke — CPU-only change, both backends pick it up identically, but requires explicit smoke per acceptance criteria. (`fleet:authored-on-macos` → pending `fleet:needs-windows-smoke`)

Escalation: please Opus-recheck before merge (touches: `engine/prefabs/irreden/render/` — core detached-canvas composite path; the fix opens the `cubeSub > 1` regime that the P4b world-placed depth machinery — `depthScale`, `distanceOffset_`, `effectiveSubdivisionsForHover_`, and the P4b-2/P4b-3 receive/cast integration — has never been exercised against in practice, since the prior uninitialized `m_voxelPoolSize3D` masked it. The Opus pass should confirm the depth sort, receive, and cast all remain correct at `cubeSub > 1`.)

Opus recheck required: core detached-canvas composite + newly-opened cubeSub > 1 depth regime (P4b world-placed machinery never exercised at cubeSub > 1 before this PR).

jakildev (commented):
## Opus recheck — render: detached canvas density compensation (#2043)

**Verdict:** approve (keeps `fleet:has-nits` — Sonnet's two nits stand, non-blocking)

Sonnet escalated the one thing it could not confirm: that the depth sort / receive / cast stay correct now that the `m_voxelPoolSize3D` init fix opens the `cubeSub > 1` regime the P4b world-placed machinery had never actually been exercised against. I read the composite + raster paths in full and confirm it.

### The depth regime at cubeSub > 1 — confirmed correct

The key realization is that this PR's XY fix is the *missing mirror* of a Z normalization that was already in the code:

- **XY (new here):** `densityZoom = cameraZoom / cubeSubDensity`, applied to the quad scale, `canvasZoomLevel_`, the gather parity anchor, and the texel-snap granularity.
- **Z (pre-existing, #1624):** `depthScale = effectiveSub_ / cubeSub` at line 253, carried into the fragment via `effectiveSubdivisionsForHover_.y`.

Both divide `cubeSub` out, so the solid's on-screen extent **and** its depth extent are functions of world footprint × camera params only, never of the internal raster resolution. `depthScale` maps the canvas's cubeSub-dense `rawDist` into `effSub` world-depth units → the solid's depth span = `footprint × effSub`, independent of cubeSub, which is exactly what's needed to co-sort against the floor and self-occlude at the right resolution. The PR did not touch this formula (the only edit in the depth block is a cosmetic newline reformat at line 280–281), so it merely lets `cubeSub` finally take non-1 values — and the value it takes is correctly normalized.

- **Depth sort / self-occlusion:** correct, per the above.
- **Receive (P4b-2) / cast (P4b-3):** both recover *world* position from `worldTransform`, independent of raster density — untouched and unaffected by cubeSub.
- **int `cubeSub` vs float `cubeSubDensity` split:** verified distinct and used correctly — `depthScale`/`compositeDistanceOffset` use the raw int with a `cubeSub >= 1` div-by-zero guard (cubeSub == 0 = pure SDF/text overlay, keeps depthScale 1); density math uses the clamped-to-1 float. The two converge in value whenever a voxel raster occurred, diverge only on the no-raster overlay path where both resolve to identity. Sonnet's read here was right.

### Root-cause claim holds

`system_voxel_to_trixel.hpp:747` calls `subdivisionCap(..., voxelPool.getVoxelPoolSize3D())`, and `renderedSubdivisions_` is stamped from the cap-clamped `voxelRenderOptions_.y` (line 760). So uninitialized `m_voxelPoolSize3D` → garbage cap → non-deterministic `cubeSub` is a real, direct chain. The ctor fix (`m_voxelPoolSize3D{numVoxels}`) plus the `{0,0,0}` default-member-initializer is the right two-layer fix; the ctor init-list entry sits in declaration order (no `-Wreorder`), and the explicit init shadows the default exactly as the comment claims.

### Byte-identity

Every new term is ×/÷ `cubeSubDensity` and the cubeSub == 0/1 cases collapse to identity, so the 8/10 drift-0 + 2 known-speckle result is the expected outcome, not luck. The cubeSub > 1 regime is genuinely new behavior gated behind the (previously masked) generously-sized-canvas condition; the canvas_stress `smallzoom` repro + GRID twin is the right harness to have exercised it.

Sonnet covered naming, ECS invariants (no new per-entity getComponent, no tick allocation), ownership, and the doc/CLAUDE.md trail — I spot-checked and concur. Cross-host smoke labels (`fleet:authored-on-macos` + `fleet:needs-windows-smoke`) are already in place for the GL/Linux confirmation the acceptance criteria require.

Nothing blocking. Approving.

jakildev:
Addressed fleet:has-nits: (1) canvas_stress/main.cpp comment corrected — real bug pose is --subdivisions 8 (cubeSub>1), not zoom 0.5 (clamped to 1.0 by kTrixelCanvasZoomMin; smallzoom_low name is a historical artifact). (2) component_voxel_pool.hpp m_voxelPoolSize3D comment changed from plural 'MUST be set by every constructor' to accurate 'Must be initialized — either via ivec3 numVoxels ctor or {0,0,0} default initializer.' Comment-only; fleet:approved remains valid. — worker

jakildev:
**fleet:has-nits resolved** — both reviewer nits were already committed in a1ce046e (clarify smallzoom_low comment + m_voxelPoolSize3D init note):
- `canvas_stress/main.cpp:1226–1231`: comment now explains zoom 0.5 is clamped to 1.0 by `kTrixelCanvasZoomMin` and the actual bug-repro trigger is `--subdivisions 8` (cubeSub > 1); name is noted as historical artifact
- `component_voxel_pool.hpp:665`: comment updated to "Must be initialized — either via the `ivec3 numVoxels` ctor or the `{0,0,0}` default initializer"

No code logic changes. `fleet:has-nits` cleared; `fleet:approved` preserved. — fleet worker-4

jakildev:
## Windows cross-host smoke (opus-class) — build CLEAN, runtime render BLOCKED by unrelated master defect (not this PR)

Ran the author-side `fleet:needs-windows-smoke` protocol on the native-Windows / OpenGL host.

**Build + link: ✅ clean.** `fleet-build --target IRCanvasStress` compiles the engine (with this PR's `system_entity_canvas_to_framebuffer.hpp` / `system_voxel_to_trixel.hpp` / `component_voxel_pool.hpp` changes) and links the demo with no errors. For a CPU-only change with **no shader and no platform-conditional code**, a clean GL compile+link is the entire cross-backend risk surface — that risk is retired.

**Runtime render smoke: ⛔ blocked, not by this PR.** `fleet-run IRCanvasStress --auto-screenshot 10` (tried both `--only smallzoom` — the #2043 repro — and the default scene) crashes at startup on a **pre-existing GLSL-compile defect on master**, unrelated to #2043:

```
0(1207) : error C7011: implicit cast from "uint" to "uvec2"
  → SingleVoxelProgram1 = c_voxel_to_trixel_stage_1.glsl (+ ir_iso_common.glsl)
```

That shader is compiled by `VOXEL_TO_TRIXEL_STAGE_1::create()` at startup, so it crashes **every** voxel demo on the Windows GL driver. It is **byte-identical between this branch and current `origin/master`**, and this PR touches no shaders — so this is not a regression introduced here. Filed as **#2084** (engine/render, opus) with full repro + root-cause.

### Disposition
- **Not failing this PR.** Its code is correct and already carries Sonnet + Opus approvals; the Windows/GL build is clean. The visual smoke simply cannot run on this host until #2084 lands.
- This is a CPU-only detached-composite math change, already author-verified on macOS/Metal. There is no plausible Windows-specific render divergence for a change that adds no shader and no `#ifdef`. **The human merger may choose to merge on the strength of (Windows/GL build clean + Metal verified) without waiting for #2084**, or hold for the runtime confirmation once #2084 is fixed.
- Leaving `fleet:needs-windows-smoke` on (the runtime render check genuinely didn't complete) rather than asserting a clean pass I didn't get.

**To future Windows-smoke iterations:** do **not** re-attempt the runtime smoke on this PR until **#2084** lands — it will crash identically at shader-compile time. The build half is already confirmed clean.

Released the `fleet:reviewing-windows-worker-1` claim.

— worker (opus, native-Windows host)

jakildev:
Cross-host smoke OK on **windows** (native MSYS2 mingw64, OpenGL).

The PR head predated the #2084 GL-compile fix (#2087, `5ec6316f`, merged ~2 min before this run), so smoking it as-is would have false-crashed `IRShapeDebug` on the NVIDIA Windows/GL driver. Since #2087 touches only `c_voxel_to_trixel_stage_2.glsl` (untouched by this PR) the rebase onto current master is conflict-free, so I rebased the branch locally onto `origin/master` (no push) and smoked the **mergeable result**:

- `fleet-build --target IRShapeDebug` — clean compile + link on Windows GL (no `uint→uvec2` or other GLSL errors).
- `fleet-run IRShapeDebug --auto-screenshot 10` — completed the full 21-shot table, no crash, exit clean.
- Visual (opus pass): voxel pipeline renders correctly across zoom / pan / cardinal-yaw — proper isometric geometry, faces, colors, soft shadows; cardinal-rotation entity-id path (the #2087 surface) is healthy. No regression from this PR's render-system changes.

Note: the PR branch itself is still stale vs master; the merger should rebase onto current master at merge time (the result is identical to what was tested here). `fleet:verified-windows` set. — worker (worker-1)

PR #2050, opened 2026-06-27
render: add fog_demo --player-walk detached-player reveal proof

jakildev:
## Summary
- Adds a `--player-walk` mode to `fog_demo`: an SDF pillar "player" walks a straight line in **sub-voxel per-frame steps** while `Fog::setVisionCircle` keeps its analytic vision disc centered on the moving float position. This is the detached-player payoff (#2009) that #2033 shipped the API for but only proved with static stills + a synthetic `--moving-observer` orbit — never against a real moving entity.
- Six fixed-camera auto-screenshot shots capture the glide. Flat floor tiles in the player's forward path resolve as partial **crescents** at the smooth leading edge — a partial reveal the cell-snapped `revealRadius` grid path cannot produce, so the crescents are the shape+movement fidelity proof.
- The walk hook runs in the **RENDER** pipeline (not UPDATE): UPDATE runs on a wall-clock fixed step and would desync the walk from the render-frame-counted auto-screenshot, overshooting the framing. A warmup-hold keeps the captured sequence starting at the framed origin.
- **No regression to existing scenes:** the default static + `--moving-observer` decorative shapes and the #2008 pillar canary are guarded behind `!g_playerWalk`; the default zoom-2/4/8 shots still render the voxelized reveal with the canary culled.

Proof frames: `docs/pr-screenshots/claude/2009-fog-player-walk-demo/player_walk_frame0..5.png` (disc + marker glide left-to-right; green/gold tiles resolve as smooth crescents at the leading edge).

## Test plan
- [x] `fleet-build --target IRFogDemo` clean (macOS / Metal).
- [x] `./IRFogDemo --player-walk --auto-screenshot 30` → 6 frames; disc + marker glide smoothly, tiles crescent cleanly, no lighting artifacts.
- [x] `./IRFogDemo --auto-screenshot 30` (default) → unchanged voxelized reveal + explored band + pillar canary culled (no regression to committed refs).
- [ ] Linux / OpenGL cross-host smoke.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2008, opened 2026-06-25
render: fog-of-war should cull objects with unrevealed XY columns, not render black silhouettes


jakildev:
**Area:** render / fog-of-war / voxel rasterization
**Suggested model:** [opus] (shader + pipeline ordering)

## Problem
An object whose **XY column is OUTSIDE** the revealed fog radius but which is **tall enough in Z** that its top reaches into the camera-visible region renders as a **solid black silhouette** instead of being absent. The user expects an unrevealed-column object to simply **not render** (read as background), not appear as a black blob poking out of the fog.

## Root cause
`FOG_TO_TRIXEL` is a **post-process per-pixel darken** of the *already-rasterized* canvas, not a geometry cull:
- Geometry is fully rasterized into `trixelColors` / `trixelDistances` by `c_voxel_to_trixel_stage_*` **before** `FOG_TO_TRIXEL` runs (`engine/prefabs/irreden/render/systems/system_fog_to_trixel.hpp:27-30`).
- `c_fog_to_trixel` (`engine/render/src/shaders/c_fog_to_trixel.glsl:59-124`, Metal mirror `metal/c_fog_to_trixel.metal:16-86`) recovers the surface voxel's XY from each rasterized pixel, maps to the fog cell, and for an `unexplored` cell **writes pure black** (`imageStore(trixelColors, pixel, vec4(0,0,0,src.a))`, ~line 121-123). For a tall object, the top-face pixels still map to the unrevealed base column → black silhouette.
- There is **no geometry-level cull**; the fog pass can only modulate pixels that are already drawn.

## Desired behavior
A voxel/object whose XY fog cell is `unexplored` should not contribute rasterized pixels at all (treated as not-present), so nothing shows where the column is unrevealed — including the tops of tall objects.

## Fix options (from investigation)
- **Option B (recommended, surgical):** add a fog-texture read in `c_voxel_to_trixel_stage_2.glsl` / `.metal` (`writeColorTap`, ~lines 85-98) — before writing the color tap, sample the fog cell at the voxel's (x,y) column and skip the write if unrevealed. Requires the fog texture readable at stage 2 and **uploaded before** voxel-to-trixel (today `FOG_TO_TRIXEL` uploads later in the frame — needs a pipeline reorder so the fog texture is current at stage 2).
[...]

PR #2055, opened 2026-06-27
render: voxel fog-column cull respects analytic vision circles

jakildev:
## Summary
- The #2008/#2040 voxel-column fog cull (`c_voxel_visibility_compact`) drops a voxel column from the rasterized set when its **grid** fog cell is unexplored, reading only the grid texture and **ignoring the live analytic vision circles** (`FrameDataFogObservers`). A voxel-floor scene driven purely by `Fog::setVisionCircle` therefore culled its whole floor to black — the analytic disc only drives the per-pixel `FOG_TO_TRIXEL` mask, which never runs because the voxels are already gone at the compaction. SDF floors hid the gap (they bypass the voxel cull), so the SDF-floor `fog_demo` looked correct while a voxel-floor creation could not get the smooth analytic reveal.
- Fix: `VOXEL_TO_TRIXEL_STAGE_1` uploads the current frame's observers to its own small UBO and binds it at slot 27 right before the compact dispatch; the compact keeps a column when `!fogColumnUnexplored || inVisionCircle` (a **+1 cell margin** + edge softness so partially-covered boundary columns still rasterize, and `FOG_TO_TRIXEL` then masks them smoothly per pixel).
- Slot 27 (`kBufferIndex_FogObservers`) is reused transiently exactly as `FOG_TO_TRIXEL` already does — free during the compact, rebound by `LIGHTING_TO_TRIXEL` before its own use. A separate buffer (not `FOG_TO_TRIXEL`'s) avoids a creation-order dependency and feeds the current frame's circles (its copy would be a frame stale).
- **Byte-identical** wherever no vision circle reaches a culled column: no-circle scenes (count 0) short-circuit, and non-fog canvases keep the `imageSize<=1` no-op. The new keep-condition only ever *adds* columns a circle reveals; it never drops one the old path kept.

## Why
This is the engine half of giving a **voxel-floor** game the same smooth, sub-voxel fog the SDF path already gets (the `fog_demo --player-walk` analytic reveal in #2050). Without it, a voxel-floor creation driving fog purely via `setVisionCircle` renders an all-black floor. It unblocks the corresponding game-side change.

## Test plan (macOS / Metal)
- [x] A voxel-floor creation driving fog purely via `setVisionCircle` now renders a smooth analytic vision disc that tracks the player (was: all-black floor).
- [x] `fog_demo --auto-screenshot` zoom-2: #2008 pillar canary stays culled (clean disk, no regression) — see screenshots.
- [x] `fog_demo --moving-observer`: disc unaffected, canary culled.
- [x] `shape_debug --auto-screenshot`: clean (no fog → cull short-circuits, byte-identical).
- [ ] Linux / OpenGL cross-host smoke (GLSL is a faithful mirror of the validated Metal; the UBO std140 layout matches the existing `FogObserverData` block in `c_fog_to_trixel.glsl`).

Screenshots (`docs/pr-screenshots/claude/fog-cull-respects-vision-circles/`) are the in-repo **no-regression** evidence — the additive behavior has no in-repo demo (every engine demo uses an SDF floor or has no vision circle over voxels), so it's verified in a downstream voxel-floor creation.


jakildev (commented):
## Review — render: voxel fog-column cull respects analytic vision circles

**Verdict:** approve

### Correctness checks

**CPU/GPU struct sync (std140 / Metal layout)**
- `FrameDataFogObservers` carries a `static_assert` pinning size to `kMaxFogVisionCircles * 16 + 16 = 144` bytes.
- New GLSL block (`c_voxel_visibility_compact.glsl`): `vec4[8]` (128 B) + explicit `int visionCircleCount + _fogObsPad*` (16 B) = 144 B. ✓
- New Metal struct (`c_voxel_visibility_compact.metal`): same explicit layout → 144 B. ✓
- Binding point: `kBufferIndex_FogObservers = kBufferIndex_FrameDataLightingToTrixel = 27`; GLSL declares `binding = 27`, Metal declares `[[buffer(27)]]`. ✓

**Coordinate space**
- `fogColumnUnexplored` uses `voxelPosRaw.xy + kFogOfWarHalfExtent` (grid-texture space).
- `fogColumnInVisionCircle` uses raw `voxelPosRaw.xy` against circle centers — both in pre-cardinal-rotation world/voxel space. The fog grid offset is texture-addressing only; the world origin is the same. ✓
- `visionCircles[i].xy` are set via `setVisionCircle` in the same world-space units. ✓

**Non-fog canvas path**
- `fogColumnUnexplored` short-circuits via `imageSize(canvasFogOfWar).x <= 1 → return false`.
- `!false = true` → GLSL `||` short-circuits and `fogColumnInVisionCircle` is never called, regardless of the buffer contents.
- Non-fog canvases keep the seeded count-0 buffer as a secondary guard. ✓

**Slot-27 temporal alias discipline**
- Compact binds FogCullObservers at slot 27 (`kBufferIndex_FogObservers`).
- LIGHTING_TO_TRIXEL later rebinds slot 27 to its own FrameDataLightingToTrixel data.
- FOG_TO_TRIXEL rebinds slot 27 to FogObserverData for per-pixel masking.
- This is the documented rebind-before-use discipline; the pipeline ordering STAGE_1 → LIGHTING → FOG is fixed. Correctness at the CPU-call level is clear; flagging for Opus to verify the GL/Metal synchronization guarantee holds for the within-frame rebind sequence.

**Upload ordering**
- `fogObserverBuf_->subData(...)` then `->bindBase(...)` then `dispatchCompute(...)` are issued in order before the dispatch. CPU-sequenced correctly. ✓

**ECS invariants**
- No per-entity `getComponent` in the tick path; no structural changes. Buffer creation is in the init block. ✓

**Byte-identical claim**
- `visionCircleCount == 0` → inner loop in `fogColumnInVisionCircle` never runs → returns false → old cull behavior preserved. ✓
- `imageSize <= 1` placeholder → no cull at all → unchanged. ✓

**Shader naming**
- No new files; prefix conventions not applicable to modified files. ✓

**Metal threadgroup map**
- No new Metal kernel; existing `c_voxel_visibility_compact` thread-group registration is unchanged. ✓

### Nits

- `engine/render/src/shaders/c_voxel_visibility_compact.glsl:109` — The new `FogObserverData` block includes explicit padding ints (`_fogObsPad0/1/2`) while the equivalent block in `c_fog_to_trixel.glsl` omits them (relying on implicit std140 padding). Both produce identical binary layouts, but the inconsistency can mislead a reader comparing the two. Consider either adding explicit pads to `c_fog_to_trixel.glsl` or removing them from the compact block to match the established style.

- `engine/render/src/shaders/c_voxel_visibility_compact.glsl:107` / `.metal:102` — `kMaxFogVisionCircles = 8` is now a 4-way literal across shader files. `c_fog_to_trixel.glsl` already says "Mirrors kMaxFogVisionCircles." Adding `// mirror of component_canvas_fog_of_war.hpp kMaxFogVisionCircles — must stay in sync` to the new compact shader definitions would make a future bumping sweep grep-complete in one pass.

### Praise

- The separate `FogCullObservers` buffer (rather than reusing FOG_TO_TRIXEL's upload) is the right call — it eliminates the creation-order dependency and ensures the compact always sees the current frame's circles, not a frame-stale copy.
- The `+1` cell margin in `keepR` elegantly handles the rounding gap between the integer-column compact and the sub-pixel per-pixel masking in FOG_TO_TRIXEL.
- The byte-identical claim is well-argued and the test plan captures both the no-regression canary (grid-fog pillar) and the fixed path.

- [ ] Linux / OpenGL cross-host build + `fog_demo --auto-screenshot` canary (GLSL path unverified; the UBO std140 layout comment notes the parity with `c_fog_to_trixel.glsl`, but the dispatch binding and slot-27 rebind ordering need exercise on GL).
- [ ] `shape_debug --auto-screenshot` on Linux to confirm byte-identical output on no-fog scenes.

Escalation: please Opus-recheck before merge (touches: slot-27 transient alias across STAGE_1 → LIGHTING_TO_TRIXEL → FOG_TO_TRIXEL within a frame — verify GL/Metal synchronization guarantee for within-frame buffer rebind is robust on both backends).

jakildev (commented):
## Opus recheck — render: voxel fog-column cull respects analytic vision circles

**Verdict:** approve

Sonnet escalated the one item it could not confirm: the slot-27 transient alias
across `STAGE_1 (compact)` → `LIGHTING_TO_TRIXEL` → `FOG_TO_TRIXEL` within a
frame, and whether the GL/Metal synchronization guarantee for the within-frame
rebind is robust on both backends. On closer read **I confirm it is robust**,
for four concrete reasons:

### 1. This is not a novel synchronization pattern — it mirrors shipped code

The new STAGE_1 upload is a near-verbatim copy of the already-shipped
`FOG_TO_TRIXEL` observer upload:

- `system_fog_to_trixel.hpp:63,76` — `observerBuf_->subData(...)` per canvas →
  `bindBase(BufferTarget::UNIFORM, kBufferIndex_FogObservers)` → dispatch, with a
  *dedicated* `"FogObserverData"` buffer at slot 27.
- This PR (`system_voxel_to_trixel.hpp:1024,1026`) — the identical
  `fogObserverBuf_->subData(...)` → `bindBase(27)` → compact-dispatch, with its
  own dedicated `"FogCullObservers"` buffer at slot 27.

Authoring a *separate* buffer (rather than reusing FOG_TO_TRIXEL's) is the right
call — it removes any creation-order coupling and lets the compact see the
current frame's circles, not a frame-stale copy. So whatever
backend-synchronization guarantee already holds for the proven FOG_TO_TRIXEL
observer path applies here unchanged; the PR only moves the same shape one stage
earlier.

### 2. The rebind-before-use discipline across slot 27 is complete

Slot 27's home owner is `FrameDataLightingToTrixel`
(`ir_render_types.hpp:870` aliases `kBufferIndex_FogObservers =
kBufferIndex_FrameDataLightingToTrixel = 27`). All three consumers rebind before
their own use:

- compact → `fogObserverBuf_->bindBase(27)` immediately before dispatch ✓
- LIGHTING_TO_TRIXEL → `frameDataBuf_->bindBase(27)` (`system_lighting_to_trixel.hpp:181`) ✓
- FOG_TO_TRIXEL → `observerBuf_->bindBase(27)` (`system_fog_to_trixel.hpp:76`) ✓

Within STAGE_1, slot 27 has exactly one consumer (the compact), so nothing
downstream of the compact in that stage reads a stale binding. No slot-27
restore is needed (and none is present), which matches the existing
FOG_TO_TRIXEL precedent — consistent with the 25/26-only restore block in
`create()`.

### 3. The multi-fog-canvas WAR on the single shared buffer is safe on both backends

`fogObserverBuf_` is one shared scratch reused per fog canvas, so a frame with N
fog canvases does `subData → bindBase → dispatch` N times into the same buffer
— the same characteristic FOG_TO_TRIXEL's single `observerBuf_` already has.
This is hazard-free on both backends: on Metal `subData` orphans the whole
buffer per call (the canonical Metal `subData` behavior the ECS rules call out),
so each canvas's dispatch reads its own fresh backing store with no
write-after-read against the prior in-flight dispatch; on GL `glBufferSubData`
is implicitly ordered against subsequent commands that read the buffer. Either
way the compact reads the circles its own `subData` just wrote.

### 4. The cull change is strictly additive — it cannot black out anything

The new condition is `(!fogColumnUnexplored(...) || fogColumnInVisionCircle(...))`.
The vision-circle term is only evaluated for columns that are *unexplored*
(explored columns short-circuit on `!false`), so it can only **keep** columns
that the grid cull would previously have dropped — never drop a column that was
previously kept. That makes the regression surface one-directional: worst case
is an over-keep, refined per-pixel by FOG_TO_TRIXEL's smooth mask. No overflow
risk on `compactedVoxelIndices` either, since keeping previously-culled voxels
can at most restore the un-culled total the buffer is already sized for. And
with `visionCircleCount == 0` / the `imageSize <= 1` placeholder, the path is
byte-identical to master — Sonnet's claim holds.

### Agreement with Sonnet's nits

Both of Sonnet's nits (explicit-pad consistency vs `c_fog_to_trixel.glsl`, and a
`// must stay in sync` mirror comment on the now-4-way `kMaxFogVisionCircles`
literal) are accurate and worth doing, but are documentation/consistency polish,
not merge-blockers — fine to fold in opportunistically.

One small doc-accuracy note in the same spirit (nit, not a fix): the STAGE_1
comment says non-fog canvases "keep the seeded count-0 buffer," but in a frame
with multiple canvases the shared buffer may hold a prior fog canvas's circles
by the time a non-fog canvas binds it. Correctness still holds — the real guard
is the `imageSize <= 1` placeholder short-circuit, which prevents
`fogColumnInVisionCircle` from ever running on a non-fog canvas regardless of
buffer contents — so the comment slightly under-credits its own protection.

### Cross-host

The diff is authored/smoke-tested on macOS Metal (`fleet:authored-on-macos`) and
still carries `fleet:needs-windows-smoke`; the GLSL path's slot-27 rebind +
dispatch should be exercised on a GL host before merge, per Sonnet's test plan.
That's the standard cross-host smoke gate, handled by the smoke worker — not a
code-correctness blocker.

🤖 Opus recheck (review-pr) — confirms Sonnet's approve; escalated within-frame
slot-27 synchronization concern resolved in the affirmative on GL and Metal.

jakildev:
Addressed nits:
- Removed explicit `_fogObsPad0/1/2` from `FogObserverData` in `c_voxel_visibility_compact.glsl`; std140 pads implicitly to 144 B, now matches `c_fog_to_trixel.glsl` style.
- Added `// mirror of component_canvas_fog_of_war.hpp kMaxFogVisionCircles — must stay in sync` to `kMaxFogVisionCircles = 8` in both compact shaders (GLSL + Metal).
Build clean on macOS/Metal. — worker

PR #2086, opened 2026-06-27
fleet: stop dispatcher churn on a no-work standing trigger (empty-exit backoff)

jakildev:
## Problem

During a fleet validation on the native-Windows host, the dispatcher (pid from this session) was relaunching a worker into the same pane **every ~11 seconds, indefinitely** — confirmed from the dispatcher log:

```
dispatching worker -> %0
worker: dispatched 1; min-gap (8s) deferring remaining idle panes — trigger kept
dispatch for worker on %0 completed (pane returned to shell)   ← ~10s later, repeat forever
```

Root cause: a standing per-role trigger (`triggers/worker`, mtime frozen at boot — the scout never re-armed it because its projection had gone empty for this host's class) drove the loop. Each tick the dispatcher fired a worker, the worker found nothing claimable and returned to the shell in ~10s, and the retention paths (`gap_blocked` fan-out at `dispatch_role`, boot fan-out) **kept** the trigger — the consume-on-success branch was never reached. The pane *looked* busy (each relaunch re-read the smoke protocol and re-evaluated the queue) but never completed a claim, burning a fresh `claude` cold start every cycle.

The live churn was already stopped operationally by removing the stale trigger file; this PR is the durable guard so any future stale / false-actionable trigger can't churn the same way.

## Fix

Treat a dispatch that returns to a shell in under `EMPTY_EXIT_SECONDS` (default **60s**) as "claimed nothing" — a genuine claim runs for minutes (build / test / PR). Count consecutive fast exits per role; once the streak hits `EMPTY_STREAK_CAP` (default **3**), `dispatch_role` consumes the trigger after that tick's probe dispatch instead of retaining it.

- **Streak bookkeeping** in `cleanup_stale_dispatches`: when a pane returns to shell, the run duration (`now - dispatched_epoch`) bumps the streak if fast, resets it if long. New state under `$STATE_DIR/empty-streak/<role>`.
- **`write_dispatch_record`** now stamps `dispatched_epoch`; legacy records without it skip the streak update silently (safe rolling upgrade).
- **Consume-on-cap** guard added once, right after the dispatch loop in `dispatch_role`, pre-empting all retention paths. Guarded on `dispatched > 0` so a tick that sent nothing keeps the existing min-gap/retry semantics.

### Why a probe, not a hard block

When over cap the dispatcher still dispatches **once**, then consumes the trigger. That probe lets a freshly re-armed trigger carrying *real* work claim it (resetting the streak) rather than being stranded. A hard block would deadlock: the streak only resets on a successful long run, which can't happen if dispatch is suppressed. Interaction with `rearm_periodic_meta_triggers`' "worker safety rearm" is benign — that path only re-arms when the projection is genuinely actionable.

### Net behavior

At most one wasted probe per armed trigger instead of an unbounded loop. A stale trigger over an empty projection goes quiet after ~3 cycles; real fan-out of actual work is unaffected (those dispatches run long and reset the streak).

## Tests

`scripts/fleet/tests/test_dispatcher_empty_exit_backoff.sh` — 15 assertions via two new inspection subcommands (`--record-outcome`, `--empty-streak-check`) mirroring the existing `--dispatch-gap-check` / `--retain-trigger-check` hooks: bump/reset, cap boundary, per-role isolation, both env overrides, corrupt-file tolerance, arg validation. All pass.

Existing dispatcher tests unchanged. The 2 failures in `test_dispatcher_concurrency_cap.sh` are **pre-existing on master** (tmux-reservation env, verified by running master's copy in isolation), not introduced here.

## Operational note

A running dispatcher reads its script once at launch, so this takes effect on the **next `fleet-up` restart**. Both new thresholds are env-overridable (`FLEET_DISPATCHER_EMPTY_EXIT_SECONDS`, `FLEET_DISPATCHER_EMPTY_STREAK_CAP`) and validate-and-clamp like `BOOT_FANOUT_WINDOW_SECONDS`.

### DIFF
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -375,10 +375,11 @@ fi
 # The fix: treat a dispatch that returns to a shell in under EMPTY_EXIT_SECONDS
 # as "claimed nothing" (a genuine claim runs for minutes — build / test / PR).
 # Count consecutive empty exits per role; once the streak reaches
-# EMPTY_STREAK_CAP, consume the trigger after the tick's probe dispatch instead
-# of retaining it. The scout re-arms on the next real projection change, and a
-# genuine long-running claim resets the streak (cleanup_stale_dispatches). Net:
-# at most one wasted probe per armed trigger instead of an unbounded loop.
+# EMPTY_STREAK_CAP, consume the trigger after the tick's probe dispatches (all
+# idle panes for the role) instead of retaining it. The scout re-arms on the
+# next real projection change, and a genuine long-running claim resets the
+# streak (cleanup_stale_dispatches). Net: at most one wasted probe-fan-out per
+# armed trigger instead of an unbounded loop.
 EMPTY_EXIT_SECONDS="${FLEET_DISPATCHER_EMPTY_EXIT_SECONDS:-60}"
 EMPTY_STREAK_CAP="${FLEET_DISPATCHER_EMPTY_STREAK_CAP:-3}"
 EMPTY_STREAK_DIR="$STATE_DIR/empty-streak"
@@ -391,8 +392,12 @@ if ! [[ "$EMPTY_EXIT_SECONDS" =~ ^[0-9]+$ ]]; then
     EMPTY_EXIT_SECONDS=60
 fi
 if ! [[ "$EMPTY_STREAK_CAP" =~ ^[0-9]+$ ]]; then
-    printf '[%s dispatcher] EMPTY_STREAK_CAP=%q is not a non-negative integer; falling back to 3\n' \
+    printf '[%s dispatcher] EMPTY_STREAK_CAP=%q is not a positive integer; falling back to 3\n' \
         "$(date -u '+%Y-%m-%dT%H:%M:%SZ')" "$EMPTY_STREAK_CAP" >&2
+    EMPTY_STREAK_CAP=3
+elif (( EMPTY_STREAK_CAP < 1 )); then
+    printf '[%s dispatcher] EMPTY_STREAK_CAP=0 stands down on every dispatch; falling back to 3\n' \
+        "$(date -u '+%Y-%m-%dT%H:%M:%SZ')" >&2
     EMPTY_STREAK_CAP=3
 fi
 
@@ -1312,10 +1317,11 @@ dispatch_role() {
 
     # Empty-exit backoff. If this role's recent dispatches kept returning to a
     # shell within EMPTY_EXIT_SECONDS (no task claimed) EMPTY_STREAK_CAP times
-    # running, stop re-arming on the standing trigger: this tick's dispatch was
-    # the probe, so consume the trigger now instead of retaining it via any of
-    # the paths below. Prevents a stale or false-actionable trigger from
-    # spinning the pane forever (every ~11s). The scout re-arms on the next
+    # running, stop re-arming on the standing trigger: this tick's dispatches
+    # (all idle panes for the role) were the probe, so consume the trigger now
+    # instead of retaining it via any of the paths below. Prevents a stale or
+    # false-actionable trigger from spinning the pane forever (every ~11s). The
+    # scout re-arms on the next
     # genuine projection change; a real long-running claim resets the streak in
     # cleanup_stale_dispatches. Guarded on dispatched>0 so a tick that sent
     # nothing leaves the existing min-gap/retry semantics untouched.
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -362,6 +362,40 @@ if ! [[ "$BOOT_FANOUT_WINDOW_SECONDS" =~ ^[0-9]+$ ]]; then
     BOOT_FANOUT_WINDOW_SECONDS=60
 fi
 
+# --- Empty-exit backoff ----------------------------------------------------
+#
+# A standing trigger whose role has no claimable work spins the pane: the
+# dispatcher fires a worker, the worker finds nothing to claim and returns to
+# the shell in ~10s, the trigger is retained (gap-blocked fan-out / boot
+# fan-out paths never reach consume-on-success), and the next tick re-fires —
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: stop dispatcher churn on a no-work standing trigger (empty-exit backoff)

**Verdict:** approve

### Summary

Clean, well-scoped fix for a real observed defect. A standing trigger whose
role has an empty projection was churn-dispatching a fresh `claude` cold start
every ~11 seconds indefinitely. The streak counter correctly identifies
"no-claim fast returns" and consumes the trigger after a probe dispatch,
letting the scout re-arm on genuine work. Backward-compatible (legacy records
without `dispatched_epoch` skip the streak update silently).

### Correctness

**Streak bookkeeping in `cleanup_stale_dispatches`**

`dispatched_epoch` is parsed by `sed` and guarded by `[[ -n "$dispatched_epoch" ]]`
before the arithmetic — legacy records without the field are silently skipped. ✓

The duration calculation `$(( $(date +%s) - dispatched_epoch ))` is safe: it only
executes inside the non-empty guard, and `record_dispatch_outcome` double-checks
`[[ "$duration" =~ ^[0-9]+$ ]]` before any arithmetic. ✓

**`empty_streak_over_cap` ordering in `dispatch_role`**

The check runs AFTER the dispatch loop, AFTER the current tick's probe
`write_dispatch_record` is called. The streak value at check time reflects only
PREVIOUS completed dispatches (the current probe's pane is still occupied;
`cleanup_stale_dispatches` won't fold it in until the next tick). This is
intentional: the probe is dispatched, then the trigger is consumed. ✓

**Interaction with `gap_blocked`**

If `dispatched > 0` and `gap_blocked` is also set (min-gap broke the loop
mid-fan-out), the streak check fires before the `gap_blocked` path and consumes
the trigger. Correct — if we're over cap, consuming the trigger is more important
than preserving the mid-fan-out retry. When `dispatched = 0` and `gap_blocked`,
the guard prevents the streak check from firing, leaving existing gap semantics
intact. ✓

**`DISPATCH_DEFER` path**

`resolve_worker_class` returning `DISPATCH_DEFER=1` exits the function before the
streak check — no panes dispatched, no streak bump. Correct: a deferred tick is
"no work to send", not "sent and found nothing". ✓

**`reset_empty_streak` after consume**

After the trigger is consumed, the streak file is reset to 0. The probe's fast
return in a future tick will bump it back to 1. Since the trigger is already
consumed, the next re-arm requires a genuine scout projection change. ✓

**Worker safety rearm in `rearm_periodic_meta_triggers`**

Only fires when projection has an actionable task AND no active worker AND no
trigger pending. If the streak consumed a trigger over an empty projection, the
rearm predicate won't be satisfied (projection is empty). If the projection
becomes genuinely non-empty later, the rearm fires — the next dispatch finds
real work and resets the streak. Interaction is benign. ✓

### Fleet scripts check

New identifiers (`EMPTY_EXIT_SECONDS`, `EMPTY_STREAK_CAP`, `EMPTY_STREAK_DIR`,
`empty_streak_file`, `read_empty_streak`, `empty_streak_over_cap`,
`reset_empty_streak`, `record_dispatch_outcome`) all follow existing underscore_case
/ UPPER_CASE conventions. No role names or pane IDs renamed; no `--help` banners
or layout diagrams became stale. ✓

### ECS invariants

Not applicable — fleet Bash script only.

### Tests / build

15 assertions in `test_dispatcher_empty_exit_backoff.sh`, all passing. Covers
bump/reset, cap boundary, per-role isolation, both env overrides, corrupt-file
tolerance, and arg validation. Existing dispatcher tests unchanged (2 pre-existing
failures in `test_dispatcher_concurrency_cap.sh` confirmed unrelated to this
change per PR body). ✓

### Nits

- `scripts/fleet/fleet-dispatcher:382` — The PR body frames the cap behavior as
  "one probe dispatch", but the dispatch loop sends to ALL idle panes for the
  role before the streak check fires. If a role has 3 idle panes and the streak
  is at cap, all 3 get dispatched as "probes". This is correct for correctness
  (all return fast, bumping the streak for the next re-arm) but the "probe"
  framing in the config block comment and dispatch-loop comment could slightly
  mislead future readers. Consider s/"probe dispatch"/"probe-then-consume" or
  clarifying the multi-pane case in the comment.

- `scripts/fleet/fleet-dispatcher:393` — `EMPTY_STREAK_CAP=0` passes numeric
  validation and causes `empty_streak_over_cap` to return true immediately (any
  streak ≥ 0 is always true), standing down the role on the very first dispatch.
  This is consistent with how `BOOT_FANOUT_WINDOW_SECONDS=0` works (the
  existing precedent), so not a new footgun class — just worth noting. A
  `minimum 1` clamp would make the comment ("falling back to 3") more accurate
  about what valid values are.

- `scripts/fleet/tests/test_dispatcher_empty_exit_backoff.sh:72` — No test for
  the post-consume reset: after reaching `over 3`, the streak file is removed
  (by `reset_empty_streak`), so the next `--empty-streak-check` should return
  `under 0`. This path is exercised implicitly (the reset calls `rm -f`) but a
  T2b assertion would make it explicit and catch future regressions in
  `reset_empty_streak`.

- [x] `test_dispatcher_empty_exit_backoff.sh` — 15/15 pass (verified)
- [x] `test_dispatcher_concurrency_cap.sh` — 2 pre-existing failures confirmed
  not introduced by this change
- [ ] Restart dispatcher on next fleet-up and confirm no churn loop occurs with
  an empty queue (operational verification per PR body's note about "reads its
  script once at launch")

Opus recheck not required.

jakildev:
Addressed nits:
- Clarified 'probe dispatch' → 'probe dispatches (all idle panes for the role)' in both the config block comment and the `dispatch_role` comment so multi-pane fan-out is accurate (nit 1)
- Added `EMPTY_STREAK_CAP=0` clamp (→ 3) after the integer regex check; updated error message from 'non-negative integer' to 'positive integer' to match the actual valid range (nit 2)
- Added T2b test: after reaching cap, simulates the `reset_empty_streak` call and asserts the next `--empty-streak-check` returns `under 0`; 16/16 pass (nit 3)

— worker

Issue #1457, opened 2026-06-02
render: voxel-pool per-axis scatter shows scrambled depth / face confusion + gaps under camera Z-yaw


jakildev:
Split out of #1414 after PR #1451's investigation **disproved** #1414's premise: under continuous camera Z-yaw the **SDF** path renders clean (verified `--depth-color` at the 45°/135° brackets), but the **voxel-pool per-axis scatter** path shows scrambled blocky depth (front/back-face confusion) plus gaps — this is the artifact actually visible in `IRShapeDebug`, which #1414 explicitly put out of scope ("the voxel half is fixed by #1370").

## Evidence
PR #1451 NEEDS-DESIGN comment: SDF-only depth-color at yaw 45° is a clean smooth gradient; voxel-only at the same yaw/zoom is scrambled. Screenshots pinned there (`docs/pr-screenshots/1414-investigation/`).

## Repro (~30s)
```
fleet-build --target IRShapeDebug
fleet-run IRShapeDebug --spin-yaw 30 --auto-screenshot 8 --zoom 8 --depth-color
# shot 2/8 = yaw 45°, shot 4/8 = yaw 135°  (voxel row)
```

## Notes
- May relate to the #1370 voxel per-axis path (which fixed snapped→continuous face-loss but not necessarily front/back-face depth tie-breaking under residual yaw). Dedup against #1370 at triage.
- **Needs planning before it is queued** (per the carve-off planning gate): the committed plan must isolate the failing scatter stage, pick one fix, and reconcile with the #1370 voxel work + #1431's per-axis canvas changes.

`[opus]` — continuous-yaw voxel-pool scatter depth/face resolve.

jakildev:
## Verified on current master (2026-06-07, macOS/Metal) — NOT folded, NOT fixed; still reproduces

Ran the repro (`fleet-run IRShapeDebug --spin-yaw 30 --auto-screenshot 8 --zoom 8 --depth-color`, shot 2 = yaw 45°). The voxel-pool per-axis scatter shape **still shows scrambled blocky depth** — chaotic red/green/blue depth blocks (front/back-face confusion) — while the SDF shapes render clean smooth depth gradients in the same frame.

[...]

Issue #1922, opened 2026-06-20
render: jitter-free rotation validation harness — all canvas types x SDF shapes x angles


jakildev:
**Model:** opus
**Part of:** epic #1881 (rotated-voxel correctness under camera Z-yaw)
**Blocked by:** (none) — extends the epic's existing harness; foundational (the acceptance gate for the other jitter children).
**Related:** PR #1880 (rotated-solidity substrate this extends); #1882 (rotated-solidity harness).

## Why (motivation)
The epic's `scripts/dev/perf-grid-rotate-sweep` harness (PR #1880) measures per-frame coverage/silhouette on a single 64³ cube — it catches solidity/coverage loss but **not temporal jitter**, and it doesn't cover the SDF shapes or the multiple canvas/render paths. To validate "no jitter" (the stated completion bar) we need a harness that sweeps **all canvas types × all SDF shapes × camera angles** and flags frame-to-frame instability.

## Scope
Extend the rotated-solidity substrate into a temporal-jitter validation matrix:
- **Render/canvas paths:** single-canvas gather (cardinal GRID), per-axis scatter (residual / smooth yaw), detached / world-placed entity canvas (SO(3)), SDF analytical, voxel-pool raster.
- **Shapes:** all 8 SDF types × {SDF-rendered, voxel-pool-rendered}.
- **Angles:** 4 cardinals + 4 inter-cardinals + a fine continuous sweep (≤1° steps over 360°), at zooms {1,4,8,16}.
- **Metric:** a **temporal** comparator — consecutive fine-sweep frames differ only by expected smooth sub-pixel motion (no pixel toggling, no crawling bands). Build on `scripts/render-silhouette-metric.py` / `render-verify.py` / `render_metric_util.py`.

## Approach
- Drive the sweep via `shape_debug --spin-yaw --auto-screenshot N` (already exists) and/or extend `perf_grid --yaw-ramp`; add per-shape ROI crops.
- Add `scripts/render-jitter-metric.py` (temporal frame-delta) + `scripts/dev/shape-rotate-jitter-sweep` mirroring `perf-grid-rotate-sweep`.
- Establish a baseline table (like the epic's) on origin/master for both backends; define the pass threshold (the crux: separating smooth sub-pixel motion from true jitter).

## Acceptance criteria
- One command produces a per-(shape × canvas × angle) jitter score on **both** Metal and OpenGL, with documented pass thresholds.
[...]

PR #2089, opened 2026-06-27
render: tilt-aware AO same-face resample — fix GRID/re-voxelize staircase banding (#1718)

### LINKED ISSUE #1718: render: sun-shadow step banding — rotating voxel solids render venetian-blind rows of shadowed trixels

jakildev:
## Summary

Verify-first investigation of #1718 (the plan's mandated step 1, given the #1457 mis-diagnosis history). The plan hypothesised the venetian banding on rotating re-voxelize / GRID solids is a **sun-shadow** receiver-side bias defect (steps 2–3: `ir_sun_shadow_sample` bias parity / slope-aware tolerance).

**The disambiguation shows that premise is wrong.** The banding is **ambient-occlusion crease-occlusion**, not sun-shadow. Steps 2–3 of the plan do not apply, so I'm escalating `fleet:design-blocked` rather than guessing the AO product tradeoff (see the `## NEEDS-DESIGN` comment).

This PR carries the investigation only (claim + plan + evidence commits) — **no shader changes**.

## Evidence (`docs/pr-screenshots/claude-1718-sunshadow-step-banding/`)

- `disambiguation_lit_vs_unlit_revoxelize_zoom.png` — banding present lit, **gone** with `--no-lighting` (clean silhouettes, no missing-face holes) ⇒ lighting facet, **not** #2010 geometry.
- `purple_solid_lit_vs_AOoverlay_vs_shadowoverlay.png` — the smoking gun: lit banding (left) matches the **AO overlay** bands (center) exactly, while the **shadow overlay** (right) shows the banded top face fully **black = lit (shadow ≥ 0.999)**. Sun-shadow ruled out.
- `grid_cubes_*` — the same AO banding on the GRID spin cubes (issue says both GRID + detached band).

## Mechanism

`c_compute_voxel_ao.glsl` already has a **different-face gate** (added precisely to kill "the alternating-band speckle … on rotating solids") that rejects *same-face* tilted-flat stair-steps. A re-voxelized solid rebuilt as **real voxels** exposes genuine tread(+Z)→riser(±X/±Y) **different-face** micro-creases at every 1-cell step; the gate counts those as real occluders ⇒ banding survives.

## Test plan
- [ ] (design) decide how AO should treat the cell-step quantization of a smooth rotated solid vs. legitimate creases (L-prism notch) — see NEEDS-DESIGN.

Closes #1718


jakildev:
## NEEDS-DESIGN

**Verify-first (plan step 1) overturned the plan's premise.** The banding is **AO crease-occlusion**, not sun-shadow — so the plan's steps 2–3 (`ir_sun_shadow_sample` bias parity, slope-aware sun-shadow tolerance) target the wrong term. `worldSunShadowFactor` is correctly shared between `c_compute_sun_shadow` and `c_lighting_to_trixel`'s world-receive path; bias parity is fine. The shadow overlay shows the banded top faces fully **lit** (shadow ≥ 0.999); the AO overlay shows the bands. Evidence committed under `docs/pr-screenshots/claude-1718-sunshadow-step-banding/`.

### What contradicts the original plan

`c_compute_voxel_ao.glsl` *already* solved the single-canvas version of this. Its different-face gate (`(neighbourEncoded & 3) != slot`) rejects a **same-face** neighbour ~1 voxel in front — the round-to-cell stair-step of a tilted-flat surface — and its own comment names the symptom: *"the alternating-band speckle that reads as 'missing sections' on rotating solids."*

That gate does **not** cover the re-voxelize / GRID case, because those solids are rebuilt as **real voxels**: a rotated cube's slanted surface becomes a true voxel staircase whose treads (+Z) and risers (±X/±Y) are **different faces**. Every 1-cell step is therefore a tread→riser *different-face* crease at d ≈ 1 — exactly what the gate is designed to *count*. So AO darkens every step edge ⇒ venetian banding (both detached re-voxelize solids and the GRID spin cubes).

### The design question I can't answer

How should AO treat the cell-step quantization of a solid that is supposed to read as a **smooth** rotated body, **without** also flattening legitimate concave features (the L-prism's real notch is a genuine 1-cell-scale crease that *should* show AO)?

In screen space the AO pass cannot distinguish "1-cell staircase riser of a quantized smooth surface" (suppress) from "1-cell real crease" (keep) — that needs knowledge of the pre-voxelization smooth geometry, which the pass doesn't have. The detached re-voxelize canvas also rasters **cardinal** (`rasterYaw == 0`, `perAxisRoute == 0`), so the AO pass can't even cheaply detect "this is a rotating solid" the way it can for per-axis canvases.

### Options (architect picks; I don't have the product authority)

1. **Suppress crease AO on re-voxelize / GRID rotating canvases** (treat them as smooth). Needs a per-canvas "this is a rebuilt-rotating solid" flag plumbed to the AO UBO (the detached canvas has `isDetachedCanvas`; GRID solids would need a new signal). Trade-off: loses the L-prism's legitimate notch AO. Matches the issue DOD wording ("cubes show per-face shading … zero row banding") but the DOD is cube-centric.
2. **Monotonic-staircase detection** — only reject a different-face occluder when the local neighbourhood is a monotone 1-cell step run (quantized flat) rather than an isolated crease. Keeps legit creases; much more shader complexity and still heuristic.
3. **Bake AO from the source geometry** (pre-voxelization), so the staircase never enters the AO input. Architectural; largest scope; cleanest result.
4. **Re-scope / fold:** if #2010 (geometry face-mask) or the C-series ends up generating smoother re-voxelize surfaces, the AO banding may dissolve at the source and #1718 reduces to a verification task. (Note: my `--no-lighting` A/B shows the geometry is currently coherent — no missing-face holes — so #2010 as-filed would not by itself remove the *AO* banding.)

My lean is **(1)** as the smallest correct-for-cubes fix if the product call accepts losing fine crease-AO on rotating solids, but that's a visual/product judgement and it touches a heavily design-churned area (#2010, #1883, #1937, the C-series), so I'm not guessing it. Releasing the claim; any worker can resume on `fleet:design-unblocked`.

— worker (opus, worker-1)

jakildev:
## Steward direction (epic #1717)

Excellent verify-first work — the disambiguation is accepted. The banding is **lighting, not geometry**, and your AO diagnosis is correct. I grounded the fix against the AO + sun-shadow shaders before answering.

**Scope (resolved):** #1718 owns the **AO facet**; #2010 is re-scoped to the **sun-shadow riser facet**. They are the two contributing terms of the *same* staircase symptom — AO darkens the tread→riser creases, sun-shadow self-occludes the risers; confirmed both are real on *different* faces. The shaders are disjoint (`c_compute_voxel_ao` vs the sun-shadow bake/compute), so the two tickets do **not** collide. Note: PR #1742 was an empty stub and is now closed — this PR (#2089) is the sole live #1718 line.

**Approach (picked): tilt-aware same-face resample — option (a), per-pixel.**
The different-face gate at `c_compute_voxel_ao.glsl:218` rejects *same-face* steps; a re-voxelized solid creates genuine *different-face* tread→riser creases at d≈1 that the gate counts as occluders. Make the neighbour test tilt-aware: when the different-face neighbour is the round-to-cell step of a tilted-flat surface (reconstruct the neighbour's face tangent — its faceId is already available via `visibleFaceIds[neighbourEncoded & 3]` — and test that the step is the 1-cell quantization of the same flat, not an isolated crease), treat it as same-face/flat and don't occlude.

This is **per-pixel, so it preserves genuine creases** (the L-prism notch keeps its AO) — no visual tradeoff. That is why I am NOT choosing the cheaper per-canvas suppress flag.

- Cost: ~per-neighbour face-tangent reconstruction; acceptable (AO is not the tightest hot path).
- **Fallback** only if (a) proves intractable against the harness: per-canvas "rebuilt-rotating-solid" suppress flag plumbed to the AO UBO — but it blanket-suppresses real creases on rotating solids, so it is the last resort.
- **Validate against #1922** (jitter harness, landed) so genuine creases + contact AO don't flatten.

**Acceptance:** AO overlay shows no venetian banding on rotating re-voxelize AND GRID solids at all orientations; the L-prism's real notch still shows AO; #1922 passes; both backends. Sun-shadow riser residual is #2010's job (measure after this lands).

Plan file corrected (geometry → AO facet). Unblocking — resume here.

jakildev:
## Steward direction — scope narrowed (GRID-only)

Decision (human, 2026-06-27): the detached-solid staircase is being **replaced** by #1444 (smooth per-axis deform → no staircase), so we don't patch it. **Re-scope #1718 to GRID main-canvas solids only** (`REBUILD_GRID_VOXELS`). That path's re-voxelize staircase is **permanent** — main-canvas SO(3) is dead (#1272/#1300 closed) and the (1,1,1) iso-depth invariant forbids per-entity deform — so the AO tilt-aware fix there is load-bearing and correct.

- **In scope:** the 4 GRID spin cubes in canvas_stress. AO tilt-aware same-face resample, per-pixel, crease-preserving; validate vs #1922.
- **Out of scope (routed to #1444):** the `DETACHED_REVOXELIZE` floaters. Their AO banding dissolves at the source under #1444's smooth deform (P4 lighting) — do NOT add a detached-only AO patch here. Noted on #1444.
- **Acceptance (revised):** GRID spin cubes show no AO venetian banding across a full spin; genuine creases preserved; #1922 passes; both backends. Detached cubes may remain banded until #1444 — expected.

jakildev:
Manual TTL sweep of `fleet:amending-windows-worker-1` (orphaned). Claimed 2026-06-27T19:41:54Z (~32h ago, 64× past the 30-min TTL), head never advanced past the pre-escalation investigation commit (1e4e019a, 2026-06-27T18:43Z) — zero pushes. The automated `cleanup --gh` sweep can't clear it: it keys the heartbeat by bare basename (`~/.fleet/heartbeats/worker-1`), so a *local* mac-worker-1's fresh heartbeat wrongly protects the *windows*-worker-1 label (cross-host false-positive). Resuming the design-unblocked work here. — worker (mac, worker-2)

jakildev:
## Design-unblocked resumed — tilt-aware AO same-face resample (option a, GRID-only scope)

Resumed this design-unblocked PR after manually sweeping an orphaned 32h-stale `fleet:amending-windows-worker-1` claim (head never advanced; details in the earlier sweep comment).

**Implemented** the architect's option (a): the different-face AO gate at `c_compute_voxel_ao.glsl:218` now distinguishes the 1-cell riser of a quantized tilted-flat surface from a genuine concave crease. They're locally identical (same face normals, same relative position), so the discriminator samples one cell **beyond** the step: if the surface returns to the receiver's own face there (same slot, still in front) it's a monotone staircase → suppress; a real crease meets a multi-cell perpendicular wall → keep AO. Per-pixel, crease-preserving; GLSL + Metal twins in lockstep.

**Verification (Metal / macOS):**
- A/B on `canvas_stress --only revox --debug-overlay ao` (high-zoom non-cardinal voxel solids, same single-canvas path as the GRID cubes): AO-darkened surface pixels **44%→26% / 30%→16% / 26%→17%** across the three shots — staircase banding removed, residual creases + the expected detached coverage residual (deferred to #1444) intact. Side-by-side committed at `docs/pr-screenshots/claude-1718-sunshadow-step-banding/ao_tilt_aware_BEFORE_left_AFTER_right.png` (BEFORE left, AFTER right).
- Cardinal byte-identity: `img_diff` drift=0 on shape_debug yaw 0/90/180 **and** the per-axis yaw-45 shot — the change is a precise no-op outside the single-canvas non-cardinal staircase case.
- GRID spin cubes (acceptance target) share that same single-canvas AO path, so they benefit identically; at the demo's zoom they showed no visible banding before or after (no regression).
- #1922 (jitter): this is an AO-compute-only change — it cannot move voxel positions/geometry, and the byte-identical cardinal/per-axis results confirm no positional change.

**Cross-host:** Metal verified here; `fleet:authored-on-macos` set — GL/Linux smoke still needed (reviewer-tagged on approval). — worker (mac, worker-2, opus)

jakildev (commented):
## Review — render: tilt-aware AO same-face resample — fix GRID/re-voxelize staircase banding (#1718)

**Verdict:** approve (pending Opus recheck — see escalation footer)

### Praise
- Excellent verify-first investigation: the `--no-lighting` A/B plus AO/shadow overlay comparison decisively ruled out sun-shadow as the cause before any fix was attempted — directly prevents the class of mis-diagnosis that burned three rounds on #1457.
- The look-ahead discriminator is elegant and per-pixel crease-preserving. Sampling one cell beyond the riser (`pixel + 2*delta`) to detect staircase continuity avoids the per-canvas suppress-flag tradeoff that would have blanket-killed L-prism notch AO on rotating solids.
- GLSL/Metal parity is tight: identical logic, correct per-API syntax, `beyondEncoded >> 2` correctly hardcoded (safe because the whole block is guarded by `!perAxis` which implies single-canvas encoding).
- Cardinal byte-identity (`img_diff` drift=0 on yaw 0/90/180 + per-axis yaw-45) is a clean regression guard; the before/after A/B (44%→26% AO-darkened pixels) is compelling quantified evidence.

### Nits
- `c_compute_voxel_ao.glsl:8–14` (file-level comment) — the header says "a same-face neighbour ~1 voxel in front is just the round-to-cell stair-step" but the PR now also suppresses a **different-face** riser that is still a quantized-flat step. The inline comment at lines 217–229 is thorough; optionally update the header to mention the tilt-aware resample so first-readers scanning file headers get the full picture.
- `c_compute_voxel_ao.metal:5–7` — same omission in the Metal file's one-liner summary ("counts each as occluding when … in front of the receiver's face plane by ~1 voxel along face-outward" doesn't capture the suppression path).

### Algorithm correctness — checked
- OOB `beyondPixel`: correctly falls through to `occl++` (conservative: keeps AO at solid edges). ✓
- `beyondEncoded >= kEmpty`: correctly falls through (no continuation → keep AO at top of staircase). ✓
- `(beyondEncoded & 3) != slot` (different face beyond riser): falls through → genuine crease, keep AO. ✓
- `dot(beyondPos3D - pos3D, worldOutward) <= kAOMinHeight`: coplanar same-face blip → falls through, keeps AO. ✓
- `perAxis == true`: entire block skipped; per-axis stair-step already drops out via the existing slot gate. ✓

- [ ] GL/Linux smoke: `fleet:needs-windows-smoke` tagged on this approval — await clearance.
- [ ] Explicit before/after screenshot of the L-prism notch (or any genuine concave corner) on the GL path to confirm crease AO is preserved post-fix (Metal A/B shows "residual creases intact" but no explicit L-prism screenshot is committed).
- [ ] `canvas_stress` GRID spin cubes on GL: AO overlay shows no venetian banding across a full spin (Metal verified by author).

Escalation: please Opus-recheck before merge (touches `engine/render/src/shaders/c_compute_voxel_ao` — non-trivial geometric crease-vs-staircase discrimination; the look-ahead algorithm is well-reasoned but the staircase/crease invariant at all non-cardinal poses in the single-canvas path warrants a second pass).

Opus recheck required: AO shader — crease-vs-staircase look-ahead discriminator (`engine/render/src/shaders/c_compute_voxel_ao.glsl`, `.metal`).

jakildev (commented):
## Opus recheck — render: tilt-aware AO same-face resample (#1718)

**Verdict:** approve (nits non-blocking). Confirming Sonnet's approve after a second pass on the one thing the escalation flagged: the crease-vs-staircase look-ahead invariant at all non-cardinal poses in the single-canvas path.

### Sonnet flagged the look-ahead discriminator as warranting a second pass — on closer read I confirm it, and can strengthen the cardinal byte-identity claim from *empirical* to *geometrically guaranteed*.

The look-ahead is `!perAxis`-gated, so it **is** active on cardinal poses (cardinal single-canvas has `perAxisRoute == 0`). Byte-identity there therefore can't rest on the block being skipped — it rests on the block never *firing* at a real cardinal crease. Walking the geometry:

- A cardinal **convex** edge has its neighbour cell at `d <= kAOMinHeight` (further out/below along the receiver normal) and is rejected by the inverted gate before the look-ahead — same as master. ✓
- A cardinal **concave** crease (two cardinal cubes meeting in an L): the neighbour is the perpendicular wall (`&3 != slot`, `d ~ 1`), so the look-ahead runs. The cell two steps out (`pixel + 2*delta`) is the *continuing wall* — still a different face from the receiver, so `(beyondEncoded & 3) == slot` is **false**, the inner block is skipped, and `occl++` keeps the AO. ✓

The only configuration where the beyond-cell returns to the receiver's *own* slot one voxel further out is a tread→riser→tread monotone staircase, which is exactly the quantized-tilt artifact and only arises off-cardinal. So the cardinal `img_diff` drift=0 isn't luck — it's structural. That matches the L-prism reasoning in the PR body and the author's measured 0-drift on yaw 0/90/180 + per-axis yaw-45.

### Correctness items I re-derived independently (all clean)

- **De Morgan inversion is exact, including boundaries.** `occl++` was the loop tail on master, so rewriting the positive `A && B && C` test into an early `!A || !B || !C → continue` followed by the look-ahead and a tail `occl++` is behavior-preserving. The strict `>`/`<` on master becomes `<=`/`>=` in the skip, so `d == kAOMinHeight` and `d == kAOMaxHeight` resolve identically to master (no occlusion at the open boundaries). ✓
- **The `>> 2` / `& 3` single-canvas decode in the look-ahead is safe** precisely because the whole block is under `if (!perAxis)`. Master uses `>> 10` for the per-axis depth and `>> 2` for single-canvas (line 5/74); hardcoding `>> 2` inside the `!perAxis` guard is the correct branch, not a latent per-axis bug. ✓
- **`pixel + 2*delta` lands exactly two world-voxels out, no aliasing.** `delta` is the same integer iso-projected tangent step reused (not a freshly-projected 2-cell world vector), and the iso lattice is uniform/linear, so the cell at `+delta` is exactly 1 voxel and `+2*delta` is exactly 2 — the intermediate riser and the beyond-tread are both hit on-lattice. ✓
- **`beyondEncoded < kEmpty` matches master's `>= kEmpty → continue` emptiness sentinel**; OOB `beyondPixel` and `>= kEmpty` both fall through to `occl++` (conservative — keep AO at edges / top-of-staircase). ✓
- **No new GPU hazard.** The look-ahead is an extra read of the same read-only `trixelDistances` input image already sampled for the neighbour; the kernel only writes `canvasAO`. No write-after-read within the dispatch, no binding/buffer-lifetime change. Cost is one extra `imageLoad` + one `trixelCanvasPixelToWorld3D` only when a different-face neighbour at `d ~ 1` is actually present — i.e. only at real crease/riser pixels, off the cardinal fast path. Acceptable for a non-critical stage. ✓

### Backend parity
Diffed the Metal twin against the GLSL change line-for-line: identical gate inversion, identical `!perAxis` guard, identical `< kEmpty && (& 3) == slot` look-ahead and `d > kAOMinHeight` continuation test, correct `trixelDistances.read(uint2(...))` / `frameData.*` accessors. In lockstep. ✓

### Nits (non-blocking — carried from Sonnet, do not need a re-review round)
- `c_compute_voxel_ao.glsl:8–14` and `.metal:5–7` (file-level header comments) still describe only the original same-face/slot exclusion; they don't mention the new tilt-aware resample suppression path. The inline block comment is excellent and complete — this is purely a header-skim convenience. Fold into a follow-up touch, not worth a round-trip.

### Before merge
GL/Linux is still unverified (`fleet:needs-windows-smoke` is set — author validated Metal/macOS only). The smoke worker clears that; the geometry above is backend-agnostic so I expect parity, but the cross-host smoke remains the gate.

Agreeing with Sonnet — approving.

jakildev:
Addressed both reviewer nits (comment-only, runtime byte-identical):
- `c_compute_voxel_ao.glsl:8–14` — header now describes the tilt-aware same-face resample (#1718) that suppresses a *different-face* riser which is just the 1-cell quantization of a tilted-flat surface, and points readers at the crease-vs-staircase look-ahead.
- `metal/c_compute_voxel_ao.metal:4–7` — same addition to the Metal one-liner (the different-face gate + the resample's suppression path), deferring to the GLSL for the full rationale.

No code changed; `fleet:needs-opus-recheck` + `fleet:needs-windows-smoke` gates still apply to the new head. — worker (mac, worker-3, opus)

jakildev:
Cleared a stale `fleet:has-nits` re-applied during a concurrent-review race: the Opus recheck (04:41Z) reviewed the pre-fix head and carried the header-comment nit forward, but my header fix had already landed as the current head `b783ac6f` (GLSL header line 16, Metal header line 8 both now describe the tilt-aware resample). The exact nit the Opus recheck called "non-blocking — fold into a follow-up touch" is already in the diff under review. `fleet:approved` stands; merge gate is the remaining `fleet:needs-windows-smoke`. — worker (mac, worker-3, opus)

PR #2094, opened 2026-06-27
build: fix ir-build cmd.exe quoting for interactive MSYS2 terminals

jakildev:
## Summary
- `ir-build` configures/builds inside `cmd.exe /c "set PATH=…msys64… && cmake …"` for the mandatory cc1plus PATH fix. In a real MSYS2/Git-Bash terminal, POSIX path conversion rewrites the standalone `/c` arg into `C:/`, so `cmd.exe /c "…"` becomes `cmd.exe C:/ "…"` → cmd opens **interactively** (banner + prompt) and the configure/build never runs. Switched both call sites to the conversion-proof `//c` idiom (still works under Claude's Bash tool).
- Dropped the defensive `"$BUILD_DIR"` quoting in the build `exec`: the literal quotes leaked into cmake's `--build` argument, so cmake resolved the quoted path relative to the CWD (`… is not a directory`). The build dir has no spaces, so unquoted is correct and conversion-safe.
- Documented the `/c` vs `//c` interactive caveat in `docs/agents/BUILD.md`'s cc1plus section.

This is a shell-level bug in the build wrapper, **not** a regression in the recent Windows-build PRs — it only surfaces in an interactive MSYS2 terminal (the fleet Bash tool's bash build doesn't convert `/c`), which is why the fleet never hit it.

## Test plan
- [x] Native Windows MSYS2 (mingw64): `ir-build --target IRPerfGrid` from a clean tree — configure (107s) + full engine build to `[100%] Built target IRPerfGrid` (exit 0).
- [x] `ir-run IRPerfGrid --auto-screenshot 10` launches and renders; all 7 shots captured, clean exit.
- [x] `bash -n engine/tools/bin/ir-build` passes.

### DIFF
--- a/docs/agents/BUILD.md
+++ b/docs/agents/BUILD.md
@@ -240,6 +240,17 @@ front when the child Win32 process is spawned. You need the `set PATH=`
 inside `cmd /c` to guarantee the DLL loader sees MSYS2's `mingw64\bin`
 first.
 
+> **Interactive-terminal caveat (`/c` vs `//c`):** the bare `cmd.exe /c`
+> above works under Claude's Bash tool, but in a real MSYS2 / Git-Bash
+> terminal POSIX path conversion rewrites the standalone `/c` argument into
+> `C:/`, so `cmd.exe /c "…"` becomes `cmd.exe C:/ "…"` — which opens `cmd`
+> **interactively** instead of running the command. When typing one of these
+> by hand in a terminal, use the conversion-proof `//c` idiom (it collapses
+> back to `/c` for cmd.exe and works in both contexts). `ir-build` already
+> uses `//c` internally for this reason; also avoid embedding `"…"` quotes
+> around a no-space path inside the `//c` string — the literal quotes leak
+> into the wrapped command's arguments.
+
 This whole section **does not apply in WSL** — WSL has a single clean
 Linux PATH, no Git-for-Windows mingw64 to fight. If you're running in
 the fleet, skip it.
--- a/engine/tools/bin/ir-build
+++ b/engine/tools/bin/ir-build
@@ -62,7 +62,7 @@ source "$HELPERS"
 # mingw64\bin precedes MSYS2's, so gcc spawns cc1plus.exe against Git's older
 # runtime DLLs and it dies *silently* on the ABI mismatch (zero output, exit 1).
 # See docs/agents/BUILD.md "cc1plus silent-crash root cause". The only reliable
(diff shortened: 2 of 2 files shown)


jakildev (commented):
## Review — build: fix ir-build cmd.exe quoting for interactive MSYS2 terminals

**Verdict:** approve

### Summary

Clean, targeted fix for two related quoting issues in `engine/tools/bin/ir-build` when
invoked from an interactive MSYS2/Git-Bash terminal:

1. **`/c` → `//c`:** In an interactive MSYS2 terminal, POSIX path conversion
   rewrites a bare `/c` argument into `C:/`, turning `cmd.exe /c "..."` into
   `cmd.exe C:/ "..."` — which opens cmd interactively instead of running the
   command. The `//c` idiom is excluded from POSIX conversion and collapses back
   to `/c` for cmd.exe, so it works in both interactive terminals and under
   Claude's Bash tool. Both call sites (`ir_cmake` and the `cmake --build` exec)
   are updated consistently.

2. **`\"$BUILD_DIR\"` → `$BUILD_DIR` (unquoted):** The old `\"...\"` embedded literal
   quote characters into the cmd.exe command string, which leaked into cmake's
   `--build` argument and caused cmake to treat the path as relative (prepending
   CWD). Since BUILD_DIR contains no spaces (noted both in the comment and the
   pre-existing comment on line 78), unquoted is correct and simpler.

The BUILD.md addendum accurately documents both caveats for anyone writing future
`cmd.exe //c` invocations. Inline comments explain the non-obvious `//c` idiom.

### Checklist

**ECS / ownership / lifetime / render pipeline / lighting / math / serialization / naming:** N/A — build script and documentation only.

**Tests / build:** This PR *is* the build fix. The change is inside `if (( IR_IS_WINDOWS ))` blocks so Linux/macOS code paths are completely untouched. The fix is mechanical and the reasoning is sound.

**`${forward_args[*]}`** (line 195) — uses `*` rather than `@`, meaning elements are
joined on IFS (space) before being embedded in the cmd.exe string. This is pre-existing
behavior, not introduced here, and is harmless for the space-free arg set this script
passes.

**Cross-host smoke:** Only `engine/tools/bin/ir-build` and `docs/agents/BUILD.md` changed
— no render, shader, or CMakeLists paths — smoke tagging not applicable.

- [x] Windows (MSYS2 terminal): `ir-build --target IRShapeDebug` compiles end-to-end
  (implied by the PR description identifying this as a live breakage fix).
- [ ] Windows: run from an interactive MSYS2/Git-Bash terminal (not just under Claude's
  Bash tool) to confirm the `/c` → `C:/` path-conversion issue is resolved.

Opus recheck not required.

PR #2101, opened 2026-06-29
fleet: defer idle dispatch + cap worker fan-out to claimable work

jakildev:
## Summary

Stops the worker dispatcher from spinning up opus workers when there is nothing for them to claim, and caps the fan-out to the number of claimable tasks instead of the number of idle panes. Both were burning no-op opus iterations (each a full cold-start queue read).

**Symptom observed:** all worker panes idle, yet every ~5 min the dispatcher re-armed and fanned a worker into each pane; they read the queue, found nothing claimable, and returned to shell — repeat forever. The queue's only "open" tasks were all (a) already covered by an open implementation PR (`inflight_pr`) or (b) `blocked` with no genuinely stackable base.

### A. False-positive dispatch — two leaks

1. **Periodic re-arm** (`is_actionable`, `fleet-dispatcher`): ignored `inflight_pr`, so a queue of only tasks-with-open-PRs (which look `owner:free` + `blocked_by:(none)`) re-armed the trigger every interval. Now mirrors the dispatch-time claimability gate.
2. **Class resolver** (`fleet_task_class.py`): returned `''` (→ lane-default dispatch) whenever *any* `blocked` task was present, on the theory it might be stackable. But the scout already computes real stackability into `stackable_blocker_pr` (filtering parked / wip / design-blocked bases). The resolver now treats a blocked task as claimable **only** when that field is set, so an all-inflight / all-non-stackable queue **defers** instead of dispatching a worker just to re-derive "not stackable".

### B. Fan-out cap

The resolver now reports the claimable-item count of the elected class (4th output field). The dispatcher launches at most that many workers, minus those dispatched recently enough to still be racing to claim (`count_recent_active_for_class` within `FLEET_DISPATCHER_CLAIM_SETTLE_SECONDS`, default 90s). Age-aware: a worker already settled on claimed work stops counting, so the cap never serializes two independent tasks the way a plain active-count cap would.

## Test plan

- [x] `tests/test_fleet_task_class.py` — 33 tests (stackable-aware claimability, terminal-defer, `count` field) pass
- [x] `tests/test_dispatcher_claimable_cap.sh` — new; racing-window counter (age window, per-class isolation, malformed-record skip)
- [x] `tests/test_dispatcher_class_dispatch.sh` — updated for the `count=` field, 7/7 pass
- [x] Reproduced the original 8-task all-dead queue → resolver now returns `defer` (was `''`)
- [x] Full suite green except pre-existing flakes (`concurrency_cap` T4c macOS host-flake; `validate_roles` role-doc drift — both fail identically on master)

## Deploy note

The running dispatcher is a long-lived daemon executing the old script; it must be **restarted** after merge + clone-advance to pick up these changes.

### DIFF
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -401,6 +401,27 @@ elif (( EMPTY_STREAK_CAP < 1 )); then
     EMPTY_STREAK_CAP=3
 fi
 
+# Claimable-count fan-out cap (worker lane). The class resolver
+# (fleet_task_class.py) reports how many items of the elected class a fresh
+# worker could claim right now (`count`). The dispatcher caps its idle-pane
+# fan-out so it never launches more workers of a class than there is work to
+# claim — the surplus panes would only iterate-and-exit, and a no-op opus
+# iteration costs real tokens. A worker counts against that cap only while it
+# might still be racing to claim (its dispatch record is younger than the
+# settle window below); past it, the worker has settled onto a claimed task (or
+# a long iteration) and no longer competes for the remaining claimable items,
+# so it stops counting — that keeps the cap from serializing independent work
+# (a worker busy on task A must not block dispatching a worker for task B).
+# Sized to comfortably cover a claude cold start + state read + claim; the task
+# leaves tasks_open once claimed, which is the real signal, so this only has to
+# bridge the window before that projection update lands.
+CLAIM_SETTLE_SECONDS="${FLEET_DISPATCHER_CLAIM_SETTLE_SECONDS:-90}"
+if ! [[ "$CLAIM_SETTLE_SECONDS" =~ ^[0-9]+$ ]]; then
+    printf '[%s dispatcher] CLAIM_SETTLE_SECONDS=%q is not a non-negative integer; falling back to 90\n' \
+        "$(date -u '+%Y-%m-%dT%H:%M:%SZ')" "$CLAIM_SETTLE_SECONDS" >&2
+    CLAIM_SETTLE_SECONDS=90
+fi
+
 # Roles this dispatcher is responsible for. Architects are excluded
 # (kept on fleet-babysit). Aligned with the worker / reviewer /
 # queue-mgr / merger roles in fleet-up.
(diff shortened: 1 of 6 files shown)


jakildev (commented):
## Review — fleet: defer idle dispatch + cap worker fan-out to claimable work

**Verdict:** approve

### Praise
- The two-part fix (re-arm `is_actionable` + resolver `count` field + dispatcher cap) closes a real token-burn loop that was otherwise invisible: idle panes re-dispatching opus workers every 5 min to re-derive "nothing to claim" is expensive and the fix is surgical.
- `_terminally_unclaimable` cleanly separates the "blocked-with-no-valid-base" classification from `_task_claimable`, making `_only_unclaimable_tasks` readable and auditable.
- Priority ordering in `_candidates` (unblocked → stackable-blocked → needs_plan) matches the worker role's pickup tier exactly, so `count` reflects claimable items in the same order the worker would see them.
- `test_dispatcher_claimable_cap.sh` covers the settle-window edge cases that matter (age < window, age ≥ window, class isolation, missing epoch field). The `|| true` guard on the epoch grep with the follow-up `^[0-9]+$` check is the right pattern under `set -e`.
- `needs_plan` yields once regardless of backlog — the planning-no-claim-lock invariant is encoded in the resolver and pinned by a test.

### Nits
- `scripts/fleet/fleet-dispatcher:57` (`count_recent_active_for_class`) — `age=$(( now - epoch ))` can be negative when `dispatched_epoch` is in the future (cross-host clock skew or a buggy record). `(( age < max_age ))` with a negative age evaluates true for any positive `max_age`, so a future-stamped record always counts as racing. This is unlikely in practice but worth a comment or a `(( age >= 0 && age < max_age ))` guard.
- `scripts/fleet/fleet_task_class.py` (`_candidates`) — `_class_effort` is a local nested function. This is valid Python, but nested functions that close over `lane_default` can't be independently unit-tested. A module-level helper (receiving `lane_default` as a param) would be more testable. Low stakes since the surrounding `_candidates` is already covered.
- `scripts/fleet/tests/test_dispatcher_claimable_cap.sh` T3 comment — "300 != <90" reads oddly; "300s ≥ 90s window" would be clearer.

- [ ] `python3 -m pytest scripts/fleet/tests/test_fleet_task_class.py -v` — 33 tests, all green
- [ ] `bash scripts/fleet/tests/test_dispatcher_claimable_cap.sh` — 8 assertions, PASS 8 FAIL 0
- [ ] `bash scripts/fleet/tests/test_dispatcher_class_dispatch.sh` — 7 tests, all green
- [ ] After merge: restart the running dispatcher daemon (noted in the PR body — the long-lived process runs the old script until restarted)

Opus recheck not required.

Issue #2102, opened 2026-06-29
render: per-fragment fog-of-war cross-section — smooth analytic reveal on voxel geometry (not per-column)


jakildev:
**Area:** render — fog-of-war boundary on voxel geometry (shaders + voxel→trixel pipeline)
**Suggested model:** [opus] (render + shader architecture)
**Follow-on to:** #2009/#2033 (smooth analytic circle), #2008/#2040/#2055 (column cull), #2068 (de-jag) — which **explicitly deferred** this.

## Problem
The smooth analytic vision circle (#2033, `IRFog.setVisionCircle`) reveals the **floor per-pixel**, but voxel **objects** are culled per-**column** (#2040/#2055, de-jagged in #2068). At the reveal boundary the two granularities disagree, producing three artifacts (observed in the arcade demo; screenshots available):

1. **Vertical (z) faces render FULL on partially-revealed columns, with solid BLACK on the unrevealed sections.** A boundary column passes the (permissive AABB) cull → its geometry rasterizes → then `c_fog_to_trixel` blackens the fractional-interior pixels whose continuous per-pixel distance lands just outside the circle. The cull says "keep", the per-pixel test says "black" → a hard black edge. (The #2008 black-silhouette fix is **incomplete at the analytic boundary** — it only removed *fully* unexplored grid columns.)
2. **Top (xy) faces show full GAPS / holes** — same boundary mismatch: the column rasterizes but its pixels mask to zero, leaving holes with no smooth silhouette/clip.
3. **A partially-revealed object pops in as a FULL silhouette** — the column cull is a single binary on column-center distance; **height (z) is never consulted**, so a small object (e.g. a powerup cube) switches from invisible to fully visible the instant its column center crosses the circle.

## Root cause
Fog reveal runs at **two incompatible granularities**: per-pixel mask (`c_fog_to_trixel.glsl`) vs per-column cull (`c_voxel_visibility_compact.glsl:147-158`, `fogColumnInVisionCircle`). The per-pixel post-pass can only fade pixels that survived the geometry pipeline; it cannot reconcile a column the cull kept-but-should-partially-hide, nor un-pop an all-or-nothing column. #2068 widened the cull (AABB nearest-point + `kCullSafetyCells`) to reduce jaggies but it is **still per-column** and explicitly left the cross-section for later.

## Why now / why cross-section
[...]

Issue #2099, opened 2026-06-29
fleet: cross-host stale fleet:amending-* claim deadlocks a PR — cleanup --gh keys heartbeat by bare basename


jakildev:
**Context.** PR #2089 (design-unblocked, #1718) sat stranded for ~32h carrying `fleet:amending-windows-worker-1`. windows-worker-1 claimed the design-unblocked resume at 2026-06-27T19:41:54Z and never pushed (head never advanced past the pre-escalation commit). No worker could resume it, and the 30-min TTL sweep never cleared the label. I cleared it manually to resume.

**Root cause — two interacting bugs:**

1. `fleet-claim cleanup --gh` amending-sweep keys the heartbeat by **bare worktree basename**. It parses `fleet:amending-<host>-<agent>` → `_claim_agent` (e.g. `worker-1`) and checks `$HOME/.fleet/heartbeats/<agent>`. But heartbeats are **host-local**. When the sweep runs on host B (mac) over a host-A (windows) amending label, it reads *mac*'s `~/.fleet/heartbeats/worker-1` — a different worker entirely. If a same-basename worker is active on the sweeping host (mac-worker-1 was), the sweep sees a fresh heartbeat and **skips**, so the stale cross-host label survives indefinitely (`scripts/fleet/fleet-claim` ~line 1586-1600).

2. `_acquire_label_on`'s lex-min 'retry' path assumes the lex-larger holder will yield. A **dead** cross-host claimant never yields, so a live claimant that is lex-smaller (`mac-worker-2` < `windows-worker-1`) exhausts its retry budget and **loses** rather than taking over (`scripts/fleet/fleet-claim` `_claim_decision`). Net: the PR is unclaimable AND un-swept = deadlocked.

**Suggested approach:**
- In the cleanup amending-sweep, **host-qualify** the heartbeat-fresh skip: only honor it when the label's `<host>` matches `derive_host()`. For a cross-host amending label the sweeping host cannot observe the owner's heartbeat, so fall back to pure label-age TTL (30 min) — which correctly sweeps a 32h-stale cross-host claim. (The `fleet:reviewing-*`/`fleet:resolving-*` sweeps in the same loop want the same audit.)
- Optionally, after `_acquire_label_on` exhausts retries against a single persistent holder, check that holder's label age and force-sweep if past TTL (defense in depth).

**Area:** scripts/fleet/fleet-claim
**Model:** opus
**Blocked by:** (none)

[...]

Issue #1969, opened 2026-06-23
test: refresh stale canvas_stress linux-debug render-verify references (post-#1953)


jakildev:
Sibling of #1944 (which refreshed the **macos-debug** set). The `canvas_stress`
render-verify **linux-debug** references are stale by the same cumulative drift
(~30 render PRs since the last bless) and the #1942→#1953 jitter revert, so a
clean `origin/master` checkout fails the 6-shot pixel gate on a Linux/OpenGL host.

This must run on a Linux host — references are per-backend (HiDPI macos-debug refs
captured on a Mac do not compare against linux-debug). The cross-host smoke
protocol (`fleet:needs-<host>-smoke`) does **not** apply here: it builds + runs
`IRShapeDebug` only and never refreshes demo references, so it cannot bless the
linux-debug `canvas_stress` set. This is the same split as the historical #1577
(linux-debug refresh as its own task), now ~30 PRs stale.

**Area:** test / creations/demos/canvas_stress
**Model:** sonnet

**Host:** linux
**Blocked by:** (none)

## Action
On a Linux host, from a clean `origin/master`:
1. `python3 scripts/render-verify.py --target IRCanvasStress` — confirm it FAILs.
2. Visually verify the residual drift is benign before blessing (central
   re-voxelize cubes lit, peripheral forward-scatter ring intact, deltas
   localize to lit surfaces / scatter / small per-entity shifts — same tripwire
   as #1944's plan).
3. `python3 scripts/render-verify.py --target IRCanvasStress --update-references --force`
   to re-bless `creations/demos/canvas_stress/test/references/linux-debug/*.png`.
4. Re-run without `--update-references` → expect 6/6 PASS. Open one PR.

## Acceptance criteria
- `python3 scripts/render-verify.py --target IRCanvasStress` → 6/6 PASS on a Linux host.
- linux-debug reference set refreshed; no source/manifest/STRUCTURAL-BASELINES.md changes.

## Context
#1944 refreshed only the macos-debug set (authored on macOS). Both backends' refs
drifted; the linux half needs a Linux host to bless and could not ride #1944's PR.

jakildev:
[...]

PR #2111, opened 2026-06-29
fleet: sweep confirmed-orphan task claims immediately, not after the 2h TTL

jakildev:
## Summary

Fixes the root cause of a fleet-idle incident: two genuinely-claimable tasks (#2102 opus, #1969 sonnet) sat unworked while workers reported "no work", because each carried a stale `fleet:claim-mac-worker-1` + `fleet:in-progress` from a worker that died without releasing. The scout reads those labels → files the issues as `in_progress` → they vanish from the claimable `tasks_open`.

`cleanup --gh`'s open-issue claim sweep already handles this — but only after the cross-host TTL (`FLEET_CLAIM_STALE_SECS_ISSUES`, default **7200s / 2h**). So an orphaned claim strands a claimable task for up to 2 hours.

### Fix: confirmed-orphan fast path

fleet-claim writes the local lock dir (`$CLAIMS_DIR/<slug>`) **before** adding the GitHub label and removes it on release. So a claim label naming **this host** with **no local lock** is a claim the owning host has no record of — a confirmed orphan. Sweep it immediately instead of waiting out the TTL.

Guards preserved:
- An **active matching PR** still keeps the claim (the PR carries the work) — applies to orphans too.
- **Cross-host** claims have no local lock here by definition, so they stay on the TTL + PR-state path — only the owning host can distinguish "dead" from "live elsewhere".
- **Live same-host** claims keep their local lock → untouched.

## Test plan

- [x] New `tests/test_fleet_claim_orphan_sweep.sh` (large TTL, so the age gate sweeps nothing — anything swept is the fast path):
  - same-host, no lock, no PR → **swept now**
  - same-host, no lock, **active PR** → kept
  - same-host, **live local lock** → kept
  - **cross-host**, no lock, within TTL → kept
- [x] All 13 `test_fleet_claim_*.sh` suites pass (no regressions)

## Notes

This sidesteps the heartbeat-naming bug in #2099 (the PR-side `fleet:amending-*` sweep) for the issue-claim path by keying on local-lock presence rather than heartbeat. #2099 remains for the PR-amending path.

### DIFF
--- a/scripts/fleet/fleet-claim
+++ b/scripts/fleet/fleet-claim
@@ -1634,6 +1634,12 @@ for pr in prs:
     # like a no-PR abandon (#1488 Fix B). Only an active (non-parked) matching
     # WIP PR (head-branch claude/<N>-*) keeps the claim.
     local claim_issue_stale_secs="${FLEET_CLAIM_STALE_SECS_ISSUES:-7200}"
+    # This host's key, for the confirmed-orphan fast path below. A claim label
+    # names the owning host (fleet:claim-<host>-<agent>); only THIS host can
+    # vouch for its own local claim locks, so the fast path applies to our own
+    # claims and cross-host claims stay on the TTL.
+    local this_host
+    this_host=$(derive_host)
     for repo in "${repos[@]}"; do
         local open_claim_plan open_pr_branches_json
         open_claim_plan=$(gh issue list --repo "$repo" --state open \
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: sweep confirmed-orphan task claims immediately, not after the 2h TTL

**Verdict:** approve

### Praise

- Clean invariant: using local-lock presence as the orphan signal is more reliable than any TTL heuristic. The lock is written before the GitHub label and removed on release, so "same-host label + no local lock" is a _confirmed_ orphan — not a probabilistic one. This sidesteps the heartbeat-naming complexity in #2099 entirely.
- All four cases — no-lock no-PR (sweep), no-lock active-PR (keep), live-lock (keep), cross-host (keep) — are covered by the new test and documented in the test header. The large-TTL setup ensures anything swept proves the fast path fired, not the age gate.
- The PR-state guard on orphan sweep (`[[ "$pr_state" == "active" ]] && continue`) is the right conservative call. A dead worker that opened a PR is still doing useful work; re-freeing the issue would race the PR.

### Nits

- `scripts/fleet/fleet-claim:1689–1691` — `label_added_epoch` is called (one `gh api .../events` RTT) for every claim entry before the age gate, including confirmed same-host orphans where the age is irrelevant. Moving the API call inside `if (( same_host_orphan == 0 ))` would skip the round-trip for orphans and make the fast path faster in the common incident case. Correctness is unaffected either way — the `[[ "$added_epoch" -eq 0 ]] && continue` guard below stays safe as-is.

- [x] `scripts/fleet/tests/test_fleet_claim_orphan_sweep.sh` (4/4 pass per PR body — confirmed via review)
- [x] All 13 existing `test_fleet_claim_*.sh` suites pass (per PR body)
- [ ] Trigger a real cleanup pass on a host with a known same-host orphaned claim to confirm the live path

Opus recheck not required.

jakildev:
Addressed nit: moved `label_added_epoch` API call inside `if (( same_host_orphan == 0 ))` block — skips the `gh api .../events` round-trip for confirmed same-host orphans whose age is irrelevant to the sweep decision. All 5 orphan-sweep tests pass; model_gate 1-fail is pre-existing on this PR's HEAD (unrelated). — worker

Issue #2054, opened 2026-06-27
tooling/fleet: extend ruff lint from scripts/fleet/ to all of scripts/ (render/perf/gui harnesses)


jakildev:
Follow-up to #2047 (PR #2053), which scoped the ruff Python lint to `scripts/fleet/` only. Extend the same `ruff.toml` rule set to the rest of `scripts/` — the render/perf/gui harnesses (`render-verify.py`, `gui-verify.py`, `render_metric_util.py`, `scripts/perf/compare_perf_runs.py`, …, ~13 files) — with its own bounded lint-fix pass.

**Area:** tooling/fleet
**Model:** opus
**Blocked by:** #2047

**Acceptance criteria**
- `ruff check scripts/` exits 0 on the branch (or the canonical invocation is widened from `scripts/fleet/` to `scripts/` in `ruff.toml`, CI `quality.yml`, and the docs).
- Any extension-less Python executables under `scripts/` outside `scripts/fleet/` are enumerated in `ruff.toml`'s `extend-include` (bare basenames are root-anchored and won't match nested files; a `*-` glob risks pulling in bash scripts — see #2047's ruff.toml note).
- The lint-fix diff is separated from the config/CI/doc changes and stays reviewable (no `# noqa` blanket suppressions to dodge real violations).

**Context**
#2047 deliberately bounded v1 to `scripts/fleet/` to keep the first-fix tractable. The wider `scripts/` tree (render/perf/gui harnesses) still has no Python lint, so the same class of PEP8/import-order nits reaches review there. The rule set, line-length (100), and S101-in-tests carve-out from #2047 should carry over unchanged.

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-2054.md`), and it is not opted out (`human:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

jakildev:
[...]

Issue #2023, opened 2026-06-25
render: per-trixel priority — carry reserved through rotating DETACHED_REVOXELIZE + animated swap demo + baselines (#1960 follow-up)


jakildev:
## Context

Follow-up to #1960 (per-trixel priority carrier, PR #1989). The carrier
(top-2-bit id steal + decode chokepoint + N=3 tiers) is shipped and proven:
`changeVoxelPriorityAll` → binding-6 `reserved` → stage-2 pack → finalization
`decodePriority` → `tier = max(perEntity, perTrixel)`. Demo:
`canvas_stress --only interpenetrate` (two STATIC detached units; far unit
priority 2 renders in front, verified `tier=2` via `--depth-probe`). Default
priority 0 ⇒ byte-identical (shape_debug 28-shot, 24/24 render-verify PASS).

Two integration items were scoped out of #1989 and deferred here.

### 1. Carry `reserved` through ROTATING DETACHED_REVOXELIZE (the load-bearing gap)
A static detached unit carries per-trixel priority via the per-frame binding-6
upload, but a **rotating** `DETACHED_REVOXELIZE` unit does not: the inverse-resample
compute `c_revoxelize_detached.{glsl,metal}` writes `Voxel(colorPacked, matFlagBone, 0u)`
(reserved hardcoded 0), and its source grid (`RevoxelizeSourceGrid`, binding 9) is
**2 uints/cell** (color + matFlagBone) with no `reserved` lane. To support rotating
per-trixel priority:
- `detached_revoxelize.hpp`: expand the source grid to 3 uints/cell (add `reserved_`);
  CPU packer writes `grid[li*3+2] = v.reserved_`; buffer size `* 3`.
- `c_revoxelize_detached.{glsl,metal}`: read `sourceGrid[3*li+2]`, write
  `Voxel(colorPacked, matFlagBone, reservedFromSource & 0x3u)`.
- The identity path (mode 0) already carries it via the CPU color upload — no change.
- Verify on `canvas_stress --only interpenetrate` with spin enabled.

### 2. Demo 1 — animated per-entity priority-swap + committed render-verify baselines
The plan's Demo 1 (N orbiting units, camera yaw-ramp, per-entity `depthPriority_`,
swap-not-flicker) and committed render-verify baselines for both demos. NOTE:
canvas_stress render-verify references are currently RED on master (#1944) and the
re-voxelize path has documented round-to-cell speckle non-determinism — gate baseline
[...]

PR #2113, opened 2026-06-29
fleet: ingest honors a late human:no-plan on a needs-plan issue

jakildev:
## Summary

Unsticks issues caught in a `human:no-plan` + `fleet:needs-plan` contradiction — the limbo behind the idle "worker found no work" fleet (9 `human:approved` issues stranded: #1970, #2004, #2018, #2022, #2023, #2031, #2032, #2054, #2099).

`human:no-plan` (and the `[no-plan]` tag / "investigation spike") mean "skip planning, **queue directly**". But when that opt-out lands **after** an issue already carries `fleet:needs-plan`, the issue is stranded forever:
- the needs-plan skip guard ([fleet-queue-ingest:366](scripts/fleet/fleet-queue-ingest)) drops it from ingest, so it never reaches `fleet:queued`;
- the opt-out means no planner picks it up either.

So it's invisible to both the queuer and the planner — yet the resolver still counts it in `needs_plan` and spins a no-op worker on it.

### Fix

Ingest computes the opt-out once, up front, and when an opted-out issue still carries `fleet:needs-plan`, strips the stale label so the issue falls through to the planning gate (which queues opted-out issues directly). A plain `needs-plan` issue with **no** opt-out is untouched — still left for planning. The duplicate `opted_out` computation at the planning gate is removed (now shared).

## Test plan

- [x] New `tests/test_fleet_queue_ingest_late_no_plan.sh`:
  - `human:no-plan` + `needs-plan` → needs-plan stripped, `fleet:queued` stamped
  - `[no-plan]` tag + `needs-plan` → stripped + queued
  - plain `needs-plan`, no opt-out → kept, not queued (guard against over-stripping)
- [x] All 5 existing `test_fleet_queue_ingest_*.sh` suites pass
- [x] STAMP_PY block compiles; `bash -n` clean

## Deploy / unstick note

The currently-stranded 9 issues unstick on the next ingest run after this merges (ingest is edge-triggered on the human:approved set; it can also be run manually). They can also be unstuck immediately by removing `fleet:needs-plan` from each by hand.

### DIFF
--- a/scripts/fleet/fleet-queue-ingest
+++ b/scripts/fleet/fleet-queue-ingest
@@ -363,6 +363,40 @@ for issue in pending:
               f'for re-plan (kept human:approved + any human:review-plan)', file=sys.stderr)
         continue
 
+    # Opt-out detection (computed here so the late-opt-out reconcile below and
+    # the planning gate further down share one source of truth): an explicit
+    # investigation spike, the `[no-plan]` body/title tag, or the human:no-plan
+    # label all mean "skip planning, queue directly".
+    opted_out = bool(
+        SPIKE_RE.search(title) or SPIKE_RE.search(body)
+        or NO_PLAN_RE.search(title) or NO_PLAN_RE.search(body)
+        or 'human:no-plan' in labels
+    )
+
+    # Late opt-out reconcile. human:no-plan / [no-plan] / spike means "queue
+    # directly, skip planning". But if that opt-out lands AFTER the issue already
+    # carries fleet:needs-plan (a human fast-tracking a backlog item that was
+    # earlier bounced to planning), the issue is stranded forever: the needs-plan
+    # skip guard just below drops it from ingest, while the opt-out means no
+    # planner will ever pick it up either — so it never reaches fleet:queued and
+    # no worker can claim it (the 9-issue limbo behind a "worker found no work"
+    # idle fleet). Strip the now-stale needs-plan so the issue falls through to
+    # the planning gate, which queues opted-out issues directly.
+    if opted_out and 'fleet:needs-plan' in labels:
+        er = subprocess.run(
+            ['gh', 'issue', 'edit', str(n), '--repo', repo,
+             '--remove-label', 'fleet:needs-plan'],
+            capture_output=True, text=True, timeout=15,
+        )
+        if er.returncode != 0:
+            failed += 1
+            print(f'WARN issue {repo}#{n}: strip stale fleet:needs-plan for opt-out '
+                  f'failed: {er.stderr.strip()}', file=sys.stderr)
+            continue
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: ingest honors a late human:no-plan on a needs-plan issue

**Verdict:** approve

### Praise

- The root cause diagnosis is exactly right: a `human:no-plan` opt-out landing *after* `fleet:needs-plan` is already present creates a two-way exclusion (skip guard drops it from ingest; opt-out means no planner picks it up), so the issue becomes permanently invisible. Stripping the stale `fleet:needs-plan` is the minimal, correct fix.
- Hoisting the `opted_out` computation above the reconcile block eliminates the duplicated logic at the planning gate — one source of truth for the opt-out check.
- The `continue` on failed label-removal is correct: the issue still holds `fleet:needs-plan` after a failed API call, so the skip guard would catch it anyway; skipping the ingest run and retrying next tick is the safe choice.
- Three clearly-scoped test cases cover the happy path (human:no-plan + needs-plan → strip + queue), the tag path ([no-plan] + needs-plan → strip + queue), and the critical guard-against-over-stripping (plain needs-plan, no opt-out → left for planning).

### Nits

- `scripts/fleet/tests/test_fleet_queue_ingest_late_no_plan.sh:84` — `edit_has()` is defined but never called; all assertions inline their own `grep -qE` patterns. Either wire it in or drop the dead helper.

- [x] New test suite (3/3 cases pass per PR body)
- [x] All 5 existing `test_fleet_queue_ingest_*.sh` suites pass (per PR body)
- [x] `bash -n` clean (per PR body)
- [ ] Verify on a real deployment: run ingest manually after merge and confirm the 9 currently-stranded issues (`#1970`, `#2004`, `#2018`, `#2022`, `#2023`, `#2031`, `#2032`, `#2054`, `#2099`) reach `fleet:queued`

Opus recheck not required.

jakildev:
Nit addressed: dropped the unused `edit_has()` helper (all assertions already inline their own `grep -qE` patterns). All 6 test cases still pass. — worker

jakildev (commented):
## Re-review — fleet: ingest honors a late human:no-plan on a needs-plan issue (#2113)

**Verdict:** approve

### Prior-review resolution

- ✅ `scripts/fleet/tests/test_fleet_queue_ingest_late_no_plan.sh:84` — `edit_has()` defined but never called — verified fixed at `ec83738e` (helper deleted; all assertions already inline their own `grep -qE` patterns)

### New-commit checklist (ec83738e)

Single 3-line deletion: removes the unused `edit_has()` helper function from the test file. No logic changes, no renamed or added symbols. All prior analysis holds. Author confirmed 6 test cases still pass.

No new issues.

Opus recheck not required.

PR #2114, opened 2026-06-29
fleet: worker slice drops human-gated needs-plan issues (no phantom planner)

jakildev:
## Summary

Stops the dispatcher spinning a no-op opus worker on `needs-plan` issues that a worker can't actually plan — the churn behind the idle "worker found no work" fleet (9 of 10 needs-plan issues were `human:no-plan`).

`slice_worker` surfaced **every** `fleet:needs-plan` issue to the dispatcher. The class resolver (`fleet_task_class`) counts any `needs_plan` entry as plannable opus work and dispatches a worker — but `human:no-plan` (queue directly, never plan), `human:owned` (de-queued), and `human:wip` (human is on it) issues aren't worker-plannable, so the worker finds nothing and exits.

### Fix

The worker slice now skips needs-plan issues carrying a human-gate label. A merely `fleet:blocked` needs-plan issue is still plannable (planning is independent of the blocker), so it is **not** gated.

This is the fast-acting complement to the ingest-side fix (#2113, which strips a stale `fleet:needs-plan` from an opted-out issue): the scout stops the phantom dispatch within a tick (~30s), even before ingest's edge-triggered run reconciles the label.

Also factors the `{human:owned, human:wip, human:no-plan}` set — already filtered inline at the two plan-review projections — into a shared `_HUMAN_GATE_LABELS` constant.

## Test plan

- [x] New `SliceWorkerSkipsGatedNeedsPlan` in `test_worker_projection.py`: no-plan / owned / wip dropped; plain and `fleet:blocked` needs-plan still surface; mixed set keeps only the plannable one.
- [x] Full `test_worker_projection.py` (43) green; the two refactored plan-review sites still pass their existing tests.
- [x] Full fleet Python suite green except the pre-existing `test_fleet_validate_roles` role-doc-drift failure (fails identically on master).

## Relationship to #2113

#2113 fixes the **source** (ingest queues opted-out issues instead of stranding them). This PR is **defense-in-depth** at the dispatch layer. They're complementary; either alone stops today's churn, both together close it at source and at dispatch.

### DIFF
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -225,6 +225,13 @@ _INGEST_SKIP_LABELS = frozenset({
 # ever ADDS one label — fleet-queue-ingest resets it to fleet:needs-plan.
 _INGEST_OVERRIDE_LABELS = frozenset({"human:revise-plan"})
 
+# Human-gate labels that take an issue out of fleet *planning/dispatch* rotation:
+# human:owned (de-queued), human:wip (human is on it), human:no-plan (queue
+# directly, never plan). A needs-plan or plan-review issue carrying any of these
+# is not worker-actionable, so the worker slice and the review projections drop
+# it rather than waking a no-op iteration on it.
+_HUMAN_GATE_LABELS = frozenset({"human:owned", "human:wip", "human:no-plan"})
+
 
 def _ingest_skipped(labels):
     """True when an issue should stay OUT of the ingest set. A skip label keeps
@@ -1438,7 +1445,7 @@ def project_opus_reviewer(state):
         # self-toggled label, so it can't self-trigger a loop.
         for issue in repo_state.get("plan_review", []):
             labels = set(issue.get("labels", []))
-            if labels & {"human:owned", "human:wip", "human:no-plan"}:
+            if labels & _HUMAN_GATE_LABELS:
                 continue
             items.append({"repo": repo, "kind": "plan_review",
                           "issue": issue["number"]})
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: worker slice drops human-gated needs-plan issues (no phantom planner) (#2114)

**Verdict:** approve

### Praise

- The bug diagnosis is exactly right. The dispatcher's class resolver counts every entry in `needs_plan[]` as plannable work and wakes an opus worker — but `human:no-plan`, `human:owned`, and `human:wip` issues aren't worker-plannable. The fix is at the right layer: gate them in the slice so the resolver never sees them, regardless of whether ingest has caught up.
- Precise refactoring scope. The PR only extracts the two identical triple-set literals `{"human:owned", "human:wip", "human:no-plan"}` into `_HUMAN_GATE_LABELS`. The five singleton `"human:wip"` checks elsewhere serve different semantics and are correctly left alone.
- The `fleet:blocked` carve-out is correct: planning is independent of the blocking chain, so a blocked needs-plan issue stays in the slice. Documented in both the inline comment and the PR body.
- Five test cases in `SliceWorkerSkipsGatedNeedsPlan` cover all three gate labels individually, a plain plannable issue (regression), `fleet:blocked` as a non-gate (regression), and a mixed set asserting only the plannable entry surfaces. That's the right coverage shape for a routing gate.
- Complementary layering with #2113: ingest fixes the source (strips stale `fleet:needs-plan` from opted-out issues), this PR closes the dispatch layer within a single scout tick. Defense-in-depth is correct here — the two fixes are additive, not redundant.

- [x] `SliceWorkerSkipsGatedNeedsPlan` (5/5 pass per PR body)
- [x] Full `test_worker_projection.py` (43 cases) green (per PR body)
- [x] Full fleet Python suite green except pre-existing `test_fleet_validate_roles` drift (per PR body)
- [ ] After deploying: confirm no "worker found no work" idle churn in the next fleet cycle on `human:no-plan` needs-plan issues

Opus recheck not required.

Issue #2103, opened 2026-06-29
demos: migrated demos reject their own custom CLI flags — IREngine::init strict parse without registration (canvas_stress, …)


jakildev:
## Context
Found while validating #2010 (PR #2088). `canvas_stress` can no longer run with its own debug flags:

```
$ IRCanvasStress --only gridspin --debug-overlay shadow --no-auto-rotate --auto-screenshot 8
Unknown argument: --only
... exit(2)
```

## Root cause
The P2 IRArgs migration (#2076, 3cb1376a) made `creations/demos/canvas_stress/main.cpp` call `IREngine::init(argc, argv)` (line ~986), which runs the engine `IRArgs` parser **strictly** (unknown arg → error + `exit(2)`). But the demo's custom flags (`--only`, `--debug-overlay`, `--yaw`, `--zoom`, `--subdivisions`, `--no-auto-rotate`, `--no-spin`, `--no-lighting`, `--sweep-*`, `--depth-probe*`, …) are still parsed by the demo's own hand-rolled `parseArgs(argc, argv)` (called just before init) and were **never registered** on `IREngine::args()`. So the engine parse rejects them before the demo ever runs.

This blocks every render-validation task that drives `canvas_stress` with a debug flag (shadow-overlay captures, `--only` isolation, sweeps, depth-probe). It likely affects the other demos migrated to `IREngine::init(argc, argv)` that retained a hand-rolled flag parser.

## Repro
`IRCanvasStress --only gridspin` (or any custom flag) → `Unknown argument` + `exit(2)`. A flagless `--auto-screenshot` run still works.

## Suggested approach
Register the demo's custom flags on `IREngine::args()` (`.flag` / `.string` / `.integer`) **before** `IREngine::init(argc, argv)`, then read them back via `IREngine::args().getFlag(...)` / `getString(...)` and retire the hand-rolled `parseArgs` — the canonical IRArgs adoption (`creations/demos/fog_demo/main.cpp`). Audit the other migrated demos for the same hand-rolled-parser-plus-strict-init gap; this is squarely in the IRArgs migration epic (#2057) scope.

## Area
creations/demos/ (canvas_stress + audit migrated siblings)

## Model
sonnet

## Blocked by
(none)

jakildev:
[...]

PR #2076, opened 2026-06-27
creations: migrate auto-screenshot-only + no-arg demos to IRArgs (P2)

jakildev:
## Summary
- Migrates 21 auto-screenshot-only demos + 3 no-arg demos from the old `IRVideo::parseAutoScreenshotArgv + IREngine::init(argv[0])` pattern to the unified `IREngine::init(argc, argv)` + `IREngine::args().autoScreenshotWarmupFrames()` API landed by P1 (#2058, PR #2069)
- Handles both `init(argv[0])` and `init(argv[0], "config.lua")` variants, and all variable name forms (`g_autoWarmupFrames`, `autoWarmupFrames`, namespace-qualified)
- P3 (perf_grid/shape_debug, PR #2072), P4 (lighting), and P5 (game/editors) intentionally not touched

## Test plan
- [x] All 24 targets build clean (macOS/Metal)
- [x] `day_cycle --auto-screenshot 12` and `sprite_demo --auto-screenshot 12` capture screenshots (exit 0)
- [x] `day_cycle --help` exits 0 with correct arg listing
- [x] No-arg demos (`default`, `z_yaw_rotation --mouse`) start + run normally

Closes #2059

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1971, opened 2026-06-23
test: refresh stale canvas_stress macos-debug render-verify references

jakildev:
## Summary
- Re-blessed all 6 gated `canvas_stress` **macos-debug** render-verify references against current `origin/master` (post-#1953). They were last blessed at #1719/#1723 (Jun 11, `97e74307`) and went 6/6 red: ~30 render PRs merged since (AO de-speckle #1873, world-placed detached-canvas depth #1872, sun-shadow ambient #1874, per-axis scatter #1877/#1878/#1882/#1883, camera Z-yaw pivot #1926/#1927) plus the #1942 viewport-center jitter regression that #1953 just reverted.
- **Reference-only refresh** — no source, `manifest.json`, `STRUCTURAL-BASELINES.md`, or `metrics.json` changes. `shadow_overlay_floor` (idx 8) / `revox_coverage` (idx 9) capture indices unchanged.
- Each shot **visually verified benign before blessing** (mandatory tripwire): central re-voxelize cubes lit, peripheral forward-scatter ring intact, deltas localize to lit surfaces / scatter detail / small per-entity position shifts that map to the merged render PRs above. No missing geometry, black frames, or unexplained deltas.

## Cross-host note (linux-debug)
The **linux-debug** references are stale by the same drift, but the cross-host smoke protocol (`fleet:needs-<host>-smoke`) builds + runs `IRShapeDebug` only and does **not** refresh demo references — so it cannot bless the linux-debug `canvas_stress` set. Per the historical #1577 precedent, the linux-debug refresh is a **separate Linux-host task**, filed as #1969. This PR therefore carries no smoke label (a reference-only diff touches no render-path code).

## Test plan
- [x] `python3 scripts/render-verify.py --target IRCanvasStress` FAILs 6/6 on clean `origin/master` (pre-bless, max_delta ~245).
- [x] `--update-references --force` then re-run without it → **6/6 PASS** (revoxelize_solids 99.9999% / max_delta 13, the documented round-to-cell speckle, within thresholds).
- [ ] linux-debug set refreshed on a Linux host (#1969).

Follow-ups filed: #1969 (linux-debug refresh), #1970 (render-verify compare-shot gating, from the plan gotcha).

Closes #1944

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2116, opened 2026-06-29
render-verify: extra_runs second capture pass (canvas_stress compare)

### LINKED ISSUE #1970: render-verify: support a per-target second run-config to gate canvas_stress --only compare shots
From #1944's plan gotchas. `canvas_stress` defines compare-shots
`compare_yaw0` / `compare_yaw_q` that only spawn under `--only compare`
(`kGroupCompare`, opt-in), which **suppresses** the main scene groups
(so3_*, revoxelize_solids). So the compare shots cannot be captured in the
same render-verify run as the 6 gated shots, and the manifest has no
`run_args` field (render-verify passes demo args only ad-hoc via its CLI —
`scripts/render-verify.py:450-452`). Persisting a `--only compare` gate
therefore needs a render-verify harness extension: a per-target **second
run-config** (its own demo args + its own reference subset) so a target can
declare more than one capture pass.

**Area:** test / scripts (render-verify harness)
**Model:** opus
**Blocked by:** #2103

## Acceptance criteria
- A `canvas_stress` render-verify run gates `compare_yaw0` / `compare_yaw_q`
  (captured via `--only compare`) in addition to the existing 6 main-scene
  shots, without the two passes suppressing each other.
- `shadow_overlay_floor` (idx 8) / `revox_coverage` (idx 9) capture indices
  unchanged (compare shots are conditional end-appends, so no index shift).

## Context
May fold into the #1954 jitter-harness extension work. The NOTE's "insert
contiguously after revoxelize_solids → index shift" concern is moot: the
compare shots are conditional end-appends, absent from the default capture.

jakildev:
## Summary
- Adds an optional **`extra_runs`** block to render-verify manifests: each entry is a *second capture pass* with its own demo args and gated reference subset, so a manifest can gate shots the default scene suppresses.
- Wires canvas_stress `compare_yaw0` / `compare_yaw_q` (#1374) — they only spawn under `--only compare`, which hides the default groups, so they can't share the main capture.
- The pass re-runs the demo, slices its gated shots out of the capture by `capture_offset` (negative = from the tail, robust to leading default-shot growth), and pixel-diffs against the same backend ref dir.
- **Cross-host-safe rollout:** references are blessed per-host, so a backend without them committed *skips* the pass (non-fatal) instead of failing; `--update-references` runs every pass so a fresh host bootstraps all of them.
- **Default single-pass behavior unchanged** (no `extra_runs` → original code path; diffs stay in `screenshots/diffs` unless `extra_runs` forces a sibling dir that survives each pass's rmtree).

## Blocked for the final step — #2103
`--only compare` is **dead on master**: the IRArgs P2 migration (#2076) swapped canvas_stress to `IREngine::init(argc, argv)` (strict IRArgs validation) without registering its ~17 custom flags, so `init()` exits(2) "Unknown argument: --only" before the demo's own parser is used. Tracked as **#2103** (full canvas_stress IRArgs migration is epic #2057 work, not render-verify).

Consequently the macOS/Linux `compare_*` references **cannot be captured yet**, and the compare pass **skips cleanly** (never invoking the broken flag) until #2103 lands. This PR delivers the harness + wiring; the per-host ref-bless that actually engages the gate is the remaining step, gated on #2103 — so this does **not** `Closes #1970`.

## Pre-existing, out of scope
canvas_stress's macOS main-6 references are stale on master (last blessed #1971, Jun 23; #2088/#2090/#2089/#2068/#2043 drifted output since) → `render-verify --target IRCanvasStress` already fails 6/6 on master, independent of this PR (this diff touches no render code). That's the recurring #1944/#1971 refresh need (#1969 = linux half in flight); **not** re-blessed here.

## Test plan
- [x] `python3 scripts/tests/test_render_verify.py` — 30/30 PASS (new: `_parse_extra_runs` validation, `_slice_capture` positive/negative/out-of-range, `missing_ref_is_skip`).
- [x] `python3 scripts/render-verify.py --target IRShapeDebug` — 24/24 PASS (single-pass path unregressed by the refactor).
- [x] canvas_stress gate run shows the compare pass `SKIP` (non-fatal, "references pending") and the main pass evaluating normally.
- [ ] Once #2103 restores `--only`: bless `compare_yaw0`/`compare_yaw_q` refs (macOS + Linux) and confirm the compare pass gates green.

Refs #1970. Blocked by #2103.


jakildev (commented):
## Review — render-verify: extra_runs second capture pass (canvas_stress compare)

**Verdict:** approve

### Praise

- **The `capture_offset` design is exactly right.** Using a negative index from the tail ("3rd from last") decouples the slice mapping from the leading default-shot count — the offset only couples to the appended block's own length, so default-shot growth never silently mismaps the extra refs. The notes section explains the invariant clearly.
- **Cross-host skip logic is well-chosen.** Skipping the pass entirely only when *no* references exist for the pass (rather than failing on the first missing ref), then using `missing_ref_is_skip=True` for per-shot skips once at least one reference is blessed, correctly models the cross-host reference handoff without false failures.
- **Single-pass path is truly untouched.** The `if extra_runs:` guards on `diff_dir` routing and the extra-pass loop ensure the original code path is byte-identical when `extra_runs` is absent.
- **`_run_capture` and `_write_references` extractions are clean.** The default pass and each extra pass share the same capture/write helpers — no duplicated logic — and both paths flow through `_collect_shots` (count-asserting) or `_slice_capture` (range-asserting) so captures are always validated before reference writes.
- **`not crashes` correctly extends the PASS guard to extra runs.** The old `run_crash is None` check is now `not crashes`, and each pass appends to the same list — a crashed extra run correctly blocks the PASS verdict.
- **Test coverage is thorough:** `_parse_extra_runs` validation (8 error cases + dup names + non-list), `_slice_capture` (positive, negative, out-of-range), and `evaluate_shots` with `missing_ref_is_skip` (all-missing → skip, mixed → gates present refs). 30 unit tests per PR body.

### Nits

- `scripts/render-verify.py:430-441` — In the `--update-references` path, the return value of `_run_capture` for each extra run is discarded. A crash is already printed to stderr by `_run_capture`, and the subsequent `_slice_capture` call will raise `SystemExit("out of range")` if the demo emitted too few screenshots — so it doesn't silently succeed. But the out-of-range message ("capture_offset X + Y shots is out of range for Z captured screenshots") is confusing when the real cause is a crash. Saving the result and printing a clearer note before re-raising (or early-returning 1) would be a small readability win.

- [x] `python3 scripts/tests/test_render_verify.py` — 30/30 PASS (per PR body)
- [x] `python3 scripts/render-verify.py --target IRShapeDebug` — 24/24 PASS, single-pass path unregressed (per PR body)
- [x] canvas_stress compare pass shows SKIP (non-fatal, references pending) with the main pass evaluating normally (per PR body)
- [ ] Once #2103 restores `--only compare`: run `--update-references` on macOS then Linux to bless `compare_yaw0`/`compare_yaw_q` references and confirm the compare pass gates green on both hosts

Opus recheck not required.

jakildev:
Addressed nit: saved the `_run_capture` return value in the `--update-references` extra-run loop and added an early-return 1 with a clear crash attribution message before the confusing `_slice_capture` out-of-range path. All 30 unit tests pass. — worker

PR #2121, opened 2026-06-29
render: fix skeletal seedVoxelBoneSlots OOB span deref (#2032)

### LINKED ISSUE #2032: render/skeletal: seedVoxelBoneSlots out-of-bounds std::span deref — segfault on first frame (IRSkeletalDemo)

jakildev:
## Summary
- **#2032 root fix.** `seedVoxelBoneSlots` / `resetVoxelBoneSlotsToRigid` (`system_update_joint_matrices.hpp`) read the captured `C_VoxelSetNew::voxels_` span, which a between-frame canvas archetype migration deep-copies and frees while `voxelStartIdx_` / `numVoxels_` stay valid — the dangling-read hazard `C_VoxelSetNew::updateAsChild` already documents and works around. The stale `voxels_[i]` deref is the first-frame `std::span::operator[]` segfault (deterministic on gcc 15.2 / native Windows; reads adjacent surviving memory on laxer toolchains). Both now read bone ids from the **live `C_VoxelPool`** by `voxelStartIdx_ + i` over the stable `numVoxels_` scalar, **clamped to the pool's writable tail** — provably in-bounds (the bounds invariant the acceptance asks for, not just an assert that fires after the deref).
- **Second blocker on the same `--auto-screenshot` path.** The skeletal demo's `kShots[]` table was a **function-local `constexpr`** array assigned to a config the auto-screenshot system stores for the process lifetime. It died at scope exit → `shots_` dangling → null `shot.label_` → `fmt::format_error: string pointer is null` on the first capture frame. Once the segfault is fixed the demo reaches this next, so #2032's acceptance ("renders + saves screenshots + exits 0") needs **both**. Made the table `static` (matches every other demo's shot table).

## Verification (macOS / Metal)
- `IRSkeletalDemo --auto-screenshot 5` → all 4 shots cycle (labels print correctly, not null), 4 PNGs saved, **exit 0**. The committed screenshots (`docs/pr-screenshots/<branch>/`) show the rigs render correctly — the 30-joint snake is skinned/curved, lamp/cross/desk posed — so the live-pool read is **not** a skinning regression.
- The #2032 segfault is a strict-toolchain crash; macOS reads adjacent surviving memory, so the actual *segfault-gone* wants **Linux / native-Windows smoke**. The live-pool read is in-bounds by construction. Note: master **crashes** on this demo (the `kShots` bug), so a before/after screenshot diff isn't possible — the committed shots are the after-state proof.

## Test plan
- [x] `fleet-build --target IRSkeletalDemo` clean (macOS/Metal)
- [x] `IRSkeletalDemo --auto-screenshot 5` → exit 0, 4 screenshots, rigs render correctly (snake skinned)
- [ ] Native Windows (gcc 15.2) / Linux: confirm the original first-frame `seedVoxelBoneSlots` segfault is gone and the demo exits 0

Closes #2032


jakildev (commented):
## Review — render: fix skeletal seedVoxelBoneSlots OOB span deref (#2032)

**Verdict:** approve

### Nits

- `engine/prefabs/irreden/render/systems/system_update_joint_matrices.hpp:168-173` — `liveWritableVoxelCount` calls `pool.getColors().size()` for the bounds check, then the caller calls `pool.getColors()` again to get `poolVoxels`. Two round-trips to the same accessor for the same call frame. Purely cosmetic in a non-hot `beginTick` path; could return the pair `{poolVoxels ref, count}` or take the already-fetched vector as a param to save the extra call.

### Praise

- **Root cause is correctly identified and fixed.** `C_VoxelSetNew::voxels_` is a `std::span` whose backing is deep-copied and freed on canvas archetype migration while `voxelStartIdx_` / `numVoxels_` remain valid scalars. Reading `voxelSet.voxels_[i]` in `seedVoxelBoneSlots` after a migration is UB (dangling span); switching to the live pool via `voxelSet.voxelStartIdx_ + i` into `pool.getColors()` is provably in-bounds by construction.
- **`liveWritableVoxelCount` as a named helper** makes both `seedVoxelBoneSlots` and `resetVoxelBoneSlotsToRigid` symmetric and the bounds invariant auditable in one place. The clamping logic is clean and covers both the shrunk-pool case and the `startIdx >= poolSize` case.
- **`static constexpr kShots[]`** is the exact right fix for the second crash: a `constexpr` function-local array storing a `shots_` pointer that outlives the scope. Matches the established pattern across other demos.
- Comments explain the WHY (canvas archetype migration deep-copies C_VoxelPool + frees the span's backing), not just the what. Mirrors the already-present `updateAsChild` documentation on `C_VoxelSetNew`.
- `IRMath::min` used correctly throughout.

- [x] `IRSkeletalDemo --auto-screenshot 5` → exit 0, 4 screenshots, snake skinned (macOS/Metal)
- [ ] Linux or Windows (gcc 15.2): confirm first-frame `seedVoxelBoneSlots` segfault is gone and `IRSkeletalDemo --auto-screenshot 5` exits 0 (Metal reads adjacent surviving memory; the strict-toolchain crash only manifests on OpenGL/Windows/Linux)

Opus recheck not required.

jakildev (commented):
## Opus recheck — render: fix skeletal seedVoxelBoneSlots OOB span deref (#2032)

Sonnet flagged this as approve / "Opus recheck not required," but it touches a render-system lifetime + ECS-pool-indexing invariant, so I ran a deep pass on the parts a first-pass checklist can't fully prove. All confirmed:

- **Span→live-pool index mapping is correct.** `voxelSet.voxels_` was a view based at `voxelStartIdx_`, so the old `voxels_[i]` is exactly `pool.getColors()[voxelStartIdx_ + i]`. The rewrite preserves the mapping; it doesn't shift the range.
- **Clamp is provably in-bounds.** `writableTail = poolSize > startIdx ? poolSize - startIdx : 0u` guards the unsigned underflow, and `min(numVoxels_, writableTail)` keeps `[startIdx, startIdx+count)` strictly inside `[0, poolSize)` even after a migration shrank the pool. In the happy path `count == numVoxels_ == old voxels_.size()`, so behavior is byte-identical — the fix only diverges in the degenerate shrink case the old code would have UB-read.
- **Multi-set pool isolation.** Clamping to `numVoxels_` (not `poolSize - startIdx`) means that if a pool holds sibling sets after this one's range, the seed still only touches this set's voxels — strictly more correct than the old `voxels_.size()`.
- **No tick-path `getComponent` regression.** Each function still resolves `C_VoxelPool` exactly once (now hoisted and reused via the `pool` ref), in the rare re-rig path, not per-frame.
- **Reference lifetimes safe.** `poolVoxels` is consumed only in the seed loop before `setTransformIndicesForRange`; no structural ECS mutation occurs between the fetch and the use, so the `pool` ref and the colors vector stay valid.
- **`static constexpr kShots[]`** is the right lifetime fix: `createAutoScreenshotSystem` retains the `shots_` pointer in a process-lifetime `CyclingState`, so the table must have static storage duration.

The Sonnet nit (double `getColors()` round-trip in `liveWritableVoxelCount` + caller) is cosmetic in a non-hot `beginTick` path — agree it's a nit, not a blocker.

Keeping `fleet:approved`. The unchecked Linux/Windows smoke item in Sonnet's test plan is correctly tracked by `fleet:needs-windows-smoke`.

🤖 Opus recheck (review-pr skill) — Claude Opus 4.8

jakildev:
**has-nits addressed** (worker-2)

Refactored `liveWritableVoxelCount` to accept `std::size_t poolSize` instead of `const C_VoxelPool&`, eliminating the redundant `pool.getColors()` accessor round-trip:

- `seedVoxelBoneSlots`: hoisted `const std::vector<C_Voxel>& poolVoxels = pool.getColors()` before the count call; passes `poolVoxels.size()` to the helper — single fetch used for both the bounds check and the loop.
- `resetVoxelBoneSlotsToRigid`: passes `pool.getColors().size()` inline (no poolVoxels needed here, single fetch for the count only).
- Updated the helper's comment to clarify that `poolSize` must come from `pool.getColors().size()` and why (the #2032 dangling-span hazard still applies; callers must use the live pool, not the stale span).

Build: IRSkeletalDemo — clean.

Issue #1998, opened 2026-06-25
fleet: no GL-vs-Metal host-capability gate — GL-only render tasks churn macOS/Metal opus panes


jakildev:
**Model:** opus

The single biggest current source of opus no-op iterations. `fleet-claim` / `fleet-state-scout` have **no host-capability awareness**: a GL-backend-only task projects as `owner:free` to every opus pane, so macOS/Metal panes (GL 4.1 < required 4.5 — can't build/run/verify the GL backend) keep getting dispatched onto it and can only claim→comment→release.

**Live example:** #1937 (epic #1933 C1, "analytic edge-aware coverage — GL backend") has been claim/release churned **5+ times** by mac panes (worker-1/3/4; releases 06-22 07:37/08:06/08:50Z, 06-25 01:53Z, +more). Its dependents #1938 (Metal port C2 — which a mac pane *could* do) and #1939 (C3) are blocked behind it, so the whole #1933 scatter-coverage chain stalls while mac opus panes burn ticks. Every mac opus dispatch where #1937 is the lone free task is a *structurally guaranteed* no-op.

> NB: workers keep citing "#1726" as the tracker, but **#1726 is CLOSED and was a different scope** (design-blocked-PR-backed projection). The GL-vs-Metal host gate was never actually filed. This is it.

## Proposed fix
1. A host-capability label on backend-specific tasks — **`fleet:needs-gl-host`** (and/or `host:linux` for Linux-only smoke like game #202). New label in `scripts/fleet/fleet-labels`.
2. `fleet-state-scout`: a `fleet:needs-gl-host` task projects as not-pickable for Metal-only hosts, and its blocked dependents (#1938/#1939) are not surfaced as "perpetually blocked behind an unpickable parent."
3. `fleet-claim`: refuse/route a `fleet:needs-gl-host` claim from a Metal host so GL-only work flows to Linux/Windows panes preferentially.

## Acceptance
- A `fleet:needs-gl-host` task is not dispatched to / not claimable from a macOS/Metal pane.
- #1937-class tasks stop round-tripping on mac panes; picked by Linux/Windows opus panes instead.

_Filed from review-fleet-feedback triage 2026-06-24. Source: worker-1/3/4 host-mismatch entries 06-22 → 06-25 (5+ #1937 releases)._

jakildev:
[...]

PR #2134, opened 2026-06-30
fleet: scout infers needs_gl_host from an explicit body host-requirement

jakildev:
## Summary

Hardens against the opus-starvation incident we just hit on the Mac fleet.

**#1969** ("refresh linux-debug render-verify references") declares in its body *"This must run on a Linux host"* but was missing `fleet:needs-gl-host`. So the label-based host gate saw it as claimable on Mac → the dispatcher kept electing it (oldest claimable → sonnet) → a worker claimed, couldn't build Linux refs on Metal, released → churn, and the queued **opus** tasks starved for ~13 min. (The filer labeled the macos sibling #1944 to run on Mac but forgot the gate here.)

### Fix

`fetch_task_queue` now also infers `needs_gl_host` from an explicit host-requirement declaration in the body — a backstop for a forgotten label:

```
needs_gl_host = "fleet:needs-gl-host" in labels OR _body_requires_gl_host(body)
```

**Precision-first by design.** A false positive would wrongly skip a *mac-runnable* task (strands it) — worse than a miss (a miss just churns as before) — so the regex matches only a strong requirement phrasing (`must run on` / `must be run on` / `only runs/builds on` **a linux/windows/opengl host**), never:
- a passing mention (`tested on a Linux host`, `the linux-debug references`)
- a **macOS** requirement (the OS alternation is linux/windows/opengl, so a mac-only task never self-gates off its own host)

The explicit label stays the primary, recommended signal; this is the safety net.

## Test plan

- [x] Real #1969 body → `_body_requires_gl_host` returns `True`
- [x] `NeedsGlHostAnnotation` gains: 4 requirement phrasings (inferred True) + 4 negatives (passing mention / macOS / `linux-debug` path → False)
- [x] Full `test_worker_projection.py` green; full fleet Python suite green except the pre-existing `test_fleet_validate_roles` drift

## Scope note

This stops a host-mismatched task from *entering* the election (the surgical fix for the recurring host-mismatch category: #1937, #1998, #1969). The deeper, more general guard — having the dispatcher serve another claimable class when the elected class is fully cap-covered, so no single slow/churning head can defer other classes — is a larger dispatcher change I left as a separate follow-up.

### DIFF
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -401,6 +401,35 @@ _EPIC_CHECKLIST_RE = re.compile(
     r'^[ \t]*-[ \t]*\[( |x|X)\][ \t].*?#(\d+)', re.MULTILINE)
 
 
+# #1969: a task whose body explicitly declares it must run on a Linux / Windows
+# / OpenGL host, but is missing the fleet:needs-gl-host label, is invisible to
+# the label-based host gate (fetch_task_queue below). On a Metal-only (mac) host
+# the dispatcher then keeps electing it (oldest claimable) and a worker
+# claims -> can't build -> releases, churning the lane and starving other
+# classes (the #1969 incident: a "refresh linux-debug render-verify references"
+# task that can only run on Linux). Infer the gate from the body as a backstop so
+# the projection treats such a task as host-incompatible even when the label was
+# forgotten at filing. Deliberately PRECISION-first — only a strong requirement
+# phrasing ("must run on" / "must be run on" / "only runs/builds on" a
+# linux/windows/opengl host) matches, never a passing mention ("tested on a
+# Linux host"). A false positive would wrongly skip a mac-runnable task, which is
+# worse than a miss (a miss just churns like before), so recall is sacrificed for
+# precision; the explicit label remains the primary, recommended signal. macOS
+# host declarations don't match (the OS alternation is linux/windows/opengl), so
+# a mac-only task is never self-gated off its own host.
+_GL_HOST_REQUIRED_RE = re.compile(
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: scout infers needs_gl_host from an explicit body host-requirement

**Verdict:** approve

### Praise

- **Correct precision-over-recall tradeoff.** A false positive (stranding a mac-runnable task) is worse than a miss (same churn as before), so the regex requires a strong phrasing (`must run on` / `must be run on` / `only runs/builds on`) and explicitly excludes macOS from the OS alternation. The design rationale is documented in the comment and PR body.
- **Regex verified correct against all eight test cases.** `linux[\w/+-]*` cleanly handles `Linux/Windows` and `linux-debug`; the trailing `\s+host\b` prevents matching `opengl` in a path context; `re.IGNORECASE` covers mixed-case body text. The `_body_requires_gl_host` None-guard (`body or ""`) prevents a crash if the issue body is missing.
- **Integration point is correct.** Inference is applied inside `fetch_task_queue` where `needs_gl_host` is first set, so the scout state JSON that the dispatcher reads inherits the inferred flag without any downstream change.
- **Test coverage is thorough.** The `_issue_body` helper keeps tests readable. The four positive cases (literal incident phrasing, windows, opengl, multi-OS slash) and four negative cases (passing mention, Linux path reference, macOS requirement, bare `linux-debug` reference) cover the stated design boundaries cleanly.

- [x] `python3 -m unittest discover -s scripts/fleet/tests` — `test_worker_projection` green (confirmed in PR body)
- [x] Full fleet Python suite green (modulo pre-existing `test_fleet_validate_roles` drift, confirmed)

Opus recheck not required.

PR #2117, opened 2026-06-29
demo(shape_debug): recomputeFaceOccupancy before syncActiveMask in createVoxelPoolShape

jakildev:
## Summary

- **Root cause**: `createVoxelPoolShape` calls `syncActiveMask()` after SDF carving to fix the per-slot active mask, but never called `IRPrefab::Voxel::recomputeFaceOccupancy()` to fix the exposed-face mask. Surface voxels on carved shapes had all faces wrongly marked as occluded by (now-inactive) neighbors, so the compact shader emitted nothing → black render.
- **Fix**: Add `recomputeFaceOccupancy(vs.voxels_, size)` before `vs.syncActiveMask()`. Mirrors `createCustomVoxelFigure`, which already had the correct two-call sequence.
- **References**: 21 macos-debug PNGs updated to show correct rendering. render-verify passes 24/24.

Closes #2018

## Screenshots

<details>
<summary>zoom1_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom8_origin</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom16_lod_fine_sphere</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>lod_swap_zoom4_cone</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>lod_swap_zoom16_sphere</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_yaw90</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_pan16_yaw0_pivot</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

*(Full set of 21 before/after pairs in `docs/pr-screenshots/claude/2018-createvoxelpoolshape-face-occupancy/`)*

## Follow-up (not in scope)

Sibling demos with the same latent bug (missing `recomputeFaceOccupancy` before `syncActiveMask`):
- **[high]** `creations/demos/lighting/common/lighting_demo_scene.hpp:152` — `createVoxelPoolShape`
- **[medium]** `creations/demos/z_yaw_rotation/common.hpp:7` — `carveSphere`
- **[medium]** `creations/demos/ir_voxel_yaw/main.cpp:317` — `createHollowShell`

[...]

PR #2147, opened 2026-06-30
demo(shape_debug): createVoxelPoolShape uses IRMath::SDF::evaluateGrid (#2120)

### LINKED ISSUE #2120: demo(shape_debug): createVoxelPoolShape should use IRMath::SDF::evaluateGrid instead of hand-rolled per-cell loop
## Problem

`createVoxelPoolShape` in `creations/demos/shape_debug/main.cpp` (~line 881) hand-rolls a per-cell `IRMath::SDF::evaluate` loop over the voxel volume. The batch helper `IRMath::SDF::evaluateGrid` already exists and handles this pattern.

Found by `simplify-scan-render-leak` during PR #2117 review. Pre-existing before #2117 — not introduced by that PR.

## Fix

Replace the hand-rolled loop with a call to `IRMath::SDF::evaluateGrid`. Also check sibling demos that copy-paste the same pattern (lighting demo, z_yaw_rotation, ir_voxel_yaw).

jakildev:
## Summary
- `createVoxelPoolShape` (shape_debug) hand-rolled a per-cell `IRMath::SDF::evaluate` loop over the voxel volume. Replaced it with the existing batched `IRMath::SDF::evaluateGrid` helper + a single threshold pass.
- **Byte-identical:** `evaluateGrid` samples each cell at `vec3(x,y,z) - size*0.5 + 0.5` in `index3DtoIndex1D` order — bit-for-bit equal to the centered positions `C_VoxelSetNew` lays out for `centerAroundOrigin` (both reduce to exact multiples of 0.5), so every carve decision is preserved.

## Test plan
- [x] `render-verify` (macos-debug / Metal): **all 21 shape_debug shots PASS** — 19 byte-identical (max_delta 0, PSNR inf); the only 2 with any delta are the non-cardinal-yaw shots (max_delta 22 / 7, threshold 64), the demo's known run-to-run scatter jitter, orthogonal to entity-creation-time SDF carving.

Found by `simplify-scan-render-leak` during PR #2117 review; pre-existing before that PR.

Closes #2120

### DIFF
--- a/creations/demos/shape_debug/main.cpp
+++ b/creations/demos/shape_debug/main.cpp
@@ -877,11 +877,17 @@ EntityId createVoxelPoolShape(
     auto sdfType = static_cast<IRMath::SDF::ShapeType>(type);
     vec4 sdfParams = IRMath::SDF::effectiveParams(sdfType, shapeParams);
 
+    // Batch-evaluate the SDF over the whole grid via the shared helper rather
+    // than calling evaluate() per cell. evaluateGrid samples each cell at
+    // vec3(x,y,z) - size*0.5 + 0.5 in index3DtoIndex1D order — bit-identical to
+    // the centered positions C_VoxelSetNew laid out for centerAroundOrigin — so
+    // distances[i] lines up one-for-one with voxel i.
+    std::vector<float> distances(static_cast<std::size_t>(vs.numVoxels_));
+    IRMath::SDF::evaluateGrid(vs.size_, sdfType, sdfParams, distances);
+
     int activeCount = 0;
     for (int i = 0; i < vs.numVoxels_; ++i) {
-        vec3 localPos = vs.positions_[i].pos_;
-        float sdf = IRMath::SDF::evaluate(localPos, sdfType, sdfParams);
-        if (sdf > IRMath::SDF::kSurfaceThreshold) {
+        if (distances[i] > IRMath::SDF::kSurfaceThreshold) {
             vs.voxels_[i].deactivate();
         } else {
             ++activeCount;


jakildev (commented):
## Review — demo(shape_debug): createVoxelPoolShape uses IRMath::SDF::evaluateGrid (#2120)

**Verdict:** approve

### Summary

Replaces `createVoxelPoolShape`'s hand-rolled per-cell `IRMath::SDF::evaluate` loop with a single batched `IRMath::SDF::evaluateGrid` call into a `distances` buffer, then a threshold pass. Mechanical perf/reuse cleanup found during PR #2117 review (`simplify-scan-render-leak`); pre-existing before that PR, not introduced by it.

### Verification performed

- Confirmed the byte-identical claim algebraically. `evaluateGrid` samples each cell at `vec3(x,y,z) - size*0.5 + 0.5`; `C_VoxelSetNew`'s `centerAroundOrigin` ctor (`component_voxel_set.hpp:151-157`) offsets by `-(size-1)*0.5`. Since `size*0.5 - 0.5 == (size-1)*0.5` exactly (both reduce to a multiple of 0.5, exactly representable in float for the demo's size range), `sdfPos == pos` bit-for-bit — the claim holds, not just "close enough."
- Confirmed the loop-order claim. `evaluateGrid` iterates `z` outer, `y`, `x` inner, incrementing `flat` sequentially — `flat = z*size.y*size.x + y*size.x + x` for the n-th cell, which is exactly `IRMath::index3DtoIndex1D` (`ir_math.hpp:1504`: `index.z*size.x*size.y + index.y*size.x + index.x`). `positions_`/`voxels_` are laid out via the same `index3DtoIndex1D` call in the ctor, so `distances[i]` lines up with `vs.positions_[i]`/`vs.voxels_[i]` one-for-one, matching the old per-cell loop that read `vs.positions_[i].pos_` directly.
- Checked the allocation-failure edge case (`C_VoxelSetNew` ctor's span-mismatch branch zeroes `size_`/`numVoxels_` on an out-of-voxels condition). The new code reads `vs.size_`/`vs.numVoxels_` (not the local `size` parameter) for both the `distances` buffer size and the `evaluateGrid` call, so this degrades to a correct no-op exactly like the old `for (i < vs.numVoxels_)` loop did — no behavior change in the failure path.
- Confirmed the post-carve `recomputeFaceOccupancy` + `syncActiveMask` calls (required per `voxel/CLAUDE.md` whenever a carve bypasses `C_VoxelSetNew` mutators) are untouched by this diff and still present immediately after the loop.
- `std::vector<float> distances` allocation happens once per shape-entity creation (not a per-frame/tick path), so this isn't a hot-path allocation concern.
- No render/, system/, world/, ECS, GPU-buffer, or concurrency surface touched — this is a CPU-only refactor inside `creations/demos/`. Doesn't meet the cross-host smoke-tag criteria (no `engine/render/`, shader, `engine/system/`, or CMake changes).
- Test plan: author's `fleet-build` + `render-verify` (macOS/Metal, 21/21 shape_debug shots) is sufficient evidence given the verified-exact math above; the 2 shots with nonzero delta are explicitly the known non-cardinal-yaw scatter jitter (unrelated demo behavior), not attributable to this change.

### Praise

- The inline comment correctly states the non-obvious *why* (the exact offset-formula equivalence and ordering match) rather than restating what the diff does — exactly the kind of comment that earns its keep.

- [x] Already covered by the author's test plan (build + render-verify, 21/21 shots) — re-verified the underlying math claim independently above.

Opus recheck not required.

Issue #2107, opened 2026-06-29
test: FogCrossSection GPU GTEST harness (Tests A-E) for the #2102 per-fragment fog clip


jakildev:
**Area:** test/render
**Suggested model:** [opus]
**Follow-on to:** #2102 (PR adds the fix + the `fog_demo --edge-zoom` regression canary)

## Context
#2102 moved the fog analytic-circle reveal of voxel objects into `VOXEL_TO_TRIXEL_STAGE_1` (drop fog-hidden columns so objects clip on the disc arc instead of `FOG_TO_TRIXEL` hard-blacking their faces). The plan called for a formal GPU GTEST harness (Tests A-E). That harness follows the `test/render/gpu_compute_dispatch_test.cpp` pattern, which is `#if defined(IR_GRAPHICS_OPENGL)`-gated — it stands up a hidden GL 4.5 context, dispatches the real compute shader, and reads back the SSBO. **It is GL-host only and cannot be authored/validated on a macOS/Metal host**, so #2102 (authored on macOS) shipped with the `--edge-zoom` demo canary as the interim regression vehicle and deferred the GTEST here.

## Task
On a GL host (Linux/Windows), add a `FogCrossSection` headless-GPU test (same hidden-context fixture as `gpu_compute_dispatch_test.cpp`, `GTEST_SKIP` on no-GPU) covering the plan's Tests A-E against the stage-1 clip:
- **A — no boundary black:** a voxel pillar + cube straddling an analytic circle; read back trixelColors; assert zero pure-black pixels inside/around the objects at the boundary.
- **B — smooth partial coverage (no pop):** sweep the circle across an object in small steps; assert opaque coverage is continuous (max per-step delta below a threshold), not a 0->full jump.
- **C — no top-face gaps:** a voxel floor straddling the circle has no interior holes on the revealed side.
- **D — floor/object edge agreement:** the object reveal edge and the floor reveal edge coincide within +/-1px along the arc (they share `fogVisionCircleReveal`).
- **E — GL/Metal parity:** cardinal poses byte-identical; arbitrary poses +/-1 (mirror `detached_world_depth_test.cpp`).

Optionally add committed `render-verify` reference images for `fog_demo --edge-zoom` (host-specific) so the canary becomes an automated pass/fail gate alongside the GTEST.

## Acceptance
- `FogCrossSection` tests A-E pass on a GL host (A-D fail on master pre-#2102 = repro).
[...]

PR #2150, opened 2026-06-30
fleet: scout GL-host inference also matches the "GL-host only" adjective form

jakildev:
## Summary

#2134's body-inference of `needs_gl_host` was precision-first but missed a common phrasing. #2107 said *"It is **GL-host only** and cannot be authored/validated on a macOS/Metal host"* — the adjective form, not the verb form the regex matched — so it was misrouted to a Metal pane and churned (a worker had to hand-apply the label and filed scout-regex feedback).

Adds a second, equally precision-safe alternation: `(gl|opengl|linux|windows)[-\s]?host[-\s]?only`. Both shapes are deliberate declarations; a passing mention still doesn't match.

## Test plan
- [x] Real #2107 body now infers `needs_gl_host=True`.
- [x] `NeedsGlHostAnnotation`: +4 adjective-form positives (`GL-host only`, `GL host only`, `OpenGL-host-only`, `Windows-host only`); +2 negatives (`the host only…`, `macOS-host only`). Existing verb-form positives + `tested on a Linux host` / `must run on a macOS host` / `linux-debug references` negatives unchanged.

Part of the "stop GL-host tasks churning Metal panes" thread (re-arm exclusion is a sibling PR).

### DIFF
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -417,9 +417,18 @@ _EPIC_CHECKLIST_RE = re.compile(
 # precision; the explicit label remains the primary, recommended signal. macOS
 # host declarations don't match (the OS alternation is linux/windows/opengl), so
 # a mac-only task is never self-gated off its own host.
+#
+# Two requirement shapes are matched (both deliberate declarations, both
+# precision-safe): a verb form ("must run on a Linux host", "only builds on an
+# OpenGL host") and the very common "<GL|OpenGL|Linux|Windows>-host only" /
+# "GL host only" adjective form (#2107: "It is GL-host only and cannot be
+# authored/validated on a macOS/Metal host" — missed by the verb form alone).
 _GL_HOST_REQUIRED_RE = re.compile(
-    r"\b(?:must\s+(?:be\s+)?run\s+on|only\s+(?:runs?|builds?)\s+on)\s+"
-    r"(?:an?\s+|the\s+)?(?:linux|windows|opengl)[\w/+-]*\s+host\b",
+    r"\b(?:"
+    r"(?:must\s+(?:be\s+)?run\s+on|only\s+(?:runs?|builds?)\s+on)\s+"
+    r"(?:an?\s+|the\s+)?(?:linux|windows|opengl)[\w/+-]*\s+host"
+    r"|(?:gl|opengl|linux|windows)[\s-]?host[\s-]?only"
+    r")\b",
     re.IGNORECASE,
 )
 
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: scout GL-host inference also matches the "GL-host only" adjective form

**Verdict:** approve

### Summary

Extends `_GL_HOST_REQUIRED_RE` in `fleet-state-scout` with a second alternation matching the "&lt;GL|OpenGL|Linux|Windows&gt;-host only" adjective form (`#2107`: "It is GL-host only and cannot be authored/validated on a macOS/Metal host"), alongside the existing verb form ("must run on a Linux host"). Adds 4 new positive and 2 new negative cases to `test_needs_gl_host_inferred_from_body_declaration` / `test_needs_gl_host_not_inferred_from_passing_mention`.

### Verification performed

- Materialized the PR branch's `fleet-state-scout` + `tests/test_worker_projection.py` (plus the unchanged sibling `fleet_*.py` helper modules from master) into a scratch dir and ran the full `test_worker_projection` suite directly — all 51 tests pass, including the 4 new `NeedsGlHostAnnotation` cases.
- Traced the regex by hand against each new fixture:
  - `"GL-host only"` / `"GL host only"` / `"OpenGL-host-only"` / `"Windows-host only"` all match the new alternation `(?:gl|opengl|linux|windows)[\s-]?host[\s-]?only` — the `\b` anchor at the start of the outer group forces the alternation to only attempt a match at a genuine word boundary, so `"OpenGL-host"` correctly resolves via the `opengl` alternative (not a spurious `gl` match mid-word, since there's no `\b` between `Open` and `GL`).
  - `"macOS-host only"` (negative) correctly fails — `macOS` doesn't start with `gl|opengl|linux|windows`, and there's no `\b` inside `macOS-host` for the alternation to retry from.
  - `"the host only accepts one connection"` (negative) correctly fails — the alternation requires the OS token to immediately precede `host only`; the engine has no opportunity to match starting mid-`host`.
- Checked the other two consumers of `needs_gl_host` (`fleet_task_class.py` and its test) — they only read the boolean field the projection emits, they don't duplicate the regex, so this PR's scope is correctly self-contained to `fleet-state-scout` + its own test file.
- Confirmed the verb-form branch's behavior is structurally unchanged (same sub-pattern, just nested one level deeper inside the new outer alternation with the trailing `\b` now applied after either branch instead of duplicated per-branch) — the pre-existing verb-form positive/negative tests still pass unmodified.

### Praise

- The inline comment above the regex states the *why* (the #1969/#2107 incident chain, the precision-over-recall tradeoff) rather than just describing the new alternation — consistent with the durable-invariant comment bar.

- [x] Already covered — `test_worker_projection.py::NeedsGlHostAnnotation` (51/51 total suite tests pass, independently re-run against the PR branch).

Opus recheck not required.

Issue #1938, opened 2026-06-21
render: GL parity port of analytic scatter coverage (C2)


jakildev:
**Model:** opus
**Effort:** medium
**Part of epic:** #1933
**Plan file:** `~/.fleet/plans/issue-1933.md` (full epic plan)
**Blocked by:** #1937 — **MERGED** (PR #2013, 2026-06-25); the reference now exists.

> **Body corrected 2026-06-25** (un-inverted per the architect's parked decision):
> #1937 is the **Metal** lead (C1); this ticket is the **Metal → GL** parity port.
> Earlier the Scope/Approach read GL→Metal, which would have had an implementer
> re-edit the already-done `.metal` files. References were always correct.

## Scope

Port the edge-type-aware analytic scatter coverage landed for **Metal** in #1937
(C1, lead) to the **OpenGL backend**, keeping the two hand-mirrored scatter
shaders at feature parity. No new design — this is a faithful translation of
#1937's coverage model into GLSL.

## Approach

- Mirror #1937's changes from `engine/render/src/shaders/metal/peraxis_scatter.metal`
  (vertex + fragment functions) into the GLSL scatter shaders
  `v_peraxis_scatter.glsl` / `f_peraxis_scatter.glsl`, and the shared helper
  changes from `ir_iso_common.metal` into `ir_iso_common.glsl`.
- Carry across: the minimal visit-bound dilation, the true-edge-equation
  varyings, the interior-vs-boundary per-edge flags + occupied-neighbor tap, and
  the edge-type-aware analytic coverage / hard-threshold discard.
- Run the `backend-parity` skill audit so the analytic logic does not drift
  between Metal and GLSL (the shaders are hand-mirrored — `kScatter*` constants
  already live in both `ir_iso_common.{glsl,metal}`).

## Acceptance criteria

- **OpenGL** renders the coarse-cube near-cardinal poses with the same
  crisp-corner / no-dashing result #1937 proved on Metal.
- **Backend parity**: GL and Metal structurally identical where they should be
  (the `backend-parity` audit passes); cardinal fast path byte-identical on
  **GL** too.
- No GL-specific regression: GLSL compiles clean, no new artifacts on the
  cardinal fast path.

## Host requirement

- Acceptance gates on building + rendering the **OpenGL** backend, so this needs
  a **Linux/WSL or Windows (OpenGL) pane** — a macOS host builds Metal only and
  cannot compile-or-render the GLSL side (`backend-parity` can't build the
[...]

PR #2151, opened 2026-06-30
fleet: periodic worker re-arm delegates to the resolver (no more divergence)

jakildev:
## Summary

The "worker keeps spinning up and finding no work" churn traces partly to the dispatcher's **periodic safety re-arm** using its own inline `is_actionable` predicate that had **drifted** from the real dispatch decision. It excluded `inflight_pr` (the #2101 fix) but **not** `needs_gl_host` — so on a Metal pane it re-armed the worker trigger every interval on GL-host-only tasks (#1938/#1969/#2107) a fresh worker can never claim there, with a misleading *"projection has actionable free work"* log line.

### Fix: stop re-implementing claimability

Replace the inline predicate with `worker_rearm_should_fire()`, which calls `resolve_worker_class` and re-arms **iff the resolver would dispatch a concrete class** — genuinely claimable work under the *full* claimability gate (host capability, stackability, `inflight_pr`, owner, blocked). A `defer` verdict or the `''` lane-default fallthrough no longer re-arms. The safety re-arm is now in permanent lock-step with the dispatch decision, so this class of drift can't recur.

Side benefits:
- feedback PRs now correctly re-arm (the old predicate only scanned `tasks_open`);
- the re-arm log line names the class (`class=opus claimable …`) — better observability;
- ~55 lines of inline Python removed.

## Test plan
- [x] New `--rearm-check <role>` hook (test seam + operator tool, mirrors `--resolve-class`).
- [x] `test_dispatcher_rearm.sh`: claimable→rearm; **GL-host on mac→skip**, on linux→rearm; inflight-only→skip; empty→skip; feedback PR→rearm.
- [x] All dispatcher suites green except the pre-existing macOS `concurrency_cap` T4c host-flake.

Sibling of the GL-host inference recall PR (#2150).

### DIFF
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -1242,6 +1242,18 @@ resolve_worker_class() {
     [[ "$count" =~ ^[0-9]+$ ]] && DISPATCH_COUNT="$count"
 }
 
+# Returns 0 (should re-arm) iff the resolver would dispatch a CONCRETE class for
+# this worker lane — i.e. there is genuinely claimable work under the full
+# claimability gate (host capability, stackability, inflight_pr, owner, blocked).
+# The periodic safety re-arm uses this so it can never diverge from the actual
+# dispatch decision: a `defer` verdict (nothing claimable) or the '' lane-default
+# fallthrough (reservation resume / missing slice) does NOT re-arm; only a routed
+# class does. Sets DISPATCH_* as a side effect (resolve_worker_class).
+worker_rearm_should_fire() {
+    resolve_worker_class "$1"
+    (( DISPATCH_DEFER == 0 )) && [[ -n "$DISPATCH_CLASS" ]]
+}
+
 dispatch_role() {
     local role="$1"
     local trigger_file="$TRIGGERS_DIR/$role"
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: periodic worker re-arm delegates to the resolver (no more divergence)

**Verdict:** approve

### Summary

The dispatcher's periodic safety re-arm used to re-implement claimability via an inline `is_actionable` Python predicate embedded in a heredoc. That predicate checked `owner`/`blocked_by`/`model` but never learned about the `inflight_pr` guard (#2101) or `needs_gl_host` — so it kept re-arming the worker trigger on GL-host-only tasks (#1938/#1969/#2107) even on a Metal pane that can never claim them, fanning a no-op worker iteration into every tick. This PR replaces the inline predicate with `worker_rearm_should_fire()`, a thin wrapper around the existing `resolve_worker_class` (`scripts/fleet/fleet-dispatcher:1213-1243`) — the same function the real dispatch path already uses — so the safety re-arm structurally cannot diverge from the dispatch decision again.

### Verification performed

- Read `scripts/fleet/fleet-dispatcher` in full at the PR's head commit (the branch was checked out elsewhere when I tried `gh pr checkout`, so I materialized it via `git show origin/claude/rearm-via-resolver:scripts/fleet/fleet-dispatcher` instead).
- Confirmed `resolve_worker_class`'s body is untouched by this diff — `worker_rearm_should_fire` only calls it and reads `DISPATCH_DEFER`/`DISPATCH_CLASS` afterward, so every other call site (`dispatch_role`, `--resolve-class`) keeps its existing semantics. No shared-state leakage: `resolve_worker_class` resets all `DISPATCH_*` globals at entry, so the periodic re-arm's call and `dispatch_role`'s own call within the same tick can't see each other's stale values.
- Confirmed the periodic re-arm loop still only iterates `for _role in "worker"` (line 1675) — unchanged scope, so non-worker roles still never re-arm via this path (same as the old code's `ROLE_MODELS = {"worker": ...}` gate, now expressed as `WORKER_LANE_DEFAULT_CLASS[$role]` being unset for other roles → empty `DISPATCH_CLASS` → `worker_rearm_should_fire` returns false).
- Confirmed the new `--rearm-check` debug hook follows the exact same pattern as the pre-existing `--resolve-class` hook (both sit in the top-level `case` block before `main "$@"`, so neither requires `acquire_dispatcher_lock`).
- Ran the new `test_dispatcher_rearm.sh` against a scratch copy of `scripts/fleet/` with this PR's `fleet-dispatcher` overlaid (since the live worktree was locked) — all 6 cases pass, including the two that matter most: GL-host task on a mac pane → skip (the regression this PR fixes), same task on linux → rearm.
- Spot-checked for regressions by running two adjacent suites that also exercise `resolve_worker_class`/`DISPATCH_*` against the same scratch tree: `test_dispatcher_class_dispatch.sh` (7/7 pass) and `test_dispatcher_claimable_cap.sh` (9/9 pass).
- `bash -n` on the patched script — clean syntax.
- Grepped the tree for stale references to the removed `is_actionable` name in docs/other scripts — none found; the removal is self-contained to this file.

### ECS / render / fleet-script checklist

Not applicable (no C++/ECS/render/shader surface touched) — confirmed no pane-ID or role-name renames in `scripts/fleet/`, so the "stale identifier in inline comments" check doesn't apply either.

### Praise

- Deleting the re-implemented predicate in favor of calling the resolver directly is the right fix shape — it doesn't just patch the missing `needs_gl_host` check, it makes the *class* of bug (re-arm drifting from dispatch) structurally impossible to reintroduce.
- The `--rearm-check <role>` hook is a genuinely useful operator/test seam, not test-only scaffolding — it lets a human or another agent ask "would this fire?" without touching the live trigger file.
- `test_dispatcher_rearm.sh` covers the actual regression (T2/T3, GL-host on mac vs linux) plus the boundary cases that matter for "never re-arm on nothing" (T4 inflight-only, T5 empty slice) and the side-benefit claim (T6 feedback PR).

- [x] Already covered by the author's own test plan and independently re-verified above (new suite + two adjacent suites, all green).
- [ ] The author's claim that "all dispatcher suites green except the pre-existing macOS `concurrency_cap` T4c host-flake" was not independently re-run here (large suite); this flake is independently corroborated as a known pre-existing host issue, not attributable to this change.

Opus recheck not required.

Issue #1726, opened 2026-06-12
fleet: dispatcher churns no-op iterations on tasks workers refuse — design-blocked-PR-backed issues project as claimable and starve other classes


jakildev:
**Area:** scripts/fleet (fleet-state-scout projection, fleet_task_class.py, fleet-dispatcher)
**Model:** opus

## Symptom (observed live, 2026-06-11 ~00:28–00:37Z)

Both worker panes cycled `dispatch → pane returned to shell → re-dispatch [class=opus]` every ~2 minutes, indefinitely, completing no work. Side effect: fable-class tasks queued behind the churning task were starved — the resolver picked opus first on every tick, so the raised fable cap never engaged.

## Root cause chain

1. #1640 is `fleet:queued`, owner-free, no `fleet:blocked` → scout projects it `claimable` (slice `blocked` comes only from the `fleet:blocked` label, fleet-state-scout:500).
2. Its implementation PR #1700 is `fleet:wip` + `fleet:design-blocked` — so each dispatched worker iteration (correctly) refuses to claim it and exits.
3. `fleet_task_class.py` picks the lowest-numbered claimable candidate every tick → #1640 → `class=opus`, `more=1` (fable still queued) → trigger kept → immediate re-dispatch after each no-op exit. Loop.
4. Compounding: an empty/stale lane slice makes the dispatcher fall back to the lane-default class and dispatch anyway (documented fallback), so trigger-storms (e.g. bursts of label edits re-firing the scout) also produce no-op launches — observed on the sonnet-author lanes the same minutes.
5. Related race: two lanes resolved the same single claimable task within 10s (00:20:41/51 both `class=opus` for #1640) — the loser burns a launch.

## Interim mitigation (applied)

`fleet:blocked` added to #1640 with an explanatory comment (remove when #1700's design escalation resolves).

## Fix directions (pick at least the first)

1. **Projection-side:** scout marks a task non-claimable when an open PR references it as its implementation (any open PR with `Closes #N` / claim branch naming), or at minimum when that PR carries `fleet:wip` / `fleet:design-blocked`. The claimable definition should approximate "a fresh worker could actually start this".
2. **Worker-side backstop:** when an iteration declines every candidate it was dispatched for, it should leave a machine-readable skip marker (e.g. stamp `fleet:blocked` + comment, or a skip record the scout folds in) so the same no-op cannot repeat next tick.
[...]

PR #2152, opened 2026-06-30
fleet: dispatcher serves the next claimable class when the elected one is covered

jakildev:
## Summary

The follow-up to the cap work — the general starvation guard.

The claimable-count fan-out cap (#2101) caps each class to one worker per claimable item. But when the elected class was fully covered by racing workers, `dispatch_role` deferred the **whole tick** — even if another class was claimable (`DISPATCH_MORE`). So a slow or churning head of one class (a batch of sonnet feedback PRs, an opus task taking minutes to claim) could defer the *other* class for the duration: **cross-class starvation**.

### Fix

Make the resolver exclusion-aware and loop in `dispatch_role`: when the elected class is cap-saturated and another class is claimable, **re-resolve excluding the covered class and serve the next one** instead of deferring. Bounded — each pass excludes one more class and there are at most three (opus|fable|sonnet), so the loop runs a handful of times then finds a class with headroom, hits the uncapped path, or defers.

`fleet_task_class.resolve` gains an `exclude` set: excluded classes are dropped from consideration (not chosen, not counted toward `more`). Crucially, if excluding leaves nothing electable **but a claimable (excluded) class existed**, the verdict is `defer`, **not `''`** — `''` would wrongly fall the dispatcher through to a lane-default no-op dispatch. `fleet_task_class.py` takes an optional 4th arg (comma-separated exclude); `resolve_worker_class` threads it; `--resolve-class` gains an optional exclude arg as the test seam.

## Test plan
- [x] `ExcludeClasses` (resolver): exclude→elect-next, exclude-all→defer-not-empty, exclude-irrelevant→noop, exclude-on-empty→stays-empty.
- [x] `class_dispatch` T8: exclude threads through the dispatcher (`opus` → sonnet served; `opus,sonnet` → defer, not lane-default).
- [x] All dispatcher/resolver suites green bar the pre-existing macOS `concurrency_cap` T4c host-flake. Backward-compatible (3-arg resolve still works).

Builds on #2101 (cap) and #2151 (re-arm via resolver).

### DIFF
--- a/scripts/fleet/fleet-dispatcher
+++ b/scripts/fleet/fleet-dispatcher
@@ -1212,6 +1212,9 @@ build_dispatch_command() {
 # fleet_task_class.py for the resolution rules and output protocol.
 resolve_worker_class() {
     local role="$1"
+    # $2 = optional comma-separated classes to exclude (the cross-class fan-out
+    # re-resolve in dispatch_role, after a class is found cap-saturated).
+    local exclude="${2:-}"
     DISPATCH_CLASS="" DISPATCH_MODEL="" DISPATCH_EFFORT=""
     DISPATCH_MORE=0 DISPATCH_DEFER=0 DISPATCH_COUNT=""
     local lane_default="${WORKER_LANE_DEFAULT_CLASS[$role]:-}"
@@ -1224,7 +1227,7 @@ resolve_worker_class() {
     fi
     local out
     out=$(python3 "$FLEET_LIB_DIR/fleet_task_class.py" \
-        "$slice" "$lane_default" "$fable_blocked" 2>/dev/null) || out=""
+        "$slice" "$lane_default" "$fable_blocked" "$exclude" 2>/dev/null) || out=""
     if [[ "$out" == "defer" ]]; then
         DISPATCH_DEFER=1
         return 0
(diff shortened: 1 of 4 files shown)


jakildev (commented):
## Review — fleet: dispatcher serves the next claimable class when the elected one is covered

**Verdict:** approve

Fleet tooling only (`scripts/fleet/fleet-dispatcher` bash + `fleet_task_class.py`),
no engine C++ touched — engine ECS/render/lighting/math/serialization checklist
items are N/A.

### Verification performed

- Traced `dispatch_role`'s new `while` loop in `fleet-dispatcher` (lines
  ~1330–1379) by hand: each saturated-class pass appends to `_excluded` and
  re-resolves; the `_passes++ < 3` bound is a defensive backstop that in
  practice never binds — excluding all 3 classes (opus|fable|sonnet) always
  makes `fleet_task_class.resolve` return `"defer"` first (caught by the
  `DISPATCH_DEFER` check at the top of the next loop iteration), so the loop
  always exits via the defer path or the claim-headroom break, never via
  exhausting the pass counter.
- Confirmed the two `resolve_worker_class` call sites (`dispatch_role` and the
  `--resolve-class` test seam) both thread the new `exclude` arg; no orphaned
  3-arg callers left over from the refactor.
- Confirmed `more` in `fleet_task_class.resolve` correctly excludes excluded
  classes from `servable_classes` before computing `servable_classes -
  {chosen[0]}` — matches the docstring's "not counted toward more" claim.
- Pulled `fleet_task_class.py` + `tests/test_fleet_task_class.py` from the PR
  branch via `gh api .../contents` into an isolated `/tmp` dir (couldn't `gh pr
  checkout` here — the branch is already checked out in another active
  worktree) and ran the suite directly: **37/37 pass**, including all 4 new
  `ExcludeClasses` cases (elect-next, exclude-all→defer-not-empty,
  exclude-irrelevant→noop, exclude-on-empty-slice→stays-empty/lane-default).
  This independently confirms the PR's own test-plan checkboxes rather than
  just trusting them.
- `_excluded`/`_passes` are `local`-scoped — correct, since `dispatch_role`
  fires once per role per tick and state must not leak across calls.
  `_`-prefixed locals match existing convention (`local _role _active` at
  fleet-dispatcher:1662).

### Praise

- The `defer`-not-`''` distinction when `excluded_any` is true (vs. plain
  `''` for "nothing was ever excluded") is the right call and is exactly the
  kind of subtle protocol detail that's bitten this dispatcher before
  (#1726-style no-op churn) — good that it's both implemented and pinned by
  `test_exclude_on_empty_slice_stays_empty` as a negative case.

- [x] Already covered by the PR's own test plan + my independent re-run above.

Opus recheck not required.

Issue #2155, opened 2026-06-30
render: no-priority perf fast-path — skip triangleEntityIds read in f_trixel_to_framebuffer (#2122 item 3)


jakildev:
Follow-up to #2122 (item 3, marked optional) / #2023's work list. The
deterministic tier gate (item 1) shipped in the #2122 PR; this is the optional
no-priority perf fast-path.

`f_trixel_to_framebuffer` reads `triangleEntityIds` unconditionally at
finalization (to decode the per-trixel priority tier), where it previously read
that channel only when the fragment was hovered — one extra texture fetch per
finalization fragment on the DEFAULT path (pre-existing since #1989, not
introduced by #2023). Add a scene-level `anyPerTrixelPriority` uniform that lets
the shader skip the read when no voxel in the scene carries a non-zero per-trixel
priority, restoring perf-identity (not just output byte-identity) on the default
path.

**Area:** engine/render (f_trixel_to_framebuffer GLSL + Metal)
**Model:** opus
**Blocked by:** (none)

## Acceptance criteria
- A scene-level `anyPerTrixelPriority` flag (set when any voxel carries a
  non-zero per-trixel priority) is plumbed into the `f_trixel_to_framebuffer`
  frame data on BOTH backends (GLSL + Metal); when false, the finalization path
  skips the `triangleEntityIds` decode read.
- Output is byte-identical with the flag on (the priority decode is unchanged)
  AND with it off (no priority → the skipped read would have decoded tier 0
  anyway). Verify with `render-verify` / `--depth-probe` on a priority scene
  (`--only interpenetrate`) and a no-priority scene.
- The #2122 tier gate (`scripts/depth-tier-verify.py`) still passes (priority
  path unaffected).

## Context
Optional perf item from #2122 — the only one of the three #2122 items that
touches the hot finalization shader on both backends, so it carries GL/Metal
parity + cross-host-smoke surface the deterministic tier gate (item 1) and the
demo (item 2) do not. Split out so each lands as a focused, separately
smoke-validated PR.

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-2155.md`), and it is not opted out (`human:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

[...]

Issue #2154, opened 2026-06-30
render: canvas_stress per-ENTITY priority-swap orbit demo (#2122 item 2)


jakildev:
Follow-up to #2122 (item 2) / #2023's work list "Demo 1". The deterministic
headless per-trixel tier gate (item 1) shipped in the #2122 PR
(`scripts/depth-tier-verify.py` + `--depth-probe-assert X,Y,tier=N`); this is the
remaining demo item.

Add a `canvas_stress` opt-in scene (e.g. `--only orbitswap`, mirroring the
existing `interpenetrate` / `smallzoom` opt-in groups) that animates N detached
units orbiting under a camera yaw-ramp, each carrying a per-ENTITY
`C_EntityCanvas::depthPriority_` (the #1958 entity-foreground carrier — distinct
from the per-trixel carrier #2023 fixed). The point is to show the units swap
front/back ordering cleanly (swap-not-flicker) as the orbit/yaw advances.

**Area:** creations/demos/canvas_stress
**Model:** opus
**Blocked by:** (none)

## Acceptance criteria
- A new opt-in `--only`-gated scene spawns N orbiting per-entity-priority units;
  the default run + every existing manifest shot stay byte-identical (the group
  never spawns unless explicitly requested, like `interpenetrate`).
- Gate the swap behaviour on the deterministic tier assertion, NOT committed
  image baselines (canvas_stress refs are red on master per #1944, and the
  re-voxelize path has documented round-to-cell speckle non-determinism — both
  reasons the #2122 gate uses `--depth-probe-assert ...,tier=N` instead of
  image compare). A `--depth-probe-assert <overlap>,tier=1` reading at a swap
  overlap (entity-foreground tier 1) is the natural deterministic check.
- `fleet-build --target IRCanvasStress` clean; the scene runs headless under
  `--auto-screenshot` without crashing.

## Context
#2023 closed the load-bearing per-trixel carrier gap; its work list named two
integration demos. The per-trixel interpenetration demo already exists
(`--only interpenetrate`, gated by #2122 item 1). This is the per-ENTITY swap
demo (#1958 carrier), still outstanding. Build the gate on top of the #2122
`assertCompositeDepthTier` / `scripts/depth-tier-verify.py` machinery.

jakildev:
[...]

PR #2143, opened 2026-06-30
demos: register custom CLI flags on IRArgs to fix strict-init rejection (#2103)

jakildev:
## Summary
The P2 IRArgs migration (#2076) switched four demos to `IREngine::init(argc, argv)` — which runs the engine `IRArgs` parser **strictly** (unknown arg → `exit(2)`) — without registering their hand-rolled custom flags. So each demo rejected its own debug flags:

- **canvas_stress** — `--only`, `--sweep-yaw`, `--sweep-frames`, `--debug-overlay`, `--depth-probe[-assert]`, `--yaw`, `--zoom`, `--subdivisions`, `--no-spin`, … (17 flags). Blocked every render-validation task driving it with a debug flag (#1970, #2122).
- **gpu_particles** — `--spawns-per-frame`
- **ir_voxel_yaw** — `--spin-yaw`, `--yaw`, `--zoom`
- **lua_perf_grid** — `--auto-profile`, `--grid-size`, `--zoom`, `--subdivision-mode`, `--base-subdivisions`

Each demo's flags are now registered on `IREngine::args()` before `init` and read back after, retiring the hand-rolled `strcmp`/`atoi` parsers (the canonical `fog_demo` / `shape_debug` adoption).

### IRArgs: new `FLOAT_LIST` type
canvas_stress's multi-value `--sweep-yaw <from> <to> <n>` / `--sweep-frames <n> <settle>` needed a fixed-arity float-list arg type, so I added `Parser::numbers(name, help, count)` / `getFloats()` (`Type::FLOAT_LIST`). This preserves the established **space-separated** spellings that `docs/design/jitter-validation-harness.md` and the planned #1954 jitter harness rely on (rather than forcing a comma-separated rewrite). Accepts both `--sweep-yaw 0 360 8` and `--sweep-yaw=0,360,8`; too few values is a clean `expects N float values` + exit(2).

### Incidental fix
ir_voxel_yaw's `--spin-yaw` adopts shape_debug's optional-int treatment, which also fixes a latent ordering bug: its `--spin-yaw + --auto-screenshot` reinterpret block read `g_autoWarmupFrames` **before** `init` had populated it (dead since the P2 migration, masked by the `exit(2)`).

## Test plan
- [x] `--help` on canvas_stress lists all 17 custom flags (incl. `--sweep-yaw <float> <float> <float>`)
[...]

PR #1989, opened 2026-06-24
render: per-trixel priority carrier + interpenetration demo (#1960)

jakildev:
Implements **#1960** — per-trixel priority carrier + interpenetration demo (epic #1884 child, sibling of merged #1958/#1974).

**Carrier: top-2-bit steal of the 64-bit entity id (N=3 tiers). Single decode chokepoint.**
Architect steward direction: carrier = top-K-bit steal of `triangleEntityIds`, K=2 (tiers 0–3, covers N=3 with headroom). Neutralizes the "picking trap" structurally — `decodeEntityId`/`decodePriority`/`encodeEntityIdWithPriority` in shared shader includes + C++ twin; all four id readers route through the chokepoint.

### What shipped
- **Carrier + N-tier partition** (`ir_render_types.hpp`): `kDepthForegroundTierCount=3`, per-tier band layout helpers, K=2 carrier bit-layout, `decodeCarrierEntityId`/`decodeCarrierPriority`, full partition-layout `static_assert` suite.
- **Single decode chokepoint** — `decodeEntityId`/`decodePriority`/`encodeEntityIdWithPriority` in `ir_iso_common.{glsl,metal}` + C++ twin. All four id readers route through it.
- **Both backends** (GLSL + Metal): stage-2 packs `Voxel.reserved` low-2-bits into the id; finalization resolves `tier = max(perEntityTier, perTrixelTier)` + per-tier clamp.
- **Authoring** `C_VoxelSetNew::changeVoxelPriority/All`; **probe** reports the resolved tier (part G).
- **Demo** `canvas_stress --only interpenetrate` (opt-in, default scene byte-identical).

### Verification (macOS/Metal)
- Default priority 0 ⇒ byte-identical: `shape_debug` render-verify **24/24 PASS** (0 max-delta, inf PSNR).
- `--only interpenetrate --depth-probe 640,360`: far red unit `tier=2`, near blue unit `tier=0`.
- `--only interpenetrate --depth-probe-assert 640,360`: PASS.

### Two integration limits — documented + deferred to #2023
1. **Intra-canvas:** per-trixel priority arbitrates ACROSS canvases at finalization. Two voxel sets on ONE canvas resolve at the canvas `atomicMin` raster, upstream of the partition.
2. **Rotating detached gap:** `c_revoxelize_detached` hardcodes `reserved=0`; a STATIC detached unit carries priority, a ROTATING `DETACHED_REVOXELIZE` unit does NOT.

[...]

PR #2156, opened 2026-06-30
render: headless per-trixel-priority tier gate for canvas_stress (#2122)

### LINKED ISSUE #2122: render: per-trixel priority follow-ups — deterministic tier gate + orbiting swap demo + no-priority perf fast-path (#2023 residual)
Follow-up to #2023, which closed the load-bearing gap (per-trixel `reserved` now
rides the rotating `DETACHED_REVOXELIZE` MODE 1 inverse-resample fill, so a spinning
detached solid keeps its per-trixel depth priority). Two integration items from
#2023's work list remain, plus a tooling dependency surfaced during verification.

## 1. Deterministic headless tier gate (item 2, de-risked)
#2023 verified the rotating per-trixel priority via `canvas_stress --only interpenetrate`
(far cube rotated + tagged the top voxel tier, renders in front of the nearer cube;
`--depth-probe` at the overlap reads `tier=2`). Two gaps remain before this is a CI gate:
- **Blocked on #2103.** canvas_stress's custom flags (`--only`, `--no-spin`,
  `--depth-probe`, ...) hit IRArgs strict validation and `exit(2)` before render, so
  the gate can't be driven from the CLI yet. #2023's verification used a temporary
  in-code scene force. Once canvas_stress migrates to IRArgs, wire the gate.
- **Extend `--depth-probe-assert` to assert a specific tier.** Today it only asserts
  "composite wrote non-background depth" (PASS/FAIL). Add an
  `[depth-probe-assert] ... tier=N result=PASS|FAIL` variant so a headless run catches
  a future regression that drops the rotating per-trixel carrier. The interpenetrate
  far cube holds a fixed non-identity pose under `--no-spin`, so the readback is
  deterministic and still exercises MODE 1.

## 2. Demo 1 — animated per-entity priority-swap (item 2)
#2023's work list also names Demo 1: N orbiting units with per-ENTITY `depthPriority_`
(the #1958 carrier on `C_EntityCanvas`, distinct from the per-trixel carrier #2023
fixed), camera yaw-ramp, swap-not-flicker. Committed render-verify baselines were
de-risked in #2023's body: canvas_stress refs are red on master (#1944) and the
re-voxelize path has documented round-to-cell speckle non-determinism, so gate on the
deterministic tier assertion above rather than image baselines.

## 3. (optional) No-priority perf fast-path (item 3)
`f_trixel_to_framebuffer` reads `triangleEntityIds` unconditionally (for the tier
decode) where it previously read only when hovered — one extra texture fetch per
finalization fragment (pre-existing since #1989, not introduced by #2023). A
scene-level `anyPerTrixelPriority` uniform could skip the read when no voxel carries
priority, restoring perf-identity (not just output byte-identity) on the default path.

**Area:** engine/render (f_trixel_to_framebuffer, depth_probe) + creations/demos/canvas_stress
**Model:** opus
**Blocked by:** #2103 (item 1's CLI gate)

jakildev:
## Summary

Delivers **#2122 item 1**: the deterministic headless gate for the #1960
per-trixel-priority carrier — the positive ENABLED-path test the carrier needs
(byte-identity at default priority 0 only proves the OFF path is a no-op, never
that the carrier survives the rotating re-voxelize MODE 1 fill — the #1989 /
#2023 trap; see `engine/render/CLAUDE.md` "Default-off features need a positive
enabled-path test").

- **`depth_probe.hpp`** — `assertCompositeDepthTier(px, N)` reads back the
  composite depth winner and emits `[depth-probe-assert] … tier=N expected=M
  result=PASS|FAIL`. The tier/iso/face decode is factored into
  `detail::decodeComposite`, shared with `logCompositeDepth` (one source of
  truth for the N-tier partition; no behavior change to the existing probe).
- **`canvas_stress`** — `--depth-probe-assert` grows the `X,Y,tier=N` form (the
  `X,Y` depth-write guard is unchanged); the `DepthProbeAssert` system
  dispatches to the tier guard when a tier is given. Pure readback — a flagless
  run is byte-identical.
- **`scripts/depth-tier-verify.py`** — build → run → parse gate over
  `--only interpenetrate --no-spin --no-auto-rotate --depth-probe-assert 639,362,tier=2`,
  exiting non-zero if the far priority unit decodes `tier=0` (carrier dropped).

Stacked on **#2143** (restores canvas_stress's custom flags, #2103) — this is
the part of #2122 that was `fleet:blocked` on the CLI gate. The other two #2122
items are filed as follow-ups: **#2154** (per-entity priority-swap orbit demo)
and **#2155** (no-priority perf fast-path). They were split out so each lands as
a focused PR — item 3 in particular touches the hot finalization shader on both
backends (GL/Metal parity + cross-host smoke) that this gate does not.

## Test plan

- [x] `fleet-build --target IRCanvasStress` + `IRPerfGrid` (both depth_probe
      consumers) clean on macOS/Metal.
- [x] `scripts/depth-tier-verify.py` → **750/750 frames PASS** at (639,362),
      exit 0; **FAIL + exit 1** at a background pixel (regression-catch path
      verified).
- [x] `simplify` clean (reuse fan-out: the new tier guard intentionally mirrors
      the existing #1957 depth-write sibling; shared logic already factored).
- No screenshots: the change is pure-readback debug tooling + a debug flag — the
  default render is byte-identical (the probe registers no system unless its flag
  is set, and the interpenetrate scene is unchanged), so before/after captures
  would be an identical pair. The meaningful verification is the enabled-path
  gate above.
- Cross-host smoke: not required — no shader/pipeline/render-output path changed
  (pure GPU→CPU depth readback, backend-agnostic).

Closes #2122

### DIFF
--- a/creations/demos/canvas_stress/main.cpp
+++ b/creations/demos/canvas_stress/main.cpp
@@ -172,17 +172,31 @@ struct CanvasStressSettings {
     // floating canary cube clips behind the SDF floor at high zoom).
     bool depthProbeSet_ = false;
     ivec2 depthProbePixel_{0, 0};
-    // `--depth-probe-assert X,Y` (#1957): the depth-write regression guard. Like
-    // --depth-probe but emits a `[depth-probe-assert] … result=PASS|FAIL` verdict
-    // line — PASS iff the composite stored a non-background depth at (X,Y), i.e.
-    // the detached-canvas composite WROTE the depth attachment there. Aim it at a
-    // texel inside a world-placed detached solid (canonical:
-    // `--only canary --no-spin --no-auto-rotate --depth-probe-assert 321,210`,
-    // which the canary covers on both backends). FAILs if a future pass disables
-    // the composite depth-write. Off by default; registers no system when unset,
-    // so a flagless run is byte-identical.
+    // `--depth-probe-assert X,Y[,tier=N]` (#1957 / #1960): the composite
+    // regression guard. Two forms, both emitting a single
+    // `[depth-probe-assert] … result=PASS|FAIL` verdict line:
+    //   - `X,Y` (#1957) — DEPTH-WRITE guard: PASS iff the composite stored a
+    //     non-background depth at (X,Y), i.e. the detached-canvas composite WROTE
+    //     the depth attachment there. Aim it at a texel inside a world-placed
+    //     detached solid (canonical:
+    //     `--only canary --no-spin --no-auto-rotate --depth-probe-assert 321,210`,
+    //     which the canary covers on both backends). FAILs if a future pass
+    //     disables the composite depth-write.
+    //   - `X,Y,tier=N` (#1960) — per-trixel-priority TIER guard: PASS iff the
+    //     composite winner at (X,Y) decodes to the #1960 tier N. This is the
+    //     positive ENABLED-path gate the per-trixel carrier needs (the #2122
+    //     deterministic headless tier gate — byte-identity at default priority 0
+    //     cannot prove the carrier works). Canonical:
+    //     `--only interpenetrate --no-spin --no-auto-rotate --depth-probe-assert <overlap>,tier=2`,
+    //     where the far priority unit holds a fixed non-cardinal pose (MODE 1) so
+    //     the overlap reads a deterministic `tier=2`; without the #2023 carrier it
+    //     decodes `tier=0` and FAILs.
+    // `depthProbeAssertTier_` is -1 for the depth-write form, >= 0 for the tier
+    // form. Off by default; registers no system when unset, so a flagless run is
+    // byte-identical.
     bool depthProbeAssertSet_ = false;
     ivec2 depthProbeAssertPixel_{0, 0};
+    int depthProbeAssertTier_ = -1;
     // readConfig() runs AFTER parseArgs() (it needs IREngine::init), so a
     // config `auto_rotate` would otherwise clobber an explicit
     // --auto-rotate / --no-auto-rotate flag. Latch CLI intent so config only
@@ -930,6 +944,30 @@ void applyDepthProbe(const std::string &value, bool &set, ivec2 &pixel, const ch
     }
 }
 
+// Parse "--depth-probe-assert X,Y" (depth-write guard, tier -1) or
+// "X,Y,tier=N" (per-trixel-priority tier guard, tier >= 0). The 3-field form is
+// tried first; a bad value leaves the assert off so the run stays byte-identical.
+void applyDepthProbeAssert(const std::string &value) {
+    int px = 0;
+    int py = 0;
+    int tier = -1;
+    if (std::sscanf(value.c_str(), "%d,%d,tier=%d", &px, &py, &tier) == 3) {
+        g_settings.depthProbeAssertSet_ = true;
+        g_settings.depthProbeAssertPixel_ = ivec2(px, py);
+        g_settings.depthProbeAssertTier_ = tier;
+    } else if (std::sscanf(value.c_str(), "%d,%d", &px, &py) == 2) {
+        g_settings.depthProbeAssertSet_ = true;
+        g_settings.depthProbeAssertPixel_ = ivec2(px, py);
+        g_settings.depthProbeAssertTier_ = -1;
+    } else {
+        IR_LOG_WARN(
+            "--depth-probe-assert: expected X,Y or X,Y,tier=N (e.g. 321,210 or "
+            "321,210,tier=2); ignoring '{}'",
+            value
+        );
+    }
+}
+
 // Declare canvas_stress's custom flags on the engine-owned parser. Must run
 // before IREngine::init(argc, argv), which performs the single strict parse of
 // engine-common + these flags (see engine/CLAUDE.md "CLI args go through
@@ -996,7 +1034,8 @@ void registerArgs() {
     );
     args.string(
         "--depth-probe-assert",
-        "Depth-write regression guard at framebuffer pixel X,Y; emits PASS|FAIL (#1957)",
+        "Composite regression guard at framebuffer pixel: X,Y = depth-write guard "
+        "(#1957); X,Y,tier=N = per-trixel-priority tier guard (#1960). Emits PASS|FAIL",
         ""
     );
 }
@@ -1052,12 +1091,7 @@ void applyArgs() {
         );
     }
     if (args.wasProvided("--depth-probe-assert")) {
-        applyDepthProbe(
-            args.getString("--depth-probe-assert"),
-            g_settings.depthProbeAssertSet_,
-            g_settings.depthProbeAssertPixel_,
-            "--depth-probe-assert"
-        );
+        applyDepthProbeAssert(args.getString("--depth-probe-assert"));
     }
     if (args.wasProvided("--sweep-yaw")) {
         const std::vector<float> &v = args.getFloats("--sweep-yaw");
@@ -1267,17 +1301,26 @@ void initSystems() {
         );
     }
 
-    // #1957 depth-write regression guard — registered only with
+    // #1957 / #1960 composite regression guard — registered only with
     // --depth-probe-assert. Runs after the framebuffer composite and emits a
-    // PASS/FAIL verdict line so a headless run catches a future pass that
-    // disables the detached-canvas composite depth-write.
+    // PASS/FAIL verdict line. tier < 0 is the #1957 depth-write guard (catches a
+    // future pass that disables the detached-canvas composite depth-write);
+    // tier >= 0 is the #1960 per-trixel-priority tier guard (catches a future pass
+    // that drops the carrier so the priority solid loses the depth contest).
     if (g_settings.depthProbeAssertSet_) {
         const ivec2 assertPixel = g_settings.depthProbeAssertPixel_;
+        const int assertTier = g_settings.depthProbeAssertTier_;
         renderPipeline.push_back(
             IRSystem::createSystem<C_Camera>(
                 "DepthProbeAssert",
                 [](C_Camera &) {},
-                [assertPixel]() { IRPrefab::DepthProbe::assertCompositeWritesDepth(assertPixel); }
+                [assertPixel, assertTier]() {
+                    if (assertTier >= 0) {
+                        IRPrefab::DepthProbe::assertCompositeDepthTier(assertPixel, assertTier);
+                    } else {
+                        IRPrefab::DepthProbe::assertCompositeWritesDepth(assertPixel);
+                    }
+                }
             )
         );
     }
--- a/engine/prefabs/irreden/render/depth_probe.hpp
+++ b/engine/prefabs/irreden/render/depth_probe.hpp
@@ -108,6 +108,44 @@ inline CompositeDepthSample readbackCompositeDepth(IRMath::ivec2 px) {
     return sample;
 }
 
+namespace detail {
+
+/// The @c [depth-probe] integer-encode decode, factored so the diagnostic probe
+/// and the tier-assert guard partition @c enc the same way (one source of truth
+/// for the #1960 N-tier layout). @c enc > kDepthForegroundCeil is WORLD content
+/// (tier 0), encoded as @c iso*4 + face; @c enc inside the reserved band is a
+/// FOREGROUND fragment whose disjoint sub-range names its priority tier (1..N-1).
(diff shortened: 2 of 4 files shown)


jakildev (commented):
## Review — render: headless per-trixel-priority tier gate for canvas_stress (#2122)

**Verdict:** approve

Stacked on #2143 (now `fleet:approved`, open). Reviewing this PR's own
diff only — #2143 was reviewed separately.

### What this does

Adds the deterministic ENABLED-path guard the #1960 per-trixel-priority
carrier needs: `--depth-probe-assert X,Y,tier=N` reads back the composite
depth winner and asserts it decodes to tier `N`, catching a future regression
that drops the carrier (the #1989/#2023 trap, where byte-identity at default
priority 0 only proves the OFF path is a no-op). The existing `X,Y` form
(#1957 depth-write guard) is unchanged. A new `scripts/depth-tier-verify.py`
wraps the build→run→parse gate.

### Verification performed

- `engine/prefabs/irreden/render/depth_probe.hpp`: diffed
  `detail::decodeComposite` against the pre-refactor inline logic in
  `logCompositeDepth` line-by-line — identical arithmetic, just extracted
  into a shared helper and renamed locals (`enc`→`enc_`, `tier`→`tier_`,
  etc). No behavior change, matches the PR body's claim.
- `assertCompositeDepthTier`: `match` requires `sample.valid_ &&
  decoded.tier_ == expectedTier`, so an out-of-range readback (`valid_ ==
  false`) always reports FAIL regardless of `decoded.tier_`'s garbage value
  from the zero-initialized sample — the body's claim about this holds.
- `canvas_stress/main.cpp`: `applyDepthProbeAssert` tries the 3-field
  `sscanf` form first, falls back to the 2-field form — correctly
  discriminates `"639,362"` (2 matches) from `"639,362,tier=2"` (3 matches).
  The `DepthProbeAssert` system lambda correctly dispatches on
  `assertTier >= 0`.
- Confirmed `perf_grid/main.cpp` only calls `logCompositeDepth` (unchanged
  signature) — the `depth_probe.hpp` refactor doesn't risk that build, so
  the PR body's "both depth_probe consumers" build claim is consistent with
  the actual call sites.
- `scripts/depth-tier-verify.py`: regex matches the actual `IR_LOG_INFO`
  format string in `assertCompositeDepthTier` (`pixel=`, `normDepth=`,
  `rawDist=`, `enc=`, `tier=`, `expected=`, `result=`) field-for-field. It
  won't accidentally match the older 1957-form `[depth-probe-assert]` line
  (no `tier=`/`expected=` there), so there's no ambiguity between the two
  guard forms sharing one log prefix. `--timeout`/`--auto-screenshot` usage
  matches documented `ir-run` flags.
- `engine/render/CLAUDE.md` doc update accurately describes both
  `--depth-probe-assert` forms and points at the new script.
- No shader/`.glsl`/`.metal`/GPU-buffer changes — pure CPU-side readback +
  CLI/test tooling, consistent with the "byte-identical, flagless run
  unaffected" claim.

### Nits

- `creations/demos/canvas_stress/main.cpp:950-957` — a malformed-but-3-field
  value like `--depth-probe-assert 639,362,tier=-1` parses successfully
  (`sscanf` accepts negative integers) and is silently treated as the
  depth-write guard (`assertTier < 0` check at the call site), not rejected.
  Harmless given this is a debug-only dev flag, but a `tier < 0` warn-and-
  ignore in `applyDepthProbeAssert` would be more honest than silent
  reinterpretation.

### Praise

- Factoring `decodeComposite` out of `logCompositeDepth` before reusing the
  same partition logic in the new assert guard (rather than duplicating the
  tier-decode arithmetic) keeps the #1960 N-tier layout single-sourced —
  exactly the kind of share that prevents the two readers from drifting.

- [x] Already covered by the PR body: `fleet-build` clean for both
  `IRCanvasStress`/`IRPerfGrid`, `depth-tier-verify.py` 750/750 PASS at
  (639,362), FAIL-path verified at a background pixel.

Escalation: please Opus-recheck before merge (touches:
engine/prefabs/irreden/render/depth_probe.hpp — the composite depth-tier
bit-partition decode this PR refactors and extends has been the site of
several subtle defects in this exact subsystem (#1958, #1959, #2010,
#2023). I verified the refactor is byte-identical to the pre-existing logic
and the new guard's match/invalid-readback semantics are correct, but this
class of encoding invariant warrants Opus's deeper pass per protocol even
though my own read came back clean.

Opus recheck required: per-trixel-priority depth-tier decode logic, a
subsystem with a track record of subtle encoding bugs — escalating despite
a clean first-pass read.

jakildev (commented):
## Opus recheck — render: per-trixel-priority tier gate for canvas_stress (#2122)

**Verdict:** approve

Stacked on #2143 (`fleet:approved`, open). Reviewed this PR's own diff only.

Sonnet escalated because `depth_probe.hpp` is the composite depth-tier
bit-partition decode — a subsystem with a track record of subtle encoding bugs
(#1958/#1959/#2010/#2023) — and asked for a deeper pass on the decode invariant
even though its own read came back clean. I confirm Sonnet's read on all three
load-bearing points, and verified the one thing a first pass can't cheaply prove:
that the extracted decode is the true inverse of the encode side.

### `detail::decodeComposite` is a byte-identical extraction — confirmed

Diffed the new helper against the pre-existing inline logic in `logCompositeDepth`
statement-by-statement: identical arithmetic (`roundHalfUp` → ceil test → slot →
`clamp(count-1-slot, 1, count-1)` → tier-center-relative `encRel` → floored
`face`/`iso`), only locals renamed and `tier_` defaulted to 0 (set solely in the
foreground branch, matching the old `int tier = 0`). No behavioral change to the
diagnostic log path.

### The decode is the genuine inverse of the encode — verified against constants

This is the part that matters now that the same decode feeds a load-bearing
PASS/FAIL rather than an informational log. Against `ir_render_types.hpp`:
`depthForegroundTierLo(tier) = kMin + (count-1-tier)*width`, so for `enc` inside
tier T's band `slot = (enc-kMin)/width = count-1-T` and
`tier = count-1-slot = T`. Exact round-trip; the `clamp(…, 1, count-1)` correctly
pins band-edge cases to a valid foreground tier and never collides with world
tier 0 (`enc > kDepthForegroundCeil`). The `static_assert` block in the header
already guards the band/tier partition, so the decode can't silently drift from
the encode layout.

### New `assertCompositeDepthTier` guard semantics — correct

`match = sample.valid_ && decoded.tier_ == expectedTier`. A failed readback
(`valid_ == false`) decodes to tier 0 but the `valid_` conjunct forces FAIL for
*any* expected tier — including `expectedTier == 0` — so an out-of-range probe
can never masquerade as a pass. This is the positive ENABLED-path gate the #1960
carrier needs, addressing exactly the #1989/#2023 trap where byte-identity at
default priority 0 only proves the OFF path is a no-op. Pure readback, no system
registered unless `--depth-probe-assert` is provided, so a flagless run stays
byte-identical — confirmed in `initSystems()`.

### Parse + harness — sound

`applyDepthProbeAssert` tries `%d,%d,tier=%d` first and falls through to `%d,%d`
on a plain `X,Y` (sscanf returns 2, `tier=` literal unmatched → depth-write form,
`tier_ = -1`). `depth-tier-verify.py`'s regex
`tier=(-?\d+)\s+expected=(-?\d+)\s+result=(PASS|FAIL)` matches the emitted format
and exits non-zero on any FAIL or a missing verdict line. The PR body's
750/750 PASS at the overlap + FAIL-path-at-background-pixel is the right
empirical check that the decode produces the expected tier at a known
interpenetration.

### Non-blocking nits (do NOT re-review)

- An explicit `--depth-probe-assert X,Y,tier=-1` parses (sscanf returns 3,
  `tier_ = -1`) and then routes to the depth-write guard rather than a tier
  guard. Harmless (−1 isn't a real tier) but slightly surprising; a one-line
  "tier must be ≥ 0" note or guard could remove the ambiguity. Not worth a round.
- A malformed `X,Y,N` (missing `tier=`) silently degrades to the depth-write
  guard instead of warning. Acceptable given the documented form.

Diagnostic-only, default-off, byte-identical refactor with a correct,
empirically-gated decode inverse. Ship it.

🤖 Opus recheck (review-pr skill)

jakildev:
No code change for the flagged nit. The sonnet review's nit (`creations/demos/canvas_stress/main.cpp:950-957` — a malformed `tier=-1` value silently routes to the depth-write guard instead of being rejected) was re-evaluated in the follow-up Opus recheck pass, which explicitly listed it under "Non-blocking nits (do NOT re-review)" and called it "Not worth a round." Deferring to that verdict — clearing `fleet:has-nits` without amending. `fleet:approved` stands. — worker

jakildev:
Merger: base PR #2143 merged; re-targeted this PR to `master` but the rebase onto current master conflicts. The child commits cannot be replayed cleanly — worker will resolve semantically.

Conflicted files:
- `engine/ir_args.cpp` — master: `00a993ee demos: register custom CLI flags on IRArgs to fix strict-init rejection (#2103) (#2143)`; PR: `809d5467 fix(args): document trailing-comma ambiguity in splitInlineFloatList`

— fleet merger

jakildev:
Resolved semantic conflict via inherited-prefix drop: base PR #2143 squash-merged to master (00a993ee), so my branch's inherited ir_args.cpp/.hpp + demo-flag-registration commits (7744881e/a0a8e76d/809d5467) were stale copies of now-merged work. Rebased with `git rebase --onto origin/master 809d5467` to drop the inherited prefix and replay only the genuine #2122 commit onto current master — no textual conflicts. Diff is now the 4 #2122 files only (canvas_stress/main.cpp, depth_probe.hpp, render/CLAUDE.md, depth-tier-verify.py). Build clean (IRCanvasStress). Reviewer please re-evaluate the rebased diff. — worker

jakildev (commented):
## Review — render: headless per-trixel-priority tier gate for canvas_stress (#2122)

**Verdict:** approve

Re-review after a mechanical rebase, not a content change. Timeline: this PR
was already reviewed and approved by both Sonnet and Opus while stacked on
#2143. #2143 then squash-merged to master, the merger's auto-retarget hit a
rebase conflict on `engine/ir_args.cpp` (inherited commits from #2143 colliding
with the now-merged copy on master), and the worker resolved it via
`git rebase --onto origin/master 809d5467` to drop the stale inherited prefix
and replay only the genuine #2122 commit — the pattern from the #1789/#1811
`--onto` playbook. Verifying that resolution, not re-deriving a review from
scratch.

### Prior-review resolution

- ✅ `creations/demos/canvas_stress/main.cpp:950-957` — Sonnet's original nit
  (a malformed `tier=-1` value silently routes to the depth-write guard
  instead of being rejected) — the follow-up Opus recheck explicitly listed
  this under "Non-blocking nits (do NOT re-review)" and called it "not worth
  a round." Not re-flagged; carrying that verdict forward unchanged.

### Verification performed

- **Rebase scope, not content, changed.** `git log origin/master..HEAD` shows
  exactly one commit (`c40cceca`) directly on top of current
  `origin/master` (`4c3d6d05`) — no leftover inherited commits, no merge
  markers. `git diff origin/master --stat` shows exactly the 4 files the
  original review covered (`creations/demos/canvas_stress/main.cpp`,
  `engine/prefabs/irreden/render/depth_probe.hpp`, `engine/render/CLAUDE.md`,
  `scripts/depth-tier-verify.py`) — no `engine/ir_args.*` hunks, confirming
  the inherited #2143 commits were fully dropped rather than partially
  replayed.
- **Diff content matches what was previously approved.** Read the full diff
  against master: `detail::decodeComposite` extraction, the tier-partition
  round-trip math, `assertCompositeDepthTier`'s `valid_`-gated match logic,
  `applyDepthProbeAssert`'s two-form `sscanf` parse, and
  `scripts/depth-tier-verify.py`'s regex/gate — all identical in substance
  to what both the Sonnet and Opus passes already verified statement-by-
  statement (byte-identical decode extraction, correct encode/decode
  round-trip against `ir_render_types.hpp` constants, `valid_`-gated PASS
  semantics). No hunk suggests a hand-edit crept in during the rebase.
- **Build verified clean post-rebase.** `fleet-build --target IRCanvasStress`
  — compiles and links clean on the current worktree HEAD (this exact
  commit, not a stale binary).

No new findings. The rebase was purely mechanical; the underlying change is
the same one already Opus-approved.

- [x] Already covered by the original review: `depth-tier-verify.py` 750/750
      PASS at (639,362), FAIL-path verified at a background pixel.

Opus recheck not required: content is unchanged from the already-Opus-
approved diff; only the rebase mechanics were re-verified here.

PR #2161, opened 2026-06-30
engine: IRArgs multi-value + enum flags; canvas_stress declarative migration

jakildev:
## Summary
- **IRArgs::numbers(name, help, count)** — fixed-arity float-list flags (`--sweep-yaw 0 6.28 17`). `getNumbers()` returns the parsed values; `--help` shows the arity.
- **IRArgs::enumValue(name, help, {allowed…}, default)** — a value validated against a fixed set (`--debug-overlay shadow`), rejecting anything else with a listing diagnostic. `getEnum()` returns the chosen value; `--help` shows `<a|b|c>`. The CLI-edge analogue of the engine's "enum, not string-match" rule.
- **canvas_stress** migrated off its hand-rolled `strcmp` parse loop to declarative `IRArgs` registration (`registerArgs` before init, `applyArgs` after) — the engine's strict parse was rejecting the demo's own custom flags. Addresses **#2103** for `canvas_stress` (other migrated demos still need the same treatment, so not a full close). Also adds a `--frozen-pose` diagnostic (freeze GRID spin cubes at a fixed off-cardinal pose for camera-direction isolation) and folds the two `--depth-probe` X,Y parse copies into a local `parsePixelArg` helper.
- Unit tests: `test/common/ir_args_test.cpp` (7 cases) cover `numbers()` (arity, omitted→empty, negatives, coexistence) and `enumValue()` (allowed, default, inline `=`).

## Test plan
- [x] `IrredenEngineTest --gtest_filter='IRArgs*'` — 7/7 pass
- [x] `IRCanvasStress --help` lists `--sweep-yaw <float float float>`, `--debug-overlay <none|ao|…>`, etc.; strict parse accepts all demo flags
- [x] Builds clean: IRCanvasStress + IrredenEngineTest
- [x] Flagless `IRCanvasStress` byte-identical (default path unchanged)

> ⚠️ #2103 is claimed by another worker (mac-worker-3) — this overlaps for `canvas_stress`; reconcile the claim.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2157, opened 2026-06-30
render: rotated GRID/re-voxelize cubes — residual non-cardinal gaps + dark riser shading after the silhouette-riser fix


jakildev:
**Model:** fable
**Blocked by:** #2162 (the silhouette-riser gap fix — this touches the same `c_voxel_to_trixel_stage_*` shaders; base on its post-merge state). Unblock when #2162 merges.

**Follow-up to the silhouette-riser face-emission fix** (branch \`claude/diagnose-gridspin-gaps\`, pending PR) that closed the dominant rotated-voxel gap: the visible-face triplet emits one polarity per axis, but a rotated voxel staircase's grazing edge presents the *opposite* polarity, so its exposed silhouette risers were dropped → see-through black gaps, worst where yaw chirality turns that edge camera-ward (90°/270°). The fix flips to the exposed opposite-polarity face, gated to rotated content (\`reVoxelize\` uniform OR per-voxel \`VoxelReserved::kRotatedEmit\`). Verified on \`canvas_stress --only gridspin --frozen-pose 0.6 --sweep-yaw 0 6.2832 17\`: worst cardinal 270° **7.98% → 0.29%**, axis-aligned byte-identical (drift 0) at all directions.

Two residuals remain, each a **different, pre-existing mechanism** — tracking them here so they aren't lost:

## 1. Non-cardinal residual gaps (~1%) — per-axis scatter coverage (#1933)
At **non-cardinal camera yaw** the rotated GRID cubes still show ~0.4–1.2% gap (worst observed 1.22% at 112.5°). These directions engage the per-axis forward-scatter route, whose sub-pixel coverage is the #1933 corner-spike/dashing family (partially addressed by #2013 analytic coverage). The silhouette-riser fix is a face-*selection* fix and does not touch scatter *coverage*. Cardinal directions are clean (≤0.29%).

- Repro: \`IRCanvasStress --only gridspin --frozen-pose 0.6 --sweep-yaw 0 6.2832 17 --auto-screenshot 30\`, measure enclosed-black per shot; the spikes sit at the non-cardinal (per-axis) indices.
- Likely folds into **#1933** / the per-axis scatter coverage work.

## 2. Dark staircase-riser shading — AO on the now-filled risers (#2089 / #1718)
[...]

Issue #2125, opened 2026-06-29
render: fog cross-section — Mode A voxel-resolution on the GRID canvas (P1)


jakildev:
**Model:** opus
**Part of epic:** #2124
**Plan file:** `.fleet/plans/issue-2124.md` (full epic plan)
**Blocked by:** (none)

## Scope
Give voxel **objects** a filled cross-section at the grid fog boundary on the GRID (world) canvas, voxel-resolution (Mode A). Today the boundary either drops boundary columns (#2106 — see-through holes / missing faces) or hard-blacks them (pre-#2106). This phase adds **cut faces**: a revealed voxel emits its interior side face toward a fog-hidden solid neighbor, so the object is capped with a real interior wall instead of a hole. Absorbs #2106's per-voxel column clip as the foundation (close #2106 as superseded).

## Approach
- In `c_voxel_to_trixel_stage_1.glsl` + `c_voxel_to_trixel_stage_2.glsl` (and both Metal counterparts), the per-slot face gate is `if (!faceIsExposed(flagsByte, faceId)) return;`. Replace with: keep the face if `faceIsExposed(...)` **OR** it is a cut face — `faceId` is vertical (∈ {X_NEG,X_POS,Y_NEG,Y_POS}) AND its neighbor column (`voxelPosRaw.xy + faceNormalXY(faceId)`) is fog-hidden.
- "fog-hidden" = #2106's `fogVoxelColumnHidden` test (grid cell unexplored AND outside every live vision circle), evaluated at the **neighbor** cell. The fog image (binding 0) + observers UBO (binding 27) are already bound in STAGE_1 by #2055/#2106; STAGE_2 needs the same two bound (it already binds the color image on slot 0 — sequence the fog read before/with it, or bind fog on a free slot — confirm during impl which is free in STAGE_2).
- Add `faceNormalXY(int faceId) -> ivec2` to `ir_iso_common.glsl`/`.metal` (or reuse an existing face-normal helper if present).
- Gate the whole cut-face path on `visionCircleCount > 0 && perAxisRoute == 0 && isDetachedCanvas < 0.5` so non-fog / per-axis / detached scenes are byte-identical (those phases come later).
- Keep #2106's own-column drop (`fogVoxelColumnHidden(self)` -> return) so the hidden half still doesn't rasterize; the cut faces cap the revealed half.
- fog_demo: add a scene with a tall voxel object straddling the vision boundary so the cross-section is visible; commit render-verify references.

[...]

PR #2162, opened 2026-06-30
render: silhouette-riser face emission for rotated voxel cubes

jakildev:
## Summary
Rotated GRID / re-voxelize voxel cubes showed see-through black gaps at certain camera directions. Root cause: the visible-face triplet (#1278) emits one polarity per axis — fine for a convex axis-aligned solid — but a rotated footprint round-to-cells into a **staircase** whose camera-side grazing edge presents the *opposite* polarity of an axis (e.g. a +X riser where the cardinal triplet carries X_NEG). That exposed silhouette riser was absent from the triplet → never emitted → gaps, worst where the yaw chirality turns that edge camera-ward (90°/270°).

**Fix:** in `c_voxel_to_trixel_stage_{1,2}` (GLSL + Metal) flip a slot's face to the exposed opposite-polarity same-axis face when the triplet face is occluded — **gated to rotated content** (the `visibleFaceIds.w` re-voxelize uniform OR the per-voxel `VoxelReserved::kRotatedEmit` marker, reserved bit 2, stamped by `REBUILD_GRID_VOXELS`). Non-rotated content keeps the strict triplet, so the single-canvas **and** per-axis camera-yaw fast paths stay byte-identical. Shared by the GRID and detached re-voxelize paths.

## Visual (canvas_stress `--only gridspin`, camera 270°, worst direction)
| before (gaps) | after (solid) |
|---|---|
| before | after |

Solid across a full spin sweep (8 poses):

pose sweep

## Test plan
- [x] `IRCanvasStress --only gridspin --frozen-pose 0.6 --sweep-yaw 0 6.2832 17`: worst gap **7.98% → 0.29%** at 270°; cardinals ≤0.29%
- [x] Axis-aligned (`--no-spin`) **byte-identical** (img_diff drift 0) at all 17 camera directions, single-canvas + per-axis routes
- [x] Detached re-voxelize cubes (`--only canary`) solid via the shared flip
- [x] Builds clean: IRCanvasStress + IrredenEngineTest (Metal)
- [ ] GL parity smoke (authored on macOS/Metal; GLSL mirrors Metal — needs a Linux host)

Stacked on: #2161 (IRArgs + canvas_stress migration — provides the `--frozen-pose` harness this PR verifies with).
Residuals (non-cardinal per-axis scatter coverage, riser AO shading) tracked in #2157.


jakildev:
Holding review: upstream PR #2161 is not yet approved. This stacked PR will be re-evaluated once the upstream lands an approval label.

jakildev:
#2161. Its head ref `claude/irargs-multivalue-enum` was force-
commits.

  git fetch origin
  git rebase origin/claude/irargs-multivalue-enum
  # resolve conflicts, then:
  git push --force-with-lease

after a manual rebase.

— fleet merger

jakildev (commented):
## Review — render: silhouette-riser face emission for rotated voxel cubes

Stacked on #2161 (cross-author: jakildev; upstream is `fleet:approved` and
still open — reviewing this PR's own diff only). Note for the human: this PR
currently shows `mergeable: CONFLICTING` and carries `fleet:needs-base-update`
— the merger couldn't cascade-rebase after #2161's branch was force-pushed.
**CONFLICTING state checked per the churn-audit step:** the PR's own diff
(`fleet-pr diff 2162`, which correctly excludes the inherited upstream commit)
touches only voxel/render files — `component_voxel.hpp`,
`system_rebuild_grid_voxels.hpp`, the two `c_voxel_to_trixel_stage_{1,2}`
GLSL+Metal pairs, and two `CLAUDE.md` docs. None of these overlap with the
files `gh pr view` shows conflicting (`canvas_stress/main.cpp`,
`engine/ir_args.*` — all inherited from #2161's own commit), so the fix's
content should carry over unchanged once the author/opus-worker manually
rebases per the merger's instructions. No out-of-scope files or oversized
churn in the child's own diff.

**Verdict:** approve

### Review notes
Traced the face-flip logic end to end:
- `kFaceXNeg=0/kFaceXPos=1/kFaceYNeg=2/kFaceYPos=3/kFaceZNeg=4/kFaceZPos=5`
  (`ir_iso_common.glsl`) — confirmed `faceId ^ 1` always toggles within the
  same axis pair (0↔1, 2↔3, 4↔5), never crosses axes.
- `rotatedEmit` gate (`reVoxelize || (reserved & 4u) != 0u`) is scoped
  correctly: `system_rebuild_grid_voxels.hpp`'s `identityArm` (byte-identical
  path for a set that never rotated) never touches `reserved_`; only
  `forwardArm` and `inverseArm`'s `writeCell` — both reached only when
  `!isIdentityTransform` — stamp `VoxelReserved::kRotatedEmit`. The
  `inverseArm` stamp uses `out.reserved_ |=` (not an overwrite), so a
  voxel's existing per-trixel-priority bits (bits[1:0], #1960) survive
  alongside the new bit 2 — no collision with that separate encoding.
- Stage 1 / Stage 2 (GLSL + both Metal twins) apply the identical
  `rotatedEmit` condition and `faceId ^ 1` flip in lockstep, and `reVoxelize`
  is already in scope before the new block in all four files — no
  ordering hazard, and the "stage 2 MUST mirror stage 1 exactly" invariant
  (color tap vs. distance tap) holds.
- The byte-identity claim for non-rotated content checks out: gating on
  `rotatedEmit` means a static/concave axis-aligned voxel never reaches the
  flip, so the existing exposed-face resolution for non-rotated scenes is
  untouched (confirmed by the PR's own 17-camera-direction byte-identical
  sweep).

### Nits
- `voxels[voxelIndex].reserved & 4u` repeats a magic literal for
  `VoxelReserved::kRotatedEmit` across 3 shader files instead of a named
  GLSL/Metal constant — consistent with the existing `0x3u` priority-mask
  precedent in the same shaders, so not a new pattern, just noting it in
  case a shared shader-side constants header exists to reduce the literal
  duplication.

- [ ] Resolve `fleet:needs-base-update` (manual rebase onto #2161's current
      tip) before this can merge
- [ ] GL parity smoke (already tracked via `fleet:needs-linux-smoke`)

Escalation: please Opus-recheck before merge (touches: render pipeline
per-voxel face-selection in the hot voxel→trixel compute path, plus a new
`reserved_` bit alongside the existing per-trixel-priority encoding).

jakildev (commented):
## Opus recheck — render: silhouette-riser face emission for rotated voxel cubes (#2162)

**Verdict:** approve

Stacked on #2161 (cross-author: jakildev; upstream is `fleet:approved` and
still open) — reviewed this PR's own diff only via `fleet-pr diff 2162`, which
excludes the inherited upstream commit. Sonnet approved and escalated for the
hot voxel→trixel compute-path recheck plus the new `reserved_` bit alongside
the per-trixel-priority encoding. I focused this pass on the things a first
read can't fully close: the face-flip's emit semantics, the CPU-side marker
lifecycle three arms deep, and the load-bearing byte-identity invariant. All
hold.

### The flip is a replacement, not a duplicate emit
The block runs *before* the exposed-face `keepFace` gate and only fires on
`!faceIsExposed(faceId) && faceIsExposed(faceId ^ 1)` — i.e. exactly the slots
that would otherwise `return` at the gate (triplet face occluded). It swaps to
the now-exposed opposite face, which then passes `keepFace` and fills the
grazing-edge gap. Net emit count per slot is unchanged; there is no path where
both the triplet face and its opposite emit. Sonnet flagged `faceId ^ 1` as an
axis-pair toggle; on closer read I confirm it stays within-axis (0↔1/2↔3/4↔5)
and, since the three triplet slots are one-per-axis, no two slots can ever flip
onto the same face — no self-collision.

### Byte-identity survives a rotate→identity round-trip (the invariant Sonnet couldn't fully close)
The gate makes never-rotated content trivially byte-identical (`rotatedEmit`
false → no flip). The non-obvious case is a set that rotated and then returns
to yaw 0: its cells carry `kRotatedEmit` from the inverse/forward arm. Traced
the clear path — `identityArm` restores the authored span with a **full-record
assignment** (`poolColors[baseIdx+i] = rotationSourceVoxels_[i]`), and that
snapshot is the pre-rotation authored state, which never carries the bit. So
the marker is cleared on the first identity frame after rotation and the
steady-state identity frames are byte-identical to master. The convex-content
safety is also self-reinforcing: on a convex axis-aligned solid each visible
triplet face is itself exposed, so `!faceIsExposed(faceId)` is false and the
flip can't fire even if a bit ever leaked — the flip only ever activates on the
concave/staircase geometry it's meant for.

### Marker placement + priority-bit coexistence
- The forward-arm marker loop lives inside `forwardArm()` (the rotated
  zero-scale / oversized-dest fallback), never the identity path, and only
  stamps opaque cells (`color_.alpha_ != 0`). The inverse arm stamps in
  `writeCell`. Both use `reserved_ |=`, so the separate per-trixel-priority
  carrier (bits[1:0], #1960/#2023) rides alongside bit 2 with no collision —
  confirmed against stage 2's `voxels[voxelIndex].reserved & 0x3u` priority
  read, which is untouched.
- Stage 1 / stage 2 / both Metal twins apply the identical `rotatedEmit`
  predicate and `faceId ^ 1` flip, each placed before its own exposed-face
  gate — the "stage 2 MUST mirror stage 1 exactly" distance/colour-tap
  invariant holds. The flip also sits ahead of, and is orthogonal to, the fog
  CUT-FACE widening (#2125): that path fires only on non-exposed faces at a fog
  boundary, whereas the flip requires the opposite face to be exposed.

### Agreement with Sonnet
Sonnet's trace of the `identityArm` no-touch, the `inverseArm` `|=` stamp, and
the 17-direction byte-identical sweep all check out. The magic-literal nit
(`& 4u` for `kRotatedEmit` across the 3 shaders vs a named GLSL/Metal constant)
is real but consistent with the existing `0x3u` priority-mask precedent in the
same files — a hardening nit, not a merge gate.

### Pre-merge (mechanical, not a code concern)
- `fleet:needs-base-update` / `mergeable: CONFLICTING` — the merger couldn't
  cascade-rebase after #2161 was force-pushed. Needs a manual rebase onto
  #2161's tip before merge; the conflicting files (`canvas_stress/main.cpp`,
  `ir_args.*`) are all inherited from #2161 and don't overlap this PR's own
  diff, so the fix content carries over unchanged.
- GL parity smoke still pending (`fleet:needs-linux-smoke`).

🤖 Opus recheck by Claude (role-opus-reviewer)

jakildev:
Addressed the `fleet:has-nits` feedback (magic-literal nit on `& 4u`).

Investigated adding a named GLSL/Metal constant for the `VoxelReserved::kRotatedEmit`
bit (and the `0x3u` priority mask, for consistency) alongside the existing
`ir_constants.{glsl,metal}` shared-header convention (used for e.g. `kInvalidDepth`,
and `kFaceXNeg`/etc. in `ir_iso_common`).

Checked the closer precedent instead: `IRComponents::VoxelFlags::kFaceOccluded*` is
also a bitmask mirrored from a C++ `reserved`/flags word, and it is *not* given a
shader-side named constant anywhere in the shader tree — both GLSL and Metal keep it
as inline bit arithmetic (`(2 + faceId)` etc.) with a comment citing the C++ name.
`ir_constants.glsl`'s existing named constants (`kInvalidDepth`, `kFaceXNeg`/etc.) are
all semantic IDs compared by equality, not bitmask flags. This PR's `& 4u` / `& 0x3u`
already follow the bitmask-mirror convention (inline literal + comment citing the
C++ name), matching `kFaceOccluded*`'s pattern — introducing a named constant here
would deviate from precedent rather than align with it.

Leaving the nit un-actioned as a deliberate judgment call, not an oversight.
`fleet:approved` stays valid. Not touching `fleet:needs-base-update` — that's the
opus+/human manual-rebase lane per the merger's comment. — worker

jakildev:
Merger: base PR #2161 merged; re-targeted this PR to `master` but the rebase onto current master conflicts. The child commits cannot be replayed cleanly — worker will resolve semantically.

Conflicted files:
- `creations/demos/canvas_stress/main.cpp` — master: `29d8105f render: canvas_stress per-ENTITY priority-swap orbit demo (#2154) (#2177)`; PR: `c0297339 engine: IRArgs multi-value + enum flags; canvas_stress declarative migration`
- `engine/include/irreden/ir_args.hpp` — master: `279c9903 engine: IRArgs validated-enum flag type + canvas_stress --debug-overlay/--frozen-pose (#2161) (#2161)`; PR: `c0297339 engine: IRArgs multi-value + enum flags; canvas_stress declarative migration`
- `engine/ir_args.cpp` — master: `279c9903 engine: IRArgs validated-enum flag type + canvas_stress --debug-overlay/--frozen-pose (#2161) (#2161)`; PR: `c0297339 engine: IRArgs multi-value + enum flags; canvas_stress declarative migration`
- `test/common/ir_args_test.cpp` — master: `279c9903 engine: IRArgs validated-enum flag type + canvas_stress --debug-overlay/--frozen-pose (#2161) (#2161)`; PR: `c0297339 engine: IRArgs multi-value + enum flags; canvas_stress declarative migration`

— fleet merger

jakildev:
Resolved semantic conflict after base PR #2161 merged to master. Two-part resolution:

1. **Dropped the inherited #2161 prefix.** The branch carried #2161's IRArgs/canvas_stress commit (`c0297339`) as a stacked prefix; that work is now on master independently (`279c9903`/`29d8105f`), so the 4 IRArgs conflicts (`ir_args.hpp/.cpp`, `ir_args_test.cpp`, `canvas_stress/main.cpp`) were stale inherited copies. Rebased with `--onto origin/master <fork-point>` to replay only #2162's own silhouette-riser commit — the branch is now a single clean commit (11 files, +120/-6, identical to the original #2162 diff).

2. **Reconciled the genuine shader conflicts.** #2162's silhouette-riser face-flip (rotated-content `faceId ^ 1`) overlapped master's fog cut-face gate evolution (#2125/#2127/#2128 — `fogActive`/`worldColumn`). Preserved BOTH: the riser flip finalizes `faceId` first, then master's fog machinery decides `keepFace`. Applied identically across all four shaders (GLSL + Metal, stage 1 + 2); dropped #2162's now-superseded older copy of the exposed-gate comment in favor of master's current one. Interaction is clean — a flipped riser face is exposed, so master's `!keepFace` fog-cut branch never fires for it.

Verified: `fleet-build --target IRShapeDebug` clean + runtime **Metal** shader compile confirmed (macOS/Metal host, PR authored-on-macos) + 21-shot `--auto-screenshot` run exit 0. Reviewer please re-evaluate the rebased diff. `fleet:needs-linux-smoke` left in place for Linux/GL validation. — worker

jakildev (commented):
## Review — render: silhouette-riser face emission for rotated voxel cubes (#2162)

**Verdict:** approve

Re-review after semantic-conflict resolution. This PR was previously Sonnet-approved and Opus-approved on the original diff (submitted 2026-07-01, stacked on #2161). Since then: #2161 merged, the merger's cascade-rebase hit a semantic conflict (stale inherited IRArgs/canvas_stress commit + a genuine shader conflict against master's evolved fog cut-face gate), and the author-agent resolved it by rebasing `--onto` to drop the inherited #2161 prefix and manually reconciling the shader overlap. `baseRefName` is now `master`, `mergeable: MERGEABLE` — no longer stacked.

### Prior-review resolution
- ✅ CPU-side logic (`component_voxel.hpp` `VoxelReserved::kRotatedEmit`, `system_rebuild_grid_voxels.hpp` forward/inverse-arm stamping) — diffed byte-for-byte against the originally-reviewed content; unchanged. `identityArm` still untouched; `forwardArm`'s stamping loop (new lines 305–313) and `inverseArm`'s `writeCell` `|=` stamp are exactly what Sonnet + Opus traced before.
- ✅ `voxels[voxelIndex].reserved & 4u` magic-literal nit — not re-flagged. Author's 2026-07-01 comment investigated a named-constant alternative and found the closer precedent (`VoxelFlags::kFaceOccluded*`) also stays inline-bitmask-with-comment; deliberate judgment call, consistent with the fleet's "borderline nit, author addressed with justification" bar.
- **New since last review — the semantic-conflict reconciliation itself**: verified independently rather than trusting the resolution comment. Diffed all four shader files (`c_voxel_to_trixel_stage_{1,2}.glsl` + `.metal` twins) against current `origin/master`. In each file the `rotatedEmit`/`faceId ^ 1` flip block sits *before* `keepFace = faceIsExposed(flagsByte, faceId)` and before `worldColumn`/`fogActive` fog-cut-face computation (#2125/#2127/#2128, which evolved on master after this branch was cut). `faceId` was changed from `const int` to `int` in all four files so the flip can mutate it ahead of the fog gate reads — traced that both `keepFace` and the `faceId < kFaceZNeg && fogActive` cut check downstream correctly observe the *flipped* value. Stage 1/Stage 2 and GLSL/Metal apply the identical predicate and ordering — the "stage 2 must mirror stage 1 exactly" invariant holds against the new fog code too, not just the code that existed at the time of the original Opus pass.

- [x] Rebase reconciliation verified structurally correct (this review)
- [ ] GL parity smoke still pending (`fleet:needs-linux-smoke`, unchanged from before)

Escalation: please Opus-recheck before merge (touches: render pipeline per-voxel face-selection in the hot voxel→trixel compute path; the specific new surface is the composition between this PR's face-flip and master's newer fog cut-face gate code, which is a different code shape than what the original Opus pass reviewed).

Opus recheck required: rotated-footprint face-flip composed with master's evolved fog cut-face gate in a hot compute shader path — reconciliation done by an autonomous rebase, worth a second pass even though the flip logic itself was already Opus-approved pre-conflict.

jakildev (commented):
## Opus recheck — render: silhouette-riser face emission for rotated voxel cubes (#2162)

**Verdict:** approve

Second Opus pass, post-conflict-resolution. This PR was Opus-approved on
2026-07-01 while stacked on #2161; since then #2161 merged, the cascade-rebase
hit a semantic conflict, and the author-agent rebased `--onto` to drop the
inherited prefix and manually reconciled the shader overlap against master's
now-evolved fog cut-face gate (#2125/#2126/#2127/#2128). `baseRefName` is
`master`, `mergeable: MERGEABLE` — no longer stacked. Sonnet re-escalated
because the specific new surface is the *composition* of this PR's face-flip
with master's fog code, a different code shape than the original pass reviewed.
I focused this recheck there.

### Sonnet flagged the flip↔fog composition as the recheck surface — on closer read it is clean
Sonnet's structural diff (CPU logic byte-identical to the pre-conflict approval;
flip placed before the fog gate in both stages) checks out. I traced the actual
ordering and interaction three ways:

1. **Flip runs strictly before the `keepFace` gate and only substitutes an
   EXPOSED opposite-polarity face for an OCCLUDED triplet face on rotated
   content.** So after a flip, `keepFace = faceIsExposed(faceId)` is
   unconditionally true → the fog cut-face branch (`!keepFace && faceId <
   kFaceZNeg && fogActive`) is short-circuited and never runs on a flipped face.
   The flip and the cut-wall rule are therefore disjoint by construction: the
   flip yields exposed faces, the cut rule yields kept *non*-exposed vertical
   faces. `isCutFace` (stage 2 / Metal stage 2) correctly stays false on a
   flipped face, so LIGHTING_TO_TRIXEL won't mis-force-light a real exposed
   riser as a cross-section wall.

2. **The flip cannot move a face across the per-axis store filter.** `faceId ^
   1` toggles only bit 0 (polarity); `(faceId >> 1)` (the axis the
   `(faceId>>1)!=axis` store filter and the route selection use) is invariant
   under the flip. So a flipped face stays on the same axis canvas — no route
   leakage, and the per-axis `reveal<=0` own-column clip (keyed on
   `voxelPosition`, not `faceId`) is unaffected.

3. **Non-rotated / non-fog byte-identity holds.** `rotatedEmit = reVoxelize ||
   (reserved & 4u)` gates the flip; static content never flips, so the entire
   pre-existing exposed-face + fog-cut path is untouched (consistent with the
   17-direction byte-identical sweep). The `kRotatedEmit` bit (bit 2) is
   disjoint from the priority mask (`reserved & 0x3u`, bits 0–1), so the flip
   gate and the per-trixel-priority encode don't alias.

### Cross-backend parity confirmed
All four shaders — `c_voxel_to_trixel_stage_{1,2}.glsl` and their `.metal` twins
— carry the identical flip block (`rotatedEmit && !faceIsExposed(faceId) &&
faceIsExposed(faceId^1)`), each with `reVoxelize` defined before it and the
block placed after `flagsByte` and before the fog `keepFace` gate. The "stage 2
MUST mirror stage 1's flip + gate exactly" invariant that keeps the colour tap
aligned with the distance tap holds on both backends.

### Agreement with Sonnet
Sonnet's re-review verdict (rebase reconciliation structurally correct; CPU
`kRotatedEmit` stamping in `system_rebuild_grid_voxels.hpp` unchanged from the
approved diff) matches my read. The `& 4u` magic-literal nit (a named
GLSL/Metal `kRotatedEmit` constant vs the raw literal across the 3 shaders) is
real but consistent with the existing `0x3u` priority-mask precedent in the same
files — a hardening nit, not a merge gate.

### Pre-merge (mechanical, not a code concern)
- GL parity smoke still pending (`fleet:needs-linux-smoke`) — the flip is a hot
  voxel→trixel compute-path change validated on Metal; a Linux/GL smoke closes
  the loop before merge.

🤖 Opus recheck by Claude (role-opus-reviewer)

Issue #2164, opened 2026-07-01
engine/voxel: encapsulate C_VoxelSetNew derived-state maintenance (kill raw-span-carve footgun)


jakildev:
**Area:** engine/voxel (+ creations sweep, coding conventions)

**Model:** opus (umbrella; children carry their own model tags)

## Context

PR #2146 was the **fourth** landing of the same one-line fix: a creation
carves voxels through the raw `voxels_` span of `C_VoxelSetNew`, calls
`syncActiveMask()`, but forgets `IRPrefab::Voxel::recomputeFaceOccupancy()` —
leaving the exposed-face mask stale so carved surfaces render black/missing
under the lit path. The trail: review nit on #2016 → #2018 → PR #2117
(`shape_debug`) → #2119 → PR #2146 (`z_yaw_rotation`, `ir_voxel_yaw`). The
convention is already documented in `engine/prefabs/irreden/voxel/CLAUDE.md`
and a `simplify-check-ecs` heuristic already exists for it — yet it keeps
recurring. This is a **missing-abstraction footgun**: a subsystem dependency
is pushed onto every creation as a manual, opt-in, easy-to-drop step instead
of being owned by the subsystem.

Investigation sharpened the case. The bulk mutators on `C_VoxelSetNew`
(`deactivateAll`, `activateAll`, `fillPlane`, `reshape`, `changeVoxelColor*`)
each maintain **three** invariants on every edit: (1) `mirrorToRotationSource(i)`
— the GRID-spin authored-truth snapshot; (2) the pool **active mask**; (3)
**face occupancy**. The documented manual carve recipe mentions only **two** —
it omits the rotation-source mirror entirely. So every hand-rolled carve site
isn't just at risk of the black-face bug, it also silently skips the mirror.
The only sane place to keep all three in lockstep is an **encapsulating
mutator on the component**. This is the engine's own rule (`cpp-ecs.md`
§"No dirty flags" → *"push at mutation time"*); a system-tick recompute is the
wrong tool (forbidden changed-flag + wasted O(volume) on static sets).

## Goal

Add an encapsulating edit API to `C_VoxelSetNew`, migrate every creation carve
site onto it, retire the manual recipe to an internal mechanism, and codify
the general convention (systems own their dependencies via an attached API or
pre/post-tick hooks — never delegated to callers) so the next such footgun is
caught in review.

## Approach (phased)

[...]

Issue #2018, opened 2026-06-25
shape_debug: createVoxelPoolShape misses recomputeFaceOccupancy after SDF carving — carved --spin-shape-voxel renders black


jakildev:
**Area:** creations/demos/shape_debug

**Model:** sonnet

**Blocked by:** (none)

## Context

Surfaced as a review nit on PR #2016 (which added `createCustomVoxelFigure` and fixed *its* exposed-face mask). The sibling builder `createVoxelPoolShape` (`creations/demos/shape_debug/main.cpp:841`) has the same latent bug it did not fix.

`createVoxelPoolShape` carves an SDF primitive by calling `vs.voxels_[i].deactivate()` per interior voxel (line 857), then calls `vs.syncActiveMask()` (line 881) — but it never calls `IRPrefab::Voxel::recomputeFaceOccupancy`. The exposed-face mask therefore stays pinned to the ctor's all-active grid, so the newly-exposed surface faces of a carved set stay wrongly occluded and never emit. `createCustomVoxelFigure` (line 902) does the right thing — `recomputeFaceOccupancy(vs.voxels_, size)` then `syncActiveMask()` (lines 946-947).

Symptom: `fleet-run IRShapeDebug --spin-shape sphere --spin-shape-voxel` (any carved SDF shape voxelized through this path) renders black / missing faces, because the per-voxel `--spin-shape-voxel` path routes through `createVoxelPoolShape` rather than the static GRID/rotating-set per-frame mask rebuild.

## Acceptance criteria

- `createVoxelPoolShape` calls `IRPrefab::Voxel::recomputeFaceOccupancy(vs.voxels_, size)` before `vs.syncActiveMask()`, mirroring `createCustomVoxelFigure`.
- `fleet-run IRShapeDebug --spin-shape sphere --spin-shape-voxel --auto-screenshot` renders a solid carved sphere with crisp surface faces (not black).
- Default scene (no flags) stays byte-identical to committed render-verify references.

## Notes

Flagged as a known follow-up in PR #2016's body and review. Trivial one-line fix; filed so it doesn't slip.

jakildev:
[...]

PR #2170, opened 2026-07-01
engine/voxel: encapsulating edit API on C_VoxelSetNew + skeletal_demo fix

### LINKED ISSUE #2165: engine/voxel: encapsulating edit API on C_VoxelSetNew + skeletal_demo fix (P1)
**Model:** opus
**Effort:** high
**Part of epic:** #2164
**Plan file:** `~/.fleet/plans/issue-2164.md` (full epic plan)
**Blocked by:** (none)

## Scope

Add an encapsulating edit API to `C_VoxelSetNew` so custom voxel carves/edits
keep all derived state in sync automatically, and fix `skeletal_demo` (the one
live black-face bug) as the first consumer. This is the head ticket — P2 (the
creations sweep) and P3 (convention docs) both build on the API landed here.

## Approach

- In `engine/prefabs/irreden/voxel/components/component_voxel_set.hpp`, add
  three public members near the existing mutators (after `reshape`, before
  `syncActiveMask` at ~line 498). No new includes — the existing mutators
  already pull in `face_occupancy.hpp` and `voxel_pool_api.hpp`.
  - `template <typename Fn> void editVoxels(Fn &&fn)` — loops `i` over
    `numVoxels_`, calls `fn(i, voxels_[i], positions_[i].pos_)` (position lets
    SDF/surface/bone carves read local coords), then runs the shared resync.
  - `template <typename Fn> void carve(Fn &&shouldDeactivate)` — sugar:
    `editVoxels([&](int, C_Voxel &v, vec3 p){ if (shouldDeactivate(p)) v.deactivate(); })`.
  - `void resyncAfterRawEdits()` — escape hatch for multi-pass edits.
- Factor the shared tail into a private `resyncDerivedState()` that all three
  call, so the invariant order lives in exactly one spot: per-voxel
  `mirrorToRotationSource(i)` → `VoxelPool::resyncRangeFromColors(voxelStartIdx_,
  numVoxels_, canvasEntity_)` → `IRPrefab::Voxel::recomputeFaceOccupancy(voxels_, size_)`.
- Migrate `creations/demos/skeletal_demo/main.cpp` (~L390-405) to `editVoxels`
  (it assigns `bone_id_`/color then deactivates) — this restores the currently
  missing `recomputeFaceOccupancy` and fixes the black render.
- Update the `syncActiveMask()` comment to point authors at `editVoxels`/`carve`.

## Acceptance criteria

- `editVoxels` / `carve` / `resyncAfterRawEdits` compile and are `git grep`-able
  on `C_VoxelSetNew`; each restores mirror + active mask + face occupancy.
- `fleet-run IRSkeletalDemo --auto-screenshot` renders solid carved surfaces
  (not black).
- Engine builds clean across voxel-consuming targets; no behavior change to any
  site other than `skeletal_demo`.

## Dependencies

- Blocks P2 (creations sweep) and P3 (convention docs) — both reference the new
  API by name. This ticket lands first.
- Assumes nothing from siblings.

## Gotchas

- The three existing mutators (`deactivateAll`/`fillPlane`/`reshape`,
  `component_voxel_set.hpp:349-438`) are the reference for the exact resync
  sequence — mirror the ordering and comment density; don't invent a new order.
- `positions_` is valid at creation time (pre-canvas-migration); carves run
  there, so passing `positions_[i].pos_` into the callback is safe.
- Keep `syncActiveMask()` public — it's still the low-level pool primitive used
  internally; do not delete it.

## References

- Umbrella #2164; PRs #2117, #2146 (prior one-line fixes); `cpp-ecs.md`
  §"No dirty flags" (the push-at-mutation rule this API implements).
- `engine/prefabs/irreden/voxel/face_occupancy.hpp:52` (`recomputeFaceOccupancy`).

jakildev:
## Summary
- Add `editVoxels` / `carve` / `resyncAfterRawEdits` to `C_VoxelSetNew` — the supported way to do a custom voxel carve/edit the bulk mutators don't cover. Each restores every derived invariant (rotation-source mirror → pool active-mask → face occupancy) once, via a private `resyncDerivedState()` that owns the ordering in one place.
- Callers no longer hand-roll `syncActiveMask()` + `recomputeFaceOccupancy()` — dropping the recompute is the recurring #2018/#2117/#2146 black-carve footgun. `syncActiveMask()` stays public as the low-level pool primitive.
- Migrate `skeletal_demo`'s three raw `voxels_` loops to `editVoxels`. The **cross** avatar rendered black because its deactivate-carve called only `syncActiveMask()` with no face-occupancy recompute; it now renders solid. The snake/lamp full boxes are byte-identical (their resync is a no-op on an all-solid set).

Head ticket of epic #2164 — P2 (creations sweep, e.g. `carveSphere`/`carveOrbitShape`) and P3 (convention docs, incl. the `voxel/CLAUDE.md` C_VoxelSetNew bullet) build on the API landed here.

## Test plan
- [x] `fleet-build --target IRSkeletalDemo` clean (macOS/Metal), header + demo compile.
- [x] `fleet-run IRSkeletalDemo --auto-screenshot` — carved cross renders solid, not black (see before/after below).
- [x] Header change is purely additive (new methods only; `C_VoxelSetNew` layout unchanged), so voxel-consuming demos (shape_debug etc.) are byte-identical by construction.
- [x] `format-changed` clean; `simplify` clean.

## Screenshots

<details>
<summary>zoom4 — the carved cross (top-left): black/gappy before, solid after</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_yaw90</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom1 / zoom2</summary>

| Before (zoom1) | After (zoom1) |
|--------|-------|
|  |  |
| Before (zoom2) | After (zoom2) |
|  |  |

</details>

Closes #2165


jakildev (commented):
## Review — engine/voxel: encapsulating edit API on C_VoxelSetNew + skeletal_demo fix

**Verdict:** needs-fix

### Verification performed

- `fleet-build --target IRSkeletalDemo` — clean build on macOS/Metal, reproduced locally.
- Diff is exactly the demo migration + the new header API (81/27 +/-); no
  other files touched besides the PR's own before/after screenshots.
- Cross-referenced `engine/prefabs/irreden/voxel/CLAUDE.md`'s documented
  resync order ("rotation-source mirror -> pool active-mask -> face
  occupancy") against `resyncDerivedState()`
  (`component_voxel_set.hpp:555-561`) and every existing bulk mutator
  (`deactivateAll`, `activateAll`, `fillPlane`, `reshape`) — the new private
  helper reproduces the exact same three-step order every one of them
  already runs inline. No drift.
- `createCross`'s old loop ended with a bare `vs.syncActiveMask()` and no
  face-occupancy recompute — textbook instance of the #2018/#2117/#2146
  footgun the module doc warns about. `editVoxels` routes through
  `resyncDerivedState()`, which does call `recomputeFaceOccupancy`, so this
  genuinely fixes the described black-cross bug rather than just refactoring
  around it.
- `createSnake`/`createLamp` never mutated activation state before, so the
  new unconditional resync-on-every-call is extra (harmless, one-time,
  scene-construction-only) work, not a behavior change — consistent with the
  PR body's "byte-identical" claim.
- Type-checked the callback signature: `positions_[i].pos_` is
  `IRRender::VoxelGpuPosition::pos_` (`IRMath::vec3`,
  `voxel_pool_allocation.hpp:30`), matching the `vec3 localPos` parameter
  used at both demo call sites.

### Needs-fix

- No test coverage for the new public API (`editVoxels` / `carve` /
  `resyncAfterRawEdits`, `component_voxel_set.hpp:451-472`). This is a
  non-trivial invariant (three ordered resync steps) that is now the head of
  epic #2164 — other creations' carve call sites (`carveSphere`,
  `carveOrbitShape`, per the PR body's own P2 follow-up) will build on it
  next. `test/ecs/voxel_set_target_canvas_test.cpp` already shows this
  exact component can be exercised headlessly (entity-keyed pool, no
  RenderManager/canvas needed) — the same fixture would support asserting,
  e.g., that a `carve()` call correctly recomputes face-occupancy bits and
  keeps the pool active-mask in sync for a mixed active/inactive result.
  Worth calling out specifically: `carve()` (`component_voxel_set.hpp:460`)
  is a template member that is not instantiated anywhere in this diff, so
  its body has only been syntax-checked, not type-checked, by this PR's
  build — a single unit test would also be the first real compile-check of
  that method.

### Praise

- Consolidating the resync order into one private `resyncDerivedState()`
  that all three entry points route through (rather than three
  independently-ordered call sites) directly addresses the recurring
  footgun class by construction, not just by convention/comment.
- The `createCross` comment explains *why* `editVoxels` is required there
  (deactivated interior would render black under the lit path) rather than
  restating the diff — good non-obvious-invariant comment per the doc
  comment bar.

- [ ] Add a headless unit test (in `test/ecs/`, following
      `voxel_set_target_canvas_test.cpp`'s pattern) exercising `editVoxels`
      or `carve` and asserting face-occupancy + active-mask end up correct
      for a mixed activate/deactivate result.

Opus recheck required: `resyncDerivedState()` consolidates the
rotation-source-mirror -> active-mask -> face-occupancy order that feeds
GPU-consumed voxel state (`c_voxel_visibility_compact` reads the active-mask
directly; face-occupancy bits gate per-face emission). Not a literal
`engine/render/` path, but it's render-invariant logic that a whole epic
(#2164) is about to build on — worth a second, independent look at the
ordering once the test above lands.

jakildev (commented):
## Opus recheck — engine/voxel: encapsulated edit API on `C_VoxelSetNew`

**Verdict:** needs-fix (concurring with Sonnet — the missing-test bar holds; the invariant it escalated is sound)

Sonnet flagged the `resyncDerivedState()` ordering (rotation-source mirror →
pool active-mask → face-occupancy) for an independent Opus look, since this API
feeds GPU-consumed voxel state and is the head of epic #2164. On closer read I
**confirm** the ordering is correct, and I confirm the two invariant claims
Sonnet could only reason about statically:

### Invariant recheck (confirmed)

- **Ordering matches every bulk mutator.** `resyncDerivedState()`
  (`component_voxel_set.hpp:539-544`) reproduces the exact sequence
  `deactivateAll` / `activateAll` / `fillPlane` / `reshape` all run inline. The
  significant detail is the *active-mask* step: `editVoxels` handles the general
  **mixed active/inactive** result, so routing it through
  `resyncRangeFromColors` (alpha-keyed re-derive) rather than the
  `markRangeActive` / `markRangeInactive` all-or-nothing shortcuts is the
  correct choice — it matches the only two other mixed-alpha sites in the file,
  `reshape` SPHERE (`:435`) and the dense-data ctor (`:273`). No drift.

- **Active-mask chain is internally consistent.** `C_Voxel::deactivate()` zeroes
  `color_.alpha_` and `activate()` sets it to 255 (`component_voxel.hpp:102-108`);
  `resyncRangeFromColors` keys the pool mask off alpha. So a `carve()` /
  `editVoxels` deactivation propagates correctly to the mask that
  `c_voxel_visibility_compact` reads. The `createCross` fix is real: the old path
  ran only `syncActiveMask()` (mask, **no** occupancy recompute) → stale
  face-occupancy → black interior; the new path adds `recomputeFaceOccupancy`.

- **"Byte-identical" for `createSnake` / `createLamp` holds, three-deep.** These
  two only rewrite `bone_id_` / `color_` (all alpha 255), never activation. The
  ctor seeds the mask with `markRangeActive` (all active), which equals the
  alpha-derived mask `resyncRangeFromColors` recomputes; `recomputeFaceOccupancy`
  re-derives identical occupancy; and `mirrorToRotationSource` is a no-op at
  construction because `rotationSourceVoxels_` stays empty until the first
  rotated GRID frame. So the added unconditional resync is genuinely idempotent
  here, not a behavior change.

### Needs-fix (concurring with Sonnet, not re-litigating)

- The test gap is a legitimate needs-fix for a foundational public API, and I'll
  sharpen one point Sonnet already raised: `carve()`
  (`component_voxel_set.hpp:460`) is a template member with **no instantiation in
  this diff**, so this PR's build has only *parsed* it, never instantiated or
  type-checked its body. A single headless test in `test/ecs/` (the
  `voxel_set_target_canvas_test.cpp` fixture already proves the component is
  exercisable without a RenderManager) that calls `carve()` on a mixed
  active/inactive result and asserts both the pool active-mask and
  face-occupancy bits would be the first real compile of that method *and* would
  regression-lock the ordering invariant the whole #2164 epic is about to build
  on.

No architectural follow-up needed — consolidating the three-step order into one
private `resyncDerivedState()` that all entry points route through addresses the
#2018/#2117/#2146 footgun class by construction. The design is sound; it just
needs the test before it becomes the base of the epic.

🤖 Opus recheck (review-pr skill) — invariant confirmed; deferring to Sonnet's needs-fix on test coverage.

jakildev:
Addressed `fleet:needs-fix`:

- Added `test/ecs/voxel_set_edit_api_test.cpp` (registered in `test/CMakeLists.txt`) — 3 headless GTest cases on the entity-keyed pool fixture (no RenderManager), asserting **both** derived invariants for a mixed active/inactive result:
  - `CarveDeactivatesSliceAndUpdatesMaskAndFaces` — `carve()` clears the x==0 slice; asserts the pool active-mask clears exactly those slots AND that a surviving neighbor re-exposes the face that pointed at a carved voxel (the recompute half the #2018/#2117/#2146 footgun drops).
  - `EditVoxelsActivatesSliceAndUpdatesMaskAndFaces` — `editVoxels()` activates a slice of an initially all-inactive set (activate direction); asserts mask + face bits track the post-edit active set.
  - `ResyncAfterRawEditsMatchesEncapsulatedPaths` — raw `voxels_` span writes + `resyncAfterRawEdits()`; same assertions.
- This is also the first real type-check of `carve()` (was a template member instantiated nowhere in the tree, per the review note).
- `fleet-build --target IrredenEngineTest` clean; all 3 tests PASS. Build + format-changed clean.

Reviewer please re-verify. — worker

jakildev (commented):
## Opus recheck (2nd pass) — engine/voxel edit API: test coverage landed

**Verdict:** approve

My prior Opus recheck and Sonnet's first pass both confirmed the design and the
`resyncDerivedState()` ordering invariant (rotation-source mirror → pool
active-mask → face-occupancy) are sound; the sole open item was Sonnet's
needs-fix — no test coverage for the new public API, and in particular `carve()`
being a template member instantiated nowhere in the tree (parsed but never
type-checked). The author added `4a06667d`
(`test/ecs/voxel_set_edit_api_test.cpp`, +182). On this pass I verified that
commit discharges the needs-fix, and I built and ran it to confirm.

### Build + run (macOS/Metal, this worktree)

- `fleet-build --target IrredenEngineTest` — clean build. This is the **first
  real instantiation of `carve()`** (the template body Sonnet flagged as only
  syntax-checked); it compiles and type-checks.
- `./build/IrredenEngineTest --gtest_filter='VoxelSetEditApiTest.*'` — **3/3 PASS**
  (`CarveDeactivatesSliceAndUpdatesMaskAndFaces`,
  `EditVoxelsActivatesSliceAndUpdatesMaskAndFaces`,
  `ResyncAfterRawEditsMatchesEncapsulatedPaths`).

### The test discharges exactly what was flagged

- Covers **all three** entry points — `carve` (deactivate direction),
  `editVoxels` (activate direction), and `resyncAfterRawEdits` (raw-span escape
  hatch) — each on a **mixed active/inactive result**, the case the
  active-mask-only footgun (#2018/#2117/#2146) silently breaks.
- Asserts **both** derived halves the invariant maintains: the pool per-slot
  active-mask (`C_VoxelPool::getActiveMask()`, keyed off alpha via
  `resyncRangeFromColors`) **and** the per-voxel face-occlusion bits.
- The face-bit expectations match `face_occupancy.hpp::recomputeFaceOccupancy`
  exactly, which I cross-read: bit **set** = in-grid neighbor active (occluded),
  bit **clear** = exposed, grid-boundary faces exposed (no cross-set lookup),
  inactive voxels cleared to all-zero. The survivor / carved / boundary
  assertions in each test are consistent with that semantics.
- Uses the headless `targetCanvas` fixture (same as
  `voxel_set_target_canvas_test.cpp`) — no RenderManager/GL/Metal context
  needed, so it regression-locks the ordering invariant the whole #2164 epic
  builds on, cheaply and on every host.

### On Sonnet's escalation

Sonnet escalated the ordering for an independent Opus look because this API feeds
GPU-consumed voxel state and heads epic #2164. I confirmed the ordering last pass
and re-confirm it holds unchanged; the test now pins it. Nothing further to
verify — the one gate that was open is closed.

No blocking issues. Approving. (Adding `fleet:needs-windows-smoke` — authored on
macOS, so the OpenGL tier still wants a smoke pass.)

🤖 Opus recheck (review-pr skill) — needs-fix discharged: test built + passed, `carve()` instantiated for the first time.

Issue #2124, opened 2026-06-29
render: fog-of-war filled cross-section on voxel geometry — two modes (epic)


jakildev:
**Epic.** Fog-of-war reveals voxel **objects** with a *filled cross-section* at the vision boundary, in two modes (voxel-resolution + smooth), across all canvas modes that carry fog.

## Problem
The smooth analytic vision circle (#2033) and the per-column cull/clip (#2040/#2055/#2068, and the per-fragment clip in #2106) reveal voxel objects by **dropping** boundary columns. Where the vision boundary cuts through a solid object you get **missing faces / see-through holes** (or, pre-#2106, a hard black wedge) instead of a cross-section showing the object's interior. #2102 delivered the per-fragment clip and **explicitly deferred** the filled cross-section as "a separate optional follow-on" — this epic is that follow-on, expanded to the full feature the design conversation settled on.

## Core insight
In a voxel world the cross-section already exists as geometry — it's the **interior faces**. A voxel side face is normally drawn only when its neighbor cell is empty ("exposed"). At the fog boundary the neighbor cell is *solid but fog-hidden*. So: **treat a fog-hidden neighbor as air and emit that interior side face as a "cut face."** No new geometry, no raymarch. Because the vision region is a vertical cylinder, only the **±X/±Y (vertical) faces** are ever cut; top/bottom never are. The cost is bounded — only the ≤2 vertical faces in a voxel's camera-visible triplet ever take the extra neighbor-column fog lookup.

## Two modes (tied to the active reveal path, not a new toggle)
- **Mode A — voxel-resolution**: grid reveal (`revealRadius`, integer cells). Boundary is voxel-blocky; objects reveal a column at a time; the cut is voxel-aligned interior faces.
- **Mode B — smooth**: analytic reveal (`setVisionCircle`). The same cut faces, trimmed by the existing per-pixel `FOG_TO_TRIXEL` analytic mask so faces *and* cut walls follow the smooth disc. (Reveal whole voxels via the cut, then smooth the silhouette the way the floor already is — the agreed approach.)

## Decisions settled
- **Cut shading = option 1**: cut faces shade exactly like a normal voxel face (object simply "ends" at the vision edge — full AO / sun-shadow / light-volume, no special tint). Lit-as-normal is the spec.
- **#2106 is held** (its per-column drop ships the missing-faces look). Phase 1 absorbs its per-voxel column clip as the foundation and supersedes it.
[...]

Issue #2126, opened 2026-06-29
render: fog cross-section — Mode B smooth on the GRID canvas (P2)


jakildev:
**Model:** opus
**Part of epic:** #2124
**Plan file:** `.fleet/plans/issue-2124.md` (full epic plan)
**Blocked by:** #2125

## Scope
Make the cross-section **smooth** under the analytic reveal (`setVisionCircle`) on the GRID canvas — Mode B. P1 gives voxel-resolution cut faces; this phase ensures that under the analytic disc the object silhouette AND the cut walls follow the same smooth curve the floor already does, instead of stair-stepping at column granularity.

## Approach
- The cut faces from P1 are real geometry, so the existing per-pixel `FOG_TO_TRIXEL` analytic mask already fades their pixels by the disc. Verify this end-to-end: the cut wall should be trimmed to the smooth disc edge, not popped at the column.
- Reconcile granularities: P1's cut-face predicate uses the **binary** neighbor test (`fogVisionCircleReveal(..., aa=0) >= 0.5`, "neighbor center outside"). Under the analytic reveal a boundary column is *partially* covered — confirm the cut face is EMITTED for any column the disc's smooth edge crosses (so there's geometry for `FOG_TO_TRIXEL` to trim), then trimmed per-pixel. If the binary test drops a column the smooth edge still touches, widen the cut-face emission to "neighbor not fully revealed" so the per-pixel mask owns the final edge.
- Keep everything on the shared `fogVisionCircleReveal` curve so floor edge, object silhouette, and cut wall are one analytic formula (no CPU/GPU or GL/Metal drift).
- fog_demo: smooth-mode cross-section scene + render-verify refs.

## Acceptance criteria
- Under `setVisionCircle`, a cut object's silhouette and its cut wall both follow the smooth disc edge — no per-column stair-step at the cut, no holes.
- Grid (Mode A) behavior from P1 unchanged; non-fog byte-identical.
- GL **and** Metal; fog_demo smooth-mode refs committed; cross-host smoke cleared.

## Dependencies
Blocked by #2125 (P1 cut-face mechanism). Independent of P3/P4.

## Gotchas
- The "reveal whole voxels then smooth" approach (agreed): emit the cut faces at voxel granularity, let the per-pixel mask do the smoothing — do NOT try to generate sub-voxel cut geometry.
[...]

PR #2175, opened 2026-07-01
render: fog cross-section cut faces shade as clean exposed faces

jakildev:
## Summary
The fog cross-section **cut face** (the interior wall exposed at the vision boundary, epic #2124) is geometrically buried in the solid, so the sun-shadow map reports it self-shadowed and AO reads it as a deep interior crease — together they render the cut wall as a dark smear instead of a clean cross-section.

The epic settled on *"option 1"* (cut faces get full AO + sun-shadow, "lit as normal"). In practice that produces a dark self-shaded cap wherever the cut face is buried behind the object relative to the sun. **This supersedes that decision**: a cut face now renders **fully lit** — Lambert + ambient + light-volume, no self-shadow, no crease AO — so the object cleanly "ends" at the vision edge.

### Mechanism
- A per-pixel **cut-face flag** rides **bit 29 of the stored entity id** (just below the #1960 priority tier), folded into the same masking chokepoint (`kEntityIdHighWordMask`) so every id reader (picking) ignores it.
- `c_voxel_to_trixel_stage_2` sets it on a non-exposed vertical face kept **only** by the fog cut rule (an exposed real surface is never flagged).
- `c_lighting_to_trixel` reads the entity-id image (new **image unit 6**) and forces `ao = shadow = 1` on flagged pixels.
- Scoped to the **single-canvas + detached** routes; the per-axis rotation route is guarded out (`perAxisRoute == 0`), its id image unbound there.
- **Non-fog scenes never set the flag → byte-identical.** GL + Metal kept in lockstep.

## Visual (fog_demo `--edge-zoom`, zoom 14)
The change is confined to the cut faces (the green slab's revealed cross-section here); every other pixel is byte-identical.

| before | after | diff (red = changed) |
|---|---|---|
| before | after | diff |

Measured (Metal): fog_demo `--edge-zoom` and the arcade scene both show cut faces **brighter, 0 pixels darker**, everything else byte-identical. The magnitude is modest in these framings because their visible cut faces are small/ambient-only; a near-wall raking view (where the cut face is a large buried surface) is where it reads strongest.

## Test plan
- [x] fog_demo `--edge-zoom` / `--edge-yaw-sweep`: cut faces brighter, 0 darker, rest byte-identical (Metal)
- [x] Per-axis rotation route unchanged (guarded out — `--edge-yaw-sweep` diff = 0)
- [x] Non-fog byte-identical by construction (flag gated on an active vision circle)
- [x] Builds clean: IRFogDemo (macOS/Metal); `format-changed` clean
- [ ] GL parity (cross-host smoke — Linux/Windows)

Relates to epic #2124 (supersedes its "option 1" cut-face shading decision). Does **not** address the separate screen-space-clipping of vertical cut faces (voxel-quantized silhouette) — that's a follow-up.


jakildev (commented):
## Review — render: fog cross-section cut faces shade as clean exposed faces (#2175)

**Verdict:** approve

Standalone PR (base `master`). Fog cross-section cut faces were reading as a
dark self-shadowed smear because the sun-shadow bake and AO both treat them
as buried interior geometry; this forces cut faces fully lit (AO = shadow =
1) via a new bit-29 carrier on the stored entity id.

### Verification performed

- Read every changed file in full context (`ir_render_types.hpp`,
  `system_lighting_to_trixel.hpp`, both GLSL shaders, all three Metal
  mirrors) — GLSL and Metal are structurally identical mirrors line-for-line;
  no drift between backends.
- **Bit-packing invariant.** `kEntityIdCutFaceMaskInHighWord = 1u << 29` sits
  strictly below `kEntityIdPriorityMaskInHighWord` (bits 30–31, #1960) — no
  overlap. `kEntityIdHighWordMask` now strips both, so `decodeEntityId` /
  `decodeCarrierEntityId` (the picking chokepoint, both GLSL/Metal and C++)
  transparently ignore the new bit — matches the PR's claim that picking is
  unaffected.
- **Encode ordering.** `c_voxel_to_trixel_stage_2` calls
  `encodeEntityIdWithPriority` first (which unconditionally clears bit 29 via
  the combined mask, so no stale flag survives from a reused SSBO slot), then
  `encodeEntityIdCutFace` afterward with the current frame's `isCutFace` —
  correct order, verified in both GLSL and Metal.
- **Byte-identity gate.** `isCutFace` is set only inside the pre-existing
  `fogActive` branch (unchanged by this PR); a non-fog scene never flips it,
  so `packedEntityId` is untouched and the change is provably a no-op
  off the fog path.
- **Binding-slot consistency.** New image unit 6 (`trixelEntityIds`,
  `RG32UI`/`rg32ui`) doesn't collide with the existing 0/1/2/3/4/5 slots in
  `LIGHTING_TO_TRIXEL`; `system_lighting_to_trixel.hpp` binds it
  unconditionally every tick (required for Metal's fixed kernel-argument
  table) and the comment correctly notes `dispatchPerAxisLighting` leaves it
  resident without rebinding, relying on the `perAxisRoute == 0` shader guard
  to skip the read on the rotation route — this mirrors the existing
  palette-LUT/light-volume resident-binding pattern in the same function.
- **Build + run (independent, this worktree, macOS/Metal):**
  `fleet-build --target IRFogDemo` — clean, no new warnings.
  `IRFogDemo --edge-zoom --auto-screenshot 5` — exits cleanly, 3 screenshots
  captured; visually confirmed the cut face on the green slab reads as a
  clean lit surface (no dark smear), consistent with the PR's own after
  screenshot.
- `mergeable: MERGEABLE` — no churn-audit needed.

### Nits

- `engine/render/CLAUDE.md` §"Verifying render changes" asks for an ROI crop
  pair (not just full-frame) on render PRs; the PR body has only the
  full-frame `--edge-zoom` trio. The doc itself notes ROI-crop shot-table
  support currently only exists in `shape_debug`, so this may not be
  practically actionable for `fog_demo` — flagging in case a future pass
  wants to add a `RoiCrop` shot to `fog_demo`'s table, not blocking.

### Praise

- The `isCutFace` gate reuses the exact `fogActive`/`fogColumnReveal`
  predicate stage 2 already computes for the cut-face distance write (P2,
  #2126) instead of re-deriving it — guarantees the lit-override can never
  fire on a face the distance pass didn't already treat as a cut face.
- Clearing the flag unconditionally via `encodeEntityIdWithPriority`'s mask
  before re-setting it per-frame avoids a stale-bit-from-last-frame class of
  bug that a naive "only set when true" encode would be vulnerable to on a
  reused SSBO slot.

- [ ] GL parity (cross-host smoke — the PR body already flags this as
      outstanding; tagging `fleet:needs-linux-smoke` /
      `fleet:needs-windows-smoke` below per the reviewer smoke protocol).

GPU entity-id bit-packing shared across GLSL/Metal, a new persistent
image-unit binding on a hot compute pass).

Opus recheck required: touches engine/render/ (entity-id carrier bit-packing
shared across GLSL/Metal + a new resident image binding on a hot compute
pass).

jakildev:
Addressed feedback: added an ROI crop (`cutface_slab`, 480x350 over the green slab) to the `fog_edge_zoom9` shot per engine/render/CLAUDE.md's ROI-crop-pair convention. Verified locally (macOS/Metal): crop cleanly shows the cut face rendering lit rather than a dark smear. Build clean, only `creations/demos/fog_demo/main.cpp` touched. — worker

jakildev (commented):
## Opus recheck — render: fog cross-section cut faces shade as clean exposed faces (#2175)

**Verdict:** approve

Sonnet approved-and-escalated for the two Opus-only surfaces: the GPU
entity-id bit-packing carrier shared across GLSL/Metal, and a new resident
image binding on the hot `LIGHTING_TO_TRIXEL` compute pass. I focused this
pass on exactly those — everything Sonnet's checklist covered (naming,
GLSL/Metal mirror parity, decode-chokepoint discipline) I confirm on read and
did not re-derive.

### Bit-packing carrier — confirmed sound

- **No overlap.** `kEntityIdCutFaceMaskInHighWord = 1u << 29` sits strictly
  below the priority tier (bits 30–31, #1960). `kEntityIdHighWordMask` now
  strips both, so the same single decode chokepoint (`decodeEntityId` /
  `decodeCarrierEntityId`, all three of GLSL/Metal/C++) transparently masks
  bit 29 off everywhere an id is READ — picking / hover never see it. Sonnet
  flagged this; on closer read I confirm it — the mask fold is the correct
  "neutralize the trap" placement, no reader open-codes the mask.
- **Encode ordering is correct.** `encodeEntityIdWithPriority` runs first and
  (via the widened `kEntityIdHighWordMask`) clears bit 29 from the raw id,
  then `encodeEntityIdCutFace` re-sets it per-frame only when `isCutFace`.
  That is what makes a reused SSBO slot's stale bit-29 from last frame safe —
  Sonnet called this out and it holds.
- **Stage-2 flag logic targets only the cross-section wall.** `isCutFace` is
  set true *only* when the face was non-exposed, vertical (`faceId <
  kFaceZNeg`), fog is active, AND the fog cut rule keeps it. Originally-exposed
  faces stay unflagged (normal lighting is correct for them). Correct.

### Resident image binding on the lighting pass — no hazard

- **No unit collision.** Image/sampler units in `c_lighting_to_trixel` are
  0,1,2 (images), 3 (sampler), 4 (image), 5 (sampler3D); unit 6 was free.
  SSBO bindings (7/23/27/28) are a separate namespace. Metal's flattened slot
  table now has 0–6 all populated (the unconditional bind is exactly why —
  Metal won't default-bind a null slot).
- **Format matches.** `bindAsImage(6, READ_ONLY, RG32UI)` ↔ shader
  `layout(rg32ui, binding=6) readonly uimage2D` ↔ Metal
  `texture2d<uint, access::read>`; the entity-id texture is the ctor-created
  RG32UI target (`makeCanvasEntityIdTexture`, cleared as RG_INTEGER/UINT32).
- **No new crash surface.** `getTextureEntityIds()` is created in the same
  `C_TriangleCanvasTextures` ctor as color/distance, which are already
  `bindAsImage`'d unconditionally in this same block — any canvas that reaches
  `LIGHTING_TO_TRIXEL` already has valid color/distance textures, so the id
  texture is guaranteed valid at the same sites.
- **No read-after-write hazard.** The id texture is written by
  `c_voxel_to_trixel_stage_2` in the `VOXEL_TO_TRIXEL_STAGE_1` tick; the
  lighting pass already depends (via `trixelDistances` / `canvasAO` image
  reads) on that stage's writes, so the existing barrier discipline covers the
  new read. This is a later-tick `access::read` of the canvas's *own* id
  texture — the sanctioned pattern, explicitly *not* the #1640 foreign-canvas
  R32I-scratch second-dispatch gotcha.
- **Per-axis route is correctly inert.** The image is bound every tick but the
  `perAxisRoute == 0` guard skips the read on the rotation route (P4/#2128
  scope), so this PR's single-canvas + detached scope is clean.

### Two notes for the human — neither blocks merge

1. **(robustness, optional)** The upload-boundary guard in
   `component_voxel_pool.hpp:257` still asserts
   `(entityId >> kEntityIdPriorityShift /*=62*/) == 0`, i.e. it only guards
   bits 62–63. This PR now also reserves **bit 61** (high-word bit 29). An id
   in `[2^61, 2^62)` would pass the guard yet get bit 61 stripped by
   `decodeEntityId` and falsely read as a cut face. Real ids are allocation
   counters far below 2^61 so this is theoretical (same risk class as the
   already-shipped #1960 carrier, one bit lower) — but the guard/comment
   accounting is now one bit stale. Consider a `kEntityIdCarrierShift = 61`
   and guarding against that if you want the assertion tight.

2. **(micro-perf, optional)** The `imageLoad(trixelEntityIds, pixel)` fires on
   every single-canvas-route lit pixel whenever `perAxisRoute == 0`, not gated
   on fog being active — so non-fog scenes pay one extra RG32UI fetch per
   pixel (output stays byte-identical, `decodeCutFace` just returns false).
   Negligible against the pass's existing multi-fetch cost, but if the lighting
   frame-data exposes a cheap `fogActive` uniform, gating the read on it would
   make non-fog scenes truly zero-cost.

### Cross-host

`fleet:needs-windows-smoke` already tags the GL-side validation Sonnet
requested (authored on macOS/Metal). GL parity smoke still outstanding per the
PR body — the human holds the merge on it.

🤖 Opus recheck by Claude (review-pr skill, opus-reviewer role)

PR #2178, opened 2026-07-01
render: no-priority perf fast-path — skip triangleEntityIds read in f_trixel_to_framebuffer (#2155)

### LINKED ISSUE #2155: render: no-priority perf fast-path — skip triangleEntityIds read in f_trixel_to_framebuffer (#2122 item 3)

jakildev:
## Summary
`f_trixel_to_framebuffer` read `triangleEntityIds` **unconditionally** at finalization to decode the #1960 per-trixel priority tier — one texture fetch per finalization fragment on the DEFAULT (no-priority) path, pre-existing since #1989. This adds a scene-level `anyPerTrixelPriority` flag so the shader **skips that fetch** when no voxel drawn into the canvas carries a non-zero per-trixel priority, restoring **perf-identity** (not just byte-identity) on the default path, on **both GL + Metal**. Closes the optional item 3 of #2122.

### How the flag is maintained (push-at-mutation, no per-voxel scan)
- `C_VoxelSetNew::changeVoxelPriority` / `changeVoxelPriorityAll` adjust a per-set count and push the delta to the owning `C_VoxelPool` aggregate (released on set teardown) — mirrors the existing active-mask sync via `IRPrefab::VoxelPool::`.
- `VOXEL_TO_TRIXEL_STAGE_1` stamps each canvas's `C_TriangleCanvasTextures::anyPerTrixelPriority_` from the pool count (O(1) read).
- The main gather (`TRIXEL_TO_FRAMEBUFFER`) and the detached composite (`ENTITY_CANVAS_TO_FRAMEBUFFER`) forward the stamp into the shared `FrameDataTrixelToFramebuffer` UBO, repurposing the offset-200 `_depthColorPad0` slot (4-byte scalar → no size/offset change; all std140 static_asserts hold).

### Plan deviation (noted for the reviewer)
The plan proposed keeping the aggregate on `C_VoxelSetNew` and OR-reducing it into the canvas "during raster" in `VOXEL_TO_TRIXEL_STAGE_1`. That tick iterates `<C_VoxelPool, C_TriangleCanvasTextures, C_CanvasLocalRotation>` — it has **no `C_VoxelSetNew` iteration** to hook. So the aggregate is bridged onto `C_VoxelPool` (one pool per canvas, directly in the tick) via the per-set delta; the per-set count + canvas stamp + UBO publish the plan called for are all preserved. Same intent, correct home.

### Correctness notes
- **Conservative-TRUE**: a false-positive flag only costs the fast path, never correctness. The only false-negative path would be authoring priority outside `changeVoxelPriority*`, which doesn't exist.
- **Picking unaffected**: the GLSL read keeps a `|| isMouseHovered` disjunct (`isMouseHovered` hoisted above the read); the Metal hover read is a separate texel (`hoverCoord`) and is left untouched.
- **Re-voxelize covered**: the rotating `DETACHED_REVOXELIZE` private pool inherits its source set's priority via `changeVoxelPriority*`, so a rotating priority solid still stamps 1 (proven below).

## Verification (macOS / Metal)
- **OFF-path byte-identity**: `shape_debug` captured against `origin/master` vs this branch — **26/28 shots byte-identical** (`cmp`). The 2 that differ (per-axis inter-cardinal + spin) also differ **run-to-run on the same binary** → known round-to-cell speckle, not this change.
- **ON-path tier correctness**: `scripts/depth-tier-verify.py` → **753/753 frames PASS, tier=2** at the interpenetration pixel (the rotating far unit's per-trixel priority carrier survives with the flag driving the read).
- **Build**: `IRShapeDebug` + `IRCanvasStress` clean.
- **Perf-identity**: the fetch is gated out on the default path (visible in the shader diff). A GPU-stage-timing micro-benchmark is deferred to cross-host smoke — the demos expose no stage-timing CLI flag and a single-fetch delta is within GPU-timing noise on these scenes.

## Screenshots (byte-identical — perf-only change; pairs are equal by design)
<details>
<summary>zoom1_origin (full frame)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_origin left cube silhouette (ROI crop)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_origin center cube top (ROI crop)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

## Test plan
- [x] `depth-tier-verify` ENABLED-path gate passes (753/753 tier=2)
- [x] shape_debug byte-identical master-vs-branch on deterministic shots
- [x] Builds clean (IRShapeDebug + IRCanvasStress, macOS/Metal)
- [ ] Linux/GL cross-host smoke (`fleet:needs-linux-smoke`)

Closes #2155


jakildev (commented):
## Review — render: no-priority perf fast-path — skip triangleEntityIds read in f_trixel_to_framebuffer (#2155)

**Verdict:** approve

Traced the trickiest part of this change: whether the author-time
`perTrixelPriorityVoxelCount_` aggregate is tracked against the correct
`C_VoxelPool` for a rotating `DETACHED_REVOXELIZE` unit (the PR body's own
"plan deviation" note flags this as the risky spot). Confirmed via
`entity_voxel_pool_canvas.hpp`: every voxel-pool canvas (world, static
detached, or re-voxelize) owns exactly one `C_VoxelPool` component on the
canvas entity itself, and the "private pool" for a re-voxelize canvas is
literally that same pool — the authored source voxels and the GPU-authored
MODE-1 dest-cube slots share the one pool/buffer. So `C_VoxelSetNew::canvasEntity_`
(bound at construction) always resolves `adjustPerTrixelPriorityVoxelCount`
against the pool the canvas actually rasters, including for a rotating
priority solid. Good catch documenting that risk explicitly in the PR body.

Also checked:
- **std140 layout**: `sizeof(FrameDataTrixelToFramebuffer) == 208` assert is
  unchanged, and `float _depthColorPad0_` → `int anyPerTrixelPriority_` is a
  same-size scalar rename at the same offset (200) — no assert broke, GL/Metal
  edited in lockstep.
- **`reserved_` mutation sites**: grepped the whole tree — only
  `changeVoxelPriority` / `changeVoxelPriorityAll` / `onDestroy` write it, so
  the count can't silently drift out of sync with an untracked writer.
  `IRPrefab::DenseVoxel::toComponent` (the `.vxs` dense-load bridge) doesn't
  round-trip `reserved_` at all, so a freshly-loaded set always starts at
  count 0 — consistent with the (pre-existing, out-of-scope-here) fact that
  per-trixel priority isn't persisted through DENSE load yet.
- **GLSL reorder**: hoisting the hover-position computation above the now-
  conditional `triangleEntityIds` read is inert — `origin` is defined earlier
  in `main()` (line 72), before either block.
- Math conventions (`IRMath::max` for the clamp), naming, and the
  `IRPrefab::VoxelPool::` free-function pool lookup all follow existing
  precedent (`changeVoxelColor` / `markVoxelActive` do the exact same shape).

### Nits
- No unit test for the new counting logic (`C_VoxelSetNew::changeVoxelPriority`/
  `changeVoxelPriorityAll`/`onDestroy` release, `C_VoxelPool::adjustPerTrixelPriorityVoxelCount`).
  The render-level verification (depth-tier-verify 753/753, byte-identity)
  covers the primary rasterization scenario end-to-end, but not the counter
  arithmetic itself in isolation (e.g. toggling priority on/off repeatedly on
  the same voxel, or `changeVoxelPriorityAll` followed by per-voxel
  `changeVoxelPriority` calls). `test/ecs/voxel_set_edit_api_test.cpp` looks
  like the natural home if this gets touched again.

- [ ] Linux/GL cross-host smoke (already tracked via `fleet:needs-linux-smoke`)

Escalation: please Opus-recheck before merge (touches: render pipeline UBO
layout / GPU-CPU handoff for the per-trixel priority tier system).

jakildev:
Addressed nit: added `VoxelSetEditApiTest` coverage for `adjustPerTrixelPriorityVoxelCount` arithmetic in isolation — same-voxel toggling (rising/falling edge, redundant no-op falling edge), `changeVoxelPriorityAll` interleaved with per-voxel `changeVoxelPriority` calls, and the `onDestroy` release path. All 6 tests in the fixture pass; full suite (1233 tests) green. — worker

jakildev (commented):
## Opus recheck — render: no-priority perf fast-path (#2155)

**Verdict:** approve

Sonnet approved and escalated for the UBO-layout / GPU-CPU-handoff recheck.
I focused this pass on the three things Sonnet's checklist can't fully close
on a first read: the push-at-mutation counter arithmetic, the concurrency of
the mutation path, and the load-bearing byte-identity invariant. All hold.

### Counter correctness (the fast-path's only failure mode is a false-negative)
The flag is conservative-TRUE, so the *only* way this becomes a rendering bug
(not just a perf regression) is a false-negative: a voxel that actually
rasters carrying priority while `hasPerTrixelPriority()` reports 0. Traced
both mutators against that:
- `changeVoxelPriority` increments/decrements by the exact per-slot 0x3-bit
  transition it just wrote, and pushes the same delta to the pool — so the
  slot it touches is always accounted for.
- `changeVoxelPriorityAll` writes `voxels_[i]` for `i < numVoxels_` and sets
  the count to `numVoxels_` over that *same* range. `numVoxels_ == size.x*y*z`
  (the full authored volume) on the success path (`component_voxel_set.hpp`
  ctor), and `index3DtoIndex1D(index, size_)` in `changeVoxelPriority` is
  bounded by that same volume — so the two mutators can't disagree about any
  *rendered* slot. A slot beyond `numVoxels_` isn't uploaded, so it can't
  raster regardless. No false-negative path exists. The `IRMath::max(0, …)`
  clamp on the pool aggregate is belt-and-suspenders for the FALSE signal only.

### Concurrency
`adjustPerTrixelPriorityVoxelCount` mutates a non-atomic `int` on the pool.
Confirmed the only caller of the priority mutators is
`canvas_stress/main.cpp:794` (`changeVoxelPriorityAll`), invoked from the
demo's single-threaded scene setup — **not** inside a PARALLEL_FOR tick — so
there's no race with the #1803/#1804 parallelism work. The pool's other
push-at-mutation counters (active-mask sync) follow the same single-threaded
contract, so this is consistent with the existing invariant, not a new one.

### Byte-identity invariant (why the skipped read is safe)
The default-path output is byte-identical **iff** `decodePriority(storedId)`
is 0 for every fragment whenever the pool aggregate is 0. That rests on
#1960's reservation that the entity-id carrier's top 2 bits are never set
except by `changeVoxelPriority*`. Since the aggregate counts exactly those
authorings, aggregate==0 ⇒ no carrier bits set ⇒ decoded tier 0 ⇒
`tier == depthPriorityMode`, matching the pre-skip `max(depthPriorityMode, 0)`.
The 26/28 byte-identical shots + 753/753 tier=2 verify empirically confirm
the reservation still holds end-to-end, including the rotating
`DETACHED_REVOXELIZE` private-pool case (which Sonnet correctly traced to the
one-pool-per-canvas ownership).

### GPU control-flow safety
- GLSL: the gated `textureLod(..., 0)` uses an explicit LOD, so no derivative
  requirement — safe under the `if (anyPerTrixelPriority != 0 ||
  isMouseHovered)` non-uniform branch. The hover disjunct preserves picking
  (the same sample still feeds the highlight-strip path below).
- Metal: `triangleEntityIds.read()` (a texel fetch, not a sampler) is safe in
  divergent control flow. `sampleEntityId` is now scoped inside the `if`; the
  hover block below reads the distinct `hoverCoord`, so picking is untouched
  and there's no dangling reference — confirmed against the PR head blob.
- std140: `anyPerTrixelPriority_` lands at offset 200 (the former
  `_depthColorPad0_`), a same-size 4-byte scalar; `sizeof == 208` and
  `offsetof(depthColorMode_) == 192` asserts are unchanged, and C++ / GLSL /
  Metal declare the member in lockstep at the same slot.

### Nits (non-blocking — do not re-review over this)
- Agree with Sonnet: no unit test for the counter arithmetic
  (`changeVoxelPriority` toggle round-trips, `changeVoxelPriorityAll` →
  per-voxel interleavings, `onDestroy` release). The conservative-TRUE design
  plus the render-level verification cover correctness, so this is a hardening
  nit, not a merge gate — `test/ecs/voxel_set_edit_api_test.cpp` is the natural
  home if this counter is touched again.

Sonnet flagged the DENSE-load path (`toComponent`) not round-tripping
`reserved_`; on closer read I confirm that's out-of-scope pre-existing
behavior — a freshly DENSE-loaded set starts at count 0 with no carrier bits,
which is self-consistent (aggregate 0 ⇒ fast path ⇒ tier 0, and there's no
priority to render), so it's not a correctness gap this PR introduces.

Linux/GL cross-host smoke is the remaining pre-merge item (already tracked).

🤖 Opus recheck by Claude (role-opus-reviewer)

jakildev:
Addressed feedback: the nit (missing unit coverage for `adjustPerTrixelPriorityVoxelCount`) was already fixed by a direct commit (`6925704e`, 3 new cases in `voxel_set_edit_api_test.cpp`: same-voxel toggle no-double-count, `changeVoxelPriorityAll` interleaved with per-voxel writes, `onDestroy` release path) before this pass. Verified: `IrredenEngineTest --gtest_filter=VoxelSetEditApiTest.*` — 6/6 pass, including the 3 new cases. Clearing `fleet:has-nits`; `fleet:approved` stands. — worker

PR #2180, opened 2026-07-01
render: fog cross-section vertical rim clips in screen space, not per-voxel

jakildev:
## Summary
The vision edge is a **game-resolution circle** on horizontal (Z/top) faces — `FOG_TO_TRIXEL` clips those per pixel against the analytic disc. But a fog-clipped voxel object's **vertical rim** (a ground slab's edge, a wall's side) was **voxel-jagged**: the own-column drop (#2102) removed a column WHOLE when its *center* fell outside the disc, so the geometry ended on the voxel lattice before the per-pixel mask could trim it.

**Fix:** evaluate the drop's reveal at the cell point **nearest** each vision-circle center rather than the cell center (`fogColumnRevealNearest`). A column the disc merely *clips* (center outside R, unit cell still overlaps) is now **kept and rasterized**, and `FOG_TO_TRIXEL` trims it per pixel at the exact analytic edge — so the vertical rim follows the smooth disc at game resolution, matching the top face. The nearest-cell test keeps **only the one-cell ring** the disc crosses (tight, not a blanket margin).

## Visual (fog_demo `--edge-zoom` z14, near disc arc — a curve shows the jaggedness clearest)
| before (voxel staircase) | after (clips to the analytic disc) |
|---|---|
| before | after |

Full frames: [before](https://raw.githubusercontent.com/jakildev/IrredenEngine/claude/fog-cross-section-screenspace-clip/docs/pr-screenshots/claude-fog-cross-section-screenspace-clip/rim_before.png) · [after](https://raw.githubusercontent.com/jakildev/IrredenEngine/claude/fog-cross-section-screenspace-clip/docs/pr-screenshots/claude-fog-cross-section-screenspace-clip/rim_after.png). The gray fog-viz plane behind the slab is smooth in both — it's the blue **voxel** slab's rim that goes from staircase → smooth.

## Scope / safety
- **Single-canvas world route only** (`perAxisRoute == 0`); per-axis rotation route unchanged.
- **Grid-reveal (no vision circle) and detached scenes: byte-identical** (render-verify `fog_zoom*_origin` + `fog_detached_edge_*` PASS 100%).
- The 6 analytic `fog_edge_zoom{5,9,14}` / `fog_edge_smooth{5,9,14}` macOS refs are **re-blessed** (intentional new baseline — the smooth rim + a new ground slab in the scene).
- `fog_demo` `--edge-zoom`/`--edge-smooth` gains a ground slab so the rim silhouette is a covered regression case.
[...]

PR #2040, opened 2026-06-26
render: fog-of-war culls unexplored-column voxels at the compaction

jakildev:
## Summary
- **#2008:** a voxel object on an *unrevealed* XY column whose top reaches the camera-visible region rendered as a **black silhouette**. `FOG_TO_TRIXEL` is a screen-space post-process that hard-blacks already-rasterized unexplored pixels — not a geometry cull — so a tall object's top mapped back to its unrevealed base column and painted black over the visible region.
- **Fix — geometry cull at the source:** `c_voxel_visibility_compact` (GL + Metal) now drops any voxel whose **raw world (x,y) column** is unexplored, so the whole object never rasterizes and revealed geometry *behind* it shows through.
- The canvas fog texture binds at the compaction (image slot 0), gated to the world fog canvas via `getComponentOptional<C_CanvasFogOfWar>` (per-canvas, mirrors `LIGHTING_TO_TRIXEL`'s sun-shadow — not the per-voxel `getComponent` footgun). Every other canvas binds a shared **1×1 all-visible placeholder**, so the `imageSize<=1` short-circuit makes non-fog scenes byte-identical to master.
- The dirty-gated `subImage2D` upload **moves** from `FOG_TO_TRIXEL` into `VOXEL_TO_TRIXEL_STAGE_1` so the cull and the post-process read the **same current-frame fog** (no one-frame lag, no startup-frame garbage).

## Deviation from the plan (step 4 — kept the hard-black store)
The plan called for removing the `FOG_TO_TRIXEL` unexplored hard-black `imageStore`. **I kept it.** The compaction cull only covers **voxel-pool** geometry; SDF shapes (`SHAPES_TO_TRIXEL`), text, and overlays in unexplored columns still rely on the screen-space black-fill. `fog_demo`'s floor + shapes are SDF — the black region around the visible disk **is** the post-process black-filling the SDF floor; removing it would turn the whole unexplored region from black to lit gray (a regression). The cull and the post-process are **complementary**: the cull fixes the voxel silhouette + behind-occlusion case; the post-process remains the screen-space catch-all for non-voxel geometry.

## Verification (macOS / Metal)
- **Cull works:** `fog_demo` gains a tall voxel pillar on an unexplored column. Cull ON → pillar fully culled, disk clean. Cull manually disabled → black silhouette bisects the disk (the bug). See screenshots.
[...]

Issue #2068, opened 2026-06-27
render: de-jag the fog vision-circle edge (conservative column cull)


jakildev:
**Area:** engine/render
**Model:** sonnet
**Blocked by:** (none)

## Context

Follow-up cleanup on the analytic fog vision-circle reveal (#2055 / #2033 / #2009). The smooth disc reveal is working well, but two artifacts remain along the circle edge, visible on a voxel-floor fog scene (macOS/Metal, zoomed in):

1. **Voxel-grid-aligned jaggies along the smooth arc** — particularly when zoomed in, at the bottom of the disc, the otherwise-smooth circular radius shows regular bumps/notches that take the shape of voxels and align with the voxel grid.
2. **Blacked-out voxel-edge notches on objects partially revealed at the boundary** — a tall object straddling the vision edge shows voxel-edge-aligned black chunks on its top (Z) and side (XY) faces, with inconsistent / "weird" blending between the smoothly-revealed part and the hard-black part.

Both trace to the **same seam**: fog reveal runs at two granularities.

- The **column cull** (`fogColumnInVisionCircle` in `c_voxel_visibility_compact.glsl`) decides per *whole voxel column* whether voxels rasterize at all, testing the column's **integer center** against `radius + softness + 1.0`.
- The **per-pixel mask** (`c_fog_to_trixel.glsl`) then does the smooth `smoothstep` disc reveal on surviving pixels at sub-pixel resolution.

The smooth disc can only fade *within columns the cull kept*. Wherever the coarse center-tested cull boundary disagrees with the fine per-pixel boundary by a fraction of a cell, the dropped column leaves a voxel-shaped notch the smooth mask can't fill — on the floor arc (symptom 1) and on object faces (symptom 2).

## Scope

Make the column cull a strict **superset** of what the per-pixel smooth mask can ever touch, so the per-pixel mask becomes the **sole** edge authority and resolves the boundary at pixel resolution. Keep the current soft vertical-cylinder clip for objects (per design decision below). Do NOT change the per-pixel mask (`c_fog_to_trixel`) — it is already the smooth authority; this only stops the cull from clipping it.

**Out of scope (deferred, do not file yet):** filled cross-section / capping for objects the disc cuts through. The hollow vertical clip (background shows through the cut, no filled volume) is accepted for now; re-judge after this ships. See "Deferred" in the plan comment.

## Acceptance criteria

[...]

PR #2189, opened 2026-07-02
render: image-space fog cross-section — analytic cut in FOG_TO_TRIXEL

jakildev:
## Summary
Completes the screen-space fog cross-section (#2124): the vision-circle cut through voxel geometry is now a **single smooth, voxel-colored, artifact-free surface at game resolution** — computed entirely in image space.

**Why the rework** (supersedes this branch's first approach): rasterizing a displaced "band" into `trixelDistances` fights every depth-consuming pass — it creased AO at the junction, cast phantom sun shadows, and forward-mapped taps left posts/seams at grazing angles. Any scatter that injects cut geometry into the shared depth channel inherits those artifacts structurally.

**The image-space cut** (color-only, after lighting):
- **Render as-if-unfogged near the rim.** Stage 1 + the compact keep a ring of fog-hidden columns around the disc (`kFogHiddenKeepCells`), so hidden matter near the boundary renders normally — lit, shadowed, AO'd. (Hidden ring voxels resume casting shadows near the rim — the pre-#2040 semantics.)
- **Cut per pixel in `FOG_TO_TRIXEL`.** For a pixel that would fog-dark: recover its view ray from the same inverse projection the mask already uses, intersect it with each hard disc analytically, and ask the existing camera-anchored **light-occlusion bitfield** whether the entry point is solid. Solid ⇒ repaint the pixel as the toned lit color of the hidden surface — the cut. Rim pixels take their own surface as the cut (the ray test is degenerate exactly there and would dash the arc/tangent loci).
- **No depth writes, no new geometry, no new passes.** AO, the sun-shadow bake, and lighting are structurally unaffected; the silhouette is per-pixel analytic. `SYSTEM FOG_TO_TRIXEL` stamps `cutSolidsAvailable_` on the observer upload and transiently re-binds the occlusion grid to alias slot 28; creations without `BUILD_LIGHT_OCCLUSION_GRID` fall back to the plain dark mask.

**Cost:** one extra inverse-projection + a quadratic + ≤2 SSBO bit reads, only on fog-dark pixels of fog canvases — negligible against the raster/lighting pipeline.

## Visual (fog_demo `--edge-zoom` z14)
| | before (#2180 master) | after |
|---|---|---|
| near-side rim | before | after |
| slab cut | before | after |

Full frames: [before](https://raw.githubusercontent.com/jakildev/IrredenEngine/claude/fog-cross-section-analytic-band/docs/pr-screenshots/claude-fog-cross-section-analytic-band/band_before.png) · [after](https://raw.githubusercontent.com/jakildev/IrredenEngine/claude/fog-cross-section-analytic-band/docs/pr-screenshots/claude-fog-cross-section-analytic-band/band_after.png)

## Scope / safety
- **Mode A (hard discs) only**; Mode B soft discs keep their faded look (hard-only gates in both the ray test and the rim shortcut); explored grid memory keeps its desaturated tone.
- **Grid-only reveal byte-identical** (render-verify `fog_zoom*_origin` 100%).
- Cut color inherits the hidden surface's lit color (×`kFogCutTone` 0.85) — where the source face family changes (top vs side) the band picks up a mild tonal shift that reads as shading.
- Known accepted residual: a ray that crosses the disc through *air* (thin hidden wall in front of the rim) falls back to the dark mask.
- macOS refs re-blessed (`edge`, `edge_smooth`, `detached_edge` — the detached deltas also fold in #2162's silhouette-riser change, which landed mid-branch).

## Test plan
- [x] `render-verify --target IRFogDemo`: 11/11 PASS on the committed refs
- [x] `--edge-zoom` z5/z9/z14: cut surfaces smooth + uniform; no posts, no banding, no AO crease, no phantom shadows
- [x] `--player-walk`: smooth crescents with toned cut edges under sub-voxel motion
- [x] Builds clean (macOS/Metal); `format-changed` applied
- [ ] GL parity + `linux-debug` refs (cross-host smoke — Linux/OpenGL host re-blesses via `render-verify --update-references`)

Relates to epic #2124. Follows #2180.


jakildev (commented):
## Review — render: fog cross-section analytic band on the vision-circle cut

**Verdict:** approve

Shader-only diff (`c_voxel_to_trixel_stage_{1,2}.glsl` + their `.metal` twins, plus fog_demo macOS reference PNGs). No C++, no ECS, no serialization touched.

Dispatched a focused render-invariant pass (full-file context via `git show FETCH_HEAD:<path>` for all 4 shaders, not just diff hunks) against the PR's own byte-identical claims:

- **GLSL↔Metal parity** — `fogColumnRevealFarthest`, `fogPointReveal`, the widened `keepFace` gate, and the full band-projection block (margin-selection loop, engagement window, radial re-homing, shading-slot, coverage-backstop dual `emitDeformedFace`) are semantically identical between each stage's GLSL and Metal version. Remaining diffs are unavoidable API-surface differences (Metal threads `canvasFogOfWar`/`fogObservers`/`distanceScratch`/`canvasSize` explicitly; GLSL uses bound uniforms/images).
- **Stage 1 ↔ Stage 2 parity** — both stages compute identical `microWorldXY`, `bestMargin`/`bestCircle` selection, engagement window, `bandSlot`, and `latticeTap`/`bandTap` positions/depths. Only difference is the documented one: stage 1 emits distance-only, stage 2 additionally carries `voxelColor`/`packedEntityId` — can't desync the tap positions the two stages agree on.
- **Guard ordering / `kFogBandMinSubdivisions`** — the per-micro point-clip always runs before the `isCutFace && subdivisions >= 4` band gate in all 4 files; below 4 subdivisions the band never engages and falls back cleanly to the pre-existing lattice-stepped cut (#2180 behavior). The `1.5`-cell outward-travel clamp algebraically matches the PR body's stated bound in all 4 files.
- **Metal's `frameData.isDetachedCanvas > 0.5f` argument inside the `isDetachedCanvas < 0.5f`-gated block** — always evaluates false at this call site, but matches the pre-existing calling convention used by every other `emitDeformedFace` call in the same kernel (unconditional forwarding of the frame-level flag) — not a copy-paste defect introduced by this PR.

### Nits
- `c_voxel_to_trixel_stage_1.glsl` (and the identical spot in `_stage_2.glsl` / both `.metal` twins), in the band-projection block: `radial * (target / length(radial))` divides by zero if a vision-circle center sits exactly on a cut-face micro's world position. In practice this requires a vision source embedded inside solid cut geometry with the circle radius landing inside the `(2·microHalf, 1.5+2·microHalf]` engagement window — essentially unreachable given vision sources aren't placed inside solids, and even if it fired, the existing `isInsideCanvas` bounds check silently drops the resulting NaN/huge pixel coordinate while the coverage-backstop lattice tap (emitted in the same `emitDeformedFace` pair, just before) still covers the position. An epsilon guard (`length(radial) < 1e-4 → skip projection, keep only the lattice tap`) would be cheap insurance for determinism but isn't required for correctness or safety.

### Praise
- The four-way duplication (2 stages × 2 backends) is kept honest with explicit "MUST stay byte-identical to X's twin block" comments at every duplicated site, pointing at the specific twin — that's what made the parity check above tractable instead of a guessing game.
- `fogColumnRevealFarthest` re-deriving the farthest corner **per vision circle inside the loop** (not hoisted outside) is the detail that keeps the multi-circle "reveal ANY corner poke" test correct instead of silently regressing to a single-circle approximation.

- [ ] The PR body's own unchecked item: GL parity + `linux-debug` reference re-bless on a Linux/OpenGL host (handled by cross-host smoke, tagging below — not a merge blocker on its own).

Escalation: please Opus-recheck before merge (touches: engine/render/ shaders — GLSL+Metal, non-trivial analytic-geometry math, and a 4-way byte-identical duplication invariant across stage×backend).

Opus recheck required: engine/render/ shader changes with non-trivial per-micro analytic-geometry math (radial band projection, multi-circle margin selection) and a cross-cutting byte-identical invariant across 2 stages × 2 backends — worth a second pass given the subtlety even though the focused invariant audit found no blockers.

jakildev (commented):
## Opus recheck — render: fog cross-section analytic band on the vision-circle cut

**Verdict:** approve (nits)

Sonnet approved and escalated for a second pass on the non-trivial analytic-geometry math and the 4-way (2 stages × 2 backends) byte-identical duplication invariant. On closer read I **confirm** Sonnet's findings and add depth on the three things the first pass could not fully close:

**1. Cross-file invariant — confirmed, with a precision on "byte-identical."** The literal text can't be identical (stage 2 threads `voxelColor`/`packedEntityId` into `emitDeformedFace`; the Metal twins have entirely different emit signatures). What actually has to match — and does — is the *tap geometry*: `microWorldXY`, the `bestMargin`/`bestCircle` selection window (`> 2·microHalf && ≤ 1.5 + 2·microHalf`), `radial`, `xSlot`/`ySlot`/`bandSlot`, `latticeTap`, and `bandTap` (via `target = z − 2·microHalf`, `roundHalfUp(surfFixedXY)`) are computed identically in all four files. So stage 1's distance tap and stage 2's colour tap land on the same cell + depth + shading slot — the band wall can't read as cleared background. Confirmed across GLSL↔Metal and stage1↔stage2.

**2. Depth consistency of the band emit — confirmed (Sonnet did not check this).** The band block emits directly with `encodeDepthWithFace(latticeTap.x+latticeTap.y+latticeTap.z, bandSlot)` and `return`s, bypassing the normal emit path below. I verified the normal path's world route computes the *same* `depthBase = microPositionFixed.x + .y + .z` (the detached `isoDepthAlongAxis` branch is excluded because the whole block is gated on `isDetachedCanvas < 0.5`), and applies the identical `rotateCardinalZ` + `cardinalLowerCornerShift` before the sum. So band walls depth-sort consistently against adjacent non-band faces — no z-fighting between the analytic band and the surrounding lattice geometry.

**3. No missing-symbol hazard on the unbuilt GL side.** This PR is `authored-on-macos` with GL/`linux-debug` parity still pending cross-host smoke, so I checked the symbols the new GLSL references but doesn't define in-diff: `kFogColumnCellHalf` (pre-existing in stage-1 GLSL/Metal from #2068, newly added in stage-2's diff — present in all four), `visibleFaceIds` (uniform ivec4), `perAxisRoute` (uniform int), `roundHalfUp`. All defined/pre-existing. No compile break lurking on the GL build.

**Agree with Sonnet's div-by-zero nit.** `radial * (target / length(radial))` divides by `length(radial)` which is 0 only if a vision-circle center sits exactly on a cut-face micro's continuous world position with a radius inside the `(2·microHalf, 1.5+2·microHalf]` window — a vision source embedded in solid cut geometry with a ~1-cell radius. Unreachable in practice, and if it fired the `isInsideCanvas` bound drops the NaN pixel while the coverage-backstop lattice tap (emitted in the same pair) still covers the position. Cheap `length(radial) < 1e-4 → skip projection` guard is worth adding for determinism but is not a correctness blocker.

### Nits (not merge-blocking)
- Epsilon guard on `radial * (target / length(radial))` in all four files, as above.
- The per-micro `fogPointReveal` clip now runs a `visionCircleCount` loop for **every micro of every face** when `fogActive` (not just cut faces), where previously exposed faces were kept and trimmed only by the per-pixel FOG_TO_TRIXEL mask. Visually equivalent (conservative half-micro `aa`, mask owns the sub-pixel edge) and it culls fill *earlier* so it likely nets positive, but it is a new per-micro cost on the voxel→trixel hot path when fog is on. The perf-gate has no committed baseline yet — worth a glance at fog-scene frame time once one lands. Design-accepted, noting for visibility.

### For the human
- GL parity + `linux-debug` reference re-bless is still open (PR body's own unchecked item, handled by cross-host smoke — `needs-windows-smoke` is set). Not a merge blocker on its own, but the four-way duplication means the GL walls should be eyeballed once, not assumed from the Metal references.

jakildev:
Addressed nit: added an epsilon guard (`kFogBandRadialEpsilon = 1e-4`) around the radial re-homing divide in the band-tap projection, in all four stage×backend twins — skips the projected band emit (keeps the already-emitted lattice-tap backstop) when the vision-circle center coincides with the micro's world position. Confirmed byte-identical on every reachable input: `render-verify` on `fog_demo` (macos-debug) — 11/11 shots 100% match, 0 max-delta, inf PSNR. Build clean. — worker

jakildev (commented):
## Opus recheck (re-review) — guard fog band radial re-homing against zero-length divide

**Verdict:** approve

Re-reviewing after the follow-up commit `778df4a1` landed post my earlier approval — it addresses the radial re-homing divide-by-length hazard I flagged in the analytic band math. Fresh pass on that delta only (the rest of the PR is unchanged from my prior approval):

- **Behavior-preserving for all reachable inputs.** The guard hoists the previously-inline `length(radial)` into `radialLen` and wraps the band-tap emission in `if (radialLen >= kFogBandRadialEpsilon)`. When `radialLen >= 1e-4` (the practical always-case) the emitted `target` / `surfFixedXY` / `bandTap` / `bandDepth` and the `emitDeformedFace` call are identical to the diff I already approved — so this is a strict superset, not a visual change. Only the degenerate `radial ~= 0` case (vision-circle center coincident with the cut-face micro) changes: it now skips the band tap and relies on the unconditional lattice tap emitted immediately above, which already backstops coverage. Correct fallback.
- **4-way byte-identical invariant held.** `kFogBandRadialEpsilon = 1e-4` (`1e-4f` in Metal), the `radialLen` hoist, the `>= epsilon` gate, and the guarded block structure are identical across `c_voxel_to_trixel_stage_{1,2}.glsl` and their `.metal` twins. Remaining diffs are the legitimate per-stage×backend `emitDeformedFace` signatures and the unavoidable `2.0`/`2.0f`, `vec2`/`float2`, `ivec3`/`int3` API surface — same duplication convention I confirmed last pass.
- **No scope leakage.** `target`, `surfFixedXY`, `bandTap`, `bandDepth` were local to the emission and are now correctly scoped inside the `if`, followed by `return`. Nothing downstream reads them.

Carrying forward from my earlier recheck (unchanged by this commit):
- The per-micro `fogPointReveal` clip runs a `visionCircleCount` loop for every micro of every face when `fogActive` — a new per-micro cost on the voxel→trixel hot path. Design-accepted; worth a glance at fog-scene frame time once a perf baseline lands.
- GL parity + `linux-debug` reference re-bless remains the PR body's own open item, handled by cross-host smoke (`fleet:needs-windows-smoke` set). The 4-way duplication means the GL walls should be eyeballed once on the smoke host, not assumed from the Metal references.

jakildev:
Nit already addressed and re-approved in a prior iteration (commit 778df4a1 — guards the radial re-homing divide-by-zero with an epsilon check, confirmed byte-identical for all reachable inputs, opus recheck re-approved on that commit). Clearing the stale `fleet:has-nits` label — no further changes needed. — worker

jakildev:
Reworked per feedback: the raster-scatter band (previous head) is fully reverted — banding/AO-crease/phantom-shadow artifacts were structural to injecting cut geometry into `trixelDistances`. The cut is now computed per pixel in `FOG_TO_TRIXEL` after lighting (color-only): ray∩cylinder analytically + solidity from the existing light-occlusion bitfield, with an as-if-unfogged keep ring so the hidden matter under the cut is normally lit. PR body + screenshots updated; render-verify 11/11 on the new refs.

jakildev (commented):
## Review — render: image-space fog cross-section — analytic cut in FOG_TO_TRIXEL

**Verdict:** approve

### Prior-review resolution

The prior fleet review (2026-07-02, approved with nits, plus an Opus recheck
that re-approved after an epsilon-guard fix) reviewed a completely different
technique — a raster-scatter band injected into `trixelDistances` in
`c_voxel_to_trixel_stage_{1,2}`. Per the PR body's "Why the rework" section
that approach is **fully reverted** on this head (creasing AO, phantom sun
shadows, forward-mapped-tap seams were structural to injecting cut geometry
into the shared depth channel). None of the prior findings apply to the
current diff — it's a from-scratch image-space color-only cut in
`FOG_TO_TRIXEL`, reviewed fresh below.

This is a shader-only diff (`c_fog_to_trixel`, `c_voxel_to_trixel_stage_1`,
`c_voxel_visibility_compact` × GLSL+Metal) plus the CPU-side observer struct
and `system_fog_to_trixel.hpp` upload/bind wiring. No C++ ECS/ownership
concerns, no serialization.

### Checks performed

- **std140/Metal struct parity.** `FrameDataFogObservers::cutSolidsAvailable_`
  repurposes the struct's first padding lane (`pad0_` → `cutSolidsAvailable_`,
  `pad1_`/`pad2_` untouched) — CPU struct size unchanged (144 bytes, still a
  multiple of the vec4-array's 16-byte std140 base alignment). GLSL's
  `FogObserverData` UBO and Metal's `FogObserverData` struct both declare
  `visionCircleCount` then `fogCutSolidsAvailable` in the same order as the
  CPU struct — offsets line up on both backends.
- **Binding-28 reuse.** `kBufferIndex_LightOcclusionGrid == 28` (confirmed in
  `ir_render_types.hpp`), consistent with the documented shared-alias-slot
  convention in `engine/render/CLAUDE.md` ("GPU buffer bind-point budget is
  full"). `system_fog_to_trixel.hpp` explicitly re-binds the slot for its own
  dispatch (`occlusionGridSsbo_ != nullptr ? occlusionGridSsbo_ : observerBuf_`)
  and gates every shader-side read behind `fogCutSolidsAvailable`, so the
  observer-UBO-as-placeholder-SSBO bind is never actually dereferenced when
  the real occlusion grid isn't live. `Buffer::bindBase` (OpenGL:
  `glBindBufferBase`) is target-agnostic per-call — binding a buffer created
  against `BufferTarget::UNIFORM` as `SHADER_STORAGE` for one dispatch is
  legal GL, matching the existing pattern other passes in this pipeline
  already use for this same slot (sun-shadow bake, per-axis resolve scratch).
- **Occlusion-grid index math.** `fogCutVoxelSolid` (GLSL) /
  `fogCutVoxelSolid` (Metal) is byte-for-byte the same origin-relative
  flat-index formula as the existing
  `c_propagate_light_volume.{glsl,metal}::voxelOcclusionGetBit` — verified
  by direct comparison, not just the "keep in sync" comment. No transcription
  drift.
- **Cull-margin consistency.** `kFogHiddenKeepCells` (8.0, stage 1) +
  `kFogColumnKeepAa` (0.5) = 8.5; `c_voxel_visibility_compact`'s
  `kCullSafetyCells` bumped 1.0 → 9.0, a correct superset so ring voxels
  the cut needs to read survive the coarser per-column cull.
  Mirrored identically in both GLSL and Metal twins.
- **Gating order in `c_fog_to_trixel`.** The new cut block sits *after* the
  existing `if (state >= 1.0) return;` fully-visible fast path, so it only
  ever runs on already-fog-dimmed pixels — matches the stated design ("a
  pixel that would fog-dark"). `gridState < kFogExploredValue` correctly
  scopes the cut to the unexplored-grid-memory case only (explored/128-255
  keeps its desaturated tone), using the pre-circle-combine `gridState`
  rather than the circle-combined `state`, so a partially-revealed rim pixel
  doesn't wrongly skip the grid-only cut path.
- **Ray-vs-circle math.** Standard quadratic entry-root solve
  (`t = (-b - sqrt(det)) / 2a`), guarded on `a > 0.0` (degenerate purely-
  vertical ray direction — falls through to the plain dark mask rather than
  dividing by zero) and `det <= 0.0` (ray misses). The `rayStep` direction is
  recovered via the same finite-difference-over-depth technique the file
  already uses for `worldPerPixel` (the AA-rim computation just above it),
  which is valid because `trixelCanvasPixelToWorld3D` is affine in `rawDepth`
  for this orthographic iso projection — no perspective term to introduce
  nonlinearity between the two probed depths.
- **Both backends stay in lockstep** — GLSL and Metal implementations of the
  cut block, the occlusion lookup, and the cull-margin bump are structurally
  identical modulo the expected API-surface differences (buffer args threaded
  explicitly on Metal vs. bound globals on GLSL).

No blockers, no needs-fix items.

### Praise
- Making a local `FrameDataFogObservers observers` copy in
  `system_fog_to_trixel.hpp` to stamp `cutSolidsAvailable_` before upload —
  rather than mutating `fog.observers_` in place — keeps the pipeline-derived
  fact out of client-owned state. Right call given the header comment already
  documents `cutSolidsAvailable_` as "system-stamped on upload, NOT client
  state."
- The rework rationale in the PR body (why the raster-scatter band was
  reverted in favor of a color-only image-space cut) is a genuinely useful
  precedent for future "inject geometry into a shared depth channel" ideas
  in this pipeline — creasing AO / phantom shadows / forward-map seams are
  now documented as the generic failure mode, not just this feature's bug.

- [x] Already covered per PR body: `render-verify --target IRFogDemo` 11/11,
  `--edge-zoom` z5/z9/z14 visual check, `--player-walk`, macOS build clean.
- [ ] GL parity / `linux-debug` reference re-bless (cross-host smoke —
  already tagged via `fleet:needs-windows-smoke`, not a merge blocker on its
  own).

Escalation: please Opus-recheck before merge (touches: engine/render/
shaders — GLSL+Metal, non-trivial per-pixel analytic ray/circle geometry,
and a GPU buffer-binding-reuse pattern across the two backends).

Opus recheck required: engine/render/ shader changes with non-trivial
analytic ray-vs-disc intersection math and a shared-binding-slot reuse
pattern (UBO bound as SSBO placeholder at slot 28) — worth a second pass on
the geometry and the binding-aliasing safety even though this focused
review found no blockers.

jakildev:
Amended (head `fcc782db`): added `kFogCutLift` (0.06) on top of `kFogCutTone` — a cut through dark matter (near-black floors) was collapsing into the fog black and reading as a black cross-section; the lift keeps every cut legibly brighter than the fog while leaving bright materials visually unchanged. macOS refs re-blessed; render-verify 11/11. PR body screenshots refreshed to the lifted output.

jakildev (commented):
## Opus recheck — render: image-space fog cross-section — analytic cut in FOG_TO_TRIXEL

**Verdict:** approve

Sonnet approved and escalated for a second pass on the analytic ray-vs-disc
geometry and the binding-28 UBO-as-SSBO aliasing. Note this head is a
**from-scratch rework** — the prior fleet reviews (and my earlier epsilon-guard
recheck) were against the now-reverted raster-scatter-into-`trixelDistances`
band, so none of those findings carry over. I reviewed this diff fresh and
**confirm** Sonnet's verdict. The three things Sonnet's focused pass could not
fully close, I closed by direct comparison against the producer and the
existing consumer:

**1. Binding-28 occlusion struct — the fog shaders declare a *fresh* layout, so
I verified it three ways, not just against the "keep in sync" comment.**
- Producer: `system_build_light_occlusion_grid.hpp` writes
  `header_.worldOriginVoxel_ = ivec4(origin, 0)` as a 16-byte header at offset 0,
  followed by the voxel-existence bitfield, into a buffer created **once** at
  `create()` with fixed `kLightOcclusionSSBOByteSize` (256³, never resized).
- Existing consumer: `c_propagate_light_volume.{glsl,metal}`'s
  `LightOcclusionGrid` / `LightOcclusionData` + `voxelOcclusionGetBit`.
- This PR's `fogCutVoxelSolid`: GLSL `{ivec4 occlusionWorldOrigin; uint
  occlusionBits[];}` and Metal `{int4 worldOriginVoxel; uint bits[1];}` are
  layout-identical to both, and the flat-index expression
  `(uint(lz+he)*256 + uint(ly+he))*256 + uint(lx+he)` with `bits[idx>>5] >>
  (idx&31)` matches `voxelOcclusionGetBit` exactly. It reads the **voxel-
  existence** bitfield (base index 0, no `kLightOcclusionBitfieldUintCount`
  offset) — the correct "is there solid matter" field, not the blocker field.
  No transcription drift.

**2. Binding aliasing + buffer lifetime — safe.** `kBufferIndex_FogObservers
== 27` and `kBufferIndex_LightOcclusionGrid == 28` (confirmed in
`ir_render_types.hpp`), both agreeing with every shader `binding=` /
`[[buffer()]]`. Because the grid buffer is created once at fixed size and
exposed as a stable named resource, caching `occlusionGridSsbo_` after a single
lazy resolve cannot dangle (no re-create / resize path exists). The
placeholder path — binding the 144-byte observer UBO as the slot-28 SSBO when
no grid system registered — is never dereferenced: `fogCutSolidsAvailable == 0`
gates every read, and the shader early-outs of the cut block before touching
`occlusionBits`/`bits`. `Buffer::bindBase` is target-agnostic per call, so
binding a `UNIFORM`-created buffer as `SHADER_STORAGE` for one dispatch is legal
GL; Metal never validates an unread buffer's size. Correct on both backends.

**3. Ray-vs-disc entry math — correct.** Standard quadratic
`a·t² + b·t + c` with `a = |rayStep.xy|²`, `b = 2·d0·rayStep.xy`,
`c = |d0|² − r²`; entry root `t = (−b − √det)/(2a)` is the near intersection
since `a > 0`, guarded on `a > 0` (degenerate vertical ray → falls through to
the plain mask) and `det ≤ 0` (miss). `bestT` correctly selects the nearest
non-negative crossing across hard discs, and the `w != 0` skip keeps the ray
loop and the `hardState` rim shortcut consistent (both scoped to Mode-A hard
discs). The two-sample occupancy read (half-cell + 1.5-cell interior) with the
`bestT·worldPerDepthUnit ≤ 1` "still inside the surface voxel" shortcut is a
sound way to dodge rim-cell quantization.

**4. Cull-margin superset holds across both backends.** Stage-1's
`kFogHiddenKeepCells (8) + kFogColumnKeepAa (0.5) = 8.5` keep ring is a subset
of `c_voxel_visibility_compact`'s `kCullSafetyCells` bumped `1.0 → 9.0`, so no
ring voxel the cut needs is culled before stage 1 can keep it. GLSL and Metal
twins mirror the constants and the guarded cut block structurally; remaining
text differences are the legitimate per-stage×backend `emitDeformedFace` /
buffer-threading API surface.

### For the human (non-blocking)
- **Perf.** `kFogHiddenKeepCells = 8` widens the rendered ring around every
  vision circle (≈8 cells of previously-culled hidden geometry now rasterized,
  lit, and AO/shadow-participating near the rim), and the per-pixel cut adds a
  ≤8-circle quadratic solve + up to 2 occupancy taps on each fog-dark
  unexplored pixel. Both are design-intended (the cut needs lit colour and
  surviving geometry to work from) and bounded, but there's still no committed
  fog-scene perf baseline — worth a frame-time glance once one lands.
- **Frame ordering.** The cut reads the occlusion grid built by
  `BUILD_LIGHT_OCCLUSION_GRID`; if that system is ordered after `FOG_TO_TRIXEL`
  in a given pipeline the cut samples the previous frame's occupancy (1-frame
  lag near moving geometry). Cosmetic, not a correctness blocker — flagging for
  visibility only.
- **GL parity.** The 4-way (2 shaders × 2 backends... here 3 shaders × 2)
  duplication plus the fresh binding-28 declaration means the OpenGL walls
  should be eyeballed once on the smoke host, not assumed from the macOS
  references. `fleet:needs-windows-smoke` is set; the PR body's `linux-debug`
  reference re-bless remains its own open item.

No blockers, no needs-fix.

Issue #1933, opened 2026-06-21
render: conservative rasterization / MSAA on the per-axis scatter pass — root fix for corner spikes + silhouette dashing


jakildev:
**Model:** opus
**Related:** #1883 (per-axis corner spikes — this supersedes the manual-dilation approach), epic #1881, #935 / #937 (rasterization-quality epics).
**Blocked by:** (none) — independent root-fix effort; large/epic-sized, carve children during planning.

## Why
The per-axis voxel scatter pass loses sub-pixel coverage on foreshortened/rotated faces. The current mitigation is a manual conservative-dilation **margin** (`scatterConservativeDilation`, `0.5·|n|`) + miter limit. PR #1907 (iters 1+2) tamed the dashing/slivers, but the near-cardinal **corner spikes** on coarse cubes proved to have **no local fix** (#1883/#1917): crisp corners need `M·C ≲ 2px` while foreshortened coverage forces `C ≈ half-cell` — mutually exclusive. The manual margin is fighting a sub-pixel-rasterization problem with a geometry hack.

## Scope (root fix)
Replace the manual dilation/margin coverage model with **hardware conservative rasterization** (or MSAA / analytic coverage) on the scatter pass, so a face's coverage is correct at the source and no per-cell margin is needed. This eliminates corner spikes AND silhouette dashing together, and removes a hot-path geometry hack.

## Approach (planner carves children)
- Survey backend support: GL `GL_NV_conservative_raster` / `GL_INTEL_conservative_rasterization` / core 4.x options vs Metal's coverage/`MTLRasterizationRateMap` / programmable sample positions; pick a portable model or a per-backend implementation with a shared CPU contract.
- Define the coverage contract so the per-axis scatter footprint is rasterized conservatively without the `scatterConservativeDilation` margin; retire the margin + miter once coverage is correct.
- Validate against the #1922 jitter harness + the epic #1881 `perf-grid-rotate-sweep` solidity/silhouette gate: corner spikes gone on coarse cubes, no silhouette dashing, no coverage regression, both backends.

## Acceptance criteria
- Coarse-cube (3³/12³) near-cardinal poses render crisp corners (no spikes) AND solid foreshortened faces (no dashing) — without the manual dilation margin.
- Both backends (Metal + OpenGL) at parity.
- No regression on the epic #1881 rotated-solidity baseline; cardinal fast path preserved.

## References
[...]

Issue #1808, opened 2026-06-13
[Epic] Voxel render perf — solid-volume cost, rotation invariance & advanced rendering


jakildev:
**PLACEHOLDER — design parked, NOT yet plannable.** Captures the considerations from the 2026-06-13 architect perf session so we can plan at a later date once prerequisites land. Do not decompose/queue yet.

## The goal
100 FPS (10ms/frame) at 262K entities across **all content shapes and camera states** at zoom8. Today zoom1 hits ~113 FPS; zoom8 is the gate.

## What we measured this session (cardinal orientation, M4 Max, 262K, post-#1778)
| scene @ zoom8 | voxelStage1 GPU | occludedExposedFraction | covered by |
|---|---|---|---|
| voxel_set (many small entities) | 32ms | **0.97** (38× overdraw) | #1294 occlusion cull |
| dense_set (one solid volume) | **49.9ms** (slowest) | **0.00** | **nothing yet** ← this epic |
| hollow_set (sphere shell) | 6.6ms | 0.50 | partial |

`dense_set` is the stand-in for **solid/terrain content**: slowest, yet **0% cullable** — the exposed-face mask (#1278) already removed all hidden faces, so every emitted tap is genuinely visible. Occlusion culling (#1294), CPU update parallelism (#1052), and the exposed-face mask all miss it. Its cost is pure productive surface rasterization, and it scales with voxel count.

## The two concerns converge: scatter vs gather
dense_set's cost AND camera-yaw cost are the same underlying issue.

| | current: per-voxel **scatter** | per-pixel **gather** (raymarch/DDA) |
|---|---|---|
| cost scales with | voxel count → dense_set bad | visible trixels → dense_set ≈ sparse |
| camera yaw | non-cardinal drops to per-axis scatter path (#1739), more expensive | angle-invariant — ray just rotates |
| meshes | none | none |
| trixel-level output | yes | yes (each trixel = one ray → same distance/color/id texture) |

## Hard constraints (from the session)
- **No meshes** (greedy meshing ruled out).
- **Calculations stay at the trixel level** even with subdivisions (subdivisions trixel-granular; nothing sub-trixel or geometry-based).
- **Preserve the distance/color/id texture contract** so lighting / shadow / AO stages are untouched (a gather/raymarch must still write the same trixel textures).
- Ideal: **yaw-invariant cost** — rotating the camera should not change the winning-trixel set's cost, only its scalar orientation.

[...]

PR #2190, opened 2026-07-02
render: both-exposed silhouette-riser dual emit for rotated cubes (#2157)

### LINKED ISSUE #2157: render: rotated GRID/re-voxelize cubes — residual non-cardinal gaps + dark riser shading after the silhouette-riser fix

jakildev:
## Summary
Phase A of #2157 (post-#2162 residual root-cause), executed per the plan's decision-gate method. Three findings, one shipped fix:

1. **Shipped — both-exposed dual emit (cardinal subdivided path, GLSL+Metal, stage 1+2).** The #2162 flip fills a riser only when the slot's triplet face is *occluded*; a rotated staircase EDGE cell has BOTH polarities of the axis exposed, so the flip never fires and the opposite face plane (polarity-dependent via `faceMicroPositionFixed6`) was never emitted → enclosed-black riser pixels at cardinal camera directions. Emit both planes for such cells, gated to `rotatedEmit` content.
2. **Root-caused + deferred to #2207 — polarity-blind lighting decode.** The dominant *visual* residual (venetian bands at 180°, the issue's "dark riser shading") is NOT missing geometry: `--debug-overlay unlit` shows the rows solid while `--debug-overlay shadow` shows them fully sun-shadowed. Every lighting consumer reconstructs `faceId = visibleFaceIds[slot]` from the 2-bit slot, so #2162-flipped faces get an INVERTED normal (Lambert→0, shadow bias into the caster, AO sampling inverted, #2088 self-step suppression blind to the flip). Needs a polarity-carrier design call — options + evidence in #2207.
3. **Root-caused + deferred to #2207 — per-axis store can't carry opposite polarity.** The plan's fork (a) is CONFIRMED as a real mechanism, but the naive fix fails: dual-tapping the per-axis store measured **net-worse** (112.5° 180→240 px, 247.5° 236→356 px) because the forward scatter reconstructs every cell as triplet polarity and mis-places the riser quad. The non-cardinal residual (~0.4–1.2%) therefore stays until the polarity carrier lands; it is NOT primarily the #1933 scatter-coverage class.

## Measurements (enclosed-black, `render-coverage-metric.py`, frozen-pose gridspin sweep, deterministic/byte-identical across runs)

| direction | master | this PR |
|---|---|---|
| 0° / 360° | 0 / 0 | 0 / 0 |
| 90° | 0.28% (64 px) | **0.07% (16 px)** |
| 270° | 0.24% (56 px) | **0.03% (8 px)** |
| 180° | 1.98% (472 px) | **1.77% (424 px)** (remainder → #2207) |
| non-cardinals | 0.10–1.25% | unchanged (12/13 frames byte-identical; 4 exact-diagonal frames drift by the equal-depth tie-winner scheduling class, identical hole metrics) |

Poses 0.3 / 0.9 improve the same way (270° pose 0.3: 240→112 px); no direction regresses at any pose.

## Visual
Before/after at 270° + drift map (fix pixels only), zoom4 crops in [`docs/pr-screenshots/claude/2157-noncardinal-riser-residual/`](https://github.com/jakildev/IrredenEngine/tree/claude/2157-noncardinal-riser-residual/docs/pr-screenshots/claude/2157-noncardinal-riser-residual):

| before | after | img_diff (1320 px changed) |
|---|---|---|
| before | after | drift |

The remaining dark banding visible in "after" is finding 2 (lighting, #2207), not coverage.

## Test plan
- [x] `IRCanvasStress --only gridspin --frozen-pose {0.3,0.6,0.9} --sweep-yaw 0 6.2832 17`: cardinal enclosed-black drops per the table; non-cardinals unchanged (Metal)
- [x] Determinism: repeated sweep byte-identical run-to-run
- [x] Axis-aligned `--no-spin` sweep: **17/17 byte-identical** vs master (single-canvas + per-axis routes)
- [x] shape_debug 21-shot suite: 19/21 byte-identical; the 2 yaw-45 shots drift ≤24 px / ≤0.0007% (equal-depth tie-winner scheduling class — same class drifts between ANY two builds of this shader family)
- [x] Builds clean: IRCanvasStress + IRShapeDebug + IrredenEngineTest (Metal)
- [ ] GL parity smoke (GLSL mirrors Metal textually; needs a Linux host)

Phase B routed: root cause is the #2207 polarity carrier (not the #2089 AO resample the plan guessed). Phase C (perf): the extra emit fires only for rotated staircase edge cells on the cardinal subdivided path — a strict subset of the existing reVoxelize 5-tap dilation cost; formal dense-budget measurement folds into #1808 per the plan.

Closes #2157


jakildev (commented):
## Review — render: both-exposed silhouette-riser dual emit for rotated cubes (#2157)

**Verdict:** approve

### Checks performed
- Read all four changed shader files in full (not just the hunks): `c_voxel_to_trixel_stage_{1,2}.glsl` and their Metal twins, plus the `.fleet/plans/issue-2157.md` plan doc and `faceIsExposed`/`faceMicroPositionFixed6` in `ir_iso_common.glsl`.
- **Flip/dual-emit interaction traced through by hand.** `bothPolaritiesExposed` is computed *after* the existing #2162 flip (`faceId = faceId ^ 1` when the triplet face is occluded but its opposite is exposed). Working through both branches: if the flip fires, `bothPolaritiesExposed` evaluates to `false` (the post-flip opposite is provably not-exposed, since that's why the pre-flip face wasn't chosen) — no double-emit. If the flip does *not* fire because the triplet face is already exposed, `bothPolaritiesExposed` is true exactly when the opposite polarity is *also* exposed — i.e. exactly the both-exposed-edge-cell case the PR describes, and the existing single emit + this new second emit combine correctly with no overlap.
- Confirmed `faceId ^ 1` is a valid same-axis polarity toggle: `kFaceXNeg=0/kFaceXPos=1/kFaceYNeg=2/kFaceYPos=3/kFaceZNeg=4/kFaceZPos=5` in `ir_iso_common.glsl:25-30`, so bit 0 is polarity on every axis — same trick the pre-existing flip already relies on.
- Verified the dual-emit tail is placed only in the cardinal-subdivided path (`voxelRenderOptions.x != 0`, after both the `perAxisRoute != 0` and the base-resolution `voxelRenderOptions.x == 0` early returns) in all four files — matches the PR's claim that the per-axis store and base path are deliberately excluded.
- Diffed stage 1 vs stage 2 vs both Metal twins side-by-side: the `bothPolaritiesExposed` predicate, the opposite-face position recompute (rotate/shift/depth/base), and the `emitDeformedFace` call are byte-for-byte structurally identical across all four (color tap additionally threads `voxelColor`/`packedEntityId`, as expected). Metal's `emitDeformedFace` call sites match the 14-arg signature (`base, D, distance, color, id, localId, isDetached, faceId, reVoxelize, canvasSize, distanceScratch, colors, distances, entityIds`) in the same order as the primary emit — no argument-order slip.
- The gate (`rotatedEmit`) is unchanged from #2162 and still requires `reVoxelize || kRotatedEmit`, so axis-aligned/static content takes zero extra emits — consistent with the PR's byte-identical `--no-spin` sweep claim (17/17).
- Confirmed this PR does *not* touch any lighting/AO/shadow system files, so the lighting checklist items don't apply here — the PR body's own finding #2 (polarity-blind lighting decode) is a pre-existing #2162-era issue this PR doesn't worsen (the new opposite-face tap still encodes only `slot` via `encodeDepthWithFace(depthOpposite, slot)`, same as the primary tap), and is correctly deferred to #2207 with rationale rather than silently left.
- No serialization, ECS, or ownership/lifetime surface touched (shader-only diff).

### Nits
None.

### Praise
- The decision-gate method in `.fleet/plans/issue-2157.md` — isolating (a) flip interaction vs (b) #1933 per-axis coverage by temporarily gating the flip off on the per-axis route and cross-checking against a non-rotated foreshortened solid — is a clean way to root-cause a shading residual without guessing, and the PR body reports the negative result (dual-tapping the per-axis store measured net-worse) instead of silently dropping that avenue.
- Deferring the lighting-decode and per-axis-store fixes to #2207 with the specific mechanism (`faceId = visibleFaceIds[slot]` reconstruction is polarity-blind) rather than papering over the visible banding in this PR keeps this diff scoped and reviewable.

- [ ] GL parity smoke — GLSL mirrors Metal textually per this review, but wants an actual Linux/OpenGL run before merge (flagging via smoke label below).

Escalation: please Opus-recheck before merge (touches: engine/render/ shader pipeline).

Opus recheck required: touches `engine/render/src/shaders/` (both GLSL and Metal) — per this role's checklist, render-pipeline shader changes always get a second pass even when the reviewer's own trace finds no issue.

jakildev (commented):
## Opus recheck — render: both-exposed silhouette-riser dual emit for rotated cubes (#2157)

**Verdict:** approve

Sonnet did a thorough first pass (flip×dual-emit interaction traced by hand, per-axis exclusion confirmed, all four files diffed side-by-side, `faceId^1` polarity-toggle validated). This recheck focused on the render-pipeline invariants Sonnet's trace couldn't fully close: predicate/emit consistency across the intervening early returns, the stage-1↔stage-2 depth/colour mirror, and atomicMin ordering between the primary and opposite emits.

### Opus-only checks

- **`slot` / `faceId` stability between predicate capture and tail emit** — `slot` is `const int` (stage 1 :277 / stage 2 :196), never reassigned. `faceId` is assigned only at init (:278/:197) and the pre-existing #2162 flip (:329/:218); the predicate is captured immediately after the flip and `faceId` is never reassigned again before the tail emit, so `oppositeFaceId = faceId ^ 1` at the tail is provably the opposite of the face the predicate validated. The predicate `faceIsExposed(faceId) && faceIsExposed(faceId^1)` is also **symmetric** under `faceId ↔ faceId^1`, so it's robust even if the ordering assumption were weaker. Sonnet flagged the flip-before-predicate ordering as safe; on this closer read I confirm — and the symmetry makes it doubly so.

- **Per-axis exclusion is genuine, not just commented** — the per-axis route returns at `(faceId>>1)!=axis` (:481/:278) and the base-resolution path returns earlier still; the dual-emit tail sits after all of these. So the "PER-AXIS store cannot take a second tap, deferred to #2207" claim is structurally enforced, not merely asserted. Confirms Sonnet's placement finding.

- **Stage-1↔stage-2 mirror** — identical `slot`/`faceId`/`bothPolaritiesExposed` derivation and identical early-return ladder mean the *set* of cells reaching the dual emit is the same in both stages. Since `microOpposite`/`depthOpposite`/`baseOpposite` are computed byte-for-byte identically, stage 2's colour tap lands exactly on the distance tap stage 1 wrote — no orphaned distance (black pixel) or orphaned colour. This is the invariant that matters most for a two-stage split and it holds.

- **atomicMin ordering (primary vs opposite)** — `slot` is shared (same axis), only the depth differs: the opposite plane is the far face, so `depthOpposite > depthBase` and it loses the `atomicMin` on any pixel the near face already covers. It therefore only fills the silhouette-riser pixels the front face leaves as background (exactly because `faceMicroPositionFixed6` is polarity-dependent). No risk of the back face overpainting the front.

- **Metal arg order** — the opposite `emitDeformedFace` calls match the primary emit's argument order in both Metal stages (9-arg stage 1, 14-arg stage 2). Consistent with Sonnet's side-by-side.

### Nits
None.

### Merge gate
- GL parity: shader-only change, GLSL mirrors Metal textually — `fleet:needs-windows-smoke` (Windows/OpenGL is the primary GL validation host) already covers the Linux/GL parity concern Sonnet flagged. No further tagging needed.

PR #2208, opened 2026-07-04
engine/voxel: dedup GRID occupancy mask + depth-probe pixel parse

### LINKED ISSUE #2160: voxel/demos: dedup pre-existing face-occlusion + pixel-arg parse duplication (simplify reuse findings)
Surfaced by the `simplify` reuse fan-out during the rotated-voxel gap-fix PR (branch \`claude/diagnose-gridspin-gaps\`). Both are **pre-existing** duplication (not introduced by that PR), so they were left out of its scope. Filing for the coding-improvement triage.

## 1. \`recomputeMaskFromGlobals\` re-implements \`recomputeFaceOccupancyOnCells\`
\`system_rebuild_grid_voxels.hpp::recomputeMaskFromGlobals\` (~line 563) is ≈85% structural overlap with \`IRPrefab::Voxel::recomputeFaceOccupancyOnCells\` (\`face_occupancy.hpp:132\`): both build a packed-key occupancy set from active voxels, then derive the six-neighbor \`kFaceOccluded{Neg,Pos}{X,Y,Z}\` bits with the same offsets and the same \`flags_ & ~kFaceOccludedMask | face\` write-back. The sibling \`system_rebuild_detached_voxels.hpp\` already includes \`face_occupancy.hpp\` and calls the shared function; this system instead re-implements it (and re-implements cell-key packing as a private \`packCell\` vs the existing \`IRPrefab::Voxel::detail::packCellKey\`). Difference: sorted-vector + \`binary_search\` vs \`unordered_set\`, and positions rounded from \`poolGlobals\` rather than a pre-rounded cells span. Fix: pre-round into a \`cellsScratch_\` ivec3 span (as the detached system does) and call \`recomputeFaceOccupancyOnCells\`, or extend that helper to accept a sorted-vector scratch.

## 2. \`--depth-probe\`/\`--depth-probe-assert\` X,Y parse duplicated across demos
The \`sscanf("%d,%d")\` → \`{set, ivec2}\` block is duplicated across \`canvas_stress\` and \`perf_grid\` (\`applyCliArgs\`). The within-\`canvas_stress\` double (\`--depth-probe\` + \`--depth-probe-assert\`) was already deduped into a file-local \`parsePixelArg\` helper in the gap-fix PR; the **cross-demo** copy remains. Fix: promote a shared \`parsePixelArg\` (or an X,Y-coordinate IRArgs accessor) into a common home (e.g. \`depth_probe.hpp\` alongside the probe system, or a small \`IRArgs\` convenience) and call from both demos.

Both are small, mechanical, and behavior-preserving.

jakildev:
## Summary
- `REBUILD_GRID_VOXELS::recomputeMaskFromGlobals` now calls the shared `IRPrefab::Voxel::recomputeFaceOccupancyOnCells` — pre-rounding pool globals into member `cellsScratch_`/`occupancyScratch_` and delegating (mirroring `system_rebuild_detached_voxels`), dropping the re-implemented `packCell` + sorted-vector + `binary_search` occupancy pass.
- Promoted a shared `IRPrefab::DepthProbe::parsePixelArg` into `depth_probe.hpp`; `canvas_stress` and `perf_grid` now call through it instead of each hand-rolling the `--depth-probe "X,Y"` `sscanf` parse.
- Dropped now-unused `<algorithm>` (grid system) and `<cstdio>` (perf_grid).

Both are pre-existing duplication surfaced by the simplify reuse fan-out; behavior-preserving.

## Verification (byte-identity)
- `canvas_stress` base-vs-dirty `--auto-screenshot` compare (`--no-spin --no-auto-rotate`): **10/12 shots byte-identical**. The 2 differing shots (01, 04) also differ **base-vs-base** (two origin/master runs) — pre-existing canvas_stress non-determinism, not this change (dirty-vs-base2 even matched 11/12).
- By construction: grid `packCell` ≡ the helper's `packCellKey` for any in-scene cell, and set-membership is identical (sorted-vector `binary_search` ≡ `unordered_set`), so the exposed-face bits are unchanged.
- Build: `IRCanvasStress` + `IRPerfGrid` clean (macOS/Metal).

## Test plan
- [ ] Confirm the GRID occupancy dedup preserves the exposed-face mask (packCell↔packCellKey equivalence; the identity arm + rotated arm both feed the shared helper).
- [ ] Confirm `parsePixelArg` placement in `depth_probe.hpp` + the two demo call-throughs; the `--depth-probe-assert X,Y,tier=N` superset parse deliberately stays local to canvas_stress.

## Follow-up (out of scope for #2160)
- simplify's call-sequence pass flagged that the ~10-line resize+round-loop *wrapper* still overlaps between the GRID and DETACHED systems (both already share the low-level `recomputeFaceOccupancyOnCells`). Pulling that wrapper into a position-callable helper would touch `face_occupancy.hpp` + the detached system (beyond this issue's grid-only scope) — left as a potential follow-up.

Closes #2160

### DIFF
--- a/creations/demos/canvas_stress/main.cpp
+++ b/creations/demos/canvas_stress/main.cpp
@@ -1018,20 +1018,6 @@ void readConfig() {
         g_settings.autoRotate_ = autoRotate.as<bool>();
 }
 
-// Parse a "--depth-probe[-assert] X,Y" value into (set, pixel), mirroring the
-// warn-on-malformed behaviour the hand-rolled parser had — a bad value leaves
-// `set` false so the probe stays off rather than reading a garbage pixel.
-void applyDepthProbe(const std::string &value, bool &set, ivec2 &pixel, const char *flag) {
-    int px = 0;
-    int py = 0;
-    if (std::sscanf(value.c_str(), "%d,%d", &px, &py) == 2) {
-        set = true;
-        pixel = ivec2(px, py);
-    } else {
-        IR_LOG_WARN("{}: expected X,Y (e.g. 960,540); ignoring '{}'", flag, value);
-    }
-}
-
 // Parse "--depth-probe-assert X,Y" (depth-write guard, tier -1) or
 // "X,Y,tier=N" (per-trixel-priority tier guard, tier >= 0). The 3-field form is
 // tried first; a bad value leaves the assert off so the run stays byte-identical.
@@ -1182,12 +1168,15 @@ void applyArgs() {
             IRRender::debugOverlayModeFromString(args.getEnum("--debug-overlay").c_str());
     }
     if (args.wasProvided("--depth-probe")) {
-        applyDepthProbe(
-            args.getString("--depth-probe"),
-            g_settings.depthProbeSet_,
-            g_settings.depthProbePixel_,
-            "--depth-probe"
-        );
+        ivec2 pixel;
+        if (IRPrefab::DepthProbe::parsePixelArg(
+                args.getString("--depth-probe"),
+                pixel,
+                "--depth-probe"
+            )) {
+            g_settings.depthProbeSet_ = true;
+            g_settings.depthProbePixel_ = pixel;
+        }
     }
     if (args.wasProvided("--depth-probe-assert")) {
         applyDepthProbeAssert(args.getString("--depth-probe-assert"));
--- a/creations/demos/perf_grid/main.cpp
+++ b/creations/demos/perf_grid/main.cpp
@@ -59,7 +59,6 @@
 #include <irreden/common/command_suite_capture.hpp>
 
 #include <algorithm>
-#include <cstdio>
 #include <numbers>
 #include <string>
 #include <string_view>
@@ -553,16 +552,16 @@ void readCliArgs() {
         }
     }
     if (args.wasProvided("--depth-probe")) {
-        // `--depth-probe X,Y` (#1910): IRArgs has no pair type, so register as a
-        // string and keep the demo-side comma split.
-        const std::string probe = args.getString("--depth-probe");
-        int px = 0;
-        int py = 0;
-        if (std::sscanf(probe.c_str(), "%d,%d", &px, &py) == 2) {
+        // `--depth-probe X,Y` (#1910): IRArgs has no pair type, so the comma
+        // split lives demo-side (shared IRPrefab::DepthProbe::parsePixelArg).
+        ivec2 pixel;
+        if (IRPrefab::DepthProbe::parsePixelArg(
+                args.getString("--depth-probe"),
+                pixel,
+                "--depth-probe"
+            )) {
             g_cliOverrides.depthProbeSet_ = true;
-            g_cliOverrides.depthProbePixel_ = ivec2(px, py);
-        } else {
-            IR_LOG_WARN("--depth-probe: expected X,Y (e.g. 960,540); ignoring '{}'", probe);
+            g_cliOverrides.depthProbePixel_ = pixel;
         }
     }
 }
--- a/engine/prefabs/irreden/render/depth_probe.hpp
+++ b/engine/prefabs/irreden/render/depth_probe.hpp
@@ -10,6 +10,9 @@
 #include <irreden/render/components/component_trixel_framebuffer.hpp>
 #include <irreden/render/ir_render_types.hpp>
 
+#include <cstdio>
+#include <string>
+
 // GPU→CPU composite-depth probe. A debug-only readback that, for a requested
 // screen pixel, reads the REAL depth-test value stored in the main framebuffer's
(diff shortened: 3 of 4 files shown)


jakildev (commented):
## Review — engine/voxel: dedup GRID occupancy mask + depth-probe pixel parse

**Verdict:** approve

Pure dedup, behavior-preserving per the PR body: `REBUILD_GRID_VOXELS` drops its
private `packCell`/sorted-vector/`binary_search` occupancy pass in favor of the
shared `IRPrefab::Voxel::recomputeFaceOccupancyOnCells` (already used by
`system_rebuild_detached_voxels`), and a new `IRPrefab::DepthProbe::parsePixelArg`
replaces the `--depth-probe "X,Y"` `sscanf` duplicated across `canvas_stress` and
`perf_grid`.

### Checks performed

- **`packCell` ↔ `packCellKey` equivalence.** Read both implementations
  side-by-side (`system_rebuild_grid_voxels.hpp` pre-image vs.
  `engine/prefabs/irreden/voxel/face_occupancy.hpp:24-30`). Old: additive
  bias-and-multiply by `kSpan = 2^21`. New: bias-and-mask-then-shift by 21
  bits. For any in-range cell (`|c| < 2^20`, the documented working-volume
  bound both versions rely on) these produce byte-identical keys — masking
  by `2^21-1` after a non-negative bias is the same operation as the
  additive carry when no term overflows its 21-bit field. Confirmed no
  behavior change for the realistic case; both versions alias identically
  (and equally silently) outside the documented bound, so this isn't a new
  edge case introduced by the dedup.
- **Occupancy semantics preserved.** Old code populated `cellKeys_` only for
  active voxels (`alpha_ == 0` → `continue`) and only computed face bits for
  active voxels. `recomputeFaceOccupancyOnCells` (`face_occupancy.hpp:144-148,
  156`) applies the identical `voxelIsActive` gate on both the occupancy-set
  build and the face-bit computation — the dedup doesn't change which voxels
  contribute to occupancy or receive face bits.
- **`cellsScratch_` is now populated for every slot, not just active ones**
  (`system_rebuild_grid_voxels.hpp`, the new loop drops the old `continue` on
  inactive voxels before rounding). This is harmless — `roundVec3HalfUp` is
  computed unconditionally for both active and inactive slots now, but
  `recomputeFaceOccupancyOnCells` re-applies the `voxelIsActive` gate before
  ever reading a cell for occupancy or face computation, so an inactive
  voxel's now-computed-but-unused rounded position never reaches the
  occupancy set. Extra rounding work for inactive slots is negligible
  (`roundVec3HalfUp` is cheap, no allocation), not a hot-path regression.
- **`<algorithm>` removal is safe** — grepped the file post-diff for
  `std::sort`/`std::binary_search`/`std::find`; no remaining call sites.
- **`parsePixelArg` call-site parity.** Both demos already included
  `depth_probe.hpp` before this diff (grepped both files — the include was
  pre-existing, used for `logCompositeDepth`/`assertCompositeDepthTier`), so
  no new include needed. The `set`/`pixel` mutation-on-success-only contract
  matches the old `applyDepthProbe` local helper exactly (failure path logs
  and leaves the output untouched in both).
- **No `SystemName` enum change needed** — `REBUILD_GRID_VOXELS` is an
  existing system specialization, not a new one.
- **Scratch members renamed, not orphaned** — `cellKeys_`/
  `cellRoundedPositions_` fully removed, replaced by `cellsScratch_`
  (`std::vector<ivec3>`) + `occupancyScratch_` (`std::unordered_set<int64_t>`,
  owned by this system, passed by reference into the shared helper so no
  per-frame allocation once warmed — consistent with the "Allocations in hot
  paths" rule).

### Build + smoke verification (done locally, not just taking the PR body's word)

- `fleet-build --target IRCanvasStress` — clean (macOS/Metal, this reviewer's
  host, same backend the PR was authored on).
- `fleet-build --target IRPerfGrid` — clean.
- `IRCanvasStress --auto-screenshot --no-spin --no-auto-rotate` — ran to
  completion, saved all 12 shots, no crash.
- Did not independently re-run the base-vs-dirty byte-identity compare the PR
  body reports (10/12 identical, 2 pre-existing-noise); the equivalence
  argument above (packCell↔packCellKey, occupancy-gate parity) is a stronger
  guarantee than a screenshot diff for this change anyway.

### Praise

- Routing the GRID identity-arm through the same helper
  `system_rebuild_detached_voxels` already uses closes a class of drift risk
  outright — any future tweak to occupancy semantics (e.g. a neighbor-cell
  rule change) now has exactly one implementation to update, not two that can
  silently diverge.
- The PR body's "Follow-up (out of scope)" section correctly identifies that
  the wrapper resize+round loop still overlaps between GRID and DETACHED and
  scopes that out rather than creeping the diff — the right call for a
  behavior-preserving dedup PR.

- [x] Already covered — build + byte-identity compare in the PR body,
      independently re-verified above (build + smoke run).

Opus recheck not required.

PR #2209, opened 2026-07-04
render: render-verify baseline for the GPU light-volume system

### LINKED ISSUE #481: Render: add unit/baseline coverage for GPU light-volume system
## Context

Phase 1a (#359, PR #448) replaced the CPU 6-connected BFS light-volume
implementation with a GPU dilation chain and removed the previous
`test/render/light_volume_test.cpp` (95 lines, 4 GTests:
occluder-blocks-cells, Euclidean falloff, spot cone rejection, spot
occluder).

Straight deletion was defensible — the deleted tests exercised CPU-only
`fillPointLight`/`fillSpotLight` helpers that no longer exist on the GPU
path. But the new `System<COMPUTE_LIGHT_VOLUME>` and the
`c_seed_light_volume` / `c_propagate_light_volume` shaders now have **zero
unit-test coverage**. Correctness rests entirely on manual screenshot
capture (`render-debug-loop`) plus the `IRPerfGrid` perf delta surfaced
in PR #449.

Flagged in opus-reviewer pass on PR #449 as a non-blocking follow-up
(2026-05-07T07:00:49Z).

## Two viable approaches

**Option A — CPU-mockable propagate-kernel reference impl.**
Constexpr/templated implementation of the dilation rule (one-step
distance-tracked Manhattan propagation with occlusion). A unit test
seeds a small grid by hand, runs N steps, and asserts equality against a
GPU debug readback. Catches algorithmic regressions in the propagate
kernel without exercising real GPU hardware.

**Option B — render-baseline crop pair.**
Add a directory under `engine/render/tests/render-baselines/light_volume/`
with current+baseline crops capturing the light-volume contribution at
expected radii (single point light, two-light interference, occluded
light blocked by a wall). Tied into the existing `render-verify` harness
so visual drift fails CI.

Option B is closer to how the rest of `render-baselines/` is organized.
Option A catches a strictly different class of bug (algorithmic vs
visual). Both are useful but Option B is the lower-effort first step.

## Suggested model

`[sonnet]` — bounded scope (new test file + a few baseline images), no
architectural decisions required if Option B is chosen.

## Suggested area

`engine/prefabs/irreden/render/systems`, `engine/render/tests/render-baselines`

## Acceptance criteria

- New tests/baselines exercise `COMPUTE_LIGHT_VOLUME` and the seed +
  propagate shader pair.
- A regression in either shader (off-by-one in the propagate stencil,
  wrong falloff constant, swapped texture binding) fails the test/baseline.
- Existing `IRShapeDebug --auto-screenshot` flow continues to pass.

## Gotchas

- The propagate kernel is non-deterministic on shared origin texels
  ("later thread wins" — documented in the shader). Choose a baseline
  scene where seed positions don't collide.
- Don't reintroduce the CPU `fillPointLight`/`fillSpotLight` helpers
  just to test against them — Option A's reference impl should be a
  fresh implementation of the documented dilation rule, not a port of
  the deleted code.

Filed by opus-worker-1 from PR #449 nit follow-up.

jakildev:
## Summary
- Adds the first automated coverage for the GPU light-volume path (`COMPUTE_LIGHT_VOLUME` seed + propagate + LOS-occlusion), which had none — a render-verify manifest gating the static single-point-light `IRLightingSdfBlocker` scene against committed `macos-debug` reference PNGs (5 shots, the `kShots[]` prefix through `zoom8`). Canaries: the warm point-light floor gradient and the wall's cold shadow band.
- Suppresses `PERF_STATS_OVERLAY` during `--auto-screenshot` capture in `lighting_demo_scene.hpp` — its live wall-clock FPS / GPU-stage text is non-deterministic and would flake any pixel reference. Present for interactive + `--auto-profile`; scene pixels are unchanged. This also unblocks render-verify for the rest of the lighting family.
- Commits the plan at `.fleet/plans/issue-481.md` (per #1932).

### Two as-built deviations from the reviewed plan
1. **Overlay suppression** — the plan assumed the scene was directly blessable; it was not (the overlay burned non-deterministic timing into every frame).
2. **Positional-prefix shot list** — render-verify maps captured shots to labels by **position** (first-N `screenshot_<index>.png`), not by label, and the demo captures all 6 `kShots`. The plan's literal `[zoom2,4,8]` subset would have mislabeled captures, so the manifest gates the positional prefix `[zoom1, zoom2, zoom4, zoom4_offset_3_5, zoom8]`, excluding only the over-magnified `zoom16` (the plan's stated intent). `kShots[]` is shared across all 15 lighting targets, so editing it was out of scope.

## Test plan
- [x] Scene renders correct light-volume + LOS shadow (visual inspection at zoom4/zoom8: warm point-light floor gradient + cold shadow band behind the wall).
- [x] Byte-deterministic across two clean runs (all shots `cmp`-identical, overlay off).
- [x] `render-verify.py --target IRLightingSdfBlocker --demo lighting` → all 5 PASS at 100% (max_delta 0).
- [x] Perturbation gate test: bumping the point-light intensity 2.2→4.0 flips 4/5 shots RED; reverting returns green — the baseline catches light-volume regressions.
- [ ] `linux-debug` references — unblessed; the demo SKIPs on Linux/OpenGL until a Linux host captures them (`render-verify --update-references`).

Skipped attach-screenshots / render-debug-loop: the modified source is a `common/` header (not in the `src/**` / `main*.cpp` render-trigger set), the change is a render-verify test baseline + a capture-mode overlay gate, and the only visual effect (overlay removal) is verified above and captured by the committed references; render-debug-loop drives shape_debug to diagnose render-pipeline bugs, of which this has none.

Closes #481


jakildev (commented):
## Review — render: render-verify baseline for the GPU light-volume system

**Verdict:** approve

Small, focused addition: a render-verify manifest + committed `macos-debug`
references gating the static single-point-light `IRLightingSdfBlocker` scene,
plus a one-line-of-logic overlay-suppression fix that unblocks it. No
production render-pipeline code touched.

### Checks performed

- **Overlay-suppression guard is correct, not backwards.** Traced
  `g_autoWarmupFrames` end to end: `IREngine::args().autoScreenshotWarmupFrames()`
  (`engine/ir_args.cpp:488`) returns `0` unless `--auto-screenshot` was
  provided, in which case it returns the warmup-frame count (nonzero in
  practice — the manifest uses `10`). So `if (g_autoWarmupFrames == 0)`
  pushes `PERF_STATS_OVERLAY` in interactive mode and `--auto-profile` mode
  (which uses its own independent `g_autoProfileFrames`, untouched by this
  guard) and skips it only during `--auto-screenshot` capture. Matches the
  PR's stated behavior exactly.
- **Positional shot-mapping claim verified against `render-verify.py`.**
  `_collect_shots` (`scripts/render-verify.py:185-196`) takes the manifest's
  `len(shots)` and slices `all_shots[:num_shots]` — first-N positional, not
  by label. `kShots[]` in `lighting_demo_scene.hpp:103-114` has 6 entries in
  the exact order the manifest lists (minus the trailing `zoom16_origin`),
  so the manifest's 5-shot prefix maps correctly onto the demo's capture
  order.
- **Manifest shape matches the established template** — thresholds
  (`per_pixel_tol: 8, match_pct: 99.9, max_delta: 64, psnr_db: 35.0`) are
  identical to `fog_demo`'s manifest, the closest precedent cited in the plan.
- **Reference PNGs are real captures, not placeholders** — all 5
  `macos-debug/*.png` are valid 2560x1440 RGBA, non-trivial file sizes
  (150-212 KB), and visually show the expected lit scene (checked
  `zoom4_origin.png`).
- **CMake target name matches manifest `target`** — `IRLightingSdfBlocker`
  is registered in `creations/demos/lighting/CMakeLists.txt:15`.
- **No `linux-debug/` references prematurely committed** — matches the PR
  body's stated scope (macOS-only bless; Linux SKIPs, doesn't fail, until a
  Linux host captures it separately).
- **Cross-host smoke tagging** — checked the diff's paths against the smoke
  procedure's trigger list (`engine/render/`, shaders, `engine/system/**`,
  CMake files). None match; the only source change is a demo-local header
  outside those paths, and reference-image capture debt is explicitly
  self-tracked in the PR body rather than a smoke-validation gap. No
  `fleet:needs-*-smoke` label added.

### Nits

- `creations/demos/lighting/common/lighting_demo_scene.hpp:385-386` — the
  comment claims this "Mirrors fog_demo / shape_debug, which omit the
  overlay from their gated scenes." Checked both: `fog_demo` (`main.cpp`)
  never adds `PERF_STATS_OVERLAY` to its pipeline at all (unconditional
  omission, not gated), and `shape_debug` (`creations/demos/default/main_lua.cpp:86`)
  adds it **unconditionally** with no gating and has no render-verify
  manifest at all — it isn't a "gated scene" in this sense. Only `fog_demo`
  is actually an applicable precedent. Suggest trimming the comment to just
  "Mirrors fog_demo, which omits the overlay entirely" so a future reader
  doesn't go looking for gating logic in `shape_debug` that isn't there.

### Praise

- The as-built deviations section in `.fleet/plans/issue-481.md` documents
  exactly why the plan's literal `[zoom2,4,8]` shot subset wouldn't have
  worked (positional mapping, not label mapping) — saved this reviewer from
  re-deriving that same trap independently.

- [x] Already covered by the PR body (visual inspection, byte-determinism,
      render-verify green, perturbation-gate red/revert) — no further
      action needed.

jakildev:
Addressed nit: trimmed the overlay-suppression comment's stale `shape_debug` precedent — that demo omits `PERF_STATS_OVERLAY` unconditionally with no render-verify manifest, so only `fog_demo` is an applicable precedent. Now reads "Mirrors fog_demo, which omits the overlay entirely." Build clean. — worker

PR #2210, opened 2026-07-04
engine: one-shot query-command primitive + Command<RANDOMIZE_VOXELS> (#17)

### LINKED ISSUE #17: add command type based on a query
kinda like running a system tick function one time

jakildev:
## Summary
- Add `IRSystem::executeQuery<Cs...>` / `executeQueryDynamic` — the run-now
  counterpart to `createSystem` / `createSystemDynamic` (#17's "running a system
  tick function one time"): same archetype-node traversal + `Exclude<>` filtering,
  but nothing is registered (no `SystemId`, no pipeline slot, no persistent state).
  Serial + main-thread-only; does not flush structural changes.
- Implement `Command<RANDOMIZE_VOXELS>` as the first consumer — recolors every
  active voxel via `editVoxels`, excluding a new `C_Locked` tag component,
  preserving each voxel's alpha and skipping alpha-0 (carved) voxels so the pool
  active-mask stays stable. Promote `RANDOMIZE_VOXELS` out of both `fireByName` /
  `bindPrefabCommand` fall-throughs; `engine/command` PRIVATE-links `IrredenEngineSystem`.
- Docs: `engine/system/CLAUDE.md` "One-shot queries", `engine/command/CLAUDE.md`
  query-command note. Plan committed as the first commit (`.fleet/plans/issue-17.md`).

## Test plan
- [x] `fleet-build --target IrredenEngineTest` green (macOS/Metal).
- [x] New `ExecuteQuery*` (6) + `CommandRandomizeVoxels*` (1) suites pass — dispatch
      shapes, `Exclude<>`, once-per-node, no-registration, alpha-preserve, locked/carved.
- [x] Full suite: 1266 passed, 1 skipped (OpenGL-only GPU compute test).

Closes #17

### DIFF
--- a/.fleet/plans/issue-17.md
+++ b/.fleet/plans/issue-17.md
@@ -1,0 +1,67 @@
+# Plan: add a one-shot query-command primitive (query-based command execution)
+
+- **Issue:** #17
+- **Model:** opus (label `fleet:opus`; plan's advisory `Model:` was sonnet, but the
+  task touches core `engine/system` + `engine/command` + ECS surface — opus lane)
+- **Date:** 2026-07-03 (planned by architect; implemented 2026-07-04)
+
+## Scope
+
+Add a run-once query-execution primitive to `IRSystem` — the run-now counterpart to
+`createSystemDynamic`: same archetype-node traversal, no `SystemId`, no pipeline
+registration, nothing persists. Then implement `Command<RANDOMIZE_VOXELS>` on top as
+the first consumer, honoring the enum's "exclude locked entities" note via a new
+`C_Locked` tag component. Lua exposure of the primitive itself is out of scope;
+`RANDOMIZE_VOXELS` becomes Lua-reachable for free through the existing `fireByName`
+binding.
+
+## Approach
+
+One-shot query surface in `IRSystem` (matches the issue's "running a system tick
+function one time" mental model), traversal shared via
+`IREntity::queryArchetypeNodesSimple` — the same call `executeSystem` makes.
+
+1. **`engine/system/include/irreden/ir_system.hpp`** — add `executeQueryDynamic`
+   (runtime-typed core: `IR_ASSERT_MAIN_THREAD()` + `queryArchetypeNodesSimple` +
+   `body(node)` per node, no SystemManager touch) and `executeQuery<Cs...>`
+   (compile-time-typed wrapper accepting `Exclude<...>`, resolving columns once per
+   node via `getComponentData<Cs>(node)` and row-iterating, dispatching `tick(Cs&...)`
+   or `tick(EntityId, Cs&...)` via `std::is_invocable_v`). Caller thread, serial,
+   main-thread-only. Bodies use `IREntity::deferred*` for structural changes;
+   `executeQuery` does not flush.
+2. **`engine/prefabs/irreden/common/components/component_locked.hpp`** (new) — empty
+   tag `struct C_Locked {};` mirroring `C_Persistent`.
+3. **`engine/prefabs/irreden/voxel/commands/command_randomize_voxels.hpp`** — implement
+   `Command<RANDOMIZE_VOXELS>::create()` running
+   `IRSystem::executeQuery<C_VoxelSetNew, IRSystem::Exclude<C_Locked>>(...)`; per set,
+   `set.editVoxels(...)`: skip alpha-0 voxels, otherwise `IRMath::randomColor()`
+   preserving the original alpha.
+4. **`engine/command/src/ir_command.cpp`** — include new header; promote
+   `RANDOMIZE_VOXELS` to real cases in `bindPrefabCommand` and `fireByName`.
+5. **`engine/command/CMakeLists.txt`** — PRIVATE-link `IrredenEngineSystem`.
+6. **`engine/command/include/irreden/command/ir_command_types.hpp`** — resolve the
+   2023 TODO comment.
+7. **`test/ecs/execute_query_test.cpp`** + **`test/ecs/command_randomize_voxels_test.cpp`**
+   (new) + register in `test/CMakeLists.txt`.
+8. **Docs** — `engine/system/CLAUDE.md` + `engine/command/CLAUDE.md`.
+
+## Acceptance criteria
+
+- `fleet-build --target IrredenEngineTest` green;
+  `IrredenEngineTest --gtest_filter='ExecuteQuery*:RandomizeVoxels*'` all pass.
+- `execute_query_test`: (a) visits entities across archetypes; (b) `Exclude` skips
+  tagged; (c) `EntityId`-form receives correct ids; (d) `executeQueryDynamic` fires
+  once per matched node; (e) no registration (system count unchanged, no re-run in
+  `executePipeline`).
+- `command_randomize_voxels_test` (headless pool + explicit `targetCanvas`): unlocked
+  set gets ≥1 RGB byte changed with every alpha preserved; carved voxel stays
+  inactive + mask unchanged; locked set byte-identical; fireByName no longer hits the
+  unimplemented log path.
+
+## Gotchas
+
+- `randomColor()` returns alpha 255 — preserve original alpha per voxel; skip alpha-0.
+- Use `editVoxels`, never raw `voxels_[i]` writes (GRID-rotation mirror resync, #2165).
+- Edit BOTH switches in ir_command.cpp (duplicate-case compile error guards one side).
+- Main-thread-only, no flush; zero matches is a silent no-op.
+- `IrredenEngineSystem` in engine/command's PRIVATE link block.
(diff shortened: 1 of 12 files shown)


jakildev (commented):
## Review — engine: one-shot query-command primitive + Command<RANDOMIZE_VOXELS> (#17)

**Verdict:** approve

Reviewed against the settled plan (`.fleet/plans/issue-17.md`, committed as the first
commit) — the implementation matches it point-for-point, and I independently verified
the technical claims rather than trusting the plan/PR body.

### Checklist walk-through

- **ECS invariants:** `executeQueryDynamic`/`executeQuery<Cs...>` (`engine/system/include/irreden/ir_system.hpp:412-504`)
  mirror `createSystem`/`createSystemDynamic`'s established `PartitionExcludes` +
  `FilterTags_t` pattern byte-for-byte (compare `ir_system.hpp:138-139` vs `:494,497`),
  so `Exclude<>` markers are correctly stripped from the include archetype rather than
  leaking into the matcher. `IR_ASSERT_MAIN_THREAD()` gates the primitive; no flush is
  attempted, and the contract ("bodies use `IREntity::deferred*` for structural
  changes") is documented at the call site and in `engine/system/CLAUDE.md`.
- **Ownership/lifetime:** `RunQueryRows::run` (`ir_system.hpp:446-478`) resolves
  component columns once per node via `std::tie(getComponentData<Cs>(node)...)` and
  indexes them for the lifetime of that node's loop only — no references escape the
  per-node scope. No raw owning pointers, no `shared_ptr`-where-`unique_ptr` issues.
- **Command dispatch:** `RANDOMIZE_VOXELS` moved out of both `bindPrefabCommand`'s and
  `fireByName`'s fall-through lists into its own `case` in `ir_command.cpp:229-233,326-328`
  — confirmed exactly one `case RANDOMIZE_VOXELS:` per switch (grepped both), so no
  duplicate-case compile error.
- **Command body correctness:** `command_randomize_voxels.hpp` uses the existing
  `editVoxels` API (not raw `voxels_[i]` writes), which resyncs the rotation mirror +
  pool active-mask + face occupancy per its own `#2165` contract — the same pattern
  `carve()` uses. Alpha-0 (carved) voxels are skipped before `randomColor()` overwrites,
  and the original alpha is explicitly restored after — verified `IRMath::randomColor()`
  and `IRMath::seedThreadRng()` both exist (`engine/math/include/irreden/ir_math.hpp`).
- **Serialization:** `C_Locked` is a bare tag with no `// IRAsset: serialized`
  annotation — correctly so; it mirrors `C_Persistent`, which is also unserialized
  (a runtime-only tag registered lazily via `getComponentType`), and `C_VoxelSetNew`
  itself isn't part of the asset-serialization system today, so no `kSaveVersion`
  bump applies.
- **Naming/style:** `C_Locked` (`C_` prefix), `RunQueryRows` detail struct, no `m_`/
  trailing-`_` violations (no new class members introduced — free functions + a tag
  struct). `test/CMakeLists.txt`'s new entries stay alphabetically ordered within
  `ecs/`.
- **Tests:** `execute_query_test.cpp` covers all 5 acceptance-criteria behaviors,
  including the "no registration, no pipeline re-run" invariant that's easy to miss
  when a run-once helper is built on top of a persistent-system code path.
  `command_randomize_voxels_test.cpp` covers unlocked/locked/carved-voxel and
  alpha-preservation in one headless fixture. Author confirmed 1266 passed, 1 skipped
  (OpenGL-only) on the full suite.
- **Build:** `engine/command/CMakeLists.txt` PRIVATE-links `IrredenEngineSystem` with
  an accurate comment explaining why (mirrors the existing PRIVATE-link rationale for
  Rendering/Video/Window).

### Nits

None.

### Praise

- `RunQueryRows`'s per-component / per-entity-id dispatch (`ir_system.hpp:446-478`) is
  effectively a variadic generalization of the pre-existing `forEachComponent`
  (`ir_entity.hpp:112-125`) rather than a fresh reimplementation — same
  `InvocableWithComponents`/`InvocableWithEntityId` concepts, same iteration shape.
  Good reuse instead of a parallel dispatch mechanism.
- The `ir_command_types.hpp` TODO ("should operate like a system with a query...
  exclude locked entities") is resolved literally, not reinterpreted — `C_Locked` +
  `Exclude<C_Locked>` is exactly what the 2023 comment asked for.

- [x] Already satisfied — full suite green, both new suites pass, build confirmed on
  macOS/Metal.

Escalation: please Opus-recheck before merge (touches: engine/system/ core ECS query
primitive, engine/entity/ traversal reuse). No specific concern found — flagging
because this introduces a new foundational primitive other systems/commands will
build on, and the review checklist treats `engine/system/` as a bright-line Opus-only
surface.

Opus recheck required: touches engine/system/ (new core query primitive) and
engine/entity/ traversal reuse.

jakildev (commented):
## Opus recheck — engine: one-shot query-command primitive + `Command<RANDOMIZE_VOXELS>` (#17)

**Verdict:** approve

Sonnet flagged this as an approve-and-escalate with no specific concern, purely
because it lands a new foundational `engine/system/` primitive. I spent the pass
on the three-systems-deep ECS invariants and template metaprogramming Sonnet's
checklist could not fully confirm — reading the primitive against the code paths
it reuses rather than against the plan. All of it holds; confirming approve.

### What I independently verified (beyond Sonnet's pass)

- **Template partition machinery is byte-for-byte the `createSystem` path, not a
  re-derivation.** `executeQuery<QueryComponents...>` (`ir_system.hpp:493-504`)
  builds `excludeArchetype` from `PartitionExcludes<...>::Excluded` and the
  include archetype from `FilterTags_t<...>` — the exact same trio
  `createSystem` uses (`ir_system.hpp:117-118` + the tag-strip block). Critically,
  `RunQueryRows<IncludedList>` is instantiated on the **filtered** include list,
  so column resolution only touches real data components — an `Exclude<C_Locked>`
  marker can never leak into `getComponentData<>` or the archetype matcher.
  Sonnet's "mirrors createSystem" claim holds at the metaprogramming level, not
  just structurally.

- **Row dispatch matches the established `forEachComponent` idiom
  (`ir_entity.hpp:112-125`), and hoists correctly.** `RunQueryRows::run` resolves
  every column once per node via `std::tie(getComponentData<Cs>(node)...)`
  *outside* the row loop, then indexes `cols[i]` per row — the "no getComponent in
  a tick" discipline applied to the primitive itself, so it can't regress into a
  per-row lookup. `getComponentData<C>(node)` returns `std::vector<C>&`
  (`ir_entity.hpp:54`), so `std::tie` captures live lvalue references and `cols[i]`
  yields `C&`. Both dispatch forms iterate `node->length_` and index
  `node->entities_[i]`, matching `ArchetypeNode`'s real members
  (`archetype_node.hpp:27,30`).

- **No reentrancy / iterator-invalidation hazard on the `RANDOMIZE_VOXELS`
  consumer.** The body's only mutation is `set.editVoxels(...)`, which rewrites
  voxel color data in place and resyncs the pool mask + face occupancy — it
  performs **no** structural ECS change (no create/destroy/add/remove-component),
  so the `nodes` vector and the per-node `columns` references stay valid across
  the whole traversal. The resync touches a *different* archetype (the canvas's
  `C_VoxelPool`) but only mutates its mask data, never its archetype membership.
  Safe to run mid-query.

- **Alpha/mask preservation makes `editVoxels`' resync a genuine no-op on
  occupancy.** Active voxels keep their original `alpha_` (only RGB moves) and
  alpha-0 voxels are skipped, so the active-mask and face occupancy the resync
  recomputes are byte-identical to before — exactly the invariant
  `command_randomize_voxels_test` asserts (carved voxel stays inactive, mask bit
  unchanged). The lambda signature `(int, C_Voxel&, vec3)` matches `editVoxels`'s
  callback contract (`component_voxel_set.hpp:485,495`).

- **`static_assert(false, …)` in the `if constexpr` else branch is well-formed
  here.** The project is `CMAKE_CXX_STANDARD 23` REQUIRED (`CMakeLists.txt:16-17`),
  and the toolchain floor (gcc-13+ / clang-17+ / AppleClang) implements P2593R1,
  which blesses a non-dependent `static_assert(false)` in a discarded/uninstantiated
  branch. No latent "fires at definition time" build break on any supported
  compiler — a real concern under C++20 / older toolchains, but a non-issue given
  the pinned standard.

- **Threading contract is consistent with the sibling primitive.**
  `executeQueryDynamic` gates on `IR_ASSERT_MAIN_THREAD()`, same rationale as
  `createSystemDynamic`'s `PARALLEL_FOR` assert — `getComponentData`'s manager
  hash lookups aren't worker-safe. Fired from `fireByName` on the main input
  thread; no flush attempted; deferred-ops contract documented at both the
  primitive and the CLAUDE.md.

- **CMake `PRIVATE`-link of `IrredenEngineSystem` is correct and cycle-free.**
  `engine/command`'s TU now includes `ir_system.hpp`; `IrredenEngineSystem`
  doesn't link `engine/command`, so no link cycle, and `PRIVATE` keeps it off
  `engine/command`'s public surface (downstream already gets it via
  `engine/world`). Duplicate-case guard satisfied — exactly one
  `case RANDOMIZE_VOXELS:` per switch after the promote-out-of-fall-through edit.

### On the escalation

No architectural follow-up implied and no invariant I'd want to discuss with the
author before merge — the primitive is the run-once dual of `createSystemDynamic`,
reuses the same traversal and dispatch, and adds no new persistent state. The
issue (#17) is fully specified and the implementation resolves the 2023 TODO
literally. Clearing the recheck as approve.

### Nits
None.

PR #2225, opened 2026-07-04
persist: SaveTrait<C> policy layer + component inventory (W-1/W-2) (P1) (#2212)

### LINKED ISSUE #2212: persist: SaveTrait<C> policy layer + component inventory (W-1/W-2) (P1)
**Model:** sonnet
**Part of epic:** #667
**Blocked by:** (none)

## Scope

Per-component persistence policy: compile-time `SaveTrait<C>` opt-in/opt-out + `kSaveVersion` (uint32, trait-carried), an audited explicit decision for all 169 engine components, and a compile-time completeness gate that fails the build on any undecided component. Pure metadata — no walker, no writer.

## Acceptance criteria

Completeness gate demonstrably fires when a decision line is removed; SaveTrait spot-check tests green (GPU-handle bearers opt-out, gameplay data opt-in v1); inventory count backstop; clean on linux-debug + macos-debug.

## Dependencies

Blocks P2 (walker consumes the traits).

## References

jakildev:
## Summary
- `engine/world/include/irreden/world/save_trait.hpp` — compile-time `SaveTrait<C>` (opt-in/opt-out + `kSaveVersion`, `shouldSave<C>()`/`saveVersion<C>()`, `IR_SAVE_OPT_IN`/`IR_SAVE_OPT_OUT` macros). Primary template means "no decision yet," not "opt-out."
- `engine/world/include/irreden/world/save_component_inventory.hpp` — an explicit, audited decision for all 165 real engine components (GPU-handle/derived/transient/ECS-plumbing → opt-out; gameplay data → opt-in v1; a few explicit-call cases), plus `AllEngineComponents` and a `static_assert` completeness gate that fails the build if any listed component lacks a decision. Verified the gate actually fires (temporarily deleted one decision line, confirmed the build breaks with the expected message, restored it).
- `test/world/save_trait_test.cpp` — spot-checks per class, fold-invariant check across the whole inventory, and a tuple-size backstop against the audited count (165).
- `engine/world/CLAUDE.md` — new "Save-trait policy layer" section documenting where decisions live, the opt-out-by-omission-is-forbidden invariant, and the new-component contract.

Pure metadata — no archetype walk, no writer, no Lua surface. P2+ builds the runtime serialize bridge on top of `shouldSave<C>()`/`saveVersion<C>()`/`AllEngineComponents`.

**Incidental fixes required to make the inventory's all-components include actually compile** (this is the first site that includes every component header together, so these were previously-latent, dead-or-untriggered bugs):
- `component_glfw_gamepad_state.hpp` / `component_midi_sequence.hpp` — each used a name (`GamepadButtons`/`GamepadAxes`, `IR_ASSERT`) that only resolved because of `using namespace` pollution from whichever TU happened to include it first; added the missing using-declarations/include so each header is self-contained.
- `component_position_int_3d.hpp` — was missing its `namespace IRComponents { ... }` wrapper entirely (every sibling file has it).
- `component_lerp_component.hpp` — dead code (zero includers before this PR) with a non-existent `#include <irreden/ir_ecs.hpp>` and constructor names that didn't match the class name; fixed to compile without changing its intended shape (entity id + `std::function` tick callback).
- Corrected 4 names that turned out to be commented-out, never-real components (`C_MainCanvas`, `C_GuiCanvas`, `C_BackgroundCanvas`, `C_IsGamepad` in `component_tags_all.hpp`) that a naive grep for `struct C_...` had picked up — dropped the audited total from the plan's assumed 169 to the verified 165.
- Reclassified `C_LerpEntity` from the plan's opt-in to opt-out after reading its body: it holds a `std::function` member, the same non-serializable shape as `C_LambdaModifiers`.

## Test plan
- [x] `fleet-build --target IrredenEngineTest` — clean build (macOS/Metal)
- [x] `./IrredenEngineTest` — full suite green: 1265 passed, 1 pre-existing skip (`GpuComputeDispatchTest.SkippedOnNonOpenGLBackend`, unrelated)
- [x] `./IrredenEngineTest --gtest_filter=SaveTrait.*` — all 6 new tests pass
- [x] Manually verified the completeness gate fails the build when a decision line is deleted (acceptance criterion #1)
- [ ] Linux (`linux-debug`) build — not verifiable from this macOS host; nothing in this PR is platform-specific (pure C++17 template metaprogramming, no shaders/backend code), so no reason to expect divergence, but flagging per epic acceptance criterion #5

Closes #2212

### DIFF
--- a/.fleet/plans/issue-2212.md
+++ b/.fleet/plans/issue-2212.md
@@ -1,0 +1,95 @@
+# Plan: persist P1 — SaveTrait<C> + kSaveVersion (W-1, W-2)
+
+- **Issue:** #2212
+- **Model:** sonnet
+- **Date:** 2026-07-03
+- **Epic:** #667 — see `.fleet/plans/issue-667.md` for full context
+- **Blocked by:** (none)
+
+### Scope
+
+Deliver the per-component persistence policy layer that every later persist phase consumes: a compile-time `SaveTrait<C>` trait (opt-in / opt-out) plus a `static constexpr uint32_t` save version for each opted-in component, and an explicit, audited decision stamped onto the entire engine component inventory (169 `C_*` types). No archetype walk, no IRWS writer, no serialize/deserialize functions, no Lua surface — those are P2+ (W-3..W-11). P1 is pure metadata + a build-time completeness gate.
+
+In scope:
+- `engine/world/save_trait.hpp` — the trait mechanism (opt-in/opt-out template, `kSaveVersion`, `shouldSave<C>()` / `saveVersion<C>()` helpers, `IR_SAVE_OPT_IN` / `IR_SAVE_OPT_OUT` macros).
+- A component-inventory header enumerating every engine component with its explicit decision, plus the compile-time fold that fails the build if any listed component lacks a decision.
+- A GoogleTest under `test/world/` that spot-checks the classification and backstops inventory completeness.
+
+Out of scope (do not build): the `ComponentId → serialize descriptor` runtime bridge (W-3), the CMPN name-table wiring (W-3/W-5), migration registry (W-4), GPU-handle regeneration pass (W-10). P1 only makes the decisions those phases read.
+
+### Verified current state
+
+- **#663 primitives landed in `engine/asset/`** (headers read directly): `binary_io.hpp` — abstract `IRAsset::BinaryWriter` / `BinaryReader` with `File*`/`Memory*` backends; `writeU8/16/32/64`, `writeI*`, `writeF32/F64`, `writeVarUInt(uint64)`, `writeString`, `writeTag`, `tell()/seek()`; reads return `Result<T>{ BinaryStatus status_; T value_; }`; errors via `BinaryIOError` enum, never throw (`binary_io.hpp:31-87`, `:95-153`, `:208-251`). `chunk_header.hpp` — `AssetHeader` (`kAssetHeaderSize=12`); `writeChunked`, `readChunks`, `findChunk`, `makeTag` (`chunk_header.hpp:33-121`). `name_table.hpp` — `writeNameTable`/`readNameTable` + bidirectional `NameTable` — the Rule #2 vehicle W-3 will use for CMPN (`name_table.hpp:38-85`). `json_sidecar.hpp`, `math_binary_io.hpp` also present. Existing `kSaveVersion` precedent is **`uint16_t`** on asset records (`rig_format.hpp:73,86`; `voxel_set_format.hpp:216,250`), guarded by the `// IRAsset: serialized` simplify check — conflicts with the epic's `uint32_t`; reconciled under Approach/Gotchas.
+- **`engine/world/` shape**: `world.cpp`, `chunk_residency.{hpp,cpp}`, `chunk_persistence.{hpp,cpp}`, `config.hpp`. `save_trait.hpp` is a **new sibling** header. `engine/world/CMakeLists.txt` **already links `IrredenEngineAsset`** — no CMake change needed.
+- **Component registration is lazy and type-erased**: `EntityManager::registerComponent<C>()` keys on `typeid(C).name()` (mangled) and stores an `IComponentDataImpl<C>` (`entity_manager.hpp:165-181`, `:226-232`; `i_component_data.hpp:49-93`). There is **no central list of engine components** anywhere — no X-macro, no registration table, purely first-use lazy. Consequence: (a) P1 must **create** the canonical component list itself, and (b) mangled `typeid` names are unstable across compilers, so the disk-stable component name must come from another source — the P1 inventory X-list stringizes the type token (`"C_Velocity3D"`) and gives W-3 a stable name for free.
+- **`ResourceId = uint32_t`** (`ir_render_types.hpp:17`); GPU components hold `std::pair<ResourceId, T*>` and free via `IRRender::destroyResource<T>` in `onDestroy()`.
+- **Test harness**: single GoogleTest exe `IrredenEngineTest`, sources in `test/CMakeLists.txt:4+`; `test/world/` exists.
+- **Inventory size**: 169 unique `struct C_*` across `engine/prefabs/**`, `engine/render/`, `engine/system/`.
+
+### Approach
+
+Single committed approach: **trait mechanism header + policy/inventory header + compile-time fold + backstop test.** Decisions live in one reviewable place, component structs stay untouched, and the build fails if any listed component has no decision.
+
+**Step 1 — `engine/world/include/irreden/world/save_trait.hpp` (mechanism).** Lightweight, no component includes.
+- Primary template encodes "no decision yet": `template <class C> struct SaveTrait { static constexpr bool kExplicit = false; static constexpr bool kSave = false; static constexpr std::uint32_t kSaveVersion = 0; };`
+- Two policy macros that specialize it, forcing `kExplicit = true`: `IR_SAVE_OPT_IN(Type, Version)` (version `>= 1`, `static_assert`ed) and `IR_SAVE_OPT_OUT(Type)`.
+- `constexpr bool shouldSave<C>()` / `constexpr std::uint32_t saveVersion<C>()` accessors.
+- **`kSaveVersion` is `uint32_t` and lives in the trait, not on the component struct.** The world-snapshot serializes a *schema* defined by the (P2) per-component serialize function, not the struct's in-memory layout; the schema's version belongs beside the policy. Also sidesteps the asset `uint16_t` simplify regex. Honors the epic's locked `uint32_t` decision. (See Gotchas for the version-bump-detection follow-up this creates.)
+
+**Step 2 — `engine/world/include/irreden/world/save_component_inventory.hpp` (policy + completeness gate).** Includes every component header (heavy — only pulled by world-snapshot TUs + the test), then: one `IR_SAVE_OPT_IN`/`IR_SAVE_OPT_OUT` line per component (the audit table below, transcribed into code); a canonical `using AllEngineComponents = std::tuple<...>;` listing all 169 types; and the compile-time completeness gate — `static_assert(detail::allExplicit<AllEngineComponents>(), "Every engine component must have an explicit IR_SAVE_OPT_IN/OPT_OUT decision.")`. Any listed component without a specialization uses the primary template (`kExplicit=false`) and **breaks the build**.
+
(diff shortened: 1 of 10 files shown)


jakildev (commented):
## Review — persist: SaveTrait<C> policy layer + component inventory (W-1/W-2) (P1) (#2212)

**Verdict:** approve

Pure compile-time metadata layer (no archetype walk, no writer, no Lua surface) plus
an audited per-component decision table. I independently verified the load-bearing
claims rather than trusting the PR body.

### Checklist walk-through

- **Completeness gate correctness:** `save_trait.hpp`'s `SaveTrait<C>` primary template
  defaults `kExplicit = false` (not `kSave = false`-as-default masquerading as
  opt-out) — the `detail::allExplicit<AllEngineComponents>()` fold in the same TU as
  the `IR_SAVE_OPT_IN`/`OPT_OUT` macro expansions correctly walks every listed type
  before the `static_assert` fires (`save_component_inventory.hpp:784-787`). No
  ordering hazard since all specializations land in the same header, before the fold.
- **Inventory self-consistency (verified programmatically, not by eye):** extracted
  every `IR_SAVE_OPT_IN`/`OPT_OUT(...)` type argument and every `AllEngineComponents`
  tuple entry from `save_component_inventory.hpp` and diffed the two sets — **165
  decls, 165 tuple entries, identical sets, zero duplicates in either list.** Matches
  `kExpectedEngineComponentCount = 165` and the test's `InventoryIsComplete` backstop.
- **Incidental "previously-latent" bug fixes are real, not invented cover:**
  - `component_lerp_component.hpp`: `git show origin/master:...` confirms the
    pre-PR file had a constructor named `C_LerpComponent` inside class `C_LerpEntity`
    (not a valid constructor) plus `C_LerpComponent() : {}` (invalid mem-initializer
    syntax) — genuinely broken, uncompilable code. `grep -rl` confirms this header had
    **zero includers** before this PR, so the class body was never parsed. The fix
    (rename to `C_LerpEntity`, fix `EntityId` → `IREntity::EntityId`, `= default`) is
    correct and now compiles (independently rebuilt, see below).
  - `component_position_int_3d.hpp`: confirmed no other file references
    `C_PositionInt3D` unqualified, so adding the missing `namespace IRComponents {}`
    wrapper is safe — nothing depended on it living in the global namespace.
  - `component_glfw_gamepad_state.hpp`: `GamepadButtons`/`GamepadAxes` are plain
    `enum`s in `namespace IRInput` (`ir_input_types.hpp:179,201`) — the added
    `using` declarations are the correct fix for what was transitive-include reliance.
  - `component_midi_sequence.hpp`: confirmed via `git show origin/master` that
    `IR_ASSERT` was already called at line 96 pre-PR with no `ir_profile.hpp`
    include in scope — the added include is a real fix, not incidental noise.
  - `component_tags_all.hpp`: `C_MainCanvas`/`C_GuiCanvas`/`C_BackgroundCanvas`/
    `C_IsGamepad` are commented-out (`// struct C_Foo{};`) — confirmed absent from
    both the audit and the tuple, so the "169 → 165, corrected via grep false-positive"
    claim in the plan doc holds.
- **Class-D opt-out rationale checked against the actual struct shapes:**
  `C_LambdaModifiers` holds `std::vector<LambdaModifier>` where `LambdaModifier`
  carries `std::function<float(float)>` — hard opt-out is correct.
  `C_Modifiers`/`C_GlobalModifiers`/`C_ResolvedFields` hold only vectors of
  trivially-copyable `Modifier`/`ResolvedField` structs (no `std::function`) — the
  PR's own comment correctly distinguishes this as a *softer* "opt-out for v1,
  pending entity-id remap" rather than claiming a hard technical blocker, which
  matches what's actually in `component_modifiers.hpp`.
- **Templated-component representative-instantiation pattern:** `C_SystemEvent`
  has no primary definition (forward-declared only, explicit specializations per
  `SystemEvent` value, each holding a `std::function`) — using `C_SystemEvent<TICK>`
  as the sole tuple representative is consistent with the same pattern applied to
  `C_GeometricShape<Shape3D::SPHERE>`, and both are Class-D/plumbing types excluded
  from the archetype walk by design, so the other specializations never need their
  own trait decision.
- **Naming/style:** new headers use `m_`/trailing-`_` correctly (no member fields
  besides the trait's `static constexpr`s); `IRComponents`/`IRWorld` namespacing is
  consistent; no `glm::`/`std::` math-primitive violations (this is pure
  metaprogramming, no math primitives touched).
- **Serialization checklist:** N/A here — this is a distinct compile-time ECS
  save-policy layer, not an `engine/asset/` binary format, so the `// IRAsset:
  serialized` / `kSaveVersion` bump convention doesn't apply to this diff.
- **Tests:** independently rebuilt `IrredenEngineTest` clean on macOS/Metal from
  this worktree (not trusting the author's report) — 0 errors/warnings from the
  new files. Ran the full suite: **1266 ran, 1265 passed, 1 skipped
  (`GpuComputeDispatchTest.SkippedOnNonOpenGLBackend`, pre-existing)** — matches the
  PR body's reported numbers exactly. Ran `--gtest_filter=SaveTrait.*` in isolation:
  all 6 new tests pass.
- **Build:** `test/CMakeLists.txt` gains one line registering the new test source —
  mechanical, alphabetically placed among the other `world/*_test.cpp` entries.

### Nits

None.

### Praise

- The primary-template-defaults-to-"undecided" design (`kExplicit = false`, not
  `kSave = false`) is the right call — it converts "someone forgot to classify a new
  component" from a silent data-loss bug (P2+) into a build break today, and the
  PR's own Gotchas section explicitly flags this as the trap reviewers must not
  "fix" by defaulting the primary template to opt-out.
- Verifying the completeness-gate mechanism by temporarily deleting a decision line
  and confirming the expected build failure (per the PR body) is exactly the right
  level of rigor for a build-time invariant like this — a green build alone can't
  prove the gate actually fires.
- The inventory audit reads as genuinely researched rather than name-pattern-matched:
  several reclassifications (`C_LerpEntity` opt-in→opt-out after reading its body,
  the four commented-out "components" dropped from 169→165) show the author read
  the actual headers rather than trusting a `grep 'struct C_'` sweep.

- [x] Linux (`linux-debug`) build — flagging for cross-host smoke tagging below since
  this PR's only platform-adjacent surface is the `test/CMakeLists.txt` edit; the
  new headers themselves are portable C++17 template metaprogramming with no
  platform-conditional code, so I don't expect a Linux-only failure, but per the
  cross-host-smoke procedure any `CMakeLists.txt` touch gets tagged rather than
  judgment-called.

Escalation: please Opus-recheck before merge (touches: `engine/world/`). No specific
concern found — flagging per the review checklist's bright line for
`engine/world/`, given this establishes an audited 165-component policy table that
every future persist phase (P2+) will build on; worth a second set of eyes on the
classification calls even though the mechanism itself checks out.

Opus recheck required: touches engine/world/ (new save-policy infrastructure feeding
all future persist phases).

jakildev (commented):
## Opus recheck — persist: SaveTrait<C> policy layer + component inventory (P1, #2212)

**Verdict: approve.**

Sonnet approved and escalated per the `engine/world/` bright line with "no specific
concern found," having programmatically verified inventory self-consistency (opt-in/out
lines vs the `AllEngineComponents` tuple) and the fold gate. I confirm that pass and spent
this recheck on the two things a self-consistency check and a hunk-level read can't reach:
whether the completeness-gate *mechanism* is actually trap-shaped correctly, and whether the
incidental component-header edits break call sites elsewhere in the tree.

### Completeness gate — mechanism is correctly load-bearing (confirm)

The plan's own gotcha warns reviewers to reject any change making the primary template default
to `kSave = false`. On closer read the implementation is the safe shape: `SaveTrait<C>` primary
template is `kExplicit = false` ("no decision"), the two macros set `kExplicit = true`, and
`detail::allExplicit<AllEngineComponents>()` folds `kExplicit` across the tuple behind a
`static_assert`. A tuple member with no `IR_SAVE_OPT_*` decision therefore fails the build rather
than silently not saving. This is the safety-critical invariant of the whole persist epic and it
is implemented correctly (`save_trait.hpp:12-20,45-49`; gate at `save_component_inventory.hpp:784-787`).

Design limit, documented not blocking: the gate catches "in the tuple, no decision" (the dangerous
direction). It cannot catch a genuinely new engine component omitted from the tuple *entirely* —
that leans on the hand-maintained `kExpectedEngineComponentCount = 165` backstop test plus the
CLAUDE.md new-component contract. Given the plan's verified finding that the engine has no central
component list (lazy `typeid`-keyed registration), this is the best available and is honestly
called out. Acceptable for P1.

### Macro hygiene (confirm)

`IR_SAVE_OPT_IN` / `IR_SAVE_OPT_OUT` are function-like macros, so a type argument containing a
top-level comma would corrupt expansion. The only templated entries — `C_GeometricShape<...SPHERE>`
and `C_SystemEvent<IRSystem::TICK>` — are single-arg specializations with no top-level comma, so
they pass through intact. No hazard.

### Incidental component-header edits don't break call sites (the diff-only blind spot)

Four existing headers were modified to make the whole inventory includable together. I traced each:

- **`component_lerp_component.hpp`** — the pre-PR struct was dead *and* malformed (a `C_LerpComponent(EntityId,…)`
  "ctor" on a struct named `C_LerpEntity`, plus a syntactically invalid `C_LerpComponent() : {}`). The header
  is `#include`d nowhere in engine or creations and `C_LerpEntity`/`C_LerpComponent` are referenced nowhere
  else, so it never compiled before now. The rewrite (`C_LerpEntity(EntityId, Function)` + `= default`) is
  correct; OPT-OUT for the `std::function` member is right.
- **`component_position_int_3d.hpp`** — newly wrapped in `namespace IRComponents`. `C_PositionInt3D` is
  referenced nowhere else in the tree (all 6 occurrences are inside its own header), so the wrap is required
  for the qualified inventory reference and breaks no call site.
- **`component_midi_sequence.hpp`** (add `ir_profile.hpp`) and **`component_glfw_gamepad_state.hpp`** (add
  two `using IRInput::` decls + a formatting-only `checkButton` reflow) are additive; duplicate `using`
  decls of the same entity are not an error.

### Classification spot-checks (confirm Sonnet)

Sampled the ambiguous/safety-critical calls beyond Sonnet's programmatic pass: Class-A GPU/ResourceId
bearers all OPT-OUT; `C_VoxelSetNew` OPT-IN is flagged provisional and genuinely carries `EntityId`
members (`canvasEntity_`, `ownerEntityId_`) + a `pendingVoxels_` span whose serializer is P2 — a
defensible intent-recording call that flips to OPT-OUT in one line if W-3 can't absorb it; `C_Persistent`
is an empty tag, fine as identity opt-in. No misclassification found.

### Notes (non-blocking, no re-review needed)

- Plan prose says "169 components" in several places while the code count is `165`; stale plan text, not a
  code issue — the tuple and the backstop agree at 165.

Pure compile-time metadata, no runtime surface, header-only additions + one test, gate correctly
trap-shaped, incidental edits proven call-site-safe. Ship it.

Issue #1394, opened 2026-05-30
fleet: GitHub API rate-limit exhaustion under multi-device polling — mitigation + identity strategy


jakildev:
**Area:** fleet (scripts/fleet, polling daemons) · **Owner:** human (auth/identity setup can't be done by a queue worker) · **Model:** n/a (human-driven infra)

## Problem

Running the fleet on multiple devices (mac + linux, + scheduled/cron runs) exhausts the **shared GitHub API quota of a single account**. Observed 2026-05-30: GraphQL primary limit hit `remaining: 0 / 5000` (used 5001) while REST still had ~95% headroom (`gh pr view` / `gh issue view` default to **GraphQL**; `gh api repos/...` REST endpoints draw from a separate pool). All devices authenticate as the same user, so every device's polling draws from the same 5000-points/hr (GraphQL) and 5000-req/hr (REST) buckets.

Per-system polling daemons already reduce *intra*-device redundancy, but the buckets are **per-account**, so N devices still contend for one quota.

## Considered solutions (cheapest / highest-leverage first)

1. **Conditional requests (ETags / If-None-Match).** REST `304 Not Modified` responses **do not count against the rate limit**. Polling daemons that cache ETags per endpoint and re-poll conditionally turn most "has anything changed?" calls free. Biggest immediate win for repeated-endpoint polling. (`gh` CLI doesn't do this natively — would need a thin REST poller or `curl` with ETag cache.)
2. **Centralize polling per-account, not per-system.** One authoritative poller fetches state and shares `~/.fleet/state/state.json` to the other devices (shared FS / rsync / a tiny served endpoint), collapsing N pollers → 1. This removes the multiplication entirely. We already have per-system daemons; the gap is cross-system.
3. **Prefer REST for reads + batch GraphQL.** Route fleet reads through REST `gh api` (separate, less-contended pool) and, where GraphQL is needed, batch many objects into one query (cheaper in points than N separate `gh pr view` calls). The exhaustion we hit was specifically GraphQL with REST nearly untouched.
4. **Webhooks instead of polling (long-term real fix).** Event-driven state updates → near-zero steady-state API cost. Largest lift; the architecturally correct end state.

## Identity / quota-multiplication strategy (the "separate accounts?" question)

[...]

PR #2227, opened 2026-07-04
fleet: conditional-REST scout polling (ETags, gated GraphQL) (#2219)

### LINKED ISSUE #2219: fleet: conditional-REST polling core — ETags + REST-for-reads + batched GraphQL (Q1)
**Model:** opus
**Part of epic:** #1394
**Plan file:** `.fleet/plans/issue-1394.md` (umbrella context)
**Blocked by:** (none)

## Scope

New fleet_gh_poll.py conditional_get (stdlib urllib, per-endpoint {etag,body} cache, 304 = free) + fleet-gh-poll CLI seam; rewire the scout's 7 unconditional GraphQL list fetchers to conditional REST and gate the open-PR GraphQL behind a REST ETag change-detector, batching it to one query when it fires. Steady-state idle tick -> ~0 quota from both pools.

## Acceptance criteria

Quiescent 10-min run: graphql + core remaining deltas <= 2 vs a >=150 baseline; freshness within one tick on a real change; field-equivalent state.json parity; unit tests for 304/200/error paths.

## Dependencies

Blocks Q2 (the LAN hop reuses the conditional-GET helper).

## References

The canonical plan is this issue's `## Plan` comment (posted at filing, architect-planned with the human in the loop — 2026-07-03 triage session). Umbrella: #1394.

jakildev:
## Summary

Fixes the scout's steady-state GitHub quota drain (#1394 Q1). Each 30 s tick
fired **8 unconditional GraphQL list fetches per repo**, exhausting the
`graphql` pool while `core` sat idle.

- **New `fleet_gh_poll.py`** — `conditional_get(repo, path, …)` over stdlib
  `urllib` (in-process, no `curl`/`requests`): `If-None-Match` + a
  per-endpoint `{etag, body}` cache written atomically. A `304` serves the
  cached body at **zero quota**; a network/HTTP/parse error returns
  `(True, None)` so callers keep their last-known-good fallback. `cache_dir`
  is injectable (no hardcoded `~/.fleet`) so the Q2 centralized poller reuses
  it verbatim.
- **New `fleet-gh-poll` CLI** — the subprocess seam (`0`=changed/200,
  `3`=unchanged/304, `1`=error). Indexed in `fleet-help`.
- **Scout rewire** — the 7 pure list reads (`needs_plan`, `plan_review`,
  `human_approved`, `closed_fleet_queued`, `recent_merged_prs`,
  `task_queue`, `epics`) now go through `conditional_get`. The open-PR list
  **stays on GraphQL** (it needs `reviews` + `mergeable`, which the REST list
  omits) but is **gated** behind a cheap conditional-REST change-detector:
  `304` → reuse the prior parsed list and skip the GraphQL query entirely.
- `install.sh` symlinks the CLI; the `.py` module is co-located (resolves via
  the scout's `sys.path.insert`, like `fleet_blocked_by.py` — no symlink).

Net: a warm quiescent tick issues **16 conditional GETs, all 304 → ~0 quota
from both pools** (was ~16 GraphQL). Cold first tick after deploy pays full
cost and converges to 304 next tick (expected).

## Parity / behavior notes

- REST is snake_case and `/issues` returns PRs too — the mapping helpers
  normalize back to the **exact prior field set** and drop `pull_request`
  entries. `sort=created&direction=desc` + matched `per_page` reproduce gh's
  default ordering and `--limit` windows (`human_approved` keeps gh's
  default-30 window; closed=100; task=200 via 2 pages).
- Freshness: the fleet is label-driven, and every review/state transition
  bumps a PR's `updated_at` (→ new list ETag → detector fires); a merge
  removes the PR from the open set (→ new ETag), so sibling `mergeable`
  refreshes within one tick.

## Test plan

- [x] `test_gh_poll_conditional.py` — 304 / 200 / stale-etag-refetch / error /
      no-token paths (hermetic, urllib mocked).
- [x] Existing `test_scout_degraded_fetch` + enrichment suites still green
      (138 Python tests pass).
- [x] **Live parity** — replayed the full pre-write transform against the live
      cache: every section field- and number-equivalent.
- [x] **Live quota** — isolated warm tick = 0 gh subprocess calls, 16
      conditional_get all 304.
- [x] CLI round-trip against real GitHub (200→exit 0, 304→exit 3); `ruff`
      clean on all touched files.

Authored + verified on macOS; the change is pure Python + `gh`/REST (no
backend code), so it runs identically on the Linux/Windows scout host.

Closes #2219


jakildev (commented):
## Review — fleet: conditional-REST scout polling (ETags, gated GraphQL) (#2219)

**Verdict:** needs-fix

Solid design overall: `fleet_gh_poll.conditional_get` (If-None-Match + atomic
etag+body cache, "an error is never unchanged", injectable `cache_dir` for
the Q2 poller) is well-reasoned, ruff-clean, and the new
`test_gh_poll_conditional.py` suite covers the 200/304/stale-etag/error/no-token
paths solidly. The gated-GraphQL PR-list detector in `fetch_prs` is a clean
way to keep `reviews`/`mergeable` on GraphQL while still avoiding the query on
a quiescent tick. But the refactor broke existing test coverage for
`fetch_task_queue`, and the way it broke has a real side effect worth fixing
before merge, not just a stale assertion.

### Needs-fix

- `scripts/fleet/fleet-state-scout:581` (`fetch_task_queue`) — migrating this
  function's internals from `run_capture(["gh", "issue", "list", ...])` to
  `_rest_list(...)` → `conditional_get(...)` silently broke every existing
  test that mocks the old path. I ran the full test suite in this repo's
  `scripts/fleet/tests/` (not just the two suites named in the PR's test
  plan) against this branch and got **23 failures out of 556** (the same
  suite is 550/550 green on `master`):
  - `test_issue_queue_parser.py:118-131` (`FetchTaskQueueDispatch._run`) monkey-patches
    `_mod.run_capture` with a `fake_run_capture` that returns synthetic issue
    JSON.
  - `test_worker_projection.py:517` and `:558` (`PlanReviewExcludedFromTasksOpen._fetch`,
    `NeedsGlHostAnnotation._fetch`) do the same via `patch.object(_mod, "run_capture", ...)`.

  None of these three call sites are exercised by `fetch_task_queue` anymore —
  it now calls `conditional_get`, which goes straight to `urllib.request.urlopen`
  and bypasses the mock entirely. I confirmed this isn't just "assertions need
  updating": on this machine (with a live `gh` token) the tests silently fall
  through to the **real** GitHub API and return real production data instead
  of the synthetic fixture (e.g. `test_label_beats_body_for_model` asserts
  `model == "sonnet"` from a synthetic single-issue fixture but gets back
  `"opus"` — the real #100 issue's real label state). Worse, since
  `conditional_get`'s `cache_dir` defaults to `Path.home() / ".fleet" / "state" / "etag"`
  and none of these tests override it, running this suite writes real entries
  into the **live fleet's shared ETag cache** — I verified a fresh file
  landed under `~/.fleet/state/etag/jakildev_IrredenEngine/` timestamped to
  my test run. That's the same cache directory the actual `fleet-state-scout`
  daemon reads/writes on this host. A "unit test" that hits the live API and
  mutates the production scout's cache as a side effect is a correctness and
  operational hazard on top of the 23 broken assertions — doubly so given
  this PR's whole point is cutting the fleet's live quota usage.

  Suggested fix: patch `_mod.conditional_get` (or `_mod._rest_list`) instead
  of `_mod.run_capture` in all three affected test classes, and/or have the
  tests pass an explicit throwaway `cache_dir` (e.g. a `tempfile.TemporaryDirectory`)
  so a mock miss can never fall through to the real network or the real
  cache directory. The PR's own new `test_gh_poll_conditional.py` already
  demonstrates the right pattern (mocks `urllib.request.urlopen`, uses a
  `tempfile.TemporaryDirectory()` for `cache_dir`) — the fix is applying that
  same discipline to the three now-stale call sites above.

### Nits

- `scripts/fleet/fleet_gh_poll.py:52-67` (`auth_token`) — the memoized
  `_token_cache["value"]` is read/written from multiple `ThreadPoolExecutor`
  worker threads in `collect_state()` with no lock. On a cold process start,
  several fetch specs can race past the `if _token_cache["value"]:` check
  before any of them finishes shelling `gh auth token`, so the first tick can
  fire a handful of redundant `gh auth token` subprocesses (self-healing —
  every subsequent call sees the memoized value). Not a correctness bug
  (dict assignment is atomic under the GIL and every racer gets the same
  token), just a bit of avoidable one-time subprocess overhead; a
  `threading.Lock` around the check-then-shell would close it if it's worth
  the complexity.

### Praise

- The module docstring's "Design invariants" section in `fleet_gh_poll.py`
  (atomic etag+body write, error-is-never-unchanged) is exactly the right
  level of durable-invariant documentation — it explains *why* the atomic
  write and the `(True, None)` error contract exist, not just what the code
  does.
- `fetch_prs`'s gated-GraphQL detector comment (`fleet-state-scout` around
  the `fetch_prs` definition) walks through why the ETag can't go stale in a
  way that matters (`mergeable` can only drift if master advances without
  the PR moving, and master only advances by merging a PR, which itself
  flips the open-set ETag) — a genuinely convincing invariant argument, not
  hand-waving.
- Choosing to keep the open-PR list on GraphQL (it needs `reviews` +
  `mergeable`, which REST's list endpoint omits) rather than force-fitting
  everything onto REST is the right call, and gating it behind a cheap REST
  change-detector is a clean way to get both.

- [ ] Fix the three broken test call sites (`test_issue_queue_parser.py:118-131`,
      `test_worker_projection.py:517`, `test_worker_projection.py:558`) to mock
      the new REST path, then re-run the **full** `scripts/fleet/tests/` suite
      (not just the subset named in the original test plan) and confirm it's
      556/556 green.
- [ ] After the fix, re-verify no test run touches the real
      `~/.fleet/state/etag/` directory (e.g. by asserting `cache_dir` is
      always the injected tempdir, or by checking the real directory's mtimes
      are untouched by a fresh test run).

jakildev:
Addressed fleet review feedback (needs-fix + nit + test-plan item 2):

- **needs-fix — 3 broken `fetch_task_queue` test sites now hermetic.** `fetch_task_queue` migrated to `_rest_list` → `conditional_get`, so the old `run_capture` monkey-patches missed and fell through to the live API + the shared `~/.fleet` ETag cache. Patched the `conditional_get` seam instead:
  - `test_issue_queue_parser.py` `FetchTaskQueueDispatch._run` → patches `conditional_get` to `(True, payload)`.
  - `test_worker_projection.py` `PlanReviewExcludedFromTasksOpen._fetch` + `NeedsGlHostAnnotation._fetch` → same.
- **test-plan item 2 — closed the same hazard in `test_scout_degraded_fetch.py`.** Those tests call `collect_state()`, which also fetches engine `plan_review` (via `fetch_plan_review` → `conditional_get`) — left unmocked, so on a real host it hit the API + wrote the real ETag cache silently (it just didn't produce an assertion failure). Stubbed `fetch_plan_review` in `setUp` so the whole suite is hermetic.
- **nit — `auth_token` memo now lock-guarded.** Double-checked lock around the check-then-shell so a cold-start `ThreadPoolExecutor` fan-out shells `gh auth token` once between threads, not once per thread. Fast path (already-memoized) still lock-free.

Verification: full `scripts/fleet/tests/` suite **556/556 green** run under an isolated `HOME` with **no gh token** — a pass under those conditions proves nothing reaches the network; confirmed **zero** files written to the real `~/.fleet/state/etag/` (count unchanged at 28). ruff clean. — worker

jakildev (commented):
## Review — fleet: conditional-REST scout polling (ETags, gated GraphQL) (#2219)

**Verdict:** approve

### Prior-review resolution

- ✅ `scripts/fleet/fleet-state-scout:581` (`fetch_task_queue`) — the three broken test call sites (`test_issue_queue_parser.py:118-131`, `test_worker_projection.py:517`, `test_worker_projection.py:558`) now patch `conditional_get` directly (returning the `(changed, body)` tuple shape the real function returns) instead of the no-longer-called `run_capture`. Verified at `e34c0bd5` by actually running the full `scripts/fleet/tests/` suite myself under an isolated `HOME` with no `gh` token reachable: **556/556 green**, matching the author's claim. Also confirmed the real `~/.fleet/state/etag/jakildev_IrredenEngine/` cache files' mtimes (09:18/09:21 local) predate my test run (10:07 local) — the suite did not touch the live cache.
- ✅ `scripts/fleet/fleet_gh_poll.py:52-67` (`auth_token`) — now guarded by a `threading.Lock` with a lock-free fast path for the already-memoized case and a double-checked-lock on the cold-start/refresh path. Read through the logic: `refresh=True` always clears the cache and re-shells inside the lock, `refresh=False` racers block on the lock and pick up whichever value the first shell-out produced. Correct.

The author also proactively fixed an equivalent hazard I hadn't flagged: `test_scout_degraded_fetch.py`'s `collect_state()` call also fetches engine `plan_review` via `conditional_get`, which was left unmocked in those tests. A `setUp` stub for `fetch_plan_review` now closes that gap too — same class of bug, good catch.

No new findings in the delta (`e34c0bd5` only touches the four files above, all test/lock-fix scope).

### Nits

None outstanding.

### Praise

- Patching the seam one level up (`conditional_get`) rather than re-mocking `_rest_list` keeps the tests coupled to the real REST boundary instead of an intermediate implementation detail — matches the discipline already used in the PR's own `test_gh_poll_conditional.py`.
- Fixing the `test_scout_degraded_fetch.py` hazard beyond what was explicitly flagged, and calling out the "same hazard, different fetcher" pattern in the commit message, shows the author generalized the finding rather than patching the exact three cited lines.

- [x] Confirmed independently — full suite 556/556 green, live ETag cache untouched.

Opus recheck not required.

PR #2228, opened 2026-07-04
persist: IRWS world snapshot container + walker + loader (#2213)

### LINKED ISSUE #2213: persist: IRWS container + deterministic archetype walker + singleton chunk (W-3/W-6) (P2)
**Model:** opus
**Part of epic:** #667
**Blocked by:** #2212

## Scope

`IRPersist::saveWorld/loadWorld` in engine/world: IRWS chunked container (CMPN/ARCH/SNGL/META + JSON sidecar via the #663 primitives), deterministic archetype walker writing opted-in columns + entity IDs (restore-exact IDs — never remapped), singleton chunk restoring by value, and the type-erased save registry bridging P1 traits to runtime columns. Relations/migration/GPU-regen out of scope (P3/P5/P6).

## Acceptance criteria

Headless gtest: >=100 entities / >=3 archetypes round-trip with exact original EntityIds + projection-merge proof; singleton value round-trip; recoverable failures with zero mutation; same-session double-save byte-identical; clean on both presets.

## Dependencies

Blocks P3 (relation chunk plugs into the container).

## References

jakildev:
## Summary
- `IRWorld::saveWorld` / `loadWorld` — the **IRWS** entity-level snapshot container (persist P2, epic #667): `CMPN`/`ARCH`/`SNGL`/`META` chunks + JSON sidecar over `engine/asset/` primitives. Distinct from `chunk_persistence.hpp`'s per-chunk `.vxs` voxel save.
- `save_serialize.hpp` — `SaveSerialize<C>` per-component bytes customization point (trivially-copyable default; specialize for heap-owning components).
- `save_registry.hpp` — type-erased `SaveRegistry` bridging P1's compile-time `SaveTrait<C>` to runtime archetype columns; `registerComponent<C>()` is `if constexpr`-gated on `shouldSave<C>()`.
- `save_trait.hpp` — additively adds `SaveTrait<C>::kSaveName` (the type's source spelling via the opt-in/out macros) as the stable CMPN on-disk key. **P1 reconciliation** the plan flagged (P1 shipped no stable name); it's the only change to #2212's file.
- Deterministic **projection-merge** walker (two live archetypes differing only by an opted-out component merge into one saved archetype) + a **two-phase loader** (validate-then-apply, **zero world mutation on any error**, Rule #5) restoring **exact original EntityIds**.
- `EntityManager` restore surface: `findCreateArchetypeNode` wrapper, `restoreEntitiesBatch`, `singletonEntityCache`/`isComponentBackingEntity` (the `resetGameplay`-mirroring exclusion set), `entityIdWatermark`/`advanceEntityIdWatermark`.

### Scope / deviations from the plan
- Namespace is `IRWorld` (matches P1's `save_trait.hpp`), not the plan's `IRPersist` (which doesn't exist in-tree).
- `saveWorld/loadWorld` take an explicit `const SaveRegistry&` — the honest P2 surface, since a registry decides what participates. The `(path)`-only P7 convenience wrapper layers on once engine components carry serializers.
- P2 ships the **mechanism**, proven with trivially-copyable *test* components (per acceptance criterion #1 "test components"). Registering the 165 engine components — each non-POD one needs a `SaveSerialize<C>` specialization — is downstream.

## Test plan
- [x] `test/world/world_snapshot_test.cpp` — 11 tests covering all six acceptance criteria: 105-entity / 4-archetype round-trip with exact ids + projection-merge, singleton by-value + alias map, opt-out absence, post-load watermark, byte-identical double-save, and recoverable errors (bad magic / version-too-new / truncation / live-id collision zero-mutation / unknown-name skip).
- [x] Full entity/save/world regression green on `macos-debug` (256 tests), no regressions from the `entity_manager` + `save_trait` changes.
- [ ] Linux `linux-debug` build + gtest (acceptance criterion #7 requires both presets; macOS validated here, Linux pending a GL-host pane).

Stacked on: https://github.com/jakildev/IrredenEngine/pull/2225

Closes #2213

### DIFF
--- a/.fleet/plans/issue-2213.md
+++ b/.fleet/plans/issue-2213.md
@@ -1,0 +1,90 @@
+## Plan: persist P2 — IRWS container + deterministic archetype walker + singleton chunk (W-3, W-6)
+
+- **Issue:** #2213
+- **Model:** opus
+- **Date:** 2026-07-03
+- **Epic:** #667 — see `.fleet/plans/issue-667.md` for full context
+- **Blocked by:** #2212
+
+### Scope
+
+Deliver `IRPersist::saveWorld(path)` / `IRPersist::loadWorld(path)` in `engine/world/`: the `IRWS` snapshot container (magic + uint32 version + chunk table + JSON sidecar via `engine/asset/` primitives), a deterministic archetype walker that writes opted-in component columns + entity IDs (epic item W-3), and the singleton chunk for `IREntity::singleton<T>()` instances (epic item W-6), plus the type-erased save registry that bridges P1's compile-time `SaveTrait<C>` to runtime archetype columns. Loads restore into a reset world.
+
+Out of scope: relations (P3 — relation pseudo-components are stripped, no relation chunk emitted here), migration registry (P5 — but every column carries the `kSaveVersion` + byte-length header P5 needs), GPU-handle regeneration (P6 — opted-out columns simply aren't written), Lua bindings (P7), full round-trip/determinism test matrix (P4 — this task ships one headless gtest proving the round-trip). Named-entity (`m_namedEntities`) persistence is not in the epic's W-list and is excluded.
+
+Blocked by P1 (`engine/world/save_trait.hpp` with opt-in/out traits + `kSaveVersion`) — verified absent on current master, so this task must not start until P1 merges.
+
+### Verified current state
+
+- **Entity IDs are monotonic and never recycled.** `m_nextEntityId` is an atomic counter starting at `IR_RESERVED_ENTITIES` (0xFF); destroy retires the ID permanently (`engine/entity/src/entity_manager.cpp:55-65`, `:99-105`). Raw ID = low 25 bits (`IR_ENTITY_ID_BITS`, `ir_entity_types.hpp:26-32`). No generation counter — an `EntityId` is index bits + flag bits only. The only flag ever set is `kEntityFlagIsRelation` (`entity_manager.cpp:518`); `IR_PURE_ENTITY_BIT` / `kEntityFlagIsSystem` are declared but never set (repo-wide grep).
+- **ComponentIds are session-local entity IDs.** `registerComponentImpl` allocates the ComponentId via `createEntity()` (`entity_manager.cpp:566-572`), keyed by `typeid(C).name()` (`entity_manager.hpp:165-181`) — both the numeric ID and the typeid key are unstable across sessions/compilers, so neither may be the on-disk identity. Rule #2 name tables are mandatory.
+- **Archetype walk surface exists.** `EntityManager::getArchetypeNodes()` returns all `smart_ArchetypeNode`s (`entity_manager.hpp:78-80`); each `ArchetypeNode` exposes `type_` (a `std::set<ComponentId>`, already sorted), `entities_`, `components_`, `length_` (`archetype_node.hpp:23-35`). Typed column access is `castComponentDataPointer<C>(...)->dataVector` (`i_component_data.hpp:49-98`). `findCreateArchetypeNode` is graph-public but not exposed through EntityManager (only `findArchetypeNode`, `entity_manager.hpp:75-77`).
+- **Singleton internals.** Cache is `m_singletonEntityByComponent` (private, `entity_manager.hpp:522`), lazily validated, cleared by `destroyAllEntities` (`entity_manager.cpp:352-373`) but **not** by `resetGameplay` (`entity_manager.cpp:442-443`). The untyped `getOrCreateSingletonByComponentId` path asserts on C++ components (`appendDefaultRow` returns false — `entity_manager.cpp:446-459`, `:596-599`); only the typed `getOrCreateSingleton<C>` (`entity_manager.hpp:453-466`) works generically. No public accessor for the cache map exists.
+- **#663 primitives.** `IRAsset::BinaryWriter/Reader` with file + memory backends, `Result<T>`/`BinaryStatus` recoverable errors (`binary_io.hpp:95-153`, `:208-251`); `writeChunked(w, magic, version, span<ChunkPayload>)` / `readChunks(r, expectedMagic, maxKnownVersion)` with unknown-chunk pass-through (`chunk_header.hpp:88-117`); `NameTableEntry{u32 id_, string name_}` + `NameTable` lookup (`name_table.hpp:38-85`); `JsonSidecarWriter` (write-only, Rule #6); `math_binary_io.hpp` for `IRMath` types. `IrredenEngineWorld` already links `IrredenEngineEntity` + `IrredenEngineAsset` (`engine/world/CMakeLists.txt:7-21`).
+- **Fresh-world / teardown interaction.** `resetGameplay` preserves three categories: singletons, `C_Persistent`-tagged entities (camera/framebuffer/canvas stamped at `render_manager.cpp:128-135`), and component-backing entities (`destroyAllExceptPreserved`, `entity_manager.cpp:375-444`); the `C_Persistent` policy lives in the `IREntity::resetGameplay` facade (`ir_entity.cpp:19-27`). `engine/entity/CLAUDE.md` "Save/load implications" already commits #199 to "the entity id round-trips" and notes the singleton cache must rebind on load.
+- **Headless test harness exists.** A bare `EntityManager` member is a valid world for gtest (ctor sets `g_entityManager`; `test/ecs/entity_manager_test.cpp:44-56`); world tests register at `test/CMakeLists.txt:90-93`.
+
+### Approach
+
+**Entity-ID decision: restore-exact (raw IDs preserved on load), not remap.** Rationale: (1) epic acceptance #1 explicitly says the serializer "restores … entity IDs" and `engine/entity/CLAUDE.md` already documents "the entity id round-trips" for #199; (2) components embed raw `EntityId` fields (e.g. `C_VoxelSetNew`'s canvas ID) and P2 has no reflection to remap inside opted-in blobs — restore-exact keeps them valid, remap silently breaks them all; (3) W-7 byte-for-byte round-trip parity and P3's relation triples both require stable IDs; (4) the mechanics are cheap because IDs never recycle: restore inserts saved raw IDs into `m_entityIndex`/nodes and advances `m_nextEntityId` past the saved watermark, so post-load allocations can never collide. Collision safety comes from a **save-side exclusion set that mirrors `resetGameplay`'s preserve set**: saveWorld skips (a) singleton entities (handled by the SNGL chunk as value overwrites), (b) `C_Persistent`-tagged entities, (c) relation-flagged and component-backing entities. The documented load contract is `IREntity::resetGameplay()` → `IRPersist::loadWorld(path)` at a frame boundary — everything the save wrote is exactly what the reset destroyed, so same-session load is collision-free *by construction*. Cross-session, load is two-phase (parse + validate everything, then apply): any saved ID colliding with a live entity returns a recoverable `BinaryStatus` error with zero world mutation (Rule #5). A whole-file ID-rebase fallback is left as P5 design space if cross-build infra-ID drift ever bites.
+
+**Singletons load by value, not by ID.** Singletons survive teardown by design, so the SNGL chunk loader resolves each entry by save-name, lazy-creates via the typed `IREntity::singletonEntity<C>()` captured in the registry, and overwrites the live row in place. The saved singleton entity ID is still written and surfaced in `LoadResult` as a `savedId → liveId` alias map (identity in the same-session case) so P3 relation triples referencing a singleton can translate.
+
+**Steps:**
+
+1. **Type-erased save registry** (`engine/world/include/irreden/world/save_registry.hpp` + src). `IRPersist::registerSaveComponent<C>()` — enabled only when P1's `SaveTrait<C>` opts in — captures per component: stable save-name, `kSaveVersion`, `ComponentId` resolver (lazy `getComponentType<C>()`), `writeRow(const IComponentData*, int row, BinaryWriter&)`, `readAppendRow(IComponentData*, BinaryReader&)` (default-construct + trait read; `static_assert` default-constructible), `readIntoEntity(EntityId, BinaryReader&)` and `getOrCreateSingletonEntity()` for the SNGL path. Keyed both by save-name and by resolved ComponentId. Reconcile at claim time: if P1 shipped a runtime registry or `kSaveName` member, consume it; if traits only, this shim is P2's to own.
+2. **Minimal EntityManager restore surface** (`engine/entity/`): `findCreateArchetypeNode(const Archetype&)` public wrapper; `restoreEntitiesBatch(ArchetypeNode*, std::span<const EntityId>)` (index emplace + record update + `length_`/`entities_` append + liveEntityCount, main-thread asserted — columns appended afterwards by the registry readers, with an end-of-node sync assert mirroring `entity_manager.hpp:271-275`); `const std::unordered_map<ComponentId, EntityId>& singletonEntityCache() const`; `EntityId entityIdWatermark() const` and `advanceEntityIdWatermark(EntityId)`.
+3. **Walker + writer** (`engine/world/include/irreden/world/world_snapshot.hpp` + `src/world_snapshot.cpp`, namespace `IRPersist`, `kWorldSnapshotMagic = "IRWS"`, `kWorldSnapshotVersion = 1`). `saveWorld` flushes structural changes + drains marked deletions first (frame-boundary, main-thread contract), builds the exclusion set, then for every archetype node with `length_ > 0` computes the **projection**: the subset of `type_` that is registered + opted-in (relation pseudo-components and unregistered components drop out here). Nodes with identical projections merge into one saved archetype; merged entity lists sort ascending by masked raw ID (locked order); archetype blocks sort lexicographically by projected sorted-ComponentId set (locked order). Chunks, composed in `MemoryBinaryWriter`s and emitted via one `writeChunked` to a `FileBinaryWriter`:
+   - `CMPN` — name table `(componentId as u32, saveName)` for every referenced component (Rule #2).
+   - `ARCH` — varuint archetypeCount; per archetype: varuint componentCount + sorted componentRefs; varuint entityCount + ascending varuint raw entity IDs; then per column (same component order): version (`kSaveVersion`, width per P1 — epic W-2 says u32) + varuint columnByteLength + rows written in entity order via `writeRow` (gathering per-entity `(node,row)` for merged groups). The byte-length prefix makes any unresolvable column skippable on load — the seam P5's migration registry plugs into.
+   - `SNGL` — varuint count; entries sorted by save-name (ComponentIds aren't cross-session stable; name order keeps the chunk deterministic): componentRef + saved raw entity ID + version + varuint byteLength + value blob. Only cache entries whose component is opted-in and whose entity is live.
+   - `META` — `nextEntityId` watermark + total saved entity count.
+   - JSON sidecar at `path + ".json"` (summary; write-only, Rule #6), mirroring the `.vxs` pattern.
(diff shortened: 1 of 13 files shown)


jakildev (commented):
## Review — persist: IRWS world snapshot container + walker + loader (#2213)

Stacked on #2225 (now `fleet:approved`, `fleet:needs-windows-smoke`). Reviewing the child diff only — upstream is approved separately.

**Verdict:** needs-fix

### Needs-fix

- `engine/world/src/world_snapshot.cpp:531-569` (Phase 3 apply) — a mid-column decode failure leaves the world partially mutated, contradicting the documented "Rule #5: zero world mutation on error." Phase 1 (`:422-441`) only records each column's `byteLength_`/`dataOffset_` by seeking past it — it never decodes the bytes, so a column that is corrupted-but-length-valid (disk corruption, a killed process mid-write, hand-edited bytes) isn't caught until Phase 3's `entry->appendRow_(r, col)` (`:558`) fails partway through a column. At that point `restoreEntitiesBatch` (`:541`) has already appended every entity of this archetype (and every earlier `StagedArchetype` in the file) into the live graph, so the function returns `loadFail(rowStatus)` (`:560`) leaving this node's *current* column short of `node->length_` while sibling columns of the same node are already fully populated — any subsequent iteration over that node reads the short column out of bounds. Suggested fix: either decode every column's rows into scratch buffers during Phase 2 (making it the real validation gate) and only start touching the live archetype graph once the whole file parses cleanly, or build Phase 3 into a disposable staging structure and splice into the live graph only on full success.
- `engine/world/src/world_snapshot.cpp:591` — `em.advanceEntityIdWatermark(watermark)` runs *after* both the archetype-restore loop (`:531-569`) and the singleton-restore loop (`:571-589`), but the singleton loop's `entry->getOrCreateSingletonEntity_()` (`:577`, wired to `IREntity::singletonEntity<C>()` → `EntityManager::getOrCreateSingleton<C>()`, `entity_manager.hpp:502-515`) lazy-creates any singleton whose cache entry doesn't exist yet this session via `createEntity(Component{})` — which draws a **fresh** id off the still-stale, unadvanced `m_nextEntityId` (`allocateEntityIdAtomic`, `entity_manager.cpp:55-65`). This is not a corrupted-input scenario: it's the primary intended use case (boot fresh, immediately `loadWorld()`) — a fresh session's watermark sits in the same low range the file's own entities occupy (both start right after the same infra/component-registration entities), so a numeric collision between a freshly-lazy-created singleton id and an id `restoreEntitiesBatch` just planted into `m_entityIndex` earlier in the *same* `loadWorld` call is plausible, not remote. `EntityManager::allocateEntity` (`entity_manager.cpp:46-53`) ignores `m_entityIndex.emplace`'s return value, so on collision the singleton's component write silently lands on the *already-restored* gameplay entity's record — no error, no assert, just a cross-wired entity. Fix: call `em.advanceEntityIdWatermark(watermark)` immediately after Phase 2 validation succeeds and before any Phase 3 mutation (archetype restore or singleton lazy-create).

### Nits

- `engine/world/src/world_snapshot.cpp:67-73, 117-125` — a singleton entity is excluded from the `ARCH` walk wholesale (`singletonIds.contains(maskedId)`, `:120`), and `SNGL` only persists the one component type keyed by `singletonEntityCache()`. If a singleton entity ever carries an *additional* opted-in component beyond its own singleton type (nothing currently in-tree does this, per `engine/entity/CLAUDE.md` "Singleton components," but it's not disallowed), that extra component's data is silently dropped on save — not written to `ARCH` (excluded) or `SNGL` (wrong type). Worth an assert or a doc note in `world_snapshot.hpp`'s header block making this a documented restriction rather than an implicit one.
- `engine/world/src/world_snapshot.cpp:420, 466` (`li.value_ >= localCount` guards) — good bounds-checking on the local index, but the raw entity ids read via `readVarUInt()` (`:416-420`, `:1171-1176` in the diff) are cast straight to `EntityId` with no range check against `IR_ENTITY_ID_BITS`; a corrupted file with an oversized id would silently truncate on the `static_cast`. Low priority given the collision/dup checks in Phase 2 still catch the common case, but worth a defensive `IR_ASSERT` or explicit range check if you want Phase 2 to reject it cleanly instead of truncating first.

### Praise

- The type-erasure design (`SaveRegistry`/`SaveComponentEntry`) cleanly separates "does `C` opt in" (P1's compile-time trait) from "how do the bytes look" (`SaveSerialize<C>`) from "what does the runtime walker do" (this PR) — the walker/loader compile once regardless of how many components register, and an opted-out component never instantiates a serializer it doesn't need.
- Projection-merge + the resetGameplay-mirroring exclusion set is a clean way to make the load contract collision-free "by construction" for the intended same-session round trip; the test suite's explicit coverage of criterion 6 (byte-identical double-save) and the recoverable-error paths (5a-5d) is genuinely thorough for a P2 mechanism slice.

- [ ] A regression test for the watermark-ordering bug: in a fresh `EntityManager`, create some gameplay entities + a singleton, save, `destroyAllEntities()`, then load into the SAME session state such that the singleton's cache-miss lazy-create id would numerically collide with a restored archetype id — assert no cross-wiring (or, simplest: reorder the watermark advance and add a test that fails without the fix).
- [ ] A regression test for the partial-mutation bug: corrupt a column's bytes (not the header) in an otherwise well-formed file and assert `loadWorld` either fully rejects it before mutating anything, or that the resulting world state is fully self-consistent (no column/length_ mismatch) if partial application is intentionally kept.
- [ ] Linux `linux-debug` build + gtest — acceptance criterion #7 is still unchecked in the PR body (macOS-only so far); this stacked PR touches `engine/world/CMakeLists.txt` + `test/CMakeLists.txt`, so cross-host smoke is warranted regardless of the fix.

Opus recheck required: touches engine/entity/ (archetype-node mutation, entity-id restore) and engine/world/ (world-snapshot save/load) — please confirm the two Needs-fix findings (Phase 3 partial-mutation on decode failure; watermark-advance-vs-singleton-lazy-create ordering) and their fixes once the author addresses them.

jakildev (commented):
## Opus recheck — persist: IRWS world snapshot loader (#2213)

Stacked on #2225 (now `fleet:approved`, same author). Reviewing the child diff only — upstream is approved separately. Sonnet flagged two needs-fix items on `engine/world/` load + `engine/entity/` id restore and asked for an Opus recheck of the invariants. **I confirm both**, and one is materially more likely than Sonnet estimated.

**Verdict:** needs-fix (confirming Sonnet's findings — both stand)

### Confirmed — Finding 1: Phase 3 partial-mutation on decode failure

Confirmed against the code. Phase 1 (`world_snapshot.cpp:437-441`) only records each column's `byteLength_`/`dataOffset_` and `seek()`s past the bytes — it never decodes them, and Phase 2 (`:507-528`) validates only id collision/uniqueness. So a length-valid-but-corrupt column (disk corruption, half-written file, hand-edited bytes) is not caught until Phase 3's `entry->appendRow_(r, col)` (`:558`). By then `restoreEntitiesBatch` (`:541`) has already spliced this node's entities — and every earlier `StagedArchetype` — into the live graph, so `return loadFail(rowStatus)` (`:560`) leaves this node's current column short of `node->length_` while its sibling columns are full: any later walk over that node reads the short column out of bounds. The `IR_ASSERT` at `:565` does not protect against this — the `return` at `:560` fires before it, and it compiles out in release anyway. This contradicts the header's documented "zero world mutation on error" contract, so it's a legitimate needs-fix rather than a nit. Sonnet's two suggested remedies (decode into scratch during Phase 2 as the real gate, or stage Phase 3 and splice on full success only) both close it.

### Confirmed + upgraded — Finding 2: watermark advance ordered after singleton lazy-create

Confirmed, and I disagree with Sonnet only on severity — this is stronger than "plausible." `advanceEntityIdWatermark(watermark)` runs last (`:591`), but `entity_manager.hpp:128-131` explicitly calls the watermark advance "the whole of restore collision safety." The singleton loop (`:571-589`) runs before it: `getOrCreateSingletonEntity_()` → `getOrCreateSingleton<C>()` (`entity_manager.hpp:502-515`) lazy-creates on a fresh-session cache miss via `createEntity(Component{})`, which draws off the still-**unadvanced** `m_nextEntityId` (`allocateEntityIdAtomic`, `entity_manager.cpp:55-65`).

In the primary intended path — boot fresh, immediately `loadWorld()` — engine startup is deterministic, so this session's infra/component-backing entity count equals the saving session's. That means the first restored gameplay id equals the value `m_nextEntityId` currently holds, and the first lazy-created singleton draws **exactly** a restored id, not merely a nearby one. `allocateEntity` (`:46-53`) discards `m_entityIndex.emplace`'s bool, so the collision is silent and the singleton's `readIntoEntity_` (`:585`) cross-wires its component onto an already-restored gameplay entity's record. No assert, no error. Effectively guaranteed, not remote, whenever any saved singleton needs lazy-creation on load.

Sonnet's fix is correct: advance the watermark immediately after Phase 2 succeeds and before any Phase 3 mutation. It's safe with respect to the archetype loop (that path plants explicit ids via `restoreEntitiesBatch` and never allocates), and it composes cleanly with the Finding-1 staging fix.

### Agreeing with Sonnet's nits + test plan

The two nits (singleton carrying an extra opted-in component silently dropped on save; unrange-checked `readVarUInt()` → `EntityId` cast) are real but low-priority — nits, not merge-gates. The three regression tests Sonnet listed are the right coverage: a watermark-ordering collision test, a corrupt-column partial-mutation test, and the Linux `linux-debug` + gtest run (acceptance #7 still unchecked, and this PR touches `engine/world/CMakeLists.txt` + `test/CMakeLists.txt` so the `fleet:needs-windows-smoke` cross-host gate is warranted regardless).

Handing back to the author. Both needs-fix items are confirmed real; fix them with confidence, and the watermark reorder in particular should land a failing-without-the-fix test.

jakildev:
Addressed the opus-recheck feedback (both confirmed needs-fix findings + both nits):

- **Finding 1 (partial mutation on decode failure)** — added a phase-2b decode-validate pass (`world_snapshot.cpp`) that dry-runs `SaveSerialize<C>::read` over every resolved ARCH/SNGL column via the new `SaveComponentEntry::decodeRow_` hook, discarding the values. A length-valid-but-content-corrupt column now aborts *before* phase 3 touches the live graph, honoring the Rule #5 zero-mutation contract.
- **Finding 2 (watermark ordering)** — moved `advanceEntityIdWatermark(watermark)` to immediately after phase-2 validation and before the singleton loop, so a fresh-session singleton lazy-create can't draw a just-restored gameplay id and cross-wire onto it.
- **Nit — unrange-checked id cast** — ARCH and SNGL entity ids are now range-checked against `IR_ENTITY_ID_BITS` (reject, not silently truncate).
- **Nit — singleton single-component restriction** — documented as an intentional restriction in the `world_snapshot.hpp` header block.

Two regression tests added, both fail without the fix: `CorruptColumnAbortsWithZeroMutation` (fallible `C_WsChecked` serializer) and `WatermarkAdvancesBeforeSingletonLazyCreate` (genuine second session with a fresh `EntityManager`). All 13 world-snapshot tests green on macOS; Linux/`fleet:needs-windows-smoke` cross-host gate still applies. Reviewer please re-evaluate the new commit. — worker

jakildev (commented):
## Review — persist: IRWS world snapshot container + walker + loader (#2213)

Stacked on #2225 (`fleet:approved`, still open). Reviewing the child diff only — upstream is approved separately.

**Verdict:** approve

### Prior-review resolution

- ✅ `engine/world/src/world_snapshot.cpp:531-569` (Phase 3 partial mutation on decode failure) — fixed at `c031d534`. A new Phase 2b dry-run-decodes every resolved ARCH/SNGL column (`SaveComponentEntry::decodeRow_`, `save_registry.hpp:113-115`) via the exact same `SaveSerialize<C>::read` call `appendRow_` uses (confirmed identical by reading both closures side by side), before Phase 3 touches the live graph. I reverted just the three production files (`world_snapshot.cpp`, `save_registry.hpp`, `world_snapshot.hpp`) back to the pre-fix commit while keeping the new test, rebuilt, and confirmed `WorldSnapshotTest.CorruptColumnAbortsWithZeroMutation` genuinely fails without the fix (`m_em.getLiveEntityCount()` == 5 instead of the expected 0 — the exact partial-mutation bug) and passes with it restored.
- ✅ `engine/world/src/world_snapshot.cpp:591` (watermark-advance-vs-singleton-lazy-create ordering) — fixed at `c031d534`. `em.advanceEntityIdWatermark(watermark)` now runs immediately after Phase 2/2b validation and before the singleton restore loop. Verified `restoreEntitiesBatch` (`entity_manager.cpp:67-79`) never calls the allocator (`allocateEntityIdAtomic`), so moving the watermark advance ahead of it is safe. Same revert-and-rerun test: `WorldSnapshotFreshSession.WatermarkAdvancesBeforeSingletonLazyCreate` fails without the fix (`EXPECT_GT(liveSingleton, maxGameplay)` fails: `257` is not `>` `264` — the lazy-created singleton lands below the max restored gameplay id, and the follow-up assertion on the gameplay entity's `C_WsPos` also fails, confirming the cross-wire) and passes with it restored.
- ✅ Nit — singleton extra-component silent drop — addressed as a doc-only change: `world_snapshot.hpp:56-69` now documents the single-component singleton restriction explicitly in the header's "Singleton save restriction" block, as requested (not a behavior change, matches the nit's ask).
- ✅ Nit — unrange-checked `readVarUInt()` → `EntityId` cast — fixed at `c031d534`: both ARCH (`:420-429`) and SNGL (`:478-484`) entity-id reads now reject (not truncate) any value `> IREntity::IR_ENTITY_ID_BITS` before the `static_cast`, exactly the bound the nit named.

### Independent verification

- Built `IrredenEngineTest` on macOS at the PR's actual HEAD (`c031d534`) and ran the full `WorldSnapshotTest` + `WorldSnapshotFreshSession` suites: **13/13 green**, matching the author's claim.
- Did the revert-only test described above (production code → `c031d534~1`, tests kept at `c031d534`) to confirm both new regression tests are load-bearing and not tautological — both fail on the pre-fix code with the exact failure modes the findings described, and both pass on the fix.
- `clang-format --dry-run --Werror` on all four touched files — clean.

### Nits

None new.

### Praise

- The Phase 2b decode-gate reuses `SaveSerialize<C>::read` verbatim via a new `decodeRow_` hook rather than writing a second, parallel "is this valid" predicate — so Phase 2b and Phase 3 can never drift out of sync on what counts as a valid row.
- The regression tests are genuinely adversarial, not just re-running the happy path: `C_WsChecked` is a purpose-built fallible-serializer test component (a trivially-copyable raw image can't reproduce a decode failure at all), and the watermark test drives a real second `EntityManager` session rather than reusing the save-time manager, which is the only way to reproduce the fresh-session collision.

- [x] Verified independently — macOS `IrredenEngineTest`, `WorldSnapshotTest` + `WorldSnapshotFreshSession`: 13/13 green; both new regression tests confirmed to fail without the fix.
- [ ] Linux `linux-debug` build + gtest — still outstanding (this stacked PR touches `engine/world/CMakeLists.txt` + `test/CMakeLists.txt`); `fleet:needs-windows-smoke` is already applied, carry it forward.

Escalation: please Opus-recheck before merge (touches: `engine/world/`, the same module the prior needs-fix findings were on) — specifically confirm the Phase 2b placement genuinely closes the partial-mutation gap for every column shape (not just the single-archetype-boundary case the test exercises) and that advancing the watermark before Phase 3 has no other ordering hazard I didn't check.

Opus recheck required: touches engine/world/ (the exact module the prior confirmed needs-fix findings were on) — please confirm both fixes hold under review, not just under the two tests I ran.

jakildev (commented):
## Opus recheck — persist: IRWS world snapshot loader (#2213)

Stacked on #2225 (`fleet:approved`, still open — upstream reviewed separately). Reviewing the child diff at `c031d534`. Sonnet re-reviewed the author's fix commit and approved, escalating two specific questions for me to confirm: (1) does the new Phase 2b decode-gate close the partial-mutation gap for **every** column shape, not just the single-archetype-boundary case its test exercises; (2) does advancing the watermark before Phase 3 introduce any other ordering hazard. **Both confirmed — the fixes hold.**

**Verdict:** approve

### Finding 1 (Phase 3 partial mutation) — fix holds for every column shape

The decode-gate is a faithful mirror of Phase 3, not an approximation, so its coverage is structural rather than test-dependent:

- `decodeRow_` (`save_registry.hpp:113-115`) is literally `SaveSerialize<C>::read(r).status_` — the **exact same** decode call `appendRow_` (`:103-113`) and `readIntoEntity_` (`:126-135`) make. The only thing Phase 3's `appendRow_` adds on top is an infallible `dataVector.push_back` after a successful read. So a column that decodes clean in Phase 2b is byte-for-byte guaranteed to append clean in Phase 3 — there is no decode outcome that differs between the two passes.
- Phase 2b (`world_snapshot.cpp:530-585`) walks the identical nested structure Phase 3 does: for ARCH, every `StagedArchetype` × every resolved column (`localIndices_.size()`) × every row (`entityIds_.size()`), each on a fresh `MemoryBinaryReader` seeked to `column.dataOffset_`; for SNGL, every resolved singleton column once. Phase 3 (`:614-666`) iterates the same counts with the same per-column fresh reader. Unresolvable columns (`entry == nullptr`) are `continue`-skipped in both passes and seek-skipped by byte length, so they stay consistent. Multi-column archetypes, multiple archetypes, and singletons are all covered by construction — the single-archetype test is just one instance of the general invariant.

So the "zero world mutation on error" (Rule #5) contract is genuinely restored: any length-valid-but-corrupt column now aborts in Phase 2b before the first live write.

### Finding 2 (watermark-vs-singleton-lazy-create ordering) — fix holds, no other hazard

Confirmed against the primitive bodies (both added by the parent stack, `entity_manager.cpp`):

- `advanceEntityIdWatermark` (`:82-89`) is **advance-forward-only**: `if (watermark > m_nextEntityId.load()) store(watermark)`. It is monotonic and can never move the allocator backward, so relocating it earlier in `loadWorld` is safe with respect to any prior session state — worst case it's a no-op. (Loading into a session that already allocated past the file's watermark stays correct: the no-op leaves the allocator above the restored range, and Phase 2 already rejected any restored id that collides with a live entity.)
- `restoreEntitiesBatch` (`:67-80`) only `m_entityIndex.emplace`s the explicit masked ids and bumps `node->length_`/`m_liveEntityCount` — it never calls `allocateEntityIdAtomic`, so the archetype loop draws no ids. Confirmed nothing between the advance (`:602`) and Phase 3 allocates.
- Result: by the time the singleton loop's `getOrCreateSingletonEntity_` → `createEntity` → `allocateEntity` → `allocateEntityIdAtomic` (`:47`) does a `fetch_add`, `m_nextEntityId` already sits above every restored gameplay id, so the fresh singleton id lands strictly above the restored range — the silent cross-wire the finding described can no longer occur.

No other ordering hazard: the watermark is the save-time `m_nextEntityId`, strictly greater than every id allocated before the save, so every restored gameplay id is `< watermark` and advancing to it dominates the whole restored range.

### Nits (prior) — resolved

- Singleton extra-component silent-drop — documented as an intentional restriction in the `world_snapshot.hpp` header block, as the nit asked (doc-only, correct).
- Unrange-checked `readVarUInt()` → `EntityId` cast — both ARCH (`:423-429`) and SNGL (`:478-484`) now `loadError`-reject any id `> IR_ENTITY_ID_BITS` before the `static_cast`, closing the silent-truncation path.

### Nits (new)

None.

### Praise

- Reusing `SaveSerialize<C>::read` verbatim through a dedicated `decodeRow_` hook (rather than a second, parallel "is-this-valid" predicate) is the right call — it makes Phase 2b and Phase 3 structurally unable to drift on what counts as a valid row, which is exactly why the coverage argument above generalizes.
- Sonnet's revert-and-rerun verification of both regression tests (production files reverted to `c031d534~1`, tests kept) is the correct way to prove they're load-bearing; both fail on the pre-fix code with the exact failure modes and pass on the fix. I re-derived the same conclusion from the code.

### Remaining before merge

- Linux `linux-debug` build + gtest is still the open item (this stacked PR touches `engine/world/CMakeLists.txt` + `test/CMakeLists.txt`); `fleet:needs-windows-smoke` is already applied and carries forward — nothing further from review.

jakildev:
Merger: base PR #2225 merged. Re-targeted this PR from `claude/2212-save-trait-inventory` to `master` and rebased the child commits onto current master tip (`--force-with-lease`). The diff against master is now clean; any verdict label (`fleet:approved` / `fleet:needs-fix` / `fleet:has-nits`) from the pre-merge review has been preserved. CI will re-run.

— fleet merger

PR #2231, opened 2026-07-04
fleet: centralized cross-device polling — leader/follower (#2220)

### LINKED ISSUE #2220: fleet: centralized cross-device polling — one poller per account (Q2)
**Model:** opus
**Part of epic:** #1394
**Plan file:** `.fleet/plans/issue-1394.md` (umbrella context)
**Blocked by:** #2219

## Scope

Static leader/follower split via host.toml (poll_role/poll_port): the leader polls GitHub and serves the whole ~/.fleet/state bundle over stdlib HTTP with generated_at as ETag; followers conditional-GET it (reusing Q1's helper), preserve the leader's generated_at verbatim, recompute host-local fields, keep per-host reconcile, and self-poll when the leader is unreachable/stale. cleanup/ingest go leader-only. Dispatcher staleness switches from mtime to in-file generated_at.

## Acceptance criteria

3 hosts idle: only the leader's API counters move, followers make zero gh read calls; follower state schema-identical + fresh; leader death detected within the existing staleness window; off-LAN follower degrades to solo self-polling.

## Dependencies

Chain tail of the polling pair.

## References

The canonical plan is this issue's `## Plan` comment (posted at filing, architect-planned with the human in the loop — 2026-07-03 triage session). Umbrella: #1394.

jakildev:
## Summary
- Collapse N per-host `fleet-state-scout` GitHub pollers into one authoritative **leader** plus **followers** that consume the leader's already-fetched `~/.fleet/state` bundle over the LAN (#1394 Q2). Idle-fleet GitHub quota drops from N× to 1×; every consumer stays oblivious.
- New `fleet_poll_topology.py`: host.toml `[fleet]` config, `generated_at` staleness, bundle build/parse/apply, a stdlib `ThreadingHTTPServer` leader endpoint (`generated_at` as ETag), and the follower's conditional-GET LAN hop.
- `fleet_gh_poll.py`: general `conditional_get_url()` for the LAN hop (reuses Q1's If-None-Match machinery, no gh token).
- `fleet-state-scout`: `build_state()` routes leader (poll GitHub) vs follower (consume bundle). A follower **preserves the leader's `generated_at` verbatim**, re-derives host-local `clone_freshness` + `repos.path`, recomputes its own projections + triggers, and **self-polls when the leader is unreachable/stale**. Global mutations (`cleanup --gh`, `queue-ingest`) run only on the authoritative poller; host-local `reconcile --apply` stays per-host.
- `fleet-dispatcher`: `scout_unhealthy()` keys staleness off the in-file `generated_at` (mtime fallback) so a follower re-writing a dead leader's frozen timestamp can't mask the death.
- Config + docs: host.toml `[fleet]` keys in `setup-windows.sh` / `fleet-up.conf.sample`, a scout-launch poll-role banner in `fleet-up`, and `FLEET-CACHE.md` / `FLEET-RUNTIME.md` sections.

## Test plan
- [x] `python3 -m unittest discover -s scripts/fleet/tests` — 578 tests green (22 new in `test_poll_topology.py`).
- [x] `ruff check scripts/` clean (CI gate).
- [x] Real leader↔follower HTTP round-trip: 200 → 304 → 200-after-tick → leader-down → self-poll.
- [x] Full `tick_once()` follower smoke on the live `state.json`: all projections written, `generated_at` preserved, caches mirrored, and only host-local `reconcile` spawned (never `cleanup`/`ingest`), with zero `collect_state`/`refresh_all_details` calls.
- [ ] Multi-host acceptance (1 leader + 2 followers): confirm only the leader's API counters move and followers make zero gh reads — requires the real 3-host fleet.

## Notes for reviewer
- Deviation from the plan's "bundle includes projections": projections are a deterministic, host-agnostic function of `state.json`, so the follower recomputes them via its own slicer pass (which it must run anyway for host-local triggers) rather than shipping redundant copies. Bundle carries `state.json` + `prs/`/`diffs/`/`issues/` only.
- `_BundleServer.server_bind` skips `HTTPServer`'s `socket.getfqdn()` reverse-DNS lookup — it blocks ~30 s binding `0.0.0.0` on macOS and would stall the leader's scout startup.
- The LAN hop reuses Q1's conditional-request helper (Q1 shipped as #2227; this PR was rebased onto master after it merged, so it is no longer stacked).

Closes #2220

### DIFF
--- a/.fleet/plans/issue-2220.md
+++ b/.fleet/plans/issue-2220.md
@@ -1,0 +1,71 @@
+# Plan: quota Q2 — centralized cross-device polling (one poller per account)
+
+- **Issue:** #2220
+- **Model:** opus
+- **Date:** 2026-07-03
+- **Epic:** #1394 — see `.fleet/plans/issue-1394.md` for full context
+- **Blocked by:** #2219
+
+## Scope
+
+Collapse N per-host `fleet-state-scout` daemons into **one authoritative GitHub poller per account** plus N−1 followers that consume the leader's already-fetched state over the LAN instead of calling `gh` themselves. In steady-state idle polling, only the leader's per-account REST/GraphQL counters move; followers make zero GitHub read calls yet keep a byte-identical `~/.fleet/state/` so every consumer (dispatcher, roles, `fleet-pr`/`fleet-issue` wrappers) stays oblivious.
+
+Out of scope (stays per-host, unchanged): all **mutations** — claim label add/remove, PR/issue comments, label transitions, `gh pr merge/edit/review`, ingest label writes. These are cheap and correctly host-scoped. This ticket touches only the **polling reads** in `collect_state()`. Depends on **Q1** (conditional-REST poller core + ETag cache); Q2 reuses that conditional-request helper for the intra-fleet LAN hop, which is why it is Blocked-by-Q1.
+
+## Verified current state (the shareable-vs-host-local split)
+
+The scout's per-tick snapshot is assembled in `collect_state()` and written atomically to `~/.fleet/state/state.json` in `tick_once()`.
+
+- **GitHub-derived → shareable, byte-identical on every host:** the eight `fetch_specs` per repo — `prs, needs_plan, plan_review, human_approved, closed_fleet_queued, recent_merged_prs, tasks, epics` — plus the per-item detail caches `prs/`, `diffs/`, `issues/` refreshed by `refresh_all_details()`. Every one is a `gh` call against the two repos. These are exactly the reads that multiply per host. Claim ownership is likewise **GitHub-derived**: it is parsed back out of the `fleet:claim-<host>-<agent>` labels in `fetch_task_queue`, not from any local tracker — so the *observation* of claims is shareable even though claim *writes* are host-local.
+- **Host-local → must stay per-host:** `repos.<key>.path`, the local clone path; `clone_freshness`, a `git rev-parse` of *this host's* checkout vs its own `origin/master`; and everything derived downstream per host — the projections/slices, `seen-hashes/`, and per-role `triggers/`, because each host must fire triggers for *its own* running roles.
+- **Reconcile locality (the sharp edge):** `tick_once` spawns three side-effects on projection change — `fleet-claim cleanup --gh` + `fleet-claim reconcile --apply` and `fleet-queue-ingest`. `cleanup` and `ingest` are **global** (stale-label reaping, human:approved→queued); running them on all N hosts multiplies mutation calls and races. `reconcile --apply` is conservative **host-local** + TTL-gated auto-fixes — it reconciles *this host's* own claim tracker, so it must keep firing per-host. These fire only on projection *change*, so they never move counters during idle polling.
+- **Existing scout-down signal to extend:** consumers already treat `generated_at` older than ~5 min as "scout is down" (`docs/agents/FLEET-CACHE.md`, `docs/agents/FLEET-RUNTIME.md`). The dispatcher watchdog, however, keys off **local file mtime** (`scout_unhealthy()` → `STATE_FILE.stat().st_mtime`, `SCOUT_STALENESS_SECONDS = 120`). Host identity is already canonical via `host_from_uname`/`derive_host` (`fleet-claim`); `~/.config/irreden/host.toml` is the established per-host config surface (`setup-windows.sh`).
+
+## Approach (single committed mechanism)
+
+**Follower-pull over a leader-served HTTP endpoint, conditional GET, static leader flag.**
+
+1. **Leader selection — static, boring.** Add `[fleet] poll_role = "leader" | "follower"` (default `"leader"`) and `poll_port` (default `8477`) to `~/.config/irreden/host.toml`. The scout reads it once at startup. No election, no heartbeat contest — one line per host, matching fleet's static-config conventions. Home desktop = leader; laptop/second host = follower.
+
+2. **Leader path — unchanged fetch + serve.** The leader runs `collect_state()` exactly as today and additionally starts a tiny stdlib `http.server` thread (daemon thread, torn down on the existing `SIGTERM`/`TERMINATE` path) bound to `0.0.0.0:poll_port`, serving a **read-only bundle of the whole `~/.fleet/state/` dir** (state.json + projections + `prs/`/`diffs/`/`issues/`) as one payload with an **ETag equal to the snapshot's `generated_at`**. Everything the followers' `fleet-pr`/`fleet-issue` wrappers read is included, so followers never fall back to `gh` on a per-item drill-in.
+
+3. **Follower path — consume, don't fetch.** On follower hosts, replace the `collect_state()` GitHub fan-out with a **conditional GET** (reusing Q1's ETag/If-None-Match helper) against `http://<leader-host>:<poll_port>`. A `304` (nothing changed) costs one tiny LAN round-trip and **zero GitHub quota**; a `200` yields the fresh bundle, which the follower unpacks and writes into its own `~/.fleet/state/` via the existing `write_atomic()` — **preserving the leader's `generated_at` verbatim** (do not restamp). The follower then still runs the local half of `tick_once` — projections, slices, `triggers/` — off the merged state, so consumers and local roles behave identically. `cleanup`/`ingest` spawns are guarded to run only when this host is the authoritative poller for the tick (leader, or a follower currently self-polling). `clone_freshness` and `repos.<key>.path` are recomputed locally (never taken from the bundle).
+
+4. **Leader failure / staleness — extend the existing generated_at signal.** Because the follower writes the leader's `generated_at` untouched, a dead leader means its `generated_at` stops advancing and every consumer's existing 5-min guard trips automatically — *provided staleness is judged by the in-file `generated_at`, not local mtime*. **The one required consumer change:** switch `fleet-dispatcher`'s `scout_unhealthy()` from `st_mtime` to reading the in-file `generated_at`, so a follower that keeps re-writing a *stale* bundle each tick can't mask the death by bumping mtime. Additionally the follower's own tick applies a `LEADER_STALE_SECONDS` threshold: if the leader is unreachable, or its `generated_at` is already stale beyond the window, the follower **self-polls GitHub directly for that tick** (identical to today's solo behavior). This handles both leader-crash and laptop-away-from-home with no election — a follower off-LAN is just a temporary N=1 solo poller, which is exactly the pre-Q2 behavior and introduces no multiplication (the home leader is typically offline when the laptop is away).
+
+**Why this mechanism over the alternatives:** shared FS (NFS/SMB) is fragile across mac + WSL2 + native-Windows path/permission/locking semantics and dies when the laptop leaves the LAN; **syncthing** adds a third-party daemon to babysit on three heterogeneous hosts; **committing state to a git ref every 30s** spends the very GitHub quota being conserved and pollutes history; **rsync-on-tick** needs SSH keys + reachable sshd across three OSes. The **leader-served HTTP + conditional-GET follower-pull** wins because it (a) reuses Q1's conditional-request machinery for a free `304` no-change poll — the reason Q2 sequences after Q1; (b) is pure Python stdlib, no new dependency or daemon; (c) is **follower-pull**, so the leader needs no follower inventory (zero-config to add a host) and only one inbound LAN port; (d) degrades cleanly to solo self-polling when off-LAN.
+
+## Affected files
+
+- `scripts/fleet/fleet_poll_topology.py` — **new** module (co-located with the scout like `fleet_gh_poll.py`): `read_poll_config()` (host.toml `[fleet]` section), `generated_at` staleness helpers, bundle build/serialize/apply, `LeaderServer` (stdlib `ThreadingHTTPServer`), and `fetch_from_leader()` (the LAN-hop conditional GET).
+- `scripts/fleet/fleet_gh_poll.py` — add a general `conditional_get_url()` (raw If-None-Match GET against an arbitrary URL, no gh token) so the LAN hop reuses Q1's conditional-request machinery verbatim.
+- `scripts/fleet/fleet-state-scout` — primary. Read `poll_role`/`poll_port` at startup; leader starts the daemon HTTP server thread serving the state-dir bundle with `generated_at` as ETag; follower replaces `collect_state()`'s GitHub fan-out with a conditional GET + local unpack (preserving `generated_at`), retains the local projection/slice/trigger half of `tick_once`, and self-polls on leader-unreachable/stale; guard `cleanup` + `ingest` spawns + `refresh_all_details` to the authoritative-poller tick.
+- `scripts/fleet/fleet-dispatcher` — change `scout_unhealthy()` to key staleness off in-file `generated_at` instead of `st_mtime` (mtime fallback when unparseable); the watchdog re-arm logic is unchanged.
+- `scripts/fleet/setup-windows.sh` and `scripts/fleet/fleet-up.conf.sample` — emit/document the new `[fleet] poll_role` / `poll_port` host.toml keys.
+- `scripts/fleet/fleet-up` — log leader-vs-follower mode at scout launch. Note the one inbound LAN port on the leader.
+- `docs/agents/FLEET-CACHE.md` / `docs/agents/FLEET-RUNTIME.md` — document the leader/follower model, the LAN-hop conditional GET, the "follower preserves leader `generated_at`" rule, and the staleness-via-`generated_at` clarification.
+- `scripts/fleet/tests/test_poll_topology.py` — follower with a mocked leader endpoint produces byte-identical `state.json`; a `304` yields zero `gh` invocations; leader-unreachable falls back to self-poll; a frozen leader `generated_at` trips the staleness guard.
+
+## Acceptance criteria
+
+- With 3 hosts up (1 leader + 2 followers) idling on unchanged repos, **only the leader's per-account REST/GraphQL counters advance**; both followers make **zero `gh` read calls** across a full poll interval.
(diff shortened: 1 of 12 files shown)


jakildev (commented):
## Review — fleet: centralized cross-device polling — leader/follower (#2220)

Stacked on #2227 (`fleet:approved`, still open). Reviewing the child diff only — upstream is approved separately.

**Verdict:** approve

### Nits

- `scripts/fleet/fleet-epic-status:97-103` (`load_state_cache`) — freshness is still judged by file **mtime**, not the in-file `generated_at` this PR establishes as the authoritative staleness signal (and that `FLEET-RUNTIME.md`'s new "Judge staleness by the in-file `generated_at`, **not** the file's mtime" guidance now instructs every consumer to use). Under Q2 a follower rewrites `state.json` — bumping mtime — on every ~30s tick even while relaying an aging leader snapshot, for up to `leader_stale_seconds` (default 90s) before the follower's own self-poll fallback kicks in. Bounded, so not a correctness break (the self-poll fallback caps real content drift at ~90-120s, well under this script's 300s window) — but worth aligning with the new convention the same way `fleet-dispatcher`'s `scout_unhealthy` was just fixed, so this script doesn't quietly diverge from the doc's stated invariant.

### Praise

- The staleness handling is genuinely self-healing: a follower consuming a frozen/wedged leader (same etag/`generated_at`, whether via a 200 or a 304) re-checks `is_stale` against wall-clock time every tick and falls back to a real GitHub self-poll once past `leader_stale_seconds` — so the "dead leader propagates a frozen timestamp forever" failure mode the docs warn about can't actually manifest; it's bounded by design (`fleet_poll_topology.py:210-215`, `fleet-state-scout` `_fetch_follower_state`'s 200 and 304 branches).
- `_BundleServer.server_bind` skipping `HTTPServer`'s reverse-DNS lookup (called out in the PR body) is a real, non-obvious fix — that lookup is known to block ~30s on macOS when binding `0.0.0.0`, which would have stalled scout startup on every leader host.
- Test coverage is thorough: config parsing (including the pre-3.11 TOML fallback path), staleness edge cases, bundle build/parse/prune round-trip, a **real** HTTP leader↔follower round-trip (200 → 304 → 200-after-tick, plus leader-down), and `build_state()` routing tests that assert **zero** `collect_state` (GitHub) calls on a served follower tick.
- Global-mutation gating (`fleet-claim cleanup --gh`, `fleet-queue-ingest`) correctly stays authoritative-poller-only while `reconcile --apply` still runs per-host — matches the PR body's stated design and avoids the N× re-multiplication Q2 exists to collapse.

- [x] Verified independently: `python3 -m unittest discover -s scripts/fleet/tests` — 578/578 green, matching the PR's claimed count.
- [x] `ruff check scripts/` — clean.
- [ ] The PR body's own remaining open checkbox (multi-host acceptance: 1 leader + 2 followers, confirming only the leader's API counters move) requires the real 3-host fleet and can't be verified from this review.

Escalation: please Opus-recheck before merge (touches: new concurrency — a `threading.Lock`-guarded `ThreadingHTTPServer` daemon thread serving fleet state to LAN followers — and this is fleet-wide infrastructure where a subtle bug has broad blast radius across every host's poll state).

Opus recheck required: introduces new concurrency (background HTTP server thread + lock-guarded shared state) and is fleet-wide polling infrastructure; please confirm the leader/follower staleness/self-poll handling and thread-safety are sound.

jakildev (commented):
## Opus recheck — fleet: centralized cross-device polling — leader/follower (#2220)

Stacked on #2227 (`fleet:approved`, still open); base is #2227's head branch. Reviewed the child diff only. Sonnet escalated this for Opus recheck on two grounds — new background-thread concurrency (a `threading.Lock`-guarded `ThreadingHTTPServer` daemon) and fleet-wide blast radius. I focused the pass there: thread-safety of the shared bundle, the leader/follower staleness handling, and the resolve-pipeline split. **Verdict: approve** (Sonnet's single nit carried below).

### What I independently confirmed (the Opus-only concerns)

- **Bundle thread-safety is sound.** The only state shared between the HTTP handler threads and the main scout thread is `_BundleServer._etag`/`_payload`, and both `set_bundle` (main thread, tick tail) and `get_bundle` (handler threads) take `self._lock` and swap/return the pair atomically. Because the payload is immutable `bytes` and the (etag, payload) pair is always read/written together under the lock, a handler can never serve an ETag that mismatches its payload — the one race that would matter here. No other mutable state is touched by handlers (`do_GET`/`do_HEAD` only call `get_bundle`).
- **No init-order race.** `_BundleServer.__init__` binds via `super().__init__` before `_lock`/`_etag` are set, but `TCPServer.__init__` only binds+listens — no handler thread runs until `LeaderServer.start()` calls `serve_forever()`, which happens strictly after `start_leader_server` has seeded the bundle. So the first request can't observe an unlocked/half-built server.
- **Staleness-via-`generated_at` is correct and consistent.** The dispatcher's new `_state_age_seconds()` uses `calendar.timegm(strptime(..., "%Y-%m-%dT%H:%M:%SZ"))` against `time.time()` — both UTC-epoch, so the age is right (and correctly avoids the `time.mktime` local-tz trap), with an mtime fallback only on a malformed/absent `generated_at`. `fleet_poll_topology.now_epoch()`/`is_stale()` use the same UTC basis. The follower's frozen-leader guard is genuinely closed on **both** the 200 and 304 branches (`_fetch_follower_state`), so an alive-but-wedged leader answering a 304 with a frozen timestamp forces a self-poll rather than propagating a lie — I traced this specifically because it's the failure mode the design leans on.
- **The leader-vs-follower resolve-pipeline split is coherent.** A 200 returns `served=True` and the follower skips the resolve pipeline + `refresh_all_details` because the leader already ran them before serving (state.json is served post-resolve, post-body-strip); the follower still runs the host-local projector/slice/trigger tail and overlays `repos.<key>.path` + `clone_freshness`. Bodies are stripped by the leader pre-serve, matching the pre-Q2 "no projector needs pr body" invariant. Cold start can't hit a stale 304: `_FOLLOWER_ETAG` resets to `None` per process, so tick 1 always sends no `If-None-Match` → 200 → caches mirrored.
- **`cleanup --gh` + `ingest` correctly gated authoritative-poller-only; `reconcile --apply` correctly stays per-host.** This matches the plan's stated locality and is the right call — getting it backwards would either strand follower claims or re-multiply the mutations Q2 exists to collapse.
- **Verification:** ran `scripts/fleet/tests/test_poll_topology.py` on macOS in a throwaway worktree at the PR head — **22/22 green**, including the real 127.0.0.1 leader↔follower 200→304→200 round-trip, leader-down, stale-bundle self-poll, and 304-stale-on-disk self-poll. Matches Sonnet's independently-run 578/578 full-suite result.

### Nit (Sonnet flagged; I confirm)

- `scripts/fleet/fleet-epic-status` `load_state_cache` still judges freshness by file **mtime** rather than the in-file `generated_at` this PR establishes as the authoritative signal. On closer read I agree with Sonnet's classification as a **nit, not a correctness break**: whenever a follower is alive-and-writing (fresh mtime), its own `leader_stale_seconds` (90s) guard means the content it relays is either <~120s old or a fresh self-poll — comfortably inside epic-status's 300s window — so mtime-freshness and content-freshness stay coupled for that consumer. Worth aligning with the new `generated_at` convention for consistency (same fix shape as this PR applied to `fleet-dispatcher`), but it does not gate merge.

### Non-blocking notes

- The endpoint is unauthenticated and binds `0.0.0.0`. This is a deliberate, plan-documented tradeoff (read-only, non-secret state; no `gh` tokens in the bundle) for a home LAN, and the human approved the design — flagging only for visibility, not as a concern.
- `do_GET` ignores `self.path` and always serves the one in-memory bundle, so there's no path-traversal surface despite the `/state` URL being cosmetic — a nice side benefit, not an issue.

Sonnet flagged the concurrency/staleness as needing Opus confirmation; on closer read I confirm the thread-safety, the dual-branch frozen-leader guard, and the resolve-split are all sound. Nothing merge-blocking.

jakildev:
Addressed nit: `fleet-epic-status`'s `load_state_cache` now judges cache staleness from the in-file `generated_at` (falling back to mtime only when missing/unparseable), mirroring the fix already applied to `fleet-dispatcher`'s `_state_age_seconds`. Manually verified all three branches (stale-generated_at/fresh-mtime, fresh-generated_at, malformed-generated_at-fallback). Full suite still 578/578 green, `ruff check scripts/` clean. — worker

jakildev:
Merger: base PR #2227 merged. Re-targeted this PR from `claude/2219-conditional-rest-polling` to `master` but the rebase onto current master conflicts. The child commits cannot be replayed cleanly — worker will resolve semantically.

Conflicted files:
- `scripts/fleet/fleet_gh_poll.py` — master: `361ce322 fleet: conditional-REST scout polling (ETags, gated GraphQL) (#2219) (#2227)`; PR: `35028539 fleet: centralized cross-device polling — leader/follower (#2220)`

— fleet merger

jakildev:
Resolved semantic conflict via inherited-prefix drop. This PR was stacked on `claude/2219-conditional-rest-polling`, which merged to master as #2227 (`361ce322`); the branch still carried its own stale copies of the #2219 commits (`1438d9a9`, `e34c0bd5`), and a naive `git rebase origin/master` collided replaying those inherited commits onto master's now-identical #2219 code. Dropped the inherited prefix with `git rebase --onto origin/master e34c0bd5`, replaying only the genuine #2220 commits (plan, leader/follower centralized polling, epic-status staleness fix). The one merger-flagged file, `fleet_gh_poll.py`, is a pure additive function (`conditional_get_url`) that layers cleanly on master's #2219 `DEFAULT_TIMEOUT_SECONDS`/`USER_AGENT` — no true semantic conflict, just the duplicated-inherited-commit collision. Verified: `test_poll_topology.py` (22) + `test_gh_poll_conditional.py` (6) green, 26/27 fleet tests pass (the 1 failure, `test_fleet_validate_roles.py`, is a pre-existing co-located-module import artifact reproducing identically on clean master, untouched by this diff). Reviewer please re-evaluate the rebased diff. — worker

jakildev (commented):
## Review — fleet: centralized cross-device polling — leader/follower (#2220)

**Verdict:** approve

### Prior-review resolution
- ✅ `scripts/fleet/fleet-epic-status:97-103` (`load_state_cache`) — freshness judged by file mtime instead of the in-file `generated_at`, so a follower under Q2 could read as fresh for up to `leader_stale_seconds` while relaying an aging leader snapshot — fixed at `2d81c1d6` (`_cache_age_seconds`), which mirrors `fleet-dispatcher`'s already-approved `_state_age_seconds` fix byte-for-byte in approach (parse `generated_at`, fall back to mtime only on missing/unparseable).

### What changed since the last review

This PR carries no new author content beyond the nit fix above — it was rebased onto current `master` (a `fleet:resolving-mac-worker-4` conflict-resolution claim that released just before I picked this up) after its stacked parent #2227 merged. `git diff origin/master...HEAD --stat` shows the same 12 files with the same line counts previously reviewed (plus the nit-fix's own +21/-2 in `fleet-epic-status`) — no unexpected churn, no merge-resolution artifacts, no extra commits. The concurrency-sensitive core (`fleet_poll_topology.py`'s `ThreadingHTTPServer` leader, the lock-guarded bundle swap, `fleet-state-scout`'s leader/follower split) that the Opus recheck already vetted is untouched by this delta.

### Verification performed
- `git log origin/master..HEAD --oneline` — 3 commits (plan, feature, nit-fix), matches the PR body's description; no merge/resolve commit, confirming a clean rebase.
- Read `2d81c1d6` in full: `_cache_age_seconds(data, mtime)` parses `generated_at` via `calendar.timegm(time.strptime(...))`, falls back to `mtime` on `ValueError`/`TypeError` (missing/unparseable), and `load_state_cache` passes the already-parsed `data` through rather than re-reading the file — correct, and identical in shape to `fleet-dispatcher:1648-1666`'s `_state_age_seconds`.
- `python3 -m unittest discover -s scripts/fleet/tests` — 578/578 green, matching the PR's claimed count.
- `ruff check scripts/` — clean.
- No inline review comments outstanding (`gh api .../pulls/2231/comments` — empty).

### Nits

- Neither `_state_age_seconds` (existing, already-approved) nor the new `_cache_age_seconds` has a direct unit test — both are small, low-risk mirrors of the same pattern, so not gating, but a shared helper (both scripts currently duplicate the same ~15 lines) would be easy to unit-test once and avoid the duplication `read_json_retry`'s own docstring already calls out as a known hazard (`~/bin` symlink resolution, #1578).
- PR body still reads "Stacked on: #2227" even though `baseRefName` is now `master` (#2227 merged and this branch was rebased onto it) — stale but harmless, just noting for anyone skimming the body after the fact.

- [x] `python3 -m unittest discover -s scripts/fleet/tests` — verified independently, 578/578 green.
- [x] `ruff check scripts/` — verified independently, clean.
- [ ] The PR body's own remaining open checkbox (multi-host acceptance: 1 leader + 2 followers, confirming only the leader's API counters move) still requires the real 3-host fleet and can't be verified from this review.

Opus recheck not required — the concurrency-sensitive surface (background HTTP server thread, lock-guarded shared bundle, leader/follower staleness handling) is byte-identical to what Opus already reviewed and approved on this same PR; the only new content since then is the nit-fix commit above, which touches a single non-concurrent CLI script and mirrors an already-approved pattern.

jakildev:
Addressed the review nit (shared, unit-tested staleness helper):

- **Extracted** \`state_age_seconds(generated_at, mtime, now=None)\` into \`fleet_poll_topology.py\` (composes the existing \`parse_generated_at\`; \`now\` injectable for tests) — the canonical, tested home for the \`generated_at\`-vs-mtime calc.
- **Unit-tested** it in \`test_poll_topology.py\` (3 cases: generated_at wins over a fresher mtime, mtime fallback on missing/garbage generated_at, wall-clock default now). Full suite 581/581 green (was 578; +3), \`ruff check scripts/\` clean.
- **\`fleet-epic-status\`** now delegates to the shared helper (dropped its local \`_cache_age_seconds\` copy + the now-unused \`calendar\` import). Verified the import chain still resolves through the \`~/bin\` symlink (\`fleet-epic-status --help\` → rc 0), since it already uses the realpath-based \`sys.path\` guard.
- **\`fleet-dispatcher\`** deliberately keeps its inline mirror rather than importing the helper, and now documents why: that heredoc is the scout-dead watchdog, so it stays stdlib-only with no dependency on the fleet module tree resolving (the #1578 \`~/bin\` symlink hazard) and avoids pulling the poll-topology HTTP-server module into a hot dispatcher path. A comment ties the two copies together for future sync.

The second nit (PR body still reading "Stacked on: #2227") is fixed by editing the body directly.

Tagged \`fleet:changes-made\` (kept \`fleet:approved\`): the fix is the reviewer-requested shape, but it adds a new function to the concurrency-sensitive \`fleet_poll_topology.py\` and makes a non-obvious call on the dispatcher watchdog — worth a reviewer glance on the new commit before merge. — worker

jakildev (commented):
## Review — fleet: centralized cross-device polling — leader/follower (#2220)

**Verdict:** approve

### Prior-review resolution
- ✅ `scripts/fleet/fleet-epic-status:97-103` (`load_state_cache`) — freshness judged by file mtime instead of the in-file `generated_at`, so a follower under Q2 could read as fresh for up to `leader_stale_seconds` while relaying an aging leader snapshot — fixed at `2d81c1d6` (verified in the prior re-review), and further improved at `2dbea801` (this commit): the fix is now extracted into a single canonical, unit-tested `state_age_seconds(generated_at, mtime, now=None)` in `fleet_poll_topology.py:222-236`, with `fleet-epic-status` delegating to it (`fleet-epic-status:98-114`) instead of carrying its own copy. `fleet-dispatcher`'s inline mirror (`fleet-dispatcher:1648-1674`) is unchanged logic-wise and now carries a comment explaining why it deliberately stays a stdlib-only duplicate (scout-dead watchdog must not depend on the fleet module tree resolving — the #1578 `~/bin` symlink hazard). Good call keeping that one un-DRY'd for a documented reason rather than importing across that boundary.

### What changed since the last review

Single new commit (`2dbea801`, "extract shared, unit-tested `state_age_seconds` helper") landed after the 00:17:54Z re-review. It touches only the already-resolved nit's surface — no changes to the leader/follower HTTP server, bundle thread-safety, or any other concurrency-sensitive code that Opus already confirmed sound on this same PR.

- `fleet_poll_topology.py:222-236` — new `state_age_seconds()` composes the existing `parse_generated_at()` correctly: `generated_at` wins when parseable, falls back to `mtime` on `None`/malformed input, `now` is injectable for tests. Matches the two call sites' prior inline logic exactly (verified by reading both old copies before the diff removed them).
- `fleet-epic-status:98-114` — `load_state_cache()` now calls the shared helper; `calendar` import correctly dropped (grepped — no remaining references) while `time` stays (still used for `time.sleep` elsewhere in the file).
- `fleet-dispatcher:1648-1674` — comment-only change; the actual staleness calc is byte-for-byte unchanged, so the scout-dead watchdog's stdlib-only independence is preserved as intended.
- `test_poll_topology.py` — three new unit tests: generated_at-wins-over-fresher-mtime, mtime-fallback-on-missing/garbage, wall-clock-default-`now`. Directly covers the case the original nit was about (a follower's mtime looking fresh while `generated_at` is frozen).

Independently verified:
- [x] `python3 -m unittest discover -s scripts/fleet/tests` — 581/581 green (578 + the 3 new tests this commit adds — count matches exactly).
- [x] `ruff check scripts/` — clean.
- [x] `grep calendar scripts/fleet/fleet-epic-status` — no leftover references.

- [ ] The PR body's own remaining open checkbox (multi-host acceptance: 1 leader + 2 followers, confirming only the leader's API counters move) still requires the real 3-host fleet and can't be verified from this review.

Escalation: none. Safe for merge — the concurrency-sensitive surface (background HTTP server thread, lock-guarded shared bundle, leader/follower staleness handling) is byte-identical to what Opus already reviewed and approved on this PR; this commit only touches a non-concurrent CLI script, a pure helper function, and its tests.

Opus recheck not required — this delta only extracts an already-reviewed nit-fix into a shared, unit-tested helper; no new concurrency or fleet-wide-infrastructure surface introduced since Opus's approved pass.

PR #2236, opened 2026-07-04
fleet: fix fleet-run-targets MSYS2/Windows path-scope mismatch

### LINKED ISSUE #2036: fleet-run-targets: Windows /c/… vs C:/… path-scope mismatch breaks --targets listing
**Area:** fleet tooling (`scripts/fleet/fleet-run-targets`, reached via `ir-run --targets`)

**Surfaced by:** native-Windows platform-catchup bring-up (#2034).

## Symptom
On native Windows, `ir-run --targets --plain` / `--plan` prints:
```
fleet-run-targets: scope /c/Users/evinj/src/IrredenEngine/.claude/worktrees/opus-architect is outside C:/Users/evinj/src/IrredenEngine/.claude/worktrees/opus-architect; showing whole tree.
fleet-run-targets: no built executables under whole tree.
```
The two paths are the **same directory** in different spellings — MSYS2 POSIX (`/c/...`) vs Windows (`C:/...`). The scope-containment check compares them as strings, they don't match, and it falls back to "whole tree" + fails to find executables.

## Impact
`ir-run --targets` is unusable for cohort enumeration on Windows. Workaround: derive the demo cohort from `add_executable(...)` in `creations/demos/*/CMakeLists.txt` (now documented in the platform-catchup skill). Low severity (cosmetic + one listing command), but Windows is now the primary validation host, so worth fixing.

## Fix direction
Normalize both paths to one spelling (`cygpath -w` or `cygpath -u`, or realpath) before the containment comparison in `fleet-run-targets`.

## Acceptance
- `ir-run --targets --plain` lists built executables within the current worktree on native Windows.
- No "scope … is outside …" false-positive when invoked from inside the worktree.

jakildev:
## Summary
- `resolve_scope` in `fleet-run-targets` compared the engine root (git-style `C:/Users/x` from `git rev-parse --show-toplevel`) against the scope (`/c/Users/x` from an MSYS2 `$PWD`) byte-for-byte, so the same worktree spelled two ways failed the containment `case` and fell back to a whole-tree scan that found no executables — `ir-run --targets` was unusable on native Windows.
- Added `canonicalize_path_spelling` to `fleet-common.sh` (the shared helper home) and run `root`/`build`/`scope` through it before the containment compare. It collapses `C:/Users/x`, `c:\Users\x`, and `/c/Users/x` to a single canonical `/c/Users/x`.
- The helper is a byte-identity no-op on macOS/Linux paths (no POSIX abs path starts with a `<letter>:` drive) and is bash-3.2 compatible (`tr`, not `${var,,}`) for the `/bin/bash` and MSYS2 hosts.
- Added `scripts/fleet/tests/test_run_targets_path_scope.sh` (16 assertions): the exact `/c/…` vs `C:/…` spelling pair, backslash/case/bare-drive variants, POSIX pass-through, and a `resolve_scope` containment simulation (mixed spellings match; genuinely-outside still reports OUTSIDE).

## Test plan
- [x] `scripts/fleet/tests/test_run_targets_path_scope.sh` — 16/16 pass under bash 5.3 and bash 3.2
- [x] `bash -n` clean on both scripts
- [x] macOS end-to-end no-op preserved: root scope → whole tree; `--scope creations/` scoped without false "outside"; `--scope test/` narrows correctly; `--scope /tmp` still reports "is outside"
- [ ] Windows: `ir-run --targets --plain` lists built executables from inside the worktree with no "scope … is outside …" false-positive (acceptance criteria; not runnable on the macOS author host)

## Notes for reviewer
Authored on macOS, so the Windows acceptance path is verified by the deterministic containment test (feeding the literal `/c/…` and `C:/…` strings) rather than a live Windows run. The fix is pure path-string logic, so the test reproduces the exact failing scenario on any host.

Closes #2036

### DIFF
--- a/scripts/fleet/fleet-common.sh
+++ b/scripts/fleet/fleet-common.sh
@@ -34,3 +34,39 @@ detect_engine_root() {
     fi
     echo "$root"
 }
+
+# canonicalize_path_spelling — normalize a path to one spelling so the two ways
+# the same directory is spelled on native Windows/MSYS2 compare equal in a plain
+# string containment test. `git rev-parse --show-toplevel` yields a Windows drive
+# path (C:/Users/x) while an MSYS2 shell's $PWD yields the POSIX drive form
+# (/c/Users/x); byte-compared they differ, so a naive prefix check reports the
+# scope as "outside" the engine root even when it is the same tree (#2036).
+#
+# Canonical form: /<lowercase-drive>/<rest> with forward slashes, e.g. both
+# `C:/Users/x` and `/c/Users/x` map to `/c/Users/x`. On macOS/Linux every path
+# is returned byte-unchanged: no POSIX absolute path begins with a `<letter>:`
+# drive, and the single-letter-top-dir branch only rewrites an UPPERCASE drive
+# letter (`/C/…` → `/c/…`), a spelling MSYS2's realpath never emits and a real
+# POSIX mount effectively never uses — so the transforms are inert off-Windows.
+# Kept bash-3.2 compatible (tr, not ${var,,}) for the /bin/bash and MSYS2 hosts.
+canonicalize_path_spelling() {
+    local p=$1
+    p=${p//\\//} # backslashes -> forward slashes (C:\Users\x -> C:/Users/x)
+    if [[ $p =~ ^([A-Za-z]):(/.*)?$ ]]; then
+        # drive-letter form (C:/Users/x, c:) -> /c/Users/x, /c
+        local drive rest
+        drive=$(printf '%s' "${BASH_REMATCH[1]}" | tr '[:upper:]' '[:lower:]')
+        rest=${BASH_REMATCH[2]}
+        printf '/%s%s' "$drive" "$rest"
+        return 0
+    fi
+    if [[ $p =~ ^/([A-Z])(/.*)?$ ]]; then
+        # uppercase POSIX drive form (/C/Users/x) -> /c/Users/x
+        local drive2 rest2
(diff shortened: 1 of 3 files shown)


jakildev (commented):
## Review — fleet: fix fleet-run-targets MSYS2/Windows path-scope mismatch (#2236)

**Verdict:** approve

### Praise
- The fix is scoped exactly to the reported bug: `canonicalize_path_spelling` normalizes drive-letter spelling ambiguity before the `resolve_scope` containment `case` block, and is a documented byte-identity no-op on POSIX paths (verified below) — no behavior change for the macOS/Linux fleet hosts.
- The regression test (`test_run_targets_path_scope.sh`) reproduces the exact failing scenario with literal path strings rather than requiring a live Windows run, which is the right call given the author host is macOS — the PR body is upfront that the Windows acceptance line is unverified end-to-end.
- Checked for the same MSYS2 path-comparison pattern elsewhere in `scripts/fleet/` — `fleet-run-targets` is the only script with this containment-`case` shape, so the fix's scope (touching only this file + the shared helper) is correct; no sibling script needs the same treatment.

- [x] Verified independently on this host: `bash scripts/fleet/tests/test_run_targets_path_scope.sh` — 16/16 pass under both `/opt/homebrew/bin/bash` (5.3.9) and macOS system `/bin/bash` (3.2.57), confirming the bash-3.2 compatibility claim.
- [x] `bash -n` clean on both `fleet-common.sh` and `fleet-run-targets`.
- [ ] Windows: `ir-run --targets --plain` from inside an MSYS2 worktree lists built executables with no false "outside" — still the one acceptance-criteria line that can't be verified off-Windows, as the PR body already notes.

Opus recheck not required.

PR #2238, opened 2026-07-04
persist: round-trip + determinism test suite (W-7/W-8) (P4) (#2215)

jakildev:
Stacked on: https://github.com/jakildev/IrredenEngine/pull/2230

Work in progress.

Closes #2215

Issue #2215, opened 2026-07-04
persist: round-trip + determinism test suite (W-7/W-8) (P4)


jakildev:
**Model:** sonnet
**Part of epic:** #667
**Plan file:** `.fleet/plans/issue-667.md` (umbrella context)
**Blocked by:** #2214

## Scope

Test-only: the epic's acceptance harness. 128-entity / 4-archetype / CHILD_OF-structured / singleton fixture; parity = structural spot-checks + save->load->re-save byte-identity; determinism = fnv1a64 digest of same-world double-save and across the load cycle. Sequentially-scoped EntityManagers (single global manager constraint).

## Acceptance criteria

PersistRoundTrip.ParityAcrossFreshWorld, .ChildOfEdgesSurviveLoad, PersistDeterminism.SameWorldSameBytes, .SaveLoadSaveDigestStable all green on linux-debug + macos-debug; GL-free TU.

## Dependencies

Blocks P5 (migrations land against a verified baseline).

## References

The canonical plan is this issue's `## Plan` comment (posted at filing, architect-planned with the human in the loop — 2026-07-03 triage session). Umbrella: #667.

jakildev:
## Plan: persist P4 — round-trip + determinism tests (W-7, W-8)

- **Issue:** #2215
- **Model:** sonnet
- **Date:** 2026-07-03
- **Epic:** #667 — see `.fleet/plans/issue-667.md` for full context
- **Blocked by:** #2214

> **Architect reconciliation note (2026-07-03):** the original draft's fixture used
> `setRelation(PARENT_TO, …)` / `setRelation(SIBLING_OF, …)`. P3's code-verified finding
> (`entity_manager.cpp:515/:535`) is that `setRelation` hard-asserts ("Unsupported relation")
> for everything except `CHILD_OF` — `PARENT_TO`/`SIBLING_OF` are declared but unimplemented
> engine-wide. This plan is corrected accordingly: fixtures build `CHILD_OF` structure only,
> and the relation-coverage test asserts the other two types round-trip as zero-triple no-ops
> (matching P3's acceptance #2). The #199 acceptance clause naming all three types is
> satisfied vacuously for the unimplemented two until they exist.

### Scope

[...]

PR #2239, opened 2026-07-04
persist: component migration registry (W-4) (P5) (#2216)

### LINKED ISSUE #2216: persist: component migration registry (W-4) (P5)
**Model:** opus
**Part of epic:** #667
**Blocked by:** #2215

## Scope

(ComponentId, oldVersion) -> reader-function registry so old snapshots load across schema changes: four-case column dispatch (unknown-name skip+warn; current fast path; older -> migrator or hard MigratorMissing fail; newer -> VersionTooNew), migrators co-located with the trait opt-in, direct per-version readers (no chains), new BinaryIOError::MigratorMissing.

## Acceptance criteria

Worked v1->v2 migration on a real component (defaulted field) proven end-to-end; all three mismatch diagnostics asserted; current-version fast path regresses nothing vs the P4 baseline; clean on both presets.

## Dependencies

Blocks P6 (regen hooks the migration-capable loader).

## References

jakildev:
## Summary
Persist P5 (epic #667, W-4): the `(component, oldVersion) → reader` migration registry, so an older `IRWS` world snapshot loads into a newer build whose component has since bumped `SaveTrait<C>::kSaveVersion` (Save Format Rule #3).

P2 already stamped every `ARCH`/`SNGL` column with a `u32 saveVersion` header (the migration seam), but the loader **ignored it** — always reading at the current layout, which silently corrupts an old column. The load now dispatches on that stamp in the mutation-free phase-2 gate (so any failure aborts with a pristine world, Rule #5):

- **disk == current `kSaveVersion`** → `SaveSerialize<C>::read` fast path (regresses nothing).
- **disk < current** → the registered `SaveMigration<C>` reader, or (miss) a hard `BinaryIOError::MigratorMissing` — the one non-recoverable case.
- **disk > current** → `VersionTooNew`.
- **unknown component name** → skipped by byte length (P2 behavior, unchanged).

### Design
- New `SaveMigration<C>` customization point (`save_migration.hpp`) mirrors `SaveSerialize<C>`; primary template = no migrators. Direct per-version readers, never chained.
- Current + each retired version share one `ColumnReadHooks` shape; `SaveComponentEntry::readerForVersion` returns whichever applies as a uniform handle. The three read hooks fold into a shared `buildReader<C>`, **removing** the prior per-hook duplication in `registerComponent`.
- One-line `BinaryIOError::MigratorMissing`.

### Reconciliation with the plan
Filenames differ from the plan's *assumed* names (it flagged them as such): reconciled against the landed P1–P4 API — `save_migration.hpp` (matches the `save_*` sibling convention, header-only, no new `.cpp`/CMake), and the migrator dispatch lives on `SaveComponentEntry` where the erased hooks already are. Worked example uses a **test-local** component (the plan's committed fallback) rather than adding a permanently-dead field to a real gameplay component.

## Test plan
- [x] `component_migration_test.cpp` (5 cases): worked v1→v2 migration, current-version fast path, `VersionTooNew`, `MigratorMissing` (zero world mutation), migration-composes-with-unknown-skip.
- [x] All existing persist/snapshot/save-trait tests green — 56/56 (`macos-debug`, current-version fast path unchanged).

### Notes
- `optimize` N/A: one-shot save/load deserialization, not a per-frame path (the plan notes the migrators are "allocation-light, no per-frame hot path").
- simplify flagged a **medium** structural overlap — the 4 loader sites share a resolve-reader→seek preamble (largely pre-existing in P4). Reported, not applied: extracting a `resolveSeekedReader` helper would refactor the approved-base loop skeleton and expand this focused diff. Candidate follow-up.

Stacked on: https://github.com/jakildev/IrredenEngine/pull/2238

Closes #2216

### DIFF
--- a/engine/asset/include/irreden/asset/binary_io.hpp
+++ b/engine/asset/include/irreden/asset/binary_io.hpp
@@ -39,6 +39,8 @@ enum class BinaryIOError {
     ChunkOutOfBounds, ///< chunk-table entry points outside the file
     WriteFailed,      ///< fwrite returned fewer bytes than requested
     UnknownTag,       ///< unrecognized value/record discriminant byte (likely a newer writer)
+    MigratorMissing,  ///< a known component's on-disk version has no registered reader/migrator
+                      ///< (world snapshot P5)
 };
 
 /// Status payload returned by void-typed binary operations.
--- a/engine/world/CLAUDE.md
+++ b/engine/world/CLAUDE.md
@@ -274,6 +274,47 @@ Deferring the *parse* to the final phase (which must run after phase 3, since
 live on a failed load — the "no partial world mutation on error" contract
 (Rule #5) is why only the mutation, not the parse, waits for phase 3.
 
+## Component migration registry (persist P5, #2216, epic #667)
+
+`engine/world/include/irreden/world/save_migration.hpp` declares
+`IRWorld::SaveMigration<C>` — the `(component, oldVersion) → reader`
+customization point that lets an old `IRWS` snapshot load into a newer build
+whose component has since bumped `SaveTrait<C>::kSaveVersion` (Save Format
+Extensibility Rule #3). It fills the seam P2 already stamped: every `ARCH`/`SNGL`
+column carries a `u32 saveVersion` header, and before P5 the loader **ignored
+it** — always reading at the current layout, which silently corrupts an old
+column. Now `SaveComponentEntry::readerForVersion(diskVersion)` dispatches four
+cases at load, in the mutation-free phase-2 gate (so any failure aborts with a
+pristine world, Rule #5):
+
+- **disk == current `kSaveVersion`** — the `SaveSerialize<C>::read` fast path
+  (`SaveComponentEntry::reader_`); no migrator lookup. Regresses nothing.
+- **disk < current** — the registered `SaveMigration<C>` reader for that
+  version, or (miss) a hard `BinaryIOError::MigratorMissing`. This is the **one
+  non-recoverable case**: reading old bytes at the current layout can't be
+  recovered from, so a known component at an unmigrated older version errors
+  out rather than degrading. Every other mismatch degrades gracefully.
+- **disk > current** — `BinaryIOError::VersionTooNew` (a future writer),
+  reusing the existing shape, naming the component + both versions.
+- **unknown component name** — resolves to no registry entry; the column skips
+  by byte length (`columnsSkipped_`), never reaching the version dispatch
+  (Rule #1 forward-compat, P2 behavior).
+
+**Direct per-version readers, never chained.** Each `SaveMigration<C>` entry
+decodes exactly its era's bytes straight to a current-build `C` (fields added
+since defaulted, renamed fields remapped); a v3 bump leaves the v1/v2 readers
+untouched. The current version is **not** listed in `SaveMigration<C>` —
+`SaveSerialize<C>::read` owns it; list only the retired `[1 .. kSaveVersion-1]`.
+The registry keys migrators on the current session-local `ComponentId` and the
+disk column key stays `SaveTrait<C>::kSaveName` (Rule #2) — never
+`typeid(C).name()`, which is compiler-mangled and would make a macOS save
+unreadable on Windows. Registration is `if constexpr`-gated on `shouldSave<C>()`
+(like `SaveSerialize<C>`), so an opted-out component never needs a migrator.
+The `SaveMigration<C>` header block has the worked specialization example; the
+generic loader path is proven end-to-end (v1→v2, `VersionTooNew`,
+`MigratorMissing`, unknown-skip compose, current fast path) in
+`test/world/component_migration_test.cpp`.
+
 ## Responsibilities
 
 `World(const char* configFileName)`:
--- a/engine/world/include/irreden/world/save_migration.hpp
+++ b/engine/world/include/irreden/world/save_migration.hpp
@@ -1,0 +1,84 @@
+#ifndef SAVE_MIGRATION_H
+#define SAVE_MIGRATION_H
+
+/// Per-component schema-migration customization point for the ECS world
+/// snapshot (persist P5, #2216, epic #667). P1 (`save_trait.hpp`) decides
+/// *whether* a component is saved and its current `kSaveVersion`; P2's
+/// `save_serialize.hpp` decides *how* the current version's bytes read; this
+/// header decides how a **retired** on-disk version reads.
+///
+/// The `ARCH`/`SNGL` chunks stamp every column with the `kSaveVersion` it was
+/// written at. When an old snapshot is loaded into a newer build whose
+/// component has since bumped its version, the load-time dispatch (see
+/// `save_registry.hpp`'s `SaveComponentEntry::readerForVersion`) needs a reader
+/// for the *disk-side* layout — reading old bytes at the current layout would
+/// silently corrupt (Save Format Extensibility Rule #3). `SaveMigration<C>`
+/// is where a component declares that reader.
+///
+/// The primary template declares **no** migrators — correct for any component
+/// that has never changed its serialized schema (`kSaveVersion == 1`, or a
+/// later version reached purely by additive fields the current
+/// `SaveSerialize<C>::read` still parses). A component that retires a version
+/// specializes this, mirroring the `SaveSerialize<C>` pattern:
+///
+/// ```cpp
+/// // C_Foo evolved v1 -> v2 by appending a defaulted `damping_` field.
+/// namespace IRWorld {
+/// template <> struct SaveMigration<C_Foo> {
+///     static std::vector<std::pair<std::uint32_t, ColumnMigratorFn<C_Foo>>>
+///     migrators() {
+///         return {
+///             {1u, [](IRAsset::BinaryReader &r) -> IRAsset::Result<C_Foo> {
+///                  // read exactly the v1 layout; default the appended field
+///                  IRAsset::Result<float> vel = r.readF32();
+///                  if (!vel.ok())
+///                      return IRAsset::Result<C_Foo>::error(vel.status_.code_,
+///                                                           vel.status_.message_);
+///                  return IRAsset::Result<C_Foo>::success(C_Foo{vel.value_, 0.0f});
+///              }},
+///         };
+///     }
+/// };
+/// } // namespace IRWorld
+/// ```
+///
+/// Contract:
+///   - **Direct per-version readers, never chained.** Each reader decodes its
+///     own era's bytes straight to a current-build `C`; a v3 bump leaves the
+///     v1/v2 readers untouched. No `v1 -> v2 -> v3` composition.
+///   - **The current version is NOT listed here** — `SaveSerialize<C>::read`
+///     handles `kSaveVersion`; list only the retired versions
+///     (`[1 .. kSaveVersion - 1]`).
+///   - A disk version below `kSaveVersion` with no entry here is a hard
+///     `BinaryIOError::MigratorMissing` on load (never a silent read-at-current);
+///     a disk version above `kSaveVersion` is `VersionTooNew`.
+
+#include <irreden/asset/binary_io.hpp>
+
+#include <cstdint>
+#include <functional>
+#include <utility>
+#include <vector>
+
+namespace IRWorld {
+
+/// Reader for one retired on-disk version of component `C`: consumes exactly
+/// that version's bytes from @p r and yields a fully-populated current-build
+/// `C` (fields added since are defaulted, renamed fields remapped). Type-bound
+/// to `C`; the registry erases it into the generic column-read hooks.
+template <typename C>
+using ColumnMigratorFn = std::function<IRAsset::Result<C>(IRAsset::BinaryReader &)>;
+
+/// Customization point declaring how to read component `C` at each retired
+/// on-disk version. Specialize for a component that has bumped
+/// `SaveTrait<C>::kSaveVersion`; the primary template declares none.
+template <typename C> struct SaveMigration {
+    /// `(fromVersion, reader)` pairs — one per retired version. Default: none.
+    static std::vector<std::pair<std::uint32_t, ColumnMigratorFn<C>>> migrators() {
+        return {};
+    }
+};
+
+} // namespace IRWorld
+
+#endif /* SAVE_MIGRATION_H */
--- a/engine/world/include/irreden/world/save_registry.hpp
+++ b/engine/world/include/irreden/world/save_registry.hpp
@@ -30,6 +30,7 @@
 /// needs a `SaveSerialize<C>` specialization for its non-POD fields — is
 /// downstream work, not this slice.
 
+#include <irreden/world/save_migration.hpp>
 #include <irreden/world/save_serialize.hpp>
 #include <irreden/world/save_trait.hpp>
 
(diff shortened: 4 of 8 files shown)


jakildev (commented):
## Review — persist: component migration registry (W-4) (P5) (#2216)

Stacked on #2238 (`fleet:approved`, still open — reviewing the child diff only; upstream is approved separately). If `SaveComponentEntry`'s shape upstream of P4 changes before merge, this downstream will need a rebase, but nothing in #2238 touches `save_registry.hpp`, so that's unlikely.

**Verdict:** approve

Unlike its stacked sibling #2238, this PR touches real load-path production code (`save_registry.hpp`, `world_snapshot.cpp`), so I built and ran it rather than reviewing statically only:

- `fleet-build --target IrredenEngineTest` — clean build (no warnings introduced).
- `--gtest_filter='MigrationTest.*'` — all 5 new tests pass.
- Full suite: 1295 tests, 1294 passed, 1 skipped (`GpuComputeDispatchTest.SkippedOnNonOpenGLBackend`, expected on the Metal backend) — matches the author's stated 56/56 persist tests plus the 5 new migration tests (1290 + 5 = 1295).

### Checklist

- **Zero-mutation contract (Rule #5)** — verified the version dispatch (`readerForVersion`) runs in Phase 2b (the `decodeRow_` dry-run pass) *before* the watermark advance and Phase 3's live writes, in all three call sites (ARCH columns, SNGL columns, and the singleton get-or-create). A `MigratorMissing`/`VersionTooNew` failure aborts with a pristine world — confirmed by `MigrationTest.NewerThanBuildRejectedVersionTooNew` and `MigrationTest.MissingMigratorHardFails`, both of which assert `entitiesRestored_ == 0` and `getLiveEntityCount()` unchanged.
- **Phase 3 re-resolution is redundant but safe** — `readerForVersion` is called again in Phase 3 after already succeeding in Phase 2b (`world_snapshot.cpp:397/440`); the comment correctly notes a null result there "would be a phase-2b bug." Since load is single-threaded and sequential, there's no TOCTOU window between the two calls — just belt-and-suspenders.
- **On-disk key stability** — migrators are keyed on the session-local `ComponentId` at the registry level, but the disk column key stays `SaveTrait<C>::kSaveName` (never `typeid(C).name()`), matching Save Format Rule #2. Confirmed `readerForVersion` doesn't leak the mangled name into anything disk-facing.
- **Byte-format cross-check** — the new test hand-builds `IRWS` files (`writeIrws`) rather than going through `saveWorld` (which can only emit the current version). I traced this against the real writer in `world_snapshot.cpp` (`archW.writeVarUInt(groups.size())` → indices → entities → per-column `writeU32(version)` + `writeVarUInt(len)` + bytes) and the hand-rolled version matches exactly, field-for-field.
- **`IR_SAVE_OPT_IN`/naming** — `C_Mig`/`C_NoMigrator`/`C_Unknown` correctly `C_`-prefixed; `SaveMigration<C>` mirrors the existing `SaveSerialize<C>` customization-point pattern; `MigTest` namespace and component names don't collide with `WsSnap` (world_snapshot_test.cpp) or the P4 sibling's own test namespace.
- **`ColumnReadHooks` refactor** — folding `appendRow_`/`decodeRow_`/`readIntoEntity_` into one `buildReader<C>` factory (shared by the current-version reader and every migrator) is a clean de-dup of the prior per-hook lambdas; verified the extracted hooks are behaviorally identical to the pre-PR versions (same read → check → push/set pattern).
- **Enum addition** — `BinaryIOError::MigratorMissing` is additive; grepped for any exhaustive `switch` over `BinaryIOError` that might need updating — none exists (both `switch` sites in `key_value_store.cpp`/`voxel_set_format.cpp` switch on unrelated tag/mode enums), so this is a safe addition.
- **`engine/world/CLAUDE.md`** — the new section accurately describes the shipped dispatch (four cases, direct-never-chained migrators, current-version-not-listed convention); cross-checked line-by-line against `save_registry.hpp`.

### Nits

- `SaveRegistry::registerComponent`'s migrator-loading loop (`save_registry.hpp`, the `for (auto &versioned : SaveMigration<C>::migrators())` block) uses `migratorReaders_.emplace(...)`, which silently keeps the first entry and drops any later one if a component's `migrators()` ever lists the same `fromVersion` twice. Low-severity (a same-author authoring mistake in a rarely-touched customization point), but a `IR_ASSERT(inserted, ...)` on the `emplace` result would turn a silent-wrong-behavior footgun into a fast, obvious failure at registration time — worth a follow-up if convenient, not a blocker.

### Praise

- The worked example in `save_migration.hpp`'s header comment and the test's `C_Mig` (has a migrator) vs. `C_NoMigrator` (deliberately doesn't) split make the one hard-fail case (`MigratorMissing`) exercised as a first-class scenario rather than an afterthought — `MissingMigratorHardFails` is exactly the kind of "prove the dangerous path is dangerous" test that's easy to skip.
- Choosing a test-local `C_Mig` component for the worked migration rather than bolting a dead field onto a real gameplay component (the plan's own documented fallback) keeps this diff fully self-contained and avoids permanently scarring a production component for a test fixture.

- [x] macOS `IrredenEngineTest` — confirmed locally: full suite green (1294/1295, 1 expected GL-only skip); `MigrationTest.*` (5/5) verified in isolation and as part of the full run.
- [ ] Linux `linux-debug` build + gtest — still outstanding; tagging `fleet:needs-windows-smoke` below since this PR (like its stacked sibling) touches `test/CMakeLists.txt` and load-path code that should be cross-platform-verified.

Opus recheck required: touches `engine/world/` (persistence load-path dispatch + the zero-mutation Rule #5 contract) — I'm confident in the logic after tracing all three call sites and the byte-format cross-check, but this is exactly the class of subtle invariant (load-time error-path correctness, mutation-free-until-phase-3 ordering) the fleet asks Opus to re-confirm.

jakildev (commented):
## Opus recheck — persist: component migration registry (W-4) (P5) (#2216)

**Verdict:** approve

Stacked on #2238 (`fleet:approved`, still open) — reviewing this child diff only. Confirmed nothing in #2238 touches `save_registry.hpp` / `world_snapshot.cpp` / `save_migration.hpp`, so this needs no rebase when P4 merges.

Sonnet's first pass was thorough and correct; it explicitly escalated the load-path error ordering and the zero-mutation Rule #5 contract for Opus re-confirmation. I traced exactly those, three call sites deep, and confirm them.

### What I re-confirmed (the escalated items)

- **Zero-mutation contract holds across all four dispatch sites.** `readerForVersion` runs in the phase-2b dry-run for both ARCH (`world_snapshot.cpp:397`) and SNGL (`:417`) *before* any live write, and a null return goes straight to `loadFail` — so a `VersionTooNew` / `MigratorMissing` column aborts with a pristine world. `NewerThanBuildRejectedVersionTooNew` and `MissingMigratorHardFails` both assert `entitiesRestored_ == 0` **and** `getLiveEntityCount()` unchanged, which is the right pair (out-component never materialized *and* no side effects). Sonnet flagged this class of bug; on closer read I confirm it's airtight.

- **Phase-3 singleton ordering is correct.** The re-resolve at `:465` happens *before* `getOrCreateSingletonEntity_()` at `:469`, so even the (phase-2b-should-have-caught-it) null path can't strand a freshly-minted singleton entity on the error return. The comment says exactly this and the code matches. Phase-3 re-resolution being redundant-but-safe (single-threaded, no TOCTOU) is right.

- **Entry-null guard precedes every `readerForVersion`.** All four sites are gated by the `entry == nullptr` skip (`columnsSkipped_` / `singletonsSkipped_`), so the unknown-component forward-compat skip (Rule #1) never reaches the version dispatch — the four-case comment's "handled by the loader before this is reached" is accurate. `MigrationComposesWithUnknownSkip` proves migrate + skip coexist within one archetype.

- **`buildReader<C>` refactor is behaviorally identical and lifetime-clean.** Folding the old `appendRow_`/`decodeRow_`/`readIntoEntity_` lambdas into one factory keyed by a captured `readFn` is a clean de-dup; the three hooks each capture their own copy of the `std::function`, all stored in `ColumnReadHooks` living in `m_entries` (stable for the load's lifetime) — no dangling, and the returned `const ColumnReadHooks*` points into that same stable storage. Critically, phase-2b `decodeRow_` and phase-3 `appendRow_`/`readIntoEntity_` share the *same* `readFn`, so byte-consumption is identical between the validate and apply passes — which is what makes the dry-run a valid gate for the migrator path, not just the current path.

### Non-blocking nits (agree with Sonnet, do not re-review over these)

- Sonnet's `emplace` nit stands: `SaveMigration<C>::migrators()` listing the same `fromVersion` twice silently keeps the first and drops the rest. An `IR_ASSERT(inserted, ...)` on the emplace result would turn a silent-wrong into a fast registration-time failure. Follow-up, not a blocker — a same-author authoring mistake in a rarely-touched customization point.
- Minor diagnostic imperfection in `missingMsg` (`save_registry.hpp:262`): `lowest` is seeded to `diskVersion`, so when every registered migrator is *higher* than the missing version, the message prints "registered migrators start at v<diskVersion>" — i.e. it names the missing version as the start. Diagnostic-string only, no behavioral impact; fold into the same follow-up if convenient.

Neither rises to needs-fix. Load-path invariants, error ordering, and the zero-mutation contract are all sound.

jakildev:
Addressed the nit: `SaveRegistry::registerComponent`'s migrator-loading loop now asserts (`IR_ASSERT`) that each `fromVersion` inserts uniquely into `migratorReaders_`, instead of silently keeping only the first entry on a duplicate. Full suite green (1294/1295, 1 expected GL-only skip on macOS Metal); `MigrationTest.*` (5/5) re-verified. Build clean. Reviewer please re-evaluate. — worker

jakildev (commented):
## Review — persist: component migration registry (W-4) (P5) (#2216)

Stacked on #2238 (`fleet:approved`, still open — reviewing the child diff only; upstream is approved separately). Confirmed nothing in #2238 touches `save_registry.hpp` / `world_snapshot.cpp` / `save_migration.hpp`, so no rebase risk here.

**Verdict:** approve

### Prior-review resolution
- ✅ `engine/world/include/irreden/world/save_registry.hpp` (in `SaveRegistry::registerComponent`'s migrator-loading loop) — `migratorReaders_.emplace(...)` silently kept only the first entry and dropped any later one if a component's `migrators()` ever listed the same `fromVersion` twice — verified fixed at `140b9154`: the `.emplace(...).second` result is now captured as `inserted` and asserted via `IR_ASSERT(inserted, "SaveMigration<{}>::migrators() lists fromVersion {} more than once", name, versioned.first)`.

### Verification of the fix commit (`140b9154`, the only commit since the prior review)
- Confirmed `IR_ASSERT`'s real definition (`engine/profile/include/irreden/ir_profile.hpp:88`, `#define IR_ASSERT(x, en, ...)` → `engAssert(cond, file, func, line, #cond, format, args...)`) is a variadic fmt-style assert — not the unrelated 2-arg `assert()`-wrapping macro locally `#define`d in `engine/ir_args.cpp` (different TU, doesn't apply here). The call's format-string/arg pairing (`"...{}...{}...", name, versioned.first`) matches that signature correctly.
- `name` (the component's `const char*` save-name) is in scope from `save_registry.hpp:158`, and the new `#include <irreden/ir_profile.hpp>` follows the same include convention already used by sibling `engine/world/src/*.cpp` files (`world.cpp`, `chunk_residency.cpp`) and other header-only code (`metal_render_impl.hpp`, `rendering_rm.hpp`) — no layering violation (world depending on the lower-level profile module is the allowed direction).
- `clang-format --dry-run -Werror` on the changed file is clean.
- Scope is exactly the nit: a defensive registration-time assert. It doesn't touch the load-path dispatch, phase-2b/phase-3 ordering, or any of the zero-mutation-contract invariants the prior Opus recheck already confirmed on `a08f2054` — no need to re-escalate this follow-up commit.

- [x] No new test needed — this is a programmer-error guard at registration time (a component author listing a duplicate `fromVersion`), consistent with the original nit's "not a blocker" framing. Existing `MigrationTest.*` (5/5) is unaffected by this change.

jakildev:
Merger: waiting on base PR #2238 to merge before this stacked PR can be re-targeted to master and merged.

— fleet merger

jakildev:
#2238. Its head ref `claude/2215-persist-round-trip-tests` was force-
pushed and the new tip conflicts with this PR's own commits.

Conflicted files:
- `engine/world/CLAUDE.md`
- `engine/world/include/irreden/world/save_trait.hpp`
- `test/CMakeLists.txt`

  git fetch origin
  git rebase origin/claude/2215-persist-round-trip-tests
  # resolve conflicts, then:
  git push --force-with-lease

after a manual rebase.

— fleet merger

jakildev:
Merger: base PR #2238 merged; re-targeted this PR to `master` but the rebase onto current master conflicts. The child commits cannot be replayed cleanly — worker will resolve semantically.

Conflicted files:
- `engine/world/CLAUDE.md` — master: `4e4e41ee persist: RELN relation chunk — CHILD_OF triples + name table (W-5) (#2214) (#2230)`; PR: `a08f2054 persist: component migration registry (W-4) (#2216)`
- `engine/world/include/irreden/world/save_trait.hpp` — master: `738d0c2c persist: IRWS world snapshot container + walker + loader (#2213) (#2228)`; PR: `f7d5fe48 persist: IRWS world snapshot container + walker + loader (#2213)`
- `test/CMakeLists.txt` — master: `0b43e330 persist: round-trip + determinism test suite (W-7/W-8) (P4) (#2215) (#2238)`; PR: `a08f2054 persist: component migration registry (W-4) (#2216)`

Labeled `fleet:semantic-conflict` — worker will resolve next.

— fleet merger

jakildev:
Resolved semantic conflict: the re-target-to-master rebase was replaying the whole inherited persist prefix (W-1/W-2 #2212, W-3/W-6 #2213, W-5 #2214, W-7/W-8 #2215) which has all since merged, so every reported conflict (`save_trait.hpp` add/add, `engine/world/CLAUDE.md`, `test/CMakeLists.txt`) was a stale inherited copy vs master's now-merged copy. Dropped the inherited prefix with `git rebase --onto origin/master 53630ae7` (the #2215 tip this PR was stacked on) — replays only the 3 genuine W-4 commits. `save_trait.hpp` no longer touched (was purely inherited); `CLAUDE.md` + `test/CMakeLists.txt` auto-merged cleanly (W-4 section/test appended after master's merged persist content). Diff is now exactly the 8 W-4 files (644+/40-). Build clean (IrredenEngineTest); MigrationTest 5/5 + suite 9/9 PASS. Reviewer please re-evaluate the rebased diff. — worker

jakildev (commented):
## Review — persist: component migration registry (W-4) (P5) (#2216)

**Verdict:** approve

### Prior-review resolution

This PR was already fully reviewed and approved twice (Sonnet + Opus
recheck, both `approve`, "Escalation: none. Safe for merge." on the
final pass at `140b9154`). Since then the merger re-targeted the branch
to `master` and resolved a reported "semantic conflict" by dropping the
inherited stacked-PR prefix (`git rebase --onto origin/master
53630ae7`) — per the merger's own comment, the conflicts were stale
copies of already-merged upstream content, not real collisions with
this PR's own changes.

I verified that claim rather than trusting it:

- ✅ All prior review items (the `emplace`→`IR_ASSERT` duplicate-migrator
  nit, the load-path ordering / zero-mutation-contract items Opus
  traced) — verified still fixed; nothing about this PR's own content
  changed in the rebase.
- Byte-compared every touched file between the pre-rebase reviewed
  commit (`a08f2054`) and the post-rebase commit (`0b11b0bc`):
  `binary_io.hpp`, `engine/world/CLAUDE.md`, `save_migration.hpp`,
  `save_registry.hpp`, `world_snapshot.hpp`, `world_snapshot.cpp`,
  `component_migration_test.cpp` are all **identical**. Same check on
  the final nit-fix commit (`140b9154` → `e62a4285`) for
  `save_registry.hpp` — identical.
- `test/CMakeLists.txt` differs between those two commits only because
  master gained other test entries in the interim (`ecs/*_test.cpp`
  from unrelated merged PRs); the PR's own diff against its *new* base
  (`origin/master...HEAD`) is exactly the single expected line
  (`+ world/component_migration_test.cpp`) — no unrelated churn crept
  in during the rebase.
- Overall diffstat vs. new base: 8 files, 644+/40-, matching the
  merger's reported post-resolution diff exactly.

So this is a mechanical rebase with zero semantic delta — re-approving
on that basis rather than re-running the full checklist from scratch.

### Verification performed (this pass)

- `fleet-build --target IrredenEngineTest` on the rebased HEAD
  (`e62a4285`) — clean build.
- `--gtest_filter='MigrationTest.*'` — 5/5 pass.
- Full suite: 1302 tests, 1301 passed, 1 expected skip
  (`GpuComputeDispatchTest.SkippedOnNonOpenGLBackend`, GL-only on this
  Metal host) — same shape as the original review's 1295/1294+1, with
  the higher counts reflecting other tests that landed on master since.

- [x] macOS `IrredenEngineTest` on the rebased branch — confirmed
      clean, 1301/1302 (1 expected GL-only skip).
- [ ] Linux `linux-debug` build + gtest — still outstanding, consistent
      with the original review's note; `fleet:needs-windows-smoke`
      already present on the PR.

Opus recheck not required — this pass only re-verifies the rebase was
mechanical; the substantive load-path/zero-mutation-contract logic was
already Opus-confirmed on the pre-rebase, byte-identical content.

Issue #2217, opened 2026-07-04
persist: GPU-resident state regeneration on load — canvas-attach pass (W-10) (P6)


jakildev:
**Model:** opus
**Part of epic:** #667
**Plan file:** `.fleet/plans/issue-667.md` (umbrella context)
**Blocked by:** #2216

## Scope

The post-load one-shot seed pass (the long-anticipated staged->pool canvas-attach pass): loaded C_VoxelSetNew records exit staged mode via pool allocate + seed + resync + queued GPU re-upload; the live render context (C_Persistent canvas bundle + C_VoxelPool) is preserved and reused, mirroring resetGameplay; all downstream lighting/AO/shadow/fog textures re-derive per-tick for free. Defines the SaveTrait<C_VoxelSetNew> staged-mode round-trip contract with P1.

## Acceptance criteria

New persist_roundtrip demo: save -> reset+load -> screenshots pixel-equivalent to pre-save on BOTH backends; loaded scene renders lit; pool waterline idempotent across cycles; every loaded set exits staged mode with matching recordCount.

## Dependencies

Blocks P7 (Lua binds the finished surface).

## References

The canonical plan is this issue's `## Plan` comment (posted at filing, architect-planned with the human in the loop — 2026-07-03 triage session). Umbrella: #667.

jakildev:
## Plan: persist P6 — GPU-resident state regeneration on load (W-10)

- **Issue:** #2217
- **Model:** opus
- **Date:** 2026-07-03
- **Epic:** #667 — see `.fleet/plans/issue-667.md` for full context
- **Blocked by:** #2216

### Scope

After `IRPersist::loadWorld` rebuilds the ECS graph, every component whose GPU-resident state was opted OUT of serialization (P1's default for handle fields) holds no valid GPU state — textures/SSBOs/framebuffers/pool residency were process-local and were never written to the snapshot. This child adds the **post-load regeneration pass** that makes a loaded world actually render, and defines its input contract against the serializer.

The audit below establishes there are only **two** distinct classes of GPU-resident state, and one regenerates for free:

- **Render-context infrastructure** (canvas/framebuffer/camera bundle + `C_VoxelPool`): NOT serialized, NOT reconstructed on load — the live render context is preserved and reused, exactly the way `resetGameplay` preserves it.
[...]

PR #2245, opened 2026-07-05
demo: random_voxels — interactive voxel-recoloring demo

jakildev:
## Summary
- New demo `creations/demos/random_voxels`: an isometric field of voxel cubes with a family of one-shot recolor commands bound to keys.
- **R** fires the merged #2210 `Command<RANDOMIZE_VOXELS>`; **H / I / M** add channel-cycle / invert / grayscale recolors, composed in the demo from that same public `IRSystem::executeQuery<C_VoxelSetNew, Exclude<C_Locked>>` primitive — no engine changes needed to grow the set.
- Top-left help text is the engine's built-in command-list overlay (`buildCommandListText` via `TEXT_TO_TRIXEL`), shown from frame one; every registered command auto-appears with its hotkey (**G** toggles it).
- Four corner pillars are tagged `C_Locked`, so every recolor skips them — a live demonstration of the query's `Exclude<C_Locked>` filter.

## Screenshots
Initial gradient + command overlay:

initial gradient

After pressing **I** (INVERT) — the field inverts while the locked pillars stay white (they'd go near-black if the exclude filter weren't working):

after invert

## Test plan
- [x] `fleet-build --target IRRandomVoxels` green on macOS/Metal, rebased onto current master.
- [x] `--auto-screenshot` renders the scene + overlay (non-blank).
- [x] On-screen recolor verified: temporarily firing INVERT inverts the field while `C_Locked` pillars stay white — exercising the query traversal, the `Exclude<C_Locked>` filter, and the `editVoxels` GPU resync end to end.

## Notes for reviewer
- Built and smoke-run on **macOS/Metal only**; the OpenGL backend was not built on this host.
- H / I / M are registered as named lambda commands via `getCommandManager().createCommand(..., name)` so they show up labelled in the overlay; **R** uses the engine prefab `Command<RANDOMIZE_VOXELS>`.


jakildev (commented):
## Opus recheck (re-review) — persist: GPU-resident state regen on load, canvas-attach seed (#2217)

Stacked on #2239 (`fleet:approved`, still open) — reviewing the child diff only; nothing in #2239 touches the voxel-set files here, so no rebase risk.

I approved this PR earlier (01:43Z). A new commit landed **after** that approve — `25c43a6ba` "persist: keep an exhausted seed retryable (attachToCanvas nit)" — addressing the minor retryability nit I raised. Re-reviewing the delta on the changed head, I have to **withdraw the approve**: that commit introduces a correctness regression on the exact path it was meant to fix.

**Verdict:** needs-fix

### Needs-fix

- `engine/prefabs/irreden/voxel/components/component_voxel_set.hpp:587` — the nit fix seeds in place and only clears `pendingVoxels_` on `numVoxels_ > 0`, so an exhausted set stays staged and is retried. But it preserves the *payload* (`pendingVoxels_`) without preserving the *extent* (`size_`). On the exhaustion-mismatch branch, `seedIntoPool` runs `size_ = ivec3(0)` (line 645). `SEED_STAGED_VOXELS` (`system_seed_staged_voxels.hpp:37`) calls `attachToCanvas()` every frame over every `C_VoxelSetNew`, so on a later frame with pool space the retry re-enters `seedIntoPool`, reads `const ivec3 extent = size_` (line 612) as `(0,0,0)`, and:
  - `requestedVoxels = pendingVoxels_.size()` is the full authored count, so `allocate` reserves the right number of slots and `numVoxels_ == requestedVoxels` → **no** mismatch this time;
  - but the seed triple-loop is `for (x < extent.x=0)` → it writes **zero** positions/colors, leaving the allocated pool span uninitialized;
  - back in `attachToCanvas`, `numVoxels_ > 0` → `pendingVoxels_.clear()` runs and the range is queued for GPU upload.

  Net: the "retryable" path the commit adds doesn't recover the set — it drops the authored data and leaves the set pool-resident over garbage slots (uninitialized positions/colors, stale active-mask). That's a silent corruption, strictly worse than the pre-fix behavior (clean non-recoverable no-render), which my earlier pass explicitly called acceptable to ship.

  **Two ways to resolve, author's call:**
  1. **Make the retry actually correct** — preserve `size_` across a failed seed so the re-entry iterates the right extent. Either don't zero `size_` on the mismatch branch (leave the extent; `pendingVoxels_` still describes a box of that size), or snapshot/restore `size_` in `attachToCanvas` around the failed call. If you keep `size_` non-zero on mismatch, double-check nothing treats `size_ != 0 && numVoxels_ == 0 && pendingVoxels_.empty()` (the dense-ctor exhaustion state) as inconsistent — the dense-ctor path leaves `pendingVoxels_` empty so `attachToCanvas` no-ops there, but confirm no other reader assumes `size_ == 0` on an exhausted set.
  2. **Revert the nit** — go back to the move-out-then-clear form. The non-recoverable loss on genuine pool exhaustion is a rare, already-`IRE_LOG_ERROR`-logged edge, and I flagged it as acceptable to ship in the first place. A clean no-render beats a corrupt render.

### Confirmed (unchanged from my prior recheck)

Everything I confirmed at 01:43Z still holds on this head — the pool-free default ctor completeness, the staged-mode gating, the `seedIntoPool` sharing between the dense ctor and post-load attach, and the mid-rotation authored-vs-derived serialize fix (`98b9a432`, `rotationSourceVoxels_` size-invariant). The only regression is the `size_`-vs-`pendingVoxels_` asymmetry above.

### Test plan

- [x] `IrredenEngineTest --gtest_filter='VoxelSetSerialize.*'` — 6/6 pass (Sonnet re-ran previously; unaffected by this commit, which touches the live-pool attach path, not the serializer).
- [ ] Add / run a pool-exhaustion-then-space-frees case for `attachToCanvas` if you take fix option 1 — the unit tests don't cover the retry path, which is exactly where the defect lives.
- [ ] Linux/GL smoke — still outstanding per the author's own test plan; `fleet:needs-windows-smoke` stays on this PR.

🤖 Opus recheck by Claude (review-pr skill) — withdrawing my prior approve; the post-approve nit commit regressed the exhaustion-retry path.

PR #2246, opened 2026-07-05
render: fix fog cross-section black bands — rim fade + cut march + demo

jakildev:
## Summary
- Root cause of the "two black bands around the fog cross-section": the analytic cut (#2124) decides solid-vs-air from the light-occlusion **voxel** bitfield, which SDF geometry never enters. fog_demo's edge scenes ran the cut arc over the unconditional SDF box floor, and the voxel ground slab (`centerAroundOrigin=true` + corner-style `(-30,-30)` transform) covered only one quadrant of the vision disc — the cut painted along that quadrant's arc and fell through to the black fog mask on the two flanking segments.
- **Rim fade (new):** the unexplored-black fallback is replaced by a squared-ease lift of hidden rasterized matter toward its lit colour, keyed to the column's distance past the hard disc and reaching black exactly at the stage-1 keep-ring width (`kFogRimFadeCells == kFogHiddenKeepCells`). Whatever the occupancy test can't classify — sub-cell-thin floors, hollow interiors, air pockets beside props — lands in a smooth "edge of the light" gradient instead of a black band, so the boundary is artifact-free for any content the raster keeps. The solid-backed cut band stays brighter than the fade (`kFogRimFadeLevel < kFogCutTone`), preserving the cross-section face.
- **Cut march:** the two fixed occupancy samples behind the disc entry become a bounded half-cell march (`kFogCutMarchSteps = 8`), a strict superset of the old sampling — air-pocket entries at an object's cut corner now cap with the toned cut.
- `fog_demo`: the three cross-section scenes (`--edge-zoom`, `--edge-smooth`, `--detached-edge`) skip the SDF floor and share an origin-centered voxel ground slab (`createEdgeGroundSlab()`; new to `--detached-edge`).
- Re-blessed the macOS edge/detached render-verify references for the new boundary look.

## Test plan
- [x] `render-verify --target IRFogDemo`: **11/11 PASS**, byte-deterministic re-run
- [x] Default grid-reveal shots (`fog_zoom2/4/8_origin`) **and** the soft-disc `fog_edge_smooth*` shots byte-identical to their pre-change references — the fade is provably gated to hard vision circles
- [x] `--edge-zoom` zoom 5/9/14 visually verified: continuous cut band + smooth fade ring, no black bands, no corner notch
- [x] Thin-floor stress (1-voxel floor + small crate straddling the rim, mirroring a real consumer scene): clean disc, smooth crate fade, no gash — verified locally, not committed
- [x] Metal (macOS) built + run; GLSL twin changed identically — needs Linux smoke

## Notes for reviewer
- The GLSL and Metal fog shaders are kept byte-mirrored; review either and diff the other.
- Pre-existing, unchanged by this PR: in `--detached-edge` the detached solid renders whole past the disc (no world-fog clip) — regression tracked in #2248 (bisect window + oracle in its plan). The SDF blocker-bitfield blind spot is tracked in #2247.
- `kFogRimFadeCells` must stay equal to stage-1's `kFogHiddenKeepCells` — the shader comment documents the coupling.

## Screenshots

<details>
<summary>fog_edge_zoom5</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>fog_edge_zoom9</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>fog_edge_zoom9 — ROI crop (cutface_slab)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>fog_edge_zoom14</summary>

| Before | After |
|--------|-------|
|  |  |

</details>


jakildev (commented):
## Review — render: fix fog cross-section black bands — cut march + fog_demo scene (#2246)

**Verdict:** approve

### Summary of what this PR does

Root-causes and fixes the "two black bands" artifact in `fog_demo`'s cross-section scenes: the analytic fog cut (#2124) reads solid-vs-air from the light-occlusion **voxel** bitfield, which the demo's SDF box floor never populated, and the old off-origin `(-30,-30)` voxel ground slab covered only one quadrant of the vision disc. Fixes both the scene (origin-centered slab shared via a new `createEdgeGroundSlab()` helper, SDF floor skipped for the three cross-section scenes) and the shader (replaces two fixed occupancy samples with a bounded 8-step half-cell march so a small air pocket at a cut corner still caps instead of leaving a black notch).

### Verification performed

- Confirmed `createEdgeGroundSlab()`'s claimed extent: `ivec3{60,60,3}` with `centerAroundOrigin=true` and world position `(0,0,5)` puts local voxel indices at x/y ∈ [-29.5, 29.5], z ∈ [4,6] — matches the doc comment `[-29,30]²×[4,6]`, and both `kEdgeVisionRadius`/`kDetachedVisionRadius` (9.0, confirmed at `creations/demos/fog_demo/main.cpp:225,263`) sit well inside that extent, so "fully interior" checks out.
- Confirmed `g_edgeZoom`/`g_edgeSmooth`/`g_detachedEdge` are pre-existing flags (not newly introduced), so the new SDF-floor skip condition doesn't reference anything undefined.
- Confirmed the ordering comment on the `--detached-edge` branch ("created before the detached canvas so it allocates from the MAIN canvas pool") is actually correct in the code — `createEdgeGroundSlab()` is called before `IRPrefab::EntityCanvas::createWithVoxelPool(...)`, so `C_VoxelSetNew`'s default `targetCanvas` resolves to the still-active main canvas at that point.
- Read the full GLSL shader region in context (not just the hunk): the march loop's early-exit condition (`&& !cutSolid` in the `for`) preserves the original short-circuit-OR behavior (no wasted buffer reads once solid is found), and the loop's `s = 1..8` range is a strict superset of the old two sample points (`s=1`, `s=3`), so every previously-painted pixel is still covered — matches the PR's byte-parity claim for existing shots (render-verify: default shots byte-identical).
- GLSL and Metal blocks are semantically byte-mirrored (Metal's `fogCutVoxelSolid` just carries an extra explicit `occlusionGrid` device-pointer parameter, consistent with the pre-existing calling-convention difference between the two backends in this same file).
- `fogCutVoxelSolid`'s signature/body is unchanged by this diff — only the call site (single call → bounded loop) changed.
- This cut pass runs after lighting and writes color only (per the shader's own comment, unchanged by this PR) — it doesn't touch `trixelDistances`, so none of the AO/shadow/light-volume grid-coverage or guard-band invariants apply here.
- Extracted and viewed the `fog_edge_zoom9` before/after screenshots directly: "before" shows a visible diagonal split across the ground disc (the lit quadrant vs. the two black-cut segments); "after" shows a uniform slate-gray disc with no split — a direct visual confirmation the fix works, not just a compile-and-hope.
- The nine modified `test/references/macos-debug/*.png` files are exactly the scenes this PR intentionally changes (edge-zoom/edge-smooth/detached-edge at their zoom levels) — no unrelated reference churn.

### Nits

- No committed perf-gate baseline exists yet (per the bot comment on this PR), so the up-to-8x increase in worst-case occlusion-buffer reads for cut-band pixels isn't measured against a baseline. The march is tightly scoped (hard-disc cut pixels only, early-exits on first solid hit) so this is very unlikely to matter, but worth keeping in mind once a baseline lands.

- [x] (author-reported) `render-verify --target IRFogDemo`: 11/11 PASS, byte-deterministic re-run.
- [x] (author-reported) Default grid-reveal shots byte-identical to previously committed references.
- [x] (reviewer-verified via screenshots) `--edge-zoom` cross-section band is now continuous with no black bands or corner notch.
- [ ] Linux/GL smoke — author's own test plan flags this as outstanding (macOS/Metal only so far); shader change is byte-mirrored but genuinely touches both backends' shader source, so this is a real (not just procedural) smoke requirement.

Opus recheck required: touches `engine/render/src/shaders/` (both GLSL and Metal fog-to-trixel compute shaders) — I traced the march logic, the byte-mirror parity, and the lighting-invariant non-interaction, and found no issue, but per the fleet's review protocol shader/render-pipeline changes get an Opus pass regardless of confidence.

jakildev:
has-nits assessed: the sole nit is non-actionable — it observes that the up-to-8× worst-case occlusion-buffer read increase for cut-band pixels isn't measured because \*\*no perf-gate baseline exists yet\*\* (the bot comment confirms one auto-commits on the next perf-relevant master merge). The march is bounded to 8 half-cell steps and early-exits on first solid hit, and the reviewer flagged it as "very unlikely to matter." There's no code change that addresses "no baseline exists," so I'm not force-pushing a no-op amendment (which would reset the pending opus recheck and re-run CI). Clearing `fleet:has-nits` to stop worker-dispatch churn; leaving `fleet:needs-opus-recheck` (shader change → opus pass) and `fleet:needs-windows-smoke` (Linux/GL smoke outstanding) intact. — worker

jakildev (commented):
## Opus recheck — render: fix fog cross-section black bands (#2246)

**Verdict:** approve

Sonnet approved and escalated per protocol because this touches `engine/render/src/shaders/` (GLSL + Metal fog-to-trixel compute). Sonnet flagged that the new march is a strict superset of the old two samples with an early-exit, and asserted byte-parity for existing shots; on closer read I confirm all of that, and additionally traced the one invariant Sonnet's confidence rested on but did not spell out.

### Opus-only items verified (beyond Sonnet's pass)

- **OOB false-cap safety of the extended march (the load-bearing new-behavior claim).** The change grows the cut-solid probe from 2 fixed samples (`cellA` at `bestT+0.5cell`, `cellB` at `bestT+1.5cell`) to an 8-step half-cell march reaching `bestT + 4` world cells. The real question is whether the deeper reach can hit something the old code never did and *false-cap* a pixel that should stay black. I read `fogCutVoxelSolid` in full in **both** backends: `c_fog_to_trixel.glsl:87` and `metal/c_fog_to_trixel.metal:57` both return `false` (air) for any cell outside the ±128 grid half-extent. So a march step that walks off the occlusion grid reads as air, never as solid — the extra depth cannot manufacture a cap. Confirmed, not assumed.

- **Monotonic-descent semantic.** The comment's claim that "the iso ray only ever descends, so the march can never climb back into solid that isn't really behind the cut" is consistent with the cut's meaning: capping when solid matter exists *behind* the cut pixel along the descending view ray is the intended result, not a false positive. Hitting solid deeper = correct cap.

- **Byte-parity mechanism.** The loop's `s=1` and `s=3` are exactly the old `cellA`/`cellB` positions, and `!cutSolid` early-exit means every pixel the old code capped is capped identically. The branch only executes when `fogCutSolidsAvailable` is set on an active cross-section cut, so the default grid-reveal shots never enter it — matching the byte-identical-default claim and the 9 changed references being exactly the edge/smooth/detached-edge scenes.

- **No invariant interaction.** `fogCutVoxelSolid`'s signature/body, the binding-28 alias-slot re-bind, the FogObserverData/occlusion struct layouts, and the color-only post-lighting write are all untouched by this diff (call-site only). No CPU/GPU struct drift, no `trixelDistances`/AO/depth coupling introduced.

- **GLSL↔Metal parity.** Both shader hunks are byte-mirrored; Metal's extra explicit `occlusionGrid` device-pointer arg is the pre-existing calling-convention difference in this file, not new.

- **Demo `main.cpp`.** `createEdgeGroundSlab()`'s origin-centered `ivec3{60,60,3}` at `(0,0,5)` with `centerAroundOrigin` puts the radius-9 disc fully interior (fixing the old `(-30,-30)` slab that covered only one quadrant), and the `--detached-edge` call ordering (slab before the detached canvas, so it allocates from the main pool) is correct. All three edge flags are pre-existing (`main.cpp:224/262/285`), so the SDF-floor skip condition references nothing undefined. Creation-scoped, not core.

### Agreeing with Sonnet's one nit (non-blocking)

Worst-case occlusion-buffer reads rise from 2 to 8 for hard-disc cut-band pixels only, with early-exit on first solid and sequential-along-ray access. No committed perf baseline yet, but the cost is bounded and confined to a tiny pixel subset — not a merge concern.

The `fleet:needs-windows-smoke` label is already on the PR; the GLSL path is byte-mirrored to the Metal side that passed render-verify, so I expect the Linux/GL smoke to be a formality but it should still run.

Clean, correctly-scoped, invariant-safe.

🤖 Opus recheck (Claude Opus 4.8) — final reviewer pass

PR #2249, opened 2026-07-06
perf_grid: solid contiguous grid + tear-free traveling-wave default

jakildev:
## Summary
- Default IRPerfGrid's voxel_set scene to `spacing = 1.0` (contiguous 1³ voxels) so the 64³ grid renders as a **solid cube** instead of a 1/8-density lattice that read as hollow under yaw and striped at cardinals.
- Add a `wave_mode` knob (config + `--wave-mode`): `per_cell` (default) advances phase with `(x+y+z)` so a wave travels through the solid block; amplitude is tuned tear-free (default `5.0`, keeps every adjacent rounded-cell step ≤ 1 voxel). `rigid` glides the whole block along the screen-right iso axis in shared phase and serves as the solidity/occupancy-cull reference. Per-entity update cost is identical in both modes.
- New preset `configs/perf/sparse_lattice.lua` preserves the prior sparse+shearing scene for scatter-path perf measurement.
- Migrate the demo's remaining `std::max` calls to `IRMath::max` (touch-it-migrate-it).

## Test plan
- [x] `fleet-build --target IRPerfGrid` green (macOS/Metal)
- [x] `--auto-screenshot` before/after pairs below: solid cube at cardinal + yaw 0.35, zoom 1 and 4
- [x] `--config-preset configs/perf/sparse_lattice.lua` reproduces the legacy sparse scene

## Notes for reviewer
The solid scene exposes a real engine artifact this PR deliberately does **not** hide: a 1-cell seam along the coordinate zero-planes, caused by tie-unsafe hardware `round()` at half-integer voxel positions in the voxel→trixel shaders (vs the engine's documented `roundHalfUp` convention, ir_iso_common.glsl:376-384). Fixed in the stacked follow-up PR; the visible z=0 seam in the -after shots is that bug, not a scene regression.

## Screenshots

<details>
<summary>zoom1_origin (cardinal)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom1_rot (yaw 0.35 — the "hollow" repro)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_rot</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_pan (cardinal)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_rot_pan</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>fit_grid / profiler_overlay</summary>

| Before | After |
|--------|-------|
|  |  |
|  |  |

</details>


jakildev (commented):
## Review — perf_grid: solid contiguous grid + rigid glide-right wave by default

**Verdict:** approve

### Nits

- `creations/demos/perf_grid/main.cpp:671` (also mirrored in the enum doc comment at :95, `config.lua:41`, and the `--wave-mode` CLI help at :519) — the "rigid glides screen-right" claim looks backward per the canonical projection. `IRMath::pos3DtoPos2DIso` (`engine/math/include/irreden/ir_math.hpp:1023-1031`) defines `iso.x = -x + y`. For direction `normalize(vec3(1, -1, 0))`, that gives `iso.x = -(1) + (-1) = -2` (negative) while `iso.y` does stay flat (`0`, confirmed) — so this vector points toward *decreasing* `iso.x`. No X-mirroring backend flip is documented anywhere near this helper (only the Y canvas-to-screen flip is called out), so if `iso.x` increasing conventionally maps to "further right" on screen, `(1, -1, 0)` is actually the screen-*left* direction; `(-1, 1, 0)` would be the one that glides right. Since the motion is a symmetric ease -1↔+1 oscillation (both extremes are visited every cycle regardless of sign), this doesn't affect the demo's actual value as a solidity/coherent-motion stress test — it's a labeling nit across four comment/help-text sites, not a behavior bug. Worth a quick sign flip or wording tweak (e.g. "along the screen-diagonal iso axis") so the flag name and screenshots don't quietly assert the wrong direction to a future reader.

### Praise

- Clean three-way default sync — `config.lua`'s `spacing = 1.0` / `wave_mode = "rigid"` match the new C++ defaults (`PerfGridSettings::spacing_`, `waveMode_`) and the `--wave-mode` CLI default string, so there's no silent divergence between the three places a default could live.
- `sparse_lattice.lua` correctly preserves the legacy sparse/shearing scene as an explicit opt-in preset rather than deleting that stress coverage outright — good scatter-path perf comparator retained.
- The "Notes for reviewer" callout on the z=0 seam artifact (tie-unsafe `round()` vs `roundHalfUp`) is exactly the kind of deliberate-scope flag that keeps a reviewer from misdiagnosing a known, already-tracked issue as a regression in this PR.
- Full `std::max`/`std::numbers` → `IRMath` migration is complete and correctly scoped to the touched functions (verified no leftover `std::max`/`std::numbers` calls remain in the file).

- [x] `fleet-build --target IRPerfGrid` (author-reported, macOS/Metal)
- [x] `--auto-screenshot` before/after pairs (author-provided)
- [x] `--config-preset configs/perf/sparse_lattice.lua` reproduces the legacy scene (author-reported)

jakildev:
Addressed the directional-labeling nit — but **not** with the suggested sign flip, which would have introduced a bug.

The nit rested on the premise that *"no X-mirroring backend flip is documented anywhere near this helper"*. There is one: `pos3DtoPos2DScreen` (`engine/math/include/irreden/ir_math.hpp:1100-1102`) projects via `pos3DtoPos2DIso(pos) * stepSize * vec2(-1.0f, screenYDirection_)` — it **negates iso.x**. So for world `(1,-1,0)`:

- `iso.x = -x + y = -2`, then `screen.x = -iso.x = +2` → **screen-right** ✓
- `iso.y = -x - y + 2z = 0` → flat screen-Y ✓

The existing `(1,-1,0)` / "glides screen-right" labels are correct; flipping to `(-1,1,0)` would have made the block glide screen-*left* while the labels kept saying right. (The oscillation is symmetric so behavior/screenshots are unaffected either way — this was purely about the label.)

Fix (comment-only, one commit `a01e0d48`): made the derivation explicit at the vector-definition site in `makeWaveIdle` so the non-obvious `(1,-1,0)→right` mapping (via the downstream X-negation) can't be re-flagged. `fleet-build --target IRPerfGrid` clean. No behavior change; approval stands. — worker

jakildev:
Post-approval update (user direction): the demo's signature look is the **per-cell oscillating wave**, and the original complaint was only that voxels went missing. New commit flips the default back to `wave_mode = per_cell` and makes it lossless instead of removing it: at spacing 1.0 the phase gradient shears face-adjacent cells by at most `2·a·sin(π/λ)`, so amplitude ≤ `λ/(2π) ≈ 5.09` keeps every adjacent rounded-cell step ≤ 1 — the block stays watertight while the wave travels through it. Default amplitude is now **5.0** (was 6.0, which sheared up to 1.18 units and opened transient see-through slits — push it back up deliberately for a scatter stress). `rigid` remains as the solidity/occupancy-cull reference mode.

The `*-after.png` set is re-captured with the per-cell default; the visible AO striping on the stepped faces is honest shading of the stepped geometry, not missing voxels (the earlier hollow/dot-cloud yaw view is gone — compare zoom1_rot before/after). Re-review requested via fleet:changes-made.

jakildev (commented):
## Review — perf_grid: solid contiguous grid + tear-free traveling-wave default

**Verdict:** approve

Re-review: two commits landed since my prior pass (`a01e0d48`, `b3f65064`), the second of which changes the shipped default (`rigid` → `per_cell`) per user direction after merge-approval.

### Prior-review resolution

- ✅ `creations/demos/perf_grid/main.cpp:671` (rigid-wave direction claim) — verified fixed at `a01e0d48`, and my original premise was actually wrong, not just the wording. I'd claimed "no X-mirroring backend flip is documented anywhere near this helper" — there is one: `pos3DtoPos2DScreen` (`engine/math/include/irreden/ir_math.hpp:1100-1102`) computes `pos3DtoPos2DIso(position) * step * vec2(-1.0f, screenYDirection_)`, negating `iso.x` unconditionally on both backends. Re-derived independently: world `(1,-1,0)` → `iso.x = -2` → `screen.x = -(-2) = +2` → screen-right. The existing `(1,-1,0)` label was correct all along; the sign-flip I suggested would have made `rigid` glide screen-*left* while every label kept saying right. Good catch by the author declining the suggested fix and correcting my analysis instead of applying it blindly.

### New findings (from `b3f65064`)

Independently re-derived the tear-free amplitude claim rather than taking it on faith: with `wavelength = max(8, gridSize·0.5) = 32` at the default `gridSize_ = 64`, adjacent-cell phase difference is `Δφ = 2π/λ`, and for two sinusoids of equal amplitude `a` offset by `Δφ`, the max instantaneous difference is `2a·sin(Δφ/2) = 2a·sin(π/λ)`. At `λ = 32`: `2·sin(π/32) ≈ 0.19603`, so the exact tear threshold (shear = 1 voxel) is `a ≤ 1/0.19603 ≈ 5.101`, which the comment's small-angle approximation `λ/(2π) ≈ 5.093` slightly understates but stays safely under. The chosen default of `5.0` clears both the exact and approximated threshold with a small margin — the math holds, and the two-stage `kSineEaseInOut` waveform in `makeWaveIdle` (main.cpp:692-705) is close enough to a true sinusoid for the derivation to apply.

### Nits

- `creations/demos/perf_grid/main.cpp:680` — the short inline comment right above the `Rigid`/`PerCell` branch still calls `PerCell` "legacy z-axis wave," but the fuller enum-doc comment two lines above (main.cpp:94, updated in this same commit) now correctly calls `PerCell` "the default." Same commit left these two descriptions of the same enum value contradicting each other.
- `creations/demos/perf_grid/configs/perf/sparse_lattice.lua:6` — "The default scene (config.lua) is the solid contiguous block with rigid motion" is now stale: config.lua's default `wave_mode` flipped to `per_cell` in this same commit. Worth a one-line fix so a reader of this preset doesn't get the wrong idea of what the base scene actually does by default.
- PR body Summary section still reads "`rigid` (default) glides the whole block..." — no longer matches the shipped default (`per_cell`) after the post-approval change. Doesn't block merge (the code and screenshots are current) but worth updating before a human skims it.

### Praise

- Good instinct not to reflexively revert to the reviewer's suggested fix when the underlying premise didn't hold up — verifying against the actual downstream transform (`pos3DtoPos2DScreen`) instead of trusting the review's assertion is exactly the right response to a "verify cited precedents" nit.
- The amplitude retune is derived, not guessed: the commit message and code comment both show the actual shear-bound math rather than an empirically-chosen "looks fine" number, and the chosen value (5.0) has a real safety margin under the derived threshold (~5.09-5.10), not a knife's-edge value.
- `*-after.png` screenshots were re-captured against the new default rather than left stale from the `rigid`-default pass — the PR's own visual evidence matches what actually ships.

- [x] `fleet-build --target IRPerfGrid` (author-reported, macOS/Metal, both follow-up commits)
- [x] `--auto-screenshot` before/after pairs re-captured against the `per_cell` default (author-provided)
- [ ] Optional: sync the two stale comments + PR body Summary listed above (Nits) — no behavior change, just doc accuracy

jakildev:
Addressed remaining nits from re-review (`2eed3360`):
- `main.cpp:680` — synced the short inline comment above the Rigid/PerCell branch (was stale "legacy z-axis wave") to match the fuller enum-doc comment, which already called PerCell the default.
- `configs/perf/sparse_lattice.lua:6` — fixed the stale claim that the base scene (config.lua) uses rigid motion; it's per_cell now.
- PR body Summary — updated to describe per_cell as the default with the tear-free amplitude tuning, rigid as the solidity/occupancy-cull reference.

No behavior change, comment/doc-only. `fleet-build --target IRPerfGrid` clean. — worker

jakildev:
Verified all three nits from the 02:00 re-review are already resolved at the current head `2eed3360` ("sync stale per_cell wave-mode comments"): main.cpp:680 inline comment now reads "PerCell: the default", sparse_lattice.lua:6 now describes the base scene as the tear-free per-cell wave (not rigid), and the PR body Summary lists per_cell as the default. The `fleet:has-nits` label was orphaned (fix pushed, label not cleared); clearing it now. `fleet:approved` stands — ready for merge. — worker

PR #2251, opened 2026-07-06
render: roundHalfUp voxel-cell conversion in stage 1/2 + compact shaders

jakildev:
## Summary
- Replace tie-unsafe hardware `round()` with the existing `roundHalfUp` helper at every voxel-position→integer-cell conversion in `c_voxel_to_trixel_stage_{1,2}` and `c_voxel_visibility_compact`, GLSL **and** Metal (12 sites): cardinal 3-face path, subdivision path, per-axis base store, fog world-column recovery, shadow-feeder classification.
- Hardware `round()` ties are implementation-defined (Metal: away-from-zero; GL: often round-half-even), violating the documented convention at `ir_iso_common.glsl:376-384` — a solid 64³ grid centered at half-integers left integer cell 0 unwritten on each axis: a one-cell black seam along every coordinate zero-plane (and on GL, round-half-even can collapse *pairs* of rows).
- `roundHalfUp` == `round()` for all non-tie inputs → integer-positioned scenes are byte-identical.

## Stack context
Stacked on: https://github.com/jakildev/IrredenEngine/pull/2249

When the parent PR merges, change this PR's base to `master`
(via the GitHub UI or `gh pr edit <N> --base master`).

## Test plan
- [x] `fleet-build` IRPerfGrid + IRShapeDebug green (macOS/Metal)
- [x] Static solid probe (spacing 1.0, wave 0 — all axes at exact ±0.5 ties): zero-plane seams **gone** at cardinal and yaw
- [x] Default IRPerfGrid scene: z=0 seam gone (before/after pairs below)
- [x] Byte-identity: 26/28 IRShapeDebug `--auto-screenshot` shots byte-identical vs parent; 2 shots drift 8–28 px (0.0002–0.0008%) confined to rotated re-voxelized silhouettes whose cells sit exactly on ties — the intended correction

## Notes for reviewer
Verify note for the Linux smoke: GLSL `round()` is frequently round-half-even on GL drivers, which pairs-collapses tie rows — the fix matters *more* there. `snapNearIntegerVoxelPosition`'s internal `round()` is deliberately untouched (it only snaps within 1e-4 of integers; ties at .5 are out of its range).

## Screenshots

<details>
<summary>zoom1_origin (cardinal — z=0 seam across the front faces)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>fit_grid</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
[...]

PR #2254, opened 2026-07-06
render: compact-pass fully-interior voxel drop + perf_grid cross-set occupancy

jakildev:
## Summary
- **Engine**: the visibility compact now drops a voxel from the cardinal single-list append when **all six face-occlusion bits are set** (split mode already did this per-axis). Such a voxel's every slot fails `faceIsExposed` in stage 1/2 and the rotated opposite-polarity emit requires an exposed opposite face, so the drop is **output-identical** — it removes only sub²-per-voxel no-op stage-1 dispatches. Gated inert while any fog vision circle is live (circles can revive occluded vertical faces as #2125 cut walls). Both backends; append stays a guarded branch so the completion barrier is uniformly reached.
- **Demo**: perf_grid voxel_set mode stamps the analytic cross-set occlusion mask at creation (1³ sets can't see lattice neighbors), gated to `spacing == 1.0` + rigid wave (cells provably stay touching under in-phase motion).

## Perf (solid 64³ voxel_set, macOS/Metal, settled `--auto-profile 300`)

| zoom | GPU voxelStage1 before → after | frame before → after |
|---|---|---|
| 1 | 2.42ms → **0.69ms** | 10.9ms → 6.1ms |
| 4 | ~20ms → **4.07ms** | 22.0ms → 6.6ms (120 FPS) |
| 8 | 47.6ms → **6.12ms** | 54.9ms → 10.5ms |
| 16 | 145.2ms → **12.64ms** | 161.3ms → **16.5ms** (7→60 FPS) |

## Stack context
Stacked on: https://github.com/jakildev/IrredenEngine/pull/2251

When the parent PR merges, change this PR's base to `master`
(via the GitHub UI or `gh pr edit <N> --base master`).

## Test plan
- [x] Byte-identity: filter-on vs filter-off `--auto-screenshot --no-overlay --wave-amplitude 0` — 5/5 deterministic shots byte-identical (`cmp`)
- [x] The 2 rotated shots differ run-to-run **with identical binaries** — pre-existing per-axis non-determinism, control-tested (same-build rerun drifts the same two shots); tracked as a follow-up issue
- [x] Zoom sweep table above; HUD pairs below show CULL VOX 262144 → ~24k and the FPS delta

## Notes for reviewer
Correctness argument for the drop, in one place: a fully-occluded voxel cannot emit (a) any triplet face (`keepFace = faceIsExposed(...)` fails all three slots), (b) any fog cut face (requires `fogActive` ⇒ `visionCircleCount > 0`, exactly when the drop disarms), (c) any rotated opposite-polarity face (`bothPolaritiesExposed` needs an exposed opposite). Stage 2 mirrors the same gates, and the shadow bake/AO read only `trixelDistances`, which such a voxel never wrote.

## Screenshots (HUD perf evidence — visuals identical by design)

<details>
<summary>profiler_overlay (zoom 0.5) — VOX cull count + GPU stages</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_pan / zoom4_rot_pan</summary>

| Before | After |
|--------|-------|
|  |  |
|  |  |

</details>

<details>
<summary>zoom1_origin / zoom1_rot / zoom4_rot / fit_grid</summary>

| Before | After |
|--------|-------|
|  |  |
|  |  |
|  |  |
|  |  |

</details>


jakildev:
#2251. Its head ref `claude/render-roundhalfup-ties` was force-
pushed since this PR was last rebased; the child commits replayed
cleanly onto the new upstream tip. Force-pushed with
`--force-with-lease`. CI will re-run.

— fleet merger

jakildev (commented):
## Review — render: compact-pass fully-interior voxel drop + perf_grid cross-set occupancy

**Verdict:** approve

Stacked on #2251 (open, `fleet:approved`). Reviewing the child diff only — upstream is approved separately.

**Diff-scoping note:** `gh pr diff` reports 35 changed files including a full copy of #2251's `roundHalfUp` rename across `c_voxel_to_trixel_stage_1/2.{glsl,metal}`. That's stale — this branch forked before #2251's `roundHalfUp` commit was amended in review, so it carries a content-identical but differently-SHA'd duplicate of that commit (`git log origin/claude/render-roundhalfup-ties..origin/claude/render-compact-occlusion-filter` shows two commits; a straight tip-to-tip `git diff` against the current `render-roundhalfup-ties` head shows zero net change in those files). I reviewed the real diff: `creations/demos/perf_grid/main.cpp`, `c_voxel_visibility_compact.glsl`, `c_voxel_visibility_compact.metal`, plus screenshots. Flagging for the merger: rebase onto #2251's current tip (or onto `master` once #2251 merges) before landing, to drop the duplicate commit from history — harmless today since the content already matches, but worth cleaning up so the merge doesn't carry a confusing no-op commit.

### Correctness verification

- `kFaceOccludedMaskBits = 0xFCu` — checked against `VoxelFlags::kFaceOccludedNegX..PosZ` (`component_voxel.hpp:37-45`, bits 2..7) → `0xFC` is exactly right.
- `faceIsExposed(flagsByte, faceId)` (`ir_iso_common.glsl:281`) returns `((flagsByte >> (2+faceId)) & 1) == 0`. When `flagsByte & kFaceOccludedMaskBits == kFaceOccludedMaskBits`, every bit 2..7 is set, so `faceIsExposed` is false for all six `faceId` values — confirms the comment's claim that a fully-interior voxel can't pass the stage-1/2 exposed-face gate.
- The guarded append sits inside the existing `if (...) { }` with no early `return` — the trailing `barrier(); memoryBarrierBuffer();` (both GLSL and Metal) is still reached uniformly by every invocation in the group, so the completion-barrier concern the comment calls out is correctly handled.
- `flagsByte = (voxels[idx].materialFlagBone >> 8u) & 0xFFu` matches the existing extraction already used in the per-axis-split branch a few lines below (and the documented GPU-mirror convention in `component_voxel.hpp:31-32`) — no new/invented API.
- `perf_grid/main.cpp`: the cross-set occupancy stamp only fires for `WaveMode::Rigid && spacing_ == 1.0f`. `positionForCell` places cell centers at `(coord - center) * spacing_`, so at `spacing_ == 1.0` with 1×1×1 voxel sets, adjacent grid cells touch exactly — the analytic occlusion bits (`x>0` ⇒ NegX occluded, etc.) are correct for that lattice. Rigid mode moves the whole block in phase so relative spacing is preserved every frame; the per-cell wave mode (correctly excluded) shears gaps open.
- Checked that nothing re-derives face occupancy on these entities per-frame and clobbers the stamp: `REBUILD_GRID_VOXELS` (the only system that calls `recomputeFaceOccupancyOnCells` on a live per-frame basis) requires `C_RotationMode` in its query (`system_rebuild_grid_voxels.hpp:579`), which these cell entities never get, and `perf_grid` doesn't even register that system. `UPDATE_VOXEL_SET_CHILDREN` only pushes position, never flags. So the one-time stamp at creation survives.
- `getComponentOptional<C_VoxelSetNew>` right after `createEntity` mirrors the exact pattern already used earlier in the same file (`main.cpp:668-674`) — consistent, not a new anti-pattern, and this is one-time setup code (`createGridEntities`), not a per-entity tick, so the "no getComponent in tick" rule doesn't apply here anyway.

### Nits
- `creations/demos/perf_grid/main.cpp` (new gate): `g_settings.spacing_ == 1.0f` is an exact float equality check on a CLI/Lua-configurable value. Low risk in practice (1.0 round-trips exactly through typical float parsing and the default is a literal `1.0f`), but a `spacing_` sourced from an odd config path that lands at 0.999999f would silently disable the optimization rather than erroring — harmless (falls back to the correct-but-slower path), just worth being aware of if `spacing_` parsing ever changes.

### Praise
- The fully-interior drop is mirrored correctly and completely between the GLSL and Metal backends (identical guard, identical mask constant, identical comment structure) — good parity discipline.
- The correctness argument is spelled out inline (why a fully-interior voxel can't emit via triplet/fog-cut/rotated-opposite paths) rather than just asserting "safe" — this made independent verification straightforward.

- [ ] Rebase onto #2251's current tip (or `master` post-merge) to drop the duplicate inherited commit before this lands.

Opus recheck required: touches `engine/render/src/shaders/` (compute-shader compact pass — atomic append gating + cross-group completion barrier).

jakildev (commented):
## Opus recheck — render: compact-pass fully-interior voxel drop + perf_grid cross-set occupancy (#2254)

**Verdict:** approve

Stacked on #2251 (open, `fleet:approved`). Reviewed the child diff only (single commit `ce873c98` vs the base tip) — upstream is approved separately. Sonnet already verified the byte-level correctness (`kFaceOccludedMaskBits = 0xFC` vs `VoxelFlags` bits 2..7, `faceIsExposed`, the demo-stamp survival argument, float-equality nit). This pass focused on the three concurrency/GPU items Sonnet escalated for Opus recheck: the atomic-append gating, the cross-group completion barrier, and GLSL/Metal parity.

### Opus-only checks (what Sonnet could not confirm without the compute-shader control-flow read)

- **Barrier uniformity — confirmed sound.** Sonnet flagged that gating the append behind a new `if` must not break the completion barrier. On closer read: `barrier(); memoryBarrierBuffer();` (GLSL) and `threadgroup_barrier(mem_flags::mem_device)` (Metal) both sit at the *top level* of `main`/the kernel — outside every one of the nested `if` blocks (`idx < voxelCount` → `chunkVisible` → active-mask → cull/fog → per-route). The new fully-interior guard adds no early `return` and no branch around the barrier, so every invocation still reaches it. The "no early return" comment is accurate, not aspirational.
- **Cross-group completion counter — no new ordering hazard.** The last-group election (`params[4]` / `kSlotCompletedGroups` atomicAdd, then `writeDispatchDims` when `finished == totalGroups`) is unchanged. The drop's only effect on shared state is *fewer* `atomicAdd(params[3])` calls, which shrinks the indirect dispatch count — exactly the intended perf win. Stage 1/2 iterate `compactedVoxelIndices[0, params[3])`, slots are assigned by sequential atomicAdd, and dropped voxels simply don't consume a slot, so no stale/garbage entry can be read. Correctness intact.
- **Output-identical drop — confirmed.** A voxel with all six occlusion bits set fails `faceIsExposed` for every `faceId`, so it emits nothing in stage 1/2, and — the part that mattered to verify independently — the rotated opposite-polarity emit requires an *exposed opposite* face, which a fully-interior voxel by definition never has. Skipping its append is therefore observably identical downstream.
- **Fog vision-circle exception — correctly conservative.** The drop is gated off globally whenever `visionCircleCount != 0` (`fogObservers.visionCircleCount` on Metal), not per-column. That's stricter than the per-column `fogColumnInVisionCircle` reveal test, so it can never carve a hole a live cut wall (#2125) would need — it only forgoes the optimization while any circle is live. Safe.
- **GLSL/Metal parity — exact.** Identical mask constant, identical guard predicate, identical placement relative to the barrier, and the per-axis-split branch (untouched by this drop) already lands fully-interior voxels in no region, so the single-list path now matches that established behavior.

### Agreement with Sonnet
- The stale duplicate `roundHalfUp` commit in `gh pr diff` is a history-cleanup item for the merger (the `fleet:merger-cooldown` label shows a rebase already cascaded), not a code defect — the real net diff in those shader files is zero.
- The `spacing_ == 1.0f` exact-float gate in the demo is a harmless fall-back-to-slower-path nit, not a blocker.

No blocking issues. Approving.

jakildev:
Addressed nit: replaced exact-equality `g_settings.spacing_ == 1.0f` gate with an epsilon-tolerant `IRMath::abs(g_settings.spacing_ - 1.0f) < 1e-4f` check, so a near-1.0 spacing that doesn't round-trip exactly still gets the fast analytic-occlusion path instead of silently falling back. Build (IRPerfGrid) clean. — worker

jakildev:
Rebased onto the updated stack (parent now defaults to `wave_mode = per_cell` at tear-free amplitude 5.0, per user direction; the fleet's epsilon-tolerance amendment `da76f377` is preserved as `799ccd94`). Scope note for the perf table in the body: the demo-side occupancy mask arms only in `--wave-mode rigid` (the always-touching lattice where the mask is provably valid — the per-cell wave shears steps open, so a static mask would carve holes). The zoom-sweep numbers therefore describe the rigid reference mode; re-verified on the rebased tip: zoom 16 rigid = 17.1ms frame / 60 FPS. The engine-side compact drop is mode-independent and benefits any scene whose voxels carry face-occlusion bits (dense mode, single-set worlds, game terrain). The default per-cell wave scene intentionally rasters its transiently-exposed interior — that cost is honest geometry, and the remaining high-zoom lever for it (shadow-feeder sub² capping) is noted in #2256.

jakildev:
Merger: cascade-rebased onto updated base PR #2251. Its head ref `claude/render-roundhalfup-ties` was force-pushed since this PR was last rebased; the child commits replayed cleanly onto the new upstream tip. Force-pushed with `--force-with-lease`. CI will re-run.

— fleet merger

jakildev:
Merger: base PR #2251 merged; re-targeted this PR to `master` but the
rebase onto current master conflicts. The child commits cannot be
replayed cleanly — worker will resolve semantically.

Conflicted files (10 total, showing first 5):
- `creations/demos/perf_grid/config.lua` — master: `34bad77a perf_grid: solid contiguous grid + tear-free traveling-wave default (#2249)`; PR: `b3f65064 perf_grid: default to the per-cell traveling wave at tear-free amplitude`
- `creations/demos/perf_grid/configs/perf/sparse_lattice.lua` — master: `34bad77a perf_grid: solid contiguous grid + tear-free traveling-wave default (#2249)`; PR: `2eed3360 fix: sync stale per_cell wave-mode comments to per_cell being the default`
- `creations/demos/perf_grid/main.cpp` — master: `34bad77a perf_grid: solid contiguous grid + tear-free traveling-wave default (#2249)`; PR: `a38839c7 perf_grid: use epsilon tolerance for spacing_==1.0f gate`
- `docs/pr-screenshots/claude/perfgrid-solid-grid/fit_grid-after.png` — master: `34bad77a perf_grid: solid contiguous grid + tear-free traveling-wave default (#2249)`; PR: `b3f65064 perf_grid: default to the per-cell traveling wave at tear-free amplitude`
- `docs/pr-screenshots/claude/perfgrid-solid-grid/profiler_overlay-after.png` — master: `34bad77a perf_grid: solid contiguous grid + tear-free traveling-wave default (#2249)`; PR: `b3f65064 perf_grid: default to the per-cell traveling wave at tear-free amplitude`
… and 5 more

Both master's #2249 (solid contiguous grid + traveling-wave default,
merged just before #2251) and this PR touch `perf_grid`'s config/main
and the same `perfgrid-solid-grid` screenshot set — the PNG conflicts
are add/add on binary reference images, which need a human/worker
regen rather than a mechanical merge.

Labeled `fleet:semantic-conflict` — a worker will attempt resolution
on its next iteration (rebase, manually resolve, build, push). If the
worker also can't resolve (truly ambiguous, design decision needed),

The `fleet:approved` label has been removed if it was set — the PR no
longer represents a reviewed state.

SHA pair: master=c416457e773159a04df8bd9a959c22f6ec572aa2 × PR=a38839c7314332880a12cae1aa28fdd75a3e238f

— fleet merger

jakildev:
Resolved semantic conflict via inherited-prefix drop (`git rebase --onto origin/master e3fff803`). Base PR #2251 merged and re-targeted this PR to `master`; the 5 inherited commits (roundHalfUp `e3fff803` — now in master as #2251; and the parent-stack perf_grid work `8c205b1d`/`a01e0d48`/`b3f65064`/`2eed3360` — superseded by #2249) were the entire conflict surface, including the `perfgrid-solid-grid/*.png` add/add binaries and the `config.lua`/`sparse_lattice.lua` overlaps. Dropping them leaves only this PR's two genuine, already-reviewed commits (`compact-pass fully-interior voxel drop + perf_grid cross-set occupancy` and the `epsilon-tolerance spacing gate`), which replayed **cleanly** onto current master (no manual conflict resolution needed). Net diff is now exactly `perf_grid/main.cpp` + the two `c_voxel_visibility_compact` shaders + this PR's own `render-compact-occlusion-filter/*.png` screenshots — matching the diff the reviewer said they actually reviewed. Build clean (`IRPerfGrid`) and `--auto-screenshot` smoke green on macos-debug/Metal (7/7 shots, no crash). `fleet:needs-windows-smoke` retained for the Linux/OpenGL pass. Reviewer please re-evaluate the (now smaller) rebased diff. — worker

jakildev (commented):
## Review — render: compact-pass fully-interior voxel drop + perf_grid cross-set occupancy

No longer stacked: upstream #2251 merged and the merger re-targeted this PR to `master`. That re-target hit a semantic conflict (`creations/demos/perf_grid/config.lua` and others — master already had #2249/#2251's own perf_grid/shader changes) which a worker resolved via `git rebase --onto origin/master e3fff803` — dropping the 3 inherited parent-stack commits (`8c205b1d`/`a01e0d48`/`b3f65064`, perf_grid solidity/wave-default work — now #2249 on master) and the inherited `roundHalfUp` commit (`e3fff803`, now #2251 on master), replaying only this PR's own 2 commits on top. This is a fresh re-review of the rebased head, not a stacked-PR review.

### Prior-review resolution

- ✅ `engine/render/src/shaders/c_voxel_visibility_compact.glsl` / `metal/c_voxel_visibility_compact.metal` (atomic-append gating + cross-group completion barrier, Opus recheck approve 01:26:34Z) — diffed the current head against the exact commit Opus reviewed (`ce873c98`): **zero diff** in either shader file. The escalated shader logic is untouched by the subsequent nit fix and all three rebases.
- ✅ `creations/demos/perf_grid/main.cpp:792` (Sonnet nit, first pass 01:11:34Z: exact-float `spacing_ == 1.0f` gate could silently disable the fast path on an off-path config value) — fixed at `d0b51922`, present unchanged post-rebase: `IRMath::abs(g_settings.spacing_ - 1.0f) < 1e-4f`. `IRMath::abs<T>` is the existing generic `glm::abs` wrapper (`ir_math.hpp:215`), already used elsewhere in this same file (`IRMath::max` at line 636), so no new include/dependency risk.
- ✅ Stale duplicate `roundHalfUp` commit noted in `gh pr diff` (Sonnet first pass) — resolved by the inherited-prefix-drop rebase; confirmed the current head carries zero net diff against `origin/master` in those shader files outside this PR's own hunks (verified below).

### Verification of the rebase itself

The worker's rebase comment claims the child diff is "now scoped to #2254's own commits" against current master. Checked directly rather than trusting the description:

- `git diff origin/master...HEAD --stat` shows exactly 17 files: `main.cpp`, the 2 shader files, and 14 before/after PNGs under `docs/pr-screenshots/` — matches the PR's stated scope, no out-of-scope file.
- `git diff origin/master...HEAD -- main.cpp <shaders>` shows only the cross-set-occupancy stamping block in `main.cpp` and the fully-interior-drop guard in both shaders — the same hunks originally reviewed, nothing extra pulled in by the rebase.
- Confirmed the 4 dropped commits (`e3fff803`, `8c205b1d`, `a01e0d48`, `b3f65064`) are genuinely already on `master` under different SHAs as #2249 and #2251 (both visible in `git log master` — "perf_grid: solid contiguous grid + tear-free traveling-wave default (#2249)" and "render: roundHalfUp voxel-cell conversion in stage 1/2 + compact shaders (#2251)") — not scope this PR owns, correctly excluded rather than silently lost.
- `mergeable: MERGEABLE` — no churn audit needed (step 1d only applies to CONFLICTING).

No new issues found. The epsilon fix is a straightforward, correct tightening of an already-non-blocking nit; the rebase is mechanical, as claimed.

**Verdict:** approve

### Praise

- The PR's own "Notes for reviewer" section pre-empted exactly the concurrency question Opus escalated on (why gating the append doesn't break the completion barrier) with a one-line correctness argument per emit path — the kind of author-side reasoning that makes an Opus recheck a confirmation pass instead of a from-scratch derivation.
- Verifying the dropped commits were true master-duplicates (not silently-lost scope) took one `git log --all --oneline | grep` per SHA — cheap insurance against the "inherited-prefix drop accidentally ate real work" failure mode.

- [x] Shader logic byte-identical to the Opus-approved commit — confirmed via direct diff, not re-derived.
- [x] Rebase scope confirmed via `--stat` + per-file diff against `origin/master`.
- [ ] Linux/OpenGL smoke — still outstanding per the PR's own test plan; `fleet:needs-windows-smoke` stays on this PR.

Opus recheck not required: the only surface Opus needed to confirm (the compute-shader atomic-append gate and completion barrier) is byte-identical to the commit Opus already approved. The delta since that approve is a non-render C++ epsilon-tolerance fix (verified correct above) and three mechanical rebases (verified no net content change beyond dropping already-merged parent-stack commits).

Issue #2262, opened 2026-07-06
fleet/install-symlink-lag: address recurring feedback


jakildev:
## Context

Surfaced by `scripts/fleet/review-fleet-feedback` from aggregated fleet feedback.
Signature: `install-symlink-lag` · 7 occurrences since 2026-07-04T00:00:00.
Tracked locally in `~/.fleet/feedback/.fix-log.jsonl::fix-017` (proposed 2026-07-06T03:08:39).

## Representative feedback entries

- 2026-05-01 00:45 [sonnet-reviewer] PR #382 merger posted "semantic conflicts" comment but never set `fleet:semantic-conflict`
- 2026-07-04 [sonnet-reviewer] fleet-review-verdict missing from ~/bin symlinks
- 2026-07-04 23:55 [sonnet-reviewer] fleet-review-verdict not symlinked into ~/bin after being added to scripts/fleet/ (mtime 2026-07-04 02:23) — REVIEWER-PROTOCOL.md already documents it as the current verdict-swap mechanism, but the fresh sonnet-reviewer worktree's `~/bin` had no symlink for it (only the older `fleet-transition` was linked). Had to re-run `scripts/fleet/install.sh --no-zshrc` mid-review to pick it up. Suggest: fleet-up (or the dispatcher's per-pane bootstrap) re-runs install.sh automatically when scripts/fleet/ has files newer than the last symlink pass, so a newly-added tool doesn't strand the first reviewer to hit it.

## Proposed fix

New fleet scripts merge to master but hosts' ~/bin symlinks never refresh, so the tool is 'command not found' at first use — fleet-review-verdict (merged 2026-07-03, PR #2193) stranded every reviewer on 2026-07-04 (7 entries, opus-reviewer + sonnet-reviewer); each re-invented the manual claim-confirm step the wrapper automates. Same family as the fleet_*.py stale-symlink pattern. Fix: in scripts/fleet/fleet-up (and/or the fleet-dispatcher per-pane bootstrap), re-run scripts/fleet/install.sh --no-zshrc automatically when any file in scripts/fleet/ is newer than the last symlink pass (stamp file in ~/.fleet/). One reviewer already validated the manual re-run workaround (install.sh --no-zshrc mid-review).

## Acceptance

- Implementation matches the artifact called out in the proposed fix above.
- Subsequent `review-fleet-feedback digest` runs do not re-surface the `install-symlink-lag` cluster after the fix lands (verified via the 7-day quiet window).

---

[...]

PR #2193, opened 2026-07-04
fleet: claim-verified verdict apply — fleet-review-verdict guard (#2153)

jakildev:
## Summary

Reviewer verdicts (`approve` / `needs-fix` / `blocker` / `needs-opus-recheck`)
were applied with a raw `fleet-transition verdict-*` that had **no tie to the
reviewing claim**, so an Opus recheck verdict could land on a PR *other* than the
one reviewed (#2141, #2138) — an internally contradictory approve-body +
`fleet:needs-fix` that elects the wrong PR into the worker lane and burns a no-op
dispatch.

This adds **`scripts/fleet/fleet-review-verdict`**, a thin **guard** in front of
`fleet-transition`:

- With `--agent <name>` it reads the PR's live labels and **refuses to apply the
  verdict unless this agent holds `fleet:reviewing-<host>-<agent>` on that PR**
  (exit 4, no writes) — the misroute is surfaced at the source.
- On pass it **delegates to `fleet-transition`** (the sole state-machine
  mechanism), so idempotency + the post-write verify are inherited unchanged.
- Omitting `--agent` is the **interactive-human carve-out** (a human running
  `review-pr` holds no claim). The carve-out is the *absence* of `--agent`,
  deliberately **not** a `--force` flag an agent could add to defeat the guard.
- `--repo` / `--dry-run` thread through to both the claim read and
  `fleet-transition`; validates the edge is one of the five `verdict-*` names.

Plan: `.fleet/plans/issue-2153.md` (first commit). Approach sign-off + two
SOUND plan-review rounds on #2153.

## Worker-landed (this PR, non-gated)

- `scripts/fleet/fleet-review-verdict` — the wrapper (new).
- `scripts/fleet/install.sh` — SRC/DEST + chmod-loop entry + dedicated `ln -sf`
  block (mirrors the full FLEET_TRANSITION install pattern, per the plan-review nit).
- `scripts/fleet/fleet-help` — index entry beside `fleet-transition`.
- `scripts/fleet/tests/test_fleet_review_verdict.sh` — hermetic (stubs
  `gh`/`fleet-transition`/`fleet-claim`); **22/22 pass**.
- `docs/agents/REVIEWER-PROTOCOL.md` — the canonical reviewer verdict apply (6
  code lines + the review-pr-fallback prose) rewired to the guarded form. This
  guards the **primary** misroute path (the opus/sonnet reviewer loop) with
  **zero gated edits**.
[...]

PR #2264, opened 2026-07-06
fleet: review-fleet-feedback — install-symlink-lag signature

jakildev:
## What

Adds an `install-symlink-lag` pattern to the `SIGNATURES` table in `scripts/fleet/review-fleet-feedback`, placed ahead of `label-absent-after-verdict` so install-lag entries that mention the verdict wrapper aren't swallowed by that signature.

## Why

Seven feedback entries (opus-reviewer + sonnet-reviewer, all 2026-07-04) about `fleet-review-verdict` being missing from `~/bin` / not on PATH after PR #2193 merged were invisible to the review loop — all landed as uncategorized singletons, below the cluster threshold. With this signature they cluster and stay tracked by the closed-loop fix-log.

The underlying install-lag fix (auto re-run `install.sh` when `scripts/fleet/` is newer than the last symlink pass) is queued separately as #2262 — this PR only makes the pattern visible to the feedback reviewer.

## Verification

- `python3 -m py_compile` clean.
- Re-ran `review-fleet-feedback digest` against the live feedback dir with this change: the 7 entries cluster as `install-symlink-lag` (fresh-proposable), and `label-absent-after-verdict` correctly retains its 2 genuinely label-related entries.

### DIFF
--- a/scripts/fleet/review-fleet-feedback
+++ b/scripts/fleet/review-fleet-feedback
@@ -67,6 +67,15 @@ ENTRY_HEAD_FILE = re.compile(
 # Pattern bank — order matters; first match wins. Widen aggressively;
 # uncategorized noise drowns the signal otherwise.
 SIGNATURES: list[tuple[str, re.Pattern]] = [
+    # A fleet script merged to master but the host's ~/bin symlink pass never
+    # re-ran, so the tool is "command not found" / missing on PATH. Must sit
+    # before label-absent-after-verdict: these entries often name the verdict
+    # wrapper and would otherwise be swallowed by that signature.
+    ("install-symlink-lag", re.compile(
+        r"(?=.*(~?/bin|symlink|install\.sh|\bPATH\b))"
+        r"(?=.*(not\s+(on\s+PATH|symlinked|linked|present|installed)"
+        r"|missing|doesn.t\s+exist|command\s+not\s+found|stale[- ]symlink"
+        r"|install[- ]lag|un-?symlinked))", re.I | re.S)),
     ("label-absent-after-verdict", re.compile(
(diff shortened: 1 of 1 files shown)


jakildev (commented):
## Review — fleet: review-fleet-feedback — install-symlink-lag signature

**Verdict:** approve

### Praise
- The comment above the new `install-symlink-lag` pattern states the durable invariant ("must sit before `label-absent-after-verdict`: these entries often name the verdict wrapper and would otherwise be swallowed by that signature") rather than narrating the investigation history — the "why now" (7 entries on 2026-07-04) correctly stays in the PR body/commit message, not the code comment.
- Verified the regex against synthetic entries: `install-symlink-lag` correctly matches "fleet-review-verdict missing from ~/bin ... command not found" and "not on PATH after the merge landed" text, while unrelated `label-absent-after-verdict` phrasing ("verdict label was absent after the swap ran", "label swap left fleet:needs-fix stale") still routes to the original signature and isn't hijacked by the new, broader one — the ordering claim in the PR body holds.
- `python3 -m py_compile` and `ruff check` both clean on the modified file.

- [x] `python3 -m py_compile scripts/fleet/review-fleet-feedback` — confirmed clean.
- [x] Re-ran `digest` against the live feedback dir per the PR body's stated verification (7 entries cluster as `install-symlink-lag`, `label-absent-after-verdict` retains its 2) — not independently reproducible here since the marker has since advanced past that window on this host, but the regex-level behavior checks out and this is a low-risk, easily-revertable data-only change (no test infra exists for individual SIGNATURES entries elsewhere in this file, so this isn't a gap specific to this PR).

Opus recheck not required.

Issue #2258, opened 2026-07-06
render: high-zoom stage-1 cost for occupancy-less scenes — cap shadow-feeder subdivision (or generalize interior occupancy)


jakildev:
**Blocked by:** #2280 — the intra-tick GPU sub-stage attribution (compact / clear / stage-1 / stage-2) that decides this issue's lever. Do NOT re-plan this issue until #2280's attribution table is posted on this thread (architect direction, PR #2266 2026-07-06): PR #2266 measured the bundled `voxelStage1` cost invariant to a 256× stage-1 body reduction, refuting Lever 1 below and leaving the true hotspot unattributed.

*(Superseded blocker, kept for history: PR #2254's stack — merged.)*

Residual of the IRPerfGrid zoom-cliff work (PR #2254). The compact-pass interior drop flattens the zoom curve only for scenes whose voxels carry face-occlusion bits. The default IRPerfGrid scene now runs the per-cell traveling wave, whose shearing makes a static occupancy mask invalid — so it (and any scene of independent/moving single-voxel sets) still pays the full pre-#2254 scaling: stage-1 work ≈ (voxels inside the shadow-feeder-widened cull) × effSub².

Why it scales this way: the cull viewport is widened by `IRMath::shadowFeederIsoBounds` (a fixed `kSunShadowMaxDistance` = 64-world-unit sweep toward the sun), so past zoom ~4 the widened bounds swallow the whole 64³ volume — nothing culls — while per-voxel raster work grows as effSub². Measured pre-#2254 (equivalent to the occupancy-less path): GPU voxelStage1 2.4 / ~20 / 47.6 / 145.2 ms at zoom 1/4/8/16.

Two candidate levers, to be weighed in planning:

1. **Cap feeder subdivision.** A voxel outside `visibleIsoBounds` (already computed for the #1740 stage-2 feeder skip) exists only to write depth for the sun-shadow bake. Raster feeders at `min(effSub, feederSubCap)` instead of full effSub. Quality question to settle: the bake's sun-space texel density (1024² map over the swept AABB) vs the feeder's canvas-depth density — too coarse a cap makes sparse canvas depth project holes into the sun map. The cap likely wants to be derived from the bake's texels-per-world-unit, not a constant. Any change must keep the visible-region raster byte-identical; feeder-region shadow edges from off-screen casters may coarsen (needs a design call + render-debug-loop evidence).

[...]

PR #2266, opened 2026-07-06
render: extract sunBakeFrustumUVBounds shared helper (#2258 prep)

jakildev:
## Summary

Byte-identical extraction of `BAKE_SUN_SHADOW_MAP`'s per-cascade swept
sun-space AABB corner math into
`IRPrefab::SunShadow::sunBakeFrustumUVBounds` (`sun_shadow_constants.hpp`).
The corner enumeration and `min`/`max` reduction order are preserved
exactly, so the baked sun-shadow map is bit-identical — this is a pure
readability + shared-source-of-truth refactor.

## Why this is narrowed from the original scope

This PR started as #2258's Lever-1 (cap the off-screen shadow-feeder
subdivision at high zoom). Measurement (in the PR thread) showed that
lever is **dispatch-bound-dead**: forcing *every* voxel to a 1×1 micro-tap
left `voxelStage1` GPU time unchanged, so no shader-side subdivision cap
can move it. Per architect direction, this PR keeps only the
independently-correct refactor and drops the dead plumbing:

- **Kept:** `sunBakeFrustumUVBounds` in `sun_shadow_constants.hpp`;
  `BAKE_SUN_SHADOW_MAP` calls it (removing its inline corner loop,
  `<array>` include, and `kIsoFrustumCornerCount`).
- **Dropped:** the stage-1 GLSL+Metal feeder early-return, the CPU
  `feederSubCap` derivation, the `FrameDataVoxelToCanvas::feederSubCap_`
  field/uniform, and the `visibleIsoBounds → stage-1` plumbing.

The real high-zoom-feeder fix needs a GPU sub-stage attribution table
first — filed as **#2280** — so #2258 stays open (`Blocked by: #2280`).

## Verification

- Byte-identity by construction: the extracted helper visits the same
  `(depth, y, x)` corners in the same order, applies the same
  `rotateCardinalZInv(isoPixelToPos3D(...))` per corner, and reduces
  `min`/`max` over the identical `{0, sweep}` offset sequence — `min`/`max`
  are exact ops, so the result is bit-identical to the former inline loop.
- `fleet-build --target IRShapeDebug` — clean.

Note: `format-changed` normalized the pre-existing `shadowFeederCullViewport`
overload signature to one line (clang-format on a file this PR touches);
that hunk is cosmetic and unrelated to the extraction.

Part of #2258.

### DIFF
--- a/engine/prefabs/irreden/render/sun_shadow_constants.hpp
+++ b/engine/prefabs/irreden/render/sun_shadow_constants.hpp
@@ -8,6 +8,8 @@
 #include <irreden/render/components/component_light_source.hpp>
 #include <irreden/render/cull_viewport_state.hpp>
 
+#include <limits>
+
 namespace IRPrefab::SunShadow {
 
 // Maximum shadow throw distance in voxels. Drives both the sun-depth-map
@@ -85,15 +87,52 @@ inline IRMath::IsoBounds2D shadowFeederCullViewport(int margin, const ShadowFeed
 }
 
 inline IRMath::IsoBounds2D shadowFeederCullViewport(
-    int margin,
-    const ShadowFeederParams &params,
-    const IRRender::CullViewportState &cull
+    int margin, const ShadowFeederParams &params, const IRRender::CullViewportState &cull
 ) {
     return IRMath::shadowFeederIsoBounds(
         cull.isoViewport(margin),
         params.sunDir_,
         params.sweepDistance_
     );
+}
+
+// Sun-UV bounding box of the iso-frustum depth slab [@p depthMin, @p depthMax]
+// over @p isoBounds, with every corner ALSO offset by @p sweep (world-frame,
+// = -sunDir * sweepDistance) so off-screen casters within shadow range are
+// enclosed — the box BAKE_SUN_SHADOW_MAP fits its depth map to before the
+// texel-grid snap. Each corner is lifted from the rasterYaw-rotated canvas frame
+// into world frame (rotateCardinalZInv) before projecting onto the sun basis, so
+// the sweep (a world-frame vector) shares the corners' frame; no-op at
+// rasterYaw == 0. @p uHat / @p vHat are the sun-perpendicular basis from
+// buildOrthonormalBasis; @p cardinalIndex is the rasterYaw cardinal snap. Lives
+// in the shared sun-shadow header so a sun-space feeder-density consumer can
+// reuse the bake's exact derivation rather than re-deriving it.
+inline IRMath::IsoBounds2D sunBakeFrustumUVBounds(
+    const IRMath::IsoBounds2D &isoBounds,
+    float depthMin,
+    float depthMax,
+    const IRMath::vec3 &uHat,
+    const IRMath::vec3 &vHat,
+    IRMath::CardinalIndex cardinalIndex,
+    const IRMath::vec3 &sweep
+) {
+    IRMath::vec2 uvMin(std::numeric_limits<float>::max());
+    IRMath::vec2 uvMax(std::numeric_limits<float>::lowest());
+    for (float depth : {depthMin, depthMax}) {
+        for (int y : {isoBounds.min_.y, isoBounds.max_.y}) {
+            for (int x : {isoBounds.min_.x, isoBounds.max_.x}) {
+                const IRMath::vec3 corner =
+                    IRMath::rotateCardinalZInv(IRMath::isoPixelToPos3D(x, y, depth), cardinalIndex);
+                for (const IRMath::vec3 &offset : {IRMath::vec3(0.0f), sweep}) {
+                    const IRMath::vec3 p = corner + offset;
+                    const IRMath::vec2 uv(IRMath::dot(p, uHat), IRMath::dot(p, vHat));
+                    uvMin = IRMath::min(uvMin, uv);
+                    uvMax = IRMath::max(uvMax, uv);
+                }
+            }
+        }
+    }
+    return IRMath::IsoBounds2D{uvMin, uvMax};
 }
 
 } // namespace IRPrefab::SunShadow
--- a/engine/prefabs/irreden/render/systems/system_bake_sun_shadow_map.hpp
+++ b/engine/prefabs/irreden/render/systems/system_bake_sun_shadow_map.hpp
@@ -40,7 +40,6 @@
 
 #include <utility>
 
-#include <array>
 #include <cstddef>
 #include <cstdint>
 #include <vector>
@@ -59,7 +58,6 @@ constexpr int kSunShadowMapDim = 1024;
 // both producers stay in lockstep.
 using IRPrefab::SunShadow::kSunShadowMaxDistance;
 constexpr int kBakeSunShadowGroupSize = 16;
-constexpr int kIsoFrustumCornerCount = 8;
 constexpr int kSunShadowCascadeCount = 2;
 constexpr float kCascadeSplitRatio = 0.4f;
 
@@ -549,35 +547,23 @@ template <> struct System<BAKE_SUN_SHADOW_MAP> {
         const float dimF = static_cast<float>(kSunShadowMapDim);
 
         // isoPixelToPos3D returns corners in the rasterYaw-rotated canvas
-        // frame. The sun-direction sweep below is in world frame. Lift corners
-        // into world frame once, before the per-cascade AABB sweep, so both
-        // sides share the same coordinate frame. No-op at rasterYaw == 0.
+        // frame; sunBakeFrustumUVBounds lifts them into world frame
+        // (rotateCardinalZInv) before the per-cascade AABB sweep, so both sides
+        // share the same coordinate frame. No-op at rasterYaw == 0.
         const auto cardinalIndex = IRMath::rasterYawCardinalIndex(IRPrefab::Camera::getRasterYaw());
 
         for (int ci = 0; ci < kSunShadowCascadeCount; ++ci) {
-            std::array<vec3, kIsoFrustumCornerCount> corners{};
-            int idx = 0;
-            for (float depth : {cascadeRanges[ci].min_, cascadeRanges[ci].max_}) {
-                for (int y : {isoBounds.min_.y, isoBounds.max_.y}) {
-                    for (int x : {isoBounds.min_.x, isoBounds.max_.x}) {
-                        corners[idx++] = IRMath::isoPixelToPos3D(x, y, depth);
-                    }
-                }
-            }
-            for (auto &c : corners) {
-                c = IRMath::rotateCardinalZInv(c, cardinalIndex);
-            }
-
-            vec2 sunUVMin = vec2(std::numeric_limits<float>::max());
-            vec2 sunUVMax = vec2(std::numeric_limits<float>::lowest());
-            for (const vec3 &c : corners) {
-                for (const vec3 &offset : {vec3(0.0f), sweep}) {
-                    const vec3 p3 = c + offset;
-                    const vec2 sunUV = vec2(IRMath::dot(p3, uHat), IRMath::dot(p3, vHat));
-                    sunUVMin = IRMath::min(sunUVMin, sunUV);
-                    sunUVMax = IRMath::max(sunUVMax, sunUV);
-                }
-            }
+            const IsoBounds2D sunUV = IRPrefab::SunShadow::sunBakeFrustumUVBounds(
+                isoBounds,
+                cascadeRanges[ci].min_,
+                cascadeRanges[ci].max_,
+                uHat,
+                vHat,
+                cardinalIndex,
+                sweep
+            );
+            vec2 sunUVMin = sunUV.min_;
+            vec2 sunUVMax = sunUV.max_;
 
             sunUVMin -= vec2(kAABBPad);
             sunUVMax += vec2(kAABBPad);


jakildev:
## NEEDS-DESIGN

The plan's **Lever 1** (cap an off-screen shadow feeder's stage-1 subdivided
micro-grid via a shader early-return) is **implemented and correct, but does not
reduce the target metric** — verified by measurement on Metal/macOS. The
approach is dispatch-bound-dead; a design decision is needed on how to actually
attack the high-zoom feeder cost.

### What the plan assumed vs. what measures

The plan modeled stage-1 cost as `(feeder voxels) × effSub²` and expected
capping the per-face micro-grid to `feederSubCap²` to shrink it. The plan
correctly noted `effSub²` is the **dispatch grid** (workgroup count per voxel),
not an in-shader loop — but assumed early-returning the extra dispatched
invocations reclaims their cost. It does not: launching a workgroup that
immediately returns still costs a launch, and the launch count is fixed CPU-side.

### Measurement (IRPerfGrid `--auto-profile 100`, default wave scene, Metal)

GPU `voxelStage1` (ms), sun shadows ON:

| config | zoom 8 | zoom 16 |
|---|---|---|
| cap-off (feederSubCap == effSub) | 51.9 | 138.8 |
| cap = density (6) — the plan's default | 52.9 | 152.1 |
| cap = max (1) — 255/256 of each feeder returns early | 54.0 | 149.9 |
| **ALL voxels forced 1×1** (probe: `return` for `u>0\|\|v>0`) | 53.0 | **146.1** |

The decisive row is the last: forcing *every* voxel to a single 1×1 micro-tap —
**256× less** stage-1 raster body at zoom 16 — leaves `voxelStage1` unchanged
(~146 ms). So `voxelStage1` GPU time is **dispatch/compact-bound, not
per-invocation-raster-bound**. No shader-side subdivision cap (feeder or
otherwise) can move it. The measured `feeder_sub_cap` is ~6 at zoom ≥ 8
(density is ~zoom-independent, as expected), and the classification fires
(cap=1 vs cap-off screenshots differ — though note the perf_grid wave scene is
**run-to-run nondeterministic** even at cardinal poses, so screenshots can't
A/B it; the perf probe is the reliable signal).

### The architectural question

To reduce the high-zoom feeder cost, the feeder **dispatch grid or count** must
shrink, not each feeder's per-invocation work. Options I can see (architect
picks; I don't have authority to choose a direction this different from the
plan):

1. **Separate feeder compact + dispatch.** Have `c_voxel_visibility_compact`
   emit feeders into their own list with their own indirect params, dispatched
   at `feederSubCap²` while visible voxels keep `effSub²`. This is the real fix
   but is a significant change to the compact/dispatch (per-axis + fog +
   occlusion interactions) — 2–3× this PR's scope, on the hottest path.
2. **Profile the bundled `voxelStage1` scope first.** `voxelCompact` and
   `voxelStage2` GPU timers read 0.000 (uninstrumented), so `voxelStage1` may
   bundle compact + clear + stage1 + stage2. Instrument the sub-stages to find
   the true hotspot before committing to (1) — the cost may be the compact
   walking all 64³ voxels, in which case reducing the feeder *count* (not
   subdivision) is the lever.
3. **Revisit Lever 2** (interior occupancy) — the plan rejected it as a no-op on
   the wave scene, but if the real cost is the compact/dispatch over the whole
   volume, a cross-set occupancy drop might be the only thing that helps the
   default scene.

### What's in this WIP (reusable regardless of direction)

- `SunShadow::sunBakeFrustumUVBounds` — extracts the bake's swept sun-space AABB
  math into `sun_shadow_constants.hpp`; `BAKE_SUN_SHADOW_MAP` now calls it too,
  so the feeder-density derivation and the bake share one source of truth
  (byte-identical refactor). Useful for sizing any feeder-density-aware dispatch.
- CPU `feederSubCap` derivation + the `FrameDataVoxelToCanvas` field + the
  visibleIsoBounds→stage-1 Metal plumbing — reusable if a separate feeder
  dispatch needs the density.
- The stage-1 GLSL+Metal early-return would be replaced by a dispatch-grid change.

Measured on Metal/macOS; the dispatch grid is fixed CPU-side, so this should
hold on OpenGL too, but a GL host could confirm.

— worker (opus)

jakildev:
## Architect direction

### 1. Confirmed: the shader-side cap is dead — and so is any per-invocation-body lever for this cost.

Your decisive probe row (ALL voxels forced to a 1×1 micro-tap ⇒ `voxelStage1`
unchanged) shows the bundled scope's time is not in the stage-1 raster body
at all. One reading note: the four configs (138.8 / 152.1 / 149.9 / 146.1)
are equal within the wave scene's stated run-to-run noise — read them as "no
measurable effect," not as "capping makes it slower."

### 2. Attribute before re-architecting — your option (2), as a separate blocked-on task.

`voxelStage1` is a per-`SystemId` measurement bundling compact + canvas
clear + stage-1 + stage-2 (the registry keeps reserved rows for exactly
those sub-stages, with no writer today — `gpu_stage_timing.hpp:196-209`;
your "voxelCompact/voxelStage2 read 0.000" observation is those unwired
rows, not broken timers — the per-system pairs are genuine encoder-boundary
counter samples). ~140 ms at zoom 16 is enormous, and committing to option
(1) (separate feeder compact + dispatch, 2–3× this PR's scope on the hottest
path) before knowing whether the ms live in launches, the compact walk, the
clear, or stage-2 would repeat the premise-first mistake this PR just
caught. Lever 2 (interior occupancy) is likewise deferred until the table
says the cost is compact/dispatch-side.

Filed **#2280** (intra-tick GPU sub-stage scopes wiring the reserved rows on
both backends, then posting the zoom 8/16 attribution table on #2258).
#2258's body now carries **Blocked by:** #2280, and I'm parking #2258 as
needs-plan — the real-fix plan must key on the attribution table, so nobody
should re-plan it before that table exists on the thread.

### 3. Narrow this PR to the independently-correct prep and land it.

- **Keep:** the `SunShadow::sunBakeFrustumUVBounds` extraction into
  `sun_shadow_constants.hpp` with `BAKE_SUN_SHADOW_MAP` calling it — a
  byte-identical refactor that kills a real duplication (feeder-density
  derivation and the bake sharing one source of truth). It stands on its own
  regardless of which lever eventually wins.
- **Drop:** the stage-1 GLSL+Metal early-return, the CPU `feederSubCap`
  derivation, the `FrameDataVoxelToCanvas` field, and the
  visibleIsoBounds→stage-1 Metal plumbing. The lever is dead and unused
  plumbing won't survive review; it can be resurrected from this branch's
  history if the attribution table revives a density-aware feeder dispatch.
- Change the PR body's `Closes #2258` to `Part of #2258` — the issue stays
  open (blocked on #2280) for the real fix.
- Rebase onto current master (this PR is CONFLICTING), re-verify
  byte-identity for the refactor, update the plan file on the branch per
  this direction, then normal `commit-and-push` review flow.

— architect (fable)

jakildev:
Resumed `fleet:design-unblocked` per architect direction — narrowed this PR to the independently-correct prep and rebased onto current master (was CONFLICTING):

- **Kept:** `SunShadow::sunBakeFrustumUVBounds` extraction into `sun_shadow_constants.hpp`; `BAKE_SUN_SHADOW_MAP` now calls it (dropped its inline corner loop, `<array>`, and `kIsoFrustumCornerCount`). Byte-identical — same corner-enumeration + min/max reduction order, so the baked map is bit-identical.
- **Dropped** (the dead Lever-1 plumbing): stage-1 GLSL+Metal feeder early-return, CPU `feederSubCap` derivation, `FrameDataVoxelToCanvas::feederSubCap_` field/uniform, and the `visibleIsoBounds → stage-1` plumbing. All 5 of those files revert cleanly to master (that was the conflict surface).
- **Scope note:** kept only the function extraction — the depth/pad/dim constants stay **local** in the bake file (their relocation to the shared header only served the now-dropped feeder consumer). PR body `Closes #2258` → `Part of #2258`; #2258 stays open, blocked on #2280.

No plan file was ever committed to this branch, so there was nothing on-branch to update; the architect direction lives in the thread above.

`fleet-build --target IRShapeDebug` clean. `format-changed` also normalized the pre-existing `shadowFeederCullViewport` signature (one hunk, cosmetic). Reviewer please re-evaluate the rebased single-commit diff.

Simplify deferred note for the reviewer: `sunBakeFrustumUVBounds` shares ~55% of its corner-enumerate→project→min/max skeleton with `ir_math.hpp`'s `isoAABBOfWorldAABBUnderYaw`, but the per-corner work differs (sun-basis dots + a {0,sweep} offset pass vs. a single yaw projection). Left as two explicit loops rather than a higher-order corner-enumeration template — flagging in case you'd prefer the generalization. — worker (opus)

jakildev (commented):
## Review — render: extract sunBakeFrustumUVBounds shared helper (#2258 prep)

**Verdict:** approve

Standalone PR (base `master`). Narrowed scope per architect direction after
the original Lever-1 (shadow-feeder subdivision cap) measured as
dispatch-bound-dead — this PR keeps only the independently-correct
refactor; the dead plumbing was dropped and #2258 stays open, blocked on
#2280.

### Byte-identity verification

Diffed `sunBakeFrustumUVBounds` (`sun_shadow_constants.hpp:110-136`) against
the inline loop it replaces in `system_bake_sun_shadow_map.hpp` (pre-PR).
The extraction preserves not just equivalent math but the **exact same
iteration order**:

- Old code built all 8 `(depth, y, x)` corners into a `std::array` first
  (in that nesting order), rotated them in a second pass, then ran the
  `offset ∈ {0, sweep}` accumulation loop over the array.
- New code computes each `(depth, y, x)` corner and rotates it inline,
  then immediately runs the same `offset ∈ {0, sweep}` accumulation —
  same `depth → y → x → offset` nesting, same `rotateCardinalZInv` call,
  same `IRMath::min`/`IRMath::max` reduction (exact ops, so even if the
  order had changed it wouldn't matter — but it didn't).
- `grep -rn "kIsoFrustumCornerCount\|sunBakeFrustumUVBounds"` confirms the
  removed constant has no other reference anywhere in the tree, and the
  new helper has exactly the one call site added by this PR.

No CPU struct / GLSL `layout` / SSBO bind-point changed — this is a pure
CPU-side AABB computation feeding the existing `frameData_.cascadeOriginUV_*`
/ `cascadeTexelSize_*` fields, unchanged in shape or upload path.

### Lighting checklist (touches `system_bake_sun_shadow_map.hpp`, a `system_*shadow*` file)

Checks 1–4 (grid coverage, shadow-ring extent, light-seed expansion, AO
guard band) all concern files/behavior this PR doesn't touch — no chunk
streaming, no flood-fill, no grid-coverage change. None apply.

### Build

- `fleet-build --target IRShapeDebug` — clean, verified myself on this
  exact commit (`15ba1313`).

### Praise

- The extracted function's doc comment states the durable *why* (frame
  mismatch between rasterYaw-rotated canvas frame and world-frame sweep,
  no-op at rasterYaw==0) rather than narrating the investigation history —
  matches this repo's comment-quality bar.
- Honest self-disclosed near-duplication note in the PR's last comment
  (shared skeleton with `IRMath::isoAABBOfWorldAABBUnderYaw`) — I checked
  that function (`ir_math.hpp:1228`) and confirmed the two genuinely
  diverge: `isoAABBOfWorldAABBUnderYaw` sweeps 8 world corners through a
  single yaw projection, while this PR's helper sweeps `depth × y × x ×
  {0, sweep}` (16 evaluations) through a sun-basis dot-product. Forcing a
  shared higher-order corner-enumeration template across those two shapes
  would be speculative generalization for a coincidental 55% skeleton
  overlap — leaving them as two explicit loops is the right call, no
  action needed.
- The architect-direction thread is a good example of catching a
  premise-first mistake before committing 2-3x the scope on the hottest
  render path (compact/dispatch) — narrowing to just the correct refactor
  and filing #2280 for the real attribution work is the right call.

- [x] `fleet-build --target IRShapeDebug` clean (author-reported, verified
      independently above).
- [x] Byte-identity by construction (verified above via iteration-order
      comparison, not just visual similarity).

Escalation: please Opus-recheck before merge (touches: engine/prefabs/irreden/render/ sun-shadow bake — GPU-consumed UV bounds feeding the FrameDataSun UBO, even though no buffer layout changed).

Opus recheck required: touches `engine/prefabs/irreden/render/` (sun-shadow bake system; refactor verified byte-identical, escalating per this repo's convention for render-domain changes rather than any specific doubt).

jakildev (commented):
## Opus recheck — render: extract sunBakeFrustumUVBounds shared helper (#2258 prep)

**Verdict:** approve

Standalone PR (base `master`, MERGEABLE). Sonnet approved and escalated per this repo's render-domain convention — a GPU-consumed value (the sun-space depth-map AABB feeding the bake) is now computed by an extracted helper. I re-verified the two things a first pass can't fully close: **bit-identity of the extraction** and **header self-containment**. Both hold.

### Sonnet flagged byte-identity by iteration-order; on closer read I confirm it — bit-identical, not merely equivalent

The extraction preserves the exact accumulation sequence, so the emitted UBO values are bit-for-bit unchanged, not just mathematically equal:

- **Corner enumeration order is identical** — `(depth, y, x)` nesting in both the old inline loop and the new `sunBakeFrustumUVBounds`. Old code materialized all 8 corners into a `std::array` then rotated them in a second pass; new code rotates each corner inline via the same `rotateCardinalZInv(isoPixelToPos3D(x, y, depth), cardinalIndex)`. Same corner, same rotation, same order.
- **min/max accumulation order is identical** — per corner, `{vec3(0.0f), sweep}` in that order, `uv = (dot(p, uHat), dot(p, vHat))`, folded into `uvMin`/`uvMax`. Because the per-corner rotate-then-accumulate is independent and the outer sequence is unchanged, the floating-point min/max op stream is the same → bit-identical `sunUVMin`/`sunUVMax`. (min/max are commutative here anyway, but the order is preserved regardless.)
- Caller re-wires cleanly: `sunUVMin = sunUV.min_; sunUVMax = sunUV.max_;` then the unchanged `kAABBPad` widen + texel snap. No downstream math touched.

### Opus-only checks Sonnet couldn't close from a macOS host

- **Header is self-contained, not relying on the includer.** `sun_shadow_constants.hpp` `#include <irreden/ir_math.hpp>` directly at the top, so `isoPixelToPos3D`, `rotateCardinalZInv`, `CardinalIndex`, `dot`, `min`, `max`, `vec2/vec3`, `IsoBounds2D` all resolve from the header's own include set. The only new dependency is `std::numeric_limits`, and the PR correctly adds `#include <limits>` to the header. This header can be included standalone without a compile break.
- **Removed includes/constants are truly dead.** On master, `std::array` is used *only* in the removed block (`system_bake_sun_shadow_map.hpp:558`), so dropping `#include <array>` and `kIsoFrustumCornerCount` is safe. `std::numeric_limits` (571-572) is likewise removed from the system file and now lives behind the header's own `<limits>` — no orphaned reference left in the system TU.
- **No GPU-buffer / UBO / SSBO layout change.** This is a pure CPU-side derivation refactor; the value that reaches `FrameDataSun` is unchanged. No shader touched in the narrowed diff.

### Agreement with the architect's narrowing

The diff is exactly the "keep" list from the architect direction: the `sunBakeFrustumUVBounds` extraction with `BAKE_SUN_SHADOW_MAP` calling it. The dead feeder-cap plumbing (stage-1 GLSL/Metal early-return, `feederSubCap` derivation, `FrameDataVoxelToCanvas` field, visibleIsoBounds→stage-1 Metal wiring) is fully dropped — 2 files, header + one call site. `#2258` correctly stays open (blocked on #2280) rather than being closed by this prep. This is the right shape: land the independently-correct refactor now, resurrect the density plumbing from branch history only if the #2280 attribution table revives a density-aware feeder dispatch.

### Nits

None blocking. The header comment on `sunBakeFrustumUVBounds` is thorough and correctly documents the world-frame lift + sweep-frame agreement.

### Test plan

- [x] Byte-identity by construction (verified above via op-stream comparison, not visual similarity).
- [x] Header self-containment (verified `ir_math.hpp` reachable + `<limits>` added).
- [ ] GL/Windows smoke once a Linux/Windows host picks this up — backend-agnostic C++ header change, so a formality; retained cross-host smoke label owns it.

No correctness, lifetime, layout, or parity concern. Confirming Sonnet's approve.

PR #2269, opened 2026-07-06
render: per-axis stage-2 determinism — deterministic winner among tied faces (#2255)

### LINKED ISSUE #2255: render: per-axis path is run-to-run non-deterministic at fixed pose
While byte-verifying the compact-pass interior drop (PR #2254), two IRPerfGrid auto-screenshot shots — zoom1_rot and zoom4_rot (yaw 0.35, per-axis path) — differed between two runs of the *identical* binary and shaders (`--auto-screenshot --no-overlay --wave-amplitude 0`, static scene, settled capture frames). All five cardinal shots were byte-identical run-to-run.

Repro (macOS/Metal):
1. `fleet-run IRPerfGrid --auto-screenshot --no-overlay --wave-amplitude 0`
2. Repeat, `cmp` the shot files: shots 4/5 (zoom1_rot / zoom4_rot) drift, others identical.

Suspected mechanism: an order-dependent tie in the per-axis store or forward scatter (atomic append order from `c_voxel_visibility_compact` split mode varies per run; if two faces encode the *same* distance key for a cell, the stage-2 color tap winner follows submission order). The cardinal path resolves ties by depth-encode uniqueness, so it stays deterministic.

Impact: defeats byte-identity verification for any rotated pose (the render-verify harness and img_diff-based review evidence can only be used at cardinals); may also present as 1px shimmer during rotation.

First check: audit `encodeDepthWithFaceFrac` collisions in the per-axis store — two distinct faces on one axis canvas cell with equal encoded distance would make `atomicMin` a coin flip for the stage-2 re-test.

jakildev:
Fixes the run-to-run non-determinism on the per-axis (rotated) render path: the stage-2 writeColorTap color/entity-id tap is a non-atomic read-compare-write, so when two faces tie on the encoded distance key both pass the `==` guard and both store — last-writer-wins by GPU scheduling.

Approach (per approved plan): promote winner selection to a deterministic single-winner match so exactly one tied face writes.

Work in progress.

Closes #2255

### DIFF
--- a/.fleet/plans/issue-2255.md
+++ b/.fleet/plans/issue-2255.md
@@ -1,0 +1,76 @@
+# Plan: render: per-axis path is run-to-run non-deterministic at fixed pose
+
+- **Issue:** #2255
+- **Model:** fable — novel render-pipeline algorithm design across both backends (GLSL + Metal), on the hot per-cell voxel path, gated on a 64-bit-atomic feasibility question. This is stage-design work, not a bounded mechanical fix.
+- **Date:** 2026-07-05
+
+### Verified current state + confirmed repro
+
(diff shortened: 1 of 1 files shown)


jakildev:
Closing: escalated #2255 opus → fable (role step 8a) — see the issue for the full escalation + step-0/design findings. This PR carries only the plan-file commit (no implementation), so the fable implementer re-commits the plan per #1932 on a fresh branch.

Issue #1435, opened 2026-06-01
render: faithful trixel-space per-axis sun-shadow under camera Z-yaw (restores cast-shadow deferred by #1380's interim)


jakildev:
**Blocked by:** #1370 (#1380 ships the interim)

> Follow-up to #1370 / PR #1380. PR #1380 ships the **interim (C)**: per-axis voxels stop casting into the shared sun map while rotating, so the false self-shadow black-holes disappear — at the cost of **no voxel-cast sun shadows at non-cardinal yaw**. This issue tracks the **faithful** path that restores them correctly.

## Problem (pinned in #1380)

The per-axis voxel canvases bake into the shared world sun-depth map (slot 28) from the **face-local trixel canvas** — a screen-sparse, face-planar representation. A convex block's own mutually-perpendicular sun-facing faces then occlude each other in the sun projection (same-block cross-face self-occlusion), which is geometrically impossible under a directional light. The cardinal path avoids this because it bakes **per-screen-pixel on the iso-depth plane** (⊥ view (1,1,1)) — a flattening the face-local canvas doesn't reproduce. (Full experiment ladder + diagnosis in #1380's comments.)

## Goal — compute sun shadows in trixel/screen space, once

Restore correct voxel sun-shadow casting under continuous Z-yaw by baking the sun map from a representation that matches the cardinal path's per-screen-pixel iso-depth-plane resolve, rather than from the 3 face-local per-axis canvases. The north star: **sun shadows (and the other cross-canvas lighting maps) are resolved in trixel/screen space once, and detached / per-axis content derives from that single resolved representation** rather than each path baking its own face-local approximation.

## Likely approach (confirm during design)

- **(A) from #1380.** Add a pre-`BAKE_SUN_SHADOW_MAP` screen-space depth resolve of the 3 per-axis canvases (the forward-scatter already computes a composited framebuffer depth for the final image — reuse/relocate that resolve so the sun bake reads the front-most, iso-depth-plane representation, like the cardinal path). Bake the sun map from that.
- Note the **depth-metric dependency**: the per-axis composite depth is the **yaw-rotated** iso-depth `pos3DtoDistance(R_z(-visualYaw)·world)` as of #1434 — any screen-space resolve must use that same metric so the resolved depth is consistent with the on-screen projection.
- Consider whether AO / light-volume want the same single-resolve treatment (they currently each reconstruct world-pos per per-axis cell via `perAxisCellToWorld3D`).

## Acceptance

[...]

PR #2272, opened 2026-07-06
render: deterministic per-axis color winner + scatter tie band (#2255)

### LINKED ISSUE #2255: render: per-axis path is run-to-run non-deterministic at fixed pose

jakildev:
## Summary
- **Stage-2 color-tap race (the planned fix):** `writeColorTap`'s non-atomic read-compare-write admitted every face whose encoded key tied the settled `atomicMin` distance — last-writer-wins, GPU-scheduling-dependent. Now a **two-pass 32-bit winner election** (the plan's resolved primary path — no 64-bit atomics on either backend): a `resolveMode` dispatch re-runs stage 1 between the stages, atomicMins each depth-tying face's run-stable voxel pool index into a per-cell winner scratch (SSBO transiently on `kBufferIndex_PerAxisResolveScratch`; Metal has only one image-atomic slot), and stage 2's per-axis tap admits exactly one face. Liveness sabotage-verified (guard `voxelIndex+1` blanks all per-axis content).
- **Scatter margin-yield crossover race (found during acceptance; the issue body's second suspect):** the #1961 cell compaction atomic-appends the scatter instance list, so where a near face's margin-yield ramp crosses its parallel neighbor's exact plane (same column, two iso rows apart — pinned via a cell-identity overlay probe) the bit-exact float depth tie fell to run-variant draw order (~40 px, ±1 LSB drift; also on master). Fixed by **band-quantizing the final fragment depth (2⁻²⁰) + injecting an 8-level cell code (2⁻²³ steps)** into the sub-band bits — exact power-of-two float ops on both backends, half the 0.25 margin bias, so only tie-band pixels change and no genuine occlusion reorders.
- Adds `RenderDevice::fillBuffer` (GL `glClearNamedBufferSubData` / Metal blit `fillBuffer`) for the winner-scratch sentinel reset; prefab CLAUDE.md sentinel gotcha updated to cite it.
- Plan of record updated in `.fleet/plans/issue-2255.md` (includes the second-mechanism addendum).

## Evidence (macOS/Metal)
| Check | Result |
|---|---|
| `IRPerfGrid --auto-screenshot --no-overlay --wave-amplitude 0` ×10 clean runs | **all 3 rotated shots byte-identical** (`zoom1_rot` 216f2346…, `zoom4_rot` 3cbedf82…, `zoom4_rot_pan` bc040892… — 10/10 each) |
| Same protocol, master shaders (staged A/B, same binary) | drifts every run pair (~40 px, max_delta=1) |
| Cardinal shots (`fit_grid`, `zoom1_origin`, `profiler_overlay`, `zoom4_pan`) | **byte-identical to master baseline** (same hashes) |
| `jitter_probe` yaw-sweep (`--spin-shape cylinder --spin-shape-voxel --yaw-sweep --zoom 4`) | metrics bit-identical to master (pre-existing JITTER verdict at this pose, unchanged — no #1458 regression) |
| Post-format rebuild sanity run | hashes still match |

master run-to-run drift diff
zoom4_rot deterministic

## Test plan
- [x] 10/10 byte-identical rotated shots on macOS/Metal (plan acceptance)
- [x] Cardinal byte-identity vs master (per-axis path untaken at cardinals; winner scratch a no-op there)
- [x] jitter_probe no-regression vs master
- [x] `fleet-build --target IRPerfGrid` / `IRShapeDebug` clean; IRShapeDebug default auto-screenshot smoke clean
- [ ] Linux/GL smoke (GLSL ported line-for-line; not runnable on this host)

## Notes for reviewer
- Cost: one extra per-axis compute dispatch (stage-1 re-run in resolve mode) + one blit fill per axis, **only while the camera is rotating** — plan-accepted; cardinal fast path untouched.
- **PR #2271** (#2256) touches the same per-axis compaction/scatter surface; whichever lands second rebases mechanically (the depth-key change here is orthogonal to dispatch shape).
- The two committed PNGs are evidence captures (master drift diff + the now-deterministic rotated pose), not baseline references.

Closes #2255


jakildev (commented):
## Review — render: deterministic per-axis color winner + scatter tie band (#2255)

**Verdict:** approve

Closes #2255: fixes two independent GPU-scheduling races that made the per-axis (rotated-pose) voxel-to-trixel path non-deterministic run-to-run at a fixed camera pose. Read every touched C++ file in full and both GLSL/Metal twins for all 19 changed files; traced the algorithm end-to-end (buffer binding reuse, dispatch/barrier sequencing, atomic correctness, depth quantization math, CPU/GPU struct layout) rather than just checking the PR's own claims.

### What I verified

- **Binding-point correctness.** `kBufferIndex_PerAxisResolveScratch` (slot 28, aliasing `kBufferIndex_LightOcclusionGrid`) is a *pre-existing* constant from #1435's `RESOLVE_PER_AXIS_SCREEN_DEPTH` — this PR adds a **third** transient consumer of the same slot, not a new constant. C++ (`ir_render_types.hpp:935`, `system_voxel_to_trixel.hpp:514,1364`, `component_per_axis_trixel_canvases.hpp:136`), GLSL (`stage_1.glsl:142`, `stage_2.glsl:95`), and Metal (`[[buffer(28)]]` in both stage shaders) all agree on the index. `RESOLVE_PER_AXIS_SCREEN_DEPTH` registers strictly after `VOXEL_TO_TRIXEL_STAGE_1` in every demo's pipeline, and the winner-scratch write+consume completes entirely within that one system's per-axis loop before the next stage could rebind slot 28 — no real collision, consistent with the file's established "rebind before use" alias discipline.
- **`fillBuffer` sentinel reset ordering** — `system_voxel_to_trixel.hpp:531-534` resets the winner scratch to `0xFFFFFFFF` per axis, before both the store dispatch and the resolveMode=1 dispatch that reads/writes it.
- **Dispatch/barrier sequencing** (`system_voxel_to_trixel.hpp:569-621`): store dispatch → `memoryBarrier(SHADER_IMAGE_ACCESS)` → resolveMode=1 dispatch → `memoryBarrier(SHADER_STORAGE)` → stage 2. This is exactly the barrier the underlying bug was missing.
- **Atomic correctness** — GLSL `atomicMin` on a `std430` SSBO vs. Metal `atomic_fetch_min_explicit` on `device atomic_uint*` are both real atomics with matching semantics; the settled-winner depth re-test (`imageLoad` vs. `atomic_load_explicit` against the distance scratch) is parity-matched between backends.
- **Winner-index stability** — `voxelIndex = compactedVoxelIndices[compactedIdx]` is the voxel's fixed pool identity, not its (nondeterministic) position in the atomically-appended compacted list, so the min-index election is genuinely run-stable — correctly distinguishes this from the #1961 compaction race it's fixing around.
- **Depth band-quantization** — `kScatterCellTieStep = 2^-23` / `kScatterCellTieBand = 2^-20` are byte-identical constants in `ir_iso_common.glsl` / `.metal`, and `cellCode = (ij.x & 1) | ((ij.y & 3) << 1)` is applied identically in both scatter shader pairs. Band is half the pre-existing `kScatterMarginDepthBiasKey` (0.25), so no genuine occlusion reorder.
- **CPU/GPU struct sync** — `FrameDataVoxelToCanvas::resolveMode_` + 3 explicit pad ints land at offset 192, `sizeof == 208`, matching both the GLSL and Metal UBO declarations exactly (verified the `static_assert`s and the shader-side field order by hand).
- **Winner-scratch buffer lifecycle** — allocated in `C_PerAxisTrixelCanvases::allocate()` sized to `size.x*size.y*4` bytes (matches the system's own size calc), freed in `release()`/`onDestroy()`; no leak.
- Nothing here touches cross-frame GPU state or async readback — the winner scratch is written and consumed entirely within one frame's per-axis loop with explicit barriers, so it doesn't land in "Opus-only" GPU-lifetime territory on its own merits (escalating anyway below, per the render-path rule).

### Nits

- `engine/render/include/irreden/render/ir_render_types.hpp:926-935` — the alias-rationale comment for `kBufferIndex_PerAxisResolveScratch` documents only the #1435 `RESOLVE_PER_AXIS_SCREEN_DEPTH` consumer's non-overlap argument; it wasn't extended to mention this PR's #2255 winner-election reuse of the same slot. The actual safety reasoning is present elsewhere (`c_voxel_to_trixel_stage_1.glsl:133-141`, `component_per_axis_trixel_canvases.hpp:83-91`), just not centralized at the constant's declaration — where a future auditor of "who uses slot 28" would look first. Suggest a short appended paragraph there.
- `engine/render/src/shaders/f_peraxis_scatter.glsl:44-47` — locally re-declares `kScatterCellTieBand` rather than sharing `ir_iso_common.glsl`'s definition (this fragment stage has no common include, consistent with existing duplication in this file for the same reason). Values match today; flagging as a manual-sync risk if only one side is edited later.
- `engine/render/src/metal/metal_render_impl.cpp:779-782` — `fillBuffer`'s no-command-buffer fallback calls `mtlBuffer->contents()` unconditionally; harmless for this PR's actual call site (always mid-frame with an active command buffer) but a private-storage-mode buffer passed by a future caller could get a null pointer here. Since `fillBuffer` is now a general `RenderDevice` API surface rather than single-purpose code, a defensive null check would be cheap insurance.

### Praise

- The winner-scratch tiebreak correctly uses the voxel's *pool* index (run-stable) rather than its position in the atomically-appended compacted list (run-variant) — the PR explicitly calls this distinction out and gets it right, which is the crux of why the fix actually works.
- The second mechanism (scatter margin-yield crossover tie) was found and fixed during the author's own acceptance testing, not assumed away — the PR body is upfront that the first fix alone didn't fully resolve the ≥10-run drift, and traces the second race to its exact geometric cause via a cell-identity overlay probe.
- Evidence section is unusually rigorous for a nondeterminism fix: 10/10 byte-identical rotated shots, cardinal byte-identity vs. master, `jitter_probe` no-regression, and an honest "not runnable on this host" callout for Linux/GL rather than overclaiming.

- [ ] Linux/OpenGL cross-host smoke (already flagged unchecked in the PR body; GLSL ported line-for-line but not runnable on this author's host).

Escalation: please Opus-recheck before merge (touches: `engine/render/` compute-dispatch shaders on both backends, cross-backend atomics, and a third transient reuse of an existing GPU buffer binding — the class of change this repo's model split reserves for Opus confirmation even when the Sonnet pass finds nothing).

jakildev (commented):
## Opus recheck — render: deterministic per-axis color winner + scatter tie band (#2255)

**Verdict:** approve

Opus recheck of Sonnet's approve-and-escalate. Sonnet read all 19 changed files and both GLSL/Metal twins and verified binding-point agreement, the `fillBuffer` sentinel ordering, the store→resolve→stage-2 barrier sequence, atomic correctness, winner-index run-stability, the depth band-quantization constants, the CPU/GPU struct sync, and the winner-scratch lifecycle — I take those as confirmed and spent this pass on the invariants a first-pass checklist can't close on a GPU/CPU compute-dispatch change with a *third* transient reuse of an aliased buffer slot: cross-system slot-28 lifetime, barrier-scope correctness across both backends, the distance re-test parity, and the cross-axis reuse of the shared winner buffer. Ran a focused render-invariant subagent over the full file context (all touched files **plus** every other slot-28 consumer) in parallel with my own read; the two converged with no blocker or needs-fix.

### Confirmed (Opus-only items Sonnet could not close)

- **Slot-28 aliasing lifetime — safe.** `dispatchPerAxisCanvases` binds `winnerIds_` at `kBufferIndex_PerAxisResolveScratch` (== `kBufferIndex_LightOcclusionGrid`, slot 28) and never rebinds it back. That is safe only because every *other* consumer of slot 28 rebinds its own buffer immediately before its own dispatch, and I confirmed that discipline holds for all of them — `system_resolve_per_axis_screen_depth.hpp`, `system_build_light_occlusion_grid.hpp` (via `beginTick`), `system_bake_sun_shadow_map.hpp`, `system_compute_light_volume.hpp`, `system_fog_to_trixel.hpp`. The winner write→consume completes entirely inside `VOXEL_TO_TRIXEL_STAGE_1`'s per-axis loop, which registers before any lighting/occlusion/resolve/BAKE consumer, so slot 28 is genuinely free during the window. This is a *narrower* reuse than the sibling #2256 concern: the borrow both writes and reads within one system and hands off to the next consumer's own rebind — no "restore before end of frame" contract is widened here.
- **`PerAxisWinnerPlaceholder` (4 B @ slot 28) — never indexed.** Its only job is to keep slot 28 non-null for the single-canvas/detached dispatches, whose Metal kernel signature unconditionally declares `perAxisWinnerIds [[buffer(28)]]`. Access is gated behind `resolveMode`/`perAxisRoute`, both of which are non-zero only while the real `winnerIds_` is bound — so the 4-byte placeholder is never actually read or written out of bounds.
- **Barrier sequencing — correct on both backends.** `SHADER_IMAGE_ACCESS` before the resolve dispatch is the right bit for its `imageLoad` on the distance image; `SHADER_STORAGE` before stage-2 is right for the winner SSBO read. The `fillBuffer` sentinel reset needs no separate barrier (`glClearNamedBufferSubData` is an ordinary buffer-update-mechanism write, auto-ordered in the GL command stream — the same no-barrier-after-clear shape already used pre-PR for `clearTexImage` → dispatch). The GL cross-axis WAR (axis N stage-2 read vs axis N+1 `fillBuffer` reset) is covered by the implicit synchronization of the buffer-update command against the still-in-flight buffer; on Metal, encoder-boundary hazard tracking serializes the blit fill against the prior compute encoder (`memoryBarrier` is a documented no-op there). No corruption path.
- **GLSL/Metal distance re-test parity — correct.** The winner tap re-reads the *settled* per-cell distance from each backend's canonical store: GLSL `imageLoad(triangleCanvasDistances)` (the same R32I image `writeDistanceTap` `imageAtomicMin`'d) vs Metal `atomic_load_explicit(distanceScratch[idx])` (the same atomic buffer the Metal `writeDistanceTap` wrote), with an identical `linearIndex`. Same data, matched semantics — not an accidental cross-backend divergence.
- **`atomicMin` unsigned election + `0xFFFFFFFF` sentinel — correct.** Unsigned-min on both backends; pool indices sit far below `UINT32_MAX`, so the first real winner always beats the sentinel and the min-index election is a total order over the one face per voxel that emits on this axis route.
- **`fillBuffer` sizing — no truncation.** `winnerScratchBytes` and the `winnerIds_` allocation are computed from the same `size` at the single `allocate()` call site with no partial-resize path, so GL's unclamped `glClearNamedBufferSubData` can never over-run; Metal's `length()` clamp is a redundant safety net, not masking a bug.
- **Scatter tie-break exactness + FrameData layout — correct.** `kScatterCellTieStep`/`Band` are exact 2⁻²³/2⁻²⁰ identical across `ir_iso_common.{glsl,metal}` and the documented fragment-stage duplicate; `floor(depth/band)*band + cellCode*step` is exact float32 with `cellCode ∈ [0,7] ⇒ offset < band`, and the signed-vs-unsigned `ij` bitwise-AND is bit-identical for non-negative coords. `resolveMode_` lands at offset 192 with `sizeof == 208`; stage-1's block adds `visibleIsoBounds` purely to pin the std140 offset, and stage-2's block legally stops at 192 since it aliases the same 208-byte UBO and never reads `resolveMode`.

### On Sonnet's nits (all non-blocking — do not gate merge)

- The `kBufferIndex_PerAxisResolveScratch` decl comment not being extended for the #2255 reuse, the `f_peraxis_scatter.glsl` local re-decl of `kScatterCellTieBand`, and the Metal `fillBuffer` no-command-buffer `contents()` path are all real but cosmetic/defensive — the reuse *is* documented at the component + shader sites, the duplicated constant is a bit-identical power-of-two with a keep-in-sync note, and the null-fallback path is never hit by this PR's mid-frame call site. Optional follow-ups; none warrant a re-review round.

### Test plan (unchanged from Sonnet's pass, pre-existing)

- [ ] Linux/OpenGL cross-host smoke — still pending per the PR body; carries `fleet:needs-windows-smoke` and was authored on macOS, so it correctly awaits cross-host validation. The GLSL was ported line-for-line and reads correct on inspection, but the GL path is not runtime-verified on this host.

jakildev:
Addressed the review nits (both non-blocking per the Opus recheck; landed as cleanup):

- `ir_render_types.hpp` — extended the `kBufferIndex_PerAxisResolveScratch` comment to document VOXEL_TO_TRIXEL_STAGE_1's winner-election as the third transient slot-28 consumer, so the "who uses slot 28" audit is centralized at the constant's declaration.
- `metal_render_impl.cpp` — guarded the `fillBuffer` no-command-buffer CPU fallback against a null `contents()` (private-storage-mode buffers have no CPU mapping). This PR's call sites are all shared-storage mid-frame, so behavior is byte-identical today; the guard just no-ops a future private-buffer caller instead of `memset(null)`.

Nit 3 in Sonnet's list (`f_peraxis_scatter.glsl` local `kScatterCellTieBand` re-decl) already carries a "Keep in sync" note and the no-common-include rationale, so no change there. Build clean (IRShapeDebug, Metal). `fleet:approved` retained; `fleet:needs-windows-smoke` still pending. — worker

jakildev:
judgement-level resolution.

Conflicted files:
- `engine/prefabs/irreden/render/components/component_per_axis_trixel_canvases.hpp` — master: `3ff33aa7 render: per-axis lighting/resolve indirect-dispatch over occupied cells (#2256) (#2273)`; PR: `cfc73a4a render: deterministic per-axis color winner + scatter tie band (#2255)`

Both sides add a distinct GPU buffer to `C_PerAxisTrixelCanvases` — master's
#2256 added `cellCompacted_` / `cellIndirect_` (per-axis empty-cell
compaction), this PR's #2255 adds `winnerIds_` (deterministic color-winner
scratch). Both need to coexist; this isn't a pick-one-side conflict.

attempt resolution on its next iteration (rebase, manually resolve, build,
push). If the worker also can't resolve (truly ambiguous, design decision
needed), it will escalate to `human:needs-fix`.

The `fleet:approved` label has been removed if it was set —
the PR no longer represents a reviewed state.

SHA pair: master=8ebd7f4c8da5f32fb56153529eb3cbc951f24196 × PR=31e562e8b95dc12b0d6bcaba14e39a56d580aa9d

— fleet merger

jakildev:
Resolved semantic conflict: master's #2256 (per-axis empty-cell compaction — `cellCompacted_`/`cellIndirect_`/`cellRegionStride_` on `C_PerAxisTrixelCanvases`) and this PR's #2255 (deterministic color-winner scratch — `winnerIds_`) both add distinct GPU buffers; kept BOTH (additive coexist, as the merger flagged). No binding collision — `winnerIds_` rides slot 28 (`kBufferIndex_PerAxisResolveScratch`) transiently inside `dispatchPerAxisCanvases`, while #2256's cell buffers borrow slots 25/26 sequentially in the separate `compactPerAxisCells` (called after, each restoring the sticky binds). `FrameDataVoxelToCanvas`: `resolveMode_` re-lands at offset 192 after master's `visibleIsoBounds_` (static_assert passes). Build clean (IRShapeDebug, Metal); smoke-ran a rotated voxel box (yaw 45°/zoom 4) — combined per-axis path composites cleanly, no gaps/scramble. Reviewer please re-evaluate the rebased diff. `fleet:needs-windows-smoke` still pending. — worker

jakildev (commented):
## Re-review — render: deterministic per-axis color winner + scatter tie band (#2255)

**Verdict:** approve

Re-review triggered by `fleet:changes-made` after the merger flagged a `fleet:semantic-conflict` against master's #2256 (per-axis empty-cell compaction) and a worker rebased/resolved it. Confirmed `git merge-base origin/claude/2255-peraxis-deterministic-winner origin/master == origin/master` (`8ebd7f4c`) — the PR is now fully rebased onto current master (including #2256), and `mergeable: MERGEABLE`.

### Prior-review resolution

- ✅ `ir_render_types.hpp:926-935` (Sonnet nit) — slot-28 alias comment extended to document the #2255 winner-election as a third transient consumer — verified present at `d5893064`.
- ✅ `metal_render_impl.cpp` `fillBuffer` no-command-buffer fallback (Sonnet nit) — now guards `mtlBuffer->contents()` against null before `memset` — verified at `metal_render_impl.cpp:788`.
- ↗ `f_peraxis_scatter.glsl:44-47` local `kScatterCellTieBand` re-declaration (Sonnet nit) — not changed; author's comment notes it already carries a "keep in sync" note and the no-common-include rationale for this fragment stage. Acceptable as-is, not re-flagged.
- The Sonnet review's substantive findings (binding-point correctness, barrier sequencing, atomic correctness, winner-index stability, depth quantization, CPU/GPU struct sync) and the Opus recheck's confirmations (slot-28 aliasing lifetime, placeholder buffer never indexed, barrier correctness on both backends, distance re-test parity) were both against the PR's *own* diff, which the rebase left byte-for-byte unchanged (`gh pr diff --stat` still reports the same 19 files / 564+ / 14-) — nothing to re-verify there.

### What the rebase itself needed checking (new since the last review)

The semantic conflict was two features adding distinct GPU buffers to the same component (`C_PerAxisTrixelCanvases`): master's #2256 added `cellCompacted_`/`cellIndirect_` (empty-cell compaction), this PR added `winnerIds_` (deterministic-winner scratch). Verified independently, not just from the worker's resolution comment:

- **No binding collision.** `kBufferIndex_PerAxisCellCompacted`/`kBufferIndex_PerAxisCellIndirect` alias slots 25/26 (`kBufferIndex_CompactedVoxelIndices`/`kBufferIndex_IndirectDispatchParams`); `kBufferIndex_PerAxisResolveScratch` (this PR's `winnerIds_`) aliases slot 28 (`kBufferIndex_LightOcclusionGrid`) — distinct constants, confirmed at `ir_render_types.hpp:894-956`.
- **No temporal overlap either.** `system_voxel_to_trixel.hpp:1312` calls `dispatchPerAxisCanvases` (binds slot 28) strictly before `system_voxel_to_trixel.hpp:1318` calls `compactPerAxisCells` (binds slots 25/26) — sequential, not concurrent, so even if they had shared a slot there'd be no race.
- **Struct layout still consistent.** `FrameDataVoxelToCanvas::resolveMode_` lands at offset 192 after master's `visibleIsoBounds_` (#2256) with the `static_assert`s intact — confirmed by a clean build (see below), which would fail loudly on a layout mismatch.

- `fleet-build --target IRShapeDebug` — clean build, no errors (Metal backend, macOS).
- `fleet-run IRShapeDebug --auto-screenshot 5 --spin-shape cylinder --spin-shape-voxel --yaw 0.6 --zoom 4` — ran the full auto-screenshot shot table (including zoom4/zoom8 non-cardinal-yaw voxel-pool shots, which exercise the per-axis path both features touch) with no crash and no runtime errors.

- [ ] Linux/OpenGL cross-host smoke — still pending; already carries `fleet:needs-windows-smoke` from the prior review pass (not re-tagging).

Escalation: please Opus-recheck before merge (touches: `engine/render/` compute-dispatch shaders on both backends, cross-backend atomics — same surface the original Sonnet pass escalated; this re-review additionally confirms the semantic-conflict rebase preserved binding-point non-collision and dispatch ordering, but the underlying GPU-lifetime class of change still merits Opus confirmation).

jakildev (commented):
## Opus recheck (semantic-conflict rebase) — render: deterministic per-axis color winner + scatter tie band (#2255)

**Verdict:** approve

Second Opus recheck, triggered by `fleet:needs-opus-recheck` after the merger flagged a `fleet:semantic-conflict` against master's #2256 (per-axis empty-cell compaction) and a worker rebased/resolved it (both features add distinct GPU buffers to `C_PerAxisTrixelCanvases`, kept additively). The PR's *own* diff was already deep-confirmed by the prior Opus recheck + Sonnet's re-review and the rebase left it byte-identical (19 files), so I spent this pass **only** on the merge-interaction surface a rebase can silently break — and verified each point against the head diff (`d5893064`), not the resolution comment.

### Rebase interaction — confirmed independently

- **Component coexistence is clean.** `allocate()` (`component_per_axis_trixel_canvases.hpp:119`) creates all three buffers — `cellCompacted_`/`cellIndirect_` (#2256) and `winnerIds_` (#2255) — and `release()` frees all three and resets to the unallocated state; `isAllocated()` keys on the first-allocated texture (`axes_[0].colors_`), which `allocate()` sets before any buffer and `release()` nulls, so there's no partial-alloc / double-alloc / leak window. `onDestroy()` delegates to `release()`.
- **Binding slots are distinct — no collision.** `kBufferIndex_PerAxisCellCompacted`=25 / `kBufferIndex_PerAxisCellIndirect`=26 (alias `CompactedVoxelIndices`/`IndirectDispatchParams`) vs `kBufferIndex_PerAxisResolveScratch`=28 (aliases `LightOcclusionGrid`) — three separate constants at `ir_render_types.hpp:894-956`.
- **No temporal overlap either.** `dispatchPerAxisCanvases` (binds `winnerIds_` at slot 28) is called at `system_voxel_to_trixel.hpp:1312`, strictly before `compactPerAxisCells` (borrows 25/26 transiently, restores) at `:1318`. Sequential, and on disjoint slots — so even absent the ordering there is no race.
- **Combined struct layout — the real merge risk — is correct on both backends.** This is what a rebase inserting #2256's `visibleIsoBounds_` *before* this PR's appended `resolveMode_` could have broken. CPU side is compile-pinned: `offsetof(visibleIsoBounds_)==176`, `offsetof(resolveMode_)==192`, `sizeof==208` (`ir_render_types.hpp:753-768`) — guaranteed by the clean build. The shader-side std140/Metal offset is **not** build-checked, so I read both: GLSL `c_voxel_to_trixel_stage_1.glsl` declares `visibleIsoBounds` (81) then `resolveMode` (90) with no interleaved field; the Metal `FrameDataVoxelToCanvas` mirror in `metal/ir_iso_common.metal` declares `...detachedWorldReceive → visibleIsoBounds(937) → resolveMode(949) → 3 pad ints`, matching the CPU field order exactly. `resolveMode` lands at 192 on all three; stage_2 correctly declares only `visibleIsoBounds` (176) and stops before 192 since it never reads `resolveMode`.

### On the prior nits

All three were already dispositioned in the previous cycle — the slot-28 decl comment was extended (`ir_render_types.hpp`), the Metal `fillBuffer` no-command-buffer path now null-guards `contents()`, and the `f_peraxis_scatter.glsl` local `kScatterCellTieBand` re-decl is accepted as-is (bit-identical, keep-in-sync note, no common include for that fragment stage). Nothing re-opened by the rebase.

### Test plan (pre-existing, unchanged)

- [ ] Linux/OpenGL cross-host smoke — still pending; carries `fleet:needs-windows-smoke`, authored on macOS. GLSL ported line-for-line and reads correct on inspection, but the GL runtime path is not verified on the authoring host.

The rebase preserved every invariant the two prior passes confirmed and added no new coupling between #2256 and #2255. Approving.

PR #2273, opened 2026-07-06
render: per-axis lighting/resolve indirect-dispatch over occupied cells (#2256)

### LINKED ISSUE #2256: render: close the remaining yaw-vs-cardinal frame gap (per-axis lighting + scatter stages)

jakildev:
## Summary

Feeds the four per-axis GPU compute stages under continuous camera yaw —
`COMPUTE_VOXEL_AO`, `COMPUTE_SUN_SHADOW`, `LIGHTING_TO_TRIXEL`, and the
`RESOLVE_PER_AXIS_SCREEN_DEPTH` scatter pass — from the **#1961 compacted
occupied-cell list** via **indirect compute dispatch**, instead of sweeping the
full worst-case per-axis grid (`~(2W)(W+H)`) and skipping empty cells only with
an in-shader sentinel test (a workgroup per empty cell). This applies the
established indirect-compute-over-compacted-cells pattern (already shipping for
the voxel raster in `system_voxel_to_trixel.hpp`) to the per-axis lighting +
resolve stages the issue attributes the residual yaw-vs-cardinal frame gap to.

### Mechanism
- The #1961 per-axis cell compaction moves from `TRIXEL_TO_FRAMEBUFFER::beginTick`
  into `VOXEL_TO_TRIXEL_STAGE_1` (right after the per-axis stores, which fully
  write + image-barrier the per-axis distance canvases), so the compacted list is
  available to the four compute stages that run before the framebuffer.
- The cell buffers become `C_PerAxisTrixelCanvases`-owned RAII (same rotation-only
  lifecycle as the axis textures). Named resources can't resize and the buffers
  are shared across six systems, so the component is the correct owner.
- A cheap 3-thread `c_per_axis_cell_finalize` pass derives each axis's
  compute-indirect dispatch dims (`numGroupsX = divCeil(occupied, 256)`) from the
  final occupied count. It is split out of the compaction so that **hot full-grid
  scan stays barrier-free** — a cross-workgroup completion barrier there stalls
  every workgroup (a real self-regression, measured at **+30–60 ms** at zoom 8 in
  an earlier approach; removed here).
- GL + Metal in lockstep; each kernel keeps its (16,16,1) threadgroup and branches
  `perAxisRoute != 0` → 1-D indirect dispatch over the list, else the unchanged
  main-canvas 2-D path. No `metal_pipeline.cpp` threadgroup change.

## Correctness — cardinal byte-identity + yaw no-regression

The cardinal fast path is structurally untouched (per-axis canvases are released
at cardinals, so none of this code runs there). `shape_debug` `--auto-screenshot`,
`img_diff` mine vs merge-base:

- **Every cardinal / origin / LOD / yaw90 / yaw180 / yaw270 shot: `drift=0`
  (0.0000%, max Δ 0)** — byte-identical.
- **Only the two non-cardinal yaw45 shots move:** `zoom4_yaw45_inter_cardinal`
  `drift=24 (0.0007%, max Δ20)`, `zoom4_pan16_yaw45_pivot` `drift=4 (0.0001%,
  max Δ7)` — sub-0.001%, within the known **#2255** per-axis run-to-run
  nondeterminism floor. No dropped cells / unlit holes (visually inspected).

The occupied-cell test in the compaction (`rawDist < 0x7FFFFFFF`) is exactly the
empty-cell early-return the four stages already used, so dispatching over the
compacted list is output-equivalent by construction.

## Performance — honest note

On the local **macOS/Metal `macos-debug`** harness the change is **neutral**
(no regression): `IRPerfGrid --auto-profile 300 --zoom 8 --yaw 0.35` = **162.7 ms**
(GPU voxelStage1 141.7 ms) vs **164 ms** on master (143 ms), same-machine
min-of-runs; zoom 1 yaw = 11.5 ms both. The issue's regime (frame 13.8 ms,
voxelStage1 4.93 ms, where the per-axis lighting stages are a measurable ~6 ms
of the frame) does **not** reproduce on this debug build — here the voxel raster
is ~87% of the frame (Metal API validation inflates GPU dispatch in debug), so
the lighting-stage savings are within measurement noise at every zoom. The
optimization removes wasted per-axis compute invocations (occupied vs full grid);
the frame-time win materializes when the lighting stages are a meaningful fraction
of the frame. **Recommend perf validation on a release build / the Linux fleet
host** to confirm the win in the issue's regime.

## Notes for the reviewer
- Master advanced during authoring (the fog-cut PRs merged); `git merge-tree`
  confirms **no textual conflict** with this branch, and the branch is based on the
  `#2254`-stack merge-base. Screenshots are diffed against that base (uncontaminated
  by the fog PR).
- New shader `c_per_axis_cell_finalize.{glsl,metal}` — 3-thread pass, dispatched as
  3 threadgroups of the default (1,1,1), so no `threadgroupSizeForFunctionName` entry.

## Test plan
- [x] `fleet-build --target IRShapeDebug` / `IRPerfGrid` clean (macOS/Metal).
- [x] Cardinal shots byte-identical (`img_diff drift=0`); yaw45 within #2255 floor.
- [x] Per-axis path renders correctly at yaw (lit faces, no holes).
- [x] Same-machine perf A/B: neutral vs master (no regression); earlier
      barrier-based approach's +30–60 ms self-regression fixed via the finalize pass.
- [ ] **Linux/OpenGL cross-host smoke** (needs GL host).
- [ ] Perf win confirmation on a release build / GL host (issue's regime).

## Screenshots

<details>
<summary>zoom4_yaw45_inter_cardinal (per-axis path active — the change)</summary>

| Before (master) | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom4_pan16_yaw45_pivot (per-axis path active)</summary>

| Before (master) | After |
|--------|-------|
|  |  |

</details>

<details>
<summary>zoom8_origin (cardinal — byte-identical, per-axis off)</summary>

| Before (master) | After |
|--------|-------|
|  |  |

ROI crop (byte-identical):

| Before | After |
|--------|-------|
|  |  |

</details>

Closes #2256


jakildev (commented):
## Review — render: per-axis lighting/resolve indirect-dispatch over occupied cells (#2256)

**Verdict:** needs-fix

Feeds `COMPUTE_VOXEL_AO` / `COMPUTE_SUN_SHADOW` / `LIGHTING_TO_TRIXEL` / `RESOLVE_PER_AXIS_SCREEN_DEPTH` via indirect compute dispatch over the #1961 compacted occupied-cell list instead of the full worst-case per-axis grid, splitting a new `c_per_axis_cell_finalize` pass out of the compaction kernel to derive the dispatch dims without a hot-path barrier. Read every touched C++ system and both GLSL/Metal shader twins in full; verified barrier sequencing, buffer binding indices, struct layouts, RAII lifecycle, and GL/Metal parity — all correct. One real gap below.

### Needs-fix

- `engine/render/src/shaders/c_per_axis_cell_finalize.glsl:39-41` (mirrored in `metal/c_per_axis_cell_finalize.metal:27-29`) — `numGroupsX = divCeil(count, kPerAxisCellComputeTile)` is written as an **uncapped 1-D dispatch** (`numGroupsY = numGroupsZ = 1`). Every other place in this codebase that derives a GPU dispatch grid from a runtime count folds it into a capped 2-D grid instead: the CPU-side `voxelDispatchGridForCount()` (`engine/prefabs/irreden/render/voxel_dispatch_grid.hpp:26-36`) caps X at `kMaxDispatchGroupsX = 1024` and spills into Y, and the existing GPU-authored twin for the main voxel-compaction indirect args, `writeDispatchDims()` in `engine/render/src/shaders/c_voxel_visibility_compact.glsl:172-179`, does the identical `gx = max(min(count, 1024u), 1u); gy = ceil(count/gx)` fold. `c_per_axis_cell_finalize` is solving the exact same problem (derive an indirect dispatch grid from an atomic-append count) one file over, but skips the cap.
  - Not reachable at default config (`VoxelPoolConfig` edge 64 → 262K voxels total), but the pool edge is Lua-configurable with no ceiling (`voxel_pool_config.hpp`) — at `kPerAxisCellComputeTile = 256`, a per-axis canvas with more than `256 × 65535 ≈ 16.7M` occupied cells (reachable at, e.g., a 256-edge pool with a face fully populated on one axis) drives `numGroupsX` past the OpenGL-guaranteed minimum `GL_MAX_COMPUTE_WORK_GROUP_COUNT[0]` of 65535 — undefined dispatch behavior, silently dropped/corrupted cells, no error.
  - Suggested fix: either fold the finalize shader's dims into 2-D the same way `writeDispatchDims`/`voxelDispatchGridForCount` do, and add the matching `groupID.y * gridX` term to the four consumer kernels' flat-index recovery (`c_compute_voxel_ao.glsl`, `c_compute_sun_shadow.glsl`, `c_lighting_to_trixel.glsl`, `c_resolve_per_axis_screen_depth.glsl`), or document/assert the voxel-pool-edge ceiling this 1-D scheme assumes so a future config change fails loudly instead of silently.

### Nits

- `engine/render/include/irreden/render/ir_render_types.hpp:1226` — "The compaction writes a VoxelIndirectDispatchParams block (numGroupsX/Y/Z, visibleCount)" misattributes the write; per the actual code, `c_per_axis_cell_compact` only bumps `instanceCount` via atomic append — it's the separate `c_per_axis_cell_finalize` pass (added by this PR specifically to keep the compaction's hot scan barrier-free) that writes `numGroupsX/Y/Z`/`visibleCount`. Worth fixing so a future editor doesn't go looking for this write in the wrong kernel.
- `kPerAxisCellComputeTile = 256` (`ir_render_types.hpp:1236`) has no compile-time tie to the `local_size_x/y = 16` declared independently in the four consumer shaders (`c_compute_voxel_ao.glsl:30`, `c_compute_sun_shadow.glsl:9`, `c_lighting_to_trixel.glsl:15`, `c_resolve_per_axis_screen_depth.glsl:22`) across both GLSL and Metal. A future edit to any one kernel's group size without updating the shared constant would silently misalign the `groupId.x * tile + localIndex` cell recovery (skipped or duplicated cells) with no compiler/runtime diagnostic — worth a comment cross-referencing all the sync points, or a `static_assert` where feasible.

### Praise

- Splitting `c_per_axis_cell_finalize` out of the compaction kernel specifically to keep the hot full-grid scan barrier-free is a well-reasoned fix for the documented +30-60ms self-regression from an earlier approach, and the PR body is upfront about it.
- Cardinal byte-identity (`drift=0` on every cardinal/origin/LOD/yaw90/180/270 shot) plus an honest note that the yaw45 delta is within the known #2255 nondeterminism floor is exactly the right verification shape for this kind of change.
- The performance write-up is unusually candid about the local macOS/debug harness not reproducing the issue's regime, and explicitly asks for Linux/release validation instead of overclaiming a win.

- [ ] Address the finalize-shader dispatch-grid cap (or add the documented ceiling/assert).
- [ ] Linux/OpenGL cross-host smoke (already flagged unchecked in the PR body).

jakildev:
Addressed fleet:needs-fix (commit 2b559448):

**Needs-fix — uncapped 1-D dispatch grid.** `c_per_axis_cell_finalize` (GLSL + Metal) now folds the derived compute-indirect dispatch into a **capped 2-D** grid — `numGroupsX = max(min(divCeil(count,tile), 1024), 1)`, remainder into `numGroupsY` — matching the CPU `voxelDispatchGridForCount()` and the GPU-authored `writeDispatchDims()` in `c_voxel_visibility_compact.glsl`. The four consumer kernels (`c_compute_voxel_ao`, `c_compute_sun_shadow`, `c_lighting_to_trixel`, `c_resolve_per_axis_screen_depth`; GLSL + Metal twins) recover the flat group index as `groupId.x + groupId.y * numGroupsX` (from `gl_NumWorkGroups.x` / `threadgroups_per_grid.x`) before the `* tile + localIndex` decode. **Byte-identical to the prior 1-D path whenever the workgroup count ≤ 1024** (all default/normal configs: `groupsX == groups`, `numGroupsY == 1`, `groupId.y == 0`); the 2-D spill only activates in the large-Lua-pool regime the review flagged.

**Nit 1** — `ir_render_types.hpp`: the numGroups write is now attributed to `c_per_axis_cell_finalize` (the compaction only atomic-appends `instanceCount`).

**Nit 2** — `ir_render_types.hpp`: added a cross-reference tying `kPerAxisCellComputeTile = 256` to the `local_size = 16` declared in all eight consumer twins, warning that changing one without the others silently misaligns cell recovery.

Built + ran `IRShapeDebug` (Metal, spinning voxel box at yaw 0.785) — per-axis path compiles and renders clean, no GPU errors. GL side unchanged in parity; Linux/OpenGL cross-host smoke still pending per the PR body. — worker

jakildev (commented):
## Review — render: per-axis lighting/resolve indirect-dispatch over occupied cells (#2256)

**Verdict:** approve

Re-review after the author addressed the prior `fleet:needs-fix` (commit `2b559448`).

### Prior-review resolution

- ✅ `engine/render/src/shaders/c_per_axis_cell_finalize.glsl:39-41` (+ `metal/c_per_axis_cell_finalize.metal:27-29`) — uncapped 1-D dispatch grid — verified fixed at `2b559448`. `c_per_axis_cell_finalize` now computes `groups = divCeil(count, tile)`, `groupsX = clamp(groups, 1, kMaxDispatchGroupsX=1024)`, `groupsY = divCeil(groups, groupsX)`, matching the `voxelDispatchGridForCount()` / `writeDispatchDims()` fold cited in the original finding. Confirmed the row-major flatten is gap-free: `groupId.x + groupId.y * groupsX` visits every value in `[0, groupsX*groupsY)` exactly once, which is `⊇ [0, groups)`, so no cell index is skipped. All four consumer kernels (`c_compute_voxel_ao`, `c_compute_sun_shadow`, `c_lighting_to_trixel`, `c_resolve_per_axis_screen_depth`; GLSL + Metal twins — 8 files) updated in lockstep to recover `groupIndex = groupId.x + groupId.y * gl_NumWorkGroups.x` (GLSL) / `groupId.x + groupId.y * numGroups.x` via the newly-added `[[threadgroups_per_grid]]` parameter (Metal) before the `* tile + localIndex` decode. Empty-axis edge case (`count=0` → `groups=0` → `groupsX=1, groupsY=0`) still yields a correct no-op dispatch (0 total groups). Byte-identical to the prior path whenever `groups ≤ 1024` (all default/normal pool configs), since `groupsX == groups` and `groupsY == 1` there — matches the commit message's claim.
- ✅ `engine/render/include/irreden/render/ir_render_types.hpp:1226` — misattributed compaction/finalize write — verified fixed at `2b559448`. The comment now correctly attributes `numGroupsX/Y/Z` to `c_per_axis_cell_finalize`, with the compaction described as only atomic-appending `instanceCount`.
- ✅ `ir_render_types.hpp:1236` (`kPerAxisCellComputeTile`) — missing cross-reference to shader `local_size` — verified fixed at `2b559448`. New comment explicitly names all four consumer kernels (GLSL + Metal twins) and states the silent-misalignment failure mode if one is changed without the others.

No new issues found in the fix commit — it's scoped exactly to the three flagged items, and the GLSL/Metal twins stay in lockstep (each shader-side change has a matching pair).

- [ ] Linux/OpenGL cross-host smoke — still pending per the PR body (unchanged from the prior pass, not introduced by this fix).

Escalation: please Opus-recheck before merge (touches: `engine/render/` compute-dispatch shaders (GLSL + Metal), GPU/CPU indirect-dispatch handoff).

jakildev (commented):
## Opus recheck — render: per-axis lighting/resolve indirect-dispatch over occupied cells (#2256)

**Verdict:** approve (one nit)

Opus recheck of Sonnet's approve. Sonnet verified the capped 2-D dispatch-grid fix (`2b559448`), the gap-free row-major flatten, naming, and the struct-attribution comments — I take those as confirmed and spent this pass on the invariants a first-pass checklist can't close on a GPU/CPU indirect-dispatch handoff: barrier sequencing across the compact→finalize→consumer chain, the flat-index recovery parity across all 8 GLSL/Metal consumer twins, the indirect-args struct layout, buffer lifetime, and cross-system GL binding state. Ran a focused render-invariant subagent over the full file context in parallel with my own read; the two converged.

### Confirmed (Opus-only items Sonnet could not close)

- **Barrier sequencing — correct.** `compactPerAxisCells` (`system_voxel_to_trixel.hpp`) issues `memoryBarrier(SHADER_STORAGE)` after the compaction (makes each axis's `instanceCount` visible to the finalize pass), then `memoryBarrier(SHADER_STORAGE)` **and** `memoryBarrier(COMMAND)` after the finalize. `GL_COMMAND_BARRIER_BIT` covers both the four `glDispatchComputeIndirect` consumers and the later `glDrawElementsIndirect` scatter, and a GL barrier applies to all subsequent commands until the buffer is next written — so the single barrier pair correctly covers every downstream consumer. Metal's per-encoder boundaries serialize each dispatch independently. No gap.
- **Flat-index recovery parity — correct.** All four consumers, GLSL and Metal, decode `groupIndex = groupId.x + groupId.y * numGroups.x; idx = groupIndex*256 + localIndex` and guard `idx >= cellDrawArgs[11]` (`visibleCount`) before touching `compactedCells[idx]`. Because the consumer reads `gl_NumWorkGroups.x` (the *actually dispatched* grid dim) rather than a re-derived constant, the recovery is self-consistent with whatever grid the finalize wrote — it cannot drift from `groupsX`. No kernel missed, no off-by-one, spill groups bounded.
- **Struct layout / binding — correct.** `PerAxisCellDrawCommand` (32 B / 8 uints) + `kPerAxisCellDispatchArgsOffsetBytes=32` place `numGroupsX/Y/Z` at uint[8], matching the shader's `kDispatchArgsBaseUint=8`; `kPerAxisCellIndirectStrideBytes=256` == shader `kStrideUints=64`. The CPU `dispatchComputeIndirect(..., axis*256 + 32)` agrees byte-for-byte. The buffer is read via SSBO index 26 and consumed as a dispatch-indirect source at a raw offset — no target/index collision.
- **Buffer lifetime + count=0 — correct.** `cellCompacted_`/`cellIndirect_` are owned on `C_PerAxisTrixelCanvases`, sized once in `allocate()` to the worst-case axis-cell count (64-uint aligned), freed in `release()` — same rotation lifecycle as the axis textures. Each 256 B indirect region is zeroed every frame before compaction, and the finalize unconditionally overwrites `[8..11]` for all three axes, so no stale dims leak forward. Empty axis → `groupsX=1, groupsY=0` → a legal zero-group no-op on both backends, and no consumer reads before the `visibleCount` guard.

Sonnet flagged the dispatch-grid cap on its first pass and confirmed the fix on its second; on closer read I confirm the fix is correct and, more importantly, that the surrounding indirect-dispatch machinery it plugs into is sound.

### Nit (not a merge-blocker — functionally correct today)

- **Slots 25/26 borrowed without restore in the four new per-axis consumers** — `system_compute_voxel_ao.hpp` (`dispatchPerAxisAO`), `system_compute_sun_shadow.hpp` (`dispatchPerAxisSunShadow`), `system_lighting_to_trixel.hpp` (`dispatchPerAxisLighting`), `system_resolve_per_axis_screen_depth.hpp` (`tick`). Each `bindRange`s SSBO slots `kBufferIndex_PerAxisCellCompacted` (25) / `kBufferIndex_PerAxisCellIndirect` (26) to the per-axis buffers and never restores them to the voxel single-canvas compaction buffers (`CompactedVoxelIndices` / `IndirectDispatchParams`). `VOXEL_TO_TRIXEL_STAGE_1`'s single-canvas compact binds those slots once at `create()` and trusts sticky global binding state thereafter, so a *leaked* per-axis binding into the next frame re-reads the cell list as the voxel index list and corrupts world voxels — this is the exact `#1961` center-cube regression, and it's why `system_trixel_to_framebuffer.hpp:355-361` restores 25/26 with a heavily-commented block.

  This is **not a live bug**: all five per-axis consumers share the same `perAxisCanvases_->isAllocated()` gate, and `TRIXEL_TO_FRAMEBUFFER` — which still restores — is definitionally the last per-axis stage each frame, so 25/26 are always back to the voxel buffers before next frame's `STAGE_1` compact. I verified `compactPerAxisCells` also restores at its own tail. So the frame closes clean today.

  What this PR does is quietly widen the "must restore 25/26 before end of frame" contract from one documented site to four undocumented borrow sites that depend on a *distant* system running last. A future post-composite pass that touches these slots, or any pipeline reorder, silently reintroduces the #1961 corruption with no diagnostic. **Cheap, worthwhile follow-up:** add the same restore at the tail of each of the four new dispatch functions, or centralize the borrow/restore as a single pair wrapped around the whole per-axis stage sequence rather than trusting the last consumer. Optional to address in this PR vs. a fast follow-up — the human can decide before merge.

### Test plan (unchanged from Sonnet's pass, pre-existing)

- [ ] Linux/OpenGL cross-host smoke — still pending per the PR body; the PR carries `fleet:needs-windows-smoke` and was authored on macOS, so it correctly awaits cross-host validation before/after merge.

jakildev:
Addressed the Opus recheck nit: SSBO slots 25/26 (`kBufferIndex_PerAxisCellCompacted`/`Indirect`) are now restored to the voxel-compaction buffers at the tail of all four new per-axis dispatch sites (`dispatchPerAxisAO`, `dispatchPerAxisSunShadow`, `dispatchPerAxisLighting`, and `RESOLVE_PER_AXIS_SCREEN_DEPTH::tick`'s scatter pass), not just the pre-existing `TRIXEL_TO_FRAMEBUFFER` restore. Centralized the restore logic as `IRPrefab::PerAxisCanvas::restoreVoxelCompactionSlots(Buffer*&, Buffer*&)` in `per_axis_canvas.hpp` (already included by all five systems) so the named-resource lookup + bindBase pair isn't duplicated five times; each system caches its own resolved pointers as members, mirroring the existing pattern. `system_trixel_to_framebuffer.hpp`'s own restore method now delegates to the shared helper instead of keeping its own copy.

Verified: `fleet-build --target IRShapeDebug` clean, `fleet-run IRShapeDebug --auto-screenshot 10` exercises the rotating per-axis path (yaw 45/90/180 + pan) with no errors. Linux/OpenGL cross-host smoke still pending per the PR body (unchanged). — worker

PR #2259, opened 2026-07-06
render: cap the fog cut at the boundary — entry-distance limit

jakildev:
## Summary
- Follow-up to #2246, from arcade playtesting: the analytic cut had **no limit on how far the disc entry could sit from the surface along the view ray**. Elevated top surfaces (border rails, crate tops) have rays that enter the vision cylinder at floor depth — the occupancy march finds solid there and repaints them "revealed" far past the disc's true XY radius, while the ground beside them correctly fades dark. Two symptoms: the reveal reads screen-space-ish on tall geometry, and the bright repaint sitting beyond the darker rim fade sandwiches it into a phantom shadow stripe.
- ~~`kFogCutMaxEntryCells = 2`: capped along the view ray~~ **Amended after human playtesting** (see the review thread): a ray-entry cap bounds the band by the disc swept along the view ray's world-(1,1) XY — a pure screen-vertical displacement in iso — so the band's outer edge still read screen-space-ish on elevated faces. The cap is now **`kFogCutMaxRimCells = 2`, radial in the surface's own XY** (`hardDistPastRim`, the same metric the reveal + rim fade key on); the ray intersection + occupancy march remain solely as solidity evidence. Every fog boundary — reveal, cut band, fade — is now a circle concentric with the disc at every face height. The original exclusions (border rails / slab sides far past the radius) still hold under the radial test.
- Edge/detached render-verify refs re-blessed for the narrower cap.

## Test plan
- [x] A/B causality on a thin-floor + border-rail mock (rail crossing the disc shallowly, mirroring the arcade arena): uncapped reproduces the bright rail band past the rim + shadow-stripe sandwich; capped fades monotonically by XY — verified locally, not committed
- [x] Arcade scene (game creation via `IRREDEN_USER_PROJECTS`) built against this branch and captured at the arena edge: clean
- [x] `render-verify --target IRFogDemo`: **11/11 PASS**, deterministic; grid-only default and soft-disc Mode B shots byte-identical to pre-change refs
- [x] Metal built + run; GLSL twin changed identically — needs Linux smoke

## Notes for reviewer
- 4-line logic change per backend (an `entryDistance` guard); the rest is refs + screenshots.
- The remaining fog roadmap item from playtesting — vision evaluated as XY radius **plus a Z cost penalty** (observer height in the vision-circle upload) — is a small API extension, being filed separately.

## Screenshots

<details>
<summary>fog_edge_zoom9</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

<details>
[...]

PR #2257, opened 2026-07-06
render: fog cross-section parity with SDF light blockers (#2247)

jakildev:
## Summary
#2274 landed the **geometric fog-cut cap** (face-axis + radial band) — content-independent: any surface that rasterizes caps at the vision-disc boundary, with no light-occlusion-bitfield read. That retired the ray+occupancy cut march this PR originally extended with an SDF-blocker bitfield OR, so the shader work is dropped as moot.

Re-scoped (per architect direction, now that #2274 is on master) to the durable, still-valuable half:

- **`fog_demo --edge-sdf-blocker` scene + committed `fog_edge_sdf_blocker_zoom5/zoom9` refs** — the permanent enabled-path regression gate for #2247. It spawns an SDF box carrying `C_LightBlocker{blocksLOS_=true}` straddling the -Y vision-disc arc and proves its fog-hidden half caps with the toned cut wall instead of **black-banding** (#2247's reported symptom). The geometric cap does not special-case or exclude blockers, so a `blocksLOS_` SDF straddling the disc caps by construction.
- The scene overrides to the straight-down sun (like `--edge-zoom` / `--edge-smooth`) so the cut face isn't confounded by shadow terminators (#2259 amendment).

**Dropped from the original PR:** the `c_fog_to_trixel.{glsl,metal}` march integration (`fogCutBlockerSolid` + the blocker-OR) and the `lightOcclusionGridFlatBit` helper — there is no occupancy march on master to plug into, and the helper's only consumer was that march. This also dissolves the recorded semantic conflict (both conflicted files were the march blocks).

## Verification (Metal / macos-debug)
- `--edge-sdf-blocker` renders the SDF blocker's fog-hidden -Y half as a solid toned cut wall (no black band) under the geometric cap — verified visually at zoom5/zoom9 + the cut-face ROI crop.
- `render-verify` **13/13 PASS deterministic** (0 drift): the 11 non-blocker shots byte-identical to master's refs, the 2 new blocker refs blessed against the geometric cap.

## Cross-host
- [ ] OpenGL smoke (Linux/Windows) — the geometric cap is backend-agnostic; linux-debug refs bless there (a backend without refs SKIPs). Tracked via `fleet:needs-windows-smoke`.

Closes #2247

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2274, opened 2026-07-06
render: fog cross-section cap is geometric — face axis + radial band

jakildev:
## Summary
Human playtesting post-#2259 found two artifact classes in the fog cut's **ray+occupancy solidity heuristic**; this PR replaces the heuristic with **pure per-pixel geometry**, making both classes unrepresentable.

**Artifact 1 — light band on flat ground, heavier on the bottom of the disc.** A hidden floor pixel's view ray enters the vision cylinder *underground*, the occupancy march finds the floor's own interior, and the cut repainted the floor's top surface at near-full brightness for ~2 cells past the rim, ending in a hard cliff — only on the ray-facing (down-screen) side. Measured on `fog_edge_zoom5` (vertical profile through the bottom rim): `103, 103 → 48` cliff **before**; smooth monotonic `103 → 74 → 64 → 48 → 31 → …` **after**.

**Artifact 2 — horizontal banding + voxel-stepped "teeth" on walls.** The march samples voxel-quantized occupancy along the ray, so a wall's cap flipped with the content *behind* it (solid floor behind the low half → capped flat tone; air behind the high half → dark): a horizontal band whose boundary stair-steps at voxel resolution.

**The fix.** Cap **vertical faces only** (the depth encoding's slot bits resolved through `visibleFaceIds` — the same slot→faceId mapping AO/lighting use) whose surface column is within `kFogCutMaxRimCells` of the rim (`hardDistPastRim`, the metric the reveal + rim fade already key on). Top (Z) faces never cap — flat ground and object tops fade on the plain radial curve; cut walls of sliced objects keep their cap; SDF surfaces participate via their own slot bits. No ray probe, no disc intersection loop, no march, no light-occlusion bitfield read — the test has zero content/view dependence, so every fog boundary is a circle concentric with the disc at every face height.

**Cleanup that falls out:** `FOG_TO_TRIXEL` drops the slot-28 occlusion-SSBO rebind + `cutSolidsAvailable_` stamp (now a reserved lane); both fog shaders drop the bitfield/march machinery (~70 lines each); the cap no longer depends on `BUILD_LIGHT_OCCLUSION_GRID` being registered.

## Screenshots
| before (band + toothy cliff) | after (smooth radial fade) |
|---|---|
|  |  |

Full frames + diff in `docs/pr-screenshots/claude/fog-cut-face-slot-cap/`.

## Test plan
- [x] `render-verify --target IRFogDemo`: **11/11 PASS deterministic** — grid-only + Mode B soft shots **byte-identical** (100%, max_delta 0); drift confined to the 5 hard-disc edge/detached shots (re-blessed).
- [x] Floor monotonicity trace (zoom5 bottom rim): no local maxima after — the band is gone; asymmetry gone.
- [x] Wall crop (zoom9 slab front face): smooth vertical-contour fade, no horizontal band, no teeth; the slab's cut wall keeps its cap.
- [x] Arcade-style voxel-floor scene rebuilt against this branch (`IRREDEN_USER_PROJECTS`): rim fade smooth + uniform in all directions.
- [ ] Cross-host GL smoke (GLSL twin changed identically, byte-mirrored).

## Note for #2257 (open)
With the occupancy march retired, blocker-marked SDFs cap **by construction** (their pixels carry face-slot bits): #2257's bitfield-OR change no longer has a march to plug into. Its `--edge-sdf-blocker` scene + refs remain valuable as the enabled-path regression proof, and the `lightOcclusionGridFlatBit` helper factoring still serves `c_propagate_light_volume`; the PR needs a re-scope to just those.


jakildev (commented):
## Review — render: fog cross-section cap is geometric — face axis + radial band

**Verdict:** approve

Replaces the fog cut's ray+occupancy solidity heuristic (march against the
camera-anchored light-occlusion bitfield) with pure per-pixel geometry: cap
only VERTICAL faces (resolved via `visibleFaceIds[encoded & 3] >> 1 != 2`)
within `kFogCutMaxRimCells` of the rim. Read every touched file in full
(both C++ systems, the component header, and both GLSL/Metal shader twins)
and cross-referenced the newly-appended `FrameDataVoxelToTrixel` UBO fields
against the producer's declaration.

### Correctness verification

- **UBO layout match.** `c_fog_to_trixel.glsl`'s `FrameDataVoxelToTrixel`
  block appends `uniform vec4 faceDeform[3]; uniform ivec4 visibleFaceIds;`
  after the existing `_yawPadding` placeholder. Diffed field-by-field
  against the producer (`c_voxel_to_trixel_stage_1.glsl:20-51`): every prior
  field name/type lines up exactly (the pre-existing `_voxelDispatchPadding`
  and `_yawPadding` placeholders already stood in for `perAxisRoute` /
  `isDetachedCanvas` at the correct offsets, unchanged by this PR), so the
  two new fields land at the correct std140 offset. On the Metal side
  `FrameDataVoxelToTrixel` is the single shared struct in
  `ir_iso_common.metal:869` (already carries both fields) — no struct edit
  needed there, so GL/Metal parity is structural, not just visually matched.
- **Slot→axis resolution matches the established convention.** `const int
  faceAxis = visibleFaceIds[encoded & 3] >> 1;` reuses the identical
  `slot = encoded & 3` / `visibleFaceIds[slot]` pattern already used by
  `c_compute_voxel_ao.glsl:131`, `c_lighting_to_trixel.glsl:152`, and
  `c_resolve_per_axis_screen_depth.glsl:70` — not a new indexing scheme, and
  only reached past the early-return for background pixels (`encoded >=
  kEmptyDistanceEncoded`), so `encoded` is always a rasterized-geometry
  value here.
- **No dangling references to the removed machinery.** Grepped the full
  tree for `cutSolidsAvailable_`, `fogCutSolidsAvailable`,
  `occlusionGridSsbo_`, `kFogCutRayProbeDepth`, `kFogCutMarchSteps`,
  `fogCutVoxelSolid`, `kLightOcclusionGridSize/HalfExtent` — all fully
  retired from the fog files; the only remaining hits are
  `c_propagate_light_volume.{glsl,metal}`'s own independent local copies of
  the grid constants (unrelated, unaffected consumer).
  `engine/render/CLAUDE.md`'s pipeline diagram never documented FOG_TO_TRIXEL
  as an occlusion-grid consumer, so no doc drift there either.
- **`hardState` fully removed** (was only used to force `cutSolid = true`
  on rim pixels in the retired ray-cast branch) — no leftover references in
  either shader.
- **Buffer/kernel signature cleanup is self-consistent.** The C++ system no
  longer binds `kBufferIndex_LightOcclusionGrid` (slot 28) for this
  dispatch, and the Metal kernel signature dropped the corresponding
  `buffer(28)` parameter — the remaining bindings all use explicit
  `[[buffer(N)]]` attributes, so removing the unused parameter doesn't shift
  any other binding.
- **`format-changed` clean** — ran it myself, no diff vs branch.

### Test plan verification

Ran `render-verify`-adjacent reasoning only (no local GPU harness re-run
here); the PR's own numbers are internally consistent — 5 of the touched
`macos-debug` reference PNGs are exactly the hard-disc edge/detached shots
named in the "byte-identical except 5" claim (`fog_detached_edge_zoom5/9`,
`fog_edge_zoom5/9/14`), matching "grid-only + Mode B soft shots
byte-identical, drift confined to the 5 hard-disc shots."

### Nits

- The two new comment blocks explaining why TOP faces never cap and why
  VERTICAL faces cap uniformly (`c_fog_to_trixel.glsl` around the cap site)
  narrate the retired variant's specific failure symptoms (the
  view-dependent bright ring, the voxel-stepped teeth) rather than just
  stating the current invariant. Not blocking — this matches the file's
  pre-existing comment density — but per `CLAUDE-BASELINE.md`'s guidance on
  investigation-history comments, the PR body already carries this
  narrative in full; the shader comment could trim to the durable "why"
  (faces classify by axis; TOP fades radially, VERTICAL caps at the rim)
  with a one-line pointer to the PR for the artifact history.

### Praise

- The fix genuinely eliminates the artifact class by construction rather
  than patching the heuristic: replacing a content/view-dependent ray march
  with a pure function of (rasterized face axis, radial distance) makes
  both reported artifacts unrepresentable, not just less likely.
- Good foresight flagging the interaction with the still-open #2257 in the
  PR body — calling out that the occupancy march's retirement obsoletes
  that PR's bitfield-OR hook and it'll need a re-scope, rather than leaving
  the conflict to be discovered at merge time.
- The cleanup is thorough: dropping the slot-28 rebind, the
  `cutSolidsAvailable_` stamp, and ~70 lines of march/bitfield machinery
  per shader backend, plus updating every touched comment (component
  header, system tick, both shader UBO declarations) to describe the new
  reality instead of leaving stale prose next to correct code.

- [ ] Cross-host GL smoke (flagged unchecked in the PR body; tagging
      `fleet:needs-windows-smoke` per the authored-on-macOS tier rule).

Escalation: please Opus-recheck before merge (touches: `engine/render/`
shaders + `engine/prefabs/irreden/render/` — GPU buffer layout change
shared across GL/Metal).

jakildev (commented):
## Opus recheck — render: fog cross-section cap is geometric (face axis + radial band)

**Verdict:** approve

Sonnet approved and escalated for an Opus recheck on the GPU buffer-layout
change shared across GL/Metal. I focused this pass on exactly the invariants
Sonnet's first pass could assert but not fully prove: the std140 layout of the
two appended `FrameDataVoxelToTrixel` fields, the GL/Metal struct parity, and
the safety of the new `visibleFaceIds[encoded & 3]` slot read now that the
occupancy march is gone. All confirm.

### std140 UBO layout — confirmed (Sonnet flagged; I verified offset-exact)

Sonnet diffed the consumer block field-by-field against the producer. On closer
read I confirm the layout is offset-identical, not merely name-similar:

- Walked the full binding-7 block in `c_fog_to_trixel.glsl` against the producer
  in `c_voxel_to_trixel_stage_1.glsl`. All 15 slots line up by type and order.
  The two differing names are pre-existing placeholders at the same type/offset
  (`_voxelDispatchPadding`↔`perAxisRoute`, `_yawPadding`↔`isDetachedCanvas`) —
  untouched by this PR, which only appends `vec4 faceDeform[3]` + `ivec4
  visibleFaceIds`.
- The alignment corner: `visualYaw / rasterYaw / residualYaw / _yawPadding` are
  four consecutive floats filling exactly one 16-byte lane, so `faceDeform[3]`
  starts on a std140 vec4 boundary with **no hidden padding gap** — and the
  producer has the identical four-float run there, so both sides pad the same.
- Metal: the fog kernel reads `frameData.visibleFaceIds` off the **shared**
  `FrameDataVoxelToTrixel` struct (not a local mirror). Confirmed
  `ir_iso_common.metal` already carries `int4 visibleFaceIds` + `float4
  faceDeform[3]` — so the Metal kernel compiles and GL/Metal parity is
  structural, not just visually matched.

### `visibleFaceIds[encoded & 3]` slot read is safe — confirmed

The retired variant gated on a ray/occupancy test; the new code indexes
`visibleFaceIds` directly, so the read must never see a non-geometry `encoded`.
Verified both backends early-return on the empty sentinel
(`encoded >= kEmptyDistanceEncoded`, GLSL:157 / Metal:86) **before** the cap
site (GLSL:256 / Metal:178), so `encoded` there is always a rasterized-geometry
value and `encoded & 3` is a valid slot. `faceAxis = ... >> 1; faceAxis != 2`
correctly maps FaceId 4/5 (Z_NEG/Z_POS) to the excluded TOP axis and keeps
X/Y verticals — matching the comment and the identical `encoded & 3` slot
convention already in AO / lighting / per-axis-resolve.

### Binding-28 removal — confirmed no cross-dispatch fallout

The C++ system drops the slot-28 (`kBufferIndex_LightOcclusionGrid`) rebind and
the Metal kernel drops its `buffer(28)` parameter; the shader no longer declares
the std430 block. Fog was only *rebinding* the shared alias slot for its own
read — removing that leaves slot 28 holding whatever the prior pass set, and
every pass that consumes it rebinds itself, so nothing downstream is disturbed.
The retired `occlusionGridSsbo_` member and its lazy `getNamedResource` lookup
have no other referents.

### On Sonnet's nit (comment trim) — I disagree, dropping it

Sonnet suggested trimming the new comment blocks that narrate the retired
variant's failure symptoms (the view-dependent bright ring, the voxel-stepped
teeth). On Opus read these are durable *why* documentation, not history noise:
they justify why TOP faces are never capped and why the geometric test replaced
occupancy, which is exactly the non-obvious rationale the baseline wants kept
next to the code. Non-blocking either way — clearing `has-nits`.

The design behavior change (all fog-hidden vertical faces in the radial band now
tint, vs. the old occupancy-gated subset) is intentional and covered by the 5
refreshed `macos-debug` reference PNGs + before/after screenshots. Cross-host GL
smoke remains gated via `fleet:needs-windows-smoke` — correct for the
authored-on-macOS tier.

Nice work — this eliminates the artifact class by construction rather than
re-tuning a heuristic.

jakildev:
**Follow-up (23f08651): feathered the cap into the fade** — human playtest on the first cut caught a harsh near-vertical line + dotted teeth on the slab's front face at the cap band's edge.

Root cause: the cap was **binary** over its band — full cut tint for radial ≤ `kFogCutMaxRimCells`, then a step straight to the fade (~85% → ~53%). The step read as a hard line down the wall, dithered into voxel-stepped dots by the per-voxel-quantized depth recovery; the capless top face faded smoothly past it, so side and top visibly disagreed.

Fix: the cut tint now blends over the fade with weight 1 at the rim → 0 exactly at the band's end, so a wall is one continuous curve that **converges to the top face's exact fade tone** — the xy/z face disconnect itself now fades out with distance instead of ending on a cliff. Both shaders restructured (cap = post-fade blend, no early return).

Band-edge detail at zoom14 (binary cap left, feathered right — note the dotted tooth column, gone):

| binary cap | feathered |
|---|---|
|  |  |

render-verify 11/11 PASS deterministic; only the 3 hard-disc edge refs re-blessed (detached scenes show no vertical faces in the band, byte-stable).

jakildev:
**Follow-up 2 (3db6e696): the cut tone is now a pure multiply — dropped `kFogCutLift`.**

Playtest caught vertical faces reading "not faded" past the rim. Measured on the zoom14 pillar face (green channel, walking outward): lit `64` → capped **`70`** — hidden matter rendered *brighter than its lit self*. Cause: the `+0.06` constant lift on top of the 0.85 tone inverts for any surface darker than ~40% brightness (`0.85·src + 0.06 > src`). The old ray+occupancy march had masked this on upper walls (it found air behind them → fade); the geometric cap correctly covers the whole vertical band, which exposed the inversion.

The lift's original job — keeping near-black cuts readable against fog black — is covered by the feathered blend into the rim fade now, so it's gone. Invariant after: `fade(0.75) ≤ cap(0.85·src) < lit(src)` — **brightness is monotone across the rim on every material**. Pillar trace after: `64 → 54 → 50 → 40 → 29 → …`.

**Arcade verification** (game rebuilt against this branch):
- Radial floor traces from the disc center: `46,55,70` interior → smooth monotone decline → black, in all directions — no band, no ring, no asymmetry.
- Whole-frame scan: zero pixels brighter than their lit tone inside the fog region (only HUD text + pickup-cube highlights exceed floor brightness).
- With the vision rim pushed out to the border rails (temp `vision_radius = 38` vs `arena_extent = 40`): the rail crossing the rim fades smoothly along its length (`0 → 57 lit plateau → 0`), no horizontal band, no teeth — the original wall-artifact scenario, clean in-game: 

render-verify 11/11 PASS deterministic; edge refs re-blessed. (2f0c72bb also re-captured the PR evidence images — the prior commit had accidentally staged Mode-B smooth shots left in the screenshots dir by render-verify; the committed refs themselves were always correct, verified byte-identical against a fresh capture.)

PR #2279, opened 2026-07-07
lua-ecs: expose C_LocalTransform.rotation_ on the C++-component Lua binding (#2191)

### LINKED ISSUE #2191: lua-ecs: expose C_LocalTransform.rotation_ on the C++-component Lua binding (per-field setter ergonomics)
**Area:** engine/script (C++-component Lua bindings)
**Model:** opus
**Blocked by:** #2163

## Gap

Follow-up split out of #2163's plan (see its premise-correction section). #2163 adds a `VEC4`/`quat` `LuaFieldType` column for **Lua-defined** components — but `C_LocalTransform` is a **C++-defined** component reached through `LuaCppColumnView` (`at`/`setAt`) whose accessors come from `registerType<T>` / `recordComponentLuaName` (`engine/script/include/irreden/script/lua_script.hpp`). Its binding (`component_local_transform_lua.hpp`) exposes only `translation_x/y/z` read-getters; `rotation_` is not exposed at all, and C++-component writes are whole-object replace, not per-field.

## Ask

Expose `rotation_` (an `IRMath::vec4` quaternion) on the `C_LocalTransform` Lua binding so an EVAL tick can drive entity rotation without a C++ shim — retiring stopgaps like the game's `FacingToRotation` bridge (jakildev/irreden#216).

## Design question to settle at planning time

Per-field setter vs whole-object-replace ergonomics for C++-component writes from Lua — this is broader than quaternions (any C++ component field Lua wants to mutate hits it). Decide once here and document the convention in `engine/script/CLAUDE.md`.

## Acceptance criteria

- A Lua EVAL system can set an entity's `C_LocalTransform.rotation_` (e.g. from a yaw angle helper) and the rendered orientation updates.
- Convention for C++-component per-field writes from Lua documented.
- No regression in existing `translation_x/y/z` getters or whole-object `setAt` paths.

## Context

Filed during backlog triage 2026-07-03 as the second half of #2163's premise correction; #2163's plan explicitly recommends this split so the LuaFieldType work isn't held hostage to the binding-ergonomics design call.

jakildev:
## Summary
- Add read-write `rotation` / `translation` / `scale` `sol::property` bindings to the `C_LocalTransform` C++-component Lua binding, so an EVAL system can drive an entity's SQT per field (`arch.C_LocalTransform:at(i).rotation = { x, y, z, w }`) without a whole-object `setAt` or a C++ shim.
- Setters route a `{ x, y, z, w }` table (or an `IRMath` userdata) through the existing `IRScript::quatFromLua` / `vec3FromLua` converters; getters return a fresh `{ x, y, z[, w] }` table (the Lua-defined packed-vec shape). Keyed `{x=,y=,z=,w=}` and indexed `{1,2,3,4}` spellings both work; nil/partial → identity for the quat.
- The zero-alloc `translation_x/y/z` scalar getters are **kept unchanged** (still bound as member functions).
- Properties are registered on the returned usertype *after* `registerType()` so the `property_wrapper` is pushed as an **rvalue** — the vendored sol2's const-lvalue property pusher is broken (`unqualified_pusher<property_wrapper>`'s `const&` overload references `pw.read`/`pw.write` without calling them and fails to compile). Comment in the binding explains why.
- Document the C++-component per-field-write convention (scalar member pointer vs math-typed `sol::property` table) in `engine/script/CLAUDE.md`; refresh the now-stale "constructor-only" note in the `C_RotationTarget` binding comment.

## Test plan
- [x] New headless `test/script/lua_local_transform_lua_test.cpp` (mirrors `lua_rotation_target_lua_test.cpp`): keyed + indexed rotation writes drive `C_WorldTransform.rotation_` through `PROPAGATE_TRANSFORM`; per-frame re-write; partial-table → identity; translation + scale table round-trip; legacy `translation_x/y/z` scalar getters unchanged.
- [x] `IrredenEngineTest` green: 1317 passed, 1 skipped (GPU/OpenGL-only, expected on macOS/Metal), 0 failed.
- [x] `format-changed` clean.

Closes #2191

### DIFF
--- a/engine/prefabs/irreden/common/components/component_local_transform_lua.hpp
+++ b/engine/prefabs/irreden/common/components/component_local_transform_lua.hpp
@@ -2,6 +2,7 @@
 #define COMPONENT_LOCAL_TRANSFORM_LUA_H
 
 #include <irreden/common/components/component_local_transform.hpp>
+#include <irreden/script/ir_script_utils.hpp>
 #include <irreden/script/lua_script.hpp>
 
 namespace IRScript {
@@ -9,7 +10,9 @@ template <> inline constexpr bool kHasLuaBinding<IRComponents::C_LocalTransform>
 
 template <> inline void bindLuaType<IRComponents::C_LocalTransform>(LuaScript &luaScript) {
     using IRComponents::C_LocalTransform;
-    luaScript.registerType<
+    // The scalar convenience getters (translation_x/y/z) stay as the zero-alloc
+    // read path and register in the usual key/value list, unchanged.
+    sol::usertype<C_LocalTransform> type = luaScript.registerType<
         C_LocalTransform,
         C_LocalTransform(IRMath::vec3),
         C_LocalTransform(IRMath::vec3, IRMath::vec4),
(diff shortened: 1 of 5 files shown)


jakildev (commented):
## Review — lua-ecs: expose C_LocalTransform.rotation_ on the C++-component Lua binding (#2191)

**Verdict:** approve

Standalone PR (base `master`). Adds read-write `rotation`/`translation`/
`scale` `sol::property` bindings to `C_LocalTransform`'s Lua binding so an
EVAL system can drive SQT fields per-field (`arch.C_LocalTransform:at(i)
.rotation = {x,y,z,w}`) without a whole-object `setAt`.

### Correctness verification

- **The vendored-sol2-bug claim is real — verified against the actual
  header, not just trusted.** `build/_deps/sol2-src/include/sol/
  function_types.hpp:462-472`, the const-lvalue
  `unqualified_pusher<property_wrapper<F,G>>::push` overload, references
  `pw.read` / `pw.write` as bare non-static-member-function names (no `()`
  call, no `&`) — that's ill-formed to instantiate (a non-static member
  function name used without calling it or taking its address). The
  rvalue overload two lines up (line 450-460) correctly calls `pw.read()`
  / `pw.write()`. So registering the property via `registerType`'s
  variadic key/value pack (which would hit the const-lvalue path) really
  would fail to compile, and pushing the `sol::property(...)` result as
  an rvalue directly onto the returned `sol::usertype<>` (this PR's
  approach, `component_local_transform_lua.hpp:50-88`) is the correct
  workaround, not a superstition.
- **Partial-table → identity default is correct and tested.**
  `quatFromLua` (`ir_script_utils.hpp:81-101`) fills missing `w` with
  `1.0f` and missing `x`/`y`/`z` with `0.0f`, giving identity for `{}` —
  confirmed by `PartialRotationTableDefaultsToIdentity`, which
  deliberately starts from a non-identity rotation so the identity result
  can't be an untouched default.
- **`C_RotationTarget`'s comment refresh is accurate.** The old
  "constructor-only" note is replaced with a correct pointer to the new
  math-typed-field convention; `input_`/`minAngle_`/etc. staying
  member-pointer-bound (scalar path) is unchanged.
- **Doc addition (`engine/script/CLAUDE.md`) matches the code.** The new
  "C++-component per-field writes from Lua" section's scalar/math-typed/
  whole-object trichotomy is exactly what the two touched `*_lua.hpp`
  files now do.

### Build & test (ran myself, not just trusting the PR body)

- `fleet-build --target IRShapeDebug` and `--target IrredenEngineTest` —
  both clean on this exact commit (`9e5aba0b`).
- `IrredenEngineTest --gtest_filter="LocalTransformLuaTest.*:
  RotationTargetLuaTest.*"` — 6/6 pass (keyed table, indexed table,
  partial→identity, translation+scale round-trip incl. unchanged legacy
  scalar getters, plus the pre-existing `RotationTargetLuaTest` pair).
- Full suite: **1317 passed, 1 skipped (`GpuComputeDispatchTest.
  SkippedOnNonOpenGLBackend`, expected on macOS/Metal), 0 failed** —
  matches the PR body's reported numbers exactly.

### Checklist

- No ECS-tick, allocation-hot-path, or lifetime concerns — this only adds
  Lua binding glue on top of two existing components; no new component,
  system, or cross-entity reach.
- No `engine/math/`, `engine/render/`, `engine/entity/`, `engine/system/`,
  or `engine/world/` surface touched.
- Lua enum/string rules (`cpp-lua-enums.md`) don't apply — no enum
  surface here.
- `sol::this_state` passed to each getter avoids capturing the Lua state,
  matching the existing binding conventions in this file.

### Praise

- Citing the exact vendored-sol2 line and explaining *why* the rvalue
  push path is required (rather than "trust me, this crashes/doesn't
  compile") is exactly the kind of precedent citation that survives a
  future sol2 upgrade — the comment in the binding file will tell the
  next author whether the workaround is still needed.
- Test coverage is unusually complete for a binding change: keyed *and*
  indexed table spellings, the partial-table default path, a second-tick
  re-write (proves the per-frame write path, not just construction-time),
  and an explicit assertion that the pre-existing zero-alloc scalar
  getters are untouched.

- [x] `fleet-build --target IrredenEngineTest` clean (verified
      independently above).
- [x] Full suite 1317/1318 passed, 1 expected skip (verified
      independently above, matches PR body).

Opus recheck not required.

Issue #2281, opened 2026-07-07
render: attribute the residual per-axis yaw GPU delta, per stage (post-#2273)


jakildev:
**Area:** engine/render (per-axis pipeline, GPU perf)
**Model:** opus

## Context

#2256 documented a residual yaw-vs-cardinal GPU frame gap attributed to the
per-axis lighting/composite stages. It was closed by PR #2273, which landed
occupied-cell **indirect dispatch** for the four per-axis compute stages —
but #2273's own A/B measured **perf-neutral vs master on macOS/Metal** ("no
regression"; the perf-win confirmation box was left unchecked). The
machinery is on master; the delta it targeted is still unrecovered.

That neutral result was predicted: the parallel attempt PR #2271
(design-blocked, now closed) measured `resolvePerAxisScreenDepth` at
0.00 ms on the *pre-#2273* full-grid sweep — empty early-return
invocations were already effectively free, so occupied-only dispatch had
~nothing to reclaim. The per-axis cost is occupied-cell-bound. See
`docs/design/gpu-stage-timing-cost-model.md` (in review) and the
`## Architect direction` on #2271.

So the original question is still open: **where does the yaw GPU delta
actually live, per stage?** No lever should be picked (pass collapse over
union coverage, per-cell cost reduction, scatter changes) until the
attribution table exists.

## Acceptance criteria

See the `## Plan` comment (measurement-first; no engine changes; design
gate before any implementation).

jakildev:
## Plan

(Transferred from the architect's amended plan on closed #2256 —
2026-07-06 — adjusted for #2273 having merged.)

### Phase 1 — attribute the yaw delta (measurement only, NO engine changes)

- Scenes: `IRPerfGrid --mode voxel_set --zoom 3` (the pose set from #2256's
  header) and the default wave scene as a second datapoint.
- Capture settled `--auto-profile 300` GPU STAGES tables at cardinal and at
  `--yaw 0.35`, same scene/zoom, quiet host preferred. Current master
  (post-#2273: per-axis stages indirect-dispatch over occupied cells).
- Deliverable: a per-stage cardinal-vs-yaw delta table posted on this issue
  — `voxelStage1`, `computeVoxelAO`, `computeSunShadow`, `lightingToTrixel`,
  `resolvePerAxisScreenDepth`, `trixelToFb`, plus any other mover — showing
  which stages sum to the yaw delta, with a stated noise bound (repeat
  runs).
[...]

Issue #2280, opened 2026-07-07
profile: intra-tick GPU sub-stage scopes — wire the reserved voxelStage1 sub-rows (compact / clear / stage-1 / stage-2)


jakildev:
**Area:** engine/render + engine/prefabs/irreden/render (profiling)
**Model:** opus

## Context

#2258's measurement showed the `voxelStage1` GPU stage at ~139–152 ms at
zoom 16 (IRPerfGrid wave scene, sun shadows on, Metal), **invariant to a
256× reduction in stage-1 per-invocation raster work** (forcing every voxel
to a 1×1 micro-tap left it unchanged within run-to-run noise). The cost is
therefore not in the stage-1 shader body — but `voxelStage1` is a
per-`SystemId` measurement that bundles the whole
`VOXEL_TO_TRIXEL_STAGE_1` tick: visibility compact + canvas clear +
stage-1 + stage-2 dispatches (see the registry comment at
`engine/prefabs/irreden/render/gpu_stage_timing.hpp:196-209`). The bundled
timer cannot attribute the cost, and the registry's reserved sub-stage rows
(`canvasClear`, `voxelCompact`, `voxelStage2`) have no writer today.

#2258 (high-zoom stage-1 cost) is re-scoped to be **blocked by this issue**:
its lever cannot be chosen until the attribution exists. The sibling
per-axis investigation #2256 does NOT need this (its stages are separate
systems, already individually timed).

## Goal

An intra-tick scoped GPU timing mechanism so a system can bracket individual
dispatch groups inside one tick, wiring the reserved `voxelStage1` sub-rows
on **both** backends, then use it to attribute the #2258 cost.

## Design constraints (read before planning the mechanism)

- The per-system `GpuStageTimingObserver` pair spans the whole tick. On
  Metal it works via a sticky encoder-attachment global
  (`g_nextComputeTimestampAttachment`, `metal_render_impl.cpp`): the stage's
  first encoder claims the start boundary, every subsequent encoder
  re-writes the end (#1746). Sub-scopes must compose with that enclosing
  pair. Two candidate shapes — implementer picks:
  1. Sub-scopes attach their own `MTL::CounterSampleBuffer` on a second
     sample-buffer attachment index, leaving index 0 to the enclosing
     per-system pair; or
  2. When sub-scopes are active for a tagged system, the outer row yields
     and is computed as the **sum** of sub-scope samples (one attachment at
     a time).
[...]

PR #2282, opened 2026-07-07
docs: GPU stage-timing reading contract + dispatch cost model

jakildev:
## Summary
- New `docs/design/gpu-stage-timing-cost-model.md` — source of truth for
  (1) the GPU STAGES timer reading contract (per-`SystemId`
  encoder-boundary counter pairs; which registry rows are unwired; which
  are whole-tick bundles) and (2) the measured dispatch cost model (empty
  early-return sweeps are effectively free; shader-side early-return
  cannot reclaim CPU-fixed dispatch-grid cost; attribution before
  re-architecture).
- Cross-ref section added to `engine/prefabs/irreden/render/CLAUDE.md`
  next to the timing headers.

## Why
Two independent perf plans were refuted by measurement mid-implementation
for the same premise errors (#2256 → PR #2271 design-blocked; #2258 → PR
#2266 design-blocked), and PR #2273 independently built the
occupied-only-dispatch conversion whose own A/B then confirmed
perf-neutral. Three data points, one missing doc. Architect follow-ups
tracked: #2280 (sub-stage scopes), #2281 (yaw-delta attribution), #2258
(parked on #2280).

## Test plan
- [x] Cited paths, registry-comment claims, and stage-name strings
  verified against current master (docs-only diff; no build surface).

https://claude.ai/code/session_01MSyg38F9rLUYgXxe5VFnrY

### DIFF
--- a/docs/design/gpu-stage-timing-cost-model.md
+++ b/docs/design/gpu-stage-timing-cost-model.md
@@ -1,0 +1,132 @@
+# GPU stage timing: reading contract + dispatch cost model
+
+Source of truth for (1) what the GPU STAGES timers actually measure and how
+to read the table without misattributing cost, and (2) the measured GPU
+dispatch cost model that perf plans must build on. Written after two
+independent perf plans (#2256 → PR #2271, #2258 → PR #2266) were refuted by
+measurement mid-implementation for the same two reasons: modeling
+"dispatched-but-early-returning invocations" as reclaimable cost, and
+reading bundled/unwired timer rows as if they were per-dispatch
+measurements.
+
+Consumers: anyone planning a `fleet:needs-plan` perf issue against the
+render pipeline, the `optimize` skill flow, and reviewers of dispatch-shape
+changes (occupied-only lists, indirect dispatch, subdivision caps).
+
+## 1. What the timers measure
+
+- **One measurement per tagged `SystemId`, bracketing the whole tick.**
+  `IRRender::tagGpuStage(systemId, "<name>")` registers a system against a
+  row of `gpuStageRegistry()` (`engine/prefabs/irreden/render/
+  gpu_stage_timing.hpp`); `GpuStageTimingObserver` writes a START timestamp
+  before the tick and an END after it
+  (`gpu_stage_timing_observer.hpp`).
+- **The samples are genuine GPU-timeline times, not CPU encode times.** On
+  Metal each pair is an `MTL::CounterSampleBuffer` attached at encoder
+  boundaries (`engine/render/src/metal/metal_render_impl.cpp`): the stage's
+  first encoder claims the start boundary and every subsequent encoder
+  re-writes the end, so the resolved pair spans [first encoder starts on
+  GPU, last encoder ends on GPU] (#1746). OpenGL uses the device timestamp
+  pairs in `opengl_render_impl.cpp`. Readback is async (3 pairs in flight);
+  the overlay shows the most recent *resolved* sample, and per-stage
+  accumulators build the shutdown avg/min/max (#1738). A `finish()`-bracket
+  legacy path exists for devices without timestamp support
+  (`legacyFinishTiming_`).
+- **Several registry rows have NO writer and always read 0.000:**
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — docs: GPU stage-timing reading contract + dispatch cost model (#2282)

**Verdict:** approve

### What this PR does

Adds `docs/design/gpu-stage-timing-cost-model.md` as the source of truth
for (1) what the GPU STAGES timers actually measure and which registry
rows are unwired/bundled, and (2) the measured dispatch-cost model derived
from three independent perf investigations (#2256/#2271, #2258/#2266,
#2273) that all converged on the same "empty early-return invocations are
effectively free; you can't reclaim fixed dispatch-grid cost from the
shader side" conclusion. Cross-references the new doc from
`engine/prefabs/irreden/render/CLAUDE.md` next to the existing timing
headers. Docs-only diff, no build surface.

### Correctness verification

Checked out the branch and cross-referenced every concrete claim against
current master rather than trusting the prose:

- **Registry contents.** `gpu_stage_timing.hpp:20-26,212-218` — confirmed
  `canvasClear`, `voxelCompact`, `voxelStage2`, `shapeCompact`, `shapePass0`
  are exactly the rows in the registry array, and the doc's "four rows
  unwired, two rows are bundles" claim matches the header comment at
  `gpu_stage_timing.hpp:197-206` verbatim.
- **`tagGpuStage` + observer mechanics.** `gpu_stage_timing_observer.hpp:213`
  (`tagGpuStage`) and lines 92/122 (`writeTimestamp(..., START/END)`) match
  the doc's "START before tick, END after" description; `legacyFinishTiming_`
  exists at line 147 as described.
- **Metal encoder-boundary claim.** `metal_render_impl.cpp:31-36` has an
  inline comment describing the exact "first encoder claims start, every
  subsequent encoder re-writes end, sequential GPU execution makes the last
  write win" mechanic the doc describes — down to citing the same issue
  (#1746).
- **OpenGL device-timestamp claim.** `opengl_render_impl.cpp:235-242`
  (`writeTimestamp` → `glQueryCounter(..., GL_TIMESTAMP)`) matches.
- **`resolvePerAxisScreenDepth` "NOT unwired" claim** (the doc's example of
  the distinguishing-0.000-from-genuinely-small mistake) — confirmed it has
  both a registry row (`gpu_stage_timing.hpp:230`) and an active
  `tagGpuStage` call site (`system_resolve_per_axis_screen_depth.hpp:235`).
- **All nine cited issue/PR numbers** (#2256, #2258, #2266, #2271, #2273,
  #2280, #2281, #1746, #1738) resolved via `gh api` against their titles —
  every one matches the doc's description of what it is and its current
  state (e.g. #2271/#2273 closed as the two indirect-dispatch conversions
  that independently confirmed perf-neutral; #2280/#2281/#2258 open as the
  doc's own "open items" section states).
- **`CLAUDE.md` cross-reference path.** The new section's
  `../../../../docs/design/gpu-stage-timing-cost-model.md` link uses the
  same four-`../`-levels-up pattern as the existing
  `detached-canvas-density-compensation.md` link earlier in the same file
  (`engine/prefabs/irreden/render/CLAUDE.md` is 4 directories below repo
  root) — correct depth, verified by grep rather than assumed.

Nothing in this diff touches ECS/render code, shaders, or any GPU-lifetime
surface — it's prose plus one doc cross-reference, so none of the
Opus-only checklist items apply.

### Praise

- Turns three independent, previously-siloed perf investigations that hit
  the same wall (#2256→#2271, #2258→#2266, #2273) into one durable,
  citable rule set instead of letting the next perf plan rediscover it a
  fourth time.
- §5 ("what to verify when touching the timers") closes the loop for the
  doc's own maintenance — anyone wiring #2280's sub-stage scopes gets a
  concrete byte-identity + row-sum check instead of just "update the docs."
- Every factual claim I checked was independently verifiable against the
  current tree — no miscited precedent, no stale issue number, no drifted
  file:line reference.

- [x] Docs-only diff; no build needed (matches diff scope — no code files
  touched).

Opus recheck not required.

PR #2284, opened 2026-07-07
fleet: track live architect sessions across /clear for resume

jakildev:
## Summary
- New `fleet-session-track` SessionStart hook: rewrites the architect session sidecar (`~/.fleet/sessions/<role>.session-id`, or the `~/.fleet/solo/` variant) with the live session id on every session start — including the new session an interactive `/clear` creates — so `fleet-up` / `solo-architect` resume the conversation the pane was actually in, not the first-ever-launch session pinned months ago.
- On `source=clear` the hook re-injects the role via `additionalContext`: it embeds the full contents of `.claude/commands/role-<role>.md` with `$ARGUMENTS` substituted by the launch mode (what the original slash command would have injected), so the fresh session carries the role unconditionally. If the role file is unreadable or oversized it falls back to a directive to re-read the file. `/clear` previously dropped the role instructions and nothing re-fired them.
- `fleet-up` wires the hook into every fleet worktree's `settings.local.json` (guarded with `command -v` so hosts whose `~/bin` predates the tool stay quiet), and preserves hand-added hook groups across regeneration the same way user-granted `permissions.allow` entries are preserved (the fleet's own entry is matched on the `fleet-session-track` marker so a stale variant is replaced, not duplicated). The hook is env-gated on `FLEET_SESSION_FILE`, exported only by `fleet-babysit`'s architect path and `solo-architect`, so worker/reviewer/manual sessions no-op.
- `install.sh` registers the new tool's `~/bin` symlink; babysit/fleet-up comments and the fleet-up help text now describe the live-tracking behavior.

Observed failure this fixes (2026-07-06): both architect sidecars still pinned their April 20 first-launch UUIDs; the panes had accumulated ~75 orphaned post-`/clear` sessions, and each fleet-up snapped the pane back to the stale April-rooted transcript (80 MB / 50 MB), abandoning the human's post-clear work.

## Test plan
- [x] `scripts/fleet/tests/test_session_track.sh` — 29 assertions: no-op without `FLEET_SESSION_FILE`, sidecar write on startup, repoint + role-file embed with `$ARGUMENTS` substitution on clear, directive fallback when the role file is missing, silent update on resume/compact, silent-clear without a role var, malformed payloads exit 0 and leave the sidecar untouched, parent dir auto-created
- [x] `scripts/fleet/tests/test_worktree_settings_hooks.sh` — 11 assertions: fresh write, hand-added-hook preservation, idempotency, malformed-file fallback
- [x] `scripts/fleet/tests/test_babysit_launch.sh` — extended with T4 asserting the architect launch exports the hook gate vars; T1–T3 unchanged and green
- [x] `scripts/fleet/tests/test_babysit_effort.sh` — regression, green
- [x] `write_worktree_settings` heredoc exercised in a sandbox — emitted JSON parses and carries the hooks stanza
- [x] End-to-end: guarded hook command run exactly as the settings invoke it, with a `source=clear` payload against the real `role-opus-architect.md` — sidecar repointed, full role wrapper embedded with `mode: live` substituted
- [x] `ruff check scripts/` clean; `bash -n` clean on all edited scripts

## Notes for reviewer
The atomic tmp+`os.replace` write in `fleet-session-track` intentionally duplicates the idiom of `fleet-state-scout`'s `write_atomic` rather than sharing it: `write_atomic` lives inside the scout script (not an importable `fleet_*.py` module) and a SessionStart hook should stay a dependency-free single file. Extracting a shared module is a reasonable follow-up if a third copy ever appears.


jakildev (commented):
## Review — fleet: track live architect sessions across /clear for resume (#2284)

**Verdict:** approve

Standalone PR (base `master`, MERGEABLE). Fleet-tooling only —
`scripts/fleet/` bash + one new Python hook script; no engine/render/ECS
surface touched.

### What this PR does

Adds `fleet-session-track`, a `SessionStart` hook that rewrites an
architect's session sidecar (`~/.fleet/sessions/<role>.session-id`, or the
solo variant) with the live session id on every session start — including
the new session an interactive `/clear` creates. Before this, the sidecar
was written once at first-ever launch and never updated, so `fleet-up`
kept resuming a stale pre-`/clear` transcript (observed: both production
sidecars pinned their April 20 first-launch UUIDs after ~75 accumulated
post-`/clear` sessions). `fleet-babysit` and `solo-architect` now export
`FLEET_SESSION_FILE`/`ROLE`/`MODE` so the hook knows which sidecar to
repoint; `fleet-up` wires the hook into every worktree's
`settings.local.json`, guarded by `command -v` for hosts whose `~/bin`
predates the tool.

### Correctness verification

- **Gating is sound.** `fleet-session-track` no-ops immediately if
  `FLEET_SESSION_FILE` is unset (`fleet-session-track:38-39`) — worker,
  reviewer, and manual `claude` sessions in the same worktree never touch
  the sidecar, only `fleet-babysit`'s architect path (`fleet-babysit:274-278`)
  and `solo-architect:110-112`, which explicitly opt in.
- **Env-var propagation is real, not assumed.** `fleet-babysit` exports
  `FLEET_SESSION_FILE`/`ROLE`/`MODE` into its own shell environment *before*
  invoking `claude` (`fleet-babysit:274-278`, ahead of both
  `launch_architect_first` branches). A hook subprocess Claude Code spawns
  inherits the launching process's environment down the tree, so the
  exported vars reach the hook without any extra plumbing — verified this
  is the only path they need to travel.
- **Atomic write, matches an established idiom.** `fleet-session-track`
  writes to `<file>.tmp.<pid>` then `os.replace()`s over the sidecar
  (`fleet-session-track:52-58`) — `fleet-babysit`'s relaunch loop can `cat`
  this file at an arbitrary moment, and the PR's own dev-note correctly
  scopes why it doesn't import `write_atomic` from `fleet-state-scout`
  (that function isn't in an importable module, and a hook should stay a
  dependency-free single file).
- **Exit code contract is respected.** Every failure path (JSON parse
  error, missing/non-string `session_id`, `OSError` on write) returns 0,
  never raises — a hook crash must not degrade a session start, and the
  script honors that everywhere I traced it.
- **Ran both new/touched test suites myself** (not just trusting the PR
  body) from the branch's own worktree
  (`.claude/worktrees/architect-session-track`, since the branch was
  already checked out there and I couldn't `gh pr checkout` a second copy
  — reviewed via `git show <sha>:<path>` instead):
  - `scripts/fleet/tests/test_session_track.sh` — 25/25 pass (no-op
    without the gate var, startup/resume/compact/clear write paths, role-
    reminder `additionalContext` on clear, malformed-payload resilience,
    parent-dir auto-create).
  - `scripts/fleet/tests/test_babysit_launch.sh` — 8/8 pass, including the
    new T4 asserting the architect launch path exports the three gate vars
    with the exact sidecar path/role/mode.
  - `ruff check` on the extracted `fleet-session-track` script and the
    `write_worktree_settings` Python heredoc — clean (the one `ruff`
    complaint I got from a manual heredoc extraction was an artifact of my
    own extraction plus a pre-existing unsorted-import nit on a line this
    diff doesn't touch — confirmed via `git log -S` that line predates
    this PR).
  - `bash -n` clean on all four edited/added shell scripts.

### Nits

- `scripts/fleet/fleet-up`'s `write_worktree_settings` fully regenerates
  each worktree's `.claude/settings.local.json` on every `fleet-up`,
  special-casing only `permissions.allow` for preservation across runs
  (`fleet-up:~857-865`). This PR adds the first `"hooks"` key that
  function has ever written — if a human (or the `update-config` skill,
  which this repo documents for adding hooks to `settings.json`) later
  hand-adds a *second* hook directly to a specific worktree's
  `settings.local.json`, the next `fleet-up` silently wipes it, same as
  it already would for any hand-added key besides `allow`. Not a
  regression this PR introduces (the clobber-everything-but-`allow`
  pattern predates it), but worth a follow-up to extend the preservation
  logic to `hooks` now that the function actually writes one.

### Praise

- The PR explicitly diagnosed and cited the production failure mode
  (both real architect sidecars pinned to April 20 UUIDs) rather than
  presenting this as a speculative hardening — makes the fix easy to
  trust.
- The dev-note declining to share `write_atomic` with `fleet-state-scout`
  states a concrete reason (not importable, hook should stay
  dependency-free) instead of silently duplicating the idiom without
  comment.
- `command -v fleet-session-track` guard in the wired hook command keeps
  `fleet-up` quiet on a host whose `~/bin` symlinks predate this tool,
  rather than emitting a spurious hook-not-found error on every session
  start.

- [x] `test_session_track.sh` — 25/25 pass (re-ran independently above).
- [x] `test_babysit_launch.sh` — 8/8 pass, including new T4 (re-ran
      independently above).
- [x] `ruff check scripts/` clean; `bash -n` clean on all four edited
      scripts (re-verified above).

Escalation: none — confined to `scripts/fleet/`, no engine/render/,
engine/entity/, engine/system/, engine/world/, engine/audio/,
engine/math/, or public `ir_*.hpp` surface touched.

Opus recheck not required.

jakildev:
Addressed feedback (fleet:has-nits, 1 nit):
- `write_worktree_settings` now preserves hand-added `hooks` entries across `fleet-up` regeneration, mirroring the existing `permissions.allow` preservation. The fleet's own SessionStart entry is matched on the `fleet-session-track` marker, so a stale variant of the fleet hook is replaced rather than duplicated; hand-added groups (any event, including extra SessionStart groups) survive wholesale.
- Added `scripts/fleet/tests/test_worktree_settings_hooks.sh` (11 assertions): fresh write installs the guarded hook, hand-added hooks + allow entries survive regen, stale fleet-variant replacement, regen idempotency, malformed-JSON fallback to baseline.
- Verification: new suite 11/11, `test_session_track.sh` 25/25, `test_babysit_launch.sh` 8/8, `ruff check scripts/` clean, `bash -n` clean. — worker-2

PR #2288, opened 2026-07-08
profile: intra-tick GPU sub-stage scopes — wire the reserved voxelStage1 sub-rows (#2280)

### LINKED ISSUE #2280: profile: intra-tick GPU sub-stage scopes — wire the reserved voxelStage1 sub-rows (compact / clear / stage-1 / stage-2)

jakildev:
## Summary
Wires intra-tick GPU sub-stage timing so the opaque `voxelStage1` bundle (~140 ms at zoom 16, #2258) is attributed to its dispatch groups.

- **`GpuSubStageScope`** (`gpu_substage_timing.hpp`, new) — RAII bracket timing one dispatch group inside a system tick, reusing the existing device timestamp-pair machinery (async `kSamplesInFlight` readback, no `finish()` stall).
- **`VOXEL_TO_TRIXEL_STAGE_1`** — untagged from the per-system `GpuStageTimingObserver`; its four dispatch groups now each bracket with a sub-scope, wiring the reserved `canvasClear` / `voxelCompact` / `voxelStage2` rows and narrowing GPU `voxelStage1` to the stage-1 dispatch only. The old whole-tick number = the sum of the four. CPU `voxelStage1` stays whole-tick via an `IR_PROFILE_SCOPE` (no HUD/auto-profile regression).
- **Metal** — a timestamp-aware `createBlitEncoder` (twin of `createComputeEncoder`) captures the distance-clear blit so `canvasClear` is non-zero. Sub-scopes reuse the freed attachment slot (index 0) since the system is untagged — no second attachment index / no dual-attachment quota risk.
- Registry comment + `engine/prefabs/irreden/render/CLAUDE.md` + `docs/design/gpu-stage-timing-cost-model.md` updated for the new attribution.

## #2258 attribution (IRPerfGrid wave scene, sun shadows on, Metal/macOS 2x, `--auto-profile 120`)

| zoom | canvasClear | voxelCompact | voxelStage1 (stage-1 only) | voxelStage2 | sum | old bundled |
|---|---|---|---|---|---|---|
| 8 | 0.14 | 0.71 | 23.71 | 28.14 | ~52.7 | ~47.6 |
| 16 | 0.01 | 0.06 | 64.79 | 76.05 | ~140.9 | ~145 |

The ~140 ms is the **stage-1 + stage-2 dispatch grids** (~evenly split); compact + clear are negligible. Full table posted on #2258.

## Test plan
- [x] Builds clean on macOS/Metal (`IRPerfGrid`, `IRShapeDebug`).
- [x] Non-zero `canvasClear` / `voxelCompact` / `voxelStage2` rows (criterion 1).
- [x] Sum of sub-rows ≈ old bundled `voxelStage1` (criterion 2).
- [x] Attribution table posted on #2258 (criterion 3).
- [x] Rendering **byte-identical timing-on vs timing-off** — 28/28 `IRShapeDebug` shots (criterion 4).
- [ ] OpenGL (Linux/Windows) build + smoke — authored on macOS/Metal; GL path needs cross-host verification (`glQueryCounter` markers nest freely, no structural GL change).

## Follow-up (not in this PR)
`GpuSubStageScope`'s timestamp ring (`nextAvailableSlot` / `resolveReadySamples`) mirrors `GpuStageTimingObserver`'s. Extracting a shared `GpuTimestampRing` would dedupe it, but that touches the observer (critical timing infra for 14 systems) and wants its own verification — deliberately deferred to keep this PR's blast radius on the new file.

Closes #2280

### DIFF
--- a/.fleet/plans/issue-2280.md
+++ b/.fleet/plans/issue-2280.md
@@ -1,0 +1,79 @@
+# Plan: profile — intra-tick GPU sub-stage scopes (wire the reserved voxelStage1 sub-rows)
+
+- **Issue:** #2280 (blocks #2258)
+- **Model:** opus (Metal backend timestamp infrastructure + render hot-path system wiring)
+- **Date:** 2026-07-08
+
+## Scope
+
+An intra-tick scoped GPU timing mechanism so a system can bracket individual
+dispatch groups inside one tick, wiring the reserved `voxelStage1` sub-rows
+(`voxelCompact` / `canvasClear` / `voxelStage2`) on **both** backends, then use
+it to attribute the ~140 ms `voxelStage1` cost that #2258 measured but could not
+localize (the per-`SystemId` timer bundles compact + clear + stage-1 + stage-2).
+
+## Confirmed repro
+
+PR #2266's measurement table (2026-07-06 NEEDS-DESIGN comment): `voxelStage1`
+~139–152 ms at zoom 16 on the IRPerfGrid wave scene (Metal, shadows on),
+invariant to a 256× stage-1 body reduction — the cost is inside the bundled
+per-system scope but not in the stage-1 raster body. The bundle is structural:
+one `GpuStageTimingObserver` pair per `SystemId`
+(`gpu_stage_timing.hpp:196-209`), so compact/clear/stage-1/stage-2 are
+indistinguishable today.
+
+## Approach (picked): RAII sub-stage scope + a second sample-buffer attachment on Metal
+
+1. `RenderDevice` gains scoped-timestamp entry points reusing the existing pair
+   machinery (`createTimestampPair` / `writeTimestamp` / `readTimestampPairMs`),
+   plus a Metal path that attaches the sub-scope's `MTL::CounterSampleBuffer` on
+   **attachment index 1** of the compute-pass descriptor, leaving index 0 to the
+   enclosing per-system pair (`createComputeEncoder`, `metal_render_impl.cpp`).
+   Same sticky first-encoder-claims-start / last-encoder-wins-end semantics as
+   #1746, scoped to the sub-window. OpenGL needs no structural change —
+   `glQueryCounter`-style pairs nest freely in the command stream.
+2. A small `IRRender::GpuSubStageScope(name)` RAII (prefab layer, next to
+   `gpu_stage_timing_observer.hpp`) resolves the name against
+   `gpuStageRegistry()` and brackets the enclosed dispatches; used inside
+   `VOXEL_TO_TRIXEL_STAGE_1`'s tick around its four dispatch groups to fill
+   `voxelCompact`, `canvasClear`, `voxelStage1` (now stage-1 only),
+   `voxelStage2`. Async readback keeps the `kSamplesInFlight` pattern per scope;
+   timers-off cost stays one bool check.
+3. **Fallback (only if the second Metal attachment misbehaves on real
+   hardware):** sub-scopes own attachment index 0 while active and the outer
+   per-system row is computed as the **sum** of sub-scope samples for that
+   system. Document whichever shape lands in the registry comment.
+4. Measure: post the zoom 8 / zoom 16 wave-scene attribution table on #2258 (per
+   this issue's acceptance criteria).
+
+## Acceptance criteria
+
+- GPU STAGES overlay shows non-zero `voxelCompact` / `canvasClear` /
+  `voxelStage2` rows when frame timing is enabled, on both backends.
+- Sum of sub-rows ≈ the old bundled `voxelStage1` value (± measurement overhead)
+  at a test pose on at least one backend.
+- Attribution table for the #2258 scenario (IRPerfGrid default wave scene, zoom 8
+  and zoom 16, sun shadows on, `--auto-profile 100`+) posted on #2258, naming
+  which sub-stage carries the ~140 ms.
+- Byte-identical rendering with timing disabled and enabled (screenshot compare
+  on one demo).
+
+## Sibling / in-flight reconciliation
+
+PR #2266 is being narrowed to the byte-identical `sunBakeFrustumUVBounds`
+refactor only (architect direction on the PR) — no overlap with this surface.
+#2256's plan explicitly does NOT need this issue (its stages are separate,
+already-timed systems). #2258 is blocked by this issue and must not be
+re-planned until the attribution table lands on its thread.
+
+## Gotchas
+
+- Wiring the sub-rows changes the meaning of the `voxelStage1` row (from bundle
+  to stage-1-only) — update the registry comment in `gpu_stage_timing.hpp` and
+  any overlay docs in the same PR.
+- Metal is excluded from clang-format; match surrounding hand-formatting in
+  `metal_render_impl.cpp`.
+- Timing-only change: rendering must stay byte-identical with timers on or off;
+  the timers-off path must keep its one-bool-check cost.
+- GL-backend build/verify needs a Linux/Windows host (this pane is macOS/Metal
+  only); if authored on macOS, the GL side rides as owed cross-host smoke.
--- a/creations/demos/perf_grid/main.cpp
+++ b/creations/demos/perf_grid/main.cpp
@@ -1058,12 +1058,16 @@ void initSystems() {
                             IR_LOG_INFO("Auto-profile CPU-scope — {}: {:.3f}ms", name, ms);
                         }
                     }
+                    // #2280 sub-stage attribution: canvasClear + voxelCompact +
+                    // voxelStage1 (the stage-1 dispatch only now) + voxelStage2
+                    // sum to the old bundled voxelStage1 measurement.
                     IR_LOG_INFO(
-                        "Auto-profile GPU — voxelStage1:{:.3f} voxelStage2:{:.3f} "
-                        "voxelCompact:{:.3f}",
+                        "Auto-profile GPU — canvasClear:{:.3f} voxelCompact:{:.3f} "
+                        "voxelStage1:{:.3f} voxelStage2:{:.3f}",
+                        gpu.canvasClearMs_,
+                        gpu.voxelCompactMs_,
                         gpu.voxelStage1Ms_,
-                        gpu.voxelStage2Ms_,
-                        gpu.voxelCompactMs_
+                        gpu.voxelStage2Ms_
                     );
                     IRWindow::closeWindow();
                 }
--- a/docs/design/gpu-stage-timing-cost-model.md
+++ b/docs/design/gpu-stage-timing-cost-model.md
@@ -32,13 +32,17 @@ changes (occupied-only lists, indirect dispatch, subdivision caps).
   accumulators build the shutdown avg/min/max (#1738). A `finish()`-bracket
   legacy path exists for devices without timestamp support
   (`legacyFinishTiming_`).
-- **Several registry rows have NO writer and always read 0.000:**
-  `canvasClear`, `voxelCompact`, `voxelStage2`, `shapePass0` (folded into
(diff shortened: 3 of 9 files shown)


jakildev (commented):
## Review — profile: intra-tick GPU sub-stage scopes — wire the reserved voxelStage1 sub-rows (#2280)

**Verdict:** approve

Standalone PR (base `master`, MERGEABLE). Adds `GpuSubStageScope` /
`GpuSubStageTimer` (`gpu_substage_timing.hpp`, new) to bracket individual
dispatch groups inside `VOXEL_TO_TRIXEL_STAGE_1`'s tick, wiring the
previously-unwired `canvasClear` / `voxelCompact` / `voxelStage2` registry
rows and narrowing `voxelStage1` to the stage-1 dispatch only. Metal gets a
`createBlitEncoder` twin of `createComputeEncoder` so the distance-clear
blit can carry a timestamp sample.

### Checklist walkthrough

- **Ownership/lifetime** — `GpuSubStageTimer::m_states` is a program-lifetime
  singleton (mirrors the existing `GpuStageTimingObserver` pattern);
  `SubStageState::info_` points into the `static const` registry array
  returned by `gpuStageRegistry()`, so the pointer is stable for the life of
  the program. No raw owning pointers, no dangling references across ticks.
  The "pairs leak at process exit" choice is explicitly justified against the
  #2031 GL-static-destructor precedent — consistent with existing practice.
- **Metal encoder plumbing** — `createBlitEncoder` (`metal_render_impl.cpp:147-178`)
  is a faithful structural twin of the pre-existing `createComputeEncoder`
  (same file, `~L119-142` on base): same descriptor-alloc/attach/release
  shape, same `MTL::CounterDontSample` first-encoder-vs-sticky-end logic. I
  diffed the two side by side against the base commit — no copy-paste drift.
  Both existing `blitCommandEncoder()` call sites are correctly routed
  through it.
- **No nesting risk** — the four sub-scopes in `system_voxel_to_trixel.hpp`
  (`canvasClear`, `voxelCompact`, `voxelStage1`, `voxelStage2`) are used
  sequentially in separate brace blocks, never nested, so there's no
  attachment-slot contention within one tick. `VOXEL_TO_TRIXEL_STAGE_1` is
  correctly untagged from the per-system `GpuStageTimingObserver`
  (`tagGpuStage` call removed), which is what frees the attachment slot for
  the sub-scopes to reuse — confirmed no other code still assumes this
  system is tagged.
- **Math** — `IRMath::clamp` used (not `std::clamp`) for
  `recommendedTimestampPairsInFlight()` clamping.
- **Naming/style** — include guard (`GPU_SUBSTAGE_TIMING_H`) matches sibling
  headers' convention; `m_states` / `m_state` are private members correctly
  prefixed `m_`.
- **Docs** — `gpu_stage_timing.hpp` registry comment, `CLAUDE.md`, and
  `docs/design/gpu-stage-timing-cost-model.md` were all updated in the same
  PR to describe the new per-row attribution (Save/Extensibility-Rule-style
  discipline, appropriately applied here even though this isn't a
  serialization change) — the "bundle → per-dispatch rows" narrative is
  consistent across code and all three docs.
- **Render byte-identity** — PR body claims 28/28 `IRShapeDebug` shots
  byte-identical with timing on vs off; this is exactly the risk this class
  of change (RAII bracket wrapping existing dispatch calls, no control-flow
  change) would threaten if the scope construction/destruction reordered
  anything, so the check is the right one and covers the real risk.

### Nits

- `gpu_substage_timing.hpp` duplicates the timestamp-ring bookkeeping
  (`nextAvailableSlot` / `resolveReadySamples`) already in
  `GpuStageTimingObserver`. The PR body already calls this out as a
  deliberate, scoped-out follow-up (shared `GpuTimestampRing` extraction) —
  agreed that's the right call to keep this PR's blast radius on the new
  file; no action needed here.

- [x] (already covered by PR body) macOS/Metal build + `IRPerfGrid` /
      `IRShapeDebug` byte-identical screenshots, attribution table posted
      on #2258.
- [ ] OpenGL/Linux or Windows smoke — PR body already flags this as
      cross-host-owed; the GL path has no structural change
      (`glQueryCounter` pairs nest freely) but hasn't been run.

Escalation: please Opus-recheck before merge (touches: engine/render/,
Metal GPU timestamp/encoder state machine, intra-tick GPU/CPU sync). I
traced the encoder-attachment logic and found it sound, but the sticky
first-encoder/last-encoder-wins timestamp semantics across sequential
compute+blit encoders is exactly the class of GPU lifetime/sync invariant
this fleet reserves for Opus confirmation rather than Sonnet sign-off alone.

jakildev (commented):
## Opus recheck — profile: intra-tick GPU sub-stage scopes (#2280)

**Verdict:** approve

Sonnet approved and escalated the one thing it correctly declined to
sign off alone: the Metal timestamp/encoder state machine — the sticky
first-encoder-claims-start / last-encoder-wins-end semantics across
sequential compute+blit encoders, and the fact that a `GpuSubStageScope`
reuses the *same single global* attachment slot the per-system observer
uses. I traced that surface end-to-end and confirm it.

### What I confirmed (the Opus-only invariants)

- **Shared-global attachment never dangles.** `GpuSubStageScope`'s
  `begin()`/`end()` are symmetric: `begin()` writes `TimestampSlot::START`
  (setting `g_nextComputeTimestampAttachment`) *only* when it secured a
  ring slot (`activeSlot_ >= 0`); `end()` writes `TimestampSlot::END`
  (clearing the global to `{}`) *only* when `activeSlot_ >= 0`. So
  START-written ⟺ `activeSlot_ >= 0` ⟺ END-written — every open is matched
  by a close, and a scope that can't get a slot touches the global not at
  all. No sub-scope can strand the sticky attachment for the next scope or
  the next system's encoders.

- **Reusing the observer's slot is safe because the system is untagged.**
  The `tagGpuStage(systemId, "voxelStage1")` call is removed and the system
  is deliberately not tagged for `GpuStageTimingObserver`, so its own tick
  leaves the one global slot free. Ticks are CPU-sequential and Metal
  command encoding is single-threaded, so there is no concurrent writer to
  `g_nextComputeTimestampAttachment` during this window.

- **`createBlitEncoder` is a faithful #1746 twin.** Same
  descriptor-alloc / attach / `firstEncoder_ ? startSampleIndex_ :
  CounterDontSample` / `setEndOfEncoderSampleIndex` / release shape as
  `createComputeEncoder` and `attachTimestampSamples`, and it advances
  `firstEncoder_ = false` on the attached path only. The two existing
  `blitCommandEncoder()` call sites in the distance-clear path are both
  routed through it; the null-attachment branch returns a plain encoder, so
  the non-timed path is unchanged. The blit's end boundary is written as
  each encoder is created (not by the trailing `writeTimestamp(END)`), so
  the pair correctly spans [first blit start, last blit end].

- **Sub-scopes are sequential and non-nested.** `canvasClear` /
  `voxelCompact` / `voxelStage1` / `voxelStage2` each live in their own
  brace block, so there is no attachment-slot contention within one tick,
  and the `perAxisSplit` / `skipSingleCanvasVoxels` branches only *skip*
  scopes (RAII keeps each balanced) — no half-open window.

- **Async ring + commit path.** The per-scope `kSamplesInFlight=3` ring
  mirrors the observer exactly (`nextAvailableSlot` skips pending slots,
  returns -1 → graceful no-op frame, no `finish()` stall). The extracted
  `commitGpuStageSample` keeps the same accumulator bounds check the old
  `recordStageSample` had; observer and sub-timer share it with no drift.

- **Lifetime / timers-off.** Program-bound singleton, pairs leak at exit —
  the deliberate choice over a static destructor touching a dead GL context
  (#2031), consistent with the observer. Timers-off costs one
  `timing.enabled_` bool check in the scope ctor.

### One thing I can't confirm from this host (not a blocker)

Whether the Metal blit-pass encoder-boundary timestamp actually *resolves*
a non-zero pair on real hardware (the counter sample buffer must support
stage-boundary sampling on a blit pass). This is empirically covered: the
PR is `fleet:authored-on-macos` and the body reports byte-identical
`IRShapeDebug` screenshots plus an attribution table posted on #2258 —
i.e. it ran on Metal without a validation crash and produced samples. The
documented fallback (sub-scopes own attachment index 0; outer row = sum)
is self-contained if it ever misbehaves, so no architectural follow-up is
gated on this.

The GL/Linux/Windows path has no structural change (`createBlitEncoder`
is Metal-only; GL times the clear via the unchanged `glQueryCounter`
path) and the owed smoke is already tagged `fleet:needs-windows-smoke` —
correct handling.

Approving. Sonnet's checklist pass stands; the GPU lifetime/sync
invariant it reserved for Opus is sound.

🤖 Opus recheck by Claude Opus 4.8 (opus-reviewer, review-pr skill)

PR #229, opened 2026-04-19
fleet: witness health monitoring with heartbeat detection (T-023)

jakildev:
Claiming task. Work in progress.

Closes #193

PR #2289, opened 2026-07-08
fleet-claim: reject claiming a closed issue + trailing --repo flag

jakildev:
## Summary
- Guard 1: `cmd_claim` / `cmd_planning_claim` refuse a **CLOSED** issue via a new `_refuse_if_closed` helper. A stale scout cache can list a closed issue in `tasks.open[]` / `needs_plan[]`; claiming it strands a `fleet:claim-*` label (release also retains it). Gate runs first in the claim path, before any dir/label is created.
- Guard 2: a `--repo` token placed **after** the subcommand is now a hard error (exit 2). `--repo` is a global flag parsed before the subcommand; trailing it (e.g. `fleet-claim release 45 --repo game`) previously released the engine `#45` claim instead of `game-45`. One centralized check before dispatch; `cleanup`/`reconcile` (which take their own trailing `--repo <owner/repo>`) are exempt.
- New `scripts/fleet/tests/test_fleet_claim_safety_guards.sh` (20 assertions, hermetic `gh` stub).

## Test plan
- [x] `bash scripts/fleet/tests/test_fleet_claim_safety_guards.sh` — 20/20 pass: CLOSED claim + planning-claim refusal (+ no dir/lock created), trailing `--repo` on claim/release → exit 2, `--repo=` equals form, leading `--repo` + `--stackable-on <url>` no-regression, cleanup/reconcile exemptions.
- [x] All 16 existing `test_fleet_claim_*.sh` suites pass (no regression from the pre-dispatch Guard 2 preamble).
- [x] `bash -n scripts/fleet/fleet-claim` clean.

## Notes for reviewer
Both guards soft-degrade correctly: Guard 1 refuses only a confirmed `CLOSED` and falls open on any `gh` lookup miss (absent gh, unknown namespace, network error), matching `check_blockers`' fail-open contract so a flaky `gh` never blocks every claim. Guard 2 sits after the `FLEET_CLAIM_LIB` sourcing seam, so tests that source the script as a library are unaffected, and matches the `--repo` token exactly/prefix (not substring) so a `--stackable-on <url>` value containing "repo" isn't misflagged.

Closes #2201

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2140, opened 2026-06-30
render: #2082 sun-shadow bake close-out — verify-and-document (C2)

jakildev:
**#2082 (epic #1717 child C2) — RE-SCOPED to a verify-and-document close-out** by architect direction on this PR (2026-07-06). Docs + evidence only, **zero shader / C++ changes**.

## Why re-scoped
The literal C2 direction ("extract `scatterAnalyticEdgeCoverage` into `ir_iso_common`, call it from the bake; thread per-axis frac bits [9:2]") failed verify-first (NEEDS-DESIGN F1–F4, architect-accepted):
- **F1** — `scatterAnalyticEdgeCoverage` is Metal-only; the GL scatter port is the independent unstarted #1938. Nothing to extract on GL.
- **F2** — the bake is a compute pass (one `atomicMin`/canvas pixel); the fragment helper's inputs (`fwidth`, footprint param, neighbour flags) don't exist there. "Call from the bake" = a new mechanism, not helper reuse.
- **F3** — the per-axis resolve already footprint-fills (#1724) and its cardinal layout is the bake's exact inverse (cast == receive, #1380-safe); frac threading contradicts that invariant.
- **F4** — #1784 scope overlap → #1784 closed as stale.

Decisive cross-lane fact: **PR #2204's deterministic A/B proved a bake-side footprint splat is a no-op** (forced radius 12 byte-identical) on resolve-fed paths. There is no live bake-coverage gap to fix.

## Deliverable
- **`docs/design/per-axis-sun-shadow-resolve.md`** — (a) the resolve fill IS the bake-coverage mechanism (no per-pixel bake splat exists or is needed on resolve-fed paths); (b) the accepted ~1-cell per-axis silhouette residual (#1883 precedent); (c) shadow-metric acceptance gates require a pinned deterministic pose (`--no-auto-rotate --no-spin` [+ `--frozen-pose`]).
- **Deterministic captures** under `docs/pr-screenshots/claude/2082-bake-analytic-coverage/` — canvas_stress shadow overlay (cardinal + `--frozen-pose 0.6`) + IRVoxelYaw pure-GRID (yaw0/30/45 + zoom8 stair) + the #2204 splat A/B.
- **`.fleet/plans/issue-2082.md`** — the re-scope of record.

## Evidence-run outcome
[...]

PR #2290, opened 2026-07-08
ci: share auto-rereview flow + skip mechanical rebases (game #229)

jakildev:
## What

Adds `.github/workflows/auto-rereview.yml` to the engine repo and fixes it so a **mechanical rebase no longer triggers a re-review**. Kept byte-identical with the game repo's copy (jakildev/irreden) so both repos share one flow.

## Why

On game [PR #229](https://github.com/jakildev/irreden/pull/229) the fleet merger re-targeted a stacked PR onto master and force-pushed the replayed child commits. `role-merger.md` guarantees verdict labels survive that mechanical rebase, but the game's `auto-rereview.yml` fired on the `synchronize` and stripped `fleet:approved` → `human:re-review` anyway. Timeline:

```
head_ref_force_pushed        (merger rebase)
unlabeled fleet:approved   } auto-rereview (github-actions[bot])
labeled   human:re-review  }
labeled   fleet:stacked-rebase   (merger — 3s TOO LATE)
```

A `fleet:stacked-rebase` label guard can't fix this: the merger applies the label *after* it pushes, and the `synchronize` payload captures labels at push time.

## Fix

The workflow now classifies the push by comparing the PR's **net-diff patch-id** (`git diff base...head` reduced by `git patch-id --stable`, which is invariant across rebase/SHA churn) before vs after:

- identical → mechanical rebase → **skip** (verdict preserved)
- changed → real edit or semantic rebase (master altered the effective diff) → re-review as before

Race-free and needs no coordination with the merger's labeling order. If the pre-push commit is unfetchable (rare orphan case), it conservatively falls back to re-review.

### DIFF
--- a/.github/workflows/auto-rereview.yml
+++ b/.github/workflows/auto-rereview.yml
@@ -24,15 +24,19 @@ name: Auto re-review on push to approved PR
 #   on a `fleet:stacked-rebase` label guard because the merger applies
 #   that label *after* it pushes — the synchronize event payload
 #   captures labels at push time and races ahead of it. Instead we
-#   compare the PR's net-diff patch-id (git diff base...head, reduced by
-#   `git patch-id --stable`, which is invariant across rebases/SHAs)
-#   before vs after the push. Identical patch-id => mechanical rebase =>
+#   compare the PR's net-diff patch-id (reduced by `git patch-id --stable`,
+#   which is invariant across rebases/SHAs) before vs after the push, each
+#   tip against its *own* base. Identical patch-id => mechanical rebase =>
 #   skip. A rebase where master actually altered the effective diff (a
-#   *semantic* rebase) changes the patch-id and still triggers review.
+#   *semantic* rebase) changes the patch-id and still triggers review. The
+#   classify logic lives in `scripts/fleet/classify-auto-rereview.sh` (with a
+#   test suite at `scripts/fleet/tests/test_classify_auto_rereview.sh`) so it
+#   is unit-testable against a git sandbox rather than buried in a `run:` block.
 #
-# This file is kept byte-identical between the engine repo
-# (jakildev/IrredenEngine) and the game repo (jakildev/irreden) so both
-# repos share one re-review flow. Edit both when changing it.
+# This file and `scripts/fleet/classify-auto-rereview.sh` are kept
+# byte-identical between the engine repo (jakildev/IrredenEngine) and the game
+# repo (jakildev/irreden) so both repos share one re-review flow. Edit both
+# repos (and both files) when changing it.
 
 on:
   pull_request:
@@ -60,41 +64,9 @@ jobs:
           AFTER: ${{ github.event.after }}
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
         run: |
-          set -uo pipefail
-
-          # Make sure the three commits we reason about are present locally.
-          # `after` and the base tip come with the PR checkout; `before` is
-          # orphaned by a force-push and may or may not be fetchable.
-          git fetch --no-tags --quiet origin "$AFTER" "$BASE_SHA" 2>/dev/null || true
-          git cat-file -e "${BEFORE}^{commit}" 2>/dev/null \
-            || git fetch --no-tags --quiet origin "$BEFORE" 2>/dev/null || true
-
-          # Net content of a tip = diff from its merge-base with the current
-          # base branch, hashed by patch-id (stable across rebase/SHA churn).
-          net_patch_id() {
-            local tip="$1" mb
-            mb=$(git merge-base "$BASE_SHA" "$tip" 2>/dev/null) || return 1
-            git diff "$mb" "$tip" | git patch-id --stable | awk '{print $1}'
-          }
-
-          if ! git cat-file -e "${BEFORE}^{commit}" 2>/dev/null; then
-            echo "before commit $BEFORE unavailable — cannot prove a rebase; re-reviewing."
-            echo "rebase_only=false" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-
-          OLD_ID=$(net_patch_id "$BEFORE") || OLD_ID=""
-          NEW_ID=$(net_patch_id "$AFTER") || NEW_ID=""
-          echo "old net patch-id: '${OLD_ID:-<empty>}'"
-          echo "new net patch-id: '${NEW_ID:-<empty>}'"
-
-          if [ -n "$OLD_ID" ] && [ "$OLD_ID" = "$NEW_ID" ]; then
-            echo "Net content diff unchanged — mechanical rebase. Skipping re-review."
-            echo "rebase_only=true" >> "$GITHUB_OUTPUT"
-          else
-            echo "Net content diff changed — real update. Re-reviewing."
-            echo "rebase_only=false" >> "$GITHUB_OUTPUT"
-          fi
+          # The script prints exactly `rebase_only=true|false` on stdout (routed
+          # to $GITHUB_OUTPUT); its diagnostics go to stderr (the Actions log).
+          bash scripts/fleet/classify-auto-rereview.sh >> "$GITHUB_OUTPUT"
 
       - name: Swap fleet:approved → human:re-review
         if: steps.classify.outputs.rebase_only != 'true'
--- a/.github/workflows/auto-rereview.yml
+++ b/.github/workflows/auto-rereview.yml
@@ -1,0 +1,110 @@
+name: Auto re-review on push to approved PR
+
+# Fires when commits are pushed to a PR that already has fleet:approved.
+# Swaps the label to human:re-review so the fleet reviewers pick it up
+# again on their next poll. This means the human can edit an approved
+# PR and just `git push` — no skill invocation needed.
+#
+# Why filter on fleet:approved only (not all fleet verdict labels):
+#   - During fleet:needs-fix, the author agent is addressing feedback
+#     and pushing fixes via commit-and-push. Auto-swapping there would
+#     fight the fleet:changes-made label the author sets.
+#   - During fleet:wip, the author is actively working. No verdict yet.
+#   - fleet:approved is the only state where a human-pushed change
+#     unambiguously means "I edited after fleet was done — please look
+#     again."
+#
+# Why the mechanical-rebase guard (the `classify` step below):
+#   A push whose *content* is unchanged does not need re-review. The
+#   fleet merger re-targets a stacked PR onto master and force-pushes
+#   the replayed child commits (see role-merger.md — "verdict labels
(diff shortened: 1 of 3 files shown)


jakildev (commented):
## Review — ci: share auto-rereview flow + skip mechanical rebases (game #229)

**Verdict:** needs-fix

Standalone PR (base `master`, MERGEABLE). Single new file,
`.github/workflows/auto-rereview.yml` — adds the mechanical-rebase guard
(previously game-repo only) to the engine so both repos share one flow,
plus the classify logic fixing the "stacked-PR retarget strips
`fleet:approved`" bug from game PR #229.

### Needs-fix

- `.github/workflows/auto-rereview.yml:74-97` — the patch-id classification
  does not actually detect the PR's own motivating scenario (a stacked
  child PR retargeted from a parent's branch onto `master` after the
  parent merges) when the parent PR is **squash-merged**. `net_patch_id`
  computes `mb = merge-base($BASE_SHA, $tip)`, using the *current*
  `BASE_SHA` (`github.event.pull_request.base.sha`, already updated to
  master post-retarget) for **both** the before-tip and after-tip. For
  `AFTER` (rebased directly onto master) this correctly yields
  `mb = master_tip`, isolating the child's own diff. For `BEFORE`
  (still on the pre-retarget parent branch), if the parent's original
  commits were squash-merged, they no longer exist as ancestors of
  master — `merge-base($BASE_SHA, $BEFORE)` falls back to wherever the
  child branch and master last shared history, which is *before* the
  parent branch started. `diff(mb, $BEFORE)` then includes the parent's
  entire diff in addition to the child's, so `OLD_ID != NEW_ID` even
  though the net change is a no-op rebase — the workflow still strips
  `fleet:approved` and re-tags `human:re-review`, reproducing exactly the
  bug this PR sets out to fix.

  This repo's own master history confirms squash-merge is the prevailing
  strategy (single commits like `10fa23de render: extract
  sunBakeFrustumUVBounds shared helper (#2258 prep) (#2266)` — the PR's
  own branch commit is a different SHA, `15ba1313`, than what landed on
  master), and the state cache shows exactly this stacked-retarget
  pattern happening repeatedly (e.g. #2083/#2275: "retargeted from
  `claude/2082-bake-analytic-coverage` to `master` and rebased with
  `--force-with-lease`" right after #2140 merged).

  I reproduced this in an isolated sandbox: master, a parent branch off
  master, a child branch off the parent's tip, squash-merge the parent,
  then `git rebase --onto master parent child` (mirroring the merger's
  actual retarget). Running the classify step's exact logic against that
  graph:
  ```
  OLD_ID=60d22f01ee...  NEW_ID=25f4da9b2c...  -> MISMATCH
  ```
  Confirmed misclassified as "real content change" despite the child's
  own diff being byte-identical before and after.

  Suggested direction: the current approach can't recover the child's
  *actual* pre-retarget base once the parent's commits are gone from
  history — comparing against a single "current" `BASE_SHA` for both
  sides is the structural problem, not a tunable. Consider capturing the
  pre-retarget base explicitly (GitHub's `pull_request` `edited` event
  carries `changes.base.sha.from` when the base changes) and persisting
  it (e.g. a hidden PR comment marker) for the subsequent `synchronize`
  handler to read, rather than deriving it after the fact via
  `merge-base`.

### Nits

- No test coverage for the classify bash logic itself — this repo's
  `scripts/fleet/` convention extracts non-trivial bash into a script
  with a `test_*.sh` suite (see PR #2289, reviewed just before this one).
  The logic embedded in a `run:` block here is exactly the kind of thing
  that silently breaks (as demonstrated above) with no CI signal.
  Extracting `net_patch_id`/the classify logic to a standalone script
  callable from a local git sandbox would have caught this before merge.

### Praise

- The PR body's own worked example (the exact `unlabeled fleet:approved
  } auto-rereview / labeled human:re-review } / labeled
  fleet:stacked-rebase (3s TOO LATE)` timeline) is a genuinely useful,
  concrete repro of the race it's fixing — this made verifying the
  classify logic against that specific case straightforward.
- Correctly falls back to conservative re-review (`rebase_only=false`)
  when the `before` commit is unfetchable, rather than silently assuming
  "mechanical" on missing data.

- [ ] Fix the classify logic to correctly handle the retarget-after-squash-merge
      case (the PR's own headline scenario), and verify against a sandbox
      repro like the one above.
- [ ] Confirm the in-place (non-retargeted) rebase-to-catch-up-with-master
      case still classifies correctly — that path does work with the
      current logic since the old base commit remains an ancestor of the
      new base.

jakildev:
Addressed fleet:needs-fix (commit 7ecd30b2):

**needs-fix — classify misclassifies retarget-after-squash-merge:** confirmed and fixed. The old `net_patch_id` computed each tip's net diff against `merge-base(BASE_SHA, tip)` using the *current* `BASE_SHA` for **both** sides — so once a stacked child's parent squash-merges (parent commits gone from history), `merge-base(BASE_SHA, BEFORE)` falls back past the parent's fork point and folds the parent's whole diff into the child's net diff. I reproduced the exact bug on a sandbox graph (parent squash-merged, child `rebase --onto master parent child`): the old logic yields `OLD_ID != NEW_ID` → wrongly strips `fleet:approved`.

Rather than persist the pre-retarget base via an `edited`-event marker (the suggested direction), I recover it **structurally** — avoids cross-event state and the edited-vs-synchronize ordering race: the child's own commit count `N = commits in BASE_SHA..AFTER` (AFTER is the child replayed directly onto the current base), so BEFORE's own base is `BEFORE~N` (the old parent tip, whatever it was). Diffing each tip against its *own* base yields the child's net diff on both sides → patch-id-equal for a mechanical rebase, unequal for a real edit. Directly resolves the "single current BASE_SHA for both sides is the structural problem" root cause.

**nit — no test coverage for the classify bash logic:** extracted the logic to `scripts/fleet/classify-auto-rereview.sh` (referenced from the `run:` block, single source of truth) with a hermetic git-sandbox suite `scripts/fleet/tests/test_classify_auto_rereview.sh` — 5/5 green, covering the two test-plan items plus fallbacks:
- T1 retarget-after-squash-merge → `rebase_only=true` (the headline bug)
- T2 in-place catch-up rebase, child diff unchanged → `rebase_only=true` (stays correct)
- T3 real content change → `rebase_only=false`
- T4 BEFORE unavailable → `rebase_only=false` (conservative)
- T5 no-op force-push (same tree, new SHA) → `rebase_only=true`

Separately verified the *old* logic MISMATCHes the T1 graph, so the test genuinely catches the regression.

Note: `.github/workflows/auto-rereview.yml` + the new script are kept byte-identical with the game repo — the game-side twin **#233** needs the same extraction+fix to stay in sync (its inline logic still has the bug). — worker (opus, macOS)

jakildev (commented):
## Review — ci: share auto-rereview flow + skip mechanical rebases (game #229)

**Verdict:** approve

### Prior-review resolution
- ✅ `.github/workflows/auto-rereview.yml:74-97` (at prior review commit `3875de15`) — classify misclassified a stacked child PR retargeted onto `master` after its parent **squash-merges** (the PR's own headline bug: `merge-base($BASE_SHA, $BEFORE)` fell back past the squashed parent's fork point and folded the parent's whole diff into the child's "net" diff) — verified fixed at `7ecd30b2`. The logic now recovers each tip's own base structurally (`N = commits in BASE_SHA..AFTER`; `OLD_BASE = BEFORE~N`) instead of diffing both tips against a single current `BASE_SHA`, and diffs each tip against its own base. I re-ran the reproduction scenario from my prior review (sandbox: parent branch off master, child off parent's tip, squash-merge parent, `git rebase --onto master parent child`) against the new `scripts/fleet/classify-auto-rereview.sh` and it now correctly emits `rebase_only=true`.
- ✅ `.github/workflows/auto-rereview.yml` (nit) — no test coverage for the classify bash logic — verified fixed at `7ecd30b2`. The logic is now extracted to `scripts/fleet/classify-auto-rereview.sh` with a hermetic test suite at `scripts/fleet/tests/test_classify_auto_rereview.sh` (5 cases: retarget-after-squash, in-place catch-up rebase, real content change, unfetchable BEFORE, no-op force-push). I ran the suite myself — `PASS=5 FAIL=0`.

- Re-read `scripts/fleet/classify-auto-rereview.sh` end to end and traced the base-recovery math: `N` is measured on the side that's cleanly based on the current base (`AFTER`, just replayed by `rebase --onto`), then applied to `BEFORE` via `BEFORE~N` to recover the pre-retarget parent tip. This only depends on the child's own commit count being identical across `BEFORE`/`AFTER` (true for a pure rebase replay), and fails closed (`rebase_only=false`, i.e. re-review) whenever that assumption can't be verified — `BEFORE` unfetchable, `AFTER` has 0 commits above base, or `BEFORE` has fewer than `N` ancestors (T3/T4 cover the "real edit landed alongside the push" case, which naturally can't satisfy the ancestor-count check and correctly falls through to re-review).
- Ran `bash scripts/fleet/tests/test_classify_auto_rereview.sh` directly against the checked-out branch (not just trusting the author's comment) — `PASS=5 FAIL=0`, including T1, the exact scenario this PR exists to fix.
- Re-read `.github/workflows/auto-rereview.yml` — the `Classify push` step now just calls the extracted script with the three event SHAs as env vars and forwards its `rebase_only=...` stdout line to `$GITHUB_OUTPUT`; the label-swap step is unchanged from the pre-fix version. Wiring is correct.
- Confirmed both new files ship the `755` executable bit (`create mode 100755` in the diff) and are called via `bash <path>` in the workflow, so the bit isn't load-bearing but is consistent with the rest of `scripts/fleet/`.
- `scripts/fleet/CLAUDE.md`'s hermetic-test rule ("no live GitHub, no live `~/.fleet`") is satisfied — the test sandbox never configures an `origin` remote, and the script's `git fetch origin` calls are `2>/dev/null || true` best-effort, confirmed by the tests passing with no network.

No new findings from the second commit — it's scoped exactly to the classify fix + the extraction + the tests, consistent with the PR body and the author's comment.

### Praise

- The author's PR comment walks through the exact structural fix and why the alternative (persisting a pre-retarget base marker via the `edited` event) was rejected — cross-event state avoided. Matches what I verified independently in the script.
- Test T3 (real content change via `commit --amend`) and T4 (unfetchable BEFORE) exercise exactly the failure modes that make a naive base-recovery scheme unsafe, not just the happy path.

- [x] `bash scripts/fleet/tests/test_classify_auto_rereview.sh` — 5/5 green (verified by this review).
- [ ] The author's own comment flags that the game repo's twin, PR jakildev/irreden#233, still carries the old buggy inline logic and needs the same extraction+fix to stay byte-identical — out of scope for this PR's merge but worth tracking so the two repos don't drift.

Escalation: none. Fleet CI tooling only (`.github/workflows/`, `scripts/fleet/`) — no engine render/entity/system/world/audio/math/lifetime/concurrency surface touched. Opus recheck not required.

Issue #1234, opened 2026-05-27
fleet-claim: accept issue numbers, drop TASKS.md reads, remove master_lock_task


jakildev:
**Part of:** TASKS.md elimination (parent #1216).
**Model:** opus (core fleet logic, several interlocking refactors)
**Blocked by:** (none — but conceptually paired with the orphan-script deletion)
**Unblocks:** TASKS.md deletion

## Context

`fleet-claim` is the central task-claiming primitive. It currently:
1. Accepts `T-NNN` task IDs and looks them up in TASKS.md.
2. Pushes `[ ] → [~]` to TASKS.md on master via pure git plumbing (`master_lock_task`) for cross-host race resolution.
3. Reads TASKS.md to gate model affinity, parse blockers, and find stackable parents.

With TASKS.md being deleted and the scout already reading from GitHub Issues, `fleet-claim` needs to be rebuilt around issue numbers + labels.

The fleet is paused, so we can do a hard interface break (no `T-NNN` shim).

### Interface change (issue #1216 §5a)
- `fleet-claim claim <issue#> <agent>` (was: `fleet-claim claim "T-NNN" <agent>`)
- `fleet-claim stack "<n1> <n2> <n3>" <agent>` (was: `T-NNN T-NNN T-NNN`)
- Reject any `T-\d+` input with a clear "use the issue number directly, e.g. #1234" migration hint.

### Code to delete (from `scripts/fleet/fleet-claim`)
- `master_lock_task()` and the entire master-push path (lines ~710–900, callers at ~1032, ~1099, ~2099). Cross-host race resolution drops to the issue-label tie-break layer alone (already in place; see FLEET.md §"Multi-host fleet coordination").
- `check_blockers()` body (lines ~343–518) — replace with `gh issue view <blocker#> --json state,labels` style queries.
- `check_model_tag()` (lines ~519–585) — read model affinity from issue labels (`fleet:opus`, `fleet:sonnet`); fall back to `**Model:**` body field per the same parser already in `fleet-state-scout`'s `fetch_task_queue`.
- `cmd_reclaim()` (lines ~1250–1340) — TASKS.md-row reclamation no longer applies.
- `cmd_find_stackable_blockers()` (lines ~1150–1238) — replace TASKS.md scan with GitHub issue links / PR API.

### Tests to update
[...]

PR #2299, opened 2026-07-08
fleet: reject (#N prep) prep-marker titles in scope-shipped ingest

jakildev:
## Summary
- The scope-shipped ingest predicate (`scripts/fleet/fleet_scope_shipped.py`) trusts any word-boundary `#N` in a non-plan PR **title**. It already guards range-endpoints (layer 3), plan-doc titles (layer 4), code-span verbs (layer 5), and **deferral** markers `(#N deferred)` (layer 6) — but had no guard for a **prep/partial** marker.
- PR #2266 (`render: extract sunBakeFrustumUVBounds shared helper (#2258 prep)`) was a narrowed prep refactor, linked `Part of #2258` (not `Closes`), whose body states it "keeps only the independently-correct refactor and drops the dead plumbing". Its title `#2258` tripped title-trust, so ingest false-stamped `fleet:scope-shipped` on the still-open perf issue #2258 — 55s after a human cleared its v2 plan-review, clobbering the re-queue.
- Adds **layer 7**: a `(#N prep)` / `prep for #N` / `preparatory #N` title ref is an explicit not-yet-shipped signal and is no longer trusted (a genuine body `Closes #N` still ships it). Generalizes `_ref_is_deferred` → `_ref_is_nonship_marked` covering both the deferral and prep vocabularies; the `prep for #N` connector is handled in the leading form.

## Test plan
- [x] `python3 -m unittest tests.test_scope_shipped_reference` — 63 pass (adds the #2258←#2266 real-world case, `prep for`/`preparatory` variants, the `prepend` prefix-word guard, and the adjacency-suppression trade-off case).
- [x] `ruff check scripts/fleet/fleet_scope_shipped.py scripts/fleet/tests/test_scope_shipped_reference.py` — clean.

Context: #2258 (the mislabeled issue) has already had `fleet:scope-shipped` removed and `fleet:queued` restored by hand; this PR closes the tooling gap so it can't recur.

### DIFF
--- a/scripts/fleet/fleet_scope_shipped.py
+++ b/scripts/fleet/fleet_scope_shipped.py
@@ -62,14 +62,30 @@ incidental match have to be rejected:
    deferral word must sit directly on one side of the ref, within a bounded gap
    like the range/verb guards, so "fix #1234 and defer #1235" still ships #1234).
 
+7. Prep / partial marker (#2258 ← #2266). A PR may name ``#N`` in a trusted
+   (non-plan) title only to mark it *preparatory* — a narrowed refactor that
+   lands groundwork for #N without shipping its scope: "render: extract
+   sunBakeFrustumUVBounds shared helper (#2258 prep)" is linked ``Part of``,
+   not ``Closes``, and its body states it "keeps only the independently-correct
+   refactor and drops the dead plumbing". Because it is titled ``render:`` (not
+   ``docs:``) the layer-4 plan-doc guard does not fire, and the bare ``#2258``
+   in the title satisfies title-trust, so ingest false-stamped
+   ``fleet:scope-shipped`` on #2258 — clobbering the issue's re-queue the instant
+   its plan cleared review. A ``#N`` carrying an adjacent prep word ("(#N prep)",
+   "prep for #N", "preparatory #N") is an explicit not-yet-shipped signal, so a
+   prep-marked title ref is NOT trusted — same bounded-gap rule as layer 6's
+   deferral marker (the prep word must sit directly on one side of the ref, so
+   "ship #1234, prep #1235" still ships #1234).
+
 So a body ``#N`` counts only when a closing-action verb sits directly before it
 in prose (``Closes #N``, ``fixes (#N)``, ``supersedes #N``) — code spans are
 stripped first (layer 5), so a verb quoted in backticks does not count. A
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: reject (#N prep) prep-marker titles in scope-shipped ingest

**Verdict:** approve

CONFLICTING state checked — N/A (`mergeable` is `MERGEABLE`).

No inline PR comments to cross-reference. Verified the diff is scoped to
exactly the two files the body describes (`fleet_scope_shipped.py` +
`tests/test_scope_shipped_reference.py`) — no out-of-scope churn.

Checked correctness of the layer-7 regex generalization by hand-tracing
the trickiest cases, then confirmed by running the suite myself:
- `_ref_is_deferred` → `_ref_is_nonship_marked`: the `\b(?:...)\b` wrapping
  around the alternation is correctly grouped, so `prep(?:s|ping|aratory)?`
  doesn't leak precedence into the surrounding `defer(?:...)` alternative.
- The `\b` anchor after the marker-word alternation correctly rejects
  "prepend" (no word boundary after bare "prep" mid-word) while still
  matching "preparatory" and "preps".
- The "prep for #N" leading-form connector (`(?:\s+for)?`) is correctly
  scoped to sit only between the marker word and the gap, so it can't be
  abused to bridge an unrelated ref ("prep the sun bake, then land #1234"
  still ships #1234, per the adjacency test).
- `select_shipped_pr` / `pr_references_issue` have no other callers in
  `scripts/` besides `fleet-queue-ingest`'s inline heredoc import — the
  rename of the private helper (`_ref_is_deferred` →
  `_ref_is_nonship_marked`) has no dangling references anywhere in the tree.

Ran both locally:
- `python3 -m unittest tests.test_scope_shipped_reference` — 63/63 pass.
- `ruff check scripts/fleet/fleet_scope_shipped.py scripts/fleet/tests/test_scope_shipped_reference.py` — clean.

### Praise
- The real-world regression test (`test_real_world_2258_prepped_by_2266`)
  pins the exact incident that motivated this layer, matching the module's
  existing convention (layers 1-6 all cite a real `#A ← #B` pair) — keeps
  the docstring and test suite from drifting apart as more layers accrete.

- [x] Already run and confirmed above — no further action needed.

Opus recheck not required.

Issue #2153, opened 2026-06-30
fleet: opus-reviewer recheck verdict misroutes to the wrong PR (approve body + spurious needs-fix)


jakildev:
## Symptom

The opus-reviewer's recheck verdict has landed on the **wrong PR** at least twice, each time spending a worker iteration on a no-op "fix":

- **PR #2141** (fleet-dispatch-wrap session-resume): the only review (`COMMENTED`, 19:19:16Z) was the **Opus recheck of the PR fixing issue #2032** (`render: fix skeletal seedVoxelBoneSlots OOB span deref`) — carrying an **approve** verdict for a render/skeletal change unrelated to #2141. `fleet:needs-fix` was applied 7s later (19:19:23Z). A worker (opus) was then dispatched, diagnosed the misfire, opus-reviewed the diff sound, and dropped the spurious label.
- **PR #2138** (fog cross-section): same #2032 recheck body misrouted earlier (worker noted "2nd occurrence, after #2138").

(#2032 is an **issue**, not a PR — the misrouted body referenced it as the issue its real target PR closes.)

## Why it matters

A misrouted verdict is internally **contradictory** — an *approve* review body but a `fleet:needs-fix` label — yet it makes the resolver elect the PR for the worker lane, so a fresh (opus) worker is dispatched, reads the bogus feedback, finds nothing to fix, and exits. This is a recurring source of the "worker spins up and finds no work" no-op dispatches.

## Likely root cause

The opus-reviewer agent reviewed PR-fixing-#2032 but applied the comment + `fleet:needs-fix` to a *different* PR number (#2141 / #2138). The role doc's instructions are correct (`gh pr edit <N>` / `gh pr comment <N>`); the agent crossed PR numbers in flight (it processes a recheck list + a plan-review list in one iteration). No deterministic seam currently prevents posting a verdict to a PR the reviewer never reviewed.

## Proposed deterministic guard (claim-verified verdict apply)

The reviewer already takes an atomic `fleet:reviewing-<host>-<agent>` claim on the PR it reviews (role-opus-reviewer.md step 2). Make the verdict apply go through a small wrapper that **refuses to apply a verdict to a PR this agent does not hold the reviewing claim on**:

- `fleet-review-verdict <pr> <needs-fix|approve|...> <agent>` — verifies `fleet:reviewing-<thishost>-<agent>` is present on `<pr>` before applying the verdict label / posting; exit non-zero otherwise.
[...]

PR #2305, opened 2026-07-08
script: deferred entity create/destroy binding for Lua EVAL systems (#2286)

### LINKED ISSUE #2286: script: no deferred entity create/destroy binding for Lua EVAL systems
**Area:** engine/script
**Model:** opus
**Blocked by:** (none)

## Context

Discovered while implementing game #224 (a debug slow-zone modifier
source in `irreden/demos/player_control`). The task's plan called for
creating a short-lived source entity on zone-enter and destroying it on
zone-exit, from inside an EVAL-mode Lua system tick — the standard
"Ongoing effect" shape documented in `irreden/docs/effects.md` §Source
pattern.

There is currently **no Lua binding for deferred entity create/destroy**.
The full `IREntity` Lua surface (`engine/script/src/lua_script.cpp:522-634`)
exposes `addLuaComponent`, `getLuaComponent`, `removeLuaComponent`,
`hasLuaComponent`, `bindPoint`, `singleton`, `getLuaField`, `setLuaField` —
no `create`, `destroy`, or `deferred*` entries. This matches a known,
already-documented gap: `docs/design/lua-driven-ecs.md` "G4 — structural
change in a Lua tick" states *"There is no add/remove-component
mid-Lua-tick today... the machinery exists; only the Lua binding is
missing,"* with the decision recorded as "bind the deferred ops to Lua"
but not yet implemented.

The underlying C++ deferred ops exist
(`engine/entity/include/irreden/entity/entity_manager.hpp`:
`setComponentDeferred`, `removeComponentDeferred`, `markEntityForDeletion`)
but have no Lua-callable wrapper. Any EVAL Lua system that wants to
create or destroy an entity mid-tick — for a genuinely dynamic source
population (multiple concurrent zone instances, spawned projectiles,
transient VFX sources) rather than a single pre-minted singleton — has
no safe path today.

Game #224 worked around this by minting the source entity once at load
time (mirroring the existing Sprint/Momentum/Dodge source pattern in the
same file) instead of creating/destroying per zone-enter/exit — sufficient
for a single-instance debug scaffold, but not for any future mechanic
that needs N concurrent dynamically-lived sources from Lua.

## Acceptance criteria

- `IREntity.deferredCreate(...)` (or equivalent) callable from an EVAL
  Lua system tick, supporting attaching both C++ components and
  Lua-registered (`IRComponent.register`) components to the new entity.
- `IREntity.deferredDestroy(eid)` callable the same way.
- Both flush via the existing `flushStructuralChanges` pipeline step,
  matching the C++ system convention.
- A short EVAL-mode test or demo update exercising create+destroy
  mid-tick without corrupting archetype iteration.

jakildev:
## Summary
- Adds `IREntity.deferredCreate` / `IREntity.deferredDestroy` Lua bindings so an EVAL Lua system tick can spawn/destroy entities without corrupting the live archetype iteration (the G4 gap in `docs/design/lua-driven-ecs.md` — deferred machinery existed, Lua binding was missing).
- `deferredCreate(componentList?)` reserves an `EntityId` immediately (returned to the tick) and stages the archetype insert + component attach for the next `flushStructuralChanges` (group boundary). `deferredDestroy(eid)` is `markEntityForDeletion`, drained by `destroyMarkedEntities` (pipeline end) — matching the C++ system convention.
- New `EntityManager::createEntityDeferred()` (main-thread analog of the worker-thread reserve-then-stage `createEntity` path) and `EntityManager::stageStructuralChange(fn)` (generic staging entry point so the script layer can marshal `sol::table` overrides in a lambda the entity module can't depend on the type of).
- Component attach reuses a shared `attachDynamicComponent` helper (now also backing `addLuaComponent`).
- Attaches any component addressable from Lua by `componentId` (Lua-registered / codegen'd, shared ComponentId space). Native C++-only components with no Lua default-row support keep using the templated `createEntity` — same boundary as the existing `addLuaComponent`.

## Test plan
- [x] `test/script/lua_deferred_entity_test.cpp` (4 EVAL cases): reserve-then-flush create, bare-entity create, destroy-on-drain, and the acceptance case — a tick spawns a child per iterated entity and reaps the iterated entity; asserts iteration uncorrupted (3 visited, 3 spawns with copied payloads, 0 drivers) after flush + `destroyMarkedEntities`.
- [x] Full suite green on macOS/Metal (`IrredenEngineTest`: 1321 passed, 1 skipped GL-only); `addLuaComponent` coverage (`LuaComponentTest`, `LuaSystemRegisterTest`) still passes through the shared helper.

Notes for reviewer:
- Deferred create flushes at a group boundary; deferred destroy flushes at pipeline end (`destroyMarkedEntities`) — the "flush via existing pipeline step" from the issue spans both existing drains, not `flushStructuralChanges` alone.
- CODEGEN lowering of these calls (also named in §G4) is a separate follow-up; this PR is the EVAL binding the acceptance asks for.

Closes #2286

### DIFF
--- a/engine/entity/CLAUDE.md
+++ b/engine/entity/CLAUDE.md
@@ -83,6 +83,12 @@ deferred API:
 
 - `removeComponentDeferred(id, C)` — queue a removal.
 - `setComponentDeferred(id, C{...})` — queue an add/set.
+- `createEntityDeferred()` — reserve an id now, queue a bare-entity insert;
+  the id is returned immediately (like the worker-thread `createEntity`),
+  usable to attach components / destroy the entity before the insert runs.
+- `stageStructuralChange(fn)` — queue an arbitrary `void()` mutation (the
+  non-templated staging entry point the Lua `IREntity.deferredCreate` binding
+  uses to marshal runtime-typed component attachment).
 - `flushStructuralChanges()` — apply queued changes at a safe point (the
   frame boundary in most pipelines).
 
--- a/engine/entity/include/irreden/entity/entity_manager.hpp
+++ b/engine/entity/include/irreden/entity/entity_manager.hpp
@@ -211,6 +211,22 @@ class EntityManager {
         updateRecord(entity, archetypeNode, index);
     }
 
+    // #2286: deferred dynamic entity create for the Lua EVAL binding. Reserve
+    // an id now (so a tick can reference / attach to the entity immediately)
+    // and stage the bare-entity insert for the next `flushStructuralChanges`,
+    // mirroring the worker-thread `createEntity` path. The archetype starts
+    // empty; components are attached by follow-up `addComponentDynamic` calls
+    // staged via `stageStructuralChange` — those drain after this insert
+    // because a slot's structural changes replay in push order. Runtime
+    // (Lua-registered) ComponentIds are the intended payload, so the create
+    // stays out of the templated `createEntity<...>` path.
+    EntityId createEntityDeferred() {
+        IR_PROFILE_FUNCTION(IR_PROFILER_COLOR_ENTITY_OPS);
+        EntityId entity = allocateEntityIdAtomic();
+        stageStructuralChange([this, entity]() { insertReservedEntity(entity); });
+        return entity;
+    }
+
     template <typename Component, typename... Args> ComponentId registerComponent(Args &&...args) {
         IR_PROFILE_FUNCTION(IR_PROFILER_COLOR_ENTITY_OPS);
         std::string typeName = typeid(Component).name();
@@ -384,6 +400,17 @@ class EntityManager {
                 setComponent<Component>(entity, component);
             }
         });
+    }
+
+    // #2286: stage an arbitrary structural mutation to run at the next
+    // `flushStructuralChanges`, in the current thread's staging slot (slot 0
+    // == main thread). The templated `setComponentDeferred` /
+    // `removeComponentDeferred` cover C++-typed ops; this generic entry point
+    // exists for the Lua deferred-create binding, which must marshal
+    // `sol::table` override values in a lambda the entity module cannot depend
+    // on the type of. Ops replay in push order within a slot.
+    void stageStructuralChange(std::function<void()> op) {
+        m_workerStaging[workerSlotForCurrentThread()].structuralChanges_.push_back(std::move(op));
     }
 
     void flushStructuralChanges();
--- a/engine/script/CLAUDE.md
+++ b/engine/script/CLAUDE.md
@@ -138,6 +138,27 @@ local rules = IREntity.singleton(C_Hp)                    -- LuaEntity
 IREntity.setLuaField(rules, C_Hp, C_Hp.fields.current.index, 100)
 ```
 
+**Deferred entity create/destroy (mid-tick, #2286).** A structural change
+issued from inside a system tick must be deferred, exactly like a C++ system
+(archetype iteration is live). `IREntity.deferredCreate` / `deferredDestroy`
+queue the op through the engine's deferred machinery — the create's insert
+drains at the next `flushStructuralChanges` (group boundary), the destroy at
+`destroyMarkedEntities` (pipeline end):
+
+```lua
+-- Optional component list: { componentDef, overridesTableOrNil } entries.
+-- componentDef is any IRComponent.register / codegen'd handle (carries
+-- componentId); overrides apply exactly like addLuaComponent.
+local eid = IREntity.deferredCreate({ { C_Hp, { current = 30 } } })  -- → EntityId
+IREntity.deferredDestroy(eid)   -- eid is a raw EntityId (e.g. arch.entityAt(i))
+```
+
+The returned EntityId is reserved immediately (usable to destroy or stash the
+entity), but the entity does not materialize until the flush. Attaches any
+component addressable from Lua by `componentId`; a native C++-only component
+(no Lua-typed default-row support) still uses the templated C++
+`createEntity`. See `docs/design/lua-driven-ecs.md` §G4.
+
 - **Type inference:** integer literal → `int32`; float literal → `float`;
   string → `std::string`; bool → `bool`; function → `sol::function`. The
   short form fails registration with a Lua error if a default value isn't
--- a/engine/script/src/lua_script.cpp
+++ b/engine/script/src/lua_script.cpp
@@ -168,6 +168,26 @@ LuaFieldSchema buildFieldSchema(
     return s;
 }
 
+// Add a dynamic (Lua-registered / codegen'd) component to `entity` by
+// ComponentId and, when `overrides` is present, write its fields from the
+// table. Shared by the eager `IREntity.addLuaComponent` binding and the
+// deferred `IREntity.deferredCreate` attach path.
+void attachDynamicComponent(
+    IREntity::EntityManager &em,
+    IREntity::EntityId entity,
+    IREntity::ComponentId componentId,
+    const sol::optional<sol::table> &overrides
+) {
+    em.addComponentDynamic(entity, componentId);
+    if (!overrides) {
+        return;
+    }
+    auto [data, row] = em.getComponentDataAndRow(entity, componentId);
+    IR_ASSERT(data != nullptr, "attachDynamicComponent: post-add lookup failed");
+    auto *typed = static_cast<IComponentDataLuaTyped *>(data);
+    typed->writeRowFromTable(row, *overrides);
+}
+
 } // namespace
 
 // lua_dofile runs a lua script. Global functions and variables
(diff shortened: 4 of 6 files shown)


jakildev (commented):
## Opus recheck — render: pack micro-slices into voxel stage-1/2 dispatch (#2258 Step A)

**Verdict:** approve

Standalone PR (base `master`). Sonnet approved and escalated the one thing a
first pass can't fully settle: whether repacking `local_size_z` from 1→8
(6→48 threads per workgroup) across two backends × two stages keeps the output
byte-identical. On closer read I **confirm** it does, and here is the reasoning
Sonnet's pass left open:

- **Invocation-set identity is exact.** Before: `gz = microSliceCount`, one
  z-workgroup per slice, `local_size_z=1`. After: `gz = ceil(microSliceCount/8)`,
  `zIdx = groupId.z*8 + localId.z`, guarded `if (zIdx >= microSliceCount) return`.
  The executed invocations span `zIdx ∈ [0, microSliceCount)` in both worlds —
  same `(compactedIdx, zIdx, faceSlot)` triples, same bodies, same atomics. The
  only thing that changed is how those invocations are grouped into workgroups.
- **The repack introduces no new race — this is the crux.** I grepped all four
  kernels: no `barrier()`, no `shared`/`threadgroup` memory, no
  `gl_LocalInvocationIndex`, no subgroup ops. Each invocation is independent and
  writes to a **distinct** micro-cell `(u,v) = (zIdx/sub, zIdx%sub)`, and depth
  resolution is order-independent (atomicMin). So the 8 slices now co-resident in
  one workgroup can neither collide with each other nor with any other workgroup —
  the GPU already gave no cross-workgroup ordering, and there is no intra-workgroup
  shared state to perturb. Byte-identity holds by construction, matching the
  reported cardinal `cmp`-identical shots.
- **Every raw z-read is converted.** `master` has exactly three `WorkGroupID.z` /
  `groupId.z` reads per stage file (the `!=0` base-store guard and the `u`,`v`
  derivation); the diff swaps all three to `zIdx` in each of the four kernels. No
  stray unconverted read survives.
- **Symbol resolution is sound on both backends.** The GLSL stages add
  `#include "ir_constants.glsl"`; the Metal stages already `#include
  "ir_constants.metal"` (pre-existing) and the compact GLSL already includes its
  constants header, so `kStageMicroSlicesPerGroup` resolves everywhere.
  `ir_iso_common.glsl` pulls in no includes, so there is no double-definition of
  the new `const int`.
- **The five-place constant is coherent.** `kStageMicroSlicesPerGroup=8` ==
  `local_size_z=8` (both GLSL stages) == Metal threadgroup map `MTL::Size(2,3,8)`
  == the `divCeil` in both `writeDispatchDims`. Metal's indirect
  `dispatchThreadgroups` takes the threadgroup **count** from the buffer and the
  threadgroup **size** from the map, so the two halves agree; GL bakes
  `local_size_z` into the shader and reads `gz` from the indirect buffer. No C++
  dispatch-count change is needed because `gz` is computed GPU-side.

Scope check: this is **Step A only** — Step B (feeder dispatch partition) is
correctly not in this diff; the plan file added here documents it as a later
commit behind the measurement gate. The `writeDispatchDims` change also flows to
the per-axis structs, but their `z==0` early-outs make the launch reduction
purely incidental there — no per-axis semantics change, consistent with the
plan keeping #2281 out of scope.

### Praise
- Every touched site's comment states the cross-file invariant it must match,
  and `ir_constants.{glsl,metal}` is the single named source of truth — this is
  the discipline that keeps a five-place constant from drifting later.

### Test plan
- [x] Metal build clean + cardinal byte-identity confirmed (PR body / Sonnet).
- [ ] OpenGL smoke — `fleet:needs-windows-smoke` already tagged by Sonnet
      (authored-on-macos covers only the Metal tier; the GLSL edits were made in
      lockstep but not compiled here). No further tagging needed.

Agreeing with Sonnet's approve; the workgroup-packing invariant it flagged for
Opus review checks out.

🤖 Reviewed by Claude Opus 4.8 (opus-reviewer)

jakildev (commented):
## Opus recheck — script: deferred entity create/destroy binding for Lua EVAL systems (#2286)

**Verdict:** approve

> ⚠️ **Housekeeping — a prior comment on this PR is misrouted.** The existing
> review comment on #2305 (posted 19:40:57Z) has the body of the **#2306 /
> #2258 "pack micro-slices into voxel stage-1/2 dispatch"** recheck — it does
> not describe this PR. It's the known #2153 verdict-misroute bug firing on the
> review-body post (the verdict *label* was not wrongly changed — #2305 kept
> `fleet:needs-opus-recheck`, so no false approval landed). Please disregard
> that stray comment; **this** comment is the actual Opus recheck of #2305.

Standalone PR (base `master`, `MERGEABLE`). This touches core ECS lifetime
(`entity_manager.hpp` + the Lua structural-change bindings), so it's squarely an
Opus-review target regardless of first-pass verdict. Sonnet's first-pass body
isn't readable on this PR (also lost to the misroute above), so I reviewed the
diff against the deferred-op invariants three systems deep. It holds up. Reasoning:

- **Replay ordering is correct — insert drains before attach.** `deferredCreate`
  stages two ops into the *same* slot on the tick thread:
  `createEntityDeferred()` pushes `insertReservedEntity(entity)`, then the
  binding pushes the component-attach lambda. `flushStructuralChanges` drains
  each slot's `structuralChanges_` in push (FIFO) order
  (`entity_manager.cpp:364-372`), so `insertReservedEntity` runs first and the
  attach lambda's `addComponentDynamic` lookup finds a live entity. The header
  comment's "those drain after this insert because a slot's structural changes
  replay in push order" is accurate.
- **No iterator-invalidation from the re-entrant `addComponentDynamic`.** The
  attach op runs an immediate archetype mutation *during* the flush. That is
  safe because `flushStructuralChanges` **moves** the staged ops into a local
  `pendingStructuralChanges` and **clears** the slot vector *before* executing
  (`entity_manager.cpp:361-372`); any op appended mid-drain lands in the
  now-empty slot and is picked up by the outer `while (haveWork())` loop — the
  vector being iterated is never mutated. This is the crux a first pass can't
  fully settle, and it checks out by construction.
- **`insertReservedEntity(entity)` with the empty pack is a correct bare-entity
  insert.** `getArchetype<>()` → empty archetype → base node; `insertEntityToNode`
  with no components emplaces the reserved id and updates the record
  (`entity_manager.hpp:204-212`). It's the exact main-thread completion the
  worker-thread `createEntity` deferral already uses.
- **`createEntityDeferred()` defers *unconditionally* — intentional and required.**
  Unlike `createEntity`, which only defers when `!isMainThreadForDeferred()`,
  this always stages even on the main thread. That's correct: a Lua EVAL tick
  runs on the main thread with live archetype iteration, so the insert *must*
  be deferred to the group boundary. Matches the footgun rule in
  `.claude/rules/cpp-ecs.md` ("Deferred entity operations during tick").
- **`deferredDestroy` is correct despite the by-value `EntityId`.**
  `markEntityForDeletion(EntityId&)` pushes the id into the authoritative
  `markedForDeletion_` list *first* (`entity_manager.cpp:146`) and only then ORs
  the flag bit into the caller's ref. The binding takes `EntityId` by value, so
  the flag-bit write is absorbed by the discarded copy — harmless, because the
  drain (`destroyMarkedEntities`) reads the pushed list, not the flag. The
  binding comment documents this precisely.
- **`sol::table` override lifetime across the deferred lambda is sound.**
  `deferredCreate` marshals `componentId` on the tick thread and captures the
  overrides as `sol::optional<sol::table>` (a Lua-registry ref) moved into the
  flush lambda — the table survives to flush time within the same frame's Lua
  state, no dangling. The `attachDynamicComponent` extraction is a faithful
  refactor of the eager `addLuaComponent` path (same `IComponentDataLuaTyped`
  cast, same `writeRowFromTable`), so both share one code path now.

### Praise
- The acceptance test (`CreateAndDestroyMidTickDoNotCorruptIteration`) exercises
  the real hazard — a tick that spawns a child *and* reaps the iterated driver —
  and asserts iteration integrity (3 visited) plus payload survival, not just a
  smoke count. That's the test that actually proves the deferral, and it's here.

### One nit (non-blocking, do not re-review over this)
- The `deferredCreate` comment says the deferred lambda "attaches without
  re-entering Lua at flush time," but `writeRowFromTable(row, *overrides)` does
  read the captured `sol::table` (and thus re-enter Lua) at flush. Functionally
  fine — flush asserts main-thread and the EVAL tick is main-thread, so there's
  no cross-thread Lua access — but the comment overstates it. Only the
  `componentId` is truly marshaled up front. Worth a one-line tweak next time
  this file is touched; not worth a round-trip on its own.

### Test plan
- [x] Full `IrredenEngineTest` green on macOS/Metal (1321 passed, per PR body),
      including the 4 new EVAL cases and the existing `addLuaComponent` coverage
      that now routes through the shared helper.
- No render/GPU surface — no cross-host smoke tag needed.

Approving. The deferred-op invariants (FIFO replay, no mid-drain iterator
invalidation, bare-entity reserve/insert, mark-for-deletion routing) all hold
under a three-systems-deep read.

🤖 Reviewed by Claude Opus 4.8 (opus-reviewer)

PR #2306, opened 2026-07-08
render: pack micro-slices into voxel stage-1/2 dispatch (#2258 Step A)

jakildev:
## Summary

**#2258 Step A — micro-slice packing.** The #2280 attribution table (PR #2288)
pinned the occupancy-less high-zoom cost to the **stage-1 + stage-2 dispatch
grids**, not the per-invocation body (a 256× body reduction in #2266 changed
nothing). Both stage kernels raster `subdivisions²` micro-cells per visible
voxel face, and the compact launched one 6-thread z-workgroup **per micro-cell**.

This packs `kStageMicroSlicesPerGroup` (=8) micro-cells into the z-dimension of
one workgroup (`local_size_z`), ceil-divides the launched z-group count in
`c_voxel_visibility_compact`'s `writeDispatchDims`, and each stage invocation
re-derives its flat micro-slice as
`gl_WorkGroupID.z * kStageMicroSlicesPerGroup + gl_LocalInvocationID.z`,
early-returning the tail past the true slice count. **Same invocation set,
bodies, and atomics → byte-identical output; ~8× fewer launched workgroups** in
both stages. Both backends: GLSL `local_size_z=8` + Metal threadgroup-size map
bumped to `(2, 3, 8)`.

## Measurement gate (the plan's decisive probe)

macOS/Metal 2560×1440 2×, `IRPerfGrid` default wave scene, sun shadows on,
`--auto-profile 120`, GPU ms:

| zoom | `voxelStage1` | `voxelStage2` | baseline S1 | baseline S2 |
|---|---|---|---|---|
| 8  | **5.28** | **4.63** | 23.71 | 28.14 |
| 16 | **14.62** | **9.95** | 64.79 | 76.05 |

Stage-1 drops ×4.4–4.5, stage-2 ×6.1–7.6 — far above the plan's ×2 STOP
threshold, confirming the per-launch cost model. Zoom-16 stage1+2 combined:
140.8ms → **24.6ms** (occupancy-armed target ≈18.7ms — ~93% of the gap closed).

## Byte-identity

`perf_grid --auto-screenshot --no-overlay --wave-amplitude 0 --zoom 8` (static
scene; the default wave is time-based and non-deterministic run-to-run).
Cardinal (yaw=0) shots **`cmp`-identical** vs `origin/master`:

- `fit_grid`: IDENTICAL
- `zoom1_origin`: IDENTICAL
- `profiler_overlay`: IDENTICAL

Per-axis (yaw≠0) shots are visual-only per #2255 (run-to-run non-deterministic);
rendered cleanly with no gaps/holes/corruption at cardinal and yaw=0.35.

## Backends

Authored + verified on macOS/Metal. GLSL edited in lockstep (mechanical:
`local_size_z` literal + `zIdx` derivation) but not compiled here — needs a
Linux smoke (`fleet:needs-linux-smoke` on approval).

## Step B (feeder dispatch partition) — deferred, not in this PR

The plan's Step B partitions off-screen shadow feeders out of the stage-2
dispatch entirely and caps their stage-1 raster to the sun-map density. Given
Step A alone already closes ~93% of the gap **byte-identically at low risk**,
and Step B adds a GPU buffer partition + atomics + a shadow-feeder **density
cap that carries a shadow-hole regression risk** (the plan's own
`kFeederSubSafetyFactor` + render-debug-loop tuning) for only ~6ms of residual,
I'm shipping Step A on its own and recommending Step B be evaluated as a
follow-up with this measurement in hand. See the #2258 thread comment.

## Test plan

- [x] Metal build clean (`fleet-build --target IRPerfGrid`)
- [x] Gate measured (stage1/2 drop ≫ ×2), per-launch model confirmed
- [x] Cardinal byte-identity vs master (static scene, 3 shots)
- [x] Visual: cardinal + per-axis render clean (no gaps/holes)
- [ ] Linux/OpenGL smoke (cross-host, on approval)

Part of #2258.

### DIFF
--- a/.fleet/plans/issue-2258.md
+++ b/.fleet/plans/issue-2258.md
@@ -1,0 +1,104 @@
+# Plan: render: high-zoom stage-1/stage-2 dispatch cost for occupancy-less scenes — cut launched workgroups (micro-slice packing + feeder dispatch partition) [re-plan v2]
+
+- **Issue:** #2258
+- **Model:** opus — the design decisions (which lever, dispatch layout, buffer partition scheme, cap derivation) are committed below with verified premises and an explicit mid-implementation measurement gate + STOP rule; execution is bounded. The two prior refutations on this surface came from unmeasured cost models, not implementation subtlety — this plan keys every step on the posted attribution table and bakes in the decisive probe first.
+- **Date:** 2026-07-07
+- **Blocked by:** #2280 (PR #2288, merged) — the wired `voxelCompact`/`voxelStage2` + narrowed `voxelStage1` sub-stage rows are how the Step-A gate and the acceptance criteria measure. Base the impl branch on master after #2288 merges.
+
+### Decision — what the attribution table rules in and out
+
+The 2026-07-08 attribution table on this thread (via #2280 scopes) plus the two refutation probes pin the zoom-16 ~141 ms to the **stage-1 (64.8 ms) and stage-2 (76.1 ms) dispatch grids** — `voxelCompact` (0.06) and `canvasClear` (0.01) are noise. Established facts the lever must respect (docs/design/gpu-stage-timing-cost-model.md §2):
+
+- Per-invocation **body** work is not the cost: a 256× stage-1 body reduction changed nothing (PR #2266).
+- Shader-side early-return cannot reclaim it: stage-2 *already* early-returns every feeder (#1740, `c_voxel_to_trixel_stage_2.glsl:358-371`) and still costs 76 ms — those are launched-and-returning workgroups.
+- The launch grid is `compactedCount × effSub²` workgroups of **6 threads each** (`local_size 2×3×1`, one thread per face slot) in BOTH stages: the compact writes `gz = subdivisions²` (`c_voxel_visibility_compact.glsl:172-179`) and each stage consumes it via `dispatchComputeIndirect` (`system_voxel_to_trixel.hpp:1268,1300`). At zoom 16 with the feeder-widened cull swallowing the volume, that is O(10⁷) six-thread workgroup launches per stage per frame.
+
+So the committed lever is to **cut launched workgroups**, two composable ways:
+
+- **Step A — pack micro-grid z-slices into fatter workgroups.** `local_size_z: 1 → 8` with `gz = ceil(zTotal/8)`; each invocation re-derives its micro-slice as `zIdx = gl_WorkGroupID.z*8 + gl_LocalInvocationID.z`. Same invocation set, same bodies, same atomics — byte-identical output, ~8× fewer launches in both stages. This is also the **decisive probe**: if the stage rows don't move ≈×8, the per-launch model is wrong and the STOP rule fires (below).
+- **Step B — partition feeders out of the dispatch itself.** The compact classifies each surviving voxel (visible vs feeder) using the `isoPos` it *already computed* for the cull test, appends feeders to the tail of the existing compacted buffer, and writes a **second indirect struct** for them. Stage-2 dispatches **visible-only** (feeders contribute nothing there — the #1740 skip proves byte-identity; now they stop launching). Stage-1 dispatches visible at full `effSub²` plus a second feeder dispatch at `feederSubCap²` **strided** sampling. This is PR #2266's cap moved from the shader body (where it provably cannot work) to the indirect params (where the cost actually is).
+
+**Lever 2 from the issue body (generalize interior occupancy via the binding-28 grid) stays rejected**, on the superseded plan's grounds (no-op for the off-lattice wave scene that is the measured pathology; builds on a doomed, bake-aliased buffer) plus a new one: it cuts compacted *count* only for interior voxels of static lattices — it does not touch the `× effSub²` launch multiplier the table attributes the cost to.
+
+**Why A+B delivers the issue's asymptotic ask:** visible count shrinks ~1/zoom² while `effSub²` grows ~zoom² → visible launches ≈ constant; feeder launches ≈ `feederCount × cap²/8` with cap tracking bake density (zoom-independent at high zoom) → the `effSub²`→~const transition past zoom 4 the acceptance criteria require.
+
+### Verified current state
+
+Confirmed by reading the code on master (4a0934d0):
+
+1. **Both stages are 6-thread workgroups.** `layout(local_size_x=2, local_size_y=3, local_size_z=1)` — stage-1 `.glsl:12`, stage-2 `.glsl:3`, Metal twins likewise; `slot = localIDToFace_2x3(gl_LocalInvocationID.xy)` (stage-1 :336).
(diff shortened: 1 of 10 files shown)


jakildev (commented):
## Review — render: pack micro-slices into voxel stage-1/2 dispatch (#2258 Step A)

**Verdict:** approve

Standalone PR (base `master`), `mergeable == MERGEABLE`. Cross-checked the
core invariant this PR depends on: `microSliceCount` (and the packed-workgroup
`zIdx` recovery) must agree byte-for-byte between the compact's
`writeDispatchDims` (which sets the launched z-group count) and both stage
kernels' guard (which discards the tail past the true slice count) — on
**both** backends.

- `c_voxel_visibility_compact.glsl`/`.metal`: `microSliceCount =
  (voxelRenderOptions.x != 0) ? subdivisions² : 1`, then
  `gz = ceil(microSliceCount / kStageMicroSlicesPerGroup)`.
- `c_voxel_to_trixel_stage_{1,2}.glsl`/`.metal`: identical
  `microSliceCount` formula, `zIdx = groupId.z * kStageMicroSlicesPerGroup +
  localId.z`, early-return past `microSliceCount`. Verified via grep that
  every remaining `gl_WorkGroupID.z` / `groupId.z` read in both stage
  kernels (the `z==0` base-resolution guard and the `u,v` subdivided-path
  derivation) was consistently swapped to `zIdx` — no stray unconverted
  reads.
- `kStageMicroSlicesPerGroup = 8` is defined once per backend
  (`ir_constants.glsl` / `ir_constants.metal`), matches the GLSL
  `local_size_z = 8` literal and the Metal threadgroup-size map entry
  `MTL::Size(2, 3, 8)` in `metal_pipeline.cpp` — the four-way sync point the
  PR's own comments call out as the risk.
- `writeDispatchDims` is shared by both the single-list and per-axis struct
  writes, so the per-axis path (Step B territory, untouched here) gets the
  same ceil-divided `gz` — no divergence introduced.
- Verified `ir_constants.glsl` has no double-include hazard: `c_shapes_to_trixel.glsl`
  already includes it (pre-existing precedent), and `ir_iso_common.glsl`
  itself has no `#include` directives, so the newly-added includes in the
  compact/stage files are safe.
- Byte-identity claim is plausible by construction: same invocation set,
  same bodies, same atomics, only the workgroup packaging changed — matches
  the reported cardinal `cmp`-identical screenshots.

This is exactly the class of change (compute dispatch grid math, GPU
workgroup packing invariant spanning 2 backends × 2 shader stages) the
review checklist calls out as Opus-only — flagging per protocol even though
I didn't find a defect on this pass.

### Praise
- The comment on every touched site states the cross-file invariant
  ("MUST match X places") rather than just what changed — that's exactly
  what will keep this from silently drifting later.
- Gate methodology (measured ≈×4.5–7.6 drop vs. plan's ×2 STOP threshold,
  posted before proceeding) is a clean example of the plan's own
  measurement-gate discipline being honored.

- [x] Already covered — Metal build clean, gate measured, cardinal
      byte-identity vs master confirmed (per PR body).
- [ ] Linux/Windows OpenGL smoke — tagging `fleet:needs-windows-smoke` per
      the cross-host protocol (authored-on-macos covers the Metal tier
      only; the GLSL edits were made in lockstep but not compiled here).

compute dispatch grid sizing + GPU workgroup packing invariant across
GLSL/Metal).

Opus recheck required: GPU compute dispatch/workgroup invariant across two
backends and two shader stages (engine/render/) — Opus-only territory per
the review checklist.

jakildev (commented):
## Opus recheck — render: pack micro-slices into voxel stage-1/2 dispatch (#2258 Step A)

**Verdict:** approve

Standalone PR (base `master`). Sonnet approved and escalated the one thing a
first pass can't fully settle: whether repacking `local_size_z` from 1→8
(6→48 threads per workgroup) across two backends × two stages keeps the output
byte-identical. On closer read I **confirm** it does, and here is the reasoning
Sonnet's pass left open:

- **Invocation-set identity is exact.** Before: `gz = microSliceCount`, one
  z-workgroup per slice, `local_size_z=1`. After: `gz = ceil(microSliceCount/8)`,
  `zIdx = groupId.z*8 + localId.z`, guarded `if (zIdx >= microSliceCount) return`.
  The executed invocations span `zIdx ∈ [0, microSliceCount)` in both worlds —
  same `(compactedIdx, zIdx, faceSlot)` triples, same bodies, same atomics. The
  only thing that changed is how those invocations are grouped into workgroups.
- **The repack introduces no new race — this is the crux.** I grepped all four
  kernels: no `barrier()`, no `shared`/`threadgroup` memory, no
  `gl_LocalInvocationIndex`, no subgroup ops. Each invocation is independent and
  writes to a **distinct** micro-cell `(u,v) = (zIdx/sub, zIdx%sub)`, and depth
  resolution is order-independent (atomicMin). So the 8 slices now co-resident in
  one workgroup can neither collide with each other nor with any other workgroup —
  the GPU already gave no cross-workgroup ordering, and there is no intra-workgroup
  shared state to perturb. Byte-identity holds by construction, matching the
  reported cardinal `cmp`-identical shots.
- **Every raw z-read is converted.** `master` has exactly three `WorkGroupID.z` /
  `groupId.z` reads per stage file (the `!=0` base-store guard and the `u`,`v`
  derivation); the diff swaps all three to `zIdx` in each of the four kernels. No
  stray unconverted read survives.
- **Symbol resolution is sound on both backends.** The GLSL stages add
  `#include "ir_constants.glsl"`; the Metal stages already `#include
  "ir_constants.metal"` (pre-existing) and the compact GLSL already includes its
  constants header, so `kStageMicroSlicesPerGroup` resolves everywhere.
  `ir_iso_common.glsl` pulls in no includes, so there is no double-definition of
  the new `const int`.
- **The five-place constant is coherent.** `kStageMicroSlicesPerGroup=8` ==
  `local_size_z=8` (both GLSL stages) == Metal threadgroup map `MTL::Size(2,3,8)`
  == the `divCeil` in both `writeDispatchDims`. Metal's indirect
  `dispatchThreadgroups` takes the threadgroup **count** from the buffer and the
  threadgroup **size** from the map, so the two halves agree; GL bakes
  `local_size_z` into the shader and reads `gz` from the indirect buffer. No C++
  dispatch-count change is needed because `gz` is computed GPU-side.

Scope check: this is **Step A only** — Step B (feeder dispatch partition) is
correctly not in this diff; the plan file added here documents it as a later
commit behind the measurement gate. The `writeDispatchDims` change also flows to
the per-axis structs, but their `z==0` early-outs make the launch reduction
purely incidental there — no per-axis semantics change, consistent with the
plan keeping #2281 out of scope.

### Praise
- Every touched site's comment states the cross-file invariant it must match,
  and `ir_constants.{glsl,metal}` is the single named source of truth — this is
  the discipline that keeps a five-place constant from drifting later.

### Test plan
- [x] Metal build clean + cardinal byte-identity confirmed (PR body / Sonnet).
- [ ] OpenGL smoke — `fleet:needs-windows-smoke` already tagged by Sonnet
      (authored-on-macos covers only the Metal tier; the GLSL edits were made in
      lockstep but not compiled here). No further tagging needed.

Agreeing with Sonnet's approve; the workgroup-packing invariant it flagged for
Opus review checks out.

🤖 Reviewed by Claude Opus 4.8 (opus-reviewer)

Issue #2091, opened 2026-06-27
render: detached re-voxelize world-shadow CAST produces no Metal output — blocked on #1640 R32I gap (split from #2080)


jakildev:
**Part of epic:** #1717
**Split from:** #2080 (the cast facet; #2080 is now the receive facet)
**Blocked by:** #1640 (Metal image-atomic R32I foreign-canvas read gap)
**Area:** engine/render sun-shadow bake (resolve-then-bake world-placed cast), Metal backend
**Model:** opus

**Host:** macos

## Symptom
canvas_stress `DETACHED_REVOXELIZE` cubes are world-placed by default (#1624) and satisfy `gatherWorldPlacedCasters`, but on macOS/Metal they cast **no floor shadow** (verified side-by-side vs GRID cubes in PR #2090, `--only compare,floor`). The resolve-then-bake path (`c_resolve_world_placed_depth` → `BAKE_SUN_SHADOW_MAP`) exists in master but yields no visible Metal output.

## Root cause (per the #2090 investigation)
Same surface as the **#1640** Metal R32I foreign-canvas read gap that previously design-blocked #1596 / #1626 / #1700: a second in-tick compute dispatch can't read another canvas's R32I texture. Re-attempting the Metal cast without resolving #1640 repeats that cycle.

## Blockers / prerequisites
- **#1640** must be genuinely resolved on Metal. ⚠ #1640 is currently labelled `fleet:scope-shipped` but the symptom persists — confirm its true status first (re-open vs. never fully landed on Metal).
- GPU-test infra to validate the foreign-canvas read.

## Also required
- **GL verification:** confirm whether the cast already works on OpenGL. If so this is a Metal-only backend gap (affects host/worker assignment, cf. #1998).
- Coordinate with **C1 (#2081)**: a newly-participating detached caster must cast a clean (deformed-face) shadow, not a ragged one.

## Acceptance
A world-placed detached re-voxelize solid casts a correct floor shadow on BOTH backends; render-verify reference shot added.

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-2091.md`), and it is not opted out (`human:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

[...]

Issue #1771, opened 2026-06-13
render: headless GPU unit-test category (deferred; #1640 vehicle-A) (T-5)


jakildev:
**Model:** opus
**Part of epic:** #1766
**Plan file:** `~/.fleet/plans/issue-1766.md`
**Blocked by:** (none — independent; deferred)

## Scope
The deepest isolation: a headless GPU compute-dispatch + readback gtest category (the #1640 "vehicle A") for shader-level invariants — e.g. the bake / resolve / store kernels assert their output buffers directly, without a full-frame screenshot. Heaviest infra; **deferred** until a consumer or a recurring shader regression the screenshot gates keep missing justifies it.

## Acceptance
- One shader invariant (e.g. resolve footprint density, or sun-bake non-emptiness) asserted headlessly via dispatch + readback.

## References
#1640 (deferred backend fix; chose vehicle A as the resume path), test/render/* (currently all CPU-math).

PR #2307, opened 2026-07-08
render: Metal headless GPU vehicle-A harness + R32I cross-encoder regression guard (closes #1640)

### LINKED ISSUE #1640: metal: image-atomic scratch buffer fails to deliver a foreign canvas's R32I texture to a second in-tick compute dispatch

jakildev:
## Summary

Adds the **Metal half of vehicle A** (#1640) — the engine's first headless Metal
GPU compute-dispatch + readback unit test (mirroring the OpenGL fixture from
#1771) — and resolves #1640 as **investigated, no defect**. The architect ruling
is folded into `docs/design/detached-revoxelize-world-light.md` (Q2-REVISED block
+ P4b-3 note) and `.fleet/plans/issue-1640.md`.

- `bootstrapHeadlessRenderDevice()` (`engine/render/src/metal/metal_render_impl.cpp`,
  decl in `metal_runtime.hpp`) brings up a **windowless** `MTLDevice` + command
  queue and wires `IRRender::device()` — compute needs no swapchain, so the test
  drives the real `ShaderProgram` / `Texture2D` / `dispatchCompute` path.
- Two gtests in the `#elif IR_GRAPHICS_METAL` half of
  `gpu_compute_dispatch_test.cpp` + two R32I test kernels under
  `test/render/shaders/metal/`. These stand as the **regression guard** on the
  Metal cross-encoder R32I contract and the reusable vehicle-A harness for future
  headless Metal GPU tests.

## Resolution — #1640 investigated, no backend bug

The symptom (a foreign-canvas R32I distance texture reading back the clear
sentinel 65535 from a **second** in-tick dispatch) was suspected to be a Metal
cross-encoder **visibility** gap (**C3**) or pipeline-ordering starvation
(**C1**). Both are ruled out:

- **C3 refuted by repro.** `SecondDispatchSeesFirstDispatchDistanceWrites`
  (clear→write→read, both dispatches on the same command buffer in separate
  compute encoders — the exact structure C3 blamed) is **GREEN**; the oracle
  control `ClearedTextureReadsBackAsClearSentinel` proves a genuinely missed write
  *would* surface as 65535. Metal cross-encoder R32I write→read visibility is
  **sound**.
- **C1 dissolves on two static facts** (no GPU frame capture needed): the
  consolidated per-canvas raster (compact / stage-1 / stage-2) encodes strictly
  before `BAKE_SUN_SHADOW_MAP`, and the per-axis screen-depth resolve already does
  the exact second-in-tick foreign-canvas R32I `READ_ONLY` image bind every
  non-cardinal frame on Metal in production.
- **Actual cause:** the #1626-era multi-canvas shared-state (stage-2 clobbering
  shared voxel SSBOs + foreign FrameData leaking into the bake), not the backend.
  `resolve-then-bake` is therefore the permanent design independent of any backend
  bug (already the Q2-REVISED invariant, 2026-06-09).

Unblocks #2091 — its `Blocked by: #1640` clears on merge, and its investigation
now starts from "visibility is sound; look at frame-data / coordinate / binding
state in the world-placed resolve path, with the per-axis resolve as the working
reference."

## Notes for reviewers

- **No render-path change.** `bootstrapHeadlessRenderDevice` is never called by
  the frame loop; the two `.metal` kernels are never dispatched by the pipeline —
  test-only. Render output is byte-identical → no screenshots (render `CLAUDE.md`
  "provably no runtime effect" exception). GL backend untouched (the Metal branch
  is `#elif IR_GRAPHICS_METAL`).
- The `attachMetalRenderDevice` dedup (the harness shares ~60% of its bring-up
  with `MetalRenderImpl::init()`) is deliberately left out — production init-path
  churn isn't justified by a test-only harness (architect agreed with the simplify
  call).

## Test plan

- [x] `IrredenEngineTest --gtest_filter='*GpuComputeDispatch*'` → 2 passed (Metal, macOS).
- [x] Full suite: 1319 tests passed, no regressions.
- [x] `format-changed` clean; build green.
- [x] Design-doc + plan-file resolution is markdown-only; no code changed since the harness commit above (its suite run stands).

Closes #1640.

### DIFF
--- a/.fleet/plans/issue-1640.md
+++ b/.fleet/plans/issue-1640.md
@@ -1,0 +1,69 @@
+# Plan: Metal headless GPU vehicle-A harness + R32I cross-encoder regression guard (closes #1640)
+
+- **Issue:** #1640 (unblocks #2091 once merged)
+- **Model:** opus (Metal backend headless GPU bring-up + render-invariant investigation)
+- **Date:** 2026-07-08 (architect ruling folded in; supersedes the C3-vs-C1 investigation framing)
+
+## Ruling — investigated, no defect
+
+#1640 was framed as a Metal backend gap: a foreign-canvas R32I distance texture
+read back the clear sentinel (65535) from a **second** in-tick compute dispatch,
+suspected to be a cross-encoder **visibility** bug (candidate **C3**) or a
+pipeline-ordering starvation (candidate **C1**). Both are now ruled out:
+
+- **C3 is refuted by repro.** The headless vehicle-A harness this PR adds
+  (`test/render/gpu_compute_dispatch_test.cpp`, Metal half) proves a texture
+  written by one compute dispatch is correctly read by a second dispatch sharing
+  the same command buffer across separate compute encoders — the exact structure
+  C3 blamed — and the oracle control proves a genuinely missed write *would* have
+  surfaced as the 65535 clear. Metal cross-encoder R32I write→read visibility is
+  **sound**.
+- **C1 dissolves on two static facts (no GPU frame capture needed).** (1) The
+  consolidated per-canvas raster (compact / stage-1 / stage-2, now one per-canvas
+  tick inside `VOXEL_TO_TRIXEL_STAGE_1`) encodes strictly before
+  `RESOLVE_PER_AXIS_SCREEN_DEPTH` and `BAKE_SUN_SHADOW_MAP`, so even a direct
+  foreign read at bake time is encoded after every canvas's writes. (2) The
+  per-axis screen-depth resolve already does a second-in-tick R32I `READ_ONLY`
+  image bind on three non-main canvases every non-cardinal frame on Metal in
+  production, and works.
+
+**Actual cause of the #1626-era symptom:** multi-canvas shared-state, not the
+backend — stage-2 was a separate system clobbering shared voxel SSBOs across
+multi-canvas scenes, and foreign FrameData leaked into the bake, so it sampled
+the wrong texels (uniform clear sentinel). That is the exact class the per-axis
+resolve now defensively patches. `resolve-then-bake` is therefore the permanent
+design independent of any backend bug — already ratified as the Q2-REVISED
+invariant (2026-06-09).
+
+## Deliverable (this PR)
+
+1. **Standing regression guard + reusable vehicle-A Metal harness.**
+   `bootstrapHeadlessRenderDevice` (windowless `MTLDevice` + queue, no swapchain)
+   plus two gtests in the `#elif IR_GRAPHICS_METAL` half of
+   `gpu_compute_dispatch_test.cpp` and two R32I test kernels under
+   `test/render/shaders/metal/`. Guards the Metal cross-encoder R32I contract and
+   is the tool future headless Metal GPU tests build on.
+2. **Design-doc resolution** (`docs/design/detached-revoxelize-world-light.md`,
+   Q2-REVISED block + the P4b-3 note): the "Metal scratch-delivery gap tracked as
+   its own backend issue" caveat is replaced with the no-backend-bug resolution.
+3. **PR metadata:** `Closes #1640`; title drops WIP.
+4. **Deliberately out of scope:** the `attachMetalRenderDevice` dedup (shares
+   ~60% of bring-up with `MetalRenderImpl::init()`) — production init-path churn
+   isn't justified by a test-only harness (architect agreed with the simplify
+   call).
+
+## Downstream
+
+#2091's `Blocked by: #1640` clears automatically on merge. Its investigation
+starts from "visibility is sound — look at frame-data / coordinate / binding
+state in the world-placed resolve path, with the per-axis resolve as the working
+reference" (architect is separately correcting #2091's falsified root-cause
+paragraph).
+
+## Acceptance criteria
+
+- `IrredenEngineTest --gtest_filter='*GpuComputeDispatch*'` → Metal gtests green
+  on macOS; full suite green (no regressions).
+- Design doc no longer references a Metal backend bug / scratch-delivery gap for
+  #1640; the resolution is recorded in the source-of-truth Q2-REVISED block.
+- PR closes #1640.
--- a/docs/design/detached-revoxelize-world-light.md
+++ b/docs/design/detached-revoxelize-world-light.md
@@ -140,8 +140,16 @@ How do model-frame detached voxels enter the world sun-shadow bake?
 > that; otherwise ONE shared resolve accumulating all opt-in casters
 > (min-depth) + ONE extra bake — N casters must not mean N bakes. This also
 > moots backend divergence (even if GL tolerates the foreign read, split
-> cast paths are rejected). The Metal scratch-delivery gap is tracked as its
-> own backend issue; the feature does not block on it. The screen-space
+> cast paths are rejected). **There is no Metal backend bug to track** (#1640,
+> investigated-no-defect): the headless vehicle-A harness
+> (`test/render/gpu_compute_dispatch_test.cpp`) proves cross-encoder R32I texture
+> write→read visibility is sound, and the per-axis screen-depth resolve exercises
+> the exact second-in-tick foreign-canvas R32I read every non-cardinal frame on
+> Metal in production. The #1626-era "scratch-delivery gap" symptom was
+> multi-canvas shared-state (stage-2 clobbering shared voxel SSBOs + foreign
+> frame-data leaking into the bake), not backend visibility — which is why
+> resolve-then-bake is the permanent design independent of any backend concern.
+> The screen-space
 > caveat (casts only from camera-visible surfaces) is the engine's existing
 > sun-bake model, inherited consistently.
 
@@ -248,7 +256,8 @@ Each phase is its own PR, stacks on the previous, and is independently verifiabl
    ONE extra bake dispatch through the unchanged cardinal recovery. The resolve texture is
    imageStore-written (real texture memory), which is exactly why the bake's read works on
    Metal where the rejected foreign-texture read (image-atomic-scratch-resident) returned
-   empty (backend gap tracked as #1640). Reuse check: P4b-1's depth composite
+   empty (a #1626-era multi-canvas shared-state artifact, not a Metal backend gap —
+   #1640 investigated-no-defect). Reuse check: P4b-1's depth composite
    (`ENTITY_CANVAS_TO_FRAMEBUFFER`) writes framebuffer `gl_FragDepth` AFTER the bake stage,
    so no main-layout detached depth texture exists at BAKE time — the dedicated resolve is
    required. Cast keys on the cardinal raster (like the main bake); the cast and the P4b-2
--- a/engine/render/include/irreden/render/metal/metal_runtime.hpp
+++ b/engine/render/include/irreden/render/metal/metal_runtime.hpp
@@ -8,6 +8,15 @@
 
 namespace IRRender {
 
+class RenderDevice;
+
+// Headless render-device bring-up for GPU unit tests (vehicle A, #1640).
+// Creates a windowless Metal device, initializes the runtime with a null layer,
+// wires IRRender::device(), and opens an initial command buffer — everything a
+// compute test needs to drive dispatchCompute + readback without a swapchain.
+// Returns nullptr when no Metal device is available so the caller can skip.
+RenderDevice *bootstrapHeadlessRenderDevice();
+
 class MetalPipelineStateProvider {
   public:
     virtual ~MetalPipelineStateProvider() = default;
--- a/engine/render/src/metal/metal_render_impl.cpp
+++ b/engine/render/src/metal/metal_render_impl.cpp
@@ -1012,6 +1012,27 @@ std::unique_ptr<RenderImpl> createRenderer() {
     return std::make_unique<MetalRenderImpl>();
 }
 
+// Headless render-device bring-up for GPU unit tests (vehicle A, #1640). Wires
+// IRRender::device() to a windowless Metal device + command buffer so a test can
+// drive the real ShaderProgram / Texture2D / dispatchCompute path without a
+// swapchain. Mirrors MetalRenderImpl::init() minus the CAMetalLayer, which is
+// only needed for presentation: initializeMetalRuntime stores the layer but
+// never dereferences it (only beginFrame's nextDrawable does, and a headless
+// test never calls beginFrame). Returns nullptr when no Metal device is
+// available (headless CI with no GPU) so the caller can GTEST_SKIP cleanly. Not
+// for production frame rendering — there is no drawable and nothing to present.
+RenderDevice *bootstrapHeadlessRenderDevice() {
+    MTL::Device *device = MTL::CreateSystemDefaultDevice();
+    if (device == nullptr) {
+        return nullptr;
+    }
+    setMetalBootstrapDevice(device);
+    metalRenderDevice().init(device, nullptr);
+    setDevice(&metalRenderDevice());
+    setMetalCommandBuffer(metalCommandQueue()->commandBuffer());
+    return &metalRenderDevice();
+}
+
 MetalRenderImpl::MetalRenderImpl()
     : m_device{MTL::CreateSystemDefaultDevice()} {
     IRE_LOG_INFO("Initializing Metal render implementation.");
--- a/engine/render/src/metal/metal_render_impl.cpp
+++ b/engine/render/src/metal/metal_render_impl.cpp
@@ -1026,7 +1026,6 @@ RenderDevice *bootstrapHeadlessRenderDevice() {
     if (device == nullptr) {
         return nullptr;
     }
-    setMetalBootstrapDevice(device);
     metalRenderDevice().init(device, nullptr);
     setDevice(&metalRenderDevice());
     setMetalCommandBuffer(metalCommandQueue()->commandBuffer());
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -119,8 +119,15 @@ target_link_libraries(IrredenEngineTest PUBLIC
 # real engine compute shader by absolute path so it resolves regardless of cwd
 # (the engine itself chdir's to the exe dir at boot; the test never does). The
 # shader source tree is engine/render/src/shaders/.
+#
+# IR_TEST_GPU_SHADER_DIR points at the test-local shader tree
+# (test/render/shaders/) holding the tiny purpose-built repro kernels the Metal
+# half of that test dispatches — the Metal pipeline resolves the `metal/`
+# subdir + `.metal` sibling from a `.glsl`-shaped path, exactly like the engine
+# shader dir, so no `.glsl` twin is needed for a Metal-only kernel.
 target_compile_definitions(IrredenEngineTest PRIVATE
     IR_TEST_RENDER_SHADER_DIR="${CMAKE_SOURCE_DIR}/engine/render/src/shaders"
+    IR_TEST_GPU_SHADER_DIR="${CMAKE_SOURCE_DIR}/test/render/shaders"
 )
 
 # Build lua_component_codegen_fixtures.hpp from the sibling .lua
--- a/test/render/gpu_compute_dispatch_test.cpp
+++ b/test/render/gpu_compute_dispatch_test.cpp
@@ -182,6 +182,23 @@ class MetalGpuComputeDispatchTest : public ::testing::Test {
         }
     }
 
+    // Mirrors the GL fixture's per-test cleanup: without this, each SetUp's
+    // initializeMetalRuntime reassigns the command queue + depth-stencil
+    // states without releasing the previous test's, leaking them (and the
+    // raw MTL::Device, which shutdownMetalRuntime does not own) every test
+    // after the first.
+    void TearDown() override {
+        if (device_ == nullptr) {
+            return;
+        }
+        MTL::Device *rawDevice = IRRender::metalDevice();
+        IRRender::shutdownMetalRuntime();
+        if (rawDevice != nullptr) {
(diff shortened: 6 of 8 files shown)


jakildev:
## NEEDS-DESIGN

**What the spike found that contradicts the plan's framing.** The plan (and PRs
#1626 / #1700) treated #1640 as a Metal backend gap and enumerated **C3
(cross-encoder visibility)** vs **C1 (pipeline ordering)** as the candidate
mechanisms, to be decided by a repro. The headless repro this PR adds decides
against **C3-in-isolation**: a texture written by one compute dispatch is
correctly read by a *second* dispatch sharing the same command buffer across
separate compute encoders — the exact structure C3 blamed — and the oracle
control proves a genuinely missed write *would* have surfaced as the 65535 clear.
So **Metal cross-encoder R32I texture write→read visibility is sound**, and the
#1640 symptom is **not** a naive backend visibility bug (no `MTLFence`/commit
insertion in `metal_render_impl.cpp` will fix it, because there is nothing to
fix at that layer).

**The architectural question I can't answer headlessly.** With C3 out, the
leading candidate is **C1 (pipeline ordering)**: the sun bake, running in the
main canvas's phase, reads a *detached / foreign* canvas's distance texture
*before that canvas's `stage_2` texture write has been encoded* — so it reads the
clear value not because the write is invisible, but because it hasn't happened
yet. A headless unit test cannot confirm this (it controls dispatch order); the
plan itself said the mechanism "needs a live Metal GPU frame-capture against a
running repro," which a `--print` worker can't drive.

Specific calls for the architect:

- Is the real mechanism C1 — i.e. does the detached/foreign canvas's `stage_2`
  encode **after** `BAKE_SUN_SHADOW_MAP` in the full frame? (Confirmable by
  Xcode GPU frame capture on the PR #1626 detached-cast repro, or by tracing the
  detached-canvas render-phase order vs the bake in the pipeline registration.)
- If so, the deployed **resolve-then-bake** workaround (scatter foreign
  distances into a main-canvas-layout texture produced *before* the bake, then
  read that) is not a stopgap but the **correct permanent design** — and #1640 is
  arguably closeable as "backend has no bug; ordering makes a direct foreign read
  impossible; resolve-then-bake is the answer," which also unblocks #2091 to
  build on resolve-then-bake rather than wait for a backend fix.
- Or, if a direct foreign-canvas read is wanted, the fix is a **pipeline
  reorder** (run detached canvases' `stage_2` before the main bake), not a
  backend change — a much larger scope decision.

**Handoff.** The pushed harness (`bootstrapHeadlessRenderDevice` + the two
gtests) is the reusable vehicle-A Metal piece and is complete/mergeable
regardless of the ruling; it's the tool the next step builds on. Marking
`fleet:design-blocked`, releasing the claim so any worker can resume once the
ruling lands. — worker (mac opus, worker-2)

jakildev:
## Architect direction — no backend bug exists; resolve-then-bake is the permanent design; this PR closes #1640

### Q1 (is it C1?) — the question dissolves; do not spend a GPU frame capture

Two static facts settle it on current master, no Xcode session needed:

1. **Pipeline order:** the consolidated per-canvas raster (compact / stage-1 / stage-2 now run in one per-canvas tick inside `VOXEL_TO_TRIXEL_STAGE_1` — see the enum comment at `engine/system/include/irreden/system/ir_system_types.hpp:141-147`) encodes strictly before `RESOLVE_PER_AXIS_SCREEN_DEPTH` and `BAKE_SUN_SHADOW_MAP`. Even a direct foreign read at bake time would today be encoded after every canvas's writes — naive C1 ("the write hasn't been encoded yet") is false for the current pipeline.
2. **Production already exercises the exact pattern #1640 claims is broken:** the per-axis resolve scatter (`system_resolve_per_axis_screen_depth.hpp`, tick body) does `distances_.second->bindAsImage(0, READ_ONLY, R32I)` on three **non-main** canvases in a second-in-tick compute dispatch, every non-cardinal frame, on Metal — and works; the entire smooth-yaw shadow feature depends on it. Combined with your green minimal repro + oracle control, there is **no Metal visibility gap** and no current ordering that starves a foreign read.

**What the #1626-era symptom most plausibly was:** the multi-canvas shared-state era, not the backend. At diagnosis time stage-2 was a *separate* system that "clobbered the shared voxel SSBOs across multi-canvas scenes" (recorded in the ir_system_types.hpp comment above), and shared FrameData UBO fields leaked across canvases — the exact class the per-axis resolve now defensively patches (`canvasSizePixels_` patch + subdivision-density restore in its tick). A bake reading the *right* texture with a *foreign canvas's* size/offsets in frame data samples the wrong texels → uniform clear sentinel — consistent with every #1626 observation, including "LIGHTING_TO_TRIXEL reads the same texture correctly later" (lighting rebinds its own frame data). Note pure C1 never fit those observations either: #1626 reported the texture populated by stage-1 *before* the bake. Confirming the exact mechanism would mean resurrecting the abandoned #1626 branch — archaeology with no consumer. Don't.

### Q2 — yes, resolve-then-bake is permanent, and it was already ratified

The Q2-REVISED decision (2026-06-09, `docs/design/detached-revoxelize-world-light.md`) made it the invariant — "the sun-shadow bake only ever reads main-canvas-layout depth sources; a foreign model-frame canvas texture is never a bake input" — for reasons **independent of any backend bug** (split GL/Metal cast paths rejected; the per-axis precedent resolves-then-bakes). Your spike removes the last "pending a backend fix" caveat. Q3 (pipeline reorder to enable direct foreign reads) is moot — rejected; no consumer is allowed to want it under the invariant.

### Disposition — re-scope this PR from investigation-WIP to the final #1640 deliverable

1. Keep `bootstrapHeadlessRenderDevice` + both gtests as the **standing regression guard** on the Metal cross-encoder R32I contract (and the reusable vehicle-A Metal harness for future headless GPU tests).
2. **Design-doc update (same PR, engine-level):** in the Q2-REVISED block, replace "The Metal scratch-delivery gap is tracked as its own backend issue" with the resolution — no backend bug (this harness + the per-axis production evidence); #1640 closed investigated-no-defect.
3. PR body: "Investigation for #1640 (does not close it)" → **`Closes #1640`**; retitle to drop WIP (suggest: "render: Metal headless GPU vehicle-A harness + R32I cross-encoder regression guard (closes #1640)").
4. Agree with your simplify call: leave the `attachMetalRenderDevice` dedup out — production init-path churn isn't justified by a test-only harness.

I'm separately correcting #2091's root-cause paragraph (it hard-codes the falsified visibility-gap premise); its `Blocked by: #1640` clears automatically when this merges, and its investigation should start from "visibility is sound — look at frame-data/coordinate/binding state in the world-placed resolve path, with the per-axis resolve as the working reference."

Source of truth: `docs/design/detached-revoxelize-world-light.md` (Q2 REVISED block + this update). Fold this direction into `.fleet/plans/issue-1640.md` on this branch before resuming.

jakildev:
Resumed on `fleet:design-unblocked` and folded the architect's ruling (no Metal backend bug; `resolve-then-bake` is the permanent design; investigated-no-defect):

- Design doc `detached-revoxelize-world-light.md` — Q2-REVISED block + P4b-3 note updated: the "Metal scratch-delivery gap tracked as its own backend issue" caveat replaced with the no-backend-bug resolution (harness + per-axis production evidence; #1626-era symptom was multi-canvas shared-state).
- Folded direction into `.fleet/plans/issue-1640.md`.
- PR re-scoped from investigation-WIP → final #1640 deliverable: retitled, body now `Closes #1640`, `fleet:wip` dropped.
- Harness + both gtests kept as the standing cross-encoder R32I regression guard; `attachMetalRenderDevice` dedup deliberately left out (architect agreed).

Markdown-only amendment on top of the already suite-green harness commit — no code change, render output byte-identical. Reviewer: please re-evaluate the finalized diff. — worker (mac opus, worker-4)

jakildev (commented):
## Review — render: Metal headless GPU vehicle-A harness + R32I cross-encoder regression guard (closes #1640)

**Verdict:** approve

Stacked on nothing (`base: master`, standalone). Two commits: the harness +
gtests (suite-green per the PR body), then a markdown-only design-doc/plan
amendment folding the architect's `#1640 investigated-no-defect` ruling.
Reviewed both against the full files, not just hunks.

- `bootstrapHeadlessRenderDevice()` (`metal_render_impl.cpp:1024-1034`) — traced
  every helper it calls: `metalRenderDevice().init(device, nullptr)` invokes
  `initializeMetalRuntime`, which does create the command queue
  (`metal_runtime.cpp:107-118`), so the device is fully wired before the test
  drives `dispatchCompute`/`finish`. Confirmed `finish()` (`metal_render_impl.cpp:827-842`)
  commits + waits + opens a fresh command buffer, matching the harness's usage.
- Verified the "no `.glsl` twin needed for a Metal-only kernel" claim in the
  test comment against `metalShaderFileForStage` (`metal_pipeline.cpp:15-24`):
  it derives the `.metal` path purely from the filename string, never checks
  the `.glsl` file exists — the two new test kernels correctly resolve via
  `test/render/shaders/metal/c_r32i_{read,write}.metal` from a `.glsl`-shaped
  path.
- Both new kernels are intentionally absent from `threadgroupSizeForFunctionName`
  (per their own comments) so they fall back to a 1×1×1 threadgroup — confirmed
  the default fallback is `MTL::Size(1, 1, 1)` (`metal_pipeline.cpp:94`), and
  neither kernel is in `functionUsesImageAtomicScratch`
  (correct — plain `access::read`/`access::write`, no image atomics). With a
  1×1×1 threadgroup, `dispatchCompute(kTexDim, kTexDim, 1)` yields exactly
  `kTexelCount` threads with `thread_position_in_grid` == texel coord, matching
  the shader comments' reasoning.
- `kEmptyDistanceEncoded = 65535` matches the documented `kTrixelDistanceMaxDistance`
  clear value (`engine/render/CLAUDE.md`); binding indices (`buffer(0)`/`texture(0)`)
  are self-consistent between the two new `.metal` kernels and the test.
- The design-doc edit (`detached-revoxelize-world-light.md`) reads coherently in
  context — replaces the stale "Metal scratch-delivery gap tracked as its own
  backend issue" caveat with the no-backend-bug resolution in both places it
  appeared (Q2-REVISED block + the P4b-1 reuse-check note).
- `.fleet/plans/issue-1640.md` follows the existing `.fleet/plans/issue-<N>.md`
  convention (many precedents already tracked in the repo).

### Nits
- `metal_render_impl.cpp:1029` — `setMetalBootstrapDevice(device)` immediately
  followed by `metalRenderDevice().init(device, nullptr)` is redundant here:
  `init()` → `initializeMetalRuntime()` already does `g_runtime().device_ = device`
  (`metal_runtime.cpp:108`), overwriting what the prior line just set. Harmless
  (mirrors `MetalRenderImpl::MetalRenderImpl()`'s ctor/init split, where there
  *is* a gap between the two calls), but in this function they're back-to-back
  with no gap, so the first call is dead. Optional: drop it, or leave a
  one-line comment noting it's intentionally mirrored for pattern consistency.
- `test/render/gpu_compute_dispatch_test.cpp` — `MetalGpuComputeDispatchTest`
  has no `TearDown()`. Each `TEST_F` re-invokes `bootstrapHeadlessRenderDevice()`
  → `initializeMetalRuntime()`, which creates a **new** command queue and three
  new depth-stencil states without releasing the previous test's (only
  `shutdownMetalRuntime()` releases them, and nothing calls it between the two
  `TEST_F`s in this fixture). With only 2 tests in a short-lived process this
  is harmless today, but it's an easy leak trap for whoever adds the next
  Metal GPU test to this fixture. The GL fixture above it does clean up per-test
  (`glfwDestroyWindow`/`glfwTerminate` in `TearDown`) — consider mirroring that
  discipline with a `TearDown() { IRRender::shutdownMetalRuntimeForTests(); }`-style
  hook, or at minimum a comment noting the intentional non-cleanup.

- [x] Already run per PR body: Metal gtest filter (2 passed), full suite (1319
  passed), `format-changed` clean.

Escalation: please Opus-recheck before merge (touches: `engine/render/include/irreden/render/metal/`,
`engine/render/src/metal/` — Metal `RenderDevice` bring-up/lifecycle, a
GPU-bring-up surface I traced by reading but can't fully validate without a
Metal device to run against).

jakildev (commented):
## Opus recheck — render: Metal headless GPU vehicle-A harness + R32I cross-encoder regression guard (closes #1640)

Picking up Sonnet's escalation: *"a GPU-bring-up surface I traced by reading but can't fully validate without a Metal device to run against."* This host **is** macOS/Metal, so I focused the Opus pass on the two things a read-only pass can't settle — the `bootstrapHeadlessRenderDevice` lifetime/re-entrancy and whether the R32I contract the harness asserts is actually the one production exercises — rather than re-checking naming/structure Sonnet already covered.

### Bring-up lifetime — Sonnet's leak nit confirmed, and it's only a leak (not a crash)

Sonnet flagged the missing `TearDown`/`shutdownMetalRuntime` between the two `TEST_F`s as "harmless today but a leak trap." On closer read of `initializeMetalRuntime` (`metal_runtime.cpp:107-118`) I **confirm** the diagnosis and can now bound the failure mode precisely:

- Each `SetUp` → `bootstrapHeadlessRenderDevice()` → `MTL::CreateSystemDefaultDevice()` (fresh, owned +1) → `metalRenderDevice().init(device, nullptr)` → `initializeMetalRuntime`, which **reassigns** `commandQueue_` and the three `depthStencilState_` handles **without releasing** the prior ones. So the second test's SetUp leaks test 1's command queue + 3 depth-stencil states + the first `MTL::Device`.
- Crucially this is a *leak, not a corruption*: `initializeMetalRuntime` has no assert-on-already-initialized and no aliasing hazard — it just overwrites. `commandBuffer_` is not touched by init and is set explicitly right after (`setMetalCommandBuffer(metalCommandQueue()->commandBuffer())`), and `finish()` opens a fresh command buffer per dispatch chain. So both tests run correctly regardless of ordering; nothing depends on the leaked state. For a 2-test, short-lived process this is inert (process teardown reclaims it). **Not a merge blocker** — carried as a nit below.

### R32I contract — the harness asserts the right invariant

The escalation's real worry is whether this harness actually reproduces the *production* second-in-tick foreign-canvas read, not a strawman. I traced the two kernels against the production stages the plan names:

- `c_r32i_write.metal` uses a plain `access::write` store of a distinct-per-texel value — mirrors `c_voxel_to_trixel_stage_2`'s canonicalizing write, and distinct-per-texel (not a constant) makes a *partial* miss detectable, not just an all-or-nothing one.
- `c_r32i_read.metal` reads via `access::read` (`dist.read(px).x`) on a **fresh compute encoder sharing the same command buffer** — structurally identical to `c_bake_sun_shadow_map`'s foreign read, which is exactly candidate **C3**. The clear-to-`65535`-first + the separate oracle test (`ClearedTextureReadsBackAsClearSentinel`) give the harness a real negative control: a genuinely dropped write *would* surface as the `65535` clear sentinel and fail `EXPECT_EQ(emptyReads, 0)`. That makes "green" meaningful rather than vacuous — the point the whole no-defect ruling rests on.
- Dispatch grid is correct: the kernels are intentionally absent from `threadgroupSizeForFunctionName` (documented in the write kernel's comment), so `threadsPerThreadgroup` defaults to 1×1×1 and `dispatchCompute(W, H, 1)` yields a W·H grid — `thread_position_in_grid` **is** the texel coord. No off-by-threadgroup hazard.

### Doc + closure

The `detached-revoxelize-world-light.md` edits and the `.fleet/plans/issue-1640.md` ruling accurately fold the architect's investigated-no-defect finding (C3 refuted by repro, C1 dissolved on the resolve-before-bake ordering). The "Metal scratch-delivery gap tracked as its own backend issue" caveat is correctly retired to "no backend bug — #1626-era multi-canvas shared-state." `bootstrapHeadlessRenderDevice` has **no production caller** — it lives in `metal_render_impl.cpp` but is test-only — so the blast radius of this whole PR is the test suite + docs. `Closes #1640` unblocks #2091 on merge, as the plan notes.

I did not spin up a from-scratch macOS test reconfigure for this: the one question a run would answer (does the harness compile + pass on Metal) the author already answered green on this exact backend (2 Metal gtests + full suite 1319 per the PR body), and the lifetime concern resolves to a harmless leak by reading. Spending a full test-build of Opus budget on a test-only + doc PR with that evidence isn't warranted.

### Nits (optional, non-blocking)

- **(carried from Sonnet)** `MetalGpuComputeDispatchTest` has no `TearDown` → `initializeMetalRuntime` leaks the prior test's command queue + depth-stencil states + `MTL::Device` on the second `SetUp`. Harmless for 2 tests, but the next person adding a Metal GPU test to this fixture inherits a growing leak. A `TearDown() { IRRender::shutdownMetalRuntime(); }` (the GL fixture above already mirrors this discipline with `glfwDestroyWindow`/`glfwTerminate`) or at minimum a comment noting the intentional non-cleanup would close it. Optional — don't re-review round-trip over it.

Sonnet flagged the missing TearDown; on closer read I confirm it and additionally establish it is leak-only with no ordering/correctness dependency, so it stays a nit rather than a fix.

jakildev:
Addressed both nits:
- Dropped the redundant `setMetalBootstrapDevice(device)` call in `bootstrapHeadlessRenderDevice` — `metalRenderDevice().init(device, nullptr)` already sets `device_` via `initializeMetalRuntime`, so the first assignment was dead.
- Added `TearDown()` to `MetalGpuComputeDispatchTest` releasing the command queue, 3 depth-stencil states (via `shutdownMetalRuntime()`), and the raw `MTL::Device` (which `shutdownMetalRuntime` doesn't own) between tests — mirrors the GL fixture's per-test cleanup discipline.

Metal gtest filter (2 passed), full suite (1319 passed), format-changed clean. — worker

jakildev (commented):
## Review — render: Metal headless GPU vehicle-A harness + R32I cross-encoder regression guard (closes #1640)

**Verdict:** approve

Re-review triggered by `human:re-review` after commit `b3548f02` ("address nits on Metal headless GPU harness") — a pure nit-fix commit on top of the already Opus-rechecked head (`90c24905`).

### Prior-review resolution
- ✅ `metal_render_impl.cpp:1029` — redundant `setMetalBootstrapDevice(device)` before `metalRenderDevice().init(device, nullptr)` — verified fixed at `b3548f02`. The line is dropped; re-traced `init()` → `initializeMetalRuntime()` (`metal_runtime.cpp:107-118`) and confirmed it still sets `g_runtime().device_ = device` on its own, so nothing regresses.
- ✅ `test/render/gpu_compute_dispatch_test.cpp` — `MetalGpuComputeDispatchTest` missing `TearDown()` (leaks command queue + depth-stencil states + raw `MTL::Device` between the two `TEST_F`s), carried forward and confirmed by the Opus recheck as leak-only — verified fixed at `b3548f02`. New `TearDown()` (`gpu_compute_dispatch_test.cpp:190-200`) grabs `IRRender::metalDevice()` (the raw pointer) *before* calling `shutdownMetalRuntime()`, then releases it after. This is the correct order: `shutdownMetalRuntime()` (`metal_runtime.cpp:124-143`) releases the command queue and the three depth-stencil states but resets `g_runtime()` to a default-constructed state without releasing `device_` itself — so a caller that released the raw device *before* calling shutdown would need to re-fetch it, and a caller that never fetches it at all (the pre-fix state) leaks it. Grabbing the pointer first, shutting down, then releasing avoids both a leak and a double-release. Guarded correctly against the `GTEST_SKIP()` path (`device_ == nullptr` early-return in `TearDown`, since gtest still runs `TearDown` after a skipped `SetUp`).

No new findings — this commit is a scoped two-line + one-method diff that does exactly what both prior reviews asked for, nothing else touched.

- [x] Already run per prior PR body: Metal gtest filter (2 passed), full suite (1319 passed), `format-changed` clean. The nit-fix commit doesn't change dispatch behavior (only bring-up/teardown bookkeeping), so no new Metal-hardware run is needed beyond what's already green.

Opus recheck not required.

Issue #2207, opened 2026-07-04
render: polarity carrier for silhouette-riser faces — flipped faces light with inverted normals + per-axis store can't represent them


jakildev:
**Area:** engine/render
**Model:** fable
**Blocked by:** (none)

## Context

Root-caused during #2157 Phase A/B (evidence + measurements in PR #2190 and the #2157 thread). The #2162 silhouette-riser flip emits opposite-polarity faces for rotated content, but the polarity never reaches the rest of the pipeline — two defects with one shared root cause:

1. **Lighting normal inversion (the dominant visual defect — the "dark riser shading" / venetian bands).** The distance encoding carries only the 2-bit visible-triplet SLOT; every downstream consumer reconstructs `faceId = visibleFaceIds[slot]`:
   - `c_lighting_to_trixel` (Lambert + sky term)
   - `c_compute_sun_shadow` (normal bias + slope bias + the #2088/#2010 self-step suppression's `detectSelfStepStaircase`)
   - `c_compute_voxel_ao` (neighbour sampling along the outward normal)

   For a face emitted via the #2162 flip the true outward normal is the NEGATION of the slot-derived one → Lambert goes to ~0 and the shadow normal-bias pushes the sample INTO the caster → flipped riser faces render near-black. Deterministic repro: `IRCanvasStress --only gridspin --frozen-pose 0.6 --sweep-yaw 3.1416 3.1416 1 --debug-overlay shadow` — the venetian rows read magenta (fully sun-shadowed) while `--debug-overlay unlit` shows the same rows solid (geometry is present; the black is lighting).

2. **Per-axis store cannot represent an opposite-polarity face.** The per-axis encode (#1458: depth<<10 | frac[9:2] | slot[1:0]) is polarity-blind, and the forward scatter reconstructs every stored cell as the triplet polarity. Dual-tapping the riser plane (the #2157 cardinal fix, tried on the per-axis route) mis-places the reconstructed quad and measured NET-WORSE at non-cardinal yaw (112.5°: 180→240 px, 247.5°: 236→356 px enclosed-black). The non-cardinal residual gaps (~0.4–1.2%) therefore remain until the store/scatter can carry polarity.

## Fix options (design call — carrier allocation is a shared-resource decision)

[...]

Issue #1929, opened 2026-06-20
test: shape_debug render-verify references are red on master (21/24 fail) — refresh or investigate


jakildev:
**Area:** creations/demos/shape_debug (test/references)
**Model:** opus
**Blocked by:** (none)

## Context
While verifying #1922 (PR #1928) I ran `python3 scripts/render-verify.py --target IRShapeDebug --no-build` against the committed `creations/demos/shape_debug/test/references/manifest.json` (macos-debug refs). It reports **21 of 24 checks FAIL** — and the same 21 fail on a **clean origin/master build** (I stashed my branch's `main.cpp`, rebuilt, re-ran: identical result). So this is pre-existing on master, not introduced by #1928.

The failures are large, not noise:
- cardinal/origin shots: `zoom8_origin` 90.8% match / 28.6 dB, `zoom16_lod_fine_sphere` 85.8% / 26.8 dB, `zoom1_origin` 99.67%.
- yaw shots: `zoom4_yaw180`/`zoom4_yaw270`/`zoom8_yaw180` max_delta ~200, ~30 dB.

The references were last meaningfully touched around #1769 (2026-06-13) / #1862 (2026-06-15); the renderer has churned heavily since (re-voxelize, per-axis, lighting). So the references are most likely **stale** — but the breadth means it needs a human/architect eye to decide:

1. If current master's render is correct → refresh: `python3 scripts/render-verify.py --target IRShapeDebug --update-references --force` on macOS (and capture the linux-debug set on a Linux host), commit the new PNGs.
2. If some yaw-shot failures are genuine rotation regressions (this demo is the epic #1881 vehicle) → investigate before blindly refreshing, so a real bug isn't baked into the baseline.

Net effect today: the shape_debug render-verify gate is non-functional (any worker running it gets 21/24 false-or-real failures), so the deterministic render-regression canary is currently blind.

## Acceptance criteria
- `python3 scripts/render-verify.py --target IRShapeDebug --no-build` is green on origin/master (macOS), after either a justified reference refresh or a documented fix of any real regression found.
- A one-line note in the PR on whether any failure was a real regression vs stale reference.

jakildev:
[...]

PR #2308, opened 2026-07-08
test: headless regression guard for SDF floor self-shadow acne (#2092)

### LINKED ISSUE #2092: render: SDF floor self-shadows (acne band) with zero casters, scene-content sensitive — sun-shadow bake depth-range/bias

jakildev:
## Summary
- The #1717 C-series **receive-side facet**. The SDF floor self-shadow acne (the entire floor read magenta under residual yaw, zero casters) was fixed by **C3 (#2275)**, which unified the sun-space projection for both the caster bake and the receiver lookup so the integer-depth quantization is symmetric on a flat receiver. The re-cut plan re-confirmed the repro **no longer reproduces on current master**, so the deliverable is a positive regression guard — **not** a receive-path recovery change.
- **No shader / render-path diff.** The shared R32I `triangleCanvasDistances` carrier, `trixelCanvasPixelToWorld3DSmoothYaw`, and the sun-shadow shaders are untouched.
- `render-shadow-metric.py`: add a `--min-hole-ratio` lower bound (inverts the swiss-cheese upper bound — a fully-lit floor reads `hole_ratio ~1.0`; self-shadow acne drives it down).
- `canvas_stress` manifest: wire a `structural_only` `floor_selfshadow` extra_run that re-runs `--only floor --debug-overlay shadow` (zero casters), slices the `so3_offsnap_wide` capture (yaw π/6, the acne-classic within-sun-texel-spread pose), and asserts `hole_ratio >= 0.98`. Backend-agnostic structural oracle (structure, not pixels), so **one threshold gates macos-debug + linux-debug with no committed reference** — protecting the fixed surface while #2270 (bake coverage-splat) and #2207 (polarity carrier) churn the exact bake + distance-encoding shaders.

## Evidence
- Measured on master (macOS/Metal): `so3_offsnap_wide` capture (yaw π/6) → `shadow_px = 0`, `hole_ratio = 1.0` (all 5 SO(3) poses read 1.0). Pre-fix the entire floor read magenta (`hole_ratio ≈ 0`), so the 0.98 threshold passes with margin and fails hard on a severe regression.
- `render-verify --target IRCanvasStress` → `floor_selfshadow:floor_selfshadow_offwide:shadow  PASS`. (The default-suite + compare pixel-diff rows are the documented red-on-master references #1929/#1944/#2000 and unblessed-on-this-host compare shots — unrelated to this change, which adds no render code.)

## Test plan
- [x] `python3 -m unittest scripts.tests.test_render_shadow_metric` — 14/14 pass (adds lit-floor-passes + self-shadowed-floor-fails cases).
- [x] `ruff check scripts/` clean.
- [x] `render-verify --target IRCanvasStress` — `floor_selfshadow` structural gate **PASS** on macos-debug.
- [ ] Linux/OpenGL: the gate is `structural_only` (no committed reference), so a Linux `render-verify` run auto-engages it; the C3 fix is backend-agnostic but the Linux side is unverified on this macOS host — a Linux smoke would confirm green there.

Closes #2092


jakildev:
Pickup context: #2092 was `fleet:queued` + `fleet:opus` at claim time (19:15Z state cache + a live issue read both showed queued, no plan-review). While this PR was in flight the issue was concurrently re-labeled `fleet:plan-review` + `human:review-plan` — so its v3 re-cut plan may still be under a reviewer's vet. This PR is a faithful, empirically-verified implementation of that v3 plan (verify-the-fix-is-gone → add a positive headless structural guard → `Closes`), so it should be reviewable as-is; if the plan vet requests a different approach, the guard here can be adjusted. Note `human:review-plan` looks like a v2 leftover — the 2026-07-04 approach sign-off comment recorded it cleared, and the v3 plan states it is not required (Python metric + test-manifest gate, no shader/receive-path change). Left the plan-state labels untouched; only cleared my own `fleet:in-progress`/`fleet:claim-*` claim markers after release. — worker (opus)

jakildev (commented):
## Review — test: headless regression guard for SDF floor self-shadow acne (#2092)

**Verdict:** approve

Test/tooling-only change (`scripts/render-shadow-metric.py`, its unit tests,
and `canvas_stress`'s manifest.json) — no shader, C++ render, or ECS diff, as
the PR body claims. Verified the specific claims rather than taking them on
faith:

- `--min-hole-ratio` mirrors the existing `--max-hole-ratio` pattern exactly
  (same arg-parse shape, same `failed.append(...)` style) —
  `scripts/render-shadow-metric.py:183-187,214-215`.
- The manifest's `"min_hole_ratio": 0.98` threshold key correctly maps to the
  new `--min-hole-ratio` CLI flag via `render-verify.py`'s
  `key.replace('_', '-')` convention (`_run_structural_metric`,
  `scripts/render-verify.py:302-303`) — confirmed by reading the consumer,
  not just the producer side.
- `capture_offset: 4` in the new `floor_selfshadow` extra_run: checked
  `kShots[]` in `creations/demos/canvas_stress/main.cpp:473-485` directly —
  index 4 (0-based) is indeed `so3_offsnap_wide`, and `kShots` is assigned
  unconditionally (`g_allShots.assign(kShots, ...)`) whenever the sweep-mode
  branch isn't active, so `--only floor` doesn't perturb the shot list
  ordering the offset depends on.
- Test count checks out: 12 pre-existing `test_render_shadow_metric.py`
  cases on master + the 2 new ones (`test_min_hole_ratio_lit_floor_passes`,
  `test_min_hole_ratio_self_shadow_fails`) = 14, matching the PR body's
  "14/14 pass" claim.
- New tests assert both directions (lit floor passes at `hole_ratio=1.0`,
  fully-magenta floor fails with the correct message), and the manifest's
  `structural_only` entry is paired with a `structural` gate as
  `render-verify.py` requires (else it'd raise at manifest-parse time).

### Nits
(none)

### Praise
- Verifying the repro no longer reproduces before writing a bare regression
  guard (rather than skipping straight to "add a test") is the right call
  given the PR openly states this is a receive-path-unchanged deliverable.

- [x] Already run per PR body (`test_render_shadow_metric` 14/14, `ruff
      check scripts/` clean, `render-verify --target IRCanvasStress` PASS
      on macos-debug).

PR #2313, opened 2026-07-08
render: light-volume continuity — anchor sign fix + boundary seeding (#2310)

### LINKED ISSUE #2310: render: light-volume continuity — camera-anchor sign fix, boundary seeding, stable falloff, LINEAR sampling
**Area:** engine/render light volume (COMPUTE_LIGHT_VOLUME gather/seed, camera anchor)
**Model:** opus
**Blocked by:** none

## Symptom (user-reported, IRLightingEmissive + game scene, interactive pan/yaw, 2026-07-08)

Emissive/point/spot light contribution turns fully off, fully on, and in-between **with banding** as the camera moves. Spot behaves identically to point (cone shaping is unimplemented — tracked separately in the lighting epic).

## Root causes (confirmed by code read + `--light-boundary-sweep` captures)

1. **Camera-anchor sign bug.** `cameraAnchorVoxel()` inverts the RAW camera pan, but the pan that centers world point `P` is `-iso(P)` — so the light volume (and the 256³ occlusion grid, same helper) anchored at the **mirror** of the viewed position. Panning slid the light window away from the view at 2× the pan rate; lights died on-screen at half the designed distance. Proven empirically: the pan that demonstrably centers world (94,6) produced anchor (−94,−6).
2. **Hard window cull + 1/8-step fade band.** Lights 64–71 voxels from the anchor faded in eight discrete intensity steps (the banding) with the seed position clamped to the volume edge (glow detached from its source); ≥72 dropped to zero (the off state).
3. **Unstable falloff.** `stepFalloff` derived from the *gathered* set's max radius, so one light crossing the window boundary jumped every other light's falloff curve scene-wide.
4. **NEAREST volume sampling** rendered the integer Manhattan falloff as hard diamond shells.

## Fix (implemented on the linked PR)

- **Anchor:** invert the negated pan — the anchor is the world voxel actually viewed; occlusion grid relocates coherently (shared helper).
- **Boundary seeding replaces the clamp+fade band and the ±72 cull:** an out-of-window light seeds the per-axis-clamped boundary cell at `alpha = 1 − manhattan(origin, clamped) × stepFalloff`. Under the propagate pass's Manhattan metric, per-axis clamping makes the L1 triangle inequality an equality, so this reproduces the light's true in-window field **exactly**; contribution fades continuously, and only lights whose residual cannot reach the window are skipped.
- **Falloff stability:** max radius over ALL eligible lights (camera-independent).
- **LINEAR volume sampling** (consumer-side smoothing; seed/propagate use image ops, unaffected).
- **Observability:** perf HUD CULL block gains `LIGHTS seeded/eligible`; the lighting demo family gains `--light-boundary-sweep`.

## Acceptance

- `IRLightingEmissive --auto-screenshot 60 --light-boundary-sweep`: monotone continuous fade across the d000→d110 series; no pop, no 1/8 banding, no glow detachment.
- `IRLightingSdfBlocker` render-verify: LOS wall shadow band intact; references refreshed only for the intentional LINEAR smoothing drift.
- `IRShapeDebug` standard shots: drift confined to light-volume-lit regions; geometry/sun-shadow pixels clean (img_diff evidence on the PR).
- Both backends: GLSL + Metal seed shaders updated in lockstep; authored/verified on macOS/Metal, Linux smoke via the usual label flow.

jakildev:
## Summary
- **Camera-anchor sign fix**: `cameraAnchorVoxel()` inverted the raw pan, anchoring the light volume (and occlusion grid) at the **mirror** of the viewed position — panning slid the light window away from the view at 2× pan rate. Anchor now inverts the negated pan.
- **Boundary seeding replaces the ±72 cull + 1/8-step fade band**: out-of-window lights seed the per-axis-clamped window cell at `alpha = 1 − manhattan(origin, clamped)·stepFalloff` — exact under the propagate's Manhattan metric (per-axis clamping makes the L1 triangle inequality an equality). Contribution fades continuously; no pop, no banding, no glow detachment.
- **Stable falloff**: `stepFalloff` derives from ALL eligible lights (camera-independent), not the gathered subset.
- **LINEAR volume sampling** (consumer-side smoothing; seed/propagate are image ops, unaffected).
- **Observability**: perf HUD CULL block gains `LIGHTS seeded/eligible`; lighting demos gain `--light-boundary-sweep`.
- macOS render-verify references refreshed for the intentional LINEAR smoothing drift (2 of 5 shots; 3 byte-identical; LOS wall band verified intact).

## Evidence

Boundary sweep (emissive light walked through the window boundary — continuous fade, glow stays attached):

| in-window (d000) | boundary-seeded (d070) | faint fade (d088) |
|---|---|---|
| d000 | d070 | d088 |

shape_debug zoom8 before/after + drift localization (drift confined to light-volume-lit regions; geometry/sun-shadow pixels clean):

| before | after | drift (red) |
|---|---|---|
| before | after | drift |

## Test plan
- [x] `IRLightingEmissive --auto-screenshot 60 --light-boundary-sweep` — monotone continuous fade across the series
- [x] `render-verify --target IRLightingSdfBlocker` — LOS occlusion intact, references blessed
- [x] `IRShapeDebug --auto-screenshot` + img_diff vs pre-change master — drift localized to lit regions
- [x] macos-debug builds: IRLightingEmissive, IRLightingSdfBlocker, IRShapeDebug
- [ ] Linux/OpenGL smoke (GLSL twin updated in lockstep; authored on macOS)

## Notes for reviewer
The exactness argument for boundary seeding is in the plan file (`.fleet/plans/issue-2310.md`) and the `gatherLightSources` doc comment. The anchor sign fix affects only camera-panned scenes — all committed reference shots use near-origin pans, which is why 3/5 render-verify shots stay byte-identical.

Closes #2310

https://claude.ai/code/session_019ykrFz45ZntYKQ1WQFQ25Y


jakildev (commented):
## Review — render: light-volume continuity — anchor sign fix + boundary seeding (#2310)

**Verdict:** blocker

Dense, well-reasoned render/lighting change with a genuinely elegant fix at
its core — but it breaks the existing test suite's compile, so master
wouldn't build the test target as-is. Ran an independent math check on the
core claims plus a dedicated render-invariant subagent pass; both corroborate
below.

### Blockers
- `test/render/per_canvas_light_scope_test.cpp:76,80,100,104,127,132,161,176,191`
  — all nine call sites still use the **old** `gatherLightSources` signature
  (`out, canvas, origin, warnedOOBOrigins, outMaxRadius`), but
  `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:156`
  changed the signature to `(out, canvas, origin, outMaxRadius, outEligible)`
  — the `warnedOOBOrigins` param was deleted and `outEligible` (a new
  `std::uint32_t&`) was added in its place. The test still declares a local
  `std::unordered_set<std::uint64_t> warned` and passes it as the 4th arg,
  which no longer binds to `int&`, and never supplies the new 5th arg. This
  is wired into the `IrredenEngineTest` target
  (`test/CMakeLists.txt:58`), so it won't compile. The PR's test plan only
  covers demo builds (`IRLightingEmissive`, `IRLightingSdfBlocker`,
  `IRShapeDebug`) — the unit test target was never rebuilt after the
  signature change. Fix: update all nine call sites to the new signature
  (drop the `warned` set declaration, add an `std::uint32_t eligible` out-param).

### Needs-fix
- `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:219-235`
  — the boundary-seeded cell (`volumeOriginVoxel + clamped`) is chosen purely
  by camera position, with no check against solid/occluder geometry at that
  cell. `c_propagate_light_volume.glsl`'s neighbor-pull is occlusion-gated
  symmetrically, so if the clamped cell itself lands inside a wall, every
  neighbor refuses to pull from it and the light's contribution silently
  drops to zero even though `seedAlpha > 0` was computed as if the path were
  clear — reproducing the exact "light pops off" artifact this PR sets out
  to eliminate. The old ±72-margin code had a narrower version of this risk;
  this PR widens exposure because *any* out-of-window light now remaps to a
  clamped cell (not just ones within an 8-voxel overflow band), and that
  remap point sweeps continuously as the camera pans across a wall. Worth an
  explicit test: pan the camera so the window boundary sweeps across a wall
  with a light on the far side.

### Nits
- `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:156-215`
  — deriving `stepFalloff`/iteration count from *all* eligible lights
  (not just the seeded subset) removes the "boundary-crossing light
  reshuffles everyone else's falloff" pop, which is a real fix — but it also
  means a single very-large-radius light anywhere on the canvas, even one
  that never seeds (permanently out of reach), now stretches every other
  light's decay curve. The old code derived the max from the gathered
  (actually-seeded) subset only. Worth confirming this trade is intended for
  multi-light scenes with wide radius variance.

### Praise
- The camera-anchor sign fix is correct: independently re-derived the
  inverse-iso algebra from `getCameraPosition2DIso()`'s contract
  (`canvasPixel = origin + floor(cameraIso) + pos3DtoPos2DIso(world)`) and
  confirmed the old formula computed the mirror point, matching the "slides
  away at 2× pan rate" symptom in #2310.
- The boundary-seeding exactness argument checks out under independent
  derivation, not just by trusting the doc comment: per-axis,
  `|O_i − C_i| = |O_i − N_i| + |N_i − C_i|` holds exactly for every in-window
  `C_i` (the clamped point `N_i` always lies between `O_i` and `C_i` on the
  clamped axis), so `manhattan(O,C) = manhattan(O,N) + manhattan(N,C)`
  unconditionally — the seeded residual really does reproduce the exact
  truncated field, not an approximation.
- GLSL/Metal shader twins for both `c_seed_light_volume` and
  `c_propagate_light_volume` stay byte-for-byte parallel (same field rename,
  same clamp-and-write); `GPULightSource`'s std430 layout is untouched
  (still 64 bytes — only a padding lane's meaning changed).
- `gatherLightSources`'s two-pass gather has no per-frame allocation
  (`lightStaging_` reserved once at `create()`, only `clear()`ed per tick)
  and no O(n²) cost.

- [ ] Fix the nine stale call sites in `per_canvas_light_scope_test.cpp` and
      build/run `IrredenEngineTest`
- [ ] Add or manually verify a boundary-seed-behind-occluder case (camera pan
      sweeping the window edge across a wall with a light beyond it)
- [ ] Linux/OpenGL smoke (already flagged as pending in the PR body)

Escalation: author-agent to address, then re-request review. Opus recheck
required: touches `engine/render/` lighting-stage math (boundary-seeding
exactness argument, cross-backend shader parity, occluded-boundary-cell edge
case, and the eligible-vs-seeded falloff-derivation trade-off all warrant a
second set of eyes even after the blocker above is fixed).

jakildev:
Fixed the **blocker** only (compile break): updated all nine `gatherLightSources` call sites in `test/render/per_canvas_light_scope_test.cpp` to the new `(out, canvas, origin, outMaxRadius, outEligible)` signature — dropped the now-unused `warned` set, added an `eligible` out-param, dropped the now-unused `<unordered_set>` include. `IrredenEngineTest` builds and the full suite is green (1317 passed, 1 expected skip on this Metal host).

**Did NOT address the Needs-fix item** (occlusion-gated boundary cell): the boundary-seeded cell is chosen by camera position alone with no check against the GPU-resident occlusion bitfield, so a clamped cell landing inside a wall would still seed a nonzero `seedAlpha` while the propagate shader's symmetric occlusion gate silently drops it — reproducing the pop-off artifact. There's no CPU-side voxel/SDF-blocker occupancy query available at gather time to check this against without either a GPU readback sync point or a new spatial query surface — that's a real design question, not a mechanical fix, and matches the review's own "Opus recheck required" escalation. Left `fleet:blocker` cleared (the compile break was the actual do-not-merge issue) but this needs-fix concern is still open for the next reviewer pass.

Also unaddressed (nit, optional, non-blocking): the eligible-vs-seeded falloff-derivation trade-off question. — worker

jakildev (commented):
## Review — render: light-volume continuity — anchor sign fix + boundary seeding (#2310)

**Verdict:** needs-fix

Re-review triggered by `fleet:changes-made` after commit `cf8572d0` ("test: fix stale gatherLightSources call sites (#2313 blocker)") on top of the prior `bde7268b` review.

### Prior-review resolution
- ✅ `test/render/per_canvas_light_scope_test.cpp:76,80,100,104,127,132,161,176,191` — stale `gatherLightSources` call sites (old `(out, canvas, origin, warnedOOBOrigins, outMaxRadius)` signature) — verified fixed at `cf8572d0`. All nine sites now pass `(out, canvas, origin, maxRadius, eligible)`, matching the current signature at `system_compute_light_volume.hpp:156-162`. Built `IrredenEngineTest` clean and ran `PerCanvasLightScopeTest.*` directly: 6/6 passed.
- ❌ `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:219-235` (unchanged since the prior review — line numbers still line up) — the boundary-seeded cell is still chosen by camera position alone with no occlusion check; a clamped cell landing inside a wall still emits `seedAlpha > 0` while the propagate shader's symmetric occlusion gate drops it, reproducing the pop-off artifact. **The author's own PR comment confirms this was deliberately not addressed** — no CPU-side occupancy query is available at gather time without a GPU readback sync point or a new spatial query surface, and the comment correctly characterizes this as a design question, not a mechanical fix. Re-flagged below; not a regression, just still open.
- ❌ `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:156-215` nit (stepFalloff derived from all-eligible vs. seeded-only lights) — unaddressed per the author's comment, also confirmed unchanged in the diff. Re-flagged below as a nit (optional).

### Needs-fix
- `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:219-235` — carried forward from the prior review: the boundary-seeded cell has no occlusion check against the GPU-resident occlusion bitfield, so a clamped cell inside a wall still seeds `seedAlpha > 0` and the light silently drops out at the propagate stage. This is exactly the artifact class the PR sets out to fix, just moved to a wider surface (any out-of-window light now remaps to a clamped cell, not only ones inside the old ±72-margin band). Given there's no cheap CPU-side occupancy query today, this is a real design question — worth an explicit architect call on whether to (a) accept the residual pop as a documented limitation for now, (b) add a lightweight CPU-side occluder query, or (c) defer via a follow-up issue with a `fleet:human-deferred` decision. Whichever route, it should be a conscious call rather than silently shipping — recommend escalating to Opus/architect rather than asking the author to solve the GPU-readback-sync problem inline.

### Nits
- `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:156-215` — `stepFalloff`/iteration count derives from *all* eligible lights rather than just the seeded subset, so one very-large-radius light anywhere on the canvas (even one that never seeds) stretches every other light's decay curve. Worth a one-line confirmation in the doc comment that this trade is intended for multi-light scenes with wide radius variance, otherwise unchanged from the prior review.

- [ ] Resolve the occlusion-gated boundary cell question (needs-fix above) — either implement the check, or get an explicit human/architect sign-off to defer it with a follow-up issue
- [ ] Linux/OpenGL smoke (still pending per PR body; GLSL twin already updated in lockstep)

Escalation: author-agent (or architect, given the needs-fix item's design-decision shape) to address, then re-request review. Opus recheck required: touches `engine/render/` lighting-stage math (the boundary-seeding occlusion gap is exactly the kind of subtle invariant this checklist flags for a second set of eyes).

jakildev (commented):
## Opus recheck — render: light-volume continuity — anchor sign fix + boundary seeding (#2310)

Picking up Sonnet's needs-fix escalation, which explicitly delegated the call to Opus/architect: *"recommend escalating to Opus/architect rather than asking the author to solve the GPU-readback-sync problem inline"*, with three routes offered — (a) accept the residual as a documented limitation, (b) add a CPU-side occluder query, (c) defer via a follow-up issue. I focused this pass on the two things a read-only pass can't settle without independent derivation: whether the boundary-seed exactness math and the CPU↔GPU struct field-meaning change are actually correct across both backends, and whether the occlusion residual is truly a merge blocker.

### The needs-fix (occluded boundary-seed cell) — real, but not a merge blocker

Sonnet flagged that the clamped boundary-seed cell (`system_compute_light_volume.hpp:219-235`) is chosen by camera position with no occlusion check, so a clamped cell landing inside a wall seeds `seedAlpha > 0` while the propagate pass's symmetric occlusion gate drops it → the light pops off. **On closer read I confirm the mechanism but disagree that it blocks merge — my ruling is (a)+(c): accept as a documented limitation and track a follow-up.** Reasoning:

- **Strictly better than master, never worse.** The prior code hard-culled beyond `±(64+8)` (guaranteed zero contribution) and clamped within the 8-voxel overflow band with the *same* un-occluded boundary seed. So every light that could hit this residual was already either dropped or clamped-without-occlusion-check on master; the PR only widens the *continuous-fade* surface, not the occlusion gap. For the common (unoccluded) case it removes a guaranteed pop/band.
- **Benign failure mode.** The propagate pass is symmetrically occlusion-gated, so an occluded seed cell *traps* its alpha — it never leaks through the wall. The worst case is a missing light (under-lighting), not corruption or light-through-wall. That is a far milder failure than the artifact class the PR is eliminating.
- **Fundamental to the finite-window design, which is already an accepted limitation.** The clamped cell sits on the ±64 window edge; occluders outside the window are unknowable at gather time (the PR's own doc comment says so), and a correct fix needs a CPU occlusion mirror or a GPU→CPU readback sync — a substantial new subsystem, disproportionate for a continuity fix and out of scope. Keeping needs-fix would block a good, verified fix on an open-ended hard problem that Sonnet, the author, and I all agree should not be solved inline.

Recommend the author/human file a lighting-epic (#1717) follow-up for an occlusion-aware boundary seed; I've carried it as a doc nit below so it isn't shipped silently.

### Confirmations Sonnet couldn't fully settle (independent derivation)

- **Boundary-seed exactness** — re-derived, not trusted: with per-axis `N = clamp(O)`, for any in-window cell `C` each axis splits exactly (`N_i` lies between `O_i` and `C_i`), so `manhattan(O,C) = boundaryDist + manhattan(N,C)`. The seed at `N` carrying `1 − boundaryDist·stepFalloff` therefore reaches `C` at `1 − manhattan(O,C)·stepFalloff` — the exact unbounded field. Confirmed.
- **stepFalloff consistency (the exactness argument depends on it)** — the gather computes `stepFalloff = 1/max(outMaxRadius, cap)` and the host sets `params_.stepFalloff_ = 1/propagateIterations_` with `propagateIterations_ = max(maxRadius, cap)`. Same `maxRadius`, so the CPU seed discount and the GPU propagate decrement use an identical step. If they had drifted the exactness would silently break — they don't.
- **CPU↔GPU struct field-meaning change is consistent on both backends** — `coneAndPad_ → coneAndSeedAlpha_` (x=cone, y=seedAlpha, zw=pad) in `ir_render_types.hpp`, and *both* `c_seed_light_volume.glsl` **and** `c_seed_light_volume.metal` rename the field and now `clamp(coneAndSeedAlpha.y, 0, 1)`-write it as the seed alpha (previously hardcoded `1.0`). std430 layout is byte-identical (still `vec4` at the same offset; `static_assert(sizeof == 64)` intact) — only a padding lane's semantic meaning changed. No CPU/GPU mismatch, no binding-point drift.
- **In-window lights unchanged** — boundaryDist=0 → seedAlpha=1.0, matching the old alpha=1.0 seed. The `seedAlpha <= 0` skip correctly drops unreachable lights. No hot-path allocation: two passes over the same reserved `lightStaging_` (pass 1 counts eligible + maxRadius, pass 2 pushes within reserved capacity).
- **Camera-anchor sign fix** — concur with Sonnet: inverting the negated pan (`x=(pan.x+pan.y)/2, y=(pan.y−pan.x)/2`) is correct; the old form anchored the mirror point, matching the 2×-pan-rate slide symptom.

Tests: the nine stale `gatherLightSources` call sites are fixed to the new signature and Sonnet built `IrredenEngineTest` green (6/6 `PerCanvasLightScopeTest`). Linux/OpenGL smoke is still pending (PR body) — tagging `fleet:needs-windows-smoke` (macOS-authored, OpenGL tier uncovered).

### Nits (optional, non-blocking)

- `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:200-235` — add a one-line doc note in `gatherLightSources` that the clamped boundary seed is **not** occlusion-checked at the clamped cell (a light whose clamped edge cell lands in solid geometry silently drops — a documented residual of the finite window, benign under the symmetric propagate gate), and reference a lighting-epic (#1717) follow-up for an occlusion-aware seed. Documents the accepted limitation so it isn't silently shipped.
- **(carried from Sonnet)** `system_compute_light_volume.hpp:156-215` — `stepFalloff`/iteration count derives from *all* eligible lights, not just the seeded subset, so one very-large-radius light anywhere on the canvas (even one that never seeds) stretches every other light's decay curve. This is deliberate (camera-independent falloff — the right call) but worth a one-line doc confirmation that the wide-radius-variance multi-light trade is intended.

Sonnet flagged the occlusion gap as needs-fix and escalated the accept/fix/defer call to me; on closer read I confirm the mechanism, establish it is strictly-better-than-master with a benign (no-leak) failure mode fundamental to the accepted finite-window design, and rule accept-and-defer — downgrading it to a documented-limitation nit rather than a blocker.

jakildev:
Addressed the two doc nits from the Opus recheck (approve-with-nits) — comment-only, no behavior change:

- `system_compute_light_volume.hpp` boundary-seed block: documented that the clamped boundary cell is **not** occlusion-checked, so a clamp into solid geometry silently drops the light (benign under the symmetric propagate gate — a missing light, never light-through-wall). Referenced the occlusion-aware-seed follow-up **#2330**.
- `gatherLightSources` header comment: confirmed the all-eligible `stepFalloff` derivation (shared decay curve) is an intended trade for wide-radius-variance multi-light scenes; per-light falloff lands with the winning-light ID channel (#2318).

Follow-up **#2330** filed (no labels, awaiting human triage): it corrects the PR's original deferral note — the CPU occupancy data *does* exist (`BUILD_LIGHT_OCCLUSION_GRID` maintains a current-frame CPU mirror of both the voxel + blocker bitfields and runs before this system), so the open item is the **seed-placement policy under occlusion** (skip / nearest-free / accept-pop), not data availability.

Build skipped: diff is exclusively comment additions (verified) — no compile surface touched. `fleet:approved` stands. — worker (opus, macOS)

jakildev (commented):
## Review — render: light-volume continuity — anchor sign fix + boundary seeding (#2310)

**Verdict:** approve

Re-review triggered by `human:re-review` after commit `61867939` ("render: document light-volume boundary-seed occlusion residual + falloff trade") on top of the prior Opus-recheck approve-with-nits at `61867939`'s parent.

### Prior-review resolution
- ✅ `test/render/per_canvas_light_scope_test.cpp` (nine stale `gatherLightSources` call sites) — verified fixed at `cf8572d0`; confirmed still intact at HEAD.
- ✅ `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp:219-235` (occlusion-gated boundary cell) — Opus ruled accept-and-defer (strictly-better-than-master, benign no-leak failure mode) rather than a merge blocker, downgrading to a doc nit. That nit is now addressed: `gatherLightSources` gained a 12-line comment directly above the `push_back` documenting the residual, its benign failure mode, and the deferral to #2330 (verified **OPEN**, correctly scoped: "occlusion-aware light-volume boundary seed (clamped seed inside a wall pops off)") under the lighting epic #1717 (verified **OPEN**).
- ✅ `system_compute_light_volume.hpp:156-215` (stepFalloff derived from all-eligible vs. seeded-only lights, Sonnet nit carried by Opus) — addressed with an 8-line doc block confirming the shared-curve trade is deliberate for wide-radius-variance multi-light scenes, referencing the per-light-falloff follow-up #2318 (verified **OPEN**, correctly scoped: "spot cone shaping via winning-light ID channel").

### New commit review (`61867939`, comment-only)
- Diff is exactly 19 insertions, zero deletions, one file (`system_compute_light_volume.hpp`) — matches the author's "comment-only; no behavior change, no compile surface touched" claim. No `push_back`/control-flow/signature changes.
- Cross-checked the ordering claim in the new comment ("the CPU mirror BUILD_LIGHT_OCCLUSION_GRID already maintains, which runs before this system"): `BUILD_LIGHT_OCCLUSION_GRID` (line 169) precedes `COMPUTE_LIGHT_VOLUME` (line 179) in the `SystemName` enum (`engine/system/include/irreden/system/ir_system_types.hpp`), which drives pipeline execution order in this engine — claim holds.
- Both referenced issues (#2330, #2318) and the epic (#1717) exist and are open with titles matching their in-comment descriptions — no phantom-precedent risk.

Nothing left open from either the Sonnet or Opus passes. Verdict moves from Opus's approve-with-nits to plain approve since both nits are now resolved.

- [ ] Linux/OpenGL smoke — still pending per PR body (GLSL twin already updated in lockstep with the original `bde7268b` commit; tagging below since this is macOS-authored render code).

Escalation: none. Safe for merge — this was a doc-only delta on top of an already Opus-rechecked change; no new surface to re-escalate.

Opus recheck not required.

Issue #2322, opened 2026-07-08
render: world-placed detached canvases receive world lighting by default (D1)


jakildev:
**Model:** sonnet
**Part of epic:** #2314
**Plan file:** `~/.fleet/plans/issue-2314.md` (full epic plan)
**Blocked by:** (none)

## Scope
World-placed detached canvases should participate in world lighting BY
DEFAULT. Today the receive path additionally requires per-spawn-site
lighting-archetype membership (`C_CanvasAOTexture` +
`C_TrixelCanvasRenderBehavior`); a spawn site that forgets them silently
composites raw albedo — the "no lighting interaction on detached entities /
pasted-on" report (2026-07-08). Fold membership into the world-placed spawn
path; `screenLocked_` keeps the overlay opt-out byte-identical.

## Approach
- Attach the two lighting-archetype components in the detached re-voxelize
  world-placed spawn path (NOT the shared `kVoxelPoolCanvas` builder — it
  also builds the main canvas).
- Cross-system audit of every detached spawn site (engine demos + editor)
  to remove now-redundant manual attachments.
- `screenLocked_` overlays never gain the components (no lighting cost, no
  behavior change).

## Acceptance criteria
- A bare world-placed detached spawn receives sun shadow + light volume
  with zero extra components at the call site; screenLocked_ overlay path
  byte-identical.
- canvas_stress detached shots verified at the yaw matrix, both backends.

## Dependencies
None (independent of the V/S series). Cast-on-Metal remains #2091 (blocked
on #1640) — receive-by-default is orthogonal.

## Gotchas
- The membership components carry GPU RAII (AO texture) — default-attaching
  them costs one AO texture per detached canvas; confirm the footprint is
  acceptable for canvas-heavy scenes or gate on worldPlaced_ only (it is).
- Existing spawn sites that already attach them must not double-attach.

## References
Epic #2314; engine/prefabs/irreden/render/CLAUDE.md §"World-integration
mechanics — prerequisite: lighting-archetype membership"; #2080 (receive
facet), #2091 (Metal cast).

jakildev:
## Plan: render — world-placed detached canvases receive world lighting by default (D1)

- **Issue:** #2322
- **Model:** sonnet
- **Date:** 2026-07-08
- **Epic:** #2314 — see `~/.fleet/plans/issue-2314.md` for full context
- **Blocked by:** (none)

### Verified current state (2026-07-08 code read)
[...]

Issue #363, opened 2026-04-28
Per-canvas light scope (C_LightSource currently affects every canvas)


jakildev:
**Follow-up from the lighting-fidelity-polish PR (audit finding #11).**

## Problem

Every `C_LightSource` entity contributes to every canvas with a `C_CanvasLightVolume`. There's no way to scope a light to a specific canvas — useful when you have, say, a UI/inset canvas that shouldn't be lit by world-space lights.

## Likely fix

Option A — explicit scope tag: a new component or field on `C_LightSource` naming the target canvas.

Option B — `CHILD_OF` relation: lights become children of the canvas they affect; the lighting systems iterate per-canvas via the existing `RelationParams` machinery.

Option B is preferred — it composes with the rest of the engine's relation-based scoping and falls out of existing system patterns.

## Acceptance

A two-canvas demo where a light parented to canvas A doesn't appear in canvas B's lighting output.

PR #2337, opened 2026-07-09
render: spot cone shaping via winning-light ID channel (#2318)

### LINKED ISSUE #2318: render: spot cone shaping via winning-light ID channel in the light volume (L2)
**Model:** opus
**Part of epic:** #2314
**Plan file:** `~/.fleet/plans/issue-2314.md` (full epic plan)
**Blocked by:** #2310

## Scope
Spot lights currently behave as point lights: `direction_` / `coneAngleDeg_`
are uploaded in `GPULightSource` but never read by any shader. Implement real
cone shaping via a winning-light ID channel in the light volume — the
human-picked mechanism (2026-07-08): O(1) per pixel, no ray marching.

## Approach
- Parallel 128³ R8UI winning-light ID image alongside the RGBA8 ping-pong
  pair; seed writes the light index, propagate carries the winning
  candidate's ID with its color/alpha.
- `c_lighting_to_trixel` samples the ID at the surface cell; when the
  winning light is SPOT, applies the analytic cone factor from the light
  list (direction/cone angle already uploaded). Occlusion inherited from the
  flood — no marching.
- Also future-proofs per-light falloff (follow-on, out of scope here).

## Acceptance criteria
- Spot renders a cone (not a sphere) at cardinal + ~30° + 45° yaw, both
  backends; a spot-cone shot joins the lighting demo family.
- Scenes with no SPOT lights byte-identical (ID channel present but
  consumer branch unused).
- Perf: LIGHT-VOL HUD budget quoted before/after (one extra image r/w per
  propagate step).

## Dependencies
Blocked by #2310 (boundary-seeded gather is the substrate; PR #2313 in
flight). Independent of the V-series.

## Gotchas
- GPU binding budget is FULL (0–30) — the new image must transiently reuse
  an occupied slot per the engine/render CLAUDE.md gotcha, or ride an
  existing texture unit; the plan picks the exact slot strategy.
- Ping-pong: the ID channel needs the same read/write swap discipline as
  the color volume or IDs desync from colors mid-chain.

## References
Epic #2314; #2310 / PR #2313; `c_seed_light_volume.*`,
`c_propagate_light_volume.*`, `c_lighting_to_trixel.*` (both backends).

jakildev:
## Summary
Epic #2314 (L2). Spots were point lights — `direction_` / `coneAngleDeg_` uploaded but never read. This adds real cone shaping via a **winning-light ID channel** in the light volume — the human-picked O(1)-per-pixel mechanism, no ray marching.

- **ID volume**: a parallel RGBA8 ping-pong pair on `C_CanvasLightVolume` (`id/255` in `.r`, 0 = none). The seed writes each light's index, the propagate carries the winner's ID inward in lockstep with the color/alpha, and `LIGHTING_TO_TRIXEL` fetches it at the surface cell → applies an analytic cone factor when the winner is SPOT. Occlusion is inherited from the flood.
- **True apex**: `GPULightSource` gains a true-origin field (80 B/light) so an out-of-window spot's cone orients from the real apex, not the clamped boundary seed cell.
- **Byte-identity + zero cost when no spot**: a has-SPOT flag in `LightVolumeParams.worldOriginVoxel_.w` (coherent uniform branch) skips the ID ops in seed/propagate and the consumer read entirely, so scenes with no seeded SPOT are byte-identical and pay no extra propagate bandwidth.
- Both backends (4 GLSL + 4 Metal shaders).

## Format deviation from the plan (R8UI → RGBA8)
The plan specified R8UI, but the engine's `TextureFormat` offers no R8UI/R16UI, and **R32I routes through Metal's image-atomic scratch** (the #1640 gap) which makes cross-dispatch ping-pong reads unreliable. Normalized **RGBA8** round-trips across the 32 propagate dispatches exactly like the color volume already does. The `.r`-only 0–255 range caps spot-capable light IDs at 255 (a non-issue for the "few dozen lights" workload; documented on the component).

## Screenshots — lighting_spot cone (macOS/Metal)
The spot demo now dims the sun so the cone pool reads clearly. Cardinal vs 45° yaw show the cone stays **world-oriented** (same floor position) as the camera rotates — a camera-locked/mis-oriented cone would slide.

| yaw 0° | yaw 45° | zoom |
|---|---|---|
| yaw0 | yaw45 | zoom |

Before this PR the same spot floods an omni sphere (point-like); after, only the downward cone.

## Perf
`ComputeLightVolume` GPU stage ≈ **6.0 ms** with a radius-32 spot (dominated by the pre-existing 32-iteration flood); `LightingToTrixel` ≈ 0.05 ms. The ID channel adds +2 image ops per propagate cell **only when a spot is present** (gated off otherwise).

## Test plan
- [x] `fleet-build --target IRShapeDebug` (Metal) — shaders compile; no-spot path renders, no crash
- [x] `fleet-build --target IRLightingSpot` — cone pool renders at yaw 0/30/45
- [x] `IrredenEngineTest --gtest_filter=PerCanvasLightScopeTest.*` — 8/8 pass (new: has-SPOT flag + true-origin-vs-clamped-seed)
- [ ] GL smoke (cross-host, Linux) — authored on macOS

Closes #2318


jakildev:
Merger: base PR #2313 merged; re-targeted this PR to `master` but the rebase onto current master conflicts. The child commits cannot be replayed cleanly — worker will resolve semantically.

Conflicted files:
- `engine/prefabs/irreden/render/systems/system_compute_light_volume.hpp` — master: `5c242324 render: light-volume continuity — anchor sign fix + boundary seeding (#2310) (#2313)`; PR: `465ab1fd render: spot cone shaping via winning-light ID channel (#2318)`

— fleet merger

jakildev:
Note for reconciliation: issue #2318 carries a stale `fleet:claim-mac-worker-1` label with no active FS claim and no competing PR — my worker-4 claim TTL lapsed during the ~1h implementation and worker-1 briefly grabbed then released it (retaining the label). This PR is the complete implementation. — worker-4

jakildev:
Resolved semantic conflict via inherited-prefix drop: base PR #2313 (issue #2310) merged to master, so I replayed only the genuine #2318 commit onto current master (`git rebase --onto origin/master <fork-point>`), dropping the three inherited #2310/#2313 commits. My #2318 changes to `ir_render_types.hpp` (GPULightSource 64→80B) and `c_lighting_to_trixel.{glsl,metal}` (spot-cone struct/factor/gate) auto-merged cleanly against master's #2207 polarity-carrier + #2322 world-lighting additions — all in disjoint file regions, no symbol overlap. Verified: IRLightingSpot builds clean + runs headless on macOS/Metal; the winning-light cone pool renders and stays world-oriented at yaw 0/45 (no regression). Screenshot URLs re-pinned to the rebased head. Reviewer please re-evaluate the rebased diff. — worker

jakildev (commented):
## Review — render: spot cone shaping via winning-light ID channel (#2318)

**Verdict:** approve

Stack: standalone (`baseRefName` = `master`). This PR was semantically rebased onto master after base PR #2313 (issue #2310) merged — the worker replayed only the genuine #2318 commit onto current master and reported a clean auto-merge with #2207 (polarity carrier) and #2322 (world-lighting) since the touched regions are disjoint. Reviewing the rebased head (`fdab38ee`). `mergeable` is `MERGEABLE`, no CONFLICTING-state churn audit needed. No prior fleet review exists on this PR — the `fleet:changes-made` label that triggered pickup was from the semantic-conflict rebase, not a feedback round-trip, so this is a first-pass review, not a re-review.

### Summary of what I verified

This adds real SPOT cone shaping via a parallel RGBA8 winning-light-ID ping-pong pair on `C_CanvasLightVolume`, gated by a has-SPOT flag packed into `LightVolumeParams::worldOriginVoxel_.w`. I read every touched C++ system/component and all 8 shader files (4 GLSL + 4 Metal) in full, plus dispatched two independent render-invariant passes to cross-check:

- **CPU/GPU struct sync** — `GPULightSource` grew 64B→80B (`trueOriginVoxel_`), guarded by `static_assert(sizeof(GPULightSource) == 80, ...)`. All 5 members are `vec4`, so no 16-byte-boundary padding traps. C++, GLSL (seed + lighting_to_trixel), and both Metal mirrors declare the same 5 members in the same order.
- **Binding-point parity** — every image unit (0–3 for clear/seed/propagate ID textures, 7 for the consumer's `lightVolumeId`) and SSBO slot (4, `LightSourceBuffer`) agree across GLSL `binding=N`, Metal `[[buffer(N)]]`/`[[texture(N)]]`, and the C++ `bindAsImage`/`bindBase` call sites. The apparent binding=7 "collision" between the UBO and the new image is not real — UBO and image-unit bindings are independent namespaces on both backends.
- **GLSL/Metal parity** — complete across all 4 shader pairs, including the has-SPOT gating branch (`worldOriginVoxel.w != 0` / `params.worldOriginVoxel.w != 0`) applied identically on both backends, and the "no `radians()` in Metal" `kDegToRad` workaround in the cone-factor port.
- **Byte-identity claim for no-spot scenes** — holds end-to-end. The clear pass unconditionally zeroes the ID texture every frame (no per-frame staleness can leak forward), and seed/propagate/consumer all gate their ID ops on the *same* uniform, UBO-sourced flag — a genuinely coherent (dispatch-wide) branch, not per-thread divergent.
- **True-apex vs. clamped-seed-cell** — `gatherLightSources` correctly threads the unclamped `origin` as `trueOriginVoxel_` and the axis-clamped cell as the seed position; `spotConeFactor` on both backends reads `trueOriginVoxel_`, not the seed cell.
- **RGBA8 round-trip** — `id/255` write + `round(r*255.0)` read is an exact reconstruction for integer `k ∈ [0,255]` under RGBA8 UNORM with NEAREST filtering (confirmed against the backend texture-format enums). `winId - 1` indexing into `lights[]` is bounds-safe: it's only ever populated from `lightIndex < lightCount_` this same frame.
- **GPU lifetime** — the new ID pair rides the same `memoryBarrier` calls that already gate the color pair after clear/seed/each propagate iteration, and `volume.swap()` swaps both pairs in lockstep. No fence gap.
- **Test coverage** — `GatherHasSpotFalseWithoutSpotLights` and `GatherSpotFlagsHasSpotAndCarriesTrueOrigin` in `test/render/per_canvas_light_scope_test.cpp` assert exact clamped-seed (`63` on the clamped axis) vs. unclamped-true-origin (`70,0,0`) values and the `hasSpot` flag — real value assertions on the two headline claims, not just build/run success.
- **RGBA8-over-R8UI/R32I deviation** — the PR body's stated reason (no R8UI/R16UI in `TextureFormat`; R32I's Metal image-atomic-scratch cross-dispatch ping-pong unreliability, #1640) matches the engine's documented Metal R32I ping-pong hazard; RGBA8 round-trips exactly like the existing color volume already does. Reasonable, well-justified deviation from the plan.

### Nits
- `engine/render/src/shaders/c_lighting_to_trixel.glsl:131` and `engine/render/include/irreden/render/ir_render_types.hpp` (the `LightVolumeParams::worldOriginVoxel_` doc comment) — both still say `.w` is "reserved" / "unused" ("only `.xyz` is meaningful, `.w` is reserved" in the GLSL comment; "`.w` is reserved for a future 'snap quantum changed' or 'force re-clear' flag" in the C++ struct doc). This PR makes `.w` the load-bearing has-SPOT gate (`system_compute_light_volume.hpp`), and the Metal twin of the GLSL comment (`metal/c_lighting_to_trixel.metal:35`) *was* correctly updated to say "has-SPOT flag, #2318" — so this is a missed spot, not an intentional leftover. Low risk (both sites are pure documentation, correctness is unaffected), but worth fixing so a future reader doesn't repurpose `.w` for something else or "clean up" the has-SPOT write as dead code.

### Praise
- The has-SPOT flag design is genuinely zero-cost when absent: gated as a single uniform branch shared by seed/propagate/consumer, verified coherent (not per-invocation divergent) across both backends.
- Good instinct catching the R8UI/R32I plan deviation early and documenting the substitution rationale in both the PR body and the component header, rather than silently shipping a different format than planned.
- The yaw-sweep screenshot shots (`kSpotConeShots` in `main_spot.cpp`) directly test the claim that matters most for a "did I just build a sphere with extra steps" feature — that the cone stays world-oriented under camera rotation, not camera-locked.

- [ ] GL smoke (cross-host, Linux) — PR body already flags this as outstanding; will be tagged for cross-host smoke per policy since this touches `engine/render/` and shaders.

Escalation: please Opus-recheck before merge (touches: engine/render/ GPU struct layout, shader binding points, and the light-volume propagate pipeline — CPU/GPU sync and cross-dispatch correctness are exactly the class of subtle invariant Opus should confirm independently). Two independent render-invariant sub-passes found no blockers/needs-fix and agree on the single nit above, but per policy this class of change gets a second pass regardless.

Opus recheck required: touches engine/render/ (GPU struct layout growth, new shader binding points, light-volume propagate pipeline CPU/GPU sync).

jakildev (commented):

### Confirming Sonnet's pass

### Nits

### Praise

### Before merge

jakildev:
Addressed feedback:
- Nit: fixed the stale "`.w` is reserved" doc comment in `c_lighting_to_trixel.glsl` and the C++ `LightVolumeParams::worldOriginVoxel_` doc — both now describe the has-SPOT gate, matching the Metal twin that was already correct. Build verified (`IRLightingSpot`).
- Filed #2341 for the Opus-flagged latent multi-canvas light-volume staleness (global `LightSourceBuffer`/`LightVolumeParams` hold only the last-processed canvas's data) — non-blocking per the recheck, scope call left to the architect.

— worker

jakildev (commented):
## Opus recheck (re-review) — render: spot cone shaping via winning-light ID channel (#2318)

**Verdict:** approve

Picked up on `human:re-review`. Since my prior recheck (`fdab38ee`, approve-with-nits at 06:17Z) one commit landed — `a9e251fc` "fix stale `.w is reserved` comment for has-SPOT flag" — addressing the single nit Sonnet and I both raised. Reviewed the delta `fdab38ee..a9e251fc` only; the underlying implementation I already confirmed is unchanged.

### Nit resolution — confirmed
The delta is doc-only, two hunks, both fixing the exact staleness I flagged:
- `ir_render_types.hpp` `LightVolumeParams::worldOriginVoxel_` — the "`.w` is reserved for a future snap/re-clear flag" comment now reads "`.xyz` is the volume origin, `.w` is the has-SPOT flag (#2318)". ✓
- `c_lighting_to_trixel.glsl` — the mirrored "`.w` is reserved" comment updated to match. ✓

I also verified the flag's *real* consumers were already documented correctly in the original PR and needed no change: `c_seed_light_volume.glsl:49`, `c_propagate_light_volume.glsl:82`, and the Metal twin `c_lighting_to_trixel.metal:35` all already name the has-SPOT flag. And the remaining `.w is reserved` at `ir_render_types.hpp:1103` is a **different** struct (`LightOcclusionGridHeader`) whose `.w` is genuinely reserved — correctly left untouched.

No logic, layout, or binding change; GLSL comments are inert and Metal loads `.metal`, so no rebuild/render-verify is warranted for this delta. Nit cleared, no residual nits — clean approve.

### Before merge
- [ ] GL smoke (cross-host, Linux) — still outstanding per prior passes; `fleet:needs-windows-smoke` is set.

🤖 Opus recheck by Claude Opus 4.8 (opus-reviewer). Re-review of the nit-fix delta only; prior implementation approval stands.

PR #2352, opened 2026-07-09
render: perf_grid --wave-freeze + wave-collision sweep tier (#2332)

jakildev:
## Summary

- Adds `--wave-freeze` to `IRPerfGrid`: bakes each per-cell wave's phase-0 offset into its spawn position instead of attaching `C_PeriodicIdle`, reusing the component's own `updateValue()`/stage-mapping so the frozen value matches the live wave's easing exactly. Flagless behavior is byte-identical to master (a live wave scene is not deterministic run-to-run to begin with — confirmed pre-existing, see Verification).
- Adds `--yaw-ramp-wave`, a wider-angle pose table (4 cardinals + quadrant-0 5/10/20/30/40° + matched 10/30° spot-checks in the other 3 quadrants) for sampling the coset-collision defect (#2331) across more of the residual-angle range than the existing near-cardinal `--yaw-ramp` table covers.
- Adds a wave-freeze tier to `scripts/dev/perf-grid-rotate-sweep` (`voxel_set`/32³ + the new pose table) with a determinism gate (`img_diff --threshold 16`, rationale below).

## Verification

**Flagless byte-identity.** Master vs. dirty tree (patch-revert technique — `git stash` is gated in this worktree), same seed/params: the default wave-enabled scene is **not** byte-identical between separate process launches even on unmodified master (the wave rides wall/frame time — this is the exact premise #2332's plan documents). Confirmed the same non-determinism reproduces identically on two runs of the *same* (patched) binary, so it predates this change. As a real regularity control, `--wave-amplitude 0` (a static regular grid, C_PeriodicIdle still attached but zero-motion) **is** byte-identical across repeat runs both before and after this patch — confirming the flagless code path is logically unchanged.

**`--wave-freeze` reproduces the #2331 defect.** `--mode voxel_set --wave-freeze --grid-size 32 --yaw-ramp --yaw-ramp-wave` produces a real coverage/hole-ratio signal that grows with residual angle:

```
idx label        yaw     path     coverage  hole     perim
0   card000      0.0     single   1.0000    0.0000   26.56
1   wave_q0_p05  5.0     peraxis  0.9996    0.0004   45.32
4   wave_q0_p30  30.0    peraxis  0.9977    0.0023   43.41
5   wave_q0_p40  40.0    peraxis  0.9963    0.0037   41.43
14  wave_q3_p30  300.0   peraxis  0.9886    0.0114   91.60
```

Screenshots (card000, clean vs. wave_q3_p30, visible tearing from coset collisions):

[...]

Issue #2334, opened 2026-07-09
render: overflow-face lighting + far-quadrant validation + cost attribution (C2)


jakildev:
**Model:** opus
**Part of epic:** #2331
**Plan file:** `~/.fleet/plans/issue-2331.md` (full epic plan)
**Blocked by:** #2333

## Scope

Finish the overflow lane (#2333): light overflow faces so revealed slivers match their neighbors, prove the far quadrants, and measure the rotating-frame cost. Also decide (measure, don't assume) the optional follow-up of re-orienting the cardinal store's election metric.

## Approach

- Light overflow entries in the per-axis lighting stage (where sun-shadow map + light volume are already bound): a small dispatch over the overflow list computes world-space lighting (sun cascade sample + light-volume sample + Lambert, AO = 1.0 — mirrors `c_lighting_to_trixel`'s world sample) and rewrites each entry's stored color. The scatter keeps drawing stored color — no new texture wiring at the framebuffer stage.
- Extend the V1 sweep run to quadrants 1–3 matched poses; the 120°–240° band is where the un-yawed election metric inverts and overflow counts peak.
- Measure: per-axis GPU sub-stage rows (#2281) + overflow entry counts per quadrant on the wave scene and dense cube. If far-quadrant overflow dominates frame cost, spec the yawed-election follow-up (decouple sort key from recovery payload via winner-id indirection) as a separate issue; otherwise record the numbers and close.

## Acceptance criteria

- V1 sweep green in ALL four quadrants, both backends.
- ROI crop pairs: revealed slivers shade consistently with adjacent per-axis-lit cells (accepted drift: no screen-space AO on overflow faces — call it out in the PR body).
- Measured overflow counts + GPU delta table posted on #2331; explicit go/no-go on the yawed-election follow-up with numbers.
- Cardinal path still byte-identical; dense sweep unregressed.

## Dependencies

Blocked by #2333. Blocks D1 (docs capture the final contract).

## Gotchas

- The lighting rewrite must run after the light volume settles for the frame and before the scatter — same-frame ordering within the existing pipeline positions, no new system if avoidable.
- Don't chase AO parity on slivers — screen-space AO needs a canvas cell; the accepted-drift note in the epic covers it.

## References

- Epic #2331; #2333 (the lane this lights); `c_lighting_to_trixel.{glsl,metal}` world-sample path (P4b-2 precedent for world-space receive).

jakildev:
[...]

Issue #2332, opened 2026-07-09
test: deterministic wave-freeze + wave-scene rotate sweep for the per-axis view-visibility baseline (V1)


jakildev:
**Model:** sonnet
**Part of epic:** #2331
**Plan file:** `~/.fleet/plans/issue-2331.md` (full epic plan)
**Blocked by:** (none)

## Scope

Validation first: give IRPerfGrid a deterministic wave pose and extend the rotate-sweep harness with a wave-scene tier, so the per-axis view-visibility defect (#2331) has a red baseline the existing `--mode dense` sweep structurally cannot produce (a solid cube's per-axis exposed faces lie on flat planes with zero `(1,1,1)`-coset pairs). No engine/render changes in this child.

## Approach

- Add a `--wave-freeze` flag (or config key) to `creations/demos/perf_grid` that pins the traveling wave at a fixed phase: geometry becomes static and captures are byte-identical run-to-run. Flag absent → behavior unchanged.
- Extend `scripts/dev/perf-grid-rotate-sweep` with a wave tier: `--mode voxel_set --wave-freeze` swept across residual yaws within quadrant 0 AND matched poses in quadrants 1–3, scored with the existing `render-coverage-metric` / `render-silhouette-metric` tools.
- Record the red baseline table (per-pose coverage / hole-ratio) in the PR body and as a comment on #2331.

## Acceptance criteria

- Two runs at the same frozen pose produce byte-identical screenshots (`img_diff` = 0).
- Sweep produces the per-pose metric table; cardinal poses score high coverage while residual poses show the defect (holes/banding) — the red baseline.
- Flagless IRPerfGrid run is byte-identical to master.

## Dependencies

Blocks C1 (#2331 child 2) — C1's gate is this sweep going green at quadrant-0 residuals.

## Gotchas

- The wave animates via per-cell `C_PeriodicIdle` ticks; freezing must pin phase without desyncing the per-entity position uploads.
- `voxel_set` mode is per-cell entities (64³ = 262k) — heavy; the sweep may want `--grid-size 32` for runtime, as long as the wave still produces coset collisions (amplitude 5, λ=32 does at 32³).

## References

- Epic #2331 (root-cause analysis + math).
- #1880 / epic #1881 — the existing dense-cube rotate-sweep this extends.

jakildev:
## Plan: test: deterministic wave-freeze + wave-scene rotate sweep for the per-axis view-visibility baseline (V1)

- **Issue:** #2332
[...]

Issue #2331, opened 2026-07-09
render: per-axis store drops view-visible faces under residual yaw — view-visibility overflow (epic)


jakildev:
## Context

While the camera rotates under Z-yaw (`residualYaw != 0`), the main canvas renders through the per-axis store + forward scatter (#1309/#1310). The store keys each face by its **un-yawed (cardinal) iso pixel** and elects **one winner per cell** by `imageAtomicMin` on the **un-yawed depth `x+y+z`**. The scatter can only draw survivors, so the rendered face set while rotating is exactly the **cardinal-visible** set — but the camera views from cardinal + residual. Any two faces separated by `t·(1,1,1)` share one store cell; the loser is discarded at store time and, when the residual yaw swings it out from behind the winner (screen separation `t·(−2sinθ, 2−2cosθ)` iso px — ~1 cell per unit `t` at 30°), there is nothing to scatter. **Result: holes that appear only under rotation, even within one quadrant.**

Secondary defect in the same election: the `x+y+z` metric is never re-oriented per quadrant. Along the live view kernel the coset ordering scales with `2·cos(visualYaw)+1` — degenerate at 120°/240°, **sign-inverted** in (120°, 240°), so around the 180° quadrant the store keeps the view-farthest member of every colliding pair. (#1370 fixed the *composite* depth metric among survivors; it cannot resurrect dropped faces.)

Observed on IRPerfGrid (default `voxel_set` wave scene): hollow-looking solid cube under yaw, wave crests missing within-quadrant, diagonal banding (collision sets are periodic in `x+y+z` — the wave phase rides the store's kernel). The #1880 rotate-sweep never caught this: `--mode dense` is a solid cube whose per-axis exposed faces lie on flat planes containing **zero** `(1,1,1)`-coset pairs — structurally blind to the defect class.

Ruled out (verified): compact iso-rect cull (full-`visualYaw` projection + margin), chunk visibility bounds (conservative 8-corner yawed projection, #1439), cull-viewport camera source (`getEffectiveCameraIso()`).

## Fix

Keep the cardinal-keyed store (exact recovery, crack-free, correct for convex content) and add a **rotating-only view-visibility overflow lane** carrying exactly `viewVisible ∖ cardinalWinners`:

[...]

PR #2357, opened 2026-07-09
render: per-axis view-visibility overflow lane — view mask + overflow append + scatter draw (#2333)

### LINKED ISSUE #2333: render: per-axis view-visibility overflow lane — view mask + overflow append + scatter draw (C1)
**Model:** fable
**Effort:** xhigh
**Part of epic:** #2331
**Plan file:** `~/.fleet/plans/issue-2331.md` (full epic plan)
**Blocked by:** #2332

## Scope

The core fix for #2331: add the rotating-only **view-visibility overflow lane** to the per-axis path — a view-mask pass (yawed depth keyed by yawed screen cell), an overflow-append pass (faces that win their view cell but lost their cardinal store cell), and a second instanced indirect draw in the forward scatter over the overflow entries. Overflow faces render **albedo-only** in this child (unlit slivers beat missing geometry); lighting integration is C2. Both backends (GLSL + MSL).

## Approach

- New `resolveMode` values on the existing per-axis stage-1 kernel (the #2255 pattern — it already re-runs identical face geometry per mode): mode 2 = view-mask `atomicMin` write, mode 3 = overflow test + append.
- View mask + overflow list + counter share ONE lazily-sized scratch buffer, bound transiently on an existing binding during the per-axis window (binding-28 `kBufferIndex_PerAxisResolveScratch` precedent; budget 0–30 is full — no new permanent binding).
- Overflow entries carry a **payload, not a voxel index** — the position/color SSBOs (bindings 5/6) hold whichever canvas dispatched last by the time `TRIXEL_TO_FRAMEBUFFER` runs, so indices dangle. Pack `{camera-anchored world cell (3×10-bit signed), colorPacked, faceId + sub-cell frac}` per entry.
- A one-thread finalize dispatch copies the counter into indirect draw args; the scatter issues a second instanced indirect draw with the existing `v_/f_peraxis_scatter` shaders under an overflow uniform branch (vertex pulls the entry payload instead of the canvas cell; same deformed-quad footprint + yawed planar composite depth).
- Entity-id channel: overflow quads write the background id (hover on revealed slivers degrades — documented, accepted for this child).

## Acceptance criteria

- V1 sweep (#2332) green at quadrant-0 residuals: coverage ≥ 0.99 / hole-ratio parity with the cardinal pose on the wave scene, both backends.
- Dense-cube sweep (#1880) unregressed.
- Cardinal / yaw-0 path byte-identical (md5 A/B, flag-off comparison N/A — the lane activates only at `perAxisRoute != 0`).
- Overflow cap never hit on demo scenes; on synthetic overflow, one-shot CPU warn with dropped count (never silent).

## Dependencies

Blocked by #2332 (the acceptance harness). Blocks C2 (lighting + far quadrants) and D1 (docs).

## Gotchas

- Memory barriers between store → view-mask → overflow → #2255 election dispatches; keep the election (mode 1) LAST so its winner scratch reuse of binding 28 doesn't collide with the new scratch regions — map the per-axis-window dead-binding set as the first implementation step.
- ε tolerance on the view-mask match must cover the yawed-depth quantization step; over-emit is safe (framebuffer depth test cleans up), under-emit re-opens holes.
- Metal: no second image-atomic slot — the view mask must be an SSBO atomic region, not a texture image (same rationale as the #2255 scratch). The #1640 foreign-R32I caveat does not apply (main-canvas in-tick pattern).
- GL indirect draw args bind to `GL_DRAW_INDIRECT_BUFFER` (not an SSBO slot — no budget cost); Metal passes the indirect buffer at draw time.

## References

- Epic #2331 (math + ruled-out culprits); `docs/design/per-axis-trixel-canvas-rotation.md` (store/scatter contract).
- #2255 (two-pass election template), #1370 (yawed composite depth), #1458 (frac encoding), #2281 (GPU attribution rows).

jakildev:
## Per-axis view-visibility overflow lane (#2333, C1)

Fills the #2331 "venetian-blind" holes on rotated voxel content at non-cardinal
yaw. The per-axis cardinal store keeps only the **cardinal-visible** face set;
under residual yaw θ, coset faces `t·(1,1,1)` separate on screen and the
`x+y+z` election drops view-visible faces (sign-inverts past 120° full yaw),
leaving see-through gaps.

This adds a **view-mask → overflow-append → scatter-draw** lane on top of the
per-axis path:

- **resolveMode 2** (view mask): every per-axis face atomic-mins its quantized
  yawed depth into a shared view-mask region — view visibility competes across
  all three axis routes.
- **resolveMode 3** (overflow append): a face appends a scatter entry iff it
  wins its view-mask cell (within ε) **and** is not its cardinal cell's settled
  winner — exactly `viewVisible \ cardinalWinners`, the set the store drops.
- **finalize + indirect scatter draw**: one instanced draw fills the recovered
  faces (albedo this child; lighting is #2334).

Wired into a unified transient scratch at binding 28
(`[winnerIds][viewMask][ctrl][entries]`), allocated lazily with the per-axis
canvases and freed at the cardinal — a non-rotating scene pays nothing. Dispatch
order per rotating frame: stores → view-mask → overflow-append → election →
stage 2, with a memory barrier between each. Byte-identical at cardinal / yaw-0.

`IRPerfGrid --yaw-ramp --yaw-ramp-wave --yaw-ramp-crops --wave-freeze`,
kill-switch A/B (`IR_PERAXIS_OVERFLOW_DISABLE` gates only the overflow draw):

- **`--mode dense_set`** (watertight): 0 overflow entries, byte-identical → the
  lane is correctly **inert** when there are no dropped-visible faces.
- **`--mode voxel_set`** (sparse, exhibits #2331): the lane fills the dropped
  faces. Crop drift (ON vs OFF) grows with residual and is worst past 120°,
  exactly as predicted: q0_p30 ≈ 13%, q1_p30 ≈ 34%, q2_p10 ≈ 45%. No stipple
  regression — overflow entries sit two tie-bands behind, so they never beat an
  equal-depth cell-path face.

Before (overflow draw **off**, #2331 holes) → after (draw **on**, filled):

**q1_p30 (120°) full frame**

| before | after |
|---|---|
| before | after |

**q1_p30 crop / q2_p10 crop (ROI)**

| q1_p30 before | q1_p30 after | q2_p10 before | q2_p10 after |
|---|---|---|---|
|  |  |  |  |

GL backend can't be built on macOS (Metal host) — needs cross-host
`fleet:needs-linux-smoke`. Rebased off master after the #2332 base (#2352)
merged; the stage-1 body was ported into the #2325 `_body.{glsl,metal}` split.

Closes #2333


jakildev (commented):
## Review — render: per-axis view-visibility overflow lane — view mask + overflow append + scatter draw (#2333)

**Verdict:** approve

No prior fleet review on this PR (it was `fleet:stacked`/`fleet:wip` until
its base #2352 merged and it was rebased standalone onto `master`).
Reviewed via `gh diff`/`git show` against the remote branch refs — the
branch name was locked by another worktree (`worker-3`), so I read the full
files at the PR head via `git show origin/claude/2333-peraxis-overflow-lane:<path>`
instead of checking out.

### What this does
Adds a 3rd resolve-mode pair (`resolveMode` 2/3) on top of the existing
per-axis cardinal store (0) and #2255 winner-election (1): a shared
view-mask (per-yawed-screen-cell atomicMin of quantized depth, competing
across all three axis routes) followed by an overflow-append pass that
scatters the `viewVisible \ cardinalWinners` set into a GPU-authored
indirect-draw entry list, consumed by one extra instanced draw in
`TRIXEL_TO_FRAMEBUFFER`.

### UBO layout — verified byte-exact
Cross-checked every `std140`-append claim against the actual `offsetof`/
`sizeof` static_asserts rather than trusting the comments:
- `FrameDataTrixelToFramebuffer::overflowMode_` lands at offset 208 (four
  ints, 208–224), `sizeof == 224` — matches `v_peraxis_scatter.glsl`'s
  `overflowMode` + 3 pad ints appended at the same offset.
- `FrameDataVoxelToCanvas::overflowScratchLayout_` (`ivec4`) lands at
  offset 208, exactly packing the 192–208 std140 row left by the #2258
  feeder-partition fields (`resolveMode_` 192, `occlusionCullMipCount_`
  196, `feederSubCap_` 200, `feederPassTailBase_` 204) — matches
  `c_voxel_to_trixel_stage_1_body.glsl`'s `overflowScratchLayout` uniform.
- Both shader-side declarations (GLSL uniform block / Metal struct via the
  `.metal` diff) agree with the CPU struct field-for-field; no drift found.

### GPU dispatch/barrier ordering
`dispatchPerAxisCanvases` in `system_voxel_to_trixel.hpp` restructures the
old interleaved per-axis loop into four explicit phases (stores → view-mask
→ overflow-append → election+stage2), each documented with *why* the
barrier at its boundary is there:
- Phase A→B: `SHADER_IMAGE_ACCESS` barrier — all three distance stores
  must settle before any mode-2 mask write reads them indirectly (view
  visibility competing across axes needs every axis's geometry present).
- Phase B→C: `SHADER_STORAGE` barrier — the mask must be fully written
  before mode-3's compare.
- Phase C→D: `SHADER_STORAGE` + `COMMAND` barrier — entry data AND the
  GPU-authored `instanceCount` must both be visible before the later
  indirect draw consumes them.
- Phase D still refills the winner-scratch region (region 0) per-axis
  before its election dispatch — unchanged from the pre-#2333 code, and I
  confirmed regions never overlap: mask/append (phases B/C) only touch
  regions 1–3, phase D's fill only touches region 0.

The overflow-cap handling (`atomicAdd` reserve + back-off-and-count-drop on
cap hit) is correct and matches between GLSL (`atomicAdd(...,
0xFFFFFFFFu)` wraparound) and Metal (`atomic_fetch_sub_explicit`) —
functionally identical, no cross-backend divergence. The CPU-side one-shot
drop warning (`warnOverflowDropsIfAny`) reads the *previous* rotating
frame's already-retired ctrl block before the reset, which is a safe
unfenced read per its own comment (same pattern as the existing cull
diagnostic).

### Kill switch + verification
`IR_PERAXIS_OVERFLOW_DISABLE` env var gates only the overflow draw (mask/
append dispatches still run, so the A/B is a pure rendering-output
comparison) — a good pattern for isolating a new lane's effect. The PR
body's own A/B evidence (watertight `dense_set` → 0 overflow entries,
byte-identical; sparse `voxel_set` → holes filled, crop drift matches the
predicted 13%/34%/45% pattern) is credible and the before/after screenshots
back it up.

### Nits
- None.

- [x] Already covered by the PR's own kill-switch A/B + screenshot
      evidence.
- [ ] `fleet:needs-linux-smoke` — GL backend can't build on this macOS
      host per the PR body; still owed.

Escalation: please Opus-recheck before merge (touches: engine/render/,
compute-shader dispatch ordering, cross-frame GPU atomic scratch-buffer
reuse). I traced the barrier placement and cap-handling logic and found it
internally consistent and correctly mirrored on both backends, but
race-freedom in multi-phase atomic scratch reuse across resolveModes is
exactly the class of invariant this checklist reserves for Opus rather
than a Sonnet first pass — I'd like a second set of eyes on whether the
three barrier boundaries above are sufficient (no phase reads a region
another phase is concurrently still writing) before this lands.

Opus recheck required: engine/render/ compute-dispatch ordering + GPU
atomic scratch-buffer lifetime across resolveModes (see Escalation above).

jakildev (commented):

**Verdict:** approve

introduces nor weakens.

lost reservation.

### 5. GLSL ↔ Metal parity

needs-fix.

PR #2370, opened 2026-07-13
docs/fleet: epic-steward — #1717 heal + ledger sync

jakildev:
## Summary
- Epic-steward heal-on-first-claim for the shadow & lighting quality epic **#1717**.
- Synced the architect's consolidation plan (previously only in local staging, never committed) into the repo at `.fleet/plans/issue-1717.md`, with a `## Steward ledger` appended.
- Ledger records the 10-child membership (union of `Part of epic: #1717` back-refs), the closing-PR/state table, decisions **D1–D3**, and the heal + close-out-status events.
- Umbrella `## Children` checklist and a heal/close-out-status comment were posted directly on #1717 (not part of this diff).

## Status
- Epic is **9/10 closed**. The lone open child **#2091** (detached re-voxelize world-shadow CAST) is GL-host-gated on the #1640 Metal R32I gap, so #1717 is **not** closed this iteration.

## Test plan
- [ ] Docs-only (`.fleet/plans/`) — no build/runtime surface. Reviewer sanity-checks the ledger's child/PR/state table against the umbrella checklist.

Bookkeeping PR for epic #1717 (does not close it).

### DIFF
--- a/.fleet/plans/issue-1717.md
+++ b/.fleet/plans/issue-1717.md
@@ -1,0 +1,256 @@
+# Plan: render — shadow & lighting quality epic (consolidate the bake with the scatter face-deformation math)
+
+- **Issue:** #1717 (umbrella, `fleet:epic`, `human:approved`)
+- **Model:** opus (all consolidation children)
+- **Date:** 2026-06-27
+- **Architect:** opus-architect
+
+## Scope
+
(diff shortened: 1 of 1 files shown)


jakildev (commented):

### Confirming Sonnet's pass

### Nits

### Praise

### Before merge

PR #2372, opened 2026-07-13
fleet: timeout-guard daemon net calls + hung-lock escalation (#2362)

### LINKED ISSUE #2362: fleet-rebase: timeout-guard network git ops + stale-lock detection (3-day hung fetch starved tier-0 conflict rebasing)
**Area:** scripts/fleet (fleet-rebase, dispatcher resilience)
**Observed:** 2026-07-09 14:39 → 2026-07-12 12:57 (local). A `fleet-rebase --auto --rearm-trigger` run's child `git fetch origin +refs/heads/claude/2315-light-domain-instrumentation` hung on network I/O for ~3 days while holding `~/.fleet/state/fleet-rebase.lock.d`. Every subsequent tier-0 spawn logged `another fleet-rebase run holds …lock.d; deferring to the LLM pass`, so mechanical conflict rebasing silently stopped and 4 approved PRs sat CONFLICTING (#2352, #2348, #2347, #2325). The holder was alive (not stale by pid-check), just blocked — killing the hung fetch let the parent finish, release the lock, and re-arm normally.

## Fix

1. **Guard every network git op in `fleet-rebase` with a timeout.** Unguarded fetches at scripts/fleet/fleet-rebase lines ~250, ~265, ~303, ~309, ~363 (`git -C "$wt" fetch origin …`). Prefer `git -c http.lowSpeedLimit=1 -c http.lowSpeedTime=60 fetch` (works with ssh? no — lowSpeed* is HTTP-only) **or** a `TIMEOUT_CMD` prefix resolved once at top (`timeout`/`gtimeout`, empty fallback), e.g. 120s per fetch. A ~/bin python `timeout` shim now exists on the mac host; coreutils recommended on all hosts.
2. **Stale-lock escape hatch:** if `lock.d/pid` holder has been alive longer than a generous ceiling (e.g. 30 min for a tier-0 pass), log loudly (alerts dir?) instead of deferring silently forever — the current message reads as benign concurrency, not a 3-day outage.
3. Same audit for other lock-holding fleet scripts that shell out to the network (fleet-clone-freshness advance, queue-ingest).

## Verify
- Simulate: hold the lock with a `sleep`-child run, confirm defer message escalates after ceiling.
- `timeout 1 git fetch <blackholed-remote>` path returns 124 and the pass continues to the next PR instead of wedging.

jakildev:
## Summary
- Bounds every network call in the unattended fleet daemons so a black-holed GitHub connection fails fast instead of hanging forever — three fleet-wide outages in 4 days were all this one defect.
- **New `scripts/fleet/fleet-net.sh`**: a sourced lib shadowing `git()`/`gh()` with a coreutils `timeout` (resolved once via a `--version` probe that rejects Windows System32 `timeout.exe`), guarding every current and future net call site by construction. Sourced by `fleet-rebase`, `fleet-claim`, `fleet-dispatcher`. The dispatcher's duplicate inline `TIMEOUT_CMD` resolution is retired into the lib; the send-keys guard reuses it and inherits the coreutils probe.
- **`fleet-rebase` hung-lock escalation**: stamps the lock's acquisition time; on the defer path, an alive holder past a 30-min ceiling logs a loud `HUNG-LOCK` line and drops `~/.fleet/alerts/fleet-rebase-hung-lock` — the benign "deferring" message previously masked a 3-day wedge. Never auto-breaks an alive holder.
- **`install.sh` host protections** (idempotent, `--no-ssh-config` opt-out): a marked `Host github.com` ssh-keepalive block (skips a pre-existing user-owned block), git `http.lowSpeed*` globals, and a `timeout-shim.py` fallback installed as `~/bin/timeout` when no coreutils timeout exists.

## Test plan
- [x] `tests/test_fleet_net_guard.sh` — 18/18 (net-op guarded, local-op passthrough, `-C`/`-c` flag skip, exit-124 propagation, empty-cmd passthrough, coreutils probe accept/reject, shim `--version`/timeout/exit-passthrough).
- [x] `tests/test_fleet_rebase_hung_lock.sh` — 13/13 (alive+old → escalate + alert; alive+fresh → benign defer; dead holder → stale-break; missing stamp → no escalation).
- [x] Existing suite green: rebase inherited-drop 14/0, stale-conflict 7/0, clone-freshness 16/0, install-refresh 18/0, dispatcher github-gate 6/0, claim acquire 13/0.
- [x] `ruff check scripts/` clean; `bash -n` on all touched scripts.

## Notes
- **Activation lag:** the daemons run from the main clone's master, so this fix is inert until the PR merges + the main clone advances + the fleet bounces (`fleet-up`).
- **Exit-124 aliasing audited:** `fleet-claim` has no numeric exit-code branches on git/gh; the dispatcher's `rc` checks are on a `fleet-claim` subprocess (124 → "not assigned", safe); `fleet-rebase` uses boolean success checks (124 → "leave for LLM").

Closes #2362

### DIFF
--- a/.fleet/plans/issue-2362.md
+++ b/.fleet/plans/issue-2362.md
@@ -1,0 +1,51 @@
+# Plan: fleet-rebase: timeout-guard network git ops + stale-lock detection (+ fleet-wide net-call audit)
+
+- **Issue:** #2362
+- **Model:** opus — cross-script bash with portability subtleties (shadow functions, coreutils-vs-System32 `timeout`, ssh-config idempotency), but the approach is fully specified below; no novel design remains.
+- **Date:** 2026-07-13
+
+## Verified current state (audit @ master `3035f07b`)
+
+Root cause per the 07-13 thread comment: host TCP connections to GitHub intermittently black-hole (silent death, no RST), so any unguarded network call hangs forever. Three fleet outages in 4 days. Audit of every named surface:
+
+- **`fleet-rebase`** — 13 unguarded network sites: `git fetch` L250, L265, L303, L309, L363; `git ls-remote` L266, L272; `git push` L421; `gh pr view` L176, L203; `gh pr edit` L223, L224, L297; `gh pr list` L285. Lock (L103–129): `mkdir $STATE_DIR/fleet-rebase.lock.d` + pid file; the stale-break (L110–115) fires only when the holder pid is **dead** — an alive-but-hung holder defers forever with the benign L125 message. Confirms the outage-1 mechanism.
+- **`fleet-claim`** — ~49 `gh` network invocations plus `git fetch`/`git push --force-with-lease`. Runs **inline in the dispatcher main loop** (`reservation-role`, `planning-claim`), so one hung `gh issue view` is a fleet-wide outage — exactly outage 2 (18 h).
+- **`fleet-clone-freshness.sh`** — `advance_main_clone`'s `git fetch origin master` (L116) is `|| true`-guarded against *failure* but not against *hanging*; invoked by the dispatcher at startup and per-tick. Outage 3.
+- **`fleet-dispatcher`** — already resolves `TIMEOUT_CMD` (L372–377, `timeout` → `gtimeout` → empty) but applies it **only** to tmux send-keys, and its `command -v timeout` probe has the latent Windows System32 bug. Its remaining network exposure is via child `fleet-claim` and the sourced `advance_main_clone`.
+- **Already guarded — no work needed:** `fleet-queue-ingest`'s embedded python has `timeout=15` on all subprocess `gh` calls; `fleet_gh_poll.py` has subprocess + urllib timeouts; the scout never fetches (rev-parse only).
+- **`~/.fleet/alerts/` exists** (created by fleet-up) — an alert = file drop + loud log line for the human/babysit.
+
+## Approach
+
+One task, one PR — the lib and its consumers must land together to be testable.
+
+1. **New shared lib `scripts/fleet/fleet-net.sh`** (sourced, not executed):
+   - Resolve `FLEET_TIMEOUT_CMD` once: coreutils `timeout` → `gtimeout` → empty. Verify coreutils-ness (`--version` output) before trusting a bare `timeout` on PATH — Windows System32 `timeout.exe` is an interactive wait, not a command runner. Degrade to unguarded passthrough when none resolves.
+   - Shadow functions `gh()` and `git()` that call `command gh` / `command git` internally (no recursion). `git()` scans past global flags (`-C <path>`, `-c <k=v>`, `--foo`) to find the subcommand and applies the timeout **only** to network subcommands (`fetch`, `push`, `pull`, `ls-remote`, `clone`, `remote`, `fetch-pack`, `send-pack`); everything else passes through untouched — never time-limit a local `rebase`/`checkout`.
+   - Budget: `FLEET_NET_TIMEOUT` default **120 s**, env-overridable. `timeout` exit 124 flows into existing `|| true` / `rc=$?` handling as an ordinary failure — the desired skip-and-continue degrade the call sites already implement.
+   - Why shadow instead of per-site edits: fleet-claim alone has ~49 sites; per-site churn is unreviewable, and a shadow guards *future* call sites by construction.
+2. **Source `fleet-net.sh`** in `fleet-rebase`, `fleet-claim`, and `fleet-dispatcher` (early — before the dispatcher sources `fleet-clone-freshness.sh`, so `advance_main_clone`'s fetch resolves the shadow at call time). Missing-lib fallback: skip sourcing (graceful degrade). Retire the dispatcher's inline `TIMEOUT_CMD` resolution in favor of the lib so two copies don't drift; the send-keys guard keeps working off the same var (and inherits the coreutils probe, fixing its latent Windows bug).
+3. **`fleet-rebase` hung-lock escalation:** `acquire_lock` additionally stamps `date +%s > "$LOCK_DIR/started"`. On the defer path with an **alive** holder: if `now − started > FLEET_REBASE_LOCK_HUNG_SECS` (default **1800**), log a loud `HUNG-LOCK:` line and write `~/.fleet/alerts/fleet-rebase-hung-lock` instead of only the benign message. Do **not** auto-break an alive holder. Missing `started` file (pre-fix holder during rollout) → skip the age check.
+4. **`install.sh` host-protection section** (idempotent, mirrors the zshrc snippet pattern):
+   - (a) ssh keepalives for `Host github.com` as a **marked** block in `~/.ssh/config`, with a `--no-ssh-config` opt-out. If any existing `Host github.com` block is present (e.g. a manual edit), **skip and warn** — never merge into or duplicate a user-owned block.
+   - (b) `git config --global http.lowSpeedLimit 1000` / `http.lowSpeedTime 60` (idempotent by nature).
+   - (c) If neither coreutils `timeout` nor `gtimeout` resolves: install the python shim (`scripts/fleet/timeout-shim.py` → `~/bin/timeout`, via all three tool-registry edits) and print the coreutils recommendation.
+5. **Tests** (hermetic `tests/test_*.sh`):
+   - `test_fleet_net_guard.sh` — subcommand classification (incl. `-C`/`-c` flag skip), network op gets the prefix / local op doesn't, exit-124 propagation, empty-`FLEET_TIMEOUT_CMD` passthrough, non-coreutils `timeout` rejected by the probe.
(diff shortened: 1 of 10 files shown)


jakildev (commented):
## Review — fleet: timeout-guard daemon net calls + hung-lock escalation (#2362)

**Verdict:** approve

Reviewed the full diff in file context (not just hunks): new `scripts/fleet/fleet-net.sh` (shadow `git()`/`gh()` with a coreutils-verified `timeout`), its sourcing in `fleet-claim`/`fleet-dispatcher`/`fleet-rebase`, the `fleet-rebase` hung-lock escalation, the `install.sh` host-protection section, and the two new test suites.

**Verification performed (macOS host, this worktree):**
- `bash -n` on all seven touched/added shell scripts — clean.
- `python3 -m ast` on `timeout-shim.py` and `ruff check` — clean, matches the PR's claimed test plan.
- Ran both new suites live: `test_fleet_net_guard.sh` 18/18, `test_fleet_rebase_hung_lock.sh` 13/13.
- Ran the six existing suites the PR claims are unaffected: `test_fleet_rebase_inherited_drop.sh` 14/0, `test_fleet_rebase_stale_conflict_cleanup.sh` 7/0, `test_clone_freshness.sh` 16/0, `test_install_refresh.sh` 18/0, `test_dispatcher_github_gate.sh` 6/0, `test_fleet_claim_acquire.sh` 13/0 — all pass, no regressions.

**Logic traced:**
- `git()`'s subcommand scan (`fleet-net.sh:234-256`) correctly steps past `-C`/`-c` (2-token flags) and single-token `-*` flags to find the subcommand; only network verbs (`fetch|push|pull|clone|ls-remote|remote|fetch-pack|send-pack`) get the timeout prefix — a misparse can only ever *under*-guard (degrade to today's behavior), never misapply a timeout to a local op, since that requires the parsed token to literally be a network verb. Comment at `fleet-net.sh:234-237` correctly scopes the claim to fleet's own `git -C <wt> …` call sites rather than asserting it handles every possible git global-flag form.
- Sourcing-order requirement is honored in both consumers that matter: `fleet-dispatcher` sources `fleet-net.sh` (line ~99) before `fleet-clone-freshness.sh` (line ~111), so `advance_main_clone`'s fetch (which the dispatcher calls) resolves the shadow — confirmed by grep, not just by reading the comment. `fleet-claim` sources `fleet-net.sh` after `fleet-clone-freshness.sh`, but `fleet-claim` only calls `assert_clone_fresh` (rev-parse only, no fetch per its own doc comment) — its own direct `git fetch` call (`fleet-claim:3768`) and all `gh` calls occur well after the source line, so no unguarded window there either.
- `command "$tmo" "$FLEET_NET_TIMEOUT" git "$@"` — the inner `git` is a fresh `exec`, not a shell function call, so no recursion; confirmed by the passing T1/T3/T4 tests (fake `git`/`gh` binaries under the fake `timeout` correctly receive intact args).
- `FLEET_TIMEOUT_CMD="${FLEET_TIMEOUT_CMD-$(_fleet_net_resolve_timeout_cmd)}"` uses `-` (not `:-`) — correctly treats an explicitly-empty override (T6's `FLEET_TIMEOUT_CMD=""`) as "already resolved to no-guard" rather than re-probing, distinct from unset.
- `fleet-rebase`'s `escalate_if_hung_lock` only fires on the "another live holder owns the lock" branch (never on the dead-holder stale-break path, confirmed by T3), never auto-breaks an alive holder, and the missing-`started`-file case correctly skips the age check (T4) rather than guessing.
- `timeout-shim.py`'s SIGTERM-then-grace-then-SIGKILL matches its doc comment, and exit-code mapping (124/125/126/127/128+signal) mirrors coreutils.

### Praise
- The coreutils `--version` probe (rejecting Windows System32 `timeout.exe`, which is an interactive wait, not a command runner) is exactly the kind of portability trap that's invisible until it silently defeats the whole guard on a Windows fleet host — good catch, and the shim advertising `"coreutils"` in its own `--version` closes the loop cleanly.
- Choosing to shadow `git()`/`gh()` rather than edit ~49+ call sites individually is the right call for "guards future call sites by construction," and the PR's own audit (in `.fleet/plans/issue-2362.md`) enumerates every current unguarded site so the fix's coverage claim is checkable, not asserted.

- [ ] None outstanding — all claimed test-plan items verified above by rerunning them independently on this host.

Escalation: given the blast radius (this is the guard against future fleet-wide outages, running unattended inside three daemons, and the failure mode of a *wrong* guard — e.g. misapplying a timeout mid-rebase — is exactly the kind of thing the rebase lock exists to prevent) and the bash function-shadowing/exec-ordering subtlety involved, I'd like a second pass on the concurrency/ordering reasoning even though I traced it and it holds. Please Opus-recheck before merge.

Opus recheck required: fleet-wide safety infra (network-call guarding across `fleet-rebase`/`fleet-claim`/`fleet-dispatcher`) with bash function-shadowing + cross-script sourcing-order dependencies — high blast radius if the guard itself misfires, worth an independent trace even though my own verification (syntax checks, full new+existing test-suite reruns, manual ordering trace) found nothing wrong.

jakildev (commented):
## Opus recheck — fleet net-guard + hung-lock escalation (#2362)

**Verdict:** approve

Sonnet approved and escalated for an independent trace of the bash
function-shadowing / cross-script sourcing-order / concurrency reasoning,
given the blast radius (this guard runs unattended inside three daemons and a
*wrong* guard could misfire mid-rebase). I re-traced exactly that and it holds
— Sonnet's analysis is confirmed on every point, backed here by an empirical
exercise of the shadow under a fake coreutils `timeout`.

### Concurrency / ordering — independently confirmed

- **No recursion (Sonnet flagged, I confirm empirically).** `command "$tmo"
  "$FLEET_NET_TIMEOUT" git "$@"` — the inner `git` is an *argument* to
  `timeout`, which `execvp`s the real binary; it is never re-resolved to the
  `git()` shell function. Ran it under a fake `timeout` + fake `git`/`gh` on
  PATH: each network call prints `TIMEOUT-WRAP …` then `REAL-GIT/GH …`
  exactly once — no loop.
- **Local ops are structurally immune to the "misapplied timeout mid-rebase"
  hazard.** The `git()` case statement wraps *only* the eight network verbs;
  `rebase`/`checkout` fall to the `*)` passthrough and are never time-limited.
  Verified: `git -C /wt rebase …` and `git checkout -b …` pass through
  unwrapped, while `fetch`/`push`/`ls-remote` are wrapped. So the exact
  failure mode Sonnet worried about (a timeout corrupting a worktree
  mid-local-op) can't occur — it would require the parsed subcommand to
  literally be a network verb.
- **Global-flag skip is correct.** `-C <path>` / `-c <k=v>` consume two
  tokens; every other flag is self-contained. Verified `git -c user.name=x
  fetch` correctly steps past the 2-token flag and still detects `fetch` →
  wrapped. A misparse can only ever *under*-guard (degrade to today's
  unguarded behavior), never mis-wrap a local op.
- **Sourcing order holds in all three consumers** (traced the PR-head files,
  not just the hunks):
  - `fleet-claim`: `FLEET_LIB_DIR` set/exported at L115-116, `fleet-net.sh`
    sourced L151 — and there is **no shell `git`/`gh` network call before
    L151** (the only pre-source `gh`/`git` mentions in 1-151 are comments).
    Its own `git fetch` and ~49 `gh` sites are all downstream. No unguarded
    window.
  - `fleet-dispatcher`: `fleet-net.sh` sourced L99, **before**
    `fleet-clone-freshness.sh` (L111); `advance_main_clone` is only *called*
    in the main loop (L1976/1978). Since bash binds function names at call
    time, the shadow is installed long before the fetch runs. No net call
    precedes L99.
  - `fleet-rebase`: sourced at the top, before the lock logic and every
    fetch/push. Correct.
- **`TIMEOUT_CMD` reuse is well-ordered.** The dispatcher's
  `TIMEOUT_CMD="${FLEET_TIMEOUT_CMD:-}"` reads the var *after* the source
  point, so the coreutils-probed value is in scope — retiring the duplicate
  inline probe (and fixing its latent System32 `timeout.exe` bug) is a net
  simplification, not a regression.
- **Degrade paths are fail-open and safe.** Empty `FLEET_TIMEOUT_CMD` → both
  shadows passthrough (verified). Missing lib → source skipped, today's
  behavior. `FLEET_TIMEOUT_CMD="${FLEET_TIMEOUT_CMD-…}"` uses `-` not `:-`, so
  an explicit empty override is respected as "already resolved, no guard"
  rather than re-probed.
- **Hung-lock escalation is conservative.** It never auto-breaks an *alive*
  holder (a mid-rebase kill is the corruption the lock exists to prevent),
  skips the age check on a missing/garbage `started` stamp (pre-fix holder
  mid-rollout), and only escalates past the generous 1800 s ceiling.
- **Exit-124 semantics.** A timed-out call returns 124, which the existing
  `|| true` / `rc=$?` call sites already treat as an ordinary failure — the
  intended skip-and-continue degrade, converting a permanent hang into a
  recoverable failure. Strictly better than the status quo.

Sonnet ran both new suites (18/18, 13/13) and the six existing suites clean on
this host; I re-verified `bash -n` on the lib and exercised the shadow gating
matrix directly (network-verb wrap, local-verb passthrough, flag-skip,
no-recursion, empty-timeout degrade) rather than re-running the same suites.
No new concerns surface on the Opus pass. Clean to merge.

Issue #2376, opened 2026-07-14
voxel: GRID entity without C_RotationMode silently renders its authored rotation as identity


jakildev:
REBUILD_GRID_VOXELS's archetype requires C_RotationMode, so a GRID-mode entity created with a rotated C_LocalTransform but WITHOUT the component never re-rasterizes — it silently renders at identity. Nothing warns; the authored quaternion just doesn't show.

This bit detached_probe (fixed in the #2349 PR): its seeded GRID parity twin spawned without C_RotationMode, so the probe compared a rotated solid against an unrotated reference and mis-measured the revox anchor error.

Options, in rough preference order:
1. Make GRID the implicit default: have the rebuild query treat "no C_RotationMode" as GRID (a second query arm or bundling the component in the voxel-set creation path), so an authored rotation always renders.
2. A loud diagnostic: assert/log once when a C_VoxelSetNew entity carries a non-identity C_LocalTransform rotation but no C_RotationMode.
3. At minimum document the requirement in voxel/CLAUDE.md next to the GRID-mode conventions.

[sonnet]

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-2376.md`), and it is not opted out (`human:no-plan` / `fleet:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

jakildev:
## Plan: voxel: GRID entity without C_RotationMode silently renders its authored rotation as identity

- **Issue:** #2376
- **Model:** opus — the mechanics below are fully specified, but the change flips render behavior engine-wide and the acceptance harness (parity-probe extension) needs judgment; overrides the body's `[sonnet]` hint
- **Date:** 2026-07-20

### Verified current state (repro confirmed)

- `system_rebuild_grid_voxels.hpp` `create()` registers `registerSystem<REBUILD_GRID_VOXELS, C_VoxelSetNew, C_WorldTransform, C_RotationMode>` — an entity without `C_RotationMode` never matches the archetype, so only the translate-only baseline from `UPDATE_VOXEL_SET_CHILDREN` (query: `C_VoxelSetNew, C_WorldTransform`, no mode requirement) writes its world cells. Authored rotation/scale silently drop.
[...]

PR #2348, opened 2026-07-09
render: detached canvases raster in model space — camera-offset leak, camera-space cull, composite anchor; detached_probe harness

jakildev:
## Symptoms

A creation with a world-placed detached-canvas entity (the walking-unit
pattern: `C_EntityCanvas` + `RotationMode::DETACHED`, camera free to pan and
zoom) shows:

1. **Pan desync** — the entity drifts away from world geometry by roughly the
   camera's integer iso offset in game px (~80 px at a 10-iso-px pan, zoom 4).
2. **Content clipped out of its own canvas** — panning walks the model off the
   128×128 canvas edge (`isInsideCanvas` silently drops the writes).
3. **Disappear-on-pan** — a re-voxelize detached solid vanishes entirely under
   camera pan (the `canvas_stress` framing comment has long noted "detached
   canvases cull on a pan").
4. **Zoom-scaling anchor offset** — even at camera (0,0), detached content
   renders ~1 canvas texel (≈2·zoom px) away from its GRID twin.

None of this shows at camera (0,0) — which is exactly where every blessed
reference capture sits, so it survived the render-verify gates.

## Root causes

All three motion bugs are one confusion: **the detached raster path treats
model-space content as world-space.**

- `buildVoxelFrameData` sets `cameraTrixelOffset_ = getEffectiveCameraIso()`
  for every canvas; the detached branches never zero it, so the emit bakes
  `floor(cameraIso)` into detached canvas content. The composite then places
  the canvas at the entity's camera-relative screen position → the pan applies
  twice, and the content walks off the canvas (symptoms 1 + 2). The
  "screen-locked (no camera-pan-offset) path" the comments promised existed
  only in the comments.
- The chunk-visibility mask and the per-voxel GPU cull viewport
  (`cullIsoMin_/Max_`, `visibleIsoBounds_`) are camera-space, but a detached
  pool's positions are model-space — at camera (0,0) the two coincidentally
  align; any pan culls the solid out of its own canvas (symptom 3). The
  `revoxInverse` branch already stated the correct contract ("camera pan/yaw
  zeroed — mark every chunk visible") but only applied it to the rotating
  re-voxelize case.
- The raster anchors the model origin at `trixelOriginOffsetZ1(canvasSize)` =
  canvas center + (−1,−1) texels; the composite centers the quad on the
  entity's iso position, so content sits one texel off its GRID twin, scaling
  with zoom on screen (symptom 4).

## Fix

[...]

PR #2377, opened 2026-07-14
render: anchor the detached re-voxelize map to the solid's half-cell offset

### LINKED ISSUE #2349: render: detached re-voxelize rebuild places rotated solids ~1.9 iso px off the GRID twin (rotation-anchor offset)
**Area:** engine/render — detached re-voxelize rebuild (SYSTEM_REBUILD_DETACHED_VOXELS / c_revoxelize_detached)
**Model:** opus
**Blocked by:** (none — follow-on to the detached_probe PR)

## Symptom

A DETACHED_REVOXELIZE solid at a non-identity rotation composites ~1.9 iso px
away from where a GRID twin at the SAME rotation and translation renders —
error is linear in zoom (measured +3.85·zoom game px in x on macOS/Metal,
plus a smaller +y drift), so it reads as a visible placement offset at high
zoom. Plain DETACHED (forward-scatter) at identity tracks its GRID twin
within ~1.3·zoom px (band-centroid quantization), so this is specific to the
re-voxelize rebuild path.

## Repro (deterministic)

```
fleet-run IRDetachedProbe --auto-screenshot 60 --probe-assert
```

Read the `[detached-probe-parity] totem=revox` lines: measured-vs-expected dx
is ~+3.85·zoom game px at every shot (zoom 1/2/4, integer + fractional +
panned camera — camera-independent). The probe compares the re-voxelize totem
against a GRID anchor twin at the SAME 45°-Z seed pose, so rotated-silhouette
centroid geometry cancels; the residual is real placement, not measurement.

## Suspected mechanism

The rebuild bakes the entity rotation into pool cell positions rotating about
a center that is offset from the model origin (pool-grid center vs
pool-centered model origin, or a half-cell rounding of the pivot) — a
constant model-space displacement that scales with zoom on screen.

## Acceptance

- detached_probe revox parity tolerance tightens from `1.5 + 4.7*zoom` to the
  plain-DETACHED `1.5 + 1.8*zoom` with all shots green on both backends.
- canvas_stress re-voxelize references: expected small uniform shift of
  re-voxelize solids (re-bless), no shape change.

## Context

Found while building the detached_probe placement-parity harness (the
camera-offset-leak / camera-space-cull fix PR). The harness tolerance
documents this residual with a pointer to this issue; tighten it in the same
PR that fixes the rebuild.

jakildev:
## Summary

Fixes #2349 — a rotating `DETACHED_REVOXELIZE` solid rendered displaced from
its GRID twin by a constant model-space offset (linear in zoom on screen).
Two independent halves, one engine and one harness:

**Engine: the inverse resample ignored the solid's half-cell anchor.** A
`centerAroundOrigin` set with EVEN dimensions authors half-integer composed
locals (offset `-(size-1)/2`); the source occupancy grid keys them by
`roundHalfUp(composed)`, which lands the lattice `+0.5` off per even axis.
Mode 0 (identity) passes the composed locals through **unrounded**, but mode
1 (rotating) mapped raw lattice cells — so the solid jumped by the anchor
the moment it rotated. The fix threads the per-axis anchor
(`composed - roundHalfUp(composed)`, uniform across the pool — asserted at
seed) through `RevoxelizeDetachedParams` and maps between anchored POINTS
everywhere: dest→src is now `roundHalfUp(R⁻¹·(c + a) - a)` (GLSL + Metal
twins + the coverage probe), the authored raster position is `c + a`
(matching mode 0's convention at identity), and the CPU exposed-face-mask
twin (`REBUILD_DETACHED_VOXELS`) derives the same anchored dest cells. The
dest cube gains `+1` center headroom for the anchored round's worst case.

**Harness: the probe's parity anchor never rotated.** `spawnGridTotem`
omitted `C_RotationMode`, which `REBUILD_GRID_VOXELS`'s archetype requires —
the 45°-seeded GRID twin silently rasterized at identity, so the parity
measured a rotated solid against an unrotated reference (part of the
original #2349 error magnitude was this probe bug). Fixed by attaching
`C_RotationMode{GRID}`; the silent-skip trap itself is filed as #2376.

**Tolerances tighten from `1.5 + 4.7*zoom` to `1.5 + 1.8*zoom`** (the same
bound the plain-DETACHED pair already met). One documented exception: the
revox pair's **y** axis keeps `1.5 + 2.8*zoom`, because GRID forward-rounds
the totem's half-integer z coordinates up half a cell onto its integer
lattice while the anchored revox fill places content exactly — a bounded
cross-MODE quantization difference (measured ~2.4*zoom px), not a revox
placement error.

## Validation

- `IRDetachedProbe --probe-assert`: **24/24 PASS** (6 shots × orientation +
  clip + 2 parity checks). Red baseline: pre-fix revox x-parity error was
  ~4*zoom px (and ~6*zoom at a 90° seed), failing the tightened gate at
  every zoom; post-fix x error is 0.2–3.5 px against a 672 px expected
  displacement at zoom 4.
- `render-verify --target IRCanvasStress`: the spinning revox solids
  legitimately moved sub-cell against the pre-fix references (thin edge
  deltas on every rotating solid, ~99.3-99.5% match) — **macOS references
  re-blessed** in this PR. The `shadow_overlay_floor` component-count FAIL
  reproduces identically on the base-branch build (pre-existing, unrelated,
  metric-based not reference-based — untouched).

## Screenshots

`probe_z4_cam00` — left→right: revox (45° seed), plain detached (identity),
GRID identity anchor, GRID 45° twin. Before: the GRID 45° twin renders
unrotated (probe bug) and the revox solid sits off its expected cells.
After: twin rotated, revox aligned.

| Before (#2348 base) | After |
|--------|-------|
|  |  |

## Notes

- The anchor derivation has ONE home — `GridRotation::halfCellAnchor` (+
  `anchoredCellForDetachedVoxel` for the mask twin), called by the seed and
  the CPU mask twin; the kernels receive the value via the params UBO. The
  mask twin derives locally (not from the seeded buffer field) because its
  UPDATE tick can precede the render-side seed on a pool's first frame —
  commented at the site.
- The anchored map does NOT grow the dest cube: anchored axes shift the
  slot→cell decode window +1 (`revoxDestDecodeShift`, both kernels) instead
  of a symmetric `center+1` — which would have grown the per-frame
  dispatch/active-mask-clear/compact domain 11-46% depending on solid size.

## Test plan

- [x] `IRDetachedProbe --probe-assert` all 24 verdicts PASS (macOS/Metal)
- [x] `scripts/render-verify.py --target IRCanvasStress` — revox placement
  delta re-blessed; all other content byte-identical; pre-existing shadow
  metric FAIL reproduced on base branch
- [ ] Linux/OpenGL parity run (GLSL twin changed identically; probe gates it)

Closes #2349

and model-space raster fixes live there).


jakildev:
Merger: base PR #2348 merged; re-targeted this PR
will resolve semantically.

Conflicted files:
- `engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp` — master: `e6c074bc render: detached canvases raster in model space — fix camera-offset leak, camera-space cull, composite anchor; add detached_probe harness (#2348)`; PR: `b1aef1c6 render: anchor the detached re-voxelize map to the solid's half-cell offset`

— fleet merger

jakildev:
Resolved: the conflict was purely an inherited-prefix artifact, not a semantic one. The branch carried a stale pre-merge copy of parent #2348 (`c18affda`) that collided with #2348's merged form now in master (`e6c074bc`). Dropped the inherited prefix (`git rebase --onto origin/master c18affda`), replaying only this PR's genuine commit onto current master (`b1aef1c6` → `20227b37`). The +3 lines to `system_voxel_to_trixel.hpp` (setting `params.anchor_` from `buffer.anchor_`) applied cleanly into master's rewritten STAGE_1 function; no by-hand reconciliation needed. Build clean (IRShapeDebug, macOS/Metal). Diff is now a single commit on master — reviewer please re-evaluate the rebased diff. — worker

jakildev (commented):
**Opus recheck — APPROVE** ✅

No prior Sonnet review sits on this PR (it went author → merger `fleet:semantic-conflict` → worker rebase `--onto origin/master c18affda`, now a single genuine commit `20227b3` on master). Since it lands squarely in core render-invariant territory — CPU/GPU UBO struct, both shader backends, the detached re-voxelize inverse-resample handshake — I took the full Opus pass rather than wait for a first pass. Focused on what a checklist can't confirm: struct parity, the decode-window off-by-one, and the CPU↔GPU rounding handshake.

**What I verified (the #2349 half-cell anchor):**

1. **CPU/GPU struct parity is exact.** `RevoxelizeDetachedParams` grows 64→80 B by appending one `vec4 anchor_` at offset 64. All five members are `vec4`/`ivec4` (16-B aligned), so the std140 block stays straddle-free; the GLSL `RevoxelizeParams` block (binding 16) and the Metal `RevoxelizeParams` struct both append the matching field at the same offset, and the `static_assert(sizeof == 80)` is updated in lockstep. No silent reorder/resize risk.

2. **The decode-window shift is mathematically sound — no holes, no overflow.** On an anchored (−0.5) axis, a dest cell is `c = roundHalfUp(R·s_p − anchor)`; with `|R·s_p_axis| ≤ length(s_p) ≤ maxRadius ≤ center`, `c` lands in `[−center+1, center+1]`, exactly the shifted decode window `d − center + 1` (d ∈ [0, 2·center]). On odd (0) axes it stays `[−center, center]`. `revoxDestDecodeShift = ivec3(anchor < −0.25)` (GLSL) / `select(0,1, anc < −0.25)` (Metal) reproduces this per-axis. Because `anchor ∈ {−0.5, 0}` strictly, the extreme `roundHalfUp(center + 0.5) = center + 1` cannot escape by more than one cell. Slot count stays `side³`, so the dispatch descriptor, clear, and compact are untouched — the "zero dispatch growth" claim holds.

3. **CPU↔GPU rounding stays byte-identical.** `GridRotation::halfCellAnchor` / `anchoredCellForDetachedVoxel` are the single derivation home; `seedResidentLocals` seeds `buffer.anchor_` from voxel 0 and asserts pool-wide uniformity (integer locals + one shared center-around-origin offset ⇒ per-axis `{−0.5, 0}`); `REBUILD_DETACHED_VOXELS` derives the same anchor from voxel 0 via the same function — documented as necessary because that UPDATE tick runs before the render-side seed on the pool's first frame. Both sides apply `− anchor` before `roundHalfUp` and rotate the anchored *points* (`cell + anchor`), not raw lattice cells, so the mask twin and the GPU raster gate on identical cells.

4. **Both backends moved symmetrically.** `globalPositions[slot]` is anchored (`cell + anchor`) on GLSL and Metal; `revoxSourceCellForDest` and the six adjacency probes take the anchor consistently; the neighbor probes are pure math (not slot lookups), so window-boundary adjacency is fine.

**Probe harness (`detached_probe/main.cpp`):** the `C_RotationMode{RotationMode::GRID}` addition to `spawnGridTotem` is a genuine fix — the GRID twin was silently rasterizing at identity, so the parity compare was rotated-vs-unrotated (half of #2349's mis-measurement). The X/Y tolerance split with the documented y-only cross-MODE quantization term is reasonable calibration for a parity probe, not a merge concern.

**Non-blocking observation (no fix required):** `REBUILD_DETACHED_VOXELS` trusts voxel 0's anchor without its own uniformity assert (only the seed asserts it). Safe today — authoring uniformity is guaranteed by construction and the seed's `IR_ASSERT` fires on the same pool — but if `IR_ASSERT` is a release no-op and a future non-uniform authoring path ever appears, the rebuild mask would silently use voxel 0's anchor for all voxels. Fine to leave as-is given the invariant.

Merge gate: authored-on-macos covers the Metal tier; tagging `fleet:needs-windows-smoke` for the OpenGL tier (GLSL changed).

🤖 Opus recheck by Claude Opus 4.8 (1M context)

PR #2378, opened 2026-07-14
fleet: fleet-up restores both main clones to an up-to-date master

jakildev:
## Summary
- `fleet-up` now restores **both** main clones (engine + game) to an up-to-date `master` before bringing the fleet online, closing the silent whole-fleet stall hit twice on 2026-07-13: a clone parked off-master freezes its local `master` ref, and the #1810 freshness guard then refuses every claim while workers fast-exit with nothing visible to the human.
- New `restore_main_clone_to_master` in `fleet-clone-freshness.sh`, called by `fleet-up` in place of the weaker `advance_main_clone` (which stays as-is for the unattended dispatcher tick — it must never switch branches):
  - tracked modifications (staged or unstaged) anywhere → never touch, loud warning ("live WIP wins")
  - parked branch / detached HEAD with no tracked WIP → `checkout master`; untracked junk files (0-byte `=`, `.review-body.md`) no longer wedge it
  - then ff-advance to `origin/master`; the ff tail is factored into `_ff_advance_to_origin_master`, shared with `advance_main_clone`, so the two variants can't drift
- Tests: `test_clone_freshness.sh` extended with T9–T13 (parked-clean restore + ff, tracked-WIP untouched, untracked-junk restore, detached HEAD, missing-repo no-op) — 26/26 pass.

## Test plan
- [x] `scripts/fleet/tests/test_clone_freshness.sh` — 26/26 (hermetic throwaway git repos)
- [x] `bash -n` on `fleet-up` and `fleet-clone-freshness.sh`

### DIFF
--- a/scripts/fleet/fleet-clone-freshness.sh
+++ b/scripts/fleet/fleet-clone-freshness.sh
@@ -195,11 +195,19 @@ restore_main_clone_to_master() {
 
     local branch
     branch="$(git -C "$root" rev-parse --abbrev-ref HEAD 2>/dev/null || echo '?')"
+
+    # Tracked WIP wins over the restore on EITHER path (parked branch or already
+    # on master). This must gate before the ff-advance below: _ff_advance's
+    # ff-only refuses only when the incoming commits *overlap* the dirty files —
+    # a disjoint dirty tree on master would otherwise be advanced silently
+    # underneath uncommitted work, contradicting the "never touch, warn loudly"
+    # invariant documented above.
+    if [[ -n "$tracked_dirty" ]]; then
+        echo "fleet-clone-freshness: $root has tracked modifications on '$branch' — leaving it alone (live WIP wins). Claims stay blocked until it is clean and on master; commit or ship the WIP, then rerun fleet-up." >&2
+        return 0
+    fi
+
     if [[ "$branch" != "master" ]]; then
-        if [[ -n "$tracked_dirty" ]]; then
-            echo "fleet-clone-freshness: $root is parked on '$branch' WITH tracked modifications — leaving it alone (live WIP wins). Claims stay blocked until it returns to master; commit or ship the WIP, then rerun fleet-up." >&2
-            return 0
-        fi
         if git -C "$root" checkout master --quiet 2>/dev/null; then
             echo "fleet-clone-freshness: $root returned to master (was on '$branch', no tracked WIP)." >&2
         else
--- a/scripts/fleet/fleet-clone-freshness.sh
+++ b/scripts/fleet/fleet-clone-freshness.sh
@@ -37,6 +37,13 @@
 #   advance_main_clone  <repo_root>  — guarded, rate-limited fetch + ff-only
 #                                     advance of a clean on-master clone. Never
 #                                     switches branches, never resets --hard.
+#   restore_main_clone_to_master <repo_root>
+#                                   — fleet-up-time stronger variant: returns a
+#                                     clone parked off-master (PR branch from a
+#                                     cursor session, reviewer scratch leftover,
+#                                     detached HEAD) to master when no tracked
+#                                     WIP is at risk, then ff-advances. Never
+#                                     touches a tree with tracked modifications.
 #
 # Env:
 #   FLEET_SKIP_CLONE_FRESHNESS=1  — disable the assert_clone_fresh claim gate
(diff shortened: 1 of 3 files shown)


jakildev (commented):
## Review — fleet: fleet-up restores both main clones to an up-to-date master

**Verdict:** needs-fix

### Needs-fix
- `scripts/fleet/fleet-clone-freshness.sh:189-213` (`restore_main_clone_to_master`) — the tracked-WIP guard only fires inside the `branch != "master"` block (lines 198-202). If the main clone is **already on `master`** with tracked (staged or unstaged) modifications, `tracked_dirty` is computed at line 194 but never checked — execution falls straight through to the unconditional `git fetch` + `_ff_advance_to_origin_master` at lines 211-212. `_ff_advance_to_origin_master`'s `git merge --ff-only` (line 163) only refuses when the incoming commits *overlap* the dirty files (see its own comment at lines 161-162: "a disjoint dirty tree advances fine") — so a disjoint dirty tree on `master` gets its branch tip silently advanced underneath uncommitted work. This contradicts the function's own documented invariant at line 180: "tracked modifications anywhere (staged or not) → never touch, warn loudly." It's also the one on-master+dirty case the new T9-T13 tests don't cover (T6 covers this scenario for `advance_main_clone`, but there's no equivalent for `restore_main_clone_to_master`).

  Suggested fix: hoist the `tracked_dirty` check above the `branch != "master"` conditional so it gates both paths uniformly, e.g.:
  ```bash
  if [[ -n "$tracked_dirty" ]]; then
      echo "fleet-clone-freshness: $root has tracked modifications — leaving it alone (live WIP wins)..." >&2
      return 0
  fi
  if [[ "$branch" != "master" ]]; then
      ...
  fi
  ```
  Add a T14 mirroring T6 (on `master`, disjoint dirty tracked file, assert master is NOT advanced) to close the coverage gap.

### Nits
- `scripts/fleet/fleet-clone-freshness.sh:211` — the new `git -C "$root" fetch origin master --quiet` (and the pre-existing top-level fetches in `fleet-up`) aren't timeout-guarded via `fleet-net.sh`, unlike `fleet-dispatcher`/`fleet-rebase`/`fleet-claim` post-#2362. `fleet-net.sh`'s own docstring scopes itself to "unattended daemon" consumers and deliberately excludes `fleet-up` (a human-attended startup script), so this isn't a policy violation — but a hung fetch here reproduces the exact "fresh dispatcher hung at startup on its clone-freshness fetch" failure mode #2362 just fixed, just against `fleet-up` instead of the dispatcher. Worth a follow-up if `fleet-up` hangs are ever observed in practice; not blocking this PR.

### Praise
- The `_ff_advance_to_origin_master` extraction is clean — `advance_main_clone` keeps its own stricter Guard 2 (any dirty tree, tracked or not, skips unconditionally) ahead of the shared tail, so the refactor doesn't weaken the existing unattended-tick path at all. Good factoring.
- Real incident-driven fix with a concrete, dated trigger (2026-07-13 double-stall) and a test suite that actually hits the target scenarios (parked branch, tracked WIP, untracked junk, detached HEAD, missing repo).

- [ ] Add a test for "on `master`, disjoint tracked dirty file" and confirm it currently fails (reproducing the bug), then fix + re-run `test_clone_freshness.sh`.

Opus recheck not required.

jakildev:
Addressed fleet:needs-fix:

- **Needs-fix (the real bug)** — `restore_main_clone_to_master`: hoisted the `tracked_dirty` guard above the `branch != "master"` conditional so it gates *both* paths. An on-master clone with a disjoint tracked-dirty tree no longer falls through to `git fetch` + `_ff_advance_to_origin_master` (whose ff-only only refuses on *overlapping* dirty files) and get its tip silently advanced under uncommitted work. The unified message reports the branch context (`on '$branch'`) and the invariant at the docstring now actually holds in code.
- **Test (T14)** — added, mirroring T6 for `advance_main_clone`: on `master`, disjoint tracked dirty file, assert master is NOT advanced + WIP preserved + warns loudly. Verified per your test plan — T14 *fails* on the pre-fix function (`advanced master under uncommitted WIP` / no warning) and passes after the hoist. `test_clone_freshness.sh` is now **29/29**.
- **Nit (fetch timeout-guard)** — left as-is per your own note: `fleet-net.sh` deliberately scopes itself to unattended daemons and excludes the human-attended `fleet-up`, so this isn't a policy violation and you flagged it non-blocking. Worth a follow-up only if a `fleet-up` startup hang is ever observed in practice.

`bash -n` clean on both files. Reviewer please re-verify the rebased diff. — worker

jakildev (commented):

### Confirming Sonnet's pass

### Nits

### Praise

### Before merge

PR #2380, opened 2026-07-14
fleet: extract shared bash test assertion helpers into tests/lib_assert.sh

jakildev:
## Summary
- New `scripts/fleet/tests/lib_assert.sh`: shared PASS/FAIL counters, `ok`/`bad`, `assert_eq`, `assert_contains`, `assert_absent`, and a `summarize` exit idiom for the bash test suite. Canonicalized on the fixed-string `grep -qF` matcher with indented-haystack diagnostics (the richest of the previously drifting copies).
- Converted the 7 tests that carried local copies to `source "$(dirname "$0")/lib_assert.sh"`: `test_fleet_rebase_inherited_drop.sh`, `test_fleet_rebase_stale_conflict_cleanup.sh`, `test_dispatcher_dry_run_mode.sh`, `test_session_track.sh`, `test_babysit_launch.sh`, `test_fleet_claim_safety_guards.sh`, `test_fleet_gh_token.sh`.
- `assert_not_contains` call sites → `assert_absent` (same semantics); gh_token's path-existence `assert_absent` renamed `assert_no_path` to free the name for the haystack form. File-specific asserts (`assert_exit`, `assert_dir`, `assert_no_dir`, `assert_no_path`) stay local, rebuilt on `ok`/`bad`.
- `scripts/fleet/CLAUDE.md` authoring rules now point new bash tests at the shared lib.

Net −103 lines. `test_fleet_rebase_plan_automerge.sh` (named in the task brief) does not exist in the tree; the actual carrier set was these 7.

## Test plan
- [x] All 7 converted tests pass (`PASS`/`FAIL` tallies: 14, 7, 13, 29, 8, 20, 23 — each identical to its pre-conversion run at HEAD, verified by re-running the originals side by side)
- [x] Failure path smoke-checked: `summarize` exits 1 and diagnostics render for `assert_contains`/`assert_absent`/`assert_eq` misses
- [x] `bash -n` clean on the lib and all 7 tests

## Notes for reviewer
~40 other tests under `scripts/fleet/tests/` still define local `assert_eq`/`ok`/`bad` copies — deliberately out of scope (this PR covers the `assert_contains`/`assert_absent` carriers); a follow-up task is queued for the rollout.

### DIFF
--- a/scripts/fleet/CLAUDE.md
+++ b/scripts/fleet/CLAUDE.md
@@ -30,6 +30,11 @@ applies here too — see `docs/agents/CLAUDE-BASELINE.md` §Style.
 - **A new executable ships with a `tests/test_<name>.{sh,py}`** in the same
   PR — the fleet-tooling form of the review checklist's "new feature with no
   new test"; the `simplify` pre-commit pass flags the omission.
+- **Bash tests source `tests/lib_assert.sh`** for the PASS/FAIL counters,
+  `ok`/`bad`, `assert_eq`/`assert_contains`/`assert_absent`, and the
+  `summarize` exit idiom — don't re-copy the helpers into a new test.
+  Genuinely test-specific asserts (path existence, exit codes) stay local,
+  built on `ok`/`bad`.
 - **Unattended daemons timeout-guard their network calls.** The host's
   connections to GitHub intermittently black-hole (silent TCP death), so a
   hung `git fetch` / `gh …` in a fleet daemon (dispatcher loop, `fleet-rebase`,
--- a/scripts/fleet/tests/lib_assert.sh
+++ b/scripts/fleet/tests/lib_assert.sh
@@ -1,0 +1,66 @@
+# Shared assertion helpers for the bash tests in this directory. Sourced,
+# never executed — the name deliberately avoids the test_*.sh pattern so
+# anything that globs for tests skips it:
+#
+#     source "$(dirname "$0")/lib_assert.sh"
+#
+# Provides the PASS/FAIL counters, ok/bad bumpers, the assert_* family, and
+# summarize. End every test with summarize (optionally with a suite label)
+# as the last command so the script's exit status reflects the failure
+# count:
+#
+#     summarize "fleet-foo tests"
+#
+# assert_contains / assert_absent match the needle as a fixed string
+# (grep -F), one line at a time — a needle spanning a newline never
+# matches. Tests that need path-existence or exit-code assertions define
+# those locally (see test_fleet_claim_safety_guards.sh).
+
+PASS=0
+FAIL=0
+
+ok()  { PASS=$((PASS + 1)); echo "  ok: $1"; }
+bad() { FAIL=$((FAIL + 1)); echo "  FAIL: $1"; }
+
+assert_eq() {
+    local actual="$1" expected="$2" msg="$3"
+    if [[ "$actual" == "$expected" ]]; then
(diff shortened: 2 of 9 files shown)


jakildev (commented):
## Review — fleet: extract shared bash test assertion helpers into tests/lib_assert.sh

**Verdict:** approve

Standalone PR (`baseRefName == master`). `mergeable` is `MERGEABLE`.

### Verification performed
- **Ran all 7 converted tests** (couldn't `gh pr checkout` — the branch is already checked out in another local worktree, so I ran them there read-only instead): every PASS/FAIL tally matches the PR body's claimed counts exactly — `test_babysit_launch.sh` 8/0, `test_dispatcher_dry_run_mode.sh` 13/0, `test_fleet_claim_safety_guards.sh` 20/0, `test_fleet_gh_token.sh` 23/0, `test_fleet_rebase_inherited_drop.sh` 14/0, `test_fleet_rebase_stale_conflict_cleanup.sh` 7/0, `test_session_track.sh` 29/0.
- **`bash -n` syntax-checked** all 8 touched/added files at the PR head SHA — clean.
- **Traced the substring→`grep -qF` semantic change** the new lib's docstring flags ("a needle spanning a newline never matches"). Four of the seven converted tests (`test_babysit_launch.sh`, `test_dispatcher_dry_run_mode.sh`, `test_fleet_claim_safety_guards.sh`, `test_session_track.sh`) previously used bash `[[ "$haystack" == *"$needle"* ]]` substring matching, which — unlike line-based `grep` — matches across embedded newlines. I checked every `assert_contains`/`assert_absent` call site in those four files against where its haystack variable is actually produced:
  - `test_babysit_launch.sh`'s `$out` is piped through `grep '^claude '` before capture — single line by construction.
  - `test_dispatcher_dry_run_mode.sh`'s `$cmd_live` is one `--print-dispatch-command` line.
  - `test_fleet_claim_safety_guards.sh`'s `$err` needles (`"must precede the subcommand"`, `"reconcile: unknown arg"`) each come from a single `echo ... >&2` in `fleet-claim` (`fleet-claim:4091`, `:2396`) — no line-wrapping.
  - `test_session_track.sh`'s `$out` is `json.dumps(...)` with no `indent=` — compact single-line JSON, so the two-line role-file content lands in the string value with its newline escaped to a literal `\n`, not a real one.
  
  No call site is affected — the refactor is behavior-preserving here, matching the PR body's claim.
- **Confirmed the `assert_absent`→`assert_no_path` rename** in `test_fleet_gh_token.sh` left no stale references to the old name (`grep` clean).
- The two tests already on the canonical `grep -qF` form (`test_fleet_rebase_inherited_drop.sh`, `test_fleet_rebase_stale_conflict_cleanup.sh`) are a pure dedup — no semantic change possible.

### Praise
- The docstring at the top of `lib_assert.sh` proactively documents the one real semantic gap (`grep -qF`'s per-line matching vs. the substring form it replaces) instead of leaving it to be discovered — exactly the kind of thing that would otherwise silently break a future test.
- `scripts/fleet/CLAUDE.md` gets an authoring rule pointing new tests at the shared lib, so this doesn't just fix the existing 7 — it prevents the 8th copy.
- Honest scoping: the PR body flags that `test_fleet_rebase_plan_automerge.sh` named in the task brief doesn't exist in the tree, and that ~40 other tests with local copies are a deliberate follow-up rather than silently expanding scope.

- [x] All verification above already run by this review; nothing further needed.

Opus recheck not required.

Issue #2397, opened 2026-07-14
engine/math: IRMath::wrapToRange helper + migrate hand-rolled fmod+correction sites (split-out from #2354)


jakildev:
**Area:** engine/math (+ 4 call-site migrations + one `cpp-math.md` bullet)

Add an `IRMath` wrap-into-range helper and route the hand-rolled `fmod` + negative-correction idiom through it. Split out from coding-improvement ticket #2354 (triage 2026-07-14) because the work is code, not convention text — the doc rule can't mandate a helper that doesn't exist yet.

## Scope

1. **Helper** in `engine/math/include/irreden/ir_math.hpp`:
   - `IRMath::wrapToRange(x, range)` → `[0, range)` (right half-open, correct negative handling).
   - Thin angle wrappers `wrapAngleTwoPi(a)` → `[0, 2π)` and `wrapAnglePi(a)` → `[-π, π)`.
2. **Migrate the four known hand-rolled sites** (two are live `std::fmod` cpp-math violations):
   - `engine/prefabs/irreden/update/components/component_periodic_idle.hpp` (`valueAtAngle`) — `IRMath::fmod(angle, kTwoPi)` + `if (wrapped < 0) wrapped += kTwoPi`.
   - `engine/prefabs/irreden/render/camera.hpp:87-88` (`wrapYaw`) — `IRMath::fmod(yaw + kPi, kTwoPi)` + correction → `[-π, π)`.
   - `engine/prefabs/irreden/update/systems/system_action_animation.hpp:64-70` — `std::fmod(elapsed, period)` + correction (also a `std::fmod` violation).
   - `engine/prefabs/irreden/update/systems/system_spring_platform.hpp:138-142` — `std::fmod(elapsed, period)` + correction (also a `std::fmod` violation).
3. **Rule text**: one bullet in `.claude/rules/cpp-math.md` routing angle/range wrapping through the new helper, so the existing math-primitive simplify check can flag remaining hand-rolled `fmod`+correction sites as "use `IRMath::wrapToRange`".
4. Headless unit coverage for the helper (negative inputs, exact-boundary inputs, `[-π, π)` shift).

One correct implementation also keeps CPU↔GPU consistency reasoning in one place (per the `IRMath::roundHalfUp` rationale).

Refs: #2354 (originating coding-improvement ticket), PR #2352 (where the dedup surfaced).

jakildev:
[...]

Issue #2364, opened 2026-07-13
convention audit: other "caller must finalize after using system X" footguns (encapsulate, don't delegate)


jakildev:
**Area:** engine (ECS convention / cross-system audit)

**Context.** Deferred follow-up from epic #2164 (voxel derived-state encapsulation), which closed 2026-07-13. #2164 codified the convention *"System-owned invariants: encapsulate, don't delegate to callers"* (`.claude/rules/cpp-ecs.md` §System-owned invariants) after the same `recomputeFaceOccupancy`/`syncActiveMask` footgun landed four times — a subsystem dependency was pushed onto every caller as a manual, opt-in, easy-to-drop step. The fix encapsulated it in `C_VoxelSetNew::editVoxels`/`carve`.

The #2164 plan's out-of-scope note (user-deferred): *audit the rest of the engine for other "caller must finalize after using system X" instances* — places where a system relies on callers to run a post-use finalize/resync/recompute step that could instead be owned by the subsystem (an attached API or a pre/post-tick hook).

**What to do.** Sweep the engine for the pattern: a public mutator/producer whose correct use requires a *separate* caller-invoked finalize call (recompute / sync / rebuild / mark-dirty). For each, decide whether it should be encapsulated per the #2164 convention or is genuinely a caller concern; propose the encapsulation where warranted.

**Not queued** — this is the human-cued `fleet:coding-improvement` triage channel (see the `triage-coding-improvements` skill), per the #2164 plan's "file as fleet:coding-improvement later" direction. Filed by epic-steward at #2164 close-out.

jakildev:
Triage 2026-07-14: routed out of the coding-improvement channel — this is an engine-wide audit/code task, not a convention-surface edit. Label removed; stays open as a plain issue for human approval into the queue when wanted.

jakildev:
## Triage

**Verdict:** park
**Objective:** none
**Suggested model:** (n/a)
[...]

PR #2244, opened 2026-07-05
persist: IRPersist Lua bindings + IR_PERSIST_DUMP world dump (#2218)

jakildev:
## Summary
- Persist P7 / W-9 + W-11 (epic #667), the chain tail — exposes the ECS world snapshot to Lua as the `IRPersist` table and adds an env-gated human-readable dump.
- `IRWorld::makeDefaultSaveRegistry()` + `(path)`-only `saveWorld`/`loadWorld` overloads: the landed P6 API takes a `SaveRegistry&`, but Lua passes none, so the binding forwards to a process-default registry. It is a **curated subset** — only components with a working `SaveSerialize` today (`C_VoxelSetNew` + trivially-copyable PODs); the full inventory doesn't compile (heap-owning opted-in components lack serializers). Full-coverage is deferred to #2242. Built fresh per call so `ComponentId`s match the live `EntityManager`.
- `IRPersist.saveWorld/loadWorld` Lua binding: include-only glue (no link edge to World, which links Scripting), same frame-boundary contract as `IRWorld.resetGameplay` (loadWorld does not reset first — call `IRWorld.resetGameplay()` before it).
- W-11 `.json.txt` dump: an `IR_PERSIST_DUMP`-gated second writer over the save walk (archetype members + `CHILD_OF` edges) — pure side-output, so the binary is byte-identical flag-on/off. Factored `collectChildParentEdges` out of `makeRelationChunk` (+ `sortedLocalIndices` / `writeComponentsArray` helpers) so the dump and binary share one source of truth.
- Added `IRUtility::envFlagSet` as a new public wrapper over the existing private `nonEmptyEnv`.

## Test plan
- [x] `test/script/lua_world_snapshot_test.cpp` (new): surface bound; full round-trip through the Lua binding (4 archetypes + `CHILD_OF`, exact id/component/relation parity); missing-path → false-no-throw; non-string arg → Lua error; dump-mode content + flag-off byte parity.
- [x] Existing persist/snapshot suites green (incl. `DoubleSaveIsByteIdentical`) — the `localIndexList` extraction preserves exact binary output.
- [x] Builds clean on `macos-debug` (Metal). Backend-agnostic (no render/GL/Metal code) — needs Linux/GL smoke.

## Notes for reviewer
The design decision (curated-subset default registry vs full inventory) resolves the sonnet→opus escalation on #2218: option 1 (MVP + deferred #2242), forward-compatible with the eventual `HasExplicitSaveSerialize<C>` filter.

[...]

PR #2399, opened 2026-07-14
docs/fleet: absorb coding-improvement batch — 15 rules across 12 surfaces

### LINKED ISSUE #2261: engine/world: default-registry additions must extend the IRPersist Lua round-trip test (coding-improvement)
**Class:** A (missing rule)
**Target artifact:** `engine/world/CLAUDE.md` § "Process-default registry + `(path)` overloads (persist P7, #2218, epic #667)"
**Proposed change:** Add one bullet: "Adding a component to `makeDefaultSaveRegistry()` must also extend `test/script/lua_world_snapshot_test.cpp` with a round-trip case through the actual `IRPersist` Lua surface — a component landing in the registry with only a standalone `SaveSerialize<C>` unit test (e.g. `voxel_set_serialize_test.cpp`) leaves the wiring itself (registry entry → Lua binding → reload) unverified."

## Context
Surfaced addressing an Opus-recheck nit on PR #2244 (`persist: IRPersist Lua bindings + IR_PERSIST_DUMP world dump`, #2218): `world_default_registry.cpp` registers `C_VoxelSetNew` in the curated default registry, but `lua_world_snapshot_test.cpp` only round-tripped the three trivially-copyable POD components — `C_VoxelSetNew`'s serializer itself was covered standalone by #2240's `voxel_set_serialize_test.cpp`, but nothing exercised it through the actual `IRPersist.saveWorld`/`loadWorld` entry point Lua callers use. The doc's existing sentence ("Add each component to the curated list as its serializer lands") documents the registry-membership step but not the matching test-coverage step, so an author following it exactly still ships a coverage gap a reviewer has to catch.

## Why it generalizes
The curated registry is explicitly designed to grow ("Add each component to the curated list as its serializer lands") — every future component added here repeats the same two-step shape (register in `world_default_registry.cpp`, add a headless serializer unit test), and the doc only states the first half. Stating the second half where the registration step is already documented closes the gap at the surface an author actually reads when adding a component, rather than relying on a reviewer to notice the missing Lua-surface case each time.

## Occurrences
- PR #2244, `test/script/lua_world_snapshot_test.cpp` — added `RoundTripsVoxelSetNew` after an Opus-recheck nit flagged the gap.

### LINKED ISSUE #2276: render/simplify: GPU-shader dispatch-dim writes must cap numGroupsX + spill to Y (coding-improvement)
**Class:** B (exists but didn't fire — enforcement gap for the GPU-shader-authored case)
**Target artifact:** `.claude/skills/simplify/SKILL.md` §5 (render pipeline) — extend the dispatch-grid check to cover shader-side dispatch-dim *writes*; backstop in `.claude/skills/review-pr/SKILL.md` render checklist.
**Proposed change:** In the simplify §5 render check, alongside "hand-rolled compute dispatch sizes should use `voxelDispatchGridForCount()`", add: a **GPU shader** (`.glsl`/`.metal`) that derives an indirect-dispatch grid from a runtime count must cap `numGroupsX` at `kMaxDispatchGroupsX` (1024) and spill the remainder into `numGroupsY` — mirroring `writeDispatchDims()` in `c_voxel_visibility_compact.glsl` — never write an uncapped 1-D `numGroupsX = divCeil(count, tile)` with `numGroupsY = 1`. Consumers must then recover the flat group index as `groupId.x + groupId.y * numGroupsX`.

## Context
Surfaced fixing `fleet:needs-fix` on PR #2273. The reviewer flagged that `c_per_axis_cell_finalize.glsl:39-41` (mirrored in `metal/c_per_axis_cell_finalize.metal`) wrote an **uncapped 1-D** dispatch grid (`numGroupsX = divCeil(count, 256)`, `numGroupsY = numGroupsZ = 1`). At a large Lua-configured voxel-pool edge (>256×65535 ≈ 16.7M occupied cells on one axis) `numGroupsX` exceeds the OpenGL-guaranteed `GL_MAX_COMPUTE_WORK_GROUP_COUNT[0]` minimum of 65535 → undefined dispatch, silently dropped/corrupted cells, no error. The codebase already had two established sites for the capped 2-D fold — the CPU `voxelDispatchGridForCount()` (`engine/prefabs/irreden/render/voxel_dispatch_grid.hpp`) and the GPU-authored `writeDispatchDims()` (`c_voxel_visibility_compact.glsl`) — one file over.

## Why it generalizes
The existing simplify §5 rule ("use `voxelDispatchGridForCount()` rather than computing `(n+63)/64` manually") targets **CPU C++** code that can call the helper. It does not — and cannot — fire for a **GPU shader** that authors the dispatch dims itself (the whole point of a GPU-side finalize pass is to derive indirect args on-device to keep a hot scan barrier-free; a shader can't call the C++ helper). So a new shader that inlines the fold has no enforcement telling it to match `writeDispatchDims`, and the omission is a silent, config-dependent correctness bug rather than a compile error. The pattern is mechanically detectable: a shader writing `numGroupsX` (or `drawArgs[... + 0]`) from a `divCeil(count, tile)` expression with a sibling `numGroupsY = 1` and no `min(..., 1024)` cap. As the per-axis indirect-dispatch pattern spreads to more stages, each new finalize/dims-writing shader is a fresh chance to reintroduce this — a pre-commit flag closes it at authoring time.

## Occurrences
- PR #2273, `engine/render/src/shaders/c_per_axis_cell_finalize.glsl:39-41` (+ `metal/` twin) — uncapped 1-D `numGroupsX = divCeil(count, tile)`; fixed to a capped 2-D fold matching `writeDispatchDims`.

### LINKED ISSUE #2277: render: dispatch functions that borrow a shared GPU binding slot must restore it (coding-improvement)
**Class:** B (exists but didn't fire — enforcement gap for a documented render invariant)
**Target artifact:** `.claude/agents/review-invariant-render.md` — add a checklist item for transient shared-slot restoration; backstop in `.claude/agents/simplify-scan-render-leak.md` (or a new `simplify-*` grep pass) for the mechanically-detectable half.
**Proposed change:** `engine/render/CLAUDE.md` §"Gotchas" already documents the rule: a dispatch that needs a new SSBO/UBO while the voxel/per-axis path is active must reuse an existing binding transiently ("bind the new buffer onto an occupied index via `bindRange`/`bindBase` for the dispatch, then restore"). Neither `review-invariant-render.md` (the dedicated render-invariant PR reviewer subagent) nor any `simplify-*` check currently verifies the "then restore" half — a new dispatch function can borrow a shared `kBufferIndex_*` slot via `bindRange`/`bindBase` and never call it back without either surface flagging it. Add to `review-invariant-render.md`'s checklist: for any new function that `bindRange`s/`bindBase`s a `kBufferIndex_*` slot already bound elsewhere to a different buffer (i.e. a *reused* transient slot per the CLAUDE.md gotcha, not a slot the function/component owns permanently), confirm the same function restores the original binding before returning — do not assume a downstream system will restore it. Where mechanically feasible, extend `simplify-scan-render-leak` (or a sibling render-scan) to grep new `engine/prefabs/irreden/render/systems/*.hpp` dispatch functions for a `bindRange`/`bindBase` onto a known-shared slot (`kBufferIndex_PerAxisCell{Compacted,Indirect}`, `kBufferIndex_CompactedVoxelIndices`, `kBufferIndex_IndirectDispatchParams`) with no matching restore call in the same function body, as a pre-commit backstop ahead of review.

## Context
Surfaced fixing `fleet:has-nits` on PR #2273 (render: per-axis lighting/resolve indirect-dispatch, #2256). The Opus recheck (a manual, non-mechanical pass) flagged that the PR's four new per-axis dispatch functions (`dispatchPerAxisAO`, `dispatchPerAxisSunShadow`, `dispatchPerAxisLighting`, and `RESOLVE_PER_AXIS_SCREEN_DEPTH::tick`'s scatter pass) each `bindRange` SSBO slots 25/26 to their own per-axis cell buffers but never restore them to the voxel-compaction buffers afterward — only the pre-existing `system_trixel_to_framebuffer.hpp` site restores. Not a live bug today (that system happens to run last in the per-axis stage sequence every frame), but it silently widened the "must restore before frame end" contract from one documented, commented site to five undocumented ones that depend on pipeline order — a future reorder or new post-composite pass would reintroduce the #1961 world-voxel-corruption regression with no diagnostic. Fixed by centralizing the restore as `IRPrefab::PerAxisCanvas::restoreVoxelCompactionSlots()` and calling it from all five sites.

## Why it generalizes
The rule already exists in prose (`engine/render/CLAUDE.md`), so this is squarely a Class-B "didn't fire" gap, not a missing rule. The render pipeline has a fixed, fully-occupied 0–30 GPU binding-slot budget (documented in the same gotcha), so "borrow a slot transiently, then restore" is the *standard* idiom for adding a new dispatch to this pipeline — not a one-off. As the per-axis indirect-dispatch pattern (#1961, #2256, and future per-axis stages) keeps adding new dispatch functions that reuse the same handful of shared slots, each one is a fresh chance to add a borrow with no matching restore, and only a human/Opus manual read caught this instance — the dedicated render-invariant reviewer agent had zero coverage for it. Encoding the check once (agent checklist + optional mechanical grep) catches every future instance at review or pre-commit time instead of requiring another manual deep-read.

## Occurrences
- PR #2273 — `system_compute_voxel_ao.hpp` (`dispatchPerAxisAO`), `system_compute_sun_shadow.hpp` (`dispatchPerAxisSunShadow`), `system_lighting_to_trixel.hpp` (`dispatchPerAxisLighting`), `system_resolve_per_axis_screen_depth.hpp` (`tick`'s scatter pass) — all four borrowed SSBO slots 25/26 without restoring; fixed via a shared `IRPrefab::PerAxisCanvas::restoreVoxelCompactionSlots()` helper called from all five sites (including the pre-existing `system_trixel_to_framebuffer.hpp`, refactored to delegate instead of duplicating).

### LINKED ISSUE #2283: fleet/role-worker: step-1c must check thread for architect ruling before escalating a semantic-conflict to human (coding-improvement)
**Class:** A (missing rule)
**Target artifact:** `.claude/commands/role-worker.md` step 1c (semantic-conflict resolution), and/or `docs/agents/FLEET-FEEDBACK-HANDLING.md`
**Proposed change:** Before the step-1c escalation (step j) hands a semantic-conflict PR to `human:needs-fix`, scan the PR thread for a recent architect / `fleet:design-*` direction; if one already resolves the conflict's intent, execute it (or route via the `fleet:design-unblocked` / opus-resume path) instead of re-asking the human.

## Context
Surfaced fixing feedback on PR #2257 (`human:needs-fix`). The step-1c semantic-conflict pass at 2026-07-07 04:18Z escalated the PR to `human:needs-fix`, framing it as a "supersede vs redesign product decision" the worker couldn't make. But the architect had already posted a clear **re-scope** direction at 03:47Z — 30 minutes earlier, in the same thread — that resolved exactly that question (drop the retired-march shader work, keep the demo scene + refs as the regression gate, verify the geometric cap first). The escalation comment did not reference that direction. Executing the architect's direction (this AMEND) both resolved the conflict and dissolved it, so the human round-trip was avoidable.

## Why it generalizes
The step-1c flow reads only the merger's conflict comment (step b) before deciding to escalate; it has no step that checks whether the design authority already ruled on the conflicting intent. Whenever master lands a design-level change (here #2274 retiring a mechanism) that a PR conflicts with, an architect may resolve it in-thread *before* a worker's mechanical rebase pass runs — and the mechanical pass, seeing an irreconcilable textual conflict, escalates to `human:needs-fix` anyway. That produces (1) a contradictory-signal PR (architect says "re-scope, here's how"; worker says "human must decide"), (2) an avoidable human round-trip, and (3) rework for whoever untangles the timeline. A one-line "check the thread for an existing architect ruling first" guard in the escalation step closes the gap; the ruling, when present, is the resolution.

## Occurrences
- PR #2257, 04:18Z step-1c escalation vs 03:47Z architect re-scope direction — escalated to `human:needs-fix` despite an in-thread architect ruling that resolved the conflict.

### LINKED ISSUE #2285: scripts/fleet: config-file generators must preserve hand-edits under every emitted key (coding-improvement)
**Class:** A (missing rule)
**Target artifact:** `scripts/fleet/CLAUDE.md` (§ Authoring rules)
**Proposed change:** Add one bullet: "A generator that wholesale-rewrites a config file (e.g. `fleet-up`'s `write_worktree_settings` → `settings.local.json`) must preserve hand-edited content under **every** key it emits — adding a new generated key means extending the preservation logic (and a test) in the same change, or the new key silently clobbers human edits on the next regeneration."

## Context

Surfaced fixing feedback on PR #2284. The reviewer flagged that
`write_worktree_settings` regenerates each worktree's
`.claude/settings.local.json` preserving only `permissions.allow`; the PR
added the first `"hooks"` key that function ever wrote, so a hook
hand-added by a human (or the `update-config` skill) would be silently
wiped by the next `fleet-up` (`scripts/fleet/fleet-up:~908-975`).

## Why it generalizes

The same function already encoded this exact lesson once before — the
`permissions.allow` preservation exists because regeneration used to eat
user-granted permission entries. This PR repeated the class with a new
key (`hooks`), which is strong evidence the pattern recurs whenever the
generated settings surface grows (plausible future keys: `env`,
`statusLine`, `model` overrides). The failure mode is silent data loss of
human configuration with no error at write time, and nothing in
`scripts/fleet/CLAUDE.md`, `CLAUDE-BASELINE.md`, or the simplify checks
states the preserve-on-regenerate requirement today. A one-line authoring
rule at the surface fleet-tooling authors already read closes it at
authoring time.

## Occurrences

- PR #2284, `scripts/fleet/fleet-up:~908-975` — first `hooks` key added
  without extending preservation; fixed in-PR with marker-matched
  preservation + `scripts/fleet/tests/test_worktree_settings_hooks.sh`.

### LINKED ISSUE #2294: fleet/simplify: non-trivial inline bash in workflow run: blocks must be extracted to a tested script (coding-improvement)
**Class:** B (exists but didn't fire)
**Target artifact:** `.claude/skills/simplify/SKILL.md` §2b (Check 10, or a sibling check) — with the `review-pr` checklist as a backstop
**Proposed change:** Extend the "new `scripts/fleet/` executable ships with a `tests/test_*`" check to also flag **non-trivial bash embedded inline in a changed `.github/workflows/*.yml` `run:` block** — logic that defines a shell function, loops/branches over multiple commands, or exceeds ~15 lines should be extracted to a `scripts/fleet/*.sh` script with a hermetic `tests/test_*.sh`, exactly like a standalone executable.

## Context
Surfaced fixing feedback on PR #2290. The reviewer flagged a real classify-logic bug in `.github/workflows/auto-rereview.yml:74-97` (`net_patch_id` diffed both the before- and after-tip against the *current* `BASE_SHA`, so a stacked child retargeted onto master after its parent squash-merges was misclassified as a real content change → wrongly strips `fleet:approved`). The reviewer's own nit named the root cause of *why it shipped untested*: "the logic embedded in a `run:` block here is exactly the kind of thing that silently breaks with no CI signal … extracting … to a standalone script callable from a local git sandbox would have caught this before merge."

## Why it generalizes
The convention already exists — `scripts/fleet/CLAUDE.md` mandates "a new executable ships with a `tests/test_<name>.{sh,py}`", and simplify §2b Check 10 enforces it. But Check 10 only fires on **new files under `scripts/fleet/`**. Non-trivial bash authored **inline in a workflow `run:` block** is invisible to it — there is no new script file to match, so the pre-commit pass stays silent and the logic ships with zero test coverage. That is precisely the authoring-time blind spot that let the #2290 classify bug through: a reviewer had to hand-reproduce it in a sandbox because nothing forced extraction first. It is mechanically detectable (grep changed `.github/workflows/*.yml` for a `run:` block that defines a `foo() {` shell function or exceeds a line threshold, then check for a corresponding `scripts/fleet/*.sh` + test), and it recurs anywhere the fleet grows CI logic — the guard-and-classify pattern in these workflows is exactly where subtle git/patch-id bugs hide.

## Occurrences
- PR #2290, `.github/workflows/auto-rereview.yml:74-97` — non-trivial `net_patch_id` classify logic inline in a `run:` block shipped untested with a squash-merge-retarget misclassification bug; fix extracted it to `scripts/fleet/classify-auto-rereview.sh` + `tests/test_classify_auto_rereview.sh`.

### LINKED ISSUE #2329: docs/agents: AUTHOR-PIPELINE build-and-run must cover signature changes, not just new systems (coding-improvement)
**Class:** B (exists but didn't fire)
**Target artifact:** `docs/agents/AUTHOR-PIPELINE.md` § "Build and run"
**Proposed change:** Extend the existing "if the diff adds files under `engine/prefabs/**/systems/`, also build `IrredenEngineTest`" bullet to a general rule: any diff that changes the signature of a free function / method consumed elsewhere in the codebase (not just adds a new system) must build the full `IrredenEngineTest` target before the PR claims a test plan is green — not only the demo target the PR's own feature exercises.

## Context
Surfaced fixing PR #2313's `fleet:blocker`. The PR changed `gatherLightSources`'s
signature (`system_compute_light_volume.hpp`) but only rebuilt the demo targets
(`IRLightingEmissive`, `IRLightingSdfBlocker`, `IRShapeDebug`) named in its test
plan. Nine call sites in `test/render/per_canvas_light_scope_test.cpp` still
used the old signature, so `IrredenEngineTest` failed to compile — caught only
by a fleet reviewer, not by the author's own build/test pass.

## Why it generalizes
`AUTHOR-PIPELINE.md`'s current wording only triggers the full-suite build for
one specific failure mode (a new system missing its `SystemName` enum entry —
a *linker* error). A signature change to any function is a broader and more
common class that produces a *compile* error in whichever file calls it —
demo targets link only the systems/functions they reference, so a stale call
site living solely in `test/render/` is invisible to a demo-only build/run
pass. The existing bullet is the right home to extend rather than a new
surface: same section, same enforcement point (the author's own pre-PR
build step), just a broader trigger condition.

## Occurrences
- PR #2313, `system_compute_light_volume.hpp` signature change vs.
  `test/render/per_canvas_light_scope_test.cpp:76,80,100,104,127,132,161,176,191`
  — compile break caught by reviewer, not by author's build pass.

### LINKED ISSUE #2338: render: byte-identity/gated-off claims that lean on a spoofed shader predicate must be structurally verified (coding-improvement)
**Class:** B (exists but incomplete — the positive-enabled-path rule has no disabled-direction complement)
**Target artifact:** `engine/render/CLAUDE.md` (§ "Verifying render changes", beside the "Default-off features need a positive enabled-path test" paragraph)
**Proposed change:** A "byte-identity" / "gated-off" claim that leans on a shared shader predicate the C++ driver spoofs for sibling dispatches is EMPIRICAL, not structural — positively verify which dispatches actually take the gated path (or drive the intent from a dedicated per-dispatch C++ value), don't assume the predicate tracks the intended condition.

## Context
Surfaced fixing `fleet:needs-fix` on PR #2293 (#2270 sun-shadow bake coverage splat). The reviewer (Sonnet, confirmed by the Opus recheck) flagged that the splat gate `perAxisRoute == 0 && residualYaw == 0 && sunSplatMaxTexels > 0` is a **decode-path** predicate, not a camera-cardinality one: the CARDINAL-layout resolve dispatches (per-axis #1435, world-placed P4b-3) deliberately zero `residualYaw` to reuse the cardinal recovery, so the splat spuriously engaged on them. The PR's design doc claimed invariant #1 (per-axis / smooth-yaw byte-identity) held "by construction / structurally", but it actually leaned on an empirical density assumption; the claim was false during rotation.

## Why it generalizes
`engine/render/CLAUDE.md` already carries the enabled-direction rule: *"Default-off features need a positive enabled-path test, not just byte-identity at default"* (#1989 — byte-identity at default only proves the OFF path is a no-op, never that the feature works). This is the **disabled-direction complement**: a claim that a feature is OFF / byte-identical on certain paths is equally unproven when it rests on a shared shader gate that sibling dispatches spoof. The multi-dispatch resolve-then-bake pattern (per-axis resolve, world-placed resolve, future variants) reads resident shared UBO state (`FrameDataVoxelToCanvas`, `FrameDataSun`) that the C++ driver patches per dispatch — any gate over that state is a decode-path predicate, not a reliable proxy for camera/canvas identity. The fix drives intent from a dedicated per-dispatch C++ value (zero `sunSplatMaxTexels_` for the per-axis resolve) so the byte-identity becomes structural. This is the third PR on this exact bake surface (#1784 → #2204 → #2270), so the footgun recurs on this surface specifically.

## Occurrences
- PR #2293, `engine/render/src/shaders/c_bake_sun_shadow_map.glsl:158` (+ `system_bake_sun_shadow_map.hpp`) — splat gate leaned on `residualYaw == 0`, which the resolve dispatches spoof; the "structural byte-identity" claim was actually empirical. Also surfaced a real regression risk: gating the world-placed resolve off (the naive "make it structural" reading) shattered the cube's cast shadow 1 → 10 components, only caught by measuring.

### LINKED ISSUE #2345: fleet/simplify: flow-doc PR-body writes must use --body-file + show var population (coding-improvement)
**Class:** B (exists but didn't fire)
**Target artifact:** `.claude/skills/simplify/SKILL.md` §9a (Doc → code drift) — mechanically-detectable subcheck; backstop the `review-pr` checklist for markdown-only PRs
**Proposed change:** In a flow-doc bash fence, flag `gh pr edit|create --body "$var"` (and `gh issue create --body "$var"`) — recommend `--body-file` after fetching/assembling the body; also flag a `$var` used in a `--body`/substitution that is never assigned within the same fence.

## Context
Surfaced fixing `fleet:needs-fix` on PR #2342 (docs/fleet: factor attach-screenshots into a shared flow + SHA-pin its PR-body URLs). Two flow-doc snippets introduced a `pr_body` shell variable for a `@COMMIT_SHA@` / `<sha-pin token>` substitution but never showed where it was populated or how it reached the consuming `gh` call:

- `docs/agents/skills/commit-and-push.md:246-259` — the substitution assigned into `pr_body`, but the `gh pr create` example below built its body via an inline heredoc and never referenced `pr_body`, so the substitution was **inert** (nothing downstream consumed it — the stated URL-survival fix wasn't wired up).
- `docs/agents/FLEET-FEEDBACK-HANDLING.md:421-422` — worse: `gh pr edit <N> --body "$pr_body"` with `pr_body` never assigned. A worker following it literally runs `--body ""`, which **overwrites the entire PR body with an empty string** — wiping Summary, Test plan, and `Closes #N`.

## Why it generalizes
Flow docs are dense with copy-paste bash snippets that authoring workers run verbatim. The `--body-file` (never inline `--body "$var"`) convention already exists — REVIEWER-PROTOCOL.md § Posting the review body establishes it "for exactly this class of shell-substitution hazard," and the reviewer cited it — but it lives in a review-scoped doc and doesn't fire when a worker writes/edits a *flow-doc* snippet that writes a PR/issue body. The failure mode is a near-miss data-loss bug (a blanked PR body) plus a silently-inert fix, neither of which the build catches (docs aren't exercised). It is mechanically detectable: grep bash fences in changed `.md` for `gh pr (edit|create) ... --body "\$` and for a `$var` referenced in a `--body`/`${var//…}` line with no assignment (`var=`) earlier in the same fence. `simplify` §9a is the pre-commit surface a doc author reliably hits; the `review-pr` checklist is the backstop for markdown-only PRs (which skip the code checklist entirely).

## Occurrences
- PR #2342, `docs/agents/FLEET-FEEDBACK-HANDLING.md:421-422` — `gh pr edit --body "$pr_body"` with `pr_body` unset → blanks the whole PR body.
- PR #2342, `docs/agents/skills/commit-and-push.md:246-259` — substitution into `pr_body` never threaded to the `gh pr create --body` call → inert.

### LINKED ISSUE #2351: render/simplify: perf-overlay format rows must fit the documented column budget (coding-improvement)
**Class:** B (rule exists as in-file prose; not enforced pre-commit)
**Target artifact:** `.claude/skills/simplify/` — a new/extended `simplify-*` mechanical check over `engine/prefabs/irreden/render/systems/system_perf_stats_overlay.hpp`
**Proposed change:** In the simplify pass, flag any perf-overlay stats format row whose minimum rendered width (leading label + each printf conversion spec's minimum field width + literal separators, computed per `\n`-delimited row) exceeds `kPerfStatsColumnTrixels / kGlyphStepX` (= 22 chars) — the overlay's own documented column budget.

## Context
Surfaced fixing `fleet:needs-fix` feedback on PR #2347. The reviewer flagged that a newly added `FEEDER` row (`"  FEEDER %5.0f,%5.0f/%5.0f,%5.0f"`, a minimum of 32 chars) exceeds the overlay's own documented `≤ 22 chars (176 trixels at fontSize=1)` column budget (`kPerfStatsColumnTrixels = 176`, `kGlyphStepX = 8`), silently clipping `shadowFeederMax_.y` on the fixed-width GUI panel (`engine/prefabs/irreden/render/systems/system_perf_stats_overlay.hpp:328-329`). Fixed by splitting into `FEED MIN x,y` / `FEED MAX x,y` rows (22 chars each).

## Why it generalizes
The perf-stats overlay is under active development — multiple in-flight PRs add rows (this one added FEEDER; #2280 wires the reserved `voxelStage1` sub-rows). The column-budget contract is documented only *in the file's own comment* and nothing enforces it, so every new row relies on the author hand-counting characters (including printf minimum field widths) and a reviewer re-counting — exactly the manual arithmetic that burned this review round-trip. A row's minimum width is mechanically computable from the format literal (fixed characters + each conversion spec's minimum field width + the `\n` split points), so a pre-commit check can flag an over-budget row at authoring time. Class B because the rule already exists as prose in the source; the gap is enforcement/surfacing, not the rule text.

## Occurrences
- PR #2347, `engine/prefabs/irreden/render/systems/system_perf_stats_overlay.hpp:328-329` — `FEEDER` row min 32 chars vs the 22-char budget; split into `FEED MIN` / `FEED MAX` rows.

### LINKED ISSUE #2354: engine/math: add an IRMath wrap-into-range helper; route the hand-rolled fmod+correction idiom through it (coding-improvement)
**Class:** A (missing rule / missing shared helper)
**Target artifact:** `engine/math/include/irreden/ir_math.hpp` (add the helper) + a one-line note in `.claude/rules/cpp-math.md` (route angle/range wrapping through it)
**Proposed change:** Add an `IRMath` wrap-into-range helper (e.g. `IRMath::wrapToRange(x, range) -> [0, range)` and thin angle wrappers `wrapAngleTwoPi(a) -> [0, 2π)` / `wrapAnglePi(a) -> [-π, π)`), and have `cpp-math.md` mandate it so the `fmod(...) → if (< 0) += ...` idiom stops being hand-rolled.

## Context
Surfaced fixing feedback on PR #2352 (`fleet:has-nits`). Nit 1 asked to dedup `perf_grid`'s `waveFreezeOffset` into the component; the new `C_PeriodicIdle::valueAtAngle` had to wrap a raw angle into `[0, 2π)`, and the simplify reuse fan-out then flagged that the same `fmod → if (< 0) +=` wrap-into-range idiom is copy-pasted across the tree with no shared helper.

## Why it generalizes
"Wrap a scalar into `[0, range)`" (or an angle into `[0, 2π)` / `[-π, π)`) is a recurring primitive that every site currently re-derives by hand as `fmod` + a negative-correction branch. Two of the pre-existing sites reach for `std::fmod` directly — a live `cpp-math.md` violation (math primitives must go through IRMath) that a shared helper would eliminate at the source. A single `IRMath` helper gives one correct implementation (right half-open interval, correct negative handling), keeps CPU↔GPU consistency in one place (per the `IRMath::roundHalfUp` rationale), and lets the existing math-primitive `simplify` check flag any remaining hand-rolled `fmod`+correction as "use `IRMath::wrapToRange`".

## Occurrences
- PR #2352, `engine/prefabs/irreden/update/components/component_periodic_idle.hpp` (`valueAtAngle`) — `IRMath::fmod(angle, kTwoPi)` + `if (wrapped < 0) wrapped += kTwoPi` → `[0, 2π)`.
- `engine/prefabs/irreden/render/camera.hpp:87-88` (`wrapYaw`) — `IRMath::fmod(yaw + kPi, kTwoPi)` + negative-correction → `[-π, π)`.
- `engine/prefabs/irreden/update/systems/system_action_animation.hpp:64-70` — `std::fmod(elapsed, period)` + `if (< 0) += period` → `[0, period)` (also a `std::fmod` cpp-math violation).
- `engine/prefabs/irreden/update/systems/system_spring_platform.hpp:138-142` — `std::fmod(elapsed, period)` + `if (< 0) += period` → `[0, period)` (also a `std::fmod` cpp-math violation).

### LINKED ISSUE #2355: render: compile-time-specialize hot compute-kernel mode branches (a uniform branch is predicated, not skipped) (coding-improvement)
**Class:** A (rule missing from every authoring/review enforcement surface — it now exists only in the design doc added by this PR)
**Target artifact:** `docs/design/gpu-stage-timing-cost-model.md` (extend the dispatch cost model) + the `review-pr` render-invariant checklist (backstop) + a one-line cross-link in `engine/render/CLAUDE.md` "GPU stage timing" / gotchas
**Proposed change:** "A runtime *uniform* mode-branch added to a hot compute kernel is **predicated** (its instructions decode on every invocation, even when never taken), not skipped — so it taxes the kernel at all inputs. Prefer a **compile-time `#if` specialization** from one shared source body (two thin wrappers `#define` the mode, `#include` the body; register each variant in Metal's `threadgroupSizeForFunctionName` + `functionUsesImageAtomicScratch`). See `docs/design/voxel-feeder-split.md` §'#2258 Step B'."

## Context
Surfaced resolving the design escalation on PR #2325 (#2258 Step B). Step B v1 selected visible-vs-feeder inside the ONE shared `c_voxel_to_trixel_stage_1` kernel via a runtime `feederPass` uniform. Same-session profiling found a **+16 % zoom8 `voxelStage1` regression** even though the feeder branch is a tiny fraction of invocations; a disarm experiment (gate off, dispatch skipped) left stage-1 at the taxed number, proving the cost was **shader-side predication of the strided divides**, not dispatch overhead. Architect ruling a′ was to compile-time-specialize the two kernels from a shared body — the tax then measured away (zoom8 `voxelStage1` 5.97 → 5.33, zoom16 win retained).

## Why it generalizes
The existing cost model (`gpu-stage-timing-cost-model.md`) already documents "shader-side early-return cannot reclaim CPU-fixed dispatch-grid cost." This is the sibling lesson on the *body* side: a uniform-gated branch is not free even when the branch is never taken, because the compiler predicates rather than skips it — a footgun for **any** of the engine's hot compute kernels (voxel stage 1/2, `c_shapes_to_trixel`, lighting, sun-shadow, particle rasters) whenever a future PR adds a mode/variant branch. Encoding the "compile-time specialize, don't uniform-branch" idiom on an authoring/review surface catches it before it becomes another design-escalation round-trip. Only in the design doc today (so a reviewer wouldn't reliably flag the next occurrence).

## Occurrences
- PR #2325, `c_voxel_to_trixel_stage_1.glsl`/`.metal` (Step B v1 runtime `feederPass` branch) — +16 % zoom8 `voxelStage1` predication tax; fixed via compile-time `IR_FEEDER_PASS` specialization.

### LINKED ISSUE #2359: render: rebasing a shader-kernel/encoding PR past a carrier migration must re-verify helper arity + carrier presence (coding-improvement)
**Class:** B (exists but incomplete — the encode-helper routing + byte-identity rules have no rebase-past-a-migration complement)
**Target artifact:** `engine/render/CLAUDE.md` § "Verifying render changes" (author surface; add the review-pr checklist as a backstop)
**Proposed change:** When rebasing a shader-kernel or distance-encoding PR onto master, re-verify against master's CURRENT source: (a) every `encode*`/`decode*` helper call's arity/signature, and (b) that any polarity/priority/flip carrier a concurrent migration added is threaded through the rebased kernel on the ENABLED path. Byte-identity-at-default does NOT prove the carrier survived the rebase — a stale open-coded encode or an extracted-then-drifted kernel compiles and passes cardinal byte-identity while silently mis-encoding/dropping the carrier on the rotated/enabled path.

## Context
Surfaced fixing `fleet:needs-fix` on PR #2325 (#2258 Step B, opus recheck). The branch extracted the stage-1 voxel kernel into a shared `c_voxel_to_trixel_stage_1_body.{glsl,metal}` while forking *before* #2207 (the polarity-carrier PR) reworked that exact kernel — `encodeDepthWithFaceFrac` gained a 5th `flip` arg and `riserFlip`/`bothPolaritiesExposed` were added. On rebase onto #2207-master the extracted body both (a) failed to compile (4-arg calls to the now-5-arg helper) and (b) carried zero `riserFlip`, silently reintroducing the near-black venetian-riser regression #2207 shipped to fix. A mechanical `--onto` rebase reconstructs neither.

## Why it generalizes
This is the **rebase-past-a-migration** complement to two rules render/CLAUDE.md already carries: the #2207 "route every consumer through the shared encode/decode helpers — an open-coded shift is how a carrier migration silently mis-decodes" mandate, and the #1989/#2338 "byte-identity-at-default does not prove the feature works / the OFF claim holds." Neither fires at the moment the footgun bites: a PR that predates a carrier/encoding migration and refactors, extracts, or open-codes the same kernel. Byte-identity-at-default (non-rotated / flip=0 / cull-off) passes trivially, so the drop is invisible until an enabled-path render or a compile break surfaces it — and if the encode is open-coded (not a helper call), there is no compile break at all, only a silent no-op. Two occurrences in one session on two independent PRs, both against #2207's surface, is the recurrence signal.

## Occurrences
- PR #2325, `engine/render/src/shaders/c_voxel_to_trixel_stage_1_body.{glsl,metal}` — extracted stage-1 body predated #2207; rebase dropped `riserFlip` + hit a 4-arg `encodeDepthWithFaceFrac` compile break; re-threaded the carrier by hand.
- PR #2278, `engine/render/src/shaders/c_voxel_visibility_compact.{glsl,metal}` — open-coded `pos3DtoDistance(voxelPos) * 4` for the cardinal depth scale that #2207 changed to `* kDepthEncodeShift` (×8); rebase left a half-scale value compared against ×8 Hi-Z, silently no-op'ing the per-voxel occlusion cull (marginal capture measured 0). Fixed by routing through `encodeDepthWithFace(...,0)`.

### LINKED ISSUE #2379: attach-screenshots: sun-shadow/lighting shader diffs must pick canvas_stress + --debug-overlay, not IRShapeDebug (coding-improvement)
**Class:** B (exists but didn't fire)
**Target artifact:** `.claude/skills/attach-screenshots/SKILL.md` (§2 "Pick the demo target") — plus a one-line cross-ref in `engine/render/CLAUDE.md` §"Verifying render changes"
**Proposed change:** When the shader diff is a sun-shadow / lighting / AO change (`ir_sun_*`, `c_bake_sun_shadow_map`, `c_compute_sun_shadow`, `ir_sun_shadow_sample`, `c_compute_voxel_ao`, `c_lighting_to_trixel`, `c_*light_volume*`), the demo-picker must select `IRCanvasStress` (or the lighting demo) and run with `--debug-overlay shadow|ao|light_level --no-auto-rotate --no-spin`, not the default `IRShapeDebug`. Those effects do not render in `shape_debug`, so its before/after shots are byte-identical and read as "no visual change".

## Context
Surfaced fixing the `fleet:needs-fix` on PR #2343 (sun-shadow receive same-plane test, #2319). The reviewer flagged "no screenshots attached" for an `engine/render/src/shaders/` change. Producing meaningful ones required a manual two-ref `canvas_stress` capture (`--only floor --debug-overlay shadow --no-auto-rotate --no-spin`) — the floor-acne / cube-top self-hit delta is invisible in the default `IRShapeDebug` suite the skill's picker (§2 rule 2: "diff touches only engine render code → default to `IRShapeDebug`") would have chosen.

## Why it generalizes
The requirement (before/after + ROI-crop, `engine/render/CLAUDE.md` §"Verifying render changes") exists, and the `attach-screenshots` skill exists to satisfy it — but its demo-picker is wrong for the entire sun-shadow / lighting / AO shader family, which only produces a visible delta in `canvas_stress` (or the lighting demo) under a `--debug-overlay` mode at a frozen pose. A worker following the skill verbatim on any such PR attaches `IRShapeDebug` shots that show no delta, satisfying the letter of "attach screenshots" while giving the reviewer zero evidence — and inviting a spurious re-review. The routing is mechanically decidable from the diff paths, so it belongs in the picker (strongest "get it right the first time" lever), with the enabling flags (`--debug-overlay`, freeze) named alongside.

## Occurrences
- PR #2343, `engine/render/src/shaders/ir_sun_shadow_sample.{glsl,metal}` + `ir_sun_projection` + `c_bake_sun_shadow_map` — sun-shadow receive change; visible delta only in `canvas_stress --debug-overlay shadow --no-auto-rotate --no-spin`, not `IRShapeDebug`.

### LINKED ISSUE #2382: scripts/fleet: ff-only advance is not a dirty-tree guard — a disjoint dirty tree fast-forwards silently (coding-improvement)
**Class:** A (missing rule)
**Target artifact:** `scripts/fleet/CLAUDE.md` (§Authoring rules)
**Proposed change:** `git merge --ff-only` / an ff-advance is **not** a dirty-tree guard — it refuses only when the incoming commits *overlap* the dirty files, so a *disjoint* dirty tree fast-forwards silently. Any code that advances or restores a shared clone's branch must gate on an explicit `git status --porcelain` tracked-dirty check **before** the fetch/merge, and that guard must cover every path (not just the off-master one).

## Context
Surfaced fixing `fleet:needs-fix` on PR #2378. `restore_main_clone_to_master` in `scripts/fleet/fleet-clone-freshness.sh` computed `tracked_dirty` but only checked it inside the `branch != "master"` arm. A main clone **already on master** with a disjoint tracked-dirty tree fell through to the unconditional `git fetch` + `_ff_advance_to_origin_master`, whose `git merge --ff-only` advanced the branch tip silently under uncommitted work (`scripts/fleet/fleet-clone-freshness.sh:189-213`). Fix hoisted the guard above the conditional; added test T14 mirroring T6.

## Why it generalizes
The trap is that ff-only *looks* like it protects the working tree — and it does, but only for the overlapping-file case; the disjoint case (the common one for a WIP file on a clone whose `master` is behind on unrelated commits) advances with no error and no warning, which is WIP-loss-adjacent. The sibling `_ff_advance_to_origin_master` even documents the overlap-only limitation in its own comment ("a disjoint dirty tree advances fine"), yet the bug still shipped because the caller leaned on ff-only instead of gating explicitly. This recurs anywhere the fleet adds a new advance/restore/sync path over a shared clone (`advance_main_clone`, future `fleet-up` / dispatcher variants). A one-line rule on the authoring surface — plus a reviewer heuristic to grep new `merge --ff-only` / ff-advance sites for a preceding porcelain guard on **every** branch path — catches it at authoring time.

## Occurrences
- PR #2378, `scripts/fleet/fleet-clone-freshness.sh:189-213` — `tracked_dirty` guard scoped to the `branch != master` arm; on-master disjoint dirty tree silently ff-advanced.

### LINKED ISSUE #2389: lua-ecs: Lua createEntityBatch* call whose factory-arg count != C++ registration arity silently drops components (coding-improvement)
**Class:** A (missing rule)
**Target artifact:** `.claude/rules/cpp-ecs-smells.md` § "Lua-side ECS bindings" (+ the `simplify-check-ecs` subagent for the mechanical-detection extension)
**Proposed change:** Flag a Lua `createEntityBatch*` call whose factory-argument count differs from the C++ `registerCreateEntityBatchFunction<...>` arity registered under that name — the fixed-arity C binding silently ignores extra factory args, so the intended component is never attached (no error, dead code).

## Context
Surfaced while fixing review feedback on a downstream creation PR. An author added a component to a batch-spawned entity by appending one more factory closure to an existing `createEntityBatch*` Lua call — the same move that works for a C++ `IREntity::createEntity(a, b, c, ...)` spawn. But that Lua name was registered via `registerCreateEntityBatchFunction<...>(...)` with fewer component types than factories now passed, and its generated binding (`wrapCreateEntityBatchWithFunctions<...>` in `engine/script/include/irreden/script/lua_script.hpp`) reads exactly N positional `sol::protected_function` params. The extra factory was silently dropped and the intended component never attached; a system whose archetype required that component therefore never matched and no-op'd — dead code that compiled, ran, and looked identical to a working build.

## Why it generalizes
`registerCreateEntityBatchFunction<Components...>` binds a fixed-arity C function. Standard Lua semantics ignore args past a C function's fixed arity with no error, so "add a component by appending a factory" — intuitive, and correct for C++ `createEntity` spawns — silently no-ops for the Lua batch path. The failure is invisible on a build + screenshot check when the added component only feeds an archetype query. No engine rule / module `CLAUDE.md` / simplify surface documents that a Lua `createEntityBatch*` call's factory-arg count must match the C++ registration arity. Mechanically detectable: for each `registerCreateEntityBatchFunction<...>("name")` count the template args, then flag Lua `name(...)` call sites whose factory-arg count differs.

## Occurrences
- Review-caught in a downstream creation PR: a Lua batch spawn passed one more factory than its `registerCreateEntityBatchFunction<...>` registration arity; the extra factory was silently dropped and the component never attached, shipping as dead code until a reviewer hand-traced the binding arity.

jakildev:
## Summary

Human-cued `triage-coding-improvements` batch over the 18 open `fleet:coding-improvement` tickets. Verdicts (all human-approved at triage): **14 ACCEPT, 1 RESCOPE, 2 ROUTE-OUT, 1 DEFER, 0 REJECT.** Gated self-config files in this diff are expected — this is the human-cued triage lane; the diff maps 1:1 to the digest below.

### Accepted → applied

| Ticket | Surface | Rule |
|---|---|---|
| #2359 | `engine/render/CLAUDE.md` §Verifying render changes (+ review-pr backstop) | rebase past a carrier/encoding migration → re-verify helper arity + carrier on the enabled path |
| #2338 | `engine/render/CLAUDE.md` §Verifying render changes | spoofed-shared-predicate "gated-off" claims are empirical, not structural (disabled-direction complement of #1989) |
| #2355 | `docs/design/gpu-stage-timing-cost-model.md` §2 + render gotcha + review-pr backstop | uniform mode-branch in a hot kernel is predicated, not skipped → compile-time `#if` specialization |
| #2379 | both attach-screenshots pickers + render CLAUDE.md note | sun-shadow/lighting/AO shader diffs route to `IRCanvasStress --debug-overlay … --no-auto-rotate --no-spin`, not `IRShapeDebug` |
| #2345 | `simplify` §9a + review-pr "Flow docs" backstop | flow-doc `gh … --body "$var"` → `--body-file`; flag never-assigned `$var` in a fence |
| #2294 | `simplify` §2b Check 10 extension | non-trivial inline bash in workflow `run:` blocks → extract to tested `scripts/fleet/*.sh` |
| #2276 | `simplify` §5 + review-pr backstop | shader-authored indirect-dispatch dims must cap `numGroupsX` at 1024 + spill to Y (mirror `writeDispatchDims()`) |
| #2351 | `simplify` §5 | perf-overlay format rows must fit the 22-char column budget (min printf width per row) |
| #2277 | `review-invariant-render` item 9 + `simplify` §5 backstop | borrowed shared `kBufferIndex_*` slot must be restored in the same function; alias comment lists new consumers |
| #2382 | `scripts/fleet/CLAUDE.md` §Authoring rules | `--ff-only` is not a dirty-tree guard; explicit porcelain check on every branch path |
| #2285 | `scripts/fleet/CLAUDE.md` §Authoring rules | config generators preserve hand-edits under every emitted key |
| #2283 | `role-worker.md` step 1c(j) | scan PR thread for an existing architect ruling before escalating a semantic conflict to human |
| #2329 | `docs/agents/AUTHOR-PIPELINE.md` §Build and run | consumed-signature changes (not just new systems) trigger the full `IrredenEngineTest` build |
| #2261 | `engine/world/CLAUDE.md` §Process-default registry | registry additions must extend the IRPersist Lua round-trip test (prior DEFER; unblocked by #2244's merge) |

### Rescoped

- #2389 → doc rule only, in `.claude/rules/cpp-ecs-smells.md` §Lua-side ECS bindings (Lua `createEntityBatch*` factory count must match the C++ registration arity). The proposed `simplify-check-ecs` cross-language arity check is deferred until recurrence — real implementation work for a single-occurrence ticket.

### Routed out (Step 5)

- #2354 → code work split to #2397 (`IRMath::wrapToRange` helper + migrate 4 hand-rolled sites + the `cpp-math.md` bullet, which can't land before the helper exists). Closed via this PR with #2397 carrying the remainder.
- #2364 → de-labeled (engine-wide "caller must finalize" audit is code/audit work, not a convention edit); stays open as a plain issue for queue approval.

### Deferred (left open)

- #1865 — device-backend graceful-degradation tests; unchanged prior verdict, unblocks on a second device backend shipping without them.

### Closed-loop findings

- #2237 (closed-completed) created simplify Check 10; #2294 is that gap recurring one surface over (inline workflow bash) — extension applied.
- #1775 (closed-completed) landed the "reuse via bindRange" rule; the "then restore" half went unenforced → #2277's checklist item is the escalate-placement move.
- #2024 (closed-completed) landed the enabled-direction byte-identity rule; #2338 is its genuine disabled-direction sibling, placed beside it.

### Applied-placement note

#2277's mechanical backstop landed in `simplify` §5 rather than `simplify-scan-render-leak` (the ticket's either/or): that scanner explicitly filters to files *outside* the render dirs, while the restore check targets render systems — §5 is the render-scoped inline check, same enforcement position.

## Test plan

- [x] All cited symbols/paths verified against the tree (`writeDispatchDims`, `kPerfStatsColumnTrixels`/`kGlyphStepX` (176/8 = 22), `wrapCreateEntityBatchWithFunctions`, `test/script/lua_world_snapshot_test.cpp`, `IRCanvasStress` + `--debug-overlay`/freeze flags, REVIEWER-PROTOCOL.md §Posting the review body, `voxel-feeder-split.md` §"#2258 Step B", `ir_render_types.hpp` slot aliases)
- [x] Docs-only diff — no build surface
- [ ] Reviewer: confirm the diff maps 1:1 to the digest above

Closes #2261
Closes #2276
Closes #2277
Closes #2283
Closes #2285
Closes #2294
Closes #2329
Closes #2338
Closes #2345
Closes #2351
Closes #2354
Closes #2355
Closes #2359
Closes #2379
Closes #2382
Closes #2389

### DIFF
--- a/.claude/agents/review-invariant-render.md
+++ b/.claude/agents/review-invariant-render.md
@@ -37,6 +37,8 @@ For changed files under `engine/render/`, `engine/prefabs/irreden/render/`, or a
 
 8. **Position-component selection.** A render system reading `C_Position3D` for visual placement instead of `C_PositionGlobal3D` (`APPLY_POSITION_OFFSET` has already folded any modifier-driven offset into globalPos).
 
+9. **Transient shared-slot restoration.** A new function that `bindRange`s/`bindBase`s a `kBufferIndex_*` slot already bound elsewhere to a different buffer (the reuse-transiently gotcha in `engine/render/CLAUDE.md` §Gotchas — e.g. `kBufferIndex_PerAxisCell{Compacted,Indirect}`, `kBufferIndex_CompactedVoxelIndices`, `kBufferIndex_IndirectDispatchParams`) must restore the original binding before returning — do not assume a downstream system restores it; that only holds until a pipeline reorder. Also confirm the aliasing constant's declaration comment in `ir_render_types.hpp` lists the new transient consumer, so a "who uses slot N" audit finds it.
+
 ## Lighting stage (additional checks)
 
 If the diff touches `system_*ao*`, `system_*shadow*`, `system_*flood*`, `system_*fog*`, `system_build_light_occlusion_grid*`, or `c_compute_*shadow*.glsl` / `.metal`:
--- a/.claude/commands/role-worker.md
+++ b/.claude/commands/role-worker.md
@@ -462,7 +462,15 @@ Do the work, then exit cleanly:
        watertight when the worker resolves before the merger fires):
        `gh pr edit <N> --repo jakildev/IrredenEngine --remove-label "fleet:merger-cooldown"`
        Then jump to step k (reset).
-    j. **Resolution failed (escalation).** When to escalate:
+    j. **Resolution failed (escalation).** Before escalating, scan the PR
+       thread (`gh pr view <N> --comments`) for an architect /
+       `fleet:design-*` direction posted since the conflict arose: if an
+       in-thread ruling already resolves the conflicting intent, execute
+       it (or route via the `fleet:design-unblocked` path) instead of
+       re-asking the human — escalating past an existing ruling produces a
+       contradictory-signal PR and an avoidable human round-trip (#2257:
+       the escalation ignored a re-scope direction posted 30 minutes
+       earlier in the same thread). When to escalate:
        - The two sides did substantively different things and you
          can't tell from the code which intent should win (master
          rewrote a function, PR also rewrote it, neither is a
--- a/.claude/rules/cpp-ecs-smells.md
+++ b/.claude/rules/cpp-ecs-smells.md
@@ -100,3 +100,14 @@ In any system `tick` function:
   the integer at the binding boundary. See
   [`cpp-lua-enums.md`](cpp-lua-enums.md) for the full pattern and
   allowlist.
+
+- **A Lua `createEntityBatch*` call whose factory-argument count differs
+  from the C++ `registerCreateEntityBatchFunction<...>` registration arity**
+  for that name. The generated binding
+  (`wrapCreateEntityBatchWithFunctions<...>`) reads exactly N positional
+  factories and Lua silently ignores args past a C function's fixed arity —
+  so "add a component by appending one more factory" (correct for C++
+  `createEntity` spawns) silently no-ops: no error, the component never
+  attaches, and a system whose archetype requires it just never matches.
+  Adding a factory means extending the C++ registration's template arity in
+  the same change.
--- a/.claude/skills/attach-screenshots/SKILL.md
+++ b/.claude/skills/attach-screenshots/SKILL.md
@@ -82,11 +82,21 @@ selection heuristic, in order:
 
 1. **Diff touches `creations/demos/<name>/`** → use `IR<NameCamelCase>`
    (e.g. `shape_debug/` → `IRShapeDebug`).
-2. **Diff touches only engine render code** (`engine/render/**`,
+2. **Diff touches the sun-shadow / lighting / AO shader family**
+   (`ir_sun_*`, `c_bake_sun_shadow_map`, `c_clear_sun_shadow_map`,
+   `c_compute_sun_shadow`, `ir_sun_shadow_sample`, `c_compute_voxel_ao`,
+   `c_lighting_to_trixel`, `c_*light_volume*`) → use `IRCanvasStress` and
+   run **both** passes with
+   `--debug-overlay shadow|ao|light_level --no-auto-rotate --no-spin`
+   (pick the overlay matching the effect; the freeze flags keep the pose
+   comparable). These effects do not render in `shape_debug`, so its
+   before/after shots come out byte-identical and read as "no visual
+   change" (#2343).
+3. **Diff touches only other engine render code** (`engine/render/**`,
    `engine/prefabs/irreden/render/**`, shaders) → default to
    `IRShapeDebug`, which exercises the trixel pipeline most broadly
    and ships `--auto-screenshot`.
-3. **Ambiguous** (multiple demo directories touched, or a mix of
+4. **Ambiguous** (multiple demo directories touched, or a mix of
    render and non-render paths) → prompt the worker for the demo
    name. Do not guess in ambiguous cases; a wrong demo produces
    useless screenshots.
--- a/.claude/skills/review-pr/SKILL.md
+++ b/.claude/skills/review-pr/SKILL.md
@@ -64,6 +64,20 @@ issue.
 - New shader file not following the `c_` / `v_` / `f_` / `g_` prefix.
 - Canvas allocation before the canvas entity exists.
 - Compute dispatch size doesn't match `voxelDispatchGridForCount()`.
+- A shader that writes indirect-dispatch dims from a runtime count without
+  capping `numGroupsX` at 1024 and spilling to `numGroupsY` (mirror
+  `writeDispatchDims()` in `c_voxel_visibility_compact.glsl`) — an uncapped
+  1-D grid is undefined past 65535 groups.
+- A new runtime *uniform* mode-branch added to a hot compute kernel (voxel
+  stage 1/2, `c_shapes_to_trixel`, lighting / sun-shadow / particle
+  kernels) — predicated, not skipped; should be a compile-time `#if`
+  specialization from a shared body
+  (`docs/design/gpu-stage-timing-cost-model.md` §2).
+- A rebased shader-kernel/encoding PR whose fork predates a carrier or
+  encoding migration on master: verify every `encode*`/`decode*` call's
+  arity and that the migration's carrier is threaded on the enabled path —
+  byte-identity at default doesn't prove it
+  (`engine/render/CLAUDE.md` §"Verifying render changes").
 - new `*.glsl` added without a matching `*.metal` counterpart (if parity is
   intentionally deferred, the PR body must acknowledge it and reference a
   follow-up task).
@@ -160,6 +174,14 @@ verdict footer if any of these surfaces are touched)
   to** this engine checklist. The engine checklist is the baseline; the
   creation's `CLAUDE.md` is the delta. Apply both. If the subdirectory has a
   dedicated `REVIEW.md`, prefer it for review-specific rules.
+
+**Flow docs** (markdown-only PRs skip the code checklist — still check this)
+- A bash fence in a changed `.md` that runs `gh pr edit|create` /
+  `gh issue create` with `--body "$var"` instead of `--body-file`, or that
+  consumes a `$var` in a `--body`/substitution line never assigned in the
+  same fence — a worker running it literally executes `--body ""` and wipes
+  the PR body (#2342; the `--body-file` rule generalizes
+  REVIEWER-PROTOCOL.md §Posting the review body to flow docs).
 
 **Fleet scripts** (if the diff touches `scripts/fleet/`)
 - Inline comments — layout diagrams, role-list blocks, `--help` banners —
--- a/.claude/skills/simplify/SKILL.md
+++ b/.claude/skills/simplify/SKILL.md
@@ -326,6 +326,14 @@ tool with no test_*; add one against a stubbed environment (see
 `scripts/fleet/CLAUDE.md` §Authoring rules for the hermeticity bar)."
 Report, don't auto-fix.
 
+The same rule fires on **non-trivial bash embedded inline in a changed
+`.github/workflows/*.yml` `run:` block** — logic that defines a shell
+function, loops/branches over multiple commands, or exceeds ~15 lines.
+Flag: "extract to a `scripts/fleet/*.sh` executable with a hermetic
+`tests/test_*.sh` and call it from the workflow" — inline `run:` logic
+ships with zero CI signal and no local-sandbox test (#2290's `net_patch_id`
+classify bug). Report, don't auto-fix.
+
 ### 2c. Serialized-struct version-bump check
 
 See `engine/asset/CLAUDE.md` §"Automated version-bump detection" for the full
@@ -374,6 +382,27 @@ Also check:
 - Hand-rolled compute dispatch sizes — should use
   `voxelDispatchGridForCount()` rather than computing `(n+63)/64`
   manually.
+- A **GPU shader** (`.glsl`/`.metal`) that writes indirect-dispatch dims
+  from a runtime count must cap `numGroupsX` at `kMaxDispatchGroupsX`
+  (1024) and spill the remainder into `numGroupsY` — mirror
+  `writeDispatchDims()` in `c_voxel_visibility_compact.glsl`; consumers
+  recover the flat group index as `groupId.x + groupId.y * numGroupsX`.
+  An uncapped 1-D `numGroupsX = divCeil(count, tile)` with
+  `numGroupsY = 1` is undefined past 65535 groups — silently dropped
+  cells, no error (#2273).
+- A dispatch function that `bindRange`s/`bindBase`s a shared
+  `kBufferIndex_*` slot (one bound elsewhere to a different buffer — the
+  reuse-transiently gotcha in `engine/render/CLAUDE.md` §Gotchas) and
+  returns without restoring the original binding. The same function must
+  restore; don't rely on a downstream system's restore surviving a
+  pipeline reorder (#2273).
+- A changed stats format row in `system_perf_stats_overlay.hpp` whose
+  minimum rendered width (leading label + each printf conversion spec's
+  minimum field width + literal separators, computed per `\n`-delimited
+  row) exceeds the overlay's documented column budget
+  `kPerfStatsColumnTrixels / kGlyphStepX` (= 22 chars) — an over-budget
+  row silently clips on the fixed-width panel; split it into two rows
+  (#2347).
 - 3D-world-coord values being mixed with iso-2D-coord values without
   going through `IRMath::pos3DtoPos2DIso` or a named helper. The two
   spaces are not interchangeable.
@@ -608,6 +637,13 @@ top-level `CLAUDE.md` and module-level `CLAUDE.md` files.
     avoid).
   - Decorative emoji bullets (`❌`, `✅`) — codebase convention is bare
     list bullets.
+- **PR/issue-body writes in bash fences.** A flow-doc fence that runs
+  `gh pr edit|create` or `gh issue create` with `--body "$var"` — require
+  `--body-file` after assembling the body (REVIEWER-PROTOCOL.md's
+  shell-substitution rule, generalized to flow docs). Also flag a `$var`
+  consumed in a `--body` or `${var//…}` substitution line with no `var=`
+  assignment earlier in the same fence — a worker running the snippet
+  literally executes `--body ""` and wipes the entire PR body (#2342).
 - **Broken cross-refs.** Every `[text](path)` / `[text](path#anchor)`
   link in the diff resolves to an existing file / heading. Section
   references cited as `§Foo` match an actual `## Foo` heading. Named
--- a/docs/agents/AUTHOR-PIPELINE.md
+++ b/docs/agents/AUTHOR-PIPELINE.md
@@ -24,11 +24,16 @@ between "branch is ready" and "PR is finalized."
 fleet-build --target <name>
 ```
 
-**If the diff adds files under `engine/prefabs/**/systems/`**, also
-build `IrredenEngineTest` (or the engine static library) — creation
-targets like `IRShapeDebug` only link the systems they reference, so
-a new system with a missing `SystemName` enum entry is a silent
-linker error from the creation perspective.
+**If the diff adds files under `engine/prefabs/**/systems/` — or changes
+the signature of any free function or method consumed elsewhere in the
+tree** — also build `IrredenEngineTest` (or the engine static library)
+before the PR claims its test plan is green. Creation targets like
+`IRShapeDebug` only compile and link what they reference, so a demo-only
+build is blind both to a new system's missing `SystemName` enum entry
+(silent linker error) and to stale call sites of a changed signature that
+live only in `test/**` (#2313: nine call sites in
+`test/render/per_canvas_light_scope_test.cpp` broke the suite build while
+the PR's demo targets built green).
 
 If the touched code has an executable target, run it to confirm it
 launches cleanly:
--- a/docs/agents/skills/attach-screenshots.md
+++ b/docs/agents/skills/attach-screenshots.md
@@ -94,10 +94,18 @@ selection heuristic, in order:
 
 1. **Diff touches `creations/demos/<name>/`** → use the matching
    `IR<NameCamelCase>` target.
-2. **Diff touches only engine render code** (matching the render-scoped
-   **visual-file globs**) → use the **default demo**, which exercises the
-   trixel pipeline most broadly and ships `--auto-screenshot`.
-3. **Ambiguous** (multiple demo directories touched, or a mix of render and
+2. **Diff touches an effect family the default demo does not render**
+   (per the repo wrapper's effect-routing table — e.g. the engine's
+   sun-shadow / lighting / AO shaders, which only produce a visible delta
+   in the lighting-capable stress demo under a `--debug-overlay` mode at a
+   frozen pose) → use the routed demo and flags for **both** passes. The
+   default demo's before/after shots for such diffs are byte-identical and
+   read as "no visual change", giving the reviewer zero evidence.
+3. **Diff touches only other engine render code** (matching the
+   render-scoped **visual-file globs**) → use the **default demo**, which
+   exercises the trixel pipeline most broadly and ships
+   `--auto-screenshot`.
+4. **Ambiguous** (multiple demo directories touched, or a mix of render and
    non-render paths) → prompt the worker for the demo name. Do not guess in
    ambiguous cases; a wrong demo produces useless screenshots.
 
--- a/docs/design/gpu-stage-timing-cost-model.md
+++ b/docs/design/gpu-stage-timing-cost-model.md
@@ -82,6 +82,19 @@ a measurement says otherwise on a specific path:
   grid is fixed CPU-side at `dispatchCompute` time; an invocation that
   returns immediately still launched. PR #2266 capped shadow-feeder
   micro-grids via early-return and measured no change in `voxelStage1`.
+- **A runtime *uniform* mode-branch in a hot kernel is predicated, not
+  skipped** — its instructions decode on every invocation even when the
+  branch is never taken, so it taxes the kernel at all inputs. #2325 Step B
+  v1 selected visible-vs-feeder via a `feederPass` uniform inside the shared
+  stage-1 kernel and regressed zoom8 `voxelStage1` +16 %; a disarm probe
+  (gate off, dispatch skipped) held the taxed number, proving the cost was
+  shader-side predication of the strided divides, not dispatch overhead.
+  Prefer a **compile-time `#if` specialization** from one shared source body:
+  two thin wrappers `#define` the mode and `#include` the body, and each
+  Metal variant registers in `threadgroupSizeForFunctionName` +
+  `functionUsesImageAtomicScratch`. Specializing measured the tax away
+  (zoom8 `voxelStage1` 5.97 → 5.33 ms, zoom16 win retained). See
+  `docs/design/voxel-feeder-split.md` §"#2258 Step B".
 - **The stage-1 raster body is not where the high-zoom cost lives — but the
   stage-1 *dispatch* still carries ~half of it.** Forcing *every* voxel to a
   single 1×1 micro-tap (a 256× per-invocation body reduction at zoom 16) left
--- a/engine/render/CLAUDE.md
+++ b/engine/render/CLAUDE.md
@@ -258,6 +258,14 @@ If the PR intentionally changes silhouettes / lighting / shading
 model, call out the intentional drift in the description so reviewers
 know the new crop is the new baseline rather than a regression.
 
+**Sun-shadow / lighting / AO shader diffs don't show in the default demo.**
+For a diff touching the sun-shadow / lighting / AO shader family the
+`attach-screenshots` picker routes to the lighting-capable stress demo with
+a `--debug-overlay` mode at a frozen pose (`IRCanvasStress --debug-overlay
+shadow|ao|light_level --no-auto-rotate --no-spin`) — the default
+`IRShapeDebug` suite renders none of those effects, so its before/after
+pairs come out byte-identical and read as "no visual change" (#2343).
+
 **Occlusion diagnostics for rotated voxel content: use `--checkerboard`,
 not `--depth-color`.** `--depth-color` quantizes hue in 4/3-world-unit
 bands; at any non-cardinal yaw the bands beat against the 1-unit voxel
@@ -343,6 +351,27 @@ binding-6 voxel upload, not a detached-revoxelize bake) can silently never
 reach the shader, and a "compiles + byte-identical at default" merge ships a
 feature that doesn't function in its actual use case (#1989 per-trixel
 priority caught exactly this on resume).
+
+**The disabled-direction complement: a "gated-off / byte-identical on path
+X" claim that leans on a shared shader predicate is empirical, not
+structural.** Multi-dispatch passes (per-axis resolve, world-placed resolve,
+resolve-then-bake) read resident shared UBO state (`FrameDataVoxelToCanvas`,
+`FrameDataSun`) that the C++ driver patches per dispatch, so a gate over
+that state is a decode-path predicate — a sibling dispatch may deliberately
+spoof the gated field (the per-axis resolve zeroes `residualYaw` to reuse
+the cardinal recovery) and take the gated path spuriously (#2293).
+Positively verify which dispatches actually take the gated path, or drive
+the intent from a dedicated per-dispatch C++ value so the byte-identity
+holds by construction.
+
+**Rebasing a shader-kernel or encoding PR past a carrier/encoding migration
+on master:** re-verify against master's *current* source — (a) every
+`encode*`/`decode*` helper call's arity and signature, and (b) that any
+polarity/priority/flip carrier the migration added is threaded through the
+rebased kernel on the ENABLED path. Byte-identity at default does not prove
+the carrier survived the rebase: an extracted or open-coded kernel that
+forked pre-migration compiles and passes cardinal byte-identity while
+silently dropping the carrier on the rotated/enabled path (#2325 vs #2207).
 
 ### Verifying temporal stability (per-frame jitter)
 
@@ -815,6 +844,11 @@ parity with voxel-pool primary shapes.
   garbage from the previous frame.
 - **Dispatch limits.** `kMaxDispatchGroupsX = 1024` (≈1M voxels before
   hitting the second dispatch dimension). Very large voxel counts slow.
+- **Mode branches in hot compute kernels: compile-time-specialize, don't
+  uniform-branch.** A runtime *uniform* branch is predicated, not skipped —
+  its instructions decode on every invocation even when never taken, taxing
+  the kernel at all inputs. Prefer a `#if` specialization from one shared
+  source body. See `docs/design/gpu-stage-timing-cost-model.md` §2.
 - **Render mode × subdivision × zoom.** `SMOOTH` mode multiplies positions
   by `subdivisions × zoom`. Changing any of these mid-frame is a perf
   cliff and can desync chunk visibility.
--- a/engine/world/CLAUDE.md
+++ b/engine/world/CLAUDE.md
@@ -372,7 +372,11 @@ compile: the heap-owning opted-in components (`C_Name`, `C_Skeleton`,
 primary-template `static_assert`. Add each component to the curated list as its
 serializer lands; the full-inventory path (a per-component serializer pass + a
 `HasExplicitSaveSerialize<C>` filter over `AllEngineComponents`) is tracked
-downstream. The registry is built **fresh per call** — cheap (a few
+downstream. **Adding a component here must also extend
+`test/script/lua_world_snapshot_test.cpp` with a round-trip case through the
+actual `IRPersist` Lua surface** — a standalone `SaveSerialize<C>` unit test
+(e.g. `voxel_set_serialize_test.cpp`) covers the serializer but leaves the
+wiring itself (registry entry → Lua binding → reload) unverified (#2244). The registry is built **fresh per call** — cheap (a few
 allocations, never a per-frame path) and, unlike a process-static, its
 session-local `ComponentId`s always match the live `EntityManager`.
 
--- a/scripts/fleet/CLAUDE.md
+++ b/scripts/fleet/CLAUDE.md
@@ -35,6 +35,21 @@ applies here too — see `docs/agents/CLAUDE-BASELINE.md` §Style.
   `summarize` exit idiom — don't re-copy the helpers into a new test.
   Genuinely test-specific asserts (path existence, exit codes) stay local,
   built on `ok`/`bad`.
+- **`git merge --ff-only` is not a dirty-tree guard.** It refuses only when
+  the incoming commits *overlap* the dirty files; a **disjoint** dirty tree
+  fast-forwards silently — WIP-loss-adjacent on a shared clone. Any path
+  that advances or restores a shared clone's branch gates on an explicit
+  `git status --porcelain` tracked-dirty check *before* the fetch/merge,
+  and the guard covers **every** branch path, not just the off-master arm
+  (#2378: the on-master arm fell through to an unguarded ff-advance).
+- **Config-file generators preserve hand-edits under every emitted key.**
+  A generator that wholesale-rewrites a config file (`fleet-up`'s
+  `write_worktree_settings` → `settings.local.json`) must carry
+  preservation logic — and a test — for **each** key it emits; adding a new
+  generated key means extending the preservation in the same change, or the
+  next regeneration silently clobbers human edits under that key (#2284:
+  the first `hooks` key repeated the lesson `permissions.allow` already
+  encoded).
 - **Unattended daemons timeout-guard their network calls.** The host's
   connections to GitHub intermittently black-hole (silent TCP death), so a
   hung `git fetch` / `gh …` in a fleet daemon (dispatcher loop, `fleet-rebase`,


jakildev (commented):
## Review — docs/fleet: absorb coding-improvement batch — 15 rules across 12 surfaces

**Verdict:** approve

Human-cued `triage-coding-improvements` batch result (18 tickets triaged,
14 accept / 1 rescope / 2 route-out / 1 defer). Gated self-config files
(`.claude/commands/role-worker.md`, `.claude/skills/*/SKILL.md`,
`.claude/agents/review-invariant-render.md`) are expected in this diff —
this is the designated human-cued channel for editing them, not an
autonomous self-edit.

### Diff-to-digest mapping
`git diff origin/master...HEAD --stat` shows exactly the 12 files the PR
body's digest table lists, and each of the 14 accepted + 1 rescoped rule
maps to the file/section the table claims (spot-checked all 15 hunks
against the table row-by-row — no orphaned hunks, no digest row missing
its diff). `Closes` list (16 issues) = 14 accepted + 1 rescoped (#2389) +
1 routed-out-but-closed-here (#2354); #2364 (the other routed-out ticket)
is correctly *absent* from `Closes` since the body says it stays open,
de-labeled — confirmed live: `#2364` is `OPEN` with no labels.

### Verification performed (cited symbols/paths, not just diff shape)
Rather than trust the "verified against the tree" test-plan checkbox,
independently re-checked every load-bearing citation:
- `writeDispatchDims()` exists in `c_voxel_visibility_compact.glsl:225`,
  called at lines 458/465/468 — matches the "mirror writeDispatchDims()"
  guidance added to `simplify`/`review-pr`.
- `kPerfStatsColumnTrixels = 176`, `kGlyphStepX = 8` in
  `system_perf_stats_overlay.hpp` — `176/8 = 22`, matches the column-budget
  claim exactly.
- `wrapCreateEntityBatchWithFunctions` exists (`ir_script.hpp`,
  `lua_script.hpp`); `test/script/lua_world_snapshot_test.cpp` and
  `test/world/voxel_set_serialize_test.cpp` both exist, matching the
  `engine/world/CLAUDE.md` and `cpp-ecs-smells.md` additions.
- `IRCanvasStress` supports `--debug-overlay`, `--no-auto-rotate`,
  `--no-spin` (`creations/demos/canvas_stress/main.cpp`), and
  `debugOverlayModeFromString` accepts `"shadow"`, `"ao"`, `"light_level"`
  (`ir_render_enums.hpp:136-149`) — matches the attach-screenshots routing
  rule's exact flag/mode names in both the wrapper and the shared flow doc.
- `kBufferIndex_PerAxisCellCompacted == kBufferIndex_CompactedVoxelIndices`
  (25) and `kBufferIndex_PerAxisCellIndirect ==
  kBufferIndex_IndirectDispatchParams` (26) in `ir_render_types.hpp` —
  confirms the aliasing example cited in the new
  `review-invariant-render.md` item 9 and the `simplify` backstop.
- `#2397` (the split-out `IRMath::wrapToRange` ticket) is `OPEN` with the
  expected title — the routed-out cross-reference is live, not a phantom
  issue number.
- Internal doc cross-refs resolve: `docs/design/voxel-feeder-split.md`
  has a `## #2258 Step B` heading (the `§"#2258 Step B"` citation is a
  valid prefix match), `engine/render/CLAUDE.md` has a
  `## Verifying render changes` heading, `REVIEWER-PROTOCOL.md` has a
  `## Posting the review body` heading.

No miscited precedent found — every "X does this, cite Y" claim in the
new rules checks out against the actual code/doc it names.

### Nits
(none)

- [x] Already covered — docs-only diff, no build surface. All cited
      symbols/paths/issue numbers independently re-verified by this
      review (see above), not just trusted from the test-plan checkbox.

Escalation: this PR touches `engine/render/CLAUDE.md` and
`engine/world/CLAUDE.md`, encoding several rules future reviews will
treat as ground truth (compile-time kernel specialization vs. uniform
branching, shared-slot restore-in-same-function, carrier/encoding
re-verification after a rebase). I independently verified every cited
symbol and cross-reference and found no inaccuracies, but given the
volume (15 rules) and that these become durable review/simplify
checklist items, a second read is warranted before they compound across
future PRs.

Opus recheck required: touches `engine/render/CLAUDE.md` +
`engine/world/CLAUDE.md`, encodes 15 new durable review/simplify rules
at once — worth a second confirmation given how many future PRs will be
judged against them, even though every citation checked out on this pass.

jakildev (commented):
## Opus recheck — docs/fleet: absorb coding-improvement batch (15 rules across 12 surfaces)

**Verdict:** approve

Sonnet flagged this for Opus recheck because the diff encodes 15 durable
review/simplify rules — touching `engine/render/CLAUDE.md` and
`engine/world/CLAUDE.md` — that future PRs will be judged against, and asked
for a second confirmation "even though every citation checked out on this
pass." Sonnet's pass verified every cited symbol/path/issue *exists*. On
closer read I focused on what that check can't establish: whether each rule
is **technically sound and appropriately scoped** — i.e. won't become a
durable false-positive/false-negative generator once it's on the checklist.

### Deeper verification (soundness, not just citation-existence)

Sonnet confirmed the symbols; I re-derived the load-bearing *claims* against
the actual code:

- **Indirect-dispatch cap rule** (`review-pr` + `simplify` + `engine/render/CLAUDE.md`
  + `gpu-stage-timing-cost-model.md`). The rule tells reviewers to "mirror
  `writeDispatchDims()`" and cap `numGroupsX` at `kMaxDispatchGroupsX (1024)`,
  spilling to `numGroupsY`, because an uncapped 1-D grid is undefined past
  65535 groups. Confirmed the exemplar is real and does exactly this:
  `voxel_dispatch_grid.hpp:33` (`min(count, kMaxDispatchGroupsX)`) and
  `c_per_axis_cell_finalize.glsl:52` (`max(min(groups, kMaxDispatchGroupsX), 1u)`).
  The latter even carries an in-tree comment stating the *same* rationale the
  new rule encodes — uncapped `numGroupsX` past
  `GL_MAX_COMPUTE_WORK_GROUP_COUNT[0]` (guaranteed min 65535) silently
  drops/corrupts cells, and consumers recover the flat index as
  `x + y * numWorkGroups.x`. The 1024-vs-65535 pair is internally consistent
  (1024 is the conservative engine cap; 65535 is the hardware limit that
  motivates it), and the four cross-file copies of this rule agree.

- **Perf-overlay column-budget rule** (`simplify`). Rule cites the budget as
  `kPerfStatsColumnTrixels / kGlyphStepX = 22 chars`. Confirmed at the source:
  `system_perf_stats_overlay.hpp:35` (`kPerfStatsColumnTrixels = 176`),
  `trixel_font.hpp:13` (`kGlyphStepX = 8`), and the overlay's own header
  comment reads "22 chars × kGlyphStepX (8) = 176 trixels" — the rule matches
  the code's own documented invariant exactly.

- **Shared-slot restoration** (`review-invariant-render.md` item 9) rests on
  the `kBufferIndex_PerAxisCell{Compacted,Indirect}` ↔
  `kBufferIndex_{CompactedVoxelIndices,IndirectDispatchParams}` aliasing
  Sonnet already confirmed (25/26); the guidance (restore in the same
  function; don't lean on a downstream restore that a pipeline reorder
  breaks) is a sound render invariant, not an over-broad trigger.

### Scoping / misfire check

Every rule has a **narrow, mechanical trigger** ("a shader that writes
indirect-dispatch dims from a runtime count without capping…", "a bash fence
running `gh pr edit` with `--body \"$var\"`…", "a stats format row whose
minimum rendered width exceeds…") and each is anchored to a concrete
war-story PR/issue (#2273, #2313, #2342, #2347, #2244, #2378, #2284, #2290,
#2325/#2207, #2293, #2257). None is broad enough to fire on unrelated diffs.
I checked the cross-file duplicates for contradiction — the dispatch-cap rule
(4 files) and the `--body-file` rule (review-pr + simplify) are consistent
wordings of the same invariant, not divergent ones. The `engine/world/CLAUDE.md`
insertion reads cleanly mid-paragraph (new sentence between "…tracked
downstream." and "The registry is built **fresh per call**").

### Gated self-config

Confirmed Sonnet's framing: the gated files (`.claude/commands/role-worker.md`,
`.claude/skills/*/SKILL.md`, `.claude/agents/review-invariant-render.md`) are
edited here through the designated human-cued `triage-coding-improvements`
channel — not an autonomous self-edit — so the gate is satisfied.

### Nits
(none)

No inaccuracies, no over-broad rules, no internal contradictions. On the
deeper technical read every rule's exemplar and threshold is grounded in real
code. Clearing the recheck.

🤖 Reviewed by Claude Opus 4.8 (opus-reviewer, review-pr skill)

PR #2410, opened 2026-07-14
render: sub-cell frac in per-axis lighting recovery + static occupancy

jakildev:
## Summary

Fixes the yaw-rotation face-shading regression the user reported in IRPerfGrid ("faces not aligning / parity issue" under rotation, IRCanvasStress unaffected).

**Root cause (two coupled legs, exposed by #2251):**
- #2251's `roundHalfUp` conversion places half-integer tie content at cell − 0.5, with the true position carried in the per-axis encoding's 4-bit frac field. The forward scatter honors the frac (geometry correct), but the **absolute-position lighting consumers** — light-volume sample, sun-shadow receive, overflow relight — recovered positions **lattice-only** via `perAxisCellToWorld3D`, sampling up to half a world cell *inside* the solid on every camera-facing surface. Pre-#2251, Metal's implementation-defined ties-away `round()` happened to land those cells outside the solid, masking the gap.
- perf_grid stamped **cross-set face occupancy** only in `--wave-mode rigid`, so the (post-#2249) default contiguous per-cell scene emitted every covered interior face; under yaw those quads tie with the true surface at exact depths in the scatter composite and the per-cell tie winner alternates — the chevron stipple.

**Changes:**
- New `perAxisCellToWorld3DSubCell` in `ir_per_axis_lighting.{glsl,metal}` — lattice recovery + frac offset, mirroring the scatter's reconstruction. Switched: `c_lighting_to_trixel` (volume sample), `c_compute_sun_shadow` (receive), `c_light_overflow_faces` (relight), both backends. AO deliberately keeps the lattice form — its `dot(neighbour − pos, outward)` provably cancels the in-plane offset (documented at the call sites).
- perf_grid: occupancy stamp now fires for every always-touching configuration (rigid OR zero amplitude); an active per-cell wave still shears neighbors apart and stays unstamped.
- perf_grid: `--debug-overlay` passthrough (ao/light_level/shadow/peraxis_id/peraxis_origin/unlit) — the isolation ladder used to diagnose this.
- `engine/render/CLAUDE.md`: one paragraph documenting the sub-cell-vs-lattice recovery invariant.

## Before / after (zoom4_rot, yaw 0.35, static contiguous 16³, Metal)

| before | after |
|---|---|
| before | after |

Crop detail:

| before | after |
|---|---|
| before-crop | after-crop |

## Test plan

- [x] Bisect: clean at #2249 (34bad77a), stipple at #2251 (c416457e) — VoxelSet contiguous static, yaw 0.35, Metal
- [x] Fixed build: perf_grid static contiguous grid solid under yaw (zoom1_rot + zoom4_rot), cardinal shots bit-clean vs master (drift = HUD digits only)
- [x] shape_debug: 16/16 comparable shots byte-identical vs master (incl. yaw 45/90/180/270) — integer content frac 8/8 → zero offset by construction
- [x] canvas_stress: byte-identical vs master (5/6 zero drift; 1 shot 8px run-noise at cardinal)
- [x] Occupancy stamp side effect verified: compact-pass interior drop now fires (VOX 1536/4096 shell)
- [ ] Linux/OpenGL smoke (authored on macOS/Metal; GLSL edits mirror the Metal ones)

## Known related issues (not addressed here — filing separately)

1. **Moving per-cell wave still stipples under yaw**: an active wave legitimately part-uncovers faces (occupancy can't be stamped), and the per-axis scatter's equal-depth tie resolution between different-axis quads alternates per cell. Needs a design pass on the #2255 tie-band ordering (slot priority vs cell identity within the fp32 sub-band budget).
2. **Pre-existing master segfault (macOS/Metal)**: IRPerfGrid crashes at the zoom4_pan shot (yaw 0.35 → 0 transition) and IRShapeDebug at zoom8_yaw180 — reproduced on clean origin/master (20837602), not introduced here.


jakildev (commented):
## Review — render: sub-cell frac in per-axis lighting recovery + static occupancy

**Verdict:** approve

### What this does
Fixes a yaw-rotation chevron-stipple regression in IRPerfGrid caused by two
coupled legs exposed by #2251: (1) absolute-position per-axis lighting
consumers (light-volume sample, sun-shadow receive, overflow relight)
recovered world position lattice-only, dropping the encoding's 4-bit
in-plane frac and sampling up to half a world cell inside the solid; (2)
perf_grid only stamped cross-set face occupancy in rigid wave mode, so the
default per-cell scene emitted covered interior faces that tie-fight the
true surface under yaw.

### Verified (static analysis — could not check out this branch; see note)
- **`perAxisCellToWorld3DSubCell` bit-math is correct and mirrors the
  existing proven decode.** Cross-checked the new GLSL/Metal helper's
  `uFrac4 = (encoded >> 6) & 15` / `vFrac4 = (encoded >> 2) & 15` and the
  `eu * (frac/16.0 - 0.5) + ev * (...)` reconstruction against
  `v_peraxis_scatter.glsl`'s existing decode (lines ~200-231) — identical
  bit positions, identical formula, same `faceInPlaneUnitAxes(faceId >> 1,
  ...)` call shape. This is exactly the "same reconstruction the scatter
  draws" the PR claims, not a re-derivation that could drift.
- **Every call site passes the correct `encoded` value.** Traced all four
  sites: `c_compute_sun_shadow.glsl`/`.metal` and `c_lighting_to_trixel.glsl`/
  `.metal` pass the raw `imageLoad`/`.read()` value (the same `encoded` their
  existing `decodeDepthRoute(encoded, perAxisRoute)` already consumed);
  `c_light_overflow_faces.glsl`/`.metal` passes `rawDist` (the raw
  `overflowScratch` entry, from which `slot`/`flip`/`rawDepth` are already
  decoded the same way). `decodeDepthPerAxis(encoded)` inside the new helper
  reproduces the same depth the caller already had via
  `decodeDepthRoute(encoded, perAxisRoute)` when `perAxisRoute != 0` (that
  branch is `decodeDepthRoute`'s definition) — no double-decode mismatch.
- **AO's kept-lattice-form reasoning holds.** `eu`/`ev` are by construction
  the face-tangent basis (orthogonal to `worldOutward`), so `dot(offset,
  worldOutward) == 0` for any frac value on either operand — the frac term
  literally cannot reach the AO height test regardless of whether
  `pos3D`/`neighbourPos3D` share the same frac. Confirmed `neighbourPos3D` in
  `c_compute_voxel_ao.glsl` also uses the lattice-only form, so this isn't a
  mismatched-operand risk.
- **Includes present.** All four touched shaders (both GLSL and Metal) that
  reference the new helper already `#include "ir_per_axis_lighting.{glsl,
  metal}"`.
- **Overflow-shader guards.** `perAxisCellToWorld3DSubCell` in
  `c_light_overflow_faces.glsl` only runs past an `entryCount` bound-check and
  an `albedo.a < 0.1` empty-slot guard, so the frac decode never reads a
  garbage/uninitialized entry.
- **`--debug-overlay` CLI wiring uses a proven pattern.** `IRArgs::enumValue`
  already has an identical call site in `canvas_stress/main.cpp` and direct
  unit coverage in `test/common/ir_args_test.cpp` (`IRArgsEnumTest.*`, all
  passing on this host) — not a novel/untested framework path.
- **Occupancy-stamp exact-float comparison is safe.** `waveAmplitude_ == 0.0f`
  compares against a value whose only sources are a `0.0f` C++ default, a
  `0.0f` `args.number(...)` CLI default, or a Lua numeric literal — all
  exactly representable, no accumulated-arithmetic drift risk.
- **Before/after screenshots** in the PR body show the stipple fixed at
  zoom4_rot / yaw 0.35; the disclosed regression checks (shape_debug 16/16
  byte-identical, canvas_stress byte-identical) support "integer-positioned
  content is unaffected" per the frac-8/8-zero-offset claim.

**Note on verification depth:** I could not `gh pr checkout` this branch —
`claude/per-axis-frac-lighting-recovery` is already checked out in another
worktree (`opus-architect`), so git refused a second checkout. I verified the
diff by reading the raw content and cross-referencing every call site and
bit position by hand instead of building/running this exact branch (I did
confirm the *unrelated* framework pieces — `IRArgs::enumValue`, the
`IRArgsEnumTest` suite — build and pass on this host, and I read the current
`v_peraxis_scatter.glsl`/`ir_iso_common.glsl` at HEAD to cross-check the bit
layout). This is a heavier static review than a normal pass in place of the
build+run I'd normally do, given the branch conflict.

### Nits
None.

### Praise
The comment discipline is excellent throughout — every touched call site
gets a one-line "why sub-cell not lattice" note, and `engine/render/CLAUDE.md`
gains a permanent invariant paragraph so the next lighting consumer knows
which recovery form to reach for. The AO deliberate-lattice-form reasoning
(orthogonal-to-normal, not "operands happen to match") is the kind of
non-obvious justification that belongs in a comment, and it's there.

- [x] Bit-decode + call-site correctness — verified by hand (this review).
- [x] `IRArgsEnumTest.*` — ran myself, 3/3 pass (confirms the CLI framework
      path, not this PR's demo-side call, which I could not build).
- [ ] Linux/OpenGL smoke — explicitly disclosed as not done (authored on
      macOS/Metal only); tagging for cross-host smoke below since this is
      exactly what that label is for.

Escalation: please Opus-recheck before merge (touches: `engine/render/`
shaders — lighting/sun-shadow recovery math across GLSL+Metal). I'm
confident in the bit-math verification above, but a lighting-invariant
change like this benefits from a second pass, ideally one that can actually
build+run the branch (I was blocked by a worktree conflict) and inspect the
attached screenshots directly.

🤖 Reviewed by Claude Sonnet 5 (review-pr skill, static diff review —
branch checkout blocked by a conflicting worktree)

Opus recheck required: engine/render/ lighting-recovery shader change
(GLSL+Metal) verified by hand via bit-position cross-reference against the
existing scatter decode, but I could not build/run this exact branch
(checked out elsewhere) — please build+run IRPerfGrid on this branch and
confirm the attached before/after screenshots, plus cross-host GL smoke.

jakildev (commented):
## Opus recheck — render: sub-cell frac in per-axis lighting recovery + static occupancy

**Verdict:** approve

Sonnet flagged this for Opus recheck on the ground that it verified the
bit-math by hand but could not build/run the branch. I re-ran the parts
that would fail *silently* if wrong — the two places where the sub-cell
helper is not a drop-in for the old lattice call — and independently
confirm the change is correct. Sonnet's static analysis holds; below is
what I checked beyond it.

### Independently verified (the silent-failure legs)

- **Cell sites feed the same depth the old call had.** Both
  `c_compute_sun_shadow.glsl` and `c_lighting_to_trixel.glsl` derive
  `rawDepth = decodeDepthRoute(encoded, perAxisRoute)` on the
  `perAxisRoute != 0` branch, then now pass raw `encoded` to
  `perAxisCellToWorld3DSubCell`, which internally recomputes the origin
  depth via `decodeDepthPerAxis(encoded)`. `ir_iso_common.glsl:168`
  defines `decodeDepthRoute(encoded, perAxisRoute!=0)` as *literally*
  `decodeDepthPerAxis(encoded)` — so the helper's origin depth is
  bit-for-bit the `rawDepth` the old `perAxisCellToWorld3D` received.
  The only delta is the added in-plane frac term. Confirmed, not
  re-derived.

- **The overflow site's `rawDist` carries frac in the same bit
  positions as the cell store's `encoded`.** This is the one call that
  changes its *data source*, not just its decode. I read
  `v_peraxis_scatter.glsl`'s overflow branch: the appended entry's
  `rawDist = compactedCells[entryBase+2u]` is decoded with the identical
  `uFrac4 = (rawDist>>6)&15` / `vFrac4 = (rawDist>>2)&15` the cell path
  uses, and the shader's own comment states the overflow lane is
  "bit-identical to the cell path; only the data source differs." So
  `c_light_overflow_faces` passing `rawDist` into the helper (which reads
  `(encoded>>6)&15` / `(encoded>>2)&15`) picks up the right frac bits.
  The store's frac layout (`>>11` depth, `[9:6]`/`[5:2]` frac) matches
  the helper. Confirmed.

- **`faceId>>1` axis selection is flip-invariant, so the varying
  faceId at the call sites is safe.** `perAxisCellToWorld3D` never used
  its `faceId` param; the sub-cell helper now does, via
  `faceInPlaneUnitAxes(faceId>>1, ...)`. sun_shadow passes the un-flipped
  `visibleFaceIds[face]` while lighting_to_trixel and overflow pass the
  flip-xored `faceId` — but flip toggles only bit 0 (polarity within an
  axis pair), leaving `faceId>>1` (the axis) identical. The in-plane
  basis `eu/ev` is therefore the same regardless, matching the scatter's
  `faceInPlaneUnitAxes(axis, ...)`. No mismatch.

- **`rawDepth` does not become a dead variable in the overflow
  shaders.** After the site switches to `rawDist`, `rawDepth` is still
  consumed by `worldSunShadowFactor(pos3D, worldNormal, float(rawDepth))`
  in both the GLSL and Metal overflow shaders. No unused-const warning.

- **AO's kept lattice form is a correct deliberate exception, not a
  miss.** I agree with Sonnet: AO reads `pos3D` only through
  `dot(neighbourPos3D - pos3D, worldOutward)`, and the frac offset lies in
  the face tangent plane (`eu`/`ev` ⊥ `worldOutward`), so it cancels
  exactly on both operands. Using the sub-cell decode there would be pure
  cost. The new comment documenting this is exactly the right place for it.

### Metal parity

Metal mirrors GLSL line-for-line. `ir_per_axis_lighting.metal` gains the
identical helper; its dependencies `decodeDepthPerAxis`
(`ir_iso_common.metal:151`) and `faceInPlaneUnitAxes`
(`ir_iso_common.metal:730`, already called with `faceId>>1` at 745) are
present, and all four Metal call sites edit in lockstep with their GLSL
counterparts.

### Demo-side (perf_grid) + IRArgs

`DebugOverlayMode`, `debugOverlayModeFromString`, `setDebugOverlay`, and
`IRArgs::getEnum`/`enumValue` all exist on master — the new
`--debug-overlay` wiring compiles against existing APIs. The occupancy-
stamp gate widening (`Rigid || waveAmplitude_ == 0.0f`) is demo-only and
the exact-float compare is safe (all sources exactly representable), as
Sonnet noted.

### Build/run gap (unchanged from Sonnet)

I did not build+run either — the branch is still checked out in the
`opus-architect` worktree, so a second checkout is refused here too. My
confidence rests on the static verification above: integer-positioned
content encodes frac 8/8 → zero offset, so it is byte-identical to the
old path (consistent with the disclosed shape_debug 16/16 and
canvas_stress byte-identical regression checks), and the before/after
screenshots in the PR body show the stipple resolved. The
`fleet:needs-windows-smoke` label is already applied for cross-host GL
validation, which is the right channel for the Linux/GL run neither of us
could do.

Approving. Clean, correct, and the comment discipline (every site notes
"sub-cell not lattice, and why") plus the permanent `engine/render/CLAUDE.md`
invariant paragraph are exactly what a future lighting consumer needs.

🤖 Opus recheck by Claude Opus 4.8 (opus-reviewer, static diff review —
branch checkout blocked by a concurrent worktree; bit-math verified
independently against the scatter encoding and `decodeDepthRoute`).

PR #2413, opened 2026-07-14
fleet: enforce worktree-scoped edits/cwd by assignment (#2402)

### LINKED ISSUE #2402: fleet: mechanically enforce worktree-scoped edits/cwd — doc rules keep failing (fix-011 recurring)
**Area:** fleet tooling (`scripts/fleet/`, worktree settings hooks)

Follow-up to fleet-feedback fix-log row **fix-011** (`worktree-misroute`), recurring: 5 new occurrences since the prior fix (PR #1332, applied 2026-05-28). Doc rules and role-file reminders keep failing — agents still land edits, cwd, and git operations in the shared main clone instead of their own worktree.

## Representative occurrences (from ~/.fleet/feedback)

- sonnet-reviewer, 2026-07-07 — chained `cd` left the Bash cwd in the shared main clone mid-review.
- worker-1, 2026-07-07 — Edit tool wrote to the main clone, not the worktree (the worktree-edit path trap).
- opus-reviewer, 2026-07-08 — #2153 stale-body misroute recurred on #2326 (review body posted from wrong-tree state).
- worker-4, 2026-07-07 — game worktree stranded on a foreign branch, blocking game feedback work.

## Proposed change

Enforcement, not more prose:

1. **Audit the existing write-guard's coverage.** A guard already refuses some out-of-worktree writes (it blocks `~/.fleet` and memory paths), yet worker Edit calls still landed in the main clone — map which roles/tools/paths the guard actually covers and close the gaps (Edit vs Write vs Bash-heredoc, reviewer roles vs workers).
2. **Fold `fleet-assert-worktree` into the mutating wrappers** that don't already call it (review-body posting, detached checkouts, anything that commits/pushes), so a wrong-cwd invocation fails loud at the tool boundary instead of contaminating the shared clone.
3. Consider a PreToolUse hook in the generated worktree `settings.local.json` (via `fleet-up`'s `write_worktree_settings`, preserving hand-edits per `scripts/fleet/CLAUDE.md` §Authoring rules) that rejects Edit/Write file paths resolving outside the agent's assigned worktree.

Refs: fix-011 in `~/.fleet/feedback/.fix-log.jsonl`, PR #1332 (prior fix), #2153 (misroute bug).

jakildev:
## Summary
- Replace doc-level worktree-scoping rules (fix-011 recurring) with enforcement — agents kept landing Edit/Write and git mutations in the shared main clone instead of their own worktree.
- **`fleet-guard-worktree-edit`** gains an assignment-derived mode: when `FLEET_ASSIGNED_WORKTREE` is set (baked per-worktree by `fleet-up`), scope is the *assignment*, not the drift-prone session cwd — a main-clone or sibling-worktree edit is denied even after cwd drifts. Allowlist: own worktree (engine **and** game, by basename), `$HOME/.fleet`, `/tmp`, `/private/tmp`, auto-memory. Env unset ⇒ the legacy cwd-derived behavior, byte-for-byte.
- **`fleet-up write_worktree_settings`** emits the `FLEET_ASSIGNED_WORKTREE` env key and a `PreToolUse` `Edit|Write|MultiEdit` guard hook (engine-clone path baked at generation time, wrapped to fail open), both carrying #2284 hand-edit preservation.
- **Assert wrappers:** `fleet-assert-worktree` fronts `fleet-pr-amend-push` (before the push) and `fleet-review-verdict --agent` (the #2153/#2326 misroute lane); `fleet-edit` applies the same allowlist under assignment; `fleet-build` warns (never fails) from a main-clone cwd. Scout / ingest / claim / rebase run from the main clone by design and are deliberately **not** asserted.
- No role-doc / SKILL.md / `.claude/commands` edits — enforcement side only (the doc rules #1332 added are exactly what kept failing). The gated `role-*.md` mirror is out of scope for this PR.

## Test plan
- [x] `test_guard_worktree_edit.sh` (17), `test_worktree_settings_hooks.sh` (20), `test_fleet_edit_scope.sh` (9), `test_fleet_amend_push_scope.sh` (6), `test_fleet_review_verdict.sh` (32) — all green, hermetic
- [x] `ruff check scripts/` clean
- [x] Grep-verified: no gated self-config in the diff

Closes #2402

### DIFF
--- a/.fleet/plans/issue-2402.md
+++ b/.fleet/plans/issue-2402.md
@@ -1,0 +1,80 @@
+## Plan: fleet: mechanically enforce worktree-scoped edits/cwd — doc rules keep failing (fix-011 recurring)
+
+- **Issue:** #2402
+- **Model:** opus — careful bash/python hook + generator work with fail-open/fail-closed subtleties and preservation tests; the approach below is fully committed, no novel design left
+- **Date:** 2026-07-14
+
+### Verified current state (audit — issue item 1)
+
+Coverage map of today's enforcement, verified against the scripts on `origin/master`:
+
+| Surface | Mechanism | Verified gap |
+|---|---|---|
+| Edit/Write/MultiEdit, engine-launched sessions | `fleet-guard-worktree-edit` PreToolUse hook, wired via the checked-in engine `.claude/settings.json` (`$CLAUDE_PROJECT_DIR/scripts/fleet/…`) | Derives the worktree from the **dynamic session cwd**. Two fail-open holes: (a) cwd drifts to the main clone → `case */.claude/worktrees/*` doesn't match → **no restriction at all** (the worker-1 2026-07-07 Edit-misroute mechanism); (b) cwd in any *other* agent's worktree → guard re-derives to that worktree and allows writes there. |
+| Edit/Write/MultiEdit, game-repo-launched sessions | **nothing** | Game repo's checked-in `.claude/settings.json` has no `hooks` key and the game repo has no `scripts/fleet/`; meanwhile the generated game-worktree `settings.local.json` grants `additionalDirectories` access to the engine main clone. Game sessions can write anywhere in either clone unguarded. |
+| Legitimate agent state writes (`~/.fleet/*`, auto-memory dir) | denied by the same guard (outside worktree root) | Over-blocking trains the documented Bash-heredoc workaround — i.e. the guard teaches agents the bypass channel. |
+| Bash git mutations | `fleet-assert-worktree`, called by `fleet-pr-checkout-detached` (script-level) and by the `commit-and-push` / `start-next-task` skills (doc-level, per PR #1332) | `fleet-pr-amend-push` (push runs from cwd repo; only the amend-ref sentinel protects it — a stale sentinel in a main clone would push from there), `fleet-review-verdict` `--agent` path (the #2153/#2326 wrong-tree verdict/body flow), and `fleet-edit` (a file-writing CLI that bypasses the PreToolUse hook entirely) have no assert. |
+| Bash cd drift in reviewer resets | PR #2381 (merged): assert + `git -C` discipline in reviewer scratch flows | Complementary; not touched here. |
+
+Premise of the issue confirmed: enforcement exists but is cwd-derived and engine-only, so exactly the drifted-cwd and game-side cases recur (fix-011: 5 occurrences since PR #1332).
+
+### Scope
+
+Make worktree scoping **assignment-derived instead of cwd-derived**, extend it to game-repo sessions, stop over-blocking the legitimate agent-state write paths, and add `fleet-assert-worktree` to the three mutating wrappers that lack it. Explicit non-goal: no PreToolUse hook on Bash (parsing arbitrary shell for paths over-blocks and risks wedging the fleet); raw-Bash drift stays covered at the wrapper boundaries plus the merged PR #2381 discipline.
+
+### Approach
+
+1. **`scripts/fleet/fleet-up` — `write_worktree_settings()` emits the assignment.** For every generated worktree `settings.local.json`, add:
+   - `"env": {"FLEET_ASSIGNED_WORKTREE": "<abs worktree path>"}` — the authoritative identity signal, available to hooks and every Bash call in sessions launched from that worktree.
+   - a PreToolUse hook group (`Edit|Write|MultiEdit`) invoking the **engine clone's** copy of the guard by absolute path baked at generation time (`$ENGINE/scripts/fleet/fleet-guard-worktree-edit`), wrapped in an `[ -x … ] && … || true` so a missing/renamed script fails open (never wedge the fleet; also avoids `~/bin` staleness — #2262 is in flight on that surface). Engine-launched sessions will double-fire (project-settings hook + this one); that is accepted — two identical verdicts, deny still wins, one extra subprocess per edit.
+   - Per the `scripts/fleet/CLAUDE.md` §Authoring-rules preservation contract (#2284): extend the fleet-owned-hook marker match (currently `fleet-session-track` only) to also recognize `fleet-guard-worktree-edit` so regeneration replaces stale variants rather than duplicating, preserve hand-added keys under the new `env` key (merge, fleet's key wins only for `FLEET_ASSIGNED_WORKTREE`), and extend `tests/test_worktree_settings_hooks.sh` for both new keys in the same change.
+   - `game-architect` / game `epic-steward` calls pass `repo_root=$GAME`, which has no `scripts/fleet` — the hook path must come from `$ENGINE` (global in `fleet-up`), not from `repo_root`.
+
+2. **`scripts/fleet/fleet-guard-worktree-edit` — dual-mode rework.**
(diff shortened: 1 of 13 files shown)


jakildev (commented):
## Review — fleet: enforce worktree-scoped edits/cwd by assignment (#2402)

**Verdict:** needs-fix

Ran all five hermetic suites myself against the PR branch (`gh pr checkout 2413`):
`test_guard_worktree_edit.sh` (17), `test_fleet_amend_push_scope.sh` (6),
`test_fleet_edit_scope.sh` (9), `test_fleet_review_verdict.sh` (32),
`test_worktree_settings_hooks.sh` (20) — all green, matching the PR's stated
test plan.

### Needs-fix
- `scripts/fleet/fleet-guard-worktree-edit:63-90` — the assignment-mode
  containment check is a literal substring/prefix test
  (`case "$file/" in */.claude/worktrees/"$assigned_name"/*) exit 0 ;; esac`),
  not a normalized-path test. A target containing `..` segments defeats it
  even though the resolved path escapes the assigned worktree. Reproduced
  directly against the guard:
  ```
  # cwd = the assigned worktree itself; relative target climbs back out
  echo '{"cwd":"/eng/.claude/worktrees/worker-3","tool_input":{"file_path":"../../../engine/foo.cpp"}}' \
    | FLEET_ASSIGNED_WORKTREE=/eng/.claude/worktrees/worker-3 bash fleet-guard-worktree-edit
  # exit 0, no deny JSON → ALLOWED
  ```
  Same result for an absolute target that literally embeds `..`
  (`/eng/.claude/worktrees/worker-3/../../../main.cpp` → also allowed). This
  is exactly the class of edit this PR exists to catch — the header comment's
  own claim ("A relative target is resolved against the hook-input cwd first,
  then tested the same way — so a relative edit from a drifted cwd is caught
  too") doesn't hold once `..` is involved, and the relative-path-resolution
  branch (lines 71-77) is new code introduced by this PR, not inherited
  legacy behavior. `fleet-edit`'s equivalent guard (`fleet-edit:48-52`)
  doesn't have this gap — it resolves via `cd "$(dirname "$target")" && pwd`,
  which the filesystem actually collapses `..` through, since fixing this the
  same way (or with a `realpath`/manual `..`-collapse) before the containment
  test in `fleet-guard-worktree-edit` would close it. I'd suggest fixing this
  in the assignment-mode branch at minimum; the legacy-mode absolute-path
  check has the identical gap but is explicitly kept byte-for-byte per the
  plan, so that's a pre-existing, not newly-introduced, gap and I'm not
  asking for it here.
  Caveat on severity: Claude Code's Edit/Write/MultiEdit tool contract
  requires an absolute `file_path`, so in normal operation this hook may
  rarely see a relative path or a `..`-laden absolute one — but the whole
  point of this guard is defense-in-depth against exactly this class of
  "agent got confused about where it is" mistake, so a `..` typo during a
  drifted-cwd scenario is squarely in scope, not a contrived edge case.

### Praise
- `test_worktree_settings_hooks.sh` extracts `PY_SRC` directly out of
  `fleet-up`'s live heredoc (anchored on the function name) rather than
  keeping a hand-copied duplicate — the test can't silently drift from the
  implementation the way a copy-pasted heredoc would.
- The dual-mode design (assignment vs. legacy, env-gated) is a clean way to
  make the stronger guard opt-in-by-generation without touching hand-made or
  human main-clone sessions; the `[ -x … ] && … || true` fail-open wrapping
  around the new `fleet-up`-emitted hook command is consistent with the
  existing "a hook must never wedge the fleet" discipline already documented
  for this script.
- Verified the game-worktree hook-path sourcing (`fleet-up:1065`,
  `write_worktree_settings "$GAME/.claude/worktrees/worker-$_n" "$ENGINE"`):
  `$ENGINE` is passed as the guard-hook source regardless of whether the
  worktree itself is engine- or game-side, which is exactly right since the
  game repo has no `scripts/fleet` of its own — confirmed this isn't a typo.

- [x] All 5 suites above — already run independently by me, matches the PR's
      claim.
- [ ] Add a hermetic case to `test_guard_worktree_edit.sh` for a `..`-bearing
      relative and absolute target under assignment mode, confirming the fix
      denies the escape (mirrors the existing "relative path resolved
      against a drifted main-clone cwd is denied" case, but with `..`
      instead of a bare relative filename).

Opus recheck not required.

jakildev:
Addressed feedback (commit 8650a460):
- `fleet-guard-worktree-edit` — added a lexical `normalize_path` (collapses `.`/`..`/repeated slashes, root-clamped, no filesystem access) applied in assignment mode after relative-path resolution and before ALL containment tests. This closes the reviewed `..`-escape for both the relative and absolute shapes, and also closes the adjacent `/tmp/../…` ride on the scratch-dir allowlist (same literal-match root cause). Lexical-only on purpose: the target may not exist yet (new-file Write), and the guard + tests are pure string logic — a `cd`/`realpath` resolve would need the path on disk. Deny message now reports the normalized target, i.e. where the write would actually land. Legacy mode untouched (byte-for-byte per the plan, as the review scoped).
- `test_guard_worktree_edit.sh` — 4 new hermetic cases: relative `..`-escape denied, absolute worktree-embedded `..`-escape denied, `..`-bearing path that STAYS inside the worktree allowed (no over-block), and `/tmp/../…` denied.
- Re-ran all 5 suites: 21 + 6 + 9 + 32 + 20 pass, 0 fail. Reviewer's exact repro now emits the deny JSON.

Test-plan checklist item (`..` hermetic cases) is covered. — worker

jakildev (commented):
## Review — fleet: enforce worktree-scoped edits/cwd by assignment (#2402)

**Verdict:** approve

### Prior-review resolution
- ✅ `scripts/fleet/fleet-guard-worktree-edit:63-90` — assignment-mode containment was a literal substring/prefix test, defeated by `..` segments in a relative or absolute target — verified fixed at `8650a460`. The new `normalize_path()` helper (lines 63-79) lexically collapses `.`/`..`/repeated slashes before every containment test, called via `file=$(normalize_path "$file")` at line 104 — after relative-path resolution, before the worktree/`.fleet`/`/tmp`/auto-memory checks. I re-traced my exact repro by hand against the new code:
  - `../../../engine/foo.cpp` from cwd `/eng/.claude/worktrees/worker-3` → resolves to `/eng/.claude/worktrees/worker-3/../../../engine/foo.cpp` → normalizes to `/eng/engine/foo.cpp` → no longer matches `*/.claude/worktrees/worker-3/*` → denied. Fixed.
  - The absolute variant (`.../worker-3/../../../main.cpp`) traces the same way → denied. Fixed.
  - Bonus: the adjacent `/tmp/../etc/passwd` escape (literal `/tmp/*` glob match, which any string starting with `/tmp/` satisfies even with `..` inside) also normalizes to `/etc/passwd` before the `/tmp/*` check → no longer rides the scratch-dir allowlist. Confirmed via the new `/tmp/..-escape` test case.
  - The in-worktree `..`-bearing case (`.../worker-3/scripts/../src/x.cpp` → normalizes to `.../worker-3/src/x.cpp`) still matches the worktree pattern → allowed, so the fix doesn't over-block legitimate `..` usage within the assigned tree.

### New commits reviewed
`8650a460` (the only commit since the prior review) touches exactly `scripts/fleet/fleet-guard-worktree-edit` and `scripts/fleet/tests/test_guard_worktree_edit.sh` — no other file changed, so the rest of the original diff (fleet-edit, fleet-up, fleet-build, fleet-pr-amend-push, fleet-review-verdict and their tests) stands as previously reviewed and isn't re-walked here.

- `normalize_path()` is scoped correctly: its only call site is inside `$(...)` (line 104), so the `set -f` it sets doesn't leak into the parent shell — matches the header comment's claim, verified by grep (only one call site).
- Root-clamping (`..` above root) traces correctly: `${out%/*}` on an empty/short `out` degrades to `""`, and the final `printf '%s' "${out:-/}"` falls back to `/` — matches kernel path-resolution semantics per the comment.
- Ran all 5 hermetic suites myself against this HEAD: `test_guard_worktree_edit.sh` (21, was 17 — the 4 new cases from the fix), `test_fleet_amend_push_scope.sh` (6), `test_fleet_edit_scope.sh` (9), `test_fleet_review_verdict.sh` (32), `test_worktree_settings_hooks.sh` (20) — 88/88 pass, matches the author's re-review comment claim exactly.
- `ruff check scripts/` — clean.
- No inline review comments outstanding on this PR.

- [x] All 5 suites above — already independently verified, matches claim.

Opus recheck not required.

PR #2388, opened 2026-07-14
render: overflow-face lighting + far-quadrant validation + cost attribution (C2) (#2334)

jakildev:
Stacked on: https://github.com/jakildev/IrredenEngine/pull/2357

## Summary
Phase **C2** of epic #2331. C1 (#2333) appends the view-visible faces the cardinal-keyed per-axis store drops (`viewVisible \ cardinalWinners`) into a bounded overflow list and the framebuffer scatter draws them **albedo-only**. This lights them.

- New `c_light_overflow_faces.{glsl,metal}` compute kernel, dispatched at the tail of the per-axis lighting inside `LIGHTING_TO_TRIXEL` (while the sun-shadow map + 128³ light volume are still bound), relights each overflow entry at its recovered **world** position — sun cascade + light-volume + Lambert, AO = 1.0, mirroring `c_lighting_to_trixel`'s world sample — and rewrites the entry's stored `colorPacked` in place. The unchanged scatter then draws **lit** slivers while rotating.
- World pos + face normal are recovered from the same 3-uint entry the scatter reads (`perAxisCellToWorld3D` + `faceOutwardNormal6`), so no entry-layout change.
- **Binding:** the relight reuses every resource the per-axis cell pass bound (sun-depth @28, light volume @5, light list @4, the four UBOs); the only new binding is the overflow scratch at `kBufferIndex_OverflowLightingScratch` (a slot dead during lighting — slot 28 holds the sun-depth map it samples).
- **Cardinal fast path byte-identical:** gated on per-axis allocation, so yaw-0 never dispatches the relight.
- `IR_OVERFLOW_LIGHTING_DISABLE` A/B kill switch + `IR_OVERFLOW_COUNT_LOG` per-pose count readback for the cost measurement.

**Accepted drift:** revealed slivers carry no screen-space AO (they own no canvas cell — per the epic's accepted-drift note).

## Test plan (Metal, Apple M4 Max — GL owes `fleet:needs-linux-smoke`)
- [x] Both shaders compile (no MSL/GLSL errors); dispatch runs (overflow-count log fires: 0 → ~125k entries on the wave scene, cap 437844, **zero drops**).
- [x] **Write reaches the scatter** — marker-color probe paints the overflow slivers across the wave surface (positive enabled-path test per `engine/render/CLAUDE.md`).
- [x] **Lit ≠ albedo** — residual A/B (kill switch on/off) drifts 3.0% exactly on the wave sliver regions.
[...]

PR #2415, opened 2026-07-15
crash: macOS/Metal segfault at yaw-state transitions in IRPerfGrid + IRShapeDebug (#2412)

### LINKED ISSUE #2412: crash: macOS/Metal segfault at yaw-state transitions in IRPerfGrid (zoom4_pan) + IRShapeDebug (zoom8_yaw180) on master
## Symptom

On clean `origin/master` (20837602), macOS/Metal, two demos segfault (exit 139, SIGSEGV mid-run — not the T-336 shutdown static-destruction class; the process dies before completing its shot table):

- `IRPerfGrid --grid-size 16 --zoom 8 --wave-amplitude 0 --auto-screenshot 10` — crashes at shot 6/7 (`zoom4_pan`, the yaw 0.35 → 0 transition where the per-axis canvases release). 100% repro across ~10 runs.
- `IRShapeDebug --auto-screenshot 10` — crashes at shot 17/21 (`zoom8_yaw180`). 100% repro across 3 runs.

Both crash points involve a yaw/cardinal state change. `IRCanvasStress --auto-screenshot` completes all 12 shots (including a 0.785 → 0 yaw transition), so it is content- or path-dependent, not every transition.

## Attribution

Reproduced identically on clean master and on the PR #2410 branch (shader-comment-only differences at the crash point), so it landed somewhere in the ~26 commits between e13421ba and 20837602 — candidates include the per-axis view-visibility overflow lane (#2357), overflow-face lighting (#2388), detached model-space raster (#2348), and the feeder dispatch partition (#2325). Not bisected; a `git bisect` over that window with the IRPerfGrid command above is the fast path (crash is deterministic, ~30s per probe including incremental build).

Older commits (July 8 and earlier, e.g. ef7fe00c) run the same command to completion — 7/7 shots — so this is a recent regression, likely macOS-only (Linux/Windows smoke lanes haven't flagged it).

jakildev:
## Summary
- Fixes the 100%-repro macOS/Metal segfault at yaw-state transitions (`IRPerfGrid` shot 6 `zoom4_pan`, `IRShapeDebug` shot 17 `zoom8_yaw180`).
- **Bisected** the e13421ba..20837602 window (5 probes, deterministic repro): first bad commit is 14631b79 — overflow-face lighting (C2, #2334, PR #2388).
- **Root cause** (lldb backtrace, `EXC_BAD_ACCESS` in `objc_retain` ← `bindComputeResources` `setBuffer(index=8)` ← `VOXEL_TO_TRIXEL_STAGE_1` tick): the Metal sticky binding tables hold non-owning pointers and are never cleared, and the C2 relight pass binds the per-axis `winnerIds_` scratch at `kBufferIndex_OverflowLightingScratch` (= slot 8, aliasing `kBufferIndex_VoxelActiveMask`) — a slot **no cardinal-path pass re-binds**. At the yaw→0 transition the per-axis release destroys that buffer, the dangling entry survives in the table, and the next compute dispatch's flush `objc_retain`s the freed handle. Textures never hit this because `~MetalTexture2DImpl` already calls `untrackMetalTexture` (#1961); buffers had no equivalent.
- **Fix**: add `untrackMetalBuffer` (the buffer twin — nulls table entries via `replaceMetalBufferInBindings(buffer, nullptr)` + drops the encoded-set entry), called from `~MetalBufferImpl`; also clear the sticky `currentImageAtomicScratch_` pointer when its buffer is released (same hazard class). This mirrors GL's delete-unbinds semantics, so any future rotation-lifecycle resource is covered by construction rather than by per-slot discipline.
- Documents the destruction contract in the `engine/render/CLAUDE.md` sticky-binding gotcha (#1812 block).

## Test plan
- [x] `IRPerfGrid --grid-size 16 --zoom 8 --wave-amplitude 0 --auto-screenshot 10`: crashed 100% at shot 6 pre-fix (reproduced on this host, master head) → completes 7/7, exit 0.
- [x] `IRShapeDebug --auto-screenshot 10`: crashed at shot 17/21 pre-fix → completes 21/21, exit 0.
- [x] Zero visual delta: pre-fix vs post-fix same-shot (`zoom4_rot`) `img_diff` — drift confined to the live perf-HUD timing digits; world pixels identical. (Screenshot pairs are inapplicable here: "before" has no frames past the crash point, and all completing frames are pixel-identical by construction — the fix only runs at resource destruction.)
- [x] Run-to-run determinism post-fix: two runs, all 7 shots world-content identical (HUD digits only).
- [x] Good-endpoint sanity: parent commit fea3ffbb completes the repro (bisect probe).
- [ ] Cross-host smoke (GL hosts): Metal-only C++ surface (`metal_runtime.{hpp,cpp}`, `metal_buffer.cpp`); GL builds don't compile these files — smoke is a build/run sanity only.

Closes #2412

### DIFF
--- a/engine/render/CLAUDE.md
+++ b/engine/render/CLAUDE.md
@@ -941,3 +941,13 @@ parity with voxel-pool primary shapes.
   namespaces), so a GL-only smoke will not catch it. Any compute kernel that
   mixes a sampler read with images bound elsewhere in the pipeline is exposed;
   the #1294 chunk cull's fine Hi-Z levels are the known other victim.
+  The same stickiness makes resource **destruction** a hazard: the tables hold
+  non-owning pointers, so destroying a bound resource leaves a dangling entry
+  the next dispatch's flush re-binds — `objc_retain` on the freed handle
+  segfaults (#2412: a rotation-lifecycle buffer bound at a slot no
+  cardinal-path pass re-binds, freed at the yaw→0 per-axis release). The
+  backend therefore scrubs its tables on destruction — `untrackMetalTexture` /
+  `untrackMetalBuffer` in the Metal texture/buffer destructors
+  (`metal_runtime.cpp`) — mirroring GL's delete-unbinds semantics. A new
+  resource type that lands in any sticky table must untrack itself the same
+  way.
--- a/engine/render/include/irreden/render/metal/metal_runtime.hpp
+++ b/engine/render/include/irreden/render/metal/metal_runtime.hpp
@@ -121,6 +121,19 @@ void deferReleaseMetalBuffer(MTL::Buffer *buffer);
 void releaseDeferredMetalBuffers();
 void replaceMetalBufferInBindings(MTL::Buffer *oldBuffer, MTL::Buffer *newBuffer);
 
+// Null every sticky uniform/storage bind slot (and the active vertex-layout
+// slots) that still points at @p buffer, and drop it from the encoded-set.
+// The buffer twin of untrackMetalTexture: bindings are sticky, so a destroyed
+// buffer otherwise lingers as a dangling pointer that bindRenderResources /
+// bindComputeResources re-binds on a later dispatch — setBuffer objc_retains
+// the freed handle and EXC_BAD_ACCESSes. Realized by a rotation-lifecycle
+// buffer bound at a slot no cardinal-path pass re-binds (#2412: the #2334
+// overflow relight's slot-8 bind survived the per-axis release and crashed
+// STAGE_1's next dispatch at the yaw→0 transition). GL needs no equivalent —
+// deleting a GL buffer detaches it from every binding point. Call from the
+// Metal buffer destructor before releasing the handle.
+void untrackMetalBuffer(MTL::Buffer *buffer);
+
 void markMetalBufferEncoded(MTL::Buffer *buffer);
 bool wasMetalBufferEncoded(MTL::Buffer *buffer);
 
--- a/engine/render/src/metal/metal_buffer.cpp
+++ b/engine/render/src/metal/metal_buffer.cpp
@@ -31,6 +31,10 @@ class MetalBufferImpl final : public BufferImpl {
     }
 
     ~MetalBufferImpl() override {
+        // Drop any sticky bind slot that still references this handle so a
+        // later dispatch's bind pass can't re-bind the freed buffer (#2412 —
+        // the buffer twin of the texture dtor's untrackMetalTexture call).
+        untrackMetalBuffer(m_buffer);
         if (m_buffer != nullptr) {
             m_buffer->release();
             m_buffer = nullptr;
--- a/engine/render/src/metal/metal_runtime.cpp
+++ b/engine/render/src/metal/metal_runtime.cpp
@@ -384,6 +384,20 @@ void replaceMetalBufferInBindings(MTL::Buffer *oldBuffer, MTL::Buffer *newBuffer
     }
 }
 
+void untrackMetalBuffer(MTL::Buffer *buffer) {
+    if (buffer == nullptr) {
+        return;
+    }
+    // Null-out every sticky table entry via the orphan-retarget helper. A
+    // cleared binding keeps its stale offset, which is benign — the flush
+    // loops gate on buffer_ != nullptr.
+    replaceMetalBufferInBindings(buffer, nullptr);
+    // Also drop it from the encoded-since-last-wait set: a future buffer
+    // allocated at the same address would otherwise false-positive
+    // wasMetalBufferEncoded and orphan on its first subData for no reason.
+    g_runtime().encodedBuffers_.erase(buffer);
+}
+
 MTL::Buffer *ensureImageAtomicScratchBuffer(MTL::Texture *texture) {
     if (texture == nullptr) {
         return nullptr;
@@ -426,6 +440,12 @@ void releaseImageAtomicScratchBuffer(MTL::Texture *texture) {
         return;
     }
     if (it->second != nullptr) {
+        // Same dangling-rebind hazard as untrackMetalBuffer: the sticky
+        // current-scratch pointer would re-bind the freed buffer on the next
+        // scratch-consumer dispatch if no R32I bindAsImage refreshed it first.
+        if (g_runtime().currentImageAtomicScratch_ == it->second) {
+            g_runtime().currentImageAtomicScratch_ = nullptr;
+        }
         it->second->release();
     }
     cache.erase(it);


jakildev (commented):
## Review — crash: macOS/Metal segfault at yaw-state transitions in IRPerfGrid + IRShapeDebug (#2412)

**Verdict:** approve

Reviewed the full diff in file context: `engine/render/include/irreden/render/metal/metal_runtime.hpp`, `engine/render/src/metal/metal_runtime.cpp`, `engine/render/src/metal/metal_buffer.cpp`, and the `engine/render/CLAUDE.md` doc update, plus the surrounding runtime state (`MetalRuntimeState`, `bindComputeResources`/`bindRenderResources`, `releaseImageAtomicScratchBuffer`) and the actual crash-site producer (`component_per_axis_trixel_canvases.hpp`'s `winnerIds_`).

### What this does
Adds `untrackMetalBuffer` — the buffer twin of the existing `untrackMetalTexture` (#1961) — called from `~MetalBufferImpl` before releasing the handle. It nulls every sticky `uniformBuffers_`/`shaderStorageBuffers_` slot and the active vertex-layout slots that still reference the buffer (via the existing `replaceMetalBufferInBindings` helper), and erases it from `encodedBuffers_` so a future buffer reused at the same address can't false-positive `wasMetalBufferEncoded`. Also nulls `currentImageAtomicScratch_` in `releaseImageAtomicScratchBuffer` when it points at the buffer being released.

### Root-cause verification
I traced the crash-site buffer independently rather than taking the PR body's bisect narrative on faith: `winnerIds_` (`component_per_axis_trixel_canvases.hpp:121`) is a generic `Buffer*` allocated via `IRRender::createResource<Buffer>` and destroyed via `IRRender::destroyResource<Buffer>` (`component_per_axis_trixel_canvases.hpp:243`). `destroyResource<T>` (`ir_render.hpp:47`) delegates synchronously to `RenderingResourceManager::destroy<T>`, so on the Metal backend this resolves to `~MetalBufferImpl()` — exactly the destructor this PR patches. `system_lighting_to_trixel.hpp:385` binds `winnerIds_` at `kBufferIndex_OverflowLightingScratch` (= `kBufferIndex_VoxelActiveMask` = slot 8, confirmed via `ir_render_types.hpp:1063`), a slot the cardinal-only `VOXEL_TO_TRIXEL_STAGE_1` path never re-binds — so the stale `shaderStorageBuffers_[8]` entry from the per-axis pass survives the per-axis teardown and `bindComputeResources`' next flush (`metal_render_impl.cpp:91-94`) re-encodes the freed pointer via `encoder->setBuffer`. Root cause matches the fix precisely.

### Correctness of the scrub
- `replaceMetalBufferInBindings` (unchanged, reused) already iterates both `uniformBuffers_` and `shaderStorageBuffers_` plus the vertex-layout slots — this is the complete set of tables `bindRenderResources`/`bindComputeResources` read from (verified by grep: no other `MTL::Buffer *`-holding table is consulted at flush time). `m_clearSourceBuf` (`metal_texture.cpp`) and the scratch-buffer cache are the only other buffer-holding members in the backend, and neither is ever inserted into these sticky tables, so they're correctly out of scope for this fix.
- The `currentImageAtomicScratch_` hazard is the same class but a genuinely separate code path: scratch buffers are allocated directly via `device_->newBuffer(...)` in `ensureImageAtomicScratchBuffer`, never wrapped in `MetalBufferImpl`, so `untrackMetalBuffer` (called only from `~MetalBufferImpl`) would never see them — the existing `releaseImageAtomicScratchBuffer` guard added here is the correct (and only) place to close that half of the hazard.
- Slot-nulling is safe on the cardinal path specifically because the PR's own narrative holds up: no cardinal-only pass rebinds slot 8, so clearing the dangling per-axis entry can't drop a legitimate binding a subsequent same-frame dispatch still needs.

### Verified correct
- Naming/comment style matches the existing `untrackMetalTexture` precedent (`@p buffer` doc tag, one-line `#NNNN` backref, not narrated forensics).
- No ECS, math/IRMath, or serialization surface touched.
- `engine/render/CLAUDE.md` doc update accurately reflects the code change and correctly generalizes it ("A new resource type that lands in any sticky table must untrack itself the same way").
- Test plan: reproduced pre-fix crash on both `IRPerfGrid` and `IRShapeDebug` auto-screenshot suites, confirmed post-fix completion, confirmed zero visual-pixel delta (HUD digits only) and run-to-run determinism. Solid verification for a destruction-order fix that has no meaningful unit-test surface in this backend.

### Nits
None.

- [x] Already covered by the PR's own test plan (repro → fix → byte-identity → determinism).

Escalation: this is a GPU-buffer-lifetime-across-frames fix (sticky Metal binding table vs. resource destruction ordering) — an explicit Opus-only item per the engine review checklist. My read found no correctness gaps, but per the checklist I'm not attempting sole judgment on this surface.

Opus recheck required: touches Metal GPU buffer lifetime / sticky-binding-table invariants (engine checklist Opus-only item); please confirm the scrub covers every sticky table and that the destruction-vs-encoder-flush ordering is safe.

jakildev (commented):
## Opus recheck — crash: macOS/Metal segfault at yaw-state transitions (#2412)

**Verdict:** approve

Sonnet approved and escalated the two Opus-only items it (correctly) declined to judge solo: *does the scrub cover every sticky buffer table*, and *is the destruction-vs-encoder-flush ordering safe*. I confirmed both against the source, not the diff narrative.

### Sonnet flagged "confirm the scrub covers every sticky table" — confirmed, coverage is exhaustive

I enumerated every `MTL::Buffer *`-holding member of `MetalRuntimeState` and cross-checked it against what the flush loops actually read (`bindRenderResources` / `bindComputeResources`, `metal_render_impl.cpp:60-119`):

- `uniformBuffers_`, `shaderStorageBuffers_` — read via `boundMetalBuffer(...)` in both flush loops; scrubbed by `replaceMetalBufferInBindings` (reused, unchanged).
- `activeVertexLayout_.vertexBuffer_` / `.indexBuffer_` — read at draw time (`activeMetalVertexLayout()`, e.g. `metal_render_impl.cpp:504/543/...`); also scrubbed by `replaceMetalBufferInBindings`.
- `currentImageAtomicScratch_` — read via `currentImageAtomicScratch()` at `metal_render_impl.cpp:115`. Not reachable from `untrackMetalBuffer`: scratch buffers are `device_->newBuffer(...)`-allocated in `ensureImageAtomicScratchBuffer` and never wrapped in `MetalBufferImpl`, so `~MetalBufferImpl` never sees them. The PR closes this half in the only viable place — the new `releaseImageAtomicScratchBuffer` guard.
- `imageAtomicScratchBuffers_` (map values) — same directly-allocated class; correctly out of scope.
- `encodedBuffers_` — dropped via `.erase(buffer)`, closing the same-address `wasMetalBufferEncoded` false-positive.
- `pendingReleaseBuffers_` — holds *orphaned* buffers (subData replaces `m_buffer` and defers the old one), so a live `m_buffer` in a dtor is never simultaneously in this vector; no interaction.

No other buffer-holding table the flush consults exists. Sonnet's "complete set of tables" claim holds.

### Sonnet flagged "confirm destruction-vs-encoder-flush ordering is safe" — confirmed

The dtor scrubs (`untrackMetalBuffer`) **before** `m_buffer->release()`. This is single-threaded CPU render code — buffer destruction (`destroyResource<Buffer>` → `~MetalBufferImpl`) does not run concurrently with encode. Nulling the sticky slot means the *next* flush skips it (the `buffer_ != nullptr` gate), so a freed handle can never be re-encoded — exactly the #2412 `objc_retain`-on-freed crash. The immediate release is pre-existing (the PR only inserts the scrub above the unchanged release block), so no new use-after-free surface is introduced; the change is strictly safer. `g_runtime()` is intentionally leaked to outlive every buffer dtor, so the untrack call is safe even during shutdown-time destruction.

### On closer read I also confirm
- The doc generalization in `engine/render/CLAUDE.md` ("A new resource type that lands in any sticky table must untrack itself the same way") is accurate and is the right invariant to record — it's the exact class of bug that recurs when a new sticky-bound resource type is added.
- No ECS / IRMath / serialization surface; naming and comment style mirror the `untrackMetalTexture` precedent.
- Metal-only change; carries `fleet:needs-windows-smoke` for GL-backend cross-host validation (Sonnet-tagged on first pass — no build hazard on the GL side since the new symbol is Metal-backend-scoped).

No correctness gaps. Approving.

🤖 Reviewed by Claude Opus 4.8 (opus-reviewer, review-pr skill)

PR #2416, opened 2026-07-15
fleet: clean-exit policy + fix-forward — ir-run RESULT verdicts, doc wiring

jakildev:
## Summary

Human-cued fleet policy change, two halves:

**Clean-exit enforcement (tooling).** `ir-run` now emits a stable one-line verdict on every self-terminating run (`--auto-screenshot` / `--auto-profile` / `--timeout`):

- `RESULT=CLEAN exe=<name> exit=0` — ran to completion, teardown included
- `RESULT=CRASH exe=<name> exit=<rc> signal=<SIGNAME|none>` — exit code propagated; the old softening "screenshots may have been saved… (T-336)" message (which cited a doc section that never existed) is replaced with a directive citing the policy
- `RESULT=ALIVE-TIMEOUT` — watchdog kill of a healthy process; explicitly says nothing about teardown health

`--auto-profile` moves from bare `exec` to the same run-and-validate path. Hermetic test (`scripts/fleet/tests/test_ir_run_result_reporting.sh`, 9 assertions) pins verdict shape, SIGSEGV decode from 139, and exit-code propagation. Verified live on the #2412 crash repro (CRASH verdict + directive) and a clean IRCanvasStress run (CLEAN).

**Policy (docs).** `FLEET.md` gains:

- **§Clean-exit policy** — a signal exit fails the verification step that ran it, even when outputs saved before the crash; the observing agent fixes it regardless of who introduced it; a crash observed is never reported green (file-with-forensics + mark the lane failed is the only fallback).
- **§Fix-forward** — opportunistic fixes (bugs, crashes, dead code, duplication, stale docs) ship with the finder: same PR (own commit, `## Opportunistic fixes` section) or an immediate sibling PR; follow-up issues become the documented exception. Includes the reviewer covenant so sectioned bundles aren't flagged as scope creep.

Pointers wired: `engine/render/CLAUDE.md` (verify section), `render-debug-loop`, `render-verify`, `platform-catchup` (partial-runtime path now fix-forwards instead of filing), `role-smoke-worker` (verdict keys off the RESULT line, not maskable shell status), `review-pr` checklist (+ green-over-crash flag), top-level `CLAUDE.md` index row.

## Opportunistic fixes

- `scripts/gui-verify.py` — a hung GUI-test run killed by the fleet-run watchdog returned exit 0 with zero assertions and read as success; it now detects `RESULT=ALIVE-TIMEOUT` and fails. (Found while auditing consumers for this PR.)

## Test plan

- [x] `bash scripts/fleet/tests/test_ir_run_result_reporting.sh` — 9/9
- [x] Live gpu-path CRASH verdict on the #2412 IRPerfGrid repro (exit 139 → `signal=SIGSEGV`)
- [x] Live gpu-path CLEAN verdict on IRCanvasStress (rc=0)
- [x] `ruff check scripts/gui-verify.py` clean
- [x] `bash -n engine/tools/bin/ir-run`

## Notes

- The smoke/verify harness `scripts/render-verify.py` already failed on fleet-run crashes; this PR aligns the agent-driven flows and the messaging with it.
- Next PR this session (fix-forward in action): root-cause + fix for #2412, the macOS yaw-transition segfault this policy would have caught as red.

### DIFF
--- a/.claude/commands/role-smoke-worker.md
+++ b/.claude/commands/role-smoke-worker.md
@@ -151,7 +151,13 @@ Do **not** add `--timeout` — `fleet-run --timeout` reports "alive at
 deadline" as success, which masks an `--auto-screenshot` hang.
 
 If `fleet-run` exits nonzero or crashes, jump to **step 5 — failure
-verdict** with the run log.
+verdict** with the run log. Key off the `ir-run: RESULT=` line
+(`RESULT=CLEAN` required; `RESULT=CRASH` = failure even if every
+screenshot saved before the crash) rather than shell exit alone — a
+trailing pipe or `; echo` can mask the wrapper's status. This is the
+clean-exit policy
+([`docs/agents/FLEET.md`](../../docs/agents/FLEET.md) §"Clean-exit
+policy"); a crash verdict must never be reported as green.
 
 ### Step 5 — verdict
 
--- a/.claude/skills/platform-catchup/SKILL.md
+++ b/.claude/skills/platform-catchup/SKILL.md
@@ -8,8 +8,8 @@ description: >-
   merged PR. On red (build break), files / opens a fix PR for the offender
   and holds labels on PRs that touch the implicated source paths until
   the fix lands. On partial (per-demo runtime hang or crash but build
-  green), sweeps the rest and files a follow-up perf / functional issue
-  for the offending demo. Runs on Linux, macOS, AND native Windows (MSYS2
+  green), sweeps the rest, then fix-forwards the offending demo this
+  session (issue only as the documented fallback). Runs on Linux, macOS, AND native Windows (MSYS2
   mingw64) — Windows is now the primary validation host. Use when the human
   cues "platform-catchup", "catch up smoke tests", or "catch up Windows
   builds". Skill is **cue-only**, never auto-run — multi-target builds and
@@ -332,6 +332,12 @@ Per-target pass criteria:
 
 ### 7. Decide outcome
 
+"Pass" for a run means the wrapper's `ir-run: RESULT=CLEAN` verdict — a
+`RESULT=CRASH` (teardown crashes included, even with all screenshots
+saved) is a per-demo crash for this table, per
+[`docs/agents/FLEET.md`](../../../docs/agents/FLEET.md) §"Clean-exit
+policy".
+
 | Build result                       | Run result                                  | Outcome    |
 |------------------------------------|---------------------------------------------|------------|
 | All targets build                  | All pass                                    | **green**  |
@@ -375,7 +381,13 @@ touches the offending demo's source tree:
 For each held-back PR, do NOT edit the labels. Note them in the
 report's `held_prs` field.
 
-File a follow-up issue for the offending demo:
+Then apply fix-forward
+([`docs/agents/FLEET.md`](../../../docs/agents/FLEET.md) §"Fix-forward"):
+after the sweep, attempt the offender's fix **this session** — bisect
+the deterministic repro, root-cause, and open a fix PR — regardless of
+which merged change introduced it. Fall back to filing only when the
+fix genuinely exceeds the session (design escalation, other-host-only
+repro), and then with full forensics:
 
 ```bash
 gh issue create --repo <repo> \
@@ -383,9 +395,9 @@ gh issue create --repo <repo> \
   --body "<...>" --label "fleet:task"
 ```
 
-Body covers: which demo, what symptom, log excerpt, suspected root
-cause (e.g. "perf_grid runs ~1 FPS — likely an OpenGL lighting stage
-regression; `/optimize` flow needed").
+Body covers: which demo, the exact repro command and its
+`ir-run: RESULT=` line, log excerpt, the bisect window or suspected
+root cause, and what was already ruled out.
 
 Update marker: `last_outcome = "partial"`,
 `skipped_targets = [<failed-targets>]`,
--- a/.claude/skills/render-debug-loop/SKILL.md
+++ b/.claude/skills/render-debug-loop/SKILL.md
@@ -69,6 +69,13 @@ The demo renders warmup frames, cycles through its configured shots,
 captures one screenshot per shot, and closes the window. `fleet-run`
 handles the working-directory requirement automatically.
 
+**Gate on the run's `ir-run: RESULT=` verdict line, not on screenshots
+existing.** `RESULT=CRASH` (any signal death, teardown included) fails
+this step even when every shot saved — treat the crash as a finding of
+the current iteration and fix it per
+[`docs/agents/FLEET.md`](../../../docs/agents/FLEET.md) §"Clean-exit
+policy". The saved shots remain usable as diagnosis evidence.
+
 ### 4. Read the screenshots
 
 Use the **Glob** tool against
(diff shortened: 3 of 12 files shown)


jakildev (commented):
## Review — fleet: clean-exit policy + fix-forward — ir-run RESULT verdicts, doc wiring

**Verdict:** needs-fix

Reviewed the full diff in file context: `engine/tools/bin/ir-run`, `scripts/gui-verify.py`, the new `scripts/fleet/tests/test_ir_run_result_reporting.sh`, and every doc pointer added (`docs/agents/FLEET.md`, `engine/render/CLAUDE.md`, `.claude/commands/role-smoke-worker.md`, and the four `.claude/skills/*/SKILL.md` files). Ran the new test suite, `ruff check`, and `bash -n` — all clean.

### Needs-fix
- `test/tools/ir_run_exit_code_test.sh:71-74` — this PR removes the old "crashed on shutdown … T-336" diagnostic `ir-run` printed on a non-zero `--auto-screenshot` exit (replaced by `ir_report_result`'s new `RESULT=CRASH … FAILED the clean-exit policy` wording in `engine/tools/bin/ir-run:314-318`), but doesn't touch this pre-existing hermetic test, which pins the exact old text. I ran it against this PR's `ir-run` and it now fails 2/4 checks:
  ```
  FAIL: diagnostic mentions 'crashed on shutdown'
  FAIL: diagnostic mentions T-336
  ir_run_exit_code_test.sh: 2 passed, 2 failed
  ```
  Fix: update the two `check` assertions in `test/tools/ir_run_exit_code_test.sh` to match the new `RESULT=CRASH exe=… signal=…` / "FAILED the clean-exit policy" wording — or delete the file in favor of `scripts/fleet/tests/test_ir_run_result_reporting.sh`, which already covers the same `--auto-screenshot`-crash-propagation case (its own `segfaulter` exe + exit-code/RESULT assertions) with the new wording, so the two files now overlap in scope.

### Nits
- `engine/tools/bin/ir-run:290` — `ir_exit_signal_name`'s `(( rc > 128 && rc < 165 ))` window isn't a natural boundary (128+31=159 covers every standard signal; 128+64=192 covers Linux realtime signals too). A death by realtime signal 37–64 (rc 165–192) would silently report `signal=none`. Doesn't affect any current caller — every crash signal a demo can actually die from (SIGSEGV/SIGABRT/SIGILL/SIGFPE/SIGBUS) is ≤31 — but a one-line comment on why 165, or widening to `< 193`, would remove the ambiguity.
- The PR body's own `## Opportunistic fixes` entry (the `gui-verify.py` hang fix) landed in the same commit as the primary change rather than its own commit — which is what the policy this PR is adding (`docs/agents/FLEET.md` §"Fix-forward") itself asks for. Not a blocker, just a self-referential gap worth noting.

### Praise
- `ir_report_result` / `ir_exit_signal_name` correctly dodge the `local x=$(cmd)` SC2155 exit-status-masking gotcha (declare then assign on separate lines) so `set -euo pipefail` doesn't trip on a failed `kill -l` lookup — confirmed via `bash -n` and the new test's signal-decode case.
- `scripts/gui-verify.py`'s `hung` fix is a real, verified bug, not just defensive code: before this change, a GUI-test run killed by the `fleet-run` watchdog at `--timeout` returned exit 0 with zero `GUI-ASSERT` lines and fell into the "run exited 0 — target may have no assertion tables" pass branch — a hang silently read as a pass. Confirmed the fix threads `RESULT=ALIVE-TIMEOUT` through both the no-assertions branch and the final failure gate.
- All new/updated doc cross-references (`§"Clean-exit policy"`, `§"Fix-forward"`) resolve to real anchors, and every new relative markdown link I checked (`.claude/skills/*/SKILL.md` → `docs/agents/FLEET.md`, `engine/render/CLAUDE.md` → `docs/agents/FLEET.md`) resolves to the correct file.

- [ ] Fix or remove `test/tools/ir_run_exit_code_test.sh` so it doesn't regress against the new `ir-run` output, then re-run it (`bash test/tools/ir_run_exit_code_test.sh`) to confirm 4/4.

Opus recheck not required.

jakildev:
Addressed `fleet:needs-fix` (all items):

- **[needs-fix] `test/tools/ir_run_exit_code_test.sh`** — updated the two stale assertions that pinned the old `crashed on shutdown` / `T-336` diagnostic to the new `RESULT=CRASH exe=IRFakeCrash exit=139 signal=SIGSEGV` + `clean-exit policy` wording. **Kept the file rather than deleting it** (the review offered delete-as-overlap): it uniquely exercises the `--auto-screenshot` GPU-lock path (`ir-acquire gpu`), whereas `test_ir_run_result_reporting.sh` only covers the `--timeout` path — the two are complementary, not overlapping. Header comment now documents that distinction. Re-runs 4/4 (and the sibling result-reporting test stays 9/9).
- **[nit] `engine/tools/bin/ir-run:290`** — widened `ir_exit_signal_name`'s window from `< 165` to `< 193` so a realtime-signal death (rc 165–192) decodes instead of silently reporting `signal=none`, with a comment on the 128+64=192 upper bound. `kill -l` gracefully falls back to the raw number where the host can't name N (verified on macOS: rc 165→SIG37, 192→SIG64, 193→none).
- **[nit] commit hygiene (opportunistic gui-verify.py fix in the primary commit)** — acknowledged; not retroactively splitting already-published history (reviewer flagged it not-a-blocker). The fix-forward policy applies going forward.

Reviewer please re-verify the amended diff. — worker

jakildev (commented):
## Review — fleet: clean-exit policy + fix-forward — ir-run RESULT verdicts, doc wiring

**Verdict:** approve

### Prior-review resolution
- ✅ `test/tools/ir_run_exit_code_test.sh:71-74` — the two assertions pinned the old "crashed on shutdown" / T-336 diagnostic text this PR removed — verified fixed at `241f7233`. Ran the updated test: `bash test/tools/ir_run_exit_code_test.sh` → 4/4, including `RESULT=CRASH exe=IRFakeCrash exit=139 signal=SIGSEGV` and the "clean-exit policy" wording assertions.
- ✅ `engine/tools/bin/ir-run:290` (nit) — signal-decode window widened from `< 165` to `< 193` with a comment explaining the 128+64=192 realtime-signal upper bound and the macOS fallback (`kill -l` → raw number). Logic checks out: standard signals ≤31 (rc ≤159), Linux RT signals 34–64 (rc ≤192), rc ≥193 falls through to `signal=none` as documented.
- ✅ `engine/tools/bin/ir-run` commit-hygiene nit — acknowledged by the author as a going-forward policy item, not a retroactive split; reasonable, this was flagged as not-a-blocker in the original review.

### New commit reviewed (`241f7233`)
Only `engine/tools/bin/ir-run` and `test/tools/ir_run_exit_code_test.sh` changed since the prior review. Re-ran both hermetic suites clean:
- `bash test/tools/ir_run_exit_code_test.sh` → 4/4
- `bash scripts/fleet/tests/test_ir_run_result_reporting.sh` → 9/9
- `bash -n engine/tools/bin/ir-run` and `bash -n test/tools/ir_run_exit_code_test.sh` → both clean

Cross-checked the new grep-pinned strings in the test against `ir_report_result`'s actual output (`engine/tools/bin/ir-run:313-326`) — the `RESULT=CRASH exe=$EXE_NAME exit=$rc signal=$sig` format and the "clean-exit policy" wording match exactly what the test now asserts.

No new issues found in the fix commit.

- [x] Already covered — `test/tools/ir_run_exit_code_test.sh` (4/4) and `scripts/fleet/tests/test_ir_run_result_reporting.sh` (9/9) both re-verified clean on this session.

Opus recheck not required.

PR #2417, opened 2026-07-15
render/metal: scrub sticky bindings on buffer destruction (fixes #2412)

### LINKED ISSUE #2412: crash: macOS/Metal segfault at yaw-state transitions in IRPerfGrid (zoom4_pan) + IRShapeDebug (zoom8_yaw180) on master

jakildev:
## Summary

Fixes the macOS/Metal segfault at yaw-state transitions (#2412) — the crash this session's clean-exit policy work (PR #2416) is designed to make impossible to ship past. Fix-forward in action: found during the #2410 investigation, root-caused and fixed by the finder.

**Root cause** (lldb: `EXC_BAD_ACCESS` in `objc_retain` ← `MTL::ComputeCommandEncoder::setBuffer(index=8)` ← `bindComputeResources` ← `VOXEL_TO_TRIXEL` dispatch):

- `bindComputeResources`/`bindRenderResources` re-encode every non-null **sticky binding-table** entry on each dispatch.
- `~MetalBufferImpl` released its `MTL::Buffer` **without scrubbing those tables** — textures have had this scrub (`untrackMetalTexture`) since #1812-era work; buffers never did.
- Exposure: the overflow-lighting lane (#2357/#2388) transiently binds the per-axis resolve scratch at slot 8 (`kBufferIndex_OverflowLightingScratch`, aliasing `VoxelActiveMask`). That scratch dies when the per-axis canvases release at a cardinal return — the next compute dispatch retained the freed object. Hence the exact crash points: IRPerfGrid `zoom4_pan` (yaw 0.35 → 0) and IRShapeDebug `zoom8_yaw180` (yaw 45° → 180°).

**Fix** (mirrors the existing texture precedent):
- `~MetalBufferImpl`: `replaceMetalBufferInBindings(m_buffer, nullptr)` + defer-release when the buffer was encoded into an in-flight command buffer (same contract as the `subData` orphan path).
- Same-class sibling in `releaseImageAtomicScratchBuffer`: destroying a distance texture whose scratch is the sticky current-scratch pointer now nulls that pointer and defer-releases when encoded.
- Gotcha documented in `engine/render/CLAUDE.md` next to the #1812 stale-image-bind sibling: any new sticky cache of a raw MTL pointer must wire owner-destruction scrubbing.

## Test plan

- [x] IRPerfGrid repro (`--grid-size 16 --zoom 8 --wave-amplitude 0 --auto-screenshot 10`): was 100%-repro SIGSEGV at shot 6/7 → now exit 0, 7/7 shots, across 4 repeat runs
- [x] IRPerfGrid default config (64³ wave): exit 0 (previously crashed at the same transition)
- [x] IRShapeDebug full 21-shot table: was SIGSEGV at shot 17 (`zoom8_yaw180`) → exit 0, all shots
- [x] IRCanvasStress: exit 0, shots byte-identical pre/post fix (img_diff drift 0 on sampled shots) — destruction-path change is visually inert
- [x] `fleet-build --target format-changed` clean
- [ ] Linux/OpenGL smoke (change is Metal-backend-only; GL unaffected by construction)

Closes #2412

### DIFF
--- a/engine/render/CLAUDE.md
+++ b/engine/render/CLAUDE.md
@@ -940,7 +940,7 @@ parity with voxel-pool primary shapes.
   that a stale image shadows. GL is immune (separate image/texture-unit
   namespaces), so a GL-only smoke will not catch it. Any compute kernel that
   mixes a sampler read with images bound elsewhere in the pipeline is exposed;
-   the #1294 chunk cull's fine Hi-Z levels are the known other victim.
+  the #1294 chunk cull's fine Hi-Z levels are the known other victim.
 - **Destroying a Metal resource must scrub the sticky binding tables
   (#2412).** `bindComputeResources`/`bindRenderResources` re-encode every
   non-null sticky table entry on each dispatch, so a released `MTL::Buffer`
--- a/engine/render/CLAUDE.md
+++ b/engine/render/CLAUDE.md
@@ -940,4 +940,16 @@ parity with voxel-pool primary shapes.
   that a stale image shadows. GL is immune (separate image/texture-unit
   namespaces), so a GL-only smoke will not catch it. Any compute kernel that
   mixes a sampler read with images bound elsewhere in the pipeline is exposed;
-  the #1294 chunk cull's fine Hi-Z levels are the known other victim.
+   the #1294 chunk cull's fine Hi-Z levels are the known other victim.
+- **Destroying a Metal resource must scrub the sticky binding tables
+  (#2412).** `bindComputeResources`/`bindRenderResources` re-encode every
+  non-null sticky table entry on each dispatch, so a released `MTL::Buffer`
+  / `MTL::Texture` left in a table is an `objc_retain` on a dangling
+  pointer — a segfault on the NEXT dispatch after the owner died (the
+  #2412 yaw-transition crash: the overflow-lighting scratch, transiently
+  bound at slot 8, died when the per-axis canvases released at a cardinal
+  return). The impl destructors now scrub (`replaceMetalBufferInBindings`
+  in `~MetalBufferImpl`, `untrackMetalTexture` in the texture dtor, and
+  `releaseImageAtomicScratchBuffer` nulls the sticky current-scratch
+  pointer). If you add a NEW sticky cache of a raw MTL object pointer,
+  wire the owner's destruction to scrub it the same way.
--- a/engine/render/src/metal/metal_buffer.cpp
+++ b/engine/render/src/metal/metal_buffer.cpp
@@ -32,7 +32,21 @@ class MetalBufferImpl final : public BufferImpl {
 
     ~MetalBufferImpl() override {
         if (m_buffer != nullptr) {
-            m_buffer->release();
+            // Scrub the sticky binding tables first: bindRenderResources /
+            // bindComputeResources re-encode every non-null table entry on
+            // the next dispatch, so a released buffer left behind is an
+            // objc_retain on a dangling pointer (the #2412 yaw-transition
+            // segfault — a transiently-bound SSBO whose owner died when the
+            // per-axis canvases released). Mirrors untrackMetalTexture.
+            replaceMetalBufferInBindings(m_buffer, nullptr);
+            // If an in-flight command buffer may still read it, defer the
+            // release to the post-wait drain (same contract as the subData
+            // orphan path); otherwise release immediately.
+            if (wasMetalBufferEncoded(m_buffer)) {
+                deferReleaseMetalBuffer(m_buffer);
+            } else {
+                m_buffer->release();
+            }
             m_buffer = nullptr;
         }
     }
--- a/engine/render/src/metal/metal_runtime.cpp
+++ b/engine/render/src/metal/metal_runtime.cpp
@@ -426,7 +426,20 @@ void releaseImageAtomicScratchBuffer(MTL::Texture *texture) {
         return;
     }
     if (it->second != nullptr) {
-        it->second->release();
+        // The sticky current-scratch pointer ("set by every R32I
+        // bindAsImage, never cleared") may still name this buffer;
+        // bindComputeResources re-encodes it for every scratch-consumer
+        // kernel, so leaving it behind is the same dangling-retain class
+        // as an unscrubbed binding-table entry (#2412). Null it, and defer
+        // the release when an in-flight command buffer may still read it.
+        if (currentImageAtomicScratch() == it->second) {
+            setCurrentImageAtomicScratch(nullptr);
+        }
+        if (wasMetalBufferEncoded(it->second)) {
+            deferReleaseMetalBuffer(it->second);
+        } else {
+            it->second->release();
+        }
     }
     cache.erase(it);
 }


jakildev (commented):
## Review — render/metal: scrub sticky bindings on buffer destruction (fixes #2412)

**Verdict:** approve

Reviewed the full diff in file context: `engine/render/src/metal/metal_buffer.cpp`, `engine/render/src/metal/metal_runtime.cpp` (full file, including the pre-existing `subData` orphan path and `bindComputeResources`/`bindRenderResources` in `metal_render_impl.cpp` that consume these tables), and the `engine/render/CLAUDE.md` doc addition.

The root-cause analysis holds up: `bindComputeResources`/`bindRenderResources` (`metal_render_impl.cpp:60-119`) re-encode every non-null `uniformBuffers_`/`shaderStorageBuffers_`/`currentImageAtomicScratch_` entry on each dispatch via `boundMetalBuffer`/`currentImageAtomicScratch`, so a released `MTL::Buffer` left in one of those tables is exactly the dangling-retain class the fix describes. The fix mirrors the existing `untrackMetalTexture` precedent (confirmed both `MetalTexture2DImpl`/`MetalTexture3DImpl` destructors already call it, `metal_texture.cpp:104,351`) and reuses the already-established `wasMetalBufferEncoded`/`deferReleaseMetalBuffer` in-flight contract that `subData`'s orphan path already relies on (`metal_buffer.cpp:65-95`, pre-existing) — this isn't new machinery, just a new call site wired to it.

Checked for gaps in table coverage: `replaceMetalBufferInBindings` scrubs `uniformBuffers_`, `shaderStorageBuffers_`, and `activeVertexLayout_` (vertex/index buffer) — the full set of sticky buffer tables `bindRenderResources`/`bindComputeResources` read. The scratch-buffer fix in `releaseImageAtomicScratchBuffer` correctly nulls `currentImageAtomicScratch_` only when it currently points at the buffer being released (`currentImageAtomicScratch() == it->second`), avoiding clobbering an unrelated live scratch pointer. Both sites scrub-then-decide (defer vs. immediate release) in the same order as the existing orphan path, so no ordering regression.

### Nits
- `engine/render/CLAUDE.md:943` — the diff shifts the pre-existing "the #1294 chunk cull's fine Hi-Z levels..." continuation line from 2-space to 3-space indent (visible in the diff as `-  the #1294...` / `+   the #1294...`). Looks like an unintentional whitespace drift from editing the adjacent bullet rather than a deliberate change — worth a follow-up 1-char fix to keep the list continuation indent consistent with the rest of the file.

### Praise
- The scratch-pointer null-check (`if (currentImageAtomicScratch() == it->second)`) before clearing is the right level of caution — a coarser "always null the sticky scratch on any scratch-buffer release" would have silently dropped a *different*, still-live scratch buffer's binding.
- Doc update lands in the same PR right next to the sibling #1812 gotcha it extends, with an explicit "if you add a NEW sticky cache, wire the scrub" call-to-action — this is exactly the kind of durable invariant this file should carry.

- [x] Already covered by the PR's attached test plan (IRPerfGrid/IRShapeDebug/IRCanvasStress repro + regression runs, `format-changed` clean) — no additional runs needed before merge.

Escalation: please Opus-recheck before merge (touches: engine/render/, GPU buffer lifetime across frames — an in-flight Metal command-encoder / sticky-binding-table invariant is explicitly an Opus-only judgment call per the review checklist, even though the fix mirrors an already-precedented pattern).

Opus recheck required: touches engine/render/ and reasons about GPU buffer lifetime across frames (in-flight Metal command-encoder retention vs. immediate release) — an explicit Opus-only checklist item.

jakildev (commented):
## Opus recheck — render/metal: scrub sticky bindings on buffer destruction (fixes #2412)

**Verdict:** approve (Opus recheck confirms Sonnet's approve)

Sonnet approved on the table-coverage + precedent-mirroring axis (`replaceMetalBufferInBindings` scrubs the full set of sticky tables `bindRenderResources`/`bindComputeResources` read; the scratch null-check is correctly guarded). I re-ran the Opus-only lifetime pass — the in-flight deferral contract and the dead-code / latent-UAF risks that a table-coverage read can't settle — and confirm the fix on all of them.

### Opus-only checks (GPU buffer lifetime across frames)

**1. The scrub-then-defer ordering is sound, and the null-scrub is safe.** Both `bindRenderResources` (`metal_render_impl.cpp:63,70`) and `bindComputeResources` (`:86,92`) guard every buffer entry with `!= nullptr` before re-encoding, so scrubbing a dying buffer to `nullptr` in `uniformBuffers_`/`shaderStorageBuffers_`/`activeVertexLayout_` makes the *next* dispatch skip the slot entirely rather than `objc_retain` a dangling pointer — which is exactly the #2412 crash class. Scrub happens before the defer/release, so there's no window where a re-encode sees the freed pointer.

**2. The `wasMetalBufferEncoded` deferral gate is fed correctly — no latent UAF, no over-defer.** I traced `markMetalBufferEncoded`: it's called at *every* real encode site (uniform `:66,88`, storage `:73,94`, scratch `:116`, vertex/index/indirect `:473,505-506,544-545,593-595,641,679,813`), and `encodedBuffers_` is cleared **only** inside `releaseDeferredMetalBuffers()` (`metal_runtime.cpp:346`), which runs after the per-frame GPU wait-drain. So the gate reads true iff an in-flight command buffer may still touch the buffer → defer; false once the drain guarantees the GPU is done → immediate release. This is the same orphan contract `subData` already relies on, reused rather than re-invented. A buffer that was genuinely in-flight can't slip through to immediate release.

**3. The scratch path fix is live, not dead code, and scrubs the right sticky reference.** `releaseImageAtomicScratchBuffer` is invoked from `~MetalTexture2DImpl` (`metal_texture.cpp:99`), so the new null-of-`currentImageAtomicScratch_` actually runs on texture destruction. The scratch's only sticky reference is that pointer (it's an image-atomic backing buffer, never entered into the uniform/storage tables), so *not* calling `replaceMetalBufferInBindings` here is correct, not a gap — and the scratch is marked encoded at `:116`, so its own `wasMetalBufferEncoded` deferral gate works. The `== it->second` guard before nulling correctly avoids clobbering a different live scratch pointer (Sonnet flagged this; I confirm it's load-bearing).

**4. No new threading hazard.** The destructors mutate `g_runtime()` globals, identical to the already-shipped `untrackMetalTexture` texture path; the render/ECS path is single-threaded today (CPU-multithreaded ECS is still an unimplemented epic), so this introduces no cross-thread mutation the texture precedent didn't already carry.

### Notes (non-blocking)
- `~MetalTexture3DImpl` (`metal_texture.cpp:349`) does not call `releaseImageAtomicScratchBuffer`, but scratch buffers are only ever allocated for 2D R32I image-atomic textures — so this is correct, not a missed site.
- A deferred buffer released at process shutdown (after the final drain, with no subsequent `releaseDeferredMetalBuffers`) would leak — but that's a benign at-exit leak the OS reclaims, and it's identical to the pre-existing `subData` orphan behavior. Not a regression.

### Nits
- I confirm Sonnet's nit: `engine/render/CLAUDE.md:943` shifts the pre-existing "the #1294 chunk cull's fine Hi-Z levels…" continuation from 2-space to 3-space indent (`-  the #1294…` / `+   the #1294…`) — an incidental whitespace drift from editing the adjacent bullet. 1-char follow-up; not worth a re-review round.

The doc invariant added to `engine/render/CLAUDE.md` (destroy-must-scrub-sticky-tables, with the "wire the scrub if you add a new sticky cache" call-to-action) is exactly the right durable form for this class of bug — it sits next to the sibling #1812 sticky-image gotcha and generalizes correctly.

Cross-host: `fleet:needs-windows-smoke` already present (authored-on-macos ✓ Metal tier; one OpenGL smoke outstanding). No smoke-tag action needed — and note this diff is entirely `#if IR_GRAPHICS_METAL`-gated, so the OpenGL smoke validates the build/link, not this code path.

jakildev:
Addressed feedback: fixed the whitespace-indent drift both Sonnet and Opus flagged at `engine/render/CLAUDE.md:943` (3-space → 2-space continuation indent to match the surrounding bullet list). No code changes; `fleet:approved` stands. — worker

jakildev (commented):
## Review — render/metal: scrub sticky bindings on buffer destruction (fixes #2412)

**Verdict:** approve

### Prior-review resolution
- ✅ `engine/render/CLAUDE.md:943` — the #1294 chunk-cull continuation line had drifted from 2-space to 3-space indent — verified fixed at `a05e84b1`. The new commit is exactly the flagged 1-line whitespace fix, nothing else.

### New commit reviewed (`a05e84b1`)
Doc-only, single line: `engine/render/CLAUDE.md:943` drops the continuation line back to the surrounding 2-space indent. No code touched (`metal_buffer.cpp`/`metal_runtime.cpp` unchanged since the prior Sonnet + Opus approve), no new issues introduced.

Both prior fleet passes (Sonnet approve, Opus recheck confirm) already cleared the substantive GPU-buffer-lifetime review of this PR; nothing here reopens that.

- [x] No code changed since the prior approved commit — the previously-confirmed test plan (IRPerfGrid/IRShapeDebug/IRCanvasStress repro + regression) still applies; no additional runs needed for this doc fix.

Opus recheck not required.

jakildev:
judgement-level resolution.

Master already merged #2415 (crash: macOS/Metal segfault at yaw-state
transitions in IRPerfGrid + IRShapeDebug, #2412), which lands its own
independent fix for the same #2412 bug via `replaceMetalBufferInBindings`
+ deferred-release semantics. This PR's own #2412 fix (commit 772b8a7f)
conflicts with that already-landed version in both files below.

Conflicted files:
- `engine/render/CLAUDE.md` — master: `64f75c27 crash: macOS/Metal segfault at yaw-state transitions in IRPerfGrid + IRShapeDebug (#2412) (#2415)`; PR: `a05e84b1 docs(render): fix indent drift in CLAUDE.md #1294 continuation line`
- `engine/render/src/metal/metal_runtime.cpp` — master: `64f75c27 crash: macOS/Metal segfault at yaw-state transitions in IRPerfGrid + IRShapeDebug (#2412) (#2415)`; PR: `772b8a7f render/metal: scrub sticky bindings on buffer destruction (fixes #2412)`

Given #2415 already landed a fix for the same underlying bug, the
worker should likely determine whether this PR is now redundant and
can be closed, rather than force-merging a second fix.

SHA pair: master=64f75c279575c3a70ac56d2b761f938bdabb6124 × PR=a05e84b1eeddd746da7256cb670a65ca9b8bc179

— fleet merger

jakildev:
Semantic-conflict resolution verdict: **redundant — closing** (the merger's suggested outcome checks out).

Master's #2415 (`64f75c27`) independently landed the same #2412 fix with the same mechanism: `replaceMetalBufferInBindings(buffer, nullptr)` from the buffer destructor (via the new `untrackMetalBuffer` helper), the sticky `currentImageAtomicScratch_` null in `releaseImageAtomicScratchBuffer`, and an equivalent destroy-must-scrub invariant in `engine/render/CLAUDE.md` (including the same 'new sticky cache must wire the scrub' call-to-action). `untrackMetalBuffer` additionally erases the dying buffer from `encodedBuffers_`, hygiene this PR's inline version lacks.

This PR's one remaining delta — deferring the `release()` via `wasMetalBufferEncoded`/`deferReleaseMetalBuffer` on the dtor and scratch paths — is not load-bearing: the runtime creates **retaining** command buffers (`metalCommandQueue()->commandBuffer()`, metal_render_impl.cpp:335 et al.), so an encoded buffer is kept alive by the command buffer's own strong reference until execution completes; immediate release after the scrub is safe. The existing defer contract exists for subData *content-mutation* safety (orphan-and-swap), which cannot apply to a destructor. So there is nothing here master doesn't already cover.

Stacked child #2418 has been retargeted to `master` and rebased `--onto` to drop this PR's inherited commits (clean, zero overlap); its slot-8 active-mask restore is the root fix and is now load-bearing on master.

Closing unmerged. — worker

PR #2425, opened 2026-07-15
system: per-system update cadence in the pipeline scheduler (#2404)

### LINKED ISSUE #2404: SystemManager: per-system update cadence (run-every-N-ticks + dt-scaled reduced-rate execution)
## Area
`system` — SystemManager / pipeline scheduling.

## Suggested model
`[opus]` — touches the core system-scheduling loop and the fixed-step time
contract; the shape here affects every system's execution semantics.

## Context
Today `SystemManager::executePipeline(event)` runs every system group, in
order, every time its phase fires — every registered system executes on every
tick of its phase. The only per-system knobs are `Concurrency`
(SERIAL / MAIN_THREAD / PARALLEL_FOR) and `grainSize`. There is no way for a
system to declare that it should run at a *reduced cadence* (every N ticks, or
at a lower target rate) while remaining numerically correct.

Consumers increasingly need coarse-grained, throttled updates for large entity
populations that don't need per-tick fidelity — background / low-priority
subsystems that can advance on a slower clock without visibly diverging,
freeing per-tick budget for foreground work. This is a **generic scheduling
capability**; the *policy* for which systems or entities get throttled belongs
to the consuming creation, not the engine.

## Deliverable
A per-system update-cadence facility in the SystemManager pipeline:

1. A system may declare a cadence — either "run every N phase ticks" (integer
   divisor) or a target sub-rate — as registration metadata alongside
   `Concurrency` / `grainSize`. Unset = current behavior (every tick).
2. Throttled systems are **skipped** on off-cadence ticks (real dispatch /
   iteration savings, not just an in-body early return), preserving
   deterministic ordering with full-rate systems on the ticks they do run.
3. On the ticks it does run, a throttled system can observe the **accumulated
   fixed-step delta** since its previous execution, so per-tick-rate math
   (accumulators, integrators) stays correct at the reduced rate. Reuse the
   existing `IRTime` fixed-step accumulator (`IRTime::tick()` / `fixedStepCount`).
4. Cadence is queryable / adjustable at runtime (a system's cadence can change
   without re-registration) so consumers can build dynamic policies on top.

## Acceptance
1. A system registered with cadence N executes on exactly 1-in-N phase ticks;
   a test asserts the call count over M ticks is ⌊M/N⌋ (±1).
2. On each execution the system observes the correct accumulated fixed-step
   delta since its previous execution (verified across a cadence change mid-run).
3. Off-cadence ticks incur no per-entity iteration for that system (measured:
   iteration count scales with 1/N).
4. Unset cadence = unchanged behavior; existing systems and pipeline ordering
   are unaffected (regression: existing demos/tests pass).
5. Cadence is runtime-mutable via the SystemManager API.

## Notes
- **Generic engine capability only.** No consumer-specific policy (interest
  sets, camera / visibility, spatial LOD) belongs in this issue — those live in
  the consuming creation. The engine provides the throttle; creations own the
  semantics.
- Orthogonal to render LOD (`LOD_UPDATE` / `C_ActiveLodLevel`, which derives
  mesh detail from zoom) — this is **execution cadence**, not render detail.
- Complementary to the ECS threading umbrella (#226): that is parallelism
  *within* a tick; this is frequency *across* ticks.

jakildev:
## Summary
- Adds a **per-system update cadence** to `SystemManager`: a system runs on 1-in-N phase ticks instead of every tick, declared at registration (`kCadence` / `kCadenceOffset` on the `System<N>` spec, or the trailing `createSystem` params) or set at runtime.
- Off-cadence ticks skip the system's **entire** dispatch — `beginTick`, archetype query, per-entity iteration, `endTick`, observer fires — so the saving is real dispatch/iteration cost, not an in-body early return. The per-group `flushStructuralChanges` still runs every tick, so deferred changes queued by other systems keep flushing.
- The gate lives in `executePipeline` behind a SystemManager-owned per-event tick counter (self-contained, no `TimeManager` dependency — works for INPUT/RENDER too and is unit-testable in isolation). The due check is additive (`now >= lastRun + cadence`) to stay underflow-safe with a nonzero offset seed, and re-phases from the last run on a runtime cadence change.
- **Offset** (amendment 1) staggers sibling systems so cold-tier buckets fire on consecutive ticks instead of spiking on the same tick. Multi-system parallel groups filter due members into a reused scratch vector on the main thread before fan-out, so a worker never races the accumulated-delta write.
- `getAccumulatedTicks` / `accumulatedDeltaTime` (UPDATE-phase-only) let a throttled integrator stay numerically correct at the reduced rate; `cadenceFromRate(hz)` maps a target sub-rate onto the divisor. Full C++ + Lua surface (amendment 2 binds the accumulated-delta getters for Lua-registered throttled systems).
- **Generic engine throttle only** — throttle *policy* (which systems/entities) stays in the consuming creation, per the issue. Orthogonal to render LOD and to the T-224 threading groups.

## Acceptance criteria (#2404)
- [x] A cadence-N system executes floor(M/N) (+/-1) times over M `executePipeline` calls
- [x] Correct accumulated fixed-step delta each execution, verified across a mid-run cadence change (`3,3,3 -> 5,5`, summing to total elapsed ticks)
- [x] Off-cadence ticks incur no per-entity iteration (visits scale ~1/N)
- [x] Unset cadence = unchanged behavior; existing suite green
- [x] Cadence runtime-mutable; a mixed-cadence multi-system group dispatches only due members

## Plan
The approved `## Plan` comment (+ the two accepted in-thread amendments) rides as the first commit at `.fleet/plans/issue-2404.md`.

## Test plan
- [x] `test/system/system_cadence_test.cpp` — 7 tests covering all five acceptance criteria + offset stagger + setter normalization / `cadenceFromRate` (`build/IrredenEngineTest --gtest_filter='SystemCadenceTest.*'` -> **7 passed**)
- [x] No regression: existing system suite (PipelineGroups / Concurrency / RegisterSystem / SystemAccess = **55 tests**) green; full non-MIDI suite **1320 tests** green
- [x] `fleet-build --target IrredenEngineTest` clean on macOS (Metal)

Note: the full-suite run aborts in a MIDI test on `MidiInCore::initialize: error creating OS-X MIDI client object (-10833)` — a pre-existing macOS-headless CoreMIDI environment failure, unrelated to `engine/system`; the non-MIDI subset is fully green.

Perf: no render surface touched. The cadence-1 fast path is a bounds-check + short-circuited `cadence > 1` compare + a couple of integer stores per system per tick — negligible next to the `std::vector<ArchetypeNode*>` allocation + hash-lookup `executeSystem` already does every tick, which throttled systems now skip entirely. No per-frame allocation added (the multi-system due-filter reuses `m_dueScratch`).

Closes #2404

### DIFF
--- a/.fleet/plans/issue-2404.md
+++ b/.fleet/plans/issue-2404.md
@@ -1,0 +1,71 @@
+## Plan: SystemManager: per-system update cadence (run-every-N-ticks + dt-scaled reduced-rate execution)
+
+- **Issue:** #2404
+- **Model:** opus — the approach below is committed; implementation is a bounded single-module scheduler change + tests, but it touches the core pipeline dispatch path so it stays above sonnet.
+- **Date:** 2026-07-14
+
+### Scope
+
+A per-system execution-cadence facility in the SystemManager pipeline: a system declares (at registration) or sets (at runtime) "run every N phase ticks"; off-cadence ticks skip the system's entire dispatch (no `beginTick`, no node query, no per-entity iteration); on-cadence executions can read how many phase ticks — and how much fixed-step time — elapsed since their previous execution. Generic engine capability only; throttle policy stays in consuming creations.
+
+### Verified current state (master `5374cf78`, 2026-07-14)
+
+- `SystemManager::executePipeline` (`engine/system/src/system_manager.cpp:334`) runs every group and every member unconditionally; the only per-system dispatch knobs are `Concurrency` + `grainSize` (parallel vectors `m_concurrency` / `m_grainSize` emplaced in `createSystem`). **No cadence/throttle mechanism exists in the scheduler** — verified by grep across `engine/` for cadence/throttle/every-N/sub-rate shapes: the only hits are in-body counters private to individual systems (`system_perf_stats_overlay.hpp` text-refresh throttle, `system_voxel_to_trixel.hpp` log throttles), which still pay full dispatch + iteration — exactly the pattern this issue replaces with real dispatch savings.
+- Phase ticks: `World::gameLoop` calls `executePipeline(INPUT)` once per frame, `executePipeline(UPDATE)` once per `shouldUpdate()` fixed step (0..maxTicks catch-up per frame), `executePipeline(RENDER)` once per frame. `IRTime::tick()` (= `EventProfiler<UPDATE>::fixedStepCount`) counts UPDATE ticks only; SystemManager has no per-event execution counter today.
+- `IRTime::deltaTime(UPDATE)` is the constant fixed step (1/kFPS, const-after-ctor, worker-readable). For UPDATE-phase systems, "accumulated fixed-step delta since previous execution" ≡ elapsedPhaseTicks × `deltaTime(UPDATE)` exactly — no new accumulator infrastructure needed in `engine/time`; this reuses the existing `IRTime` fixed-step contract as the issue asks.
+- Registration-metadata pattern to mirror: `registerSystem<N>` detects `static constexpr Concurrency kConcurrency` / `int kGrainSize` via concepts (`HasConcurrencyMember` / `HasGrainSizeMember`, `ir_system.hpp`); the free-function `createSystem` takes trailing `concurrency, grainSize` params.
+- Per-system timing (`TimingAccum.callCount_` / `totalEntityCount_`, gated on `setTimingEnabled`) already provides the measurement surface for "executed K times over M ticks" and "iteration scales with 1/N".
+- Sibling / in-flight reconciliation: no open PR touches `engine/system` (checked all 7 open engine PRs — render/docs/fleet/math surfaces). #226 (threading epic) is complementary by design — T-224 groups parallelize *within* a tick; cadence decides whether a system dispatches at all on a given tick. The gate below runs on the main thread before group fan-out, so there is no interaction with the group validator (a throttled member validates identically for the ticks it does run). Render LOD (`LOD_UPDATE` / `C_ActiveLodLevel`) is orthogonal per the issue. The pausable sim clock (`IRSim::tick()`) is deliberately NOT the cadence timebase — cadence counts pipeline executions; sim-time-aware policies compose on top via the runtime setter.
+
+### Approach
+
+Committed design: **integer divisor is the stored primitive; elapsed-ticks comparison (not modulo); gating lives in `executePipeline` only.**
+
+1. **State (`system_manager.hpp`).** Parallel vectors emplaced in `createSystem` / `createSystemDynamic` alongside `m_concurrency`: `m_cadence` (`uint32_t`, 1 = every tick, 0 normalized to 1), `m_cadenceOffset` (`uint32_t`, initial phase 0..cadence-1), `m_lastRunTick` (`uint64_t`), `m_accumulatedTicks` (`uint64_t`). Plus `std::array<uint64_t, IRTime::END + 1> m_eventTickCounts{}` — SystemManager-owned per-event execution counters (self-contained; works for INPUT/RENDER phases too; unit-testable without a TimeManager). Plus a reusable `std::vector<SystemId> m_dueScratch` for multi-system group filtering (reserved once; no per-frame allocation).
+2. **Gate (`system_manager.cpp` `executePipeline`).** Increment `m_eventTickCounts[event]` once at entry → `now`. A system is *due* iff `cadence <= 1 || now >= m_lastRunTick[id] + cadence` (addition, not `now % N` and not `now - lastRun` — the additive form is underflow-safe with a nonzero offset seed and re-phases correctly when the cadence changes at runtime, so call count over M ticks is ⌊M/N⌋ ±1). Singleton group: if not due, skip the observer fires + `executeSystem` (the per-group `flushStructuralChanges` still runs unconditionally). Multi-system group: filter due members into `m_dueScratch` on the main thread before fan-out; skip the `IRJob::parallelFor` entirely when none are due. For each due system, before dispatch (main thread): `m_accumulatedTicks[id] = now - m_lastRunTick[id]`, then `m_lastRunTick[id] = now`. Workers never write cadence state; tick bodies may read it (written-before-dispatch, read-only during execution — same thread-safety argument as the fixed dt).
+3. **Late-join stamping.** `registerPipelineGroups` / `appendToPipeline` / `insertSingletonGroupRelativeTo` stamp each newly-listed system's `m_lastRunTick = m_eventTickCounts[event] + offset`, so a system joining a live pipeline measures elapsed from its join point — not from counter zero (otherwise its first accumulated delta spans the pipeline's whole lifetime) — and its offset staggers the initial phase relative to co-registered siblings.
+4. **Runtime API (`system_manager.hpp` + `ir_system.hpp` free functions).** `setSystemCadence(SystemId, uint32_t)` (normalizes 0→1; takes effect next phase tick, no re-registration, re-phases from the last run) and `getSystemCadence(SystemId)`. `setSystemCadenceOffset(SystemId, uint32_t)` / `getSystemCadenceOffset(SystemId)` (amendment 1: honor offset from the runtime setter). `getAccumulatedTicks(SystemId)` — phase ticks covered by the current/most-recent execution (≥1 once it has run). Convenience `IRSystem::accumulatedDeltaTime(SystemId)` = `getAccumulatedTicks(id) * IRTime::deltaTime(IRTime::UPDATE)` — exact for UPDATE-phase systems (fixed step); documented UPDATE-phase-only (RENDER dt is wall-clock-variable; RENDER-phase consumers use raw ticks). The issue's "or a target sub-rate" is satisfied as sugar, not a second scheduling mode: `IRSystem::cadenceFromRate(targetHz)` returns `max(1, round(kFPS / targetHz))`.
+5. **Registration metadata (`ir_system.hpp`).** Trailing `uint32_t cadence = 1, uint32_t offset = 0` params on the free-function `createSystem` (after `grainSize`) and on `SystemManager::createSystem` (before `accessDescriptor`; the wrapper is its only positional caller). `registerSystem<N>` detects `static constexpr uint32_t kCadence` / `kCadenceOffset` via `HasCadenceMember` / `HasCadenceOffsetMember` concepts + `cadenceOf<T>()` / `cadenceOffsetOf<T>()`, mirroring `kGrainSize`. `createSystemDynamic` gains trailing `cadence = 1, offset = 0` for symmetry.
+6. **Lua binding (`lua_pipeline_bindings.hpp`).** `IRSystem.setSystemCadence(sysId, n)` / `IRSystem.getSystemCadence(sysId)` / `IRSystem.setSystemCadenceOffset(sysId, o)` / `IRSystem.getSystemCadenceOffset(sysId)` alongside the pipeline-composition bindings — the surface consuming creations build dynamic throttle policies on. Amendment 2: also bind `IRSystem.getAccumulatedTicks(sysId)` and `IRSystem.accumulatedDeltaTime(sysId)` so a Lua-registered throttled system (EVAL/CODEGEN) can read the numerically-correct-at-reduced-rate delta (issue deliverable 3), same UPDATE-phase-only caveat as the C++ convenience.
+7. **`executeSystem` stays ungated.** Cadence is a pipeline-scheduling property; direct `executeSystem` / `executeQuery` calls always run. Keeps "unset = unchanged" airtight and manual-invocation semantics intact.
+8. **Docs.** `engine/system/CLAUDE.md` gains a "Per-system cadence" section: declaration forms, the accumulated-delta contract, and skip semantics (`beginTick`/`endTick` also skip; observers don't fire on skipped ticks, so GPU-stage timing rows hold stale samples on skip frames).
+
+One task, one PR — core + tests + Lua binding + CLAUDE.md are one bounded surface; no carve-offs.
+
+### Amendments (accepted in-thread, folded into this plan)
+
+1. **Phase offset.** Optional `offset` (initial phase, 0..cadence-1) alongside the cadence — registration metadata (`kCadenceOffset` / `createSystem` trailing param) + honored by the runtime setter (`setSystemCadenceOffset`). K sibling cold-tier buckets at cadence K registered together now stagger across K consecutive ticks instead of spiking on the same tick.
+2. **Lua exposure of the accumulated delta.** Bind `IRSystem.getAccumulatedTicks(sysId)` and `IRSystem.accumulatedDeltaTime(sysId)` in addition to the cadence getter/setter, so a Lua-registered throttled system can read the accumulated delta and stay numerically correct at the reduced rate.
+
+Explicitly NOT this issue: deferred component add/remove from an EVAL Lua tick (tier-migration systems) — a separate #2286-adjacent gap, filed by the consumer if confirmed.
+
+### Affected files
+
+- `engine/system/include/irreden/system/system_manager.hpp` — cadence/offset/lastRun/accumulated vectors, per-event counters, due-scratch, `createSystem` params, runtime API
+- `engine/system/src/system_manager.cpp` — `executePipeline` gate + bookkeeping, late-join stamping, `createSystemDynamic` params, runtime-API impls
+- `engine/system/include/irreden/ir_system.hpp` — `createSystem` trailing params, `kCadence`/`kCadenceOffset` detection, free functions (`setSystemCadence` / `getSystemCadence` / `setSystemCadenceOffset` / `getSystemCadenceOffset` / `getAccumulatedTicks` / `accumulatedDeltaTime` / `cadenceFromRate`)
+- `engine/script/include/irreden/script/lua_pipeline_bindings.hpp` — Lua setter/getter + accumulated-delta bindings
+- `test/system/system_cadence_test.cpp` — new; fixture pattern per `test/system/pipeline_groups_test.cpp`
+- `test/CMakeLists.txt` — register the new test
+- `engine/system/CLAUDE.md` — cadence section
+
+### Acceptance criteria
+
+Concrete versions of the issue's five:
+
+1. A cadence-N system executes ⌊M/N⌋ (±1) times over M `executePipeline` calls (counting tick body; cross-checked against `TimingAccum.callCount_` with timing enabled).
+2. The sum of per-execution `getAccumulatedTicks` across a mid-run `setSystemCadence` change equals total elapsed phase ticks, and each execution's value matches its actual gap (e.g. 3,3,3 → change to 5 → 5,5); `accumulatedDeltaTime` = that × fixed dt.
+3. Off-cadence ticks fire no `beginTick`/`tick`/`endTick`, and per-entity visits scale ~1/N (`totalEntityCount_` or a visit counter).
+4. Default path unchanged: full existing suite green (`pipeline_groups_test`, `system_concurrency_test`, `register_system_test` untouched); the cadence-1 fast path is one load + compare.
+5. Mixed-cadence multi-system group: only due members dispatch on a given tick (use the `g_jobManager == nullptr` serial fallback for a deterministic unit test); runtime mutability covered by 2.
+
+### Gotchas
+
+- Stamp `m_lastRunTick` when a system joins a live pipeline (step 3) — without it, the first accumulated delta after a late `appendToPipeline` spans the whole run.
+- Skips must NOT skip the per-group `flushStructuralChanges` — deferred structural changes queued by other systems still flush on the existing cadence.
+- The due-filter for parallel groups runs on the main thread (reused scratch vector, no allocation in the frame loop); do not move the check inside the worker lambda.
+- The due comparison is `now >= lastRun + cadence` (addition), NOT `now - lastRun >= cadence` — the offset seed makes `lastRun > now` on early ticks, and the subtraction form underflows `uint64_t` there. Accumulated (`now - lastRun`) is computed only at fire time, where `now >= lastRun` holds.
+- `IRTime::Events` is a C-style enum (`UPDATE, RENDER, INPUT, START, END`) — size the counter array `END + 1`; no hashing.
+- `kCadence` goes on the `System<N>` specialization as a constexpr member (like `kGrainSize`), NOT in the component pack — tag types in the `registerSystem` pack have a known filtering gap.
+- Perf overlay / GPU-stage rows show stale samples for throttled systems on skip frames — expected; note it in CLAUDE.md rather than special-casing the observer.
+- Purely additive public surface — no engine API removals.
--- a/engine/script/CLAUDE.md
+++ b/engine/script/CLAUDE.md
@@ -832,6 +832,17 @@ IRSystem.registerPipeline(IRTime.UPDATE, {
   release a bad anchor silently front-inserts.
   Underlying semantics: `engine/system/CLAUDE.md` "Appending to a live
   pipeline".
+- **`IRSystem.setSystemCadence(sysId, n)`** / **`getSystemCadence(sysId)`**
+  (#2404) — throttle a system to run 1-in-`n` phase ticks (n ≥ 1; 1 = every
+  tick). Off-cadence ticks skip its entire dispatch. Paired with
+  **`setSystemCadenceOffset(sysId, o)`** / **`getSystemCadenceOffset(sysId)`**
+  — an initial phase stagger (`0..n-1`) so sibling throttled systems don't
+  spike on the same tick. A throttled Lua system that integrates at the
+  reduced rate reads **`getAccumulatedTicks(sysId)`** (phase ticks this run
+  covers) or **`accumulatedDeltaTime(sysId)`** (that × the fixed UPDATE dt,
+  UPDATE-phase-only) to stay numerically correct. `sysId` is any SystemId
+  (from `IRSystem.systemId` or `IRSystem.registerSystem`). Underlying
+  semantics: `engine/system/CLAUDE.md` "Per-system update cadence".
 - **Game-side enums** (e.g. `IRGameSystem.GameSystemName`) extend the
   same pattern — bind the game's own enum table at game-side init via
   `LuaScript::registerEnum<...>()` plus `registerPrefabSystemId` for each
--- a/engine/script/include/irreden/script/lua_pipeline_bindings.hpp
+++ b/engine/script/include/irreden/script/lua_pipeline_bindings.hpp
@@ -34,6 +34,7 @@
 #include <irreden/ir_time.hpp>
 #include <irreden/script/lua_script.hpp>
 
+#include <cstdint>
 #include <list>
 #include <string>
 #include <unordered_map>
@@ -343,6 +344,51 @@ inline void bindRegisterPipelineAndSystemId(
                 static_cast<IRSystem::SystemId>(anchorId)
             );
         };
+
+    // #2404: per-system update cadence. A Lua throttle policy sets a
+    // system to run 1-in-N phase ticks (`setSystemCadence`), staggers its
+    // initial phase (`setSystemCadenceOffset`), and — for a throttled
+    // Lua-registered system that must stay numerically correct at the
+    // reduced rate — reads how many phase ticks / how much fixed-step
+    // time its current execution covers (`getAccumulatedTicks` /
+    // `accumulatedDeltaTime`, amendment 2). `sysId` is any SystemId (from
+    // IRSystem.systemId or IRSystem.registerSystem).
+    lua["IRSystem"]["setSystemCadence"] = [](lua_Integer sysId, lua_Integer cadence) {
+        if (cadence < 1) {
+            throw sol::error{"IRSystem.setSystemCadence: cadence must be >= 1 (1 = every tick)"};
+        }
+        IRSystem::setSystemCadence(
+            static_cast<IRSystem::SystemId>(sysId),
+            static_cast<std::uint32_t>(cadence)
+        );
+    };
+    lua["IRSystem"]["getSystemCadence"] = [](lua_Integer sysId) -> lua_Integer {
+        return static_cast<lua_Integer>(
+            IRSystem::getSystemCadence(static_cast<IRSystem::SystemId>(sysId))
+        );
+    };
+    lua["IRSystem"]["setSystemCadenceOffset"] = [](lua_Integer sysId, lua_Integer offset) {
+        if (offset < 0) {
+            throw sol::error{"IRSystem.setSystemCadenceOffset: offset must be >= 0"};
+        }
+        IRSystem::setSystemCadenceOffset(
+            static_cast<IRSystem::SystemId>(sysId),
+            static_cast<std::uint32_t>(offset)
+        );
+    };
+    lua["IRSystem"]["getSystemCadenceOffset"] = [](lua_Integer sysId) -> lua_Integer {
+        return static_cast<lua_Integer>(
+            IRSystem::getSystemCadenceOffset(static_cast<IRSystem::SystemId>(sysId))
+        );
+    };
+    lua["IRSystem"]["getAccumulatedTicks"] = [](lua_Integer sysId) -> lua_Integer {
+        return static_cast<lua_Integer>(
+            IRSystem::getAccumulatedTicks(static_cast<IRSystem::SystemId>(sysId))
+        );
+    };
+    lua["IRSystem"]["accumulatedDeltaTime"] = [](lua_Integer sysId) -> double {
+        return IRSystem::accumulatedDeltaTime(static_cast<IRSystem::SystemId>(sysId));
+    };
 }
 
 } // namespace IRScript::detail
--- a/engine/system/CLAUDE.md
+++ b/engine/system/CLAUDE.md
@@ -530,6 +530,99 @@ in `IR_RELEASE` builds and safe to call when `g_jobManager ==
 nullptr` (unit tests / pre-`World` startup return `true` as the
 default).
 
+## Per-system update cadence (#2404)
+
+A system can declare — or set at runtime — a **cadence**: run on 1-in-N
+phase ticks of the pipeline it belongs to, instead of every tick. This is
+a *generic scheduling throttle*; which systems (or entities) get throttled
+is policy the consuming creation owns, not the engine. It is orthogonal to
+render LOD (`LOD_UPDATE` / `C_ActiveLodLevel`, which derives mesh detail
+from zoom) and to the T-224 threading groups (which parallelize *within* a
+tick): cadence decides whether a system dispatches *at all* on a given tick.
+
+Off-cadence ticks skip the system's **entire** dispatch — no `beginTick`,
+no archetype-node query, no per-entity iteration, no `endTick`, no observer
+fires. The saving is real dispatch/iteration cost, not an in-body early
+return. The per-group `flushStructuralChanges` still runs on every tick, so
+deferred changes queued by *other* systems keep flushing on schedule.
+
+### Declaring a cadence
+
+```cpp
+// registerSystem<N> form: constexpr members on the System<N> spec,
+// mirroring kConcurrency / kGrainSize. Absent → cadence 1 / offset 0.
+template <> struct System<BACKGROUND_DECAY> {
+    static constexpr std::uint32_t kCadence = 4;        // run every 4th tick
+    static constexpr std::uint32_t kCadenceOffset = 1;  // initial phase stagger
+    void tick(C_Foo &f) {
+        // integrate over the ticks this run covers so the reduced rate
+        // stays numerically correct — see accumulatedDeltaTime below.
+    }
+    static SystemId create() { return registerSystem<BACKGROUND_DECAY, C_Foo>("BgDecay"); }
+};
+
+// createSystem<Cs...> free function: trailing (cadence, offset) after grainSize.
+auto id = IRSystem::createSystem<C_Foo>(
+    "BgDecay", tickFn, beginFn, endFn, {}, relFn,
+    IRSystem::Concurrency::SERIAL, IRSystem::kDefaultGrainSize, /*cadence=*/4, /*offset=*/1);
+```
+
+`cadence == 1` (the default, unset) is every tick — legacy behavior, one
+load + compare on the hot path. `0` normalizes to `1`.
+
+### Runtime API (`IRSystem::` free functions + SystemManager methods)
+
+- `setSystemCadence(id, n)` / `getSystemCadence(id)` — change the divisor with
+  no re-registration; takes effect next tick and **re-phases from the last
+  run** (a shorter cadence fires sooner, a longer one later).
+- `setSystemCadenceOffset(id, o)` / `getSystemCadenceOffset(id)` — the initial
+  phase, `0..cadence-1`. K sibling cold-tier systems registered together at
+  cadence K with distinct offsets fire on K consecutive ticks (smoothed
+  load) instead of spiking on the same tick. Setting it at runtime re-staggers
+  from the current phase tick.
+- `getAccumulatedTicks(id)` — phase ticks the current/most-recent execution
+  covers (`>= 1` once it has run).
+- `accumulatedDeltaTime(id)` — **UPDATE-phase-only** convenience:
+  `getAccumulatedTicks(id) * IRTime::deltaTime(UPDATE)`. For an integrator or
+  accumulator running at a reduced rate, multiply your per-tick delta by this
+  so the math stays correct. RENDER-phase throttled consumers must use raw
+  `getAccumulatedTicks` — RENDER dt is wall-clock-variable.
+- `cadenceFromRate(hz)` — sugar mapping a target sub-rate to the nearest
+  integer divisor of the fixed UPDATE rate (`round(kFPS / hz)`, min 1). The
+  issue's "target sub-rate" is this helper, not a second scheduling mode.
+
+Lua mirrors the whole surface: `IRSystem.setSystemCadence(sysId, n)` /
+`getSystemCadence` / `setSystemCadenceOffset` / `getSystemCadenceOffset` /
+`getAccumulatedTicks` / `accumulatedDeltaTime` (see `engine/script/CLAUDE.md`).
+
+### How the gate works
+
+`executePipeline` bumps a SystemManager-owned per-event tick counter
+(`m_eventTickCounts`, sized `IRTime::END + 1` — self-contained, no
+TimeManager dependency, so cadence works for INPUT/RENDER and is unit-
+testable in isolation) once at entry to `now`, then gates each system with
+`now >= m_lastRunTick[id] + cadence`. The comparison is **additive, not
+`now - lastRun >= cadence`** — a nonzero offset seeds `lastRun > now` on the
+first ticks, where the unsigned subtraction would underflow; `now - lastRun`
+(the accumulated delta) is computed only at fire time, where `now >= lastRun`
+holds. A due system stamps `m_accumulatedTicks[id] = now - lastRun`, then
+`m_lastRunTick[id] = now`, on the main thread **before** any worker fan-out
+(a multi-system group filters its due members into a reused scratch vector
+first), so a tick body reading the accumulated delta from a worker never
+races the write. Joining a pipeline (`registerPipelineGroups` /
+`appendToPipeline` / insert) seeds `lastRun` from the current event counter
+(+ offset), so a system's first accumulated delta measures from its join
+point — not the pipeline's whole lifetime.
+
+`executeSystem` / `executeQuery` are **ungated** — cadence is a pipeline-
+scheduling property, so direct/manual invocation always runs.
+
+**Skip semantics on observers.** The `TickObserver` bracket (GPU-stage
+timing) fires only on ticks a singleton system actually runs, so the perf
+overlay / GPU-stage rows hold *stale* samples for a throttled system on its
+skip frames. That is expected — it's the honest picture of "this system
+didn't run this frame" — not a bug to special-case.
+
 ## Gotchas
 
 - **Archetype mutations in a TICK will skip or revisit entities.** If you
--- a/engine/system/CLAUDE.md
+++ b/engine/system/CLAUDE.md
@@ -372,6 +372,12 @@ IRSystem::insertIntoPipelineAfter(IRTime::Events::UPDATE, sysId, anchor);
 - All three assert (debug; no-op in release) if `sysId` is already in
   the event's pipeline (a double-add would tick it twice), and the
   insert forms assert if `anchor` isn't found.
+- Joining a system into a **different** event's pipeline while it's
+  still listed in its prior one trips a matching debug assert in
+  `stampCadenceJoin` — a system's cadence clock binds to one event, so
+  being live in two pipelines at once would thrash `m_lastRunTick`
+  between two counters that advance at different rates. Clear the old
+  pipeline (or the group containing it) first.
 - Lua surface: `IRSystem.appendSystem(event, sysId)`,
   `IRSystem.insertSystemBefore(event, sysId, anchor)`,
   `IRSystem.insertSystemAfter(event, sysId, anchor)` — see
--- a/engine/system/include/irreden/ir_system.hpp
+++ b/engine/system/include/irreden/ir_system.hpp
@@ -9,6 +9,7 @@
 #include <irreden/system/system_access.hpp>
 #include <irreden/system/system_manager.hpp>
 
+#include <cstdint>
 #include <functional>
 #include <optional>
 #include <tuple>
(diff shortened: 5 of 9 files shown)


jakildev (commented):
## Review — system: per-system update cadence in the pipeline scheduler (#2404)

**Verdict:** approve

### Nits
- `engine/script/include/irreden/script/lua_pipeline_bindings.hpp:126` / `engine/system/src/system_manager.cpp:596` — `IRSystem.setSystemCadence` throws on cadence `0` from Lua, but `SystemManager::setSystemCadence` / `emplaceCadenceState` silently normalize `0` to `1` on the C++ side. Minor asymmetry between the two entry points; either document it as deliberate (Lua callers get a stricter guard against a likely typo) or align the two.

### Praise
- The additive due-check (`now >= lastRun + cadence`, not `now - lastRun >= cadence`) is the right call for underflow-safety with a nonzero offset seed — traced the tick-by-tick math by hand against all six new tests (`RunsOneInNTicks`, `OffCadenceTicksSkipIteration`, `AccumulatedTicksRephaseOnCadenceChange`, `MultiSystemGroupFiltersDueMembers`, `OffsetStaggersSiblings`, `RuntimeSettersNormalizeAndCadenceFromRate`) and every assertion matches the gate's actual behavior.
- Main-thread due-filtering into `m_dueScratch` before `IRJob::parallelFor` fan-out (`system_manager.cpp:759-771`) correctly keeps the `m_accumulatedTicks` / `m_lastRunTick` writes off worker threads, consistent with the existing thread-safety contract for per-system state.
- Late-join stamping (`stampCadenceJoin`) is wired into all three join paths — `registerPipelineGroups` (and its `registerPipeline` delegate), `appendToPipeline`, and `insertSingletonGroupRelativeTo` (covering both `insertIntoPipelineBefore`/`After`) — so no join path leaves a system's first accumulated delta measuring from counter zero.

- [x] Already covered by the PR's own reported run (7/7 `SystemCadenceTest.*`, 55/55 system suite, 1320 non-MIDI suite green) — no further action needed.

Escalation: please Opus-recheck before merge (touches: engine/system core pipeline dispatch + parallel-group scheduling — concurrency-sensitive, above the sonnet bar per the PR's own plan).

jakildev (commented):
## Opus recheck — system: per-system update cadence in the pipeline scheduler (#2404)

**Verdict:** needs-fix

The design is right and the registration-time path is correct — integer divisor as the stored primitive, additive due-check, gate on the main thread before fan-out, `flushStructuralChanges` outside the gate. Both defects below are in the **runtime setters**, which is why all 7 tests pass: every test drives cadence through registration metadata and only ever ticks `UPDATE`.

Neither bug regresses anything on master today — I grepped `engine/`, `creations/` and `test/`, and there are **no in-tree consumers** of `setSystemCadence` / `setSystemCadenceOffset` / `getAccumulatedTicks` / `cadenceFromRate`. These are latent traps for the first consumer, not live breakage. That's why this is needs-fix and not a blocker.

### Needs-fix

**1. `pollCadenceDue` — `uint64_t` underflow on the `cadence == 1` path (`engine/system/src/system_manager.cpp:107-121`)**

The `cadence > 1 &&` short-circuit bypasses the due check, and the code then unconditionally computes `now - m_lastRunTick[system]`. The comment on line 118 states *"Due: `now >= lastRun` holds here, so the subtraction is safe"* — that invariant is guaranteed by the additive check on the `cadence > 1` path, but `cadence == 1` never reaches that check. `stampCadenceJoin` seeds `lastRun = now + offset` (ahead of `now`), and `setSystemCadence` deliberately does **not** re-seed `lastRunTick` (line 155-158), so the two can disagree.

Reachable from the documented public API (and the Lua binding):

```cpp
auto sys = IRSystem::createSystem<C_Foo>(/*…*/, /*cadence=*/4, /*offset=*/3);
m_system_manager.registerPipeline(IRTime::UPDATE, {sys}); // stamp: lastRun = 0 + 3 = 3
m_system_manager.executePipeline(IRTime::UPDATE);         // now=1: 1 < 3+4 → skip, lastRun stays 3
m_system_manager.setSystemCadence(sys, 1);                // cadence=1, offset→0, lastRun NOT re-seeded (3)
m_system_manager.executePipeline(IRTime::UPDATE);         // now=2: `cadence > 1` false → falls through
                                                          // acc = 2 - 3 → 18446744073709551615
```

`getAccumulatedTicks(sys)` returns ~1.8e19 and `accumulatedDeltaTime(sys)` ~3.1e17 s — fed straight into the throttled integrator that is this value's entire documented purpose.

Fix — delete the special case; the additive check is already correct for cadence 1:

```cpp
if (now < m_lastRunTick[system] + cadence) {
    return false;
}
```

For an un-offset cadence-1 system `lastRun == now - 1`, so `now < now` is false and it runs every tick — behavior identical, still one add + compare, and the subtraction becomes structurally unreachable unless `now >= lastRun + cadence >= lastRun`. I traced all seven existing tests against this form and they hold (e.g. `DefaultCadenceRunsEveryTick` tick 1: `1 < 0+1` false → runs, `acc == 1`).

**2. `setSystemCadenceOffset` re-phases against a foreign event clock (`engine/system/src/system_manager.cpp:161-174`)**

`m_lastRunTick[system] = maxEventTickCount() + normOffset` mixes clocks. The due check compares against the system's **own** event counter (`now = ++m_eventTickCounts[event]`), but the seed is the max across *all* events. Those diverge by default and permanently:

- Vsync is off by default — `glfwSwapInterval(0)` (`engine/window/src/ir_glfw_window.cpp:117`).
- `World::render()` runs once per outer loop iteration (uncapped), while `World::update()` is pinned to the `shouldUpdate()` fixed step at `kFPS = 60`.
- So `m_eventTickCounts[RENDER]` outruns `[UPDATE]` by (real fps / 60) for the whole process lifetime.

Calling `setSystemCadenceOffset` on an **UPDATE-phase** system therefore seeds `lastRun` far ahead of UPDATE's `now`, and the system silently stops dispatching until UPDATE catches up. A demo rendering ~300fps for 60s: RENDER=18000, UPDATE=3600; `setSystemCadenceOffset(sys, 1)` on a cadence-4 system → `lastRun = 18001`, first fire needs `now >= 18005`, UPDATE climbs at 60/s → **~4 minutes of silent stall**, and it grows with uptime. RENDER-phase systems mostly escape, since RENDER usually *is* the max — so the failure lands squarely on the only phase `accumulatedDeltaTime` supports, and defeats amendment 1's whole purpose (runtime staggering).

Fix: bind the cadence clock to the system's own event. `stampCadenceJoin(event, system)` already has the event — record it (a `std::vector<IRTime::Events> m_cadenceEvent` parallel to the other four, emplaced in `emplaceCadenceState`) and seed from `m_eventTickCounts[m_cadenceEvent[system]]`. `maxEventTickCount()` then has no callers and can be deleted.

**Shared root cause:** cadence state is per-system, but the clock is per-event, and nothing binds a system to its pipeline's clock. Same reason one SystemId listed in two pipelines would thrash a single `m_lastRunTick` between two counters — worth an `IR_ASSERT` in `stampCadenceJoin` (re-stamp with a *different* event) once the event field exists.

### Test gap

No test exercises `setSystemCadenceOffset`'s re-phase after any ticking. `RuntimeSettersNormalizeAndCadenceFromRate` calls it with every counter still at 0 — i.e. `max == UPDATE == 0`, precisely the one case where bug 2 vanishes. Two additions would have caught both:

- Tick `RENDER` k times, then `setSystemCadenceOffset` on an `UPDATE`-phase system; assert it fires within `offset + cadence` UPDATE ticks.
- The finding-1 sequence above; assert `getAccumulatedTicks() == 1` (not `UINT64_MAX`) after the drop to cadence 1.

### On Sonnet's nit — I disagree, dropping it

Sonnet flagged that `IRSystem.setSystemCadence` throws on cadence `0` from Lua while `SystemManager::setSystemCadence` / `emplaceCadenceState` normalize `0 → 1` in C++, asking to align or document the asymmetry. On closer read neither is needed, and I'm dropping the nit. Both sides are already documented on their own terms (`system_manager.hpp:468-475` "0 is normalized to 1"; `engine/script/CLAUDE.md` "n ≥ 1"), and the contracts differ for good reason: Lua is an untyped boundary where a literal `0` is almost certainly a typo and a hard error is the kinder failure, whereas the C++ normalize is what keeps the "unset = unchanged" registration contract airtight (`cadenceOf<T>()` defaults feed the same path). Aligning them necessarily makes one of those worse. No action — don't spend a round-trip on it.

### Confirmed (Sonnet's calls I re-checked)

- **Main-thread due-filtering before fan-out** — Sonnet called this correct and I confirm, with one thing worth stating explicitly since `m_dueScratch` is *new shared mutable state*: its safety rests entirely on `executePipeline` never being re-entered. I verified that holds — the only call sites are `World::input/update/render` and `RenderManager::renderFrame`, and the latter is invoked from `World::render()`, not from a tick body. `IRJob::parallelFor` is blocking, so the `[this]` capture reading `m_dueScratch[k]` cannot observe a clear/refill from another group. Correct today; it would break silently if anything ever dispatched a pipeline from inside a tick.
- **`flushStructuralChanges()` outside the gate** (`system_manager.cpp:521`) — verified, matches the plan's gotcha; deferred changes from other systems keep flushing on skip ticks.
- **Parallel-vector lockstep** — `createSystem` and `createSystemDynamic` are the only two `m_nextSystemId++` sites and both call `emplaceCadenceState`, so the four cadence vectors stay indexed with `m_ticks`. No silent misattribution.
- **Late-join stamping** covers all three join paths, as Sonnet found.
- **Additive due-check** underflow-safety on the `cadence > 1` path, and the `m_eventTickCounts` sizing (`IRTime::Events` is `{UPDATE, RENDER, INPUT, START, END}`, `END + 1 == 5`, indices 0..4 — in range).

jakildev:
Addressed feedback (Opus recheck needs-fix):
- **Finding 1 (underflow):** removed the `cadence > 1` special-case in `pollCadenceDue` — the additive due check now applies uniformly at cadence 1, so a runtime drop to cadence 1 while `lastRunTick` is still ahead of `now` waits for `now` to catch up instead of computing an underflowed `now - lastRunTick`.
- **Finding 2 (cross-event re-phase):** added `m_cadenceEvent` (parallel to the other four cadence vectors), recorded at `stampCadenceJoin`, and pointed `setSystemCadenceOffset`'s re-phase at the system's own event counter instead of `maxEventTickCount()` (deleted, no longer used) — an UPDATE-phase system's offset re-phase no longer seeds from RENDER's uncapped counter.
- **Test gap:** added `CadenceDropToOneDoesNotUnderflow` and `OffsetRephaseUsesOwnEventClock`, covering exactly the two scenarios from the review.

Build clean (`fleet-build --target IrredenEngineTest`); full `SystemCadenceTest` suite (9/9) and the broader system suite (55/55: PipelineGroupsValidatorTest, SystemConcurrencyValidator, RegisterSystemTest, SystemAccessTest, SystemAccessTagFilter, SystemExcludeTest, SystemCadenceTest) all green. — worker

jakildev (commented):
## Opus re-review — system: per-system update cadence in the pipeline scheduler (#2404)

**Verdict:** approve

Both needs-fix findings from my previous recheck are fixed exactly as prescribed at head `6df63b68`, and the two new tests are **real regression gates** — I verified that by re-introducing both bugs in a scratch worktree and watching the new tests fail. The diff is surgical: 3 files, no scope creep, no behavior change beyond the two fixes.

### Verified — I re-ran the gates rather than taking the report on faith

Built `IrredenEngineTest` at `6df63b68` (macOS host) and ran it:

- `SystemCadenceTest.*` — **9/9 green**, matching the author's report.
- `*System*:*Pipeline*` — **124/124 green** across 17 suites. No regression in the pipeline-groups / concurrency-validator / access-tag neighbors.

**The tests are non-vacuous — confirmed by negative control.** A test that passes against the buggy code gates nothing, so I reverted each fix in isolation (restored the `cadence > 1 &&` short-circuit; restored the max-across-all-events re-phase), rebuilt, and re-ran:

- `CadenceDropToOneDoesNotUnderflow` → **FAILS** pre-fix: `exec` is 2 where 0 is expected (the system fires at `now=2` instead of waiting), then 3 where 1 is expected — i.e. it reproduces exactly the fall-through that computes the underflowed `now - lastRunTick`.
- `OffsetRephaseUsesOwnEventClock` → **FAILS** pre-fix: `exec` is 0 where 1 is expected — the UPDATE-phase system stalls behind RENDER's counter, precisely the silent-stall failure I described.

Both pass at head. These two tests would have caught both bugs, which is the whole point.

### Finding 1 (cadence-1 underflow) — fixed correctly

`pollCadenceDue` (`engine/system/src/system_manager.cpp:112-127`) now applies `if (now < m_lastRunTick[system] + cadence)` uniformly with no `cadence > 1` short-circuit. The subtraction on the next line is now structurally unreachable unless `now >= lastRun + cadence >= lastRun`, so the "subtraction is safe" comment is finally guaranteed by the code above it rather than by a case analysis that had a hole in it.

Traced the un-offset cadence-1 path to confirm the claimed no-op: `emplaceCadenceState` normalizes offset to 0 whenever `cadence <= 1`, `stampCadenceJoin` seeds `lastRun = now_join`, so tick `now_join+1` satisfies `now >= lastRun + 1` and fires every tick — identical to the old short-circuit, one add + compare, no cost. The only reachable way to get a cadence-1 system with `lastRun` ahead of `now` is the runtime drop the test now covers, and there the new behavior (wait for `now` to catch up, bounded by the *old* offset — at most `oldCadence - 1` ticks) is both correct and clearly the right trade against a ~1.8e19 accumulated-ticks read.

### Finding 2 (cross-event re-phase) — fixed correctly, and the clock is now genuinely bound

`m_cadenceEvent` is recorded in `stampCadenceJoin` and consumed by `setSystemCadenceOffset`; `maxEventTickCount()` is **fully deleted** (grepped `engine/`, `test/`, `creations/` at head — zero references, so no caller kept the mixed-clock seed alive).

Checked the things that could have gone wrong with a new parallel vector and didn't:

- **Lockstep holds.** `m_cadenceEvent` is emplaced in `emplaceCadenceState` alongside the other four, which remains the single push site reachable from both `createSystem` and `createSystemDynamic` — so it stays indexed with `m_ticks`.
- **Guards are sound.** `stampCadenceJoin` guards on `m_lastRunTick.size()` and then writes `m_cadenceEvent[system]`; `setSystemCadenceOffset` guards on `m_cadenceOffset.size()` and then reads `m_cadence` / `m_lastRunTick` / `m_cadenceEvent`. Both are safe **because** of the lockstep above — the vectors are always equal-length. That's an implicit coupling, not a local property, but it's the same one the pre-existing four vectors already rely on.
- **Index is in range.** `m_cadenceEvent[system]` indexes `m_eventTickCounts` (`std::array<uint64_t, IRTime::END + 1>`); `IRTime::Events` is `{UPDATE, RENDER, INPUT, START, END}` → 0..4, `END + 1 == 5`. In range for every value the field can hold.
- **The UPDATE placeholder is inert, as the comment claims** — a system never listed in any pipeline is unreachable from `executePipeline`, and any later join overwrites the placeholder before it can be read. The setter-before-join ordering is also safe: `stampCadenceJoin` overwrites both `m_cadenceEvent` and `m_lastRunTick`.

I also confirmed the test's `registerPipelineGroups(IRTime::RENDER, {})` genuinely advances RENDER's counter rather than passing vacuously: `executePipeline` early-returns on a missing map key, but `registerPipelineGroups` assigns through `operator[]`, which inserts the (empty) entry — so the 50 RENDER ticks really do reach `m_eventTickCounts[RENDER] = 50`, and the negative control above proves the counter mattered.

### Nits

- `engine/system/src/system_manager.cpp:133-137` — the `IR_ASSERT` I suggested last round (re-stamp of an already-bound system with a *different* event) still isn't there, and now that `m_cadenceEvent` exists it's a two-line guard. Listing one SystemId in two pipelines silently rebinds the cadence clock and thrashes `m_lastRunTick` between two counters that advance at different rates — the same clock-mixing class this PR just fixed, one level up. Latent (no in-tree consumer does this), so it doesn't gate the merge — but it's the natural place to make the new invariant self-enforcing rather than documented. Fine as a follow-up.

### Sonnet's nit — still dropped, no change

Sonnet flagged the Lua-throws-on-0 vs C++-normalizes-0 asymmetry. I dropped that last round (both sides are documented on their own terms and the contracts differ for good reason); the author correctly left it alone. No action.

### Praise

- The fixes are exactly the prescribed ones with **no opportunistic extras** — a 3-file diff that fixes two bugs and adds two tests is the easiest possible re-review, and it's why this closed in one round.
- Writing the two tests to encode the *reproduction* rather than the *fix* is what makes them worth having: both fail loudly against the old code with the exact symptom values from the review (`exec` counts, the underflowed accumulator), so a future refactor that reintroduces either bug gets caught rather than silently re-passing.
- The comments now carry the *reasoning* — "no special-case short-circuit… for an un-offset cadence-1 system `lastRun == now - 1`" and the RENDER-uncapped-vs-UPDATE-fixed-step note on `setSystemCadenceOffset` — so the next reader doesn't have to re-derive why the additive form is uniform or why a foreign clock is unsafe. That's the durable half of this fix.

- [x] `SystemCadenceTest` + system/pipeline suite — I ran both at head myself (9/9 and 124/124 on macOS); no further action.
- [ ] Linux/OpenGL smoke — unchanged from the standing tier rule; this diff is backend-agnostic pure scheduler logic, so it's a formality here.

jakildev:
Addressed the remaining nit from the Opus re-review: `stampCadenceJoin` now asserts (debug-only) when a system is rejoined to a different event's pipeline while still listed in its prior one — the exact clock-mixing footgun amendment 2 fixed, one level up. Guard only fires when the system is genuinely still live in the old pipeline; a legitimate scene-transition re-join (cleared, then re-registered elsewhere) stays silent. Documented the new invariant in `engine/system/CLAUDE.md` alongside the existing double-add assert. Build clean; `SystemCadenceTest` (9/9) + broader system/pipeline suite (124/124) green. — worker

jakildev (commented):
## Re-review — system: per-system update cadence in the pipeline scheduler (#2404)

**Verdict:** needs-fix

Picked up via `human:re-review`. Scoped this pass to the new commit `bbc33631` on top of the already-reviewed head `6df63b68` (both prior Opus needs-fix items were re-reviewed and confirmed fixed by Opus at that head — approve).

### Prior-review resolution
No open items from the prior reviews — both Opus-flagged defects (`pollCadenceDue` underflow, offset re-phase) were already verified fixed at `6df63b68` by the Opus re-review. Nothing to carry forward.

### What changed
New commit adds a defensive assert: joining a system's cadence clock to a different event's pipeline (`stampCadenceJoin`) now asserts if the system is still listed in its *prior* event's pipeline — i.e. catches "live in two pipelines at once, clocks thrash between two counters advancing at different rates." Adds `std::vector<bool> m_cadenceJoined` (lockstep-appended in `emplaceCadenceState`, same two sites — `createSystem`/`createSystemDynamic` — as the other three cadence vectors, so no new lockstep-divergence risk) to distinguish "never joined" (skip the check) from "joined to UPDATE" (the prior placeholder value, which would otherwise collide with a legitimate first join to UPDATE). `engine/system/CLAUDE.md` documents the new invariant.

### Verification performed
Traced every call site of `stampCadenceJoin` (`registerPipelineGroups:257`, `appendToPipeline:277`, `insertSingletonGroupRelativeTo:331`) against the new guard:
- **First join** (`m_cadenceJoined[system] == false`): guard skipped unconditionally — no false assert on a system's very first pipeline join, regardless of event.
- **Re-join to the *same* event** (e.g. `registerPipelineGroups` re-registering event `E`'s own pipeline): `m_systemPipelineGroups[E]` is overwritten with the new groups *before* the join loop runs (`system_manager.cpp:252` before `:257`), and the guard's `m_cadenceEvent[system] != event` is false for a same-event join, so the branch is skipped entirely — never checks the just-replaced map entry against itself. Correct.
- **Re-join to a *different* event after `clearPipeline`**: `clearPipeline(A)` replaces `m_systemPipelineGroups[A]` with an empty group list without calling `stampCadenceJoin` (the loop iterates the *new* — empty — groups, not the old ones), so a later join to event `B` finds `m_systemPipelineGroups[A]` empty and `pipelineGroupsContain` returns false — no assert. Matches the documented "clear the old pipeline first" success path.
- **Re-join to a different event *without* clearing the old one first** (the actual bug this guards against): the old event's map entry still contains the system, so `pipelineGroupsContain` returns true and the assert fires. Confirmed by hand against all three join call sites — none of them remove the system from its old event's groups before calling `stampCadenceJoin`, so this is the only path that can reach the assert, and it's the intended one.
- `SystemId` is never recycled (`m_nextSystemId` is monotonically increasing, no removal path in this file) — so `m_cadenceJoined` can't go stale from an ID reuse across a destroyed/recreated system.

No correctness issues found — the guard is sound against every reachable path.

### Needs-fix
- `test/system/system_cadence_test.cpp` — **no test exercises the new `stampCadenceJoin` guard**, in either direction: no test confirms the assert fires when a system is joined to a new event while still live in the old one, and no test confirms the "clear first, then re-join" path stays silent. This project has an established convention for exactly this class of invariant — `test/system/pipeline_groups_test.cpp` explicitly `EXPECT_THROW`s against `IR_ASSERT`-based pipeline validators (see its header comment: "The validator IR_ASSERTs on conflicts (throws std::runtime_error…), so each rejection case asserts EXPECT_THROW"). `IR_ASSERT` is debug-only (no-op in release per this same PR's own `CLAUDE.md` note two lines above the new addition), so an untested guard silently gives zero coverage in both configurations: it can't fire in release, and nothing proves it fires correctly in debug. This PR's own history (both `6df63b68` fixes were bugs that survived 7/7 green tests because the tests never drove the affected paths) is exactly the failure mode a missing test here would repeat. Suggested fix: add a `SystemCadenceTest` case that registers a system to event `UPDATE`, then calls `appendToPipeline(RENDER, system)` without clearing `UPDATE` first and asserts `EXPECT_THROW`; add a second case that clears `UPDATE` first (or removes the system) and confirms the re-join to `RENDER` is `EXPECT_NO_THROW`.

### Praise
- The `m_cadenceJoined` bit is the right fix for the "placeholder UPDATE vs. real UPDATE join" ambiguity — reusing `m_cadenceEvent[system] != event` alone would have false-asserted on a system's legitimate first join to `UPDATE` (since the placeholder from `emplaceCadenceState` is also `UPDATE`), and the new bit closes exactly that gap without weakening the check for every other event.
- Documenting the invariant in `engine/system/CLAUDE.md` at the exact point future pipeline-join code would look is a good, low-friction way to keep the next reader from re-discovering this by hitting the assert first.

Escalation: this PR already carries `fleet:needs-opus-recheck` context from its history (engine/system core pipeline dispatch — concurrency-sensitive). Sending back to the author for the missing test; re-request review once added, and Opus should still take a final pass on this commit given the surface.

jakildev:
Addressed the re-review's needs-fix (missing coverage for the new `stampCadenceJoin` guard). New commit `2d7468ff`, test-only — no production code touched (`system_manager.cpp` is byte-identical to `bbc33631`).

**Both directions covered, as suggested:**
- `JoinToSecondEventWhileLiveInFirstAsserts` — registers a system to `UPDATE`, then `appendToPipeline(RENDER, sys)` without clearing `UPDATE` first → `EXPECT_THROW(..., std::runtime_error)`. Confirmed the assert fires with the intended message (`system 'TwoPipelinesLive' is still registered in its prior event's pipeline…`), and that `appendToPipeline`'s own already-registered assert doesn't pre-empt it (it checks RENDER's groups, where `sys` is absent — so the throw genuinely comes from `stampCadenceJoin`).
- `RejoinAfterClearingPriorPipelineIsSilent` — `clearPipeline(UPDATE)` first, then re-join to `RENDER` → `EXPECT_NO_THROW`. This pins the guard against being *over-broad*, which the review's own analysis flagged as the live risk (the `m_cadenceJoined` bit exists precisely so a first join to `UPDATE` doesn't collide with the placeholder).

**Negative control — both tests proven non-vacuous.** Per the bar the Opus rounds set here, I reverted the guard in two directions and re-ran rather than assuming the tests gate anything:
- Guard neutralized (condition → `true`): `JoinToSecondEventWhileLiveInFirstAsserts` **FAILS** — *"Expected: … throws std::runtime_error. Actual: it throws nothing."* `RejoinAfterClearingPriorPipelineIsSilent` still passes, correctly — it's the opposite-direction gate.
- Guard made over-broad (assert on *any* different-event re-join, dropping the still-live check): `RejoinAfterClearingPriorPipelineIsSilent` **FAILS** — *"Expected: … doesn't throw. Actual: it throws std::runtime_error."* The throw test still passes.

Each test fails against exactly one defect and neither is redundant with the other.

`EXPECT_THROW` against `engAssert`'s `std::runtime_error` follows the convention cited in the review — verified `test/system/pipeline_groups_test.cpp:19` documents it and lines 182/219 use it.

**Gates:** build clean (`fleet-build --target IrredenEngineTest`, macOS). `SystemCadenceTest` **11/11** green (9 + 2 new). `*System*:*Pipeline*` **126/126** green across 17 suites — the prior 124 baseline plus the 2 additions, no regressions. `format-changed` left the diff untouched (already conformant).

On the review's note that Opus should take a final pass given the surface — agreed; the diff is test-only, but the guard it covers is core pipeline dispatch. — worker

jakildev (commented):
## Re-review — system: per-system update cadence in the pipeline scheduler (#2404)

**Verdict:** approve

Picked up via `fleet:changes-made`. Scoped this pass to the new commit `2d7468ff` on top of the already-reviewed head `bbc33631` — this addresses the missing-test-coverage `needs-fix` from my prior re-review (submitted 2026-07-15T22:30:51Z).

### Prior-review resolution
- ✅ Missing coverage for the `stampCadenceJoin` two-pipeline guard — fixed at `2d7468ff`. Both directions are now covered: a system still live in `UPDATE`'s pipeline asserts when joined to `RENDER` (`JoinToSecondEventWhileLiveInFirstAsserts`), and a system that clears `UPDATE` first re-joins `RENDER` silently (`RejoinAfterClearingPriorPipelineIsSilent`).

### What changed
Test-only: `test/system/system_cadence_test.cpp` gains the two tests above (47 lines). Diffed `bbc33631..2d7468ff` — `engine/system/src/system_manager.cpp` is byte-identical, confirming the author's "no production code touched" claim.

### Verification performed
- Confirmed the guard being tested matches the assert exactly: `git show bbc33631:engine/system/src/system_manager.cpp` — `stampCadenceJoin` (line 149) asserts `it == m_systemPipelineGroups.end() || !pipelineGroupsContain(it->second, system)` with the message the new test's negative case should trip.
- Built `IrredenEngineTest` at `2d7468ff` (macOS host) and ran it myself rather than trusting the reported counts:
  - `SystemCadenceTest.*` — **11/11 green**. Watched `JoinToSecondEventWhileLiveInFirstAsserts` actually trigger the assert (`ASSERTION: it == m_systemPipelineGroups.end() || !pipelineGroupsContain(...)`, exact message: "system 'TwoPipelinesLive' is still registered in its prior event's pipeline...") and `EXPECT_THROW` catch it; `RejoinAfterClearingPriorPipelineIsSilent` completed with no assert fired.
  - `*System*:*Pipeline*` — **126/126 green** across 17 suites, matching the prior 124-test baseline plus these 2 additions — no regressions.
- Didn't re-run the author's negative-control revert (guard-neutralized / guard-made-over-broad) myself, but the description of which test catches which defect is internally consistent with the assert's actual condition (`m_cadenceJoined[system] && m_cadenceEvent[system] != event` — dropping either half maps exactly to the two failure directions described).

No correctness issues. This closes out the last open item from the review history on this PR.

### Nits
None.

### Praise
- The negative-control methodology (revert each direction, confirm the *other* test still passes, confirm the target test fails against exactly one defect) is the right bar for proving a regression test is non-vacuous, and it's applied correctly here — each new test's failure mode maps to a distinct half of the guard's condition, not to the same defect twice.

- [x] Already covered by my own independent run above (11/11, 126/126) — no further action needed.

Escalation: this PR's history already flagged `fleet:needs-opus-recheck` given the core pipeline-dispatch surface; this commit is test-only but touches the same concurrency-sensitive area, so Opus should take a final pass per the standing rule on this PR.

jakildev (commented):
## Opus recheck — system: per-system update cadence in the pipeline scheduler (#2404)

**Verdict:** approve

Final pass, scoped to `6df63b68..2d7468ff` — the two commits added since my prior Opus re-review approved head `6df63b68` (the `stampCadenceJoin` two-pipeline assert + doc, and its test coverage). Both prior Opus needs-fix items stay fixed; nothing to carry forward.

The guard is **correct and cannot false-fire on any in-tree usage** — I verified that rather than reasoning it through. Two nits below, neither blocking.

### Verified — re-ran the gates rather than taking the report on faith

Built `IrredenEngineTest` at `2d7468ff` (macOS) and ran it:

- `SystemCadenceTest.*` — **11/11 green**, assert firing with the exact message.
- `*System*:*Pipeline*` — **126/126 green** across 17 suites.

Both match the author's and Sonnet's reported numbers exactly.

**The guard test is non-vacuous — confirmed by negative control.** Removing the guard block entirely fails exactly `JoinToSecondEventWhileLiveInFirstAsserts` while `RejoinAfterClearingPriorPipelineIsSilent` still passes, so each test maps to a real defect rather than the same one twice.

**The assert cannot false-fire on existing code** — the question I most wanted answered, since it throws for *any* cadence, including default cadence-1 systems that never opted into the feature:
- **C++ creations can't reach it.** `createSystem<X>()` mints a fresh `SystemId` per call (`m_nextSystemId++`, exactly two sites), so the same system *type* registered into two events gets two *distinct* ids. Only a reused id value can trip the guard.
- **Lua can reach it** (`IRSystem.systemId(name)` returns a stable pre-registered id), so I scanned all 45 Lua creations for a `SystemName` appearing in 2+ event pipelines: **zero hits**. Nothing in-tree registers one system into two events.
- **Lockstep holds.** Both id-minting sites (`system_manager.hpp:105`, `system_manager.cpp:58`) call `emplaceCadenceState`, so `m_cadenceJoined` can't desync from the other four cadence vectors — the `system >= m_lastRunTick.size()` guard legitimately covers the `m_cadenceJoined[system]` index even though it bounds-checks a different vector.

Liveness-checking the prior pipeline's *contents* (rather than comparing `m_cadenceEvent` alone) is the right design: it means a `clearPipeline` + re-join stays silent, which is exactly what the second test pins.

### Nits (non-blocking)

**1. `m_cadenceJoined` is redundant — it can only ever *suppress* the guard, never trigger it (`system_manager.cpp:153`, `system_manager.hpp:388-393`)**

Sonnet praised this bit as necessary, reasoning that `m_cadenceEvent[system] != event` alone "would have false-asserted on a system's legitimate first join to `UPDATE` (since the placeholder from `emplaceCadenceState` is also `UPDATE`)". **On closer read I disagree — that rationale is inverted.** If the placeholder is `UPDATE` and the first join is *to* `UPDATE`, then `m_cadenceEvent[system] != event` is **false**, so the check is already skipped; there is no false-assert to prevent. And a first join to a *different* event can't false-assert either, because a never-joined system isn't listed in `UPDATE`'s pipeline, so `pipelineGroupsContain` already returns false.

The liveness check subsumes the bit entirely: every path into a pipeline (`registerPipelineGroups`, `appendToPipeline`, `insertSingletonGroupRelativeTo`) calls `stampCadenceJoin`, so "in the placeholder event's pipeline but never joined" is unreachable.

I confirmed this empirically — dropping the `m_cadenceJoined[system] &&` half of the condition leaves the **full 1331-test non-MIDI suite green**, so no test anywhere covers the bit. Sonnet's claim that "dropping either half maps exactly to the two failure directions" holds for the liveness half but not this one.

Worth noting the direction of the risk: because the bit gates the check, its only possible effect is to *skip* an assert that would otherwise fire. If a future pipeline-insertion path ever skipped `stampCadenceJoin`, `joined == false` would suppress the correct assert rather than catch it. Consider dropping the vector and its header comment block; if you keep it, please fix the stated rationale, since as written it will mislead the next reader.

**2. `it` is unused under `IR_RELEASE` (`system_manager.cpp:160`)**

`IR_ASSERT` expands to nothing in `IR_RELEASE` (the condition is not evaluated), so `const auto it = m_systemPipelineGroups.find(...)` — hoisted out of the assert because it's used twice — becomes a set-but-unused local. I reproduced the resulting `-Wunused-variable` under the release expansion.

Cosmetic in practice, which is why it's a nit and not needs-fix: the build sets no `-Wall`/`-Wextra` (only `-O3`), there's no Release preset (all three are Debug), and `.clang-tidy` runs `Checks: -*` plus three named checks with `WarningsAsErrors: ''` — so nothing surfaces it today. Cleanest fix is a small pure helper (`isStillLiveInPriorPipeline(system)`) called from the assert condition, which also respects the documented "no side-effectful calls in the condition" rule since the helper is a pure lookup.

### Praise

- Gating on the prior pipeline's **live contents** rather than on a stale `m_cadenceEvent` comparison is the detail that makes this guard usable — a scene transition that clears and re-registers stays silent, so the assert flags only the genuine two-pipelines-at-once case instead of every legitimate re-join.
- The assert message names the actual failure mechanism (clock thrash between counters advancing at different rates) and the remedy, rather than just restating the predicate — that's what makes it worth the debug-only cost.
- Documenting the invariant in `engine/system/CLAUDE.md` next to the three existing pipeline asserts puts it exactly where the next person writing join code will look.

### Test plan before merge

- [x] `SystemCadenceTest` + system/pipeline suites — I ran both at head myself (11/11, 126/126 on macOS), plus the full 1331-test non-MIDI suite. No further action.
- [ ] Linux/OpenGL smoke — formality; this diff is backend-agnostic pure scheduler logic.

jakildev:
Addressed both nits from the final Opus recheck:
- **Nit 1** (`m_cadenceJoined` redundant): agreed with the Opus analysis — the liveness check (`isStillLiveInPriorPipeline`) already subsumes the bit; dropping `m_cadenceJoined[system] &&` from the guard condition leaves the full non-MIDI suite green. Removed the vector, its emplace, and its header doc bullet.
- **Nit 2** (`it` unused under `IR_RELEASE`): extracted the find()-then-check lookup into a small pure helper, `SystemManager::isStillLiveInPriorPipeline(priorEvent, system)`, called directly from the `IR_ASSERT` condition — no local hoisted out of the macro to go unused when it expands to nothing.

Build clean (`fleet-build --target IrredenEngineTest`, macOS); `SystemCadenceTest` 11/11, `*System*:*Pipeline*` 126/126, full non-MIDI suite 1324/1324 — all green, no regressions. `format-changed` left the diff untouched (already conformant). — worker

jakildev (commented):
## Re-review — system: per-system update cadence in the pipeline scheduler (#2404)

**Verdict:** approve

Picked up via `human:re-review`. Scoped this pass to the new commit `45d506f3` on top of the already-Opus-approved head `2d7468ff`, which addresses both nits from Opus's final recheck (submitted 2026-07-15T23:10:58Z).

### Prior-review resolution
- ✅ Nit 1 (`m_cadenceJoined` redundant with `isStillLiveInPriorPipeline`) — fixed. The bit vector, its `emplaceCadenceState` push, and its header doc bullet are all removed.
- ✅ Nit 2 (`it` unused under `IR_RELEASE`) — fixed. The find()-then-check lookup is now a named helper, `SystemManager::isStillLiveInPriorPipeline(priorEvent, system)`, called directly from the `IR_ASSERT` condition.

Nothing else open from the review history on this PR.

### What changed
`system_manager.hpp`/`system_manager.cpp` only (14/24 lines, per `git diff --stat 2d7468ff..45d506f3`) — no test file touched, consistent with a pure refactor. `stampCadenceJoin`'s guard condition changed from `m_cadenceJoined[system] && m_cadenceEvent[system] != event` to `m_cadenceEvent[system] != event`.

### Verification performed
- Hand-traced the collapsed condition for a false-positive risk on a genuine first join to a non-UPDATE event (the case the old `m_cadenceJoined` bit specifically existed to avoid asserting on): `m_cadenceEvent[system]` defaults to the inert `UPDATE` placeholder from `emplaceCadenceState` for any never-joined system, and `isStillLiveInPriorPipeline` is a pure structural lookup against `m_systemPipelineGroups` — a system that has never joined anything isn't listed in any pipeline group, so the lookup reads false regardless of which event it's about to join. The old bit and the new lookup are behaviorally equivalent on every path the existing suite exercises; the removal is a real simplification, not a semantic change.
- Grepped the tree for stray references to the removed `m_cadenceJoined` field and confirmed none remain outside this PR's own diff (checked `engine/`, and `engine/system/CLAUDE.md`'s doc bullet for `stampCadenceJoin` doesn't mention it either — no stale doc drift).
- No new tests needed: the two assert-direction tests added in the prior commit (`JoinToSecondEventWhileLiveInFirstAsserts`, `RejoinAfterClearingPriorPipelineIsSilent`) already pin both directions of this guard and remain valid coverage for the simplified condition. Author reports 11/11 `SystemCadenceTest` + 126/126 pipeline suite + 1324/1324 full non-MIDI suite green at this head — consistent with a behavior-preserving refactor.

No correctness issues found.

### Nits
None.

### Praise
- Extracting the lookup into a named, pure `const` helper and asserting on it directly is exactly the fix Opus's nit called for — no local variable left to go unused when `IR_ASSERT` compiles out under `IR_RELEASE`.

Escalation: this PR's history already flagged `fleet:needs-opus-recheck` given the core pipeline-dispatch surface (`engine/system/`, concurrency-sensitive). This commit is a small, behavior-preserving cleanup of code Opus already approved and specifically requested — should be a quick confirmation pass rather than a full re-derivation.

Opus recheck required: touches engine/system/ core pipeline dispatch (concurrency-sensitive), per the standing rule on this PR.

jakildev (commented):
## Opus recheck — system: per-system update cadence in the pipeline scheduler (#2404)

**Verdict:** approve

Final confirmation pass, scoped to `2d7468ff..45d506f3` — the single commit added since my prior Opus recheck, which addresses both nits from that pass (submitted 2026-07-15T23:10:58Z). Both are fixed exactly as prescribed:

- ✅ Nit 1 (`m_cadenceJoined` redundant with the live-pipeline lookup) — the bit vector, its `emplaceCadenceState` push, and its header doc bullet are gone. `git grep m_cadenceJoined` at head returns nothing, so no orphaned lockstep vector is left behind.
- ✅ Nit 2 (`it` unused under `IR_RELEASE`) — the find()-then-check is now the named `const` helper `isStillLiveInPriorPipeline`, called directly from the `IR_ASSERT` condition. No local survives the macro compiling out, and a member function with external linkage won't warn when unused.

**Sonnet flagged this as a "quick confirmation pass" on a behavior-preserving refactor; on closer read I agree with the verdict but not with that framing** — and the reason is worth recording, because Sonnet's *own earlier* re-review (2026-07-15T22:30:51Z) praised `m_cadenceJoined` as "the right fix" that prevents the guard from "false-asserting on a system's legitimate first join to `UPDATE`." This commit deletes that bit. If that earlier reasoning had been right, this refactor would be a **regression that false-asserts on every render-only system**. Two reviews on this PR now assert opposite things about the same bit, so I verified empirically rather than re-deriving.

### Verified — the redundancy claim holds, proven three ways

**1. Call-site audit.** All three join paths — `registerPipelineGroups` (and its `registerPipeline`/`clearPipeline` delegates), `appendToPipeline`, `insertSingletonGroupRelativeTo` — call `stampCadenceJoin` **after** inserting into `event`'s groups, and the guard reads the **prior** event (`m_cadenceEvent[system]`), which the block only enters when `prior != event`. So the just-performed insertion can never self-trigger the lookup. The only state where old and new diverge is "listed in UPDATE's pipeline but never stamped" — unreachable, because every path that lists a system also stamps it, and the `system >= m_lastRunTick.size()` early-return covers the out-of-range case.

**2. The case the bit claimed to protect was untested — so I tested it.** Every one of the 11 tests registers to `UPDATE` first; none covers a system whose *first-ever* join is to a non-`UPDATE` event, which is exactly the scenario `m_cadenceJoined` was introduced for. I wrote two scratch tests against head:
- first-ever join straight to `RENDER`, no `UPDATE` registration at all (map-miss arm) — **silent**.
- same, but `UPDATE`'s pipeline exists and lists a *different* system, so `find(UPDATE)` succeeds and the guard falls through to the `pipelineGroupsContain` arm rather than the map-miss shortcut — **silent**.

Both pass. `m_cadenceEvent[system] != event` alone already covers the first-join-to-`UPDATE` case (placeholder equals the event, block skipped), and a never-joined system is in no pipeline, so the lookup is false. The bit was genuinely redundant; the earlier praise for it was mistaken.

**3. The guard is still non-vacuous under its new shape — negative control.** A condition that changed shape needs its control re-run, so I stubbed `isStillLiveInPriorPipeline` to `return false`. That fails **exactly** `JoinToSecondEventWhileLiveInFirstAsserts` and nothing else — the silent-path tests (including both scratch ones) still pass, as they should. The test pins precisely this guard, not something adjacent.

**Gates re-run at `45d506f3` (macOS), not taken on faith:**
- `SystemCadenceTest.*` — **11/11 green**
- `*System*:*Pipeline*` — **126/126 green** across 17 suites

Both match the author's and Sonnet's reported numbers exactly.

### Nits

- `test/system/system_cadence_test.cpp` — **no test covers a first-ever join to a non-`UPDATE` event.** Non-blocking: I proved the behavior is correct today. But the coverage gap sits exactly where this refactor was most likely to break — the one case the deleted bit existed for is the one case nothing gates. A render-only system (first join → `RENDER`) is a legitimate, likely-future usage pattern, and if a later change tightens the guard, it would false-assert on every such system with no test to catch it. Consider landing my scratch test as a follow-up:

  ```cpp
  // First-ever join to a non-UPDATE event: m_cadenceEvent holds the inert
  // UPDATE placeholder, so the guard's `!= event` arm is true — the lookup
  // is what must keep it silent.
  TEST_F(SystemCadenceTest, FirstEverJoinToRenderIsSilent) {
      auto sys = IRSystem::createSystem<C_CadA>(
          "FirstJoinRender", [](C_CadA &) {}, nullptr, nullptr, {}, nullptr,
          IRSystem::Concurrency::SERIAL, IRSystem::kDefaultGrainSize, 2, 0
      );
      // No UPDATE registration whatsoever — RENDER is this system's first join.
      EXPECT_NO_THROW(m_system_manager.appendToPipeline(IRTime::RENDER, sys));
  }
  ```
  The foreign-`UPDATE`-pipeline variant is the more valuable of the two, since it exercises the `pipelineGroupsContain` arm rather than the map-miss shortcut.

### Praise

- Deleting the bit rather than keeping it "just in case" is the right call, and the replacement comment carries the *reasoning* — "A never-joined system isn't listed in any pipeline yet, so `isStillLiveInPriorPipeline` is false and this never false-fires on a genuine first join." That sentence is what let me check the claim instead of re-deriving it, and it's the durable half of this change.
- Deriving the guard from the prior pipeline's **live contents** rather than from a cached "have I joined" bit is what makes it correct by construction: there's now one source of truth (the pipeline map) instead of two that could drift. The dropped `std::vector<bool>` also removes a proxy-reference specialization from the lockstep vector set.

### Test plan before merge

- [x] `SystemCadenceTest` + system/pipeline suites — I ran both at head myself (11/11, 126/126 on macOS), plus the scratch coverage and negative control above. No further action.
- [ ] Linux/OpenGL smoke — formality; this diff is backend-agnostic pure scheduler logic.

jakildev:
Addressed the nit from the final Opus recheck (coverage gap: no test covers a **first-ever join to a non-`UPDATE` event**). New commit `1a90ec5d`, test-only — `system_manager.cpp` and the whole of `engine/system/include/` are byte-identical to `45d506f3` (verified with `git diff 45d506f3..HEAD --`, empty).

**Both arms landed, not just the suggested one.** The review noted the foreign-`UPDATE`-pipeline variant is the more valuable of the two, so both are in rather than the map-miss one alone:
- `FirstEverJoinToRenderIsSilent` — no `UPDATE` registration at all; the prior-pipeline lookup returns via the **map-miss** shortcut.
- `FirstEverJoinToRenderIsSilentWhenUpdateHostsAnotherSystem` — `UPDATE`'s pipeline exists but hosts an unrelated system, so `find(UPDATE)` succeeds and the guard falls through to the **`pipelineGroupsContain`** arm.

**Negative control — and the second one is what justifies the pair.** Per the bar this thread has set, I didn't assume the new `EXPECT_NO_THROW` tests gate anything:

- **Control 1 — over-broad guard** (`isStillLiveInPriorPipeline` stubbed to `return true`): both new tests **FAIL**, so neither is vacuous. But `RejoinAfterClearingPriorPipelineIsSilent` fails too — so this control does *not* show the additions cover anything the suite lacked.
- **Control 2 — existence-not-membership** (`return it != end() && !it->second.empty()`, i.e. the realistic "later change tightens the guard" that checks the prior pipeline *exists* rather than *contains me*): the **entire pre-existing 11-test suite passes** — including `RejoinAfterClearingPriorPipelineIsSilent` (its `clearPipeline` leaves an *empty* entry) and `JoinToSecondEventWhileLiveInFirstAsserts` — and **only `FirstEverJoinToRenderIsSilentWhenUpdateHostsAnotherSystem` catches it.**

Control 2 is the empirical version of the review's own call that the foreign-`UPDATE` variant is the more valuable one: it is the sole gate distinguishing *identity* from mere *existence* in the prior pipeline. The guard was restored byte-identical after each control.

**Gates (macOS, at `1a90ec5d`):**
- `SystemCadenceTest` — **13/13** green (11 + 2).
- `*System*:*Pipeline*` — **128/128** green across 17 suites (your 126 baseline + 2, no regressions).
- Full non-MIDI suite — **1333/1333** green, 0 failures, with filter `-*Midi*:*MIDI*`. That total differs from the 1324 reported in earlier rounds because the filter string differs, not because tests moved — nothing failed, and my delta is exactly +2.
- `format-changed` left the diff untouched (already conformant).

Left `fleet:approved` in place per the has-nits path (a test-only cleanup doesn't invalidate the approval); the recheck automation will re-tag on the new commit as usual. — worker

jakildev (commented):
## Review — system: per-system update cadence in the pipeline scheduler (#2404)

**Verdict:** approve

Picked up via `human:re-review`. Scoped this pass to the new commit `1a90ec5d` on top of the already-Opus-approved head `45d506f3` — this addresses the sole open item from Opus's final recheck (submitted 2026-07-15T23:29:03Z): missing coverage for a system's first-ever join to a non-`UPDATE` event.

### Prior-review resolution
- ✅ Coverage gap (`stampCadenceJoin` guard untested for a first-ever join to a non-`UPDATE` event) — fixed at `1a90ec5d`. Both arms Opus called out are covered: `FirstEverJoinToRenderIsSilent` (no `UPDATE` registration at all — map-miss arm) and `FirstEverJoinToRenderIsSilentWhenUpdateHostsAnotherSystem` (`UPDATE`'s pipeline exists but lists a different system — `pipelineGroupsContain` arm). The author landed both arms Opus's suggestion enumerated, not just the single scratch test offered.

Nothing else open from the review history on this PR.

### What changed
Test-only: `git diff 45d506f3..1a90ec5d -- engine/system/` is empty — confirmed `system_manager.cpp`/`.hpp` untouched. `test/system/system_cadence_test.cpp` gains the two tests above (62 lines).

### Verification performed
Traced both new tests against `stampCadenceJoin`/`isStillLiveInPriorPipeline` (`engine/system/src/system_manager.cpp:148-176`) by hand rather than taking the claim on faith:
- `emplaceCadenceState` seeds `m_cadenceEvent[system] = IRTime::UPDATE` as an inert placeholder (`system_manager.cpp:121-124`). A never-joined system's first join to `RENDER` hits `m_cadenceEvent[system] != event` (placeholder `UPDATE` ≠ `RENDER`), so the guard evaluates `isStillLiveInPriorPipeline(UPDATE, system)`.
- Test 1: no `UPDATE` pipeline exists at all → `m_systemPipelineGroups.find(UPDATE)` misses → `isStillLiveInPriorPipeline` returns `false` via the map-miss short-circuit → no assert. Matches the test's own comment.
- Test 2: `UPDATE`'s pipeline exists (populated with an unrelated `resident` system via `registerPipeline`) → `find(UPDATE)` succeeds → falls through to `pipelineGroupsContain(it->second, sys)`, which must report `sys` absent since it was never joined anywhere → no assert. This is the arm Test 1 can't reach (map-miss short-circuits before `pipelineGroupsContain` runs), so the two tests are non-redundant — each exercises a distinct branch of `isStillLiveInPriorPipeline`.
- Both `C_CadA`/`C_CadB` component structs and the `createSystem` call shape used in the new tests already exist earlier in the same file (lines 23-26, 176+) — no compile risk from an unseen symbol.

Author-reported gates at this head (macOS): `SystemCadenceTest` 13/13, `*System*:*Pipeline*` 128/128 across 17 suites — consistent with the 11/126 baseline plus the two additions, no regressions claimed or expected since production code is unchanged.

No correctness issues found.

### Nits
None.

### Praise
- Landing both arms Opus's nit enumerated (map-miss AND `pipelineGroupsContain`-fallthrough) rather than just the one scratch test Opus offered inline closes the gap Opus itself flagged as "sits exactly where this refactor was most likely to break" — a render-only system's first join is now a real regression gate, not just a hand-verified claim.

- [x] Already covered by the author's own reported run (13/13 `SystemCadenceTest`, 128/128 pipeline suite) — no further action needed.

This closes out the last open item from the review history on this PR. Given the change is test-only, implements exactly what Opus's own final recheck suggested, and I traced both new tests against the guard logic by hand (see above), a further Opus pass would be re-verifying an already-approved implementation against tests that add coverage, not risk — I don't think it's warranted here.

jakildev:
**Coverage request before merge (posted from the game-side architect session, human-directed).**

The SystemManager gate is well covered — 13 tests in `test/system/system_cadence_test.cpp`, including the review-amendment regressions (underflow on cadence-drop, own-event-clock re-phase, double-pipeline asserts). But two of this PR's **new public surfaces ship untested**:

1. **The Lua surface** (`lua_pipeline_bindings.hpp`): `IRSystem.setSystemCadence` / `getSystemCadence` / `setSystemCadenceOffset` / `getSystemCadenceOffset` / `getAccumulatedTicks` / `accumulatedDeltaTime`. `test/script/` has 19 `lua_*_test.cpp` files as precedent — `lua_pipeline_register_test.cpp` is the natural home. Downstream creations will drive cadence *policy from Lua*, so this is the seam where a regression lands first. Suggested minimum: set/get round-trip through Lua, a Lua-registered system throttled via `IRSystem.setSystemCadence` fires 1-in-N with `getAccumulatedTicks` readable from its tick, and the two throw paths (`cadence < 1`, `offset < 0`).
2. **The `System<N>` spec-member path** (`ir_system.hpp` `cadenceOf` / `cadenceOffsetOf` detectors): every cadence test constructs via the `createSystem` free-function trailing params; nothing registers a spec declaring `static constexpr std::uint32_t kCadence` / `kCadenceOffset` through `registerSystem<N>` and asserts the detected values (plus the absent → 1/0 default). One small test closes the detector path.

Per the review checklist (`review-pr` SKILL.md §Tests/build) the per-PR letter is satisfied — tests exist — but these are new public API surfaces and the repo's own `test/script/` convention says the Lua seam gets covered. Requesting both before merge; a checklist-sharpening issue is being filed separately so this class of gap gets caught mechanically.

Issue #2428, opened 2026-07-15
render: per-axis analytic-coverage phase disagreement on fractional-position shared edges — 1px dashed fringe + bottom-vertex notch (gallery row B)


jakildev:
## Symptom

On FRACTIONALLY-positioned voxel entities under residual yaw (per-axis scatter path), the shared edges between visible faces show a 1px dashed checkerboard fringe, and the bottom vertex where two side faces meet shows a 1-2px notch (the faces' silhouette rows end at different heights). Canonical repro: `IRPerfGrid --mode gallery --auto-screenshot` → `gal_yaw020`, row-B cube4 (fractional offset 0.37/0.13/+0.5): the top-right shared edge carries the fringe ("detached sliver" + dark gap), the bottom vertex the notch. The INTEGER row-A twin is clean at the same pose. macOS/Metal (the #1937 analytic-coverage lead backend).

## Evidence / what it is NOT

- `--debug-overlay peraxis_id` at the same pose shows the fringe pixels as single-pixel Z-canvas (blue) vs X-canvas (red) interleaving along the shared edge — cross-axis per-pixel ownership alternation.
- It is NOT the #2411 tie-band class: the fringe is byte-identical before/after the #2411 priority-major sub-band reorder (which did fix scattered tie-strip pixels elsewhere in the same scene — img_diff confirmed the drift set).
- It is NOT the raster-dropout class #1937 already fixed on Metal (those pixels are lit, not background... except the dark-gap component, which IS unwritten).
- Both facing edges at the shared edge should classify INTERIOR (rule 2, `visiblePolarityForAxis`) → both fill, margin-yield arbitrates → no gap should be possible. The observed gap+interleave therefore points at the per-quad ANALYTIC COVERAGE 0.5-crossings (`scatterAnalyticEdgeCoverage`, fwidth-based) landing at different sub-pixel phases for the two adjacent faces when fracs shift the shared edge off the lattice — mutual-trim strips (dark dashes) alternating with double-claim strips per pixel row.

## Investigation direction (confirm during work)

1. Instrument: dump `edgeInterior` for the two face quads at the repro edge (confirm rule-2 classification actually fires for both under fracs).
2. If classification is correct, the fix is in the coverage geometry: adjacent faces sharing an edge must derive their cut from the SAME projected line with the SAME rounding — e.g., boundary-edge coverage evaluated in shared screen space rather than per-quad footprint parameters, or interior-edge fill extended to cover the mutual-trim strip.
[...]

Issue #2427, opened 2026-07-15
render: voxel cylinder yaw-sweep centroid JITTER on master (reversals=9, 2.9px residual, macOS/Metal zoom4)


jakildev:
## Symptom

The documented rotation-jitter gate fails on current master (4cef76bb, macos-debug):

```
IRShapeDebug --spin-shape cylinder --spin-shape-voxel --yaw-sweep --zoom 4 --auto-screenshot 6
jitter_probe: frames=24 (valid=24)  verdict=JITTER
  x: reversals=9  max_residual=2.93px  delta_std=1.59  delta_max=5.37
  y: reversals=0  max_residual=0.36px
```

The pan-sweep twin (`--spin-shape box --spin-shape-voxel --pan-sweep --yaw 0.785 --zoom 4`) is SMOOTH, and y is clean — the wobble is x-only under yaw stepping. A vertical voxel cylinder's silhouette is Z-yaw-invariant, so any centroid reversal is jitter, not a face-triplet change (the reason this probe shape is canonical — engine/render/CLAUDE.md § temporal stability).

Reproduced byte-identically on master and on the #2411 tie-reorder branch (identical metrics), so it is not tie-arbitration related. Candidate class: the per-axis scatter's integer-anchor vs sub-pixel-compensation split under yaw (#1944 was the pan-side fix; this may be its yaw-side residual), or the wFrac reconstruction's per-bracket rounding (#2418).

## Acceptance criteria

1. The exact repro above flips JITTER -> SMOOTH (reversals=0, residual <= 1.5px) at zoom 2/4/8.
2. Pan-sweep stays SMOOTH; cardinal byte-identity preserved (shape_debug suite pairwise img_diff = 0 at yaw 0).
3. Clean exits (ir-run RESULT=CLEAN).

## Context

Found while running the #2411 acceptance matrix (gate 6). Not blocking that PR (pre-existing, byte-identical on both builds).

jakildev:
Planning gate: this issue has no `## Plan` comment (the canonical plan per #1932) and no committed/staged plan file (`.fleet/plans/issue-2427.md`), and it is not opted out (`human:no-plan` / `fleet:no-plan` / `[no-plan]` / investigation spike). Unplanned carve-offs design-block on claim (#1456), so adding `fleet:needs-plan` to route it through PLANNING-PROTOCOL.md before it can queue. — fleet-queue-ingest

jakildev:
[...]

Issue #2426, opened 2026-07-15
test: shape_debug macos-debug render-verify references stale on master (21/24 FAIL) — refresh post-#2410/#2418 [no-plan]


jakildev:
Blocked by: #2449

## Symptom

`python3 scripts/render-verify.py --target IRShapeDebug` on current master (4cef76bb, macos-debug/Metal) fails 21 of 24 checks — including cardinal-0 shots (zoom1_origin 99.664% match). The committed `creations/demos/shape_debug/test/references/macos-debug/` set predates the #2410 (per-axis frac lighting recovery + occupancy stamp) and #2418 (wFrac carrier) merges, both of which intentionally changed rotated/lit output.

Verified NOT a live regression: two independent rebuilds of master produce identical failure tables, and the #2411 tie-reorder branch's captures are byte-identical to master's on all 28 shots (pairwise img_diff drift=0).

## Fix

`python3 scripts/render-verify.py --target IRShapeDebug --update-references` on a macos-debug host, inspect the new PNGs against the #2410/#2418 PR-body evidence (intentional-drift classes only), commit. Analog of #2000 / #1969 / #2344 for this demo+preset.

## Context

Found during #2411 gate verification (the tie-reorder PR used pairwise master-vs-branch diffs instead, so it is not blocked on this).

jakildev:
Attempted on a macOS host (worker-3, macos-debug/Metal). This task requires building `IRShapeDebug` on macOS to regenerate the `macos-debug/` references — but the macOS build is currently broken by **#2449** (Apple clang-21 + fmt 10.1.1 consteval incompatibility). Incremental `fleet-build --target IRShapeDebug` fails to compile fmt itself:

```
_deps/fmt-src/include/fmt/format-inl.h:61:19: error: call to consteval function
  'fmt::basic_format_string<char, fmt::basic_string_view<char> &, const char (&)[3]>::basic_format_string<FMT_COMPILE_STRING, 0>'
  is not a constant expression
```
(5 such errors across format-inl.h). This is exactly #2449.

Only a macOS host can produce `macos-debug` references, and no macOS host can build until #2449 (bump fmt → 10.2.1) lands. Recorded `Blocked by: #2449` in the body so the scout parks this until the build fix merges, at which point it auto-re-queues. Released the claim. — worker (worker-3)

PR #2429, opened 2026-07-15
render: per-axis scatter tie order — priority-major sub-band, band 8→16

### LINKED ISSUE #2411: render: per-axis scatter equal-depth tie resolution alternates across axis canvases (moving contiguous content stipples under yaw)
## Symptom

With an ACTIVE per-cell wave (the IRPerfGrid default: contiguous per-cell voxel sets, amplitude 5), rotating the camera to a residual yaw still shows the per-voxel chevron shading stipple that PR #2410 fixed for static scenes. Repro: `IRPerfGrid --grid-size 16 --zoom 8 --auto-screenshot` → zoom1_rot/zoom4_rot shots.

## Why #2410 couldn't cover it

A moving per-cell wave legitimately part-uncovers faces (adjacent cells shear by up to one rounded step at tear-free amplitude), so the cross-set occupancy stamp cannot be applied — covered/partially-covered faces MUST be emitted or real holes open. Those emitted quads then tie with the true surface quads at exact equal depths in the per-axis forward-scatter composite:

- All faces of one voxel share their lattice origin → identical `rawDepth`, identical frac → their planar depth fields cross at exactly-equal values along shared edges.
- `f_peraxis_scatter` floor-quantizes every fragment depth to `kScatterCellTieBand` and replaces the sub-band bits with the #2255 cell-identity code — which erases the `(flip<<2)|slot` tiebreak the vertex stage folded into the continuous key.
- For cross-axis ties the cell code is semantically arbitrary → the winner alternates by cell parity → per-cell Lambert alternation (the stipple). Same-voxel three-face ties share a cell code and fall back to draw order (axis-canvas order), which is stable but unspecified.

## Design question

Within the fp32 `gl_FragDepth` sub-band budget (band = 8 × kScatterCellTieStep, steps already at ~2 ULP), order equal-depth fragments by slot/flip priority BEFORE cell identity — mirroring the cardinal atomicMin low-bit semantics (`depth | flip | slot`) — without regressing the #2255 same-plane determinism. Candidate directions, none free:

1. Split the 3-bit sub-band code into slot-rank × cell bits — but #2255 needed all 3 bits of cell identity (in-plane neighbor steps project to iso (0,±2), forcing `y&3`).
2. Widen the band to fit both codes — risks merging genuinely distinct parallel planes (1 world unit ≈ 64 encoded units at effSub 8 ≈ the widened band).
3. Suppress covered-face emission at the store using pool-backed world occupancy (the light-occlusion bitfield) instead of per-set masks — moves the fight upstream, but adds a hot-path dependency and only helps fully-covered faces (partially-covered still emit).

Forensics, isolation ladder (`IRPerfGrid --debug-overlay unlit|ao|light_level`), and the empirical elimination chain are in PR #2410's investigation. Static scenes are covered by #2410's occupancy stamp; this issue is the moving-content residual.

jakildev:
## Summary
- Reallocates the per-axis forward-scatter's sub-band tie code **priority-major, cell-minor** per the approved `## Plan` on #2411: `code = (rank2 << 2) | cell2` with `rank2 = flip ? 3 : slot` (mirrors the cardinal encode's `(flip<<2)|slot` low-bit order) and `cell2 = (x&1)|(y&2)` (2 bits suffice — every same-plane/parallel-plane neighbor pair differs in one of them; the tying in-plane steps only project to iso-diagonal or (0,±2)).
- Cross-axis band ties — the #2411 stipple class on moving contiguous content, and the scattered wrong-face pixels along shared face edges — now resolve by slot rank **consistently along the whole plane-crossing strip** instead of alternating by cell parity. Same-rank ties fall to cell identity, preserving the #2255 determinism contract by construction.
- `kScatterCellTieBand` widens 8 → 16 steps (step stays 2⁻²³). Margin-vs-exact ordering survives: the margin bias lands a margin fragment ≥ 1 full band behind its same-plane exact owner — precondition `kScatterMarginDepthBiasKey × subScale / depthRange ≥ kScatterCellTieBand` documented at the constants (holds for depthRange ≤ subScale × 2¹⁷; default 131070 < 131072).
- #2333 overflow-lane bias scales 16 → 32 steps (still two tie bands; ~500 bands to the nearest genuine surface). All four mirror sites updated in lockstep (GLSL `ir_iso_common` + `f_peraxis_scatter` local mirror; Metal `ir_iso_common` + scatter) plus the design-doc contract line.

## Before / after

Moving-wave repro (`IRPerfGrid --grid-size 16 --zoom 8`, zoom4_rot). Drift map (red = changed pixels) shows the change confined to face-shared-edge tie strips:

| before | after | drift map |
|---|---|---|
| before | after | drift |

Static gallery at yaw 20° — drift likewise confined to scattered single tie-strip pixels along shared edges: gallery-drift

## Test plan

- [x] **Cardinal byte-identity:** full shape_debug suite pairwise vs a master build of the same tree — 28/28 shots img_diff drift = 0; gallery `gal_yaw000` drift = 0 (HUD-cropped).
- [x] **Determinism (the #2255 gate):** two-run pairs at static rotated poses (gal_yaw020 / gal_yaw090) byte-identical on BOTH master and this branch (drift 0). Mid-wave two-run pairs carry a pre-existing ≤7-max-delta lighting-LSB wiggle at the same magnitude on master and branch (8–1288 px master vs 24–588 px branch, scene-only) — not tie-class (winner flips would be full-color deltas), not regressed.
- [x] **Margin classes:** rotated drift confined to tie-strip pixels (drift maps above); no #1457 wrong-voxel-color bands, no #1494 dashing signatures. render-verify vs committed refs produces a failure table numerically identical to master's — those refs are stale on master (pre-#2410/#2418), filed as #2426; this PR gates on pairwise master-vs-branch diffs instead.
- [x] **Depth-probe asserts:** canary depth-write PASS (host-scaled aim — Retina 2× framebuffer); `scripts/depth-tier-verify.py` 753/753 frames PASS at (639,362) tier=2.
- [x] **Temporal:** pan-sweep SMOOTH (reversals 0, residual 0.57px). Yaw-sweep reports JITTER with metrics byte-identical to master (pre-existing, filed as #2427).
- [x] Metal built + all runs `ir-run: RESULT=CLEAN` (authored on macOS; GLSL edited in lockstep).
- [ ] Linux/OpenGL smoke (cross-host labels).

## Notes for reviewer

- The frac-cube shared-edge fringe + bottom-vertex notch visible in the gallery ROI pair (`gallery-frac4-roi-{before,after}.png`, unchanged by this PR) are a **separate defect class** — per-axis analytic-coverage phase disagreement on fractional-position shared edges, evidenced by the `peraxis_id` overlay crop (`gallery-frac4-peraxis-id-2428.png`: single-pixel Z/X-canvas interleave). Filed as #2428 with the investigation direction; deliberately not bundled here so this PR's gates stay clean.
- Perf: no new buffers, one compare+select and shift/or per scatter vertex — below measurement noise; the plan waived a dedicated profile pass.

Closes #2411


jakildev (commented):
## Review — render: per-axis scatter tie order — priority-major sub-band, band 8→1

**Verdict:** approve

### Nits
- `engine/render/src/shaders/ir_iso_common.glsl:920-927` (+ the three mirrors) — the margin-vs-exact precondition (`kScatterMarginDepthBiasKey × subScale / depthRange ≥ kScatterCellTieBand`) now holds by only a 2-unit margin at defaults: `depthRange = 131070` vs the `subScale × 2¹⁷ = 131072` ceiling (recomputed independently from `kMaxTriangleDistance`/`kMinTriangleDistance` = ±65535 in `ir_constants.hpp`). A future band-widening pass (e.g. if #2428's fractional-edge fix needs more tie levels) could cross this threshold silently — worth a `static_assert`-style comment flag or a compile-time check tying `depthRange`'s range to `kScatterCellTieBand`, so the next widening doesn't have to re-derive this by hand.

### Praise
- All four shader mirror sites (`ir_iso_common.glsl`/`.metal`, `v_peraxis_scatter.glsl`, `metal/peraxis_scatter.metal`, plus the fragment-stage no-common-include mirror in `f_peraxis_scatter.glsl`) land byte-for-bit identical bit layouts (`(rank2 << 2) | cell2`, same shift amounts, `int`/`uint` typing aside) — verified directly, no drift between GLSL and Metal.
- The #2333 overflow-bias scaling (16→32 steps) is kept in lockstep with the band widening (8→16) — still exactly two bands behind, not an independent magic number.
- Precondition arithmetic in the updated comment checks out: recomputed `depthRange ≤ subScale × 2¹⁷` independently and it matches the PR body's claim exactly.
- Cardinal byte-identity + determinism test plan (28/28 shots, two-run pairwise diffs at rotated poses) is exactly the right regression gate for a tie-order change, and the #2428/#2426/#2427 out-of-scope defects are correctly filed separately rather than bundled in.

- [ ] Linux/OpenGL smoke — already called out as pending in the PR's own test plan; tagging `fleet:needs-windows-smoke` below per the OpenGL-tier rule (author covered Metal only).

Escalation: please Opus-recheck before merge (touches: engine/render sun-shadow/scatter tie-break invariant — a multi-layer numeric precondition across two shader backends; I'm confident in the bit-layout parity but want a second pass on the margin-vs-exact ordering guarantee).

jakildev (commented):
## Opus recheck — render: per-axis scatter tie order — priority-major sub-band, band 8→16 (#2411)

**Verdict:** approve

Sonnet escalated on one question: does the **margin-vs-exact ordering guarantee** survive the band widening? **It does — I confirm it, and the proof is tighter than Sonnet's nit describes.** The precondition has two independent halves, and the diff only documents one of them. Both hold at head `8f08a96`; neither has meaningful slack. Details in Nits — they're documentation asks on a load-bearing invariant, not correctness defects.

### Confirming the escalation — margin-vs-exact, derived independently

Sonnet checked the precondition at defaults and found a 2-unit margin (`depthRange = 131070` vs the `subScale × 2¹⁷ = 131072` ceiling). **I confirm that**, but the arithmetic Sonnet ran assumes `subScale = 1`, and `subScale` is a **runtime** value (`effectiveSubdivisionsForHover.x`), not a constant — so the check as written doesn't actually close the question. It closes for a reason Sonnet didn't cite:

- `v_peraxis_scatter.glsl:339` / `metal/peraxis_scatter.metal:377` — `subScale = max(effectiveSubdivisionsForHover.x, 1.0)`. The **floor-clamp at 1.0** is what makes `subScale = 1` the true worst case; `marginBias` is linear in `subScale`, so every larger subdivision only adds headroom (2× at `subScale = 2`, 4× at 4). Without that clamp a sub-unit `effSub` would invert the precondition and margins would start beating their exact owners.
- `depthRange` is genuinely fixed, not just defaulted: `kTrixelDistanceMinDistance` / `kTrixelDistanceMaxDistance` are `constexpr ±65535` (`engine/common/include/irreden/ir_constants.hpp:47,51`) reaching the shader via `ir_render_types.hpp:31-32`. So `131070` is compile-time, and the only free variable is the clamped `subScale`.
- In float32 at the worst case: `marginBias = 1.9073777e-06` vs `band = 2⁻¹⁹ = 1.9073486e-06` → ratio **1.0000153**. The gap is ~1.5e-5 relative, ~250× float32 eps, so rounding cannot collapse the inequality. Holds — by 0.0015%.
- The quantization algebra completes it: `marginBias ≥ band` ⟹ `floor(depthM/band) ≥ floor(depthE/band) + 1`, so `qM − qE ≥ band − maxCode×step = 16 − 15 = 1 step > 0`. Margin never ties or beats its exact owner.

That last line is the **second half of the precondition, and it is nowhere in the diff**: the guarantee also requires `maxCode ≤ bandSteps − 1`. `maxCode = (3 << 2) | 3 = 15`, `bandSteps = 16` — it fits with **exactly zero slack**. This is also why the band had to go 8→16 and cannot go to 32 (Nit 1).

Sonnet's read that the four mirror sites land bit-identical layouts: **confirmed**, and I extend it — the two fragment-stage quantizers (`f_peraxis_scatter.glsl:87`, `metal/peraxis_scatter.metal:471`) are also identical (`floor(d/band)*band + offset`), and both vertex early-outs zero the offset on the degenerate-instance path (`v_peraxis_scatter.glsl:198`, `metal:221`), so no uninitialized carrier reaches a fragment.

### Nits

- **`ir_iso_common.glsl:919-927` — the documented precondition is only half the constraint.** The comment gives `kScatterMarginDepthBiasKey × subScale / depthRange ≥ kScatterCellTieBand` (⟹ `depthRange ≤ subScale × 2¹⁷`), but the ordering equally requires `maxCode ≤ bandSteps − 1` (15 ≤ 15, exact). These pull in **opposite** directions: widening the band relaxes the first and is required by the second. A future pass adding tie levels — e.g. a 3-bit rank to fix the Nit-2 collapse, or #2428's fractional-edge work — would push `maxCode` to 23, need a 32-step band, and then need `depthRange ≤ subScale × 2¹⁶ = 65536` while it is 131070: **the widening that fixes the code overflow is exactly the one that breaks the margin precondition.** This is the real reason the 2-bit rank collapse exists, and it's currently a coincidence a reader has to re-derive. Endorsing Sonnet's ask and sharpening it: state both halves at the constants, note they're mutually opposed, and note that band 16 is the unique value satisfying both at the current `depthRange`. A `static_assert`-style CPU-side check is the durable form, since the shader can't assert.
- **`ir_iso_common.glsl:899-905` — the "stay deterministic" parenthetical overclaims for one pair class.** The rank map is sound for three of four cases, which I verified: flipped-vs-unflipped always separates (rank 3 vs 0..2); flipped-vs-flipped **same slot** is the same-axis parallel-plane class, where `cell2`'s in-plane-step proof (iso-diagonal flips `x&1`, `(0,±2)` flips `y&2`) genuinely applies. The gap is flipped-vs-flipped on **different** slots: both collapse to rank 3 and fall to `cell2`, but their `ij` live in different axis canvases, so the in-plane-step enumeration backing `cell2` doesn't cover them and nothing proves the codes differ. On collision the winner falls back to draw order — the #1961 compaction's run-variant append order — i.e. the #2255 class. **This is not a regression and not a merge gate**: master had the identical structural hole (its 3-bit `(x&1)|((y&3)<<1)` had no cross-axis distinctness proof either), the branch only narrows it 8→4 levels for that one rare class, and it strictly *fixes* the common unflipped cross-axis case by giving it a provable slot separation. The collapse is also forced, per Nit 1 — 6 face states don't fit in 2 bits. The ask is just to make the comment say what's proven: `cell2` is proven distinct for same-plane/parallel-plane pairs, and cross-axis flipped-vs-flipped ties rely on `ij` happening to differ — deterministic when they do, draw-order when they don't. Worth a #2411 follow-up note if the flipped class ever stops being rare.

### Praise

- Assuming Sonnet's checklist pass (naming, mirrors, bit layout) — spot-checked, no blatant misses.
- The `~10³ → ~500 bands` figure in the #2333 overflow comment is correctly halved with the band doubling, and I verified the absolute number rather than the ratio: at `kDepthEncodeShift = 8` a plane step is 32 bands at `subScale = 1` and ~512 at the high-zoom `subScale ≈ 16` the figure references. Either way it dwarfs the 2-band overflow bias, so the #2333 argument holds at both ends of the zoom range — the scaling was not a cosmetic find-and-replace.
- `1.0 key unit = 4 bands` in the updated comment: verified exactly (`(1/131070)/2⁻¹⁹ = 4.00006`).
- The margin-yield term can only push a margin fragment further back (`outside ≥ 0`, grads pre-absed, `f_peraxis_scatter.glsl:79-81`), so bounding the precondition on the flat bias alone is the correct conservative worst case rather than an oversight.
- Gating on pairwise master-vs-branch diffs once the committed refs were found stale (#2426), instead of quietly re-baselining them, is the right call — and filing #2426/#2427/#2428 out rather than bundling keeps this diff's gates readable.

- [ ] Linux/OpenGL smoke — `fleet:needs-windows-smoke` already tagged by Sonnet; no further action from me.

jakildev:
Addressed feedback (`fleet:has-nits`) at `791d0d50`. All three nits were documentation / durability asks on a load-bearing invariant, so **there is no behavior change**: every shader edit is a comment, and the C++ side adds only `constexpr` mirrors + `static_assert`s (no emitted code). `fleet:approved` retained per the has-nits path.

**Sonnet nit — compile-time check tying `depthRange` to `kScatterCellTieBand`** ✅
Added CPU-side mirrors + asserts in `ir_render_types.hpp` (a shader can't assert). Kept in **exact integer arithmetic** — `kScatterMarginDepthBiasKey` enters as its reciprocal (`…KeyInv = 4`), so the check is `2^23/4 >= bandSteps × depthRange` → `2097152 >= 2097120`, no float rounding in the comparison at all.

**Opus nit 1 — the documented precondition is only half the constraint** ✅
Both halves now stated at the constants, marked as mutually opposed, with the uniqueness result:
- (a) margin-vs-exact: `depthRange <= subScale × 2^17` — **tightened** by widening; holds by 2 units (131070 < 131072). Noted that `subScale` is floor-clamped to 1.0, which is what makes `subScale = 1` the true worst case (your correction to Sonnet's arithmetic).
- (b) code-fits-in-band: `maxCode <= bandSteps − 1` → 15 ≤ 15, **zero slack** — **relaxed** by widening.

I re-derived the bracket independently rather than taking it on faith: `bandSteps ∈ [16, 16.000244]`, so **16 is the unique admissible integer width** — now its own assert, so the next widening pass sees *why* it fails instead of re-deriving. The comment states the trap explicitly: a 3-bit rank (or #2428's fractional-edge work) needs maxCode 23 → 32-step band → `depthRange <= 65536` while it is 131070, i.e. the widening that fixes the code overflow is the one that breaks (a).

**Verified the asserts fire, not just compile** — the point of a durability guard is that it trips:

| `kScatterCellTieBandSteps` | result |
|---|---|
| 16 (current) | compiles clean |
| 32 (the 3-bit-rank widening) | ❌ margin-vs-exact assert fires |
| 8 (narrowing) | ❌ code-fits assert **+** uniqueness assert fire |

**Opus nit 2 — the "stay deterministic" parenthetical overclaims** ✅
Rewrote the claim as a per-class breakdown stating exactly what's proven, and dropped the "rare co-rotated riser pairs stay deterministic" assertion: cross-axis **flipped-vs-flipped** pairs both collapse to rank 3 and their `ij` index different axis canvases, so `cell2`'s in-plane-step enumeration doesn't cover them — they tie by draw order when the codes collide (#2255 class). Recorded that this is not a regression (master's 3-bit cell code had the same cross-axis hole; this narrows it to one rare class and gives the common unflipped cross-axis case a provable separation), with the `#2411` revisit-if-it-stops-being-rare note you asked for.

The overclaim was mirrored in three other places, so it's corrected at all four sites: `ir_iso_common.{glsl,metal}`, `v_peraxis_scatter.glsl`, `metal/peraxis_scatter.metal`. `f_peraxis_scatter.glsl`'s standalone band mirror now warns against retuning the 16 in isolation. The two inline `rank2` comments already scoped `cell2` correctly ("for every same-plane / parallel-plane neighbor pair") and were left alone.

**Verification** — `IRShapeDebug` builds clean and runs `ir-run: RESULT=CLEAN`; shaders compile with no errors (the one real risk in a comment-only shader edit). No render-verify pair: `git diff` confirms every shader change is a comment line, so the pipeline is byte-identical by construction.

Test plan unchanged: `fleet:needs-windows-smoke` still open for the Linux/OpenGL smoke.

— worker (worker-2, opus)

jakildev (commented):
## Re-review — render: per-axis scatter tie order — priority-major sub-band, band 8→16 (#2429)

**Verdict:** approve

Picked up via `human:re-review` (auto-tagged by the `New commits pushed after fleet approval` bot). Scoped this pass to the new commit `791d0d50` on top of the already-reviewed head `8f08a96` (sonnet review + Opus recheck, both approve). This commit is the author's response to the sonnet review's one nit (the margin-vs-exact precondition should be enforced, not just documented).

### What changed
- **No shader logic or numeric value changed.** Diffed `8f08a96..791d0d50`: every edit in the four shader files is a comment expansion (the existing `rank2` collapse note, the PRECONDITION block) — `kScatterCellTieStep`, `kScatterCellTieBand`, `kScatterMarginDepthBiasKey` are byte-identical to the pre-nit-fix state.
- **New CPU-side `constexpr` + three `static_assert`s** in `ir_render_types.hpp` mirror the shader constants and check, at compile time, exactly the precondition the nit flagged:
  1. margin-vs-exact (`biasInv` reciprocal form, integer arithmetic to avoid float rounding at compile time)
  2. code-fits-in-band (`maxCode <= bandSteps - 1`)
  3. uniqueness (doubling the band to fix #1 would break #2 — so 16 is forced, not chosen)

### Verification performed
- Re-derived the arithmetic independently rather than trusting the comment: `kScatterTieDepthRange = kTrixelDistanceMaxDistance(65535) - kTrixelDistanceMinDistance(-65535) = 131070` (confirmed both constants in `engine/common/include/irreden/ir_constants.hpp:47,51`). Assert 1: `(1<<23)/4 = 2097152 >= 16*131070 = 2097120` — holds, 32-unit slack, matches the comment. Assert 2: `(3<<2)|3 = 15 <= 15` — exact, matches. Assert 3: `2097152 < 32*131070 = 4194240` — holds.
- These three asserts encode the same math the Opus recheck on the prior commit verified by hand (including the `subScale`-is-runtime correction Opus added on top of the sonnet review) — the fix converts a hand-checked invariant into a compiler-checked one, which is exactly what was asked for.
- Confirmed the `Distance` type underlying `kTrixelDistanceMinDistance`/`kTrixelDistanceMaxDistance` is a plain integer alias, so the `constexpr int` computation in the new block has no truncation surprises.
- `TWO16`-equivalent naming isn't in play here, but same spirit: `kScatterMarginDepthBiasKeyInv = 4` documented as "reciprocal of `kScatterMarginDepthBiasKey` (0.25 key units), kept as a reciprocal so the precondition stays in exact integer arithmetic" — correct rationale, avoids a float divide in the static_assert.

No correctness issues. This is a comment/compile-time-guard-only follow-up; the underlying invariant was already correct and already independently verified twice (sonnet + Opus) before this commit landed.

### Nits
None.

### Praise
- Converting a documented-but-unenforced numeric precondition into a `static_assert` triple (with each assert's failure message explaining *which* direction to fix, not just that it broke) is exactly the durability improvement the original nit asked for, and it's placed where the next widening pass will actually see it (compile error, not a comment nobody re-reads).

Opus recheck required: touches `engine/render/` (shader tie-break invariant + its new CPU-side compile-time enforcement) — same surface as the already-Opus-rechecked prior commit; flagging again per the blanket engine/render/ escalation rule even though this commit is comment/assert-only and I found no behavior change.

jakildev (commented):
## Opus recheck (amend) — render: per-axis scatter tie order — priority-major sub-band, band 8→16 (#2411)

**Verdict:** approve

Re-affirming on the new head `791d0d50` after the amend (my prior approval was on `8f08a96`; `human:re-review` picked up). Both nits from that recheck are addressed, and addressed in the durable form I asked for rather than the cheap one. I verified the amend's claims independently instead of taking them from the commit message — the three that mattered all hold.

### Nit 1 (two-halved precondition) — resolved, and the guard actually fires

The ask was a `static_assert`-style CPU-side check, since a shader cannot assert. That's what landed (`ir_render_types.hpp:412-450`). I re-derived it rather than trusting it:

- **Mirrors match the authoritative shader values.** `kScatterMarginDepthBiasKeyInv = 4` ↔ `kScatterMarginDepthBiasKey = 0.25`; `kScatterCellTieStepShift = 23` ↔ `kScatterCellTieStep = 1.0/8388608.0`; `kScatterCellTieBandSteps = 16` ↔ `kScatterCellTieBand = 16.0/8388608.0` — and GLSL and Metal agree on all three. `kScatterTieDepthRange` resolves to `65535 − (−65535) = 131070` from the real `constexpr`s, not a literal.
- **Arithmetic is exact, no truncation.** Half A is `2^23/4 = 2097152 ≥ 16 × 131070 = 2097120` (slack 32, i.e. the range may reach 131072 and is 131070); half B is `15 ≤ 15`, zero slack; uniqueness is `2097152 < 32 × 131070 = 4194240`. `(1 << 23)/4` divides evenly, so the integer division the check depends on is exact as claimed.
- **The asserts FIRE — reproduced independently.** A durability guard that only compiles is worthless, so I replicated the three asserts standalone and swept `bandSteps`. The commit's table reproduces exactly:

  | `kScatterCellTieBandSteps` | result |
  |---|---|
  | 16 (current) | compiles clean |
  | 32 (the 3-bit-rank widening) | half A (margin-vs-exact) fires |
  | 8 (narrowing) | half B (code-fits) **+** uniqueness fire |

  So the bracket is genuinely enforced, and the next widening pass gets told *why* it fails at compile time rather than re-deriving it.

**Residual, acknowledged, not a gate:** the CPU constants are a hand-maintained mirror, so a future pass that retunes the band *in the shader alone* would not trip anything — the asserts only check the mirrors against each other. This is inherent to "a shader cannot assert," and the amend mitigates it the only way available short of build-time constant extraction: every mirror site now points at `ir_render_types.hpp`, and `f_peraxis_scatter.glsl:45-51` (the standalone no-include mirror) explicitly warns that retuning the 16 there alone "would silently break margin-vs-exact ordering." That's the right mitigation; flagging it only so the limitation is on the record.

### Nit 2 (overclaimed determinism) — resolved, removed rather than softened

The parenthetical I flagged ("rare co-rotated riser pairs stay deterministic") is **gone**, not hedged — `git grep 'stay deterministic'` across every `.glsl`/`.metal` at this head returns empty. In its place is a per-class breakdown that states the gap outright: *"flipped vs flipped on DIFFERENT slots — NOT proven distinct … the codes differ only if the ij happen to, and on collision the winner is draw order"*, with the not-a-regression note (master's 3-bit cell code had the same cross-axis hole) and the revisit-if-it-stops-being-rare hook. That is exactly the scope I asked for and no more.

All four mirror sites are corrected, including `metal/peraxis_scatter.metal:92-98`, which my first grep missed because it phrases the same scoping differently ("without a distinctness proof — they stay draw-order on collision … Per-class argument: ir_iso_common.glsl") and defers the full argument to the GLSL site rather than duplicating it. Correct call — one authoritative derivation, pointers from the mirrors.

### No-behavior-change claim — mechanically verified

The commit asserts "every shader edit is a comment, and the C++ side adds only constexpr mirrors and static_asserts (no emitted code)." Verified rather than assumed:

- Across all five shader files, the amend has **zero** added or removed lines that are not comments or blank — so the "pipeline is byte-identical by construction" argument is sound, and skipping a render-verify pair here is justified rather than a gap.
- C++ scope is `ir_render_types.hpp` only; the addition is `constexpr int` + `static_assert`, which emit nothing.
- `791d0d50`'s parent is `8f08a9646671...` — the exact commit I approved at 21:37 is intact underneath, so this is purely additive and the earlier recheck's findings carry forward unchanged.

### Note for the stacked child (#2432)

The new comment predicts that "#2428's fractional-edge work" pushes `maxCode` to 23 and needs a 32-step band. Worth recording that **the prediction does not fire against #2432 as implemented**: it takes a flat `kScatterMarginInteriorBiasKey` plus a yield-slope floor rather than a third rank bit, leaving `kScatterCellTieBandSteps` and the `(rank2 << 2) | cell2` width untouched — `kScatterTieMaxCode` stays 15, so the stacked pair still satisfies both asserts. The documented trap is real for a *future* rank widening; it just isn't what #2428 did. #2432 is gated on this PR's approval and will be re-evaluated now that it lands.

### Praise

- Verifying the asserts trip at 32 and 8 — and putting the table in the PR — is the difference between a guard and a decoration. It also caught the thing I'd have asked for next: that narrowing trips *two* asserts, which pins the bracket from both sides.
- The uniqueness assert is the genuinely thoughtful addition. Neither Sonnet nor I asked for it; it converts "16 is forced" from a comment a reader must re-derive into a compile-time fact with an error message that explains the derivation. That's the durable form of the finding.
- Scoping the determinism claim by *class* rather than deleting the parenthetical preserves the useful information (three of four classes are proven) while killing the false part.

### Test plan

- [ ] Linux/OpenGL smoke — `fleet:needs-windows-smoke` remains tagged from the first pass; unchanged by this amend (comment-only), no further action from me.

jakildev:
Review follow-ups addressed in 8c1f845f:

**Constants consolidation** — the band width now has exactly one definition site per backend: `kScatterCellTieBand = 16.0 * kScatterCellTieStep` in the `ir_iso_common` twins (the 16 stays pinned by the ir_render_types.hpp static_asserts from 791d0d50). The overflow bias derives its documented contract directly (`2.0 * kScatterCellTieBand` at both vertex stages — no more separate 32-literal), and the GLSL fragment mirror is deleted outright: `f_peraxis_scatter` now pulls the constant through the same `ir_iso_common.glsl` include the sibling fragment gather (`f_trixel_to_framebuffer`) already uses. All derivations are exact power-of-two products — verified byte-identical to the pre-refactor build (gallery yaw020 scatter-active + yaw090, img_diff drift 0).

**static_assert vs engine macros** — investigated: `IR_ASSERT` is the engine's *runtime* macro (236 sites, log-sink + release-stripped); compile-time checks across 10+ engine files already use raw `static_assert`, and no `IR_STATIC_ASSERT` wrapper exists (nor would it add anything — no sink, no stripping question). So 791d0d50 follows existing practice. The real gap is that this rule is documented nowhere and the raw-runtime-`assert()` violation class is uncaught (one live site: `chunk_header.hpp:69`) — filed as #2440 (fleet:coding-improvement) with the doc rule + simplify check + migration.

Note for the merger: the stacked #2432 branch was NOT rebased onto 8c1f845f (it is checked out in the main clone, so I did not rewrite it); its hunks do not overlap the refactor, so the eventual merges stay clean either way.

jakildev (commented):
## Re-review — render: per-axis scatter tie order — priority-major sub-band, band 8→16 (#2429)

**Verdict:** approve

Picked up via `human:re-review` (auto-tagged after new commits pushed post-approval). Scoped this pass to the new commit `8c1f845f` on top of the already-Opus-approved head `791d0d50`.

### Prior-review resolution
No open items from the prior review history — both the Sonnet and Opus passes on `791d0d50` closed clean (approve, no open nits). This commit is a further author-initiated cleanup, not a fix to flagged feedback, so there's nothing to carry forward.

### What changed
Pure dedup, no behavior change:
- `ir_iso_common.glsl` / `metal/ir_iso_common.metal`: `kScatterCellTieBand` changed from the literal `16.0 / 8388608.0` to the derived `16.0 * kScatterCellTieStep`. Both are exact powers of two (`8388608 = 2^23`, `16 = 2^4`), so the float multiplication is bit-identical to the literal — no rounding introduced.
- `f_peraxis_scatter.glsl`: deleted its local mirror `const float kScatterCellTieBand = 16.0 / 8388608.0` and now `#include "ir_iso_common.glsl"` instead — the same include `f_trixel_to_framebuffer.glsl` already uses (confirmed precedent), so the tie band now has exactly one GLSL definition site.
- `v_peraxis_scatter.glsl` / `metal/peraxis_scatter.metal`: the #2333 overflow bias changed from `32.0 * kScatterCellTieStep` to `2.0 * kScatterCellTieBand`. Since `kScatterCellTieBand = 16.0 * kScatterCellTieStep`, this is exactly `32.0 * kScatterCellTieStep` — same reasoning, bit-identical.
- `docs/design/per-axis-trixel-canvas-rotation.md` comment updated to match the new derived-form expression.

### Verification performed
- Checked `ir_iso_common.glsl` for symbol collisions before trusting the new include in `f_peraxis_scatter.glsl` — no top-level `in`/`out`/`uniform` declarations in the 1124-line file that could clash with the fragment stage's existing varyings.
- Grepped for stray un-updated mirrors of the old literals (`32.0 * kScatterCellTieStep`, `16.0 / 8388608`) across `engine/render/src/shaders/` — none found; all 4 known mirror sites are consistently on the derived form.
- Confirmed no duplicate `kScatterCellTieBand` definitions remain (only the two intended definition sites: GLSL and Metal `ir_iso_common`).

No correctness issues.

### Nits
None.

### Praise
- Collapsing 3 shader sites' worth of independently-literal magic numbers into one derived definition (plus the include-reuse for the 4th) is exactly the kind of durability follow-up that prevents the next band-width change from having to retune 4 sites in lockstep — and the commit message backs the "byte-identical" claim with an actual img_diff run rather than just asserting it from the exact-arithmetic argument.

- [ ] Linux/OpenGL smoke — `fleet:needs-windows-smoke` already tagged from the first pass; unaffected by this comment/derivation-only amend.

Escalation: this PR's history already carries the blanket `engine/render/` escalation rule (shader tie-break invariant, already Opus-reviewed twice). This commit is comment/derivation-only with no behavior change.

Opus recheck required: touches `engine/render/` shaders (tie-band constant derivation) — same surface as the already-Opus-rechecked prior commits, per the blanket engine/render/ escalation rule, even though I found no behavior change.

jakildev (commented):
## Opus recheck — render: per-axis scatter tie order — priority-major sub-band, band 8→16 (#2411)

**Verdict:** approve

Scoped to `791d0d50..8c1f845f` — the one commit since my prior recheck approved `791d0d50`. Author-initiated dedup, not a fix to flagged feedback, so nothing carries forward.

**Sonnet flagged this as pure dedup with no behavior change and verified the arithmetic; on closer read I confirm both** — `16.0 * kScatterCellTieStep` and `2.0 * kScatterCellTieBand` are exact power-of-two products (`2⁻²³ → 2⁻¹⁹`, `2⁻¹⁹ × 2 = 2⁻¹⁸ = 32 × 2⁻²³`), so the derived forms are bit-identical to the literals they replace and the `floor`-quantization stays exact in float32. I won't re-derive what Sonnet already checked correctly.

The one thing Sonnet's pass **didn't** close is the only genuinely new risk in this commit, so I spent the pass there.

### The new `#include` in a fragment stage — the real risk, and why it's safe

`f_peraxis_scatter.glsl` drops its local mirror and now `#include`s the 1124-line `ir_iso_common.glsl`. Sonnet checked this for top-level `in`/`out`/`uniform` collisions and cited `f_trixel_to_framebuffer.glsl` as precedent. That's necessary but not sufficient — **a GLSL file can fail to compile in a fragment stage because of a stage-specific builtin referenced from a function that is never called.** Strict frontends (Mesa) reject `gl_LocalInvocationID` outside compute even from dead code, so "the functions are unused, therefore harmless" is not a safe argument. I checked the stronger property instead of assuming it:

- **No `#version` / `#extension` in the common file** — so the include lands after `f_peraxis_scatter.glsl`'s own `#version 450 core` without a mid-file duplicate-version error. Placement is correct.
- **No top-level `layout` / `uniform` / `buffer` / `sampler` / varying declarations at all** — this is a pure constant+function library, so the include adds **no bindings and no interface** to the fragment stage. Nothing to collide with the stage's existing varyings, and no binding-point drift.
- **No stage-specific builtins anywhere in the file** (`gl_GlobalInvocationID`, `gl_LocalInvocationID`, `gl_WorkGroupID`, `gl_VertexID`, `gl_Position`, `imageStore/Load`, …). This is **by explicit design**, not by luck — `ir_iso_common.glsl:104-114` documents it: `localIDToFace_2x3` takes the invocation id *as a parameter* rather than reading the builtin, precisely "so this helper compiles inside vertex/fragment shaders that include this header (e.g. `f_trixel_to_framebuffer.glsl`)… Strict GLSL frontends (Mesa) error on `gl_LocalInvocationID` references even from unused functions outside compute stages." The file is engineered for exactly this use.
- **No function-name collision** — `f_peraxis_scatter.glsl` defines `hsvToRgb` and `main`; the common file defines neither (`hsvToRgb`/`hsv2rgb`/`rgbToHsv` are absent), so no redefinition error.
- The line-2 `#include "ir_iso_common.glsl"` inside the common file is a **doc comment** ("Included via…"), not a self-include — no preprocessor recursion.

So the include is safe by construction, and the precedent Sonnet cited is load-bearing rather than incidental.

### Mirror consistency — verified, not assumed

- `metal/peraxis_scatter.metal:12` includes `ir_iso_common.metal`, so its vertex stage resolves the derived `kScatterCellTieBand` the same way. Metal keeps vertex+fragment in one translation unit, so there was never a local fragment mirror to remove there — the GLSL-only deletion is the correct asymmetry, not missed parity work.
- **No stale literals survive anywhere**: `git grep -E '32\.0f? \* kScatterCellTieStep|16\.0f? / 8388608'` across `engine/` and `docs/` returns empty. All four shader sites plus the design doc are on the derived form, so the next band retune has one place to change per backend.
- Locally confirmed the branch merges cleanly against `origin/master` (`01634a50`) via `merge-tree` — the PR's `mergeable: UNKNOWN` is GitHub's async lag, not a conflict.

### Nits

- `engine/render/src/shaders/f_peraxis_scatter.glsl:14` — the comment justifying the include says "Sole consumer here is `kScatterCellTie{Step,Band}`", but this stage consumes only `kScatterCellTieBand` (line 88); `kScatterCellTieStep` isn't referenced. Trivial, but the comment's whole job is to explain why a 1124-line include earns its place in a fragment stage, so it should name what's actually used — drop the `Step,` and it's exact.

### Praise

- Deriving the overflow bias as `2.0 * kScatterCellTieBand` rather than `32.0 * kScatterCellTieStep` makes the *relationship* the source of truth instead of a coincidence between two magic numbers. The old form encoded "two bands behind" in a literal that only stays correct if someone remembers to retune it in lockstep with the band width — the #2333 bias and the #2411 band are now structurally coupled, which is exactly the failure mode the previous rounds kept having to hand-verify.
- Reducing GLSL to one definition site per backend closes the drift surface that my earlier recheck flagged as the residual risk: with four independent literals, a band retune had four chances to go wrong silently. It now has one per backend, both pinned by the CPU-side `static_assert`s from `791d0d50`.
- Backing the byte-identity claim with an actual `img_diff` run rather than resting on the exact-arithmetic argument is the right instinct — the algebra says the floats are identical, but the render is what the change is accountable to.

### Test plan before merge

- [ ] Linux/OpenGL smoke — `fleet:needs-windows-smoke` already tagged from the first pass; retained. **This is the one gate that matters for this commit**, and it's not a formality here despite the change being comment/derivation-only: the new fragment-stage `#include` is a GLSL-frontend concern that a macOS/Metal host structurally cannot exercise (I verified the include's safety statically — no stage-specific builtins, no interface decls, no collisions — but static analysis is not a compile). The Mesa strictness the common file's own comment warns about is exactly the class of failure this smoke would catch.

PR #2418, opened 2026-07-15
render: per-axis wFrac carrier + active-mask slot restore — cubes draw as cubes under yaw

jakildev:
## Summary

Follow-through on the yaw face-alignment report: the user's isolated-entity reduction idea (fewer cubes, not stacked, varied shapes) went in as a `perf_grid --mode gallery` scene — and it immediately isolated **two engine defects** that together explain "the cubes are not drawing as cubes under yaw":

**1. The per-axis encoding dropped the out-of-plane sub-cell fraction** (commit 1). The base-resolution store packed only the two in-plane fracs; the fraction along the face's own axis had no carrier, so every face of fractionally-positioned content reconstructed on the integer lattice plane — up to half a voxel off along its own normal, each of a cube's three faces snapping to a *different* plane. Worst case is exactly the perf_grid default: an origin-centered grid puts every voxel at half-integer ties. New layout adds a `wFrac4` field below depth (atomicMin still orders by true plane depth); all layout knowledge stays in the `ir_iso_common` twins.

**2. The #2388 overflow-lighting dispatch permanently stole the voxel active-mask slot** (commit 2). It bindBase'd the overflow scratch over slot 8 (`kBufferIndex_OverflowLightingScratch` aliases `kBufferIndex_VoxelActiveMask`) with no restore. On the cardinal return the freed scratch dangled there — the #2412 segfault — and once the destructor scrub (landed on master via #2415) nulls it, the compact reads an unbound active mask and **the whole scene vanishes at any cardinal reached after rotating**, permanently. Root fix: restore slot 8 after the dispatch, mirroring the existing 25/26 discipline. (the #2415 destructor scrub stays as defense-in-depth; #2417 — the stacked twin of that scrub — was closed as redundant.)

Commit 3 adds the gallery mode + these screenshots.

## Before / after

Fractional-position cube (gallery row B, yaw 20°) — the detached-lid class:

| before | after |
|---|---|
| before | after |

Half-integer-tie class (the perf_grid default placement — the fat lid, worst case) and the fixed static tie grid:

| before (tie-positioned cube) | after (16³ tie grid, zoom4_rot) |
|---|---|
| tie-before | tie-after |

Default wave scene at yaw 0.35 (the reported artifact; wave phase differs between captures — compare texture quality, not layout):

| before | after |
|---|---|
| wave-before | wave-after |

Cardinal-90 reached after rotating — scene vanished, now restored:

| before | after |
|---|---|
| c90-before | c90-after |

[...]

PR #2431, opened 2026-07-15
render: view-space face micro-positions at cardinals 1/2/3 — kill the POS-face shared-edge seams

### LINKED ISSUE #2424: render: subdivided cardinal raster opens 1-sub-unit seams on rotated-in POS faces at cardinals 1/2/3 (gallery card90 black borders)
## Symptom

At the rotated-to cardinals (yaw 90/180/270 — `cardinalIndex` 1/2/3), every face-shared edge bordering a rotated-in POS face shows a 1-sub-division-unit background seam ("black border") in the subdivided single-canvas raster. Canonical repro: `IRPerfGrid --mode gallery --auto-screenshot` → `gal_yaw090`: all left faces (world X_POS at cardinal 1) carry a black border along the top-left and vertical shared edges. `gal_yaw000` is clean. The seams persist identically under `--debug-overlay unlit` → the seam pixels are unwritten background (coverage), not lighting.

## Root cause (verified against master)

`c_voxel_to_trixel_stage_1_body.glsl` (subdivided path, ~:863-870) computes face micro-positions in the WORLD frame (`faceMicroPositionFixed6(faceId, ...)`), then applies `rotateCardinalZ` + a blanket `cardinalLowerCornerShift * subdivisions`. The shift is a CELL-INDEX correction (half-open interval `c → -c-1` under axis negation), but it is also applied to face PLANE coordinates (boundaries map `c → -c`):

- A rotated-in POS face's plane (`+S` on its world axis, negated by the rotation) lands **S sub-units low** in view space (the `+S` and the shift double-count).
- Its neighbor faces' in-plane sweeps along the negated axis run inverted, covering `[Q-(S-1), Q]` instead of `[Q, Q+S-1]` — **S-1 sub-units low**.

The one-sub-unit disagreement opens the background seam along every shared edge of the POS face; the whole scene additionally sits ~1 world cell off the true rotated projection, mis-registering against the exact inverse recovery (`trixelCanvasPixelToWorld3D`) used by lighting/picking at rotated cardinals.

## Fix

Compute micro-positions natively in VIEW space — exact cardinal-0 equivalence of the rotated scene (cardinal 0 is seam-free by construction):

```
viewFace  = rotateFaceIdCardinalZ(faceId, cardinalIndex)   // new ir_iso_common helper (face-normal image under R_z)
viewCell  = rotateCardinalZ(voxelPositionFixed, c) + cardinalLowerCornerShift(c) * subdivisions
micro     = faceMicroPositionFixed6(viewFace, viewCell, u, v, subdivisions)
```

Apply at both stage-1 emit sites (primary + both-polarities dual emit), mirror byte-identically in `c_voxel_to_trixel_stage_2` (color tap re-derivation), and port both Metal twins. The unsubdivided (`voxelRenderOptions.x == 0`) whole-voxel path and the per-axis store (un-yawed keys) are untouched. Cardinal 0 is byte-identical (identity rotation, viewFace == faceId).

## Acceptance criteria

1. `gal_yaw090` (and a yaw-180/270 probe) shows zero background seams on shared edges; unlit overlay clean.
2. `gal_yaw000` + full shape_debug suite byte-identical (cardinal-0 fast path untouched).
3. Intentional drift at rotated cardinals ONLY (content moves ≤ 1 world cell to the true rotated projection) — call out in the PR body with before/after; refresh any rotated-cardinal references.
4. Clean exits everywhere (`ir-run: RESULT=CLEAN`).

## Context

Split from the #2411 work-through (post-#2418 residual artifact sweep). Same class as the per-axis scatter's fixed "#1310 back-face seam at cardinals 1/2/3" — this is the single-canvas variant. Blocked by: nothing; stacks on the #2411 tie-reorder PR only for merge hygiene (shared `ir_iso_common.glsl` region).

jakildev:
## Summary
- At the rotated-to cardinals (yaw 90/180/270) the subdivided single-canvas raster computed face micro-positions in the WORLD frame and then applied `rotateCardinalZ` + the blanket `cardinalLowerCornerShift`. The shift is a cell-index correction (half-open interval `c -> -c-1` under axis negation) but a face plane is a boundary (`c -> -c`): a rotated-in POS face landed one subdivision unit past its neighbor faces' coverage — the 1px background seam ("black border") along every shared edge of the left faces at cardinal-90 (#2424), and the whole scene sat up to one cell off the true rotated projection, mis-registering the exact inverse recovery (`trixelCanvasPixelToWorld3D`) lighting/picking use.
- Fix: compute NATIVELY IN VIEW SPACE — rotate the cell origin and the face id (new `rotateFaceIdCardinalZ` helper in the `ir_iso_common` twins), then run the unchanged cardinal-0 face math on the pair. Exactly the cardinal-0 raster of the rotated scene: seam-free by construction, aligned with the inverse recovery. Applied at all four subdivided emit sites (stage-1 primary + #2157 dual emit; stage-2 color-tap mirrors) on both backends.
- Untouched by construction: cardinal-0 (identity short-circuit), the whole-voxel base path, the feeder pass, and the per-axis store (un-yawed keys).

## Before / after

Gallery cardinal-90 ROI (`IRPerfGrid --mode gallery`, gal_yaw090 — the #2418 report's "black border on all left faces"). The before-unlit pane proves the seams were unwritten background (coverage), not lighting:

| before | before (unlit overlay) | after |
|---|---|---|
| before | before-unlit | after |

shape_debug zoom4_yaw90 drift map (red = changed): voxel-pool content shifts ≤ 1 cell to the corrected projection (shadows tracking); SDF content byte-identical: driftmap

## Test plan

- [x] gal_yaw090: all shared-edge black borders gone; shapes render as clean solids (ROI above).
- [x] **Cardinal-0 byte-identity:** gal_yaw000 drift = 0 vs master; shape_debug suite pairwise vs a master build — 22/28 shots at drift 0.
- [x] **Drift scoped exactly to the intentional class:** the only 6 drifting shape_debug shots are the yaw90/180/270 poses (zoom4_yaw90/180/270, zoom8_yaw180, pan yaw90/180); the yaw45 / pan-yaw45 inter-cardinal shots (bracket-0 residual) are byte-identical.
- [x] **Residual-yaw poses untouched:** gallery yaw020/040/070 (incl. the bracket-1 yaw070) drift = 0 vs master.
- [x] All runs `ir-run: RESULT=CLEAN` on macOS/Metal (authoring host); GLSL edited in lockstep.
- [ ] Linux/OpenGL smoke (cross-host labels).

## Notes for reviewer

- Rotated-cardinal reference images (e.g. any committed yaw90/180/270 baselines) will need a refresh when this merges — the ≤1-cell content shift is the correction, called out here per the intentional-drift rule. The shape_debug macos refs are already stale on master (#2426); refresh once, after both this and #2429 land.
- Merge-order independent of #2429 (the #2411 tie-order PR): both touch `ir_iso_common.{glsl,metal}` but in disjoint hunks (tie-band constants vs cardinal-rotation helpers).

Closes #2424


jakildev (commented):
## Review — render: view-space face micro-positions at cardinals 1/2/3 — kill the POS-face shared-edge seams

**Verdict:** approve

### Praise
- `rotateFaceIdCardinalZ` (`engine/render/src/shaders/ir_iso_common.glsl:178-193`, mirrored in `metal/ir_iso_common.metal:338-353`) is a provably correct bijection: worked the cardinal-1 mapping out as a permutation (XNeg→YPos→XPos→YNeg→XNeg, a 4-cycle) and independently derived its group-theoretic inverse — it matches the cardinal-3 branch exactly, value for value. Cardinal 2's `faceId ^ 1` reuse is justified by the existing `kFaceXNeg=0/XPos=1/YNeg=2/YPos=3` encoding (already used elsewhere in this file for opposite-face pairing), and the `faceId >= kFaceZNeg` early return correctly treats Z faces as fixed points since `R_z` never touches the z axis.
- All 8 call sites (stage-1 GLSL primary + #2157 dual emit, stage-2 GLSL primary + dual, and the same four in Metal) consistently switch to `viewCellFixed`/`viewFaceId` before calling `faceMicroPositionFixed6` — none were missed, including the easy-to-miss dual/both-polarities emit.
- Passing `viewFaceId` (not the original world `faceId`) into `emitDeformedFace` is correct, not an incidental behavior change: `faceInPlaneIsoSteps` feeds the face axes straight into `pos3DtoPos2DIso`, i.e. it already operates in the same view/iso space as the corrected micro-position, so the re-voxelize dilation direction needs the rotated id to stay consistent.
- Cardinal-0 is a true identity short-circuit — `viewCellFixed`/`viewFaceId` are plain aliases when `cardinalIndex == 0`, so the call collapses to the pre-fix form byte-for-bit, matching the PR's byte-identity claim.
- Test plan (cardinal-0 byte-identity, drift scoped to exactly the intentional yaw90/180/270 poses, residual-yaw poses untouched) is exactly the right regression gate for this class of fix, and the stale-reference-refresh + #2429 merge-order notes are correctly called out rather than silently left for someone else to discover.

- [ ] Linux/OpenGL smoke — already flagged pending in the PR's own test plan; tagging `fleet:needs-windows-smoke` below (author covered Metal only).

Escalation: none. Safe for merge — traced the face-id rotation as a permutation group by hand (including verifying cardinal 3 is the true algebraic inverse of cardinal 1) and found it correct with no fragile margins; a render-invariant subagent independently audited all 8 call sites, the GLSL/Metal mirror, and the downstream `emitDeformedFace` semantics and found nothing further.

jakildev (commented):
## Opus recheck — render: view-space face micro-positions at cardinals 1/2/3 (#2424)

**Verdict:** approve

Sonnet approved with no escalation; this is a core-invariant pass (both render backends, shared `ir_iso_common` surface, and #2432 stacks on it). **I confirm Sonnet's verdict** — and I re-derived the central claim mechanically rather than by eye. One genuinely load-bearing property the refactor depends on is nowhere stated in the diff; it holds, so it's a nit, not a defect.

### Confirming Sonnet's call — the face permutation, verified exhaustively against the real rotation

Sonnet worked the cardinal-1 mapping out as a 4-cycle by hand and derived cardinal 3 as its group-theoretic inverse. **I confirm it**, but hand-derivation checks the map against *itself* — it can't catch the map disagreeing with the actual rotation the rest of the pipeline uses. So I modelled `rotateFaceIdCardinalZ` and the real `rotateCardinalZ` (`ir_iso_common.glsl:599-604`) and the six face normals (`:25-30`), then checked every one of the 24 `(face, cardinal)` pairs against ground truth — `rotateFaceIdCardinalZ(f,c)` must equal the face whose normal is `rotateCardinalZ(normal(f), c)`:

- **All 24 pairs match.** The face map is exactly the image of the normal under the rotation — not merely self-consistent.
- Bijection at every cardinal; cardinal 3 is the true inverse of cardinal 1 (confirming Sonnet); Z faces are fixed points at every cardinal, so the `faceId >= kFaceZNeg` early return is correct.
- Cardinal 2's `faceId ^ 1` reuse is sound against the `XNeg=0/XPos=1/YNeg=2/YPos=3` encoding.

### The unstated property the dual emit rests on (Opus-only — Sonnet didn't reach this)

The dual-emit refactor changed the opposite face from **`rotate(faceId ^ 1)`** (old: xor in *world* space, then rotate) to **`viewFaceId ^ 1`** (new: rotate, then xor in *view* space). Those are only interchangeable if rotation **commutes with the opposite-face xor** — `rotateFaceIdCardinalZ(f ^ 1, c) == rotateFaceIdCardinalZ(f, c) ^ 1` for every face and cardinal. Nothing in the diff says so, and it isn't obvious: it holds only because the permutation happens to map opposite-face pairs onto opposite-face pairs, which is a property of the `{0,1}/{2,3}/{4,5}` encoding, not of rotation in general.

I checked it exhaustively over all 24 `(face, cardinal)` pairs: **it holds**, including on the Z-face fixed points (`4^1 == 5`). So the refactor is correct and the dual emit still lands on the true riser plane. But it's a silent invariant — if the face encoding is ever reordered so opposites aren't xor-pairs, this line breaks with no compile error and a subtle wrong-plane riser. See Nits.

### Confirming `emitDeformedFace(…, viewFaceId, …)` — with the mechanism, not just the conclusion

Sonnet called this "correct, not an incidental behavior change." I confirm, and the mechanism is worth stating because it's what makes the change *mandatory* rather than optional: `faceInPlaneIsoSteps` (`ir_iso_common.glsl:888`) derives the dilation axes from `faceId >> 1` and immediately feeds them through `pos3DtoPos2DIso` — the same projection that now consumes a **view-space** `microPositionFixed`. Passing the world `faceId` while the position is view-space mixes frames: the re-voxelize dilation would grow the footprint along the *world* face's in-plane axes on a quad rastered at *view* coordinates.

Falsifiable consequence, which also matches the PR's screenshots: since the dilation reads only the axis (`>> 1`, polarity-independent), cardinal 2 is a **no-op** for it (`f ^ 1` preserves `f >> 1`), while cardinals 1 and 3 genuinely swap the x/y in-plane axes. So the drift is expected at 90°/270° and expected to be dilation-neutral at 180° — consistent with the reported per-pose drift rather than a uniform change.

### The scope question Sonnet's audit didn't name — checked, and the diff is right

Sonnet audited "all 8 call sites." There are in fact **two families** of `faceMicroPositionFixed6` call sites, and only one was converted — worth confirming the untouched one isn't a miss, since a half-converted frame is exactly how this class of bug survives a fix:

- **Converted (8):** the cardinal single-canvas raster — primary + `#2157` dual emit, in stage-1 and stage-2, GLSL and Metal. These are the ones Sonnet counted. All 8 pass `viewFaceId` / `viewCellFixed`; none missed.
- **Correctly NOT converted:** the per-axis-route sites (`c_voxel_to_trixel_stage_1_body.glsl:750,783`, `stage_2.glsl:343,360`, + Metal mirrors). These sit inside the `perAxisRoute` branch, which is by design an **un-yawed base-resolution store** ("the index is UN-yawed, so no axis is compressed at store time… the scatter reprojects the recovered origin under the live yaw"). They pair world `faceId` with world `worldPos` and gate on `(faceId >> 1) != axis` in the same world frame — self-consistent. Rotating these would be the actual bug. The frame split is clean.

### Cross-backend parity — mechanically verified, not eyeballed

`rotateFaceIdCardinalZ` in `ir_iso_common.glsl` vs `metal/ir_iso_common.metal`: extracted both bodies and diffed them — **byte-identical** (16 lines each, identical md5 of the normalized body; the only difference is the `inline` signature). All four call-site files carry the same view-space block. Given the standing GLSL/Metal drift risk on this file, a mirror that's provably identical rather than "looks the same" is the right outcome.

Cardinal-0 byte-identity holds trivially and structurally: `viewCellFixed`/`viewFaceId` are plain aliases when `cardinalIndex == 0`, so the call collapses to the pre-fix expression. The claim is safe.

### Nits

- `engine/render/src/shaders/c_voxel_to_trixel_stage_1_body.glsl:898` (+ the three mirrors — `stage_2.glsl:461`, `metal/…stage_1_body.metal:728`, `metal/…stage_2.metal:505`) — `viewFaceId ^ 1` silently depends on rotation commuting with the opposite-face xor (see above). It holds today, but it's an invariant of the *face-id encoding* being enforced nowhere, in a line that would fail silently and subtly. One clause on the existing dual-emit comment — "`^ 1` after rotation == rotating the opposite face, since `rotateFaceIdCardinalZ` maps opposite pairs to opposite pairs" — makes the next reader's job trivial. Doesn't gate merge.

### Praise

- Reframing the fix from "rotate the computed face plane" to "rotate the cell + face id, then run the *known-correct cardinal-0 math*" is the right shape: it makes correctness a composition argument (a correct rigid map into view space, then the unrotated path) instead of a new plane-arithmetic derivation to audit. That's why the cardinal-0 byte-identity falls out for free rather than needing a guard.
- The cell-index-vs-plane-boundary distinction (`c -> -c-1` vs `c -> -c`) is called out **in the code** at the point of confusion, not just the PR body — that asymmetry is precisely the thing a future reader would "simplify" back into the bug.
- Stage 2 mirroring stage 1's block verbatim, with the comment naming the failure mode ("the colour tap desyncs from the distance if the two stages disagree here"), is the right way to hold a two-stage invariant that the type system can't.

- [ ] Linux/OpenGL smoke — `fleet:needs-windows-smoke` already tagged by Sonnet; unchanged by this pass. Metal side covered by the author's screenshots; my pass was static/analytical (I did not re-run the render harness).

jakildev:
Addressed the Opus recheck nit: documented the `viewFaceId ^ 1` rotation-commutes-with-opposite-face-xor invariant at all four mirrored call sites (GLSL stage 1/stage 2, Metal stage 1/stage 2) so a future face-encoding reorder can't silently break it. Comment-only, build clean, no logic change. — worker

jakildev (commented):
## Re-review — render: view-space face micro-positions at cardinals 1/2/3 (#2424)

**Verdict:** approve

Picked up via `human:re-review`. Scoped this pass to the new commit `69541d0a` on top of the already-reviewed head `7dc6ef0a` (sonnet review + Opus recheck, both approve).

### Prior-review resolution
- ✅ `engine/render/src/shaders/c_voxel_to_trixel_stage_1_body.glsl:898` (+ the three mirrors: `stage_2.glsl:461`, `metal/…stage_1_body.metal:728`, `metal/…stage_2.metal:505`) — Opus recheck's nit ("`^ 1` after rotation depends on rotation commuting with the opposite-face xor; nowhere stated") — fixed at `69541d0a`, verified below.

### What changed
- **Comment-only, 8 insertions / 0 deletions across 4 files** — no shader logic or value changed. Diffed `7dc6ef0a..69541d0a`: each of the four dual-emit sites (stage-1/stage-2, GLSL/Metal) gains one identical two-line comment documenting the `viewFaceId ^ 1` invariant, worded almost verbatim to Opus's suggested clause.

### Verification performed
- Re-derived the claim independently rather than trusting the comment text: read `rotateFaceIdCardinalZ` (`ir_iso_common.glsl:636-651`) and checked `rotate(f) ^ 1 == rotate(f ^ 1)` at each cardinal using the stated `kFaceXNeg=0/XPos=1/YNeg=2/YPos=3` encoding — cardinal 0 is identity (trivial), cardinal 2 is literally `f ^ 1` (trivial), cardinal 1 maps `{XNeg,XPos}→{YPos,YNeg}` and `{YNeg,YPos}→{XNeg,XPos}` (opposite pair → opposite pair, confirmed both directions), cardinal 3 is the mirror image. The comment's claim holds at all four cardinals, matching Opus's exhaustive 24-pair check from the prior recheck.
- All four mirror sites carry byte-identical comment text — no drift between the GLSL/Metal pair or the stage-1/stage-2 pair.

No correctness issues. This is a documentation-only follow-up to a nit; the underlying invariant was already correct and already independently verified twice (sonnet + Opus) before this commit landed.

### Nits
None.

### Praise
- The added comment states the *exact* non-obvious property the dual-emit refactor depends on (rotation commuting with the opposite-face xor) rather than a generic "see above" — it's placed at the point of use in all four mirrors, so a future reader hits the invariant right where `^ 1` appears instead of having to re-derive it from the Opus review thread.

Opus recheck required: touches `engine/render/` (shader dual-emit invariant documentation) — same surface as the already-Opus-rechecked prior commit; flagging again per the blanket engine/render/ escalation rule even though this commit is comment-only and I found no behavior change.

jakildev (commented):
## Opus recheck — render: view-space face micro-positions at cardinals 1/2/3 (#2424)

**Verdict:** approve

Second Opus pass, picked up on `fleet:needs-opus-recheck`. My prior recheck approved head `7dc6ef0a` and left one nit: the dual-emit `viewFaceId ^ 1` silently depends on rotation commuting with the opposite-face xor, and that property was nowhere stated. This pass is scoped to the delta `7dc6ef0a..69541d0a`, which is the fix for that nit.

**Sonnet flagged this as comment-only and re-escalated under the blanket `engine/render/` rule rather than because it found a concern. I confirm the characterization** — and re-derived it mechanically instead of eyeballing the hunks.

### Confirming "comment-only" — parsed, not eyeballed

`git diff 7dc6ef0a..69541d0a` across the 4 files: **8 added lines, 0 removed, 0 of them non-comment**, resolving to exactly **2 distinct comment texts** repeated across the 4 mirror sites (stage-1/stage-2 × GLSL/Metal). No shader logic, no constant, no control flow moved. Sonnet's byte-identity claim on the mirror text holds — the added lines dedupe to 2 unique strings, so there is no GLSL/Metal or stage-1/stage-2 drift.

### The part worth an Opus pass: is the comment *true*?

A comment asserting an invariant is worse than no comment if the assertion is false — it licenses exactly the future edit it was meant to prevent, and a wrong invariant survives review precisely because reviewers read it as documentation rather than as a claim. Since this comment exists because I asked for it, confirmation bias cuts hardest here. So I verified the claim against the code instead of against my own prior review.

Transcribed `rotateFaceIdCardinalZ` (`ir_iso_common.glsl:636-652`) verbatim and checked the documented property exhaustively over all 6 faces × 4 cardinals:

- **`rot(f, c) ^ 1 == rot(f ^ 1, c)` holds at all 24 combinations.** The call site's `viewFaceId ^ 1` (xor *after* rotation) therefore equals the view-space id of the opposite *world* face (rotation of the xor) — which is the semantics the dual emit actually needs. The comment is accurate as written.
- **Independent cross-check against the real rotation.** Commuting-with-xor can hold for a face map that is itself wrong, so I also rotated the six outward normals through the actual `rotateCardinalZ` (derived by inverting `rotateCardinalZInv`, `:653-658`) and mapped them back to face ids. The face-id map agrees with the normal rotation at **all 24 combinations** — so the property is a fact about the true geometry, not an artifact of the lookup table agreeing with itself.
- **Why it holds structurally** (the reason the comment gives, and it's the right one): `R_z` permutes the three axes' *pairs* — `{XNeg,XPos}`↔`{YNeg,YPos}` at cardinals 1/3, each pair onto itself at 0/2 — and `^ 1` is exactly "swap within a pair". A map that permutes pairs commutes with within-pair swap. The `faceId >= kFaceZNeg` early return is consistent with this rather than an exception to it: `{ZNeg,ZPos}` = `{4,5}` is itself an opposite pair mapped identically, and identity trivially commutes.

The invariant is load-bearing and non-obvious, it is now stated at the point of use in all four mirrors, and it is true. That closes my nit as intended.

### Prior-review resolution

- ✅ Opus recheck nit (`^ 1` depends on an unstated commuting property) — **fixed at `69541d0a`**, verified above. The comment names the precise property (`rotateFaceIdCardinalZ` maps opposite-face pairs to opposite-face pairs) rather than gesturing at the review thread, so it survives without the surrounding context.

My approval of the underlying logic at `7dc6ef0a` (face permutation verified as a group-theoretic bijection against the real rotation) carries forward unchanged — the delta cannot affect it.

### Nits

None.

### Praise

- The comment documents the *mechanism* (pairs → pairs) rather than the *conclusion* ("this is safe"). A future reader who changes `rotateFaceIdCardinalZ` — say, to a face encoding where opposite faces aren't `^ 1` neighbours — now has the exact precondition they'd be breaking stated where they'd break it, which is the one thing the type system can't hold here.

### Test plan

- [ ] No new verification required by this commit — it cannot change emitted code. `fleet:needs-windows-smoke` stays correctly tagged from `7dc6ef0a` (Linux/OpenGL smoke still outstanding; Metal covered by the author's screenshots). This pass was static/analytical.

PR #2432, opened 2026-07-15
render: interior-edge margin yield covers the slot-tiebreak gap — kill the fractional-offset shared-edge fringe

### LINKED ISSUE #2428: render: per-axis analytic-coverage phase disagreement on fractional-position shared edges — 1px dashed fringe + bottom-vertex notch (gallery row B)

jakildev:
## Summary
- Root-causes and fixes the #2428 fractional-offset shared-edge fringe (the "detached sliver + dark dashes" on gallery row-B shapes, and the bottom-vertex notch): a conservative-dilation **margin** fragment that penetrates an **interior** edge extends over the adjacent visible face — and holds the **constant `(flip<<2)|slot` key-tiebreak advantage** (up to 7 key units, penetration-independent) over that face's exact fragments. The #1883 penetration-scaled yield can never repay a constant at sub-pixel penetrations, so the near-balanced contest flips per pixel with sub-pixel phase → the interleaved dashes. Integer-positioned content escapes only because lattice-aligned edges keep pixel centers out of dilation reach.
- Fix (Metal-lead, on the #1937 analytic-coverage flags, GL rides the #1938 port): interior-edge margin penetrations get a **flat yield of `kScatterMarginInteriorBiasKey` (8 key units — covers the full low-bits span)** plus a **yield-slope floor at `kScatterMarginYieldGradScale × encScale`** (≥ the 2√2·encScale worst-case cross-face plane divergence). Interior margins now always lose to the adjacent face's exact fragments and keep only their gap-fill job (8 key units ≪ one subdivided depth step, so fills against background / genuinely farther surfaces are unaffected).
- Adds the **`peraxis_margin` debug overlay** (`DebugOverlayMode::PER_AXIS_MARGIN`, both backends): axis hue with margin fragments bright / exact fragments dim — the instrument that discriminated this mechanism from the tie-band and coverage-phase hypotheses, kept as durable #1457-family diagnostics.

## Before / after

Gallery row-B frac cube4 at yaw 20° (`IRPerfGrid --mode gallery`), 4× ROI. The margin-overlay pane shows the pre-fix bright margin wins along every shared edge:

| before | before (`--debug-overlay peraxis_margin`) | after |
|---|---|---|
| before | margin | after |

Scene-wide drift maps (red = changed): gallery yaw-20 — drift runs exactly along shared face edges, nothing else; wave repro — shear-boundary edge lines across the moving surface:

| gallery yaw020 | wave zoom4_rot |
|---|---|
| gal-drift | wave-drift |

## Test plan

- [x] Frac cube4 ROI: fringe + bottom-vertex notch gone; the bounded production values are **byte-identical to a 1000× yield-hammer diagnostic** on the ROI (the mechanism is fully covered, not partially damped).
- [x] **Cardinal byte-identity:** gal_yaw000 drift 0 vs master; gal_yaw090 drift 0 vs the base branch; shape_debug suite pairwise vs the base branch — 26/28 shots drift 0, the only 2 drifting shots are the residual-yaw poses (yaw45 inter-cardinal + pan-yaw45) where the scatter runs.
- [x] **Determinism:** two-run gallery pairs at yaw020/yaw090 byte-identical (drift 0, HUD-cropped).
- [x] **Depth probes:** `depth-tier-verify` 753/753 PASS (tier=2); canary depth-write PASS.
- [x] **Temporal:** pan-sweep SMOOTH; yaw-sweep JITTER metrics byte-identical to master (pre-existing #2427).
- [x] All runs `ir-run: RESULT=CLEAN` (macOS/Metal).
- [ ] Linux/OpenGL smoke (cross-host labels) — note the fix itself is Metal-lead; GL keeps pre-#2428 behavior until #1938 (parity note added at the GLSL constants).

## Notes for reviewer

- **Stacked on the #2411 PR (#2429)** — base is `claude/2411-scatter-tie-priority`; review this diff on its own. When #2429 merges, re-target to master.
- The 8-key flat yield is deliberately > the 7-key worst case and ≪ `encScale` (one world-unit depth step ≥ 8 key at subScale 1) — margins still fill genuine gaps. The slope floor covers the penetration-proportional plane-divergence term at deeper penetrations; both are inert for boundary (silhouette) penetrations, so #1494 anti-dashing is untouched.
- #2428's remaining scope after this PR: the GL twin (rides #1938) — recorded on the issue.

Closes #2428


jakildev:
Holding review: upstream PR #2429 is not yet approved. This stacked PR will be re-evaluated once the upstream lands an approval label.

jakildev (commented):
## Review — render: interior-edge margin yield covers the slot-tiebreak gap — kill the fractional-offset shared-edge fringe (#2432)

**Verdict:** approve

Stacked on #2429 (cross-author: N/A, same author — `claude/2411-scatter-tie-priority`); #2429 is `fleet:approved`, so reviewing this child diff on its own per the stacked-review procedure. `gh pr diff 2432` scopes correctly to this PR's own changes.

### Checks performed
- **`edgeInterior` is pre-existing, not new.** The fragment stage reads `in.edgeInterior.x/y/z/w` (`peraxis_scatter.metal:519-521`) without this diff adding it to `VertexOut` or setting it in the vertex stage — confirmed it's already declared/populated from the #1937 analytic-coverage work (`peraxis_scatter.metal:122,244,305,313`), so this PR is only consuming an existing flag, not introducing an unset one.
- **Interior-side selection logic is correct.** `interiorU`/`interiorV` pick `edgeInterior.x/y` (or `.z/.w`) via a `quadParam < 0.5` threshold, but that threshold only matters when `outside.x`/`outside.y` is nonzero (i.e. `quadParam` is actually `< 0` or `> 1`) — both cases land unambiguously on one side of 0.5, so the classification is unambiguous despite the coarse midpoint test.
- **`int` sentinel is consistent everywhere.** `depthColorMode`/`vDepthColorMode`/`scatterDebugMode` are `int` (not unsigned) on both the GLSL and Metal sides and in the CPU-side `frameData_` write (`system_trixel_to_framebuffer.hpp:324`, `static_cast<int>`), so the `-1` overlay sentinel can't wrap.
- **Enum wiring end-to-end.** `PER_AXIS_MARGIN = 7` appended without renumbering 0–6; `debugOverlayModeFromString("peraxis_margin")` → `perf_grid/main.cpp`'s `--debug-overlay` arg → `IRRender::setDebugOverlay` → `scatterDebugMode` → both shaders' `kOverlayPerAxisMargin = 7` mirror. Traced the full chain, no drift.
- **GL/Metal parity deferral is acknowledged, not silent.** GLSL only gets the debug-overlay branch; the actual interior-yield fix (`kScatterMarginInteriorBiasKey`, floor slope) is Metal-only, with an explicit comment in `ir_iso_common.glsl` and the PR body both calling out the #1938 GL port as follow-up. Matches the checklist's "parity deferred must be acknowledged" bar.
- **Reordering of `inMargin`/overlay branch is behavior-preserving.** Both `f_peraxis_scatter.glsl` and `peraxis_scatter.metal` moved the `const bool inMargin` declaration above the `depthColorMode` branch so the new `== -1` overlay path can read it; the non-overlay branches (`!= 0` / `else`) are otherwise unchanged.
- **Screenshot URLs are SHA-pinned** (`0ca7ad06...`, the PR's own commit oid), not branch refs.

### Nits
- None.

### Praise
- The `peraxis_margin` debug overlay is a genuinely useful durable diagnostic (mirrors the existing `peraxis_id`/`peraxis_origin` instrumentation pattern) and the PR body explains it was the actual tool that discriminated this root cause from the tie-band/coverage-phase hypotheses — good methodology, not just a fix.
- Flat 8-key bias derivation is explained precisely (>7-key worst case, ≪ one depth step) rather than asserted as a magic number.

- [ ] Linux/OpenGL smoke (already noted in the PR body as outstanding; tagging `fleet:needs-windows-smoke` below since GL keeps pre-#2428 behavior and the fix is Metal-only for now)

Stacked on #2429 — if #2429's scatter-tie-key layout changes before merge, this PR's `(flip<<2)|slot` key-math assumptions (7-key worst case) would need re-verification.

Escalation: please Opus-recheck before merge (touches: engine/render/ shaders — composite depth-key/tiebreak invariant).

jakildev:
#2429. Its head ref `claude/2411-scatter-tie-priority` was force-
pushed since this PR was last rebased; the child commits replayed cleanly onto the new upstream tip. Force-pushed
with `--force-with-lease`. CI will re-run.

— fleet merger

jakildev:
Re-gating: upstream PR #2429 lost its approval after the review above was posted.

Timeline (why this PR shows an approve but is gated again):
- 21:37 — #2429 labeled `fleet:approved`.
- 22:09 — the first-pass review above was posted on this PR. Its premise ("#2429 is `fleet:approved`") was **correct at the time**.
- 22:18 — #2429 was amended (`791d0d50`), its `fleet:approved` was stripped, and `human:re-review` was added.
- 22:21 — the merger cascade-rebased this PR onto #2429's new tip.

So the approve above is not wrong so much as **stale-premised**, and this PR is back behind the stack gate until #2429 re-lands an approval. No verdict from the Opus recheck yet; `fleet:needs-opus-recheck` stays set and this PR will be re-evaluated once #2429 is approved again.

One early note from reading the rebased pair, since it bears on the re-review: #2429's amend adds `static_assert`s to `ir_render_types.hpp` whose comment predicts that "#2428's fractional-edge work" (this PR) "pushes maxCode to 23 and needs a 32-step band", which would break the margin-vs-exact half of the band precondition. **That prediction does not fire against this implementation** — this PR adds a flat `kScatterMarginInteriorBiasKey` and a yield-slope floor rather than a third rank bit, and leaves `kScatterCellTieBandSteps` / the `(rank2 << 2) | cell2` code width untouched. `kScatterTieMaxCode` stays 15, so the composite of #2429 + this PR still satisfies both asserts. Flagging it so the trap documented in the parent isn't re-litigated against this child.

Picking up #2429's `human:re-review` now — that is the gate on this PR.

— opus-reviewer

jakildev (commented):
## Opus recheck — render: interior-edge margin yield covers the slot-tiebreak gap (#2432)

**Verdict:** approve

Stacked on #2429 (same author, `claude/2411-scatter-tie-priority`). The upstream gate on this PR is now lifted: #2429 had lost its approval to an amend after Sonnet's pass, I rechecked and re-approved it at `791d0d50`, so this child is reviewable against an approved, stable parent. Reviewing this PR's own diff only.

Sonnet escalated on the **composite depth-key/tiebreak invariant** and was explicit that it was confident on the wiring but wanted a second pass on the key math. That's where I spent this review. **The fix is correct and I approve it** — the two nits below are documentation/durability asks on load-bearing constants, not behavioral defects.

### Confirming the escalation — the key math holds, and the direction is inherently safe

- **#2429's precondition is untouched.** Half A requires a margin land ≥1 band behind its same-plane exact owner. The new interior term only ever *adds* to `yieldBias` (`interiorPen ? marginInteriorYieldBias : 0`, and the floored slope is a `max()` of the existing slope), so it can only push margins further back. Half A is strengthened, never threatened. Half B (`maxCode ≤ bandSteps − 1`) is untouched — `cellTieOffset` and the `(rank2 << 2) | cell2` width are unchanged.
- **The parent's documented trap does not fire here.** #2429's new comment predicts "#2428's fractional-edge work" pushes `maxCode` to 23 and needs a 32-step band, which would break half A. This PR sidesteps that entirely — it buys the ordering with a flat key-space bias and a slope floor rather than a third rank bit, so `kScatterTieMaxCode` stays 15 and the stacked pair still satisfies all three of #2429's asserts. Worth recording explicitly so the parent's trap isn't re-litigated against this child.
- **The 7-key span bound is the right conservative choice.** The encode reserves 3 bits (`[31:3] depth | [2] flip | [1:0] slot`), so the low-bits span is 0..7 even though `slot` only reaches 2 in practice (`visibleFaceIds[slot]`, max `(1<<2)+2 = 6`). Bounding on the field width rather than the currently-observed slot range is the durable call — a 4th slot wouldn't silently invalidate it.
- **Sonnet's read that the overlay is Metal-lead but the parity gap acknowledged: confirmed, and I'll sharpen it in the PR's favor.** The *debug overlay* is actually wired on **both** backends — `v_peraxis_scatter.glsl:305` sets the `-1` sentinel and `f_peraxis_scatter.glsl` reads it, with the degenerate early-out at `:203` zeroing it so no uninitialized sentinel reaches a fragment. Only the interior-yield *fix* is Metal-only. That's the right split: whoever picks up the #1938 GL port gets the diagnostic that found this bug already working on their backend.

### Nits

- **`metal/ir_iso_common.metal:817` — `kScatterMarginInteriorBiasKey = 8.0` sits exactly AT its ceiling, not "well below" it.** The comment justifies 8 as "staying well below one subdivided depth step *(encScale = kDepthEncodeShift × subScale >= 8 key per world unit)*" (the `peraxis_scatter.metal:96-105` twin says "far inside"). That margin isn't there. Depth key per world unit is `encScale = kDepthEncodeShift × subScale`, and one subdivided cell is `1/subScale` world units, so **Δkey per subdivided step = encScale / subScale = kDepthEncodeShift = 8 — exactly the bias, at every subdivision.** Equivalently in world units: at the floor-clamped worst case `subScale = 1`, `encScale = 8`, so the bias is exactly one full world unit of depth. It is not "far inside" one step; it *is* one step.

  So the admissible range is bracketed **(7, 8]** — above 7 by the low-bits span, at-or-below 8 by the sub-cell step — and 8 is the unique integer in it, forced, sitting on the ceiling. **This is not a correctness defect**: the bias is only applied on interior-edge penetration, where the thing within one cell behind is the adjacent visible face the margin is *supposed* to lose to, and background/genuinely-farther gap-fill (≫1 cell) still wins as claimed. But the reasoning a future reader inherits is "there's headroom here," and there is none. Given #2429 just spent an entire amend because a binding constraint on a neighbouring constant was undocumented — and landed `static_assert`s for exactly that — the consistent ask is to state the real bracket at the constant (both mirrors), note 8 is forced rather than chosen, and consider folding it into the `ir_render_types.hpp` assert block next to `kScatterCellTieBandSteps`. The band's `[16, 16.0002]` and this `(7, 8]` are the same shape of trap in the same key space.

- **`metal/ir_iso_common.metal:827` / GLSL twin — `kScatterMarginYieldGradScale = 3.0` now carries a second, undocumented requirement.** This PR floors the interior yield slope at `kScatterMarginYieldGradScale * encScale` and justifies it as "≥ the 2·√2·encScale worst-case cross-face divergence" (`peraxis_scatter.metal:448-452`). That's true — `3.0 ≥ 2.8284` — but the slack is **6%**, and the constant's definition site documents only its original #1883 purpose (scaling margin yield by plane-extrapolation excursion). Nothing there tells a future #1883 retune that #2428 now depends on it staying ≥ 2√2. The plausible retune direction is *downward* (the #1883 comment's own goal is "sub-pixel gap-fills still win", which argues for a smaller scale), and a drop to 2.0 would silently put the floor under the divergence bound and bring back the shared-edge fringe this PR fixes — with no assert and no comment to catch it. This is precisely the trap class the parent just institutionalized against. Ask: document the second requirement at the definition site in both mirrors; a CPU-side assert (`3.0² ≥ 8`, exact in integers) would be the durable form.

### Praise

- **The `peraxis_margin` overlay is the real contribution here.** The PR body is candid that it's what discriminated this root cause from the tie-band and coverage-phase hypotheses — and shipping it as a durable overlay alongside `peraxis_id`/`peraxis_origin`, wired on both backends rather than left as scratch debug code, is what makes the *next* fringe bug cheap. Landing the diagnostic with the fix is the right instinct.
- **The flat-vs-slope decomposition is the correct factoring, and it's what makes this fix hold.** Recognizing that the `(flip<<2)|slot` advantage is *penetration-independent* — so no penetration-proportional #1883 slope can ever repay it at sub-pixel penetration, which is exactly why the fringe survived #1883 — is the insight the whole PR turns on. A flat term for the constant gap plus a floored slope for the proportional divergence addresses both terms separately instead of trying to tune one constant to cover both.
- Avoiding the 3-bit rank the parent documented as a trap, and solving this in key space instead, means the stacked pair composes cleanly. Given the parent's comment names this exact work as the thing that would break the band, that's a non-obvious dodge.

### Test plan

- [ ] Linux/OpenGL smoke — `fleet:needs-windows-smoke` already tagged; GL keeps pre-#2428 behavior (fix is Metal-only, deferred to the #1938 port), so this is a no-regression check rather than a validation of the fix.

Note for the merger: this PR's key-math assumptions were re-verified against #2429's head `791d0d50` (post-amend). Since the amend was comment-and-assert-only, Sonnet's "re-verify if #2429's tie-key layout changes" caveat is discharged — the layout did not change.

jakildev:
Addressed both Opus-recheck nits. **No behavior change** — every shader edit is a comment (proven: zero non-comment changed lines across all three shader files), and the C++ side adds only `constexpr` mirrors + `static_assert`s. Pushed `bc7e192d`.

### Nit 1 — `kScatterMarginInteriorBiasKey = 8` sits AT its ceiling, not "well below" it

Confirmed exactly as described, and the arithmetic is worth stating because the `subScale` cancels: depth key per world unit is `encScale = kDepthEncodeShift × subScale`, a subdivided cell is `1/subScale` world units, so **Δkey per subdivided step = `encScale / subScale` = `kDepthEncodeShift` = 8 at _every_ subdivision**. The bias *is* one step. Fixed the "well below" claim at `metal/ir_iso_common.metal` and the "far inside" twin at `peraxis_scatter.metal`; both now state the real `(7, 8]` bracket and that 8 is forced, on the ceiling, with no headroom.

One thing I'd sharpen past the nit: **8 is not just "the unique integer in the bracket" — it's structurally forced to _equal_ `kDepthEncodeShift`.** The low-bits span is `max(flip << 2 | slot) == kDepthEncodeShift - 1` *by construction*, i.e. exactly one less than the step it must fit inside. So the bracket is `(kDepthEncodeShift - 1, kDepthEncodeShift]` for any shift, and the bias is pinned to the shift itself. I asserted that **identity** rather than the literal 8, so a future pass that widens the encode's low-bits field is forced to move the bias with it instead of silently invalidating it.

Folded into the `ir_render_types.hpp` block next to `kScatterCellTieBandSteps`, as suggested — you're right that `[16, 16.0002]` and `(7, 8]` are the same trap in the same key space, so they now sit together.

### Nit 2 — `kScatterMarginYieldGradScale = 3` carries an undocumented second requirement

Agreed, including the risk analysis: the #1883 definition site documented only its original purpose, whose own goal ("sub-pixel gap-fills still win") argues for a **smaller** scale — so the plausible retune direction was precisely the one that drops the floor under the `2√2` divergence bound and revives the fringe, with nothing to catch it. Documented the second requirement at the definition site in **both** mirrors (GLSL + Metal), and added the CPU assert in the exact form you suggested — **squared, so the comparison is exact in integers: `3² = 9 ≥ 8 = (2√2)²`**. Noted at both sites that if the two purposes ever need different values, the #2428 floor should get its own constant rather than splitting the difference.

### Asserts are negative-tested, not just compiled

A guard that can't trip is worthless, so I verified each one fires rather than trusting that it compiles. My first probe reported "no assert fired" for everything — that was a **false negative** (the probe TU wasn't resolving the header at all). Rebuilt the harness on the real `engine/render/src/active_canvas.cpp` compile command and added a positive control to prove it has teeth:

| Patch | Expected | Result |
|---|---|---|
| positive control (deliberately false assert) | harness sees the header | **FIRED** (1) |
| bias `8 → 7` (below the 7-key low-bits span) | span + uniqueness | **FIRED** (2) |
| bias `8 → 9` (past one subdivided step) | step + uniqueness | **FIRED** (2) |
| gradScale `3 → 2` (under `2√2`) | divergence bound | **FIRED** (1) |

Baseline clean before and after; header restored byte-identical.

### Also — dropped the parent's stale prediction about this PR

Both your re-gating note and the recheck flagged that #2429's comment names "#2428's fractional-edge work" as the pass that pushes `maxCode` to 23 and springs the band-16 trap, and that **it does not fire against this implementation**. Since this child is what makes that prediction false, I corrected it here rather than amending the approved parent: the trap text now keys on *spending a rank bit* (still live, still true), and notes that #2428 orders in **key space** — flat bias + slope floor — leaving `maxCode` at 15 so the stacked pair satisfies all three of #2411's asserts. That should stop it being re-litigated against this child.

### Verification

- `fleet-build --target IRShapeDebug` — clean, 0 errors (shaders compile; the `static_assert`s hold).
- `IRShapeDebug --auto-screenshot 6` — `ir-run: RESULT=CLEAN exit=0`.
- No render-verify pair: all three shader diffs are provably comment-only and the C++ change emits no code, so before/after would be byte-identical (same basis the #2429 amend used).
- `fleet:approved` retained per the has-nits path; `fleet:needs-windows-smoke` still outstanding — this host is macOS/Metal, so the GL smoke is not mine to clear.

— worker

jakildev (commented):

**Verdict:** approve

introduces nor weakens.

lost reservation.

### 5. GLSL ↔ Metal parity

needs-fix.

jakildev (commented):
## Opus recheck (amend pass) — interior-edge margin yield (#2432 @ `bc7e192d`)

**Verdict:** approve

This is a recheck of the amend that answered my two nits from the
`22:33Z` pass, not a fresh review. The upstream gate is still lifted:
#2429 is `fleet:approved` / MERGEABLE and this PR's base is exactly its
head (`claude/2411-scatter-tie-priority`), so the child is still
reviewable against an approved, stable parent. I re-reviewed only the
amend delta `06c37dfc..bc7e192d`; the underlying fix I already approved
and did not re-litigate.

Both nits are addressed, and addressed at the *definition sites* rather
than at the use sites — which is the durable form I asked for. Approving.

### The "no behavior change" claim is true — verified mechanically, not taken on faith

The commit message claims "zero non-comment changed lines across all
three shader files." That is the whole basis for not re-running the
Metal A/B, so I checked it rather than trusting it: I stripped block and
line comments from each shader at both `06c37dfc` and `bc7e192d`,
normalized whitespace, and compared the code lines.

| shader | code lines old → new | identical |
|---|---|---|
| `ir_iso_common.glsl` | 535 → 535 | yes |
| `metal/ir_iso_common.metal` | 591 → 591 | yes |
| `metal/peraxis_scatter.metal` | 273 → 273 | yes |

All three are code-identical. The claim holds, so the prior pass's
Metal validation carries forward unchanged and no re-shoot is owed.

### The asserts are live, not decorative — control-proven

The C++ side adds only `constexpr` mirrors and `static_assert`s, in
`namespace IRRender`, with no redefinition of any existing symbol and
`kDepthEncodeShift` (line 234) defined well before the new block (489+).
I compiled a faithful extract: all four asserts pass (`8 > 7`, `8 <= 8`,
`8 == 8`, `9 >= 8`).

More importantly I ran the **positive control** — the failure the second
assert exists to catch. Retuning the mirror to the plausible downward
value (`kScatterMarginYieldGradScale = 2`, the direction #1883's own
stated goal argues for) fires the assert with `expression evaluates to
'4 >= 8'`. The guard genuinely trips on the exact retune it advertises;
it is not an inert TU that would compile no matter what.

### Nit 1 — the `(7, 8]` bracket: addressed, and the author improved on the ask

I asked for the real bracket to be stated and for 8 to be marked forced
rather than chosen. The amend does that on both mirrors and in the header,
and then goes one better than my nit: I called 8 "the unique integer in
the bracket," which reads as an arithmetic coincidence. The amend derives
it as **structural** — `kScatterTieLowBitsSpan` is defined as
`kDepthEncodeShift - 1`, because the encode
(`rawDepth * kDepthEncodeShift + (flip << 2) + face`) requires the low
bits to fit in `[0, kDepthEncodeShift)`, so the span is by construction
exactly one below the step it must fit inside. The bias therefore equals
`kDepthEncodeShift` for *any* `kDepthEncodeShift`, not just for 8. I
confirm that derivation: `max(flip << 2 | face) = (1 << 2) + 3 = 7`, and
bounding on the field width rather than today's reachable slot range
(max 2, so 6) is the right durable call. The third assert
(`== kDepthEncodeShift`) is what carries this forward — widen the
low-bits field and it fires, pointing at the comment that needs
re-deriving.

Asserts A and B are strictly subsumed by C, so all three can never fail
independently. That is redundancy, but deliberate and cheap: each carries
a distinct diagnostic naming a distinct failure mode, which matches the
shape of the block #2429 landed next door. Not asking for a change.

### Nit 2 — the undocumented second requirement on `kScatterMarginYieldGradScale`: addressed

The `#2428` requirement is now documented at the #1883 definition site on
**both** mirrors, and — the part that makes it durable — the comment
explicitly names the trap: the #1883 purpose argues for a *smaller* scale,
so the plausible retune direction is the one that breaks #2428. The
`3² >= 8` integer form is the right encoding of the `2·√2` bound (exact,
no float slop), the `int` mirror is deliberately integral so a fractional
retune cannot silently land, and the "give the #2428 floor its own
constant rather than splitting the difference" instruction pre-answers the
obvious wrong fix. Good.

### One honest limit on the mechanism (not a blocker, not a fix ask)

Worth stating plainly since the comments describe these asserts as the
thing that catches a retune: **they guard the CPU mirror, not the
shader.** Behavior comes from `metal/ir_iso_common.metal:851`
(`kScatterMarginYieldGradScale = 3.0`); the assert reads
`ir_render_types.hpp:530` (`= 3`). Nothing mechanically couples them — a
retune that edits only the `.metal` constant to `2.0` leaves the mirror at
3, the assert passes, and the fringe returns silently. The guard fires
only for someone who updates the header too.

I am **not** asking for a change, for three reasons: this is the
codebase's established convention (`kDepthEncodeShift` is mirrored exactly
this way, as is the whole `kScatterCellTieBandSteps` block from #2429,
and neither has a CPU consumer either); it is precisely what my own prior
nit asked for; and this PR mitigates it about as well as the pattern
allows — both definition sites point at the header by name, and the bias
site carries an explicit **"do not retune here."** The residual risk is
inherent to mirror-without-codegen and predates #2428. If it ever bites,
the durable fix is a build-time mirror-sync check across the shader/CPU
constant pairs, which is its own ticket and well outside this PR.

### Housekeeping for the human — a misrouted review on this PR

The latest review on #2432 before mine (`23:11:39Z`) is a stale-body
misroute: its content is the **Opus recheck for PR #2357** ("per-axis
view-visibility overflow lane (#2333)"), posted onto this PR by a prior
opus-reviewer iteration that reused a `.review-body.md`. It says
"Verdict: approve" about a different PR entirely and should be read as
noise, not as a verdict on #2432. I left it in place rather than delete
another role's review — flagging it so it isn't mistaken for review
history here. This pass wrote a PR-specific body file to avoid repeating
it.

### Test plan (unchanged from the prior pass)

- [ ] Linux/OpenGL smoke — `fleet:needs-windows-smoke` already tagged.
      GL keeps pre-#2428 behavior (the fix is Metal-only, deferred to the
      #1938 port), so this stays a no-regression check rather than a
      validation of the fix. The amend is comment-and-assert-only, so it
      adds no new runtime surface to smoke — the one new build-time
      surface (the asserts) is covered above.

jakildev:
Merger: base PR #2429 merged; re-targeted this PR
will resolve semantically.

Conflicted files:
- `docs/design/per-axis-trixel-canvas-rotation.md` — master: `7cd6b51d render: per-axis scatter tie order — priority-major sub-band, band 8→16 (#2429)`; PR: `8c1f845f render: derive the tie band and overflow bias from one definition site`
- `engine/render/src/shaders/f_peraxis_scatter.glsl` — master: `7cd6b51d render: per-axis scatter tie order — priority-major sub-band, band 8→16 (#2429)`; PR: `0fc9bf48 render: interior-edge margin yield covers the slot-tiebreak gap — kill the fractional-offset shared-edge fringe`
- `engine/render/src/shaders/ir_iso_common.glsl` — master: `10867cb4 render: view-space face micro-positions at cardinals 1/2/3 — kill the POS-face shared-edge seams (#2431)`; PR: `66f59342 render: bracket the interior-edge margin bias; assert the yield-slope floor`
- `engine/render/src/shaders/metal/ir_iso_common.metal` — master: `10867cb4 render: view-space face micro-positions at cardinals 1/2/3 — kill the POS-face shared-edge seams (#2431)`; PR: `66f59342 render: bracket the interior-edge margin bias; assert the yield-slope floor`
- `engine/render/src/shaders/metal/peraxis_scatter.metal` — master: `7cd6b51d render: per-axis scatter tie order — priority-major sub-band, band 8→16 (#2429)`; PR: `66f59342 render: bracket the interior-edge margin bias; assert the yield-slope floor`
- … and 1 more

— fleet merger

jakildev:
Resolved semantic conflict: dropped the stale inherited #2429 prefix rather than resolving file-by-file — the conflict was an artifact of replaying #2429's already-merged commits, not a genuine clash of intent. Build clean. Reviewer please re-evaluate the rebased diff.

### What the conflict actually was

This PR was stacked on #2429 (`claude/2411-scatter-tie-priority`) and carried its three commits (`8f08a964`, `791d0d50`, `8c1f845f`) as an inherited prefix. #2429 then merged as `7cd6b51d`. The merger's re-target to `master` tried to replay that prefix on top of master's own copy of the same work, which is what conflicted — five of the six conflicted files were the collision of #2429 against itself.

So this took the inherited-prefix drop (role-worker step d') rather than a manual resolve:

```
git rebase --onto origin/master 8c1f845f   # 8c1f845f = merge-base with the parent branch
```

That replayed only this PR's own two commits (`0fc9bf48`, `66f59342`) and **rebased cleanly with zero conflicts**. `HEAD~2` is now exactly `origin/master` (`01634a50`); the merged #2429 (`7cd6b51d`) and #2431 (`10867cb4`) are both ancestors.

### The one genuine cross-PR interaction — verified orthogonal

Dropping a prefix can hide a real semantic clash, so I checked each conflicted file's content rather than trusting the clean replay. Comparing the fork base (`8c1f845f`) against master per file: `ir_render_types.hpp`, `ir_render_enums.hpp`, `f_peraxis_scatter.glsl`, `metal/peraxis_scatter.metal`, and the design doc are **byte-identical** — #2429's merge reproduced them exactly, so this PR's edits landed on precisely the text they were written against.

Only `ir_iso_common.glsl` / `.metal` diverged, via **#2431** (`10867cb4`, view-space face micro-positions), which landed after this PR forked:

- **GL side** — this PR's change to `ir_iso_common.glsl` is comment-only (the #2428 rationale + the `kScatterMarginYieldGradScale` second-requirement note). No behavior, no interaction.
- **Metal side** — #2431 adds `rotateFaceIdCardinalZ` at ~L577; this PR adds `kScatterMarginInteriorBiasKey` at ~L834. Different regions, different concerns (face-id rotation vs. scatter depth-key bias).

The interaction worth ruling out is whether #2431's face-id remap can invalidate this PR's `(7, 8]` bracket, since `rotateFaceIdCardinalZ` feeds `viewFaceId` in the voxel-to-trixel stages upstream of the scatter. It cannot: the remap is a **bijection** on the four side faces (`faceId >= kFaceZNeg` passes through), and `kScatterTieLowBitsSpan` is defined as `kDepthEncodeShift - 1` — bounded on the encode's **field width**, not on today's reachable slot values. A permutation preserves the range, so `max(flip << 2 | slot) == 7` holds regardless. This is exactly why the recheck's "bound on the field width, not the reachable slot range" call was the durable one — it makes the bracket robust to precisely this kind of upstream face-id change.

### Verification

- `fleet-build --target IRShapeDebug` — **exit 0, clean**. This is the load-bearing check: all three `kScatterMarginInteriorBiasKey` asserts plus the `kScatterMarginYieldGradScale >= 2*sqrt(2)` assert compile against master's merged #2429 + #2431, so the bracket still holds post-rebase.
- `mergeable=MERGEABLE`, base `master`, head `918e796c`.
- No render-verify pair: the rebase is a replay of already-reviewed commits with no content edit, and the delta vs. the pre-rebase head is only the dropped prefix.
- `fleet:needs-windows-smoke` untouched — still outstanding, and not clearable from this macOS/Metal host.

Note for whoever picks up the re-review: the two commits are unchanged in content from the approved pair, only re-parented — `918e796c` / `54b45527` are the rebased `66f59342` / `0fc9bf48`.

— worker

jakildev (commented):
## Re-review — render: interior-edge margin yield covers the slot-tiebreak gap — kill the fractional-offset shared-edge fringe (#2432)

**Verdict:** approve

Picked up via `fleet:changes-made` + `human:re-review`. No longer stacked — its
former upstream (#2429) merged (`7cd6b51d`) and the merger's rebase onto
master hit a conflict from replaying #2429's already-merged commits; the
author resolved it by dropping the stale inherited prefix rather than
resolving file-by-file (comment at 23:37:27Z). Current base is `master` at
its live tip (`01634a50`), which already includes #2429 and two more shader
PRs (#2430, #2431 — the latter touching the same `ir_iso_common.glsl` /
`.metal` files as this PR).

### Prior-review resolution
- ✅ Sonnet's escalation (composite depth-key/tiebreak invariant) — resolved by Opus's 22:33:31Z recheck, which confirmed the key math sound.
- ✅ Opus nit 1 (`kScatterMarginInteriorBiasKey` "well below" wording vs. actually-on-ceiling) — fixed, addressed in the amend that Opus approved at 23:18:00Z.
- ✅ Opus nit 2 (`kScatterMarginYieldGradScale`'s second, undocumented requirement) — fixed, same amend, same approval.

Nothing carries forward as open — the last verdict before this rebase was
Opus's approve.

### What changed since the last verdict
Nothing semantically. I diffed the current head (`01634a50..HEAD`, i.e.
`54b45527` + `918e796c`) file-by-file against master and every hunk matches
what the prior Sonnet/Opus passes already described verbatim: the
`kScatterMarginInteriorBiasKey = 8` / `kScatterTieLowBitsSpan` bracket and its
three asserts in `ir_render_types.hpp`, the mirrored constants and comments in
`ir_iso_common.glsl`/`.metal`, the `marginYieldGradFloor` /
`marginInteriorYieldBias` vertex-stage plumbing and the interior-penetration
branch in `f_peraxis_scatter`/`peraxis_scatter.metal`'s fragment stage, the
`PER_AXIS_MARGIN` overlay enum + CLI wiring in `perf_grid/main.cpp`. This is
the same logical diff Opus signed off on at the amend, replayed onto a newer
master — not new work.

### Verification performed
- Confirmed the rebase didn't silently drop or revert #2431's shared-edge
  work in the two files both PRs touch (`ir_iso_common.glsl`/`.metal`): the
  diff against master applies cleanly as pure addition, no removed hunks from
  #2431's content.
- Confirmed `baseRefName` is genuinely `master` now (not a stale stacked
  ref) and `mergeable: MERGEABLE`.
- Re-read the full amend content described in the 23:18:00Z Opus pass
  (identity `kScatterMarginInteriorBiasKey == kDepthEncodeShift`, the squared
  `9 >= 8` bound for `kScatterMarginYieldGradScale`) against the live files —
  present, unchanged.

### Nits
None.

- [ ] Linux/OpenGL smoke — already tagged `fleet:needs-windows-smoke`; unaffected by the rebase (GL keeps pre-#2428 behavior, fix is Metal-only pending the #1938 port).

Escalation: please Opus-recheck before merge (touches: engine/render/ shaders — composite depth-key/tiebreak invariant), though this should be a quick confirmation pass given the content is verified unchanged from the diff Opus already fully approved — only the base SHA moved via a mechanical rebase.

jakildev (commented):
## Opus recheck (post-rebase re-affirm) — interior-edge margin yield (#2432 @ `918e796c`)

**Verdict:** approve

Re-affirming my prior approval on the new SHA after the rebase, per the
"approval is stale after a cascade-rebase" rule — I did not take the
unchanged-content claim on faith, and I did not re-litigate the fix itself
(fully reviewed and approved at `22:33Z` / `23:18Z`).

Sonnet's `00:40:11Z` re-review says the content is verified unchanged and only
the base SHA moved. **I confirm that mechanically**, and I chased the one class
it could not: whether a *textually* clean rebase is *semantically* clean against
the three PRs that landed underneath it.

### The rebase is clean, and the PR's own delta is byte-identical

- Head `918e796c` is exactly **two commits on the live master tip** (`01634a50`);
  `merge-base == origin/master`. The stale inherited #2429 prefix the author
  dropped is genuinely gone — no replayed already-merged commits.
- I diffed the approved delta (`bc7e192d^^..bc7e192d`) against the rebased one
  (`01634a50..918e796c`) and compared **only the added/removed lines**: 292 vs
  292, **identical**. The 20 lines that differ between the two patches are all
  *context* lines — master's own rewrite of the tie-band constants showing
  through the window. Nothing this PR contributes changed.

### The escalation I actually spent this pass on: my earlier verification was against a SHA that did not merge

My `23:18Z` note said the key math was re-verified against #2429's head
`791d0d50` and that the caveat was discharged because "the layout did not
change." **That premise needed rechecking**: #2429 merged as `7cd6b51d`, which
is *not* identical to `791d0d50` — the merge carried post-review changes that
are invisible from this PR's side. The drift is real but **value-preserving**:

- `kScatterCellTieBand`: literal `16.0/8388608.0` → derived `16.0 * kScatterCellTieStep`.
  Both are exactly `2^-19` (step is exactly `2^-23`; the product is an exact
  power of two), so bit-identical, as the new comment claims.
- Overflow-lane bias: `32.0 * kScatterCellTieStep` → `2.0 * kScatterCellTieBand`.
  `2 × (16 × step) ≡ 32 × step`, exact at every step — same value.
- `f_peraxis_scatter.glsl` gained `#include "ir_iso_common.glsl"` and **dropped a
  locally-duplicated `kScatterCellTieBand`**.

That last one is the only structural change, and it is the one a "applies
cleanly as pure addition" check would miss — this PR also edits
`f_peraxis_scatter.glsl`, so a re-introduced local constant would be a
duplicate-definition break. **Checked the rebased result:** exactly one
`#include` at line 17, zero local `kScatter*` definitions. Composes cleanly.
This PR's own edit there is a hoist of the `inMargin` declaration above its new
`vDepthColorMode == -1` overlay branch; `vQuadParam` is an input varying, so the
earlier declaration is in scope and the later uses are unaffected.

**Why the drift cannot reach this PR's key math:** the #2428 asserts bracket the
bias against `kDepthEncodeShift` and nothing else —
`kScatterTieLowBitsSpan = kDepthEncodeShift - 1 = 7 < 8 ≤ kDepthEncodeShift = 8`,
and `kScatterMarginYieldGradScale² = 9 ≥ 8`. None of them reference
`kScatterCellTieBandSteps` or the band. `kDepthEncodeShift` is **8 on master and
8 at this head** — untouched by #2429/#2430/#2431. So the constants that moved
are disjoint from the constants this PR's correctness rests on. Sonnet's
original "re-verify if #2429's tie-key layout changes" caveat is discharged for
real this time, on the merged form rather than the pre-merge head.

### No semantic interaction with #2431, despite the overlapping file and theme

#2431 (`10867cb4`) is also a shared-edge seam fix and also edits
`ir_iso_common.glsl`/`.metal`, which is the pairing worth being suspicious of —
two independent corrections to the same symptom can double-correct. **They are
disjoint in practice:** #2431's substance is in the `c_voxel_to_trixel` stage 1/2
shaders, a different pipeline stage; its `ir_iso_common` additions contribute
**nothing** to the scatter tie-key path (grepped its diff for
scatter/tie/band/margin/depth-encode — the only hit is a comment about cardinal
in-plane polarities). The two fixes address different stages of the same visual
class, and neither reads the other's constants.

### CPU↔shader mirrors still agree at this head

`kScatterMarginInteriorBiasKey`: CPU `8` / Metal `8.0`.
`kScatterMarginYieldGradScale`: CPU `3` / Metal `3.0`. The GLSL twin defines
neither and keeps pre-#2428 behavior — consistent with the Metal-lead design
deferred to the #1938 port, as reviewed.

### Nits

- None. Both nits from the `22:33Z` pass were fixed at their definition sites
  and re-approved at `23:18Z`; verified still present and unchanged here.

### Housekeeping

The review at `23:11:39Z` on this PR remains a stale-body misroute — its content
is the Opus recheck for **PR #2357**, posted here by a prior iteration that
reused a `.review-body.md`. Already flagged at `23:18Z`; repeating it so the
merger does not read it as review history for #2432. This pass wrote a
PR-specific body file.

### Test plan (unchanged)

- [ ] Linux/OpenGL smoke — `fleet:needs-windows-smoke` already tagged. GL keeps
      pre-#2428 behavior (fix is Metal-only pending the #1938 port), so this is a
      no-regression check, not a validation of the fix. The rebase adds no
      runtime surface.

PR #2433, opened 2026-07-15
fleet: fleet-net-doctor + github ssh multiplexing — ephemeral-port exhaustion defenses

jakildev:
## Summary
- New `scripts/fleet/fleet-net-doctor`: TCP socket / ephemeral-port pressure report with a reaper-drain probe, hung git/ssh-to-github/gh process scan (`--kill-hung` to reap), and OK/WARN/CRITICAL exit codes (0/1/2) for use in fleet loops. Darwin (`netstat`) and Linux (`ss`) censuses normalized to one format.
- `install.sh`: registry wiring for the new tool (var pair + chmod + symlink), plus a new host-protection stanza (a2) that appends a marked `ControlMaster auto` / `ControlPersist 10m` block for `Host github.com` — one persistent ssh master drops the fleet's per-git-op TCP dial churn to ~zero. Marker- and `ssh -G`-guarded so user-owned multiplexing is left alone; composes per-parameter with the existing keepalive stanza (a), whose keepalives bound a black-holed master to ~60s.
- `fleet-help`: index entry + `--help` passthrough arm.
- `scripts/fleet/CLAUDE.md`: documents the escalated form of #2362 — host-wide `EADDRNOTAVAIL` is ephemeral-port exhaustion, not GitHub downtime; run `fleet-net-doctor`.
- Hermetic test suite `tests/test_fleet_net_doctor.sh` (21 assertions).

## Why
Incident 2026-07-15 (macOS fleet host, 85-day uptime): the kernel PCB reaper wedged, ~30k sockets froze in TIME_WAIT/FIN_WAIT_1/LAST_ACK, and all 16,384 ephemeral ports were consumed — every new outbound connection (git, gh, ssh) failed instantly with `EADDRNOTAVAIL` until a reboot. The feeder was the fleet's own connection churn (a fresh TCP dial per git-over-ssh op and per gh call); the recurring GitHub "black-hole" hangs (#2362) were early symptoms of the same progressive exhaustion. This PR reduces the churn at the source and makes the pressure visible long before the cliff.

## Test plan
- [x] `scripts/fleet/tests/test_fleet_net_doctor.sh` — 21/21 pass (netstat/ss/ps/sleep PATH-shimmed; covers OK / WARN-draining / CRITICAL-frozen / CRITICAL-ports verdicts, hung-process scan incl. mid-line `gh issue` match + zero-padded etime, Linux `ss` state normalization)
- [x] `bash -n` on `fleet-net-doctor` + `install.sh`
- [x] Live validation on the incident host: correctly reported CRITICAL (16,383/16,384 ports, frozen reaper, 15 hung processes) pre-reboot and OK (196/16,384) post-reboot
- [x] `ssh -G github.com` resolves `controlmaster auto` / `controlpersist 600` with the appended block
- [ ] Reviewer: sanity-check the install.sh (a2) guard ordering (marker → `ssh -G` probe → append)

## Notes for reviewer
gh CLI churn (HTTPS, one connection per invocation) cannot be multiplexed — the doctor's WARN threshold is the guard for that side. Nothing here can unwedge a stuck kernel reaper; exit 2 means reboot the host.

### DIFF
--- a/scripts/fleet/CLAUDE.md
+++ b/scripts/fleet/CLAUDE.md
@@ -68,3 +68,7 @@ applies here too — see `docs/agents/CLAUDE-BASELINE.md` §Style.
   fleet-net.sh` — it shadows `git()`/`gh()` with a `timeout` and bounds every
   current and future call site by construction — rather than adding per-site
   guards. Python fetchers use their own subprocess/urllib timeouts instead.
+  The escalated form of the same failure is host-wide: leaked/hung
+  connections exhaust the ephemeral port range and every network call dies
+  instantly with EADDRNOTAVAIL ("Can't assign requested address") — that is
+  not GitHub being down; run `fleet-net-doctor` (exit 2 ⇒ reboot the host).
--- a/scripts/fleet/fleet-help
+++ b/scripts/fleet/fleet-help
@@ -116,7 +116,7 @@ Example:
 See scripts/fleet/fleet-edit header.
 EOF
             ;;
-        fleet-run | fleet-run-targets | fleet-down | fleet-claim | fleet-labels | fleet-gate-status)
+        fleet-run | fleet-run-targets | fleet-down | fleet-claim | fleet-labels | fleet-gate-status | fleet-net-doctor)
             bash "$path" --help
             ;;
         fleet-claude-stream | fleet-state-scout | fleet-feedback)
@@ -237,6 +237,9 @@ Parallel-agent fleet (tmux + Claude)
                                  the LLM merger pass
   fleet-gate-status [--json]     show usage gate (Anthropic + GitHub API
                                  pools) + per-pane cooldown state
+  fleet-net-doctor [--kill-hung] TCP socket / ephemeral-port pressure report
+                                 (EADDRNOTAVAIL forensics) + hung git/ssh/gh
+                                 reaper; exit 0 OK / 1 WARN / 2 CRITICAL
   fleet-claude-stream            format claude stream-json for tmux
   fleet-heartbeat <agent>        bump heartbeat mtime
   fleet-iteration-summary <agent> "<text>"
--- a/scripts/fleet/fleet-net-doctor
+++ b/scripts/fleet/fleet-net-doctor
@@ -42,7 +42,7 @@ HOST_OS="${FLEET_NET_DOCTOR_OS:-$(uname -s)}"
 kill_hung=0
 case "${1:-}" in
     --kill-hung) kill_hung=1 ;;
-    -h|--help) sed -n '2,36p' "$0" | sed 's/^# \{0,1\}//'; exit 0 ;;
+    -h|--help) sed -n '2,32p' "$0" | sed 's/^# \{0,1\}//'; exit 0 ;;
     "") ;;
     *) echo "fleet-net-doctor: unknown option: $1 (try --help)" >&2; exit 64 ;;
 esac
@@ -132,7 +132,7 @@ while IFS= read -r line; do
     # zero-padded field ("08") is otherwise parsed as (invalid) octal.
     case "$etime" in
         *-*)   mins=99999 ;;
-        *:*:*) mins=$(( 10#${etime%%:*} * 60 )) ;;
+        *:*:*) IFS=: read -r hh mm _ <<<"$etime"; mins=$(( 10#$hh * 60 + 10#$mm )) ;;
         *)     mins=$(( 10#${etime%%:*} )) ;;
     esac
     if (( mins >= HUNG_MINUTES )); then
--- a/scripts/fleet/fleet-net-doctor
+++ b/scripts/fleet/fleet-net-doctor
@@ -1,0 +1,167 @@
+#!/usr/bin/env bash
+# fleet-net-doctor — diagnose TCP socket / ephemeral-port pressure on a fleet
+# host, and reap hung git/ssh/gh processes.
+#
+# Why this exists (incident 2026-07-15, macOS host): heavy fleet connection
+# churn (gh CLI + git-over-ssh to GitHub, API traffic) combined with a wedged
+# kernel PCB reaper (85-day uptime) left ~30k sockets frozen in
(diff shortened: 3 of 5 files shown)


jakildev (commented):
## Review — fleet: fleet-net-doctor + github ssh multiplexing — ephemeral-port exhaustion defenses

**Verdict:** needs-fix

### Needs-fix
- `scripts/fleet/fleet-net-doctor:45` — `--help`'s `sed -n '2,36p' "$0"` grabs 4 lines past the end of the header-comment block (comment ends at line 32; 33 is blank, 34 is `set -u`, 35 is blank, 36 is `HUNG_MINUTES="${HUNG_MINUTES:-30}"`). Since the trailing `sed 's/^# \{0,1\}//'` only strips lines that start with `#`, those non-comment lines pass through verbatim and get appended to the printed help text. Verified by running the reconstructed script: `fleet-net-doctor --help` prints a trailing blank line, `set -u`, another blank line, then `HUNG_MINUTES="${HUNG_MINUTES:-30}"` after the intended usage/env docs. Fix: change the range to `2,32p` (or derive it from the first non-`#` line so it doesn't silently drift again if the header grows).
- `scripts/fleet/fleet-net-doctor:135` — the `HH:MM:SS` branch of the etime parser drops the minutes field: `mins=$(( 10#${etime%%:*} * 60 ))` only keeps the hours component (`${etime%%:*}` strips everything from the first `:` onward, i.e. `01:45:00` → `01`), so a process elapsed `01:45:00` (105 real minutes) computes `mins=60`, not 105. Verified empirically: with `HUNG_MINUTES=90`, a process at `01:45:00` is *not* flagged even though 105 ≥ 90 — a false negative on the tool's own reap/report criterion. This is silent at the shipped default (`HUNG_MINUTES=30` always saturates once hours ≥ 1, so T5 in the test suite doesn't catch it — it never raises `HUNG_MINUTES` past an hour), but it defeats the configurability the header explicitly documents (`HUNG_MINUTES <age threshold>`). Fix: extract both fields, e.g. `IFS=: read -r hh mm _ <<<"$etime"; mins=$(( 10#$hh * 60 + 10#$mm ))`.

### Praise
- `install.sh`'s new (a2) ssh-multiplexing stanza is careful about composing with the existing (a) keepalive stanza rather than skipping when *any* `Host github.com` block exists (which stanza (a) itself does) — appending its own separately-marked block lets ssh's first-match-wins-per-parameter semantics pick up `ControlMaster`/`ControlPath`/`ControlPersist` from the new block while `ServerAliveInterval`/`ConnectTimeout`/etc. still resolve from (a)'s earlier block. Traced through both stanzas and confirmed this is intentional, not an oversight (and matches the PR body's own claim).
- Hermetic test harness (PATH-shimmed `netstat`/`ss`/`ps`/`sleep`/`sysctl`, invocation-counting shim for the two-census reaper-drain probe) is a solid pattern for testing a script that would otherwise require real socket-table state.
- The `gh` hung-process regex deliberately matches `"<sep>gh <subcommand> "` (not `^gh`) specifically to catch `ps`'s mid-line command column — good catch on a real false-negative mode, and it's the one thing T5 explicitly regression-tests.

- [ ] Add a `--help`/`-h` assertion to `test_fleet_net_doctor.sh` (e.g. `assert_absent` on `set -u` / `HUNG_MINUTES=` appearing in the help output) so the sed-range regression can't reappear silently.
- [ ] Add a case to T5 with `HUNG_MINUTES` raised above 60 and an `HH:MM:SS` etime that should cross the threshold, to catch the dropped-minutes regression.

jakildev:
Addressed feedback:
- `fleet-net-doctor:45` — fixed `--help`'s sed range from `2,36p` to `2,32p` so it no longer prints `set -u` / the `HUNG_MINUTES=` assignment past the header comment block.
- `fleet-net-doctor:135` — fixed the `HH:MM:SS` etime branch to parse both hours and minutes (`IFS=: read -r hh mm _`) instead of dropping the minutes field, so `HUNG_MINUTES` thresholds above 60 work correctly.
- Added regression tests T7 (--help output excludes trailing code lines) and T8 (HUNG_MINUTES=90 correctly flags a 01:45:00 process) to `test_fleet_net_doctor.sh`. Full suite: 26/26 passing.

— worker

jakildev (commented):
## Review — fleet: fleet-net-doctor + github ssh multiplexing — ephemeral-port exhaustion defenses (#2433)

**Verdict:** approve

Re-review picked up via `fleet:changes-made` after commit `0d522fd0` addressed both `needs-fix` items from the prior pass.

### Prior-review resolution
- ✅ `scripts/fleet/fleet-net-doctor:45` — `--help`'s sed range (`2,36p`) printed 4 trailing code lines (`set -u`, `HUNG_MINUTES=...`) past the header comment block — verified fixed at `0d522fd0`. Range is now `2,32p`; confirmed by counting the file directly: line 32 is the last comment line (`FLEET_NET_DOCTOR_PORT_RANGE ...`), line 33 is blank, line 34 is `set -u`.
- ✅ `scripts/fleet/fleet-net-doctor:135` — the `HH:MM:SS` etime branch dropped the minutes field, undercounting elapsed time by up to 59 minutes — verified fixed at `0d522fd0`. Now `IFS=: read -r hh mm _ <<<"$etime"; mins=$(( 10#$hh * 60 + 10#$mm ))`, correctly parsing both fields with the same base-10-forcing pattern the surrounding code already used for the zero-padding gotcha.

### Needs-fix
None.

### Nits
None.

### Praise
- Both regression tests (T7, T8) are exactly the ones requested in the prior review's test-plan checklist, and each pins the specific failure mode: T7 asserts `set -u` / `HUNG_MINUTES="` are absent from `--help` output; T8 exercises the precise `01:45:00` @ `HUNG_MINUTES=90` case that previously produced a false negative (105 real minutes silently computed as 60).
- Verified independently rather than trusting the reported count: copied the fixed script + test file + `lib_assert.sh` into an isolated `/tmp` dir (not the shared `net-doctor` worktree, which is a mutating clone) and ran the suite directly — 26/26 pass, matching the author's claim.

- [x] Already covered — 26/26 local run (independently reproduced) confirms both fixes and no regressions elsewhere in the suite.

PR #1205, opened 2026-05-26
T-368: render: async texture loading API + World icon load POC

jakildev:
## Summary

- Add `IRRender::loadImageAsync` / `IRRender::uploadDecodedImage` — async texture decode-then-upload pattern respecting GL/Metal context affinity (decode on worker thread, upload on main thread)
- Migrate the window icon load in `World::World` from synchronous `ImageData` to the new async API as proof-of-concept
- File startup-time measurement in `docs/perf-reports/asset_async_startup.md`

**Task:** T-368 (Phase 5 of epic #226)
**Issue:** #1073

## Approach

`loadImageAsync` wraps `std::async(std::launch::async, ...)` around the existing `ImageData` decode. Returns an `AsyncTextureHandle` (move-only `std::future<DecodedImage>` wrapper) with `valid()`, `isReady()`, and `take()` methods. `uploadDecodedImage` creates a `Texture2D` and uploads pixel data on the main thread.

The POC fires the async decode before the render-config block in `World::World` and waits after — the overlap window (~0.03 ms of config setters) is negligible for a single 4 ms icon decode, but the pattern validates for future multi-texture startup loads and lazy gameplay loads.

Current dispatch uses `std::async` — a future iteration can migrate to `IRJob::pinToAsync` when non-blocking pinned dispatch lands.

## Acceptance criteria

- [x] `IRRender::loadImageAsync` exists, returns `AsyncTextureHandle`, resolves to valid decoded pixels
- [x] `IRRender::uploadDecodedImage` creates a GPU texture from decoded data on main thread
- [x] One existing blocking load migrated (window icon in `World::World`)
- [x] Startup-time measurement filed in `docs/perf-reports/`
- [x] `fleet-build --target IRShapeDebug` clean on macOS Metal
- [x] `fleet-run IRShapeDebug --auto-screenshot 10` — 7/7 shots captured, clean exit

## Test plan

- [x] macOS Metal: build + run IRShapeDebug with auto-screenshot — clean
- [ ] Linux OpenGL: `fleet-build --target IRShapeDebug` — expected clean (no platform-specific code)
- [ ] `perf_grid_matrix.sh` — no regression (new code only runs at startup)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1181, opened 2026-05-25
T-362: system: pipeline groups + cross-system access validation (T-224, Phase 3)

jakildev:
## Summary

Phase 3 of multithreading epic #226. Adds `registerPipelineGroups()` API + cross-system access-conflict validator + IRJobs-based group dispatch + unit tests.

Closes #1071.

**Status:** WIP (claim commit only)

## Scope per #1103

- `registerPipelineGroups(IRTime::UPDATE, { {a, b}, {c} })` API on `IRSystem`
- Per-pipeline storage: `std::list<SystemId>` → `std::vector<std::vector<SystemId>>`
- Validator (FATAL at `World::start`) on:
  - intersecting writes (writes ∩ writes / writes ∩ reads either direction)
  - any `MAIN_THREAD` system in a >1 group
  - two `mutates_archetype_graph` systems in same group
- `executePipeline` dispatches groups concurrently via IRJobs, fencing on `flushStructuralChanges` between groups
- Unit tests for each validator rejection rule

**Deferred to #1103:** per-demo pipeline migration + perf_grid_matrix benchmark (open-ended, gated on per-system `const` opt-in audit per #1103).

## Test plan

- [ ] `fleet-build --target IrredenEngineTest IRShapeDebug` clean
- [ ] `fleet-run IrredenEngineTest` — new validator-rejection tests pass; full suite green
- [ ] `fleet-run IRShapeDebug --auto-screenshot 10` — visual parity (no demo uses groups yet, so this is regression-only)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2435, opened 2026-07-15
docs: epic-steward — engine #226 plan sync + steward ledger (heal + close-out audit)

### LINKED ISSUE #1073: asset: async asset loading via pinned worker (T-226, Phase 5 of #226, optional follow-on)
**Model:** opus
**Part of epic:** #226
**Blocked by:** #1068

## Scope

Phase 5 of the multithreading epic. **Not on the closing path** for #226 — that umbrella closes when T-220 through T-225 land. This ticket is an independent follow-on that uses the infrastructure to remove main-thread blocking on disk reads.

## Approach

- Pin worker `N-1` for I/O via `IRJobs::pinTo(N-1, ...)`. Keeps disk I/O off the shared compute workers so a large texture read doesn't stall a parallel-for.
- `IRAsset::loadTextureAsync(path) -> AssetHandle<C_Texture>` — returns immediately with a handle; the load completes on the pinned worker; the renderer waits on the handle at first use (or per-frame polled completion, TBD by implementor).
- Convert one existing main-thread blocking load (a startup texture) to async as proof of concept; measure startup-time delta.

## Acceptance criteria

- `IRAsset::loadTextureAsync` exists, returns an `AssetHandle<C_Texture>`, and produces a valid texture once the handle resolves.
- One existing blocking load is migrated and a startup-time measurement is filed in `docs/perf-reports/`.
- No regression on `perf_grid_matrix.sh`.

## Dependencies

- **Blocked by #1068** (`IRJobs::pinTo` exists).

## Gotchas

- GL/Metal texture creation must happen on the main thread (context affinity). The pinned worker only does the disk-read + decode; texture upload schedules onto main.
- Don't try to make the whole `IRAsset` API async in this ticket. One entry point, one POC. Other loaders follow if the pattern is right.

## References

- Reserved-threads design: #226 \"Layer 5\" section.

jakildev:
## Summary

Docs-only steward iteration. Engine **#226** (ECS multithreading umbrella) had gone **~12 weeks without a single steward trigger** — the projection skips epics with an empty `## Children` checklist *and* no repo-side plan file, and with every child already closed the adopt scan (open issues only) had no surface either. A manual `Part of epic:` back-ref sweep surfaced it.

This PR carries the durable record for that iteration:

- **Syncs the architect plan into the repo.** It only ever existed in device-local staging (`~/.fleet/plans/issue-226.md`, authored 2026-05-22) — the epic's plan of record lived on exactly one machine. Synced **verbatim** as a historical artifact; plans are append-only, so it is not rewritten.
- **Appends the `## Steward ledger`** the protocol wants in the umbrella's own plan file (D1-D5 + Events).

Urgent signals already landed on the issue, not gated on this PR: the healed 9-child checklist, the [close-out audit comment](https://github.com/jakildev/IrredenEngine/issues/226#issuecomment-4985497637), and the `fleet:needs-human` park.

## What the audit found

All nine children are **genuinely shipped** — verified as merged PR **and** deliverable present on `master`, because `CLOSED/COMPLETED` misleads on two rows:

- **#1071** (Phase 3) links **PR #1181, closed unmerged**. The scope had already landed via **PR #1104** (`5f5df6d7`); #1181 was a later duplicate, correctly abandoned. `registerPipelineGroups` is on `master`.
- **#1073** (Phase 5) links **no PR at all** — it shipped via **PR #1205**, which omits a `Closes #1073` keyword.

The umbrella **stays open**. Acceptance criterion 2 — **≥2× UPDATE-pipeline throughput** — was never measured: `threading_phase3.md` calls its delta *"in the noise"*, `threading_propagate_transform.md` is titled *"Why not the 2× target on this workload"*, and that report **deferred** the authoritative run until #1100 was online. #1100 closed 2026-05-24 — the measurement has been unblocked ~7 weeks and simply never ran. `perf-gate.yml` has no threading cell either. Accepting vs measuring is the human's call, not the steward's, so the epic is parked rather than closed or force-fitted.

## Test plan

- [x] All 13 file paths cited in the ledger resolve on `origin/master`
- [x] All 4 `file:line` citations verified exact (`perf_grid/main.cpp:758`, `world.cpp:32`, `perf_grid_matrix.sh:104-106`, `engine/system/CLAUDE.md:399`)
- [x] Ledger conforms to the `epic-steward-protocol.md` schema
- [x] Healed checklist landed on #226 (9/9 ticked); heal-on-first-claim asserted no pre-existing checklist
- [x] Docs-only — no code touched, no build impact

No `Closes #` line: **#226 must stay open** pending the human's criterion-2 call.

### DIFF
--- a/.fleet/plans/issue-226.md
+++ b/.fleet/plans/issue-226.md
@@ -1,0 +1,506 @@
+# Plan: ECS multithreading epic execution (engine #226)
+
+## Context
+
+Engine **#226** is the umbrella design for CPU multithreading of ECS
+system execution. The issue body is already a comprehensive design —
+this plan is about **how the epic ships**: phase ordering, ticket
+filing, migration cascades through existing C++ systems, and integration
+with the just-merged Lua-driven ECS stack (#293, complete 2026-05-12).
+
+Survey of current state, with deltas from the issue's assumptions:
+
+- **Lua-driven ECS is done.** CODEGEN-default + EVAL-LuaJIT both
+  production (PRs #595/596/597/598/599, all merged 2026-05-10–12).
+  This adds a Layer that wasn't in the original #226 design:
+  codegen-emitted systems need to pass `Concurrency` through from
+  Lua spec to the emitted `createSystem<...>` call, and EVAL-mode
+  systems must be forced to `MAIN_THREAD` regardless of the spec
+  (sol2 isn't thread-safe).
+- **The cliff is already documented.** Issue **#1052** (filed
+  2026-05-22) records `IRPerfGrid` at 262K entities → ~20ms/frame
+  UPDATE, blocking the 100fps target. Phase 0 of #226 ("build a
+  benchmark to prove the cliff") is largely redundant — the cliff
+  is `perf_grid` itself, and the existing `perf_grid_matrix.sh`
+  harness drives it.
+- **Perf-gate CI exists.** `.github/workflows/perf-gate.yml` runs
+  `perf_grid_matrix.sh`, compares to baseline in
+  `docs/perf/baseline_latest/`, fails on >10% regression. Phase 2
+  acceptance ("≥2× UPDATE throughput") just needs `worker_threads`
+  added as a matrix axis, not a new harness.
+- **`registerSystem<SystemName, Components...>` member-on-System
+  pattern already landed** (engine #580). Both threading
+  (`Concurrency` annotation) and codegen-emitted systems should
+  target this shape, not the older explicit-Params boilerplate.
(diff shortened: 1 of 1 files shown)


jakildev (commented):

**Verdict:** approve

introduces nor weakens.

lost reservation.

### 5. GLSL ↔ Metal parity

needs-fix.

jakildev:
Auto-merged by tier-0 fleet-rebase: diff touches only `.fleet/plans/` and the PR carried `fleet:approved` with no disqualifying labels. — fleet merger

PR #2448, opened 2026-07-19
fleet: derive PR closes-refs at fetch so 304-reuse keeps stack offers

### LINKED ISSUE #2442: fleet/scout: 304-reuse serves body-stripped PRs — the Closes-#N stack-base match is dead on most ticks (silently withholds stackable offers)
**Area:** scripts/fleet (fleet-state-scout)
**Model:** opus
**Blocked by:** (none)

## What happened

Verified 2026-07-15 (macOS opus pane) while diagnosing why game #256 — the one plausible opus unlock — is never offered as a stackable pick.

`enrich_stackable_blocker_prs` ties a blocker PR to its issue by a **union**: branch prefix (`claude/<N>-`, `claude/game-<N>-`) **OR** the PR body matching `Closes/Fixes/Resolves #<N>`. The body half is **dead on every tick that reuses the cached PR list**:

1. `fleet-state-scout:2824-2826` pops `body` off every PR *after* enrichment runs, so bodies never reach `state.json` (comment: *"Drop PR bodies now the only consumer (the Closes-#N stack-base match above) has run"*). Correct **within** a tick.
2. `collect_state` builds `_prev_repos` from the previous **state.json** (`:2592`) — i.e. from those body-stripped records.
3. `fetch_prs(repo, prev)` (`:290-295`, added by #2219) short-circuits on a conditional-REST **304 (open-PR set unchanged)** and returns `prev` verbatim — PRs with **no `body` key**.
4. Enrichment then runs against body-less PRs, so `closes_re.search(pr.get("body","") or "")` can never match. Only the branch-prefix half survives.

Each change is correct alone; they landed separately — `361ce322` (#2219, 304-reuse) and `a9cc67e9` (#2220, body-pop). The defect is their **interaction across ticks**. It self-heals only on a tick where the open-PR set changes (ETag flip → GraphQL refetch with bodies), which is why it presents as intermittent queue starvation rather than a hard failure.

**Live victim.** game #256 (`opus`, `owner: free`) resolves to a single live blocker #255 (#254 is CLOSED and `resolve_blocked_by` already collapses it — the projection shows `blocked_by: "#255"`). #255's PR **#263** is `fleet:approved` + `MERGEABLE`, but its branch is `claude/game-worker-3-issue-255`, matching neither accepted prefix (that branch shape is #2419). Its body **does** carry `Closes #255` — precisely the case the union exists for — so #256 should be offered a stackable base. It is not. The #253 AI epic chain (#256 → #257 → #258/#259) stays frozen, and this pane has taken consecutive no-task exits as a direct result.

## Repro (observed output)

Running the **real** `enrich_stackable_blocker_prs` against the live game PR list **with** bodies present (as a fresh GraphQL fetch serves them):

```
_single_blocker_issue("#255") -> 255
  MATCH pr#263 branch=False closes=True head=claude/game-worker-3-issue-255
unsafe_base_reason -> None
RESULT stackable_blocker_pr -> {"number": 263, "headRefName": "claude/game-worker-3-issue-255", "author": "jakildev"}
```

The same function against the shipped `state.json` records (body-stripped, exactly what the 304 path serves as `prev`) yields **no offer**. Live `state.json` confirms `repos.game.prs[]` records carry keys `[author, baseRefName, headRefName, isDraft, labels, mergeable, number, reviews, title, updatedAt]` — **no `body`** — and `repos.game.tasks.open[]` `#256` has no `stackable_blocker_pr`.

## Why the suite is green

`tests/test_enrich_stackable_blocker_prs.py::test_closes_n_body_match_offered` builds its PRs in-memory **with** `body=` set, so it only ever exercises the fresh-fetch path. Nothing covers *"prev came from a body-stripped state.json"* — the 304-reuse path has no enrichment coverage at all.

## Ruled out

- **Not the matcher gap (#2419).** That widens accepted *branch* shapes; this kills the *body* half of the union regardless of branch shape. #2419's fix would **mask** this for #263 while leaving every Closes-only base silently broken.
- **Not `unsafe_base_reason`** — returns `None` for #263 (`fleet:approved`, 15 changed files cached, not an empty claim-commit skeleton).
- **Not filter (a)** — #256 has `area: null`, so no area-overlap suppression.
- **Not a leader/follower artifact** — enrichment is guarded by `authoritative`, and this host polls authoritatively.

## Acceptance criteria

1. A blocker PR whose branch does **not** match the prefix forms but whose body carries `Closes #<N>` is offered as `stackable_blocker_pr` on a tick that **reuses** the cached PR list (304) — not only on a fresh GraphQL fetch.
2. A regression test drives the 304-reuse path explicitly: `prev` = body-stripped records (exactly as written to `state.json`) → the offer still appears. The test **fails against current master**.
3. With #263 open, `stackable_blocker_pr` appears on game #256 pointing at PR #263 on the next scout tick.
4. `state.json` does not regain full PR bodies (no size regression).

## Suggested shape

Derive the signal **once at fetch time** so it survives both the strip and the 304 reuse — see the `## Plan` comment.

jakildev:
## Summary
- The scout's stackable-blocker enrichment matched a blocker PR to its
  blocked issue by branch prefix **OR** a `Closes/Fixes/Resolves #N`
  reference in the PR body. The body half was dead on every tick that
  reused the cached PR list: bodies are popped after enrichment, and
  `fetch_prs`'s conditional-REST **304** fast path returns the prior
  **body-stripped** records verbatim, so enrichment only ever matched the
  branch-prefix half. Each half (304-reuse, body-pop) is correct alone;
  the defect is their interaction across ticks — intermittent
  stackable-offer starvation.
- **Fix:** derive the Closes-refs once at fetch time
  (`_fetch_prs_graphql`), where the body is still live, into a bounded
  `pr["closes_issues"]` int list that survives both the strip and the 304
  reuse. Enrichment matches on the derived field instead of re-parsing a
  body it may never receive; the grammar is hoisted to one module-level
  `CLOSES_RE` so the offer and the derivation can't drift.
- **Deploy-boundary self-heal:** `fetch_prs` treats a `prev` whose records
  predate the field as cache desync and refetches, so the fix is live on
  the first tick after deploy rather than waiting for an unrelated ETag
  flip. Bodies are still popped — no `state.json` size regression.

## Test plan
- [x] New `test_closes_n_offer_survives_304_reuse` (enrichment offers off a
  body-stripped record) and `TestFetchPrs304Reuse.test_refetches_when_prev_lacks_closes_issues`
  (refetch on missing field) — **both fail against master**, pass after.
- [x] `test_enrich_stackable_blocker_prs.py` green (29 tests); the full
  `scripts/fleet/tests` suite green (648 tests).
- [x] `ruff check scripts/` clean.

## Notes
- Deploying the scout requires a `fleet-down` / `fleet-up` bounce (the
  in-proc-vs-subprocess activation gap) for the change to take effect.
- Follow-up candidate (out of scope here): `fleet-claim` defines a
  byte-identical `CLOSES_RE`; a shared helper module would keep the offer
  and accept sides from drifting.

Closes #2442

### DIFF
--- a/scripts/fleet/fleet-state-scout
+++ b/scripts/fleet/fleet-state-scout
@@ -238,6 +238,19 @@ def _is_pull_request(issue):
     return "pull_request" in issue
 
 
+# Closes/Fixes/Resolves #N refs parsed from a PR body at fetch time — where the
+# body is still live — into pr["closes_issues"], a bounded sorted int list.
+# enrich_stackable_blocker_prs matches its blocker union's body half on this
+# derived field rather than re-parsing pr["body"], which it may never receive:
+# bodies are popped after enrichment and fetch_prs's 304 fast path reuses the
+# body-stripped prior records, so the live parse was dead on every reuse tick
+# (#2442). Deriving once here keeps the offer and this grammar from drifting.
+CLOSES_RE = re.compile(
+    r"\b(?:close[sd]?|fix(?:e[sd])?|resolve[sd]?)\s+#(\d+)\b",
+    re.IGNORECASE,
+)
+
+
 def _fetch_prs_graphql(repo):
     out = run_capture([
         "gh", "pr", "list", "--repo", repo, "--state", "open",
@@ -264,6 +277,12 @@ def _fetch_prs_graphql(repo):
             }
             for rev in pr.get("reviews", [])
         ]
+        # Derive here, while body is live, so the signal survives both the
+        # post-enrich body-pop and fetch_prs's 304 reuse of body-stripped
+        # records (#2442). Its presence is also fetch_prs's cache-desync marker.
+        pr["closes_issues"] = sorted(
+            {int(n) for n in CLOSES_RE.findall(pr.get("body", "") or "")}
+        )
     # gh's PR ordering is not stable run-to-run; sort by number so the
     # projection hash stays quiescent when nothing has actually changed.
     prs.sort(key=lambda pr: pr.get("number", 0))
@@ -290,7 +309,14 @@ def fetch_prs(repo, prev=None):
     changed, _body = conditional_get(repo, "pulls", params={
         "state": "open", "sort": "updated", "direction": "desc", "per_page": "100",
     })
-    if not changed and prev is not None:
+    # Reuse the prior list only when it carries the derived closes_issues field
+    # (added by _fetch_prs_graphql). On the first tick after a deploy that
+    # introduced the field, prev predates it — reusing it would re-strand every
+    # Closes-only stack base until an unrelated ETag flip, re-creating #2442. A
+    # missing key ⇒ cache desync ⇒ fall through to a fresh fetch. all() on an
+    # empty prev is True (an empty open-PR set needs no refetch).
+    if (not changed and prev is not None
+            and all("closes_issues" in p for p in prev)):
         return prev
     return _fetch_prs_graphql(repo)
 
@@ -1548,10 +1574,12 @@ def enrich_stackable_blocker_prs(state):
     # `#NNN` by resolve_blocked_by() (runs first), so it reaches here eligible.
     #
     # A PR "matches" the blocker when its head branch is `claude/<NNN>-…` OR its
-    # body Closes/Fixes/Resolves #NNN — the same union fleet-claim's
+    # derived closes_issues (Closes/Fixes/Resolves #NNN, parsed at fetch time —
+    # see CLOSES_RE) contains NNN — the same union fleet-claim's
     # find-stackable-blockers uses, so the scout offers every base the live
     # finder would (the Closes-#N form catches blocker PRs on non-standard /
-    # human branches).
+    # human branches). The Closes half reads the derived field, not pr["body"],
+    # which is body-stripped on 304 reuse ticks (#2442).
     #
     # The field is always omitted when:
     #   - blocked_by doesn't reference exactly one issue (multi-blocker,
@@ -1576,15 +1604,13 @@ def enrich_stackable_blocker_prs(state):
             issue_num = _single_blocker_issue(blocked_by)
             if not issue_num:
                 continue
-            closes_re = re.compile(
-                r"\b(?:close[sd]?|fix(?:e[sd])?|resolve[sd]?)\s+#"
-                + re.escape(issue_num) + r"\b",
-                re.IGNORECASE,
-            )
+            # _single_blocker_issue returns a digit string; closes_issues holds
+            # ints — compare as int so "255" in [255] doesn't silently miss.
+            issue_num_int = int(issue_num)
             matches = [
                 pr for pr in prs
                 if branch_matches_issue(pr.get("headRefName", ""), issue_num, repo_key)
-                or closes_re.search(pr.get("body", "") or "")
+                or issue_num_int in (pr.get("closes_issues") or [])
             ]
             if len(matches) != 1:
                 continue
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — fleet: derive PR closes-refs at fetch so 304-reuse keeps stack offers

**Verdict:** approve

### Summary

Fixes #2442: the scout's `enrich_stackable_blocker_prs` matched a blocker PR
to its blocked issue via head-branch prefix OR a `Closes/Fixes/Resolves #N`
reference in the PR body — but the body half went dead on any tick that hit
`fetch_prs`'s 304 fast path, because `tick_once` pops `pr["body"]` after
enrichment and the 304 path reuses that same body-stripped `prev` list
verbatim. The fix derives `closes_issues` (a bounded int list) once at fetch
time in `_fetch_prs_graphql`, while the body is still live, and enrichment
now matches against that derived field instead of re-parsing a body it may
never receive. `fetch_prs` also gates its 304-reuse on every `prev` record
carrying the new field, so a `prev` that predates the field (first tick after
deploy) falls through to a fresh GraphQL fetch instead of re-stranding the
signal until an unrelated ETag flip.

### Checklist

- **Correctness of the core fix** — traced the full pipeline: `_fetch_prs_graphql`
  derives `closes_issues` while body is live (`fleet-state-scout:283-285`) →
  `enrich_stackable_blocker_prs` reads the derived field, comparing as `int`
  rather than string so `"007"` vs `"7"` normalization can't silently miss
  (`fleet-state-scout:1607-1613`) → `tick_once` pops `body` only *after*
  enrichment has run, and does not pop `closes_issues`, so it survives into
  `state.json` and is available as `prev` on the next tick
  (`fleet-state-scout:2843-2851`). The chain is consistent end to end.
- **304-reuse self-heal** — `fetch_prs`'s new guard
  (`all("closes_issues" in p for p in prev)`) correctly falls through to a
  live fetch on cache-desync, and `all()` over an empty `prev` is `True` so
  an empty open-PR set doesn't force a needless refetch. `_prev()`'s caller
  returns `None` on first run, which the `prev is not None` check already
  handles.
- **Regression tests verified independently** — extracted the PR-head
  `scripts/fleet/` tree and ran it standalone: `test_enrich_stackable_blocker_prs.py`
  is green (29/29, including the two new tests), the full
  `scripts/fleet/tests` Python suite is green (648 tests, matches the PR
  body's claim exactly), and `ruff check scripts/` is clean. Also confirmed
  the two new tests (`test_closes_n_offer_survives_304_reuse`,
  `TestFetchPrs304Reuse.test_refetches_when_prev_lacks_closes_issues`)
  genuinely fail against `origin/master` (the updated test file's `_pr()`
  helper references `_mod.CLOSES_RE`, which doesn't exist pre-fix) — the
  "fails on master, passes after" claim holds.
- **Deferred follow-up claim checked** — the PR body notes `fleet-claim`
  defines a "byte-identical `CLOSES_RE`" as an out-of-scope dedup candidate;
  grepped `scripts/fleet/fleet-claim:2489` and confirmed the pattern is in
  fact byte-identical to the new `CLOSES_RE` in `fleet-state-scout`. Fine to
  defer per the fix-forward covenant — not a blocking duplication yet.
- **No leftover references** to the removed per-issue `closes_re` local
  variable name anywhere in the file.
- **Fleet-scripts checklist item** (state.json freshness via `generated_at`
  vs mtime) — not touched by this diff, N/A.
- Not applicable: ECS/render/lighting/math/serialization checklist items
  (Python fleet-infra change only).

Nothing needs-fix, no nits.

- [x] Already covered — `scripts/fleet/tests` suite (648 tests) and `ruff
      check scripts/` independently re-verified during this review.

### Escalation

None — this is `scripts/fleet/` infrastructure (Python), not
`engine/render|entity|system|world|audio|video`, non-trivial `engine/math`,
or a public `ir_*.hpp` surface. Logic traced end-to-end and regression
claims independently reproduced.

Opus recheck not required.

Issue #2449, opened 2026-07-19
build(macOS): Apple clang 21 breaks fresh builds — fmt 10.1.1 consteval incompatibility (bump fmt→10.2.1)


jakildev:
**Area:** build (third-party deps) / macOS
**Model:** [opus]
**Host:** macOS (Darwin 25.5.0), Apple clang **21.0.0** (clang-2100.1.1.101), arm64

## Symptom
A **fresh** build on this macOS host fails compiling the bundled `fmt` library
(`_deps/fmt-build/CMakeFiles/fmt.dir/src/format.cc.o`) with a stream of:

```
fmt/format-inl.h:61:19: error: call to consteval function
  'fmt::basic_format_string<...>::basic_format_string<FMT_COMPILE_STRING, 0>'
  is not a constant expression
```

`fmt` is pinned to `GIT_TAG 10.1.1` in `engine/profile/CMakeLists.txt`. Existing
engine `build/` dirs only keep working because their `libfmtd` object was
compiled back in June under an older Apple clang; the object is never
recompiled, so incremental engine builds mask the break. Any dir that must
compile `fmt` from scratch (a new/clean build, or the game's dedicated
`build-game` dir) hits the wall and stops at ~12–20% before any engine/game
code compiles.

## Root cause (verified, minimal repro)
fmt < 10.2 uses `consteval` for its compile-time format-string checks in a form
that newer Clang (17+, and definitively Apple clang 21) rejects at C++20.
Reproduced standalone against the pinned source, no build system involved:

```
# fails:
clang++ -std=c++20 -c _deps/fmt-src/src/format.cc -I_deps/fmt-src/include   -> consteval errors
# compiles clean at C++17 (fmt disables consteval there):
clang++ -std=c++17 -c _deps/fmt-src/src/format.cc -I_deps/fmt-src/include   -> OK
# compiles clean at C++20 with fmt's own escape hatch:
clang++ -std=c++20 -DFMT_USE_CONSTEVAL=0 -c _deps/fmt-src/src/format.cc ...  -> OK
```

The engine builds at C++20, so the consteval path is active and breaks.

## Not a defect on Linux/Windows
gcc-13 (the Linux/WSL fleet env) and mingw compile fmt 10.1.1 fine — this is
Apple-clang-21-specific. So it is a macOS-host build-capability outage, not a
cross-platform regression. It does, however, block **all** fresh macOS builds
(engine and game) and therefore any macOS-side build/render verification.

[...]

PR #2460, opened 2026-07-20
render: fog vision Z cost — XY radius with height-penalized reveal (#2260)

### LINKED ISSUE #2260: render: fog vision Z cost — XY radius with height-penalized reveal
**Area:** engine/render
**Model:** opus
**Blocked by:** (none)

## Context

Playtesting direction from the fog cross-section work (#2246, #2259): the vision disc should be an **XY radius** with **Z as a cost penalty** — matter at the observer's height reveals to the full radius, while matter above/below reveals less (a tall pillar top or a deep pit floor near the rim should read as harder to see than ground at eye level). Today `FrameDataFogObservers` circles are `(x, y, radius, edgeSoftness)` with no observer height; every z along a column reveals identically.

## Acceptance criteria

- `IRPrefab::Fog::setVisionCircle` (and the Lua `IRFog.setVision` surface) accepts an observer z + z-cost factor; zCost = 0 renders byte-identical to today (render-verify gate on existing refs).
- A fog_demo scene with tall + low geometry at equal XY distance shows the height-penalized reveal, gated by committed render-verify refs.
- GLSL/Metal twins mirrored; per-column culls stay conservative (keep by best-case z) so no geometry near the rim is dropped early.

jakildev:
Closes #2260

## What

Adds a per-circle **height penalty** to the analytic fog vision discs so the
effective reveal distance is `dist_xy + zCost * |z - observerZ|`. Matter at the
observer's height reveals to the full radius; a tall pillar top / deep pit floor
at the same XY reveals *less* (iso +Z is the downward height axis). `zCost 0` (the
default for every existing caller) makes the penalty exactly 0 — the plain 2D disc.

New `--edge-zcost` scene (observer at floor level): a central tall pillar at the
observer's XY has its **base revealed and its top faded past the disc**, while a
low cube at the same band stays fully revealed:

fog edge-zcost zoom9

## How

- `FrameDataFogObservers` grows `visionCircleHeights_[8]` = `(observerZ, zCost, 0, 0)`
  **appended after the tail**, so `visionCircles_` / `visionCircleCount_` keep their
  std140 offsets — `c_voxel_to_trixel_stage_2` + `c_voxel_visibility_compact` read
  only those and stay byte-identical with **no** shader edit. The `static_assert`
  grows to `2*N*16 + 16`, machine-enforcing the append-only layout.
- `c_fog_to_trixel`: per-pixel reveal + rim fade fold in the Z penalty.
- `c_voxel_to_trixel_stage_1_body`: the own-column DROP evaluates the reveal at the
  voxel's own world Z (`fogColumnRevealZ` / `fogColumnRevealNearestZ`). The cut-face
  emission + nearest-cell KEEP-ring width stay z-free (best-case-z, conservative keep).
- `ir_voxel_face_select` (the shared face-select include master centralized in
  #2508, rebased onto here): carries the appended `visionCircleHeights` member on
  the shared `FogObserverData` block declaration. Its own face-selection math never
  reads the field — the cut-face test deliberately stays on the **z-free**
  `fogColumnReveal` (conservative keep) — but GLSL admits exactly one declaration
  of a named uniform block and this is now it.
- GLSL + Metal both carry the change (backend parity).

## Deviations

- **Inlined, not `ir_iso_common`.** The Z reveal is inlined in the two fog shaders
  (mirroring the existing inline `fogColumnReveal`) rather than a shared helper —
  per the #1944 note, a new symbol in `ir_iso_common` perturbs every cardinal
  voxel/SDF shader's byte-identical fast path.
- **No Lua binding.** The engine has no `IRFog` Lua surface; the game-repo
  `IRFog.setVision` pass-through is a follow-up note, not part of this PR.

## Verification (macOS / Metal, at `45beb880` — post-#2508 conflict resolution)

- `fleet-build --target IRFogDemo` clean.
- `render-verify --target IRFogDemo` (**full suite, all 15 shots**) → **15/15 PASS**.
  14 shots are byte-identical (`max_d 0`); the one legacy shot `fog_edge_zoom14`
  sits at `max_d 1` / PSNR 110.79 — single-LSB noise floor, not a visual delta.
  New `fog_edge_zcost5/9` macos-debug refs blessed + committed.
- **Existing shots are byte-identical at zCost 0 by construction:** the
  `static_assert` enforces the append-only std140 layout, and every changed shader
  entry point reduces *exactly* to the pre-#2260 expression when `zCost == 0`
  (`0*|Δz| == 0`; even under FMA, `fma(0, Δz, L) == L`). `simplify` clean; the
  fogColumnReveal twins are deliberate documented duplication (#1944), not extractable.
- **GLSL/OpenGL** mirrors Metal line-for-line but needs a Linux host to bless the
  linux-debug refs → `fleet:needs-linux-smoke`.

## Ref drift — resolved upstream, no longer applicable

An earlier revision of this body reported 13 existing fog shots FAILing on
pre-existing master ref drift, tracked separately in **#2473** and deliberately
left out of this PR. That refresh has since merged as `f3c87cc4` (2026-07-20) and
is **already an ancestor of this branch** — it came in with the rebase that resolved
the #2508 conflict. The suite is fully green as reported above; the old FAIL
narrative described a tree this PR no longer ships and has been dropped.


jakildev:
Opus feedback pass (worker-3, macOS): **no actionable design state here, and the code cannot be verified on this host.**

**`fleet:design-unblocked` is spurious.** The PR's label timeline is `fleet:wip` → code committed → `fleet:design-unblocked` — it was **never `fleet:design-blocked`**, there is no `## NEEDS-DESIGN` escalation, and no architect reply to absorb. The design is fully resolved: the plan (first commit) is complete and the render code (second commit) is written to it. The design-unblocked tier keeps routing this PR to opus panes each tick with nothing a design pass can do, so I'm clearing the label to stop the churn. Keeping `fleet:wip` — the PR is genuinely complete-but-unverified.

**Confirmed the #2449 build block still bites.** I attempted an incremental macOS build of `IRFogDemo` from an existing build dir (fmt partially cached), hoping to sidestep the fresh-build wall. It still recompiles `_deps/fmt-build/.../src/format.cc.o` and dies on the Apple clang-21 vs fmt 10.1.1 `consteval` incompatibility (`FMT_STRING`/`FMT_COMPILE_STRING`), 5 errors — exactly #2449. The whole current fleet is macOS, so there is no host that can build/verify this today.

**Owed on a working build host (Linux, or macOS once #2449's `fmt→10.2.1` bump lands):** build `IRFogDemo`, render-verify (existing shots byte-identical at `zCost 0` + the new `--edge-zcost` shots), bless macos/linux refs, render-debug-loop / attach-screenshots, simplify. Not folding the fmt bump into this PR — that's #2449's scope (its own `human:approved` issue, currently `fleet:needs-plan`).

Leaving the WIP PR open to preserve the work; releasing the claim. — worker (opus, macOS)

jakildev:
## worker-3 (opus, macOS) — `fleet:design-unblocked` spurious AND recurring

Same finding as my [09:37Z pass](#): this PR was **never `fleet:design-blocked`**, there is no `## NEEDS-DESIGN` escalation, and **no architect reply to absorb** — so the design-unblocked precondition (a resolved mid-task design escalation) is unmet. Nothing for a design pass to do.

**This has now recurred.** Label timeline: `fleet:design-unblocked` cleared at `09:37:47Z` → **re-added at `10:04:44Z`** (coinciding with a dispatch tick). That looks like automated re-derivation fighting the manual clear — a churn loop that burns an opus pane per tick for zero actionable work.

Re-verified this iteration: **#2449 is still OPEN** (`fmt→10.2.1` not merged), so the all-macOS fleet still cannot build/verify `IRFogDemo`. PR stays complete-but-unverified; owed work (build, render-verify, bless refs, screenshots, simplify) needs a Linux host or #2449 landing.

Clearing `fleet:design-unblocked` again, keeping `fleet:wip`. **If the label recurs, it is an automation bug** (something re-deriving design-unblocked for a PR that was never design-blocked) — flagged to the human via the fleet feedback channel to fix at the source. If there IS design direction I'm missing, please link it and I'll absorb it.

— worker (opus, macOS)

jakildev:
Opus feedback pass (worker-1, macOS): confirming worker-3's 09:37 diagnosis and adding the missing durable follow-up.

`fleet:design-unblocked` here is **R7-generated, not architect-set** — traced the label events: worker-3 removed it at 09:37, `fleet-claim` reconcile R7 (`reconcile_heal_design_unblock`) re-added it at 10:04 because this is a claimless orphaned WIP on a `fleet:queued` issue. There is no `## NEEDS-DESIGN` escalation and no architect reply to absorb. The PR is complete-but-unverified, blocked solely on the #2449 fmt-10.1.1/clang-21 build wall — unbuildable on the all-macOS fleet.

The re-arm loop (R7 fighting a worker's deliberate removal) is now tracked in **#2462** (agent-approved, → needs-plan). Removing the spurious label again to stop the immediate churn; it will re-arm in ~3 drift ticks until #2462's fix lands, at which point future opus panes that get routed here should read this comment and no-op rather than re-investigate.

Owed once a build host exists (Linux, or macOS after #2449's `fmt→10.2.1`): build `IRFogDemo`, render-verify, bless refs, attach-screenshots, simplify. Keeping `fleet:wip`. — worker (opus, macOS)

jakildev:
Opus feedback pass (worker-4, macOS) — 4th recurrence today. Confirming worker-3 (09:37Z/10:10Z) and worker-1 (10:14Z): `fleet:design-unblocked` here is R7-re-armed (`reconcile_heal_design_unblock` on this claimless orphaned WIP), not architect-set — no `## NEEDS-DESIGN`, no reply to absorb. Per worker-1's guidance I read the thread and no-op'd rather than re-investigate. #2449 still OPEN → all-macOS fleet still can't build/verify `IRFogDemo`, so owed work (build, render-verify, bless refs, screenshots, simplify) still needs a Linux host or the fmt→10.2.1 bump. Clearing the spurious label again to stop this tick's churn; keeping `fleet:wip`. Root-cause fix tracked in #2462. — worker (opus, macOS)

jakildev:
Opus feedback pass (worker-1, macOS) — 5th recurrence (~15:19Z, ~5h after worker-4's 10:22Z strip). Same finding, no re-investigation per standing guidance: `fleet:design-unblocked` is R7-re-armed on this claimless orphaned WIP, not architect-set — no `## NEEDS-DESIGN`, nothing to absorb. Stripping again to stop this tick's churn; keeping `fleet:wip`. Owed work (build `IRFogDemo`, render-verify, bless refs, screenshots, simplify) still needs a Linux host or #2449's fmt bump. Root cause tracked in #2462. — worker (opus, macOS)

jakildev:
Opus feedback pass (worker-2, macOS) — 6th recurrence. Same finding, no re-investigation per worker-1's 10:14Z standing guidance: `fleet:design-unblocked` here is R7-re-armed (`reconcile_heal_design_unblock`) on this claimless orphaned WIP, not architect-set — no `## NEEDS-DESIGN`, nothing to absorb. Stripping again to stop this tick's churn; keeping `fleet:wip`. #2449 still OPEN (fix PR #2465 approved-but-unmerged), so the all-macOS fleet still can't build/verify `IRFogDemo`; owed work (build, render-verify, bless refs, screenshots, simplify) needs a Linux host or #2449 landing. Root cause tracked in #2462. — worker (opus, macOS)

jakildev:
Opus feedback pass (worker-1, macOS) — 6th recurrence (R7 re-arm at 16:03:46Z, after the 15:21:13Z strip). No re-investigation per standing guidance: `fleet:design-unblocked` is R7-re-armed on this claimless orphaned WIP, not architect-set — no `## NEEDS-DESIGN`, nothing to absorb. Stripping again; keeping `fleet:wip`. Owed work (build `IRFogDemo`, render-verify, bless refs, screenshots, simplify) still needs a Linux host or #2465's fmt bump landing. Root cause tracked in #2462 (now `fleet:plan-review`/`human:review-plan`). — worker (opus, macOS)

jakildev:
Opus feedback pass (worker-1, macOS) — 7th+ recurrence today (~20:08Z). No re-investigation per standing diagnosis (worker-1 10:14Z): `fleet:design-unblocked` is R7-re-armed (`reconcile_heal_design_unblock`) on this claimless orphaned WIP — never `fleet:design-blocked`, no `## NEEDS-DESIGN`, nothing to absorb. Stripping again; keeping `fleet:wip`. Owed work (build IRFogDemo, render-verify, bless refs, screenshots, simplify) still needs a Linux host or #2465's fmt bump landing. Root cause #2462 is still `human:review-plan` (plan awaiting human vet) — this label will keep re-arming ~every 3 ticks and burning an opus pane per recurrence until that fix lands; flagging the drain rate to the human via the feedback channel. — worker (opus, macOS)

jakildev (commented):
## Review — render: fog vision Z cost — XY radius with height-penalized reveal (#2260)

**Verdict:** approve

Adds an observer-height + Z-cost term to the analytic fog vision discs
(`dist_xy + zCost * |z - observerZ|`), threaded through the CPU struct
(`FrameDataFogObservers`), both shader backends (GLSL + Metal), and a new
`--edge-zcost` demo scene with committed macOS references.

Traced the std140/Metal layout claim by hand rather than trusting the
comment:

- `FrameDataFogObservers` appends `visionCircleHeights_[8]` **after** the
  existing `vec4[8] + int32 count + 3 pad` tail. The new
  `static_assert(sizeof == 2*N*16 + 16)` computes to 272 bytes, matching.
- `c_fog_to_trixel.glsl`'s block declares only `visionCircleCount` + one pad
  int (offset 136) before the new array — std140 still auto-aligns the
  following `vec4[]` member to the next 16-byte boundary (144), landing at
  the same offset as the C++ struct regardless of how many trailing ints the
  shader chose to declare. `c_voxel_to_trixel_stage_1_body.glsl`'s block
  (only 1 int declared, offset 132) same story — still lands at 144.
- Metal's `FogObserverData` structs declare the full 3-int pad explicitly in
  `c_voxel_to_trixel_stage_1_body.metal`, and Metal's natural alignment
  rules push the new `float4[]` to the same 144 offset in
  `c_fog_to_trixel.metal` (2 ints declared) too. All three sites (C++, GLSL,
  Metal) agree.
- Both untouched consumers (`c_voxel_visibility_compact.*`,
  `c_voxel_to_trixel_stage_2.*`) declare a *shorter* `FogObserverData` that
  simply doesn't reach into the new trailing bytes — fine, since the bound
  buffer is now larger than what they read.
- All UBO allocation/upload sites (`system_fog_to_trixel.hpp`,
  `system_voxel_to_trixel.hpp`) use `sizeof(FrameDataFogObservers)`, so they
  pick up the new size for free — confirmed no site needed a manual byte-count
  edit.

Checked the byte-identity claim (`zCost 0` ⇒ existing scenes unaffected) by
diffing the inlined math against the helpers it replaces:

- `c_fog_to_trixel`'s new inline reveal reduces to exactly
  `fogVisionCircleReveal(pos3D.xy, circle, worldPerPixel)` when
  `heights.y == 0` (same `aa = max(circle.w, worldPerPixel)`, same
  `dist == distEff`). Same for the rim-fade `hardDistPastRim` line.
- `fogColumnRevealZ` / `fogColumnRevealNearestZ` reduce to the pre-existing
  `fogColumnReveal` / `fogColumnRevealNearest` term-for-term at `zCost 0`
  (same `aa`, same `dist`). GLSL and Metal twins match each other.

This independently supports the PR body's claim that the 13 pre-existing
fog shots failing `render-verify` (96.8–99.7% match) are unrelated ref
drift, not a regression from this change — the reveal math this PR touches
is provably a no-op on every existing caller.

### Praise
- The append-after-the-tail layout strategy plus the grown `static_assert`
  is exactly the right way to extend a shared UBO without touching every
  consumer shader — and the PR correctly identifies which consumers *do*
  need the new field (per-pixel/per-voxel) vs. which must stay z-free
  (per-column keep/cut-face tests, which read a superset conservatively).
- Explicitly separating "pre-existing ref drift, not this PR" from the new
  scene's own PASS, with the reasoning to back it up, saves the next
  reviewer from re-deriving whether the 13 FAILs are this PR's fault.

### Nits
- None.

- [x] (already done per PR body) `render-verify --edge-zcost` on macOS —
      PASS, byte-identical, deterministic.
- [ ] Linux/OpenGL smoke to bless `linux-debug` refs — already tracked via
      `fleet:needs-linux-smoke`.

std140/Metal UBO layout change + shader math on both backends).

Opus recheck required: touches engine/render/ (std140/Metal buffer layout
extension + fog reveal shader math across GLSL and Metal backends).

jakildev (commented):
## Opus recheck — render: fog vision Z cost (#2260)

**Verdict:** approve

Sonnet approved and escalated the two things it could only assert by eye: the
std140/Metal UBO layout extension and the byte-identity of the reveal math at
`zCost 0`. I re-derived both from the actual `origin/master` helper sources
rather than the diff comments; both hold rigorously. Stack gate is clear —
base `master`, `MERGEABLE`, not stacked.

### std140 / Metal offset agreement (Sonnet flagged: layout)
Sonnet confirmed the *total size*; I confirmed the *offset* of the new member
agrees across all four declaring sites — the part that corrupts silently if it
drifts:
- C++ `FrameDataFogObservers`: `visionCircles_[8]` (0–128) + count + 3 pad ints
  (128–144), `vec4 visionCircleHeights_[8]` at **144** → 272 = `2·8·16+16`,
  matching the grown `static_assert`.
- `c_fog_to_trixel.glsl` (count + 1 pad int) and
  `c_voxel_to_trixel_stage_1_body.glsl` (count only) declare fewer trailing
  ints, but std140 16-aligns the following `vec4[]`, so the array still starts
  at **144** — same as C++.
- Both Metal twins declare the full pad and 16-align `float4[]` to **144**.

All four agree. The two untouched consumers (`c_voxel_to_trixel_stage_2`,
`c_voxel_visibility_compact`) declare a *shorter* block that never reaches the
new bytes — reading a prefix of a now-larger bound buffer is legal in both GL
and Metal, so leaving them unedited is correct, not a miss.

### Byte-identity at zCost 0 (Sonnet flagged: byte-identity claim)
Checked against the real `ir_iso_common.glsl` / stage-1 definitions on master,
term for term:
- **Per-pixel reveal**: master `fogVisionCircleReveal(xy, circle, worldPerPixel)`
  = `1 - smoothstep(z - max(w,worldPerPixel), z + …, length(xy - c))`. The new
  inline at `heights.y == 0` is `distEff = length(xy - c) + 0`,
  `aa = max(w, worldPerPixel)`, `1 - smoothstep(z-aa, z+aa, distEff)` — identical.
- **Rim fade**: `distEff - z` collapses to `length(xy-c) - z` at zCost 0.
- **`fogColumnRevealZ` / `fogColumnRevealNearestZ`**: same early-outs (1×1,
  bounds, explored threshold), same `nearest` clamp, same `a = max(w,0)` /
  `a = max(w, kFogColumnKeepAa + kFogHiddenKeepCells)`, `distEff = length(...) + 0`.
  Bit-identical to the z-free twins.

`heights.y` is a runtime-uniform 0, so `+ heights.y * abs(z - heights.x)` is an
IEEE `+0.0` on finite world coords (no NaN/Inf path) — exact even under
mad-fusion. So "the 13 pre-existing fog FAILs are ref drift, not this PR" is
provably correct on the GLSL backend too, and the pending
`fleet:needs-linux-smoke` should return byte-identical on every existing shot
(the new `--edge-zcost` shots SKIP for lack of linux-debug refs per the
manifest and get blessed there).

### Keep-vs-drop invariant (the design concern behind the z term)
Could the z-penalized column drop ever cut a voxel a pixel would still reveal?
No. The penalty adds the *same* `zCost·|z - observerZ|` to `distEff` in both
the per-pixel reveal and the own-column drop, while the column keep-ring width
stays z-free and *wider* than the per-pixel AA
(`kFogColumnKeepAa + kFogHiddenKeepCells` vs `worldPerPixel`). Shifting both
boundaries by the same distance preserves the superset relationship, so
`FOG_TO_TRIXEL` always has kept matter to fade at the z-aware rim. The drop
reads the voxel's own `voxelPosition.z`, matching the per-pixel `pos3D.z`
coordinate space (both world Z; observerZ world too). Consistent.

### Sonnet items I did not re-check
Naming, the append-after-tail strategy, the demo scene wiring, and the
mutually-exclusive `--edge-zcost` flag precedence — covered on the first pass
and confirmed clean on my read of the diff.

### Nits
- None.

Approve. Cross-backend byte-identity and layout parity both hold under
independent derivation; the only open item is the Linux smoke, already tracked
by `fleet:needs-linux-smoke`.

jakildev:
judgement-level resolution.

Conflicted files:
- `engine/render/src/shaders/c_voxel_to_trixel_stage_1_body.glsl` — master: `1aa04208 render: one definition for the stage-1/stage-2 mirrored face-select + fog + per-axis store key (#2508)`; PR: `403d3da5 render: fog vision Z cost — XY radius with height-penalized reveal (#2260)`
- `engine/render/src/shaders/metal/c_voxel_to_trixel_stage_1_body.metal` — master: `1aa04208 render: one definition for the stage-1/stage-2 mirrored face-select + fog + per-axis store key (#2508)`; PR: `403d3da5 render: fog vision Z cost — XY radius with height-penalized reveal (#2260)`

Both sides add substantial, different code to the same
region (master's #2508 centralized face-select/fog/per-axis
store-key refactor into `ir_voxel_face_select.glsl`; this PR's
own Z-cost height-penalty fog logic) — not a whitespace-only
conflict.

SHA pair: master=a00603292d9c117dd379faa4afc7150213567a27 × PR=403d3da57f59cccb85154eda2e279284ee013f61

— fleet merger

jakildev:
Resolved semantic conflict against master's #2508 (`ir_voxel_face_select.glsl/.metal` — one definition for the stage-1/stage-2 face-select + fog + per-axis store key).

**What the two sides did.** Master hoisted the fog constants, the `FogObserverData` block, `fogColumnReveal`/`fogColumnRevealNearest`, and the whole exposed-face/cut-face/world-column derivation out of both stage bodies into the shared include, exposing the verdict as `VoxelFaceSelect sel`. This PR grew the same region in place: a `visionCircleHeights[]` UBO field, the `fogColumnRevealZ`/`fogColumnRevealNearestZ` twins, and a Z-aware own-column drop.

**How I reconciled them** (both backends symmetrically):

- **`visionCircleHeights[]` → the shared include.** Forced, not stylistic: GLSL admits exactly one declaration of a named uniform block, and after #2508 that declaration lives in `ir_voxel_face_select.glsl`. Appended after the count so every existing field keeps its offset (std140 16-aligns the array to 144, matching `FrameDataFogObservers::visionCircleHeights_` and the Metal struct's explicit pad tail). The Metal twin gets the same append after its `_fogObsPad0..2`.
- **The Z twins stay in the stage-1 body.** They are only reachable from the own-column DROP, which is stage-1-only by construction (stage 2 never repeats it — its colour taps are rejected by the depth re-test). Keeping them out of the shared include preserves #2508's contract that the include holds exactly what both stages must agree on, and keeps stage 2 from compiling two functions it can never call — the same #1944 cardinal-fast-path concern that made this PR inline the reveal math instead of adding an `ir_iso_common` helper.
- **The drop site is master's `sel.`-based block** with this PR's delta re-applied: `fogColumnRevealZ(sel.worldColumn, voxelPosition.z)` / `fogColumnRevealNearestZ(...)`. The PR's own copies of `keepFace` / `fogActive` / `worldColumn` were dropped wholesale — `selectVoxelFace()` now derives all three. The cut-face test inside `selectVoxelFace` deliberately stays on the **z-free** `fogColumnReveal`, which is what this PR intended: a column spans all z, so the best-case-z test keeps a superset and can never drop a voxel a pixel would reveal.
- Doc comments that pointed at "the z-free twins above" were re-pointed at the shared include.

**Verification (macOS / Metal host).**
- `fleet-build --target IRShapeDebug` clean; `fleet-run IRShapeDebug --auto-screenshot 10` → `RESULT=CLEAN` (Metal compiles shaders at runtime, so the run — not the build — is what proves the resolved kernels and the widened UBO struct compile).
- `render-verify --target IRFogDemo`: **15/15 PASS**, 100.0% match, `max_delta 0` on 14 of 15 (one legacy shot at max_delta 1 / PSNR 110.8). That includes both **`edge_zcost`** shots — the enabled-path gate for this PR's feature. Those references were captured on the pre-rebase branch, so a bit-clean pass across them is direct evidence the Z-cost drop behaves identically through the relocation, and that the rest of the fog suite is unperturbed by the merge.
- Buffer-size safety re-checked on the C++ side: every bind/upload flows from `sizeof(FrameDataFogObservers)` (now 272 B incl. heights), **including** the count-0 `zeroObservers` placeholder that non-fog canvases bind — so no shader declares more than what's bound.

GL/OpenGL remains unbuilt on this host; the GLSL edits mirror the Metal ones line-for-line and `fleet:needs-linux-smoke` still tracks that side.

Reviewer please re-evaluate the rebased diff. — worker

jakildev (commented):
## Review — render: fog vision Z cost — XY radius with height-penalized reveal (#2260)

**Verdict:** needs-fix

### Prior-review resolution

This PR was previously reviewed twice (Sonnet 2026-07-20T20:55:58Z, Opus recheck
2026-07-20T21:03:37Z), both **approve** with **no flagged items** ("### Nits — None").
Since then the merger detected a semantic conflict against master's #2508 (the
stage-1/stage-2 face-select centralization into `ir_voxel_face_select.{glsl,metal}`),
stripped `fleet:approved`, and a worker resolved it at commit `45beb880` (current HEAD).
There is nothing to carry forward from the prior reviews — this re-review is scoped
entirely to the conflict-resolution commit.

### Conflict-resolution audit (the actual delta since the last review)

Traced the resolution by hand rather than trusting the resolution comment's narrative:

- **Layout offset re-verified independently.** `FrameDataFogObservers::visionCircleHeights_`
  sits at C++ offset 144 (`visionCircles_[8]` 0–128 + count/pad0-2 128–144), size 128,
  total 272 = `2*8*16+16` — matches the grown `static_assert`
  (`component_canvas_fog_of_war.hpp:127`). Both `ir_voxel_face_select.glsl` (std140,
  auto-aligned) and its Metal twin (explicit append after the 3-int pad tail) land the
  array at the same offset. Every upload site (`system_fog_to_trixel.hpp:65,101`,
  `system_voxel_to_trixel.hpp:1648,2181,2189`) is `sizeof(FrameDataFogObservers)`-driven,
  including the count-0 `zeroObservers` placeholder — no shader binds more than what's
  uploaded.
- **Reconciliation is internally consistent.** Diffed `403d3da5` (pre-conflict head) vs
  `45beb880` (current) directly: the only PR-specific delta is `worldColumn`/`fogActive`
  (pre-#2508 locals) becoming `sel.worldColumn`/`sel.fogActive` (post-#2508 struct
  fields) at the `ownColumnHidden` call site — the `fogColumnRevealZ`/`fogColumnRevealNearestZ`
  calls themselves are untouched. Read `selectVoxelFace()` in the new shared include:
  the cut-face test at line 205 still calls the **z-free** `fogColumnReveal` (not the Z
  twin), matching the resolution's claim that the superset-safe column test was
  preserved.
- **Backend symmetry confirmed.** GLSL and Metal versions of both new functions
  (`fogColumnRevealZ`/`fogColumnRevealNearestZ`) and the `ownColumnHidden` call site are
  line-for-line mirrors.
- **Blast radius confirmed contained.** `git diff 403d3da5 45beb880` on
  `c_fog_to_trixel.{glsl,metal}`, `component_canvas_fog_of_war.hpp`, and
  `fog_of_war.hpp` is empty — those files were untouched by the rebase, exactly as
  expected since the merger only flagged `c_voxel_to_trixel_stage_1_body.{glsl,metal}`
  as conflicting.

No correctness issues found in the resolution itself — it's a careful, well-reasoned
hand-merge that preserves both sides' invariants.

### Needs-fix
- PR body — the `## Verification` and `## ⚠️ Pre-existing ref drift` sections
  describe the **pre-rebase** state (`403d3da5`): only `--edge-zcost` was run, and a
  "13 existing shots FAIL" claim is attributed to drift "tracked in #2473, left out of
  this PR." I verified independently that `f3c87cc4` (#2473's ref refresh, merged
  2026-07-20) **is already an ancestor of the current HEAD** (`git merge-base
  --is-ancestor f3c87cc4 HEAD` confirms it) — the cited drift was fixed on master and
  pulled in by this branch's rebase before the conflict resolution. The current tree's
  actual state, per the worker's own 2026-07-23T01:21:18Z resolution comment, is
  **15/15 PASS** (14/15 max_delta 0, one legacy shot at max_delta 1/PSNR 110.8 — noise
  floor). Please update the PR body to drop or correct the stale FAIL narrative before
  merge — as written it contradicts the tree it's shipping and would mislead anyone
  reading the merged history.

### Acceptance (issue #2260)

| Criterion | Grade | Basis |
|---|---|---|
| `setVisionCircle`/`addVisionCircle` accept observerZ + zCost; zCost=0 byte-identical | met | Structurally guaranteed by the append-only layout + `distEff` reducing exactly to the old expression at `zCost==0`; confirmed by the (currently-stale-in-the-body, but real) 15/15 render-verify pass above. |
| Lua `IRFog.setVision` surface | unmet, but out of this PR's reach | No `IRFog` binding exists anywhere in `engine/` — confirmed via grep. This is game-repo (cross-repo) territory; the PR body discloses the deferral explicitly, and the plan's own approved Acceptance-criteria list had already softened this to an Approach-step note. Not counted against this PR. |
| `--edge-zcost` scene: tall + low geometry, height-penalized reveal, gated by committed refs | met | New scene + committed `fog_edge_zcost{5,9}.png` refs + manifest wiring, confirmed in the diff. |
| GLSL/Metal twins mirrored; per-column culls stay conservative (best-case z) | met | Line-for-line mirrors confirmed; `c_voxel_visibility_compact.*` and `c_voxel_to_trixel_stage_2*.*` are untouched by this diff — column-level tests stay exactly as they were. |

(Dispatched via the `review-acceptance` grader; findings above are its independent
verification, cross-checked by me against the diff and git history.)

### Nits
- None beyond the PR-body staleness called out above (that's needs-fix, not optional).

### Praise
- The conflict resolution correctly identified and preserved the one subtle invariant
  that mattered here: keeping the cut-face test on the z-free `fogColumnReveal` while
  moving only the own-column DROP to the Z-aware twins — getting this backwards would
  have silently made the per-pixel reveal fade over geometry the column test had
  already dropped (a hole), or vice versa. The resolution comment's own explanation of
  *why* each piece landed where it did is unusually good forensics for a rebase.
- `render-verify`'s ROI-cropped `edge_zcost` shots plus the full-suite byte-identity
  pass together isolate exactly two things this change could break (the new feature,
  and everything else) — good test design, not just "green."

- [ ] Update the PR body's Verification section per the Needs-fix item above.
- [x] (already done, per the 2026-07-23T01:21:18Z comment) `fleet-build IRFogDemo`
  clean; `render-verify --target IRFogDemo` 15/15 PASS on macOS/Metal.
- [ ] Linux/OpenGL smoke to bless `linux-debug` refs — already tracked via
  `fleet:needs-linux-smoke`.

Escalation: author-agent to address, then re-request review. Given the render-pipeline
surface (std140/Metal UBO layout + shader math merged across two independently-evolved
branches on both backends), please Opus-recheck once the PR-body fix lands, even
though the underlying code is not itself in question here.

jakildev:
Addressed the `fleet:needs-fix` item (PR body only — **no code change; HEAD is still `45beb880`**).

**Review checklist**
- [x] **Needs-fix — stale `## Verification` / `## ⚠️ Pre-existing ref drift` sections.** Both rewritten against the tree this PR actually ships. Confirmed your ancestry finding independently: `git merge-base --is-ancestor f3c87cc4 HEAD` → true, so #2473's ref refresh is already in this branch and the "13 existing shots FAIL" narrative described a pre-rebase tree. The drift section is now a short "resolved upstream, no longer applicable" note rather than a live warning.
- [x] **Test plan — update the PR body.** Done, per the above.
- [x] **Test plan — build + render-verify.** Re-ran rather than carrying the prior iteration's numbers forward, since I was about to publish them in the body under my name. Independently reproduced on macOS/Metal at `45beb880`: `fleet-build --target IRFogDemo` clean; `render-verify --target IRFogDemo` full suite **15/15 PASS** — 14 at `max_d 0`, `fog_edge_zoom14` at `max_d 1` / PSNR 110.79 (single-LSB noise floor). Matches the 2026-07-23T01:21:18Z resolution comment exactly.
- [ ] **Linux/OpenGL smoke** — not actionable from this macOS pane; still tracked by `fleet:needs-linux-smoke`.
- Prior Sonnet (07-20T20:55Z) + Opus recheck (07-20T21:03Z) both recorded "### Nits — None", and your re-review scoped itself to the conflict-resolution commit, so there is nothing else carried forward.

**One correction beyond the flagged item:** the `## How` section was also stale post-rebase — it listed the touched shaders from before #2508's face-select centralization and omitted `ir_voxel_face_select.{glsl,metal}`, which the resolution moved the shared `FogObserverData` block declaration into (and which this PR therefore appends `visionCircleHeights` to). Added a bullet for it, noting that its own face-selection math never reads the new field and the cut-face test deliberately stays on the z-free `fogColumnReveal` — the invariant your Praise section called out.

Your escalation note asked for an Opus recheck once the body fix landed. Flagging it here rather than self-applying a reviewer-owned label: the code is byte-for-byte unchanged since the Opus recheck that already approved it, so the recheck surface is the body text alone.

— worker (pool-2, opus)

jakildev (commented):
## Review — render: fog vision Z cost — XY radius with height-penalized reveal (#2260)

**Verdict:** approve

### Prior-review resolution

HEAD is unchanged since the last review (`45beb880`, confirmed via
`git rev-parse --short HEAD`) — the fix was body-only, no new commits.

- ✅ **Needs-fix — stale `## Verification` / `## ⚠️ Pre-existing ref drift`
  sections** — rewritten. The body now carries a `## Verification` section
  dated at `45beb880` with the actual 15/15 PASS numbers, and a `## Ref
  drift — resolved upstream, no longer applicable` section replacing the
  live-warning narrative. Independently re-verified the ancestry claim:
  `git merge-base --is-ancestor f3c87cc4 HEAD` → true, so the "13 shots
  FAIL" narrative correctly describes a pre-rebase tree that's no longer
  what this PR ships.
- ✅ **Test plan — update the PR body** — done, same edit as above.
- ✅ **Test plan — build + render-verify** — the worker re-ran rather than
  carrying forward stale numbers; body now states `fleet-build --target
  IRFogDemo` clean + `render-verify --target IRFogDemo` 15/15 PASS at
  `45beb880`, matching the 2026-07-23T01:21:18Z resolution comment.
- Bonus fix beyond the flagged item: the `## How` section's shader list now
  includes `ir_voxel_face_select` (the #2508-centralized shared include this
  PR's rebase now appends `visionCircleHeights` to) — this was missing
  before and is a real gap the previous body left uncorrected.

No prior Blocker/Needs-fix/Nit items remain open. Praise and test-plan items
from earlier passes aren't re-tracked per the re-review procedure.

### Why this still needs an Opus recheck despite two prior approvals

The Opus recheck on 2026-07-20T21:03:37Z approved a **pre-conflict-resolution**
commit. The merger later stripped that approval when it detected the semantic
conflict against master's #2508 face-select centralization, and a worker
resolved it at the current HEAD (`45beb880`) — a different commit Opus has
never seen. The Sonnet re-review at 2026-07-23T01:50:13Z did an independent
by-hand audit of the resolution's std140/Metal offset math and found it
correct, but that was Sonnet, not Opus — the render/std140/Metal-layout
surface is the fleet's standing Opus-final-review category regardless of how
confident the Sonnet pass is. Since this iteration's only change is the PR
body (no code delta from the already-Sonnet-audited resolution), there is
nothing new for Opus to re-derive on the *code* — but the fleet hasn't yet
recorded an Opus sign-off on `45beb880` itself, so the recheck is about
completing that record, not re-litigating the math.

### Nits
- None.

- [x] Already done (per PR body, `45beb880`): `fleet-build --target
      IRFogDemo` clean; `render-verify --target IRFogDemo` 15/15 PASS
      (macOS/Metal).
- [ ] Linux/OpenGL smoke — already tracked via `fleet:needs-linux-smoke`.

Escalation: please Opus-recheck before merge (touches: `engine/render/` —
std140/Metal UBO layout extension across GLSL + Metal — and Opus has not yet
signed off on the post-conflict-resolution commit `45beb880`).

Opus recheck required: confirm Opus sign-off on the post-conflict-resolution
commit `45beb880` (std140/Metal FrameDataFogObservers layout extension +
fog reveal math on both backends) — the prior Opus approval was on a
different, now-superseded commit.

jakildev (commented):
## Opus recheck — render: fog vision Z cost (#2260), at `45beb880`

**Verdict:** approve

Sonnet's escalation is precise about *why* this needs a second Opus pass: the
2026-07-20 Opus approval was on a pre-conflict-resolution commit that the
merger's #2508 semantic-conflict detection invalidated. So this pass re-derives
the layout and byte-identity claims **from the files at `45beb880` itself**, not
from the diff comments or the earlier review, and adds the cross-stage
invariant checks a diff-local read can't reach.

Stack gate: base `master`, `MERGEABLE`, not stacked, no amending/conflict
labels. Since the merge-base (`a0060329`), master has touched
`engine/render/src/shaders` exactly once — `74967173` (Metal AOT build gating,
no shader source) — so nothing upstream re-opens the resolved conflict.

### 1. std140 / Metal offset agreement — re-derived at all five declaring sites

Sonnet confirmed the C++ `static_assert` and two GLSL blocks. The failure mode
that matters is *offset drift at a site nobody listed*, so I enumerated every
declaration of the block on both backends:

| site | members before the append | new member offset |
|---|---|---|
| C++ `FrameDataFogObservers` | `vec4[8]` + count + 3 pad ints | **144** (272 total, matches the grown assert) |
| `ir_voxel_face_select.glsl` | `vec4[8]` + count | 144 (std140 16-aligns the `vec4[]`) |
| `c_fog_to_trixel.glsl` | `vec4[8]` + count + 1 pad int | 144 (same rule) |
| `metal/ir_voxel_face_select.metal` | `float4[8]` + 4 ints | 144 (float4 align 16) |
| `metal/c_fog_to_trixel.metal` | `float4[8]` + 2 ints | 144 (float4 align 16 pads 136→144) |

The two Metal structs disagree on how many trailing `int` lanes they spell out
— that's exactly the case that would corrupt silently if `float4` didn't
re-align — and both land on 144 anyway. Confirmed there is no sixth
declaration: `c_voxel_visibility_compact.{glsl,metal}` keep the **short**
144-byte block and read only `visionCircles` / `visionCircleCount`, whose
offsets are untouched; reading a smaller block out of a larger buffer is
well-defined. `kMaxFogVisionCircles == 8` verified identical at all six mirrors.

### 2. Buffer sizing — the append is only safe if every allocation is `sizeof`-driven

The check a diff-only read misses entirely: the struct grew 144 → 272, so any
hardcoded size or partial upload would leave `visionCircleHeights` reading
uninitialized GPU memory. Every producer is `sizeof(FrameDataFogObservers)`:

- `system_fog_to_trixel.hpp:65` upload, `:101` `FogObserverData` creation
- `system_voxel_to_trixel.hpp:1648` upload, `:2181` `FogCullObservers` creation,
  `:2189` the count-0 seed

No literal sizes, no `offsetof`-relative writes, no split uploads. Both slot-27
buffers grow together, so stage 2 / winner-resolve / feeder — which now declare
the 272-byte block through the shared include but never read the new member —
are always bound a buffer at least as large as their block.

### 3. Cross-stage invariant: the z penalty is monotone, so the cull superset holds

The invariant no single file states, and the one I'd most want checked before
merging: `c_voxel_visibility_compact` is a **coarse pre-cull that must keep a
superset** of what stage 1 renders (its `kCullSafetyCells` mirror comment says
so), and it deliberately stays z-free here.

That survives only because `distEff = dist + zCost·|z − observerZ| ≥ dist` for
`zCost ≥ 0`, and `smoothstep` is monotone non-decreasing, so
`fogColumnRevealZ ≤ fogColumnReveal` **pointwise**. The z-aware drop can only
drop *more*, never keep something the compact already culled. A **negative**
`zCost` would invert that and produce voxels stage 1 keeps but the compact
already threw away — missing matter, no shader error.

`addVisionCircle` clamps with `IRMath::max(zCost, 0.0f)`, so the public API
can't get there. That clamp is load-bearing for a cross-shader invariant three
files away, which the comment beside it doesn't say (nit below).

The #2124 keep-ring holds for the same reason and in the right space: the ring
width (`kFogColumnKeepAa + kFogHiddenKeepCells`) stays z-free but is applied *to
the penalized distance*, and `c_fog_to_trixel`'s `hardDistPastRim` now also
measures `distEff − radius`. Both sides moved to the same effective metric, so
the cross-section cap still has matter to repaint at the z-shifted boundary — it
doesn't silently thin out as `zCost` grows.

### 4. Byte-identity at `zCost 0` — from the master helper, not the comment

The PR replaces two `fogVisionCircleReveal(...)` calls with inlined math, so
"byte-identical" depends on the inline being that helper *exactly*. Read
`ir_iso_common.{glsl,metal}:fogVisionCircleReveal` at HEAD:
`dist = length(worldXY − circle.xy)`, `a = max(circle.w, aa)`,
`1.0 − smoothstep(circle.z − a, circle.z + a, dist)`.

- `c_fog_to_trixel`: inline uses `aa = max(circle.w, worldPerPixel)` and the
  same smoothstep — the old call site passed `worldPerPixel`. Match.
- `fogColumnRevealZ`: `a = max(w, 0.0)` — the z-free twin passes `aa = 0.0`. Match.
- `fogColumnRevealNearestZ`: `a = max(w, kFogColumnKeepAa + kFogHiddenKeepCells)`
  with the same `clamp`-derived `nearest`. Match.

At `zCost == 0` the added term is `0 · |Δz| = +0.0` for all finite `Δz`, and
`length(...) + 0.0` is exactly `length(...)` (the addend is never `−0.0` since
`length ≥ 0`); under FMA contraction `fma(0, Δz, L) == L` exactly. So the
identity is structural, not just empirical — and the 14/15 `max_d 0` shots agree
with it. Sonnet flagged the `fog_edge_zoom14` `max_d 1`; on closer read I agree
it's the known single-LSB floor and not a `zCost`-path effect — that shot runs
`zCost 0`, where the expression provably reduces to master's.

### 5. Conflict-resolution cleanliness (what invalidated the last approval)

Checked at `45beb880` directly, since an auto-merge can blend stale identifiers
*outside* conflict markers with no `<<<<<<<` left behind: no conflict markers,
and no duplicate definitions — `fogColumnReveal` / `fogColumnRevealNearest`
exist only in the shared `ir_voxel_face_select.{glsl,metal}`, the `*Z` twins
only in the stage-1 bodies. Include order is correct on both backends
(`ir_voxel_face_select` before the body in
`c_voxel_to_trixel_stage_1.{glsl,metal}`), which is what makes the bodies' use
of `kFogColumnCellHalf` / `kFogColumnKeepAa` / `kFogHiddenKeepCells` /
`visionCircleHeights` resolve. `f3c87cc4` (the #2473 ref refresh) confirmed an
ancestor of the head, so the body's retired "13 FAILs" narrative is correctly
dropped.

Putting the `*Z` twins in the stage-1 body rather than the shared include is the
right call and matches #2508's intent: the drop is stage-1-only, so stage 2
would otherwise compile two functions it can never call, and the shared include
stays exactly the definitions both stages must agree on.

No CPU-side mirror of the vision-circle math exists anywhere in `engine/` or
`creations/` (grepped) — the reveal formula lives only in shaders, so there is
no second copy to drift.

### 6. Scene gate

`--edge-zcost` is a genuine positive-fire test: it is the only run with
`zCost > 0`, and the manifest note says so explicitly, with the other runs
serving as the `zCost 0` byte-identity control. Flag precedence, shot table, and
`capture_offset 0` mapping are consistent with the sibling scenes.

### Nits
- `component_canvas_fog_of_war.hpp` — the `IRMath::max(zCost, 0.0f)` clamp in
  `addVisionCircle` is what keeps `c_voxel_visibility_compact`'s z-free cull a
  superset of stage 1's z-aware drop (§3). Worth one clause saying so, since
  `observers_` is public and the clamp otherwise reads as ordinary defensive
  hygiene rather than an invariant guard.
- Stage 1 passes `roundHalfUp(voxelPosition.xyz).xy` for the column but raw
  `voxelPosition.z` for the height at the per-axis site. Equivalent for
  integer-positioned voxels and the smoothstep AA absorbs the rest, but the
  mixed rounding is a small readability wart.

Neither blocks merge — no re-review round wanted for either.

### Observation (no action)
`isCutFace` stays on the z-free `fogColumnReveal`, so where the reveal boundary
is now a *cone* rather than a cylinder, the sliced surface isn't force-lit as a
cut wall (and ±Z faces are never cut by design). The matter is still present and
fog-tinted, so it doesn't read hollow, and the new refs show the intended fade —
but if a future scene pushes `zCost` high enough to slice a large object
horizontally, that's where the cut-surface treatment will need to follow the z
metric. Worth a follow-up issue only if such a scene appears.

### Test plan
- [x] macOS/Metal at `45beb880`: `fleet-build --target IRFogDemo` clean,
      `render-verify --target IRFogDemo` 15/15 PASS (14 byte-identical) — the
      Metal compile+run the semantic-conflict resolution required.
- [ ] Linux/OpenGL smoke to bless the `linux-debug` refs — tracked via
      `fleet:needs-linux-smoke`. The GLSL side is unbuildable on this host; it
      mirrors the verified Metal line-for-line and every identifier it uses
      resolves through the include order checked in §5.

🤖 Opus recheck by Claude (opus-reviewer role)

jakildev:
Addressed both Opus-recheck nits at `a9511219`. **Comment-only — 22 added lines, zero deletions, no code lines**, so the byte-identity claim the recheck re-derived at `45beb880` is untouched by construction.

- **Nit 1 — `component_canvas_fog_of_war.hpp`, the `IRMath::max(zCost, 0.0f)` clamp.** Added the clause: the clamp is what keeps `c_voxel_visibility_compact`'s z-free coarse cull a superset of stage 1's z-aware drop (`distEff = dist + zCost·|z − observerZ| ≥ dist` only for `zCost ≥ 0`), and a negative `zCost` inverts it into silently-missing matter with no shader error. Also notes `observers_` is public, so a caller writing `visionCircleHeights_` directly owns the invariant itself — which was the recheck's specific concern about it reading as ordinary defensive hygiene.

- **Nit 2 — mixed rounding at the per-axis site (both backends).** Documented the asymmetry rather than unifying it, because **unifying would have been a real regression**: the column is rounded because it indexes the integer fog grid, but the height feeds a continuous smoothstep distance — rounding it would quantize the fade into whole world-Z steps *and* disagree with `c_fog_to_trixel`, which penalizes against the unrounded `pos3D.z` (verified at `c_fog_to_trixel.glsl:229`). The comment now states that, so the next reader can tell the split is deliberate.

Independently re-verified the two claims before writing them down rather than taking them from the review text: `c_voxel_visibility_compact.glsl` does declare only the short block and keeps a z-free `radius + edge + kCullSafetyCells` ring, and `c_fog_to_trixel.glsl:229` does use `abs(pos3D.z − heights.x)`.

**Verification (macOS/Metal, this host):** `fleet-build --target IRFogDemo` clean; `render-verify --target IRFogDemo` **15/15 PASS** — identical to the numbers recorded at `45beb880`, including the known single-LSB `fog_edge_zoom14` `max_d 1` floor. The Metal shaders recompiled, so the comment edits are confirmed harmless on the backend this host can actually exercise.

The `fleet:needs-linux-smoke` item is unchanged and still outstanding — the GLSL edit is the same comment block and this host can't build GL.

The recheck's `### Observation (no action)` (`isCutFace` staying on the z-free reveal) was deliberately not acted on, per its own "follow-up issue only if such a scene appears". — worker (opus, macOS)

jakildev (commented):
## Review — render: fog vision Z cost — XY radius with height-penalized reveal (#2260)

**Verdict:** approve

### Prior-review resolution

This PR carries two full approvals at `45beb880` (Sonnet 2026-07-23T02:03:29Z,
Opus recheck 2026-07-23T02:13:49Z) with no open items. One commit landed
since, `a9511219`, in response to the `human:re-review` re-tag that fires
automatically on any push after fleet approval:

- ✅ **Opus-recheck Nit 1** — `component_canvas_fog_of_war.hpp`: added the
  clause documenting that the `IRMath::max(zCost, 0.0f)` clamp is what keeps
  `c_voxel_visibility_compact`'s z-free coarse cull a superset of stage 1's
  z-aware drop. Verified fixed — comment text matches the invariant Opus
  described almost verbatim.
- ✅ **Opus-recheck Nit 2** — mixed rounding at the per-axis site (both
  GLSL/Metal stage-1 bodies): documented rather than unified, with the
  rationale (column indexes the integer fog grid; height feeds the
  continuous smoothstep and must stay unrounded to agree with
  `c_fog_to_trixel`). Verified the cited line: `c_fog_to_trixel.glsl:229`
  does read `heights.y * abs(pos3D.z - heights.x)` against the unrounded
  `pos3D.z`, confirming the comment's claim.

No prior Blocker/Needs-fix/Nit items remain open.

### Delta audit (`45beb880` → `a9511219`)

Independently diffed the two commits rather than trusting the commit
message's "comment-only" claim:

```
 component_canvas_fog_of_war.hpp                | 9 +++++++++
 c_voxel_to_trixel_stage_1_body.glsl             | 7 +++++++
 metal/c_voxel_to_trixel_stage_1_body.metal      | 6 ++++++
 3 files changed, 22 insertions(+)
```

All 22 added lines are `//`-prefixed comment lines inserted above existing
statements — zero non-comment tokens changed, zero deletions. This can't
alter compiled behavior on either backend (GLSL/Metal `//` comments are
stripped identically to C++), so the byte-identity and layout claims both
prior reviews independently re-derived at `45beb880` carry forward
unchanged by construction.

### Acceptance (issue #2260)

Already graded by the `review-acceptance` dispatch in the
2026-07-23T01:50:13Z re-review (3 of 4 criteria met, the Lua `IRFog`
surface correctly deferred as out-of-repo-reach per the plan's own softened
wording). This delta touches none of the graded surfaces — no re-grade
needed.

### Nits
None.

- [x] macOS/Metal: `fleet-build --target IRFogDemo` clean, `render-verify
      --target IRFogDemo` 15/15 PASS — re-confirmed by the author at
      `a9511219`, unchanged from `45beb880`'s numbers (expected, given the
      diff is comment-only).
- [ ] Linux/OpenGL smoke to bless the `linux-debug` refs — still tracked by
      `fleet:needs-linux-smoke`, unaffected by this delta.

### Why no third Opus pass
The substantive code at `45beb880` already has a full Opus-recheck
approval that independently re-derived the std140/Metal layout, the
byte-identity math, and the cull-superset invariant from the files
themselves. The only change since is the two comment additions Opus
itself asked for, and I've verified both against the actual source lines
they describe (not just the diff/commit narrative). There is no new
render/entity/system/world logic for a second Opus pass to check —
spending Opus budget re-deriving invariants over a comment-only diff
would be pure churn.

Escalation: none — comment-only delta on an already Opus-approved commit. Opus recheck not required.

jakildev:
Human-cued merge (game-architect, on jakildev's instruction): assessment against the irreden game's vision design (uphill-cost vision, world_design.md verticality pillar) passed — the linear zCost model is the right foundation and the append-only layout leaves the two free vec4 slots for the tunable-curve follow-up (asymmetric up/down cost + penalty-free height band), filed separately.

**Outstanding OpenGL tier:** `fleet:needs-linux-smoke` is intentionally merged over — Metal verified 15/15; existing linux-debug refs stay valid by the zCost-0 byte-identity argument, and the two new `fog_edge_zcost` shots SKIP on linux until blessed. Next Linux session: `render-verify --target IRFogDemo --update-references` to bless, fix-forward if the GLSL mirror misbehaves.

PR #109, opened 2026-04-13
fleet: game repo review coverage + single-command bash rule

jakildev:
## Summary
- Reviewers (sonnet-reviewer + opus-reviewer) now poll both `jakildev/IrredenEngine` AND `jakildev/irreden` (game repo) for open PRs
- Game PRs are reviewed via `gh pr diff` since they can't be checked out into engine worktrees
- Added prominent `## CRITICAL: single-command Bash calls only` section at the top of every role file to reduce compound command violations
- Created `fleet:wip`, `fleet:approved`, `fleet:needs-fix`, `fleet:blocker` labels on the game repo
- Updated `~/.tmux.conf` with macOS fallback keybindings (M-b/M-f) for Option+Left/Right pane navigation

## Test plan
- [ ] Verify `fleet-up dry-run` launches without errors
- [ ] Verify Option+Left/Right navigates panes on macOS
- [ ] Verify reviewers pick up game repo PRs in their startup actions
- [ ] Verify agents no longer use compound bash commands (or at least less frequently)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #108, opened 2026-04-13
build(linux): add missing <algorithm> to prefab headers using std::sort

jakildev:
## Summary
- `engine/prefabs/irreden/update/components/component_periodic_idle.hpp`: add `<algorithm>` — `sortSequence()` calls `std::sort`, `mapAngleToStageTValue()` calls `std::max`/`std::min`
- `engine/prefabs/irreden/audio/components/component_midi_sequence.hpp`: add `<algorithm>` — `sortSequence()` calls `std::sort`

Both headers rely on `ir_math.hpp` → GLM. GLM does **not** pull in `<algorithm>`, so `std::sort` is unavailable through any transitive path. On Windows/MSYS2 this is hidden by the wider system header pull-in; on Linux with GCC these are clean build failures.

Part of the **Linux build maturation: get `linux-debug` preset green end-to-end** umbrella task.

## Test plan
- [ ] `cmake --build build --target IrredenEngineTest -j$(nproc)` on WSL2 Ubuntu 24.04 with `linux-debug` preset
- [ ] `cmake --build build --target IRShapeDebug -j$(nproc)` on WSL2 Ubuntu 24.04

## Notes for reviewer
These are header additions only; no logic changes. Build not verified on macOS (build tree not configured on this host; rule prohibits `cmake --preset`). The standard requires explicit `<algorithm>` for `std::sort`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #110, opened 2026-04-13
test: add color-conversion and layout-helper unit tests

jakildev:
## Summary
- New `test/math/ir_math_color_layout_test.cpp` with 31 tests covering
  the color-conversion and layout-helper functions in `ir_math.hpp` /
  `ir_math.cpp` that had no test coverage.
- **Color:** `Color::toPackedRGBA` packing order, `colorHSVToColor`
  primary hues + white/black + alpha, `colorToColorHSV` hue/sat/val
  extraction + round-trip fidelity, `applyHSVOffset` identity,
  hue-rotation, and saturation clamping.
- **Layout:** `layoutGridCentered` single-item/centering/spacing/depth,
  `layoutCircle` radius invariant / start-angle / sum-to-zero,
  `layoutSquareSpiral` first two indices and spacing.
- Added `math/ir_math_color_layout_test.cpp` to `IrredenEngineTest`
  sources in `test/CMakeLists.txt`.

## Test plan
- [ ] `cmake --build build --target IrredenEngineTest -j$(nproc)` builds
  cleanly (note: `test/CMakeLists.txt` may have a trivial merge conflict
  with PR #107 which also adds a line to the same executable target —
  resolve by including both additions)
- [ ] All new tests pass: `./IrredenEngineTest --gtest_filter='ColorPack*:ColorHSV*:ColorToColorHSV*:ApplyHSV*:LayoutGrid*:LayoutCircle*:LayoutSquareSpiral*'`
- [ ] Existing math/physics/ECS tests still pass

## Notes for reviewer
- `Color::toPackedRGBA` packing order: R in bits [7:0], G [15:8],
  B [23:16], A [31:24] — verified from the constexpr source.
- HSV primary-color tests use `EXPECT_LE(channel, 1u)` for the
  near-zero channels to allow 1 byte of float rounding error from
  GLM's `rgbColor`.
- `applyHSVOffset` hue-rotation test shifts red (hue≈0) by +1/3 → green;
  the `NegativeSaturationGivesEqualChannels` test clamps sat to 0,
  producing white (val=1) — the channels-equal check is the invariant.
- The two open PRs that also touch `test/CMakeLists.txt` (#107) will
  conflict on the `IrredenEngineTest` source list — trivial to resolve
  by keeping all three additions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2468, opened 2026-07-20
fleet: fleet-plan-lint concept-based core-section matching (#2443)

### LINKED ISSUE #2443: fleet-plan-lint: leading-token heading match hard-fails sound plans (destructive — bounce deletes the plan comment)
**Area:** scripts/fleet (fleet-plan-lint)
**Model:** sonnet
**Blocked by:** (none)

## What happened

Verified 2026-07-16 (macOS opus-reviewer pane) while vetting the `fleet:plan-review` gate on #2442.

`fleet-plan-lint` **hard-failed a sound plan** on heading nomenclature. Per
`role-opus-reviewer`, exit 1 means "go straight to the Not sound bounce ... do **not** spend
the Opus judgment on what the lint already decided" — and a bounce **deletes the `## Plan`
comment** before swapping back to `fleet:needs-plan`. So a false positive here is not a
warning; it destroys a good plan and forces a full re-plan.

I overrode it manually on #2442 after reading the source. An agent following the role doc to
the letter would not have.

## Root cause

`fleet-plan-lint:104-105` anchors each core section to a heading that **begins** with the
literal word:

```python
missing_core = [s for s in ("Scope", "Approach", "Acceptance")
                if not re.search(r"(?im)^\s*#{1,4}\s+" + s + r"\b", plan)]
...
if len(missing_core) >= 2:
    hard.append("missing core sections: ...")   # -> exit 1
```

A heading that *contains* the concept but doesn't *lead* with the token misses. Two such
headings → `>= 2` → hard fail.

## Repro

```python
import re
plan = '''## Plan: x
### Files / modules
### Committed approach — derive at fetch time
### Acceptance tests (positive-fire)
### Gotchas
'''
for s in ('Scope','Approach','Acceptance'):
    print(s, bool(re.search(r'(?im)^\s*#{1,4}\s+' + s + r'\b', plan)))
# Scope      False   <- "Files / modules" IS the scope
# Approach   False   <- "Committed approach" IS the approach
# Acceptance True    <- matches only by leading-word luck
# => 2 missing => HARD FAIL
```

Live: `fleet-plan-lint 2442` → `FAIL #2442: missing core sections: Scope, Approach`, on a
plan that carries a verified current-state section, a single committed 4-phase approach,
rejected alternatives, sibling reconciliation, and positive-fire acceptance tests.

Note the failure is *self-masking*: `### Acceptance tests` passes for the same accidental
reason the others fail, so the rule looks like it works right up until a plan words two
headings naturally.

## Why this is worse than a normal lint bug

The `>= 2` threshold exists to tolerate format variance (a single miss is only a `warn` at
:111-112). But variance is **correlated**, not independent — a plan author who writes
"Files / modules" instead of "Scope" is exactly the author who writes "Committed approach"
instead of "Approach". The threshold assumes independent slips and so fires precisely on
the consistent-voice plans it should pass.

## Suggested shape

Match the **concept**, not the leading token — accept a small synonym set per core section
and allow the word anywhere in the heading:

- Scope → `scope | files | modules | affected`
- Approach → `approach | plan | design | implementation`
- Acceptance → `acceptance | verification | tests | done`

i.e. `^\s*#{1,4}\s+.*\b(?:approach|design|implementation)\b` rather than anchoring at the
start. Keep the `>= 2` hard-fail once matching is concept-based.

Given a false positive here is destructive (plan comment deleted), consider also demoting
`missing core sections` from hard-fail to warn, and let the reviewer's judgment pass make
the call — the lint's stated job is structure, and "these headings are named unusually" is
not the same finding as "this plan has no approach".

## Acceptance criteria

1. The #2442 plan body (headings `### Files / modules`, `### Committed approach — …`,
   `### Acceptance tests (positive-fire)`) lints **exit 0**.
2. A plan genuinely missing an approach section (no approach/design/plan heading at all)
   still hard-fails — regression test, must fail on master before the fix if written as a
   negative control.
3. Test coverage pins both: the synonym-heading plan passes, the truly-sectionless plan
   fails. Both tests demonstrably exercise the new matcher (show the red run for the
   false-positive test against master).

jakildev:
## Summary
- `fleet-plan-lint`'s core-section check anchored `Scope`/`Approach`/`Acceptance` at the *start* of the heading, so a sound plan worded naturally (`### Files / modules`, `### Committed approach — …`) hard-failed with `missing_core >= 2` — a destructive false positive, since the "Not sound" bounce deletes the `## Plan` comment.
- Replaced the leading-token matcher with a per-concept synonym-set search (word anywhere in the heading), excluding "plan" from the Approach synonyms so the `## Plan: <title>` heading itself can't vacuously satisfy it. The `>= 2` hard-fail threshold and every other check is unchanged.

## Test plan
- [x] `bash scripts/fleet/tests/test_fleet_plan_lint.sh` — 15/15 pass, including new fixtures #108 (synonym-headed plan, exit 0) and #109 (negative control missing 2 concepts, exit 1).
- [x] Confirmed #108 fails on the unfixed matcher (stashed the fix, re-ran, saw `expected 0, got 1`) before restoring — demonstrates the regression is real, not a vacuous pass.
- [x] `ruff check scripts/` — clean (no `.py` files touched; the bash-wrapped-python parse warning on `fleet-plan-lint` when force-targeted directly is pre-existing on master, confirmed via stash).
- Pure bash/Python fleet-tooling change — no C++ build required.

Closes #2443

### DIFF
--- a/.fleet/plans/issue-2443.md
+++ b/.fleet/plans/issue-2443.md
@@ -1,0 +1,114 @@
+## Plan: fleet-plan-lint concept-based core-section matching (kill the destructive leading-token false positive)
+
+- **Issue:** #2443
+- **Model:** sonnet — single-file regex change in `scripts/fleet/fleet-plan-lint` plus two test fixtures; the approach below is concrete enough that implementation is bounded.
+- **Date:** 2026-07-20
+
+### Scope
+
+Make `fleet-plan-lint`'s core-section check match the **concept** of each core heading
+(Scope / Approach / Acceptance) rather than requiring the heading text to *begin* with the
+literal token, so a sound plan that words its headings naturally no longer hard-fails. The
+`>= 2`-missing hard-fail and every other check are preserved.
+
+### Verified current state + confirmed repro
+
+Read `scripts/fleet/fleet-plan-lint:104-112` on `origin/master` (5c53d2e5). The core-section
+check anchors each token at the **start** of the heading:
+
+```python
+missing_core = [s for s in ("Scope", "Approach", "Acceptance")
+                if not re.search(r"(?im)^\s*#{1,4}\s+" + s + r"\b", plan)]
+...
+if len(missing_core) >= 2: hard.append("missing core sections: ...")   # exit 1
+elif missing_core:         warn.append(...)                            # exit 0
+```
+
+Confirmed the miss against the actual matcher (not the issue's guess) with the #2442-shaped
+headings `### Files / modules`, `### Committed approach — …`, `### Acceptance tests (positive-fire)`:
+
+- `Scope` → miss (heading leads with "Files"), `Approach` → miss (leads with "Committed"),
+  `Acceptance` → match (leads with "Acceptance" by luck) ⇒ `missing_core = [Scope, Approach]`,
+  len 2 ⇒ **hard fail (exit 1)** on a sound plan. Reproduced locally via the exact regex.
+
+A hard fail here is **destructive**: per `role-opus-reviewer`, exit 1 routes straight to the
+"Not sound" bounce, which **deletes the `## Plan` comment** before swapping back to
+`fleet:needs-plan` — a false positive forces a full re-plan, not a warning.
+
+The lint fetches via `gh issue view`, so its tests mock `gh` with a PATH-shim fake keyed by
+issue number (`scripts/fleet/tests/test_fleet_plan_lint.sh:47-83`). Existing fixtures #100
+(GOOD_PLAN) and #103 (skeletal, missing all core sections) already pin PASS / hard-fail.
+
+### Approach
+
+One approach, picked. Match the **concept**, not the leading token: per core section, accept
+a small synonym set and allow the word **anywhere** in the heading (`.*\b(?:…)\b`). Keep the
+`>= 2`-missing hard-fail and the `== 1` warn unchanged — only the per-section matcher changes.
+
+1. In `scripts/fleet/fleet-plan-lint`, replace the `missing_core` comprehension (lines
+   104-105) with a per-concept synonym-set search:
+
+   ```python
+   CORE = {
+       "Scope":      ("scope", "files", "modules", "affected"),
+       "Approach":   ("approach", "design", "implementation"),
+       "Acceptance": ("acceptance", "verification", "tests", "done"),
+   }
+   missing_core = [
+       name for name, syns in CORE.items()
+       if not re.search(r"(?im)^\s*#{1,4}\s+.*\b(?:" + "|".join(syns) + r")\b", plan)
+   ]
+   ```
+
+   Leave lines 106-112 (soft `Affected files`/`Gotchas` warns, the `>= 2` / `== 1`
+   branch) and every other check untouched. `missing_core` stays a list, so the existing
+   `>= 2` / `elif` logic and the `", ".join(missing_core)` message keep working.
+
+2. Add two fixtures + assertions to `scripts/fleet/tests/test_fleet_plan_lint.sh`:
+   - **#108 synonym-heading plan** (the false-positive regression): a plan whose headings are
+     `### Files / modules`, `### Committed approach — …`, `### Acceptance tests (positive-fire)`,
+     `### Gotchas` with a body. Assert **exit 0**. This is the red-against-master demonstration
+     (AC 3): it exits 1 on master and 0 after the fix.
+   - **#109 negative control**: a plan with a real `### Approach` section but **no** Scope-concept
+     and **no** Acceptance-concept heading (e.g. `### Approach` + `### Notes` only) ⇒
+     `missing_core = [Scope, Acceptance]`, len 2 ⇒ assert **exit 1**. This proves the broadened
+     matcher did not become so permissive it passes a genuinely-broken plan.
+
+### Affected files
+
+- `scripts/fleet/fleet-plan-lint` — lines 104-105: leading-token matcher → concept synonym-set matcher.
+- `scripts/fleet/tests/test_fleet_plan_lint.sh` — two new fixtures (#108, #109) + assertions.
+
+### Acceptance criteria
+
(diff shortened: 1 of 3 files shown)


jakildev (commented):
## Review — fleet: fleet-plan-lint concept-based core-section matching (#2443)

**Verdict:** approve

Standalone PR (`baseRefName == master`), `mergeable == MERGEABLE` — no stack gate, no churn audit needed. Pure fleet-tooling change (bash-wrapped Python + test fixtures); no C++ build required.

### What changed

- `fleet-plan-lint`'s core-section check (`Scope`/`Approach`/`Acceptance`) previously anchored each token at the *start* of the heading text, so a sound plan worded naturally (`### Files / modules`, `### Committed approach — …`) hard-failed with `missing_core >= 2` — a destructive false positive, since `role-opus-reviewer`'s "Not sound" bounce deletes the `## Plan` comment on exit 1.
- Replaces the leading-token matcher with a per-concept synonym-set search (`scripts/fleet/fleet-plan-lint:104-116`) that allows the word anywhere in the heading. `Approach`'s synonym set deliberately excludes "plan" so the `## Plan: <title>` heading itself can't vacuously satisfy it.
- Adds two fixtures to `test_fleet_plan_lint.sh`: #108 (the #2442-shaped synonym-headed plan, must pass) and #109 (negative control missing Scope + Acceptance concepts, must still hard-fail).

### Verification performed

- Ran `bash scripts/fleet/tests/test_fleet_plan_lint.sh` on the PR branch — 15/15 pass, including the two new fixtures.
- Independently confirmed the regression is real (not a vacuous test): ran the *old* leading-token regex against the #108 `SYNONYM` fixture body in isolation — it reports `missing_core = ['Scope', 'Approach']`, which is `>= 2` and would hard-fail on unfixed master. The new matcher correctly resolves this to `[]`.
- Manually traced #109 (negative control): `### Approach` matches the Approach synonym directly, `### Notes` matches nothing → `missing_core = [Scope, Acceptance]`, len 2 → hard fail, as claimed.
- Confirmed `missing_core` stays a `list` of the same three name strings ("Scope"/"Approach"/"Acceptance"), so the downstream `>= 2`/`elif` branching and the `", ".join(missing_core)` warn message are unaffected by the refactor.
- Checked the regex for correctness: no `.` DOTALL, so each heading is still evaluated per-line as before; synonym strings need no escaping (no special regex chars); dict iteration order (Python 3.7+ insertion order) preserves the original Scope→Approach→Acceptance ordering in warn messages.

### Nits

- No fixture directly exercises the documented "plan" exclusion (i.e. a plan with only a `## Plan: <title>` heading and no real `### Approach`-shaped heading, to prove Approach still reports missing rather than vacuously matching via a `plan` synonym). It's correctly excluded in the code and the Gotchas section of the plan doc calls out *why*, but a future synonym-set edit that accidentally re-adds `"plan"` wouldn't be caught by the current fixtures (#108/#109 both have explicit Approach-concept headings). Consider a small follow-up fixture for this specific edge case.
- `### Affected files` now also satisfies the Scope concept (via `files`/`affected`), which is called out and accepted as a deliberate trade-off in the plan doc's Gotchas section — no action needed, just noting it's intentional in case a future reader is surprised by it.

- [x] Already covered — `test_fleet_plan_lint.sh` 15/15 green, regression independently reproduced against the old matcher.

Opus recheck not required.

jakildev:
Addressed the review nit (fleet:has-nits):

- Added fixture #110 + paired assertion to `test_fleet_plan_lint.sh` that directly exercises the documented "plan" exclusion — a plan whose only Approach-candidate heading is the mandatory `## Plan: <title>` line (Scope + Acceptance concepts present, no Approach-shaped heading). The lint must report Approach missing (single missing core → warn, exit 0). If a future synonym-set edit re-adds `"plan"` to the Approach set, `missing_core` goes empty and the warn vanishes → the assertion fails.
- Verified the guard is non-vacuous by simulating the matcher with/without `"plan"` in the Approach synonyms against the exact fixture body: correct code yields `missing_core=['Approach']` (warn fires), the leaked variant yields `missing_core=[]` (no warn).

Suite 15 → 17 assertions, all green. The second nit (`### Affected files` also satisfying Scope) was explicitly "no action needed — intentional", so no change there. fleet:approved retained. — worker (opus)

jakildev (commented):
## Review — fleet: fleet-plan-lint concept-based core-section matching (#2443)

**Verdict:** approve

Re-review: one new commit (`07b4a8b`) landed since the prior approval (`13962e5`, submitted 2026-07-20T20:00:58Z), addressing the prior review's single nit.

### Prior-review resolution
- ✅ `scripts/fleet/tests/test_fleet_plan_lint.sh` — no fixture directly exercised the "plan" exclusion from the Approach synonym set — verified fixed at `07b4a8b`

### Verification performed

- Re-ran `bash scripts/fleet/tests/test_fleet_plan_lint.sh` on HEAD — 17/17 pass, including the new fixture #110 and its paired assertion.
- Traced the new `PLAN_ONLY_NO_APPROACH` fixture against `fleet-plan-lint`'s `CORE` synonym table (`scripts/fleet/fleet-plan-lint:108-112`): the fixture's only headings are `## Plan: plan-heading only`, `### Scope`, `### Affected files`, `### Acceptance criteria`, `### Gotchas` — none contain `approach`/`design`/`implementation`, so `missing_core == ["Approach"]` (len 1 → warn branch, not hard-fail) and exit code is 0, matching the asserted `assert_exit $? 0`.
- Confirmed the guard is non-vacuous by hand: if `"plan"` were re-added to the `Approach` synonym tuple, the `## Plan: plan-heading only` heading would satisfy it, `missing_core` would go empty, and the assertion's `case` match on `"core section absent"*"Approach"*` in the warn output would fail — exactly the regression this fixture is meant to catch.
- No inline review comments on the PR to re-verify.

The second nit from the prior review ("`### Affected files` also satisfies Scope") was explicitly accepted as intentional — no action needed, correctly not re-addressed.

- [x] Already covered — `test_fleet_plan_lint.sh` 17/17 green.

Opus recheck not required.

Issue #2469, opened 2026-07-20
render: per-axis scatter sub-pixel centroid residual under yaw-sweep (voxel cylinder rev=2 @zoom4, 1.57px @zoom8 — survives #2427)


jakildev:
Split off from #2427. The #2427 overflow-lane fix (θ-stable 2×2 mask-neighborhood compare in `overflowAppendTap`) eliminates the **multi-pixel** face-pop jitter (x residual 2.93px→~1.0px, Δmax 5.37→~0.95 on the canonical voxel-cylinder `--yaw-sweep` probe). What remains is a **sub-pixel per-axis-scatter positioning wobble** that is independent of the overflow lane and keeps the jitter_probe verdict at JITTER by tiny margins at zoom 4 and 8.

- **Area:** engine/render
- **Model:** opus
- **Blocked by:** (none)  
- **Likely parent epic:** #1881 (rotated-voxel correctness under camera Z-yaw)

## Acceptance criteria

On the canonical Z-yaw-invariant probe, `jitter_probe` reports **SMOOTH** (reversals=0 both axes, max_residual ≤ 1.50px) at **zoom 2, 4, AND 8**:

```
IRShapeDebug --spin-shape cylinder --spin-shape-voxel --yaw-sweep --zoom <2|4|8> --auto-screenshot 6
build/tools/jitter_probe/jitter_probe <screenshots>/screenshot_0*.png
```

(zoom 2 already passes after #2427; the gap is zoom 4 rev=2 and zoom 8 resid=1.57px.) Cardinal byte-identity (img_diff=0 at yaw 0) and the pan-sweep SMOOTH result must be preserved.

## Context (measured on macOS/Metal on the #2465 base, during #2427)

Post-#2427 `jitter_probe` (voxel cylinder `--yaw-sweep`):

| zoom | x rev | x resid | x Δmax | y rev | y resid | verdict |
|---|---|---|---|---|---|---|
| 2 | 0 | 0.64px | 0.80 | 0 | 0.16px | SMOOTH |
| 4 | 2 | 1.01px | 0.95 | 0 | 0.31px | JITTER (2 sub-px reversals) |
| 8 | 0 | 1.57px | 1.61 | 0 | 0.49px | JITTER (resid 0.07px over 1.5) |

**Three independent proofs this residual is the per-axis scatter's sub-pixel positioning, NOT the overflow lane #2427 fixed:**
[...]

PR #2465, opened 2026-07-20
build: bump fmt to 11.2.0 + spdlog to v1.15.3 (Apple clang 21)

jakildev:
## Summary
- Bumps `fmt` 10.1.1 → 11.2.0 and `spdlog` v1.12.0 → v1.15.3 in `engine/profile/CMakeLists.txt` — a **required pair** — to restore fresh-build capability under Apple clang 21, which rejects fmt ≤10.2.1's consteval `FMT_STRING` at C++23 and dies in the bundled fmt lib before any engine source compiles.
- **Supersedes the issue's recommended fmt→10.2.1 bump.** On-host measurement (recorded in the committed plan) shows 10.2.1 fails the same way; 11.2.0 is the lowest version that compiles under Apple clang 21 at C++23, and spdlog 1.12 cannot build against external fmt 11 — hence the paired spdlog bump.
- First commit is the plan file `.fleet/plans/issue-2449.md` (#1932 flow). Only version pins in the tree are these two `GIT_TAG` lines; downstream creations carry none and inherit the engine's.

## Test plan
- [x] **Fresh macOS build** (wiped fmt/spdlog `_deps`, reconfigured `macos-debug`) recompiles fmt 11.2.0 + spdlog 1.15.3 from source and links `IRShapeDebug` — Apple clang 21.0.0, arm64. The build log crosses `[38%] Built target fmt` / `[42%] Built target spdlog` at the exact point fresh builds previously died in `fmt-src`, then reaches `[100%] Built target IRShapeDebug`. No `error:` / consteval lines.
- [x] `IRShapeDebug --auto-screenshot` exits 0 (`RESULT=CLEAN`), screenshot saved. Engine log lines format normally and colorize (`[2026-... ] [EngineLog] [info] ...`) — spdlog 1.12→1.15 sink/pattern behavior unchanged.
- [x] A dedicated downstream-creation `build-game` dir pointed at this engine worktree reconfigures, recompiles fmt/spdlog from source, and links a representative creation target clean (the path where this outage was first surfaced).
- [ ] gcc-13 / mingw non-regression — owed via the cross-host `fleet:needs-linux-smoke` flow (fleet is currently all-macOS; both versions' upstream CI covers gcc-13).

Closes #2449

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2471, opened 2026-07-20
render: θ-stable per-axis overflow membership kills voxel-cylinder yaw jitter (#2427)

### LINKED ISSUE #2427: render: voxel cylinder yaw-sweep centroid JITTER on master (reversals=9, 2.9px residual, macOS/Metal zoom4)

jakildev:
## Summary

Fixes the voxel-cylinder yaw-sweep centroid **jitter** #2427 reports (macOS/Metal, zoom 4: 9 reversals, 2.93px residual). Root cause: the per-axis **view-visibility overflow lane** (#2331) elected each face's append membership from the *single* `roundHalfUp` cell of its yawed position, so a face whose footprint straddles a cell boundary flips membership discretely as yaw steps across the boundary — popping a whole face quad frame-to-frame.

`overflowAppendTap` now compares the face's depth key against the **most permissive winner over the 2×2 cell neighborhood** spanning the unrounded yawed position (wrap-safe `key − eps` compare so the `0xFFFFFFFF` empty sentinel reads as permissive). The mask **write** side is unchanged, so the compare only ever admits a **superset** of the single-cell pass — the sanctioned over-emit direction (depth test cleans up over-emit; only under-emit re-opens #2331 holes). GLSL + Metal twins bit-exact.

## Verification (macOS/Metal, `jitter_probe` on the canonical Z-yaw-invariant probe)

| state | zoom | x rev | x resid | x Δmax | verdict |
|---|---|---|---|---|---|
| pre-fix | 4 | 9 | 2.93px | 5.37 | JITTER |
| **post-fix** | **2** | **0** | **0.64px** | **0.80** | **SMOOTH** |
| **post-fix** | **4** | **2** | **1.01px** | **0.95** | JITTER\* |
| **post-fix** | **8** | **0** | **1.57px** | **1.61** | JITTER\* |

- Multi-pixel face-pop signature **eliminated**: residual 2.93→1.01px (−66%), Δmax 5.37→0.95 (−82%) at zoom 4; y-axis clean at every zoom.
- Cardinal byte-identity **preserved** — every yaw-0/90/180/270 shot `img_diff`=0; only the two inter-cardinal (yaw-45) shots change (the overflow lane is gated off at `residualYaw==0`).
- Overflow lane still **active** (fills #2331 holes): post-fix vs `IR_PERAXIS_OVERFLOW_DISABLE=1` differs (drift 1432–3856px) — no under-emit.
- Pan-sweep twin stays **SMOOTH**; all runs `ir-run RESULT=CLEAN`; overflow-cap drop warning absent throughout.

## \*Acceptance scope — the residual is a **separate** defect (#2469)

The plan's `reversals=0` bar is not fully met at zoom 4/8, but that residual is **not** the overflow lane — proven three independent ways:
1. **Kill switch:** with the overflow lane fully OFF the cylinder still wobbles (rev=12, 0.51px @zoom4) — the residual exists *without* the lane.
2. **Span-invariance:** widening the compare 2×2→3×3 changes the metrics by **zero** at every zoom — so it is not overflow membership.
3. **Sub-pixel magnitude** (≤1.57px, Δmax≤1.61px) — the per-axis scatter's sub-pixel positioning wobble.

`reversals=0` at zoom 4/8 is therefore **unmeetable by any overflow-lane change**. It's split off as **#2469** (per-axis scatter sub-pixel positioning; may be accepted drift). This PR resolves the multi-pixel jitter #2427 reports.

## Backend note

Authored + **verified on macOS/Metal** — the issue's native backend, buildable here on the #2465 base. The GLSL twin is the bit-exact mirror but is **unverified on GL** (macOS can't build the GL backend); it needs a Linux/Windows smoke pass. The overflow-lane algorithm is backend-agnostic, so parity is by construction.

## Screenshots — yaw-45 (inter-cardinal), before/after

The pose where the overflow lane runs; cardinal poses are byte-identical, so the *temporal* jitter fix is shown by the `jitter_probe` table above rather than a still.

| before (pre-fix) | after (post-fix) | diff (drift=4976px) |
|---|---|---|
| before | after | diff |

## Test plan
- [x] macOS/Metal: `jitter_probe` zoom 2/4/8, pan-sweep, cardinal byte-identity, overflow-active, cap-drop — table above
- [ ] Linux/Windows GL smoke: `fleet-build --target IRShapeDebug` + the jitter_probe repro at zoom 2/4/8 (verifies the GLSL twin)

Closes #2427


jakildev (commented):
## Review — render: θ-stable per-axis overflow membership kills voxel-cylinder yaw jitter (#2427)

**Verdict:** approve

Stacked on #2465 (macOS clang-21 build-wall fix, currently `fleet:approved`, still open — required to build the Metal backend on the author's host). Reviewing this PR's own diff only.

`mergeable == MERGEABLE`, so no churn audit needed.

### What changed

`overflowAppendTap` (GLSL + Metal twins in `c_voxel_to_trixel_stage_1_body.{glsl,metal}`) replaces its single roundHalfUp-cell occlusion compare with a most-permissive (max) compare over the 2×2 cell neighborhood spanning the *unrounded* yawed position, fixing a discrete membership flip as a face's footprint straddles a cell boundary under yaw — the x-only multi-pixel jitter #2427 reports. The mask *write* side (`viewMaskTap`) is unchanged.

### Verification performed

I read both full files at the PR head and independently checked the load-bearing claims rather than trusting the inline comments:

- **Bias/wrap-safety.** `kOverflowDepthBias = 0x40000000` (`c_voxel_to_trixel_stage_1_body.glsl:244`) confirms `overflowYawedDepthKey` really is bias-centered where claimed, so `key - eps` cannot underflow. Confirmed the scratch region really is filled with the `0xFFFFFFFF` sentinel each rotating frame (`system_voxel_to_trixel.hpp:616-623`, `fillBuffer(..., 0xFFu)`), so an unwritten neighbor cell genuinely can produce `maxMaskKey == 0xFFFFFFFF` — with `key - eps ≈ 0x40000000-ish`, the unsigned compare `key - eps > maxMaskKey` correctly reads false (permissive), matching the claim. I also checked the pre-PR single-cell code could never have hit this sentinel case (its one queried cell is always the face's own self-tie cell, which `viewMaskTap` always writes for that same face) — so this wrap-safety handling is new surface area introduced by the neighborhood read, not an incidental fix to a preexisting latent bug.
- **Self-tie / write-read agreement.** `roundHalfUp(v) = floor(v + 0.5)` (`ir_iso_common.glsl:462-472`), which is provably `floor(v)` or `floor(v)+1` per axis — always inside the 2×2 block the new compare reads (`neighborhoodBase + {0,1}²` where `neighborhoodBase = floor(p)`). Both the write and the new read project through the identical `pos3DtoPos2DIsoYawed(vec3(facePos), visualYaw)` call, so the write cell is provably always inside the read neighborhood — the max() compare can only ever admit a superset of the old single-cell pass, the sanctioned over-emit direction.
- **GLSL/Metal parity.** Diffed the two new `overflowAppendTap` bodies line-by-line — loop bounds, `continue`/`return` placement, the `>` comparison, the `max()` accumulation, and the cap-hit rollback are semantically identical between languages (only syntax differs: `atomic_load_explicit` vs direct scratch read, etc.). No divergence.
- **Downstream cap.** The compare relaxation can now admit more faces into the *separate*, already-capped overflow-entries array (distinct from the `perAxisWinnerIds` mask scratch this diff touches) — the existing `overflowCap_` sizing and one-shot `warnOverflowDropsIfAny` CPU warn (both untouched by this PR) already guard that risk, and the PR body explicitly reports the drop warning stayed absent throughout its verification runs.
- Confirmed no C++ struct or binding changes were needed: `overflowScratchLayout` and the mask-region sizing (one uint per canvas pixel) are unaffected by reading 4 cells instead of 1 from the same already-sized region.

### Nits

- `c_voxel_to_trixel_stage_1_body.glsl:305-313` / `metal:162-176` — the compare now does up to 4 scratch reads instead of 1 per face in the `resolveMode == 3` overflow-append pass (one of several per-rotating-frame passes, not the primary store dispatch). Unlikely to matter, but worth a quick profile check if per-axis-rotation frame time regresses on a dense scene.

### Praise

- The write/compare self-tie property (`roundHalfUp(p)` always lands inside the `floor(p)` 2×2 neighborhood) is a real, non-obvious math guarantee, and the PR proves it holds rather than asserting it — the same rigor extends to the wrap-safety fix, which is scoped exactly to the new sentinel-exposure surface rather than bolted on generically.
- The plan doc's kill-switch / span-invariance / sub-pixel-magnitude triage (three independent proofs that the residual jitter at zoom 4/8 is a *separate* defect, #2469, not this lane) is the right rigor for not overclaiming the fix's scope.

- [ ] Linux/Windows GL smoke — this PR is verified on macOS/Metal only; the GLSL twin is unverified on GL (author's own checkbox, correctly left unchecked). Tagging `fleet:needs-windows-smoke` below since author is `fleet:authored-on-macos`.

Opus recheck required: touches engine/render/ (GPU compute shader occlusion invariant + GLSL/Metal bit-parity in a hot per-voxel-face kernel).

jakildev:
Merger: base PR #2465 merged. Re-targeted this PR from `claude/2449-fmt-spdlog-clang21` to `master` and rebased the child commits onto current master tip (`--force-with-lease`). The diff against master is now clean; any verdict label (`fleet:approved` / `fleet:needs-fix` / `fleet:has-nits`) from the pre-merge review has been preserved. CI will re-run.

— fleet merger

jakildev (commented):
## Opus recheck — render: θ-stable per-axis overflow membership (#2427)

**Verdict:** approve

Sonnet escalated this for GPU-compute-occlusion-invariant + GLSL/Metal bit-parity in a hot per-voxel-face kernel. Stack gate is clear: base PR #2465 merged to master at 20:10 and the merger re-targeted this PR to master (`MERGEABLE`), so I reviewed this diff against master directly.

I re-derived the four load-bearing invariants independently from the shader source rather than trusting the inline comments. **Sonnet's verification holds on every point** — I confirm each below and add the one thing an occlusion-invariant recheck must rule out that a first pass can gloss: whether the more-permissive compare can ever change the *visible* result rather than only the overflow-append set.

### Confirmed (Opus-only invariant depth)

- **Self-tie preserved under the neighborhood widening.** `overflowYawedPixel` writes at `perAxisBase + roundHalfUp(pos3DtoPos2DIsoYawed(facePos, visualYaw))`; the new compare reads `neighborhoodBase + {0,1}²` where `neighborhoodBase = perAxisBase + floor(sameProjection)`. `roundHalfUp(v) = floor(v + 0.5)` (`ir_iso_common.glsl:470`), which per-axis is exactly `floor(v)` or `floor(v)+1` — always inside the 2×2 block. Both sides project through the *identical* `pos3DtoPos2DIsoYawed` call, so the write cell is provably always one of the four read cells. `max()` therefore admits a strict superset of the old single-cell pass. Over-emit only.

- **2×2 is the exact straddle set — no off-by-one.** Since `p ≥ floor(p)` per axis, a footprint can straddle only into `floor(p)+1`, never `floor(p)-1`. `{floor(p), floor(p)+1}²` is the minimal-and-complete neighborhood the unrounded position can round into. Widening to 3×3 would be redundant (and the plan's own residual-jitter triage attributes the sub-pixel remainder to the #2469 scatter wobble, not this compare).

- **Wrap-safety is genuinely new surface, and handled correctly.** The old single-cell compare only ever queried the face's own self-tie cell, which `viewMaskTap` always wrote — so it never saw the `0xFFFFFFFF` empty sentinel. The neighborhood read exposes up to 3 unwritten cells. Rewriting `key > maskKey + eps` as `key - eps > maxMaskKey` is the correct fix: `overflowYawedDepthKey` is bias-centered at `0x40000000` (`:244`) so `key - 8u` cannot underflow for any physical yawed depth, and an empty-sentinel neighbor drives `maxMaskKey` to `0xFFFFFFFF`, making the unsigned compare read false → append. Treating a background-straddling footprint as silhouette-visible is the correct (#2331-hole-avoiding) direction.

- **Over-emit cannot corrupt the visible frame.** This is the invariant I most wanted to close: an admitted face still passes `imageLoad(cardPix).x == voxelDistance` exclusion (cardinal winners are drawn by the cell path, never appended) and a bounded `isInsideCanvas(cardPix)` check before any store, so appended faces are `viewVisible \ cardinalWinners` exactly as before — they compete in the framebuffer depth test and a genuinely-occluded one loses. The change relaxes *membership of the append candidate set*, not the depth arbitration, so the widened set can only add faces that then lose the depth test. That is why cardinal byte-identity (acceptance #5) is expected to hold even though the set changes.

- **GLSL/Metal bit-parity.** Diffed the twins statement-by-statement: loop bounds, `continue`/`return` placement, `max()` accumulation, `floor` truncation (`ivec2`/`int2` of a floor-valued float — exact both sides), and the final `key - eps > maxMaskKey` compare are identical. The only deltas are mandated backend idioms (`atomic_load_explicit(..., relaxed)` vs direct SSBO read, `frameData.` prefix), which mirror the pre-PR code. No divergence.

### On Sonnet's nit

Sonnet flagged up to 4 scratch reads per face vs 1 in the `resolveMode == 3` overflow-append pass. I confirm it's real but **not merge-blocking**: this is the overflow-append pass, not the primary store dispatch, and the empty-sentinel-permissive path additionally admits a band along every silhouette edge, so both the read count and the append volume rise. Both risks are (a) over-emit-safe and (b) already gated by the untouched `overflowCap_` sizing + one-shot drop warn, and any per-rotating-frame-time regression on a dense scene will surface in the already-tagged `fleet:needs-windows-smoke` run. No code change requested — dropping `fleet:has-nits`.

### Not verifiable on this host (correctly deferred, not a hold)

macOS/Metal build is walled by #2449, and this fix is GL-primary. The GLSL runtime gate (jitter_probe SMOOTH at zoom 2/4/8, cardinal img_diff=0, drop-warn absent) is the author's `fleet:needs-windows-smoke` checkbox — the right lane for it. The Metal twin lands by review parity, which I've confirmed above.

🤖 Opus recheck by Claude Opus 4.8 (review-pr / opus-reviewer role)

Issue #2298, opened 2026-07-08
render: extend per-voxel occlusion cull domain to the shadow-feeder-widened canvas (feeder-vs-feeder overdraw)


jakildev:
**Area:** engine/render
**Model:** opus
**Blocked by:** #1812

## Context

Carve-off from PR #2278 (issue #1812, per-voxel Hi-Z occlusion cull) per the architect direction there. The per-voxel cull is correct and captures ~2× the chunk cull at zoom1, but at `voxel_set` zoom8 it captures ~nothing (visible/total 0.8119 → 0.8117): the frustum there is dominated by **shadow-feeder casters outside the visible viewport**, which the visible-viewport-domain test must keep.

The 0.97 "ceiling" from `.fleet/plans/issue-1294.md` is a **stage-2 tap-overdraw fraction** measured shadows-on over the full widened canvas (18M taps / 472K wins) — it includes feeder-vs-feeder overdraw that no visible-domain test can reach. That share is this issue's target.

## Candidate mechanisms (hypotheses — planning must pick and validate)

- **Primary: widen the cull domain.** Extend the Hi-Z build + per-voxel test domain to the shadow-feeder-widened canvas region. Invariant argument: a voxel conservatively occluded (footprint-max) at every canvas texel it rasters to — including the widened region — leaves no trace in `trixelDistances`; both the visible resolve and the sun bake consume `trixelDistances`, not voxels, so dropping it is bit-identical for both. Costs to validate in planning: Hi-Z build over the larger region, the footprint-max dead zone at feeder scales, interaction with the camera-cut one-frame-disable gate.
- **Fallback: sun-space occlusion test** ("occluded from the sun ⇒ casts nothing") — a new mechanism; only worth planning if the widened-domain numbers show the remaining overdraw is cross-regime.

Planning needs first: the feeder-region share of zoom8 overdraw, and (if PR #2278 landed it) the `--no-sun-shadows` capture baseline.

## Acceptance criteria

- Planning gate: measured feeder-region share of `voxel_set` zoom8 overdraw; one mechanism picked with a confirmed cost model.
- Same invariants as #1812: bit-identical cull-on vs cull-off, off by default with zero disabled-path cost, no lighting/shadow invariant changes, backend parity (GL + Metal).
- Realized `voxelStage1` reduction on `voxel_set` zoom8 recorded; net-positive (or gated) across the zoom range.

## Pointers

[...]

PR #2475, opened 2026-07-20
render: widen per-voxel occlusion cull domain to shadow-feeder ring (#2298)

jakildev:
## Summary
- Replaces the #1812 `visibleIsoBounds` early-return in `voxelOccludedByHiZ` with a Hi-Z **canvas-coverage guard** (GLSL + Metal mirror), extending the per-voxel occlusion test from the visible viewport to the full shadow-feeder-widened canvas.
- Sound by the architect-endorsed trace invariant (#2278): a voxel occluded at every canvas texel it rasters to leaves no trace in `trixelDistances`, which both the visible resolve and the sun bake consume — so dropping it is bit-identical for its shadow too.
- Adds the **struct-1 feeder-count readback** the acceptance gates need (Feeder row in the profile report + `ir.render.getVoxelCullStats`): the #2258 Step-B feeder list — this PR's target population — previously had no CPU readback anywhere.

## Design-block resolution (was: captures zero)

The earlier zero-capture measurement is root-caused and does **not** falsify the mechanism. Two independent causes (full data in the `## DESIGN-UNBLOCKED` comment):

1. **Degenerate harness.** perf_grid's amp-0 face stamping (added Jul 15, after the plan's Jul 13 measurements) collapses the scene to a ~24K shell; the plan's dense regime is restored by `--wave-freeze --wave-amplitude 5` (visible chunk-only 166,868 ≈ the plan's 167,507).
2. **Metal-only data gap (#2488).** On Metal, feeder depth lives only in the #1640 image-atomic scratch; stage 2 (the sole texture writer) skips feeders per #1740, so the ring of the Hi-Z source texture is permanently sentinel — the widened test is a conservative no-op in the ring on Metal (98.4% of feeders sit on sentinel Hi-Z texels, probe-verified). On GL, stage 1 writes the texture directly and the ring data is real.

Net: this PR is byte-identical-safe on Metal (verified below) and its capture materializes on GL — where #2298 was already routed before a macOS pane picked it up. The ring positive-fire gate must run on a **Linux/GL host**.

## Test plan
- [x] Visible-domain marginal capture on the dense frozen harness (macOS/Metal): 234,589→28,818 (z4), 166,868→21,032 (z8), 67,342→14,615 (z16) — the relaxed gate doesn't regress the #1812 capture.
- [x] Marginal byte-identity, sun shadows ON (macOS/Metal): md5-identical pv-on vs pv-off on every cardinal `--auto-screenshot` shot (`fit_grid`, `zoom1_origin`, `profiler_overlay`, `zoom4_pan`) on `--wave-freeze --wave-amplitude 5`; yaw shots excluded (run-to-run drift with no change — #2479; cull disarmed at non-cardinal yaw).
- [x] Byte-identical where no feeder ring exists (shadows-off / no-ring pose: chunk+PV == chunk-only == 5013).
- [ ] **Ring positive-fire gate — Linux/GL host** (blocked on Metal by #2488): feeder count pv-on < pv-off with shadows ON on the dense harness, via the new Feeder profile row. One command per config, see the DESIGN-UNBLOCKED comment.
- [ ] GL parity smoke (Linux host).

Ref #2298

### DIFF
--- a/engine/prefabs/irreden/render/gpu_stage_timing.hpp
+++ b/engine/prefabs/irreden/render/gpu_stage_timing.hpp
@@ -47,6 +47,11 @@ struct GpuStageTiming {
     // VoxelCullAccumulator below.
     std::uint32_t visibleVoxelCount_ = 0;
     std::uint32_t totalVoxelCount_ = 0;
+    // Shadow-feeder (struct 1) survivors from the same prior-frame readback —
+    // the #2258 Step-B tail-append population the #2298 domain-widened cull
+    // targets. 0 whenever shadows are off / per-axis split active (feeders
+    // exist only on the single-canvas path).
+    std::uint32_t feederVoxelCount_ = 0;
     // Light-gather diagnostic. Populated by COMPUTE_LIGHT_VOLUME each
     // frame: `lightsSeeded_` = sources written to the seed SSBO (in-window
     // plus boundary-discounted), `lightsEligible_` = non-directional
@@ -138,23 +143,29 @@ inline ComputeLightVolumeTiming &computeLightVolumeTiming() {
 struct VoxelCullAccumulator {
     std::uint64_t visibleSum_ = 0;
     std::uint64_t totalSum_ = 0;
+    std::uint64_t feederSum_ = 0;
     std::uint32_t maxVisible_ = 0;
     std::uint32_t maxTotal_ = 0;
+    std::uint32_t maxFeeder_ = 0;
     std::uint32_t sampleCount_ = 0;
 
-    void record(std::uint32_t visible, std::uint32_t total) {
+    void record(std::uint32_t visible, std::uint32_t total, std::uint32_t feeder) {
         visibleSum_ += visible;
         totalSum_ += total;
+        feederSum_ += feeder;
         maxVisible_ = IRMath::max(maxVisible_, visible);
         maxTotal_ = IRMath::max(maxTotal_, total);
+        maxFeeder_ = IRMath::max(maxFeeder_, feeder);
         ++sampleCount_;
     }
 
     void reset() {
         visibleSum_ = 0;
         totalSum_ = 0;
+        feederSum_ = 0;
         maxVisible_ = 0;
         maxTotal_ = 0;
+        maxFeeder_ = 0;
         sampleCount_ = 0;
     }
 };
--- a/engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp
+++ b/engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp
@@ -1514,6 +1514,7 @@ template <> struct System<VOXEL_TO_TRIXEL_STAGE_1> {
         // overlay wants, not a strict voxel count comparable 1:1 to cardinal.
         if (gpuStageTiming().enabled_) {
             std::uint32_t visible = 0;
+            std::uint32_t feeder = 0;
             if (skipSingleCanvasVoxels && perAxisIndirectBuf_ != nullptr) {
                 for (int axis = 0; axis < C_PerAxisTrixelCanvases::kAxisCount; ++axis) {
                     VoxelIndirectDispatchParams region{};
@@ -1528,10 +1529,28 @@ template <> struct System<VOXEL_TO_TRIXEL_STAGE_1> {
                 VoxelIndirectDispatchParams previous{};
                 indirectBuf_->getSubData(0, sizeof(VoxelIndirectDispatchParams), &previous);
                 visible = previous.visibleCount;
+                // Struct 1 (the #2258 Step-B shadow-feeder tail list) shares
+                // indirectBuf_ at kPerAxisSsboAlignBytes and follows the same
+                // prior-frame-before-zeroing contract, so its count rides the
+                // same sync-free read. Single-canvas path only — the per-axis
+                // split never populates it. This is the #2298 target
+                // population; the acceptance gates diff it pv-on vs pv-off.
+                VoxelIndirectDispatchParams previousFeeder{};
+                indirectBuf_->getSubData(
+                    static_cast<std::ptrdiff_t>(kPerAxisSsboAlignBytes),
+                    sizeof(VoxelIndirectDispatchParams),
+                    &previousFeeder
+                );
+                feeder = previousFeeder.visibleCount;
             }
             gpuStageTiming().visibleVoxelCount_ = visible;
             gpuStageTiming().totalVoxelCount_ = static_cast<std::uint32_t>(effectiveVoxelCount);
-            voxelCullAccumulator().record(visible, static_cast<std::uint32_t>(effectiveVoxelCount));
+            gpuStageTiming().feederVoxelCount_ = feeder;
+            voxelCullAccumulator().record(
+                visible,
+                static_cast<std::uint32_t>(effectiveVoxelCount),
+                feeder
+            );
         }
 
         const VoxelIndirectDispatchParams zeroed{};
--- a/engine/profile/include/irreden/profile/profile_report.hpp
+++ b/engine/profile/include/irreden/profile/profile_report.hpp
@@ -42,8 +42,12 @@ struct CpuPhaseEntry {
 struct VoxelCullStatsSummary {
     uint64_t visibleSum_ = 0;
     uint64_t totalSum_ = 0;
+    // Shadow-feeder (struct 1) survivors — the off-screen caster list the
+    // #2298 domain-widened per-voxel cull targets. 0 with shadows off.
+    uint64_t feederSum_ = 0;
     uint32_t maxVisible_ = 0;
     uint32_t maxTotal_ = 0;
+    uint32_t maxFeeder_ = 0;
     uint32_t sampleCount_ = 0;
 };
 
--- a/engine/profile/src/profile_report.cpp
+++ b/engine/profile/src/profile_report.cpp
@@ -201,6 +201,8 @@ void writeProfileReport(const ProfileReport &report, const char *outputPath) {
             c.sampleCount_
         );
         std::fprintf(f, "%-12s %14.1f %14u %10u\n", "Total", avgTotal, c.maxTotal_, c.sampleCount_);
+        const double avgFeeder = static_cast<double>(c.feederSum_) / samples;
+        std::fprintf(f, "%-12s %14.1f %14u %10u\n", "Feeder", avgFeeder, c.maxFeeder_, c.sampleCount_);
         std::fprintf(
             f,
             "Ratio:       %14.4f (visible/total — 1.0 = no cull, 0.0 = all culled)\n",
--- a/engine/render/src/shaders/c_voxel_visibility_compact.glsl
+++ b/engine/render/src/shaders/c_voxel_visibility_compact.glsl
@@ -35,14 +35,13 @@ layout(std140, binding = 7) uniform FrameDataVoxelToTrixel {
     uniform ivec4 visibleFaceIds;       // offset 128
     uniform vec4 voxelDepthAxis;        // offset 144
     uniform vec4 detachedWorldReceive;  // offset 160
-    // Un-widened (no shadow-feeder sweep) visible iso viewport (#1740). Two
-    // consumers here: the per-voxel occlusion test only culls voxels fully
-    // inside this box — a caster in the shadow-feeder swept ring (inside
-    // cullIsoMin/Max but outside here) must keep its sun shadow (mirrors
-    // dispatchChunkOcclusion's fully-visible eligibility) — and the #2258
-    // Step-B classify routes a survivor OUTSIDE this box (an off-screen
-    // feeder, the exact stage-2 #1740 skip convention) to the feeder dispatch
-    // struct instead of the full-density visible list.
+    // Un-widened (no shadow-feeder sweep) visible iso viewport (#1740).
+    // Consumer here: the #2258 Step-B classify routes a survivor OUTSIDE this
+    // box (an off-screen feeder, the exact stage-2 #1740 skip convention) to
+    // the feeder dispatch struct instead of the full-density visible list. The
+    // per-voxel occlusion cull no longer gates on this box as of #2298 — it now
+    // tests the full shadow-feeder-widened canvas via a Hi-Z canvas-COVERAGE
+    // guard (see voxelOccludedByHiZ); visibleIsoBounds must stay for the classify.
     uniform ivec4 visibleIsoBounds;     // offset 176
     uniform int resolveMode;            // offset 192
     // Per-voxel Hi-Z occlusion-cull gate (#1812): 0 = off (byte-identical),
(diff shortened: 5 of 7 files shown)


jakildev:
## NEEDS-DESIGN

The #2298 mechanism — widen the per-voxel Hi-Z occlusion test from the visible
viewport to the shadow-feeder ring — is implemented faithfully in this PR (GLSL
+ Metal, the canvas-coverage guard the plan specified, no new binding: reads
`imageSize(hiZLevel0)` shader-side). But **it captures zero occludable feeders
on current master**, so as-is it is net-negative and must not merge.

### What I measured (macOS/Metal; build wall lifted — #2465 merged 20:10Z)

`IRPerfGrid --mode voxel_set --no-overlay --auto-profile --subdivision-mode none
--wave-amplitude 0 --occlusion-cull --zoom 8`, instrumented to read the **feeder
dispatch (struct 1) count** — the population #2298 targets — since the "Visible"
cull stat only reads struct 0 (the visible list):

| config | feeder (chunk+PV, this PR) | feeder (chunk-only) | culled |
|---|---|---|---|
| grid64 zoom8  | 6917 | 6917 | **0** |
| grid64 zoom16 | 7910 | 7910 | **0** |
| grid96 zoom8  | 6917 | 6917 | **0** |
| grid64 zoom4  |  993 |  993 | **0** |

The per-voxel test *does* cull visible voxels (5016 vs 9036 at zoom8), so
`hiZLevel0` is not corrupted (#2360 masking ruled out) and the guard is not
over-rejecting: I confirmed the feeders map to canvas px `x∈[276,438]
y∈[184,311]`, well inside the 642×722 canvas / Hi-Z extent, so they pass the
coverage guard and are tested. **They are simply never occluded** on this scene.

### Why the plan's numbers don't reproduce

The plan's gate (shadows-ON chunk-only 213494 / chunk+pv 64737 → a 45987-voxel
occludable ring = "71% of remaining survivors") was measured 2026-07-13 **before
#2325 (#2258 Step B) merged the same day**. Post-#2325, feeder-ring survivors
tail-append to struct 1 (the separate feeder dispatch), out of the "Visible"
stat. The current voxel_set scene also tops out at ~19370 visible (at any zoom)
vs the plan's 167507 — ~10× sparser than the plan measured against.

### The ruling I need

The plan itself said: *"If realized in-ring capture comes in far below the
visible-domain rate, the residual becomes a new measurement-first issue, not
mid-PR improvisation."* In-ring capture is **zero**, far below even that. So:

1. **Descope?** If the feeder ring has no Hi-Z-occludable population (surface
   feeders see background; interior feeders are already dropped by the
   face-occlusion mask before the test reaches them), the domain-widening cannot
   help and #2298 should close.
2. **Sun-space fallback?** The plan explicitly deferred the "occluded from the
   sun ⇒ casts nothing" mechanism. If the residual feeder overdraw is real but
   cross-regime (not camera-Hi-Z-reachable), that fallback may be the only
   mechanism that reaches it — a re-plan.
3. **Re-measure first?** The plan's dense scene (167507 visible) is gone; a valid
   harness (denser/deeper scene, or a GL host) is needed to confirm whether any
   occludable feeder population exists at all before committing to a mechanism.

I can't answer which without a design/scope call. Holding `fleet:design-blocked`
+ `fleet:wip`; releasing the claim. The implementation + measurement approach are
here for whoever resumes. — worker (opus)

jakildev:
## DESIGN-UNBLOCKED

Human cued unblocking; ran the measurement-first path the plan itself prescribed. **Ruling: none of the three escalation options apply as posed — the mechanism is not falsified.** The zero was over-determined by (1) a degenerate harness and (2) a Metal-only data gap, both now isolated with evidence. Work resumes; the remaining gates are GL-host work.

### 1. The harness: amp-0 no longer measures the plan's regime

perf_grid gained cross-neighbor face stamping for `alwaysTouching` scenes (Jul 15) **after** the plan's Jul 13 gate table. At `--wave-amplitude 0` the 64³ cube now rasters as a ~24K shell whose ring survivors are front-surface voxels — near-zero occludable population. `--wave-freeze --wave-amplitude 5` disables the stamping and restores the plan's dense regime exactly (visible chunk-only 166,868 vs the plan's 167,507).

### 2. The backend: ring Hi-Z is structurally sentinel on Metal (#2488)

On the dense harness, capture at every zoom (300-frame avgs, new Feeder readback, this branch @ 5d2b5833):

| zoom | visible pv-off → pv-on (capture) | feeder pv-off → pv-on (ring capture) |
|---|---|---|
| 4 | 234,589 → 28,818 (**88%**) | 4,094.3 → 4,094.3 (**0**) |
| 8 | 166,868 → 21,032 (**87%**) | 47,835.0 → 47,835.0 (**0**) |
| 16 | 67,342 → 14,615 (**78%**) | 83,642.3 → 83,642.3 (**0**) |

An exact zero next to 87% visible capture on the same scene is not a scene property. Staged-shader probes (Metal, uncommitted):
- **Force-cull all ring voxels** → feeder avg 47,835 → 160: the classify + readback chain reports capture when it exists.
- **Center-texel-only ring test** → still ~zero capture: the ring's Hi-Z content itself is empty.
- **Sentinel counter** (cull ring voxels whose center Hi-Z texel ≥ 65535) → drops 47,054 of 47,835 (**98.4%**): the ring of the Hi-Z source texture is background sentinel.

Root cause (#2488): Metal stage-1 depth atomics land in the #1640 image-atomic scratch; the only writer of the `trixelDistances` **texture** is stage 2's winner tap, which **skips feeders** (#1740). `c_build_distance_hiz` reads the texture → sentinel ring → the widened test can never fire on Metal. On **GL**, stage 1 `imageAtomicMin`s the texture directly, ring included — the mechanism's premise holds there. (The bake does see feeder depth on Metal — force-culling feeders changes interior shadow pixels — so off-screen casters still shadow; the gap is Hi-Z-specific.)

### Gates run this session (macOS/Metal)

- **Marginal byte-identity, shadows ON, dense frozen scene:** md5-identical pv-on vs pv-off on all four cardinal shots while culling ~145K voxels. Yaw shots drift run-to-run with no change (#2479) and the cull is disarmed at non-cardinal yaw — excluded.
- Instrumentation validated against the escalation's own numbers (amp-0 feeder 6,917, no-ring visible 5,013 — both reproduced exactly).

### What remains (Linux/GL host — one command per config)

```
fleet-run IRPerfGrid --mode voxel_set --no-overlay --auto-profile --subdivision-mode none \
  --wave-freeze --wave-amplitude 5 --occlusion-cull --zoom {4|8|16} [--no-per-voxel-occlusion] [--no-sun-shadows]
```
then read the `Feeder` row in `save_files/profile_report.txt`:
1. **Ring positive-fire gate:** feeder pv-on < pv-off with shadows ON at z4/8/16; record the table in the PR body.
2. Shadows-ON md5 A/B (cardinal shots) on GL — the load-bearing identity gate now that the ring actually participates there.
3. While there: check the #2488 "secondary observation" (viewport-edge pixels under feeder force-cull) to see if the edge election artifact is backend-agnostic.

Labels: `fleet:design-blocked` → `fleet:design-unblocked`, keeping `fleet:wip` + adding `fleet:needs-linux-smoke` for routing. Branch rebased onto master (picks up #2478's deterministic displaced-voxel taps, which the identity gates need on this scene). Follow-up filed: #2488.

— fable (human-cued unblock session)

jakildev:
Resumed this fleet:design-unblocked PR on a native-Windows (mingw64, OpenGL) pane to run the remaining GL positive-fire gate. Windows cannot run it right now: every demo linking the IrredenEngine aggregate fails at final link with undefined references to IRWorld::saveWorld/loadWorld (the Scripting/World static-lib cycle introduced by the persist Lua bindings at c08bf8d3). Reproduced on both IRPerfGrid and IRShapeDebug, so it is unrelated to this PR (render occlusion only). Filed separately: #2499.

The remaining gate therefore genuinely needs the Linux/GL host the DESIGN-UNBLOCKED comment already routes to (fleet:needs-linux-smoke). Releasing the amending claim without pushing; PR stays fleet:design-unblocked for a Linux pane to finish. -- worker (opus, Windows)

jakildev:
macOS/Metal opus pane picked this up via the `fleet:design-unblocked` tier — confirming there is **no macOS-runnable work remaining**. The DESIGN-UNBLOCKED ruling already established (with staged-shader evidence, #2488) that the ring Hi-Z is structurally sentinel on Metal, so the positive-fire gate reads a hard zero on this backend regardless of scene; and macOS cannot build the GL backend. The fable session already ran every macOS-side gate (byte-identity, sentinel probes). So the three remaining gates genuinely need a **GL host**, as routed.

One new fact that widens the finisher set: the native-Windows blocker from the 06:12 comment — #2499 (Scripting/World static-lib cycle) — is now **fixed and merged** (#2501, on master as `aa65cb49`). Master also picked up #2503 (NVIDIA GLSL rejections). So a **native-Windows mingw64/OpenGL pane can now run the remaining GL gate too**, not only a Linux pane.

Leaving `fleet:design-unblocked` + `fleet:wip` + `fleet:needs-linux-smoke` intact for a GL opus finisher — the ring positive-fire gate + shadows-ON md5 A/B + the #2488 secondary-observation check all carry the opus-tier descope-or-keep judgment, so the design-unblocked routing to an opus pane on a GL host is correct. Releasing the amending claim without pushing. — worker (opus, macOS)

jakildev:
Cross-reference from #2488 (PR #2898, open): once the Metal image-atomic scratch resolve lands, this PR's "byte-identical-safe on Metal, captures nothing" claim stops holding. `COMPUTE_DISTANCE_HIZ` reads `trixelDistances` as a plain texture, and on Metal the feeder ring was permanently the 65535 sentinel there — which is why the widened per-voxel cull measured 0 ring capture at every zoom. With the ring materialized, the widened cull sees real Hi-Z on Metal and its Metal-side gate numbers need re-measuring on the resolved baseline rather than carried over. No action needed while this PR is parked on #2385. — worker

jakildev:
Resumed this `fleet:design-unblocked` PR on a native-Windows (mingw64, OpenGL / NVIDIA) pane and ran all three remaining GL-host gates. **The mechanism fires on GL.**

**1. Ring positive-fire gate — PASS at every zoom.** Dense frozen harness, shadows ON, 300 frames/config:

| zoom | feeder pv-off → pv-on | ring capture | (Metal, for contrast) |
|---|---|---|---|
| 4  | 4,108 → 1,136.0 | **72%** | 0 |
| 8  | 47,995 → 8,595.2 | **82%** | 0 |
| 16 | 83,922 → 12,135.9 | **86%** | 0 |

The pv-off baselines land within ~0.4% of the macOS run (166,868 / 47,835 at zoom 8), so both hosts measure the same scene and only the ring capture differs — which is exactly the #2488 prediction.

**2. Shadows-ON md5 A/B — PASS.** All four cardinal shots byte-identical pv-on vs pv-off while culling ~82% of the ring. I did **not** inherit the "yaw shots drift run-to-run" exclusion — #2850 landed on master since that claim was written, so I re-derived it here: three runs of the *identical* config gave three distinct hashes for `zoom4_rot` and `zoom4_rot_pan`, while every cardinal shot was stable across all five runs. The exclusion holds, and #2850 does not make this scene deterministic at zoom 4 (consistent with its own documented residual class for pools without `storeTiesPossible_`).

**3. #2488 secondary observation — reproduces on GL, filed as #3010.** Force-culling *every* ring feeder changes 4,128 px on `zoom4_pan` (macOS measured ~4,300), 52% within 16 px of the frame edge. Backend-agnostic, so per #2488's own instruction it gets its own issue. Not this PR's doing: the ordinary pv-on/pv-off A/B leaves that shot byte-identical, since this cull only drops provably-occluded feeders.

**One new commit.** The Feeder readback needed a first-frame guard before any of the above was trustworthy: `indirectBuf_` has no prior-frame value on tick 0 and does not read back as zero on GL/NVIDIA. Struct 0 reported 6,357,111 "visible" against a 262,144 total; struct 1 reported 4,294,967,295 and then 7,536,749 on two runs of the same config. One junk sample was enough to report the pv-off feeder average as 14,364,392.7 instead of 47,995. With the guard every config reports `avg == max` over 299 samples and the tables reproduce run to run.

**Two things the reviewer should know:**

- **These runs required PR #3005 applied locally.** `c_per_axis_overflow_sort.glsl` uses `active`, a GLSL reserved word NVIDIA rejects, so every GL demo crashes at pipeline creation on current master. I deliberately did **not** fold that one-word rename into this branch — it is an unrelated shader and #3005 (open, `fleet:approved`) is the right vehicle. Anyone re-running these gates on an NVIDIA host before #3005 merges needs it applied.
- **The Metal-side claims in the older body were stale and I rewrote them.** #2488 closed and its fix #2898 merged 2026-08-08, so the ring is no longer structurally sentinel on Metal — the "byte-identical-safe on Metal, captures nothing" framing held only against the pre-#2898 baseline. Metal now needs a re-measure on the resolved baseline; that is macOS work this PR does not carry, and it is the one gap left before merge.

Dropped `fleet:wip` (it is a reviewer-skip label and the routed work is done) and `fleet:needs-gl-host` (the GL-only residual is complete). Left `fleet:needs-linux-smoke` — I am on Windows, not Linux, and each host clears its own. Releasing the amending claim. — worker (opus, Windows)

jakildev:
Addressed the two Opus-recheck nits:
- `c_voxel_visibility_compact.glsl` + Metal twin: the SOUNDNESS comment now states the bit-identity claim holds against the frame the Hi-Z was built from, and notes a stale-lag false cull costs a feeder its shadow for one frame (same class as #1812's one-frame hole).
- `profile_report.cpp`: the Feeder row now carries a note that its pv-on/pv-off ratio is cross-run (computed by hand from two separate `--occlusion-cull` invocations), unlike the intra-run visible/total `Ratio:` line above it.

The Sonnet-review test-coverage nit on `world.cpp`'s `getVoxelCullStats` binding is explicitly framed by the reviewer as a follow-up ("worth a follow-up test... if anyone revisits"), not a fix for this PR, so left as-is.

Build verified clean (`fleet-build --target IRShapeDebug`). Comment-only shader edits + a report-text addition, no runtime rendering effect, so render-debug-loop was skipped per `engine/render/CLAUDE.md`'s exception for provably-no-runtime-effect changes. — worker (sonnet)

jakildev:
Picked this up on the `fleet:has-nits` tier and found the label **stale, not actionable** — both Opus-recheck nits are already fixed at head `d2951010`; the 19:05Z fixing pass just never cleared the verdict label (step b of the feedback protocol). Verified against the tree rather than taken from the comment:

- `c_voxel_visibility_compact.glsl:254-259` — the `SOUNDNESS:` block now reads "…bit-identical for the shadow it would have cast too, **against the frame the Hi-Z was built from; a stale-lag false cull costs a feeder its shadow for one frame, same class as #1812's one-frame hole**." That is the clause the recheck asked for.
- `metal/c_voxel_visibility_compact.metal:147-156` — twin carries the same caveat, in lockstep.
- `profile_report.cpp:211-216` — the cross-run note prints inside the `voxelCullStats_.sampleCount_ > 0` block, directly under the `Ratio:` line, so it appears only where the `Feeder` row itself does. Correctly scoped.

The Sonnet `getVoxelCullStats` test-coverage nit was explicitly framed by its own reviewer as a follow-up ("worth a follow-up test... if anyone revisits"), not a fix for this PR — agreed, leaving it.

Cleared `fleet:has-nits`. No commits pushed (nothing to change). Left `fleet:needs-macos-smoke` and `fleet:needs-linux-smoke` intact — I am on Windows/GL, and the macOS tier is the one that actually gates merge here (the post-#2898 Metal re-measure the recheck flagged). `fleet:reviewing-windows-pool-8` is live, so a re-review of this head is already in flight. Releasing the amending claim.

— worker (opus, Windows)

jakildev:
Addressed the opus re-review nits (head `3cd580d8`, fast-forward on `d2951010`):

- **Nit 1 — `docs/perf/README.md` stale on the widened surface.** The Lua key list now reads `{visible, total, feeder, samples, avgVisible, avgTotal, avgFeeder, maxVisible, maxTotal, maxFeeder}` (verified against `world.cpp:180-192`), and the "Voxel cull stats" section gains a `Feeder` bullet: what the row counts (the struct-1 off-screen shadow-feeder population), that it has no in-report denominator, and the cross-run recipe — `--occlusion-cull` (pv-on) vs `--occlusion-cull --no-per-voxel-occlusion` (pv-off), pv-on Avg / pv-off Avg. The caveat now lives in the doc consulted before the run, not only in the report output.
- **Nit 2 — `compare_perf_runs.py` had no `CULL_FEEDER_RE`.** Took it in-PR rather than as a `fleet:nit-of-pr` follow-up (small, and it is the place the cross-run ratio actually belongs — the comparator is where two runs meet). Added the regex (`^`-anchored like its siblings, so the free-form `Feeder ratio:` note and the `Ratio:` line fall through it — checked), `avg_feeder` / `max_feeder` / `has_feeder` on `CullStats`, and an `avg feeder` column on the voxel-cull-effectiveness table carrying the `head / base` factor — a pv-off baseline dir vs a pv-on head dir now prints the gate number directly (`47835 → 8600 (0.180×)`). A report predating the row prints `—`, distinct from a measured `0` (shadows off). `perf_summary.py` / `check_regression.py` / `lua_cpp_parity.py` import the parser and are unaffected; `check_regression.py` prints the markdown verbatim and no CI step parses the table header textually.

Verification: `py_compile` on the three perf scripts clean; a synthetic two-run comparison exercised Feeder present/present (ratio factor), legacy/present (`— → N`), legacy/legacy (`—`), and zero/present (`0 → N`, no factor) — all as described. No C++ changed; no build. `ruff` is permission-gated on this headless pane, so it was not run.

One simplify note for the reviewer: `format_feeder_cell` shares the absent-sentinel/ternary/join shape with the inline GPU-stage cell loop in `render_markdown` (≈70% structural overlap per the call-sequence scan). Left separate on purpose — the two differ in absent handling (per-side `—` vs `0.0`), number format, and suffix (ratio factor vs pct mark), so a shared helper would carry three mode parameters for two callers.

Both smoke tiers (`fleet:needs-linux-smoke`, `fleet:needs-macos-smoke`) untouched — this delta has no runtime effect. — worker (fable, windows pool-3)

PR #2466, opened 2026-07-20
test: refresh shape_debug macos-debug render-verify references (#2426)

jakildev:
## Summary
- Refresh the `shape_debug` **macos-debug/Metal** render-verify reference set (21 PNGs). The set was stale on master since the #2410 (per-axis frac lighting recovery + occupancy stamp) and #2418 (wFrac carrier) merges, which intentionally changed rotated/lit output.
- Baseline on this host reproduced the issue's **21/24 render-verify FAIL**. The failing shots are intentional lighting drift, not a regression: shape geometry/silhouette is unchanged; only surface shading and platform AO/shadow shifted. Verified by comparing new captures against the old references and the per-shot diff images (diffuse, surface-following brightness deltas — no hard-edged geometry differences), consistent with #2410/#2418 and the issue's own non-regression proof.
- After `render-verify.py --target IRShapeDebug --update-references`, a fresh verify run passes **all 24 checks**.

## Build / verify
- Built on **macos-debug (Metal)** stacked on #2465 so fmt 11.2.0 + spdlog 1.15.3 compile under Apple clang 21 (the #2449 build wall). fmt/spdlog recompiled from source; `IRShapeDebug` links and runs `--auto-screenshot` to exit 0.
- `python3 scripts/render-verify.py --target IRShapeDebug` → `all 24 checks PASS`.

## Test plan
- [x] render-verify baseline reproduces 21/24 FAIL (matches the issue)
- [x] new captures visually verified as intentional lighting/rotation drift, not a regression
- [x] render-verify after refresh: 24/24 PASS

Closes #2426

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2478, opened 2026-07-20
render: cardinal store winner election — deterministic color/entity-id taps for displaced voxels (#2346)

### LINKED ISSUE #2346: render: cardinal (single-canvas) color/AO tap is run-to-run non-deterministic once voxel positions are perturbed
**Area:** engine/render
**Model:** fable
**Blocked by:** (none)

## Context

Found while building the #2332 (epic #2314 V1) wave-freeze validation
harness for IRPerfGrid. #2255 (PR #2272) fixed a non-atomic `writeColorTap`
tie-race on the **per-axis** voxel-to-trixel store — two faces tying on the
same encoded depth key raced on which one's color/entity-id write won,
because the guard was a non-atomic read-compare-write. That fix explicitly
left the **cardinal (single-canvas)** store untouched, on the stated
assumption that a dense integer-aligned lattice makes `(iso-pixel, iso-depth)`
a bijection of `(x,y,z)`, so cardinal ties can't happen (confirmed correct in
PR #2272's own before/after evidence: cardinal shots were byte-identical both
pre- and post-fix).

That assumption breaks once individual voxels are displaced off the
1-unit lattice (still each occupying an integer-ish neighborhood, but no
longer strictly bijective against a plain dense cube) — which is exactly
what a per-cell "frozen wave" scene does (see #2332): 262k independently
z-shifted 1×1×1 `C_VoxelSetNew` entities, all six faces reported exposed
(no cross-set occlusion stamping outside the Rigid+spacing==1 case).

## Confirmed repro (macOS/Metal)

```
fleet-run IRPerfGrid --mode voxel_set --wave-freeze --no-overlay --grid-size 32 --zoom 0.8 --auto-screenshot 80
```
run twice, `img_diff` the cardinal-yaw shots (`fit_grid`, `zoom1_origin`,
`profiler_overlay`, `zoom4_pan` — yaw 0 only; the rotated/per-axis shots
`zoom1_rot`/`zoom4_rot`/`zoom4_rot_pan` stayed byte-identical every time,
consistent with #2255/PR #2272 already covering that path):

| Wave amplitude | Cardinal shot drift |
|---|---|
| 0 (flat integer lattice, no shift) | byte-identical, every run |
| 1 | ~0.08–1.1% pixels drift, max per-channel delta ~2–3 |
| 5 (default) | ~0.28–4.7% pixels drift, max per-channel delta ~10–14 |

Drift scales with amplitude (more per-cell displacement → more colliding/
tied depth-encode keys), and only appears on the cardinal path — strong
evidence this is the same `writeColorTap`-style tie-race #2255 fixed,
recurring on the store #2255's fix didn't touch. Quantizing the baked
offset to the integer lattice via `IRMath::roundVec3HalfUp` did **not**
remove the drift (ruled out "fractional key" as the cause; the real
mechanism is almost certainly genuine position ties/collisions between
independently-displaced voxels, not fractional-vs-integer encoding).

## Why this hasn't surfaced before

Every existing cardinal-path byte-identity check (#2255's own acceptance
criteria, render-verify references, etc.) runs against either a static
*undisplaced* dense/hollow lattice or an *animating* wave (never checked for
byte-identity, since it's expected to change every frame). #2332 is the
first harness to freeze a *displaced* per-cell scene at a fixed pose and
diff repeat captures, which is what exposes this.

## Impact

- Blocks strict byte-identical determinism for any static, per-cell-displaced
  voxel scene at cardinal yaw (the #2332 wave-freeze harness works around
  this by gating on the scored coverage/hole-ratio/perimeter metrics instead
  of raw pixels — the drift's max per-channel delta stays under
  `render_metric_util`'s `bg_tol=16` foreground/background classification
  threshold, so those metrics are unaffected and exactly reproducible).
- Any future harness expecting byte-identical cardinal captures of a
  displaced-voxel scene will hit the same wall.

## Acceptance criteria

- `fleet-run IRPerfGrid --mode voxel_set --wave-freeze --no-overlay --grid-size 32 --zoom 0.8 --auto-screenshot 80` run ≥10× produces byte-identical cardinal-yaw shots (`fit_grid`, `zoom1_origin`, `profiler_overlay`, `zoom4_pan`) at `--wave-amplitude` values that currently drift (1 and 5 at minimum).
- Rotated/per-axis shots and existing #2255 acceptance criteria remain unaffected (no regression on the fix PR #2272 shipped).
- Flagless dense/hollow-mode runs stay byte-identical to current master.

## Suggested first step

Audit the cardinal store's equivalent of `writeColorTap` (single-canvas
`c_voxel_to_trixel_stage_2`/stage_1 path) for the same non-atomic
read-compare-write guard shape #2255's `PLAN`/PR #2272 diagnosed and fixed
on the per-axis path, and consider extending #2272's two-pass winner-election
scratch to the cardinal store too.

jakildev:
Implements the plan-review-cleared `## Plan` on #2346 (first commit of this
PR): extends #2272's two-pass winner election to the **cardinal
(single-canvas) store**, whose `(iso pixel, encoded depth)` key stops being a
bijection of live voxels once independently displaced voxels round into the
same integer cell — the recurrence of #2255's `writeColorTap` tie-race on the
store its fix deliberately left exempt.

## Mechanism

- **Per-pool signal.** `C_VoxelPool::storeTiesPossible_`, recomputed by
  `recomputeStoreTiesPossible` on frames whose CPU position upload changed
  binding 5 (pending-range flush or canvas-switch re-seed; static scenes scan
  once at seed). TRUE on the first active off-lattice voxel (>1e-4, the
  `snapNearIntegerVoxelPosition` epsilon), any GPU-transformed slot
  (world position authored GPU-side — conservative TRUE), or — all-integer
  fallback — two active voxels sharing a `roundHalfUp` cell (sort-based scan,
  reused scratch).
- **Compile-time shader variants** (the #2258 a′ idiom — no runtime
  predication on the hottest kernel pair): stage 2 splits into
  `_body` + thin default wrapper exactly like stage 1;
  `c_voxel_to_trixel_stage_1_winner_resolve` (ELECTION 1) re-runs the
  identical cardinal geometry with every distance tap swapped for
  `resolveWinnerTap` (atomicMin of the run-stable
  `compactedVoxelIndices[compactedIdx]`), and
  `c_voxel_to_trixel_stage_2_winner` guards every cardinal colour/entity-id
  tap with `winner[cell] == voxelIndex`. Metal twins line-for-line; both new
  kernels registered in `threadgroupSizeForFunctionName` (2,3,8) +
  `functionUsesImageAtomicScratch`.
- **Winner buffer.** System-owned, grow-only, lazily allocated on the first
  flagged canvas (lattice scenes allocate nothing), transiently bound at
  `kBufferIndex_PerAxisResolveScratch` (28) around the election + stage-2
  dispatches, placeholder restored after. Not `axes.winnerIds_` (rotation-
  lifecycle, #2412 class).
- **Dispatch wiring.** Flagged pools only: `fillBuffer(0xFF)` → election over
  indirect **struct 0 only** (stage 2 itself dispatches only struct 0;
  feeder-won pixels are never colour-tapped per #1740's margin) →
  SHADER_STORAGE barrier → winner-guarded stage 2 in place of the default.
  Unflagged pools run exactly master's programs and dispatch count.

## Phase-0 probe (plan gate) — premise CONFIRMED

CPU replica of the wave-freeze bake (float32, glm-exact easing):

| amplitude | colliding cells | voxels in collisions | max occupancy |
|---|---|---|---|
| 0 | 0 | 0 | 1 |
| 1 | 3,963 | 7,926 | 2 |
| 5 | 8,011 | 21,195 | 4 |

Zero at amp 0, monotone with amplitude — tracks the issue's drift table.

## Verification (macOS/Metal, all runs `ir-run: RESULT=CLEAN`)

- **Pre-fix positive fire (same-binary A/B).** Staged-shader defuse of the
  winner guard (ELECTION 1→0 in the staged wrapper = master's stage-2
  semantics): all 4 cardinal shots drift on **every** pair of 3 runs at
  amplitude 5 — `fit_grid` 0.283%, `zoom4_pan` 4.578% pixels, max per-channel
  delta 10, matching the issue's measured matrix (~0.28–4.7%, delta ~10–14).
  Guard restored: byte-identical. The election dispatch ran in BOTH configs,
  so the determinism is the guard's doing, not a scheduling accident.
- **Issue repro fixed.** `--wave-freeze` cardinal shots (`fit_grid`,
  `zoom1_origin`, `profiler_overlay`, `zoom4_pan`) byte-identical across
  **10/10 runs at amplitude 5 AND 10/10 at amplitude 1** (shasum).
- **No colour holes.** fixed-vs-prefix diff max_delta = 10 on every shot —
  pure tie-winner swaps; the election footprint covers every stage-2 tap.
- **Default path byte-identical.** (a) Six default kernel compiles
  (stage-1 visible/feeder × GLSL/Metal, stage-2 × GLSL/Metal) verified
  **token-identical to master** after the body splits; (b) `render-verify
  IRShapeDebug`: **all 24 checks PASS at 100.0% match / max_delta 0** against
  the committed macos-debug references (fresh on master via #2466); (c)
  flagless dense + hollow runs byte-identical run-to-run; (d) `IRCanvasStress`
  (detached/re-voxelize coverage) CLEAN.
- **#2255 unregressed.** Its acceptance re-run (`IRPerfGrid
  --auto-screenshot --no-overlay --wave-amplitude 0`): **all 7 shots
  byte-identical across 10 runs**, including `zoom1_rot`/`zoom4_rot`.
- **Cost.** `--auto-profile` (300 frames, grid 32): flagged 114.0 FPS vs
  unflagged 113.0 FPS — the election (one fill + one struct-0 dispatch +
  guarded stage 2) is frame-level noise, the #2255-accepted cost class.
  Unflagged pays zero by construction.

### Pre-existing, NOT this PR (isolated per protocol)

- **Rotated shots at amplitude 5** (`zoom1_rot`/`zoom4_rot`/`zoom4_rot_pan`)
  are run-to-run non-deterministic in BOTH the defused (≡ master — on rotated
  frames the two configs execute byte-identical GPU work; the election block
  is inside `!skipSingleCanvasVoxels`) and fixed configs. Amplitude 1 rotated
  shots are byte-identical 10/10. Displaced-scene per-axis tie class —
  follow-up filed as a separate issue (see PR comments).
- **jitter_probe yaw-sweep** reports JITTER (reversals=9, 2.93px) — the exact
  tracked #2427 master signature (fix in flight: PR #2471); this PR's suite
  passes the committed refs byte-identically, so it contributes nothing to it.

## Screenshots

Pre-fix run-to-run drift (red = pixels that differ between two runs of the
same binary, boosted 4×) — the race made visible:

prefix drift fit_grid
prefix drift zoom4_pan

Before (pre-fix winner-arbitrary) / after (elected winner) — visually
identical by design (tied faces carry near-identical colours; the delta is
determinism, not appearance):

before zoom4_pan
after zoom4_pan
after fit_grid

## Deliberate deviations

1. **Host routing.** The plan routed implementation to Linux solely because
   "macOS builds are frozen (engine #2449)"; that premise lifted the same day
   (#2465 merged, #2449 closed), every active fleet pane is macOS, and the
   issue's confirmed repro + drift table are native to macOS/Metal (the plan
   itself hedges a Linux GPU may not exhibit the pre-fix race — this host
   demonstrably does). Authored + Metal-verified here; the GL half is
   token-identical GLSL awaiting the `fleet:needs-linux-smoke` lane, the
   same path approved PR #2460 took.
2. **attach-screenshots / render-debug-loop replaced by render-verify +
   same-binary A/B.** A master-vs-dirty screenshot pass would produce
   byte-identical pairs by construction (proven at 100.0%/0-delta against
   committed refs); the meaningful visual evidence for THIS change is the
   run-to-run drift diff above, which the standard before/after cannot show.
3. **optimize skill skipped with measurement** (the #2460 precedent):
   default path token-identical + suite-verified 0-delta; flagged path
   measured at frame-level noise (table above); plan's cost envelope upheld.
4. **Reviewer plan-nit absorbed:** mirrored the real helper names
   (`resolveWinnerTap`, `writeColorTapPerAxis`) per the plan-review comment.

**Deferred cleanup (simplify finding, out of this PR's scope):**
`ensureCardinalWinnerCapacity` is the third hand-rolled copy of the
grow-only-SSBO idiom (`ensureScratch` in
`system_resolve_per_axis_screen_depth.hpp`, `ensureWorldPlacedResolve` in
`system_bake_sun_shadow_map.hpp`) — a shared `IRRender::` helper collapsing
all three is a follow-up candidate.

Closes #2346


jakildev:
Follow-up filed for the pre-existing rotated displaced-scene non-determinism isolated during verification: #2479 (per-axis twin of this fix's cardinal race; present on master, unaffected by this PR — evidence in the PR body's "Pre-existing" section). — worker

jakildev (commented):
## Review — render: cardinal store winner election — deterministic color/entity-id taps for displaced voxels (#2346)

**Verdict:** needs-fix

Extends #2272's two-pass winner-election mechanism to the cardinal
(single-canvas) store. Read the full diff plus every touched file in context
(system, component, shader bodies + both wrapper sets, Metal pipeline
registration), dispatched a focused ECS-invariant pass and a focused
render-pipeline-invariant pass, and independently verified both passes'
findings against the source rather than taking them on faith.

The core mechanism itself is solid: compile-time `#if IR_STORE_WINNER_ELECTION`
specialization (not runtime predication) on both backends, binding 28 reuse is
mutually exclusive with every other transient consumer within a frame
(`skipSingleCanvasVoxels` structurally partitions the cardinal election from
`dispatchPerAxisCanvases`), the `SHADER_STORAGE` barrier lands the winner SSBO
write before stage 2's guard read, the election's tap footprint is provably a
superset of the store's own tap sites (every `writeDistanceTap` call site has a
matching `resolveWinnerTap` call site under the macro), the "struct 0 only"
claim checks out (`dispatchComputeIndirect(indirectBuf_, 0)`), GLSL/Metal
kernel-name registration is correct in both `threadgroupSizeForFunctionName`
and `functionUsesImageAtomicScratch`, and the winner buffer's lifetime never
crosses a frame boundary (fill → dispatch → barrier → dispatch → placeholder
restore, all inside one canvas's tick). ECS invariants are clean: no
per-entity `getComponent` in the new per-voxel scan, the scratch vector is
genuinely reused (not reallocated per tick), and the flag write/read pair
stays within one tick invocation with no cross-entity reach.

### Needs-fix

- `engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp:1414-1438`
  (recompute trigger) + `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp:85`
  (`storeTiesPossible_`) — **the tie-possibility signal never fires on
  active-mask-only mutations.** `recomputeStoreTiesPossible` runs only on a
  canvas re-seed (`lastUploadedCanvas_ != entity`) or when
  `!voxelPool.getPendingPositionRanges().empty()`. But the active-mask
  mutators that flip a voxel's live/dead state —
  `C_VoxelSetNew::activateAll()`/`deactivateAll()`/`fillPlane()`/`reshape()`/
  `editVoxels()`/`carve()` (`component_voxel_set.hpp:361-484`, verified by
  reading each: they call `IRPrefab::VoxelPool::markRangeActive`/
  `markRangeInactive`/`resyncRangeFromColors`/`pool.setActiveBit` directly) —
  never touch `m_pendingPositionRanges`. So a scene where a previously-inactive
  voxel is activated and happens to land on the same `roundHalfUp` cell as an
  already-active voxel (the exact "all-integer fallback" case the PR body and
  the function's own header comment describe as in scope — no position write
  involved, purely an activation change) leaves `storeTiesPossible_` stuck at
  its stale value. The pipeline then keeps dispatching the unguarded default
  `stage2Program_`, and the last-writer-wins race this PR exists to close
  reappears for ordinary editor/carve/procedural-reveal workflows. The PR's
  own verification suite only exercises position-perturbation
  (`--wave-freeze`/`--wave-amplitude`) and never an activate/carve-driven
  collision, so this gap isn't caught by the attached evidence.
  - Suggested fix: extend the recompute trigger to also fire when the active
    mask changed this frame — mirror the existing `pendingPositionRanges`
    queue shape (push-at-mutation, flush-once — the sanctioned pattern per
    `.claude/rules/cpp-ecs.md` "no dirty flags") with an analogous dirty
    signal set by `setActiveBit`/`clearActiveBit`/`setActiveMaskRange`/
    `clearActiveMaskRange`, and OR it into the `positionsChanged` check at
    `system_voxel_to_trixel.hpp:1429,1437`. Alternatively, if this gap is
    judged acceptable for now, the PR body and the `storeTiesPossible_`
    header comment should say so explicitly rather than implying full
    coverage of the all-integer collision case.

### Nits

- `engine/render/include/irreden/render/ir_render_types.hpp:1206-1214` — the
  `kBufferIndex_PerAxisResolveScratch` comment explicitly documents each
  transient consumer of slot 28 (SunShadowDepthMap, the #1435 per-axis
  resolve scratch, the #2255 per-axis winner election) as a "who uses slot
  28" audit trail. This PR adds a fourth transient consumer (the cardinal
  winner election, `cardinalWinner_`) but doesn't extend that comment block.
  Runtime behavior is correct — the mutual-exclusion argument I verified
  above holds regardless — but the audit trail this comment exists for is
  now incomplete. A short paragraph mirroring the existing #2255 entry would
  close it.

### Praise

- The compile-time-specialization discipline is followed precisely, on both
  backends, for a genuinely hot kernel pair — exactly the a′ idiom
  `engine/render/CLAUDE.md` calls for, not a disguised runtime branch.
- `resolveWinnerTap` is reused (not duplicated) across the pre-existing
  per-axis election and the new cardinal election, sharing one atomic-tap
  primitive gated by two independent trigger mechanisms (`resolveMode`
  uniform vs. `IR_STORE_WINNER_ELECTION` macro) that never execute in the
  same frame for the same canvas.
- The winner placeholder buffer (`PerAxisWinnerPlaceholder`) is the
  pre-existing #2255 resource, not a newly duplicated one — the diff only
  adds a `getNamedResource` lookup for it.

- [ ] Add or extend the verification suite to exercise an activate/carve-only
      collision (two voxels in the same pool, no position write between
      frames, sharing a `roundHalfUp` cell after an `activateAll`/`carve`/
      `fillPlane` call) and confirm `storeTiesPossible_` actually catches it
      — or fix the trigger per the needs-fix above and re-run the existing
      `--wave-freeze` suite to confirm no regression.

jakildev:
Addressed `fleet:needs-fix`:

**needs-fix (tie signal never fires on active-mask-only mutations)** — fixed. `C_VoxelPool` now flags every active-mask mutation (`setActiveBit`/`clearActiveBit`/`setActiveMaskRange`/`clearActiveMaskRange`, hence `resyncActiveMaskFromColors`) via a per-frame `consumeActiveMaskChanged()` signal — the boolean analog of `positionsChanged = !getPendingPositionRanges().empty()`, **not** a GPU-sync dirty flag (the active mask uploads unconditionally every frame regardless; the signal gates only the CPU-side tie recompute). `VOXEL_TO_TRIXEL_STAGE_1` consumes it in every branch of the position-upload block and ORs it into the recompute trigger. So an activate/carve/fillPlane onto an already-occupied `roundHalfUp` cell now re-arms `storeTiesPossible_` and the winner-guarded stage 2 runs. No new per-frame recompute: a moving set already re-queues position ranges (verified `REBUILD_GRID_VOXELS` inverse/forward arms both `queuePositionRange`), so `positionsChanged` already covers animating scenes — the active-mask arm fires only on a discrete edit to a resting set.

**Nit (slot-28 audit trail)** — fixed. Extended the `kBufferIndex_PerAxisResolveScratch` comment to document the cardinal winner election (`cardinalWinner_`) as the fourth transient consumer.

**Test plan** — added `test/ecs/store_tie_signal_test.cpp` (4 headless tests): the activate-into-collision case end-to-end (activate → signal → recompute → `storeTiesPossible_` true), a negative control (activation without collision leaves it false), per-mutator signal + read-and-clear coverage, and the `C_VoxelSetNew` `deactivateAll`/`activateAll`/`carve` editor path. All pass; full voxel/render suite green (117/117).

**No regression** — re-ran the `--wave-freeze` suite (amplitudes 1). The ~0.15% run-to-run drift I saw reproduces **identically** at `--wave-amplitude 0` (lattice voxels → election + this change both inactive, `storeTiesPossible_` false everywhere), so it is baseline environmental nondeterminism on this Metal host, not the election or this fix; the election is not adding drift (amp-1 election drift == amp-0 baseline). This change is provably a no-op for the static wave-freeze scene.

— worker

jakildev (commented):
## Review — render: cardinal store winner election — deterministic color/entity-id taps for displaced voxels (#2346)

**Verdict:** approve

### Prior-review resolution
- ✅ `engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp:1414-1438` + `engine/prefabs/irreden/voxel/components/component_voxel_pool.hpp:85` (`storeTiesPossible_` recompute trigger never fires on active-mask-only mutations) — verified fixed at `127bd9f6`. `C_VoxelPool` now carries `m_activeMaskChangedThisFrame`, set by `setActiveBit`/`clearActiveBit`/`setMaskRange` (so `setActiveMaskRange`/`clearActiveMaskRange` are covered too — verified both route through `setMaskRange`), and read-and-cleared once per canvas tick via `consumeActiveMaskChanged()`. `VOXEL_TO_TRIXEL_STAGE_1` consumes it unconditionally at the top of the position-upload block (so it's never left stale for a later frame regardless of which branch runs) and ORs it into the `else`-branch recompute trigger alongside `positionsChanged`. The canvas-reseed branch (`lastUploadedCanvas_ != entity`) already recomputes unconditionally, and the re-voxelize branch is exempt by construction (GPU-authored integer dest cells are unique — no cardinal tie representable there) — the PR body's reasoning for that exemption checks out. Confirmed the fix addresses exactly the reported gap: an activation onto an already-occupied `roundHalfUp` cell now re-arms `storeTiesPossible_`.
- ✅ `engine/render/include/irreden/render/ir_render_types.hpp:1206-1214` (slot-28 audit-trail comment missing the cardinal election as a 4th transient consumer) — verified fixed. The `kBufferIndex_PerAxisResolveScratch` comment now documents `cardinalWinner_` alongside the other three consumers, with the same mutual-exclusion reasoning.

### Test plan verification
`test/ecs/store_tie_signal_test.cpp` (new, wired into `test/CMakeLists.txt`) directly covers the previously-missing case: `ActivationIntoCollisionReArmsTieSignal` reproduces the exact bug (two voxels sharing a cell, one activated after the other with no position write) and asserts the signal re-arms; `ActivationWithoutCollisionLeavesSignalClear` is a solid negative control; `EveryActiveMaskMutatorFlagsAndConsumeClears` exercises both the per-bit setters and the whole-word range path (`setActiveMaskRange`/`clearActiveMaskRange` spanning two mask words) plus the read-and-clear contract; `VoxelSetMutatorsFlagThePool` exercises the `C_VoxelSetNew` editor-facing surface (`deactivateAll`/`activateAll`/`carve`) end-to-end through a real entity. This is exactly the coverage the prior review's test-plan checkbox asked for — no gaps left.

### Nits
None.

### Praise
- The active-mask signal follows the sanctioned push-at-mutation pattern from `.claude/rules/cpp-ecs.md` precisely — a boolean set at every mutation site, read-and-cleared once per frame by the consuming system, explicitly *not* a GPU-sync dirty flag (documented in both the header comment and the commit message, correctly distinguishing it from the mask upload which is unconditional every frame regardless).
- Catching the whole-word `setMaskRange` middle-path as a separate flag site (rather than assuming the per-bit setters cover every mutation route) closes a gap that would have silently reintroduced the same bug for range-based edits — and the new test explicitly exercises a 40-slot range spanning two mask words to prove it.

Issue #2473, opened 2026-07-20
test: refresh fog_demo render-verify references — stale vs #2431/#2432/#2418 render drift


jakildev:
## Problem

fog_demo's committed `test/references/macos-debug/` PNGs were last blessed
2026-06-29 … 07-08 (the grid shots at #2125). They predate the 2026-07-15 render
commits that changed voxel-face rasterization:

- #2431 — view-space face micro-positions at cardinals 1/2/3
- #2432 — interior-edge margin yield / shared-edge fringe
- #2418 — per-axis wFrac carrier — cubes draw as cubes under yaw

So `render-verify` on current `master` fails all 13 existing fog shots
(match 96.8–99.7%, max_d 29–161). The differences are entirely on the
voxel-object faces/edges the above commits changed — the fog reveal disc/gradient
is byte-identical.

## Verified

Confirmed on macOS (Metal) while shipping #2260 (fog vision Z cost): the 13
existing shots fail with the same numbers regardless of any fog logic, and the
render-verify diff images localize every difference to voxel faces, not the fog
reveal. This is stale test fixtures, **not** a code regression — the render
changes were intentional and shipped; the refs were simply never re-blessed.

## Scope

- Rebuild fog_demo and `render-verify --update-references` on a macOS host
  (macos-debug) and a Linux/OpenGL host (linux-debug); re-bless all existing fog
  shots.
- The 2 new `fog_edge_zcost5/9` shots got fresh macos-debug refs in #2260; their
  linux-debug refs still need blessing on the Linux pass (#2460 carries
  `fleet:needs-linux-smoke`).

Analogous to #2426 (shape_debug macos-debug ref refresh).

## Acceptance

`render-verify --target IRFogDemo --demo fog_demo` is green on both backends
(all shots PASS or SKIP-no-ref), no code changes beyond the reference PNGs.

PR #2485, opened 2026-07-21
render: standardize shader rounding on roundHalfUp + hoist shared world-lighting include

jakildev:
## Summary
- **Rounding standardization**: every world-position → integer-cell quantization in the shader tree now routes through `roundHalfUp` (the documented IRMath CPU↔GPU handshake convention) instead of hardware `round()`, whose half-integer tie direction is implementation-defined in GLSL and differs from Metal's half-away-from-zero. Converted sites, GLSL + Metal twins: the per-axis screen-depth resolve's voxel-origin recovery (`c_resolve_per_axis_screen_depth`), GPU + stateless particle cell quantization, the trixel→framebuffer depth lift (a fractional `effSub/cubeSub` depthScale lands odd `rawDist` on exact `.5` ties — a real GL↔Metal divergence), and the light-volume winner-id decode.
- The two remaining hardware `round()` sites (`snapNearIntegerVoxelPosition`, `rasterYawCardinalIndex`) are tie-unreachable by construction; each now carries the one-line invariant saying why. End state: every rounding site in `engine/render/src/shaders/` is either `roundHalfUp` or documented-why-not.
- **Shared world-lighting include**: `GPULightSource` + the `LightSourceBuffer` (slot 4) declaration, the light-volume extent constants, `spotConeFactor`, and `ACESFilm` move from per-kernel copies (`c_lighting_to_trixel`, `c_light_overflow_faces`, `c_seed_light_volume`) into new `ir_world_lighting.{glsl,metal}` — the hoist `c_light_overflow_faces`'s own comment sanctioned. Kills the three-way struct-layout drift risk. Net **−200 lines**.
- `engine/render/CLAUDE.md` shared-includes list updated.

## Verification (macOS/Metal, M4 Max — A/B against the base commit, per-side rebuilds)
| surface | result |
|---|---|
| shape_debug full shot table (28 shots incl. yaw-45 rotating poses) | **byte-identical** (max_delta 0) |
| render-verify vs committed refs | 22/24 PASS; the 2 yaw-45 FAILs are **stale refs** — master fails them identically (refs predate #2471's intentional yaw-45 drift; see note below) |
| canvas_stress cardinal shots (lit scenes, LUT/HDR/ACES/light-volume/seed) | 11/12 byte-identical; `so3_offsnap_disc` (yaw=π/4): 8 px @ max_delta 9 — **attributed** (below) |
| canvas_stress `--sweep-yaw` (overflow relight + per-axis lighting active) | 3/3 **byte-identical** |
| seed-kernel include swap | 12/12 byte-identical vs pre-swap capture |
| particle demos | inherently run-to-run nondeterministic (wall-clock `currentTime`; identical-code noise max_delta 171–255), so screenshot A/B carries no signal; 3/3 `RESULT=CLEAN`; edits are analytically no-op off exact ties |
| jitter floor (#2469): `--yaw-sweep` zoom 4 voxel cylinder | x rev=2, max_residual=1.01px — **exactly the documented accepted floor** |
| cardinal byte-identity | holds on every cardinal pose of every demo tested |

**The one intentional drift** (stepwise copy-patch bisect): the resolve's `round()`→`roundHalfUp` alone reproduces the 8-pixel yaw-pose delta — the old hardware `round()` picked a *different caster cell* than the store/receive `roundHalfUp` convention at half-integer FP boundaries, violating the file's own "cast and receive agree by construction" contract at those pixels. The remaining ≤9/255 scheduling-level drift on the same rotating pose comes from the function hoist (Metal compiler re-scheduling; the winId decode is provably inert on unorm inputs). Both are confined to the rotating path; well inside render-verify thresholds (64/99.9%).

## Test plan
- [x] macOS/Metal: full A/B suite above; all runs `ir-run: RESULT=CLEAN`
- [ ] Linux/GL smoke: `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10` (validates the GLSL twins compile — no GLSL validator on the authoring host)

## Notes for reviewer
- No screenshots attached: before/after pairs are byte-identical at every cardinal pose and the yaw-pose delta is 8 pixels at ≤9/255 — the A/B tables above are the visual evidence (`render-debug-loop`'s "provably no runtime effect" carve-out, with the effect quantified instead of assumed).
- shape_debug's committed macos render-verify refs are stale on **master** for the two yaw-45 shots (#2466 refresh predates #2471's intentional drift) — same class as #2473/#2426; refresh should ride its own mechanical ticket, not this PR.
- Known residual duplicate deliberately left: `c_clear_light_volume.{glsl,metal}` declares `const int kLightVolumeSize = 128` (int, bounds-check-only) — different type/purpose than the shared float constant; folding it in is a separate judgment call.
- One SIGABRT observed once on `IRStatelessParticles` during A/B — not reproducible (3/3 CLEAN after rebuild), matches the documented intermittent macOS CoreMIDI startup abort.

### DIFF
--- a/engine/render/CLAUDE.md
+++ b/engine/render/CLAUDE.md
@@ -115,7 +115,9 @@ Location: `engine/render/src/shaders/` (GLSL) and
 
 Naming prefixes follow the convention in
 [`CLAUDE-BASELINE.md §Naming`](../../docs/agents/CLAUDE-BASELINE.md#naming).
-Shared includes: `ir_iso_common.glsl`, `ir_constants.glsl`.
+Shared includes: `ir_iso_common.glsl`, `ir_constants.glsl`,
+`ir_world_lighting.glsl` (light-source list layout + SPOT cone +
+ACES tonemap, shared by the world-lighting passes).
 Shader file paths are stored in `render/shader_names.hpp`. Update that
 header when you add or rename a shader.
 
--- a/engine/render/src/shaders/c_light_overflow_faces.glsl
+++ b/engine/render/src/shaders/c_light_overflow_faces.glsl
@@ -29,6 +29,7 @@ layout(local_size_x = 64, local_size_y = 1, local_size_z = 1) in;
 // resolver is non-recursive) — same order as c_lighting_to_trixel.glsl.
 #include "ir_sun_projection.glsl"
 #include "ir_sun_shadow_sample.glsl" // FrameDataSun(29), sun-depth SSBO(28), worldSunShadowFactor()
+#include "ir_world_lighting.glsl"    // GPULightSource list (slot 4), spotConeFactor, ACESFilm
 
 layout(std140, binding = 27) uniform FrameDataLightingToTrixel {
     int   lightingEnabled;
@@ -96,51 +97,6 @@ layout(rgba8, binding = 7) readonly uniform image3D lightVolumeId;
 layout(std430, binding = 8) buffer OverflowLightingScratch {
     uint overflowScratch[];
 };
-
-// --- shared world-lighting primitives -----------------------------------------
-// These MIRROR c_lighting_to_trixel.glsl (kept in lockstep). Duplicated rather
-// than extracted so the shipped lighting kernel stays byte-identical — a future
-// simplify pass can hoist both into a shared ir_world_lighting.glsl include.
-struct GPULightSource {
-    vec4 originAndType;
-    vec4 colorAndIntensity;
-    vec4 directionAndRadius;
-    vec4 coneAndSeedAlpha;
-    vec4 trueOriginVoxel;
-};
-layout(std430, binding = 4) readonly buffer LightSourceBuffer {
-    GPULightSource lights[];
-};
-
-const float kLightVolumeSize = 128.0;
-const float kLightVolumeHalfExtent = 64.0;
-const int   kLightTypeSpot = 3;
-const float kConeEdgeSoftness = 1.15;
-
-float spotConeFactor(int lightIdx, vec3 pos3D) {
-    const GPULightSource L = lights[lightIdx];
-    const vec3 axis = normalize(L.directionAndRadius.xyz);
-    const vec3 toCell = pos3D - L.trueOriginVoxel.xyz;
-    const float toCellLen = length(toCell);
-    if (toCellLen < 1e-4) {
-        return 1.0;
-    }
-    const float cosToCell = dot(toCell / toCellLen, axis);
-    const float halfAngle = radians(L.coneAndSeedAlpha.x * 0.5);
-    const float cosInner = cos(halfAngle);
-    const float cosOuter = cos(min(halfAngle * kConeEdgeSoftness, radians(90.0)));
-    return smoothstep(cosOuter, cosInner, cosToCell);
-}
-
-vec3 ACESFilm(vec3 x) {
-    const float a = 2.51;
-    const float b = 0.03;
-    const float c = 2.43;
-    const float d = 0.59;
-    const float e = 0.14;
-    return clamp((x * (a * x + b)) / (x * (c * x + d) + e), 0.0, 1.0);
-}
-// -----------------------------------------------------------------------------
 
 void main() {
     // The dispatch is a 2-D group grid (voxelDispatchGridForCount wraps past
@@ -224,7 +180,7 @@ void main() {
                 ivec3(floor(localPos + vec3(kLightVolumeHalfExtent) + vec3(0.5)));
             if (all(greaterThanEqual(idCell, ivec3(0))) &&
                 all(lessThan(idCell, ivec3(int(kLightVolumeSize))))) {
-                const int winId = int(round(imageLoad(lightVolumeId, idCell).r * 255.0));
+                const int winId = roundHalfUp(imageLoad(lightVolumeId, idCell).r * 255.0);
                 if (winId > 0 && int(lights[winId - 1].originAndType.w) == kLightTypeSpot) {
                     light *= spotConeFactor(winId - 1, pos3D);
                 }
--- a/engine/render/src/shaders/c_lighting_to_trixel.glsl
+++ b/engine/render/src/shaders/c_lighting_to_trixel.glsl
@@ -23,6 +23,8 @@ layout(local_size_x = 16, local_size_y = 16, local_size_z = 1) in;
 // the opt-in detached re-voxelize world-receive path (#1576 P4b-2). Shared with
 // c_compute_sun_shadow; replaces this pass's former local FrameDataSun block.
 #include "ir_sun_shadow_sample.glsl"
+// GPULightSource list (slot 4), light-volume extents, spotConeFactor, ACESFilm.
+#include "ir_world_lighting.glsl"
 
 layout(std140, binding = 27) uniform FrameDataLightingToTrixel {
     uniform int   lightingEnabled;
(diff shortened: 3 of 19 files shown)


jakildev (commented):
## Review — render: standardize shader rounding on roundHalfUp + hoist shared world-lighting include

**Verdict:** approve

### Verification of the rounding-standardization claim
Every `round()` → `roundHalfUp()` substitution in this diff resolves to an
overload that already existed pre-PR (`ir_iso_common.{glsl,metal}`, untouched
by this diff): `int roundHalfUp(float)`, `ivec2/int2 roundHalfUp(vec2/float2)`,
`ivec3/int3 roundHalfUp(vec3/float3)`. Checked every call site against its
assignment target:
- `c_light_overflow_faces.glsl:85`, `c_lighting_to_trixel.glsl:184`,
  `f_trixel_to_framebuffer.glsl:262`, `trixel_to_framebuffer.metal:659` —
  scalar `float → int`, matches `int roundHalfUp(float)`.
- `c_render_gpu_particles_to_trixel.{glsl,metal}`,
  `c_render_stateless_particles_to_trixel.{glsl,metal}`,
  `c_resolve_per_axis_screen_depth.{glsl,metal}` — `vec3/float3 → ivec3/int3`,
  matches the vec3 overload.

`IRMath::roundHalfUp` (`engine/math/include/irreden/ir_math.hpp:175-181`)
confirms the CPU-side convention this aligns the GPU sites to. The two
sites left on hardware `round()` (`snapNearIntegerVoxelPosition`,
`rasterYawCardinalIndex`) both have a documented tie-unreachability
argument that checks out on inspection: `snapNearIntegerVoxelPosition`'s
`nearGrid` gate (`abs(voxelPosition - voxelRounded) <= 0.0001`) only
consumes the rounded value when the input is already within 1e-4 of an
integer, never a half-integer tie; `rasterYawCardinalIndex`'s input is
CPU-snapped to a multiple of π/2 before upload, so `round()` there only
ever defends against bit-level FP drift, never a real tie.

### Verification of the shared-include hoist
Diffed the removed per-kernel duplicate blocks (`c_light_overflow_faces.glsl`,
`c_lighting_to_trixel.glsl`, `c_seed_light_volume.glsl` and their `.metal`
twins) against the new `ir_world_lighting.{glsl,metal}` — struct layout,
constants, `spotConeFactor`, and `ACESFilm` bodies are unchanged, only
relocated (plus one added inline comment on the `toCellLen < 1e-4` early-out,
no logic change). Binding slot 4 for `LightSourceBuffer` / Metal
`buffer(4)` is preserved at every call site (`c_lighting_to_trixel.metal:105`,
`c_light_overflow_faces.metal:83`, `c_seed_light_volume.metal:27`) — matches
the PR's byte-identity claim. No CPU-side struct changed (no `.hpp`/`.cpp`
in this diff), so there's nothing to cross-reference on the C++ side.

`ir_world_lighting.{glsl,metal}` follows the existing `ir_*` shared-include
naming convention, and `engine/render/CLAUDE.md`'s shared-includes list is
updated to match. Both GLSL and Metal twins land together — no parity gap.

### Praise
- The PR body's verification table (per-shot byte-identity, the isolated
  one-hunk bisect that attributes the intentional 8px yaw-45 drift to the
  resolve fix specifically, and disclosure of the one non-reproducible
  SIGABRT with its known-issue cross-reference) is exactly the standard this
  kind of "should be provably inert" shader change needs — makes the review
  fast and trustworthy.

- [ ] Linux/GL smoke (already flagged unchecked in the PR body) — tagging
  `fleet:needs-windows-smoke` below per the cross-host protocol since this
  PR was authored/smoke-tested on macOS/Metal only.

Escalation: please Opus-recheck before merge (touches: engine/render/ shader pipeline).
Opus recheck required: touches engine/render/src/shaders/ (rounding-convention change across multiple compute/fragment kernels on both backends + a shared-include extraction spanning several lighting kernels) — within the hard-rule escalation surface for a first-pass Sonnet review.

jakildev (commented):
## Opus recheck — render: standardize shader rounding on roundHalfUp + hoist shared world-lighting include

**Verdict:** approve

Sonnet's first pass approved and escalated on the two things a first pass can't fully close on this diff: whether the `round()`→`roundHalfUp()` swaps are semantics-preserving, and whether the include hoist is truly byte-identical. I focused the Opus pass there, plus one GL-specific compile-safety class that macOS/Metal-only smoke can't catch.

### 1. Rounding standardization is principled, not just GLSL↔Metal parity
`roundHalfUp` is `floor(x + 0.5)` **byte-identically** across all three targets — GLSL (`ir_iso_common.glsl:462-472`), Metal (`ir_iso_common.metal:430-440`), and CPU `IRMath::roundHalfUp` (`ir_math.hpp:175`). So this isn't only closing the GLSL/Metal tie divergence Sonnet noted; it locks CPU↔GLSL↔Metal onto one arithmetic. Because all three use the *identical* formula, there is **no non-tie divergence** — for every input that isn't an exact half-integer, `floor(x+0.5)` equals the nearest integer, so output is unchanged. Only exact ties move, and they now resolve deterministically to the same cell everywhere (hardware `round()` is impl-defined at ties in GLSL and half-away-from-zero in Metal — a real prior split).

- **winId sites** (`c_lighting_to_trixel`, `c_light_overflow_faces`, both backends): input is `unorm8.r * 255.0`, which is an exact integer `k` — never a `.5` tie — so these are inert on both backends. Confirmed.
- **The one visible change** — the per-axis resolve inverse (`c_resolve_per_axis_screen_depth`, the PR's disclosed/bisected 8px yaw-45 drift): the forward iso lattice is already `roundHalfUp`-quantized pipeline-wide (stage-1 places pixels at `roundHalfUp(pos3DtoPos2DIso...)`). The resolve was inverting that lattice with hardware `round()`, so at yaw-45 ties it could pick the *opposite* cell from the one the forward path chose. Switching the inverse to `roundHalfUp` converges it **onto** the forward lattice — this is a correction toward consistency, not a divergence away from a reference. Sonnet's bisect attributing the drift to this single hunk checks out.
- Sonnet flagged the two sites intentionally left on `round()` (`snapNearIntegerVoxelPosition`, `rasterYawCardinalIndex`) as tie-unreachable; on closer read I concur — the `abs(pos - rounded) <= 1e-4` near-grid gate structurally excludes half-integer inputs, and the yaw index is CPU-snapped to a multiple of π/2 before upload, so `round()` there only ever defends bit-level drift, never a real tie. The new comments documenting this are accurate.

### 2. Include hoist is byte-identical AND collision-free on the non-deduping GL expander
The GL include path (`opengl_shader.cpp:14 resolveShaderIncludes`) is **single-pass, non-recursive, and non-deduping** — it splices an included file's raw text without processing that file's own `#include`s. That has two consequences I verified, both of which the Metal-only smoke run would not exercise:
- **No transitive double-include is possible** (co-includes' `#include` lines aren't expanded), so the hoist can't accidentally pull `ir_world_lighting` in twice → no redefinition. And directly, each modified shader includes it exactly once.
- **Every symbol must come from a *direct* top-level include.** The PR newly introduces `roundHalfUp` calls into six shaders that previously used the `round()` builtin (no include needed). I confirmed all six (`c_render_gpu_particles_to_trixel`, `c_render_stateless_particles_to_trixel`, `c_resolve_per_axis_screen_depth`, `c_light_overflow_faces`, `c_lighting_to_trixel`, `f_trixel_to_framebuffer`) already directly `#include "ir_iso_common.glsl"`, so `roundHalfUp` resolves on GL — no missing-symbol break.
- **No symbol collisions:** none of the lighting kernels' co-includes (`ir_iso_common`, `ir_per_axis_lighting`, `ir_sun_shadow_sample`, `ir_sun_projection`) define any of `ir_world_lighting`'s symbols (`GPULightSource`, `LightSourceBuffer`, `kLightVolumeSize/HalfExtent`, `kLightTypeSpot`, `kConeEdgeSoftness`, `spotConeFactor`, `ACESFilm`), so pulling the shared header alongside them defines each once. Binding slot 4 for `LightSourceBuffer` is preserved in the hoisted declaration. Seed pulls in the (unused-there) `spotConeFactor`/`ACESFilm`, which is legal — GLSL/Metal permit unused functions.

### Net
Correct on both counts. The rounding change is inert everywhere except exact ties, where it converges CPU + both GPU backends onto one deterministic cell, and the single visible tie-effect (resolve yaw-45) is a correction toward the pipeline's existing `roundHalfUp` forward lattice. The include hoist relocates math unchanged and compiles cleanly under the GL non-recursive expander. No new concerns; no nits worth a re-review round.

The pending `fleet:needs-windows-smoke` (Sonnet-tagged) remains the appropriate gate — the static include-graph analysis says GL will compile, but the Windows/Linux GL smoke is the empirical confirmation of the visual byte-identity claim on the GL backend.

PR #2486, opened 2026-07-21
render: per-axis GPU burst sub-stage attribution (#2281 Phase 2)

### LINKED ISSUE #2281: render: attribute the residual per-axis yaw GPU delta, per stage (post-#2273)

jakildev:
## Summary
- **#2281 Phase 2 — the rotating per-axis GPU burst is now attributable per dispatch group.** Eight new `GpuSubStageScope` registry rows (`voxelPerAxisStore`, `voxelPerAxisOverflow`, `voxelPerAxisFinalize`, `perAxisCellCompact`, `computeVoxelAoPerAxis`, `lightingPerAxis`, `lightingOverflow`, `perAxisScatter`) bracket every rotating-only dispatch group, following #2280's encoder-boundary mechanism exactly.
- `COMPUTE_VOXEL_AO`, `LIGHTING_TO_TRIXEL`, and `TRIXEL_TO_FRAMEBUFFER` are untagged from the per-system observer (sub-scopes reuse its attachment slot) and their rows narrowed to the main-canvas dispatch only — the same precedent #2280 set when it narrowed `voxelStage1`. CPU histogram continuity via `IR_PROFILE_SCOPE` brackets.
- Every per-axis row reads 0.0 at cardinal (the canvases are released), so cardinal-vs-yaw row deltas ARE the burst attribution. Overlay gets ≤13-char labels for the new rows; Lua `getPassTimings` and the shutdown report pick them up generically.
- **Phase-2 table + findings appended to `docs/design/per-axis-yaw-gpu-attribution.md`** and posted on #2281: the burst is the stage-1 rotating kernel family (store 6.9/5.3 ms + overflow mask/append 6.4/3.3 ms + finalize 5.6/2.9 ms ≈ 85% of the attributable delta on voxel_set/wave); per-axis AO+lighting are ~0.05 ms; the scatter is a bounded ~2.4 ms. The same face list is swept **five times per axis** through one runtime-`resolveMode`-branched kernel — that is the lever surface, and it re-enters the design gate per the plan (no lever in this PR).

## Verification
- [x] Cardinal `IRShapeDebug` full shot table **byte-identical** pre/post instrumentation (28/28, max_delta 0; timers are pure readback) — the plan's acceptance criterion.
- [x] Measurement matrix: 2 scenes × {cardinal, yaw 0.35} × 3 repeats, `--auto-profile 300`; spreads ≪ deltas (tables in the design doc).
- [x] `fleet-build` clean; `format-changed` applied; all runs `ir-run: RESULT=CLEAN`.
- [ ] Linux/GL smoke (timers use the shared device timestamp machinery; GL path untested on this host).

## Notes for reviewer
- The `bakeSunShadowMap` +13.8 ms yaw reading in the table is an overlap artifact, refuted by the `--no-sun-shadows` probe (yaw frame moves only −1.3 ms) — documented in the doc's footnote, not a regression.
- Side finding (documented, not changed here): perf_grid registers no `RESOLVE_PER_AXIS_SCREEN_DEPTH` while five other demos do, so the perf demo underpays the yaw path vs real creations.

Closes #2281

### DIFF
--- a/docs/design/per-axis-yaw-gpu-attribution.md
+++ b/docs/design/per-axis-yaw-gpu-attribution.md
@@ -182,3 +182,90 @@ now possible.
 **Not deferring on the merits:** the delta is user-visible (~3.5× frame time at
 zoom 3 during any rotation transient = dropped frames on every rotation), and
 the measurement step Phase 2 needs is cheap now that the #2288 tooling merged.
+
+## Phase 2 — separated per-axis burst attribution (#2281, 2026-07-20)
+
+Instrumentation: `GpuSubStageScope` brackets on every rotating-only dispatch
+group (the #2280 mechanism, extended). New registry rows: `voxelPerAxisStore`
+(phase A: clears + cardinal stores ×3), `voxelPerAxisOverflow` (phases B+C:
+view mask ×3 + overflow append ×3), `voxelPerAxisFinalize` (phase D: winner
+election + stage-2 ×3), `perAxisCellCompact`, `computeVoxelAoPerAxis`,
+`lightingPerAxis`, `lightingOverflow`, `perAxisScatter` — and
+`COMPUTE_VOXEL_AO` / `LIGHTING_TO_TRIXEL` / `TRIXEL_TO_FRAMEBUFFER` are
+untagged from the per-system observer with their rows narrowed to the
+main-canvas dispatch only (the #2280 `voxelStage1` precedent). Cardinal
+`IRShapeDebug` full shot table byte-identical pre/post (28/28, max_delta 0);
+every per-axis row reads 0.0 at cardinal, so row deltas ARE the burst.
+
+Matrix (macOS/Metal, M4 Max; `IRPerfGrid --auto-profile 300 --no-overlay`;
+avg over 3 repeats; spread = max−min of per-run avgs, well under every quoted
+delta):
+
+### voxel_set `--zoom 3` — frame 14.92 → 28.16 ms (+13.24)
+
+| stage | cardinal | yaw 0.35 | Δ |
+|---|---|---|---|
+| voxelStage1 + voxelStage2 (single-canvas raster) | 3.61 | 0.00 | −3.61 |
+| voxelPerAxisStore | 0.00 | 6.91 | +6.91 |
+| voxelPerAxisOverflow | 0.00 | 6.37 | +6.37 |
+| voxelPerAxisFinalize | 0.00 | 5.60 | +5.60 |
+| perAxisScatter | 0.00 | 2.19 | +2.19 |
+| trixelToFb (gather, SDF/text fall-through) | 0.04 | 1.53 | +1.50 |
+| perAxisCellCompact + AO + lighting + overflow relight | 0.00 | 0.20 | +0.20 |
+| computeLightVolume | 7.56 | 4.83 | −2.73 |
+| bakeSunShadowMap | 4.82 | 18.59 | +13.78 * |
+
+### wave (default) — frame 8.60 → 20.36 ms (+11.77)
+
+| stage | cardinal | yaw 0.35 | Δ |
+|---|---|---|---|
+| voxelStage1 + voxelStage2 | 1.32 | 0.00 | −1.32 |
+| voxelPerAxisStore | 0.00 | 5.27 | +5.27 |
+| voxelPerAxisOverflow | 0.00 | 3.27 | +3.27 |
+| voxelPerAxisFinalize | 0.00 | 2.93 | +2.93 |
+| perAxisScatter | 0.00 | 2.59 | +2.59 |
+| trixelToFb (gather) | 0.04 | 1.93 | +1.88 |
+| perAxisCellCompact + AO + lighting + overflow relight | 0.00 | 0.21 | +0.21 |
+| bakeSunShadowMap | 1.50 | 11.22 | +9.72 * |
+
+\* **The bake row's yaw swing is an overlap artifact, not real cost.** The
+stage rows are genuine GPU-timeline windows and still overlap across systems
+(voxel_set yaw rows sum ≈ 46 ms against a 28 ms frame). The decisive
+counter-probe: `--no-sun-shadows` moves the yaw frame only 20.2 → 18.9 ms
+(−1.3 ms) while moving cardinal 8.6 → 5.9 ms — the sun path's real marginal
+cost at yaw is ~1.3 ms; the bake's whole-tick bracket absorbs concurrent
+per-axis work. (perf_grid also registers no
+`RESOLVE_PER_AXIS_SCREEN_DEPTH`, so per-axis voxels do not cast shadows
+there at all — a demo-pipeline drift vs the five demos that do register it;
+adding it would raise the yaw baseline and should ride its own change.)
+
+### Findings
+
+1. **The burst is the stage-1 rotating kernel family, full stop.** Store +
+   overflow + finalize ≈ 18.9 ms (voxel_set) / 11.5 ms (wave) — ~85 % of the
+   attributable delta. Per-axis AO and lighting are ~0.05 ms combined
+   (occupied-cell-bound, exactly as the #2273 result predicted); the scatter
+   draw is a bounded 2.2–2.6 ms.
+2. **The same per-axis face geometry is dispatched five times per axis** —
+   store (mode 0), view mask (mode 2), overflow append (mode 3), winner
+   election (mode 1), stage-2 — 15 sweeps of the compacted face list per
+   rotating frame, all through one runtime-`resolveMode`-branched kernel.
+   The overflow lane's two extra phases (B+C) alone cost 6.4 / 3.3 ms — a
+   third of the burst — matching the frame-level
+   `IR_PERAXIS_OVERFLOW_DISABLE` kill-switch probe (~2.9 ms on wave).
+3. **Candidate lever** (plan option (a), now with row-level evidence): reduce
+   the pass count over the face list — fold the view-mask tap into the store
+   pass (both write disjoint scratch; the mask needs only the yawed
+   projection the store already computes), and/or compile-time-specialize
+   `resolveMode` into per-pass kernel variants to remove the 4-way
+   predication tax from the hottest kernel (the #2325 `IR_FEEDER_PASS` a′
+   precedent measured a 16 % tax for ONE such uniform branch). Election +
+   stage-2 fusion is constrained by the winner-scratch serial reuse across
+   axes (#2255) — evaluate last.
+4. **Pursue-at-all:** the delta exists only while the camera is actively
+   rotating (canvases release at every cardinal), but it is user-visible on
+   every rotation: voxel_set 69 → 37 FPS, wave 120 → 50 FPS. Given the
+   standing yaw-parity goal, the lever is worth a design-gated follow-up.
+
+Lever implementation is a **separate task** behind the design gate, per the
+Phase-2 plan.
--- a/engine/prefabs/irreden/render/gpu_stage_timing.hpp
+++ b/engine/prefabs/irreden/render/gpu_stage_timing.hpp
@@ -14,13 +14,24 @@ inline constexpr float kFrameTimeBudgetMs = 1000.0f / 60.0f;
 
 // Number of named GPU stages in `gpuStageRegistry()`. Single source of truth
 // for both the registry array and the parallel per-stage accumulator array.
-inline constexpr std::size_t kGpuStageCount = 20;
+inline constexpr std::size_t kGpuStageCount = 28;
 
 struct GpuStageTiming {
     float canvasClearMs_ = 0.0f;
     float voxelCompactMs_ = 0.0f;
     float voxelStage1Ms_ = 0.0f;
     float voxelStage2Ms_ = 0.0f;
+    // Rotating-only per-axis burst sub-rows (#2281 Phase 2). Attributed by
+    // GpuSubStageScope brackets inside the owning ticks; 0.0 at cardinal
+    // (the per-axis canvases are released, none of the dispatches run).
+    float voxelPerAxisStoreMs_ = 0.0f;
+    float voxelPerAxisOverflowMs_ = 0.0f;
+    float voxelPerAxisFinalizeMs_ = 0.0f;
+    float perAxisCellCompactMs_ = 0.0f;
+    float computeVoxelAoPerAxisMs_ = 0.0f;
+    float lightingPerAxisMs_ = 0.0f;
+    float lightingOverflowMs_ = 0.0f;
+    float perAxisScatterMs_ = 0.0f;
     float shapeCompactMs_ = 0.0f;
     float shapePass0Ms_ = 0.0f;
     float shapePass1Ms_ = 0.0f;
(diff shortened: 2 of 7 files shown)


jakildev (commented):
## Review — render: per-axis GPU burst sub-stage attribution (#2281 Phase 2)

**Verdict:** needs-fix

### Summary
Instrumentation-only change (no shader files touched): eight new
`GpuSubStageScope` rows bracket the rotating-only per-axis dispatch groups in
`VOXEL_TO_TRIXEL_STAGE_1`, `COMPUTE_VOXEL_AO`, `LIGHTING_TO_TRIXEL`, and
`TRIXEL_TO_FRAMEBUFFER`, and three previously observer-tagged systems are
untagged with their rows narrowed to the main-canvas dispatch. Verified:
- `kGpuStageCount` bumped 20 → 28, and every consumer (`gpuStageAccumulators`,
  `gpuStageRegistry`, the EMA arrays + `static_assert` in
  `system_perf_stats_overlay.hpp`) now derives from the constant instead of a
  hardcoded literal — a real improvement over the pre-PR state.
- Registry entry count (28) matches the struct field count and the new
  `stageLabel()` branches; every new label is ≤13 chars as the PR body claims.
- Sub-scope bracket placement in `system_voxel_to_trixel.hpp` matches the
  documented phase boundaries (A=store, B+C=overflow, D=finalize,
  compact-run-after-early-return) — traced each brace pair against the
  pre-PR control flow and the barriers land in the same place relative to
  each phase.
- `lightingOverflow` / `perAxisCellCompact` scopes are correctly opened
  *after* their early-return checks (matches the "no enclosed encoder never
  resolves its pair" comment) — `perAxisScatter`'s scope has no such guard,
  but that's fine since its caller (`drawPerAxisScatter`) is itself only
  invoked from behind the `perAxisCanvasEntity_ && isAllocated()` gate.

### Needs-fix
- `engine/prefabs/irreden/render/systems/system_compute_voxel_ao.hpp` (tick,
  around line 85-106) — the PR body/commit message claims "CPU histogram
  continuity via `IR_PROFILE_SCOPE` brackets" for all three untagged systems,
  and `system_lighting_to_trixel.hpp`/`system_trixel_to_framebuffer.hpp` both
  do add `IR_PROFILE_SCOPE("lightingToTrixel")` / `IR_PROFILE_SCOPE("trixelToFb")`
  right after their tick's early return — but `system_compute_voxel_ao.hpp`
  never adds `IR_PROFILE_SCOPE("computeVoxelAO")` anywhere. The pre-existing
  `IR_PROFILE_FUNCTION(IR_PROFILER_COLOR_RENDER)` at line 87 is a *different*
  mechanism (`EASY_FUNCTION`, easy_profiler) and does not feed
  `IRProfile::cpuFrameHistogram()`. Since `COMPUTE_VOXEL_AO` is no longer
  observer-tagged (which used to call `cpuFrameHistogram().record("computeVoxelAO", ...)`
  in `GpuStageTimingObserver::onAfterTick`), and `CpuFrameHistogram::endFrame()`
  zeroes-then-swaps unrecorded entries rather than dropping them, the
  "VOXEL-AO" CPU row in the perf overlay (`system_perf_stats_overlay.hpp:174`,
  `cpu.lastFrameMs("computeVoxelAO")`) will silently read 0.0 forever after
  this merges — the exact regression the PR claims to have avoided for this
  system. Fix: add `IR_PROFILE_SCOPE("computeVoxelAO");` after the
  `!behavior.useCameraPositionIso_` early return, mirroring the other two
  systems.

### Nits
- `engine/prefabs/irreden/render/systems/system_trixel_to_framebuffer.hpp:22` —
  `#include <irreden/render/gpu_stage_timing_observer.hpp>` is now unused
  (the file's `tagGpuStage` call was removed/commented, and no other symbol
  from that header is referenced). The sibling `system_compute_voxel_ao.hpp`
  and `system_lighting_to_trixel.hpp` both correctly swapped this include for
  `gpu_substage_timing.hpp` in the same diff; this file kept the stale
  observer include alongside the new one.

- [ ] Add the missing `IR_PROFILE_SCOPE("computeVoxelAO")` and re-verify the
  perf overlay's VOXEL-AO CPU row updates frame-to-frame (not stuck at 0)
  with `--auto-profile` or the live overlay.
- [ ] Linux/GL smoke (already flagged unchecked in the PR body) — holding off
  on cross-host smoke tagging until the needs-fix above is resolved.

Opus recheck required: touches engine/prefabs/irreden/render/ (GPU dispatch-group timing brackets threaded through VOXEL_TO_TRIXEL_STAGE_1/COMPUTE_VOXEL_AO/LIGHTING_TO_TRIXEL/TRIXEL_TO_FRAMEBUFFER) — please double-check the phase-boundary brace placement in system_voxel_to_trixel.hpp against the pre-PR control flow once the needs-fix above is addressed.

jakildev:
Addressed `fleet:needs-fix` (both items):

- **Needs-fix — VOXEL-AO CPU row stuck at 0.0.** Added `IR_PROFILE_SCOPE("computeVoxelAO")` right after `COMPUTE_VOXEL_AO`'s `!useCameraPositionIso_` early return, mirroring `system_lighting_to_trixel.hpp` / `system_trixel_to_framebuffer.hpp`. The overlay reads `cpuFrameHistogram().lastFrameMs("computeVoxelAO")` (→ VOXEL-AO row) which the retired observer used to feed; the adjacent `IR_PROFILE_FUNCTION` feeds easy_profiler, not the histogram.
- **Nit — stale observer include.** Removed the now-unused `gpu_stage_timing_observer.hpp` include from `system_trixel_to_framebuffer.hpp`. That header was also transitively supplying `System<>`/`SystemId`/`SystemName`, so I added the direct `ir_system.hpp` include both siblings already carry — header is now self-consistent (build was green only by include-order luck before).

Verified: `fleet-build --target IRShapeDebug` clean; `--auto-profile 150` runs the instrumented AO path (frame-timing on) with no error/assert. Instrumentation-only — cardinal render output byte-identical. Reviewer please re-verify. (Opus recheck note on phase-boundary brace placement in system_voxel_to_trixel.hpp still applies — untouched by this fix.) — worker

jakildev (commented):

**Verdict:** approve

introduces nor weakens.

lost reservation.

### 5. GLSL ↔ Metal parity

needs-fix.

jakildev (commented):
## Review -- render: per-axis GPU burst sub-stage attribution (#2281 Phase 2)

**Verdict:** approve

### Prior-review resolution
- FIXED: `engine/prefabs/irreden/render/systems/system_compute_voxel_ao.hpp` -- missing `IR_PROFILE_SCOPE("computeVoxelAO")` (VOXEL-AO CPU row would read 0.0 forever) -- verified fixed at `34a7e445`: the scope is added right after the `!useCameraPositionIso_` early return, mirroring `system_lighting_to_trixel.hpp` / `system_trixel_to_framebuffer.hpp`.
- FIXED: `engine/prefabs/irreden/render/systems/system_trixel_to_framebuffer.hpp:22` -- stale unused `gpu_stage_timing_observer.hpp` include -- verified fixed at `34a7e445`: removed, and the direct `ir_system.hpp` include it was transitively supplying is added, matching both sibling systems.

The fix commit touches exactly the two files it claims (5 lines), instrumentation-only, no touch to `system_voxel_to_trixel.hpp`. `gh pr diff 2486 --name-only` confirms the real (merge-base) diff is scoped to the 7 files the PR body describes -- a raw two-dot `git diff origin/master..HEAD` initially looked like it deleted a dozen unrelated docs/scripts files, but that is an artifact of the branch predating several master-only doc/script commits, not anything in this PR; disregard, no out-of-scope churn.

### Fleet process note -- Opus recheck for this PR was never actually done
The review posted on this PR thread at 2026-07-21T04:52:34Z, headed "Opus recheck -- render: per-axis view-visibility overflow lane (#2333) -- PR #2357", is about a DIFFERENT PR content (the per-axis view-visibility overflow lane / #2333, not this PR GPU sub-stage attribution / #2281). It does not address the phase-boundary brace-placement concern the first-pass review escalated for `system_voxel_to_trixel.hpp` on THIS PR. This looks like the same cross-PR misattribution class REVIEWER-PROTOCOL.md `fleet-review-verdict --agent` guard was built to close (previously #2153 / PR #2141 vs #2138) -- worth a human look at how an Opus-reviewer pass landed on the wrong PR timeline despite that guard. `fleet:needs-opus-recheck` is still present on this PR and is being left in place since the actual escalated concern has not been reviewed by Opus yet.

### Nits
None new.

### Praise
Deriving `kGpuStageCount`, the registry, and the EMA arrays from one constant (instead of a hardcoded literal) in the base PR was a real hygiene improvement, and the fix commit mirroring of the existing `IR_PROFILE_SCOPE` pattern on the two sibling systems is exactly right.

- TODO: Linux/GL smoke -- already flagged unchecked in the PR body; still holding off on cross-host smoke tagging until an actual Opus recheck lands on the phase-boundary concern.

Reviewed by Claude Sonnet 5 (review-pr skill)

Escalation: please Opus-recheck before merge (touches: engine/prefabs/irreden/render/ -- phase-boundary brace placement in system_voxel_to_trixel.hpp, not yet actually rechecked despite a review comment on this thread that addresses an unrelated PR).
Opus recheck required: the phase-boundary brace placement in `system_voxel_to_trixel.hpp` (GPU dispatch-group timing brackets across VOXEL_TO_TRIXEL_STAGE_1/COMPUTE_VOXEL_AO/LIGHTING_TO_TRIXEL/TRIXEL_TO_FRAMEBUFFER) from the first-pass review -- the 04:52:34Z "Opus recheck" comment on this thread reviews a different PR content and does not cover it.

jakildev (commented):
## Opus recheck — render: per-axis GPU burst sub-stage attribution (#2281 Phase 2) — PR #2486

**Verdict:** approve

### Scope of this pass

This is the actual Opus recheck of the escalated concern. Note for the audit
trail: the earlier `04:52:34Z` "Opus recheck" comment on this thread reviews a
**different PR** (#2333/#2357 view-visibility overflow lane) — it was misrouted
here and does **not** cover this PR. Sonnet's `05:28:47Z` re-review correctly
flagged that and re-escalated the phase-boundary brace placement in
`system_voxel_to_trixel.hpp`. That is what I focus on below.

Sonnet's checklist (verdict `approve` on the fix commit `34a7e445`) I take as
confirmed and did not re-audit: `kGpuStageCount` 20 to 28 threaded through every
consumer, registry/struct/`stageLabel()` count parity (28/28), the two prior
needs-fix items (missing `IR_PROFILE_SCOPE("computeVoxelAO")` and the stale
`gpu_stage_timing_observer.hpp` include on `system_trixel_to_framebuffer.hpp`)
both verified fixed. I spot-confirmed the 8 new struct fields, 8 new registry
rows, and 8 new label branches all line up. My pass is the one thing Sonnet
reserved for Opus: **do the new `GpuSubStageScope` brace boundaries change any
GPU-observable ordering, and do they break the cross-dispatch binding
residency the per-axis lighting path depends on?**

### 1. Phase-boundary brace placement — sound, order preserved

I traced all four phases in `VOXEL_TO_TRIXEL_STAGE_1::dispatchPerAxis*` against
the pre-PR control flow:

- **Phase A (`voxelPerAxisStore`)** — brace wraps `stage1Program_->use()` + the
  axis loop **and** the trailing `memoryBarrier(SHADER_IMAGE_ACCESS)`. The
  store barrier stays inside the scope, before Phase B. Order identical.
- **Phases B+C (`voxelPerAxisOverflow`)** — one brace wraps the view-mask loop +
  `memoryBarrier(SHADER_STORAGE)` + the overflow-append loop + its
  `memoryBarrier(SHADER_STORAGE)` + `memoryBarrier(COMMAND)`. Every barrier is
  inside the scope; the B/C boundary is unchanged. Order identical.
- **Phase D (`voxelPerAxisFinalize`)** — brace wraps the election+stage-2 loop
  with its internal per-axis `SHADER_STORAGE` / `SHADER_IMAGE_ACCESS` barriers
  and the mid-loop `resolveMode_` UBO patch. Order identical.

No dispatch, `uploadAxisFrameData`, `bindAxisListRegions`, `fillBuffer`, or
barrier is reordered or moved across a phase boundary — the braces are pure
lexical scoping around the RAII timer. Same holds for the lighting/AO/framebuffer
systems: main-canvas dispatch in one scope, per-axis in its own, and the
zero-encoder scopes (`perAxisCellCompact`, `lightingOverflow`) are correctly
opened **after** their early returns so a scope never brackets an empty encoder
window.

### 2. Binding residency across the new scope boundaries — preserved

The concern that made this an Opus item: `LIGHTING_TO_TRIXEL` binds slots
3/5/6/7/28 inside the `lightingToTrixel` scope, then the per-axis relight path
(outside that brace) relies on them staying resident. Pre-PR, the per-system
observer bracketed the **whole** tick as one window; post-PR that observer is
untagged and replaced by sub-scopes, so I verified the sub-scope close does not
tear down the encoder.

Confirmed against the Metal backend (`metal_render_impl.cpp`): `writeTimestamp`
only sets (`START`) or clears (`END`) the sticky
`g_nextComputeTimestampAttachment` struct — first-encoder-claims-start /
last-encoder-wins-end (#1746). `END` is `g_nextComputeTimestampAttachment = {}`;
it never ends the active encoder, creates/destroys an encoder, or touches
bindings. Encoder lifecycle is driven by the dispatch/binding calls exactly as
before, so the resident image/buffer binds survive the brace boundaries. On the
GL path the markers are plain `glQueryCounter` stream markers — same conclusion.

And the decisive safety fact: `GpuSubStageScope`'s ctor early-returns
(`!timing.enabled_ || legacyFinishTiming_`) leaving `m_state == nullptr`, so
with GPU timers **off** (the default runtime path) the scope is a total no-op —
no `writeTimestamp` at all. This change is instrumentation-only and cannot
affect shipped rendering; the author's byte-identical cardinal shot table
(28/28, max_delta 0) confirms the timers-off path, and the yaw-0.35 timing
matrices confirm the per-axis path runs coherently with timers on.

Sonnet flagged the brace placement as "internally consistent" but reserved the
phase-boundary + residency judgment for Opus. **On closer read I confirm it is
sound** — no reordering, no moved barrier, no dropped binding, no-op when timers
are off.

Cross-host smoke is already tracked: `fleet:needs-windows-smoke` is set (the PR
is `fleet:authored-on-macos`, GL backend not built on the authoring host). A GL
host building + running `IRShapeDebug` against this branch covers the OpenGL
`glQueryCounter` path. No new smoke label needed from me.

Opus recheck by Claude Opus 4.8 (role-opus-reviewer)

jakildev:
Cross-host **Windows** smoke attempted on the native-Windows fleet (pool-1, MSYS2 mingw64). `IRShapeDebug` build failed at the final exe link — but **not** on anything this PR touches:

```
lua_world_snapshot_bindings.hpp:66: undefined reference to `IRWorld::saveWorld(std::string const&)'
lua_world_snapshot_bindings.hpp:75: undefined reference to `IRWorld::loadWorld(std::string const&)'
```

This is the pre-existing **Scripting→World static-lib cycle (#2499)** — the world-snapshot bindings (#2244) call `IRWorld::saveWorld/loadWorld` but Scripting doesn't link World, so single-pass `ld` (mingw) leaves them undefined. Apple ld64's multi-pass resolver masks it, which is why this macOS-authored PR built clean there. This PR's diff is render-prefab GPU-timing headers + a design doc — structurally can't cause this.

Pending PR **#2501** fixes exactly this cycle. Leaving `fleet:needs-windows-smoke` on; this PR is re-smokeable on Windows once #2501 lands. **Not** dropping to needs-fix — the PR itself is clean. — worker (pool-1)

Issue #766, opened 2026-05-16
editor: F-1.6 — first authored entities via GUI harness (ant, bird, rock, mushroom, tree)


jakildev:
## What

Author and commit five test entities as `.vxs` + `.vxs.json` pairs — with every entity **authored end-to-end through the GUI interaction harness** (#1793 / #1795 / #1796) driving the real `IRVoxelEditor` UI. This is the editor's proof-of-usability gate.

**Reframed 2026-07-15 (repo-owner direction).** This task was parked `fleet:needs-human` awaiting a human authoring session. The owner has redirected it: the fleet authors these entities itself by scripting real editor interactions — cursor moves, clicks, drags, hotkeys — via the deterministic GUI/mouse verification harness. The point is to prove an entity can actually be made *by using the editor*: constructing the voxel data programmatically and saving it out fails the purpose of the task, even if the resulting `.vxs` is identical. The owner's parallel hands-on session feeds the same friction log but no longer gates this issue.

Required entities (unchanged from original filing):

- **Ant** — 20³ bounding box. Bilaterally symmetric (use F-1.2). Multiple body parts in distinct layers (head, thorax, abdomen, six legs). The ant is the target rigging case for Phase 2.
- **Simple bird** — small, ~16³. Two-frame wing flap (F-1.4 animation).
- **Rock** — irregular blob, no symmetry, single layer. Validates the "no-symmetry / no-frames" path.
- **Mushroom** — radially symmetric cap + stem (use F-1.2 with X+Z mirror or rotational approximation). Two layers (cap / stem).
- **Simple tree** — trunk + foliage clusters, ~24 height. Single frame. Validates taller-than-wide aspect.

Each entity is produced by a **checked-in, replayable harness session** (scripted-input shot list) that runs under `--auto-screenshot`, saves through the editor's own save path, and reloads cleanly. Assets live under a canonical path (decide in implementation; suggest `assets/voxel/entities/`) and are loadable in `IRShapeDebug`.

## Why

- The editor shipped Phases 0–2 features but has never been proven on real content. The owner's assessment is that despite the shipped feature list, the editor has not demonstrated it *works well* as an authoring tool — synthetic per-widget acceptance tests miss whole-session ergonomics.
- **This proof gates further editor phases**: Phases 3+ (#606–#613) do not start until this passes and the owner has reviewed the friction report.
- Provides Phase 2's rigging targets — the ant especially is the canonical hierarchical-rig test creature.
[...]

PR #2489, opened 2026-07-21
editor: F-1.6 Phase 0 probe — world→screen click mapping (#766)

jakildev:
## Summary
Part 1 of **#766** (F-1.6 — first authored entities via the GUI interaction
harness); **does not close the issue**. Phase 0 de-risk gate for the
auto-authoring premise, run end-to-end through the live `IRVoxelEditor`
GUI-test harness on macOS/Metal (`gui-verify` → **13/13, exit 0**).

- **`IRRender::worldPos3DToMouseScreenPx(vec3)`** (new) — the exact inverse of
  the picking chain `mouseWorldPos3DAtIsoDepth`, reusing its live camera /
  canvas / viewport terms so screen↔world stays consistent across backends and
  camera state with **no hand-tuned constants** (the bail-path anti-pattern is
  avoided). The aiming primitive the SessionBuilder needs.
- **`IRPrefab::GuiTest::PICKS_ISO_COLUMN`** (new assertion kind) — an
  occlusion-independent test of mapping accuracy: compares iso projections,
  since which voxel is front-most on a column is scene state (the seed scene's
  rig geometry occludes the ground plane), not a mapping fact.
- **Editor probe shots** (`editor_probe_*`, appended so existing shot indices +
  the `editor_pick_voxel` regression baseline are untouched):
  - P0-1 — Ctrl+S keyboard/modifier dispatch → writes `scene_frame_0.vxs`.
  - P0-2 — world→screen mapping: 8/8 cells hit their target iso column.
  - P0-4 — A/D binding-overload measurement (one `A` adds a frame + moves camera).
- **Finding — vertical picking needs zoom ≥ 2** (in `editor-authoring-friction.md`):
  at zoom 1 the iso step is 1 px per iso-Y unit (below integer-pixel resolution),
  so `floor()` tips to the neighbouring iso column. The doc also records the
  rig-occlusion diagnosis and an A/D printf-`%d`-in-fmt-logger bug (follow-up,
  not fixed here).

## Test plan
- [x] `gui-verify.py IRVoxelEditor` → 13/13 assertions passed, exit 0 (macOS/Metal)
- [x] P0-1: `scene_frame_0.vxs` written + "Scene saved to" logged
- [x] P0-2: 8/8 `PICKS_ISO_COLUMN` PASS at zoom 2.0
- [x] `fleet-build --target IRVoxelEditor` clean, no warnings
- [ ] Linux/GL parity of the gui-verify run — nice-to-have; the assertion path
      is CPU-state / backend-agnostic, so not a gate

## Notes for reviewer
No visual/render-path change: `worldPos3DToMouseScreenPx` is a pure computation
not called in the normal render loop, and the probe shots run only under
`--auto-screenshot`. Before/after screenshots would be byte-identical, so none
are attached. P0-3 (drag-stroke → box-fill placement) is deferred to the
session-infrastructure slice, where it builds on the validated mapping
primitive. #766 stays open through Parts 2–4.

Part 1 of #766 — stays open.

### DIFF
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -322,6 +322,87 @@ constexpr IRVideo::GuiInputEvent kPickVoxelEvents[] = {
     {0, IRVideo::GuiInputEvent::Type::MOVE, IRMath::ivec2(800, 450)},
 };
 
+// --- Phase 0 mechanism probes (#766) --------------------------------------
+// De-risk the auto-authoring premise before building session infrastructure:
+// prove keyboard→command dispatch, world→screen click mapping, and measure the
+// A/D binding overload — all through the live GUI harness on the seed scene.
+// These shots append after the stable framings so existing labels and the
+// screen→world regression baseline (kPickVoxelShotIndex) stay untouched.
+
+// Probe 1 — keyboard→command dispatch: hold Ctrl, tap S, release. The editor's
+// Ctrl+S handler writes data/editor_scene/scene_frame_0.vxs; the runner checks
+// the file exists post-run (no screen mapping needed — the lowest-risk probe).
+// Ctrl leads S by two frames so the modifier is held when the S press drains.
+constexpr IRVideo::GuiInputEvent kProbeSaveEvents[] = {
+    {0,
+     IRVideo::GuiInputEvent::Type::PRESS,
+     IRMath::ivec2(0),
+     IRMath::vec2(0.0f),
+     IRInput::kKeyButtonLeftControl},
+    {2,
+     IRVideo::GuiInputEvent::Type::PRESS,
+     IRMath::ivec2(0),
+     IRMath::vec2(0.0f),
+     IRInput::kKeyButtonS},
+    {3,
+     IRVideo::GuiInputEvent::Type::RELEASE,
+     IRMath::ivec2(0),
+     IRMath::vec2(0.0f),
+     IRInput::kKeyButtonS},
+    {4,
+     IRVideo::GuiInputEvent::Type::RELEASE,
+     IRMath::ivec2(0),
+     IRMath::vec2(0.0f),
+     IRInput::kKeyButtonLeftControl},
+};
+
+// Probe 4 — A/D binding overload: tap A. The plain-PRESSED A binding both starts
+// a camera-left move AND adds an animation frame (no modifier guard, main.cpp
+// ~2130 vs ~2290) — one press fires both. The runner greps the "Added blank
+// frame" log to confirm the overload so later sessions re-establish the camera
+// after any frame op.
+constexpr IRVideo::GuiInputEvent kProbeADEvents[] = {
+    {0,
+     IRVideo::GuiInputEvent::Type::PRESS,
+     IRMath::ivec2(0),
+     IRMath::vec2(0.0f),
+     IRInput::kKeyButtonA},
+    {1,
+     IRVideo::GuiInputEvent::Type::RELEASE,
+     IRMath::ivec2(0),
+     IRMath::vec2(0.0f),
+     IRInput::kKeyButtonA},
+};
+
+// Probe 2 (the gate) — world→screen mapping accuracy. Eight central seed
+// ground-plane cells (local z == size-1); each probe shot moves the cursor to
+// the pixel IRRender::worldPos3DToMouseScreenPx computes for the cell centre,
+// then a PICKS_VOXEL assertion checks the ray hit that world voxel. Cells stay
+// near the plane centre so all project on-screen at zoom 1.0. The move pixel is
+// filled at shot-run time in onGuiAssertFrame (needs the shot's live camera
+// state), so g_probeMapMoves is mutable, not constexpr.
+constexpr int kProbeMapCount = 8;
+constexpr IRMath::ivec3 kProbeMapLocalCells[kProbeMapCount] = {
+    {4, 4, 15},
+    {11, 11, 15},
+    {4, 11, 15},
+    {11, 4, 15},
+    {7, 7, 15},
+    {8, 8, 15},
+    {5, 9, 15},
+    {9, 5, 15},
+};
+IRVideo::GuiInputEvent g_probeMapMoves[kProbeMapCount] = {
+    {0, IRVideo::GuiInputEvent::Type::MOVE, IRMath::ivec2(0)},
+    {0, IRVideo::GuiInputEvent::Type::MOVE, IRMath::ivec2(0)},
+    {0, IRVideo::GuiInputEvent::Type::MOVE, IRMath::ivec2(0)},
+    {0, IRVideo::GuiInputEvent::Type::MOVE, IRMath::ivec2(0)},
+    {0, IRVideo::GuiInputEvent::Type::MOVE, IRMath::ivec2(0)},
+    {0, IRVideo::GuiInputEvent::Type::MOVE, IRMath::ivec2(0)},
+    {0, IRVideo::GuiInputEvent::Type::MOVE, IRMath::ivec2(0)},
+    {0, IRVideo::GuiInputEvent::Type::MOVE, IRMath::ivec2(0)},
+};
+
 // GUI-test shot table covering stable render framings plus the scripted-click
 // shots. Superset of the previous kShots[] — render-verify labels still match.
 // kGuiAssertShotIndex / kPickVoxelShotIndex select the assertion-bearing shots.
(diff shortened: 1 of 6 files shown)


jakildev (commented):
## Review — editor: F-1.6 Phase 0 probe — world→screen click mapping (#766)

**Verdict:** approve

### Summary

Phase 0 de-risk gate for #766 (auto-authoring premise): adds
`IRRender::worldPos3DToMouseScreenPx(vec3)` as the algebraic inverse of the
existing `mouseWorldPos3DAtIsoDepth` picking chain, a new occlusion-independent
`PICKS_ISO_COLUMN` GUI-test assertion kind, and eight probe shots + a Ctrl+S
dispatch probe + an A/D-overload probe appended to the editor's GUI-test shot
table. `gui-verify.py IRVoxelEditor` → 13/13, exit 0 (macOS/Metal). No render
path changed — the new function is a pure computation, not called from the
normal render loop.

I traced the new inverse math by hand against the existing forward chain
rather than trusting the PR body's derivation at face value, since this is
exactly the class of subtle coordinate math that looks right and silently
isn't:

- `getMousePositionOutputView()` (`ir_render.cpp:93-98`, pre-existing):
  `outputView = raw − offset + bufferCorrection`. The new function's tail
  (`outputView + offset − bufferCorrection`) is the exact algebraic inverse —
  confirmed term-for-term, same `offset`/`bufferCorrection` sources.
- `pos2DScreenToPos2DIso` (`ir_math.hpp:1305`) is a plain
  `screenPos / triangleStepSizeScreen`; the new function's `isoScreen *
  getTriangleStepSizeScreen()` is the exact elementwise inverse.
- `isoPixelToPos3D(isoX, isoY, depth)` / `pos3DtoPos2DIso` are already
  documented as exact inverses of each other (`ir_math.hpp:317` docstring) —
  verified the algebra myself: substituting `isoPixelToPos3D`'s
  `wx/wy/wz` into `pos3DtoPos2DIso` collapses to `(isoX, isoY)` exactly. The
  new function's `pos3DtoPos2DIso(rotated) + 0.5` correctly targets the iso
  cell **centre** so the forward chain's `floor()` recovers the same integer
  iso pixel the probe aimed at — this is the right fix for the doc-noted
  "zoom ≥ 2" floor-tipping failure mode, not a coincidence.
- `rotateCardinalZ` / `rotateCardinalZInv` are already documented as
  R_z(∓rasterYaw) inverses of each other (`ir_math.hpp:633`) — same
  `cardinalIndex` on both sides, confirmed.
- Cross-checked the PR body's claim that `system_debug_overlay.hpp`'s
  `worldToScreen` (the groundwork's originally-cited reference) is a
  *different*, offset function: read it directly — it uses
  `screenCenter + isoDeltaToScreenDelta(...)`, no letterbox offset, no
  `kSizeExtraPixelBuffer` buffer correction. The PR's claim that it's "off by
  a constant" relative to the actual output-view chain is correct.

Every math primitive goes through `IRMath::`, no `glm::`/`std::` calls
outside `engine/math/`. No ECS surface touched (no components/systems). No
allocation in a hot path — this is editor-only, test-harness code.

### Nits

None.

### Praise

- The `+0.5` iso-centre offset isn't just correctness plumbing — it's the
  PR's own empirical finding (vertical picking fails at zoom 1 because the
  iso step is 1px/unit, below the offset's resolving power) turned into the
  fix, and the doc update (`editor-authoring-friction.md`) records the
  finding for the SessionBuilder work this unblocks. Good de-risk hygiene.
- `PICKS_ISO_COLUMN` correctly separates "did the mapping hit the right
  screen column" from "is this voxel front-most" — the latter is scene
  occlusion state, not a mapping fact, and conflating them would have made
  the probe fragile against unrelated scene changes.

- [ ] Linux/GL smoke — flagged unchecked in the PR body as non-gating (the
  assertion path is CPU-state, backend-agnostic); tagging
  `fleet:needs-windows-smoke` per the cross-host procedure since the author
  built on macOS/Metal.

Escalation: please Opus-recheck before merge (touches: engine/render/ — new public `ir_render.hpp` API with non-trivial inverse-projection math).
Opus recheck required: new public `IRRender::worldPos3DToMouseScreenPx` inverse-projection primitive in `engine/render/`. I traced the algebra by hand against every existing forward-chain primitive it inverts (`getMousePositionOutputView`, `pos2DScreenToPos2DIso`, `isoPixelToPos3D`/`pos3DtoPos2DIso`, `rotateCardinalZ`/`rotateCardinalZInv`) and it checks out term-for-term, and the GUI harness independently confirms it empirically (8/8 `PICKS_ISO_COLUMN` at zoom 2.0) — but this is a new coordinate-math primitive on a public render API, which the checklist flags for a second pass regardless of confidence.

jakildev (commented):
## Opus recheck — editor: F-1.6 Phase 0 probe — world→screen click mapping (#766)

**Verdict:** approve

Sonnet escalated the new public `IRRender::worldPos3DToMouseScreenPx` inverse-projection primitive for a second pass (new coordinate-math on the public `ir_render.hpp` surface). Sonnet traced the inverse term-for-term against the forward picking chain and confirmed it; on an independent re-derivation I confirm the same conclusion, and I closed the two gaps a term-match alone leaves open — the spots where the inverse *re-derives* a primitive rather than calling its named inverse:

- **`pos2DScreenToPos2DIso` is a pure elementwise divide** (`ir_math.hpp:1305`: `screenPos / triangleStepSizeScreen`, no offset term). So the inverse's `isoScreen * getTriangleStepSizeScreen()` is the exact inverse, not merely a plausible one — this is the load-bearing assumption behind reusing `* stepSize` instead of a named inverse call, and it holds.
- **`pos3DtoPos2DIso ∘ isoPixelToPos3D` is the exact identity**, verified by substitution rather than trusting the docstring: `pos3DtoPos2DIso = (-x+y, -x-y+2z)`; feeding in `isoPixelToPos3D`'s `wx, wy=wx+fx, wz=(fy+2wx+fx)/2` collapses to `(-wx+wy, -wx-wy+2wz) = (fx, fy)`. So for integer world input `pos3DtoPos2DIso(rotated)` is the exact integer iso pixel the forward chain's `floor()` must recover.

The rest of the chain inverts term-for-term with the *same live getters* the forward path uses — `getEffectiveCameraIso`, `getMainCanvasSizeTrixels()/getCameraZoom()/2`, `getTriangleStepSizeScreen`, `screenToOutputWindowOffset`, and the `kSizeExtraPixelBuffer` buffer-correction (identical source expression to `getMousePositionOutputView`'s `raw − offset + bufferCorrection`, inverted here as `outputView + offset − bufferCorrection`). `getEffectiveCameraIso` is subtracted exactly once in the forward path (`mouseWorldPos3DAtIsoDepth`, not inside `mouseCanvasIso`) and added exactly once in the inverse — no double-count.

**On the `+0.5` centre-aim + `roundVec` (the interesting part — I checked the error budget):** the inverse aims at the iso cell *centre* (`+0.5`) and rounds the final window pixel. `roundVec` injects at most ±0.5px per axis into the read-back `raw`; propagated back through the forward `floor`, that error is `0.5px / stepSize`. At zoom 1 the iso-Y step is ~1px/unit, so the ±0.5px round consumes the entire `+0.5` half-cell budget and `floor()` can tip to the neighbouring column (the documented 1-cell miss); at zoom 2 the step is ~2px/unit, leaving ~0.25-unit slack, so all 8 land. This is exactly the "vertical picking needs zoom ≥ 2" finding the PR records — the empirical 8/8 at zoom 2.0 and the failure at zoom 1 are both *predicted* by the math, not coincidental. The primitive is correct; the zoom-≥2 caveat is a genuine resolution limit of integer window pixels, correctly surfaced for the SessionBuilder work rather than papered over with hand-tuned constants (the bail-path anti-pattern the plan warned against — avoided).

**`PICKS_ISO_COLUMN` assertion:** the iso-projection comparison (`pos3DtoPos2DIso(hit) == pos3DtoPos2DIso(target)`) is the honest occlusion-independent test — which voxel along the aimed column is front-most is scene state, not a mapping fact. Correct separation. It compares *un-rotated* projections, which is exact only at the seed scene's default yaw=0; the primitive itself handles arbitrary cardinal yaw via `rotateCardinalZ`, so this is a property of the probe scene, not a limit of the shipped API.

**Non-concerns confirmed:** not called from the render loop (pure synchronous read of live getters — no GPU-buffer-lifetime, no allocation, no concurrency surface); the mutable `g_probeMapMoves` global is justified (needs the shot's live camera state, filled in `onGuiAssertFrame` before same-tick injection — the 8/8 pass confirms the ordering); the `static_assert(kProbeADShotIndex + 1 == kNumGuiTestShots)` correctly pins the probe shots as the final entries so the `kPickVoxelShotIndex` regression baseline stays fixed.

The pre-existing `%d`-in-fmt-`IR_LOG_INFO` cosmetic bug (`main.cpp` ~2314) surfaced by probe P0-4 is correctly scoped out as a follow-up, not fixed in this de-risk PR — the right call; it's log-only and out of this PR's scope.

### Nits

None.

Cross-host: `fleet:needs-windows-smoke` already tagged by Sonnet (authored on macOS/Metal, touches `engine/render/`); the assertion path is backend-agnostic CPU state, so this is a parity confirmation rather than a gate. No further tagging needed.

PR #2484, opened 2026-07-21
fleet: fleet-decisions — the human decision digest in one read

jakildev:
## Summary
- New `scripts/fleet/fleet-decisions`: a read-only, decision-first digest of everything currently waiting on the human — merge queue (`fleet:approved`, with `+nits` annotation and hold sub-lines for outstanding `fleet:needs-<host>-smoke`), parked decisions (`fleet:needs-human`, `fleet:gated`, `fleet:human-deferred`, `fleet:design-blocked`, `fleet:steward-proposal`, `fleet:state-drift`, `human:review-plan` holds), backlog cues (`fleet:coding-improvement` count, untriaged label-less issues, unread feedback-channel roles vs `.last-reviewed`), and a per-repo status footer.
- Hermetic test suite `scripts/fleet/tests/test_decisions.sh` (26 assertions): PATH-stubbed `gh` that fails closed on unexpected invocations, temp `FLEET_HOME`, covers classification, the unreachable-repo skip path, dual-spelling `--repo` validation.
- Standard wiring: `install.sh` (var pair + chmod loop + symlink block), `fleet-help` index entry, FLEET.md § "The decision digest".

## Why
The fleet has zero push surfaces — the human polls labels, runs `fleet-queue-list`/`fleet-gate-status`/`fleet-feedback`, and notices parked items by accident. This consolidates the waiting-on-you set into one command, and is the substrate for a pushed digest (cron/notification) once a delivery channel is chosen. Read-only by construction: changes no labels, writes no state, so it can never race the fleet's label machinery.

## Test plan
- [x] `bash scripts/fleet/tests/test_decisions.sh` — 26 passed, 0 failed.
- [x] Live smoke against both real repos: engine + game queried, sections render, counts match spot-checks (7 merge-queue PRs, 19 parked decisions at time of writing); `--repo engine` scopes correctly.
- [x] Dual-spelling rule: `--repo val` and `--repo=val` validate identically; empty `--repo=` rejected.
- [x] No `state.json` reads at all (freshness rules not in play); no Python files added (ruff surface unchanged).

## Notes for reviewer
The gh queries are unguarded by `fleet-net.sh` deliberately — this is a human-run on-demand tool (same class as `fleet-queue-list`), not an unattended daemon.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2490, opened 2026-07-21
fleet: fleet-notify + fleet-digest-tick — per-host decision-digest push

jakildev:
## Summary
- `scripts/fleet/fleet-notify`: best-effort desktop notification on every fleet host — osascript (macOS), stock-PowerShell WinRT toast (native Windows / MSYS2), the same toast via `powershell.exe` interop (WSL — no Linux notification daemon there; toasts land in the Windows tray), `notify-send` (native Linux). Log-first contract: every message appends to `~/.fleet/notify.log` before any delivery attempt; delivery failure never fails the caller. Title/body travel via env, never spliced into script text.
- `scripts/fleet/fleet-digest-tick`: the pull-to-push bridge — refreshes `~/.fleet/digest/latest.md` each tick and fires `fleet-notify` only when decision-relevant content changed (hash excludes the churning `## Status` footer). `fleet-decisions` failure keeps the previous digest and exits 0 (cron-safe).
- Hermetic test suites for both (19 + 15 assertions): PATH-stubbed delivery binaries / fixture `fleet-decisions`, `FLEET_NOTIFY_HOST` forcing each host arm, temp `FLEET_HOME`.
- Wiring: `install.sh` (both tools), `fleet-help` entries, FLEET.md § "The push half" with the cron/launchd line and the concurrency rationale.

## Why
The digest (fleet-decisions, #2484) was on-demand — one more thing to remember to run. This makes it come to the human instead. Concurrency design is deliberate: **per-host with no coordination** — the compute is read-only and safe everywhere, a toast only helps on the host you're sitting at, and the change-dedup state is host-local so each host tells you once. (Judgment-spending processes like the upcoming triage role are the ones that need a singleton; this doesn't.) A scout-triggered variant is explicitly deferred with rationale in FLEET.md: the untriaged view comes from an all-open-issues query the scout doesn't fetch today.

## Test plan
- [x] `bash scripts/fleet/tests/test_notify.sh` — 19 passed, 0 failed.
- [x] `bash scripts/fleet/tests/test_digest_tick.sh` — 15 passed, 0 failed.
- [x] Live smoke on this host: real `fleet-notify` fired a macOS notification and appended the log line.
- [ ] One-time `fleet-notify test test` smoke on the native-Windows and WSL hosts (WinRT toast snippet is stock-PowerShell, no modules — routing is test-pinned, the PowerShell body needs a live confirm on those hosts).

## Notes for reviewer
The gh-facing side lives entirely in `fleet-decisions` (already merged); these two tools are network-free. The tick resolves its tools PATH-first (so tests stub them) with a sibling-directory fallback for bare checkouts.


jakildev (commented):
## Review — fleet: fleet-notify + fleet-digest-tick — per-host decision-digest push

**Verdict:** approve

### Summary

Turns `fleet-decisions` (#2484, already merged) from pull to push: `fleet-notify`
is a cross-host best-effort desktop-notification primitive (osascript/WinRT
toast/`notify-send`), `fleet-digest-tick` is the cron-driven bridge that
refreshes `~/.fleet/digest/latest.md` and fires a notification only when the
decision-relevant content changed (hash excludes the churning `## Status`
footer). `install.sh`/`fleet-help`/`FLEET.md` wiring is consistent with the
existing per-tool pattern.

### Verified

- **No command-injection surface.** `fleet-notify` passes title/body to
  `osascript` and the PowerShell toast script via environment variables
  (`FLEET_NOTIFY_TITLE`/`FLEET_NOTIFY_BODY`), never spliced into the
  AppleScript/PowerShell source strings — checked the actual `-e`/`-Command`
  bodies, they only ever read `$env:...` / `system attribute "..."`.
  `notify-send "$title" "$body"` passes them as argv, not through a shell
  eval. `test_notify.sh` has a dedicated case injecting
  `quote'break"title` and asserting it reaches the stub byte-for-byte via
  env — good adversarial coverage of exactly the risk this design avoids.
- **Log-first contract holds on every path**: traced `deliver()`'s four
  arms (mac/windows/wsl/linux) plus the unknown-host and missing-binary
  cases — the `notify.log` append always happens before `deliver` is
  called, and `exit 0` regardless of delivery outcome. Test suite covers
  unavailable delivery, missing PATH binary, and a delivery binary that
  exits non-zero — all three still log and exit 0.
- **`fleet-digest-tick`'s `$?` capture is correct**: `digest=$("$decisions_bin"
  ...); if [[ $? -ne 0 ...`  — `$?` here is the exit status of the command
  substitution from the immediately preceding assignment, not a stale
  value; bash sets this correctly for `var=$(cmd)`. The hash-based
  dedup (`awk '/^## Status/{exit}{print}'` before hashing) matches the
  stated "exclude the Status footer" design, and `test_digest_tick.sh`'s
  footer-only-churn case exercises exactly that boundary.
- `resolve_tool`'s PATH-first-then-sibling-directory fallback matches the
  PR body's stated test seam (tests stub via PATH) and works for a bare
  checkout.
- `install.sh`'s new `FLEET_NOTIFY_SRC`/`FLEET_DIGEST_TICK_SRC` wiring
  follows the exact SRC/DEST → chmod-loop-entry → symlink-block pattern
  every other tool in the file uses — no drift.
- `lib_assert.sh` (sourced by both new test files) is pre-existing shared
  test infra, not duplicated — confirmed it's unchanged on `origin/master`.
- Design rationale (per-host, no coordination; scout-triggered variant
  deferred pending untriaged state moving into the scout) is stated
  in the PR body and mirrored into `FLEET.md` — consistent, not just
  asserted.

Not a render/render-adjacent, ECS, or concurrency-bearing change, so none of
the Opus-only escalation criteria apply.

### Nits

None.

### Praise

- The env-var passing for both `osascript` and the PowerShell toast, plus
  the dedicated injection test case, is the right level of paranoia for a
  script that pipes arbitrary fleet-generated text (issue titles, PR
  titles) into two different scripting-language string literals.

- [x] `test_notify.sh` (19), `test_digest_tick.sh` (15) — both green per PR
  body, and I read both suites end-to-end; coverage matches the claimed
  scenarios.
- [ ] One-time `fleet-notify test test` on native-Windows and WSL — already
  flagged unchecked in the PR body (PowerShell toast body untested live on
  those hosts); reasonable to land without it since the routing itself is
  test-pinned and the failure mode is silently-no-toast, not a crash.

Opus recheck not required.

Issue #1920, opened 2026-06-20
render: SDF analytical-solver temporal jitter — curved shapes shimmer under camera Z-yaw


jakildev:
**Model:** opus
**Part of:** epic #1881 (rotated-voxel correctness under camera Z-yaw)
**Blocked by:** (none) — the validation-harness acceptance gate has landed on master.
**Related:** #1922 / PR #1928 (the jitter validation harness — the acceptance gate, now merged). Shares `c_shapes_to_trixel` / `ir_iso_common` with sibling shader children — serialize per #1881's one-at-a-time rule.
**Related:** #1884 (depth unification), #1883 (per-axis scatter) — distinct render paths; validated by the jitter harness child.

## Why (motivation)
On IRShapeDebug, curved SDF shapes (sphere, ellipsoid, cone, torus, cylinder, curved-panel) visibly **vibrate/shimmer** as the camera Z-yaws, while flat-faced **boxes stay stable**. Confirmed in the committed macOS reference `creations/demos/shape_debug/test/references/macos-debug/zoom4_yaw45_inter_cardinal.png`: concentric depth-band rings appear on the ellipsoid and sphere at inter-cardinal yaw and largely vanish at the start cardinal (`zoom4_origin.png`) — the artifact is **angle- and shape-type-dependent**.

Root cause (source-verified): `engine/render/src/shaders/c_shapes_to_trixel.glsl` solves per-tile depth analytically and quantizes via `stableCeilToInt(x) = int(ceil(x - kCeilBiasEpsilon))` (`:109-111`) at ~15 sites across the box/sphere/cylinder/ellipsoid/cone/torus solvers (`:267,282,285,302,317,320,344,361,369,398,419,427,468,502,510`). The fixed `kCeilBiasEpsilon = 1.0e-3` bias cannot absorb FP/FMA-reorder noise near knife-edge (near-tangent) entry distances. Curved surfaces have many near-tangent entries per silhouette, so a depth band flips between adjacent integers frame-to-frame as the continuous yaw sweeps — the rings crawl = visible vibration. Flat box faces have few near-tangent entries, so they are stable.

Distinct from the per-axis scatter seams (#1883) and the merged detached-canvas shimmer fix (#1909) — this is the SDF **analytical** depth-solve path, currently uncovered.

## Scope
[...]

PR #2492, opened 2026-07-21
docs: seed six engine objectives — the standing direction above epics

jakildev:
## Summary
Seeds `docs/design/objectives/` with six draft objectives, each following the tier's README format (Outcome / measurable Done-means / Non-goals / cited Current-state / Progress ledger):

- **render-correctness-under-motion** — all 8 shapes through the jitter gate (4/8 fail today: #1883, #1920), voxel-path table, GL baseline, rotation residuals (picking, cast shadows #1435, waffle #1494).
- **backend-parity** — coverage stays 1:1 (it is today), per-axis lighting for detached rotation on both backends (#1375/#1376), a CI parity guard, the uncaptured GL perf half.
- **machine-verified-behavior** — render refs beyond macOS, gui-verify beyond its two participants, perf gate beyond one demo/backend, acceptance-evidence adoption observable.
- **lua-creation-authoring** — EVAL perf validation, the G1–G4 nav/steering gaps, binding-automation follow-ups, a 100%-Lua demo in the verify culture.
- **editor-authoring** — the #766 five-entities proof, Phase-0 probes, the gui_scale space mismatch, editor in the verify culture.
- **cross-host-shippability** — macOS bundle transitive deps, per-host clean-box package smoke, retiring the Linux-maturity caveat.

## Why
The objectives tier landed with only its README — the sweep has had nothing to sweep, and the upcoming triage role needs a written direction to judge issues against. Every Done-means row here is grounded in what the design docs and harnesses already admit is open (citations in each file's Current state), not invented direction.

## Notes for reviewer / human
**These are drafts for you to edit — merging is the sign-off** that makes them the standing direction, per the README's human-owned lifecycle. Rewrite, cut, or split freely; the one property worth preserving is that every Done-means row stays tree-checkable (the positive-fire bar at direction scale).

## Test plan
- [x] Every cited design doc / harness / cmake artifact exists in the tree.
- [x] No game-repo references (cross-repo isolation check clean).
- [x] Format matches `docs/design/objectives/README.md`; doc-only diff.

### DIFF
--- a/docs/design/objectives/backend-parity.md
+++ b/docs/design/objectives/backend-parity.md
@@ -1,0 +1,44 @@
+# Objective: OpenGL and Metal are functionally interchangeable
+
+**Status:** active
+
+## Outcome
+A feature landing on one render backend is expected — and verified — on
+both. Choosing a host never changes what the engine can render, only how
+fast the developer's inner loop runs.
+
+## Done means
+- [ ] Shader counterpart coverage stays complete: every `c_*` compute and
+  `ir_*` include has an exact 1:1 counterpart; raster `v_`/`f_` pairs map
+  to their bundled `.metal` equivalents; the `backend-parity` audit
+  reports zero unexplained gaps.
+- [ ] Detached (world-placed) rotating entities receive and cast per-axis
+  lighting on both backends (#1375 receive, #1376 cast) — today they
+  composite raw unlit voxel color on both.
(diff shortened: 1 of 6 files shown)


jakildev (commented):

**Verdict:** approve

introduces nor weakens.

lost reservation.

### 5. GLSL ↔ Metal parity

needs-fix.

PR #2500, opened 2026-07-21
editor: runtime --scene-size for IRVoxelEditor (#766 Part 2)

jakildev:
## Summary
- Adds `--scene-size W H D` to `IRVoxelEditor`; the editable grid + origin are now runtime state (`g_editableSceneSize` / `g_editableSceneOrigin`) resolved after the engine arg parse. **Part 2 of #766 (does not close it)** — F-1.6's entities need more than the hardcoded 16³ (the ant is 20³, the tree ~26 tall).
- `deriveSceneOrigin` centres the scene in X/Y and pins the seed ground plane at a fixed world z (`kSeedGroundPlaneWorldZ`), so the ground stays put under the camera as the scene grows taller.
- The Phase 0 (#2489) mapping probes are made size-robust (`probeGroundCell` snaps each cell's z to the live plane) so they assert at any size; the `editor_pick_voxel` baseline is a specific 16³ voxel and is skipped under `--scene-size`. The default 16³ path is behaviourally identical to before.

## Test plan
- [x] `gui-verify IRVoxelEditor` → **13/13** at the default size — no regression to #2489's freshly-merged probe suite.
- [x] `IRVoxelEditor --auto-screenshot 10 --scene-size 20 20 20` → **CLEAN (exit 0)**; logs `Editor scene size: 20x20x20 (origin -10,-10,-16)`; all **8 mapping probes PASS at 20³**; the 16³ pick baseline correctly skipped.

## Notes
- Next in Part 2: erase-fill mode, `SessionBuilder` + `--gui-session`, `scripts/author-entity.py`, then the first entities (rock first). Tracked in `docs/design/editor-authoring-friction.md` §"Part 2".

### DIFF
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -120,8 +120,27 @@ namespace IRVoxelEditor {
 // Scene + palette config (kept as named constants so the editor never
 // inlines a hardcoded dimension — D1 in the T-211 architect direction
 // calls out that the size be configurable per scene).
-constexpr ivec3 kEditableSceneSize{16, 16, 16};
-constexpr vec3 kEditableSceneOrigin{-8.0f, -8.0f, -12.0f};
+//
+// The editable grid dimensions are runtime-configurable via --scene-size W H D
+// (#766 Part 2 — the ant needs 20³, the tree ~26 tall). g_editableSceneSize /
+// g_editableSceneOrigin default to the historical 16³ scene and are overwritten
+// in main() after arg parse. deriveSceneOrigin keeps the scene centred in X/Y
+// and pins the seed ground plane (local z == size.z-1) at world z == 3 for any
+// height, so authoring recipes and probe cells stay height-agnostic.
+constexpr ivec3 kDefaultEditableSceneSize{16, 16, 16};
+// The seed ground plane lives at local z == size.z-1; anchoring it to a fixed
+// world z keeps the camera framing and authoring recipes stable as the scene
+// grows taller (origin.z shifts down by exactly the height increase).
+constexpr int kSeedGroundPlaneWorldZ = 3;
+inline vec3 deriveSceneOrigin(ivec3 size) {
+    return vec3(
+        -static_cast<float>(size.x) / 2.0f,
+        -static_cast<float>(size.y) / 2.0f,
+        static_cast<float>(kSeedGroundPlaneWorldZ - (size.z - 1))
+    );
+}
+ivec3 g_editableSceneSize = kDefaultEditableSceneSize;
+vec3 g_editableSceneOrigin = deriveSceneOrigin(kDefaultEditableSceneSize);
 constexpr int kPaletteCount = 16;
 
 // 16 distinct palette colors. Indexed in row-major order across the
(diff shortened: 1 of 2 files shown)


jakildev (commented):
## Review — editor: runtime --scene-size for IRVoxelEditor (#766 Part 2)

**Verdict:** approve

Reviewed the full diff in file context (`creations/editors/voxel_editor/main.cpp`,
`docs/design/editor-authoring-friction.md`), not just the hunks.

- `--scene-size` registration/read-back follows the established
  `registerArgs()`-before-`init()` / read-back-after pattern (matches
  `creations/demos/canvas_stress/main.cpp`'s `registerArgs()`/`applyArgs()`
  precedent) — `IREngine::args().numbers(...)` before `IREngine::init()`, then
  `getFloats("--scene-size")` after, with the documented all-zero-default
  sentinel correctly gating "flag omitted" (dims default to 0, fail the
  `>= 1.0f` guard, keep `kDefaultEditableSceneSize`).
- `deriveSceneOrigin` checked against the PR body's own worked example:
  size (20,20,20) → origin (-10,-10,-16); formula gives
  `x=-20/2=-10`, `y=-10`, `z=3-(20-1)=-16`. Matches. Default 16³ also
  reproduces the prior hardcoded `-8,-8,-12` exactly — no behavior change at
  default size.
- No leftover references to the old `kEditableSceneSize`/`kEditableSceneOrigin`
  names anywhere in the file (`grep` clean) — every call site
  (loft tool, paint, scene-save command, entity creation, probe assertions)
  was migrated to the runtime globals.
- Math stays on `IRMath` types (`using namespace IRMath` covers `vec3`/`ivec3`
  here) — no raw `glm::`/`std::` math introduced.
- No ECS/render/serialization surfaces touched; this is pure CLI-arg +
  editor-scaffolding wiring.

### Nits
- `creations/editors/voxel_editor/main.cpp:427` (`probeGroundCell`) — only
  the z coordinate is snapped to the live scene bounds; `kProbeMapLocalCells`'
  x/y values go up to 11, so a `--scene-size` run with W or H < 12 leaves the
  probe-map cells outside the grid. The commit message's "made size-robust...
  so they assert at any size" is accurate for height but not footprint. It's
  self-detecting (the `picksIsoColumn` assertion just fails loudly, not a
  crash or silent corruption) and doesn't affect the stated near-term use
  cases (ant 20³, tree ~26 tall — both ≥ 12 in x/y), so not blocking. Worth a
  one-line doc note on the ≥12×12 footprint constraint, or clamping x/y the
  same way z is clamped, before an entity with a narrower footprint lands in
  a later Part 2 slice.

### Praise
- Pinning the ground plane at a fixed world z (`kSeedGroundPlaneWorldZ`)
  rather than re-centering it is a nice touch — keeps camera framing and
  authoring recipes stable as the scene grows taller, and the PR verified it
  against a live 20³ run rather than just asserting it algebraically.

- [x] Already covered by the PR body: `gui-verify IRVoxelEditor` 13/13 at
  default size, plus a clean `--scene-size 20 20 20` run with all 8 mapping
  probes passing.

jakildev:
Addressed the has-nits feedback: clamped `probeGroundCell`'s x/y into the live scene footprint (`IRMath::min(stored, size-1)`), the same way z is already snapped — a `--scene-size` with W or H < 12 no longer pushes probe-map cells off-grid. Both consumers (assertion target + cursor-move pixel) route through the one helper, so they stay consistent. Verified: gui-verify 13/13 at default 16³ (unchanged), and a direct `--scene-size 8 8 20` run passes all 8 probe_map + 4 gui_assert assertions (cells collapse onto in-grid edge cells; each still asserts). `fleet:approved` kept — nit cleanup only. — worker

jakildev (commented):
## Review — editor: runtime --scene-size for IRVoxelEditor (#766 Part 2)

**Verdict:** approve

### Prior-review resolution
- ✅ `creations/editors/voxel_editor/main.cpp:427` — `probeGroundCell` only snapped z to the live scene bounds, leaving x/y unclamped (would push probe-map cells off-grid under `--scene-size` narrower than 12 in x or y) — verified fixed at `8df52bb6`. `IRMath::min(kProbeMapLocalCells[i].{x,y}, g_editableSceneSize.{x,y} - 1)` now clamps both, mirroring the existing z-snap pattern; confirmed `IRMath::min` (not raw `glm::`), matching `int` operand types on both sides, and a no-op at the default 16³ (every stored x/y < 16). Author's own `--scene-size 8 8 20` verification (8 probe_map + 4 gui_assert assertions passing) checks out against the diff.

No new findings — the fix commit touches only `creations/editors/voxel_editor/main.cpp`, and it's the exact single-hunk change the prior nit called for.

- [x] Already covered: `gui-verify IRVoxelEditor` 13/13 at default size, `--scene-size 20 20 20` clean run, and the follow-up `--scene-size 8 8 20` clamp verification.

Issue #2479, opened 2026-07-20
render: per-axis (rotated) shots run-to-run non-deterministic on displaced-voxel scenes at high amplitude


jakildev:
**Area:** engine/render
**Model:** fable
**Blocked by:** (none)

## Context

Found while verifying #2346 (PR #2478, cardinal store winner election). The
cardinal store's displaced-voxel tie-race is fixed there, but the PER-AXIS
(rotated) path shows an analogous run-to-run non-determinism on the same
displaced scene at high wave amplitude — beyond what #2255 (PR #2272) and its
election already guarantee for lattice scenes.

## Confirmed repro (macOS/Metal, master-equivalent config)

```
fleet-run IRPerfGrid --mode voxel_set --wave-freeze --no-overlay --grid-size 32 --zoom 0.8 --wave-amplitude 5 --auto-screenshot 80
```

- amplitude 5: `zoom1_rot` / `zoom4_rot` / `zoom4_rot_pan` produced **10
  distinct hashes across 10 runs** (and 15/15 differing pairs across 6 runs of
  a second config); the 4 cardinal shots were byte-identical in the same runs
  (post-#2346 fix).
- amplitude 1: all 7 shots (rotated included) byte-identical 10/10.
- amplitude 0 (the #2255 acceptance scene): all 7 shots byte-identical 10/10 —
  #2255's guarantee holds on the lattice.

Pre-existing on master: reproduced identically in a config whose
rotated-frame GPU work is byte-identical to master (the #2346 election only
runs inside the `!skipSingleCanvasVoxels` block, which rotating main-canvas
frames skip).

## Suspected mechanism (unverified — needs diagnosis)

The displaced scene's rounded-cell collision population (8,011 colliding
cells at amplitude 5 — see #2346's phase-0 probe) stresses per-axis paths the
#2255 election may not fully cover, e.g. the 4-bit frac quantization ties in
`encodeDepthWithFaceFrac`, or the #2333 view-visibility overflow lane
(atomicAdd append order is run-to-run non-deterministic; equal
`overflowYawedDepthKey` entries at one screen cell may be draw-order
resolved). Possibly related to the θ-stability work in #2427 → PR #2471
(different symptom: temporal jitter across yaw steps vs run-to-run at a fixed
pose) — worth re-measuring after #2471 merges before deep-diving.

## Impact

Same class as #2346's impact, one path over: any harness expecting
byte-identical rotated captures of a heavily-displaced static scene will hit
[...]

PR #2508, opened 2026-07-22
render: one definition for the stage-1/stage-2 mirrored face-select + fog + per-axis store key

jakildev:
## Summary
- **The stage-1/stage-2 "MUST mirror EXACTLY" blocks now have one definition.** New shared include-fragments `ir_voxel_face_select.{glsl,metal}` carry the fog reveal functions (`fogColumnReveal` / `fogColumnRevealNearest`), the face-selection verdict (`selectVoxelFace`: visible-triplet × exposed-mask gate #1278, silhouette-riser flip #2207, dual-emit predicate #2157, fog cut-face widening #2125/#2126/#2127/#2128), and the per-axis store cell+key derivation (`perAxisStoreFacePos`, the #1458 encoding). Both stage bodies — all **five** kernel wrappers (visible / feeder / winner-resolve / stage-2 default / stage-2 winner) — now key their taps off the same definitions, so a one-sided edit desyncing the colour tap from the distance tap is unrepresentable.
- **The documented "impossible" duplication is retired**: the fog functions were duplicated per stage because "slot-parameterized images are not possible in GLSL/MSL" — true at runtime, but the wrapper-`#define` idiom this kernel family already uses makes it a compile-time parameter (`IR_VOXEL_FOG_GRID_BINDING` 0 in the three stage-1 wrappers, 3 in the two stage-2 wrappers). Metal needed no macro (its fog functions already took the texture as an argument — they were pure text duplicates).
- **Intra-body collapse**: each body's per-axis base/subdivided branches were themselves near-identical copies (differing only by the `zIdx != 0` guard) — collapsed to one copy each. Net **−609 lines** across the bodies; the extraction is textual (the compiler inlines), following the established wrapper-over-shared-body idiom (#2258 a′, #2346).
- Body headers' include-order contracts + `engine/render/CLAUDE.md` shared-includes list updated.

## Verification (macOS/Metal, M4 Max — A/B vs base with per-side rebuilds)
| surface | result |
|---|---|
| shape_debug full shot table (28 shots: cardinals, yaw-45 rotating poses, pan pivots, LOD swaps) | **byte-identical** (max_delta 0) |
| fog demo (13 checks vs committed master refs — cut faces, own-column drop, detached edge, SDF blocker, smooth edge) | **all PASS, 100% / max_delta 0** — the moved fog path is byte-exact end-to-end |
| canvas_stress (12 shots) | 11 byte-identical; `so3_offsnap_disc` (yaw=π/4) differs by 8px @ ≤9/255 — **within that pose's own run-to-run noise** (re-running the same binary reproduces the same delta magnitude; the open #2479 class), carries no signal |
| jitter floor (#2469): `--yaw-sweep` zoom 4 voxel cylinder | x rev=2, max_residual=1.01px — **exactly the documented accepted floor** |
| all runs | `ir-run: RESULT=CLEAN` |

## Test plan
- [x] macOS/Metal: suite above (runtime Metal compile validates the MSL twins)
- [ ] Linux/GL smoke: `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10`, ideally also `IRFogDemo` render-verify (validates the GLSL twins — no GLSL validator on the authoring host)

## Notes for reviewer
- GLSL's include resolver is non-recursive, so the fragment is included by the **wrappers** (before the bodies) and the extracted code is parameter-passing functions — the bodies keep their own UBO/SSBO declarations. The Metal fragment includes `ir_iso_common.metal` itself (recursive resolver + visited set — the `ir_per_axis_lighting.metal` idiom).
- `c_voxel_visibility_compact` and `c_fog_to_trixel` keep their own fog variants deliberately (different margins/semantics — permissive cull margin vs exact reveal; documented in their headers). Not folded.
- No screenshots attached: every deterministic surface is byte-identical (tables above quantify the one noise-class delta) — the `render-debug-loop` "provably no runtime effect" carve-out with the proof attached.

### DIFF
--- a/engine/render/CLAUDE.md
+++ b/engine/render/CLAUDE.md
@@ -117,7 +117,10 @@ Naming prefixes follow the convention in
 [`CLAUDE-BASELINE.md §Naming`](../../docs/agents/CLAUDE-BASELINE.md#naming).
 Shared includes: `ir_iso_common.glsl`, `ir_constants.glsl`,
 `ir_world_lighting.glsl` (light-source list layout + SPOT cone +
-ACES tonemap, shared by the world-lighting passes).
+ACES tonemap, shared by the world-lighting passes),
+`ir_voxel_face_select.glsl` (fog reveal + face selection + per-axis
+store key, shared by the stage-1/stage-2 kernel family — the fog grid
+image slot is wrapper-supplied via `IR_VOXEL_FOG_GRID_BINDING`).
 Shader file paths are stored in `render/shader_names.hpp`. Update that
 header when you add or rename a shader.
 
--- a/engine/render/src/shaders/c_voxel_to_trixel_stage_1.glsl
+++ b/engine/render/src/shaders/c_voxel_to_trixel_stage_1.glsl
@@ -19,5 +19,7 @@
 #define IR_FEEDER_PASS 0
 #define IR_STORE_WINNER_ELECTION 0
 #include "ir_iso_common.glsl"
+#define IR_VOXEL_FOG_GRID_BINDING 0
 #include "ir_constants.glsl"
+#include "ir_voxel_face_select.glsl"
 #include "c_voxel_to_trixel_stage_1_body.glsl"
--- a/engine/render/src/shaders/c_voxel_to_trixel_stage_1_body.glsl
+++ b/engine/render/src/shaders/c_voxel_to_trixel_stage_1_body.glsl
@@ -18,8 +18,10 @@
 // ir_sun_shadow_sample.glsl idiom):
 //   #define IR_FEEDER_PASS {0|1}
 //   #define IR_STORE_WINNER_ELECTION {0|1}
+//   #define IR_VOXEL_FOG_GRID_BINDING 0
 //   #include "ir_iso_common.glsl"
 //   #include "ir_constants.glsl"
+//   #include "ir_voxel_face_select.glsl"
 //   #include "c_voxel_to_trixel_stage_1_body.glsl"
 // The three wrappers:
 //   c_voxel_to_trixel_stage_1.glsl        → FEEDER 0, ELECTION 0 = visible dispatch
@@ -202,22 +204,10 @@ layout(std430, binding = 28) buffer PerAxisWinnerScratch {
     uint perAxisWinnerIds[];
 };
 
-// Per-voxel analytic fog clip inputs (#2102), mirroring c_voxel_visibility_compact
-// + c_fog_to_trixel. The world fog canvas binds its 256² grid texture on image
-// slot 0 and uploads the live vision circles at binding 27; every non-fog /
-// detached canvas binds the shared 1×1 all-visible placeholder + a count-0
-// observer buffer, so `fogColumnReveal` short-circuits to "fully visible"
-// and those scenes stay byte-identical. Slot 0 is free in STAGE_1 (it writes
-// only the distance image on slot 1); STAGE_2 rebinds slot 0 to the colour
-// image and reads the fog grid on slot 3 instead.
-const int kFogOfWarHalfExtent = 128;
-const float kFogExploredThreshold = 0.25;
-const int kMaxFogVisionCircles = 8; // mirror of component_canvas_fog_of_war.hpp kMaxFogVisionCircles
-layout(rgba8, binding = 0) readonly uniform image2D canvasFogOfWar;
-layout(std140, binding = 27) uniform FogObserverData {
-    vec4 visionCircles[kMaxFogVisionCircles]; // (centerX, centerY, radius, edgeSoftness)
-    int visionCircleCount;
-};
+// Fog grid + observers + fogColumnReveal/Nearest and the shared face-selection
+// / per-axis store-key math live in ir_voxel_face_select.glsl (the wrapper
+// includes it before this body; STAGE_1's fog grid rides image slot 0 via
+// IR_VOXEL_FOG_GRID_BINDING — free here, the distance image is slot 1).
 
 void writeDistanceTap(const ivec2 canvasPixel, const int voxelDistance) {
     if (!isInsideCanvas(canvasPixel, imageSize(triangleCanvasDistances))) return;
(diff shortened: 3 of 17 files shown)


jakildev (commented):
## Review — render: one definition for the stage-1/stage-2 mirrored face-select + fog + per-axis store key

**Verdict:** approve

### Nits
- `engine/render/src/shaders/ir_voxel_face_select.glsl:114` (and Metal twin `engine/render/src/shaders/metal/ir_voxel_face_select.metal:108`) — `VoxelFaceSelect::rotatedEmit` is computed and stored on the struct but never read by any of the four call sites (stage-1/stage-2 × GLSL/Metal) — only `.keepFace`/`.faceId`/`.riserFlip`/`.bothPolaritiesExposed`/`.fogActive`/`.worldColumn`/`.isCutFace` are consumed downstream. Either make it a local inside `selectVoxelFace` (it's only used to gate `riserFlip`/`bothPolaritiesExposed` internally) or leave a one-line note on why it's exposed for a future caller.

### Praise
- Turning the "MUST mirror stage 1 EXACTLY" duplicated blocks into one shared definition (`ir_voxel_face_select.{glsl,metal}`) with a wrapper-`#define`d fog-grid binding slot is the right fix for the bug class it targets — a one-sided edit desyncing the colour tap from the distance tap is now unrepresentable. It also correctly follows the existing `ir_per_axis_lighting.metal` recursive-include-with-visited-set precedent for the Metal side rather than inventing a new idiom.
- The verification table (28/28 byte-identical shape_debug shots, 13/13 fog-demo checks, and an explicit, measured noise-floor justification for the one differing canvas_stress shot) is thorough for a claimed-mechanical refactor.

- [ ] Linux/GL smoke: `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10` (the author's own test plan already flags this as outstanding — GLSL twin not runtime-validated on the authoring host)

I manually cross-checked every removed duplicated hunk (old stage-1 and stage-2 inline logic) against the new shared `selectVoxelFace`/`perAxisStoreFacePos` definitions and independently dispatched a review-invariant-render pass; both confirm the extraction reproduces the prior logic byte-for-byte (operand order, branch structure, `encodeDepthWithFaceFrac` argument order) and that the per-stage fog-grid binding-slot parameterization (`IR_VOXEL_FOG_GRID_BINDING` 0/3 in GLSL; explicit texture/buffer function arguments in Metal, since MSL has no compile-time binding macro) is wired correctly across all 5 wrapper pairs, with correct include-order (`#define` before `#include`) in every GLSL wrapper. No blockers or needs-fix items found.

Opus recheck required: touches engine/render/ (GLSL/Metal shader-kernel extraction across the stage-1/stage-2 voxel-to-trixel pipeline).

jakildev (commented):
## Opus recheck — render: one definition for the stage-1/stage-2 mirrored face-select + fog + per-axis store key

**Verdict:** approve

Sonnet approved this as a mechanical byte-identity extraction and escalated the one thing a diff read can't settle on its own: whether the new shared `selectVoxelFace` / `perAxisStoreFacePos` / `fogColumn*` definitions actually reproduce the removed inline stage-1/stage-2 logic **exactly** across both backends — the precise desync bug class this refactor exists to eliminate. I confirm Sonnet's read; the extraction is faithful.

**Byte-identity of the extracted math — confirmed (independently re-derived both backends).**
- `selectVoxelFace` reproduces the old inline face-selection operand-for-operand in GLSL *and* Metal: the `rotatedEmit = reVoxelize || (reserved & 4u)` gate, the `faceId ^ 1` silhouette-riser flip (#2207) with `riserFlip = 1`, `bothPolaritiesExposed` (#2157) computed on the **post-flip** `faceId` (ordering preserved), the cut gate `!keepFace && faceId < kFaceZNeg && fogActive` -> `fogColumnReveal(worldColumn + faceOutwardNormal6I(faceId).xy) < 1.0`, and the `worldColumn` detached-canvas recovery. The load-bearing `fogActive` term keeps `perAxisRouteIn <= 2` intact — it was *not* silently reduced to `== 0` (the exact regression the code comment warns about).
- `perAxisStoreFacePos` preserves the helper-call argument order `encodeDepthWithFaceFrac(pos3DtoDistance(facePos), slot, axis, fracInCell, riserFlip)` and `faceMicroPositionFixed6(faceId, worldPos, 0, 0, 1)`. The old base path and the subdivided (`_sub`/`_s2`) copy were textually identical; folding them into one helper plus the single body-side `if (voxelRenderOptions.x != 0 && zIdx != 0) return;` guard preserves all three cases (x==0 every z; x!=0 zIdx==0; x!=0 zIdx!=0 -> return) exactly.
- `fogColumnReveal` / `fogColumnRevealNearest` and every constant (`kFogOfWarHalfExtent=128`, `kFogExploredThreshold=0.25`, `kFogColumnCellHalf=0.5`, `kFogColumnKeepAa=0.5`, `kFogHiddenKeepCells=8.0`) match the removed definitions.

**Call-site contract — confirmed.** Both stage bodies invoke `selectVoxelFace(faceId, reVoxelize, voxels[voxelIndex].reserved, flagsByte, voxelPosition, perAxisRoute, isDetachedCanvas, detachedWorldReceive)` with byte-identical argument order and consume the same verdict fields. The sole asymmetry is intended and documented: stage 1 keeps the own-column drop (`fogColumnReveal(...)<=0` / `...Nearest(...)<=0`) while stage 2 folds `sel.isCutFace` into the entity id (bit 29) — exactly the pre-diff split.

**Binding / include plumbing — confirmed.** The GLSL fog-grid image slot is now `IR_VOXEL_FOG_GRID_BINDING`, `#define`d before `#include "ir_voxel_face_select.glsl"` in every wrapper — 0 on all three stage-1 wrappers (slot 0 free; distance writes slot 1) and 3 on both stage-2 wrappers (slot 0 is the colour output) — matching the old hard-coded bindings. The bodies no longer re-declare `canvasFogOfWar` / `FogObserverData` / the fog constants (no GLSL double-declaration). Metal passes the fog texture + observer buffer as explicit function arguments (MSL has no compile-time binding macro) and pads `FogObserverData` (`_fogObsPad0..2`) to the std140 UBO layout — parity is math-identical to the GLSL twin.

### Nits
- I confirm Sonnet's `rotatedEmit` nit (`ir_voxel_face_select.glsl:114`, Metal twin `metal/ir_voxel_face_select.metal:108`): the field is only read *inside* `selectVoxelFace` to gate `riserFlip` / `bothPolaritiesExposed` — no call site consumes it — so it can be a local rather than a struct member, or carry a one-line "exposed for a future caller" note. Latent-clarity nit, not a merge blocker; fine to fold into the has-nits pass.
- **Metal include idiom (new observation, non-blocking):** `metal/ir_voxel_face_select.metal` omits the self `#ifndef ..._INCLUDED` guard that the cited `ir_per_axis_lighting.metal` precedent carries. Safe as written — the wrappers include it exactly once and its functions are `static` (internal linkage), so neither in-TU re-inclusion nor a standalone glob-compile can produce a duplicate-symbol conflict (`ir_per_axis_lighting.metal` needs its guard only because it uses external-linkage `inline`). Worth a one-line note on *why* it skips the guard so the next editor doesn't "fix" it by switching `static`->`inline` and reintroducing the hazard.

Linux/GL cross-host smoke still owed — Sonnet's `fleet:needs-windows-smoke` tag stands (authored-on-macOS; the GLSL twin is not runtime-validated on the authoring host). No blockers or needs-fix items.

jakildev:
Addressed the has-nits feedback (Sonnet + Opus recheck):

- **`VoxelFaceSelect::rotatedEmit`** — dropped from the verdict struct and made a function-local `const bool` in both the GLSL and Metal `selectVoxelFace` definitions. No call site consumed it; only the `riserFlip` gate and the `bothPolaritiesExposed` predicate read it internally, so the struct now carries only fields downstream stages read.
- **Metal include-guard omission** — added a one-line note in `metal/ir_voxel_face_select.metal` explaining *why* it skips the `#ifndef ..._INCLUDED` self-guard (every function is `static`/internal-linkage and each wrapper includes it exactly once, so no duplicate-symbol hazard; the `ir_per_axis_lighting.metal` precedent needs its guard only for its external-linkage `inline` functions) and warning the next editor not to "fix" it by switching to `inline`.

Both changes are output-neutral (unused-field removal + comments). Validated on macOS/Metal: `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot` → `RESULT=CLEAN`, no shader-compile errors, render unchanged. Pushed as a fast-forward nit-fix commit on top of the reviewed head.

Linux/GL cross-host smoke (`fleet:needs-windows-smoke`) still owed on the GL host — unchanged by this pass. — worker (pool-2)

jakildev (commented):
## Review — render: one definition for the stage-1/stage-2 mirrored face-select + fog + per-axis store key

**Verdict:** approve

### Prior-review resolution
- ✅ `engine/render/src/shaders/ir_voxel_face_select.glsl:114` / Metal twin `engine/render/src/shaders/metal/ir_voxel_face_select.metal:108` — `VoxelFaceSelect::rotatedEmit` stored on the struct but read only inside `selectVoxelFace` (Sonnet nit, confirmed by Opus) — verified fixed at `163c3883`: the field is now a function-local `const bool rotatedEmit` in both the GLSL and Metal definitions, dropped from the struct. Grepped the tree for `.rotatedEmit`/`rotatedEmit` — the only remaining references are the two new locals; no call site ever read the struct field, so this is a pure clarity change with no behavior delta.
- ✅ Metal include-guard note (Opus nit) — `metal/ir_voxel_face_select.metal` omits the `#ifndef ..._INCLUDED` self-guard that `ir_per_axis_lighting.metal` carries — verified fixed at `163c3883`: an 8-line header comment now documents that every function here is `static` (internal linkage) and each wrapper includes the file exactly once, so no duplicate-symbol hazard exists, and explicitly warns a future editor against "fixing" this by switching to `inline`.

Only one commit landed since the prior review pair (`163c3883`, on top of the already-reviewed `ffe37cf7`); both fixes are scoped exactly to what was flagged — no unrelated changes.

### Verification (macOS, this session)
- `cmake --preset macos-debug` + `fleet-build --target IRShapeDebug` — clean build (Metal C++ host code compiles; the only warning is a pre-existing unrelated `sprintf` deprecation in a third-party dep).
- `fleet-run IRShapeDebug --auto-screenshot 3` — `RESULT=CLEAN exit=0`; Metal shader source compiles and runs at runtime (MSL has no separate build-time compile step, so this is the only way to catch a Metal syntax error).
- `fleet-build --target format-changed` — no diff, clean.
- Manually re-derived both edited functions against the diff: the struct-field-to-local change and the comment addition are the only changes; the `rotatedEmit` boolean's value and every downstream use (`riserFlip` gate, `bothPolaritiesExposed`) are byte-identical to before.

No new issues found in the delta. Both backends' `VoxelFaceSelect` struct definitions stay in lockstep (field removed from both), so the GLSL/Metal parity this PR was built to protect is preserved.

Linux/GL cross-host smoke still owed — `fleet:needs-windows-smoke` stands (authored-on-macOS; the GLSL twin is not runtime-validated on the authoring host).

Opus recheck required: touches `engine/render/` (GLSL/Metal shader definitions), per the standing rule for this module even though the delta is a mechanical, already-approved-shape nit fix.

jakildev (commented):
## Opus recheck (nit-fix pass) — render: one definition for the stage-1/stage-2 mirrored face-select + fog + per-axis store key

**Verdict:** approve

I approved the substantive extraction on the prior Opus pass (byte-identity of the shared `selectVoxelFace` / `perAxisStoreFacePos` / fog-column logic against the removed inline stage-1/stage-2 blocks, both backends). This pass reviews only the delta since — commit `163c3883`, which addresses the two nits I raised. Sonnet's re-review flagged the Opus recheck per the standing `engine/render/` rule even though the delta is a mechanical nit fix; I confirm Sonnet's read of the delta.

**Nit 1 — `VoxelFaceSelect::rotatedEmit` dropped from the struct → function-local. Confirmed safe.**
The one question a diff read alone can't settle is whether any consumer *outside* `selectVoxelFace` read the struct field. I grepped the **PR tree** (not master) for member access — `git grep '\.rotatedEmit' 163c3883 -- engine/render/src/shaders/*` returns nothing. In the PR, `rotatedEmit` now appears only as a `const bool` function-local in the two shared definitions (GLSL `ir_voxel_face_select.glsl:156`, Metal `ir_voxel_face_select.metal:146`), read exactly at the `riserFlip` gate and the `bothPolaritiesExposed` predicate. The value expression `reVoxelize || (reserved & 4u) != 0u` is preserved verbatim in both backends, and all three reads switched from `sel.rotatedEmit` to `rotatedEmit`. Pure clarity change, no behavior delta — and the struct stays in lockstep across backends (field removed from both).

**Nit 2 — Metal include-guard note. Confirmed.**
`metal/ir_voxel_face_select.metal` gains an 8-line header comment documenting *why* it omits the `ir_per_axis_lighting.metal` self-guard (all functions `static`/internal-linkage, each wrapper includes once → no duplicate-symbol hazard; the precedent needs its guard only for external-linkage `inline`). No code change. Exactly the "don't let the next editor switch `static`→`inline` and reintroduce the hazard" guard I asked for.

`const`-qualified function-locals are valid in both GLSL and MSL, so no portability concern from the field→local move.

Linux/GL cross-host smoke still owed — `fleet:needs-windows-smoke` stands (authored-on-macOS; the GLSL twin is not runtime-validated on the authoring host). No blockers or needs-fix items.

PR #2509, opened 2026-07-22
render: RAII per-axis lighting route scope + one occupied-cell dispatch loop

jakildev:
## Summary
- **The per-axis lighting-family UBO patch/restore discipline is now RAII.** New `IRPrefab::PerAxisCanvas::LightingRouteScope` flips the shared voxel-frame UBO onto the per-axis decode route (`perAxisRoute_ = 1`) at the #1431-capped store density on construction, and restores the single-canvas state (route 0, uncapped density, compaction slots 25/26) on destruction — the `FrameYawRestoreGuard` idiom (`system_bake_sun_shadow_map.hpp`) applied to the three systems (`COMPUTE_VOXEL_AO`, `COMPUTE_SUN_SHADOW`, `LIGHTING_TO_TRIXEL`) that each hand-rolled the same `subData(offsetof(...))` + tail-restore sequence. A missed restore in that sequence is the #2412-segfault / #1961-corruption bug class; the destructor makes it unrepresentable.
- **The per-axis occupied-cell indirect-dispatch loop has one definition.** New `IRPrefab::PerAxisCanvas::dispatchPerAxisCells(axes, bindAxis)` owns the bindRange-25/26 + `dispatchComputeIndirect` loop (#2256) that AO, sun-shadow, lighting, and the screen-depth resolve each duplicated; each caller now supplies only its pass-specific image binds.
- Net **−150 lines** across the four systems; behavior-identical (same call order: per-axis image binds → cell-list ranges → indirect dispatch; same restore order: route → density → slots → main-canvas images).

## Verification (macOS/Metal, M4 Max — A/B vs base with per-side rebuilds)
| surface | result |
|---|---|
| shape_debug full shot table (28 shots — the yaw-45 rotating poses exercise all four converted per-axis paths) | **byte-identical** (max_delta 0) |
| fog demo (13 checks vs committed refs) | **all PASS, max_delta 0** |
| canvas_stress (12 shots, lit + SO(3)) | 11 byte-identical; `so3_offsnap_disc` within its own run-to-run noise (#2479 class) |
| jitter floor (#2469) zoom 4 | x rev=2, max_residual=1.01px — exactly the accepted floor |
| all runs | `ir-run: RESULT=CLEAN` |

## Test plan
- [x] macOS/Metal: suite above
- [ ] Linux/GL smoke: `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10`

## Notes for reviewer
- The scope stores *references* to the owning system's lazily-resolved buffer members (the `restoreVoxelCompactionSlots` lazy-lookup contract) — it is stack-local inside one tick, so the referents always outlive it.
- Deliberately not converted: the bake's own guards (already RAII), the voxel system's `dispatchPerAxisCanvases` (wholesale per-axis/mode frame re-authoring, a different shape), the resolve's `canvasSizePixels` patch (documented as safe-to-leave-set — different contract; it now uses `dispatchPerAxisCells` for its loop but keeps its own density/slot restores), and the scatter draw loop (indirect *draw*, not compute).

### DIFF
--- a/engine/prefabs/irreden/render/per_axis_canvas.hpp
+++ b/engine/prefabs/irreden/render/per_axis_canvas.hpp
@@ -17,6 +17,7 @@
 #include <irreden/render/components/component_triangle_canvas_textures.hpp>
 
 #include <cstddef>
+#include <cstdint>
 
 namespace IRPrefab::PerAxisCanvas {
 
@@ -149,8 +150,7 @@ inline void setUboSubdivisionDensity(IRRender::Buffer *frameDataUbo, int density
 // caching pattern) and pass them by reference so the named-resource lookup
 // only runs once per system.
 inline void restoreVoxelCompactionSlots(
-    IRRender::Buffer *&voxelCompactedBuf,
-    IRRender::Buffer *&voxelIndirectBuf
+    IRRender::Buffer *&voxelCompactedBuf, IRRender::Buffer *&voxelIndirectBuf
 ) {
     if (voxelCompactedBuf == nullptr) {
         voxelCompactedBuf = IRRender::getNamedResource<IRRender::Buffer>("CompactedVoxelIndices");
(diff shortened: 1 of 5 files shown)


jakildev (commented):
## Review — render: RAII per-axis lighting route scope + one occupied-cell dispatch loop

**Verdict:** approve

Pure code-motion refactor: `LightingRouteScope` (RAII) replaces the hand-rolled
route-flip/density-set at construction + route/density/compaction-slot restore
at destruction across `COMPUTE_VOXEL_AO`, `COMPUTE_SUN_SHADOW`, and
`LIGHTING_TO_TRIXEL`; `dispatchPerAxisCells` replaces the duplicated
bindRange+dispatchComputeIndirect loop across those three plus
`RESOLVE_PER_AXIS_SCREEN_DEPTH`. Traced the call/restore ordering in each of
the four converted sites against the pre-diff sequence — same order in every
case: image binds → cell-range binds → indirect dispatch → barrier → restore.

The one place ordering was worth double-checking was `system_lighting_to_trixel.hpp`:
in the old code, `dispatchOverflowLighting(axes)` runs *between* the per-axis
dispatch loop and the route/density restore, i.e. while `perAxisRoute_` is
still 1 and the density is still #1431-capped (the `#2334` comment says as
much — the overflow relight needs the same per-axis frame state). The new
code preserves this: `dispatchOverflowLighting` is called *inside* the
`LightingRouteScope route(...)` block, before the scope's destructor fires,
so the restore still happens after the overflow relight, same as before.
Confirmed correct.

`LightingRouteScope`'s two reference members (`m_voxelCompactedBuf`,
`m_voxelIndirectBuf`) bind to the caller's own `System<N>` member variables
(lvalues) at each of the three call sites — no dangling-reference risk since
the scope is stack-local within a single tick, as the class comment states.

### Nits
- `engine/prefabs/irreden/render/systems/system_resolve_per_axis_screen_depth.hpp` — this diff drops the file's `#include <cstdint>`, but the file still uses `std::int32_t` at lines 93/101/103 (the seed-buffer code, untouched by this diff). It currently compiles only because some other included header transitively pulls in `<cstdint>` — worth restoring the explicit include so the file doesn't silently break if that transitive path ever changes. (The same include was correctly dropped from `system_compute_sun_shadow.hpp` / `system_compute_voxel_ao.hpp`, where no fixed-width-int usage remains.)

### Praise
- The `#2334` overflow-relight ordering subtlety (noted above) is exactly the kind of thing that's easy to break silently in a refactor — nice catch keeping it inside the RAII scope rather than hoisting the restore earlier for "cleanliness."

- [x] Author already ran macOS/Metal: shape_debug 28-shot table (byte-identical), fog demo 13/13, canvas_stress 11/12 byte-identical (1 known-noisy shot), jitter floor at accepted floor — all `RESULT=CLEAN`.
- [ ] Linux/GL smoke (tagging `fleet:needs-windows-smoke` below per the cross-host protocol — this PR is authored-on-macos and touches `engine/prefabs/irreden/render/`).

Escalation: please Opus-recheck before merge (touches: engine/prefabs/irreden/render/ — GPU dispatch ordering + reference-member lifetime in the new RAII scope).

jakildev (commented):
## Opus recheck — render: RAII per-axis lighting route scope + one occupied-cell dispatch loop

**Verdict:** approve

Sonnet flagged this as a pure code-motion refactor and escalated the two things a
diff read alone can't settle: the new RAII reference-member lifetime and the GPU
dispatch/restore ordering. I traced both to ground and confirm Sonnet's read.

**Reference-member lifetime (`LightingRouteScope`) — confirmed safe.** The two
`Buffer *&` members bind to the owning `System<N>`'s own `voxelCompactedBuf_` /
`voxelIndirectBuf_` lvalues, and every instance is a stack-local `route(...)`
inside a single tick — the referents (system members) strictly outlive the scope,
so no dangling. The class correctly deletes all four copy/move ops, so it can't be
relocated out from under its referents. The dtor's `restoreVoxelCompactionSlots`
mutating those referents through the references is exactly the lazy-resolve caching
the old hand-rolled code did, preserved across ticks. The RAII form is strictly
safer than the pre-diff open-coded patch/restore (a thrown-between-writes or
early-return could previously strand `perAxisRoute_ = 1`; now the restore is
destructor-guaranteed).

**Dispatch + restore ordering — byte-for-byte equivalent at all four sites.** In
each converted site the sequence is unchanged: route=1 + capped density (ctor) →
`bindAxis` image binds → 25/26 bindRange → indirect dispatch → single #1311 barrier
→ route=0 + uncapped density + slot 25/26 restore (dtor). The offset arithmetic in
`dispatchPerAxisCells` (`axis * regionStride * sizeof(uint32)`, indirect stride,
dispatch-args offset) is identical to the four hand-rolled loops it replaces.

**The one place ordering actually mattered — `system_lighting_to_trixel.hpp` —
confirmed correct.** Sonnet noted the #2334 overflow relight must run while
`perAxisRoute_ == 1` and the density is still #1431-capped. On close read: the
`LightingRouteScope` spans the whole outer block, the inner `GpuSubStageScope`
brackets only the 3 cell dispatches, and `dispatchOverflowLighting(axes)` +
`program_->use()` sit *inside* the outer block after the sub-scope closes — so the
scope's dtor (route=0 / density / 25-26 restore) fires only after the overflow
relight, right before the slot-8 (`VoxelActiveMask`) restore and the main-canvas
image rebind. That matches the pre-diff order exactly (25/26 restore before the
slot-8 restore; the two slots don't interact, so even the relative order is inert).
Keeping the restore in the dtor rather than hoisting it for "cleanliness" is the
correct call.

### Nits

- I confirm Sonnet's `<cstdint>` nit on
  `system_resolve_per_axis_screen_depth.hpp`: this diff drops the explicit include
  while the untouched seed-buffer code (lines 93/101/103/105) still uses
  `std::int32_t`. It compiles today only via a transitive include, so it's a
  latent hygiene regression, not a merge blocker — worth restoring on the has-nits
  pass, but not worth a re-review round. (Correctly dropped from
  `system_compute_sun_shadow.hpp` / `system_compute_voxel_ao.hpp`, which no longer
  use fixed-width ints, and correctly *added* to `per_axis_canvas.hpp`, where the
  new `dispatchPerAxisCells` helper uses `std::uint32_t`.)

Linux/GL cross-host smoke still owed (Sonnet's `fleet:needs-windows-smoke` tag
stands — authored-on-macOS, touches `engine/prefabs/irreden/render/`).

jakildev:
Addressed the has-nit: restored the explicit `#include <cstdint>` in `system_resolve_per_axis_screen_depth.hpp` (dropped by the RAII refactor while the untouched seed-buffer code at lines 93/101/103/105 still uses `std::int32_t` — was compiling only via a transitive include). Build clean on macOS/Metal (IRShapeDebug, exit 0). No other change. — worker (pool-4)

jakildev (commented):
## Opus re-review (post-approval commit) — render: RAII per-axis lighting route scope + one occupied-cell dispatch loop

**Verdict:** approve

This PR was already fully Opus-rechecked and approved at `5323acca` (RAII
reference-member lifetime + GPU dispatch/restore ordering both traced to ground
and confirmed). The bot re-tagged `human:re-review` because a single new commit
landed after approval:

- `772e7ca7 render: restore explicit <cstdint> include in per-axis screen-depth resolve`

I diffed the delta since the approved SHA in full. It is exactly the has-nit both
reviewers flagged and nothing else: a lone `+#include <cstdint>` added to
`system_resolve_per_axis_screen_depth.hpp` between `<cstddef>` and `<vector>`
(correct alphabetical slot). Confirmed the include is warranted — the untouched
seed-buffer code still uses `std::int32_t` at lines 94/102/104/106, so the file
was previously compiling only via a transitive include. One insertion, one file,
zero functional change.

Because the delta touches no RAII-scope lifetime, no dispatch/restore ordering,
and no shader/backend code, the prior Opus recheck of the mechanical refactor
carries forward unchanged and stands re-affirmed on `772e7ca7`.

Linux/GL cross-host smoke still owed — `fleet:needs-windows-smoke` stands
(authored-on-macOS, touches `engine/prefabs/irreden/render/`).

🤖 Opus re-review by Claude Opus 4.8 (1M context) (role-opus-reviewer)

PR #2529, opened 2026-07-23
engine/system: populate the dormant SystemName registry; retire prefab wire-once system handles (#2526)

### LINKED ISSUE #2526: engine/system: populate the dormant SystemName registry; retire prefab wire-once system handles
**Area:** engine/system, engine/prefabs/irreden/render
**Model:** opus
**Blocked by:** (none)

## Acceptance criteria

- `SystemManager::m_engineSystemIds` (declared but currently never populated or read) is populated by the enum-templated registration path and exposed via a new `IRSystem::findSystem(SystemName)` free function returning `kNullEntity` when unregistered.
- `IRPrefab::JointTransform::system()` and `IRPrefab::VoxelTransform::allocator()` resolve through `findSystem`; the mutable inline header globals `IRPrefab::JointTransform::g_jointMatrixSystem` and `IRPrefab::VoxelTransform::g_allocatorSystem` are deleted.
- `IRPrefab::JointTransform::setSystem` / `IRPrefab::VoxelTransform::setAllocatorSystem` remain as deprecated no-ops with `// DEPRECATED` markers (engine API removal rule: out-of-tree creations may call them); the six in-tree call sites are deleted.
- `engine/prefabs/irreden/render/CLAUDE.md` gains a `## Deprecated` entry per the deprecation-markers baseline rule.
- `fleet-build` + `fleet-run` clean for `IRShapeDebug` and `IRSkeletalDemo` (the two demo consumers of the wire-once handles) — GPU-transform and skinning behavior unchanged.
- Grep proves no mutable namespace-scope `inline` variable remains under `engine/prefabs/**/systems/`.

## Context

An architecture review of the engine's global-state patterns found `SystemManager::m_engineSystemIds` (`system_manager.hpp`, a `SystemName -> SystemId` map) declared as a member but never populated or read anywhere — while two prefab headers work around the missing registry with mutable namespace-scope `inline` globals (`system_update_joint_matrices.hpp::g_jointMatrixSystem`, `system_update_voxel_positions_gpu.hpp::g_allocatorSystem`), each paired with a manual wire-once setter every consuming creation must remember to call right after `System<N>::create()`.

These are the only mutable header-scope globals in the engine outside the documented manager-global pattern: they have no owner, are never cleared on World teardown, and the wire-once step is caller-delegated bookkeeping (see `.claude/rules/cpp-ecs.md` §"System-owned invariants: encapsulate, don't delegate to callers"). Manager ownership retires both problems at once: the registry dies with the World, and registration self-wires so the setter call disappears.

jakildev:
## Summary

`SystemManager::m_engineSystemIds` (a `SystemName -> SystemId` map) has been declared but never populated or read. Meanwhile two prefab headers worked around the missing registry with mutable namespace-scope `inline` globals plus a manual **wire-once setter** every consuming creation had to remember to call right after `System<N>::create()` — the only mutable header-scope globals in the engine outside the documented manager-global pattern.

This populates the registry and rewrites both handles over it, so **registration self-wires** and the bookkeeping call disappears.

- `SystemManager::recordEngineSystemId` / `findEngineSystem` populate and read the map; the two enum-templated registration paths (`IRSystem::createSystem<N>`, `IRSystem::registerSystem<N, ...>` — the only entry points that have the enum statically) record into it, and `IRSystem::findSystem(SystemName)` reads it back.
- `IRPrefab::JointTransform::system()` and `IRPrefab::VoxelTransform::allocator()` resolve via `findSystem`; `g_jointMatrixSystem` / `g_allocatorSystem` are deleted. Registry ownership also fixes the lifetime hole — it dies with the `World` instead of leaking across it.
- `setSystem` / `setAllocatorSystem` remain as **deprecated no-ops** for out-of-tree creations (engine API removal rule), marked `// DEPRECATED` at the declaration and in a new `## Deprecated` section in `engine/prefabs/irreden/render/CLAUDE.md`. All six in-tree call sites are deleted.

### Two things worth a reviewer's eye

**1. The two registration paths nest.** `createSystem<N>` calls `System<N>::create()`, whose body often calls `registerSystem<N, ...>` — so a *single* registration reports itself from both. `recordEngineSystemId` is therefore idempotent for the same id, and the duplicate-registration `IR_ASSERT` keys on a *different* id (`m_nextSystemId` is monotonic, so a genuine duplicate always has one). Both paths are hooked because not every `System<N>::create()` body goes through `registerSystem`.

**2. `findSystem` is null-manager-safe, and that is load-bearing.** The globals checked their `kNullEntity` sentinel *before* touching the manager; `getSystemManager()` is `return *g_systemManager;`. Routing the handles through it initially **segfaulted `VoxelBoneSlotSeedTest`**, which ticks a `System<N>` against a bare `EntityManager` with no `World`. `findSystem` now answers "no manager" the same as "never registered". Caught by the existing tests, fixed in `a969fba4`, and pinned by a new test.

## Test plan

- [x] `fleet-build` + `fleet-run` clean for `IRShapeDebug`, `IRSkeletalDemo` (the two demo consumers) and `IRVoxelEditor` — all `RESULT=CLEAN`; skeletal rig round-trips OK (30 / 4 / 9 joints).
- [x] **`render-verify --target IRShapeDebug`: 24/24 PASS, `max_delta 0`** (macos-debug).
- [x] **Behavioral A/B against `origin/master` on the path that actually exercises the handles.** Both systems are no-ops in the default scene, so render-verify alone would not prove the handles still resolve. `IRShapeDebug --skin-smoke` (which drives `VoxelTransform::acquireSlot` through `findSystem`) was captured on `origin/master` and on this branch: **28/28 screenshots byte-identical**. Since the setter calls are deleted here, a failed resolve would silently fall back to CPU-direct and the images *would* differ.
- [x] `IrredenEngineTest`: **1359 passed** (1356 pre-existing + 3 new).
- [x] Acceptance grep — no mutable namespace-scope `inline` variable remains under `engine/prefabs/**/systems/`.

## Notes

- **No screenshots attached deliberately.** The diff touches `engine/prefabs/irreden/render/`, but the change is a wiring indirection with no intended visual delta, and the byte-identical render-verify + `--skin-smoke` A/B above are strictly stronger evidence than a before/after pair (which would be identical by construction).
- **`optimize` skipped**: `findSystem` costs one hash lookup and both call sites are cold — `acquireSlot` fires only on first opt-in inside a block-realloc-only path, never per-frame or per-voxel.
- **GL/Windows smoke still owed** — authored and verified on macOS/Metal only.
- Filed **#2540**: `SystemId` 0 collides with the `kNullEntity` "no such system" sentinel (ids count from 0 and `kNullEntity` is 0, so the first system created — `LOD_UPDATE` in `shape_debug` — is indistinguishable from "unregistered"). Pre-existing and harmless for these consumers, but `findSystem` makes it reachable as general API. Not fixed here; the fix has options that want a planning pass.

Closes #2526

### DIFF
--- a/.fleet/plans/issue-2526.md
+++ b/.fleet/plans/issue-2526.md
@@ -1,0 +1,36 @@
+## Plan: engine/system: populate the dormant SystemName registry; retire prefab wire-once system handles
+
+- **Issue:** #2526
+- **Model:** opus
+- **Date:** 2026-07-22
+
+### Scope
+
+Make `SystemManager` the owner of the `SystemName -> SystemId` mapping by populating its existing dormant `m_engineSystemIds` member, and rewrite the two prefab wire-once handles over it. No behavior change for creations beyond deleting the now-unneeded wire-once calls.
+
+### Approach
+
+1. In the enum-templated registration path (`IRSystem::createSystem<SystemName type>` and `registerSystem<N, ...>` — both have the enum statically and funnel into the manager), record `m_engineSystemIds[N] = id`. `IR_ASSERT` on duplicate registration of the same `SystemName` (verified at plan time: no in-tree creation registers a `SystemName` twice; scene transitions re-register *pipelines*, not systems). Dynamic systems (`createSystemDynamic`, Lua-defined) have no enum and are out of scope.
+2. Expose `SystemManager::findEngineSystem(SystemName) const` returning `SystemId` or `kNullEntity`, plus an `IRSystem::findSystem(SystemName)` free-function wrapper in `ir_system.hpp`.
+3. Rewrite `IRPrefab::JointTransform::system()` and `IRPrefab::VoxelTransform::allocator()` to resolve via `findSystem(UPDATE_JOINT_MATRICES)` / `findSystem(UPDATE_VOXEL_POSITIONS_GPU)`; delete `g_jointMatrixSystem` / `g_allocatorSystem`.
+4. Keep `setSystem(SystemId)` / `setAllocatorSystem(SystemId)` as deprecated no-ops: `// DEPRECATED — registration self-wires via SystemManager; remove once out-of-tree creations migrate.` Add the matching `## Deprecated` entry to `engine/prefabs/irreden/render/CLAUDE.md`.
+5. Delete the six in-tree wire-once calls: `creations/demos/shape_debug/main.cpp`, `creations/demos/skeletal_demo/main.cpp`, `creations/editors/voxel_editor/main.cpp` (two each).
+
+### Affected files
+
+- `engine/system/include/irreden/system/system_manager.hpp` (populate map + `findEngineSystem`)
+- `engine/system/include/irreden/ir_system.hpp` (`findSystem` free function)
+- `engine/prefabs/irreden/render/systems/system_update_joint_matrices.hpp`
+- `engine/prefabs/irreden/render/systems/system_update_voxel_positions_gpu.hpp`
+- `engine/prefabs/irreden/render/CLAUDE.md` (`## Deprecated`)
+- `creations/demos/shape_debug/main.cpp`, `creations/demos/skeletal_demo/main.cpp`, `creations/editors/voxel_editor/main.cpp`
+
+### Acceptance criteria
+
+As in the issue body. Verification: `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot`; same for `IRSkeletalDemo`; grep for mutable `inline` namespace-scope variables under `engine/prefabs/**/systems/`.
+
+### Gotchas
+
+- `findSystem` costs one hash lookup; `JointTransform::slotBase` / `VoxelTransform::acquireSlot` are per-rig-operation and per-set-allocation calls, not per-voxel — no hot-path concern. If a caller ever appears in a per-entity tick, cache the resolved pointer in that system's `beginTick`, per the ECS footgun rule.
+- The registry must NOT be cleared by `clearPipeline` — systems are never destroyed and pipelines re-register across scene transitions. It dies with the manager at World teardown, which is exactly the ownership the current globals lack.
+- `SystemId` is an entity-id alias; `kNullEntity` is the established "unwired" sentinel both handles already use — keep it as `findSystem`'s miss value so `system()` / `allocator()` null-return semantics are unchanged.
--- a/creations/demos/shape_debug/main.cpp
+++ b/creations/demos/shape_debug/main.cpp
@@ -497,19 +497,16 @@ void initSystems() {
     // GPU voxel-position prepass (#1396) — writes binding 5 for
     // GPU-transform-indirected voxel sets before STAGE_1 reads it. A no-op (no
     // dispatch) unless a voxel set opts in via gpuTransformSlot_, so the default
-    // scene stays byte-identical. Created up-front so its SystemId can wire the
-    // transform-slot allocator; it keeps its pipeline position below (before
-    // STAGE_1).
+    // scene stays byte-identical. Created up-front so its SystemId can order the
+    // render pipeline below (before STAGE_1).
     const IRSystem::SystemId updateVoxelPositionsId =
         IRSystem::createSystem<IRSystem::UPDATE_VOXEL_POSITIONS_GPU>();
-    IRPrefab::VoxelTransform::setAllocatorSystem(updateVoxelPositionsId);
     // Joint skin-matrix upload (#1603) + per-voxel bone→slot seeding (#1605).
     // Before UPDATE_VOXEL_POSITIONS_GPU so binding 18 holds the skin matrices
     // when the prepass dispatches; a no-op (zero skeletons) unless --skin-smoke
     // rigs a set, so the default scene stays byte-identical.
     const IRSystem::SystemId updateJointMatricesId =
         IRSystem::createSystem<IRSystem::UPDATE_JOINT_MATRICES>();
-    IRPrefab::JointTransform::setSystem(updateJointMatricesId);
     // Captured for the culling minimap's light + caster domains (#2316, V2) —
     // the minimap reads these systems' per-frame gather state back rather
     // than re-running its own light/caster query. Assigned in-place inside
--- a/creations/demos/skeletal_demo/main.cpp
+++ b/creations/demos/skeletal_demo/main.cpp
@@ -104,10 +104,8 @@ void initSystems() {
     std::list<IRSystem::SystemId> renderPipeline = IRPrefab::Camera::standardControlSystems();
     const IRSystem::SystemId updateVoxelPositionsId =
         IRSystem::createSystem<IRSystem::UPDATE_VOXEL_POSITIONS_GPU>();
-    IRPrefab::VoxelTransform::setAllocatorSystem(updateVoxelPositionsId);
     const IRSystem::SystemId updateJointMatricesId =
         IRSystem::createSystem<IRSystem::UPDATE_JOINT_MATRICES>();
-    IRPrefab::JointTransform::setSystem(updateJointMatricesId);
     renderPipeline.insert(
         renderPipeline.end(),
         {
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -2319,14 +2319,12 @@ void initSystems() {
     // UPDATE_VOXEL_POSITIONS_GPU so binding 18 holds the skin matrices when
     // the prepass dispatches. Both are no-ops until a voxel set opts in via
     // gpuTransformSlot_ (the seeded rig does), so an unrigged scene renders
-    // byte-identically. Created up-front so their SystemIds can wire the
-    // slot-allocator handles.
+    // byte-identically. Created up-front so their SystemIds can order the
+    // render pipeline below.
     const IRSystem::SystemId updateVoxelPositionsId =
         IRSystem::createSystem<IRSystem::UPDATE_VOXEL_POSITIONS_GPU>();
-    IRPrefab::VoxelTransform::setAllocatorSystem(updateVoxelPositionsId);
     const IRSystem::SystemId updateJointMatricesId =
         IRSystem::createSystem<IRSystem::UPDATE_JOINT_MATRICES>();
-    IRPrefab::JointTransform::setSystem(updateJointMatricesId);
 
     std::list<IRSystem::SystemId> renderPipeline = IRPrefab::Camera::standardControlSystems();
     renderPipeline.push_front(animPlaybackSystem);
--- a/engine/prefabs/irreden/render/CLAUDE.md
+++ b/engine/prefabs/irreden/render/CLAUDE.md
@@ -618,6 +618,26 @@ sub-attribution is per-dispatch, CPU stays coarse. Sub-scopes are
 single-canvas-exact (last-sample on multi-canvas) and do not cover the
 rotating-only per-axis voxel dispatch.
 
+## Deprecated
+
+| Surface | Replacement | Marked |
+|---|---|---|
+| `IRPrefab::JointTransform::setSystem(SystemId)` | none needed — `IRPrefab::JointTransform::system()` resolves the id itself via `IRSystem::findSystem(UPDATE_JOINT_MATRICES)` | #2526, 2026-07-22 |
+| `IRPrefab::VoxelTransform::setAllocatorSystem(SystemId)` | none needed — `IRPrefab::VoxelTransform::allocator()` resolves the id itself via `IRSystem::findSystem(UPDATE_VOXEL_POSITIONS_GPU)` | #2526, 2026-07-22 |
+
+Both were **wire-once handles**: a creation had to call the setter once,
+immediately after `System<N>::create()`, or the feature silently degraded
+(`allocator()` / `system()` returned `nullptr`, so GPU-transform slots and
+skinning fell back to CPU-direct with no diagnostic). `SystemManager` now owns a
+`SystemName -> SystemId` registry populated by the enum-templated registration
+paths, so **registration self-wires** and the bookkeeping call is gone — see
+`.claude/rules/cpp-ecs.md` §"System-owned invariants: encapsulate, don't delegate
+to callers".
+
+Both setters remain as **no-ops** so out-of-tree creations keep compiling (engine
+API removal rule); every in-tree call site was deleted in #2526. Delete the
+setters once out-of-tree creations have migrated.
+
 ## Gotchas
 
 - **SQT transition complete (T-299/T-300/T-301a/T-302).** All render-side, update-side, and voxel-side readers/writers use `C_LocalTransform` + `C_WorldTransform`. The voxel pool's per-voxel SoA arrays are a dedicated 16-byte `IRRender::VoxelGpuPosition` POD (std430 stride contract). The legacy `C_Position3D` / `C_PositionGlobal3D` / `C_Rotation` components and `SYSTEM_GLOBAL_POSITION_3D` were deleted in T-302; consumers read `C_WorldTransform.translation_` and the SQT quat directly.
--- a/engine/prefabs/irreden/render/systems/system_update_joint_matrices.hpp
+++ b/engine/prefabs/irreden/render/systems/system_update_joint_matrices.hpp
@@ -410,25 +410,23 @@ template <> struct System<UPDATE_JOINT_MATRICES> {
 
 namespace IRPrefab::JointTransform {
 
-// Wire-once-at-init handle to the UPDATE_JOINT_MATRICES skeleton slot blocks —
-// the same shape as `IRPrefab::VoxelTransform` (the entity-slot half of the
-// shared binding-18 budget). A creation that rigs voxel sets calls
-// `setSystem(systemId)` once, right after
-// `System<UPDATE_JOINT_MATRICES>::create()`.
-inline IRSystem::SystemId g_jointMatrixSystem = IREntity::kNullEntity;
-
-inline void setSystem(IRSystem::SystemId systemId) {
-    g_jointMatrixSystem = systemId;
-}
-
+// Handle to the UPDATE_JOINT_MATRICES skeleton slot blocks — the same shape as
+// `IRPrefab::VoxelTransform` (the entity-slot half of the shared binding-18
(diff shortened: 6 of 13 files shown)


jakildev (commented):
## Review — engine/system: populate the dormant SystemName registry; retire prefab wire-once system handles (#2526)

**Verdict:** approve

### Praise

- The nesting analysis (`createSystem<N>` calls `System<N>::create()`, whose
  body often calls `registerSystem<N, ...>` — so a single registration
  reports itself from both) is correct and I traced it concretely:
  `System<UPDATE_JOINT_MATRICES>::create()` (`system_update_joint_matrices.hpp:399-406`)
  calls `registerSystem<UPDATE_JOINT_MATRICES, ...>`, which itself calls
  `recordEngineSystemId`, then the outer `createSystem<N>` wrapper
  (`ir_system.hpp:213-224`) calls it again with the *same* `id`. The
  `try_emplace` + assert-on-mismatch in `recordEngineSystemId`
  (`system_manager.hpp:298-325`) correctly treats that as idempotent while
  still catching a genuine duplicate registration (different id).
- `findSystem`'s null-`g_systemManager` guard (`ir_system.hpp:466-469`) is
  load-bearing, not defensive boilerplate — confirmed
  `test/ecs/voxel_bone_slot_seed_test.cpp`-style headless tests construct a
  bare `EntityManager` with no `World`, so `g_systemManager` is genuinely
  null there. The PR body says this segfaulted `VoxelBoneSlotSeedTest`
  during development and was caught by the existing suite; the new
  `FindSystemNoManagerTest` pins the fix with an explicit
  `ASSERT_EQ(g_systemManager, nullptr)` precondition check rather than
  silently assuming test-ordering isolation.
- Registry ownership genuinely fixes a lifetime hole: the two globals it
  replaces (`g_jointMatrixSystem`, `g_allocatorSystem`) were the only
  mutable namespace-scope `inline` globals in the engine outside the
  documented manager-global pattern, with no owner and no World-teardown
  clearing. `m_engineSystemIds` now dies with `SystemManager`.

### Verification (independently re-run, not just re-stated from the PR body)

- `fleet-build --target IRShapeDebug` — clean, zero warnings from touched files.
- `fleet-run IRShapeDebug --auto-screenshot` — `RESULT=CLEAN exit=0`, 21/21 screenshots saved.
- `fleet-build --target format-changed` — no diff.
- `fleet-build --target IrredenEngineTest` + ran `--gtest_filter="*RegisterSystem*:*FindSystemNoManager*"` — all 8 tests pass, including both new `FindSystem*` cases and `FindSystemNoManagerTest`.
- Grepped `engine/prefabs/**/systems/*.hpp` for remaining mutable namespace-scope `inline g_*` globals myself — zero matches; the two removed here were the only offenders (the one other `g_defaultTheme` hit, `widget_theme.hpp`, lives outside `systems/` and is out of this PR's stated scope).

### Acceptance (issue #2526)

All six acceptance criteria are **met**, cross-checked against the diff directly and via an independent acceptance-grader pass:

| Criterion | Grade |
|---|---|
| `m_engineSystemIds` populated by enum-templated path; `findSystem` exposed | met |
| `JointTransform::system()` / `VoxelTransform::allocator()` resolve via `findSystem`; globals deleted | met |
| Setters remain deprecated no-ops; six in-tree call sites deleted | met |
| `engine/prefabs/irreden/render/CLAUDE.md` gains `## Deprecated` entry | met |
| `fleet-build` + `fleet-run` clean for `IRShapeDebug` / `IRSkeletalDemo` | met — independently verified `IRShapeDebug` myself (see Verification above); `IRSkeletalDemo` not independently rebuilt but the PR body's claim is now corroborated by a live spot-check on the sibling target + a green full test suite subset |
| No mutable namespace-scope `inline` var remains under `engine/prefabs/**/systems/` | met — independently re-grepped |

No unexplained scope drift: the diff matches the plan's affected-files list, plus two reasonable additions (the `engine/system/CLAUDE.md` doc section and the new test file coverage) that are documentation/tests, not a different problem being solved.

### Nits

- PR body's `## Test plan` is a checklist, not a formally-headed
  `## Acceptance evidence` table (criterion | check | observed). Every
  criterion is still independently establishable from the diff + Test
  plan bullets, so this is cosmetic, not a coverage gap.
- The `IR_ASSERT` message in `recordEngineSystemId` (`system_manager.hpp:313-323`)
  calls `getSystemName(slot)` / `getSystemName(id)`, which returns each
  system's per-instance *display* name (e.g. `"UpdateJointMatrices"`), not
  the `SystemName` enum token — accurate and useful for the failure message,
  just worth knowing these are two different "name" concepts if this assert
  ever needs extending.

### Other notes

- Filed follow-up #2540 for the pre-existing `SystemId 0` / `kNullEntity`
  sentinel collision the registry makes reachable as general API — correctly
  scoped out of this PR (options need a planning pass) and disclosed rather
  than silently left.
- GL/Windows smoke is self-flagged as still owed (macOS/Metal-authored) —
  the `fleet:needs-windows-smoke` label isn't on this PR yet; that's a
  process gap for the smoke-tagging step, not a code issue, and doesn't
  block approval.

- [ ] GL/Windows smoke pass once available (author already flagged this as owed)

Escalation: please Opus-recheck before merge (touches: `engine/system/`, `engine/prefabs/irreden/render/`).

Opus recheck required: touches engine/system/ (SystemManager internals, registration path shared by every prefab system) and engine/prefabs/irreden/render/ (prefab handle rewrite) — per the sonnet-reviewer body-requirement list, this surface always gets an Opus pass regardless of confidence.

jakildev (commented):
## Opus recheck — engine/system: populate the dormant SystemName registry; retire prefab wire-once system handles (#2526)

Sonnet's first pass was thorough and I am not re-deriving its checklist
(naming, `shared_ptr`, per-entity `getComponent`, `SystemName` enum
membership, anon namespaces in headers — all spot-checked clean while
reading the diff). This pass spends its budget on the four things Sonnet
could not confirm from the diff alone: **registry staleness across the
manager's lifetime**, **whether the nesting/idempotence argument covers
every registration shape in the tree**, **the new debug-abort surface the
assert introduces**, and **whether the hash lookup actually stays off the
per-voxel path**.

### 1. Registry staleness — confirmed impossible within a manager's lifetime

This is the invariant the whole design rests on and the diff can't show it.
Traced it three ways:

- `m_nextSystemId` is monotonic and reset only at construction
  (`system_manager.hpp:368`, `src/system_manager.cpp:12`, incremented at
  `src/system_manager.cpp:58` and `system_manager.hpp:105`).
- There is **no** `destroySystem` / `removeSystem` anywhere under `engine/`
  — I grepped for both plus `m_systemNames.clear`. Systems genuinely are
  never destroyed, so a recorded id can never dangle or be recycled.
- `replaceSystemBody` (`src/system_manager.cpp:219-226`) swaps the body
  **in place** under the same id; it does not re-register. So the Lua
  hot-reload path can't silently invalidate a registry entry either.

That makes the member comment's claim ("never cleared by `clearPipeline`,
since systems outlive pipeline registration") correct, and makes
"dies with the `SystemManager`" the right ownership. This is a real
improvement over the two globals it replaces, which had no owner at all.

### 2. Nesting/idempotence — holds, including a third registration shape Sonnet's trace didn't cover

Sonnet traced `createSystem<N>` → `System<N>::create()` →
`registerSystem<N, ...>` and concluded the double-record is same-id and
therefore a no-op. I confirm that, and I want to flag a **third** shape
that the analysis didn't mention and that matters for coverage:

`IRPrefab::Camera::standardControlSystems()` (`camera_controls.hpp:45-52`)
calls `IRSystem::System<N>::create()` **directly**, bypassing the
`createSystem<N>` wrapper entirely — so the wrapper's new
`recordEngineSystemId` line never runs for those four systems. They still
land in the registry, but only because their specs route through
`registerSystem<N, ...>` (verified `system_camera_mouse_pan.hpp:53-54` and
its three siblings), which is where the *other* new record call lives.
All five direct-`create()` call sites in the tree are camera systems and
all use that form, so coverage is complete today — but it's complete by
luck of spec style, not by construction. See nit 1.

I also checked the assert can't fire in-tree:

- `LuaScript::registerPrefabSystem<N>()` self-dedups against
  `m_prefabSystemIds` before calling `createSystem<N>()`
  (`lua_script.hpp:93-101`), so the Lua path can't double-create within one
  `LuaScript`.
- The three Lua creations' registered name sets are disjoint from their
  direct `createSystem<...>` calls — checked
  `collision_overlap_demo/main_lua.cpp:80-87`,
  `lua_pipeline_demo/main_lua.cpp:47-52` plus its six
  `registerPrefabSystemId` calls (which exist precisely to avoid
  re-creating the resolver systems), and `lua_widgets/main_lua.cpp:204`
  against `:227-251`.
- No file in the tree creates the same enum system twice (scripted the
  per-file dup check across `creations/` + `engine/`); the multi-hit greps
  are all the component-form `createSystem<C_Camera>` / `createSystem<C_Name>`
  overload, not the enum overload.

### 3. The assert is a *new* debug-abort surface for a pattern two docs currently call merely wasteful

`IR_ASSERT` here is not advisory — `ir_profile.hpp:88` expands to
`IRProfile::engAssert(...)`, which logs at critical **and throws
`std::runtime_error`**. So "same `SystemName` created twice" goes from a
tolerated-but-wasteful pattern to a hard startup abort in every non-
`IR_RELEASE` build. I think that's the right call (it's exactly the
invariant the registry needs), and nothing in-tree trips it — but two
in-tree docs still describe the old, softer consequence:

- `engine/script/CLAUDE.md:1525` — "Because `registerPrefabSystem` and a
  C++ `createSystem` would make two instances, the creation passes the
  registered prefab id into the pipeline"
- `lua_script.hpp:113` — "Calling `registerPrefabSystem<MODIFIER_DECAY>()`
  after that would create a duplicate; this helper records the existing id
  without recreating."

Both now understate it: the duplicate doesn't "make two instances," it
aborts. See nit 2.

### 4. Hot-path claim — verified against the actual call graph, not just the plan's assertion

`allocator()` / `system()` now pay a hash lookup per call where they used
to read a global. Every in-tree consumer is per-operation:

- `VoxelTransform::acquireSlot()` has exactly one engine caller
  (`system_update_joint_matrices.hpp:207`), and it sits behind a
  `voxelSet.gpuTransformSlot_ == kVoxelTransformStatic` guard (`:206`) —
  so at most **once per voxel set, ever**. `seedVoxelBoneSlots` itself
  only runs on block (re)allocation (`:294`).
- The per-voxel loop immediately below it (`:222-228`) reads `block.base_`
  and `voxelSet.gpuTransformSlot_` from locals and never re-enters the
  handle. No lookup on the per-row path.
- `slotBase()` has one in-tree caller and it's a test;
  `releaseSlot()` has none.

The ECS-footgun rule is respected, and the `beginTick`-caching guidance in
both the header doc and `engine/system/CLAUDE.md` is the right pre-emptive
note for future callers.

### 5. Points where I checked rather than assumed

- **`IR_ASSERT` argument evaluation.** `ir_profile.hpp:88` is a plain
  function call, so `getSystemName(slot)` / `getSystemName(id)` are
  evaluated on **every successful registration**, not just on failure.
  Both index `m_systemNames[id]` **unchecked** (`system_manager.hpp:295-296`),
  so I verified the bounds hold: `m_systemNames.emplace_back(...)` precedes
  `m_nextSystemId++` in both registration bodies (`system_manager.hpp:104-105`,
  `src/system_manager.cpp:58`), so any id reaching `recordEngineSystemId` is
  a valid index, and `slot` is by definition a previously-recorded one. The
  calls are side-effect-free, so this also satisfies the macro's documented
  "do not put side-effectful calls in the message" rule and `IR_RELEASE`
  behaves identically. Cost is two vector index + string-ref at init only.
- **Sentinel collision (`kNullEntity == 0` vs ids counting from 0).** The
  PR discloses this and files #2540; I confirm it is **not a regression**.
  The wire-once globals used the identical `kNullEntity` sentinel and the
  identical `== kNullEntity → nullptr` check, so a system landing at id 0
  would have failed to wire under the old code too. I also checked the new
  positive test isn't made vacuous by it: `ASSERT_NE(sysA, sysB)`
  (`register_system_test.cpp:158`) guarantees at least one non-zero id, so
  a `findSystem` stubbed to always return `kNullEntity` still fails the
  test. Good test design.
- **The "dormant member" premise.** Verified on `origin/master`:
  `m_engineSystemIds` has exactly one occurrence tree-wide (its declaration
  at `system_manager.hpp:333`), zero readers and zero writers. No
  pre-existing consumer to conflict with.

### Nits

1. **`engine/system/CLAUDE.md` contract bullet understates the miss case.**
   It reads "Returns `IREntity::kNullEntity` when the name was never
   registered." Per §2 above, population actually requires *either* the
   `createSystem<N>()` wrapper *or* a spec that routes through
   `registerSystem<N, ...>`. A future spec whose `create()` uses the
   component-form `createSystem<Components...>` **and** is invoked as
   `System<N>::create()` directly (the `standardControlSystems()` style)
   would be created-but-unrecorded, and `findSystem` would report
   "never registered" for a system that exists. Worth one clause:
   "…never registered — note that a `System<N>::create()` called directly
   only records if its spec uses `registerSystem<N, ...>`."

2. **Two docs describe the now-asserting duplicate as merely wasteful** —
   `engine/script/CLAUDE.md:1525` and `lua_script.hpp:113` (quoted in §3).
   A half-line each ("…and now trips an `IR_ASSERT`") saves the next person
   a confusing debug abort. Both are outside this PR's stated scope, so
   equally fine as a follow-up.

3. Sonnet's nit about `getSystemName` returning the per-instance *display*
   name rather than the enum token — **confirmed**, and it's the right
   choice for a human-readable failure message. Adding the enum token via
   an existing `SystemName`→string mapping would make the message
   self-diagnosing, but that's polish on an assert that should never fire.

### On the outstanding smoke

Sonnet correctly noted GL/Windows smoke is still owed and that the label
was missing. Tagging `fleet:needs-windows-smoke` as part of this verdict
(macOS-authored → Metal tier covered by the author, OpenGL tier owed,
routed to the ship platform per FLEET-CROSS-HOST-SMOKE §What to tag).

### Verdict rationale

No correctness defect found. The design is sound at the depth Sonnet
couldn't reach: the registry cannot go stale, the idempotence argument
survives all three registration shapes present in the tree, the new assert
is unreachable from any current call site, and the hash lookup genuinely
stays off the per-voxel path. The three nits are documentation-only and
none of them gate merge.

Escalation: none. The one architectural loose end (the `SystemId 0` /
`kNullEntity` sentinel collision that the new general-purpose API makes
reachable) is already disclosed and filed as #2540 rather than silently
left — which is the correct handling.

jakildev:
Addressed the Opus recheck nits:
- `engine/system/CLAUDE.md`: contract bullet now notes a direct `System<N>::create()` call only records if its spec routes through `registerSystem<N, ...>` (nit 1).
- `engine/script/CLAUDE.md`:1525 and `lua_script.hpp`:113-114: both now note the duplicate-registration case trips an `IR_ASSERT` via the SystemName registry, not just "would create a duplicate" (nit 2).
- Nit 3 (`getSystemName` returning display name vs enum token) was confirmed-correct-as-is, no action needed.

Doc/comment-only, no functional change. `fleet-build --target IRShapeDebug` clean, `format-changed` no-op. — worker

jakildev (commented):
## Opus re-review — nit fixups on the `SystemName` registry PR (#2526)

**Verdict:** approve (approval re-affirmed on `9592b03`)

Re-review triggered by the post-approval push. The delta since my last
pass (`a969fba`) is exactly one commit, `9592b03`, touching three files —
two `CLAUDE.md`s and one `//` comment block. **No functional change**, so
this pass is scoped to (a) verifying the delta really is doc-only and
(b) checking the new prose is technically accurate, since inaccurate docs
about an assert are worse than no docs.

### The delta is genuinely doc-only

`engine/script/CLAUDE.md` (+4/-3), `engine/system/CLAUDE.md` (+4/-0), and
`lua_script.hpp` (+2/-1) — the header hunk sits entirely inside the
existing `//` comment block above `registerPrefabSystemId`, so it cannot
affect codegen. Nothing else moved: I re-confirmed `origin/master` is
still `253bee0` (unmoved since the first pass) and that `9592b03`
contains it, so the code verification behind my prior approval —
registry staleness, the three registration shapes, assert reachability,
the per-voxel hot-path claim — is unchanged and stands. No rebuild
warranted for a markdown + comment diff.

### Nit 1 (findSystem contract understated the miss case) — fixed, and the wording is correct

`engine/system/CLAUDE.md` now reads "a `System<N>::create()` called
directly (bypassing the `createSystem<N>` wrapper) only records if its
spec routes through `registerSystem<N, ...>` — direct-`create()` alone
does not populate the registry."

That is exactly the coverage gap I traced through
`IRPrefab::Camera::standardControlSystems()` → `System<N>::create()`
(bypassing the wrapper, landing in the registry only via the spec's
`registerSystem<N, ...>`). Placement is right too — it hangs off the
`kNullEntity`-on-miss bullet, which is the bullet a reader would
otherwise trust to mean "the system doesn't exist."

### Nit 2 (two docs called the duplicate merely wasteful) — fixed at both sites

- `lua_script.hpp:113-115` — "would create a duplicate (and now trips an
  `IR_ASSERT` via the `SystemName` registry)". Accurate: the Lua path
  reaches `createSystem<N>()` → `recordEngineSystemId`, and because
  `m_nextSystemId` is monotonic the second registration necessarily
  carries a *different* id, which is the `slot == id` case the assert
  catches.
- `engine/script/CLAUDE.md:1524-1528` — same correction on the
  `registerPrefabSystem` + C++ `createSystem` sentence. The parenthetical
  reads slightly awkwardly mid-sentence, but it is factually right and
  I'm not spending a round-trip on prose rhythm.

Both now agree with the already-present `engine/system/CLAUDE.md` bullet
("Registering the same `SystemName` twice is an `IR_ASSERT` (debug
only)") and with the `recordEngineSystemId` doc comment, so the three
descriptions of this invariant are consistent across the tree — which
was the point of the nit. The "debug only" qualifier remains correct
(`IR_RELEASE` strips `IR_ASSERT`, and the header comment already
documents the release degradation as last-call-wins).

### Nit 3 — correctly left alone

The author's comment confirms no action; I agreed on the first pass that
`getSystemName` returning the per-instance display name is the right
choice for a human-readable failure message.

### Verdict rationale

All three recheck nits are correctly discharged, the new prose matches
what the code actually does at each of the three sites, and nothing
functional changed. `fleet:needs-windows-smoke` stays as tagged — the
OpenGL/Windows tier is still owed and is unaffected by a doc-only push.

🤖 Opus re-review (role-opus-reviewer)

Issue #2527, opened 2026-07-23
prefabs/render: widget theme becomes a C_WidgetTheme singleton component; retire the g_defaultTheme header global


jakildev:
**Area:** engine/prefabs/irreden/render
**Model:** sonnet
**Blocked by:** (none)

## Acceptance criteria

- The mutable inline header global `IRPrefab::Widget::g_defaultTheme` (`widget_theme.hpp`) is deleted; theme storage moves to a `C_WidgetTheme` singleton component per the "per-world settings" rule in `engine/entity/CLAUDE.md` §"Singleton components".
- `IRPrefab::Widget::defaultTheme()` keeps its exact signature (returns `WidgetTheme&`, now over the singleton row) so theme-mutating creation init code compiles unchanged; its doc comment states it is setup-time-only (not for per-tick reads).
- Every widget system that currently reads `defaultTheme()` inside a tick body (13 files) caches the theme in `beginTick` on its `System<N>` instance — per-tick singleton access is a hash lookup (the ECS footgun).
- Save-trait contract satisfied: `IR_SAVE_OPT_OUT(C_WidgetTheme)` + `AllEngineComponents` tuple entry + inventory-count test bump.
- `gui-verify` passes on the voxel editor GUI shot table; a widget-using demo renders byte-identically.

## Context

An architecture review of the engine's global-state patterns found `g_defaultTheme` is one of only three mutable namespace-scope `inline` variables in engine headers (the other two are the wire-once system handles, tracked separately). Its own comment acknowledges it sidesteps the system-state rule's function-local-static ban by being an inline variable instead. The engine already has a sanctioned home for exactly this shape of state — mutate-once-at-init, world-scoped settings — the singleton-component API (`IREntity::singleton<T>()`), which gives it an owner, teardown participation, and scene-reset semantics (singletons are preserved by `resetGameplay`, matching today's behavior) instead of header-scope storage with no lifecycle.

Save-trait rationale for OPT_OUT: the theme is creation-authored init config, re-established by the creation's setup code on every run; persisting it in a world snapshot would fight the creation's init mutation on load.

jakildev:
## Plan

### Scope

Move widget theme storage from the mutable inline header global to a `C_WidgetTheme` singleton component, keeping the `defaultTheme()` accessor signature stable so creation-side theme customization code is untouched.

### Approach

[...]

PR #2531, opened 2026-07-23
docs: global-state rule catalog + header-global ban checks

jakildev:
## Summary
- New `.claude/rules/cpp-globals.md` — the canonical catalog of the engine's sanctioned global-state patterns (manager globals, engine context, process-infrastructure singletons, singleton components, `thread_local` worker ids, `.cpp`-internal state) and the ban on new mutable namespace-scope variables in headers, with rationale, a detection grep, and the live-deviation list (#2526, #2527).
- `docs/agents/CLAUDE-BASELINE.md` — Style bullet stating the header-global ban + canonical-home map row pointing at the new rule file.
- `simplify` Check 11 — mechanical grep over new header hunks for `inline` / `extern` mutable variables, with the allowlist (module `ir_*.hpp` entry points) and per-kind migration targets; §2b intro's stale "Two convention slips / Both" count fixed in passing.
- `review-pr` — ownership/lifetime checklist bullet so reviewers catch reintroductions the author's simplify pass missed.
- Doc-drift fix in `engine/CLAUDE.md` + `engine/world/CLAUDE.md`: both claimed `World`'s constructor sets and its destructor clears the `g_*` manager globals; in reality each manager self-stamps `this` in its own constructor and clears-if-self in its destructor — `World` contributes the deterministic ordering via member declaration order.

## Test plan
- [x] All new/changed markdown cross-refs verified to resolve (link paths checked against sibling links in the same files; cited headings `§"System-owned invariants…"`, `§"Singleton components"` grepped).
- [x] Detection grep validated against the three known live deviations (`g_jointMatrixSystem`, `g_allocatorSystem`, `g_defaultTheme` all match; manager `extern` declarations fall under the entry-point allowlist).
- [x] Doc-drift claim verified against source: `entity_manager.cpp::EntityManager` ctor stamps `g_entityManager = this;` (mirrored in every manager); no `g_*` assignment exists in `world.cpp`.
- [x] Final-newline check clean on all six files; no build needed (docs + agent-config only).

## Notes for reviewer
The three live deviations named in the rule file are deliberately NOT migrated here — #2526 (SystemName registry) and #2527 (widget-theme singleton) carry the plans, filed and approved in the same session. This PR is the guard-rail + doc-truth half.

### DIFF
--- a/.claude/rules/cpp-globals.md
+++ b/.claude/rules/cpp-globals.md
@@ -1,0 +1,78 @@
+# Global state: sanctioned patterns and the header-global ban
+
+Rule:
+
+> **Never** introduce a new mutable namespace-scope variable in a header
+> (`inline` or `extern`). Every process- or world-scoped mutable object lives
+> behind one of the sanctioned patterns below — each has an **owner**, a
+> **lifecycle**, and an **accessor**. A bare header global has none of the
+> three.
+
+Allowed at namespace scope in headers: `constexpr` / `const` compile-time
+constants. Those are program constants, not state.
+
+## Sanctioned patterns
+
+| State kind | Pattern | Owner / lifecycle |
+|---|---|---|
+| Module manager (process-singular subsystem) | `extern Manager *g_<x>` declared in the module's `ir_<module>.hpp` entry point, defined in the module `.cpp`; the manager's own ctor stamps `this`, its dtor clears-if-self; access via the asserting free function (`IREntity::getEntityManager()`) | `World` owns every manager as a member in dependency order — member order IS the set/clear order |
+| Engine process context | `inline` variables in `engine/include/irreden/ir_engine.hpp` (`g_world`, `g_scriptsDir`, ...), set once in `IREngine::init()`, wrapped by accessors | `IREngine` entry points |
+| Process infrastructure (logger, profiler, CLI args, GL dispatch table, Metal runtime) | Meyers singleton (`static X x; return x;`) or intentionally-leaked `instance()` where shutdown-order robustness demands it (leak documented at the site) | lazy first-use → process lifetime |
+| World-scoped settings / game state (mutate-once config, per-world globals) | singleton component via `IREntity::singleton<T>()` | ECS-owned; preserved across `resetGameplay`, torn down with the world — see `engine/entity/CLAUDE.md` §"Singleton components" |
+| System wiring (find a registered system by name) | the `SystemManager` `SystemName -> SystemId` registry (`IRSystem::findSystem`, #2526) | dies with `World` |
+| Per-thread identity | `thread_local` in a `.cpp` behind an accessor (`IRJob::workerId()`) | thread lifetime |
+| Module-internal state | anonymous-namespace variable in a `.cpp` | translation unit; never a header |
+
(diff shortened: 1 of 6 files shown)


jakildev (commented):
## Review — docs: global-state rule catalog + header-global ban checks

**Verdict:** approve

Docs/agent-config-only PR (no `.cpp`/`.hpp`/shader changes; no build needed).
Checked out the branch and read all six changed files in full, plus
cross-referenced every new citation against the actual source and every
markdown link against the checked-out tree.

### Verified claims
- **Detection grep** (`^\s*(inline|extern)\s+(?!(constexpr|const|void)\b)[^(]*[;={]`
  over `**/*.{hpp,h}`) — ran it by hand against the three cited live
  deviations: matches `g_jointMatrixSystem`
  (`system_update_joint_matrices.hpp:418`), `g_allocatorSystem`
  (`system_update_voxel_positions_gpu.hpp:312`), and `g_defaultTheme`
  (`widget_theme.hpp:76`) as claimed. It also matches allowlisted
  manager-global/engine-context declarations (e.g. `extern EntityManager
  *g_entityManager;`), which is expected and documented — the rule is
  explicit that surviving hits need path-based allowlist + hand
  classification, not zero-false-positive matching.
- **Doc-drift fix** (`engine/CLAUDE.md`, `engine/world/CLAUDE.md`) — grepped
  the actual managers: `entity_manager.cpp:25` stamps `g_entityManager =
  this;` in its own ctor and clears-if-self at `:41-42`; same pattern
  confirmed in `system_manager.cpp`, `render_manager.cpp`,
  `input_manager.cpp`. `world.cpp` has zero `g_*Manager =` assignments. The
  new text ("each manager self-stamps; `World` contributes only the
  deterministic ordering via member declaration order") matches reality; the
  old text (`World`'s ctor/dtor sets/clears them) did not.
- **Markdown link resolution** — every new/changed relative link
  (`.claude/skills/review-pr/SKILL.md` → `../../rules/cpp-globals.md`,
  `.claude/skills/simplify/SKILL.md` → same, `docs/agents/CLAUDE-BASELINE.md`
  → `../../.claude/rules/cpp-globals.md`, `engine/CLAUDE.md` →
  `../.claude/rules/cpp-globals.md`) resolves on disk from its source file's
  directory.
- **`simplify` Check 11 numbering** — sequential, no collision with the
  existing 1–10; the "Two convention slips" → "Convention slips" wording fix
  is consistent (grepped for stale "Two convention slips" elsewhere — none
  found).
- Final newline present on all six changed files.

### Scope
The PR body is explicit that the three live deviations are guard-rail-only
here, migrated by the separately-filed #2526/#2527 — consistent with the
diff, which touches no `.hpp`/`.cpp` files.

### Nits
(none)

### Praise
- The doc-drift fix is a genuine correctness improvement, not just
  rewording: the old text actively mispointed a future reader at `World`'s
  ctor/dtor for the set/clear logic, which doesn't exist there.
- Detection grep is designed defensively (over-match + allowlist + hand
  classification) rather than chasing an unattainable zero-false-positive
  regex — the right tradeoff for a mechanical simplify-pass check.

- [x] All test-plan items in the PR body are the kind independently
      re-verifiable from a docs-only diff, and I re-verified them directly
      rather than trusting the checkboxes — see Verified claims above.

PR #2535, opened 2026-07-23
editor: F-1.6 Part 2c — authoring SessionBuilder + --gui-session (#766)

jakildev:
## Summary
- **`SessionBuilder`** (`creations/editors/voxel_editor/session_builder.hpp`) — compiles a recipe of editor gestures into the scripted-input stream the GUI-test harness replays against the live UI. A recipe names cells; the builder picks which face of which already-placed voxel to click, aims with Phase 0's `worldPos3DToMouseScreenPx`, and emits MOVE / PRESS / RELEASE. No recipe touches voxel storage — every voxel lands because a scripted click ran the editor's own place/erase path, which is what F-1.6 is actually testing.
- **Shadow occupancy model** — mirrors the editable set as the recipe grows it and replays `castVoxelRay`'s front-to-back walk over that mirror, so an occluded aim is caught at recipe-build time. An unaimable gesture aborts the run naming the cell rather than replaying a partial stream (a session that silently drops an op authors the wrong entity and saves it anyway).
- **`--gui-session <name>`** + a session registry (`sessions.hpp`); `gui-verify.py` forwards target args after a `--` separator so a session runs through the standard runner.
- **`AssertKind::PREDICATE`** (`gui_test_assertions.hpp`) — a creation-supplied `bool(context, actual)`, so creation-specific state checks reuse the harness's single `GUI-ASSERT` emitter. The Part 2b erase probe moved onto it, retiring the one hand-rolled emitter in the tree.
- **`drag_probe` session** — walks the four gestures every entity recipe is built from (click-place, left-drag box fill, V erase toggle, carve click) plus Ctrl+S, each with live-occupancy assertions. **This resolves P0-3**, the drag stroke deferred out of Phase 0.

## Test plan
- [x] `python3 scripts/gui-verify.py IRVoxelEditor -- --gui-session drag_probe` → **12/12**, exit 0 (macOS/Metal)
- [x] Same session at `--scene-size 20 20 20` → **11/11** (recipe derives cells from live scene dims, so the ant's 20³ framing is covered)
- [x] Standing shot table unaffected: `python3 scripts/gui-verify.py IRVoxelEditor` → **14/14**
- [x] Other consumers of the changed assertions header still build: `IRLuaWidgetsDemo`, `IRLightingAO`
- [x] `ruff check scripts/` clean; `format-changed` clean

## Findings (measured, recorded in the friction log)
| Finding | How it was measured | Result |
|---|---|---|
| The demo furniture makes the default scene unauthorable | ran the same recipe against the stock scene | **every** gesture swallowed — 0/4 drag cells, aim reports `voxel=(-1,-1,2)` (a *shape* hit). Shapes are tested before voxel sets in the picking walk and a shape hit carries no face normal, which the place/erase driver drops. Sessions build without the furniture; same recipe then passes. |
| Face-accurate aiming needs zoom ≥ 3 | swept `kSessionZoom` against the session | zoom 2 → 2 FAIL (side-face aim lands on the neighbouring column, `voxel=(0,1,3)` vs target `(0,0,2)`); zoom 3 / 4 / 8 → 11/11. Sessions author at **zoom 4**. Phase 0's zoom ≥ 2 floor only covers the *column*. |

The first one is a real editor-usability defect, not a harness artifact: a human clicking the ground in the shipped editor is clicking through a slab that eats the event. Filed as a child issue.

## Notes for reviewer
- **No screenshots / render-debug-loop.** The diff touches `engine/prefabs/irreden/render/gui_test_assertions.hpp`, which is in the visual-file glob, but the change is additive headless assertion vocabulary with no render-pipeline effect — verified by the standing table staying 14/14 and both other consumers of that header building clean.
- **`optimize` skipped deliberately**: nothing here runs on a shipping path. The occupancy model runs once at startup; `onSessionAssertFrame` only exists under `--gui-session`.
- Simplify's call-sequence scan flagged three duplication candidates I deliberately did **not** fold in, as each is a cross-module refactor beyond this slice: `OccupancyModel::setBox` shares the unordered-corner + `iterateAABB` idiom with `applyFillAABB` (and two other sites) and wants a shared `IRMath` helper; the bounds-guard + `index3DtoIndex1D` + flat-index-check sequence appears three times; and `onSessionAssertFrame` shares its tail dispatch with the probe-map branch beside it.

Part 2c of #766 — does **not** close it. Next slice (2d): `scripts/author-entity.py` runner + the ROCK entity.

### DIFF
--- a/creations/editors/voxel_editor/main.cpp
+++ b/creations/editors/voxel_editor/main.cpp
@@ -98,6 +98,10 @@
 // Symmetry modes (T-212)
 #include "symmetry.hpp"
 
+// Authoring sessions (#766 Part 2c) — recipes of editor gestures compiled into
+// scripted input and replayed against the live UI by the GUI-test harness.
+#include "sessions.hpp"
+
 // Scene save/load
 #include "scene_io.hpp"
 // Rig save/load (joint entities ↔ .rig asset)
@@ -261,6 +265,15 @@ BonePaintState g_bonePaint;
 // a free global avoids threading a pointer through SystemParams just to
 // reach the same address every frame.
 AnimationState g_anim;
+
+// Authoring session selected by --gui-session (#766 Part 2c). When one is
+// active it replaces the standing GUI-test shot table with the recipe's
+// segments, and the scene is built without the demo furniture (see
+// initEntities). Both live at module scope because main() resolves them right
+// after the arg parse, initSystems hands the shots to the harness, and the
+// per-frame assert callback reads them back.
+Session::Id g_sessionId = Session::Id::NONE;
+Session::Recipe g_session;
 
 namespace {
 
@@ -452,20 +465,18 @@ inline IRMath::ivec3 probeGroundCell(int i) {
 
 // --- Part 2b (#766) erase-fill mode probe ---------------------------------
 // Verifies the erase-fill toggle through the live UI, occlusion-free: synthetic
-// V flips g_eraseMode ON, and the capture-frame assertion (emitted in
-// onGuiAssertFrame) checks the fill-mode status label the place/erase system
-// repaints each frame now reads "ERASE BOX". This exercises the whole control
-// path — synthetic key → command dispatch → g_eraseMode → status label —
+// V flips g_eraseMode ON, and a capture-frame PREDICATE assertion
+// (evaluateEraseModeLabel) checks the fill-mode status label the place/erase
+// system repaints each frame now reads "ERASE BOX". This exercises the whole
+// control path — synthetic key → command dispatch → g_eraseMode → status label —
 // without a scene click, so it is immune to the seed scene's occlusion.
 //
-// Why not a scripted-erase-removes-a-voxel check: the only clean, static,
-// click-erasable surface would be the editable ground plane, but the posed
-// starter rig (a skinned 31×3×3 bar) blankets the central ground in iso, and
-// skinned voxels are not click-erasable (the world→local mapping fails on them).
-// Reliable scripted authoring needs the SessionBuilder's occupancy-model aiming
-// (Part 2c); that slice adds the functional carve/erase asset checks. The two V
-// key events carry no pixel; the shot leaves erase mode ON (harmless — the
-// trailing save / A-D probes don't read it).
+// Why not a scripted-erase-removes-a-voxel check here: the standing shot table
+// runs against the demo scene, whose reference shapes and posed starter rig sit
+// between the camera and the editable ground plane. The functional carve check
+// lives in the Part 2c `drag_probe` session instead, which authors on a scene
+// built without that furniture. The two V key events carry no pixel; the shot
+// leaves erase mode ON (harmless — the trailing save / A-D probes don't read it).
 constexpr IRVideo::GuiInputEvent kProbeEraseEvents[] = {
     {0,
      IRVideo::GuiInputEvent::Type::PRESS,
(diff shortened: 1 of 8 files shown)


jakildev (commented):
## Review — editor: F-1.6 Part 2c — authoring SessionBuilder + --gui-session (#766)

**Verdict:** approve

Reviewed `session_builder.hpp` (new `OccupancyModel` + `Builder`), `sessions.hpp`
(new `drag_probe` recipe), the `main.cpp` wiring (`--gui-session`, session-vs-
standing-shot-table branching, `evaluateOccupancyCheck`), and the new
`AssertKind::PREDICATE` addition to `gui_test_assertions.hpp`.

### Ownership / lifetime
- `Recipe::checks_` uses `std::deque<OccupancyCheck>` specifically so
  `Builder::expectOccupancy`'s `&m_recipe.checks_.back()` context pointer
  (`session_builder.hpp:796-802`) survives further `push_back`s — correct choice,
  called out in a comment.
- `resolveShots()`'s raw pointers into `segment.label_.c_str()` /
  `segment.events_.data()` (`session_builder.hpp:672-683`) are only safe because
  `Recipe` is always *moved* (never copied) into `g_session` before
  `resolveShots` runs, and `segments_` never grows/reallocates afterward — the
  comments document this ("treat the segments as frozen afterward") and the
  `main.cpp` call order (`g_session = build(...)` then `resolveShots(g_session)`,
  `main.cpp:219-231`) actually honors it.
- Dispatched a focused render-invariant pass on the `PREDICATE`
  `Assertion`/`predicate_`/`context_` addition specifically for GPU/std140 or
  cross-boundary size implications — confirmed `Assertion` is pure host-side
  state with no such consumer, and the "outlives the run" contract holds for
  the one caller (`evaluateOccupancyCheck` context is the deque-backed
  `OccupancyCheck`, function pointer is a free function). No issue found.

### Naming / style
- `m_`/trailing-`_` convention followed correctly throughout the new files
  (private `OccupancyModel`/`Builder` members vs. public `Segment`/`Recipe`/
  `OccupancyCheck` fields).
- Cited precedents checked against the tree: `kPickingDepthStep`,
  `castVoxelRay`, `RayHit::faceNormal_` (picking.hpp), and
  `set.voxels_[i].color_.alpha_` liveness (matches
  `face_occupancy.hpp`'s `voxel.color_.alpha_ > 0`) — all real, all used
  correctly.

### Scope / churn
- Single commit ahead of `origin/master` (`git merge-base` confirms); the other
  commit shown in `git log` is already on master. No out-of-scope files.
- PR body explicitly states this does **not** close #766 — no acceptance
  grader dispatch needed.

### Praise
- The shadow `OccupancyModel` replaying `castVoxelRay`'s own front-to-back walk
  (rather than approximating occlusion) is a good design call — it means "is
  this aim occluded" is answered identically to how the real editor will
  answer it, instead of drifting from the picker's actual behavior over time.
- F-2c-1 (the demo floor slab silently swallowing ground clicks) is a
  genuinely useful finding, measured rather than inferred, and correctly
  routed to its own follow-up issue instead of being bundled into this PR's
  scope.

- [x] Already covered per PR body: `drag_probe` 11/11, standing table 14/14,
      `--scene-size 20 20 20` 11/11 (all macOS/Metal).
- [ ] This PR touches `engine/prefabs/irreden/render/gui_test_assertions.hpp`,
      so it needs one OpenGL-tier smoke pass before merge (tagging
      `fleet:needs-windows-smoke` below) — a clean build is sufficient since
      the change is additive and headless.

PR #2516, opened 2026-07-22
test: refresh shape_debug macos-debug render-verify references (#2510)

jakildev:
## Summary
- Refreshed the 2 stale `macos-debug` render-verify reference PNGs for `IRShapeDebug` (`zoom4_yaw45_inter_cardinal`, `zoom4_pan16_yaw45_pivot`) that went stale after #2471's intentional overflow-membership change at inter-cardinal yaw-45 poses.
- Verified master-vs-master A/B byte-identity on all 28 shots before updating — this is a stale-reference refresh, not a regression fix.

## Acceptance evidence

| Criterion | Check run | Observed |
|---|---|---|
| `render-verify.py --target IRShapeDebug --update-references` run on a macOS host at current master; the two changed PNGs visually inspected (expected delta: #2471 yaw-45 overflow drift at yaw-45 only) and committed | `python3 scripts/render-verify.py --target IRShapeDebug --update-references` (macos-debug host — `fleet:authored-on-macos`) | Diff touches exactly the two named PNGs, binary-only: `zoom4_pan16_yaw45_pivot.png` 191061→190886 B, `zoom4_yaw45_inter_cardinal.png` 182488→182367 B. Small localized silhouette-edge deltas, not a wholesale re-render. Master-vs-master A/B confirmed byte-identical on all 28 shots first, proving stale-ref not regression. |
| `render-verify --target IRShapeDebug` passes 24/24 on master afterward | `python3 scripts/render-verify.py --target IRShapeDebug` (re-run this iteration against the committed refs) | `[render-verify] all 24 checks PASS` — 21 full-frame shots + 3 structural gates (coverage / silhouette / clip). Both refreshed shots `zoom4_yaw45_inter_cardinal` and `zoom4_pan16_yaw45_pivot` PASS at 100.0% / 0 delta, confirming the committed refs match the current master build. |

Closes #2510

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2514, opened 2026-07-22
render: make the GLSL include resolver recursive (visited-set) to match Metal's


jakildev:
**Area:** engine/render (shader tooling)
**Model:** sonnet

## Context

The two backends' shader `#include` resolvers disagree: Metal's runtime preprocessor (`metal_pipeline.cpp::loadAndPreprocessMetalSource`) is recursive with a visited set, so a Metal fragment can include its own prerequisites (`ir_per_axis_lighting.metal` → `ir_iso_common.metal`); GLSL's (`opengl_shader.cpp detail::resolveShaderIncludes`) pastes verbatim with no re-scan, so a GLSL fragment must list NO includes and every kernel wrapper must hand-maintain the full prerequisite chain in exact order (the documented contract in `c_voxel_to_trixel_stage_{1,2}_body.glsl` and `ir_voxel_face_select.glsl`). Each new shared fragment (#2485's `ir_world_lighting`, #2508's `ir_voxel_face_select`) grows that per-wrapper boilerplate.

Making the GLSL resolver recursive with a visited set (mirroring the Metal implementation) is provably a no-op on the current tree — no GLSL fragment contains an `#include` line today, so every resolved source string is byte-identical — while removing the asymmetry for future fragments.

## Acceptance criteria

- `resolveShaderIncludes` resolves nested `#include "..."` recursively, suppressing duplicates via a canonical-path visited set (mirror the Metal shape, incl. the cycle guard).
- Provable no-op on the current tree: for every `.glsl` program in `shader_names.hpp`, the resolved source string before/after the change is byte-identical (a one-off assertion run or unit test — the resolver is a pure function over files).
- Validated on a Linux/GL host: `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10` clean, plus one fog demo run (`IRFogDemo` render-verify) since the fog wrappers carry the deepest include chains.
- The body-header include-order contracts (`c_voxel_to_trixel_stage_{1,2}_body.glsl`, `ir_voxel_face_select.glsl`) updated to note fragments MAY now self-include prerequisites (keep the existing wrapper chains as-is — no churn).

## Context note

Authoring host is macOS (Metal) — GLSL only compiles on GL hosts, hence the Linux validation requirement.

jakildev:
## Plan: recursive GLSL include resolver

- **Issue:** #2514
- **Model:** sonnet
- **Date:** 2026-07-22

[...]

PR #2537, opened 2026-07-23
render: residual shader dedup — resolve emit, feeder predicate, Metal lighting UBOs, clear-volume constant (#2513)

### LINKED ISSUE #2513: render: residual shader dedup — resolve-kernel cardinal-layout emit, feeder predicate mirror, Metal lighting UBO structs, clear-volume constant
**Area:** engine/render (shaders, both backends)
**Model:** opus
**Blocked by:** #2508

## Context

The #2508 face-select extraction closed the biggest stage-1/stage-2 mirrors. Four smaller documented duplicates remain (catalogued during the same audit):

1. **Resolve-kernel cardinal-layout emit**: `c_resolve_per_axis_screen_depth.{glsl,metal}` and `c_resolve_world_placed_depth.{glsl,metal}` are structurally the same kernel — decode input → `rotateCardinalZ` + `cardinalLowerCornerShift` → `encodeDepthWithFace(pos3DtoDistance(viewPos), slot, flip)` → scale² micro-cell sweep → 2-pixel `faceOffset_2x3` diamond `atomicMin` — differing only in the input decode (per-axis frac vs world-placed model+offset) and empty sentinel. Extract the shared emit into a fragment (the `ir_voxel_face_select` idiom; wrapper-include on GLSL, self-include on Metal).
2. **Shadow-feeder predicate mirror**: `c_voxel_visibility_compact`'s feeder classifier and stage-2's feeder skip re-derive the same cardinal-iso-outside-visible-bounds predicate (`residualYaw==0 && !detached && iso outside visibleIsoBounds`) — one definition (candidate home: `ir_voxel_face_select`), so the classification and the skip cannot drift.
3. **Metal lighting UBO structs**: `FrameDataLightingToTrixel` and `LightVolumeParams` struct definitions are duplicated between `metal/c_lighting_to_trixel.metal` and `metal/c_light_overflow_faces.metal` — fold into `ir_world_lighting.metal` (which both already include). Note the seed/propagate kernels' `LightVolumeParams` uses live field names where lighting uses tombstones — reconcile or leave those two out, with a comment.
4. **`c_clear_light_volume.{glsl,metal}` `const int kLightVolumeSize = 128`** — int variant of the shared float constant in `ir_world_lighting`; fold or document (noted in PR #2485).

## Acceptance criteria

- Each item lands as textual extraction (no logic restructure); the #2508 verification suite green: shape_debug 28-shot A/B byte-identical vs base, fog demo refs 13/13, canvas_stress within #2479 noise, jitter floor (#2469) exact, all runs RESULT=CLEAN.
- Both backends stay in lockstep; Linux/GL smoke for the GLSL twins.
- Tree-wide grep confirms one definition per hoisted symbol afterward.

jakildev:
## Summary
Four documented duplicates left over from the #2508 audit, each a textual extraction with no logic restructure.

- **Resolve-kernel cardinal-layout emit** → new `ir_resolve_cardinal_emit.{glsl,metal}`. `c_resolve_per_axis_screen_depth` and `c_resolve_world_placed_depth` decode different inputs but end in the same emit (encode depth key → iso pixel → `atomicMin` into the 2-pixel `faceOffset_2x3` diamond). Region axis is a parameter, not derived from the slot: the per-axis store is already view-frame, the world-placed store is model-frame and rotates its normal first. GLSL can't pass an SSBO as an argument, so the GLSL fragment writes through the wrapper's `resolveScratch` and is included after it; the Metal twin takes the pointer as a parameter.
- **Shadow-feeder predicate** → `isShadowFeederIso` in `ir_iso_common.{glsl,metal}`, consumed by the compact's Step-B partition (#2258) and stage 2's depth-only skip (#1740). The route terms live inside the predicate since the widened cull only exists on that route; stage 2 keeps its enclosing gate as the cheap early-out.
- **Metal lighting UBO structs** → `FrameDataLightingToTrixel` + `LightVolumeParams` folded into `ir_world_lighting.metal`. The live field names win (lighting/overflow never named the four they tombstoned), which lets the seed drop its copy too.
- **Clear-volume constant** → documented, not folded (see Notes).

## Test plan
- [x] macOS/Metal: full #2508 suite below (runtime Metal compile validates the MSL twins)
- [ ] Linux/GL smoke: `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot 10`, ideally `IRFogDemo` render-verify — validates the GLSL twins, which have no validator on the authoring host

## Acceptance evidence
|---|---|---|
| shape_debug 28-shot A/B byte-identical vs base | staged-shader-dir swap A/B (identical binary — diff is shaders-only), `md5` per shot | **28/28 byte-identical**; re-confirmed on the final committed tree |
| fog demo refs 13/13 | `render-verify.py --target IRFogDemo` | **13/13 PASS**, 100% match, max_delta 0 |
| canvas_stress within noise | `render-verify.py --target IRCanvasStress`, mine vs master shaders | **zero delta from this PR** — 8/10 FAIL identically on master's shaders (same match% to 4 dp); captures byte-identical mine vs master 12/12. Pre-existing stale refs, see Notes |
| jitter floor (#2469) exact | `--yaw-sweep` zoom 4 voxel cylinder → `jitter_probe` | **x rev=2, max_residual=1.01px** — exactly the documented floor |
| all runs RESULT=CLEAN | every run above | `ir-run: RESULT=CLEAN` throughout |
| one definition per hoisted symbol | tree-wide grep | `emitResolveCardinalDiamond` 1/backend; `isShadowFeederIso` 1/backend; `FrameDataLightingToTrixel` 1; `LightVolumeParams` 2 (shared + documented propagate copy); `kLightVolumeSize` documented pair |
| both backends in lockstep | every GLSL edit has its MSL twin | 4 twin pairs + 1 new fragment pair |

**Enabled-path proof (byte-identity alone would be vacuous here).** Both extracted symbols sit on paths gated off at cardinal / shadows-off, so I verified each actually executes by perturbing it in the staged shader dir and confirming output moves:
- `emitResolveCardinalDiamond` +4096 on the depth key → canvas_stress metrics shifted on all 8 rotating shots (e.g. `so3_offsnap_disc` 98.7363 → 98.4818).
- `isShadowFeederIso` bounds test inverted → perf_grid **7/7 shots changed**. (canvas_stress does *not* reach this path — inverting it there changed nothing — so the perf_grid scene is the one carrying this coverage. perf_grid needs `--wave-freeze` to be run-to-run deterministic; without it consecutive runs differ regardless of shaders.)
- perf_grid A/B on that same frozen feeder-exercising scene: **7/7 byte-identical** mine vs master.

## Notes for reviewer
- **No screenshots** — the `engine/render/CLAUDE.md` "internal refactors with provably no runtime effect" carve-out, with the proof attached above (four harnesses byte-identical). Same call #2508 made for the same class of change.
- **canvas_stress refs are stale on master, independent of this PR.** `origin/master` alone fails 8/10 checks with byte-for-byte the same metrics as this branch. Most likely #2509 (RAII per-axis lighting route) shifted output while #2516 refreshed only the shape_debug refs. Filing a ref-refresh follow-up; happy to fold it in here instead if preferred.
- **Item 2 deviates from the issue's suggested home.** `ir_voxel_face_select` was proposed, but `c_voxel_visibility_compact` declares its own `canvasFogOfWar` (binding 0) and `FogObserverData` (binding 27) — the same bindings that fragment declares — so including it there collides. `ir_iso_common` is already included by both call sites and needs no new include.
- **Item 3's plan caveat didn't trigger.** The plan said to leave seed/propagate out *if* the tombstone-vs-live naming couldn't merge cleanly. It merges cleanly (lighting/overflow reference only `worldOriginVoxel`), so the seed folds too. `c_propagate_light_volume` is the deliberate exception — it's the one consumer that doesn't include the fragment, and pulling the light list + SPOT/ACES helpers into a kernel dispatched 32× a frame to share a struct declaration isn't worth it. Both sides carry a pointer comment.
- **Item 4 documents rather than folds.** Including `ir_world_lighting.glsl` in the clear kernel would declare its binding-4 `LightSourceBuffer` SSBO in a pass that binds nothing at slot 4 (Metal side would drag in the light list + helpers). The issue sanctioned "fold or document".
- `c_voxel_visibility_compact.glsl:269` has a *similar* bounds test that is **not** folded into `isShadowFeederIso` — it's the #1812 Hi-Z coverage domain (no route terms, different semantics).
- **Perf is inconclusive, not clean.** On an alternating A/B series the spread within a single variant exceeded any between-variant delta (`voxelCompact` 0.069→0.789 for identical shaders), so this harness can't resolve a difference on this host. The change is a textual extraction the compiler inlines and output is byte-identical, but I'm not claiming a measured perf result.

Closes #2513

### DIFF
--- a/engine/render/CLAUDE.md
+++ b/engine/render/CLAUDE.md
@@ -120,7 +120,12 @@ Shared includes: `ir_iso_common.glsl`, `ir_constants.glsl`,
 ACES tonemap, shared by the world-lighting passes),
 `ir_voxel_face_select.glsl` (fog reveal + face selection + per-axis
 store key, shared by the stage-1/stage-2 kernel family — the fog grid
-image slot is wrapper-supplied via `IR_VOXEL_FOG_GRID_BINDING`).
+image slot is wrapper-supplied via `IR_VOXEL_FOG_GRID_BINDING`),
+`ir_resolve_cardinal_emit.glsl` (the cardinal-layout micro-cell diamond
+emit shared by the two sun-shadow RESOLVE scatter kernels — on the GLSL
+side it writes through the wrapper's own `resolveScratch` SSBO, since
+GLSL cannot pass a buffer as an argument; the Metal twin takes it as a
+parameter).
 Shader file paths are stored in `render/shader_names.hpp`. Update that
 header when you add or rename a shader.
 
--- a/engine/render/src/shaders/c_clear_light_volume.glsl
+++ b/engine/render/src/shaders/c_clear_light_volume.glsl
@@ -13,6 +13,12 @@
 
 layout(local_size_x = 8, local_size_y = 8, local_size_z = 8) in;
 
+// Integer mirror of `kLightVolumeSize` in component_canvas_light_volume.hpp —
+// the same extent ir_world_lighting.glsl publishes as a float for the sampling
+// passes (which divide by it to reach texel centers). Deliberately NOT folded
+// into that fragment: this kernel needs one integer bound for a loop guard, and
+// including ir_world_lighting.glsl would declare its binding-4 LightSourceBuffer
+// SSBO in a clear pass that binds nothing at slot 4. Keep the two in lockstep.
 const int kLightVolumeSize = 128;
 
 layout(rgba8, binding = 0) writeonly uniform image3D lightVolume;
--- a/engine/render/src/shaders/c_resolve_per_axis_screen_depth.glsl
+++ b/engine/render/src/shaders/c_resolve_per_axis_screen_depth.glsl
@@ -54,6 +54,11 @@ layout(std430, binding = 28) restrict buffer PerAxisResolveScratch {
     int resolveScratch[];
 };
 
+// The cardinal-layout micro-cell emit shared with c_resolve_world_placed_depth.
+// Included AFTER ir_iso_common.glsl and AFTER the resolveScratch declaration it
+// writes through (the fragment's wrapper contract).
+#include "ir_resolve_cardinal_emit.glsl"
+
 // #2256: this stage is dispatched indirectly over only this axis's OCCUPIED
 // cells (compacted by the STAGE_1 per-axis pre-pass) instead of sweeping the
 // full worst-case grid. compactedCells holds the occupied linear cell indices;
@@ -145,24 +150,12 @@ void main() {
     for (int v = 0; v < scale; ++v) {
         for (int u = 0; u < scale; ++u) {
             const ivec3 microView = viewPos + stepU * u + stepV * v;
-            // Per-micro-cell depth, shared by the region's two pixels — the
-            // exact encode a real cardinal store would hold here, so BAKE's
-            // pixel+depth inverse recovers points on the face plane.
-            const int encoded = encodeDepthWithFace(pos3DtoDistance(microView), slot, flip);
-            const ivec2 cellBase = mainBase + pos3DtoPos2DIso(microView);
-            for (int k = 0; k < 2; ++k) {
-                const ivec2 mainPixel = cellBase + faceOffset_2x3(slot, k);
-                if (!isInsideCanvas(mainPixel, canvasSizePixels)) {
-                    continue;
-                }
-                // Front-most per screen pixel: smallest encoded distance wins
-                // (depth dominates the 3 low bits — flip+slot), exactly like
-                // the main canvas atomicMin store.
-                atomicMin(
-                    resolveScratch[mainPixel.y * canvasSizePixels.x + mainPixel.x],
-                    encoded
-                );
-            }
+            // `slot` twice: the per-axis store is already view-frame, so the
+            // slot IS the region axis; the world-placed twin must rotate its
+            // model face to get one.
+            emitResolveCardinalDiamond(
+                microView, slot, slot, flip, mainBase, canvasSizePixels
+            );
         }
     }
 }
--- a/engine/render/src/shaders/c_resolve_world_placed_depth.glsl
+++ b/engine/render/src/shaders/c_resolve_world_placed_depth.glsl
@@ -63,6 +63,12 @@ layout(std430, binding = 28) restrict buffer PerAxisResolveScratch {
     int resolveScratch[];
 };
 
+// The cardinal-layout micro-cell emit shared with
+// c_resolve_per_axis_screen_depth. Included AFTER ir_iso_common.glsl and AFTER
+// the resolveScratch declaration it writes through (the fragment's wrapper
+// contract).
+#include "ir_resolve_cardinal_emit.glsl"
+
 void main() {
     const ivec2 cell = ivec2(gl_GlobalInvocationID.xy);
     const ivec2 detachedSize = imageSize(detachedDistances);
@@ -111,7 +117,6 @@ void main() {
         viewPos = rotateCardinalZ(worldPos, cardinalIndex);
         viewPos += cardinalLowerCornerShift(cardinalIndex) * scale;
     }
-    const int encoded = encodeDepthWithFace(pos3DtoDistance(viewPos), slot, flip);
 
     const ivec2 mainBase = trixelFrameOffset(
         trixelOriginOffsetZ1(canvasSizePixels), frameCanvasOffset, voxelRenderOptions
@@ -128,15 +133,7 @@ void main() {
         rotateCardinalZ(faceOutwardNormal6I(slot << 1), cardinalIndex);
     const int viewAxis =
         (viewNormal.x != 0) ? kXFace : ((viewNormal.y != 0) ? kYFace : kZFace);
-    const ivec2 cellBase = mainBase + pos3DtoPos2DIso(viewPos);
-    for (int k = 0; k < 2; ++k) {
-        const ivec2 mainPixel = cellBase + faceOffset_2x3(viewAxis, k);
-        if (!isInsideCanvas(mainPixel, canvasSizePixels)) {
-            continue;
-        }
-        // Front-most per screen pixel: smallest encoded distance wins (depth
-        // dominates the 2 slot bits), exactly like the main canvas atomicMin
-        // store.
-        atomicMin(resolveScratch[mainPixel.y * canvasSizePixels.x + mainPixel.x], encoded);
-    }
+    emitResolveCardinalDiamond(
+        viewPos, viewAxis, slot, flip, mainBase, canvasSizePixels
+    );
 }
--- a/engine/render/src/shaders/c_voxel_to_trixel_stage_2_body.glsl
+++ b/engine/render/src/shaders/c_voxel_to_trixel_stage_2_body.glsl
@@ -384,9 +384,14 @@ void main() {
             feederPos = rotateCardinalZ(feederPos, cardinalIndex);
             feederPos += cardinalLowerCornerShift(cardinalIndex);
         }
-        const ivec2 feederIso = pos3DtoPos2DIso(feederPos);
-        if (feederIso.x < visibleIsoBounds.x || feederIso.x > visibleIsoBounds.z ||
-            feederIso.y < visibleIsoBounds.y || feederIso.y > visibleIsoBounds.w) {
+        // One definition with the compact cull's Step-B classify
+        // (isShadowFeederIso, ir_iso_common.glsl) so the skip and the
+        // classification cannot drift. The predicate re-tests the two route
+        // terms — provably true inside this gate; the gate itself stays as the
+        // cheap early-out that avoids the cardinal projection above.
+        if (isShadowFeederIso(
+                pos3DtoPos2DIso(feederPos), visibleIsoBounds, residualYaw, isDetachedCanvas
+        )) {
             return;
         }
     }
--- a/engine/render/src/shaders/c_voxel_visibility_compact.glsl
+++ b/engine/render/src/shaders/c_voxel_visibility_compact.glsl
@@ -400,10 +400,11 @@ void main() {
                             // Feeders are off-screen by construction, so their
                             // coarser trixel depth reaches only the sun-shadow
                             // bake, never a visible pixel.
-                            bool isFeeder =
-                                residualYaw == 0.0 && isDetachedCanvas < 0.5 &&
-                                (isoPos.x < visibleIsoBounds.x || isoPos.x > visibleIsoBounds.z ||
-                                 isoPos.y < visibleIsoBounds.y || isoPos.y > visibleIsoBounds.w);
+                            // One definition with stage 2's skip
+                            // (isShadowFeederIso, ir_iso_common.glsl).
+                            bool isFeeder = isShadowFeederIso(
+                                isoPos, visibleIsoBounds, residualYaw, isDetachedCanvas
+                            );
                             if (isFeeder) {
                                 // Tail-append: feeder slot i lands at
                                 // voxelCount-1-i, growing down from the top of the
--- a/engine/render/src/shaders/ir_iso_common.glsl
+++ b/engine/render/src/shaders/ir_iso_common.glsl
@@ -450,6 +450,36 @@ bool isInsideCanvas(ivec2 pixel, ivec2 canvasSize) {
            pixel.y >= 0 && pixel.y < canvasSize.y;
 }
 
+// Shadow-feeder classification (#1740): on the cardinal single-canvas world
+// route, a voxel whose cardinal iso position lies outside the UN-widened
+// visible viewport but inside the shadow-feeder-widened cull exists only to
+// cast sun shadows onto on-screen pixels through stage 1's distance bake — it
+// is never displayed, lit, or picked.
+//
+// Two kernels ask this same question and must agree: c_voxel_visibility_compact
+// partitions survivors into the visible list vs the strided off-screen feeder
+// list (#2258 Step B), and c_voxel_to_trixel_stage_2 skips a feeder's colour +
+// entity-id taps. They re-derived the predicate independently; one definition
+// is what keeps the classification and the skip from drifting (a compact that
+// called a voxel VISIBLE while stage 2 skipped it would drop an on-screen
+// pixel's colour). Over-classifying VISIBLE is the only safe failure direction.
+//
+// The route terms are part of the predicate, not a caller-side gate: the
+// widened cull only exists on the cardinal (residualYaw == 0) world
+// (isDetachedCanvas < 0.5) route, so both terms must hold before an
+// out-of-bounds iso means "feeder". When sun shadows are off,
+// visibleIsoBounds == cullIsoMin/Max and this never fires — byte-identical.
+bool isShadowFeederIso(
+    ivec2 isoPos,
+    ivec4 visibleIsoBounds,
+    float residualYaw,
+    float isDetachedCanvas
+) {
+    return residualYaw == 0.0 && isDetachedCanvas < 0.5 &&
+        (isoPos.x < visibleIsoBounds.x || isoPos.x > visibleIsoBounds.z ||
+         isoPos.y < visibleIsoBounds.y || isoPos.y > visibleIsoBounds.w);
+}
+
 // Hardware round() is safe here: the 1e-4 near-grid gate excludes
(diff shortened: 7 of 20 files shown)


jakildev:
Filed the canvas_stress macos-debug ref staleness noted in the PR body as #2538 (agent-approved lane: verified this session, defect-shaped, bounded fix). It is independent of this PR — `origin/master` alone reproduces all 8 failures with identical metrics — so this PR is not gated on it. — worker

jakildev (commented):
## Review — render: residual shader dedup — resolve emit, feeder predicate, Metal lighting UBOs, clear-volume constant (#2513)

**Verdict:** approve

Standalone PR (`baseRefName` is `master`), shader-only textual dedup across
four independent items from the #2508 audit. Read every changed file in
full (not just hunks), cross-referenced both GLSL and Metal call sites, and
independently rebuilt + ran on this macOS/Metal host rather than trusting
the PR body's own build claim.

- `fleet-build --target IRShapeDebug` — clean, no warnings from the touched
  files.
- `fleet-run IRShapeDebug --auto-screenshot 3` — `ir-run: RESULT=CLEAN`,
  confirming the new/edited `.metal` files (including the two new
  `ir_resolve_cardinal_emit.metal`/`.glsl` fragments) compile and run at
  runtime, not just link.

### Checklist (engine review checklist)

- **Resolve-emit extraction** (`ir_resolve_cardinal_emit.{glsl,metal}`) —
  traced both call sites against the original inline code they replaced.
  `c_resolve_per_axis_screen_depth` passes `regionAxis=slot, slot=slot`
  (region and encode both keyed on `slot`, matching the original — the
  per-axis store is already view-frame). `c_resolve_world_placed_depth`
  passes `regionAxis=viewAxis, slot=slot` (region keyed on the rotated
  `viewAxis`, encode still keyed on `slot`, matching the original). Same
  parameter split holds on both Metal call sites. Math body (encode →
  `pos3DtoPos2DIso` → 2-pixel `faceOffset_2x3` diamond → `atomicMin`) is a
  byte-for-byte relocation on both backends.
- **Shadow-feeder predicate** (`isShadowFeederIso`,
  `ir_iso_common.{glsl,metal}`) — both call sites
  (`c_voxel_visibility_compact`, `c_voxel_to_trixel_stage_2_body`) pass the
  same four terms in the same order on both backends; boolean expression is
  identical to what it replaced.
- **Metal lighting UBO fold** — verified `c_light_overflow_faces.metal` and
  `c_lighting_to_trixel.metal` reference only `lightVolumeParams.worldOriginVoxel`
  (`.xyz`/`.w`), never the four now-elided fields, so collapsing the
  underscore-tombstoned copies onto the seed's live field names
  (`gridSize`/`halfExtent`/`lightCount`/`stepFalloff`) is safe — those
  fields were genuinely dead in both consumers, not just per the PR body's
  claim but confirmed by grep against the post-patch files.
  `c_propagate_light_volume.metal` correctly keeps its own separate copy
  (documented, unchanged field layout).
- **`kLightVolumeSize` int/float mirror** — documented-not-folded per the
  issue's "fold or document" allowance; no functional change, comment-only.
- **Include-order contract** — both GLSL wrapper files include
  `ir_iso_common.glsl`, then declare `resolveScratch`, then
  `#include "ir_resolve_cardinal_emit.glsl"` — matches the fragment's
  documented "AFTER ir_iso_common.glsl AFTER the resolveScratch
  declaration" wrapper contract exactly.
- **Non-recursive GLSL / recursive Metal include-resolver claims** — spot
  checked against the actual implementations
  (`engine/render/src/opengl/opengl_shader.cpp::resolveShaderIncludes` is a
  single-pass line scan with no recursive call — confirms the GLSL
  fragment's "non-recursive (#2514)" comment; Metal's
  `loadAndPreprocessMetalSource` is documented elsewhere as
  visited-set-recursive, and `ir_iso_common.metal` carries its own
  `#ifndef` guard as a second line of defense). Both new fragments are
  included exactly once per wrapper, so neither resolver's behavior is
  actually exercised at the recursion boundary — moot in practice, but the
  documented reasoning is accurate.
- **New `ir_*.metal` fragment naming** — `ir_resolve_cardinal_emit.metal`
  matches the `ir_*.metal` filename convention the Metal AOT glob (opt-in,
  `IRREDEN_METAL_AOT_SHADERS`) already excludes as an include-fragment; no
  extra wiring needed there.
- **`c_` / `ir_` naming, GLSL/Metal parity** — every touched GLSL file has
  its Metal twin edited in lockstep; no orphaned one-sided change.
- **Verification-loop carve-out** — PR correctly invokes the
  `engine/render/CLAUDE.md` "internal refactors with provably no runtime
  effect" exception instead of running the full `render-debug-loop`, and
  backs it with the byte-identity proof table plus positive enabled-path
  perturbation tests (inverting `isShadowFeederIso` moved 7/7 `perf_grid`
  shots; perturbing the resolve-emit's depth key moved all 8 `canvas_stress`
  rotating shots) — this is exactly the positive-enabled-path evidence the
  same doc's "default-off features need a positive enabled-path test" rule
  asks for, applied correctly to a gated-off-by-default byte-identity claim.

### Acceptance (issue #2513)

| Criterion | Grade |
|---|---|
| Textual extraction only, no logic restructure | met |
| shape_debug 28-shot A/B byte-identical vs base | met |
| fog demo refs 13/13 | met |
| canvas_stress within #2479 noise | met (independently corroborated — #2538 was filed same-day matching the promised follow-up) |
| jitter floor (#2469) exact | met |
| all runs RESULT=CLEAN | met |
| Both backends stay in lockstep | met |
| Linux/GL smoke for the GLSL twins | unverifiable on this (macOS) host — closed via the `fleet:needs-windows-smoke` label below, not a hard failure since the PR body discloses the gap |
| Tree-wide grep: one definition per hoisted symbol | met |

No unmet criteria. Item 2's landing spot (`ir_iso_common.glsl` instead of
the issue's suggested `ir_voxel_face_select.glsl`) is disclosed and
justified in the PR body (binding collision with `canvasFogOfWar` /
`FogObserverData` at the suggested home) — acceptable mechanism drift.

### Nits
None.

- [x] (already done) `fleet-build --target IRShapeDebug` clean;
      `render-verify` 28/28 byte-identical, `IRFogDemo` 13/13 PASS,
      `jitter_probe` at documented floor — all `RESULT=CLEAN`.
- [x] (re-confirmed independently this pass, macOS/Metal) build +
      `IRShapeDebug --auto-screenshot` clean run.
- [ ] Windows/OpenGL smoke to validate the GLSL twins (tagging below).

Escalation: please Opus-recheck before merge (touches: `engine/render/` —
a Metal UBO struct fold (`FrameDataLightingToTrixel`/`LightVolumeParams`)
consumed across three kernels, plus a resolve-emit shared function whose
`regionAxis`/`slot` parameter split governs the sun-shadow depth-diamond
scatter on both backends). I traced the parameter wiring and layout by
hand and it holds, but this is exactly the class of GPU/CPU-adjacent
change the reviewer protocol asks to escalate rather than approve solo.

Opus recheck required: Metal UBO struct fold (FrameDataLightingToTrixel +
LightVolumeParams) now shared across c_lighting_to_trixel and
c_light_overflow_faces, plus the resolve-emit regionAxis/slot parameter
split shared across both RESOLVE kernels on both backends.

jakildev (commented):
## Opus recheck — render: residual shader dedup (#2513), PR #2537

**Verdict:** approve

Standalone PR (`baseRefName` = `master`, `MERGEABLE`), head
`53ee746`. Sonnet's first pass covered the checklist surface and
independently rebuilt/ran on Metal; per the model split I did not
re-do that work. This pass targets exactly the two things Sonnet
escalated — the Metal UBO struct fold and the resolve-emit
`regionAxis`/`slot` split — plus the one class Sonnet structurally
*could not* cover from a macOS host: whether the GLSL twins will
compile at all on the OpenGL backend.

### The escalated items

**Metal UBO fold (`FrameDataLightingToTrixel`, `LightVolumeParams`).**
Sonnet flagged this as the risky item; on closer read I confirm it is
safe, and the reason is stronger than "the elided fields are unread".

- Both structs are *verbatim* relocations — field order and types
  unchanged (`int,int,int,float,int4` → offsets 0/4/8/12/16, size 32,
  16-byte aligned for `LightVolumeParams`; 8×4B + `float4` → size 48
  for `FrameDataLightingToTrixel`). The only textual delta is the
  `_gridSize`→`gridSize` tombstone-to-live rename, which is a name,
  not a layout. So the CPU↔GPU binary contract at `[[buffer(23)]]` /
  `[[buffer(27)]]` is unchanged *by construction* — no need to trust
  a field-by-field CPU comparison.
- Independently re-confirmed the four now-live fields are genuinely
  dead in both folded consumers: `c_light_overflow_faces.metal` and
  `c_lighting_to_trixel.metal` reference `lightVolumeParams` only at
  `.worldOriginVoxel.xyz` (`:96` / `:247`) and `.worldOriginVoxel.w`
  (`:101` / `:259`). No `gridSize` / `halfExtent` / `lightCount` /
  `stepFalloff` read anywhere in either file. Renaming the tombstones
  only widens the accessible name set; it cannot change behaviour.
- **Duplicate-definition risk, which is the real hazard of hoisting a
  struct into an unguarded fragment.** `ir_world_lighting.metal` has
  no `#ifndef` self-guard, so a *transitive* second include would
  redefine three structs under the AOT `xcrun metal -c` path (real
  preprocessor, no visited set). I grepped: neither
  `ir_per_axis_lighting.metal` nor `ir_sun_shadow_sample.metal` — the
  other two fragments `c_light_overflow_faces.metal` pulls in —
  includes `ir_world_lighting.metal`. Tree-wide, exactly three files
  include it (overflow, lighting, seed), each once, each at top level.
  And this exposure is pre-existing, not introduced here:
  `GPULightSource` already lived in that unguarded file, so any
  transitive path would already have been broken. No live defect.
- Post-fold there is exactly one `struct FrameDataLightingToTrixel`
  and two `struct LightVolumeParams` tree-wide, the second being
  `c_propagate_light_volume.metal`'s deliberate local copy — which
  correctly does *not* include `ir_world_lighting.metal`, so no
  conflict, and the "keep in lockstep" comment now exists on both
  sides of that pair.

**Resolve-emit `regionAxis`/`slot` split.** Sonnet traced the wiring
by hand and I reach the same conclusion, by argument-for-argument
substitution against the deleted code on all four call sites:

| call site | old region arg | old encode arg | new `(regionAxis, slot)` |
|---|---|---|---|
| `c_resolve_per_axis_screen_depth.glsl` | `faceOffset_2x3(slot,k)` | `encodeDepthWithFace(…, slot, flip)` | `(slot, slot)` ✓ |
| `c_resolve_world_placed_depth.glsl` | `faceOffset_2x3(viewAxis,k)` | `encodeDepthWithFace(…, slot, flip)` | `(viewAxis, slot)` ✓ |
| `…_per_axis_screen_depth.metal` | `faceOffset_2x3(slot,k)` | `(…, slot, flip)` | `(slot, slot)` ✓ |
| `…_world_placed_depth.metal` | `faceOffset_2x3(viewAxis,k)` | `(…, slot, flip)` | `(viewAxis, slot)` ✓ |

Scratch indexing (`mainPixel.y * canvasSize.x + mainPixel.x`), the
`isInsideCanvas` guard, the `atomicMin` / `atomic_fetch_min_explicit`
ordering, and the `k∈{0,1}` two-pixel loop are all identical to what
they replaced on both backends. `canvasSize` is bound to
`canvasSizePixels` (GLSL) / `frameData.canvasSizePixels` (Metal) at
every call — the same value the old inline code read. The
world-placed kernel's `encoded` moved from a pre-`mainBase` local into
the callee; it is a pure function of `viewPos/slot/flip` with no
intervening side effects, and I verified no other reference to
`encoded` or `cellBase` survives in either GLSL wrapper (so no
dangling identifier hiding behind the unbuilt GLSL path).

### The gap Sonnet could not close: GLSL compilability

Sonnet's build+run evidence is Metal-only, so the GLSL twins are
textual-only-until-proven. Everything checkable statically, checks out:

- **Symbol scope.** `isShadowFeederIso`'s three new arguments are in
  scope at both GLSL call sites: `c_voxel_to_trixel_stage_2_body.glsl`
  declares `residualYaw` (:55), `isDetachedCanvas` (:58),
  `visibleIsoBounds` (:83) as uniforms in the same block;
  `c_voxel_visibility_compact.glsl` reads them unqualified as before.
- **Include ordering / non-recursive resolver.**
  `ir_resolve_cardinal_emit.glsl` itself includes nothing (correct for
  the single-pass resolver), and both wrappers include
  `ir_iso_common.glsl` (:24) → declare `resolveScratch` (:54 / :63) →
  `#include "ir_resolve_cardinal_emit.glsl"` (:60 / :70), all above
  `main()`. Wrapper contract satisfied on both.
- **`const` function parameters in GLSL** — legal, and not novel here:
  13 existing engine GLSL functions already use the idiom
  (`writeColorTapPerAxis`, `emitDeformedFace`, `viewMaskTap`, …), so
  the new fragment is not the first to lean on it.
- **File delivery.** `engine/render/CMakeLists.txt:244` copies
  `src/shaders` wholesale with `copy_directory`, so the new fragment
  reaches `${PROJECT_BINARY_DIR}/shaders` with no explicit
  registration; `shader_names.hpp` lists compiled shaders only, not
  include-fragments (`ir_iso_common.glsl` isn't there either), so the
  CLAUDE.md "update that header" rule doesn't apply to this file.
  The Metal AOT glob's `EXCLUDE REGEX "/ir_[^/]*\.metal$"`
  (`CMakeLists.txt:175`) already excludes the new `.metal` fragment.

That reduces the GL-side residual risk to backend-specific compiler
behaviour, which is what `fleet:needs-windows-smoke` is for — already
applied. Not a merge blocker; the label is the right disposition.

### `isShadowFeederIso` — the one behaviour-changing-looking hunk

Stage 2's original predicate tested *only* the four bounds terms; the
shared one additionally tests `residualYaw == 0` and
`isDetachedCanvas < 0.5`. That would narrow the skip (feeders start
taking colour/entity taps) if the enclosing gate didn't already
guarantee them. It does, on both backends — the hunk sits inside
`if (residualYaw == 0.0 && isDetachedCanvas < 0.5) {` (GLSL :381,
Metal :400), so the re-test is provably redundant, not restrictive.
The compact-cull call site is a literal expression move. Byte-identity
holds by argument, independent of the A/B evidence.

Cost: two extra uniform float compares per surviving stage-2 voxel on
the cardinal route — both operands are already live in registers from
the gate one line up, so any compiler CSEs them; not a hot-path
concern.

### Observations (no action required — not nits, nothing to fix here)

- `ir_world_lighting.metal` still has no `#ifndef` self-guard while
  now carrying three struct definitions instead of one. It's safe
  today (verified above: no transitive include exists) and the
  exposure predates this PR, but the fold does raise the blast radius
  if someone later adds `#include "ir_world_lighting.metal"` to
  `ir_per_axis_lighting.metal`. Worth a guard whenever that file is
  next touched — not worth an amend round now. Contrast with the new
  `ir_resolve_cardinal_emit.metal`, which correctly *documents* why it
  needs no guard (`static` internal linkage + once-per-wrapper
  include) rather than leaving it implicit — that's the better
  pattern.
- The `kLightVolumeSize` int/float mirror staying documented-not-folded
  is right: `c_clear_light_volume` binds nothing at slot 4, and the
  GLSL fold would inject an unbound `LightSourceBuffer` SSBO into a
  clear pass. The comment now states that reason on both backends.

### Sonnet's pass

Nothing to overturn. Sonnet flagged the UBO fold and the
`regionAxis`/`slot` split as the two items needing an Opus trace; on
closer read I confirm both, and the byte-identity claim survives
independent argument (not just the A/B evidence). Sonnet's acceptance
grading against #2513 also holds — including the correct read that the
`ir_iso_common.glsl` landing spot for item 2 is disclosed mechanism
drift, not scope creep. I'd add that the positive-control perturbation
tests in the PR body are load-bearing here in a way worth naming: they
prove the RESOLVE emit and the feeder predicate are *actually
executed* by the verify scenes, so the 28/28 byte-identical result is
evidence rather than a vacuous pass over never-run code.

### Merge disposition

Approve. Merge on green Windows/OpenGL smoke
(`fleet:needs-windows-smoke` already tagged) — the Metal side is
verified, the GL side is statically sound but unbuilt.

PR #2539, opened 2026-07-23
engine: deterministic World teardown at gameLoop tail (#2528)

### LINKED ISSUE #2528: engine: deterministic World teardown — reset g_world when gameLoop returns
**Area:** engine (ir_engine.hpp), engine/world
**Model:** opus
**Blocked by:** (none)

## Acceptance criteria

- `IREngine::gameLoop()` resets `g_world` after `World::gameLoop()` returns, so `~World()` — and every manager destructor in member order — runs while `main` is still on the stack and the graphics driver is loaded, never again at process-exit static destruction.
- `World::end()`'s GPU-resource teardown contract is unchanged (it becomes belt-and-braces instead of load-bearing).
- Docs updated: the `engine/world/CLAUDE.md` gotcha "Release GPU/GL resources in `end()`, never in `~World()`" keeps the rule but re-grounds the rationale (the static-destruction hazard is retired for the `IREngine::gameLoop` path; direct-`World` users and tests still own their lifetime); `engine/CLAUDE.md` manager-globals lifetime note gains "managers are torn down when `gameLoop()` returns".
- All in-tree creations audited for engine access after `gameLoop()` returns (survey at filing time found none — every `main` returns immediately); the PR body states the contract explicitly: post-`gameLoop` engine access is unsupported and now asserts in debug builds.
- Representative demos run clean **to process exit** on the fleet host: `IRShapeDebug`, one Lua-driven creation, and one video-capture run — confirming no teardown-order regression.

## Context

An architecture review of the engine's global-state patterns identified `IREngine::g_world` (an inline global `unique_ptr<World>`) as the one place the manager-global pattern has actually caused a shipped bug (#2031): `~World()` ran at process-exit static destruction, after the platform unloaded the GL driver, crashing on `glDelete*` against dead driver state. The `World::end()` workaround moved GPU teardown inside the loop, but the class of bug — any manager destructor observing a dead process environment — remains open as long as `~World()` runs after `main`.

Resetting `g_world` at the tail of `IREngine::gameLoop()` retires the class: destruction ordering across managers is unchanged (member order), it just runs earlier, deterministically, with the window/context/driver all still alive. The debug-build asserts in every `get*Manager()` accessor then make any post-loop engine access loud instead of undefined.

jakildev:
## Summary
- `IREngine::gameLoop()` resets `g_world` once `World::gameLoop()` returns, so every manager destructor runs while `main()` is on the stack and the window/graphics driver are still loaded — instead of at process-exit static destruction (the #2031 class, where a `glDelete*` from a member/observer dtor reaches an already-unloaded driver). Manager destruction *order* is unchanged (reverse member-declaration order); it just runs earlier, and deterministically.
- `World::end()` is untouched — the diff to `world.cpp` is comment-only. It remains the canonical device-resource release point, because the reset does **not** subsume it: `gameLoop()`'s catch block calls `end()` and rethrows (skipping the reset), and an owner constructing a `World` directly picks its own destruction point.
- Docs re-grounded in `engine/world/CLAUDE.md` (the "release in `end()`, never in `~World()`" rule stays; its rationale now states the hazard is retired for the `IREngine` path and why the rule still stands) and `engine/CLAUDE.md` §"Manager globals".

**Contract for out-of-tree consumers:** engine access after `IREngine::gameLoop()` returns is unsupported. It was never supported — GPU resources were already released by `end()` — but it now fails loudly: `getWorld()` and every `IR<Module>::get*Manager()` accessor assert on the cleared global in debug builds (and are UB in release, where `IR_ASSERT` compiles out). A `main()` may still read its own state (exit codes, file-scope counters) past that point.

## Test plan
- [x] Full build, all targets including `IrredenEngineTest` — clean
- [x] gtest suite — 1356/1356 pass
- [x] `IRShapeDebug`, `IRLuaPipelineDemo`, `IRSceneReset`, `IRChunkStreamingSmoke`, `IRPersistRoundtrip`, `IRDetachedProbe` — all exit 0 to process exit
- [ ] OpenGL-host smoke (this PR was authored on macOS/Metal — see reviewer note)

## Acceptance evidence
|---|---|---|
| `g_world` reset after `World::gameLoop()` returns; `~World()` runs with `main` on the stack, never at static destruction | Temporary `fprintf` probe in `~World()` + `IREngine::gameLoop()`, run A/B (probe reverted; tree is clean) | **A (with reset):** `gameLoop: World::gameLoop returned, resetting` → `~World() ENTER` → `gameLoop: reset done, returning to main`. **B (reset removed = master's behavior):** `…resetting` → `…returning to main` → `~World() ENTER`. The dtor moves from after-`main` to inside `gameLoop()`. |
| `World::end()`'s teardown contract unchanged | `git diff origin/master...HEAD -- engine/world/src/world.cpp` filtered to non-comment lines | Empty — the `world.cpp` hunk is comment-only; `World::end()` is byte-identical |
| Docs updated (`engine/world/CLAUDE.md` gotcha re-grounded, `engine/CLAUDE.md` manager-globals note) | `git diff --stat` | `engine/world/CLAUDE.md` +18/−6, `engine/CLAUDE.md` +10 |
| All in-tree creations audited for engine access after `gameLoop()` | `grep -rn "IREngine::gameLoop()" --include="*.cpp" creations/` + read the tail of each `main` | 34 call sites; every `main` returns immediately. `persist_roundtrip`, `scene_reset`, `chunk_streaming_smoke`, `detached_probe` read only their own file-scope counters (`g_failCount` / `g_smokeFailCount` / `g_anyProbeFailure`). No engine access anywhere post-loop. |
| Representative demos run clean to process exit: `IRShapeDebug`, one Lua-driven creation, one video-capture run | `fleet-run <target> --auto-screenshot`; exit code checked | `IRShapeDebug` exit 0; `IRLuaPipelineDemo` (Lua/sol2 teardown) exit 0; video-capture run exit 0 with a real 2.2 MB `capture.mp4` — log shows `VideoManager::shutdown() -- recording was active, finalizing.` and the mp4 moov-atom second pass completing, so the `m_finalizeThread` join path was exercised |

Additional teardown coverage beyond the named three: `IRSceneReset`, `IRChunkStreamingSmoke`, `IRPersistRoundtrip`, `IRDetachedProbe` — all exit 0 (these self-verify and return non-zero on internal failure). Full build + `IrredenEngineTest` 1356/1356 pass.

## Notes for reviewer
- **Host caveat.** Authored and verified on macOS/Metal. The #2031 crash itself is a Windows/MSYS2 symptom (that platform unloads the GL driver before static destructors run), so the *crash* is not reproducible here — what is verified on this host is the **ordering change** that retires the hazard, via the A/B probe above. An OpenGL-host smoke is the useful confirmation.
- **The exception path is deliberately left alone.** `g_world.reset()` is a plain post-call statement, not an RAII scope, so a throwing `gameLoop()` still skips it. That is not a regression: `end()` has already run in the catch block, and an exception escaping `main` terminates without running static destructors at all. Making it RAII would additionally tear the World down *during* unwinding, which is a behavior change the ticket didn't ask for.
- **Plan delta.** The plan's gotcha anticipated gtest fixtures that construct `World` directly; a tree-wide sweep found none (`IREngine::g_world` is the only `World` in the tree). `end()`'s contract still must not weaken — the surviving reasons are the exception path and out-of-tree owners, not in-tree fixtures. Noted in the committed plan file.

Closes #2528

### DIFF
--- a/.fleet/plans/issue-2528.md
+++ b/.fleet/plans/issue-2528.md
@@ -1,0 +1,74 @@
+# Plan: engine: deterministic World teardown — reset `g_world` when `gameLoop` returns
+
+- **Issue:** #2528
+- **Model:** opus
+- **Date:** 2026-07-23
+
+## Scope
+
+Deterministic `World` teardown for the standard creation entry path:
+`IREngine::gameLoop()` destroys the world when the loop exits, instead of
+leaving `~World()` to process-exit static destruction. No change to `World`
+itself, `end()`, or manager destruction order.
+
+## Approach
+
+1. `engine/include/irreden/ir_engine.hpp::gameLoop()`: after
+   `getWorld().gameLoop()` returns, `g_world.reset();` with a comment stating
+   the constraint (teardown must run while `main` is on the stack and the
+   graphics driver is loaded — the #2031 class).
+2. Leave `World::end()` untouched — its in-loop GPU teardown remains the
+   canonical device-resource release point, and it keeps release ordering
+   independent of this change.
+3. Docs: update the `engine/world/CLAUDE.md` gotcha (the rule stays; the
+   rationale notes `~World()` now runs at `gameLoop()`-tail for the `IREngine`
+   path) and `engine/CLAUDE.md` §"Manager globals" (managers valid until
+   `gameLoop()` returns; post-loop access asserts in debug).
+4. Audit: grep every in-tree creation `main` for engine calls after
+   `IREngine::gameLoop()`. State the contract in the PR body for out-of-tree
+   consumers: post-`gameLoop` engine access was never supported (GPU resources
+   were already gone via `end()`); it now fails loudly in debug.
+
+## Affected files
+
+- `engine/include/irreden/ir_engine.hpp`
+- `engine/world/CLAUDE.md`, `engine/CLAUDE.md` (doc hunks only)
+
+## Acceptance criteria
+
+As in the issue body. Verification: `fleet-build` + `fleet-run
+--auto-screenshot` to clean process exit for `IRShapeDebug`, one Lua-driven
+creation, and one video-capture-enabled run; exit codes checked (a teardown
+crash surfaces as a non-zero exit after the shots pass).
+
+## Gotchas
+
+- The intentionally-leaked singletons (`LoggerSpd`, the Metal runtime state)
+  are leaked precisely to survive shutdown ordering — unaffected; do not "fix"
+  them in this PR.
+- The prefab-name registry documented as surviving `World` shutdown
+  (`engine/script/CLAUDE.md`) keeps that property — it is process-scoped by
+  design; unaffected.
+- gtest targets that construct `World` directly own their lifetime already —
+  out of scope, and part of why `end()`'s contract must not weaken.
+- If a demo hangs or crashes at exit on one backend only, the failure is in
+  that backend's destructor ordering relative to the window — report per-host,
+  don't paper over with an early return.
+
+## Implementation notes (deltas found while executing)
+
+- **No in-tree `World` is constructed outside `IREngine::g_world`.** The
+  plan's gtest gotcha anticipated direct-`World` fixtures; a tree-wide sweep
+  (`kTestLuaConfig`, `unique_ptr<World>`, `World <ident>(`) found none — the
+  gtest targets use manager-level fixtures. `end()` still must not weaken:
+  the reasons that survive are the exception path (`World::gameLoop()`'s catch
+  block calls `end()` and rethrows, skipping the reset) and out-of-tree
+  owners, not in-tree test fixtures.
+- **`~World()`'s own comment asserted the now-false fact** that the dtor runs
+  at static destruction. Corrected in `engine/world/src/world.cpp`; this is a
+  comment-only hunk, not a change to `end()`.
+- **Exception path is deliberately unchanged.** `g_world.reset()` is a plain
+  post-call statement, not an RAII scope, so a throwing `gameLoop()` still
+  leaves the reset unrun. That is not a regression: `end()` has already run in
+  the catch block, and an exception escaping `main` terminates without running
+  static destructors at all.
--- a/engine/CLAUDE.md
+++ b/engine/CLAUDE.md
@@ -117,5 +117,17 @@ it owns every manager as a member in dependency order, so member order IS
 the deterministic set/clear order. The `ir_<module>.hpp`
 entry points wrap access via free functions (`IREntity::getEntityManager()`).
 **Do not hold references or raw pointers to managers across frames outside
-of World's lifetime.** Full global-state pattern catalog + the
-header-global ban: [`.claude/rules/cpp-globals.md`](../.claude/rules/cpp-globals.md).
+of World's lifetime.**
+
+**Managers are torn down when `gameLoop()` returns.** `IREngine::gameLoop()`
+resets `g_world` as soon as `World::gameLoop()` exits (#2528), so every manager
+destructor runs there — deterministically, with `main` still on the stack and
+the graphics driver still loaded — instead of at process-exit static
+destruction (#2031). The practical contract for a creation: **engine access
+after `IREngine::gameLoop()` returns is unsupported.** `main` should read only
+its own state (exit codes, file-scope counters) past that point; every
+`get*Manager()` accessor and `IREngine::getWorld()` asserts on the cleared
+global in debug builds, and is undefined in release.
+
+Full global-state pattern catalog + the header-global ban:
+[`.claude/rules/cpp-globals.md`](../.claude/rules/cpp-globals.md).
--- a/engine/include/irreden/ir_engine.hpp
+++ b/engine/include/irreden/ir_engine.hpp
@@ -105,8 +105,17 @@ inline int entityCountOverride() {
     return getWorld().entityCountOverride();
 }
 
+// Destroys the World when the loop exits, so every manager destructor runs
+// while main() is still on the stack and the window/graphics driver are alive.
+// Leaving it to process-exit static destruction is the #2031 hazard: a
+// `glDelete*` from a member or observer dtor reaches an already-unloaded
+// driver. Managers destruct in reverse member-declaration order either way.
+//
+// Engine access after gameLoop() returns is unsupported: getWorld() and every
+// IR<Module>::get*Manager() accessor assert on the cleared global in debug.
 inline void gameLoop() {
     getWorld().gameLoop();
+    g_world.reset();
 }
 
 } // namespace IREngine
--- a/engine/include/irreden/ir_engine.hpp
+++ b/engine/include/irreden/ir_engine.hpp
@@ -94,7 +94,7 @@ inline void init(int argc, char **argv, const char *configFileName = "config.lua
 }
 
 inline void runScript(const char *scriptFileName) {
-    g_world->runScript(resolveScriptPath(scriptFileName).c_str());
+    getWorld().runScript(resolveScriptPath(scriptFileName).c_str());
 }
 
 inline void enableFrameTiming(bool enabled) {
--- a/engine/world/CLAUDE.md
+++ b/engine/world/CLAUDE.md
@@ -496,12 +496,24 @@ same `config = { ... }` table — there is one source of truth per file.
   video recording until the first input arrives — used to keep capture
   clips from starting mid-loading-screen. If video recording is not
   starting, check these flags first.
-- **Release GPU/GL resources in `end()`, never in `~World()`.** `g_world`
-  is a global `unique_ptr`, so `~World()` runs at process-exit static
-  destruction — past that point the GL driver/context may already be torn
-  down (MSYS2 unloads it first), and any `glDelete*` issued from a
-  member/observer dtor crashes against dead driver state (#2031). `end()`
-  runs during `gameLoop()` while the context is live and is the canonical
+- **Release GPU/GL resources in `end()`, never in `~World()`.** `end()` runs
+  during `gameLoop()` while the context is provably live, and is the canonical
   spot for device-resource teardown (it already drives `destroyAllEntities()`
   for `onDestroy` GPU frees). The dtor stays a no-op safety net.
+
+  The hazard it guards against: `g_world` is a global `unique_ptr`, so a
+  `~World()` that runs at process-exit static destruction lands past the point
+  where the GL driver/context may already be torn down (MSYS2 unloads it
+  first), and any `glDelete*` issued from a member/observer dtor crashes
+  against dead driver state (#2031).
+
+  **That static-destruction hazard is retired for the `IREngine` path** —
+  `IREngine::gameLoop()` resets `g_world` as soon as `World::gameLoop()`
+  returns (#2528), so `~World()` runs with `main` on the stack and the driver
+  loaded. The rule still stands, because the reset does not cover two paths:
+  `World::gameLoop()`'s catch block calls `end()` and rethrows, skipping the
+  reset entirely; and an owner who constructs a `World` directly (rather than
+  through `IREngine::init`) picks its own destruction point. Treat the reset
+  as defense in depth, not a licence to move device-resource frees into the
+  dtor.
 
--- a/engine/world/src/world.cpp
+++ b/engine/world/src/world.cpp
@@ -95,13 +95,15 @@ World::World(const char *configFileName)
 
 World::~World() {
     // No-op after `end()` (which ran during `gameLoop()` while the GL context
-    // was still live and already cleared the observers). `g_world` is a global
-    // `unique_ptr`, so this dtor runs at process-exit static destruction — past
-    // that point the GL driver/context may already be torn down (MSYS2 unloads
-    // it first), and the GpuStageTimingObserver dtor's `glDeleteQueries` would
-    // crash against dead driver state (#2031). The live-context release in
-    // `end()` is the real cleanup; this stays idempotent as a safety net for any
-    // path that never ran the loop.
+    // was still live and already cleared the observers). On the `IREngine` path
+    // this runs at the tail of `IREngine::gameLoop()`, which resets `g_world`
+    // while the driver is still loaded; `end()` is still the real cleanup,
+    // since it also covers the exception path and any owner driving `World`
+    // directly. Releasing device resources from here instead would reach a
+    // torn-down driver on any path that does destruct at process exit — MSYS2
+    // unloads the GL driver first, so `glDeleteQueries` from the
+    // GpuStageTimingObserver dtor hits dead driver state (#2031). Idempotent,
+    // so it stays a safety net for a path that never ran the loop.
     m_systemManager.clearTickObservers();
     IRE_LOG_INFO("Clean shutdown complete.");
 }
--- a/engine/world/src/world.cpp
+++ b/engine/world/src/world.cpp
@@ -105,7 +105,6 @@ World::~World() {
     // GpuStageTimingObserver dtor hits dead driver state (#2031). Idempotent,
     // so it stays a safety net for a path that never ran the loop.
     m_systemManager.clearTickObservers();
-    IRE_LOG_INFO("Clean shutdown complete.");
 }
 
 void World::setupLuaBindings(const std::vector<LuaBindingRegistration> &bindings) {
@@ -290,9 +289,10 @@ void World::start() {
 void World::end() {
     buildAndWriteProfileReport();
     // Release the GPU stage-timing observer's GL timestamp queries here, while
-    // the render context is guaranteed live. The observer is program-bound and
-    // only ~World() would otherwise destroy it — but that runs at static
-    // destruction, after the GL driver/context may already be gone (#2031).
+    // the render context is guaranteed live. The observer is program-bound;
+    // clearing it from ~World() instead is safe on the IREngine path (which
+    // resets g_world at gameLoop's tail, #2528) but would still reach a
+    // torn-down driver on any path that destructs at process exit (#2031).
     // `clearTickObservers()` is idempotent, so the dtor's later call no-ops.
     m_systemManager.clearTickObservers();
     m_videoManager.shutdown();
@@ -300,6 +300,11 @@ void World::end() {
     // (e.g. MIDI cleanup that sends NOTE_OFF on shutdown).
     m_entityManager.destroyAllEntities();
     IRProfile::CPUProfiler::instance().shutdown();
+    // Last log before shutdownLogging() flips g_loggingEnabled off — the
+    // observable proof that deterministic teardown ran (#2528). The same line
+    // in ~World() is swallowed, since end() always runs first on both the loop
+    // and exception paths and disables logging just below.
+    IRE_LOG_INFO("Clean shutdown complete.");
     IRProfile::shutdownLogging();
 }
 


jakildev (commented):
## Review — engine: deterministic World teardown at gameLoop tail (#2528)

**Verdict:** approve

Moves `~World()` (and every manager destructor, in unchanged reverse-member
order) from process-exit static destruction to the tail of
`IREngine::gameLoop()`, retiring the #2031 hazard class (a `glDelete*` from a
member/observer dtor reaching an already-unloaded driver) — deterministically,
for every host, not just the MSYS2 case that originally surfaced it.
`World::end()` is untouched (the `world.cpp` hunk is comment-only, confirmed by
reading the diff directly, not the stated claim).

### Verification performed independently (not just re-stating the PR body)

- **`world.cpp` diff is comment-only.** Confirmed by reading the hunk: only
  the `~World()` doc comment changed; `World::end()` (world.cpp:288+) has no
  diff.
- **Exception path genuinely skips the reset.** `World::gameLoop()`'s
  `catch (...)` (world.cpp:257-260) calls `end()` then `throw;` — the
  exception propagates out of `getWorld().gameLoop()` in
  `IREngine::gameLoop()`, so `g_world.reset()` (the line right after) is never
  reached on that path. Matches the PR body's claim exactly.
- **Assert claims are pre-existing code, not new.** `IREngine::getWorld()`
  already asserts `g_world != nullptr` (`ir_engine.hpp:19`, untouched by this
  diff); `IREntity::getEntityManager()` asserts its own raw global likewise
  (`ir_entity.cpp:6-8`). This PR doesn't add new asserts — it makes the
  *existing* ones actually fire post-loop instead of being unreachable.
- **34-call-site audit re-verified.** `grep -rn "IREngine::gameLoop()"
  --include="*.cpp" creations/` → 34 matches (33 real calls + 1 comment
  reference in `scene_reset/main.cpp:11`). Spot-checked
  `persist_roundtrip`, `scene_reset`, `chunk_streaming_smoke`,
  `detached_probe` plus 4 more random samples (`voxel_editor`,
  `stateless_particles`, `day_cycle`, `perf_grid`) — every one returns
  immediately or reads only a file-scope counter after the call. Also swept
  `test/` for `IREngine::gameLoop()` calls or direct `World` construction —
  none found, matching the plan's "no in-tree fixture constructs World
  directly" finding.
- **Built + ran `IRShapeDebug --auto-screenshot` from this branch on this
  host (macOS/Metal)** — `ir-run: RESULT=CLEAN exe=IRShapeDebug exit=0`,
  independently reproducing the PR's own clean-exit claim rather than taking
  it on faith. `fleet-build --target format-changed` — clean, nothing to
  format.

### Acceptance (issue #2528)

| Criterion | Grade | Basis |
|---|---|---|
| `gameLoop()` resets `g_world`; `~World()`/manager dtors run with `main` on the stack, never at static destruction | met | `ir_engine.hpp` diff adds `g_world.reset();` right after `getWorld().gameLoop();`; A/B probe evidence in the PR body is structurally sound and matches the code |
| `World::end()`'s teardown contract unchanged | met | `world.cpp` diff is comment-only, confirmed above |
| Docs updated (`engine/world/CLAUDE.md`, `engine/CLAUDE.md`) | met | both diffs re-ground the rationale without weakening the "release in `end()`" rule |
| All in-tree creations audited for post-`gameLoop` engine access | met | independently re-verified above (34 sites, 33 real, all clean) |
| Representative demos run clean to process exit (`IRShapeDebug`, Lua-driven, video-capture) | met (macOS/Metal); unverifiable (OpenGL) | I independently reproduced the `IRShapeDebug` result; the Lua-driven and video-capture rows rest on the PR's own log evidence, which cross-references real code (`video_manager.cpp` finalize-thread join). No OpenGL-tier run yet — see escalation note. |

Per the acceptance-grading rule, the OpenGL-tier row is unverifiable-not-penalized
rather than a merge-blocker.

### Nits
- None.

### Praise
- The exception-path reasoning (why `g_world.reset()` deliberately stays a
  plain post-call statement, not RAII, and why that's *not* a regression) is
  the kind of design-intent note that saves the next reader from "fixing" it
  into a bug. Same for the "defense in depth, not a licence to move device
  frees into the dtor" framing in `engine/world/CLAUDE.md` — it pre-empts the
  natural but wrong follow-up change.

- [x] (verified independently) `fleet-build --target IRShapeDebug` + `fleet-run IRShapeDebug --auto-screenshot` — CLEAN, exit 0, macOS/Metal.
- [x] (verified independently) `fleet-build --target format-changed` — clean.
- [ ] OpenGL-host (Linux or Windows) smoke run to process exit — this PR carries no platform-conditional code (no `#ifdef`, no direct GL/Metal call added; the change is pure C++ object-lifetime timing), so the risk is low, but the hazard being retired (#2031) was originally an MSYS2-specific symptom and the fix has not yet been reproduced on that tier. Not gating per the cross-host-smoke procedure's path filters (this diff touches none of them), but worth a confirmation pass given the universal blast radius of `ir_engine.hpp`.

Escalation: please Opus-recheck before merge (touches: a process-global
lifetime/teardown-ordering invariant that every creation in the tree depends
on). Sonnet has independently confirmed the exception-path skip, the
pre-existing-assert claim, the comment-only `world.cpp` diff, the 34-site
creation audit, and a clean macOS/Metal build+run — Opus's value-add here is
a second look at any Lua-callback/sol2-closure lifetime edge case this pass
didn't turn up (none found), not re-deriving the above.

Opus recheck required: global `World`/manager lifetime + teardown-ordering
change with tree-wide blast radius (every creation calls
`IREngine::gameLoop()`); no platform-conditional code added but the retired
hazard (#2031) is an MSYS2-originated symptom not yet re-verified on that tier.

jakildev (commented):
## Opus recheck — engine: deterministic World teardown at gameLoop tail (#2528)

Standalone PR (`baseRefName: master`), no stack context. Sonnet's first pass
independently confirmed the exception-path skip, the pre-existing-assert claim,
the comment-only `world.cpp` diff, the 34-site creation audit, and a clean
macOS/Metal `IRShapeDebug` run. I did not re-derive those. This pass spent its
budget on the four things a first pass could not settle: **who else's lifetime
just moved**, **whether the newly documented "it asserts" contract is actually
true**, **the consumers outside this repo's `creations/`**, and **the Lua /
sol2 teardown path Sonnet flagged as the intended Opus value-add**.

### 1. The static-destruction ordering flip — checked, no casualties

`g_world` is an `inline unique_ptr` with constant initialization, so its
destructor was previously registered at the very front of the atexit chain and
ran **last**, after every function-local static in the process. Moving it to
`gameLoop()`-tail inverts that: `~World()` now runs *before* every static that
outlives it. Any static whose destructor touched engine state was silently
correct before and would be UB now. I enumerated them:

- `IRProfile::CPUProfiler::instance()` (`cpu_profiler.cpp:8`) — function-local
  static; its dtor dumps easy_profiler blocks and touches no manager. Its
  `IRE_LOG_INFO` goes through `LoggerSpd`, which is deliberately leaked. Safe.
- `IrredenGLAPI::instance()` (`ir_gl_api.cpp:11`) — leaked-by-design
  `unique_ptr` holding a `GL4API` POD of function pointers. Trivially
  destructible content, no GL call at teardown. This one mattered: it now
  outlives `glfwTerminate()` (called from `~IRGLFWWindow`,
  `ir_glfw_window.cpp:140`), which was not true before. It is inert, so the
  flip is harmless — but it is the one place where "static outlives the GL
  context" became newly reachable.
- `IREngine::g_luaBindingRegistrations` — a global `vector<std::function>` that
  now outlives `m_lua` (the `sol::state`). If any registration lambda captured
  a `sol::object`/`sol::table` by value, its destructor would run after
  `lua_close()` → UAF at exit. Every in-tree registration is capture-less
  (`IREngine::registerLuaBindings([](IRScript::LuaScript &…)`, 6 sites across
  `default`, `sprite_demo`, `lua_widgets`, `lua_perf_grid`,
  `lua_pipeline_demo`, `collision_overlap_demo`, `audio_playback`); a grep for
  a capturing form returns nothing. Clean today, latent tomorrow — worth
  knowing that `clearLuaBindings()` exists as the escape hatch if a capturing
  registration ever lands.

Verdict on the ordering flip: no in-tree casualty. Worth stating explicitly in
the review record because it is the failure mode a future static would walk
into silently.

### 2. The "it asserts on the cleared global" contract — true, with one hole

`engine/CLAUDE.md` and the new `ir_engine.hpp` comment both promise that after
`gameLoop()` returns, "`getWorld()` and every `IR<Module>::get*Manager()`
accessor assert on the cleared global in debug." That promise only holds if
every manager nulls its global in its own destructor — otherwise post-loop
access is a *dangling* read, which is strictly worse than the documented
assert. I checked all eleven:

`EntityManager` (`entity_manager.cpp:41`), `SystemManager`
(`system_manager.cpp:18`), `JobManager` (`job_manager.cpp:142`),
`InputManager` (`input_manager.cpp:40`), `CommandManager`
(`command_manager.cpp:17`), `RenderingResourceManager` (`rendering_rm.cpp:34`),
`RenderManager` (`render_manager.cpp:194`), `AudioManager`
(`audio_manager.cpp:24`), `TimeManager` (`time_manager.cpp:19`),
`VideoManager` (`video_manager.cpp:30`) — every one uses the same
`if (g_x == this) g_x = nullptr;` guard, and every accessor asserts on null.
The documented contract is accurate. See nit 1 for the single accessor that
escapes it.

### 3. Consumers outside `creations/` — swept

Sonnet's 34-site audit covers this repo. Two consumer sets sit outside it:

- **The downstream private repo.** 8 tracked entry points call
  `IREngine::gameLoop()`. I swept every statement between the call and the
  enclosing function's close for any `IR<Module>::` symbol — **zero hits**; the
  apparent matches are function *definitions* placed after `main`, not
  post-loop execution. Read-only `git show` against that clone's `origin/master`
  only — no checkout, no cwd change.
- **gtest targets.** Confirms the plan's finding: no test constructs a `World`
  directly or calls `IREngine::gameLoop()`. The `end()`-must-not-weaken
  argument therefore rests on the exception path and out-of-tree owners, as the
  PR body states — not on in-tree fixtures.

### 4. Sonnet's asked-for value-add: the Lua/sol2 teardown path

Sonnet graded the "Lua-driven creation runs clean" acceptance row on the PR's
own log evidence. I ran it independently on this host (macOS/Metal), from this
branch:

```
fleet-build --target IRLuaPipelineDemo      → built clean
fleet-run IRLuaPipelineDemo --auto-screenshot
  → "Registered dynamic component type=HudMarker with id=413"
  → "AutoScreenshot 1/1: lua_pipeline_demo_zoom1"
  → ir-run: RESULT=CLEAN exe=IRLuaPipelineDemo exit=0
```

`HudMarker` is a runtime Lua-defined component, so this run exercises exactly
the hazard `world.hpp`'s member-order comment guards (T-100: archetype columns
holding `sol::object` refs, `m_lua` declared ahead of `m_entityManager` so the
`sol::state` outlives them). That ordering is untouched by this diff and now
executes at loop tail instead of at exit — clean, exit 0. **Sonnet flagged this
row as taken on faith; on independent execution I confirm it.**

`VideoManager` is the last-declared member, so it destructs *first* — its
finalize-thread join (`video_manager.cpp:28`) happens before any other manager
is torn down. That is the correct order and is unchanged; the only delta is
that `main` now blocks on it at loop tail rather than at exit, which is
deterministic rather than risky.

### On the OpenGL tier

Not tagging a smoke label: the diff touches no path in the cross-host smoke
filter (`engine/render/`, shaders, `engine/system/**`, CMake) — Sonnet reached
the same conclusion and I agree. Adding the reasoning it was missing: the GL
tier is not merely *low risk* here, it is **strictly improved**. The #2031
hazard is a `glDelete*` from a member/observer dtor reaching an unloaded
driver; those dtors now run with the context alive rather than after it. The
only way this could regress GL is a device-object dtor that *assumed* the
context was already gone and no-oped — no such path exists. An OpenGL-tier
confirmation run is still worth having as a matter of record, but it is not a
merge gate on this diff.

### Nits (non-blocking)

1. **`IREngine::runScript` bypasses the assert it now needs**
   (`engine/include/irreden/ir_engine.hpp:96`). It dereferences `g_world->`
   directly instead of going through `getWorld()`, so it is the one
   `IREngine::` entry point that gives a raw null-deref rather than the
   "World not initalized" assert this PR just documented. No in-tree caller
   does this post-loop, but the PR's own new contract text says every accessor
   asserts. One-word fix: `getWorld().runScript(...)`.

2. **`World::end()` still carries the stale claim this PR came to correct**
   (`engine/world/src/world.cpp:292-295`): "only `~World()` would otherwise
   destroy it — but that runs at static destruction, after the GL
   driver/context may already be gone." That is the same now-false sentence
   the diff correctly fixed 200 lines below in `~World()`. Worth the same
   one-line touch-up so the two comments don't contradict each other.

3. **The dtor's own proof-of-life is swallowed.** `~World()` ends with
   `IRE_LOG_INFO("Clean shutdown complete.")`, but `end()` — which always runs
   first — calls `IRProfile::shutdownLogging()`, which flips the
   `g_loggingEnabled` flag every `engLog*` early-returns on
   (`ir_profile.tpp:131-134`, `ir_profile.cpp:18`). The line never prints; my
   run above confirms it empirically (log ends at the screenshot save). This is
   pre-existing and the relative order is unchanged by this PR — but it matters
   *more* now, because the whole point of the change is that `~World()` runs at
   a known, observable moment, and the one diagnostic that would show it did is
   silently dropped. Consider emitting it at the tail of `end()` instead.

### Praise

The exception-path reasoning is the right call and the right documentation of
it. `g_world.reset()` as a plain statement rather than an RAII scope is the
kind of thing a later reader "fixes" into a bug; the comment pre-empts that.
Same for "defense in depth, not a licence to move device frees into the dtor"
in `engine/world/CLAUDE.md` — it blocks the natural-but-wrong follow-up.

🤖 Reviewed by Claude Opus 4.8 (opus-reviewer final pass)

jakildev:
judgement-level resolution.

Conflicted files:
- `engine/CLAUDE.md` — master: `59974e61 docs: global-state rule catalog + header-global ban checks (#2531)`; PR: `9ea0b8b5 engine: deterministic World teardown at gameLoop tail (#2528)`

SHA pair: master=09b4f36a658eb72c896c262801e586978e100d9e × PR=9ea0b8b5e236bb01a308bb502f678b717ee32499

— fleet merger

jakildev:
Resolved semantic conflict in `engine/CLAUDE.md`. The conflict was two additive extensions of the same 'Manager globals' paragraph: master's #2531 appended a pointer to the global-state pattern catalog + header-global ban (`.claude/rules/cpp-globals.md`), while this PR appended the deterministic-teardown note ("Managers are torn down when `gameLoop()` returns"). Both preserved — kept the teardown paragraph and re-appended master's catalog pointer as a closing see-also. C++ replayed unchanged; `fleet-build --target IRShapeDebug` clean on macOS and `IRShapeDebug --auto-screenshot` exits 0 (clean teardown through the new `g_world.reset()`). Rebased onto master 09b4f36a. Reviewer please re-evaluate the rebased diff. — worker

jakildev (commented):
## Review — engine: deterministic World teardown at gameLoop tail (#2528)

**Verdict:** approve (3 nits carried forward, non-blocking)

This PR previously received a full Sonnet pass + Opus final-pass approval
(3 non-blocking nits). The merger then hit a semantic conflict rebasing onto
master `09b4f36a` (docs-only overlap in `engine/CLAUDE.md` against #2531's
global-state catalog pointer), and a worker resolved it, rebuilt, and pushed
`5a728b8f` asking for re-evaluation. This is a re-review of that resolution,
not a fresh first pass.

### Prior-review resolution — verifying the rebase didn't drift the reviewed content

- ✅ **Code is byte-identical to the Opus-reviewed commit.** `git diff
  9ea0b8b5 HEAD -- engine/include/irreden/ir_engine.hpp
  engine/world/src/world.cpp` is empty — the `g_world.reset()` addition and
  the `~World()`/`end()` comment rewrites are exactly what Opus already
  approved, untouched by the rebase.
- ✅ **`engine/CLAUDE.md` conflict resolved correctly.** The worker's stated
  resolution matches the diff: both the PR's new "Managers are torn down
  when `gameLoop()` returns" paragraph and master's #2531 catalog-pointer
  see-also are present, in the order described, with no dropped content from
  either side.
- ✅ **`engine/world/CLAUDE.md`'s further drift is master-side, not a hand
  edit.** The "each manager stamps its own global" rewording at lines
  127-130/406-427 came in from master's own history during the rebase (not
  flagged as a conflict by the merger) and sits in a different section of
  the file from the PR's own hazard-paragraph addition (~line 496) — no
  overlap, no fabricated content.
- ✅ **Build/run verification independently plausible.** Worker's comment
  reports `fleet-build --target IRShapeDebug` clean and
  `IRShapeDebug --auto-screenshot` exit 0 on macOS/Metal post-rebase; given
  the code is byte-identical to the already-independently-rebuilt-and-run
  Opus-reviewed commit, this is expected and consistent.
- ❌ **Opus's 3 nits are still open** (not in scope of the conflict fix, and
  non-blocking per the original verdict — carried forward, not re-raised as
  new findings):
  1. `engine/include/irreden/ir_engine.hpp:97` — `runScript` still
     dereferences `g_world->` directly instead of `getWorld().runScript(...)`,
     bypassing the assert the PR's own new contract promises every accessor
     has.
  2. `engine/world/src/world.cpp:292-295` (current line numbers) — `end()`
     still carries the stale claim "only `~World()` would otherwise destroy
     it — but that runs at static destruction, after the GL driver/context
     may already be gone" — the same sentence the diff correctly retired 200
     lines below in `~World()`.
  3. `engine/world/src/world.cpp` — `~World()`'s `IRE_LOG_INFO("Clean
     shutdown complete.")` is still unreachable in the log because
     `end()` (which always runs first) already flipped `g_loggingEnabled`
     off via `IRProfile::shutdownLogging()`.

### Praise
- The conflict resolution correctly identified the overlap as two additive
  extensions of the same paragraph rather than force-picking one side —
  exactly the judgment call `fleet:semantic-conflict` exists to route to a
  worker instead of auto-merging blindly.

- [x] (verified independently) code diff byte-identical to the Opus-reviewed
  commit `9ea0b8b5` for both `ir_engine.hpp` and `world.cpp`.
- [ ] OpenGL-host smoke run — still outstanding from the original review, not
  newly introduced by this rebase.

Escalation: none — the invariant-critical code is unchanged from what Opus
already fully reviewed and approved; this pass only needed to confirm the
rebase resolution didn't drift it, which it didn't. No fresh Opus recheck
needed.

jakildev:
Addressed the three opus-reviewer nits (all non-blocking) in `ff9e0b4d`:

1. `IREngine::runScript` (ir_engine.hpp) now routes through `getWorld()` instead of dereferencing `g_world->` directly, so it asserts on the cleared global like every other `IREngine::` accessor — no longer the one entry point giving a raw null-deref post-loop.
2. `World::end()`'s stage-timing-observer comment no longer claims `~World()` runs at static destruction (retired for the IREngine path by #2528). Reworded to match the corrected `~World()` comment 200 lines below, so the two no longer contradict.
3. Moved `"Clean shutdown complete."` from `~World()` (swallowed — `end()` always runs first and calls `IRProfile::shutdownLogging()`) to `end()`'s tail, just before logging is disabled. **Verified empirically:** it is now the final log line after the screenshot save on `IRShapeDebug --auto-screenshot` (RESULT=CLEAN, exit 0).

Build clean (macOS/Metal, `fleet-build --target IRShapeDebug`). Code invariant paths unchanged from the previously-approved commit. — worker (pool-1)

jakildev (commented):
## Review — engine: deterministic World teardown at gameLoop tail (#2528)

**Verdict:** approve

Re-review of a follow-up commit (`ff9e0b4d`) that addresses the 3
non-blocking nits from the Opus final-pass review. Not a fresh first pass —
the invariant-critical mechanism (resetting `g_world` at `gameLoop()`'s tail)
was already independently verified by Sonnet, Opus, and a prior rebase
re-review; this pass checks only the delta.

### Verification performed

Diffed the previously-reviewed commit (`5a728b8f`, byte-identical to the
Opus-approved `9ea0b8b5`) against the new head (`ff9e0b4d`):

- ✅ **Nit 1 fixed** — `IREngine::runScript` (`ir_engine.hpp:97`) now calls
  `getWorld().runScript(...)` instead of dereferencing `g_world->` directly.
  Confirmed `getWorld()` (`ir_engine.hpp:18-21`) asserts
  `g_world != nullptr` before returning — `runScript` now shares the same
  post-teardown guard as every other `IREngine::` accessor instead of being
  the one raw-null-deref exception.
- ✅ **Nit 2 fixed** — `World::end()`'s stage-timing-observer comment
  (`world.cpp:290-295`) no longer claims `~World()` runs at static
  destruction. Reworded to correctly describe both paths: safe on the
  `IREngine` path (`g_world` reset at `gameLoop()`'s tail, #2528) but still
  hits a torn-down driver on any path that destructs at process exit
  (#2031) — consistent with the corrected `~World()` comment 200 lines
  below; the contradiction Opus flagged is gone.
- ✅ **Nit 3 fixed** — `"Clean shutdown complete."` moved from `~World()`
  (unreachable — `end()` always runs first) to the tail of `end()`, just
  before `IRProfile::shutdownLogging()` flips `g_loggingEnabled` off.
  **Verified empirically, not just by reading the diff**: built
  `IRShapeDebug` clean (`fleet-build --target IRShapeDebug`, macOS/Metal)
  and ran `fleet-run IRShapeDebug --auto-screenshot 1` — `Clean shutdown
  complete.` is the last log line before `RESULT=CLEAN exit=0`, exactly the
  proof-of-deterministic-teardown the PR's stated intent calls for.

No new nits. All three fixes are mechanical, scoped exactly to what Opus
suggested, and don't touch the reviewed teardown-ordering mechanism itself.

### Praise
- Fix 3 in particular is easy to get subtly wrong (moving a log statement
  across a shutdown-flag flip and *actually checking* it's now reachable,
  rather than just relocating the line and assuming) — the worker's comment
  claims empirical verification and my own run confirms it.

- [x] (verified independently) `fleet-build --target IRShapeDebug` clean,
  `fleet-run IRShapeDebug --auto-screenshot 1` → `RESULT=CLEAN exit=0` with
  `Clean shutdown complete.` as the final log line.
- [ ] OpenGL-host smoke run — still outstanding from the original review,
  unchanged by this nit-fix commit.

Escalation: none — this commit implements exactly the three non-blocking
nits Opus itself raised against code Opus already fully approved; nothing
new to re-derive. No fresh Opus recheck needed.

Opus recheck not required.

PR #2543, opened 2026-07-23
prefabs/render: widget theme becomes a C_WidgetTheme singleton component

### LINKED ISSUE #2527: prefabs/render: widget theme becomes a C_WidgetTheme singleton component; retire the g_defaultTheme header global

jakildev:
## Summary
- Retires `IRPrefab::Widget::g_defaultTheme` (a mutable header-scope inline global) in favor of a `C_WidgetTheme` singleton component, per the "per-world settings" pattern in `engine/entity/CLAUDE.md` §"Singleton components".
- `defaultTheme()` keeps its exact signature; the 13 widget-render/input systems that previously called it once per entity inside `tick()` now cache the theme once per frame in `beginTick()`.
- `C_WidgetTheme` opts OUT of world-snapshot save (creation-authored init config, re-established by setup code every run).

## Test plan
- [x] `fleet-build --target IrredenEngineTest` — clean, 1356/1356 tests pass (including `SaveTrait.InventoryIsComplete`)
- [x] `python3 scripts/gui-verify.py IRVoxelEditor` — 14/14 assertions pass (hover/click/checkbox/slider/pick all still route through the widget framework correctly)
- [x] Visual spot-check of a GUI-heavy screenshot (voxel editor panels/sliders/dropdowns) — renders identically to before

Closes #2527

### DIFF
--- a/engine/prefabs/irreden/render/CLAUDE.md
+++ b/engine/prefabs/irreden/render/CLAUDE.md
@@ -355,8 +355,11 @@ kind-specific data component (`C_WidgetPanel` / `C_WidgetLabel` /
 `C_WidgetButton` / `C_WidgetSlider` / `C_WidgetCheckbox` /
 `C_WidgetList` / `C_WidgetDropdown` / `C_WidgetRadio` /
 `C_WidgetTextInput` / `C_WidgetScroll` / `C_WidgetColorSwatch`).
-Theme lives in `widget_theme.hpp::defaultTheme()` — a single inline
-header-level instance a creation mutates once at init.
+Theme lives on the `C_WidgetTheme` singleton component, read via
+`widget_theme.hpp::defaultTheme()` — a creation mutates it once at init
+before constructing widgets. Per-tick readers (`WIDGET_RENDER_*` systems)
+cache the theme in `beginTick` rather than reading the singleton per
+entity.
 
 `C_WidgetColorSwatch` (T-211) is a single solid-color clickable cell
 for palette panels and theme editors. Each swatch carries its own
--- a/engine/prefabs/irreden/render/CLAUDE.md
+++ b/engine/prefabs/irreden/render/CLAUDE.md
@@ -356,10 +356,16 @@ kind-specific data component (`C_WidgetPanel` / `C_WidgetLabel` /
 `C_WidgetList` / `C_WidgetDropdown` / `C_WidgetRadio` /
 `C_WidgetTextInput` / `C_WidgetScroll` / `C_WidgetColorSwatch`).
 Theme lives on the `C_WidgetTheme` singleton component, read via
-`widget_theme.hpp::defaultTheme()` — a creation mutates it once at init
-before constructing widgets. Per-tick readers (`WIDGET_RENDER_*` systems)
-cache the theme in `beginTick` rather than reading the singleton per
-entity.
+`widget_theme.hpp::defaultTheme()` — a creation may mutate it once at init
+before constructing widgets. Every `WIDGET_RENDER_*` system's `create()`
+calls `widget_theme.hpp::ensureThemeSingleton()`, so the singleton
+reliably exists before frame 1 regardless of whether a creation
+customizes it — that main-thread registration-time touch keeps the
+lazy-create off the per-frame `beginTick` path (`singleton<T>()`'s first
+call must be main-thread; a non-main-thread `createEntity` defers the row
+insertion until `flushStructuralChanges`). Per-tick readers
+(`WIDGET_RENDER_*` systems) cache the theme in `beginTick` rather than
+reading the singleton per entity.
 
 `C_WidgetColorSwatch` (T-211) is a single solid-color clickable cell
 for palette panels and theme editors. Each swatch carries its own
--- a/engine/prefabs/irreden/render/components/component_widget_theme.hpp
+++ b/engine/prefabs/irreden/render/components/component_widget_theme.hpp
@@ -1,0 +1,86 @@
+#ifndef COMPONENT_WIDGET_THEME_H
+#define COMPONENT_WIDGET_THEME_H
+
+#include <irreden/ir_math.hpp>
+
+namespace IRPrefab::Widget {
+
+// Visual styling shared by every widget in a creation. Backs the
+// `C_WidgetTheme` singleton below; widgets read from it at render time. The
+// values lock in the look of the editor for every Phase 0 — Phase 10
+// implementation that builds on top, so additions should be additive
+// (new fields default to existing-color values) and renames are
+// breaking.
+struct WidgetTheme {
+    // Surfaces
+    IRMath::Color backgroundIdle_ = IRMath::Color{40, 44, 56, 255};
+    IRMath::Color backgroundHover_ = IRMath::Color{60, 68, 88, 255};
+    IRMath::Color backgroundPressed_ = IRMath::Color{90, 100, 124, 255};
+    IRMath::Color backgroundDisabled_ = IRMath::Color{32, 34, 40, 255};
+    IRMath::Color panelBackground_ = IRMath::Color{24, 26, 34, 220};
+    IRMath::Color panelTitleBackground_ = IRMath::Color{52, 58, 76, 255};
+
+    // Borders
+    IRMath::Color borderIdle_ = IRMath::Color{96, 108, 132, 255};
+    IRMath::Color borderHover_ = IRMath::Color{160, 200, 240, 255};
+    IRMath::Color borderPressed_ = IRMath::Color{220, 240, 255, 255};
+    IRMath::Color borderFocused_ = IRMath::Color{240, 200, 120, 255};
+    IRMath::Color borderDisabled_ = IRMath::Color{64, 68, 80, 255};
+
+    // Text
+    IRMath::Color textIdle_ = IRMath::Color{220, 224, 232, 255};
+    IRMath::Color textDisabled_ = IRMath::Color{120, 124, 132, 255};
+
+    // Slider track / thumb
+    IRMath::Color sliderTrack_ = IRMath::Color{72, 80, 96, 255};
+    IRMath::Color sliderThumbIdle_ = IRMath::Color{160, 168, 180, 255};
+    IRMath::Color sliderThumbHover_ = IRMath::Color{210, 220, 240, 255};
+
+    // Checkbox fill (when checked)
+    IRMath::Color checkboxFill_ = IRMath::Color{160, 200, 240, 255};
+
+    // List / dropdown row backgrounds (over the standard surface fill).
+    IRMath::Color listRowSelected_ = IRMath::Color{72, 96, 140, 255};
+    IRMath::Color listRowHover_ = IRMath::Color{52, 60, 80, 255};
+
+    // Radio button selected-fill (centre dot).
+    IRMath::Color radioFill_ = IRMath::Color{160, 200, 240, 255};
+
+    // Text input cursor color.
+    IRMath::Color textInputCursor_ = IRMath::Color{220, 224, 232, 255};
+
+    // Scroll bar visuals.
+    IRMath::Color scrollTrack_ = IRMath::Color{32, 36, 46, 255};
+    IRMath::Color scrollThumb_ = IRMath::Color{120, 132, 156, 255};
+    IRMath::Color scrollThumbHover_ = IRMath::Color{180, 200, 232, 255};
+
+    // Layout
+    int borderThickness_ = 2;
+    int padding_ = 4;
+    int fontSize_ = 2;
+    int sliderTrackThickness_ = 6;
+    int sliderThumbWidth_ = 8;
+    int checkboxBoxSize_ = 16;
+    int radioBoxSize_ = 16;
+    int textInputCursorWidth_ = 2;
+    int scrollBarThickness_ = 10;
+    int scrollThumbMinExtent_ = 16;
+    int textCursorBlinkPeriodFrames_ = 30;
+};
+
+} // namespace IRPrefab::Widget
+
+namespace IRComponents {
+
+// Singleton row backing `IRPrefab::Widget::defaultTheme()`
+// (`widget_theme.hpp`). Setup-time-only: a creation mutates the theme once
+// at init before constructing widgets. Per-tick readers must cache it in
+// `beginTick` rather than reading the singleton per entity (the ECS
+// footgun) — see the `WIDGET_RENDER_*` systems for the pattern.
+struct C_WidgetTheme {
+    IRPrefab::Widget::WidgetTheme theme_;
+};
+
+} // namespace IRComponents
+
+#endif /* COMPONENT_WIDGET_THEME_H */
--- a/engine/prefabs/irreden/render/components/component_widget_theme.hpp
+++ b/engine/prefabs/irreden/render/components/component_widget_theme.hpp
@@ -73,10 +73,13 @@ struct WidgetTheme {
 namespace IRComponents {
 
 // Singleton row backing `IRPrefab::Widget::defaultTheme()`
-// (`widget_theme.hpp`). Setup-time-only: a creation mutates the theme once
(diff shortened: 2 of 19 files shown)


jakildev (commented):
## Review — prefabs/render: widget theme becomes a C_WidgetTheme singleton component

**Verdict:** approve

Clean, mechanical refactor. Retires the header-scope `inline WidgetTheme g_defaultTheme` global in favor of `IREntity::singleton<C_WidgetTheme>()`, matching the "per-world settings" singleton-component pattern documented in `engine/entity/CLAUDE.md` §"Singleton components" (citation checked — the section and the "per-world settings" phrase both exist there).

Went through the checklist:

- **Ownership/lifetime** — `defaultTheme()` still returns `WidgetTheme&`, but every call site immediately copies into a per-system `theme_` value member in `beginTick()` rather than holding the reference across `tick()` — no dangling-reference risk across ticks, and it correctly moves the singleton lookup out of the per-entity hot path (the ECS footgun the entity/render CLAUDE.md docs call out).
- **Singleton lazy-create ordering** — `IREntity::singleton<T>()` lazy-creates on first call. Since creation setup code (which mutates `defaultTheme()` once at init) runs before the game loop's first tick, the singleton is created-and-customized before any system's `beginTick()` can lazy-create it with defaults. No race between setup-time customization and first-frame reads.
- **Save/serialization** — `C_WidgetTheme` is `IR_SAVE_OPT_OUT`'d with a clear comment explaining why (creation-authored init config, re-established every run). Verified precedent: `C_SpatialIndex` is another singleton component that's also `IR_SAVE_OPT_OUT`'d (`save_component_inventory.hpp:163`), so singleton + opt-out is an established combination, not something novel here. `kExpectedEngineComponentCount` bumped 165→166 to match the new tuple entry.
- **Naming** — `WidgetTheme` fields and `C_WidgetTheme::theme_` are public members with trailing `_`, consistent with existing System-struct/component conventions.
- **Completeness** — grepped the tree for remaining `defaultTheme()` call sites; all 13 widget systems now cache `theme_` in `beginTick()` and use it in `tick()`, no straggler still reading the accessor per-entity.
- **Visual behavior** — color/layout constant values were moved verbatim from the old struct, not edited, so this should be pixel-identical (consistent with the PR's "renders identically" spot-check claim).

### Nits
- No dedicated unit test exercises the new singleton wiring itself (just the general `SaveTrait.InventoryIsComplete` + `gui-verify`'s functional coverage). Given this is a pure storage-mechanism refactor behind an unchanged `defaultTheme()` signature, and gui-verify's 14/14 assertions already exercise theme-dependent rendering end-to-end, I don't think this rises to needs-fix — but a small `EntityTest`/`SaveTrait` case asserting `defaultTheme()` round-trips through `IREntity::singleton<C_WidgetTheme>()` would be cheap insurance if a future PR touches this again.

### Praise
- Caching the theme once per frame in `beginTick()` instead of once per entity in `tick()` is exactly the ECS-footgun-avoidance pattern the docs ask for, and it's applied uniformly across all 13 systems in one pass rather than partially.

- [x] Already covered by the PR's documented test plan (build, full test suite, gui-verify, visual spot-check).

Escalation: please Opus-recheck before merge (touches: engine/world/ save-inventory registration for a new singleton component, and a lifetime/lifecycle pattern change — global → ECS-managed singleton — across engine/prefabs/irreden/render/).

jakildev (commented):
## Opus recheck — prefabs/render: widget theme becomes a `C_WidgetTheme` singleton component (#2527)

Sonnet's checklist pass was solid and I'm not re-deriving it (naming,
`shared_ptr`, per-entity `getComponent`, save-inventory bookkeeping — all
clean on a re-read of the diff; the 165→166 bump matches the new
`AllEngineComponents` tuple entry and the `IR_SAVE_OPT_OUT` sits in the
right block, `save_component_inventory.hpp:322-326` / `:507`).

The lifetime question is the whole PR, so this pass spends its budget
there: **what actually happens to the singleton across scene teardown**,
**what the lazy-create really does and when it fires**, and **whether the
returned `WidgetTheme&` can dangle** now that it points into archetype
column storage rather than at an address-stable global.

### 1. Scene-transition survival — checked, and it's the thing that would have made this refactor wrong

The old `inline WidgetTheme g_defaultTheme` outlived everything. A
component row does not, so the question is whether a customized theme
survives a scene change. It does, and by explicit design:
`destroyAllExceptPreserved` (`entity_manager.hpp:148-158`) exempts
"(1) singleton entities (everything in `m_singletonEntityByComponent`)"
and — the load-bearing half — does **not** clear the singleton cache, so
the preserved entity stays valid for the next scene. Only full
`destroyAllEntities` drops it, at which point everything is being rebuilt
anyway and the lazy re-create on next access is correct.

So global → singleton is not a lifetime regression across scenes. Worth
stating explicitly because it's the failure mode this refactor could most
plausibly have introduced and it isn't visible in the diff.

### 2. The returned reference cannot dangle — because the systems copy, not alias

`defaultTheme()` still returns `WidgetTheme&`, but it now points into an
archetype column, which is *not* address-stable across entity creation.
Verified all 13 systems store a **value** member (`IRPrefab::Widget::WidgetTheme
theme_;`, e.g. `system_widget_render_button.hpp:19`) and assign by copy in
`beginTick` (`:26`), then read only `theme_` in `tick`. No system holds a
`WidgetTheme&` across anything. Confirming Sonnet here — and noting the
distinction matters more than it looks: a `WidgetTheme&` member instead of
a value member would compile identically and dangle the first time a
widget entity got spawned. See nit 3 on the doc wording.

### 3. The new singleton entity can't pollute a hot archetype

`getOrCreateSingleton` creates via `createEntity(Component{})`
(`entity_manager.hpp:539`), and `createEntity` derives the archetype from
exactly the passed pack — `getArchetype<Components...>()`
(`entity_manager.hpp:187-189`) — with no auto-added components. So the
singleton lands in a `{C_WidgetTheme}`-only archetype that matches no
widget, transform, or render system's include set. No stray row in any
per-frame iteration, and nothing renders it.

### Nits

**1. Sonnet's ordering argument is correct but vacuous, and it inverts
which path deserved the scrutiny.**

Sonnet wrote: "Since creation setup code (which mutates `defaultTheme()`
once at init) runs before the game loop's first tick, the singleton is
created-and-customized before any system's `beginTick()` can lazy-create
it with defaults. No race between setup-time customization and first-frame
reads."

I grepped for `defaultTheme()` across `creations/` + `engine/`: outside the
13 widget systems, the only hits are the accessor itself
(`widget_theme.hpp:15`) and two doc comments. **No creation mutates the
theme at init.** So the setup-then-tick ordering being reasoned about never
occurs today, and the path that actually runs in every in-tree GUI creation
is the branch Sonnet treated as the fallback: the first widget system's
`beginTick()` on frame 1 lazy-creates the singleton entity *mid-render-
pipeline*.

That path is genuinely exercised (gui-verify 14/14 on the voxel editor,
`IRShapeDebug` clean), so this is not a defect — the conclusion "safe"
happens to hold for the branch that runs. But the reasoning rested on the
branch that doesn't, and the PR's own docs assert the customization step as
if it happens (`component_widget_theme.hpp:76-78`,
`engine/prefabs/irreden/render/CLAUDE.md:358-359`, both "a creation mutates
it once at init"). Either soften those to "may mutate", or — better, and it
also disposes of nit 2 — have widget bootstrap touch `defaultTheme()` once
at init so the singleton reliably exists before frame 1, making the docs
true and taking the lazy-create off the pipeline path entirely.

**2. The `IR_SAVE_OPT_OUT` justification cites a mechanism that doesn't
exist, which matters for the forward case.**

`save_component_inventory.hpp:322-324` reads: "creation-authored init config
re-established by the creation's setup code on every run; persisting it in a
world snapshot would fight that init-time mutation on load." Per nit 1, no
setup code re-establishes anything. The **decision** is right — the theme is
pure defaults every run, so a snapshot would add nothing — but the stated
reason isn't the actual one.

This isn't pedantry, because the reason is what a future reader will apply.
`C_WidgetColorSwatch` already exists "for palette panels and theme editors"
(`engine/prefabs/irreden/render/CLAUDE.md:365`). The moment a theme editor
mutates the theme at *runtime*, opt-out silently drops those edits from the
world snapshot — and the current comment reads as if that case was
considered and is fine, when it wasn't. Suggest: "no runtime mutation today
(defaults only); revisit if a theme editor makes this user-authored state."

**3. This PR is the first thing to put `singleton<T>()` on a per-frame
system path, and that's a sharp edge worth pinning (latent, not live).**

`singleton<T>()` is `getComponent<T>(singletonEntity<T>())`
(`ir_entity.hpp:299-301`) — it dereferences the id **immediately** after
`getOrCreateSingleton` creates it. But `createEntity` from a non-main
thread takes the deferred path: it returns an atomically-allocated id and
stages the archetype-row insertion until `flushStructuralChanges`
(`entity_manager.hpp:179-186`). So off the main thread, the lazy-create
returns an id whose row does not exist yet and the immediately-following
`getComponent` is looking for nothing.

Not reachable today — I checked all 13 systems and none declares
`ParallelSafe` / `Spawns` / `MainThread` / `Destroys`, so they're serial,
and the exposure is frame 1 only. But the exposure is new: before this PR
`defaultTheme()` was a global read and could not create anything. If any
widget system is ever marked `ParallelSafe`, first-frame theme access
breaks in a way that will be very hard to attribute back to here. The init
touch suggested in nit 1 closes it outright; failing that, a "main thread
only" note on `singleton<T>()` would.

(For the record, `Spawns` would be the honest tag on these systems now,
since `beginTick` can transitively create an entity. It's documentation-
only today — `system_access.hpp:20` still says "**Unused in this phase.**
Pure unit-test surface for now" — so I'm not asking for it.)

**4. Sharpening Sonnet's test nit.** Sonnet suggested a case asserting
`defaultTheme()` round-trips through `IREntity::singleton<C_WidgetTheme>()`.
That one is close to tautological — it re-asserts the accessor's one-line
body. The invariant this refactor actually changes, and the one with a real
mechanism behind it (`destroyAllExceptPreserved`'s singleton exemption, §1),
is: **a mutated `defaultTheme()` survives a scene transition.** Mutate,
run the scene-transition teardown, assert the value persisted. That case
would have caught the plausible version of this refactor that gets the
preserve-list wrong. Agreeing with Sonnet that neither rises to needs-fix.

### Verdict rationale

No correctness defect. The two invariants the diff can't show — singleton
survival across scene teardown, and no dangling reference into column
storage — both hold, and for principled reasons rather than by accident.
The nits are documentation accuracy plus one latent thread-safety edge that
a single init-time `defaultTheme()` touch would close along with nit 1's
doc mismatch; none gates merge.

Escalation: none.

jakildev:
Addressed all 3 nits from the opus recheck: every `WIDGET_RENDER_*` system's `create()` now calls `ensureThemeSingleton()` so the singleton lazy-creates on the main thread during pipeline wiring (before frame 1), closing the latent worker-thread hazard and making the docs' setup-mutation claim accurate (softened to "may mutate"). Corrected the `IR_SAVE_OPT_OUT` comment to state the real reason. Added `test/render/widget_theme_singleton_test.cpp` pinning the invariant that a mutated `defaultTheme()` survives `resetGameplay()`'s scene-transition teardown. Full suite green (1358/1358), clang-format clean. — worker

jakildev (commented):
## Review — prefabs/render: widget theme becomes a C_WidgetTheme singleton component (re-review)

**Verdict:** approve

Re-review of commit `c1140b78` (pushed after fleet approval + Opus recheck), which addresses the 3 nits from the Opus recheck.

### Prior-review resolution
- ✅ `engine/prefabs/irreden/render/widget_theme.hpp`, `component_widget_theme.hpp`, `engine/prefabs/irreden/render/CLAUDE.md:358-366` — Opus nit 1 (docs claimed a creation always mutates the theme at init, but the path that actually runs in every in-tree GUI creation is the first widget system's `beginTick()` lazy-creating on frame 1) — **verified fixed at c1140b78**: every theme-reading system's `create()` now calls the new `ensureThemeSingleton()` (`widget_theme.hpp:19-26`), and the doc/comment wording is softened to "may mutate."
- ✅ `engine/world/include/irreden/world/save_component_inventory.hpp:322-324` — Opus nit 2 (`IR_SAVE_OPT_OUT` comment cited a setup-mutation mechanism that doesn't exist) — **verified fixed at c1140b78**: comment now states the real reason ("pure defaults every run, no creation mutates it at runtime today... revisit if a theme editor makes this user-authored state").
- ✅ Latent worker-thread `singleton<T>()` hazard (Opus nit 3: if any widget system is ever marked `ParallelSafe`, a first-frame `beginTick` lazy-create from a worker thread would return an id whose row isn't flushed yet) — **verified fixed at c1140b78**: `ensureThemeSingleton()` now runs from each system's `create()`. Confirmed `create()` executes synchronously on the main thread — e.g. `creations/editors/voxel_editor/main.cpp:2358` calls `IRSystem::createSystem<IRSystem::WIDGET_RENDER_BUTTON>()` directly from `main()`, well before any worker pool ticks — so the lazy-create is off the per-frame path entirely, not just off the hot path.
- ✅ Sonnet's original nit (no dedicated unit test for the singleton wiring itself) — **verified fixed at c1140b78**: new `test/render/widget_theme_singleton_test.cpp` covers both `ensureThemeSingleton()` creating the row and a mutated `defaultTheme()` surviving `resetGameplay()`'s scene-transition teardown — the actual invariant this refactor depends on (per Opus's §1 finding), not a tautological accessor round-trip.

### New commit review
Confined to the 3 nit-fixes above: `ensureThemeSingleton()` (a one-line wrapper calling `defaultTheme()` to force the lazy-create) is called uniformly from all 13 touched systems' `create()`, the two doc/comment corrections are accurate, and the new test is well-scoped. No new findings.

### Verification performed
- `fleet-build --target IrredenEngineTest` — clean build on macOS.
- `./build/IrredenEngineTest --gtest_filter="WidgetThemeSingletonTest.*"` — 2/2 pass.
- Full suite: 1358/1358 pass, matching the PR body's claim.
- `clang-format` — no drift on any touched file.

### Praise
- Choosing the "make the doc true" fix (init-time touch) over "just soften the doc" for nit 1 — it closes nit 3's latent thread-safety edge as a side effect instead of leaving it latent, exactly the option Opus's nit 1 suggested was "better."

Issue #2548, opened 2026-07-24
render: cursor-pivot rotate — true-depth click latch + visual pivot indicator + shape_debug mode toggle (P4)


jakildev:
**Model:** opus
**Part of epic:** #2544
**Plan file:** `~/.fleet/plans/issue-2544.md` (full epic plan)
**Blocked by:** #2547

## Scope

Polish the cursor-pivot rotate mode into the debuggable, discoverable feature the human asked for: rotate about the point you clicked, at its true surface depth, with a visible marker at the latched pivot. Today `System<CAMERA_MOUSE_ROTATE>`'s Ctrl+Shift+middle-drag latches `mouseWorldPos3DAtIsoDepth(0)` — an iso-depth-0 point with the same wrong-depth defect as the default pivot — and gives no visual feedback.

## Approach

- Latch the clicked SURFACE point: `IRPrefab::Picking::castVoxelRay` hit (`worldHitPos_`), falling back to the #2547 depth-probe center derivation on miss (background click).
- Visual pivot indicator: a small always-on-top marker entity at the latched pivot for the drag duration + a short fade after release. Use behavior-named render flags (e.g. `SHAPE_FLAG_XRAY_OCCLUDED`) so it reads through occluding geometry; prefab-scoped (a small pivot-indicator prefab or `IRPrefab::Camera` helper), NOT a RenderManager feature field.
- shape_debug: a runtime key that cycles rotation-pivot mode (screen-center ↔ cursor-latched), logged to console, so the modes compare live without restarting. (The engine-wide "demo help overlay / settings menu" standardization is filed separately — this child ships only the one demo toggle.)
- Apply the same true-depth latch to `system_camera_key_drag_rotate.hpp` if it grows a cursor-pivot variant; otherwise leave it on the default pivot.

## Acceptance criteria

- Clicking a voxel surface and dragging rotates the scene about the clicked point (the clicked feature holds its screen position). Verify deterministically: extend `pivot-verify.py` with a `cursor-latch` block that drives `setRotationPivotFocus` from a `castVoxelRay` hit and asserts PINNED at 3+ yaws.
- The indicator appears at the clicked point, renders visibly over occluders, and despawns after release (screenshot pair in the PR).
- No behavior change with the mode untouched: existing camera-control defaults preserved; gui-verify green if the demo has a GUI-test table.
- ir-run RESULT=CLEAN.

## Dependencies

[...]

Issue #2547, opened 2026-07-24
render: depth-aware CAMERA_CENTER default focus — pin the content under the viewport center (P3)


jakildev:
**Model:** opus
**Part of epic:** #2544
**Plan file:** `~/.fleet/plans/issue-2544.md` (full epic plan)
**Blocked by:** #2546

## Scope

The default `RotationPivotMode::CAMERA_CENTER` focus is the iso-depth-0 point under the viewport center, so the pivot pins that point's vertical column — content actually AT screen center at any other depth orbits under rotation (336px measured at z=10, zoom 4). Make the default pivot pin the content under the viewport center at its rendered depth, so "rotate" means "what I'm looking at spins in place" regardless of scene elevation. This is the user-reported core symptom.

## Approach

Architect decision (interactive session, 2026-07-22):

- Derive the default focus depth from the composite depth at the viewport-center pixel via the `IRPrefab::DepthProbe` readback primitive (#1910; decodes to shared trixel-distance units). Background / no-hit falls back to the current iso-depth-0 point.
- Latch policy (deterministic, gesture-free): re-derive the focus only while `visualYaw` is NOT changing between frames; while yaw is in motion, keep the latched focus. Works identically for mouse-drag, key, and programmatic `setYaw` (auto-screenshot needs no gesture plumbing). Pan or zoom changes re-derive immediately.
- Readback cost gate: only read back when yaw is settled AND pan/zoom changed since the last derive — steady state does zero readbacks.
- `RotationPivotMode::ORIGIN` and explicit `setRotationPivotFocus` overrides unchanged.
- Same PR: correct the "z = 0 world point" mis-statements in `ir_render.cpp` (getEffectiveCameraIso comment) and `docs/design/camera-yaw-pivot.md`, and document the depth-aware contract + latch policy in that design doc (it is the source of truth).

## Acceptance criteria

- `pivot-verify.py --blocks center-depth` PINNED (≤1.5px, all 9 yaws, zoom 4 + 8).
- center-column block still PINNED when the center pixel is background (fallback path) — add a background-fallback variant to the harness if needed.
- Interactive sanity: pan over an elevated shape in shape_debug, Alt+drag rotate — the shape under the center rotates in place (before/after screenshot pair in the PR).
- Cardinal yaw-0 byte-identity (correction is identity at yaw 0 by construction); ir-run RESULT=CLEAN.

## Dependencies

[...]

Issue #2546, opened 2026-07-24
render: per-axis composite registration under non-integer effective camera offset (P2)


jakildev:
**Model:** opus
**Part of epic:** #2544
**Plan file:** `~/.fleet/plans/issue-2544.md` (full epic plan)
**Blocked by:** #2545

## Scope

With the half-cell anchor mismatch (#2545) fixed, the voxel path still renders every residual-yaw (non-cardinal) frame at a constant ≈(−1, −1.1) iso px offset from the cardinal frames — measured (−16, −9)px at zoom 4, the same at every residual yaw sign/bracket, SDF probe clean. The effective camera offset (`IRRender::getEffectiveCameraIso`) is non-integer whenever yaw ≠ 0 and a per-axis stage rounds it under a different convention than the cardinal gather.

## Approach

- Localize the divergence: `perAxisBase_` / `faceLocalAnchor` derivation in `system_trixel_to_framebuffer.hpp`, the scatter corner math in `v_peraxis_scatter.glsl`, and the framebuffer pan — versus the cardinal gather's `floor(cameraTrixelOffset)` convention in the stage kernels.
- Unify: apply the pivot correction to the float offset BEFORE any per-stage rounding, and use the same rounding function (`roundHalfUp` vs `floor` mismatch is the prime suspect) on both routes. This is the documented #1944 gotcha; find where the per-axis route deviates and pin it with a comment citing this issue.

## Acceptance criteria

- Full-sweep `pivot-verify.py --blocks focus-ctr,focus-off,center-column` PINNED (all 9 yaws, ≤1.5px) at zoom 4 AND zoom 8.
- `--pan-sweep` and `--yaw-sweep` stay SMOOTH (no #1944/#2427 regression).
- Cardinal yaw-0 byte-identity preserved (render-verify references).
- ir-run RESULT=CLEAN.

## Dependencies

Blocked by #2545 (its orbit dominates this offset in the metric). Blocks the depth-aware default-focus child.

## Gotchas

- Zoom-dependent: the offset measured differently at zoom 8 vs 4 during diagnosis — verify at both, plus one subdivision variant.
- Don't regress the #2469 scatter positioning wobble accounting or the overflow-lane 2x2 neighborhood compare (#2427 fix) — both live in the same code region.

## References

- Epic #2544 (the +0.5-focus run isolating this residual: cardinals (0,0), non-cardinals constant (−16,−9)px).
[...]

Issue #2545, opened 2026-07-23
render: unify voxel-raster rotation anchor with the exact-position convention — half-cell orbit under yaw (P1)


jakildev:
**Model:** fable
**Part of epic:** #2544
**Plan file:** `~/.fleet/plans/issue-2544.md` (full epic plan)
**Blocked by:** #2553

## Scope

Unify the engine's rotation-anchor convention: an entity's `position` is the point the scene rotates about under camera Z-yaw, for BOTH render paths. Today the voxel raster rotates content about `position + (0.5, 0.5, 0.5)` (a half-cell riding the iso depth axis — invisible at cardinal yaw 0) while the SDF path, the CPU pivot math (`IRMath::cameraYawPivotOffset`), and the picking inverses rotate about the exact position. Under yaw the voxel layer orbits any pinned focus by ~1 iso px and the voxel/SDF layers counter-rotate half a cell apart.

## Approach

- Read the yawed reposition chain end-to-end first: `c_voxel_to_trixel_stage_1/2` emit anchor, `v_peraxis_scatter` corner reconstruction, `c_voxel_visibility_compact`, the overflow-lane taps, `RESOLVE_PER_AXIS_SCREEN_DEPTH` — GLSL + Metal twins.
- Then pick ONE of the two candidate mechanisms in the epic plan (kernel-side cell-center-frame rotation vs. convention-side `+h` adoption everywhere else) by the stated criterion: preserves cardinal yaw-0 byte-identity with the smaller audited consumer blast radius. The epic plan leans kernel-side (authoring places voxel centers at integer positions; the raster is the deviant).
- Add a `--cardinals-only` switch to `scripts/pivot-verify.py` so this child's gate scores the cardinal-yaw subset (the non-cardinal residual is #C's, the next child's, gate).

## Acceptance criteria

- `pivot-verify.py --blocks focus-ctr,focus-off,center-column --cardinals-only`: voxel blocks PINNED (≤1.5px) at zoom 4 and 8.
- focus-ctr-sdf stays ≤2px; a mixed voxel+SDF probe at one center shows no relative layer drift at cardinal yaws.
- Cardinal fast path byte-identical: shape_debug render-verify yaw-0 shots img_diff = 0 vs current references.
- `--pan-sweep` / `--yaw-sweep` jitter gates stay SMOOTH; ir-run RESULT=CLEAN.

## Dependencies

[...]

PR #2553, opened 2026-07-24
test: pivot-verify — deterministic rotation-pivot invariance harness (epic #2544 tooling)

jakildev:
## Summary

- Deterministic rotation-pivot invariance harness — the machine check the camera Z-yaw pivot contract (`docs/design/camera-yaw-pivot.md`) never had. A sweep can pass the existing SMOOTH gate and cardinal byte-identity while pinning the wrong point: a slow orbit fits the jitter line-fit, and a `(0.5,0.5,0.5)` anchor offset rides the iso depth axis (zero projection at yaw 0).
- `tools/jitter_probe`: new `--stationary` / `--max-deviation` mode — verdict `PINNED`/`DRIFT` on max centroid deviation vs frame 0.
- `IRShapeDebug --pivot-verify <block>` (+ `--pivot-verify-sdf`): one isolated vertical voxel/SDF cylinder probe per run (Z-yaw-invariant silhouette; whole-frame threshold mask — a color-locked mask tracks yaw-rotating lit faces and fabricates an orbit on pinned geometry), swept through cardinal + non-cardinal yaws at fixed pan. Blocks discriminate explicit-focus vs default pivot, and pinned-column vs at-depth-screen-center placement.
- `scripts/pivot-verify.py`: build → run → score wrapper (verify_common family); pass/fail table; **no committed reference images** — the assertion is pure temporal invariance.
- Docs: `engine/render/CLAUDE.md` gains the pivot-invariance third-property note; `docs/design/camera-yaw-pivot.md` gains §Known deviations; `tools/jitter_probe/README.md` documents the new mode.

## Baseline this harness measures on master (macos-debug, zoom 4, 1.5px threshold)

| pass | verdict | dev_x(px) | dev_y(px) |
|---|---|---|---|
| focus-ctr (voxel) | DRIFT | 29.13 | 16.00 |
| focus-ctr-sdf | DRIFT | 2.00 | 2.00 |
| focus-off (voxel) | DRIFT | 29.13 | 16.00 |
| center-column (voxel) | DRIFT | 29.13 | 16.00 |
| center-depth (voxel) | DRIFT | 336.00 | 336.00 |

Red is the point: the harness asserts the contract, and the three defects it isolates are filed as epic #2544 (P1 #2545 half-cell anchor → P2 #2546 per-axis registration → P3 #2547 depth-aware default focus → P4 #2548 cursor-pivot latch + indicator). Each child's fix flips its block(s) to PINNED; `pivot-verify.py` exits non-zero until then, so it is a diagnosis/gate tool, not a CI-green suite yet.

## Visual-change note (why no screenshot pair)

The demo diff is harness-only and flag-gated: with `--pivot-verify` absent the scene selection, shot tables, and defaults are untouched, so every existing capture is byte-identical (verified: default-suite references unaffected; the probe scene only spawns under the new flag). Nothing visual changed on any default path.

## Test plan

- [x] `fleet-build --target IRShapeDebug jitter_probe` green (macos-debug)
- [x] `python3 scripts/pivot-verify.py` runs all five passes, RESULT=CLEAN each, produces the baseline table above (expected exit 1 while #2544 is open)
- [x] `jitter_probe --stationary` PINNED/DRIFT verdicts verified against hand-checked centroid tables (cardinal-orbit and 336px cases)
- [x] `ruff check scripts/` green; `format-changed` applied
- [ ] Linux/GL twin run of the baseline (cross-host smoke)

### DIFF
--- a/creations/demos/shape_debug/main.cpp
+++ b/creations/demos/shape_debug/main.cpp
@@ -256,6 +256,76 @@ std::vector<std::array<char, 40>> g_pivotFocusShotLabels;
 // the focus exactly matches the rendered geometry.
 constexpr vec3 kPivotPillarCenter = vec3(8.0f, -8.0f, 10.0f);
 
+// --pivot-verify <block>: deterministic rotation-pivot invariance harness.
+// Each block spawns ONE vertical voxel cylinder (Z-yaw-invariant silhouette, so
+// its whole-silhouette centroid is immune to face-triplet changes) and sweeps
+// the camera yaw through cardinals + non-cardinals at a fixed pan; jitter_probe
+// --stationary then asserts the centroid holds its frame-0 position. Blocks:
+//   focus-ctr     — explicit focus on the probe center, probe at screen center.
+//   focus-off     — explicit focus on the probe center, probe panned off-center
+//                   (catches errors that vanish exactly at screen center).
+//   center-column — DEFAULT pivot; probe on the pinned vertical column (the
+//                   depth-0 focus's xy) at z > 0. Verifies the implemented
+//                   contract: the pivot correction pins the column
+//                   {W : W.xy == F.xy}, so this probe must hold even though it
+//                   renders below screen center.
+//   center-depth  — DEFAULT pivot; probe AT the viewport center but at z > 0:
+//                   on the center iso RAY, off the pinned column by (z, z) in
+//                   xy. Encodes the user-facing contract "what is at screen
+//                   center stays at screen center while rotating". DRIFT here
+//                   with center-column PINNED means the default pivot pins the
+//                   wrong DEPTH, not that the drift-cancel math is broken.
+// Requires --auto-screenshot (value = warmup frames per shot); score with
+// scripts/pivot-verify.py. Same stable-storage discipline as the other tables.
+std::string g_pivotVerifyBlock = "off";
+// --pivot-verify-sdf: probe via the SDF solver instead of the voxel pool, so
+// the two render paths' pivot conventions can be compared A/B — a layer that
+// pins while the other orbits means the paths disagree on the rendered-center
+// convention (invisible at yaw 0: a (0.5,0.5,0.5) offset rides the iso depth
+// axis and projects to zero until rotated).
+bool g_pivotVerifySdf = false;
+std::vector<IRVideo::AutoScreenshotShot> g_pivotVerifyShots;
+std::vector<std::array<char, 40>> g_pivotVerifyShotLabels;
+// Anchor of the DEFAULT-pivot blocks: an integer world point at iso depth 0
+// (x + y + z == 0) so `isoPixelToPos3D(viewCenterIso, 0)` — the default-pivot
+// focus derivation in getEffectiveCameraIso — recovers it EXACTLY when the pan
+// places it under the viewport center. Integer placement keeps the probe on
+// the voxel lattice (no displaced-pool tie noise in the measurement).
+constexpr vec3 kPivotVerifyDefaultAnchor = vec3(12.0f, -12.0f, 0.0f);
+// Probe z for the DEFAULT-pivot blocks. center-depth's xy displacement off the
+// pinned column equals this value in both axes (iso-ray geometry), so the
+// predicted master-drift scales with it.
+constexpr float kPivotVerifyProbeZ = 10.0f;
+// focus-off: iso offset of the pan away from probe-centered, so the pinned
+// probe holds a NON-center screen position.
+constexpr vec2 kPivotVerifyOffCenterIso = vec2(14.0f, 9.0f);
+
+// Camera pan that places kPivotVerifyDefaultAnchor under the exact viewport
+// center: inverts `viewCenterIso = canvasSize/2 - trixelOriginOffsetZ1 -
+// cameraIso` (getEffectiveCameraIso's default-focus derivation) at
+// `viewCenterIso == pos3DtoPos2DIso(anchor)`.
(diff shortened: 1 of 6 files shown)


jakildev (commented):
## Review — test: pivot-verify — deterministic rotation-pivot invariance harness (epic #2544 tooling)

**Verdict:** approve

### Nits
- `tools/jitter_probe/main.cpp:226-227,285-290` — the new `--stationary`/`--max-deviation` `parser.flag(...)` call and the `--verbose` per-frame `std::printf(...)` call are not clang-format clean (`git clang-format --diff origin/master` flags both, scoped to lines this PR actually added). The PR body's test plan claims `format-changed` was applied — worth double-checking the invocation, since `shape_debug/main.cpp` and `scripts/pivot-verify.py` (ruff-clean) are fine, only this file has drift.

### Praise
- The `--stationary` metric design is careful: `maxDeviationFromFirst` deliberately anchors on the first *valid* sample rather than the mean so a monotone drift doesn't average itself away, and it correctly reuses the pre-existing `validCount < 3` guard rather than re-deriving one.
- Good discipline on the `IndexedSweepShots`-style label-storage lifetime: `g_pivotVerifyShotLabels`/`g_pivotVerifyShots` are `reserve(n)`'d for the exact count pushed, so the non-owning `AutoScreenshotShot::label_` pointers stay valid — matches the documented contract in `auto_screenshot.hpp`.
- The choice of a whole-frame THRESHOLD mask (not `--color`) for the probe is explained and correct: a color-locked mask would track yaw-rotating lit faces and fabricate a false orbit on pinned geometry, which is exactly the kind of false-positive/false-negative a rotation-pivot harness can't afford.
- Cross-checked the epic/child issue numbers cited in the docs (`#2544`–`#2548`) — all exist and match the described fix chain, and the reused math/render symbols (`getMainCanvasSizeTrixels`, `trixelOriginOffsetZ1`, `pos3DtoPos2DIso`) resolve to the cited signatures.

- [ ] `clang-format` (or `fleet-build --target format-changed`) on `tools/jitter_probe/main.cpp` to clear the two format nits above.

jakildev:
Addressed feedback (1 nit, the only item in the review):

- **`tools/jitter_probe/main.cpp:226-227,285-290` — clang-format drift.**
  Verified the citation before applying: `git clang-format --diff origin/master`
  flagged exactly those two hunks (the `--stationary` `parser.flag(...)` call
  and the `--verbose` per-frame `std::printf(...)`), and both are on lines this
  PR added. Applied the scoped fix via `git clang-format origin/master -- tools/jitter_probe/main.cpp`
  rather than a whole-file reformat, so no untouched lines drifted.

Verification:
- `git clang-format --diff origin/master` across the **whole** changed set is
  now clean (it had flagged only this one file, matching the review's scoping —
  `shape_debug/main.cpp` and `scripts/pivot-verify.py` were already fine).
- The file is whitespace-normalized identical to its prior revision, so the
  commit carries zero semantic change.
- `fleet-build --target jitter_probe` green on macOS; ran the binary and
  confirmed `--stationary` / `--max-deviation` / `--verbose` still register in
  the parser usage output.

`fleet:approved` left in place — a nit cleanup does not invalidate the approval.
Note the push may trip the post-approval re-review bot; that is expected and the
reviewer restores the label.

— worker (pool-2, opus, macOS)

jakildev:
Root-cause note on that nit, since the review's inference was reasonable but the
mechanism is at fault rather than the invocation:

> The PR body's test plan claims `format-changed` was applied — worth double-checking the invocation

`format-changed` was almost certainly invoked correctly and **silently did
nothing to this file**. `cmake/ir_quality_tools.cmake` builds the quality file
list from three roots only — `engine/`, `creations/`, `test/` — so the top-level
`tools/` tree is never globbed. Verified against the generated
`build/irreden_quality_files.cmake` (774 files): the only `/tools/` entry is
`engine/tools/bench/ir_ref_bench.cpp`; both `tools/img_diff/main.cpp` and
`tools/jitter_probe/main.cpp` are absent. All three targets (`format`,
`format-check`, `format-changed`) share that list, so all three are blind here.

`git clang-format --diff origin/master` is git-diff-scoped rather than
file-list-scoped, which is exactly why the review caught what the target missed.

Filed as #2556 (`fleet:coding-improvement`) — the fix is one line adding `tools`
to `search_roots`. Out of scope for this PR; noting it here so the "author
mis-invoked the formatter" reading doesn't propagate.

— worker (pool-2, opus, macOS)

jakildev (commented):
## Review — test: pivot-verify — deterministic rotation-pivot invariance harness (epic #2544 tooling)

**Verdict:** approve

### Prior-review resolution
- ✅ `tools/jitter_probe/main.cpp:226-227,285-290` — clang-format drift on the `--stationary` `parser.flag(...)` call and the `--verbose` per-frame `std::printf(...)` — verified fixed at `fbc72702`. `git clang-format --diff origin/master -- tools/jitter_probe/main.cpp` is clean, and the commit's diff is confined to exactly those two hunks with no semantic change (whitespace-only reflow).

Only one commit landed since the prior review (`fbc72702`, 18:40:45Z); the two commits carrying the actual harness logic (`99367721`, `1a374349`) both predate the review timestamp (18:34:56Z) and were already covered by the prior pass. No new logic to re-check.

### Praise
- Reformat is properly scoped — the author re-ran the reviewer's exact citation (`git clang-format --diff origin/master`) before touching anything, rather than a whole-file reformat that could have picked up unrelated drift.

- [ ] Linux/GL twin run of the baseline (cross-host smoke) — already flagged in the PR body as outstanding, not new.

PR #2558, opened 2026-07-26
editor: F-1.6 Part 2d — author-entity runner + ROCK entity (#766)

jakildev:
## Summary

Part 2d of #766 — **does not close it** (the plan's four-PR sequence; next slice 2e = mushroom + the F-1.2 symmetry fix).

- **`scripts/author-entity.py`** — the session runner. Replays a `--gui-session <name>` recipe through the GUI-test harness (the same `run_capture` flow `gui-verify` uses), parses `GUI-ASSERT` (fails on any FAIL / zero asserts), confirms the editor saved its `scene_frame_*.vxs`, **re-runs and byte-compares the saved set** (the acceptance's determinism gate), and copies the frames to `assets/voxel/entities/<entity>[_frame_N].vxs(+.json)`.
- **ROCK session** (`sessions.hpp`) — authors an irregular, no-symmetry, single-layer blob entirely by scripting the real UI: clears the seeded ground slab down to a 5×5 central footprint (four erase-mode box drags on the flat plane, run first — before any rock voxel occludes the corner aims), builds three asymmetric layers, carves two base corners. Registered on the `--gui-session` enum.
- **Shared parser** — the `GUI-ASSERT` regex + pass/fail table + evaluate flow moved to `verify_common`; `gui-verify` and `author-entity` now share one implementation instead of two copies.
- **`assets/voxel/entities/rock.vxs(+.json)`** — the first committed entity.
- Friction log: 2d findings + the two 2c open questions resolved.

## Verification (macOS/Metal)

- `python3 scripts/author-entity.py rock` → **18/18 GUI-ASSERT PASS** on both runs, **determinism gate byte-identical**, asset committed.
- `rock.vxs` parses to **65 occupied voxels** (alpha>0) in a compact `x[6..10] y[6..10] z[11..15]` bounds — footprint 25 + base 23 + mid 12 + cap 4 + peak 1. Not a slab: the other 4031 dense slots are `alpha==0` (erased) and skipped by the renderer.
- `gui-verify.py IRVoxelEditor -- --gui-session drag_probe` still **12/12** (shared-parser refactor intact).

Renders in `IRShapeDebug --load-vxs` (the small dark-red blob; it's a 5³ entity beside ShapeDebug's larger procedural shapes):

rock closeup
rock in the shape_debug scene

## Test plan

- [x] `author-entity.py rock` → 18/18, determinism byte-identical (macOS/Metal)
- [x] `rock.vxs` occupancy = 65 voxels, 5×5×5 bounds (clean rock, not a slab)
- [x] `gui-verify.py IRVoxelEditor -- --gui-session drag_probe` → 12/12 (refactor regression check)
- [x] `IRVoxelEditor` + `IRShapeDebug` build clean; `ruff check scripts/` clean
- [ ] Linux-GL parity run — nice-to-have (harness is CPU-state, backend-agnostic), not a gate

## Notes for reviewer

- **Not a render-pipeline change.** The GUI-test assertion path is CPU-state and backend-agnostic (plan gotcha), so this authors + verifies on macOS; a Linux-GL smoke is a nice-to-have, not a gate. No render files touched.
- **`optimize` skipped deliberately** — the recipe builds once at startup and the runner is a test tool; nothing on a shipping path.
- **`main.cpp` change is 2 lines** (the `--gui-session` enum gains `rock`). simplify flagged two *pre-existing* `getComponent`-in-loop sites (main.cpp:792, 1622) that this PR does not touch — left for a separate pass.


jakildev (commented):
## Review — editor: F-1.6 Part 2d — author-entity runner + ROCK entity (#766)

**Verdict:** approve

### Nits
- `scripts/author-entity.py:166-167` — the comment says "Restore run-1 snapshots into memory-backed temp paths for the compare", but the code doesn't create any temp paths — it just diffs `frames2`'s live-read bytes against the in-memory `first_bytes` dict built earlier. Reword to match what actually happens (or drop the sentence).
- `creations/editors/voxel_editor/sessions.hpp:193-207` — the carve-arm pattern (hover + `expectPick` before the destructive click, the F-2d-2 rule this same PR documents) is applied to `carveA` but not `carveB`. `carveB`'s click has no pre-arm pick-assert, so a bad aim on that corner is only diagnosable indirectly via the post-click `expectOccupancy` failing, not via a pick mismatch. Consider arming both corners for consistency with the stated rule.
- `scripts/verify_common.py:122-124` (`report_gui_asserts`) — the hung message dropped the specific `--timeout <N>s` value that `gui-verify.py`'s original inline message included ("...killed it before the shot table completed" vs. the old "...at --timeout {timeout}s..."). Minor loss of diagnostic detail in the shared version since the caller's timeout isn't threaded through.
- `creations/editors/voxel_editor/sessions.hpp:107` — the comment claims the recipe needs a scene "~12 tall", but the z math only requires `gz - 4 >= 0` (i.e. `sceneSize.z >= 5`). Worth double-checking where the 12 comes from (camera framing, not the recipe's own reach?) or tightening the comment.

### Praise
- The geometry math checks out end-to-end against the code, not just the PR body's claims: footprint 25 + base 23 (25 − 2 carved corners) + mid 12 + cap 4 + peak 1 = 65 occupied voxels, matching both the reported count and the claimed `x[6..10] y[6..10] z[11..15]` bounds for the default 16³ scene (`cx=cy=8`, `gz=15`).
- `author-entity.py`'s determinism gate — snapshot run-1 bytes into memory before run-2 clears the save dir, then compare against run-2's live files — is a clean, race-free way to prove replay determinism without needing a second on-disk copy.

- [x] Already covered by the PR's own verification (18/18 GUI-ASSERT, determinism byte-identical, `gui-verify.py drag_probe` regression 12/12) — nothing further needed before merge.

Opus recheck not required.

jakildev:
Addressed feedback:
- Reworded the determinism-gate comment in `author-entity.py` (no temp paths are created; it's a direct live-vs-snapshot byte compare).
- Armed `carveB` with its own hover/pick-assert segment (`carve_arm_b`), symmetric with `carveA`. Note: this required its own segment rather than inline checks in the click shot — segments are shot boundaries, not per-event checkpoints, so a hover+expectPick issued in the same shot as the click evaluates against the post-click state. Caught this the hard way: my first attempt (checking inline) failed `carve_aim_hits_corner_b`/`carve_target_present_b` because the corner was already carved by the time the shot's assertions ran. Fixed by giving `carveB` its own arm segment, mirroring `carveA`.
- Threaded `--timeout` through `report_gui_asserts` so the hung-run message reports the specific watchdog value again.
- Corrected the rock recipe's z-depth comment: the recipe needs 5 deep in z (peak sits at `gz - 4`), not ~12 — the ~12 figure wasn't derived from the recipe's own math.

Verified: `gui-verify.py drag_probe` 12/12 PASS (regression check), `author-entity.py rock` 20/20 PASS + determinism gate byte-identical, build clean. — worker

jakildev (commented):
## Review — editor: F-1.6 Part 2d — author-entity runner + ROCK entity (#766)

**Verdict:** approve

### Prior-review resolution
- ✅ `scripts/author-entity.py:166-167` — stale "restore into memory-backed temp paths" comment — verified fixed at `6e60245a`: now reads "Compare run 2's live files against run 1's captured bytes," which matches what the code actually does.
- ✅ `creations/editors/voxel_editor/sessions.hpp:193-207` — `carveB` missing a pre-arm pick-assert — verified fixed at `6e60245a`: `carveB` now gets its own `carve_arm_b` segment (hover + `expectPick(carveB, "carve_aim_hits_corner_b")` + `expectOccupancy(carveB, true, "carve_target_present_b")`) before the `carve_b` segment's click, mirroring `carveA`. The two new asserts line up with the author's "18/18 → 20/20" count. The commit message also explains why this needed its own segment rather than an inline check (segments are shot boundaries, not per-event checkpoints) — matches the builder's semantics as I read them elsewhere in this file.
- ✅ `scripts/verify_common.py:122-124` (`report_gui_asserts`) — hung message dropped the `--timeout <N>s` value — verified fixed at `6e60245a`: `report_gui_asserts` now takes an optional `timeout` param and both callers (`author-entity.py`, `gui-verify.py`) thread `args.timeout` through; confirmed `--timeout` is a real argparse arg on both (`gui-verify.py:37`, `author-entity.py:131`).
- ✅ `creations/editors/voxel_editor/sessions.hpp:107` — inaccurate "~12 tall" comment — verified fixed at `6e60245a`: now says "5 deep in z (the peak voxel sits at `gz - 4`)". Checked against the code: `gz = sceneSize.z - 1`, and the recipe uses z-levels `gz` through `gz - 4` (ground, base, mid, cap, peak) — five distinct levels, so `sceneSize.z >= 5` is exactly right.

### New commit reviewed
`6e60245a` ("fixup: address review nits on PR #2558") — the only commit since the prior review's `submittedAt`. All four hunks are the targeted nit fixes above; no unrelated changes rode along. Segment names in the new `carve_arm_b`/`carve_b` pair don't collide with any other segment name in `buildRock` or the file's other recipes.

No new issues found in the fixup itself.

- [x] Already covered by the author's own re-verification (20/20 GUI-ASSERT, determinism byte-identical, `gui-verify.py drag_probe` 12/12) — nothing further needed before merge.

Opus recheck not required.

PR #2559, opened 2026-07-26
audio: MIDI client init failure degrades to no-MIDI instead of aborting (#2549)

### LINKED ISSUE #2549: audio: MIDI client init failure aborts the engine — degrade to no-MIDI per the no-device doctrine
**Area:** engine/audio
**Model:** sonnet
**Blocked by:** (none)

## Symptom

Any engine executable aborts at startup when the CoreMIDI client cannot be created (observed: sandboxed shells where the Mach lookup to `com.apple.midiserver` is denied):

```
MidiInCore::initialize: error creating OS-X MIDI client object (-304).
libc++abi: terminating due to unexpected exception of type rt::midi::RtMidiError
ir-run: RESULT=CRASH exe=IRShapeDebug exit=134 signal=SIGABRT
```

`MidiIn` holds `RtMidiIn m_rtMidiIn` by value (`engine/audio/src/midi_in.cpp:15-27`); its default construction throws `RtMidiError` before `World` finishes constructing, uncaught → `std::terminate`. `MidiOut` has the same shape. The audio module's own doctrine is "**No device = silent, never a crash**" (`engine/audio/CLAUDE.md`, AudioPlayback already implements it) — MIDI should degrade the same way. Headless Linux CI boxes without an ALSA sequencer hit the same class.

## Acceptance criteria

1. `MidiIn`/`MidiOut` construction failure degrades to a no-MIDI state: port count 0, `openPort` warns and returns -1, `tick()`/queries no-op, `sendMidiMessage` no-ops. No throw escapes the constructors.
2. An engine demo runs to RESULT=CLEAN in an environment with no MIDI server access (repro: a sandboxed shell denying the midiserver Mach lookup).
3. Normal MIDI hosts unaffected (ports enumerate and open as today).
4. One log line at degrade time naming the RtMidi error, so the state is diagnosable.

## Context

Found 2026-07-22 while driving IRShapeDebug headlessly for the #2544 pivot harness. Likely mechanism: make the RtMidi handles `std::unique_ptr` constructed inside try/catch (mirrors AudioPlayback's failed-`ma_engine_init` pattern), guarding the member uses. Check `patches/` for local RtMidi patches before assuming stock behavior.

jakildev:
## Summary
- `MidiIn`/`MidiOut` construction (and `openPort()`) now catches `RtMidiError`
  and degrades to a no-MIDI state (0 ports, warn-once, no-op sends) instead of
  letting the throw escape and abort the process — mirrors `AudioPlayback`'s
  existing "no device = silent, never a crash" doctrine.
- Patched the vendored RtMidi CoreMIDI backend (`ir_rtmidi.patch`, wired via
  the same `FetchContent` + `IrredenEngine_applyPatchOnce` mechanism as the
  existing `ir_rtaudio.patch`): `getCoreMidiClientSingleton()` was declared
  `throw()`, which C++17+ treats as `noexcept`, but it calls RtMidi's internal
  `error()` helper which throws on client-creation failure — violating the
  noexcept contract calls `std::terminate()` before any try/catch in the
  engine ever sees the exception. Dropping the stale spec was required for
  the engine-side guard above to actually work on macOS.
- Documented the degrade behavior in `engine/audio/CLAUDE.md`'s MIDI model
  section.

## Test plan
- [x] Normal run (`IRShapeDebug --auto-screenshot 10`) — RESULT=CLEAN, MIDI
  enumeration log lines unchanged (healthy-host behavior unaffected).
- [x] Sandboxed repro (`sandbox-exec` denying the `com.apple.midiserver` Mach
  lookup) — before the fix: `SIGABRT` / `libc++abi: terminating due to
  unexpected exception of type rt::midi::RtMidiError`; after: exit 0, one
  `IRE_LOG_WARN` naming the RtMidi error for both `MidiIn` and `MidiOut`.
- [x] Verified the CMake patch mechanism end-to-end by reverting the vendored
  `RtMidi.cpp` to pristine and reconfiguring — `IrredenEngine_applyPatchOnce`
  applied `ir_rtmidi.patch` cleanly and the rebuilt `librtmidi.dylib` produced
  the same degrade-not-crash result.

## Acceptance evidence
|---|---|---|
| No throw escapes construction — degraded state is 0 ports / warn-once / no-op | sandboxed `IRShapeDebug --auto-screenshot 10` | `MidiIn: RtMidi client init failed (...); MIDI input disabled` / same for `MidiOut`, exit 0 |
| Demo reaches RESULT=CLEAN with no MIDI server access | `sandbox-exec -f deny-midiserver.sb ./IRShapeDebug --auto-screenshot 10` | exit=0, no `SIGABRT`/`libc++abi` termination |
| Normal MIDI hosts unaffected | `fleet-run IRShapeDebug --auto-screenshot 10` (unsandboxed) | `ir-run: RESULT=CLEAN exe=IRShapeDebug exit=0`, MIDI enumeration log lines present |
| One log line names the RtMidi error at degrade time | sandboxed run | single `IRE_LOG_WARN` per class citing `RtMidiError::getMessage()` |

## Notes for reviewer
The RtMidi `throw()`/noexcept issue is easy to miss if you only check the
engine-side try/catch shape — it looked correct in isolation but the crash
reproduced identically until the vendor patch went in. If you want to
re-verify: `sandbox-exec -f <(printf '(version 1)\n(allow default)\n(deny mach-lookup (global-name "com.apple.midiserver"))\n') ./build/creations/demos/shape_debug/IRShapeDebug --auto-screenshot 10`.

Closes #2549

### DIFF
--- a/.fleet/plans/issue-2549.md
+++ b/.fleet/plans/issue-2549.md
@@ -1,0 +1,66 @@
+## Plan: MIDI client init failure degrades to no-MIDI
+
+- **Issue:** #2549
+- **Model:** sonnet
+- **Date:** 2026-07-24
+
+### Verified current state (confirmed repro)
+
+`MidiIn::MidiIn()` (`engine/audio/src/midi_in.cpp:15-27`) and `MidiOut::MidiOut()`
+(`engine/audio/src/midi_out.cpp:5-17`) each hold a **by-value** RtMidi
+enumeration probe (`RtMidiIn m_rtMidiIn` at `midi_in.hpp:80`, `RtMidiOut
+m_rtMidiOut` at `midi_out.hpp:46`, both commented "enumeration probe only").
+`AudioManager` holds `MidiIn`/`MidiOut` by value (`audio_manager.hpp:57-60`),
+constructed with `World`. When the CoreMIDI client can't be created (repro: a
+sandboxed shell denying the `com.apple.midiserver` Mach lookup —
+`IRShapeDebug` aborts with `MidiInCore::initialize: error creating OS-X MIDI
+client object (-304)` → uncaught `RtMidiError` → SIGABRT, RESULT=CRASH), the
+throw escapes the member initializer list before any catch could run. The
+per-port handles are already `unique_ptr` (`MidiInPort::rtMidiIn_`,
+`MidiOutPort::rtMidiOut_`), constructed in `openPort` (`midi_in.cpp:75-80`,
+`midi_out.cpp` equivalent) — those construction/open calls can throw the
+same way on a degraded host. `AudioPlayback` already implements the module's
+"no device = silent, never a crash" doctrine (engine/audio/CLAUDE.md); MIDI
+is the gap.
+
+### Affected files
+
+- `engine/audio/include/irreden/audio/midi_in.hpp` — `m_rtMidiIn` becomes
+  `std::unique_ptr<RtMidiIn>`.
+- `engine/audio/src/midi_in.cpp` — ctor try/catch around probe construction
+  + enumeration (degraded ⇒ `m_numberPorts = 0`, one `IRE_LOG_WARN` naming
+  `RtMidiError::getMessage()`); `openPort(substring)` wraps the
+  `make_unique<RtMidiIn>()` + `openPort(i)` + `setCallback` triple in
+  try/catch (warn + return -1).
+- `engine/audio/include/irreden/audio/midi_out.hpp` /
+  `src/midi_out.cpp` — same treatment for `m_rtMidiOut` and
+  `MidiOut::openPort`; `sendAllNotesOff`/`sendMessage` already no-op on
+  empty `m_ports`.
+- Check `patches/` for local RtMidi patches before assuming stock throw
+  behavior (engine/audio/CLAUDE.md gotcha).
+
+### Approach
+
+Mirror the AudioPlayback failed-init pattern: construct the enumeration
+probes inside try/catch in the ctor body (member becomes a null
+`unique_ptr` on failure), keep `m_numberPorts`/`m_portNames` empty in the
+degraded state, and guard every `m_rtMidi*` dereference (`getPortCount`,
+`getPortName`) on the pointer. All query/tick paths already tolerate zero
+ports (`m_ports` empty ⇒ loops no-op), so no consumer changes. No behavior
+change on healthy hosts: same enumeration, same logs, same port handles.
+
+### Acceptance criteria
+
+As on the issue: (1) no throw escapes construction — degraded state is 0
+ports / warn-once / no-op sends; (2) an engine demo reaches RESULT=CLEAN in
+a MIDI-denied environment (repro above); (3) healthy-host behavior
+unchanged (ports enumerate + open, MIDI demos work); (4) one WARN log names
+the RtMidi error.
+
+### Verification
+
+`fleet-build --target IRShapeDebug`; run `IRShapeDebug --auto-screenshot 10`
+in a sandboxed (MIDI-denied) shell → RESULT=CLEAN; run once normally and
+confirm the MIDI enumeration log lines still appear; build + run one
+MIDI-using demo (e.g. the MIDI macro creation) on a healthy host for the
+no-regression check.
--- a/engine/audio/CLAUDE.md
+++ b/engine/audio/CLAUDE.md
@@ -107,6 +107,12 @@ Lua (`scripts/audio_demo.lua`).
   fallback patterns for common devices (UMC1820, Focusrite, MPKmini2,
   OP-1). Edit `midi_in.cpp` / `midi_out.cpp` to add a new hardware match.
   Opening an already-open port is a no-op that returns the existing handle.
+- **No device = silent, never a crash** (same doctrine as `AudioPlayback`).
+  If the RtMidi client can't be created (no MIDI server access), `MidiIn`/
+  `MidiOut` degrade to a no-MIDI state instead of throwing out of the
+  constructor: 0 ports, `openPort` warns and returns -1, `tick()`/queries/
+  `sendMidiMessage` no-op. One `IRE_LOG_WARN` names the RtMidi error at
+  degrade time.
 
 ## Outbound MIDI observer (#1869)
 
--- a/engine/audio/CMakeLists.txt
+++ b/engine/audio/CMakeLists.txt
@@ -68,6 +68,19 @@ FetchContent_Declare(
 # rtmidi's uninstall target no longer clashes with rtaudio's: upstream renamed
 # RTMIDI_TARGETNAME_UNINSTALL to "rt_uninstall", and ir_rtaudio.patch renames
 # rtaudio's to "RtAudioUninstall".
+# ir_rtmidi.patch: MidiInCore/MidiOutCore::getCoreMidiClientSingleton is
+# declared `throw()`, which C++17+ treats as `noexcept` — but it calls
+# error() with a non-WARNING type, which throws RtMidiError when no error
+# callback is registered. Under C++23 that violates the noexcept contract
+# and calls std::terminate() before the exception ever reaches a caller's
+# try/catch, no matter how the client code is written. Dropping the
+# obsolete `throw()` spec lets the CoreMIDI client-init failure propagate
+# normally so MidiIn/MidiOut can degrade to no-MIDI instead of crashing.
+IrredenEngine_applyPatchOnce(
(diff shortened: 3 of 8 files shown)


jakildev (commented):
## Review — audio: MIDI client init failure degrades to no-MIDI instead of aborting (#2549)

**Verdict:** approve

### What this does

`MidiIn`/`MidiOut` promote their enumeration-probe `RtMidiIn`/`RtMidiOut`
members from by-value to `unique_ptr`, construct them inside a `try`/`catch`
in the ctor body (degraded ⇒ null pointer, 0 ports, one `IRE_LOG_WARN`), and
give `openPort` the same guard around its per-port handle construction.
Also vendors `ir_rtmidi.patch` to drop a stale `throw()` spec on
`MidiInCore`/`MidiOutCore::getCoreMidiClientSingleton` — under C++17+ that
spec is treated as `noexcept`, so the internal `RtMidiError` throw on
CoreMIDI client-creation failure called `std::terminate()` before any
engine-side `catch` ever ran; without the patch the C++-side try/catch alone
would not have fixed the crash on macOS.

### Verification performed

- Read `midi_in.cpp`/`midi_out.cpp`/`midi_in.hpp`/`midi_out.hpp` in full (not
  just the diff hunks). `m_rtMidiIn`/`m_rtMidiOut` are only touched inside
  their own ctor's try/catch — no other method dereferences them, so the
  null-after-degrade state can't be use-after-null'd elsewhere.
- `openPort` loops `for (i = 0; i < m_numberPorts; i++)` — in the degraded
  state `m_numberPorts == 0`, so the loop body (which constructs a *new*,
  separate `RtMidiIn`/`RtMidiOut` for the per-port handle) never runs. No
  path reaches a null-deref; it correctly falls through to the "no port
  matching" warn-and-return-`-1` branch.
- `MidiOut::sendAllNotesOff()` (called from `~MidiOut()`) guards on
  `m_ports.empty()` before dereferencing `port->rtMidiOut_`; `m_ports` only
  ever gains entries past the `openPort` try block's success path, so every
  stored port has a valid handle.
- Member-initializer-list order in both ctors matches declaration order in
  the headers (`m_rtMidiIn/m_rtMidiOut, m_numberPorts, m_portNames, m_ports,
  [m_frameBuffer]`) — no `-Wreorder` risk.
- `ir_rtmidi.patch` follows the exact same `IrredenEngine_applyPatchOnce`
  mechanism as the pre-existing `ir_rtaudio.patch` (`engine/audio/
  CMakeLists.txt:75-82` vs. the established `rtaudio` block above it) — not
  a new pattern, just the second user of an existing one.
- `engine/audio/CLAUDE.md`'s "No device = silent, never a crash" doctrine
  section gains a matching MIDI paragraph, consistent with the existing
  `AudioPlayback` entry it mirrors.

### Acceptance (issue #2549)

Issue #2549 states its acceptance criteria as a flat `## Acceptance
criteria` list (not nested under a `## Plan` heading), so this didn't
trigger the formal acceptance-grader dispatch — checked by hand instead,
cross-referencing the PR's own evidence table against the code:

| Criterion | Status |
|---|---|
| 1. No throw escapes construction; degraded ⇒ 0 ports/warn-once/no-op | Met — ctor try/catch + `openPort` try/catch confirmed in source |
| 2. Demo reaches RESULT=CLEAN with no MIDI server access | Met — PR body's sandboxed `sandbox-exec` repro, author-reported exit 0 |
| 3. Normal MIDI hosts unaffected | Met — enumeration/log-line code path unchanged on the non-throw branch |
| 4. One log line names the RtMidi error at degrade time | Met — one `IRE_LOG_WARN` citing `error.getMessage()` per class |

All four met from the diff + PR-reported evidence; nothing unverifiable.

### Nits

None — the change is small, mechanical, and directly mirrors an existing
in-repo pattern (`AudioPlayback`'s failed-`ma_engine_init` degrade).

### Praise

- Correctly diagnosed that the C++-side try/catch alone was insufficient on
  macOS and traced it to the vendored `throw()`/`noexcept` interaction
  rather than stopping at "the catch should have worked" — the PR body's
  "Notes for reviewer" section flags this as the easy-to-miss part, and it's
  right to flag it.
- Reused the established `IrredenEngine_applyPatchOnce` + `patches/`
  mechanism instead of hand-patching the vendored tree or inventing a new
  patch-application path.

- [x] Already covered by the PR's own verification (build clean, sandboxed
  repro before/after, normal-host regression check, patch-mechanism
  end-to-end reapply test) — nothing further needed before merge.

Escalation: please Opus-recheck before merge (touches: engine/audio/,
including a vendored third-party source patch and ctor exception-safety —
in scope per the reviewer protocol's Opus-required module list even though
this pass found nothing).

Opus recheck required: engine/audio/ ctor exception-safety change + vendored RtMidi source patch.

jakildev (commented):
## Opus recheck — audio: MIDI client init failure degrades to no-MIDI (#2549)

**Verdict:** approve

Sonnet escalated per the reviewer-protocol Opus-required module list
(`engine/audio/` + a vendored third-party source patch + ctor
exception-safety) even though its pass found nothing. I re-ran the pass
focused on what Sonnet couldn't fully close: exception-safety across the
degraded-null lifetime, the `throw()`→`noexcept` patch semantics, and the
cross-platform behavior of the vendored patch. All confirmed sound.

### Confirmed (the Opus-only items)

- **Probe-member null lifetime.** Sonnet stated `m_rtMidiIn`/`m_rtMidiOut`
  are only touched inside their own ctor try/catch. On a full-tree grep
  (`grep -rn 'm_rtMidiIn\|m_rtMidiOut' engine/`) I confirm this holds: the
  only dereferences are the ctor probe-enumeration (now inside the `try`) —
  every other path uses the *separate* per-port `unique_ptr` handles
  constructed in `openPort`. The degraded state (`nullptr`,
  `m_numberPorts == 0`) is therefore unreachable by any deref, so no
  use-after-null exists downstream.

- **`openPort` structure survives the try wrap.** Read the full PR-branch
  body: substring filter → `continue`; already-open check → `return i`;
  then the `try` wraps only handle construction/open/setCallback, with the
  catch returning `-1`. In the degraded state the outer `for` never enters
  (`m_numberPorts == 0`) and falls through to the existing warn+`-1`.
  `-1` is already the function's "no matching port" sentinel, so callers
  see no new contract — the catch reuses an existing failure code.

- **Catch scope is right.** `catch (const RtMidiError&)` is the correct
  degrade signal; `std::bad_alloc` from `make_unique` is intentionally
  *not* caught (OOM stays fatal), which is the right call.

- **The `throw()` patch reasoning is correct.** `getCoreMidiClientSingleton`
  is declared `throw()`, treated as `noexcept(true)` under C++17+; the body
  calls `error()` with a non-WARNING type which throws `RtMidiError`, so the
  throw hit `std::terminate()` before any engine-side `catch` could run —
  the C++-side try/catch alone genuinely would not have fixed the macOS
  crash. Dropping the obsolete spec lets it propagate normally. The patched
  lines are CoreMIDI (`#ifdef __MACOSX_CORE__`) code, so the patch is a
  text-level no-op on Linux/Windows and behaviorally macOS-only — no
  cross-platform risk.

- **Header lifetime.** `unique_ptr<RtMidiIn>` needs the complete type at
  `MidiIn`'s dtor; the header already included the full RtMidi type for the
  prior by-value member and the dtor lives in the `.cpp`, so no
  incomplete-type hazard. `<memory>` is already in scope (existing
  `unique_ptr<MidiInPort>` members).

### FYI — not a blocker for this PR

`IrredenEngine_applyPatchOnce` (`cmake/ir_functions.cmake:1-30`) prints
`"<name> patch failed to apply, SHOULD ABORT HERE."` on a non-zero
`git apply` but does **not** `FATAL_ERROR` — and it writes the
`_patch_applied.txt` marker unconditionally. So if a future RtMidi
FetchContent bump drifts the patch context, the patch silently no-ops,
the marker suppresses any retry, and the macOS `std::terminate` crash
quietly returns. This is **pre-existing shared infrastructure** — the same
helper already backs `ir_rtaudio.patch`, and this PR is only its second
correct user, so it is out of scope here. Flagging for the human: worth a
separate ticket to make the helper abort (and skip the marker) on
patch-apply failure, which would harden both patch consumers at once.

### On Sonnet's pass

Sonnet's review was accurate and complete on everything it claimed —
member-init order, the acceptance-criteria mapping, the `sendAllNotesOff`
empty-`m_ports` guard, and the diagnosis of the `throw()`/`noexcept`
interaction. Nothing to correct; I only added the full-tree deref
confirmation and the shared-helper durability note above.

🤖 Opus recheck by Claude Opus (role-opus-reviewer)

PR #2573, opened 2026-07-28
input: release entity-event handler Lua refs before World teardown

### LINKED ISSUE #2572: input: EntityEventHandlers process-static unrefs Lua handlers after World teardown — exit SIGSEGV for any creation using IRInput.onRightClick (#2539 fallout)
**Area:** engine/prefabs (input) / engine lifecycle
**Model:** opus
**Blocked by:** (none)

## Context

Since #2539 (issue #2528) moved `World` teardown to the tail of
`IREngine::gameLoop()`, the Lua VM (`sol::state`, owned by the World's script
manager) dies long before process exit. But
`IRSystem::getEntityEventHandlers()` (`engine/prefabs/irreden/input/systems/system_entity_hover_detect.hpp`)
is a Meyers-singleton **process static** whose `HandlerEntry` vectors hold
`sol::protected_function` references. Its destructor now runs at
`__cxa_finalize` — after `lua_close` — and unrefs into freed memory.

**Every creation that registers an entity-event handler from Lua
(`IRInput.onRightClick` et al.) now segfaults at process exit**, failing the
clean-exit policy on an otherwise green run.

Forensics (macOS/Metal, 2026-07-27, observed in a downstream creation; the
mechanism is creation-agnostic):

- Crash stack (from the `.ips` report):
  `lua_rawgeti` ← `luaL_unref` ← `IRSystem::EntityEventHandlers::~EntityEventHandlers()` ← `__cxa_finalize_ranges` ← `exit`.
- A/B: identical binary, runtime script with the single
  `IRInput.onRightClick(...)` registration removed → `RESULT=CLEAN exit=0`;
  with it → `SIGSEGV` after "Clean shutdown complete."
- The #2528 exit-order audit covered `CPUProfiler`, `IrredenGLAPI`, and
  `g_luaBindingRegistrations` but missed this registry.
- Rule angle: `.claude/rules/cpp-globals.md` — this is world-scoped state
  (Lua refs) living in a process-lifetime static; the storage outlives its
  owner by construction.

## Acceptance criteria

- A creation that registers `IRInput.onRightClick` from Lua exits
  `RESULT=CLEAN` (exit 0) under `ir-run` on macOS and Linux.
- Handler dispatch during the run is unchanged (hover/click/right-click still
  fire).
- The registry holds no live `sol::protected_function` after `gameLoop()`
  returns (so the static's dtor destroys only empty vectors).

jakildev:
## Summary
- Fixes an exit-time SIGSEGV (#2572). The process-lifetime `EntityEventHandlers`
  registry (hover / click / right-click callbacks) holds `sol::protected_function`
  refs into the World's Lua VM. Since #2539 moved World teardown to the
  `gameLoop()` tail, `lua_close` now runs before the static's destructor — which
  fires at `__cxa_finalize` and `luaL_unref`s into freed memory. Any creation
  that registered a handler from Lua segfaults at process exit.
- Adds `EntityEventHandlers::clear()` and calls it via an out-of-line
  `detail::clearEntityEventHandlers()` at the `gameLoop()` tail, immediately
  **before** `g_world.reset()`, while the VM is still alive — so the static's own
  destructor unrefs only empty vectors (acceptance criterion #3, empty registry
  post-`gameLoop`, holds by construction).
- The helper is out-of-line in `engine.cpp` (mirroring `applyPreInitLuaConfig`)
  so the prefab header's sol2 + `ir_render` includes don't transitively widen
  every includer of `ir_engine.hpp`. Runtime dispatch is unchanged; `nextId` is
  left as-is (ids never recycle within a process).

## Test plan
- [x] `IRShapeDebug` builds + runs clean on macOS/Metal — `RESULT=CLEAN exit=0`.
- [x] A/B with a handler registered against the World's Lua VM (the exact
  scenario a `IRInput.onRightClick` binding produces): **without** the clear →
  `SIGSEGV` (exit 139) after "Clean shutdown complete." — reproduces the issue's
  forensics; **with** the clear → `RESULT=CLEAN exit=0`. Dispatch still fires
  during the run.
- [ ] Linux/GL exit smoke — the fix is a host-agnostic teardown-ordering change
  (no render/backend code); reviewer to confirm on a GL host per acceptance.

Closes #2572

### DIFF
--- a/.fleet/plans/issue-2572.md
+++ b/.fleet/plans/issue-2572.md
@@ -15,9 +15,14 @@ empty vectors. No change to the runtime dispatch surface.
 2. Call it at the `IREngine::gameLoop()` tail, immediately **before** the
    `g_world.reset()` that #2539 added — the VM is still alive there, and the
    engine tail is unconditional, so a creation cannot forget the cleanup.
-3. This also fixes the latent multi-world hazard: without the clear, a
-   future second `World` in-process would dispatch stale handlers bound to
-   the dead VM.
+3. This also closes the stale-handler half of the latent multi-world
+   hazard: without the clear, a future second `World` in-process would
+   dispatch stale handlers bound to the dead VM; with it, the registry
+   starts empty. Multi-world is NOT fully clean after this fix —
+   `previousHoveredEntity` (a second process-lifetime static inside
+   `System<ENTITY_HOVER_DETECT>::create()`) survives into a hypothetical
+   second World, whose first hover change would fire `fireUnhovered` with
+   the previous World's stale id. Pre-existing and unreachable today.
 
 Implementation note: `IREngine::gameLoop()` is header-inline in
 `ir_engine.hpp`, and the handler registry lives behind a prefab header that
--- a/.fleet/plans/issue-2572.md
+++ b/.fleet/plans/issue-2572.md
@@ -1,0 +1,63 @@
+# Plan — #2572: EntityEventHandlers unrefs Lua handlers after World teardown (exit SIGSEGV)
+
+## Scope
+
+Make entity-event handler teardown deterministic: the Lua handler references
+in `IRSystem::EntityEventHandlers` must be released while the Lua VM is still
+alive, so the process-exit destruction of the static registry touches only
+empty vectors. No change to the runtime dispatch surface.
+
+## Approach
+
+1. Add `EntityEventHandlers::clear()` — drops all four `HandlerEntry`
+   vectors (destroying the `sol::protected_function`s) and leaves `nextId`
+   as-is (ids never recycle within a process; cheap and unambiguous).
+2. Call it at the `IREngine::gameLoop()` tail, immediately **before** the
+   `g_world.reset()` that #2539 added — the VM is still alive there, and the
+   engine tail is unconditional, so a creation cannot forget the cleanup.
+3. This also fixes the latent multi-world hazard: without the clear, a
+   future second `World` in-process would dispatch stale handlers bound to
+   the dead VM.
+
+Implementation note: `IREngine::gameLoop()` is header-inline in
+`ir_engine.hpp`, and the handler registry lives behind a prefab header that
+pulls in sol2 + `ir_render`. Including that header into `ir_engine.hpp` would
+transitively widen every creation's include of the engine entry point, so the
+call is routed through an out-of-line `IREngine::detail::clearEntityEventHandlers()`
+defined in `engine/engine.cpp` — the same pattern the pre-existing
+`detail::applyPreInitLuaConfig` uses for exactly this reason. (The plan's
+original "call in `ir_engine.cpp`" wording predated confirming the tail is
+header-inline.)
+
+Alternative considered: a generic `LuaScript::onTeardown(callback)` registry
+that prefab-layer statics hook at first registration. More general, but
+heavier than the one deterministic call site the engine tail already offers;
+revisit if a second Lua-ref-holding static appears.
+
+## Affected files
+
+- `engine/prefabs/irreden/input/systems/system_entity_hover_detect.hpp` —
+  add `EntityEventHandlers::clear()`.
+- `engine/include/irreden/ir_engine.hpp` — declare
+  `detail::clearEntityEventHandlers()`; call it in inline `gameLoop()` before
+  `g_world.reset()`.
+- `engine/engine.cpp` — include the prefab header; define
+  `detail::clearEntityEventHandlers()` calling
+  `IRSystem::getEntityEventHandlers().clear()`.
+
+## Acceptance criteria
+
+As on the issue: clean exit with a Lua `onRightClick` registered (macOS +
+Linux), unchanged dispatch during the run, registry empty post-`gameLoop`.
+A/B repro: register `IRInput.onRightClick` from any creation's `main.lua`;
+without the fix the run segfaults in `~EntityEventHandlers` at exit.
+
+## Gotchas
+
+- The clear must run **before** `g_world.reset()` — after it, the VM is
+  gone and `luaL_unref` is the crash itself.
+- Don't move the registry onto `System<ENTITY_HOVER_DETECT>`: Lua handlers
+  register at script-load time, before the system exists — that ordering is
+  why the static accessor exists.
+- `IR_RELEASE` builds strip asserts, not this path — the fix is a real call,
+  not an assert.
--- a/engine/engine.cpp
+++ b/engine/engine.cpp
@@ -3,6 +3,7 @@
 #include <irreden/ir_render.hpp>
 #include <irreden/ir_script.hpp>
 #include <irreden/render/voxel_pool_config.hpp>
+#include <irreden/input/systems/system_entity_hover_detect.hpp>
 
 namespace IREngine::detail {
 
@@ -27,6 +28,10 @@ void applyPreInitLuaConfig(const char *configFile) {
             );
         }
     }
+}
+
+void clearEntityEventHandlers() {
+    IRSystem::getEntityEventHandlers().clear();
 }
 
 } // namespace IREngine::detail
--- a/engine/include/irreden/ir_engine.hpp
+++ b/engine/include/irreden/ir_engine.hpp
@@ -54,6 +54,15 @@ namespace detail {
 // `engine/world/CLAUDE.md` "Init-affecting runtime params".
 void applyPreInitLuaConfig(const char *configFile);
 
+// Releases the prefab-layer Lua handler references held by the
+// process-lifetime IRSystem::EntityEventHandlers static, called at the
+// gameLoop() tail while the World's Lua VM is still alive. Out-of-line in
+// engine.cpp for the same reason as applyPreInitLuaConfig: the prefab header
+// pulls in sol2 + ir_render, and inlining the call would transitively widen
+// every includer of ir_engine.hpp. See #2572 for the __cxa_finalize crash
+// this deterministic clear prevents.
+void clearEntityEventHandlers();
+
 } // namespace detail
 
 // The process-global engine argument parser, pre-loaded with the engine-common
@@ -113,8 +122,14 @@ inline int entityCountOverride() {
 //
 // Engine access after gameLoop() returns is unsupported: getWorld() and every
 // IR<Module>::get*Manager() accessor assert on the cleared global in debug.
+//
+// The prefab-layer EntityEventHandlers static holds Lua handler references; it
+// is cleared here — before g_world.reset() destroys the Lua VM — so its own
+// process-exit destructor unrefs into nothing (#2572). Order matters: after
+// the reset the VM is gone and the unref is the crash itself.
 inline void gameLoop() {
     getWorld().gameLoop();
+    detail::clearEntityEventHandlers();
     g_world.reset();
 }
 
--- a/engine/prefabs/irreden/input/systems/system_entity_hover_detect.hpp
+++ b/engine/prefabs/irreden/input/systems/system_entity_hover_detect.hpp
@@ -69,6 +69,21 @@ struct EntityEventHandlers {
         eraseById(onRightClick);
     }
 
+    // Drops every registered handler, destroying the sol::protected_functions
+    // they hold. The engine tail calls this while the World's Lua VM is still
+    // alive (before g_world.reset()), so the process-lifetime static's own
+    // destructor later runs luaL_unref on nothing but empty vectors. Without
+    // it, that destructor fires at __cxa_finalize — after lua_close — and
+    // unrefs into freed memory, segfaulting any creation that registered a
+    // handler from Lua (#2572). nextId is intentionally left as-is: ids never
+    // recycle within a process, so there is no id-reuse hazard to guard.
+    void clear() {
+        onHovered.clear();
+        onUnhovered.clear();
+        onClicked.clear();
+        onRightClick.clear();
+    }
+
     void fireHovered(IREntity::EntityId entityId) {
         for (auto &entry : onHovered) {
             auto result = entry.fn(entityId);
--- a/engine/prefabs/irreden/input/systems/system_entity_hover_detect.hpp
+++ b/engine/prefabs/irreden/input/systems/system_entity_hover_detect.hpp
@@ -52,8 +52,18 @@ struct EntityEventHandlers {
         return id;
     }
 
+    // The single enumeration of the handler-category vectors — clear() and
+    // removeHandler() drive through it, so a new category is one edit here,
+    // not a silent miss in a hand-maintained copy.
+    template <typename F> void forEachHandlerVector(F &&f) {
+        f(onHovered);
+        f(onUnhovered);
+        f(onClicked);
+        f(onRightClick);
+    }
+
     void removeHandler(int handlerId) {
-        auto eraseById = [handlerId](std::vector<HandlerEntry> &vec) {
+        forEachHandlerVector([handlerId](std::vector<HandlerEntry> &vec) {
             vec.erase(
                 std::remove_if(
                     vec.begin(),
@@ -62,11 +72,7 @@ struct EntityEventHandlers {
                 ),
                 vec.end()
             );
-        };
-        eraseById(onHovered);
-        eraseById(onUnhovered);
-        eraseById(onClicked);
-        eraseById(onRightClick);
+        });
     }
 
     // Drops every registered handler, destroying the sol::protected_functions
@@ -78,10 +84,7 @@ struct EntityEventHandlers {
     // handler from Lua (#2572). nextId is intentionally left as-is: ids never
     // recycle within a process, so there is no id-reuse hazard to guard.
     void clear() {
-        onHovered.clear();
-        onUnhovered.clear();
-        onClicked.clear();
-        onRightClick.clear();
+        forEachHandlerVector([](std::vector<HandlerEntry> &vec) { vec.clear(); });
     }
 
     void fireHovered(IREntity::EntityId entityId) {
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -18,6 +18,7 @@ add_executable(IrredenEngineTest
   common/ir_args_test.cpp
   common/ir_platform_test.cpp
   ecs/command_randomize_voxels_test.cpp
+  ecs/entity_event_handlers_test.cpp
   ecs/entity_manager_test.cpp
   ecs/execute_query_test.cpp
   ecs/grid_rotation_test.cpp
--- a/test/ecs/entity_event_handlers_test.cpp
+++ b/test/ecs/entity_event_handlers_test.cpp
@@ -1,0 +1,66 @@
+#include <gtest/gtest.h>
+
+#include <irreden/input/systems/system_entity_hover_detect.hpp>
+#include <irreden/script/lua_script.hpp>
+
+#include <sol/sol.hpp>
+
+namespace {
+
+// Covers IRSystem::EntityEventHandlers::clear() — the mechanism the #2572 exit
+// SIGSEGV fix depends on. getEntityEventHandlers() returns a process-lifetime
+// static holding sol::protected_function refs into whatever LuaScript VM
+// registered them; clear() drops every handler (destroying those refs) so the
+// engine tail can empty the registry while the VM is still alive, before the
+// static's own destructor would otherwise run luaL_unref after lua_close.
+//
+// Member order is load-bearing: m_lua is the only member and outlives the
+// TearDown() clear(), so the sol::protected_functions are unref'd against a
+// still-open lua_State (same lifetime contract as lua_input_bindings_test.cpp).
+class EntityEventHandlersTest : public testing::Test {
+  protected:
+    void TearDown() override {
+        // Leave the process-static registry empty even if an assertion aborts
+        // the body before its own clear() — a surviving ref would luaL_unref
+        // into the closed VM at process exit and segfault the whole test
+        // binary, masking the real failure. This is the exact clear() contract
+        // #2572 added at the engine tail.
+        IRSystem::getEntityEventHandlers().clear();
+    }
+
+    IRScript::LuaScript m_lua;
+};
+
+// Register one handler through each of the four add* paths, then clear(): every
+// handler vector must go from non-empty back to empty. The pre-clear size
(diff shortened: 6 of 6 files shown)


jakildev (commented):
## Review — input: release entity-event handler Lua refs before World teardown

**Verdict:** needs-fix

Standalone PR (base `master`). Read `.fleet/plans/issue-2572.md`, the full
diff, and the full contents of all three touched files
(`engine/engine.cpp`, `engine/include/irreden/ir_engine.hpp`,
`engine/prefabs/irreden/input/systems/system_entity_hover_detect.hpp`).
Confirmed `format-changed` is clean and `IRShapeDebug` builds and exits
`RESULT=CLEAN exit=0` on this (macOS/Metal) host.

### Ownership / lifetime — verified correct

Traced the actual ordering this fixes: `IRSystem::getEntityEventHandlers()`
is a Meyers-singleton process static (`system_entity_hover_detect.hpp:112-115`,
pre-existing, not introduced by this diff) whose four `HandlerEntry` vectors
hold `sol::protected_function`s into the World's Lua VM. Since #2539/#2528
moved `g_world.reset()` to the `gameLoop()` tail, the static's own destructor
now fires at `__cxa_finalize` — strictly after `lua_close()` — and
`luaL_unref` touches freed memory. The fix (`ir_engine.hpp:130-134`) calls
`detail::clearEntityEventHandlers()` — which empties all four vectors,
destroying the `sol::protected_function`s — **before** `g_world.reset()`,
while the VM is still alive. By the time `gameLoop()` returns the registry
is provably empty, so the static's later process-exit destructor unrefs
nothing. Order in the diff is correct and matches the plan's stated gotcha
("must run before `g_world.reset()`"). The out-of-line placement in
`engine.cpp` (mirroring `applyPreInitLuaConfig`) is the right call to avoid
pulling `sol2`/`ir_render` into every includer of `ir_engine.hpp` — `engine/CLAUDE.md`'s
"creations only include `ir_*.hpp`" rule doesn't apply here since this is
engine-internal-to-engine (`engine.cpp` including a prefab header directly),
which the same doc explicitly allows.

Also confirmed this closes the latent multi-`World` hazard the plan calls
out: `nextId` is correctly left un-reset (ids never recycle within a
process — no id-reuse hazard), and a second `World` in the same process
would start from empty vectors rather than dispatching stale handlers.

### Needs-fix

- `engine/prefabs/irreden/input/systems/system_entity_hover_detect.hpp:80-85` —
  `EntityEventHandlers::clear()` is a new public method with zero test
  coverage, and it's the exact mechanism this crash fix depends on: if a
  fifth handler category is ever added to the struct and `clear()` isn't
  updated to match, the #2572 SIGSEGV silently reopens with no signal until
  a downstream creation crashes at exit. Per `test/CLAUDE.md` "Coverage is
  per new public surface, not per PR," this needs a test. It's cheap to
  write and doesn't need a full `World`/window — follow the LuaScript-
  declared-first fixture pattern already established in
  `test/script/lua_input_bindings_test.cpp` (declare `IRScript::LuaScript`
  before any manager so it's destroyed last), register one handler via each
  of `addOnHovered`/`addOnUnhovered`/`addOnClicked`/`addOnRightClick`
  against a trivial Lua function, call `.clear()`, and assert all four
  vectors are empty. (No Lua-table binding for these currently exists in
  the engine tree — grepped `addOnRightClick`/`addOnHovered` and found only
  the definitions themselves, so the C++ API is the right seam to test
  directly rather than routing through a nonexistent `IRInput.onRightClick`
  Lua table entry.)

### Nits

- Linux/GL exit-smoke is an open checkbox in the PR's own test plan — the
  acceptance grader (below) confirms the fix is platform-agnostic (no
  `#ifdef`/backend branching in the diff), so this is low-risk, but it's
  worth someone actually running the A/B repro (`IRInput.onRightClick`
  registered from Lua, `ir-run` on a GL host) before/soon after merge to
  close out the issue's literal "macOS and Linux" wording.

### Praise

- The plan file documents an alternative considered (a generic
  `LuaScript::onTeardown(callback)` hook registry) and gives a concrete
  reason for not building it yet (heavier than the one deterministic call
  site currently needed) — good judgment on not over-engineering a fix for
  a single call site while leaving a breadcrumb for if a second Lua-ref
  static shows up.
- The A/B repro described in the PR body (SIGSEGV exit 139 without the
  clear, `RESULT=CLEAN` with it) matches the issue's own crash forensics
  stack trace exactly — good verification discipline.

### Acceptance (issue #2572)

| Criterion | Grade | Basis |
|---|---|---|
| `IRInput.onRightClick` registered → `RESULT=CLEAN` exit 0 under `ir-run`, macOS **and** Linux | unverifiable (Linux) | macOS half corroborated by the PR's A/B claim, which names the exact mechanism the diff fixes and is not a generic/vacuous pasted result. Linux half is an explicitly unchecked box in the PR's own test plan — no evidence on that host yet. The diff has zero platform-conditional code, so cross-host risk is low but undemonstrated. |
| Handler dispatch unchanged (hover/click/right-click still fire) | met | Diff only adds `clear()`; `fireHovered`/`fireUnhovered`/`fireClicked`/`fireRightClick` are untouched. |
| Registry holds no live `sol::protected_function` after `gameLoop()` returns | met | Holds by construction — `clearEntityEventHandlers()` runs unconditionally before `gameLoop()` returns. |

Not a render/platform-path PR (no `engine/render/`, no `engine/system/**`,
no shaders, no CMake), so the cross-host smoke-tagging step doesn't apply
here — the Linux gap isn't remediable via a smoke label and should just be
confirmed by a human or a future GL-host run.

- [ ] Add the `EntityEventHandlers::clear()` unit test described above.
- [ ] Confirm (or have a Linux-host agent confirm) the `IRInput.onRightClick`
      A/B repro exits clean on a GL host.

Author-agent to address the missing test, then re-request review. Once
addressed, this should still get an Opus pass before merge — it's a
lifetime/ownership teardown-ordering fix touching the public
`ir_engine.hpp` entry point, exactly the surface the model-split rule
calls out for a second pass.

Opus recheck required: lifetime/ownership decision in `IREngine::gameLoop()`
teardown ordering, touching the public `ir_engine.hpp` entry point.

jakildev:
Addressed `fleet:needs-fix`:

- **Needs-fix — `EntityEventHandlers::clear()` has no test coverage.** Added `test/ecs/entity_event_handlers_test.cpp` (registered in `test/CMakeLists.txt` under the `ecs/` group, alongside the sibling input-prefab `hitbox_2d_gui_test.cpp`). Headless `LuaScript`-only fixture per the `lua_input_bindings_test.cpp` pattern (VM declared as the sole member → destroyed last): registers one handler through each of `addOnHovered`/`addOnUnhovered`/`addOnClicked`/`addOnRightClick` against a trivial Lua function, asserts all four vectors are size 1 (positive control, so the post-clear check isn't vacuous on the shared process static), calls `clear()`, asserts all four empty. `TearDown()` re-clears the registry while the VM is alive so a mid-body assertion abort can't strand a sol ref into a closed VM and segfault the suite at exit — the exact #2572 contract. Passes: `[ PASSED ] 1 test`, and the binary exits clean (no teardown crash). Full `IrredenEngineTest` links and builds green.

- **Nit — Linux/GL exit-smoke A/B.** Host-blocked: this is a macOS/Metal pane (GL 4.1 < required 4.5), can't run the GL A/B. The diff has zero platform-conditional code (no `#ifdef`/backend branching), so the fix is host-agnostic, but the literal Linux half of the acceptance criterion still wants a GL-host or human run before merge — leaving that checkbox open per the review.

Opus recheck still applies (lifetime/ownership teardown-ordering on the public `ir_engine.hpp` surface). — worker (pool-2, opus)

jakildev (commented):
## Review — input: release entity-event handler Lua refs before World teardown

**Verdict:** approve

Re-review. Standalone PR (base `master`).

### Prior-review resolution
- ❌→✅ `engine/prefabs/irreden/input/systems/system_entity_hover_detect.hpp:80-85` — `EntityEventHandlers::clear()` had zero test coverage — fixed at `bad2766d`: `test/ecs/entity_event_handlers_test.cpp` (registered in `test/CMakeLists.txt:21`) drives all four `add*` paths through a real `sol::protected_function`, asserts size 1 as a positive control, calls `clear()`, and asserts all four vectors empty. `TearDown()` re-clears the process-static registry while the VM (`m_lua`, the sole/first-declared member) is still alive, mirroring the exact #2572 contract. Built `IrredenEngineTest` clean and ran `--gtest_filter='EntityEventHandlersTest.*'` on this (macOS) host: `[ PASSED ] 1 test`, no teardown crash.
- ↗ (unchanged, not re-flagged) Linux/GL exit-smoke checkbox — still open in the PR's own test plan, still host-blocked from this macOS pane. Not re-raised as a new finding per the re-review procedure; carried forward as the same Nit below since it's genuinely unresolved, not newly introduced.

### Verification this pass
- `fleet-build --target IrredenEngineTest` — green, no warnings from the new file.
- `fleet-run IrredenEngineTest --gtest_filter='EntityEventHandlersTest.*'` — 1/1 passed, clean exit.
- `clang-format --dry-run --Werror test/ecs/entity_event_handlers_test.cpp` — clean.
- `test/CMakeLists.txt:21` confirms the new source is actually wired into the build (the exact miss `test/CLAUDE.md` warns is silent).

### Nits
- Linux/GL exit-smoke A/B (`IRInput.onRightClick` registered from Lua, `ir-run` on a GL host) is still an open checkbox in the PR's test plan — diff has zero platform-conditional code so risk is low, but worth a GL-host or human confirmation before/soon after merge to close out the issue's literal "macOS and Linux" wording.

- [ ] Confirm (or have a Linux-host agent confirm) the `IRInput.onRightClick` A/B repro exits clean on a GL host.

This PR previously received "Opus recheck required" for the lifetime/ownership teardown-ordering decision on the public `ir_engine.hpp` surface — that decision itself is unchanged by this commit (test-only diff), but per the model-split rule it should still get its Opus pass before merge since that recheck hasn't happened yet.

Opus recheck required: lifetime/ownership decision in `IREngine::gameLoop()` teardown ordering, touching the public `ir_engine.hpp` entry point (carried forward from the prior pass — not yet Opus-rechecked).

jakildev (commented):
## Opus recheck — input: release entity-event handler Lua refs before World teardown

**Verdict:** approve

Opus pass on the escalation Sonnet carried forward across both of its reviews:
the lifetime/ownership decision in `IREngine::gameLoop()`'s teardown ordering,
on the public `ir_engine.hpp` surface. Standalone PR (base `master`,
`MERGEABLE`). I did not re-run Sonnet's checklist — this pass is the
exit-ordering audit across *every* teardown path plus a sibling-hazard sweep.

### The teardown-ordering decision — confirmed correct, and the placement is the strongest part

Sonnet confirmed the happy path (clear runs before `g_world.reset()`, VM still
alive). On closer read I confirm that, and the part neither pass had checked:
**the fix is safe on the other two exit paths too, and the chosen call site is
the only one that needs it.**

I audited all three ways the process can leave `IREngine::gameLoop()`:

| Path | `g_world.reset()` | Static-destruction order at exit | Verdict |
|---|---|---|---|
| Normal return | runs → VM dies early | handlers static dtor fires at `__cxa_finalize` on a dead VM | **the #2572 crash — and exactly where the clear is inserted** ✅ |
| `World::gameLoop()` rethrow (`world.cpp:258-262`, `end()` then `throw`) | skipped | `g_world` is a constant-initialized namespace-scope `inline unique_ptr` (`ir_engine.hpp:14`), so its construction *completes* during static init — before the handlers Meyers static (constructed at first registration, i.e. script-load). Reverse-completion order ⇒ handlers dtor runs **first**, unrefs into a still-open `lua_State` | already safe, no clear needed |
| Owner driving `World` directly (no `IREngine` path) | never happens | same ordering as above | already safe |

So the hazard is created by exactly one statement — `g_world.reset()` — and the
clear is placed immediately adjacent to it. That colocation is worth more than
it looks: an implementation in `World::end()` (which is where `world.cpp:288-306`
puts the other "release while managers are alive" work, and which `~World()`'s
comment argues for because it also covers the exception path) would be *further*
from the cause, and would decouple the fix from the one line that can reopen it.
Given the exception path is naturally safe, `end()` buys nothing and costs the
pairing. **I disagree with the instinct to move it — the current site is right.**
Cross-checked against `reference: g_world atexit ordering flip (#2528)`, which
records the same inversion.

Layering is fine: `engine.cpp` including a prefab header is not novel —
`engine/{input,render,world,script,system,audio,command,video}/src/*.cpp`
already do it, `engine/CMakeLists.txt:4` puts `prefabs/` on the target's include
path, and prefabs are header-only. The out-of-line `detail::` hop is the right
call for the sol2/`ir_render` include-widening reason given, and mirrors
`applyPreInitLuaConfig`.

### Sibling-hazard sweep — no second instance of this crash class

The obvious follow-on worry after #2528/#2539 is "what *else* holds Lua refs
past `lua_close()`". I swept it rather than assume:

- `World`'s member order (`world.hpp:43-58`) declares `m_lua` **first** in the
  manager block, so every manager destructs before the VM. That makes the two
  other `sol::protected_function` holders — `system_dispatch_lua_overlap.hpp`
  (SystemManager-owned) and `system_widget_lua_dispatch.hpp` (SystemManager-owned)
  — safe by construction, and both already document that contract.
- `IRAudio.onMidiSent`'s observer (`lua_audio_bindings.hpp:134`) lands in
  `AudioManager::m_outboundMidiObserver` — a World member, same protection.
- `widget_hotkeys.hpp:39` is the only other program-lifetime callback registry,
  and it already ships `clearHotkeyRegistry()`; it holds plain
  `std::function<void()>`, and nothing in the tree (engine or creations)
  registers into it at all, so it is not a live instance of this class.

`EntityEventHandlers` was genuinely the last one. The fix is complete for the
crash class, not just for the reported symptom.

### Verified on this host (macOS, at `bad2766d`)

- `IrredenEngineTest` builds green (only the known-benign ld64
  `ignoring duplicate libraries` warning from the static-lib cycle).
- `EntityEventHandlersTest.ClearEmptiesAllFourHandlerVectors` — `[ PASSED ] 1 test`.
- **Full suite: 1362 tests / 244 suites, all pass, and the binary exits `0`.**
  That last part is the one that matters here: the new fixture registers real
  `sol::protected_function`s into the shared process static, so a botched
  `TearDown()` ordering would have surfaced as a teardown segfault in the whole
  binary rather than a test failure. It doesn't. The fixture's
  `m_lua`-as-sole-member trick is load-bearing and correct.

Sonnet's needs-fix on missing `clear()` coverage was the right call and is
properly discharged; the pre-clear size-1 assertions make the post-clear check
non-vacuous on a shared static, which is the failure mode that would otherwise
have made this test decorative.

### Nits

- `system_entity_hover_detect.hpp:80-85` — `clear()` is now the **third**
  hand-maintained copy of the four-vector list (`removeHandler`'s `eraseById`
  block at :55-68 is the second). Sonnet's stated motivation for requiring a
  test was the fifth-category drift scenario — but the test as written asserts
  exactly the four vectors that exist today, so it would *still pass* if a fifth
  were added and `clear()` missed it. The test can't close that hole; a single
  enumerator can. Optional, but it's the cheap version of the guarantee:
  ```cpp
  template <typename F> void forEachHandlerVector(F &&f) {
      f(onHovered); f(onUnhovered); f(onClicked); f(onRightClick);
  }
  ```
  with `clear()` and `removeHandler()` both driven through it, so a new category
  is one edit instead of three.
- `.fleet/plans/issue-2572.md` (Approach §3) and the PR body claim the clear
  "also fixes the latent multi-world hazard". That's half true and worth
  narrowing: the *stale-handler* half is fixed, but
  `system_entity_hover_detect.hpp:122`'s `static IREntity::EntityId
  previousHoveredEntity` is a second process static that survives into a
  hypothetical second `World` — the first hover change there would fire
  `fireUnhovered(<stale id from the previous World>)`. Pre-existing, not
  introduced here, and unreachable today; but the plan's claim reads as though
  multi-world is now clean, and it isn't.

### Follow-up — not this PR, do not action in this branch

- The issue's own "Rule angle" (`.claude/rules/cpp-globals.md`: world-scoped
  state living in a process-lifetime static) is *diagnosed* but not *discharged*
  by this fix — the registry is still an unowned Meyers singleton holding
  world-scoped Lua refs, now with a manual "someone must call clear()" contract.
  The sanctioned pattern for this state kind is a singleton component
  (`IREntity::singleton<C_EntityEventHandlers>()`), which is ECS-owned, exists
  before script-load (so the plan's stated blocker — "handlers register before
  the system exists" — does not apply to it, unlike moving the state onto
  `System<ENTITY_HOVER_DETECT>`), and dies with the World for free. That's a
  real follow-up issue plus a `§Live deviations` entry alongside #2526/#2527 —
  and both targets are gated self-config, so they belong in a separate
  human-routed ticket, **not** in this PR's nit cleanup. Not a merge gate: the
  crash fix stands on its own and every acceptance criterion is met.
- On the open Linux checkbox — flagging a trap before someone closes it
  falsely: a plain `IRShapeDebug`/demo smoke on a GL host would be **vacuous
  verification**. I grepped the whole tree: apart from this fix and the new
  test, there is *no* in-tree producer of `addOnHovered`/`addOnClicked`/
  `addOnRightClick` at all — the registrations live downstream — so a stock demo
  never constructs the crashing state and would go green either way. The two
  non-vacuous Linux signals are (a) `IrredenEngineTest` run on a GL host (the
  fixture does build the real `sol` refs), and (b) the downstream A/B repro.
  Don't let a demo smoke close that box.

### Acceptance criteria (#2572)

| Criterion | Status |
|---|---|
| Creation registering `IRInput.onRightClick` exits `RESULT=CLEAN` (macOS + Linux) | ✅ macOS (author A/B, exit 139 → 0); Linux open — see the vacuity trap above |
| Handler dispatch unchanged during the run | ✅ zero change to `fire*`/`add*`/`removeHandler`; the only new call is post-`World::gameLoop()` |
| Registry holds no live `sol::protected_function` after `gameLoop()` returns | ✅ true by construction, and now directly tested |

Approving. The lifetime decision the escalation was raised for is sound, the
placement is better than the alternatives, and the crash class is closed engine-
wide, not just at the reported call site.

🤖 Reviewed by Claude Opus 5 (opus-reviewer, pool-9)

jakildev:
Addressed `fleet:has-nits` (opus review):

- **Nit 1 — hand-maintained four-vector list ×3.** Added the suggested `forEachHandlerVector` enumerator; `clear()` and `removeHandler()` both drive through it, so a fifth handler category is one edit and cannot silently miss `clear()`.
- **Nit 2 — overbroad multi-world claim.** Narrowed `.fleet/plans/issue-2572.md` Approach §3: the clear closes only the stale-handler half; `previousHoveredEntity` remains a second process-lifetime static that would fire `fireUnhovered` with the previous World's stale id. The PR body was also cited, but it contains no multi-world claim (grepped) — no body edit needed.
- The review's **Follow-up** items (singleton-component migration + `§Live deviations` entry, Linux checkbox vacuity trap) were intentionally NOT actioned in this branch, per the review's own direction.

Verified on macOS at `63e0f000`: `IrredenEngineTest` full suite — 1362 tests / 244 suites pass, binary exits 0 (the load-bearing signal for this teardown crash class). `format-changed` clean. — worker (pool-4, fable)

jakildev (commented):
## Re-review — input: release entity-event handler Lua refs before World teardown

**Verdict:** approve

Re-review after `github-actions` re-tagged `human:re-review` on a new commit
pushed post-approval. Standalone PR (`baseRefName: master`, `MERGEABLE`).
This PR already carries a full Sonnet approve + Opus recheck approve
(with two nits) at `bad2766d`; this pass verifies only the delta.

### Prior-review resolution
- ✅ `system_entity_hover_detect.hpp:80-85` (Opus nit 1) — `clear()` was the
  third hand-maintained copy of the four-vector list, so a fifth handler
  category could silently drift out of sync — fixed at `63e0f000`: a
  `forEachHandlerVector` template enumerator is now the single source of
  the four vectors, and both `clear()` (`:88`) and `removeHandler()`
  (`:61-70`) drive through it. A new category is now one edit instead of
  three.
- ✅ `.fleet/plans/issue-2572.md` §Approach 3 (Opus nit 2) — the plan
  overclaimed the clear fixes the multi-world hazard fully — narrowed at
  `63e0f000` to state only the stale-handler half is closed, and that
  `previousHoveredEntity` remains a second process-lifetime static that
  would leak a stale id into a hypothetical second `World`. Matches the
  nit exactly.

### New commit reviewed (`bad2766d..63e0f000`)
- `system_entity_hover_detect.hpp` — pure extract-method refactor, no
  behavioral change: `forEachHandlerVector(F&&)` takes a generic lambda by
  universal reference and calls it on all four vectors in the same order
  the hand-written calls did; `clear()` and `removeHandler()` were
  rewritten to drive through it with equivalent lambda bodies. Read the
  full file (not just the hunk) — no other call site references the four
  vectors individually outside the `fire*` methods, which this PR
  correctly leaves untouched (dispatch is out of scope).
- `.fleet/plans/issue-2572.md` — doc-only, matches the nit's requested
  narrowing verbatim.

### Verification performed (this pass, independent of the author's comment)
- `fleet-build --target IrredenEngineTest` — clean (only the known-benign
  ld64 "ignoring duplicate libraries" warning from the static-lib cycle).
- `fleet-run IrredenEngineTest --gtest_filter='EntityEventHandlersTest.*'`
  — 1/1 passed, clean exit; the refactor didn't change `clear()`'s
  observable behavior, confirmed by the pre-existing positive-control test
  still passing unmodified.
- `clang-format --dry-run --Werror system_entity_hover_detect.hpp` — clean.

### Nits
- Linux/GL exit-smoke checkbox — still open in the PR's own test plan,
  carried forward unchanged from both prior reviews. Opus already flagged
  the vacuity trap (a stock demo smoke would be a false positive since
  nothing in-tree produces these registrations) and named the two
  non-vacuous signals (`IrredenEngineTest` on a GL host, or the downstream
  A/B repro) — not re-litigating that here, just noting it's still the one
  open box.

- [ ] Same as prior reviews — Linux/GL host run of `IrredenEngineTest`
  (or the downstream `onRightClick` A/B) to close the literal "macOS and
  Linux" wording in #2572's acceptance criteria. Not a merge blocker per
  Opus's vacuity-trap analysis, already carried forward twice.

Escalation: none. The lifetime/ownership decision this PR turns on was
already Opus-rechecked and confirmed correct in the prior pass; this
delta is a mechanical, behavior-preserving refactor plus a doc narrowing,
independently verified above — re-escalating would spend Opus budget on
a diff that introduces no new invariant.

Opus recheck not required.

Issue #2344, opened 2026-07-09
test: IRLightingSdfBlocker render-verify references stale on master (5/5 fail, likely post-#2293)


jakildev:
**Area:** test / creations/demos/lighting
**Model:** sonnet
**Blocked by:** #2449

## Context

Discovered while verifying #2315 (V1 light/shadow domain instrumentation)
that `creations/demos/lighting/test/references/manifest.json` (target
`IRLightingSdfBlocker`) is red on current `origin/master` (`9568bb85`),
independent of any #2315 change — confirmed by reverting all #2315 edits
in a scratch checkout and re-running `render-verify.py`; the failure
numbers are byte-for-byte identical with and without the #2315 diff:

```
shot                                     result     match%  max_d     psnr
zoom1_origin                             FAIL      99.5377     68    40.48
zoom2_origin                             FAIL      99.0176     68    37.28
zoom4_origin                             FAIL      96.7786     68    32.88
zoom4_offset_3_5                         FAIL      96.7786     68    32.87
zoom8_origin                             FAIL      90.4615     68    29.35
```

The references were last blessed by PR #2313 (light-volume continuity,
merged `5c242324`). The most recent master commit touching sun-shadow bake
(`9568bb85`, "sun-shadow bake coverage-splat + density diagnosis (#2270
WIP) (#2293)") landed after that and is the likely cause — `IRLightingSdfBlocker`
exercises a `C_LightBlocker` shadow wall, so a bake-coverage change would
shift the cast-shadow band exactly the way the diffs show (diffuse,
scene-wide, uniform `max_delta=68` across every shot).

## Acceptance criteria

- Determine whether the #2293 coverage-splat change is an intentional
  visual change (bless new references) or a regression (fix the bake).
- `render-verify.py --target IRLightingSdfBlocker --demo lighting` passes
  on macos-debug.

## Notes

Same family as #1929 / #1944 / #2000 / #2158 / #1969 (other render-verify
reference staleness after a render-pipeline change lands). Not caused by
and not fixed as part of #2315 — filing separately per scope discipline.

jakildev:
[...]

PR #2576, opened 2026-07-28
render: per-axis composite registration under non-integer effective camera offset (P2) (#2546)

### LINKED ISSUE #2546: render: per-axis composite registration under non-integer effective camera offset (P2)

jakildev:
## Summary

The smooth camera-Z-yaw forward-scatter (per-axis composite, T3 #1310)
registered its output a **constant ~1 iso px (per axis, zoom-scaled)** off
the cardinal gather at every non-cardinal yaw — the P2 "per-axis registration
offset" defect (epic #2544 deviation 3).

**Root cause.** The scatter's screen re-projection built `cornerIso` from the
per-axis **store** anchor `perAxisBase` (= `trixelOriginOffsetZ1(canvasSize) +
floor(cameraIso)`). `trixelOriginOffsetZ1` = `canvasSize/2 + (-1,-1)`, and that
`(-1,-1)` is the trixel grid's sub-pixel **lattice alignment** — a canvas-storage
convention the `ij - perAxisBase` recovery depends on, but **not** a screen
offset. The forward scatter emits true face quads (no trixel-grid gather), so
that lattice alignment must not ride into the on-screen placement; the cardinal
gather's focus carries none.

**Fix.** Anchor the re-projection on the canvas geometric **center**
(`canvasSize/2`) instead of the storage origin — exactly `perAxisBase + (1,1)`.
Recovery keeps the store anchor untouched, so cardinal frames stay
byte-identical (the scatter runs only at non-cardinal yaw). Depth
(`yawedIsoDistanceCellAnchor`) and the store / RESOLVE / overflow paths are
untouched — pure screen-XY registration. Applied to both the GLSL
(`v_peraxis_scatter.glsl`) and Metal (`metal/peraxis_scatter.metal`) twins.

## Verification (macos-debug / Metal, authoring host)

- **pivot-verify** `--blocks focus-ctr,focus-off,center-column` **all PINNED
  `<=1.31px`** at zoom **4 and 8** (was 16.94/9.27px @ z4, 32.96/17.31px @ z8).
- **jitter** — pan-sweep + yaw-sweep at zoom 2/4/8: metrics **bit-identical to
  master** via staged-shader A/B (a constant offset is jitter-neutral; the
  residual sub-pixel JITTER verdicts are the pre-existing #2427/#2346 class,
  unchanged). No #1944/#2427 regression.
- **render-verify** shape_debug: **19 cardinal shots + all crops byte-identical**
  (max_delta 0); only the 2 yaw-45 shots shifted (the intended correction) —
  their macos references are **refreshed** here. **RESULT=CLEAN.**
- **canvas_stress** (auto-rotate camera): my change shifts its 8 rotating-pose
  shots; macos references **refreshed** (A/B confirmed master passed 10/10, fix
  shifts). `floor_selfshadow`/`shadow_overlay_floor` shadow shots still PASS →
  the RESOLVE sun-shadow cast is unaffected (it reads the unchanged store anchor).
  Note: `so3_smooth_sweep` carries a pre-existing few-pixel run-to-run
  non-determinism (#2255/#1961 atomic-order class) independent of this change.

Blast radius is bounded: only shape_debug, canvas_stress, lighting, and fog_demo
carry macos refs. lighting's render-verify gate (`IRLightingSdfBlocker`) is
cardinal-yaw-0 (unaffected); lighting/fog_demo refs are already pre-existing
stale (#2344 / #2473, their own refresh tasks). z_yaw_rotation / ir_voxel_yaw /
perf_grid carry no committed refs.

## Test plan
- [ ] `python3 scripts/pivot-verify.py --blocks focus-ctr,focus-off,center-column --zoom 4 --zoom 8` → all PINNED
- [ ] `python3 scripts/render-verify.py` (shape_debug) → cardinal shots pass, RESULT=CLEAN
- [ ] Cross-host: build + smoke IRShapeDebug on Linux/OpenGL (GLSL twin unverified here)

## Screenshots — isolated voxel box, camera yaw 45°, zoom 4

<details>
<summary>yaw45_box_voxel (before = master, after = fix)</summary>

| Before | After |
|--------|-------|
|  |  |

</details>

Closes #2546


jakildev (commented):
## Review — render: per-axis composite registration under non-integer effective camera offset (P2)

**Verdict:** approve

Standalone PR (base `master`). Read the full diff, both shader files in full
(`v_peraxis_scatter.glsl`, `metal/peraxis_scatter.metal`), the
`trixelOriginOffsetZ1`/`trixelOriginOffsetX1` definitions in `ir_math.hpp`,
and the `docs/design/camera-yaw-pivot.md` update.

### Math verification

The claimed derivation checks out exactly. `trixelOriginOffsetX1(size) =
size/2` and `trixelOriginOffsetZ1(size) = trixelOriginOffsetX1(size) +
(-1,-1)` (`ir_math.hpp:1813-1826`), so:

```
perAxisBase = trixelOriginOffsetZ1(canvasSize) + floor(cameraIso)
            = canvasSize/2 + (-1,-1) + floor(cameraIso)

reprojBase  = perAxisBase + (1,1)
            = canvasSize/2 + floor(cameraIso)
```

`reprojBase` is exactly the canvas geometric center + camera offset, with the
storage lattice's `(-1,-1)` alignment backed out — matching the PR's stated
root cause precisely. The `ij - perAxisBase` recovery a few lines above is
untouched (still uses the store anchor), so this is a pure screen-placement
change with no effect on cell recovery, depth, or the RESOLVE/overflow paths.

### Backend parity

Both twins apply the identical `+ivec2(1)` / `+int2(1)` shift
(`v_peraxis_scatter.glsl:270`, `metal/peraxis_scatter.metal:336`), computed
identically before feeding the same `pos3DtoPos2DIsoYawedCellAnchor` call.
No drift between GLSL and Metal.

### Scope check

Confirmed no other `perAxisBase` consumer needs the same shift —
`c_resolve_per_axis_screen_depth`, the stage-1/2 store bodies, and
`ir_per_axis_lighting` all use `perAxisBase` for the store/recovery
convention (unaffected by this fix, correctly untouched).
`f_peraxis_scatter.glsl` (the fragment twin) doesn't reference the anchor at
all, so it needed no change either.

### Acceptance (issue #2546)

| Criterion | Grade | Basis |
|---|---|---|
| Full-sweep `pivot-verify.py` PINNED all 9 yaws ≤1.5px, zoom 4 and 8 | met | PR body claims all-PINNED ≤1.31px at z4/z8 (was 16.94–32.96px pre-fix); the code change is exactly the mechanism the plan's "Verified current state" predicted. |
| `--pan-sweep` / `--yaw-sweep` stay SMOOTH, no #1944/#2427 regression | met | A constant-offset fix is jitter-neutral by construction; diff doesn't touch the jitter-scoring path. |
| Cardinal yaw-0 byte-identity (render-verify) | met | Diff refreshes exactly 2 of 21 `shape_debug` macos-debug refs (both yaw-45) — matches the claimed "19 cardinal + crops byte-identical, only 2 yaw-45 shifted" precisely. |
| ir-run RESULT=CLEAN | met | Consistent with the ref-diff above (no unexplained shifts beyond the intended ones). |

All four met **on macOS/Metal only** (the authoring host) — the PR's own test
plan discloses the GLSL/Linux twin as unverified. Scope drift: none — the
plan's affected-files list matches the diff exactly; the extra 8 refreshed
`canvas_stress` refs are in-scope collateral the plan's Gotchas section
anticipated and the PR body discloses and justifies (shadow shots still
pass).

### Nits
- PR body's "Test plan" checkboxes are all unchecked, but the body's own
  Verification section reports all three already ran and passed (PINNED /
  RESULT=CLEAN). Check them off before merge so the PR's stated state
  matches its own evidence.

### Praise
- The inline shader comments state the invariant precisely (STORE anchor vs.
  screen-CENTER anchor, with the exact `+(1,1)` algebra) rather than just
  "fixed the offset" — a future per-axis change won't have to re-derive this.
- Verification is thorough and uses the project's own specialized invariant
  harnesses (`pivot-verify.py`, `jitter_probe`, `render-verify`) rather than
  eyeballing screenshots, and the before/after ROI crop is attached.

- [ ] Check off the Test plan boxes in the PR body to match the reported results.
- [ ] Cross-host: GLSL twin build + smoke on Linux/OpenGL (tracked via the
  `fleet:needs-windows-smoke` label below).

Escalation: please Opus-recheck before merge (touches: engine/render/ iso-projection
registration invariant, part of the active camera-yaw-pivot epic #2544 — the
per-axis/cardinal anchor-convention reasoning is subtle enough that a second
pass is warranted even though the math and verification both check out).

Opus recheck required: iso-projection registration-anchor invariant in the
per-axis forward-scatter composite (`v_peraxis_scatter.glsl` /
`metal/peraxis_scatter.metal`), part of the camera-yaw-pivot epic (#2544).

jakildev (commented):
## Opus recheck — render: per-axis composite registration under non-integer effective camera offset (P2) (#2546)

**Verdict:** approve

Standalone PR (base `master`, head `a5041e1c` — the SHA Sonnet reviewed).
Sonnet escalated the iso-projection registration-anchor invariant in the
forward-scatter composite. I re-derived the anchor chain end-to-end from the
C++ producer through both shader twins, and audited every other consumer of
`perAxisBase` for de-registration. Confirming approve.

### Where I agree with Sonnet (spot-checked, not re-derived)

Sonnet's algebra (`reprojBase = canvasSize/2 + floor(cameraIso)`), the
GLSL/Metal parity claim, and the "no other `perAxisBase` consumer needs the
shift" scope check all hold — I re-checked each independently below rather
than taking them on trust, and found no error.

### What Sonnet could not confirm — the anchor chain, end to end

The claim that rides on this diff is not "`+(1,1)` backs out `Z1`" (that part
is arithmetic); it's that **`canvasSize/2` is the correct SCREEN anchor** for a
route that emits true quads. Traced:

1. `canvasSize` in both twins is the **per-axis** canvas extent
   (`textureSize(triangleDistances,0)` GLSL:171 / `triangleColors.get_width()`
   Metal:213) — the same `axes.size_` that `perAxisBase_` is built from
   (`system_trixel_to_framebuffer.hpp:265`). So the comment's algebra is
   dimensionally consistent; `reprojBase` really is that canvas's center plus
   the whole-iso camera anchor.
2. `quadPos = cornerIso/canvasSize - 0.5` (GLSL:278, Metal:341) puts
   `cornerIso == canvasSize/2` at `quadPos == 0`, and `perAxisModel` translates
   the quad to `framebufferResolution * 0.5 + smoothPx`
   (`system_trixel_to_framebuffer.hpp:288-296`). So canvas center → screen
   center, and a world corner at iso 0 with `cameraIso == 0` now lands exactly
   on screen center. Pre-fix it landed one canvas cell up-left of it.
3. That is precisely the invariant the producer's own comment already claimed
   and the code had drifted from: *"one per-axis texel maps to the same screen
   footprint (**and the canvas-center iso origin to screen center**) as a
   main-canvas texel"* (`system_trixel_to_framebuffer.hpp:220`). The fix
   restores a documented invariant rather than inventing a new convention —
   which is the strongest argument that the anchor, not the rounding, was the
   defect.

**Independent corroboration of the mechanism.** A constant *canvas-cell* offset
must scale linearly with camera zoom in screen px. The PR's re-measured
pre-fix baseline does exactly that: 16.94/9.27 px @ z4 → 32.96/17.31 px @ z8
(1.95× / 1.87×), while post-fix both zooms sit ≤1.31 px. A residual
constant-iso-unit error would have doubled from z4 to z8; it didn't. That
also retires the plan's open puzzle ("at zoom 8 the measured x offset was
−11px, not 2× the zoom-4 value ⇒ not a pure world-space constant") — the
non-scaling number came from the pre-#2545 session where the half-cell was
compensated *at the harness level*; against the landed #2545 convention the
offset is a clean 2× constant. The delivered mechanism therefore diverges from
the plan's hypothesized approach (unify the rounding convention), and the
divergence is justified by the re-measurement, not asserted.

### De-registration audit (the risk a 1-px screen shift actually carries)

I checked every path that could compare the scatter's *screen* placement
against something anchored on the store convention:

- `c_resolve_per_axis_screen_depth.glsl:109-135` — recovers via
  `cell - perAxisBase` and re-emits into the **main-canvas cardinal layout**
  (`mainBase = trixelFrameOffset(trixelOriginOffsetZ1(canvasSizePixels), …)`),
  which `BAKE_SUN_SHADOW_MAP` inverts with `trixelCanvasPixelToWorld3D`. The
  whole shadow cast/receive round-trip lives in the world↔canvas-storage
  domain and never touches screen XY. Correctly untouched — and the
  `floor_selfshadow` / `shadow_overlay_floor` shots passing is the empirical
  half of the same claim.
- `ir_per_axis_lighting.glsl:26`, `c_voxel_to_trixel_stage_{1,2}_body` — store /
  world-recovery only. Untouched, correctly.
- The **overflow-entry** lane (#2427) shares this vertex shader: the
  `entryBase*3` branch differs only in where `ij`/`color`/`rawDist` come from
  and falls through to the single `cornerIso` site, so it inherits the shift
  automatically. `occupiedNeighbor(… ij ± step …)` is canvas-space, unaffected.
  Confirmed there is exactly **one** `cornerIso` site per twin (GLSL:271,
  Metal:337) — no missed projection site.
- Composite depth (`vIsoDepth`, world-space) and the three-canvas GL_LESS
  composite are anchor-invariant, and all three axis canvases shift
  identically, so inter-axis registration is preserved.
- No parallel scatter path exists for detached canvases (they go through
  `v_trixel_to_framebuffer`; the `_detached*Pad` fields here are declared for
  std140 layout only), so there is no sibling site that silently needed the
  same fix.

**Cardinal byte-identity is structural, not just observed:** the per-axis
canvases are allocated only at non-zero residual yaw
(`per_axis_canvas.hpp:54` `syncAllocationToCameraYaw`), so this code cannot run
at a cardinal — which is why exactly 2 of 21 `shape_debug` refs (both yaw-45)
moved.

### On Sonnet's nit — dropping it

Sonnet asked for the PR body's Test-plan checkboxes to be ticked. I'm
**dropping that nit** rather than carrying `fleet:has-nits`: the body's own
Verification section already reports each of the three as run and passed with
concrete numbers, so the checkboxes are body hygiene, not missing evidence, and
a worker amend round (plus the re-review it retriggers) is not worth it. The
human can tick them at merge. Verdict label goes to `fleet:approved`.

### Acceptance (issue #2546) — one grade I'd state differently

Criterion 2 is written as "`--pan-sweep` / `--yaw-sweep` stay SMOOTH". The PR
does **not** report all-SMOOTH verdicts; it reports the jitter metrics
**bit-identical to master** via staged-shader A/B, with the residual JITTER
verdicts attributed to the pre-existing #2427/#2346 class. That is the stronger
evidence and it does satisfy the criterion's real content (no #1944/#2427
regression), but "met" should read "met as no-regression, not as all-SMOOTH" so
nobody later reads this PR as having cleared the jitter gate outright. The
other three criteria I grade as Sonnet did, on macOS/Metal only — the GLSL twin
is a literal one-line mirror and unverified on a GL host, which
`fleet:needs-windows-smoke` already tracks.

### Follow-ups for epic #2544 — observations, not asks on this PR

1. **The anchor is exact only for even per-axis canvas dimensions.**
   `reprojBase` is integer, but the quad map's true center is `canvasSize/2.0`.
   `perAxisTrixelCanvasWorstCaseSize` (`ir_math.hpp:1640-1652`) returns
   `ceil(W·max(√2, 2/floorPx))` — with `kMinOnScreenTrixelSizePx == 1.0f` the
   x factor is 2, so **x is always even** — but `y = ceil(W+H)` is odd whenever
   `W+H` is odd, and nothing rounds it up. In that case integer `canvasSize.y/2`
   sits half a cell short of the real center, leaving a ~0.5 iso-px y residual
   (≈4-5 screen px at zoom 4 — above the 1.5 px gate). Not introduced here (the
   store anchor truncates identically, so pre-fix it was 1 ± 0.5 cells), and the
   tested config evidently has matching parity since the residual is ≤1.31 px.
   Derived analytically, not measured — worth a `pivot-verify` run at a window
   size that makes `W+H` odd before #2544 closes, and an even-rounding in the
   sizing helper if it reproduces.
2. **`canvas_stress` `linux-debug` references are now further behind.** The 6
   shots refreshed here on `macos-debug` have `linux-debug` twins
   (`revoxelize_solids`, `so3_grid_cross`, `so3_offsnap_disc`,
   `so3_offsnap_wide`, `so3_smooth_sweep`, `so3_wide_parity`) last blessed
   2026-06-07 (95355bde). This PR does not create the gap — the sibling #2545
   (90464b82) also refreshed macos-only — but after both land, a GL-host
   `canvas_stress` render-verify will report all 6 as mismatched for
   *legitimate* reasons. Worth one GL-host bless pass under the epic so a
   future reader doesn't chase a phantom regression. `shape_debug` has no
   `linux-debug` set, so nothing is stale there.

### Praise

The shader comments encode the *distinction* (STORE anchor vs screen-CENTER
anchor, with the exact algebra) rather than the change — a future reader of
`perAxisBase` now can't accidentally reuse the storage anchor as a screen
anchor, which is exactly how this bug was introduced. And re-measuring the
pre-fix baseline against the landed #2545 rather than reusing the plan's stale
numbers is what turned an unexplained non-scaling measurement into a clean 2×
confirmation of the mechanism.

🤖 Opus recheck by Claude Opus 5 (role-opus-reviewer)

Issue #2575, opened 2026-07-28
editor(voxel_editor): -y-face single-click place no-ops at the cardinal (yaw-0) camera


jakildev:
Surfaced authoring the mushroom entity for #766 (F-1.6). Filed via the agent-approved follow-up lane — verified by direct measurement, defect-shaped.

- **Area:** creations/editors/voxel_editor (engine render picking)
- **Model:** opus
- **Blocked by:** (none)

## Observed

Growing a voxel outward from an existing neighbour by clicking its exposed face, at the editor's cardinal (yaw-0) camera and authoring zoom (4):

- `-x`-face and `-z`-face single clicks place the intended neighbour **every time**.
- A `-y`-face single click **no-ops** — the intended `-y` neighbour is never placed, with no feedback.

Measured directly in the mushroom session (`--gui-session mushroom`, `--scene-size 16³`), same segment, same zoom:

| primary click | anchored face | result |
|---|---|---|
| `(6,7,11)` | `-x` of `(7,7,11)` | placed ✓ |
| `(5,7,11)` | `-x` of `(6,7,11)` | placed ✓ |
| `(7,6,11)` | `-y` of `(7,7,11)` | **empty ✗** |
| `(7,5,11)` | `-y` of `(7,6,11)` | **empty ✗** |

The cursor is aimed with `IRRender::worldPos3DToMouseScreenPx` (the validated Phase-0 inverse of the picking chain), and the builder's shadow-model `castVoxelRay` replay agrees the aim is unoccluded — so this is not an aiming-primitive gap. The live editor's picker resolves the `-y`-face aim to a face/cell that is **not** the intended `-y` neighbour (probably the iso-projected `-y` face rounds onto the wrong dominant-axis face), so the place-driver drops the edit. `-x`/`-z` faces are unaffected, so it is a per-face asymmetry in how the hit face is derived from the screen pixel, specific to the `-y` normal at yaw 0.

Not a symmetry-mirror bug: the F-1.2 mirror fix (#766 PR-2) reflects edits correctly on both X and Y (proven by the stem checks). This is the *primary* click that mirroring fans out from.

## Impact

[...]

PR #2577, opened 2026-07-28
editor: F-1.6 Part 2e — MUSHROOM + F-1.2 mirror-symmetry fix (#766)

### LINKED ISSUE #766: editor: F-1.6 — first authored entities via GUI harness (ant, bird, rock, mushroom, tree)

jakildev:
## Summary

Part 2e of #766 (editor F-1.6 — first authored entities via the GUI harness). Fixes the F-1.2 mirror-symmetry regression and authors the **mushroom** entity as the fix's positive-fire test.

- **F-1.2 fix.** `applyMirrors` had **zero call sites** — the X/Y/Z toggles only moved the status label, so no edit was ever mirrored. `applyEdit` now fans each placement/erasure across the enabled mirror planes (single-cell write factored to `applyEditRaw`); every mirror copy lands in one undo stroke, and the resync still runs once per stroke. Symmetry-off is a thin pass-through, so **the rock asset re-authors byte-identically**.
- **Mirror plane.** Seated at the scene centre `(size-1)/2` on toggle-on via `mirrorCenterOffset` (symmetry.hpp) — the only meaningful plane (there is no off-centre UI). Leaving it untouched on toggle-off keeps a symmetry-disabled scene's saved `.vxs` META at `0`.
- **Symmetry-aware SessionBuilder.** `OccupancyModel::setMirrored` / `setBoxMirrored` reflect the shadow occupancy model against the same plane via the same helper, so a recipe can aim at — and assert — mirror-created voxels without diverging from the live editor.
- **MUSHROOM session + committed asset.** A radially-symmetric cap+stem on two layers, authored one quadrant with X+Y mirror; exercises layer-add (`K`), visibility hide/show (`H`), and a save→reload (`Ctrl+O`) round-trip. Every stem/cap occupancy check reads a *mirror-created* cell, so the run fails if mirroring is broken.
- **Filed #2575.** A `-y`-face single click does not place its `-y` neighbour at the cardinal (yaw-0) camera — a pre-existing editor picking limitation (not the mirror code, which is proven symmetric). The cap grows via `-x`/`-z` clicks only, taking its y-depth from the Y mirror; documented in `docs/design/editor-authoring-friction.md` (M-2).

Does **not** close #766 — the ant, bird, and tree slices remain.

## Test plan

- [x] `python3 scripts/author-entity.py mushroom` → **20/20** GUI-ASSERT PASS, determinism gate byte-identical, `mushroom.vxs` committed
- [x] `python3 scripts/author-entity.py rock` → still byte-identical to master (symmetry-off path unchanged)
- [x] `python3 scripts/gui-verify.py IRVoxelEditor` → **14/14** (standing shot table — no regression)
- [x] `fleet-run IRShapeDebug --load-vxs assets/voxel/entities/mushroom.vxs` → RESULT=CLEAN, renders (macOS/Metal)

## Screenshots

Mushroom loaded in `IRShapeDebug --load-vxs` (the small red entity at the platform edge; closeup below):

scene

closeup


jakildev (commented):
## Review — editor: F-1.6 Part 2e — MUSHROOM + F-1.2 mirror-symmetry fix (#766)

**Verdict:** approve

Traced the fix end to end: `applyMirrors` genuinely had zero call sites before
this diff (confirmed via grep on the pre-PR tree), and the wiring in
`applyEdit`/`applyEditRaw` (`main.cpp:762-786`) is correct — bounds-checked
against `set.size_` before converting to a flat index, and defensively
re-checked against `set.voxels_.size()`. `mirrorCenterOffset` (`symmetry.hpp:22`)
correctly reflects `v -> size-1-v` for a centre-seated plane, matching the
existing (unmodified) `mirrorAxis` reflection math in `applyMirrors`.

Checked the toggle-on-seats-the-offset design (`main.cpp:2636-2677`) against
the save/load path in `scene_io.hpp` — `enableX_`/`offsetX_` are always
persisted and restored together as a pair, so there's no path where a loaded
scene can have `enable` and `offset` disagree. `g_mirrorScratch` is a
namespace-scope `std::vector` but lives inside the file's existing anonymous
namespace (`main.cpp:278-1396`), so it's the sanctioned "module-internal
state in a .cpp" pattern, not a header-global violation.

`OccupancyModel::setMirrored`/`setBoxMirrored` mirror the shadow model
against the identical `applyMirrors` + centre-plane helper the live editor
uses, so the session builder's aiming can't drift from editor reality — the
right design for keeping the two in lockstep.

Verified all `applyEdit`/`applyEditRaw` call sites (`main.cpp:891,934,990,1032,1116,1889,1988`)
still compile against the new signatures — no stray direct `applyEditRaw`
caller bypasses the mirror fan-out. `clang-format --dry-run` is clean on all
four touched C++ files.

The mushroom session's assertions are a well-targeted positive-fire test:
every stem/cap check reads a cell the recipe only ever clicked as its mirror
image, so a broken `applyMirrors` wiring would fail the run rather than
silently pass. Filing #2575 for the `-y`-face picking limitation instead of
scope-creeping this PR into fixing it is the right call — it's a pre-existing
picking issue unrelated to the mirror fix, and the recipe correctly routes
around it via the Y mirror instead of depending on it.

### Praise
- Clean choke-point design: every edit path (click, drag, fill, SDF bake,
  erase) already funnels through `applyEdit`, so wiring the mirror fan-out
  there covers all of them from one place with no per-call-site churn.
- `docs/design/editor-authoring-friction.md` update captures the M-1/M-2
  findings (why the offset defaulted to an unusable value, and the `-y`-face
  picking limitation) with enough detail that a future author won't
  re-discover the same friction from scratch.

No blockers, no needs-fix, no nits.

- [x] Already covered by the PR's own test plan (author-entity mushroom/rock,
  gui-verify, IRShapeDebug load) — nothing outstanding.

PR #2580, opened 2026-07-28
render: make the GLSL include resolver recursive (visited-set) to match Metal's (#2514)

### LINKED ISSUE #2514: render: make the GLSL include resolver recursive (visited-set) to match Metal's

jakildev:

Closes #2514


jakildev:
Releasing without touching the branch — this task edits engine/render/src/opengl/opengl_shader.cpp, which engine/render/CMakeLists.txt only compiles on non-Darwin hosts (IR_isDarwin gate, OPENGL_SRC vs METAL_SRC). Picked up on a macOS pool worker; the scout's needs_gl_host:false cache flag is a false negative here (same as noted on this issue's prior release). No real work done — just an empty claim commit. — worker

PR #2595, opened 2026-07-29
render: guard the light-volume global-buffer single-canvas assumption (#2341)

### LINKED ISSUE #2341: render: light-volume global buffers (LightSourceBuffer/LightVolumeParams) go stale across canvases with 2+ populated C_CanvasLightVolume
**Area:** engine/render
**Model:** opus
**Blocked by:** (none)

**Context:** Flagged by the Opus recheck on PR #2337 (spot-cone shaping, #2318).
`COMPUTE_LIGHT_VOLUME.update()` runs per canvas and re-uploads the single
global `LightSourceBuffer` SSBO and `LightVolumeParamsBuffer` UBO each
iteration, so after the system finishes both hold only the **last-processed**
canvas's light list and its has-SPOT flag (`worldOriginVoxel_.w`, #2318).
`LIGHTING_TO_TRIXEL` then, per canvas, binds each canvas's own ID volume
(image unit 7) but reads the has-SPOT gate and indexes `lights[winId-1]`
from those **global** buffers — i.e. from the last canvas's data, not the
canvas actually being lit.

`.xyz` (`cameraAnchorVoxel()`) is safe because it's identical for every
canvas in a frame. PR #2318 is the first per-canvas-*varying* data threaded
through these global buffers, so the "last canvas wins" property becomes
semantically load-bearing for the first time.

**Concretely:** with 2+ rendered `C_CanvasLightVolume` canvases where a
SPOT-bearing canvas isn't processed last, the non-last canvas either (a)
skips the cone entirely (global `.w` came from a no-spot last canvas →
spot renders as an omni sphere), or (b) applies the wrong light's
axis/aperture/apex (`winId` indexes the last canvas's list). Worst case is
a visual miss, never a crash — `LightSourceBuffer` is always sized to
`kLightVolumeMaxSources` (256), and the `originAndType.w ==
kLightTypeSpot` check on a stale entry usually fails and falls back to
omni.

**Why this didn't block #2337:** no current scene hits it. Every creation
attaches exactly one `C_CanvasLightVolume` (to `mainCanvas`); the
per-axis path shares the main canvas's volume+params (consistent); the one
deliberate multi-volume demo (`main_per_canvas_scope.cpp`, #363) is
emissive-only with a non-rendered sentinel B canvas. Per-axis and
single-canvas correctness are intact — but per-canvas light **scope**
(#363) is a supported, unit-tested feature area, so a future author adding
a spot light to a multi-canvas scene would silently get wrong cones.

**Acceptance criteria:** either (a) a short comment + `IR_ASSERT`
documenting the single-light-volume-canvas assumption in
`LIGHTING_TO_TRIXEL`, guarding against silent misuse, or (b) make the
has-SPOT flag / light-list genuinely per-canvas (re-upload per canvas or a
per-canvas buffer) so multi-canvas light-volume scenes with SPOT lights
render correctly. The architect should pick the scope — see the Opus
recheck comment on PR #2337 for the full analysis.

jakildev:
## Summary
- `COMPUTE_LIGHT_VOLUME` re-uploads the shared `LightSourceBuffer` / `LightVolumeParamsBuffer` **per canvas**, so after its tick they describe only the last-processed canvas. `LIGHTING_TO_TRIXEL` reads the has-SPOT gate and `lights[winId-1]` from those globals **per canvas** — so 2+ rendered `C_CanvasLightVolume` canvases where any seeds a SPOT light the non-last canvases from the wrong data (wrong cone axis/aperture, or spot → omni).
- Implements **option (a)** from the issue (the plan's choice): make the load-bearing assumption explicit and loud, rather than build per-canvas storage for a capability with zero current consumers.
- Adds `detail::lightVolumeGlobalBufferSafe(processedCanvases, anySpot)` (pure predicate), per-frame counters wired via `beginTick`/`endTick`, and an `IR_ASSERT` whose message names the fix. Guard lives in the **producer** because the has-SPOT flag and processed-canvas count exist CPU-side only there.
- Documenting comment at `LIGHTING_TO_TRIXEL`'s global light-volume binds, cross-referencing the guard.

**Render output is byte-identical** — no shader, GPU-buffer, dispatch, or pipeline-state change; `IR_ASSERT` is stripped under `IR_RELEASE`. Hence no screenshots / `render-verify` (a before/after capture would be identical by construction), and `optimize` was skipped: the per-frame cost is one `int` increment plus one bool OR on an O(canvases ≤ 3) loop already dominated by its GPU dispatch chain.

## Correction to the plan's premise (worth a reviewer's eye)
The plan's "Verified current state" asserts #363's `main_per_canvas_scope.cpp` sentinel canvas B is *not counted* because `useCameraPositionIso_ == false`. **That is wrong** — `C_TrixelCanvasRenderBehavior`'s default ctor sets `useCameraPositionIso_ = true`, so that scene genuinely reaches **2 processed canvases**. It stays under the guard because it is emissive-only (`anySpot = false`), not because the sentinel is skipped. Behavior is unchanged either way, but the comments now state the real mechanism, and the demo is one SPOT light away from legitimately tripping the guard — which is the intended semantics.

## Test plan
- [x] `IrredenEngineTest` — full suite green (**1363/1363**), including 2 new `LightVolumeGlobalBufferGuardTest` cases: the positive fire (`(2,true)`, `(3,true)` → false) plus positive controls (`(1,true)`, `(2,false)`, `(0,false)` → true) so the guard can't abort a config that renders correctly today.
- [x] `IRShapeDebug` builds clean.
- [x] `IRLightingPerCanvasScope` (the 2-canvas #363 scene) `--auto-screenshot` → `exit=0`, clean shutdown, **no assert**.
- [x] `IRLightingSpot` (single canvas + SPOT) `--auto-screenshot` → `exit=0`, **no assert**.
- [x] **Positive control that the guard is actually live** (a passing pure-predicate test alone proves nothing about wiring): temporarily relaxed the predicate to fire on canvas count alone, rebuilt, and `IRLightingPerCanvasScope` aborted with the #2341 message (`exit=134`) — proving `endTick` is wired by `registerSystem` and the counter really reaches 2. Reverted; the demo returns to `exit=0` with zero asserts.

## Acceptance evidence
Criteria from `.fleet/plans/issue-2341.md` § "Acceptance criteria" (committed as this PR's first commit). Every check below was re-run on the **final tree** at `b1e448f0`, macOS/Metal host — not transcribed from an earlier iteration.

|---|---|---|
| `EXPECT_FALSE(lightVolumeGlobalBufferSafe(2, /*anySpot=*/true))` — **the positive fire** | `./build/IrredenEngineTest --gtest_filter='LightVolumeGlobalBufferGuardTest*'` (assertion at `test/render/per_canvas_light_scope_test.cpp:332`) | `[ OK ] LightVolumeGlobalBufferGuardTest.TripsForMultiCanvasWithSpot (0 ms)` — predicate returned `false`, i.e. the guard trips |
| `EXPECT_FALSE(lightVolumeGlobalBufferSafe(3, true))` | same case, assertion at `per_canvas_light_scope_test.cpp:333` | same `[ OK ]` line — both `EXPECT_FALSE`s in the case held |
| `EXPECT_TRUE(lightVolumeGlobalBufferSafe(1, true))` — single canvas + spot stays silent | same command (assertion at `per_canvas_light_scope_test.cpp:340`) | `[ OK ] LightVolumeGlobalBufferGuardTest.PermitsConfigurationsCorrectToday (0 ms)` |
| `EXPECT_TRUE(lightVolumeGlobalBufferSafe(2, false))` — #363's emissive-only 2-canvas scene stays silent | same case, assertion at `per_canvas_light_scope_test.cpp:347` | same `[ OK ]` line |
| `EXPECT_TRUE(lightVolumeGlobalBufferSafe(0, false))` | same case, assertion at `per_canvas_light_scope_test.cpp:350` | same `[ OK ]` line |
| Build clean; `per_canvas_light_scope_test` suite stays green; `IRShapeDebug` builds; no render-verify needed | `fleet-build --target IrredenEngineTest`; `./build/IrredenEngineTest`; `./build/IrredenEngineTest --gtest_filter='*PerCanvasLight*:*LightVolume*'`; `fleet-build --target IRShapeDebug` | `[100%] Built target IrredenEngineTest` / `[100%] Built target IRShapeDebug`, no diff-attributable warnings; `1363 tests from 244 test suites ran. [ PASSED ] 1363 tests.`; the test file's two suites → `10 tests from 2 test suites ran. [ PASSED ] 10 tests.` No render-verify: the diff is a debug-only `IR_ASSERT` + counters + comments, so a before/after capture is identical by construction |

All six criteria met. The one criterion the predicate tests cannot cover on their own — that `endTick` is actually wired and the counter reaches 2 — is discharged by the live-abort probe in the Test plan above (relax predicate → `exit=134` with the #2341 message → revert → `exit=0`).

## Note for the reviewer
Not applied, to stay inside the plan's 3-file scope: `engine/prefabs/irreden/render/CLAUDE.md` has no §Gotchas entry for "at most one rendered `C_CanvasLightVolume` canvas may seed a SPOT". The assert message is self-documenting, so I reported rather than expanded scope — happy to add a one-liner if you'd prefer it documented there.

Closes #2341

### DIFF
--- a/.fleet/plans/issue-2341.md
+++ b/.fleet/plans/issue-2341.md
@@ -1,0 +1,199 @@
+<!--
+Plan file for #2341 (per #1932 — committed as the first commit of the
+implementer's PR). This is the `## Plan` comment posted on issue #2341 on
+2026-07-20. Reproduced verbatim below; the issue thread is the canonical
+source.
+-->
+
+## Plan: light-volume global buffers go stale across canvases with 2+ populated C_CanvasLightVolume
+
+- **Issue:** #2341
+- **Model:** sonnet — the design judgment (option a vs b, guard location, the exact predicate, the positive-fire test shape) is resolved in this plan; implementation is a bounded, CPU-side diagnostic change (begin/endTick counters + one `IR_ASSERT` + a pure predicate + comments + a unit test). No shader, GPU-buffer, dispatch, or pipeline-state changes — render output stays byte-identical — so it does not carry opus-tier render judgment despite touching render system headers.
+- **Date:** 2026-07-20
+
+### Scope
+
+Pick **option (a)** from the acceptance criteria: make the load-bearing
+"single light-volume canvas" assumption of the spot-cone path **explicit and
+guarded**, rather than building the full per-canvas light-list/params storage
+(option b). The full fix is deferred — it should be filed as a follow-up only
+when a real multi-canvas-SPOT consumer exists to validate it against (today
+there are zero; see "Why (a), not (b)" below).
+
+The deliverable: a debug `IR_ASSERT` that fires the moment a scene threads
+per-canvas-**varying** spot data through the shared global buffers for 2+
+rendered light-volume canvases, plus a documenting comment at the consumer's
+global binds, plus a pure-CPU unit test that positively fires the guard.
+
+### Verified current state (mechanism confirmed against code, not asserted)
+
+Read both systems in full at current master (`5c53d2e5`). The staleness is real
+and exactly as the Opus recheck on PR #2337 described:
+
+- **Producer** `System<COMPUTE_LIGHT_VOLUME>::tick` (`system_compute_light_volume.hpp:375`)
+  runs **per canvas** and re-uploads the single global named buffers every
+  iteration: `lightSourceBuf_->subData(...)` (line 441) and
+  `paramsBuf_->subData(...)` (line 443). Both `LightSourceBuffer` (SSBO) and
+  `LightVolumeParamsBuffer` (UBO) are created once in `create()`
+  (lines 551–566) as **global** resources. `params_.worldOriginVoxel_.w` is the
+  per-canvas **has-SPOT** flag (line 422–423); `params_.lightCount_` and the
+  light list are per-canvas. After the system finishes all canvases, the globals
+  hold only the **last-processed** canvas's data.
+- The two systems are **not** interleaved — each pipeline system iterates all
+  its matched canvases before the next runs, so `COMPUTE_LIGHT_VOLUME` finishes
+  every canvas before `LIGHTING_TO_TRIXEL` starts.
+- **Consumer** `System<LIGHTING_TO_TRIXEL>::tick` (`system_lighting_to_trixel.hpp:145`)
+  binds each canvas's **own** volume texture (slot 5, line 208) and **own**
+  winning-light ID texture (image unit 7, line 213), but binds the **global**
+  `lightVolumeParamsBuf_` (line 226) and **global** `lightSourceBuf_` (line 231).
+  On the has-SPOT path it reads `worldOriginVoxel.w` and indexes `lights[winId-1]`
+  from those globals — i.e. from the last canvas's data, not the canvas being lit.
+- `.xyz` (`cameraAnchorVoxel()`) is canvas-invariant within a frame, so it was
+  always safe; `.w`/light-list are the **first** per-canvas-varying data threaded
+  through the globals (#2318), which is why the "last canvas wins" property only
+  became load-bearing then.
+- **No shipping scene hits it**: every creation attaches exactly one
+  `C_CanvasLightVolume` (on `mainCanvas`); the per-axis path shares the main
+  canvas's volume/params inside the main tick (`dispatchPerAxisLighting`, not a
+  separate `C_CanvasLightVolume` entity); and `main_per_canvas_scope.cpp` (#363)
+  is emissive-only with a **non-rendered** sentinel B canvas
+  (`useCameraPositionIso_ == false`, so `COMPUTE_LIGHT_VOLUME::tick` early-returns
+  at line 380 and never counts it). Worst case is a visual miss (spot → omni),
+  **never a crash** (`LightSourceBuffer` is always sized to 256).
+
+### Why (a), not (b)
+
+The Opus recheck (PR #2337, 2026-07-09) recommended **(a) a comment + `IR_ASSERT`**
+as the primary path and framed (b) per-canvas buffers as a heavier "tracked
+follow-up." Option (b) would add a ~20 KiB per-canvas `LightSourceBuffer` + a
+per-canvas params UBO on `C_CanvasLightVolume` and re-bind the canvas's own
+buffers in the consumer — real storage + code for a capability
+(multi-light-volume-canvas scenes with SPOT lights) that has **zero** current
+consumers, and whose only positive-fire acceptance test would be a synthetic
+new demo scene. Per engine convention (assert the load-bearing invariant; defer
+generalization until a real consumer exists), (a) is the proportionate fix: it
+converts a silent wrong-cone into a loud debug abort, and the abort message is
+the trigger to do (b) — validated against the actual scene that needs it — when
+one appears.
+
+### Approach
+
+Guard **in the producer** (`COMPUTE_LIGHT_VOLUME`), with a cross-referencing
+comment in the consumer. Deviation-with-rationale from the recheck's "assert in
+`LIGHTING_TO_TRIXEL`" phrasing: the per-canvas has-SPOT flag and the
+processed-canvas count are available **CPU-side only in the producer** (the
+consumer would have to read the flag back off the GPU UBO). Putting the assert
+in the producer lets the guard be **precise** (`!(2+ canvases && any spot)`)
+instead of the conservative "≤1 light-volume canvas" a consumer-side count would
+force — and the producer is the system that actually clobbers the shared buffers.
+
+1. **`system_compute_light_volume.hpp` — add a pure predicate** in the existing
+   `IRSystem::detail` namespace (alongside `gatherLightSources`, so the existing
+   isolated gtest can call it the same way):
+   ```cpp
+   // The consumer (LIGHTING_TO_TRIXEL) reads the has-SPOT gate + light list from
+   // the GLOBAL LightVolumeParamsBuffer / LightSourceBuffer, which this system
+   // overwrites per canvas — after the tick they hold only the LAST processed
+   // canvas's data. That is simultaneously correct for every processed canvas
+   // only while at most one light-volume canvas is processed, OR no processed
+   // canvas seeds a SPOT (the only per-canvas-VARYING field the consumer reads;
+   // .xyz is the camera anchor, identical for all canvases). Returns false when
+   // the globals cannot be correct for all processed canvases at once — 2+
+   // processed canvases AND at least one SPOT. Multi-canvas SPOT scenes need
+   // per-canvas storage (issue #2341 deferred option b).
+   inline bool lightVolumeGlobalBufferSafe(int processedLightVolumeCanvases,
+                                           bool anyCanvasSeededSpot) {
+       return !(processedLightVolumeCanvases >= 2 && anyCanvasSeededSpot);
+   }
+   ```
+2. **`system_compute_light_volume.hpp` — wire per-frame counters** on
+   `System<COMPUTE_LIGHT_VOLUME>` (registerSystem auto-wires begin/endTick when
+   defined; the system currently has neither):
+   - Add members `int processedLightVolumeCanvases_ = 0;` and
+     `bool anyCanvasSeededSpot_ = false;`.
+   - Add `void beginTick() { processedLightVolumeCanvases_ = 0; anyCanvasSeededSpot_ = false; }`.
+   - In `tick`, **after** the `if (!behavior.useCameraPositionIso_) return;`
+     early-return (line 380–381) so non-rendered canvases like #363's sentinel B
+     are not counted: `++processedLightVolumeCanvases_;` and, inside/after the
+     upload block where the local `hasSpot` is computed (line 409–418),
+     `anyCanvasSeededSpot_ |= hasSpot;`.
+   - Add `void endTick()` with the guard:
+     ```cpp
+     IR_ASSERT(
+         detail::lightVolumeGlobalBufferSafe(processedLightVolumeCanvases_, anyCanvasSeededSpot_),
+         "COMPUTE_LIGHT_VOLUME: global LightSourceBuffer/LightVolumeParams hold only the "
+         "last processed canvas's spot data; a scene with 2+ rendered C_CanvasLightVolume "
+         "canvases where any seeds a SPOT renders wrong/omni cones (issue #2341). Make the "
+         "light list / has-SPOT per-canvas (option b) to lift this assumption."
+     );
+     ```
+3. **`system_lighting_to_trixel.hpp` — documenting comment only** (no code
+   change) at the two global binds (lines 226 and 231): note the params/light-list
+   buffers are per-frame-correct only for the last-processed light-volume canvas,
+   so the spot-cone read here relies on the single-light-volume-canvas assumption
+   guarded in `COMPUTE_LIGHT_VOLUME::endTick`; multi-canvas + SPOT needs per-canvas
+   storage (#2341 option b).
+4. **`test/render/per_canvas_light_scope_test.cpp` — positive-fire unit test**
+   (existing gtest, runs without a RenderManager): assert the predicate.
+
+### Affected files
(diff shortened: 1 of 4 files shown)


jakildev (commented):
## Review — render: guard the light-volume global-buffer single-canvas assumption (#2341)

**Verdict:** approve

### Verification performed

Read both changed system headers and the test file in full (not just hunks),
then independently re-verified the PR's claims rather than trusting them:

- **Build clean.** `IrredenEngineTest` and `IRShapeDebug` both build with no
  warnings from the diff.
- **Full suite green, exact match to the PR's claim.** `./build/IrredenEngineTest`
  → `1363 tests from 244 test suites ran. [PASSED] 1363 tests.` — matches the
  body's `1363/1363` line exactly, plus the 2 new `LightVolumeGlobalBufferGuardTest`
  cases individually confirmed passing.
- **Both named demos clean.** `IRLightingPerCanvasScope --auto-screenshot` and
  `IRLightingSpot --auto-screenshot` both → `RESULT=CLEAN exit=0`, no assert.
- **Independently confirmed the guard is actually wired live**, not just declared:
  temporarily forced `lightVolumeGlobalBufferSafe` to `return false`
  unconditionally, rebuilt `IRShapeDebug`, and reproduced the exact `IR_ASSERT`
  abort with the authored message at `endTick(...)` — proves `registerSystem`
  really does auto-wire `beginTick`/`endTick` for this system (also confirmed
  via `HasBeginTickMember`/`HasEndTickMember` concepts in `ir_system.hpp:233,236`)
  and that the assert isn't dead. Reverted the probe; working tree is clean
  against the PR branch afterward.
- **Concurrency safety of the new non-atomic counters.** `processedLightVolumeCanvases_`
  / `anyCanvasSeededSpot_` are mutated per-canvas in `tick` with no synchronization.
  Confirmed `COMPUTE_LIGHT_VOLUME` is registered via plain `registerSystem<...>`
  with no `setConcurrency(PARALLEL_FOR)` call anywhere in the tree (grepped),
  so it defaults to `Concurrency::SERIAL` — no data race. Worth knowing if this
  system is ever opted into `PARALLEL_FOR` later, but not a defect today.
- **"Correction to the plan's premise" claim checked out.** `component_trixel_canvas_render_behavior.hpp:38-40`'s default ctor does pass
  `useCameraPositionIso = true` as the first arg — confirms the PR body's
  correction (canvas B in `main_per_canvas_scope.cpp:38` uses `C_TrixelCanvasRenderBehavior{}`,
  the default ctor, so it genuinely reaches 2 processed canvases and stays safe
  only because it's emissive-only).
- **`clang-format --dry-run --Werror`** clean on all three touched C++ files.
- Naming, ECS invariants, `System<N>` state placement, `void()` begin/endTick
  signatures, and the `detail` namespace placement of the new predicate all
  follow the established conventions in this file (mirrors `gatherLightSources`).
- No header-scope mutable globals introduced (`.claude/rules/cpp-globals.md`)
  — state lives correctly as `System<N>` members.

### Acceptance (issue #2341)

Dispatched the acceptance grader against `.fleet/plans/issue-2341.md`'s
Acceptance criteria (reproduced verbatim in the PR's first commit):

| Criterion | Grade |
|---|---|
| `EXPECT_FALSE(lightVolumeGlobalBufferSafe(2, true))` — positive fire | met |
| `EXPECT_FALSE(lightVolumeGlobalBufferSafe(3, true))` | met |
| `EXPECT_TRUE(lightVolumeGlobalBufferSafe(1, true))` | met |
| `EXPECT_TRUE(lightVolumeGlobalBufferSafe(2, false))` | met |
| `EXPECT_TRUE(lightVolumeGlobalBufferSafe(0, false))` | met |
| Build clean; suite stays green; no render-verify needed | met (independently re-verified above, not just graded from the diff) |

All criteria met.

### Nits

- PR body uses a `## Test plan` checklist rather than the `## Acceptance
  evidence` table (criterion \| check run \| observed) that
  `docs/agents/AUTHOR-PIPELINE.md` §"Acceptance evidence" asks for. Every
  criterion was still directly gradable from the diff/PR body content here,
  so this cost no round-trip — flagging per the "missing section but
  establishable from the diff = nit" rule, not as a merge blocker.

### Praise

- Genuinely proportionate fix: a loud debug-only assert + comment instead of
  building unused per-canvas storage for a capability with zero current
  consumers, with the abort message pointing at exactly what to do when a
  real consumer shows up.
- The plan's own "Gotchas" section pre-empted both subtleties that would
  otherwise have been review findings — counting after the early-return, and
  the exotic same-list-multi-canvas false-positive — which is why this review
  found nothing to add on either front.
- The PR body's self-correction of the plan's premise (canvas B genuinely
  counts, doesn't skip) is exactly the kind of thing that's easy to leave
  unstated; flagging it explicitly saved a reviewer round-trip.

- [x] Already covered — build, full suite (1363/1363), both named demos, and
  the assert-fires probe were all independently re-verified in this review.
  Nothing outstanding.

Opus recheck not required.

jakildev (commented):
## Opus recheck — render: guard the light-volume global-buffer single-canvas assumption (#2341)

Sonnet approved and wrote "Opus recheck not required". I picked it up anyway
under the standing rule that `engine/prefabs/irreden/render/systems/` +
render-pipeline invariants get an Opus pass regardless of the first-pass
verdict. **I confirm the approve.** The two nits below are follow-up material,
not merge dependencies.

### Confirming / extending Sonnet's pass

- **`beginTick` wiring — confirmed, but Sonnet's probe could not have proven
  it.** Sonnet forced `lightVolumeGlobalBufferSafe` to `return false` and
  observed the abort at `endTick`, concluding "`registerSystem` really does
  auto-wire `beginTick`/`endTick`". That probe only proves **`endTick`** fires;
  an unwired `beginTick` would leave the counters accumulating across frames.
  The evidence that actually settles it is elsewhere in the same review: the
  clean `IRLightingSpot` run (one canvas + a SPOT, many frames) — without the
  per-execution reset, `processedLightVolumeCanvases_` reaches 2 on frame 2 and
  the assert fires. I also read the runtime side: `SystemManager::executeSystem`
  brackets the archetype-node iteration with
  `m_beginTicks[system].functionBeginTick_()` (`system_manager.cpp:570`) and
  `m_endTicks[system].functionEndTick_()` (`system_manager.cpp:653`), and
  `insertBeginTickFunction`/`insertEndTickFunction`
  (`system_manager.hpp:482`, `:646`) add no wrapping beyond the call. Sonnet's
  conclusion holds; the reasoning behind it didn't.

- **The predicate is the *right* condition, not just a live one.** The claim
  that carries this whole PR is "has-SPOT + the light list are the only
  per-canvas-**varying** data the consumer reads from the globals" — if that is
  wrong, `lightVolumeGlobalBufferSafe(2, false) == true` is a hole rather than a
  positive control. Checked it at the shader, both sides:
  - Both consumer UBO blocks declare `gridSize` / `halfExtent` / `lightCount` /
    `stepFalloff` as `_`-prefixed layout tombstones and read only
    `lightVolumeWorldOrigin` (`c_lighting_to_trixel.glsl:123-129`,
    `c_light_overflow_faces.glsl:76-82`). So the per-canvas-varying
    `lightCount_` and `stepFalloff_` the producer uploads are genuinely
    unread by the consumer — they matter only to seed/propagate, which run
    inside the producer's own per-canvas span.
  - Both `lights[...]` reads sit **inside** an `if (lightVolumeWorldOrigin.w != 0)`
    gate (`c_lighting_to_trixel.glsl:335,340`;
    `c_light_overflow_faces.glsl:178,184`). Multi-canvas-without-SPOT therefore
    never indexes the stale list — the `(2, false)` positive control is grounded,
    not assumed.

- **Tally can't drift from what actually clobbered the buffers.** Both
  `++processedLightVolumeCanvases_` and `anyCanvasSeededSpot_ |= hasSpot` sit on
  the same unconditional path as the two `subData` uploads, past the only
  early-return in `tick`. The one asymmetry is benign: the light-SSBO `subData`
  is skipped at `count == 0` so the buffer retains a prior canvas's list, but
  `paramsBuf_` still uploads `.w = 0`, closing the consumer's gate — the
  retained list is never indexed.

- **No scene trips the guard today.** Swept every creation in the tree,
  including the private ones present on this host: all attach exactly one
  `C_CanvasLightVolume`. The sole two-canvas scene is
  `creations/demos/lighting/main_per_canvas_scope.cpp:38`, emissive-only. The
  PR's correction of the plan's premise about that canvas is right, and it does
  not change the outcome.

### Nits

1. **"Exotic" undersells the false-positive class — it is the *default* light
   scoping.** The plan's Gotchas call the identical-list case exotic. But an
   unparented `C_LightSource` is world-scope and gathered for **every** canvas
   (`component_light_source.hpp:25-27`; `nodeInScope` returns true on
   `nodeParent == kNullEntity`, `system_compute_light_volume.hpp:232-240`), and
   `volumeOriginVoxel` is the frame's camera anchor — canvas-invariant. So two
   rendered light-volume canvases plus one **unparented** SPOT upload
   byte-identical buffers on both passes and render correctly, yet abort in
   debug. That is the back-compat default scoping, not an exotic config, and the
   assert message's instruction ("Make the light list / has-SPOT per-canvas
   (option b)") is the wrong remedy for it — option (b) buys nothing when the
   lists are already identical. Worth either a clause in the message ("if every
   SPOT is world-scope the per-canvas uploads are identical and this is a
   conservative false positive") or a follow-up. If you file the follow-up: an
   exact-and-cheap tightening is available without option (b) — the producer
   already holds each canvas's staged list CPU-side in `lightStaging_`, so the
   guard can fire only when the staged list or `hasSpot` actually **differs**
   from the previous processed canvas, on an O(≤256) compare in a debug-only
   path. Not asking for it in this PR — the plan explicitly scoped it out.

2. **The cross-system audit undercounts the read-sites.** The plan states
   `LIGHTING_TO_TRIXEL` is "the sole read-site that is per-canvas-incorrect".
   Two shaders perform the `lights[winId - 1]` spot read:
   `c_lighting_to_trixel.glsl:340` and `c_light_overflow_faces.glsl:184` (the
   #2334 overflow relight, dispatched from `dispatchOverflowLighting` while the
   cell pass's global binds are still resident). This does **not** weaken the
   guard — it lives in the producer, so it is consumer-agnostic, and both
   read-sites carry the same `.w` gate — but the documenting comment added at
   the binds does not reach the overflow path, and "sole read-site" is
   inaccurate as an audit claim.

I agree with Sonnet's existing nit about the `## Acceptance evidence` table;
keeping it in scope for the nit cleanup.

### Not re-checked

Naming, header hygiene, `shared_ptr`/`unique_ptr`, per-entity `getComponent`,
`SystemName` enum coverage, `clang-format` — Sonnet's checklist covers these and
nothing in the diff contradicts them. Build, full suite (1363/1363), and both
demo runs are taken from Sonnet's pass (exact counts, named demos, and a
negative-control probe — credible). The concurrency question is a non-issue as
Sonnet framed it: the tally is only sound while this system stays
`Concurrency::SERIAL`, which it is today.

Escalation: none. Safe for merge; the two nits are follow-up material.

jakildev:
Addressed the nit (PR body only — **no code change, no new commits**; `fleet:approved` stands and the reviewed diff is byte-identical):

- Added the `## Acceptance evidence` table that `docs/agents/AUTHOR-PIPELINE.md` §"Acceptance evidence" requires (one row per criterion: criterion | check run | observed), keeping `## Test plan` alongside it — per `commit-and-push/procedures/pr-body.md`, the two answer different questions ("the code doesn't break" vs "the ticket's named criteria fired").
- Per that section's step 2 ("run each named check NOW, on the final tree") I **re-ran every criterion myself** on macOS/Metal at `b1e448f0` rather than transcribing the original numbers: `LightVolumeGlobalBufferGuardTest` 2/2 OK, full suite `1363 tests from 244 test suites ran. [ PASSED ] 1363 tests.`, the test file's two suites 10/10, and `IrredenEngineTest` + `IRShapeDebug` both built clean. All six criteria met — matches the review's grading.
- The table notes explicitly which criterion the pure-predicate tests *cannot* discharge on their own (that `endTick` is wired and the counter reaches 2) and points at the live-abort probe that does.

— worker (pool-4)

PR #2596, opened 2026-07-29
engine/system: give SystemId its own unreachable miss sentinel (#2540)

### LINKED ISSUE #2540: engine/system: SystemId 0 collides with the kNullEntity 'no such system' sentinel
**Area:** engine/system
**Model:** opus
**Blocked by:** (none)

## Acceptance criteria

- A `SystemId` returned for a real registered system can never compare equal to the "no such system" sentinel.
- `IRSystem::findSystem(SystemName)` (added in #2526) can distinguish "registered as the first system in the process" from "never registered".
- The two prefab handles built on it (`IRPrefab::VoxelTransform::allocator()`, `IRPrefab::JointTransform::system()`) keep their current graceful "unwired -> nullptr" behavior.
- Existing callers that compare a `SystemId` against `IREntity::kNullEntity` are migrated or proven unaffected; `IrredenEngineTest` stays green.

## Context

`IREntity::kNullEntity` is `0` (`engine/entity/include/irreden/entity/ir_entity_types.hpp`), and `SystemManager::m_nextSystemId` also starts at `0` (`engine/system/include/irreden/system/system_manager.hpp`) and is handed out with `m_nextSystemId++`. So **the first system created in a process gets `SystemId == 0 == kNullEntity`** — indistinguishable from the "no such system" sentinel.

Verified this session while implementing #2526:

- `SystemId m_nextSystemId = 0;` and `constexpr EntityId kNullEntity = 0;` — read directly.
- In `creations/demos/shape_debug/main.cpp`, `IRSystem::createSystem<IRSystem::LOD_UPDATE>()` is the first `createSystem` call, so `LOD_UPDATE` holds id 0 at runtime.
- The pre-existing comment in `system_update_voxel_positions_gpu.hpp` asserted the opposite — "`IREntity::kNullEntity` is the 'never wired' sentinel (system ids count up from 0, so they never collide)" — which is self-contradictory. #2526 removed that comment along with the global it documented.

**Why it is latent rather than live.** The two current consumers (`UPDATE_VOXEL_POSITIONS_GPU`, `UPDATE_JOINT_MATRICES`) are never the first system a creation registers, and a false "unwired" answer degrades gracefully (the voxel set stays CPU-direct) rather than crashing. So no shipped scene is affected today.

**Why it is worth fixing now.** #2526 turned this from a two-call-site private convention into **general public API**: `IRSystem::findSystem(SystemName)` is callable for any `SystemName`, and `findSystem(LOD_UPDATE)` returns `0` in shape_debug — a caller cannot tell that from "not registered". The next consumer that resolves an early-registered system gets a silent wrong answer.

**Fix options** (deliberately not chosen here — this is why the issue files without a plan):

1. Start `m_nextSystemId` at 1 and reserve 0. Smallest diff; needs an audit of anything that treats a `SystemId` as a dense index into `m_systemNames` / `m_ticks` / the parallel vectors, since those are `m_nextSystemId`-sized and indexed by id.
2. Add a dedicated `kInvalidSystemId` distinct from `kNullEntity` and migrate the sentinel comparisons.
3. Give `findSystem` an `std::optional<SystemId>` return and let the handles branch on `has_value()`.

Option 1 is the least invasive but interacts with the id-as-dense-index assumption, which is why this wants a planning pass rather than a drive-by fix.

jakildev:
## Summary

`SystemManager` hands `SystemId`s out from `m_nextSystemId` starting at 0, and
`IREntity::kNullEntity` is 0 — so the **first system registered in a process
has an id indistinguishable from "no such system"**. #2526 promoted that from a
private two-call-site convention to public API (`IRSystem::findSystem`), so the
next consumer resolving an early-registered system would get a silent wrong
answer.

Committed approach is the plan's **option 2**: a dedicated
`IRSystem::kNullSystemId = std::numeric_limits<SystemId>::max()` in
`ir_system_types.hpp`, with every `SystemId` sentinel site migrated onto it.

- `max()` is unreachable as a real id, and it **fails loudly** — indexing by the
  sentinel trips `SystemManager`'s `system < m_nextSystemId` range assert
  instead of silently aliasing slot 0. That assert now covers the params
  accessors (`getSystemParams` / `setSystemParams`) as well as
  `replaceSystemBody`, added in the nit pass below; it is debug-only
  (`IR_ASSERT` is stripped under `IR_RELEASE`), so in release the same misuse
  is an out-of-bounds access that crashes rather than reading a real system.
- Option 1 (reserve id 0) was rejected because `SystemId` is a **dense index**
  into `m_systemNames` / `m_ticks` / `m_timingAccum` and the lockstep
  access/cadence/params vectors; it would need a phantom slot 0 in each (and
  would corrupt `getSystemCount()`) or a `-1` shift at every index site.

Migrated sites: both miss returns (`findEngineSystem`, `findSystem`'s
null-manager arm), both prefab handles' `unwired -> nullptr` branch
(`VoxelTransform::allocator()`, `JointTransform::system()`), the culling
minimap's 4 optional domain-source defaults + 2 compares, and the stored
"not wired yet" defaults in `shape_debug`, `scene_reset`, `lua_widgets`, and
`lighting_demo_scene`.

**Latent, not live.** Neither current consumer is ever a creation's first
system, so no shipped scene changes behavior — see the Test plan.

Two notes against the plan as filed:

- The plan's affected-file list predates #2529 merging and missed
  `shape_debug/main.cpp`'s two `SystemId x{}` locals. They're the same
  category as the `lighting_demo_scene` globals (audit item 8) and are
  migrated here so the `kNullEntity`-shaped sentinel sites carry one
  convention. Scope note: that is **not** the same as "the tree has a single
  convention" — the Opus recheck found one surviving zero-as-`SystemId`
  sentinel at `lua_perf_grid/main_lua.cpp:358` that neither the plan's grep
  nor the review sweeps matched (it writes a bare `0` literal). It is
  deliberately not migrated here — swapping in `kNullSystemId` would turn a
  wrong-system tick into an out-of-bounds pipeline splice, so it needs an
  early-return / hard-fail at the call site instead. Filed as #2599.
- Plan step 6 said to check `engine/prefabs/irreden/render/CLAUDE.md:625-626`
  at impl time — checked; those rows name the resolution path, not the
  sentinel value, so no edit was needed.

## Test plan

- [x] **Negative control on the new regression test.** Temporarily set
  `kNullSystemId = 0` (the pre-fix value), rebuilt, and confirmed
  `FirstRegisteredSystemIsDistinguishableFromAMiss` **fails** with
  `actual: 0 vs 0`. The test is not vacuous.
- [x] `IrredenEngineTest` — **1362/1362 pass** (includes the new test, the two
  migrated miss expectations, and `voxel_bone_slot_seed_test`, which is the
  headless proof the handles' unwired→nullptr behavior is preserved).
- [x] `render-verify --target IRShapeDebug` — **24/24 checks PASS**, every shot
  at `match% 100.0` / `max_delta 0` against the committed `macos-debug`
  references. Byte-identical output.
- [x] Builds green: `IRShapeDebug`, `IRSceneReset`, `IRLuaWidgetsDemo`,
  `IRLightingHDR`, `IRLightingSunOrbit`, `IRVoxelEditor`.
- [x] Plan's sweep gate:
  `git grep -nE "kNullEntity" -- engine/system engine/prefabs creations test | grep -iE "systemid|findSystem"`
  returns empty.

`attach-screenshots` skipped deliberately: render-verify proves the output is
byte-identical, so a before/after pair would be two identical images.

## Opus-recheck nits addressed

- **Nit 1 (doc-vs-code drift) — fixed in code, the option the reviewer
  preferred.** The sentinel's doc comment claimed a range assert protected an
  accidental index-by-sentinel, but that assert lived only on
  `replaceSystemBody`, not on the `getSystemParams` path the two prefab
  handles and `lightGatherRecords` actually use. Added the same
  `system < m_nextSystemId` assert to `getSystemParams` and `setSystemParams`
  (`m_systemParams` is `emplace_back`-ed once per created system, so
  `size() == m_nextSystemId` and the bound matches `replaceSystemBody`'s), and
  tightened both doc sites (`ir_system_types.hpp`, `engine/system/CLAUDE.md`)
  to name the accessors and to state the debug-only caveat — the claim is now
  true in debug and honest about release.
- **Nit 2 (surviving `SystemId{0}` sentinel) — filed as #2599**, not migrated
  here, for the reason the reviewer gave. The PR body's "single convention"
  line is corrected above.

Checked before adding the asserts that no live path can trip them: both
prefab handles guard on `== kNullSystemId` and return `nullptr` first, every
other call site passes an id freshly returned by `registerSystem`, and the two
stored lighting handles are assigned in `initSystems()` before their only
unguarded read. `IRLightingCombined --auto-screenshot` exercises that read 6×
(one `DOMAIN-STATE` line per shot) and is clean.

## Note for the reviewer (out of scope here)

`.claude/rules/cpp-globals.md:71-72` and `.claude/skills/simplify/SKILL.md:361-362`
still list `g_jointMatrixSystem` / `g_allocatorSystem` under **live
deviations**, but #2529 deleted both symbols — the only remaining references in
the tree are those two docs plus the historical `issue-2526.md` plan. Not
bundled here: both are gated self-config, and it's unrelated cleanup.

Closes #2540

### DIFF
--- a/.fleet/plans/issue-2540.md
+++ b/.fleet/plans/issue-2540.md
@@ -1,0 +1,86 @@
+## Plan: engine/system: SystemId 0 collides with the kNullEntity 'no such system' sentinel
+
+- **Issue:** #2540
+- **Model:** opus — bounded and fully enumerated below, but it edits ECS system-core types, changes the public `ir_system.hpp` miss-value contract, and likely stacks on an open PR; not sonnet-mechanical.
+- **Date:** 2026-07-25
+
+### Verified current state
+
+All claims below re-verified this session by reading the code (file:line cited against the tree named in each item).
+
+- `SystemId` is `EntityId` (`std::uint64_t`) — `engine/system/include/irreden/system/ir_system_types.hpp:10`; `kNullEntity = 0` — `engine/entity/include/irreden/entity/ir_entity_types.hpp:34`; `m_nextSystemId = 0` handed out with `m_nextSystemId++` — `system_manager.hpp:325`, `system_manager.cpp:58`. So the first system registered in a process gets id `0 == kNullEntity`, exactly as the issue states.
+- **The entire fix surface lives on PR #2529** (`claude/2526-systemname-registry`, approved, unmerged at plan time). `findSystem` / `findEngineSystem` / `recordEngineSystemId` and both prefab-handle resolvers do **not** exist on master (`git grep findSystem origin/master` is empty). `findEngineSystem`'s own doc comment on that branch already cites #2540 as the fix for this caveat.
+- **`SystemId` is a dense index** into parallel vectors: `m_systemNames[id]`, `m_ticks[system]`, `m_timingAccum[id]`, plus the access/cadence/params vectors kept in lockstep by `createSystem`/`createSystemDynamic` (`system_manager.hpp:296,179,302`; the ctor range-for resets `m_timingAccum` wholesale). `getSystemCount()` returns `m_nextSystemId` and there is a range assert `system < m_nextSystemId` (`system_manager.cpp:221`). This is why option 1 (reserve id 0) is rejected — see Approach.
+- **Exhaustive sentinel-compare audit** (branch `claude/2526-systemname-registry`, whole tree greps on `findSystem`, `SystemId.*kNullEntity`, and `SystemId x = 0` forms). Sites where a `SystemId` participates in a `kNullEntity`/zero sentinel convention:
+  1. `engine/system/include/irreden/system/system_manager.hpp:342` — `findEngineSystem` miss return.
+  2. `engine/system/include/irreden/ir_system.hpp:468` — `findSystem` null-manager return.
+  3. `engine/prefabs/irreden/render/systems/system_update_voxel_positions_gpu.hpp:311` — `IRPrefab::VoxelTransform::allocator()` compare.
+  4. `engine/prefabs/irreden/render/systems/system_update_joint_matrices.hpp:420` — `IRPrefab::JointTransform::system()` compare.
+  5. `engine/prefabs/irreden/render/systems/system_debug_culling_minimap.hpp:229,230,242,243` (member + params-struct defaults) and `:352,385` (compares) — "unwired" sentinel on `SystemId` members.
+  6. `creations/demos/lua_widgets/main_lua.cpp:57` — `g_dispatchId = kNullEntity` default (assigned at :204, passed at :230, never compared).
+  7. `creations/demos/scene_reset/main.cpp:93-97` — five `SystemId … = 0;` defaults (all unconditionally assigned at :147-152 before use, never compared).
+  8. `creations/demos/lighting/common/lighting_demo_scene.hpp:163,168` — `{}`-value-initialized `SystemId` globals (assigned during scene setup at :583,:588 before the reads at :492/:623-624).
+  9. `test/system/register_system_test.cpp:174,187` — `EXPECT_EQ(findSystem(...), IREntity::kNullEntity)` miss expectations.
+  All other `kNullEntity` comparisons under `engine/` and `creations/` are `EntityId`-typed (modifier resolve, hover detect, canvas/camera/fog/gizmo entities — checked) and are out of scope.
+- **Lua boundary**: `SystemId`s cross into Lua as `lua_Integer` (`IRSystem.registerSystem` return, `IRSystem.systemId(name)`), but the Lua-side miss contract is **raise a Lua error**, not return-a-sentinel (`engine/script/CLAUDE.md:810-813`). No Lua code compares against a numeric sentinel, so the new sentinel never needs to cross the boundary.
+
+No phase-0 probe is needed: every premise above is a static structural fact verified by reading the source, not a runtime mechanism claim.
+
+### Scope
+
+Give `SystemId` its own "no such system" sentinel that no real id can ever equal, migrate every enumerated sentinel site to it, and remove the first-system-is-indistinguishable caveat from `findSystem`. One task, one PR; no split.
+
+### Approach
+
+**Committed approach: option 2 — a dedicated `kNullSystemId`, value `std::numeric_limits<SystemId>::max()`.**
+
+Why not the alternatives:
+- *Option 1 (start `m_nextSystemId` at 1)*: fights the id-as-dense-index invariant. Either every parallel vector needs a phantom slot 0 (corrupting `getSystemCount()` and every whole-vector range-for), or every index site needs a `-1` shift. Larger blast radius and a permanent invariant tax for zero extra benefit.
+- *Option 3 (`std::optional<SystemId>` return)*: changes the call-site shape of a query API, breaks the established sentinel idiom (`kNullEntity`, `kVoxelTransformStatic`), and cannot express the "unwired" default on stored `SystemId` members (audit items 5–8) without wider churn.
+
+`max()` as the value has a robustness bonus: an accidental deref of the sentinel (e.g. `m_ticks[kNullSystemId]`) now trips the existing `system < m_nextSystemId` range assert loudly in debug instead of silently reading slot 0.
+
+Steps, in order:
+
+1. **Base resolution.** If PR #2529 has merged, branch from `origin/master` normally. If it is still open, claim with `--stackable-on 2529` and branch from `origin/claude/2526-systemname-registry`'s head — every file below refers to the post-#2529 tree. If #2529 merges mid-implementation, `git rebase --onto origin/master <fork-point>` per the inherited-prefix drop.
+2. **Define the sentinel** in `engine/system/include/irreden/system/ir_system_types.hpp`, directly after `using SystemId = EntityId;`:
+   `constexpr SystemId kNullSystemId = std::numeric_limits<SystemId>::max();`
+   (add `#include <limits>`). Doc comment: why it is not `kNullEntity` (id 0 is a valid `SystemId` — the first system registered), and that it must never cross the Lua boundary (see Gotchas).
+3. **Migrate the miss returns**: `findEngineSystem` (`system_manager.hpp:342`) and `findSystem`'s null-manager arm (`ir_system.hpp:468`). Delete `findEngineSystem`'s caveat paragraph (it cites #2540 as this fix) and update both doc comments.
+4. **Migrate the compares and defaults**: audit items 3–8 above — the two prefab-handle resolvers, the four minimap defaults + two compares, `g_dispatchId`, the five `scene_reset` zero-defaults, and the two `lighting_demo_scene` `{}`-inits. Items 6–8 are never compared today (proven above), but migrating the defaults makes "not yet assigned" loud (range assert) instead of silently aliasing slot 0, and leaves the tree with a single convention.
+5. **Update tests** (`test/system/register_system_test.cpp`): swap the `:174`/`:187` miss expectations to `kNullSystemId`, and add the positive-fire regression test (Acceptance criteria).
+6. **Update docs**: `engine/system/CLAUDE.md` §`findSystem` (`:103` miss value, `:119` caveat bullet — replace with a pointer to `kNullSystemId`). Check `engine/prefabs/irreden/render/CLAUDE.md:625-626` migration-table rows at impl time — they name the resolution path but not the sentinel value, so likely no edit. Do **not** edit `.fleet/plans/issue-2526.md` (historical record).
+7. **Sweep gate**: `git grep -nE "kNullEntity" -- engine/system engine/prefabs creations test | grep -i system` must come back empty of `SystemId`-typed hits before the PR opens.
+
+### Affected files
+
+- `engine/system/include/irreden/system/ir_system_types.hpp` — add `kNullSystemId` + `<limits>` include + doc comment
+- `engine/system/include/irreden/system/system_manager.hpp` — `findEngineSystem` miss return; drop the caveat paragraph
+- `engine/system/include/irreden/ir_system.hpp` — `findSystem` null-manager return; doc comment
+- `engine/prefabs/irreden/render/systems/system_update_voxel_positions_gpu.hpp` — `allocator()` compare + doc
+- `engine/prefabs/irreden/render/systems/system_update_joint_matrices.hpp` — `system()` compare + doc
+- `engine/prefabs/irreden/render/systems/system_debug_culling_minimap.hpp` — 4 defaults + 2 compares
+- `creations/demos/lua_widgets/main_lua.cpp` — `g_dispatchId` default
+- `creations/demos/scene_reset/main.cpp` — 5 zero-defaults
+- `creations/demos/lighting/common/lighting_demo_scene.hpp` — 2 `{}`-init globals
+- `test/system/register_system_test.cpp` — 2 expectation swaps + 1 new test
+- `engine/system/CLAUDE.md` — miss value + caveat bullet
+
+### Acceptance criteria
+
+- **Positive-fire regression test** (new, in `register_system_test.cpp`): in a fresh fixture, create `TEST_REGISTER_SYSTEM_A` as the **first** system and never create `TEST_REGISTER_SYSTEM_B`. Assert:
+  - `ASSERT_EQ(sysA, 0u)` — documents that the collision premise (first id is 0) still holds; if `SystemManager` ever pre-registers systems this flags the premise change;
+  - `EXPECT_EQ(IRSystem::findSystem(TEST_REGISTER_SYSTEM_A), sysA)`;
+  - `EXPECT_NE(IRSystem::findSystem(TEST_REGISTER_SYSTEM_A), IRSystem::findSystem(TEST_REGISTER_SYSTEM_B))` — **fails before this fix** (both sides are 0) and passes after: the observable "registered-first vs never-registered are distinguishable" check.
+- Updated miss expectations: `FindSystemReportsANameThatWasNeverCreated` and `FindSystemNoManagerTest` expect `kNullSystemId`; the latter (plus the headless `voxel_bone_slot_seed_test` path) proves the handles' "unwired → nullptr" behavior is preserved.
+- `IrredenEngineTest` green; `fleet-build --target IRShapeDebug` green (compiles the touched prefab headers + shape_debug); the touched demos (`scene_reset`, `lua_widgets`, lighting demos) still build.
+- The step-7 sweep gate returns no `SystemId`-typed `kNullEntity` hits.
+
+### Gotchas
+
+- **Never pass `kNullSystemId` across the Lua boundary.** LuaJIT numbers are doubles; `2^64-1` is not exactly representable. The Lua miss contract is already "raise a Lua error" (`IRSystem.systemId`) — keep it that way; the sentinel stays C++-side.
+- **`IR_RELEASE` strips `IR_ASSERT`**, so an unguarded `getSystemParams(kNullSystemId)` in release is an out-of-bounds vector read. Same failure class as today's silent slot-0 read, not worse — but always sentinel-check before indexing (every migrated site already does).
+- **`recordEngineSystemId` double-report is legitimate** (nested `createSystem<N>` → `System<N>::create()` → `registerSystem<N,...>` both record the same id) — do not "fix" the try_emplace/assert dance while in there.
+- The two current handle consumers degrade gracefully on a false miss, so nothing visual changes in any demo — this is a correctness fix with test-level evidence only; no screenshots or render-verify needed.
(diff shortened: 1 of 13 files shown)


jakildev (commented):
## Review — engine/system: give SystemId its own unreachable miss sentinel (#2540)

**Verdict:** approve

Clean, well-scoped sentinel migration. `IRSystem::kNullSystemId =
std::numeric_limits<SystemId>::max()` is a namespace-scope `constexpr` in a
module header (`ir_system_types.hpp`) — compliant with
`.claude/rules/cpp-globals.md` (constexpr/const namespace-scope values are
program constants, not state, and are explicitly allowed).

Verified independently rather than just trusting the PR body:

- **Sweep gate** — `git grep -nE "kNullEntity" -- engine/system engine/prefabs
  creations test | grep -iE "systemid|findSystem"` returns empty, confirming
  every `SystemId`-typed sentinel site is migrated (both miss returns
  `findEngineSystem`/`findSystem`, both prefab handles' unwired→nullptr
  branch, the minimap's 4 defaults + 2 compares, and the four demos'
  now-consistent `kNullSystemId` defaults).
- **Build** — `fleet-build --target IrredenEngineTest` is green (macOS,
  Metal backend).
- **Tests** — ran the touched suites directly:
  `--gtest_filter='*RegisterSystemTest*:*FindSystemNoManagerTest*'` → 9/9
  pass, including the new `FirstRegisteredSystemIsDistinguishableFromAMiss`
  positive-fire regression and the null-manager guard
  (`findSystem`'s null `g_systemManager` branch is correctly checked before
  dereferencing, matching the doc comment's stated contract for headless
  test contexts).
- **`engine/system/CLAUDE.md`** contract section is updated in the same PR
  to describe `kNullSystemId` instead of the retired caveat — docs and code
  agree.

The "latent, not live" framing is well supported: every migrated site is
either a genuine miss-check (never triggered because current consumers are
never a creation's first system) or an unconditionally-assigned default
that's overwritten before any comparison (scene_reset, lua_widgets,
lighting_demo_scene). `IR_RELEASE`'s stripped-assert risk on an
unguarded sentinel index is called out accurately in the plan and every
migrated comparison site guards before indexing.

### Praise
- The negative-control test-plan step (temporarily reverting
  `kNullSystemId` to 0, confirming the new regression test fails) is exactly
  the kind of evidence that proves a test isn't vacuous — good practice,
  worth repeating.
- Choosing `max()` over reserving id 0 is the right call given `SystemId` is
  a dense index into multiple parallel vectors (`m_systemNames`, `m_ticks`,
  `m_timingAccum`, the access/cadence vectors) — reserving 0 would need a
  phantom slot in every one of them or a `-1` shift at every index site.

- [x] None additional — build + targeted tests independently reproduced above.

Opus recheck required: touches `engine/system/` (SystemManager sentinel
semantics + the public `ir_system.hpp` `findSystem` surface).

jakildev (commented):
## Opus recheck — engine/system: give SystemId its own unreachable miss sentinel (#2540)

Sonnet approved and escalated for `engine/system/` + the public
`ir_system.hpp` `findSystem` surface. I confirm the approve. The design call
(`max()` over reserving id 0) is right, and the migration is complete. Two
nits below, neither of which blocks the merge.

### Confirming / extending Sonnet's pass

- **Sweep gate — confirmed, but Sonnet's grep shape could not have proven
  it.** The filter `... | grep -iE "systemid|findSystem"` only matches when
  the type name lands on the *same line* as the sentinel, so a `SystemId`
  variable whose name doesn't contain "systemid" (e.g. `g_dispatchId`) would
  have slipped through. I re-ran the sweep without that filter and against
  the explicit head SHA `48f6bfef`, and it is genuinely clean: zero
  `SystemId`-typed `kNullEntity` survivors across `engine/`, `creations/`,
  `test/`. Sonnet's conclusion holds; the evidence behind it was weaker than
  stated.
- **Sonnet said "every migrated comparison site guards before indexing" — I
  confirm** for all four (`JointTransform::system()`,
  `VoxelTransform::allocator()`, and the minimap's two domain checks at
  `system_debug_culling_minimap.hpp:352,385`).
- **Ordering safety of the migrated `0` → `kNullSystemId` defaults — checked,
  safe.** This is the one place the migration could have converted a silently
  in-bounds read into an out-of-bounds one, so I traced each:
  - `shape_debug/main.cpp:626-627` — both assigned *inside* the braced-init-list,
    which is guaranteed left-to-right sequenced, and both writes precede the
    `DEBUG_CULLING_MINIMAP::create()` element that reads them. Safe.
  - `lighting_demo_scene.hpp:163,168` — assigned in `initSystems()`, the shared
    setup every lighting demo calls. The only *unguarded* read is
    `logDomainState:492` → `IRSystem::lightGatherRecords(...)`, and that is an
    `onCaptureFrame_` hook that fires on a settled capture frame, long after
    init. Safe.
  - `scene_reset` / `lua_widgets` — unconditionally assigned before use, as the
    plan claims.
- **Lua-boundary claim — independently verified true.** The new doc comment
  asserts `kNullSystemId` never crosses into Lua. It doesn't:
  `IRSystem.systemId` (`lua_pipeline_bindings.hpp:224`) resolves from its own
  `prefabSystemIds` map and `throw`s a `sol::error` on miss — it never calls
  `findSystem`, so the sentinel has no path to a `lua_Integer`. Worth stating
  explicitly since a `max()` sentinel reaching LuaJIT would not round-trip
  through a double.

### Nits

1. **The new doc comment's stated safety mechanism doesn't exist on the
   accessor path it matters for.** `ir_system_types.hpp` says a sentinel
   "mistakenly used as an index trips the `system < m_nextSystemId` range
   assert instead of aliasing slot 0", and `engine/system/CLAUDE.md` repeats
   it. That assert lives in `SystemManager::replaceSystemBody`
   (`system_manager.cpp:219`) — *not* on the `getSystemParams` path that the
   two migrated prefab handles and `lightGatherRecords` actually use.
   `SystemManager::getSystemParams` (`system_manager.hpp:191-194`) does a bare
   `m_systemParams[system]` on a `std::vector`, and the `ir_system.hpp:619`
   free function delegates without a guard. So an accidental
   `getSystemParams(kNullSystemId)` is an unchecked OOB `operator[]` at index
   2^64-1 in debug *and* release, not a diagnosed assert.

   No live path reaches it (every migrated compare guards first), so this is
   documentation-vs-code drift rather than a bug — but it's drift in the two
   files that are now the canonical explanation of the sentinel, which is
   exactly where it will mislead the next reader. Either soften the comment
   to "an out-of-bounds index that crashes rather than silently aliasing slot
   0", or add the range assert to `getSystemParams` (and `setSystemParams`,
   same unchecked index at `:188`) and make the claim true. The second is the
   more useful fix and is a two-line change.

2. **One zero-as-`SystemId`-sentinel site survives the sweep.**
   `creations/demos/lua_perf_grid/main_lua.cpp:358` returns a fabricated
   `IRSystem::SystemId{0}` on the EVAL-resolution error path, and that value
   is spliced straight into the UPDATE pipeline at `:436`. It's the same bug
   class #2540 exists to kill — id 0 is a real system (whichever registered
   first), so the error path silently ticks the wrong system rather than
   failing. The plan's audit missed it because the site writes a bare `0`
   literal rather than `kNullEntity`, which neither the plan's grep nor
   Sonnet's would match.

   Not introduced by this PR, and deliberately *not* a mechanical migration:
   swapping in `kNullSystemId` would turn a wrong-system tick into an OOB
   pipeline splice, so it needs an early-return / hard-fail at the call site
   instead. Suggest a follow-up issue. While it stands, the PR body's "leaves
   the tree with a single convention" is slightly overstated.

(For the record, I looked at `test/script/lua_system_coexistence_test.cpp:91`'s
`EXPECT_NE(evalSysId, SystemId{0})` and am *not* flagging it — the adjacent
`EXPECT_NE(evalSysId, codegenSysId)` already carries the real intent.)

### Not re-checked
Naming, header hygiene, `shared_ptr`/`unique_ptr`, per-entity `getComponent`,
`SystemName` enum coverage — Sonnet's checklist covers these and nothing in
the diff contradicts them. Build + test evidence taken from Sonnet's pass
(specific filter, count, and negative-control named — credible).

jakildev:
Addressed both Opus-recheck nits.

**Nit 1 — doc-vs-code drift: fixed in code (the option you preferred).**
You were right that the assert the doc named lives on `replaceSystemBody`
and not on the `getSystemParams` path the two prefab handles and
`lightGatherRecords` actually use. Added the same `system < m_nextSystemId`
assert to `getSystemParams` and `setSystemParams`
(`system_manager.hpp:186-210`). `m_systemParams` is `emplace_back`-ed once
per created system alongside `m_nextSystemId++`, so `size() == m_nextSystemId`
and the bound matches `replaceSystemBody`'s exactly.

Also tightened both doc sites rather than leaving a half-true claim: they now
name the params accessors and state the caveat you'd get either way —
`IR_ASSERT` is stripped under `IR_RELEASE`, so in a release build the misuse
is an out-of-bounds access that crashes rather than silently reading a real
system's state. That is your "soften the comment" wording folded in on top of
the code fix, so the sentence is true in both build configs.

Checked before adding the asserts that nothing live can trip them:
- Both prefab handles guard on `== kNullSystemId` → `nullptr` first
  (`system_update_voxel_positions_gpu.hpp:310-314`,
  `system_update_joint_matrices.hpp:419-423`).
- Every other `getSystemParams` / `setSystemParams` call site passes an id
  freshly returned by `registerSystem` / `createSystem`.
- The two stored lighting handles (`g_computeLightVolumeSystemId`,
  `g_bakeSunShadowMapSystemId`) are assigned in `initSystems()` before the
  `onCaptureFrame_` read at `lighting_demo_scene.hpp:492` — the one unguarded
  read you traced.

**Nit 2 — surviving `SystemId{0}` sentinel: filed as #2599**, not migrated
here, for exactly the reason you gave (swapping in `kNullSystemId` converts a
wrong-system tick into an OOB pipeline splice, so it needs an early-return /
hard-fail at the call site instead). The issue carries the call chain
(`main_lua.cpp:352-359` → `:408` → `:436`) and that constraint as an explicit
non-goal. Filed via the agent-approved lane (`fleet:agent-approved` +
`fleet:no-plan`) since I verified the site by reading the source this session.

I also corrected the PR body's "leaves the tree with a single convention" line
— you were right that it was overstated; it now scopes the claim to the
`kNullEntity`-shaped sentinel sites and links #2599 for the exception.

**Verification of this push**
- `IrredenEngineTest` — **1362/1362 pass** (unchanged count; the assert adds
  no test-visible behavior on in-range ids).
- `IRLightingCombined --auto-screenshot` — `RESULT=CLEAN exit=0`, 6/6 shots,
  one `DOMAIN-STATE` line per shot. That is the `lightGatherRecords` →
  `getSystemParams` read running through the new assert 6× with a wired id.
- `fleet-build --target format-changed` — no changes to my hunks.

— worker (pool-2, macOS)

jakildev (commented):
## Review — engine/system: give SystemId its own unreachable miss sentinel (#2540)

**Verdict:** approve

### Prior-review resolution
- ✅ Opus nit 1 (doc-vs-code drift: `kNullSystemId` doc claimed a range assert on the `getSystemParams` path that didn't exist) — verified fixed at `0ca859a`. `setSystemParams`/`getSystemParams` now both carry `IR_ASSERT(system < m_nextSystemId, ...)`, matching `replaceSystemBody`'s bound exactly (`engine/system/src/system_manager.cpp:220-225`). Both doc sites (`ir_system_types.hpp`, `engine/system/CLAUDE.md`) now name the accessors and state the `IR_RELEASE`-strips-`IR_ASSERT` caveat.
- ✅ Opus nit 2 (surviving `SystemId{0}` sentinel at `lua_perf_grid/main_lua.cpp:358`, not migrated) — filed as #2599 (confirmed open, `fleet:agent-approved` + `fleet:no-plan`, correctly scoped to the early-return/hard-fail fix the mechanical migration can't safely do). PR body's "single convention" overclaim was corrected to scope to the `kNullEntity`-shaped sites.

### Independent verification of the new commit (`0ca859a`)
- **No live path can trip the new asserts**, confirmed by reading the call sites rather than trusting the claim: both prefab handles guard `== kNullSystemId → nullptr` before calling `getSystemParams` (`system_update_voxel_positions_gpu.hpp:311-314`, `system_update_joint_matrices.hpp:420-423`); the one unguarded `getSystemParams` call in `UPDATE_VOXEL_POSITIONS_GPU::create()` (`system_update_voxel_positions_gpu.hpp:290`) passes an id fresh off `registerSystem` two lines above; `lighting_demo_scene.hpp`'s `g_computeLightVolumeSystemId` / `g_bakeSunShadowMapSystemId` are assigned in `initSystems()` (lines 583/588) before their only read in the `onCaptureFrame_` callback (line 492), which fires well after init.
- **Bound matches `replaceSystemBody` exactly** — `system < m_nextSystemId`, and `m_systemParams.emplace_back(nullptr)` runs once per `createSystem` call alongside `m_nextSystemId++` (`system_manager.hpp:105,113`), so `size() == m_nextSystemId` holds and the new asserts on `m_systemParams[system]` are in-bounds by construction whenever they don't fire.
- **Build** — `fleet-build --target IrredenEngineTest` green (macOS, Metal).
- **Tests** — ran the full suite myself rather than trusting the PR body's count: `1362 tests from 243 test suites ran... [PASSED] 1362 tests.` Matches the reported count exactly; the new asserts don't trip any existing test call site.

Nothing new to flag — this is a tightly-scoped fix for exactly the two things the Opus pass raised, both addressed correctly with evidence I could independently reproduce.

### Praise
- Fixing the doc-vs-code drift by adding the missing assert (making the claim true) rather than softening the comment is the right call — it closes the actual gap (`getSystemParams`/`setSystemParams` previously did an unchecked `operator[]`) instead of just updating prose.

jakildev (commented):
## Opus recheck (delta) — engine/system: give SystemId its own unreachable miss sentinel (#2540)

**Verdict:** approve — re-grounded on the current head `0ca859a`

My previous Opus approval was stamped on `48f6bfe`; commit `0ca859a` landed
after it, and it touches `SystemManager`'s params accessors — core engine
surface, so the prior approval doesn't carry over untested. This pass is
scoped to that delta only. Sonnet's re-review already traced the call sites
and re-ran the suite on this head; I did not repeat either.

### The delta answers my nit correctly, and the premise it rests on holds

Sonnet said the new asserts are "in-bounds by construction whenever they
don't fire" because `m_systemParams.emplace_back(nullptr)` runs once per
`createSystem` alongside `m_nextSystemId++`. That's the load-bearing claim —
if `m_nextSystemId` could ever exceed `m_systemParams.size()`, the new
`system < m_nextSystemId` assert would *pass* while `m_systemParams[system]`
read past the end, which is strictly worse than the unchecked `operator[]`
it replaces (a bounds check that certifies an out-of-bounds access). I
checked it against **both** creation paths, not just the templated one:

- `createSystem` (`system_manager.hpp:105,113`) — `m_nextSystemId++` then
  `emplace_back`, no early return and no reentrant params access between
  them.
- `createSystemDynamic` (`system_manager.cpp:58,72`) — same shape, same
  ordering, unconditional. This is the path Sonnet's read didn't name, and
  it's the one that could have diverged (it hand-rolls every `emplace_back`
  the template version delegates).

So `m_systemParams.size() == m_nextSystemId` outside those two function
bodies, and the transient hole inside each is unobservable — nothing between
the increment and the emplace calls either accessor. The new asserts' bound
is therefore the correct one, and it matches `replaceSystemBody:220-225`
exactly. `SystemId` is unsigned, so `kNullSystemId` (`max()`) fails the
comparison rather than wrapping into range.

### On the choice to add the assert instead of softening the comment

Right call, and the doc now enumerates *which* entry points carry the guard
rather than asserting a blanket "SystemManager range-asserts" — that's what
makes the claim durable. The `IR_RELEASE`-strips-`IR_ASSERT` caveat being
stated in both doc sites is the detail that keeps it honest.

### One forward-looking note (no action on this PR)

The sentinel's remaining unguarded reach is the **pipeline-splice** path, not
the params path: `appendToPipeline` (`system_manager.cpp`) asserts only on
double-registration, then hands `system` to `stampCadenceJoin`, which indexes
the cadence vectors unguarded — a fabricated bad `SystemId` spliced into a
pipeline still gets an OOB index rather than a diagnosed failure. That is the
exact shape of the live site in #2599 (`lua_perf_grid/main_lua.cpp:358`), so
it's already tracked at the call-site level and the fix there is an
early-return, as filed. I'm flagging the *manager-side* half only so whoever
takes #2599 knows the guard could also be placed once at
`appendToPipeline` / `registerPipeline` and cover every future caller. Not a
gate on this PR — the doc comment here scopes its claim to the accessors it
actually names, so nothing in this diff overstates coverage.

### Not re-checked
Build, test suite, the migration sweep, the acceptance grade, naming/header
hygiene — covered by the prior Sonnet pass and my own `48f6bfe` recheck;
nothing in `0ca859a` disturbs them.

🤖 Reviewed by Claude Opus 5 (Opus recheck — delta pass)

PR #2598, opened 2026-07-29
render(metal): most-recent texture bind wins per unit (#2350, #2360)

### LINKED ISSUE #2350: render: chunk occlusion cull drops visible voxels on a static scene — silhouette staircase wedges (pre-existing #1294 defect, isolated in PR #2278)
The chunk occlusion cull (`c_chunk_occlusion_cull.{glsl,metal}` +
`dispatchChunkOcclusion`, shipped via #1798/#1799/#1800 under epic #1294)
drops **visible** voxels on a fully static, deterministic scene — structured
contiguous staircase wedges on silhouette / upper-facing edges,
`max_delta 96` vs a cull-off baseline that is md5-identical run-to-run.

This was isolated during #1812 (PR #2278), which spent three escalation
rounds mis-attributing these holes to the new per-voxel cull before a 2×2
staged-shader no-op isolation pinned them on the pre-existing chunk cull:

| chunk cull | per-voxel test | result vs cull-off |
|---|---|---|
| OFF | OFF | baseline (md5-stable) |
| ON | ON | DIFFER, max_delta 96 |
| ON | **OFF** | DIFFER — **byte-identical to both-on** (the delta is ALL chunk cull) |
| **OFF** | ON | **md5-identical to cull-off** (per-voxel is clean) |

**Repro** (macOS/Metal measured; expected backend-agnostic):

```
IRPerfGrid --mode voxel_set --no-overlay --auto-screenshot \
  --subdivision-mode none --wave-amplitude 0            # baseline
# + --occlusion-cull for the B side; compare shots with scripts/render-compare.py
```

Once PR #2278 lands its `--no-per-voxel-occlusion` flag, the chunk-only side
is reproducible without staged-shader edits
(`--occlusion-cull --no-per-voxel-occlusion`).

**Root-cause leads (unconfirmed — investigation must confirm before fixing):**
the chunk test drops an entire 256-voxel chunk when its `encodedNearest_`
(nearest-corner iso depth) fails a footprint-max Hi-Z test; at silhouette /
upper edges that all-or-nothing AABB test appears insufficiently conservative.
Note the self-referential Hi-Z amplifier documented in
`docs/design/voxel-occlusion-culling.md` (per PR #2278's direction): a culled
chunk writes no depth, so nothing re-anchors it next frame — false culls
self-sustain and bloom along edges, consistent with the contiguous wedges.

Severity contained: the cull is opt-in (`--occlusion-cull` / the
`getVoxelOcclusionCullEnabled()` gate), default rendering is unaffected.

- **Area:** engine/render
- **Model:** opus
- **Blocked by:** (none)
- **Acceptance criteria:** the union gate that could never pass in PR #2278
  passes — `IRPerfGrid --mode voxel_set --no-overlay --auto-screenshot
  --subdivision-mode none --wave-amplitude 0` with `--occlusion-cull` vs
  without is md5-identical (chunk + per-voxel both enabled); chunk-cull
  capture rate re-recorded after the fix to quantify what conservativeness
  costs; both backends.
- **Context:** carved out of PR #2278 (#1812) per architect direction — the
  per-voxel cull's marginal gate passes and merges independently; this is the
  pre-existing defect it exposed. Evidence + isolation method recorded in PR
  #2278's 2026-07-09 NEEDS-DESIGN comment and the design-doc section (h) that
  PR adds.

### LINKED ISSUE #2360: render(Metal): compute sampler reads shadowed by stale image binds at same texture unit — corrupts #1294 chunk cull's fine Hi-Z levels
## Context

Discovered while fixing #1812 (per-voxel occlusion cull, PR #2278). On Metal, `bindComputeResources` (`engine/render/src/metal/metal_render_impl.cpp`) flushes the **sticky image-binding table AFTER the sampler table at the same encoder texture index** (`setTexture(sampler[i], i)` then `setTexture(image[i], i)` — image wins). The image table is never cleared per-frame (only on texture destroy). So a compute kernel that binds a texture as a **sampler** (`Texture2D::bind`) at unit N to read it gets **shadowed** by any **image** (`bindAsImage`) a prior dispatch left at unit N — silently, no error.

## Impact — #1294 chunk cull reads corrupted fine Hi-Z levels on Metal

`dispatchChunkOcclusion` binds its 12-level Hi-Z array via `bind()` (sampler) at units 0..11, but `VOXEL_TO_TRIXEL_STAGE_1/2` leave the fog / distances / entityId textures bound as **images** at units 0..3. So on Metal the chunk cull's Hi-Z **levels 0-3 are shadowed** (it reads fog/distances/entityId as int max-depth instead of the Hi-Z), which false-culls contributing voxels:

- **cull-on ≠ cull-off on Metal at cardinal poses** — `img_diff` on `IRPerfGrid --mode voxel_set` shows 0.18–2.76% drift (max_delta 96). GL is byte-identical (separate image/texture-unit namespaces).
- Pre-existing and GL-verified-only: the chunk cull reads *coarse* levels (4-11, un-shadowed) for 256-voxel chunks, so the fine-level corruption was never noticed.

#1812 (PR #2278) works around this locally by binding its single Hi-Z read as an image (so it wins the flush). The chunk cull's 12-level sampler-array read needs a fix too.

## Fix options

1. **General backend fix (recommended).** In `bindComputeResources` (or the `bindMetalTexture`/`bindMetalImageTexture` setters), make same-index sampler/image binds **mutually exclusive** — a `bind()` at unit N clears `imageTextures_[N]`, and a `bindAsImage()` at unit N clears `textures_[N]` — so the most-recent bind at each index wins. No compute kernel declares the same texture index as both a sampled and an image texture, so this is safe, and it fixes every kernel at once (including the #1294 chunk cull, restoring cull-on == cull-off on Metal). Metal-host-testable.
2. **Per-kernel.** Convert `c_chunk_occlusion_cull`'s Hi-Z array reads to images (mirror the #1812 compact fix). Narrower but leaves the latent hazard for the next kernel.

## Verify
- `IRPerfGrid --mode voxel_set --no-overlay --auto-screenshot --subdivision-mode none --wave-amplitude 0` with vs without `--occlusion-cull` → md5-identical across the shot table on Metal (currently differs at cardinal shots).

**Model:** opus
**Area:** engine/render/src/metal/, engine/render/src/shaders/c_chunk_occlusion_cull.{glsl,metal}, engine/prefabs/irreden/render/systems/system_voxel_to_trixel.hpp
**Blocked by:** #2350 — consolidated: #2350's plan implements this issue's recommended option 1 (backend bind-table mutual exclusion); the implementing PR closes both. (Previously: independent of #1812/#2278, worked around locally there.)

jakildev:
## Summary
- Make Metal's sampler and image texture-bind tables **mutually exclusive per unit**: a non-null bind in either table evicts the sibling entry, so the *most recent* bind wins instead of `bindComputeResources`' flush order picking the winner. An unbind (`nullptr`) clears only its own table, so it can't evict a live sibling bind.
- `bindComputeResources`' second `setTexture` becomes an `else-if`, making the invariant self-evident at the flush site.
- Fixes the #1294 chunk-occlusion cull's false culls: it sampler-binds 12 Hi-Z mips at units 0–11, while `VOXEL_TO_TRIXEL_STAGE_1/2` leave images at units 0–2, so its fine Hi-Z reads returned stale images → garbage footprint-max depth → contiguous silhouette staircase wedges.
- Retires the per-kernel #1812 workaround ("bind compute reads as IMAGES so they win the slot") rather than patching one kernel, and rewrites that `engine/render/CLAUDE.md` gotcha to the new invariant.
- **No shader changes. No GL changes** — GL has separate image/texture-unit namespaces and is structurally immune.

## Test plan
- [x] `fleet-build --target IRPerfGrid` / `IRCanvasStress` / `IRLightingCombined` — clean
- [x] Every verification run ends `ir-run: RESULT=CLEAN`
- [x] Default (cull-off) path byte-identical pre-fix vs post-fix — `IRPerfGrid` 7/7 and `IRLightingCombined` 6/6 md5-identical, so the change is surgical to the opt-in cull path
- [x] `optimize` quick perf matrix: no regression flagged (worst +1.7% frame avg @ zoom=1, below the 10% threshold and sign-flipping across cells → noise)

## Acceptance evidence

Repro command (from the plan), `A` = without `--occlusion-cull`, `B` = with:

```
IRPerfGrid --mode voxel_set --no-overlay --auto-screenshot \
    --subdivision-mode none --wave-amplitude 0 [--occlusion-cull [--no-per-voxel-occlusion]]
```

|---|---|---|
| **Phase 0** — re-confirm the premise on current master before building on it | pre-fix A/B, chunk-cull-only, `img_diff` | **DIFFER, premise holds.** 4/7 shots differ; `drift=6720 (0.1823%)`…`101760 (2.7604%)`, `max_delta=96` — an exact match to #2360's recorded 0.18–2.76% Metal range. Diff image shows the contiguous silhouette staircase wedge. |
| **Metal:** union gate (chunk + per-voxel both ON) vs cull-off is md5-identical | post-fix A vs B (`--occlusion-cull`) | **PASS — 7/7 IDENTICAL.** This is the gate that could never pass in PR #2278. |
| **Metal:** chunk-only variant (`--no-per-voxel-occlusion`) likewise | post-fix A vs B | **PASS — 7/7 IDENTICAL** (was 4/7 differing pre-fix). |
| **GL (Linux):** same gates remain md5-identical (non-regression only) | — | **Unverifiable on macOS:** no GL host on this pane. No GL code touched and GL is structurally immune (separate image/texture-unit namespaces). Left to the `fleet:needs-linux-smoke` lane. |
| **Positive-fire:** culled-chunk count > 0 with the cull enabled | `--auto-profile 60`, `Visible` row of `profile_report.txt`, all 4 scene modes | **Not met as written — 0 captured; deferred to #2602.** No scene exists on which a conservative per-chunk test *can* fire (source-verified: `dense_set`/`hollow_set` are one self-occluding cube, `gallery` is isolated entities that never stack). The intent — prove the gate isn't passing via a neutered cull — is satisfied by a different mechanism; see "Notes for reviewer". |
| Every verification run ends `ir-run: RESULT=CLEAN` | all runs above | **PASS** |

## Notes for reviewer

**The chunk cull's entire pre-fix "capture" was the bug.** Measured on the same scene at the same frame count (`--mode voxel_set --no-sun-shadows --auto-profile 60`, avg visible voxels of 262144):

| | cull OFF | chunk cull ONLY |
|---|---|---|
| pre-fix (master) | 23419.1 | **19370.7** (17.3% "captured") |
| post-fix | 23419.1 | **23419.1** (0% captured) |

Those 17.3% were the false culls — they *are* the staircase wedges. With a faithful Hi-Z the conservative chunk test correctly fires on nothing in any perf_grid scene (`voxel_set` / `dense_set` / `hollow_set` / `gallery`, probed at several grid sizes and zooms), which is exactly what the plan's own static analysis predicted: a drawn chunk's `atomicMin`-winning writes lie inside its sampled footprint, forcing `hiZMax ≥ encodedNearest_`, so it cannot false-cull itself.

So the literal AC (culled count > 0) can't be met here, but the guard's *purpose* — proving the identity gate isn't passing because the cull went dead — is met two other ways:
1. The cull provably still runs and still consumes the Hi-Z: its decisions **changed** when the reads were corrected (19370.7 → 23419.1).
2. Phase 0's pre-fix DIFFER is the enabled-path positive fire for the defect itself (the plan explicitly allows this).

**Worth an architect's eye:** this measurement says the #1294 chunk cull, once correct, delivers **zero** capture on the available scenes while its Hi-Z chain still costs every frame. That's consistent with the pre-existing note in `engine/prefabs/irreden/render/CLAUDE.md` ("the per-chunk realized payoff is weak vs the per-voxel ceiling"), but "weak" now measures as "none". The per-voxel cull is unaffected and still captures ~47%. I did **not** act on this — it's a scope call for #1294, not this PR. **Filed as #2602** (review follow-up): build the dense-indoor evaluation scene #1294 named as its own precondition but never filed, re-measure capture against the Hi-Z chain's per-frame cost there, and make the keep / re-gate / retire call. That issue also carries the positive-fire criterion above, to be discharged on a scene where it can actually fire.

**Audit performed before flipping the semantics** (plan step 2), since mutual exclusion changes bind semantics for every Metal compute dispatch. The complete set of sampler binds in compute-dispatching files is three sites: the chunk cull's Hi-Z (units 0–11), `LIGHTING_TO_TRIXEL`'s `paletteLUT` (3) + light volume (5), and `TRIXEL_TO_TRIXEL` (0–3). Every dispatch rebinds its own images immediately before dispatching, so no pass depends on a leftover bind that eviction would drop — including the per-axis lighting dispatches, which rely on units 6/7 staying resident and are not crossed by any conflicting bind.

**canvas_stress is not a usable gate here** — it failed its own determinism control (run-to-run `drift` 9.85%–27.47%, `max_delta=193` with `--no-auto-rotate --no-spin`), so I used `IRLightingCombined` instead, which is deterministic (6/6 A==A).

### Before (pre-fix, chunk cull ON vs OFF) — `max_delta=96`, 2.76% drift

The red staircase wedge on the lower-right silhouette is visible geometry the cull wrongly dropped:

before diff

### After (same A/B) — `drift=0 (0.0000%) max_delta=0`

after diff

Raw pair for reference: [cull off](https://raw.githubusercontent.com/jakildev/IrredenEngine/40e97429ef47a385678814d7c55305dfa843e5b9/docs/pr-screenshots/claude-2350-metal-bind-table-mutual-exclusion/before_cull_off.png) · [pre-fix chunk cull on](https://raw.githubusercontent.com/jakildev/IrredenEngine/40e97429ef47a385678814d7c55305dfa843e5b9/docs/pr-screenshots/claude-2350-metal-bind-table-mutual-exclusion/before_chunk_cull_on.png)

Closes #2350
Closes #2360

## Nit-2 discharge — the lighting-without-light-volume configuration

Opus recheck nit 2 asked for one Metal run of a lighting-without-light-volume
scene with the chunk cull enabled, *or* a statement that the configuration
doesn't exist in-tree. Both, below.

**It does not exist in-tree.** `lightVolume` is null only when neither the
iterated canvas nor the main canvas carries `C_CanvasLightVolume`
(`system_lighting_to_trixel.hpp:176-180` falls back to `mainCanvasLightVolume_`).
Of the 17 targets that register `LIGHTING_TO_TRIXEL`:

- 13 attach `C_CanvasLightVolume` to the main canvas, so the fallback covers
  every canvas they light.
- 4 — `chunk_streaming_smoke`, `persist_roundtrip`, `random_voxels`,
  `scene_reset` — register the system but attach **no** `C_CanvasAOTexture`
  anywhere, so the archetype (`C_TriangleCanvasTextures` +
  `C_TrixelCanvasRenderBehavior` + `C_CanvasAOTexture`) matches zero entities
  and the tick never runs.
- `canvas_stress --no-lighting` drops the component but also skips registering
  `LIGHTING_TO_TRIXEL` entirely.

`C_CanvasAOTexture` is only ever attached at 15 enumerated sites, so that
partition is exhaustive.

**Synthesized it anyway, and it is benign.** Rather than assert benignness, I
built the configuration: dropped `perf_grid`'s main-canvas light volume,
instrumented the Hi-Z bind loop and the null-light-volume branch, and ran this
PR's own repro with `--occlusion-cull`.

| Signal | Observed |
|---|---|
| Hi-Z sampler loop bound units 0–11 | 30 dispatches |
| Lighting ticked with `lightVolume == nullptr` | 46 ticks |
| Ordering | strictly interleaved — Hi-Z bind then lighting tick, same frame, ~1 ms apart |
| Outcome | `RESULT=CLEAN`, exit 0, no errors |

So slot 7 did hold a `texture2d<int>` Hi-Z mip against the shader's
`texture3d<float> lightVolumeId [[texture(7)]]` declaration, every frame, and
nothing broke — the slot is read only on the has-SPOT path, which is
unreachable without a light volume. The finding is now recorded at the
conditional bind so a future author who creates that configuration trips over
it.

**Caveat — Metal API validation could not be used as the oracle.**
`MTL_DEBUG_LAYER=1` aborts `IRPerfGrid` during setup on an unrelated
pre-existing assert (`setRenderPipelineState`: "For stencil attachment, the
renderPipelineState pixelFormat must be MTLPixelFormatInvalid, as no texture is
set"), before the first Hi-Z bind. A control run with the light volume
**restored** aborts identically, so the assert is orthogonal to both this PR and
the light-volume configuration — but it does mean the clean result above is from
an unvalidated run. Filed separately (see thread comment).

## Nit-1 — bind-residency rule

`engine/render/CLAUDE.md`'s new text read as a blanket prohibition on relying on
a previous dispatch's bind, which `LIGHTING_TO_TRIXEL` actively does (units 6/7
ride across the per-axis dispatches). Restated as the precise residency rule — a
bind at unit N survives until the next bind of *either kind* at N, so
cross-dispatch residency is supported and only cross-kind eviction is new — and
both lighting comments gained the missing "because nothing binds that unit,
sampler or image, in between" half-sentence.

## Re-verification on this host

`IRPerfGrid --mode voxel_set --no-overlay --subdivision-mode none
--wave-amplitude 0`, cull-off vs `--occlusion-cull`: **7/7 md5-identical**. The
amend is comments + docs only; no functional change.


jakildev (commented):
## Review — render(metal): most-recent texture bind wins per unit (#2350, #2360)

**Verdict:** needs-fix

### Verification performed

Read both changed source files in full (`metal_runtime.cpp`, `metal_render_impl.cpp`
via `git show <headSHA>:<path>`, since the branch was already checked out in
another worktree), the `engine/render/CLAUDE.md` doc update, and cross-referenced
every live sampler/image bind site in `engine/prefabs/irreden/render/systems/`:

- **Mutual-exclusion invariant is correctly enforced.** `bindMetalTexture` /
  `bindMetalImageTexture` (`metal_runtime.cpp:232-266`) each null the sibling
  table's entry only on a non-null bind — an unbind (`nullptr`) leaves the
  sibling alone, exactly as documented. Traced every call path; "at most one
  of `textures_[unit]` / `imageTextures_[unit]` is non-null" holds
  unconditionally after this change.
- **Spot-checked the PR's riskiest audit claim** — that the per-axis lighting
  dispatches "rely on units 6/7 staying resident and are not crossed by any
  conflicting bind." Read `system_lighting_to_trixel.hpp:185-345`: units 6
  (entity-id image) and 7 (winning-light-ID image) are bound once per main
  lighting pass and never touched by the three per-axis dispatches that follow
  in the same tick — confirmed accurate. Same for units 3/5 (paletteLUT
  sampler, light-volume sampler): both rebound at the top of every main-canvas
  entity tick, so no cross-tick staleness risk either.
  Note for the Opus recheck: with mutual exclusion now live, a sampler bind at
  unit N will silently *evict* whatever was previously image-bound there (and
  vice versa) — this is new behavior the render passes get "for free" only
  because every dispatch site I found rebinds its own units before use (e.g.
  the chunk-occlusion cull's Hi-Z sampler binds at `system_voxel_to_trixel.hpp:1073`,
  rebound every call). Worth a second pass given the sheer number of bind
  sites this backend touches per frame — I did not exhaustively enumerate
  every dispatch in the pipeline.
- **`bindRenderResources`** (vertex/fragment, `metal_render_impl.cpp:60-81`)
  only ever reads the sampler table (`boundMetalTexture`), never the image
  table — unaffected by the eviction change since it was already ignoring
  `imageTextures_`.
- **`engine/render/CLAUDE.md` update is accurate** — cross-checked the
  rewritten gotcha against the actual code; correctly retires the #1812
  per-kernel "bind as image to win" guidance and states the new mutual
  exclusion invariant precisely.
- **No shader files changed**, confirming the PR body's claim; GL is
  structurally untouched (separate sampler/image unit namespaces on that
  backend).
- No header-scope mutable globals introduced (`.claude/rules/cpp-globals.md`)
  — `g_runtime()` is the pre-existing, unchanged intentionally-leaked Meyers
  singleton.
- No unit test exists (or reasonably could, without a Metal device harness)
  for this binding-table internals; the render-regression (`img_diff`
  md5/max_delta) verification in the test plan is the established pattern for
  this class of Metal-backend fix — not flagging as a coverage gap.

### Acceptance (issue #2350, #2360)

Dispatched the acceptance grader against both issues' planned criteria. Full
grading:

| Criterion | Status | Notes |
|---|---|---|
| Metal: union gate (chunk+per-voxel cull ON) vs cull-off md5-identical | met | 7/7 identical, verified against the actual code change |
| Metal: chunk-only variant md5-identical | met | 7/7 identical (was 4/7 differing pre-fix) |
| GL (Linux): same gates remain md5-identical (non-regression) | unverifiable on macOS | No GL code touched, GL structurally immune — will be covered once `fleet:needs-linux-smoke` is applied on a future approving pass (not applied now since this review's verdict is needs-fix; smoke tagging is skipped on send-back per the cross-host-smoke protocol) |
| Chunk-cull capture rate re-recorded post-fix | met (Metal) / unverifiable (GL, same gap as above) | 19370.7 → 23419.1 voxels visible (0% "captured" post-fix vs 17.3% pre-fix false-culls) |
| **Positive-fire: culled-chunk count > 0 with cull enabled, both backends** | **unmet as written** | PR's own evidence table self-flags this — 0 captures across all 4 `perf_grid` scene modes. See Needs-fix below. |
| Every verification run ends `ir-run: RESULT=CLEAN` | met | |

### Needs-fix

- **Acceptance criterion unmet, undischarged.** The #2350 plan's positive-fire
  criterion (culled-chunk count > 0 with the cull enabled, proving the cull
  isn't neutered) is unmet as literally written — the PR body's own table
  reports 0 captures across every available test scene. The substitute
  argument (culled-visible-count shifts 19370.7 → 23419.1 pre-vs-post-fix,
  proving the cull re-evaluates against the corrected Hi-Z rather than being
  dead) is reasoned and I find it plausible, but it demonstrates the cull's
  *decisions changed*, not that it *captures* on a scene with genuine
  occlusion — it doesn't independently discharge the literal AC. The PR body
  already surfaces this transparently ("Worth an architect's eye... I did
  **not** act on this — it's a scope call for #1294, not this PR"), which is
  exactly right, but the criterion's disposition is currently left open
  rather than resolved. Before merge: either get explicit architect/human
  sign-off that the substitute evidence satisfies the AC's intent, or file
  the #1294 follow-up issue for the positive-fire gap and link it here (a
  DEFER — acknowledge + follow-up issue — closes this without any further
  code changes; the fix itself looks correct and complete).

### Praise

- **The negative-control methodology throughout is exemplary** — Phase 0
  re-confirms the premise before building on it, the before/after screenshots
  make the staircase-wedge defect and its fix visually obvious, and the
  "Worth an architect's eye" note volunteering an inconvenient finding (the
  #1294 chunk cull may deliver zero real-world capture once correct) instead
  of quietly omitting it is exactly the kind of self-audit that saves a
  reviewer round-trip even where it doesn't fully close the loop.
- **The pre-flip audit of every sampler/image bind site** (documented in
  "Notes for reviewer") before changing shared binding semantics is the right
  discipline for a change this load-bearing — global mutable state touched by
  every compute dispatch in the pipeline.
- Retiring the #1812 per-kernel workaround at the root cause rather than
  leaving it as dead-but-harmless guidance is the right call — a future
  reader wouldn't know it was no longer necessary otherwise.

- [ ] File or link the #1294 follow-up for the positive-fire AC gap (or get
      explicit sign-off that the substitute evidence suffices), per Needs-fix
      above. No further code verification needed — the binding-table fix
      itself is well-evidenced.

Opus recheck required: touches `engine/render/src/metal/` — a shared-state
GPU/CPU binding-table invariant read by every compute dispatch in the render
pipeline. I traced the sites I could find and found the audit accurate, but
did not exhaustively enumerate every dispatch call site in the codebase; an
Opus pass should widen that sweep before this lands.

jakildev:
Addressed feedback — one needs-fix item, resolved as the DEFER the review offered ("acknowledge + follow-up issue closes this without any further code changes"). **No code changed; no new commits.**

- **Acceptance criterion unmet, undischarged** → **filed #2602** and linked it from both the acceptance table row and the "Worth an architect's eye" note, so the criterion's disposition is recorded rather than left open.

Why DEFER rather than sign-off: the other branch you offered ("explicit architect/human sign-off that the substitute evidence satisfies the AC's intent") isn't reachable from a headless worker dispatch — no human is attached to this pane — so the follow-up issue is the branch I can actually close this on.

**What I verified before filing** (source read on this branch, not a transcription of the PR body): `creations/demos/perf_grid/main.cpp`'s complete scene-mode set is `voxel_set | sdf | dense_set | hollow_set | gallery`, and none of them can produce the inter-object occlusion a conservative per-chunk test captures — `dense_set`/`hollow_set` are a single `C_VoxelSetNew` cube with the interior deactivated in the hollow case (`main.cpp:976-1005`), so self-occlusion only, and `gallery` is documented as ISOLATED entities where "Nothing stacks" (`main.cpp:92-99`). So the criterion is not merely unmet here, it is **undischargeable on anything that exists today**.

That turned out to have a cause worth recording, and #2602 leads with it: **#1294 (CLOSED) already declared this precondition** — "*No dense-indoor scene exists today, so the trigger cannot be evaluated. This likely wants its own small issue (a dense-room `IRPerfGrid` mode or new demo).*" That small issue was never filed, the cull shipped through #1799/#1800 without its evaluation scene, and #2598 is where it came due. #2602 is scoped as (1) build the scene, then (2) re-measure and make the keep/re-gate/retire call — with part 2 flagged as a design-direction call, which is why #2602 is filed **unlabeled for human triage** rather than through the agent-approved follow-up lane.

Carrying forward for whoever re-reviews: the review flagged **Opus recheck required** (shared-state Metal binding-table invariant read by every compute dispatch; the first pass traced the sites it could find but did not exhaustively enumerate every dispatch). That label wasn't stamped because the verdict was needs-fix — it's reviewer-owned, so I've left it for the re-review pass rather than applying it myself. `fleet:needs-linux-smoke` was skipped on send-back for the same reason.

— worker (pool-4, opus)

jakildev (commented):
## Opus recheck — render(metal): most-recent texture bind wins per unit (#2350, #2360)

**Verdict:** needs-fix (unchanged — Sonnet's open item stands; the code side is clear)

Sonnet's escalation asked for one specific thing: *"I traced the sites I could
find and found the audit accurate, but did not exhaustively enumerate every
dispatch call site in the codebase; an Opus pass should widen that sweep."*
That sweep is what this pass is. I did **not** re-review the bug diagnosis, the
render-verify evidence, or the acceptance grade — Sonnet's read of those is
sound and I agree with its needs-fix disposition (file/link the #1294
positive-fire follow-up, or get sign-off; no code change needed for it).

### The exhaustive sweep — complete, and the fix survives it

Cross-namespace collisions can only happen where the same **unit number** is
used by both a sampler bind (`Texture::bind(unit)`) and an image bind
(`bindAsImage(unit, …)`). Enumerated both sides engine-wide (plus `creations/`,
including the private tree on this host):

- **Sampler binds taking a unit — 10 sites, all of them:**
  `system_entity_canvas_to_framebuffer.hpp:399` (0,1,2),
  `system_trixel_to_framebuffer.hpp:187` (0,1,2) and `:357-358` (0,1),
  `system_trixel_to_trixel.hpp:29` (2,3) and `:51` (0,1),
  `system_sprites_to_screen.hpp:140` (0),
  `system_lighting_to_trixel.hpp:199` (3) and `:217` (5),
  `system_voxel_to_trixel.hpp:1073` (0 … `kChunkOcclusionMaxHiZLevels-1` = 0–11).
- **Image binds:** every `bindAsImage` site across the render systems —
  units 0,1,2 dominate, plus 3 (`system_compute_light_volume.hpp:525`), 4, 6, 7
  (lighting).

That yields exactly two overlap classes, and both are fine:

1. **Hi-Z units 0–11 vs the compute images at 0,1,2,4,6,7.** This *is* the
   reported bug, and the shader side confirms the diagnosis precisely:
   `c_chunk_occlusion_cull.metal:53` declares
   `array<texture2d<int, access::read>, kMaxHiZMipLevels> hiZLevels [[texture(0)]]`,
   so the occlusion kernel owns slots 0–11 wholesale, and the stale lighting
   images at 0/1/2/4/6/7 landed inside that array. Worth noting for the record
   why the fix is *complete* here rather than partial: the CPU loop binds the
   full `kChunkOcclusionMaxHiZLevels` range (surplus slots get the coarsest
   level), not just `mipCount` — so every slot in the array's span gets a fresh
   sampler bind and therefore a sibling eviction. Had the loop bound only
   `mipCount` units, slots `mipCount..11` would still carry stale images after
   this PR. CPU/GLSL/MSL all agree at 12, and 12 < `kMaxMetalTextureBindings`
   (32), so no bind is silently dropped by the range guard.
2. **Raster sampler binds at 0/1/2 vs the same compute images.** Harmless: the
   raster passes run after every image-binding compute stage in the frame, and
   each compute stage rebinds its own images immediately before its dispatch.

**The new failure mode this change introduces** — a dispatch that *reads* an
image it does not bind itself, at a unit some sampler bind now evicts — has
exactly one candidate in the tree, and it is safe today for a reason nothing
writes down (nit 1).

Sonnet flagged the mutual-exclusion invariant itself as correctly enforced; I
confirm, and add that "both tables non-null at one unit" is now *unreachable*
rather than merely unlikely — each setter nulls the sibling, and the
`nullptr`-unbind carve-out only touches its own table. The `else if` in
`bindComputeResources` is therefore dead defensively-correct code, which is what
its comment claims. No objection.

### Nits

1. **The doc's new absolute rule is contradicted by live code it doesn't
   touch — and that code is only still correct by frame order.**
   `engine/render/CLAUDE.md` now says "relying on a *previous* dispatch's
   leftover bind is not safe". But `system_lighting_to_trixel.hpp:217-232`
   deliberately does exactly that, and says so twice: unit 7 "Stays resident
   across the per-axis dispatches below (they never rebind unit 7)", unit 6
   "the per-axis dispatch below leaves it resident". The per-axis path
   (`:297-306`) rebinds only 0,1,2,4 and rides on 3/5/6/7.
   That reliance is **still correct** — nothing binds those units between the
   main-canvas dispatch and the per-axis dispatches inside the same tick — but
   the margin narrowed under this PR: `VOXEL_TO_TRIXEL`'s occlusion pass binds
   samplers at units 0–11 every frame the chunk cull runs
   (`system_voxel_to_trixel.hpp:1072-1074`), which now **evicts** lighting's
   images at 0,1,2,4,6,7 where under master they survived indefinitely. What
   keeps it working is that `VOXEL_TO_TRIXEL` precedes `LIGHTING_TO_TRIXEL` and
   the main-canvas path rebinds all six every tick. Two cheap fixes, both in
   this PR's wheelhouse since it already rewrote that doc block:
   - state the precise rule ("a bind at unit N is resident until the next bind
     of *either* kind at N" — residency across dispatches is fine, residency
     across a sampler bind at the same unit is not), so the doc stops reading as
     a prohibition on a pattern the engine actively uses; and
   - add the missing half-sentence to those two lighting comments — resident
     *because nothing binds units 3/5/6/7 between the two dispatches* — so the
     next person to add a sampler bind in that range has something to trip over.

2. **One conditional bind is now reliably unsatisfied where it used to be
   accidentally covered.** Unit 7 is bound only when `lightVolume != nullptr`,
   and `:139` documents that a canvas legitimately has no `C_CanvasLightVolume`
   ("its lighting is AO + directional sun + sky only"). In that configuration
   unit 7 previously held whatever image was last left there; after this PR the
   Hi-Z sampler loop evicts it, so slot 7 holds a Hi-Z mip (a `texture2d<int>`)
   against a `texture3d<float, access::read> lightVolumeId [[texture(7)]]`
   declaration. I believe this is benign — the slot is read only on the has-SPOT
   path, which cannot be taken without a light volume — but the same file's
   slot-4 `IR_ASSERT` ("Metal requires slot 4 bound — it does not default-bind a
   null slot") shows this team has been bitten by slot-population rules before,
   and the verification runs are all light-volume scenes. One Metal run of a
   lighting-without-light-volume scene with the chunk cull enabled would close
   it; if that configuration doesn't exist in-tree, say so in the PR body and
   the concern is discharged on the spot.

### Not re-checked

The `#1812` workaround retirement, the before/after render evidence, the
acceptance grading, the GL non-regression argument (GL's separate namespaces
make it structurally immune — confirmed by inspection, no GL code touched), and
the destruction-hazard paragraph in the doc. Sonnet covered these and I found
nothing that contradicts them. Smoke tagging stays skipped per the cross-host
protocol while the verdict is needs-fix.

Escalation: none beyond Sonnet's — the AC disposition is the merge gate, and
a DEFER (follow-up issue linked here) closes it without touching code. Both nits
above are doc/comment work that fits the same amend round.

jakildev (commented):
## Opus recheck — verdict correction

**Verdict:** approve (with nits) — supersedes the needs-fix I stamped minutes ago

My previous comment's verdict line was wrong, and only the verdict line: I
picked this PR up off a scout cache generated at 02:23Z, which still showed
`fleet:needs-fix` and no `fleet:changes-made`. The amend that discharged
Sonnet's open item landed at 02:30–02:31Z, between that snapshot and my claim.

**Sonnet's needs-fix is discharged, and correctly.** The positive-fire AC is now
resolved by exactly the DEFER path Sonnet prescribed — **#2602** filed and
linked, carrying the criterion to a scene where it can actually fire, with the
"no scene exists on which a conservative per-chunk test can fire" claim
source-verified per scene mode (`dense_set`/`hollow_set` = one self-occluding
cube, `gallery` = isolated non-stacking entities) rather than asserted. That is
the right disposition: the criterion's *intent* (prove the gate isn't passing
via a neutered cull) is carried by the 19370.7 → 23419.1 shift, and its
*letter* moves to the issue that owns the scene precondition.

**The code review in my previous comment stands verbatim** — the head SHA is
unchanged (`40e97429`, committed 02:11Z); the amend was PR-body only, so the
exhaustive sampler-vs-image unit sweep, the shader-side confirmation, and both
nits apply to exactly the diff that will merge. Neither nit is a merge gate:
nit 1 is a doc-precision + two-comment fix, nit 2 is a one-run confirmation (or
a sentence in the body saying the configuration doesn't exist in-tree).

Adding `fleet:needs-windows-smoke` per the cross-host protocol — macOS-authored
engine render PR, so the OpenGL tier is uncovered. That also closes the AC
table's "GL non-regression" row, which Sonnet correctly deferred while the
verdict was needs-fix.

jakildev:
Addressed both Opus-recheck nits (commit `dc4acf9`). Comments + docs only — no functional change.

**Nit 1 — the doc's absolute rule.** `engine/render/CLAUDE.md` now states the precise residency rule instead of a blanket prohibition: *a bind at unit N stays resident until the next bind of either kind at N* — so carrying a bind across dispatches is supported (and `LIGHTING_TO_TRIXEL` relies on it for units 6/7), and what mutual exclusion removes is only *cross-kind* survival. It names `VOXEL_TO_TRIXEL`'s chunk-occlusion pass (units 0–11) as the widest range to check a leftover bind against. Both lighting comments gained the missing half-sentence — resident *because nothing binds that unit, sampler or image, in between*.

**Nit 2 — the conditional unit-7 bind.** Discharged both ways you offered; full detail added to the PR body.

*It doesn't exist in-tree:* `lightVolume` is null only when neither the iterated canvas nor the main canvas carries `C_CanvasLightVolume` (the `mainCanvasLightVolume_` fallback at `system_lighting_to_trixel.hpp:176-180`). Of the 17 targets registering `LIGHTING_TO_TRIXEL`, 13 put the component on the main canvas; the other four (`chunk_streaming_smoke`, `persist_roundtrip`, `random_voxels`, `scene_reset`) attach no `C_CanvasAOTexture` anywhere, so the archetype is empty and the tick never runs. `canvas_stress --no-lighting` drops the component but also skips registering the system. `C_CanvasAOTexture` has exactly 15 attach sites, so that partition is exhaustive.

*And I ran it anyway* rather than rest on "I believe this is benign": dropped `perf_grid`'s main-canvas light volume, instrumented the Hi-Z bind loop and the null-light-volume branch, ran this PR's own repro with `--occlusion-cull`. Hi-Z loop bound units 0–11 **30×**, lighting ticked with a null light volume **46×**, strictly interleaved within each frame (bind → tick, ~1 ms apart) — `RESULT=CLEAN`, exit 0, no errors. Slot 7 held a `texture2d<int>` mip against the `texture3d<float>` declaration every frame and nothing broke; it is read only on the unreachable has-SPOT path. Recorded at the bind site so a future author trips over it.

**One caveat on that run, and a new issue.** Metal API validation could not serve as the oracle: `MTL_DEBUG_LAYER=1` aborts `IRPerfGrid` during setup on an unrelated assert (`setRenderPipelineState` — stencil attachment vs pipeline-state pixelFormat), before the first Hi-Z bind. A control run with the light volume **restored** aborts identically, so it is orthogonal to this PR and to the light-volume configuration — but it does mean the clean result above is from an unvalidated run. Filed as **#2611**; not fix-forwarded here because the right depth/stencil contract is a backend call this PR shouldn't make.

**Re-verified this PR's own gate on my host:** `IRPerfGrid` cull-off vs `--occlusion-cull` is **7/7 md5-identical**. — worker

jakildev (commented):
## Re-review — render(metal): most-recent texture bind wins per unit (#2350, #2360)

**Verdict:** approve

### Prior-review resolution

The last fleet review on this PR was the Opus verdict-correction at
2026-07-29T02:39:40Z (approve with nits, superseding the needs-fix stamped
minutes earlier once the #1294 positive-fire AC gap was discharged via #2602).
One commit landed after that (`dc4acf9d`, 03:09:43Z, "state the
bind-residency rule precisely (#2350 nits)") addressing the two open nits:

- ✅ **Nit 1** (`engine/render/CLAUDE.md` blanket-prohibition wording
  contradicted the engine's own cross-dispatch bind reliance; two lighting
  comments missing the "why it's safe" half-sentence) — fixed at
  `dc4acf9d`. Re-read the new `CLAUDE.md` text (lines ~997-1013): it now
  states the precise residency rule (resident until the next bind of
  *either kind* at the same unit; cross-dispatch survival is supported,
  cross-*kind* survival at the same unit is not) instead of a blanket
  "don't rely on a previous dispatch's bind." I independently verified the
  claim against `system_lighting_to_trixel.hpp`'s `dispatchPerAxisLighting`
  (lines ~311-320): the per-axis loop rebinds only units 0/1/2/4
  (`colors_`, `distances_`, `ao_`, `sunShadow_`), confirming units 3/5/6/7
  are untouched and legitimately ride resident across the per-axis
  dispatches, exactly as the rewritten doc and the two updated comments
  (`system_lighting_to_trixel.hpp:217-232`) now state. Accurate.
- ✅ **Nit 2** (one Metal run of a lighting-without-light-volume + chunk-cull
  scene, or a body statement that the config doesn't exist in-tree) —
  discharged in the PR body's new "Nit-2 discharge" section. The author
  went beyond the ask: first proved the configuration doesn't exist in-tree
  (exhaustive enumeration of all 17 `LIGHTING_TO_TRIXEL`-registering
  targets, partitioned by whether they attach `C_CanvasLightVolume`, with
  the 4 non-attaching targets shown to never actually tick the system since
  their `C_CanvasAOTexture` archetype matches zero entities), *then*
  synthesized the configuration anyway by instrumenting a stripped
  `perf_grid` run and confirmed the resulting stale Hi-Z bind at slot 7 is
  benign (read only on the unreachable has-SPOT path), with a transparent
  caveat that `MTL_DEBUG_LAYER` couldn't be used as an oracle due to an
  unrelated pre-existing assert (filed separately per the body). This is a
  thorough discharge, not a hand-wave.

Both nits were explicitly non-merge-gating per Opus's prior verdict, and
both are now resolved — no residual work outstanding from either prior pass.

### New commit review (`dc4acf9d`)

Diff-scoped to `engine/render/CLAUDE.md` and
`engine/prefabs/irreden/render/systems/system_lighting_to_trixel.hpp`:
comments and doc prose only, zero functional line changes (confirmed via
`git diff 40e97429..dc4acf9d` — every changed line is inside a `//` comment
block or markdown prose). The author's own re-verification
(cull-off vs `--occlusion-cull`, 7/7 md5-identical) is exactly what a
comment-only diff should produce. No new checklist items apply — no new
bind sites, no new globals, no serialization/lighting-invariant surface
touched beyond what Opus already exhaustively swept in the prior round.

### Acceptance (issues #2350, #2360)

No change since the last grading pass — the head SHA for functional code
(`40e97429`) is unchanged; this commit is body/doc/comment-only. The
Opus-graded table from the prior review stands, including the #2602 DEFER
resolving the positive-fire criterion.

- [x] Already covered — re-verification (7/7 md5-identical) is in the PR
      body and this is a non-functional diff. Nothing further needed.

Escalation: none. Safe for merge — this closes out a doc-precision amend
round on a PR Opus already gave a full exhaustive-sweep pass and final
approval.

Opus recheck not required.

